PR #642, opened 2020-08-04
Add Pod Security Policy to deploy

ttpcodes:
This PR adds an additional file to `deploy/` that creates the necessary resources for Longhorn to be deployed with Pod Security Policy. This file creates a Pod Security Policy with the necessary privileges for Longhorn to operate correctly, along with the necessary Role and Role Binding for Longhorn's Service Accounts to use this policy.

This PR is based off of the Pod Security Policy that is proposed in longhorn/longhorn#1634.


yasker:
So it's not replacing the previous `privileged` flag?

ttpcodes:
The Pod Security Policy doesn't replace the need for the `privileged` flag, but rather allows the Pod to run as `privileged` in clusters that have Pod Security Policy enabled.

PR #647, opened 2020-08-11
controller, deploy, scheduler, types: support Data Locality option to keep a replica local to the engine

PhanLe1010:
### Migrate replica to the same node as engine: 
Give the user the option to keep a local replica to be on the same node as the volume engine. If a volume is attached to a node that has no replica, the Volume Controller will start rebuilding the replica on the node after the volume is attached. Once the rebuilding process is done, it will randomly remove one of the other replicas to keep the replica count as specified.

Currently, we support 2 data locality modes: `disabled` and `best-effort`. We allow users to change the global default data locality setting in Longhorn UI. The global setting should only function as a default value, like replica count. It doesn't change any existing volume's setting. We also allow users to change data locality setting for individual volume and expose the data locality setting as a parameter on the Storage Class

In the future, we might consider the modifying/extending scheduler approaches or adding `PV soft affinity` by summit a PR into Kubernetes upstream (See more discussion in the issue longhorn/longhorn#1045). However, we only move replica in v1.1.0

longhorn/longhorn#1045


controller/volume_controller.go
```diff
@@ -1141,10 +1142,120 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
...
+	if !hasLocalReplica {
+		if _, err := vc.addLocalReplicaOnSameNodeAsEngine(v, e, rs); err != nil {
+			return err
+		}
+		return nil
+	}
+
+	// check if there are have extra healthy replicas
+	healthyCount := 0
+	for rName, mode := range e.Status.ReplicaModeMap {
```
shuo-wu, line 1185: We can directly verify if a replica is healthy by checking `r.Spec.HealthyAt != "" && r.Spec.FailedAt  == ""`

controller/volume_controller.go
```diff
@@ -1141,10 +1142,120 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
...
+	}
+
+	return nil
+}
+
+// return true if successfully added, false otherwise
+func (vc *VolumeController) addLocalReplicaOnSameNodeAsEngine(v *longhorn.Volume, e *longhorn.Engine, rs map[string]*longhorn.Replica) (bool, error) {
+	// Create a new replica struct. We don't create a replica in DS using createReplica() directly
+	// because we may need to delete the new replica if it fails to ScheduleReplicaToNode.
+	// This prevent UI from repeatedly show creating/deleting the new replica
```
shuo-wu, line 1220: Do you mean a scheduling failed replica will be created each time when the volume is attached?

controller/volume_controller.go
```diff
@@ -1141,10 +1142,120 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
...
+		if keepLocalReplicaPolicy {
+			if err := vc.handleKeepLocalReplicaOnSameNodeAsEngine(v, e, rs); err != nil {
+				return err
+			}
+		}
 	}
 	return nil
 }
 
+func (vc *VolumeController) handleKeepLocalReplicaOnSameNodeAsEngine(v *longhorn.Volume, e *longhorn.Engine, rs map[string]*longhorn.Replica) error {
```
shuo-wu, line 1159: This function naming is a little bit weird. I think a simpler naming would be better. e.g., `ReconcileLocalReplica`

controller/volume_controller.go
```diff
@@ -1141,10 +1142,120 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
 		if oldState != v.Status.State {
 			vc.eventRecorder.Eventf(v, v1.EventTypeNormal, EventReasonAttached, "volume %v has been attached to %v", v.Name, v.Status.CurrentNodeID)
 		}
+
+		keepLocalReplicaPolicy, err := vc.ds.GetSettingAsBool(types.SettingNameKeepLocalReplicaOnSameNodeAsEngine)
```
shuo-wu, line 1146: You can put this setting check into the following function, then add extra checks and move function `handleKeepLocalReplicaOnSameNodeAsEngine` into the function `syncVolume` rather than `reconcileVolumeState`.

PhanLe1010:
@shuo-wu 
> I am thinking if we really need to actively delete the extra redundant replica after the local replica becoming running. Right now we will retain the replica when a volume is scaled up then scaled down.

I also thought about this and I think that we need to delete the extra redundant replica in this case. The reason is that pod can move around the cluster. Each time it moves to a new node, Longhorn will migrate a local replica to the node. In the long run, the pod possibly moves around the whole cluster and if we don't delete the extra replica, we will end up with having replica on all nodes. This destroys the purpose of having a local replica which is keeping the network usage in the cluster down. How do you think?

controller/volume_controller.go
PhanLe1010: thanks. Fixed

controller/volume_controller.go
PhanLe1010: Thanks. Fixed

controller/volume_controller.go
PhanLe1010: Thanks. Fixed

controller/volume_controller.go
PhanLe1010: Yeah. It is kind of. 

Assume that the local node is not available for scheduling replica. When the users turn on `KeepLocalReplicaOnSameNodeAsEngine`, Longhorn will repeatedly try to migrate a local replica.  

If we create a replica in DS using createReplica() directly, the following are repeatedly run:
1. create a new replica
2. schedule the new replica to the local node
3. if fail to schedule (in this case, it will fail because we assume local node is not available for scheduling), we have to delete the newly created replica because if we don't `ReconcileVolumeState` will schedule it to some other node latter on.

Step 1 and 3 cause a problem on Longhorn UI. The UI will show new replica immediately after step 1. Then UI deletes it immediately after step 3. The whole process is repeated again and again and users will see flashing on UI.

We can avoid this UI problem by not creating a replica in DS using createReplica() directly in step 1. Create an in-memory replica struct instead. As the result we don't have to do step 3 when fail to schedule.

shuo-wu: Got it.

shuo-wu:
> @shuo-wu
> 
> > I am thinking if we really need to actively delete the extra redundant replica after the local replica becoming running. Right now we will retain the replica when a volume is scaled up then scaled down.
> 
> I also thought about this and I think that we need to delete the extra redundant replica in this case. The reason is that pod can move around the cluster. Each time it moves to a new node, Longhorn will migrate a local replica to the node. In the long run, the pod possibly moves around the whole cluster and if we don't delete the extra replica, we will end up with having replica on all nodes. This destroys the purpose of having a local replica which is keeping the network usage in the cluster down. How do you think?

Yeah. It's better to remove the redundant replicas in this case. And it would be better to if we add priority for the replica removal. e.g., Remove the redundant replicas if there are more than 1 replicas on the same zone/node/disk.

shuo-wu:
We can also implement the `Kubernetes Scheduler Extender` for the new volumes after the setting is enabled. Then I can apply 2 modes for the new related pods: restrict data locality or best effort.

controller/volume_controller.go
```diff
@@ -1145,6 +1150,118 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
...
+	if !keepLocalReplicaPolicy {
+		return nil
+	}
+
+	if v.Status.State != types.VolumeStateAttached {
+		return nil
+	}
+
+	if e == nil {
+		return fmt.Errorf("BUG: handleKeepLocalReplicaWithEngine needs a valid engine")
```
shuo-wu, line 1167: `return nil`
IIRC, for the newly created volume, the related engine will be created in following `ReconcileVolumeState()`.

client/generated_longhorn_condition.go
```diff
@@ -1,87 +0,0 @@
-package client
-
-const (
-	LONGHORN_CONDITION_TYPE = "longhornCondition"
-)
-
-type LonghornCondition struct {
```
joshimoo, line 7: Just double checking, we don't need this file anymore?

manager/volume.go
```diff
@@ -698,3 +711,29 @@ func (m *VolumeManager) UpdateReplicaCount(name string, count int) (v *longhorn.
...
+
+	v, err = m.ds.GetVolume(name)
+	if err != nil {
+		return nil, err
+	}
+
+	oldDataLocalityMode := v.Spec.DataLocality
+	v.Spec.DataLocality = dataLocalityMode
+
+	v, err = m.ds.UpdateVolume(v)
```
joshimoo, line 732: consider making it a noop when the current data locality setting is the same as the requested.

types/setting.go
```diff
@@ -267,6 +270,24 @@ var (
 		Default:     "3",
 	}
 
+	SettingDefinitionDefaultDataLocality = SettingDefinition{
+		DisplayName: "Default Data Locality",
+		Description: "We say a Longhorn volume has data locality if there is a local replica of the volume on the same node as the pod which is using the volume.\n\n" +
+			"This setting specifies the default data locality mode when a volume is created from the Longhorn UI. For Kubernetes configuration, update the `dataLocality` in the StorageClass\n\n" +
+			"The available data locality modes are: \n\n" +
+			"- **disable** is the default mode. There may be a local replica of the volume on the same node as the consuming pod or there may be not. Longhorn doesn't do anything\n" +
```
joshimoo, line 278: I think it's `may not be` instead of `may be not` ?

controller/volume_controller.go
```diff
@@ -1026,6 +1031,114 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
 	return nil
 }
 
+func (vc *VolumeController) ReconcileLocalReplica(v *longhorn.Volume, e *longhorn.Engine, rs map[string]*longhorn.Replica) error {
+	if v.Spec.DataLocality != types.DataLocalityBestEffort {
```
joshimoo, line 1035: Consider checking the disable case (== disable), since you are foreseeing an enforced case. And there will only ever be 1 disable case.

controller/volume_controller.go
```diff
@@ -1026,6 +1031,114 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
...
+
+	if e == nil {
+		return fmt.Errorf("BUG: ReconcileLocalReplica needs a valid engine")
+	}
+
+	if !vc.isSafeToModifyReplica(v, e, rs) {
+		return nil
+	}
+
+	// hasLocalReplica indicate whether there exist a replica on the same node as engine
```
joshimoo, line 1051: add private util func for this

controller/volume_controller.go
```diff
@@ -1026,6 +1031,114 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
...
+	hasLocalReplica := false
+	for _, r := range rs {
+		if r.Spec.NodeID == e.Spec.NodeID {
+			hasLocalReplica = true
+		}
+	}
+
+	if !hasLocalReplica {
```
joshimoo, line 1059: call util func here, since doesn't seem that you depend on this variable anywhere else.

controller/volume_controller.go
```diff
@@ -1026,6 +1031,114 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
...
+		if _, err := vc.addLocalReplicaOnSameNodeAsEngine(v, e, rs); err != nil {
+			return err
+		}
+		return nil
+	}
+
+	// check if there are have extra healthy replicas
```
joshimoo, line 1066:
```suggestion
	// check if there are to many healthy replicas
```

controller/volume_controller.go
```diff
@@ -1026,6 +1031,114 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
...
+	}
+
+	if healthyCount <= v.Spec.NumberOfReplicas {
+		return nil
+	}
+
+	// Randomly delete extra non-local healthy replicas
+	// Sometime ReconcileLocalReplica() is called more than once with the same input (v,e,rs).
+	// To make the deleting operation idempotent and prevent deleting more replica than needed,
+	// we always delete the replica with the smallest name. This doesn't destroy the randomness nature of the
```
joshimoo, line 1081:
```suggestion
	// we always delete the replica with the smallest name. This doesn't destroy the random nature of the
```

controller/volume_controller.go
```diff
@@ -1026,6 +1031,114 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
...
+	// deleting operation because replica's name contains a random hashcode.
```
joshimoo, line 1082:
```suggestion
	// deletion operation because each replica's name contains a random hashcode.
```

controller/volume_controller.go
```diff
@@ -1026,6 +1031,114 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
...
+		return false, nil
+	}
+
+	// save the new replica to DS, the new replica will be started by ReconcileVolumeState() later
+	r, err := vc.ds.CreateReplica(scheduledReplica)
+	if err != nil {
+		return false, err
+	}
+	rs[r.Name] = r
+	logrus.Infof("Added local replica %v of volume %v on the same node as engine. Engine is on node: %v", r.Name, v.Name, e.Spec.NodeID)
```
joshimoo, line 1122:
```suggestion
	logrus.Infof("Added local replica %v of volume %v on the same node %v as engine.", r.Name, v.Name, e.Spec.NodeID)
```

joshimoo (commented):
in principle looks good to me, added some suggestions.

client/generated_longhorn_condition.go
PhanLe1010: This code is auto-generated by https://github.com/longhorn/longhorn-manager/blob/master/client/README.md
After following the above step, this file is gone. I am not sure if it is a problem.
How do you think?  @shuo-wu

controller/volume_controller.go
PhanLe1010: I did the same initially but I observed a problem when upgrading Longhorn. Some volumes may not have field `v.Spec.DataLocality` and because I checked for `v.Spec.DataLocality == types.DataLocalityDisable`, those volumes still pass the check even though they don't have `v.Spec.DataLocality==types.DataLocalityBestEffort`

controller/volume_controller.go

controller/volume_controller.go

controller/volume_controller.go

controller/volume_controller.go

controller/volume_controller.go

controller/volume_controller.go

types/setting.go
PhanLe1010: Thank you! Fixed

manager/volume.go
PhanLe1010: Thank you! Fixed

controller/volume_controller.go
```diff
@@ -1026,6 +1031,116 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
...
+		if r.Spec.FailedAt == "" && r.Spec.HealthyAt != "" {
+			healthyCount++
+		}
+	}
+
+	if healthyCount <= v.Spec.NumberOfReplicas {
+		return nil
+	}
+
+	// Randomly delete extra non-local healthy replicas
```
shuo-wu, line 1070: Can we make this cleanup more intelligent? e.g., if soft anti-affinity is enabled and there are multiple replicas on the same node, deleting one of those replicas would be better

controller/volume_controller.go
```diff
@@ -1504,6 +1619,28 @@ func (vc *VolumeController) createReplica(v *longhorn.Volume, e *longhorn.Engine
 	return vc.ds.CreateReplica(replica)
 }
 
+func (vc *VolumeController) createReplicaStruct(v *longhorn.Volume, e *longhorn.Engine, rs map[string]*longhorn.Replica) *longhorn.Replica {
```
shuo-wu, line 1622: will it be better to rename this to `CreateReplicaManifest`?

shuo-wu (changes requested):
It seems that the new field `v.Spec.DataLocality` can be empty. Is it acceptable?

controller/volume_controller.go
joshimoo: Good point, add a private utility function isDataLocalityDisabled that treats the empty spec as disabled :)

joshimoo: Good point add a private utility function isDataLocalityDisabled that treats the empty case as disabled.
Then call it here and any other place where you explicitly check.

controller/volume_controller.go
joshimoo: good name suggestion be careful on the casing, this should be private.
ie `createReplicaManifest`

joshimoo (changes requested):

controller/volume_controller.go
shuo-wu: Oh. Sorry for the typo.

types/resource.go
```diff
@@ -28,6 +28,13 @@ const (
 	VolumeFrontendEmpty    = VolumeFrontend("")
 )
 
+type DataLocality string
+
+const (
+	DataLocalityDisable    = DataLocality("disable")
```
yasker, line 34: `disabled` instead of `disable`

manager/volume.go
```diff
@@ -190,6 +190,18 @@ func (m *VolumeManager) Create(name string, spec *types.VolumeSpec) (v *longhorn
...
 	}
+
+	if string(spec.DataLocality) == "" {
+		defaultDataLocalityMode, err := m.GetSettingValueExisted(types.SettingNameDefaultDataLocality)
+		if err != nil {
+			return nil, errors.Wrapf(err, "cannot get valid mode for setting default data locality mode for volume: %v", name)
+		}
+		spec.DataLocality = types.DataLocality(defaultDataLocalityMode)
+	}
+	if err := types.ValidateDataLocalityMode(spec.DataLocality); err != nil {
```
yasker, line 201: Need to be consistent with the name, e.g. DataLocality instead of DataLocalityMode.

`Locality` is a noun so you don't need to mention `mode` every time.

csi/util.go
```diff
@@ -66,6 +67,13 @@ func getVolumeOptions(volOptions map[string]string) (*longhornclient.Volume, err
 		vol.NumberOfReplicas = int64(nor)
 	}
 
+	if dataLocalityMode, ok := volOptions["dataLocality"]; ok {
```
yasker, line 70: `if locality, ok :=` is good enough.

controller/volume_controller.go
PhanLe1010: Thank you! Fixed

api/model.go
```diff
@@ -21,31 +21,30 @@ import (
 
 type Volume struct {
 	client.Resource
-
```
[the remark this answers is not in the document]
PhanLe1010: Thank you! Fixed

controller/volume_controller.go
PhanLe1010: Thank you! I will try this

controller/volume_controller.go
PhanLe1010: Thank you! Fixed

controller/volume_controller.go
```diff
@@ -1026,6 +1031,116 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
 	return nil
 }
 
+func (vc *VolumeController) ReconcileLocalReplica(v *longhorn.Volume, e *longhorn.Engine, rs map[string]*longhorn.Replica) error {
```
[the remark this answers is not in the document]
PhanLe1010: Thank you! Fixed

csi/util.go
PhanLe1010: Thank you! Fixed

manager/volume.go
PhanLe1010: Thank you! Fixed

types/resource.go
PhanLe1010: Thank you! Fixed

controller/volume_controller.go
PhanLe1010: @yasker How do you think about this?

yasker: Yes, we should do this. Though I cannot think of a case other than soft anti-affinity we need to cover.

PhanLe1010: So, we don't care about zone and we just consider replicas on the same node/disk as better candidates for deletion?

PhanLe1010: @shuo-wu @yasker I updated replica cleanup logic. How do you think about this logic?

yasker: LGTM.

controller/volume_controller.go
```diff
@@ -1026,6 +1031,176 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
...
+		return deletionCandidates, nil
+	}
+
+	// if all replicas are on different nodes, prefer to delete replicas on the same zone
+	deletionCandidates = findValueWithBiggestLength(zoneToReplicaMap)
+	if len(deletionCandidates) > 1 {
+		return deletionCandidates, nil
+	}
+
+	// if all replicas are on different zones, return all replicas' names in the input RS
```
shuo-wu, line 1129:
```
if all replicas are on same zones
```

shuo-wu (changes requested):

client/generated_longhorn_condition.go
shuo-wu: It shouldn't be a problem then.

PhanLe1010:
> For the old volumes, it seems that the new field `v.Spec.DataLocality` can be empty. Is it acceptable?

@shuo-wu It should be ok. When `v.Spec.DataLocality`  is empty, we treat it the same as having `v.Spec.DataLocality=disabled`

controller/volume_controller.go
PhanLe1010: I am not sure I understand this. Should it be `if all replicas are on different zones` because we checked for replicas in the same zone before come to this?

shuo-wu: Oh sorry for the misunderstanding... This should be good.

shuo-wu (approved):

yasker:
There is a conflict that needs to be fixed.

PhanLe1010:
@yasker The conflict is resolved

yasker:
I assume @joshimoo is OK with merging this since all the issues raised have been solved.

joshimoo (approved):
LGTM, sorry for the hold up.

PR #655, opened 2020-08-18
Added comments to exported functions and methods in `datastore/uncached.go`

c3y1huang:
Added comments to exported functions and methods.

### DIFF
--- a/datastore/uncached.go
+++ b/datastore/uncached.go
@@ -6,57 +6,70 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
+// GetLonghornEventList returns a list of longhorn events for the given namespace
 // Event list is uncached but current lister doesn't field selector, so we have no choice but read from API directly
-
 func (s *DataStore) GetLonghornEventList() (*corev1.EventList, error) {
 	return s.kubeClient.CoreV1().Events(s.namespace).List(metav1.ListOptions{FieldSelector: "involvedObject.apiVersion=longhorn.io/v1beta1"})
 }
 
 // The following APIs are not cached, should only be used in e.g. the support bundle
 
+// GetAllPodsList returns a list of pods for the given namespace
 func (s *DataStore) GetAllPodsList() (runtime.Object, error) {
 	return s.kubeClient.CoreV1().Pods(s.namespace).List(metav1.ListOptions{})
 }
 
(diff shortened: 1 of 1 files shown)


datastore/uncached.go
```diff
@@ -6,57 +6,70 @@ import (
...
+// GetLonghornEventList returns a list of longhorn events for the given namespace
 // Event list is uncached but current lister doesn't field selector, so we have no choice but read from API directly
-
 func (s *DataStore) GetLonghornEventList() (*corev1.EventList, error) {
 	return s.kubeClient.CoreV1().Events(s.namespace).List(metav1.ListOptions{FieldSelector: "involvedObject.apiVersion=longhorn.io/v1beta1"})
 }
 
 // The following APIs are not cached, should only be used in e.g. the support bundle
 
+// GetAllPodsList returns a list of pods for the given namespace
```
joshimoo, line 17:
```suggestion
// GetAllPodsList returns an uncached list of pods for the given namespace directly from the api server.
// Using cached informers should be prefered and direct retrieval from the api server should only be used for one shot tasks for example support bundle creation.
```

@c3y1huang Thank you for the PR's it's appreciated 👍 
Would you mind adding more details (Why, Why, When, etc) for example to the comments, like in the above suggestion?
Currently your comments only describe the `What` part which is self explanatory based on the Function name.

c3y1huang: Thank you for the review and suggestion. I've addressed in https://github.com/longhorn/longhorn-manager/pull/655/commits/05580128f7e6ad437ded4cd3ddddcc40d59e8b49, PTAL.

yasker:
Thanks for the PR @c3y1huang. I noticed there are four similar PRs for the `datastore`. Can we aggregate them into one PR? We can have multiple commits in the PR, one for each file.

Also, we need to squash the commits for the same functionality (e.g. two commits in this commit) into one commit.

c3y1huang:
> Thanks for the PR @c3y1huang. I noticed there are four similar PRs for the `datastore`. Can we aggregate them into one PR? We can have multiple commits in the PR, one for each file.
> 
> Also, we need to squash the commits for the same functionality (e.g. two commits in this commit) into one commit.

Sure, closing this and supersede by https://github.com/longhorn/longhorn-manager/pull/658.

PR #661, opened 2020-08-26
Refactor to add support for instance based logging

joshimoo:
This adds a base controller that takes care of setting up the initial instance based logger so that we always have the controller name and node fields.

longhorn/longhorn#1724
longhorn/longhorn#1736

### DIFF
--- a/app/uninstall.go
+++ b/app/uninstall.go
@@ -104,8 +104,11 @@ func uninstall(c *cli.Context) error {
 		persistentVolumeClaimInformer, kubeNodeInformer, priorityClassInformer,
 		kubeClient, namespace)
 
+	logger := logrus.StandardLogger()
+
 	doneCh := make(chan struct{})
 	ctrl := controller.NewUninstallController(
+		logger,
 		namespace,
 		c.Bool(FlagForce),
 		ds,
--- a/controller/controller_manager.go
+++ b/controller/controller_manager.go
@@ -80,6 +80,8 @@ func StartControllers(stopCh chan struct{}, controllerID, serviceAccount, manage
 	volumeAttachmentInformer := kubeInformerFactory.Storage().V1beta1().VolumeAttachments()
 	priorityClassInformer := kubeInformerFactory.Scheduling().V1().PriorityClasses()
 
+	logger := logrus.StandardLogger()
+
 	ds := datastore.NewDataStore(
 		volumeInformer, engineInformer, replicaInformer,
 		engineImageInformer, nodeInformer, settingInformer, imInformer,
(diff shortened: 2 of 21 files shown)


controller/generic_controller.go
```diff
@@ -0,0 +1,28 @@
...
+
+import (
+	"github.com/sirupsen/logrus"
+	"k8s.io/client-go/util/workqueue"
+)
+
+type genericController struct {
+	name   string
+	logger logrus.FieldLogger
+	queue  workqueue.RateLimitingInterface
```
yasker, line 11: Why do you decide to include the queue here? I don't think this is aiming to be a controller framework, but a logger for the controllers.

controller/engine_image_controller_test.go
```diff
@@ -73,7 +74,10 @@ func newTestEngineImageController(lhInformerFactory lhinformerfactory.SharedInfo
 		persistentVolumeClaimInformer, kubeNodeInformer, priorityClassInformer,
 		kubeClient, TestNamespace)
 
+	// TODO: [#1724] improve the logging setup/infrastructure further down the line
```
yasker, line 77: Do we need this `TODO` line here?

controller/controller_manager.go
```diff
@@ -81,7 +81,7 @@ func StartControllers(stopCh chan struct{}, controllerID, serviceAccount, manage
 	priorityClassInformer := kubeInformerFactory.Scheduling().V1().PriorityClasses()
 
 	// TODO: [#1724] improve the logging setup/infrastructure further down the line
-	logger := logrus.StandardLogger()
+	logger := logrus.StandardLogger().WithField("node", controllerID)
```
yasker, line 84: I am debating if we should include it here since every manager is running on different nodes so the information is embedded in the manager's name. But on the other side, searching will be easier with the nodeID.

joshimoo: having the dedicated node field for each log output, seems like a good idea?
I don't see no harm in this and it will make it easier to correlate logs from different components.

controller/engine_image_controller_test.go
joshimoo: I will remove the TODOs

controller/generic_controller.go
joshimoo: Clarified during discussion that the idea is not to build a controller framework.
This can be conceptually thought of an abstract base class, the reason to include the queue is to keep the name in sync.
I will rename the file to `base-controller` instead :)

app/uninstall.go
```diff
@@ -104,8 +104,12 @@ func uninstall(c *cli.Context) error {
 		persistentVolumeClaimInformer, kubeNodeInformer, priorityClassInformer,
 		kubeClient, namespace)
 
+	// TODO: [#1724] improve the logging setup/infrastructure further down the line
```
yasker, line 107: Still has the TODO comment?

Same below.

controller/engine_controller.go
```diff
@@ -107,6 +108,8 @@ func NewEngineController(
 	eventBroadcaster.StartRecordingToSink(&v1core.EventSinkImpl{Interface: v1core.New(kubeClient.CoreV1().RESTClient()).Events("")})
 
 	ec := &EngineController{
+		baseController: newBaseController("longhorn-engine", logger),
```
yasker, line 111: just call it `engine` instead of `longhorn-engine`. It's an unnecessary prefix since it's produced by the Longhorn manager.

Same below.

shuo-wu (commented):
I think we need to use replace the log invocation like `logrus.Infof()` with `logger.Infof()` in those controllers then?

controller/engine_controller.go
joshimoo: Agree, I kept the prior names but did think that the prefix is unnecessary.

joshimoo:
@shuo-wu see message in channel also there is an epic  #1724 that links to issues for each controller refactor.
We want to leave these for the moment, since they are excellent choices for first time contributors.

controller/engine_controller.go
yasker: After discussion with @joshimoo , I realize the controller name is also used in the event recorder, which might cause confusion if we remove the `longhorn-` prefix it that. So I am good with keeping the `longhorn-` prefix and keep controller name same across different logging fields.

joshimoo:
addressed issues and rebased.

yasker:
Done

PR #662, opened 2020-08-26
Refactor engine controller to use structured logging

joshimoo:
By switching to structured logging and removing the common element
information from the individual log messages it becomes much simpler to
correlate a log statement with some code flow. The structured log fields
also allow us to easily filter logs based on specific element keys.

depends on longhorn/longhorn-manager#661
longhorn/longhorn#1724
longhorn/longhorn#1725

### DIFF
--- a/controller/engine_controller.go
+++ b/controller/engine_controller.go
@@ -78,6 +78,8 @@ type EngineController struct {
 }
 
 type EngineMonitor struct {
+	logger logrus.FieldLogger
+
 	namespace     string
 	ds            *datastore.DataStore
 	eventRecorder record.EventRecorder
@@ -163,8 +165,8 @@ func (ec *EngineController) Run(workers int, stopCh <-chan struct{}) {
 	defer utilruntime.HandleCrash()
 	defer ec.queue.ShutDown()
 
-	logrus.Infof("Start Longhorn engine controller")
-	defer logrus.Infof("Shutting down Longhorn engine controller")
+	ec.logger.Info("Start Longhorn engine controller")
+	defer ec.logger.Info("Shutting down Longhorn engine controller")
 
 	if !controller.WaitForCacheSync("longhorn engines", stopCh, ec.eStoreSynced, ec.imStoreSynced) {
 		return
@@ -202,14 +204,15 @@ func (ec *EngineController) handleErr(err error, key interface{}) {
 		return
 	}
 
+	log := ec.logger.WithField("engine", key)
 	if ec.queue.NumRequeues(key) < maxRetries {
-		logrus.Warnf("Error syncing Longhorn engine %v: %v", key, err)
+		log.WithError(err).Warn("Error syncing Longhorn engine")
 		ec.queue.AddRateLimited(key)
 		return
 	}
 
 	utilruntime.HandleError(err)
-	logrus.Warnf("Dropping Longhorn engine %v out of the queue: %v", key, err)
+	log.WithError(err).Warn("Dropping Longhorn engine out of the queue")
 	ec.queue.Forget(key)
 }
 
@@ -226,10 +229,11 @@ func (ec *EngineController) syncEngine(key string) (err error) {
 		return nil
 	}
 
+	log := ec.logger.WithField("engine", name)
 	engine, err := ec.ds.GetEngine(name)
 	if err != nil {
 		if datastore.ErrorIsNotFound(err) {
-			logrus.Infof("Longhorn engine %v has been deleted", key)
+			log.Info("Engine has been deleted")
 			return nil
 		}
 		return err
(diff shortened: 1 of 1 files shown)


joshimoo:
This depends on #661 and I will rebase this once #661 has been merged.

controller/engine_image_controller_test.go
```diff
@@ -2,6 +2,7 @@ package controller
 
 import (
 	"fmt"
+	"github.com/sirupsen/logrus"
```
yasker, line 5: Add a blank.

joshimoo:
Please only review the last commit (engine-controller) the other code is covered by #661 (please add review comment there instead) and will not be part of this pr once #661 is committed

controller/engine_controller.go
```diff
@@ -562,6 +564,7 @@ func (ec *EngineController) isMonitoring(e *longhorn.Engine) bool {
 func (ec *EngineController) startMonitoring(e *longhorn.Engine) {
 	stopCh := make(chan struct{})
 	monitor := &EngineMonitor{
+		logger:           ec.logger.WithField("engine", e.Name),
```
shuo-wu, line 567: Would `engine-monitor` be better?

controller/engine_controller.go
```diff
@@ -607,8 +610,8 @@ func (ec *EngineController) stopMonitoring(e *longhorn.Engine) {
 }
 
 func (m *EngineMonitor) Run() {
-	logrus.Debugf("Start monitoring %v", m.Name)
-	defer logrus.Debugf("Stop monitoring %v", m.Name)
+	m.logger.Debug("Start monitoring engine")
```
shuo-wu, line 613: `Start monitoring`/`Stop monitoring` would be enough.

controller/engine_controller.go
```diff
@@ -911,7 +914,7 @@ func syncWithRestoreStatus(engine *longhorn.Engine, rsMap map[string]*types.Rest
 
 	for _, status := range rsMap {
 		if status.Error != "" {
-			logrus.Warnf("Need to wait for the restore error handling for engine %v before the restore invocation: %v", engine.Name, status.Error)
+			log.Warnf("need to wait for the restore error handling for engine %v before the restore invocation: %v", engine.Name, status.Error)
```
shuo-wu, line 917:
```
log.Warnf("need to wait for the restore error handling before the restore invocation: %v", status.Error)
```

controller/engine_controller.go
```diff
@@ -1017,13 +1021,13 @@ func checkSizeBeforeRestoration(engine *longhorn.Engine, ds *datastore.DataStore
...
 			// The volume controller will update `engine.Spec.VolumeSize` later then trigger expansion call
-			logrus.Infof("engine monitor: Prepare to expand the size from %v to %v for DR volume %v", v.Spec.Size, bvSize, v.Name)
+			log.Infof("Prepare to expand the size from %v to %v for DR volume %v", v.Spec.Size, bvSize, v.Name)
 			v.Spec.Size = bvSize
 			if _, err := ds.UpdateVolume(v); err != nil {
 				if !datastore.ErrorIsConflict(err) {
 					return false, err
 				}
-				logrus.Debugf("engine monitor: Retrying update the volume %v size before restore", v.Name)
+				log.Debugf("Retrying size update for DR volume %v before restore", v.Name)
```
shuo-wu, line 1030: Can we use `WithField("volume", e.Spec.VolumeName)`? Similar below.

shuo-wu (changes requested):

controller/engine_controller.go
joshimoo: yes, we can add it as a field.

controller/engine_controller.go
joshimoo: It's better to have `Start monitoring engine` where engine is defined as a field.
This way a search in the logs will be specific.

If you use a generic `start monitoring` in all the different components you will most likely end up with other components.
(ie instance-manager, backupstore, etc)

controller/engine_controller.go
joshimoo: The object we are concerned with is a `engine` not the monitor.

joshimoo: Example, log print from below to make it more clear what it mean by that.
`log.Warnf("need to wait for the restore error handling before the restore invocation: %v", status.Error)`
this is concerned with a `engine` not a monitor, the monitor itself is only an internal construction that is not exposed to the outside.

shuo-wu: Why is the `engine` preferred?

joshimoo: See comment above?
We want to correlate actions across components that are involved with the `engine` object.

Maybe this helps?
It wouldn't make much sense calling it `engine-monitor` here then in the longhorn-engine/instance-manager project the field would be called `engine`?

joshimoo: Let me flip the question, why do you think `engine-monitor` would be better?

shuo-wu: Got it! Sorry for missing the comment, since the page was not refreshed when I replied to you.

shuo-wu: I was thinking the original logger is based on the controller, hence this derived logger should base on the monitor.

joshimoo:
addressed issues and rebased.

joshimoo:
Rebased and Ready for review, added some more withError calls.

shuo-wu (approved):

PR #663, opened 2020-08-27
Make daemon support IPv6 adresses

Frankkkkk:
Actually the longhorn manager doesn't support well IPv6 adresses. For
example, in an IPv6-only k8s cluster we can see the following logs:

>  level=info msg="Listening on 2001:db8:caca:8273:187:4:ffff:4f5a:9500

which is missing brackets. The pod fails its readiness probe and never
becomes available.

PS: please excuse me for any fuckups. It's actually my first time writing a line in Go.
Cheers

### DIFF
--- a/types/types.go
+++ b/types/types.go
@@ -3,6 +3,7 @@ package types
 import (
 	"encoding/json"
 	"fmt"
+	"net"
 	"os"
 	"path/filepath"
 	"regexp"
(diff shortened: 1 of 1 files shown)


yasker:
Thanks @Frankkkkk . We don't have an IPv6 infrastructure to validate all the fixes at the moment, but as long as it doesn't impact the IPv4, we can merge them.

joshimoo:
@Frankkkkk thank you for the contribution, it's appreciated :)

PR #664, opened 2020-08-27
controller: Clean up extra failed to scheduled replica when no eviction requested

boknowswiki:
This will clean up the extra failed to scheduled replica when there is
no eviction requested and the healthy replicas count is not less then
the volume spec replica numbers.

https://github.com/longhorn/longhorn/issues/1679

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -570,13 +570,18 @@ func (vc *VolumeController) ReconcileEngineReplicaState(v *longhorn.Volume, e *l
 }
 
 func (vc *VolumeController) cleanupCorruptedOrStaleReplicas(v *longhorn.Volume, rs map[string]*longhorn.Replica) error {
-	hasHealthyReplicas := false
+	var healthyCount int
+	hasEvictionRequestedReplicas := false
+
 	for _, r := range rs {
 		if r.Spec.FailedAt == "" && r.Spec.HealthyAt != "" {
-			hasHealthyReplicas = true
-			break
+			healthyCount++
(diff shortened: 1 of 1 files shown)


controller/volume_controller.go
```diff
@@ -592,8 +597,18 @@ func (vc *VolumeController) cleanupCorruptedOrStaleReplicas(v *longhorn.Volume,
 		}
 
 		if r.Spec.FailedAt == "" {
+			// Cleanup extra failed to scheduled replica
+			// when no eviction requested
+			if healthyCount >= v.Spec.NumberOfReplicas &&
```
shuo-wu, line 602: How about using the following logic:
```
// skip running replicas
if r.Spec.HealthyAt != "" {
    continue
}
// when the volume is not Healthy, a new replica will be still recreated if there is a cleanup
if healthyCount < v.Spec.NumberOfReplicas {
    continue
}
// avoid the clean up during the eviction
if hasEvictionRequestedReplicas {
    continue
}
```

Then you don't need to delete the replica here.

boknowswiki: Following your logic, I prefer to put them together like this:
`
//comments here
if r.Spec.HealthyAt != "" || healthyCount < v.Spec.NumberOfReplicas || hasEvictionRequestedReplicas {
        continue
}
`
what do you think?

boknowswiki: Updated this part, please take a look. Thanks.

shuo-wu (approved):

PR #665, opened 2020-08-27
types: fix a typo in Default Data Locality Setting

PhanLe1010:
longhorn/longhorn#1045

### DIFF
--- a/types/setting.go
+++ b/types/setting.go
@@ -278,8 +278,8 @@ var (
 		Description: "We say a Longhorn volume has data locality if there is a local replica of the volume on the same node as the pod which is using the volume.\n\n" +
 			"This setting specifies the default data locality when a volume is created from the Longhorn UI. For Kubernetes configuration, update the `dataLocality` in the StorageClass\n\n" +
 			"The available modes are: \n\n" +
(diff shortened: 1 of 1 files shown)


types/setting.go
```diff
@@ -278,7 +278,7 @@ var (
 		Description: "We say a Longhorn volume has data locality if there is a local replica of the volume on the same node as the pod which is using the volume.\n\n" +
 			"This setting specifies the default data locality when a volume is created from the Longhorn UI. For Kubernetes configuration, update the `dataLocality` in the StorageClass\n\n" +
 			"The available modes are: \n\n" +
-			"- **disable** is the default mode. There may be a local replica of the volume on the same node as the consuming pod or there may not be. Longhorn doesn't do anything\n" +
+			"- **disabled** is the default mode. There may be a local replica of the volume on the same node as the consuming pod or there may not be. Longhorn doesn't do anything\n" +
```
yasker, line 281: `- **disabled**. This is the default option. There may or may not be a replica on the same node as the attached volume (workload)\n`

`- **best-effort**. This option instructs Longhorn to try to keep a replica on the same node as the attached volume (workload). Longhorn will not stop the volume, even if it cannot keep a replica local to the attached volume (workload) due to environment limitation, e.g. not enough disk space, incompatible disk tags, etc.\n`

PhanLe1010: Thank you! I fixed it

PR #671, opened 2020-09-01
Add role/cluster binding for nfs-provisioner

khushboo-rancher:
https://github.com/longhorn/longhorn/issues/1606

I'm not sure about ```service``` ```longhorn-nfs-provisioner``` that it should be included here or not.

Signed-off-by: khushboo-rancher <60111667+khushboo-rancher@users.noreply.github.com>

### DIFF
--- a/deploy/install/01-prerequisite/02-serviceaccount.yaml
+++ b/deploy/install/01-prerequisite/02-serviceaccount.yaml
@@ -56,3 +56,59 @@ subjects:
 - kind: ServiceAccount
   name: longhorn-service-account
   namespace: longhorn-system
+- kind: ServiceAccount
+  name: longhorn-nfs-provisioner
+  namespace: longhorn-system
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: longhorn-nfs-provisioner
+  namespace: longhorn-system
+---
+kind: Service
+apiVersion: v1
+metadata:
+  name: longhorn-nfs-provisioner
+  namespace: longhorn-system
+  labels:
+    app: longhorn-nfs-provisioner
+spec:
+  # hardcode a cluster ip for the service
+  # so that on delete & recreate of the service the previous pv's still point
+  # to this nfs-provisioner, pick a new ip for each new nfs provisioner
+  clusterIP: 10.43.111.111
+  ports:
+    - name: nfs
+      port: 2049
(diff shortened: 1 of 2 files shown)


shuo-wu:
The RBAC is already in `examples/rwx`. Maybe we can put the enhanced PSP in that directory if the existing PSP is editable.

khushboo-rancher:
As per my understanding, going forward to address this https://github.com/longhorn/longhorn/issues/1470, we can modify the existing podsecuritypolicy.yaml and add serviceAccount ```longhorn-nfs-provisioner```.

@joshimoo Any thoughts?

deploy/install/01-prerequisite/02-serviceaccount.yaml
```diff
@@ -56,3 +56,59 @@ subjects:
...
+  name: longhorn-nfs-provisioner
+  namespace: longhorn-system
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: longhorn-nfs-provisioner
+  namespace: longhorn-system
+---
+kind: Service
```
joshimoo, line 69: Service does not belong here.

joshimoo (changes requested):
The service account change looks good, but since we anyway give the PSP the permissions.
We might just want to only use a single service-account @yasker ?

From a security perspective it's always better to have lowest common security profiles, but we don't really do that at the moment, so it might be okay to just have a single account for now.

yasker:
I am fine with using the same service account.

And the service part should be in the NFS provisioner deployment, at least before we have the HA RWX.

yasker:
Also, need to update the sign-off-by.

yasker:
Since we're merging the PSP/role binding from `examples/rwx`, we should also update the `examples/rwx` to use the Longhorn service account, etc. Otherwise, this change has no users.

khushboo-rancher:
Discussed with @yasker, we need to change the ```example/rwx``` and doc of rwx too. We can wait till the story ```HA rwx``` is delivered and then merge this to have all the relevant changes at one time.

joshimoo:
Closing this, since the changes are no longer necessary.
Once the native rwx support is merged, https://github.com/longhorn/longhorn-manager/pull/778
We can then remove the rwx provisioner example, and transition to using the native rwx support.

PR #672, opened 2020-09-01
Refact replica clean up with supporting data-locality and eviction requested

boknowswiki:
1. Revert back the replica clean up for corrupted or staled case.
    2. Refactor replica clean up for failed to scheduled case.
    3. Add replica clean up support for data-locality enabled and eviction requested case.

https://github.com/longhorn/longhorn/issues/1679

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -395,7 +395,7 @@ func (vc *VolumeController) syncVolume(key string) (err error) {
 		return err
 	}
 
-	if err := vc.cleanupCorruptedOrStaleReplicas(volume, replicas); err != nil {
+	if err := vc.cleanupReplicas(volume, engine, replicas); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


controller/volume_controller.go
```diff
@@ -604,9 +604,11 @@ func (vc *VolumeController) cleanupCorruptedOrStaleReplicas(v *longhorn.Volume,
 			// Skip running healthy replicas.
 			// Skip degraded volume.
 			// Skip if eviction is going.
+			// Skip if datalocality has been enabled
 			if r.Spec.HealthyAt != "" ||
 				healthyCount < v.Spec.NumberOfReplicas ||
-				hasEvictionRequestedReplicas {
+				hasEvictionRequestedReplicas ||
+				!isDataLocalityDisabled(v) {
```
yasker, line 611: Should we limit to the replica that exists on the same attached node of v? cc @PhanLe1010

shuo-wu:
Not sure why we need this PR? Can you clarify what happens (without this PR) if the data locality is enabled?

PhanLe1010:
Right now, the PR only delete extra healthy replica when EvictionRequested or DataLocality are on. Should we also delete extra healthy replica in normal cases to keep a consistent behavior as discussed with @yasker and @boknowswiki yesterday?

boknowswiki:
The document update has been submitted on the [same PR as 298 document PR](https://github.com/longhorn/website/pull/186).

boknowswiki:
> Right now, the PR only deletes extra healthy replica when EvictionRequested or DataLocality are on. Should we also delete extra healthy replica in normal cases to keep a consistent behavior as discussed with @yasker and @boknowswiki yesterday?

I've thought about that as well. To make sure we are talking about the scenario.
User case with 3 worker nodes cluster, not regarding to eviction and data-locality.
1. Create a volume with 3 replicas. This will succeed.
2. User change the replica number to 2. In this case, there is one extra healthy replica.

Do you mean we should remove this extra replica? @PhanLe1010 

I think we should not remove it automatically. User may have plan with it, because we don't know which one user may want to delete or user may just want to leave it there. Since this extra replica is not driven by us, we should not touch it.

Please let me know if my understand is not correct or you have different ideas. Thanks.

PhanLe1010:
@boknowswiki 
Yeah, I am referring about the above scenario. Because we are deleting extra healthy replicas when DataLocality and Eviction are on, I am thinking whether we should do the same in normal cases so that users have uniform expectation `Longhorn will always maintain exactly NumberOfReplicas replicas`. 

Having extra healthy replica laying around is not always a good thing. It creates more load on the network. So if the users decrease the NumberOfReplicas, it is more likely that they want to decrease the actual number of replicas.

>  User may have plan with it or user may just want to leave it there
If the users have a plan with the replica, I am not sure why they decrease the NumberOfReplicas.

>  we don't know which one user may want to delete 
Yeah, I agree with this one.

controller/volume_controller.go
```diff
@@ -624,16 +623,89 @@ func (vc *VolumeController) cleanupCorruptedOrStaleReplicas(v *longhorn.Volume,
...
+
+	for _, r := range rs {
+		if r.Status.EvictionRequested == true {
+			hasEvictionRequestedReplicas = true
+		}
+	}
+
+	if healthyCount >= v.Spec.NumberOfReplicas {
+		for _, r := range rs {
+			if !hasEvictionRequestedReplicas {
```
yasker, line 650: Hmm? Are we going to have the same issue of repeatedly deleting the local replica in this case?

yasker (changes requested):

controller/volume_controller.go
boknowswiki: Hi Sheng,

Sorry, I don't fully understand your question. This is for eviction requested volume, in the case of there is no enough resource to evict the replica, we will have one failed to scheduled replica waiting for user ether to add more resource(node) or cancel the eviction. Thanks.

yasker: I meant, are we going to keep the unscheduled replica due to data locality?

PhanLe1010:
> Hmm? Are we going to have the same issue of repeatedly deleting the local replica in this case? I thought we're going to change the scheduling mechanism for the local replica to be the same as others, or are we still checking it separately?
> I meant, are we going to keep the unscheduled replica due to data locality?

we will discuss this over zoom.

yasker (approved):

PhanLe1010:
Memory Note: In this PR, we don't delete extra healthy replicas when DataLocality and Eviction are off.

PR #673, opened 2020-09-02
api: Fix return error message to UI

boknowswiki:
Eviction is only allowed on scheduling disabled disk or node. Return
error message to user if trying to evict on a scheduled disk or node, or
enabled scheduling on an eviction requested disk or node.

https://github.com/longhorn/longhorn/issues/298

### DIFF
--- a/api/node.go
+++ b/api/node.go
@@ -70,7 +70,7 @@ func (s *Server) NodeUpdate(rw http.ResponseWriter, req *http.Request) error {
 	// Only scheduling disabled node can be evicted
 	// Can not enable scheduling on an evicting node
 	if n.EvictionRequested == true && n.AllowScheduling != false {
-		return errors.Wrap(err, "Need to disable scheduling on this node for node eviction, or cancel eviction to enable scheduling on this node.")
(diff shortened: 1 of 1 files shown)


api/node.go
```diff
@@ -70,7 +70,7 @@ func (s *Server) NodeUpdate(rw http.ResponseWriter, req *http.Request) error {
 	// Only scheduling disabled node can be evicted
 	// Can not enable scheduling on an evicting node
 	if n.EvictionRequested == true && n.AllowScheduling != false {
-		return errors.Wrap(err, "Need to disable scheduling on this node for node eviction, or cancel eviction to enable scheduling on this node.")
+		return errors.New("need to disable scheduling on this node for node eviction, or cancel eviction to enable scheduling on this node")
```
yasker, line 73: Use `fmt.Errorf`, and add node/disk name to the error message.

yasker (changes requested):

api/node.go
boknowswiki: Sure, will fix it. Thanks.

shuo-wu (approved):

yasker (approved):

PR #674, opened 2020-09-03
Introduce a new CRD 'Disk'

shuo-wu:
longhorn/longhorn#1269


shuo-wu:
1. Need to update UI since the UI doesn't work after applying this PR.
2. Will verify and update integration tests later.

controller/disk_controller.go
```diff
@@ -0,0 +1,541 @@
...
+		dc.eventRecorder.Eventf(disk, v1.EventTypeNormal, EventReasonDelete, "Deleting disk %v", disk.Name)
+		return dc.ds.RemoveFinalizerForDisk(disk)
+	}
+
+	existingDisk := disk.DeepCopy()
+	defer func() {
+		if err != nil {
+			return
+		}
+		if disk.Status.State != existingDisk.Status.State {
```
yasker, line 268: Why are we checking the state here rather than nodeID directly?

controller/disk_controller.go
```diff
@@ -0,0 +1,541 @@
...
+		}
+		// Found multiple disks in the same Fsid
+		if isDuplicate {
+			diskFailureReason = string(types.DiskConditionReasonDiskFilesystemChanged)
+			diskFailureMessage = fmt.Sprintf("Disk has same file system ID %v as other disks", info.Fsid)
+			return nil
+		}
+		disk.Status.FSID = info.Fsid
+	} else {
+		if disk.Status.FSID != info.Fsid {
```
yasker, line 374: disk FSID can change (https://github.com/longhorn/longhorn/blob/master/enhancements/20200331-replace-filesystem-id-key-in-disk-map.md#summary). That's why we introduced the disk UUID. We shouldn't use fsid to identify the disk.

controller/disk_controller.go
```diff
@@ -0,0 +1,541 @@
...
+			return nil
+		}
+	}
+
+	diskUUID := ""
+	diskConfig, err := dc.getDiskConfig(disk.Spec.Path)
+	if err != nil {
+		if !types.ErrorIsNotFound(err) {
+			diskFailureReason = string(types.DiskConditionReasonNoDiskInfo)
+			diskFailureMessage = fmt.Sprintf("Failed to get disk config: %v", err)
```
yasker, line 386: What if the config doesn't exist?

controller/disk_controller.go
```diff
@@ -0,0 +1,541 @@
...
+			if err != nil {
+				diskFailureReason = string(types.DiskConditionReasonNoDiskInfo)
+				diskFailureMessage = fmt.Sprintf("Failed to generate disk config: %v", err)
+				return nil
+			}
+			diskUUID = diskConfig.DiskUUID
+			log.Debugf("New disk is introduced for Longhorn, a disk config file with UUID %v will be generated", diskUUID)
+		}
+		if disk.Status.DiskUUID != diskUUID {
+			log.Debugf("Set disk UUID to %v", diskUUID)
```
yasker, line 404: Is this an existing disk?

yasker (changes requested):

controller/disk_controller.go
```diff
@@ -0,0 +1,541 @@
...
+	if err != nil {
+		return err
+	}
+	if isNodeDown {
+		diskFailureReason = string(types.DiskConditionReasonNodeUnknown)
+		diskFailureMessage = fmt.Sprintf("The connected node %v is down or deleted", disk.Spec.NodeID)
+		return nil
+	}
+	// Prevent the corner case:
+	// The corresponding ndoe is up but the ownership of the disk hasn't been transferred.
```
c3y1huang, line 347:
```suggestion
	// The corresponding node is up but the disk ownership hasn't been transferred.
```

controller/disk_controller.go
shuo-wu: In that case, a config file will be generated later. And due to `if !types.ErrorIsNotFound(err) {`, this line won't be executed.

controller/disk_controller.go
shuo-wu: This means the disk hasn't been used as a Longhorn disk. And Longhorn needs to generate UUID for it.

shuo-wu:
All (modified) integration tests except for `test_recurring_job_labels` get passed. And that failed test seems to be not related to this PR.

upgrade/v102to110/upgrade.go
```diff
@@ -0,0 +1,129 @@
...
+		return err
+	}
+
+	return nil
+}
+
+func createDisksAndUpdateNodes(namespace string, lhClient *lhclientset.Clientset) (err error) {
+	defer func() {
+		err = errors.Wrapf(err, upgradeLogPrefix+"upgrade instance manager failed")
+	}()
```
PhanLe1010, line 41: Is the log message correct? I got into this log message while upgrading and it was a bit confusing.

upgrade/v102to110/upgrade.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+func generateFixedDiskNameForExistingDisks(nodeName, diskID string) string {
+	return fmt.Sprintf("%s-%s", diskID, util.GetStringChecksum(strings.TrimSpace(nodeName))[:8])
+}
+
+func doReplicaUpgrade(namespace string, lhClient *lhclientset.Clientset) (err error) {
+	defer func() {
+		err = errors.Wrapf(err, upgradeLogPrefix+"upgrade instance manager failed")
+	}()
```
PhanLe1010, line 108: Same as https://github.com/longhorn/longhorn-manager/pull/674/files#r493271476

upgrade/v102to110/upgrade.go
```diff
@@ -0,0 +1,129 @@
...
+	return nil
+}
+
+func getNodeLabelSelector(nodeName string) string {
+	return fmt.Sprintf("%s=%s", types.LonghornNodeKey, nodeName)
+}
+
+func generateFixedDiskNameForExistingDisks(nodeName, diskID string) string {
+	return fmt.Sprintf("%s-%s", diskID, util.GetStringChecksum(strings.TrimSpace(nodeName))[:8])
+}
```
PhanLe1010, line 103: @shuo-wu This function is problematic when there is a fail-to-schedule replica in the cluster.  In this case, the diskID  of the  fail-to-schedule replica is empty, so the return value will have format `-xxxxxxxx`. That is invalid for Kubernetes label value.  Please see this error for more details:
```
time="2020-09-23T07:39:23Z" level=error msg="Upgrade failed: upgrade CRD failed: upgrade from v1.0.2 to v1.1.0: upgrade instance manager failed: Replica.longhorn.io \"testvol3-r-12d412a9\" is invalid: metadata.labels: Invalid value: \"-cf83e135\": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')"
```

shuo-wu: This implementation has been deprecated. And thank you for pointing out the error!

yasker:
I noticed we have `disk.Spec.Path` stored in the disk object. Does it work if the disk was mounted on a different directory on the node later? Do we update the path accordingly?

yasker:
Also, there is no way to `Add/delete` disk in the UI.

shuo-wu:
> Also, the UI `Connect Disk` part is a bit weird. We can rename it back to `Add Disk`, and add another option `Add disk` as drop down for the node option with prefilled node ID. I was confused for a bit that I cannot add disk in the new UI.

The path can be determined once the disk object is created. I forgot to update the `disk.Spec.Path` when a node is trying to reuse a disk object.

The UI image I am using is `wangsiye/longhorn-ui:3234846`. It currently looks good except for one minor issue. I will ask SIye to update it once the backend is merged.

yasker:
If the `Path` remain at spec, then the node controller needs to update it. Otherwise, disk controller can update it once detect the change from the node, just like it updates the nodeID.

shuo-wu:
OK. Avoiding too many disks related operations in the node controller seems better. I will update the implementation.

yasker:
The disk object enqueue part need to be updated like what's done in https://github.com/longhorn/longhorn-manager/pull/725.

shuo-wu:
Close this PR since the design changes.

PR #675, opened 2020-09-03
CSI snapshotter implementation

joshimoo:
This is a straight forward implementation of the csi snapshotter based on the current backup interaction pattern.
There are some concerns about the volume.BackupStatus since longhorn has a 1 - many relationship from snapshot to backup.
While the csi snapshotter assumption is a 1-1 relationship.

The csi snapshot restore part uses the same mechanism as the previous storageClass fromBackup approach.
Currently there is an issue longhorn/longhorn#1754 with that part but it's unrelated to the CSI snapshotter implementation.

The csi snapshots require a prefix, I changed the default from `snapshot` to `csi`, this unfortunately makes the longhorn snapshot ui look bad, @yasker maybe we should consider prefixing the longhorn snapshots with `snapshot` as well then have the UI strip the prefix for displaying the name?

Prerequisite:
- make sure your cluster contains the below crds, rancher rke did not deploy them for me.
https://github.com/kubernetes-csi/external-snapshotter/tree/master/client/config/crd 

- make sure your cluster contains the snapshot controller, rancher rke did not deploy it for me.
https://github.com/kubernetes-csi/external-snapshotter/tree/master/deploy/kubernetes/snapshot-controller

modify the above yaml files for the snapshot controller if you don't want it in the default namespace.

### New Snapshot (dynamic provisioning)

- create a volume + pvc + pv named `test-vol`
- you can trigger snapshots via the below snapshot_pvc.yaml
- you can trigger restores via the below restore_pvc_snapshot.yaml

snapshot_pvc.yaml
```yaml
apiVersion: snapshot.storage.k8s.io/v1beta1
kind: VolumeSnapshot
metadata:
  name: test-vol-snapshot
spec:
  volumeSnapshotClassName: longhorn
  source:
    persistentVolumeClaimName: test-vol
```

restore_pvc_snapshot.yaml
```yaml
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: test-vol-restore
spec:
  storageClassName: longhorn
  dataSource:
    name: test-vol-snapshot
    kind: VolumeSnapshot
    apiGroup: snapshot.storage.k8s.io
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 2Gi
```

### Existing Longhorn Backup (pre provisioning)

existing_backup.yaml
```yaml
apiVersion: snapshot.storage.k8s.io/v1beta1
kind: VolumeSnapshotContent
metadata:
  name: test-existing-backup
spec:
  volumeSnapshotClassName: longhorn
  driver: driver.longhorn.io
  deletionPolicy: Delete
  source:
    # NOTE: change this to point to an existing backup on the backupstore
    snapshotHandle: bs://test-vol/backup-625159fb469e492e
  volumeSnapshotRef:
    name: test-snapshot-existing-backup
    namespace: default
```

snapshot_existing.yaml
```yaml
apiVersion: snapshot.storage.k8s.io/v1beta1
kind: VolumeSnapshot
metadata:
  name: test-snapshot-existing-backup
spec:
  volumeSnapshotClassName: longhorn
  source:
    volumeSnapshotContentName: test-existing-backup
```

restore_existing_backup.yaml
```yaml
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: test-restore-existing-backup
spec:
  storageClassName: longhorn
  dataSource:
    name: test-snapshot-existing-backup
    kind: VolumeSnapshot
    apiGroup: snapshot.storage.k8s.io
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 2Gi
```

longhorn/longhorn#304


csi/deployment.go
```diff
@@ -19,7 +19,7 @@ import (
 
 const (
 	DefaultCSIAttacherImage            = "longhornio/csi-attacher:v2.0.0"
-	DefaultCSIProvisionerImage         = "longhornio/csi-provisioner:v1.4.0"
+	DefaultCSIProvisionerImage         = "longhornio/csi-provisioner:v1.6.0"
```
yasker, line 23: Will this be upgraded automatically if the user upgrades Longhorn?

Also, does provisioner v1.6.0 has requirements on Kubernetes version?

app/driver.go
```diff
@@ -221,6 +239,11 @@ func deployCSIDriver(kubeClient *clientset.Clientset, lhClient *lhclientset.Clie
 		return err
 	}
 
+	snapshotSupportEnabled, err := util.IsKubernetesVersionAtLeast(kubeClient, types.CSISnapshotterMinVersion)
```
yasker, line 242: We need to think about what's our expectation for the snapshotter controller. I would prefer us to deploy it if possible, though strictly speaking it's not a Longhorn component.

csi/controller_server.go
```diff
@@ -335,12 +374,176 @@ func (cs *ControllerServer) GetCapacity(context.Context, *csi.GetCapacityRequest
...
+	backupVolume, err := cs.apiClient.BackupVolume.ById(csiVolumeName)
+	if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+	backupListOutput, err := cs.apiClient.BackupVolume.ActionBackupList(backupVolume)
+	if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+
+	// NOTE: csi-snapshots assume a 1 to 1 relationship, longhorn allows for multiple backups of a snapshot
```
yasker, line 399: CSI snapshot are Longhorn backups, so it's still one to one?

csi/controller_server.go
```diff
@@ -335,12 +374,176 @@ func (cs *ControllerServer) GetCapacity(context.Context, *csi.GetCapacityRequest
...
+	var backup *longhornclient.Backup
+	for _, b := range backupListOutput.Data {
+		if b.SnapshotName == csiSnapshotName {
```
yasker, line 402: We should check the backup name/id rather than the snapshot name.

yasker (changes requested):
We can use the `backup` prefix?

Also, even it's called `snapshot` in CSI, it's essentially the `backup` in longhorn, so it's still one to one mapping.

yasker:
Btw, what's the usage for the prefix? Is it suppose to be vendor specific?

csi/controller_server.go
```diff
@@ -335,12 +374,176 @@ func (cs *ControllerServer) GetCapacity(context.Context, *csi.GetCapacityRequest
...
+// encodeSnapshotID cncodes the backup volume as part of the snapshotID
+// otherwise we need to iterate over all backup volumes,
+// when trying to find a backup for deletion or restoration
+func encodeSnapshotID(volumeName string, backupName string) string {
+	return fmt.Sprintf("backupstore/%s/%s", volumeName, backupName)
+}
+
+// decodeSnapshotID splits up the snapshotID back into it's components
+// backupType will be used once we implement the VolumeBackup crd
+func decodeSnapshotID(snapshotID string) (backupType string, volumeName string, backupName string) {
```
c3y1huang, line 520:
```suggestion
func decodeSnapshotID(snapshotID string) (backupType, volumeName, backupName string) {
```

app/driver.go
joshimoo: Agree, but that can be a separate PR, this way we can test the csi snapshotter interactions already.

csi/controller_server.go
joshimoo: csi snapshot(1) -> (1)longhorn snapshot(1) -> (n) backups

csi/controller_server.go
joshimoo: That's one of the problems with the current backup design I am alluding to, we don't know the backupID.

joshimoo:
> Btw, what's the usage for the prefix? Is it suppose to be vendor specific?

No it's just used as a prefix, for the [final snapshot name](https://github.com/kubernetes-csi/external-snapshotter/blob/42650bce93ddaaa8dbf6fdb21a38835533eee9f2/pkg/sidecar-controller/csi_handler.go#L73l), this might be used by google/aws to filewatch a directory for snapshots then upload them to final resting place.

csi/deployment.go
joshimoo: Same kubernetes [minimum version 1.13](https://github.com/kubernetes-csi/external-provisioner/releases/tag/v1.6.0) requirement as  the currently used [provisioner](https://github.com/kubernetes-csi/external-provisioner/releases/tag/v1.4.0)

For the default installation yes, but it's possible that the users have pinned their images (airgap),
in which case they need to update their images anyway if they want to use the csi-snapshot support.

csi/controller_server.go
joshimoo: Thanks I prefer to specify the types explicitly, It makes it easier to see what is what, it also makes it easier to refactor the params down the line.
What's your reasoning for suggesting this change?

yasker: I am fine with either way. Use `xxx, xxx, xxx <type>` might be more of a Go thing.

yasker:
`longhornio/csi-provisioner:v1.6.0` and `longhornio/csi-snapshotter:v2.1.1` have been pushed to the Docker Hub.

csi/controller_server.go
yasker: I see. So you're trying to use/match the CSI snapshot name with the snapshot name in Longhorn. So if there is any backup with this snapshot name, then we will use that backup. Otherwise, we will create a new snapshot with the name and back it up.

I didn't notice this part of logic in the LEP, so I didn't realize we're actively using the CSI snapshot name.

Will it make things easier if we just keep the CSI snapshot name in the CSI? So any new snapshot call will create a new snapshot/backup, no matter what the name is?

yasker:
Should we upgrade all the CSI components to v1.6.0 instead?

csi/controller_server.go
c3y1huang: Np, more of a preference thing. To me, cleaning up the duplication of the same type names sometimes could help to make it a bit easier to read.

csi/controller_server.go
joshimoo: This would not work, the csi-snapshot-request name is generated by the csi snapshotter, based on the configured prefix + `Snapshot.uuid` this information is also used to generate the `SnapshotContent.name = snapcontent-uuid`

example: `kubectl get volumesnaphots test-snapshot-pvc`:
```yaml
kind: VolumeSnapshot
metadata:
  name: test-snapshot-pvc
  uid: aec4bdc3-3244-443e-a26d-4cd48b3e27af
spec:
  source:
    persistentVolumeClaimName: test-vol
  volumeSnapshotClassName: longhorn
status:
  boundVolumeSnapshotContentName: snapcontent-aec4bdc3-3244-443e-a26d-4cd48b3e27af
  readyToUse: false
```

the csi-snapshotter will always call the longhorn csi plugin with the same csi snapshot request name for a specific `Snapshot` object.

So we have to use the csi snapshotter generated name, for the longhorn snapshot or introduce a CRD (VolumeBackup) so that we have a persisted way of associating `csi - snapshot - backup`

If we would ignore the name, we would end up generating a new snapshot for each successive call, since a backup can take a long time for very big volumes, that would not be the desired behavior.

One might consider storing the csi name in memory, but that wouldn't work since there can be multiple csi-plugins also that wouldn't survive crashes of the csi-plugin. Basically the same problem we currently have with the backupStatus information. Long term the only way out of this is the backup refactor tracked here longhorn/longhorn#1761

yasker: We've discussed this and I am good with the solution.

Also, need to update `snapshotHandle` schema.

csi/controller_server.go
joshimoo: Thanks for the discussion, I removed the duplicate type parameters.
I checked the code base and we seems to remove them in most places, it's better to have a consistent code base.
When in rome do like the romans :)

I generally prefer to be more verbose the wider the scope of the variables/parameters is.
As example using `i` as a loop variable is totally fine with me, using `i` as a function parameter is discussable

csi/controller_server.go
joshimoo: Updated the schema for the `snapshotID` to use `bs://volumeName/backupName`, inside of the csi driver it's called `snapshotID`, the final VolumeSnapshotContent calls it `snapshotHandle`. Since the csi driver only deals with the request/response pairs I decided to stick with `snapshotID`

joshimoo:
In regard to updating the images, we use the following csi images:
```
quay.io/k8scsi/csi-attacher:v2.0.0
quay.io/k8scsi/csi-node-driver-registrar:v1.2.0
quay.io/k8scsi/csi-provisioner:v1.6.0
quay.io/k8scsi/csi-resizer:v0.3.0
quay.io/k8scsi/csi-snapshotter:v2.1.1
```

The csi-resizer could be updated, but is independent and not required for the csi snapshot functionality.
It also introduced a change in behavior in version 1.0 which is the latest version of it.

joshimoo:
I changed the snapshot prefix back to the default `snapshot` since whether we use `csi` or `snapshot` makes no difference from the ui perspective, by using `snapshot` we keep it generic and could change the longhorn snapshot scheme to fit in the future but is not required.

For the ui we can filter out the `snapshot` prefix if we want to do so.

yasker (approved):

PR #677, opened 2020-09-03
controller: Add eviction support for detached volume

boknowswiki:
This change adds auto attach for eviction requested volume and auto
detach it once the eviction is done.

https://github.com/longhorn/longhorn/issues/1704

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -570,6 +570,16 @@ func (vc *VolumeController) getHealthyReplicaCount(rs map[string]*longhorn.Repli
 	return healthyCount
 }
 
+func (vc *VolumeController) hasReplicaEvictionRequested(rs map[string]*longhorn.Replica) bool {
+	for _, r := range rs {
+		if r.Status.EvictionRequested == true {
+			return true
+		}
+	}
+
+	return false
+}
+
 func (vc *VolumeController) cleanupReplicas(v *longhorn.Volume, e *longhorn.Engine, rs map[string]*longhorn.Replica) error {
 	if err := vc.cleanupCorruptedOrStaleReplicas(v, rs); err != nil {
 		return err
@@ -637,13 +647,7 @@ func (vc *VolumeController) cleanupCorruptedOrStaleReplicas(v *longhorn.Volume, 
 
 func (vc *VolumeController) cleanupFailedToScheduledReplicas(v *longhorn.Volume, rs map[string]*longhorn.Replica) (err error) {
 	healthyCount := vc.getHealthyReplicaCount(rs)
-	hasEvictionRequestedReplicas := false
-
-	for _, r := range rs {
-		if r.Status.EvictionRequested == true {
-			hasEvictionRequestedReplicas = true
-		}
-	}
+	hasEvictionRequestedReplicas := vc.hasReplicaEvictionRequested(rs)
 
 	if healthyCount >= v.Spec.NumberOfReplicas {
 		for _, r := range rs {
@@ -824,7 +828,7 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
 	if err := vc.checkForAutoAttachment(v, e, rs, allScheduled); err != nil {
 		return err
 	}
-	if err := vc.checkForAutoDetachment(v, e, allScheduled); err != nil {
+	if err := vc.checkForAutoDetachment(v, e, rs, allScheduled); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


controller/volume_controller.go
```diff
@@ -821,10 +825,13 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
 		return err
 	}
 
-	if err := vc.checkForAutoAttachment(v, e, rs, allScheduled); err != nil {
+	hasEvictionRequestedReplicas := vc.hasReplicaEvictionRequested(rs)
+
+	if err := vc.checkForAutoAttachment(v, e, rs, allScheduled, hasEvictionRequestedReplicas); err != nil {
```
yasker, line 830: Both auto-attach and detach function can have replicas as a parameter, so we don't need to check the condition outside then pass it in.

controller/volume_controller.go
```diff
@@ -1636,8 +1645,9 @@ func (vc *VolumeController) checkForAutoAttachment(v *longhorn.Volume, e *longho
 	// Do auto attachment for:
 	//   1. restoring/DR volumes.
 	//   2. offline expansion.
+	//	 3. Eviction requested on this volume.
```
yasker, line 1648: Wrong indent.

controller/volume_controller.go
```diff
@@ -1667,6 +1683,7 @@ func (vc *VolumeController) checkForAutoDetachment(v *longhorn.Volume, e *longho
 		v.Status.CurrentNodeID = ""
 		return nil
 	}
+
```
yasker, line 1686: ?

yasker (changes requested):

controller/volume_controller.go
boknowswiki: then the new function will be called twice. Each once in auto-attach and auto-detach, is that ok?

controller/volume_controller.go
boknowswiki: I also noticed this, but in my local environment, the numbers are aligned. Let me double check on this.

controller/volume_controller.go
boknowswiki: will remove this new line. Thanks.

controller/volume_controller.go
yasker: It's normally tab vs space.

controller/volume_controller.go
yasker: That's fine since it's just a status check, doesn't have side effects.

boknowswiki: Ok, will fix it.

boknowswiki: Sorry my bad. auto-detach doesn't have replicas as parameter passed in. Should I add the rs new parameter or keep this way? Thanks.

yasker: We can just add `rs` parameter. It's still a small change.

controller/volume_controller.go
```diff
@@ -1617,7 +1621,8 @@ func (vc *VolumeController) reconcileVolumeSize(v *longhorn.Volume, e *longhorn.
 	return nil
 }
 
-func (vc *VolumeController) checkForAutoAttachment(v *longhorn.Volume, e *longhorn.Engine, rs map[string]*longhorn.Replica, allScheduled bool) error {
+func (vc *VolumeController) checkForAutoAttachment(v *longhorn.Volume,
```
yasker, line 1624: Line breaks? Not sure if it's necessary.

Same below.

yasker (commented):
Small coding style issue, otherwise LGTM.

@shuo-wu ?

controller/volume_controller.go
boknowswiki: Sure, just changed them back. Thanks.

yasker (dismissed):

controller/volume_controller.go
```diff
@@ -1657,6 +1662,10 @@ func (vc *VolumeController) checkForAutoDetachment(v *longhorn.Volume, e *longho
 
 	// Do auto-detachment for non-restore/DR volumes.
 	if !v.Status.RestoreRequired && !v.Status.IsStandby {
+		// Don't for auto-detachment if the evivtion is going on.
```
shuo-wu, line 1665: Put it outside of this `if`. e.g.:
```
	if v.Status.ExpansionRequired {
		return nil
	}
       
        if vc.hasReplicaEvictionRequested(rs) {
		return nil
	}		
```

And there is a typo `evivtion` in the comment.

shuo-wu (changes requested):

controller/volume_controller.go
boknowswiki: Fixed, thanks.

yasker:
Since @shuo-wu 's comment has been addressed, I assume it's OK to merge this PR.

yasker (approved):

PR #678, opened 2020-09-04
controller, scheduler: Refactor data locality code

PhanLe1010:
Added a new field HardNodeAffinity in replica's spec to keep track of which node this replica must be scheduled to.
Refactor the replica scheduler so that it considers the HardNodeAffinity.
Modify the fail-to-scheduled replica cleanup logic so that it avoids the local replica.

longhorn/longhorn#1045

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -378,10 +378,6 @@ func (vc *VolumeController) syncVolume(key string) (err error) {
 			return err
 		}
 
-		if err := vc.reconcileLocalReplica(volume, engine, replicas); err != nil {
-			return err
-		}
-
 		if err := vc.updateRecurringJobs(volume); err != nil {
 			return err
 		}
@@ -432,7 +428,7 @@ func (vc *VolumeController) EvictReplicas(v *longhorn.Volume,
 	for _, replica := range rs {
 		if replica.Status.EvictionRequested == true &&
 			healthyCount == v.Spec.NumberOfReplicas {
-			if err = vc.replenishReplicas(v, e, rs); err != nil {
+			if err := vc.replenishReplicas(v, e, rs, ""); err != nil {
 				logrus.Errorf("Failed to create new replica for replica eviction err %v", err)
 				vc.eventRecorder.Eventf(v, v1.EventTypeWarning,
 					EventReasonFailedEviction,
@@ -534,6 +530,14 @@ func (vc *VolumeController) ReconcileEngineReplicaState(v *longhorn.Volume, e *l
 		if err := vc.EvictReplicas(v, e, rs, healthyCount); err != nil {
 			return err
 		}
+
+		// Migrate local replica when Data Locality is on
+		if v.Status.State == types.VolumeStateAttached && !isDataLocalityDisabled(v) && !hasLocalReplicaOnSameNodeAsEngine(e, rs) {
+			if err := vc.replenishReplicas(v, e, rs, e.Spec.NodeID); err != nil {
+				return err
+			}
+		}
+
 	} else { // healthyCount < v.Spec.NumberOfReplicas
 		v.Status.Robustness = types.VolumeRobustnessDegraded
 		if oldRobustness != types.VolumeRobustnessDegraded {
@@ -549,7 +553,7 @@ func (vc *VolumeController) ReconcileEngineReplicaState(v *longhorn.Volume, e *l
 		//   2. the volume is expanding size.
 		if !((v.Status.IsStandby || v.Status.RestoreRequired) && cliAPIVersion < engineapi.CLIVersionFour) &&
 			v.Spec.Size == e.Status.CurrentSize {
-			if err = vc.replenishReplicas(v, e, rs); err != nil {
+			if err := vc.replenishReplicas(v, e, rs, ""); err != nil {
 				return err
 			}
 		}
@@ -648,8 +652,8 @@ func (vc *VolumeController) cleanupFailedToScheduledReplicas(v *longhorn.Volume,
 	if healthyCount >= v.Spec.NumberOfReplicas {
 		for _, r := range rs {
 			if !hasEvictionRequestedReplicas {
-				if r.Spec.HealthyAt == "" && r.Spec.FailedAt == "" &&
-					r.Spec.NodeID == "" {
+				if r.Spec.HealthyAt == "" && r.Spec.FailedAt == "" && r.Spec.NodeID == "" &&
+					(isDataLocalityDisabled(v) || r.Spec.HardNodeAffinity != v.Status.CurrentNodeID) {
 					logrus.Infof("Cleaning up failed to scheduled replica %v", r.Name)
 					if err := vc.deleteReplica(r, rs); err != nil {
 						return errors.Wrapf(err, "cannot cleanup failed to scheduled replica %v", r.Name)
@@ -682,7 +686,7 @@ func (vc *VolumeController) cleanupExtraHealthyReplicas(v *longhorn.Volume, e *l
 		}
 	}
 
-	// Clean up extra healthy replica with data-locallity finished
+	// Clean up extra healthy replica with data-locality finished
 	healthyCount = vc.getHealthyReplicaCount(rs)
 
 	if healthyCount > v.Spec.NumberOfReplicas &&
@@ -694,7 +698,7 @@ func (vc *VolumeController) cleanupExtraHealthyReplicas(v *longhorn.Volume, e *l
 		}
 
 		// Randomly delete extra non-local healthy replicas in the preferred candidate list rNames
-		// Sometime reconcileLocalReplica() is called more than once with the same input (v,e,rs).
+		// Sometime cleanupExtraHealthyReplicas() is called more than once with the same input (v,e,rs).
 		// To make the deleting operation idempotent and prevent deleting more replica than needed,
 		// we always delete the replica with the smallest name.
 		sort.Strings(rNames)
@@ -775,7 +779,7 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
 
 	if len(rs) == 0 {
 		// first time creation
-		if err = vc.replenishReplicas(v, e, rs); err != nil {
+		if err := vc.replenishReplicas(v, e, rs, ""); err != nil {
 			return err
 		}
 	}
@@ -801,16 +805,21 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
 			return err
 		}
 		if scheduledReplica == nil {
-			logrus.Errorf("unable to schedule replica %v of volume %v", r.Name, v.Name)
-			v.Status.Conditions = types.SetCondition(v.Status.Conditions,
-				types.VolumeConditionTypeScheduled, types.ConditionStatusFalse,
-				types.VolumeConditionReasonReplicaSchedulingFailure, "")
+			if r.Spec.HardNodeAffinity == "" {
+				logrus.Errorf("unable to schedule replica %v of volume %v", r.Name, v.Name)
+				v.Status.Conditions = types.SetCondition(v.Status.Conditions,
+					types.VolumeConditionTypeScheduled, types.ConditionStatusFalse,
+					types.VolumeConditionReasonReplicaSchedulingFailure, "")
+			} else {
+				logrus.Errorf("unable to schedule replica %v of volume %v with HardNodeAffinity = %v", r.Name, v.Name, r.Spec.HardNodeAffinity)
+				v.Status.Conditions = types.SetCondition(v.Status.Conditions,
+					types.VolumeConditionTypeScheduled, types.ConditionStatusFalse,
+					types.VolumeConditionReasonLocalReplicaSchedulingFailure, "")
+			}
 			allScheduled = false
-			// no need to continue, since we won't able to schedule
-			// more replicas if we failed this one
-			break
+		} else {
+			rs[r.Name] = scheduledReplica
 		}
-		rs[r.Name] = scheduledReplica
 	}
 	if allScheduled {
 		v.Status.Conditions = types.SetCondition(v.Status.Conditions,
@@ -1155,33 +1164,6 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
 	return nil
 }
 
-func (vc *VolumeController) reconcileLocalReplica(v *longhorn.Volume, e *longhorn.Engine, rs map[string]*longhorn.Replica) error {
-	if isDataLocalityDisabled(v) {
-		return nil
-	}
-
-	if v.Status.State != types.VolumeStateAttached {
-		return nil
-	}
-
-	if e == nil {
-		return fmt.Errorf("BUG: ReconcileLocalReplica needs a valid engine")
-	}
-
-	if !vc.isSafeToModifyReplica(v, e, rs) {
-		return nil
-	}
-
-	if !hasLocalReplicaOnSameNodeAsEngine(e, rs) {
-		if _, err := vc.addLocalReplicaOnSameNodeAsEngine(v, e, rs); err != nil {
-			return err
-		}
-		return nil
-	}
-
-	return nil
-}
-
 func (vc *VolumeController) getPreferredReplicaCandidatesForDeletion(rs map[string]*longhorn.Replica) ([]string, error) {
 	diskToReplicaMap := make(map[string][]string)
 	nodeToReplicaMap := make(map[string][]string)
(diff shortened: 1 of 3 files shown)


controller/volume_controller.go
```diff
@@ -432,7 +428,7 @@ func (vc *VolumeController) EvictReplicas(v *longhorn.Volume,
 	for _, replica := range rs {
 		if replica.Status.EvictionRequested == true &&
 			healthyCount == v.Spec.NumberOfReplicas {
-			if err = vc.replenishReplicas(v, e, rs); err != nil {
+			if err = vc.replenishReplicas(v, e, rs, ""); err != nil {
```
yasker, line 431: Why we're using `err = ` instead of `err :=` here? And there are similar cases below.

controller/volume_controller.go
```diff
@@ -549,7 +553,7 @@ func (vc *VolumeController) ReconcileEngineReplicaState(v *longhorn.Volume, e *l
 		//   2. the volume is expanding size.
 		if !((v.Status.IsStandby || v.Status.RestoreRequired) && cliAPIVersion < engineapi.CLIVersionFour) &&
 			v.Spec.Size == e.Status.CurrentSize {
-			if err = vc.replenishReplicas(v, e, rs); err != nil {
+			if err = vc.replenishReplicas(v, e, rs, ""); err != nil {
```
yasker, line 556: Why we're using `err = ` instead of `err :=` here?

controller/volume_controller.go
```diff
@@ -775,7 +779,7 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
 
 	if len(rs) == 0 {
 		// first time creation
-		if err = vc.replenishReplicas(v, e, rs); err != nil {
+		if err = vc.replenishReplicas(v, e, rs, ""); err != nil {
```
yasker, line 782: Why we're using `err = ` instead of `err :=` here?

controller/volume_controller.go
```diff
@@ -534,6 +530,14 @@ func (vc *VolumeController) ReconcileEngineReplicaState(v *longhorn.Volume, e *l
 		if err := vc.EvictReplicas(v, e, rs, healthyCount); err != nil {
 			return err
 		}
+
+		// Migrate local replica when Data Locality is on
+		if v.Status.State == types.VolumeStateAttached && !isDataLocalityDisabled(v) && !hasLocalReplicaOnSameNodeAsEngine(e, rs) {
+			if err = vc.replenishReplicas(v, e, rs, e.Spec.NodeID); err != nil {
```
yasker, line 536: Why we're using `err = ` instead of `err :=` here?

controller/volume_controller.go
```diff
@@ -1242,63 +1220,22 @@ func isDataLocalityDisabled(v *longhorn.Volume) bool {
...
 }
 
-// hasLocalReplicaOnSameNodeAsEngine returns whether there exist a replica on the same node as engine
+// hasLocalReplicaOnSameNodeAsEngine returns true if one of the following condition is satisfied:
+// 1. there exist a replica on the same node as engine
+// 2. there exist a replica with HardNodeAffinity set to engine's NodeID
 func hasLocalReplicaOnSameNodeAsEngine(e *longhorn.Engine, rs map[string]*longhorn.Replica) bool {
 	for _, r := range rs {
-		if r.Spec.NodeID == e.Spec.NodeID {
+		if e.Spec.NodeID != "" && r.Spec.NodeID == e.Spec.NodeID || r.Spec.HardNodeAffinity == e.Spec.NodeID {
```
yasker, line 1228: Seems the precedence is wrong here? Should it be `e.Spec.NodeID != "" && (r.Spec.NodeID == e.Spec.NodeID || r.Spec.HardNodeAffinity == e.Spec.NodeID)` ?

controller/volume_controller.go
```diff
@@ -1317,6 +1254,15 @@ func (vc *VolumeController) replenishReplicas(v *longhorn.Volume, e *longhorn.En
 		return nil
 	}
 
+	if hardNodeAffinity != "" && getRebuildingReplicaCount(e) == 0 {
+		r, err := vc.createReplica(v, e, rs, hardNodeAffinity)
+		if err != nil {
+			return err
+		}
+		rs[r.Name] = r
+		return nil
```
yasker, line 1263: hmm? Should we skip all the logic below? This should be automatically taken care of.

yasker (changes requested):

controller/volume_controller.go
PhanLe1010: Oops. My bad. I am sorry!

controller/volume_controller.go
PhanLe1010: Thank you! Fixed

controller/volume_controller.go
PhanLe1010: Thank you! Fixed

controller/volume_controller.go
PhanLe1010: Thank you! Fixed

controller/volume_controller.go
PhanLe1010: Thank you! Fixed

controller/volume_controller.go
PhanLe1010: Do you mean that we can skip this logic?
```
if hardNodeAffinity != "" && getRebuildingReplicaCount(e) == 0 {
 		r, err := vc.createReplica(v, e, rs, hardNodeAffinity)
 		if err != nil {
 			return err
 		}
 		rs[r.Name] = r
 		return nil
```

yasker: I meant we shouldn't need to `return nil` here.

PhanLe1010: Yeah, we don't need to return nil. Thank you

controller/volume_controller.go
```diff
@@ -648,8 +652,8 @@ func (vc *VolumeController) cleanupFailedToScheduledReplicas(v *longhorn.Volume,
 	if healthyCount >= v.Spec.NumberOfReplicas {
 		for _, r := range rs {
 			if !hasEvictionRequestedReplicas {
-				if r.Spec.HealthyAt == "" && r.Spec.FailedAt == "" &&
-					r.Spec.NodeID == "" {
+				if r.Spec.HealthyAt == "" && r.Spec.FailedAt == "" && r.Spec.NodeID == "" &&
+					(isDataLocalityDisabled(v) || r.Spec.HardNodeAffinity != e.Spec.NodeID) {
```
shuo-wu, line 656: Can we directly use `r.Spec.HardNodeAffinity != v.Status.CurrentNodeID` here?

controller/volume_controller.go
```diff
@@ -811,6 +815,7 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
 			break
 		}
 		rs[r.Name] = scheduledReplica
+
```
shuo-wu, line 818: Remove this empty line

shuo-wu (changes requested):

controller/volume_controller.go
PhanLe1010: Yes, we can. Fixed. Thank you

controller/volume_controller.go
PhanLe1010: Thank you! Fixed

PhanLe1010:
@yasker @shuo-wu 

When testing, I encountered this problem with the original push:

1. Cluster of 3 nodes. Disable scheduling for node-3
1. Create a vol with 1 replica, dataLocality = best-effort. The replica is scheduled on node-1
1. Attach vol to node-3. There is a fail-to-schedule replica with `HardNodeAffinity=node-3`
1. Increase numberOfReplica to 2. The replica set is unchanged (e.g. one on node-1 and one failed on node-3). 
1. Increase numberOfReplica to 3. The replica set is unchanged (e.g. one on node-1 and one failed on node-3). 
1. Turn off data locality by set `dataLocality=disabled` for the vol. The replica set is unchanged (e.g. one on node-1 and one failed on node-3). 

However, I am expecting there is a new replica on node-2 at step 5 and 6. Is it correct?

The reason that the set of replicas is unchanged in steps 5 and 6 is: the failed-to-schedule replica on node-3 is blocking the creation of new replica. Therefore, I made 2 new changes since the original push:

1. add 1 more check `r.Spec.NodeID != ""`: https://github.com/longhorn/longhorn-manager/blob/0d85d9f8343fe411d8cd3f522697ab70ba6b0d8a/controller/volume_controller.go#L1314
1. Don't stop when failing to schedule a replica, keep going and schedule other replicas: https://github.com/longhorn/longhorn-manager/blob/0d85d9f8343fe411d8cd3f522697ab70ba6b0d8a/controller/volume_controller.go#L807

How do you think?

controller/volume_controller.go
```diff
@@ -801,16 +805,21 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
...
-			logrus.Errorf("unable to schedule replica %v of volume %v", r.Name, v.Name)
-			v.Status.Conditions = types.SetCondition(v.Status.Conditions,
-				types.VolumeConditionTypeScheduled, types.ConditionStatusFalse,
-				types.VolumeConditionReasonReplicaSchedulingFailure, "")
+			if r.Spec.HardNodeAffinity == "" {
+				logrus.Errorf("unable to schedule replica %v of volume %v", r.Name, v.Name)
+				v.Status.Conditions = types.SetCondition(v.Status.Conditions,
+					types.VolumeConditionTypeScheduled, types.ConditionStatusFalse,
+					types.VolumeConditionReasonReplicaSchedulingFailure, "")
+			} else {
```
PhanLe1010, line 813: Update Volume's condition to notify user when Longhorn fail to keep a local replica.

controller/volume_controller.go
```diff
@@ -801,16 +805,21 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
...
+				logrus.Errorf("unable to schedule replica %v of volume %v with HardNodeAffinity = %v", r.Name, v.Name, r.Spec.HardNodeAffinity)
+				v.Status.Conditions = types.SetCondition(v.Status.Conditions,
+					types.VolumeConditionTypeScheduled, types.ConditionStatusFalse,
+					types.VolumeConditionReasonLocalReplicaSchedulingFailure, "")
```
PhanLe1010, line 817: I update volume's condition to notify user when Longhorn fail to keep a local replica.

boknowswiki:
> @yasker @shuo-wu
> 
> When testing, I encountered this problem with the original push:
> 
> 1. Cluster of 3 nodes. Disable scheduling for node-3
> 2. Create a vol with 1 replica, `dataLocality = best-effort`. The replica is scheduled on node-1
> 3. Attach vol to node-3. There is a fail-to-schedule replica with `HardNodeAffinity=node-3`
> 4. Increase numberOfReplica to 2. The replica set is unchanged (e.g. one on node-1 and one failed on node-3).
> 5. Increase numberOfReplica to 3. The replica set is unchanged (e.g. one on node-1 and one failed on node-3).
> 6. Turn off data locality by set `dataLocality=disabled` for the vol. The replica set is unchanged (e.g. one on node-1 and one failed on node-3).
> 
> However, I am expecting there is a new replica on node-2 at step 5 and 6. Is it correct?
> 
> The reason that the set of replicas is unchanged in steps 5 and 6 is: the failed-to-schedule replica on node-3 is blocking the creation of new replica. Therefore, I made 2 new changes since the original push:
> 
> 1. We don't consider an unscheduled replica as a healthy replica. So, add 1 more check `r.Spec.NodeID != ""`: https://github.com/longhorn/longhorn-manager/blob/0d85d9f8343fe411d8cd3f522697ab70ba6b0d8a/controller/volume_controller.go#L1314
> 2. Before, we stop scheduling other replicas when we fail to schedule a replica because we know that other replicas would also fail. However, in the current situation, when we fail to schedule a replica, it doesn't necessarily mean that other replicas would also fail. Therefore, don't stop when failing to schedule a replica, keep going and schedule other replicas: https://github.com/longhorn/longhorn-manager/blob/0d85d9f8343fe411d8cd3f522697ab70ba6b0d8a/controller/volume_controller.go#L807
> 
> How do you think?

I am also expecting there is a new replica on node-2 at step 5 and 6. I think your idea is good. Say how @yasker and @shuo-wu think.

boknowswiki:
LGTM.

boknowswiki (approved):

yasker:
@PhanLe1010 I expected the same, the failure to schedule a local replica shouldn't block other replicas from scheduling.

Also, on step 4, we should have another replica rebuilding too? Since failed to schedule local replica shouldn't be counted.

controller/volume_controller.go
```diff
@@ -1363,14 +1312,16 @@ func (vc *VolumeController) getReplenishReplicasCount(v *longhorn.Volume, rs map
...
+		return 0
+	}
 	return v.Spec.NumberOfReplicas - usableCount
 }
 
 func (vc *VolumeController) hasEngineStatusSynced(e *longhorn.Engine, rs map[string]*longhorn.Replica) bool {
 	healthyCount := 0
 	for _, r := range rs {
-		if r.Spec.FailedAt == "" {
+		if r.Spec.FailedAt == "" && r.Spec.NodeID != "" {
```
yasker, line 1324: We should change the name `healthyCount` since it's not really a healthy count. It's more like a `connectedReplicaCount`.

yasker (changes requested):

controller/volume_controller.go
PhanLe1010: Thank you! Fixed

PhanLe1010:
@yasker 
> Also, on step 4, we should have another replica rebuilding too? Since failed to schedule local replica shouldn't be counted.

Yeah. I will fix this now

PhanLe1010:
@yasker 
> Also, on step 4, we should have another replica rebuilding too? Since failed to schedule local replica shouldn't be counted.
 
Fixed

controller/volume_controller.go
```diff
@@ -1358,24 +1307,32 @@ func getRebuildingReplicaCount(e *longhorn.Engine) int {
 func (vc *VolumeController) getReplenishReplicasCount(v *longhorn.Volume, rs map[string]*longhorn.Replica) int {
 	usableCount := 0
 	for _, r := range rs {
+		// The failed to schedule local replica shouldn't be counted
+		if r.Spec.HealthyAt == "" && r.Spec.FailedAt == "" && r.Spec.NodeID == "" &&
```
yasker, line 1311: Put `!isDataLocalityDisabled(v)` in the first condition so we know it's designed to deal with the volume with data locality.

yasker (approved):
LGTM, except for the minor coding style issue.

yasker:
> @yasker
> 
> > Also, on step 4, we should have another replica rebuilding too? Since failed to schedule local replica shouldn't be counted.
> 
> Yeah. I will fix this now. This is a bit tricky 😂

We should add those two cases in the automation test case as well.

controller/volume_controller.go

PhanLe1010:
> We should add those two cases in the automation test case as well.

Yes, I will

controller/volume_controller.go
```diff
@@ -1317,8 +1258,16 @@ func (vc *VolumeController) replenishReplicas(v *longhorn.Volume, e *longhorn.En
...
+	if hardNodeAffinity != "" && getRebuildingReplicaCount(e) == 0 {
+		r, err := vc.createReplica(v, e, rs, hardNodeAffinity)
+		if err != nil {
+			return err
+		}
+		rs[r.Name] = r
+	}
+
 	replenishCount := vc.getReplenishReplicasCount(v, rs)
-	if len(rs) != 0 && replenishCount != 0 {
```
PhanLe1010, line 1321: marked point

controller/volume_controller.go
```diff
@@ -1358,24 +1307,32 @@ func getRebuildingReplicaCount(e *longhorn.Engine) int {
...
+			v.Status.CurrentNodeID != "" && r.Spec.HardNodeAffinity == v.Status.CurrentNodeID {
+			continue
+		}
 		// Skip the replica has been requested eviction.
 		if r.Spec.FailedAt == "" && (!r.Status.EvictionRequested) {
 			usableCount++
 		}
 	}
 
+	if v.Spec.NumberOfReplicas < usableCount {
```
PhanLe1010, line 1321: we need this check because when user decrease NumberOfReplicas, this function might return a negative number. With the combination with the check https://github.com/longhorn/longhorn-manager/pull/678/files#r483810861, the Longhorn will create infinite number of replicas.

yasker (approved):

yasker:
It seems @shuo-wu 's comments have been addressed as well.

PR #679, opened 2020-09-09
Prometheus support, Part 1

PhanLe1010:
Added the following metrics:
1. longhorn_volume_capacity_bytes
1. longhorn_volume_actual_size_bytes
1. longhorn_node_stats_node_status

longhorn/longhorn#1180


api/router.go
```diff
@@ -1,6 +1,7 @@
 package api
 
 import (
+	"github.com/longhorn/longhorn-manager/monitoring"
```
joshimoo, line 4: add new section after `apierrors "k8s.io/apimachinery/pkg/api/errors"` a new line acts as a separator between the sections.

monitoring/backup_stats_number_failed_backups_metric.go
```diff
@@ -0,0 +1,47 @@
...
+	defer func() {
+		err = errors.Wrap(err, "failed to get values for numberFailedBackupsMetric")
+	}()
+
+	volumeLists, err := c.lhClient.LonghornV1beta1().Volumes(c.namespace).List(metav1.ListOptions{})
+	if err != nil {
+		return err
+	}
+
+	for _, v := range volumeLists.Items {
```
joshimoo, line 40: Don't store these persistently, in the objects, this will create a lot of update and writes which are completely unnecessary. 

Metrics should only exist at runtime in memory => hashmap
Prometheus can collect these, from your metrics endpoints.

manager/volume.go
```diff
@@ -741,3 +741,39 @@ func (m *VolumeManager) UpdateDataLocality(name string, dataLocality types.DataL
 	logrus.Debugf("Updated volume %v data locality from %v to %v", v.Name, oldDataLocality, v.Spec.DataLocality)
 	return v, nil
 }
+
+func (m *VolumeManager) IncBackupFailedCount(name string) (v *longhorn.Volume, err error) {
```
joshimoo, line 745: Remove these methods and add a metrics interface, then have metric class implementations in a datastore class.

example:
metrics.backup.succeeded(volumeName)
metrics.backup.failed(volumeName)

metrics is a singleton instance that contains references to the actual metric interface implementation.
this looks nice and makes it easy to split the metrics across different files.

Don't store these persistently, in the objects, this will create a lot of update and writes which are completely unnecessary. See below on how to store these in memory.

Also don't return any errors from these methods, metrics are best effort only.

monitoring/backup_stats_number_succeed_backups_metric.go
```diff
@@ -0,0 +1,47 @@
+package monitoring
+
+import (
+	"github.com/pkg/errors"
```
joshimoo, line 4: add different sections for import

monitoring/backup_stats_number_failed_backups_metric.go
joshimoo: applies to all the other metrics as well, also see implementation comment above for the increment methods etc.

monitoring/collector.go
```diff
@@ -0,0 +1,105 @@
+package monitoring
+
+import (
+	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta1"
+	lhclientset "github.com/longhorn/longhorn-manager/k8s/pkg/client/clientset/versioned"
+	"github.com/longhorn/longhorn-manager/types"
+	"github.com/pkg/errors"
```
joshimoo, line 7: add different sections for import

monitoring/collector.go
```diff
@@ -0,0 +1,105 @@
...
+	"os"
+)
+
+type Collector struct {
+	kubeClient    *clientset.Clientset
+	lhClient      *lhclientset.Clientset
+	namespace     string
+	currentNodeID string
+
+	log *logrus.Logger
```
joshimoo, line 23: Use logger as name to be consistent with the different controllers.
use logrus.FieldLogger as type, and pass in the base logger in the new NewCollector() call
see format of the different controllers as example.

Then you can add a method, if you need per metric logging with different fields for example.
log := getLoggerForMetrics(collector.logger, metric)

monitoring/monitoring.go
```diff
@@ -0,0 +1,22 @@
+package monitoring
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
```
joshimoo, line 4: add different sections.

joshimoo (changes requested):
Will review the rest after you had a chance to address this feedback.

api/router.go
PhanLe1010: Thank you! Fixed

api/snapshot.go
```diff
@@ -143,7 +143,11 @@ func (s *Server) SnapshotRevert(w http.ResponseWriter, req *http.Request) (err e
 }
 
 func (s *Server) SnapshotBackup(w http.ResponseWriter, req *http.Request) (err error) {
+	var volName string
 	defer func() {
+		if err != nil && volName != "" {
+			_, err = s.m.IncBackupFailedCount(volName)
```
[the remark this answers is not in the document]
PhanLe1010: Thank you! Fix so that we only count at at the manager level

manager/volume.go
PhanLe1010: Follow your suggestion, I created a new collector for each type (volumeCollector, backupCollector, nodeCollector, etc..) and have a common baseCollector. 

> Don't store these persistently, in the objects, this will create a lot of update and writes which are completely unnecessary. See below on how to store these in memory.

We will discuss more about this on the meeting tommorow

monitoring/backup_stats_number_failed_backups_metric.go
```diff
@@ -0,0 +1,47 @@
+package monitoring
+
+import (
+	"github.com/pkg/errors"
```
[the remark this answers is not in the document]
PhanLe1010: Thank you! This file was deleted

monitoring/backup_stats_number_failed_backups_metric.go
PhanLe1010: We will discuss more about this on the meeting tommorow

monitoring/backup_stats_number_succeed_backups_metric.go
PhanLe1010: Thank you! This file was deleted

monitoring/collector.go
PhanLe1010: Thank you! Fixed

monitoring/monitoring.go
PhanLe1010: Thank you! Fixed

monitoring/collector.go
PhanLe1010: Thank you! Fixed to use logging structure

PhanLe1010:
Thank you Joshua! I have fixed 4/5 suggestions.

Things that I fixed:
1. logging structure
1. use DS instead of calling etcd API directly in all metrics
1. Create a new collector for each type (volumeCollector, backupCollector, nodeCollector, etc..) and have a common baseCollector. This structure is similar to controllers. We have volumeController, nodeController, etc.. which have a commend baseController
1. rewrite the nodeStatus metrics so that it use 1 more dimension for different type of conditions. e.g longhorn_node_status{node_id="node1", condition="ready"}, longhorn_node_status{node_id="node1", condition="schedulabe"}, etc..

One thing that I would like to discuss more:
The best way to export information about backups. Option 1: expose the number of failed backup and succeed backups for each volume. Option 2: expose backup status for each backup as Portworx does: https://docs.portworx.com/reference/metrics/ 

- If we chose Option 1 above, what is the best way to present the metric to users. And what would be the best way to persist the data across diffrent scrapes.
- Option 2 seems to be very helpful from users' perspective but hard to implement with the current Longhorn's implementation

monitoring/volume_collector.go
```diff
@@ -0,0 +1,64 @@
...
+
+func (vc *VolumeCollector) Describe(ch chan<- *prometheus.Desc) {
+	ch <- vc.volumeCapacityMetric.Desc
+	ch <- vc.volumeActualSizeMetric.Desc
+}
+
+func (vc *VolumeCollector) Collect(ch chan<- prometheus.Metric) {
+	volumeLists, err := vc.ds.ListVolumesRO()
+	if err != nil {
+		vc.logger.Errorf("error during scrape %v", err)
```
joshimoo, line 56: WithError(err)

joshimoo (commented):
We can discuss the backup stuff in zoom later.

monitoring/node_stats_node_status_metric.go
```diff
@@ -0,0 +1,76 @@
+package monitoring
+
+import (
+	"github.com/longhorn/longhorn-manager/types"
```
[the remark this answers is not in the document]
PhanLe1010: This is outdate. This file was deleted

monitoring/node_stats_node_status_metric.go
```diff
@@ -0,0 +1,76 @@
...
+func (m nodeStatusMetric) describe() *prometheus.Desc {
+	return m.Desc
+}
+
+func (m nodeStatusMetric) getMetricValues(ch chan<- prometheus.Metric, c *Collector) (err error) {
+	defer func() {
+		err = errors.Wrap(err, "failed to get values for nodeStatusMetric")
+	}()
+
+	nodeLists, err := c.lhClient.LonghornV1beta1().Nodes(c.namespace).List(metav1.ListOptions{
```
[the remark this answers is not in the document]
PhanLe1010: This is outdate. This file was deleted

monitoring/monitoring.go
```diff
@@ -0,0 +1,37 @@
...
+func MustRegister(collector prometheus.Collector) {
+	longhornCustomRegistry.MustRegister(collector)
+}
+
+// Handler returns an http.Handler for longhornCustomRegistry, using default HandlerOpts
+func Handler() http.Handler {
+	return promhttp.HandlerFor(longhornCustomRegistry, promhttp.HandlerOpts{})
+}
+
+func InitMonitoringSystem(logger logrus.FieldLogger, ds *datastore.DataStore, currentNodeID string) {
```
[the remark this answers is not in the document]
PhanLe1010: Thank you for sharing! It is awsome!

monitoring/node_collector.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+func (nc *NodeCollector) Describe(ch chan<- *prometheus.Desc) {
+	ch <- nc.newNodeStatusMetric.Desc
+}
+
+func (nc *NodeCollector) Collect(ch chan<- prometheus.Metric) {
+	nodeList, err := nc.ds.ListNodesRO()
+	if err != nil {
+		nc.logger.Errorf("error during scrape %v", err)
```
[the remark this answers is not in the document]
PhanLe1010: Thank you! I fixed it

monitoring/volume_collector.go
PhanLe1010: Thank you! Fixed

datastore/longhorn.go
```diff
@@ -1032,6 +1032,11 @@ func (s *DataStore) ListNodes() (map[string]*longhorn.Node, error) {
 	return itemMap, nil
 }
 
+// ListNodesRO returns a list of all Nodes for the given namespace
```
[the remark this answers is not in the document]
PhanLe1010: Thank you for sharing! I fixed it

PhanLe1010:
Removed the backup_collector file. This PR only contains volume capacity, volume actual size, and node status metrics.

PhanLe1010:
@joshimoo Can you please take a look at this PR again?

monitoring/types.go
```diff
@@ -0,0 +1,22 @@
...
+)
+
+const (
+	longhornName = "longhorn"
+
+	subsystemVolume      = "volume"
+	subsystemBackupStats = "backup_stats"
+	subsystemNodeStats   = "node_stats"
+
+	prometheusNodeIDLabel    = "node_id"
```
joshimoo, line 14: remove the `id` suffix and just use `node, volume, condition, condition_reason`

monitoring/types.go
```diff
@@ -0,0 +1,22 @@
...
+import (
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+const (
+	longhornName = "longhorn"
+
+	subsystemVolume      = "volume"
+	subsystemBackupStats = "backup_stats"
+	subsystemNodeStats   = "node_stats"
```
joshimoo, line 12: remove the `stats` suffix and just use `node`
this way it will look nicer and be easier to read:

`node_status` vs `node_stats_node_status`

this also makes sense since all our per component collectors are individual metrics and not aggregates.
If we want to add different aggregated statistics we could add a collector like this instead:
`stats_nodes`
`stats_volumes`

But at the moment I don't see any direct need for this, since once all individual metrics are tracked via prometheus the user can use the prometheus query language to generate the aggregates themselves.
Which is much better then us generating them.

monitoring/types.go
```diff
@@ -0,0 +1,22 @@
...
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+const (
+	longhornName = "longhorn"
+
+	subsystemVolume      = "volume"
+	subsystemBackupStats = "backup_stats"
```
joshimoo, line 11: remove the backup related field for now, since we are not doing it till later.

monitoring/node_collector.go
```diff
@@ -0,0 +1,77 @@
...
+
+func (nc *NodeCollector) collectNodeStatus(ch chan<- prometheus.Metric, nodeList []*longhorn.Node) {
+	for _, node := range nodeList {
+		if node.Name != nc.currentNodeID {
+			continue
+		}
+
+		for _, conditionType := range []string{types.NodeConditionTypeReady, types.NodeConditionTypeSchedulable, types.NodeConditionTypeMountPropagation} {
+			val := 0
+			switch condition := types.GetCondition(node.Status.Conditions, conditionType); condition.Status {
```
joshimoo, line 66: Use binary conditions:
False/Unkown => 0
True => 1

monitoring/node_collector.go
```diff
@@ -0,0 +1,77 @@
...
+			case types.ConditionStatusTrue:
+				val = 1
+			case types.ConditionStatusFalse:
+				val = 2
+			case types.ConditionStatusUnknown:
+				val = 3
+			}
+			ch <- prometheus.MustNewConstMetric(nc.newNodeStatusMetric.Desc, nc.newNodeStatusMetric.Type, float64(val), nc.currentNodeID, strings.ToLower(conditionType))
```
joshimoo, line 74: Add the condition_reason values as a dimension.
This allows for example to find how often a node became not ready because of network failure, cpu failure, etc.

monitoring/node_collector.go
```diff
@@ -0,0 +1,77 @@
...
+
+}
+
+func (nc *NodeCollector) collectNodeStatus(ch chan<- prometheus.Metric, nodeList []*longhorn.Node) {
+	for _, node := range nodeList {
+		if node.Name != nc.currentNodeID {
+			continue
+		}
+
+		for _, conditionType := range []string{types.NodeConditionTypeReady, types.NodeConditionTypeSchedulable, types.NodeConditionTypeMountPropagation} {
```
joshimoo, line 64: Instead of checking for specific conditions, return all node conditions, by iterating over the node.Status.Conditions map.

joshimoo: The same approach can be used for Volume Conditions, Disk Conditions, etc down the line.

monitoring/node_collector.go
```diff
@@ -0,0 +1,77 @@
...
+
+	"github.com/longhorn/longhorn-manager/datastore"
+	"github.com/longhorn/longhorn-manager/types"
+
+	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta1"
+)
+
+type NodeCollector struct {
+	*baseCollector
+	newNodeStatusMetric metricInfo
```
joshimoo, line 18: Why is this called `newNodeStatusMetric`?
could it be called just `statusMetric` since it's private and inside of the NodeCollector?

monitoring/volume_collector.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/sirupsen/logrus"
+
+	"github.com/prometheus/client_golang/prometheus"
+
+	"github.com/longhorn/longhorn-manager/datastore"
+)
+
+type VolumeCollector struct {
+	*baseCollector
+	volumeCapacityMetric   metricInfo
```
joshimoo, line 13: maybe just `capacityMetric`?

monitoring/volume_collector.go
```diff
@@ -0,0 +1,65 @@
...
+	volumeActualSizeMetric metricInfo
```
joshimoo, line 14: maybe just `sizeMetric`?

monitoring/volume_collector.go
```diff
@@ -0,0 +1,65 @@
...
+			"Configured size in bytes for this volume",
+			[]string{prometheusNodeIDLabel, prometheusVolumeIDLabel},
+			nil,
+		),
+		Type: prometheus.GaugeValue,
+	}
+
+	vc.volumeActualSizeMetric = metricInfo{
+		Desc: prometheus.NewDesc(
+			prometheus.BuildFQName(longhornName, subsystemVolume, "actual_size_bytes"),
```
joshimoo, line 38: maybe just `used_bytes` ? 

This can be used to get the ratio of used / capacity and do alerting in the case when there are to many snapshots for a volume for example. (i.e. once ratio > 5 for example)

joshimoo (commented):
as discussed in zoom, looks good to me in general, some smaller naming suggestions below.

monitoring/volume_collector.go
joshimoo: I see elsewhere you use the term `usage` instead, so let's stick with that?

monitoring/types.go
PhanLe1010: Thank you! Fixed

monitoring/types.go
PhanLe1010: Thank you! Fixed

monitoring/types.go
PhanLe1010: Thank you! Fixed

monitoring/node_collector.go
PhanLe1010: Thank you! Fixed

monitoring/node_collector.go
PhanLe1010: Thank you! Fixed

monitoring/node_collector.go
PhanLe1010: Thank you! Fixed

monitoring/node_collector.go
PhanLe1010: Opps. this is a typo. Thank you! Fixed

monitoring/volume_collector.go
PhanLe1010: Thank you! Fixed

monitoring/volume_collector.go
PhanLe1010: Thank you! Fixed

monitoring/volume_collector.go
PhanLe1010: Thank you! Fixed

monitoring/node_collector.go
```diff
@@ -0,0 +1,72 @@
...
+		if node.Name != nc.currentNodeID {
+			continue
+		}
+
+		for _, condition := range node.Status.Conditions {
+			val := 0
+			if condition.Status == types.ConditionStatusTrue {
+				val = 1
+			}
+			ch <- prometheus.MustNewConstMetric(nc.statusMetric.Desc, nc.statusMetric.Type, float64(val), nc.currentNodeID, strings.ToLower(condition.Type), condition.Reason)
```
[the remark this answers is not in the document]
joshimoo: As an alternative you can also just use
`defer func() { recover() }` at the top of the collectNodeStatus function.
This will swallow the panic, and stop execution of the collectNodeStatus function.

In there you could then log once,  instead of per node / condition / etc.
This is probably a simpler design, since generally these shouldn't panic anyway.

monitoring/volume_collector.go
```diff
@@ -0,0 +1,65 @@
...
+
+func (vc *VolumeCollector) Collect(ch chan<- prometheus.Metric) {
+	volumeLists, err := vc.ds.ListVolumesRO()
+	if err != nil {
+		vc.logger.WithError(err).Warn("error during scrape")
+		return
+	}
+	for _, v := range volumeLists {
+		if v.Status.OwnerID == vc.currentNodeID {
+			ch <- prometheus.MustNewConstMetric(vc.capacityMetric.Desc, vc.capacityMetric.Type, float64(v.Spec.Size), vc.currentNodeID, v.Name)
```
joshimoo, line 61: see comment above about panicking, I think using the `defer func { recover() }` approach is okay for the collection methods.

joshimoo (changes requested):
Added concerns about the panicking behavior.

monitoring/volume_collector.go
PhanLe1010: Thank you! Fixed on the https://github.com/longhorn/longhorn-manager/pull/684 PR

monitoring/node_collector.go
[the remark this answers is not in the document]
PhanLe1010: Yeah!!! Thank you! Fixed on the https://github.com/longhorn/longhorn-manager/pull/684 PR

monitoring/node_collector.go
```diff
@@ -0,0 +1,72 @@
...
+	statusMetric metricInfo
+}
+
+func NewNodeCollector(
+	logger logrus.FieldLogger,
+	nodeID string,
+	ds *datastore.DataStore) *NodeCollector {
+
+	nc := &NodeCollector{
+		baseCollector: newBaseCollector("node", logger, nodeID, ds),
```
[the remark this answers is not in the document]
PhanLe1010: Thank you! Fixed on the https://github.com/longhorn/longhorn-manager/pull/684 PR

monitoring/monitoring.go
```diff
@@ -0,0 +1,35 @@
...
+)
+
+// longhornCustomRegistry exposes Longhorn metrics only.
+// We use longhornCustomRegistry to get rid of all default Prometheus go-client metrics
+var longhornCustomRegistry = prometheus.NewRegistry()
+
+// MustRegister registers the provided Collectors with the longhornCustomRegistry and
+// panics if any error occurs.
+func MustRegister(collector prometheus.Collector) {
+	longhornCustomRegistry.MustRegister(collector)
```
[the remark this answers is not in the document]
PhanLe1010: Thank you! Fixed on the https://github.com/longhorn/longhorn-manager/pull/684 PR

joshimoo:
superseded by PR #684

PR #680, opened 2020-09-10
Refactor setting controller for structured log

c3y1huang:
Refactor setting controller to address structured logging implementation.

Ref:
* https://github.com/longhorn/longhorn/issues/1732
* https://github.com/longhorn/longhorn-manager/pull/661

### DIFF
--- a/controller/setting_controller.go
+++ b/controller/setting_controller.go
@@ -251,6 +251,7 @@ func (sc *SettingController) syncBackupTarget() (err error) {
 			// already monitoring
 			return nil
 		}
+		sc.logger.Infof("Restarting backup store monitor because backup target changed from %v to %v", sc.bsMonitor.backupTarget, targetSetting.Value)
 		sc.bsMonitor.Stop()
 		sc.bsMonitor = nil
 		manager.SyncVolumesLastBackupWithBackupVolumes(nil,
--- a/controller/setting_controller.go
+++ b/controller/setting_controller.go
@@ -134,8 +134,8 @@ func (sc *SettingController) Run(stopCh <-chan struct{}) {
 	defer utilruntime.HandleCrash()
 	defer sc.queue.ShutDown()
 
-	logrus.Infof("Start Longhorn Setting controller")
-	defer logrus.Infof("Shutting down Longhorn Setting controller")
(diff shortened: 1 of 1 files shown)


joshimoo (changes requested):
Changes look good, while doing this consider places where you can improve the logs by adding additional logging.

In this case log anytime the backup target / backupstore monitor get's restarted, if that had been done before it would have been quite apparent that there is a bug in the backupstore monitor :)

see for explanation of the bug: https://github.com/longhorn/longhorn/issues/1547#issuecomment-678781631

joshimoo:
I will let @PhanLe1010 review this, since he will have to implement logging for the metrics side and this gives a quick contrast between the old style implementation and the new style via instance based loggers.

PhanLe1010:
It looks good to me.

c3y1huang:
> Changes look good, while doing this consider places where you can improve the logs by adding additional logging.
> 
> In this case log anytime the backup target / backupstore monitor get's restarted, if that had been done before it would have been quite apparent that there is a bug in the backupstore monitor :)
> 
> For an explanation of the bug see: [longhorn/longhorn#1547 (comment)](https://github.com/longhorn/longhorn/issues/1547#issuecomment-678781631)

Not sure if I've addressed entirely, PTAL https://github.com/longhorn/longhorn-manager/pull/680/commits/79b86b834275afcb0f211a57610679741d08ba6c.

controller/setting_controller.go
```diff
@@ -251,6 +251,7 @@ func (sc *SettingController) syncBackupTarget() (err error) {
 			// already monitoring
 			return nil
 		}
+		sc.logger.Warn("Restart backup store monitoring")
```
joshimoo, line 254:
```suggestion
		sc.logger.Info("restarting backup store monitor because backup target changed from %v to %v", sc.bsMonitor.backupTarget, targetSetting.Value)
```

joshimoo (approved):
@c3y1huang Sorry for the late response, can you change the backupstore change message to include the backup target information. Then this can be merged.

joshimoo (approved):

joshimoo:
@c3y1huang merged thanks for opening and working on this PR, it's appreciated 👍

PR #684, opened 2020-09-15
app, datastore, deploy, monitoring: Prometheus support, part 2

PhanLe1010:
Added the following metrics:
    1. longhorn_instance_manager_requests_cpu_millicpu
    2. longhorn_instance_manager_requests_memory_bytes
    3. longhorn_instance_manager_usage_cpu_millicpu
    4. longhorn_instance_manager_usage_memory_bytes
    5. longhorn_manager_usage_cpu_millicpu
    6. longhorn_manager_usage_memory_bytes

longhorn/longhorn#1180


monitoring/instance_manager_collector.go
```diff
@@ -0,0 +1,174 @@
...
+	"github.com/longhorn/longhorn-manager/types"
+)
+
+type InstanceManagerCollector struct {
+	*baseCollector
+
+	kubeMetricsClient *metricsclientset.Clientset
+	namespace         string
+
+	instanceManagerUsageCPUMetric      metricInfo
```
joshimoo, line 24: How about flipping these and just using the below?

cpuUsageMetric`
`cpuRequestMetric`
`memoryUsageMetric`
`memoryRequestMetric`

monitoring/types.go
```diff
@@ -7,13 +7,18 @@ import (
...
+	subsystemVolume          = "volume"
+	subsystemBackupStats     = "backup_stats"
+	subsystemNodeStats       = "node_stats"
+	subsystemInstanceManager = "instance_manager"
+	subsystemManager         = "manager"
 
-	prometheusNodeIDLabel    = "node_id"
-	prometheusVolumeIDLabel  = "volume_id"
-	prometheusConditionLabel = "condition"
+	prometheusNodeIDLabel            = "node_id"
```
joshimoo, line 16: you can just use NodeLabel ..... InstanceManagerLabel, ManagerLabel, etc

monitoring/types.go
```diff
@@ -7,13 +7,18 @@ import (
...
 	longhornName = "longhorn"
 
-	subsystemVolume      = "volume"
-	subsystemBackupStats = "backup_stats"
-	subsystemNodeStats   = "node_stats"
+	subsystemVolume          = "volume"
+	subsystemBackupStats     = "backup_stats"
+	subsystemNodeStats       = "node_stats"
+	subsystemInstanceManager = "instance_manager"
+	subsystemManager         = "manager"
```
joshimoo, line 14: These can then be remove and instead use NodeLabel, InstanceManagerLabel, etc from below.

monitoring/monitoring.go
```diff
@@ -26,10 +33,33 @@ func Handler() http.Handler {
...
+	}
+
+	config, err := clientcmd.BuildConfigFromFlags("", kubeconfigPath)
+	if err != nil {
+		return errors.Wrap(err, "unable to get client config. Failed to init the monitoring system")
+	}
+
+	kubeMetricsClient, err := metricsclientset.NewForConfig(config)
+	if err != nil {
+		return errors.Wrap(err, "unable to get k8s metrics client. Failed to init the monitoring system")
```
joshimoo, line 51: Do we want to fail all metrics collection in this case or just proceed with the setup then, log an error in the 
`InstanceManagerCollector` as well as the `ManagerCollector`?

The constructor can then just have an if kubeMetricsClient is nil -> log
and the collect method can just have an  if kubeMetricsClient is nil -> do nothing

If we are saying that metrics are mandatory, then failing here is okay otherwise if metrics are best effort then we should proceed with regular longhorn functionality.

monitoring/manager_collector.go
```diff
@@ -0,0 +1,89 @@
...
+	"github.com/longhorn/longhorn-manager/datastore"
+)
+
+type ManagerCollector struct {
+	*baseCollector
+
+	kubeMetricsClient *metricsclientset.Clientset
+	namespace         string
+
+	managerUsageCPUMetric    metricInfo
```
joshimoo, line 22: how about `cpuUsageMetric` and `memoryUsageMetric`?

monitoring/manager_collector.go
```diff
@@ -0,0 +1,89 @@
...
+
+	mc := &ManagerCollector{
+		baseCollector:     newBaseCollector("manager", logger, nodeID, ds),
+		kubeMetricsClient: kubeMetricsClient,
+		namespace:         namespace,
+	}
+
+	mc.managerUsageCPUMetric = metricInfo{
+		Desc: prometheus.NewDesc(
+			prometheus.BuildFQName(longhornName, subsystemManager, "usage_cpu_millicpu"),
```
joshimoo, line 41: mentioned in LEP how about `cpu_usage_millicpu` and `cpu_request_millicpu` basically index by HW component then specific metric.

monitoring/instance_manager_collector.go
joshimoo: basically index by HW component then specific metric.

monitoring/instance_manager_collector.go
```diff
@@ -0,0 +1,174 @@
...
+
+	imc := &InstanceManagerCollector{
+		baseCollector:     newBaseCollector("instance-manager", logger, nodeID, ds),
+		kubeMetricsClient: kubeMetricsClient,
+		namespace:         namespace,
+	}
+
+	imc.instanceManagerUsageCPUMetric = metricInfo{
+		Desc: prometheus.NewDesc(
+			prometheus.BuildFQName(longhornName, subsystemInstanceManager, "usage_cpu_millicpu"),
```
joshimoo, line 45: Same, how about `cpu_usage_millicpu`?
Index by HW component then specific metric, same for the other fields below?

joshimoo: If you look at the sentence below, my point might be more clear?
the `cpu usage` vs the `usage cpu / usage of cpu`

app/daemon.go
```diff
@@ -131,7 +131,9 @@ func startManager(c *cli.Context) error {
 
 	m := manager.NewVolumeManager(currentNodeID, ds)
 
-	monitoring.InitMonitoringSystem(logger, currentNodeID, ds)
+	if err := monitoring.InitMonitoringSystem(logger, currentNodeID, ds, kubeconfigPath); err != nil {
```
joshimoo, line 134: See my question below, are metrics best effort or are we saying they are required for longhorn operations?
This is something to add to the LEP as well

joshimoo (commented):
LGTM in general, added some suggestions and a question.

monitoring/instance_manager_collector.go
PhanLe1010: Thank you! Fixed

monitoring/types.go
PhanLe1010: Thank you! Fixed

monitoring/types.go
PhanLe1010: I like to keep the `subsystem*` prefix for those labels because it makes the naming convention clear. e.g:
```
prometheus.BuildFQName(longhornName, InstanceManagerLabel, "cpu_usage_millicpu")
```
How do you think?

monitoring/monitoring.go
PhanLe1010: Great point! I agree that all metrics should be best-effort. Added the logic to skip instantiating InstanceManagerCollector and ManagerCollector  when failing to create kubeMetricsClient.

monitoring/manager_collector.go
PhanLe1010: Thank you! Fixed

monitoring/manager_collector.go
PhanLe1010: Thank you! Fixed

monitoring/instance_manager_collector.go
PhanLe1010: Thank you! Fixed

app/daemon.go
PhanLe1010: Great point! I agree that all metrics should be best-effort. Fixed InitMonitoringSystem so that it doesn't return an error.

monitoring/types.go
joshimoo: Fair enough, how about this instead?
instead of using `NodeLabel` and `subsystemNode` just use one field called `NodeComponent` for both?
Since you are effectively tracking metrics for a node, volume, manager, instance_manager.

If you prefer though you can also keep the subsystem but then consider putting the assignment on the same line as the NodeLabel
subsystemNode = nodeLabel = "node" if possible?

I have no hard feelings about this, as long as the actual value is the same as the individual component labels.

monitoring/instance_manager_collector.go
```diff
@@ -0,0 +1,174 @@
...
+	"github.com/longhorn/longhorn-manager/types"
+)
+
+type InstanceManagerCollector struct {
+	*baseCollector
+
+	kubeMetricsClient *metricsclientset.Clientset
+	namespace         string
+
+	instanceManagerCPUUsageMetric      metricInfo
```
joshimoo, line 24: Remove the instanceManager prefix, these fields are internal, so they don't add additional value.
Since this is an InstanceManagerCollector :)

joshimoo: it will also look nicer :)

`cpuUsageMetric` vs `CPUUsageMetric`

monitoring/manager_collector.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/longhorn/longhorn-manager/datastore"
+)
+
+type ManagerCollector struct {
+	*baseCollector
+
+	kubeMetricsClient *metricsclientset.Clientset
+	namespace         string
+
+	managerCPUUsageMetric    metricInfo
```
joshimoo, line 22: remove manager prefix same as with the other collectors.

joshimoo (commented):
LGTM remove the additional prefixes, then we should be good to go.

monitoring/instance_manager_collector.go
PhanLe1010: Thank you! Fixed

monitoring/manager_collector.go
PhanLe1010: Thank you! Fixed

monitoring/instance_manager_collector.go
```diff
@@ -0,0 +1,186 @@
...
+
+	wg.Wait()
+}
+
+func (imc *InstanceManagerCollector) collectActualUsage(ch chan<- prometheus.Metric, wg *sync.WaitGroup) {
+	defer wg.Done()
+	defer func() {
+		err := recover()
+		if err != nil {
+			imc.logger.WithField("panic error", err).Warn("panic during collecting metrics")
```
joshimoo, line 110: better to use withError since we dealt with the panic.

monitoring/instance_manager_collector.go
```diff
@@ -0,0 +1,186 @@
...
+	go imc.collectRequestValues(ch, &wg)
+
+	wg.Wait()
+}
+
+func (imc *InstanceManagerCollector) collectActualUsage(ch chan<- prometheus.Metric, wg *sync.WaitGroup) {
+	defer wg.Done()
+	defer func() {
+		err := recover()
+		if err != nil {
```
joshimoo, line 109: combine the declartion and if evaluation.

monitoring/instance_manager_collector.go
```diff
@@ -0,0 +1,186 @@
...
+	default:
+		return ""
+	}
+}
+
+func (imc *InstanceManagerCollector) collectRequestValues(ch chan<- prometheus.Metric, wg *sync.WaitGroup) {
+	defer wg.Done()
+	defer func() {
+		err := recover()
+		if err != nil {
```
joshimoo, line 156: combine

monitoring/instance_manager_collector.go
```diff
@@ -0,0 +1,186 @@
...
+			imc.logger.WithField("panic error", err).Warn("panic during collecting metrics")
```
joshimoo, line 157: withError

monitoring/manager_collector.go
```diff
@@ -0,0 +1,98 @@
...
+
+func (mc *ManagerCollector) Describe(ch chan<- *prometheus.Desc) {
+	ch <- mc.cpuUsageMetric.Desc
+	ch <- mc.memoryUsageMetric.Desc
+}
+
+func (mc *ManagerCollector) Collect(ch chan<- prometheus.Metric) {
+	defer func() {
+		err := recover()
+		if err != nil {
```
joshimoo, line 70: combine

monitoring/manager_collector.go
```diff
@@ -0,0 +1,98 @@
...
+			mc.logger.WithField("panic error", err).Warn("panic during collecting metrics")
```
joshimoo, line 71: withError

monitoring/manager_collector.go
```diff
@@ -0,0 +1,98 @@
...
+		}
+	}()
+
+	managerPodName, err := os.Hostname()
```
joshimoo, line 75: is the os hostname always guaranteed to be the manager pod name?

monitoring/volume_collector.go
```diff
@@ -52,11 +52,19 @@ func (vc *VolumeCollector) Describe(ch chan<- *prometheus.Desc) {
 }
 
 func (vc *VolumeCollector) Collect(ch chan<- prometheus.Metric) {
+	defer func() {
+		err := recover()
+		if err != nil {
+			vc.logger.WithField("panic error", err).Warn("panic during collecting metrics")
```
joshimoo, line 58: combine, withError

monitoring/node_collector.go
```diff
@@ -57,6 +57,13 @@ func (nc *NodeCollector) Collect(ch chan<- prometheus.Metric) {
 }
 
 func (nc *NodeCollector) collectNodeStatus(ch chan<- prometheus.Metric, nodeList []*longhorn.Node) {
+	defer func() {
+		err := recover()
+		if err != nil {
```
joshimoo, line 62: combine, withError

monitoring/monitoring.go
```diff
@@ -2,34 +2,81 @@ package monitoring
...
 
-func InitMonitoringSystem(logger logrus.FieldLogger, currentNodeID string, ds *datastore.DataStore) {
+func InitMonitoringSystem(logger logrus.FieldLogger, currentNodeID string, ds *datastore.DataStore, kubeconfigPath string) {
 	vc := NewVolumeCollector(logger, currentNodeID, ds)
 	nc := NewNodeCollector(logger, currentNodeID, ds)
 
-	MustRegister(vc)
-	MustRegister(nc)
+	if err := Register(vc); err != nil {
+		logger.WithError(err).Warn("failed to register collector")
```
joshimoo, line 40: withField("collector", subsystemTypeVolume)

same for the other ones below.

joshimoo (commented):
Looks good to me, address the mentioned suggestions.
Then squash Part 1 and two into a single commit.

joshimoo (approved):
LGTM, great work on building this feature and addressing the review suggestions :)

PR #685, opened 2020-09-16
508 replace revision counter manager

boknowswiki:
https://github.com/longhorn/longhorn/issues/508


controller/volume_controller.go
```diff
@@ -876,6 +881,14 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
 		v.Status.Robustness = types.VolumeRobustnessFaulted
 		v.Status.CurrentNodeID = ""
 
+		// Since all replica failed, mark engine controller salvage requested
+		e.Spec.SalvageRequested = true
+		logrus.Infof("All replicas are failed, set engine salvageRequested to %v", e.Spec.SalvageRequested)
+		if e.Status.SalvageExecuted {
+			logrus.Warnf("Engine has executed salvage, execute salvage again")
+			e.Status.SalvageExecuted = false
```
shuo-wu, line 889: Do not update engine status outside of the engine controller.

shuo-wu (changes requested):
Need to handle the case that the replica fails to start after the auto salvage.

boknowswiki:
> Need to handle the case that the replica fails to start after the auto salvage.

I think it will keep auto salvage.

go.mod
```diff
@@ -23,6 +23,7 @@ require (
 	github.com/evanphx/json-patch v4.5.0+incompatible // indirect
 	github.com/gogo/protobuf v1.3.1 // indirect
 	github.com/golang/groupcache v0.0.0-20191002201903-404acd9df4cc // indirect
+	github.com/golang/protobuf v1.3.3-0.20190920234318-1680a479a2cf
```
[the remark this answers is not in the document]
boknowswiki: I think this is updated during update the client. Anything wrong with it? Thanks.

controller/volume_controller.go
boknowswiki: Oh, i see. I will fix it. Thanks.

controller/volume_controller.go
```diff
@@ -876,6 +881,10 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
 		v.Status.Robustness = types.VolumeRobustnessFaulted
 		v.Status.CurrentNodeID = ""
 
+		// Since all replica failed, mark engine controller salvage requested
+		e.Spec.SalvageRequested = true
```
shuo-wu, line 885: Don't we need to check `autoSalvage` before setting this field?

shuo-wu (changes requested):
It's better to add unit tests for instance handler/volume controller.

controller/volume_controller.go
boknowswiki: Yes, good catch! Fix it. Thanks.

shuo-wu:
Where are the unit tests?

boknowswiki:
> Where are the new unit tests?
Oh, sorry, I think I need to sync with you about this part. Let's have a quick zoom tonight, is that ok? Thanks.

boknowswiki:
@shuo-wu , I have added two unit tests for volume controller, please take a look and let me know. Thanks.

controller/volume_controller_test.go
```diff
@@ -820,6 +821,72 @@ func (s *TestSuite) TestVolumeLifeCycle(c *C) {
...
+	tc = generateVolumeTestCaseTemplate()
+	tc.volume.Spec.NodeID = TestNode1
+	tc.volume.Status.State = types.VolumeStateDetached
+	tc.volumeAutoSalvage = "true"
+
+	for _, e := range tc.engines {
+		e.Status.CurrentState = types.InstanceStateStopped
+	}
+	for _, r := range tc.replicas {
+		r.Spec.FailedAt = getTestNow()
```
shuo-wu, line 868: The init state should be:
```
r.Status.CurrentState = types.InstanceStateError
r.Spec.healthyAt = getTestNow()
......
```

Then the desired state should be:
```
r.Spec.FailedAt = getTestNow()
r.Spec.DesiredState = types.InstanceStateStopped
......
```

shuo-wu (changes requested):

controller/volume_controller_test.go
boknowswiki: I think to trigger the auto-salvage, all replicas need to be failed from [this code](https://github.com/longhorn/longhorn-manager/blob/38c9c683ce63bf29afac0b0c8649626a6102dae0/controller/volume_controller.go#L869-L873). This is why I set the init state with 'FailedAt'.

shuo-wu: The prerequisite of the auto salvage is that all replicas become error. Considering the following workflow:
1. All replica processes become `error` simultaneously.
2. The instance handler in Replica Controller will update the `replica.status.state` to `error` accordingly. (Here, `replica.Spec.FailedAt` is still empty. And I think this should be the init state of this unit test.)
3. The Volume Controller get triggered. In the **single** sync loop:
    1. In `ReconcileEngineReplicaState`, `replica.Spec.FailedAt` will be set.
    2. In `ReconcileVolumeState`, the replicas will be salvaged immediately. Both `v.Status.CurrentNodeID` and `replica.Spec.FailedAt` [will be unset](https://github.com/longhorn/longhorn-manager/blob/38c9c683ce63bf29afac0b0c8649626a6102dae0/controller/volume_controller.go#L910).
    3. `e.Spec.DesireState = types.InstanceStateStopped` 
    ......

The state I provided above maybe not completely precise/correct. But you cannot use the transit state in the sync loop as the init state of a unit test

boknowswiki: Hi Shuo, thanks for explaining this in details. And it totally make sense. I have updated the PR, please take a look. Thanks.

controller/volume_controller_test.go
```diff
@@ -820,6 +821,87 @@ func (s *TestSuite) TestVolumeLifeCycle(c *C) {
...
+		expectRs[r.Name] = r
+	}
+	tc.expectReplicas = expectRs
+
+	testCases["volume revision counter disabled - engine and replica revision counter disabled"] = tc
+
+	// Salvage Requested
+	tc = generateVolumeTestCaseTemplate()
+	tc.volume.Spec.NodeID = TestNode1
+	tc.volume.Status.State = types.VolumeStateDetached
```
shuo-wu, line 861: `tc.volume.Status.State = types.VolumeStateAttached`

shuo-wu (changes requested):
Others LGTM

controller/volume_controller_test.go
boknowswiki: Hi Shuo, I think it need to be detached [code is here](https://github.com/longhorn/longhorn-manager/blob/e4f9956d28fde33027eb3dd9b69f3456cea55c17/controller/volume_controller.go#L884), since this is one sync call. In the real case, it can be attached, and need to become detached first, then be salvaged. Please correct me if i am wrong. Thanks.

shuo-wu: In the init state, the salvaged hasn't been executed. And how can a `detached` volume suddenly becomes `detaching`.

boknowswiki: I think you are right, probably there is something wrong with my PR, let me check it again. Thanks.

boknowswiki: @shuo-wu , I have updated my PR, also fixed [this part](https://github.com/longhorn/longhorn-manager/blob/d94f2f28d8b76041650b0d979e88825f7b240368/controller/volume_controller.go#L953-L959), please take a look and let me know. Thanks.

controller/volume_controller.go
```diff
@@ -954,7 +976,7 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
 		// the final state will be determined at the end of the clause
 		if newVolume {
 			v.Status.State = types.VolumeStateCreating
-		} else {
+		} else if v.Status.State != types.VolumeStateDetached {
```
boknowswiki, line 979: @shuo-wu , Here is the fix for the unit test why it changed from detached to detaching. Please help take a look. Thanks.

boknowswiki:
@shuo-wu , based on our discussion last night, I have updated the PR to make the `SalvageRequested` more general. Please take a look. Thanks.

controller/volume_controller.go
```diff
@@ -904,17 +910,27 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
...
+						// check revision counter of replicas
+						if v.Spec.RevisionCounterDisabled {
+							msg := fmt.Sprintf("Replica %v of volume %v has been automatically salvaged", r.Name, v.Name)
+							vc.eventRecorder.Event(v, v1.EventTypeWarning, EventReasonAutoSalvaged, msg)
+							salvaged = true
+						}
 					}
 				}
-				if salvaged {
+				if salvaged || e.Spec.SalvageRequested {
```
shuo-wu, line 933: Due to line 916, the result of this check is always `true`. It is meaningless. 

I think you can remove line 926 `if v.Spec.RevisionCounterDisabled {` then revert (the modification of) this line.

controller/volume_controller.go
```diff
@@ -904,17 +910,27 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
...
 					if util.TimestampWithinLimit(lastFailedAt, r.Spec.FailedAt, AutoSalvageTimeLimit) {
 						r.Spec.FailedAt = ""
-						log.WithField("replica", r.Name).Warn("Automatically salvaged volume replica")
-						msg := fmt.Sprintf("Replica %v of volume %v has been automatically salvaged", r.Name, v.Name)
-						vc.eventRecorder.Event(v, v1.EventTypeWarning, EventReasonAutoSalvaged, msg)
-						salvaged = true
+						log.WithField("replica", r.Name).Warn("Automatically salvaging volume replica")
+						// Revision counter enabled case, engine will
+						// check revision counter of replicas
+						if v.Spec.RevisionCounterDisabled {
```
shuo-wu, line 926: Why do we need this check? Then `salvaged` cannot be `true` if the revision counter is enabled. That doesn't make sense.

controller/volume_controller.go
```diff
@@ -1028,6 +1043,10 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
 			vc.eventRecorder.Eventf(v, v1.EventTypeNormal, EventReasonDetached, "volume %v has been detached", v.Name)
 		}
 		// Automatic reattach the volume if PendingNodeID was set
+		// TODO: need to revisit for CurrentNodeID and PendingNodeID update
+		// The reason is usually when CurrentNodeID is set, the volume
+		// should be in attaching or attached state, currently
+		// CurrentNodeID is set when volume is in detached state.
```
shuo-wu, line 1049: Add one more statement:
And we will implicitly assume `v.status.CurrentNodeID` has been the final state of this sync loop if this code block is reached. Hence updating `v.status.CurrentNodeID` here is weird.

shuo-wu (changes requested):

controller/volume_controller.go
boknowswiki: Sorry, missed `v.Spec.RevisionCounterDisabled` for `e.Spec.SalvageRequested`. Will update it. Thanks.

controller/volume_controller.go
boknowswiki: Sorry, should be `if !v.Spec.RevisionCounterDisabled {`. Will update. Thanks.

controller/volume_controller.go
boknowswiki: Will do. Thanks.

controller/volume_controller.go
shuo-wu: Thinking about a corner case here, if there is no usable failed replica, `v.Status.PendingNodeID` will be still set as long as the revision counter is disabled. But actually, the re-attaching will get stuck there when there is no available replica. It's a little bit weird...

shuo-wu (approved):

PR #687, opened 2020-09-22
Prometheus support, part 3

PhanLe1010:
Added the following metrics:
1. `longhorn_node_count_total` : We use this one to detect and alert when a node is down. e.g when the number of ready nodes < total number of nodes. This metric is necessary because when a node goes down, Longhorn manager on the node stops reporting so we cannot get the NotReady status directly from that Longhorn manager.
2. `longhorn_volume_state`: To display and alert about state of volume
3. `longhorn_volume_robustness`:  To display and alert about robustness of volume. e.g: send an alert when volume becomes fault.

With the 3 newly added metrics we are able to draw this dashboard: https://grafana.com/grafana/dashboards/13032.

longhorn/longhorn#1180

### DIFF
--- a/monitoring/volume_collector.go
+++ b/monitoring/volume_collector.go
@@ -6,13 +6,18 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 
 	"github.com/longhorn/longhorn-manager/datastore"
+	"github.com/longhorn/longhorn-manager/types"
+
+	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta1"
 )
 
 type VolumeCollector struct {
 	*baseCollector
 
-	capacityMetric metricInfo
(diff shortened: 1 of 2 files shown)


monitoring/volume_collector.go
```diff
@@ -1,6 +1,8 @@
 package monitoring
 
 import (
+	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta1"
```
yasker, line 4: Wrong location for import.

yasker (changes requested):

monitoring/volume_collector.go
PhanLe1010: Sorry, I forgot the imports statements again. Fixed it

yasker (approved):

PR #688, opened 2020-09-23
controller: Fix the unexpected auto detachment during the restoring

shuo-wu:
longhorn/longhorn#1807


controller/engine_controller.go
```diff
@@ -942,10 +943,11 @@ func syncWithRestoreStatus(log logrus.FieldLogger, engine *longhorn.Engine, rsMa
 				}
 			}
 		}
```
yasker, line 945: Can we just say if any replica is still restoring, then `isConsensual = false`?

yasker (approved):

PR #689, opened 2020-09-24
Prometheus Support, Part 4

PhanLe1010:
Added the following metrics:
1. longhorn_node_cpu_capacity_millicpu
2. longhorn_node_cpu_usage_millicpu
3. longhorn_node_memory_capacity_bytes
4. longhorn_node_memory_usage_bytes

We use the above metrics to detect when there is cpu/memory high pressure on worker nodes. This will help when debugging cases such as engines die unexpectedly due to cpu pressure. 

Also, change the metric name `longhorn_volume_usage_bytes` to `longhorn_volume_actual_size_bytes` to avoid the confusion between the size of the filesystem (usage_bytes) vs used space by Longhorn block device (actual size)

longhorn/longhorn#1180

### DIFF
--- a/monitoring/node_collector.go
+++ b/monitoring/node_collector.go
@@ -2,31 +2,41 @@ package monitoring
 
 import (
 	"strings"
+	"sync"
 
 	"github.com/sirupsen/logrus"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	metricsclientset "k8s.io/metrics/pkg/client/clientset/versioned"
 
 	"github.com/prometheus/client_golang/prometheus"
 
 	"github.com/longhorn/longhorn-manager/datastore"
 	"github.com/longhorn/longhorn-manager/types"
-
-	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta1"
 )
 
 type NodeCollector struct {
 	*baseCollector
 
+	kubeMetricsClient *metricsclientset.Clientset
+
 	statusMetric             metricInfo
 	totalNumberOfNodesMetric metricInfo
+	cpuUsageMetric           metricInfo
+	cpuCapacityMetric        metricInfo
+	memoryUsageMetric        metricInfo
+	memoryCapacityMetric     metricInfo
 }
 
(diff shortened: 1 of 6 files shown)


monitoring/volume_collector.go
```diff
@@ -41,8 +41,8 @@ func NewVolumeCollector(
 
 	vc.sizeMetric = metricInfo{
 		Desc: prometheus.NewDesc(
-			prometheus.BuildFQName(longhornName, subsystemVolume, "usage_bytes"),
-			"Used storage space in bytes for this volume",
+			prometheus.BuildFQName(longhornName, subsystemVolume, "actual_size_bytes"),
```
yasker, line 44: Separate this change into another commit.

monitoring/node_collector.go
```diff
@@ -49,33 +59,90 @@ func NewNodeCollector(
...
-		nc.logger.WithError(err).Warn("error during scrape")
-		return
-	}
+	var wg sync.WaitGroup
+
+	wg.Add(1)
+	go nc.collectNodeStatus(ch, &wg)
+
+	wg.Add(1)
+	nc.collectTotalNumberOfNodes(ch, &wg)
```
yasker, line 121: hmm? Those are single-threaded so why you need waitgroup? Did you mean to call those as goroutines?

Also, you can avoid passing the waitgroup into the function, since waitgroup is just a flow control e.g.
``` 
wg.Add(1)
go func() {
    defer wg.Done()
    call_func()
}()
```

yasker (changes requested):
Also, change the commit title to something else. Otherwise, it's hard to tell from the commit title what does this commit do. (it was the same as previous commits)

monitoring/node_collector.go
PhanLe1010: My bad. Yes I meant to call those funcs as goroutines. Also fixed to avoid passing the waitgroup.

monitoring/volume_collector.go
PhanLe1010: Thank you. Done

PhanLe1010:
Changed the commit name

yasker (approved):

PR #690, opened 2020-09-25
Fix bug that the uninstaller may get stuck there

shuo-wu:
longhorn/longhorn#1820

### DIFF
--- a/controller/uninstall_controller.go
+++ b/controller/uninstall_controller.go
@@ -70,8 +70,8 @@ func NewUninstallController(
 	c := &UninstallController{
 		baseController: newBaseControllerWithQueue("longhorn-uninstall", logger,
 			workqueue.NewNamedRateLimitingQueue(workqueue.NewMaxOfRateLimiter(
-				workqueue.NewItemExponentialFailureRateLimiter(500*time.Millisecond, 5*time.Second),
-				&workqueue.BucketRateLimiter{Limiter: rate.NewLimiter(rate.Limit(10), 100)},
+				workqueue.NewItemExponentialFailureRateLimiter(100*time.Millisecond, 2*time.Second),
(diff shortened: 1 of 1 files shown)


controller/uninstall_controller.go
```diff
@@ -70,8 +70,8 @@ func NewUninstallController(
 	c := &UninstallController{
 		baseController: newBaseControllerWithQueue("longhorn-uninstall", logger,
 			workqueue.NewNamedRateLimitingQueue(workqueue.NewMaxOfRateLimiter(
-				workqueue.NewItemExponentialFailureRateLimiter(500*time.Millisecond, 5*time.Second),
-				&workqueue.BucketRateLimiter{Limiter: rate.NewLimiter(rate.Limit(10), 100)},
+				workqueue.NewItemExponentialFailureRateLimiter(100*time.Millisecond, 2*time.Second),
```
yasker, line 73: We can just use `EnhancedDefaultControllerRateLimiter` unless it doesn't fit for this scenario.

yasker (changes requested):

controller/uninstall_controller.go
shuo-wu: The parameter (max delay is 1000s) of `NewItemExponentialFailureRateLimiter` in `EnhancedDefaultControllerRateLimiter` doesn't fit the uninstaller. You can see the issue post for details.

yasker: Should we have a max delay of 1000s for other controllers? It was copied from `DefaultControllerRateLimiter`.

shuo-wu: In other controllers, the keys are typically different. Hence the `EnhancedDefaultControllerRateLimiter ` should be fine.
But in the uninstall controller, the key is always `uninstall`. Keeping enqueuing the same key will be considered as handling failure then trigger the exponential backoff. Then the uninstaller may get blocked.

yasker (approved):

PR #691, opened 2020-09-28
csi: added support for  `kublet_volume_*` metrics

PhanLe1010:
Modified Longhorn CSI plugin to support `NodeGetVolumeStats`  RPC according to [CSI spec](https://github.com/container-storage-interface/spec/blob/master/spec.md#nodegetvolumestats). 

This allows Kubelet to query Longhorn CSI plugin and report the status of filesystems of PVCs bounded to Longhorn volumes. Users can then scrape Kubelet metric endpoints to obtains PVCs' status. 

Kubelet exposes [the following metrics](https://github.com/kubernetes/kubernetes/blob/4b24dca228d61f4d13dcd57b46465b0df74571f6/pkg/kubelet/metrics/collectors/volume_stats.go#L27):

1. kubelet_volume_stats_capacity_bytes
1. kubelet_volume_stats_available_bytes
1. kubelet_volume_stats_used_bytes
1. kubelet_volume_stats_inodes
1. kubelet_volume_stats_inodes_free
1. kubelet_volume_stats_inodes_used

After testing, I notice that Kubelet does not query the Longhorn CSI plugin for the volumes with `volumeMode: Block`. Therefore, we don't have to check whether a volume is in block mode before calling `unix.Statfs`.

longhorn/longhorn#1821

### DIFF
--- a/csi/node_server.go
+++ b/csi/node_server.go
@@ -3,13 +3,20 @@ package csi
 import (
 	"context"
 	"fmt"
-	"github.com/container-storage-interface/spec/lib/go/csi"
-	"github.com/sirupsen/logrus"
-	"google.golang.org/grpc/codes"
-	"google.golang.org/grpc/status"
-	"k8s.io/kubernetes/pkg/util/mount"
 	"os"
 	"path/filepath"
+	"strconv"
+
+	"github.com/container-storage-interface/spec/lib/go/csi"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+
+	"golang.org/x/sys/unix"
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
+	"k8s.io/kubernetes/pkg/util/mount"
 
 	longhornclient "github.com/longhorn/longhorn-manager/client"
 	"github.com/longhorn/longhorn-manager/types"
@@ -28,6 +35,7 @@ func NewNodeServer(apiClient *longhornclient.RancherClient, nodeID string) *Node
 		caps: getNodeServiceCapabilities(
 			[]csi.NodeServiceCapability_RPC_Type{
 				csi.NodeServiceCapability_RPC_EXPAND_VOLUME,
+				csi.NodeServiceCapability_RPC_GET_VOLUME_STATS,
 			}),
 	}
 }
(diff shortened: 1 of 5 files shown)


csi/node_server.go
```diff
@@ -197,8 +204,48 @@ func (ns *NodeServer) NodeUnstageVolume(
...
+func (ns *NodeServer) NodeGetVolumeStats(ctx context.Context, req *csi.NodeGetVolumeStatsRequest) (*csi.NodeGetVolumeStatsResponse, error) {
+	existVol, err := ns.apiClient.Volume.ById(req.GetVolumeId())
+	if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+	if existVol == nil {
+		msg := fmt.Sprintf("NodeGetVolumeStats: the volume %q not exists", req.GetVolumeId())
+		logrus.Warn(msg)
+		return nil, status.Error(codes.NotFound, msg)
+	}
```
PhanLe1010, line 216: CSI spec requires to return `codes.NotFound` if the volume does not exist. But, I am worried that this part might affect etcd performance.

joshimoo: NotFound is the correct return the CO (i.e. kubernetes) is then responsible for trying again with exponential backoff.
Any specific concern?

csi/util.go
```diff
@@ -152,3 +164,24 @@ func detectFileSystem(devicePath string) (string, error) {
 	}
 	return strings.Trim(strings.TrimPrefix(strings.TrimSpace(items[2]), "TYPE="), "\""), nil
 }
+
+func getStatistics(volumePath string) (volumeStatistics, error) {
```
joshimoo, line 168: return a pointer to volumeStatistics

csi/util.go
```diff
@@ -152,3 +164,24 @@ func detectFileSystem(devicePath string) (string, error) {
...
+	var statfs unix.Statfs_t
+	// See http://man7.org/linux/man-pages/man2/statfs.2.html for details.
+	err := unix.Statfs(volumePath, &statfs)
+	if err != nil {
+		return volumeStatistics{}, err
```
joshimoo, line 173: return nil

csi/node_server.go
```diff
@@ -197,8 +204,48 @@ func (ns *NodeServer) NodeUnstageVolume(
...
+
+	volumePath := req.GetVolumePath()
+	if volumePath == "" {
+		return nil, status.Error(codes.InvalidArgument, "NodeGetVolumeStats volume Volume Path cannot be empty")
+	}
+
```
joshimoo, line 222: in block mode you can just return the longhorn volume capacity.

```go
	return &csi.NodeGetVolumeStatsResponse{
		Usage: []*csi.VolumeUsage{
			&csi.VolumeUsage{
				Total:     // Longhorn volume capacity
				Unit:      csi.VolumeUsage_BYTES,
			},
		},
	}, nil
```

joshimoo (commented):
LGTM in general

csi/node_server.go
PhanLe1010: regard to the comment https://github.com/longhorn/longhorn-manager/pull/691#discussion_r496068884
> NotFound is the correct return the CO (i.e. kubernetes) is then responsible for trying again with exponential backoff.
Any specific concern?

My concern is that every time users scrape Kubelet metrics, we call apiClient. This might affect etcd performance when use scrapes metrics too often.

csi/util.go
PhanLe1010: Thank you! Fixed

csi/util.go
PhanLe1010: Thank you! Fixed

csi/node_server.go
joshimoo: Good thinking ahead 👍 
In this case, we are not actually calling against the kubernetes api (ETCD).

The api client is a client against the longhorn api, and will call the below method on the longhorn side:
https://github.com/longhorn/longhorn-manager/blob/93f30d51e817e17f0331908ce81f52f2b8271062/api/volume.go#L65

Which will then retrieve the volume from the datastore via the volume-manager
https://github.com/longhorn/longhorn-manager/blob/93f30d51e817e17f0331908ce81f52f2b8271062/api/volume.go#L79

csi/node_server.go
PhanLe1010: The CSI spec does not mention that the volume is in block mode when `volumePath == ""`. It said:
```
  // It can be any valid path where volume was previously
  // staged or published.
  // It MUST be an absolute path in the root filesystem of
  // the process serving this request.
  // This is a REQUIRED field.
  string volume_path = 2;
```
I am thinking that even when the volume is in block mode, it still has a `valid path where volume was previously staged or published`. How do you think?

csi/node_server.go
PhanLe1010: Oh. I see. Thank you so much for the explanation!

csi/node_server.go
joshimoo: No idea why it linked to that specific line, I wanted to link to the empty space.

joshimoo: @yasker do we keep track of whether a volume is used in block mode?

PhanLe1010: Thank you, I see. 

I am thinking about how to detect when the volume is used as a raw block device in the consuming pod.

My first idea is that we can check to see if the filesystem is `tmpfs` and conclude that the volume is a block device bc it does not have a filesystem? There is a catch with this method: use might use the volume as raw block device in the pod's spec but they created a filesystem on it inside the application which we are unaware of. 

My second idea is we find the pod which use the volume and read the information form the pod's spec. This method seems to be ineffective bc we need to call Kubernetes APIs.

How do you think ?

joshimoo: You should be able to do a unix.stat against the path, and if `st_rdev` is set then it's a block device.
https://linux.die.net/man/2/stat
Can you give this a try?

Then just refactor the get stats method to return the volumestats response, either for a block device or a mounted filesystem.

PhanLe1010: Thank you Joshua! I will try it

PhanLe1010: By the way, when testing, I notice that Kubelet does not query the Longhorn CSI plugin for the volumes with `volumeMode: Block`.

csi/util.go
```diff
@@ -152,3 +164,24 @@ func detectFileSystem(devicePath string) (string, error) {
 	}
 	return strings.Trim(strings.TrimPrefix(strings.TrimSpace(items[2]), "TYPE="), "\""), nil
 }
+
+func getStatistics(volumePath string) (*volumeStatistics, error) {
+	var statfs *unix.Statfs_t
+	// See http://man7.org/linux/man-pages/man2/statfs.2.html for details.
+	err := unix.Statfs(volumePath, statfs)
```
PhanLe1010, line 171: I made a fundamental mistake here which costs me a few hours of debugging. I should have allocated a struct `var statfs unix.Statfs_t` instead of a pointer. Because I only allocated a pointer and the initial value of the pointer is `nil`,  the `unix.Statfs` failed to dereference the `nil` pointer.

joshimoo: Your assessment is correct, the unix.Statfs function expects a pointer to a previously allocated struct, which it will end up filling.

csi/node_server.go
PhanLe1010: @joshimoo Thank you for the suggestion! The `unix.stat` func works. I implemented it slightly differently.

joshimoo (approved):
LGTM

PR #692, opened 2020-09-28
controller: Clean up the Longhorn node when the kube node is gone

shuo-wu:
longhorn/longhorn#1232

### DIFF
--- a/controller/kubernetes_node_controller.go
+++ b/controller/kubernetes_node_controller.go
@@ -198,11 +198,17 @@ func (knc *KubernetesNodeController) syncKubernetesNode(key string) (err error) 
 
 	kubeNode, err := knc.ds.GetKubernetesNode(name)
 	if err != nil {
-		if datastore.ErrorIsNotFound(err) {
-			logrus.Errorf("Kubernetes node %v has been deleted", key)
-			return nil
+		if !datastore.ErrorIsNotFound(err) {
+			return err
 		}
-		return err
+		logrus.Errorf("Kubernetes node %v has been deleted", key)
+	}
+
+	if kubeNode == nil {
+		if err := knc.ds.DeleteNode(name); err != nil {
+			return err
+		}
+		return nil
 	}
 
 	defer func() {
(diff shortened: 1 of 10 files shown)


shuo-wu:
This PR is pretty straightforward. Not sure if I miss anything, and if we still need a LEP.

controller/kubernetes_node_controller.go
```diff
@@ -198,11 +198,17 @@ func (knc *KubernetesNodeController) syncKubernetesNode(key string) (err error)
...
-			logrus.Errorf("Kubernetes node %v has been deleted", key)
-			return nil
+		if !datastore.ErrorIsNotFound(err) {
+			return err
 		}
-		return err
+		logrus.Errorf("Kubernetes node %v has been deleted", key)
+	}
+
+	if kubeNode == nil || kubeNode.DeletionTimestamp != nil {
```
yasker, line 207: Should we count `DeletionTimestamp != nil`? What if the manager is still running on the node?

Also, do we need to do the replica cleanup after the node deletion?

shuo-wu: OK. Maybe I cannot count `DeletionTimestamp != nil`. And there are some issues/corner cases for the replicas on the node. I will need to re-check those cases.

shuo-wu:
The following cases should be tested:
1. When the node is gone,
    1. -[ ] The volume attached to the node should become and keep the state `Faulted`. And there is no need to detach the volume, which is similar to the node-down case.
        1. -[ ] The only replica is on the removed node, the volume won't be salvaged.
    2. -[ ] The volume is not attached to the removed node. The replica on the removed node should become ERROR.
2. After the node is removed,
    1. -[ ] The replicas on the removed node cannot be salvaged.
    2. -[ ] All volumes can be detached and reattached manually.
3. -[ ] If a node with the same node is added to the cluster, all volumes can be recovered automatically unless the only replica becomes ERROR

yasker (approved):

yasker:
LGTM. Let's wait for the test result.

PR #694, opened 2020-09-28
Add namespace in example pvc/pod

khushboo-rancher:
https://github.com/longhorn/longhorn/issues/1830

Signed-off-by: khushboo-rancher <khushboo@rancher.com>

### DIFF
--- a/examples/deployment.yaml
+++ b/examples/deployment.yaml
@@ -15,6 +15,7 @@ apiVersion: v1
 kind: PersistentVolumeClaim
 metadata:
   name: mysql-pvc
+  namespace: default
 spec:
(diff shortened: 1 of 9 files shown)


examples/deployment.yaml
```diff
@@ -51,9 +53,6 @@ spec:
               - /var/lib/mysql/lost+found
           initialDelaySeconds: 5
           periodSeconds: 5
-        env:
```
khushboo-rancher, line 54: This env with the name 'MYSQL_ROOT_PASSWORD' was added twice, so deleted one.

joshimoo (approved):
LGTM

PR #695, opened 2020-09-29
deploy: Add `additionalPrinterColumn` to CRDs

PhanLe1010:
This allows `kubectl` to display summary information about Longhorn CRs upon listing.

The good news is this does not impact upgrade. Kubernetes allows editing CRDs' spec on the flight.

We can discuss what fields should be exposed in this PR

longhorn/longhorn#1480

### DIFF
--- a/deploy/install/01-prerequisite/03-crd.yaml
+++ b/deploy/install/01-prerequisite/03-crd.yaml
@@ -15,6 +15,26 @@ spec:
     singular: engine
   scope: Namespaced
   version: v1beta1
+  additionalPrinterColumns:
+  - name: State
+    type: string
+    description: The current state of the engine
+    JSONPath: .status.currentState
+  - name: Node
+    type: string
+    description: The node that the engine is on
+    JSONPath: .spec.nodeID
+  - name: InstanceManager
+    type: string
+    description: The instance manager of the engine
+    JSONPath: .status.instanceManagerName
+  - name: Image
+    type: string
+    description: The current image of the engine
+    JSONPath: .status.currentImage
+  - name: Age
+    type: date
+    JSONPath: .metadata.creationTimestamp
   subresources:
     status: {}
 ---
@@ -35,6 +55,30 @@ spec:
     singular: replica
   scope: Namespaced
   version: v1beta1
+  additionalPrinterColumns:
+  - name: State
+    type: string
+    description: The current state of the replica
+    JSONPath: .status.currentState
+  - name: Node
+    type: string
+    description: The node that the replica is on
+    JSONPath: .spec.nodeID
+  - name: Disk
+    type: string
+    description: The disk that the replica is on
+    JSONPath: .spec.diskID
+  - name: InstanceManager
+    type: string
+    description: The instance manager of the replica
+    JSONPath: .status.instanceManagerName
+  - name: Image
+    type: string
+    description: The current image of the replica
+    JSONPath: .status.currentImage
+  - name: Age
+    type: date
+    JSONPath: .metadata.creationTimestamp
   subresources:
     status: {}
 ---
@@ -55,6 +99,14 @@ spec:
     singular: setting
   scope: Namespaced
   version: v1beta1
+  additionalPrinterColumns:
+  - name: Value
+    type: string
+    description: The value of the setting
+    JSONPath: .value
+  - name: Age
+    type: date
+    JSONPath: .metadata.creationTimestamp
 ---
 apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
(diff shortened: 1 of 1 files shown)


yasker:
@PhanLe1010 do you have an example of output for each CRD type?

PhanLe1010:
@yasker One moment, please. I just cleaned up Longhorn. Bringing up a new installation for testing.

PhanLe1010:
Screen Shot 2020-09-29 at 4 46 15 PM

deploy/install/01-prerequisite/03-crd.yaml
```diff
@@ -15,6 +15,26 @@ spec:
...
   version: v1beta1
+  additionalPrinterColumns:
+  - name: State
+    type: string
+    description: The current state of the engine
+    JSONPath: .status.currentState
+  - name: Node
+    type: string
+    description: The node that the engine is on
+    JSONPath: .status.ownerID
```
yasker, line 26: This should be `spec.NodeID`.

deploy/install/01-prerequisite/03-crd.yaml
```diff
@@ -35,6 +55,26 @@ spec:
...
   version: v1beta1
+  additionalPrinterColumns:
+  - name: State
+    type: string
+    description: The current state of the replica
+    JSONPath: .status.currentState
+  - name: Node
+    type: string
+    description: The node that the replica is on
+    JSONPath: .status.ownerID
```
yasker, line 66: Same here, should be `spec.NodeID`.

deploy/install/01-prerequisite/03-crd.yaml
```diff
@@ -35,6 +55,26 @@ spec:
...
+  - name: InstanceManager
```
yasker, line 67: Add a field `DiskID`.

deploy/install/01-prerequisite/03-crd.yaml
```diff
@@ -55,6 +95,18 @@ spec:
     singular: setting
   scope: Namespaced
   version: v1beta1
+  additionalPrinterColumns:
+  - name: Value
+    type: string
+    description: The value of the setting
+    JSONPath: .value
+  - name: Generation
```
yasker, line 103: Don't need `Generation`

deploy/install/01-prerequisite/03-crd.yaml
```diff
@@ -73,6 +125,22 @@ spec:
     singular: volume
   scope: Namespaced
   version: v1beta1
+  additionalPrinterColumns:
+  - name: Size
```
yasker, line 129: Add `Robustness`, and condition `Scheduled`.

deploy/install/01-prerequisite/03-crd.yaml
```diff
@@ -113,6 +201,22 @@ spec:
     singular: node
   scope: Namespaced
   version: v1beta1
+  additionalPrinterColumns:
+  - name: Ready
+    type: string
+    description: Indicate whether the node is ready
+    JSONPath: .status.conditions['Ready']['status']
+  - name: AllowReplicaScheduling
```
yasker, line 209: `AllowScheduling` is enough.

deploy/install/01-prerequisite/03-crd.yaml
```diff
@@ -133,5 +237,21 @@ spec:
...
   version: v1beta1
+  additionalPrinterColumns:
+  - name: Type
+    type: string
+    description: The type of the instance manager (engine or replica)
+    JSONPath: .spec.type
+  - name: Node
+    type: string
+    description: The node that the instance manager is running on
+    JSONPath: .status.ownerID
```
yasker, line 248: Spec.nodeID.

yasker (changes requested):

deploy/install/01-prerequisite/03-crd.yaml
PhanLe1010: Thank you! Fixed

deploy/install/01-prerequisite/03-crd.yaml
PhanLe1010: Thank you! Fixed

deploy/install/01-prerequisite/03-crd.yaml
PhanLe1010: Thank you! Added

deploy/install/01-prerequisite/03-crd.yaml
PhanLe1010: Thank you! Removed

deploy/install/01-prerequisite/03-crd.yaml
PhanLe1010: Thank you! Added

deploy/install/01-prerequisite/03-crd.yaml
PhanLe1010: Thank you! Fixed

deploy/install/01-prerequisite/03-crd.yaml
PhanLe1010: Thank you! Fixed

PhanLe1010:
Updated the example out at https://github.com/longhorn/longhorn-manager/pull/695#issuecomment-701051583

PhanLe1010:
Why do we choose  `.spec.nodeID` over `.spec.ownerID`? Is it because sometimes `.status.ownerID` may not be the node that the object is on?

yasker:
@PhanLe1010 Yes. OwnerID only indicates which controller is managing the object. It doesn't have to be the same node as the volume/engine is attached to.

deploy/install/01-prerequisite/03-crd.yaml
yasker: well, I think it's a misunderstanding.

I meant rename the field to `AllowScheduling`, not `AllowReplicaScheduling`.

We can keep the `scheduable` condition.

yasker: Other than this, LGTM.

PhanLe1010: Sorry. I misunderstand it. Fixed

yasker (approved):

PR #696, opened 2020-10-05
app, controller: allow recurring backup for detached volumes

PhanLe1010:
In the current Longhorn implementation, users cannot do recurring backup when volumes are detached. 
This PR gives the users an option to do recurring backup even when volumes are detached.

The design overview is documented at the PR https://github.com/longhorn/longhorn/pull/1851

longhorn/longhorn#1509

### DIFF
--- a/app/volume.go
+++ b/app/volume.go
@@ -1,7 +1,9 @@
 package app
 
 import (
+	"encoding/json"
 	"fmt"
+	"math/rand"
 	"os"
 	"sort"
 	"time"
@@ -30,7 +32,11 @@ const (
 
 	SnapshotPurgeStatusInterval = 5 * time.Second
 
-	WaitInterval = 5 * time.Second
+	WaitInterval        = 5 * time.Second
+	VolumeAttachTimeout = 300 // 5 minutes
+
+	jobTypeSnapshot = string("snapshot")
+	jobTypeBackup   = string("backup")
 )
 
 func SnapshotCmd() cli.Command {
@@ -97,28 +103,27 @@ func snapshot(c *cli.Context) error {
 	}
 
 	backup := c.Bool(FlagBackup)
-	job, err := NewJob(managerURL, volume, snapshotName, labelMap, retain)
+	job, err := NewJob(managerURL, volume, snapshotName, labelMap, retain, backup)
 	if err != nil {
 		return err
 	}
-	if backup {
-		return job.backupAndCleanup()
-	}
-	return job.snapshotAndCleanup()
+
+	return job.run()
 }
 
 type Job struct {
 	lhClient     lhclientset.Interface
 	namespace    string
-	volume       *longhornclient.Volume
+	volumeName   string
 	snapshotName string
 	retain       int
+	jobType      string
 	labels       map[string]string
 
 	api *longhornclient.RancherClient
 }
 
-func NewJob(managerURL, volumeName, snapshotName string, labels map[string]string, retain int) (*Job, error) {
+func NewJob(managerURL, volumeName, snapshotName string, labels map[string]string, retain int, backup bool) (*Job, error) {
 	namespace := os.Getenv(types.EnvPodNamespace)
 	if namespace == "" {
 		return nil, fmt.Errorf("Cannot detect pod namespace, environment variable %v is missing", types.EnvPodNamespace)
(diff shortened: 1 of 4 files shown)


client/generated_volume.go
```diff
@@ -120,6 +120,8 @@ type VolumeOperations interface {
 	ActionSnapshotPurge(*Volume) (*Volume, error)
 
 	ActionSnapshotRevert(*Volume, *SnapshotInput) (*Snapshot, error)
+
+	ActionUpdateDataLocality(*Volume, *UpdateDataLocalityInput) (*Volume, error)
```
yasker, line 124: I don't think we need this in this PR? If we still want the action, we can update it in another PR.

api/model.go
```diff
@@ -537,7 +537,8 @@ func volumeSchema(volume *client.Schema) {
 		},
 
 		"updateDataLocality": {
-			Input: "UpdateDataLocalityInput",
+			Input:  "UpdateDataLocalityInput",
```
yasker, line 540: Unrelated change.

app/volume.go
```diff
@@ -31,6 +33,10 @@ const (
 	SnapshotPurgeStatusInterval = 5 * time.Second
 
 	WaitInterval = 5 * time.Second
+	timeout      = 300 // 5 minutes
```
yasker, line 36: Give `timeout` a name, e.g. `VolumeAttachTimeout`.

app/volume.go
```diff
@@ -41,6 +47,10 @@ func SnapshotCmd() cli.Command {
 				Name:  FlagManagerURL,
 				Usage: "Longhorn manager API URL",
 			},
+			cli.StringFlag{
+				Name:  FlagCurrentNode,
```
yasker, line 51: Why do we need this?

app/volume.go
```diff
@@ -274,12 +346,35 @@ func (job *Job) backupAndCleanup() (err error) {
...
 
+	volume, err := volumeAPI.ById(volumeName)
+	if err != nil {
+		return errors.Wrapf(err, "could not get volume %v", volumeName)
+	}
+
+	// Check the size of VolumeHead, if it is empty, skip the recurring backup.
+	// We don't want to overwrite the old backups with new identical backups.
+	// This happens when the volume is detached for a long time and has no new data.
+	volumeHead, err := volumeAPI.ActionSnapshotGet(volume, &longhornclient.SnapshotInput{
```
yasker, line 363: Separate the volume head size check into another commit. It will affect all the recurring jobs, not only detached one.

app/volume.go
```diff
@@ -386,3 +481,21 @@ func (job *Job) GetVolume(name string) (*longhorn.Volume, error) {
 func (job *Job) UpdateVolumeStatus(v *longhorn.Volume) (*longhorn.Volume, error) {
 	return job.lhClient.LonghornV1beta1().Volumes(job.namespace).UpdateStatus(v)
 }
+
+// waitForVolumeToBeInState timeout in second
+func (job *Job) waitForVolumeToBeInState(state string, timeout int) (*longhornclient.Volume, error) {
```
yasker, line 486: just call it `waitForVolumeState`

app/volume.go
yasker: OK I see. But what if the specified node is down? Seems better if we randomly choose a ready node.

controller/volume_controller.go
```diff
@@ -1940,7 +1943,8 @@ func (vc *VolumeController) createCronJob(v *longhorn.Volume, job *types.Recurri
 								},
 							},
 							ServiceAccountName: vc.ServiceAccount,
-							RestartPolicy:      v1.RestartPolicyOnFailure,
+							//RestartPolicy:      v1.RestartPolicyOnFailure,
+							RestartPolicy: v1.RestartPolicyNever,
```
yasker, line 1947: ?

yasker (changes requested):

api/model.go
PhanLe1010: Thank you. Fixed

client/generated_volume.go
PhanLe1010: Thank you. Fixed

app/volume.go
PhanLe1010: Thank you. Fixed

app/volume.go
PhanLe1010: Thank you. Fixed

app/volume.go
PhanLe1010: Thank you. Fixed

controller/volume_controller.go
PhanLe1010: My bad. I forgot to remove debugging code

app/volume.go
```diff
@@ -386,3 +459,56 @@ func (job *Job) GetVolume(name string) (*longhorn.Volume, error) {
...
+				readyNodeList = append(readyNodeList, node.Name)
+			}
+		}
+	}
+
+	if len(readyNodeList) == 0 {
+		return "", fmt.Errorf("cannot find a ready node")
+	}
+
+	return readyNodeList[rand.Int()%len(readyNodeList)], nil
```
yasker, line 532: Use `intn` instead.

yasker (changes requested):

app/volume.go
PhanLe1010: Thank you. Fixed

yasker (approved):

yasker:
LGTM. Waiting for the test result.

PR #698, opened 2020-10-06
app: modify the recurring job so that it only happens when volume has new data since the last backup/snapshot

PhanLe1010:
We do recurring backup jobs when:
1. `volume-head` has new data
2. `volume-head` doesn't new data but:
   1. last snapshot != last backup's snapshot

We do recurring snapshot jobs when:
1. `volume-head` has new data
1.  recurring snapshot is triggered by recurring backup jobs

longhorn/longhorn#1858

### DIFF
--- a/app/volume.go
+++ b/app/volume.go
@@ -216,8 +216,10 @@ func (job *Job) run() (err error) {
 	}
 
 	if job.jobType == jobTypeBackup {
+		logrus.Infof("Running recurring backup for volume %v", volumeName)
 		return job.backupAndCleanup()
 	}
+	logrus.Infof("Running recurring snapshot for volume %v", volumeName)
 	return job.snapshotAndCleanup()
 }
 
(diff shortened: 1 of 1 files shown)


app/volume.go
```diff
@@ -354,21 +455,12 @@ func (job *Job) backupAndCleanup() (err error) {
...
-	}
-	volumeHeadSize, err := strconv.ParseInt(volumeHead.Size, 10, 64)
+	shouldDo, err := job.shouldDoRecurringBackup(volume)
 	if err != nil {
 		return err
 	}
-	if volumeHeadSize == 0 {
-		logrus.Infof("Skipping the recurring backup because volume %v has no new data", volumeName)
+	if !shouldDo {
+		logrus.Infof("Skipping taking backup because volume %v is either empty or doesn't have new data since the last backup", volumeName)
```
yasker, line 463: `volume is empty`? should be `volume head`?

yasker (changes requested):

app/volume.go
PhanLe1010: When volume head is empty we might still take new backup (e.g. the parent of volume-head is not backuped yet)

Over here, I am referring to the case when volume is just created. The volume is empty (there isn't `lastSnapshot` and the `volume-head` is empty). We cannot compare the  `lastSnapshot`(which is `nil`) with `lastBackup`'s snapshot in this case but we still skip the reccurring backup in this case.

app/volume.go
```diff
@@ -293,11 +304,101 @@ func (job *Job) snapshotAndCleanup() (err error) {
...
+			oldestEle = ele
+		}
+	}
+
+	return oldestEle
+}
+
+// getLastSnapshot return the last snapshot of the volume exclude the volume-head
+// return nil, nil if volume doesn't have any snapshot other than the volume-head
+func (job *Job) getLastSnapshot(volume *longhornclient.Volume) (*longhornclient.Snapshot, error) {
```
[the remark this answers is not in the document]
PhanLe1010: Thank you for the `snapshot chain` suggestion! I fixed it to use the parent of `volume-head`

Yeah! I miss-used the word oldest :)

yasker (approved):

PR #699, opened 2020-10-09
app: change GetK3SCmdlineScript to adapt to the command line separator r in k3s v1.19.2+k3s1

PhanLe1010:
The new script will detect both:
 k3s versions before v1.19 (the cmdline is separated by `\000`)
 and k3s versions after v1.19 (the cmdline is separated by `\040`)

longhorn/longhorn#1861

### DIFF
--- a/app/get_proc_arg.go
+++ b/app/get_proc_arg.go
@@ -50,8 +50,10 @@ const (
         if [ ! -f $proc/cmdline ]; then
           continue
         fi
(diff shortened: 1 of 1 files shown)


PhanLe1010:
I am still looking into `why k3s changed the separator` and `whether it will only affect v1.19 in the future`

yasker (approved):

PR #700, opened 2020-10-12
Fix datastore object creation returning pointers to the cached objects

joshimoo:
The issue here is that in the case where we do creation verification,
we return a reference to the `readOnly` object from the lister directly.


yasker (approved):
Nice catch.

PR #703, opened 2020-10-13
Add missed informers for the uninstaller

shuo-wu:
Sometimes the CSIDriver object or deployments are not cleaned up immediately after the deletion timestamps are set, there is no way to trigger the uninstaller. Then the uninstallation will get stuck.

The root cause of this case is different from the 1st case mentioned in https://github.com/longhorn/longhorn/issues/1820

### DIFF
--- a/app/uninstall.go
+++ b/app/uninstall.go
@@ -94,6 +94,7 @@ func uninstall(c *cli.Context) error {
 	persistentVolumeClaimInformer := kubeInformerFactory.Core().V1().PersistentVolumeClaims()
 	kubeNodeInformer := kubeInformerFactory.Core().V1().Nodes()
 	priorityClassInformer := kubeInformerFactory.Scheduling().V1().PriorityClasses()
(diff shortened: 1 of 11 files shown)


app/uninstall.go
```diff
@@ -94,6 +94,7 @@ func uninstall(c *cli.Context) error {
 	persistentVolumeClaimInformer := kubeInformerFactory.Core().V1().PersistentVolumeClaims()
 	kubeNodeInformer := kubeInformerFactory.Core().V1().Nodes()
 	priorityClassInformer := kubeInformerFactory.Scheduling().V1().PriorityClasses()
+	csiDriverInformer := kubeInformerFactory.Storage().V1beta1().CSIDrivers()
```
yasker, line 97: CSI driver informer? the commit message only mentioned the deployments.

Also, why the sync period doesn't help?

shuo-wu: In this case, all related CRs, deployements and daemonsets are gone hence there is no event for the uninstaller controller, the 30s-force-sync doesn't work.

yasker (approved):

PR #706, opened 2020-10-14
Refactor uninstall controller for structured log

c3y1huang:
Refactor setting controller to address structured logging implementation.

https://github.com/longhorn/longhorn/issues/1733
https://github.com/longhorn/longhorn-manager/pull/661

### DIFF
--- a/controller/uninstall_controller.go
+++ b/controller/uninstall_controller.go
@@ -171,11 +171,7 @@ func (c *UninstallController) Run() error {
 	startTime := time.Now()
 	c.logger.Info("Uninstalling...")
 	defer func() {
-		log := c.logger.WithFields(
-			logrus.Fields{
-				"runtime": time.Now().Sub(startTime),
-			},
-		)
+		log := c.logger.WithField("runtime", time.Now().Sub(startTime))
 		log.Info("Uninstallation completed")
 	}()
 	go wait.Until(c.worker, time.Second, c.stopCh)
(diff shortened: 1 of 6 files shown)


controller/uninstall_controller.go
```diff
@@ -470,17 +485,19 @@ func (c *UninstallController) deleteInstanceManagers(instanceManagers map[string
 		err = errors.Wrapf(err, "Failed to delete instance managers")
 	}()
 	for _, im := range instanceManagers {
-		logFields := logrus.Fields{
-			"type": "instanceManager",
-			"name": im.Name,
-		}
+		log := c.logger.WithFields(
```
joshimoo, line 488: Add a getLoggerForInstanceManager function to the instance-manager-controller, same idea as above for the engine.
The idea is to have a getLoggerX for each crd type.

controller/uninstall_controller.go
```diff
@@ -491,50 +508,62 @@ func (c *UninstallController) deleteInstanceManagers(instanceManagers map[string
...
 			}
-			logrus.WithFields(logFields).Infof("Removed finalizer")
+			log.Info("Removed finalizer")
 		}
 	}
 	return
 }
 
 func (c *UninstallController) deleteManager() (bool, error) {
+	log := c.logger.WithFields(
```
joshimoo, line 525: For the DaemonSet/ReplicaSet you can add the getLoggerForX functions to this controller for now.
the idea is that the fieldName is `daemonSet`, with the value of the daemonsSet name.

controller/uninstall_controller.go
```diff
@@ -551,54 +580,72 @@ func (c *UninstallController) deleteDriver() (bool, error) {
 	}
 	wait := false
 	for _, name := range deploymentsToClean {
+		log := c.logger.WithFields(
```
joshimoo, line 583: See above for deamonSet

joshimoo (changes requested):
Thanks for the pr, I added some suggestions.

c3y1huang:
> Thanks for the pr, I added some suggestions.

Thanks for the review, I've addressed all in https://github.com/longhorn/longhorn-manager/pull/706/commits/0cf8db6d12d8453bc601cdd320a27f1f4db6abaf.

joshimoo (approved):
LGTM, for the daemonset/deployment case you could have used references to the actual deployment/daemonset but since we currently only use this in the uninstall controller we can refactor that part later :)

joshimoo:
@c3y1huang Thanks for the PR and the addition of the new logger functions :)

PR #707, opened 2020-10-14
add PodDisruptionBudgets to protect engine instance manager pod from being deleted when node is drained

PhanLe1010:
The idea is that InstanceManager controller will create a PDB for the engine instance manager pod on the current node.

When the current node is drained, the PDB blocks Kubernetes from evicting the correspoding engine instance manager pod. Only when all volumes are detached , then InstanceManager deletes the PDB and allows Kubernetes to finish the drain action.

In v1.1.0, users can use the following step to drain/shutdown nodes during maintaining:
(Let's say that we want to do maintaining on node-x)
1. Evict all replicas from the node-x (by feature https://github.com/longhorn/longhorn/issues/298)
1. Drain the node: `kubectl drain`. It doesn't have to be force drain. (by this feature and https://github.com/longhorn/longhorn/issues/1286)
1. Do maintaining work
1. terminate the node and bring up a new one or uncordon the current node

Credit: This PR is heavily based on @tksm 's idea: https://github.com/longhorn/longhorn/issues/1631#issuecomment-687055380

longhorn/longhorn#1631


controller/instance_manager_controller.go
```diff
@@ -419,6 +425,92 @@ func (imc *InstanceManagerController) syncInstanceManager(key string) (err error
...
+	imPDB, err := imc.ds.GetPDBRO(imc.getPDBName(im))
+	if err != nil && !datastore.ErrorIsNotFound(err) {
+		return err
+	}
+
+	// Delete the PodDisruptionBudget for the instance manager if all of the following conditions are meet:
+	// 1. The current node is unschedulable. This is an signal that the node is being cordoned/drained.
+	// 2. There is no instance process inside the instance manager
+	// 3. There exist a PodDisruptionBudget for this instance manager
+	if unschedulable && len(im.Status.Instances) == 0 {
```
yasker, line 447: What if the user want to depend on the rebuilding on the existing replica feature to quickly reboot the node without migrating all the replicas to other nodes? Wait for replicas to finish eviction can take a long time.

The PDB always makes sense for the engine since it should be moved with the workload pod. But It really depends on the user if they want to apply it to the replicas. If they want, it's likely they will perform the replica eviction first.

yasker (changes requested):

controller/instance_manager_controller.go
PhanLe1010: @yasker Great point!

Replica eviction feature will be handled replica migration. 
Blindly waiting for replica migration will take long time. 

I will modify the code so that Longhorn only use PDB to protect engine IM pods

controller/instance_manager_controller.go
```diff
@@ -419,6 +425,92 @@ func (imc *InstanceManagerController) syncInstanceManager(key string) (err error
...
+
+	imPDB, err := imc.ds.GetPDBRO(imc.getPDBName(im))
+	if err != nil && !datastore.ErrorIsNotFound(err) {
+		return err
+	}
+
+	// Delete the PodDisruptionBudget for the instance manager if all of the following conditions are meet:
+	// 1. The current node is unschedulable. This is an signal that the node is being cordoned/drained.
+	// 2. There is no instance process inside the instance manager
+	// 3. There exist a PodDisruptionBudget for this instance manager
```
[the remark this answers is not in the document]
PhanLe1010: Thank you. Fixed

controller/instance_manager_controller.go
```diff
@@ -419,6 +425,92 @@ func (imc *InstanceManagerController) syncInstanceManager(key string) (err error
...
+		return false, err
+	}
+	return kubeNode.Spec.Unschedulable, nil
+}
+
+func (imc *InstanceManagerController) deleteInstanceManagerPDB(im *longhorn.InstanceManager) error {
+	err := imc.ds.DeletePDB(imc.getPDBName(im))
+	if err != nil {
+		if datastore.ErrorIsNotFound(err) {
+			imc.logger.Warn("PDB %s has been deleted", imc.getPDBName(im))
```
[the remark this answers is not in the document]
PhanLe1010: Thank you! Fixed

controller/instance_manager_controller.go
PhanLe1010: I fixed it

controller/instance_manager_controller.go
```diff
@@ -443,6 +550,34 @@ func (imc *InstanceManagerController) enqueueInstanceManagerPod(pod *v1.Pod) {
...
+			return
+		}
+		utilruntime.HandleError(fmt.Errorf("couldn't get instance manager: %v", err))
+		return
+	}
+
+	for _, im := range ims {
+		imc.enqueueInstanceManager(im)
+	}
+}
```
PhanLe1010, line 579: I noticed that the InstanceManager is slow to react when node changes status (schedulable/unschedulable).

Therefore, I added to the `kubeNodeInformer` so that InstanceManager react faster.

controller/instance_manager_controller.go
```diff
@@ -181,6 +186,17 @@ func NewInstanceManagerController(
...
 	})
 
+	kubeNodeInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		UpdateFunc: func(oldObj, cur interface{}) {
+			curNode := cur.(*v1.Node)
+			imc.enqueueKubernetesNode(curNode)
+		},
+		DeleteFunc: func(obj interface{}) {
+			node := obj.(*v1.Node)
+			imc.enqueueKubernetesNode(node)
```
PhanLe1010, line 196: I noticed that the InstanceManager is slow to react when node changes status (schedulable/unschedulable).

Therefore, I added to the kubeNodeInformer so that InstanceManager react faster.

yasker (approved):
LGTM. Let me know if I can merge it.

PhanLe1010:
@yasker 
Integration core tests passed.

PR #708, opened 2020-10-15
allow `kubectl drain` without the --force flag

PhanLe1010:
`kubectl drain` by default is blocked when there is orphan pod on draining the node.

By adding the field `controller: true` to the `ownerReferences` of instance manager pods, we allow the user to run `kubectl drain` without the --force flag

Now users can run something like:
```
kubectl drain phan-cluster-v25-worker-1 --delete-local-data=true --force=false --grace-period=-1 --ignore-daemonsets=true --timeout=120s
```
longhorn/longhorn#1286

### DIFF
--- a/upgrade/v102to110/upgrade.go
+++ b/upgrade/v102to110/upgrade.go
@@ -1,0 +1,80 @@
+package v102to110
+
+import (
+	"fmt"
+
+	"github.com/pkg/errors"
+
+	v1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	clientset "k8s.io/client-go/kubernetes"
+
+	lhclientset "github.com/longhorn/longhorn-manager/k8s/pkg/client/clientset/versioned"
+	"github.com/longhorn/longhorn-manager/types"
+)
+
+// This upgrade is needed because we add one more field `controller: true`
+// to the ownerReferences of instance manager pods so that `kubectl drain`
+// can work without --force flag.
+// Therefore, we need to updade the field for all existing instance manager pod
+// Link to the original issue: https://github.com/longhorn/longhorn/issues/1286
+
+const (
+	upgradeLogPrefix = "upgrade from v1.0.2 to v1.1.0: "
+)
+
+func UpgradeInstanceManagerPods(namespace string, lhClient *lhclientset.Clientset, kubeClient *clientset.Clientset) (err error) {
+	defer func() {
+		err = errors.Wrapf(err, upgradeLogPrefix+"upgrade instance manager pods failed")
+	}()
+
+	imPodsList, err := kubeClient.CoreV1().Pods(namespace).List(metav1.ListOptions{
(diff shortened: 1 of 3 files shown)


upgrade/v102to110/upgrade.go
```diff
@@ -0,0 +1,79 @@
...
+func upgradeInstanceMangerPodOwnerRef(pod *v1.Pod, kubeClient *clientset.Clientset, namespace string) (err error) {
+	metadata, err := meta.Accessor(pod)
+	if err != nil {
+		return err
+	}
+
+	podOnwerRefs := metadata.GetOwnerReferences()
+	isController := true
+	needToUpdate := false
+	for ind, ownerRef := range podOnwerRefs {
```
yasker, line 60: Need to check if it's owned by Longhorn, in case some other components are using the field as well.

upgrade/v102to110/upgrade.go
```diff
@@ -0,0 +1,79 @@
...
+	return nil
+}
+
+func upgradeInstanceMangerPodOwnerRef(pod *v1.Pod, kubeClient *clientset.Clientset, namespace string) (err error) {
+	metadata, err := meta.Accessor(pod)
+	if err != nil {
+		return err
+	}
+
+	podOnwerRefs := metadata.GetOwnerReferences()
```
yasker, line 57: typo, `podOwnerRefs`

yasker (changes requested):

yasker:
Also, we need to be mindful about `--timeout=120s`. Not sure if PodDistruptionBudget can be affected.

PhanLe1010:
> Also, we need to be mindful about --timeout=120s. Not sure if PodDistruptionBudget can be affected.

How are you thinking that this could affect PodDistruptionBudget of engines?

yasker:
> > Also, we need to be mindful about --timeout=120s. Not sure if PodDistruptionBudget can be affected.
> 
> How are you thinking that this could affect PodDistruptionBudget of engines?

I've checked the reference, it should be fine. I was worried about kubectl will take some actions (e.g. remove PodDistruptionBudget) after the timeout. But it looks like the drain will simply fail after the timeout.

PhanLe1010:
Thank you!

upgrade/v102to110/upgrade.go
PhanLe1010: Thank you! Fixed

upgrade/v102to110/upgrade.go
PhanLe1010: Thank you! Fixed

upgrade/v102to110/upgrade.go
```diff
@@ -0,0 +1,80 @@
...
+		podOwnerRefs[ind] = ownerRef
+	}
+
+	if !needToUpdate {
+		return nil
+	}
+
+	metadata.SetOwnerReferences(podOwnerRefs)
+
+	if pod, err = kubeClient.CoreV1().Pods(namespace).Update(pod); err != nil {
```
yasker, line 75: Need to do `_, err :=`. Otherwise, `pod` can be nil if the call failed, then cause a NPE.

yasker (changes requested):

upgrade/v102to110/upgrade.go
PhanLe1010: Great catch! Thank you. Fixed

yasker (approved):

PR #710, opened 2020-10-15
Add disable replica rebuild setting to pause cross cluster rebuild

boknowswiki:
If disabled replica rebuild, the eviction and data locality feature
won't work, but restore disaster recovery volume feature and currently
rebuilding replica would still work as expected.

https://github.com/longhorn/longhorn/issues/307

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -1280,6 +1280,16 @@ func hasLocalReplicaOnSameNodeAsEngine(e *longhorn.Engine, rs map[string]*longho
 // It will count all the potentially usable replicas, since some replicas maybe
 // blank or in rebuilding state
 func (vc *VolumeController) replenishReplicas(v *longhorn.Volume, e *longhorn.Engine, rs map[string]*longhorn.Replica, hardNodeAffinity string) error {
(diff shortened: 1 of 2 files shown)


controller/volume_controller.go
```diff
@@ -543,6 +543,15 @@ func (vc *VolumeController) ReconcileEngineReplicaState(v *longhorn.Volume, e *l
 			vc.eventRecorder.Eventf(v, v1.EventTypeNormal, EventReasonHealthy, "volume %v became healthy", v.Name)
 		}
 
+		// If disabled replica rebuild, skip eviction and data locality.
+		disableReplicaRebuild, err := vc.ds.GetSettingAsBool(types.SettingNameDisableReplicaRebuild)
```
yasker, line 547: Can we add the logic in `replenishReplicas` instead? We can check if the `rs` is empty to decide if it's the first time creation.

yasker (changes requested):

controller/volume_controller.go
boknowswiki: Will fix it. Thanks.

yasker (approved):

PR #715, opened 2020-10-17
Upgrade default csi images

boknowswiki:
https://github.com/longhorn/longhorn/issues/6

### DIFF
--- a/csi/deployment.go
+++ b/csi/deployment.go
@@ -19,11 +19,11 @@ import (
 )
 
 const (
-	DefaultCSIAttacherImage            = "longhornio/csi-attacher:v2.0.0"
-	DefaultCSIProvisionerImage         = "longhornio/csi-provisioner:v1.6.0"
-	DefaultCSIResizerImage             = "longhornio/csi-resizer:v0.3.0"
-	DefaultCSISnapshotterImage         = "longhornio/csi-snapshotter:v2.1.1"
-	DefaultCSINodeDriverRegistrarImage = "longhornio/csi-node-driver-registrar:v1.2.0"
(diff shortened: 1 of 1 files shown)


yasker (approved):

csi/deployment.go
```diff
@@ -19,11 +19,11 @@ import (
...
-	DefaultCSIAttacherImage            = "longhornio/csi-attacher:v2.0.0"
-	DefaultCSIProvisionerImage         = "longhornio/csi-provisioner:v1.6.0"
-	DefaultCSIResizerImage             = "longhornio/csi-resizer:v0.3.0"
-	DefaultCSISnapshotterImage         = "longhornio/csi-snapshotter:v2.1.1"
-	DefaultCSINodeDriverRegistrarImage = "longhornio/csi-node-driver-registrar:v1.2.0"
+	DefaultCSIAttacherImage            = "longhornio/csi-attacher:lh-10142020"
+	DefaultCSIProvisionerImage         = "longhornio/csi-provisioner:lh-10132020"
+	DefaultCSIResizerImage             = "longhornio/csi-resizer:lh-10152020"
+	DefaultCSISnapshotterImage         = "longhornio/csi-snapshotter:lh-10132020"
+	DefaultCSINodeDriverRegistrarImage = "longhornio/csi-node-driver-registrar:lh-10132020"
```
joshimoo, line 26: Should have used year month day instead.
That way it would be easy to check if there is a newer version and it's universal across different culture formats

yasker: I was thinking that but it's probably better to stick with upstream version. I've filed https://github.com/longhorn/longhorn/issues/1886 to discuss.

PR #717, opened 2020-10-19
feature: Add AllowVolumeCreationWithDegradedAvailability

yasker:
The setting allows user to create a volume with less than specified
number of replicas, so user can use the default Longhorn configuration
without issues in one or two nodes cluster.

It is recommended to disable the feature when using Longhorn in
production environment.

Notice this PR also adds the ability to restore volume/DR volume with degraded availability.

https://github.com/longhorn/longhorn/issues/1701

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -813,7 +813,7 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
 		rs[r.Name] = r
 	}
 
-	allScheduled := true
+	scheduled := true
 	for _, r := range rs {
 		// check whether the replica need to be scheduled
 		if r.Spec.NodeID != "" {
@@ -835,12 +835,12 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
 					types.VolumeConditionTypeScheduled, types.ConditionStatusFalse,
 					types.VolumeConditionReasonLocalReplicaSchedulingFailure, "")
 			}
-			allScheduled = false
+			scheduled = false
 		} else {
 			rs[r.Name] = scheduledReplica
 		}
 	}
-	if allScheduled {
+	if scheduled {
 		v.Status.Conditions = types.SetCondition(v.Status.Conditions,
 			types.VolumeConditionTypeScheduled, types.ConditionStatusTrue, "", "")
 	} else if v.Status.CurrentNodeID == "" {
@@ -860,6 +860,7 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
 				v.Status.Conditions = types.SetCondition(v.Status.Conditions,
 					types.VolumeConditionTypeScheduled, types.ConditionStatusTrue, "",
 					"Reset schedulable due to allow volume creation with degraded availability")
+				scheduled = true
 			}
 		}
 	}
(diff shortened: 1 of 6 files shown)


controller/volume_controller.go
```diff
@@ -1745,7 +1746,7 @@ func (vc *VolumeController) checkForAutoDetachment(v *longhorn.Volume, e *longho
 	}
 	if e.Spec.RequestedBackupRestore != "" && e.Spec.RequestedBackupRestore == e.Status.LastRestoredBackup &&
 		!v.Spec.Standby &&
-		((cliAPIVersion >= engineapi.CLIVersionFour && !isPurging && (v.Status.Robustness == types.VolumeRobustnessHealthy || (v.Status.Robustness == types.VolumeRobustnessDegraded && !allScheduled))) ||
+		((cliAPIVersion >= engineapi.CLIVersionFour && !isPurging && (v.Status.Robustness == types.VolumeRobustnessHealthy || (v.Status.Robustness == types.VolumeRobustnessDegraded && !scheduled))) ||
```
yasker, line 1749: @shuo-wu I am uncertain about what does this line meant, and why `allScheduled` is used here.

shuo-wu: The only case that the volume can be auto detached during `Degraded` state:
Longhorn had tried to rebuild the volume but it cannot be complete. In other words, it's caused by a scheduling failure.

yasker: OK, it's hard to understand when you say it in this way... Let me rephrase:

We can automatically detach the volume when:
1. Either the volume become healthy
2. Or the volume is degraded, but all the possible replicas are complete the restore.

Before, since the `allScheduled` indicates if there is any schedule failed schedule, so it may make sense to use it to check if we've completed the restore for all the possible replica.

However, what if there are three replicas:
1. Replica 1: restore completed
2. Replica 2: in the rebuilding process
3. Replica 3: failed scheduling.

`allScheduled` would be false, but do we still need to wait for the replica 2 to complete the rebuild?

shuo-wu: Yes. Longhorn will rely on the condition `e.Spec.RequestedBackupRestore == e.Status.LastRestoredBackup` to guarantee that all running replicas have finished the restore.

yasker: If `e.Spec.RequestedBackupRestore == e.Status.LastRestoredBackup` guarantees all the running replicas are fine, then I don't understand why we want to check`allScheduled && Degraded`.

I can understand if `degraded && allScheduled` happened, we need to wait for the additional replica rebuild to complete.

But `degraded && !allScheduled` doesn't mean all the replica rebuilds/restores are completed. As I said in the previous case, it can be the second replica just get scheduled and haven't been added to the engine yet, while the third replica results in the `allScheduled == false`.

shuo-wu:
```
If e.Spec.RequestedBackupRestore == e.Status.LastRestoredBackup guarantees all the running replicas are fine, then I don't understand why we want to checkallScheduled && Degraded.
```
There is a corner case that the all **running** replicas finish the restore and the new replica object hasn't been created. `allScheduled && Degraded` is used to handle this case.

```
As I said in the previous case, it can be the second replica just get scheduled and haven't been added to the engine yet, while the third replica results in the allScheduled == false.
```
I didn't realize this corner case. Thank you for pointing out. It seems that we need to directly check all replicas here for this case.

yasker:
Core integration test passed.

PR #722, opened 2020-10-21
Change volume auto remount logic

PhanLe1010:
**Overview:**
Instead of manually finding and remounting all mount points of the volume, we kill the pod so that Kubernetes handles those works. This PR also solved the issue that remount doesn't support subpath (e.g. when pod use subpath in PVC)

**Explanation:**
The idea is that `VolumeControler` will set `LastRemountRequestAt` when the volume needs to remount. The `KubernetesPodController` will compare `LastRemountRequestAt` with the pod's `podStartTime`. If pod's `startTime` < `vol.Status.RemountRequestAt`, `KubernetesPodController` kills the pod. We don't kill the pod immediately though (see the code for more detail). After `KubernetesPodController` kills the pod, there is no need to pass the information back to `VolumeController` because there is no need to reset the field `LastRemountRequestAt` which is just an event in the past.

I didn't provide a global setting for users to choose whether they want Longhorn to auto delete workload pods when a remount is required because:
1. The deletion is safe. Longhorn only deletes a pod if the pod has a controller. This makes sure that the pod will be restarted by its controller. This is similar to `kubectl drain`. 
1. Because we removed the manual auto-remount logic, the workload pod must be restarted to be able to use the volume. 

**Other considered solution:**
`findmnt` command can be used to detect whether a mount point is mounting to the root of a filesystem or a directory(subpath) inside the filesystem. However, manually doing remount is cumbersome and leaves duplicated mount points on the host.

longhorn/longhorn#1719


PhanLe1010:
Integration core tests passed

controller/kubernetes_pod_controller.go
```diff
@@ -227,6 +233,55 @@ func (kc *KubernetesPodController) handlePodDeletionIfNodeDown(pod *v1.Pod, node
...
+		if err != nil {
+			return err
+		}
+
+		timeNow := time.Now()
+		delayDuration := time.Duration(int64(5)) * time.Second
+
+		if podStartTime.Before(remountRequestAt) && timeNow.After(remountRequestAt.Add(delayDuration)) {
+			gracePeriod := int64(0)
+			err := kc.kubeClient.CoreV1().Pods(pod.Namespace).Delete(pod.Name, &metav1.DeleteOptions{
```
yasker, line 270: Is it a force deletion(grace period 0)? Should we allow the pod to shutdown properly to release the resources?

types/resource.go
```diff
@@ -104,23 +104,23 @@ type WorkloadStatus struct {
...
+	CurrentImage         string               `json:"currentImage"`
+	KubernetesStatus     KubernetesStatus     `json:"kubernetesStatus"`
+	Conditions           map[string]Condition `json:"conditions"`
+	LastBackup           string               `json:"lastBackup"`
+	LastBackupAt         string               `json:"lastBackupAt"`
+	PendingNodeID        string               `json:"pendingNodeID"`
+	FrontendDisabled     bool                 `json:"frontendDisabled"`
+	RestoreRequired      bool                 `json:"restoreRequired"`
+	RestoreInitiated     bool                 `json:"restoreInitiated"`
+	LastRemountRequestAt string               `json:"lastRemountRequestAt"`
```
yasker, line 120: `RemountRequestedAt` should be good enough. Notice it's `requested`.

controller/kubernetes_pod_controller.go
PhanLe1010: Thank you! 
Yeah, we should not force delete

types/resource.go
PhanLe1010: Thank you! Fixed

yasker (approved):

PR #723, opened 2020-10-21
Add global setting for auto cleanup systen generated snapshot

boknowswiki:
https://github.com/longhorn/longhorn/issues/1669

### DIFF
--- a/types/setting.go
+++ b/types/setting.go
@@ -69,6 +69,7 @@ const (
 	SettingNameDisableReplicaRebuild                       = SettingName("disable-replica-rebuild")
 	SettingNameSystemManagedPodsImagePullPolicy            = SettingName("system-managed-pods-image-pull-policy")
 	SettingNameAllowVolumeCreationWithDegradedAvailability = SettingName("allow-volume-creation-with-degraded-availability")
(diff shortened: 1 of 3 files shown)


types/setting.go
```diff
@@ -503,6 +506,16 @@ var (
 		ReadOnly:    false,
 		Default:     "true",
 	}
+
+	SettingDefinitionAutoCleanupSystemGeneratedSnapshot = SettingDefinition{
+		DisplayName: "Auto Cleanup System Generated Snapshot",
```
yasker, line 511: `Automatically Cleanup ...`

yasker (changes requested):
Also need to add to `deploy/install/01-prerequisite/04-default-setting.yaml`

boknowswiki:
> Also need to add to `deploy/install/01-prerequisite/04-default-setting.yaml`

Oh, yes. Sorry about that. I will add it. Thanks for reminder.

types/setting.go
boknowswiki: Will fix it, thanks.

yasker (approved):

PR #724, opened 2020-10-21
Add a global setting for user to choose whether they want Longhorn todelete workload pod automatically when the volume is detached unexpectly

PhanLe1010:
longhorn/longhorn#1719


yasker (changes requested):
Need to update `default-settings.yaml` as well.

Also, remember to add the setting to the doc.

PhanLe1010:
My bad. 
I updated the `default-settings.yaml`
Updated document
Resolved conflix

yasker (approved):

PR #725, opened 2020-10-21
Fix missed deletion events crash the longhorn-manager

joshimoo:
If an informer misses a deletion event, the final state of the object
will be unknown and instead the event will be a marker object of type
cache.DeletedFinalStateUnknown with a sub Obj that contains the last
cached state.

Previously we didn't handle the cache.DeletedFinalStateUnkown object and instead always used a type assertion for the corresponding longhorn object.

longhorn/longhorn#1900
related: #1877


joshimoo:
Added some more team members for reference on how to correctly deal with the informer event handlers.
The change on the filter functions is a good case of where being more specific is nicer :)

yasker (approved):

PR #731, opened 2020-10-23
Prevent upgrade failure due to changed StorageClass

PhanLe1010:
We move the default Longhorn storageclass into manager's upgrading path so that it is not managed by Helm anymore. This helps to prevent the failure while upgrading Longhorn using Helm and the Longhorn default storageclass is accidentally modified by users.

Longhorn #1527

longhorn/longhorn#1527


PhanLe1010:
I am re-testing all the cases

types/types.go
```diff
@@ -68,6 +68,7 @@ const (
 
 	DeprecatedProvisionerName = "rancher.io/longhorn"
 	DepracatedDriverName      = "io.rancher.longhorn"
+	LonghornStorageClassName  = "longhorn"
```
yasker, line 71: `DefaultStorageClassName`.

upgrade/upgrade_storageclass.go
```diff
@@ -0,0 +1,100 @@
...
+
+	decode := scheme.Codecs.UniversalDeserializer().Decode
+	obj, _, err := decode(data, nil, nil)
+	if err != nil {
+		return nil, errors.Wrapf(err, "error while decoding YAML object")
+	}
+
+	sc, ok := obj.(*storagev1.StorageClass)
+	if !ok {
+		return nil, fmt.Errorf("%v is not a storageclass yaml", settingPath)
```
yasker, line 80: What will happen if the storage yaml is invalid and result in upgrade failure? Will the manager restart automatically?

yasker (changes requested):

upgrade/upgrade_storageclass.go
PhanLe1010: No. The upgrade will not fail. It will skip the new storageclass YAML file and keep the old one. I consider updating `longhorn` storageclass as a best-effort action. If it fail, the users will likely to notice it and take some manual interventions. 
In this case, restart manager will not solve the problem because the next time it wakes up it will see the same invalid storageclass yaml file.

PhanLe1010: I think that there is only 1 case that we need to stop the upgrade which is: we successfully deleted the existing `longhorn` storageclass but fail to create a new one. We need to stop the world in this case because if we don't workloads that are using `longhorn` storageclass will not be able to scale up.

types/types.go
PhanLe1010: Thank you. Fixed

upgrade/upgrade_storageclass.go
yasker: I don't follow. The current behavior is the manager going to restart, right? So that's upgrade failure.

What are we going to stop? All the managers would be done if the upgrade failed and user need to fix the config.

PhanLe1010: > I don't follow. The current behavior is the manager going to restart, right? So that's upgrade failure.

I am sorry for the confusion. No, the current behavior is the manager will not restart. It skips upgrading storageclass and continue. The return error is swallowed here: https://github.com/longhorn/longhorn-manager/blob/926e7f9d023375629236e96f60092a9a41a360ae/upgrade/upgrade_storageclass.go#L30

> What are we going to stop when you said stop the world? All the managers would be done if the upgrade failed and user needs to fix the config.

By saying `stop the world`, I mean that when `a manager pod successfully deleted the existing longhorn storageclass but failed to create a new one.`, that manager will be restarted. The next time it run, it will try to recreate the storageclass.

yasker: The problem of not stopping the manager is, the user will never notice the storage class error. We don't have anything to indicate the wrong configuration of storageclass except for the error in the log. And no user will look into the log unless something bad happens.

Another question is, what if the user changes the storageclass config map after the upgrade? Or does user suppose to change the storageclass config map after the upgrade? I am wondering should we keep the storageclass up to date to the configmap, not only for the upgrade process, but also during the normal operation.

PhanLe1010: @yasker 

> The problem of not stopping the manager is, the user will never notice the storage class error. We don't have anything to indicate the wrong configuration of storageclass except for the error in the log. And no user will look into the log unless something bad happens.

I absolutely agree with this! I never look at the log unless something went wrong.  :)))) I fixed the code so that it will stop the manager incase the storageclass yaml file is invalid.


> Another question is, what if the user changes the storageclass config map after the upgrade? Or does user suppose to change the storageclass config map after the upgrade? I am wondering should we keep the storageclass up to date to the configmap, not only for the upgrade process, but also during the normal operation.

No, the users are not supposed to change the storageclass config map after the upgrade. It will confuse Helm and create the same problem as what we are having right now with the storageclass.

yasker: Though any change to the configmap will be picked up by the manager upon restart, would that be a problem? ConfigMap should be changeable from helm's perspective? Since no field is prohibited from updating.

PhanLe1010: I agree!

PhanLe1010:
A new PR for this issue is at https://github.com/longhorn/longhorn-manager/pull/735. Close this PR

PR #732, opened 2020-10-23
Try to reuse failed replica during rebuilding

shuo-wu:
longhorn/longhorn#1304

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -442,6 +442,9 @@ func (vc *VolumeController) EvictReplicas(v *longhorn.Volume,
 func (vc *VolumeController) ReconcileEngineReplicaState(v *longhorn.Volume, e *longhorn.Engine, rs map[string]*longhorn.Replica) (err error) {
 	defer func() {
 		err = errors.Wrapf(err, "fail to reconcile engine/replica state for %v", v.Name)
+		if v.Status.Robustness != types.VolumeRobustnessDegraded {
+			v.Status.LastDegradedAt = ""
+		}
 	}()
 
 	if e == nil {
@@ -501,6 +504,7 @@ func (vc *VolumeController) ReconcileEngineReplicaState(v *longhorn.Volume, e *l
 				// from replica failed during rebuilding
 				if r.Spec.HealthyAt == "" {
 					r.Spec.HealthyAt = vc.nowHandler()
+					r.Spec.RebuildRetryCount = 0
 					rs[rName] = r
 				}
 				healthyCount++
@@ -535,6 +539,7 @@ func (vc *VolumeController) ReconcileEngineReplicaState(v *longhorn.Volume, e *l
 	} else { // healthyCount < v.Spec.NumberOfReplicas
 		v.Status.Robustness = types.VolumeRobustnessDegraded
 		if oldRobustness != types.VolumeRobustnessDegraded {
+			v.Status.LastDegradedAt = vc.nowHandler()
 			vc.eventRecorder.Eventf(v, v1.EventTypeNormal, EventReasonDegraded, "volume %v became degraded", v.Name)
 		}
 
@@ -1115,7 +1120,9 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
 				}
 				r.Spec.DesireState = types.InstanceStateStopped
 			} else if r.Spec.FailedAt == "" && r.Spec.EngineImage == v.Status.CurrentImage {
-				r.Spec.DesireState = types.InstanceStateRunning
+				if r.Status.CurrentState == types.InstanceStateStopped {
+					r.Spec.DesireState = types.InstanceStateRunning
+				}
 			}
 			rs[r.Name] = r
 		}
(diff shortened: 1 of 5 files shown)


controller/volume_controller.go
```diff
@@ -1315,29 +1322,61 @@ func (vc *VolumeController) replenishReplicas(v *longhorn.Volume, e *longhorn.En
...
 		replenishCount = 1
-		currentRebuilding := getRebuildingReplicaCount(e)
-		if currentRebuilding != 0 {
-			return nil
-		}
 	}
+
 	for i := 0; i < replenishCount; i++ {
-		r, err := vc.createReplica(v, e, rs, "")
+		hasPotentialReusableReplica, reusableFailedReplica, err := vc.scheduler.CheckAndReuseFailedReplica(rs, v, hardNodeAffinity)
```
yasker, line 1338: is `hasPotentialReusableReplica` same as `reusableFailedReplica != nil`? I don't see the difference in the flow below.

controller/volume_controller.go
```diff
@@ -1315,29 +1322,61 @@ func (vc *VolumeController) replenishReplicas(v *longhorn.Volume, e *longhorn.En
...
+			reusableFailedReplica.Spec.RebuildRetryCount++
+			rs[reusableFailedReplica.Name] = reusableFailedReplica
+		} else {
+			newReplicaRequired := false
+			// Creating new replica immediately when:
+			// 1. data locality is required
+			// 2. the volume is a new volume
+			// 3. replica eviction happens
+			// 4. there is no potential reusable replica
+			if hardNodeAffinity != "" || v.Status.Robustness != types.VolumeRobustnessDegraded || !hasPotentialReusableReplica {
```
yasker, line 1355: Same above. Do we need to check `hasPotentialReusableReplica` since we are already in `reusableFailedReplica == nil` clause?

controller/volume_controller.go
```diff
@@ -1315,29 +1322,61 @@ func (vc *VolumeController) replenishReplicas(v *longhorn.Volume, e *longhorn.En
...
+				newReplicaRequired = true
+			}
+			// Otherwise Longhorn will relay the new replica creation then there is a chance to reuse failed replicas later.
+			if !newReplicaRequired && v.Status.Robustness == types.VolumeRobustnessDegraded {
```
yasker, line 1359: Merge with `else` clause with the previous `if` statement.

controller/volume_controller.go
```diff
@@ -1315,29 +1322,61 @@ func (vc *VolumeController) replenishReplicas(v *longhorn.Volume, e *longhorn.En
...
+			}
+			if newReplicaRequired {
+				log.Debugf("A new replica will be created during rebuilding")
+				r, err := vc.createReplica(v, e, rs, hardNodeAffinity)
+				if err != nil {
+					return err
+				}
+				rs[r.Name] = r
+			} else {
+				log.Debugf("New replica replenishment is delayed")
```
yasker, line 1377: We can give a timestamp here `delayed until xxx`.

scheduler/replica_scheduler.go
```diff
@@ -286,6 +292,173 @@ func (rcs *ReplicaScheduler) scheduleReplicaToDisk(replica *longhorn.Replica, di
 		replica.Name, replica.Spec.NodeID, replica.Spec.DiskID, replica.Spec.DataPath)
 }
 
+// CheckAndReuseFailedReplica will check:
+//     1. If there is currently reusable failed replica
```
yasker, line 296: What's the difference between currently reusable and potentially reusable?

scheduler/replica_scheduler.go
```diff
@@ -286,6 +292,173 @@ func (rcs *ReplicaScheduler) scheduleReplicaToDisk(replica *longhorn.Replica, di
...
+		return false, true
+	}
+	if !rcs.checkTagsAreFulfilled(diskSpec.Tags, v.Spec.DiskSelector) {
+		return false, true
+	}
+
+	return true, true
+}
+
+func (rcs *ReplicaScheduler) chooseReusableReplica(nodeInfo map[string]*longhorn.Node, rs map[string]*longhorn.Replica, nodeReusableReplicaMap map[string]*longhorn.Replica, volume *longhorn.Volume) (*longhorn.Replica, error) {
```
yasker, line 369: Logic and flow here are similar to replica scheduling, except this is checking for an existing replica to see if it satisfied the current scheduling policy. We don't need to unify them now but it's better if we can unify the flow later so when we change the scheduling policy in the future, we won't need to modify two pieces of the code.

yasker (changes requested):

controller/volume_controller.go
yasker: According to Shuo
```
potential reusable replica means the failed replica may be able to reuse it later but it’s not valid now due to node/disk down issue.
```

Those are the replicas we need to wait for `ReplicaReplenishmentWaitInterval` for.

yasker: Make sure add those to the comments.

yasker (approved):

PR #734, opened 2020-10-26
controller: Keep syncing IP when IM is running

shuo-wu:
longhorn/longhorn#1880

If somehow the ETCD doesn't work correctly (may due to master reboot) and the instance manager pods are restarted with different IPs during the period, the InstanceManager Controller may miss the instance manager IP sync.


joshimoo (commented):
This changes the IP assignment from edge triggered to level based, which makes sense.
Are there other places where we are doing edge triggered updates?

yasker (approved):

yasker:
@joshimoo we might need a thorough code walkthrough to check those. Filed https://github.com/longhorn/longhorn/issues/1916 to track those.

PR #735, opened 2020-10-26
move Longhorn storageclass management into setting controller to prevent upgrade failure due to changed storageclass

PhanLe1010:
We move the default Longhorn storageclass into setting controller so that it is not managed by Helm anymore. This helps to prevent the failure while upgrading Longhorn using Helm and the Longhorn default storageclass  is accidentally modified by users.

Now, the storageclass ConfigMap will be the source of true. Any changes to this ConfigMap will result in creation of new longhorn storageclass.

longhorn/longhorn#1527


controller/setting_controller.go
```diff
@@ -403,6 +428,90 @@ func (sc *SettingController) updatePriorityClass() error {
...
+
+	storageclass, err = sc.ds.CreateStorageClass(storageclass)
+	if err != nil {
+		if apierrors.IsAlreadyExists(err) {
+			return nil
+		}
+		return err
+	}
+
+	sc.logger.Infof("Updated Longhorn storagecclass: %v", storageclass)
```
yasker, line 472: typo. `StorageClass`.

deploy/uninstall/uninstall.yaml
```diff
@@ -64,7 +64,7 @@ spec:
     spec:
       containers:
       - name: longhorn-uninstall
-        image: longhornio/longhorn-manager:master
+        image: phanle1010/longhorn-manager:sc-12
```
yasker, line 67: ?

controller/setting_controller.go
```diff
@@ -403,6 +428,90 @@ func (sc *SettingController) updatePriorityClass() error {
...
+				return nil
+			}
+			return nil
+		}
+		return err
+	}
+
+	storageclass, err := buildStorageClassManifestFromConfigMap(storageConfigMap)
+	if err != nil {
+		return err
```
yasker, line 447: So a failure to parse the config map will stop the setting controller from reconciliation?

Also, the extra setting name is weird. Now I think it's better to have `ConfigMapController` to take in the change instead.

yasker (changes requested):

controller/setting_controller.go
PhanLe1010: Yes, I agree that the extra setting is a bit weird.

controller/setting_controller.go
```diff
@@ -403,6 +428,90 @@ func (sc *SettingController) updatePriorityClass() error {
...
 }
 
+func (sc *SettingController) updateDefaultStorageClass() error {
+	storageConfigMap, err := sc.ds.GetConfigMapRO(types.DefaultStorageClassConfigMapName)
+	if err != nil {
+		if datastore.ErrorIsNotFound(err) {
+			// The storageclass configmap is deleted therefore we delete the actual storageclass
+			err := sc.ds.DeleteStorageClass(types.DefaultStorageClassName)
+			if err != nil && !datastore.ErrorIsNotFound(err) {
+				return nil
```
c3y1huang, line 438: Better to return the error instead?

PhanLe1010:
@c3y1huang 

Since we decided to to create a new controller to handle storageclass configmap. I have closed this PR and create a new one at https://github.com/longhorn/longhorn-manager/pull/738

Thank you for your review! Can we discuss about your review at https://github.com/longhorn/longhorn-manager/pull/738 instead?

PhanLe1010:
Close this PR because creating a replacement one at https://github.com/longhorn/longhorn-manager/pull/738

controller/setting_controller.go
PhanLe1010: Yeah. We should return err here. Thank you! Fixed at https://github.com/longhorn/longhorn-manager/pull/738

controller/setting_controller.go
```diff
@@ -403,6 +428,90 @@ func (sc *SettingController) updatePriorityClass() error {
...
+
+	sc.logger.Infof("Updated Longhorn storagecclass: %v", storageclass)
+
+	return nil
+}
+
+func buildStorageClassManifestFromConfigMap(cfm *v1.ConfigMap) (*storagev1.StorageClass, error) {
+	data, ok := cfm.Data["storageclass.yaml"]
+	if !ok {
+		return nil, fmt.Errorf("cannot find storageclass.yaml inside the default storageclass configmap")
```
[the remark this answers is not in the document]
PhanLe1010: It doesn't seems to have a func `fmt.Error`

datastore/kubernetes.go
```diff
@@ -176,6 +177,23 @@ func (s *DataStore) DeletePod(name string) error {
...
+func (s *DataStore) GetStorageClassRO(scName string) (*storagev1.StorageClass, error) {
+	return s.storageclassLister.Get(scName)
+}
+
+// DeleteStorageClass deletes StorageClass with the given name
+func (s *DataStore) DeleteStorageClass(scName string) error {
+	return s.kubeClient.StorageV1().StorageClasses().Delete(scName, nil)
+}
+
+// CreateStorageClass creates StorageClass with the given name
```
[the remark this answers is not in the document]
PhanLe1010: Thank you! Fixed

PR #736, opened 2020-10-26
Add native rwx support

joshimoo:
Draft PR for preliminary review. 
This is raw and I need to clean this up before final review.
There is also some stuff missing that still needs to be added.


yasker (changes requested):

shuo-wu (changes requested):
1. The lifecycle of share managers are unclear. When will a share manager get deleted/cleaned up? How do we handle the old share manager when the share manager image is updated?
2. I am confused about the relationship between Longhorn volumes and share managers. It seems that one share managers can handle multiple volumes, which is similar to one instance manager against the related instances. 
3. Can we add unit tests for share-manager-controller

shuo-wu (changes requested):
1. The lifecycle of share managers are unclear. When will a share manager get deleted/cleaned up? How do we handle the old share manager when the share manager image is updated?
2. I am confused about the relationship between Longhorn volumes and share managers. It seems that one share managers can handle multiple volumes, which is similar to one instance manager against the related instances. 
3. Can we add unit tests for share-manager-controller?

shuo-wu:
when there is no workload using a share volume, can this volume get detached? If share volumes cannot be detached, we may need to add some restrictions in volume-controller.

joshimoo:
This is for the multiple volumes per share manager approach which we are currently no longer using, therefore I am closing this for now. Since we will need to refactor this anyway if we decide to continue with this approach.

PR #737, opened 2020-10-26
Fix csi 'ControllerPublishVolume' checking attachment condition

boknowswiki:
Fix inconsistent attachment condition checking.
Split device frontend checking with attachment checking.

https://github.com/longhorn/longhorn/issues/1712

### DIFF
--- a/csi/controller_server.go
+++ b/csi/controller_server.go
@@ -268,8 +268,13 @@ func (cs *ControllerServer) ControllerPublishVolume(ctx context.Context, req *cs
 		return nil, status.Errorf(codes.Aborted, "The volume %s is %s", req.GetVolumeId(), existVol.State)
 	}
 
+	// Check volume frontend settings
+	if existVol.DisableFrontend || existVol.Frontend != string(types.VolumeFrontendBlockDev) {
+		return nil, status.Errorf(codes.InvalidArgument, "ControllerPublishVolume: there is no block device frontend for volume %s", req.GetVolumeId())
+	}
+
 	// the volume is already attached make sure it's to the same node as this
-	if existVol.State == string(types.VolumeStateAttached) {
+	if existVol.State == string(types.VolumeStateAttached) && existVol.Controllers[0].Endpoint != "" {
 		if !existVol.Ready || len(existVol.Controllers) == 0 {
 			return nil, status.Errorf(codes.Aborted,
 				"The volume %s is already attached but it is not ready for workloads", req.GetVolumeId())
@@ -632,6 +637,9 @@ func (cs *ControllerServer) waitForVolumeState(volumeID string, state types.Volu
 				return notFoundReturn
 			}
 			if existVol.State == string(state) {
+				if state == types.VolumeStateAttached && existVol.Controllers[0].Endpoint == "" {
+					continue
+				}
 				return true
 			}
 		}
--- a/csi/node_server.go
+++ b/csi/node_server.go
@@ -57,10 +57,17 @@ func (ns *NodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 	if len(existVol.Controllers) != 1 {
 		return nil, status.Errorf(codes.InvalidArgument, "There should be only one controller for volume %s", req.GetVolumeId())
 	}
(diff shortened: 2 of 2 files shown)


csi/controller_server.go
```diff
@@ -269,7 +269,7 @@ func (cs *ControllerServer) ControllerPublishVolume(ctx context.Context, req *cs
 	}
 
 	// the volume is already attached make sure it's to the same node as this
-	if existVol.State == string(types.VolumeStateAttached) {
+	if existVol.State == string(types.VolumeStateAttached) && existVol.Controllers[0].Endpoint == "" {
```
yasker, line 272: `Endpoint == ""`??

Also, why didn't we check if the frontend is disabled or not, or if frontend is blockdev or not? Just like we do in the node publish volume.

yasker (changes requested):

csi/controller_server.go
boknowswiki: Will fix it. Thanks.

yasker (approved):

csi/controller_server.go
```diff
@@ -268,8 +268,13 @@ func (cs *ControllerServer) ControllerPublishVolume(ctx context.Context, req *cs
...
 	}
 
+	// Check volume frontend settings
+	if existVol.DisableFrontend || existVol.Frontend != string(types.VolumeFrontendBlockDev) {
+		return nil, status.Errorf(codes.InvalidArgument, "ControllerPublishVolume: there is no block device frontend for volume %s", req.GetVolumeId())
+	}
+
 	// the volume is already attached make sure it's to the same node as this
-	if existVol.State == string(types.VolumeStateAttached) {
+	if existVol.State == string(types.VolumeStateAttached) && existVol.Controllers[0].Endpoint != "" {
```
joshimoo, line 277: in the case where endpoint == ""
we are going to call attach which isn't valid in the case where volume.State == attached.
see https://github.com/longhorn/longhorn-manager/blob/master/api/model.go#L910
this will then throw an internal error, below in the else clause.

csi/controller_server.go
```diff
@@ -268,8 +268,13 @@ func (cs *ControllerServer) ControllerPublishVolume(ctx context.Context, req *cs
...
 		if !existVol.Ready || len(existVol.Controllers) == 0 {
```
joshimoo, line 278: add || existVol.Controllers[0].Endpoint == "" to this volume ready for workload check.
Since that will return aborted, which is the code for operation in progress.

Might even look into only having the volume be ready once an endpoint is available, but I would have to look more at the code to make sure that this would be the correct behavior. cc @yasker

csi/controller_server.go
```diff
@@ -632,6 +637,9 @@ func (cs *ControllerServer) waitForVolumeState(volumeID string, state types.Volu
 				return notFoundReturn
 			}
 			if existVol.State == string(state) {
+				if state == types.VolumeStateAttached && existVol.Controllers[0].Endpoint == "" {
```
joshimoo, line 640: this is sort of a hack, but okay for now.
I have a refactor for this state checking function in my branch.

csi/node_server.go
```diff
@@ -57,10 +57,17 @@ func (ns *NodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
...
-		existVol.Frontend != string(types.VolumeFrontendBlockDev) || existVol.Controllers[0].Endpoint == "" {
+
+	// Check volume frontend settings
+	if existVol.DisableFrontend || existVol.Frontend != string(types.VolumeFrontendBlockDev) {
 		return nil, status.Errorf(codes.InvalidArgument, "There is no block device frontend for volume %s", req.GetVolumeId())
 	}
+
+	// Check volume attachment status
+	if existVol.State != string(types.VolumeStateAttached) || existVol.Controllers[0].Endpoint == "" {
+		return nil, status.Errorf(codes.InvalidArgument, "Volume %s hasn't been attached yet", req.GetVolumeId())
```
joshimoo, line 68: I would return code.FailedPrecondition here, since the precondition that the volume is attached isn't completed yet.

PR #738, opened 2020-10-27
Prevent upgrade failure due to changed storageclass

PhanLe1010:
We create a ConfigMap controller to manage the default Longhorn storageclass. Now, Longhorn storageclass is not managed by Helm anymore. This helps to prevent the failure while upgrading Longhorn using Helm and the Longhorn default storageclass is accidentally modified by users.

Now, the storageclass ConfigMap will be the source of true. Any changes to this ConfigMap will result in creation of new longhorn storageclass.

longhorn/longhorn#1527


controller/kubernetes_configmap_controller.go
```diff
@@ -0,0 +1,246 @@
...
+	return nil
+}
+
+func (kc *KubernetesConfigMapController) reconcileDefaultStorageClass(namespace, cfmName string) error {
+	// The default storageclass configmap must be in the same namespace as the current KubernetesConfigMapController
+	if namespace != kc.namespace {
+		return nil
+	}
+
+	storageCFM, err := kc.ds.GetConfigMap(kc.namespace, types.DefaultStorageClassConfigMapName)
```
yasker, line 155: If the configmap is not the one we want, we don't need to go through this.

There can be other config maps.

controller/kubernetes_configmap_controller.go
```diff
@@ -0,0 +1,246 @@
...
+
+	storageclass, err = kc.ds.CreateStorageClass(storageclass)
+	if err != nil {
+		if apierrors.IsAlreadyExists(err) {
+			return nil
+		}
+		return err
+	}
+
+	kc.logger.Infof("Updated the default Longhorn storageclass: %v", storageclass)
```
yasker, line 195: `StorageClass`. It's a noun.

controller/kubernetes_configmap_controller.go
```diff
@@ -0,0 +1,246 @@
...
+		}
+		return err
+	}
+
+	storageclass, err := buildStorageClassManifestFromConfigMap(storageCFM)
+	if err != nil {
+		return err
+	}
+
+	existingSC, err := kc.ds.GetStorageClassRO(types.DefaultStorageClassName)
```
yasker, line 173: Is it possible the existing storage class has default value filled in but the one in the configmap doesn't, so we will constantly update the storageclass object in the system?

yasker (changes requested):

controller/kubernetes_configmap_controller.go
```diff
@@ -0,0 +1,246 @@
...
+
+	kc.logger.Infof("Updated the default Longhorn storageclass: %v", storageclass)
+
+	return nil
+}
+
+func buildStorageClassManifestFromConfigMap(cfm *v1.ConfigMap) (*storagev1.StorageClass, error) {
+	data, ok := cfm.Data["storageclass.yaml"]
+	if !ok {
+		return nil, fmt.Errorf("cannot find storageclass.yaml inside the default storageclass configmap")
```
c3y1huang, line 203: How about using errors.New() instead?

yasker: It's more common to always use `fmt.Errorf` instead of switch between the two, since `errors.New()` doesn't provide a formatted output.

controller/kubernetes_configmap_controller.go
PhanLe1010: Thank you. Fixed

controller/kubernetes_configmap_controller.go
PhanLe1010: Thank you. Fixed

controller/kubernetes_configmap_controller.go
PhanLe1010: Yeah, it is possible. 
I am thinking of 2 ways:
1. Autofill the default values for the storage class built from the config map. (This one is not future-proof  because Kubernetes may introduce new fields in the future)
1. Update the config map to contains the newly updated storage class. This method seems like a circle.

PhanLe1010: I changed the logic so that we don't update the storage class by default unless there is a change in one of the nine field of the StorageClass struct. 

In the future, if Kubernetes introduces a new field to StorageClass, we can update the function `needToUpdateStorageClass` to check the new field.

yasker: Is there any Kubernetes function to fill in the default values for StorageClass?

PhanLe1010: @yasker 
I couldn't find a function that do that.

I found an interesting approach:  store a hash value of the ConfigMap's data field into an annotation of the storageclass.  If this hash value change -> we delete/create the storageclass.

yasker: We don't need to store the hash value. We can store the whole config we want to apply in the annotation as a JSON string, e.g.

`storageclass.longhorn.io/applied: {xxxx}`

PhanLe1010: Yeah, that will also work. But why not `hash`? :) Git use `hash` to comparing extensively

yasker: yeah, hash works too. The only downside is hash doesn't show what's the applied version, but it should be fine in our case.

controller/kubernetes_configmap_controller.go
```diff
@@ -0,0 +1,248 @@
...
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/tools/record"
+	"k8s.io/kubernetes/pkg/controller"
+
+	"github.com/longhorn/longhorn-manager/datastore"
+	"github.com/longhorn/longhorn-manager/types"
+)
+
+const (
+	lastAppliedStorageConfigLabelKeySuffix = "last-applied-configuration"
```
yasker, line 29: let's do `last-applied-configmap`

PhanLe1010: Thank you. Fixed

yasker (dismissed):
LGTM. Waiting for core test to pass.

controller/kubernetes_configmap_controller.go
```diff
@@ -0,0 +1,248 @@
...
+func (kc *KubernetesConfigMapController) reconcileDefaultStorageClass(namespace, cfmName string) error {
+	if namespace != kc.namespace || cfmName != types.DefaultStorageClassConfigMapName {
+		return nil
+	}
+
+	storageCFM, err := kc.ds.GetConfigMap(kc.namespace, types.DefaultStorageClassConfigMapName)
+	if err != nil {
+		if datastore.ErrorIsNotFound(err) {
+			// The StorageClass ConfigMap was deleted therefore we delete the actual StorageClass
+			err := kc.ds.DeleteStorageClass(types.DefaultStorageClassName)
```
joshimoo, line 160: I might be missing something, but I feel like this is a bad idea, since we won't be able to provision any volumes afterwards.

joshimoo (changes requested):
Can you check, my comment, I might be missing something.

controller/kubernetes_configmap_controller.go
PhanLe1010: Yes, you are right! Use will not be able to provision new volume with the `longhorn` StorageClass if we delete the `longhorn` storageclass.

This is the result when users manually delete the longhorn storage ConfigMap.  I feel that it is expected because the idea is that Longhorn will sync  longhorn storage ConfigMap with the actual `longhorn` StorageClass.

How do you think?

PhanLe1010:
@yasker 
The core tests passed except for the tests in https://ci.longhorn.io/job/public/job/longhorn-tests/142/#showFailuresLink

controller/kubernetes_configmap_controller.go
joshimoo: It's a good thought, though
I feel like in the case where a config map (conceptually these are user defined settings for me) is not present we should:
- check for the existence of a longhorn storage class
-- if a class exists, do nothing
-- if no class exists, create a default class with default settings (we already have them defined in the longhorn ui)

Absence of dynamic user configuration should not lead to total system failure.
What do you think?

joshimoo: For me the config map is conceptually an override
layer0: default settings
layer1: user settings (config-map)
absence of layer1 should not lead to failure

PhanLe1010: Yeah, good point!

However, I would prefer to completely separate the storageclass setting from UI default setting. Right now, some users are confused about those two since what they change in UI doesn't reflect on the actually `longhorn` storageclass. We have been telling them that changing UI setting has nothing to do with the volumes created via CSI using `longhorn` storageclass

Moreover, if users want to delete `longhorn` storageclass? If we fall back to default settings and recreate a new one, it means that users never be able to remove `longhorn` storageclass unless they remove the setting CRDs.

joshimoo: @yasker, @PhanLe1010 and me had a good design conversation and we explored only doing this during upgrade but in the end decided to do it at runtime and treat the config map as the source of truth for the default longhorn storage class.

But we agreed no to do anything if the config map is absent.

PhanLe1010: Thanks @joshimoo !

Update the PR to do nothing (not punish users :) ) if the config map is absent.

yasker (approved):

yasker:
since @joshimoo agreed with the approach, I will merge the PR.

PR #739, opened 2020-10-27
Fix recurring job scheduling problems

PhanLe1010:
This PR does:
1. Remove the hardcoded NodeName in pod's spec of recurring job
2. Set the tolerations for the cronjob pods that are set by the user via the longhorn-settings
3. Reduce SuccessfulJobsHistoryLimit to 1 to reduce total number of pods in the cluster

Previously, we used the engine client binary on the current node to do backup. Therefore, we need to add NodeName: v.Status.CurrentNodeID. Now that we changed to use Longhorn API to do backup, the NodeName field is no longer needed.

Issues:
longhorn/longhorn#1903
longhorn/longhorn#1796
longhorn/longhorn#1793

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -50,7 +50,8 @@ var (
 )
 
 const (
-	CronJobBackoffLimit = 3
+	CronJobBackoffLimit               = 3
+	CronJobSuccessfulJobsHistoryLimit = 1
 )
 
 type VolumeController struct {
@@ -1921,6 +1922,7 @@ func (vc *VolumeController) ResolveRefAndEnqueue(namespace string, ref *metav1.O
 
 func (vc *VolumeController) createCronJob(v *longhorn.Volume, job *types.RecurringJob, suspend bool) (*batchv1beta1.CronJob, error) {
 	backoffLimit := int32(CronJobBackoffLimit)
+	successfulJobsHistoryLimit := int32(CronJobSuccessfulJobsHistoryLimit)
 	cmd := []string{
 		"longhorn-manager", "-d",
 		"snapshot", v.Name,
@@ -1935,6 +1937,10 @@ func (vc *VolumeController) createCronJob(v *longhorn.Volume, job *types.Recurri
 	if job.Task == types.RecurringJobTypeBackup {
 		cmd = append(cmd, "--backup")
 	}
+	tolerations, err := vc.ds.GetSettingTaintToleration()
+	if err != nil {
+		return nil, err
+	}
 	// for mounting inside container
 	privilege := true
 	cronJob := &batchv1beta1.CronJob{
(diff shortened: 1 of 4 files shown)


joshimoo (approved):
Changes look good to me.
For reference can you leave a comment why we required the NodeName and now no longer need it?

PhanLe1010:
Thanks Joshua! I added the comment

PhanLe1010:
Manual tests looks good. 
Running integration core tests.

joshimoo:
@PhanLe1010 no need for the code comment in this case, it was more about adding a comment to the PR / commit message. Since the PR / commit can be found via the issue refs in the commit message, as well as via annotate (git blame).

PhanLe1010:
Oh. I see. 
Deleted it. Thank you

obeyler:
@joshimoo @PhanLe1010 If you need/desire and you have a docker image with this fix I can also test it on my cluster.

controller/volume_controller.go
```diff
@@ -50,7 +50,8 @@ var (
 )
 
 const (
-	CronJobBackoffLimit = 3
+	CronJobBackoffLimit               = 3
```
obeyler, line 53: I thinks it is already the default value for cron job set by K8S 
CronJobBackoffLimit               = 3
CronJobSuccessfulJobsHistoryLimit = 1
see https://v1-18.docs.kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#cronjob-v1beta1-batch

PhanLe1010: @obeyler Thank you for looking into this!

In the link you gave, it said that the default value for `CronJobBackoffLimit = 6` , and `CronJobSuccessfulJobsHistoryLimit = 3`.

We want to reduce those number in Longhorn recurring job since they don't really need to keep that many.

PhanLe1010:
@obeyler 

It would be awesome!  Here is the images `phanle1010/longhorn-manager:re-4`

Thank you!

controller/volume_controller.go
```diff
@@ -1902,8 +1903,21 @@ func (vc *VolumeController) ResolveRefAndEnqueue(namespace string, ref *metav1.O
 	vc.enqueueVolume(volume)
 }
 
+func (vc *VolumeController) getTaintToleration() ([]v1.Toleration, error) {
```
joshimoo, line 1906: Move this function into the datastore, like we did for the `GetSettingImagePullPolicy` function.
There is also a duplicate of this in the instance-manager-controller, engine-image-controller instead make them use the datastore function that you added.

joshimoo (approved):
LGTM, added a refactor suggestion.

controller/volume_controller.go
PhanLe1010: Great idea! Thank you! Done

joshimoo (approved):
LGTM

yasker:
@obeyler This is still a dev version so please don't use the image in your production environment.

PR #741, opened 2020-10-27
Add 'TooManySnapshots' condition to volume

boknowswiki:
https://github.com/longhorn/longhorn/issues/496

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -52,6 +52,7 @@ var (
 const (
 	CronJobBackoffLimit               = 3
 	CronJobSuccessfulJobsHistoryLimit = 1
+	VolumeSnapshotsWarningThreshold   = 100
 )
 
 type VolumeController struct {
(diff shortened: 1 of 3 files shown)


types/resource.go
```diff
@@ -196,6 +198,7 @@ type EngineStatus struct {
 	RebuildStatus            map[string]*RebuildStatus `json:"rebuildStatus"`
 	Snapshots                map[string]*Snapshot      `json:"snapshots"`
 	SnapshotsError           string                    `json:"snapshotsError"`
+	SnapshotsCount           int                       `json:"snapshotsCount"`
```
yasker, line 201: Do we need this field? Just `len(Snapshots)` will give you the same result.

controller/volume_controller.go
```diff
@@ -50,7 +50,8 @@ var (
 )
 
 const (
-	CronJobBackoffLimit = 3
+	CronJobBackoffLimit             = 3
+	VolumeSnapshotsWanringThreshold = 100
```
yasker, line 54: Typo.

controller/volume_controller.go
```diff
@@ -777,6 +778,17 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
 		v.Status.ActualSize = actualSize
 	}
 
+	if e.Status.SnapshotsCount > VolumeSnapshotsWanringThreshold {
+		log.Warnf("Volume %v has too many snapshots, which is over the threshold %v", v.Name, VolumeSnapshotsWanringThreshold)
```
yasker, line 782: Don't need the extra log. It's covered by `SetCondition` automatically.

controller/volume_controller.go
```diff
@@ -777,6 +778,17 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
...
+		v.Status.Conditions = types.SetCondition(v.Status.Conditions,
+			types.VolumeConditionTypeTooManySnapshots, types.ConditionStatusTrue,
+			types.VolumeConditionReasonTooManySnapshots, fmt.Sprintf("Snapshots count is over %v", VolumeSnapshotsWanringThreshold))
```
yasker, line 785: We should show the snapshot count as well.

yasker (changes requested):

yasker (approved):

PR #744, opened 2020-10-28
Fix the issue that causes the reboot test failure for failed replica reuse feature

shuo-wu:
https://github.com/longhorn/longhorn/issues/1304


yasker:
Build failed.

Is there a more reliable way to do this? E.g. can instance manager state help? Backoff is still needed but node down sounds like a wrong reason.

yasker (approved):

PR #746, opened 2020-10-28
Recurring job is stuck in running state and never finishes if the volume is detached unexpectedly

PhanLe1010:
Problem: In the old implementation, the recurring job is stuck in running state and never finishes if the volume is detached unexpectedly in the middle of a backup. This prevents the cronjob from creating any new job in the future. Effectively, it suspends the recurring cronjob forever.

Recurring cronjob now detects when a volume is detached unexpectedly, and cancels the current running job. This makes sure that it doesn't block other running in the future.

longhorn/longhorn#1882

### DIFF
--- a/app/volume.go
+++ b/app/volume.go
@@ -33,8 +33,9 @@ const (
 
 	SnapshotPurgeStatusInterval = 5 * time.Second
 
-	WaitInterval        = 5 * time.Second
-	VolumeAttachTimeout = 300 // 5 minutes
+	WaitInterval              = 5 * time.Second
+	VolumeAttachTimeout       = 300 // 5 minutes
+	BackupProcessStartTimeout = 90  // 1.5 minutes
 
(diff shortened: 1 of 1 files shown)


app/volume.go
```diff
@@ -445,6 +446,10 @@ func (job *Job) backupAndCleanup() (err error) {
 		return err
 	}
 
+	if hasStarted, err := job.waitForBackupProcessStart(BackupProcessStartTimeout); err != nil || !hasStarted {
```
yasker, line 449: ? Can it be `err != nil` but `hasStarted`?

yasker (changes requested):

app/volume.go
PhanLe1010: It cannot. Thank you. Fixed

yasker (approved):

PR #747, opened 2020-10-29
Refactor replica controller for structured log

c3y1huang:
Refactor replica controller to address structured logging implementation.

https://github.com/longhorn/longhorn/issues/1731
https://github.com/longhorn/longhorn-manager/pull/661

### DIFF
--- a/controller/instance_manager_controller.go
+++ b/controller/instance_manager_controller.go
@@ -248,7 +248,12 @@ func (imc *InstanceManagerController) handleErr(err error, key interface{}) {
 }
 
 func getLoggerForInstanceManager(logger logrus.FieldLogger, im *longhorn.InstanceManager) *logrus.Entry {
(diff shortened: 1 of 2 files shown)


yasker:
@c3y1huang Sorry we missed this one. Can you rebase to the latest code?

@joshimoo can you review when you have time?

c3y1huang:
rebased.

yasker:
cc @joshimoo

joshimoo (approved):
LGTM, I was concerned about the fact that we already have a node field, 
https://github.com/longhorn/longhorn-manager/blob/ce66fe014abe9a558e5151ca6fd0163b22f09499/app/daemon.go#L130

But thinking some more about this, having the nodeID exposed as what is currently recorded in the relevant resources spec.nodeID seems like a good idea.

PR #749, opened 2020-10-29
Add 'node.Status.OwnerID' field to bind the node object to a node controller

boknowswiki:
This new field will prevent multiple node controllers to update the same
node object.

https://github.com/longhorn/longhorn/issues/1449

### DIFF
--- a/controller/node_controller.go
+++ b/controller/node_controller.go
@@ -269,6 +269,10 @@ func getLoggerForNode(logger logrus.FieldLogger, n *longhorn.Node) logrus.FieldL
 	return logger.WithField("node", n.Name)
 }
 
+func (nc *NodeController) isResponsibleFor(n *longhorn.Node) bool {
+	return isControllerResponsibleFor(nc.controllerID, nc.ds, n.Name, n.Name, n.Status.OwnerID)
+}
+
 func (nc *NodeController) syncNode(key string) (err error) {
 	defer func() {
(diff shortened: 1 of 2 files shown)


yasker (approved):
I am a little bit worried about this change since the function `isControllerResponsibleFor` depends on the node status itself. Let's see how it goes in the manual node down tests.

boknowswiki:
> I am a little bit worried about this change since the function `isControllerResponsibleFor` depends on the node status itself. Let's see how it goes in the manual node down tests.

Could you explain a little bit more about your concern? Thanks.

yasker:
https://github.com/longhorn/longhorn-manager/blob/master/controller/controller_manager.go#L183 depends on the node status, while the call might result in updating the node status. It seems like a loop.

yasker:
OK, now I don't know how can this possibly work.

`ownerDown, err = ds.IsNodeDownOrDeleted(currentOwnerID)` won't return `ownerDown = true` since there is no other controller indicate the node is down. So the original owner will always prevail.

yasker:
Oh, you always set the `preferredOwnerID` to the caller controller. That means the node object will no longer be controlled by the controller on the same node even if it's not down.

yasker:
No, it doesn't work in my test. A down node (NotReady) cannot be detected by Longhorn after the change. Revert the PR now.

PR #752, opened 2020-10-30
Remove DisableFrontend check in ControllerPublishVolume

c3y1huang:
DisableFrontend gets set to false during volume attach hence no need to check before.

### DIFF
--- a/csi/controller_server.go
+++ b/csi/controller_server.go
@@ -269,7 +269,7 @@ func (cs *ControllerServer) ControllerPublishVolume(ctx context.Context, req *cs
 	}
 
 	// Check volume frontend settings
-	if existVol.DisableFrontend || existVol.Frontend != string(types.VolumeFrontendBlockDev) {
+	if existVol.Frontend != string(types.VolumeFrontendBlockDev) {
 		return nil, status.Errorf(codes.InvalidArgument, "ControllerPublishVolume: there is no block device frontend for volume %s", req.GetVolumeId())
 	}
 


csi/controller_server.go
```diff
@@ -303,6 +298,11 @@ func (cs *ControllerServer) ControllerPublishVolume(ctx context.Context, req *cs
 		}
 	}
 
+	// Check volume frontend settings
+	if existVol.DisableFrontend || existVol.Frontend != string(types.VolumeFrontendBlockDev) {
```
shuo-wu, line 302: There is no need to move condition check `existVol.Frontend != string(types.VolumeFrontendBlockDev)`.

yasker: In fact, I am fine with moving both since it's weird to check the frontend value if the frontend is disabled.

shuo-wu: 1. The field `vol.DisableFrontend` is decided by the `Attach` request. Hence there is no need to check `existVol.DisableFrontend`.
2. It's better to check `existVol.Frontend != string(types.VolumeFrontendBlockDev)` before the `Attach` call. (There is no need to move this condition check.) If it's false, this function can directly error out rather than continuing call `Attach`.

shuo-wu (changes requested):

csi/controller_server.go
c3y1huang: Addressed.

shuo-wu (approved):

yasker (approved):

PR #753, opened 2020-10-30
Fix failed replica reuse issues

shuo-wu:
longhorn/longhorn#1304

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -1359,18 +1359,17 @@ func (vc *VolumeController) replenishReplicas(v *longhorn.Volume, e *longhorn.En
 			return errors.Wrapf(err, "failed to reuse a failed replica during replica replenishment")
 		}
 		if reusableFailedReplica != nil {
-			if vc.backoff.IsInBackOffSinceUpdate(reusableFailedReplica.Name, time.Now()) {
-				log.Debugf("Cannot reuse failed replica %v immediately, backoff period is %v now",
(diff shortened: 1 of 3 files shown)


shuo-wu:
This PR hasn't been tested.

controller/volume_controller.go
```diff
@@ -1359,18 +1359,17 @@ func (vc *VolumeController) replenishReplicas(v *longhorn.Volume, e *longhorn.En
 			return errors.Wrapf(err, "failed to reuse a failed replica during replica replenishment")
 		}
 		if reusableFailedReplica != nil {
-			if vc.backoff.IsInBackOffSinceUpdate(reusableFailedReplica.Name, time.Now()) {
-				log.Debugf("Cannot reuse failed replica %v immediately, backoff period is %v now",
-					reusableFailedReplica.Name, vc.backoff.Get(reusableFailedReplica.Name).Seconds())
+			if !vc.backoff.IsInBackOffSinceUpdate(reusableFailedReplica.Name, time.Now()) {
```
yasker, line 1362: Why after found a reusableFailedReplica, we might still replenish a new replica? It's unclear to me.

shuo-wu: If the failed replica reusage is in the backoff window, it's considered as unavailable. Otherwise, the replica rebuild starting time may be longer than the setting replenishment wait interval.

shuo-wu: I am thinking if it would be more straightforward to put the backoff mechanism into the scheduler function.

yasker (approved):

PR #755, opened 2020-11-01
Add 'OwnerID field in 'NodeStatus' to bind node object to a node controller

boknowswiki:
Add 'OwnerID' to 'NodeStatus' and add logic to
'isControllerResponsibleFor' to get Kubernetes node condition for node
controller.

https://github.com/longhorn/longhorn/issues/1449

### DIFF
--- a/controller/controller_manager.go
+++ b/controller/controller_manager.go
@@ -10,6 +10,7 @@ import (
 	"golang.org/x/time/rate"
 
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/informers"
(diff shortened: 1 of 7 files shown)


controller/controller_manager.go
```diff
@@ -174,15 +175,39 @@ func GetGuaranteedResourceRequirement(ds *datastore.DataStore) (*corev1.Resource
...
+					ownerDown = true
+				} else {
+					// For other error don't trade as node down.
+					// Be consistent with ds.IsNodeDownOrDeleted.
+					logrus.Warnf("Kubernetes get node %v failed: %v, don't mark as node down", name, err)
+				}
+			} else {
+				kubeConditions := kubeNode.Status.Conditions
+				for _, con := range kubeConditions {
+					if con.Type == corev1.NodeReady {
```
yasker, line 199: Node not ready can be caused by various reasons, for example, manager pod is down (https://github.com/longhorn/longhorn-manager/blob/84228670e96accb94a1597a69a0d7443b956f5ea/controller/node_controller.go#L332 ) or node has various reasons or pressure (https://github.com/longhorn/longhorn-manager/blob/84228670e96accb94a1597a69a0d7443b956f5ea/controller/node_controller.go#L385 ). 

Also, we shouldn't need `getKubeNode bool`.

This change can have more implications on how to decide which condition we should set for the node. For example, should we move the condition check to another function? We definitely won't want to duplicate the code here.

yasker:
Close this for now. We've punt https://github.com/longhorn/longhorn/issues/1449 to v1.1.1.

PR #757, opened 2020-11-02
Return error in the backup recurring job when volume is detached unexpectedly

PhanLe1010:
In the recurring backup job, we check whether the engine process is running. If it is not running, the volume is detached in the middle of the backup process. Therefore, we return error.

longhorn/longhorn#1882

### DIFF
--- a/app/volume.go
+++ b/app/volume.go
@@ -34,6 +34,7 @@ const (
 	SnapshotPurgeStatusInterval = 5 * time.Second
 
 	WaitInterval              = 5 * time.Second
+	DetachingWaitInterval     = 30 * time.Second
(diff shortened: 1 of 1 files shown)


app/volume.go
```diff
@@ -528,11 +542,24 @@ func (job *Job) waitForBackupProcessStart(timeout int) error {
 			return err
 		}
 
+		hasRunningController := false // controller is the engine process
```
yasker, line 545: Do we need this for starting the backup? Unexpected detach should result in fail to start the backup soon.

PhanLe1010: You are right. Fixed. Thank you

yasker (approved):

PR #759, opened 2020-11-04
Disk and replica migration

shuo-wu:
longhorn/longhorn#1269

### DIFF
--- a/api/model.go
+++ b/api/model.go
@@ -110,6 +110,7 @@ type Replica struct {
 	Instance
 
 	DiskID   string `json:"diskID"`
+	DiskPath string `json:"diskPath"`
 	DataPath string `json:"dataPath"`
 	Mode     string `json:"mode"`
 	FailedAt string `json:"failedAt"`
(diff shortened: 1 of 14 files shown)


yasker (dismissed):

yasker:
LGTM in general.

The only thing that worries me is when we updating replicas, we don't check if the previous node has the disk disconnected (or the node is gone), or if the replica is stopped/errored out. But that would require us to have the ability to find the disk by UUID. Not sure if it worth the complication.

shuo-wu:
1. If the node is already gone, it seems that we need to skip the replica update rather than erroring out. Otherwise, the whole Longhorn upgrade will be blocked.
2. If a disk is disconnected, the replicas using the disk are failed/potentially failed(stopped) replicas. If there is no available DiskUUID in the disk status, `DiskID` of all related replicas becomes invalid. As a result, these (failed) replicas cannot be reused. Hence I think it's fine to ignore these replicas if Longhorn cannot handle them.

yasker (approved):

PR #761, opened 2020-11-05
Add logic to handle the case RemountRequestedAt sometimes is not set when all replicas fail

PhanLe1010:
I have posted the analyst at https://github.com/longhorn/longhorn/issues/1942#issuecomment-722698783

longhorn/longhorn#1942

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -978,20 +978,20 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
 			}
 		}
 	} else { // !allFaulted
-		// Don't need to touch other status since it should converge naturally
 		if v.Status.Robustness == types.VolumeRobustnessFaulted {
 			v.Status.Robustness = types.VolumeRobustnessUnknown
+			// The volume was faulty and there are usable replicas.
+			// Therefore, we set RemountRequestedAt so that KubernetesPodController restarts the workload pod
+			v.Status.RemountRequestedAt = vc.nowHandler()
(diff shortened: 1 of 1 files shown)


PhanLe1010:
Integration core tests passed.

controller/volume_controller.go
```diff
@@ -1000,6 +1000,7 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
 					rs[r.Name] = r
 				}
 			}
+			v.Status.Robustness = types.VolumeRobustnessFaulted
```
yasker, line 1003: Do we need to set the volume state here?

PhanLe1010: Yes, we need to set it so that Longhorn sets `v.Status.RemountRequestedAt` in the next reconciliation run at https://github.com/longhorn/longhorn-manager/blob/95b63e85741c189bdae2112cf2f2fcf7df8f8a50/controller/volume_controller.go#L981.

It also makes sense because when the engine crash, the volume should be faulty.

yasker (approved):

yasker:
@shuo-wu Can you take a look at this PR as well?

shuo-wu (dismissed):
LGTM except for the typo in the commit message

shuo-wu:
> Integration core tests passed.

It's better to run the tests in `test_ha.py` before the merge.

PhanLe1010:
`test_ha.py` tests passed

yasker (approved):

PR #763, opened 2020-11-07
Fix: volume remains attaching to the wrong node during a recurring backup

PhanLe1010:
We have discussed the option of removing the auto-reattachment feature entirely. However, I think that we shouldn't do it because:
1. We need the auto-reattachment feature in some cases. For example, when the volume is in the restoring process, if it is detached unexpectedly, Longhorn should auto-reattach the volume and continue doing the restoration. 

1. Removing the auto-reattachment feature doesn't solve the problem in this issue. The root cause of this problem is that when the volume is detached unexpectedly during a recurring backup job, the `volume.Spec.NodeID` field is not cleared (therefore, workload pod cannot uses the volume). The auto-reattachment feature lives inside VolumeController and VolumeController never changes volume's spec.

This PR proposes another approach:
1. We add a new field to the volume's spec: `LastAttachedBy`. This field indicates who is the last one that attached the volume. If the last attachment doesn't set this set field, it will be empty.
1. Change the logic of recurring backup job so that when the job automatically attaches the volume to a node, it also sets volume's `LastAttachedBy` field to be the job's name.
1. At the end of the recurring backup job, it will check to see if the volume was attached by the job's name. If so, the job tries to detach the volume.

This approach has another advantage: It doesn't only solve the problem when the volume is detached unexpectedly, but also solve the problem when the recurring job dies unexpectedly in the middle of a backup. 
Currently, if the recurring job dies unexpectedly in the middle of a backup, it cannot detach the volume that was attached by the job. The next time, the job runs, it sees that volume already attached so it will not detach the volume at the end. The end result is that the workload pod cannot use the volume forever. 

 longhorn/longhorn#1922

### DIFF
--- a/app/volume.go
+++ b/app/volume.go
@@ -34,7 +34,7 @@ const (
 	SnapshotPurgeStatusInterval = 5 * time.Second
 
 	WaitInterval              = 5 * time.Second
-	DetachingWaitInterval     = 30 * time.Second
+	DetachingWaitInterval     = 10 * time.Second
 	VolumeAttachTimeout       = 300 // 5 minutes
 	BackupProcessStartTimeout = 90  // 1.5 minutes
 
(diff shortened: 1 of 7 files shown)


app/volume.go
```diff
@@ -169,15 +169,58 @@ func NewJob(managerURL, volumeName, snapshotName string, labels map[string]strin
...
 	volumeName := job.volumeName
+	jobName, _ := job.labels[types.RecurringJobLabel]
+	if jobName == "" {
+		logrus.Warn("Missing RecurringJob label")
+		return
+	}
+
+	logrus.Infof("Handling volume detachment for volume %v", volumeName)
+	for {
+		volume, err := volumeAPI.ById(volumeName)
```
shuo-wu, line 187: What if the volume just becomes state `Detached` during the auto reattachment and when this line is executed?

shuo-wu (commented):
I still cannot get the idea of adding field `AttachedBy`.
What's the advantage compared with the solution that using a defer function to wait for attachment then issue detach request?

PhanLe1010:
> What's the advantage compared with the solution that using a defer function to wait for attachment then issue detach request?

There is one advantage that is:

Currently, if the recurring job pod dies unexpectedly in the middle of a backup, it cannot detach the volume that was attached by the job. The next time, the job runs, it sees that volume already attached so it will not detach the volume at the end. The end result is that the workload pod cannot use the volume forever.

PhanLe1010:
> For both cases, and waiting with timeout machenism may be enough but simpler.

Suppose that we wait for TIME_OUT mins. If the node that the volume was attached to goes down and comes up after TIME_OUT + 1 mins, the volume will be stuck in the wrong node forever.

app/volume.go
PhanLe1010: Awesome catch! You are right! I will fix this.

manager/volume.go
```diff
@@ -277,7 +277,7 @@ func (m *VolumeManager) Delete(name string) error {
 	return nil
 }
 
-func (m *VolumeManager) Attach(name, nodeID string, disableFrontend bool) (v *longhorn.Volume, err error) {
+func (m *VolumeManager) Attach(name, nodeID string, disableFrontend bool, attachedBy string) (v *longhorn.Volume, err error) {
```
yasker, line 280: Wrong separation of the commits. This build will fail due to the caller wasn't updated.

Each separate commit should be buildable.

yasker (changes requested):

yasker:
Also, `fix:` is enough. Don't need `fix the bug` in the commit title.

app/volume.go
PhanLe1010: Fixed

manager/volume.go
PhanLe1010: Thank you! Fixed

yasker (approved):

yasker:
LGTM. Waiting for @shuo-wu 's review.

shuo-wu (approved):

PR #764, opened 2020-11-10
upgrade k8s dependencies to v1.16.15

PhanLe1010:
upgrade k8s dependencies to v1.16.15
Also update Longhorn manager codebase to use the updated dependencies.

Note:

`k8s.io/kubernetes` is not primarily intended to be consumed as a module ([link](https://github.com/kubernetes/kubernetes/issues/79384#issuecomment-505725449)). People are moving away from depending on `k8s.io/kubernetes` directly . For example, [helm/helm](https://github.com/helm/helm/pull/6609)

We can keep using the `k8s.io/kubernetes` directly for now. In the future, when Kubernetes separates `controller` package and `utils` package into different modules, we can import those modules directly. 

longhorn/longhorn#1938


PhanLe1010:
Integration core tests passed 26 tests and failed at the test `ha_salvage_test `.
Looks like a known problem.

csi/node_server.go
```diff
@@ -136,16 +139,16 @@ func (ns *NodeServer) nodePublishMountVolume(volumeName, devicePath, targetPath,
...
 func (ns *NodeServer) nodePublishBlockVolume(volumeName, devicePath, targetPath string, mounter *mount.SafeFormatAndMount) (*csi.NodePublishVolumeResponse, error) {
 	targetDir := filepath.Dir(targetPath)
-	exists, err := mounter.ExistsPath(targetDir)
+	exists, err := hostUtil.PathExists(targetDir)
 	if err != nil {
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 	if !exists {
-		if err := mounter.MakeDir(targetDir); err != nil {
+		if err := makeDir(targetDir); err != nil {
```
yasker, line 147: Was this done in the Pod's namespace or host namespace?

csi/node_server.go
```diff
@@ -136,16 +139,16 @@ func (ns *NodeServer) nodePublishMountVolume(volumeName, devicePath, targetPath,
...
 			return nil, status.Errorf(codes.Internal, "Could not create dir %q: %v", targetDir, err)
 		}
 	}
-	if err = mounter.MakeFile(targetPath); err != nil {
+	if err = makeFile(targetPath); err != nil {
```
yasker, line 151: Same above.

go.mod
```diff
@@ -3,78 +3,79 @@ module github.com/longhorn/longhorn-manager
 go 1.13
 
 replace (
-	k8s.io/api => k8s.io/api v0.0.0-20191004102349-159aefb8556b
-	k8s.io/apimachinery => k8s.io/apimachinery v0.0.0-20191004074956-c5d2f014d689
-	k8s.io/client-go => k8s.io/client-go v11.0.1-0.20191004102930-01520b8320fc+incompatible
+	k8s.io/api => k8s.io/api v0.16.15
```
yasker, line 6: Is this the result after `go mod tidy`? Also, do we need ping so many packages here? Most of them don't look like direct dependencies.

PhanLe1010: > Is this the result after go mod tidy?
Yes, Kubernetes team tags the release to make it easier for the users to consume. e.g not having to read the longhorn hascode.

> Also, do we need ping so many packages here? Most of them don't look like direct dependencies.

Which one is not direct dependency?

csi/node_server.go
PhanLe1010: From what I am understanding, it is the same whether we execute this command in Pod's namespace or host namespace because we Bidirectionally mount `/var/lib/kubelet/pods` on the Pod to `/var/lib/kubelet/pods` on the host. How do you think?

On the other hand, we currently are using [mounter.MakeDir(targetDir)](https://github.com/longhorn/longhorn-manager/blob/9f3059094d3359dfdd00b3727989aaf9fd758732/vendor/k8s.io/kubernetes/pkg/util/mount/mount_linux.go#L389)  which does the same thing as the new one `makeDir ()`

csi/node_server.go
PhanLe1010: Same as https://github.com/longhorn/longhorn-manager/pull/764#discussion_r520897566

yasker (approved):

PR #765, opened 2020-11-10
Clean up poddisruptionbudget of non-existing instance manager pods

PhanLe1010:
longhorn/longhorn#1944

### DIFF
--- a/controller/instance_manager_controller.go
+++ b/controller/instance_manager_controller.go
@@ -426,6 +426,10 @@ func (imc *InstanceManagerController) syncInstanceManager(key string) (err error
 }
 
 func (imc *InstanceManagerController) syncInstanceManagerPDB(im *longhorn.InstanceManager) error {
+	if err := imc.cleanUpPDBOfNonExistingIMPod(); err != nil {
+		return err
+	}
+
 	if im.Status.CurrentState != types.InstanceManagerStateRunning {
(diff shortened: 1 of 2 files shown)


controller/instance_manager_controller.go
```diff
@@ -265,6 +265,11 @@ func (imc *InstanceManagerController) syncInstanceManager(key string) (err error
 	if err != nil {
 		if datastore.ErrorIsNotFound(err) {
 			logrus.Infof("Longhorn instance manager %v has been deleted", key)
+			if err := imc.ds.DeletePDB(getPDBNameFromIMName(name)); err != nil {
```
yasker, line 268: It's possible for us to miss this event. I don't think we should depend on the deletion notification. It's better to compare what we have set right now vs what's need to be set right now.

PhanLe1010: I see. Thank you!

PhanLe1010: Fixed. Thanks

PhanLe1010:
E2E tests doesn't pass. I am looking into it. `test_expansion_canceling `

PhanLe1010:
Update: E2E core tests passed except for `test_ha_salvage` which is a known problem.

yasker (approved):

PR #770, opened 2020-11-13
Fix: uninstallation takes long time to finish

PhanLe1010:
Previously, we delete node resources before deleting instance manager resources. This leads to a chaotic situation in which the ownership of instance manager resources gets transferred constantly between Longhorn managers ([see the log](https://github.com/longhorn/longhorn/issues/1653#issuecomment-670109463)). There are always conflicts when updating the instance manager resources in `etcd` and the Longhorn managers cannot remove the `finalizer` for the instance manager resources.

Now, we delete instance manager resources, then delete node resources. The problem is gone.

longhorn/longhorn#1653

### DIFF
--- a/controller/uninstall_controller.go
+++ b/controller/uninstall_controller.go
@@ -338,18 +338,18 @@ func (c *UninstallController) deleteCRDs() (bool, error) {
 		return true, c.deleteEngineImages(engineImages)
 	}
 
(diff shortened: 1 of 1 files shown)


PhanLe1010:
I think that we can skip running the E2E tests for this PR

yasker (approved):

PR #771, opened 2020-11-17
Allow user to set all kind of taint tolerations

PhanLe1010:
We allow the user to set all kinds of taint tolerations including the ones that contain `kubernetes.io` substring. This allows users to run Longhorn on control/etcd nodes.

We remember which tolerations applied by us in the annotation `longhorn.io/last-applied-tolerations`

Also, allow empty effect to enable 2 special cases for tolerations:

>An empty key with operator Exists matches all keys, values and effects which means this will tolerate everything.
An empty effect matches all effects with key key.

Also, fix the bug to allow multiple tolerations with the same key

longhorn/longhorn#1874 
longhorn/longhorn#1977

### DIFF
--- a/controller/engine_image_controller.go
+++ b/controller/engine_image_controller.go
@@ -1,6 +1,7 @@
 package controller
 
 import (
+	"encoding/json"
 	"fmt"
 	"reflect"
 	"time"
@@ -276,7 +277,10 @@ func (ic *EngineImageController) syncEngineImage(key string) (err error) {
 			return errors.Wrapf(err, "failed to get system pods image pull policy before creating engine image daemonset")
 		}
 
-		dsSpec := ic.createEngineImageDaemonSetSpec(engineImage, tolerations, priorityClass, registrySecret, imagePullPolicy)
+		dsSpec, err := ic.createEngineImageDaemonSetSpec(engineImage, tolerations, priorityClass, registrySecret, imagePullPolicy)
+		if err != nil {
+			return errors.Wrapf(err, "fail to create daemonset spec for engine image %v", engineImage.Name)
+		}
 
 		if err = ic.ds.CreateEngineImageDaemonSet(dsSpec); err != nil {
 			return errors.Wrapf(err, "fail to create daemonset for engine image %v", engineImage.Name)
@@ -550,7 +554,7 @@ func (ic *EngineImageController) ResolveRefAndEnqueue(namespace string, ref *met
 }
 
 func (ic *EngineImageController) createEngineImageDaemonSetSpec(ei *longhorn.EngineImage, tolerations []v1.Toleration,
-	priorityClass, registrySecret string, imagePullPolicy v1.PullPolicy) *appsv1.DaemonSet {
+	priorityClass, registrySecret string, imagePullPolicy v1.PullPolicy) (*appsv1.DaemonSet, error) {
 
 	dsName := types.GetDaemonSetNameFromEngineImageName(ei.Name)
 	image := ei.Spec.Image
@@ -565,9 +569,15 @@ func (ic *EngineImageController) createEngineImageDaemonSetSpec(ei *longhorn.Eng
 	}
 	maxUnavailable := intstr.FromString(`100%`)
 	privileged := true
+	tolerationsByte, err := json.Marshal(tolerations)
+	if err != nil {
+		return nil, err
+	}
+
 	d := &appsv1.DaemonSet{
 		ObjectMeta: metav1.ObjectMeta{
-			Name: dsName,
+			Name:        dsName,
+			Annotations: map[string]string{types.GetLonghornLabelKey(types.LastAppliedTolerationAnnotationKeySuffix): string(tolerationsByte)},
 		},
 		Spec: appsv1.DaemonSetSpec{
 			Selector: &metav1.LabelSelector{
@@ -642,5 +652,5 @@ func (ic *EngineImageController) createEngineImageDaemonSetSpec(ei *longhorn.Eng
 		}
 	}
 
-	return d
+	return d, nil
 }
--- a/controller/setting_controller.go
+++ b/controller/setting_controller.go
@@ -332,10 +332,7 @@ func (sc *SettingController) updateTaintToleration() error {
 		if reflect.DeepEqual(util.TolerationListToMap(lastAppliedTolerationsList), newTolerationsMap) {
 			continue
 		}
-		if err := updateTolerationForDeployment(dp, lastAppliedTolerationsList, newTolerationsList); err != nil {
-			return err
-		}
-		if _, err := sc.ds.UpdateDeployment(dp); err != nil {
+		if err := sc.updateTolerationForDeployment(dp, lastAppliedTolerationsList, newTolerationsList); err != nil {
 			return err
 		}
 	}
@@ -348,10 +345,7 @@ func (sc *SettingController) updateTaintToleration() error {
 		if reflect.DeepEqual(util.TolerationListToMap(lastAppliedTolerationsList), newTolerationsMap) {
 			continue
 		}
-		if err := updateTolerationForDaemonset(ds, lastAppliedTolerationsList, newTolerationsList); err != nil {
-			return err
-		}
-		if _, err := sc.ds.UpdateDaemonSet(ds); err != nil {
+		if err := sc.updateTolerationForDaemonset(ds, lastAppliedTolerationsList, newTolerationsList); err != nil {
 			return err
 		}
 	}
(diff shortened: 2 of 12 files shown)


PhanLe1010:
Running E2E tests

PhanLe1010:
E2E core tests passed

controller/engine_image_controller.go
```diff
@@ -254,10 +254,11 @@ func (ic *EngineImageController) syncEngineImage(key string) (err error) {
 		return errors.Wrapf(err, "cannot get daemonset for engine image %v", engineImage.Name)
 	}
 	if ds == nil {
-		tolerations, err := ic.ds.GetSettingTaintToleration()
+		setting, err := ic.ds.GetSetting(types.SettingNameTaintToleration)
```
yasker, line 257: Is there any other place using the original function?

yasker (changes requested):

controller/engine_image_controller.go
PhanLe1010: Yes, there are a few places. I modified them all except for the this place https://github.com/longhorn/longhorn-manager/blob/8af120dd7c109d81d86118e482f88a26ddf7650f/controller/volume_controller.go#L1982. I think that it is ok to keep it the same here.

yasker: Hmm, we don't need this change in cronjob?

PhanLe1010: The current design for cron job is that if any field in cron job' spec is different than the desired spec, we update the cron job to the desire state,[link](https://github.com/longhorn/longhorn-manager/blob/8af120dd7c109d81d86118e482f88a26ddf7650f/controller/volume_controller.go#L2083). 
This logic already cover the modification to tolerations list. 

I know that this would be a problem if K8s automatically fill in some default values that is not specify by the desired spec. In this case, Longhorn will keep creating/deleting cron job forever. So far we don't hit this issue yet. If we hit it, it should be an other issue?

yasker: It seems safer if we use the same annotation method? Of course, then we need to change the `deepEquals` a bit.

util/util.go
```diff
@@ -685,31 +687,47 @@ func IsKubernetesDefaultToleration(toleration v1.Toleration) bool {
...
+	objMeta.SetAnnotations(annos)
+
+	return nil
 }
 
 func TolerationListToMap(tolerationList []v1.Toleration) map[string]v1.Toleration {
 	res := map[string]v1.Toleration{}
 	for _, t := range tolerationList {
-		res[t.Key] = t
+		res[GetTolerationChecksum(t)] = t
```
shuo-wu, line 722: Add comments to explain why we need to use checksum instead.

shuo-wu (changes requested):
I think we may need to update annotations for existing workloads for v.1.10 upgrade.
And remember to check integration test `test_setting_toleration`.

util/util.go
PhanLe1010: Awesome catch!!!
Yeah, we need to be careful with v.1.10 upgrade. If we don't we might kill all instance manager pod while the volumes are in the attached state. It will lead to data corruption.

controller/engine_image_controller.go
PhanLe1010: @yasker 
Actually, the current logic for comparison is not correct. This check always fails:
https://github.com/longhorn/longhorn-manager/blob/8af120dd7c109d81d86118e482f88a26ddf7650f/controller/volume_controller.go#L2083
And  Longhorn constantly updates cronjobs every reconciliation cycle. 

I have re-write the comparison logic using annotation.

util/util.go
PhanLe1010: Thank you! Fixed

PhanLe1010:
I still working on making sure Longhorn v1.1.0 upgrading is ok.

util/util.go
shuo-wu: with the same t.Key but different operator/effect/value?

PhanLe1010: Sure! Added

shuo-wu:
Can Longhorn directly add all tolerations except for `kubernetes.io` to the annotations for existing longhorn workloads during the upgrade? What's the main blocker/concern for the upgrade part in this PR?

PhanLe1010:
> Can Longhorn directly add all tolerations except for kubernetes.io to the annotations for existing longhorn workloads during the upgrade? What's the main blocker/concern for the upgrade part in this PR? 

Yes, we can. There is no blocker here. It is just that I haven't finished the code for the upgrade part yet. I took a few hours off today.

PhanLe1010:
Test `test_setting_toleration ` passed after the fix https://github.com/longhorn/longhorn-tests/pull/486

PhanLe1010:
E2E core tests passed

util/util.go
```diff
@@ -685,31 +686,50 @@ func IsKubernetesDefaultToleration(toleration v1.Toleration) bool {
 	return false
 }
 
-func AreIdenticalTolerations(oldTolerations, newTolerations map[string]v1.Toleration) bool {
-	// modified existing tolerations
-	for name := range oldTolerations {
-		if !IsKubernetesDefaultToleration(oldTolerations[name]) && !reflect.DeepEqual(newTolerations[name], oldTolerations[name]) {
-			return false
-		}
+func GetAnnotation(obj runtime.Object, annotationKey string) (string, error) {
```
shuo-wu, line 689: I think it's better to put these 2 functions in `datastore/kubernetes`

controller/setting_controller.go
```diff
@@ -322,27 +324,62 @@ func (sc *SettingController) updateTaintToleration() error {
...
+		lastAppliedTolerationsList, err := getLastAppliedTolerationsList(dp)
+		if err != nil {
+			return err
+		}
+		if reflect.DeepEqual(util.TolerationListToMap(lastAppliedTolerationsList), newTolerationsMap) {
 			continue
 		}
-		dp.Spec.Template.Spec.Tolerations = getFinalTolerations(util.TolerationListToMap(dp.Spec.Template.Spec.Tolerations), newTolerations)
+
+		dp.Spec.Template.Spec.Tolerations = getFinalTolerations(util.TolerationListToMap(dp.Spec.Template.Spec.Tolerations),
```
shuo-wu, line 335: How about using a single function to update tolerations and the related annotations for all Longhorn workloads? There are too many details of operating tolerations here, which is not straightforward.

Similar below.

util/util.go
PhanLe1010: I think that `datastore/kubernetes` contains functions that make requests to Kube API. However, `GetAnnotation()` and `SetAnnotation()` function here only extract the data from its input parameters. Moreover, the 2 functions are used in the upgrade part which doesn't have access to DS. So, I think that they should be in `util`
How do you think?

controller/setting_controller.go
PhanLe1010: Do you mean that we use a separate function to update both tolerations and annotations?
If so, yeah, I agree!

PhanLe1010: Fixed. Thanks

PhanLe1010:
@shuo-wu  Would you able to take a look again?

util/util.go
shuo-wu: OK. It makes sense.

controller/setting_controller.go
```diff
@@ -351,6 +373,54 @@ func (sc *SettingController) updateTaintToleration() error {
 	return nil
 }
 
+func updateTolerationForDeployment(ds *appsv1.Deployment, lastAppliedTolerations, newTolerations []v1.Toleration) error {
```
shuo-wu: Is it possible to use a common function rather than 2 separate functions here? Notice that both functions will update `PodTemplateSpec` and `Annotations`

PhanLe1010: It would be something like this:
```
func (sc *SettingController) updateTolerationAndAnnotation(podTemplateSpec *v1.PodTemplateSpec, obj runtime.Object, lastAppliedTolerations, newTolerations []v1.Toleration) error {
	existingTolerationsMap := util.TolerationListToMap(podTemplateSpec.Spec.Tolerations)
	lastAppliedTolerationsMap := util.TolerationListToMap(lastAppliedTolerations)
	newTolerationsMap := util.TolerationListToMap(newTolerations)
	podTemplateSpec.Spec.Tolerations = getFinalTolerations(existingTolerationsMap, lastAppliedTolerationsMap, newTolerationsMap)
	newTolerationsByte, err := json.Marshal(newTolerations)
	if err != nil {
		return err
	}
	if err := util.SetAnnotation(obj, types.GetLonghornLabelKey(types.LastAppliedTolerationAnnotationKeySuffix), string(newTolerationsByte)); err != nil {
		return err
	}
	return nil
}
///////////////////////////////
sc.updateTolerationAndAnnotation(dp.Spec.Template, dp, lastAppliedTolerationsList, newTolerationsList)
//////////////////////////////
sc.updateTolerationAndAnnotation(ds.Spec.Template, ds, lastAppliedTolerationsList, newTolerationsList)
```

imho, the separation of the 2 functions is a bit repeating code but it is easier to read.  
However, if you prefer the compact form, it totally makes sense too. 
Please let me know which way you prefer?

shuo-wu: OK. Let's retain the current implementation

shuo-wu (approved):

yasker (approved):

PR #772, opened 2020-11-21
Move registry secret setting to outside of  default setting

PhanLe1010:
There are 2 reasons for doing this:
1. To avoid confusing users. In v1.0.2, users have to specify `Private Registry Secret` in the default setting section which is hidden by default. See more at [link](https://github.com/longhorn/longhorn/issues/1670#issuecomment-670723484). With this fix, all the private registry settings are on the same `Private Registry Setting` section. See more at [link](https://github.com/longhorn/longhorn/blob/e020e22913fd3f92c7bfc9d513a5bc86ecd7e0b8/chart/values.yaml#L98)
1. In v1.0.2, when users updating Longhorn deployment and specify a new setting for `registry secret `, it won't be applied because the old `registry secret ` already existed. With this fix, the new value of  `registry secret ` will be applied.

The approach is similar to how we handle the default engine image setting. First, users specify `privateRegistry.registrySecret` in helm chart. Longhorn chart copies  the registry secret setting to ImagePullSecrets field in Longhorn manager daemonset yaml. When manager pod starts, it lookups the field ImagePullSecrets and update the setting SettingNameRegistrySecret to the corresponding value.

longhorn/longhorn#1670

### DIFF
--- a/app/daemon.go
+++ b/app/daemon.go
@@ -144,6 +144,9 @@ func startManager(c *cli.Context) error {
 	if err := updateSettingDefaultInstanceManagerImage(m, instanceManagerImage); err != nil {
 		return err
 	}
+	if err := updateRegistrySecretName(m); err != nil {
+		return err
+	}
 
 	if err := initDaemonNode(ds); err != nil {
 		return err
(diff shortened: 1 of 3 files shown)


shuo-wu (dismissed):

yasker (changes requested):
We need to call it out at least in the commit message on `how` we did it, rather than just what the user can do now.

PhanLe1010:
Thank @yasker  I added the explanation

yasker (approved):

PR #773, opened 2020-11-24
Always set e.Spec.SalvageRequested=true when all volume's replicas fail

PhanLe1010:
Why do we need this fix?
When users disable the revision counter, Longhorn will use `e.Spec.SalvageRequested` as an indicator to decide whether it needs to choose the best replica to salvage the volume when the engine process starts. If all replicas of volumes fail, and users disable the revision counter, and `autoSalvage=true`, and we fail to set `e.Spec.SalvageRequested=true`, Longhorn will pick all replicas when the volume's engine restarts. It is problematic because replicas might not be in sync. 

How does this PR fix this issue?
When all volume's replicas fail and `autoSalvage==true`, we set `e.Spec.SalvageRequested=true` and make sure that it persists in etcd before auto-salvage the replicas. This makes sure that when the engine process restarts, it picks the correct replica for salvage.

longhorn/longhorn#1957

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -902,7 +902,20 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
 		if err != nil {
 			return err
 		}
-		// make sure the volume is detached before automatically salvage
+		// To make sure that we don't miss the `allFaulted` event, This IF statement makes sure the `e.Spec.SalvageRequested=true`
+		// persist in ETCD before Longhorn salvages the failed replicas in the IF statement below it.
+		// More explanation: when all replicas fails, Longhorn tries to set `e.Spec.SalvageRequested=true`
+		// and try to detach the volume by setting `v.Status.CurrentNodeID = ""`.
+		// Because at the end of volume syncVolume(), Longhorn updates CRs in the order: replicas, engine, volume,
+		// when volume changes from v.Status.State == types.VolumeStateAttached to v.Status.State == types.VolumeStateDetached,
+		// we know that volume RS has been updated and therefore the engine RS also has been updated and persisted in ETCD.
+		// At this moment, Longhorn goes into the IF statement below this IF statement and salvage all replicas.
+		if autoSalvage && !v.Status.IsStandby && !v.Status.RestoreRequired {
+			// Since all replica failed and autoSalvage is enable, mark engine controller salvage requested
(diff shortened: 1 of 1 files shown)


controller/volume_controller.go
```diff
@@ -902,7 +902,12 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
 		if err != nil {
 			return err
 		}
-		// make sure the volume is detached before automatically salvage
+		if autoSalvage && !v.Status.IsStandby && !v.Status.RestoreRequired {
```
yasker, line 905: I don't understand why moving this code will solve the problem.

1. What triggered the problem before?
2. Why we no longer need to check if the data exists or not.
3. Why previously the logic only applies to if the node is not down or deleted?

yasker (changes requested):

controller/volume_controller.go
PhanLe1010: > What triggered the problem before?
This was reason that triggered the problem https://github.com/longhorn/longhorn/issues/1942#issuecomment-722698783. In short, Longhorn sometime succeeds to salvage replicas and save replicas CRs to `etcd` but fail to save engine to `etcd`.

> Why we no longer need to check if the data exists or not.
> Why previously the logic only applies to if the node is not down or deleted?

We still do all of this check before salvage the failed replicas in Volume Controller. 

However, I don't think that we need to do those check for before setting ` e.Spec.SalvageRequested=true` because when all replicas fail and `autoSalvage==true`, we must set ` e.Spec.SalvageRequested=true` so that engine picks the best replica the next time it starts no matter whether the current node is down or deleted, or data exists.  If we don't set `e.Spec.SalvageRequested=true`, Longhorn will pick all non-failed replicas the next time engine starts.

PhanLe1010:
Added more explanation to the code and the commit message

controller/volume_controller.go
yasker: Talked with Phan, `v.Status.State == types.VolumeStateDetached` works as a barrier to make sure the volume status was updated. And since engine update will be done before the volume update, so it also guarantees the engine state has been updated in the etcd.

yasker (approved):
LGTM. @shuo-wu ?

PhanLe1010:
E2E core tests passed

shuo-wu (approved):

PR #778, opened 2020-12-02
1470 rwx single volume

joshimoo:


controller/share_manager_controller.go
```diff
@@ -0,0 +1,626 @@
...
+
+	if err = c.syncShareManagerPod(sm); err != nil {
+		return err
+	}
+
+	if sm.Status.State == types.ShareManagerStateRunning {
+		// running is once the pod is in ready state
+		// which means the nfs server is up and running with the volume attached
+		// the cluster service ip doesn't change for the lifetime of the volume
+		if sm.Status.Endpoint == "" {
```
shuo-wu, line 341: It's better to use level based assignment rather than edge triggered assignment for Endpoint, which is similar to https://github.com/longhorn/longhorn-manager/pull/734

controller/share_manager_controller.go
```diff
@@ -0,0 +1,626 @@
...
+
+				if volume, err = c.ds.UpdateVolume(volume); err != nil {
+					return err
+				}
+			} else {
+				// manually requeue the share manager so we can check again for container ready just in case
+				key := volume.Namespace + "/" + volume.Name
+				c.queue.AddRateLimited(key)
+			}
+		} else if sm.Status.State == types.ShareManagerStateStarting {
```
shuo-wu, line 455: Can the share manager be directly updated from state stopped to state running?

shuo-wu (dismissed):

controller/share_manager_controller.go
joshimoo: - If there is no pod we create one and transition into "starting" state

- If there is a pending pod but we are not in "starting" state, we transition to "error" state
- If there is a running pod but we are not in "starting / running" state we transition to "error" state

- If we are in "error" state we delete the pod

So no it shouldn't be able to go from stopped into running state.

yasker (dismissed):

yasker:
Need to wait for the integration test result.

shuo-wu (dismissed):

joshimoo:
e2e core test passed except `test_csi_expansion_with_replica_failure` which also failed in nightly.

yasker (approved):

PR #780, opened 2020-12-04
controller: mark instance as ERROR if the related instance manager is not found and the node is running

shuo-wu:
For an instance, nil instance manager or instance manager becoming nil means that the related node should be down or deleted. If the node is still running, it means something is wrong with this instance. Then we can directly mark it as ERROR rather than UNKNOWN so that Longhorn will clean up it automatically.

longhorn/longhorn#1118


shuo-wu:
According to the comment in the issue, the only way to trigger the instance manager not found error is by deleting instance manager CR. But deleting IM cannot test this PR since instance-manager-controller will mark instances as ERROR, too.
Hence, I am not sure if we still need this PR.

PR #787, opened 2020-12-08
Restart share manager pods, when tolerations/priority class changes

joshimoo:
There is an interesting behavior, that we can only change the toleration settings while all volumes are detached, so the user needs to scale down all rwx volumes, for the sharemanager to detach the volume.

https://github.com/longhorn/longhorn/issues/2049


joshimoo:
verified the existence of the updated annotation + tolerations manually.

yasker (approved):

PR #789, opened 2020-12-10
Turn a non shared volume into a shared volume when accessed via a RWX pvc

joshimoo:
Adds an api endpoint updateAccessMode to the volume api which is used by the csi driver to turn a non shared volume into a shared volume when accessed via a rwx pvc. it's only possible to call this api while the volume is detached.

https://github.com/longhorn/longhorn/issues/2071


joshimoo:
manually tested turning a non shared volume into a shared volume via a RWX pvc.

shuo-wu (dismissed):
In general LGTM except for missing the upgrade for existing volumes.

api/model.go
```diff
@@ -889,7 +902,7 @@ func toVolumeResource(v *longhorn.Volume, ves []*longhorn.Engine, vrs []*longhor
 		RevisionCounterDisabled: v.Spec.RevisionCounterDisabled,
 		Ready:                   ready,
 
-		Share:         v.Spec.Share,
+		AccessMode:    v.Spec.AccessMode,
```
shuo-wu, line 905: If you decide not to update field `v.Spec.AccessMode` for existing volumes, you may still need to make sure this field exposed to UI is not empty.

joshimoo:
added the upgrade part for existing volumes

joshimoo:
Added a check at the manager level to ensure volume is detached before updating the access mode.

yasker (approved):

PR #790, opened 2020-12-10
Remove host ipc from the csi plugin deployment

joshimoo:
We only require the pid, mnt, net and uts namespaces

longhorn/longhorn#2074


yasker (approved):

yasker (commented):
Has this change been tested? Our PSP also doesn't allow hostNetwork.

https://github.com/longhorn/longhorn-manager/blob/b467f4fbb9f936b91830449a125cea04c9b0b6b6/deploy/install/01-prerequisite/05-podsecuritypolicy.yaml#L12

yasker (approved):

PR #792, opened 2020-12-10
Cleanup orphaned share managers for no longer existing volumes

joshimoo:
This cleans up orphaned share managers for no longer existing volumes, for whatever reason a volume might no longer be present, while the share manager crd is still present.

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -2380,7 +2380,7 @@ func (vc *VolumeController) ReconcileShareManagerState(volume *longhorn.Volume) 
 
 	// no ShareManager create a new one
 	if sm == nil {
-		sm, err = vc.createShareManagerForVolume(volume.Name, image)
+		sm, err = vc.createShareManagerForVolume(volume, image)
 		if err != nil {
 			log.WithError(err).Errorf("Failed to create share manager %v", volume.Name)
 			return err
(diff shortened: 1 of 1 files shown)


shuo-wu (approved):
Make sense. And it's better to explain why cleaning up engine/replica is unnecessary here. (There is an owner reference set for engine/replica but not for share manager)

shuo-wu (commented):
~~Make sense. And it's better to explain why cleaning up engine/replica is unnecessary here. (There is an owner reference set for engine/replica but not for share manager)~~
Sorry for my replicate comment due to the bad network.

joshimoo:
Good point since the share manager now only controls a single volume, we can add a owner reference and have it cleaned up with the volume.

controller/volume_controller.go
```diff
@@ -242,7 +242,14 @@ func (vc *VolumeController) syncVolume(key string) (err error) {
 	volume, err := vc.ds.GetVolume(name)
 	if err != nil {
 		if datastore.ErrorIsNotFound(err) {
-			vc.logger.WithField("volume", name).Debug("Longhorn volume has been deleted")
+			log := vc.logger.WithField("volume", name)
+			log.Debug("Longhorn volume has been deleted")
+
+			if sm, _ := vc.ds.GetShareManager(name); sm != nil {
+				log.Info("Removing orphaned share manager for deleted volume")
+				_ = vc.ds.DeleteShareManager(name)
```
yasker, line 250: Failure to delete should expose a warning.

yasker (changes requested):

controller/volume_controller.go
joshimoo: Removed, since I added a volume owner reference to the share manager so kubernetes can garbage collect them.
If we miss the deletion for whatever reason.

controller/volume_controller.go
```diff
@@ -2412,12 +2412,13 @@ func (vc *VolumeController) ReconcileShareManagerState(volume *longhorn.Volume)
...
+func (vc *VolumeController) createShareManagerForVolume(volume *longhorn.Volume, image string) (*longhorn.ShareManager, error) {
 	sm := &longhorn.ShareManager{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      volume,
-			Namespace: vc.namespace,
-			Labels:    types.GetShareManagerLabels(volume, image),
+			Name:            volume.Name,
+			Namespace:       vc.namespace,
+			Labels:          types.GetShareManagerLabels(volume.Name, image),
+			OwnerReferences: datastore.GetOwnerReferencesForVolume(volume),
```
yasker, line 2421: Have we set the cascading deletion for the volume object?

joshimoo: We don't specify and therefore use the default option of orphan.
But we do have a deletion call as part of the volume deletion already, as well as a finalizer (already) on the volume object.

Same as with engine/replica, just making it consistent with the ownership reference.
So that in the case if there is a share manager crd without the associated volume for whatever reason the GC can clean it up.

yasker (approved):

PR #793, opened 2020-12-10
Add support for DR volumes in RWX mode

joshimoo:
While a volume is in standby mode there is no need for the volume to be
controlled by a share manager, since the user needs to activate a volume
before it can be used by workloads.

### DIFF
--- a/controller/share_manager_controller.go
+++ b/controller/share_manager_controller.go
@@ -388,6 +388,12 @@ func (c *ShareManagerController) syncShareManagerPod(sm *longhorn.ShareManager) 
 		return err
 	}
 
+	// no need to expose standby volumes via a share manager
+	if volume.Spec.Standby || volume.Status.IsStandby {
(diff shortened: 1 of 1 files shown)


shuo-wu:
Does a DR volume need RWX mode? DR volume is attached without frontend and there is no workload being able to use it. IMO, RWX mode should be disabled for DR volumes.

joshimoo:
@shuo-wu the share manager won't do anything if the volume is in `StandBy` mode. If I understood correctly the objective of the DR volume is to replicate a volume from another cluster, for quicker failover. Assuming the user uses the volume in RWX mode on cluster A, then he wants to fail over to cluster B it makes sense that the user has the ability to specify an access mode.

So the user only needs to go through the normal activate process. What do you think?

shuo-wu:
OK. it makes sense.

shuo-wu (approved):

PR #794, opened 2020-12-11
create pv/pvc based on the volumes access mode

joshimoo:
This also includes 2 minor fixes in separate commits.

### DIFF
--- a/datastore/kubernetes.go
+++ b/datastore/kubernetes.go
@@ -434,11 +434,16 @@ func NewPVManifestForVolume(v *longhorn.Volume, pvName, storageClassName, fsType
 		"staleReplicaTimeout": strconv.Itoa(v.Spec.StaleReplicaTimeout),
 	}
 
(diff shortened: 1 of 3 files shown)


datastore/kubernetes.go
```diff
@@ -434,7 +434,15 @@ func NewPVManifestForVolume(v *longhorn.Volume, pvName, storageClassName, fsType
 		"staleReplicaTimeout": strconv.Itoa(v.Spec.StaleReplicaTimeout),
 	}
 
-	return NewPVManifest(v.Spec.Size, pvName, v.Name, storageClassName, fsType, volAttributes)
+	// by default we create rwo manifests
```
shuo-wu, line 437: Why not directly modify the function `NewPVManifest`? Same below

joshimoo: updated to pass the access mode specifically when creating the manifest

shuo-wu (approved):

PR #795, opened 2020-12-11
Fix: volume controller constantly update the ScheduledCondition

PhanLe1010:
When a volume is created with number of replicas more than available nodes and `AllowVolumeCreationWithDegradedAvailability` is enabled, volume controller constantly updates the volume's `ScheduledCondition`. This prevents other components to update the volume's spec since there will be a lot of conflicts. As a result, volume sometimes takes a long time to attach.

This PR makes sure volume controller only update volume's condition if the condition's status has changed

Longhorn/longhorn#2073

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -383,6 +383,8 @@ func (vc *VolumeController) syncVolume(key string) (err error) {
 		}
 		// stop updating if engines and replicas weren't fully updated
 		if lastErr == nil {
+			// Make sure that we don't update condition's LastTransitionTime if the condition's values hasn't changed
+			handleConditionLastTransitionTime(&existingVolume.Status, &volume.Status)
 			if !reflect.DeepEqual(existingVolume.Status, volume.Status) {
 				// reuse err
 				_, err = vc.ds.UpdateVolumeStatus(volume)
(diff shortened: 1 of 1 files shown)


PhanLe1010:
Notice that the `Condition` struct contains the time `LastTransitionTime`. Therefore, when the old code updates the VolumeConditionTypeScheduled, even if update the  VolumeConditionTypeScheduled to the same value, the Volume object still consider as has changed since the VolumeConditionTypeScheduled's LastTransitionTime has changed. This is a bit tricky.

controller/volume_controller.go
```diff
@@ -430,6 +432,15 @@ func (vc *VolumeController) syncVolume(key string) (err error) {
 	return nil
 }
 
+// handleConditionLastTransitionTime rollback to the existing condition object if condition's value hasn't changed
+func handleConditionLastTransitionTime(existingStatus, newStatus *types.VolumeStatus) {
+	for conditionType, condition := range newStatus.Conditions {
+		if existingCondition, ok := existingStatus.Conditions[conditionType]; ok && existingCondition.Status == condition.Status {
```
yasker, line 438: Also need to check the reason and message to ensure they're the same one.

PhanLe1010: Thank you! Fixed

yasker (approved):
LGTM.

yasker:
Is this ready to be merged?

PhanLe1010:
Running E2E core tests but I think that it is safe to merge since the fix is simple

PR #798, opened 2020-12-11
Marshal empty tolerations for share manager annotation

joshimoo:
https://github.com/longhorn/longhorn/issues/2049


shuo-wu (changes requested):
The function `GetDistinctTolerations` calling is missed when applying the tolerations to the pod manifest. Otherwise, LGTM.

joshimoo:
added missing GetDistinctTolerations call

yasker (approved):

PR #799, opened 2020-12-11
No need to call attach for shared volumes from the csi driver

joshimoo:
Also includes the full error message on a mount failure for shared volumes
We can then add information to the knowledge base to help the user diagnose failures

### DIFF
--- a/csi/controller_server.go
+++ b/csi/controller_server.go
@@ -302,26 +302,6 @@ func (cs *ControllerServer) ControllerPublishVolume(ctx context.Context, req *cs
 			logrus.Infof("Changed Volume %s access mode to RWX", req.GetVolumeId())
 		}
 
-		if existVol.State == string(types.VolumeStateAttached) {
-			if !existVol.Ready || len(existVol.Controllers) == 0 || existVol.Controllers[0].Endpoint == "" {
-				return nil, status.Errorf(codes.Aborted,
-					"The volume %s is already attached but it is not ready for workloads", req.GetVolumeId())
-			}
-
-			logrus.Infof("ControllerPublishVolume: shared volume %s is attached to %s new workload on %s",
-				req.GetVolumeId(), existVol.Controllers[0].HostId, req.GetNodeId())
-		} else {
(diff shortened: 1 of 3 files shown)


csi/controller_server.go
```diff
@@ -302,26 +302,6 @@ func (cs *ControllerServer) ControllerPublishVolume(ctx context.Context, req *cs
 			logrus.Infof("Changed Volume %s access mode to RWX", req.GetVolumeId())
 		}
 
-		if existVol.State == string(types.VolumeStateAttached) {
-			if !existVol.Ready || len(existVol.Controllers) == 0 || existVol.Controllers[0].Endpoint == "" {
```
yasker, line 306: Does that mean we no longer colocate share manager with the first consumer workload?

joshimoo: Correct, since we only manage a single volume, we no longer need the information.
Whichever longhorn-manager claims, it first will get it, we keep the share manager pod always on the same node as the longhorn-manager which is the owner of the share-manager cr.

Also conceptually it would be a fair to assume since the volume is supposed to be used in rwx mode, that the workloads will be distributed across nodes.

yasker: I still feel it's better if the shared manager is at least on the node for at least one workload. Why it's better to remove this?

joshimoo: The current single share manager per volume implementation does not wait with creation of the share manager till a volume is attached. i.e. no spec on the share manager object.

We could add functionality like this again, but it makes the ownership logic more complicated, since we need to prioritize ownership of the share manager, crd to the volume.spec.NodeID the first time ownership will be claimed.

Let me know if you want to add this functionality, for 1.1 ?

I liked the idea of the longhorn manager that owns the crd also having the pod on the same node as the manager.
But it's not strictly necessary, but complicates the ownership and lifetime of the pod handling.

yasker: If the colocation is not already working, we can remove this part of code.

yasker (approved):

PR #807, opened 2020-12-16
controller: KubeNodeController stop filtering keys in the enqueue function

shuo-wu:
longhorn/longhorn#2100

### DIFF
--- a/controller/kubernetes_node_controller.go
+++ b/controller/kubernetes_node_controller.go
@@ -89,8 +89,8 @@ func NewKubernetesNodeController(
 	)
 
 	kubeNodeInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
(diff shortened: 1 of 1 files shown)


yasker:
Is it possible for this PR to disrupt our node failure handling? I am a bit concerned to put it in just before the v1.1.0 release.

PhanLe1010:
LGTM. 
So in this PR, we allow KubeNodeController controller to handle the deletion of Longhorn nodes with different ID. There might be duplicated deletion requests from multiple managers but I don't see a problem.

shuo-wu:
IIRC, duplicate deletion calls won't cause the deletion timestamp update. Hence I think it's fine. And there will be update conflicts still when controllers want to pick up an object containing field`ownerID`. 

Since merging this PR means re-test the node down & deletion case, we can push it to the next release while finding a workaround for now

yasker:
Let's punt it to the next release.

yasker (approved):

PR #808, opened 2020-12-16
package: update base image to get the latest changes

meldafrawi:
https://github.com/longhorn/longhorn/issues/2091

Signed-off-by: Mohamed Eldafrawi <mohamed.eldafrawi@rancher.com>

### DIFF
--- a/scripts/package
+++ b/scripts/package
@@ -19,6 +19,10 @@ fi
 
 trap 'rm -rf ./package/bin' exit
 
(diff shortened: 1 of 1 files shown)


meldafrawi:
Base image updated before build.


image

yasker (commented):
This change will still require any developers to update the images before the build. Can we do this in drone build instead?

yasker (approved):

PR #814, opened 2021-01-05
Fix csi plugin nil pointer for storage classes with no defined params

joshimoo:
This fixes an issue for RWX volume creation as well as volume creation
based on a snapshot where the storage class has no params defined.

longhorn/longhorn#2154


yasker (approved):

yasker:
LGTM. @shuo-wu ?

shuo-wu (approved):

PR #815, opened 2021-01-08
controller: mark replica failed when we detect a bad replica while doing SnapshotPurge

PhanLe1010:
For the error during the SnapshotPurge progress, it will be recorded by
the engine/replica process automatically. Then once there is an error
in the engine's purgeStatus, the volume controller will mark the replica
as failed.

longhorn/longhorn#1895

### DIFF
--- a/controller/events.go
+++ b/controller/events.go
@@ -15,6 +15,7 @@ const (
 	EventReasonFailedRebuilding = "FailedRebuilding"
 
 	EventReasonFailedStartingSnapshotPurge = "FailedStartingSnapshotPurge"
+	EventReasonFailedSnapshotPurge         = "FailedSnapshotPurge"
 
 	EventReasonFailedRestore = "FailedRestore"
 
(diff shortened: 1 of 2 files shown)


PhanLe1010:
Running E2E core tests

joshimoo (commented):
We treat the purge failure same as restore failure, for now that seems okay, maybe down the line we can differentiate between different failure modes as you suggested.
Let @shuo-wu review and approve

shuo-wu (commented):
Is there any purge error that can be ignored? e.g.,
https://github.com/longhorn/longhorn-manager/blob/73b66414ce281a91658d2b120a1933dd84ad99cb/controller/engine_controller.go#L1102

If NO, I am fine with directly marking the replica as failed when there is a purge error.

PhanLe1010:
> Is there any purge error that can be ignored? 
@shuo-wu I think no error can be ignored in e.Status.PurgeStatus. The error `replica is already purging snapshots` already caught in the start snapshot purge command, and it is not recorded to `e.Status.PurgeStatus`, it is just recorded in the volume's event.

PhanLe1010:
There is one E2E test that failed. It is because of the test's logic and it is tracked at https://github.com/longhorn/longhorn/issues/2167

PhanLe1010:
This PR is ready to be review

shuo-wu (approved):

PR #817, opened 2021-01-13
Introduce Backing image

shuo-wu:
longhorn/longhorn#2006


controller/node_controller.go
```diff
@@ -165,7 +165,8 @@ func isSettingStorageMinimalAvailablePercentage(obj interface{}) bool {
 		}
 	}
 
-	return types.SettingName(setting.Name) == types.SettingNameStorageMinimalAvailablePercentage
+	return types.SettingName(setting.Name) == types.SettingNameStorageMinimalAvailablePercentage ||
+		types.SettingName(setting.Name) == types.SettingNameBackingImageCleanupWaitInterval
```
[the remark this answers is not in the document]
shuo-wu: Will rename this function.

datastore/longhorn.go
```diff
@@ -919,6 +919,163 @@ func (s *DataStore) ListEngineImages() (map[string]*longhorn.EngineImage, error)
...
+		return errors.Wrapf(err, "unable to remove finalizer for backing image %v", obj.Name)
+	}
+	return nil
+}
+
+func (s *DataStore) getBackingImageRO(name string) (*longhorn.BackingImage, error) {
+	return s.biLister.BackingImages(s.namespace).Get(name)
+}
+
+func (s *DataStore) getBackingImage(name string) (*longhorn.BackingImage, error) {
```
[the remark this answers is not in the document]
shuo-wu: Make sense

api/backingimage.go
```diff
@@ -0,0 +1,80 @@
...
+	if err != nil {
+		return errors.Wrapf(err, "unable to create backing image %v for URL %v", input.Name, input.ImageURL)
+	}
+	apiContext.Write(toBackingImageResource(bi, apiContext))
+	return nil
+}
+
+func (s *Server) BackingImageDelete(rw http.ResponseWriter, req *http.Request) error {
+	id := mux.Vars(req)["name"]
+	if err := s.m.DeleteBackingImage(id); err != nil {
```
shuo-wu, line 58: Check if there is volume using it

controller/replica_controller.go
```diff
@@ -346,6 +355,30 @@ func (rc *ReplicaController) CreateInstance(obj interface{}) (*types.InstancePro
 		return nil, fmt.Errorf("missing parameters for replica process creation: %v", r)
 	}
 
+	backingImagePath := ""
+	if r.Spec.BackingImage != "" {
+		bi, err := rc.ds.GetBackingImage(r.Spec.BackingImage)
+		if err != nil {
+			return nil, err
```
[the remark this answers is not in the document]
shuo-wu: If the replica process launching fails, which is the 1st step of volume attachment, Longhorn has no way to handle it and just leave the volume keeping state `detached`. The issue you find here is actually a sub-issue of this bug.
There is already an issue tracking this bug: https://github.com/longhorn/longhorn/issues/1330

shuo-wu:
Is there a better way to generate the download name? Currently, the download pod name is `fmt.Sprintf("%s-download-%s", biName, diskUUID)`, which may exceed 63-character limitation if the bi name is not a pretty short string.

shuo-wu:
Will add a commit to update `go.mod` after https://github.com/longhorn/backupstore/pull/60 is merged

shuo-wu:
@joshimoo Do we still need the dependency `k8s.io/utils/nsenter`? The vendor files are removed automatically when I update `longhorn/backupstore` and run `go mod tidy && go mod vendor`

controller/node_controller.go
```diff
@@ -887,3 +892,47 @@ func (nc *NodeController) createInstanceManager(node *longhorn.Node, imName, ima
...
+	}
+	waitInterval := time.Duration(settingValue) * time.Minute
+
+	for _, diskStatus := range node.Status.DiskStatus {
+		uuid := diskStatus.DiskUUID
+		backingImages, err := nc.ds.ListBackingImages()
+		if err != nil {
+			return err
+		}
+		for _, bi := range backingImages {
```
joshimoo, line 910: inverse this so you process the bi's once and have an inner loop for node.Status.DiskStatus per bi.
this way you only update the bi once, after removing all matching disk entries :)

joshimoo (commented):
Left some code suggestions, I will review the backing_image_controller tomorrow.

datastore/longhorn.go
```diff
@@ -1001,6 +1156,30 @@ func (s *DataStore) GetNode(name string) (*longhorn.Node, error) {
...
+						if _, exists := node.Spec.Disks[diskName]; exists {
+							return node, diskName, nil
+						}
+					}
+				}
+			}
+		}
+	}
+
+	return nil, "", fmt.Errorf("cannot find the corresponding ready node and disk with disk UUID %v", diskUUID)
```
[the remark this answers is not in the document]
shuo-wu: We can use function `NewNotFound(qualifiedResource schema.GroupResource, name string)` in package `apierrors` to generate the Kubernetes error. But there is no node name here and we cannot input the error reason for `NotFound` errors, I prefer to use this kind of error here.

engineapi/backups.go
```diff
@@ -22,14 +22,16 @@ type BackupTarget struct {
...
-	DataStored     string
-	Messages       map[backupstore.MessageType]string
-	Backups        map[string]interface{}
+	Name             string
+	Size             string
+	Created          string
+	LastBackupName   string
+	LastBackupAt     string
+	BackingImageName string
+	BackingImageURL  string
```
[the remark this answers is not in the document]
shuo-wu: All right... I just check the code and find that modifying this means modifying all backing image PRs including backupstore, longhorn-engine, engine integration test, and manager integration test...... If this is a must-have or a better-to-have, I will do then

controller/backing_image_controller.go
```diff
@@ -0,0 +1,621 @@
...
+		}
+		node, _, err := bic.ds.GetReadyDiskNode(diskUUID)
+		if err != nil {
+			if types.ErrorIsNotFound(err) {
+				backingImage.Status.DiskDownloadStateMap[diskUUID] = types.BackingImageDownloadStateFailed
+				continue
+			}
+			return err
+		}
+		if err := bic.createDownloadPod(downloadPodName, diskUUID, node, backingImage); err != nil {
```
joshimoo, line 342: Move the node retrieval into the createDownloadPod function and only pass a diskUUID to the function
Then you can pass through the notFound error from there and handle it below as part of the error handling for the createDownloadPod below

controller/backing_image_controller.go
```diff
@@ -0,0 +1,621 @@
...
+	priorityClass, err := bic.ds.GetSetting(types.SettingNamePriorityClass)
+	if err != nil {
+		return err
+	}
+
+	imagePullPolicy, err := bic.ds.GetSettingImagePullPolicy()
+	if err != nil {
+		return err
+	}
+
```
joshimoo, line 438: do the node retrieval here, see above comment for details.

controller/backing_image_controller.go
```diff
@@ -0,0 +1,621 @@
...
+	// The image is not important.
+	// As long as the image is a Linux distribution with wget installed, it can be used in the download pod.
+	// We choose the default engine image here so that users don't need to fetch on extra image in the air-gaped environment.
+	defaultEngineImage, err := bic.ds.GetSetting(types.SettingNameDefaultEngineImage)
+	if err != nil {
+		return err
+	}
+	if defaultEngineImage.Value == "" {
+		return fmt.Errorf("default engine image not set")
+	}
```
joshimoo, line 462:
```suggestion
	if err != nil || defaultEngineImage.Value == "" {
		return fmt.Errorf("failed to retrieve default engine image")
	}
```

controller/backing_image_controller.go
```diff
@@ -0,0 +1,621 @@
...
+					StartupProbe: &corev1.Probe{
+						Handler: corev1.Handler{
+							Exec: &corev1.ExecAction{
+								Command: []string{
+									"bash", "-c",
+									"ls " + filePathInContainer + " && ls " + downloadedNotificationFlag,
+								},
+							},
+						},
+						// Wait up to 1 day for the backing image downloading.
```
joshimoo, line 512: 1 day seems like a really long time to wait for the download?
Is it better to make this a shorter time, so we eventually mark the pod as failed which will mark the download as failed + cleanup so we retry?

controller/backing_image_controller.go
```diff
@@ -0,0 +1,621 @@
...
+						FailureThreshold:    17280,
+						PeriodSeconds:       5,
+						InitialDelaySeconds: 5,
+					},
+					// Choose ReadinessProbe rather than LivenessProbe due to this Kubernetes bug:
+					// https://github.com/kubernetes/kubernetes/issues/95140
```
joshimoo, line 518: nice catch :)

controller/backing_image_controller.go
```diff
@@ -0,0 +1,621 @@
...
+							types.BackingImageDownloadStateDownloaded, backingImage.Status.DiskDownloadStateMap[diskUUID])
+						backingImage.Status.DiskDownloadStateMap[diskUUID] = types.BackingImageDownloadStateDownloaded
+					}
+				} else {
+					if backingImage.Status.DiskDownloadStateMap[diskUUID] == types.BackingImageDownloadStateDownloaded {
+						log.WithFields(logrus.Fields{"diskUUID": diskUUID, "pod": downloadPodName}).Warnf("The download state becomes %v from %v since the download pod suddenly becomes unavailable",
+							types.BackingImageDownloadStateFailed, types.BackingImageDownloadStateDownloaded)
+						backingImage.Status.DiskDownloadStateMap[diskUUID] = types.BackingImageDownloadStateFailed
+					}
+				}
```
joshimoo, line 307:
```suggestion
				downloadComplete := backingImage.Status.DiskDownloadStateMap[diskUUID] == types.BackingImageDownloadStateDownloaded
				if !downloadComplete && isReady {
					log.WithFields(logrus.Fields{"diskUUID": diskUUID, "pod": downloadPodName}).Infof("The download state becomes %v from %v since the download pod is ready",
							types.BackingImageDownloadStateDownloaded, backingImage.Status.DiskDownloadStateMap[diskUUID])
					backingImage.Status.DiskDownloadStateMap[diskUUID] = types.BackingImageDownloadStateDownloaded
				} 
				
				if downloadComplete && !isReady {
					log.WithFields(logrus.Fields{"diskUUID": diskUUID, "pod": downloadPodName}).Warnf("The download state becomes %v from %v since the download pod suddenly becomes unavailable",
							types.BackingImageDownloadStateFailed, types.BackingImageDownloadStateDownloaded)
					backingImage.Status.DiskDownloadStateMap[diskUUID] = types.BackingImageDownloadStateFailed
				}
```

Add a downloadComplete flag, that will make the control flow logic simpler.
Since this functions control flow structure should be refactored further see above comment this will also reduce the nesting further.

joshimoo (approved):
LGTM in general, left some suggestions to simplify the code.
After incorporating these suggestions feel free to merge.

engineapi/backups.go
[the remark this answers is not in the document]
joshimoo: up to you if you think it's to much work for little gain, you can leave it, we can always refactor down the line as long as the external api is solid :)

controller/backing_image_controller.go
shuo-wu: Sure, I have updated it to 3 hours

controller/backing_image_controller.go
```diff
@@ -0,0 +1,605 @@
...
+		// and the disk on the node are ready
+		if downloadPod == nil {
+			if state, exists := backingImage.Status.DiskDownloadStateMap[diskUUID]; exists {
+				log.WithField("diskUUID", diskUUID).Warnf("The disk download state is %v but the download pod is nil, will directly create a new pod then",
+					state)
+			}
+			if err := bic.createDownloadPod(downloadPodName, diskUUID, backingImage); err != nil {
+				log.WithFields(logrus.Fields{"diskUUID": diskUUID, "pod": downloadPodName}).WithError(err).Errorf("The download state becomes %v due to the download pod creation failure",
+					types.BackingImageDownloadStateFailed)
+				backingImage.Status.DiskDownloadStateMap[diskUUID] = types.BackingImageDownloadStateFailed
```
joshimoo, line 305: This will not be updated since we are returning error.
Do we want to process the other disks for this backing image even if one fails?
If so you can just set the failed state here & requeue the backing image & continue

controller/backing_image_controller.go
```diff
@@ -0,0 +1,605 @@
...
+	}
+
+	return nil
+}
+
+func getDownloadPodName(biName, diskUUID string) string {
+	return fmt.Sprintf("%s-download-%s", biName, diskUUID[:8])
+}
+
+func (bic *BackingImageController) syncBackingImageWithDownloadPods(backingImage *longhorn.BackingImage) (err error) {
```
joshimoo, line 254: Might be useful to add a doc comment with all the state transitions like I do for the share-manager-controller, up to you though. example here: https://github.com/longhorn/longhorn-manager/blob/master/controller/share_manager_controller.go#L537-L543

controller/backing_image_controller.go
```diff
@@ -0,0 +1,605 @@
...
+					types.BackingImageDownloadStateDownloaded, backingImage.Status.DiskDownloadStateMap[diskUUID])
+				backingImage.Status.DiskDownloadStateMap[diskUUID] = types.BackingImageDownloadStateDownloaded
+			}
+
+			if downloadComplete && !isReady {
+				log.WithFields(logrus.Fields{"diskUUID": diskUUID, "pod": downloadPodName}).Warnf("The download state becomes %v from %v since the download pod suddenly becomes unavailable",
+					types.BackingImageDownloadStateFailed, types.BackingImageDownloadStateDownloaded)
+				backingImage.Status.DiskDownloadStateMap[diskUUID] = types.BackingImageDownloadStateFailed
+			}
+		case corev1.PodPending:
```
joshimoo, line 339: Consider moving `PodPending` before `PodRunning` in the switch statement.

controller/backing_image_controller.go
```diff
@@ -0,0 +1,605 @@
...
+		}
+
+		// Create/Recreate the download pod for a disk if the old pod is gone
+		// and the disk on the node are ready
+		if downloadPod == nil {
+			if state, exists := backingImage.Status.DiskDownloadStateMap[diskUUID]; exists {
+				log.WithField("diskUUID", diskUUID).Warnf("The disk download state is %v but the download pod is nil, will directly create a new pod then",
+					state)
+			}
+			if err := bic.createDownloadPod(downloadPodName, diskUUID, backingImage); err != nil {
```
joshimoo, line 302: consider returning the created pod and just assigning it to the download pod, then you can remove the continue below, since the pod processing will set it to downloading based on the `pod.pending`.

joshimoo (changes requested):
LGTM, the only thing that needs to be fixed is the missing `rStoreSynced` for the cache wait.
The rest are suggestions.

controller/backing_image_controller.go
```diff
@@ -0,0 +1,605 @@
...
+			// we don't mind others coming first
+			if apierrors.IsConflict(errors.Cause(err)) {
+				return nil
+			}
+			return err
+		}
+		log.Infof("Backing Image got new owner %v", bic.controllerID)
+	}
+
+	if backingImage.Status.OwnerID != bic.controllerID {
```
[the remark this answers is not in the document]
shuo-wu: OK. Then I think I can leave this part unchanged for now?

joshimoo (approved):
LGTM

PR #821, opened 2021-01-27
Support s3 pagination

jenting:
#### Proposed Changes
Bump longhorn/backupstore version to support s3 pagination.

#### Linked Issues
https://github.com/longhorn/longhorn/issues/1904

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -47,7 +47,7 @@ require (
 	github.com/imdario/mergo v0.3.8 // indirect
 	github.com/jinzhu/copier v0.0.0-20190924061706-b57f9002281a
 	github.com/kubernetes-csi/csi-lib-utils v0.6.1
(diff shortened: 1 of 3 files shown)


yasker (changes requested):
Where is the vendor change?

jenting:
> Where is the vendor change?

The longhorn-manager does not have dependency on longhorn/backup store, should we bump the longhorn-manager dependency to make it sure align with longhorn-engine?

shuo-wu:
I think Sheng means that there is no actual dependency file update in the directory `vendor/`. Hence, this PR may be redundant.

jenting:
Close it due to no dependency need to be updated.

yasker:
Sorry for the confusion @jenting @shuo-wu . I didn't realize there is no vendor change needed. In fact, it's still better to have the backupstore dependency updated, to ensure it's on par with the engine.

yasker (approved):

PR #823, opened 2021-01-27
Set minAvailable to 0 for instance-manager-e pods

drewhemm:
* Allows for compatibility with Cluster Autoscaler

Signed-off-by: Andrew Hemming <andrew@footprintmedia.net>

### DIFF
--- a/controller/instance_manager_controller.go
+++ b/controller/instance_manager_controller.go
@@ -627,6 +627,14 @@ func (imc *InstanceManagerController) createInstanceManagerPDB(im *longhorn.Inst
 }
 
 func (imc *InstanceManagerController) generateInstanceManagerPDBManifest(im *longhorn.InstanceManager) *policyv1beta1.PodDisruptionBudget {
+	var minAvailable int32
+	switch im.Spec.Type {
+	case types.InstanceManagerTypeEngine:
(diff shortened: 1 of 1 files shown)


yasker:
@drewhemm Sorry this PR doesn't make sense. It will break the Kubernetes drain logic by allowing Kubernetes to drain Longhorn instance manager pod with the engine running on it, which will result in the crash of workload pod who is using the volume.

zbialik:
@yasker do you have any recommendation for fixing [Issue 2203](https://github.com/longhorn/longhorn/issues/2203)?

This is a pretty big blocker for those of us wanting to use longhorn with a dynamic autoscaling cluster. We already plan on dedicating data nodes for our longhorn deployment, however that doesn't fix the issue of having these e-managers scheduled onto our scaling worker node group. 

The only solution/workaround is to taint our scaling worker node group so that the e-manager's can't run on these nodes. This isn't ideal as now we have to add tolerations to all our workloads we have scaling. 

If we can't fix the compatibility with cluster-autoscaler quickly, can we provide a mechanism to enforce node scheduling of the e-manger's using nodeAffinity/nodeSelector instead of a taint on all nodes we don't want them deployed to?

PR #824, opened 2021-01-27
Compatibility with Cluster Autoscaler

drewhemm:
* Add safe-to-evict annotation to instance-manager-e pods
* Set minAvailable to 0 for instance-manager-e pods

Signed-off-by: Andrew Hemming <andrew@footprintmedia.net>

### DIFF
--- a/controller/instance_manager_controller.go
+++ b/controller/instance_manager_controller.go
@@ -627,6 +627,14 @@ func (imc *InstanceManagerController) createInstanceManagerPDB(im *longhorn.Inst
 }
 
 func (imc *InstanceManagerController) generateInstanceManagerPDBManifest(im *longhorn.InstanceManager) *policyv1beta1.PodDisruptionBudget {
+	var minAvailable int32
+	switch im.Spec.Type {
+	case types.InstanceManagerTypeEngine:
+		minAvailable = 0
+	case types.InstanceManagerTypeReplica:
+		minAvailable = 1
+	}
+
 	return &policyv1beta1.PodDisruptionBudget{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      imc.getPDBName(im),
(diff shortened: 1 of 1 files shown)


drewhemm:
This PR is a more comprehensive alternative to https://github.com/longhorn/longhorn-manager/pull/823, adding full compatibility with CA, without requiring the use of a mutating webhook such as kube-annotate

munjalpatel (approved):
LGTM!

yasker:
Similar to https://github.com/longhorn/longhorn-manager/pull/823 . This PR will break the Longhorn support for Kubernetes drain logic by allowing Kubernetes to drain Longhorn instance manager pod with the engine running on it, which will result in the crash of workload pod that is using the volume.

drewhemm:
Sounds like the instance manager pod needs to exit gracefully only after any workloads using the volume(s) have exited. Is a check like that possible? 

If no workloads are consuming volumes (which will often be the case since the manager runs on every node and not every node will have workloads consuming LH volumes), then there should be nothing additional to do.

zbialik:
@yasker do you have any recommendation for fixing Issue 2203?

This is a pretty big blocker for those of us wanting to use longhorn with a dynamic autoscaling cluster. We already plan on dedicating data nodes for our longhorn deployment, however that doesn't fix the issue of having these e-managers scheduled onto our scaling worker node group.

The only solution/workaround is to taint our scaling worker node group so that the e-manager's can't run on these nodes. This isn't ideal as now we have to add tolerations to all our workloads we have scaling.

If we can't fix the compatibility with cluster-autoscaler quickly, can we provide a mechanism to enforce node scheduling of the e-manger's using nodeAffinity/nodeSelector instead of a taint on all nodes we don't want them deployed to?

chris93111:
Any update ?

PR #825, opened 2021-01-28
Fix CSI sanity check

c3y1huang:
Fix failures from [csi-sanity](https://github.com/kubernetes-csi/csi-test/tree/master/cmd/csi-sanity) tests.

* Add volume name autocorrection.
* Remove the unimplemented capability to bypass related tests.
* Correct misuse of the error codes.
* Validate resources to fix unexpected returns.
* Address to idempotency of volume life cycle.

Relate to https://github.com/longhorn/longhorn/issues/2076

### DIFF
--- a/csi/controller_server.go
+++ b/csi/controller_server.go
@@ -282,24 +282,33 @@ func (cs *ControllerServer) isNodeReady(nodeID string) bool {
 // ControllerPublishVolume will attach the volume to the specified node
 func (cs *ControllerServer) ControllerPublishVolume(ctx context.Context, req *csi.ControllerPublishVolumeRequest) (*csi.ControllerPublishVolumeResponse, error) {
 	logrus.Infof("ControllerServer ControllerPublishVolume req: %v", req)
+
+	if req.GetNodeId() == "" {
+		msg := "ControllerPublishVolume: missing node id in request"
+		logrus.Warn(msg)
+		return nil, status.Error(codes.InvalidArgument, msg)
+	}
+
+	volumeCapability := req.GetVolumeCapability()
+	if volumeCapability == nil {
+		msg := fmt.Sprint("ControllerPublishVolume: missing volume capability in request")
+		logrus.Warn(msg)
+		return nil, status.Error(codes.InvalidArgument, msg)
+	}
+
+	if !cs.isNodeReady(req.GetNodeId()) {
+		msg := fmt.Sprintf("ControllerPublishVolume: the volume %s cannot be attached to `NotReady` node %s",
+			req.GetVolumeId(), req.GetNodeId())
+		logrus.Warn(msg)
+		return nil, status.Error(codes.NotFound, msg)
+	}
+
 	existVol, err := cs.apiClient.Volume.ById(req.GetVolumeId())
 	if err != nil {
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 	if existVol == nil {
 		msg := fmt.Sprintf("ControllerPublishVolume: the volume %s not exists", req.GetVolumeId())
-		logrus.Warn(msg)
-		return nil, status.Error(codes.NotFound, msg)
-	}
-
-	volumeCapability := req.GetVolumeCapability()
-	if volumeCapability == nil {
-		return nil, status.Error(codes.InvalidArgument, "Volume capability not provided")
-	}
-
-	if !cs.isNodeReady(req.GetNodeId()) {
-		msg := fmt.Sprintf("ControllerPublishVolume: the volume %s cannot be attached to `NotReady` node %s",
-			req.GetVolumeId(), req.GetNodeId())
 		logrus.Warn(msg)
 		return nil, status.Error(codes.NotFound, msg)
 	}
--- a/csi/controller_server.go
+++ b/csi/controller_server.go
@@ -236,6 +236,10 @@ func (cs *ControllerServer) DeleteVolume(ctx context.Context, req *csi.DeleteVol
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 
+	if !cs.waitForVolumeState(req.GetVolumeId(), "deleted", isVolumeDeleted, false, true) {
+		return nil, status.Errorf(codes.Aborted, "Failed to delete volume %s", req.GetVolumeId())
+	}
+
 	return &csi.DeleteVolumeResponse{}, nil
 }
 
@@ -715,6 +719,10 @@ func (cs *ControllerServer) ControllerExpandVolume(ctx context.Context, req *csi
 		CapacityBytes:         req.CapacityRange.GetRequiredBytes(),
 		NodeExpansionRequired: false,
 	}, nil
+}
+
+func isVolumeDeleted(vol *longhornclient.Volume) bool {
+	return vol == nil
 }
 
 func isVolumeDetached(vol *longhornclient.Volume) bool {
--- a/csi/controller_server.go
+++ b/csi/controller_server.go
@@ -247,6 +247,17 @@ func (cs *ControllerServer) ControllerGetCapabilities(ctx context.Context, req *
 
 func (cs *ControllerServer) ValidateVolumeCapabilities(ctx context.Context, req *csi.ValidateVolumeCapabilitiesRequest) (*csi.ValidateVolumeCapabilitiesResponse, error) {
 	logrus.Infof("ControllerServer ValidateVolumeCapabilities req: %v", req)
+
+	existVol, err := cs.apiClient.Volume.ById(req.GetVolumeId())
+	if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+	if existVol == nil {
+		msg := fmt.Sprintf("ValidateVolumeCapabilities: the volume %s not exists", req.GetVolumeId())
+		logrus.Warn(msg)
+		return nil, status.Error(codes.NotFound, msg)
+	}
+
 	if err := cs.validateVolumeCapabilities(req.GetVolumeCapabilities()); err != nil {
 		return nil, err
 	}
--- a/csi/controller_server.go
+++ b/csi/controller_server.go
@@ -664,7 +664,7 @@ func (cs *ControllerServer) DeleteSnapshot(ctx context.Context, req *csi.DeleteS
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 
-	if backupVolume != nil {
+	if backupVolume != nil && backupVolume.Name != "" {
 		backupVolume, err = cs.apiClient.BackupVolume.ActionBackupDelete(backupVolume, &longhornclient.BackupInput{Name: backupName})
 		if err != nil {
 			return nil, status.Error(codes.Internal, err.Error())
--- a/csi/controller_server.go
+++ b/csi/controller_server.go
@@ -410,6 +410,12 @@ func (cs *ControllerServer) ControllerPublishVolume(ctx context.Context, req *cs
 func (cs *ControllerServer) ControllerUnpublishVolume(ctx context.Context, req *csi.ControllerUnpublishVolumeRequest) (*csi.ControllerUnpublishVolumeResponse, error) {
 	logrus.Infof("ControllerServer ControllerUnpublishVolume req: %v", req)
 
+	if req.GetVolumeId() == "" {
+		msg := "ControllerUnpublishVolume: missing volume id in request"
+		logrus.Warn(msg)
+		return nil, status.Error(codes.InvalidArgument, msg)
+	}
+
 	existVol, err := cs.apiClient.Volume.ById(req.GetVolumeId())
 	if err != nil {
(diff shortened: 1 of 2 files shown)


c3y1huang:
Found some tests that might be better off skipped. Below tests attempts to create snapshots after creating volume. Since the engine won't be running for detached volumes, it seems impossible to fix without making destructive changes to the current Longhorn design.
* should create volume from an existing source snapshot
* should return appropriate values
* should succeed when creating snapshot with maximum-length name
* should succeed when requesting to create a snapshot with already existing name and same source volume ID
* should fail when requesting to create a snapshot with already existing name and different source volume ID

c3y1huang:
Tests all passed other than the tests mentioned in https://github.com/longhorn/longhorn-manager/pull/825#issuecomment-771520991.
```
csi-sanity -csi.endpoint /var/lib/kubelet/obsoleted-longhorn-plugins/driver.longhorn.io/csi.sock -ginkgo.skip="should create volume from an existing source snapshot|should return appropriate values|should succeed when creating snapshot with maximum-length name|should succeed when requesting to create a snapshot with already existing name and same source volume ID|should fail when requesting to create a snapshot with already existing name and different source volume ID"
...
Ran 42 of 77 Specs in 134.364 seconds
SUCCESS! -- 42 Passed | 0 Failed | 1 Pending | 34 Skipped
```
The `1 Pending` is expected as it is [marked](https://github.com/kubernetes-csi/csi-test/blob/master/pkg/sanity/controller.go#L267) disable.

csi/controller_server.go
```diff
@@ -254,6 +254,18 @@ func (cs *ControllerServer) isNodeReady(nodeID string) bool {
 // ControllerPublishVolume will attach the volume to the specified node
 func (cs *ControllerServer) ControllerPublishVolume(ctx context.Context, req *csi.ControllerPublishVolumeRequest) (*csi.ControllerPublishVolumeResponse, error) {
 	logrus.Infof("ControllerServer ControllerPublishVolume req: %v", req)
+	if req.GetNodeId() == "" {
```
shuo-wu, line 288: Put this check in front of this part. Or directly update the Error Code of this error instead :
```
	if !cs.isNodeReady(req.GetNodeId()) {
		msg := fmt.Sprintf("ControllerPublishVolume: the volume %s cannot be attached to `NotReady` node %s",
			req.GetVolumeId(), req.GetNodeId())
		logrus.Warn(msg)
		return nil, status.Error(codes.NotFound, msg)
	}
```

csi/controller_server.go
```diff
@@ -646,6 +646,14 @@ func (cs *ControllerServer) DeleteSnapshot(ctx context.Context, req *csi.DeleteS
 	}
 
 	_, volumeName, backupName := decodeSnapshotID(req.SnapshotId)
+	snapshot, err := cs.apiClient.Snapshot.ById(backupName)
```
shuo-wu, line 653: 1. I think it should be `snapshot, err := cs.apiClient.Snapshot.ById(req.SnapshotId)`
2. There must be an error returned from Longhorn API layer if the snapshot doesn't exist:
```
func (m *VolumeManager) GetSnapshot(snapshotName, volumeName string) (*types.Snapshot, error) {
	if volumeName == "" || snapshotName == "" {
		return nil, fmt.Errorf("volume and snapshot name required")
	}
	engine, err := m.GetEngineClient(volumeName)
	if err != nil {
		return nil, err
	}
	snapshot, err := engine.SnapshotGet(snapshotName)
	if err != nil {
		return nil, err
	}
	if snapshot == nil {
		return nil, fmt.Errorf("cannot find snapshot '%s' for volume '%s'", snapshotName, volumeName)
	}
	return snapshot, nil
}
```
How does `if snapshot == nil` works?

api/model.go
```diff
@@ -932,6 +932,7 @@ func toVolumeResource(v *longhorn.Volume, ves []*longhorn.Engine, vrs []*longhor
 			actions["pvcCreate"] = struct{}{}
 			actions["updateDataLocality"] = struct{}{}
 			actions["updateAccessMode"] = struct{}{}
+			actions["snapshotList"] = struct{}{}
```
shuo-wu, line 935: Longhorn cannot list snapshots for a detached volume...

shuo-wu (changes requested):
1. Can we consider the missing/empty volume id as a sub-case of volume not found/nil existing volume? 

2. I don't understand why waiting for volume deletion complete could fix the attachment error mentioned in commit `Fix csi sanity check - idempotent`

csi/controller_server.go
c3y1huang: The issue with the approach to check for the node would fail other tests. 

To change the err code would fail this [example](https://github.com/kubernetes-csi/csi-test/blob/master/pkg/sanity/controller.go#L961) where is expecting `codes.NotFound`.

But without change the err code would fail other tests where expecting different error code. Most negative tests check the request with the minimal mocking of the request. For [example](https://github.com/kubernetes-csi/csi-test/blob/master/pkg/sanity/controller.go#L866), this one does not mock the NodeId. 

By putting the node check right to the front not only fail by the unexpected error code, and also bypasses the request that is testing for.

However, I could still move it forward, together with the other checks...

api/model.go
c3y1huang: Arrgh..this is partial fix of the 5 skipped tests... let me remove it.

csi/controller_server.go
c3y1huang: Sounds good to me, fixed.

c3y1huang:
> 1. Can we consider the missing/empty volume id as a sub-case of volume not found/nil existing volume?

https://github.com/longhorn/longhorn-manager/pull/825#discussion_r569997490

> 2. I don't understand why waiting for volume deletion complete could fix the attachment error mentioned in commit `Fix csi sanity check - idempotent`

The idempotent tests attempt to run the volume lifecycle (Create-Publish-Unpublish-Delete) repeatedly. Since volume could still in deleting from the previous round of the test, this fails test with unexpected results.
```
Action [attach] not available

Attaching volume sanity-controller-publish-2b264483-80559 on node ip-172-30-0-21 failed
```

c3y1huang:
@shuo-wu thanks for the review, I've addressed all comments.

csi/controller_server.go
```diff
@@ -97,19 +98,33 @@ func (cs *ControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
...
-	// check for already existing volume name
-	// ID and name are same in longhorn API
-	existVol, err := cs.apiClient.Volume.ById(req.GetName())
+	// Check for already existing volume name ID and name are same in longhorn API
+	// Ensure auto-corrected volume name not missed:
+	// - Convert to lower case of the definition of a label in DNS (RFC 1123)
+	// - Trim volume name of max name length limitation
+	volName := strings.ToLower(req.GetName())
+	if len(volName) > datastore.NameMaximumLength {
+		volName = strings.TrimRight(volName[:datastore.NameMaximumLength], "-")
```
joshimoo, line 107: is it possible that 2 volumes with overlength names end up with the same truncated name?
If so then it seams that it would be possible that 2 pvc (full name) could both think that they are the owner of a single longhorn volume (truncated name)

csi/controller_server.go
```diff
@@ -208,6 +223,10 @@ func (cs *ControllerServer) DeleteVolume(ctx context.Context, req *csi.DeleteVol
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 
+	if !cs.waitForVolumeState(req.GetVolumeId(), "deleted", isVolumeDeleted, false, true) {
```
joshimoo, line 226: good idea waiting for the volume deletion before acknowledging the deletion.

datastore/longhorn.go
```diff
@@ -242,14 +242,14 @@ func checkVolume(v *longhorn.Volume) error {
 	if v.Name == "" || size == 0 || v.Spec.NumberOfReplicas == 0 {
 		return fmt.Errorf("BUG: missing required field %+v", v)
 	}
+	if len(v.Name) > NameMaximumLength {
+		logrus.Warnf("Volume name is too long %v, auto-correct to %v characters", v.Name, NameMaximumLength)
+		v.Name = strings.TrimRight(v.Name[:NameMaximumLength], "-")
+	}
```
joshimoo, line 248: same issue here, feels like this could end up with name collisions?

joshimoo (commented):
Left a comment about the name truncation, the additional validation checks seem fine.

csi/controller_server.go
```diff
@@ -604,6 +649,12 @@ func (cs *ControllerServer) DeleteSnapshot(ctx context.Context, req *csi.DeleteS
 		return nil, status.Error(codes.InvalidArgument, "Snapshot ID must be provided")
 	}
 
+	snapshot, _ := cs.apiClient.Snapshot.ById(req.SnapshotId)
+	if snapshot == nil {
+		logrus.Debugf("Snapshot %v does not exist", req.SnapshotId)
+		return &csi.DeleteSnapshotResponse{}, nil
+	}
+
```
joshimoo, line 657: This is incorrect, have a look at the decode snapshotID function.
A csi snapshot refers to a longhorn backup.

joshimoo: Check the ActionBackupDelete below and see if it returns a notFound error code, then you can evaluate the error below.

joshimoo (changes requested):
A csi snapshot refers to a longhorn backup, so you need to check for the existence of the backup and not the existence of a snapshot for the deleteSnapshot call.

csi/controller_server.go
c3y1huang: Great catch!
I've committed the fix in https://github.com/longhorn/longhorn-manager/pull/825/commits/009e20dd4e5a97671e77b2533a8d7a3a18c07ed6, this will replace the suffix of the new name with the first 8 characters of its full name checksum.

datastore/longhorn.go
c3y1huang: See https://github.com/longhorn/longhorn-manager/pull/825#discussion_r570705312.

csi/controller_server.go
c3y1huang: Thanks for the hint! I think we could check the existence of backup volume name to achieve the same result. Fix the original commit(https://github.com/longhorn/longhorn-manager/pull/825/commits/1efd1ab258a7d6dd3692a85160ba883c26762599).

c3y1huang:
@joshimoo thanks for the review, I've addressed all comments.

csi/controller_server.go
shuo-wu: OK, make sense.

Then moving this check together with `if !cs.isNodeReady(req.GetNodeId())` may be enough.

c3y1huang: I've moved this forward and it goes before the capability check. This is because the node id [test](https://github.com/kubernetes-csi/csi-test/blob/master/pkg/sanity/controller.go#L870-L873) doesn't mock capability in the request. Though they result in the same error code, they are actually in a certain sequence to ensure different tests go through the correct checks.

shuo-wu: All right, then the current implementation is good enough.

util/util.go
```diff
@@ -357,6 +357,15 @@ func SplitStringToMap(str, separator string) map[string]struct{} {
 	return ret
 }
 
+// AutoCorrectName trims the overlength name, and also replaces the name suffix
+// with 8 char from its checksum to ensure uniqueness.
+func AutoCorrectName(name string, maxLength int) string {
+	checksum := GetStringChecksum(name)
+	suffix := "-" + checksum[:8]
+	newName := strings.TrimRight(name[:maxLength-len(suffix)], "-")
```
shuo-wu, line 365: `newNamePrefix`

shuo-wu (dismissed):
In general, LGTM

util/util.go
c3y1huang: Fixed.

shuo-wu (approved):

joshimoo (changes requested):
Can you stash the changes for the volume name length limitation and open a separate pr for them after this is merged?

c3y1huang:
> Can you stash the changes for the volume name length limitation and open a separate pr for them after this is merged?

@joshimoo I've dropped the duplicated volume name fix to pr later, and kept https://github.com/longhorn/longhorn-manager/pull/825/commits/f9e864167330e009cf45c407a42936137d7c5b7c to so the other tests can still pass. Could you PTAL?

shuo-wu:
It's better to move all volume name related modifications (commit `Fix csi sanity check - volume name` and `Fix csi sanity check - CreateVolume`) to a separate PR.

c3y1huang:
> It's better to move all volume name related modifications (commit `Fix csi sanity check - volume name` and `Fix csi sanity check - CreateVolume`) to a separate PR.

Moved to https://github.com/longhorn/longhorn-manager/pull/832.

shuo-wu (approved):

joshimoo (approved):
LGTM thanks for moving out the Name changes into a separate issue.

PR #826, opened 2021-02-03
Setup environment variables to sub-process only

jenting:
#### Proposed Changes
Pass S3 specific environment variables to sub-process only.
This could prevent the sub-process environment variables affect the parent process.

For example: if the backup target S3 need to go through proxy server, then pass the additional envs to sub-process by
```go
cmd := exec.Command("backup", "ls", args...)
cmd.Env = append(os.Environ(), "HTTP_PROXY=<http-proxy>", "HTTPS_PROXY=<https-proxy>", "NO_PROXY=<no-proxy>", ...)
```

#### Linked Issues
https://github.com/longhorn/longhorn/issues/2198

### DIFF
--- a/engineapi/backups.go
+++ b/engineapi/backups.go
@@ -3,6 +3,7 @@ package engineapi
 import (
 	"encoding/json"
 	"fmt"
+	"os"
 	"path/filepath"
 	"strings"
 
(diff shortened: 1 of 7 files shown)


engineapi/backups.go
```diff
@@ -40,24 +42,46 @@ func NewBackupTarget(backupTarget, engineImage string, credential map[string]str
 	}
 }
 
-func (b *BackupTarget) LonghornEngineBinary() string {
-	return filepath.Join(types.GetEngineBinaryDirectoryOnHostForImage(b.Image), "longhorn")
+// getBackupCredentialEnv returns the environment variables as KEY=VALUE in string slice
+func getBackupCredentialEnv(backupTarget string, credential map[string]string) ([]string, error) {
```
shuo-wu, line 46: Put this modification in the 2nd commit. Or directly coalesce 2 commits. It's confusing to add this function in this commit.

shuo-wu (changes requested):

engineapi/backups.go
jenting: I removed the enhancement code out in this PR to make it easier to review.

shuo-wu (approved):

PR #828, opened 2021-02-09
Enhanced instance manager CPU reservation

shuo-wu:
longhorn/longhorn#2207

### DIFF
--- a/controller/controller_manager.go
+++ b/controller/controller_manager.go
@@ -2,7 +2,9 @@ package controller
 
 import (
 	"fmt"
+	"math"
 	"os"
+	"strconv"
 	"time"
 
 	"github.com/pkg/errors"
@@ -126,8 +128,8 @@ func StartControllers(logger logrus.FieldLogger, stopCh chan struct{}, controlle
 		volumeInformer, engineInformer, replicaInformer,
 		settingInformer, engineImageInformer, nodeInformer)
 	sc := NewSettingController(logger, ds, scheme,
-		settingInformer,
-		kubeClient, version)
+		settingInformer, nodeInformer,
+		kubeClient, namespace, version)
 	imc := NewInstanceManagerController(logger, ds, scheme,
 		imInformer, podInformer, kubeNodeInformer, kubeClient, namespace, controllerID, serviceAccount)
 	smc := NewShareManagerController(logger, ds, scheme,
@@ -174,12 +176,8 @@ func StartControllers(logger logrus.FieldLogger, stopCh chan struct{}, controlle
 	return ds, ws, nil
 }
 
-func GetGuaranteedResourceRequirement(ds *datastore.DataStore) (*corev1.ResourceRequirements, error) {
-	guaranteedCPU, err := ds.GetSetting(types.SettingNameGuaranteedEngineCPU)
-	if err != nil {
-		return nil, err
-	}
-	quantity, err := resource.ParseQuantity(guaranteedCPU.Value)
+func ParseResourceRequirement(val string) (*corev1.ResourceRequirements, error) {
+	quantity, err := resource.ParseQuantity(val)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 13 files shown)


controller/controller_manager.go
```diff
@@ -188,6 +185,60 @@ func GetGuaranteedResourceRequirement(ds *datastore.DataStore) (*corev1.Resource
...
+		return nil, err
+	}
+	kubeNode, err := ds.GetKubernetesNode(im.Spec.NodeID)
+	if err != nil {
+		return nil, err
+	}
+	switch im.Spec.Type {
+	case types.InstanceManagerTypeEngine:
+		if lhNode.Spec.EngineManagerCPURequest > 0 {
+			return GetCPUResourceRequirement(fmt.Sprintf("%.3f", float32(lhNode.Spec.EngineManagerCPURequest)/1000.0))
```
joshimoo, line 213: instead of throwing in a float, just pass a milli string, ParseQuantity will handle that correctly.

joshimoo (changes requested):
General looks good to me, but the format and type conversions should be made much simpler.
Left some suggestions :)

yasker:
I noticed that we need to use float numbers here. I am all for using an integer instead since it makes things easier. There isn't a big enough difference between 12% vs 12.5% worth our extra effort to deal with floating numbers IMO.

shuo-wu:
OK, I will update to integer instead.

controller/controller_manager.go
```diff
@@ -193,6 +191,52 @@ func GetGuaranteedResourceRequirement(ds *datastore.DataStore) (*corev1.Resource
...
+		}
+		rmCPUSetting, err := ds.GetSetting(types.SettingNameGuaranteedReplicaManagerCPU)
+		if err != nil {
+			return nil, err
+		}
+		rmCPUPercentage, err := strconv.ParseFloat(rmCPUSetting.Value, 64)
+		if err != nil {
+			return nil, err
+		}
+		return GetCPUResourceRequirement(fmt.Sprintf("%.0fm", math.Round(float64(kubeNode.Status.Allocatable.Cpu().MilliValue())*rmCPUPercentage/100.0)))
```
joshimoo, line 234:
```suggestion
		cpuRequest := lhNode.Spec.ReplicaManagerCPURequest
		if cpuRequest == 0 {
			rmCPUSetting, err := ds.GetSetting(types.SettingNameGuaranteedReplicaManagerCPU)
			if err != nil {
				return nil, err
			}
			rmCPUPercentage, err := strconv.ParseFloat(rmCPUSetting.Value, 64)
			if err != nil {
				return nil, err
			}
			cpuRequest = int(math.Round(float64(kubeNode.Status.Allocatable.Cpu().MilliValue())*rmCPUPercentage/100.0))))
		}
		return GetCPUResourceRequirement(fmt.Sprintf("%dm", cpuRequest))
```

joshimoo: Both are fine, but prefer to do the fmt in integer form.

joshimoo (approved):
I don't particularly like that we are always evaluating the two independent settings together, I think it should be fine to treat them independently and not limit based on the sum of both together.

That should make the validation logic simpler and allow for the mentioned use case with dedicated storage nodes.

Other than that looks good to me.
@yasker @shuo-wu what do you think about treating them independently?

datastore/longhorn.go
```diff
@@ -134,6 +134,25 @@ func (s *DataStore) ValidateSetting(name, value string) (err error) {
 			}
 		}
 	}
+
+	guaranteedEngineManagerCPU, err := s.GetSetting(types.SettingNameGuaranteedEngineManagerCPU)
```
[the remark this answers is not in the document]
shuo-wu: I will put this check into the `switch`.

The reason for check the sum is that:
    1. We want to guarantee that the new instance manager pods can be launched when the image is updated. Hence the sum, in theory, shouldn't be greater than 50% and we will limit it to 40% for other workload upgrade. 
    2. But at the same time we want to allow users set engine/replica manager CPU request flexibly. e.g., if some node are storage nodes, users can set replica cpu to 40% with engine cpu 0%.

joshimoo (approved):
LGTM

PR #822, opened 2021-01-27
live migration feature

joshimoo:
The live migration function works like this:

- n1 | vm1 has the volume attached (v.spec.nodeID = n1)
- n2 | vm2 requests attachment (v.spec.migrationNodeID = n2) [migrationStart]
- volume-controller brings up new engine on n2, with inactive matching replicas (same as live engine upgrade)
- csi driver polls for existence of second engine on n2 before acknowledging attach

The following detach decides whether a migration is successful or a migration rollback on the longhorn side.
- n1 | vm1 requests detach of n1 (v.spec.nodeID = n2, v.spec.migrationNodeID = "") [migrationComplete]
- n2 | vm2 requests detach of n2 (v.spec.NodeID = n1, v.spec.migrationNodeID = "") [migrationRollback]

The volume controller then cleans up the second engine and switches the active replicas to be the current engine ones.

longhorn/longhorn#2127

PR #829, opened 2021-02-12
Refactor volume attach/detach to fix rwx expansion

joshimoo:
The problem is that the expansion call disables the frontend and never enables it again which prohibits the share-manager from controlling the volume. We rely on the attach call to reenable the frontend, this pr reworks the csi + api interaction so that we call attach for each volume type and let the volume-manager do the appropriate stuff :)

depends on this pr https://github.com/longhorn/longhorn-manager/pull/822 for the live migration feature, since there was a lot of overlap in the csi/attach/detach changes.

For the review only the last 2 commits are relevant for the rwx expansion support.
longhorn/longhorn#2181

### DIFF
--- a/csi/controller_server.go
+++ b/csi/controller_server.go
@@ -321,8 +321,9 @@ func (cs *ControllerServer) publishMigratableVolume(req *csi.ControllerPublishVo
 		}
 	}
 
+	checkVolumeAttached := func(vol *longhornclient.Volume) bool { return isVolumeAvailableOn(vol, req.NodeId) }
 	return cs.publishVolume(volume, req.NodeId, func() error {
-		if !cs.waitForVolumeState(req.GetVolumeId(), "volume migration available", isVolumeMigrationAvailable, false, false) {
+		if !cs.waitForVolumeState(req.GetVolumeId(), "volume available on", checkVolumeAttached, false, false) {
 			return status.Errorf(codes.DeadlineExceeded, "Attaching volume %s on node %s failed", req.GetVolumeId(), req.GetNodeId())
 		}
 		return nil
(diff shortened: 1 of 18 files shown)


manager/volume.go
```diff
@@ -607,9 +607,13 @@ func (m *VolumeManager) Expand(volumeName string, size int64) (v *longhorn.Volum
 		return nil, fmt.Errorf("cannot expand volume before replica scheduling success")
 	}
 
+	// volume shrinking is not supported the volume is already bigger then the requested size
```
shuo-wu, line 610: Do not remove the check here since users can use UI to expand a volume, which won't be prevented by Kubernetes.

shuo-wu (changes requested):
Since Kubernetes will prevent volume shrinking, it's nice to remove the error returning from the csi plugin. But I still don't know why the last commit could fix the rwx expansion issue. 

To expand a Longhorn volume, we need to guarantee there is no one using it and the frontend should be disabled. For a rwx volume, IIRC, the frontend `blockdev` is still enabled, right?

joshimoo:
@shuo-wu the volume expansion is fixed by having the csi driver call attach/detach for each volume type including the shared-volumes. This will now reset the left over disabled Frontend from the expansion call. See this commit 
https://github.com/longhorn/longhorn-manager/pull/829/commits/76531b74d8633b3102238e16c3e875c6478a88d2

manager/volume.go
joshimoo: We don't support  shrinking, so if the user requests a smaller size then the volume already has, we can return success.
Since the wanted capacity is already covered by the available capacity.

joshimoo: This also has the added benefit of making us compatible with the kubernetes expectation for volume expansion.

csi/controller_server.go
```diff
@@ -279,20 +279,40 @@ func (cs *ControllerServer) isNodeReady(nodeID string) bool {
...
+	input := &longhornclient.AttachInput{
+		HostId:          nodeID,
+		DisableFrontend: false,
 	}
 
-	if !cs.waitForVolumeState(volume.Name, "share available", isVolumeShareAvailable, false, false) {
-		return nil, status.Errorf(codes.DeadlineExceeded, "Failed to wait for volume %v share available", req.GetVolumeId())
+	if _, err := cs.apiClient.Volume.ActionAttach(volume, input); err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
 	}
```
shuo-wu: This part is the most important part of this PR,  right? Previously Longhorn CSI plugin will do nothing but just relies on the share manager controller to automatically attach RWX volumes as long as there are workloads using it, then the field `DisableFrontend` won't be reset.

manager/volume.go
shuo-wu: I don't think it's a good idea...
1. If users try to shrink a volume, Longhorn will tell them it succeeds but with the size unchanged, which will confuse or mislead users.
2. On the other hand, the reason for volume shrinking may be that users want to save/reserve some space for other applications/places. But actually Longhorn won't release any spaces regardless of returning success.
3. Since Kubernetes will prohibit size shrinking, to make Longhorn be consistent with Kubernetes, we should error out as well. Here is the error message.
```
# persistentvolumeclaims "longhorn-simple-pvc" was not valid:
# * spec.resources.requests.storage: Forbidden: field can not be less than previous value
```

csi/controller_server.go
joshimoo: Correct we need to call attach for each volume type, so we can reset the disableFrontend flag via the longhorn api -> manager. There is some new logic in the volume manager to deal with this correctly.

manager/volume.go
shuo-wu: IMO, the size in Longhorn is always the exposed block device's actual size, rather than a request. A PVC with a small size request could use a large PV, but the PV size is always accurate, right?

joshimoo:
Closed in favor of https://github.com/longhorn/longhorn-manager/pull/844
Feedback adressed in that pr.

PR #830, opened 2021-02-18
fix nil pointer cause longhorn-manager pods crushloopbackoff (#2208)

cclhsu:
#### Proposed Changes ####

fix nil pointer cause longhorn-manager pods crushloopbackoff to resolve ticket '[BUG] All longhorn-manager pods crushloopbackoff' (#2208)

#### Types of Changes ####

Fixed Bug.

#### Verification ####

It seems that during creating, diskStatus is nil. Need to check if diskstatus existed.

time="2021-02-18T08:41:05Z" level=info msg=disk-1
time="2021-02-18T08:41:06Z" level=info msg=disk-1
time="2021-02-18T08:41:06Z" level=info msg="Diskname disk-1 has status %!b(bool=true)"

NOTE: this was sample print during implementation/verification.

#### Linked Issues ####

Refs: https://github.com/longhorn/longhorn/issues/2208

#### Further Comments ####

None

Signed-off-by: cclhsu <clark.hsu@suse.com>

### DIFF
--- a/controller/replica_controller.go
+++ b/controller/replica_controller.go
@@ -574,18 +574,10 @@ func (rc *ReplicaController) enqueueNodeChange(obj interface{}) {
 		}
 	}
 
-	// If Node eviction, add all the disks to evictionDisks.
-	// Otherwise add request eviction disk separately.
-	var evictionDisks []string
-	for diskName, diskSpec := range node.Spec.Disks {
-		if node.Spec.EvictionRequested || diskSpec.EvictionRequested {
-			evictionDisks = append(evictionDisks, diskName)
-		}
-	}
-
 	// Add eviction requested replicas to the workqueue
-	for _, diskName := range evictionDisks {
-		if diskStatus, existed := node.Status.DiskStatus[diskName]; existed {
+	for diskName, diskSpec := range node.Spec.Disks {
+		evictionRequested := node.Spec.EvictionRequested || diskSpec.EvictionRequested
+		if diskStatus, existed := node.Status.DiskStatus[diskName]; existed && evictionRequested {
 			for replicaName := range diskStatus.ScheduledReplica {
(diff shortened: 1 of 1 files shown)


controller/replica_controller.go
```diff
@@ -585,12 +585,14 @@ func (rc *ReplicaController) enqueueNodeChange(obj interface{}) {
...
-			replica, err := rc.ds.GetReplica(replicaName)
-			if err != nil {
-				return
+		if diskStatus, existed := node.Status.DiskStatus[diskName]; existed {
+			for replicaName := range diskStatus.ScheduledReplica {
+				replica, err := rc.ds.GetReplica(replicaName)
+				if err != nil {
+					return
+				}
+				rc.enqueueReplica(replica)
```
joshimoo, line 594:
```suggestion
				if replica, err := rc.ds.GetReplica(replicaName); err == nil {
					rc.enqueueReplica(replica)
				}
```
If one replica cannot be retrieved from the datastore, for example because it has been deleted in the mean time.
We still want to enqueue the other replicas so they can run a reconcile loop.

joshimoo: You can also merge the two loops and get rid of the evictionDisk slice.
```
evictionRequested := node.Spec.EvictionRequested || diskSpec.EvictionRequested
if diskStatus, existed := node.Status.DiskStatus[diskName]; existed && evictionRequested {
	for replicaName := range diskStatus.ScheduledReplica {
		if replica, err := rc.ds.GetReplica(replicaName); err == nil {
			rc.enqueueReplica(replica)
		}
	}
}

}

joshimoo (changes requested):
Since this is a enqueu dependent components on updates of the observed component (node) type handler. We still want to enqueue all the other replicas even if one of them has been already deleted for example.

joshimoo (approved):
LGTM

PR #832, opened 2021-02-23
Fix CSI sanity check - Create Volume

c3y1huang:
Part of and relate to https://github.com/longhorn/longhorn-manager/pull/825#issuecomment-783294412.

- Add volume name autocorrection.
- Correct misuse of the error codes.

Fix https://github.com/longhorn/longhorn/issues/2270

### DIFF
--- a/manager/volume.go
+++ b/manager/volume.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"reflect"
 	"sort"
+	"strings"
 
 	"github.com/pkg/errors"
 	"github.com/robfig/cron"
@@ -258,7 +259,7 @@ func (m *VolumeManager) Create(name string, spec *types.VolumeSpec) (v *longhorn
 
 	v = &longhorn.Volume{
 		ObjectMeta: metav1.ObjectMeta{
-			Name: name,
+			Name: strings.ToLower(name),
 		},
 		Spec: types.VolumeSpec{
 			Size:                    size,
--- a/manager/volume.go
+++ b/manager/volume.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"reflect"
 	"sort"
-	"strings"
 
 	"github.com/pkg/errors"
 	"github.com/robfig/cron"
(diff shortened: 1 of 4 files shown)


manager/volume.go
```diff
@@ -258,7 +259,7 @@ func (m *VolumeManager) Create(name string, spec *types.VolumeSpec) (v *longhorn
 
 	v = &longhorn.Volume{
 		ObjectMeta: metav1.ObjectMeta{
-			Name: name,
+			Name: strings.ToLower(name),
```
c3y1huang, line 262: At existing VolumeCreate, Longhorn will validate if the volume name is compliant with DNS (RFC 1123). So I assume there shouldn't be any existing volume named out of the boundary.

shuo-wu (changes requested):
1. It's better to move `strings.ToLower` into function `AutoCorrectName`
2. Explicitly call `AutoCorrectName` before invoking the datastore function `VolumeCreate`. Function `checkVolume` should be responsible for validating the name only.

joshimoo:
I agree with Shuo's comment, autocorrect should be moved at the beginning of the `func (m *VolumeManager) Create` function then have check only validate the name.

c3y1huang:
> 1. It's better to move `strings.ToLower` into function `AutoCorrectName`
> 2. Explicitly call `AutoCorrectName` before invoking the datastore function `VolumeCreate`. Function `checkVolume` should be responsible for validating the name only.

> I agree with Shuo's comment, autocorrect should be moved at the beginning of the `func (m *VolumeManager) Create` function then have check only validate the name. Other than that LGTM, feel free to merge after addressing the feedback.

@shuo-wu , @joshimoo I've addressed the comments. PTAL.

util/util.go
```diff
@@ -347,6 +347,20 @@ func SplitStringToMap(str, separator string) map[string]struct{} {
 	return ret
 }
 
+// AutoCorrectName converts name to lowercase, and
+// correct overlength name by replaces the name suffix
+// with 8 char from its checksum to ensure uniqueness.
+func AutoCorrectName(name string, maxLength int) string {
```
shuo-wu, line 353: Add a log if the name is different after the auto correction

shuo-wu (changes requested):
Need to rename the 3rd commit. Others LGTM

util/util.go
c3y1huang: Addressed.

shuo-wu (approved):

PR #836, opened 2021-03-09
Add list backup volume error handling for NFS

c3y1huang:
When the volume directory does not exist on NFS backstore should return nil instead of error.

https://github.com/longhorn/longhorn/issues/2312


shuo-wu (commented):
Can we modify function `List` in repo `backupstore` instead? IMO, this error can be handled by the binary itself.

c3y1huang:
@shuo-wu thank you for the hint. PTAL at https://github.com/longhorn/backupstore/pull/66.

c3y1huang:
Closing in favor or https://github.com/longhorn/backupstore/pull/66.

PR #840, opened 2021-03-10
fix iscsi install manifest to create config file initiatorname iscsi

cclhsu:
#### Types of Changes ####

fix iscsi install manifest to create config file initiatorname iscsi to resolve ticket '[BUG] Configuration file  does not exist iscsi using ' (#2319)

- generate `/etc/iscsi/initiatorname.iscsi`
- fix `yum makecache fast` invalid in Centos 8.x

#### Verification ####

check if `/etc/iscsi/initiatorname.iscsi` created 

#### Linked Issues ####

Refs: # https://github.com/longhorn/longhorn/issues/2319

#### Further Comments ####

Verified in Centos 7.8 and Centos 8.2

### DIFF
--- a/deploy/prerequisite/longhorn-iscsi-installation.yaml
+++ b/deploy/prerequisite/longhorn-iscsi-installation.yaml
@@ -5,7 +5,7 @@ metadata:
   labels:
     app: longhorn-iscsi-installation
   annotations:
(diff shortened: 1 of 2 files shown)


khushboo-rancher:
Can we remove the `makecache fast` directly or should we have `if else` condition to handle both versions.

cclhsu:
> Can we remove the `makecache fast` directly or should we have `if else` condition to handle both versions.

I verified in Centos 7.8 and 8.2, remove `fast` directly work for both version.

khushboo-rancher (approved):
LGTM

innobead (approved):

PR #841, opened 2021-03-11
Support AWS IAM Role

jenting:
#### Proposed Changes
- Create a new secret controller, the purpose is to update the annotation of the `iam.amazonaws.com/role` to longhorn manager and replica instance manager Pod.
- If the user performs S3 list/backup/restore/delete, the longhorn-manager checks:
  - AWS_IAM_ROLE_ARN :heavy_check_mark: & AWS_ACCESS_KEY_ID/AWS_SECRET_ACCESS_KEY :heavy_check_mark: -> :ok:
  - AWS_IAM_ROLE_ARN :heavy_check_mark: & AWS_ACCESS_KEY_ID/AWS_SECRET_ACCESS_KEY :x: -> :ok:
  - AWS_IAM_ROLE_ARN :x: & AWS_ACCESS_KEY_ID/AWS_SECRET_ACCESS_KEY :heavy_check_mark: -> :ok:
  - AWS_IAM_ROLE_ARN :x: & AWS_ACCESS_KEY_ID/AWS_SECRET_ACCESS_KEY :x: -> :x:
- Bump longhorn/backupstore version

#### Linked Issues
https://github.com/longhorn/longhorn/issues/1526

### DIFF
--- a/datastore/longhorn.go
+++ b/datastore/longhorn.go
@@ -17,10 +17,9 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/validation"
 
+	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta1"
 	"github.com/longhorn/longhorn-manager/types"
 	"github.com/longhorn/longhorn-manager/util"
-
-	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta1"
 )
 
 const (
@@ -233,6 +232,52 @@ func (s *DataStore) ListSettings() (map[types.SettingName]*longhorn.Setting, err
 	return itemMap, nil
 }
 
+// annotateAWSIAMRole ensures that the running pods of the manager as well as the replica instance managers.
+// have the correct AWS IAM role assigned to them based on the passed `awsIAMRole`
+func (s *DataStore) annotateAWSIAMRole(awsIAMRole string) error {
+	selector, err := s.getManagerSelector()
+	if err != nil {
+		return err
+	}
+	managerPods, err := s.pLister.Pods(s.namespace).List(selector)
+	if err != nil {
+		return err
+	}
+
+	selector, err = metav1.LabelSelectorAsSelector(&metav1.LabelSelector{
+		MatchLabels: types.GetInstanceManagerLabels("", "", types.InstanceManagerTypeReplica),
+	})
+	if err != nil {
+		return err
+	}
+	imPods, err := s.pLister.Pods(s.namespace).List(selector)
+	if err != nil {
+		return err
+	}
+
+	var pods []*corev1.Pod
+	pods = append(pods, managerPods...)
+	pods = append(pods, imPods...)
+
+	for _, pod := range pods {
+		val, exist := pod.Annotations[types.AWSIAMRroleAnnotation]
+		updateAnnotation := awsIAMRole != "" && awsIAMRole != val
+		deleteAnnotation := awsIAMRole == "" && exist
+		if updateAnnotation {
+			pod.Annotations[types.AWSIAMRroleAnnotation] = awsIAMRole
+		} else if deleteAnnotation {
+			delete(pod.Annotations, types.AWSIAMRroleAnnotation)
+		} else {
+			continue
+		}
+
+		if _, err = s.kubeClient.CoreV1().Pods(s.namespace).Update(pod); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
 // GetCredentialFromSecret gets the Secret of the given name and namespace
 // Returns a new credential object or error
 func (s *DataStore) GetCredentialFromSecret(secretName string) (map[string]string, error) {
@@ -242,6 +287,11 @@ func (s *DataStore) GetCredentialFromSecret(secretName string) (map[string]strin
 	}
 	credentialSecret := make(map[string]string)
 	if secret.Data != nil {
+		if err := s.annotateAWSIAMRole(string(secret.Data[types.AWSIAMRole])); err != nil {
+			return nil, err
+		}
+
+		credentialSecret[types.AWSIAMRole] = string(secret.Data[types.AWSIAMRole])
 		credentialSecret[types.AWSAccessKey] = string(secret.Data[types.AWSAccessKey])
 		credentialSecret[types.AWSSecretKey] = string(secret.Data[types.AWSSecretKey])
 		credentialSecret[types.AWSEndPoint] = string(secret.Data[types.AWSEndPoint])
--- a/datastore/longhorn.go
+++ b/datastore/longhorn.go
@@ -287,11 +287,11 @@ func (s *DataStore) GetCredentialFromSecret(secretName string) (map[string]strin
 	}
 	credentialSecret := make(map[string]string)
 	if secret.Data != nil {
-		if err := s.annotateAWSIAMRole(string(secret.Data[types.AWSIAMRole])); err != nil {
+		if err := s.annotateAWSIAMRole(string(secret.Data[types.AWSIAMRoleArn])); err != nil {
 			return nil, err
 		}
 
-		credentialSecret[types.AWSIAMRole] = string(secret.Data[types.AWSIAMRole])
+		credentialSecret[types.AWSIAMRoleArn] = string(secret.Data[types.AWSIAMRoleArn])
 		credentialSecret[types.AWSAccessKey] = string(secret.Data[types.AWSAccessKey])
 		credentialSecret[types.AWSSecretKey] = string(secret.Data[types.AWSSecretKey])
 		credentialSecret[types.AWSEndPoint] = string(secret.Data[types.AWSEndPoint])
--- a/datastore/longhorn.go
+++ b/datastore/longhorn.go
@@ -232,50 +232,57 @@ func (s *DataStore) ListSettings() (map[types.SettingName]*longhorn.Setting, err
 	return itemMap, nil
 }
 
-// annotateAWSIAMRole ensures that the running pods of the manager as well as the replica instance managers.
-// have the correct AWS IAM role assigned to them based on the passed `awsIAMRole`
-func (s *DataStore) annotateAWSIAMRole(awsIAMRole string) error {
+// AnnotateAWSIAMRoleArn ensures that the running pods of the manager as well as the replica instance managers.
+// have the correct AWS IAM role arn assigned to them based on the passed `awsIAMRoleArn`
+func (s *DataStore) AnnotateAWSIAMRoleArn(controllerID, awsIAMRoleArn string) (bool, error) {
+	update := false
+
 	selector, err := s.getManagerSelector()
 	if err != nil {
-		return err
+		return update, err
 	}
 	managerPods, err := s.pLister.Pods(s.namespace).List(selector)
 	if err != nil {
-		return err
+		return update, err
 	}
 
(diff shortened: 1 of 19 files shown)


engineapi/backups.go
```diff
@@ -42,17 +41,36 @@ func getBackupCredentialEnv(backupTarget string, credential map[string]string) (
...
+			if credential[types.AWSIAMRole] == "" {
+				if credential[types.AWSAccessKey] == "" && credential[types.AWSSecretKey] == "" {
+					return nil, fmt.Errorf("Could not backup to %s without setting credential", backupType)
+				}
+				if credential[types.AWSAccessKey] == "" && credential[types.AWSSecretKey] != "" {
+					return nil, fmt.Errorf("Could not backup to %s without setting credential access key", backupType)
+				}
+				if credential[types.AWSAccessKey] != "" && credential[types.AWSSecretKey] == "" {
+					return nil, fmt.Errorf("Could not backup to %s without setting credential secret access key", backupType)
+				}
```
joshimoo, line 54: Example one how to flatten this whole if/else nesting.
You want to check your pre conditions first and exit, this way you don't need lots of inner if/else structures and can assume you are in a valid state.

I also added the missing keys, since this way the returned error is actually actionable for the user.

```go
if backupType != types.BackupStoreTypeS3 || credential == nil { 
    return env, nil
}

var missingKeys string;
if credential[types.AWSAccessKey] == "" {
    missingKeys = append(missingKeys, string(types.AWSAccessKey)
}

if credential[types.AWSSecretKey] == "" {
    missingKeys = append(missingKeys, string(types.AWSSecretKey)
}

areKeysRequired := credential[types.AWSIAMRole] == ""
if (areKeysRequired && len(missingKeys) > 0) {
    return nil, fmt.Errorf("Could not backup to %s secret is missing keys %v", backupType, missingKeys)
}

if len(missingKeys) == 0 {
    envs = append(envs, fmt.Sprintf("%s=%s", types.AWSAccessKey, credential[types.AWSAccessKey]))
    envs = append(envs, fmt.Sprintf("%s=%s", types.AWSSecretKey, credential[types.AWSSecretKey]))
}
```

engineapi/backups.go
```diff
@@ -42,17 +41,36 @@ func getBackupCredentialEnv(backupTarget string, credential map[string]string) (
...
+			}
+
 			envs = append(envs, fmt.Sprintf("%s=%s", types.AWSEndPoint, credential[types.AWSEndPoint]))
 			envs = append(envs, fmt.Sprintf("%s=%s", types.AWSCert, credential[types.AWSCert]))
 			envs = append(envs, fmt.Sprintf("%s=%s", types.HTTPSProxy, credential[types.HTTPSProxy]))
 			envs = append(envs, fmt.Sprintf("%s=%s", types.HTTPProxy, credential[types.HTTPProxy]))
 			envs = append(envs, fmt.Sprintf("%s=%s", types.NOProxy, credential[types.NOProxy]))
 			envs = append(envs, fmt.Sprintf("%s=%s", types.VirtualHostedStyle, credential[types.VirtualHostedStyle]))
-		} else if os.Getenv(types.AWSAccessKey) == "" || os.Getenv(types.AWSSecretKey) == "" {
```
joshimoo, line 54: There seems to be a change in behavior, can you verify that we don't need to support this anymore ?

Since previously if they where already present in the env, we would proceed, now it looks like the passing of a credential is required.
Which seems reasonable, since we are also setting proxy envs, but needs to be verified that this is desired.

joshimoo (commented):
Left you some code suggestions, try to flatten your if/else structures and reduce duplication.
You want to check your pre conditions first and exit, this way you don't need lots of inner if/else structures and can assume you are in a valid state.

joshimoo (dismissed):
Check the change in behavior on the keys being present on the os.env and whether we no longer need this.

engineapi/backups.go
jenting: Thank u, it's excellent.

datastore/longhorn.go
```diff
@@ -233,6 +232,71 @@ func (s *DataStore) ListSettings() (map[types.SettingName]*longhorn.Setting, err
...
+			// delete annotation to the instance-manager-r Pods if the annotation is present
+			delete(pod.Annotations, types.AWSIAMRroleAnnotation)
+			_, err = s.kubeClient.CoreV1().Pods(s.namespace).Update(pod)
+			if err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
```
joshimoo, line 298:
```suggestion
// annotateAWSIAMRole ensures that the running pods of the manager as well as the replica instance managers.
// have the correct IAM role assigned to them based on the passed `awsIAMRole`
func (s *DataStore) annotateAWSIAMRole(awsIAMRole string) error {
	selector, err := s.getManagerSelector()
	if err != nil {
		return err
	}
	
	managerPods, err := s.pLister.Pods(s.namespace).List(selector)
	if err != nil {
		return err
	}
	
	selector, err = metav1.LabelSelectorAsSelector(&metav1.LabelSelector{
		MatchLabels: types.GetInstanceManagerLabels("", "", types.InstanceManagerTypeReplica),
	})
	if err != nil {
		return err
	}

	imPods, err := s.pLister.Pods(s.namespace).List(selector)
	if err != nil {
		return err
	}
	
	pods = append(managerPods, imPods)
	for _, pod := range pods {
		val, exist:= pod.Annotations[types.AWSIAMRroleAnnotation]
		updateAnnotation := awsIAMRole != "" && awsIAMRole != val
		deleteAnnotation := awsIAMRole == "" && exist

		if updateAnnotation  {
			pod.Annotations[types.AWSIAMRroleAnnotation] = awsIAMRole
		} else if deleteAnnotation {
			delete(pod.Annotations, types.AWSIAMRroleAnnotation)
		} else {
			continue
		}
		
		if _, err = s.kubeClient.CoreV1().Pods(s.namespace).Update(pod); err != nil {
			return err
		}
	}

	return nil
}
```

Refactor to much code duplication, by restructuring the code to the above you can easily follow the logic and add trace logs if desired :)

joshimoo (commented):
Left one more code suggestions, still need to check the change of behavior, other than that LGTM

datastore/longhorn.go
jenting: Thank u, really cool!

engineapi/backups.go
jenting: I find this https://longhorn.io/docs/1.1.0/references/examples/#restore-to-file
It'll read the environment variables `AWS_ACCESS_KEY_ID` & `AWS_SECRET_ACCESS_KEY` and `AWS_ENDPOINTS`.

The helm chart or longhorn-manager itself won't expose `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` environment variables.

datastore/longhorn.go
```diff
@@ -242,6 +287,11 @@ func (s *DataStore) GetCredentialFromSecret(secretName string) (map[string]strin
 	}
 	credentialSecret := make(map[string]string)
 	if secret.Data != nil {
+		if err := s.annotateAWSIAMRole(string(secret.Data[types.AWSIAMRoleArn])); err != nil {
```
innobead, line 290: It seems this kind of on-the-fly update should be avoided, because we will cause repated query operation to do the check. 

Let's see if it's possible to have secret controller to update the annotations on demand and make the resources update in the controller reconcile loop only.

innobead (changes requested):
In general, LGTM, but one concern is about how we manage the IAM role annotations as expected efficiently.

datastore/longhorn.go
jenting: A new secret controller added to update the annotations to longhorn manager and replica instance manager Pods.

innobead:
AWS_IAM_ROLE_ARN  ✔️  & AWS_ACCESS_KEY_ID/AWS_SECRET_ACCESS_KEY  ✔️  ->  🆗  

In this case, which will be selected to use?

types/types.go
```diff
@@ -105,10 +105,12 @@ const (
 
 	BackupStoreTypeS3 = "s3"
 
-	AWSAccessKey = "AWS_ACCESS_KEY_ID"
-	AWSSecretKey = "AWS_SECRET_ACCESS_KEY"
-	AWSEndPoint  = "AWS_ENDPOINTS"
-	AWSCert      = "AWS_CERT"
+	AWSIAMRroleAnnotation = "iam.amazonaws.com/role"
```
innobead, line 108: `AWSIAMRoleAnnotation`

datastore/longhorn.go
```diff
@@ -233,6 +232,59 @@ func (s *DataStore) ListSettings() (map[types.SettingName]*longhorn.Setting, err
...
+				pod.Annotations = make(map[string]string)
+			}
+			pod.Annotations[types.AWSIAMRroleAnnotation] = awsIAMRoleArn
+		} else if deleteAnnotation {
+			delete(pod.Annotations, types.AWSIAMRroleAnnotation)
+		} else {
+			continue
+		}
+
+		update = update || true
```
innobead, line 280: This will make the update successful if at least one pod updated successfully. Is this expected?

I think we need to return the error to let event requeued to try update again.

innobead (changes requested):
In general, LGTM

Some comments to resolve/answer.

datastore/longhorn.go
```diff
@@ -233,6 +232,59 @@ func (s *DataStore) ListSettings() (map[types.SettingName]*longhorn.Setting, err
...
+		return update, err
+	}
+
+	selector, err = metav1.LabelSelectorAsSelector(&metav1.LabelSelector{
+		MatchLabels: types.GetInstanceManagerLabels("", "", types.InstanceManagerTypeReplica),
+	})
+	if err != nil {
+		return update, err
+	}
+	imPods, err := s.pLister.Pods(s.namespace).List(selector)
```
innobead, line 255: For AWS IAM Role, we only can apply to the `pod` level instead of `process` level, so that means this will be a problem in the future for mulitple AWS credentials support.

But, I would say this is the limitation when using `assume role`, so it can be documented when we support multiple AWS credentials in the future.

datastore/longhorn.go
jenting: It's a `update` variable to determine to print log `Annotates the AWS IAM role succeeded` or not.
If one of the Pod updates failed, the error returned immediately and the event would be re-queue.

```go
        update, err := ks.ds.AnnotateAWSIAMRoleArn(ks.controllerID, awsIAMRoleArn)
	if err != nil {
		return err
	}

	if update {
		ks.logger.Infof("Annotates the AWS IAM role succeeded")
	}
```

innobead: Yes, but it's always `true` because of `update = update || true`, so I doubt the `if update` any meaning in this pod update loop?

innobead: After checking again, there is a case for non-update, so it does have some meaning :)

innobead (dismissed):
LGTM

jenting:
> AWS_IAM_ROLE_ARN  & AWS_ACCESS_KEY_ID/AWS_SECRET_ACCESS_KEY  -> 
> 
> In this case, which will be selected to use? I think it should depend on the AWS s3 client lib. (probably users will ask later)

1st: AWS_ACCESS_KEY_ID & AWS_SECRET_ACCESS_KEY
2nd: AWS_IAM_ROLE_ARN

Reference to the code
- https://github.com/aws/aws-sdk-go/blob/v1.25.16/aws/config.go#L39-L42
- https://github.com/aws/aws-sdk-go/blob/v1.25.16/aws/defaults/defaults.go#L100-L113

innobead (approved):

PR #843, opened 2021-03-11
Fix one typo for the CPU reservation

shuo-wu:
longhorn/longhorn#2325

### DIFF
--- a/types/setting.go
+++ b/types/setting.go
@@ -744,7 +744,7 @@ func ValidateInitSetting(name, value string) (err error) {
 		}
 	case SettingNameGuaranteedEngineCPU:
 		if value != "" {
-			return errors.Wrapf(err, "Cannot set a value %v for the deprecated CPU resource settings", value)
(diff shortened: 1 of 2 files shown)


shuo-wu:
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/60/

types/setting.go
```diff
@@ -744,7 +744,7 @@ func ValidateInitSetting(name, value string) (err error) {
 		}
 	case SettingNameGuaranteedEngineCPU:
 		if value != "" {
-			return errors.Wrapf(err, "Cannot set a value %v for the deprecated CPU resource settings", value)
+			return fmt.Errorf("cannot set a value %v for the deprecated settings %v", value, sName)
```
c3y1huang, line 747:
```suggestion
			return fmt.Errorf("cannot set a value %v for the deprecated setting %v", value, sName)
```

c3y1huang (dismissed):
Thanks!

innobead (approved):

PR #844, opened 2021-03-11
Fixes the RWX attach call

joshimoo:
Also added a fix in a separate commit for an issue on the CSI drivers, volume expansion idempotency. Since I was testing the RWX expansion at the same time.

https://github.com/longhorn/longhorn/issues/2316
https://github.com/longhorn/longhorn/issues/2181

@shuo-wu closed the other pr in favor of this one, kept the old error behavior except in the case where the requested size == the actual size, since that makes the call idempotent. We still error out in the longhorn api where size < actual size.
For the csi part it's expected behavior to return success in case where size < actual size and include the actual size.

### DIFF
--- a/csi/controller_server.go
+++ b/csi/controller_server.go
@@ -325,10 +325,6 @@ func (cs *ControllerServer) publishVolume(volume *longhornclient.Volume, nodeID 
 }
 
 func (cs *ControllerServer) publishSharedVolume(req *csi.ControllerPublishVolumeRequest, volume *longhornclient.Volume) (*csi.ControllerPublishVolumeResponse, error) {
-	if !volume.Ready || len(volume.Controllers) == 0 || volume.Controllers[0].Endpoint == "" {
-		return nil, status.Errorf(codes.Aborted, "The volume %s is not ready for workloads", req.GetVolumeId())
-	}
-
 	return cs.publishVolume(volume, req.NodeId, func() error {
 		if !cs.waitForVolumeState(volume.Name, "share available", isVolumeShareAvailable, false, false) {
 			return status.Errorf(codes.DeadlineExceeded, "Failed to wait for volume %v share available", req.GetVolumeId())
@@ -338,13 +334,6 @@ func (cs *ControllerServer) publishSharedVolume(req *csi.ControllerPublishVolume
 }
 
 func (cs *ControllerServer) publishMigratableVolume(req *csi.ControllerPublishVolumeRequest, volume *longhornclient.Volume) (*csi.ControllerPublishVolumeResponse, error) {
-	if volume.State == string(types.VolumeStateAttached) {
-		if !volume.Ready || len(volume.Controllers) == 0 || volume.Controllers[0].Endpoint == "" {
-			return nil, status.Errorf(codes.Aborted,
-				"The volume %s is already attached but it is not ready for workloads", req.GetVolumeId())
-		}
-	}
-
 	checkVolumeAttached := func(vol *longhornclient.Volume) bool { return isVolumeAvailableOn(vol, req.NodeId) }
 	return cs.publishVolume(volume, req.NodeId, func() error {
 		if !cs.waitForVolumeState(req.GetVolumeId(), "volume available on", checkVolumeAttached, false, false) {
@@ -355,13 +344,7 @@ func (cs *ControllerServer) publishMigratableVolume(req *csi.ControllerPublishVo
 }
 
 func (cs *ControllerServer) publishReadWriteOnceVolume(req *csi.ControllerPublishVolumeRequest, volume *longhornclient.Volume) (*csi.ControllerPublishVolumeResponse, error) {
-	// the volume is already attached make sure it's to the same node as this
 	if volume.State == string(types.VolumeStateAttached) {
-		if !volume.Ready || len(volume.Controllers) == 0 || volume.Controllers[0].Endpoint == "" {
-			return nil, status.Errorf(codes.Aborted,
-				"The volume %s is already attached but it is not ready for workloads", req.GetVolumeId())
-		}
-
 		if volume.Controllers[0].HostId != req.GetNodeId() {
 			return nil, status.Errorf(codes.FailedPrecondition,
 				"The volume %s cannot be attached to the node %s since it is already attached to the node %s",
@@ -423,13 +406,15 @@ func (cs *ControllerServer) ControllerPublishVolume(ctx context.Context, req *cs
 		return nil, status.Errorf(codes.Aborted, "The volume %s is %s", req.GetVolumeId(), existVol.State)
 	}
 
-	// Check volume frontend settings
 	if existVol.Frontend != string(types.VolumeFrontendBlockDev) {
 		return nil, status.Errorf(codes.InvalidArgument, "ControllerPublishVolume: there is no block device frontend for volume %s", req.GetVolumeId())
 	}
 
-	if !existVol.Ready {
-		return nil, status.Errorf(codes.Aborted, "The volume %s is not ready for workloads", req.GetVolumeId())
+	isAttached := existVol.State == string(types.VolumeStateAttached)
+	waitForController := isAttached && (len(existVol.Controllers) == 0 || existVol.Controllers[0].Endpoint == "")
+	if !existVol.Ready || waitForController {
+		return nil, status.Errorf(codes.Aborted, "The volume %s in state %v is not ready for workloads, is waiting on controller %v",
+			req.GetVolumeId(), existVol.State, waitForController)
 	}
 
 	if requiresSharedAccess(existVol, volumeCapability) {
--- a/csi/controller_server.go
+++ b/csi/controller_server.go
@@ -732,23 +732,52 @@ func (cs *ControllerServer) ControllerExpandVolume(ctx context.Context, req *csi
 	if existVol.State != string(types.VolumeStateDetached) {
 		return nil, status.Errorf(codes.FailedPrecondition, "Invalid volume state %v for expansion", existVol.State)
 	}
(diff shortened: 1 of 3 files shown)


joshimoo:
Ran the e2e core tests, there where some failures in regard to the backupstore volume path, which we are working on.
Manually tested attach/detach/expansion for rwx and rwo volumes.

csi/controller_server.go
```diff
@@ -325,10 +325,6 @@ func (cs *ControllerServer) publishVolume(volume *longhornclient.Volume, nodeID
 }
 
 func (cs *ControllerServer) publishSharedVolume(req *csi.ControllerPublishVolumeRequest, volume *longhornclient.Volume) (*csi.ControllerPublishVolumeResponse, error) {
-	if !volume.Ready || len(volume.Controllers) == 0 || volume.Controllers[0].Endpoint == "" {
-		return nil, status.Errorf(codes.Aborted, "The volume %s is not ready for workloads", req.GetVolumeId())
-	}
-
```
PhanLe1010, line 331: Can you explain why we can skip the check `len(volume.Controllers) == 0 || volume.Controllers[0].Endpoint == ""`?

csi/controller_server.go
```diff
@@ -732,23 +732,49 @@ func (cs *ControllerServer) ControllerExpandVolume(ctx context.Context, req *csi
...
+			engineReady = engineSize >= requestedSize && !engine.IsExpanding
+		}
+		size, _ := strconv.ParseInt(vol.Size, 10, 64)
+		return size >= requestedSize && engineReady
+	}
+
+	if !cs.waitForVolumeState(req.VolumeId, "volume expansion", volumeExpansionComplete, false, false) {
+		return nil, status.Errorf(codes.DeadlineExceeded, "Expanding volume %s existing capacity %v requested capacity %v failed",
+			req.GetVolumeId(), existingSize, requestedSize)
 	}
```
PhanLe1010: Is it a problem if it takes longer than 2 minutes to finish the expansion?
Does it timeout on the Kubernetes side? Does Kubernetes retry the expansion call?

csi/controller_server.go
joshimoo: Sure thing, I will explain it based on the examples below.
- volume detached, we can see that the engine endpoint is not available, since the engine is not running.
- volume attached in maintenance mode, we can see that the engine endpoint is not available, since the frontend is disabled.
- volume after expansion, we can see that the engine endpoint is not available and that the `frontend is still disabled`, since the expansion disables the frontend but there is nothing that re enables it after expansion, it's expected that the following attach call will re enable it, if we prevent that attach call we cannot re enable the frontend.
- volume attached, we can see that the engine endpoint is available

**Conclusion:**
An engines endpoint is only available while it's attached in non maintenance mode, using this condition outside of these circumstances is not valid to determine whether the volume is ready for workloads. 

You might be asking yourself why we need to check this at all, currently the csi driver has way to much logic and all of this should be handled by the volume-manager, see my comment in the dev channel in regard to making the driver dumber.

The goal should be to have the driver call attach on the api then poll & evaluate the result. But that requires structured error details, so we can act on them, see epic #1875 

Currently returning all these Aborted errors on transient states, just confuses the users since they think something is wrong. Instead it would be better to just call attach & poll since the poll would wait for the transient state to pass and return success as expected instead of kubernetes having to retry the whole csi call.


**volume detached**
```
controllers": [
{
"address": "",
"endpoint": "",
"hostId": "",
"instanceManagerName": "",
"name": "test-rwx-manual-e-cfd524c8",
"running": false,
},
],
"disableFrontend": false,
```

**volume attached in maintenance mode**
```
"controllers": [
{
"address": "10.42.1.251",
"endpoint": "",
"hostId": "jmoody-lh-dev-work1",
"instanceManagerName": "instance-manager-e-e16b4017",
"name": "test-rwx-maintenance-e-dfbc8076",
"running": true,
},
],
"disableFrontend": true,
```

**volume after expansion**
```
"controllers": [
{
"address": "",
"endpoint": "",
"hostId": "",
"instanceManagerName": "",
"name": "pvc-a9579ac8-0351-4c8a-8ba0-d24214fcd941-e-51fd4e2c",
"running": false,
},
],
"disableFrontend": true,
```

**volume attached**
```
"controllers": [
{
"address": "10.42.3.37",
"endpoint": "/dev/longhorn/pvc-a9579ac8-0351-4c8a-8ba0-d24214fcd941",
"hostId": "jmoody-lh-dev-work2",
"instanceManagerName": "instance-manager-e-b1eda47e",
"name": "pvc-a9579ac8-0351-4c8a-8ba0-d24214fcd941-e-51fd4e2c",
"running": true,
},
],
"disableFrontend": false,
```

csi/controller_server.go
joshimoo: Not a problem, all csi calls needs to be idempotent.
It's important to wait for the expansion to succeed otherwise you will have signaled that the system has updated the size, which will lead to the external-resizer updating the PV size. 

Now imagine what would happen if there is a crash in the middle of any of this. Now the PV and actual volume size might be out of sync. So by waiting we ensure that longhorn and the PV size are in sync.

If longhorn has already done the update, but the PV update failed then this call will be retried since the PVC.requestedSize > PV.size. That's why it's important that this call is idempotent.

joshimoo: Step all the way towards: 
https://github.com/kubernetes-csi/external-resizer/blob/372f7514451bd6f54023e8df5860f606cc73f938/pkg/controller/controller.go#L270
https://github.com/kubernetes-csi/external-resizer/blob/372f7514451bd6f54023e8df5860f606cc73f938/pkg/controller/controller.go#L437
https://github.com/kubernetes-csi/external-resizer/blob/372f7514451bd6f54023e8df5860f606cc73f938/pkg/resizer/csi_resizer.go#L177

final csi call:
https://github.com/kubernetes-csi/external-resizer/blob/372f7514451bd6f54023e8df5860f606cc73f938/pkg/csi/client.go#L122

joshimoo:
rebased and updated the csi-external-resizer timeout to be consistent with the poll timeout for the expansion call.

csi/controller_server.go
PhanLe1010: Thank you for the explanation

csi/controller_server.go
PhanLe1010: Thank you for the explanation

PhanLe1010 (approved):
LGTM

csi/deployment.go
```diff
@@ -200,6 +200,7 @@ func NewResizerDeployment(namespace, serviceAccount, resizerImage, rootDir strin
 		[]string{
 			"--v=5",
 			"--csi-address=$(ADDRESS)",
+			"--timeout=2m5s",
```
yasker, line 203: Why we specify a timeout here?

joshimoo: Since the default timeout is 10s and we now do verify that the volume expansion is completed, see comment answer to phan's question above, this is important to keep kubernetes state and longhorns state consistent.

The timeout we use for the waitForVolumeState is hardcoded at 2 minutes currently, so I add 5s extra :)
So that we can timeout the active call. This ensures that the external-resizer does not flood the csi driver with a bunch of unnecessary concurrent ExpandCalls for the same volume. Since we will return success as soon as the expansion is completed or timeout the call.

```go
	if !cs.waitForVolumeState(req.VolumeId, "volume expansion", volumeExpansionComplete, false, false) {
		return nil, status.Errorf(codes.DeadlineExceeded, "Expanding volume %s existing capacity %v requested capacity %v failed",
			req.GetVolumeId(), existingSize, requestedSize)
	}
```

csi/controller_server.go
```diff
@@ -732,23 +732,49 @@ func (cs *ControllerServer) ControllerExpandVolume(ctx context.Context, req *csi
 	if existVol.State != string(types.VolumeStateDetached) {
 		return nil, status.Errorf(codes.FailedPrecondition, "Invalid volume state %v for expansion", existVol.State)
 	}
-	volumeSize, err := strconv.ParseInt(existVol.Size, 10, 64)
+	existingSize, err := strconv.ParseInt(existVol.Size, 10, 64)
```
shuo-wu, line 735: The naming is a little bit weird... How about `currentSize`?

csi/deployment.go
shuo-wu: Can we add this explaination as a comment here (with a brief version)?

shuo-wu (changes requested):
Can we directly make the Longhorn API `Attach` idempotent? Then I don't need to check the volume state for `publishReadWriteOnceVolume`. Furthermore, we don't need functions `publishReadWriteOnceVolume`, `publishMigratableVolume`, and `publishSharedVolume` then. Instead, simply calling `attach` and waiting for the volume state (based on the volume type) is enough.

Right now, Longhorn API `Attach` is exposed and idempotent for attached RWX volumes but not for RWO volume. It's not consistent. I do remember you want to directly make it idempotent but I don't remember the details. Any blocks for the RWO part?

csi/controller_server.go
joshimoo: It is `existingSize` because it doesn't get updated and is used below in the log message.
Since we retrieve an updated volume below, therefore I find keeping existing Size better.

csi/deployment.go
joshimoo: sure thing, but we didn't add any comments when we changed the timeout value for the attacher :)

joshimoo:
@shuo-wu yes I want to make the csi driver dumber and have longhorn backend deal with attach/detach, which Is what i am working on today. But this should be merged before that just so that we can have functioning RWX volumes :)

joshimoo:
@shuo-wu rebased and added the requested comments.

shuo-wu (approved):

PR #845, opened 2021-03-12
Fix volume name not auto-correct to lowercase

c3y1huang:
Volume name not converted to lower-case.

https://github.com/longhorn/longhorn/issues/2270


shuo-wu (approved):

joshimoo (approved):
LGTM

PR #846, opened 2021-03-12
Fix backup monitor polling from each node at the same time

joshimoo:
Currently the backupstore monitor poll is processed by every node in the 
cluster, this is undesired since it leads to excessive S3 calls and cost.

This also has the side effect that each nodes backupstore monitor will 
race against each other to try to update each volumes lastBackup field. 
This leads to unnecessary conflicts and processing of the volumes.

Since the backup monitor is run on each node, but we only need a single 
update we pick a consistent random ready node for each poll run. This is
accomplished by the following steps:
- sort the candidate list, this normalizes the list across nodes.
- use a time index to derive an index into the candidate list, this 
  normalizes time differences across nodes.
- arbitrarily choose the pollInterval as your time normalization factor, 
  since this also has the benefit of doing round robin across the at the 
  time available candidate nodes.

longhorn/longhorn#1547

### DIFF
--- a/controller/setting_controller.go
+++ b/controller/setting_controller.go
@@ -7,6 +7,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"reflect"
+	"sort"
 	"time"
 
 	"github.com/pkg/errors"
@@ -46,7 +47,8 @@ var (
 type SettingController struct {
 	*baseController
 
-	namespace string
+	namespace    string
+	controllerID string
 
 	kubeClient    clientset.Interface
 	eventRecorder record.EventRecorder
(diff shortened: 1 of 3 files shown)


controller/setting_controller.go
```diff
@@ -524,22 +534,71 @@ func getFinalTolerations(existingTolerations, lastAppliedTolerations, newTolerat
 }
 
 func (bm *BackupStoreMonitor) Start() {
+	log := bm.logger.WithFields(logrus.Fields{
```
shuo-wu, line 537: Add controller ID

shuo-wu (approved):
In general LGTM

controller/setting_controller.go
joshimoo: We already have a `node` field as part of the initial logger and we use a component field for the backup-store-monitor.
Not sure what adding a controllerID field would get us?

PR #848, opened 2021-03-17
Fix bug Longhorn upgrades more than concurrentAutomaticEngineUpgradePerNodeLimit engines per nodes

PhanLe1010:
`limitAutomaticEngineUpgradePerNode()` may be called multiple times with the same inputs. Therefore, we need to sort the `upgradingCandidates` slice so that the function has deterministic finalCandidates output. This prevents selecting more than `concurrentAutomaticEngineUpgradePerNodeLimitcandidates` to upgrade at the same time.

I am not sure why the `limitAutomaticEngineUpgradePerNode()` sometime is called multiple times with the same inputs. @joshimoo , @yasker Do you have any idea? I actually don't like the current fix (sorting the list of volumes) because it may hurt Longhorn in the long run when Longhorn is used to manage million of volumes. So, if we can prevent the `limitAutomaticEngineUpgradePerNode()` from being called multiple times with the same inputs, it would be a better solution, I think.

longhorn/longhorn#2328

### DIFF
--- a/controller/engine_image_controller.go
+++ b/controller/engine_image_controller.go
@@ -353,7 +353,7 @@ func (ic *EngineImageController) syncEngineImage(key string) (err error) {
 		engineImage.Status.State = types.EngineImageStateDeployed
 	}
 
-	if err := ic.handleAutoUpgradeEngineImageToDefaultEngineImage(); err != nil {
+	if err := ic.handleAutoUpgradeEngineImageToDefaultEngineImage(engineImage.Spec.Image); err != nil {
 		log.WithError(err).Warn("error when handleAutoUpgradeEngineImageToDefaultEngineImage")
 	}
 
@@ -393,7 +393,23 @@ func (ic *EngineImageController) syncNodeDeploymentMap(engineImage *longhorn.Eng
 }
 
 // handleAutoUpgradeEngineImageToDefaultEngineImage automatically upgrades volume's engine image to default engine image when it is applicable
-func (ic *EngineImageController) handleAutoUpgradeEngineImageToDefaultEngineImage() error {
+func (ic *EngineImageController) handleAutoUpgradeEngineImageToDefaultEngineImage(currentProcessingImage string) error {
+	defaultEngineImage, err := ic.ds.GetSettingValueExisted(types.SettingNameDefaultEngineImage)
+	if err != nil {
+		return err
+	}
+
+	// To avoid multiple managers doing upgrade at the same time, only allow the
+	// manager that is responsible for the default engine image to do the upgrade
+	if currentProcessingImage != defaultEngineImage {
+		return nil
+	}
+
+	defaultEngineImageResource, err := ic.ds.GetEngineImage(types.GetEngineImageChecksumName(defaultEngineImage))
+	if err != nil {
+		return err
+	}
+
 	concurrentAutomaticEngineUpgradePerNodeLimit, err := ic.ds.GetSettingAsInt(types.SettingNameConcurrentAutomaticEngineUpgradePerNodeLimit)
 	if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


yasker:
If you meant by `limitAutomaticEngineUpgradePerNode` was called multiple times due to there are multiple nodes,  @joshimoo was working on a mechanism to only allow one node to run some logic of the setting controller, which might help. Otherwise, you need to figure out the reason since it's your code..

PhanLe1010:
Thanks @yasker . I meant it is called multiple time with the same inputs on the same node. Yeah, I will continue doing more investigation by myself.

PhanLe1010:
I noticed that this behavior happens in the volume controller as well https://github.com/longhorn/longhorn-manager/blob/0b4f6ef17c3f02f1c8f52576066abc4926ca56d2/controller/volume_controller.go#L812
I will do more investigation

joshimoo:
Sorting is the correct approach since you need to normalize the input, since it comes from an unsorted map.

**Improvements:**
Let's return a map of candidates per node here, since we already need to process all volumes.
- getVolumesForEngineImageUpgrading

Then this turns into a for,for loop but you can add an debug log like `requesting engine image update for Y volumes on node X`
https://github.com/longhorn/longhorn-manager/blob/b0b3579609ac768d76271f68f662ce2243b6cb99/controller/engine_image_controller.go#L428
And in the future if we ever need we can run the update calls in parallel. (don't do this now)

refactoring below: 
- return candidates + in progress
- remove err since we never return any
- cleanup naming + simplified logic (valid test instead of invalid test)

```go
func (ic *EngineImageController) getVolumeCandidatesForEngineUpgrade(volumes map[string]*longhorn.Volume,
	newEngineImageResource *longhorn.EngineImage) (candidates map[string][]*longhorn.Volume, inProgress map[string][]*longhorn.Volume) {
	candidates = make(map[string][]*longhorn.Volume)
	inProgress = make(map[string][]*longhorn.Volume)
	for _, v := range volumes {
		// TODO: what's the correct node to use here, we use currentNode below for the readiness check
		node := v.Status.OwnerID
		if v.Spec.EngineImage != v.Status.CurrentImage {
			inProgress[node] = append(inProgress[node], v)
			continue
		}

		canBeUpgraded := ic.canDoOfflineEngineImageUpgrade(v, newEngineImageResource) || ic.canDoLiveEngineImageUpgrade(v, newEngineImageResource)
		engineImageAvailable, _ := ic.ds.CheckEngineImageReadinessForVolume(newEngineImageResource.Spec.Image, v.Name, v.Status.CurrentNodeID)
		validCandidate := v.Spec.EngineImage != newEngineImageResource.Spec.Image && engineImageAvailable && canBeUpgraded
		if validCandidate {
			candidates[node] = append(candidates[node], v)
		}
	}

	return candidates, inProgress
}
```

limiting is easy now, pseudo code below.
```go
limitedCandidates;
for node := range candidates {
upgradeCount := len(inProgress[node])
if upgradeCount >= maxLimit {
    continue
}

// sort & slice 
limitedCandidates[node] = candidates[node].sort()[:maxLimit - upgradeCount]
}

return limitedCandidates

```




**Let's fix some other issues while we are touching the image-controller:**

for each enqueue remove these type of checks, since they prohibit ownership transfer
https://github.com/longhorn/longhorn-manager/blob/b0b3579609ac768d76271f68f662ce2243b6cb99/controller/engine_image_controller.go#L656

The controller should at least look at each resource to determine whether it's responsible for it or not, it's no the enqueues responsibility to make that determination. This is a form of optimization we can add later if necessary, but for now remove these checks from the enques the same applies for other controllers. 



Knowledge sharing:
Sorting normalizes the input for a single consumer, as long as the input is only processed on that node, this is sufficient.
If the input needs to be processed across nodes, by only a single node of the set of nodes, you need to add an additional dimension in this case time since it's assumed to be relatively stable across nodes.
Have a look at this pr, where I normalize a input across many nodes and only select a single node to process it.
https://github.com/longhorn/longhorn-manager/pull/846

joshimoo (commented):
LGTM sorting is the correct approach, after implementing the suggested improvements feel free to merge this.

PhanLe1010:
Thank you  @joshimoo !

1. I have refactored the code base on the suggested improvements
1. We don't need to sort. The problem was that there were multiple manager doing the upgrade at the same time. It is my mistake.
1. I will fix the ownership transfer in the issue https://github.com/longhorn/longhorn/issues/2377

PhanLe1010:
ping @shuo-wu
Can you help to take a look and merge this PR?

controller/engine_image_controller.go
```diff
@@ -394,72 +394,67 @@ func (ic *EngineImageController) syncNodeDeploymentMap(engineImage *longhorn.Eng
...
 	}
-	if concurrentAutomaticEngineUpgradePerNodeLimit <= 0 {
+	defaultEngineImageResource, err := ic.ds.GetEngineImage(types.GetEngineImageChecksumName(defaultEngineImage))
+	if err != nil {
+		return err
+	}
+	// Make sure that only the owner of the default engine image does the upgrade
+	if defaultEngineImageResource.Status.OwnerID != ic.controllerID {
 		return nil
 	}
```
shuo-wu, line 408: Why not directly check if the current engineimage is the default one? I prefer not to consider the ownership issue here.

shuo-wu (commented):
Others LGTM

controller/engine_image_controller.go
```diff
@@ -394,72 +394,67 @@ func (ic *EngineImageController) syncNodeDeploymentMap(engineImage *longhorn.Eng
...
+
+	for _, vs := range limitedCandidates {
+		for _, v := range vs {
+			oldEngineImage := v.Spec.EngineImage
+			v.Spec.EngineImage = defaultEngineImage
+			v, err = ic.ds.UpdateVolume(v)
+			if err != nil {
+				return err
+			}
+			ic.logger.Infof("automatically upgrades volume's engine image from %v to the default engine image %v for volume %v", oldEngineImage, defaultEngineImage, v.Name)
```
joshimoo, line 436:
```suggestion
			log := ic.logger.withField("volume", v.Name).withField("engineImage", v.Spec.EngineImage).Infof("automatically upgrade volume engine image to the default engine image %v", defaultEngineImage)
			v.Spec.EngineImage = defaultEngineImage
			v, err = ic.ds.UpdateVolume(v)
			if err != nil {
				return err
			}
```

joshimoo: using fields allows to trace on a specific volume/engineImage

controller/engine_image_controller.go
```diff
@@ -468,28 +463,25 @@ func limitAutomaticEngineUpgradePerNode(volumes map[string]*longhorn.Volume, upg
...
 			continue
 		}
-		if isReady, _ := ic.ds.CheckEngineImageReadinessForVolume(newEngineImageResource.Spec.Image, v.Name, v.Status.CurrentNodeID); !isReady {
-			continue
-		}
-		if !ic.canDoOfflineEngineImageUpgrade(v, newEngineImageResource) && !ic.canDoLiveEngineImageUpgrade(v, newEngineImageResource) {
-			continue
+		canBeUpgraded := ic.canDoOfflineEngineImageUpgrade(v, newEngineImageResource) || ic.canDoLiveEngineImageUpgrade(v, newEngineImageResource)
+		isCurrentEIAvailable, _ := ic.ds.CheckEngineImageReadinessForVolume(v.Status.CurrentImage, v.Name, v.Status.CurrentNodeID)
+		isNewEIAvailable, _ := ic.ds.CheckEngineImageReadinessForVolume(newEngineImageResource.Spec.Image, v.Name, v.Status.CurrentNodeID)
```
joshimoo, line 477: in the case where the volume is detached, what is the value of status.CurrentNodeID? should these be using the v.Status.OwnerID as well? Or do we only care about the replicas in the case where the volume is detached?

controller/engine_image_controller.go
joshimoo: I agree with this, move the check outside into the main sync loop.
Then rename this function to `handleAutoUpgradeEngineImage(newImage *longhorn.EngineImage, limit int)` pass the new image (default) and limit to this function. Then the only function that needs to retrieve settings is the main sync loop in the case where it's the owner of the default image.

https://github.com/longhorn/longhorn-manager/blob/0a2885634f0f7fcf159721132d53b501804248f3/controller/engine_image_controller.go#L356

joshimoo (commented):
LGTM, I agree with Shuo's suggestion, see my comment.
Afterwards this can be merged :)

controller/engine_image_controller.go
PhanLe1010: Yeah, I think we only care about the replica when volume is detached.

controller/engine_image_controller.go
PhanLe1010: Thank you! Fixed

controller/engine_image_controller.go
PhanLe1010: Thanks @shuo-wu 
Fixed!

controller/engine_image_controller.go
```diff
@@ -393,73 +393,70 @@ func (ic *EngineImageController) syncNodeDeploymentMap(engineImage *longhorn.Eng
...
-		oldEngineImage := v.Spec.EngineImage
-		v.Spec.EngineImage = defaultEngineImage
-		v, err = ic.ds.UpdateVolume(v)
-		if err != nil {
-			return err
+	limitedCandidates := limitAutomaticEngineUpgradePerNode(candidates, inProgress, int(concurrentAutomaticEngineUpgradePerNodeLimit))
+
+	for _, vs := range limitedCandidates {
+		for _, v := range vs {
+			ic.logger.WithField("volume", v.Name).WithField("engineImage", v.Spec.EngineImage).Infof("automatically upgrade volume engine image to the default engine image %v", defaultEngineImage)
```
shuo-wu, line 433: You can directly use `ic.logger.WithFields(logrus.Fields{"volume": v.Name, "engineImage": v.Spec.EngineImage})`

shuo-wu (approved):
In general LGTM

controller/engine_image_controller.go
joshimoo: Sounds reasonable if it's best effort, then keep it in there. 
Reasons for exposing the parameters are the ability to do unit tests against that function, but it's not important for the moment so keep it the way you had it with the settings and everything included in that single function.

controller/engine_image_controller.go
```diff
@@ -393,73 +393,70 @@ func (ic *EngineImageController) syncNodeDeploymentMap(engineImage *longhorn.Eng
...
 	if err != nil {
 		return err
 	}
-	if concurrentAutomaticEngineUpgradePerNodeLimit <= 0 {
+
+	// To avoid multiple managers doing upgrade at the same time, only allow the
+	// manager that is responsible for the default engine image to do the upgrade
+	if currentProcessingImage != defaultEngineImage {
 		return nil
 	}
```
joshimoo, line 406: This is more than just limiting to a single node, this limits to just events from the `defaultEngineImage`.
With the 30s resync period this will work for now, but might need to be looked at again later once we remove the 30s resync period.

joshimoo (dismissed):
LGTM, left a comment about a potential change in regard to removal of the 30s resync period that we might need to have another look at later.

controller/engine_image_controller.go
PhanLe1010: Yeah, this was the reason that I did the check `defaultEngineImage .Status.OwnerID == ec.ControllerID` in the previous push since it captures more events. But it still doesn't guarantee 100%. So, I switched to this recommended check to make the logic independent of the `Status.OwnerID` field.

controller/engine_image_controller.go
PhanLe1010: Thank you! Fixed

joshimoo (approved):

PR #849, opened 2021-03-19
Enable uninstall job security context privileged

jenting:
#### Symptom

```shell
kubectl apply -f https://raw.githubusercontent.com/longhorn/longhorn/v1.1.0/deploy/longhorn.yaml
kubectl create -f https://raw.githubusercontent.com/longhorn/longhorn/v1.1.0/uninstall/uninstall.yaml
kubectl get job/longhorn-uninstall -n default -w
```
The job pod in the stuck state.

#### Root Cause
According to PSP Policy Order:

> 1. PodSecurityPolicies which allow the pod as-is, without changing defaults or mutating the pod, are preferred. The order of these non-mutating PodSecurityPolicies doesn't matter.
> 2. If the pod must be defaulted or mutated, the first PodSecurityPolicy (ordered by name) to allow the pod is selected.

since the uninstall Job did not set any `securityContext`, it causes it would fall into `global-restricted-psp` first since the pod must be defaulted or mutated, the first PodSecurityPolicy (ordered by name) to allow the pod is selected.

#### Solution

- Configure the `privileged: true` to uninstall Job, to match against 
https://github.com/longhorn/longhorn-manager/blob/v1.1.0/deploy/install/02-components/01-manager.yaml#L21-L22
- Remove the new service account and uses the existing service-account create when install https://github.com/longhorn/longhorn-manager/blob/v1.1.0/deploy/install/01-prerequisite/02-serviceaccount.yaml (this matches the helm chart)

#### Issues
https://github.com/longhorn/longhorn/issues/2292

### DIFF
--- a/deploy/uninstall/uninstall.yaml
+++ b/deploy/uninstall/uninstall.yaml
@@ -1,63 +1,8 @@
-apiVersion: v1
-kind: ServiceAccount
-metadata:
-  name: longhorn-uninstall-service-account
-  namespace: default
----
-apiVersion: rbac.authorization.k8s.io/v1
-kind: ClusterRole
-metadata:
-  name: longhorn-uninstall-role
-rules:
-  - apiGroups:
-      - apiextensions.k8s.io
-    resources:
-      - customresourcedefinitions
-    verbs:
-      - "*"
-  - apiGroups: [""]
-    resources: ["pods", "persistentvolumes", "persistentvolumeclaims", "nodes", "configmaps", "secrets", "services", "endpoints"]
-    verbs: ["*"]
-  - apiGroups: ["apps"]
-    resources: ["daemonsets", "statefulsets", "deployments"]
-    verbs: ["*"]
-  - apiGroups: ["batch"]
-    resources: ["jobs", "cronjobs"]
-    verbs: ["*"]
-  - apiGroups: ["policy"]
-    resources: ["poddisruptionbudgets"]
-    verbs: ["*"]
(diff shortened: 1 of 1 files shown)


innobead (approved):
LGTM

joshimoo:
@meldafrawi @yasker is the change of moving the uninstall job into the `longhorn-system` namespace ok, was there a reason the uninstall job was ran in the `default` namespace?

joshimoo:
While reviewing this I noticed that the share-manager crds are not getting processed in the uninstall-controller.
@innobead I added this issue https://github.com/longhorn/longhorn/issues/2384 with instructions on how to fix this part.

yasker:
@joshimoo No, it's NOT OK to move the uninstaller to the `longhorn-system` namespace.

When user delete the `longhorn-system` namespace by mistake and get stuck, the Pod cannot be created in that namespace. That's why we're running the uninstaller job in default namespace.

cc @innobead @jenting

innobead:
> @joshimoo No, it's NOT OK to move the uninstaller to the `longhorn-system` namespace.
> 
> When the user deletes the `longhorn-system` namespace by mistake and gets stuck, the Pod cannot be created in that namespace. That's why we're running the uninstaller job in the default namespace.
> 
> We should also add a manual test case for this scenario.
> 
> cc @innobead @jenting

Thanks for the info.

@jenting as we discussed, this seems a fair concern, so we should make sure to `uninstall job` run by kubectl in a different namespace but still with the correct role binding with the authorized PSP.

Let's do another PR. Thanks.

PR #851, opened 2021-03-23
Keep uninstall job in default namespace

jenting:
#### Why

According to https://github.com/longhorn/longhorn-manager/pull/849#issuecomment-804234586
> When the user deletes the longhorn-system namespace by mistake and gets stuck, the Pod cannot be created in that namespace. That's why we're running the uninstaller job in the default namespace.

#### How
- Rvert the change of removing ServiceAccount, ClusterRole, and ClusterRoleBinding
- Keep the job running in `default` namespace
- Add PSP `longhorn-uninstall-psp` for ServiceAccount `longhorn-uninstall-service-account`

#### Issues
https://github.com/longhorn/longhorn/issues/2292

### DIFF
--- a/deploy/uninstall/uninstall.yaml
+++ b/deploy/uninstall/uninstall.yaml
@@ -1,8 +1,63 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: longhorn-uninstall-service-account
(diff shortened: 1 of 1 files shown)


innobead (approved):
LGTM

PR #852, opened 2021-03-24
Fixed engine image never becomes deployed after upgrading

PhanLe1010:
We recently changed the labels for engine image CR in the PR https://github.com/longhorn/longhorn-manager/pull/842. We added the label `longhorn.io/managed-by: longhorn-manager` to `engine image`, `engine image daemonset`, and `engine image daemonset pod` by updating the fucntion [GetEngineImageLabels()](https://github.com/longhorn/longhorn-manager/blob/f6ff18666ea238c4eec4bfdce9612796da1a0701/types/types.go#L234)

On the other hand, during Longhorn upgrading, we only update the label for `engine image`, `engine image daemonset` but not the `engine image daemonset pod`  of the existing engine image.

This creates a problem when we [syncNodeDeploymentMap()](https://github.com/longhorn/longhorn-manager/blob/f6ff18666ea238c4eec4bfdce9612796da1a0701/controller/engine_image_controller.go#L363), Longhorn couldn't find the pod of the old engine image using the label selector returned by [GetEngineImageLabels()](https://github.com/longhorn/longhorn-manager/blob/f6ff18666ea238c4eec4bfdce9612796da1a0701/types/types.go#L234)  and think that the old engine image is not deployed.

The propose solution in this PR is that we don't need to add the label `longhorn.io/managed-by: longhorn-manager`  to engine image daemonset pods at all (for both old and new engine image) because the engine image ds pods are not managed by Longhorn manager. It is managed by the daemonset controller. Then, we use the correct label selector to list the engine image pods (i.e., don't use function [GetEngineImageLabels()](https://github.com/longhorn/longhorn-manager/blob/f6ff18666ea238c4eec4bfdce9612796da1a0701/types/types.go#L234) but create a new function that returns the selector that doesn't include the label `longhorn.io/managed-by: longhorn-manager`)

longhorn/longhorn#2399

### DIFF
--- a/controller/controller_test.go
+++ b/controller/controller_test.go
@@ -156,12 +156,12 @@ func newEngineImageDaemonSet() *appsv1.DaemonSet {
 		},
 		Spec: appsv1.DaemonSetSpec{
 			Selector: &metav1.LabelSelector{
-				MatchLabels: types.GetEngineImageLabels(getTestEngineImageName()),
+				MatchLabels: types.GetEIDaemonSetLabelSelector(getTestEngineImageName()),
 			},
 			Template: corev1.PodTemplateSpec{
 				ObjectMeta: metav1.ObjectMeta{
 					Name:   getTestEngineImageDaemonSetName(),
-					Labels: types.GetEngineImageLabels(getTestEngineImageName()),
+					Labels: types.GetEIDaemonSetLabelSelector(getTestEngineImageName()),
 				},
 				Spec: corev1.PodSpec{
 					ServiceAccountName: TestServiceAccount,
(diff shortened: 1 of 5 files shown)


innobead (changes requested):
LGTM, but please check if some tests needed adjust accordingly although unit tests passed in this PR.

https://github.com/longhorn/longhorn-manager/blob/master/controller/controller_test.go#L164-L164
https://github.com/longhorn/longhorn-manager/blob/master/controller/engine_image_controller_test.go#L150-L150

PhanLe1010:
Thanks @innobead !
Updated the unit tests

PhanLe1010:
I am running the e2e core tests just in case

innobead (approved):
LGTM

innobead:
> Thanks @innobead !
> Updated the unit tests

:+1:  but that also means the original unit tests do not take any effect when you change the label selection logic (it should fail in some cases). Probably we need to review the unit tests later.

PR #854, opened 2021-03-25
Ownership transferring refactoring

PhanLe1010:
1. When the manager pod on the owner node of the CR is missing, we switch
the ownerID of the CR to a different node
2. Don’t cleaning up CRs base on the current ownerID. The convention is 
that ownerID only indicates which node should watch the CR. We shouldn’t 
make any other logic depend on the ownerID. This way we can freely transfer
ownership
3. Make sure that when the engine/instance manager CR is transfered to 
new ownerID, the old owner must stop monitoring them. Also, remove 
engine/instance manager monitoring from the monitoring map of engine/instance 
manager controller when stop monitoring the engine. This allow the old
owner to restart the monitoring when the CRs’ owner is transfered back to it
4. Don’t clear the engine status when switching ownership. The new owner 
will take care of continuing to update the engine. This create a smooth
owner transferring.
5. Modify the isResponsibleFor() in engine, volume, engine image controller 
so that the current owner can continue to be the owner if the preferred 
owner doesn’t have engine image deployed.
6. Make sure the replica/engine instance is able to stop when the IM is 
in error state
7. The controller should at least look at each resource to determine
whether it's responsible for it or not, it's no the enqueue functions's
responsibility to make that determination
8. Allow attaching volume if there is engine image ready on at least one of 
the replicas' nodes of the volume.  This allow volumes to be attached/auto 
reattached when there are replicas on failed/non-exist nodes while still make
sure not to fail all of the replicas

longhorn/longhorn#2377
longhorn/longhorn#2329

### DIFF
--- a/controller/engine_controller.go
+++ b/controller/engine_controller.go
@@ -91,7 +91,7 @@ type EngineMonitor struct {
 
 	controllerID string
 	// used to notify the controller that monitoring has stopped
-	monitoringRemoveCh chan string
+	monitorVoluntaryStopCh chan struct{}
 }
 
 func NewEngineController(
@@ -235,7 +235,6 @@ func (ec *EngineController) syncEngine(key string) (err error) {
 		return err
 	}
 	if !isResponsible {
-		// Not ours
 		return nil
 	}
 	if engine.Status.OwnerID != ec.controllerID {
@@ -319,7 +318,7 @@ func (ec *EngineController) syncEngine(key string) (err error) {
 		}
 	} else if ec.isMonitoring(engine) {
 		// engine is not running
-		ec.stopMonitoring(engine)
+		ec.resetAndStopMonitoring(engine)
 	}
 
 	return nil
@@ -372,9 +371,7 @@ func (ec *EngineController) enqueueInstanceManagerChange(obj interface{}) {
 	}
 
 	for _, e := range engineMap {
-		if e.Status.OwnerID == ec.controllerID {
-			ec.enqueueEngine(e)
-		}
+		ec.enqueueEngine(e)
 	}
 	return
 }
@@ -409,6 +406,7 @@ func (ec *EngineController) DeleteInstance(obj interface{}) error {
 	if !ok {
 		return fmt.Errorf("BUG: invalid object for engine process deletion: %v", obj)
 	}
+	log := getLoggerForEngine(ec.logger, e)
 
 	if err := ec.deleteInstanceWithCLIAPIVersionOne(e); err != nil {
 		return err
@@ -425,25 +423,12 @@ func (ec *EngineController) DeleteInstance(obj interface{}) error {
 			return err
 		}
 		// The related node may be directly deleted.
-		logrus.Warnf("The instance manager %v is gone during the instance %v deletion. Will do nothing for the deletion", e.Status.InstanceManagerName, e.Name)
+		log.Warnf("The engine instance manager %v is gone during the engine instance %v deletion. Will do nothing for the deletion", e.Status.InstanceManagerName, e.Name)
 		return nil
 	}
 
-	// Node down.
-	// engine.Spec.NodeID will be unset when Longhorn prepares to stop the engine.
-	// Hence the field engine.Spec.NodeID cannot be used to check if the node is down.
-	if im.Spec.NodeID != im.Status.OwnerID {
-		isDown, err := ec.ds.IsNodeDownOrDeleted(im.Spec.NodeID)
-		if err != nil {
-			return err
-		}
-		if isDown {
-			delete(im.Status.Instances, e.Name)
-			if _, err := ec.ds.UpdateInstanceManagerStatus(im); err != nil {
-				return err
-			}
-			return nil
-		}
+	if im.Status.CurrentState != types.InstanceManagerStateRunning {
+		return nil
 	}
 
 	// For the engine process in instance manager v0.7.0, we need to use the cmdline to delete the process
@@ -590,16 +575,18 @@ func (ec *EngineController) isMonitoring(e *longhorn.Engine) bool {
 
 func (ec *EngineController) startMonitoring(e *longhorn.Engine) {
 	stopCh := make(chan struct{})
+	monitorVoluntaryStopCh := make(chan struct{})
 	monitor := &EngineMonitor{
-		logger:           ec.logger.WithField("engine", e.Name),
-		Name:             e.Name,
-		namespace:        e.Namespace,
-		ds:               ec.ds,
-		eventRecorder:    ec.eventRecorder,
-		engines:          ec.engines,
-		stopCh:           stopCh,
-		expansionBackoff: flowcontrol.NewBackOff(time.Second*10, time.Minute*5),
-		controllerID:     ec.controllerID,
+		logger:                 ec.logger.WithField("engine", e.Name),
+		Name:                   e.Name,
+		namespace:              e.Namespace,
+		ds:                     ec.ds,
+		eventRecorder:          ec.eventRecorder,
+		engines:                ec.engines,
+		stopCh:                 stopCh,
+		monitorVoluntaryStopCh: monitorVoluntaryStopCh,
+		expansionBackoff:       flowcontrol.NewBackOff(time.Second*10, time.Minute*5),
+		controllerID:           ec.controllerID,
 	}
 
 	ec.engineMonitorMutex.Lock()
@@ -612,33 +599,54 @@ func (ec *EngineController) startMonitoring(e *longhorn.Engine) {
 	ec.engineMonitorMap[e.Name] = stopCh
 
 	go monitor.Run()
+
+	go func() {
+		<-monitorVoluntaryStopCh
+		ec.engineMonitorMutex.Lock()
+		delete(ec.engineMonitorMap, e.Name)
+		ec.logger.WithField("engine", e.Name).Debug("removed the engine from ec.engineMonitorMap")
+		ec.engineMonitorMutex.Unlock()
+	}()
 }
 
-func (ec *EngineController) stopMonitoring(e *longhorn.Engine) {
+func (ec *EngineController) resetAndStopMonitoring(e *longhorn.Engine) {
 	if _, err := ec.ds.ResetMonitoringEngineStatus(e); err != nil {
 		utilruntime.HandleError(errors.Wrapf(err, "failed to update engine %v to stop monitoring", e.Name))
 		// better luck next time
 		return
 	}
 
+	ec.stopMonitoring(e.Name)
+
+	return
+}
+
+func (ec *EngineController) stopMonitoring(engineName string) {
 	ec.engineMonitorMutex.Lock()
 	defer ec.engineMonitorMutex.Unlock()
 
-	stopCh, ok := ec.engineMonitorMap[e.Name]
+	stopCh, ok := ec.engineMonitorMap[engineName]
 	if !ok {
-		ec.logger.WithField("engine", e.Name).Warn("Stop monitoring engine called even though there is no monitoring")
+		ec.logger.WithField("engine", engineName).Warn("Stop monitoring engine called even though there is no monitoring")
 		return
 	}
-	close(stopCh)
 
-	delete(ec.engineMonitorMap, e.Name)
+	select {
+	case <-stopCh:
+		// stopCh channel is already closed
+	default:
+		close(stopCh)
+	}
 
 	return
 }
 
 func (m *EngineMonitor) Run() {
 	m.logger.Debug("Start monitoring engine")
-	defer m.logger.Debug("Stop monitoring engine")
+	defer func() {
+		m.logger.Debug("Stop monitoring engine")
+		close(m.monitorVoluntaryStopCh)
+	}()
 
 	ticker := time.NewTicker(EnginePollInterval)
 	defer ticker.Stop()
@@ -654,33 +662,20 @@ func (m *EngineMonitor) Run() {
 	}
 }
 
-func (m *EngineMonitor) stop(e *longhorn.Engine) {
-	if e != nil {
-		if _, err := m.ds.ResetMonitoringEngineStatus(e); err != nil {
-			utilruntime.HandleError(errors.Wrapf(err, "failed to update engine %v to stop monitoring", m.Name))
-			// better luck next time
-			return
-		}
-	}
-}
-
 func (m *EngineMonitor) sync() bool {
 	for count := 0; count < EngineMonitorConflictRetryCount; count++ {
 		engine, err := m.ds.GetEngine(m.Name)
 		if err != nil {
 			if datastore.ErrorIsNotFound(err) {
 				m.logger.Info("stop monitoring because the engine no longer exists")
-				m.stop(engine)
 				return true
 			}
 			utilruntime.HandleError(errors.Wrapf(err, "fail to get engine %v for monitoring", m.Name))
 			return false
 		}
 
-		// when engine stopped, nodeID will be empty as well
 		if engine.Status.OwnerID != m.controllerID {
-			m.logger.Infof("stop monitoring because the engine no longer has this node as ownerID: engine.Status.OwnerID=%v; m.controllerID=%v", engine.Status.OwnerID, m.controllerID)
-			m.stop(engine)
+			m.logger.Infof("stop monitoring the engine on this node (%v) because the engine has new ownerID %v", m.controllerID, engine.Status.OwnerID)
 			return true
 		}
 
@@ -1402,33 +1397,49 @@ func (ec *EngineController) UpgradeEngineProcess(e *longhorn.Engine) error {
 	return nil
 }
 
+// isResponsibleFor picks a running node that has e.Status.CurrentImage deployed.
+// We need e.Status.CurrentImage deployed on the node to make request to the corresponding engine process.
+// Prefer picking the node e.Spec.NodeID if it meet the above requirement.
 func (ec *EngineController) isResponsibleFor(e *longhorn.Engine, defaultEngineImage string) (bool, error) {
 	var err error
 	defer func() {
 		err = errors.Wrap(err, "error while checking isResponsibleFor")
 	}()
-	readyNodesWithDefaultEI, err := ec.ds.ListReadyNodesWithEngineImage(defaultEngineImage)
-	if err != nil {
-		return false, err
-	}
 
 	isResponsible := isControllerResponsibleFor(ec.controllerID, ec.ds, e.Name, e.Spec.NodeID, e.Status.OwnerID)
 
-	if len(readyNodesWithDefaultEI) == 0 {
+	// The engine is not running, the owner node doesn't need to have e.Status.CurrentImage
+	// Fall back to the default logic where we pick a running node to be the owner
+	if e.Status.CurrentImage == "" {
 		return isResponsible, nil
 	}
 
-	preferredOwnerEngineAvailable, err := ec.ds.CheckEngineImageReadiness(defaultEngineImage, e.Spec.NodeID)
+	readyNodesWithEI, err := ec.ds.ListReadyNodesWithEngineImage(e.Status.CurrentImage)
 	if err != nil {
 		return false, err
 	}
-	currentOwnerEngineAvailable, err := ec.ds.CheckEngineImageReadiness(defaultEngineImage, e.Status.OwnerID)
+	// No node in the system has the e.Status.CurrentImage,
+	// Fall back to the default logic where we pick a running node to be the owner
+	if len(readyNodesWithEI) == 0 {
+		return isResponsible, nil
+	}
+
+	preferredOwnerEngineAvailable, err := ec.ds.CheckEngineImageReadiness(e.Status.CurrentImage, e.Spec.NodeID)
 	if err != nil {
 		return false, err
 	}
-	currentNodeEngineAvailable, err := ec.ds.CheckEngineImageReadiness(defaultEngineImage, ec.controllerID)
+	currentOwnerEngineAvailable, err := ec.ds.CheckEngineImageReadiness(e.Status.CurrentImage, e.Status.OwnerID)
 	if err != nil {
 		return false, err
 	}
-	return (isResponsible && currentNodeEngineAvailable) || (!preferredOwnerEngineAvailable && !currentOwnerEngineAvailable && currentNodeEngineAvailable), nil
+	currentNodeEngineAvailable, err := ec.ds.CheckEngineImageReadiness(e.Status.CurrentImage, ec.controllerID)
+	if err != nil {
+		return false, err
+	}
+
+	isPreferredOwner := currentNodeEngineAvailable && isResponsible
+	continueToBeOwner := currentNodeEngineAvailable && !preferredOwnerEngineAvailable && ec.controllerID == e.Status.OwnerID
+	requiresNewOwner := currentNodeEngineAvailable && !preferredOwnerEngineAvailable && !currentOwnerEngineAvailable
+
+	return isPreferredOwner || continueToBeOwner || requiresNewOwner, nil
 }
--- a/controller/engine_image_controller.go
+++ b/controller/engine_image_controller.go
@@ -473,8 +473,8 @@ func (ic *EngineImageController) getVolumesForEngineImageUpgrading(volumes map[s
 			continue
 		}
 		canBeUpgraded := ic.canDoOfflineEngineImageUpgrade(v, newEngineImageResource) || ic.canDoLiveEngineImageUpgrade(v, newEngineImageResource)
-		isCurrentEIAvailable, _ := ic.ds.CheckEngineImageReadinessForVolume(v.Status.CurrentImage, v.Name, v.Status.CurrentNodeID)
-		isNewEIAvailable, _ := ic.ds.CheckEngineImageReadinessForVolume(newEngineImageResource.Spec.Image, v.Name, v.Status.CurrentNodeID)
+		isCurrentEIAvailable, _ := ic.ds.CheckEngineImageReadyOnAllVolumeReplicas(v.Status.CurrentImage, v.Name, v.Status.CurrentNodeID)
+		isNewEIAvailable, _ := ic.ds.CheckEngineImageReadyOnAllVolumeReplicas(newEngineImageResource.Spec.Image, v.Name, v.Status.CurrentNodeID)
 		validCandidate := v.Spec.EngineImage != newEngineImageResource.Spec.Image && canBeUpgraded && isCurrentEIAvailable && isNewEIAvailable
 		if validCandidate {
 			candidates[v.Status.OwnerID] = append(candidates[v.Status.OwnerID], v)
--- a/controller/engine_image_controller.go
+++ b/controller/engine_image_controller.go
@@ -207,12 +207,10 @@ func (ic *EngineImageController) syncEngineImage(key string) (err error) {
 		return err
 	}
 	if !isResponsible {
-		// Not ours
 		return nil
 	}
 
 	if engineImage.Status.OwnerID != ic.controllerID {
-		// Claim it
 		engineImage.Status.OwnerID = ic.controllerID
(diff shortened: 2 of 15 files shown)


controller/share_manager_controller.go
```diff
@@ -287,11 +287,7 @@ func (c *ShareManagerController) syncShareManager(key string) (err error) {
 		if !c.shouldClaimOwnership(sm) {
 			return nil
 		}
-
-		// on owner ship transfer we set the state to unknown,
-		// which leads to cleaning up any of the old pods
 		sm.Status.OwnerID = c.controllerID
-		sm.Status.State = types.ShareManagerStateUnknown
```
shuo-wu, line 294: Why do we need to remove this?

controller/share_manager_controller.go
```diff
@@ -480,23 +476,26 @@ func (c *ShareManagerController) syncShareManagerVolume(sm *longhorn.ShareManage
...
+	shouldDetach := isDown && (volume.Status.State == types.VolumeStateAttached || volume.Status.State == types.VolumeStateAttaching)
+	if shouldDetach {
+		log.WithField("volume", volume.Name).Infof("Requesting Volume detach from previous node %v", volume.Status.CurrentNodeID)
+		volume.Spec.NodeID = ""
+		if volume, err = c.ds.UpdateVolume(volume); err != nil {
+			return err
 		}
+	}
 
+	if volume.Status.State == types.VolumeStateDetached {
```
shuo-wu, line 496: `&& volume.Spec.NodeID == ""`

controller/share_manager_controller.go
```diff
@@ -480,23 +476,26 @@ func (c *ShareManagerController) syncShareManagerVolume(sm *longhorn.ShareManage
 		sm.Status.State = types.ShareManagerStateStarting
 	}
 
-	if volume.Spec.NodeID != sm.Status.OwnerID || volume.Status.CurrentNodeID != sm.Status.OwnerID {
```
shuo-wu, line 483: I am not sure if this part can be replaced. Since we want to make sure sm.Status.OwnerID == volume.Spec.NodeID == volume.Status.CurrentNodeID == pod.Spec.NodeName in this controller. Checking the node down only may be not enough.

manager/engineimage.go
```diff
@@ -117,6 +117,10 @@ func (m *VolumeManager) CheckEngineImageReadiness(image string, nodes ...string)
 	return m.ds.CheckEngineImageReadiness(image, nodes...)
 }
 
-func (m *VolumeManager) CheckEngineImageReadinessForVolume(image, volumeName, nodeID string) (isReady bool, err error) {
```
shuo-wu, line 120: Do we need these? I think you can directly call `m.ds.CheckEngineImageReadyOnAllVolumeReplicas`

shuo-wu (changes requested):
1. I have some doubts about function`isResponsibleFor()`:
    1. For the engine controller, it seems that we need to verify if the image the engine is using is on the node. Currently this function will blindly verify the default image.
    2. If the preferred owner has the ready engine image, can the current owner actively give up the ownership?
    3. It's better to list the strategy with comments. It's harder to understand this part...

controller/share_manager_controller.go
PhanLe1010: We don't want to clean up the IM pod when transfer ownership to a new node. Doing so would disrupt the current running workloads.

controller/share_manager_controller.go
PhanLe1010: IMHO,  we only need `volume.Spec.NodeID` == `pod.Spec.NodeName` so that the share manager pod can access the underlying Longhorn volume on the same node. The `sm.Status.OwnerID` and `volume.Status.CurrentNodeID` should be designed so that they can be moved freely to any controller node without disrupting the current running workloads that use the RWX volume. How do you think?
cc @joshimoo

manager/engineimage.go
PhanLe1010: You are right.  We don't need them. Fixed. Thank you

controller/share_manager_controller.go
PhanLe1010: Thank you. Fixed

controller/engine_controller.go
```diff
@@ -590,16 +596,18 @@ func (ec *EngineController) isMonitoring(e *longhorn.Engine) bool {
...
-		logger:           ec.logger.WithField("engine", e.Name),
-		Name:             e.Name,
-		namespace:        e.Namespace,
-		ds:               ec.ds,
-		eventRecorder:    ec.eventRecorder,
-		engines:          ec.engines,
-		stopCh:           stopCh,
-		expansionBackoff: flowcontrol.NewBackOff(time.Second*10, time.Minute*5),
-		controllerID:     ec.controllerID,
+		logger:             ec.logger.WithField("engine", e.Name),
```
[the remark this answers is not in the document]
PhanLe1010: Can you explain why we need the to add `monitorNodeID`? Since the `ec.logger` already has the `controllerID`

controller/engine_controller.go
```diff
@@ -612,33 +620,54 @@ func (ec *EngineController) startMonitoring(e *longhorn.Engine) {
...
 
 	go monitor.Run()
+
+	go func() {
+		<-monitoringRemoveCh
+		ec.engineMonitorMutex.Lock()
+		delete(ec.engineMonitorMap, e.Name)
+		ec.logger.WithField("engine", e.Name).Debug("removed the engine from ec.engineMonitorMap")
+		ec.engineMonitorMutex.Unlock()
+	}()
```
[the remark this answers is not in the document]
PhanLe1010: My current solution for engine monitoring is: 
There are 3 goroutines: the main sync goroutine, the monitoring goroutine, and the stop monitoring goroutine. 
There are 2 ways to stop monitoring an engine:
1.  The main sync goroutine closes the stopCh -> the monitoring goroutine closes the monitoringRemoveCh -> the stop monitoring goroutine removes the engine RS from the `ec.engineMonitorMap` -> Done
1. The monitoring goroutine volunteer to stop monitoring engine, it closes the monitoringRemoveCh -> the stop monitoring goroutine removes the engine RS from the `ec.engineMonitorMap` -> Done

Your suggested solution for backing image monitoring is:
There are 3 goroutines: the main sync goroutine, the monitoring goroutine, and  the [notification update](https://github.com/shuo-wu/longhorn-manager/blob/35f06eba89ae87e0ad8e67b245b74c1ee2e1ac4c/controller/backing_image_manager_controller.go#L1093) goroutine
There are 2 ways to stop monitoring a backing image:
1.  The main sync goroutine closes the stopCh and removes the backing image RS from the `ec.engineMonitorMap` -> the monitoring goroutine returns ->  the notification update goroutine returns -> Done
1. The notification update goroutine volunteers to stop monitoring backing image CR, it closes the stopCh and  removes the engine RS from the `ec.engineMonitorMap`-> the monitoring goroutine returns -> Done

The suggested solution wouldn't work if the monitoring goroutine wants to volunteer to stop monitoring the backing image. For example, if something is wrong inside the monitoring goroutine while it does [pollAndUpdateBackingImageFileMap](https://github.com/shuo-wu/longhorn-manager/blob/35f06eba89ae87e0ad8e67b245b74c1ee2e1ac4c/controller/backing_image_manager_controller.go#L1147) (i.e., the backing image's owner has moved to a new node and doesn't need to be monitored on this node), the monitoring goroutine cannot volunteer to stop monitoring the backing image. If we try to call `m.monitorStopper(m.Name)` inside `pollAndUpdateBackingImageFileMap()`, it will be blocked forever because we send [stopCH <- struct{}{}](https://github.com/shuo-wu/longhorn-manager/blob/35f06eba89ae87e0ad8e67b245b74c1ee2e1ac4c/controller/backing_image_manager_controller.go#L1066) but no goroutine would take it out.

For engine monitoring, we need the stop monitoring goroutine to be able to volunteer to stop monitoring the engine if something is wrong while doing [sync](https://github.com/longhorn/longhorn-manager/blob/f12923c00c2aab6a21b52bf311393337814902c9/controller/engine_controller.go#L686). Therefore, I think the current implementation is ok.

How do you think?

controller/engine_controller.go
```diff
@@ -474,6 +465,21 @@ func (ec *EngineController) DeleteInstance(obj interface{}) error {
...
+	// Node down.
+	// engine.Spec.NodeID will be unset when Longhorn prepares to stop the engine.
+	// Hence the field engine.Spec.NodeID cannot be used to check if the node is down.
+	isDown, err := ec.ds.IsNodeDownOrDeleted(im.Spec.NodeID)
+	if err != nil {
+		log.WithError(err).Warnf("cannot check IsNodeDownOrDeleted(%v) when checking canForceDeleteInstanceFromInstanceManager", im.Spec.NodeID)
+	}
+	return isDown
+}
+
```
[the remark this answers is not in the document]
PhanLe1010: Thank you! Fixed

controller/engine_controller.go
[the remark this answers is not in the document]
shuo-wu: Oh, sorry. Then we don't need it

controller/engine_controller.go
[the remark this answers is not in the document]
shuo-wu: Make sense. As a receiver, the monitor goroutine cannot send a stop signal.

controller/engine_controller.go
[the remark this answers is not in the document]
shuo-wu: The map cleanup for the not running instance managers is missed.

controller/share_manager_controller.go
shuo-wu: I am confused about the share manager state and the node ID issue now:
1. Can `sm.status.ownerID` be different from `volume.Spec.NodeID == pod.Spec.NodeName` while the related RWX volume is running?
2. Then when will the share manager become state unknown?

shuo-wu:
> I have some doubts about function`isResponsibleFor()`:
> 
> 1. For the engine controller, it seems that we need to verify if the image the engine is using is on the node. Currently this function will blindly verify the default image.
> 2. If the preferred owner has the ready engine image, can the current owner actively give up the ownership?
> 3. It's better to list the strategy with comments. It's harder to understand this part...

You may miss this comment

controller/engine_controller.go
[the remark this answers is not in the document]
shuo-wu: And I may need a similar logic in the backing image manager controller to actively stop monitoring in case of ownerID suddenly change.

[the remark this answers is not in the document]
shuo-wu: BTW, can you rename it to avoid misunderstanding? e.g., `monitorVoluntaryStopCh`

PhanLe1010:
Regard to https://github.com/longhorn/longhorn-manager/pull/854#pullrequestreview-624859923
> For the engine controller, it seems that we need to verify if the image the engine is using is on the node. Currently this function will blindly verify the default image.

Good catch! Thank you. I fixed it

> If the preferred owner has the ready engine image, can the current owner actively give up the ownership?

Yes. From the condition
```
(currentNodeEngineAvailable && isResponsible) ||
		(currentNodeEngineAvailable && !preferredOwnerEngineAvailable && ec.controllerID == e.Status.OwnerID) ||
		(currentNodeEngineAvailable && !preferredOwnerEngineAvailable && !currentOwnerEngineAvailable), nil
```
When the preferred owner has the ready engine image, on the current owner node, we will have:
* `isResponsible` is `false` because the preferred owner is up and running
*  `!preferredOwnerEngineAvailable ` is `false` since the preferred owner has the engine image
So the function return `false` and the current owner gives up the ownership

> It's better to list the strategy with comments. It's harder to understand this part...

I added some comments. Thank you. Yeah, the logic here is complicated because this function must make sure that it will output one and only one node when it is running on any node at any given time.

controller/share_manager_controller.go
PhanLe1010: > Can sm.status.ownerID be different from volume.Spec.NodeID == pod.Spec.NodeName while the related RWX volume is running?

Yes, it can from the code perspective. I have also tested the scenario 

> Then when will the share manager become state unknown?

It never becomes unknown from the code perspective.  We also mark it as `error` when things went wrong. I think it is ok.

controller/engine_controller.go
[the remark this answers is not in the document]
PhanLe1010: Thank you! I changed `monitoringRemoveCh ` to `monitorVoluntaryStopCh `

controller/instance_manager_controller.go
```diff
@@ -272,11 +276,14 @@ func (imc *InstanceManagerController) syncInstanceManager(key string) (err error
 		return err
 	}
 
+	log := imc.logger.WithField("instance manager", im.Name)
```
joshimoo, line 279: you can use `getLoggerForInstanceManager`

controller/engine_controller.go
```diff
@@ -409,6 +407,7 @@ func (ec *EngineController) DeleteInstance(obj interface{}) error {
 	if !ok {
 		return fmt.Errorf("BUG: invalid object for engine process deletion: %v", obj)
 	}
+	log := ec.logger.WithField("engine", e.Name)
```
joshimoo, line 410: You can use `getLoggerForEngine`

controller/instance_manager_controller.go
```diff
@@ -738,8 +750,8 @@ func (imc *InstanceManagerController) enqueueKubernetesNode(obj interface{}) {
 }
 
 func (imc *InstanceManagerController) cleanupInstanceManager(im *longhorn.InstanceManager) error {
+	log := imc.logger.WithField("instance manager", im.Name)
```
joshimoo, line 753: use `getLoggerForInstanceManager`

controller/instance_manager_controller.go
```diff
@@ -1118,8 +1138,8 @@ func (m *InstanceManagerMonitor) Run() {
 			m.lock.Unlock()
 
 			if needUpdate {
```
joshimoo, line 1140: consider inverting this `if !needUpdate { continue }`
reduces nesting.

controller/kubernetes_configmap_controller.go
```diff
@@ -32,8 +32,8 @@ const (
 type KubernetesConfigMapController struct {
 	*baseController
 
+	namespace string
 	// use as the OwnerID of the controller
```
joshimoo, line 36: just remove the comment, doesn't really add anything :)

controller/replica_controller.go
```diff
@@ -273,11 +273,11 @@ func (rc *ReplicaController) syncReplica(key string) (err error) {
 
 	log := getLoggerForReplica(rc.logger, replica)
 
+	if !rc.isResponsibleFor(replica) {
+		// Not ours
```
joshimoo, line 277: consider removing the comment, doesn't really add anything that is not expressed by.
We are not responsible for this :)

datastore/longhorn.go
```diff
@@ -967,8 +967,28 @@ func (s *DataStore) CheckEngineImageReadiness(image string, nodes ...string) (is
...
+// CheckEngineImageReadyOnAtLeastOneVolumeReplica checks if the IMAGE is deployed on the NODEID and on at least one of the the volume's replicas
+func (s *DataStore) CheckEngineImageReadyOnAtLeastOneVolumeReplica(image, volumeName, nodeID string) (isReady bool, err error) {
+	replicas, err := s.ListVolumeReplicas(volumeName)
+	if err != nil {
+		return false, fmt.Errorf("cannot get replicas for volume %v: %w", volumeName, err)
+	}
+
+	isReady, err = s.CheckEngineImageReadiness(image, nodeID)
+	if nodeID != "" && !isReady {
+		return
```
joshimoo, line 979: I feel like it's better to specify the return values explicitly instead of relying on the default values of the named params. In this case you also probably don't need the named params, since it's obvious from the usage :)

datastore/longhorn.go
```diff
@@ -967,8 +967,28 @@ func (s *DataStore) CheckEngineImageReadiness(image string, nodes ...string) (is
...
+		isReady, err = s.CheckEngineImageReadiness(image, r.Spec.NodeID)
+		if err != nil || isReady {
+			return
+		}
+	}
+	return false, nil
+}
+
+// CheckEngineImageReadyOnAllVolumeReplicas checks if the IMAGE is deployed on the NODEID as well as all the volume's replicas
+func (s *DataStore) CheckEngineImageReadyOnAllVolumeReplicas(image, volumeName, nodeID string) (isReady bool, err error) {
```
joshimoo, line 991: Consider whether you need the named params.

manager/volume.go
```diff
@@ -315,11 +315,11 @@ func (m *VolumeManager) Attach(name, nodeID string, disableFrontend bool, attach
 		return nil, err
 	}
 
-	if isReady, err := m.CheckEngineImageReadinessForVolume(v.Spec.EngineImage, v.Name, nodeID); !isReady {
+	if isReady, err := m.ds.CheckEngineImageReadyOnAtLeastOneVolumeReplica(v.Spec.EngineImage, v.Name, nodeID); !isReady {
 		if err != nil {
 			return nil, fmt.Errorf("cannot attach volume %v with image %v: %v", v.Name, v.Spec.EngineImage, err)
 		}
-		return nil, fmt.Errorf("cannot attach volume %v with image %v because the engine image %v is not deployed on the replicas' nodes or the node that the volume is going to attach to", v.Name, v.Spec.EngineImage, v.Spec.EngineImage)
+		return nil, fmt.Errorf("cannot attach volume %v with image %v because the engine image %v is not deployed on at least one of the the replicas' nodes or the node that the volume is going to attach to", v.Name, v.Spec.EngineImage, v.Spec.EngineImage)
```
joshimoo, line 322:
```suggestion
		return nil, fmt.Errorf("cannot attach volume %v because the engine image %v is not deployed on at least one of the the replicas' nodes or the node that the volume is going to attach to", v.Name, v.Spec.EngineImage)
```

controller/volume_controller.go
```diff
@@ -2549,7 +2551,9 @@ func (vc *VolumeController) isResponsibleFor(v *longhorn.Volume, defaultEngineIm
 		return false, err
 	}
 
-	return (isResponsible && currentNodeEngineAvailable) || (!preferredOwnerEngineAvailable && !currentOwnerEngineAvailable && currentNodeEngineAvailable), nil
+	return (currentNodeEngineAvailable && isResponsible) ||
```
joshimoo, line 2554: You might want to use variables like this, similar to the ControllerIsResponsibleFor, might make it easier to digest what's going on :)

```
	isPreferredOwner := currentNodeEngineAvailable && isResponsible
	continueToBeOwner := currentNodeEngineAvailable && !preferredOwnerEngineAvailable && vc.controllerID == v.Status.OwnerID
	requiresNewOwner := currentNodeEngineAvailable && !preferredOwnerEngineAvailable && !currentOwnerEngineAvailable
	return isPreferredOwner || continueToBeOwner || requiresNewOwner
```

joshimoo (commented):
Some feedback below, I will now have a deeper look at the engine_controller, instance_manager_controller, share_manager_controller will submit another review for that part.

joshimoo:
Hey, can you take a look at the following code to see how I use a cancellable context for monitoring.
And see if this might be a better implementation instead of the raw stopChannel + voluntaryStopChannel?
The actual code details are not important for the example.

link to the relevant places below:
- https://github.com/joshimoo/longhorn-manager/blob/aa8f26dd0d18d55b68e10dde5fb2ad2045ac56c2/shareapi/monitor.go#L42 (cancellable context)
- https://github.com/joshimoo/longhorn-manager/blob/aa8f26dd0d18d55b68e10dde5fb2ad2045ac56c2/shareapi/monitor.go#L65-L66 (2 go routines)
- https://github.com/joshimoo/longhorn-manager/blob/aa8f26dd0d18d55b68e10dde5fb2ad2045ac56c2/shareapi/monitor.go#L70-L71 (close function that can be called externally, that calls the cancel function)

context evaluation:
- https://github.com/joshimoo/longhorn-manager/blob/aa8f26dd0d18d55b68e10dde5fb2ad2045ac56c2/shareapi/monitor.go#L101
- https://github.com/joshimoo/longhorn-manager/blob/aa8f26dd0d18d55b68e10dde5fb2ad2045ac56c2/shareapi/monitor.go#L123

Have a look at the enable/disable/hasMonitor functions (controller)
- https://github.com/joshimoo/longhorn-manager/blob/0fde5608ea2985829b1e1a5e2a82b22aad37c03f/controller/share_manager_controller.go#L35-L68

Usage of the monitor (this one is not important since it might require to much changes to the monitor for v1.1.1)
In General our monitoring is kinda flawed, since we have the reconcile loop directly call a grpc service instead of the monitor keeping state and ensuring a requeue of the elements
If you are curious how monitoring should work, have a look at the syncCallback usage :slightly_smiling_face:
- https://github.com/joshimoo/longhorn-manager/blob/0fde5608ea2985829b1e1a5e2a82b22aad37c03f/controller/share_manager_controller.go#L415

controller/share_manager_controller.go
```diff
@@ -777,7 +779,15 @@ func (c *ShareManagerController) createPodManifest(sm *longhorn.ShareManager, an
...
+	preferredOwnerID := ""
+	pod, err := c.ds.GetPod(types.GetShareManagerPodNameFromShareManagerName(sm.Name))
+	if err == nil && pod != nil {
+		preferredOwnerID = pod.Spec.NodeName
+	}
+
+	shouldClaim := isControllerResponsibleFor(c.controllerID, c.ds, sm.Name, preferredOwnerID, sm.Status.OwnerID)
 	isCurrentNodeSchedulable := isNodeSchedulable(c.ds, c.controllerID)
 	if shouldClaim {
 		return isCurrentNodeSchedulable
```
joshimoo, line 793: isCurrentNodeSchedulable is only relevant, if the pod currently does not exists (share-manager-state == error), since you use the pod.spec.NodeName as preferred owner. If the pod already exists the schedulable status is irrelevant.

controller/share_manager_controller.go
```diff
@@ -777,7 +779,15 @@ func (c *ShareManagerController) createPodManifest(sm *longhorn.ShareManager, an
 // this scenario would lead to the node no longer being able ot be drained
 // since the volume would never be unattached which prevents the engine from being removed
 func (c *ShareManagerController) shouldClaimOwnership(sm *longhorn.ShareManager) bool {
```
joshimoo, line 781: You can rename this to isResponsibleFor same as the other controllers, since we are now doing the same more or less :)

controller/share_manager_controller.go
joshimoo: Using status.Owner as condition was a simplification for volume failure handling.
It's not strictly required that the owner for the crd is the same node as the actual pod.
So this change is fine, but comes with the caveat that it adds a couple more isNodeDown checks  :)

But overall I agree, it's okay to decouple CRD processing from pod/volume ownership.

joshimoo (commented):
Left some notes about the share-manager-controller, I think your changes in that regard are okay.
Can you check on my comment in regard to the monitoring with cancellable context to see if that can simplify/make the monitoring more robust?

PhanLe1010:
Regarding to https://github.com/longhorn/longhorn-manager/pull/854#issuecomment-812101940

> Usage of the monitor (this one is not important since it might require to much changes to the monitor for v1.1.1)
In General our monitoring is kinda suboptimal, since we have the reconcile loop directly call a grpc service instead of the monitor keeping state and ensuring a requeue of the elements
If you are curious how monitoring could be improved, have a look at the syncCallback usage

Thank you for the suggestion. I agree that it is a better design and easier to read. Yeah, it would require  a big change, we can consider refactoring the monitoring logic of the engine, instance manager in a different issue. I will create an issue to track the effort. 

> And see if this might be a better implementation instead of the raw stopChannel + voluntaryStopChannel?

Until we do the above refactoring, I don't see how the cancellable context will help to make the code cleaner comparing to the raw stopChannel + voluntaryStopChannel. I think we can wait until we do refactoring and use the cancellable context.

How do you think?

controller/engine_image_controller.go
```diff
@@ -821,5 +813,7 @@ func (ic *EngineImageController) isResponsibleFor(ei *longhorn.EngineImage) (boo
 		return false, err
 	}
 
-	return (isResponsible && currentNodeEngineAvailable) || (!currentOwnerEngineAvailable && currentNodeEngineAvailable), nil
+	return (currentNodeEngineAvailable && isResponsible) ||
+		(currentNodeEngineAvailable && ic.controllerID == ei.Status.OwnerID) ||
```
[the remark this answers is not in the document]
PhanLe1010: You are right. This change is not logically necessary. I just want to make the code format to be similar to other controllers.

controller/instance_manager_controller.go
PhanLe1010: Thank you. Fixed

controller/instance_manager_controller.go
PhanLe1010: Thank you. Fixed

controller/instance_manager_controller.go
PhanLe1010: Thank you. Fixed

controller/kubernetes_configmap_controller.go
PhanLe1010: Thank you. Fixed

controller/replica_controller.go
PhanLe1010: Thank you. Fixed

controller/share_manager_controller.go
PhanLe1010: Thank you. Fixed

controller/share_manager_controller.go
PhanLe1010: I see. But I think it is still ok that we switch to a schedulable node even if the pod already exists because:
1. Ownership transferring has a very small cost now
2. The next time the sm stops, it will have to switch to the new node anyways
How do you think?

controller/volume_controller.go
PhanLe1010: Thank you. Fixed

datastore/longhorn.go
PhanLe1010: Thank you. Fixed

datastore/longhorn.go
PhanLe1010: Thank you. Removed

controller/engine_controller.go
PhanLe1010: Thank you. Fixed

manager/volume.go
PhanLe1010: Thank you. Fixed

controller/engine_controller.go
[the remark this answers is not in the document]
shuo-wu: OK, I see

controller/instance_manager_controller.go
```diff
@@ -272,11 +276,14 @@ func (imc *InstanceManagerController) syncInstanceManager(key string) (err error
 		return err
 	}
 
+	log := getLoggerForInstanceManager(imc.logger, im)
+
+	if !imc.isResponsibleFor(im) {
+		// Not ours
```
joshimoo, line 282: don't need the comment.

controller/engine_image_controller.go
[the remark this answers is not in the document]
joshimoo: Sounds reasonable, do you want to use variables like we do for the other controllers.
isPreferredOwner, continueToBeOwner, requiresNewOwner 
I find with the named variables it makes the intentions/conditions very easy to understand.

joshimoo (approved):
LGTM once the nits are addressed this can be merged.
I will let @shuo-wu merge it unless I receive a notification that it's good to merge from him/phan.

controller/share_manager_controller.go
PhanLe1010: Thank you. Fixed

controller/instance_manager_controller.go
PhanLe1010: Thank you. Fixed

innobead (approved):
LGTM

Any other concerns @joshimoo @shuo-wu? Otherwise we can merge

shuo-wu (approved):

PR #856, opened 2021-03-30
update unsintall controller to cleanup share manager crds (#2384)

cclhsu:
#### Proposed Changes ####

update unsinstall controller to cleanup share manager crds to resolve ticket '[BUG] uninstall-controller does not cleanup share-manager crds' (#2384)

- add clean up for share-manager and settings

#### Types of Changes ####

Bug

#### Verification ####

- deploy longhorn
- deploy a rxw volume/pod
- uninstall using uninstaller from "https://raw.githubusercontent.com/longhorn/longhorn-manager/master/deploy/uninstall/uninstall.yaml"
- check log for uninstaller

```
ttime="2021-04-01T06:49:07Z" level=info msg="Found 1 share managers remaining" controller=longhorn-uninstall
time="2021-04-01T06:49:07Z" level=info msg="Marked for deletion" controller=longhorn-uninstall owner=centos-worker-0 shareManager=pvc-d6e490b0-0437-481c-ba0f-c6d8944757f4 state=running volume=pvc-d6e490b0-0437-481c-ba0f-c6d8944757f4
time="2021-04-01T06:49:07Z" level=info msg="Found 1 share managers remaining" controller=longhorn-uninstall
time="2021-04-01T06:49:07Z" level=info msg="Marked for deletion" controller=longhorn-uninstall owner= shareManager=pvc-d6e490b0-0437-481c-ba0f-c6d8944757f4 state= volume=pvc-d6e490b0-0437-481c-ba0f-c6d8944757f4
time="2021-04-01T06:49:07Z" level=info msg="Found 1 share managers remaining" controller=longhorn-uninstall
time="2021-04-01T06:49:07Z" level=info msg="Found 1 share managers remaining" controller=longhorn-uninstall
```

#### Linked Issues ####

Refs: # https://github.com/longhorn/longhorn/issues/2384

#### Further Comments ####

None

### DIFF
--- a/controller/uninstall_controller.go
+++ b/controller/uninstall_controller.go
@@ -35,6 +35,7 @@ const (
 	CRDEngineImageName     = "engineimages.longhorn.io"
 	CRDNodeName            = "nodes.longhorn.io"
 	CRDInstanceManagerName = "instancemanagers.longhorn.io"
+	CRDShareManagerName    = "sharemanagers.longhorn.io"
 	CRDBackingImageName    = "backingimages.longhorn.io"
 
 	LonghornNamespace = "longhorn-system"
@@ -67,6 +68,7 @@ func NewUninstallController(
 	engineImageInformer lhinformers.EngineImageInformer,
 	nodeInformer lhinformers.NodeInformer,
 	imInformer lhinformers.InstanceManagerInformer,
+	smInformer lhinformers.ShareManagerInformer,
 	backingImageInformer lhinformers.BackingImageInformer,
 	daemonSetInformer appsv1.DaemonSetInformer,
 	deploymentInformer appsv1.DeploymentInformer,
@@ -117,6 +119,10 @@ func NewUninstallController(
 	if _, err := extensionsClient.ApiextensionsV1beta1().CustomResourceDefinitions().Get(CRDInstanceManagerName, metav1.GetOptions{}); err == nil {
 		imInformer.Informer().AddEventHandler(c.controlleeHandler())
 		cacheSyncs = append(cacheSyncs, imInformer.Informer().HasSynced)
+	}
+	if _, err := extensionsClient.ApiextensionsV1beta1().CustomResourceDefinitions().Get(CRDShareManagerName, metav1.GetOptions{}); err == nil {
+		smInformer.Informer().AddEventHandler(c.controlleeHandler())
+		cacheSyncs = append(cacheSyncs, smInformer.Informer().HasSynced)
 	}
 	if _, err := extensionsClient.ApiextensionsV1beta1().CustomResourceDefinitions().Get(CRDBackingImageName, metav1.GetOptions{}); err == nil {
 		backingImageInformer.Informer().AddEventHandler(c.controlleeHandler())
@@ -319,6 +325,13 @@ func (c *UninstallController) deleteLease() error {
 }
 
 func (c *UninstallController) deleteCRDs() (bool, error) {
+	if shareManagers, err := c.ds.ListShareManagers(); err != nil {
+		return true, err
+	} else if len(shareManagers) > 0 {
+		c.logger.Infof("Found %d share managers remaining", len(shareManagers))
+		return true, c.deleteShareManagers(shareManagers)
+	}
+
 	if volumes, err := c.ds.ListVolumes(); err != nil {
 		return true, err
 	} else if len(volumes) > 0 {
@@ -532,6 +545,37 @@ func (c *UninstallController) deleteInstanceManagers(instanceManagers map[string
 			}
 
 			if err = c.ds.RemoveFinalizerForInstanceManager(im); err != nil {
+				err = errors.Wrapf(err, "Failed to remove finalizer")
+				return
+			}
+			log.Info("Removed finalizer")
+		}
+	}
+	return
+}
+
+func (c *UninstallController) deleteShareManagers(shareManagers map[string]*longhorn.ShareManager) (err error) {
+	defer func() {
+		err = errors.Wrapf(err, "Failed to delete share managers")
+	}()
+	for _, sm := range shareManagers {
+		log := getLoggerForShareManager(c.logger, sm)
+
+		timeout := metav1.NewTime(time.Now().Add(-gracePeriod))
+		if sm.DeletionTimestamp == nil {
+			if err = c.ds.DeleteShareManager(sm.Name); err != nil {
+				err = errors.Wrapf(err, "Failed to mark for deletion")
+				return
+			}
+			log.Info("Marked for deletion")
+		} else if sm.DeletionTimestamp.Before(&timeout) {
+			podName := types.GetShareManagerPodNameFromShareManagerName(sm.Name)
+			if err = c.ds.DeletePod(podName); err != nil {
+				return err
+			}
+			log.Infof("Removing share manager related pod %v", podName)
+
+			if err = c.ds.RemoveFinalizerForShareManager(sm); err != nil {
 				err = errors.Wrapf(err, "Failed to remove finalizer")
 				return
 			}
--- a/app/uninstall.go
+++ b/app/uninstall.go
@@ -132,6 +132,7 @@ func uninstall(c *cli.Context) error {
 		engineImageInformer,
 		nodeInformer,
 		imInformer,
+		smInformer,
 		backingImageInformer,
 		daemonSetInformer,
 		deploymentInformer,


controller/uninstall_controller.go
```diff
@@ -35,7 +35,9 @@ const (
 	CRDEngineImageName     = "engineimages.longhorn.io"
 	CRDNodeName            = "nodes.longhorn.io"
 	CRDInstanceManagerName = "instancemanagers.longhorn.io"
+	CRDShareManagerName    = "sharemanagers.longhorn.io"
 	CRDBackingImageName    = "backingimages.longhorn.io"
+	CRDSettingName         = "settings.longhorn.io"
```
innobead, line 40: I think we don't need to handle `settings` resources for now because in the current implementation, we don't need any special logic for handling the settings cleanup via finalizer, but other resources all managed by finalizer to make sure we can clean up the corresponding pods correctly.

But, I don't think we need to have that for `node` as well, but this was implemented already. Maybe we can clean up uninstaller in near future, also many duplicated code.

controller/setting_controller.go
```diff
@@ -140,6 +140,14 @@ func NewSettingController(
 	return sc
 }
 
+func getLoggerForSetting(logger logrus.FieldLogger, setting *longhorn.Setting) *logrus.Entry {
```
innobead, line 143: No need, because we don't need to handle setting resource uninstall. Just leave it to the normal resource clean up via CRD removing.

controller/uninstall_controller.go
```diff
@@ -368,6 +387,13 @@ func (c *UninstallController) deleteCRDs() (bool, error) {
 		return true, c.deleteNodes(nodes)
 	}
 
+	if settings, err := c.ds.ListSettings(); err != nil {
```
innobead, line 390: The same, no need for settings removal handling.

controller/uninstall_controller.go
```diff
@@ -480,6 +506,30 @@ func (c *UninstallController) deleteEngineImages(engineImages map[string]*longho
 	return
 }
 
+func (c *UninstallController) deleteSettings(settings map[types.SettingName]*longhorn.Setting) (err error) {
```
innobead, line 509: No need.

controller/uninstall_controller.go
```diff
@@ -541,6 +591,43 @@ func (c *UninstallController) deleteInstanceManagers(instanceManagers map[string
...
+		timeout := metav1.NewTime(time.Now().Add(-gracePeriod))
+		if sm.DeletionTimestamp == nil {
+			if err = c.ds.DeleteShareManager(sm.Name); err != nil {
+				err = errors.Wrapf(err, "Failed to mark for deletion")
+				return
+			}
+			log.Info("Marked for deletion")
+		} else if sm.DeletionTimestamp.Before(&timeout) {
+			var pod *v1.Pod
+			pod, err = c.ds.GetPod(sm.Name)
```
innobead, line 610: I don't think the pod name is `sm.Name`, can you check again and verify if the share managed pods actually deleted by creating a RWX volume.

https://github.com/longhorn/longhorn-manager/blob/master/controller/share_manager_controller.go#L703-L703

innobead (changes requested):
Some comments to resolve/fix.

controller/uninstall_controller.go
joshimoo: One can use the function `types.GetShareManagerPodNameFromShareManager` to map a share-manager name to a pod name. See https://github.com/longhorn/longhorn-manager/blob/3b83f478cdcaf70176127e73510d91ae37a9301c/types/types.go#L355-L361

controller/uninstall_controller.go
```diff
@@ -541,6 +591,43 @@ func (c *UninstallController) deleteInstanceManagers(instanceManagers map[string
...
+			if err != nil {
+				err = errors.Wrapf(err, "Could not get pod for share manager")
+			}
+			if pod != nil {
+				if err = c.ds.DeletePod(pod.Name); err != nil {
```
joshimoo, line 615: I think you can just call delete without initial retrieval of the pod?
```
podName := types.GetShareManagerPodNameFromShareManagerName(sm.Name)`
c.ds.DeletePod(podName)
```

joshimoo (commented):
The reason settings do not require the extra cleanup step is because they don't have a finalizer set. So they can be deleted directly when calling `kubectl delete -f` 

See below:
Since settings don't have a finalizer set the can be deleted directly.
`kubectl get settings system-managed-pods-image-pull-policy -o yaml -n longhorn-system`

```
apiVersion: longhorn.io/v1beta1
kind: Setting
metadata:
  creationTimestamp: "2021-02-03T12:52:35Z"
  generation: 1
  name: system-managed-pods-image-pull-policy
  namespace: longhorn-system
  resourceVersion: "8386"
  selfLink: /apis/longhorn.io/v1beta1/namespaces/longhorn-system/settings/system-managed-pods-image-pull-policy
  uid: 4fe5023c-5133-49c1-8eb6-d577a87c6394
value: if-not-present
```

controller/uninstall_controller.go
cclhsu: removed

controller/setting_controller.go
cclhsu: removed

controller/uninstall_controller.go
cclhsu: removed

controller/uninstall_controller.go
```diff
@@ -541,6 +591,43 @@ func (c *UninstallController) deleteInstanceManagers(instanceManagers map[string
...
 }
 
+func (c *UninstallController) deleteShareManagers(shareManagers map[string]*longhorn.ShareManager) (err error) {
+	defer func() {
+		err = errors.Wrapf(err, "Failed to delete share managers")
+	}()
+	for _, sm := range shareManagers {
+		log := getLoggerForShareManager(c.logger, sm)
+
+		timeout := metav1.NewTime(time.Now().Add(-gracePeriod))
```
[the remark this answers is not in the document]
cclhsu: I think this is related to race condition in uninstall() which previous resolved.

controller/uninstall_controller.go
cclhsu: changed

controller/uninstall_controller.go
cclhsu: changed

cclhsu:
I have updated according to the reviews and updated the verification steps.

controller/uninstall_controller.go
```diff
@@ -541,6 +554,33 @@ func (c *UninstallController) deleteInstanceManagers(instanceManagers map[string
...
+		if sm.DeletionTimestamp == nil {
+			if err = c.ds.DeleteShareManager(sm.Name); err != nil {
+				return errors.Wrapf(err, "Failed to mark for deletion")
+			}
+			log.Info("Marked for deletion")
+		} else if sm.DeletionTimestamp.Before(&timeout) {
+			podName := types.GetShareManagerPodNameFromShareManagerName(sm.Name)
+			c.ds.DeletePod(podName)
```
innobead, line 572: No error handling.

innobead (changes requested):

controller/uninstall_controller.go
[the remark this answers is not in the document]
innobead: > I think this is related to race condition in uninstall() which previous resolved.

What kind of race condition do u mean? I think it is just in the control loop and each iteration from the queue is sequential.

The timeout seems for the specific delete hanging of share manager related resources or clean up.

I saw 30 sec everywhere for different resource deletion, maybe we need to review again if needed. Too many duplication with some assumption.

controller/uninstall_controller.go
```diff
@@ -541,6 +554,35 @@ func (c *UninstallController) deleteInstanceManagers(instanceManagers map[string
...
+			if err = c.ds.DeleteShareManager(sm.Name); err != nil {
+				return errors.Wrapf(err, "Failed to mark for deletion")
+			}
+			log.Info("Marked for deletion")
+		} else if sm.DeletionTimestamp.Before(&timeout) {
+			podName := types.GetShareManagerPodNameFromShareManagerName(sm.Name)
+			if err = c.ds.DeletePod(podName); err != nil {
+				return err
+			}
+			log.Infof("Removing backing image related pod %v", podName)
```
innobead, line 575: Share manager pod not backing image

innobead (changes requested):

controller/uninstall_controller.go
cclhsu: changed

controller/uninstall_controller.go
cclhsu: udpated

innobead (approved):
LGTM

But we need to revisit the uninstall process n clean up a bit in the future

controller/uninstall_controller.go
```diff
@@ -319,6 +325,13 @@ func (c *UninstallController) deleteLease() error {
 }
 
 func (c *UninstallController) deleteCRDs() (bool, error) {
+	if shareManagers, err := c.ds.ListShareManagers(); err != nil {
```
joshimoo, line 328: Previously this was at the bottom of this function, and should be moved back there. 
Since it's preferable to cleanup the volumes first. otherwise the volume controller will try to recreate a share-manager.

innobead: A good suggestion! @cclhsu please help improve this part. Thanks!

PR #857, opened 2021-03-30
Refactor volume controller to change volume attachment handling

joshimoo:
The behavior is changed so that attachment is handled solely based on
v.Spec.NodeID instead of handling attachment internally via setting
v.Status.CurrentNode or v.Status.PendingNodeID

This requires dedicated controllers for operations that previously relied
on the internal attachment, the required controllers are:
- SalvageController for processing replica failures
- ExpansionController for processing volume/engine/replica size changes
- BackupController for processing DR volumes as well as restore operations

This also changes the behavior so that:
- v.Status.CurrentNodeID is only set once we are fully attached
- v.Status.CurrentNodeID is only unset once we are fully detached

This is so that the status represents a resting state (i.e. attached/detached)
since the transition states are based on the v.Spec.NodeID

longhorn/longhorn#2527

### DIFF
--- a/api/model.go
+++ b/api/model.go
@@ -980,14 +980,18 @@ func toVolumeResource(v *longhorn.Volume, ves []*longhorn.Engine, vrs []*longhor
 		Timestamp: timestamp,
 	}
 
-	actions := map[string]struct{}{}
+	// api attach & detach calls are always allowed
+	// the volume manager is responsible for handling them appropriately
+	actions := map[string]struct{}{
+		"attach": {},
+		"detach": {},
+	}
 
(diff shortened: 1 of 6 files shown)


joshimoo:
This is a draft PR with work in progress, I left a couple of todos, that I plan to address.
The big changes are the changes in behavior to attachment handling.

The relevant commit for review is [WIP change volume attachment handling](https://github.com/longhorn/longhorn-manager/pull/857/commits/2ac48fd936ee65fd4d9b37d445f76e698319e3a4) this link will stop working once I force push over this, but you can manually select that specific commit.

PR #858, opened 2021-04-01
Fix websocket traffic flood

c3y1huang:
Currently, websocket stream will return all resource data at 1-sec interval even when there is no update in the resource data. This is caused by the timestamp in all resource objects.
    
Remove the timestamp and instead returns an empty data list to the frontend reduces the duplicated data transfer and also keeps the websocket connection alive.

https://github.com/longhorn/longhorn/issues/2372

### DIFF
--- a/api/model.go
+++ b/api/model.go
@@ -2,7 +2,6 @@ package api
 
 import (
 	"strconv"
-	"time"
 
 	"github.com/rancher/go-rancher/api"
 	"github.com/rancher/go-rancher/client"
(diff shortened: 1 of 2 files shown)


joshimoo (commented):
Do we know if the timestamp was used for anything?

c3y1huang:
> Do we know if the timestamp was used for anything?

Afaik this is only used to reconcile conflicting states between ws and http, I couldn't find it anywhere else...

@ttpcodes do you know if the timestamps are used anywhere else?

innobead (dismissed):
LGTM

innobead (approved):

PR #859, opened 2021-04-01
Always call attach/detach from the csi driver + auto salvage hotfix

joshimoo:
There is two things that are required one is the csi changes to always call attach/detach.
This insures that kubernetes + longhorn volumes are in sync.
I plan on working on structured error returns for the longhorn api, so that we can use them to do appropriate error handling.
https://github.com/longhorn/longhorn/issues/1875

The other issue is the auto salvage hotfix.
Honestly I have a hard time to reason about the current volume controllers
attach/detach behavior and therefore I cannot confidently say whether this
change has any undesired side effects.

I recommend we prioritize the volume controller refactor, 
so that attach/detach decisions and state are solely based on the set v.Spec.NodeID
https://github.com/longhorn/longhorn-manager/pull/857

Issues:
- https://github.com/longhorn/longhorn/issues/2309
- https://github.com/longhorn/longhorn/issues/2329

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -2502,7 +2502,7 @@ func (vc *VolumeController) processMigration(v *longhorn.Volume, es map[string]*
 		}
 		if migrationEngine.Spec.NodeID != "" && migrationEngine.Spec.NodeID != v.Spec.MigrationNodeID {
 			return fmt.Errorf("volume %v: engine is on node %v vs volume migration on %v",
-				v.Name, migrationEngine.Spec.NodeID, v.Spec.NodeID)
+				v.Name, migrationEngine.Spec.NodeID, v.Spec.MigrationNodeID)
 		}
 		migrationEngine.Spec.NodeID = v.Spec.MigrationNodeID
 		migrationEngine.Spec.ReplicaAddressMap = replicaAddressMap
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -1029,11 +1029,17 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
 		// make sure the volume is detached before automatically salvage replicas
 		if autoSalvage && v.Status.State == types.VolumeStateDetached && !v.Status.IsStandby && !v.Status.RestoreRequired {
 			// There is no need to auto salvage (and reattach) a volume on an unavailable node
+			// If we return error we will loop forever, instead we can do auto salvage even if the volume should be detached
+			// since all auto salvage does is unset the failure state for the replicas
+			// on a single replica volume, if the replica is on a failed node, it will be skipped below
+			// and the volume will remain in the faulted state but as soon as the node of the replica comes back
+			// we can do the auto salvage operation and set the Robustness to Unknown.
+			shouldBeAttached := v.Spec.NodeID != ""
 			isNodeDownOrDeleted, err := vc.ds.IsNodeDownOrDeleted(v.Spec.NodeID)
-			if err != nil {
+			if err != nil && shouldBeAttached {
 				return err
 			}
-			if !isNodeDownOrDeleted {
+			if !isNodeDownOrDeleted || !shouldBeAttached {
 				lastFailedAt := time.Time{}
 				failedUsableReplicas := map[string]*longhorn.Replica{}
 				dataExists := false
@@ -1097,6 +1103,8 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
 					if salvaged {
 						// remount the reattached volume later if possible
 						// For the auto-salvaged volume, `v.Status.CurrentNodeID` is empty but `v.Spec.NodeID` shouldn't be empty.
+						// There shouldn't be any problems if v.Spec.NodeID is empty, since the volume is desired to be detached
+						// so we just unset PendingNodeID.
 						v.Status.PendingNodeID = v.Spec.NodeID
 						v.Status.RemountRequestedAt = vc.nowHandler()
 						msg := fmt.Sprintf("Volume %v requested remount at %v", v.Name, v.Status.RemountRequestedAt)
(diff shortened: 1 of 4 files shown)


joshimoo:
This fixes issues with salvage of multiple replicas as well, it's just easier to test with a single replica case.
- This should be tested with data locality enabled/disabled.
- This should be tested with replicaset / statefulset
- This should be tested with RWO/RWX volumes

Reproduce steps:
- create RWO|RWX volume with replica count = 1 & data locality = enabled|disabled
- create deployment|statefulset for volume
- power down node of volume/replica
- wait for pod deletion & recreation
- volume will fail to attach since volume is not ready (i.e remains faulted, since single replica is on downed node)
- power up node
- verify auto salvage finishes (i.e pod completes start)
- verify volume attached & accessible by pod (i.e test data is available)

csi/controller_server.go
```diff
@@ -381,56 +309,78 @@ func (cs *ControllerServer) ControllerPublishVolume(ctx context.Context, req *cs
...
-	waitForController := isAttached && (len(existVol.Controllers) == 0 || existVol.Controllers[0].Endpoint == "")
-	if !existVol.Ready || waitForController {
-		return nil, status.Errorf(codes.Aborted, "The volume %s in state %v is not ready for workloads, is waiting on controller %v",
-			req.GetVolumeId(), existVol.State, waitForController)
+	// TODO: JM if volume is already attached to a different node, return code `codes.FailedPrecondition`
+	//  this should be handled by the processing of the api return code
+	if !requiresSharedAccess(volume, volumeCapability) && volume.State == string(types.VolumeStateAttached) && volume.Controllers[0].HostId != req.GetNodeId() {
+		return nil, status.Errorf(codes.FailedPrecondition, "The volume %s cannot be attached to the node %s since it is already attached to the node %s",
+			req.GetVolumeId(), req.GetNodeId(), volume.Controllers[0].HostId)
 	}
```
shuo-wu, line 350: Maybe we can let this csi server blindly send attach calls even without checking readiness or frontend. But I am not sure if the upper level will do something special for some special error code like `codes.FailedPrecondition`.

csi/controller_server.go
```diff
@@ -485,45 +419,39 @@ func (cs *ControllerServer) ControllerUnpublishVolume(ctx context.Context, req *
...
-		if !cs.waitForVolumeState(req.GetVolumeId(), string(types.VolumeStateDetached), isVolumeDetached, false, true) {
+		if !cs.waitForVolumeState(req.GetVolumeId(), "volume unpublished", checkVolumeUnpublished, false, true) {
 			return status.Errorf(codes.DeadlineExceeded, "Failed to detach volume %s from node %s", req.GetVolumeId(), req.GetNodeId())
 		}
 		return nil
 	})
 }
 
+// unpublishVolume sends the actual detach request to the longhorn api and executes the passed waitForResult func
+func (cs *ControllerServer) unpublishVolume(volume *longhornclient.Volume, nodeID string, waitForResult func() error) (*csi.ControllerUnpublishVolumeResponse, error) {
```
shuo-wu, line 440: Can we use `cs.waitForVolumeState` only? I don't know why there is an extra wrapper function `waitForResult`.

controller/volume_controller.go
```diff
@@ -1029,11 +1029,17 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
...
+			// and the volume will remain in the faulted state but as soon as the node of the replica comes back
+			// we can do the auto salvage operation and set the Robustness to Unknown.
+			shouldBeAttached := v.Spec.NodeID != ""
 			isNodeDownOrDeleted, err := vc.ds.IsNodeDownOrDeleted(v.Spec.NodeID)
-			if err != nil {
+			if err != nil && shouldBeAttached {
 				return err
 			}
-			if !isNodeDownOrDeleted {
+			if !isNodeDownOrDeleted || !shouldBeAttached {
```
shuo-wu, line 1042: Can we decouple the auto salvage from the node down case or auto reattachment feature? If there is a replica candidate on a running node, we will unset `spec.failedAt`. If this replica is actually no longer reusable, we can fail the volume again next time the volume is attached.

csi/controller_server.go
joshimoo: Yeah we should but for v1.1.1 I don't have time to implement structured error for the backend.
So I handled this specific case in the csi driver for the moment with the TODO.
FailedPrecondition is dedicated to the multi attach condition.
Technically we could just call the api and return internal server error which is less specific.

csi/controller_server.go
```diff
@@ -204,7 +204,12 @@ func (cs *ControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 
-	if !cs.waitForVolumeState(resVol.Id, string(types.VolumeStateDetached), isVolumeDetached, true, false) {
+	// Do we have a better condition than this?
+	checkVolumeCreated := func(vol *longhornclient.Volume) bool {
+		return vol.State == string(types.VolumeStateDetached)
```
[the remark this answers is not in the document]
joshimoo: This is called by waitForVolumeState only if vol exists.

csi/controller_server.go
joshimoo: No waitForVolumeState only has information about the volume. (i.e. retrieves the volume)
For the actual checks we need external information form the request.
That's why I pass the anonymous function that includes all necessary conditions (req.NodeId), that internally calls waitForVolumeState.

controller/volume_controller.go
joshimoo: I am going for a surgical change for the hotfix, since it's currently hard to reason about the volume controller behavior :)
In this case less change is better, till we have a volume controller that can be better reasoned about :)

csi/controller_server.go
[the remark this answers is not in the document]
shuo-wu: If the deletion timestamp is exposed, we may need to check that field. But actually No, then we don't need to check any fields for volume creation case here.  As a result, I use this to skip field check

controller/volume_controller.go
shuo-wu: Got it. Let's re-visit this during the refactor

shuo-wu (approved):
LGTM

innobead:
Thanks for the fix!

PR #862, opened 2021-04-06
CSI-sanity check - should fail when the node does not exist

c3y1huang:
Fix error code when node does not exist.
```
    Expected
        <codes.Code>: 13
    to equal
        <codes.Code>: 5
```

https://github.com/longhorn/longhorn/issues/2076
https://github.com/longhorn/longhorn/issues/2270

### DIFF
--- a/csi/controller_server.go
+++ b/csi/controller_server.go
@@ -309,8 +309,12 @@ func (cs *ControllerServer) ControllerPublishVolume(ctx context.Context, req *cs
 		return nil, status.Error(codes.InvalidArgument, msg)
 	}
 
-	// TODO: JM if the node is not ready we need to return `codes.NotFound`
-	//  should be handled by the processing of the api return codes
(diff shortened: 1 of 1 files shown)


c3y1huang:
```
ip-172-30-0-21:/home/ec2-user/csi-test/cmd/csi-sanity # ./csi-sanity -csi.endpoint /var/lib/kubelet/obsoleted-longhorn-plugins/driver.longhorn.io/csi.sock -ginkgo.skip="should create volume from an existing source snapshot|should return appropriate values|should succeed when creating snapshot with maximum-length name|should succeed when requesting to create a snapshot with already existing name and same source volume ID|should fail when requesting to create a snapshot with already existing name and different source volume ID"

Ran 42 of 77 Specs in 144.196 seconds
SUCCESS! -- 42 Passed | 0 Failed | 1 Pending | 34 Skipped
```

innobead (approved):

PR #863, opened 2021-04-06
Fix frontend cannot determine update from data only

c3y1huang:
When there is only one piece of data left and a delete operation is performed frontend cannot tell if the data has not changed or if the data is empty.

https://github.com/longhorn/longhorn/issues/2372

### DIFF
--- a/api/stream.go
+++ b/api/stream.go
@@ -100,7 +100,10 @@ func writeList(conn *websocket.Conn, oldResp *client.GenericCollection, listFunc
 
 	resp := newResp
 	if oldResp != nil && reflect.DeepEqual(oldResp, newResp) {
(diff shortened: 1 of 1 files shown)


c3y1huang:
@smallteeths thanks for catching this, please see if this works for you (c3y1huang/research:232-lh-manager).

c3y1huang:
Discussed with @smallteeths , we could also try `action.payload.resourceType === undefined`. If that works then I will close this PR.

c3y1huang:
> Discussed with @smallteeths , we could also try `action.payload.resourceType === undefined`. If that works then I will close this PR.

@smallteeths confirmed working. close PR.

PR #868, opened 2021-04-07
Add backing image in images.txt

innobead:
ref https://github.com/longhorn/longhorn/issues/2295


c3y1huang (approved):
LGTM, consider sorting alphabetically would make it looks nicer.

innobead:
`longhornio/backing-image-manager` the name convention is different because of project folder is not `longhorn` prefix. Let's still place in the group of longhorn images first, probably we need to change the prefix later.

shuo-wu (approved):
LGTM

PR #869, opened 2021-04-08
Remove BIM Ownership Transferring and address some review comments

shuo-wu:
Directly deleting old managers and ask new managers to reuse the
existing files (if possible) are enough. The side effect of this
modification is that the existing backing image won't be shwon
during the system upgrade, which is acceptable.


shuo-wu:
I Will test this PR tomorrow.

PhanLe1010:
I like this solution! Feel more resilient than the the transfer ownership gRPC calls.

shuo-wu:
> I like this solution! Feel more resilient than the the transfer ownership gRPC calls.
> Should we also update the LEP?

Sure. I will update the LEP after passing the tests.

innobead:
Can we change to PR instead of Draft PR if it's ready to review? Thanks.

shuo-wu:
Will update the image and vendor after the related backing image manager PR gets merged.

The regression test:
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/151/

deploy/install/02-components/01-manager.yaml
```diff
@@ -31,7 +31,7 @@ spec:
         - --share-manager-image
         - longhornio/longhorn-share-manager:v1_20210406
         - --backing-image-manager-image
-        - longhornio/backing-image-manager:v1_20210407
+        - shuowu/backing-image-manager:v1_20210409
```
PhanLe1010, line 34: Don't forget to change this one back

PhanLe1010 (approved):
LGTM. Thanks for the awesome job

innobead (commented):
LGTM, but as @PhanLe1010 mentioned, needs to update the backing-image-manager image when ready before merging.

innobead (approved):
LGTM

PR #871, opened 2021-04-12
Allow users to sepecify node selector for Longhorn components

PhanLe1010:
This feature allows user to limit Longhorn on a specific set of nodes in the cluster. 

User stories:
* Deploy Longhorn in a cluster with both Linux nodes and Windows nodes
* [Longhorn in GKE](https://forums.rancher.com/t/longhorn-in-gke-node-affinity/19814)

User experience:
1. Set the node selector for user-deployed components (Longhorn manager, UI, driver deployer) in Helm or YAML file
2. Then set the node selector for managed components (everything else) in `system-managed-components-node-selector` setting

Note: Workloads that use Longhorn volume can only run on the selected nodes in the cluster.

longhorn/longhorn#2199

### DIFF
--- a/app/driver.go
+++ b/app/driver.go
@@ -189,7 +189,10 @@ func checkKubernetesVersion(kubeClient *clientset.Clientset) error {
 	return nil
 }
 
-func deployCSIDriver(kubeClient *clientset.Clientset, lhClient *lhclientset.Clientset, c *cli.Context, managerImage, managerURL string) error {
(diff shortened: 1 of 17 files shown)


innobead (changes requested):
LGTM. One thing, do we need to apply this to recurring job as well?

shuo-wu (approved):
LGTM. Except for the recurring job part.

PhanLe1010:
Thanks @innobead  and @shuo-wu . I added the cronjob. My bad

innobead (approved):

PR #872, opened 2021-04-13
Update backing image manager and bug fix

shuo-wu:
longhorn/longhorn#2416

### DIFF
--- a/controller/backing_image_manager_controller.go
+++ b/controller/backing_image_manager_controller.go
@@ -15,6 +15,7 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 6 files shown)


shuo-wu:
Regression test: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/

shuo-wu:
Since drone won't tag the BIM image like `longhornio/backing-image-manager:20210413` automatically. I just update the tag to `master` for now.

innobead:
> Since drone won't tag the BIM image like `longhornio/backing-image-manager:20210413` automatically. I just update the tag to `master` for now.

@meldafrawi can you help with that? we should release the specific tag instead of master.

innobead (commented):
LGTM, except the image tag should be a release version instead master tag as we discussed.

shuo-wu:
The image is updated. Besides, there is a TCP readiness probe introduced for the BIM pods in the last commit.

innobead (approved):

PR #873, opened 2021-04-14
Fix volume cannot be attached when one of the replicas has error IM and the node of IM is still running

PhanLe1010:
The temporary solution is: don't attempt to start replica if the corresponding default IM is `error`, `unknown`, or `stopped`

I think in the long run, we should consider refactor the replica's reconciliation logic in replica controller so that it correctly set the state of the replica

longhorn/longhorn#2475

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -1264,7 +1264,11 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
 			if err != nil {
 				return errors.Wrapf(err, "cannot find node %v", r.Spec.NodeID)
 			}
-			if nodeDown {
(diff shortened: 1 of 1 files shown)


PhanLe1010:
Running E2E test

innobead (approved):
LGTM

@PhanLe1010 could you create a following ticket for the refactoring? then we can do it later.

shuo-wu (approved):
LGTM as the workaround/hotfix. 

BTW, will the failed replica loop in reusing when the volume is running and the related replica IM cannot be running?

PhanLe1010:
Yes, failed replica will loop in reusing

PhanLe1010:
The E2E core tests passes @innobead

PR #874, opened 2021-04-14
Fix bug: Longhorn always restart the managed components when the node selector is empty

PhanLe1010:
When node selector is empty,  the object return by client-go has a `nil` map for nodeSelector. This makes the deep equal comparison always fail since the nodeSelector is an empty map. As the result, Longhorn manage restarts the managed components

longhorn/longhorn#2199

### DIFF
--- a/controller/setting_controller.go
+++ b/controller/setting_controller.go
@@ -579,6 +579,11 @@ func (sc *SettingController) updateNodeSelector() error {
 		return errors.Wrapf(err, "failed to list backing image manager pods for node selector update")
 	}
 	for _, dp := range deploymentList {
+		if dp.Spec.Template.Spec.NodeSelector == nil {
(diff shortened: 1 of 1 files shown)


innobead (commented):
LGTM, but again probably we should consider recurring job update in this function as well.

PhanLe1010:
This is not an issue with recurring jobs since we remember what we applied the last time https://github.com/longhorn/longhorn-manager/blob/1f1452a43d436e9c3a72bb59b28b0553ddf55d8c/controller/volume_controller.go#L2255-L2261

May be we should consider this approach for all other components so we don't have to deal with this kind of `nil` or `empty` map

innobead:
ya, but it's good for now. LGTM.

innobead (approved):

PR #875, opened 2021-04-14
Remove the schedule condition check from the Volume attach call

joshimoo:
This condition is already covered by the readiness check in the detached
state. I identified some issues with the way the scheduling conditions are
handled in the volume controller, but they are outside of the scope for
v1.1.1 This is a surgical fix, to allow for rwx volumes to be used even if
one replica cannot be scheduled.

The share manager currently does not check for this condition, and will
always allow for volume usage even if not all replicas could be scheduled.
After creation, this is part of the problem, since the intent of the
scheduling failure is to keep the volume in the creation state, right now
the volume controller does not do this instead, it has some hacks for the
detached state.

I plan on looking into this, when working on the volume controller refactor.

longhorn/longhorn#2463

### DIFF
--- a/manager/volume.go
+++ b/manager/volume.go
@@ -322,10 +322,6 @@ func (m *VolumeManager) Attach(name, nodeID string, disableFrontend bool, attach
 		return nil, fmt.Errorf("cannot attach volume %v because the engine image %v is not deployed on at least one of the the replicas' nodes or the node that the volume is going to attach to", v.Name, v.Spec.EngineImage)
 	}
 
-	scheduleCondition := types.GetCondition(v.Status.Conditions, types.VolumeConditionTypeScheduled)
-	if scheduleCondition.Status != types.ConditionStatusTrue {
-		return nil, fmt.Errorf("volume %v cannot be scheduled due to lack of nodes or disks satisfied the space requirement", name)
-	}
 	restoreCondition := types.GetCondition(v.Status.Conditions, types.VolumeConditionTypeRestore)
 	if restoreCondition.Status == types.ConditionStatusTrue {
 		return nil, fmt.Errorf("volume %v is restoring data", name)


khushboo-rancher:
One doubt - What happens when `Allow Volume Creation with Degraded Availability` is disable. In that case we want to check the scheduling condition, right?

joshimoo:
@khushboo-rancher the scheduling condition for creation is covered by the readiness check here: https://github.com/longhorn/longhorn-manager/blob/master/api/model.go#L907
detached + unscheduled.

The share-manager doesn't care either way, since it just sets the spec.nodeID which will make the volume attach.
The real problem is that our scheduling determination code, is hacky and the intent should be to remain in the `creating` state.
Any modification of the scheduling determination code is outside of the scope for v1.1.1 and will need to be addressed as part of the volume controller refactor.

joshimoo:
The csi driver checks for volume readiness before trying to attach:
https://github.com/longhorn/longhorn-manager/blob/master/csi/controller_server.go#L344

The volume readiness will be false, since the scheduling condition is false and the volume is in detached state.
Once the volume is attached once, we cleanup the failed to schedule replicas which will make the volume no longer be unschedulable while detached, that's the current hacky implementation. Since in that case the volume will be schedulable while attached and the attach call for RWO happens at the same time as the attachment, we don't notice this problem for RWO volumes. 

But it's really more of a hack and not intentional, fixes to the scheduling condition determination are outside of the scope for v1.1.1 and need a refactor of the volume controller.

joshimoo:
here is what it looks like, when trying to create volumes when degraded creation is disabled.
image
image

Since the share-manager doesn't care, it will set v.spec.NodeID once the volume is requested to be used by a workload.
So the volume becomes available.

For the RWO volume the workload will check the readiness condition before calling attach which will prevent usage of the volume from the csi driver before being fully scheduled. A manual attach would still work, but isn't really a concern, since the user decided to use the volume anyway through his manually initiated action.

Once degraded creation is enabled, the hack in the volume controller overwrites the scheduling flag.
https://github.com/longhorn/longhorn-manager/blob/master/controller/volume_controller.go#L952-L972
image

innobead (approved):
LGTM, and thanks @joshimoo detailed explanation :+1:

PR #878, opened 2021-04-19
Update backing image manager image

shuo-wu:


innobead:
LGTM, let's wait until the backing image manager PR merge.

innobead (approved):

PR #881, opened 2021-04-21
Fix bugs for the backing image manager

shuo-wu:


controller/backing_image_manager_controller.go
```diff
@@ -688,7 +688,7 @@ func (c *BackingImageManagerController) isEligibleForPulling(currentBIM *longhor
 			}
 			continue
 		}
-		cksum := util.GetStringChecksum(biName + bimName)
+		cksum := util.GetStringChecksum(biName + bim.Spec.DiskUUID)
```
PhanLe1010, line 691: Nice catch! Should we add a comment explain this detail since it is hard to see just from reading the code?

PhanLe1010 (approved):
LGTM

controller/backing_image_manager_controller.go
shuo-wu: Sure. Will add a comment above.

shuo-wu: Updated

PhanLe1010 (approved):

shuo-wu:
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/195/

innobead (approved):

PR #882, opened 2021-04-21
Upgrade CSI components to 1.2.0

weihanglo:
Maybe part of https://github.com/longhorn/longhorn/issues/2388

- Upgrade CSI to [1.2.0](https://github.com/container-storage-interface/spec/releases/tag/v1.2.0)
- Set minimum support Kubernetes version to 1.16.0


innobead:
cc @joshimoo

joshimoo:
@weihanglo thank you for the PR, but we are not quite ready yet to update the kubernetes components. The effort is tracked at 
https://github.com/longhorn/longhorn/issues/2757
https://github.com/longhorn/longhorn/issues/2388

So for the moment, we cannot include this PR.

PR #883, opened 2021-04-22
Increase the hardcoded NFS client timeout

joshimoo:
We use a 3s timeout, with 3 retransmission tries before failure.
After each timeout the timeout will be linearly increased (3s, 6s, 9s)
This means that an IO operation has a maximum runtime of 18s before failure.

For the HA solution we want to use hard mode, which will lead to an infinite
retry, since in that mode we want to be able to guarantee the survival of
the workload pods instead of having to terminate them.

longhorn/longhorn#2528


PhanLe1010:
Should we expose the setting to the user instead of hard-coding it?

innobead (approved):

PR #884, opened 2021-04-30
Only keep upto 2 snapshots for recurring backup jobs

PhanLe1010:
To avoid creating unnecessary extra snapshots which consume the disk space, we only keep up to 2 snapshots for recurring backup jobs (the snapshot of the last backup and the snapshot of the currently running backup)

If users want to create more snapshots, they can use recurring snapshot job

longhorn/longhorn#2231

### DIFF
--- a/app/volume.go
+++ b/app/volume.go
@@ -106,7 +106,10 @@ func snapshot(c *cli.Context) error {
 	}
 
 	backup := c.Bool(FlagBackup)
-	job, err := NewJob(managerURL, volume, snapshotName, labelMap, retain, backup)
+
+	logger := logrus.StandardLogger()
+
+	job, err := NewJob(logger, managerURL, volume, snapshotName, labelMap, retain, backup)
 	if err != nil {
 		return err
 	}
@@ -115,6 +118,7 @@ func snapshot(c *cli.Context) error {
 }
 
 type Job struct {
+	logger       logrus.FieldLogger
 	lhClient     lhclientset.Interface
 	namespace    string
 	volumeName   string
@@ -126,7 +130,7 @@ type Job struct {
 	api *longhornclient.RancherClient
 }
 
-func NewJob(managerURL, volumeName, snapshotName string, labels map[string]string, retain int, backup bool) (*Job, error) {
+func NewJob(logger logrus.FieldLogger, managerURL, volumeName, snapshotName string, labels map[string]string, retain int, backup bool) (*Job, error) {
 	namespace := os.Getenv(types.EnvPodNamespace)
 	if namespace == "" {
 		return nil, fmt.Errorf("Cannot detect pod namespace, environment variable %v is missing", types.EnvPodNamespace)
@@ -157,7 +161,17 @@ func NewJob(managerURL, volumeName, snapshotName string, labels map[string]strin
 		jobType = jobTypeBackup
 	}
 
+	logger = logger.WithFields(logrus.Fields{
+		"namespace":    namespace,
+		"volumeName":   volumeName,
+		"snapshotName": snapshotName,
+		"labels":       labels,
+		"retain":       retain,
+		"jobType":      jobType,
+	})
+
 	return &Job{
+		logger:       logger,
 		lhClient:     lhClient,
 		namespace:    namespace,
 		volumeName:   volumeName,
@@ -178,11 +192,10 @@ func (job *Job) handleVolumeDetachment() {
 	volumeName := job.volumeName
 	jobName, _ := job.labels[types.RecurringJobLabel]
 	if jobName == "" {
-		logrus.Warn("Missing RecurringJob label")
+		job.logger.Warn("Missing RecurringJob label")
 		return
 	}
 
-	logrus.Infof("Handling volume detachment for volume %v", volumeName)
 	for {
 		volume, err := volumeAPI.ById(volumeName)
 		if err == nil {
@@ -192,24 +205,26 @@ func (job *Job) handleVolumeDetachment() {
 			// !volume.DisableFrontend condition makes sure that volume is detached by this recurring job,
 			// not by the auto-reattachment feature.
 			if volume.State == string(types.VolumeStateDetached) && !volume.DisableFrontend {
-				logrus.Infof("Volume %v is detached", volumeName)
+				job.logger.Infof("Volume %v is detached", volumeName)
 				return
 			}
 			if volume.State == string(types.VolumeStateAttached) {
-				logrus.Infof("Attempting to detach volume %v from all nodes", volumeName)
+				job.logger.Infof("Attempting to detach volume %v from all nodes", volumeName)
 				if _, err := volumeAPI.ActionDetach(volume, &longhornclient.DetachInput{HostId: ""}); err != nil {
-					logrus.Infof("%v ", err)
+					job.logger.WithError(err).Info("Volume detach request failed")
 				}
 			}
 
 		} else {
-			logrus.Infof("%v ", err)
+			job.logger.WithError(err).Infof("Could not get volume %v", volumeName)
 		}
 		time.Sleep(DetachingWaitInterval)
 	}
 }
 
 func (job *Job) run() (err error) {
+	job.logger.Info("job starts running")
+
 	volumeAPI := job.api.Volume
 	volumeName := job.volumeName
 	jobName, _ := job.labels[types.RecurringJobLabel]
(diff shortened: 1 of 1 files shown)


shuo-wu (commented):
In general LGTM. Here are 2 minor suggestions

app/volume.go
```diff
@@ -157,7 +161,17 @@ func NewJob(managerURL, volumeName, snapshotName string, labels map[string]strin
 		jobType = jobTypeBackup
 	}
 
+	logger = logger.WithFields(logrus.Fields{
```
joshimoo, line 164: This is a good addition, good job thinking about this :)

app/volume.go
```diff
@@ -192,24 +205,26 @@ func (job *Job) handleVolumeDetachment() {
...
-				logrus.Infof("Volume %v is detached", volumeName)
+				job.logger.Infof("Volume %v is detached", volumeName)
 				return
 			}
 			if volume.State == string(types.VolumeStateAttached) {
-				logrus.Infof("Attempting to detach volume %v from all nodes", volumeName)
+				job.logger.Infof("Attempting to detach volume %v from all nodes", volumeName)
 				if _, err := volumeAPI.ActionDetach(volume, &longhornclient.DetachInput{HostId: ""}); err != nil {
-					logrus.Infof("%v ", err)
+					job.logger.Infof("%v ", err)
```
joshimoo, line 214:
```suggestion
					job.logger.withError(err).Info("Volume detach request failed")
```

app/volume.go
```diff
@@ -192,24 +205,26 @@ func (job *Job) handleVolumeDetachment() {
...
 				}
 			}
 
 		} else {
-			logrus.Infof("%v ", err)
+			job.logger.Infof("%v ", err)
```
joshimoo, line 219: consider adding a log message describing what failed then have the error in a withError field.

app/volume.go
```diff
@@ -293,7 +325,7 @@ func (job *Job) snapshotAndCleanup() (err error) {
 
 	defer func() {
 		if err != nil {
-			logrus.Warnf("created snapshot successfully but errored on cleanup for %v: %v", volumeName, err)
+			job.logger.Warnf("created snapshot successfully but errored on cleanup for %v: %v", volumeName, err)
```
joshimoo, line 328: consider using logger.withError for the error.

app/volume.go
```diff
@@ -353,17 +387,10 @@ func (job *Job) shouldDoRecurringSnapshot(volume *longhornclient.Volume) (bool,
 	if err != nil {
 		return false, err
 	}
-
 	// If volume-head has new data, we need to take snapshot, we need to take a new snapshot
```
joshimoo, line 390: adjust the comment

app/volume.go
```diff
@@ -412,32 +439,6 @@ func (job *Job) getVolumeHeadSize(volume *longhornclient.Volume) (int64, error)
...
-				Timestamp: t,
-			})
-		}
-	}
-	return job.getCleanupList(sts)
-}
-
 func (job *Job) getCleanupList(sts []*NameWithTimestamp) []string {
 	sort.Slice(sts, func(i, j int) bool {
 		if sts[i].Timestamp.Before(sts[j].Timestamp) {
```
joshimoo, line 444: you can just do `return sts[i].Timestamp.Before(sts[j].Timestamp)`

app/volume.go
```diff
@@ -737,3 +772,75 @@ func (job *Job) findARandomReadyNode(v *longhornclient.Volume) (string, error) {
...
+			return true
+		}
+		return false
+	})
+}
+
+// getOldNames returns a list of names from the input sts excluding the latest retainCount names
+func getOldNames(nts []NameWithTimestamp, retainCount int) []string {
+	sort.Slice(nts, func(i, j int) bool {
+		if nts[i].Timestamp.Before(nts[j].Timestamp) {
```
joshimoo, line 810: you can just do `return nts[i].Timestamp.Before(nts[j].Timestamp)`

app/volume.go
```diff
@@ -628,23 +656,30 @@ func (job *Job) shouldDoRecurringBackup(volume *longhornclient.Volume) (bool, er
...
 	}
-
 	return job.api.BackupVolume.ActionBackupGet(backupVolume, &longhornclient.BackupInput{
 		Name: volume.LastBackup,
 	})
 }
 
 func (job *Job) listBackupsForCleanup(backups []longhornclient.Backup) []string {
-	sts := []*NameWithTimestamp{}
+	sts := []NameWithTimestamp{}
```
joshimoo, line 682: moving the elements to the stack seems fine, since it's a small bounded structure.

app/volume.go
```diff
@@ -737,3 +772,75 @@ func (job *Job) findARandomReadyNode(v *longhornclient.Volume) (string, error) {
 	}
 	return readyNodeList[rand.Intn(len(readyNodeList))], nil
 }
+
+func filterSnapshots(snapshots []longhornclient.Snapshot, predicate func(snapshot longhornclient.Snapshot) bool) []longhornclient.Snapshot {
```
joshimoo, line 776: I like the usage of a predicate function :)

app/volume.go
```diff
@@ -737,3 +772,75 @@ func (job *Job) findARandomReadyNode(v *longhornclient.Volume) (string, error) {
...
+}
+
+// filterSnapshotsWithLabel return snapshots that have LabelKey and LabelValue
+func filterSnapshotsWithLabel(snapshots []longhornclient.Snapshot, labelKey, labelValue string) []longhornclient.Snapshot {
+	return filterSnapshots(snapshots, func(snapshot longhornclient.Snapshot) bool {
+		snapshotLabelValue, found := snapshot.Labels[labelKey]
+		if found && labelValue == snapshotLabelValue {
+			return true
+		}
+		return false
```
joshimoo, line 793:
```suggestion
		return found && labelValue == snapshotLabelValue
```

joshimoo (commented):
LGTM, shuo left some suggestions as well.
Good job on adding the logger with fields, the log messages could use some cleanup now, but we can do this the next time we touch this part of the code as well.

app/volume.go
PhanLe1010: Thank you. Fixed

app/volume.go
```diff
@@ -737,3 +772,75 @@ func (job *Job) findARandomReadyNode(v *longhornclient.Volume) (string, error) {
...
+	})
+
+	ret := []string{}
+	for i := 0; i < len(nts)-retainCount; i++ {
+		ret = append(ret, nts[i].Name)
+	}
+	return ret
```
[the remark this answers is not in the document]
PhanLe1010: We would have a problem if `len(nts)-retainCount < 0`

app/volume.go
PhanLe1010: Thank you. Fixed

app/volume.go
```diff
@@ -737,3 +772,75 @@ func (job *Job) findARandomReadyNode(v *longhornclient.Volume) (string, error) {
...
+	return filterSnapshots(snapshots, func(snapshot longhornclient.Snapshot) bool {
+		if _, ok := targets[snapshot.Name]; !ok {
+			return true
+		}
+		return false
+	})
+}
+
+// getOldNames returns a list of names from the input sts excluding the latest retainCount names
+func getOldNames(nts []NameWithTimestamp, retainCount int) []string {
```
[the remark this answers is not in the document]
PhanLe1010: Thank you. Fixed

app/volume.go
PhanLe1010: Thank you. Fixed

app/volume.go
PhanLe1010: I thought that the slice is still on heap because we do append() that could potentially reallocate the slice so Golang cannot determine the final size of the slice at compile time.

app/volume.go
PhanLe1010: Thank you. The function is no longer used. So, I deleted it.

app/volume.go
PhanLe1010: Thank you. Fixed

app/volume.go
PhanLe1010: Thank you. Fixed

app/volume.go
PhanLe1010: Thank you. Fixed

app/volume.go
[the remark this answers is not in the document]
shuo-wu: OK. Then the `if` check is unavoidable. The current implementation is fine.

shuo-wu (dismissed):

joshimoo (approved):
LGTM, shuo has already approved this, phan just squashed the fixup commit for the merge.

innobead:
@PhanLe1010 one question. Does it mean the snapshots will be deleted even the following backup failed? 

https://github.com/longhorn/longhorn-manager/blob/7876da85f557dfa758ef48262e595ace63fc77eb/app/volume.go#L498-L511

PR #885, opened 2021-05-01
Add PriorityClass setting to recurring jobs

PhanLe1010:
longhorn/longhorn#2444

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -2139,6 +2139,10 @@ func (vc *VolumeController) createCronJob(v *longhorn.Volume, job *types.Recurri
 	if err != nil {
 		return nil, err
 	}
+	priorityClass, err := vc.ds.GetSetting(types.SettingNamePriorityClass)
+	if err != nil {
+		return nil, err
+	}
 	nodeSelector, err := vc.ds.GetSettingSystemManagedComponentsNodeSelector()
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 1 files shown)


shuo-wu (approved):
LGTM

shuo-wu (changes requested):
Sorry. Longhorn may need to update tolerations or priority classes in the setting controller.

PhanLe1010:
> Sorry. Longhorn may need to update tolerations or priority classes in the setting controller.

Do you mean updating tolerations or priority classes for cronjob? If so, we don't need to do it setting controller because volume controller will update the cronjob when it is not in desired state https://github.com/longhorn/longhorn-manager/blob/1fbd2bf21fdcac5dbc063644c48e66acbdc633d6/controller/volume_controller.go#L2255-L2261

shuo-wu:
> > Sorry. Longhorn may need to update tolerations or priority classes in the setting controller.
> 
> Do you mean updating tolerations or priority classes for cronjob? If so, we don't need to do it setting controller because volume controller will update the cronjob when it is not in desired state
> 
> https://github.com/longhorn/longhorn-manager/blob/1fbd2bf21fdcac5dbc063644c48e66acbdc633d6/controller/volume_controller.go#L2255-L2261

OK. I forgot this part. Then this PR LGTM

shuo-wu (approved):

innobead (approved):

PR #887, opened 2021-05-04
app: Increase the timeout for the http client

shuo-wu:
The default request timeout for recurring jobs is 10s, which is
definitely not enough. If multiple recurring jobs for a volume
start simultaneously (due to the mutex lock), or some steps take
several seconds, the timeout will fail the recurring jobs.

Increasing the timeout is not able to completely avoid duplicate 
snapshots/backups if there are intensive snapshot-related
operations. But this should reduce the probability.

longhorn/longhorn#2278

Signed-off-by: Shuo Wu <shuo.wu@suse.com>

### DIFF
--- a/app/volume.go
+++ b/app/volume.go
@@ -31,6 +31,8 @@ const (
 	FlagRetain       = "retain"
 	FlagBackup       = "backup"
 
+	HTTPClientTimout = 1 * time.Minute
+
 	SnapshotPurgeStatusInterval = 5 * time.Second
 
 	WaitInterval              = 5 * time.Second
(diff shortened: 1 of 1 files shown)


shuo-wu:
I am not if there is a simple way to block a backup creation for over 10s then trigger the issue...

jenting (approved):

joshimoo (approved):
LGTM to increase the client timeout, I also think there is a backend timeout at 1 minute but I don't remember the details.

joshimoo:
For testing I think you can run a proxy between client <-> server then stall the server responses. But seems like a lot of work, with little gain, since we know that the client timeout previously was working we can assume the increase in timeout should work as well.

innobead:
Probably we need to enhance the snapshot command to support asynchronous mode to return immediately if the job gets triggered successfully instead of waiting until the job completed successfully.

https://github.com/longhorn/longhorn-manager/blob/ac065a8deb96193ffec142bcc7e7345615b57b8d/app/volume.go#L495-L543

joshimoo:
@innobead agree having the apis be async is a good design goal, for this specific case this should be covered by the backup refactoring https://github.com/longhorn/longhorn/issues/1761 since once the backup/snapshots are exposed as crds, we can operate on the crds and the only thing the api needs to do is creation/validation of these crds and the client then can poll the crd state instead of having the apis be blocking.

shuo-wu:
> For testing I think you can run a proxy between client <-> server then stall the server responses. But seems like a lot of work, with little gain, since we know that the client timeout previously was working we can assume the increase in timeout should work as well.

Yes. When I tested this part, I directly added a "sleep" in the backend side then validated if this timeout works or not. For this PR, maybe we can skip the test.

PR #888, opened 2021-05-04
use pvc to expand volume (#2422)

cclhsu:
#### Proposed Changes ####

use pvc to expand volume to resolve ticket 'Application couldn't mount the volume after expanded the volume' (#2422)

- Modify 'Expand' API to provide PVC/PV expand

#### Types of Changes ####
 
Bug fix

#### Verification ####

```
GIVEN:
- Kubernetes cluster deployed
- Longhorn deployed
- a statefulset/deployment with mounted volume deployed
WHEN:
- scale down replicas for statefulset/deployment to zero
- POST `Expand` to `longhorn_backend` 
THEN:
- wait for pvc/pv successful expand
- statefulset/deployment scale replica to 1 and disk attached successfully
```

Curl API

```
ACTION=expand
curl -sSX POST -H 'Content-Type: application/json' "http://${BACKEND_URL}:${BACKEND_PORT}/v1/volumes/${PV_NAME}?action=${ACTION}" -d '{"size":"4Gi"}' 
```

if volume is not detached will receive error like this

```
{
  "actions": {},
  "code": "Server Error",
  "detail": "",
  "links": {
    "self": "http://longhorn-backend.longhorn-system.svc.cluster.local:9500/v1/volumes/pvc-5d3833f1-5530-426f-bb8e-cb7b6aef467a"
  },
  "message": "invalid volume state to expand: pvc-5d3833f1-5530-426f-bb8e-cb7b6aef467a",
  "status": 500,
  "type": "error"
}
```

#### Linked Issues ####

Refs: https://github.com/longhorn/longhorn/issues/2422

#### Further Comments ####

### DIFF
--- a/manager/kubernetes.go
+++ b/manager/kubernetes.go
@@ -44,7 +44,7 @@ func (m *VolumeManager) PVCreate(name, pvName, fsType string) (v *longhorn.Volum
 	}
 
 	pv := datastore.NewPVManifestForVolume(v, pvName, storageClassName, fsType)
-	pv, err = m.ds.CreatePersisentVolume(pv)
+	pv, err = m.ds.CreatePersistentVolume(pv)
 	if err != nil {
 		return nil, err
 	}
@@ -89,21 +89,21 @@ func (m *VolumeManager) PVCCreate(name, namespace, pvcName string) (v *longhorn.
 		return nil, fmt.Errorf("cannot found PV %v or the PV status %v is invalid for PVC creation", ks.PVName, ks.PVStatus)
 	}
 
-	pv, err := m.ds.GetPersisentVolume(ks.PVName)
+	pv, err := m.ds.GetPersistentVolume(ks.PVName)
 	if err != nil {
 		return nil, err
 	}
 	// cleanup ClaimRef of PV. Otherwise the existing PV cannot be reused.
 	if pv.Spec.ClaimRef != nil {
 		pv.Spec.ClaimRef = nil
-		pv, err = m.ds.UpdatePersisentVolume(pv)
+		pv, err = m.ds.UpdatePersistentVolume(pv)
 		if err != nil {
 			return nil, err
 		}
 	}
 
 	pvc := datastore.NewPVCManifestForVolume(v, ks.PVName, namespace, pvcName, pv.Spec.StorageClassName)
-	pvc, err = m.ds.CreatePersisentVolumeClaim(namespace, pvc)
+	pvc, err = m.ds.CreatePersistentVolumeClaim(namespace, pvc)
 	if err != nil {
 		return nil, err
 	}
--- a/manager/volume.go
+++ b/manager/volume.go
@@ -4,11 +4,14 @@ import (
 	"fmt"
 	"reflect"
 	"sort"
+	"strconv"
 
 	"github.com/pkg/errors"
 	"github.com/robfig/cron"
 	"github.com/sirupsen/logrus"
 
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/longhorn/backupstore"
@@ -613,12 +616,50 @@ func (m *VolumeManager) Expand(volumeName string, size int64) (v *longhorn.Volum
 		return nil, fmt.Errorf("cannot expand volume before replica scheduling success")
 	}
 
-	if v.Spec.Size > size {
-		return nil, fmt.Errorf("cannot expand volume %v with current size %v to a smaller size %v", v.Name, v.Spec.Size, size)
+	size = util.RoundUpSize(size)
+
+	kubernetesStatus := &v.Status.KubernetesStatus
+	if kubernetesStatus.PVCName != "" && kubernetesStatus.LastPVCRefAt == "" {
+		pvc, err := m.ds.GetPersistentVolumeClaim(kubernetesStatus.Namespace, kubernetesStatus.PVCName)
+		if err != nil {
+			return nil, err
+		}
+
+		requestedSize := resource.MustParse(strconv.FormatInt(size, 10))
+
+		// TODO: Should check for pvc.Spec.Resources.Requests.Storage() here, once upgrade API to v0.18.x.
+		pvcSpecValue, ok := pvc.Spec.Resources.Requests[corev1.ResourceStorage]
+		if !ok {
+			return nil, fmt.Errorf("cannot get request storage")
+		}
+
+		if pvcSpecValue.Cmp(requestedSize) < 0 {
+			pvc.Spec.Resources = corev1.ResourceRequirements{
+				Requests: corev1.ResourceList{
+					corev1.ResourceStorage: requestedSize,
+				},
+			}
+
+			logrus.Infof("Persistent Volume Claim %v expansion from %v to %v requested", v.Name, v.Spec.Size, size)
+			_, err = m.ds.UpdatePersistentVolumeClaim(kubernetesStatus.Namespace, pvc)
+			if err != nil {
+				return nil, err
+			}
+
+			// return and CSI plugin call this API later for Longhorn volume expansion
+			return v, nil
+		}
+
+		// all other case belong to the CSI plugin call
+		logrus.Infof("CSI plugin call to expand volume %v", v.Name)
+
+		if pvcSpecValue.Cmp(requestedSize) > 0 {
+			size = util.RoundUpSize(pvcSpecValue.Value())
+		}
 	}
 
-	if v.Spec.Size == size {
-		logrus.Infof("Volume %v expansion is not necessary since current size %v == %v", v.Name, v.Spec.Size, size)
+	if v.Spec.Size >= size {
+		logrus.Infof("Volume %v expansion is not necessary since current size %v >= %v", v.Name, v.Spec.Size, size)
 		return v, nil
 	}
 
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -326,7 +326,7 @@ func (vc *VolumeController) syncVolume(key string) (err error) {
 		kubeStatus := volume.Status.KubernetesStatus
 
 		if kubeStatus.PVName != "" {
-			if err := vc.ds.DeletePersisentVolume(kubeStatus.PVName); err != nil {
+			if err := vc.ds.DeletePersistentVolume(kubeStatus.PVName); err != nil {
 				if !datastore.ErrorIsNotFound(err) {
 					return err
 				}
@@ -334,7 +334,7 @@ func (vc *VolumeController) syncVolume(key string) (err error) {
 		}
 
 		if kubeStatus.PVCName != "" && kubeStatus.LastPVCRefAt == "" {
-			if err := vc.ds.DeletePersisentVolumeClaim(kubeStatus.Namespace, kubeStatus.PVCName); err != nil {
+			if err := vc.ds.DeletePersistentVolumeClaim(kubeStatus.Namespace, kubeStatus.PVCName); err != nil {
 				if !datastore.ErrorIsNotFound(err) {
 					return err
 				}
--- a/csi/controller_server.go
+++ b/csi/controller_server.go
@@ -669,16 +669,11 @@ func (cs *ControllerServer) ControllerExpandVolume(ctx context.Context, req *csi
 		return nil, status.Errorf(codes.Internal, err.Error())
 	}
 
-	// we can only call the longhorn expansion api if the requested size is larger than the volume size
-	// since longhorn treats the size differently than kubernetes, in kubernetes capacity is a request
-	// to ensure that the volume has at least that amount of capacity.
 	requestedSize := req.CapacityRange.GetRequiredBytes()
-	if requestedSize > existingSize {
-		if existVol, err = cs.apiClient.Volume.ActionExpand(existVol, &longhornclient.ExpandInput{
-			Size: strconv.FormatInt(requestedSize, 10),
-		}); err != nil {
-			return nil, status.Errorf(codes.Internal, err.Error())
-		}
+	if existVol, err = cs.apiClient.Volume.ActionExpand(existVol, &longhornclient.ExpandInput{
+		Size: strconv.FormatInt(requestedSize, 10),
+	}); err != nil {
+		return nil, status.Errorf(codes.Internal, err.Error())
 	}
 
 	// kubernetes doesn't support volume shrinking and the csi spec specifies to return true
--- a/datastore/kubernetes.go
+++ b/datastore/kubernetes.go
@@ -389,45 +389,51 @@ func (s *DataStore) GetKubernetesNode(name string) (*corev1.Node, error) {
 	return s.knLister.Get(name)
 }
 
-// CreatePersisentVolume creates a PersistentVolume resource for the given
+// CreatePersistentVolume creates a PersistentVolume resource for the given
 // PersistentVolume object
-func (s *DataStore) CreatePersisentVolume(pv *corev1.PersistentVolume) (*corev1.PersistentVolume, error) {
+func (s *DataStore) CreatePersistentVolume(pv *corev1.PersistentVolume) (*corev1.PersistentVolume, error) {
 	return s.kubeClient.CoreV1().PersistentVolumes().Create(pv)
 }
 
-// DeletePersisentVolume deletes the PersistentVolume for the given
+// DeletePersistentVolume deletes the PersistentVolume for the given
 // PersistentVolume name
-func (s *DataStore) DeletePersisentVolume(pvName string) error {
+func (s *DataStore) DeletePersistentVolume(pvName string) error {
 	return s.kubeClient.CoreV1().PersistentVolumes().Delete(pvName, &metav1.DeleteOptions{})
 }
 
-// UpdatePersisentVolume updates the PersistentVolume for the given
+// UpdatePersistentVolume updates the PersistentVolume for the given
 // PersistentVolume object
-func (s *DataStore) UpdatePersisentVolume(pv *corev1.PersistentVolume) (*corev1.PersistentVolume, error) {
+func (s *DataStore) UpdatePersistentVolume(pv *corev1.PersistentVolume) (*corev1.PersistentVolume, error) {
 	return s.kubeClient.CoreV1().PersistentVolumes().Update(pv)
 }
 
-// GetPersisentVolume gets the PersistentVolume from the index for the
+// GetPersistentVolume gets the PersistentVolume from the index for the
 // given name
-func (s *DataStore) GetPersisentVolume(pvName string) (*corev1.PersistentVolume, error) {
+func (s *DataStore) GetPersistentVolume(pvName string) (*corev1.PersistentVolume, error) {
 	return s.pvLister.Get(pvName)
 }
 
-// CreatePersisentVolumeClaim creates a PersistentVolumeClaim resource
+// CreatePersistentVolumeClaim creates a PersistentVolumeClaim resource
 // for the given PersistentVolumeclaim object and namespace
-func (s *DataStore) CreatePersisentVolumeClaim(ns string, pvc *corev1.PersistentVolumeClaim) (*corev1.PersistentVolumeClaim, error) {
+func (s *DataStore) CreatePersistentVolumeClaim(ns string, pvc *corev1.PersistentVolumeClaim) (*corev1.PersistentVolumeClaim, error) {
 	return s.kubeClient.CoreV1().PersistentVolumeClaims(ns).Create(pvc)
 }
 
-// DeletePersisentVolumeClaim deletes the PersistentVolumeClaim for the
+// DeletePersistentVolumeClaim deletes the PersistentVolumeClaim for the
 // given name and namespace
-func (s *DataStore) DeletePersisentVolumeClaim(ns, pvcName string) error {
+func (s *DataStore) DeletePersistentVolumeClaim(ns, pvcName string) error {
 	return s.kubeClient.CoreV1().PersistentVolumeClaims(ns).Delete(pvcName, &metav1.DeleteOptions{})
 }
 
-// GetPersisentVolumeClaim gets the PersistentVolumeClaim from the
+// UpdatePersistentVolumeClaim expand the PersistentVolumeClaim from the
 // index for the given name and namespace
-func (s *DataStore) GetPersisentVolumeClaim(namespace, pvcName string) (*corev1.PersistentVolumeClaim, error) {
+func (s *DataStore) UpdatePersistentVolumeClaim(namespace string, pvc *corev1.PersistentVolumeClaim) (*corev1.PersistentVolumeClaim, error) {
+	return s.kubeClient.CoreV1().PersistentVolumeClaims(namespace).Update(pvc)
+}
+
+// GetPersistentVolumeClaim gets the PersistentVolumeClaim from the
+// index for the given name and namespace
+func (s *DataStore) GetPersistentVolumeClaim(namespace, pvcName string) (*corev1.PersistentVolumeClaim, error) {
 	return s.pvcLister.PersistentVolumeClaims(namespace).Get(pvcName)
 }
 


shuo-wu (commented):
My concern is if we can merge this with the old expansion API, which can make Longhorn expansion more intelligent:

1. Since the min K8s version for Longhorn is v1.16 now and the csi resizer should be installed by default. Once the expansion happens, it's better to make sure the related PVC is in sync.
2. If the volume is not detached, of course, the API call should fail.
3. If there is no PVC for the volume, Longhorn can expand the volume only.
4. If the related StorageClass doesn't support expansion, Longhorn should fail the call and inform the user.
5. If the related StorageClass doesn't exist, e.g., the volume and PVC/PV are all created by Longhorn UI, Longhorn should recreate the PVC with larger size. And expand the volume.

joshimoo:
@shuo-wu what is the reason for recreating the pvc in point 5?
```
If the related StorageClass doesn't exist, e.g., the volume and PVC/PV are all created by Longhorn UI, Longhorn should recreate the PVC with larger size. And expand the volume.
```

shuo-wu:
@joshimoo A PVC can be expanded only if it is dynamically provisioned by a StorageClass with option `allowVolumeExpansion` enabled. Otherwise, the following error will be triggered:
```
$ k edit pvc vol1
error: persistentvolumeclaims "vol1" could not be patched: persistentvolumeclaims "vol1" is forbidden: only dynamically provisioned pvc can be resized and the storageclass that provisions the pvc must support resize
You can run `kubectl replace -f /tmp/kubectl-edit-aux52.yaml` to try this update again.
```

cclhsu:
@shuo-wu I have address first 4 points in previous update.  For point 5, I am still thinking if this scenario is what we wanted: 
1. User has an original storage class with `allowVolumeExpansion: false`.
2. User scaled down workload.
3. User deleted the storage class.
4. User expand volume in Longhorn UI. Longhorn-manager will create a storage class with same name and `allowVolumeExpansion: true`.

joshimoo:
I feel like we shouldn't concern ourselves with the storage class, i.e. let kubernetes evaluate whether the pvc expansion is valid or not.

Together with this point above, I think we should consider creating the pvc/pv with the `longhorn` storage class instead of the non existing `longhorn-static` class.

@shuo-wu @cclhsu What do you think?

The PR currently doesn't evaluate the PVC size vs requested size.
Since this API endpoint will be called by the csi driver, after the user has manually updated the PVC size.
There is no need to try to update the PVC in that case.

shuo-wu:
I was thinking we can hide the old expansion API and the workflow should be like:
1. If the StorageClass does not exist: Users call the new expansion API --> Longhorn recreate the PVC with new size --> Longhorn expand the volume
2. If the StorageClass exists: Users call the new expansion API --> Longhorn update the PVC spec size --> the CSI resizer receive expansion request then call the old expansion API --> Longhorn expand the volume

If there is always an existing storageclass for PVCs, then I think we can ignore the 1st case.

cclhsu:
PVC's storage class is an immutable field.  

1. When recreating a customized storage class with the same name, and replacing the value of `allowVolumeExpansion` from `false` to `true`, storage is able to expand successfully. 
2 With current Longhorn implementation, deleting PVC will also automatically delete PV.  An effect that we definitely don't wanted when doing expanding.  I need to figure out how to disable it.

joshimoo:
That's what I was worried about, having to delete the PVC implies the deletion of the PV -> longhorn volume.
Therefore it's better to make the UI created pvc/pv use the default `longhorn` storage class, instead of the non existing `longhorn-static` class.

Also as mentioned before, I don't think we should check the value on the storage class, since it's handled by kubernetes when trying to extend the pvc. i.e. expansion will fail if the storage class does not exist or does not support expansion.

So no need for us to check explicitly.

cclhsu:
With the latest PR update,

1. Use existing API `Expand` to check if `lhv` kubernetes status has PVC.
2. If has PVC and storage class is not exist or expandable, error out.  We can expand PV when storage class set as unexpandable, but we shouldn't. In order to get storage class, we will encounter storage not found, so need to error out here.
3. If has PVC and storage class is expandable, expand PVC and PV. This will resolve the issue of `Invalid volume state attached for expansion errors in the csi-resizer.log`
4. If doesn't have PVC, expand PV.

I am not sure if we should create PVC for user, since there is different access mode, and I believe it is immutable and from UI User can create PVC.

cclhsu:
@joshimoo @shuo-wu Any additional concern with current implementation?

shuo-wu:
The question is, where Longhorn can expand PVC and PV. expanding PVC will lead to the CSI resizer call the existing Expansion API. That's why I mention `Hide the old expansion API` above.

@joshimoo What do you think about the 2 solutions commented on by Clark and me?

joshimoo:
kubernetes regular expansion flow:
pvc update -> csi resizer -> csi plugin -> longhorn api expand -> expansion done -> csi resizer will expand PV with new capacity.

Since expansion calls, need to be idempotent. I think it should be okay to use a single expansion api.
How I could see this working:

longhorn api call:
- check pvc size < requested size:  -> update pvc
- check volume size < requested size -> update volume size (this should already be implemented)

Since the first api call will update the pvc, this will trigger the csi flow above.
Which will call the expansion api, since the volume/pvc size is already update that is a noop.
Then the polling in the csi plugin will wait for successful volume size update.
Once that is done, control returns to the csi resizer which will update the PV with the new size.

In the case where a PVC is present, I don't think we need to update the PV size as part of the longhorn api call.
Since it will be the responsibility of the csi resizer to do so, there is currently no harm in doing it a head of time. 
In the case where no PVC is present, updating the PV size after the volume update seems reasonable.

Tere is some upcoming discussion on resize failure handling, here, that might change this part.
https://github.com/gnufied/enhancements/blob/master/keps/sig-storage/1790-recover-resize-failure/README.md

---------------------------------------------

@cclhsu can you explain why we need to process the storage class?
We should be able to just try updating the PVC and let kubernetes handle whether that update is valid, based on it's internal criteria (storage class exists & storage class allows for volume expansion)
No need for us to concern ourselves with that evaluation since it's already handled by kubernetes during the pvc update.

That's why I also think that we should make the restored volumes use the `longhorn` storage class so that their expansion settings are tied to the user configured settings instead of the non existing `longhorn-static` class with hardcoded params.
@shuo-wu @cclhsu what do you think about this part?

cclhsu:
> * check pvc size < requested size:  -> update pvc
> check volume size < requested size -> update volume size (this should already be implemented)

Yes, this sound close to what I did in the first implementation, where UI will issue `pvcexpand` api, and added @shou suggestion to hide `original api`.

> can you explain why we need to process the storage class?

Seem like side effect when pack everything to original expand function.  I remember I encounter a situation where pvc didn't expand as state in storage class but pv was expand, so I add check to prevent it from happening. I think we can ignore it since we are moving back to first implementation.

I have rechecking the first implementation with some updates.

From my testing I have yet to encounter the scenario I need to recover.  I will try with scenario in the document.  When recover, if user change `setting.longhorn.io/default-longhorn-static-storage-class` to another storage-class, e.g. `longhorn-xfs`.  Shouldn't we use it instead of `longhorn`?

manager/kubernetes.go
```diff
@@ -115,3 +116,60 @@ func (m *VolumeManager) PVCCreate(name, namespace, pvcName string) (v *longhorn.
...
+	pvcName := v.Status.KubernetesStatus.PVCName
+	resourceSize := resource.MustParse(size)
+	if pvcName != "" {
+		pvcNamespace := v.Status.KubernetesStatus.Namespace
+		pvc, err := m.ds.GetPersisentVolumeClaim(pvcNamespace, pvcName)
+		if err != nil {
+			return nil, err
+		}
+
+		if pvc.Status.Capacity.StorageEphemeral().Cmp(resourceSize) > 0 {
```
joshimoo, line 147: I don't think this is using the right field? you probably want to check against the requested size instead?

joshimoo (commented):
I am not sure if we need a separate call?

manager/kubernetes.go
cclhsu: I was intended to use `pvc.Status.Capacity.Storage()`, but the API is only support since v0.18.x.  Currently,  `longhorn-manager` and `longhorn-share-manager` has specified version to use v0.16.15.  Do you know the reason why we didn't update it?  Should we update it?

cclhsu: Seems like it is due to https://github.com/longhorn/longhorn/issues/1938, our minimum Kubernetes requirements is 1.16+.  Without the API, there is no reason to check it.

cclhsu:
@joshimoo @shuo-wu as show in sprint demo, any additional concerns?  Please help review.

manager/volume.go
```diff
@@ -622,6 +625,38 @@ func (m *VolumeManager) Expand(volumeName string, size int64) (v *longhorn.Volum
...
+	resourceSize := resource.MustParse(strconv.FormatInt(size, 10))
+	if pvcName != "" {
+		pvcNamespace := v.Status.KubernetesStatus.Namespace
+		pvc, err := m.ds.GetPersisentVolumeClaim(pvcNamespace, pvcName)
+		if err != nil {
+			return nil, err
+		}
+
+		// Should change to pvc.Status.Capacity.Storage(), once upgrade API to v0.18.x
+		if pvc.Status.Capacity.StorageEphemeral().Cmp(resourceSize) > 0 {
```
shuo-wu, line 638: Need to check PVC spec rather than status

manager/volume.go
```diff
@@ -622,6 +625,38 @@ func (m *VolumeManager) Expand(volumeName string, size int64) (v *longhorn.Volum
...
+			return nil, fmt.Errorf("cannot expand volume %v with current size %v to a smaller size %v", v.Name, v.Spec.Size, size)
+		}
+
+		// Should change to pvc.Status.Capacity.Storage(), once upgrade API to v0.18.x
+		if pvc.Status.Capacity.StorageEphemeral().Cmp(resourceSize) == 0 {
```
shuo-wu, line 643: Need to check both spec and status

manager/volume.go
```diff
@@ -622,6 +625,38 @@ func (m *VolumeManager) Expand(volumeName string, size int64) (v *longhorn.Volum
...
+		pvc.Spec.Resources = apiv1.ResourceRequirements{
+			Requests: apiv1.ResourceList{
+				apiv1.ResourceStorage: resourceSize,
+			},
+		}
+
+		_, err = m.ds.UpdatePersisentVolumeClaim(pvcNamespace, pvc)
+		if err != nil {
+			return nil, err
+		}
```
shuo-wu, line 657: Update the PVC only if `resourceSize` is greater than the current PVC spec size.

shuo-wu (changes requested):

manager/volume.go
```diff
@@ -4,11 +4,14 @@ import (
 	"fmt"
 	"reflect"
 	"sort"
+	"strconv"
 
 	"github.com/pkg/errors"
 	"github.com/robfig/cron"
 	"github.com/sirupsen/logrus"
 
+	apiv1 "k8s.io/api/core/v1"
```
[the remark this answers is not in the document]
cclhsu: Though I was following what  defined in `manager/kubernetes.go`.  I prefer use corev1 also. Changed.

manager/volume.go
```diff
@@ -622,6 +625,38 @@ func (m *VolumeManager) Expand(volumeName string, size int64) (v *longhorn.Volum
 		return v, nil
 	}
 
+	pvcName := v.Status.KubernetesStatus.PVCName
+	resourceSize := resource.MustParse(strconv.FormatInt(size, 10))
+	if pvcName != "" {
```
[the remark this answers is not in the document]
cclhsu: changed

manager/volume.go
cclhsu: Currently, using v0.16.15, only support pvc.Spec.XXX_Size()

manager/volume.go
cclhsu: Currently, using v0.16.15, only support pvc.Spec.XXX_Size()

manager/volume.go
cclhsu: With previous conditions check, only request pvc size greater than current size will run here.

manager/volume.go
cclhsu: We are currently using api v16.15 as in https://github.com/kubernetes/api/blob/release-1.16/core/v1/resource.go, PVC spec/status didn't support `Storage` yet.  We already check disk size in earlier code.

manager/volume.go
cclhsu: We are currently using api v16.15 as in https://github.com/kubernetes/api/blob/release-1.16/core/v1/resource.go, PVC spec/status didn't support `Storage` yet.  We already check disk size in earlier code.

manager/volume.go
```diff
@@ -613,6 +616,46 @@ func (m *VolumeManager) Expand(volumeName string, size int64) (v *longhorn.Volum
...
 	}
 
+	kubernetesStatus := &v.Status.KubernetesStatus
+	if kubernetesStatus.PVCName != "" && kubernetesStatus.LastPVCRefAt == "" {
+		pvc, err := m.ds.GetPersistentVolumeClaim(kubernetesStatus.Namespace, kubernetesStatus.PVCName)
+		if err != nil {
+			return nil, err
+		}
+
+		resourceSize := resource.MustParse(strconv.FormatInt(size, 10))
```
innobead, line 626: Suggest changing to `requestedSize`

manager/volume.go
```diff
@@ -613,6 +616,46 @@ func (m *VolumeManager) Expand(volumeName string, size int64) (v *longhorn.Volum
...
+
+		// TODO: Should check for pvc.Spec.Resources.Requests.Storage() here, once upgrade API to v0.18.x.
```
innobead, line 628: Please do map index check to avoid null pointer error.

innobead, line 628: The same, map index check

innobead, line 628: Please just do two conditions to avoid missing conditions to flow to the original logic.

if storageSpecValue.Cmp(resourceSize) < 0 {
... 
} else {
... // the current volume size is large or eqaul to the requrested size.
}

innobead (changes requested):

manager/volume.go
```diff
@@ -613,6 +616,46 @@ func (m *VolumeManager) Expand(volumeName string, size int64) (v *longhorn.Volum
...
+			}
+
+			logrus.Infof("Persistent Volume Claim %v expansion from %v to %v requested", v.Name, v.Spec.Size, size)
+			_, err = m.ds.UpdatePersistentVolumeClaim(kubernetesStatus.Namespace, pvc)
+			if err != nil {
+				return nil, err
+			}
+
+			return v, nil
+		}
```
innobead, line 656: Please just do two conditions to avoid missing conditions to flow to the original logic.

if storageSpecValue.Cmp(resourceSize) < 0 {
... 
} else {
... // the current volume size is large or eqaul to the requrested size.
}

manager/volume.go
```diff
@@ -613,6 +616,46 @@ func (m *VolumeManager) Expand(volumeName string, size int64) (v *longhorn.Volum
...
+		if err != nil {
+			return nil, err
+		}
+
+		resourceSize := resource.MustParse(strconv.FormatInt(size, 10))
+
+		// TODO: Should check for pvc.Spec.Resources.Requests.Storage() here, once upgrade API to v0.18.x.
+		storageSpecValue := pvc.Spec.Resources.Requests[corev1.ResourceStorage]
+		// TODO: Should check for pvc.Status.CapacityStorage() here, once upgrade API to v0.18.x.
+		storageStatusValue := pvc.Status.Capacity[corev1.ResourceStorage]
```
innobead, line 631: The same, map index check

manager/volume.go
```diff
@@ -613,6 +616,46 @@ func (m *VolumeManager) Expand(volumeName string, size int64) (v *longhorn.Volum
...
+	if kubernetesStatus.PVCName != "" && kubernetesStatus.LastPVCRefAt == "" {
+		pvc, err := m.ds.GetPersistentVolumeClaim(kubernetesStatus.Namespace, kubernetesStatus.PVCName)
+		if err != nil {
+			return nil, err
+		}
+
+		resourceSize := resource.MustParse(strconv.FormatInt(size, 10))
+
+		// TODO: Should check for pvc.Spec.Resources.Requests.Storage() here, once upgrade API to v0.18.x.
+		storageSpecValue := pvc.Spec.Resources.Requests[corev1.ResourceStorage]
```
innobead, line 629: Please do map index check to avoid null pointer error.

innobead (changes requested):

manager/volume.go

manager/volume.go

manager/volume.go

manager/volume.go

manager/volume.go
```diff
@@ -613,6 +616,57 @@ func (m *VolumeManager) Expand(volumeName string, size int64) (v *longhorn.Volum
...
+				return v, nil
+			}
+
+			// condition where storageSpecValue == requestedSize and v.Spec.Size < size should continue for CSI expand call
+		}
+	}
+
 	if v.Spec.Size > size {
 		return nil, fmt.Errorf("cannot expand volume %v with current size %v to a smaller size %v", v.Name, v.Spec.Size, size)
 	}
```
shuo-wu, line 672:
```suggestion
	if v.Spec.Size > size {
		return nil, fmt.Errorf("longhorn volume spec size %v is greater than requested PVC size %v", v.Spec.Size, size)
	}
	if v.Status.CurrentSize > size {
		return nil, fmt.Errorf("longhorn volume current size %v is greater than requested PVC size %v", v.Status.CurrentSize, size)
	}
	
	kubernetesStatus := &v.Status.KubernetesStatus
	if kubernetesStatus.PVCName != "" && kubernetesStatus.LastPVCRefAt == "" {
		pvc, err := m.ds.GetPersistentVolumeClaim(kubernetesStatus.Namespace, kubernetesStatus.PVCName)
		if err != nil {
			return nil, err
		}

		requestedSize := resource.MustParse(strconv.FormatInt(size, 10))

		// TODO: Should check for pvc.Spec.Resources.Requests.Storage() here, once upgrade API to v0.18.x.
		storageSpecValue, ok := pvc.Spec.Resources.Requests[corev1.ResourceStorage]
		if !ok {
			return nil, fmt.Errorf("cannot get request storage")
		}
		// TODO: Should check for pvc.Status.CapacityStorage() here, once upgrade API to v0.18.x.
		storageStatusValue, ok := pvc.Status.Capacity[corev1.ResourceStorage]
		if !ok {
			return nil, fmt.Errorf("cannot get capacity storage")
		}

		// Check if the requested size is valid for PVC expansion
		if storageSpecValue.Cmp(requestedSize) > 0 {
			return nil, fmt.Errorf("PVC spec size %v is greater than requested PVC size %v", ......)
		}
		if storageStatusValue.Cmp(requestedSize) > 0 {
			return nil, fmt.Errorf("PVC status size %v is greater than requested PVC size %v", ......)
		}

		// Expand PVC first
		if storageSpecValue.Cmp(requestedSize) < 0 {
			pvc.Spec.Resources = corev1.ResourceRequirements{
				Requests: corev1.ResourceList{
					corev1.ResourceStorage: requestedSize,
				},
			}
			if _, err = m.ds.UpdatePersistentVolumeClaim(kubernetesStatus.Namespace, pvc); err != nil {
				return nil, err
			}
			logrus.Infof()
			// CSI plugin will call this API for Longhorn volume expansion later
			return v, nil
		}	
	}
		
	// Start to do volume expansion only when storageSpecValue == requestedSize
	if v.Spec.Size == size {
		logrus.Infof()
		return v, nil
	}
		
	v.Spec.Size = size
	......
```

shuo-wu (changes requested):

manager/volume.go

manager/volume.go
```diff
@@ -614,7 +617,59 @@ func (m *VolumeManager) Expand(volumeName string, size int64) (v *longhorn.Volum
...
 
 	if v.Spec.Size > size {
-		return nil, fmt.Errorf("cannot expand volume %v with current size %v to a smaller size %v", v.Name, v.Spec.Size, size)
+		return nil, fmt.Errorf("longhorn volume %v spec size %v is greater than requested Persistent Volume Claim size %v", v.Name, v.Spec.Size, size)
+
```
innobead, line 621: Remove needless empty line.

manager/volume.go
```diff
@@ -614,7 +617,59 @@ func (m *VolumeManager) Expand(volumeName string, size int64) (v *longhorn.Volum
...
+	}
+
+	if v.Status.ActualSize > size {
```
innobead, line 624: Why we need to check status.ActualSize? I assume we only need to tackle spec.Size.

manager/volume.go
```diff
@@ -614,7 +617,59 @@ func (m *VolumeManager) Expand(volumeName string, size int64) (v *longhorn.Volum
...
+	if kubernetesStatus.PVCName != "" && kubernetesStatus.LastPVCRefAt == "" {
+		pvc, err := m.ds.GetPersistentVolumeClaim(kubernetesStatus.Namespace, kubernetesStatus.PVCName)
+		if err != nil {
+			return nil, err
+		}
+
+		requestedSize := resource.MustParse(strconv.FormatInt(size, 10))
+
+		// TODO: Should check for pvc.Spec.Resources.Requests.Storage() here, once upgrade API to v0.18.x.
+		storageSpecValue, ok := pvc.Spec.Resources.Requests[corev1.ResourceStorage]
```
innobead, line 638: Suggest change `storageSpecValue` to `pvcSpecValue`

manager/volume.go
```diff
@@ -614,7 +617,59 @@ func (m *VolumeManager) Expand(volumeName string, size int64) (v *longhorn.Volum
...
+		if !ok {
+			return nil, fmt.Errorf("cannot get request storage")
+		}
+		// TODO: Should check for pvc.Status.CapacityStorage() here, once upgrade API to v0.18.x.
+		storageStatusValue, ok := pvc.Status.Capacity[corev1.ResourceStorage]
```
innobead, line 643: Why we need to check status value?

manager/volume.go
```diff
@@ -614,7 +617,59 @@ func (m *VolumeManager) Expand(volumeName string, size int64) (v *longhorn.Volum
...
+		if !ok {
+			return nil, fmt.Errorf("cannot get capacity storage")
+		}
+
+		if storageSpecValue.Cmp(requestedSize) == 0 && storageStatusValue.Cmp(requestedSize) == 0 && v.Spec.Size == size {
+			logrus.Infof("Persistent Volume Claim %v expansion is not necessary since current size %v == %v", v.Name, v.Spec.Size, size)
+			return v, nil
+		} else if storageSpecValue.Cmp(requestedSize) > 0 {
+			return nil, fmt.Errorf("persistent volume claim %v spec size %v is greater than requested persistent volume claim size %v", kubernetesStatus.PVCName, storageSpecValue.Value(), requestedSize.Value())
+		} else if storageStatusValue.Cmp(requestedSize) > 0 {
```
innobead, line 653: Why we ned to check status value?

manager/volume.go
```diff
@@ -614,7 +617,59 @@ func (m *VolumeManager) Expand(volumeName string, size int64) (v *longhorn.Volum
...
+		if !ok {
+			return nil, fmt.Errorf("cannot get request storage")
+		}
+		// TODO: Should check for pvc.Status.CapacityStorage() here, once upgrade API to v0.18.x.
+		storageStatusValue, ok := pvc.Status.Capacity[corev1.ResourceStorage]
+		if !ok {
+			return nil, fmt.Errorf("cannot get capacity storage")
+		}
+
+		if storageSpecValue.Cmp(requestedSize) == 0 && storageStatusValue.Cmp(requestedSize) == 0 && v.Spec.Size == size {
```
innobead, line 648: Should we just check `storageSpecValue.Cmp(requestedSize) == 0`?

manager/volume.go
```diff
@@ -614,7 +617,59 @@ func (m *VolumeManager) Expand(volumeName string, size int64) (v *longhorn.Volum
...
+			logrus.Infof("Persistent Volume Claim %v expansion is not necessary since current size %v == %v", v.Name, v.Spec.Size, size)
+			return v, nil
+		} else if storageSpecValue.Cmp(requestedSize) > 0 {
+			return nil, fmt.Errorf("persistent volume claim %v spec size %v is greater than requested persistent volume claim size %v", kubernetesStatus.PVCName, storageSpecValue.Value(), requestedSize.Value())
+		} else if storageStatusValue.Cmp(requestedSize) > 0 {
+			return nil, fmt.Errorf("persistent volume claim %v status size %v is greater than requested persistent volume claim size %v", kubernetesStatus.PVCName, storageStatusValue.Value(), requestedSize.Value())
+		} else if storageSpecValue.Cmp(requestedSize) < 0 {
```
innobead, line 655: Just need to `else {` good enough. 

nit: for multiple if-else, you can probably use `switch` w/ no condition statement.

manager/volume.go
```diff
@@ -614,7 +617,59 @@ func (m *VolumeManager) Expand(volumeName string, size int64) (v *longhorn.Volum
...
+			logrus.Infof("Persistent Volume Claim %v expansion from %v to %v requested", v.Name, v.Spec.Size, size)
+			_, err = m.ds.UpdatePersistentVolumeClaim(kubernetesStatus.Namespace, pvc)
+			if err != nil {
+				return nil, err
+			}
+
+			// return and CSI plugin call this API later for Longhorn volume expansion
+			return v, nil
+		} else {
+			logrus.Infof("CSI plugin call to expand volume %v", v.Name)
```
innobead, line 671: Why we need this?

innobead (dismissed):

manager/volume.go
cclhsu: changed

manager/volume.go
cclhsu: changed

manager/volume.go
cclhsu: I remember during testing PVC status storage size didn't change till PV storage size is changed.  When failed, PVC storage request has updated value and PVC storage status has older value.  Use it to check for corner cases.

manager/volume.go
cclhsu: End User can actually do this to LHV and it does not reflect in PV and PVC.
```
LHV=test
kubectl patch -p '{"spec":{"size": "3Gi"}}' volume.longhorn.io ${LHV} --namespace longhorn-system --type=merge
```

manager/volume.go
cclhsu: removed else

manager/volume.go
cclhsu: removed else

manager/volume.go
cclhsu: removed.

manager/volume.go
cclhsu: if user manually created pvc, we can see request size 1Gi and status size is 2Gi

manager/volume.go
innobead: If the user changes the size of LHV, it should be reconciled by the controller instead of API, right? Here, it should be simple, if users request the API (by UI), we just respect the requested value. After that, all things should be handled by the CSI resizer. 

https://github.com/longhorn/longhorn-manager/blob/713e2e48b6f7fe191ccd36d33b41fb2e5a234811/controller/volume_controller.go#L1850-L1850

manager/volume.go
innobead: If users request 1G, but the existing size is bigger than it. The following logic should be handled by the Longhorn CSI plugin. 

Yes, it's good to tackle here, but actually, the backend is also taking care of. I am fine with no need to change. 

https://github.com/longhorn/longhorn-manager/blob/713e2e48b6f7fe191ccd36d33b41fb2e5a234811/csi/controller_server.go#L672-L682

manager/volume.go
innobead: As I said in https://github.com/longhorn/longhorn-manager/pull/888#discussion_r643003422, but I am fine with keeping it.

manager/volume.go
```diff
@@ -614,7 +617,58 @@ func (m *VolumeManager) Expand(volumeName string, size int64) (v *longhorn.Volum
...
+			_, err = m.ds.UpdatePersistentVolumeClaim(kubernetesStatus.Namespace, pvc)
+			if err != nil {
+				return nil, err
+			}
+
+			// return and CSI plugin call this API later for Longhorn volume expansion
+			return v, nil
+		}
+		// all other case belong to the CSI plugin call
+		logrus.Infof("CSI plugin call to expand volume %v", v.Name)
```
innobead, line 671: Because you already tackle in PVC case, you should return here and the rest of the part will be tackled by the CSI plugin. Correct?

innobead, line 671: You already tackle the PVC case, so you should return from here instead of going through the original following process.

cclhsu: We are using this function for both PVC and CSI call.  When CSI call this function, all previous condition should pass and continue with remaining function.  If we return here, CSI part of expanding function will never be called.

innobead: Thanks, missed that.

manager/volume.go
```diff
@@ -614,7 +617,58 @@ func (m *VolumeManager) Expand(volumeName string, size int64) (v *longhorn.Volum
...
+		if pvcSpecValue.Cmp(requestedSize) == 0 && pvcStatusValue.Cmp(requestedSize) == 0 && v.Spec.Size == size {
+			// lhv size does not reflect pvc/pv size
+			logrus.Infof("Persistent Volume Claim %v expansion is not necessary since current size %v == %v", v.Name, v.Spec.Size, size)
+			return v, nil
+		} else if pvcSpecValue.Cmp(requestedSize) > 0 {
+			// this is possible due to user can patch lvh, but not patch pv and pvc
+			return nil, fmt.Errorf("persistent volume claim %v spec size %v is greater than requested persistent volume claim size %v", kubernetesStatus.PVCName, pvcSpecValue.Value(), requestedSize.Value())
+		} else if pvcStatusValue.Cmp(requestedSize) > 0 {
+			// this is possible due to user manually created pvc
+			return nil, fmt.Errorf("persistent volume claim %v status size %v is greater than requested persistent volume claim size %v", kubernetesStatus.PVCName, pvcStatusValue.Value(), requestedSize.Value())
```
joshimoo, line 652: Remove all of these conditions, non of these are our concern.

The only concerns that require updates are:
if pvc.size < requestedSize: -> update PVC
if volume.size < requestedSize: -> update volume

joshimoo (changes requested):
I think you can simplify by removing additional conditions, that we are not responsible for.

manager/volume.go
```diff
@@ -614,7 +617,58 @@ func (m *VolumeManager) Expand(volumeName string, size int64) (v *longhorn.Volum
 	}
 
 	if v.Spec.Size > size {
-		return nil, fmt.Errorf("cannot expand volume %v with current size %v to a smaller size %v", v.Name, v.Spec.Size, size)
+		return nil, fmt.Errorf("longhorn volume %v spec size %v is greater than requested Persistent Volume Claim size %v", v.Name, v.Spec.Size, size)
+	}
```
joshimoo, line 621: If we want to keep this behavior, you should move it below the PVC processing.
Since that way the user can use the longhorn api to update PVC, even if the volume is already bigger.

joshimoo: I personally feel we should return success in this case instead of an error, since the API is called Expand instead of Resize.
It's true we can never shrink a volume, but I also don't see why we couldn't return success as in:

- We ensure that the capacity of the volume is at least the requested user capacity.
- In a sense we already operate on in this way, since we need to round the user supplied capacity up to a multiple of our block size.
- The other benefit of this change is that we are in sync with kubernetes behavior and don't need the special casing in the csi driver.

We discussed this previously though and since it's a change in behavior we agreed on not changing it at that point in time.
Now that we are adding additional behavior to this API I think it's fair to also consider this case. Since the api behavior is changing already -> i.e early exit because of PVC, relying on the CSI driver to update the longhorn volume size.

joshimoo: cc @shuo-wu @innobead

cclhsu: My main concern is to make sure that when user call API, workload/volume will not go into a state where it cannot attach as in the issue.
If I remember correctly, when PVC Size < LHV Size, workload's volume attachment will fail.  Therefore, request size need to be greater than LHV Size.
If we move this check after PVC expand and request size is greater than PVC size and less than LHV size, the scenario will still occur.  
I know it is good by design to move it after PVC call, but do we want to leave it to the user to resolve the issue manually?

manager/volume.go
cclhsu: Same as above. Some scenarios I found if user edit LHV/PVC manifest manually, it will cause LVH and PVC/PV spec/status size inconsistent so the purpose of it is to prevent it from happen, if it is not our concern, I am ok to remove it.

manager/volume.go
joshimoo: I don't understand your concern, can you clarify.

These are the 2 cases that require updates:
- if pvc.size < requestedSize: -> update PVC
- if volume.size < requestedSize: -> update volume

From the logic these 2 should be handled in this order:
- PVC evaluation & update
- Volume evaluation & update

You can even do both of the updates in the same api call, since the expansion call needs to be idempotent.

manager/volume.go
joshimoo: I think there might be a misunderstanding on the conceptual level here:
- A longhorn volume represents a PV
- A PVC is a claim for a PV (the right to exclusively use that PV)

It's valid if a PV.size > PVC.size, for example in a non dynamic provisioning scenario:
- cluster has a PV of size 10GB defined
- user creates a PVC with requested size 1GB, the cluster will associate that PVC (1GB) with the above PV (10GB)

From the above we can see that it should also be valid if Longhorn.Volume.size (actual capacity) > PVC.size (requested capacity)

manager/volume.go
cclhsu: I know it might not be a normal UI case, but I am a little hackish and I like to edit manifest through scripting and use UI as last resort. (And yes, UI has done some prevention) and  I am not sure if there are other scenarios like this.

Assuming original LHV/PVC size is 1Gi, editing Longhorn volume by manifest

```
kubectl patch -p '{"spec":{"size": "3Gi"}}' volume.longhorn.io ${LHV} --namespace longhorn-system --type=merge
```

See that PV/PVC didn't update and try to call longhorn API

```
curl -sSX POST -H 'Content-Type: application/json' "http://${BIND_URL}:${BIND_PORT}/v1/volumes/${PV}?action=${ACTION}" -d '{"size":"2Gi"}' | jq
```

In such scenario, request size will be greater than PVC size but less than LHV size. With current implementation, the user will receive an error.  If move after PVC checking, I remember PVC request will change but PVC/PV status will never update nor receive error. UI will display LHV as 3Gi, but workload would not able to attached.  Need to increase request size to 4Gi in UI for workload/volume to get back to normal state.

joshimoo: So what happens if you do the above? I don't expect any problem, if there is then that is a bug.
- Create Volume (1G)
- Workload scale down
- Edit lh.spec.Size via kubectl (3G)
- Wait for expansion, you can scale up before, but it will not be ready
- Workload scale up (should work, even though LH size > pvc size)
- Workload scale down
- Edit pvc.size via kubectl (2G)
- this should go through the csi expansion process
- return success PV should become (3G) while pvc (2G)
- Workload scale up (should work, even though LH size > pvc size)

Are you seeing any failure in the above process?

manager/volume.go
shuo-wu: One quick question, should we continue trying to expand Longhorn volume even if the requested size is smaller than the PVC spec size?

manager/volume.go
shuo-wu: Considering the scenario Joshua mentioned, allowing Longhorn to return success when the requested size is smaller than the volume spec size makes sense.

cclhsu: Update the code as (in the latest code)

> if pvc.size < requestedSize: -> update PVC
> if volume.size < requestedSize: -> update volume

And go through the steps

> Create Volume (1Gi)
> Workload scale down
> Edit lh.spec.Size via kubectl (3Gi)
> Wait for expansion, you can scale up before, but it will not be ready
> Workload scale up (should work, even though LH size > pvc size)
> Workload scale down
> Edit pvc.size via kubectl (2Gi)
> this should go through the csi expansion process
> return success PV should become (3Gi) while pvc (2Gi)
> Workload scale up (should work, even though LH size > pvc size)

Workload is able to attached volume and see storage volume as following in the end:

```
PVC
resources.request.storage = 2Gi
status.capacity.storage = 3Gi
PV
status.capacity.storage = 3Gi
LHV
spec.size=3221225472
```

joshimoo: Great as shuo mentioned above we can combine the v.size > requested size and v.size == requestedSize to return the current volume instead of error.

Let's have a quick sync up tomorrow morning my time then we can merge this. @shuo-wu anything missing from your perspective?

manager/volume.go
joshimoo: Yeah I think so, this case should only happen in case of restoration of a smaller volume backup 

we could also do size = max(requestedSize, PVC.requestedSize) 

To cover the restore case but it would actually need to happen as part of the controller to reconcile v.size with the PVC.size.

So we can ignore the controller part for now and do the max(x,y) approach.

------------

For the normal part in which the Cai driver will be called by reconciliation of PVC.size > pv.size this would be covered already, but doing the max approach is beneficial since it allows the user via an expand call to bring a restored volume back in line with the old PVC/PV 

On the CSi expand call since we return the volume.size which will end up in the PV. Kubernetes should end up call expand again while reconciling PVC.requested against the PV.size in the future.

--------

tldr doing size = max(pvc.requestedSize, requestedSize) should have no downaide

shuo-wu: Do you mean restore/create a volume with small size with existing old PVC? The actual storage size being smaller than the claim desired size is really weird. And doing max is tricky and misunderstanding. e.g., If the requested new size is smaller than both volume spec size and PVC spec size, Longhorn will return success for this expansion call. In the user's perspective, PVC **or/and** volume somehow can be "expanded" to a smaller size since there is no error.

manager/volume.go
cclhsu: updated

manager/volume.go
cclhsu: updated

manager/volume.go
```diff
@@ -613,12 +616,48 @@ func (m *VolumeManager) Expand(volumeName string, size int64) (v *longhorn.Volum
 		return nil, fmt.Errorf("cannot expand volume before replica scheduling success")
 	}
 
-	if v.Spec.Size > size {
-		return nil, fmt.Errorf("cannot expand volume %v with current size %v to a smaller size %v", v.Name, v.Spec.Size, size)
+	kubernetesStatus := &v.Status.KubernetesStatus
```
joshimoo, line 619: Add rounding here size = util.RoundUpSize(size)
This rounds the PVC assigned size up to align with the longhorn blockvolume size.
It's a nice to have but not a requirement.

Below before assigning to the v.spec.Size you do need to roundUp, see below comment.

manager/volume.go
```diff
@@ -613,12 +616,48 @@ func (m *VolumeManager) Expand(volumeName string, size int64) (v *longhorn.Volum
...
+			size = pvcSpecValue.Value()
+		}
 	}
 
-	if v.Spec.Size == size {
-		logrus.Infof("Volume %v expansion is not necessary since current size %v == %v", v.Name, v.Spec.Size, size)
+	if v.Spec.Size >= size {
+		logrus.Infof("Volume %v expansion is not necessary since current size %v >= %v", v.Name, v.Spec.Size, size)
 		return v, nil
 	}
```
joshimoo, line 663: Add rounding up for the volumeSize so it's a multiple of blockSize size = util.RoundUpSize(size)

joshimoo (commented):
Thanks for the hard work on this PR, once you add the roundUp of size to a multiple of blockSize, see comment this can be merged.

manager/volume.go
cclhsu: Changed. Instead of before assignment, roundup before size is used to do comparison.

manager/volume.go
cclhsu: changed

joshimoo (approved):
LGTM

joshimoo:
@shuo-wu @innobead I dismissed the stale reviews, since we discussed the solution during PR team meeting, so I can merge the PR.

PR #889, opened 2021-05-06
manager: Auto correct backing image name during creation

shuo-wu:
If the backing image name is longer than 63 char, Longhorn cannot put it in the replica label. And Longhorn cannot list the replicas using the backing image name with the name.

longhorn/longhorn#2522


jenting (approved):

innobead (approved):
LGTM

innobead:
@shuo-wu please cherry-pick this commit to v1.1.2 branch because this needs to be backported.

PR #892, opened 2021-05-11
Cherry-pick: Auto correct backing image name during creation

shuo-wu:
longhorn/longhorn#2522


shuo-wu:
Sorry for that. I have updated the base.

innobead (approved):

PR #893, opened 2021-05-17
Backing Image Upload

shuo-wu:
longhorn/longhorn#2404

Workflow:
1. Users create a backing image by uploading a local file via UI.
2. UI calls the backing image creation API then waits for backing image field `uploadAddress` to become non-empty.
3. Longhorn manager pods will ask the backing image manager pods to launch an HTTP upload server for the backing image after receiving the backing image creation call. When the upload server is launched, backing image field `uploadAddress` will be set.
4. UI starts to call the upload-related APIs for the backing image. Longhorn manager pod will forward the requests to the upload server in the backing image manager pods:
    1. Action `uploadServerStart` will ask the upload server to create a tmp file and set the total size for the backing image. This means upload start.
    2. Action `chunkPrepare` is used to check if each chunk is already uploaded or not. If the chunk is not correct, it will do clean up first. Then a chunk file will be created. This API will be called before uploading each chunk.
    3. Action `chunkUpload` is for chunk upload. This will write the actual chunk data to the chunk file.
    4. After uploading all chunks, action `chunkCoalesce` will be invoked. This will copy all chunk data into the tmp file created by `uploadServerStart` then clean up all chunk files.
    5. Finally, action `uploadServerClose` is to indicate upload complete.
5.  If the upload is interrupted. UI can retry upload for the users. With action `chunkPrepare`, Longhorn can skip re-uploading existing chunks.

Besides, for uploaded backing images. Longhorn cannot clean up all downloaded files even if the backing image timeout is reached. And for the backing image cleanup API, Longhorn needs to avoid users force deleting all downloaded files. If they want to clean up all files, they can directly delete the backing image.

### DIFF
--- a/api/forwarder.go
+++ b/api/forwarder.go
@@ -9,11 +9,17 @@ import (
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 
+	v1 "k8s.io/api/core/v1"
+
+	"github.com/longhorn/longhorn-manager/engineapi"
 	"github.com/longhorn/longhorn-manager/manager"
 	"github.com/longhorn/longhorn-manager/types"
 )
 
 type OwnerIDFunc func(req *http.Request) (string, error)
+type BackingImageUploadServerAddressFunc func(req *http.Request) (string, error)
+type GetTargetAddressFunc func(req *http.Request) (string, error)
+type ProxyRequestHandler func(address string, req *http.Request) (proxyRequired bool, err error)
 
 func OwnerIDFromVolume(m *manager.VolumeManager) func(req *http.Request) (string, error) {
 	return func(req *http.Request) (string, error) {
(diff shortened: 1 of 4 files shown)


shuo-wu:
Only need to review commit `api: Add proxy handling for backing image upload requests`. Others are actually the code of PR https://github.com/longhorn/longhorn-manager/pull/953

joshimoo (commented):
LGTM, can be merged after we merged https://github.com/longhorn/longhorn-manager/pull/953 Then this will only be a single commit.

At some point w should investigate whether we can restructure the managers so that they can handle any calls and we don't need forwarding to a manager that owns a volume for interacting with the api.

For the backing image we still would need forwarding to the internal data source pod for the moment for uploads.

Have you tested Large uploads > 1 min? 
I think I remember faintly that there was some timeouts at the api layer on the server side, but I don't remember the details.

shuo-wu:
> Have you tested Large uploads > 1 min?
> I think I remember faintly that there was some timeouts at the api layer on the server side, but I don't remember the details.

YES. Typically the timeout is controlled by the client (HTTP client or the engine binary caller). The timeout you mentioned might be the engine binary execution timeout.

innobead (approved):
LGTM

PR #896, opened 2021-05-26
Implement replica auto balance

c3y1huang:
Add a new Longhorn global setting `Replica Auto Balance` to enable detection and deletion of the unbalanced replica to achieve automatic rebalancing.

Also a new setting in volume spec `replicaAutoBalance` to enable/disable automatic replica rebalancing for individual volume.

https://github.com/longhorn/longhorn/issues/587


shuo-wu (changes requested):
Can you split this PR into more commits? e.g.:
1. Add a new setting
2. Add a new HTTP API
3. Update package `client` for the new API
4. Update csi
5. controller: Apply auto rebalance

I will continue reviewing the controller part tomorrow.

types/setting.go
```diff
@@ -302,6 +305,25 @@ var (
 		Default:     "false",
 	}
 
+	SettingDefinitionReplicaAutoBalance = SettingDefinition{
+		DisplayName: "Replica Auto Balance",
+		Description: "Enable this setting automatically rebalances replicas when discover an available node.\n\n" +
```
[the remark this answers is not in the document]
c3y1huang: Thanks, addressed.

manager/volume.go
```diff
@@ -905,6 +914,32 @@ func (m *VolumeManager) UpdateReplicaCount(name string, count int) (v *longhorn.
...
+func (m *VolumeManager) UpdateReplicaAutoBalance(name string, inputSpec types.ReplicaAutoBalance) (v *longhorn.Volume, err error) {
+	defer func() {
+		err = errors.Wrapf(err, "unable to update replica auto-balance for volume %v", name)
+	}()
+
+	v, err = m.ds.GetVolume(name)
+	if err != nil {
+		return nil, err
+	}
+
```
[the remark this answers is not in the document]
c3y1huang: Thanks, fixed

c3y1huang:
> Can you split this PR into more commits? e.g.:

Addressed.

controller/volume_controller.go
```diff
@@ -115,6 +117,7 @@ func NewVolumeController(
 		kubeClient:    kubeClient,
 		eventRecorder: eventBroadcaster.NewRecorder(scheme, v1.EventSource{Component: "longhorn-volume-controller"}),
 
+		nStoreSynced:  nodeInformer.Informer().HasSynced,
```
joshimoo, line 120: your missing the wait for CacheSync on this one, below.

client/generated_backing_image.go
```diff
@@ -11,9 +11,13 @@ type BackingImage struct {
 
 	DiskStateMap map[string]string `json:"diskStateMap,omitempty" yaml:"disk_state_map,omitempty"`
 
+	DownloadProgressMap map[string]string `json:"downloadProgressMap,omitempty" yaml:"download_progress_map,omitempty"`
```
joshimoo, line 14: why is this in your commit/pr?

api/router.go
```diff
@@ -66,16 +66,17 @@ func NewRouter(s *Server) *mux.Router {
...
-		"activate":           s.VolumeActivate,
-		"expand":             s.VolumeExpand,
-		"cancelExpansion":    s.VolumeCancelExpansion,
+		"attach":                   s.VolumeAttach,
+		"detach":                   s.VolumeDetach,
+		"salvage":                  s.VolumeSalvage,
+		"recurringUpdate":          s.VolumeRecurringUpdate,
+		"updateReplicaCount":       s.VolumeUpdateReplicaCount,
+		"updateReplicaAutoBalance": s.VolumeUpdateReplicaAutoBalance,
+		"updateDataLocality":       s.VolumeUpdateDataLocality,
```
joshimoo, line 75: nit: put a space around this, keep the replica stuff in their own sections and cleans up the diff.
I despise the go auto fmt for keeping them in line, just leads to messy diffs/git blame when trying to see when a field was added

api/model.go
```diff
@@ -22,28 +22,29 @@ import (
...
-	NodeSelector            []string               `json:"nodeSelector"`
+	Name                    string                   `json:"name"`
+	Size                    string                   `json:"size"`
+	Frontend                types.VolumeFrontend     `json:"frontend"`
+	DisableFrontend         bool                     `json:"disableFrontend"`
+	FromBackup              string                   `json:"fromBackup"`
+	NumberOfReplicas        int                      `json:"numberOfReplicas"`
+	ReplicaAutoBalance      types.ReplicaAutoBalance `json:"replicaAutoBalance"`
+	DataLocality            types.DataLocality       `json:"dataLocality"`
+	StaleReplicaTimeout     int                      `json:"staleReplicaTimeout"`
```
joshimoo, line 33: nit: put a space around these to keep the replica stuff together in their own section.

controller/volume_controller.go
```diff
@@ -771,33 +790,101 @@ func (vc *VolumeController) cleanupFailedToScheduledReplicas(v *longhorn.Volume,
...
+
+	cleaned, err = vc.cleanupDataLocalityReplicas(v, e, rs)
+	if err != nil {
+		return err
+	}
+	if cleaned {
+		return nil
+	}
+
+	return nil
```
joshimoo, line 822: nit: err will be nil when cleaned == true, no need for the spaghetti
```suggestion
	if cleaned, err = vc.cleanupEvictionRequestedReplicas(v, rs); err != nil || cleaned {
		return err
	}
	
	if cleaned, err = vc.cleanupAutoBalancedReplicas(v, rs); err != nil || cleaned {
		return err
	}
	
	if cleaned, err = vc.cleanupDataLocalityReplicas(v, e, rs); err != nil || cleaned {
		return err
	}
	
	return nil
```

controller/volume_controller.go
```diff
@@ -771,33 +790,101 @@ func (vc *VolumeController) cleanupFailedToScheduledReplicas(v *longhorn.Volume,
...
+
+func (vc *VolumeController) cleanupEvictionRequestedReplicas(v *longhorn.Volume, rs map[string]*longhorn.Replica) (bool, error) {
+	for _, r := range rs {
+		if r.Status.EvictionRequested {
+			if err := vc.deleteReplica(r, rs); err != nil {
+				vc.eventRecorder.Eventf(v, v1.EventTypeWarning,
+					EventReasonFailedEviction,
+					"volume %v failed to evict replica %v",
+					v.Name, r.Name)
+				return false, err
```
joshimoo, line 833: I am wondering if this can be done as best effort?

controller/volume_controller.go
```diff
@@ -771,33 +790,101 @@ func (vc *VolumeController) cleanupFailedToScheduledReplicas(v *longhorn.Volume,
...
+			logrus.Debugf("Evicted replica %v", r.Name)
+			return true, nil
+		}
+	}
+	return false, nil
+}
+
+func (vc *VolumeController) cleanupAutoBalancedReplicas(v *longhorn.Volume, rs map[string]*longhorn.Replica) (bool, error) {
+	log := getLoggerForVolume(vc.logger, v)
+	log = log.WithField("replicaAutoBalanceType", "delete")
```
joshimoo, line 844: nit: can be combined no need for an independent assignment.
```suggestion
	log := getLoggerForVolume(vc.logger, v).WithField("replicaAutoBalanceType", "delete")
```

controller/volume_controller.go
```diff
@@ -808,12 +895,32 @@ func (vc *VolumeController) cleanupExtraHealthyReplicas(v *longhorn.Volume, e *l
...
+	globalSetting, err := vc.ds.GetSettingValueExisted(types.SettingNameReplicaAutoBalance)
+	if err != nil {
+		return "", err
+	}
+
+	if globalSetting == string(types.ReplicaAutoBalanceIgnored) {
+		globalSetting = string(types.ReplicaAutoBalanceDisabled)
+	}
+
+	return types.ReplicaAutoBalance(globalSetting), nil
```
joshimoo, line 923: What's the difference between Ignored/Disabled?
I am guessing volumes that don't have one specified will fallback to the global setting?

types/resource.go
```diff
@@ -74,25 +83,26 @@ const (
...
-	LastAttachedBy          string         `json:"lastAttachedBy"`
-	AccessMode              AccessMode     `json:"accessMode"`
-	Migratable              bool           `json:"migratable"`
+	Size                    int64              `json:"size,string"`
+	Frontend                VolumeFrontend     `json:"frontend"`
+	FromBackup              string             `json:"fromBackup"`
+	NumberOfReplicas        int                `json:"numberOfReplicas"`
+	ReplicaAutoBalance      ReplicaAutoBalance `json:"replicaAutoBalance"`
+	DataLocality            DataLocality       `json:"dataLocality"`
+	StaleReplicaTimeout     int                `json:"staleReplicaTimeout"`
```
joshimoo, line 92: nit: put them in their own section, so the diff is smaller and we have an easier git blame time to see when a field was added.

types/types.go
```diff
@@ -432,6 +432,21 @@ func ValidateReplicaCount(count int) error {
...
+func ValidateReplicaAutoBalance(option ReplicaAutoBalance) error {
+	switch option {
+	case ReplicaAutoBalanceIgnored:
+		return nil
+	case ReplicaAutoBalanceDisabled:
+		return nil
+	case ReplicaAutoBalanceLeastEffort:
+		return nil
+	case ReplicaAutoBalanceBestEffort:
+		return nil
```
joshimoo, line 444: nit: You can use a multi value case
```suggestion
	case ReplicaAutoBalanceIgnored, ReplicaAutoBalanceDisabled, ReplicaAutoBalanceLeastEffort, ReplicaAutoBalanceBestEffort:
		return nil
```

joshimoo (changes requested):
left some nits and 2 issues.
Will have another look at the scheduling part.

One thing I don't like and it's not your fault is that the healthy replica count function does not consider the different engines.

client/generated_backing_image.go
c3y1huang: I don't know, it got automatically updated with the generator. I assumed it's something we've missed to update, and fine to keep?

shuo-wu: I think Chin-Ya just directly ran the update script and put all `client` modifications in one commit

c3y1huang: Exactly what I did. Should this be removed?

shuo-wu: No. You can rename the commit title to `Update package "client" to the latest`. 
BTW, there is no need to name the commit title as `Implement auto rebalance - xxxxxx`. The commit separation I mentioned before aims to split a big commit into smaller commits based on the functionality/package you modified. It doesn't mean that the title naming schema is fixed.

controller/volume_controller.go
c3y1huang: Thanks, fixed.

api/router.go
c3y1huang: Thanks, addressed.

api/model.go
c3y1huang: Thanks, addressed.

controller/volume_controller.go
c3y1huang: Thanks! fixed.

controller/volume_controller.go
c3y1huang: Thanks! fixed.

types/resource.go
c3y1huang: Thanks! fixed.

types/types.go
c3y1huang: Thanks, fixed.

client/generated_backing_image.go
c3y1huang: Thanks, addressed.

controller/volume_controller.go
c3y1huang: The `ignored` option is only for volume spec, when not specified in the YAML it will default to set to `ignored`, and the operation falls back to the global setting. The `ignored` option should also show as the default option in the UI. (I will create an issue for UI later).

Because we are also validating this `ignored` option against the global setting, so in any case global is set to `ignored`, it will fall back to `disabled.`

api/volume.go
```diff
@@ -299,6 +300,29 @@ func (s *Server) VolumeUpdateReplicaCount(rw http.ResponseWriter, req *http.Requ
 	return s.responseWithVolume(rw, req, "", v)
 }
 
+func (s *Server) VolumeUpdateReplicaAutoBalance(rw http.ResponseWriter, req *http.Request) error {
+	var input UpdateReplicaAutoBalanceInput
+	id := mux.Vars(req)["name"]
+
+	apiContext := api.GetApiContext(req)
+	if err := apiContext.Read(&input); err != nil {
+		return errors.Wrapf(err, "error reading recurringInput")
```
innobead, line 309: Wrong message, it should be something like `error reading replicaAutoBalance` (please help check others to fix as well if there are)

client/generated_backing_image.go
innobead, line 14: Do you change anything relatd to backing image?

controller/volume_controller.go
```diff
@@ -127,6 +130,11 @@ func NewVolumeController(
 
 	vc.scheduler = scheduler.NewReplicaScheduler(ds)
 
+	nodeInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    vc.enqueueVolume,
```
innobead, line 134: This is nodeInformer callback, why you use existing enqueueVolume?

innobead (changes requested):

client/generated_backing_image.go
c3y1huang: No, it got automatically updated from the generator script.

api/volume.go
c3y1huang: Thanks! fixed.

controller/volume_controller.go
c3y1huang: Originally I thought this would trigger enqueueVolume but seems like it's a no-op so I've removed the nodeInformer. However, I don't know what is constantly triggering this yet.

c3y1huang: The informers get resynced at per 30s interval [here](https://github.com/longhorn/longhorn-manager/blob/master/controller/controller_manager.go#L64-L71). After discussing with @innobead we might just leverage this behavior and will not add in nodeInformer since it might not be worth to add in those extra volume queues.

controller/volume_controller.go
```diff
@@ -625,6 +625,16 @@ func (vc *VolumeController) ReconcileEngineReplicaState(v *longhorn.Volume, e *l
 			}
 		}
 
+		setting, err := vc.getAutoBalancedReplicasSetting(v)
+		if err != nil {
+			return err
+		}
+		if setting != types.ReplicaAutoBalanceDisabled {
+			if err := vc.replenishReplicas(v, e, rs, ""); err != nil {
```
innobead, line 633: I assume this `replenishReplicas` will respect the result of the previous `replenishReplicas` call happing when data locality is enabled.

controller/volume_controller.go
```diff
@@ -771,33 +782,88 @@ func (vc *VolumeController) cleanupFailedToScheduledReplicas(v *longhorn.Volume,
...
-					return err
-				}
-				logrus.Debugf("Evicted replica %v", r.Name)
-				return nil
+	var cleaned bool
+	if cleaned, err = vc.cleanupEvictionRequestedReplicas(v, rs); err != nil || cleaned {
+		return err
+	}
+
+	if cleaned, err = vc.cleanupAutoBalancedReplicas(v, rs); err != nil || cleaned {
```
innobead, line 794: Suggest moving cleanupAutoBalancedReplicas after cleanupDataLocalityReplicas to keep `data locality` respected first.

controller/volume_controller.go
```diff
@@ -771,33 +782,88 @@ func (vc *VolumeController) cleanupFailedToScheduledReplicas(v *longhorn.Volume,
...
+		}
+	}
+
+	// Randomly delete extra non-local healthy replicas in the preferred candidate list rNames
+	// Sometime cleanupExtraHealthyReplicas() is called more than once with the same input (v,e,rs).
+	// To make the deleting operation idempotent and prevent deleting more replica than needed,
+	// we always delete the replica with the smallest name.
+	sort.Strings(rNames)
+	rName := rs[rNames[0]]
+	if err := vc.deleteReplica(rName, rs); err != nil {
```
innobead, line 854: Deleting 1 replica is intended. 

We should consider to have more self-explanatory methods instead of implicit behavior inside like this. Not your problem, but we can enhance later.

controller/volume_controller.go
```diff
@@ -1553,6 +1639,269 @@ func getRebuildingReplicaCount(e *longhorn.Engine) int {
...
+	adjustCount := rOverlapTopCount - rOverlapLowCount
+	if adjustCount < 2 {
+		return 0, nil
+	}
+
+	log.Debugf("Found %v replicas to balance: %v", adjustCount, topCountIDRNames)
+	return adjustCount, topCountIDRNames
+}
+
+func contains(list []string, item string) bool {
```
innobead, line 1729: This is general and reusable, so suggest moving to util package.

controller/volume_controller.go
```diff
@@ -1553,6 +1639,269 @@ func getRebuildingReplicaCount(e *longhorn.Engine) int {
 	return rebuilding
 }
 
+type replicaAutoBalanceCount func(*longhorn.Volume, map[string]*longhorn.Replica) (int, map[string][]string, error)
+
+func (vc *VolumeController) replicaAutoBalanceLeastEffort(v *longhorn.Volume,
```
innobead, line 1644: Suggest changing to `getReplicaCountForAutoBalanceLeastEffort`

controller/volume_controller.go
```diff
@@ -1553,6 +1639,269 @@ func getRebuildingReplicaCount(e *longhorn.Engine) int {
...
+	var adjustCount int
+	adjustCount, _, err = fnCount(v, rs)
+	if err != nil {
+		return 0
+	}
+	log.Debugf("Found %v replica candidate for auto-balance", adjustCount)
+	return adjustCount
+}
+
+func (vc *VolumeController) replicaAutoBalanceBestEffort(v *longhorn.Volume,
```
innobead, line 1679: Suggest changing to `getReplicaCountForAutoBalanceBestEffort`

controller/volume_controller.go
```diff
@@ -1553,6 +1639,269 @@ func getRebuildingReplicaCount(e *longhorn.Engine) int {
...
+		return 0, nil
+	}
+
+	if v.Status.Robustness != types.VolumeRobustnessHealthy {
+		log.Debugf("Cannot auto-balance volume in %s state", v.Status.Robustness)
+		return 0, nil
+	}
+
+	var unusedCount int
+	unusedCount, idOverlays, err := fnCount(v, rs)
```
innobead, line 1702: Can we have better naming for `idOverlays`?

controller/volume_controller.go
```diff
@@ -1553,6 +1639,269 @@ func getRebuildingReplicaCount(e *longhorn.Engine) int {
...
+func contains(list []string, item string) bool {
+	for _, i := range list {
+		if i == item {
+			return true
+		}
+	}
+	return false
+}
+
+func (vc *VolumeController) replicaAutoBalanceZoneCount(v *longhorn.Volume, rs map[string]*longhorn.Replica) (int, map[string][]string, error) {
```
innobead, line 1738: Change to `getReplicaCountForAutoBalanceZone`

controller/volume_controller.go
```diff
@@ -1553,6 +1639,269 @@ func getRebuildingReplicaCount(e *longhorn.Engine) int {
...
+		adjustCount = unusedCount
+	} else {
+		adjustCount = unevenCount
+	}
+	log.Debugf("Found %v zone available for auto-balance duplicates in %v", adjustCount, zoneDuplicates)
+
+	return adjustCount, zoneDuplicates, err
+}
+
+func (vc *VolumeController) replicaAutoBalanceNodeCount(v *longhorn.Volume, rs map[string]*longhorn.Replica) (int, map[string][]string, error) {
```
innobead, line 1826: Change to `getReplicaCountForAutoBalanceNode`

controller/volume_controller.go
```diff
@@ -1566,11 +1915,26 @@ func (vc *VolumeController) getReplenishReplicasCount(v *longhorn.Volume, rs map
 			usableCount++
 		}
 	}
-
-	if v.Spec.NumberOfReplicas < usableCount {
+	rCount := v.Spec.NumberOfReplicas
```
innobead, line 1918: Just have 3 conditions, suggest use `v.Spec.NumberOfReplicas` directly instead of creating another variable.

innobead (changes requested):
LGTM, but some minor requests to change.

innobead:
Because this is related to replica management, @shuo-wu or @joshimoo should review and approve before merging.

controller/volume_controller.go
c3y1huang: replica auto-balance count only happens when all replicas are deployed(`rCount == usableCount`), this shouldn't conflict with when data locality is enabled which happens when replicas are still replenishing to meet the `v.Spec.NumberOfReplicas`.

types/types.go
```diff
@@ -432,6 +432,18 @@ func ValidateReplicaCount(count int) error {
...
 
+func ValidateReplicaAutoBalance(option ReplicaAutoBalance) error {
+	switch option {
+	case ReplicaAutoBalanceIgnored,
+		ReplicaAutoBalanceDisabled,
+		ReplicaAutoBalanceLeastEffort,
+		ReplicaAutoBalanceBestEffort:
+		return nil
+	default:
+		return fmt.Errorf("invalid replica rebalance option: %v", option)
```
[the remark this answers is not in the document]
c3y1huang: Thanks, fixed.

controller/volume_controller.go
c3y1huang: Needs to keep cleanupAutoBalancedReplicas  before to ensure auto-balance best-effort replica list is respected. I've updated to have the replica list go through getPreferredReplicaCandidatesForDeletion so `data locality` will also get respected.

controller/volume_controller.go
c3y1huang: Thank you, addressed.

controller/volume_controller.go
```diff
@@ -1553,6 +1655,260 @@ func getRebuildingReplicaCount(e *longhorn.Engine) int {
...
+		rNameCount := len(rNames)
+		if len(rNames) > len(topCountIDRNames) {
+			topCountIDRNames = rNames
+			rOverlapTopCount = rNameCount
+		}
+		if rOverlapLowCount == 0 || rNameCount < rOverlapLowCount {
+			rOverlapLowCount = rNameCount
+		}
+	}
+	adjustCount := rOverlapTopCount - rOverlapLowCount
```
shuo-wu, line 1736: Should not the adjustCount equal to `rOverlapTopCount - rOverlapLowCount - 1`?

controller/volume_controller.go
```diff
@@ -1553,6 +1655,260 @@ func getRebuildingReplicaCount(e *longhorn.Engine) int {
...
+			log.Debugf("Replica %v on node %v not count for auto-balance, could get evicted", r.Spec.NodeID, r.Name)
+			continue
+		}
+
+		nZone := node.Status.Zone
+		_, exist = zoneDuplicates[nZone]
+		if exist {
+			zoneDuplicates[nZone] = append(zoneDuplicates[nZone], r.Name)
+		} else {
+			zoneDuplicates[nZone] = nil
```
shuo-wu, line 1772: Do you mean `zoneDuplicates[nZone] = []string{}`? I think the list records extra replicas only.

Similar below

controller/volume_controller.go
```diff
@@ -1553,6 +1655,260 @@ func getRebuildingReplicaCount(e *longhorn.Engine) int {
...
+
+	ei, err := vc.getEngineImage(v.Status.CurrentImage)
+	if err != nil {
+		return 0, nil, err
+	}
+
+	unusedZone := make(map[string][]string)
+	for nodeName, node := range readyNodes {
+		if util.Contains(usedZones, node.Status.Zone) {
+			log.Debugf("Cannot use node %v in zone '%v', have running replica", nodeName, node.Status.Zone)
```
shuo-wu, line 1794: The debug logs here and below are unnecessary.

shuo-wu (changes requested):
Question 1: Can the scheduler **evenly** distribute all replicas among zones/nodes? I didn't see any modifications for this part.
e.g., When there are 2 nodes, nodeA contains 3 replicas while nodeB contains 1 replica. The adjust count should be 1. Then how does the scheduler know the new replica should be in nodeB?

Question 2: Will failed replica re-usage block/intervene this feature?

controller/volume_controller.go
c3y1huang: Fixed. tested with [test_replica_auto_balance_zone_best_effort_with_data_locality](https://github.com/longhorn/longhorn-tests/pull/609).

controller/volume_controller.go
c3y1huang: Addressed.

controller/volume_controller.go
c3y1huang: Addressed.

controller/volume_controller.go
c3y1huang: Updated to`duplicatedReplicaNames`.

controller/volume_controller.go
c3y1huang: Addressed.

controller/volume_controller.go
c3y1huang: Addressed.

controller/volume_controller.go
c3y1huang: Addressed.

controller/volume_controller.go
```diff
@@ -808,12 +876,40 @@ func (vc *VolumeController) cleanupExtraHealthyReplicas(v *longhorn.Volume, e *l
...
+		}
+
+		setting = types.ReplicaAutoBalance(globalSetting)
+	}
+
+	err = types.ValidateReplicaAutoBalance(types.ReplicaAutoBalance(setting))
+	if err != nil {
+		setting = types.ReplicaAutoBalanceDisabled
+	}
+	return setting, errors.Wrapf(err, "replica auto-balance is disabled")
```
c3y1huang, line 912: > If volumeSetting or globalSetting is not the defined option, should this function fall back to return types.ReplicaAutoBalanceDisabled (with a warning) rather than blindly convert the string to type types.ReplicaAutoBalance?

Addressed, thanks! (I don't know why I can't reply to the original comment)

controller/volume_controller.go
```diff
@@ -1553,6 +1655,260 @@ func getRebuildingReplicaCount(e *longhorn.Engine) int {
...
+	defer func() {
+		if err != nil {
+			log.WithError(err).Warn("Skip replica auto-balance")
+		}
+	}()
+
+	setting, err := vc.getAutoBalancedReplicasSetting(v)
+	enabled := []string{
+		string(types.ReplicaAutoBalanceLeastEffort),
+		string(types.ReplicaAutoBalanceBestEffort),
```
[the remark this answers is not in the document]
c3y1huang: Addressed.

controller/volume_controller.go
```diff
@@ -1553,6 +1655,260 @@ func getRebuildingReplicaCount(e *longhorn.Engine) int {
...
+
+	var err error
+	defer func() {
+		if err != nil {
+			log.WithError(err).Warn("Skip replica auto-balance")
+		}
+	}()
+
+	setting, err := vc.getAutoBalancedReplicasSetting(v)
+	if setting != types.ReplicaAutoBalanceBestEffort || err != nil {
```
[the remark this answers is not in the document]
c3y1huang: Thanks, fixed.

controller/volume_controller.go
```diff
@@ -1553,6 +1655,260 @@ func getRebuildingReplicaCount(e *longhorn.Engine) int {
...
+
+	var unusedCount int
+	unusedCount, idOverlays, err := fnCount(v, rs)
+	if unusedCount == 0 && len(idOverlays) <= 1 {
+		return 0, nil
+	}
+
+	var topCountIDRNames []string
+	var rOverlapTopCount int
+	var rOverlapLowCount int
```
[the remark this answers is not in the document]
c3y1huang: Updated. Any other suggestion for naming is welcome.

controller/volume_controller.go
c3y1huang: Not sure why include `- 1`? Do you have a specific case where adjustCount is not expected without the `- 1`?

controller/volume_controller.go
c3y1huang: Addressed. and yes the list records only extra replicas so we preserve at least one.

controller/volume_controller.go
c3y1huang: Addressed.

c3y1huang:
> Question 1: Can the scheduler **evenly** distribute all replicas among zones/nodes? I didn't see any modifications for this part.
> e.g., When there are 2 nodes, nodeA contains 3 replicas while nodeB contains 1 replica. The adjust count should be 1. Then how does the scheduler know the new replica should be in nodeB?
> 

Ok, I didn't consider the case where there are more than 3 replicas per volume for HA. At the moment this PR did not deal with the scheduler so if the replica scheduler happens to schedule the adjusting replica to the original node then the replica auto-balance will delete it again and the process would repeat until the replica is adjusted evenly.

> Question 2: Will failed replica re-usage block/intervene this feature?

ATM I did not find where replicas re-usage could block/intervene this feature.

c3y1huang:
> > Question 1: Can the scheduler **evenly** distribute all replicas among zones/nodes? I didn't see any modifications for this part.
> > e.g., When there are 2 nodes, nodeA contains 3 replicas while nodeB contains 1 replica. The adjust count should be 1. Then how does the scheduler know the new replica should be in nodeB?
> 
> Ok, I didn't consider more than 3 replicas per volume for HA as a common use case. At the moment this PR did not deal with the scheduler so if the replica scheduler happens to schedule the adjusting replica to the original node then the replica auto-balance will delete it again and the process would repeat until the replica is adjusted evenly.

Made the fix by set hardNodeAffinity to nodeID that has the least replicas.

Also found a dead loop between zone and node during adjusting counts with best-effort, So the commit also dropped the best-effort replica adjustment during replenishing, but kept for cleanups so it's prioritized to be flattened out at node level.

c3y1huang:
> Made the fix by set in-function hardNodeAffinity to nodeID that has the least replicas.
> 
> Also found a dead loop between zone and node during adjusting counts with best-effort when replica numbers are high. So the commit also dropped the best-effort replica adjustment during replenishing, but kept in cleanup so it's prioritized to be flattened out at node level.

Updated with another fix to even out replicas if there are unused nodes in zones. Case [test_replica_auto_balance_zone_best_effort_with_uneven_node_in_zones](https://github.com/longhorn/longhorn-tests/pull/609) added.

controller/volume_controller.go
c3y1huang: Addressed, thanks!

controller/volume_controller.go
```diff
@@ -1553,7 +1652,286 @@ func getRebuildingReplicaCount(e *longhorn.Engine) int {
...
+		}
+	}
+	for _, r := range rs {
+		if r.Status.CurrentState != types.InstanceStateRunning {
+			continue
+		}
+
+		node, exist := readyNodes[r.Spec.NodeID]
+		if !exist {
+			log.Debugf("Replica %v on node %v not count for auto-balance, could get evicted", r.Spec.NodeID, r.Name)
```
shuo-wu, line 1780: The unnecessary debug logs here and below.

controller/volume_controller.go
shuo-wu: Let's take a look at this case again: Assuming there are 2 nodes, nodeA contains 3 replicas while nodeB contains 1 replica. Hence `leastExtraRCount` is 0 while `mostExtraRCount` is 2 if the current logic is applied, then `adjustCount` will be 2. But actually, we only need to adjust 1 replica.

shuo-wu (changes requested):
I am thinking if function `getReplicaCountForAutoBalanceBestEffort` should return a **list** of nodes that have the least replicas. And what if all affinity nodes cannot fit a new replica? e.g., there are only 2 nodes schedulable. nodeA contains 3 replicas while nodeB contains 1 replica. But the schedulable size of nodeB cannot fit one more replica.

controller/volume_controller.go
c3y1huang: Addressed.

controller/volume_controller.go
c3y1huang: You are right, the naming isn't consistent. It's more of a  diff count. Fixed!

c3y1huang:
> I am thinking if function `getReplicaCountForAutoBalanceBestEffort` should return a **list** of nodes that have the least replicas. 

Addressed.

> And what if all affinity nodes cannot fit a new replica? e.g., there are only 2 nodes schedulable. nodeA contains 3 replicas while nodeB contains 1 replica. But the schedulable size of nodeB cannot fit one more replica.

The behavior is the same as data locality, when the replica failed to schedule be cleaned up in `cleanupFailedToScheduledReplicas` and process balance -> delete will loop until it's balanced.

controller/volume_controller.go
```diff
@@ -1567,10 +1947,104 @@ func (vc *VolumeController) getReplenishReplicasCount(v *longhorn.Volume, rs map
...
+		adjustNodeAffinity := ""
+		adjustCount := vc.getReplicaCountForAutoBalanceLeastEffort(v, e, rs, vc.getReplicaCountForAutoBalanceZone)
+		if adjustCount == 0 {
+			adjustCount = vc.getReplicaCountForAutoBalanceLeastEffort(v, e, rs, vc.getReplicaCountForAutoBalanceNode)
+		}
+		var nCandidates []string
+		if adjustCount == 0 {
+			adjustCount, _, nCandidates = vc.getReplicaCountForAutoBalanceBestEffort(v, e, rs, vc.getReplicaCountForAutoBalanceNode)
+		}
+		if adjustCount == 0 && adjustNodeAffinity == "" {
```
shuo-wu, line 1965: `if adjustCount == 0` is enough

controller/volume_controller.go
```diff
@@ -1567,10 +1947,104 @@ func (vc *VolumeController) getReplenishReplicasCount(v *longhorn.Volume, rs map
...
+		return 0, ""
+	case v.Spec.NumberOfReplicas > usableCount:
+		return v.Spec.NumberOfReplicas - usableCount, ""
+	case v.Spec.NumberOfReplicas == usableCount:
+		adjustNodeAffinity := ""
+		adjustCount := vc.getReplicaCountForAutoBalanceLeastEffort(v, e, rs, vc.getReplicaCountForAutoBalanceZone)
+		if adjustCount == 0 {
+			adjustCount = vc.getReplicaCountForAutoBalanceLeastEffort(v, e, rs, vc.getReplicaCountForAutoBalanceNode)
+		}
+		var nCandidates []string
```
shuo-wu, line 1961: If the adjust count is not zero, there is no need to continue. Hence we can directly return. e.g.,:
```suggest
if adjustCount := vc.getReplicaCountForAutoBalanceLeastEffort(v, e, rs, vc.getReplicaCountForAutoBalanceZone); adjustCount != 0 {
    return adjustCount, ""
}
if adjustCount := vc.getReplicaCountForAutoBalanceLeastEffort(v, e, rs, vc.getReplicaCountForAutoBalanceNode); adjustCount != 0 {
    return adjustCount, ""
}
adjustNodeAffinity := ""
```

shuo-wu (changes requested):
I mean if `getReplenishReplicasCount` can return the candidate list then ask vc.scheduler.RequireNewReplica()` to receive the list, rather than letting `getReplenishReplicasCount` blindly picking up the 1st candidate from the list. But this involves modifying/refactor the scheduler logic. Can you check if it can make the schedule have a larger possibility to rebuild a new replica on the desired nodes. If Yes, you can add a TODO comment here and below, or directly implement it with a new commit

controller/volume_controller.go
c3y1huang: Thanks. fixed.

controller/volume_controller.go
c3y1huang: Thanks. Fixed.

c3y1huang:
> I mean if `getReplenishReplicasCount` can return the candidate list then ask vc.scheduler.RequireNewReplica()`to receive the list, rather than letting`getReplenishReplicasCount` blindly picking up the 1st candidate from the list. But this involves modifying/refactor the scheduler logic. Can you check if it can make the schedule have a larger possibility to rebuild a new replica on the desired nodes. If Yes, you can add a TODO comment here and below, or directly implement it with a new commit

IMO this is not critical and could be enhanced later. Added a TODO comment, and https://github.com/longhorn/longhorn/issues/2667.

shuo-wu (commented):
Notice that `getNodeCandidatesForAutoBalanceZone` doesn't check the schedulable size and the disks on the candidate nodes for the volumes. Maybe it's better to check it for now. Once #2667 is implemented, we can remove this function and directly ask the scheduler to take care of it.

c3y1huang:
> Notice that `getNodeCandidatesForAutoBalanceZone` doesn't check the schedulable size and the disks on the candidate nodes for the volumes. Maybe it's better to check it for now. Once #2667 is implemented, we can remove this function and directly ask the scheduler to take care of it.

Thanks, addressed.

shuo-wu (approved):

innobead (approved):

PR #905, opened 2021-05-31
Fix full data over web sockets when there is no update

c3y1huang:
- [List order caused full data transfer when no resource updated.](https://github.com/longhorn/longhorn/issues/2646)

- [Full data over WS at per 30/seconds interval seems useless.](https://github.com/longhorn/longhorn/issues/2591)

### DIFF
--- a/api/model.go
+++ b/api/model.go
@@ -812,59 +812,63 @@ func toVolumeResource(v *longhorn.Volume, ves []*longhorn.Engine, vrs []*longhor
 		}
 		backupStatus := e.Status.BackupStatus
 		if backupStatus != nil {
-			for id, status := range backupStatus {
+			ids := util.GetSortedKeysFromMap(backupStatus)
+			for _, id := range ids {
 				backups = append(backups, BackupStatus{
 					Resource:  client.Resource{},
 					Name:      id,
-					Snapshot:  status.SnapshotName,
-					Progress:  status.Progress,
-					BackupURL: status.BackupURL,
-					Error:     status.Error,
-					State:     status.State,
-					Replica:   datastore.ReplicaAddressToReplicaName(status.ReplicaAddress, vrs),
+					Snapshot:  backupStatus[id].SnapshotName,
+					Progress:  backupStatus[id].Progress,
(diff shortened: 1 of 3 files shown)


c3y1huang:
Screenshot-2021-05-31-18:24:18

api/model.go
```diff
@@ -812,59 +813,79 @@ func toVolumeResource(v *longhorn.Volume, ves []*longhorn.Engine, vrs []*longhor
 		}
 		backupStatus := e.Status.BackupStatus
 		if backupStatus != nil {
-			for id, status := range backupStatus {
+			ids := make([]string, 0, len(backupStatus))
```
innobead, line 816: Suggest to extract the sort part as a common function in util  to reuse, ex: `GetSortedKeysFromMap`.

innobead (changes requested):
LGTM, just a comment required to resolve.

api/model.go
c3y1huang: Thank you, fixed!

shuo-wu (approved):

innobead (approved):

innobead:
@c3y1huang please backport to 1.1.2. Thanks.

api/model.go
```diff
@@ -812,59 +812,63 @@ func toVolumeResource(v *longhorn.Volume, ves []*longhorn.Engine, vrs []*longhor
...
-					State:     status.State,
+					Error:     purgeStatus[replica].Error,
+					IsPurging: purgeStatus[replica].IsPurging,
+					Progress:  purgeStatus[replica].Progress,
+					State:     purgeStatus[replica].State,
 				})
 			}
 		}
 		rebuildStatus := e.Status.RebuildStatus
 		if rebuildStatus != nil {
```
joshimoo, line 861: nit: these nil checks can be removed, since a nil/empty map has 0 elements. And therefore there will be 0 iterations. Same for the other checks.

joshimoo (commented):
LGTM, Left some comments.
One could also consider adding `flattenBackupStatus(map) -> slice` functions next to the object definitions.

PR #908, opened 2021-06-03
Add functionality to "clean up" a faulty mount if `NodePublishVolume` is called

Oats87:
https://github.com/longhorn/longhorn/issues/2629


innobead (approved):
LGTM

After merging, we need to backport to 1.1.2. cc @joshimoo

joshimoo:
@Oats87 thanks for the PR, I incorporated your changes in https://github.com/longhorn/longhorn-manager/pull/910

PR #910, opened 2021-06-07
Ensure cleanup of corrupted mount points

joshimoo:
Ensure that corrupt mount points are cleaned up during NodePublishVolume calls, this currently returns an error afterwards to make sure that kubelet can cleanup on it's side. In the future we might want to consider utilizing the NodeStateVolume calls to actually deal with the block device then only do bind mounts during NodePublishVolume calls.

We should also update the node plugin (lock per volume) to keek track of currently running operations for a single volume. To prevent multiple operations from running concurrently, the correct error to return then is aborted. Since that will signal that a previous operation is in progress.

longhorn/longhorn#2629
supersedes longhorn/longhorn-manager#908

### DIFF
--- a/csi/node_server.go
+++ b/csi/node_server.go
@@ -204,8 +204,26 @@ func (ns *NodeServer) nodePublishMountVolume(volumeName, devicePath, targetPath,
 	}
 	if !notMnt {
 		if _, err := ioutil.ReadDir(targetPath); err != nil {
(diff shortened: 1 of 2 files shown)


shuo-wu (approved):
LGTM

joshimoo:
hold merge for @innobead requested review.

innobead (approved):
LGTM

innobead:
@joshimoo Please help backport to 1.1.2.

PR #911, opened 2021-06-07
Remove VA cleanup feature

joshimoo:
This removes the VA cleanup feature, since the use case is covered by the pod deletion feature.
It's better to only have a single way of accomplishing this.

I didn't touch the settings as part of this PR, since it's not important and they are just ignored.
So I can easily backport to 1.1.2

longhorn/longhorn#2060

### DIFF
--- a/controller/controller_manager.go
+++ b/controller/controller_manager.go
@@ -90,7 +90,6 @@ func StartControllers(logger logrus.FieldLogger, stopCh chan struct{}, controlle
 	cronJobInformer := kubeInformerFactory.Batch().V1beta1().CronJobs()
 	daemonSetInformer := kubeInformerFactory.Apps().V1().DaemonSets()
 	deploymentInformer := kubeInformerFactory.Apps().V1().Deployments()
-	volumeAttachmentInformer := kubeInformerFactory.Storage().V1beta1().VolumeAttachments()
 	priorityClassInformer := kubeInformerFactory.Scheduling().V1().PriorityClasses()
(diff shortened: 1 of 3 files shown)


joshimoo:
1/3 of the solution items for https://github.com/longhorn/longhorn/issues/2629

joshimoo:
Hold merge for @innobead requested review.

innobead (changes requested):
Right now, it works as ignoring the VA deletion setting, but it seems we still need to remove/clean up related configurations like `SettingDefinitionVolumeAttachmentRecoveryPolicy` from settings, manifests and charts?

joshimoo:
@innobead I didn't want to touch the setting for 1.1.2 and I currently don't have time to look into the complexity involved of removing it.

Since we should only do it for 1.2 I created this issue to track the setting removal https://github.com/longhorn/longhorn/issues/2661

After this pr is merged the setting has now effect.

innobead:
> @innobead I didn't want to touch the setting for 1.1.2 and I currently don't have time to look into the complexity involved of removing it.
> 
> Since we should only do it for 1.2 I created this issue to track the setting removal [longhorn/longhorn#2661](https://github.com/longhorn/longhorn/issues/2661)
> 
> After this pr is merged the setting has now effect.

Sounds good to me. Let's do this way first. (but we need to update the doc at least)

innobead (approved):

innobead:
@joshimoo Please help backport to 1.1.2.

PR #912, opened 2021-06-08
Roundup requested volume size at the beginning of the CSI CreateVolume() call

PhanLe1010:
By rounding up the requested volume size before checking the size of
existing volume, we ensure that the CSI volume creation operation is
retried with idempotency in the case where the requested size is not
multiple of 2*1024*1024

longhorn/longhorn#2624


joshimoo (approved):
LGTM

innobead (approved):

innobead:
@PhanLe1010 Please backport to 1.1.2. Thanks.

PR #915, opened 2021-06-09
Allow users to set startingDeadlineSeconds for CronJobs

PhanLe1010:
The startingDeadlineSeconds value helps to prevent Kubernetes
from permanently stopping the CronJob if it hasn't been running
for a long time

The startingDeadlineSeconds setting contains 2 levels:

1. The default value for all cronjobs that user can change in
setting page in Longhorn UI
2. The setting value per cronjob. This is similar to the retain
count. If the user specifies this setting, it will overwrite
the default setting.

longhorn/longhorn#2513

### DIFF
--- a/client/generated_recurring_job.go
+++ b/client/generated_recurring_job.go
@@ -15,6 +15,8 @@ type RecurringJob struct {
 
 	Retain int64 `json:"retain,omitempty" yaml:"retain,omitempty"`
 
+	StartingDeadlineSeconds int64 `json:"startingDeadlineSeconds,omitempty" yaml:"starting_deadline_seconds,omitempty"`
+
 	Task string `json:"task,omitempty" yaml:"task,omitempty"`
 }
 
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -2147,6 +2147,13 @@ func (vc *VolumeController) createCronJob(v *longhorn.Volume, job *types.Recurri
 	if err != nil {
 		return nil, err
 	}
+	startingDeadlineSeconds, err := vc.ds.GetSettingAsInt(types.SettingNameDefaultCronJobStartingDeadlineSeconds)
+	if err != nil {
+		return nil, err
+	}
+	if job.StartingDeadlineSeconds > 0 {
+		startingDeadlineSeconds = job.StartingDeadlineSeconds
+	}
 	// for mounting inside container
 	privilege := true
 	cronJob := &batchv1beta1.CronJob{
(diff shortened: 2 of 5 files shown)


PhanLe1010:
I have not run `e2e core testes` for this PR yet. Please hold on merging

joshimoo:
@PhanLe1010 did you miss this comment? https://github.com/longhorn/longhorn/issues/2513#issuecomment-847666125
What's the benefit for the user to specify starting Deadline Seconds?
Any issue with just deleting the no longer functioning cronjobs and have longhorn recreate them?

PhanLe1010:
I feel deleting the no longer functioning cronjob is kind of a shortcut/hack:
* If we go this way we would have to repeat the Kubernetes CronJon controller logic to determine which CronJob miss more than 100 runs since its last scheduled time.
* If there are many volumes that are detached for a long time and they suddenly attached at the same time, many recurring backup jobs may start at the same time (because there is no deadline specified) and it hurts the system performance. 

I prefer the startingDeadlineSeconds because it gives users more control over their CronJobs (this is not really a strong point, just feel more comfortable)

joshimoo:
Hmm I don't know if any user would know what startingDeadline seconds is even used for.
I certainly wouldn't know without studying the kubernetes documentation.
What I am trying to say it's not intuitive so might be hard to easily explain to the user what they need to set there?
Are we able to tell when a cron job is no longer capable of running, if so how?

PhanLe1010:
After discussion with @yasker , I realized that:
* I should have replied/clarified what `deleting the no longer functioning cronjob` means in the comment https://github.com/longhorn/longhorn/issues/2513#issuecomment-847666125
* I was thinking that it means we have to find out which cronjob is no longer capable of running (by counting its number of miss runs) and only delete those jobs. In this case, we would repeat Kubernetes CronJob controller logic. 
* However, Sheng pointed out that it can just mean that we delete the CronJob when the volume is detached. With this approach, we don't have the above problem. Moreover, this approach also solve an additional problem (which is the second point I mentioned in https://github.com/longhorn/longhorn-manager/pull/915#issuecomment-857888950): 
   * Currently we only suspend the CronJob when volumes are detached. Assume that there are 100 volumes are detached at the same time, we suspend 100 CronJobs. Further assume that every CronJob have 1 hour interval. After 2.5 hour since the time we detach 100 volumes, users attach 100 volume at nearly the same time. In this case, Kubernetes will start 100 CronJobs at the same time since they haven't run for a while and there is no `startingDeadlineSeconds`. I have tested this behavior by create a recurring backup job that runs every 1 min; detach the volume so Longhorn suspends the job for 2.5 mins; attach the volume; Longhorn activate the job; Kubernetes immediately start the job.

Now that I understand `deleting the no longer functioning cronjob` comment (which I should have replied earlier), [my argument](https://github.com/longhorn/longhorn-manager/pull/915#issuecomment-857888950) is no longer valid. I have create a new PR with the `deleting the no longer functioning cronjob` approach at https://github.com/longhorn/longhorn-manager/pull/916

Thanks @yasker and @joshimoo

PR #922, opened 2021-06-14
CSI driver fixes (path, locks, etc)

joshimoo:
The path cleanup might need to be modified to take care of the old invalid socket locations on existing systems.
But future systems should be working correctly now.

- Cleanup CSI Plugin paths
- Ensure only a single in flight CSI operation for a resource
- Validate csi requests on entry and improve logging

### DIFF
--- a/csi/controller_server.go
+++ b/csi/controller_server.go
@@ -54,14 +54,10 @@ func NewControllerServer(apiClient *longhornclient.RancherClient, nodeID string)
 }
 
 func (cs *ControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVolumeRequest) (*csi.CreateVolumeResponse, error) {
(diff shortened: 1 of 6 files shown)


shuo-wu:
The 1st & 2nd commits are merged in https://github.com/longhorn/longhorn-manager/pull/927.
For the 3rd one, can you provide more contexts & explanations besides statement `kubernetes cannot assumed that they are operating in the same process.`? I still don't know why we need a lock for it.

shuo-wu:
@joshimoo It seems that we forget this PR. I think we can merge it after removing the first 2 commits.

joshimoo:
Thanks @innobead I will put the csi volume locks on the back burner for the moment. They are a good addition but our driver might run into issues with nfs mounts once we enable hard mode. I plan on dealing with it at that point.

PR #927, opened 2021-06-15
kubelet csi corrupt mnt mitigations

joshimoo:
This is for v1.1.2 

- does not contain csi locking.
- contains the plugin path fixes, this might leave the obsolete plugin around, not sure in regard to updates.

I try unmounting a mount point when volume is not attached is encountered.
Since it means that any prior mount point would be useless at the moment.

Since we use soft mount, the nfs calls will return withhin 15s
I decreased the timeout of the external sidecars to 1m50s
and the final volume state polling of the csi driver to 90s, this gives the rest of the function 20s

### DIFF
--- a/csi/controller_server.go
+++ b/csi/controller_server.go
@@ -21,7 +21,8 @@ import (
 )
 
 const (
-	timeoutAttachDetach     = 120 * time.Second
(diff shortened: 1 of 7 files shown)


innobead (approved):
LGTM

PR #928, opened 2021-06-16
Increase ReadinessProbe TimeoutSeconds to 5 secs

jenting:
#### Proposal Change

- Increase the Pods' ReadinessProbe.TimeoutSeconds from 1 sec to `PodProbePeriodSeconds-1` secs (4 secs).
  - Backing image manager Pods
  - Engine image Pods
  - Share manager Pods

- Change instance manager Pod LivenessProbe to TCP probe to decrease the overhead than gRPC probe.

#### Related Issue

https://github.com/longhorn/longhorn/issues/2590#issuecomment-856389880

### DIFF
--- a/controller/engine_image_controller.go
+++ b/controller/engine_image_controller.go
@@ -762,8 +762,8 @@ func (ic *EngineImageController) createEngineImageDaemonSetSpec(ei *longhorn.Eng
 									},
 								},
 								InitialDelaySeconds: 5,
-								TimeoutSeconds:      5,
-								PeriodSeconds:       5,
+								TimeoutSeconds:      datastore.PodProbeTimeoutSeconds,
+								PeriodSeconds:       datastore.PodProbePeriodSeconds,
 							},
 							SecurityContext: &v1.SecurityContext{
 								Privileged: &privileged,
--- a/controller/engine_image_controller.go
+++ b/controller/engine_image_controller.go
@@ -762,6 +762,7 @@ func (ic *EngineImageController) createEngineImageDaemonSetSpec(ei *longhorn.Eng
 									},
 								},
 								InitialDelaySeconds: 5,
+								TimeoutSeconds:      5,
 								PeriodSeconds:       5,
 							},
 							SecurityContext: &v1.SecurityContext{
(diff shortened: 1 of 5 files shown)


controller/engine_image_controller.go
```diff
@@ -762,6 +762,7 @@ func (ic *EngineImageController) createEngineImageDaemonSetSpec(ei *longhorn.Eng
 									},
 								},
 								InitialDelaySeconds: 5,
+								TimeoutSeconds:      5,
```
joshimoo, line 765: consider using the ds consts here as well.

datastore/kubernetes.go
```diff
@@ -31,6 +31,7 @@ const (
 	KubeStatusPollInterval = 1 * time.Second
 
 	PodProbeInitialDelay             = 3
+	PodProbeTimeoutSeconds           = 5
```
joshimoo, line 34: Good idea externalizing this
- make the timeout 1 smaller then the probePeriod, just in case :)

controller/engine_image_controller.go
innobead, line 765: Use datastore.PodProbeTimeoutSeconds instead?

joshimoo (commented):
There is also the instance-manager deployments, but @shuo-wu mentioned we could use a TCP probe. 

https://github.com/longhorn/longhorn-manager/blob/779149f6f3a8556abff08e12feec6ec0646c3831/controller/instance_manager_controller.go#L842

Need to consider if the modification will lead to a restart of the IM deployments, only want to do this for new IM deployments probably.

controller/engine_image_controller.go
```diff
@@ -762,7 +762,8 @@ func (ic *EngineImageController) createEngineImageDaemonSetSpec(ei *longhorn.Eng
 									},
 								},
 								InitialDelaySeconds: 5,
```
jenting, line 764: I did not change this to use `datastore.PodProbeInitialDelay` since the value is 5 bigger than 3.

innobead (approved):

innobead:
@jenting Please help backport to 1.1.2.

PR #929, opened 2021-06-16
Add functions to get pv pvc from kubernetes client for volume expansion

cclhsu:
#### Proposed Changes ####

Add functions to get pv pvc from kubernetes client for volume expansion to resolve issue '[BUG] Application couldn't mount the volume after expanded the volume'

- Once the pvc's status become bound, tried the expansions. Only the alternate operation of expansion works on the detached volume. The sequence is 1st - fail, 2nd- pass, 3rd - fail, 4th- pass etc.

#### Types of Changes ####

Bug Fix

#### Verification ####

Use `longhorn-static` should always return error message 

```
unable to expand volume test: persistentvolumeclaims "test" is forbidden: only dynamically provisioned pvc can be resized and the storageclass that provisions the pvc must support resize
```

#### Linked Issues ####

Refs: https://github.com/longhorn/longhorn/issues/2422

#### Further Comments ####

None

### DIFF
--- a/datastore/kubernetes.go
+++ b/datastore/kubernetes.go
@@ -408,10 +408,22 @@ func (s *DataStore) UpdatePersistentVolume(pv *corev1.PersistentVolume) (*corev1
 	return s.kubeClient.CoreV1().PersistentVolumes().Update(pv)
 }
 
-// GetPersistentVolume gets the PersistentVolume from the index for the
+// GetPersistentVolumeRO gets the PersistentVolume from the index for the
 // given name
-func (s *DataStore) GetPersistentVolume(pvName string) (*corev1.PersistentVolume, error) {
+// This function returns direct reference to the internal cache object and should not be mutated.
+// Consider using this function when you can guarantee read only access and don't want the overhead of deep copies
+func (s *DataStore) GetPersistentVolumeRO(pvName string) (*corev1.PersistentVolume, error) {
 	return s.pvLister.Get(pvName)
+}
+
+// GetPersistentVolume gets a mutable PersistentVolume for the given name
+func (s *DataStore) GetPersistentVolume(pvName string) (*corev1.PersistentVolume, error) {
+	resultRO, err := s.pvLister.Get(pvName)
+	if err != nil {
+		return nil, err
+	}
+	// Cannot use cached object from lister
+	return resultRO.DeepCopy(), nil
 }
 
 // CreatePersistentVolumeClaim creates a PersistentVolumeClaim resource
@@ -432,10 +444,22 @@ func (s *DataStore) UpdatePersistentVolumeClaim(namespace string, pvc *corev1.Pe
 	return s.kubeClient.CoreV1().PersistentVolumeClaims(namespace).Update(pvc)
 }
 
-// GetPersistentVolumeClaim gets the PersistentVolumeClaim from the
+// GetPersistentVolumeClaimRO gets the PersistentVolumeClaim from the
 // index for the given name and namespace
-func (s *DataStore) GetPersistentVolumeClaim(namespace, pvcName string) (*corev1.PersistentVolumeClaim, error) {
+// This function returns direct reference to the internal cache object and should not be mutated.
+// Consider using this function when you can guarantee read only access and don't want the overhead of deep copies
+func (s *DataStore) GetPersistentVolumeClaimRO(namespace, pvcName string) (*corev1.PersistentVolumeClaim, error) {
 	return s.pvcLister.PersistentVolumeClaims(namespace).Get(pvcName)
+}
+
+// GetPersistentVolumeClaim gets a mutable PersistentVolumeClaim for the given name and namespace
+func (s *DataStore) GetPersistentVolumeClaim(namespace, pvcName string) (*corev1.PersistentVolumeClaim, error) {
+	resultRO, err := s.pvcLister.PersistentVolumeClaims(namespace).Get(pvcName)
+	if err != nil {
+		return nil, err
+	}
+	// Cannot use cached object from lister
+	return resultRO.DeepCopy(), nil
 }
 
 // GetConfigMapRO gets ConfigMap with the given name in s.namespace


manager/volume.go
```diff
@@ -633,7 +633,7 @@ func (m *VolumeManager) Expand(volumeName string, size int64) (v *longhorn.Volum
 
 	kubernetesStatus := &v.Status.KubernetesStatus
 	if kubernetesStatus.PVCName != "" && kubernetesStatus.LastPVCRefAt == "" {
-		pvc, err := m.ds.GetPersistentVolumeClaim(kubernetesStatus.Namespace, kubernetesStatus.PVCName)
+		pvc, err := m.ds.GetPersistentVolumeClaimRO(kubernetesStatus.Namespace, kubernetesStatus.PVCName)
```
joshimoo, line 636: Since you are modifying the PVC you want to use `GetPersistentVolumeClaim` then return a deep copy in that non ReadOnly function call.

While you are fixing this also return a deep copy from the `GetPersistentVolume` which is used by the PVC creation / cleanup, which again is modifying the raw instance.

joshimoo (changes requested):
- need to add deep copy for the mutable call 
- need to use the mutable call since you are modifying the state of the object.

manager/volume.go

datastore/kubernetes.go
```diff
@@ -431,12 +441,22 @@ func (s *DataStore) UpdatePersistentVolumeClaim(namespace string, pvc *corev1.Pe
...
+// GetPersistentVolumeClaimRO gets the PersistentVolumeClaim from the
 // index for the given name and namespace
-func (s *DataStore) GetPersistentVolumeClaim(namespace, pvcName string) (*corev1.PersistentVolumeClaim, error) {
+func (s *DataStore) GetPersistentVolumeClaimRO(namespace, pvcName string) (*corev1.PersistentVolumeClaim, error) {
 	return s.pvcLister.PersistentVolumeClaims(namespace).Get(pvcName)
 }
 
+// GetPersistentVolumeClaim gets the PersistentVolumeClaim for the given name and namespace
+func (s *DataStore) GetPersistentVolumeClaim(namespace, pvcName string) (*corev1.PersistentVolumeClaim, error) {
+	resultRO, err := s.kubeClient.CoreV1().PersistentVolumeClaims(namespace).Get(pvcName, metav1.GetOptions{})
```
joshimoo, line 452: Use the cache instead of the direct api retrieval.

datastore/kubernetes.go
```diff
@@ -407,12 +407,22 @@ func (s *DataStore) UpdatePersistentVolume(pv *corev1.PersistentVolume) (*corev1
...
+// GetPersistentVolumeRO gets the PersistentVolume from the index for the
 // given name
-func (s *DataStore) GetPersistentVolume(pvName string) (*corev1.PersistentVolume, error) {
+func (s *DataStore) GetPersistentVolumeRO(pvName string) (*corev1.PersistentVolume, error) {
 	return s.pvLister.Get(pvName)
 }
 
+// GetPersistentVolume gets the PersistentVolume for the given name
+func (s *DataStore) GetPersistentVolume(pvName string) (*corev1.PersistentVolume, error) {
+	resultRO, err := s.kubeClient.CoreV1().PersistentVolumes().Get(pvName, metav1.GetOptions{})
```
joshimoo, line 418: Use the cache instead of the direct api retrieval.

joshimoo (changes requested):
You are doing the deep copy now correctly, but you are retrieving from the api server directly which is unnecessary.

datastore/kubernetes.go
cclhsu: changed

datastore/kubernetes.go
cclhsu: changed

datastore/kubernetes.go
```diff
@@ -408,12 +408,22 @@ func (s *DataStore) UpdatePersistentVolume(pv *corev1.PersistentVolume) (*corev1
 	return s.kubeClient.CoreV1().PersistentVolumes().Update(pv)
 }
 
-// GetPersistentVolume gets the PersistentVolume from the index for the
+// GetPersistentVolumeRO gets the PersistentVolume from the index for the
 // given name
-func (s *DataStore) GetPersistentVolume(pvName string) (*corev1.PersistentVolume, error) {
+func (s *DataStore) GetPersistentVolumeRO(pvName string) (*corev1.PersistentVolume, error) {
```
joshimoo, line 413: You should add the below to the doc comment, to be consistent this also pops up during auto complete, so the consumer of the function can see this.

```
// This function returns direct reference to the internal cache object and should not be mutated.
// Consider using this function when you can guarantee read only access and don't want the overhead of deep copies
```

datastore/kubernetes.go
```diff
@@ -432,12 +442,22 @@ func (s *DataStore) UpdatePersistentVolumeClaim(namespace string, pvc *corev1.Pe
 	return s.kubeClient.CoreV1().PersistentVolumeClaims(namespace).Update(pvc)
 }
 
-// GetPersistentVolumeClaim gets the PersistentVolumeClaim from the
+// GetPersistentVolumeClaimRO gets the PersistentVolumeClaim from the
 // index for the given name and namespace
-func (s *DataStore) GetPersistentVolumeClaim(namespace, pvcName string) (*corev1.PersistentVolumeClaim, error) {
+func (s *DataStore) GetPersistentVolumeClaimRO(namespace, pvcName string) (*corev1.PersistentVolumeClaim, error) {
```
joshimoo, line 447: Same as above, have a look at the surrounding code + comments.
particularly the GetConfigMapRO immediately below.

joshimoo (commented):
Add the mutation warning comment.

datastore/kubernetes.go
cclhsu: changed

joshimoo (approved):
LGTM

PR #931, opened 2021-06-17
Add pprof and disable-profiler cmd flag

c3y1huang:
Add pprof and enabled by default for Longhorn manager. Users can set cmd option flag `--disable-profiler` at deployment.

https://github.com/longhorn/longhorn/issues/2696

### DIFF
--- a/api/router.go
+++ b/api/router.go
@@ -2,6 +2,7 @@ package api
 
 import (
 	"net/http"
+	"net/http/pprof"
 	"time"
 
 	"github.com/gorilla/mux"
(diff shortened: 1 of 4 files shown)


api/router.go
```diff
@@ -174,5 +176,25 @@ func NewRouter(s *Server) *mux.Router {
...
+	r.HandleFunc("/debug/pprof/cmdline", pprof.Cmdline)
+	r.HandleFunc("/debug/pprof/profile", pprof.Profile)
+	r.HandleFunc("/debug/pprof/symbol", pprof.Symbol)
+	r.HandleFunc("/debug/pprof/trace", pprof.Trace)
+	r.Handle("/debug/pprof/allocs", pprof.Handler("allocs"))
+	r.Handle("/debug/pprof/block", pprof.Handler("block"))
+	r.Handle("/debug/pprof/goroutine", pprof.Handler("goroutine"))
+	r.Handle("/debug/pprof/heap", pprof.Handler("heap"))
+	r.Handle("/debug/pprof/mutex", pprof.Handler("mutex"))
+	r.Handle("/debug/pprof/threadcreate", pprof.Handler("threadcreate"))
```
innobead, line 198: Probably we don't need to this? This should be already handled in `r.HandleFunc("/debug/pprof/", pprof.Index)`?

app/daemon.go
```diff
@@ -109,6 +114,8 @@ func startManager(c *cli.Context) error {
 
 	defaultSettingPath := os.Getenv(types.EnvDefaultSettingPath)
 
+	types.SetProfilerDisabled(c.Bool(FlagDisableProfiler))
```
innobead, line 117: Let's get the falg and pass it to the server directly instead of saving in another place. 

```
server := api.NewServer(m, wsc, profileEnabled).
```

types/setting.go
```diff
@@ -1041,3 +1041,13 @@ func UnmarshalNodeSelector(nodeSelectorSetting string) (map[string]string, error
...
+
+var disableProfiler bool = false
+
+func SetProfilerDisabled(value bool) {
+	disableProfiler = value
+}
+
+func IsProfilerDisabled() bool {
+	return disableProfiler
+}
```
innobead, line 1053: We don't need this global variable for now, so please remove them first. Thanks.

innobead (changes requested):

api/router.go
c3y1huang: I don't think this gets handled by pprof.Index. If we remove this then accessing the uri give `404 page not found`.

types/setting.go
c3y1huang: Removed.

app/daemon.go
c3y1huang: Addressed.

innobead (approved):

joshimoo (commented):
See comment my comment here: https://github.com/longhorn/longhorn/issues/2696#issuecomment-863995649

I am concerned about the current implementation expsosing this via the regular api server.

PR #933, opened 2021-06-17
Fix drain stuck forever for unreplicated volume

cclhsu:
#### Proposed Changes ####

Fix drain stuck forever for unreplicated volume to resolve issue '[BUG] kubectl drain node gets stuck forever'

#### Linked Issues ####

Refs: # https://github.com/longhorn/longhorn/issues/2673

#### Further Comments ####

None

### DIFF
--- a/controller/instance_manager_controller.go
+++ b/controller/instance_manager_controller.go
@@ -10,7 +10,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	policyv1beta1 "k8s.io/api/policy/v1beta1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


controller/instance_manager_controller.go
```diff
@@ -590,6 +590,10 @@ func (imc *InstanceManagerController) canDeleteInstanceManagerPDB(im *longhorn.I
 				hasHealthyReplicaOnAnotherNode = true
 				break
 			}
+			if r.Spec.HealthyAt == "" && r.Spec.FailedAt == "" && r.Spec.NodeID == imc.controllerID {
+				hasHealthyReplicaOnAnotherNode = true
```
shuo-wu, line 594: Please use an separate bool variable and add some comments for this case. e.g., `isUnusedReplicaOnCurrentNode`

shuo-wu (changes requested):

controller/instance_manager_controller.go
cclhsu: changed

controller/instance_manager_controller.go
```diff
@@ -585,14 +585,21 @@ func (imc *InstanceManagerController) canDeleteInstanceManagerPDB(im *longhorn.I
...
 		hasHealthyReplicaOnAnotherNode := false
+		isUnusedReplicaOnCurrentNode := false
 		for _, r := range replicas {
 			if r.Spec.HealthyAt != "" && r.Spec.FailedAt == "" && r.Spec.NodeID != imc.controllerID {
 				hasHealthyReplicaOnAnotherNode = true
 				break
 			}
+			// Fix drained node `pod's disruption budget` issue, when LHV is is created through Longhorn UI,
+			// but has not yet attached and replicated.
+			if r.Spec.HealthyAt == "" && r.Spec.FailedAt == "" && r.Spec.NodeID == imc.controllerID {
```
shuo-wu, line 596:
```
If a replica has never been started, there is no data stored in this replica, and retaining it makes no sense for HA.
Hence Longhorn doesn't need to block the PDB removal for the replica.
This case typically happens on a newly created volume that hasn't been attached to any node.
```

You can add the Github link if you prefer.

shuo-wu (changes requested):

controller/instance_manager_controller.go
cclhsu: changed

shuo-wu (approved):

PR #934, opened 2021-06-17
controller: Stop monitoring if the streaming does not work

shuo-wu:
longhorn/longhorn#2561

This is a surgery fix. IIRC, we need to refactor the monitor implementation in all controllers. @joshimoo Do you remember the issue number?


joshimoo:
resource monitoring refactor is tracked at https://github.com/longhorn/longhorn/issues/2441
with a reference to my comment here https://github.com/longhorn/longhorn-manager/pull/854#issuecomment-812101940

jenting (approved):

innobead (approved):

PR #938, opened 2021-06-18
Add pprof http endpoint

joshimoo:
To fix https://github.com/longhorn/longhorn/issues/2715

Please revert the following PR's
- https://github.com/longhorn/longhorn-manager/pull/931
- https://github.com/longhorn/longhorn-manager/pull/937

deployment reverts:
- https://github.com/longhorn/longhorn/pull/2700
- https://github.com/longhorn/longhorn/pull/2709

website reverts:
- https://github.com/longhorn/website/pull/324
- https://github.com/longhorn/website/pull/325

Afterwards this PR can be merged.
It's based on the latest commit before the above commits
https://github.com/longhorn/longhorn-manager/commit/d32898046ab984db6c88b5980bcef2f181c77a1e

See comment here: https://github.com/longhorn/longhorn/issues/2696#issuecomment-863995649

Signed-off-by: Joshua Moody <joshua.moody@suse.com>

### DIFF
--- a/app/daemon.go
+++ b/app/daemon.go
@@ -5,6 +5,8 @@ import (
 	"net/http"
 	"os"
 
+	_ "net/http/pprof" // for runtime profiling
+
 	"github.com/gorilla/handlers"
 	"github.com/sirupsen/logrus"
 	"github.com/urfave/cli"
@@ -185,7 +187,6 @@ func startManager(c *cli.Context) error {
 
 	server := api.NewServer(m, wsc)
 	router := http.Handler(api.NewRouter(server))
-
 	router = util.FilteredLoggingHandler(map[string]struct{}{
 		"/v1/apiversions":  {},
 		"/v1/schemas":      {},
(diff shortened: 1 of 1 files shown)


c3y1huang:
Reverted.

app/daemon.go
```diff
@@ -198,10 +199,19 @@ func startManager(c *cli.Context) error {
 	router = handlers.ProxyHeaders(router)
 
 	listen := types.GetAPIServerAddressFromIP(currentIP)
-	logrus.Infof("Listening on %s", listen)
+	logger.Infof("Listening on %s", listen)
 
 	go http.ListenAndServe(listen, router)
 
+	go func() {
+		debugAddress := "127.0.0.1:6060"
```
yasker, line 207: Should we allow customize debug server IP/port just in case?

joshimoo: Agree in general that we should allow customization. With appropriate documentation:

For v1.1.2 I wanted to go with the simple approach, since we then don't need to document this just yet.
For v1.2 we should try to have a consistent design for our deployments and allow the user to enable the ability to expose the endpoint outside of the local pod host.

For example an ENV that by default is set to "127.0.0.1:6060" with explanations on how to make this 
- listen to all interfaces "0.0.0.0:6060" 
- listen to just the pod interface

innobead (approved):
LGTM. Please also backport to 1.1.2.

PR #948, opened 2021-06-23
Fix regression v1.1.2 expand with static storage class

cclhsu:
#### Proposed Changes ####

Fix regression v1.1.2 expand with static storage class to resolve issue 'Application couldn't mount the volume after expanded the volume'

Get static class name from setting `default-longhorn-static-storage-class`.  If class name exist, continue with PVC expansion.  If not exist, continue with LHV expansion.

#### Types of Changes ####
 
enhancement

#### Verification ####

* Case 1: From UI, PVC expand Volume with `default-longhorn-static-storage-class` set as `longhorn-<CUSTOMIZED-NON-EXISTED-CLASS-NAME>`

Expand will success.

#### Linked Issues ####

Refs: Longhorn: #2692

#### Further Comments ####

- Issue discussion: https://rancher.slack.com/archives/C01G91NUHFY/p1624409225268200?thread_ts=1624316294.238500&cid=C01G91NUHFY

### DIFF
--- a/manager/volume.go
+++ b/manager/volume.go
@@ -11,6 +11,7 @@ import (
 	"github.com/sirupsen/logrus"
 
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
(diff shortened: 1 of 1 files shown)


manager/volume.go
```diff
@@ -637,7 +637,18 @@ func (m *VolumeManager) Expand(volumeName string, size int64) (v *longhorn.Volum
...
+		longhornStaticStorageClass, err := m.ds.GetSettingValueExisted(types.SettingNameDefaultLonghornStaticStorageClass)
+		if err != nil {
+			return nil, err
+		}
+
+		isStaticStorageClass := false
+		scName := pvc.Spec.StorageClassName
+		if _, err := m.ds.GetStorageClassRO(*scName); err != nil && *scName == longhornStaticStorageClass {
+			isStaticStorageClass = true
+		}
```
shuo-wu, line 649:
```
pvcSCName := *pvc.Spec.StorageClassName
if pvcSCName == longhornStaticStorageClass {
    if _, err := m.ds.GetStorageClassRO(pvcSCName); err != nil {
        if !apierrors.IsNotExist(err) {
            return nil, err
        }
        isStaticStorageClass = true
    }
}
```

manager/volume.go
```diff
@@ -637,7 +637,18 @@ func (m *VolumeManager) Expand(volumeName string, size int64) (v *longhorn.Volum
...
+
+		if pvcSpecValue.Cmp(requestedSize) < 0 && !isStaticStorageClass {
```
shuo-wu, line 651: I am thinking if Longhorn still needs to round up PVC spec size if a static SC is used.

shuo-wu (changes requested):

manager/volume.go

manager/volume.go
cclhsu: changed.

cclhsu: Add condition to check static storage class, since volume size can be mounted and expanded as it is

shuo-wu: You can refactor this function like the following:
```
		size, err := m.checkAndExpandPVC(kubernetesStatus.Namespace, kubernetesStatus.PVCName, size)
		if err != nil {
			return nil, err
		}
	}
......


func (m *VolumeManager) checkAndExpandPVC(namespace, pvcName string, size int64) (size int64, err error) {
	pvc, err := m.ds.GetPersistentVolumeClaim(kubernetesStatus.Namespace, kubernetesStatus.PVCName)
	if err != nil {
		return nil, err
	}

	// Backward compatibility
	pvcSCName := *pvc.Spec.StorageClassName
	if pvcSCName == longhornStaticStorageClass {
		if _, err := m.ds.GetStorageClassRO(pvcSCName); err != nil {
			if !apierrors.IsNotFound(err) {
				return nil, err
			}
			// Do nothing for old PVCs created by a static StorageClass
			return size, nil
		}
	}

	requestedSize := resource.MustParse(strconv.FormatInt(size, 10))

	// TODO: Should check for pvc.Spec.Resources.Requests.Storage() here, once upgrade API to v0.18.x.
	pvcSpecValue, ok := pvc.Spec.Resources.Requests[corev1.ResourceStorage]
	if !ok {
		return nil, fmt.Errorf("cannot get request storage")
	}
	if pvcSpecValue.Cmp(requestedSize) < 0 {
......
}
```

shuo-wu (changes requested):

manager/volume.go
cclhsu: Let me see how to change it, since part of function return `volume` instead of 'size`

cclhsu: refactored.

manager/volume.go
```diff
@@ -681,6 +657,62 @@ func (m *VolumeManager) Expand(volumeName string, size int64) (v *longhorn.Volum
 	return v, nil
 }
 
+func (m *VolumeManager) checkAndExpandPVC(namespace string, pvcName string, size int64) (s int64, waitForPVCExpansion bool, err error) {
+	waitForPVCExpansion = false
+
+	pvc, err := m.ds.GetPersistentVolumeClaim(namespace, pvcName)
+	if err != nil {
+		return -1, waitForPVCExpansion, err
```
shuo-wu, line 665: 1. The boolean arg should be the 1st arg of the output args.
2. For this function, It's better to directly specify `true` or `false` rather than using a variable in `return`.

shuo-wu (changes requested):

manager/volume.go

shuo-wu (approved):

PR #951, opened 2021-06-25
Fix regression v1.1.2 expand with static storage class

cclhsu:
#### Proposed Changes ####

Fix regression v1.1.2 expand with static storage class to resolve issue 'Application couldn't mount the volume after expanded the volume'

Get static class name from setting `default-longhorn-static-storage-class`.  If class name exist, continue with PVC expansion.  If not exist, continue with LHV expansion.

#### Types of Changes ####
 
enhancement

#### Verification ####

* Case 1: From UI, PVC expand Volume with `default-longhorn-static-storage-class` set as `longhorn-<CUSTOMIZED-NON-EXISTED-CLASS-NAME>`

Expand will success.

#### Linked Issues ####

Refs: Longhorn: #2692

#### Further Comments ####

- Issue discussion: https://rancher.slack.com/archives/C01G91NUHFY/p1624409225268200?thread_ts=1624316294.238500&cid=C01G91NUHFY

(cherry picked from commit ba88af4015a17163c89a981830cf388182265d94)

### DIFF
--- a/manager/volume.go
+++ b/manager/volume.go
@@ -11,6 +11,7 @@ import (
 	"github.com/sirupsen/logrus"
 
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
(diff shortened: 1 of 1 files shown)


innobead (approved):
LGTM. 

We need to make sure the change in the master and 1.1.2 first, even though there will be a possible change in 1.2.0.

PR #952, opened 2021-06-28
Longhorn backup refactoring

jenting:
#### Propose Changes

- Bump longhorn/backupstore to support:
  - new `backup list` command behavior.
  - new `backup inspect-volume` command.
  - new `backup head` command.
- Add new CRDs:
   - BackupTargets.
   - BackupVolumes.
   - Backups.
- Update `setting_controller`.
- Add new controllers:
  - `backup_target_controller`.
  - `backup_volume_controller`.
  - `backup_controller`.
- Add BackupVolume and Backup informer to the `websocket_controller`.

#### Issue

https://github.com/longhorn/longhorn/issues/1761


controller/setting_controller.go
```diff
@@ -626,31 +616,28 @@ func (sc *SettingController) updateNodeSelector() error {
 	return nil
 }
 
-func (bm *BackupStoreMonitor) Start() {
+func (bm *BackupStoreTimer) Start() {
```
shuo-wu, line 619: This function is responsible for blindly updating `backupTarget.spec.syncRequestedAt` only.

controller/backup_target_controller.go
```diff
@@ -0,0 +1,313 @@
...
+	backupTargetAvailable := backupTarget.Status.Available
+	backupTargetLastSyncedAt := backupTarget.Status.LastSyncedAt
+
+	defer func(backupTargetAvailable *bool, backupTargetLastSyncedAt **metav1.Time) {
+		if backupTargetAvailable == nil || backupTargetLastSyncedAt == nil {
+			log.Info("BUG: backupTargetAvailable or backupTargetLastSyncedAt is nil")
+			return
+		}
+
+		if backupTarget.Spec.BackupTargetURL == "" {
```
shuo-wu, line 193: Abstract this as a cleanup function.

controller/backup_target_controller.go
```diff
@@ -0,0 +1,313 @@
...
+					log.WithError(err).Errorf("Error deleting backup volume %s from cluster", backupVolumeName)
+				}
+			}
+		}
+
+		backupTarget.Status.Available = *backupTargetAvailable
+		backupTarget.Status.LastSyncedAt = *backupTargetLastSyncedAt
+		if _, err = btc.ds.UpdateBackupTargetStatus(backupTarget); err != nil && !datastore.ErrorIsConflict(err) {
+			log.WithError(err).Error("Error updating backup target status")
+		}
```
shuo-wu, line 212: For this part, you can follow the pattern in other controllers:
  1. Set the status field
  2. Do deep equal then determine if calling status update API is necessary.

Besides, there is no need to set the field `backupTarget.Status.Available = *backupTargetAvailable` here. If this field becomes false, the reconcile loop will return immediately.

controller/backup_target_controller.go
```diff
@@ -0,0 +1,313 @@
...
+	}(&backupTargetAvailable, &backupTargetLastSyncedAt)
+
```
shuo-wu, line 214: No owner ID check and update

controller/backup_target_controller.go
```diff
@@ -0,0 +1,313 @@
...
+	if err != nil {
+		log.WithError(err).Error("Error listing backup volumes from backup target")
+		backupTargetAvailable = false
+		backupTargetLastSyncedAt = &metav1.Time{Time: time.Now().UTC()}
+		// Ignore error to prevent enqueue
+		return nil
+	}
+
+	backupStoreBackupVolumes := sets.NewString(res...)
+	log.WithField("BackupVolumeCount", len(backupStoreBackupVolumes)).Debug("Got backup volumes from backup target")
```
shuo-wu, line 242: There is no need to print this

controller/backup_target_controller.go
```diff
@@ -0,0 +1,313 @@
...
+
+	// Get a list of all the backup volumes that exist as custom resources in the cluster
+	clusterBackupVolumes, err := btc.ds.ListBackupVolume()
+	if err != nil {
+		log.WithError(err).Error("Error listing backup volumes in the cluster, proceeding with pull into cluster")
```
shuo-wu, line 247: If you cannot list the current CRs, how can you continue the proceeding?

controller/backup_target_controller.go
```diff
@@ -0,0 +1,313 @@
...
+	} else {
+		log.WithField("clusterBackupVolumeCount", len(clusterBackupVolumes)).Debug("Got backup volumes in the cluster")
```
shuo-wu, line 249: No need to print this. We only need to log backup volume creation or deletion.

controller/backup_target_controller.go
```diff
@@ -0,0 +1,313 @@
...
+	}
+	for backupVolumeName := range backupVolumesToPull {
+		log.WithField("backupVolume", backupVolumeName).Info("Attempting to pull backup volume into cluster")
+
+		backupVolumeMetadataURL := backupstore.EncodeMetadataURL("", backupVolumeName, backupTarget.Spec.BackupTargetURL)
+		backupVolumeInfo, err := backupTargetClient.InspectBackupVolumeConfig(backupVolumeMetadataURL)
+		if err != nil || backupVolumeInfo == nil {
+			log.WithError(err).Error("Error getting backup volume metadata from backup target")
+			continue
+		}
```
shuo-wu, line 271: I think this is redundant. BackupVolumeController will be responsible for inspecting it.

controller/backup_target_controller.go
```diff
@@ -0,0 +1,313 @@
...
+		btc.queue.AddRateLimited(key)
+		return
+	}
+
+	utilruntime.HandleError(err)
+	btc.logger.WithError(err).Warnf("Dropping Longhorn backup target %v out of the queue", key)
+	btc.queue.Forget(key)
+}
+
+func (btc *BackupTargetController) reconcile(log logrus.FieldLogger, name string) (err error) {
```
shuo-wu, line 164: Typically the reconcile function receive string `key` only

controller/backup_volume_controller.go
```diff
@@ -0,0 +1,375 @@
...
+			}
+
+			if err := backupTargetClient.DeleteBackupVolume(backupVolumeName); err != nil {
+				log.WithError(err).Error("Error deleting remote backup volume")
+				return err
+			}
+		}
+
+		// Clean up Volume CR status
+		if err := manager.SyncVolumeLastBackupWithBackupVolume(backupVolumeName, nil, bvc.ds.GetVolume, bvc.ds.UpdateVolumeStatus); err != nil {
```
shuo-wu, line 247: I think it's VolumeController's responsibility to sync volume CRs with backupVolume CRs now.

controller/backup_controller.go
```diff
@@ -0,0 +1,396 @@
...
+	if err != nil {
+		return err
+	}
+
+	credential, err := manager.GetBackupCredentialConfig(bc.ds)
+	if err != nil {
+		return err
+	}
+
+	volumeManager := manager.NewVolumeManager(bc.controllerID, bc.ds)
```
shuo-wu, line 179: Do not use package `manager` in `controller`. IMO, that package is for HTTP API handling.

controller/backup_controller.go
```diff
@@ -0,0 +1,396 @@
...
+	engine, err := volumeManager.GetEngineClient(volumeName)
+	if err != nil {
+		return err
+	}
+
+	// Blocks till the backup snapshot creation has been started
+	_, err = engine.SnapshotBackup(
```
shuo-wu, line 186: I would recommend making this part as a part of an async call if possible. Since this API may block for a while.

shuo-wu (changes requested):
Some issues happen in all of the new controllers. Please follow the comments of the BackupTargetController and check other controllers.

I will re-review this PR after resolving existing comments.

engineapi/backups.go
```diff
@@ -85,147 +85,149 @@ func (b *BackupTarget) ExecuteEngineBinaryWithoutTimeout(args ...string) (string
...
 
-func (b *BackupTarget) GetVolume(volumeName string) (*BackupVolume, error) {
-	output, err := b.ExecuteEngineBinary("backup", "ls", "--volume", volumeName, "--volume-only", b.URL)
+// DeleteBackupVolume deletes the backup volume from the remote backup target
+func (b *BackupTarget) DeleteBackupVolume(volumeName string) error {
+	_, err := b.ExecuteEngineBinaryWithoutTimeout("backup", "rm", "--volume", volumeName, b.URL)
 	if err != nil {
-		if strings.Contains(err.Error(), "msg=\"cannot find ") {
-			return nil, nil
+		if types.ErrorIsNotFound(err) {
```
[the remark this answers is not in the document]
jenting: I think I did it already, am I miss it elsewhere?

controller/engine_controller.go
```diff
@@ -1025,7 +1025,8 @@ func checkSizeBeforeRestoration(log logrus.FieldLogger, engine *longhorn.Engine,
 		return false, errors.Wrapf(err, "engine monitor: Cannot generate BackupTarget for expansion check of the DR volume engine %v", engine.Name)
 	}
 
-	bv, err := backupTarget.GetVolume(engine.Spec.BackupVolume)
+	backupVolumeMetadataURL := backupstore.EncodeMetadataURL("", engine.Spec.BackupVolume, backupTarget.URL)
```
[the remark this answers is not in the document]
jenting: I'll update the wording in backupstore repo.

datastore/longhorn.go
```diff
@@ -2394,3 +2394,312 @@ func (s *DataStore) ListShareManagers() (map[string]*longhorn.ShareManager, erro
...
+	obj, err := s.lhClient.LonghornV1beta1().BackupTargets(s.namespace).UpdateStatus(backupTarget)
+	if err != nil {
+		return nil, err
+	}
+	verifyUpdate(backupTarget.Name, obj, func(name string) (runtime.Object, error) {
+		return s.GetBackupTargetRO(name)
+	})
+	return obj, nil
+}
+
```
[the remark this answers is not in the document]
jenting: No, even the Backup Target URL is empty, we keep the CR exists.

controller/engine_controller.go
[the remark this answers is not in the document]
shuo-wu: How about `EncodeBackupURL`? The keyword is `backup` rather than `config` or `metadata`

engineapi/backups.go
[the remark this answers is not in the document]
shuo-wu: I mean, you can do this replacement for all functions in `engineapi/backup.go`

controller/engine_controller.go
```diff
@@ -24,17 +27,13 @@ import (
 	"k8s.io/client-go/util/flowcontrol"
 	"k8s.io/kubernetes/pkg/controller"
 
-	imapi "github.com/longhorn/longhorn-instance-manager/pkg/api"
-	imutil "github.com/longhorn/longhorn-instance-manager/pkg/util"
-
```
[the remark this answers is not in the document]
shuo-wu: Don't forget the imports in other files. e.g., `engineapi/backups.go`

manager/engine.go
```diff
@@ -270,121 +271,135 @@ func (m *VolumeManager) GetEngineClient(volumeName string) (client engineapi.Eng
...
+			Messages:         backupVolume.Status.Messages,
+			BackingImageName: backupVolume.Status.BackingImageName,
+			BackingImageURL:  backupVolume.Status.BackingImageURL,
+		}
 	}
 
 	// side effect, update known volumes
-	SyncVolumesLastBackupWithBackupVolumes(backupVolumes, m.ds.ListVolumes, m.ds.GetVolume, m.ds.UpdateVolumeStatus)
-	return backupVolumes, nil
+	SyncVolumesLastBackupWithBackupVolumes(bvs, m.ds.ListVolumes, m.ds.GetVolume, m.ds.UpdateVolumeStatus)
```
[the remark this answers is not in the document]
jenting: The purpose is for DR volume to update its lastBackupName and lastBackupAt in the volume CR.
Probably the better way is moving it into the volume_controller.
But my original thought was to make another PR to address it.

controller/engine_controller.go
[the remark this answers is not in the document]
jenting: Do we have the import convention? Similar to below?
```golang
go-standard-library

3rd-party-library

kubernetes-library

longhorn-3rd-party-library

longhorn-manager-library

longhorn-manager-kubernetes-library
```

controller/setting_controller.go
```diff
@@ -259,69 +254,64 @@ func (sc *SettingController) syncBackupTarget() (err error) {
 	if err != nil {
 		return err
 	}
+	pollInterval := time.Duration(interval) * time.Second
```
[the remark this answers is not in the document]
shuo-wu: I just realize `ownerID` is typically in CR status. It doesn’t apply to this backup target sync function of the setting controller.
We can use the function `getResponsibleNodeID` defined in `BackupStoreTimer.Start()` to determine if a backup target CR can be updated by the current controller.

controller/backup_target_controller.go
jenting: > Besides, there is no need to set the field `backupTarget.Status.Available = *backupTargetAvailable` here. If this field becomes false, the reconcile loop will return immediately.

Not quite understand this one. I'd want to update the `status.available` once in the reconcile loop, it detects the remote backup target suddenly unavailable.

shuo-wu: I mean, you don't need this line at all:
```
backupTarget.Status.Available = *backupTargetAvailable
```

controller/setting_controller.go
```diff
@@ -259,69 +254,64 @@ func (sc *SettingController) syncBackupTarget() (err error) {
...
-		stopCh: make(chan struct{}),
+	if sc.bsTimer != nil {
+		if sc.bsTimer.backupTargetURL != targetSetting.Value ||
+			sc.bsTimer.backupTargetCredentialSecret != secretSetting.Value ||
+			sc.bsTimer.pollInterval != pollInterval {
+			// Stop backup store timer
+			sc.bsTimer.Stop()
+			sc.bsTimer = nil
+		}
 	}
```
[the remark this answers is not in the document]
jenting: This is for better user experience if the user configures backup target URL and backup target credential secret at the same time. It restarts the timer and the timer sets BackupTarget CR `spec.syncRequestAt=time.Now()`.

Otherwise, if the user configures backup target URL and backup target credential secret at the same time, the setting_controller first sets backup target URL -> creates the timer -> sets backup target credential secret, the backup_target_controller be triggered after the poll interval time reached (default need to wait 5 mins).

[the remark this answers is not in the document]
jenting: This is for better user experience if the user configures backup target URL and backup target credential secret at the same time. It restarts the timer and the timer sets BackupTarget CR `spec.syncRequestAt=time.Now()`.

Otherwise, if the user configures backup target URL and backup target credential secret at the same time, the setting_controller first sets backup target URL -> creates the timer -> sets backup target credential secret, the backup_target_controller be triggered after the poll interval time reached (default need to wait 5 mins).

controller/backup_target_controller.go
jenting: But there is probably the backup target available status from true to false.

manager/engine.go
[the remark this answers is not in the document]
shuo-wu: Yes. It's VolumeController's responsibility to update the volume status. You need to ask VolumeControlller to be informed by BackupVolume.

controller/setting_controller.go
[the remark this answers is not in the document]
shuo-wu: OK. It should be starting the timer as long as the backup target is **available** and the poll interval is set. And when the backup target spec is updated, Longhorn needs to set `spec.syncRequestedAt` immediately. Here, we don't need to care if the backupstore URL and the secret are set simultaneously.

controller/backup_target_controller.go
shuo-wu: In the below code, you will set this status field to false then return from the current reconcile loop (then jump into this defer function) once the backup target is unavailable, right? Or in other words, once the reconcile loop enters into this defer function, `backupTarget.Status.Available` is already up-to-date. 

Similar issue to `backupTarget.Status.lastSyncedAt`

controller/backup_controller.go
jenting: I'd prefer to address this in other PR since there the `engine_controller` and `volume_controller` also use package `manager` in `controller`
https://github.com/longhorn/longhorn-manager/blob/f53ebaf/controller/engine_controller.go#L1024
https://github.com/longhorn/longhorn-manager/blob/f53ebaf/controller/engine_controller.go#L1066-L1071
https://github.com/longhorn/longhorn-manager/blob/f53ebaf/controller/volume_controller.go#L2321

controller/backup_controller.go
```diff
@@ -0,0 +1,444 @@
...
+					}
+					if _, err = bc.ds.CreateBackupVolume(backupVolume); err != nil {
+						log.WithError(err).Errorf("Error creating backup volume %s into cluster", volumeName)
+					}
+				}
+
+				bc.eventRecorder.Eventf(backup, corev1.EventTypeNormal, EventReasonSucceeded,
+					"Volume %s snapshot %s label %v", volumeName, backup.Spec.SnapshotName, backup.Spec.Labels)
+				log.Debugf("Backup snapshot succeeded")
+				break
```
shuo-wu, line 276: I am thinking about how the backup volume config file gets created,. If it's created automatically when the 1st backup is ready. We can leave the backup target controller to pull it and create the CR. Then the backup target controller is the only owner of the resource. (only one that is responsible for managing BackupVolume CR lifecycle.)

controller/backup_controller.go
```diff
@@ -0,0 +1,444 @@
...
+func (bc *BackupController) getBackupVolume(log logrus.FieldLogger, backup *longhorn.Backup) (string, error) {
+	backupVolumeName, ok := backup.Labels[types.LonghornLabelBackupVolume]
+	if !ok {
+		log.Warn("Cannot find the backup volume label")
+		return "", fmt.Errorf("Cannot find the backup volume label")
+	}
+	return backupVolumeName, nil
+}
+
+func (bc *BackupController) getBackupMetadataURL(log logrus.FieldLogger, backup *longhorn.Backup, backupTargetURL string) (string, error) {
```
shuo-wu, line 294: `getBackupURL`

controller/backup_controller.go
```diff
@@ -0,0 +1,444 @@
...
+			}
+		}
+
+		// Request backup_volume_controller to reconcile BackupVolume immediately if it's the last backup
+		volumeName, err := bc.getBackupVolume(log, backup)
+		if err != nil {
+			return err
+		}
+		backupVolume, err := bc.ds.GetBackupVolume(volumeName)
+		if err == nil &&
```
shuo-wu, line 366:
```
if err != nil && ! apierrors.IsNotFound(err) {
    return err
}
if backupVolume != nil {
    // Update 
}
```

controller/backup_controller.go
```diff
@@ -0,0 +1,444 @@
...
+				// Ignore error to prevent enqueue
+				return nil
+			}
+
+			if err := backupTargetClient.DeleteBackup(backupURL); err != nil {
+				log.WithError(err).Error("Error deleting remote backup")
+				return err
+			}
+		}
+
```
shuo-wu, line 359: User a defer function to update the CR status.

manager/engine.go
```diff
@@ -149,7 +153,7 @@ func (m *VolumeManager) PurgeSnapshot(volumeName string) error {
 	return nil
 }
 
-func (m *VolumeManager) BackupSnapshot(volumeName, snapshotName, backingImageName, backingImageURL string, labels map[string]string) error {
+func (m *VolumeManager) BackupSnapshot(backupName, volumeName, snapshotName, backingImageName, backingImageURL string, labels map[string]string) error {
```
shuo-wu, line 156: We can return a CR immediately now.

manager/engine.go
```diff
@@ -158,65 +162,31 @@ func (m *VolumeManager) BackupSnapshot(volumeName, snapshotName, backingImageNam
...
-				break
-			}
-			time.Sleep(BackupStatusQueryInterval)
-		}
-
-		if err := UpdateVolumeLastBackup(volumeName, target, m.ds.GetVolume, m.ds.UpdateVolumeStatus); err != nil {
-			logrus.Warnf("Failed to update volume LastBackup for %v: %v", volumeName, err)
-		}
-	}()
 	return nil
```
shuo-wu, line 190: NIT: `return m.ds.CreateBackup(backupCR, volumeName)`

manager/volume.go
```diff
@@ -571,9 +573,8 @@ func (m *VolumeManager) Activate(volumeName string, frontend string) (v *longhor
 		return nil, fmt.Errorf("invalid frontend %v", frontend)
 	}
 
-	// ListBackupVolumes() will trigger the update for LastBackup
-	_, err = m.ListBackupVolumes()
-	if err != nil {
+	// Trigger update the LastBackup
+	if err := m.ForceSyncBackupVolume(volumeName); err != nil {
 		logrus.Warnf("failed to update LastBackup and backup volume list before activating standby volume %v: %v", volumeName, err)
 	}
```
shuo-wu, line 579: VolumeController should be responsible for sync backup fields of volumes with BackupVolume CRs. No need to do anything here now.

controller/backup_volume_controller.go
```diff
@@ -0,0 +1,390 @@
...
+func (bvc *BackupVolumeController) cleanupBackups(backupVolumeName string, fileCleanupRequired bool) error {
+	backups, err := bvc.ds.ListBackupsWithBackupVolumeName(backupVolumeName)
+	if err != nil {
+		return err
+	}
+
+	var errs []string
+	if fileCleanupRequired {
+		for _, backup := range backups {
+			backup.Spec.FileCleanupRequired = true
```
shuo-wu, line 350:
```
if !backup.Spec.FileCleanupRequired {
    // update
}
```

controller/volume_controller.go
```diff
@@ -2322,9 +2333,10 @@ func (vc *VolumeController) checkAndInitVolumeRestore(v *longhorn.Volume) error
 	if err != nil {
 		return err
 	}
-	backup, err := backupTarget.GetBackup(v.Spec.FromBackup)
+
+	backup, err := backupTarget.InspectBackupConfig(v.Spec.FromBackup)
```
shuo-wu, line 2337: Can we directly use the Backup or BackupVolume CR instead? Then we can prevent other components except for backup related controllers to call the backup APIs.

upgrade/v112to120/upgrade.go
```diff
@@ -0,0 +1,158 @@
...
+	}()
+
+	const defaultBackupTargetName = "default"
+	_, err = lhClient.LonghornV1beta1().BackupTargets(namespace).Get(defaultBackupTargetName, metav1.GetOptions{})
+	if err == nil {
+		return nil
+	}
+	if err != nil && !datastore.ErrorIsNotFound(err) {
+		return err
+	}
```
shuo-wu, line 51: NIT: Typically we will check and return the error first.

shuo-wu (changes requested):

controller/setting_controller.go
```diff
@@ -259,69 +281,87 @@ func (sc *SettingController) syncBackupTarget() (err error) {
...
-	sc.bsMonitor = &BackupStoreMonitor{
-		logger:       sc.logger.WithField("component", "backup-store-monitor"),
-		controllerID: sc.controllerID,
-
-		backupTarget:                 targetSetting.Value,
-		backupTargetCredentialSecret: secretSetting.Value,
-
-		pollInterval: time.Duration(interval) * time.Second,
+	if responsibleNodeID == sc.controllerID {
```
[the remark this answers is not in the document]
jenting: All the nodes should start the timer, and only the responsible node can update the `spec.syncRequestAt`.
This prevents if the responsible node goes down and other nodes didn't have a timer to periodically update the `spec.syncRequestAt`.

controller/setting_controller.go
```diff
@@ -626,79 +666,42 @@ func (sc *SettingController) updateNodeSelector() error {
...
-		for node := range nodes {
-			candidates = append(candidates, node)
+			log.WithError(err).Warn("Failed to select node in backup store timer")
+			return
 		}
-
-		if len(candidates) == 0 {
-			return false, fmt.Errorf("no ready nodes with engine image %v available", defaultEngineImage)
+		if responsibleNodeID != bm.controllerID {
+			return
```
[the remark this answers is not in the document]
jenting: [1]

controller/backup_target_controller.go
```diff
@@ -0,0 +1,342 @@
...
+		logrus.Fields{
+			"url":      backupTarget.Spec.BackupTargetURL,
+			"cred":     backupTarget.Spec.CredentialSecret,
+			"interval": backupTarget.Spec.PollInterval.Duration,
+		},
+	)
+}
+
+func (btc *BackupTargetController) reconcile(name string) (err error) {
+	backupTarget, err := getDefaultBackupTarget(btc.ds)
```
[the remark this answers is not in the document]
jenting: Do you think we should salvage the default BackupTarget CR back?

controller/backup_volume_controller.go
```diff
@@ -0,0 +1,390 @@
...
+	backupVolumeMetadataURL := backupstore.EncodeBackupURL("", backupVolumeName, backupTarget.Spec.BackupTargetURL)
+	configMetadata, err := backupTargetClient.GetConfigMetadata(backupVolumeMetadataURL)
+	if err != nil {
+		log.WithError(err).Error("Error getting backup volume config metadata from backup target")
+		// Ignore error to prevent enqueue
+		return nil
+	}
+	if configMetadata == nil {
+		return nil
+	}
```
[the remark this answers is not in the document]
jenting: Since the backup config creates first and the create/update backup volume config.
This makes the new backup config has the chance to pull faster.

controller/backup_volume_controller.go
jenting: If the backup volume `spec.FileCleanupRequired=true`, it means that all the corresponding backups need to set `spec.FileCleanupRequired=true`. This is for the case when the user clicks delete all backups of a backup volume.

controller/setting_controller.go
```diff
@@ -259,69 +281,87 @@ func (sc *SettingController) syncBackupTarget() (err error) {
...
+		}
 	}
-	go sc.bsMonitor.Start()
-	return nil
-}
 
-func (sc *SettingController) updateBackupstorePollInterval() (err error) {
-	if sc.bsMonitor == nil {
+	// We only start the timer if the backup target is valid
+	if _, err = manager.GenerateBackupTarget(sc.ds); err != nil {
```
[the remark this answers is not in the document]
jenting: If we already know the backup target is invalid, why do we need to start the timer?

controller/volume_controller.go
jenting: Sure, lots of files using backup APIs, I'll update it.

controller/backup_controller.go
jenting: I need to double-check it again. But If I remember correctly, the backup volume config creates once the snapshot backup is started.

controller/setting_controller.go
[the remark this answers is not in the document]
shuo-wu: The sync function mentioned above means function `syncBackupTarget`

controller/backup_target_controller.go
```diff
@@ -0,0 +1,342 @@
...
+			ObjectMeta: metav1.ObjectMeta{
+				Name: backupVolumeName,
+			},
+			Spec: types.BackupVolumeSpec{
+				SyncRequestAt: &metav1.Time{Time: time.Now().Add(time.Second).UTC()},
+			},
+		}
+		if _, err = btc.ds.CreateBackupVolume(backupVolume); err != nil {
+			log.WithError(err).Errorf("Error creating backup volume %s into cluster", backupVolumeName)
+		}
```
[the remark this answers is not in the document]
jenting: return err

controller/backup_target_controller.go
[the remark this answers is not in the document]
jenting: Add `backupTargetInformer` to catch Delete event in setting_controller when the default BackupTarget CR be deleted, salvage the CR back.

api/backup.go
```diff
@@ -95,7 +95,6 @@ func (s *Server) BackupDelete(w http.ResponseWriter, req *http.Request) error {
 	if err := s.m.DeleteBackup(input.Name, volName); err != nil {
 		return errors.Wrapf(err, "error deleting backup %v of volume %v", input.Name, volName)
 	}
-	logrus.Debugf("Removed backup %v of volume %v", input.Name, volName)
```
jenting, line 98: The delete backup log in https://github.com/longhorn/longhorn-manager/blob/777d0b64b1c504dc178e7405b23d17af8e7cfd70/engineapi/backups.go#L244 already.

shuo-wu (approved):
In general LGTM. 
Since there will be another PR soon, we can address the issues I mentioned in that PR.

controller/backup_controller.go
```diff
@@ -0,0 +1,523 @@
...
+			backup.Name, backup.Spec.SnapshotName, url,
+			backup.Spec.BackingImage, backup.Spec.BackingImageChecksum,
+			backup.Spec.Labels, credential)
+		if err != nil {
+			backup.Status.State = types.BackupStateError
+			logEvent(err)
+			return
+		}
+
+		backup.Status.State = types.BackupStateInProgress
```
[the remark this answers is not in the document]
jenting: Addressed in #985.

controller/backup_target_controller.go
```diff
@@ -0,0 +1,377 @@
...
+		log.WithError(err).Error("Error listing backup volumes from backup target")
+		return nil // Ignore error to allow status update as well as preventing enqueue
+	}
+	backupStoreBackupVolumes := sets.NewString(res...)
+
+	// Get a list of all the backup volumes that exist as custom resources in the cluster
+	clusterBackupVolumes, err := btc.ds.ListBackupVolumes()
+	if err != nil {
+		log.WithError(err).Error("Error listing backup volumes in the cluster, proceeding with pull into cluster")
+		return err
```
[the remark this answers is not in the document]
jenting: Addressed in #985.

PR #953, opened 2021-06-28
Backing image enhancement

shuo-wu:
longhorn/longhorn#2530


shuo-wu:
Adding error message reporting and retry backoff. Convert to draft for now.

datastore/longhorn.go
```diff
@@ -1301,6 +1318,158 @@ func GetOwnerReferencesForBackingImageManager(backingImageManager *longhorn.Back
...
+
+	for _, itemRO := range list {
+		// Cannot use cached object from lister
+		itemMap[itemRO.Name] = itemRO.DeepCopy()
+	}
+	return itemMap, nil
+}
+
+// ListBackingImageDataSourcesByNode returns object includes all BackingImageDataSource in namespace
+func (s *DataStore) ListBackingImageDataSourcesByNode(nodeName string) (map[string]*longhorn.BackingImageDataSource, error) {
```
joshimoo, line 1437: NIT: I like to add a generic list function that gets passed a selector, see the refactor I done for the `listBackingImageManagers(selector labels.Selector)` as an example. This way both the list Everything an ByNode both go through the same retrieval + copy code flow.

controller/backing_image_data_source_controller.go
```diff
@@ -0,0 +1,855 @@
...
+		}
+
+		// use the last known state, to enqueue, dependent objects
+		pod, ok = deletedState.Obj.(*v1.Pod)
+		if !ok {
+			return false
+		}
+	}
+
+	if pod.Labels[types.GetLonghornLabelComponentKey()] == types.LonghornLabelBackingImageDataSource {
```
joshimoo, line 612: NIT: `return pod.Labels[types.GetLonghornLabelComponentKey()] == types.LonghornLabelBackingImageDataSource`

controller/backing_image_data_source_controller.go
```diff
@@ -0,0 +1,855 @@
...
+	if bids.Status.OwnerID != m.controllerID {
+		close(m.stopCh)
+		logrus.Warnf("Stop monitoring since backing image data source %v owner %v is not the same as monitor current controller %v", m.Name, bids.Status.OwnerID, m.controllerID)
+		return
+	}
+
+	fileInfo, err := m.client.Get()
+	if err != nil {
+		syncErr = errors.Wrapf(err, "failed to get %v info from backing image data source server", m.Name)
+		logrus.Error(syncErr)
```
joshimoo, line 826: NIT: any reason not to use m.log for these?

controller/backing_image_data_source_controller.go
```diff
@@ -0,0 +1,855 @@
...
+	}
+
+	existingBIDS := bids.DeepCopy()
+	bids.Status.CurrentState = types.BackingImageState(fileInfo.State)
+	bids.Status.Size = fileInfo.Size
+	bids.Status.Progress = fileInfo.Progress
+	bids.Status.Checksum = fileInfo.CurrentChecksum
+	bids.Status.Message = fileInfo.Message
+	if !reflect.DeepEqual(bids.Status, existingBIDS.Status) {
+		if _, err := m.ds.UpdateBackingImageDataSourceStatus(bids); err != nil {
```
joshimoo, line 841: NIT: this uses the old monitoring pattern, where the monitor does the status update outside of the reconcile loop.
It's okay for now, just need to remember to refactor it later when we do https://github.com/longhorn/longhorn/issues/2441

joshimoo (commented):
I reviewed all of the code completely except the 3 controllers. Let's do an in person review of the controllers during tomorrows PR review meeting @shuo-wu

joshimoo:
These PR's that add new controllers + crds are always pretty large.
We should consider splitting them into  2 PR's.
PR1: With CRD + Boilerplate Code (datastore, etc) and boilerplate Controller (enqueues, constructor) no sync logic yet.
PR2: actual implementation of the sync logic + monitoring etc.

This way PR1 can go early into the code base with a quick check review, since it's more or less just boilerplate.
Then PR2 can be reviewed in depth or via in person code walk through, since this is where actual code will run and do stuff.

What do you guys think @innobead @shuo-wu ?

util/util.go
```diff
@@ -422,17 +422,25 @@ func GetDiskInfo(directory string) (info *DiskInfo, err error) {
...
 func RetryOnConflictCause(fn func() (interface{}, error)) (obj interface{}, err error) {
-	for i := 0; i < ConflictRetryCounts; i++ {
+	return RetryOnErrorCondition(fn, apierrors.IsConflict)
+}
+
+func RetryOnNotFoundCause(fn func() (interface{}, error)) (obj interface{}, err error) {
+	return RetryOnErrorCondition(fn, apierrors.IsNotFound)
+}
+
+func RetryOnErrorCondition(fn func() (interface{}, error), predicate func(error) bool) (obj interface{}, err error) {
```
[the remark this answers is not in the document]
shuo-wu: No. Will remove the naming.

types/types.go
```diff
@@ -162,6 +164,8 @@ const (
 	instanceManagerImagePrefix = "imi-"
 	shareManagerImagePrefix    = "smi-"
 
+	BackingImageDataSourcePodNamePrefix = "backing-image-data-source-"
```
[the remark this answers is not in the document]
shuo-wu: Sounds good. Will update it.

util/util.go
```diff
@@ -61,8 +61,8 @@ var (
 	cmdTimeout     = time.Minute // one minute by default
 	reservedLabels = []string{"KubernetesStatus", "ranchervm-base-image"}
 
-	ConflictRetryInterval = 20 * time.Millisecond
-	ConflictRetryCounts   = 100
+	APIRetryInterval = 20 * time.Millisecond
```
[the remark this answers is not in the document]
shuo-wu: No. I didn't realize the interval was too short. I will update it to 500ms.

BTW, what do you mean by `at some jitter around the time.sleep`?

engineapi/types.go
```diff
@@ -106,20 +108,17 @@ type Volume struct {
...
-	LastBackupName   string                             `json:"lastBackupName"`
-	LastBackupAt     string                             `json:"lastBackupAt"`
-	DataStored       string                             `json:"dataStored"`
-	Messages         map[backupstore.MessageType]string `json:"messages"`
-	Backups          map[string]*Backup                 `json:"backups"`
-	BackingImageName string                             `json:"backingImageName"`
-	BackingImageURL  string                             `json:"backingImageURL"`
-
-	// Deprecated
-	BaseImage string `json:"baseImage"`
```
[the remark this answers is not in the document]
shuo-wu: I think it's safe to remove it.
1. This field is never been used since we remove the backing image feature in v0.5.0. The backing image info is tracked by the new field `BackingImageName` after re-introducing backing image feature in v1.1.0.
2. Even if somehow the backup meta info json string contains this removed field, this field will be ignored during marshaling.

controller/backing_image_data_source_controller.go
shuo-wu: No. I was just used to print the log via `logrus`. I will replace all `logrus` with `m.log`.

controller/backing_image_data_source_controller.go
shuo-wu: Sure. I will add a `TODO` comment with this link first. If time allows, I will do refactor later.

shuo-wu:
> These PR's that add new controllers + crds are always pretty large.
> We should consider splitting them into 2 PR's.
> PR1: With CRD + Boilerplate Code (datastore, etc) and boilerplate Controller (enqueues, constructor) no sync logic yet.
> PR2: actual implementation of the sync logic + monitoring etc.

It's a nice suggestion. But for this PR, splitting it now will lead to tons of extra modifications and commits adjustments, which will consume lots of time. Besides, there are 2 other PRs highly rely on this PR. I think we can follow this pattern next time. 

BTW, the controller modifications/commits are in the latter part of this PR. We can focus on these commits then.

joshimoo:
@shuo-wu Agree I wasn't suggesting to do this to this PR, makes sense not to touch this one.

engineapi/types.go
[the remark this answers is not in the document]
shuo-wu: Will call it out in the comit comment.

api/backingimage.go
```diff
@@ -79,6 +103,32 @@ func (s *Server) BackingImageCleanup(rw http.ResponseWriter, req *http.Request)
...
+	apiContext.Write(toBackingImageResource(bi, bids, apiContext))
 	return nil
 }
+
+func (s *Server) prepareBackingImageDataSource(bi *longhorn.BackingImage) (*longhorn.BackingImageDataSource, error) {
+	// For deleting backing image, the related data source object may be cleaned up.
+	if bi.DeletionTimestamp != nil {
+		return nil, nil
+	}
+	return s.getBackingImageDataSourceWithRetry(bi.Name)
```
shuo-wu, line 119: Just return nil if the CR can not be found after retry

controller/backing_image_controller.go
```diff
@@ -596,28 +596,121 @@ func (bic *BackingImageController) syncBackingImageFileInfo(bi *longhorn.Backing
...
+		if !isActiveFile && !isRecordedHistoricFile && !isFirstFile {
+			// The timestamp may be removed later based on the HA requirement.
+			bi.Status.DiskLastRefAtMap[diskUUID] = util.Now()
 		}
 	}
+
+	// Determine the final cleanup entries based on HA requirement as well as
+	// the current file state map.
+	// TODO: Make `haBackingImageCount` configure when introducing HA backing image feature
+	haBackingImageCount := 1
```
shuo-wu, line 641: Move the HA check to the cleanup function rather than this timestamp setting function

controller/backing_image_controller.go
```diff
@@ -395,13 +503,43 @@ func (bic *BackingImageController) handleBackingImageManagers(bi *longhorn.Backi
...
+
+		if bids.Status.Size > 0 {
+			if bi.Status.Size == 0 {
+				bi.Status.Size = bids.Status.Size
+				bic.eventRecorder.Eventf(bi, corev1.EventTypeNormal, EventReasonUpdate, "Set size to %v", bi.Status.Size)
+			}
+			if bi.Status.Size != bids.Status.Size {
+				return fmt.Errorf("BUG: found mismatching size %v in disk %v, the size recorded in status is %v", bids.Status.Size, bids.Spec.DiskUUID, bi.Status.Size)
+			}
+		}
```
shuo-wu, line 540: Abstruct this part as a function to avoid duplicate code

controller/backing_image_controller.go
```diff
@@ -329,6 +342,99 @@ func (bic *BackingImageController) cleanupBackingImageManagers(bi *longhorn.Back
...
+	} else if bids.Spec.Started && allFilesFailed {
+		switch bids.Spec.SourceType {
+		case types.BackingImageDataSourceTypeDownload:
+			log.Info("Prepare to re-download backing image via backing image data source since all existing files become unavailable")
+			bids.Spec.Started = false
+		default:
+			log.Warnf("Cannot recover backing image after all existing files becoming unavailable, since the backing image data source with type %v doesn't support restarting", bids.Spec.SourceType)
+		}
+	}
```
shuo-wu, line 400: Use a variable with **an appropriate naming** to indicate the code logic.

joshimoo (commented):
Had a PR review session with @shuo-wu we identified 2 issues, that would need to be fixed before merging. But the rest should be fine.

- Issue01: backup image listing should function even if a backup data source resource is missing (i.e. one missing ds should not stop all backup images from listing)

- Issue02: the way cleanup of a no longer referenced image is handled needs to be changed, so there is no flip/flop on the last ref timestamp. Instead a free floating function should be designed that gets a set of input parameters that are necessary to produce a set of disks/images that are safe to remove. The reason for this is that this function can then be independently unit tested with different scenarios withhout having knowledge of the controller sync loops.

A similiar approach should in the future be used for replica scheduling decisions @c3y1huang

controller/node_controller.go
```diff
@@ -967,3 +951,84 @@ func (nc *NodeController) cleanUpBackingImagesInDisks(node *longhorn.Node) error
 
 	return nil
 }
+
+func BackingImageDiskFileCleanup(node *longhorn.Node, bi *longhorn.BackingImage, bids *longhorn.BackingImageDataSource, waitInterval time.Duration, haRequirement int) {
```
joshimoo, line 955: This is a good step forward with the addition of the unit tests.
In general I would have liked to see a split between evaluation/decision making (which to delete) & modification (operates on the prior set of decisions) then removes them from the resources.

This way Decision logic can be independently tested, but this is good enough :)
Thank you for adding it.

joshimoo (dismissed):
LGTM Just gave it a quick look on the changes.

joshimoo:
@innobead /hold merge @shuo-wu will merge this after he had a chance to do a full test.

shuo-wu:
May need to update vendor longhorn/backing-image-manager before merge.

joshimoo:
Had a review session with @shuo-wu LGTM, we identified some smaller changes.
And scheduled a PR review session for Monday, to follow up and get this merged.

innobead (approved):
In general, LGTM.

PR #963, opened 2021-07-19
Fix replica auto-balance rebuild loop in zones

c3y1huang:
This issue was raised by @kaxing at manual regressions for https://github.com/longhorn/longhorn/issues/587.
- [When there are two equal zones(with the same node numbers), replica-auto-balance will rebuild replica in between zones.](https://github.com/longhorn/longhorn/issues/587#issuecomment-881419481)
- And [one of the replica will rebuilding in between zones.](https://github.com/longhorn/longhorn/issues/587#issuecomment-882194830)

I can reproduce it manually on a cluster with 7 nodes, 3 zones, and 5 replicas. This is caused by the unnecessary hardNodeAffinity by the multiple candidates. This skips the node candidate when the replica adjust count is zero.

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -1933,6 +1933,9 @@ func (vc *VolumeController) getReplicaCountForAutoBalanceNode(v *longhorn.Volume
 	} else {
 		adjustCount = unevenCount
 	}
+	if adjustCount < 0 {
+		adjustCount = 0
+	}
 	log.Debugf("Found %v node available for auto-balance duplicates in %v", adjustCount, nodeExtraRs)
 
 	return adjustCount, nodeExtraRs, err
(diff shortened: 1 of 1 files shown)


c3y1huang:
Regression - [build#366](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/366/testReport/)

shuo-wu:
Can you have a summary of the issue as the context? There are multiple test cases. I don't know which one is related to this issue.

c3y1huang:
> Can you have a summary of the issue as the context? There are multiple test cases. I don't know which one is related to this issue.

Don't think we have a test case covering this one. This was found by @kaxing and I've updated the issue description.

joshimoo (commented):
The whole replica rebalance / data locality scheduling code is pretty convoluted and should probably be moved into the replica scheduler with functions that take all necessary inputs.

That way we could also add unit tests for these important scheduling decisions/scenarios instead of exclusively having to rely on e2e / manual verification.

This should be refactored and addressed as part of longhorn/longhorn#2667
I will leave a note in that ticket as well.

joshimoo (approved):
I don't like the code complexity surrounding this and do want to see a future refactor as part of #2667 but I will approve this under the premise that it has been tested already by QA/e2e test.

And therefore is better then leaving it in a broken state.

shuo-wu (approved):

PR #968, opened 2021-07-30
Remove duplicate engine::info call to extract volume endpoint

joshimoo:
Remove duplicate engine::info call to extract volume endpoint

Since we already require the volume info any way this is an unnecessary
process/engine client socket creation in the monitoring loop. 

In summary one engine monitor loop after these fixes will open 
`f() = 8e + 6*r`  connections the expected connection count for
a volume with 3 replicas is around 312 time-wait connection in a 
minute window.

additional explanation on why this cannot be further improved at
https://github.com/longhorn/longhorn-engine/pull/646

longhorn/longhorn#2818


innobead (approved):
LGTM

PR #970, opened 2021-07-30
Add validation for credential secret

FrankYang0529:
Issue: https://github.com/longhorn/longhorn/issues/811

### DIFF
--- a/datastore/longhorn.go
+++ b/datastore/longhorn.go
@@ -93,7 +93,7 @@ func (s *DataStore) UpdateSetting(setting *longhorn.Setting) (*longhorn.Setting,
 // ValidateSetting checks the given setting value types and condition
 func (s *DataStore) ValidateSetting(name, value string) (err error) {
 	defer func() {
-		err = errors.Wrapf(err, "fail to set settings with invalid %v", name)
+		err = errors.Wrapf(err, "fail to set the setting %v with invalid value %v", name, value)
 	}()
 	sName := types.SettingName(name)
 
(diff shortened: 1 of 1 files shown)


datastore/longhorn.go
```diff
@@ -107,6 +107,23 @@ func (s *DataStore) ValidateSetting(name, value string) (err error) {
 			}
 			return fmt.Errorf("cannot modify BackupTarget since there are existing standby volumes: %v", standbyVolumeNames)
 		}
+	case types.SettingNameBackupTargetCredentialSecret:
+		secret, err := s.kubeClient.CoreV1().Secrets(s.namespace).Get(value, metav1.GetOptions{})
```
shuo-wu, line 111:
```suggestion
secret, err := s.GetSecretRO(s.namespace, value)
```

shuo-wu (commented):
Thanks for your contribution! I think we can add the validation for other fields as well. e.g. `AWS_CERT`

https://github.com/longhorn/longhorn-manager/blob/d8b8bc23ca76da647f93e4e556f7b906ce8664a6/types/types.go#L127-L138

datastore/longhorn.go
```diff
@@ -107,6 +107,23 @@ func (s *DataStore) ValidateSetting(name, value string) (err error) {
...
+		if err != nil {
+			return err
```
PhanLe1010, line 113: return errors.Wrapf(err, "failed to list the secret %v before modifying backup target credential secret setting", value)

PhanLe1010 (changes requested):
We we also change the defer statement to:
```
	defer func() {
		err = errors.Wrapf(err, "fail to set the setting %v with invalid value %v", name, value)
	}()
```

datastore/longhorn.go
FrankYang0529: Fixed it. Thank you.

datastore/longhorn.go
FrankYang0529: Fixed it. Thank you.

shuo-wu:
@innobead I think we need to help backport this to v1.1.3 & 1.2.x

innobead:
> @innobead I think we need to help backport this to v1.1.3 & 1.2.x

YEP, I think we should for bug fixes. cc @jenting

PR #976, opened 2021-08-04
Add restore support (PV creation) support for encrypted volumes

joshimoo:
By passing `secretNamespace, secretName` as part of the PVCreate input.
We can add the necessary secret reference to the PV that longhorn creates for restore/manual created volumes.
These fields are only evaluated if the volume has been marked as `Encrypted` (i.e. v.Spec.Encrypted == true)

NOTE: we renamed the Secure field to Encrypted

longhorn/longhorn#1859


innobead (approved):
LGTM

PR #977, opened 2021-08-06
Fix the rebuilding race condition

PhanLe1010:
We use `engine.Status.CurrentReplicaAddressMap` to know which replica needed to rebuild. Need to persist the `engine.Status.CurrentReplicaAddressMap` to `etcd` before start rebuilding. Otherwise we might start rebuild and
later on `engine.Status.CurrentReplicaAddressMap` might be failed to persisted in `etcd`. As a result, we mark the rebuilding replica as UNKNOWN and remove it while it is in rebuilding process

longhorn/longhorn#2849


innobead (approved):
LGTM

shuo-wu (approved):

PR #978, opened 2021-08-07
CSI and manager part for volume cloning feature

PhanLe1010:
Implement volume cloning logic on Longhorn manager and CSI component

longhorn/longhorn#1815


innobead (approved):
In general, LGTM.

types/data_source.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+}
+
+func IsValidVolumeDataSource(vds VolumeDataSource) bool {
+	split := strings.Split(string(vds), "://")
+	if len(split) != 2 {
+		return false
+	}
+	prefix := split[0]
+	suffix := split[1]
```
shuo-wu, line 46: To avoid duplicate, you can directly use `decodeDataSourceTypeSnapshot` or `decodeDataSourceTypeVolume` based on `prefix` value

client/generated_node_input.go
```diff
@@ -1,79 +0,0 @@
-package client
```
shuo-wu, line 1: It seems that you can not delete this

csi/controller_server.go
```diff
@@ -121,13 +141,30 @@ func (cs *ControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 			return nil, status.Errorf(codes.AlreadyExists, "volume %s size %v differs from requested size %v", existVol.Name, exVolSize, reqVolSizeBytes)
 		}
 
+		if types.VolumeDataSource(existVol.DataSource).IsDataFromAVolume() {
+			if existVol.CloneStatus.State == string(types.VolumeCloneStateFailed) {
+				logrus.Errorf("CreateVolume: cannot creat volume %s: volume failed to clone", existVol.Id)
```
shuo-wu, line 146: typo `creat`

csi/controller_server.go
```diff
@@ -121,13 +141,30 @@ func (cs *ControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 			return nil, status.Errorf(codes.AlreadyExists, "volume %s size %v differs from requested size %v", existVol.Name, exVolSize, reqVolSizeBytes)
 		}
 
+		if types.VolumeDataSource(existVol.DataSource).IsDataFromAVolume() {
```
shuo-wu, line 144: I am thinking if Longhorn CSI needs to validate clone success or not. If the backup restore is not validated, it shouldn't check the volume clone either.

shuo-wu (changes requested):

controller/engine_controller.go
```diff
@@ -887,6 +887,32 @@ func (m *EngineMonitor) refresh(engine *longhorn.Engine) error {
...
+	var snapshotCloneStatusMap map[string]*types.SnapshotCloneStatus
+	if cliAPIVersion >= engineapi.CLIVersionFive {
+		if snapshotCloneStatusMap, err = client.SnapshotCloneStatus(); err != nil {
+			return err
+		}
+	}
+
+	defer func() {
+		engine.Status.CloneStatus = snapshotCloneStatusMap
+		if !reflect.DeepEqual(existingEngine.Status.CloneStatus, engine.Status.CloneStatus) {
```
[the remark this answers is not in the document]
PhanLe1010: Thank you. Fixed

controller/volume_controller.go
```diff
@@ -2375,6 +2403,93 @@ func (vc *VolumeController) checkAndInitVolumeRestore(v *longhorn.Volume) error
...
+			return nil
+		}
+		return err
+	}
+	// Wait for the source volume to be attach
+	if sourceVol.Status.State != types.VolumeStateAttached {
+		return nil
+	}
+	if snapshotName == "" {
+		volumeManager := manager.NewVolumeManager(vc.controllerID, vc.ds)
```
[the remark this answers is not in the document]
PhanLe1010: If we don't use package `manager` here, we would have to repeat the logic of creating a snapshot. How do you think?

controller/volume_controller.go
```diff
@@ -2449,6 +2574,16 @@ func (vc *VolumeController) checkForAutoDetachment(v *longhorn.Volume, e *longho
...
 
+	// While cloning is happening, don't auto-detach target volume or source volume of the cloning
+	if isTargetVolumeOfCloning(v) {
+		return nil
+	}
+	if isSourceVolOfCloning, err := vc.isSourceVolumeOfCloning(v); err != nil {
+		return err
+	} else if isSourceVolOfCloning {
+		return nil
+	}
```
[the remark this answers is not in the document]
PhanLe1010: We will consider it as an enhancement

csi/controller_server.go
PhanLe1010: If we don't validate cloning here, Longhorn CSI controller server might return OK for the second CSI `CreateVolume`  calls  even if the clone failed (e.g., the first call timeout waiting for cloning to finish and the second call immediately return OK). As the result, user get an empty volume and Kubernetes never attempt to clone the volume again. If we validate the cloning state here, Kubernetes retry to clone volume with a backoff limit until the volume is cloned. 


IMO, we should implement this logic for backup restore too. How do you think?

csi/controller_server.go
PhanLe1010: Thank you. Fixed

engineapi/enginesim.go
```diff
@@ -148,27 +148,27 @@ func (e *EngineSimulator) SimulateStopReplica(addr string) error {
 }
 
 func (e *EngineSimulator) SnapshotCreate(name string, labels map[string]string) (string, error) {
-	return "", fmt.Errorf("Not implemented")
```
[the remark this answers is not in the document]
PhanLe1010: I thought error shouldn't start with Capitalized letter?

types/data_source.go
```diff
@@ -0,0 +1,131 @@
+package types
+
+import (
+	"fmt"
+	"github.com/pkg/errors"
+	"strings"
```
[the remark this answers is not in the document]
PhanLe1010: Thank you. Fixed

types/data_source.go
```diff
@@ -0,0 +1,131 @@
...
+)
+
+const (
+	SnapPrefix = "snap"
+	VolPrefix  = "vol"
+)
+
+func NewVolumeDataSource(volumeDataSourceType string, parameters ...string) (dataSource VolumeDataSource, err error) {
```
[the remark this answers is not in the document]
PhanLe1010: Thank you. Fixed

types/data_source.go
```diff
@@ -0,0 +1,131 @@
...
+		err = errors.Wrapf(err, "cannot create new VolumeDataSource")
+	}()
+
+	switch volumeDataSourceType {
+	case VolumeDataSourceTypeVolume:
+		if len(parameters) != 1 {
+			return VolumeDataSource(""), fmt.Errorf("VolumeDataSourceTypeVolume requires 1 parameter but %v were given", len(parameters))
+		}
+		volumeName := parameters[0]
+		dataSource := VolumeDataSource(fmt.Sprintf("%v://%s", VolPrefix, volumeName))
```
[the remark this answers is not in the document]
PhanLe1010: Thank you, fixed

types/data_source.go
PhanLe1010: Thank you. Fixed

types/data_source.go
```diff
@@ -0,0 +1,131 @@
...
+	case VolPrefix:
+		return VolumeDataSourceTypeVolume
+	case SnapPrefix:
+		return VolumeDataSourceTypeSnapshot
+	default:
+		return ""
+	}
+}
+
+func (vds VolumeDataSource) IsDataFromAVolume() bool {
```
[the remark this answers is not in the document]
PhanLe1010: Thank you. Fixed

csi/controller_server.go
shuo-wu: Is there a retry limit for CSI volume creation? Due to some reason, we choose to wait for the restore success in the attachment stage rather than `CreateVolume` but I don't remember the detail. You may need to try to clone a volume that can spend a long time via CSI first.

controller/volume_controller.go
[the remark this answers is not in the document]
shuo-wu: I may prefer to repeat it in controllers.
@yasker Is there a convention for package `manager` usage?

engineapi/enginesim.go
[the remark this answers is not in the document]
shuo-wu: Yes. OK, though this modification is not related to the PR, it is not important. Let's retain this change.

client/generated_node_input.go
shuo-wu: ^^ You may miss this comment

PhanLe1010: I added it back. But I am not sure why the script doesn't recreate this file. Also, not sure why we need this file

csi/controller_server.go
PhanLe1010: Trying with 70GB full of data volume

controller/volume_controller.go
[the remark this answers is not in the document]
yasker: Don't import the `manager` package in controller. The implied relationship is wrong.

We might need some duplicated code here, but we can refactor this part once we add snapshot CRD.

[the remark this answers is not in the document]
PhanLe1010: Thanks Sheng and Shuo. I removed the dependency on `manager` package

controller/volume_controller.go
```diff
@@ -3332,3 +3469,85 @@ func (vc *VolumeController) ReconcileBackupVolumeState(volume *longhorn.Volume)
...
+	if len(es) != 1 {
+		return nil, fmt.Errorf("more than one engine exists")
+	}
+	for _, e = range es {
+		break
+	}
+	if e.Status.CurrentState != types.InstanceStateRunning {
+		return nil, fmt.Errorf("engine is not running")
+	}
```
shuo-wu, line 3531: You can use `vc.ds.PickVolumeCurrentEngine` instead

client/generated_node_input.go
shuo-wu: We can check it later. Let's retain it now.

controller/volume_controller.go
PhanLe1010: I see that it doesn't save the mount of code. That func might return `nil` so we need to check for `nil`

csi/controller_server.go
shuo-wu: I am not sure why we cannot abort a cloning volume but allow deleting a restoring volume. Will the waiting logic inside `VolumeCreate` blocks/fails the CSI volume deletion call?

For the clone retry, YES, we can make it as an enhancement.

controller/volume_controller.go
shuo-wu: It makes the code more readable a little bit. But this is not required. I will mark this as resolved.

csi/controller_server.go
PhanLe1010: It is because `VolumeCreate` haven't returned OK together with the volume ID yet so Kubernetes doesn't know which volume to delete. 

In fact, when Kubernetes wants to abort/delete the cloning volume, if it doesn't know volume ID yet, it will make another `VolumeCreate` call to get the volume ID. As the result, even if user manually clean up volume in Longhorn UI, it is recreated:
```
  //    The CO is responsible for cleaning up volumes it provisioned
  //    that it no longer needs. If the CO is uncertain whether a volume
  //    was provisioned or not when a `CreateVolume` call fails, the CO
  //    MAY call `CreateVolume` again, with the same name, to ensure the
  //    volume exists and to retrieve the volume's `volume_id` (unless
  //    otherwise prohibited by "CreateVolume Errors").
```

PhanLe1010: This doesn't happen with restoring volume because as you mentioned: we return OK for restoring volume immediately on the second call from CSI provisioner. E.g., we don't wait for restoring to finish inside `VolumeCreate` call. This is also what we want to do for volume cloning to avoid the above behavior .

shuo-wu: If the clone wait in `VolumeCreate` will block the deletion, I prefer not to wait for the cloning. "wait" here always means returning error before the clone complete.
For the restore part, I mean we should make the restore behavior consistent with the clone. If we decide to wait for the clone to complete in `VolumeCreate`, we need to add a wait for the restoring as well.

PhanLe1010: Yes, I understand "wait" means return deadline exceeded. In this case, Kubernetes will not have volume ID thus cannot delete. 

Yes, I agree that we shouldn't wait inside VolumeCreate for both clone and backup restore. I already updated the PR to reflect this.

shuo-wu: Can you add a comment explaining why Longhorn won't wait for clone/restore complete? Then this PR is good to go.

PhanLe1010: Yes, I added the comment. Thank you

innobead (approved):
As per the review results, merge on behalf of @shuo-wu .

PR #981, opened 2021-08-11
api: Fix forwarder for https requests

shuo-wu:
The forwarder simply modifying the schema of a HTTPS request to
HTTP makes the URL become invalid. Since it contradicts the
entries "X-Forwarded-Proto" and "X-Forwarded-Host" in the Header.
By removing or modifying these 2 entries, the issue gets resolved.

longhorn/longhorn#2869


innobead (approved):
LGTM

PR #982, opened 2021-08-11
Correct pointer init and copy for backup related CRs

shuo-wu:
1. The controllers need to guarantee there is no nil pointer before reconciling a resource.
2. The deep copy functions look weird. I have no time to verify if it's actually correct. I directly use the simplest way to rewrite them.


innobead (approved):
LGTM

@jenting you can follow up this tomorrow.

PR #983, opened 2021-08-11
Remove auto salvage default BackupTarget CR

jenting:
To prevent uninstall job stuck in clean up BackupTarget CR.


innobead (approved):
LGTM

PR #984, opened 2021-08-13
Update package client

c3y1huang:
Package client out-of-date.

https://github.com/longhorn/longhorn/issues/2530

### DIFF
--- a/client/generated_backup.go
+++ b/client/generated_backup.go
@@ -21,6 +21,8 @@ type Backup struct {
 
 	SnapshotName string `json:"snapshotName,omitempty" yaml:"snapshot_name,omitempty"`
 
+	State string `json:"state,omitempty" yaml:"state,omitempty"`
+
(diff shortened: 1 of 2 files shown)


client/generated_backing_image.go
```diff
@@ -11,7 +11,7 @@ type BackingImage struct {
 
 	DeletionTimestamp string `json:"deletionTimestamp,omitempty" yaml:"deletion_timestamp,omitempty"`
 
-	DiskFileStatusMap map[string]BackingImageDiskFileStatus `json:"diskFileStatusMap,omitempty" yaml:"disk_file_status_map,omitempty"`
```
shuo-wu, line 14: Do not modify this field. Same below.

shuo-wu (changes requested):

client/generated_backing_image.go
c3y1huang: Fixed.

shuo-wu (approved):

PR #985, opened 2021-08-17
Backup refactoring enhancement

jenting:
https://github.com/longhorn/longhorn-manager/pull/952#discussion_r686744828: Updates backup creation state to ETCD
https://github.com/longhorn/longhorn-manager/pull/952#discussion_r686749679: Correct log message
https://github.com/longhorn/longhorn-manager/pull/952#discussion_r686754880: Skip the last modification check and directly inspect the backup volume if there is a backup CR creation/deletion.
https://github.com/longhorn/longhorn-manager/blob/master/upgrade/v111to120/upgrade.go#L34-L35: Address TODO, creates BackupTarget CR inside the setting_controller.
https://github.com/longhorn/longhorn/issues/2870: Remove `spec.FileCleanupRequired`, delete Backup CRs then delete BackupVolume CR
https://github.com/longhorn/longhorn/issues/2888: The volume event log is filtered by volume name at the frontend, so send the event with the *longhorn.Volume object
https://github.com/longhorn/longhorn/issues/2896: Add node_informer+backup_target_informer in setting_controller to enqueue BackupTarget event when node status change _or_ backup target CR be deleted status change
https://github.com/longhorn/longhorn/issues/2910: Unset backup target in setting CR, waits backup_target_controller/backup_volume_controller to removes the BackupVolume/Backup CRs
https://github.com/longhorn/longhorn/issues/2897: Add longhorn engine version > 4 when using flag `--backup-name`

### DIFF
--- a/controller/backup_controller.go
+++ b/controller/backup_controller.go
@@ -237,11 +237,7 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
 		if backupTarget.DeletionTimestamp == nil && backupTarget.Spec.BackupTargetURL != "" &&
 			backupVolume != nil && backupVolume.DeletionTimestamp == nil {
 			// Initialize a backup target client
-			credential, err := bc.ds.GetCredentialFromSecret(backupTarget.Spec.CredentialSecret)
-			if err != nil {
-				return err
-			}
-			backupTargetClient, err := engineapi.NewBackupTargetClient(defaultEngineImage, backupTarget.Spec.BackupTargetURL, credential)
+			backupTargetClient, err := getBackupTargetClient(bc.ds, backupTarget)
 			if err != nil {
 				log.WithError(err).Error("Error init backup target client")
 				return nil // Ignore error to prevent enqueue
@@ -285,11 +281,7 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
 	// Perform backup snapshot to remote backup target
 	if backup.Spec.SnapshotName != "" && backup.Status.State == "" {
 		// Initialize a backup target client
-		credential, err := bc.ds.GetCredentialFromSecret(backupTarget.Spec.CredentialSecret)
-		if err != nil {
-			return err
-		}
-		backupTargetClient, err := engineapi.NewBackupTargetClient(defaultEngineImage, backupTarget.Spec.BackupTargetURL, credential)
+		backupTargetClient, err := getBackupTargetClient(bc.ds, backupTarget)
 		if err != nil {
 			log.WithError(err).Error("Error init backup target client")
 			return nil // Ignore error to prevent enqueue
@@ -319,11 +311,7 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
 	}
 
 	// Initialize a backup target client
-	credential, err := bc.ds.GetCredentialFromSecret(backupTarget.Spec.CredentialSecret)
-	if err != nil {
-		return err
-	}
-	backupTargetClient, err := engineapi.NewBackupTargetClient(defaultEngineImage, backupTarget.Spec.BackupTargetURL, credential)
+	backupTargetClient, err := getBackupTargetClient(bc.ds, backupTarget)
 	if err != nil {
 		log.WithError(err).Error("Error init a backup target client")
 		return nil // Ignore error to prevent enqueue
--- a/controller/backup_controller.go
+++ b/controller/backup_controller.go
@@ -226,13 +226,6 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
 		return err
 	}
 
-	if backup.Status.LastSyncedAt == nil {
-		backup.Status.LastSyncedAt = &metav1.Time{Time: time.Time{}}
-		if backup, err = bc.ds.UpdateBackupStatus(backup); err != nil {
-			return err
-		}
-	}
-
 	// Examine DeletionTimestamp to determine if object is under deletion
 	if !backup.DeletionTimestamp.IsZero() {
 		backupVolume, err := bc.ds.GetBackupVolume(backupVolumeName)
@@ -262,7 +255,7 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
 
 		// Request backup_volume_controller to reconcile BackupVolume immediately if it's the last backup
 		if backupVolume != nil && backupVolume.Status.LastBackupName == backup.Name {
-			backupVolume.Spec.SyncRequestedAt = &metav1.Time{Time: time.Now().Add(time.Second).UTC()}
+			backupVolume.Spec.SyncRequestedAt = metav1.Time{Time: time.Now().UTC()}
 			if _, err = bc.ds.UpdateBackupVolume(backupVolume); err != nil && !apierrors.IsConflict(errors.Cause(err)) {
 				log.WithError(err).Errorf("Error updating backup volume %s spec", backupVolumeName)
 				// Do not return err to enqueue since backup_controller is responsible to
@@ -273,7 +266,7 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
 		return bc.ds.RemoveFinalizerForBackup(backup)
 	}
 
-	syncTime := &metav1.Time{Time: time.Now().UTC()}
+	syncTime := metav1.Time{Time: time.Now().UTC()}
 	existingBackup := backup.DeepCopy()
 	defer func() {
 		if err != nil {
@@ -320,7 +313,7 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
 
 	// The backup config had synced
 	if !backup.Status.LastSyncedAt.IsZero() &&
-		backup.Status.LastSyncedAt.Time.After(backup.Spec.SyncRequestedAt.Time) {
+		!backup.Spec.SyncRequestedAt.After(backup.Status.LastSyncedAt.Time) {
 		return nil
 	}
 
@@ -517,7 +510,7 @@ func (bc *BackupController) backupCreation(log logrus.FieldLogger, engineClient 
 				// Request backup_volume_controller to reconcile BackupVolume immediately.
 				backupVolume, err := bc.ds.GetBackupVolume(volumeName)
 				if err == nil {
-					backupVolume.Spec.SyncRequestedAt = &metav1.Time{Time: time.Now().Add(time.Second).UTC()}
+					backupVolume.Spec.SyncRequestedAt = metav1.Time{Time: time.Now().UTC()}
 					if _, err = bc.ds.UpdateBackupVolume(backupVolume); err != nil && !apierrors.IsConflict(errors.Cause(err)) {
 						log.WithError(err).Errorf("Error updating backup volume %s spec", volumeName)
 					}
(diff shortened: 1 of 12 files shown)


innobead:
@jenting need to resolve the conflicts.

jenting:
On hold, find an issue when using NFS as backup target (it would happens with S3) + v1.2.0-preview1 engine. After backup deletion and create a new backup would failed.

jenting:
> On hold, find an issue when using NFS as backup target (it would happens with S3) + v1.2.0-preview1 engine. After backup deletion and create a new backup would failed.

Sometimes, list volumes and their backups failed when the backup target is NFS server.
I'll create another issue to tackle it.

controller/backup_controller.go
```diff
@@ -448,39 +447,41 @@ func (bc *BackupController) backupCreation(log logrus.FieldLogger, engineClient
 		biChecksum = bi.Status.Checksum
 	}
 
-	backup.Status.State = types.BackupStatePending
-	logEvent(nil)
+	backup.Status.State = types.BackupStateInProgress
```
shuo-wu, line 432: After removing state `pending`, how can this controller avoid duplicate backup calls?

controller/backup_target_controller.go
```diff
@@ -174,6 +175,24 @@ func getLoggerForBackupTarget(logger logrus.FieldLogger, backupTarget *longhorn.
 	)
 }
 
+func getBackupTargetClient(ds *datastore.DataStore, defaultEngineImage string, backupTarget *longhorn.BackupTarget) (*engineapi.BackupTargetClient, error) {
```
shuo-wu, line 178: This function doesn't need input `defaultEngineImage`. This setting won't change frequently. You can directly fetch this info via `ds`.

engineapi/backups.go
```diff
@@ -23,14 +23,6 @@ type BackupTargetClient struct {
...
 // NewBackupTargetClient returns the backup target client
 func NewBackupTargetClient(defaultEngineImage, url string, credential map[string]string) (*BackupTargetClient, error) {
-	backupType, err := util.CheckBackupType(url)
-	if err != nil {
-		return nil, err
-	}
-	if backupType == types.BackupStoreTypeS3 && credential == nil {
-		return nil, fmt.Errorf("Could not backup for %s without credential secret", types.BackupStoreTypeS3)
-	}
-
```
shuo-wu, line 33: This function won't return an error anymore.

controller/backup_controller.go
```diff
@@ -253,7 +253,7 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
 			}
 
 			if backupVolume.Status.LastBackupName == backup.Name {
-				backupVolume.Spec.SyncRequestedAt = time.Now().UTC()
```
shuo-wu: Please coalesce this commit to the previous one.

shuo-wu (changes requested):
I have no idea why you decide to revert the cascading deletion at the end of the PR. Can you explain why you choose not to use the cascading deletion?

Besides, if this is not required, please directly remove the original commit. Typically I will review PRs by commits.

shuo-wu:
BTW, please add the corresponding ticket number in each commit.

controller/setting_controller.go
```diff
@@ -97,7 +96,6 @@ func NewSettingController(
 	scheme *runtime.Scheme,
 	settingInformer lhinformers.SettingInformer,
 	nodeInformer lhinformers.NodeInformer,
-	backupTargetInfomer lhinformers.BackupTargetInformer,
```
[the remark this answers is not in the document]
jenting: It would cause the uninstall job would succeed if we auto salvage the default BackupTarget CR back.

controller/backup_volume_controller.go
```diff
@@ -290,7 +290,11 @@ func (bvc *BackupVolumeController) reconcile(backupVolumeName string) (err error
 
 	clustersSet := sets.NewString()
 	for _, b := range clusterBackups {
-		clustersSet.Insert(b.Name)
+		// Skip the Backup CR which is created from local cluster
+		// that have not synced from remote backup target yet
```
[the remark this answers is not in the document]
jenting: Yes, otherwise the in progress Backup CR would be deleted.

controller/uninstall_controller.go
```diff
@@ -382,6 +382,46 @@ func (c *UninstallController) deleteCRDs() (bool, error) {
...
+		return true, c.deleteBackupVolumes(backupVolumes)
+	}
+
+	// Delete BackupTargets before deletes engine images
+	if backupTargets, err := c.ds.ListBackupTargets(); err != nil {
+		return true, err
+	} else if len(backupTargets) > 0 {
+		c.logger.Infof("Found %d backuptargets remaining", len(backupTargets))
+		return true, c.deleteBackupTargets(backupTargets)
+	}
```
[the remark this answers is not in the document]
jenting: If deletes BackupTarget CR first, it would cause backup_volume_controller and backup_controller can't go into deletion code because we check the default BackupTarget CR at these two controllers first.

controller/backup_controller.go
jenting: If the state is not empty, we won't call engineapi.snapshotCreate because we think the backup creation called once already.

manager/engine.go
```diff
@@ -161,17 +162,35 @@ func (m *VolumeManager) BackupSnapshot(backupName, volumeName, snapshotName stri
...
+		bv := &longhorn.BackupVolume{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: volumeName,
+			},
+		}
+		backupVolume, err = m.ds.CreateBackupVolume(bv)
+		if err != nil && !apierrors.IsAlreadyExists(err) {
+			return err
+		}
+	}
```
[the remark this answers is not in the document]
jenting: There might be a case that if the BackupVolume created at here, and the backup volume config haven't be created at the remote backup target.
At the moment, the backup_target_controller think this BackupVolume CR must be deleted since it's only in the remote backup target but only within in-cluster.
If the cascading deletion also implemented, the corresponding Backup CRs will be deleted (also the remote backup config)

jenting:
> I have no idea why you decide to revert the cascading deletion at the end of the PR. Can you explain why you choose not to use the cascading deletion?
> 
> 
> 
> Besides, if this is not required, please directly remove the original commit. Typically I will review PRs by commits.
> 
> 

After implementing cascading deletion, the BackupVolme CR must created when HTTP API calls snapashotBackup endpoint.
At the moment, since the backup still in progress, the backup volume config and backup config doesn't exist in the remote backup.
When the backup_target_controller up, it found the BackupVolume CR must be deleted since it's not in remote backup target, also the Backup CR will be deleted.

Therefore, the backup creation CR will be deleted in the end.

controller/backup_controller.go
```diff
@@ -228,44 +228,41 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
...
-		credential, err := bc.ds.GetCredentialFromSecret(backupTarget.Spec.CredentialSecret)
-		if err != nil {
-			return err
-		}
-		backupTargetClient, err := engineapi.NewBackupTargetClient(defaultEngineImage, backupTarget.Spec.BackupTargetURL, credential)
-		if err != nil {
-			log.WithError(err).Error("Error init backup target client")
-			return nil // Ignore error to prevent enqueue
-		}
+		if backupTarget.Spec.BackupTargetURL != "" {
```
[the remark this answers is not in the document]
jenting: I'd like to remove finalizer if the `backupTarget.Spec.BackupTargetURL == ""`

shuo-wu:
> After implementing cascading deletion, the BackupVolme CR must created when HTTP API calls snapashotBackup endpoint.
> At the moment, since the backup still in progress, the backup volume config and backup config doesn't exist in the remote backup.
> When the backup_target_controller up, it found the BackupVolume CR must be deleted since it's not in remote backup target, also the Backup CR will be deleted.
> 
> Therefore, the backup creation CR will be deleted in the end.

Got it. Then please directly remove the commit that introduces the cascading deletion.

controller/setting_controller.go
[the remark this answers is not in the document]
shuo-wu: Hmmm. There should be a similar issue for InstanceManager CRs. How does Longhorn resolve it? 

Or can we directly ask the setting controller to remove the BackupTarget CR if the setting value being empty?

manager/engine.go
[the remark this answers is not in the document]
shuo-wu: Since you decide to remove the cascading deletion, the BackupVolume CR creation always happens in the controllers. You can ignore this comment.

controller/setting_controller.go
```diff
@@ -668,6 +658,15 @@ func (bst *BackupStoreTimer) Start() {
...
 	wait.PollUntil(bst.pollInterval, func() (done bool, err error) {
+		responsibleNodeID, err := getResponsibleNodeID(bst.ds)
+		if err != nil {
+			log.WithError(err).Warn("Failed to select node for backup store timer, will try again next poll interval")
+			return false, nil
+		}
+		if responsibleNodeID != bst.controllerID {
+			return false, nil
+		}
+
```
shuo-wu, line 669: Why moving the code into this timer function?

shuo-wu (changes requested):

controller/setting_controller.go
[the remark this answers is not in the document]
jenting: I choose 2.

controller/uninstall_controller.go
```diff
@@ -382,6 +382,48 @@ func (c *UninstallController) deleteCRDs() (bool, error) {
...
+
+	// Delete BackupTarget CRs after the BackupVolume/Backup CR be clean up
+	if backupTargets, err := c.ds.ListBackupTargets(); err != nil {
+		return true, err
+	} else if len(backupTargets) > 0 {
+		c.logger.Infof("Found %d backuptargets remaining", len(backupTargets))
+		return true, c.deleteBackupTargets(backupTargets)
+	}
+
```
shuo-wu, line 426: We don't need to delete and wait for BackupTarget CR. It will be gone with the CRD removal.

shuo-wu (changes requested):
1. Why do you still decide to remote the backup target informer from setting_controller (in the 1st commit)?
2. Please rename commit title `Add engine version check`. This title is confusing.
3. Rename commit title `Send backup creation event to *longhorn.Volume object` or split this into multiple commits. The essentital of this commit is totally irrelevant to the current title.
4. Add check `backupTarget == nil` for other controllers. And need to handle ErrorIsNotFound when using `GetBackupTargetRO`

controller/backup_controller.go
```diff
@@ -235,7 +235,7 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
 			return err
 		}
 
-		if backupTarget.DeletionTimestamp == nil && backupTarget.Spec.BackupTargetURL != "" &&
```
shuo-wu: Please retain the check `backupTarget.DeletionTimestamp == nil`. This prevents the remote files from being cleaned up if users delete the CR by accident.

shuo-wu (changes requested):
Others LGTM

controller/backup_controller.go
jenting: But, we'll auto salvage the default BackupTarget CR, isn't?

shuo-wu: OK. When this line is executed, the CR won't be `nil`.

jenting: Also, we have check the BackupTarget CR is not found and return nil before this line.
And since we remove the finalized of BackupTarget CR, so the deletion time stamp always nil from my understanding.

shuo-wu:
Please resolve the conflicts. Then it's good to go.

shuo-wu (approved):

PR #639, opened 2020-08-03
Set LeaderDurationSeconds on LeaderElectionRecord

ttpcodes:
This PR patches the `client-go` dependency in order to fix a bug in which the `LeaderDurationSeconds` is not properly set on the `LeaderElectionRecord` created in the `release()` function. This fixes the bug in longhorn/longhorn#1379 in which the `LeaderDurationSeconds` field is set to 0, leading to a failure to release the upgrade lock and delaying the upgrade process while the other Longhorn Managers wait for the lock to expire.

This PR contains an additional fix to the `LeaseSpecToLeaderElectionRecord()` function, which prevents a panic resulting from a nil pointer occurring when `LeaderDurationSeconds` is set, which is necessary for the above patch to work properly in the current version of the Kubernetes API we are using.

If the `vendor/` is overwritten, these patches can be applied again using `git am --signoff <name of patch file>`, with `vendor-add-duration-seconds-to-election-record.patch` containing the first fix and `vendor-fix-nil-pointer-for-Leader-Election-Record.patch` containing the second fix.

These patches should eventually be removed from the repository once the issues have been fixed upstream. The first patch does not have a timeline for removal, as upstream has yet to merge the PRs necessary to fix this issue:
- kubernetes/kubernetes#88192
- kubernetes/kubernetes#91942
- kubernetes/kubernetes#91966

The second patch can be removed once we upgrade to v1.17 of the Kubernetes API which contains the fix for the nil pointer issue. The code for this fix is contained in this commit: kubernetes/kubernetes@7907b29551c7ef87bbe398ac02836b4c87246d3d

PR #988, opened 2021-08-18
Longhorn v1.2.0: Bump Kubernetes, API version group, CSI component's dependency version

PhanLe1010:
- [x] Bump Kubernetes, client-go to v1.18
- [x] API version group
- [x] CSI component's dependency version 

longhorn/longhorn#2757


csi/util.go
```diff
@@ -25,16 +27,16 @@ const (
...
-// NewForcedParamsOsExec creates a osExecutor that allows for adding additional params to later occurring Run calls
-func NewForcedParamsOsExec(cmdParamMapping map[string]string) mount.Exec {
+// NewForcedParamsExec creates a osExecutor that allows for adding additional params to later occurring Run calls
+func NewForcedParamsExec(cmdParamMapping map[string]string) utilexec.Interface {
 	return &forcedParamsOsExec{
-		osExec:          mount.NewOSExec(),
+		exec:            utilexec.New(),
 		cmdParamMapping: cmdParamMapping,
 	}
 }
```
PhanLe1010, line 36: Can you double-check this change @joshimoo ? Kubenetes v1.18 moves the mount package around and I have to make this change

innobead:
This will be after RC1.

PhanLe1010:
RKE `v1.18.20`
- [ ] Running e2e test
- [ ] Test upgrade from Longhorn `v1.1.2` to `v1.2.0`

K3s  `v1.22.1-rc1+k3s1`
- [ ] Running e2e test
- [ ] Test upgrade from Longhorn `v1.1.2` to `v1.2.0`

deploy/longhorn-images.txt
```diff
@@ -1,8 +1,8 @@
-longhornio/csi-attacher:v2.2.1-lh2
-longhornio/csi-node-driver-registrar:v1.2.0-lh1
-longhornio/csi-provisioner:v1.6.0-lh2
-longhornio/csi-resizer:v0.5.1-lh2
-longhornio/csi-snapshotter:v2.1.1-lh2
+k8s.gcr.io/sig-storage/csi-attacher:v3.2.1
+k8s.gcr.io/sig-storage/csi-provisioner:v2.1.2
+k8s.gcr.io/sig-storage/csi-resizer:v1.2.0
+k8s.gcr.io/sig-storage/csi-snapshotter:v3.0.3
+k8s.gcr.io/sig-storage/csi-node-driver-registrar:v2.3.0
```
innobead, line 5: Why we don't use the latest releases of attacher (3.3.0), provisioner (v2.2.2), resizer (1.3.0) compatible with 1.18?

innobead (dismissed):
In general, LGTM. Just one question to resolve.

Also, remember:
- update the example options of CSI images in longhorn-manager/deploy/install/02-components/04-driver.yaml
- have a PR update for longhorn/longhorn chart

Need @shuo-wu or @joshimoo to review before merging.

deploy/longhorn-images.txt
innobead: Using the upstream images is good, but this also possibly brings us another challenge when dealing with vulnerability issues the future. However, these images seem all based on `distroless` image, so that should be fine.

cc @meldafrawi

PhanLe1010: > Why we don't use the latest releases of attacher (3.3.0), provisioner (v2.2.2), resizer (1.3.0) compatible with 1.18?

So we import Kubernetes v1.18 (since many of our users are still using it). This leads to the decision of choosing CSI Spec v1.2 since Kubernetes v1.18 uses CSI Spec v1.2. As a result, all CSI components need to support min CSI Version v1.2

We cannot use provisioner (v2.2.2), resizer (1.3.0). But yeah, we can use attacher (3.3.0). I was looking at [the documentation](https://kubernetes-csi.github.io/docs/external-attacher.html) so I didn't see the latest attached v3.3.o

PhanLe1010: Just take another look at [attacher v3.3.0](https://github.com/kubernetes-csi/external-attacher/releases). Actually, it is too new and has recommended Kubernetes version of 1.22. Should we still want to use it?

innobead: Thanks, but it's minor for a patch update for now. All good.

innobead: Let's just keep the version u verified already.

innobead:
> RKE `v1.18.20`
> 
> * [x]  e2e core test
> * [x]  Test upgrade from Longhorn `v1.1.2` to `v1.2.0`
> 
> K3s `v1.22.1-rc1+k3s1`
> 
> * [x]  e2e core test
> * [x]  ~Test upgrade from Longhorn `v1.1.2` to `v1.2.0`~ Not applicable since Longhorn `v1.1.2` doesn't run on Kubernetes v1.22

cc @khushboo-rancher @meldafrawi @kaxing

csi/util.go
joshimoo: Looks good, god job on the change. With them moving this stuff into a dedicated util package we can potentially replace the nsenter exec after 1.2 and push some changes to the util package.

csi/util.go
```diff
@@ -25,16 +27,16 @@ const (
 	defaultStaleReplicaTimeout = 2880
 )
 
-// NewForcedParamsOsExec creates a osExecutor that allows for adding additional params to later occurring Run calls
-func NewForcedParamsOsExec(cmdParamMapping map[string]string) mount.Exec {
+// NewForcedParamsExec creates a osExecutor that allows for adding additional params to later occurring Run calls
+func NewForcedParamsExec(cmdParamMapping map[string]string) utilexec.Interface {
 	return &forcedParamsOsExec{
-		osExec:          mount.NewOSExec(),
+		exec:            utilexec.New(),
```
joshimoo, line 33: Can you also make the exec/mount changes on the share-manager repository, since it give you a chance to go through the build process there.

joshimoo (approved):
LGTM, @PhanLe1010 can you check if we still need the leader election patch on the new kubernetes version ref: https://github.com/longhorn/longhorn-manager/pull/639

PhanLe1010:
> In general, LGTM. Just one question to resolve.
> 
> Also, remember:
> 
> * update the example options of CSI images in longhorn-manager/deploy/install/02-components/04-driver.yaml
> * have a PR update for longhorn/longhorn chart
> 
> Need @shuo-wu or @joshimoo to review before merging.

Done

csi/util.go
PhanLe1010: Sure. Added to my todo list

joshimoo:
The update for the share-manager was missed, so I am updating the share-manager kubernetes dependencies as part of
https://github.com/longhorn/longhorn/issues/2991

Which will go into v1.2.1

PR #990, opened 2021-08-19
Add BackupTarget unavailable reason and HTTP endpoint

jenting:
#### Issues

https://github.com/longhorn/longhorn/issues/2887

### DIFF
--- a/api/model.go
+++ b/api/model.go
@@ -73,6 +73,11 @@ type Snapshot struct {
 	types.Snapshot
 }
 
+type BackupTarget struct {
+	client.Resource
+	engineapi.BackupTarget
+}
+
 type BackupVolume struct {
 	client.Resource
 	engineapi.BackupVolume
@@ -343,6 +348,7 @@ func NewSchema() *client.Schemas {
 	schemas.AddType("attachInput", AttachInput{})
 	schemas.AddType("detachInput", DetachInput{})
 	schemas.AddType("snapshotInput", SnapshotInput{})
+	schemas.AddType("backupTarget", BackupTarget{})
 	schemas.AddType("backup", Backup{})
 	schemas.AddType("backupInput", BackupInput{})
 	schemas.AddType("backupStatus", BackupStatus{})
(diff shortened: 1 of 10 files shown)


api/model.go
```diff
@@ -1072,6 +1078,28 @@ func toSnapshotCollection(ss map[string]*types.Snapshot) *client.GenericCollecti
...
+			Id:    bt.Name,
+			Type:  "backupTarget",
+			Links: map[string]string{},
+		},
+		BackupTarget: engineapi.BackupTarget{
+			BackupTargetURL:  bt.Spec.BackupTargetURL,
+			CredentialSecret: bt.Spec.CredentialSecret,
+			PollInterval:     bt.Spec.PollInterval.Duration.String(),
+			Available:        bt.Status.Available,
+			Message:          bt.Status.Conditions[types.BackupTargetConditionTypeUnavailable].Message,
```
innobead, line 1097: If the condition unavailable, have a default message to avoid null pointer exception.

manager/engine.go
```diff
@@ -225,6 +225,23 @@ func (m *VolumeManager) GetEngineClient(volumeName string) (client engineapi.Eng
 	})
 }
 
+func (m *VolumeManager) ListBackupTargetsSorted() ([]*longhorn.BackupTarget, error) {
+	backupTargetMap, err := m.ds.ListBackupTargets()
+	if err != nil {
+		return []*longhorn.BackupTarget{}, err
+	}
+
+	backupTargets := make([]*longhorn.BackupTarget, len(backupTargetMap))
```
innobead, line 234: Suggest to move after the below if-err block, so just use it when needed.

innobead (changes requested):
LGTM, just few comments needed to resolve.

api/model.go
jenting: I think NPE won't happens https://play.golang.org/p/0gO7qqleBxo

innobead (approved):

PR #992, opened 2021-08-19
2867-rwx-encryption-support

joshimoo:
Add support for encrypted volumes, this works by retrieving the secret for the PV as part of the share-manager controller, the passphrase is then passed to the share-manager pod as an environment variable. From a security standpoint this isn't great but we currently have no way of securly transmitting that phasphrase, this would require a secure communication channel (grpc) to transmit the information at runtime.
REF: longhorn/longhorn#2867 longhorn/longhorn#1859

This also added new optional params "fs", "mount" to allow specifying a custom filesystem for the share-manager RWO volume. This has been requested in longhorn/longhorn#2814 These are optional and when not specified we use the same behavior as before. The fs/mount options are the ones specified as part of the storage class, since that information is persisted in the PV context. (So we treat this RWX volume the same way as an RWO volume) since it effectivly is still an RWO volume.

After discussing with @yasker I removed the forced `sync` mount option for the nfs client mount, instead it will rely on the host defaults which for most oses is async, since sync is prohibitly expensive. This addresses the performance difference between external provisioner based approach and the native longhorn RWX approach. REF longhorn/longhorn#2584 longhorn/longhorn#2549

I also added a hidden/undocumented nfsOptions parameter to the storage classes, that can be used to configure the nfs mount options, this is mostly for testing different settings withhout requiring a rebuild. This also allows us to update the share-manager independently of the csi-plugin in case we want to add support for UDP, etc. REF longhorn/longhorn#2630

### DIFF
--- a/deploy/install/02-components/01-manager.yaml
+++ b/deploy/install/02-components/01-manager.yaml
@@ -29,7 +29,7 @@ spec:
         - --instance-manager-image
         - longhornio/longhorn-instance-manager:v1_20210731
         - --share-manager-image
-        - longhornio/longhorn-share-manager:v_20210820
+        - longhornio/longhorn-share-manager:v1_20210820
         - --backing-image-manager-image
         - longhornio/backing-image-manager:v2_20210809
         - --manager-image
--- a/deploy/install/02-components/01-manager.yaml
+++ b/deploy/install/02-components/01-manager.yaml
@@ -29,7 +29,7 @@ spec:
         - --instance-manager-image
         - longhornio/longhorn-instance-manager:v1_20210731
         - --share-manager-image
-        - joshimoo/longhorn-share-manager:v_20210819
+        - longhornio/longhorn-share-manager:v_20210820
         - --backing-image-manager-image
         - longhornio/backing-image-manager:v2_20210809
         - --manager-image
--- a/deploy/install/02-components/01-manager.yaml
+++ b/deploy/install/02-components/01-manager.yaml
@@ -29,7 +29,7 @@ spec:
         - --instance-manager-image
         - longhornio/longhorn-instance-manager:v1_20210731
         - --share-manager-image
-        - longhornio/longhorn-share-manager:v1_20210416
+        - joshimoo/longhorn-share-manager:v_20210819
         - --backing-image-manager-image
         - longhornio/backing-image-manager:v2_20210809
         - --manager-image
--- a/deploy/longhorn-images.txt
+++ b/deploy/longhorn-images.txt
@@ -7,5 +7,5 @@ longhornio/backing-image-manager:v2_20210809
 longhornio/longhorn-engine:master
 longhornio/longhorn-instance-manager:v1_20210731
 longhornio/longhorn-manager:master
-longhornio/longhorn-share-manager:v1_20210416
+joshimoo/longhorn-share-manager:v_20210819
 longhornio/longhorn-ui:master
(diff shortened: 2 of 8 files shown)


deploy/install/02-components/01-manager.yaml
```diff
@@ -29,7 +29,7 @@ spec:
         - --instance-manager-image
         - longhornio/longhorn-instance-manager:v1_20210731
         - --share-manager-image
-        - longhornio/longhorn-share-manager:v1_20210416
+        - joshimoo/longhorn-share-manager:v_20210819
```
innobead, line 32:
```suggestion
        - longhornio/longhorn-share-manager:v_20210820
```

deploy/longhorn-images.txt
```diff
@@ -7,5 +7,5 @@ longhornio/backing-image-manager:v2_20210809
 longhornio/longhorn-engine:master
 longhornio/longhorn-instance-manager:v1_20210731
 longhornio/longhorn-manager:master
-longhornio/longhorn-share-manager:v1_20210416
+joshimoo/longhorn-share-manager:v_20210819
```
innobead, line 10:
```suggestion
longhornio/longhorn-share-manager:v_20210820
```

deploy/release-images.txt
```diff
@@ -7,5 +7,5 @@ longhornio/backing-image-manager:v2_20210809
 longhornio/longhorn-engine:master
 longhornio/longhorn-instance-manager:v1_20210731
 longhornio/longhorn-manager:master
-longhornio/longhorn-share-manager:v1_20210416
+joshimoo/longhorn-share-manager:v_20210819
```
innobead, line 10:
```suggestion
longhornio/longhorn-share-manager:v_20210820
```

innobead (approved):
LGTM, just need to update using the latest share manager image.

deploy/install/02-components/01-manager.yaml
```diff
@@ -29,7 +29,7 @@ spec:
         - --instance-manager-image
         - longhornio/longhorn-instance-manager:v1_20210731
         - --share-manager-image
-        - longhornio/longhorn-share-manager:v1_20210416
+        - longhornio/longhorn-share-manager:v_20210820
```
innobead, line 32:
```suggestion
        - longhornio/longhorn-share-manager:v1_20210820
```

deploy/longhorn-images.txt
```diff
@@ -7,5 +7,5 @@ longhornio/backing-image-manager:v2_20210809
 longhornio/longhorn-engine:master
 longhornio/longhorn-instance-manager:v1_20210731
 longhornio/longhorn-manager:master
-longhornio/longhorn-share-manager:v1_20210416
+longhornio/longhorn-share-manager:v_20210820
```
innobead, line 10:
```suggestion
longhornio/longhorn-share-manager:v1_20210820
```

deploy/release-images.txt
```diff
@@ -7,5 +7,5 @@ longhornio/backing-image-manager:v2_20210809
 longhornio/longhorn-engine:master
 longhornio/longhorn-instance-manager:v1_20210731
 longhornio/longhorn-manager:master
-longhornio/longhorn-share-manager:v1_20210416
+longhornio/longhorn-share-manager:v_20210820
```
innobead, line 10:
```suggestion
longhornio/longhorn-share-manager:v1_20210820
```

innobead (approved):

PR #996, opened 2021-08-21
Use support bundle kit for collecting longhorn resources

shangma:
[longhorn/longhorn#2759](https://github.com/longhorn/longhorn/issues/2759)


app/uninstall.go
```diff
@@ -15,6 +12,9 @@ import (
 	"github.com/longhorn/longhorn-manager/datastore"
 	lhclientset "github.com/longhorn/longhorn-manager/k8s/pkg/client/clientset/versioned"
 	lhinformers "github.com/longhorn/longhorn-manager/k8s/pkg/client/informers/externalversions"
+	"github.com/pkg/errors"
```
yasker, line 15: Don't move those to longhorn specific section.

controller/controller_manager.go
```diff
@@ -106,8 +107,12 @@ func StartControllers(logger logrus.FieldLogger, stopCh chan struct{}, controlle
 		engineImageInformer, nodeInformer, settingInformer,
 		imInformer, shareManagerInformer,
 		backingImageInformer, backingImageManagerInformer, backingImageDataSourceInformer,
+<<<<<<< HEAD
 		backupTargetInformer, backupVolumeInformer, backupInformer,
 		recurringJobInformer,
+=======
+		backupTargetInformer, backupVolumeInformer, backupInformer, supportBundleInformer,
+>>>>>>> update informer
```
yasker: ??? How does this passed CI?

controller/controller_manager.go
```diff
@@ -205,7 +213,11 @@ func StartControllers(logger logrus.FieldLogger, stopCh chan struct{}, controlle
 	go btc.Run(Workers, stopCh)
 	go bvc.Run(Workers, stopCh)
 	go bc.Run(Workers, stopCh)
+<<<<<<< HEAD
 	go rjc.Run(Workers, stopCh)
+=======
+	go sbc.Run(Workers, stopCh)
+>>>>>>> update informer
```
yasker: ???

controller/controller_manager.go
```diff
@@ -150,6 +155,9 @@ func StartControllers(logger logrus.FieldLogger, stopCh chan struct{}, controlle
 	bc := NewBackupController(logger, ds, scheme,
 		backupInformer,
 		kubeClient, controllerID, namespace)
+	sbc := NewSupportBundleController(logger, ds, scheme,
```
yasker, line 154: Add to the end of the list.

yasker (changes requested):
Haven't finished the review yet. But
1. random merge code in the PR.
2. Wrong Signed-off-by email.
3. Multiple lines of codes are not in the correct commits.

controller/controller_manager.go
yasker: OK saw the later commit removes it. Need to merge them and put the code in the corresponding commit.

yasker:
also need to add issue reference in the PR’s comment.

app/uninstall.go
shangma: Fixed. @yasker

controller/controller_manager.go
shangma: all cleaned up @yasker

controller/controller_manager.go
shangma: updated. @yasker

shangma:
> also need to add issue reference in the PR’s comment.

Updated. @yasker

types/resource.go
```diff
@@ -776,3 +776,15 @@ type RecurringJobSpec struct {
...
 }
+
+type SupportBundleSpec struct {
+	IssueURL    string `json:"issueURL"`
+	Description string `json:"description"`
+}
+
+type SupportBundleStatus struct {
+	State      string      `json:"state,omitempty"`
+	Filename   string      `json:"filename,omitempty"`
```
yasker, line 787: Need to use CamelCase for Go variable name and camelCase for json.
`FileName: json fileName`
`FileSize: json fileSize`

types/setting.go
```diff
@@ -195,6 +199,28 @@ var (
 		SettingNameBackingImageCleanupWaitInterval:              SettingDefinitionBackingImageCleanupWaitInterval,
 		SettingNameGuaranteedEngineManagerCPU:                   SettingDefinitionGuaranteedEngineManagerCPU,
 		SettingNameGuaranteedReplicaManagerCPU:                  SettingDefinitionGuaranteedReplicaManagerCPU,
+		SettingNameSupportBundleImage:                           SettingDefinitionBundleImage,
```
yasker, line 202: Wrong naming pattern. Should be `SettingDefinitionSupportBundleImage`

types/setting.go
```diff
@@ -195,6 +199,28 @@ var (
...
+		SettingNameSupportBundleImagePullPolicy:                 SettingDefinitionSupportBundleImagePullPolicy,
```
yasker, line 203: We shouldn't need a pull policy (see other images for example). It should be `IfNotPresent` by default for the air gap environment.

types/setting.go
```diff
@@ -195,6 +199,28 @@ var (
...
+	}
+
+	SettingDefinitionBundleImage = SettingDefinition{
+		DisplayName: "Support Bundle Image Name",
+		Description: "The image name of the support bundle",
+		Category:    SettingCategoryGeneral,
+		Type:        SettingTypeString,
+		Required:    false,
```
yasker, line 211: `Required` is yes.

types/setting.go
```diff
@@ -195,6 +199,28 @@ var (
...
+		ReadOnly:    false,
+		Default:     "rancher/harvester-support-bundle-utils:master-head",
```
yasker, line 213: I see why the image pull policy is an option now. But don't use `master-head`. We need to release (and it cannot be called `harvester-support-bundle-utils`) and pin down a released version, otherwise, API might be imcompatible.

types/types.go
```diff
@@ -180,6 +180,18 @@ const (
...
 
+const (
+	StateNone       = ""
+	StateGenerating = "generating"
+	StateError      = "error"
+	StateReady      = "ready"
+
+	SupportBundleLabelKey = "longhorn.io/supportbundle"
+
+	AppManager = "support-bundle-manager"
```
yasker, line 191: `AppManager/AppAgent`? This name is too generic.

types/types.go
```diff
@@ -180,6 +180,18 @@ const (
 	replicaManagerPrefix  = instanceManagerPrefix + "r-"
 )
 
+const (
+	StateNone       = ""
```
yasker, line 184: `StateNone` etc name is too generic.

controller/supportbundle_controller.go
```diff
@@ -0,0 +1,353 @@
+package controller
+
+import (
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/longhorn/longhorn-manager/datastore"
```
yasker, line 8: `longhorn` part is the last section.

```
built-in imports
common imports
kubernetes imports
longhorn imports.
```

controller/supportbundle_controller.go
```diff
@@ -0,0 +1,353 @@
...
+	logger logrus.FieldLogger,
+	ds *datastore.DataStore,
+	scheme *runtime.Scheme,
+	supportBundleInformer lhinformers.SupportBundleInformer,
+	kubeClient clientset.Interface,
+	controllerID string,
+	namespace string,
+	serviceAccount string) *SupportBundleController {
+	eventBroadcaster := record.NewBroadcaster()
+	eventBroadcaster.StartLogging(logrus.Infof)
```
yasker, line 63: `Infof` level is very high. What does the log look like?

api/model.go
```diff
@@ -1369,13 +1371,18 @@ type Server struct {
...
+	httpClient *http.Client
 }
 
 func NewServer(m *manager.VolumeManager, wsc *controller.WebsocketController) *Server {
 	s := &Server{
 		m:   m,
 		wsc: wsc,
 		fwd: NewFwd(m),
+		httpClient: &http.Client{
+			Timeout: 30 * time.Second,
```
yasker, line 1368: Does it mean support bundle only has 30 seconds to generate? generation a big support bundle can take hours. It cannot be done in a synchronized manner.

yasker (changes requested):

yasker:
**This shouldn't be merged into v1.2.0. This is targeting v1.2.1.**

api/model.go
shangma: This 30 second is the time for longhorn to download the support bundle zip file from the support bundle manager, which gets deployed by the deployment. So longhorn has a server to handle a lot of requests, and it uses this client to download the zip file from the support bundle manager which itself has a server running inside. @yasker

yasker: Does that mean the support bundle needs to be downloaded in 30 seconds? And why do we need to define this one-time HTTP client in the server object?

api/router.go
```diff
@@ -149,6 +149,8 @@ func NewRouter(s *Server) *mux.Router {
 		r.Methods("POST").Path("/v1/backingimages/{name}").Queries("action", name).Handler(f(schemas, action))
 	}
 
+	r.Methods("GET").Path("/v1/supportbundles/{bundleName}/download").Handler(f(schemas, s.SupportBundleDownload))
```
yasker, line 152: With the new API, how does UI initiate the support bundle generation?

util/http.go
```diff
@@ -13,3 +14,30 @@ func CopyReq(req *http.Request) *http.Request {
...
+	if err != nil {
+		return []byte(`{\"errors\":[\"Failed to parse response body\"]}`)
+	}
+	return respBody
+}
+
+func ResponseOKWithBody(rw http.ResponseWriter, obj interface{}) {
+	rw.Header().Set("Content-type", "application/json")
+	rw.WriteHeader(http.StatusOK)
+	_, _ = rw.Write(ResponseBody(obj))
```
yasker, line 29: Don't ignore the error code.

util/http.go
```diff
@@ -13,3 +14,30 @@ func CopyReq(req *http.Request) *http.Request {
...
+	rw.WriteHeader(http.StatusOK)
+}
+
+func ResponseError(rw http.ResponseWriter, statusCode int, err error) {
+	ResponseErrorMsg(rw, statusCode, err.Error())
+}
+
+func ResponseErrorMsg(rw http.ResponseWriter, statusCode int, errMsg string) {
+	rw.WriteHeader(statusCode)
+	_, _ = rw.Write(ResponseBody([]string{errMsg}))
```
yasker, line 42: Same here.

yasker (changes requested):

types/types.go
shangma: Updated

types/setting.go
shangma: Updated

types/setting.go
shangma: Updated to use the existing image pull policy

types/setting.go
shangma: Updated to use the existing image pull policy: SettingDefinitionSystemManagedPodsImagePullPolicy

types/resource.go
shangma: Updated

types/setting.go
shangma: Updated

api/supportbundle.go
```diff
@@ -0,0 +1,103 @@
...
+		return err
+	}
+
+	url := fmt.Sprintf("http://%s:8080/bundle", managerPodIP)
+	req, err := http.NewRequestWithContext(r.Context(), http.MethodGet, url, nil)
+	if err != nil {
+		util.ResponseError(rw, http.StatusInternalServerError, err)
+		return err
+	}
+	resp, err := s.httpClient.Do(req)
```
yasker, line 51: We cannot download to the mgr's local directory. Need to make it possible for the UI to download the zip file.

api/router.go
shangma: @yasker  add the original http handler back. so no ui change is needed.

util/http.go
shangma: @yasker  updated.

util/http.go
shangma: @yasker  updated.

types/resource.go
```diff
@@ -776,3 +778,52 @@ type RecurringJobSpec struct {
...
+
+	SupportBundleProgressPercentageYaml  = 20
+	SupportBundleProgressPercentageLogs  = 75
+	SupportBundleProgressPercentageTotal = 100
+)
+
+type SupportBundleSpec struct {
+	Name               string              `json:"name"`
+	FileName           string              `json:"filename"`
+	State              SuppportBundleState `json:"state"`
```
yasker, line 815: Why `state, size, error, progressPercentage` are in the spec? Those are not user-defined.

Also, how is `fileName` decided? Is it user defined?

controller/supportbundle_controller.go
```diff
@@ -0,0 +1,456 @@
...
+func (sbc *SupportBundleController) setError(sb *longhorn.SupportBundle, reason string) (*longhorn.SupportBundle, error) {
+	log := getLoggerForSupportBundle(sbc.logger, sb)
+	log.Errorf(reason)
+
+	toUpdate := sb.DeepCopy()
+	sbc.SupportBundleInitialized.False(toUpdate)
+	sbc.SupportBundleInitialized.Message(toUpdate, reason)
+
+	toUpdate.Status.State = types.SupportBundleStateError
+	return sbc.ds.UpdateSupportBundle(toUpdate)
```
yasker, line 384: Don't call `UpdateSupportBundle` more than once in the controller. Follow the pattern in https://github.com/longhorn/longhorn-manager/blob/master/controller/volume_controller.go to only update on exiting the reconciliation loop.

controller/supportbundle_controller.go
```diff
@@ -0,0 +1,456 @@
...
+	err = json.NewDecoder(resp.Body).Decode(s)
+	if err != nil {
+		return nil, err
+	}
+
+	return s, nil
+}
+
+type ManagerStatus struct {
+	// phase to collect bundle
```
yasker, line 445: No need to those comments since it's pretty self-explained, and it's also an internal structure.

controller/supportbundle_controller.go
```diff
@@ -0,0 +1,456 @@
...
+	Phase string
```
yasker, line 446: Can we use type rather than string here?

controller/supportbundle_controller.go
```diff
@@ -0,0 +1,456 @@
...
+
+	// fail to collect bundle
+	Error bool
```
yasker, line 449: Seems same as `ErrorMessage != ""`

controller/supportbundle_controller.go
```diff
@@ -0,0 +1,456 @@
...
+
+	// error message
+	ErrorMessage string
+
+	// progress of the bundle collecting. 0 - 100.
+	Progress int
+
+	// bundle filename
+	Filename string
```
yasker, line 458: `FileName`.

controller/supportbundle_controller.go
```diff
@@ -0,0 +1,456 @@
...
+
+	// bundle filesize
+	Filesize int64
```
yasker, line 461: `FileSize`

util/http.go
```diff
@@ -2,6 +2,8 @@ package util
 
 import (
 	"bytes"
+	"encoding/json"
+	"github.com/sirupsen/logrus"
```
yasker, line 6: Wrong section.

k8s/pkg/client/informers/externalversions/longhorn/v1beta1/interface.go
```diff
@@ -52,6 +52,8 @@ type Interface interface {
 	Settings() SettingInformer
 	// ShareManagers returns a ShareManagerInformer.
 	ShareManagers() ShareManagerInformer
+	// SupportBundles returns a SupportBundleInformer.
```
yasker, line 55: Should this be in the CRD type generation commit?

yasker (changes requested):

types/resource.go
```diff
@@ -776,3 +778,51 @@ type RecurringJobSpec struct {
...
+type SupportBundleSpec struct {
+	Name        string `json:"name"`
+	IssueURL    string `json:"issueURL"`
+	Description string `json:"description"`
+}
+
+type SupportBundleStatus struct {
+	State        SuppportBundleState `json:"state,omitempty"`
+	Progress     int                 `json:"progress,omitempty"`
+	FileName     string              `json:"filename,omitempty"`
```
yasker, line 824: JSON is `camelCase`, so it should be `fileName` instead of `filename`. Same below.

yasker (changes requested):
Not a complete review since I know you're working on moving the controller to the support bundle kit.

controller/supportbundle_controller.go
shangma: @yasker  updated.

innobead:
Closing, we will restart this in near future.

PR #999, opened 2021-08-23
controller: Fix a informer typo for volume controller

shuo-wu:
longhorn/longhorn#2915


innobead (approved):

innobead:
Also thanks @jenting pointed out this typo.

PR #1000, opened 2021-08-24
Fix the bug that lead to the uninstaller loop in instance manager deletion and recreation

shuo-wu:
longhorn/longhorn#2919


innobead (approved):
LGTM

innobead:
@shuo-wu need to backport to 1.1.3.

PhanLe1010:
1000th PR 🎉

PR #1001, opened 2021-08-24
[Backport][v1.1.3]Fix the bug that lead to the uninstaller loop in instance manager deletion and recreation

shuo-wu:
longhorn/longhorn#2919


shuo-wu:
I just realize that we may need to update instance managers in the upgrade path

innobead (approved):
LGTM

PR #1007, opened 2021-08-25
Lower csi sidecar log level

joshimoo:
Reduce the csi side car log level to V(2)

Which is recommended for most systems, since we don't heavily rely on the
logs from the csi sidecars, this should be fine. More information on the
available log levels below.

https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/logging.md

longhorn/longhorn#1202


innobead (approved):
LGTM

PR #1008, opened 2021-08-26
feat: add WebSocket for Backup page

jenting:
#### Proposal Change

Add WebSocket endpoint `/v1/ws/backupvolumes` and `/v1/ws/backups`.

Haven't found a way to allow passing the volume name as a parameter to filter per-backup volume's backups change. Either by:
- `/v1/ws/backups/{volName}`
- `/v1/ws/backups?volName=<volName>`
- `/v1/ws/backupvolumes/{backupName}`
    
So, the current behavior for the `backups` WebSocket endpoint is that sending all the Backup CRs to the frontend, and frontend will do filtering according to the field `backups.volumeName` when entering the page `Backup -> a Backup Volume`

This follows the same behavior as the `volume` WebSocket endpoint that also sending all the volume/engine/replica changes to the frontend. We need to improve the WebSocket data traffic sending in the future.

#### Issues

https://github.com/longhorn/longhorn/issues/2740

### DIFF
--- a/api/backup.go
+++ b/api/backup.go
@@ -6,6 +6,7 @@ import (
 	"github.com/gorilla/mux"
 	"github.com/pkg/errors"
 	"github.com/rancher/go-rancher/api"
+	"github.com/rancher/go-rancher/client"
 	"github.com/sirupsen/logrus"
 )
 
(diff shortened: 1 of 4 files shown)


innobead (approved):
LGTM

PR #1009, opened 2021-08-27
Update the BackupTarget status.lastSyncedAt only when the client is correct

jenting:
#### Proposal Changes

Updates the `backupTarget.Status.Available = <false|true>` only when the backup target client is correct.

#### Issues

https://github.com/longhorn/longhorn/issues/2887

### DIFF
--- a/controller/backup_target_controller.go
+++ b/controller/backup_target_controller.go
@@ -2,6 +2,7 @@ package controller
 
 import (
 	"fmt"
+	"reflect"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


jenting:
> > 1. I think we can directly use `time.Now()`, which provides milli-sec level precision time string. So that the timestamp format in all kinds of Longhorn CRs keeps the same. The sec level precision `metav1.Time` provides is not enough in our use case regardless of this issue.
> 
> I'll handle this part in the other PR.

Address in https://github.com/longhorn/longhorn-manager/pull/1045

PR #1014, opened 2021-08-27
Fixed bug volume cannot attach after Kubernetes node is deleted

PhanLe1010:
Don't attempt to find the IM on a down or deleted node before starting a replica.
Attempting to find the IM on a down or deleted node will always return error and block volume controller sync loop. 

longhorn/longhorn#2848


shuo-wu (approved):

shuo-wu:
need to backport to 1.1.3.

PR #1015, opened 2021-08-28
Ignore delete backup volume error if not found

jenting:
#### Proposal Change

When the backup target URL is unset, it cleans up all the backup volume CRs.
If the user unset the backup target URL, and if the other controller trigger backup_target_controller within the deleted BackupVolume CRs period, the backup_target_controller be trigger twice and there are two processes simultaneously
clean up all the backup volume CRs. It'd cause partially BackupVolume CRs to fail at clean up.

Therefore, we ignore the delete backup volume not found error.

#### Issue

https://github.com/longhorn/longhorn/issues/2913


innobead (approved):
LGTM

PR #1027, opened 2021-09-02
Fix recurringJob CRs not created during upgrade

c3y1huang:
@kaxing has managed to reproduce this on his local machine. The issue could be seen during manager re-try at failed run that is causing the volume spec to get removed before CR creation.

This PR refactors the recurringJob upgrade flow to ensure the recurringJob CRs are created before updating the storageClass and volume CRs.

This also fixed the environment-based inconsistent behavior when creating recurring job CR with null spec.groups.

https://github.com/longhorn/longhorn/issues/2961#issuecomment-909810861

### DIFF
--- a/upgrade/v110to120/upgrade.go
+++ b/upgrade/v110to120/upgrade.go
@@ -9,6 +9,7 @@ import (
 	"github.com/sirupsen/logrus"
 
 	batchv1beta1 "k8s.io/api/batch/v1beta1"
+	corev1 "k8s.io/api/core/v1"
 	storagev1 "k8s.io/api/storage/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -35,34 +36,6 @@ func UpgradeCRs(namespace string, lhClient *lhclientset.Clientset, kubeClient *c
 		return err
 	}
 	if err := upgradeInstanceManagers(namespace, lhClient, kubeClient); err != nil {
-		return err
-	}
-	return nil
-}
-
-func upgradeRecurringJobs(namespace string, lhClient *lhclientset.Clientset, kubeClient *clientset.Clientset) (err error) {
-	defer func() {
-		err = errors.Wrapf(err, upgradeLogPrefix+"upgrade recurring jobs failed")
-	}()
-
-	// Need to transfer volume spec recurringJobs to CRs because Longhorn
-	// is migrating to label-driven recurring job CRD.
-	idMapSpec := make(map[string]*types.RecurringJobSpec)
-	err = translateStorageClassRecurringJobsToSelector(namespace, kubeClient, idMapSpec)
-	if err != nil {
-		return err
-	}
-	err = translateVolumeRecurringJobsToLabel(namespace, lhClient, idMapSpec)
-	if err != nil {
-		return err
-	}
-	err = translateVolumeRecurringJobToCRs(namespace, lhClient, idMapSpec)
-	if err != nil {
-		return err
-	}
-
-	err = cleanupAppliedVolumeCronJobs(namespace, lhClient, kubeClient)
-	if err != nil {
 		return err
 	}
 	return nil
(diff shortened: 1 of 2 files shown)


upgrade/v110to120/upgrade.go
```diff
@@ -179,74 +209,93 @@ func translateStorageClassRecurringJobsToSelector(namespace string, kubeClient *
...
-	logrus.Infof(upgradeLogPrefix+"Updating %v configmap", storageClassCM.Name)
-	if _, err := kubeClient.CoreV1().ConfigMaps(namespace).Update(context.TODO(), storageClassCM, metav1.UpdateOptions{}); err != nil {
-		return errors.Wrapf(err, "failed to update %v configmap", storageClassCM.Name)
+	run.log.Info(upgradeLogPrefix + "Removing recurringJobs")
+	delete(run.storageClass.Parameters, "recurringJobs")
+
+	logrus.Infof(upgradeLogPrefix+"Updating %v configmap", run.storageClassConfigMap.Name)
+	newStorageClassYAML, err := yaml.Marshal(run.storageClass)
+	run.storageClassConfigMap.Data["storageclass.yaml"] = string(newStorageClassYAML)
+	if _, err := run.kubeClient.CoreV1().ConfigMaps(run.namespace).Update(context.TODO(), run.storageClassConfigMap, metav1.UpdateOptions{}); err != nil {
```
shuo-wu, line 269: The StorageClass ConfigMap update (include unsetting `storageClass.Parameters["recurringJobs"]`) should  not be done before the RecurringJob CR creation.

shuo-wu (changes requested):
The implementation is a little bit messy. Maybe you can re-organize the code with the following steps:
1. Get the default StorageClass ConfigMap and the StorageClass
2. Generate RecurringJob CR manifests based on SC.
3. Generate the new StorageClass ConfigMap manifest
4. Generate the new volume spec manifests
5. Create RecurringJob CRs then update the StorageClass ConfigMap and the volume spec based on the manifests.

upgrade/v110to120/upgrade.go
c3y1huang: This method is called after recurringJob CR creation.
```
	if err := run.createRecurringJobCRs(); err != nil {
		return err
	}
	if err := run.convertToSelectors(recurringJobIDs); err != nil {
		return err
	}
```

upgrade/v110to120/upgrade.go
```diff
@@ -121,37 +94,94 @@ func upgradeInstanceManagers(namespace string, lhClient *lhclientset.Clientset,
...
+
+	err = run.cleanupAppliedVolumeCronJobs()
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (run *recurringJob) translateStorageClassRecurringJobs() (err error) {
+	revertLog := run.log
```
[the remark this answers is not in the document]
c3y1huang: The purpose is to revert the extra fields added in this method.

upgrade/v110to120/upgrade.go
```diff
@@ -121,37 +94,94 @@ func upgradeInstanceManagers(namespace string, lhClient *lhclientset.Clientset,
...
+	defer func() {
+		if err == nil {
+			run.log.Info(upgradeLogPrefix + "Finished storageClass recurring job translation")
```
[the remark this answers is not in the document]
c3y1huang: I would like to include `run.log = run.log.WithField("translate", "storage-class-recurring-jobs")`. `revertLog.Debug` doesn't include the extra field.

upgrade/v110to120/upgrade.go
shuo-wu: Oh sorry. The comparison shown by the GitHub review page confused me. The implementation here is right.

upgrade/v110to120/upgrade.go
```diff
@@ -121,37 +94,94 @@ func upgradeInstanceManagers(namespace string, lhClient *lhclientset.Clientset,
 	return nil
 }
 
+type recurringJob struct {
```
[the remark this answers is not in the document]
c3y1huang: Addressed.

upgrade/v110to120/upgrade.go
```diff
@@ -260,124 +309,150 @@ func translateVolumeRecurringJobsToLabel(namespace string, lhClient *lhclientset
...
+	if len(run.recurringJobMapSpec) == 0 {
+		run.log.Debug(upgradeLogPrefix + "Found 0 recurring job")
+		return nil
+	}
+
+	for _, spec := range run.recurringJobMapSpec {
+		run.log.Infof(upgradeLogPrefix+"Creating %v recurring job CR", spec.Name)
+		newRecurringJob := &longhorn.RecurringJob{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: spec.Name,
```
shuo-wu, line 422: I just realized the CR name is stored in the spec as well... Can we remove it in the future or now?

upgrade/v110to120/upgrade.go
```diff
@@ -260,124 +309,150 @@ func translateVolumeRecurringJobsToLabel(namespace string, lhClient *lhclientset
...
+			},
+			Spec: *spec,
+		}
+		run.log.Debugf(upgradeLogPrefix+"Checking if %v recurring job CR already exists", spec.Name)
+		recurringJobClient := run.lhClient.LonghornV1beta1().RecurringJobs(run.namespace)
```
shuo-wu, line 427: NIT: move this outside of the for loop

upgrade/v110to120/upgrade.go
```diff
@@ -260,124 +309,150 @@ func translateVolumeRecurringJobsToLabel(namespace string, lhClient *lhclientset
...
+		obj, err := recurringJobClient.Get(context.TODO(), newRecurringJob.Name, metav1.GetOptions{})
+		if err == nil {
+			run.log.Debugf(upgradeLogPrefix+"Recurring job CR already exists %v", obj)
+			continue
+		}
+		if !apierrors.IsNotFound(err) {
+			run.log.Debugf(upgradeLogPrefix+"Failed to get recurring job %v", spec.Name)
```
shuo-wu, line 434:
```
run.log.Errorf()
continnue
```
Or we can error out here.

upgrade/v110to120/upgrade.go
```diff
@@ -179,74 +209,93 @@ func translateStorageClassRecurringJobsToSelector(namespace string, kubeClient *
...
+
+	if len(recurringJobIDs) == 0 {
+		run.log.Debug(upgradeLogPrefix + "Found 0 recurring job to convert to recurring job selector")
+		return
+	}
+
+	selectors := []recurringJobSelector{}
+	selectorIDs := []string{}
+	selectorParameter, ok := run.storageClass.Parameters["recurringJobSelector"]
 	if ok {
```
shuo-wu, line 238: NIT:
```suggestion
if selectorParameter, ok := run.storageClass.Parameters["recurringJobSelector"]; ok {
```

upgrade/v110to120/upgrade.go
```diff
@@ -179,74 +209,93 @@ func translateStorageClassRecurringJobsToSelector(namespace string, kubeClient *
...
+		err = errors.Wrapf(err, "convert storageClass recurringJobs to recurringJobSelector failed")
+	}()
+	run.log = run.log.WithField("action", "convert-recurring-jobs-to-selectors")
+
+	if len(recurringJobIDs) == 0 {
+		run.log.Debug(upgradeLogPrefix + "Found 0 recurring job to convert to recurring job selector")
+		return
+	}
+
+	selectors := []recurringJobSelector{}
```
shuo-wu, line 235: Can you directly use a map instead? Then you don't need to store all selector ID in `selectorIDs` then rely on `selectorIDs` to skip the duplication in `recurringJobIDs`.

This change will improve the readability of the funcion. The only extra operation is that this map (values) should be converted to a list in the end.

shuo-wu (changes requested):

upgrade/v110to120/upgrade.go
c3y1huang: I prefer not to do the extra conversion because
`run.storageClass.Parameters["recurringJobSelector"]` comes as an array. To use map instead will need to convert the existing recurringJobSelector to map and then at the end back to array again.

upgrade/v110to120/upgrade.go
c3y1huang: Thanks, fixed.

upgrade/v110to120/upgrade.go
c3y1huang: Thanks, fixed.

upgrade/v110to120/upgrade.go
c3y1huang: Thanks, fixed.

upgrade/v110to120/upgrade.go
c3y1huang: I think this PR should stick with the upgrade fix. We can look at the name in spec when we start to do https://github.com/longhorn/longhorn/issues/2865.

upgrade/v110to120/upgrade.go
shuo-wu: Got it.

upgrade/v110to120/upgrade.go
shuo-wu: OK. We can deprecate this spec field in this ticket. But please avoid using it from now on.

shuo-wu (approved):

upgrade/v110to120/upgrade.go
```diff
@@ -260,124 +308,150 @@ func translateVolumeRecurringJobsToLabel(namespace string, lhClient *lhclientset
...
+		return errors.Wrapf(err, "failed to decoding YAML string")
 	}
-	storageclass, ok := obj.(*storagev1.StorageClass)
+	sc, ok := obj.(*storagev1.StorageClass)
 	if !ok {
-		return nil, fmt.Errorf("invalid storageclass YAML string: %v", storageclassYAML)
+		return fmt.Errorf("invalid storageclass YAML string: %v", storageclassYAML)
 	}
-	return storageclass, nil
+	run.storageClass = sc
```
innobead, line 453: The function naming is quite misleading because the property update happening in `getXXX` function.` I would suggest changing to other function name like `initStorageClassProFromYAML`.

innobead (changes requested):
LGTM, but just one minor namig suggested to change.

upgrade/v110to120/upgrade.go
c3y1huang: Addressed.

innobead (approved):

innobead:
@c3y1huang please backport to 1.2.x.

PR #1028, opened 2021-09-02
Backport to v1.2.1: Explicitly set default fsType to EXT4

PhanLe1010:
This allow Kubernetes to properly changes the volume ownership
and permissions when user specify fsGroup in workload pod

longhorn/longhorn#2964


innobead (approved):
LGTM

PR #1033, opened 2021-09-08
Allow migration for degraded volumes

shuo-wu:
Longhorn/longhorn#2805

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -2861,21 +2861,14 @@ func (vc *VolumeController) getCurrentEngineAndCleanupOthers(v *longhorn.Volume,
 }
 
 func (vc *VolumeController) createAndStartMatchingReplicas(v *longhorn.Volume,
-	rs, pathToOldRunnngRs, pathToNewRs map[string]*longhorn.Replica,
+	rs, pathToOldRs, pathToNewRs map[string]*longhorn.Replica,
 	fixupFunc func(r *longhorn.Replica, obj string), obj string) error {
-	log := getLoggerForVolume(vc.logger, v)
 
-	if len(pathToNewRs) == v.Spec.NumberOfReplicas {
-		return nil
+	if len(pathToOldRs) == 0 {
+		return fmt.Errorf("no old replica for the volume")
 	}
 
-	if len(pathToOldRunnngRs) != v.Spec.NumberOfReplicas {
-		log.Debugf("Volume old healthy replica count %v doesn't match the desired replica count %v",
-			len(pathToOldRunnngRs), v.Spec.NumberOfReplicas)
-		return nil
-	}
-
-	for path, r := range pathToOldRunnngRs {
+	for path, r := range pathToOldRs {
 		if pathToNewRs[path] != nil {
 			continue
 		}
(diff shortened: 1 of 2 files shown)


controller/volume_controller.go
```diff
@@ -3054,6 +3066,45 @@ func (vc *VolumeController) processMigration(v *longhorn.Volume, es map[string]*
 	return nil
 }
 
+func (vc *VolumeController) IsReplicaUnavailable(r *longhorn.Replica) (bool, error) {
+	log := vc.logger.WithFields(logrus.Fields{
+		"replica":       r.Name,
+		"replicaNodeID": r.Spec.NodeID,
+	})
+
+	if r.Spec.FailedAt != "" {
```
joshimoo, line 3075: NIT: can combine these

controller/volume_controller.go
```diff
@@ -3054,6 +3066,45 @@ func (vc *VolumeController) processMigration(v *longhorn.Volume, es map[string]*
...
+	}
+
+	isDownOrDeleted, err := vc.ds.IsNodeDownOrDeleted(r.Spec.NodeID)
+	if err != nil {
+		log.WithError(err).Errorf("Unable to check if node %v is still running for failed replica", r.Spec.NodeID)
+		return true, err
+	}
+	if isDownOrDeleted {
+		return true, nil
+	}
```
joshimoo, line 3089: NIT: can combine these, since err is nil when `isDownOrDeleted`

controller/volume_controller.go
```diff
@@ -2995,16 +2996,27 @@ func (vc *VolumeController) processMigration(v *longhorn.Volume, es map[string]*
...
+		if err != nil {
+			return err
+		}
+		if isUnavailable {
 			continue
 		}
+		dataPath := types.GetReplicaDataPath(r.Spec.DiskPath, r.Spec.DataDirectoryName)
 		if r.Spec.EngineName == currentEngine.Name {
-			currentReplicas[dataPath] = r
+			if currentEngine.Status.ReplicaModeMap[r.Name] == types.ReplicaModeWO {
```
joshimoo, line 3012: So we still block while a rebuild is in progress, is that desired?

joshimoo (dismissed):
LGTM is it desired to block migration while a rebuild is in progress, I am thinking about the harvester case where there are more then 3 nodes. This would still fail the migration in that scenario?

joshimoo:
@shuo-wu you will also need to change this condition check as part of the attach.
So the migration can happen when the volume is in degraded state.
https://github.com/longhorn/longhorn-manager/blob/2bf1d608fb388dc062a256ab4a79dbcab588f6b2/manager/volume.go#L434-L436

controller/volume_controller.go
shuo-wu: Right. Launching the new migration replica process or adding it to the engine process will fail if the rebuilding is in progress. And I don't want to modify the longhorn-engine part, which will make things complicated.

shuo-wu:
> @shuo-wu you will also need to change this condition check as part of the attach.
> So the migration can happen when the volume is in degraded state.
> https://github.com/longhorn/longhorn-manager/blob/2bf1d608fb388dc062a256ab4a79dbcab588f6b2/manager/volume.go#L434-L436

This part is already modified in the PR.

shuo-wu:
If there is replica rebuilding for when users decide to migrate the degraded volume. The migration will be delayed but the request won't fail.

controller/volume_controller.go
shuo-wu: Although the returned boolean is true in both cases, the log will be printed only when the err is not nil. I prefer not to retain the log rather than combining these like:
```
if isDownOrDeleted, err := vc.ds.IsNodeDownOrDeleted(r.Spec.NodeID); err != nil || isDownOrDeleted{
    return true, err
}
```

joshimoo (approved):
LGTM

PR #1034, opened 2021-09-08
Track terminating pods

joshimoo:
We changed the behavior so that we only set lastPodRefAt once there are
no more pods present, previously this was already set when the last pod
went into terminating state.

I added a unit test for both cases, to verify the behavior of tracking in
terminating state and lastPodRefAt setting once there are no more pods
present for that volume.

longhorn/longhorn#2874

Signed-off-by: Joshua Moody <joshua.moody@suse.com>

### DIFF
--- a/controller/share_manager_controller.go
+++ b/controller/share_manager_controller.go
@@ -315,10 +315,7 @@ func (c *ShareManagerController) syncShareManager(key string) (err error) {
 	// update at the end, after the whole reconcile loop
 	existingShareManager := sm.DeepCopy()
 	defer func() {
-		if !reflect.DeepEqual(existingShareManager.Status, sm.Status) {
-			// we end up overwriting any prior errors
-			// but if the update works, the share manager will be
-			// enqueued again anyway
+		if err == nil && !reflect.DeepEqual(existingShareManager.Status, sm.Status) {
 			_, err = c.ds.UpdateShareManagerStatus(sm)
 		}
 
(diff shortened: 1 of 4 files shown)


joshimoo:
Testing https://github.com/longhorn/longhorn/issues/2960 to see if this fixes that issue, will have to make some additional changes.

shuo-wu (approved):

joshimoo:
Added a fix for https://github.com/longhorn/longhorn/issues/2960
by adding a stopping state, the controller will transition the share-manager into stopping state, when the volume is no longer beeing used, it will then wait for the share-manager pod to be cleaned up before transitioning into stopped state.

This breaks the previous loop of:
I.E a single pod workload, we could end up in a bad loop of
starting (new workload) -> error (remount) -> stopped (cleanup sm)

Since we would transition from stopped into starting, but the old pod has been marked for deletion but not cleaned up yet. This would signal error state, so we end up in the above cycle.

joshimoo:
/hold merge
Review please, will merge this after finishing my final test run.

controller/share_manager_controller.go
```diff
@@ -438,10 +435,10 @@ func (c *ShareManagerController) syncShareManagerVolume(sm *longhorn.ShareManage
 		// ensure volume gets detached if share manager needs to be cleaned up and hasn't stopped yet
 		// we need the isNotNeeded var so we don't accidentally detach manually attached volumes,
 		// while the share manager is no longer running (only run cleanup once)
-		if isNotNeeded && sm.Status.State != types.ShareManagerStateStopped {
+		if isNotNeeded && sm.Status.State != types.ShareManagerStateStopping && sm.Status.State != types.ShareManagerStateStopped {
```
shuo-wu, line 443: NIT: If `sm.Status.State != types.ShareManagerStateStopping && sm.Status.State != types.ShareManagerStateStopped` is used here, there is no need to re-check it below when setting `isNotNeeded` to `true`. Or vice versa.

Furthermore, I am thinking if we really need to check the share manager state before detachment. Is this check for avoiding detaching a stopped volume?

controller/share_manager_controller.go
```diff
@@ -438,10 +435,10 @@ func (c *ShareManagerController) syncShareManagerVolume(sm *longhorn.ShareManage
...
 			getLoggerForShareManager(c.logger, sm).Info("stopping share manager")
 			if err = c.detachShareManagerVolume(sm); err == nil {
-				sm.Status.State = types.ShareManagerStateStopped
+				sm.Status.State = types.ShareManagerStateStopping
 			}
```
shuo-wu, line 447: NIT: print out the error log

controller/share_manager_controller.go
joshimoo: Correct, this is so that if the share-manager is stopped we don't mess with the volume.
The extra state check is a little redundant but it helps make it more clear in this case.

controller/share_manager_controller.go
joshimoo: The error log is already handled in the detachShareManagerVolume.
The only other error that is possible is conflict, which we don't need to print.

joshimoo:
@shuo-wu @innobead my testing looks good can we merge this?

shuo-wu (approved):

innobead:
LGTM

PR #1037, opened 2021-09-10
[CSI Sanity] Auto-attach volume for K8s CSI snapshot

c3y1huang:
Some csi-sanity test fails because Longhorn doesn't support snapshot/backup when the volume is detached.

For Kubernetes CSI snapshot the volume should be fine with auto-attachment since the VolumeSnapshot spec uses PVC as the source. 

### DIFF
--- a/csi/controller_server.go
+++ b/csi/controller_server.go
@@ -13,13 +13,17 @@ import (
 	"github.com/container-storage-interface/spec/lib/go/csi"
 	"github.com/golang/protobuf/ptypes"
 	"github.com/golang/protobuf/ptypes/timestamp"
+	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
 	longhornclient "github.com/longhorn/longhorn-manager/client"
 	"github.com/longhorn/longhorn-manager/datastore"
 	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
+	lhclientset "github.com/longhorn/longhorn-manager/k8s/pkg/client/clientset/versioned"
 	"github.com/longhorn/longhorn-manager/types"
 	"github.com/longhorn/longhorn-manager/util"
 )
@@ -30,6 +34,7 @@ const (
 	tickAttachDetach        = 2 * time.Second
 	timeoutBackupInitiation = 60 * time.Second
 	tickBackupInitiation    = 5 * time.Second
+	tickBackupInterval      = 5 * time.Second
 	backupStateCompleted    = "Completed"
 
 	csiSnapshotTypeLonghornSnapshot         = "snap"
@@ -39,14 +44,18 @@ const (
 
 type ControllerServer struct {
 	apiClient   *longhornclient.RancherClient
+	lhClient    lhclientset.Interface
+	namespace   string
 	nodeID      string
 	caps        []*csi.ControllerServiceCapability
 	accessModes []*csi.VolumeCapability_AccessMode
 }
 
-func NewControllerServer(apiClient *longhornclient.RancherClient, nodeID string) *ControllerServer {
+func NewControllerServer(apiClient *longhornclient.RancherClient, lhClient lhclientset.Interface, ns, nodeID string) *ControllerServer {
 	return &ControllerServer{
 		apiClient: apiClient,
+		lhClient:  lhClient,
+		namespace: ns,
 		nodeID:    nodeID,
 		caps: getControllerServiceCapabilities(
 			[]csi.ControllerServiceCapability_RPC_Type{
@@ -442,7 +451,7 @@ func (cs *ControllerServer) ControllerPublishVolume(ctx context.Context, req *cs
 			volumeID, nodeID, volume.Controllers[0].HostId)
 	}
 
-	return cs.publishVolume(volume, nodeID, func() error {
+	return cs.publishVolume(volume, nodeID, "", func() error {
 		checkVolumePublished := func(vol *longhornclient.Volume) bool {
 			return isVolumeAvailableOn(vol, nodeID) || isVolumeShareAvailable(vol)
 		}
@@ -454,11 +463,12 @@ func (cs *ControllerServer) ControllerPublishVolume(ctx context.Context, req *cs
 }
 
 // publishVolume sends the actual attach request to the longhorn api and executes the passed waitForResult func
-func (cs *ControllerServer) publishVolume(volume *longhornclient.Volume, nodeID string, waitForResult func() error) (*csi.ControllerPublishVolumeResponse, error) {
+func (cs *ControllerServer) publishVolume(volume *longhornclient.Volume, nodeID, attachedBy string, waitForResult func() error) (*csi.ControllerPublishVolumeResponse, error) {
 	logrus.Debugf("ControllerPublishVolume: volume %s is ready to be attached, and the requested node is %s", volume.Name, nodeID)
 	input := &longhornclient.AttachInput{
 		HostId:          nodeID,
 		DisableFrontend: false,
+		AttachedBy:      attachedBy,
 	}
 
 	logrus.Infof("ControllerPublishVolume: volume %s with accessMode %s requesting publishing to %s", volume.Name, volume.AccessMode, nodeID)
@@ -595,9 +605,23 @@ func (cs *ControllerServer) createCSISnapshotTypeLonghornSnapshot(req *csi.Creat
 	if vol == nil {
 		return nil, status.Errorf(codes.NotFound, "volume %s not found", csiVolumeName)
 	}
-	// TODO: we may want to support taking snapshot of a detached volume in future
-	if vol.State != string(longhorn.VolumeStateAttached) {
-		return nil, status.Errorf(codes.FailedPrecondition, "volume %s invalid state %v for taking snapshot. Volume must be in %v state to take snapshot", vol.Name, vol.State, longhorn.VolumeStateAttached)
+	if vol.State == string(longhorn.VolumeStateDetached) {
+		log := logrus.WithFields(
+			logrus.Fields{
+				"volume":       vol.Name,
+				"snapshotType": "snapshot",
+			},
+		)
+		log.Debug("CSI snapshot attempting to attach volume to a random node")
+		nodeID, err := cs.autoAttachVolumeToRandomNode(csiSnapshotName, vol)
+		if err != nil {
+			return nil, status.Error(codes.Internal, err.Error())
+		}
+		log.Debugf("CSI snapshot attached volume to node %v", nodeID)
+
+		defer func() {
+			go cs.handleVolumeDetachment(csiVolumeName, csiSnapshotName, nodeID)
+		}()
 	}
 
 	var snapshot *longhornclient.Snapshot
@@ -636,6 +660,32 @@ func (cs *ControllerServer) createCSISnapshotTypeLonghornBackup(req *csi.CreateS
 		return nil, status.Error(codes.InvalidArgument, "Volume name must be provided")
 	} else if len(csiSnapshotName) == 0 {
 		return nil, status.Error(codes.InvalidArgument, "Snapshot name must be provided")
+	}
+
+	vol, err := cs.apiClient.Volume.ById(csiVolumeName)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to get volume %v", csiVolumeName)
+	}
+	if vol == nil {
+		return nil, status.Errorf(codes.NotFound, "volume %s not found", csiVolumeName)
+	}
+	if vol.State == string(longhorn.VolumeStateDetached) {
+		log := logrus.WithFields(
+			logrus.Fields{
+				"volume":       vol.Name,
+				"snapshotType": "backup",
+			},
+		)
+		log.Debug("CSI snapshot attempting to attach volume to a random node")
+		nodeID, err := cs.autoAttachVolumeToRandomNode(csiSnapshotName, vol)
+		if err != nil {
+			return nil, status.Error(codes.Internal, err.Error())
+		}
+		log.Debugf("CSI snapshot attached volume to node %v", nodeID)
+
+		defer func() {
+			go cs.handleVolumeDetachment(csiVolumeName, csiSnapshotName, nodeID)
+		}()
 	}
 
 	// we check for backup existence first, since it's possible that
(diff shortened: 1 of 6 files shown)


c3y1huang:
This fixes
```
[Fail] DeleteSnapshot [Controller Server] [It] should return appropriate values (no optional values added) 
[Fail] Controller Service [Controller Server] CreateVolume [It] should create volume from an existing source snapshot 
[Fail] CreateSnapshot [Controller Server] [It] should succeed when requesting to create a snapshot with already existing name and same source volume ID 
[Fail] CreateSnapshot [Controller Server] [It] should succeed when creating snapshot with maximum-length name 
```

innobead:
This will be pending until @c3y1huang back.

datastore/uncached.go
```diff
@@ -137,3 +144,47 @@ func (s *DataStore) GetAllConfigMaps() (runtime.Object, error) {
 func (s *DataStore) GetAllVolumeAttachments() (runtime.Object, error) {
 	return s.kubeClient.StorageV1().VolumeAttachments().List(context.TODO(), metav1.ListOptions{})
 }
+
+// FindARandomReadyNodeForVolumeAttachment return a random ready node to attach
+// the volume.
+// Return error if there is no ready node
+func FindARandomReadyNodeForVolumeAttachment(v *longhornclient.Volume, api *longhornclient.RancherClient) (string, error) {
```
shuo-wu, line 151: It's better not to use `longhornclient` in this package. You can get the node list before invoking this function then make the node list as an input.

shuo-wu (changes requested):

csi/controller_server.go
```diff
@@ -545,6 +546,34 @@ func (cs *ControllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
...
+		defer func() {
+			volume, err := cs.apiClient.Volume.ById(csiVolumeName)
+			if err != nil {
+				logrus.WithError(err).Errorf("volume %s unavailable", csiVolumeName)
+			}
+			if volume == nil {
+				logrus.WithError(err).Errorf("volume %s not found", csiVolumeName)
+			}
+			logrus.Debugf("CSI snapshot attempting to unattach %s volume on %s node", csiVolumeName, nodeID)
+			cs.unpublishVolume(volume, nodeID, func() error {
```
innobead, line 563: If somehow, the volume was not unpublished successfully. How is it? What users will encounter?

innobead: If the volume is already attached before creating a snapshot, the volume will be automatically detached after snapshot?

c3y1huang: AFAIK CSI snapshot can only have PVC as source where volume should have been already attached. Hense this fix is only for the csi-sanity test.

Ref: https://kubernetes.io/docs/concepts/storage/volume-snapshots/#volumesnapshots

c3y1huang: > If the volume is already attached before creating a snapshot, the volume will be automatically detached after snapshotting?

It's been handled by
```
	nodeID, err := cs.autoAttachVolumeToRandomNode(csiVolumeName)
	if err != nil {
	}
	if nodeID != "" {
```

datastore/uncached.go
c3y1huang: Fixed.

csi/controller_server.go
shuo-wu: > AFAIK CSI snapshot can only have PVC as source where volume should have been already attached.
A Longhorn volume that has a corresponding PVC can be detached.

shuo-wu: And I think the CSI plugin should check if the volume is already attached before picking up a random node (then start auto attachment).

datastore/uncached.go
shuo-wu: I mean, it's better not to use `"github.com/longhorn/longhorn-manager/client"` in ** the whole package** `datastore` rather than this go file.

csi/controller_server.go
c3y1huang: It's handled in autoAttachVolumeToRandomNode.
```
func (cs *ControllerServer) autoAttachVolumeToRandomNode(volumeName string) (nodeID string, err error) {
	volume, err := cs.apiClient.Volume.ById(volumeName)
	if err != nil {
		return "", errors.Wrapf(err, "failed to get volume %v", volumeName)
	}
	if volume.State == string(longhorn.VolumeStateDetached) {
```

shuo-wu: Oh, OK. I misread it... But can you separate the check from this function? It's like we will blindly do auto attachment for the CSI snapshot creation at the 1st glance.

datastore/uncached.go
shuo-wu: If it's hard to modify this function without introducing `"github.com/longhorn/longhorn-manager/client"`, you can directly add a similar function in csi rather than moving this function here.

csi/controller_server.go
c3y1huang: OK!

datastore/uncached.go
c3y1huang: Thanks, addressed.

csi/controller_server.go
```diff
@@ -545,6 +547,41 @@ func (cs *ControllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
...
+	}
+	nodeID := ""
+	if volume.State == string(longhorn.VolumeStateDetached) {
+		nodeID, err = cs.autoAttachVolumeToRandomNode(volume)
+		if err != nil {
+			return nil, status.Error(codes.Internal, err.Error())
+		}
+	}
+	if nodeID != "" {
+		defer func() {
```
shuo-wu, line 562:
```suggestion
	if volume.State == string(longhorn.VolumeStateDetached) {
		nodeID, err := cs.autoAttachVolumeToRandomNode(volume)
		if err != nil {
			return nil, status.Error(codes.Internal, err.Error())
		}
		defer func() {
			// auto detachment
			......
		}
	}
```

I am not sure why you prefer to use `nodeID` as the indication of volume auto attachment/detachment.

shuo-wu (changes requested):

csi/controller_server.go
c3y1huang: Fixed.

csi/controller_server.go
```diff
@@ -545,6 +547,40 @@ func (cs *ControllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
...
+	volume, err := cs.apiClient.Volume.ById(csiVolumeName)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to get volume %v", csiVolumeName)
+	}
+	if volume.State == string(longhorn.VolumeStateDetached) {
+		nodeID, err := cs.autoAttachVolumeToRandomNode(volume)
+		if err != nil {
+			return nil, status.Error(codes.Internal, err.Error())
+		}
+		if nodeID != "" {
```
shuo-wu, line 559: Question: Why can there be an empty node ID with no error returned after the auto attachment?

c3y1huang: There shouldn't be a case. Nice catch. Fixed.

csi/controller_server.go
```diff
@@ -545,6 +547,45 @@ func (cs *ControllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
...
+		nodeID, err := cs.autoAttachVolumeToRandomNode(volume)
+		if err != nil {
+			return nil, status.Error(codes.Internal, err.Error())
+		}
+		logrus.Debugf("CSI snapshot attached %v volume to %v node", volume.Name, nodeID)
+
+		defer func() {
+			volume, err := cs.apiClient.Volume.ById(csiVolumeName)
+			if err != nil {
+				logrus.WithError(err).Errorf("volume %v unavailable", csiVolumeName)
```
shuo-wu, line 565: should `return` after log printing. similar below.

shuo-wu (changes requested):

csi/controller_server.go
c3y1huang: Fixed. Thanks!

datastore/uncached.go
```diff
@@ -137,3 +142,37 @@ func (s *DataStore) GetAllConfigMaps() (runtime.Object, error) {
 func (s *DataStore) GetAllVolumeAttachments() (runtime.Object, error) {
 	return s.kubeClient.StorageV1().VolumeAttachments().List(context.TODO(), metav1.ListOptions{})
 }
+
+// FindARandomReadyNode return a random ready node.
+// Return error if there is no ready node
+func FindARandomReadyNode(ei *longhorn.EngineImage, nodeList *longhorn.NodeList) (string, error) {
```
innobead, line 148: Why do you add the function in datastore pkg? or just no good place to add it?

Just thinking if we should better organize the shared util function/service well in the future. (not a request to change here)

csi/controller_server.go
```diff
@@ -545,6 +547,48 @@ func (cs *ControllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
...
+				return
+			}
+			if volume == nil {
+				logrus.WithError(err).Errorf("volume %v not found", csiVolumeName)
+				return
+			}
+			logrus.Debugf("CSI snapshot attempting to detach %v volume on %s node", csiVolumeName, nodeID)
+			_, err = cs.unpublishVolume(volume, nodeID, func() error {
+				isSharedVolume := requiresSharedAccess(volume, nil) && !volume.Migratable
```
innobead, line 574: Why if the volume is shared volume, we don't need to wait until it is unpublished? Is it because we just assume probably it still used by other workloads? 

If that is the case, you will only do auto-attach only when the volume is detached, so why we need to have this condition to bypass the waiting?

innobead:
After this merge, we should probably update the documentation about we support CSI snapshots (backup) even a volume is detached.

datastore/uncached.go
innobead: cc @jenting @shuo-wu @joshimoo

shuo-wu: Personally, I would like to move CR CRUD functions into `datastore`. And I will consider this as a kind of `Get` function.

csi/controller_server.go
shuo-wu: Nice catch. BTW, the RWX volume will keep attached when there is **at least one workload** using it. Otherwise, we may still need to consider auto-attach/detach it.
https://github.com/longhorn/longhorn-manager/blob/24b38e1aa25c7522a39dda60cb992793b8136240/controller/share_manager_controller.go#L431-L434

c3y1huang: Fixed!

csi/controller_server.go
```diff
@@ -647,6 +690,68 @@ func (cs *ControllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
...
+	})
+	if err != nil {
+		return "", err
+	}
+	return nodeID, nil
+}
+
+// findARandomReadyNode is similar to FindARandomReadyNode but with
+// longhornclient since not possible without introducing longhorn manager client.
+func findARandomReadyNode(ei *longhornclient.EngineImage, nc *longhornclient.NodeCollection) (string, error) {
```
innobead, line 723: I still have a concern about a duplicated code even in CSI node plugin we only use longhorn client only.

Is it possible to extend longhorn manager `VolumeAttach` capability to support attaching to a random node if node id is empty? then you can simply call the volume attach method of longhorn client to request to attach to a random node instead of duplicating same code w/ different data structure here.

datastore/uncached.go
```diff
@@ -137,3 +142,37 @@ func (s *DataStore) GetAllConfigMaps() (runtime.Object, error) {
...
+	if ei.Status.State != longhorn.EngineImageStateDeployed && ei.Status.State != longhorn.EngineImageStateDeploying {
```
innobead, line 149: Do you know why we don't check `ei.Status.State != longhorn.EngineImageStateDeployed` ? because I don't see the below logic to check if the node really has the image ready or not. 

P.S. I know this is the previous implementation.

csi/controller_server.go
shuo-wu: I don't think it'sa good idea. Actually I already discussed this part with Chin-ya. Ideally it's Longhorn backend's responsibility to do auto attachment. Asking the CSI plugin (for CSI snapshot) or recurring pod (for backup recurring jobs) to send attach/detach calls is a short-term solution. We can re-visit it after refactoring the volume controller.

csi/controller_server.go
```diff
@@ -545,6 +547,47 @@ func (cs *ControllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
...
+			if err != nil {
+				logrus.WithError(err).Errorf("volume %v unavailable", csiVolumeName)
+				return
+			}
+			if volume == nil {
+				logrus.WithError(err).Errorf("volume %v not found", csiVolumeName)
+				return
+			}
+			logrus.Debugf("CSI snapshot attempting to detach %v volume on %s node", csiVolumeName, nodeID)
+			_, err = cs.unpublishVolume(volume, nodeID, func() error {
```
shuo-wu, line 573: Wait... The volume cannot be detached before the backup creation is **completed**. I think we should use a goroutine rather than a defer function here...

shuo-wu (changes requested):

csi/controller_server.go
innobead: Sure, also try to avoid any duplicated code existing when revisiting this part later.

datastore/uncached.go
jenting: Agree with Shuo

datastore/uncached.go
PhanLe1010: I agree with @innobead we should check to see if the node is ready and has engine image deployed. Similar to the logic here [ListReadyNodesWithReadyEngineImage](https://github.com/longhorn/longhorn-manager/blob/44b41a6d41886e3ff249c5375697ae8156a7194b/datastore/longhorn.go#L1913)

joshimoo (commented):
Backend should be responsible for managing attachment and all necessary actions for snapshot operations, see review comments in this issue for example:
https://github.com/longhorn/longhorn-manager/pull/1198#discussion_r789955564

csi/controller_server.go
c3y1huang: Removed this function instead use the one in datastore. The recurring job can leverage lhClient to get the nodes and engine image.

datastore/uncached.go
c3y1huang: Fixed.

csi/controller_server.go
c3y1huang: Addressed.

c3y1huang:
> Backend should be responsible for managing attachment and all necessary actions for snapshot operations, see review comments in this issue for example: [#1198 (comment)](https://github.com/longhorn/longhorn-manager/pull/1198#discussion_r789955564)

Keeping to the CSI controller server for the same reasons as [this](https://github.com/longhorn/longhorn/issues/3517#issuecomment-1046588116).

csi/controller_server.go
```diff
@@ -635,6 +645,29 @@ func (cs *ControllerServer) createCSISnapshotTypeLonghornBackup(req *csi.CreateS
...
 	}
 
+	volume, err := cs.apiClient.Volume.ById(csiVolumeName)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to get volume %v", csiVolumeName)
+	}
+	if volume.State == string(longhorn.VolumeStateDetached) {
+		log := logrus.WithFields(
+			logrus.Fields{
+				"CSISnapshot": "auto-attach",
```
shuo-wu, line 655: NIT: This k-v pair looks weird. I don't think we need it.

datastore/uncached.go
```diff
@@ -137,3 +141,32 @@ func (s *DataStore) GetAllConfigMaps() (runtime.Object, error) {
 func (s *DataStore) GetAllVolumeAttachments() (runtime.Object, error) {
 	return s.kubeClient.StorageV1().VolumeAttachments().List(context.TODO(), metav1.ListOptions{})
 }
+
+// FindARandomReadyNode return a random ready node.
+// Return error if there is no ready node
+func FindARandomReadyNode(engineImage *longhorn.EngineImage, nodes *longhorn.NodeList) (string, error) {
```
shuo-wu, line 147: NIT: `FindRandomReadyNode`

shuo-wu (approved):
In general LGTM

csi/controller_server.go
c3y1huang: Removed.

datastore/uncached.go
c3y1huang: Addressed.

shuo-wu (approved):

PhanLe1010:
We are only doing auto-attachment for createCSISnapshotTypeLonghornBackup but not createCSISnapshotTypeLonghornSnapshot in this PR. I think they should be handled the same way (either return error or doing auto-attachment for both CSI snapshot types) 

The bigger question is: have we a conclusion on doing the auto-attachment at the backend API level? I still see that this PR is doing auto-attachment in the Longhorn CSI plugin.

PhanLe1010 (changes requested):
We are only doing auto-attachment for createCSISnapshotTypeLonghornBackup but not createCSISnapshotTypeLonghornSnapshot in this PR. I think they should be handled the same way (either return error or doing auto-attachment for both CSI snapshot types)

The bigger question is: have we a conclusion on doing the auto-attachment at the backend API level? I still see that this PR is doing auto-attachment in the Longhorn CSI plugin.

c3y1huang:
> 
> The bigger question is: have we reached a conclusion on doing the auto-attachment at the backend API level? I still see that this PR is doing auto-attachment in the Longhorn CSI plugin.

After talking to @PhanLe1010 and @shuo-wu 
We have two options:
1. Consider this as a surgical fix. Merge this PR, and remove limitation in doc [here](https://github.com/longhorn/website/blob/v1.3.0/content/docs/1.3.0/snapshots-and-backups/csi-snapshot-support/csi-volume-snapshot-associated-with-longhorn-snapshot.md#current-limitation) and [here](https://github.com/longhorn/website/blob/v1.3.0/content/docs/1.3.0/snapshots-and-backups/csi-snapshot-support/csi-volume-snapshot-associated-with-longhorn-backup.md#current-limitation). 
2. Wait for longhorn term fix. Close this PR, revert [this](https://github.com/longhorn/website/pull/426) and skip the CSI sanity tests until refactoring for attaching/detaching is done. Maybe in https://github.com/longhorn/longhorn/issues/2527.

WDYT @innobead ?

innobead:
> 1. Consider this as a surgical fix. Merge this PR, and remove limitation in doc [here](https://github.com/longhorn/website/blob/v1.3.0/content/docs/1.3.0/snapshots-and-backups/csi-snapshot-support/csi-volume-snapshot-associated-with-longhorn-snapshot.md?rgh-link-date=2022-03-08T02%3A40%3A01Z#current-limitation) and [here](https://github.com/longhorn/website/blob/v1.3.0/content/docs/1.3.0/snapshots-and-backups/csi-snapshot-support/csi-volume-snapshot-associated-with-longhorn-backup.md?rgh-link-date=2022-03-08T02%3A40%3A01Z#current-limitation).
> 2. Wait for a long-term fix. Close this PR, revert [this](https://github.com/longhorn/website/pull/426) and skip the CSI sanity tests until refactoring for attaching/detaching is done. Maybe in [[REFACTOR] volume controller refactoring/split up, to simplify the control flow longhorn#2527](https://github.com/longhorn/longhorn/issues/2527).
> 
> WDYT @innobead ?

It seems we have many places about auto attach/detach (ex: recurring job, https://github.com/longhorn/longhorn-manager/commit/216493bcf97be725ce59f2b63e99728bccfbf895), and we definitely should consolidate them to make it consistent behavior to ensure all operations work correctly (attach/detach) at the end.

I don't mind introducing this but need to have a following ticket for the refactoring on purpose. Any other concerns @PhanLe1010 @joshimoo @shuo-wu ?

PhanLe1010:
I think we should delay this PR (unless there is a strong request from customers/users) and have an epic for volume auto attachment/detachment issues

innobead:
> I think we should delay this PR (unless there is a strong request from customers/users) and have an epic for volume auto attachment/detachment refactoring issues

As per the discussion, we agree with the consistent solution for auto volume attach/detach.

c3y1huang:
Close this first in favor of https://github.com/longhorn/longhorn/issues/3715.

PhanLe1010:
Update: we should have this capability now

PR #1038, opened 2021-09-13
Fix backup target status.available always false

jenting:
#### Proposal Change

If the user configures a valid backup target and the poll interval to 0 during the default engine image is not ready, the backup target `status.available=false` even the default engine image is ready.

Check the backup target available or not shouldn't depend on the poll interval setting (i.e., the `spec.syncRequestedAt` and `status.lastSyncedAt`).
Therefore, change the check of `spec.syncRequestedAt > status.lastSyncedAt` before listing remote backup volumes. (To list remote backup volumes or not depends on the poll interval setting)

#### Issues

https://github.com/longhorn/longhorn/issues/2960

### DIFF
--- a/datastore/longhorn.go
+++ b/datastore/longhorn.go
@@ -1059,7 +1059,11 @@ func (s *DataStore) CheckEngineImageReadiness(image string, nodes ...string) (is
 	if len(nodes) == 0 || (len(nodes) == 1 && nodes[0] == "") {
 		return false, nil
 	}
-	nodesHaveEngineImage, err := s.ListNodesWithEngineImage(image)
+	eiStates := map[types.EngineImageState]struct{}{
+		types.EngineImageStateDeploying: {},
+		types.EngineImageStateDeployed:  {},
+	}
+	nodesHaveEngineImage, err := s.ListNodesWithEngineImage(image, eiStates)
 	if err != nil {
 		return false, errors.Wrapf(err, "failed CheckEngineImageReadiness for nodes %v", nodes)
 	}
@@ -1767,7 +1771,7 @@ func (s *DataStore) ListNodesRO() ([]*longhorn.Node, error) {
 	return s.nLister.Nodes(s.namespace).List(labels.Everything())
 }
 
-func (s *DataStore) ListNodesWithEngineImage(image string) (map[string]*longhorn.Node, error) {
+func (s *DataStore) ListNodesWithEngineImage(image string, states map[types.EngineImageState]struct{}) (map[string]*longhorn.Node, error) {
 	nodes, err := s.ListNodes()
 	if err != nil {
 		return nil, err
@@ -1777,7 +1781,7 @@ func (s *DataStore) ListNodesWithEngineImage(image string) (map[string]*longhorn
 	if err != nil {
 		return nil, fmt.Errorf("unable to get engine image %v: %v", image, err)
 	}
-	if ei.Status.State != types.EngineImageStateDeployed && ei.Status.State != types.EngineImageStateDeploying {
+	if _, exist := states[ei.Status.State]; !exist {
 		return map[string]*longhorn.Node{}, nil
 	}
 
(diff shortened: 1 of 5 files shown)


shuo-wu (commented):
This can be a false positive: If the user sets an invalid backup target URL/secret with the poll interval keeping 0, will `backupTarget.Status.Available` be wrongly set to `true`?

Can we try the following solution instead?
1. The setting controller enqueues the backup target setting when the default engine image **becomes ready**.
2. The backup target sync function in the setting controller directly sets `spec.syncRequstedAt` for the **unavailable** backup target regardless of the poll interval. Of course, you can skip the field update when the gap between `spec.syncRequstedAt` and the current time is too short. Then the poll interval setting actually means the interval of updating backups **in a ready** backup target.

jenting:
I'd like to go this way. Currently the backup_target_controller/backup_volume_controller/backup_controller run the reconcile loop when there is a ready node with engine image (state in `deploying` or `deployed`).
But since these three controllers depend on the default engine image, so without a ready default engine image, the controller should not keep running the reconcile loop. Therefore, I think we should change the logic here:
- https://github.com/longhorn/longhorn-manager/blob/v1.2.0/controller/backup_target_controller.go#L370-L377
- https://github.com/longhorn/longhorn-manager/blob/v1.2.0/controller/backup_volume_controller.go#L386-L393
- https://github.com/longhorn/longhorn-manager/blob/v1.2.0/controller/backup_controller.go#L358-L365

To filter the nodes which are `ready` and its default engine image is in `ready` state. The code snippet is like this
```go
        readyNodesWithReadyEI, err := btc.ds.ListReadyNodesWithReadyEngineImage(defaultEngineImage)
	if err != nil {
		return false, err
	}
	// No node in the system has the default engine image in ready state
	if len(readyNodesWithReadyEI) == 0 {
		return false, nil
	}
```

After that, if the user configures a valid backup target setting and the engine image is not ready yet, we should not update the BackupTarget `status.available`.

And we should add an EngineImageInformer to enqueue default BackupTarget in backup_target_controller when the default engine image state becomes `ready`.

datastore/longhorn.go
```diff
@@ -1767,7 +1771,7 @@ func (s *DataStore) ListNodesRO() ([]*longhorn.Node, error) {
 	return s.nLister.Nodes(s.namespace).List(labels.Everything())
 }
 
-func (s *DataStore) ListNodesWithEngineImage(image string) (map[string]*longhorn.Node, error) {
+func (s *DataStore) ListNodesWithEngineImage(image string, states map[types.EngineImageState]struct{}) (map[string]*longhorn.Node, error) {
```
shuo-wu, line 1774: Just notice that this function looks weird. I think we can directly input an EngineImage CR and filter the CR outside of this function. This function is only responsible for picking up all nodes contain this EngineImage.

shuo-wu (commented):
I think we need to discuss the following 2 questions first:
1. Is the final goal of this PR to guarantee that the backup target availability detection should not be affected by Longhorn settings (Poll interval) or Longhorn interval state (EngineImage state)? 
2. Should the poll interval control the backup target availability detection?

jenting:
> I think we need to discuss the following 2 questions first:
> 
> 1. Is the final goal of this PR to guarantee that the backup target availability detection should not be affected by Longhorn settings (Poll interval) or Longhorn interval state (EngineImage state)?
> 2. Should the poll interval control the backup target availability detection?

In general, the availability should not be affected by the poll interval.

For the engine image state, if the engine image is not ready to use, it means the backup target client can't access the S3/NFS even the user configures a valid URL + credential. From the user's point of view, the URL + credential is valid. But from Longhorn's point of view (as a client), the backup target is invalid since it can't reach the remote backup target.

I'd say the backup target availability definition is the Longhorn client is able to reach the remote backup target at the moment (valid S3/NFS URL/Credential + a ready engine image in one of the nodes + have permission to list remote backup target).

WDYT?

shuo-wu:
How about the case that a backup target server suddenly gets recovered while the poll interval is 0? Should we detect the backup target server and update the status in time?

shuo-wu:
BTW, based on the modification, the existing owner ID of backup-related CRs will not be cleaned up when there is a default engine image crash. But actually, the corresponding controller won't be responsible for reconciling the CRs at this moment, which may confuse users.

jenting:
> The backup target availability you defined is affected by the Longhorn current status only, which is irrelevant to the backup target server. Then how about the case that a backup target server suddenly gets recovered while the poll interval is 0? Should we detect the backup target server and update the status in time?

I think we should, but the current code implementation can't detect this situation.
Let me think about how to tackle it.

jenting:
> BTW, based on the modification, the existing owner ID of backup-related CRs will not be cleaned up when there is a default engine image crash. But actually, the corresponding controller won't be responsible for reconciling the CRs at this moment, which may confuse users.

If all the node's engine images crash, the controller can't proceed to pull the remote backup target. Unless one of the node's engine image comes back. I don't think _why_ we should clean up the backup-related at this moment?

shuo-wu:
> If all the node's engine images crash, the controller can't proceed to pull the remote backup target. Unless one of the node's engine image comes back. I don't think _why_ we should clean up the backup-related at this moment?

The **owner recorded in the CR status** won't reconcile the CRs in this case. It means it didn't work as expected. But it's not a big issue. We can ignore it now.

shuo-wu:
> I think we should, but the current code implementation can't detect this situation.
> Let me think about how to tackle it 🤔

Maybe we can tolerate this case. Otherwise, we need to consider the counter case as well:
The backup target server suddenly becomes unavailable while the poll interval is 0.

jenting:
So, what're the left items we need to tackle for this PR/issue?

shuo-wu:
If the backup target is marked as unavailable (and `lastSyncedAt` is later than `requestedSyncedAt`) due to the engine image not being ready, Longhorn should resync it after the system upgrade.

jenting:
> If the backup target is marked as unavailable (and `lastSyncedAt` is later than `requestedSyncedAt`) due to the engine image not being ready, Longhorn should resync it after the system upgrade.

I've tested the system upgrade case. Upgrade from the Longhorn v1.2.0 to `longhorn-engine:master-head` and `longhorn-manager:this-pr`.

After upgrade, the backup_target_controller runs reconcile (the `spec.requestedSyncedAt` would be triggered, also the `status.lastSyncedAt` be updated).
I haven't check which part code trigger it (probably one of the informer in setting_controller _or_ one of the informer in backup_target_controller).

So, with longhorn v1.2.0 + `poll interval = 0` + `available = false`, perform system upgrade, in the end the backup target `availabe=true`. So, we don't have to handle it in the upgrade path.

shuo-wu (approved):
LGTM

innobead:
@jenting Please backport to 1.2.x.

PR #1040, opened 2021-09-14
Introduce a new setting to limit per node concurrent rebuilding count

shuo-wu:
longhorn/longhorn#2460

### DIFF
--- a/controller/replica_controller.go
+++ b/controller/replica_controller.go
@@ -63,6 +63,7 @@ type ReplicaController struct {
 	rStoreSynced  cache.InformerSynced
 	imStoreSynced cache.InformerSynced
 	biStoreSynced cache.InformerSynced
+	sStoreSynced  cache.InformerSynced
 
 	instanceHandler *InstanceHandler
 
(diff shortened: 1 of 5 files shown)


innobead:
@jenting please also help review if have time.

controller/replica_controller.go
```diff
@@ -420,6 +408,80 @@ func (rc *ReplicaController) CreateInstance(obj interface{}) (*types.InstancePro
...
+			r.Name, bi.Name, r.Spec.NodeID, r.Spec.DiskID, currentBackingImageState)
+		return "", nil
+	}
+	return types.GetBackingImagePathForReplicaManagerContainer(r.Spec.DiskPath, r.Spec.BackingImage, bi.Status.UUID), nil
+}
+
+func (rc *ReplicaController) CanStartRebuildingReplica(r *longhorn.Replica) (bool, error) {
+	log := getLoggerForReplica(rc.logger, r)
+
+	concurrentRebuildingLimit, err := rc.ds.GetSettingAsInt(types.SettingNameConcurrentReplicaRebuildPerNodeLimit)
```
innobead, line 451: What's the default setting for `concurrentRebuildingLimit`? I assume it will be `0`.

If it's the default setting, probably we just skip the following check and return `true` to allow the rebuild process.

innobead (changes requested):
In general, LGTM. But just one comment needs to resolve.

controller/replica_controller.go
shuo-wu: OK. I will update the default value to 0 to avoid regression.

innobead (approved):
LGTM

innobead:
@shuo-wu Please update manifests in longhorn/longhorn, longhorn/longhorn-manager.

PR #1044, opened 2021-09-15
Enforce overprovisioning limit when expanding volume size and fix the rounding issue with Storage Over Provisioning Percentage

PhanLe1010:
longhorn/longhorn#2962
longhorn/longhorn#2952

### DIFF
--- a/csi/controller_server.go
+++ b/csi/controller_server.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"reflect"
+	"regexp"
 	"strconv"
 	"strings"
 	"time"
@@ -747,6 +748,11 @@ func (cs *ControllerServer) ControllerExpandVolume(ctx context.Context, req *csi
 	if existVol, err = cs.apiClient.Volume.ActionExpand(existVol, &longhornclient.ExpandInput{
 		Size: strconv.FormatInt(requestedSize, 10),
 	}); err != nil {
+		// TODO: This manual error code parsing should be refactored once Longhorn API implements error code response
+		// https://github.com/longhorn/longhorn/issues/1875
+		if matched, _ := regexp.MatchString("cannot schedule .* more bytes to disk", err.Error()); matched {
+			return nil, status.Errorf(codes.OutOfRange, err.Error())
+		}
 		return nil, status.Errorf(codes.Internal, err.Error())
 	}
 
--- a/scheduler/replica_scheduler.go
+++ b/scheduler/replica_scheduler.go
@@ -5,6 +5,7 @@ import (
 	"sort"
 	"time"
 
+	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 
 	"github.com/longhorn/longhorn-manager/datastore"
@@ -545,8 +546,8 @@ func (rcs *ReplicaScheduler) IsSchedulableToDisk(size int64, requiredStorage int
 	// StorageAvailable = the space can be used by 3rd party or Longhorn system.
 	// There is no (direct) relationship between StorageReserved and StorageAvailable.
 	return info.StorageMaximum > 0 && info.StorageAvailable > 0 &&
-		info.StorageAvailable-requiredStorage > info.StorageMaximum*info.MinimalAvailablePercentage/100 &&
-		(size+info.StorageScheduled) <= (info.StorageMaximum-info.StorageReserved)*(info.OverProvisioningPercentage/100)
+		info.StorageAvailable-requiredStorage > int64(float64(info.StorageMaximum)*float64(info.MinimalAvailablePercentage)/100) &&
+		(size+info.StorageScheduled) <= int64(float64(info.StorageMaximum-info.StorageReserved)*float64(info.OverProvisioningPercentage)/100)
 }
 
 func (rcs *ReplicaScheduler) GetDiskSchedulingInfo(disk types.DiskSpec, diskStatus *types.DiskStatus) (*DiskSchedulingInfo, error) {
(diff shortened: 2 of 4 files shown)


scheduler/replica_scheduler.go
```diff
@@ -569,3 +570,55 @@ func (rcs *ReplicaScheduler) GetDiskSchedulingInfo(disk types.DiskSpec, diskStat
...
+		requestingSizeExpansion := delSize * diskIDToReplicaCount[diskID]
+		if !rcs.IsSchedulableToDisk(requestingSizeExpansion, 0, diskInfo) {
+			logrus.Errorf("Cannot expand replica size from %v to %v on disk %v with %+v", v.Spec.Size, newSize, diskID, diskInfo)
+			return fmt.Errorf("cannot expand replica size from %v to %v on disk %v with %+v ", v.Spec.Size, newSize, diskID, diskInfo)
+		}
+	}
+	return nil
+}
+
+func findDiskSpecAndDiskStatusInNode(diskID string, node *longhorn.Node) (types.DiskSpec, types.DiskStatus, bool) {
```
shuo-wu, line 616: NIT: To avoid confusion, please use `diskUUID` here and name the disk map key as `diskName`.

shuo-wu (commented):
The verification itself LGTM. Do we need to add the expansion size check in the volume controller? The newly added check in API layer cannot prevent direct volume spec size modification from breaking the scheduling rules.

innobead (dismissed):
LGTM

@shuo-wu need your knowledge: I am curious how confident if users update the volume spec directly, all volume functions should work? I remember we don't suggest users update the volume spec directly until it gets supported well? (this is a side question actually)

jenting (commented):
Could we add/update the unit test to cover this case?

scheduler/replica_scheduler.go
PhanLe1010: Thank you! I fixed the names

Like @innobead mentioned, I think we assume that users don't mess up with the CRs directly otherwise we would have a lot of problems not just with volume expansion.

PhanLe1010:
I looked into the unit test, I might be wrong but it seems we have to create a new unit test and it takes time. Can we create a new ticket for the unit test and implement it later because it is not top priority? How do you think @jenting @innobead?

innobead:
> I looked into the unit test, I might be wrong but it seems we have to create a new unit test and it takes time. Can we create a new ticket for the unit test and implement it later because it is not top priority? How do you think @jenting @innobead?

NP. Let’s do this way.

csi/controller_server.go
```diff
@@ -747,6 +748,11 @@ func (cs *ControllerServer) ControllerExpandVolume(ctx context.Context, req *csi
 	if existVol, err = cs.apiClient.Volume.ActionExpand(existVol, &longhornclient.ExpandInput{
 		Size: strconv.FormatInt(requestedSize, 10),
 	}); err != nil {
+		// TODO: This manual error code parsing should be refactored once Longhorn API implements error code response
+		// https://github.com/longhorn/longhorn/issues/1875
+		if matched, _ := regexp.MatchString("cannot schedule .* more bytes to disk", err.Error()); matched {
+			return nil, status.Errorf(codes.OutOfRange, err.Error())
+		}
```
PhanLe1010, line 755: @joshimoo  Can you take a look at this error code parsing as we discussed in the PR review meeting?

innobead:
We need to backport this to 1.2.x

shuo-wu:
> I am curious how confident if users update the volume spec directly, all volume functions should work? I remember we don't suggest users update the volume spec directly until it gets supported well? (this is a side question actually)

IIRC, Harvester will operate the Longhorn system via `kubectl` rather than sending HTTP requests. Although Longhorn doesn't officially support direct CR operation, we should start to consider this part and try to handle it from now on.

For this PR, I think it's nice to consider this part. It is not the top priority.

PhanLe1010:
Thanks @shuo-wu 

I added the check to volume controller. Need to do more testing before merging

PhanLe1010:
BTW, I just made a common mistake when edit the volume CR. I edit the volume size to a value that is not multiple of 4Kb. The volume stuck in attaching/detaching loop. This is one example of many things could go wrong if user directly modify the volume CR.

PhanLe1010:
Why Harvester doesn't use Longhorn API by the way?

innobead:
@PhanLe1010 I believe they understand how it works, so they go with the spec setting like recent node management. But for now, let's make the function working first, we can tackle the CR part later as @shuo-wu said not a priority for now.

csi/controller_server.go
joshimoo: I think this should be fine, maybe a little overkill with the regex but since it's one off why not :)

scheduler/replica_scheduler.go
```diff
@@ -569,3 +570,55 @@ func (rcs *ReplicaScheduler) GetDiskSchedulingInfo(disk types.DiskSpec, diskStat
...
+		}
+		diskInfo, err := rcs.GetDiskSchedulingInfo(diskSpec, &diskStatus)
+		if err != nil {
+			return fmt.Errorf("failed to GetDiskSchedulingInfo %v", err)
+		}
+		diskIDToDiskInfo[r.Spec.DiskID] = diskInfo
+		diskIDToReplicaCount[r.Spec.DiskID] = diskIDToReplicaCount[r.Spec.DiskID] + 1
+	}
+
+	delSize := newSize - oldSize
```
joshimoo, line 605: NIT: delSize could be deltaSize or deletionSize, I feel like this is good spot to be a little bit more descriptive :)

joshimoo (approved):
LGTM

joshimoo:
Harvester shouldn't need to modify the longhorn CRD directly for size operations instead they should be utilizing a PVC in block mode then updating the size there. 

For the future to prevent modifications that could lead to invalid states, we would need a mutating/validating admission hook.
Which will be a good addition for a couple of our CRD's.

scheduler/replica_scheduler.go
shuo-wu: `expandedSize` may be more precise.

shuo-wu:
> Why Harvester doesn't use Longhorn API by the way?

I am not sure what the exact reason is. Maybe we can test this question in the Slack channel. Using `kubectl` is the most popular & native way to operate Kubernetes components. In the future, we can add some validation for `kubectl` to avoid this. kind of invalid size update.

scheduler/replica_scheduler.go
PhanLe1010: Thank you! Fixed

scheduler/replica_scheduler.go
```diff
@@ -569,3 +570,55 @@ func (rcs *ReplicaScheduler) GetDiskSchedulingInfo(disk types.DiskSpec, diskStat
...
+		if err != nil {
+			return fmt.Errorf("failed to GetDiskSchedulingInfo %v", err)
+		}
+		diskIDToDiskInfo[r.Spec.DiskID] = diskInfo
+		diskIDToReplicaCount[r.Spec.DiskID] = diskIDToReplicaCount[r.Spec.DiskID] + 1
+	}
+
+	delSize := newSize - v.Spec.Size
+	for diskID, diskInfo := range diskIDToDiskInfo {
+		requestingSizeExpansion := delSize * diskIDToReplicaCount[diskID]
```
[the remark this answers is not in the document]
PhanLe1010: I changed the name. 
If there are 2 replicas of the volume on the same disk. We need `2*the-expanding-size` more space

PhanLe1010:
e2e core tests looks ok

innobead (approved):

innobead:
@PhanLe1010 please backport to 1.2.1 and 1.1.3.

PR #1045, opened 2021-09-15
Change to millisecond-level precision

jenting:
Address PR comment https://github.com/longhorn/longhorn-manager/pull/1009#issuecomment-906903691

Change the BackupTarget/BackupVolume/Backup CR from second-level precision to millisecond-level precision.
It fixes if more than two controllers set the `spec.syncRequestedAt` within a short time (within 1 second), the second one would not be handled.

### DIFF
--- a/controller/backup_controller.go
+++ b/controller/backup_controller.go
@@ -12,7 +12,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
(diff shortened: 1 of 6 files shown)


innobead (commented):
LGTM, but one thing needed to clarify for me. 

~Because you change the spec, will this impact the existing data when deserializing back to the resource w/ new spec?~

```
type Time struct {
	time.Time `protobuf:"-"`
}
```
Ok, it's embedded structure in `metav1.Time`, also actually the same as `time.Time`. It should be fine, but it's better to double confirm to avid any side effect after upgrade.

innobead (approved):

shuo-wu (commented):
What's the (string) format of `time.Now().UTC()`? After the system upgrade, does the old timestamp still exist?

jenting:
> What's the (string) format of `time.Now().UTC()`? After the system upgrade, does the old timestamp still exist?

- metav1.Time string format, for example: `2021-09-15T09:58:12Z`
- time.Time string format, for example: `2021-09-15T09:58:12.64046605Z`

The old timestamp would exist if the CR did not get updated. For example, the Backup CR won't be updated once it has been pulled from the remote backup target. So the `spec.syncRequestedAt` and `status.lastSyncedAt` will keep the second-level precision (for example: `2021-09-15T09:58:12Z`).

shuo-wu (approved):

PR #1057, opened 2021-09-18
Notify users about latest stable version of every minor release line

PhanLe1010:
longhorn/longhorn#3014

### DIFF
--- a/controller/setting_controller.go
+++ b/controller/setting_controller.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"reflect"
 	"sort"
+	"strings"
 	"time"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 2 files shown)


innobead (approved):
LGTM

innobead:
@PhanLe1010 because you introduce a new setting, please also update the manifests and charts of longhorn/longhorn, longhron/longhorn-manager.

PhanLe1010:
@innobead 
This setting is a hidden system setting. I think we don't need to modify any chart similar to the `latest version` setting

innobead:
> @innobead
> This setting is a hidden system setting. I think we don't need to modify any chart similar to the `latest version` setting

Thanks for the clarification!

PR #1058, opened 2021-09-21
upgrade: Avoid CR update if there is nothing change

shuo-wu:
This is a surgery fix for the original issue, which can not avoid
the old upgrade path intervening longhorn manager pod start.

To resolve this issue, Longhorn needs to somehow understand the
old version and the current version during system upgrade.

Longhorn/longhorn#2582

### DIFF
--- a/upgrade/v102to110/upgrade.go
+++ b/upgrade/v102to110/upgrade.go
@@ -122,7 +122,7 @@ func upgradeVolumes(namespace string, lhClient *lhclientset.Clientset) (err erro
 			v = *updatedVolume
 		}
 
-		if v.Status.Robustness == types.VolumeRobustnessDegraded {
+		if v.Status.Robustness == types.VolumeRobustnessDegraded && v.Status.LastDegradedAt == "" {
 			v.Status.LastDegradedAt = util.Now()
 			if _, err := lhClient.LonghornV1beta1().Volumes(namespace).UpdateStatus(context.TODO(), &v, metav1.UpdateOptions{}); err != nil {
 				return err


shuo-wu:
Since it's hard to reproduce the bug, I am not sure if we can find a way to test this fix.

innobead (approved):

upgrade/v102to110/upgrade.go
```diff
@@ -113,17 +113,16 @@ func upgradeVolumes(namespace string, lhClient *lhclientset.Clientset) (err erro
...
+			requireUpdate = true
 		}
-
-		if v.Status.Robustness == types.VolumeRobustnessDegraded {
+		if v.Status.Robustness == types.VolumeRobustnessDegraded && v.Status.LastDegradedAt == "" {
 			v.Status.LastDegradedAt = util.Now()
+			requireUpdate = true
+		}
+		if requireUpdate {
 			if _, err := lhClient.LonghornV1beta1().Volumes(namespace).UpdateStatus(context.TODO(), &v, metav1.UpdateOptions{}); err != nil {
```
innobead, line 126: When `requireUpdate` is true, you only update the volume status but actually, but how about the change of `v.spec.accessmode` which will not be taken effect if only updating the status.

innobead (changes requested):

innobead:
I am also curious a bit about this change which corner cases will this PR actually fix?

innobead:
> I am also curious a bit about this change which corner cases will this PR actually fix?

This is from @shuo-wu 
> The issue mentioned in the ticket are caused by this blind volume update conflict.
> E.g., when users upgrade the Longhorn from v1.1.0 to v1.1.x or v1.2.0, this volume update in the old upgrade path should not do anything But actually NO. Since this is neither deep equal check nor some flags like requireUpdate that indicate the update skip.

innobead (approved):

innobead:
@shuo-wu please backport to 1.2.x.

PR #1063, opened 2021-09-22
Deprecate setting 'disable-replica-rebuild' and merge it into the new setting

shuo-wu:
longhorn/longhorn#2460

### DIFF
--- a/upgrade/v120to121/upgrade.go
+++ b/upgrade/v120to121/upgrade.go
@@ -1,0 +1,77 @@
+package v120to121
+
+import (
+	"context"
+	"strconv"
+
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/longhorn/longhorn-manager/types"
+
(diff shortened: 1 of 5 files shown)


upgrade/v120to121/upgrade.go
```diff
@@ -0,0 +1,77 @@
...
+			return nil
+		}
+		return err
+	}
+	disabled, err := strconv.ParseBool(disableReplicaRebuildSetting.Value)
+	if err != nil {
+		logrus.Warnf(upgradeLogPrefix+"Failed to parse boolean value for setting %v, will skip checking it, error: %v", types.SettingNameDisableReplicaRebuild, err)
+		return nil
+	}
+	if disabled {
```
innobead, line 50: This should be `if !disabled`. You need to skip the following concurrent setting update if the disabled is false.

innobead (changes requested):

upgrade/v120to121/upgrade.go
shuo-wu: Yes. If this deprecated setting zs not modified, the concurrent setting should be the default value or the customized value.

innobead (approved):

innobead:
@shuo-wu Please backport to 1.2.x. Thanks.

PR #1064, opened 2021-09-22
add fs expansion support to the csi driver

joshimoo:
We use the CSI driver to expand the volume filesystem after the mount during NodeStageVolume
This is required to correctly support kubernetes cloned volumes, since NodeExpandVolume is not called after
a succesful clone of a volume with a requested bigger size.
ref: https://github.com/kubernetes/kubernetes/issues/94929

NodeExpandVolume is a mixed bag at this point probably the only reason it still exists, is for supporting online expansions
of already published volumes, which we do not support. So I left out the implementation of it in favor of the NodeStageVolume expansion pass.
ref: https://kubernetes-csi.github.io/docs/kubernetes-changelog.html#kubernetes-119

Besides the expansion I also removed the nfs/nsenter code from the csi driver instead I use a small shell script that gets called by the regular mounter, this is similiar to googles containerized mounter. And can easily be replaced with a small go program instead of my quick shell script.

https://github.com/kubernetes/kubernetes/tree/release-1.22/cluster/gce/gci/mounter

longhorn/longhorn#2794
longhorn/longhorn#2868


csi/node_server.go
```diff
@@ -386,6 +386,28 @@ func (ns *NodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
...
+	// check if we need to resize the fs
+	// this is important since cloned volumes of bigger size don't trigger NodeExpandVolume
+	// therefore NodeExpandVolume is kind of redundant since we have to do this anyway
+	// some refs below for more details
+	// https://github.com/kubernetes/kubernetes/issues/94929
+	// https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/753
+	resizer := mount.NewResizeFs(utilexec.New())
+	if needsResize, err := resizer.NeedResize(devicePath, targetPath); err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	} else if needsResize {
```
shuo-wu, line 398: NIT:
```suggestion
	needsResize, err := resizer.NeedResize(devicePath, targetPath); 
	if err != nil {
		return nil, status.Error(codes.Internal, err.Error())
	} 
	if needsResize {
```

Similar below

csi/node_server.go
```diff
@@ -386,6 +386,28 @@ func (ns *NodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
 		return nil, err
 	}
 
+	// check if we need to resize the fs
+	// this is important since cloned volumes of bigger size don't trigger NodeExpandVolume
+	// therefore NodeExpandVolume is kind of redundant since we have to do this anyway
+	// some refs below for more details
+	// https://github.com/kubernetes/kubernetes/issues/94929
+	// https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/753
+	resizer := mount.NewResizeFs(utilexec.New())
```
shuo-wu, line 395: Does the CSI need to skip this check for `nfs`? Based on the ref links you provide, it seems that the util function does not handle `nfs`. 
And we don't need to do anything on the NFS client side for the expansion. Expanding the underlying filesystem of the NFS server work directory is enough, right?

joshimoo: Correct, the share-manager is responsible for the filesystem expansion of the RWO volume that is exported via NFS.
NFS mount is handled at line 324 above this code path is only executed for regular RWO mount volumes. https://github.com/longhorn/longhorn-manager/pull/1064/files/3524e239cd7356171c93b7efe0f14d724b3ed16d..4fd4a3fbd27f9d622ac5a54b75cec31e1d3294dc#diff-e3208de2b41b1544cc85ea18e65fb55e847951129cdeee7db877af639c08e584R324

csi/node_server.go
joshimoo: I find the existing one more readable, it also reduces the scope of the variables.
But both approaches are fine, I tend to prefer the compacter code when the inner scope only has a couple of lines.

shuo-wu: OK. The suggestions I provide reduces the nesting. But in general both are fine. Let me retain the current implementation.

csi/node_server.go
shuo-wu: Got it. I forgot to check the unchanged code hidden above.

shuo-wu (approved):

shuo-wu:
Remember to update the default image for the share manager.

PR #1066, opened 2021-09-23
controller: live upgrade should not continue checking the old replica count while the new replicas are activated

shuo-wu:
Longhorn/longhorn#3052

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -2253,10 +2253,15 @@ func (vc *VolumeController) upgradeEngineForVolume(v *longhorn.Volume, e *longho
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


joshimoo:
Just checking did you test this with live migration? Since I think it uses the same createMatching function.

shuo-wu:
> Just checking did you test this with live migration? Since I think it uses the same createMatching function.

YES. I just have quick live migration tests for degraded & healthy volumes. It works fine. And just in case, I add an additional check in the migration function.

innobead (approved):
LGTM

joshimoo (approved):
LGTM, if I understand correctly it means we cannot do live upgrade / volume migration when a volume is in degraded state.

PR #1067, opened 2021-09-23
[Backport][v1.2.x]controller: live upgrade should not continue checking the old replica count while the new replicas are activated

shuo-wu:
Longhorn/longhorn#3052


joshimoo (approved):
LGTM

PR #1069, opened 2021-09-24
Fix the concurrent rebuild limit not working issue

shuo-wu:
longhorn/longhorn#2460

### DIFF
--- a/controller/replica_controller.go
+++ b/controller/replica_controller.go
@@ -67,7 +67,8 @@ type ReplicaController struct {
 
 	instanceHandler *InstanceHandler
 
-	rebuildingLock *sync.Mutex
+	rebuildingLock          *sync.Mutex
+	inProgressRebuildingMap map[string]struct{}
 }
 
 func NewReplicaController(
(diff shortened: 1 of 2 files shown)


controller/replica_controller.go
```diff
@@ -67,7 +67,8 @@ type ReplicaController struct {
 
 	instanceHandler *InstanceHandler
 
-	rebuildingLock *sync.Mutex
+	rebuildingLock          *sync.Mutex
+	inProgressRebuildingMap map[string]struct{}
```
joshimoo, line 71: NIT: you can use map[string]interface as well then just push nil. If you only care about the key and not the value.
Both approaches are fine :)

controller/replica_controller.go
```diff
@@ -98,7 +99,8 @@ func NewReplicaController(
 
 		ds: ds,
 
-		rebuildingLock: &sync.Mutex{},
+		rebuildingLock:          &sync.Mutex{},
+		inProgressRebuildingMap: map[string]struct{}{},
```
joshimoo, line 103: NIT: I would probably use `make(map[string]ValueType)` instead. The inline initialization is nice if you immediately add values. But again not important :)

innobead (approved):
LGTM

shuo-wu:
@joshimoo Thank you for your suggestions. I will revisit them next week.

PR #1071, opened 2021-09-27
upgrade: Check if the label already exists before setting it for a volume

shuo-wu:
Longhorn/longhorn#3068

### DIFF
--- a/upgrade/v111to120/upgrade.go
+++ b/upgrade/v111to120/upgrade.go
@@ -209,6 +209,9 @@ func upgradeLabelsForVolume(v *longhorn.Volume, lhClient *lhclientset.Clientset,
 	if labels == nil {
 		labels = map[string]string{}
 	}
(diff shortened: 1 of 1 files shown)


shuo-wu:
@jenting Actually I have a questions here: Will the backup volume label here be removed when the restore/DR volumes are activated? If No, will the label confuse users in the future?

jenting:
> @jenting Actually I have a question here: Will the backup volume label here be removed when the restore/DR volumes are activated? If No, will the label confuse users in the future?

No, I did not remove the label after the restore/DR volume is activated.

I was think that the label was like history to know that this volume was a restore/DR volume.
If we want a state of art, we should remove the label.

shuo-wu:
> If we want a state of art, we should remove the label.

Yeah. It's nice to have. We can consider this later.

innobead (approved):
LGTM

PR #1074, opened 2021-09-28
Fixed the log about Longhorn CSI plugin version

PhanLe1010:
longhorn/longhorn#3079

### DIFF
--- a/app/csi.go
+++ b/app/csi.go
@@ -31,11 +31,6 @@ func CSICommand() cli.Command {
 				Value: "",
 				Usage: "Longhorn manager API URL",
 			},
-			cli.StringFlag{
-				Name:  "csi-version",
-				Value: "0.3.0",
-				Usage: "CSI plugin version",
-			},
 		},
 		Action: func(c *cli.Context) {
 			if err := runCSI(c); err != nil {
@@ -51,7 +46,6 @@ func runCSI(c *cli.Context) error {
 	return manager.Run(c.String("drivername"),
 		c.String("nodeid"),
 		c.String("endpoint"),
-		c.String("csi-version"),
 		identityVersion,
 		c.String("manager-url"))
 }
--- a/csi/manager.go
+++ b/csi/manager.go
@@ -19,8 +19,8 @@ func GetCSIManager() *Manager {
 	return &Manager{}
 }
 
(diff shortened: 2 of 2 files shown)


jenting (approved):
LGTM

app/csi.go
```diff
@@ -51,7 +46,6 @@ func runCSI(c *cli.Context) error {
 	return manager.Run(c.String("drivername"),
 		c.String("nodeid"),
 		c.String("endpoint"),
-		c.String("csi-version"),
 		identityVersion,
```
shuo-wu, line 49: This version is always `dev`. Do we need to retain it?

PhanLe1010: I think it is the same as the version in Longhorn manager image (e.g., v1.1.2, v1.2.0, v1.2.1-rc2)

PhanLe1010: I guess dapper is doing some magic here to set it https://github.com/longhorn/longhorn-manager/blob/532d53d1f1d1484abfff0e1e8a440d5bb18798bf/Dockerfile.dapper#L30

shuo-wu: OK. You are right. Then this field looks good.

shuo-wu (approved):

csi/manager.go
```diff
@@ -19,8 +19,8 @@ func GetCSIManager() *Manager {
 	return &Manager{}
 }
 
-func (m *Manager) Run(driverName, nodeID, endpoint, csiVersion, identityVersion, managerURL string) error {
-	logrus.Infof("CSI Driver: %v csiVersion: %v, manager URL %v", driverName, csiVersion, managerURL)
+func (m *Manager) Run(driverName, nodeID, endpoint, identityVersion, managerURL string) error {
+	logrus.Infof("CSI Driver: %v csiVersion: %v, manager URL %v", driverName, identityVersion, managerURL)
```
joshimoo, line 23: I thought I had reviewed this one as well, must have been only the backports. I would like the csiVersion to just be named version, since it's actually the longhorn version which is unrelated to the supported csi version

joshimoo (commented):
LGTM, see comment about csiVersion -> version naming.

PhanLe1010:
Thanks @joshimoo I pushed a new PR since this one has been merged 

https://github.com/longhorn/longhorn-manager/pull/1077

PR #1075, opened 2021-09-28
[Backport: v1.1.3] Fixed the log about Longhorn CSI plugin version

PhanLe1010:
longhorn/longhorn#3079

### DIFF
--- a/csi/manager.go
+++ b/csi/manager.go
@@ -19,8 +19,8 @@ func GetCSIManager() *Manager {
 	return &Manager{}
 }
 
-func (m *Manager) Run(driverName, nodeID, endpoint, csiVersion, identityVersion, managerURL string) error {
-	logrus.Infof("CSI Driver: %v csiVersion: %v, manager URL %v", driverName, csiVersion, managerURL)
+func (m *Manager) Run(driverName, nodeID, endpoint, identityVersion, managerURL string) error {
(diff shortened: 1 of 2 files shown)


csi/manager.go
```diff
@@ -19,8 +19,8 @@ func GetCSIManager() *Manager {
 	return &Manager{}
 }
 
-func (m *Manager) Run(driverName, nodeID, endpoint, csiVersion, identityVersion, managerURL string) error {
-	logrus.Infof("CSI Driver: %v csiVersion: %v, manager URL %v", driverName, csiVersion, managerURL)
+func (m *Manager) Run(driverName, nodeID, endpoint, identityVersion, managerURL string) error {
+	logrus.Infof("CSI Driver: %v csiVersion: %v, manager URL %v", driverName, identityVersion, managerURL)
```
joshimoo, line 23:
```suggestion
	logrus.Infof("CSI Driver: %v version: %v, manager URL %v", driverName, identityVersion, managerURL)
```

joshimoo: Change it to just version, since it's not a the specific csi version instead it's the drivers version.

joshimoo (dismissed):
LGTM, change the comment though to make it clear that it's the drivers version.

csi/manager.go
PhanLe1010: Thanks @joshimoo Fixed it

innobead (approved):

PR #1076, opened 2021-09-29
Recurring snapshot/backup job supports imagePullSecrets

jenting:
https://github.com/longhorn/longhorn/issues/2432

### DIFF
--- a/controller/recurring_job_controller.go
+++ b/controller/recurring_job_controller.go
@@ -312,6 +312,12 @@ func (control *RecurringJobController) newCronJob(recurringJob *longhorn.Recurri
 	if err != nil {
 		return nil, err
 	}
+	registrySecretSetting, err := control.ds.GetSetting(types.SettingNameRegistrySecret)
+	if err != nil {
+		return nil, err
+	}
+	registrySecret := registrySecretSetting.Value
+
 	// for mounting inside container
 	cronJob := &batchv1beta1.CronJob{
 		ObjectMeta: metav1.ObjectMeta{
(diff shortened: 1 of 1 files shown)


controller/recurring_job_controller.go
```diff
@@ -377,6 +381,11 @@ func (control *RecurringJobController) newCronJob(recurringJob *longhorn.Recurri
 							Tolerations:        util.GetDistinctTolerations(tolerations),
 							NodeSelector:       nodeSelector,
 							PriorityClassName:  priorityClass.Value,
+							ImagePullSecrets: []corev1.LocalObjectReference{
+								{
+									Name: registrySecretSetting.Value,
```
innobead, line 386: If the setting is empty, what happens? should we ignore the empty case here?

jenting: Yes, we should. Thanks for the reminder.

shuo-wu (commented):
Should we check all old recurring job CRs in the upgrade path? have a correction function in this controller? Besides, there are recurring job CR creation `translateStorageClassRecurringJobs` in `v110tov120/upgrade.go`.

jenting:
> Should we check all old recurring job CRs in the upgrade path? have a correction function in this controller? Besides, there are recurring job CR creation `translateStorageClassRecurringJobs` in `v110tov120/upgrade.go`.

Probably no, because it means that the old Longhorn cluster recurring job did not work because Longhorn doesn't support `imagePullSecret`.
So I think the user did not use the private registry + private registry secret + recurring job in the end.

But it's nice to have to auto recreate the old recurring job.

shuo-wu:
Does it not work in v1.1.x as well? If No, we can skip considering the upgrade path. Otherwise, it's better to add it.

jenting:
The original issue reported is v1.1.0, so I believe it's not working.

shuo-wu (approved):

PR #1079, opened 2021-10-01
HOTFIX replicaset issues when IM crashes

joshimoo:
After evaluating different options for addressing this issue.
I decided to do a check on the staging path and on failure to do a manual NodeUnstageVolume and NodeStageVolume call as part of the NodePublishVolume call.

The issue is only present for replicasets since they bypass the NodeStage phase depending on UpdateStrategy.
I tested this patch set for:

- RWO/RWX volumes with replica sets / stateful sets
- I tested IM crashes/deletions
- I tested kubelet restarts

This fix does not work for RWO encrypyted volumes, I worked on fixes for the encrypted RWO case.
But it introduces to much risk and issues around the case where a replicaset has multiple pods on the same node so I left them out of this patch set.

longhorn/longhorn#3070

### DIFF
--- a/csi/node_server.go
+++ b/csi/node_server.go
@@ -118,6 +118,38 @@ func (ns *NodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 		return &csi.NodePublishVolumeResponse{}, nil
 	}
 
+	// we validate the staging path to make sure the global mount is still valid
+	if isMnt, err := ensureMountPoint(stagingPath, mounter); err != nil || !isMnt {
+		msg := fmt.Sprintf("NodePublishVolume: staging path is no longer valid for volume %v", volumeID)
+		logrus.Error(msg)
+
+		// HACK: normally when we return FailedPrecondition below kubelet should call NodeStageVolume again
+		//	but currently it does not, so we manually call NodeStageVolume to remount the block device globally
+		//	we currently don't reuse the previously mapped block device (major:minor) so the initial mount even after
+		//	reattachment of the longhorn block dev is no longer valid
+		logrus.Warnf("NodePublishVolume: calling NodeUnstageVolume for volume %v", volumeID)
+		_, _ = ns.NodeUnstageVolume(ctx, &csi.NodeUnstageVolumeRequest{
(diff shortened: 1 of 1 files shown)


csi/node_server.go
```diff
@@ -118,6 +118,38 @@ func (ns *NodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 		return &csi.NodePublishVolumeResponse{}, nil
 	}
 
+	// we validate the staging path to make sure the global mount is still valid
+	if isMnt, err := ensureMountPoint(stagingPath, mounter); err != nil || !isMnt {
```
shuo-wu, line 122: Quick question, except for RWO encrypted volumes, does the checks (including `ioutil.ReadDir(targetPath)`) in this function work for Block volumes? If YES, then this LGTM.

jenting (approved):
LGTM

csi/node_server.go
innobead: It seems like Block volume is handled by the above code not here, so it should not be impacted by this change?

innobead (approved):
LGTM

csi/node_server.go
joshimoo: Correct block volumes for the moment aren't handled by NodeStage phase and only handled by NodePublish phase. Since there are some inconsistencies in the documentation regardion NodeStage before for block volumes, I decided to skip implementation in NodeStage since anyway for block volumes we only create a mknod for the associated (major:minor)

PhanLe1010:
LGTM 

I am looking into Kubernetes to see why it doesn't retry nodeStage and why this problem is only happen with ReplicaSet

PhanLe1010:
Yeah, from the code perspective, it seems that Kubelet doesn't handle FailedPreCondition error: https://github.com/kubernetes/kubernetes/blob/04f747d09fd2f8fb486c90a85122c0cb7b2df782/pkg/kubelet/volumemanager/reconciler/reconciler.go#L202

This problem only happen with replica set because Kubernetes spin up the replacement pod while the old pod is still in terminating state. This make Kubelet doesn't issue the NodeStageVolume call

PR #1081, opened 2021-10-07
Fixed bug: restoring backup always point to the latest backup

PhanLe1010:
Set the `e.Spec.RequestedBackupRestore` to the correct backup in `v.Spec.FromBackup` for non-DR restoring volume

This fix should fix:
* First time restore from a correct backup
* Retry to restore from a correct backup

longhorn/longhorn#3111

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -2302,12 +2302,24 @@ func (vc *VolumeController) updateRequestedBackupForVolumeRestore(v *longhorn.Vo
 	if v.Spec.FromBackup == "" {
 		return nil
 	}
-	if !v.Status.RestoreRequired && !v.Status.IsStandby {
+
+	// If the volume is not restoring, skip setting e.Spec.RequestedBackupRestore
+	if !v.Status.RestoreRequired {
 		return nil
 	}
 
-	if v.Status.LastBackup != "" && v.Status.LastBackup != e.Spec.RequestedBackupRestore {
-		e.Spec.RequestedBackupRestore = v.Status.LastBackup
+	// For DR volume, we set RequestedBackupRestore to the LastBackup
+	if v.Status.IsStandby {
+		if v.Status.LastBackup != "" && v.Status.LastBackup != e.Spec.RequestedBackupRestore {
+			e.Spec.RequestedBackupRestore = v.Status.LastBackup
+		}
+		return nil
+	}
+
+	// For non-DR restoring volume, we set RequestedBackupRestore to the backup in v.Spec.FromBackup
+	if _, backupName, err := vc.getInfoFromBackupURL(v); err == nil && backupName != "" {
+		e.Spec.RequestedBackupRestore = backupName
+		return nil
 	}
 
 	return nil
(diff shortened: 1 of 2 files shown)


controller/volume_controller.go
```diff
@@ -2302,12 +2302,24 @@ func (vc *VolumeController) updateRequestedBackupForVolumeRestore(v *longhorn.Vo
...
 		e.Spec.RequestedBackupRestore = v.Status.LastBackup
+		return nil
+	}
+
+	// For non-DR restoring volume we set RequestedBackupRestore to the backup in v.Spec.FromBackup
+	if v.Spec.FromBackup != "" {
+		if _, backupName, err := vc.getInfoFromBackupURL(v); err == nil && backupName != "" {
+			e.Spec.RequestedBackupRestore = backupName
+			return nil
+		}
```
jenting, line 2322: 1. Will the DR volume fall into this case when/if the DR volume `spec.FromBackup` is not empty?
2. The code logic seems duplicate to https://github.com/longhorn/longhorn-manager/blob/2561e778468bca785c148b7bc700c7f46a2c723e/controller/volume_controller.go#L2678-L2684, should we add the code here?

controller/volume_controller.go
```diff
@@ -2302,12 +2302,24 @@ func (vc *VolumeController) updateRequestedBackupForVolumeRestore(v *longhorn.Vo
...
 
-	if v.Status.LastBackup != "" && v.Status.LastBackup != e.Spec.RequestedBackupRestore {
+	// For DR volume we set RequestedBackupRestore to the LastBackup
+	if v.Status.IsStandby && v.Status.LastBackup != "" && v.Status.LastBackup != e.Spec.RequestedBackupRestore {
 		e.Spec.RequestedBackupRestore = v.Status.LastBackup
+		return nil
+	}
+
+	// For non-DR restoring volume we set RequestedBackupRestore to the backup in v.Spec.FromBackup
+	if v.Spec.FromBackup != "" {
```
shuo-wu, line 2318: NIT: This field won't be empty. See line 2302 above.

innobead (dismissed):
LGTM

Only one nit from Shou is needed to rsolve but no hurts atm.

controller/volume_controller.go

controller/volume_controller.go
PhanLe1010: Thanks @shuo-wu and @jenting 

As Shuo mentioned in Slack, for the least amount of change, we can keep both the init part and this update part.

jenting: > 1. `spec.FromBackup` must be non-empty for DR/restore volume.

Okay, so the comment `For non-DR restoring volume we set RequestedBackupRestore to the backup in v.Spec.FromBackup` is a bit confused that even the DR volume might fall into this case.

shuo-wu: Looks good. For normal restore volume, use the backup recorded in `Spec.FromBackup`. For DR volumes, use the latest one instead.

innobead (approved):

jenting (approved):

PR #253, opened 2019-04-12
controller: speeded up disaster recovery by deleting VolumeAttachment in time

shuo-wu:
Feature 3: https://github.com/rancher/longhorn/issues/461

PR #1083, opened 2021-10-08
Refactor register informers flow

jenting:
#### Proposal Changes

- remove the initialization of the informers from StartControllers
- only pass the informer factories to the datastore
- the datastore create informers and exports the informers
- only pass the ds to the controllers
- all controllers ref the ds exposed informers to add event handlers
- all controllers add the informer.HasSynced to `cacheSyncs` array
- all controllers wait for cache sync by `cacheSyncs` array

The end result is that we only need to deal with:
- informer creation at `NewDatastore`, and add to `informer.HasSynced` to the `Sync` function.
- each controller adds event handlers to the informers and registers `informer.HasSynced` to the `cacheSyncs`.

### Issue
https://github.com/longhorn/longhorn/issues/2671

### DIFF
--- a/controller/kubernetes_pv_controller.go
+++ b/controller/kubernetes_pv_controller.go
@@ -91,12 +91,6 @@ func NewKubernetesPVController(
 	})
 	kc.cacheSyncs = append(kc.cacheSyncs, ds.PodInformer.HasSynced)
 
-	// after volume becomes detached, try to delete the VA of lost node
-	ds.VolumeInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
-		UpdateFunc: func(old, cur interface{}) { kc.enqueueVolumeChange(cur) },
-	})
-	kc.cacheSyncs = append(kc.cacheSyncs, ds.VolumeInformer.HasSynced)
-
 	return kc
 }
 


datastore/longhorn.go
```diff
@@ -1879,6 +1879,10 @@ func (s *DataStore) ListPodsRO() ([]*corev1.Pod, error) {
 	return s.pLister.Pods(s.namespace).List(labels.Everything())
 }
 
+func (s *DataStore) GetPodRO(namespace, name string) (*corev1.Pod, error) {
```
joshimoo, line 1882: This is good enough for now, just leaving a note here for the future.

NIT: there is kubernetes.go that contains functions for working with the kubernetes resources. Which is a part of the implementation of the datastore. It would require some additional cleanup, to not rely on the s.namespace instead as you have done here to pass the namespace explicitly.

datastore/datastore.go
```diff
@@ -31,185 +26,190 @@ var (
...
-	csiDriverInformer storageinformers_v1.CSIDriverInformer,
-	storageclassInformer storageinformers_v1.StorageClassInformer,
-	pdbInformer policyinformers.PodDisruptionBudgetInformer,
-	serviceInformer coreinformers.ServiceInformer,
-
+	kubeInformerFactory informers.SharedInformerFactory,
 	kubeClient clientset.Interface,
 	namespace string) *DataStore {
 
+	replicaInformer := lhInformerFactory.Longhorn().V1beta1().Replicas()
```
joshimoo, line 100: NIT: let's do the same as in the controllers, (informer, cache) pair.
Makes it easy to see that a cachesync add to the array is necessary when adding a new informer :)

joshimoo (commented):
LGTM good job the refactoring :) 

I see why you went with the cachesync array for the unit tests.
This can be merged after double checking the review comments.

datastore/longhorn.go
shuo-wu: NIT: I think it's fine to add one more commit in which this function can be moved to `datastore/kubernetes.go`. And it's unnecessary to open a separate PR for this movement.

controller/kubernetes_pv_controller.go
```diff
@@ -84,39 +69,33 @@ func NewKubernetesPVController(
...
+	ds.PodInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
 		AddFunc:    kc.enqueuePodChange,
 		UpdateFunc: func(old, cur interface{}) { kc.enqueuePodChange(cur) },
 		DeleteFunc: kc.enqueuePodChange,
 	})
+	kc.cacheSyncs = append(kc.cacheSyncs, ds.PodInformer.HasSynced)
 
 	// after volume becomes detached, try to delete the VA of lost node
-	volumeInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+	ds.VolumeInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
```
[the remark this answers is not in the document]
jenting: I did check the code but I can't 100% sure if there are other features that rely on the `volumeInformer`.
The code was introduced from #253.

@shuo-wu do you have any idea/feedback on this one?

shuo-wu (dismissed):

datastore/datastore.go
joshimoo: Sure you add `cacheSyncs []cache.InformerSynced` to the datastore struct.
Then after each informer, you append to the cacheSyncs array, just like in the controllers.
Then in the DS.sync function you wait for all hasSynced functions, just like in the controllers.

```go
func (s *DataStore) Sync(stopCh <-chan struct{}) bool {
	return cache.WaitForNamedCacheSync("longhorn datastore", stopCh, cacheSyncs...)
}
```

joshimoo (dismissed):
LGTM, I would like the cachesync change to the DS to make it consistent with the controller behavior and make it less likely that we miss a cachesync in the ds when a new informer is added.

PR #1084, opened 2021-10-08
manager: Add space trimming for the backing image creation input

shuo-wu:


shuo-wu:
Not sure if this backend PR is still required once we have a UI fix. If NO, feel free to close this.

PR #1085, opened 2021-10-09
Validate backing image checksum during creation

derekbit:
Make sure the given backing image checksum is SHA512 value during creation.
If the checksum is not given, skip the validation.

### DIFF
--- a/manager/backingimage.go
+++ b/manager/backingimage.go
@@ -68,6 +68,12 @@ func (m *VolumeManager) CreateBackingImage(name, checksum, sourceType string, pa
 		return nil, fmt.Errorf("invalid name %v", name)
 	}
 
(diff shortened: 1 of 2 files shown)


derekbit:
> Can we add the space trimming as well? #1084

Hello Shuo,
Seems good. That is, we can tolerate the unexpected leading whitespaces, like "    a1dcde...1a2s".
Do you mind I integrate https://github.com/longhorn/longhorn-manager/pull/1084 into one single commit?

shuo-wu:
@derekbit Yes. But you can put the content in a separate commit of this PR. Then I can close my PR.

derekbit:
Open https://github.com/longhorn/longhorn-manager/pull/1086 and close this.

PR #1086, opened 2021-10-12
Fix backing image creation input

derekbit:
The PR is to fix  https://github.com/longhorn/longhorn/issues/3133

- Trim write-spaces for the backing image creation input
- Make sure the given backing image checksum is SHA512 value during creation. If the checksum is not given, skip the validation.


innobead (approved):
LGTM

innobead:
@derekbit Please backport to 1.2.x.

PR #1089, opened 2021-10-13
Create the BackupVolume CR if it's the 1st volume backup

jenting:
#### Proposal Changes
- Create BackupVolume CR for the volume 1st backup.
- Inside backup_controller:
  - Do not run pull remote backup _or_ create a backup if the backup target is unavailable.
  - Only inspect the remote backup if it's a New/InProgress/Completed Backup CR.
- Inside backup_volume_controller:
  - Do not run pull remote backup volume if the backup target URL is unavailable.
  - Skip removes the Backup CR which is still in backup creation state (New/InProgress/Completed) and hasn't be synced from the remote backup target yet.
- Inside backup_target_controller: Skip removes the BackupVolume CR which hasn't be synced from the remote backup target yet.

#### Issues
- https://github.com/longhorn/longhorn/issues/3082
- https://github.com/longhorn/longhorn/issues/3141

### DIFF
--- a/controller/backup_volume_controller.go
+++ b/controller/backup_volume_controller.go
@@ -218,7 +218,7 @@ func (bvc *BackupVolumeController) reconcile(backupVolumeName string) (err error
 		}
 
 		// Delete the backup volume from the remote backup target
-		if backupTarget.Spec.BackupTargetURL != "" {
+		if backupTarget.Status.Available {
 			// Initialize a backup target client
 			backupTargetClient, err := getBackupTargetClient(bvc.ds, backupTarget)
 			if err != nil {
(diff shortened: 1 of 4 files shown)


controller/backup_volume_controller.go
```diff
@@ -283,8 +283,10 @@ func (bvc *BackupVolumeController) reconcile(backupVolumeName string) (err error
 	clustersSet := sets.NewString()
 	for _, b := range clusterBackups {
 		// Skip the Backup CR which is created from the local cluster and
-		// the snapshot backup hasn't be completed or pulled from the remote backup target yet
-		if b.Spec.SnapshotName != "" && b.Status.State != types.BackupStateCompleted {
+		// the snapshot backup hasn't be completed and sync with the remote backup target yet
+		// note that the error/unknown Backup CRs would be included in the clustersSet
+		if b.Status.LastSyncedAt.IsZero() && b.Spec.SnapshotName != "" &&
+			(b.Status.State == types.BackupStateNew || b.Status.State == types.BackupStateInProgress || b.Status.State == types.BackupStateCompleted) {
```
shuo-wu, line 290: Can the state be completed while `LastSyncedAt` is empty?
And it's better to use a separate function like `IsBackupInProgress` to abstract this part.

jenting:
We should consider removing the BackupVolume CRD instead of this PR's approach.
Close this PR now.

PR #1094, opened 2021-10-14
Remove deadcode

100mik:
Removing deadcode detected by `golangci-lint` on [Sonatype Lift](https://lift.sonatype.com/result/longhorn/longhorn-manager/01FHW2JKTWJ3Z4RA5MMT2A53WS?t=CustomTool%20%22golangci-lint%22%7Cdeadcode).


shuo-wu (approved):

app/driver.go
```diff
@@ -319,16 +319,6 @@ type DaemonSetOps struct {
 	kubeClient *clientset.Clientset
 }
 
-func newDaemonSetOps(kubeClient *clientset.Clientset) (*DaemonSetOps, error) {
```
shuo-wu, line 322: After rechecking the code, it seems that struct `DaemonSetOps` is not used as well. I think you can remove the struct as well as all methods/functions for it.

client/generated_node_input.go
```diff
@@ -28,12 +28,6 @@ type NodeInputOperations interface {
 	Delete(container *NodeInput) error
 }
 
-func newNodeInputClient(rancherClient *RancherClient) *NodeInputClient {
```
shuo-wu, line 31: Similar here. It seems that this whole go file is useless.

types/types.go
```diff
@@ -170,7 +170,6 @@ const (
 
 	BackingImageDataSourcePodNamePrefix = "backing-image-ds-"
 
-	shareManagerPrefix    = "share-manager-"
```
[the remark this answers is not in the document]
100mik: made this change!

app/driver.go
```diff
@@ -313,6 +315,7 @@ func deployCSIDriver(kubeClient *clientset.Clientset, lhClient *lhclientset.Clie
 
 	return nil
 }
+<<<<<<< HEAD
```
shuo-wu, line 318: Need to remove these lines.

100mik: My bad missed this. Done!

types/types.go
```diff
@@ -414,11 +415,11 @@ func GetShareManagerImageChecksumName(image string) string {
 }
 
 func GetShareManagerPodNameFromShareManagerName(smName string) string {
-	return LonghornLabelShareManager + "-" + smName
+	return shareManagerPrefix + "-" + smName
```
joshimoo, line 418: Since the prefix already includes, the `-` sign, the appending of the `-` is not longer needed.

types/types.go
```diff
@@ -414,11 +415,11 @@ func GetShareManagerImageChecksumName(image string) string {
...
 }
 
 func GetShareManagerNameFromShareManagerPodName(podName string) string {
-	return strings.TrimPrefix(podName, LonghornLabelShareManager+"-")
+	return strings.TrimPrefix(podName, shareManagerPrefix+"-")
```
joshimoo, line 422: Since the prefix already includes, the - sign, the appending of the - is not longer needed.

types/types.go
```diff
@@ -170,7 +170,8 @@ const (
 
 	BackingImageDataSourcePodNamePrefix = "backing-image-ds-"
 
-	shareManagerPrefix    = "share-manager-"
+	shareManagerPrefix = "share-manager-"
+
```
joshimoo, line 174: NIT: no need for a separate section for this variable.

joshimoo (changes requested):
LGTM, when replacing the usage with the const var, the extra `-` was overlooked. After that is fixed this can be merged.

types/types.go
100mik: Incorporated suggestions!

joshimoo (approved):
LGTM

PR #1104, opened 2021-10-14
handle the  unhandled errors.

slayer321:
no issue related to bugbash

https://lift.sonatype.com/result/longhorn/longhorn-manager/01FHW2JKTWJ3Z4RA5MMT2A53WS?t=CustomTool%20%22GoSec%22%7CG104&t=CustomTool%20%22golangci-lint%22%7Cerrcheck

### DIFF
--- a/controller/volume_controller_test.go
+++ b/controller/volume_controller_test.go
@@ -3,7 +3,6 @@ package controller
 import (
 	"context"
 	"fmt"
-	"log"
 	"math"
 	"strconv"
 	"time"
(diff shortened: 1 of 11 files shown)


controller/volume_controller_test.go
```diff
@@ -1460,7 +1461,10 @@ func (s *TestSuite) runTestCases(c *C, testCases map[string]*VolumeTestCase) {
 			n, err := lhClient.LonghornV1beta1().Nodes(TestNamespace).Create(context.TODO(), node, metav1.CreateOptions{})
 			c.Assert(err, IsNil)
 			c.Assert(n, NotNil)
-			nIndexer.Add(n)
+			err = nIndexer.Add(n)
+			if err != nil {
+				log.Println(err)
+			}
```
shuo-wu, line 1467: It's better to use `c.Assert(err, IsNil)` instead so that the unit test won't continue. Similar below

shuo-wu: I mean all errors in the unit test (`controller/xxxxx_controller.test.go`). In this PR, they are all like:
```
err = nIndexer.Add(n)
if err != nil {
	log.Println(err)
}
```

Instead, it's better to modify it to this:
```
err = nIndexer.Add(n)
c.Assert(err, IsNil)
```

slayer321: ok got it

PR #1107, opened 2021-10-20
Move backupStatus from engine controller/CR to backup controller/CR

jenting:
#### Proposal Changes
- Deprecated backupStatus in Engine CR. (We can't delete it directly since we need migration to Backup CR).
- Add resource monitor logic to backup_controller to monitor backup creation progress.
- Pass `backupName` + `replicaAddress` to fetch current backup status.
- In the upgrade, move the backupStatus from Engine CR to Backup CR.
- The HTTP `/v1/volumes` and WebSocket `/v1/ws/volumes` endpoints returns backupStatus from Backup CR instead of Engine CR.

#### Issues
- https://github.com/longhorn/longhorn/issues/3035
- https://github.com/longhorn/longhorn/issues/2441
- https://github.com/longhorn/longhorn/issues/1311
- https://github.com/longhorn/longhorn/issues/2997
- https://github.com/longhorn/longhorn/issues/3033


controller/backup_controller.go
```diff
@@ -434,99 +510,139 @@ func (bc *BackupController) backupCreation(log logrus.FieldLogger, engineClient
...
-				state = types.BackupStateUnknown
-				event(err, state, backup, volume)
-				return
-			}
-			if bks.Error != "" {
-				state = types.BackupStateError
-				event(errors.New(bks.Error), state, backup, volume)
-				return
-			}
+type BackupMonitor struct {
```
joshimoo, line 557: NIT: Consider putting the backup monitor related code into it's own file / package.

In general the backup monitor should be able to be tested independently from the backup controller.
For example via an engine sim client, we currently don't have many unit tests in this regard yet.

controller/backup_controller.go
```diff
@@ -434,99 +510,139 @@ func (bc *BackupController) backupCreation(log logrus.FieldLogger, engineClient
...
+	ticker := time.NewTicker(BackupMonitorTickPeriod)
+	defer ticker.Stop()
+	timeSinceLastSync := time.Duration(0)
+	for {
+		select {
+		case <-m.ctx.Done():
+			return
+		case <-ticker.C:
+			timeSinceLastSync += BackupMonitorTickPeriod
+			if timeSinceLastSync > BackupMonitorSyncPeriod || m.stale {
```
joshimoo, line 638: NIT: this should be >= syncPeriod otherwise you will only sync on the 3rd tick instead of the 2nd tick.

controller/backup_controller.go
```diff
@@ -434,99 +510,139 @@ func (bc *BackupController) backupCreation(log logrus.FieldLogger, engineClient
...
-		for {
-			engines, err := bc.ds.ListVolumeEngines(volume.Name)
-			if err != nil {
-				state = types.BackupStateUnknown
-				event(err, state, backup, volume)
-				return
-			}
+const (
+	BackupMonitorTickPeriod = time.Second
+	BackupMonitorSyncPeriod = 2 * time.Second
```
joshimoo, line 554: NIT: do we need such a short period? currently the engine call via binary can be still expensive.
This is less of a problem once we switch to a persistent grpc connection.

controller/backup_controller.go
```diff
@@ -279,29 +287,97 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
...
+			// https://github.com/longhorn/longhorn-engine/blob/9da3616/pkg/replica/backup.go#L20-L22
+			switch bm.engineBackupStatus.State {
+			case "in_progress":
+				backup.Status.State = types.BackupStateInProgress
+			case "complete":
+				backup.Status.State = types.BackupStateCompleted
+			case "error":
+				backup.Status.State = types.BackupStateError
+			}
+
```
joshimoo, line 338: NIT: consider moving this mapping/conversion function (engineBackupStatus -> backup.Status) into a util function.

joshimoo (commented):
In general the code looks okay. 

I am thinking about whether we actually care about the transient engine.BackupStatus information from the past. Since the source of truth is the backupStore, so for completed backups we would have them anyway.

For future backups we would have them going forward, due to the monitoring.
I am guessing doing the one time migration as part of the upgrade process might be okay, but I am not sure it's really needed.

controller/backup_controller.go
```diff
@@ -434,99 +510,139 @@ func (bc *BackupController) backupCreation(log logrus.FieldLogger, engineClient
...
+func (m *BackupMonitor) Close() {
+	m.quit()
+}
+
+func (m *BackupMonitor) syncBackups() error {
+	m.lock.Lock()
+	defer m.lock.Unlock()
+
+	engineBackupStatusList, err := m.engineClient.SnapshotBackupStatus()
```
shuo-wu, line 608: Just notice that here the output of this function is a map. But the result from the cmd is a struct. Can you double-check this part?
https://github.com/longhorn/longhorn-engine/blob/31ac10b371897346f9f3b08a731af48f5f132d70/app/cmd/backup.go#L229

controller/backup_controller.go
```diff
@@ -279,29 +287,97 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
...
+			bm := bc.hasMonitor(backup.Name)
+			if bm == nil || bm.engineBackupStatus == nil {
+				return nil
+			}
+
+			backup.Status.Progress = bm.engineBackupStatus.Progress
+			backup.Status.URL = bm.engineBackupStatus.BackupURL
+			backup.Status.Error = bm.engineBackupStatus.Error
+			backup.Status.SnapshotName = bm.engineBackupStatus.SnapshotName
+			backup.Status.ReplicaAddress = bm.engineBackupStatus.ReplicaAddress
```
shuo-wu, line 328: It's better to add a RWLock for this field.

controller/backup_controller.go
```diff
@@ -279,29 +287,97 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
...
+				if err != nil {
+					return err
+				}
+				bc.eventLog(nil, backup.Status.State, backup, volume)
+			}
+
+			// state changes from in progress to complete
+			if backup.Status.State == types.BackupStateCompleted {
+				// disable backup monitor
+				bc.disableBackupMonitor(backup)
```
shuo-wu, line 352: Putting this line here implicitly means the monitor stop is an edge-triggered operation and it happens only when the current backup state is updated from `in_progress` to `completed`. 
IMO, this should be a level-triggered operation instead: As long as the current backup reaches an end state (completed, error, not sure if unknown should be included), the monitor can be stopped.

controller/backup_controller.go
jenting: The result of the cmd is a raw string, for example:
```shell
{
	"backup-02b651e23f314cf3": {
		"progress": 100,
		"backupURL": "s3://backupbucket@minio/?backup=backup-02b651e23f314cf3\u0026volume=pvc-75631cf4-cd4e-40e7-b05f-acd87efdda0d",
		"snapshotName": "78684fb5-696f-4071-a1f6-4a5d44444d47",
		"state": "complete",
		"replicaAddress": "tcp://10.42.2.115:10000"
	},
	"backup-a2471feb35b04a1d": {
		"progress": 100,
		"backupURL": "s3://backupbucket@minio/?backup=backup-a2471feb35b04a1d\u0026volume=pvc-75631cf4-cd4e-40e7-b05f-acd87efdda0d",
		"snapshotName": "ebc45a24-c226-4c2f-b55e-732f8e004b7b",
		"state": "complete",
		"replicaAddress": "tcp://10.42.2.115:10000"
	}
}
```

Longhorn manager [unmarshal](https://github.com/longhorn/longhorn-manager/blob/f83a22c/engineapi/backups.go#L300-L301) the console output to the map.

controller/backup_controller.go
```diff
@@ -434,99 +510,139 @@ func (bc *BackupController) backupCreation(log logrus.FieldLogger, engineClient
 		biChecksum = bi.Status.Checksum
 	}
 
-	backup.Status.State = types.BackupStateInProgress
-	event(nil, backup.Status.State, backup, volume)
+	var state types.BackupState
+	_, err = engineClient.SnapshotBackup(backup.Name, backup.Spec.SnapshotName, url, biName, biChecksum, backup.Spec.Labels, credential)
```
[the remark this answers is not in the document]
jenting: No. It'll do the same operation again.

If this operation runs twice and since the input parameters are the same, so in the end, it won't create two backups.

controller/backup_controller.go
jenting: It follows the original behavior.

I agree that the gRPC call is expensive since we have to init the gRPC connection every 2 seconds.

shuo-wu:
@jenting Correctly, the backup creation call should inform the replica address as well. Besides, in the worst case, I think the cmd `BackupStatus` should be able to detect the replica that is processing the backup creation even if the caller does not mention/doesn't know it.

shuo-wu (changes requested):

controller/backup_controller.go
```diff
@@ -405,104 +407,96 @@ func (bc *BackupController) backupCreation(log logrus.FieldLogger, engineClient
...
+		backup.Status.State = backupStatus.State
 
+		// log+event when backup state changes
+		if existingBackup.Status.State != backupStatus.State {
+			event(nil, backup.Status.State, backup, volume)
+		}
+		return false, nil
+	case types.BackupStateCompleted:
+		if bm := bc.hasMonitor(backup.Name); bm != nil {
+			backupStatus := bm.GetBackupStatus()
```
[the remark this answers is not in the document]
jenting: I've updated the PR to blindly sync the CR status from the longhorn engine if the Backup CR that needs to perform snapshot backup to the remote backup target.

controller/backup_controller.go
```diff
@@ -385,146 +381,185 @@ func (bc *BackupController) getBackupVolumeName(backup *longhorn.Backup) (string
...
-			if err != nil {
-				log.WithError(err).Errorf("Error get backup")
-				return
-			}
-			existingBackup := backup.DeepCopy()
+		// Initialize a backup target client
+		backupTargetClient, err := getBackupTargetClient(bc.ds, backupTarget)
+		if err != nil {
+			log.WithError(err).Error("Error init backup target client")
+			return true, nil // Ignore error to prevent enqueue
```
innobead: Why the exact reason is for ignoring the error from `getBackupTargetClient`?

api/router.go
```diff
@@ -165,7 +165,7 @@ func NewRouter(s *Server) *mux.Router {
 	r.Path("/v1/ws/settings").Handler(f(schemas, settingListStream))
 	r.Path("/v1/ws/{period}/settings").Handler(f(schemas, settingListStream))
 
-	volumeListStream := NewStreamHandlerFunc("volumes", s.wsc.NewWatcher("volume", "engine", "replica"), s.volumeList)
+	volumeListStream := NewStreamHandlerFunc("volumes", s.wsc.NewWatcher("volume", "engine", "replica", "backup"), s.volumeList)
```
innobead, line 168: Do we need anything from UI side to adapt to this change?

engineapi/backups.go
```diff
@@ -277,32 +277,52 @@ func (e *Engine) SnapshotBackup(backupName, snapName, backupTarget, backingImage
...
+	}
+
 	output, err := e.ExecuteEngineBinary(args...)
 	if err != nil {
 		return nil, err
 	}
-	backups := make(map[string]*longhorn.BackupStatus)
-	if err := json.Unmarshal([]byte(output), &backups); err != nil {
+
+	engineBackupStatus := &longhorn.EngineBackupStatus{}
```
innobead, line 306: Because `replicaAddress` is possible to be empty string, shouldn't we still have getting all backup status in this existing case?

upgrade/v121to122/upgrade.go
```diff
@@ -0,0 +1,140 @@
+package v121to122
```
innobead, line 1: We should add this to v122to123 instead, because 1.2.2 is already released.

upgrade/v121to122/upgrade.go
```diff
@@ -0,0 +1,140 @@
...
+		if !exist {
+			continue
+		}
+
+		// Get the corresponding volume's engine
+		engines, err := lhClient.LonghornV1beta1().Engines(namespace).List(context.TODO(), metav1.ListOptions{
+			LabelSelector: types.LonghornLabelVolume + "=" + volumeName,
+		})
+		if err != nil {
+			continue
```
innobead, line 59: Suggest don't silently swallow any error message even you don't want it block the upgrade path. It's better to have a log for it.

upgrade/v121to122/upgrade.go
```diff
@@ -0,0 +1,140 @@
...
+			engine = engines.Items[0]
+		default:
+			// If more than 2 engines found, use the current volume's engine
+			v, err := lhClient.LonghornV1beta1().Volumes(namespace).Get(context.TODO(), volumeName, metav1.GetOptions{})
+			if err != nil {
+				if apierrors.IsNotFound(err) {
+					// Cannot found the correspoding volume
+					continue
+				}
+				return err
```
innobead, line 77: This will cause upgrade failed in the middle? If yes, I am thinking if we can make some pre-upgrade validation in the future in general to make sure at least the upgrade condition is OK before the upgrade (not really related to this PR)

upgrade/v121to122/upgrade.go
```diff
@@ -0,0 +1,140 @@
...
+			}
+
+			for _, e := range engines.Items {
```
innobead, line 80: It seems possible to have `engine` is still nil if no engines match the below condition?

upgrade/v121to122/upgrade.go
```diff
@@ -0,0 +1,140 @@
...
+					e.Spec.DesireState == longhorn.InstanceStateRunning &&
+					e.Status.CurrentState == longhorn.InstanceStateRunning {
+					engine = e
+					break
+				}
+			}
+		}
+
+		// No corresponding backupStatus inside engine CR
+		backupStatus, exist := engine.Status.BackupStatus[backup.Name]
```
innobead, line 91: If `engine` is nil, then null exception happens

innobead (changes requested):
In general, LGTM. But some feedbacks needed to resolve.

controller/backup_controller.go
jenting: Because the function `getBackupTargetClient` returns error when the URL+Credential pair in invalid.
Skip return error here to prevent the controller from re-enqueue it again because even after immediately reconciling again, the backup target client is still invalid.

api/router.go
jenting: No, UI focuses on the RESTful/WebSocket endpoint response.
The longhorn manager backend handles to get the `backupstatus` from Backup CR now (originally is from Engine CR).

engineapi/backups.go
jenting: If `replicaAddress `is empty, the longhorn engine will find the corresponding replica by looping all the replicas. You could refer to the longhorn engine [PR](https://github.com/longhorn/longhorn-engine/blob/77d1117f3c917be41951a1c6d4b6a51d1a2ff550/app/cmd/backup.go#L117-LL138).

upgrade/v121to122/upgrade.go
jenting: Okay!

upgrade/v121to122/upgrade.go
jenting: Okay, I'll add check if the `err` is `apierrors.IsNotFound`.
If yes, it means the Backup CR doesn't have corresponding Volume/Engine within the cluster, we should continue the next Backup CR.
If no, it means something error when List the engines, will return an error.

upgrade/v121to122/upgrade.go
jenting: I think no. It's because the Backup CR doesn't have a corresponding Volume within the cluster (Backup pulled from the remote backup target)

upgrade/v121to122/upgrade.go
jenting: I think the `engine` is not a pointer [code](https://github.com/longhorn/longhorn-manager/blob/66ea070/k8s/pkg/apis/longhorn/v1beta1/engine.go#L114), so it shouldn't be a `nil`.

upgrade/v121to122/upgrade.go
jenting: [1]

upgrade/v121to122/upgrade.go
jenting: Change to directly error out.

controller/backup_controller.go
innobead: Got it. Maybe it will be good to check the exact error type which you expect to be able to ignore?

upgrade/v121to122/upgrade.go
innobead: ;)

controller/backup_controller.go
jenting: Fair.

I'd prefer to create another PR for this one because the backup_target_controller/backup_volume_controller/backup_controller all use the same function `getBackupTargetClient `.

innobead: Sure, please do it.

innobead (approved):

innobead:
ping @shuo-wu / @joshimoo

controller/backup_controller.go
```diff
@@ -413,153 +415,122 @@ func (bc *BackupController) validateBackingImageChecksum(volName, biName string)
...
-	)
-
-	currentState := backup.Status.State
+// backupCreation calls engine's snapshot backup API or sync current engine's snapshot backup progress/state
+// if the `spec.SnapshotName` is set and `status.lastSyncedAt` is unset.
+// note that:
+// - if the CR without the spec.SnapshotName, the sourcce of truth is the remote backup target
+// - if the CR with the spec.SnapshotName, there are two source of truth:
+//   - the 1st is the engine, longhorn manager calls engine's snapshot backup API and monitoring the longhorn engine reports progress/state
+//   - the 2nd is once the engine reports snapshot backup state Completed, the source of truth becomes the remote backup target
```
shuo-wu, line 424: 1. Move these comments above `if backup.Status.LastSyncedAt.IsZero() && backup.Spec.SnapshotName != "" {`.
2. Field `spec.SnapshotName` should be set once the CR is created. How can it become empty? Do you mean the source of truth is the engine/replica if `status.lastSyncedAt` is empty?
3. If `status.lastSyncedAt` is set, it means the controller won't sync with the monitor hence the source of truth should be the remote backup target only rather than both.

controller/backup_controller.go
```diff
@@ -284,15 +283,19 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
...
+		case longhorn.BackupStateNew, longhorn.BackupStateInProgress:
+			return nil
+		case longhorn.BackupStateCompleted:
+			bc.disableBackupMonitor(backup.Name)
+		case longhorn.BackupStateError, longhorn.BackupStateUnknown:
+			backup.Status.LastSyncedAt = syncTime
+			bc.disableBackupMonitor(backup.Name)
 			return nil
 		}
 	}
```
shuo-wu: Can we refactor it like the following? The method namings here are not important...
```
syncWithReplica := backup.Status.LastSyncedAt.IsZero() && backup.Spec.SnapshotName != ""
if syncWithReplica {
    checkMonitor(backup)  // enable or disable monitor based on backup current state
    syncWIthMonitor()
    WrapUp()  // Disable monitor and update backup volume
    return
}

// The backup creation is complete, then the source of truth becomes the remote backup target
```

controller/backup_controller.go
```diff
@@ -313,17 +316,17 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
...
 	if err != nil {
-		if !strings.Contains(err.Error(), "in progress") {
-			log.WithError(err).Error("Error inspecting backup config")
-		}
+		log.WithError(err).Error("Error inspecting backup config")
 		return nil // Ignore error to prevent enqueue
 	}
+
+	backup.Status.State = longhorn.BackupStateCompleted
+	backup.Status.LastSyncedAt = syncTime
```
shuo-wu, line 324: Any explanation for this modification?

controller/backup_controller.go
```diff
@@ -413,153 +415,122 @@ func (bc *BackupController) validateBackingImageChecksum(volName, biName string)
...
+			backup.Status.Error = backupStatus.Error
+			backup.Status.SnapshotName = backupStatus.SnapshotName
+			backup.Status.ReplicaAddress = backupStatus.ReplicaAddress
+			backup.Status.State = backupStatus.State
+		}
+		if existingBackupState == backup.Status.State {
+			return
+		}
+
+		if backup.Status.Error != "" {
```
shuo-wu, line 448:
```
if existingBackup.Status.Error != backup.Status.Error {
```

controller/backup_controller.go
```diff
@@ -413,153 +415,122 @@ func (bc *BackupController) validateBackingImageChecksum(volName, biName string)
...
+			backup.Status.Progress = backupStatus.Progress
+			backup.Status.URL = backupStatus.URL
+			backup.Status.Error = backupStatus.Error
+			backup.Status.SnapshotName = backupStatus.SnapshotName
+			backup.Status.ReplicaAddress = backupStatus.ReplicaAddress
+			backup.Status.State = backupStatus.State
+		}
+		if existingBackupState == backup.Status.State {
+			return
+		}
```
shuo-wu, line 446: Do not return before printing out the error.

controller/backup_controller.go
```diff
@@ -413,153 +415,122 @@ func (bc *BackupController) validateBackingImageChecksum(volName, biName string)
...
+		backup.Status.Error = err.Error()
+		backup.Status.State = longhorn.BackupStateError
+		return err
 	}
+	backup.Status.ReplicaAddress = replicaAddress
+	backup.Status.State = longhorn.BackupStateInProgress
 
-	log.Warnf("Unknown backup state %s", backup.Status.State)
-	return true, nil
+	monitor = bc.enableBackupMonitor(backup, volumeName, engineClient, replicaAddress)
```
shuo-wu: I am thinking if we can move this API call into the monitor: When the monitor is launched and the backup current state is New, the API will be invoked. And in the next loop monitor could get the state from the engine/replica so that the controller does not need to update the state manually here.

shuo-wu (changes requested):
Please squash the last commit into the 2nd one.

upgrade/v122to123/upgrade.go
```diff
@@ -0,0 +1,140 @@
+package v122to123
```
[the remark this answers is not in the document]
jenting: I'm curious if we'll backport this PR to v1.2.x, what's the best package name it should be?
Or, for the master branch, the package name is v122tov123; for the v1.2.x branch, the package name is v122tov123?

controller/backup_controller.go
jenting: I thought that if the backup inspection succeeded (without error) but the backup info in nil, we should consider the backup synced with the remote backup target already. Therefore, updates the backup state to `Completed` and `lastSyncedAt`.

controller/backup_controller.go
jenting: We only want to emit an event and log when the backup status state change.
For example:
- New -> InProgress
- InProgress -> Completed
- New -> Error
- InProgress -> Error

controller/backup_controller.go
jenting: [1]

controller/backup_controller.go
jenting: If the Backup CR is pulled from the remote backup target, the `spec.SnapshotName` would be empty.

I think the source of truth is:
- engine/replica is `status.lastSyncedAt == ""` && `spec.SnapshotName != ""` && `status.State != Complete`.
- remote backup target is `status.lastSyncedAt == ""` && `spec.SnapshotName != ""` && `status.State == Complete` _or_ `status.lastSyncedAt == ""` && `spec.SnapshotName == ""`.

upgrade/v122to123/upgrade.go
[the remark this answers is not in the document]
jenting: Besides that, we're going to release the v123 release, the package name should also change to v123tov130?
How to tackle the release/master branch upgrade package naming?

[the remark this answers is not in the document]
shuo-wu: OK. Using different packages for v1.2.3 and v1.3.0 but with the same code is redundant. Let's retain this.

controller/backup_controller.go
shuo-wu: OK. Got it. If the backup CR is created by the backup volume controller and needs to be synced with the remote backup target immediately after creation, the snapshot name is empty.

The condition for the remote backup target is too complicated. As long as the source of truth is not the engine/replica, then it must be the remote backupstore.

jenting: Okay, I'll make the wording/condition be simplified.

controller/backup_controller.go
jenting: I had thought this before. The result would be that if the backup size is small and the backup process is finished within the poll interval (2 seconds here), then the event log will only record the backup transition state `Completed` since the state changed from New -> Completed.
If the backup process time is over than the poll interval (2 seconds), then the event log will have transition state `InProgress` and `Completed`.

If we're don't care the event log must have InProgress -> Completed state, then yes we could move the `engineClient.SnapshotBackup` API call into the `NewBackupMonitor`.

controller/backup_controller.go
jenting: Yes, we could consider it once we have a consensus on [2].

controller/backup_controller.go
shuo-wu: I don't think a backup should restrictedly follow New -> InProgress -> Completed flow. What the controller does here is **blindly** syncing the state from the engine/replica. This `blind` means we cannot control the flow.

controller/backup_controller.go
shuo-wu: How can the backup info be nil? Does this mean something is wrong with the backup config file?

jenting: In general, it's impossible that the err = nil and backupInfo = nil.

Never mind, I will revert this change.

shuo-wu: OK. You can print a warning log if `backupInfo == nil` for now.

controller/backup_controller.go
jenting: Updated!

controller/backup_controller.go
jenting: Updated.

controller/backup_controller.go
jenting: Updated.

controller/backup_controller.go
```diff
@@ -385,146 +400,164 @@ func (bc *BackupController) getBackupVolumeName(backup *longhorn.Backup) (string
...
+	bc.eventRecorder.Eventf(volume, corev1.EventTypeNormal, string(backup.Status.State),
+		"Snapshot %s backup %s label %v", backup.Spec.SnapshotName, backup.Name, backup.Spec.Labels)
 
-			if bks.Progress != 100 {
-				time.Sleep(BackupStatusQueryInterval)
-				continue
-			}
+	if backup.Status.State == longhorn.BackupStateCompleted {
+		return bc.syncBackupVolume(volume.Name)
+	}
```
shuo-wu, line 498: Moving this and the following monitor disabling into a separate function would make the flow more easy-to-understand. You can name the function something like `finishMonitoring` or `wrapUpMonitoring`.

shuo-wu (changes requested):

controller/backup_controller.go
jenting: Put here is intended to only when the state changes to Completed, call `syncBackupVolume`.

jenting: Change the below to a new function
```go
		switch backup.Status.State {
		case longhorn.BackupStateNew, longhorn.BackupStateInProgress:
			return nil
		case longhorn.BackupStateCompleted:
			bc.disableBackupMonitor(backup.Name)
		case longhorn.BackupStateError, longhorn.BackupStateUnknown:
			backup.Status.LastSyncedAt = syncTime
			bc.disableBackupMonitor(backup.Name)
			return nil
		}
```

would require an addition returned parameter to indicate whether to continue the reconcile loop _or_ immediately return it.

controller/backup_controller.go
```diff
@@ -385,146 +400,164 @@ func (bc *BackupController) getBackupVolumeName(backup *longhorn.Backup) (string
...
+	if monitor != nil {
+		return monitor, nil
+	}
+
+	bc.monitorLock.Lock()
+	defer bc.monitorLock.Unlock()
+	monitor, err := engineapi.NewBackupMonitor(bc.logger, backup, volume, backupTargetClient, biChecksum, engineClient, bc.enqueueBackupForMonitor)
+	if err != nil {
+		backup.Status.Error = err.Error()
+		backup.Status.State = longhorn.BackupStateError
```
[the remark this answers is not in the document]
jenting: I moved to the function `checkMonitor`.

controller/backup_controller.go
shuo-wu: OK. If it is not as simple as I expected, we can retain the current implementation. It's already much better than the very 1st version.

shuo-wu (approved):
LGTM. Let @joshimoo take a final look and see if the monitor refactoring is what he expects.

joshimoo (approved):
LGTM, good job on this changeset. 
Any additional changes/improvements we can make when we implement the persistent grpc connections, since we will be touching the monitoring routines in that case any way :)

shuo-wu (approved):

PR #1117, opened 2021-11-02
Apply podAntiAffinity rules to csi sidecar containers

derekbit:
To prevent the csi sidecar containers from running on one host,
apply podAntiAffinity rules to spread the workloads to different
workers.

Longhorn#2894

### DIFF
--- a/csi/deployment_util.go
+++ b/csi/deployment_util.go
@@ -82,6 +82,30 @@ func getCommonDeployment(commonName, namespace, serviceAccount, image, rootDir s
 					Tolerations:        tolerations,
 					NodeSelector:       nodeSelector,
 					PriorityClassName:  priorityClass,
+					Affinity: &v1.Affinity{
+						PodAntiAffinity: &v1.PodAntiAffinity{
+							PreferredDuringSchedulingIgnoredDuringExecution: []v1.WeightedPodAffinityTerm{
+								{
+									Weight: 1,
+									PodAffinityTerm: v1.PodAffinityTerm{
+										LabelSelector: &metav1.LabelSelector{
+											MatchExpressions: []metav1.LabelSelectorRequirement{
+												{
+													Key:      "app",
+													Operator: metav1.LabelSelectorOpIn,
+													Values: []string{
+														commonName,
+													},
+												},
+											},
+										},
+
+										TopologyKey: v1.LabelHostname,
+									},
+								},
+							},
+						},
+					},
 					Containers: []v1.Container{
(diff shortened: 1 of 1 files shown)


csi/deployment_util.go
```diff
@@ -137,6 +138,34 @@ func getCommonDeployment(commonName, namespace, serviceAccount, image, rootDir s
...
+				},
+			},
+		},
+	}
+
+	return &v1.Affinity{
+		PodAntiAffinity: &v1.PodAntiAffinity{
+			PreferredDuringSchedulingIgnoredDuringExecution: []v1.WeightedPodAffinityTerm{
+				{
+					Weight: 1,
```
jenting, line 158: Not sure should we configure the weight or not.

jenting (commented):
In general LGTM.

nit: I probably will go without the function `getAffinity` which makes it easier to read the manifest.

csi/deployment_util.go
derekbit: > nit: I probably will go without the function getAffinity which makes it easier to read the manifest.

There are a lots of brackets after merging into the manifest.
If it is acceptable in our code convention, I can remove getAffinity().

> Not sure should we configure the weight or not.

There is only one affinity rule here, so removing it and using default value are maybe okay as well.

derekbit: Resend the patch followed @jenting's suggestions.

derekbit: >>Not sure should we configure the weight or not.

>There is only one affinity rule here, so removing it and using default value are maybe okay as well.

Although the k8s doc said weight is not required, but the longhon driver deploy crashes if the weight is not provided.

innobead: Any reason for the crash?

innobead:
@derekbit PR build failed, you can check the reason.

csi/deployment_util.go
derekbit: @innobead 
log error message: `spec.template.spec.affinity.podAntiAffinity.preferredDuringSchedulingIgnoredDuringExecution[0].weight: must be in the range 1-100`

Unfortunately, the default value set by kubernetes  is 0 which is not accepted by itself.

derekbit:
> @derekbit PR build failed, you can check the reason.

jenting (approved):

innobead (approved):
LGTM

innobead:
@derekbit Please back port to 1.2.x.

PR #1123, opened 2021-11-09
Remove volume attachment recovery policy from settings

jenting:
The volume attachment recovery policy is deprecated/removed from LH v1.1.2.
We'll release v1.3.0 in the next coming release, we could consider fully removing the code from the setting.


innobead (approved):
LGTM

PR #1125, opened 2021-11-10
Migrating volume auto recovery

shuo-wu:
longhorn/longhorn#3206


shuo-wu:
Regression for core tests: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/541/

upgrade/v122to130/upgrade.go
```diff
@@ -0,0 +1,93 @@
...
+		e := &engines.Items[i]
+		if e.Spec.VolumeName == "" {
+			// Cannot do anything in the upgrade path if this case happens.
+			continue
+		}
+		if _, exists := volumeEngineMap[e.Spec.VolumeName]; !exists {
+			volumeEngineMap[e.Spec.VolumeName] = []*v1beta1.Engine{e}
+		} else {
+			volumeEngineMap[e.Spec.VolumeName] = append(volumeEngineMap[e.Spec.VolumeName], e)
+		}
```
joshimoo, line 50: NIT: you can append to a nil slice, this will create and init a slice, so should be possible to avoid the extra comparison logic
```suggestion
		volumeEngineMap[e.Spec.VolumeName] = append(volumeEngineMap[e.Spec.VolumeName], e)
```

upgrade/v122to130/upgrade.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+
+	for volumeName, engineList := range volumeEngineMap {
+		skip := false
+		for _, e := range engineList {
+			if e.Spec.Active {
+				skip = true
+				break
+			}
```
joshimoo, line 59: NIT: I find this approach more readable for tiny loops, up to you :)
```suggestion
			skip = skip || e.Spec.Active
```

datastore/longhorn.go
```diff
@@ -585,29 +585,73 @@ func checkEngine(engine *longhorn.Engine) error {
...
-		return nil, nil
+// GetCurrentEngineAndExtras pick the current Engine and extra Engines from the Engine list of a volume with the given namespace
+func GetCurrentEngineAndExtras(v *longhorn.Volume, es map[string]*longhorn.Engine) (currentEngine *longhorn.Engine, extras []*longhorn.Engine, err error) {
+	if len(es) == 1 {
+		for _, e := range es {
+			return e, nil, nil
+		}
 	}
+	for _, e := range es {
+		if e.Spec.Active {
```
joshimoo, line 596: What happens if there are multiple active engines should we overwrite? Or is there another condition we can use in that case? i.e first come first server?

datastore/longhorn.go
```diff
@@ -585,29 +585,73 @@ func checkEngine(engine *longhorn.Engine) error {
...
-	node := v.Status.CurrentNodeID
-	if node != "" {
-		for _, e := range es {
-			if e.Spec.NodeID == node {
-				return e, nil
-			}
+	oldEngineName := ""
+	for name := range es {
+		e := es[name]
+		if e.Spec.Active {
```
joshimoo, line 621: What happens if there are multiple active engines should we overwrite? Or is there another condition we can use in that case? i.e first come first server?

manager/volume.go
```diff
@@ -524,9 +526,26 @@ func (m *VolumeManager) Detach(name, nodeID string) (v *longhorn.Volume, err err
...
+		}
+		if e.Spec.DesireState != longhorn.InstanceStateRunning || e.Status.CurrentState != longhorn.InstanceStateRunning {
+			continue
+		}
+		hasAvailableReplica := false
+		for _, mode := range e.Status.ReplicaModeMap {
+			if mode == longhorn.ReplicaModeRW {
+				hasAvailableReplica = true
+				break
+			}
```
joshimoo, line 543: NIT: for small loops I like this approach, I think it makes it more readable, you also get rid of the conditional code.
```suggestion
			hasAvailableReplica = hasAvailableReplica || mode == longhorn.ReplicaModeRW
```

joshimoo (approved):
LGTM, a couple of questions and some NIT's.
You can ignore the NIT's. After double checking the questions, feel free to merge.

controller/volume_controller.go
```diff
@@ -3049,7 +3195,7 @@ func (vc *VolumeController) IsReplicaUnavailable(r *longhorn.Replica) (bool, err
 		"replicaNodeID": r.Spec.NodeID,
 	})
 
-	if r.Spec.FailedAt != "" || r.Spec.NodeID == "" || r.Spec.DiskID == "" {
+	if r.Spec.NodeID == "" || r.Spec.DiskID == "" || r.Spec.DiskPath == "" || r.Spec.DataDirectoryName == "" {
```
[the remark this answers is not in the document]
shuo-wu: No, we don't need it. 
For failed migration replicas, the controller will add them into map `migrationReplicas` then do cleanup in function `deleteInvalidMigrationReplicas`. 
For failed old replicas, we can still clone migration replicas from them since this kind of duplicating is like replica reusage. The engine will determine if the cloned migration replicas are available based on the revision counter.

datastore/longhorn.go
shuo-wu: Right now it's first come first served. I am hesitating if we need to error out when there are multiple active engines. Same below.

innobead: If possible, in general we should have debug/warn log, because it's better for debugging.

datastore/longhorn.go
```diff
@@ -585,29 +585,73 @@ func checkEngine(engine *longhorn.Engine) error {
...
-func (s *DataStore) PickVolumeCurrentEngine(v *longhorn.Volume, es map[string]*longhorn.Engine) (*longhorn.Engine, error) {
-	if len(es) == 0 {
-		return nil, nil
+// GetCurrentEngineAndExtras pick the current Engine and extra Engines from the Engine list of a volume with the given namespace
+func GetCurrentEngineAndExtras(v *longhorn.Volume, es map[string]*longhorn.Engine) (currentEngine *longhorn.Engine, extras []*longhorn.Engine, err error) {
+	if len(es) == 1 {
+		for _, e := range es {
+			return e, nil, nil
+		}
 	}
```
innobead, line 594: Should we remove this part? but just rely on the below new loop?

datastore/longhorn.go
```diff
@@ -585,29 +585,73 @@ func checkEngine(engine *longhorn.Engine) error {
...
+}
 
+// GetNewCurrentEngineAndExtras detects the new current Engine and extra Engines from the Engine list of a volume with the given namespace during engine switching.
+func GetNewCurrentEngineAndExtras(v *longhorn.Volume, es map[string]*longhorn.Engine) (currentEngine *longhorn.Engine, extras []*longhorn.Engine, err error) {
 	if len(es) == 1 {
 		for _, e := range es {
-			return e, nil
+			return e, nil, nil
 		}
 	}
```
innobead, line 616: The same, should we remove this part?

upgrade/v122to130/upgrade.go
```diff
@@ -0,0 +1,89 @@
...
+	engines, err := lhClient.LonghornV1beta1().Engines(namespace).List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		return err
+	}
+
+	volumeEngineMap := map[string][]*v1beta1.Engine{}
+	for i := range engines.Items {
+		e := &engines.Items[i]
+		if e.Spec.VolumeName == "" {
+			// Cannot do anything in the upgrade path if this case happens.
```
innobead, line 43: Do you mean the orphan engine? Could you add more explanation for `this case` in the comment? also add logging for this case.

upgrade/v122to130/upgrade.go
```diff
@@ -0,0 +1,89 @@
...
+			currentEngine = engineList[0]
+		} else {
+			v, err := lhClient.LonghornV1beta1().Volumes(namespace).Get(context.TODO(), volumeName, metav1.GetOptions{})
+			if err != nil {
+				return err
+			}
+			for i := range engineList {
+				if (v.Spec.NodeID != "" && v.Spec.NodeID == engineList[i].Spec.NodeID) ||
+					(v.Status.CurrentNodeID != "" && v.Status.CurrentNodeID == engineList[i].Spec.NodeID) ||
+					(v.Status.PendingNodeID != "" && v.Status.PendingNodeID == engineList[i].Spec.NodeID) {
```
innobead, line 72: Generally, I am thinking we better organize this kind of reusable condition check in some util libraries. Not for now, but it's good to think of it.

controller/volume_controller.go
```diff
@@ -2887,8 +2875,11 @@ func (vc *VolumeController) processMigration(v *longhorn.Volume, es map[string]*
 
 	log := getLoggerForVolume(vc.logger, v).WithField("migrationNodeID", v.Spec.MigrationNodeID)
 
-	// only process if volume is attached
-	if v.Spec.NodeID == "" || len(es) == 0 {
+	// only process if volume is attached and running
+	if v.Spec.NodeID == "" || v.Status.CurrentNodeID == "" || len(es) == 0 {
+		return nil
+	}
+	if v.Status.Robustness != longhorn.VolumeRobustnessDegraded && v.Status.Robustness != longhorn.VolumeRobustnessHealthy {
```
innobead, line 2932: So we only process the migration when the volume is not degraded? Should we add some debug log to better know why the volume can't be migrated here?

controller/volume_controller.go
```diff
@@ -587,27 +590,29 @@ func (vc *VolumeController) ReconcileEngineReplicaState(v *longhorn.Volume, e *l
...
-		}
 
-		setting, err := vc.getAutoBalancedReplicasSetting(v)
-		if err != nil {
-			vc.logger.Warnf(err.Error())
-		}
-		if setting != longhorn.ReplicaAutoBalanceDisabled {
-			if err := vc.replenishReplicas(v, e, rs, ""); err != nil {
-				return err
+			// Migrate local replica when Data Locality is on
```
innobead, line 603: Good to take care this!

innobead (changes requested):
LGTM, but few comments waiting to resolve or answer.

innobead:
> Regression for core tests: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/541/

How is the regression result?

datastore/longhorn.go
shuo-wu: OK... Let's directly error out for now. When multiple active engines are found, continuing the execution is weird...

upgrade/v122to130/upgrade.go
```diff
@@ -0,0 +1,89 @@
...
+	"github.com/pkg/errors"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta1"
+	lhclientset "github.com/longhorn/longhorn-manager/k8s/pkg/client/clientset/versioned"
+)
+
+const (
+	upgradeLogPrefix = "upgrade from v1.2.2 to v1.3.0: "
```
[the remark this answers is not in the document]
shuo-wu: Oh YES. This will be backported to v1.2.3.

upgrade/v122to130/upgrade.go
shuo-wu: Yes. Once there is an orphan engine CR, we cannot do anything in the upgrade path. I will add this explanation in the comment.

controller/volume_controller.go
shuo-wu: We only process the migration when the volume is **neither** degraded nor **healthy**. I am afraid that that kind of message will flood the log.

shuo-wu:
> > Regression for core tests: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/541/
> 
> How is the regression result?

It seems that the ci failed to init the regression test. I will redo it after updating this PR.

controller/volume_controller.go
innobead: Maybe tracing log? I know we use debug by default.

datastore/longhorn.go
shuo-wu: I am not sure if there are any unexpected cases that would lead to the only engine does not set `Spec.Active`. Let me add a warn log for this case first... It should be like:
```
if len(es) == 1 {
	for _, e := range es {
		if !e.Spec.Active {
			log.Warn("xxxxx")
		}
		return e, nil, nil
	}
}
```

innobead: I mean probably u can just remove this if block, and just leave the following new logic to handle.

shuo-wu: I mean, if I remove this block, this function may error out once somehow the only engine of a volume does not set `Spec.Active`. Then the reconcile loop will error out. This is what I want to avoid.

innobead: Got your point, just for `in case`, so the concern is `I am not sure if there are any unexpected cases that would lead to the only engine does not set Spec.Active`. .

innobead:
After the regression is ready and comments are resolved, then we should be able to merge this 👍

controller/volume_controller.go
shuo-wu: The trace log actually does not make too much sense... Do we really need that?

datastore/longhorn.go
shuo-wu: YES. Erroring out would implicitly force users to report the issue once it's really triggered. It's reasonable and good for us but may annoy users. If it's better to expose the issue as early as possible, I am fine with removing this block. What do you think?

innobead: Then, removed 🚀

shuo-wu:
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/545/

shuo-wu:
I don't know why all core tests are passed in my dev cluster but failed in the regression... This PR is ready to go anyway.

innobead (approved):

PR #1126, opened 2021-11-12
CRD structural schema

jenting:
#### Proposal Changes
- Update `k8s/generate_code.sh`
  - Use [generate-group.sh](https://github.com/kubernetes/code-generator/blob/v0.18.0/generate-groups.sh) script instead of manually install the binaries `deepcopy-gen`, `client-gen`, `lister-gen`, `informer-gen`.
  - Use [controller-gen](https://github.com/kubernetes-sigs/controller-tools/tree/v0.7.0/cmd/controller-gen) to generate the CRD structural schema.
  - Use [kustomize](https://github.com/kubernetes-sigs/kustomize/tree/kustomize/v3.10.0/kustomize) to add label to the generated CRD schema and concatenate into one `03-crd.yaml`.
- Update code to fulfill the controller-gen rule
  - The controller-gen does not allow the struct with no data type, for example, `struct{}`, `interface{}`.
  - The controller-gen does not allow the struct fields without json tag.
  - Add description of all the struct fields to make `kubectl explain <crd>` works.
  - Add `+optional` for struct fields as optional to prevent any issue during Longhorn upgrade (otherwise, the field be marked as required).
  - Add kubebuilder marker
    - Overwrite the type of the field `+kubebuilder:validation:Type=<type>`
    - Enum validation `+kubebuilder:validation:Enum`
    - Min/Max validation `+kubebuilder:validation:Minimum`/`+kubebuilder:validation:Maximum`
    - CRD short name `+kubebuilder:resource:shortName=<string>`
    - CRD subresources status `+kubebuilder:subresource:status`
    - print columns `+kubebuilder:printcolumn:name=...`

#### Issue
https://github.com/longhorn/longhorn/issues/791

#### Additional context

- We try to make `kubectl explain <crd>` be clear and in detail for the description. However, lots of fields I'm not pretty sure how to describe it. If reviewers have a good description of these fields, please leave a comment.
- Since the CRD label is generated by `kustomize` now and we can't make it happen that different CRD with the same key but different values like `longhorn-manager: Engine`, `longhorn-manager: Volume`. Since the label is used to filter the CRDs only `kubectl get crds -l longhorn-manager`, so I propose to use an empty value but with the same key `longhorn-manager`. So in the end, all Longhorn CRD label is `longhorn-manager: `.
- For all the kubebuilder markers, I have dumped them. If interested, please refer to [gist](https://gist.github.com/jenting/e8b6477d91061c223416bda2538c2d68).


k8s/pkg/apis/longhorn/v1beta1/backingimage.go
```diff
@@ -18,33 +18,48 @@ const (
...
-	Disks            map[string]string          `json:"disks"`
-	Checksum         string                     `json:"checksum"`
-	SourceType       BackingImageDataSourceType `json:"sourceType"`
-	SourceParameters map[string]string          `json:"sourceParameters"`
-
+	// +optional
+	Disks map[string]string `json:"disks"`
+	// +optional
+	Checksum string `json:"checksum"`
+	// +optional
```
shuo-wu, line 37: It's better to mark some fields are required if possible. We can do it later with a separate ticket tracking it.

k8s/pkg/apis/longhorn/v1beta1/engine.go
```diff
@@ -126,6 +126,7 @@ type EngineSpec struct {
 type EngineStatus struct {
 	InstanceStatus `json:""`
 
+	// +kubebuilder:validation:Type=string
```
shuo-wu, line 129: Why is the current size type string rather than integer?

k8s/pkg/apis/longhorn/v1beta1/backup.go
```diff
@@ -11,42 +11,62 @@ const (
...
 	Labels map[string]string `json:"labels"`
 }
 
+// SnapshotBackupStatus defines the observed state of the Longhorn backup
 type SnapshotBackupStatus struct {
+	// The owner ID which is responsible to reconcile this backup CR.
 	// +optional
 	OwnerID string `json:"ownerID"`
+	// The backup creation state.
+	// Can be "", InProgress, Completed, Error, Unknown.
```
shuo-wu, line 33: Add quotas for all enums: `Can be "", "InProgress", "Completed", "Error", "Unknown".`

k8s/pkg/apis/longhorn/v1beta1/backup.go
```diff
@@ -11,42 +11,62 @@ const (
...
+	// The volume size.
 	// +optional
 	VolumeSize string `json:"volumeSize"`
+	// The volume creation time.
 	// +optional
 	VolumeCreated string `json:"volumeCreated"`
+	// The volume's backing image name.
 	// +optional
 	VolumeBackingImageName string `json:"volumeBackingImageName"`
+	// The last time that the backup was synced into the cluster.
```
shuo-wu, line 69: `The last time that the backup was synced with the remote backup target.`

k8s/pkg/apis/longhorn/v1beta1/backuptarget.go
```diff
@@ -8,24 +8,34 @@ const (
...
 
+// BackupTargetSpec defines the desired state of the Longhorn backup target
 type BackupTargetSpec struct {
+	// The backup target URL.
 	// +optional
 	BackupTargetURL string `json:"backupTargetURL"`
+	// The backup target credential secret.
 	// +optional
 	CredentialSecret string `json:"credentialSecret"`
+	// The backup target poll interval.
```
shuo-wu, line 19: The interval that the cluster needs to run sync with the backup target.

k8s/pkg/apis/longhorn/v1beta1/backup.go
```diff
@@ -11,42 +11,62 @@ const (
...
 	// +optional
 	SnapshotName string `json:"snapshotName"`
+	// The labels of snapshot backup.
 	// +optional
 	Labels map[string]string `json:"labels"`
 }
 
+// SnapshotBackupStatus defines the observed state of the Longhorn backup
 type SnapshotBackupStatus struct {
+	// The owner ID which is responsible to reconcile this backup CR.
```
shuo-wu, line 29: `The node ID on which the controller is responsible to reconcile this backup volume CR.`

similar below and above.

deploy/install/01-prerequisite/03-crd.yaml
```diff
@@ -1,201 +1,740 @@
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.7.0
+  creationTimestamp: null
   labels:
-    longhorn-manager: Engine
-  name: engines.longhorn.io
+    longhorn-manager: ""
```
shuo-wu, line 8: Notice that the value of this generated label is missed. Is there a way to fix that?

shuo-wu (commented):
In general LGTM. There are some minor issues but we can handle them in the future.

- As I mentioned in the below comment, not all fields should be marked as `optional`.
- Some field descriptions are missed.
- It's nice to put the doc/article links you referred into the PR or the script comment.

k8s/generate_code.sh
```diff
@@ -1,29 +1,60 @@
...
+  echo "Prepare to install kustomize"
+	mkdir -p ${GOPATH}/src/github.com/kubernetes-sigs
+	pushd ${GOPATH}/src/github.com/kubernetes-sigs
+	git clone -b ${KUSTOMIZE_VERSION} git@github.com:kubernetes-sigs/kustomize.git 2>/dev/null || true
+	cd kustomize/kustomize
+	go install .
+	popd
+fi
+
+${GOPATH}/src/k8s.io/code-generator/generate-groups.sh \
```
derekbit, line 46: Avoid unexpected permission error
```suggestion
bash ${GOPATH}/src/k8s.io/code-generator/generate-groups.sh \
```

k8s/pkg/apis/longhorn/v1beta1/node.go
```diff
@@ -32,53 +32,90 @@ const (
...
+	// +optional
+	Disks map[string]DiskSpec `json:"disks"`
+	// +optional
+	AllowScheduling bool `json:"allowScheduling"`
+	// +optional
+	EvictionRequested bool `json:"evictionRequested"`
+	// +optional
+	Tags []string `json:"tags"`
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=40
```
derekbit, line 75: Why is the EngineManagerCPURequest's Maximum 40?

k8s/pkg/apis/longhorn/v1beta1/node.go
```diff
@@ -32,53 +32,90 @@ const (
...
+	// +optional
+	EngineManagerCPURequest int `json:"engineManagerCPURequest"`
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=40
```
derekbit, line 79: Same question as above

k8s/pkg/apis/longhorn/v1beta1/node.go
jenting: I check the doc and the chart questions.yaml, the max value is 40.

k8s/pkg/apis/longhorn/v1beta1/backingimage.go
jenting: Yes, that would be great. But since lots of spec fields need to check one by one which is/are required, the only way I could think now is run the E2E test but it required lots of time. 

I agree that we should enhance it in the future.

k8s/pkg/apis/longhorn/v1beta1/engine.go
jenting: The json tag specify to use string rather than integer. So, that's why we need to enforce the type with string.

I did a test, without specific to use string, it failed on json unmarshal when call List API.

k8s/pkg/apis/longhorn/v1beta1/backup.go
jenting: You mean add validation for the enum of status fields as well?

I tried to add validation for all the spec fields, but for status fields, it can only be updated by controller. Therefore, the end user should be unable to change the status field by kubectl.

If we would like to make all the status fields add enum validation, then we need to run full E2E test to double verify it as well.

deploy/install/01-prerequisite/03-crd.yaml
jenting: I have no idea by leverage kustomize tool or other tool.

However,it's a label and the purpose is to filter the CRDs used by longhorn-manager. Without, the corresponding value, we still able to get the longhorn CRDs by "kubectl get crd -l longhorn-manager".

If we really care the label's value, the only way I could think now is hand-craft the label.

jenting:
@shuo-wu Lots of fields I don't know how to add the description for it. If possible, probably you could add the comment on it and I will update the commit 🙏. Or,directly push to my branch if you like, thank u.

k8s/pkg/apis/longhorn/v1beta1/node.go
jenting: [1]

k8s/pkg/apis/longhorn/v1beta1/backup.go
shuo-wu: No. I just talked about the description here.

k8s/pkg/apis/longhorn/v1beta1/node.go
shuo-wu: After the system upgrade, there may be 2 sets of instance manager pods on each node. To avoid the instance manager pods eating up all CPU resources, we just limit the max number to 40%.

shuo-wu:
> @shuo-wu Lots of fields I don't know how to add the description for it. If possible, probably you could add the comment on it and I will update the commit 🙏. Or,directly push to my branch if you like or tackle it in the future because it only affects the display on "kubectl explain".

I know these are nice to have and won't affect the functionality. We don't need to handle them now. Please create a separate Github ticket to track it. Then everyone would help add descriptions later.

shuo-wu:
> @shuo-wu Lots of fields I don't know how to add the description for it. If possible, probably you could add the comment on it and I will update the commit 🙏. Or,directly push to my branch if you like or tackle it in the future because it only affects the display on "kubectl explain".

I know these are nice to have and won't affect the functionality. We don't need to handle them now. Please create a separate Github ticket to track it. Then everyone would help add descriptions later.

k8s/pkg/apis/longhorn/v1beta1/engine.go
shuo-wu: Not sure if removing suffix `,string` from below json tag `json:"currentSize,string"` work or not.

derekbit:
LGTM

k8s/pkg/apis/longhorn/v1beta1/engine.go
jenting: Seems no https://play.golang.org/p/hJifNZNR0l6

shuo-wu: This link is unavailable...

deploy/install/01-prerequisite/03-crd.yaml
shuo-wu: I am fine with leaving the value empty. @innobead What do you think?

k8s/pkg/apis/longhorn/v1beta1/engine.go
jenting: Well, it's available to me. Never mind, I did a quick test, it's not working.
```
package main

import (
	"encoding/json"
	"fmt"
)

type EngineStatus struct {
	CurrentSize int64 `json:"currentSize"`
}

func main() {
	es := EngineStatus{}
	data := []byte(`{"currentSize": "100"}`)

	json.Unmarshal(data, &es)
	fmt.Printf("%v\n", es.CurrentSize)
}
```

shuo-wu: Shouldn't the test be like this?
```
package main

import (
	"encoding/json"
	"fmt"
)

type EngineStatus struct {
	CurrentSize int64 `json:"currentSize"`
}

func main() {
	es := EngineStatus{}
	data := []byte(`{"currentSize": 100}`)

	json.Unmarshal(data, &es)
	fmt.Printf("%v\n", es.CurrentSize)
}
```

This test works fine. `// +kubebuilder:validation:Type=int64` indicates the input data value is an integer. `json:"currentSize"` means the data value type is not required to be string.


But once you modify the definition to `CurrentSize int64 `json:"currentSize,string"``, this does not work anymore.

jenting: But the existing `status.currentSize` in CR is a string type.

shuo-wu: You mean the existing CR field is stored as a string type in ETCD? If that's the case, it's fine to leave it here. Or we can modify this part during the CRD API version bump.

jenting: Yes, existing CR field is stored as a string type in ETCD.

Agree, we could handle it in the v1beta2. I'll create an issue for this.

shuo-wu: Sure. And you can add the issue into this Epic:
https://github.com/longhorn/longhorn/issues/3276

jenting: Or we could wait the PR https://github.com/longhorn/longhorn-manager/pull/1136 be merged, and we focus on CRD structural schema for v1beta2 version only.

Then, we could change the struct from string to int, and implement the conversion in upgrade path.

shuo-wu: There is no need to wait for it... We still need to open another PR to remove all deprecated fields after https://github.com/longhorn/longhorn-manager/pull/1136 merge.

jenting: Okay, new [issue](https://github.com/longhorn/longhorn/issues/3282) created.

shuo-wu (approved):
In general LGTM except for one issue mentioned in https://github.com/longhorn/longhorn-manager/pull/1136#discussion_r754080772

k8s/generate_code.sh
```diff
@@ -1,29 +1,63 @@
...
+  ${GROUP_VERSION} \
+  $@
+
+echo Generating CRD
+controller-gen crd paths=${APIS_DIR}/... output:crd:dir=${CRDS_DIR}
+pushd ${CRDS_DIR}
+kustomize create --autodetect 2>/dev/null || true
+kustomize edit add label longhorn-manager: 2>/dev/null || true
+popd
+kustomize build ${CRDS_DIR} > ${GOPATH}/src/${LH_MANAGER_DIR}/deploy/install/01-prerequisite/03-crd.yaml
```
innobead, line 62: It's great to see we use kustomize

deploy/install/01-prerequisite/03-crd.yaml
innobead: It's still able to filter, so I am fine with that. But it's better to see if we have any internal code impacted by this or even documentation.

k8s/pkg/apis/longhorn/v1beta1/engine.go
```diff
@@ -11,105 +11,175 @@ const (
...
-	RebuildStatus            map[string]*RebuildStatus       `json:"rebuildStatus"`
-	CloneStatus              map[string]*SnapshotCloneStatus `json:"cloneStatus"`
-	Snapshots                map[string]*Snapshot            `json:"snapshots"`
-	SnapshotsError           string                          `json:"snapshotsError"`
-	IsExpanding              bool                            `json:"isExpanding"`
-	LastExpansionError       string                          `json:"lastExpansionError"`
-	LastExpansionFailedAt    string                          `json:"lastExpansionFailedAt"`
+	InstanceStatus `json:""`
+
+	// +kubebuilder:validation:Type=string
```
innobead, line 129: It's weird to use string as a validation, why not `int64`. If needed to change, probably we need to change the json tag?

innobead:
LGTM, and after resolving my comments, then it's good to 🚀 merge.

k8s/pkg/apis/longhorn/v1beta1/engine.go
jenting: Shuo has the same question before.
It's because we store string in the ETCD.

We could change the data type to int64 in Longhorn v1beta2 version, and I had created an issue for this already.

innobead (approved):

PR #1129, opened 2021-11-16
controller: Try best to clean up instance processes

shuo-wu:
https://github.com/longhorn/longhorn/issues/3255

### DIFF
--- a/controller/engine_controller.go
+++ b/controller/engine_controller.go
@@ -411,19 +411,30 @@ func (ec *EngineController) DeleteInstance(obj interface{}) error {
 		return err
 	}
 
-	// Not assigned, safe to delete
+	var im *longhorn.InstanceManager
+	var err error
+	// Not assigned or not updated, try best to delete
 	if e.Status.InstanceManagerName == "" {
-		return nil
-	}
-
-	im, err := ec.ds.GetInstanceManager(e.Status.InstanceManagerName)
-	if err != nil {
-		if !apierrors.IsNotFound(err) {
-			return err
+		if e.Spec.NodeID == "" {
+			log.Warnf("Engine %v does not set instance manager name and node ID, will skip the actual process deletion", e.Name)
+			return nil
 		}
-		// The related node may be directly deleted.
-		log.Warnf("The engine instance manager %v is gone during the engine instance %v deletion. Will do nothing for the deletion", e.Status.InstanceManagerName, e.Name)
-		return nil
+		im, err = ec.ds.GetInstanceManagerByInstance(obj)
+		if err != nil {
+			log.Warnf("Failed to detect instance manager for engine %v, will skip the actual process deletion: %v", e.Name, err)
+			return nil
+		}
+		log.Infof("Try best to clean up the process for engine %v in instance manager %v", e.Name, im.Name)
+	} else {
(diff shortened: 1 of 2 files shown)


innobead (approved):
LGTM, this is the best effort to look up the corresonding instance manager a engine/replica belongs to.

controller/engine_controller.go
```diff
@@ -411,19 +411,26 @@ func (ec *EngineController) DeleteInstance(obj interface{}) error {
...
+	var err error
+	// Not assigned or not updated, try best to delete
+	if e.Status.InstanceManagerName == "" && e.Spec.NodeID != "" {
+		im, err = ec.ds.GetInstanceManagerByInstance(obj)
+		if err != nil {
+			log.Warnf("failed to detect instance manager for engine %v, will skip the actual process deletion: %v", e.Name, err)
+			return nil
+		}
+		log.Infof("Try best to clean up the process for engine %v in instance manager %v", e.Name, im.Name)
+	} else {
```
PhanLe1010, line 424: To avoid print out the wrong log, can we do one more check here:
```suggestion
	} else {
		if e.Status.InstanceManagerName == "" {
		    return nil
	        }
	 ...
```

controller/replica_controller.go
```diff
@@ -523,19 +523,26 @@ func (rc *ReplicaController) DeleteInstance(obj interface{}) error {
...
+	var err error
+	// Not assigned or not updated, try best to delete
+	if r.Status.InstanceManagerName == "" && r.Spec.NodeID != "" {
+		im, err = rc.ds.GetInstanceManagerByInstance(obj)
+		if err != nil {
+			log.Warnf("failed to detect instance manager for replica %v, will skip the actual process deletion: %v", r.Name, err)
+			return nil
+		}
+		log.Infof("Try best to clean up the process for replica %v in instance manager %v", r.Name, im.Name)
+	} else {
```
PhanLe1010, line 536: Same as above

PhanLe1010 (changes requested):

controller/engine_controller.go
```diff
@@ -411,19 +411,30 @@ func (ec *EngineController) DeleteInstance(obj interface{}) error {
...
-	if err != nil {
-		if !apierrors.IsNotFound(err) {
-			return err
+		if e.Spec.NodeID == "" {
+			log.Warnf("Engine %v does not set instance manager name and node ID, will skip the actual process deletion", e.Name)
+			return nil
+		}
+		im, err = ec.ds.GetInstanceManagerByInstance(obj)
+		if err != nil {
+			log.Warnf("failed to detect instance manager for engine %v, will skip the actual process deletion: %v", e.Name, err)
```
jenting, line 424: nit: do we have the conversion for the log first character should be upper case _or_ lower case?

jenting (approved):

PhanLe1010 (approved):
LGTM

controller/engine_controller.go
shuo-wu: It's better to follow the conversion. Thanks for pointing it out! I have updated the PR.

jenting (approved):

innobead (approved):

PR #1133, opened 2021-11-17
Fix UI can't set empty backup target credential secret

jenting:
#### Proposal change

If the secret is not found, do not proceed with it.
Also, remove duplicate retry on conflict since it's being handled at
- https://github.com/longhorn/longhorn-manager/blob/dbdea97/api/router.go#L61
- https://github.com/longhorn/longhorn-manager/blob/dbdea97/api/router.go#L47
- https://github.com/longhorn/longhorn-manager/blob/dbdea97/api/router.go#L25-L35

#### Issue
https://github.com/longhorn/longhorn/issues/3261

### DIFF
--- a/api/setting.go
+++ b/api/setting.go
@@ -1,7 +1,6 @@
 package api
 
 import (
-	"fmt"
 	"net/http"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


innobead (approved):
LGTM

PR #1136, opened 2021-11-17
Upgrade CRD version from v1beta1 to v1beta2

derekbit:
The PR is to upgrade CRD version from v1beta1 to v1beta2, and the following tasks are done.

- Upgrade CRD version to v1beta2
- Replace status.Conditions's map datatype with splice and refactor the related functions and test files
- Add upgrade path from v1beta1 to v1beta2
- Remove upgrade path from v1alpha1
- Refactor doAPIVersionUpgrade
  - upgradable
    - v1beta1
    - v1beta2 (include new installation)
  - non-upgradable 
    - v1alpha1
- Update crd.yaml by k8s/generate_code.go

https://github.com/longhorn/longhorn/issues/2955


jenting:
> Change v1alpha1/v1beta1 upgrade path to v1alpha1/v1beta2

Shouldn't the v1alpha1 upgrade to v1beta1, and then v1beta1 upgrade to v1beta2?

derekbit:
> > Change v1alpha1/v1beta1 upgrade path to v1alpha1/v1beta2
> 
> Shouldn't the v1alpha1 upgrade to v1beta1, and then v1beta1 upgrade to v1beta2?

Sure, I think it's feasible.

derekbit:
> Shouldn't the v1alpha1 upgrade to v1beta1, and then v1beta1 upgrade to v1beta2?

I think the difference between v1beta1 and v1beta2 is only status.Conditions, so the seems v1alpha1->v1beta1->v1beta2 is not necessary. I also handled the status.Conditions in v1alpha1 to v1beta2 path. Any concern?

joshimoo:
@innobead @shuo-wu What do you guys think, should we consider dropping any upgrade path pre GA(1.0) for the v1.3 release?
There is still some old code in the controllers that could be removed at the same time as this as well, as well as dropping the v1alpha crds.

jenting:
> > Shouldn't the v1alpha1 upgrade to v1beta1, and then v1beta1 upgrade to v1beta2?
> 
> I think the difference between v1beta1 and v1beta2 is only status.Conditions. I also handled the status.Conditions in v1alpha1 to v1beta2 path, so v1alpha1->v1beta1->v1beta2 seems not necessary. Any concern?

I thought we only need to have an upgrade from v1beta1 to v1beta2 is enough so we don't have to change the upgrade code from v070 to v121.

derekbit:
> > > Shouldn't the v1alpha1 upgrade to v1beta1, and then v1beta1 upgrade to v1beta2?
> > 
> > 
> > I think the difference between v1beta1 and v1beta2 is only status.Conditions. I also handled the status.Conditions in v1alpha1 to v1beta2 path, so v1alpha1->v1beta1->v1beta2 seems not necessary. Any concern?
> 
> I thought we only need to have an upgrade from v1beta1 to v1beta2 is enough so we don't have to change the upgrade code from v070 to v121.

Agree.
Should we remove the v1alpha1 to v1beta1 upgrade path?

shuo-wu:
Should we create an Epic to track the removal of these old codes? There was a ticket that tracks the restore/DR volume part: https://github.com/longhorn/longhorn/issues/1685

cc @innobead

innobead:
> @innobead @shuo-wu What do you guys think, should we consider dropping any upgrade path pre GA(1.0) for the v1.3 release? There is still some old code in the controllers that could be removed at the same time as this as well, as well as dropping the v1alpha crds.

YES, I think we don't need to keep maintaining v1alpha CRDs and also the upgrade path before 1.1. 

@shuo-wu YES, we can create an epic to cover all issues/stories we would like to clean up, even though this seems moving part. https://github.com/longhorn/longhorn/issues/3276

types/condition.go
```diff
@@ -28,28 +31,35 @@ func getUnknownCondition(conditionType string) longhorn.Condition {
...
 	condition.Reason = reason
 	condition.Message = message
-	conditions[conditionType] = condition
+
+	if conditionIndex == -1 {
+		conditions = append(conditions, condition)
+	} else {
+		conditions[conditionIndex] = condition
+	}
+
```
jenting, line 63: Do you think we could check the `condition.Status == longhorn.ConditionStatusUnknown` rather than return the index `-1` and check the `index == -1`?

controller/engine_image_controller_test.go
```diff
@@ -114,6 +114,17 @@ func createEngineImageDaemonSetPod(name string, containerReadyStatus bool, nodeI
 	return pod
 }
 
+func updateOrAppendCondition(conditions []longhorn.Condition, condition longhorn.Condition) []longhorn.Condition {
```
jenting, line 117: Consider move the function to `types/condition.go`.

controller/volume_controller.go
```diff
@@ -446,13 +446,15 @@ func (vc *VolumeController) syncVolume(key string) (err error) {
...
-			newStatus.Conditions[conditionType] = existingCondition
+	for i, newCondition := range newStatus.Conditions {
+		for _, existingCondition := range existingStatus.Conditions {
+			if existingCondition.Type == newCondition.Type {
+				if existingCondition.Status == newCondition.Status &&
+					existingCondition.Reason == newCondition.Reason &&
+					existingCondition.Message == newCondition.Message {
+					newStatus.Conditions[i] = existingCondition
+				}
+			}
```
jenting, line 457:
```suggestion
			if existingCondition.Type == newCondition.Type &&
				existingCondition.Status == newCondition.Status &&
				existingCondition.Reason == newCondition.Reason &&
				existingCondition.Message == newCondition.Message {
				newStatus.Conditions[i] = existingCondition
			}
```

controller/volume_controller_test.go
```diff
@@ -46,8 +46,15 @@ func setVolumeConditionWithoutTimestamp(originConditions map[string]longhorn.Con
...
 	}
-	conditions[conditionType] = condition
-	return conditions
+
+	for i := range conditions {
+		if conditions[i].Type == conditionType {
+			conditions[i] = condition
+			return conditions
+		}
+	}
```
jenting, line 55: Consider use the function `GetCondition`.

app/recurring_job.go
```diff
@@ -750,7 +750,15 @@ func (job *Job) findARandomReadyNode(v *longhornclient.Volume) (string, error) {
...
 	var readyNodeList []string
 	for _, node := range nodeCollection.Data {
-		if readyConditionInterface, ok := node.Conditions[longhorn.NodeConditionTypeReady]; ok {
+		var readyConditionInterface map[string]interface{}
+		for i := range node.Conditions {
+			con := node.Conditions[i].(map[string]interface{})
+			if con["type"] == longhorn.NodeConditionTypeReady {
+				readyConditionInterface = con
+			}
+		}
```
jenting, line 759: Consider use the function `GetCondition`.

app/recurring_job.go
```diff
@@ -750,7 +750,15 @@ func (job *Job) findARandomReadyNode(v *longhornclient.Volume) (string, error) {
...
+
+		if readyConditionInterface != nil {
```
jenting, line 761: Not sure _why_ below need to perform json marshal/unmarshal, can't we directly use the `node.Conditions`?

controller/engine_image_controller_test.go
derekbit: Sure.

app/recurring_job.go
derekbit: Hmm, need to check the generic json marshal/unmarshal logic.

app/recurring_job.go
derekbit: Sure.

controller/volume_controller_test.go
derekbit: Sure.

types/condition.go
derekbit: Seems OK after quick check.

upgrade/v1beta1/upgrade.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+
+	volumes, err := lhClientV1beta1.LonghornV1beta1().Volumes(namespace).List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		return errors.Wrap(err, "upgrade: v1beta1: unable to list volumes")
+	}
+	for _, old := range volumes.Items {
+		new := &longhorn.Volume{}
+
+		if err := copier.Copy(&new, &old); err != nil {
```
shuo-wu, line 44: Will there be an error since the field `condition` is different between old volume definition and the new one?

deploy/install/01-prerequisite/03-crd.yaml
```diff
@@ -16,6 +16,39 @@ spec:
   scope: Namespaced
   versions:
   - name: v1beta1
+    served: true
```
shuo-wu, line 19: @jenting Can `controller-gen` in your PR generate this kind of old version info?

upgrade/upgrade.go
```diff
@@ -230,6 +207,31 @@ func doAPIVersionUpgrade(namespace string, config *restclient.Config, lhClient *
...
+	if err != nil {
+		return false, errors.Wrap(err, "unable to get clientset for v1beta1")
+	}
+
+	scheme := runtime.NewScheme()
+	if err := longhorn.SchemeBuilder.AddToScheme(scheme); err != nil {
+		return false, errors.Wrap(err, "unable to create scheme for v1beta1")
+	}
+
+	setting, err := lhClient.LonghornV1beta2().Settings(namespace).Get(context.TODO(), string(types.SettingNameDefaultEngineImage), metav1.GetOptions{})
```
shuo-wu, line 221: Shouldn't this be `LonghornV1beta1`?

upgrade/upgrade.go
```diff
@@ -230,6 +207,31 @@ func doAPIVersionUpgrade(namespace string, config *restclient.Config, lhClient *
 	return nil
 }
 
+func canUpgrade(config *restclient.Config, namespace string) (bool, error) {
+	lhClient, err := lhclientset.NewForConfig(config)
```
shuo-wu, line 211: Shouldn't this clientset be `lhclientset_v1beta1` if you are checking CRD API v1beta1?

shuo-wu (commented):
- For the 1st commit, it's better to split it into 2 commits. Then we can fcous on the `condition` modification during the review:
    1. Rename the dir/package to v1beta2 and update the imports
    2. Modify the condition definition
- Is it fine to skip creating new version CRs without field `condition` in `UpgradeFromV1beta1ToV1beta2`?

api/model.go
```diff
@@ -264,9 +264,18 @@ type Node struct {
 	ReplicaManagerCPURequest int                           `json:"replicaManagerCPURequest"`
 }
 
+type DiskStatus struct {
+	Conditions       map[string]longhorn.Condition `json:"conditions"`
```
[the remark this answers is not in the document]
derekbit: Yes, we can replace all Conditions' map type with slice type. But the longhorn-ui also need to be modified. After discussion with @jenting , we think we can convert Condtions (slice) to map tyoe in model/api.go, so the ui doesn't need change.

Do you prefer consistent Conditions type?

api/model.go
```diff
@@ -264,9 +264,18 @@ type Node struct {
...
+	StorageScheduled int64                         `json:"storageScheduled"`
+	StorageMaximum   int64                         `json:"storageMaximum"`
+	ScheduledReplica map[string]int64              `json:"scheduledReplica"`
+	DiskUUID         string                        `json:"diskUUID"`
+}
+
 type DiskInfo struct {
 	longhorn.DiskSpec
-	longhorn.DiskStatus
+	DiskStatus
```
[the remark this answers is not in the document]
derekbit: Same as above. The Conditions' datatype in DiskStatus and longhorn.DiskStatus to achieve the above coversion.

deploy/install/01-prerequisite/03-crd.yaml
jenting: Do you mean for v1beta1, still uses the old CRD schema that with `x-kubernetes-preserve-unknown-fields: true`, but for new v1beta2 uses the new CRD structural schema?

upgrade/v1beta1/upgrade.go
derekbit: No, copier.Copy can copy to different struct and will skip the fields having different datatype.

upgrade/upgrade.go
derekbit: Yes. Using old lhclientset_v1beta1 is better.

upgrade/upgrade.go
derekbit: Yes, LonghornV1beta1 is better although LonghornV1beta1 and LonghornV1beta2 for setting are equivalent.

upgrade/upgrade.go
```diff
@@ -186,22 +188,38 @@ func doAPIVersionUpgrade(namespace string, config *restclient.Config, lhClient *
...
 			break
 		}
 		fallthrough
 	case types.CRDAPIVersionV1alpha1:
-		if types.CurrentCRDAPIVersion != types.CRDAPIVersionV1beta1 {
+		logrus.Infof("upgrading from %v to %v", types.CRDAPIVersionV1beta1, types.CurrentCRDAPIVersion)
+
+		if types.CurrentCRDAPIVersion != types.CRDAPIVersionV1beta2 {
 			return fmt.Errorf("cannot upgrade from %v to %v directly", types.CRDAPIVersionV1alpha1, types.CurrentCRDAPIVersion)
 		}
```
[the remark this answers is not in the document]
derekbit: Yes, I'm also confused when refine this logic. CurrentCRDAPIVersion should be always CRDAPIVersionV1beta2, right?

derekbit:
> * For the 1st commit, it's better to split it into 2 commits. Then we can fcous on the `condition` modification during the review:
>   
>   1. Rename the dir/package to v1beta2 and update the imports
>   2. Modify the condition definition

No prob. Sounds better.

> * Is it fine to skip creating new version CRs without field `condition` in `UpgradeFromV1beta1ToV1beta2`?

Yes, the old CR will be interpreted  with V1beta2 API.

controller/volume_controller_test.go
derekbit: The  data type of the condition is map[string]interface and not compatible with GetCondition's argument.
Do you have any idea for this?

jenting: for example
```go
        i, _ := types.GetCondition(conditions, conditionType)
	if i != -1 {
		return conditions
	}
	return append(conditions, longhorn.Condition{
		Type:    conditionType,
		Status:  conditionValue,
		Reason:  reason,
		Message: message,
	})
```

derekbit: @jenting  Sorry. No prob for this. The problem is for the next comment.

derekbit: @jenting Sorry, the prob is for the comment in next comment (app/recurring_job.go).

app/recurring_job.go
derekbit: @jenting The reason is the Conditions' datatype is map[string]interface{}, so need to marshal here again.

app/recurring_job.go
derekbit: @jenting The data type of the condition is map[string]interface and not compatible with GetCondition's argument.
Do you have any idea for this?

api/model.go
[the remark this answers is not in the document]
derekbit: @innobead Do you think the consistent `condition ` datatype is better?

app/recurring_job.go
jenting: No, since it's the data struct for RESTful API. And as we discussed, we don't want to change the RESTful API interface.
So, let's keep it as is.

app/recurring_job.go
jenting: Thanks for the explanation.

types/condition.go
```diff
@@ -28,28 +31,41 @@ func getUnknownCondition(conditionType string) longhorn.Condition {
...
 	condition.Status = conditionValue
 	condition.Reason = reason
 	condition.Message = message
-	conditions[conditionType] = condition
-	return conditions
+
+	return UpdateOrAppendCondition(conditions, condition)
+}
+
+func UpdateOrAppendCondition(conditions []longhorn.Condition, condition longhorn.Condition) []longhorn.Condition {
```
jenting, line 62: Since the `UpdateOrAppendCondition` is only used by engine_image_controller_test.go

I'd suggest considering changing the function to the private function and the public function is the `SetCondition` function. This makes there is only one public function for set conditions. Also, you could consider merging the `updateOrAppendCondition` code into `SetCondition` after the suggested change.

controller/backing_image_controller.go
```diff
@@ -376,11 +376,13 @@ func (bic *BackingImageController) handleBackingImageDataSource(bi *longhorn.Bac
 				return err
 			}
 			for _, node := range nodes {
-				if types.GetCondition(node.Status.Conditions, longhorn.NodeConditionTypeSchedulable).Status != longhorn.ConditionStatusTrue {
+				condition := types.GetCondition(node.Status.Conditions, longhorn.NodeConditionTypeSchedulable)
+				if condition.Status != longhorn.ConditionStatusTrue {
```
jenting, line 380: After removing the index change, I think all these changes are unnecessary. (Please revert this change across all the files)

jenting (changes requested):

types/condition.go
derekbit: Okay

upgrade/v1beta1/upgrade.go
```diff
@@ -0,0 +1,186 @@
...
+		if apierrors.IsNotFound(err) {
+			// cannot find the setting,
+			logrus.Infof("Longhorn CRD API v1beta1 not found")
+			return true, nil
+		}
+
+		return false, errors.Wrap(err, "unable to verify if version matches v1beta1")
+	}
+
+	return false, errors.Wrap(err, fmt.Sprintf("unable to upgrade from %v directly", setting.Value))
```
jenting, line 185:
```suggestion
        return false, errors.Wrapf(err, "unable to upgrade from %v directly", setting.Value)
```

upgrade/v1beta1/upgrade.go
```diff
@@ -0,0 +1,186 @@
...
+	scheme := runtime.NewScheme()
+	if err := longhorn_v1beta1.SchemeBuilder.AddToScheme(scheme); err != nil {
+		return false, errors.Wrap(err, "unable to create scheme for v1beta1")
+	}
+
+	setting, err := lhClient.LonghornV1beta1().Settings(namespace).Get(context.TODO(), string(types.SettingNameDefaultEngineImage), metav1.GetOptions{})
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			// cannot find the setting,
+			logrus.Infof("Longhorn CRD API v1beta1 not found")
```
jenting, line 178:
```suggestion
                        logrus.Info("Longhorn CRD API v1beta1 not found")
```

upgrade/v1beta1/upgrade.go
```diff
@@ -0,0 +1,186 @@
...
+	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
+	lhclientset "github.com/longhorn/longhorn-manager/k8s/pkg/client/clientset/versioned"
+)
+
+const (
+	upgradeLogPrefix = "upgrade from v1beta1 to v1beta2:"
+)
+
+func UpgradeFromV1beta1ToV1beta2(config *restclient.Config, namespace string, lhClient *lhclientset.Clientset) error {
+	lhClientV1beta1, err := lhclientset_v1beta1.NewForConfig(config)
```
jenting, line 29: Consider add at the function `UpgradeFromV1beta1ToV1beta2` beginning, so, all the below `return err` will have the same log prefix. [1]
```go
        defer func() {
		err = errors.Wrapf(err, upgradeLogPrefix+" failed")
	}()
```

upgrade/v1beta1/upgrade.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+
+	scheme := runtime.NewScheme()
+	if err := longhorn_v1beta1.SchemeBuilder.AddToScheme(scheme); err != nil {
+		return errors.Wrap(err, "unable to create scheme for v1beta1")
+	}
+
+	volumes, err := lhClientV1beta1.LonghornV1beta1().Volumes(namespace).List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		return errors.Wrap(err, "upgrade: v1beta1: unable to list volumes")
```
jenting, line 41:
```suggestion
                return errors.Wrap(err, "unable to list volumes")
```

upgrade/v1beta1/upgrade.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+	for _, old := range volumes.Items {
```
jenting, line 43: Consider splits below code to multiple functions. For example:
```go
    upgradeVolumes
    upgradeEngineImages
    upgradeBackupTargets
    upgradeNodes
```

upgrade/v1beta1/deploy/install/01-prerequisite/02-serviceaccount.yaml
```diff
@@ -0,0 +1,68 @@
+apiVersion: v1
```
jenting, line 1: TBH, I don't understand _why_ we need to have these files
```
v1beta1/deploy/*
```

upgrade/upgrade.go
```diff
@@ -147,61 +147,53 @@ func doAPIVersionUpgrade(namespace string, config *restclient.Config, lhClient *
...
+			ObjectMeta: metav1.ObjectMeta{
+				Name: string(types.SettingNameCRDAPIVersion),
+			},
+			Value: types.CurrentCRDAPIVersion,
 		}
-		fallthrough
-	case types.CRDAPIVersionV1alpha1:
-		if types.CurrentCRDAPIVersion != types.CRDAPIVersionV1beta1 {
-			return fmt.Errorf("cannot upgrade from %v to %v directly", types.CRDAPIVersionV1alpha1, types.CurrentCRDAPIVersion)
+		_, err = lhClient.LonghornV1beta2().Settings(namespace).Create(context.TODO(), crdAPIVersionSetting, metav1.CreateOptions{})
```
jenting, line 185: Would there be a possibility that the Setting `"crd-api-version"` value is empty?
If yes, then I think the `Create` API call would fail.

upgrade/v1beta1/deploy/install/01-prerequisite/02-serviceaccount.yaml
derekbit: I think we don't need, actually. Just copy the files like the v1alpha1.

api/model.go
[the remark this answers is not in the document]
shuo-wu: I prefer consistent Condition definition in the backend part. For the UI, yes, I can make convert in api/model.go.

api/model.go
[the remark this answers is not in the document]
shuo-wu: OK. You can put the explanation as a comment here.

upgrade/upgrade.go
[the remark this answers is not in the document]
shuo-wu: Yes. Since you already refactored this part. I will ignore it.

api/model.go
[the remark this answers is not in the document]
innobead: Backend good enough, because at the end, the interface for users is actually our custom resources, UI just a wrapper.

upgrade/upgrade.go
derekbit: I think we should remove Create and just return error. It's for the upgrade from v1alpha1.

shuo-wu (changes requested):
The 1st commit should be: **Copy** v1beta1 to upgrade/v1beta1. And the 2nd one should be Rename v1beta1 to v1beta2.
Otherwise, all git histories related to type definitions will be lost. Sometimes these histories are useful during debugging.

deploy/install/01-prerequisite/03-crd.yaml
shuo-wu: I mean, by using `controller-gen` in your PR, will the following v1beta1 info be reserved/generated in the v1beta2 CRD YAML file?
https://github.com/longhorn/longhorn-manager/pull/1126/commits/86eabc9a11cfde7ca46e6610248fb4d98da882ec

derekbit:
> The 1st commit should be: **Copy** v1beta1 to upgrade/v1beta1. And the 2nd one should be Rename v1beta1 to v1beta2. Otherwise, all git histories related to type definitions will be lost. Sometimes these histories are useful during debugging.

@shuo-wu Thanks for the tip.

derekbit:
@jenting @shuo-wu 
Please help to check the 2nd commit which implements https://github.com/longhorn/longhorn/issues/3295 after https://github.com/longhorn/longhorn-manager/pull/1126 is merged.
Now, v1beta1 and v1beta2 are both generated by k8s/generate_code.sh.

deploy/install/01-prerequisite/03-crd.yaml
derekbit: @shuo-wu @jenting 
I think this is just handled in my 2nd commit, please help check it. Thank you.

jenting (commented):
I'm thinking that probably we could remove the upgrade/v1beta1/deploy and upgrade/v1beta1/k8s (upgrade/v1beta1/types probably?), so all the informer/listers/... of v1beta1 and v1beta2 are under the k8s folder.

jenting:
> I'm thinking that probably we could remove the upgrade/v1beta1/deploy and upgrade/v1beta1/k8s (upgrade/v1beta1/types probably?), so all the informer/listers/... of v1beta1 and v1beta2 are under the k8s folder.

derekbit:
> > I'm thinking that probably we could remove the upgrade/v1beta1/deploy and upgrade/v1beta1/k8s (upgrade/v1beta1/types probably?), so all the informer/listers/... of v1beta1 and v1beta2 are under the k8s folder.

I think we can do that.
But now, the layout is generated  informer/listers/... are under `longhorn-manager/k8s/pkg/client/`.
Need to change to `longhorn-manager/k8s/pkg/client/v1beta2`, and the v1beta1's generated files will be moved to `longhorn-manager/k8s/pkg/client/v1beta1`. Is it available?

jenting:
> > > I'm thinking that probably we could remove the upgrade/v1beta1/deploy and upgrade/v1beta1/k8s (upgrade/v1beta1/types probably?), so all the informer/listers/... of v1beta1 and v1beta2 are under the k8s folder.
> 
> I think we can do that. But now, the layout is generated informer/listers/... are under `longhorn-manager/k8s/pkg/client/`. Need to change to `longhorn-manager/k8s/pkg/client/v1beta2`, and the v1beta1's generated files will be moved to `longhorn-manager/k8s/pkg/client/v1beta1`. Is it available?

Let me check how other open-source deal with it.

upgrade/v1beta1/upgrade.go
```diff
@@ -0,0 +1,214 @@
...
+			continue
+		}
+	}
+	return nil
+}
+
+func upgradeEngineImages(namespace string, lhClientV1beta1 *lhclientsetV1beta1.Clientset, lhClient *lhclientset.Clientset) error {
+	engineImages, err := lhClientV1beta1.LonghornV1beta1().EngineImages(namespace).List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		return errors.Wrap(err, "upgrade: v1beta1: unable to list engine images")
```
jenting, line 129:
```suggestion
                return errors.Wrap(err, "unable to list engine images")
```

upgrade/v1beta1/upgrade.go
```diff
@@ -0,0 +1,214 @@
...
+			continue
+		}
+	}
+	return nil
+}
+
+func upgradeBackupTargets(namespace string, lhClientV1beta1 *lhclientsetV1beta1.Clientset, lhClient *lhclientset.Clientset) error {
+	backupTargets, err := lhClientV1beta1.LonghornV1beta1().BackupTargets(namespace).List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		return errors.Wrap(err, "upgrade: v1beta1: unable to list engine images")
```
jenting, line 159:
```suggestion
                return errors.Wrap(err, "unable to list engine images")
```

upgrade/v1beta1/upgrade.go
```diff
@@ -0,0 +1,214 @@
...
+			continue
+		}
+	}
+	return nil
+}
+
+func upgradeNodes(namespace string, lhClientV1beta1 *lhclientsetV1beta1.Clientset, lhClient *lhclientset.Clientset) error {
+	nodes, err := lhClientV1beta1.LonghornV1beta1().Nodes(namespace).List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		return errors.Wrap(err, "upgrade: v1beta1: unable to list engine images")
```
jenting, line 189:
```suggestion
                return errors.Wrap(err, "unable to list engine images")
```

upgrade/v1beta1/upgrade.go
```diff
@@ -0,0 +1,214 @@
...
+		conditions, err := copyConditions(old.Status.Conditions)
+		if err != nil {
+			return errors.Wrap(err, "upgrade: copy volume conditions")
+		}
+		new.Status.Conditions = conditions
+
+		new, err = lhClient.LonghornV1beta2().Volumes(namespace).UpdateStatus(context.TODO(), new, metav1.UpdateOptions{})
+		if err != nil {
+			if !apierrors.IsConflict(err) {
+				return errors.Wrapf(err, "failed to convert v1beta1 to v1beta2 for %v status %v", old.Kind, old.Name)
```
jenting, line 117: The code already wrapped of these message
```go
        defer func() {
		err = errors.Wrapf(err, upgradeLogPrefix+" failed")
	}()
```

Therefore, you don't have to specify the message `convert v1beta1 to v1beta2`.

```suggestion
                                return errors.Wrapf(err, "failed to update %v status %v", old.Kind, old.Name)
```

[1]

upgrade/v1beta1/upgrade.go
```diff
@@ -0,0 +1,214 @@
...
+		conditions, err := copyConditions(old.Status.Conditions)
+		if err != nil {
+			return errors.Wrap(err, "upgrade: copy engineImage conditions")
+		}
+		new.Status.Conditions = conditions
+
+		new, err = lhClient.LonghornV1beta2().EngineImages(namespace).UpdateStatus(context.TODO(), new, metav1.UpdateOptions{})
+		if err != nil {
+			if !apierrors.IsConflict(err) {
+				return errors.Wrapf(err, "failed to convert v1beta1 to v1beta2 for %v status %v", old.Kind, old.Name)
```
jenting, line 147: [1]

upgrade/v1beta1/upgrade.go
```diff
@@ -0,0 +1,214 @@
...
+		conditions, err := copyConditions(old.Status.Conditions)
+		if err != nil {
+			return errors.Wrap(err, "upgrade: copy backupTarget conditions")
+		}
+		new.Status.Conditions = conditions
+
+		new, err = lhClient.LonghornV1beta2().BackupTargets(namespace).UpdateStatus(context.TODO(), new, metav1.UpdateOptions{})
+		if err != nil {
+			if !apierrors.IsConflict(err) {
+				return errors.Wrapf(err, "failed to convert v1beta1 to v1beta2 for %v status %v", old.Kind, old.Name)
```
jenting, line 177: [1]

upgrade/v1beta1/upgrade.go
```diff
@@ -0,0 +1,214 @@
...
+		conditions, err := copyConditions(old.Status.Conditions)
+		if err != nil {
+			return errors.Wrap(err, "upgrade: copy node conditions")
+		}
+		new.Status.Conditions = conditions
+
+		new, err = lhClient.LonghornV1beta2().Nodes(namespace).UpdateStatus(context.TODO(), new, metav1.UpdateOptions{})
+		if err != nil {
+			if !apierrors.IsConflict(err) {
+				return errors.Wrapf(err, "failed to convert v1beta1 to v1beta2 for %v status %v", old.Kind, old.Name)
```
jenting, line 207: [1]

derekbit:
> > > > I'm thinking that probably we could remove the upgrade/v1beta1/deploy and upgrade/v1beta1/k8s (upgrade/v1beta1/types probably?), so all the informer/listers/... of v1beta1 and v1beta2 are under the k8s folder.
> > 
> > 
> > I think we can do that. But now, the layout is generated informer/listers/... are under `longhorn-manager/k8s/pkg/client/`. Need to change to `longhorn-manager/k8s/pkg/client/v1beta2`, and the v1beta1's generated files will be moved to `longhorn-manager/k8s/pkg/client/v1beta1`. Is it available?
> 
> Let me check how other open-source deal with it.

Leaving a note here.
We can replace `GROUP_VERSION="longhorn:v1beta1"` in `k8s/generate_code.sh` with `GROUP_VERSION="longhorn:v1beta1,v1beta2"`, and generate v1beta1/v1beta2's clienset/informers/... to `k8s/pkg/client`. The `upgrade/v1beta2` is not needed.

upgrade/upgrade.go
```diff
@@ -148,61 +148,53 @@ func doAPIVersionUpgrade(namespace string, config *restclient.Config, lhClient *
...
+		if err != nil {
+			return errors.Wrap(err, "cannot create CRDAPIVersionSetting")
 		}
-		if err := v1alpha1.UpgradeFromV1alpha1ToV1beta1(config, namespace, lhClient); err != nil {
+		logrus.Infof("new %v installation", types.CurrentCRDAPIVersion)
```
jenting, line 190: nit
```suggestion
                logrus.Infof("New %v installation", types.CurrentCRDAPIVersion)
```

upgrade/upgrade.go
```diff
@@ -148,61 +148,53 @@ func doAPIVersionUpgrade(namespace string, config *restclient.Config, lhClient *
...
+	case types.CRDAPIVersionV1beta1:
+		logrus.Infof("upgrading from %v to %v", types.CRDAPIVersionV1beta1, types.CurrentCRDAPIVersion)
```
jenting, line 192: nit:
```suggestion
                logrus.Infof("Upgrading from %v to %v", types.CRDAPIVersionV1beta1, types.CurrentCRDAPIVersion)
```

k8s/generate_code.sh
```diff
@@ -7,7 +7,7 @@ set -o pipefail
 LH_MANAGER_DIR="github.com/longhorn/longhorn-manager"
 OUTPUT_DIR="${LH_MANAGER_DIR}/k8s/pkg/client"
 APIS_DIR="${LH_MANAGER_DIR}/k8s/pkg/apis"
-GROUP_VERSION="longhorn:v1beta1"
+GROUP_VERSION="longhorn:v1beta1,v1beta2"
```
jenting, line 10: :100:

upgrade/v122to123/upgrade.go
```diff
@@ -99,7 +99,7 @@ func upgradeBackups(namespace string, lhClient *lhclientset.Clientset) (err erro
 		backup.Status.URL = backupStatus.BackupURL
 		backup.Status.Error = backupStatus.Error
 		backup.Status.SnapshotName = backupStatus.SnapshotName
-		backup.Status.State = engineapi.ConvertEngineBackupState(backupStatus.State)
+		backup.Status.State = longhorn.BackupState(engineapi.ConvertEngineBackupState(backupStatus.State))
```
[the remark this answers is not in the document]
derekbit: engineapi.ConvertEngineBackupState return v1beta2.BackupState, but backup.Status.State is v1beta1.BackupState, so I do a type casting here.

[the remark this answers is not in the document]
derekbit: engineapi.ConvertEngineBackupState return v1beta2.BackupState, but backup.Status.State is v1beta1.BackupState, so I do type casting here.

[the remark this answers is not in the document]
derekbit: BTW, should the APIs in upgrade/vxxxtoxxx/upgrade.go be upgraded to v1beta2?
I'm not sure, so leave this to @shuo-wu's refactoring task.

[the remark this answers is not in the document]
jenting: > engineapi.ConvertEngineBackupState return v1beta2.BackupState, but backup.Status.State is v1beta1.BackupState, so I do type casting here.

:thinking: Isn't this down upgrade from v1beta2 -> v1beta1?

upgrade/upgrade.go
```diff
@@ -148,61 +148,53 @@ func doAPIVersionUpgrade(namespace string, config *restclient.Config, lhClient *
...
-			return fmt.Errorf("cannot upgrade from %v to %v directly", types.CRDAPIVersionV1alpha1, types.CurrentCRDAPIVersion)
+		_, err = lhClient.LonghornV1beta2().Settings(namespace).Create(context.TODO(), crdAPIVersionSetting, metav1.CreateOptions{})
+		if err != nil {
+			return errors.Wrap(err, "cannot create CRDAPIVersionSetting")
 		}
-		if err := v1alpha1.UpgradeFromV1alpha1ToV1beta1(config, namespace, lhClient); err != nil {
+		logrus.Infof("New %v installation", types.CurrentCRDAPIVersion)
+	case types.CRDAPIVersionV1beta1:
+		logrus.Infof("Upgrading from %v to %v", types.CRDAPIVersionV1beta1, types.CurrentCRDAPIVersion)
+		if err := v1beta1.UpgradeFromV1beta1ToV1beta2(config, namespace, lhClient); err != nil {
```
jenting, line 193: I'm thinking that this step should be put in like v123to130 package.

So, `doAPIVersionUpgrade` only upgrades the Setting `crd-api-version`.
And the corresponding conditions conversion from the map to slice should be into v123to130.
Also, we should copy existing v1beta1 CR into v1beta2 CR in v123to130 as well.

Therefore, we don't have to change the upgrade packages v070to080, v100to101, v102to110, v110to111, v110to120, v111to120, v120to121, v122to123.

In general, the package v123to130 should involve:
- Copy the existing v1beta1 CR to v1beta2 CR
- Convert the map to slice of the `status.conditions`

What do you and others think? @derekbit @innobead @shuo-wu @joshimoo

upgrade/v122to123/upgrade.go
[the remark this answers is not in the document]
derekbit: @jenting Yes. So upgrade/vxxxtoxxx/upgrade.go should use v1beta2 APIs to avoid the problem?
`upgrade/vxxxtoxxx/upgrade.go` is for v1beta1 upgrade, using v1beta2 APIs  is a bit weird.

upgrade/upgrade.go
derekbit: I agree with it. It will be unified.

derekbit: But in upgrade functions, they uses the datastore and engine, which accepts v1beta2 objects now, but some objects are got by v1beta1 APIs. They are incompatible. Any idea to resolve?

jenting: > But in upgrade functions, they uses the datastore and engine, which accepts v1beta2 objects now, but some objects are got by v1beta1 APIs. They are incompatible. Any idea to resolve?

No from my side. Let's keep your current way that directly upgrades to v1beta2. Otherwise, have an interface to converts the datastore <-> v1beta1/v1beta2 and engine <-> v1beta1/v1beta2 is too complex.

derekbit: > So, doAPIVersionUpgrade only upgrades the Setting crd-api-version.
And the corresponding conditions conversion from the map to slice should be in v123to130.
Also, we should copy existing v1beta1 CR into v1beta2 CR in v123to130 as well.

We need to upgrade v1beta1 CR to v1beta2  first because status.Conditions is changed, or we will encounter
```
time="2021-11-24T07:23:22Z" level=fatal msg="Error starting manager: upgrade CRD failed: upgrade from v0.7.0 to v0.8.0: upgrade instance manager failed: upgrade from v0.7.0 to v0.8.0: failed to list all nodes during the instance managers upgrade: v1beta2.NodeList.Items: []v1beta2.Node: v1beta2.Node.Status: v1beta2.NodeStatus.Conditions: []v1beta2.Condition: decode slice: expect [ or n, but found {, error found in #10 byte of ...|ditions\":{},\"diskSta|..., bigger context ...|rCPURequest\":0,\"tags\":[]},\"status\":{\"conditions\":{},\"diskStatus\":{\"default-disk-745ff2be8d312b52\":{\"|..."
```

derekbit:
Leaving a not here.

For future upgrade refactoring, need to separate the two scenarios

(1) CRD version upgrade
(2) CRD version not upgraded, but longhorn version upgrade

The logic of current upgrade mixes them up, so it is still messy now.

deploy/install/01-prerequisite/03-crd.yaml
shuo-wu: Is file `deploy/install/01-prerequisite/03-crd.yaml` in the 2nd commit automatically generated by the script `k8s/generate_code.sh`? I saw it included both v1beta1 and v1beta2. If YES, that would be great.

upgrade/v122to123/upgrade.go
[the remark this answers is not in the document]
shuo-wu: YES. We should use v1beta2 clientset in all `upgrade/vxxxtoxxx/upgrade.go` and you already did that.

Here we are confused why there is a type conversion `longhorn.BackupState()`.

upgrade/upgrade.go
shuo-wu: Based on the info above, I think upgrading the API version then applying vxxxtoxxx upgrade would be more straightforward in terms of implementation.

upgrade/v122to123/upgrade.go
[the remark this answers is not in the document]
derekbit: @shuo-wu  we don't need  type conversion after replacing the v1beta1 apis with v1beta2.

deploy/install/01-prerequisite/03-crd.yaml
derekbit: @shuo-wu Yes, now the all lines in `03-crd.yaml` are generated by `k8s/generate_code.sh`.

upgrade/upgrade.go
derekbit: @shuo-wu Yes. But because of the CR status change, so we need to upgrade CR from v1beta1 to v1beta2 first, or we will hit the data type mismatch error.
I think this can be addressed in upgrade path refactoring task like https://github.com/longhorn/longhorn-manager/pull/1136#issuecomment-977630063

upgrade/upgrade.go
```diff
@@ -148,61 +148,53 @@ func doAPIVersionUpgrade(namespace string, config *restclient.Config, lhClient *
...
 	case "":
-		// it can be a new installation or v1alpha1
-		isV1alpha1, err := v1alpha1.IsCRDVersionMatch(config, namespace)
-		if err != nil {
-			logrus.Warnf("Cannot verify current CRD version, assume it's not v1alpha1: %v", err)
+		// upgradable: supported version or new installation ((v1beta1, v1beta2)
+		// non-upgradable: non-supported version (v1alpha1 which cannot upgrade directly)
+		upgradable, err := v1beta1.CanUpgrade(config, namespace)
+		if !upgradable {
+			return err
```
shuo-wu, line 177: This looks weird... Should it be something like:
```
upgradable, err := v1beta1.CanUpgrade(config, namespace)
if err != nil {
	return err
}
if upgradable {
	......  // do upgrade
}

// create or update the setting
......
```

upgrade/v1beta1/upgrade.go
```diff
@@ -0,0 +1,214 @@
...
+		}
+		new.Status.Conditions = conditions
+
+		new, err = lhClient.LonghornV1beta2().Volumes(namespace).UpdateStatus(context.TODO(), new, metav1.UpdateOptions{})
+		if err != nil {
+			if !apierrors.IsConflict(err) {
+				return errors.Wrapf(err, "failed to update for %v status %v", old.Kind, old.Name)
+			}
+			logrus.Warnf("%v: update status for %v %v v1beta1 result in conflict, skipping", upgradeLogPrefix, old.Kind, old.Name)
+			continue
```
shuo-wu, line 120: I knew these codes are derived from `upgrade/v1alpha1/upgrade.go`. But I don't think we can skip here if there is a conflict. Since later on we will update the API version setting to v1beta2 then there is no chance to redo this upgrade anymore.

shuo-wu (changes requested):

shuo-wu:
> Leaving a not here.
> 
> For future upgrade refactoring, need to separate the two scenarios
> 
> (1) CRD version upgrade (2) CRD version not upgraded, but longhorn version upgrade
> 
> The logic of current upgrade mixes them up, so it is still messy now.

We always do API version upgrade then longhorn-version-based CR upgrade.
https://github.com/longhorn/longhorn-manager/blob/050cbf3d0a97c1c8d8681d611c9382bcb4b76daf/upgrade/upgrade.go#L113-L127

Maybe we can squash functions `doPodsUpgrade`, `doServicesUpgrade `, and `doDeploymentAndDaemonSetUpgrade` into `doCRUpgrade`. But I think the current flow in general is clear. Not sure which part could confuse you when you implement this PR.

upgrade/v122to123/upgrade.go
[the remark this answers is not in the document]
shuo-wu: Right. And I saw this modification was removed.

upgrade/upgrade.go
derekbit: yes, will modify it.

upgrade/v1beta1/upgrade.go
derekbit: > > Leaving a not here.
> > For future upgrade refactoring, need to separate the two scenarios
> > (1) CRD version upgrade (2) CRD version not upgraded, but longhorn version upgrade
> > The logic of current upgrade mixes them up, so it is still messy now.
> 
> We always do API version upgrade then longhorn-version-based CR upgrade.
> 
> https://github.com/longhorn/longhorn-manager/blob/050cbf3d0a97c1c8d8681d611c9382bcb4b76daf/upgrade/upgrade.go#L113-L127
> 
> Maybe we can squash functions `doPodsUpgrade`, `doServicesUpgrade `, and `doDeploymentAndDaemonSetUpgrade` into `doCRUpgrade`. But I think the current flow in general is clear. Not sure which part could confuse you when you implement this PR.

There are upgrade CR in doAPIVersionUpgrade and doCRUpgrade, so it will be confusing. 
Yes, squashing functions is a good way to avoid confusion.

derekbit: It tries to UpgradeCRFromV1beta1ToV1beta2 first. The crdAPIVersionSetting is not updated if the first failed, so it will retry again next time.

```
	case types.CRDAPIVersionV1beta1:
		logrus.Infof("Upgrading from %v to %v", types.CRDAPIVersionV1beta1, types.CurrentCRDAPIVersion)
		if err := v1beta1.UpgradeCRFromV1beta1ToV1beta2(config, namespace, lhClient); err != nil {
			return err
		}
		crdAPIVersionSetting.Value = types.CRDAPIVersionV1beta2
		if _, err := lhClient.LonghornV1beta2().Settings(namespace).Update(context.TODO(), crdAPIVersionSetting, metav1.UpdateOptions{}); err != nil {
			return errors.Wrapf(err, "cannot finish CRD API upgrade by setting the CRDAPIVersionSetting to %v", types.CurrentCRDAPIVersion)
		}
```

derekbit:
> > Leaving a not here.
> > For future upgrade refactoring, need to separate the two scenarios
> > (1) CRD version upgrade (2) CRD version not upgraded, but longhorn version upgrade
> > The logic of current upgrade mixes them up, so it is still messy now.
> 
> We always do API version upgrade then longhorn-version-based CR upgrade.
> 
> https://github.com/longhorn/longhorn-manager/blob/050cbf3d0a97c1c8d8681d611c9382bcb4b76daf/upgrade/upgrade.go#L113-L127
> 
> Maybe we can squash functions `doPodsUpgrade`, `doServicesUpgrade `, and `doDeploymentAndDaemonSetUpgrade` into `doCRUpgrade`. But I think the current flow in general is clear. Not sure which part could confuse you when you implement this PR.

There are upgrade CR in doAPIVersionUpgrade and doCRUpgrade, so it will be confusing.
Yes, squashing functions is a good way to avoid confusion.

derekbit:
Hello @jenting and @shuo-wu
The PR is updated based on the latest apis and comments. Could you please help review? Thank you.

upgrade/upgrade.go
```diff
@@ -148,61 +148,55 @@ func doAPIVersionUpgrade(namespace string, config *restclient.Config, lhClient *
...
 		return nil
 	}
 
 	switch crdAPIVersion {
 	case "":
-		// it can be a new installation or v1alpha1
-		isV1alpha1, err := v1alpha1.IsCRDVersionMatch(config, namespace)
+		// upgradable: supported version or new installation ((v1beta1, v1beta2)
+		// non-upgradable: non-supported version (v1alpha1 which cannot upgrade directly)
+		upgradable, err := v1beta1.CanUpgrade(config, namespace)
```
shuo-wu, line 175: The comment does not match the behavior of function `v1beta1.CanUpgrade(config, namespace)`: This function will return true **only when** the setting is not found.

upgrade/v1beta1/upgrade.go
shuo-wu: Sorry, this kind of volume update conflict won't fail `UpgradeCRFromV1beta1ToV1beta2`... The setting will be updated anyway...

shuo-wu:
> There are upgrade CR in doAPIVersionUpgrade and doCRUpgrade, so it will be confusing. Yes, squashing functions is a good way to avoid confusion.

The essential of both API version upgrade and Longhorn system version upgrade is CR upgrade. I think the current upgrade flow is clear enough... Do you mean you prefer the following kind of upgrade flow?
```
upgradeForVolumeCR()
    |- upgradeAPIVersion
    |- upgradeFromvxxxTovxxx
    |- upgradeFromvxxxTovxxx
upgradeForEngineCR()
    |- upgradeAPIVersion
    |- upgradeFromvxxxTovxxx
    |- upgradeFromvxxxTovxxx
```

shuo-wu:
After the discussion with Derek, we made a consensus about the expected upgrade flow, which should be like the following:
```
upgradeAPIversion
    |-doCRupgrade (not necessary every upgrade)
upgradeLHversion
    |-vxxxTovxxxUpgrade (not necessary every upgrade)
        |-doLHCRUpgrade
        |-doK8SCRUpgrade
    |-vxxxTovxxxUpgrade (not necessary every upgrade)
        |-doLHCRUpgrade
        |-doK8SCRUpgrade
    …...
```

upgrade/upgrade.go
derekbit: Yes, update the comment.

shuo-wu: This function implicitly means that if the setting CR cannot be retrieved by the v1beta2 client but is parsable for the v1beta1 client (not an empty value), the version must be v1alpha1, right? It's still a little bit confusing. I am not sure if we can improve this part.

upgrade/v1beta1/upgrade.go
```diff
@@ -0,0 +1,210 @@
...
+		new.Status.Conditions = conditions
+
+		new, err = lhClient.LonghornV1beta2().EngineImages(namespace).UpdateStatus(context.TODO(), new, metav1.UpdateOptions{})
+		if err != nil {
+			if !apierrors.IsConflict(err) {
+				return errors.Wrapf(err, "failed to update for %v status %v", old.Kind, old.Name)
+			}
+			logrus.Warnf("%v: update status for %v %v v1beta2 result in conflict, skipping", upgradeLogPrefix, old.Kind, old.Name)
+			continue
+		}
```
shuo-wu, line 147: The conflicts caused by other kinds of CRs should be returned as well. Similar below.

shuo-wu (changes requested):

upgrade/upgrade.go
```diff
@@ -148,61 +148,55 @@ func doAPIVersionUpgrade(namespace string, config *restclient.Config, lhClient *
...
 			}
-			logrus.Infof("Initialized CRD API Version to %v", types.CurrentCRDAPIVersion)
-			break
-		}
-		fallthrough
-	case types.CRDAPIVersionV1alpha1:
-		if types.CurrentCRDAPIVersion != types.CRDAPIVersionV1beta1 {
-			return fmt.Errorf("cannot upgrade from %v to %v directly", types.CRDAPIVersionV1alpha1, types.CurrentCRDAPIVersion)
+			_, err = lhClient.LonghornV1beta2().Settings(namespace).Create(context.TODO(), crdAPIVersionSetting, metav1.CreateOptions{})
+			if err != nil {
```
jenting, line 188: Consider adding a check for `!apierrors.IsAlreadyExists(err)`.

This could prevent a corner case that the CR `crd-api-version` is there but with an empty string, then in longhorn-manager upgrade path, the longhorn-manager would try to create the `crd-api-version` Setting CR and return an error since the CR already exists.

upgrade/v1beta1/upgrade.go
```diff
@@ -0,0 +1,210 @@
...
+
+		conditions, err := copyConditions(old.Status.Conditions)
+		if err != nil {
+			return errors.Wrap(err, "upgrade: copy volume conditions")
+		}
+		new.Status.Conditions = conditions
+
+		new, err = lhClient.LonghornV1beta2().Volumes(namespace).UpdateStatus(context.TODO(), new, metav1.UpdateOptions{})
+		if err != nil {
```
jenting, line 115: Should this part check the conflicts as well?

upgrade/v1beta1/upgrade.go
```diff
@@ -0,0 +1,210 @@
...
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			logrus.Infof("setting %v not found", string(types.SettingNameDefaultEngineImage))
+			return true, nil
+		}
+
+		return false, errors.Wrap(err, fmt.Sprintf("unable to get setting %v", string(types.SettingNameDefaultEngineImage)))
+	}
+
+	return false, errors.Wrapf(err, "unable to upgrade from %v directly", setting.Value)
```
jenting, line 93: Leaving a note here. Since like at v1alpha1 -> v1beta1, the longhorn-manager uses the `default-engine-image` Setting CR to determine whether it it v1alpha1 _or_ v1beta1, and the code is copy from there.

However, the code now is to handle upgrades from v1beta1 -> v1beta2, not sure should we keep this code logic here _or_ we could directly drop it.

Probably we could leave it as is, and see if there is any issue happens related to this one in the future.

upgrade/v1beta1/upgrade.go
derekbit: Always return error if there is any error and conflict and retry next time.

derekbit:
@jenting @shuo-wu @innobead 
Please help to review, thank you.

api/model.go
```diff
@@ -266,9 +266,20 @@ type Node struct {
...
+	StorageMaximum   int64                         `json:"storageMaximum"`
+	ScheduledReplica map[string]int64              `json:"scheduledReplica"`
+	DiskUUID         string                        `json:"diskUUID"`
+}
+
 type DiskInfo struct {
 	longhorn.DiskSpec
-	longhorn.DiskStatus
+	// To convert status.Condtions' slice to map type, replace longhorn.Distatus with DiskStatus,
+	// the UI does not any change.
```
jenting, line 281:
```suggestion
        // To convert CR status.Conditions from datatype map to slice, replace longhorn.DiskStatus with DiskStatus to keep it use datatype map.
        // Therefore, the UI (RESTful endpoint) does not need to do any changes.
```

api/model.go
```diff
@@ -1221,7 +1235,7 @@ func toBackupTargetResource(bt *longhorn.BackupTarget) *BackupTarget {
 			CredentialSecret: bt.Spec.CredentialSecret,
 			PollInterval:     bt.Spec.PollInterval.Duration.String(),
 			Available:        bt.Status.Available,
-			Message:          bt.Status.Conditions[longhorn.BackupTargetConditionTypeUnavailable].Message,
+			Message:          condition.Message,
```
jenting, line 1238: nit
```
Message: types.GetCondition(bt.Status.Conditions, longhorn.BackupTargetConditionTypeUnavailable).Message,
```

upgrade/v1beta1/upgrade.go
```diff
@@ -0,0 +1,287 @@
...
+
+		if err := copier.Copy(&new, &old); err != nil {
+			return err
+		}
+
+		recurringJobs, err := copyRecurringJobs(old.Spec.RecurringJobs)
+		if err != nil {
+			return errors.Wrap(err, "failed to copy volume Spec.RecurringJobs")
+		}
+		new.Spec.RecurringJobs = recurringJobs
```
jenting, line 95: Not sure why you need to copy the spec.RecurringJobs. And at below, you call `UpdateStatus`, it would not update the spec.RecurringJobs in the end. [0]

upgrade/v1beta1/upgrade.go
```diff
@@ -0,0 +1,287 @@
...
+			return errors.Wrap(err, "failed to copy volume Status.KubernetesStatus")
+		}
+
+		if err := copier.Copy(&new.Status.CloneStatus, &old.Status.CloneStatus); err != nil {
+			return errors.Wrap(err, "failed to copy volume Status.KubernetesStatus")
+		}
+
+		if err := copier.Copy(&new.Status.KubernetesStatus, &old.Status.KubernetesStatus); err != nil {
+			return err
+		}
```
jenting, line 113: Why do need to copy these fields? [1]

upgrade/v1beta1/upgrade.go
```diff
@@ -0,0 +1,287 @@
...
+
+		conditions, err := copyConditions(old.Status.Conditions)
+		if err != nil {
+			return errors.Wrap(err, "failed to copy engineImage Status.Conditions")
+		}
+		new.Status.Conditions = conditions
+
+		if err := copier.Copy(&new.Status.EngineVersionDetails, &old.Status.EngineVersionDetails); err != nil {
+			return errors.Wrap(err, "failed to copy engineImage Status.EngineVersionDetails")
+		}
```
jenting, line 149: [1]

upgrade/v1beta1/upgrade.go
```diff
@@ -0,0 +1,287 @@
...
+		new := longhorn.Node{}
+
+		if err := copier.Copy(&new, &old); err != nil {
+			return err
+		}
+
+		new.Spec.Disks, err = copyDisks(old.Spec.Disks)
+		if err != nil {
+			return errors.Wrap(err, "failed to copy node Spec.Disks")
+		}
```
jenting, line 212: [0]

upgrade/v1beta1/upgrade.go
```diff
@@ -0,0 +1,287 @@
...
+
+		new.Status.Conditions, err = copyConditions(old.Status.Conditions)
+		if err != nil {
+			return errors.Wrap(err, "failed to copy node Status.Conditions")
+		}
+
+		new.Status.DiskStatus, err = copyDiskStatus(old.Status.DiskStatus)
+		if err != nil {
+			return errors.Wrap(err, "failed to copy node Status.DiskStatus")
+		}
```
jenting, line 222: [1]

upgrade/v1beta1/upgrade.go
derekbit: Don't need this. It's for the previous error handling implementation. Forgot to clean up it.

upgrade/v1beta1/upgrade.go
derekbit: Because the data type is different, copier.Copy() cannot copy them.

jenting: But the KubernetesStatus, CloneStatus does not contain the Conditions fields inside.
Also, the status.KubernetesStatus copied twice, intended?

derekbit: > Also, the status.KubernetesStatus copied twice, intended?

Yes, already fixed

jenting (approved):
LGTM

upgrade/v1beta1/upgrade.go
```diff
@@ -0,0 +1,248 @@
...
+		return err
+	}
+	if err := upgradeBackupTargets(namespace, lhClient); err != nil {
+		return err
+	}
+	if err := upgradeNodes(namespace, lhClient); err != nil {
+		return err
+	}
+
+	logrus.Infof("%v: completed", upgradeLogPrefix)
```
jenting, line 46: nit
```suggestion
        logrus.Infof("%v completed", upgradeLogPrefix)
```

derekbit: Fixed.

jenting (dismissed):

jenting (dismissed):

jenting:
Remember to file a PR to update the crd.yaml to https://github.com/longhorn/longhorn/blob/v1.3.0/chart/templates/crds.yaml.

upgrade/v1beta1/upgrade.go
shuo-wu: As I mentioned in this [comment](https://github.com/longhorn/longhorn-manager/pull/1136#discussion_r758306872), this code may confuse the contributors who are not familiar with the API version upgrade logic & compatibility. Since this function directly returns false without checking the setting. It's better to add some comments or explicitly check `setting` instead.

shuo-wu (commented):
In general LGTM

jenting (approved):

shuo-wu (approved):

upgrade/v110to120/upgrade.go
```diff
@@ -18,7 +18,7 @@ import (
 	"sigs.k8s.io/yaml"
 
 	"github.com/longhorn/longhorn-manager/datastore"
-	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta1"
+	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
```
innobead, line 21: Why in the old upgrade path, we change to use v1beta2 instead of v1beta1 instead? I assume the new API should be used in the upgrade to 1.3.0?

upgrade/v1beta1/upgrade.go
```diff
@@ -0,0 +1,234 @@
...
+			return err
+		}
+
+		conditions, err := copyConditions(old.Status.Conditions)
+		if err != nil {
+			return errors.Wrap(err, "failed to copy engineImage Status.Conditions")
+		}
+		new.Status.Conditions = conditions
+
+		if err := copier.Copy(&new.Status.EngineVersionDetails, &old.Status.EngineVersionDetails); err != nil {
```
innobead, line 131: Does copier not support nested copy?

k8s/pkg/apis/longhorn/v1beta2/backingimage.go
```diff
@@ -0,0 +1,92 @@
...
+	// +optional
+	DiskDownloadProgressMap map[string]int `json:"diskDownloadProgressMap"`
+}
+
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:resource:shortName=lhbi
+// +kubebuilder:subresource:status
+// +kubebuilder:storageversion
+// +kubebuilder:printcolumn:name="Image",type=string,JSONPath=`.spec.image`,description="The backing image name"
```
innobead, line 73: `.spec.image` is this correct?

innobead (changes requested):
In general, LGTM, but just few comments to resolve first.

k8s/pkg/apis/longhorn/v1beta2/backingimage.go
derekbit: backingimage's ` .spec.image` also does not exist in v1.1.x and v1.2.x.
Not sure it's a typo or just not implemented.

innobead: cc @shuo-wu

upgrade/v110to120/upgrade.go
innobead: Got it, thanks.

Just curious why we don't upgrade CRs from v1beta1 to v1beta2 only just before upgrading to 1.3.0 instead of doing that at the first step before all upgrade paths?

derekbit: Discussed with @shuo-wu. It's an ordering logic consideration.
We upgrade API to v1beta2 first, and then only use the v1beta2 APIs later on.

If upgrade the CRs first, the logic will become
`upgrade longhorn version < 1.3 CRs by v1beta1` -> `upgrade API to v1beta2` -> `upgrade longhorn CRs  >= 1.3 CRs by v1beta2`.

The upgrade flow becomes complicated and duplicated.

innobead (approved):

upgrade/v110to120/upgrade.go
innobead: Thanks for the explanation

PR #1161, opened 2021-11-29
[Backport][v1.2.x]csi: Add a flag for the resizer so that it would handle completed job pods

shuo-wu:
https://github.com/longhorn/longhorn/issues/3303
https://github.com/longhorn/longhorn/issues/3339


innobead (approved):
LGTM

PR #1162, opened 2021-11-29
[Backport][v1.1.3]csi: Add a flag for the resizer so that it would handle completed job…

shuo-wu:
longhorn/longhorn#3303
longhorn/longhorn#3340


innobead:
@shuo-wu It seems we don't need `handle-volume-inuse-error` in 1.1.3 because 1.1.3 still uses 0.5.1 resizer which does not have this option/flag.

innobead:
Closing first, but feel free to reopen if we still need this.

PR #1166, opened 2021-12-02
controller: ReplicaController does not enqueue irrelavent replicas for backing image changes

shuo-wu:
Longhorn/longhorn#3242

Signed-off-by: Shuo Wu <shuo.wu@suse.com>

### DIFF
--- a/controller/replica_controller.go
+++ b/controller/replica_controller.go
@@ -762,14 +762,16 @@ func (rc *ReplicaController) enqueueBackingImageChange(obj interface{}) {
 		}
 	}
 
+	replicas, err := rc.ds.ListReplicasByNodeRO(rc.controllerID)
+	if err != nil {
+		utilruntime.HandleError(fmt.Errorf("failed to list replicas on node %v for backing image %v: %v", rc.controllerID, backingImage.Name, err))
+		return
+	}
 	for diskUUID := range backingImage.Status.DiskFileStatusMap {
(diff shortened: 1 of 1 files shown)


innobead (approved):

innobead:
@shuo-wu Please backport this to 1.2.x.

controller/replica_controller.go
```diff
@@ -762,14 +762,16 @@ func (rc *ReplicaController) enqueueBackingImageChange(obj interface{}) {
 		}
 	}
 
+	replicas, err := rc.ds.ListReplicasByNodeRO(rc.controllerID)
+	if err != nil {
+		utilruntime.HandleError(fmt.Errorf("failed to list replicas on node %v for backing image %v: %v", rc.controllerID, backingImage.Name, err))
+		return
+	}
```
PhanLe1010, line 769: when node-1 goes down, all replicas on node-1 will be handled by a new node, node-2. 

Now the replica controller node-2 only lists replicas on node-2 as mentioned in line 765. Does it mean that the replica controller on node-2 always skip handling the newly migrated replica from node-1?

shuo-wu: The reason why the replica controller needs to enqueue the backing image changes is that it needs to start the replica processes once the backing image on the node is ready. Hence, we don't need to do anything for the replicas on the down nodes.

PhanLe1010: I see. Thank you for the explantion

PR #1170, opened 2021-12-06
Handle the CRD validation error

derekbit:
- The null fields of CRs Spec are initialized correctly during creation.
- The old null fields of the CRs Spec are fixed up.
- The slices and maps in the CRs Status are set `// +nullable` because some of them are not set while calling Update.

Longhorn 3352

### DIFF
--- a/datastore/longhorn.go
+++ b/datastore/longhorn.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/jinzhu/copier"
 	"github.com/pkg/errors"
 	"github.com/robfig/cron"
 	"github.com/sirupsen/logrus"
@@ -375,6 +376,9 @@ func GetOwnerReferencesForRecurringJob(recurringJob *longhorn.RecurringJob) []me
 
 // CreateVolume creates a Longhorn Volume resource and verifies creation
 func (s *DataStore) CreateVolume(v *longhorn.Volume) (*longhorn.Volume, error) {
+	if err := initVolume(v); err != nil {
+		return nil, err
+	}
 	if err := checkVolume(v); err != nil {
 		return nil, err
 	}
@@ -415,6 +419,32 @@ func (s *DataStore) CreateVolume(v *longhorn.Volume) (*longhorn.Volume, error) {
 	}
 
 	return ret.DeepCopy(), nil
+}
+
+func initVolume(v *longhorn.Volume) error {
+	if v.Spec.DiskSelector == nil {
+		v.Spec.DiskSelector = []string{}
+	}
+	if v.Spec.NodeSelector == nil {
+		v.Spec.NodeSelector = []string{}
+	}
+	if v.Spec.RecurringJobs == nil {
+		v.Spec.RecurringJobs = make([]longhorn.VolumeRecurringJobSpec, 0)
+	}
+	for i, src := range v.Spec.RecurringJobs {
+		dst := longhorn.VolumeRecurringJobSpec{}
+		if err := copier.Copy(&dst, &src); err != nil {
+			return err
+		}
+		if dst.Groups == nil {
+			dst.Groups = []string{}
+		}
+		if dst.Labels == nil {
+			dst.Labels = make(map[string]string, 0)
+		}
+		v.Spec.RecurringJobs[i] = dst
+	}
+	return nil
 }
 
 // UpdateVolume updates Longhorn Volume and verifies update
@@ -1185,6 +1215,9 @@ func (s *DataStore) CheckEngineImageReadyOnAllVolumeReplicas(image, volumeName, 
 // CreateBackingImage creates a Longhorn BackingImage resource and verifies
 // creation
 func (s *DataStore) CreateBackingImage(backingImage *longhorn.BackingImage) (*longhorn.BackingImage, error) {
+	if err := initBackingImage(backingImage); err != nil {
+		return nil, err
+	}
 	if err := util.AddFinalizer(longhornFinalizerKey, backingImage); err != nil {
 		return nil, err
 	}
@@ -1208,6 +1241,16 @@ func (s *DataStore) CreateBackingImage(backingImage *longhorn.BackingImage) (*lo
 	}
 
 	return ret.DeepCopy(), nil
+}
+
+func initBackingImage(backingImage *longhorn.BackingImage) error {
+	if backingImage.Spec.Disks == nil {
+		backingImage.Spec.Disks = make(map[string]string, 0)
+	}
+	if backingImage.Spec.SourceParameters == nil {
+		backingImage.Spec.SourceParameters = make(map[string]string, 0)
+	}
+	return nil
 }
 
 // UpdateBackingImage updates Longhorn BackingImage and verifies update
@@ -1317,6 +1360,9 @@ func GetOwnerReferencesForBackingImage(backingImage *longhorn.BackingImage) []me
 // CreateBackingImageManager creates a Longhorn BackingImageManager resource and verifies
 // creation
 func (s *DataStore) CreateBackingImageManager(backingImageManager *longhorn.BackingImageManager) (*longhorn.BackingImageManager, error) {
+	if err := initBackingImageManager(backingImageManager); err != nil {
+		return nil, err
+	}
 	if err := util.AddFinalizer(longhornFinalizerKey, backingImageManager); err != nil {
 		return nil, err
 	}
@@ -1347,6 +1393,13 @@ func (s *DataStore) CreateBackingImageManager(backingImageManager *longhorn.Back
 	}
 
 	return ret.DeepCopy(), nil
+}
+
+func initBackingImageManager(backingImageManager *longhorn.BackingImageManager) error {
+	if backingImageManager.Spec.BackingImages == nil {
+		backingImageManager.Spec.BackingImages = make(map[string]string, 0)
+	}
+	return nil
 }
 
 // UpdateBackingImageManager updates Longhorn BackingImageManager and verifies update
@@ -1514,6 +1567,9 @@ func GetOwnerReferencesForBackingImageManager(backingImageManager *longhorn.Back
 // CreateBackingImageDataSource creates a Longhorn BackingImageDataSource resource and verifies
 // creation
 func (s *DataStore) CreateBackingImageDataSource(backingImageDataSource *longhorn.BackingImageDataSource) (*longhorn.BackingImageDataSource, error) {
+	if err := initBackingImageDataSource(backingImageDataSource); err != nil {
+		return nil, err
+	}
 	if err := util.AddFinalizer(longhornFinalizerKey, backingImageDataSource); err != nil {
 		return nil, err
 	}
@@ -1537,6 +1593,13 @@ func (s *DataStore) CreateBackingImageDataSource(backingImageDataSource *longhor
 	}
 
 	return ret.DeepCopy(), nil
+}
+
+func initBackingImageDataSource(backingImageDataSource *longhorn.BackingImageDataSource) error {
+	if backingImageDataSource.Spec.Parameters == nil {
+		backingImageDataSource.Spec.Parameters = make(map[string]string, 0)
+	}
+	return nil
 }
 
 // UpdateBackingImageDataSource updates Longhorn BackingImageDataSource and verifies update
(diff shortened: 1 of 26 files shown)


k8s/pkg/apis/longhorn/v1beta1/volume.go
```diff
@@ -197,9 +201,11 @@ type VolumeSpec struct {
 	ReplicaAutoBalance ReplicaAutoBalance `json:"replicaAutoBalance"`
 	// Deprecated. Rename to BackingImage
 	// +optional
+	// +nullable
```
jenting, line 204: This one should not necessary.

shuo-wu (commented):
One quick question: Will these fields be initialized during the CR creation if these comments are not introduced?

k8s/pkg/apis/longhorn/v1beta1/volume.go
derekbit: Yes, fixed.

derekbit:
> 



> One quick question: Will these fields be initialized during the CR creation if these comments are not introduced?

Most of the fields are initialized during the CR creation.
`nodes.status.diskStatus[].tags` is not initialized and stays `null`.

jenting (approved):
LGTM, thank you @derekbit.

k8s/pkg/apis/longhorn/v1beta1/engine.go
```diff
@@ -140,16 +140,22 @@ type EngineStatus struct {
...
+	// +nullable
 	PurgeStatus map[string]*PurgeStatus `json:"purgeStatus"`
 	// +optional
+	// +nullable
 	RebuildStatus map[string]*RebuildStatus `json:"rebuildStatus"`
 	// +optional
+	// +nullable
 	CloneStatus map[string]*SnapshotCloneStatus `json:"cloneStatus"`
 	// +optional
+	// +nullable
```
shuo-wu, line 158: Remove this.

k8s/pkg/apis/longhorn/v1beta1/node.go
```diff
@@ -41,6 +41,7 @@ type DiskSpec struct {
 	// +optional
 	StorageReserved int64 `json:"storageReserved"`
 	// +optional
+	// +nullable
```
shuo-wu, line 44: Remove this.

k8s/pkg/apis/longhorn/v1beta1/node.go
```diff
@@ -70,6 +71,7 @@ type NodeSpec struct {
 	// +optional
 	EvictionRequested bool `json:"evictionRequested"`
 	// +optional
+	// +nullable
```
shuo-wu, line 74: Remove this.

k8s/pkg/apis/longhorn/v1beta1/node.go
```diff
@@ -84,6 +86,7 @@ type NodeStatus struct {
 	// +optional
 	Conditions map[string]Condition `json:"conditions"`
 	// +optional
+	// +nullable
```
shuo-wu, line 89: Remove this.

k8s/pkg/apis/longhorn/v1beta2/backingimage.go
```diff
@@ -54,6 +54,7 @@ type BackingImageStatus struct {
 	// +optional
 	Checksum string `json:"checksum"`
 	// +optional
+	// +nullable
```
shuo-wu, line 57: Remove this.

shuo-wu (changes requested):
The reason why I prefer not to add `nullble` comment for some fields is that: the fields will be initialized as long as Longhorn finds that they are null. In other words, if a field is nullble, it means Longhorn will do something special rather than blindly initializing it when it is null.

derekbit:
> The reason why I prefer not to add `nullble` comment for some fields is that: the fields will be initialized as long as Longhorn finds that they are null. In other words, if a field is nullble, it means Longhorn will do something special rather than blindly initializing it when it is null.

Got it.

Is removing the `// +nullable` of some fields to ensure that the error can be reflected immediately if Longhorn does not handle correctly?

shuo-wu:
> Is removing the `// +nullable` of some fields to ensure that the error can be reflected immediately if Longhorn does not handle correctly?

I am not sure what Kubernetes will do exactly if these fields become null. But by design these fields should not be null.

k8s/pkg/apis/longhorn/v1beta1/node.go
derekbit: The Tags is NULL after creating the CR.
If remove the `//+ nullable`, the UpdateStatus will fail.

```
failed to update=Node.longhorn.io \“ku50-master\” is invalid: [spec.disks.default-disk-745ff2be8d312b52.tags: Invalid value: \“null\“: spec.disks.default-disk-745ff2be8d312b52.tags in body must be of type array: \“null\“, status.diskStatus: Invalid value: \“null\“: status.diskStatus in body must be of type object: \“null\“]”
```

shuo-wu: This is caused by `diskSpec.Tags` rather than `nodeSpec.Tags`. Can you add the initialization when adding a new disk to nodeSpec?

derekbit: Oh! Right.  
> Can you add the initialization when adding a new disk to nodeSpec?
No prob. Trying to add it.

jenting:
Ref to https://github.com/longhorn/longhorn/issues/791

k8s/pkg/apis/longhorn/v1beta1/backuptarget.go
```diff
@@ -18,9 +18,11 @@ type BackupTargetSpec struct {
 	CredentialSecret string `json:"credentialSecret"`
 	// The interval that the cluster needs to run sync with the backup target.
 	// +optional
+	// +nullable
```
jenting, line 21: This one seems not to be necessary?

derekbit:
Hello @shuo-wu and @jenting,

I separate the handling of Spec and Status in the last PR.

- The null fields of CRs Spec are initialized correctly during creation.
- The old null fields of the CRs Spec are fixed up.
- The slices and maps in the CRs Status are set `// +nullable` because some of them are not set and null while calling Update.

Please help to review, thank you.

datastore/longhorn.go
```diff
@@ -417,6 +421,30 @@ func (s *DataStore) CreateVolume(v *longhorn.Volume) (*longhorn.Volume, error) {
...
+	}
+	if v.Spec.NodeSelector == nil {
+		v.Spec.NodeSelector = []string{}
+	}
+	if v.Spec.RecurringJobs == nil {
+		v.Spec.RecurringJobs = make([]longhorn.VolumeRecurringJobSpec, 0)
+	}
+	for i, src := range v.Spec.RecurringJobs {
+		dst := longhorn.VolumeRecurringJobSpec{}
+		copier.Copy(&dst, &src)
```
jenting, line 436: Should we check the error?

upgrade/v1beta1/upgrade.go
```diff
@@ -120,6 +139,43 @@ func upgradeVolumes(namespace string, lhClient *lhclientset.Clientset) error {
...
+			if dst.Groups == nil {
+				dst.Groups = []string{}
+			}
+			if dst.Labels == nil {
+				dst.Labels = make(map[string]string, 0)
+			}
+			obj.Spec.RecurringJobs[i] = dst
+		}
+
+		if _, err = lhClient.LonghornV1beta1().Volumes(namespace).Update(context.TODO(), &obj, metav1.UpdateOptions{}); err != nil {
```
jenting, line 172: Should we call `LonghornV1beta2` update?

k8s/pkg/apis/longhorn/v1beta1/backuptarget.go
jenting: I think this one does not need to add `// +nullable`.

datastore/longhorn.go
derekbit: Yes. Should check it.

upgrade/v1beta1/upgrade.go
derekbit: Hmm... The Status.Conditions are not converted yet, and the Update operation also validates the Status.
Thus, using v1beta1.Update is the only way to update Spec.

k8s/pkg/apis/longhorn/v1beta1/backuptarget.go
derekbit: Yes, should remove it.

upgrade/v1beta1/upgrade.go
jenting: I see.

upgrade/v1beta1/upgrade.go
```diff
@@ -44,6 +44,21 @@ func UpgradeCRFromV1beta1ToV1beta2(config *restclient.Config, namespace string,
...
+	}
+	if err := upgradeBackingImageDataSources(namespace, lhClient); err != nil {
+		return err
+	}
+	if err := upgradeBackingImageManagers(namespace, lhClient); err != nil {
+		return err
+	}
+	if err := upgradeBackingImages(namespace, lhClient); err != nil {
+		return err
+	}
```
shuo-wu, line 61: For these CRs, it's better to check if the objects are modified before updating them.

shuo-wu (changes requested):
Others LGTM

upgrade/v1beta1/upgrade.go
derekbit: Hmm, but how to check if the objects are modified?

datastore/longhorn.go
```diff
@@ -375,6 +376,9 @@ func GetOwnerReferencesForRecurringJob(recurringJob *longhorn.RecurringJob) []me
 
 // CreateVolume creates a Longhorn Volume resource and verifies creation
 func (s *DataStore) CreateVolume(v *longhorn.Volume) (*longhorn.Volume, error) {
+	if err := mayInitVolume(v); err != nil {
```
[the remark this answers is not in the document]
derekbit: No prob.

upgrade/v1beta1/upgrade.go
jenting: > Hmm, but how to check if the objects are modified?

- copy the existing CR before doing any operation on the CR
- run reflect.DeepEqual check before call the Update/UpdateStatus API

jenting (approved):

shuo-wu (approved):

upgrade/v1beta1/upgrade.go
```diff
@@ -120,6 +140,49 @@ func upgradeVolumes(namespace string, lhClient *lhclientset.Clientset) error {
...
+				return err
+			}
+			if dst.Groups == nil {
+				dst.Groups = []string{}
+			}
+			if dst.Labels == nil {
+				dst.Labels = make(map[string]string, 0)
+			}
+			obj.Spec.RecurringJobs[i] = dst
+		}
```
innobead, line 175: Maybe you can just reuse `initXXX` here instead of repeating the implementation, but you need to make `initXXX` exported first. WDYT?

innobead (changes requested):
LGTM, but one comment needed to resolve.

upgrade/v1beta1/upgrade.go
derekbit: Agree, it's better to reuse the functions, but the `datastore/longhorn.go` and `upgrade/v1beta1/upgrade.go` handle `v1beta2` and `v1beta2` objects, respectively.
Seems no simple way to handle them into one function. In addition, `v1beta1` objects' fields are not changed anymore, but the fields in `v1beta2` might be changed in the future. May handling them be better separately?

innobead: I think what you want to say is `datastore/longhorn.go` is handling `v1beta1`.  Ok, it seems like a limitation for now except we have another reflection layer to know which type to handle but it will be an over engineering. 

Let's keep this as is.

innobead (approved):

upgrade/v1beta1/upgrade.go
derekbit: Yes, sorry for the typo.

PR #1171, opened 2021-12-06
Handle the failure of the upgrade to v1beta2

derekbit:
Add the label to the upgraded CRs.
The re-run upgrade process only needs to process the non-upgraded CRs.

Longhorn 3354

### DIFF
--- a/types/types.go
+++ b/types/types.go
@@ -84,6 +84,7 @@ const (
 	LonghornLabelBackupVolume             = "backup-volume"
 	LonghornLabelRecurringJob             = "job"
 	LonghornLabelRecurringJobGroup        = "job-group"
+	LonghornLabelCRDAPIVersion            = "crd-api-version"
 
 	LonghornLabelValueEnabled = "enabled"
 
@@ -254,6 +255,10 @@ func GetBaseLabelsForSystemManagedComponent() map[string]string {
 
 func GetLonghornLabelComponentKey() string {
 	return GetLonghornLabelKey("component")
+}
+
+func GetLonghornLabelCRDAPIVersionKey() string {
+	return GetLonghornLabelKey(LonghornLabelCRDAPIVersion)
 }
 
 func GetEngineImageLabels(engineImageName string) map[string]string {
--- a/upgrade/v1beta1/upgrade.go
+++ b/upgrade/v1beta1/upgrade.go
@@ -3,12 +3,14 @@ package v1beta1
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/jinzhu/copier"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	restclient "k8s.io/client-go/rest"
@@ -74,7 +76,9 @@ func CanUpgrade(config *restclient.Config, namespace string) (bool, error) {
 }
 
 func upgradeVolumes(namespace string, lhClient *lhclientset.Clientset) error {
-	volumes, err := lhClient.LonghornV1beta1().Volumes(namespace).List(context.TODO(), metav1.ListOptions{})
+	volumes, err := lhClient.LonghornV1beta1().Volumes(namespace).List(context.TODO(), metav1.ListOptions{
+		LabelSelector: getLabelSelectorNotEqualV1beta2APIVersion(),
+	})
 	if err != nil {
 		return errors.Wrapf(err, "unable to list volumes")
 	}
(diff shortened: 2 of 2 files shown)


types/types.go
```diff
@@ -84,6 +84,7 @@ const (
 	LonghornLabelBackupVolume             = "backup-volume"
 	LonghornLabelRecurringJob             = "job"
 	LonghornLabelRecurringJobGroup        = "job-group"
+	LonghornLabelCRDVersion               = "crd-version"
```
shuo-wu, line 87: It's better to call it (Longhorn) API version rather than CRD version...

upgrade/v1beta1/upgrade.go
```diff
@@ -74,7 +75,9 @@ func CanUpgrade(config *restclient.Config, namespace string) (bool, error) {
 }
 
 func upgradeVolumes(namespace string, lhClient *lhclientset.Clientset) error {
-	volumes, err := lhClient.LonghornV1beta1().Volumes(namespace).List(context.TODO(), metav1.ListOptions{})
+	volumes, err := lhClient.LonghornV1beta1().Volumes(namespace).List(context.TODO(), metav1.ListOptions{
+		LabelSelector: types.GetLonghornLabelCRDVersionKey() + "!=" + strings.Split(types.CRDAPIVersionV1beta2, "/")[1],
```
shuo-wu, line 79: Abstract `types.GetLonghornLabelCRDVersionKey() + "!=" + strings.Split(types.CRDAPIVersionV1beta2, "/")[1]` as an separate function.

upgrade/v1beta1/upgrade.go
```diff
@@ -192,15 +210,74 @@ func upgradeNodes(namespace string, lhClient *lhclientset.Clientset) error {
...
 		}
+		if err := addCRDVersionLabelToNode(obj, namespace, lhClient); err != nil {
+			return errors.Wrapf(err, "failed to update for %v %v", old.Kind, old.Name)
+		}
 	}
 	logrus.Info("Finished upgrading nodes")
 	return nil
 }
 
+func addCRDVersionLabelToVolume(obj *longhorn.Volume, namespace string, lhClient *lhclientset.Clientset) error {
```
shuo-wu, line 225: Maybe you use use a single function like [this](https://github.com/longhorn/longhorn-manager/blob/7bd89abdbae3c406ec56094364cf6c2da1d36d9d/datastore/longhorn.go#L316) rather than one labeling function for each CRD.

shuo-wu (changes requested):

types/types.go
jenting: longhorn.io/v1beta1 and longhorn.io/v1beta2 are extended APIs of Longhorn.
For the CRD version, it's apiextensions.k8s.io/v1beta1 and apiextensions.k8s.io/v1 from my point of view.
So, personally, I think Shou's comment is fair.

derekbit: Understood! Thanks for the explanation.

upgrade/v1beta1/upgrade.go
```diff
@@ -74,7 +76,9 @@ func CanUpgrade(config *restclient.Config, namespace string) (bool, error) {
 }
 
 func upgradeVolumes(namespace string, lhClient *lhclientset.Clientset) error {
-	volumes, err := lhClient.LonghornV1beta1().Volumes(namespace).List(context.TODO(), metav1.ListOptions{})
+	volumes, err := lhClient.LonghornV1beta1().Volumes(namespace).List(context.TODO(), metav1.ListOptions{
+		LabelSelector: GetLabelSelectorNotEqualCurrentCRDAPIVersion(),
```
shuo-wu, line 80: NIT: `GetLabelSelectorNotEqualV1beta2APIVersion`

shuo-wu (commented):
In general LGTM now.

jenting (dismissed):

jenting:
Nit: probably we could name the label the same as https://github.com/longhorn/longhorn-manager/blob/3163c45/types/setting.go#L60

So, the label would be  `longhorn.io/crd-api-version=v1beta2` _or_ `longhorn.io/crd-api-version=longhorn.io/v1beta2`.

upgrade/v1beta1/upgrade.go
```diff
@@ -192,15 +223,43 @@ func upgradeNodes(namespace string, lhClient *lhclientset.Clientset) error {
...
+		labels = map[string]string{}
+	}
+
+	labels[types.GetLonghornLabelCRDAPIVersionKey()] = strings.Split(types.CRDAPIVersionV1beta2, "/")[1]
+	metadata.SetLabels(labels)
+
+	return nil
+}
+
+func GetLabelSelectorNotEqualV1beta2APIVersion() string {
```
innobead, line 259: If not exported, suggest using the name staring with a lower case.

derekbit: Okay. Thanks for the reminder.

innobead (approved):
LGTM, just some comments to resolve

upgrade/v1beta1/upgrade.go
```diff
@@ -100,17 +104,26 @@ func upgradeVolumes(namespace string, lhClient *lhclientset.Clientset) error {
...
 		if err != nil {
 			return errors.Wrapf(err, "failed to update for %v status %v", old.Kind, old.Name)
 		}
+
+		if err := tagCRLabelCRDAPIversion(obj); err != nil {
+			return errors.Wrapf(err, "failed to add label to %v %v", old.Kind, old.Name)
+		}
+		if _, err := lhClient.LonghornV1beta2().Volumes(namespace).Update(context.TODO(), obj, metav1.UpdateOptions{}); err != nil {
+			return errors.Wrapf(err, "failed to update for %v %v", old.Kind, old.Name)
+		}
```
jenting, line 117: nit: suggestion to use `new.Kind` and `new.Name` rather than `old.Kind` and `old.Name`.

innobead (approved):

PR #1172, opened 2021-12-06
Change default setting from ConfigMap to Setting CR

jenting:
#### Proposal Change
Create a new overlay folder to customize the default setting after longhorn manager DaemonSet.
Also, the customize default setting is changed from ConfigMap to Setting CRs.

#### Issue

### DIFF
--- a/app/daemon.go
+++ b/app/daemon.go
@@ -109,21 +109,10 @@ func startManager(c *cli.Context) error {
 	}
 	kubeconfigPath := c.String(FlagKubeConfig)
 
-	defaultSettingPath := os.Getenv(types.EnvDefaultSettingPath)
-
 	if err := environmentCheck(); err != nil {
 		logrus.Errorf("Failed environment check, please make sure you " +
 			"have iscsiadm/open-iscsi installed on the host")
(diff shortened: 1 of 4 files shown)


shuo-wu (changes requested):
These codes should be removed as well:
https://github.com/longhorn/longhorn-manager/blob/7bd89abdbae3c406ec56094364cf6c2da1d36d9d/app/daemon.go#L120-L127

Besides, why do you comment the new YAML file in the 2nd commit?

jenting:
> Besides, why do you comment the new YAML file in the 2nd commit?

Well, I'd like to make sure the longhorn-manager DaemonSet deploy first and then the Setting CRs.
However, I realized that the ordering of longhorn-manager DaemonSet and Setting CRs is irrelevant because the longhorn-manager create a new Setting CR with a default value when the Setting CR does not exist.

Drop the 2nd commit.

shuo-wu (approved):
LGTM. Let's see if the system upgrade via Rancher APP would lead to user-modified setting CRs being overwritten.

innobead (approved):
LGTM

PR #1173, opened 2021-12-08
Make sure backing images can be handled even if spec.disks is not initialized

shuo-wu:
Longhorn/longhorn#3379


innobead (approved):
LGTM

innobead:
@shuo-wu I assume this fix is only for 1.1.3.

joshimoo (approved):
LGTM

PR #1177, opened 2021-12-09
Pass the DNS suffix to the replica instance manager

innobead:
Signed-off-by: Keith Lucas <keith.lucas@suse.com>


joshimoo (approved):
LGTM

joshimoo (approved):
LGTM, the commit name here is wrong since it sets the suffix on the engine pod.

joshimoo (commented):
Okay this one is wrong, the HOST_SUFFIX needs to be applied to the replica-instance-manager creation

PR #1183, opened 2021-12-13
Add  validating admission webhook for Setting CRs

jenting:
#### Proposal Change
- Leverage https://github.com/rancher/webhook to implement the validating/mutating admission webhook skeleton.
  - The webhook server runs with HTTPS protocol with port 9443
  - Configure the ValidatingWebhookConfiguration and MutatingWebhookConfiguration
    ```shell
    $ kubectl get validatingwebhookconfigurations
    NAME                         WEBHOOKS   AGE
    longhorn-webhook-validator   1          20s
    $ kubectl get mutatingwebhookconfigurations
    NAME                       WEBHOOKS   AGE
    longhorn-webhook-mutator   1          20s
    ```
- Add validating admission webhook for Setting CRs.
- Deploy the longhorn webhook server as Deployment with default 2 replicas and podAntiAffinity to `preferredDuringSchedulingIgnoredDuringExecution`.
  ```shell
  $ kubectl get pod -l app=longhorn-webhook
  NAME                                READY   STATUS    RESTARTS   AGE
  longhorn-webhook-5c7bb86877-sf4td   1/1     Running   0          4m24s
  longhorn-webhook-5c7bb86877-vhwnn   1/1     Running   0          4m22s
  ```
- The webhook server listens on port 9443 and the longhorn-backend service handles port 9443 as well.
  ```shell
  $ kubectl get service longhorn-webhook
  NAME               TYPE        CLUSTER-IP     EXTERNAL-IP   PORT(S)    AGE
  longhorn-webhook   ClusterIP   10.43.204.23   <none>        9443/TCP   11m
  ```
- The webhook certificate is auto-generated and store in the Kubernetes secret, the certificate expired time is:
  - 10 years for CA certificate
  - 1 year for server certificate
    ```shell
    $ kubectl get secret 
    NAME                                   TYPE                                  DATA   AGE
    longhorn-webhook-ca                    kubernetes.io/tls                     2      29m
    longhorn-webhook-tls                   kubernetes.io/tls                     2      29
    ```
- The uninstall job will
  - Clean up the longhorn webhook Deployment
  - Clean up the ValidatingWebhookConfiguration and MutatingWebhookConfiguration

#### Issue
https://github.com/longhorn/longhorn/issues/3241


webhook/resources/setting/validator.go
```diff
@@ -0,0 +1,50 @@
...
+		},
+	}
+}
+
+func (v *settingValidator) Create(request *admission.Request, newObj runtime.Object) error {
+	return validateSetting(newObj)
+}
+
+func (v *settingValidator) Update(request *admission.Request, oldObj runtime.Object, newObj runtime.Object) error {
+	return validateSetting(newObj)
```
shuo-wu, line 40: Need to refuse modifications happens on read-only settings. The function you are using is for **init** settings.

webhook/server/server.go
```diff
@@ -0,0 +1,187 @@
...
+
+	ValidatingWebhookName = WebhookPrefix + "-validator"
+	MutatingWebhookName   = WebhookPrefix + "-mutator"
+
+	caName   = WebhookPrefix + "-ca"
+	certName = WebhookPrefix + "-tls"
+)
+
+var (
+	port = int32(9443)
```
shuo-wu: make this as a const in `types/types.go`

shuo-wu (changes requested):
You need to update the network policy as well: 
https://longhorn.io/docs/1.2.3/references/networking/
https://github.com/longhorn/longhorn-manager/tree/master/examples/network-policy

webhook/resources/setting/validator.go
jenting: But the RESTful API setting calls the `validateSetting` function as well.
- https://github.com/longhorn/longhorn-manager/blob/3163c45/api/setting.go#L46-L62
- https://github.com/longhorn/longhorn-manager/blob/3163c45/manager/setting.go#L42-L43

If I rename the function `ValidateInitSetting` to `ValidateSetting`, would it be better?

jenting:
> You need to update the network policy as well: https://longhorn.io/docs/1.2.3/references/networking/ https://github.com/longhorn/longhorn-manager/tree/master/examples/network-policy

Thank you for pointing it out.
I check the file https://github.com/longhorn/longhorn-manager/blob/7bad73d/examples/network-policy/manager-network-policy.yaml and found that the rules are based on the label, however, I did not introduce a new label for longhorn webhook Pods. And keep using the label `app=longhorn-manager`.

webhook/resources/setting/validator.go
shuo-wu: OK. Don't you prefer to add a more restricted limitation based on the read-only field?
And I just realized that you didn't use function https://github.com/longhorn/longhorn-manager/blob/023f9f3bc704b867984de5d47a70663be93a6407/datastore/longhorn.go#L93:21

shuo-wu:
> Thank you for pointing it out. I check the file https://github.com/longhorn/longhorn-manager/blob/7bad73d/examples/network-policy/manager-network-policy.yaml and found that the rules are based on the label, however, I did not introduce a new label for longhorn webhook Pods. And keep using the label `app=longhorn-manager`.

Oh, YES. There is no need to do more about it. The current network policy does not add any limitation for the port. Sorry about that.

webhook/resources/setting/validator.go
jenting: Thanks for the reminder, I did not consider the read-only field before.

I add it to verify the read-only field modification should be refused.
Also, change to use the function https://github.com/longhorn/longhorn-manager/blob/023f9f3bc704b867984de5d47a70663be93a6407/datastore/longhorn.go#L93:21.

types/setting.go
```diff
@@ -757,6 +757,9 @@ func ValidateInitSetting(name, value string) (err error) {
 	if definition.Required && value == "" {
 		return fmt.Errorf("required setting %v shouldn't be empty", sName)
 	}
+	if definition.ReadOnly && value != "" {
+		return fmt.Errorf("setting %v is read-only", sName)
+	}
```
shuo-wu, line 762: For the setting CR creation, this check should not be applied. In other words, this check is not for **init** setting validation

jenting: I think the function name should be reworded, remove the `init` term so it won't be confused anymore.
After that, the function `ValidateSetting` would be called by
- CreateOrUpdateSetting: the RESTful API endpoint calls create/update the setting.
- validateSetting: the validating admission webhook calls create/update the Setting CR.

shuo-wu: Oh, OK. The read-only CR should neither be created/customized by users during installation nor be edited for an existing Longhorn. Putting the check here make sense, and you can rename the function if necessary

jenting: Okay, I choose to rename the function.

deploy/install/01-prerequisite/02-serviceaccount.yaml
```diff
@@ -53,6 +53,12 @@ rules:
 - apiGroups: ["metrics.k8s.io"]
   resources: ["pods", "nodes"]
   verbs: ["get", "list"]
+- apiGroups: ["apiregistration.k8s.io"]
+  resources: ["apiservices"]
+  verbs: ["*"]
+- apiGroups: ["admissionregistration.k8s.io"]
+  resources: ["mutatingwebhookconfigurations", "validatingwebhookconfigurations"]
+  verbs: ["*"]
```
shuo-wu, line 61: Do we need to add this for `longhorn-uninstall-role`?

shuo-wu (commented):
In general LGTM

deploy/install/01-prerequisite/02-serviceaccount.yaml
jenting: No, because we did not remove the `mutatingwebhookconfigurations` and `validatingwebhookconfigurations` after the uninstallation job.

webhook/server/server.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+
+	router := mux.NewRouter()
+	router.Handle(validationPath, validationHandler)
+	router.Handle(mutationPath, mutationHandler)
+	if err := s.listenAndServe(client, router, validationResources, mutationResources); err != nil {
+		return err
+	}
+
+	if err := client.Start(s.context); err != nil {
```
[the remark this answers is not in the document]
jenting: To be honest, I did a reference from https://github.com/rancher/webhook/blob/c85c8d7/pkg/server/server.go#L78-L84

deploy/install/01-prerequisite/02-serviceaccount.yaml
shuo-wu: When/Where does Longhorn operate these 2 kinds of resources?

jenting: https://github.com/longhorn/longhorn-manager/blob/87b7263df396dbdf9db9df00c51bc552767b816d/webhook/server/server.go#L96-L150

types/setting.go
jenting: Sorry, I just realized adding the checking for the ReadOnly fields is harder than I was thought.

Right now, I can't find a way to distinguish that whether the create/update comes from longhorn-manager _or_ the user.

For example, the longhorn-manager will create/update the ReadOnly field `default-engine-image`. Every time when the default-engine-image changes, the longhorn-manager updates the value
- https://github.com/longhorn/longhorn-manager/blob/5b64adfaa300d2765cfeba232b883cf35b28e1bc/app/daemon.go#L221-L228
- https://github.com/longhorn/longhorn-manager/blob/5b64adfaa300d2765cfeba232b883cf35b28e1bc/manager/setting.go#L42-L47
- https://github.com/longhorn/longhorn-manager/blob/5b64adfaa300d2765cfeba232b883cf35b28e1bc/datastore/longhorn.go#L81-L82

Then, the longhorn-webhook server receives the update request and will verify it's a ReadOnly field and should be updated.
- https://github.com/longhorn/longhorn-manager/blob/87b7263df396dbdf9db9df00c51bc552767b816d/webhook/resources/setting/validator.go#L40-L47

Therefore, both the longhorn-manager _or_ the user creates/updates the ReadOnly field would thru. the same longhorn-webhook server validator.

deploy/install/01-prerequisite/02-serviceaccount.yaml
shuo-wu: Doesn't Longhorn clean them up during uninstallation?

types/setting.go
shuo-wu: All right. Then let's remove the ReadOnly field check now...

deploy/install/01-prerequisite/02-serviceaccount.yaml
jenting: > Doesn't Longhorn clean them up during uninstallation?

Okay, I add the cleanup code in the uninstall job
- clean up the longhorn webhook Deployment
- clean up the ValidatingWebhookConfiguration and MutatingWebhookConfiguration.

types/setting.go
jenting: > All right. Then let's remove the ReadOnly field check now...

Removed.

app/daemon.go
```diff
@@ -142,10 +142,7 @@ func startManager(c *cli.Context) error {
 
 	monitoring.InitMonitoringSystem(logger, currentNodeID, ds, kubeconfigPath)
 
-	if err := ds.InitSettings(); err != nil {
-		return err
-	}
-
+	// Initialize the required settings first
 	if err := updateDefaultImageSetting(m, types.SettingNameDefaultEngineImage, engineImage); err != nil {
```
shuo-wu, line 146: NIT: It's better to rename this function as well. e.g., `createOrUpdateDefaultImageSetting`

shuo-wu (dismissed):
In general LGTM

webhook/server/server.go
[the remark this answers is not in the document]
shuo-wu: You can take a quick try and verify if you need to modify this. (e.g., calling this function before the server start, or adding a wait interval before calling this function)

deploy/install/02-components/05-webhook.yaml
```diff
@@ -0,0 +1,49 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
+    app: longhorn-webhook
+  name: longhorn-webhook
+  namespace: longhorn-system
```
derekbit, line 7: Is it better to add podAntiAffinity here?

jenting: Good suggestion. I added the `preferredDuringSchedulingIgnoredDuringExecution`.

app/daemon.go
jenting: Updated!

derekbit:
LGTM.

innobead (approved):
In general, LGTM. Just one comment.

webhook/admission/admission.go
```diff
@@ -0,0 +1,135 @@
...
+		logrus.Debugf("%s operation is rejected: %s", req, admitErr)
+		return
+	}
+
+	if len(patchOps) > 0 {
+		patchType := admissionv1.PatchTypeJSONPatch
+		patchData := fmt.Sprintf("[%s]", strings.Join(patchOps, ","))
+		response.PatchType = &patchType
+		response.Patch = []byte(patchData)
+		logrus.Debugf("%s patchOps: %s", req, patchData)
```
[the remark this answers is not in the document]
jenting: I think the default log level is [INFO](https://github.com/sirupsen/logrus/blob/79c5ab66aa2ce7d9ff7b3c437ebc22fcc519a967/logger.go#L92).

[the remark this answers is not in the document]
innobead: then all good 👍

PR #1188, opened 2021-12-21
change base image to SUSE BCI

meldafrawi:
Ref: https://github.com/longhorn/longhorn/issues/2939

Signed-off-by: Mohamed Eldafrawi <mohamed.eldafrawi@suse.com>


Dockerfile.dapper
```diff
@@ -1,27 +1,18 @@
...
-RUN wget ${!DOCKER_URL} -O docker_ce_${ARCH} && dpkg -i docker_ce_${ARCH}
+RUN zypper -n install gcc ca-certificates git wget curl vim less file python3-tox python3-devel iptables libdevmapper1_03 libltdl7 awk && \
+    rm -f /bin/sh && ln -s /bin/bash /bin/sh && \
+    zypper install -y -f docker
 
 ENV GOLANG_ARCH_amd64=amd64 GOLANG_ARCH_arm64=arm64 GOLANG_ARCH=GOLANG_ARCH_${ARCH} \
     GOPATH=/go PATH=/go/bin:/usr/local/go/bin:${PATH} SHELL=/bin/bash
 
-RUN wget -O - https://storage.googleapis.com/golang/go1.14.1.linux-${!GOLANG_ARCH}.tar.gz | tar -xzf - -C /usr/local && \
+RUN wget -O - https://storage.googleapis.com/golang/go1.13.15.linux-${!GOLANG_ARCH}.tar.gz | tar -xzf - -C /usr/local && \
```
meldafrawi, line 15: According to go.mod file, longhorn-manager is using golang v1.13

innobead: Don’t think we need to revert back to 1.13. The go version mentioned in go.mod should be min version.

innobead (dismissed):
LGTM. Will merge until the regression result ready.

Dockerfile.dapper
joshimoo: @PhanLe1010 since we updated kubernetes min version, did we end up having to update the min go version as well?
To be in line with the kubernetes/grpc/ etc versions :)

PhanLe1010: We didn't update the min go version since Kubernetes v1.18 is still using go 1.13

Dockerfile.dapper
```diff
@@ -1,27 +1,18 @@
...
-RUN wget ${!DOCKER_URL} -O docker_ce_${ARCH} && dpkg -i docker_ce_${ARCH}
+RUN zypper -n install gcc ca-certificates git wget curl vim less file python3-tox python3-devel iptables libdevmapper1_03 libltdl7 awk docker && \
+    rm -f /bin/sh && ln -s /bin/bash /bin/sh && \
+    rm -rf /var/cache/zypp/*
 
 ENV GOLANG_ARCH_amd64=amd64 GOLANG_ARCH_arm64=arm64 GOLANG_ARCH=GOLANG_ARCH_${ARCH} \
     GOPATH=/go PATH=/go/bin:/usr/local/go/bin:${PATH} SHELL=/bin/bash
 
-RUN wget -O - https://storage.googleapis.com/golang/go1.14.1.linux-${!GOLANG_ARCH}.tar.gz | tar -xzf - -C /usr/local && \
+RUN wget -O - https://storage.googleapis.com/golang/go1.14.15.linux-${!GOLANG_ARCH}.tar.gz | tar -xzf - -C /usr/local && \
```
PhanLe1010, line 15: We are using go 1.13 in go.mod but we specify go1.14.15 here. I am not sure if this is a problem

meldafrawi: fixed

Dockerfile.dapper
meldafrawi: will use the latest version of golang 1.13 which is v1.13.15, and we can discuss the upgrade separately later.

meldafrawi:
I've changed the golang version in Dockerfile.dapper to the latest v1.13.15 to match the version defined in go.mod

innobead (approved):

PR #1189, opened 2021-12-27
Add CRD conversion webhook

jenting:
#### Proposal Change
- Reference to the [kubebuilder Hub and Spoke doc](https://book.kubebuilder.io/multiversion-tutorial/conversion-concepts.html), [Kubernetes CRD versioning doc](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/), [design](https://github.com/kubernetes-sigs/kubebuilder/blob/8076235/designs/crd_version_conversion.md) and [code](https://github.com/droot/crd-conversion-example) to implement the CRD conversion webhook.
- Generate the schemaless for longhorn.io/v1beta1 CRD.
- Add the conversion webook for 
  - BackupTarget/EngineImage/Node/Volume CRs: `status.conditions` conversion between `map` (v1beta1) to `slice` (v1beta2).
  - BackingImage CR: `spec.disks` conversion between `map[string]struct{}` (v1beta1) to `map[string]string` (v1beta2).
- Remove the upgrade `status.conditions` from the map to slice code in the upgrade path because it should be handled by the CRD conversion webhook.
- Add `initContainers` in the longhorn manager DaemonSet to delay start until the longhorn webhook server is ready.

#### Issues
- https://github.com/longhorn/longhorn/issues/3265
- https://github.com/longhorn/longhorn/issues/3523
- https://github.com/longhorn/longhorn/issues/3295
- https://github.com/longhorn/longhorn/issues/3407
- https://github.com/longhorn/longhorn/issues/3410

### DIFF
--- a/controller/uninstall_controller.go
+++ b/controller/uninstall_controller.go
@@ -410,6 +410,14 @@ func (c *UninstallController) deleteCRDs() (bool, error) {
 		return true, nil
 	}
 
+	// Delete the BackupTarget CRs
+	if backupTargets, err := c.ds.ListBackupTargets(); err != nil {
+		return true, err
+	} else if len(backupTargets) > 0 {
+		c.logger.Infof("Found %d backuptargets remaining", len(backupTargets))
+		return true, c.deleteBackupTargets(backupTargets)
+	}
+
 	if engineImages, err := c.ds.ListEngineImages(); err != nil {
 		return true, err
 	} else if len(engineImages) > 0 {
@@ -539,6 +547,22 @@ func (c *UninstallController) deleteReplicas(replicas map[string]*longhorn.Repli
 				return
 			}
 			log.Info("Removed finalizer")
+		}
+	}
+	return
+}
+
+func (c *UninstallController) deleteBackupTargets(backupTargets map[string]*longhorn.BackupTarget) (err error) {
+	defer func() {
+		err = errors.Wrapf(err, "Failed to delete backup targets")
+	}()
+	for _, bt := range backupTargets {
+		log := getLoggerForBackupTarget(c.logger, bt)
+		if bt.DeletionTimestamp == nil {
+			if err = c.ds.DeleteBackupTarget(bt.Name); err != nil {
+				return errors.Wrapf(err, "Failed to mark for deletion")
+			}
+			log.Info("Marked for deletion")
 		}
 	}
 	return
--- a/k8s/pkg/apis/longhorn/v1beta2/backuptarget.go
+++ b/k8s/pkg/apis/longhorn/v1beta2/backuptarget.go
@@ -71,3 +71,7 @@ type BackupTargetList struct {
 	metav1.ListMeta `json:"metadata,omitempty"`
 	Items           []BackupTarget `json:"items"`
 }
+
+// Hub defines the current version (v1beta2) is the storage version
+// so mark this as Hub
+func (bt *BackupTarget) Hub() {}
--- a/upgrade/v1beta1/upgrade.go
+++ b/upgrade/v1beta1/upgrade.go
@@ -4,64 +4,58 @@ import (
 	"context"
 	"fmt"
 	"reflect"
-	"strings"
 
 	"github.com/jinzhu/copier"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
-	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	restclient "k8s.io/client-go/rest"
 
 	longhornV1beta1 "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta1"
-	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
-
-	"github.com/longhorn/longhorn-manager/types"
-
 	lhclientset "github.com/longhorn/longhorn-manager/k8s/pkg/client/clientset/versioned"
+	"github.com/longhorn/longhorn-manager/types"
 )
 
 const (
-	upgradeLogPrefix = "upgrade from v1beta1 to v1beta2:"
+	fixupCRLogPrefix = "fix up CRs:"
 )
 
-func UpgradeCRFromV1beta1ToV1beta2(config *restclient.Config, namespace string, lhClient *lhclientset.Clientset) (err error) {
+// FixupCRs initialize the old CR spec which is with the `null` field.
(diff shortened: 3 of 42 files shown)


jenting:
Note, the leftover is the backing image conversion, however, we need to revert the `spec.disks` struct from `map[string]string` back to `map[string]struct{}`. However, the controller-gen tool doesn't support generating the `struct{}`. Therefore, we need to find a way to skip the field. I'd prefer to file another PR to handle this issue.

k8s/pkg/apis/longhorn/v1beta2/backuptarget.go
```diff
@@ -71,3 +71,7 @@ type BackupTargetList struct {
 	metav1.ListMeta `json:"metadata,omitempty"`
 	Items           []BackupTarget `json:"items"`
 }
+
+// Hub defines the current version (v1beta2) is the storage version
+// so mark this as Hub
+func (bt *BackupTarget) Hub() {}
```
innobead, line 77: So it means we need to change the hub interface implementation to the next version of resources from v1beta2 if we have v1beta3 for example.

jenting: Once `v1beta3` introduce and it's the hub, and both v1beta1 and v1beta2 keeps, we need to change the hub interface implementation under `v1beta3` package, and add conversion from v1beta1 >-> v1beta3 in the v1beta1 package, and v1beta2 <-> v1beta3 in the v1beta2 package as well.

If only v1beta2 keeps and v1beta1 drop, we only need to add conversion from v1beta2 <-> v1beta3 in the v1beta2 package.

innobead (approved):
LGTM

k8s/pkg/apis/longhorn/v1beta2/backuptarget.go
innobead: thanks, also as the pattern we probably will adopt is just keep one convertible version like you mentioned before.

controller/uninstall_controller.go
```diff
@@ -410,6 +410,14 @@ func (c *UninstallController) deleteCRDs() (bool, error) {
 		return true, nil
 	}
 
+	// Delete the BackupTarget CRs
+	if backupTargets, err := c.ds.ListBackupTargets(); err != nil {
+		return true, err
+	} else if len(backupTargets) > 0 {
+		c.logger.Infof("Found %d backuptargets remaining", len(backupTargets))
+		return true, c.deleteBackupTargets(backupTargets)
+	}
```
derekbit, line 419: Leave a note here
```
Derek Su  3:22 PM
Why didn’t we delete the backupTarget CRs before?

JenTing Hsiao  3:24 PM
Because the BackupTarget CR without finalizer
So, it fine when we delete CRD, the CR will be deleted as well.
However, once the webhook introduce, if we keep the same that the BackupTarget CR be deleted when we delete CRD.
After the uninstallation job finished, we can't delete BackupTarget CR success anymore because the longhorn-webhook server gone.
```

derekbit:
LGTM

derekbit (dismissed):
LGTM

webhook/server/server.go
```diff
@@ -90,12 +94,45 @@ func (s *WebhookServer) listenAndServe(client *client.Client, handler http.Handl
...
+							Namespace: s.namespace,
+							Name:      webhookServiceName,
+							Path:      &conversionPath,
+							Port:      &port,
+						},
+						CABundle: secret.Data[corev1.TLSCertKey],
+					},
+					ConversionReviewVersions: []string{"v1beta2", "v1beta1"},
+				},
+			}
```
shuo-wu, line 119: It seems that we can directly add this field in the CRD YAML file rather than updating it after the installation:
```
conversion:
    strategy: Webhook
    webhook:
      clientConfig:
        service:
          name: longhorn-webhook
          namespace: longhorn-system
          path: "/v1/webhook/conversion"
          port: 9443
      conversionReviewVersions:
        - v1beta1
        - v1beta2
```

I am not sure which one is better. What do you think?

shuo-wu (commented):
I am confused about the CR upgrade after introducing commit `upgrade: remove upgrade status.conditions code`. The operations in this commit are more like v1beta2 CR initialization after the upgrade. When will Longhorn upgrade the hub/storage version from v1beta1 to v1bet2 for the old CRs? Or, does not it matter at all?

webhook/server/server.go
jenting: The `caBundle` is the CA certificate which is a byte slice.

I follow the pattern that how we register the ValidatingWebhookConfiguration and MutatingWebhookConfiguration.

shuo-wu: OK. CA certificate setting may be a problem for the YAML file. Let's retain the current implementation.

shuo-wu:
I know the conversion functions will be invoked if the current hub version of a CR does not match the version users want to get. 

My question is, should Longhorn keep the hub/storage version of all CRs identical (convert all old CRs from v1beta1 to v1beta2)? Is the answer No since we implement `convertTo`?

jenting:
> I know the conversion functions will be invoked if the current hub version of a CR does not match the version users want to get.
> 
> My question is, should Longhorn keep the hub/storage version of all CRs identical (convert all old CRs from v1beta1 to v1beta2)? Is the answer No since we implement `convertTo`?

I think is no.
Besides that, Derek had work migrating all the old CRs from v1beta1 -> v1beta2, however, the kube-manager-controller still uses the v1beta1 to access the EngineImage CR. Ref to 
- https://github.com/longhorn/longhorn/issues/3407#issuecomment-995088068
- https://github.com/longhorn/longhorn/issues/3410#issuecomment-995339068

shuo-wu:
OK. I see. Keeping the API version unchanged for old CRs makes sense.

upgrade/v1beta1/upgrade.go
```diff
@@ -4,30 +4,30 @@ import (
...
 const (
 	upgradeLogPrefix = "upgrade from v1beta1 to v1beta2:"
 )
 
+// UpgradeCRFromV1beta1ToV1beta2 initialize the old CR spec which is with the `null` field.
+// Since start from longhorn.io/v1beta2, we introduce the CRD structural schema.
+// However, for the Kubernetes < 1.20, there is an issue that it does not allow
+// the `null` field because of the CRD validation failure.
+// Reference issue: https://github.com/kubernetes/kubernetes/issues/86811`
 func UpgradeCRFromV1beta1ToV1beta2(config *restclient.Config, namespace string, lhClient *lhclientset.Clientset) (err error) {
```
shuo-wu, line 31: It's better to rename this and the below functions in this file. e.g., `FixupV1beta1CRsForConversion`

shuo-wu (commented):
In general LGTM

innobead (dismissed):

innobead (approved):

innobead:
As per the review comments, just merged.

PR #1190, opened 2022-01-03
controller: Fix NPE issue when trying to get the client for a nil engine

shuo-wu:
Longhorn/longhorn#3484


jenting (approved):
Thank you for helping fix the bug.

innobead (approved):

PR #1191, opened 2022-01-04
crd: add recurring job retain count max check

jenting:
https://github.com/longhorn/longhorn/issues/3493

### DIFF
--- a/deploy/install/01-prerequisite/03-crd.yaml
+++ b/deploy/install/01-prerequisite/03-crd.yaml
@@ -2742,6 +2742,7 @@ spec:
                 type: string
               retain:
                 description: The retain count of the snapshot/backup.
+                maximum: 50
                 minimum: 1
                 type: integer
               task:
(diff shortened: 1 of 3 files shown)


shuo-wu (commented):
LGTM. Will the creation of RecurringJob CR get failed automatically if the retain count is set to >50?

innobead (approved):
LGTM

jenting:
> LGTM. Will the creation of RecurringJob CR get failed automatically if the retain count is set to >50?

Yes, apply the following yaml manifest
```yaml
kind: RecurringJob
metadata:
  name: snap 
  namespace: longhorn-system
spec:
  cron: "0 */5 * * *"
  task: "snapshot"
  groups:
    - default
  retain: 15000
  concurrency: 1
```

would fail with error message
```shell
The RecurringJob "snap" is invalid: spec.retain: Invalid value: 15000: spec.retain in body should be less than or equal to 50
```

PR #1196, opened 2022-01-06
Fix bug volume cloning feature is broken

PhanLe1010:
The field DataSource should be a string with full information about the source of volume. For example, `snap://vol-name/snapshot-name`

longhorn/longhorn#3508

### DIFF
--- a/csi/deployment.go
+++ b/csi/deployment.go
@@ -397,7 +397,8 @@ func NewPluginDeployment(namespace, serviceAccount, nodeDriverRegistrarImage, ma
 								},
 								AllowPrivilegeEscalation: pointer.BoolPtr(true),
 							},
-							Image: managerImage,
+							Image:           managerImage,
+							ImagePullPolicy: imagePullPolicy,
 							Lifecycle: &v1.Lifecycle{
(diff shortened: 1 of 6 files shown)


jenting:
Looks like the volume cloning feature, it does not follow the type definitions https://github.com/longhorn/longhorn-manager/blob/c72e5b9654fb32541db92eb6e9a171ef19de25ef/k8s/pkg/apis/longhorn/v1beta2/volume.go#L34-L41?

PhanLe1010:
I think it is a typo in `longhorn-manager/k8s/pkg/apis/longhorn/v1beta2/volume.go`, maybe we should change the type `VolumeDataSource` to `VolumeDataSourceType`

There are 4 VolumeDataSourceType but the field volume.Spec.DataSource is intended to have the format `type-prefix://volume-name/snapshot-name`

jenting:
I see, then could you make the change in this PR as well?

PhanLe1010:
Thanks @jenting for helping with the CRDs schema.

I pushed the change

jenting (approved):
LGTM, thank you.

innobead (approved):

PR #1197, opened 2022-01-07
Change the lifecycle management of the instance manager pods

shuo-wu:
https://github.com/longhorn/longhorn/issues/2650
https://github.com/longhorn/longhorn/issues/3459

### DIFF
--- a/controller/instance_manager_controller.go
+++ b/controller/instance_manager_controller.go
@@ -1,7 +1,6 @@
 package controller
 
 import (
-	"context"
 	"encoding/json"
 	"fmt"
 	"reflect"
@@ -53,7 +52,7 @@ type InstanceManagerController struct {
 
 	cacheSyncs []cache.InformerSynced
 
-	instanceManagerMonitorMutex *sync.RWMutex
+	instanceManagerMonitorMutex *sync.Mutex
 	instanceManagerMonitorMap   map[string]chan struct{}
 
 	// for unit test
@@ -66,13 +65,12 @@ type InstanceManagerMonitor struct {
 	Name         string
 	controllerID string
 
-	instanceManagerUpdater  *InstanceManagerUpdater
-	instanceManagerNotifier *InstanceManagerNotifier
-	ds                      *datastore.DataStore
-	lock                    *sync.RWMutex
-	updateNotification      bool
-	stopCh                  chan struct{}
-	done                    bool
+	instanceManagerUpdater *InstanceManagerUpdater
+	ds                     *datastore.DataStore
+	lock                   *sync.RWMutex
+	updateNotification     bool
+	stopCh                 chan struct{}
+	done                   bool
 	// used to notify the controller that monitoring has stopped
 	monitorVoluntaryStopCh chan struct{}
 }
@@ -122,7 +120,7 @@ func NewInstanceManagerController(
 
 		ds: ds,
 
-		instanceManagerMonitorMutex: &sync.RWMutex{},
+		instanceManagerMonitorMutex: &sync.Mutex{},
 		instanceManagerMonitorMap:   map[string]chan struct{}{},
 
 		versionUpdater: updateInstanceManagerVersion,
@@ -255,8 +253,8 @@ func (imc *InstanceManagerController) syncInstanceManager(key string) (err error
 	im, err := imc.ds.GetInstanceManager(name)
 	if err != nil {
 		if datastore.ErrorIsNotFound(err) {
-			logrus.Infof("Longhorn instance manager %v has been deleted", key)
-			return nil
+			logrus.Infof("Longhorn instance manager %v has been deleted, will try best to do cleanup", name)
+			return imc.cleanupInstanceManager(name)
 		}
 		return err
 	}
@@ -281,7 +279,7 @@ func (imc *InstanceManagerController) syncInstanceManager(key string) (err error
 	}
 
 	if im.DeletionTimestamp != nil {
-		return nil
+		return imc.cleanupInstanceManager(im.Name)
 	}
 
 	existingIM := im.DeepCopy()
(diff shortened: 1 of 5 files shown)


controller/instance_manager_controller.go
```diff
@@ -296,140 +295,185 @@ func (imc *InstanceManagerController) syncInstanceManager(key string) (err error
...
-				default:
-					log.Errorf("BUG: instance manager %v has invalid type %v", im.Name, im.Spec.Type)
-					im.Status.CurrentState = longhorn.InstanceManagerStateError
-				}
-			}
+			disableMonitor = true
+		} else { // The monitor will be enabled only when the InstanceManager is running and compatible.
+			enableMonitor = true
 		}
+	} else if im.Status.CurrentState != longhorn.InstanceManagerStateUnknown { // For now Longhorn won't actively disable or enable monitoring when the InstanceManager is Unknown.
```
innobead, line 453: I assume ignoring the unknown state should be related the state should be transient.

> For some reason the state of the Pod could not be obtained. This phase typically occurs due to an error in communicating with the node where the Pod should be running.

innobead (approved):
LGTM.

controller/instance_manager_controller.go
```diff
@@ -296,140 +295,185 @@ func (imc *InstanceManagerController) syncInstanceManager(key string) (err error
...
+		if err := imc.versionUpdater(im); err != nil {
+			return err
+		}
 	}
+	return nil
+}
 
+func (imc *InstanceManagerController) syncMonitor(im *longhorn.InstanceManager) error {
+	enableMonitor := false
+	disableMonitor := false
```
joshimoo, line 446: NIT: combine these into one boolean `isMonitorRequired := im.Status.CurrentState == longhorn.InstanceManagerStateRunning` since they are exclusive of each other (true && true wouldn't make much sense)

Then you can just do this, with a bit of easy refactor, see my previous monitor suggestions
No need to check if isMonitoring the enable/disable should handle that internally.
```
if im.state == unknown {
    return nil
}

if isMonitorRequired {
   return enableMonitor(im)
} else {
   return disableMonitor(im)
}
```

controller/instance_manager_controller.go
```diff
@@ -462,7 +506,7 @@ func (imc *InstanceManagerController) syncInstanceManagerPDB(im *longhorn.Instan
 	return nil
 }
 
-func (imc *InstanceManagerController) cleanUpPDBOfNonExistingIMPod() error {
+func (imc *InstanceManagerController) cleanUpPDBForNonExistingIM() error {
```
joshimoo, line 509: NIT: for now this is okay, but in the future we might want to move this into a gc controller, since now each IM will evaluate all PDBs in the longhorn-system namespace. Also all of the functions use copies even though they are all readonly.

controller/instance_manager_controller.go
```diff
@@ -740,25 +776,20 @@ func (imc *InstanceManagerController) enqueueKubernetesNode(obj interface{}) {
...
-	log := getLoggerForInstanceManager(imc.logger, im)
-	im.Status.IP = ""
-	if len(im.Status.Instances) > 0 {
-		im.Status.Instances = map[string]longhorn.InstanceProcess{}
-	}
-	if imc.isMonitoring(im.Name) {
-		imc.stopMonitoring(im)
+func (imc *InstanceManagerController) cleanupInstanceManager(imName string) error {
+	if imc.isMonitoring(imName) {
+		imc.stopMonitoring(imName)
```
joshimoo, line 781: NIT: it's better to have just the disableMonitor call, mentioned above.

datastore/kubernetes.go
```diff
@@ -423,6 +423,16 @@ func (s *DataStore) GetKubernetesNode(name string) (*corev1.Node, error) {
 	return s.knLister.Get(name)
 }
 
+// IsKubeNodeUnschedulable checks if the Kubernetes Node resource is
+// unschedulable
+func (s *DataStore) IsKubeNodeUnschedulable(nodeName string) (bool, error) {
```
joshimoo, line 428: I used the IsNodeSchedulable from the lh node (which is synced by the manager), in the share-manager not sure what is better, probably using the kube node status directly is better.

joshimoo (approved):
LGTM some monitoring refactoring down the line would be nice, but this is better structured already.

joshimoo:
LGTM, can be merged, my comments are nits so feel free to ignore them or do later :)

controller/instance_manager_controller.go
```diff
@@ -296,140 +295,185 @@ func (imc *InstanceManagerController) syncInstanceManager(key string) (err error
...
+	if err := imc.syncMonitor(im); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// syncStatusWithPod updates the InstanceManager based on the pod current phase only,
+// regardless of the InstanceManager previous status.
+func (imc *InstanceManagerController) syncStatusWithPod(im *longhorn.InstanceManager) error {
```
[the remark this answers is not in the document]
shuo-wu: It's fine to have unit tests for this function even if the pod retrieval is inside it. It's better not to move the retrieval outside of the function since there is almost no other function that requires the pod info.

controller/instance_manager_controller.go
```diff
@@ -296,140 +295,185 @@ func (imc *InstanceManagerController) syncInstanceManager(key string) (err error
...
+	unschedulable, err := imc.ds.IsKubeNodeUnschedulable(im.Spec.NodeID)
+	if err != nil {
+		return err
+	}
+	if unschedulable {
+		return nil
+	}
+
+	if err := imc.createInstanceManagerPod(im); err != nil {
+		if !types.ErrorAlreadyExists(errors.Cause(err)) {
```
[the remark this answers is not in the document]
shuo-wu: I just moved this error handling inside the function, which could avoid going through the error hierarchy as well.

controller/instance_manager_controller.go
shuo-wu: Make sense. I will refactor this function based on your suggestions.

controller/instance_manager_controller.go
shuo-wu: Right. Besides, we can add labels for IM PDBs then do filtering here.

datastore/kubernetes.go
shuo-wu: Here we should check the kube node status.

For lh nodes, scheduling means replica allocation.
For kube nodes, scheduling means pod allocation.

controller/instance_manager_controller.go
shuo-wu: I refactored only a part of the monitor logic to fit the modification here. Further refactoring will be handled in https://github.com/longhorn/longhorn/issues/2441

controller/instance_manager_controller.go
shuo-wu: Yes. We don't know if the `unknown` is a transient state or not hence there is no need to do anything aggressively.

Once the status of the Ready condition remains `Unknown` or `False` for longer than the `pod-eviction-timeout`, the pod will be marked as terminating then the controller would mark the IM as state `error`.

https://kubernetes.io/docs/concepts/architecture/nodes/#condition

innobead (approved):

innobead:
Merged as per the discussions and updates.

controller/instance_manager_controller.go
innobead: For the gc controller as we discussed before, it's also possible for other possible orphan resources cleanup.

shuo-wu:
Should we backport this PR to v1.2.4?

innobead:
> Should we backport this PR to v1.2.4?

Yes and 1.1.4.

PR #1199, opened 2022-01-12
example: add longhorn webhook network policy yaml manifest

jenting:
#### Proposal Change

The traffic from/to the longhorn webhook server is the kube-apiserver.
The only way we could add restriction is to add the network policy of
the ingress port because we can't know each Kubernetes distro default
kube-apiserver Pod's label. Therefore, we can't add the label selector
in the network policy rule to restrict the traffic that comes from the
kube-apiserver is able to access to the longhorn webhook server.

#### Issue
https://github.com/longhorn/longhorn/issues/3513

### DIFF
--- a/examples/network-policy/webhook-network-policy.yaml
+++ b/examples/network-policy/webhook-network-policy.yaml
@@ -1,0 +1,15 @@
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: longhorn-webhook
+  namespace: longhorn-system
+spec:
+  podSelector:
+    matchLabels:
+      app: longhorn-webhook
+  policyTypes:
+  - Ingress
+  ingress:
+  - ports:
+    - protocol: TCP
+      port: 9443


examples/network-policy/webhook-network-policy.yaml
```diff
@@ -0,0 +1,15 @@
...
+  namespace: longhorn-system
+spec:
+  podSelector:
+    matchLabels:
+      app: longhorn-webhook
+  policyTypes:
+  - Ingress
+  ingress:
+  - ports:
+    - protocol: TCP
```
innobead, line 14: Does this work w/o UCP port allowed? for the case Mohamed found.

innobead (approved):
LGTM

examples/network-policy/webhook-network-policy.yaml
jenting: It's the UDP port `8472` which is the Canal/Flannel VXLAN overlay networking
https://rancher.com/docs/rancher/v2.5/en/installation/requirements/ports/#commonly-used-ports

I found that each host open the same UDP port with the command:
```
ss -ulpn
```

PR #1201, opened 2022-01-15
Fix CRD upgrade and conversion failures

derekbit:
1. CRD conversion failure of backingimages  because of the initialized variable
2. Upgrade failure from v1.1.x or below because of the empty string of volume.Spec.ReplicaAutoBalance
 
Longhorn 3265

### DIFF
--- a/upgrade/v1beta1/upgrade.go
+++ b/upgrade/v1beta1/upgrade.go
@@ -94,6 +94,9 @@ func fixupVolumes(namespace string, lhClient *lhclientset.Clientset) error {
 	for _, obj := range volumes.Items {
 		existing := obj.DeepCopy()
 
+		if obj.Spec.ReplicaAutoBalance == "" {
+			obj.Spec.ReplicaAutoBalance = longhornV1beta1.ReplicaAutoBalanceIgnored
+		}
 		if obj.Spec.DiskSelector == nil {
 			obj.Spec.DiskSelector = []string{}
 		}
--- a/k8s/pkg/apis/longhorn/v1beta1/backingimage.go
+++ b/k8s/pkg/apis/longhorn/v1beta1/backingimage.go
@@ -102,6 +102,7 @@ func (bi *BackingImage) ConvertTo(dst conversion.Hub) error {
 		}
 
 		// Copy spec.disks from map[string]struct{} to map[string]string
+		biV1beta2.Spec.Disks = make(map[string]string)
 		for name := range bi.Spec.Disks {
 			biV1beta2.Spec.Disks[name] = ""
 		}
@@ -125,6 +126,7 @@ func (bi *BackingImage) ConvertFrom(src conversion.Hub) error {
 		}
 
 		// Copy spec.disks from map[string]string to map[string]struct{}
+		bi.Spec.Disks = make(map[string]struct{})
 		for name := range biV1beta2.Spec.Disks {
 			bi.Spec.Disks[name] = struct{}{}
 		}


innobead (approved):

upgrade/v1beta1/upgrade.go
```diff
@@ -94,6 +94,9 @@ func fixupVolumes(namespace string, lhClient *lhclientset.Clientset) error {
 	for _, obj := range volumes.Items {
 		existing := obj.DeepCopy()
 
+		if obj.Spec.ReplicaAutoBalance == "" {
```
jenting, line 97: I'd recommend move this to v123 to v124 upgrade path and back port to 1.2.4 branch as well.

innobead: Sounds good but it means need to use v1beta1 in the 123 to 124 upgrade path for this update? and it’s nice to have that in 1.2.4.

jenting: Yes, and it uses v1beta1 API to operate now.

jenting:
Remember to mark this issue https://github.com/longhorn/longhorn/issues/3524 resolved as well, thank you @derekbit

upgrade/v1beta1/upgrade.go
derekbit: We still need to put the ReplicaAutoBalance in `upgrade/v1beta1/upgrade.go`, because `doCRUpgrade` is executed after `doAPIVersionUpgrade`. `doCRUpgrade` uses v1beta2 now, so the `volume` updateStatus failure occurs in v1.1.0 to v1.2.0 upgrade.

```
time="2022-01-16T08:19:37Z" level=info msg="upgrade from v1.1.0 to v1.2.0: Updating pvc-dd63e559-6809-4c40-881f-3b98ce382d02 volume labels to map[longhornvolume:pvc-dd63e559-6809-4c40-881f-3b98ce382d02 recurring-job.longhorn.io/backup-20-b5f4656e-77074ba4:enabled]" action=convert-volume-recurring-jobs-to-labels namespace=longhorn-system translate=volume-recurring-jobs
```

Another way is moving ReplicaAutoBalance to the conversion webhook. WDYT?

innobead: This is the way to do, but I would prefer making the conversion webhook focusing on the things which can't handled in the migration like non-backward compatibility. 

You should just create a `v123to124/upgrade.go`, then use `lhClient.LonghornV1beta1()` to do the migration to update ReplicaAutoBalance to the default value?

derekbit: > You should just create a v123to124/upgrade.go, then use lhClient.LonghornV1beta1() to do the migration to update ReplicaAutoBalance to the default value?

No... The `volume` read from `lhClient.LonghornV1beta1` does not have the `Spec.ReplicaAutoBalance` field, so the migration is not achievable by this way.

derekbit: Sorry. I mixed up the v1beta1 structs in v1.1.3 and master.
The current hurdle is the volume update fails on v110tov120 unless we replace all v1beta2 APIs with v1beta2 APIs in upgrade logics. The upgrade logics use datastore APIs only accepting v1beta2 objects.

So any migration impacted by validation, we now just put into fixxxxx only for the current implementation. We don't need to have backport now, because at the end, when users upgrade to 1.3.0, it will still be migrated

innobead: Thanks @derekbit , all fine now.

PR #1202, opened 2022-01-17
Fix CRD validation failure due to the null field(s)

derekbit:
- mutating webhook
  Replace the invalid values on APIs and other interfaces with valid ones.
- upgrade path
  The purpose of fix up functions are fixing the invalid fields in old CRs. The strategy is listing the CRs and then update the them. Before updating, the mutating webhook will check the invalid values and fix them.

### DIFF
--- a/deploy/install/01-prerequisite/03-crd.yaml
+++ b/deploy/install/01-prerequisite/03-crd.yaml
@@ -1445,12 +1445,14 @@ spec:
                     children:
                       additionalProperties:
                         type: boolean
+                      nullable: true
                       type: object
                     created:
                       type: string
                     labels:
                       additionalProperties:
                         type: string
+                      nullable: true
                       type: object
                     name:
                       type: string
--- a/k8s/pkg/apis/longhorn/v1beta2/engine.go
+++ b/k8s/pkg/apis/longhorn/v1beta2/engine.go
@@ -89,6 +89,7 @@ type Snapshot struct {
 	// +optional
 	Parent string `json:"parent"`
 	// +optional
+	// +nullable
 	Children map[string]bool `json:"children"`
 	// +optional
 	Removed bool `json:"removed"`
@@ -99,6 +100,7 @@ type Snapshot struct {
 	// +optional
 	Size string `json:"size"`
 	// +optional
+	// +nullable
 	Labels map[string]string `json:"labels"`
 }
 
--- a/upgrade/v1beta1/upgrade.go
+++ b/upgrade/v1beta1/upgrade.go
@@ -39,6 +39,9 @@ func FixupCRs(config *restclient.Config, namespace string, lhClient *lhclientset
 	if err := fixupNodes(namespace, lhClient); err != nil {
 		return err
 	}
+	if err := fixupEngines(namespace, lhClient); err != nil {
+		return err
+	}
 	if err := fixupRecurringJobs(namespace, lhClient); err != nil {
 		return err
 	}
@@ -158,6 +161,31 @@ func fixupNodes(namespace string, lhClient *lhclientset.Clientset) error {
 			continue
 		}
 		if _, err = lhClient.LonghornV1beta1().Nodes(namespace).Update(context.TODO(), &obj, metav1.UpdateOptions{}); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func fixupEngines(namespace string, lhClient *lhclientset.Clientset) error {
+	engines, err := lhClient.LonghornV1beta1().Engines(namespace).List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		return err
+	}
+
+	for _, obj := range engines.Items {
+		existing := obj.DeepCopy()
+
+		if obj.Spec.ReplicaAddressMap == nil {
+			obj.Spec.ReplicaAddressMap = make(map[string]string, 0)
+		}
+		if obj.Spec.UpgradedReplicaAddressMap == nil {
+			obj.Spec.UpgradedReplicaAddressMap = make(map[string]string, 0)
+		}
+		if reflect.DeepEqual(obj, existing) {
+			continue
+		}
+		if _, err = lhClient.LonghornV1beta1().Engines(namespace).Update(context.TODO(), &obj, metav1.UpdateOptions{}); err != nil {
 			return err
 		}
 	}
--- a/webhook/resources/backingimage/mutator.go
+++ b/webhook/resources/backingimage/mutator.go
@@ -1,0 +1,56 @@
+package backingimage
+
+import (
+	admissionregv1 "k8s.io/api/admissionregistration/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+
+	"github.com/longhorn/longhorn-manager/datastore"
+	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
+	"github.com/longhorn/longhorn-manager/webhook/admission"
+)
+
+type backingImageMutator struct {
+	admission.DefaultMutator
+	ds *datastore.DataStore
+}
+
+func NewMutator(ds *datastore.DataStore) admission.Mutator {
+	return &backingImageMutator{ds: ds}
+}
+
+func (b *backingImageMutator) Resource() admission.Resource {
+	return admission.Resource{
+		Name:       "backingimages",
+		Scope:      admissionregv1.NamespacedScope,
+		APIGroup:   longhorn.SchemeGroupVersion.Group,
+		APIVersion: longhorn.SchemeGroupVersion.Version,
+		ObjectType: &longhorn.BackingImage{},
+		OperationTypes: []admissionregv1.OperationType{
+			admissionregv1.Create,
+			admissionregv1.Update,
+		},
+	}
+}
+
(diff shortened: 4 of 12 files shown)


innobead (dismissed):
LGTM. We need to create another ticket to revisit these nullable properties in 1.4.0 for strict validation. 

Also, need to have a PR to update longhorn/longhorn charts after this merged.

jenting:
I'm fine with this approach.
However, we had a PR to upgrade the old CRs to initiate these fields.
Should we reconsider to use this approach and remove the upgrade code?

innobead:
> 

Probably we should keep both for 1.3.0. One is for the upgrade to make things right, but this PR is actually for the fresh install for backward compatibility at the moment. WDYT?

jenting:
I'm fine. I just think that the mutating admission webhook could achieve this as well, i.e., initiate the default value if it's null.

innobead:
> I'm fine. I just think that the mutating admission webhook could achieve this as well, i.e., initiate the default value if it's null.

It seems reasonable for me by using mutation webhook. @derekbit WDYT?

derekbit:
Yes, it's reasonable. I'll use mutation webhook and to see if there is any prob.

shuo-wu (commented):
Actually, I am confused about the recent CRD-related PRs, especially for the validation part.
1. Is the validation handled by k8s rather than the webhook we introduced?
2. Will there be similar issues for the old (v1beta1) CRs? 
3. Jenting mentioned one similar issue in https://github.com/longhorn/longhorn-manager/pull/1189#issuecomment-1011542445. Can this PR fix the issue he mentioned?

I am thinking if you can briefly summarize the materials/links/docs you referred for the CRD API version & webhook PRs. I am afraid of missing any conversions or principles for this part.

derekbit:
> Actually, I am confused about the recent CRD-related PRs, especially for the validation part.
> 
> 1. Is the validation handled by k8s rather than the webhook we introduced?
> 2. Will there be similar issues for the old (v1beta1) CRs?
> 3. Jenting mentioned one similar issue in [#1189 (comment)]
(https://github.com/longhorn/longhorn-manager/pull/1189#issuecomment-1011542445). Can this PR fix the issue he mentioned?

Yes. The validation is handled by k8s instead of the webhook. Because v1beta1 CRs is still schema-less, we won't encounter this issue.

The APIs do not validate the parameters now. For example, the `node.disks.tags` should not be `null`, but ui or user may call the APIs with `null` values. Thus, the PR aims to the backward compatibility instead of fixing something.

> 3. Jenting mentioned one similar issue in [#1189 (comment)]
(https://github.com/longhorn/longhorn-manager/pull/1189#issuecomment-1011542445). Can this PR fix the issue he mentioned?

Right. `nullable` also fixes the the issue @jenting mentioned. I cannot make sure if `null` values is good, so I still keep the `fixUpxxxx`.

> 
> I am thinking if you can briefly summarize the materials/links/docs you referred for the CRD API version & webhook PRs. I am afraid of missing any conversions or principles for this part. If this is time consuming, you can ignore it.

- main context: https://github.com/longhorn/longhorn/issues/3265. 
- Conversion webhook document: https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/
- Backward compatibility concern: https://github.com/longhorn/longhorn/issues/3538

derekbit:
> It seems reasonable for me by using mutation webhook. @derekbit WDYT?

@jenting @innobead
Looks we need another ticket for this, right?
It seems not in the scope of this issue and PR.

innobead:
@derekbit actually I would prefer fixing in this PR or you still need to refactor it later when moving to the mutation webhooks. Right?

derekbit:
Sure. I can fix in this ticket. Just a little confused about the purpose of the PR's scope.

jenting:
I'm okay to use this PR/issue.

From my perspective, I think the considering sequence to handles the CRs are:
1. Try to fix it in the CRD structural schema _or_ fix it in the webhook server. Such as null field, we could add nullable in theCRD structural schema _or_ use mutating admission webhook to initiate the field which is null.
2. If the above not possible, fix it in the longhorn manager.

derekbit:
@jenting @innobead 
We still need the `fixUpxxx` in the upgrade path.

While upgrading the longhorn-manager, longhorn-manager calls `UpdateStatus` to update `nodes`' conditions.

The `UpdateStatus` complains
`err=Node.longhorn.io \"rancher50-master\" is invalid: spec.disks.default-disk-745ff2be8d312b52.tags: Invalid value: \"null\": spec.disks.default-disk-745ff2be8d312b52.tags in body must be of type array: \"null\""`,
because the mutating webhook does not handle `UpdateStatus` and only handle `Update`, `Create`, `Delete` and `Connect`. 

The `fixUpxxx` in the update path can be simpler than before. Just `List` the old objects and  the `Update`, because the mutating webhook can help update the `null` values.

innobead:
> @jenting @innobead We still need the `fixUpxxx` in the upgrade path.
> 
> While upgrading the longhorn-manager, node-controller calls `UpdateStatus` to update `nodes`' conditions.
> 
> The `UpdateStatus` complains `err=Node.longhorn.io \"rancher50-master\" is invalid: spec.disks.default-disk-745ff2be8d312b52.tags: Invalid value: \"null\": spec.disks.default-disk-745ff2be8d312b52.tags in body must be of type array: \"null\""`, because the mutating webhook does not handle `UpdateStatus` and only handle `Update`, `Create`, `Delete` and `Connect`.
> 
> The `fixUpxxx` in the update path can be simpler than before. Just `List` the old objects and then call `Update`, because the mutating webhook can help fix the `null` values.

@derekbit will try to add subresources to the mutation config but update later.

derekbit:
I added the subresource `node/status` in the mutating webhook. Then, modified the node `Spec` in the node mutating logic while handling `UpdateStatus`, the error still happened.  The modify of the `Spec` is not accepted by Kubernetes.

```
curl -H "Authorization: Bearer  $<TOKEN>"  https://localhost:6443/apis/longhorn.io/v1beta2/namespaces/longhorn-system/nodes/rancher50-master/status -d '{"status": {"region": "Hello"}}' -H 'Content-Type: application/merge-patch+json' -X PATCH -k

{
  "kind": "Status",
  "apiVersion": "v1",
  "metadata": {

  },
  "status": "Failure",
  "message": "Node.longhorn.io \"rancher50-master\" is invalid: spec.disks.default-disk-745ff2be8d312b52.tags: Invalid value: \"null\": spec.disks.default-disk-745ff2be8d312b52.tags in body must be of type array: \"null\"",
  "reason": "Invalid",
  "details": {
    "name": "rancher50-master",
    "group": "longhorn.io",
    "kind": "Node",
    "causes": [
      {
        "reason": "FieldValueInvalid",
        "message": "Invalid value: \"null\": spec.disks.default-disk-745ff2be8d312b52.tags in body must be of type array: \"null\"",
        "field": "spec.disks.default-disk-745ff2be8d312b52.tags"
      }
    ]
  },
  "code": 422
```
In summary, `Spec` and `Status` cannot be modified simultaneously.

Thus, we still need the `fixUpxxx` logic to fix the old CRs. The purpose of the mutating webhook is to fix the `null` values from different interfaces.

derekbit:
In this commit, the fixing up/mutating invalid values logics (`fixUpxxx`) are moved to the mutating webhook.
Thus, the `//+nullable` in CRD' `Spec` can be removed.

The `fixUpxxx` functions cannot be removed, because the node controller updates the CR.status.conditions after upgrading the longhorn-manager. The `UpdateStatus` fails if CR' `Spec` has invalid values.

But we can simplify the `fixUpxxx` functions. Just read the old CRs and update them. The mutating webhook will fix up the invalid values.

cc. @innobead, @jenting, @shuo-wu

shuo-wu (commented):
I still prefer to retain the field checks in v1beta1 upgrade/fixup functions. Otherwise, if one manager pod fails at one CR update, the next manager pod needs to walk through all of the CRs one more time. e.g., there are 100 existing/old CRs in total, the 1st manager pod fails at updating the 99th CRs, then the pod will be restarted and re-update all of the 100 CRs blindly.

Actually, the checks in the v1beta1 upgrade/fixup functions are not mutexed with the checks in the mutation hooks. IMO, the checks in the mutation hooks are more like general steps that contain not only the v1beta1 related fields but also some other nullable fields.

derekbit:
> I still prefer to retain the field checks in v1beta1 upgrade/fixup functions. Otherwise, if one manager pod fails at one CR update, the next manager pod needs to walk through all of the CRs one more time. e.g., there are 100 existing/old CRs in total, the 1st manager pod fails at updating the 99th CRs, then the pod will be restarted and re-update all of the 100 CRs blindly.

Yes, your concern is valid. I will add it back.

> 
> Actually, the checks in the v1beta1 upgrade/fixup functions are not mutexed with the checks in the mutation hooks. IMO, the checks in the mutation hooks are more like general steps that contain not only the v1beta1 related fields but also some other nullable fields.

Correct. The main purpose of the checks in the mutation hooks aims to fix the invalid values from different interfaces.

shuo-wu (approved):

innobead (approved):
LGTM

@derekbit 👍

innobead:
@derekbit Please also update longhorn/longhorn manfiests.

PR #1205, opened 2022-01-19
Improve longhorn manager upgrade and starting

shuo-wu:
longhorn/longhorn#3087


innobead (dismissed):
LGTM

types/types.go
```diff
@@ -32,6 +32,8 @@ const (
 	CRDAPIVersionV1beta1  = "longhorn.io/v1beta1"
 	CRDAPIVersionV1beta2  = "longhorn.io/v1beta2"
 	CurrentCRDAPIVersion  = CRDAPIVersionV1beta2
+
+	CurrentLonghornVersion = "v1.3.0"
```
derekbit, line 36: Is it possible to read the version value in file `longhorn-manager/version` and then set to `CurrentLonghornVersion` during compilation? Then, we don't have to maintain two `version` values.

derekbit:
In general, LGTM.

types/types.go
shuo-wu: Do you mean adding an ENV in the YAML file `deploy/install/02-components/01-manager.yaml`?
BTW, what is another version value? the image tag?

derekbit: I'm just thinking if we can reuse the value in https://github.com/longhorn/longhorn-manager/blob/master/version .

shuo-wu: Oh, I never realize there is such a file in repo `longhorn-manager`... Where do we use it?

derekbit: Hmm, I have no idea.
@innobead Do you know the usage of the `version` file?

yasker: https://github.com/longhorn/longhorn-manager/blob/c09016aafa9e7f582ed2f7ecafa2d823795cd353/scripts/build#L9

upgrade/upgrade.go
```diff
@@ -217,69 +220,64 @@ func upgradeLocalNode() (err error) {
...
-	if err := v110to111.UpgradeCRs(namespace, lhClient); err != nil {
+
+	lhVersionBeforeUpgrade, err := upgradeutil.GetCurrentLonghornVersion(namespace, lhClient)
+	if err != nil {
 		return err
 	}
-	if err := v110to120.UpgradeCRs(namespace, lhClient, kubeClient); err != nil {
-		return err
+
+	if semver.Compare(lhVersionBeforeUpgrade, "v0.8.0") < 0 {
```
joshimoo, line 233: What's the result if `lhVersionBeforeUpgrade` is empty?

types/setting.go
```diff
@@ -89,6 +90,7 @@ var (
 		SettingNameStorageOverProvisioningPercentage,
 		SettingNameStorageMinimalAvailablePercentage,
 		SettingNameUpgradeChecker,
+		SettingNameCurrentLonghornVersion,
```
joshimoo, line 93: NIT: perhaps we should consider having a meta/version resource, using settings for a const is sort of a hack.
But it's okay for now, since it allows us to reuse the existing implementations for settings (crd,validation,ui,etc)

types/types.go
joshimoo: using a meta package like in the engine is a good approach :)

upgrade/upgrade.go
shuo-wu: If the 1st input is invalid or empty, the result will be `-1` :
https://go.dev/play/p/Xr2fwc-t4Nv

upgrade/upgrade.go
```diff
@@ -82,6 +83,17 @@ func upgrade(currentNodeID, namespace string, config *restclient.Config, lhClien
 	var err error
 	defer cancel()
 
+	// If the current Longhorn is already the latest version,
+	// the leader election & the whole upgrade path could be skipped.
+	lhVersionBeforeUpgrade, err := upgradeutil.GetCurrentLonghornVersion(namespace, lhClient)
+	if err != nil {
+		return err
+	}
+	if semver.Compare(lhVersionBeforeUpgrade, types.CurrentLonghornVersion) >= 0 {
```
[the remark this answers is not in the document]
shuo-wu: If the 1st input is invalid or empty, the result will be `-1` :
https://go.dev/play/p/Xr2fwc-t4Nv

Yes, sometimes this would avoid meaningless leader election. But once there are multiple longhorn manager pods starting simultaneously, all of these pods will run this check then trap into the leader election. In order to avoid duplicate upgrades in this case, I add the version comparison in function `doResourceUpgrade` as well.

types/types.go
shuo-wu: The version value used in [the build script you mentioned](https://github.com/longhorn/longhorn-manager/blob/c09016aafa9e7f582ed2f7ecafa2d823795cd353/scripts/build#L9) is from the [version script](https://github.com/longhorn/longhorn-manager/blob/840ae441597ac6af0904946db7d40916d959f9f8/scripts/version#L11-L150, which is actually using git commit or tag rather than the file we discussed. Should we improve this part? And how? Since right now once there is no git tag when building one longhorn manager image, the version number will be the commit ID. Should we differentiate the git commit ID from the version number and always use the value in https://github.com/longhorn/longhorn-manager/blob/master/version for this upgrade path refactoring?

derekbit (approved):
LGTM

shuo-wu:
@PhanLe1010 @joshimoo

upgrade/upgrade.go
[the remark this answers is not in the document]
PhanLe1010: There is a small problem. When both 1st input and 2nd input are invalid, the result is always 0. This mean we won't upgrade when deploying a new dev version
https://go.dev/play/p/0xvHZKt1Xg2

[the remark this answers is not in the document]
shuo-wu: Probably. Hence I already add one more check `meta.Version != ""` in the updated commit. PTAL.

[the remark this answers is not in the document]
PhanLe1010: I was referring to the case when `meta.Version` is `abc` and `lhVersionBeforeUpgrade` is `xyz`. In this case, will we always skip upgrading? If yes, it might be a problem for the developers. 

An other case is when both `meta.Version` and `lhVersionBeforeUpgrade` are `dev-xxx`. In this case, we will skip upgrading. It is a bit inconvenient for the developer.

May be we can do: if the version is dev version, then we always upgrade?  This created another problem too: QA cannot test this PR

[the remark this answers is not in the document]
shuo-wu: I just updated the PR. Right now the upgrade will be skipped only when both `meta.Version` and `lhVersionBeforeUpgrade` are valid and match the case.
As for the test, I will help generate multiple images for the testing.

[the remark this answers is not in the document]
shuo-wu: Besides, I have refactored the versioning part. Now the source of the version is determined by file `version` in the root directory.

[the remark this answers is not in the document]
PhanLe1010: Thank you! Great approach

PhanLe1010 (approved):
LGTM

derekbit (approved):
LGTM

jenting (approved):
LGTM

innobead (approved):
version file is actually only updated when doing a release. However, the change here still looks fine for me, because the following updating during release will not cause any problem as per your semver check.

PR #1206, opened 2022-01-24
webhook: separate conversion and admission webhook servers

derekbit:
1. The conversion webhook should be started before the admission webhook, because the operations (list and get) of longhorn resources in the admission webhook replies on the CRD conversion. We separate one webhook server to one conversion webhook and one admission webhook server to resolve the issue.

2. For admission webhook, sync datastore during initialization.

longhorn/longhorn#3548
https://github.com/longhorn/longhorn/issues/3571


app/webhook.go
```diff
@@ -59,12 +59,15 @@ func runWebhookServer(c *cli.Context) error {
 		return errors.Wrap(err, "unable to get client config")
 	}
 
-	s := server.New(context.Background(), cfg, namespace)
+	done := make(chan struct{})
```
joshimoo, line 62: NIT: Generally I prefer using a cancellable context for example, this way you can pass down the context.
And you can wait on the ctx.done channel :)

webhook/client/client.go
```diff
@@ -23,7 +23,7 @@ type Client struct {
 	Datastore *datastore.DataStore
 }
 
-func New(ctx context.Context, config *rest.Config, namespace string) (*Client, error) {
+func New(ctx context.Context, stopCh chan struct{}, config *rest.Config, namespace string) (*Client, error) {
```
joshimoo, line 26: NIT: If you use the cancellable context you also don't need to pass the extra stop channel

joshimoo (commented):
Left some nit comments, I don't really know the big picture on this one, so I will wait for the approval from @jenting

derekbit:
Upgrade from `v1.2.3 `to the `patched branch`, will encounter the chicken-egg problem because datastore use `v1beta2` APIs.
```
	kubeInformerFactory := informers.NewSharedInformerFactory(kubeClient, time.Second*30)
	lhInformerFactory := lhinformers.NewSharedInformerFactory(lhClient, time.Second*30)
	datastore := datastore.NewDataStore(lhInformerFactory, lhClient, kubeInformerFactory, kubeClient, namespace)

	go lhInformerFactory.Start(ctx.Done())
	go kubeInformerFactory.Start(ctx.Done())

	if !datastore.Sync(ctx.Done()) {
		return nil, errors.Wrap(err, "datastore cache sync up failed")
	}
```

Error
```
root@rancher50-master:~/deploy/install# kl logs longhorn-webhook-55fbcb67c4-5qzlj
2022/01/25 07:21:24 proto: duplicate proto type registered: VersionResponse
time="2022-01-25T07:21:24Z" level=info msg="Starting longhorn webhook server"
W0125 07:21:24.939930       1 client_config.go:552] Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.
I0125 07:21:24.940878       1 shared_informer.go:223] Waiting for caches to sync for longhorn datastore
E0125 07:21:24.961274       1 reflector.go:178] github.com/longhorn/longhorn-manager/k8s/pkg/client/informers/externalversions/factory.go:117: Failed to list *v1beta2.Node: v1beta2.NodeList.Items: []v1beta2.Node: v1beta2.Node.Status: v1beta2.NodeStatus.Conditions: []v1beta2.Condition: decode slice: expect [ or n, but found {, error found in #10 byte of ...|ditions":{},"diskSta|..., bigger context ...|rCPURequest":0,"tags":[]},"status":{"conditions":{},"diskStatus":{"default-disk-745ff2be8d312b52":{"|...
E0125 07:21:24.961905       1 reflector.go:178] github.com/longhorn/longhorn-manager/k8s/pkg/client/informers/externalversions/factory.go:117: Failed to list *v1beta2.EngineImage: v1beta2.EngineImageList.Items: []v1beta2.EngineImage: v1beta2.EngineImage.Status: v1beta2.EngineImageStatus.Conditions: []v1beta2.Condition: decode slice: expect [ or n, but found {, error found in #10 byte of ...|ditions":{},"control|..., bigger context ...|iAPIMinVersion":3,"cliAPIVersion":5,"conditions":{},"controll
```

Moving the validation functions to webhook seems better.

webhook/client/client.go
```diff
@@ -33,22 +35,36 @@ func New(ctx context.Context, config *rest.Config, namespace string) (*Client, e
...
+
+		if !datastore.Sync(ctx.Done()) {
+			return nil, fmt.Errorf("datastore cache sync up failed")
+		}
+
+		return &Client{
+			Clients:   *clients,
+			Datastore: datastore,
+		}, nil
+	}
```
shuo-wu, line 66: NIT: Do we still need to check if `serverType == types.WebhookTypeConversion`?

webhook/client/client.go
```diff
@@ -16,14 +17,15 @@ import (
...
+	"github.com/longhorn/longhorn-manager/types"
 )
 
 type Client struct {
 	clients.Clients
 	Datastore *datastore.DataStore
 }
 
-func New(ctx context.Context, config *rest.Config, namespace string) (*Client, error) {
+func New(ctx context.Context, config *rest.Config, namespace, serverType string) (*Client, error) {
```
shuo-wu, line 28: NIT:
I just realize that this package name is a little bit confusing. It's more than the Longhorn client used in the webhook service rather than a webhook client. But we will use it to start the webhook service in `xxxxxxxWebhookListenAndServe` as well...

webhook/client/client.go
```diff
@@ -33,22 +35,36 @@ func New(ctx context.Context, config *rest.Config, namespace string) (*Client, e
...
+
+		if !datastore.Sync(ctx.Done()) {
+			return nil, fmt.Errorf("datastore cache sync up failed")
+		}
+
+		return &Client{
+			Clients:   *clients,
+			Datastore: datastore,
+		}, nil
+	}
```
shuo-wu, line 65:
```suggestion
var ds *datastore.DataStore
if serverType == types.WebhookTypeAdmission {
    // build ds
}
return &Client{
	Clients:   *clients,
	Datastore: datastore,
}, nil


```

shuo-wu (commented):
In general LGTM

webhook/client/client.go
derekbit: Yes, seem better.

app/webhook.go
```diff
@@ -17,29 +16,39 @@ import (
...
+				logrus.Fatalf("Error starting longhorn %v webhook server: %v", webhookType, err)
 			}
 		},
 	}
 }
 
-func runWebhookServer(c *cli.Context) error {
-	logrus.Info("Starting longhorn webhook server")
+func runWebhookServer(c *cli.Context, serverType string) error {
+	logrus.Info("Starting longhorn admission webhook server")
```
jenting, line 49:
```suggestion
        logrus.Infof("Starting longhorn %s webhook server", webhookType)
```

types/types.go
```diff
@@ -35,8 +35,12 @@ const (
 )
 
 const (
-	DefaultAPIPort           = 9500
-	DefaultWebhookServerPort = 9443
+	DefaultAPIPort                     = 9500
+	DefaultConversionWebhookServerPort = 9443
+	DefaultAdmissionWebhookServerPort  = 9444
```
jenting, line 40: Can we use the same port `9443`?

jenting (commented):
Remember to update the network policy examples as well.
https://github.com/longhorn/longhorn-manager/blob/master/examples/network-policy/webhook-network-policy.yaml

innobead (approved):
LGTM.

derekbit:
> LGTM, but remember you probably want to update the resync period which seems needless.

Yes. I will have the resync period adjusted in https://github.com/longhorn/longhorn/issues/3562 .

> 
> Also, we need to update the longhorn/longhorn 1.3.0 manifest later after this PR merged.

Okay.

jenting (approved):
LGTM. Thank you for the effort.

PR #1209, opened 2022-01-27
environment_check: check per node packages

derekbit:
Longhorn 3550

### DIFF
--- a/deploy/scripts/environment_check.sh
+++ b/deploy/scripts/environment_check.sh
@@ -1,18 +1,97 @@
 #!/bin/bash
 
-dependencies() {
+export RED='\x1b[0;31m'
+export GREEN='\x1b[38;5;22m'
+export CYAN='\x1b[36m'
+export YELLOW='\x1b[33m'
+export NO_COLOR='\x1b[0m'
+
+if [ -z "${LOG_TITLE}" ]; then
+  LOG_TITLE=''
+fi
+if [ -z "${LOG_LEVEL}" ]; then
+  LOG_LEVEL="INFO"
+fi
+
+debug() {
+  if [[ "${LOG_LEVEL}" == "DEBUG" ]]; then
+    local log_title
+    if [ -n "${LOG_TITLE}" ]; then
+     log_title="(${LOG_TITLE})"
+    else
+     log_title=''
+    fi
+    echo -e "${GREEN}[DEBUG]${log_title} ${NO_COLOR}$1"
+  fi
+}
+
+info() {
+  if [[ "${LOG_LEVEL}" == "DEBUG" ]] ||\
+     [[ "${LOG_LEVEL}" == "INFO" ]]; then
+    local log_title
+    if [ -n "${LOG_TITLE}" ]; then
+     log_title="(${LOG_TITLE})"
+    else
+     log_title=''
+    fi
+    echo -e "${CYAN}[INFO] ${log_title} ${NO_COLOR}$1"
+  fi
+}
+
+warn() {
+  if [[ "${LOG_LEVEL}" == "DEBUG" ]] ||\
+     [[ "${LOG_LEVEL}" == "INFO" ]] ||\
+     [[ "${LOG_LEVEL}" == "WARN" ]]; then
+    local log_title
+    if [ -n "${LOG_TITLE}" ]; then
+     log_title="(${LOG_TITLE})"
+    else
+     log_title=''
+    fi
+    echo -e "${YELLOW}[WARN] ${log_title} ${NO_COLOR}$1"
+  fi
+}
+
+error() {
+  if [[ "${LOG_LEVEL}" == "DEBUG" ]] ||\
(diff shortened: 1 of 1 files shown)


deploy/scripts/environment_check.sh
```diff
@@ -100,10 +114,54 @@ validate_ds() {
...
+  local packages=($@)
+  local pods=$(kubectl get pods -o name | grep longhorn-environment-check)
+
+  local allFound=true
+
+  for ((i=0; i<${#packages[@]}; i++)); do
+    local package=${packages[$i]}
+
+    for pod in ${pods}; do
+      installed=`kubectl exec -it $pod -- nsenter --mount=/proc/1/ns/mnt -- bash -c "dpkg --get-selections ${package}" | grep -c 'install'`
```
innobead, line 127: This should also work for all distros we support in this script.

derekbit: Oops, sorry, forgot to handle `dpkg ...`

deploy/scripts/environment_check.sh
```diff
@@ -65,45 +144,105 @@ wait_ds_ready() {
...
+  CHECK_CMD="rpm -q"
+  packages=(nfs-utils iscsi-initiator-utils)
+  ;;
+*"suse"* )
+  CHECK_CMD="zypper se -i"
+  packages=(nfs-client open-iscsi)
+  ;;
+*)
+  CHECK_CMD=""
+  packages=()
```
innobead, line 230: If not supported OS, suggest to error out and exit.

innobead (commented):
LGTM, but it seems we don't check if multipass is running, do we need?

deploy/scripts/environment_check.sh
derekbit: Okay.

derekbit:
> LGTM, but it seems we don't check if multipass is running, do we need?

I'm thinking how to check multipath on the node. Maybe just check the `multipath.conf` has correct blacklist?
https://longhorn.io/kb/troubleshooting-volume-with-multipath/

cc. @PhanLe1010

innobead:
> > LGTM, but it seems we don't check if multipass is running, do we need?
> 
> I'm thinking how to check multipath on the node. Maybe just check the `multipath.conf` has a correct blacklist? [longhorn.io/kb/troubleshooting-volume-with-multipath](https://longhorn.io/kb/troubleshooting-volume-with-multipath/)
> 
> cc. @PhanLe1010

or just check if the multipass is running, show a warning to inform users there will be probably some problem about volumes automatically mounted? I don't think we need to specially identify the problem. WDYT?

deploy/scripts/environment_check.sh
derekbit: > > > LGTM, but it seems we don't check if multipass is running, do we need?
> > 
> > 
> > I'm thinking how to check multipath on the node. Maybe just check the `multipath.conf` has a correct blacklist? [longhorn.io/kb/troubleshooting-volume-with-multipath](https://longhorn.io/kb/troubleshooting-volume-with-multipath/)
> > cc. @PhanLe1010
> 
> or just check if the multipass is running, show a warning to inform users there will be probably some problem about volumes automatically mounted? I don't think we need to specially identify the problem. WDYT?

Good idea!

PhanLe1010:
Yeah, agree with @innobead that we can just check if multipath is running and show a warning message with the link to the GitHub issue

PhanLe1010:
I am wondering if it is possible that the host has the open-iscsi package but the iscsid is not running?

derekbit:
> I am wondering if it is possible that the host has the open-iscsi package but the iscsid is not running?

Yes, I think we need this. I just remember there is an issue that the iscsid was not started after installing the iscsi package from an user.

PhanLe1010:
Thanks @derekbit Can you help to add the check for iscsid?

derekbit:
> Thanks @derekbit Can you help to add the check for iscsid?

Sure thing. I'm working on it.

innobead (changes requested):
LGTM, just minor suggestions

PhanLe1010 (dismissed):
LGTM

innobead (approved):

jenting:
@derekbit 
Looks like we need to update the https://github.com/longhorn/longhorn/blob/master/scripts/environment_check.sh as well.

innobead:
> @derekbit Looks like we need to update the [https://github.com/longhorn/longhorn/blob/master/scripts/environment_check.sh](https://github.com/longhorn/longhorn/blob/master/scripts/environment_check.sh?rgh-link-date=2022-02-10T06%3A33%3A22Z) as well.

YES, but I am planning to keep only one place. Let me handle that.

jenting:
Besides, we could consider improving the doc output https://longhorn.io/docs/1.2.3/deploy/install/#using-the-environment-check-script of the 1.3.0 version.

PR #1210, opened 2022-02-03
Fix scalability issue and expose controller workqueue prometheus metrics

PhanLe1010:
1. Fix scalability issue by: 
    * Increase the resync for all shared informers so that we don't overwhelm the controller workqueues. One exception is the node shared informer, we keep resync to be 30s to check the real time usage of disk on nodes
    * When adding a fresh event to wokqueue, we should use Add() instead of AddRateLimited(). AddRateLimited() should only be used when a controller encounters error while processing an event and the controller want to add the event back to its workqueue for later retry. See example at https://github.com/kubernetes-csi/external-attacher/blob/master/pkg/controller/controller.go
1. Expose controller workqueue metrics

longhorn/longhorn#3083


PhanLe1010:
Note from @joshimoo : Since we are increasing the resync, before we push release that though we need to do a review cycle of all resource dependencies to ensure we have the appropriate event handlers in each controller. I remember previously that some event handlers where missing for certain resource + controller combinations.

joshimoo:
knowledge sharing there is some good conversation/analysis about this issue in the slack / confluence.
https://confluence.suse.com/display/LON/Scalability+issue+in+Longhorn+manager
https://suse.slack.com/archives/C02DR3N5T24/p1643931813495509?thread_ts=1643924747.610999&cid=C02DR3N5T24

PhanLe1010:
Because we are planning to increase the resync for all shared informers, we need to do a review cycle of all resource dependencies to ensure we have the appropriate event handlers in each controller. To evaluate whether the controllers have the appropriate event handlers, one can start by drawing up a dependency graph on the resources. The attached image is the graph that I have drawn. 
longhorn_dependency_graph



Action item: Can the team help to review this graph to see if there is any missing dependency on the controllers that you have worked with? @joshimoo @shuo-wu @jenting @c3y1huang @derekbit

controller/controller_manager.go
```diff
@@ -67,8 +67,8 @@ func StartControllers(logger logrus.FieldLogger, stopCh chan struct{}, controlle
 	//  lead to scalability problems, since we dump the whole cache of each object back in to the reconciler every 30 seconds.
 	//  if a specific controller requires a periodic resync, one enable it only for that informer, add a resync to the event handler, go routine, etc.
 	//  some refs to look at: https://github.com/kubernetes-sigs/controller-runtime/issues/521
-	kubeInformerFactory := informers.NewSharedInformerFactory(kubeClient, time.Second*30)
-	lhInformerFactory := lhinformers.NewSharedInformerFactory(lhClient, time.Second*30)
+	kubeInformerFactory := informers.NewSharedInformerFactory(kubeClient, time.Hour)
+	lhInformerFactory := lhinformers.NewSharedInformerFactory(lhClient, time.Hour)
```
innobead, line 71: I seem to remember backup related setting also rely on this resync to ensure the settings take effect. Not sure if this change will impact that, or it's not a case anymore?

innobead (approved):
LGTM, but need other reviews as well, because this will probably impact some existing implementations which potentially rely on the original 30s resync.

monitoring/workqueue/metrics.go
```diff
@@ -0,0 +1,123 @@
...
+		Subsystem: WorkQueueSubsystem,
+		Name:      AddsKey,
+		Help:      "Total number of adds handled by workqueue",
+	}, []string{"name"})
+
+	latency = prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Namespace: LonghornName,
+		Subsystem: WorkQueueSubsystem,
+		Name:      QueueLatencyKey,
+		Help:      "How long in seconds an item stays in workqueue before being requested.",
```
shuo-wu, line 45: Do you mean `being reconciled` or `being requested`?

shuo-wu (commented):
Do we need to replace the enqueue functions like [this](https://github.com/longhorn/longhorn-manager/blob/840ae441597ac6af0904946db7d40916d959f9f8/controller/volume_controller.go#L401) with rate-limited requeue functions?

controller/controller_manager.go
shuo-wu: No. Right now there will be a separate goroutine periodically update `backupTarget.spec.syncRequestedAt` then trigger backup & backupVolume CR update. But let's wait for Jenting's response as confirmation.
@jenting Can you help verify this won't affect backup-related scenarios like BackupVolume creation?

monitoring/monitoring.go
```diff
@@ -1,7 +1,7 @@
 package monitoring
 
 import (
-	"net/http"
+	"github.com/longhorn/longhorn-manager/monitoring/registry"
```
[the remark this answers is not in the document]
PhanLe1010: Thank you. Fixed

monitoring/workqueue/metrics.go
PhanLe1010: I meant `being requested`. I copied those definitions from kubernetes package :D

controller/controller_manager.go
jenting: I remember no, the backup-related scenarios would not depend on the resync time.
As Shuo's comment, the backup-related depends on the `spec.syncRequestedAt` field.

jenting: https://github.com/longhorn/longhorn-manager/blob/840ae441597ac6af0904946db7d40916d959f9f8/webhook/client/client.go#L46-L47

Please check if we need to change the resync time of the webhook server as well.
cc @derekbit

monitoring/workqueue/metrics.go
```diff
@@ -0,0 +1,123 @@
+package workqueue
+
+import (
+	"k8s.io/client-go/util/workqueue"
+
+	"github.com/longhorn/longhorn-manager/monitoring/registry"
+	"github.com/prometheus/client_golang/prometheus"
```
jenting, line 7: nit, add space between these two packages to separate the internal/external imported packages.

controller/controller_manager.go
derekbit: Is there any potential risk after increasing resync period to 1 hour?

monitoring/workqueue/metrics.go
shuo-wu: IMO, this latency means the interval from the item being enqueued to the item being poped up (for reconciliation), right?

shuo-wu:
> Actually, why do we need to handle the API error and enqueing separately for that case? Can we just error out the let the caller (`processNextWorkItem()`) requeue the item with rate limited?

Hmmm, I have not considered this issue in deep. It seems that somehow the controller does not want to return an error in this case... But I prefer to return errors then rely on `processNextWorkItem` processing.

controller/controller_manager.go
shuo-wu: @derekbit For example, if we increase this interval to 1 hour for the node controller, then Longhorn cannot detect the disk failure (via checking the disk meta file in the reconcile loop) in time. That's why there is an exception. 

Here, we want to verify if there are more cases like the node controller.

derekbit: @jenting I think the resync time of the webhook server can also increase to 1 hour, because it only reads resources such as settings from the datastore.

monitoring/workqueue/metrics.go
PhanLe1010: > IMO, this latency means the interval from the item being enqueued to the item being poped up 
Yes, I think so too. From the queue package's perspective, the item is poped up. The queue doesn't know what the consumer is doing with it (reconciliation/process/etc...). So I think the queue just knows that the item is requested.

PhanLe1010:
> Hmmm, I have not considered this issue in deep. It seems that somehow the controller does not want to return an error in this case... But I prefer to return errors then rely on processNextWorkItem processing.

@joshimoo, @yasker  Do you see any problem if we just return errors then rely on processNextWorkItem processing?

monitoring/workqueue/metrics.go
PhanLe1010: Thank you. Fixed

controller/controller_manager.go
jenting: I'd like no resync if possible.

> But, I'm not sure if there are other parts in webhooks requiring short resync period, so I will re-check the resyncPeriod after finishing mutating and validating webhooks.

Thank you.

jenting (approved):
LGTM

shuo-wu (commented):
LGTM except for the conflict handling. Let's wait for others' feedbacks.

derekbit (approved):
LGTM

innobead (approved):

innobead:
> > Hmmm, I have not considered this issue in deep. It seems that somehow the controller does not want to return an error in this case... But I prefer to return errors then rely on processNextWorkItem processing.
> 
> @joshimoo, @yasker Do you see any problem if we just return errors then rely on processNextWorkItem processing?

It seems out of scope here, this can be improved in another PR if needed.

shuo-wu:
OK. We can continue this discussion in our Slack channel.

PhanLe1010:
Thanks everyone for helping. I am back porting it to v1.2.x and v1.1.x but has some doubts so I will push the PR for those backport after some testing

joshimoo:
@PhanLe1010 @shuo-wu The reason conflicts were treated differently is probably because of the 3 errors limit before throwing the Ressource out of the queue.

Since conflicts are transitory we are guaranteed to eventually be able to process them.

We can reevaluate this error handling after we remove all permanent error loops "BUG:" cases as an example.

PhanLe1010:
Thanks @joshimoo 

Got it. To keep the current behavior, I will make a PR to add the object back with rate limited when there is a CR conflict.

PhanLe1010:
@shuo-wu, regarding to 
> Do we need to replace the enqueue functions like [this](https://github.com/longhorn/longhorn-manager/blob/840ae441597ac6af0904946db7d40916d959f9f8/controller/volume_controller.go#L401) with rate-limited requeue functions?

After looking deeper into this case, I think that adding the event with rate limited almost has the same effect has adding the event without rate limited. 

Reason: we return nil error, so even if we add the event back with the rate limited, we still forget it [here](https://github.com/longhorn/longhorn-manager/blob/555b5a38f57b1875af37614467623010f4d799b8/controller/volume_controller.go#L194). The forget call basically resets the timer for the event so it is not slowing down.

A correct way to slow down this event would be to return the error so the processNextWorkItem() doesn't call forget. This will need more evaluation as @joshimoo mentioned at https://github.com/longhorn/longhorn-manager/pull/1210#issuecomment-1036114423.

PR #1212, opened 2022-02-08
Implement storage network through gRPC proxy

c3y1huang:
https://github.com/longhorn/longhorn/issues/3546
https://github.com/longhorn/longhorn/issues/2285


manager/engine.go
```diff
@@ -225,6 +241,41 @@ func (m *VolumeManager) GetEngineClient(volumeName string) (client engineapi.Eng
 	})
 }
 
+func (m *VolumeManager) GetEngine(volumeName string) (e *longhorn.Engine, err error) {
```
shuo-wu, line 335: `GetAvailableEngine`/`GetRunningEngine`

engineapi/proxy_snapshot.go
```diff
@@ -30,6 +32,14 @@ func (p *Proxy) SnapshotGet(e *longhorn.Engine, name string) (snapshot *longhorn
 	return recv[name], nil
 }
 
+func (p *Proxy) SnapshotRevert(e *longhorn.Engine, name string) (err error) {
+	if name == VolumeHeadName {
+		return errors.Errorf("invalid operation: cannot revert to %v", VolumeHeadName)
+	}
```
shuo-wu, line 55: NIT: Should the gRPC Client handle the parameter checks?

engineapi/proxy_backup.go
```diff
@@ -51,3 +54,28 @@ func (p *Proxy) SnapshotBackupStatus(e *longhorn.Engine, backupName, replicaAddr
...
+func (p *Proxy) BackupRestore(e *longhorn.Engine, backupTarget, backupName, backupVolumeName, lastRestored string, credential map[string]string) error {
+	backupURL := backupstore.EncodeBackupURL(backupName, backupVolumeName, backupTarget)
+
+	// get environment variables if backup for s3
+	envs, err := getBackupCredentialEnv(backupTarget, credential)
+	if err != nil {
+		return err
+	}
+
+	taskError, err := p.grpcClient.BackupRestore(p.DirectToURL(e), backupURL, backupTarget, backupVolumeName, envs)
```
shuo-wu, line 72: NIT: Can the gRPC Client aggregate the errors into a single one?

shuo-wu (commented):
Will continue checking the network part tomorrow

controller/instance_manager_controller.go
```diff
@@ -1037,6 +1039,21 @@ func (imc *InstanceManagerController) createInstanceManagerPod(im *longhorn.Inst
...
+	}
+
+	nadAnnot := string(types.CNIAnnotationNetworks)
+	if storageNetwork.Value != types.CniNetworkNone {
+		err := types.ValidateStorageNetwork(storageNetwork.Value)
+		if err != nil {
+			return errors.Wrap(err, "failed to validate network-attachment-definition to create instance manager pod")
+		}
+
+		podSpec.Annotations[nadAnnot] = types.CreateCniAnnotationFromSetting(storageNetwork)
```
shuo-wu, line 1054: Will there be any impacts if users add similar annotations manually for the existing pods? I think NO, right?

shuo-wu (commented):
BTW, will you add some unit tests for the storage network part later?

engineapi/proxy.go
```diff
@@ -0,0 +1,239 @@
...
+	grpcClient *imclient.ProxyClient
+
+	lock *sync.RWMutex
+}
+
+func (p *Proxy) IsGRPC() bool {
+	return p.grpcClient != nil
+}
+
+func (p *Proxy) Start(im *longhorn.InstanceManager, logger logrus.FieldLogger, ds *datastore.DataStore) error {
```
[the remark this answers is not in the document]
c3y1huang: There will be a persistent and reusable client connection for each `controllerID`. Longhorn will launch one goroutine for that proxy client and periodically check with the `proxy.Ping`.

engineapi/proxy.go
```diff
@@ -0,0 +1,239 @@
...
+	Handle map[string]Client
+}
+
+func NewProxyHandler() *ProxyHandler {
+	return &ProxyHandler{
+		Handle: make(map[string]Client),
+	}
+}
+
+func (h *ProxyHandler) GetClient(ownerID string, im *longhorn.InstanceManager, fallBack interface{}) (Client, error) {
```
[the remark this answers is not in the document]
c3y1huang: > 1. When the caller tries to get a client, automatically create and add the client to the handler if not exist.

https://github.com/longhorn/longhorn-instance-manager/pull/105#discussion_r867605920

> 2. Modify this/Rename the interface to `CompatibleClient` instead. I mean, automatically **create** the binary client rather than asking the caller to input it if the IM is incompatible. (If the client is the binary client, Longhorn will add arg `--url` during the API invocation rather then during the client init.)
> 3. The main difference between EngineClient and the BackupTargetClient is about the credential. Can we directly use a single interface instead. When callers want to call the backup target related APIs, they need to input the credential.

Not sure if renaming to `CompatibleClient` make sense since the engine binary calls are temporary and [we've talked about deprecate it in the next major version](https://github.com/longhorn/longhorn/pull/3277#discussion_r866467405). Currently we have different inputs to get the binary client (`vc.getEngineClient(volume.Name)`, `GetClientForEngine(engine, m.engines, engine.Status.CurrentImage)`, `getBackupTargetClient(bc.ds, backupTarget)`), the idea is to reuse the current logic to get to  binary client for fallback, so we don't have to worry if the new logic for the binary client going to cause an issue.

controller/instance_manager_controller.go
```diff
@@ -512,6 +521,54 @@ func (imc *InstanceManagerController) syncMonitor(im *longhorn.InstanceManager)
...
+		log.Debugf("Adding to proxy handler")
+		imc.proxyHandler.Handle[imc.controllerID] = &engineapi.Proxy{}
+		proxy = imc.proxyHandler.Handle[imc.controllerID]
+	}
+
+	isInstanceManagerRunning := im.Status.CurrentState == longhorn.InstanceManagerStateRunning
+	if !isInstanceManagerRunning {
+		log.Warnf("Cannot proxy to gRPC while instance manager in %v state", im.Status.CurrentState)
+		return nil
+	}
```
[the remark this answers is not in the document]
c3y1huang: I prefer not to actively close the connection base on the IM status since I am not sure if the manager could still be talking to the IM proxy server. The connection is checked via `proxy.Ping` and the lifetime will be with the proxy server.

datastore/longhorn.go
```diff
@@ -193,10 +193,53 @@ func (s *DataStore) ValidateSetting(name, value string) (err error) {
...
+		}
+
+		volumesDetached, err := s.AreAllVolumesDetached()
+		if err != nil {
+			return errors.Wrapf(err, "failed to check volume detachment for %v setting update", types.SettingNameStorageNetwork)
+		}
+
+		if !volumesDetached {
+			return errors.Errorf("cannot update %v setting with volumes attached", types.SettingNameStorageNetwork)
+		}
```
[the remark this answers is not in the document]
derekbit: Yep. We can.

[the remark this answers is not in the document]
c3y1huang: If I understand correctly, the setting validation is already in the webhook. That means this part will be validated in the `(s *DataStore) ValidateSetting(name, value string)`.

controller/engine_controller.go
```diff
@@ -1299,17 +1309,19 @@ func (ec *EngineController) rebuildNewReplica(e *longhorn.Engine) error {
 	return nil
 }
 
-func doesAddressExistInEngine(addr string, client engineapi.EngineClient) (bool, error) {
-	replicaURLModeMap, err := client.ReplicaList()
+func (ec *EngineController) doesAddressExistInEngine(e *longhorn.Engine, addr string, proxy engineapi.Client) (bool, error) {
```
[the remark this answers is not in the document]
c3y1huang: Removed.

engineapi/proxy_snapshot.go
c3y1huang: Addressed.

engineapi/proxy_backup.go
c3y1huang: Addressed.

manager/engine.go
c3y1huang: Addressed.

controller/setting_controller.go
```diff
@@ -618,6 +622,51 @@ func (sc *SettingController) updateKubernetesClusterAutoscalerEnabled() error {
...
+		return err
+	}
+
+	volumesDetached, err := sc.ds.AreAllVolumesDetached()
+	if err != nil {
+		return errors.Wrapf(err, "failed to check volume detachment for %v setting update", types.SettingNameStorageNetwork)
+	}
+
+	if !volumesDetached {
+		return errors.Errorf("cannot update %v setting with volumes attached", types.SettingNameStorageNetwork)
```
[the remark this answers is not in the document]
c3y1huang: Updated.

c3y1huang:
After talking to @shuo-wu:
1. Use the instance manager image as the proxy handler key.
2. Let the proxy handler handle create/close connection instead of IM controller.
3. Since we are closing to release, we can tolerate inputting of the fallback binary client when getting the proxy client.

controller/instance_manager_controller.go
```diff
@@ -1037,6 +1039,21 @@ func (imc *InstanceManagerController) createInstanceManagerPod(im *longhorn.Inst
...
+	if err != nil {
+		return err
+	}
+
+	nadAnnot := string(types.CNIAnnotationNetworks)
+	if storageNetwork.Value != types.CniNetworkNone {
+		err := types.ValidateStorageNetwork(storageNetwork.Value)
+		if err != nil {
+			return errors.Wrap(err, "failed to validate network-attachment-definition to create instance manager pod")
+		}
```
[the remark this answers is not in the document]
c3y1huang: Removed.

types/setting.go
```diff
@@ -772,6 +775,20 @@ var (
 		ReadOnly: false,
 		Default:  "false",
 	}
+
+	SettingDefinitionStorageNetwork = SettingDefinition{
+		DisplayName: "Storage Network",
+		Description: "The storage network used for Longhorn in-cluster data traffic in <NAMESPACE>/<NETWORK-ATTACHMENT-DEFINITION> format. \n\n" +
```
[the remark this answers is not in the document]
c3y1huang: Updated.

controller/instance_manager_controller.go
c3y1huang: The annotation will only take effect when the pod is created, it has no effect on running pods.

engineapi/instance_manager.go
```diff
@@ -53,13 +54,25 @@ func GetDeprecatedInstanceManagerBinary(image string) string {
 }
 
 func CheckInstanceManagerCompatibilty(imMinVersion, imVersion int) error {
+	// Skip compatibility check for gPRC proxy implemented version so volume/engine upgrade is not blocked.
+	if imVersion < InstanceManagerProxyMinAPIVersion {
+		return nil
+	}
```
shuo-wu, line 60: Here, we can use 2 kinds of API versions instead:
`CurrentInstanceManagerAPIVersion` stands for the API version of the ProcessManager service/client hence there is no need to bump it. And `CurrentInstanceManagerProxyAPIVersion` would represent the proxy service/client version. 
Then we don't need to skip this compatibility check without blocking the process launching in old IMs.

shuo-wu (changes requested):

engineapi/proxy.go
```diff
@@ -0,0 +1,302 @@
...
+	}
+
+	log.Debugf("Adding %v client to proxy handler", image)
+
+	client := &Proxy{}
+	if err := client.Start(im, log, h.ds); err != nil {
+		return nil, err
+	}
+
+	h.Handle[image] = client
```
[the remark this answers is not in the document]
c3y1huang: Maybe we could just write lock `h.NewClient` and add the check and return client `h.Handle[image]` if exist.
So when 2 threads try to create its proxy client simultaneously, one will write lock the `h.NewClient` and the other wait until `h.NewClient` is released.
When the `h.NewClient` gets release it should find  `h.Handle[image]` exist, and return the client.

engineapi/instance_manager.go
c3y1huang: How would we get the version from the IMs? Do you mean to add in a new field in `im.Status.ProxyAPIVersion`?

engineapi/proxy.go
```diff
@@ -0,0 +1,302 @@
...
+	}
+
+	return nil
+}
+
+func (p *Proxy) NewProxyClient(im *longhorn.InstanceManager) error {
+	// Do not check the major version here. Since IM cannot get the major version without using this client to call VersionGet().
+	if im.Status.CurrentState != longhorn.InstanceManagerStateRunning || im.Status.IP == "" {
+		return errors.Errorf("invalid Instance Manager %v, state: %v, IP: %v", im.Name, im.Status.CurrentState, im.Status.IP)
+	}
```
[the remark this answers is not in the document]
c3y1huang: Addressed.

engineapi/proxy.go
```diff
@@ -0,0 +1,302 @@
...
+
+func (p *Proxy) startProxyClient(im *longhorn.InstanceManager) error {
+	logrus.Debug("Connecting to proxy server")
+	err := p.NewProxyClient(im)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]

engineapi/proxy.go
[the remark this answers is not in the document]
c3y1huang: Added lock for NewClient.

engineapi/proxy.go
```diff
@@ -0,0 +1,302 @@
...
+
+func getLoggerForProxy(logger logrus.FieldLogger, image string) *logrus.Entry {
+	return logger.WithFields(
+		logrus.Fields{
+			"instanceManagerImage": image,
+		},
+	)
+}
+
+func (h *ProxyHandler) GetClient(image string, im *longhorn.InstanceManager, fallBack interface{}) (EngineClientProxy, error) {
```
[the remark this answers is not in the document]
c3y1huang: Fixed.

engineapi/proxy.go
```diff
@@ -0,0 +1,291 @@
...
+		return nil, errors.Errorf("BUG: invalid fallback client for proxy: %v", fallBack)
+	}
+
+	client, exist := h.Handle[image]
+	if exist {
+		err := client.Ping()
+		if err == nil {
+			return client, nil
+		}
+	}
```
innobead, line 105: This is duplicated in `NewClient`

engineapi/proxy.go
```diff
@@ -0,0 +1,291 @@
...
+	return h.NewClient(log, image, im)
+}
+
+func (h *ProxyHandler) NewClient(log logrus.FieldLogger, image string, im *longhorn.InstanceManager) (Client, error) {
+	h.lock.Lock()
+	defer h.lock.Unlock()
+
+	existClient, exist := h.Handle[image]
+	if exist {
+		err := existClient.Ping()
```
innobead, line 116: What's timeout for the ping? is it possible to cause some latency for `NewClient` due to timeout ping? or it's acceptable?

types/setting.go
```diff
@@ -791,6 +809,13 @@ const (
 	SystemManagedPodsImagePullPolicyAlways       = SystemManagedPodsImagePullPolicy("always")
 )
 
+type CNI string
```
innobead, line 812: CNIAnnotation.

innobead (changes requested):

innobead:
In general, LGTM and my comments are more about naming convention. I will check the details again tomorrow.

controller/backing_image_manager_controller.go
```diff
@@ -460,6 +460,26 @@ func (c *BackingImageManagerController) syncBackingImageManagerPod(bim *longhorn
...
+
+				if storageIP == types.CniNetworkNone {
+					logrus.Debugf("Cannot find storage-network IP from pod %v to update backing image status, use cluster IP %v", pod.Name, pod.Status.PodIP)
+					storageIP = pod.Status.PodIP
+				}
+
+				if bim.Status.StorageIP != storageIP {
+					bim.Status.StorageIP = storageIP
+					logrus.Debugf("Inconsistent storage-network IP from pod %v, update backing image status IP %v", pod.Name, bim.Status.IP)
+				}
```
innobead, line 481: Can we have a util funtion to get the storage IP from the pod instead?

ex: `bim.Status.IP = getStorageIPFromPod(...)`

controller/backup_controller.go
```diff
@@ -577,15 +602,31 @@ func (bc *BackupController) hasMonitor(backupName string) *engineapi.BackupMonit
...
+
+	im, err := bc.ds.GetInstanceManager(engine.Status.InstanceManagerName)
+	if err != nil {
+		return nil, err
+	}
+
+	proxy, err := bc.ProxyHandler.GetClient(im, binClient)
+	if err != nil {
+		return nil, err
+	}
```
innobead, line 627: The same, I feel these statements can be consolidated to a util method/function.

engineapi/backups.go
```diff
@@ -22,6 +23,16 @@ const (
 	backupStateError      = "error"
 )
 
+type BackupTargetBinaryClient interface {
```
innobead, line 26: This is interface not implementation, so suggest call `BackupTargetClient`?

engineapi/engine.go
```diff
@@ -22,26 +24,30 @@ const (
...
+type EngineBinary struct {
 	name  string
 	image string
 	ip    string
 	port  int
 	cURL  string
 }
 
-func (c *EngineCollection) NewEngineClient(request *EngineClientRequest) (EngineClient, error) {
+func (c *EngineCollection) NewEngineClient(request *EngineClientRequest) (*EngineBinary, error) {
```
innobead, line 41: Why not just return `EngineClientProxy` instead of `*EngineBinary`? so we can operate on interface instead of struct.

engineapi/proxy.go
c3y1huang: We can return existing responsive client without `NewClient`.
The duplicated Ping in `NewClient` is to avoid overwriting client connections when multiple threads trying to create a proxy client simultaneously.

controller/backup_controller.go
```diff
@@ -495,13 +500,13 @@ func (bc *BackupController) checkMonitor(backup *longhorn.Backup, volume *longho
 	}
 
 	// Find the corresponding engine client
-	engineClient, err := bc.getEngineClient(volume.Name)
+	binClient, err := bc.getEngineClient(volume.Name)
```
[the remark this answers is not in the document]
c3y1huang: Updated.

engineapi/proxy.go
```diff
@@ -0,0 +1,282 @@
...
+
+// TODO: replace with EngineClientProxy interface later
+type Client interface {
+	IsGRPC() bool
+	Start(*longhorn.InstanceManager, logrus.FieldLogger, *datastore.DataStore) error
+	Stop(string) error
+	Ping() error
+}
+
+type ProxyHandler struct {
```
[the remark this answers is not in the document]

engineapi/proxy.go
c3y1huang: As discussed. Will keep the Ping for now. I will create another issue for enhancement.

engineapi/proxy.go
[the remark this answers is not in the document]
c3y1huang: After discussing with @innobead . I will create another issue for enhancement.

engineapi/proxy.go
c3y1huang: ATM the client has 3 minutes context timeout. I will create another issue for enhancement.

https://github.com/longhorn/longhorn-manager/pull/1212#discussion_r869998036

types/setting.go
c3y1huang: Updated.

controller/backing_image_manager_controller.go
c3y1huang: Addressed.

controller/backup_controller.go
c3y1huang: Addressed.

engineapi/engine.go
c3y1huang: This is used for fallback returns so we don't error out on missing methods.

controller/backup_controller.go
```diff
@@ -495,13 +520,13 @@ func (bc *BackupController) checkMonitor(backup *longhorn.Backup, volume *longho
...
 	if err != nil {
 		return nil, err
 	}
 
 	// Enable the backup monitor
-	monitor, err := bc.enableBackupMonitor(backup, volume, backupTargetClient, biChecksum, engineClient)
+	monitor, err := bc.enableBackupMonitor(backup, volume, backupTargetClient, biChecksum, binClient)
```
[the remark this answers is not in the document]
c3y1huang: Addressed.

engineapi/backup_monitor.go
```diff
@@ -46,14 +49,18 @@ type BackupMonitor struct {
...
+	biChecksum string, binClient EngineClient, engine *longhorn.Engine, proxy Client,
+	syncCallback func(key string)) (*BackupMonitor, error) {
 	ctx, quit := context.WithCancel(context.Background())
 	m := &BackupMonitor{
 		logger: logger.WithFields(logrus.Fields{"backup": backup.Name}),
 
 		backupName:   backup.Name,
 		snapshotName: backup.Spec.SnapshotName,
-		engineClient: engineClient,
+		engineClient: binClient,
```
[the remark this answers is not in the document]
c3y1huang: Not required, removed.

engineapi/proxy.go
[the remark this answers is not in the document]
c3y1huang: https://github.com/longhorn/longhorn/issues/3967

engineapi/backups.go
c3y1huang: Talked to @innobead , the name is taken as [struct](https://github.com/longhorn/longhorn-manager/blob/master/engineapi/backups.go#L25) already. Keep it for now.

innobead: Thanks, @c3y1huang . 

note: I feel our naming pattern requires some improvement and consistency especially to differentiate the interface and structure but this can be tackled later. (this is common in all languages)

engineapi/instance_manager.go
shuo-wu: Similar to the existing APIVersion field. Yes. We need to add:
1. The corresponding fields in `im.Status`
2. The corresponding fields in cli and server output

engineapi/proxy.go
shuo-wu: The below client initialization & connection are already protected by the lock. Then the race case I raised won't happen. I would recommend removing the `Ping` in `h.NewClient()`.

engineapi/proxy.go
```diff
@@ -0,0 +1,312 @@
...
+		logrus.Fields{
+			"instanceManagerImage": image,
+		},
+	)
+}
+
+func (h *EngineClientProxyHandler) GetClient(im *longhorn.InstanceManager, fallBack interface{}) (c EngineClientProxy, err error) {
+	image := ""
+	if im == nil {
+		image, err = h.ds.GetSettingValueExisted(types.SettingNameDefaultInstanceManagerImage)
```
shuo-wu, line 66: Then you need to get the IM CR for this default image. Otherwise, the below `h.NetClient` must fail and the caller has no way to get the client.

engineapi/proxy.go
```diff
@@ -0,0 +1,312 @@
...
+		err := client.Ping()
+		if err == nil {
+			return client, nil
+		}
+	}
+
+	return h.NewClient(log, image, im)
+}
+
+func (h *EngineClientProxyHandler) NewClient(log logrus.FieldLogger, image string, im *longhorn.InstanceManager) (EngineClientProxy, error) {
```
shuo-wu, line 111: Same here. Remove input `image`

engineapi/proxy.go
```diff
@@ -0,0 +1,312 @@
...
+				return
+			}
+		case <-p.stopCh:
+			return
+		}
+	}
+}
+
+func (p *Proxy) sync() bool {
+	im, err := p.ds.GetInstanceManager(p.InstanceManagerName)
```
shuo-wu, line 242: NIT: `GetInstanceManagerRO`

datastore/longhorn.go
```diff
@@ -193,10 +193,53 @@ func (s *DataStore) ValidateSetting(name, value string) (err error) {
 		if err := types.ValidateCPUReservationValues(guaranteedEngineManagerCPU.Value, guaranteedReplicaManagerCPU.Value); err != nil {
 			return err
 		}
+	case types.SettingNameStorageNetwork:
+		if err := types.ValidateStorageNetwork(value); err != nil {
+			return err
+		}
```
shuo-wu, line 199: Redundant. This is already validated in the above invocation `types.ValidateSetting(name, value)`.

shuo-wu (changes requested):

engineapi/proxy.go
c3y1huang: Not expecting im to be nil. Removed.

engineapi/proxy.go

engineapi/proxy.go

datastore/longhorn.go
c3y1huang: Removed

c3y1huang: Removed.

engineapi/proxy.go
```diff
@@ -0,0 +1,312 @@
...
+		}
+
+		if obj, ok := fallBack.(*BackupTargetClient); ok {
+			return obj, nil
+		}
+
+		return nil, errors.Errorf("BUG: invalid fallback client for proxy: %v", fallBack)
+	}
+
+	client, exist := h.Clients[image]
```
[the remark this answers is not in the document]
c3y1huang: Fixed.

engineapi/proxy.go
c3y1huang: Removed.

engineapi/instance_manager.go
c3y1huang: Addressed.

engineapi/proxy.go
```diff
@@ -0,0 +1,299 @@
...
+	hasNewOwnerID := im.Status.OwnerID != p.InstanceManagerControllerID
+	if hasNewOwnerID {
+		p.logger.Infof("Stopping proxy because the instance manager changed ownership from %v to %v", p.InstanceManagerControllerID, im.Status.OwnerID)
+		return true
+	}
+
+	return false
+}
+
+func (p *Proxy) NewProxyClient(im *longhorn.InstanceManager) error {
```
shuo-wu, line 249: NIT: do we need this function?

engineapi/proxy.go
```diff
@@ -0,0 +1,299 @@
...
+		err := client.Ping()
+		if err == nil {
+			return client, nil
+		}
+
+		log.WithError(err).Debug("Resetting proxy connection")
+		client.Stop(im.Spec.Image)
+		delete(h.Clients, im.Spec.Image)
+	}
+	h.lock.Unlock()
```
shuo-wu, line 108: Redundant. There is already a defer function above for the unlocking.

engineapi/proxy.go
```diff
@@ -0,0 +1,299 @@
...
+
+		return nil, errors.Errorf("BUG: invalid fallback client for proxy: %v", fallBack)
+	}
+
+	h.lock.Lock()
+	defer h.lock.Unlock()
+
+	client, exist := h.Clients[im.Spec.Image]
+	if exist {
+		err := client.Ping()
```
shuo-wu, line 99: Should this ping function be protected by the h.lock?

engineapi/proxy.go

engineapi/proxy.go

engineapi/proxy.go
c3y1huang: I've put the lock around delete and create of the client. WDYT?

shuo-wu: Need to protect the read as well:
```
h.lock.Lock()
client, exist := h.Clients[im.Spec.Image]
h.lock.Unlock()
```

controller/engine_controller.go
```diff
@@ -907,15 +924,18 @@ func (m *EngineMonitor) refresh(engine *longhorn.Engine) error {
...
+		if err = cloneSnapshot(engine, engineClientProxy, m.ds); err != nil {
 			return err
 		}
 	}
 
 	return nil
 }
 
-func preRestoreCheckAndSync(log logrus.FieldLogger, engine *longhorn.Engine, rsMap map[string]*longhorn.RestoreStatus, addressReplicaMap map[string]string, cliAPIVersion int, client engineapi.EngineClient, ds *datastore.DataStore) (needRestore bool, err error) {
+func preRestoreCheckAndSync(log logrus.FieldLogger, engine *longhorn.Engine,
```
shuo-wu, line 935: Does this function still need `client *engineapi.EngineBinary`? The caller will input the compatible client (returned from `proxyHandler.GetClient(im, engineCliClient)`), which may be a binary one.
Similar below.

controller/engine_controller.go
```diff
@@ -1213,7 +1233,7 @@ func (ec *EngineController) ReconcileEngineState(e *longhorn.Engine) error {
 	return nil
 }
 
-func GetClientForEngine(e *longhorn.Engine, engines engineapi.EngineClientCollection, image string) (client engineapi.EngineClient, err error) {
+func GetClientForEngine(e *longhorn.Engine, engines engineapi.EngineClientCollection, image string) (client *engineapi.EngineBinary, err error) {
```
shuo-wu, line 1236: NIT: Rename this function. e.g., `GetBinaryClientForEngine`

controller/engine_controller.go
```diff
@@ -1415,11 +1458,23 @@ func (ec *EngineController) Upgrade(e *longhorn.Engine) (err error) {
...
+	im, err := ec.ds.GetInstanceManager(e.Status.InstanceManagerName)
 	if err != nil {
 		return err
 	}
-	version, err := client.Version(false)
+
+	engineClientProxy, err := ec.proxyHandler.GetClient(im, engineCliClient)
+	if err != nil {
+		return err
+	}
```
shuo-wu, line 1475: NIT: I saw this kind of code multiple times. Abstract it as a function.

engineapi/proxy.go
```diff
@@ -0,0 +1,304 @@
...
+
+		h.lock.Lock()
+		log.WithError(err).Debug("Resetting proxy connection")
+		client.Stop(im.Spec.Image)
+
+		// TODO: do not delete the map to avoid nil memory panic. Fix this when
+		// refactoring to use goroutine for proxy client handling.
+		// delete(h.Clients, im.Spec.Image)
+
+		h.lock.Unlock()
```
shuo-wu, line 111: `h.Clients` should be init during the ProxyHandler creation. Why will there be NPE? If the function does not need to delete entries from `h.Clients`, there is no need to use a lock protecting the stop either.

shuo-wu (changes requested):

engineapi/proxy.go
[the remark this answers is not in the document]
c3y1huang: Updated. Break into `GetCompatibleClient` and `GetClient`.

engineapi/types.go
```diff
@@ -59,7 +59,7 @@ type EngineClient interface {
 	Ping() error
 
 	Name() string
-	Version(clientOnly bool) (*EngineVersion, error)
+	VersionGet(engine *longhorn.Engine, binary EngineClient, clientOnly bool) (*EngineVersion, error)
```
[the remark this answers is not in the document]
c3y1huang: Fixed.

engineapi/proxy.go
```diff
@@ -0,0 +1,304 @@
...
+	clientVersion, err := binary.VersionGet(nil, nil, true)
+	if err != nil {
+		return version, err
+	}
+
+	if clientOnly {
+		return &EngineVersion{
+			ClientVersion: clientVersion.ClientVersion,
+		}, nil
+	}
```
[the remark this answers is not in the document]
c3y1huang: Fixed.

controller/backup_controller.go
```diff
@@ -265,15 +270,14 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
 
 		if backupTarget.Spec.BackupTargetURL != "" &&
 			backupVolume != nil && backupVolume.DeletionTimestamp == nil {
-			// Initialize a backup target client
-			backupTargetClient, err := getBackupTargetClient(bc.ds, backupTarget)
+			engineClientProxy, backupTargetClient, err := getBackupTargetClients(bc.controllerID, backupTarget, bc.ProxyHandler, bc.ds, log)
```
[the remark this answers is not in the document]
c3y1huang: We don't need to fall back to backupTargetClient.

controller/backing_image_data_source_controller.go
```diff
@@ -210,6 +215,26 @@ func getLoggerForBackingImageDataSource(logger logrus.FieldLogger, bids *longhor
 	)
 }
 
+func (c *BackingImageDataSourceController) getProxyClientForBackingImageDataSource(e *longhorn.Engine, volumeName string) (engineapi.EngineClientProxy, error) {
```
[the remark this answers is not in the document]
c3y1huang: Renamed.

controller/engine_controller.go
c3y1huang: Removed.

controller/engine_controller.go
c3y1huang: Renamed.

controller/engine_controller.go
c3y1huang: Addressed.

engineapi/proxy.go
c3y1huang: There will be NPE if any caller already has the client but another caller failed to Ping and delete the map. Removed lock around `client.Stop`.

shuo-wu: Getting clients from `h.Clients` is already protected by the lock. And repeatedly calling `client.Ping()` should be fine...

controller/backup_controller.go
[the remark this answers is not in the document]
shuo-wu: I mean, can we directly remove the struct `BackupTargetClient` in `engineapi/backup.go` and the related methods?

shuo-wu (commented):
NIT: we can abstract some functions like `GetEngineClientProxyForInstanceManager` and `GetEngineClientProxyForEngine` then the callers do not need to manually get the binary client then check if the proxy client is available.

controller/backup_controller.go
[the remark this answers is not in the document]
c3y1huang: Removed.

engineapi/proxy.go
c3y1huang: Updated as discussed.

controller/backing_image_data_source_controller.go
```diff
@@ -210,6 +215,21 @@ func getLoggerForBackingImageDataSource(logger logrus.FieldLogger, bids *longhor
 	)
 }
 
+func (c *BackingImageDataSourceController) getEngineClientProxy(e *longhorn.Engine, volumeName string) (engineapi.EngineClientProxy, error) {
+	engineCollection := &engineapi.EngineCollection{}
+	engineCliClient, err := engineCollection.NewEngineClient(&engineapi.EngineClientRequest{
+		EngineImage: e.Status.CurrentImage,
+		VolumeName:  volumeName,
```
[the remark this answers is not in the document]
c3y1huang: Updated.

c3y1huang:
Test result (coretest) - [#1230](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/1230/testReport/)

shuo-wu (approved):
In general LGTM

engineapi/proxy.go
derekbit: h.lock is a RWMutex. Why not use RLock/RUnLock for read operation?
```
h.lock.RLock()
client, exist := h.Clients[im.Spec.Image]
h.lock.RUnlock()
```

PR #1215, opened 2022-02-11
Initialize backingimage and backingimagedatasource sourceType

derekbit:
Fix upgrade failure from v1.1.x to master-head.
The sourceType of backingimage and backingimagedatasource in v1.1.x is "", so initialize it in the mutating webhook.


innobead (approved):

innobead:
@derekbit we should probably need to check if we’d just remove the related empty check conditions if needed.

PR #1219, opened 2022-02-15
Add NodeInformer event handler to volume controller

PhanLe1010:
Make sure to enqueue volume for replica scheduling when there is a change on a node.

longhorn/longhorn#3626


innobead (approved):
LGTM

Waiting for @shuo-wu review.

controller/volume_controller.go
```diff
@@ -3459,6 +3466,40 @@ func (vc *VolumeController) enqueueVolumesForBackingImageDataSource(obj interfac
...
+
+		// use the last known state, to enqueue, dependent objects
+		node, ok = deletedState.Obj.(*longhorn.Node)
+		if !ok {
+			utilruntime.HandleError(fmt.Errorf("DeletedFinalStateUnknown contained invalid object: %#v", deletedState.Obj))
+			return
+		}
+	}
+
+	replicas, err := vc.ds.ListReplicasRO()
```
shuo-wu, line 3486: You can have 2 list functions: One is for the unscheduled replicas, another is for the failed replicas **on this node only**.
This is just a tiny suggestion, I am fine with the current implementation as well.

shuo-wu (commented):
In general LGTM

shuo-wu (approved):

controller/volume_controller.go
PhanLe1010: Thanks @shuo-wu ! Good idea as usual! 

I would like to keep the current implementation to verify the theory that the slow enqueueNodeChange function shouldn't affect the scalability

innobead:
@PhanLe1010 please backport this. thanks.

PhanLe1010:
As discussed in Slack, we will change the resync back to 30s for v1.1.4 and v1.2.4 since there are potentially conner cases that we cannot catch before releasing. Those are stable releases so we don't want to have mistakes. Therefore, we don't need to backport this PR to those branches

PR #1222, opened 2022-02-17
Remove unnecessary permission

jenting:
#### Proposal Change

Remove unnecessary permission. The `nsenter` only needs permissions
- `hostPID: true`
- `securityContext.privileged: true`

#### Issue
https://github.com/longhorn/longhorn/issues/3549

### DIFF
--- a/deploy/prerequisite/longhorn-iscsi-installation.yaml
+++ b/deploy/prerequisite/longhorn-iscsi-installation.yaml
@@ -15,7 +15,6 @@ spec:
       labels:
         app: longhorn-iscsi-installation
     spec:
-      hostNetwork: true
       hostPID: true
       initContainers:
       - name: iscsi-installation
--- a/deploy/prerequisite/longhorn-nfs-installation.yaml
+++ b/deploy/prerequisite/longhorn-nfs-installation.yaml
@@ -15,7 +15,6 @@ spec:
       labels:
         app: longhorn-nfs-installation
     spec:
-      hostNetwork: true
       hostPID: true
       initContainers:
       - name: nfs-installation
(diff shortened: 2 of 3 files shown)


deploy/install/02-components/03-ui.yaml
```diff
@@ -19,8 +19,6 @@ spec:
       - name: longhorn-ui
         image: longhornio/longhorn-ui:master-head
         imagePullPolicy: IfNotPresent
-        securityContext:
-          runAsUser: 0
```
innobead, line 23: Because right now our images have the root user, we don't need this actually. But I would prefer to keep it on purpose to notice we need the root for the current components. WDYT?

deploy/prerequisite/longhorn-iscsi-installation.yaml
```diff
@@ -15,7 +15,6 @@ spec:
       labels:
         app: longhorn-iscsi-installation
     spec:
-      hostNetwork: true
```
innobead, line 18: It looks YES, have you tried? the same question for nfs.

jenting: Yes, I tried.

deploy/install/02-components/03-ui.yaml
innobead: Ok, keep this

deploy/install/02-components/04-driver.yaml
```diff
@@ -79,5 +79,3 @@ spec:
 #        label-key1: "label-value1"
 #        label-key2: "label-value2"
       serviceAccountName: longhorn-service-account
-      securityContext:
```
innobead, line 82: Ok, keep this

innobead (changes requested):

innobead:
We need to backport this to longhorn/longhorn.

innobead (approved):

PR #1223, opened 2022-02-18
Fix  condition unmarshal failure & implement volume recurring job label clean up

c3y1huang:
Fix unable to parse engine image with array type condition issue. Use longhorn client instead - partial [Longhorn-3470](https://github.com/longhorn/longhorn/issues/3470).

Implement volume clean-up when the recurring job/job-group is deleted - https://github.com/longhorn/longhorn/issues/3517.

### DIFF
--- a/app/recurring_job.go
+++ b/app/recurring_job.go
@@ -148,8 +148,8 @@ func recurringJob(c *cli.Context) error {
 	}
 	logger.Infof("Found %v volumes with recurring job %v", len(filteredVolumes), jobName)
 
-	var wg sync.WaitGroup
 	concurrentLimiter := make(chan struct{}, jobConcurrent)
+	var wg sync.WaitGroup
 	defer wg.Wait()
 	for _, volumeName := range filteredVolumes {
 		wg.Add(1)
@@ -257,6 +257,11 @@ func NewJob(logger logrus.FieldLogger, managerURL, volumeName, snapshotName stri
 // It should detach the volume when:
 // 1. The volume is attached by this recurring job
 // 2. The volume state is VolumeStateAttached
+// NOTE:
+//   The volume could remain attached when the recurring job pod gets force
+//   terminated and unable to complete detachment within the grace period. Thus
+//   there is a workaround in the recurring job controller to handle the
+//   detachment again (detachVolumeAutoAttachedByRecurringJob).
 func (job *Job) handleVolumeDetachment() {
 	volumeAPI := job.api.Volume
 	volumeName := job.volumeName
--- a/app/recurring_job.go
+++ b/app/recurring_job.go
@@ -710,6 +710,10 @@ func (job *Job) GetVolume(name string) (*longhorn.Volume, error) {
 	return job.lhClient.LonghornV1beta2().Volumes(job.namespace).Get(context.TODO(), name, metav1.GetOptions{})
 }
 
+func (job *Job) GetEngineImage(name string) (*longhorn.EngineImage, error) {
+	return job.lhClient.LonghornV1beta2().EngineImages(job.namespace).Get(context.TODO(), name, metav1.GetOptions{})
+}
+
 func (job *Job) UpdateVolumeStatus(v *longhorn.Volume) (*longhorn.Volume, error) {
 	return job.lhClient.LonghornV1beta2().Volumes(job.namespace).UpdateStatus(context.TODO(), v, metav1.UpdateOptions{})
 }
@@ -740,12 +744,12 @@ func (job *Job) findARandomReadyNode(v *longhornclient.Volume) (string, error) {
 		return "", err
 	}
 
-	engineImage, err := job.api.EngineImage.ById(types.GetEngineImageChecksumName(v.CurrentImage))
+	engineImage, err := job.GetEngineImage(types.GetEngineImageChecksumName(v.CurrentImage))
 	if err != nil {
 		return "", err
 	}
-	if engineImage.State != string(longhorn.EngineImageStateDeployed) && engineImage.State != string(longhorn.EngineImageStateDeploying) {
-		return "", fmt.Errorf("error: the volume's engine image %v is in state: %v", engineImage.Name, engineImage.State)
+	if engineImage.Status.State != longhorn.EngineImageStateDeployed && engineImage.Status.State != longhorn.EngineImageStateDeploying {
+		return "", fmt.Errorf("error: the volume's engine image %v is in state: %v", engineImage.Name, engineImage.Status.State)
 	}
 
 	var readyNodeList []string
@@ -771,7 +775,7 @@ func (job *Job) findARandomReadyNode(v *longhornclient.Volume) (string, error) {
 				return "", err
 			}
 
-			if readyCondition.Status == longhorn.ConditionStatusTrue && engineImage.NodeDeploymentMap[node.Name] {
+			if readyCondition.Status == longhorn.ConditionStatusTrue && engineImage.Status.NodeDeploymentMap[node.Name] {
 				readyNodeList = append(readyNodeList, node.Name)
 			}
 		}
--- a/controller/recurring_job_controller.go
+++ b/controller/recurring_job_controller.go
@@ -196,6 +196,10 @@ func (control *RecurringJobController) syncRecurringJob(key string) (err error) 
 		log.Infof("Recurring Job got new owner %v", control.controllerID)
 	}
 
+	if recurringJob.DeletionTimestamp != nil {
+		return control.cleanupVolumeRecurringJob(recurringJob)
+	}
+
 	existingRecurringJob := recurringJob.DeepCopy()
 	defer func() {
 		if err != nil {
@@ -217,6 +221,58 @@ func (control *RecurringJobController) syncRecurringJob(key string) (err error) 
 		log.WithError(err).Error("failed to reconcile recurring job")
 	}
 
+	return nil
+}
+
+func (control *RecurringJobController) cleanupVolumeRecurringJob(recurringJob *longhorn.RecurringJob) error {
+	// Check if each group of the recurring job contains other recurring jobs.
+	// If No, it means the recurring job is the last job of the group then
+	// Longhorn will clean up this group labels for all volumes.
+	checkRecurringJobs, err := control.ds.ListRecurringJobs()
+	if err != nil {
+		return err
+	}
+	rmGroups := []string{}
+	for _, group := range recurringJob.Spec.Groups {
+		inUse := false
+		for _, checkRecurringJob := range checkRecurringJobs {
+			if checkRecurringJob.Name == recurringJob.Name {
+				continue
+			}
+			if util.Contains(checkRecurringJob.Spec.Groups, group) {
+				inUse = true
+				break
+			}
+		}
+		if !inUse {
+			rmGroups = append(rmGroups, group)
+		}
+	}
+
+	// delete volume labels
+	volumes, err := control.ds.ListVolumes()
+	if err != nil {
+		return err
+	}
+	for _, vol := range volumes {
+		jobs := datastore.MarshalLabelToVolumeRecurringJob(vol.Labels)
+		for jobName, job := range jobs {
+			if job.IsGroup {
+				if !util.Contains(rmGroups, jobName) {
+					continue
+				}
+				control.logger.Debugf("Clean up volume recurring job-group %v for %v", jobName, vol.Name)
+				if _, err := control.ds.DeleteVolumeRecurringJob(jobName, true, vol); err != nil {
+					return err
+				}
+			} else if jobName == recurringJob.Name {
+				control.logger.Debugf("Clean up volume recurring job %v for %v", jobName, vol.Name)
+				if _, err := control.ds.DeleteVolumeRecurringJob(jobName, false, vol); err != nil {
+					return err
+				}
+			}
+		}
+	}
 	return nil
 }
 
--- a/controller/recurring_job_controller.go
+++ b/controller/recurring_job_controller.go
@@ -175,6 +175,24 @@ func (control *RecurringJobController) syncRecurringJob(key string) (err error) 
 			return err
 		}
 		log.Debug("Cannot find recurring job, may have been deleted")
+
+		// The detachVolumeAutoAttachedByRecurringJob is a workaround to
+		// resolve volume unable to detach via the recurring job. The volume
+		// could remain attached via recurringjob auto-attachment when the
+		// recurring job pod gets force terminated and unable to complete
+		// detachment within the grace period.
+		// This should be handled when a separate controller is introduced for
+		// attachment and detachment handling.
+		// https://github.com/longhorn/longhorn-manager/pull/1223#discussion_r814655791
+		volumes, err := control.ds.ListVolumes()
+		if err != nil {
+			return err
+		}
+		for _, vol := range volumes {
+			if err := control.detachVolumeAutoAttachedByRecurringJob(name, vol); err != nil {
+				return err
+			}
+		}
 		return nil
 	}
 
@@ -271,6 +289,20 @@ func (control *RecurringJobController) cleanupVolumeRecurringJob(recurringJob *l
 					return err
 				}
 			}
+		}
+	}
+	return nil
+}
+
+func (control *RecurringJobController) detachVolumeAutoAttachedByRecurringJob(name string, v *longhorn.Volume) error {
+	if v.Spec.LastAttachedBy != name {
+		return nil
+	}
+	if v.Status.State == longhorn.VolumeStateAttached {
+		control.logger.Infof("requesting auto-attached volume %v to detach from node %v", v.Name, v.Spec.NodeID)
+		v.Spec.NodeID = ""
+		if _, err := control.ds.UpdateVolume(v); err != nil {
+			return err
 		}
 	}
 	return nil
--- a/datastore/longhorn.go
+++ b/datastore/longhorn.go
@@ -557,6 +557,52 @@ func (s *DataStore) ListVolumes() (map[string]*longhorn.Volume, error) {
 	return itemMap, nil
 }
 
+func MarshalLabelToVolumeRecurringJob(labels map[string]string) map[string]*longhorn.VolumeRecurringJob {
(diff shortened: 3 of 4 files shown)


c3y1huang:
After this fix the test still errors [here](https://github.com/longhorn/longhorn-tests/blob/master/manager/integration/tests/test_recurring_job.py#L800), and failed at precondition
```
Events:
  Type     Reason              Age                   From                     Message
  ----     ------              ----                  ----                     -------
  Normal   Scheduled           6m14s                 default-scheduler        Successfully assigned default/longhorn-testvol-teqvm7-dep-76f95bbb79-pshn2 to ip-172-30-0-131
  Warning  FailedAttachVolume  5m5s (x8 over 6m13s)  attachdetach-controller  AttachVolume.Attach failed for volume "longhorn-testvol-teqvm7-pv" : rpc error: code = FailedPrecondition desc = volume longhorn-testvol-teqvm7 cannot be attached to node ip-172-30-0-131 is already attached to node ip-172-30-0-140
  Warning  FailedMount         4m11s                 kubelet                  Unable to attach or mount volumes: unmounted volumes=[volv], unattached volumes=[volv kube-api-access-l9hqc]: timed out waiting for the condition
  Warning  FailedAttachVolume  118s (x2 over 4m1s)   attachdetach-controller  AttachVolume.Attach failed for volume "longhorn-testvol-teqvm7-pv" : rpc error: code = FailedPrecondition desc = volume longhorn-testvol-teqvm7 cannot be attached to node ip-172-30-0-131 is already attached to node ip-172-30-0-98
  Warning  FailedMount         113s                  kubelet                  Unable to attach or mount volumes: unmounted volumes=[volv], unattached volumes=[kube-api-access-l9hqc volv]: timed out waiting for the condition
```

manager/recurringjob.go
```diff
@@ -114,9 +114,64 @@ func (m *VolumeManager) UpdateRecurringJob(spec longhorn.RecurringJobSpec) (*lon
...
+
+	// check if other recurring jobs contains same groups
+	recurringJobs, err := m.ds.ListRecurringJobs()
+	if err != nil {
+		return err
+	}
+	rmGroups := []string{}
+	for _, group := range recurringJob.Spec.Groups {
+		inUse := 0
+		for _, obj := range recurringJobs {
```
innobead, line 142: Suggest using a meaningful object name instead of `obj`.

manager/recurringjob.go
```diff
@@ -114,9 +114,64 @@ func (m *VolumeManager) UpdateRecurringJob(spec longhorn.RecurringJobSpec) (*lon
...
+		}
+	}
+
+	// delete volume labels
+	volumes, err := m.ds.ListVolumes()
+	if err != nil {
+		return err
+	}
+	for volName, vol := range volumes {
+		list, err := m.ListVolumeRecurringJob(volName)
```
innobead, line 161: Suggest using a meaningful object name instead of `list`. In general for other places if needed.

manager/recurringjob.go
```diff
@@ -114,9 +114,64 @@ func (m *VolumeManager) UpdateRecurringJob(spec longhorn.RecurringJobSpec) (*lon
...
+	}
+	rmGroups := []string{}
+	for _, group := range recurringJob.Spec.Groups {
+		inUse := 0
+		for _, obj := range recurringJobs {
+			if util.Contains(obj.Spec.Groups, group) {
+				inUse++
+			}
+			if inUse > 1 {
+				continue
```
innobead, line 147: This should be `break`? to avoid continuing iteration if the group has been found to assign to other replica jobs.

manager/recurringjob.go
```diff
@@ -114,9 +114,64 @@ func (m *VolumeManager) UpdateRecurringJob(spec longhorn.RecurringJobSpec) (*lon
...
+			}
+		}
+		if inUse == 1 {
```
innobead, line 150: Could we use `inUse == 0` as the condition by excluding the job which is `recurringJob`? Using `==1` is not intuitive or implicit.

manager/recurringjob.go
```diff
@@ -114,9 +114,64 @@ func (m *VolumeManager) UpdateRecurringJob(spec longhorn.RecurringJobSpec) (*lon
 }
 
 func (m *VolumeManager) DeleteRecurringJob(name string) error {
+	if err := m.CleanVolumeRecurringJob(name); err != nil {
```
innobead, line 117: Should we clean up in the controller instead to react for the recurring job deletion instead of just tackling in the same call?

manager/recurringjob.go
```diff
@@ -114,9 +114,64 @@ func (m *VolumeManager) UpdateRecurringJob(spec longhorn.RecurringJobSpec) (*lon
...
+		return err
+	}
+	for volName, vol := range volumes {
+		list, err := m.ListVolumeRecurringJob(volName)
+		if err != nil {
+			return err
+		}
+		if vol.Spec.LastAttachedBy == name && vol.Status.State != longhorn.VolumeStateDetached {
+			m.Detach(volName, "")
+		}
```
innobead, line 167: Why we need this? I thought recurring job management should not impact volume management.

manager/recurringjob.go
```diff
@@ -114,9 +114,64 @@ func (m *VolumeManager) UpdateRecurringJob(spec longhorn.RecurringJobSpec) (*lon
...
+		for jobName := range list {
```
innobead, line 168: Is it job name duplicated? like the same name for a job or a group? maybe you can leverage job properties to do the condition instead of job name? WDYT?

innobead (changes requested):

app/recurring_job.go
```diff
@@ -740,12 +744,12 @@ func (job *Job) findARandomReadyNode(v *longhornclient.Volume) (string, error) {
 		return "", err
 	}
 
-	engineImage, err := job.api.EngineImage.ById(types.GetEngineImageChecksumName(v.CurrentImage))
+	engineImage, err := job.GetEngineImage(types.GetEngineImageChecksumName(v.CurrentImage))
```
innobead, line 747: Do you mean `engineImage, err := job.api.EngineImage.ById(types.GetEngineImageChecksumName(v.CurrentImage))` also error due to decoding problem?

innobead:
https://github.com/longhorn/longhorn/issues/3517#issuecomment-1046591929 cc @derekbit

app/recurring_job.go
c3y1huang: Correct.

manager/recurringjob.go
c3y1huang: Addressed.

manager/recurringjob.go
c3y1huang: Addressed, moved logic to the controller.

manager/recurringjob.go
c3y1huang: Fixed. Note logic moved to the controller.

manager/recurringjob.go
c3y1huang: Updated to use boolean instead.

manager/recurringjob.go
c3y1huang: Thanks, that makes more sense. Addressed.

manager/recurringjob.go
c3y1huang: Actually, recurring job also manages auto-attachment [here](https://github.com/longhorn/longhorn-manager/blob/master/app/recurring_job.go#L330). This was added in the case when the pod gets removed before it detaches the volume.

But since the controller now does the clean-up, this is not needed. Removed.

manager/recurringjob.go
c3y1huang: Good catch, I missed this one. Fixed and also added to the automation test case in this [PR](https://github.com/longhorn/longhorn-tests/pull/891).

c3y1huang:
Test result - https://ci.longhorn.io/job/private/job/longhorn-tests-regression/723/testReport/tests/test_recurring_job/

controller/recurring_job_controller.go
```diff
@@ -220,6 +224,51 @@ func (control *RecurringJobController) syncRecurringJob(key string) (err error)
 	return nil
 }
 
+func (control *RecurringJobController) cleanupVolumeRecurringJob(recurringJob *longhorn.RecurringJob) error {
+	// check if other recurring jobs contains same groups
```
shuo-wu, line 228: // Check if each group of the RJ contains other RJs. If No, it means the RJ is the last job of the group then Longhorn will clean up this group labels for all volumes.

datastore/longhorn.go
```diff
@@ -557,6 +557,56 @@ func (s *DataStore) ListVolumes() (map[string]*longhorn.Volume, error) {
...
+	}
+	return jobMapVolumeJob
+}
+
+func (s *DataStore) DeleteVolumeRecurringJob(volumeName string, name string, isGroup bool) (v *longhorn.Volume, err error) {
+	defer func() {
+		err = errors.Wrapf(err, "failed to delete volume recurring jobs for %v", volumeName)
+	}()
+
+	v, err = s.GetVolume(volumeName)
```
shuo-wu, line 588: NIT: Why do we re-get `v *longhorn.Volume` rather than relying on the caller to input it? If you prefer to retain the current implementation, you can use `ListVolumesByXxxxxxRO` in the caller to avoid meaningless deep copies.

controller/recurring_job_controller.go
```diff
@@ -220,6 +224,51 @@ func (control *RecurringJobController) syncRecurringJob(key string) (err error)
...
+				break
+			}
+		}
+		if !inUse {
+			rmGroups = append(rmGroups, group)
+		}
+	}
+
+	// delete volume labels
+	volumes, err := control.ds.ListVolumes()
```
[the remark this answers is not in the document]
c3y1huang: We need to check each volume and each labels in volumes to delete the labels for the enqueued recurring job. Also need to check if the recurring job is the last in the group. I prefer to keep it this way because to separate the logic into two functions means it needs to loop twice through the labels to get the names (job & job-group).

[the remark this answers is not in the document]
shuo-wu: OK. In brief, you prefer to use a single loop here.

c3y1huang:
Found some flaky test results caused by recurring job pod force termination. Added commit in this PR to fix.
Test result - https://ci.longhorn.io/job/private/job/longhorn-tests-regression/730/

datastore/longhorn.go
c3y1huang: Fixed.

controller/recurring_job_controller.go
c3y1huang: Addressed.

controller/recurring_job_controller.go
```diff
@@ -220,6 +224,53 @@ func (control *RecurringJobController) syncRecurringJob(key string) (err error)
...
+				break
+			}
+		}
+		if !inUse {
+			rmGroups = append(rmGroups, group)
+		}
+	}
+
+	// delete volume labels
+	volumes, err := control.ds.ListVolumesRO()
```
shuo-wu, line 253: When you decide to mutate the objects in this list result later, please do not use `RO`. Or you can rerun the get function before the mutator.

c3y1huang: Fixed.

c3y1huang:
Test result - https://ci.longhorn.io/job/private/job/longhorn-tests-regression/741/

shuo-wu (approved):

controller/recurring_job_controller.go
```diff
@@ -318,6 +369,10 @@ func (control *RecurringJobController) newCronJob(recurringJob *longhorn.Recurri
 	}
 	registrySecret := registrySecretSetting.Value
 
+	// Cronjob sends SIGTERM to the container when RecurringJob gets removed.
+	// Need to allow enough time for termination for recurring job auto-attached
+	// volume to finish and detach volume.
+	terminationGracePeriodSeconds := int64(types.DefaultRecurringJobTerminationGracePeriodSeconds)
```
innobead, line 375: Whats the issue for adding this?

types/types.go
```diff
@@ -114,6 +114,8 @@ const (
 	ControlPlaneName                 = "longhorn-manager"
 
 	DefaultRecurringJobConcurrency = 10
+	// DefaultRecurringJobTerminationGracePeriodSeconds set to 24 hours
```
innobead, line 117: Why 24hrs?

innobead (changes requested):

controller/recurring_job_controller.go
innobead: Ok checked again the commit history. Why not fix at e2e to wait until the volume detached but use lengthen the grace period here instead ?

app/recurring_job.go
```diff
@@ -148,6 +148,9 @@ func recurringJob(c *cli.Context) error {
 	}
 	logger.Infof("Found %v volumes with recurring job %v", len(filteredVolumes), jobName)
 
+	done := make(chan struct{})
+	util.RegisterShutdownChannel(done)
```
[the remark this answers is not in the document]
c3y1huang: We don't want to block this. I've added this so the pod shows log when cronjob sends for SIGTERM.

controller/recurring_job_controller.go
c3y1huang: When the RecurringJob gets removed, its CronJob will send SIGTERM to the recurring job pod. The pod could get force terminated when it doesn't finish within the grace period, and this is likely to result:
1. Incomplete backups.
2. Volume not detached when it is set to allow recurring-job while volume id detached.
So I feel this is a better solution than fixing the e2e.

types/types.go
c3y1huang: We should leave enough time for the backups to complete. I assume normally the backup shouldn't exceed beyond a day.

app/recurring_job.go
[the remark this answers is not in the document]
innobead: For debugging purpose? What scenario?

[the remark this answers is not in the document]
c3y1huang: When allow-recurring-job-while-volume-detached is enabled. See https://github.com/longhorn/longhorn-manager/pull/1223#discussion_r814537892.

controller/recurring_job_controller.go
innobead: What u want to achieve is to try detach the volume before the pod gets terminated. 

As discussed, think about if trying to detach the volume when receiving the termination signal instead of having a long termination period to keep the pod running until the backup finished and the volume detached, even the corresponding recurring job has been deleted. Because this will probably cause some confusion when there is a new recurring job being created for that volume, but it will never be executed successfully because a terminating job still running.

Also, a failed backup is reasonable if the recurring job is deleted, so probably we shouldn't do anything to make it work in this case.

However know your intention.

shuo-wu: The ultimate solution for this issue is volume-controller refactoring. After the refactoring, there will be a separate controller handling all kinds of attachment & detachment.

c3y1huang: > P.S. in any solution, it's still possible to fail to detach the volume at the end when passing the termination period. It's fine but need to note.

I've rethought about catching the termination signal. IMO it might not be a good solution because it still means the volumes needs to be detached within the 30 seconds grace time. It is possible for a large number of volumes the detach could exceed the grace period. So I've think moving the logic to the recurring job controller might be a better solution.

c3y1huang:
Test result - https://ci.longhorn.io/job/private/job/longhorn-tests-regression/767/testReport/tests/test_recurring_job/

controller/recurring_job_controller.go
```diff
@@ -220,6 +233,78 @@ func (control *RecurringJobController) syncRecurringJob(key string) (err error)
...
+		}
+	}
+	return nil
+}
+
+func (control *RecurringJobController) detachVolumeAutoAttachedByRecurringJob(name string, v *longhorn.Volume) error {
+	if v.Spec.LastAttachedBy != name {
+		return nil
+	}
+	// !volume.DisableFrontend condition makes sure that volume is detached by this recurring job,
```
innobead, line 292: Why not just reuse `VolumeManager::Detach`? We should better leverage existing functions instead of updating the spec directly if the operation is general.

innobead (changes requested):
In general, LGTM except the detach implemnetation.

controller/recurring_job_controller.go
c3y1huang: The logic is similar to [detachShareManagerVolume](https://github.com/longhorn/longhorn-manager/blob/master/controller/share_manager_controller.go#L399). I don't think we can call VolumeManager in the controller? Or do you think we should new a VolumeManager for this?

innobead: hmm, right now the volume manager seems the interface for the API instead. It's fine for now, but it looks like the code reuse can be improved a bit.

innobead (dismissed):

controller/recurring_job_controller.go
```diff
@@ -175,6 +175,15 @@ func (control *RecurringJobController) syncRecurringJob(key string) (err error)
 			return err
 		}
 		log.Debug("Cannot find recurring job, may have been deleted")
+		volumes, err := control.ds.ListVolumes()
+		if err != nil {
+			return err
+		}
+		for _, vol := range volumes {
+			if err := control.detachVolumeAutoAttachedByRecurringJob(name, vol); err != nil {
```
shuo-wu, line 183: Wait... The detachment function in the recurring job is not removed yet. Do you still want to add this kind of duplicate handling?
https://github.com/longhorn/longhorn-manager/blob/ac190f908d33a8c0f135167d73c3aed904f16d62/app/recurring_job.go#L260

shuo-wu (commented):
We didn't reach a consensus on the detachment logic. And it seems to be unrelated to the original ticket (title). At least I don't think we can do detachment in both the recurring job pod and the controller. Can we separate it from this PR first?

controller/recurring_job_controller.go
shuo-wu: After talking with Chin-Ya, I think the risk of this detachment logic is acceptable. It may be triggered only when the recurring job is gone but the pod does not handle the cleanup gracefully.
After adding a comment here, this PR is good to go.

innobead (approved):

controller/recurring_job_controller.go
innobead: Agreed and thanks for adding the comments for the later revisit.

PR #1227, opened 2022-02-20
Implement mutation and validation logics of resources in the admission webhook

derekbit:
The general mutation and validation logics in HTTP API functions are moved or re-implemented in the admission webhook.

[longhorn #3562](https://github.com/longhorn/longhorn/issues/3562)

### DIFF
--- a/deploy/install/02-components/05-webhook.yaml
+++ b/deploy/install/02-components/05-webhook.yaml
@@ -120,6 +120,10 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: metadata.namespace
+        - name: NODE_NAME
+          valueFrom:
+            fieldRef:
+              fieldPath: spec.nodeName
 #      imagePullSecrets:
 #      - name: ""
 #      priorityClassName:
--- a/manager/backingimage.go
+++ b/manager/backingimage.go
@@ -3,7 +3,6 @@ package manager
 import (
 	"fmt"
 	"reflect"
-	"strings"
 
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
@@ -12,11 +11,7 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
-	"github.com/longhorn/longhorn-manager/controller"
-	"github.com/longhorn/longhorn-manager/datastore"
-	"github.com/longhorn/longhorn-manager/engineapi"
 	"github.com/longhorn/longhorn-manager/types"
-	"github.com/longhorn/longhorn-manager/util"
 
 	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
 )
@@ -72,67 +67,6 @@ func (m *VolumeManager) GetBackingImageDataSourcePod(name string) (*v1.Pod, erro
 }
 
 func (m *VolumeManager) CreateBackingImage(name, checksum, sourceType string, parameters map[string]string) (bi *longhorn.BackingImage, err error) {
-	name = util.AutoCorrectName(name, datastore.NameMaximumLength)
-	if !util.ValidateName(name) {
-		return nil, fmt.Errorf("invalid name %v", name)
-	}
-
-	if len(checksum) != 0 {
-		checksum = strings.TrimSpace(checksum)
-
-		if !util.ValidateChecksumSHA512(checksum) {
-			return nil, fmt.Errorf("invalid checksum %v", checksum)
-		}
-	}
-
-	for k, v := range parameters {
-		parameters[k] = strings.TrimSpace(v)
-	}
-
-	switch longhorn.BackingImageDataSourceType(sourceType) {
-	case longhorn.BackingImageDataSourceTypeDownload:
-		if parameters[longhorn.DataSourceTypeDownloadParameterURL] == "" {
-			return nil, fmt.Errorf("invalid parameter %+v for source type %v", parameters, sourceType)
-		}
-	case longhorn.BackingImageDataSourceTypeUpload:
-	case longhorn.BackingImageDataSourceTypeExportFromVolume:
-		volumeName := parameters[controller.DataSourceTypeExportFromVolumeParameterVolumeName]
-		if volumeName == "" {
-			return nil, fmt.Errorf("invalid parameter %+v for source type %v", parameters, sourceType)
-		}
-		v, err := m.ds.GetVolume(volumeName)
-		if err != nil {
-			return nil, fmt.Errorf("failed to get volume %v before exporting backing image", volumeName)
-		}
-		if v.Status.Robustness == longhorn.VolumeRobustnessFaulted {
-			return nil, fmt.Errorf("cannot export a backing image from faulted volume %v", volumeName)
-		}
-		eiName := types.GetEngineImageChecksumName(v.Status.CurrentImage)
-		ei, err := m.ds.GetEngineImage(eiName)
-		if err != nil {
-			return nil, errors.Wrapf(err, "failed to get then check engine image %v for volume %v before exporting backing image", eiName, volumeName)
-		}
-		if ei.Status.CLIAPIVersion < engineapi.CLIVersionFive {
-			return nil, fmt.Errorf("engine image %v CLI version %v doesn't support this feature, please upgrade engine for volume %v before exporting backing image from the volume", eiName, ei.Status.CLIAPIVersion, volumeName)
-		}
-		// By default the exported file type is raw.
-		if parameters[DataSourceTypeExportFromVolumeParameterExportType] == "" {
-			parameters[DataSourceTypeExportFromVolumeParameterExportType] = DataSourceTypeExportFromVolumeParameterExportTypeRAW
-		}
-		if parameters[DataSourceTypeExportFromVolumeParameterExportType] != DataSourceTypeExportFromVolumeParameterExportTypeRAW &&
-			parameters[DataSourceTypeExportFromVolumeParameterExportType] != DataSourceTypeExportFromVolumeParameterExportTypeQCOW2 {
-			return nil, fmt.Errorf("unsupported export type %v", parameters[DataSourceTypeExportFromVolumeParameterExportType])
-		}
-	default:
-		return nil, fmt.Errorf("unknown backing image source type %v", sourceType)
-	}
-
-	if _, err := m.ds.GetBackingImage(name); err == nil {
-		return nil, fmt.Errorf("backing image already exists")
-	} else if !apierrors.IsNotFound(err) {
-		return nil, errors.Wrapf(err, "failed to check backing image existence before creation")
-	}
-
 	bi = &longhorn.BackingImage{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:   name,
@@ -154,13 +88,6 @@ func (m *VolumeManager) CreateBackingImage(name, checksum, sourceType string, pa
 }
 
 func (m *VolumeManager) DeleteBackingImage(name string) error {
-	replicas, err := m.ds.ListReplicasByBackingImage(name)
-	if err != nil {
-		return err
-	}
-	if len(replicas) != 0 {
-		return fmt.Errorf("cannot delete backing image %v since there are replicas using it", name)
-	}
 	if err := m.ds.DeleteBackingImage(name); err != nil {
 		return err
 	}
--- a/manager/node.go
+++ b/manager/node.go
@@ -1,11 +1,6 @@
 package manager
 
 import (
-	"fmt"
-	"math"
-
-	"github.com/longhorn/longhorn-manager/types"
-	"github.com/longhorn/longhorn-manager/util"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 
@@ -65,43 +60,6 @@ func (m *VolumeManager) GetNodeTags() ([]string, error) {
 }
 
 func (m *VolumeManager) UpdateNode(n *longhorn.Node) (*longhorn.Node, error) {
-	// We need to make sure the tags passed in are valid before updating the node.
-	tags, err := util.ValidateTags(n.Spec.Tags)
-	if err != nil {
-		return nil, err
-	}
-	n.Spec.Tags = tags
-
-	if n.Spec.EngineManagerCPURequest < 0 || n.Spec.ReplicaManagerCPURequest < 0 {
-		return nil, fmt.Errorf("found invalid EngineManagerCPURequest %v or ReplicaManagerCPURequest %v during node %v update", n.Spec.EngineManagerCPURequest, n.Spec.ReplicaManagerCPURequest, n.Name)
-	}
-	if n.Spec.EngineManagerCPURequest != 0 || n.Spec.ReplicaManagerCPURequest != 0 {
-		kubeNode, err := m.ds.GetKubernetesNode(n.Name)
-		if err != nil {
-			return nil, err
-		}
-		allocatableCPU := float64(kubeNode.Status.Allocatable.Cpu().MilliValue())
-		engineManagerCPUSetting, err := m.ds.GetSetting(types.SettingNameGuaranteedEngineManagerCPU)
-		if err != nil {
-			return nil, err
-		}
-		engineManagerCPUInPercentage := engineManagerCPUSetting.Value
-		if n.Spec.EngineManagerCPURequest > 0 {
-			engineManagerCPUInPercentage = fmt.Sprintf("%.0f", math.Round(float64(n.Spec.EngineManagerCPURequest)/allocatableCPU*100.0))
-		}
-		replicaManagerCPUSetting, err := m.ds.GetSetting(types.SettingNameGuaranteedReplicaManagerCPU)
-		if err != nil {
-			return nil, err
-		}
-		replicaManagerCPUInPercentage := replicaManagerCPUSetting.Value
-		if n.Spec.ReplicaManagerCPURequest > 0 {
-			replicaManagerCPUInPercentage = fmt.Sprintf("%.0f", math.Round(float64(n.Spec.ReplicaManagerCPURequest)/allocatableCPU*100.0))
-		}
-		if err := types.ValidateCPUReservationValues(engineManagerCPUInPercentage, replicaManagerCPUInPercentage); err != nil {
-			return nil, err
-		}
-	}
-
 	node, err := m.ds.UpdateNode(n)
 	if err != nil {
 		return nil, err
@@ -145,29 +103,6 @@ func (m *VolumeManager) DiskUpdate(name string, updateDisks map[string]longhorn.
 		return nil, err
 	}
 
-	originDisks := node.Spec.Disks
-
-	for name, uDisk := range updateDisks {
-		if uDisk.StorageReserved < 0 {
-			return nil, fmt.Errorf("update disk on node %v error: The storageReserved setting of disk %v(%v) is not valid, should be positive and no more than storageMaximum and storageAvailable", name, name, uDisk.Path)
-		}
-
-		tags, err := util.ValidateTags(uDisk.Tags)
-		if err != nil {
-			return nil, err
-		}
-		uDisk.Tags = tags
-		updateDisks[name] = uDisk
-	}
-
-	// delete disks
-	for name, oDisk := range originDisks {
-		if _, ok := updateDisks[name]; !ok {
-			if oDisk.AllowScheduling || node.Status.DiskStatus[name].StorageScheduled != 0 {
-				return nil, fmt.Errorf("Delete Disk on node %v error: Please disable the disk %v and remove all replicas first ", name, oDisk.Path)
-			}
-		}
-	}
 	node.Spec.Disks = updateDisks
 
 	node, err = m.ds.UpdateNode(node)
@@ -179,28 +114,6 @@ func (m *VolumeManager) DiskUpdate(name string, updateDisks map[string]longhorn.
 }
 
 func (m *VolumeManager) DeleteNode(name string) error {
-	node, err := m.ds.GetNode(name)
-	if err != nil {
-		return err
-	}
-	// only remove node from longhorn without any volumes on it
-	replicas, err := m.ds.ListReplicasByNodeRO(name)
-	if err != nil {
-		return err
-	}
-	engines, err := m.ds.ListEnginesByNodeRO(name)
-	if err != nil {
-		return err
-	}
-	condition := types.GetCondition(node.Status.Conditions, longhorn.NodeConditionTypeReady)
-	// Only could delete node from longhorn if kubernetes node missing or manager pod is missing
-	if condition.Status == longhorn.ConditionStatusTrue ||
-		(condition.Reason != longhorn.NodeConditionReasonKubernetesNodeGone &&
-			condition.Reason != longhorn.NodeConditionReasonManagerPodMissing) ||
-		node.Spec.AllowScheduling || len(replicas) > 0 || len(engines) > 0 {
-		return fmt.Errorf("could not delete node %v with node ready condition is %v, reason is %v, node schedulable %v, and %v replica, %v engine running on it", name,
-			condition.Status, condition.Reason, node.Spec.AllowScheduling, len(replicas), len(engines))
-	}
 	if err := m.ds.DeleteNode(name); err != nil {
 		return err
 	}
--- a/manager/volume.go
+++ b/manager/volume.go
@@ -14,10 +14,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
-	"github.com/longhorn/backupstore"
-
 	"github.com/longhorn/longhorn-manager/datastore"
-	"github.com/longhorn/longhorn-manager/engineapi"
 	"github.com/longhorn/longhorn-manager/scheduler"
 	"github.com/longhorn/longhorn-manager/types"
 	"github.com/longhorn/longhorn-manager/util"
@@ -146,14 +143,6 @@ func (m *VolumeManager) GetReplicasSorted(vName string) ([]*longhorn.Replica, er
 	return replicas, nil
 }
 
-func (m *VolumeManager) getDefaultReplicaCount() (int, error) {
-	c, err := m.ds.GetSettingAsInt(types.SettingNameDefaultReplicaCount)
-	if err != nil {
-		return 0, err
-	}
-	return int(c), nil
-}
-
 func (m *VolumeManager) Create(name string, spec *longhorn.VolumeSpec, recurringJobSelector []longhorn.VolumeRecurringJob) (v *longhorn.Volume, err error) {
 	defer func() {
 		err = errors.Wrapf(err, "unable to create volume %v", name)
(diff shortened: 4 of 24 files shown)


webhook/resources/node/mutator.go
```diff
@@ -35,29 +37,55 @@ func (n *nodeMutator) Resource() admission.Resource {
...
+			tags, err := util.ValidateTags(node.Spec.Tags)
+			if err != nil {
+				return nil, err
+			}
+			bytes, err := json.Marshal(tags)
+			if err != nil {
+				return nil, err
+			}
+			patchOps = append(patchOps, fmt.Sprintf(`{"op": "replace", "path": "/spec/tags", "value": %s}`, string(bytes)))
+		}
```
shuo-wu, line 67: Should Longhorn validate the tags in the validator rather than the mutator? Then is the following `json.Marshal()` required?

Similar below.

webhook/resources/engineimage/mutator.go
```diff
@@ -0,0 +1,62 @@
...
+	if image != engineImage.Spec.Image {
+		patchOps = append(patchOps, fmt.Sprintf(`{"op": "replace", "path": "/spec/image", "value": "%s"}`, image))
+	}
+
+	name := types.GetEngineImageChecksumName(image)
+	if name != engineImage.Name {
+		patchOps = append(patchOps, fmt.Sprintf(`{"op": "replace", "path": "/metadata/name", "value": "%s"}`, name))
+	}
+
+	labels := types.GetEngineImageLabels(name)
```
shuo-wu, line 54: Can we allow users to set the customized labels during this CR creation?

webhook/resources/backingimage/validator.go
```diff
@@ -0,0 +1,114 @@
...
+	default:
+		return werror.NewInvalidError(fmt.Sprintf("unknown backing image source type %v", backingImage.Spec.SourceType), "")
+	}
+
+	if _, err := b.ds.GetBackingImage(backingImage.Name); err == nil {
+		return werror.NewInvalidError("backing image already exists", "")
+	} else if !apierrors.IsNotFound(err) {
+		err = errors.Wrapf(err, "failed to check backing image existence before creation")
+		return werror.NewInvalidError(err.Error(), "")
+	}
```
shuo-wu, line 98: Is this redundant?

manager/volume.go
```diff
@@ -216,65 +207,11 @@ func (m *VolumeManager) Create(name string, spec *longhorn.VolumeSpec, recurring
...
-	}
-
-	if spec.Migratable && spec.AccessMode != longhorn.AccessModeReadWriteMany {
-		return nil, fmt.Errorf("migratable volumes are only supported in ReadWriteMany (rwx) access mode")
-	}
-
 	defaultEngineImage, err := m.GetSettingValueExisted(types.SettingNameDefaultEngineImage)
 	if defaultEngineImage == "" {
 		return nil, fmt.Errorf("BUG: Invalid empty Setting.EngineImage")
 	}
```
shuo-wu, line 213: Redundant. This is already moved into the validator.

shuo-wu (commented):
Great job!
In general LGTM. Hence I didn't take a review line-by-line. If required, I can do it tomorrow.

webhook/resources/backingimage/validator.go
derekbit: Hmm?

shuo-wu: If the backing image CR already exists before the creation, the creation function will directly return an error, right?

webhook/resources/engineimage/mutator.go
shuo-wu: Then will op `replace` overwrite the labels customized by users themselves?

derekbit: Yes. This will overwrite the labels set by users.
Do you think we need to merge the labels set by `GetEngineImageLabels` and users?
Originally, the labels are from `GetEngineImageLabels`, correct?

webhook/resources/node/mutator.go
shuo-wu: is the main mutation in the function `sort.Strings(tags)`? If YES, I think we can let this mutator take over the sorting only.

manager/volume.go
shuo-wu: Oh, sorry. It's not in the validator. We may need to move this to the validator.

derekbit: Ah, thank you. I forgot this part.

webhook/resources/backingimage/validator.go
derekbit: Good catch. It's redundant, because the creation function won't be executed if the backing image CR already exists.

webhook/resources/engineimage/mutator.go
shuo-wu: IMO, Longhorn just needs to **add and correct** the labels for the **specific** keys.

webhook/resources/node/mutator.go
derekbit: Yes. Sounds good. Will follow your suggestion.

shuo-wu (changes requested):
I think we need to remove the validations from the volume Creation/Update functions in `manager/volume.go`. Furthermore, as I mentioned in the comments, we need to re-visit and simplify the VolumeManager implementation.

derekbit:
> I think we need to remove the validations from the volume Creation/Update functions in `manager/volume.go`. Furthermore, as I mentioned in the comments, we need to re-visit and simplify the VolumeManager implementation.

Agree. 

One question while I was handling the validation function.
What's the purpose of `CheckEngineImageReadiness`?
How do we get the `currentNodeID` after moving to the webhook?

```
func (m *VolumeManager) verifyDataSourceForVolumeCreation(dataSource longhorn.VolumeDataSource, requestSize int64) (err error) {
...
		if snapName := types.GetSnapshotName(dataSource); snapName != "" {
			if _, err := m.GetSnapshot(snapName, srcVolName); err != nil {
				return err
			}
		}
	}
	return nil
}
```

```
func (m *VolumeManager) GetSnapshot(snapshotName, volumeName string) (*longhorn.Snapshot, error) {
	if volumeName == "" || snapshotName == "" {
		return nil, fmt.Errorf("volume and snapshot name required")
	}
	engine, err := m.GetEngineClient(volumeName)
	if err != nil {
		return nil, err
	}
	snapshot, err := engine.SnapshotGet(snapshotName)
```

```
func (m *VolumeManager) GetEngineClient(volumeName string) (client engineapi.EngineClient, err error) {
	var e *longhorn.Engine
...

	if isReady, err := m.ds.CheckEngineImageReadiness(e.Status.CurrentImage, m.currentNodeID); !isReady {
		if err != nil {
...
```

shuo-wu:
When `volume.spec.DataSource` is set, it indicates that the new volume is a clone of an old volume. In Longhorn, the clone is similar to rebuilding but involves 2 volumes. You can check this for details: https://github.com/longhorn/longhorn/issues/1815

manager/node.go
```diff
@@ -145,29 +103,6 @@ func (m *VolumeManager) DiskUpdate(name string, updateDisks map[string]longhorn.
...
-
-	// delete disks
-	for name, oDisk := range originDisks {
-		if _, ok := updateDisks[name]; !ok {
-			if oDisk.AllowScheduling || node.Status.DiskStatus[name].StorageScheduled != 0 {
-				return nil, fmt.Errorf("Delete Disk on node %v error: Please disable the disk %v and remove all replicas first ", name, oDisk.Path)
-			}
-		}
-	}
 	node.Spec.Disks = updateDisks
```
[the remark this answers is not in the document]
derekbit: @shuo-wu 
I will create a ticket for `volumeManager` refactoring, okay?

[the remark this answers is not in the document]
shuo-wu: Sure. But you still need to move the validation/mutation logic to the webhook in this PR......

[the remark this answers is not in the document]
derekbit: Yep.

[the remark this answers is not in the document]
derekbit: @shuo-wu 
Some **update** validation and mutation cannot be moved to the webhook, because there are conflicts between the actions. https://github.com/longhorn/longhorn/issues/3562#issuecomment-1041214241
Thus, I only move the general validation and mutation logics to the webhook.
WDYT?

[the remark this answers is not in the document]
shuo-wu: OK, YES. We just need to move the general checks to the webhook.

manager/volume.go
```diff
@@ -284,24 +161,18 @@ func (m *VolumeManager) Create(name string, spec *longhorn.VolumeSpec, recurring
...
-	if spec.DataSource != "" {
-		if err := m.verifyDataSourceForVolumeCreation(spec.DataSource, size); err != nil {
-			return nil, err
-		}
-	}
-
 	v = &longhorn.Volume{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:   name,
+			Name:   util.AutoCorrectName(name, datastore.NameMaximumLength),
```
shuo-wu, line 166: Can we correct the naming in the mutator?

webhook/resources/volume/validator.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+
+	if err := datastore.CheckVolume(volume); err != nil {
+		return werror.NewInvalidError(err.Error(), "")
+	}
+
+	return nil
+}
+
+func (v *volumeValidator) Update(request *admission.Request, oldObj runtime.Object, newObj runtime.Object) error {
```
shuo-wu, line 113: Notice that this function has both the old obj and the new obj, we can move more validations from the update APIs to this validator. e.g., once there is a change for field `Spec.NumberOfReplicas`, we need to apply these checks https://github.com/longhorn/longhorn-manager/blob/c04eac1c804597c78d5962c1758e8ed80d2abb37/manager/volume.go#L1016-L1024

And this logic can be applied to other validators. If this would take time, you can open one more PR for it.

webhook/resources/volume/validator.go
```diff
@@ -0,0 +1,149 @@
...
+			return werror.NewInvalidError(err.Error(), "")
+		}
+	}
+
+	currentNodeID, err := util.GetRequiredEnv(types.EnvNodeName)
+	if err != nil {
+		return werror.NewInvalidError("BUG: fail to detect the node name", "")
+	}
+
+	m := manager.NewVolumeManager(currentNodeID, v.ds)
```
shuo-wu, line 99: IMO, pkg `manager` is mainly used by the functions in pkg `api`. You can move the required util functions to other pkgs like `datastore` instead.

webhook/resources/volume/validator.go
derekbit: > And this logic can be applied to other validators. If this would take time, you can open one more PR for it.

I will process this part in a separate PR.

webhook/resources/volume/validator.go
```diff
@@ -0,0 +1,231 @@
...
+
+	// Check engine version before disable revision counter
+	if volume.Spec.RevisionCounterDisabled {
+		if ok, err := v.canDisableRevisionCounter(volume.Spec.EngineImage); !ok {
+			err := errors.Wrapf(err, "can not create volume with current engine image that doesn't support disable revision counter")
+			return werror.NewInvalidError(err.Error(), "")
+		}
+	}
+
+	currentNodeID, err := util.GetRequiredEnv(types.EnvNodeName)
```
shuo-wu, line 93: NIT: This can be stored in `volumeValidator`.

shuo-wu (approved):
In general LGTM

webhook/resources/node/validator.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+}
+
+func (n *nodeValidator) Update(request *admission.Request, oldObj runtime.Object, newObj runtime.Object) error {
+	oldNode := oldObj.(*longhorn.Node)
+	newNode := newObj.(*longhorn.Node)
+
+	// Only scheduling disabled node can be evicted
+	// Can not enable scheduling on an evicting node
+	if oldNode.Spec.EvictionRequested && oldNode.Spec.AllowScheduling {
```
innobead, line 48: I think this should validate again `newNode` instead of `oldNode`, because we want to validate the user request instead of the current settings.

webhook/resources/node/validator.go
```diff
@@ -0,0 +1,149 @@
...
+			oldNode.Name), "")
+	}
+
+	// We need to make sure the tags passed in are valid before updating the node.
+	_, err := util.ValidateTags(newNode.Spec.Tags)
+	if err != nil {
+		return werror.NewInvalidError(err.Error(), "")
+	}
+
+	if newNode.Spec.EngineManagerCPURequest < 0 || newNode.Spec.ReplicaManagerCPURequest < 0 {
```
innobead, line 59: Don't these have been covered by OpenAPI schema validation? `+kubebuilder:validation:Minimum`

webhook/resources/node/validator.go
```diff
@@ -0,0 +1,149 @@
...
+		if err != nil {
+			return werror.NewInvalidError(err.Error(), "")
+		}
+		allocatableCPU := float64(kubeNode.Status.Allocatable.Cpu().MilliValue())
+		engineManagerCPUSetting, err := n.ds.GetSetting(types.SettingNameGuaranteedEngineManagerCPU)
+		if err != nil {
+			return werror.NewInvalidError(err.Error(), "")
+		}
+		engineManagerCPUInPercentage := engineManagerCPUSetting.Value
+		if newNode.Spec.EngineManagerCPURequest > 0 {
```
innobead, line 75: `newNode.Spec.EngineManagerCPURequest` should be always > 0 here, so probably we don't need to have if check here anymore.

webhook/resources/node/validator.go
```diff
@@ -0,0 +1,149 @@
...
+			engineManagerCPUInPercentage = fmt.Sprintf("%.0f", math.Round(float64(newNode.Spec.EngineManagerCPURequest)/allocatableCPU*100.0))
+		}
+		replicaManagerCPUSetting, err := n.ds.GetSetting(types.SettingNameGuaranteedReplicaManagerCPU)
+		if err != nil {
+			return werror.NewInvalidError(err.Error(), "")
+		}
+		replicaManagerCPUInPercentage := replicaManagerCPUSetting.Value
+		if newNode.Spec.ReplicaManagerCPURequest > 0 {
```
innobead, line 83: Don't need this if check, because the same as `newNode.Spec.EngineManagerCPURequest`.

webhook/resources/engineimage/validator.go
```diff
@@ -0,0 +1,61 @@
...
+		OperationTypes: []admissionregv1.OperationType{
+			admissionregv1.Create,
+			admissionregv1.Delete,
+		},
+	}
+}
+
+func (e *engineImageValidator) Create(request *admission.Request, newObj runtime.Object) error {
+	engineImage := newObj.(*longhorn.EngineImage)
+	if engineImage.Spec.Image == "" {
```
innobead, line 40: Can we remove `// +optional` from `Image`? so we leave the validation to the OpenAPI scheme itself instead.

webhook/resources/engineimage/validator.go
```diff
@@ -0,0 +1,61 @@
...
+	defaultImage, err := e.ds.GetSettingValueExisted(types.SettingNameDefaultEngineImage)
+	if err != nil {
+		err = errors.Wrap(err, "unable to delete engine image")
+		return werror.NewInvalidError(err.Error(), "")
+	}
+	if engineImage.Spec.Image == defaultImage {
+		return werror.NewInvalidError("unable to delete the default engine image", "")
+	}
+	if engineImage.Status.RefCount != 0 {
+		return werror.NewInvalidError("unable to delete the engine image while being used", "")
```
innobead, line 58: Could we add `refCount` into the message?

webhook/resources/recurringjob/validator.go
```diff
@@ -0,0 +1,92 @@
...
+			admissionregv1.Create,
+			admissionregv1.Update,
+		},
+	}
+}
+
+func (r *recurringJobValidator) Create(request *admission.Request, newObj runtime.Object) error {
+	recurringJob := newObj.(*longhorn.RecurringJob)
+
+	if recurringJob.Spec.Task != longhorn.RecurringJobTypeBackup && recurringJob.Spec.Task != longhorn.RecurringJobTypeSnapshot {
```
innobead, line 42: Can we use `+kubebuilder:validation:Enum` to do schema validation instead?

webhook/resources/recurringjob/validator.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+
+	return nil
+
+}
+
+func (r *recurringJobValidator) Update(request *admission.Request, oldObj runtime.Object, newObj runtime.Object) error {
+	newRecurringJob := newObj.(*longhorn.RecurringJob)
+
+	if newRecurringJob.Spec.Task != longhorn.RecurringJobTypeBackup && newRecurringJob.Spec.Task != longhorn.RecurringJobTypeSnapshot {
```
innobead, line 72: The same here by using schema validation instead.

manager/backingimage.go
```diff
@@ -77,62 +74,6 @@ func (m *VolumeManager) CreateBackingImage(name, checksum, sourceType string, pa
 		return nil, fmt.Errorf("invalid name %v", name)
 	}
```
innobead, line 75: Do we still need this? it seems we already cover in the mutator and validator.

webhook/resources/backingimage/mutator.go
```diff
@@ -35,14 +41,61 @@ func (b *backingImageMutator) Resource() admission.Resource {
...
+	checksum := strings.TrimSpace(backingImage.Spec.Checksum)
+	patchOps = append(patchOps, fmt.Sprintf(`{"op": "replace", "path": "/spec/checksum", "value": "%s"}`, checksum))
+
+	// Handle Spec.SourceParameters
+	parameters := make(map[string]string, 0)
+	for k, v := range backingImage.Spec.SourceParameters {
+		parameters[k] = strings.TrimSpace(v)
+	}
+
+	if longhorn.BackingImageDataSourceType(backingImage.Spec.SourceType) == longhorn.BackingImageDataSourceTypeExportFromVolume {
```
innobead, line 62: BTW, could we add enum schema validation for `SourceType`?

webhook/resources/backingimage/validator.go
```diff
@@ -0,0 +1,105 @@
...
+		if ei.Status.CLIAPIVersion < engineapi.CLIVersionFive {
+			return werror.NewInvalidError(fmt.Sprintf("engine image %v CLI version %v doesn't support this feature, please upgrade engine for volume %v before exporting backing image from the volume", eiName, ei.Status.CLIAPIVersion, volumeName), "")
+		}
+
+		if backingImage.Spec.SourceParameters[manager.DataSourceTypeExportFromVolumeParameterExportType] != manager.DataSourceTypeExportFromVolumeParameterExportTypeRAW &&
+			backingImage.Spec.SourceParameters[manager.DataSourceTypeExportFromVolumeParameterExportType] != manager.DataSourceTypeExportFromVolumeParameterExportTypeQCOW2 {
+			return werror.NewInvalidError(fmt.Sprintf("unsupported export type %v", backingImage.Spec.SourceParameters[manager.DataSourceTypeExportFromVolumeParameterExportType]), "")
+		}
+	default:
+		return werror.NewInvalidError(fmt.Sprintf("unknown backing image source type %v", backingImage.Spec.SourceType), "")
```
innobead, line 88: Suggest using `enum` validation instead, so we don't need to add this default error here.

webhook/resources/backingimage/validator.go
```diff
@@ -0,0 +1,105 @@
...
+
+	return nil
+}
+
+func (b *backingImageValidator) Delete(request *admission.Request, oldObj runtime.Object) error {
+	backingImage := oldObj.(*longhorn.BackingImage)
+
+	replicas, err := b.ds.ListReplicasByBackingImage(backingImage.Name)
+	if err != nil {
+		return werror.NewInvalidError(err.Error(), "")
```
innobead, line 99: Suggest having `cannot delete backing image` as prefix for this error like you have done for others if needed.

innobead (changes requested):

webhook/resources/node/validator.go
derekbit: Yes, we can remove the checks.

webhook/resources/node/validator.go
derekbit: Should be `newNode` here.

webhook/resources/node/validator.go
derekbit: https://github.com/longhorn/longhorn-manager/blob/master/k8s/pkg/apis/longhorn/v1beta2/node.go#L79

Should replace `Minimum=0` with `exclusiveMinimum=0`

derekbit: https://github.com/longhorn/longhorn-manager/blob/master/upgrade/v110to111/upgrade.go#L115
Also need to mutate the `EngineManagerCPURequest` with zero value

webhook/resources/node/validator.go
derekbit: https://github.com/longhorn/longhorn-manager/blob/master/upgrade/v110to111/upgrade.go#L119

manager/backingimage.go
derekbit: Yep, should remove this part.

webhook/resources/backingimage/mutator.go
derekbit: Yep, we should add `enum` for these data types.

webhook/resources/volume/mutator.go
```diff
@@ -35,14 +43,149 @@ func (v *volumeMutator) Resource() admission.Resource {
...
+	patchOps = append(patchOps, fmt.Sprintf(`{"op": "replace", "path": "/spec/size", "value": "%v"}`, strconv.FormatInt(size, 10)))
+
+	defaultEngineImage, _ := v.ds.GetSettingValueExisted(types.SettingNameDefaultEngineImage)
+	if defaultEngineImage == "" {
+		return nil, werror.NewInvalidError("BUG: Invalid empty Setting.EngineImage", "")
+	}
+	patchOps = append(patchOps, fmt.Sprintf(`{"op": "replace", "path": "/spec/engineImage", "value": "%s"}`, defaultEngineImage))
+
+	// Add backup volume name label to the restore/DR volume
+	if volume.Spec.FromBackup != "" {
```
innobead, line 166: Can we move this part as part of the same condition above? `volume.Spec.FromBackup != ""`

deploy/install/02-components/05-webhook.yaml
```diff
@@ -120,6 +120,10 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: metadata.namespace
+        - name: NODE_NAME
```
innobead, line 123: Why we need this?

innobead (changes requested):
LGTM, just left some comments needed to resolve before approval. This is a good start to move out some validation logic from the existing code to the same place.

webhook/resources/engineimage/validator.go
derekbit: `Image` is required field and non-empty sting.
Remove `// +optional` and Add `// +kubebuilder:validation:MinLength:=1`

webhook/resources/node/validator.go
derekbit: https://github.com/longhorn/longhorn-manager/blob/555b5a38f57b1875af37614467623010f4d799b8/controller/controller_manager.go

If `EngineManagerCPURequest` or `ReplicaManagerCPURequest` is 0, the default value will be set.
Thus, `0` is also available value here.

webhook/resources/engineimage/validator.go
innobead: yeah, right now it's based on the error template of openAPI schema validation.

innobead (commented):
LGTM, but as discussed, need to check if enum validation work as expected first before merge.

innobead (approved):

webhook/resources/volume/mutator.go
```diff
@@ -68,5 +206,17 @@ func mutateVolume(newObj runtime.Object) (admission.PatchOps, error) {
 		}
 	}
```
jenting, line 208: Shouldn't we perform the check as lines 80-169 because the user probably changing the Volume CR.
In this case, the mutating webhook of the _Update_ operation would be called, right?

jenting (commented):
Sorry for the late review, a few questions.

webhook/resources/backingimage/validator.go
```diff
@@ -0,0 +1,103 @@
...
+
+func (b *backingImageValidator) Resource() admission.Resource {
+	return admission.Resource{
+		Name:       "backingimages",
+		Scope:      admissionregv1.NamespacedScope,
+		APIGroup:   longhorn.SchemeGroupVersion.Group,
+		APIVersion: longhorn.SchemeGroupVersion.Version,
+		ObjectType: &longhorn.BackingImage{},
+		OperationTypes: []admissionregv1.OperationType{
+			admissionregv1.Create,
```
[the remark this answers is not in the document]
derekbit: `backingImage.update` are called in the HTTP API path and the controllers, so I didn't handle it in this phase.
The validation logics in A and B are different, so need to find the common part and move to the webhook

webhook/resources/engineimage/mutator.go
```diff
@@ -0,0 +1,73 @@
...
+
+func (e *engineImageMutator) Resource() admission.Resource {
+	return admission.Resource{
+		Name:       "engineimages",
+		Scope:      admissionregv1.NamespacedScope,
+		APIGroup:   longhorn.SchemeGroupVersion.Group,
+		APIVersion: longhorn.SchemeGroupVersion.Version,
+		ObjectType: &longhorn.EngineImage{},
+		OperationTypes: []admissionregv1.OperationType{
+			admissionregv1.Create,
```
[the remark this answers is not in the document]
derekbit: Basically, nothing can be moved to `engineImage.Update` or `engineImage.UpdateStatus`..

webhook/resources/node/validator.go
```diff
@@ -0,0 +1,144 @@
...
+
+func (n *nodeValidator) Resource() admission.Resource {
+	return admission.Resource{
+		Name:       "nodes",
+		Scope:      admissionregv1.NamespacedScope,
+		APIGroup:   longhorn.SchemeGroupVersion.Group,
+		APIVersion: longhorn.SchemeGroupVersion.Version,
+		ObjectType: &longhorn.Node{},
+		OperationTypes: []admissionregv1.OperationType{
+			admissionregv1.Update,
```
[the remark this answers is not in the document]
derekbit: Yep. We probably can implement the `Create`.

webhook/resources/volume/mutator.go
```diff
@@ -35,14 +43,144 @@ func (v *volumeMutator) Resource() admission.Resource {
...
+		numberOfReplicas, err := v.getDefaultReplicaCount()
+		if err != nil {
+			err = errors.Wrap(err, "BUG: cannot get valid number for setting default replica count")
+			return nil, werror.NewInvalidError(err.Error(), "")
+		}
+		logrus.Infof("Use the default number of replicas %v", numberOfReplicas)
+		patchOps = append(patchOps, fmt.Sprintf(`{"op": "replace", "path": "/spec/numberOfReplicas", "value": %v}`, numberOfReplicas))
+	}
+
+	if string(volume.Spec.ReplicaAutoBalance) == "" {
```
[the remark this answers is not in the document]
derekbit: Yep, duplicate.

webhook/resources/volume/mutator.go
derekbit: Ideally, we should check as line 80-169 in `Update`, but there are some exclusive validations between `Update` operates. As [1], the next step is to find out more common and general checks and move them to the webhook.

webhook/resources/node/validator.go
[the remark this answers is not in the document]
derekbit: The `Node.Create` are mainly related to the mutation, so the node validation does not have `Create` function.

webhook/resources/engineimage/mutator.go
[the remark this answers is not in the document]
jenting: Okay, I was thinking we could mutate the `spec.image`, but this would cause the EngineImage CR name to be changed.

webhook/resources/node/validator.go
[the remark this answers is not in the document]
jenting: If I understand correctly, we don't have to validate the new Node CR such as the `spec.engineManagerCPURequest` and `spec.replicaManagerCPURequest` is valid or not?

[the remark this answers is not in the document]
derekbit: https://github.com/longhorn/longhorn-manager/blob/master/datastore/longhorn.go#L1767
I'm thinking if we should move this part to the mutation function.
If the node.disks is not set, then set the default disk.
But the `disks` is not a MUST set field.

@jenting, @innobead WDYT?

webhook/resources/volume/mutator.go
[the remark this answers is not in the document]
jenting: Should we remove it?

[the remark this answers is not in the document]
derekbit: Yep, will send a PR to remove it.

webhook/resources/volume/mutator.go
jenting: I see, thank you for the effort.

webhook/resources/node/validator.go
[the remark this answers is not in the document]
derekbit: > `spec.replicaManagerCPURequest`

Yes. But some of the values are validated by schema validation.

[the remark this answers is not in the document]
jenting: Do you mean this part? https://github.com/longhorn/longhorn-manager/blob/3d7e218bc74e1966360bb1aa311b9eee87684efe/datastore/longhorn.go#L1786-L1801

webhook/resources/backingimage/validator.go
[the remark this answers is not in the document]
jenting: I see, thank you for the effort.

webhook/resources/node/validator.go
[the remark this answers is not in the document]
derekbit: > Do you mean this part?
> 
> https://github.com/longhorn/longhorn-manager/blob/3d7e218bc74e1966360bb1aa311b9eee87684efe/datastore/longhorn.go#L1786-L1801

Yep!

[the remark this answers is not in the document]
jenting: > I'm thinking if we should move this part to the mutation function.
If the node.disks is not set, then set the default disk.

From the comment https://github.com/longhorn/longhorn-manager/blob/3d7e218bc74e1966360bb1aa311b9eee87684efe/datastore/longhorn.go#L1789-L1791
I think we could add to Create mutator only, but not in Update mutator.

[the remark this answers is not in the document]
derekbit: Just aware of the logic including `disks, err := types.CreateDefaultDisk(dataPath) `. 
Thus, we cannot move it to the admission webhook, or the `dataPath` will be created on a wrong node.
`initDaemonNode`, `CreateDefaultNode` and `CreateDefaultDisk` will be executed while starting the longhorn-manager daemon.

webhook/resources/engineimage/mutator.go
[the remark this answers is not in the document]
derekbit: Gotcha. Normalizing the `spec.image` and mutating it is good idea, but the previous logic does not include it. Thus, I did't do it to avoid unexpected side effects.

webhook/resources/node/validator.go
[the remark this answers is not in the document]
jenting: I see

> Just aware of the logic including `disks, err := types.CreateDefaultDisk(dataPath) `. Thus, we cannot move it to the admission webhook, or the `dataPath` will be created on a wrong node.
> 
> Currently, `initDaemonNode`, `CreateDefaultNode` and `CreateDefaultDisk` will be executed while starting the longhorn-manager daemon.

I see.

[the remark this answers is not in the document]
derekbit: > I see
> 
> > Just aware of the logic including `disks, err := types.CreateDefaultDisk(dataPath) `. Thus, we cannot move it to the admission webhook, or the `dataPath` will be created on a wrong node.
> > Currently, `initDaemonNode`, `CreateDefaultNode` and `CreateDefaultDisk` will be executed while starting the longhorn-manager daemon.
> 
> I see.

Leaving a note here.
The admission webhook is responsible for the parameters validation and mutation. Any manipulation of the resources of a node should not be allowed, and the manipulations should handle in the longhorn-manager daemon or controllers.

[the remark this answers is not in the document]
innobead: @derekbit as discussed, suggest having a note in that package like readme, so it's better to know what kind of code should be put there.

PR #1229, opened 2022-02-22
Enqueue volume for replica-auto-balance on node change

c3y1huang:
https://github.com/longhorn/longhorn/issues/3649

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -3489,12 +3489,16 @@ func (vc *VolumeController) enqueueNodeChange(obj interface{}) {
 		return
 	}
 	for _, r := range replicas {
-		if r.Spec.NodeID == "" || r.Spec.FailedAt != "" {
-			vol, err := vc.ds.GetVolumeRO(r.Spec.VolumeName)
-			if err != nil {
-				utilruntime.HandleError(fmt.Errorf("failed to get volume %v of replica %v when enqueuing node %v: %v", r.Spec.VolumeName, r.Name, node.Name, err))
-				continue
-			}
+		vol, err := vc.ds.GetVolumeRO(r.Spec.VolumeName)
+		if err != nil {
+			utilruntime.HandleError(fmt.Errorf("failed to get volume %v of replica %v when enqueuing node %v: %v", r.Spec.VolumeName, r.Name, node.Name, err))
+			continue
+		}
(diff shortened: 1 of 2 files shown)


c3y1huang:
Test result - https://ci.longhorn.io/job/private/job/longhorn-tests-regression/724/

controller/volume_controller.go
```diff
@@ -3489,12 +3489,16 @@ func (vc *VolumeController) enqueueNodeChange(obj interface{}) {
...
+		vol, err := vc.ds.GetVolumeRO(r.Spec.VolumeName)
+		if err != nil {
+			utilruntime.HandleError(fmt.Errorf("failed to get volume %v of replica %v when enqueuing node %v: %v", r.Spec.VolumeName, r.Name, node.Name, err))
+			continue
+		}
+		replicaAutoBalance, err := vc.getAutoBalancedReplicasSetting(vol)
+		if err != nil {
+			vc.logger.Warnf(err.Error())
+		}
+		if r.Spec.NodeID == "" || r.Spec.FailedAt != "" || replicaAutoBalance != longhorn.ReplicaAutoBalanceDisabled {
```
innobead, line 3501: I assume `replicaAutoBalance` will not be empty string, otherwise we need to ensure the check.

innobead (approved):
LGTM. This is an important fix. 

We need to backport to 1.2.x.

controller/volume_controller.go
c3y1huang: It will not be an empty string. Handled in [getAutoBalancedReplicasSetting](https://github.com/longhorn/longhorn-manager/blob/2387d963b7868493dc1f64dd9b3702ef1a6b3f44/controller/volume_controller.go#L903-L927).

c3y1huang:
> We need to backport to 1.2.x.

Are we adding node informer to volume controller on 1.2.x?
It seems resync had been reverted - https://github.com/longhorn/longhorn-manager/pull/1221

innobead:
> > We need to backport to 1.2.x.
> 
> Are we adding node informer to volume controller on 1.2.x? It seems resync had been reverted - #1221 and also discussed in [#1219 (comment)](https://github.com/longhorn/longhorn-manager/pull/1219#issuecomment-1040945317).

Nope, this is new code, so we don't need to backport.

PR #1230, opened 2022-02-24
Bump backing-image-manager

jenting:
https://github.com/longhorn/longhorn/issues/3343


engineapi/backing_image_manager.go
```diff
@@ -12,8 +12,8 @@ import (
 const (
 	BackingImageManagerDefaultPort = 8000
 
-	CurrentBackingImageManagerAPIVersion = 2
-	MinBackingImageManagerAPIVersion     = 2
+	CurrentBackingImageManagerAPIVersion = 3
```
innobead, line 15: Need to update the default backing image in the manifest as well. Also, create a release of https://github.com/longhorn/backing-image-manager and use that image as default. 

After that, update longhorn/longhorn as well.

innobead (changes requested):

engineapi/backing_image_manager.go
innobead: https://github.com/longhorn/backing-image-manager/releases/tag/v3_20220318

innobead (approved):

innobead (approved):

PR #1231, opened 2022-02-24
Using user-friendly name for the `type` parameter in VolumeSnapshotClass

PhanLe1010:
The type parameter in the `VolumeSnapshotClass ` will be either `snapshot`, `backup`, or `""` after this change.
The ID in the `VolumeSnapshotContent ` will be either `ss`, `bs`.

longhorn/longhorn#2534

### DIFF
--- a/csi/controller_server.go
+++ b/csi/controller_server.go
@@ -32,8 +32,10 @@ const (
 	tickBackupInitiation    = 5 * time.Second
 	backupStateCompleted    = "Completed"
 
-	csiSnapshotTypeLonghornSnapshot = "ss"
-	csiSnapshotTypeLonghornBackup   = "bs"
+	csiSnapshotTypeLonghornSnapshotLongName = "snapshot"
+	csiSnapshotTypeLonghornBackupLongName   = "backup"
+	csiSnapshotTypeLonghornSnapshot         = "ss"
+	csiSnapshotTypeLonghornBackup           = "bs"
 )
 
 type ControllerServer struct {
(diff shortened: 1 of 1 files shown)


csi/controller_server.go
```diff
@@ -32,8 +32,10 @@ const (
 	tickBackupInitiation    = 5 * time.Second
 	backupStateCompleted    = "Completed"
 
-	csiSnapshotTypeLonghornSnapshot = "ss"
-	csiSnapshotTypeLonghornBackup   = "bs"
+	csiSnapshotTypeLonghornSnapshotLongName = "snapshot"
+	csiSnapshotTypeLonghornBackupLongName   = "backup"
+	csiSnapshotTypeLonghornSnapshot         = "ss"
+	csiSnapshotTypeLonghornBackup           = "bs"
```
shuo-wu, line 38: Sorry. I was thinking we will use `snap/bak` in both CSI snapshot and Longhorn volume... 
Using 2 kinds of expression is confusing.

PhanLe1010: We already use `bs` in the csi snapshot: https://longhorn.io/docs/1.2.3/snapshots-and-backups/csi-snapshot-support/restore-a-backup-via-csi/#restore-a-backup-that-has-no-associated-volumesnapshot

jenting: But the current `bs` is for `VolumeSnapshotContent`.

For `VolumeSnapshotClass`, I realized we should change from `bs/ss` to `snap/bak` or `snapshot/backup`.

shuo-wu: We can add backward compatible code for the previous implementation. e.g.,
```go
 if csiSnapshotType == "" || csiSnapshotType == "bs" || csiSnapshotType == csiSnapshotTypeLonghornBackupLongName {
	// For backward compatibility, type "" or "bs" is considered as csiSnapshotTypeLonghornBackupLongName
	rsp, err = cs.createCSISnapshotTypeLonghornBackup(req)
}
```

innobead: it's better to have a function/method wrapper for the backward compatibility condition check instead of an inline condition

jenting:
@PhanLe1010 I believe you push to the wrong branch.

csi/controller_server.go
PhanLe1010: After discussing with the team, it seems that there were some misunderstandings about how we would agree on this fix.  Therefore, I have prepare 2 PRs for the sake of clarity:
* The first approach is this PR
* The second approach is the PR https://github.com/longhorn/longhorn-manager/pull/1240

Please let me know what you think? @joshimoo @shuo-wu @jenting @innobead

PhanLe1010:
pushed back the correct content

innobead:
Let's use https://github.com/longhorn/longhorn-manager/pull/1240 instead. Closing this.

PR #1232, opened 2022-02-24
backup: retry deletion after failing to acquire the lock

derekbit:
The resync period of the backingimage is one hour and the maxRetries is
3. Thus, the deletion of the backup in error state caused by the
shutdown of the replica during backing up will fail.

The workaround is to increase the maxRetries number and to retry the
deletion until the lock acquisition of the backup is timeout after 150
seconds.

maxRetriesOnAcquireLockError should guarantee the cumulative retry time
is larger than 150 seconds.

Longhorn 3620

### DIFF
--- a/controller/backup_controller.go
+++ b/controller/backup_controller.go
@@ -32,6 +32,16 @@ import (
 	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
 )
 
+var (
+	// maxRetriesOnAcquireLockError should guarantee the cumulative retry time
+	// is larger than 150 seconds.
+	// With the current rate-limiter in use (5ms*2^(maxRetries-1)) the following numbers represent the times
+	// a deployment is going to be requeued:
+	//
+	// 5ms, 10ms, 20ms, ... , 81.92s, 163.84s
+	maxRetriesOnAcquireLockError = 16
+)
+
 type BackupController struct {
 	*baseController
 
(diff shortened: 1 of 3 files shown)


controller/base_controller.go
```diff
@@ -6,6 +6,17 @@ import (
...
 
+var (
+	// maxRetries is the number of times a deployment will be retried before it is dropped out of the queue.
+	// With the current rate-limiter in use (5ms*2^(maxRetries-1)) the following numbers represent the times
+	// a deployment is going to be requeued:
+	//
+	// 5ms, 10ms, 20ms
+	maxRetries = 3
+
+	maxRetriesOnAcquireLockError = 16
```
jenting, line 17: It's a customized threshold for backup_controller only. It'd better put in backup_controller.go

controller/backup_controller.go
```diff
@@ -142,10 +142,18 @@ func (bc *BackupController) handleErr(err error, key interface{}) {
 		return
 	}
```
jenting, line 144: It's better to have a comment on _why_ the backup_controller catch the "failed lock" and perform the special logic.

controller/backup_controller.go
```diff
@@ -32,6 +32,14 @@ import (
 	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
 )
 
+var (
+	// maxRetriesOnAcquireLockError should guarantee the cumulative retry time
+	// is larger than 150 seconds.
+	//
+	// 5ms, 10ms, 20ms, ... , 81.92s, 163.84s
```
jenting, line 39: Could you add the math function on how it is calculated? `(5ms*2^(maxRetries-1))`

controller/backup_controller.go
```diff
@@ -142,10 +150,23 @@ func (bc *BackupController) handleErr(err error, key interface{}) {
 		return
 	}
 
-	if bc.queue.NumRequeues(key) < maxRetries {
-		bc.logger.WithError(err).Warnf("Error syncing Longhorn backup %v", key)
-		bc.queue.AddRateLimited(key)
-		return
+	// The resync period of the backingimage is one hour and the maxRetries is 3.
```
jenting, line 153: Not backingimage, it's backup.

controller/backup_controller.go
```diff
@@ -142,10 +150,23 @@ func (bc *BackupController) handleErr(err error, key interface{}) {
...
+	// Thus, the deletion failure of the backup in error state is caused by the shutdown of the replica during backing up,
+	// if the lock hold by the backup job is not released.
+	// The workaround is to increase the maxRetries number and to retry the deletion until the lock acquisition
+	// of the backup is timeout after 150 seconds.
+	if strings.Contains(err.Error(), "failed lock") {
+		if bc.queue.NumRequeues(key) < maxRetriesOnAcquireLockError {
+			bc.logger.WithError(err).Warnf("Error syncing Longhorn backup %v", key)
```
jenting, line 160: Do you think we should be able to differentiate the log message of line 160 and line 166?

derekbit: Better. I've updated the error messages.

controller/backup_controller.go
derekbit: Okay. added.

shuo-wu (approved):

jenting (approved):

innobead (approved):

PR #1234, opened 2022-02-25
Add a field to record the longhorn engine generated backup name

jenting:
#### Propose Change

The longhorn manager >= 1.2.x + longhorn engine >= 1.2.0, the backup CR name is generated in longhorn manager and pass to the longhorn engine.
The longhorn manager >= 1.2.x + longhorn engine < 1.2.0, the backup CR name is generated in longhorn manager but the longhorn engine would generate another name.
To keep backward compatibility of longhorn manager >= 1.2.x + longhorn engine < 1.2.0, stores the longhorn engine reports backup name as `engineBackupName`, and use this name to query the longhorn engine to monitor the snapshot backup process.
    
Note that, with longhorn manager >= 1.2.x + longhorn engine < 1.2.0, the backup CR would be replaced with a newer one (with the name generated by the longhorn engine) after the snapshot backup is done and synced from the remote backup target.

#### Issue
https://github.com/longhorn/longhorn/issues/3621


engineapi/backup_monitor.go
```diff
@@ -27,6 +27,13 @@ type BackupMonitor struct {
 	replicaAddress string
 	engineClient   EngineClient
 
+	// The longhorn engine < 1.2.0, the backup name is generated within it.
```
shuo-wu, line 30: For enigneImage < 1.2.0, how does Longhorn set the backup CR name? Does this mean that the CR name is different from the name generated by the longhorn engine?

jenting: For longhorn engine < 1.2.0, the backup name is generated in longhorn engine.
Yes, the Backup CR name is different to the longhorn engine < 1.2.0 generated name.

shuo-wu: Then will [this code](https://github.com/longhorn/longhorn-manager/blob/b8e67dc8112c397089041b11207ec6fa5155ac6d/controller/backup_volume_controller.go#L290-L306) create one more new backup CR in this case?

jenting: Yes, it will have 2 Backup CRs (intermediate transition state) if the snapshot backup hasn't been completed.
After the snapshot backup is completed and the backup_volume_controller syncs the remote backup target, in the end, the Backup CR would be 1.

shuo-wu: In brief, the backup CR created by users is just for triggering the operation as well as presenting the progress, right? And once the backup is done, this CR will be replaced by a newer one with the name generated by longhorn-engine.

jenting: > In brief, the backup CR created by users is just for triggering the operation as well as presenting the progress, right? And once the backup is done, this CR will be replaced by a newer one with the name generated by longhorn-engine.

Right. This is longhorn-manager >= 1.2.0 + longhorn-engine < 1.2.0 backup behavior.

engineapi/backup_monitor.go
```diff
@@ -59,14 +66,15 @@ func NewBackupMonitor(logger logrus.FieldLogger,
 
 	// Call engine API snapshot backup
 	if backup.Status.State == longhorn.BackupStateNew {
-		_, replicaAddress, err := engineClient.SnapshotBackup(backup.Name, backup.Spec.SnapshotName,
+		engineBackupName, replicaAddress, err := engineClient.SnapshotBackup(backup.Name, backup.Spec.SnapshotName,
```
innobead, line 69: It's good to see we already have backward compatibility for this, so we can have this simple solution here.

innobead (approved):
LGTM

engineapi/backup_monitor.go
shuo-wu: Then after the backup creation is complete, users cannot check the CR via `kubectl -n longhorn-system get lhb <the old backup name>` anymore, right? This may confuse users...

jenting: Yes, unless we ask the user to upgrade the engine image to 1.2.x.

However, in the end, the single source of truth would be the remote backup target.

innobead: This issue seems still explanatory but do we have a note for this in 1.2.x? If not, probably we should have somewhere in the upgrade section to avoid some questions in the future.

jenting: We have a note at the [doc](https://longhorn.io/docs/1.2.3/snapshots-and-backups/backup-and-restore/).

shuo-wu: OK. Since we already applied it with the explanation in the doc, I am fine with this now. But please add more comments to explain this CR replacement behavior here.

shuo-wu (commented):
After adding the comment, this PR is good to go.

innobead (approved):

PR #1240, opened 2022-02-28
Using user-friendly name for the csiSnapshotType

PhanLe1010:
The values will be either `snap`, `bak`,or `""` after this change.

The benefit of this approach is that both `type` parameter in `VolumeSnapshotClass` and the snapshot ID in `VolumeSnapshotContent` will have the same value (one of the 3 possible values: `snap`, `bak`,or `""`)

longhorn/longhorn#2534


shuo-wu (approved):
I prefer this solution

innobead (approved):

PR #1243, opened 2022-03-01
Reducing the workqueue depth while still keeping the resync period as 30s

PhanLe1010:
For each controller, we only need to enable resync for the handler
that handles the events of its managed object. For example, the
volume controller only needs to enable resync for volumeInformer
event handler. By doing this, we reduce the workqueue depth of the
controller while still making sure that we resync every object
of the controller every resync period (30s)

longhorn/longhorn#3083


PhanLe1010:
Push the PR for discussion with the team. Doing more testing on this PR. Please hold on merging

innobead:
This is creative, and actually, from the logic and design convention perspective, this looks reasonable, but have concerns about the unknown part and the change in the patch release, 1.2.x.  Need more reviews.

controller/backing_image_controller.go
```diff
@@ -88,23 +88,23 @@ func NewBackingImageController(
 		DeleteFunc: bic.enqueueBackingImage,
 	})
 
-	backingImageManagerInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+	backingImageManagerInformer.Informer().AddEventHandlerWithResyncPeriod(cache.ResourceEventHandlerFuncs{
```
shuo-wu, line 91: Is this resync period only applied to these specific handlers? If YES, that would be great.

PhanLe1010: > Yes, it is only applied to these specific handlers.

Some info from the code path:

1. Each AddEventHandlerWithResyncPeriod is adding a process listener to the sharedProcessor: [link](https://github.com/longhorn/longhorn-manager/blob/ad25434856d45c92af5dcbb4acd2e4413a4dd508/vendor/k8s.io/client-go/tools/cache/shared_informer.go#L476)
1. The sharedProcessor collects all the listeners that need to do resync and put them into syncingListeners: [link](https://github.com/longhorn/longhorn-manager/blob/ad25434856d45c92af5dcbb4acd2e4413a4dd508/vendor/k8s.io/client-go/tools/cache/shared_informer.go#L610-L622). The listeners that need to do resync are the ones with resyncPeriod > 0: [link](https://github.com/longhorn/longhorn-manager/blob/ad25434856d45c92af5dcbb4acd2e4413a4dd508/vendor/k8s.io/client-go/tools/cache/shared_informer.go#L762)
1. When it is the time to resync, the share sharedProcessor distribute the event to the listener that want to resync only: [link](https://github.com/longhorn/longhorn-manager/blob/ad25434856d45c92af5dcbb4acd2e4413a4dd508/vendor/k8s.io/client-go/tools/cache/shared_informer.go#L570-L583)

controller/node_controller.go
```diff
@@ -105,50 +103,45 @@ func NewNodeController(
 
 	nc.scheduler = scheduler.NewReplicaScheduler(ds)
 
-	// We want to check the real time usage of disk on nodes.
-	// Therefore, we add a small resync for the NodeInformer here
-	nodeInformer.Informer().AddEventHandlerWithResyncPeriod(cache.ResourceEventHandlerFuncs{
+	nodeInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
```
jenting, line 106: I think we could leave a comment to explain _why_ the node informer uses `AddEventHandler` rather than `AddEventHandlerWithResyncPeriod`.

jenting:
If I understand correctly, only the informers with `AddEventHandler` keeps the resync period as 30 seconds?

shuo-wu (approved):

PhanLe1010:
> If I understand correctly, only the informers with AddEventHandler keeps the resync period as 30 seconds?

Yes, it is correct

controller/node_controller.go
PhanLe1010: Agree. Added back the comment

PhanLe1010: Actually, I think we dont need the comment because it is similar to other controller: the node controller will resync node CR every 30s

jenting: I see, I just want to have a comment/history somewhere to explain why this one does not use `AddEventHandlerWithResyncPeriod`.

shuo-wu: All developers can get the details by checking the git annotation/commit comment.

jenting (approved):

innobead (approved):
LGTM. 

Originally, I am worried about the uncertain part, but right now the cascading eventing resolves my concerns. Because the resync responsibility belongs to the original resource controller, if any resource resync there, it will be cascading to inform the informer in other resource controllers which watch the changes.

PR #1245, opened 2022-03-02
Apply non-root user to longhorn-ui by adopting the change from https:…

innobead:
Apply non-root user to longhorn-ui by adopting the change from https://github.com/longhorn/longhorn/pull/3662


innobead:
Need to backport to 1.2.x.

shuo-wu (approved):

PR #1248, opened 2022-03-03
Fix upgrade failure because of the wrong conversion strategy 

derekbit:
Fix upgrade failure because of the wrong conversion strategy in crd.yaml.
The conversion strategy and webhook settings are patched while generating the CRD manifest, and the `caBundle` will be dynamically generated in the conversion webhook.

Longrhorn 3631

### DIFF
--- a/deploy/install/01-prerequisite/03-crd.yaml
+++ b/deploy/install/01-prerequisite/03-crd.yaml
@@ -335,11 +335,22 @@ kind: CustomResourceDefinition
 metadata:
   annotations:
     controller-gen.kubebuilder.io/version: v0.7.0
-  creationTimestamp: null
   labels:
     longhorn-manager: ""
   name: backingimages.longhorn.io
 spec:
+  conversion:
+    strategy: Webhook
+    webhook:
+      clientConfig:
+        service:
+          name: longhorn-conversion-webhook
+          namespace: longhorn-system
+          path: /v1/webhook/conversion
+          port: 9443
+      conversionReviewVersions:
+      - v1beta2
+      - v1beta1
   group: longhorn.io
   names:
     kind: BackingImage
@@ -675,11 +686,22 @@ kind: CustomResourceDefinition
 metadata:
   annotations:
     controller-gen.kubebuilder.io/version: v0.7.0
-  creationTimestamp: null
   labels:
     longhorn-manager: ""
   name: backuptargets.longhorn.io
 spec:
+  conversion:
+    strategy: Webhook
+    webhook:
+      clientConfig:
+        service:
+          name: longhorn-conversion-webhook
+          namespace: longhorn-system
+          path: /v1/webhook/conversion
+          port: 9443
+      conversionReviewVersions:
+      - v1beta2
+      - v1beta1
   group: longhorn.io
   names:
     kind: BackupTarget
@@ -1002,11 +1024,22 @@ kind: CustomResourceDefinition
 metadata:
   annotations:
     controller-gen.kubebuilder.io/version: v0.7.0
-  creationTimestamp: null
   labels:
     longhorn-manager: ""
   name: engineimages.longhorn.io
 spec:
+  conversion:
+    strategy: Webhook
+    webhook:
+      clientConfig:
+        service:
+          name: longhorn-conversion-webhook
+          namespace: longhorn-system
+          path: /v1/webhook/conversion
+          port: 9443
+      conversionReviewVersions:
+      - v1beta2
+      - v1beta1
   group: longhorn.io
   names:
     kind: EngineImage
@@ -1651,11 +1684,22 @@ kind: CustomResourceDefinition
 metadata:
   annotations:
     controller-gen.kubebuilder.io/version: v0.7.0
-  creationTimestamp: null
   labels:
     longhorn-manager: ""
   name: nodes.longhorn.io
 spec:
+  conversion:
+    strategy: Webhook
+    webhook:
+      clientConfig:
+        service:
+          name: longhorn-conversion-webhook
+          namespace: longhorn-system
+          path: /v1/webhook/conversion
+          port: 9443
+      conversionReviewVersions:
+      - v1beta2
+      - v1beta1
   group: longhorn.io
   names:
     kind: Node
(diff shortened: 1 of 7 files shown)


innobead (commented):
LGTM, the only thing is to keep the patch manifests clean w/ necessary settings, and other dynamic parts will be updated after the conversation webhook starts, so no need to add to the patches.

Well done, good teamwork!

innobead:
> Although we recommend the user to install/deploy Longhorn to namespace `longhorn-system`. However, we did see users install/deploy Longhorn to other namespaces.
> 
> But since this PR force the namespace to be `longhorn-system`, therefore this would make the conversion webhook server unreachable.

The namespace part should be variable when updating to longhorn/longhorn? in this repo, it should be fine as others.

derekbit:
> Although we recommend the user to install/deploy Longhorn to namespace `longhorn-system`. However, we did see users install/deploy Longhorn to other namespaces.
> 
> But since this PR force the namespace to be `longhorn-system`, therefore this would make the conversion webhook server unreachable.

A bit of curious. Other manifests also use the `longhorn-system` namespace. The upgrade of components by applying the all manifests also fails, correct?

jenting (dismissed):

PhanLe1010:
Discussed with @derekbit. Our webhook is still modifying the user deployed manifest (the caBundle field).  @derekbit Will do a test with flux to see if it has any problem

shuo-wu (commented):
Shouldn't we add field `conversion` for each CRD? 
Besides, is there an automatic way to generate the patch for `conversion`?

derekbit:
> Shouldn't we add field `conversion` for each CRD?

No, we don't need to add the conversion to  the CRD with unchanged schema.

> Besides, is there an automatic way to generate the patch for `conversion`?

After investigating the `controller-gen` and related tools, there is no automatic way to generate the `conversion`.
Pretty handy now...

shuo-wu:
> After investigating the `controller-gen` and related tools, there is no automatic way to generate the `conversion`...

OK. Then we need to add or modify the patches manually besides running the script each time we decide to add a new CRD or bump the API version...

shuo-wu (approved):

innobead (approved):

derekbit:
> Discussed with @derekbit. Our webhook is still modifying the user deployed manifest (the caBundle field). @derekbit Will do a test with flux to see if it has any problem

Steps:
1. Setup `flux2` and install the longhorn `v1.2.3` on my Kubernetes cluster by following the [steps](https://fluxcd.io/docs/get-started/).

2. The longhorn cluster is created and ready
  ```
  root@rancher50-master:~/crd_webhook/longhorn# flux get kustomizations
  NAME       	READY	MESSAGE                         	REVISION      	SUSPENDED
  flux-system	True 	Applied revision: master/b0616dc	master/b0616dc	False
  longhorn   	True 	Applied revision: v1.2.3/96a7afb	v1.2.3/96a7afb	False
  ```

3. Update the branch to `master` having this PR's commits, and then Longhorn cluster is upgraded. The pods of Longhorn works well and the flux does not pop up any error message.
  ```
  root@rancher50-master:~/longhorn-flux2# flux get kustomizations
  NAME       	READY	MESSAGE                          	REVISION       	SUSPENDED
  flux-system	True 	Applied revision: master/74d4b6b 	master/74d4b6b 	False
  longhorn   	True 	Applied revision: upgrade/685944c	upgrade/685944c	False
  ```
  
The `caBundle` of the CRDs is set correctly by the webhook server.
```
root@rancher50-master:~/longhorn-flux2# kl get crds backingimages.longhorn.io -o yaml
metadata:
  annotations:
    controller-gen.kubebuilder.io/version: v0.7.0
  creationTimestamp: "2022-03-04T12:08:32Z"
  generation: 3
  labels:
    kustomize.toolkit.fluxcd.io/name: longhorn
    kustomize.toolkit.fluxcd.io/namespace: flux-system
    longhorn-manager: ""
  name: backingimages.longhorn.io
  resourceVersion: "7923"
  uid: be76419b-55b8-48ff-a9a5-a907c8575ada
spec:
  conversion:
    strategy: Webhook
    webhook:
      clientConfig:
        caBundle: LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0tCk1JSUJpRENDQVM2Z0F3SUJBZ0lCQURBS0JnZ3Foa2pPUFFRREFqQTdNUnd3R2dZRFZRUUtFeE5rZVc1aGJXbGoKYkdsemRHVnVaWEl0YjNKbk1Sc3dHUVlEVlFRREV4SmtlVzVoYldsamJHbHpkR1Z1WlhJdFkyRXdIaGNOTWpJdwpNekEwTVRJeE56SXhXaGNOTXpJd016QXhNVEl4TnpJeFdqQTdNUnd3R2dZRFZRUUtFeE5rZVc1aGJXbGpiR2x6CmRHVnVaWEl0YjNKbk1Sc3dHUVlEVlFRREV4SmtlVzVoYldsamJHbHpkR1Z1WlhJdFkyRXdXVEFUQmdjcWhrak8KUFFJQkJnZ3Foa2pPUFFNQkJ3TkNBQVFWcXp5UDVqNTVuNkRRNmJyZy9ZVk10Zk53M2pqNGxIWjI0dkcvUzFLVwpDWjdkdDR6UkR0L0FUNExyWlRjbUR2L2tBZWFyT3VwdFVTZmZUSHZZWGFmSW95TXdJVEFPQmdOVkhROEJBZjhFCkJBTUNBcVF3RHdZRFZSMFRBUUgvQkFVd0F3RUIvekFLQmdncWhrak9QUVFEQWdOSUFEQkZBaUVBNEpQMWIzRWgKZE0wQStaWW11em9ob3IveW8xNGdpRzdpMlFjRndUT0dOdjBDSUhSWktzcTBjMEhwZ0F5YzdGMEVta2F3d1JKZApnVnFFWGt1TldZeUMzR09GCi0tLS0tRU5EIENFUlRJRklDQVRFLS0tLS0K
        service:
          name: longhorn-conversion-webhook
          namespace: longhorn-system
          path: /v1/webhook/conversion
          port: 9443
      conversionReviewVersions:
```

cc. @jenting @shuo-wu @innobead

PR #1252, opened 2022-03-03
Records engine backup name in Backup CR status

jenting:
The longhorn engine < 1.2.0, the backup name is generated within the longhorn engine. Therefore, the longhorn manager >= 1.2.3 generated Backup CR is not the same as the longhorn engine < 1.2.0 generated backup name.

To support longhorn manager >= 1.2.3 backwards compatible to longhorn engine < 1.2.0, records the engine generated backup name in Backup CR. Then, use this name to query the longhorn engine backup status to make sure both the backward compatible and forward compatible.

Note that, the Backup CR status `engineBackupName` could be deprecated once the longhorn manager no longhorn supports longhorn engine < 1.2.0.

https://github.com/longhorn/longhorn/issues/3621

### DIFF
--- a/engineapi/backup_monitor.go
+++ b/engineapi/backup_monitor.go
@@ -21,16 +21,17 @@ const (
 type BackupMonitor struct {
 	logger logrus.FieldLogger
 
-	namespace      string
-	backupName     string
-	snapshotName   string
-	replicaAddress string
-	engineClient   EngineClient
+	backupName       string
+	engineBackupName string
+	snapshotName     string
+	replicaAddress   string
+	engineClient     EngineClient
 
 	backupStatus     longhorn.BackupStatus
 	backupStatusLock sync.RWMutex
 
 	syncCallback func(key string)
+	callbackKey  string
 
 	ctx  context.Context
 	quit context.CancelFunc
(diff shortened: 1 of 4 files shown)


engineapi/backup_monitor.go
```diff
@@ -70,19 +70,17 @@ func NewBackupMonitor(logger logrus.FieldLogger,
 	}
 
 	// Call engine API snapshot backup
-	if backup.Status.State == longhorn.BackupStateNew {
```
shuo-wu, line 73: I don't understand why we need to remove this check.
- When the backup creation is not started yet, this check won't block the creation.
- When the backup creation is already in progress, the below invocation `engineClient.SnapshotBackup` would return [a (locking) error](https://github.com/longhorn/longhorn-engine/blob/cda02873ef0789a56fcca9ffcdbc5bea76f09018/vendor/github.com/longhorn/backupstore/deltablock.go#L125) during [the backup init](https://github.com/longhorn/longhorn-engine/blob/cda02873ef0789a56fcca9ffcdbc5bea76f09018/pkg/backup/main.go#L127). Since the 1st step of the backup creation is [picking up a random replica](https://github.com/longhorn/longhorn-engine/blob/cda02873ef0789a56fcca9ffcdbc5bea76f09018/pkg/sync/backup.go#L60-L65) rather than checking all replicas and see if there is a replica processing the creation.

PR #1256, opened 2022-03-04
Fix: Longhorn doesn't requeue the volume when failed to schedule or fail to add new replica

PhanLe1010:
We should requeue the volume when failed to schedule or fail
to add new replica. By doing this, the volume controller can
try those operations again. This is a quick fix. Long-term
fix may be separating the replica replenishment and scheduling
logic into a different controller

longhorn/longhorn#3626

### DIFF
--- a/scheduler/replica_scheduler.go
+++ b/scheduler/replica_scheduler.go
@@ -407,18 +407,20 @@ func (rcs *ReplicaScheduler) CheckAndReuseFailedReplica(replicas map[string]*lon
 }
 
 // RequireNewReplica is used to check if creating new replica immediately is necessary **after a reusable failed replica is not found**.
+// If creating new replica immediately is necessary, returns 0.
+// Otherwise, returns the duration that the caller should recheck.
 // A new replica needs to be created when:
 //   1. the volume is a new volume (volume.Status.Robustness is Empty)
 //   2. data locality is required (hardNodeAffinity is not Empty and volume.Status.Robustness is Healthy)
 //   3. replica eviction happens (volume.Status.Robustness is Healthy)
 //   4. there is no potential reusable replica
 //   5. there is potential reusable replica but the replica replenishment wait interval is passed.
-func (rcs *ReplicaScheduler) RequireNewReplica(replicas map[string]*longhorn.Replica, volume *longhorn.Volume, hardNodeAffinity string) bool {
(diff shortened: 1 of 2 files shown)


scheduler/replica_scheduler.go
```diff
@@ -407,18 +407,20 @@ func (rcs *ReplicaScheduler) CheckAndReuseFailedReplica(replicas map[string]*lon
...
+// If creating new replica immediately is necessary, returns true.
+// Otherwise, returns false and the duration that the caller should recheck.
 // A new replica needs to be created when:
 //   1. the volume is a new volume (volume.Status.Robustness is Empty)
 //   2. data locality is required (hardNodeAffinity is not Empty and volume.Status.Robustness is Healthy)
 //   3. replica eviction happens (volume.Status.Robustness is Healthy)
 //   4. there is no potential reusable replica
 //   5. there is potential reusable replica but the replica replenishment wait interval is passed.
-func (rcs *ReplicaScheduler) RequireNewReplica(replicas map[string]*longhorn.Replica, volume *longhorn.Volume, hardNodeAffinity string) bool {
+func (rcs *ReplicaScheduler) RequireNewReplica(replicas map[string]*longhorn.Replica, volume *longhorn.Volume, hardNodeAffinity string) (bool, time.Duration) {
```
shuo-wu, line 418: NIT: Actually, the output of this function can be changed to a single field `replenishmentWaitInterval time.Duration`

shuo-wu (commented):
One quick question related to a corner case:
If Longhorn needs to rebuild multiple replicas for a large volume, will the 2nd or subsequent rebuilding get started in time after the 1st rebuilding is complete?

PhanLe1010:
> If Longhorn needs to rebuild multiple replicas for a large volume, will the 2nd or subsequent rebuilding get started in time after the 1st rebuilding is complete?

I think yes because after replica rebuilding, the engine.Status.rebuildStatus is updated. This update leads to enqueuing the corresponding volume CR. Then replenishCount logic is run again

scheduler/replica_scheduler.go
PhanLe1010: Fixed. Please check if the fix is what you are suggesting

shuo-wu (approved):

innobead (approved):

PR #1261, opened 2022-03-04
[CSI Sanity] Should fail when the volume source snapshot is not found

c3y1huang:
Fix CSI sanity test failure.

```
    unexpected error: invalid csi snapshot type: . Must be snap or bak
    Expected
        <codes.Code>: 3
    to equal
        <codes.Code>: 5
```


shuo-wu (commented):
It seems that the CSI should return error `NotFound` once there is an empty result from `decodeSnapshotID(snapshot.SnapshotId)`

c3y1huang:
> It seems that the CSI should return error `NotFound` once there is an empty result from `decodeSnapshotID(snapshot.SnapshotId)`

Updated.

innobead (approved):

PR #1262, opened 2022-03-05
webhook: set securityContext

derekbit:
To follow the least privilege principle, the conversion and admission webhooks
are run with a non-root user (uid = 2000).

Longhorn #3701


jenting (dismissed):

derekbit:
Ping @innobead, thank you.

innobead (approved):

PR #1265, opened 2022-03-08
[CSI Sanity] Add host mount /tmp dir to longhorn-csi-plugin

c3y1huang:
CSI sanity uses [defaultCheckPath](https://github.com/kubernetes-csi/csi-test/blob/db65255108c0649001a55df9f0a09dfcec62b183/pkg/sanity/sanity.go#L543-L544) to check target path existence. Running the test from the host currently does not know about the pod /tmp directory(default target root) where csi-mount and csi-staging gets created, hence the `should remove target path`
test fails.

Add host mount /tmp directory in longhorn-csi-plugin for host to have access to
/tmp/csi-mount/ and /tmp/csi-staging/.

### DIFF
--- a/csi/deployment.go
+++ b/csi/deployment.go
@@ -457,6 +457,10 @@ func NewPluginDeployment(namespace, serviceAccount, nodeDriverRegistrarImage, ma
 									MountPath: "/sys",
 								},
 								{
+									Name:      "host-tmp", // path is required to pass csi-sanity
+									MountPath: "/tmp",
+								},
+								{
 									Name:             "host",
 									MountPath:        "/rootfs", // path is required for namespaced mounter
 									MountPropagation: &MountPropagationBidirectional,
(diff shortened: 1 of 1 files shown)


c3y1huang:
Alternative, we can also use what is already mounted [here](https://github.com/longhorn/longhorn-manager/blob/master/csi/deployment.go#L451-L458). (/dev, /sys).
And provided as a cmd-line option to specify the custom target path.
```
mkdir /dev/target;
mkdir /dev/staging;
./csi-sanity -csi.endpoint /var/lib/kubelet/plugins/driver.longhorn.io/csi.sock 
  --csi.mountdir="/dev/target/mount" 
  --csi.stagingdir="/dev/staging/mount"
  --ginkgo.focus="should remove target path"
```

c3y1huang:
Discussed with @innobead . Close this PR and use the [workaround](https://github.com/longhorn/longhorn-manager/pull/1265#issuecomment-1062454623) instead.

PR #1266, opened 2022-03-09
Enqueue the source volume of a cloning when enqueue the target volume of the cloning

PhanLe1010:
longhorn/longhorn#3692


innobead:
@PhanLe1010 do we still need this one?

PhanLe1010:
Can we keep this one. We might want to circle back here when removing the resync period

PhanLe1010:
We don't need this one any more, the new AD refactor takes care of the original issue

PR #1270, opened 2022-03-10
Mark backup to Pending state if encounter the engine RPC error code DeadlineExceeded

jenting:
Mark backup to **Pending** state if the engine SnapshotBackup API call returns RPC error code DeadlineExceeded.
The sync agent replica server _probably_ perform snapshot backup initialization succeeded in the end.

For this case, we create a backup monitor routine to periodically (2 seconds) fetch the backup status to see if the snapshot backup change initialization in the end (state change to InProgress/Completed) or keeps in an Error state. We have a maximum retry period of 10 minutes for the Pending state.

https://github.com/longhorn/longhorn/issues/3545

### DIFF
--- a/controller/backup_controller.go
+++ b/controller/backup_controller.go
@@ -505,6 +505,7 @@ func (bc *BackupController) checkMonitor(backup *longhorn.Backup, volume *longho
 	if err != nil {
 		backup.Status.Error = err.Error()
 		backup.Status.State = longhorn.BackupStateError
+		backup.Status.LastSyncedAt = metav1.Time{Time: time.Now().UTC()}
 		return nil, err
 	}
 	return monitor, nil
--- a/controller/backup_controller.go
+++ b/controller/backup_controller.go
@@ -338,7 +338,7 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
 		}
 
 		switch backup.Status.State {
-		case longhorn.BackupStateNew, longhorn.BackupStateInProgress:
+		case longhorn.BackupStateNew, longhorn.BackupStatePending, longhorn.BackupStateInProgress:
 			return nil
 		case longhorn.BackupStateCompleted:
 			bc.disableBackupMonitor(backup.Name)
--- a/engineapi/backup_monitor.go
+++ b/engineapi/backup_monitor.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"strings"
 	"sync"
 	"time"
 
@@ -48,7 +49,6 @@ func NewBackupMonitor(logger logrus.FieldLogger,
 		snapshotName: backup.Spec.SnapshotName,
 		engineClient: engineClient,
 
-		backupStatus:     backup.Status,
 		backupStatusLock: sync.RWMutex{},
 
 		syncCallback: syncCallback,
@@ -63,10 +63,29 @@ func NewBackupMonitor(logger logrus.FieldLogger,
 			backupTargetClient.URL, volume.Spec.BackingImage, biChecksum,
 			backup.Spec.Labels, backupTargetClient.Credential)
 		if err != nil {
-			m.logger.WithError(err).Warn("Cannot take snapshot backup")
-			return nil, err
+			if !strings.Contains(err.Error(), "DeadlineExceeded") {
+				m.logger.WithError(err).Warn("Cannot take snapshot backup")
+				return nil, err
+			}
+
+			// [Workaround]
+			// Special handling the RPC call return code DeadlineExceeded, mark it as Pending state.
+			// The snapshot backup initialization _probably_ succeeded in the replica sync agent server.
+			// Use the backup monitor routine to monitor the backup status stays in Error state or change
+			// to InProgress/Completed state with the maximum retry count mechanism.
+			// [TODO]
+			// Since API engineclient.SnapshotBackup is not idempotent, this controller cannot blindly
+			// retry the call when error DeadlineExceeded is triggered.
+			// Instead, it has to mark the backup as a kind of special state Pending, then relies on the
+			// backup monitor routine periodically checking if the backup creation actually started.
+			// After making the API call idempotent and being able to deprecate the old version
+			// engine image, we can remove this part.
+			// https://github.com/longhorn/longhorn/issues/3545
+			m.logger.WithError(err).Warnf("Snapshot backup timeout")
+			backup.Status.State = longhorn.BackupStatePending
 		}
 
+		m.backupStatus = backup.Status
 		m.replicaAddress = replicaAddress
 	}
 
--- a/engineapi/backup_monitor.go
+++ b/engineapi/backup_monitor.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/sirupsen/logrus"
 
+	"k8s.io/apimachinery/pkg/util/clock"
 	"k8s.io/apimachinery/pkg/util/wait"
 
 	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
@@ -17,12 +18,15 @@ import (
 
 const (
 	BackupMonitorSyncPeriod = 2 * time.Second
+
+	// BackupMonitorMaxRetryPeriod is the maximum retry period when backup monitor routine
+	// encounters an error and backup stays in Pending state
+	BackupMonitorMaxRetryPeriod = 24 * time.Hour
 )
 
 type BackupMonitor struct {
 	logger logrus.FieldLogger
 
-	namespace      string
 	backupName     string
 	snapshotName   string
 	replicaAddress string
@@ -32,9 +36,12 @@ type BackupMonitor struct {
 	backupStatusLock sync.RWMutex
 
 	syncCallback func(key string)
+	callbackKey  string
 
 	ctx  context.Context
 	quit context.CancelFunc
+
+	retryCount int
 }
 
 func NewBackupMonitor(logger logrus.FieldLogger,
@@ -42,9 +49,8 @@ func NewBackupMonitor(logger logrus.FieldLogger,
 	biChecksum string, engineClient EngineClient, syncCallback func(key string)) (*BackupMonitor, error) {
 	ctx, quit := context.WithCancel(context.Background())
 	m := &BackupMonitor{
-		logger: logger,
+		logger: logger.WithFields(logrus.Fields{"backup": backup.Name}),
 
-		namespace:    backup.Namespace,
 		backupName:   backup.Name,
 		snapshotName: backup.Spec.SnapshotName,
 		engineClient: engineClient,
@@ -52,6 +58,7 @@ func NewBackupMonitor(logger logrus.FieldLogger,
 		backupStatusLock: sync.RWMutex{},
 
 		syncCallback: syncCallback,
+		callbackKey:  backup.Namespace + "/" + backup.Name,
 
 		ctx:  ctx,
 		quit: quit,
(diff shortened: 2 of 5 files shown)


jenting:
E2E test locally with NFS backup target only.

```
# py.test -k backup -v
========================================================= test session starts ==========================================================
platform linux -- Python 3.9.10, pytest-5.3.1, py-1.11.0, pluggy-0.13.1 -- /usr/local/bin/python
cachedir: .pytest_cache
plugins: repeat-0.9.1
collected 230 items / 200 deselected / 30 selected                                                                                     

test_backing_image.py::test_backup_with_backing_image[nfs] PASSED                                                                [  3%]
test_backing_image.py::test_backup_labels_with_backing_image[nfs] PASSED                                                         [  6%]
test_backing_image.py::test_ha_backup_deletion_recovery[nfs] PASSED                                                              [ 10%]
test_backing_image.py::test_csi_backup_with_backing_image[nfs] PASSED                                                            [ 13%]
test_basic.py::test_backup_status_for_unavailable_replicas[nfs] PASSED                                                           [ 16%]
test_basic.py::test_backup_block_deletion[nfs] PASSED                                                                            [ 20%]
test_basic.py::test_dr_volume_with_backup_block_deletion[nfs] PASSED                                                             [ 23%]
test_basic.py::test_dr_volume_with_backup_block_deletion_abort_during_backup_in_progress[nfs] PASSED                             [ 26%]
test_basic.py::test_dr_volume_with_all_backup_blocks_deleted[nfs] PASSED                                                         [ 30%]
test_basic.py::test_backup_volume_list[nfs] PASSED                                                                               [ 33%]
test_basic.py::test_backup_metadata_deletion[nfs] PASSED                                                                         [ 36%]
test_basic.py::test_backup[nfs] PASSED                                                                                           [ 40%]
test_basic.py::test_backup_labels[nfs] PASSED                                                                                    [ 43%]
test_basic.py::test_deleting_backup_volume[nfs] PASSED                                                                           [ 46%]
test_basic.py::test_listing_backup_volume PASSED                                                                                 [ 50%]
test_basic.py::test_storage_class_from_backup[nfs] PASSED                                                                        [ 53%]
test_basic.py::test_backup_lock_deletion_during_restoration[nfs] PASSED                                                          [ 56%]
test_basic.py::test_backup_lock_deletion_during_backup[nfs] PASSED                                                               [ 60%]
test_basic.py::test_backup_lock_creation_during_deletion[nfs] PASSED                                                             [ 63%]
test_basic.py::test_backup_lock_restoration_during_deletion[nfs] SKIPPED                                                         [ 66%]
test_basic.py::test_backuptarget_available_during_engine_image_not_ready PASSED                                                  [ 70%]
test_csi.py::test_csi_backup[nfs] PASSED                                                                                         [ 73%]
test_csi_snapshotter.py::test_csi_volumesnapshot_restore_existing_backup[nfs-Delete-True] PASSED                                 [ 76%]
test_csi_snapshotter.py::test_csi_volumesnapshot_restore_existing_backup[nfs-Retain-False] PASSED                                [ 80%]
test_ha.py::test_ha_backup_deletion_recovery[nfs] PASSED                                                                         [ 83%]
test_ha.py::test_restore_volume_with_invalid_backupstore SKIPPED                                                                 [ 86%]
test_kubernetes.py::test_backup_kubernetes_status[nfs] PASSED                                                                    [ 90%]
test_recurring_job.py::test_recurring_job_backup[nfs] PASSED                                                                     [ 93%]
test_statefulset.py::test_statefulset_backup[nfs] PASSED                                                                         [ 96%]
test_statefulset.py::test_statefulset_recurring_backup[nfs] PASSED                                                               [100%]

====================================== 28 passed, 2 skipped, 200 deselected in 5194.66s (1:26:34) ======================================
```

engineapi/backup_monitor.go
```diff
@@ -17,6 +17,14 @@ import (
...
 const (
 	BackupMonitorSyncPeriod = 2 * time.Second
+
+	// BackupMonitorMaxRetryPeriod is the maximum retry period when backup monitor routine
+	// encounters an error and backup stays in Pending state
+	BackupMonitorMaxRetryPeriod = 10 * time.Minute
+
+	// BackupMonitorMaxRetryCount is the maximum retry count before the backup monitor routine
+	// stops monitor and backup stays in Pending state
+	BackupMonitorMaxRetryCount = int(BackupMonitorMaxRetryPeriod / BackupMonitorSyncPeriod)
```
shuo-wu, line 27: I am thinking if we can use a longer retry interval rather than the backup monitor default resync period. Besides, the max retry limit is too short for large backups. I am thinking 1 or 2 hours would be better.

innobead (commented):
In general, LGTM, but leave other reviews to do the final check like longer retry period consideration. 

@jenting Just one question: if the longhorn-manager at that node gets restarted, so the unsync ready backup resource will be moved the ownership and it will cause another snapshot creation to get triggered in one volume replica. This is the current behavior, correct? 

Then, `idempotent` is needed in the future, do we have a ticket to track this? Also, I feel we have many monitoring stuff coupled with controllers, probably we should rethink if we need to consolidate the monitoring mechanism, however, this is another topic here.

engineapi/backup_monitor.go
PhanLe1010: Do you mean 1 to 2 hours could be the time that it takes for a large backup to be `initiated`?

jenting:
> In general, LGTM, but leave other reviews to do the final check like longer retry period consideration.
> 
> @jenting Just one question: if the longhorn-manager at that node gets restarted, so the unsync ready backup resource will be moved the ownership and it will cause another snapshot creation to get triggered in one volume replica. This is the current behavior, correct?

No. The ownership would move to another node but it won't cause another snapshot backup creation because of this [condition check](https://github.com/longhorn/longhorn-manager/blob/c04eac1c804597c78d5962c1758e8ed80d2abb37/engineapi/backup_monitor.go#L61)

> Then, `idempotent` is needed in the future, do we have a ticket to track this? Also, I feel we have many monitoring stuff coupled with controllers, probably we should rethink if we need to consolidate the monitoring mechanism, however, this is another topic here.

I haven't created a ticket for this. But for the monitoring mechanism, we have an issue tracking https://github.com/longhorn/longhorn/issues/2441.

engineapi/backup_monitor.go
jenting: I implemented the code with exponential backoff retry mechanism by ref the [Kubernetes code](https://github.com/kubernetes/kubernetes/blob/a7d866f036a45cd4c4c8c1b183fa33c65ee1d28f/pkg/volume/csi/csi_attacher.go#L479-L511) when the Backup CR fails to `Pending` state.

The back off retry period would be 4s/8s/16s/32s/..., the maximum retry period is 2 hours.

shuo-wu: Let's say the max volume/backup size Longhorn supports now is 1TB. Then you can try [this cmd](https://github.com/longhorn/longhorn/issues/2507#issuecomment-857195496) with a 1TB sparse file and see how long it will take to get the fiemap. This time should be the max retry limit.

engineapi/backup_monitor.go
```diff
@@ -63,10 +70,28 @@ func NewBackupMonitor(logger logrus.FieldLogger,
...
+				m.logger.WithError(err).Warn("Cannot take snapshot backup")
+				return nil, err
+			}
+
+			// [Workaround]
+			// Special handling the RPC call return code DeadlineExceeded, mark it as Pending state.
+			// The snapshot backup initialization _probably_ succeeded in the replica sync agent server.
+			// Use the backup monitor routine to monitor the backup status stays in Error state or change
+			// to InProgress/Completed state with the maximum retry count mechanism.
+			// [TODO]
```
joshimoo, line 83: Good commenting, you further link to this PR/issue after the todo so that future readers can get additional context.

engineapi/backup_monitor.go
```diff
@@ -77,26 +102,91 @@ func NewBackupMonitor(logger logrus.FieldLogger,
...
+		resetDuration = BackupMonitorMaxRetryPeriod
+		backoffFactor = 2.0
+		jitter        = 0.0
+		clock         = clock.RealClock{}
+		retryCount    = 0
+	)
+	backoffMgr := wait.NewExponentialBackoffManager(initBackoff, maxBackoff, resetDuration, backoffFactor, jitter, clock)
+	defer backoffMgr.Backoff().Stop()
+
+	ctx, cancel := context.WithTimeout(context.Background(), BackupMonitorMaxRetryPeriod)
```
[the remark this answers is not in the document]
joshimoo: Should be able to use the monitors context as a parent, so that when the monitors context gets cancelled the child ctx can also be cancelled. Unless we need to differentiate between the 2 cases? 

Looks like we differentiate below as part of the case. So we cannot rely on the parent ctx cancellation propagation.

k8s/pkg/apis/longhorn/v1beta2/backup.go
```diff
@@ -6,6 +6,7 @@ type BackupState string
 
 const (
 	BackupStateNew        = BackupState("")
+	BackupStatePending    = BackupState("Pending")
```
joshimoo, line 9: The Pending state addition is good,

k8s/pkg/apis/longhorn/v1beta1/backup.go
```diff
@@ -6,6 +6,7 @@ type BackupState string
 
 const (
 	BackupStateNew        = BackupState("")
+	BackupStatePending    = BackupState("Pending")
```
joshimoo, line 9: NIT: technically this wasn't a valid state for v1beta1 but since it's not an ordered enum, it seems fine to keep it here :)

controller/backup_controller.go
```diff
@@ -328,6 +328,12 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
 			bc.syncBackupStatusWithSnapshotCreationTimeAndVolumeSize(volume, backup)
 		}
 
+		defer func() {
+			if backup.Status.State == longhorn.BackupStateError || backup.Status.State == longhorn.BackupStateUnknown {
+				backup.Status.LastSyncedAt = syncTime
+			}
+		}()
```
joshimoo, line 335: This is seems like a change in functionality, you are now always resetting the syncTime, when the `checkMonitor/syncWithMonitor` throws an error.

Are there any repercussions for this?

engineapi/backup_monitor.go
```diff
@@ -77,26 +102,91 @@ func NewBackupMonitor(logger logrus.FieldLogger,
...
+	if !fallbackToExponentialBackoff {
+		return
+	}
+
+	// Exponential back off timer
+	m.logger.Info("Fallback to exponential back off timer")
+	var (
+		initBackoff   = BackupMonitorSyncPeriod
+		maxBackoff    = 10 * time.Minute
+		resetDuration = BackupMonitorMaxRetryPeriod
```
joshimoo, line 131: NIT: have you considered whether we can use an instance of a ratelimiter for all monitors?
Possibly either using one of these implementations or implementing this interface our selfs, I think having a ratelimiter that we can reuse / replace during unit tests would be a nice addition?

You can pass an instance of the rate limiter monitor as part of the construction of the monitor, so all monitors can use the same one, this way we can also add additional flow control (i.e. when 100 backups all need to be synced at the same time, etc)

REF: 
https://github.com/kubernetes/client-go/blob/master/util/workqueue/default_rate_limiters.go#L27

joshimoo (commented):
NIT: have you considered whether we can use an instance of a ratelimiter for all monitors?
Possibly either using one of these implementations or implementing this interface our selfs, I think having a ratelimiter that we can reuse / replace during unit tests would be a nice addition?

You can pass an instance of the rate limiter monitor as part of the construction of the monitor, so all monitors can use the same one, this way we can also add additional flow control (i.e. when 100 backups all need to be synced at the same time, etc)

REF:
https://github.com/kubernetes/client-go/blob/master/util/workqueue/default_rate_limiters.go#L27

engineapi/backup_monitor.go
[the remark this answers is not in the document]
shuo-wu: Oh OK. We do differentiate the handling for these 2 ctx.

controller/backup_controller.go
jenting: Revert it. We should set the `status.lastSyncedAt` once the backup monitor enablement failed.

engineapi/backup_monitor.go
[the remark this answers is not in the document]
jenting: The `resetDuration` is for if the backoff is not called within it, the backoff is reset.
I was thinking we should not reset the backoff until we reach the BackupMonitorMaxRetryPeriod. (Which means it would never be reset).

engineapi/backup_monitor.go
```diff
@@ -77,26 +102,91 @@ func NewBackupMonitor(logger logrus.FieldLogger,
...
+		select {
+		case <-backoffMgr.Backoff().C():
+			retryCount++
+			m.logger.Debugf("Within exponential back off timer retry count %d", retryCount)
+			fallback, err := m.syncBackups()
+			if err != nil {
+				m.logger.Errorf("Stop monitoring because of %v in exponential back off timer retry count %d", err, retryCount)
+				return
+			}
+			if !fallback {
```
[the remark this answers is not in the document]
jenting: Reword.

k8s/pkg/apis/longhorn/v1beta1/backup.go
jenting: Since it'd be backported to v1.2.4, thus I intended to add the Pending state for v1beta1.

engineapi/backup_monitor.go
jenting: Sounds good.

The only thing we need to do is manually calculated the number of maximum retries count since we use the `BackupMonitorMaxRetryPeriod` to determine how long should the monitor give up retry.
With the rate limiter interface, we use the maximum retries count to determine how many times should the monitor give up retry, like what we did for [backup lock deletion max retry count](https://github.com/longhorn/longhorn-manager/blob/452ca6e313b7378c9bd4c3d35e1a66a6eadfbcde/controller/backup_controller.go#L43).

engineapi/backup_monitor.go
```diff
@@ -96,26 +103,91 @@ func NewBackupMonitor(logger logrus.FieldLogger,
...
+
+	ctx, cancel := context.WithTimeout(context.Background(), BackupMonitorMaxRetryPeriod)
+	defer cancel()
+
+	for {
+		select {
+		case <-backoffMgr.Backoff().C():
+			retryCount++
+			m.logger.Debugf("Within exponential back off timer retry count %d", retryCount)
+			exponentialBackoffRequired, err := m.syncBackups()
```
shuo-wu, line 149: This may error out, right? Then should we stop the monitor below?

jenting: Currently, we rely on the Backup CR `status.state` in the backup controller to stop the monitor if it's error out _or_ is completed.

k8s/pkg/apis/longhorn/v1beta1/backup.go
innobead: 1.2.5 instead

innobead (dismissed):
In general, LGTM

engineapi/backup_monitor.go
shuo-wu: Here this `monitorBackups` goroutine will exit without any auto-recovery mechanism after erroring out. 
In the controller's perspective, the monitor is still in the map but actually there is nothing periodically talking with the engine.

jenting: Agree. It'd cause the memory leak in the [monitors](https://github.com/longhorn/longhorn-manager/blob/e89fca5206c67423dafe2c5606f30d737b1af52e/controller/backup_controller.go#L57) map.

shuo-wu (dismissed):
LGTM

shuo-wu (commented):
Will the state value modification affect the existing metrics?

jenting:
> Will the state value modification affect the existing metrics?

No, because the backup metric be release at Longhorn v1.3.0.

shuo-wu (approved):

innobead (approved):

innobead:
@jenting Please help backport to 1.2.5. Thanks.

PR #1272, opened 2022-03-11
Revert "Fix CVE-2022-21698"

jenting:
The go mod lib needs to be update as well.
However, it'd encounter protocol buffer VersionResponse conflict when
bumping the protocol buffer packages which causes CI failure.

Therefore, we decide the revert the commmit a7e9fb2245b81b2166dda9bfec75cc5482dedf36.


innobead (approved):

innobead:
cc @c3y1huang

PR #1276, opened 2022-03-17
admission webhook: add README.md

derekbit:
Longhorn 3752

### DIFF
--- a/webhook/admission/README.md
+++ b/webhook/admission/README.md
@@ -1,0 +1,8 @@
+# Admission Webhook
+
+The admission webhook only validates or mutates the spec of hooked resources
+before being persistent to the database like etcd. Any resource reconciliation
+to ensure resources behave as expected should be handled in the backend as usual.
+For example, controllers or even host operations like creating and deleting
+folders at the host as usual.
+


webhook/admission/README.md
```diff
@@ -0,0 +1,8 @@
+# Admission Webhook
+
+The admission webhook validates or changes the values in a request.
+An resource request from longhorn-manager might be handled by an
+admission webhook on another node. Therefore, any manipulation of
+the resources of a node, such as creating and deleting folders,
+should be avoided. The manipulations should be handled by the 
+longhorn-manager daemon or controller.
```
innobead, line 8: How about

`
The admission webhook only validates or mutates the spec of hooked resources before resource reconsilation. Beyound that, any resource reconsilation to ensure resources behave as expected should still be handled in the backend (ex: controllers or even host operations like creating/deleting folders at host) as usual. 
`

innobead (changes requested):

webhook/admission/README.md
derekbit: Looks good, thank you.

innobead: LGTM

webhook/admission/README.md
```diff
@@ -0,0 +1,8 @@
+# Admission Webhook
+
+The admission webhook only validates or mutates the spec of hooked resources
+before being persistent to the database like etcd. Any resource reconciliation
+to ensure resources behave as expected should be handled in the backend as usual.
+For exampleex, controllers or even host operations like creating and deleting
```
innobead, line 6:
```suggestion
For example, controllers or even host operations like creating and deleting
```

innobead (approved):

PR #1279, opened 2022-03-24
Fix upgrade and admission webhook

derekbit:
- The default `matchingPolicy`, `equivalent`, in the admission webohook breaks the upgrade path. Replace `equivalent` with `exact`.
- Fix the invalid value in the upgrade path and mutating webhook as well.
- Fix uninstallation failure due to the delete validation.

[Longhorn#3777](https://github.com/longhorn/longhorn/issues/3777)


jenting (approved):

manager/backingimage.go
```diff
@@ -88,6 +88,13 @@ func (m *VolumeManager) CreateBackingImage(name, checksum, sourceType string, pa
 }
 
 func (m *VolumeManager) DeleteBackingImage(name string) error {
+	replicas, err := m.ds.ListReplicasByBackingImage(name)
```
innobead, line 91: Why we move the validation back here from the admission webhook?

derekbit: The checks in DELETE are not general. Uninstallation does not need the checks.
https://github.com/longhorn/longhorn-manager/blob/master/controller/uninstall_controller.go#L361

innobead: As discussed, `backing image validation` looks general, but the validation of node & engine images moved out from the admission look good. 

Also, suggest adding a comment to the README to mention when adding validation to deletion, we should be aware of this possible side effect.

innobead (approved):

PR #1283, opened 2022-03-31
Increase the maxRetries

derekbit:
Increate the maxRetries to 12, so the maximum cumulative retry time is 20.475 seconds.

https://github.com/longhorn/longhorn/issues/3620


derekbit:
Keep it as is.

PR #1284, opened 2022-04-01
Move minimum and maximum CRD schema validation to validating webhook

derekbit:
Fix the helm upgrade failure on kuberenetes v1.18.
Root cause: https://github.com/kubernetes/kubernetes/issues/87675
    
[Longhorn 3631](https://github.com/longhorn/longhorn/issues/3631)

### DIFF
--- a/k8s/pkg/apis/longhorn/v1beta2/recurringjob.go
+++ b/k8s/pkg/apis/longhorn/v1beta2/recurringjob.go
@@ -33,12 +33,9 @@ type RecurringJobSpec struct {
 	// +optional
 	Cron string `json:"cron"`
 	// The retain count of the snapshot/backup.
-	// +kubebuilder:validation:Minimum=1
-	// +kubebuilder:validation:Maximum=50
 	// +optional
 	Retain int `json:"retain"`
 	// The concurrency of taking the snapshot/backup.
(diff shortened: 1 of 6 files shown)


k8s/pkg/apis/longhorn/v1beta2/recurringjob.go
```diff
@@ -33,12 +33,9 @@ type RecurringJobSpec struct {
 	// +optional
 	Cron string `json:"cron"`
 	// The retain count of the snapshot/backup.
-	// +kubebuilder:validation:Minimum=1
-	// +kubebuilder:validation:Maximum=50
```
innobead, line 37: Why changes in recurring job and volume did not move to their validators?

derekbit: For node, the validation has already existed in https://github.com/longhorn/longhorn-manager/blob/master/webhook/resources/volume/validator.go#L58.

The recurringjob's upper bound is in https://github.com/longhorn/longhorn-manager/blob/master/webhook/resources/recurringjob/validator.go#L57, but the lower bound is missing.
So, I update the PR and add it explicitly.

innobead (approved):

PR #1286, opened 2022-04-07
controller: Monitor and handle the default StorageClass only

shuo-wu:
Longhorn/longhorn#3821


PhanLe1010 (approved):
LGTM.
it is a good idea checking the name first before checking the driver since the name is unique 👍

innobead (approved):

PR #1287, opened 2022-04-07
Implement orphaned data cleanup

derekbit:
- Introduce orphan CRD and its controller
- Introduce generic monitor interface
- Introduce node monitor
- Implement the orphan CRs cleanup and creation logics
- Introduce global orphan-auto-deletion setting

In the implementation, we focus on the cleanup of orphaned directories. The generic framework can be extended to different types of orphaned data in the future.
 
---

Orphaned replica directories processing flow:
- The node monitor collects the disk status and on-disk replica directories every 30 seconds and then enqueues the node resource objects.
- Then, node controller syncs the node, cleans up the orphan and its on-disk data and create orphans.
---

LEP: https://github.com/longhorn/longhorn/blob/master/enhancements/20220324-orphaned-data-cleanup.md

https://github.com/longhorn/longhorn/issues/685


k8s/pkg/apis/longhorn/v1beta2/orphan.go
```diff
@@ -0,0 +1,70 @@
...
+	// +optional
+	NodeID string `json:"nodeID"`
+	// The type of the orphaned data.
+	// Can be "replica".
+	// +optional
+	Type OrphanType `json:"orphanType"`
+
+	// The parameters of the orphaned data
+	// +optional
+	// +nullable
```
shuo-wu, line 34: Can this be nullable?

types/types.go
```diff
@@ -397,6 +399,15 @@ func GetRecurringJobLabelValueMap(labelType, recurringJobName string) map[string
 		GetRecurringJobLabelKey(labelType, recurringJobName): LonghornLabelValueEnabled,
 	}
 }
+
+func GetOrphanLabels(node, diskUUID, orphanName string) map[string]string {
```
shuo-wu, line 403: 1. Have you used `orphanName`?
2. This CRD is for general cases. But the labels here are for replica directories only. In other words, you need to label the type first.
3. Add the related label when the input is not empty.

types/types.go
```diff
@@ -421,6 +432,10 @@ func GetShareManagerImageChecksumName(image string) string {
 	return shareManagerImagePrefix + util.GetStringChecksum(strings.TrimSpace(image))[:ImageChecksumNameLength]
 }
 
+func GetOrphanChecksumName(name, diskID, nodeID string) string {
```
shuo-wu, line 435: Rename this to something like `GetChecksumNameOrphanReplicaDirectory`

webhook/resources/orphan/mutator.go
```diff
@@ -0,0 +1,78 @@
...
+
+	orphan := newObj.(*longhorn.Orphan)
+
+	// Merge the user created and longhorn specific labels
+	labels := orphan.Labels
+	if labels == nil {
+		labels = map[string]string{}
+	}
+
+	longhornLabels := types.GetOrphanLabels(orphan.Spec.NodeID, orphan.Spec.Parameters[longhorn.OrphanDiskUUID], orphan.Name)
```
shuo-wu, line 56: Check or add labels based on the type

webhook/resources/orphan/validator.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/longhorn/longhorn-manager/webhook/admission"
+	werror "github.com/longhorn/longhorn-manager/webhook/error"
+)
+
+type orphanValidator struct {
+	admission.DefaultValidator
+	ds *datastore.DataStore
+}
+
+func NewValidator(ds *datastore.DataStore) admission.Validator {
```
shuo-wu, line 21: Need to check the type and the required parameters in the validator

controller/orphan_controller.go
```diff
@@ -0,0 +1,314 @@
...
+		MatchLabels: map[string]string{
+			types.GetLonghornLabelKey(types.LonghornLabelNode): node.Name,
+		},
+	})
+	if err != nil {
+		utilruntime.HandleError(fmt.Errorf("failed to create label selector label since %v", err))
+		return
+	}
+
+	orphans, err := oc.ds.ListOrphansBySelectorRO(labelSelector)
```
shuo-wu, line 125: Can we add a function named `ListOrphansByNodeRO` instead?

controller/orphan_controller.go
```diff
@@ -0,0 +1,314 @@
...
+		if disk, ok := node.Spec.Disks[id]; ok {
+			if disk.EvictionRequested {
+				return nil
+			}
+		} else {
+			log.Debugf("Only delete the orphan resource object since disk %v on node %v does not exist", id, node.Name)
+			return nil
+		}
+
+		if node.Status.DiskStatus[id].DiskUUID == orphan.Spec.Parameters[longhorn.OrphanDiskUUID] {
```
shuo-wu, line 268: Do we have a function that checks the node & disk ready status?

controller/orphan_controller.go
```diff
@@ -0,0 +1,314 @@
...
+				return err
+			}
+		}
+		return nil
+	}
+
+	existingOrphan := orphan.DeepCopy()
+
+	orphan.Status.Conditions = types.SetCondition(orphan.Status.Conditions,
+		longhorn.OrphanConditionTypeDeletable, longhorn.ConditionStatusTrue, "", "")
```
shuo-wu, line 283: What if `orphan.Status.OwnerID != orphan.Spec.NodeID` or `oc.controllerID != = orphan.Spec.NodeID`?

Or in other words, does `deletable` here mean that the data can be cleaned up?

types/setting.go
```diff
@@ -73,6 +73,7 @@ const (
 	SettingNameBackingImageRecoveryWaitInterval             = SettingName("backing-image-recovery-wait-interval")
 	SettingNameGuaranteedEngineManagerCPU                   = SettingName("guaranteed-engine-manager-cpu")
 	SettingNameGuaranteedReplicaManagerCPU                  = SettingName("guaranteed-replica-manager-cpu")
+	SettingNameOrphanAutoDeletion                           = SettingName("orphan-auto-deletion")
```
shuo-wu, line 76: Need to add this setting CR in `deploy/install/01-prerequisite/04-default-setting.yaml`.

shuo-wu (changes requested):
Will continue reviewing the node part tomorrow

controller/node_controller.go
```diff
@@ -1021,3 +871,215 @@ func BackingImageDiskFileCleanup(node *longhorn.Node, bi *longhorn.BackingImage,
...
+		},
+	})
+	if err != nil {
+		return err
+	}
+	orphans, err := nc.ds.ListOrphansBySelectorRO(labelSelector)
+	if err != nil {
+		return errors.Wrapf(err, "unable to list orphans with label %v=%v",
+			types.GetLonghornLabelKey(types.LonghornLabelNode), nc.controllerID)
+	}
```
shuo-wu, line 931: Add a datastore function `ListOrphansByNodeIDRO`

controller/node_controller.go
```diff
@@ -1021,3 +871,215 @@ func BackingImageDiskFileCleanup(node *longhorn.Node, bi *longhorn.BackingImage,
...
+
+	for _, orphan := range orphans {
+		if orphan.Status.OwnerID != nc.controllerID {
+			continue
+		}
+
+		if orphan.Spec.NodeID == orphan.Status.OwnerID {
```
shuo-wu, line 938: Do we need to check these node ID fields here? The orphan controller should handle it and tell the orphan object final state.

controller/node_controller.go
```diff
@@ -1021,3 +871,215 @@ func BackingImageDiskFileCleanup(node *longhorn.Node, bi *longhorn.BackingImage,
...
+				id := orphan.Spec.Parameters[longhorn.OrphanDiskFsid]
+				replicaDirectoryName := orphan.Spec.Parameters[longhorn.OrphanDataName]
+				_, onDiskDataExist := onDiskReplicaDirectoryNames[id][replicaDirectoryName]
+
+				if !onDiskDataExist || isDiskDownOrEvicted(node, orphan) || isDiskUUIDChanged(node, orphan) {
+					if err := nc.ds.DeleteOrphan(orphan.Name); err != nil {
+						return errors.Wrapf(err, "failed to delete orphan %v", orphan.Name)
+					}
+				}
+			}
```
shuo-wu, line 957: Duplicate code. There is similar logic in the orphan controller. As I mentioned above, the orphan controller should tell the final state of orphan objects then the node controller just needs to send deletion requests. You can use `conditions` or a new field to represent the orphan CR state.

controller/node_controller.go
```diff
@@ -1021,3 +871,215 @@ func BackingImageDiskFileCleanup(node *longhorn.Node, bi *longhorn.BackingImage,
...
+func (nc *NodeController) createOrphan(node *longhorn.Node, replicaDirectoryName, id string) error {
+	orphanName := types.GetOrphanChecksumName(replicaDirectoryName, id, node.Name)
+
+	_, err := nc.ds.GetOrphanRO(orphanName)
+	if err == nil || (err != nil && !apierrors.IsNotFound(err)) {
+		return err
+	}
+
+	orphan := &longhorn.Orphan{
+		ObjectMeta: metav1.ObjectMeta{
```
shuo-wu, line 1026: I am thinking if adding `OwnershipReference` with cascading deletion would help simplify the logic.

controller/monitor/fake_node_monitor.go
```diff
@@ -0,0 +1,341 @@
...
+	m.state.OnDiskReplicaDirectoryNames = onDiskReplicaDirectoryNames
+}
+
+func (m *FakeNodeMonitor) syncDiskStatus(node *longhorn.Node) {
+	// Sync the disks between node.Spec.Disks and node.Status.DiskStatus.
+	if node.Status.DiskStatus == nil {
+		node.Status.DiskStatus = map[string]*longhorn.DiskStatus{}
+	}
+
+	// Aligh node.Spec.Disks and node.Status.DiskStatus.
```
shuo-wu, line 124: Same typo.

controller/node_controller.go
```diff
@@ -1021,3 +871,215 @@ func BackingImageDiskFileCleanup(node *longhorn.Node, bi *longhorn.BackingImage,
...
+func (nc *NodeController) syncOrphans(node *longhorn.Node, onDiskReplicaDirectoryNames map[string]map[string]string) error {
+	if err := nc.cleanupOrphans(node, onDiskReplicaDirectoryNames); err != nil {
+		return errors.Wrapf(err, "failed to clean up orphans")
+	}
+
+	if err := nc.createOrphans(node, onDiskReplicaDirectoryNames); err != nil {
+		return errors.Wrapf(err, "failed to create orphans")
+	}
+
+	if autoDeletion, err := nc.ds.GetSettingAsBool(types.SettingNameOrphanAutoDeletion); err == nil {
```
shuo-wu, line 905: The change of this setting should trigger the node controller reconciliation. I mean, you need to add this setting to https://github.com/longhorn/longhorn-manager/blob/08439666ca2c5cdf70fe53e82860f41eec461c56/controller/node_controller.go#L103

shuo-wu (changes requested):
I will review the test part after making a consensus over the controller & monitor implementation.

util/util.go
```diff
@@ -397,6 +420,15 @@ func GetChecksumSHA512(data []byte) string {
...
 }
 
+func GetStringChecksumSHA256(data string) string {
+	return GetChecksumSHA256([]byte(data))
+}
+
+func GetChecksumSHA256(data []byte) string {
+	checksum := sha256.Sum256(data)
+	return hex.EncodeToString(checksum[:])
+}
```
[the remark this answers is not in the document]
derekbit: Using SHA512 is fine, but the name is too long. Not sure if we will encounter the name conflict after truncating it. So, I use SHA256 here.

controller/monitor/node_monitor.go
```diff
@@ -0,0 +1,328 @@
...
+
+const (
+	NodeMonitorSyncPeriod = 30 * time.Second
+)
+
+type NodeMonitor struct {
+	*baseMonitor
+
+	stateLock sync.RWMutex
+	state     *NodeMonitorState
```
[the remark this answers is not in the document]
derekbit: Yeah. `state` is confused. Do you have a better idea about the rename monitor's `GetState()`?  Is `GetResult()` better?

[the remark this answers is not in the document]
shuo-wu: 1. Directly store disk info & dir info in the monitor struct.
2. Use functions named `GetDisksState()` and `GetDirectories()`/`GetPossibleReplicaDirectories()` would be fine. I am not sure if separating them would be better. These are minor issues overall. You can decide for yourself. 😄

util/util.go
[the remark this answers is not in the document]
shuo-wu: OK. I am fine with these.

k8s/pkg/apis/longhorn/v1beta2/orphan.go
derekbit: No. Because the properties/parameters of the orphaned data are recorded in the map, I will remove the `// +nullable`.

shuo-wu: Sure. And don't forget to re-generate the CRD YAML if necessary.

k8s/pkg/apis/longhorn/v1beta2/orphan.go
```diff
@@ -0,0 +1,70 @@
...
+	OrphanTypeReplicaDirectory = OrphanType("replica directory")
+)
+
+const (
+	OrphanConditionTypeDeletable = "Deletable"
+)
+
+const (
+	OrphanDataName = "DataName"
+	OrphanDiskFsid = "DiskFsid"
```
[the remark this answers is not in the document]
derekbit: Looks `OrphanDiskFsid ` is the ID (name) of the disk and `OrphanDiskUUID ` is the disk UUID, right?

util/util.go
```diff
@@ -787,3 +819,91 @@ func Contains(list []string, item string) bool {
...
+
+	initiatorNSPath := iscsi_util.GetHostNamespacePath(HostProcPath)
+	mountPath := fmt.Sprintf("--mount=%s/mnt", initiatorNSPath)
+	output, err := Execute([]string{}, "nsenter", mountPath, "ls", directory)
+	if err != nil {
+		return replicaDirectoryNames, err
+	}
+
+	names := strings.Split(output, "\n")
+	for _, name := range names {
```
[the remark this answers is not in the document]
derekbit: Yep. I think we can check the replica directories with replica CRs first. If the replica CR exists, don't need to check the volume metafile.

controller/monitor/node_monitor.go
```diff
@@ -0,0 +1,328 @@
...
+}
+
+func (m *NodeMonitor) syncDiskStatus(node *longhorn.Node) {
+	// Sync the disks between node.Spec.Disks and node.Status.DiskStatus.
+	if node.Status.DiskStatus == nil {
+		node.Status.DiskStatus = map[string]*longhorn.DiskStatus{}
+	}
+
+	// Aligh node.Spec.Disks and node.Status.DiskStatus.
+	for id := range node.Spec.Disks {
```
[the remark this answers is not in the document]
derekbit: Looks feasible. So, the Ready and Schedulable will be set in the node controller.

controller/monitor/monitor.go
```diff
@@ -0,0 +1,47 @@
+package monitor
```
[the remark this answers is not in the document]
derekbit: I'm thinking that we can move the backup monitor to `controller/monitor` and reuse the monitor interface in the future.  The helper functions are put in the `engineapi`, but the main monitor logics are in `controller/monitor`.

controller/node_controller.go
derekbit: Can you elaborate it more?

shuo-wu: For example, after adding [the owner reference for backing image manager pods](https://github.com/longhorn/longhorn-manager/blob/51ee963e85e20a0a7b724c96fe67c876756ffd52/controller/backing_image_manager_controller.go#L782), the pod will be cleaned up when [the related backing image manager CR is removed](https://github.com/longhorn/longhorn-manager/blob/51ee963e85e20a0a7b724c96fe67c876756ffd52/datastore/longhorn.go#L1432).

Not sure if a similar logic can be applied to Orphan CRD vs Node CRD. Maybe the Orphan CRD is not node-specific in the future.

k8s/pkg/apis/longhorn/v1beta2/orphan.go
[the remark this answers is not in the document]
shuo-wu: Oh, sorry. You are right.

util/util.go
```diff
@@ -787,3 +819,91 @@ func Contains(list []string, item string) bool {
...
+		err = errors.Wrapf(err, "cannot list replica directories in the disk %v", diskPath)
+	}()
+
+	replicaDirectoryNames = make(map[string]string, 0)
+
+	directory := filepath.Join(diskPath, "replicas")
+
+	initiatorNSPath := iscsi_util.GetHostNamespacePath(HostProcPath)
+	mountPath := fmt.Sprintf("--mount=%s/mnt", initiatorNSPath)
+	output, err := Execute([]string{}, "nsenter", mountPath, "ls", directory)
```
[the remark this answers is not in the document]
shuo-wu: Oh, btw, you add add `-maxdepth 1` and `-mindepth 1` to avoid unnecessary traversal.
```
find <Dir Path> -maxdepth 1 -mindepth 1 -type d -regextype posix-extended -regex ".*-[a-zA-Z0-9]{8}$"
```

util/util.go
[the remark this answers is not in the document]
shuo-wu: Right. I saw that the valid ones are filtered out outside of this function. But it's better to do the filtering before this.

controller/node_controller.go
shuo-wu: Correct. With the current implementation, the node controller should understand **almost all details ** of Orphan CRs, which makes the code redundant as well as future maintenance difficult. In other words, the node controller handles everything for Orphan CRs, so what the orphan controller is for.

controller/orphan_controller.go
```diff
@@ -0,0 +1,371 @@
...
+	}
+
+	ds.OrphanInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    oc.enqueueOrphan,
+		UpdateFunc: func(old, cur interface{}) { oc.enqueueOrphan(cur) },
+		DeleteFunc: oc.enqueueOrphan,
+	})
+	oc.cacheSyncs = append(oc.cacheSyncs, ds.OrphanInformer.HasSynced)
+
+	ds.NodeInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
```
innobead, line 78: Use `AddEventHandlerWithResyncPeriod` to disable resync for non-primary resource watching, the same as other controllers.

controller/orphan_controller.go
```diff
@@ -0,0 +1,371 @@
...
+
+	if !orphan.DeletionTimestamp.IsZero() {
+		defer func() {
+			if err == nil {
+				err = oc.ds.RemoveFinalizerForOrphan(orphan)
+			}
+		}()
+
+		// Make sure if the orphan node and controller ID are same.
+		// If NO, just delete the orphan resource object.
```
innobead, line 229: We probably should use logging for this reason instead of comments only.

controller/orphan_controller.go
```diff
@@ -0,0 +1,371 @@
...
+		node, err := oc.ds.GetNode(orphan.Spec.NodeID)
+		if err != nil {
+			if apierrors.IsNotFound(err) {
+				log.Debugf("Only delete the orphan resource object since node %v does not exist", node.Name)
+				return nil
+			}
+			return err
+		}
+
+		if orphan.Spec.Type == longhorn.OrphanTypeReplicaDirectory &&
```
innobead, line 243: I feel we should use `longhorn.OrphanTypeReplica` better than `OrphanTypeReplicaDirectory` because it's much straightforward, but it will be not granular a bit for sure.

controller/orphan_controller.go
```diff
@@ -0,0 +1,371 @@
...
+			types.GetCondition(orphan.Status.Conditions, longhorn.OrphanConditionTypeUnavailable).Status == longhorn.ConditionStatusFalse {
+			err := oc.deleteOrphanedData(orphan)
```
innobead, line 245: How about using the specific method name instead? like `oc.deleteOrphanedReplica` or if you want to `deleteOrphanedData` adaptively, then you should just move all type check into `deleteorphanedData`. KISS is better.

innobead (changes requested):

shuo-wu (changes requested):
Maybe we need a review session for the monitor part

controller/orphan_controller.go
```diff
@@ -0,0 +1,371 @@
...
+		if err != nil {
+			return
+		}
+
+		status := longhorn.ConditionStatusFalse
+		if reason != "" {
+			status = longhorn.ConditionStatusTrue
+		}
+
+		orphan.Status.Conditions = types.SetCondition(orphan.Status.Conditions, longhorn.OrphanConditionTypeUnavailable, status, reason, "")
```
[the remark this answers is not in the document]
derekbit: 1. Yeah, using a positive is much clearer.
2. `Available` means the data (aka replica directory) is accessible.
3. I think no, so the condition type should be more generic. `Cleanable` sounds good. How about it?

controller/orphan_controller.go
```diff
@@ -0,0 +1,371 @@
...
+	}()
+
+	if autoDeletion, err := oc.ds.GetSettingAsBool(types.SettingNameOrphanAutoDeletion); err == nil {
+		if autoDeletion {
+			reason = longhorn.OrphanConditionTypeUnavailableReasonAutoDeletionEnabled
+			return nil
+		}
+	} else {
+		return errors.Wrapf(err, "failed to get %v setting", types.SettingNameOrphanAutoDeletion)
+	}
```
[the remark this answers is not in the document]
derekbit: Do you mean the node controller deletes the orphan CRs if the setting is set to true?

controller/monitor/node_monitor.go
```diff
@@ -0,0 +1,377 @@
...
+
+func (m *NodeMonitor) prunePossibleReplicaDirectoryNames(node *longhorn.Node, diskPath string, replicaDirectoryNames map[string]string) (map[string]string, error) {
+	// Find out the orphaned directories by checking with replica CRs
+	replicas, err := m.ds.ListReplicasByNodeRO(node.Name)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, replica := range replicas {
+		delete(replicaDirectoryNames, replica.Spec.DataDirectoryName)
```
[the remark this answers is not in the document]
derekbit: Yeah, both contain replica dir name only.

webhook/resources/orphan/mutator.go
```diff
@@ -0,0 +1,88 @@
...
+	// Add labels according to the orphan type
+	var longhornLabels map[string]string
+
+	switch {
+	case orphan.Spec.Type == longhorn.OrphanTypeReplicaDirectory:
+		longhornLabels = types.GetOrphanLabelsForOrphanedDirectory(orphan.Spec.NodeID, orphan.Spec.Parameters[longhorn.OrphanDiskUUID])
+	}
+	if longhornLabels == nil {
+		return nil, werror.NewInvalidError("invalid orphan labels", "")
+	}
```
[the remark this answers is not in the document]
derekbit: We need to marshal and update the labels here, so check the label here instead of in the validator.

controller/monitor/node_monitor.go
```diff
@@ -0,0 +1,377 @@
...
+	node, err := m.ds.GetNode(m.nodeName)
+	if err != nil {
+		err = errors.Wrapf(err, "failed to get longhorn node %v", m.nodeName)
+		return err
+	}
+
+	pruneDiskStatus(node)
+
+	diskStatusMap, diskInfoMap := m.syncDiskStatus(node)
+	onDiksReplicaDirectoryNames := m.getOnDiskReplicaDirectoryNames(node)
```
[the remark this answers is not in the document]
derekbit: In `fsid2Disks` iteration, the logic will get the `getDiskConfig`  and do `generateDiskConfig`. The two operations may fail and return error, so I update the ready condition of the `diskStatus` here.

We can move the condition updates to the node controller, but we need to add other fields to record the errors, instead. Thus, to simplify it, I recorded the errors in the `diskStatus.condition`  in node monitor. Then, the node controller gets the status from monitor and update the `node.diskStatus.condition` in the node controller.

controller/monitor/node_monitor.go
```diff
@@ -0,0 +1,377 @@
...
+func (m *NodeMonitor) GetCollectedData() (interface{}, error) {
+	m.collectedDataLock.RLock()
+	defer m.collectedDataLock.RUnlock()
+
+	data := &NodeMonitorCollectedData{}
+	if err := copier.Copy(data, m.collectedData); err != nil {
+		return nil, errors.Wrapf(err, "failed to copy node monitor collected data")
+	}
+
+	return data, nil
```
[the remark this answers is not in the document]
derekbit: Hmm...? Can you elaborate it more?

controller/orphan_controller.go
```diff
@@ -0,0 +1,371 @@
...
+		if err != nil {
+			if apierrors.IsNotFound(err) {
+				log.Debugf("Only delete the orphan resource object since node %v does not exist", node.Name)
+				return nil
+			}
+			return err
+		}
+
+		if orphan.Spec.Type == longhorn.OrphanTypeReplicaDirectory &&
+			types.GetCondition(orphan.Status.Conditions, longhorn.OrphanConditionTypeUnavailable).Status == longhorn.ConditionStatusFalse {
```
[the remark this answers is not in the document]
derekbit: yeah. I will refactor and simplify this part.

controller/orphan_controller.go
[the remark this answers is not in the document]
shuo-wu: Yes. This controller does not care when to delete the CRs

manager/orphan.go
```diff
@@ -0,0 +1,36 @@
...
+}
+
+func (m *VolumeManager) ListOrphansSorted() ([]*longhorn.Orphan, error) {
+	orphanMap, err := m.ds.ListOrphans()
+	if err != nil {
+		return []*longhorn.Orphan{}, err
+	}
+
+	orphans := make([]*longhorn.Orphan, len(orphanMap))
+	orphanNames, err := sortKeys(orphanMap)
```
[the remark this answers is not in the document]
derekbit: No, this is an old implementation. I think we don't need the sort here.

controller/monitor/node_monitor.go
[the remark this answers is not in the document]
shuo-wu: I mean, can we return the struct `NodeMonitorCollectedData` rather than the struct pointer here?

controller/monitor/node_monitor.go
```diff
@@ -0,0 +1,378 @@
...
+	m.collectedData.DiskInfoMap = diskInfoMap
+	m.collectedData.OnDiskReplicaDirectoryNames = onDiskReplicaDirectoryNames
+}
+
+func (m *NodeMonitor) syncDiskStatus(node *longhorn.Node) (diskStatusMap map[string]*longhorn.DiskStatus, diskInfoMap map[string]*DiskInfo) {
+	diskStatusMap = copyDiskStatus(node.Status.DiskStatus)
+	diskInfoMap = m.getDiskInfoMap(node)
+
+	fsid2Disks := map[string][]string{}
+	for id, info := range diskInfoMap {
```
shuo-wu: `diskName, info := range diskInfoMap`

controller/monitor/node_monitor.go
```diff
@@ -0,0 +1,378 @@
...
+	volumeMetaData = "volume.meta"
+)
+
+type NodeMonitor struct {
+	*baseMonitor
+
+	nodeName string
+
+	collectedDataLock sync.RWMutex
+	collectedData     *NodeMonitorCollectedData
```
shuo-wu, line 37: I am thinking `collectedData` should be like:
```
collectedData map[string]*CollectedDiskInfo. // key is the disk name.

type CollectedDiskInfo struct {
	util.DiskInfo
	ReadyCondition longhorn.Condition. // Error will be directly recorded here
	ReplicaDirectoryNames map[string]string
}
```

controller/monitor/node_monitor.go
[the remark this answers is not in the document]
derekbit: yeah, we can.

controller/monitor/node_monitor.go
```diff
@@ -0,0 +1,378 @@
...
+func (m *NodeMonitor) updateCollectedData(diskStatusMap map[string]*longhorn.DiskStatus, diskInfoMap map[string]*DiskInfo, onDiskReplicaDirectoryNames map[string]map[string]string) {
+	m.collectedDataLock.Lock()
+	defer m.collectedDataLock.Unlock()
+
+	m.collectedData.DiskStatusMap = diskStatusMap
+	m.collectedData.DiskInfoMap = diskInfoMap
+	m.collectedData.OnDiskReplicaDirectoryNames = onDiskReplicaDirectoryNames
+}
+
+func (m *NodeMonitor) syncDiskStatus(node *longhorn.Node) (diskStatusMap map[string]*longhorn.DiskStatus, diskInfoMap map[string]*DiskInfo) {
```
shuo-wu: Then this function can be like:
1. Load the disk info
2. Get or (blindly) generate the disk UUID

Here, I prefer to ask the node controller itself to handle the duplicate fsid issue. Since this monitor actually does not know which disk should be marked as NotReady or not when multiple paths hold the same fsid. 
For example, there is only one disk on the node at the beginning and everything works fine. Later on, users add some sub-directories of the current disk path as the new disks for the node (and some of the subdir contain DiskUUID). Then Longhorn should mark all disks **except for 1st one** as NotReady. The monitor does not know the previous disk status info (which disk is already considered Ready) hence it cannot handle this case.

controller/orphan_controller.go
```diff
@@ -0,0 +1,387 @@
...
+	switch {
+	case orphan.Spec.Type == longhorn.OrphanTypeReplica:
+		reason = checkOrphanedReplicaUnavailable(node, orphan)
+	}
+
+	return nil
+}
+
+func checkOrphanedReplicaUnavailable(node *longhorn.Node, orphan *longhorn.Orphan) string {
+	id := orphan.Spec.Parameters[longhorn.OrphanDiskFsid]
```
shuo-wu: Should use `DiskUUID` instead. 
BTW, it's better to name the disk-related fields as `id`. Instead, using `fsid`, `diskName`, or `diskUUID`.

controller/orphan_controller.go
```diff
@@ -0,0 +1,387 @@
...
+	if disk, ok := node.Spec.Disks[id]; ok {
```
shuo-wu: As I mentioned, please find a matching disk by `DiskUUID` rather than `diskName` or `fsid`.

controller/orphan_controller.go
```diff
@@ -0,0 +1,387 @@
...
+
+	// Make sure if the orphan nodeID and controller ID are same.
+	// If NO, just delete the orphan resource object and don't touch the data.
+	if orphan.Spec.NodeID != oc.controllerID {
+		log.Infof("Orphan nodeID %v is different from controllerID %v, so just delete the orphan resouce object",
+			orphan.Name, oc.controllerID)
+		return nil
+	}
+
+	if types.GetCondition(orphan.Status.Conditions, longhorn.OrphanConditionTypeDataCleanable).Status ==
```
shuo-wu, line 286: Question: what if the Orphan CR is being deleted while the condition field is still empty?

controller/orphan_controller.go
```diff
@@ -0,0 +1,387 @@
...
+			"orphan": orphan.Name,
+		},
+	)
+}
+
+func (oc *OrphanController) isResponsibleFor(orphan *longhorn.Orphan) bool {
+	return isControllerResponsibleFor(oc.controllerID, oc.ds, orphan.Name, orphan.Spec.NodeID, orphan.Status.OwnerID)
+}
+
+func (oc *OrphanController) deleteOrphanedData(orphan *longhorn.Orphan) (err error) {
```
shuo-wu, line 265: NIT: rename this to `cleanup` or `doCleanupForOrphan`. When I saw word `data`, I was thinking this function would directly delete the replica directory data.

shuo-wu (commented):
Sorry for not realizing some issues during today's review session, especially for the data struct held by the node monitor. If you have any questions, we can continue the discussion tomorrow.

controller/monitor/node_monitor.go
derekbit: Okay. The fsid handling will be moved to node controller.

controller/orphan_controller.go
derekbit: weird, this is old code...Why isn't the code updated ? :(

controller/orphan_controller.go
derekbit: okay!

controller/orphan_controller.go
derekbit: This is also old codes... :(

controller/orphan_controller.go
derekbit: Good catch!
For CR operations, can the deletion happen before the creation?
If YES, I incline to delete the orphan CR only and don't touch the orphaned data.

derekbit: So, the following change will be better.

```
	if types.GetCondition(orphan.Status.Conditions, longhorn.OrphanConditionTypeDataCleanable).Status !=
		longhorn.ConditionStatusTrue {
		log.Infof("Only delete orphan %v resource object and do not delete the orphaned data", orphan.Name)
		return nil
	}
```

shuo-wu: The CR can be marked as deletion immediately after the creation. For example, when the auto-deletion setting is enabled, will the node controller wait for the condition update before sending the deletion request?

derekbit: No, the node controller won't wait for it. You are right. The node controller should check the condition before sending the deletion.

controller/node_controller.go
```diff
@@ -1018,3 +1001,215 @@ func BackingImageDiskFileCleanup(node *longhorn.Node, bi *longhorn.BackingImage,
...
+	for _, orphan := range orphans {
+		if orphan.Status.OwnerID != nc.controllerID {
+			continue
+		}
+
+		// If the on-disk data is missing, e.g. deleted by users by manual, we need to delete the orphaned CR.
+		// Orphan controller won't stat/check the on-disk data and isn't aware the of the missing replica directory,
+		// so we check and handle it here.
+		if orphan.Spec.Type == longhorn.OrphanTypeReplica {
+			onDiskDataExist, err := nc.isOnDiskDataExist(orphan, diskInfo.OrphanedReplicaDirectoryNames)
```
shuo-wu, line 1061: I think this is redundant:
    1. If the dir suddenly disappears, the orphan controller will delete the CR only.
    2. The CR creation is based on `diskInfo.OrphanedReplicaDirectoryNames`. We don't need to verify it back and force.

controller/node_controller.go
```diff
@@ -1018,3 +1001,215 @@ func BackingImageDiskFileCleanup(node *longhorn.Node, bi *longhorn.BackingImage,
...
+
+	orphans, err := nc.ds.ListOrphansRO()
+	if err != nil {
+		return errors.Wrap(err, "unable to list orphans")
+	}
+
+	for _, orphan := range orphans {
+		if orphan.Status.OwnerID != nc.controllerID {
+			continue
+		}
```
shuo-wu, line 1055: This implicitly means that Longhorn won't automatically delete the Orphan CRs on the down/removed nodes, right?

controller/node_controller.go
```diff
@@ -1018,3 +1001,215 @@ func BackingImageDiskFileCleanup(node *longhorn.Node, bi *longhorn.BackingImage,
...
+	}
+
+	collectedDiskInfo := v.(map[string]*monitor.CollectedDiskInfo)
+	if !isDiskMatched(node, collectedDiskInfo) {
+		return nil, errors.New("mismatching disks in node resource object and monitor collected data")
+	}
+	return collectedDiskInfo, nil
+}
+
+func isDiskMatched(node *longhorn.Node, collectedDiskInfo map[string]*monitor.CollectedDiskInfo) bool {
```
shuo-wu, line 1147: Print out the mismatching reason.

BTW, the mismatching should be caused by the monitor syncing up the status before the disk spec update, right? And the mismatching should disappear in the next sync.

controller/node_controller.go
```diff
@@ -1018,3 +1001,215 @@ func BackingImageDiskFileCleanup(node *longhorn.Node, bi *longhorn.BackingImage,
...
+func (nc *NodeController) checkMonitor(node *longhorn.Node) (monitor.Monitor, error) {
+	if node == nil {
+		return nil, nil
+	}
+
+	if nc.monitor != nil {
+		return nc.monitor, nil
+	}
+
+	monitor, err := monitor.NewNodeMonitor(nc.logger, nc.ds, node, nc.enqueueNodeForMonitor)
```
shuo-wu, line 1014: 1. Will the node controller take over the unknown node CR?
2. Will it create monitors for those unknown node CRs?

controller/monitor/fake_node_monitor.go
```diff
@@ -0,0 +1,174 @@
...
+	collectedData     map[string]*CollectedDiskInfo
+
+	syncCallback func(key string)
+
+	getDiskStatHandler GetDiskStatHandler
+	getDiskConfig      GetDiskConfig
+	generateDiskConfig GenerateDiskConfig
+}
+
+func NewFakeNodeMonitor(logger logrus.FieldLogger, ds *datastore.DataStore, node *longhorn.Node, syncCallback func(key string)) (*FakeNodeMonitor, error) {
```
shuo-wu, line 44: You can input `getDiskStatHandler`, `getDiskConfig`, and `generateDiskConfig`. Then you don't need this Fake Monitor.

shuo-wu (changes requested):

controller/node_controller.go
derekbit: > And the mismatching should disappear in the next sync.

yeah

controller/node_controller.go
derekbit: orphan controller does not track the existence of the orphaned directory, so the step is to check if the orphan CR's dir is still in the latest `diskInfo.OrphanedReplicaDirectoryNames`. If not, the orphan CR needs to be deleted.

controller/node_controller.go
```diff
@@ -536,153 +551,102 @@ func (nc *NodeController) enqueueKubernetesNode(obj interface{}) {
...
 				diskStatus.DiskUUID = diskUUID
 			} else { // diskStatusMap[id].DiskUUID != ""
 				if diskUUID == "" {
-					diskStatusMap[id].Conditions = types.SetConditionAndRecord(diskStatusMap[id].Conditions,
+					diskStatusMap[diskName].Conditions = types.SetConditionAndRecord(diskStatusMap[diskName].Conditions,
 						longhorn.DiskConditionTypeReady, longhorn.ConditionStatusFalse,
 						string(longhorn.DiskConditionReasonDiskFilesystemChanged),
-						fmt.Sprintf("Disk %v(%v) on node %v is not ready: cannot find disk config file, maybe due to a mount error", id, disk.Path, node.Name),
+						fmt.Sprintf("Disk %v(%v) on node %v is not ready: cannot find disk config file, maybe due to a mount error",
+							diskName, diskInfoMap[diskName].Path, node.Name),
```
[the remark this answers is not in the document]
derekbit: Yes. If the disk condition is set to false with a reason, the DiskUUID is not used and is empty.

controller/node_controller.go
```diff
@@ -691,35 +655,54 @@ func (nc *NodeController) syncDiskStatus(node *longhorn.Node) error {
...
+				usableStorage := (diskInfoMap[diskName].DiskStat.StorageAvailable / truncateTo) * truncateTo
+				node.Status.DiskStatus[diskName].StorageAvailable = usableStorage
+				node.Status.DiskStatus[diskName].StorageMaximum = diskInfoMap[diskName].DiskStat.StorageMaximum
+				diskStatusMap[diskName].Conditions = types.SetConditionAndRecord(diskStatusMap[diskName].Conditions,
+					longhorn.DiskConditionTypeReady, longhorn.ConditionStatusTrue,
+					"", fmt.Sprintf("Disk %v(%v) on node %v is ready", diskName, diskInfoMap[diskName].Path, node.Name),
+					nc.eventRecorder, node, v1.EventTypeNormal)
+			}
+			diskStatusMap[diskName] = diskStatus
+		}
```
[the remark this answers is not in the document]
derekbit: Sorry, this is redundant.

controller/monitor/node_monitor.go
```diff
@@ -0,0 +1,225 @@
...
+						diskName, disk.Path, node.Name, err))
+				continue
+			}
+		}
+
+		var orphanedReplicaDirectoryNames map[string]string
+
+		if node.Status.DiskStatus != nil &&
+			node.Status.DiskStatus[diskName] != nil &&
+			types.GetCondition(node.Status.DiskStatus[diskName].Conditions, longhorn.DiskConditionTypeReady).Status == longhorn.ConditionStatusTrue {
```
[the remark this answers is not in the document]
derekbit: > This means the node monitor relies on the previous disk status recorded in the node CR, which is unexpected.

Yes. collect the orphaned directories after the disks' status are ready. In this logic, we won't scan the non-ready disks' orphaned directories.

> Besides, Matching disks with diskName is inaccurate

I think we need to check diskPath and diskUUID here.

controller/node_controller.go
derekbit: The `OwnerID` of an orphan CR, whose node is down or removed, will be updated to another NodeID.

The orphan controller checks if the orphan CR' node is missing or not. If missing, `DataCleanable` condition is set to `False`, node controller will clean it up.

`if orphan.Status.OwnerID != nc.controllerID {` is to ensure and prevent from the mistaken deletion of data.

controller/node_controller.go
derekbit: What's the unknown node CRs?
Do you mean  `nc.controllerID != node.Name`?
If yes, won't create monitors for the unknown node CRs.

controller/node_controller.go
```diff
@@ -536,153 +551,102 @@ func (nc *NodeController) enqueueKubernetesNode(obj interface{}) {
...
-						string(longhorn.DiskConditionReasonNoDiskInfo),
-						fmt.Sprintf("Disk %v(%v) on node %v is not ready: failed to get disk config: error: %v", id, disk.Path, node.Name, err),
-						nc.eventRecorder, node, v1.EventTypeWarning)
-					continue
-				}
-			} else {
-				diskUUID = diskConfig.DiskUUID
-			}
+	for fsid, diskInfoMap := range healthyDiskInfoMap {
+		for diskName, info := range diskInfoMap {
```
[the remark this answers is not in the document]
derekbit: This logic is from the previous `syncDiskStatus`.
But why do we need to check the path matching?

api/orphan.go
```diff
@@ -0,0 +1,52 @@
...
+	orphans, err := s.m.ListOrphans()
+	if err != nil {
+		return errors.Wrap(err, "failed to list instance managers")
+	}
+
+	apiContext.Write(toOrphanCollection(orphans))
+	return nil
+}
+
+func (s *Server) orphanList(apiContext *api.ApiContext) (*client.GenericCollection, error) {
```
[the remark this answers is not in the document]
derekbit: For `orphanListStream := NewStreamHandlerFunc("orphans", s.wsc.NewWatcher("orphan"), s.orphanList)`,
not really sure if we will use it.

controller/monitor/fake_node_monitor.go
derekbit: But the collections of orphaned directory names in the two monitors are different, I have no better idea handle them.

controller/node_controller.go
```diff
@@ -536,153 +551,102 @@ func (nc *NodeController) enqueueKubernetesNode(obj interface{}) {
...
 }
 
-func (nc *NodeController) getDiskInfoMap(node *longhorn.Node) map[string]*diskInfo {
-	result := map[string]*diskInfo{}
-	for id, disk := range node.Spec.Disks {
-		info, err := nc.getDiskInfoHandler(disk.Path)
-		result[id] = &diskInfo{
-			entry: info,
-			err:   err,
+func findFailedAndHealthyDisks(collectedDataInfo map[string]*monitor.CollectedDiskInfo) (failedDiskInfo, healthyDiskInfo map[string]map[string]*monitor.CollectedDiskInfo) {
```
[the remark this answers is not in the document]
derekbit: > NIT: Actually we can filter out the invalid disks here, then blindly do the update in the following function.

Not really sure if this is better. Seems just merge the `syncDiskStatus` and this function. The logic becomes more complicated...

controller/monitor/fake_node_monitor.go
derekbit: Addressed!

controller/node_controller.go
```diff
@@ -1018,3 +998,213 @@ func BackingImageDiskFileCleanup(node *longhorn.Node, bi *longhorn.BackingImage,
...
+			if otherName != diskName && diskReady.Status == longhorn.ConditionStatusTrue {
+				return true
+			}
+		}
+	}
+
+	return false
+}
+
+func getDuplicatedDisks(diskInfoMap map[string]*monitor.CollectedDiskInfo) []string {
```
shuo-wu, line 1204: `GetDiskNamesFromDiskMap`

controller/node_controller_test.go
```diff
@@ -656,9 +739,22 @@ func (s *TestSuite) TestSyncNode(c *C) {
...
 		}
+
 		// sync node status
 		for nodeName, node := range tc.nodes {
-			err := nc.syncNode(getKey(node, c))
+			if nc.controllerID == node.Name {
+				enqueueNodeForMonitor := func(key string) {
+					nc.queue.Add(key)
+				}
+				mon, err := monitor.NewFakeNodeMonitor(nc.logger, nc.ds, node, enqueueNodeForMonitor)
```
shuo-wu, line 749: You can init this fake monitor in `newTestNodeController`

shuo-wu (commented):
Can we add a unit test for the orphan CR auto-cleanup in node_controller_test?

controller/node_controller.go
shuo-wu: OK, got what you mean. It's like the alignment between the CRs and the collected dir list. Can you add a log before this deletion/alignment?

derekbit: okay

controller/node_controller.go
derekbit: I'm thinking this logic is redundant.

controller/node_controller.go
```diff
@@ -536,190 +552,154 @@ func (nc *NodeController) enqueueKubernetesNode(obj interface{}) {
...
-					}
-					diskUUID = diskConfig.DiskUUID
-				}
-				diskStatus.DiskUUID = diskUUID
-			} else { // diskStatusMap[id].DiskUUID != ""
-				if diskUUID == "" {
-					diskStatusMap[id].Conditions = types.SetConditionAndRecord(diskStatusMap[id].Conditions,
-						longhorn.DiskConditionTypeReady, longhorn.ConditionStatusFalse,
+			if nc.isFSIDDuplicatedWithExistingReadyDisk(diskName, diskInfoMap, node.Status.DiskStatus) ||
+				len(readyDiskInfoMap[fsid]) == 1 {
```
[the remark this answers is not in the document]
derekbit: It's a bit of confusing.
The abstraction is clear. If one ready disk is found for the fsid, other disks with the same fsid are regarded as not-ready disks.

```
func isReadyDiskFound(diskInfoMap map[string]*monitor.CollectedDiskInfo) bool {
	return len(diskInfoMap) > 0
}
```

controller/node_controller_test.go
derekbit: addressed!

derekbit:
>  Can we add a unit test for the orphan CR auto-cleanup in node_controller_test?

I don't find a better way to introduce an auto-cleanup test in node_controller_test, because it's related to the on-disk data deletion in the orphan controller.
Can we address it in another ticket?

Currently, the [e2e](https://github.com/longhorn/longhorn-tests/pull/940) includes the auto-cleanup test case.

controller/node_controller.go
```diff
@@ -536,190 +552,154 @@ func (nc *NodeController) enqueueKubernetesNode(obj interface{}) {
...
-	return false
+	return nc.updateDiskStatusSchedulableCondition(node)
 }
 
-func (nc *NodeController) syncDiskStatus(node *longhorn.Node) error {
-	log := getLoggerForNode(nc.logger, node)
+func (nc *NodeController) findNotReadyAndReadyDisks(node *longhorn.Node, collectedDataInfo map[string]*monitor.CollectedDiskInfo) (notReadyDiskInfoMap, readyDiskInfoMap map[string]map[string]*monitor.CollectedDiskInfo) {
+	notReadyDiskInfoMap = make(map[string]map[string]*monitor.CollectedDiskInfo, 0)
+	readyDiskInfoMap = make(map[string]map[string]*monitor.CollectedDiskInfo, 0)
+	duplicatedDiskInfoMap := make(map[string]map[string]*monitor.CollectedDiskInfo, 0)
```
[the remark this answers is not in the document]
shuo-wu: Here, word `duplicate` is inaccurate, you will group disks by fsid regardless of duplicates first.

controller/node_controller.go
```diff
@@ -536,190 +552,154 @@ func (nc *NodeController) enqueueKubernetesNode(obj interface{}) {
...
+		nc.updateReadyDiskStatusReadyCondition(node, fsid, diskInfoMap)
 	}
 
-	return false
+	return nc.updateDiskStatusSchedulableCondition(node)
 }
 
-func (nc *NodeController) syncDiskStatus(node *longhorn.Node) error {
-	log := getLoggerForNode(nc.logger, node)
+func (nc *NodeController) fsidDiskGroupMap(node *longhorn.Node, collectedDataInfo map[string]*monitor.CollectedDiskInfo) (notReadyDiskInfoMap, readyDiskInfoMap map[string]map[string]*monitor.CollectedDiskInfo) {
```
shuo-wu, line 571: Sorry, I mean renaming the below variable rather than this function... My bad...

controller/node_controller.go
shuo-wu: The condition `DataCleanable` would tell if the data can be removed or not. It won't help the current node controller skip the handling of the orphans on other nodes...

controller/node_controller.go
derekbit: That's okay. But maybe `fsidDiskToGroupMap` is also a good name for this function.

controller/node_controller.go
derekbit: > If yes, why not directly use ListOrphansByNodeRO?

The label is still equal to the `orphan.spec.NodeID`, so I don't use `ListOrphansByNodeRO` here.

shuo-wu: You mean this check is already included in `isOnDiskDataMissing`? If there are limited Orphan CRs in the cluster, I am fine with either. Otherwise, I prefer to filter them at the beginning (via the ListOption).

shuo-wu (approved):
In general LGTM.

controller/node_controller.go
derekbit: yeah, I also prefer to filter them first.

derekbit: Addressed.

types/setting.go
```diff
@@ -757,6 +761,16 @@ var (
 		ReadOnly: false,
 		Default:  "false",
 	}
+
+	SettingDefinitionOrphanAutoDeletion = SettingDefinition{
+		DisplayName: "Orphan Auto-Deletion",
+		Description: "This setting allows Longhorn to delete the orphaned data and its corresponding resource automatically.",
```
shuo-wu, line 767: NIT: mentioning that the Orphans on the down/unknown nodes won't be cleaned up automatically

shuo-wu (approved):

types/setting.go
derekbit: Precisely, the orphan CR's `DataCleanable` condition will be set to be `false` in the orphan controller because the node is down, but the orphaned data won't be deleted.

shuo-wu: I knew. 
It's OrphanController's duty to check the condition and delete the actual data. 
For this setting, it's NodeController that relies on it to determine if a deletion timestamp can be added to the CRs.

derekbit: Got your point. Thank you.

shuo-wu (dismissed):

webhook/resources/orphan/validator.go
```diff
@@ -0,0 +1,96 @@
...
+func (o *orphanValidator) Update(request *admission.Request, oldObj runtime.Object, newObj runtime.Object) error {
+	oldOrphan := oldObj.(*longhorn.Orphan)
+	newOrphan := newObj.(*longhorn.Orphan)
+
+	if err := checkOrphanParameters(newOrphan); err != nil {
+		return werror.NewInvalidError(err.Error(), "")
+	}
+
+	if !reflect.DeepEqual(oldOrphan.Spec, newOrphan.Spec) {
+		return werror.NewInvalidError(fmt.Sprintf("orphan %v spec fileds are immutable", oldOrphan.Name), "")
```
innobead, line 65: This is a good check for immutability.

controller/node_controller.go
```diff
@@ -1018,3 +1003,250 @@ func BackingImageDiskFileCleanup(node *longhorn.Node, bi *longhorn.BackingImage,
...
+				continue
+			}
+		}
+
+		dataCleanableCondition := types.GetCondition(orphan.Status.Conditions, longhorn.OrphanConditionTypeDataCleanable)
+		if dataCleanableCondition.Status == longhorn.ConditionStatusUnknown {
+			continue
+		}
+
+		if autoDeletionEnabled || dataCleanableCondition.Status == longhorn.ConditionStatusFalse {
```
innobead, line 1080: Should it be `if autoDeletionEnabled || dataCleanableCondition.Status == longhorn.ConditionStatusTrue`?

innobead (commented):
In general, LGTM. Just have a few comments waiting to resolve, then we can merge this.

controller/node_controller.go
derekbit: When orphan controller detects a node is evicted, disk is evicted or something happened, it means the data is not accessible and not cleanable anymore. Thus `dataCleanableCondition.Status` of the orphan CR is set to  `longhorn.ConditionStatusFalse`.  

Then, this logic is to clean up the orphan CR that their data is not accessible, so check `dataCleanableCondition.Status == longhorn.ConditionStatusTrue` here.

innobead: Thanks for the explanation!

innobead (approved):

PR #1289, opened 2022-04-11
Rename rpc.proto to imrpc.proto

derekbit:
To avoid the "protobuf namespace conflicts" error, rename the protocol buffer file.

Depends on https://github.com/longhorn/longhorn-instance-manager/pull/108

https://github.com/longhorn/longhorn/issues/3796


go.mod
```diff
@@ -75,3 +75,5 @@ require (
 	sigs.k8s.io/controller-runtime v0.4.0
 	sigs.k8s.io/yaml v1.2.0
 )
+
+replace github.com/longhorn/longhorn-instance-manager v0.0.0-20210729081215-50c310f97378 => github.com/derekbit/longhorn-instance-manager v0.0.0-20220411065951-c052142954b7
```
innobead, line 79: Why using replace instead of updating the required dep directly?

innobead (approved):
LGTM

After updating this, if running `go mod tidy` then `go test or make test`, all tests passed w/o any proto registration error?

go.mod
innobead: Oh, we are still waiting for https://github.com/longhorn/longhorn-instance-manager/pull/108 😄

derekbit:
> LGTM
> 
> After updating this, if running `go mod tidy` then `go test or make test`, all tests passed w/o proto registration error anymore, right?

yeah, works as expected while doing
```
go mod tidy
go mod vendor
make or make test
```

innobead (approved):

PR #1292, opened 2022-04-11
MTLS auth

joshimoo:
tracking issue longhorn/longhorn#3839

Currently running core tests, my confidence in the manager <-> im communication isn't as high as it should be.
I would prefer if we clean up the monitoring + im communication in general before we add additional complexity to it.
The previous code/architecture wasn't built with peristent connections in mind and instead is completly built as one shot commands, which is highly inefficient.

Since for each grpc call: 
- a single connection is setup (tcp handshake + co) 
- then grpc call is triggered
- connection teardown

Some of the code smells surrounding the manager <-> im communication:

- im monitor
- instance-handler usage, same one shot idea.
- im notifier / updater (I removed these wrappers and replaced them with a direct call to the im client)

Generally In the future I would like our code base to become more testable, for example we can mock on the im monitor/client level then the full im controller should be testable same with engine/replica controller.

I left most of these refactorings for longhorn/longhorn#2441

### DIFF
--- a/controller/engine_controller.go
+++ b/controller/engine_controller.go
@@ -557,7 +557,7 @@ func (ec *EngineController) GetInstance(obj interface{}) (*longhorn.InstanceProc
 	return c.ProcessGet(e.Name)
 }
 
-func (ec *EngineController) LogInstance(obj interface{}) (*imapi.LogStream, error) {
+func (ec *EngineController) LogInstance(ctx context.Context, obj interface{}) (*imapi.LogStream, error) {
 	e, ok := obj.(*longhorn.Engine)
 	if !ok {
 		return nil, fmt.Errorf("BUG: invalid object for engine process log: %v", obj)
@@ -572,7 +572,7 @@ func (ec *EngineController) LogInstance(obj interface{}) (*imapi.LogStream, erro
 		return nil, err
 	}
 
-	return c.ProcessLog(e.Name)
+	return c.ProcessLog(ctx, e.Name)
 }
 
 func (ec *EngineController) isMonitoring(e *longhorn.Engine) bool {
--- a/controller/engine_controller.go
+++ b/controller/engine_controller.go
@@ -560,22 +560,24 @@ func (ec *EngineController) GetInstance(obj interface{}) (*longhorn.InstanceProc
 	return c.ProcessGet(e.Name)
 }
 
-func (ec *EngineController) LogInstance(ctx context.Context, obj interface{}) (*imapi.LogStream, error) {
+func (ec *EngineController) LogInstance(ctx context.Context, obj interface{}) (*engineapi.InstanceManagerClient, *imapi.LogStream, error) {
 	e, ok := obj.(*longhorn.Engine)
 	if !ok {
-		return nil, fmt.Errorf("BUG: invalid object for engine process log: %v", obj)
+		return nil, nil, fmt.Errorf("BUG: invalid object for engine process log: %v", obj)
 	}
 
 	im, err := ec.ds.GetInstanceManager(e.Status.InstanceManagerName)
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 	c, err := engineapi.NewInstanceManagerClient(im)
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 
-	return c.ProcessLog(ctx, e.Name)
+	// TODO: #2441 refactor this when we do the resource monitoring refactor
+	stream, err := c.ProcessLog(ctx, e.Name)
+	return c, stream, err
 }
 
 func (ec *EngineController) isMonitoring(e *longhorn.Engine) bool {
--- a/controller/engine_controller.go
+++ b/controller/engine_controller.go
@@ -396,6 +396,7 @@ func (ec *EngineController) CreateInstance(obj interface{}) (*longhorn.InstanceP
 	if err != nil {
 		return nil, err
 	}
+	defer c.Close()
 
 	return c.EngineProcessCreate(e.Name, e.Spec.VolumeName, e.Spec.EngineImage, frontend, e.Status.CurrentReplicaAddressMap, e.Spec.RevisionCounterDisabled, e.Spec.SalvageRequested)
 }
@@ -462,6 +463,7 @@ func (ec *EngineController) DeleteInstance(obj interface{}) error {
 	if err != nil {
 		return err
 	}
+	defer c.Close()
 	if err := c.ProcessDelete(e.Name); err != nil && !types.ErrorIsNotFound(err) {
 		return err
 	}
@@ -553,6 +555,7 @@ func (ec *EngineController) GetInstance(obj interface{}) (*longhorn.InstanceProc
 	if err != nil {
 		return nil, err
 	}
+	defer c.Close()
 
 	return c.ProcessGet(e.Name)
 }
@@ -1452,6 +1455,7 @@ func (ec *EngineController) UpgradeEngineProcess(e *longhorn.Engine) error {
 	if err != nil {
 		return err
 	}
+	defer c.Close()
 
 	engineProcess, err := c.EngineProcessUpgrade(e.Name, e.Spec.VolumeName, e.Spec.EngineImage, frontend, e.Spec.UpgradedReplicaAddressMap)
 	if err != nil {
--- a/controller/instance_handler.go
+++ b/controller/instance_handler.go
@@ -1,6 +1,7 @@
 package controller
 
 import (
+	"context"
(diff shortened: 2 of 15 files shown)


innobead:
@joshimoo Please help rebase from the master again because the dep issue has been resolved.

joshimoo:
Can be reviewed, I will need to update the im ref once https://github.com/longhorn/longhorn-instance-manager/pull/109
is merged, currently 2 bottom commits are dev commits that I will strip before merging.

controller/instance_manager_controller.go
```diff
@@ -807,6 +807,11 @@ func (imc *InstanceManagerController) createInstanceManagerPod(im *longhorn.Inst
 		return err
 	}
 
+	podSpec.Spec.Containers[0].Env = append(podSpec.Spec.Containers[0].Env,
+		v1.EnvVar{Name: "GRPC_GO_LOG_VERBOSITY_LEVEL", Value: "99"},
```
innobead, line 804: I feel this is quite aggressive. As for debugging, but what the major reason we should turn on the verbosity and use debug level instead of info?

deploy/install/02-components/01-manager.yaml
```diff
@@ -77,6 +83,11 @@ spec:
       - name: longhorn
         hostPath:
           path: /var/lib/longhorn/
+      - name: longhorn-grpc-tls
+        secret:
+          secretName: longhorn-grpc-tls
+          optional: true
```
innobead, line 89: So users need to provide the key-pair and ca if they want TLS enabled between manager and engine. If not provided, it's disabled by default?

controller/instance_manager_controller.go
```diff
@@ -955,17 +981,33 @@ func (imc *InstanceManagerController) createReplicaManagerPodSpec(im *longhorn.I
...
 		{
 			MountPath:        "/host",
 			Name:             "host",
 			MountPropagation: &hostToContainer,
 		},
+		{
+			MountPath:        types.EngineBinaryDirectoryInContainer,
+			Name:             "engine-binaries",
+			MountPropagation: &hostToContainer,
+		},
```
innobead, line 998: Why replica IM pods need to access engine binaries?

engineapi/instance_manager.go
```diff
@@ -56,12 +65,53 @@ func NewInstanceManagerClient(im *longhorn.InstanceManager) (*InstanceManagerCli
...
+		if _, err = pmClient.VersionGet(); err != nil {
+			return nil, fmt.Errorf("failed to check check version of Instance Manager Client with TLS connection Error: %w", err)
+		}
+
+		return pmClient, nil
+	}
+
+	pmClient, err := initTLSClient()
+	if err != nil {
+		// fallback to non tls client, there is no way to differentiate between im versions unless we get the version via the im client
```
innobead, line 93: I would suggest having a log for fallback behavior, it's better to understand the context at runtime.

controller/instance_handler.go
```diff
@@ -337,11 +339,13 @@ func (h *InstanceHandler) ReconcileInstanceState(obj interface{}, spec *longhorn
...
 func (h *InstanceHandler) printInstanceLogs(instanceName string, obj runtime.Object) error {
-	stream, err := h.instanceManagerHandler.LogInstance(obj)
+	ctx, cancel := context.WithCancel(context.TODO())
+	defer cancel()
+	client, stream, err := h.instanceManagerHandler.LogInstance(ctx, obj)
 	if err != nil {
 		return err
 	}
-	defer stream.Close()
+	defer client.Close()
```
innobead, line 348: Should we close the returned stream as well?

innobead (commented):
In general, LGTM. However, we should have a doc for the below items.

- mTLS is not enabled by default, but rather users need to provide the corresponding key-pair and CA. We need a doc for this.
- If mTLS is enabled before active volumes attached, how usres should expect? or should users need to detach all volumes before enabling/disabling mTLS? 

I would suggest to have an enhance proposal later if there are conditions around and related doc PRs as well.

controller/instance_manager_controller.go
joshimoo: This is marked as DEV, I left the commit in for the review but indeed to remove it before merging.
The dev commits make it easier to review/test this out as well to get a better understanding on how the connection setup/teardown reconnection etc works.

deploy/install/02-components/01-manager.yaml
joshimoo: The user needs to create the `longhorn-grpc-tls` secret before the longhorn-manager pod is started.
Since the secret is marked optional it's fine if it's not there which means tls is disabled.

controller/instance_manager_controller.go
joshimoo: They already do access this via the root host mount, I would like to unify the two im pods to only refer to `engine-binaries` instead of doing the indirect access via the root host mount.

This will allow for removal of one of the white list entries and make the pathing between im-e and im-r identical for engine calls.

But it's not a requirement and can be removed for the moment.

controller/instance_handler.go
joshimoo: Stream gets automatically closed when client is closed.
For knowledge sharing REF: https://github.com/longhorn/longhorn-instance-manager/blob/60cd930107ef19273b9c490814b69a59669e306f/vendor/google.golang.org/grpc/stream.go#L127-L143

engineapi/instance_manager.go
joshimoo: Generally agree adding the logrus ref here kinda sucks though.
I didn't want to change this to much before the proxy PR.
Afterwards I intended to add a client factory so we can have im client mock for unit tests.

joshimoo:
Agree on the requirement for documentation, I will write some up.

engineapi/instance_manager.go
```diff
@@ -56,12 +65,53 @@ func NewInstanceManagerClient(im *longhorn.InstanceManager) (*InstanceManagerCli
...
+	endpoint := "tcp://" + imutil.GetURL(im.Status.IP, InstanceManagerDefaultPort)
+
+	initTLSClient := func() (*imclient.ProcessManagerClient, error) {
+		// check for tls cert file presence
+		tlsDir := types.TLSDirectoryInContainer
+		pmClient, err := imclient.NewProcessManagerClientWithTLS(endpoint,
+			filepath.Join(tlsDir, "ca.crt"),
+			filepath.Join(tlsDir, "tls.crt"),
+			filepath.Join(tlsDir, "tls.key"),
+			"longhorn-backend.longhorn-system",
```
shuo-wu, line 78: NIT: Use constant instead

controller/engine_controller.go
```diff
@@ -560,22 +560,24 @@ func (ec *EngineController) GetInstance(obj interface{}) (*longhorn.InstanceProc
 	return c.ProcessGet(e.Name)
 }
 
-func (ec *EngineController) LogInstance(ctx context.Context, obj interface{}) (*imapi.LogStream, error) {
+func (ec *EngineController) LogInstance(ctx context.Context, obj interface{}) (*engineapi.InstanceManagerClient, *imapi.LogStream, error) {
```
shuo-wu, line 563: NIT: I am thinking if the InstanceManagerClient should be init by the caller of the interface rather than by itself. Returning the client looks weird...

shuo-wu (approved):
In general LGTM.
The overall implementations of InstanceManagerController and InstanceHandler do look bad. One more thing needs to be refactored lol.

controller/instance_manager_controller.go
shuo-wu: If you want to unify the binary path, you may need to update this as well: https://github.com/longhorn/longhorn-manager/blob/ee7a757d554ed57b4bd2487e80dba880e551333f/engineapi/instance_manager.go#L138

controller/engine_controller.go
joshimoo: Agree the controller + handler stuff needs to be refactored, I didn't want to do this as part of this issue instead we should do the refactoring after we have done the proxy + monitor implementation since I don't think we will even need the handler abstraction afterwards.

I left a more details of this as part of the commit message.

controller/instance_manager_controller.go
joshimoo: I removed this for now from this PR, since we can do the full change over in the future.
Better to have all changes be contained in that PR.

engineapi/instance_manager.go
joshimoo: Agree, done

joshimoo:
@innobead updated the PR, stripped the dev commits and updated to the new im image + dependency

innobead (approved):

PR #1294, opened 2022-04-12
[master] Make Longhorn UI Listen Binding configurable

smallteeths:
https://github.com/longhorn/longhorn/issues/3829

Signed-off-by: Siye Wang <siye@rancher.com>


innobead (approved):

innobead:
@smallteeths we also need to update longhorn/longhorn repo.

PR #1296, opened 2022-04-13
Revert the resync back to 30s

PhanLe1010:
Revert the resync back to 30s for Longhorn v1.3.0 since there are conner cases that we cannot solve in time.

We can retry increasing the resync in the v1.4.0 release. The effort is tracked at https://github.com/longhorn/longhorn/issues/3629


innobead (approved):
LGTM

PR #1299, opened 2022-04-13
Support Kubernetes cluster-autoscaler

c3y1huang:
https://github.com/longhorn/longhorn/issues/2203

### DIFF
--- a/controller/instance_manager_controller.go
+++ b/controller/instance_manager_controller.go
@@ -73,6 +73,8 @@ type InstanceManagerMonitor struct {
 	done                   bool
 	// used to notify the controller that monitoring has stopped
 	monitorVoluntaryStopCh chan struct{}
+
+	nodeCallback func(obj interface{})
 }
 
 type InstanceManagerUpdater struct {
@@ -149,7 +151,34 @@ func NewInstanceManagerController(
 	}, 0)
 	imc.cacheSyncs = append(imc.cacheSyncs, ds.KubeNodeInformer.HasSynced)
 
+	ds.SettingInformer.AddEventHandlerWithResyncPeriod(
+		cache.FilteringResourceEventHandler{
+			FilterFunc: imc.isResponsibleForSetting,
+			Handler: cache.ResourceEventHandlerFuncs{
+				UpdateFunc: func(old, cur interface{}) { imc.enqueueSettingChange(cur) },
+			},
+		}, 0)
+	imc.cacheSyncs = append(imc.cacheSyncs, ds.SettingInformer.HasSynced)
+
 	return imc
+}
+
+func (imc *InstanceManagerController) isResponsibleForSetting(obj interface{}) bool {
+	setting, ok := obj.(*longhorn.Setting)
+	if !ok {
+		deletedState, ok := obj.(cache.DeletedFinalStateUnknown)
+		if !ok {
+			return false
+		}
+
+		// use the last known state, to enqueue, dependent objects
+		setting, ok = deletedState.Obj.(*longhorn.Setting)
+		if !ok {
+			return false
+		}
+	}
+
+	return types.SettingName(setting.Name) == types.SettingNameKubernetesClusterAutoscalerEnabled
 }
 
 func isInstanceManagerPod(obj interface{}) bool {
@@ -468,19 +497,63 @@ func (imc *InstanceManagerController) syncInstanceManagerPDB(im *longhorn.Instan
 		return err
 	}
 
-	// When current node is unschedulable, it is a signal that the node is being cordoned/drained.
+	// When current node is unschedulable, it is a signal that the node is being
+	// cordoned/drained. The replica IM PDB can be delete when there is least one
+	// IM PDB on another schedulable node to protect detached volume data.
+	//
+	// During Cluster Autoscaler scale down, when a node is marked unschedulable
+	// means CA already decided that this node is not blocked by any pod PDB limit.
+	// Hence there is no need to check when Cluster Autoscaler is enabled.
 	if unschedulable {
 		if imPDB == nil {
 			return nil
 		}
-		canDelete, err := imc.canDeleteInstanceManagerPDB(im)
+
+		isPDBRequired, err := imc.canDeleteInstanceManagerPDB(im)
 		if err != nil {
 			return err
 		}
-		if !canDelete {
+
+		if !isPDBRequired {
 			return nil
 		}
+
+		imc.logger.Debugf("Node %v is marked unschedulable, try to remove %v PDB", imc.controllerID, im.Name)
 		return imc.deleteInstanceManagerPDB(im)
+	}
+
+	// If the setting is enabled, Longhorn needs to retain the least IM PDBs as
+	// possible. Each volume will have at least one replica under the protection
+	// of an IM PDB while no redundant PDB blocking the Cluster Autoscaler from
+	// scale down.
+	// CA considers a node is unremovable when there are strict PDB limits
+	// protecting the pods on the node.
+	//
+	// If the setting is disabled, Longhorn will blindly create IM PDBs for all
+	// engine and replica IMs.
+	clusterAutoscalerEnabled, err := imc.ds.GetSettingAsBool(types.SettingNameKubernetesClusterAutoscalerEnabled)
+	if err != nil {
+		return err
+	}
+
+	if clusterAutoscalerEnabled {
+		isPDBRequired, err := imc.canDeleteInstanceManagerPDB(im)
+		if err != nil {
+			return err
+		}
+
+		if !isPDBRequired {
+			if imPDB == nil {
+				return imc.createInstanceManagerPDB(im)
+			}
+			return nil
+		}
+
+		if imPDB != nil {
+			return imc.deleteInstanceManagerPDB(im)
+		}
+
+		return nil
 	}
 
 	// Make sure that there is a PodDisruptionBudget to protect this instance manager in normal case.
@@ -533,10 +606,12 @@ func (imc *InstanceManagerController) cleanUpPDBForNonExistingIM() error {
 }
 
 func (imc *InstanceManagerController) deleteInstanceManagerPDB(im *longhorn.InstanceManager) error {
-	err := imc.ds.DeletePDB(imc.getPDBName(im))
+	name := imc.getPDBName(im)
+	err := imc.ds.DeletePDB(name)
 	if err != nil && !datastore.ErrorIsNotFound(err) {
 		return err
 	}
+	imc.logger.Infof("Deleted %v PDB", name)
 	return nil
 }
 
@@ -557,7 +632,8 @@ func (imc *InstanceManagerController) canDeleteInstanceManagerPDB(im *longhorn.I
 	}
 
 	// Must wait for all volumes detached from the current node first.
-	// This also means that we must wait until the PDB of engine instance manager on the current node is deleted
+	// This also means that we must wait until the PDB of engine instance manager
+	// on the current node is deleted
 	allVolumeDetached, err := imc.areAllVolumesDetachedFromCurrentNode()
 	if err != nil {
 		return false, err
@@ -583,7 +659,8 @@ func (imc *InstanceManagerController) canDeleteInstanceManagerPDB(im *longhorn.I
 	}
 
 	// For each replica process in the current node,
-	// find out whether there is a healthy replica of the same volume on another node
+	// find out whether there is a PDB protected healthy replica of the same
+	// volume on another schedulable node.
 	for _, replica := range replicasOnCurrentNode {
 		vol, err := imc.ds.GetVolume(replica.Spec.VolumeName)
 		if err != nil {
@@ -595,25 +672,49 @@ func (imc *InstanceManagerController) canDeleteInstanceManagerPDB(im *longhorn.I
 			return false, err
 		}
 
-		hasHealthyReplicaOnAnotherNode := false
+		hasPDBOnAnotherNode := false
 		isUnusedReplicaOnCurrentNode := false
 		for _, r := range replicas {
-			if r.Spec.HealthyAt != "" && r.Spec.FailedAt == "" && r.Spec.NodeID != imc.controllerID {
-				hasHealthyReplicaOnAnotherNode = true
-				break
+			hasOtherHealthyReplicas := r.Spec.HealthyAt != "" && r.Spec.FailedAt == "" && r.Spec.NodeID != imc.controllerID
+			if hasOtherHealthyReplicas {
+				unschedulable, err := imc.ds.IsKubeNodeUnschedulable(r.Spec.NodeID)
+				if err != nil {
+					return false, err
+				}
+				if unschedulable {
+					continue
+				}
+
+				var rIM *longhorn.InstanceManager
+				rIM, err = imc.getRunningReplicaInstancManager(r)
+				if err != nil {
+					return false, err
+				}
+				if rIM == nil {
+					continue
+				}
+
+				pdb, err := imc.ds.GetPDBRO(imc.getPDBName(rIM))
+				if err != nil && !datastore.ErrorIsNotFound(err) {
+					return false, err
+				}
+				if pdb != nil {
+					hasPDBOnAnotherNode = true
+					break
+				}
 			}
 			// If a replica has never been started, there is no data stored in this replica, and
 			// retaining it makes no sense for HA.
 			// Hence Longhorn doesn't need to block the PDB removal for the replica.
 			// This case typically happens on a newly created volume that hasn't been attached to any node.
 			// https://github.com/longhorn/longhorn/issues/2673
-			if r.Spec.HealthyAt == "" && r.Spec.FailedAt == "" && r.Spec.NodeID == imc.controllerID {
-				isUnusedReplicaOnCurrentNode = true
+			isUnusedReplicaOnCurrentNode := r.Spec.HealthyAt == "" && r.Spec.FailedAt == "" && r.Spec.NodeID == imc.controllerID
+			if isUnusedReplicaOnCurrentNode {
 				break
 			}
 		}
 
-		if !hasHealthyReplicaOnAnotherNode && !isUnusedReplicaOnCurrentNode {
+		if !hasPDBOnAnotherNode && !isUnusedReplicaOnCurrentNode {
 			return false, nil
 		}
 	}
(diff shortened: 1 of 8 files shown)


datastore/kubernetes.go
```diff
@@ -430,9 +430,35 @@ func (s *DataStore) IsKubeNodeUnschedulable(nodeName string) (bool, error) {
 	if err != nil {
 		return false, err
 	}
+
+	clusterAutoscalerEnabled, err := s.GetSettingAsBool(types.SettingNameKubernetesClusterAutoscalerEnabled)
```
innobead, line 434: I think we should just allow the drain w/ cordon case triggered by cluster autoscaler to avoid unexpected following replica scheduling, so that said we just keep the current function implementation as is instead of respecting `ToBeDeletedByClusterAutoscaler` which also include the drain w/o cordon case. WDYT?

controller/instance_manager_controller.go
```diff
@@ -473,21 +473,38 @@ func (imc *InstanceManagerController) syncInstanceManagerPDB(im *longhorn.Instan
 		if imPDB == nil {
 			return nil
 		}
-		canDelete, err := imc.canDeleteInstanceManagerPDB(im)
+		imc.logger.Debugf("Node %v is marked unschedulable, try to remove %v PBD", imc.controllerID, im.Name)
```
innobead, line 482: We should separate cases for enabling CA or disabling CA instead of mixing the implementation together. 

```go
if unscheduled {
    if enabled {
        // the new case for CA awareness 
    } else {
        // the same as before
    }
}
```

types/types.go
```diff
@@ -103,6 +103,10 @@ const (
 	KubernetesTopologyRegionLabelKey      = "topology.kubernetes.io/region"
 	KubernetesTopologyZoneLabelKey        = "topology.kubernetes.io/zone"
 
+	KubernetesClusterAutoscalerSafeToEvictKey     = "cluster-autoscaler.kubernetes.io/safe-to-evict"
+	KubernetesClusterAutoscalerSafeToEvictEnabled = "true"
```
innobead, line 107: `true` is very straight, so I don't think we should use another constant for that, especially the const here just implicitly means `true` actually.

controller/setting_controller.go
```diff
@@ -563,6 +567,51 @@ func (sc *SettingController) updatePriorityClass() error {
...
+	if err != nil {
+		return err
+	}
+	newClusterAutoscalerEnabled := setting.Value
+	evictKey := types.KubernetesClusterAutoscalerSafeToEvictKey
+	deploymentList, err := sc.ds.ListDeploymentWithLabels(types.GetBaseLabelsForSystemManagedComponent())
+	if err != nil {
+		return errors.Wrapf(err, "failed to list Longhorn deployments for priority class update")
+	}
+	longhornUI, err := sc.ds.GetDeployment(types.LonghornUIDeploymentName)
```
innobead, line 581: Please use extra new line following where you finish one logic group of statements. This is kind of coding convention as the below example.

```go
    deploymentList, err := sc.ds.ListDeploymentWithLabels(types.GetBaseLabelsForSystemManagedComponent())
	if err != nil {
		return errors.Wrapf(err, "failed to list Longhorn deployments for priority class update")
	}

	longhornUI, err := sc.ds.GetDeployment(types.LonghornUIDeploymentName)
	if err != nil && !apierrors.IsNotFound(err) {
		return errors.Wrapf(err, "failed
```

datastore/kubernetes.go
innobead: After discussing with @c3y1huang , I think there are no obvious ways to let users configure `cordon before drain` in cloud provider cluster autoscaler. 

Also, the specific taint added by CA actually means the node is marked to scale down, so it indirectly means it's unscheduled for Longhorn. Let's keep the change as is first. 

ref: https://github.com/kubernetes/ingress-gce/issues/595

controller/instance_manager_controller.go
c3y1huang: Fixed.

types/types.go
c3y1huang: Removed.

controller/setting_controller.go
c3y1huang: Fixed.

datastore/kubernetes.go
c3y1huang: After talking to @innobead ,

There is an option flag [cordon-node-before-terminating](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/main.go#L185). If this option is enabled, CA will [mark the node spec unschedulable](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/utils/deletetaint/delete.go#L120-L122).
However, regardless of the option flag, CA will taint the node with `ToBeDeletedByClusterAutoscaler`.

There could be a possible corner case when the user manually tolerates all taints and the replica could still rebuild onto the to be deleted node.

We think this can be tackled later if users have further concerns.

controller/instance_manager_controller.go
```diff
@@ -468,26 +468,60 @@ func (imc *InstanceManagerController) syncInstanceManagerPDB(im *longhorn.Instan
...
+			}
+
+			node, err := imc.ds.GetNode(availableNode)
+			if err != nil {
+				return err
+			}
+			imc.logger.Debugf("Creating instance manager PBD on %v", node.Name)
+			imc.enqueueKubernetesNode(node)
+			imc.logger.Debugf("Deleting %v PDB on %v", im.Name, imc.controllerID)
+			return imc.deleteInstanceManagerPDB(im)
```
innobead, line 505: We should not just `deleteInstanceManagerPDB` unless the PDB has been created in the available node.

controller/setting_controller.go
```diff
@@ -563,6 +567,57 @@ func (sc *SettingController) updatePriorityClass() error {
...
+	evictKey := types.KubernetesClusterAutoscalerSafeToEvictKey
+
+	longhornUI, err := sc.ds.GetDeployment(types.LonghornUIDeploymentName)
+	if err != nil && !apierrors.IsNotFound(err) {
+		return errors.Wrapf(err, "failed to get %v deployment", types.LonghornUIDeploymentName)
+	}
+
+	deploymentList, err := sc.ds.ListDeploymentWithLabels(types.GetBaseLabelsForSystemManagedComponent())
+	if err != nil {
+		return errors.Wrapf(err, "failed to list Longhorn deployments for priority class update")
```
innobead, line 586: The log is incorrect.

controller/setting_controller.go
```diff
@@ -563,6 +567,57 @@ func (sc *SettingController) updatePriorityClass() error {
...
+	newClusterAutoscalerEnabled := setting.Value
+
+	evictKey := types.KubernetesClusterAutoscalerSafeToEvictKey
+
+	longhornUI, err := sc.ds.GetDeployment(types.LonghornUIDeploymentName)
+	if err != nil && !apierrors.IsNotFound(err) {
+		return errors.Wrapf(err, "failed to get %v deployment", types.LonghornUIDeploymentName)
+	}
+
+	deploymentList, err := sc.ds.ListDeploymentWithLabels(types.GetBaseLabelsForSystemManagedComponent())
```
innobead, line 584: Why we need to deal with K8s managed pods here? I thought all of them will be covered in CA already even those are created by Longhorn.

innobead (changes requested):

controller/setting_controller.go
innobead: ah, thanks @c3y1huang explanation again, it's used to add the annotation to the managed pods w/ constraints like local storage volumes.

innobead:
Except the comments above, LGTM

controller/setting_controller.go
c3y1huang: Fixed.

controller/instance_manager_controller.go
c3y1huang: Fixed, and added case to test cordon node [here](https://github.com/longhorn/longhorn-tests/pull/936/commits).

controller/instance_manager_controller.go
```diff
@@ -468,26 +468,61 @@ func (imc *InstanceManagerController) syncInstanceManagerPDB(im *longhorn.Instan
 		return err
 	}
 
+	clusterAutoscalerEnabled, err := imc.ds.GetSettingAsBool(types.SettingNameKubernetesClusterAutoscalerEnabled)
```
shuo-wu, line 477: Need an event handler for the setting CR

controller/instance_manager_controller.go
```diff
@@ -468,26 +468,61 @@ func (imc *InstanceManagerController) syncInstanceManagerPDB(im *longhorn.Instan
...
+		}
+		if canDelete {
+			if imPDB == nil {
+				return nil
+			}
+			return imc.deleteInstanceManagerPDB(im)
+		}
+		if imPDB == nil {
+			return imc.createInstanceManagerPDB(im)
+		}
```
shuo-wu, line 527: Redundant.

controller/instance_manager_controller.go
```diff
@@ -533,74 +568,106 @@ func (imc *InstanceManagerController) cleanUpPDBForNonExistingIM() error {
...
+	allInstanceRemoved, err := imc.areAllInstanceRemovedFromCurrentNode(im)
 	if err != nil {
-		return false, err
+		return false, "", err
 	}
-	if !allVolumeDetached {
-		return false, nil
+	if !allInstanceRemoved {
+		return false, "", nil
 	}
```
shuo-wu, line 645: Wait... Can we replace the volume detachment check function with the instance cleanup wait function?
After this modification, can a node with running replicas be drained even if the replicas are not the only ones of the volumes?

controller/instance_manager_controller.go
```diff
@@ -468,26 +468,61 @@ func (imc *InstanceManagerController) syncInstanceManagerPDB(im *longhorn.Instan
...
+		imc.logger.Debugf("Node %v is marked unschedulable, try to remove %v PDB", imc.controllerID, im.Name)
+		canDelete, availableNode, err := imc.canDeleteInstanceManagerPDB(im)
 		if err != nil {
 			return err
 		}
 		if !canDelete {
+			if clusterAutoscalerEnabled && im.Spec.Type == longhorn.InstanceManagerTypeReplica {
+				if availableNode == "" {
+					return errors.Errorf("cannot find healthy replica on another node")
+				}
```
shuo-wu, line 497: Should we error out here?

controller/instance_manager_controller.go
```diff
@@ -533,74 +568,106 @@ func (imc *InstanceManagerController) cleanUpPDBForNonExistingIM() error {
...
+				pdb, err := imc.ds.GetPDBRO(imc.getPDBName(rIM))
+				if err != nil && !datastore.ErrorIsNotFound(err) {
+					return false, "", err
+				}
+				if pdb != nil {
+					hasPDBOnAnotherNode = true
+					break
+				}
 			}
```
shuo-wu, line 713: 1. Can we refactor this block like the following?
```
containOtherHealthyReplicas := r.Spec.HealthyAt != "" && r.Spec.FailedAt == "" && r.Spec.NodeID != imc.controllerID
isOtherHealthyReplicaOnScheduableNode := ....
isOtherHealthyReplicaHasRuningIM := ....
isOtherHealthyReplicaHasIMPDB := 
```
2. Can the logic handle this corner case: A volume has 2 replicas on node2 and node3, and the volume is attached to node1. Then the node2 & node3 are drained (almost) simultaneously.
3. What if the controller does not enqueue `availableNode`? (This output field is a little bit confusing)

controller/setting_controller.go
```diff
@@ -563,6 +567,57 @@ func (sc *SettingController) updatePriorityClass() error {
 	return nil
 }
 
+func (sc *SettingController) updateKubernetesClusterAutoscalerEnabled() error {
```
shuo-wu, line 570: Similar to PQ or Toleration settings, should we put the IM pod annotation update in this function?

controller/instance_manager_controller.go
c3y1huang: Thanks, fixed.

controller/instance_manager_controller.go
c3y1huang: Here we are expecting this node would have the only replica IM PDB. When a node is marked unschedulable we want another available node with the volume replica to re-create the replica IM PDB.

controller/instance_manager_controller.go
```diff
@@ -149,6 +149,11 @@ func NewInstanceManagerController(
 	}, 0)
 	imc.cacheSyncs = append(imc.cacheSyncs, ds.KubeNodeInformer.HasSynced)
 
+	ds.SettingInformer.AddEventHandlerWithResyncPeriod(cache.ResourceEventHandlerFuncs{
```
shuo-wu, line 153: NIT: you can use `AddEventHandlerWithResyncPeriod` instead. See the node controller as an example.

controller/instance_manager_controller.go
shuo-wu: Yes, we should block/avoid deleting the PDB for this IM that contains the only replica of a volume. But it does not mean that we need to fail the **whole IM controller reconcile loop**.

controller/instance_manager_controller.go
c3y1huang: Removed.

controller/instance_manager_controller.go
c3y1huang: I see. Fixed.

controller/instance_manager_controller.go
c3y1huang: > 1. Can we refactor this block like the following?
> 
> ```
> containOtherHealthyReplicas := r.Spec.HealthyAt != "" && r.Spec.FailedAt == "" && r.Spec.NodeID != imc.controllerID
> isOtherHealthyReplicaOnScheduableNode := ....
> isOtherHealthyReplicaHasRuningIM := ....
> isOtherHealthyReplicaHasIMPDB := 
> ```

I've made some refactoring. PTAL.

> 
> 2. Can the logic handle this corner case: A volume has 2 replicas on node2 and node3, and the volume is attached to node1. Then the node2 & node3 are drained (almost) simultaneously.

- We do not delete PDB when a node has attached volume.
- We are not expecting CA to remove multiple nodes at once. Ref [How does scale-down work?](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#how-does-scale-down-work)
```
Cluster Autoscaler terminates one non-empty node at a time to reduce the risk of creating new unschedulable pods.
```
- CA would have made a decision to not scale down at pre-check if it's the last healthy replica available. So this logic applies to manual cordon/drain.

This logic does not handle the case of manual cordon/drain of multiple nodes simultaneously. Any suggestion to handle this? 

> 3. What if the controller does not enqueue `availableNode`? (This output field is a little bit confusing)
We will block PDB deletion.

controller/instance_manager_controller.go
c3y1huang: Updated.

controller/setting_controller.go
c3y1huang: Done.

controller/instance_manager_controller.go
c3y1huang: @shuo-wu @innobead , I've removed logic to enqueue another schedulable node. Don't think we need to re-create the PDB.

controller/instance_manager_controller.go
```diff
@@ -473,16 +505,38 @@ func (imc *InstanceManagerController) syncInstanceManagerPDB(im *longhorn.Instan
...
 		if !canDelete {
 			return nil
 		}
 		return imc.deleteInstanceManagerPDB(im)
 	}
 
+	// CA considers a node is unremovable when the node has pod that does not allow
+	// for disruptions. When this setting is enabled Longhorn will actively delete
+	// IM PDB when volume is detached, leaving single replica IM PDB to protect volume
+	// data.
```
shuo-wu, line 566: If the setting is enabled, Longhorn needs to retain as least IM PDBs as possible so that there is at least one replica for each volume while no redundant PDB blocking the CA draining.
If the setting is disabled, Longhorn will blindly create IM PDBs. When a node is being drained/cordoned while the draining won't crash volumes, the PDBs on the node will be cleaned up

c3y1huang: Reword.

innobead (commented):
In genreal LTGM, just one question left.

controller/instance_manager_controller.go
```diff
@@ -1186,6 +1329,28 @@ func (m *InstanceManagerMonitor) pollAndUpdateInstanceMap() (needStop bool) {
...
+		return false
+	}
+
+	// During volume attaching/detaching, it is likely both the engine and replica
+	// IMs enqueue at the same time. If the replica IM queued before the engine IM,
+	// then the instance in the engine manager possibly still not updated.
+	// When ClusterAutoscaler is enabled, Longhorn cannot remove the redundant
+	// replica IM PDB in this case. So enqueue the node IMs again to sync replica
+	// IM PDB.
+	if clusterAutoscalerEnabled && im.Spec.Type == longhorn.InstanceManagerTypeEngine {
```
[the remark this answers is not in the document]
c3y1huang: When engine and replica IM get enqueued simultaneously the replica IM could possibly return [here](https://github.com/longhorn/longhorn-manager/blob/master/controller/instance_manager_controller.go#L547-L552) while the engine IM is still getting updated.

controller/instance_manager_controller.go
```diff
@@ -761,6 +889,19 @@ func (imc *InstanceManagerController) enqueueKubernetesNode(obj interface{}) {
...
+func (imc *InstanceManagerController) enqueueSettingChange(obj interface{}) {
+	ims, err := imc.ds.ListInstanceManagers()
+	if err != nil {
+		utilruntime.HandleError(fmt.Errorf("failed to list instance manager on current node %v: %v",
+			imc.controllerID, err))
+		return
+	}
+
+	for _, im := range ims {
+		imc.enqueueInstanceManager(im)
```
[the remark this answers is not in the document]
c3y1huang: I might be wrong, but I think we should enqueue the IM. All IMs will catch the change when the setting gets updated. If we enqueue the node then all IMs will get enqueued + all the IMs on the IM node get enqueued again.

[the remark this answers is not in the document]
innobead: You are not wrong, but each manager has their IM controller and all of them listen settings change now, so each of them should just enqueue their own instance managers, otherwise, there will be duplicated events to the queue?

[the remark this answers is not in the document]
c3y1huang: I see, you are right.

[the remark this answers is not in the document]
c3y1huang: Fixed.

controller/setting_controller.go
```diff
@@ -563,6 +567,59 @@ func (sc *SettingController) updatePriorityClass() error {
...
+	// IM pods annotation will be handled in the instance manager controller
+
+	setting, err := sc.ds.GetSetting(types.SettingNameKubernetesClusterAutoscalerEnabled)
+	if err != nil {
+		return err
+	}
+	newClusterAutoscalerEnabled := setting.Value
```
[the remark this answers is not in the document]
c3y1huang: Fixed.

controller/instance_manager_controller.go
c3y1huang: Reverted.

innobead (dismissed):
LGTM

shuo-wu (approved):
LGTM

controller/instance_manager_controller.go
```diff
@@ -468,21 +539,64 @@ func (imc *InstanceManagerController) syncInstanceManagerPDB(im *longhorn.Instan
...
+	//
+	// During Cluster Autoscaler scale down, when a node is marked unschedulable
+	// means CA already decided that this node is not blocked by any pod PDB limit.
+	// Hence there is no need to check when Cluster Autoscaler is enabled.
 	if unschedulable {
 		if imPDB == nil {
 			return nil
 		}
+
+		imc.logger.Debugf("Node %v is marked unschedulable, try to remove %v PDB", imc.controllerID, im.Name)
```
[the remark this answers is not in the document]
c3y1huang: Addressed.

controller/instance_manager_controller.go
```diff
@@ -468,21 +539,64 @@ func (imc *InstanceManagerController) syncInstanceManagerPDB(im *longhorn.Instan
...
+	//
+	// If the setting is disabled, Longhorn will blindly create IM PDBs for all
+	// engine and replica IMs.
+	clusterAutoscalerEnabled, err := imc.ds.GetSettingAsBool(types.SettingNameKubernetesClusterAutoscalerEnabled)
+	if err != nil {
+		return err
+	}
+
+	if clusterAutoscalerEnabled {
+		canDelete, err := imc.canDeleteInstanceManagerPDB(im)
```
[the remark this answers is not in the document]
c3y1huang: Addressed.

controller/instance_manager_controller.go
```diff
@@ -468,21 +539,65 @@ func (imc *InstanceManagerController) syncInstanceManagerPDB(im *longhorn.Instan
...
+	//
+	// During Cluster Autoscaler scale down, when a node is marked unschedulable
+	// means CA already decided that this node is not blocked by any pod PDB limit.
+	// Hence there is no need to check when Cluster Autoscaler is enabled.
 	if unschedulable {
 		if imPDB == nil {
 			return nil
 		}
+
 		canDelete, err := imc.canDeleteInstanceManagerPDB(im)
```
innobead, line 554: `isPDBRequired` here as well.

c3y1huang: Fixed.

innobead:
@c3y1huang well done  👍

PR #1301, opened 2022-04-18
Fix volumeMounts missing in admission webhook

derekbit:
[Longhorn 3860](https://github.com/longhorn/longhorn/issues/3860)

### DIFF
--- a/manager/volume.go
+++ b/manager/volume.go
@@ -161,6 +161,12 @@ func (m *VolumeManager) Create(name string, spec *longhorn.VolumeSpec, recurring
 		labels[key] = types.LonghornLabelValueEnabled
 	}
 
+	if spec.DataSource != "" {
+		if err := m.verifyDataSourceForVolumeCreation(spec.DataSource, spec.Size); err != nil {
+			return nil, err
+		}
+	}
+
 	v = &longhorn.Volume{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:   name,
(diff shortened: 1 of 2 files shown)


innobead (dismissed):
LGTM.

innobead:
Just one minor note for now, because the below part will involve some communication efforts between the validating webhook (we only have 2 replicas) and volume engines, it will possibly have some cost issue if many volume cloning operations happen simultaneously, even though this is actually a corner case.

I still expect we can make the validation focus on things w/ low cost, but rather heavy cost should be moved to the longhorn manager instead. 

```go
func (v *volumeValidator) verifyDataSourceForVolumeCreation(currentNodeID string, dataSource longhorn.VolumeDataSource, requestSize int64) (err error) {
...

		if snapName := types.GetSnapshotName(dataSource); snapName != "" {
			if _, err := v.getSnapshot(currentNodeID, snapName, srcVolName); err != nil {
				return err
			}
		}
...
	return nil
}
```

innobead:
@derekbit will adjust this change a bit based on the above comment. Don't merge until the update.

manager/volume.go
```diff
@@ -954,3 +960,31 @@ func (m *VolumeManager) UpdateAccessMode(name string, accessMode longhorn.Access
...
 	return v, nil
 }
+
+func (m *VolumeManager) verifyDataSourceForVolumeCreation(dataSource longhorn.VolumeDataSource, requestSize int64) (err error) {
+	defer func() {
+		err = errors.Wrapf(err, "failed to verify data source")
+	}()
+
+	if !types.IsValidVolumeDataSource(dataSource) {
+		return fmt.Errorf("in valid value for data source: %v", dataSource)
```
innobead, line 970: typo, `invalid`.

innobead (changes requested):

innobead (approved):

PR #1304, opened 2022-04-20
Support Longhorn snapshot CRD

PhanLe1010:
longhorn/longhorn#3144


PhanLe1010:
I am wondering if we can consider not clearing the snapshot info here: https://github.com/longhorn/longhorn-manager/commit/ec1c382dc84df3c0c9b7101a04e3f40fdbc5f7b7 since it will make all snapshot CRs as lost temporary. How do you think @shuo-wu ?

webhook/resources/snapshot/validator.go
```diff
@@ -0,0 +1,91 @@
...
+	snapshot, ok := newObj.(*longhorn.Snapshot)
+	if !ok {
+		return werror.NewInvalidError(fmt.Sprintf("%v is not a *longhorn.Snaphot", newObj), "")
+	}
+
+	if _, ok := snapshot.Labels[types.LonghornLabelVolume]; !ok {
+		return werror.NewInvalidError(fmt.Sprintf("snapshot labels %v doesnot contain %v", snapshot.Labels, types.LonghornLabelVolume), "metadata.labels")
+	}
+
+	if !contains(snapshot.Finalizers, longhornFinalizerKey) {
```
derekbit, line 48: Mutator already creates the finalizer, so this is redundant.

webhook/resources/snapshot/validator.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+}
+
+func (o *snapshotValidator) Create(request *admission.Request, newObj runtime.Object) error {
+	snapshot, ok := newObj.(*longhorn.Snapshot)
+	if !ok {
+		return werror.NewInvalidError(fmt.Sprintf("%v is not a *longhorn.Snaphot", newObj), "")
+	}
+
+	if _, ok := snapshot.Labels[types.LonghornLabelVolume]; !ok {
```
derekbit, line 44: Redundant, already handeled in mutator.

webhook/resources/snapshot/validator.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	if _, ok := oldSnapshot.Labels[types.LonghornLabelVolume]; ok && newSnapshot.Labels[types.LonghornLabelVolume] != oldSnapshot.Labels[types.LonghornLabelVolume] {
+		return werror.NewInvalidError(fmt.Sprintf("label %v is immutable", types.LonghornLabelVolume), "metadata.labels")
+	}
+
+	return nil
+}
+
+func contains(strs []string, target string) bool {
```
derekbit, line 84: nit: reuse `util.Contains`

controller/snapshot_controller.go
```diff
@@ -0,0 +1,543 @@
...
+}
+
+func (sc *SnapshotController) getTheOnlyEngineCRforSnapshot(snapshot *longhorn.Snapshot) (*longhorn.Engine, error) {
+	engines, err := sc.ds.ListVolumeEngines(snapshot.Spec.Volume)
+	if err != nil {
+		return nil, errors.Wrap(err, "getTheOnlyEngineCRforSnapshot")
+	}
+	if len(engines) != 1 {
+		return nil, fmt.Errorf("getTheOnlyEngineCRforSnapshot: found more than 1 engines for volume %v", snapshot.Spec.Volume)
+	}
```
derekbit, line 446: Is `len(engines) == 0` possible ?

controller/snapshot_controller.go
```diff
@@ -0,0 +1,543 @@
...
+		}
+		if !isPurging {
+			sc.logger.Infof("start SnapshotPurge to delete snapshot %v", snapshot.Name)
+			if err := client.SnapshotPurge(); err != nil {
+				return err
+			}
+		}
+	}
+
+	return nil
```
derekbit, line 511: nit:
```suggestion
	if snapshotInfo == nil {
		return nil
	}
	
	...
	 logic for snapshotInfo != nil
	...

	return nil
```

deploy/install/01-prerequisite/03-crd.yaml
```diff
@@ -2550,6 +2550,127 @@ status:
 ---
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.7.0
+  creationTimestamp: null
+  labels:
+    longhorn-manager: ""
+  name: snapshots.longhorn.io
```
derekbit, line 2559: Also need to add `snapshots` `uninstall.yaml`

controller/snapshot_controller.go
```diff
@@ -0,0 +1,543 @@
...
+
+	//if sc.queue.NumRequeues(key) < maxRetries {
+	//	sc.logger.WithError(err).Warnf("Error syncing Longhorn snapshot %v", key)
+	//	sc.queue.AddRateLimited(key)
+	//	return
+	//}
+
+	//utilruntime.HandleError(err)
+	//sc.logger.WithError(err).Warnf("Dropping Longhorn snapshot %v out of the queue", key)
+	//sc.queue.Forget(key)
```
derekbit, line 234: Do we need the retires?

PhanLe1010: No. I think we don't need maxRetries. If there is any error, we add it back to the queue with rate limited to recheck it later.

controller/snapshot_controller.go
PhanLe1010: Could be the moment when the volume is being deleted. We delete the engine CR first. Anyways, it should be a temporary state

controller/snapshot_controller.go
PhanLe1010: Thank you. Fixed

webhook/resources/snapshot/validator.go
PhanLe1010: Thank you. Fixed

webhook/resources/snapshot/validator.go
PhanLe1010: Thank you. Fixed

webhook/resources/snapshot/validator.go
PhanLe1010: Thank you. Fixed

deploy/install/01-prerequisite/03-crd.yaml
PhanLe1010: Thank you. Fixed

controller/engine_controller.go
```diff
@@ -910,6 +924,52 @@ func (m *EngineMonitor) refresh(engine *longhorn.Engine) error {
 	return nil
 }
 
+func (m *EngineMonitor) syncSnapshotCRs(engine *longhorn.Engine) error {
+	snapshotCRs, err := m.ds.ListVolumeSnapshotsRO(engine.Spec.VolumeName)
```
innobead, line 928: Suggest changing to `ListVolumeSnapshotsByVolumeRO`

controller/snapshot_controller.go
```diff
@@ -0,0 +1,545 @@
...
+	}
+
+	return
+}
+
+func (sc *SnapshotController) Run(workers int, stopCh <-chan struct{}) {
+	defer utilruntime.HandleCrash()
+	defer sc.queue.ShutDown()
+
+	sc.logger.Info("Start Longhorn Snapshot Controller")
```
innobead, line 185: Change to "Starting ...."

innobead (changes requested):
Will continue the rest later.

controller/engine_controller.go
```diff
@@ -868,10 +870,22 @@ func (m *EngineMonitor) refresh(engine *longhorn.Engine) error {
...
-			if _, updateErr := m.ds.UpdateEngineStatus(engine); updateErr != nil {
+			e, updateErr := m.ds.UpdateEngineStatus(engine)
+			if updateErr != nil {
 				err = errors.Wrapf(err, "engine monitor: Failed to update the status for engine %v: %v", engine.Name, updateErr)
+				return
+			}
+			engine = e
+		}
+		if needToSyncSnapshotCRs {
+			if syncErr := m.syncSnapshotCRs(engine); syncErr != nil {
```
derekbit, line 886: IIUC, there are two ways to create the snapshot CRs. One is create the snapshot CRs by kubectl. Another one is create the snapshot by API, and then the snapshot CR is created in this logic. Right?

If yes, might the logic be removed if the snapshot CRD is mature in the next version?

webhook/resources/snapshot/mutator.go
```diff
@@ -0,0 +1,93 @@
...
+
+	if err := util.AddFinalizer(longhornFinalizerKey, snapshot); err != nil {
+		return nil, werror.NewInvalidError(err.Error(), "")
+	}
+	bytes, err = json.Marshal(snapshot.Finalizers)
+	if err != nil {
+		err = errors.Wrapf(err, "failed to get JSON encoding for snapshot %v finalizers", snapshot.Name)
+		return nil, werror.NewInvalidError(err.Error(), "")
+	}
+	patchOps = append(patchOps, fmt.Sprintf(`{"op": "replace", "path": "/metadata/finalizers", "value": %v}`, string(bytes)))
```
derekbit, line 80: Can replace the logic with the two new help functions https://github.com/longhorn/longhorn-manager/blob/master/webhook/common/common.go.

shuo-wu (changes requested):

controller/engine_controller.go
derekbit: I see. Thank you for the clarification.

app/recurring_job.go
```diff
@@ -171,7 +171,7 @@ func recurringJob(c *cli.Context) error {
 			})
 			log.Info("Creating job")
 
-			snapshotName := types.GetCronJobNameForRecurringJob(jobName) + "-" + util.RandomID()
+			snapshotName := sliceStringSafely(types.GetCronJobNameForRecurringJob(jobName), 0, 8) + "-" + util.UUID()
```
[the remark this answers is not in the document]
PhanLe1010: > Why we need this specific modification? or just have a createSnapshotName util function to use here.

We want to have a prefix to easily identify that the snapshot belongs to the recurring job. We also want the snapshot name to have enough entropy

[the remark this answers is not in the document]
PhanLe1010: > Will this affect the existing recurring jobs?

Yes, it will but I don't see a problem. Are you foreseeing a problem?

controller/volume_controller.go
```diff
@@ -311,6 +315,13 @@ func (vc *VolumeController) syncVolume(key string) (err error) {
 			}
 		}
 
+		for _, snap := range snapshots {
+			if snap.DeletionTimestamp == nil {
+				if err := vc.ds.DeleteSnapshot(snap.Name); err != nil {
+					return err
+				}
+			}
+		}
```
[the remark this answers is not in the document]
PhanLe1010: I was thinking about this case if we don't delete snapshot here:
1. User deletes a volume -> volume CR is deleted
2. Snapshots haven't been deleted yet
3. User create a volume with the same name
4. Snapshots might be referring to the new volume

But now rethinking about it, the case is not possible because the owerreference of the snapshot referring to the UUID of the volume. 

I will modify the logic to use `DeletePropagationForeground` thanks

controller/engine_controller.go
```diff
@@ -910,6 +924,52 @@ func (m *EngineMonitor) refresh(engine *longhorn.Engine) error {
...
+	if err != nil {
+		return err
+	}
+
+	for snapName := range engine.Status.Snapshots {
+		// Don't create snapshot CR for the volume-head snapshot
+		if snapName == "volume-head" {
+			continue
+		}
+		if _, ok := snapshotCRs[snapName]; !ok {
```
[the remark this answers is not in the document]
PhanLe1010: Thank you. I decided to use the same name for the snapshot CR as the snapshotInfo inside engine. Please take a look again

controller/engine_controller.go
```diff
@@ -910,6 +924,52 @@ func (m *EngineMonitor) refresh(engine *longhorn.Engine) error {
...
+				ObjectMeta: metav1.ObjectMeta{
+					Name: types.SnapshotInfoNameToSnapshotCRName(snapName, volume.Name),
+				},
+				Spec: longhorn.SnapshotSpec{
+					Volume:         volume.Name,
+					CreateSnapshot: false,
+				},
+			}
+			m.logger.Infof("Creating snapshot CR for the snapshot %v", snapName)
+			if _, err := m.ds.CreateSnapshot(snapCR, volume.Name); err != nil {
```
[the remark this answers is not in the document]
PhanLe1010: Thank you. Fixed

controller/engine_controller.go
```diff
@@ -910,6 +924,52 @@ func (m *EngineMonitor) refresh(engine *longhorn.Engine) error {
...
+				utilruntime.HandleError(fmt.Errorf("syncSnapshotCRs: failed to create snapshot CR %v: %v", snapCR.Name, err))
+			}
+		}
+	}
+
+	for snapName, snapCR := range snapshotCRs {
+		requestCreateNewSnapshot := snapCR.Spec.CreateSnapshot
+		alreadyCreatedBefore := snapCR.Status.CreationTime != nil
+		if _, ok := engine.Status.Snapshots[snapName]; !ok && (!requestCreateNewSnapshot || alreadyCreatedBefore) && snapCR.DeletionTimestamp == nil {
```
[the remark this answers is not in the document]
PhanLe1010: Thank you. Fixed

controller/uninstall_controller.go
```diff
@@ -372,6 +378,13 @@ func (c *UninstallController) deleteCRDs() (bool, error) {
 		return true, c.deleteVolumes(volumes)
 	}
 
+	if snapshots, err := c.ds.ListSnapshots(labels.Everything()); err != nil {
+		return true, err
+	} else if len(snapshots) > 0 {
+		c.logger.Infof("Found %d snapshots remaining", len(snapshots))
+		return true, c.deleteSnapshots(snapshots)
+	}
```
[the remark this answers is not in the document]
PhanLe1010: We deleted all volume CRs before deleting snapshot CRs in the above steps.

Since at this step the volume is already gone, we can delete any snapshot CRs in the system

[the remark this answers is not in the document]
shuo-wu: Got it. It's better to put this explanation as a comment. BTW, if there are still snapshot CRs here (after the volume cleanup), it implicitly means that these CRs are orphans, right?

controller/volume_controller.go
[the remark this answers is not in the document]
PhanLe1010: Sorry. User might use kubectl directly and Kubernetes uses background cascading deletion by default. 

Should we keep the logic here

controller/volume_controller.go
```diff
@@ -344,8 +355,12 @@ func (vc *VolumeController) syncVolume(key string) (err error) {
...
 		}
 
-		// now replicas and engines have been marked for deletion
-
+		// now snapshots, replicas, and engines have been marked for deletion
+		if snapshots, err := vc.ds.ListVolumeSnapshotsRO(volume.Name); err != nil {
+			return err
+		} else if len(snapshots) > 0 {
+			return nil
+		}
```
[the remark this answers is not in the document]
PhanLe1010: It can be delete because the volume has deletion timestamp, snapshot controller will remove finalizer

datastore/longhorn.go
```diff
@@ -3176,6 +3185,119 @@ func (s *DataStore) RemoveFinalizerForBackup(backup *longhorn.Backup) error {
...
 }
 
+// CreateSnapshot creates a Longhorn snapshot CR and verifies creation
+func (s *DataStore) CreateSnapshot(snapshot *longhorn.Snapshot, volumeName string) (*longhorn.Snapshot, error) {
+	if err := tagVolumeLabel(volumeName, snapshot); err != nil {
+		return nil, err
+	}
+	if err := util.AddFinalizer(longhornFinalizerKey, snapshot); err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
PhanLe1010: If we don't create it here, we will make 2  requests to API server:
1. One here
2. One more in the weekhook

I am trying to reduce number of API requests. How do you think?

k8s/pkg/apis/longhorn/v1beta1/engine.go
```diff
@@ -54,7 +54,7 @@ type SnapshotCloneStatus struct {
 	SnapshotName       string `json:"snapshotName"`
 }
 
-type Snapshot struct {
+type SnapshotT struct {
```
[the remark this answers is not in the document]
PhanLe1010: Thank you. Fixed

types/types.go
```diff
@@ -197,6 +197,14 @@ func GetCronJobNameForRecurringJob(name string) string {
 	return name + recurringSuffix
 }
 
+func SnapshotInfoNameToSnapshotCRName(snapshotInfoName, volumeName string) string {
```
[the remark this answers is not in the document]
PhanLe1010: Removed the function. Thank you

controller/uninstall_controller.go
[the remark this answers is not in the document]
PhanLe1010: Yeah, those CR are orphans for a very short period of time. It will automatically be cleaned up by Kubernetes because the owner (the volume) is gone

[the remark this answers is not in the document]
PhanLe1010: Update the comment. Thank you

controller/snapshot_controller.go
```diff
@@ -0,0 +1,584 @@
...
+		if err := syncSnapshotWithSnapshotInfo(snapshot, snapshotInfo, engine.Spec.VolumeSize); err != nil {
+			return err
+		}
+		if reflect.DeepEqual(existingSnapshot.Status, snapshot.Status) {
+			return nil
+		}
+		if _, err = sc.ds.UpdateSnapshotStatus(snapshot); err != nil {
+			return err
+		}
+		sc.generatingEventsForSnapshot(existingSnapshot, snapshot)
```
innobead, line 366: This part is actually the same as the below defer function. Could we move the defer function up, then it can be reused for both cases (deletion or non-deletion)?

controller/snapshot_controller.go
```diff
@@ -0,0 +1,584 @@
...
+	var engine *longhorn.Engine
+	for _, e := range engines {
+		engine = e
+		break
+	}
+	return engine, nil
+}
+
+func (sc *SnapshotController) handleSnapshotCreate(snapshot *longhorn.Snapshot, engine *longhorn.Engine) error {
+	client, err := GetClientForEngine(engine, sc.engineClientCollection, engine.Status.CurrentImage)
```
innobead, line 489: This is probably needed to change to use proxy client instead after IM proxy PR merged. The same for others related to the communication with engine process. 

cc Chinya and Shuo

controller/snapshot_controller.go
```diff
@@ -0,0 +1,584 @@
...
+			isPurging = true
+			break
+		}
+	}
+	if !isPurging {
+		sc.logger.Infof("start SnapshotPurge to delete snapshot %v", snapshot.Name)
+		if err := client.SnapshotPurge(); err != nil {
+			return err
+		}
+	}
```
innobead, line 544: I don't think you need to check at caller side, because I would expect the server will check if purging or not. 

https://github.com/longhorn/longhorn-engine/blob/c9ac9c6a3ad2ebd4389a1abf171490c2d8d4aee0/pkg/sync/rpc/server.go#L1185

innobead (changes requested):

innobead:
In general, LGTM. See if @shuo-wu has any concerns.

webhook/resources/snapshot/mutator.go
PhanLe1010: Thank you. Fixed

controller/snapshot_controller.go
```diff
@@ -0,0 +1,545 @@
...
+		// Delete the snapshot from engine process
+		if err := sc.handleSnapshotDeletion(snapshot, engine); err != nil {
+			return err
+		}
+
+		// Wait for the snapshot to be removed from engine.Status.Snapshots
+		engine, err = sc.ds.GetEngineRO(engine.Name)
+		if err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
PhanLe1010: After handling `handleSnapshotDeletion` the engine might have updated snapshot list, we get it again to see if the deleting snapshot has gone

controller/snapshot_controller.go
```diff
@@ -0,0 +1,545 @@
...
+	}
+
+	// newly created snapshotCR by user
+	requestCreateNewSnapshot := snapshot.Spec.CreateSnapshot
+	alreadyCreatedBefore := snapshot.Status.CreationTime != nil
+	if requestCreateNewSnapshot && !alreadyCreatedBefore {
+		if engine.Status.CurrentState != longhorn.InstanceStateRunning {
+			snapshot.Status.Error = fmt.Sprintf("cannot take snapshot because the volume engine %v is not running", engine.Name)
+			return nil
+		}
```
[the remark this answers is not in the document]
PhanLe1010: We only need the engine to be running in 2 cases:
1. snapshot CR  has deletion timestamp and the volume CR is still exist
2. snapshot CR is requesting a new snapshot

I am not sure how can we check the engine one time. Could you give some hints?

controller/snapshot_controller.go
```diff
@@ -0,0 +1,545 @@
...
+			if apierrors.IsNotFound(err) {
+				log.Infof("Removing finalizer for snapshot %v", snapshot.Name)
+				return sc.ds.RemoveFinalizerForSnapshot(snapshot)
+			}
+			return err
+		}
+		if !volume.DeletionTimestamp.IsZero() {
+			log.Infof("Removing finalizer for snapshot %v", snapshot.Name)
+			return sc.ds.RemoveFinalizerForSnapshot(snapshot)
+		}
```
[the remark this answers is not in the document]
PhanLe1010: Thank you. Fixed

controller/snapshot_controller.go
```diff
@@ -0,0 +1,545 @@
...
+			return err
+		}
+		if reflect.DeepEqual(existingSnapshot.Status, snapshot.Status) {
+			return nil
+		}
+		if _, err = sc.ds.UpdateSnapshotStatus(snapshot); err != nil {
+			return err
+		}
+		sc.generatingEventsForSnapshot(existingSnapshot, snapshot)
+		return nil
```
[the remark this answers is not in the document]
PhanLe1010: > I guess it's for the snapshots that cannot be purged immediately.

You are correct

I would like to seperate the code path for snapshot with deletion timestamp out of the code path for snapshot creation and updating. Yeah, the code looks longer but it simply the logic in the creation/updating path because we don't have to worry about snapshot is being deleted. 

How do you think?

controller/snapshot_controller.go
PhanLe1010: Great idea. Fixed. Thanks

controller/snapshot_controller.go
PhanLe1010: Reaching out to the engine is an expensive operation (the GRPC calls) while checking engine CR is cheap. We don't want to spam the engine process when there are many snapshot CRs being deleted happening near the same time.

controller/snapshot_controller.go
PhanLe1010: Fixed. Thank you

api/model.go
```diff
@@ -1191,11 +1191,11 @@ func toSnapshotResource(s *longhorn.Snapshot) *Snapshot {
 			Id:   s.Name,
 			Type: "snapshot",
 		},
-		Snapshot: *s,
+		SnapshotInfo: *s,
```
innobead, line 1194: This will impact the output of API, so probably will impact UI? Have you seen any problems from UI?

controller/snapshot_controller.go
innobead: Good consideration, let's keep it. Thanks.

metrics_collector/client_go_adaper/metrics.go
```diff
@@ -0,0 +1,98 @@
...
+// this file contains setup logic to initialize the myriad of places
+// that client-go registers metrics.  We copy the names and formats
+// from Kubernetes so that we match the core controllers.
+const (
+	LonghornName = "longhorn"
+)
+
+var (
+	// client metrics
+	clientGoRestClientSubsystem = "rest_client"
```
innobead, line 22: Can we name it specifically like `k8s_rest_client`?

PhanLe1010: I am not so sure. We use this one for both kubeClient and longhornClient https://github.com/longhorn/longhorn-manager/blob/73f1914b988daa121599f7ff2c4c9bbf59a40967/controller/controller_manager.go#L49-L54

innobead: ah, then all good :slightly_smiling_face:

api/model.go
innobead: As confirmed by @PhanLe1010 , it will not impact the output.

note: the response of API actually depends on the schema-registered snapshot response structure as below w/ embedded structures, so decoding still works correctly.

https://github.com/longhorn/longhorn-manager/pull/1304/files#diff-22d9efdab1c8ceb9165fd0f1b5777cec7d012c74ec1d1f1d84c9e96d918b6fc0R74-R78

innobead (approved):
LGTM, waiting for the testing result.

cc @shuo-wu

PhanLe1010:
Test plan all passed: https://github.com/longhorn/longhorn/issues/3144#issuecomment-1125684793
Scale test looks smooth

controller/controller_manager.go
```diff
@@ -46,6 +46,9 @@ func StartControllers(logger logrus.FieldLogger, stopCh chan struct{}, controlle
 		return nil, nil, errors.Wrap(err, "unable to get client config")
 	}
 
+	config.Burst = 100
+	config.QPS = 50
```
derekbit, line 50: Curious. How to determine Burst and QPS?

datastore/longhorn.go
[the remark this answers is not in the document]
derekbit: The API server will mutate resource and then create resource creation when receive the request from `LonghornV1beta2().Snapshots(s.namespace).Create()`, so there is only one API server request.

controller/snapshot_controller.go
```diff
@@ -0,0 +1,585 @@
...
+
+		namespace:              namespace,
+		controllerID:           controllerID,
+		kubeClient:             kubeClient,
+		eventRecorder:          eventBroadcaster.NewRecorder(scheme, v1.EventSource{Component: "longhorn-snapshot-controller"}),
+		ds:                     ds,
+		engineClientCollection: engineClientCollection,
+	}
+
+	sc.monitor = &SnapshotMonitor{
```
derekbit, line 97: Isn't the snapshot monitor used?

controller/snapshot_controller.go
```diff
@@ -0,0 +1,585 @@
...
+	eventRecorder record.EventRecorder
+
+	ds                     *datastore.DataStore
+	cacheSyncs             []cache.InformerSynced
+	engineClientCollection engineapi.EngineClientCollection
+
+	monitor *SnapshotMonitor
+}
+
+type SnapshotCache struct {
```
derekbit, line 47: Isn't this struct used?

controller/controller_manager.go
derekbit: Another question: do we need to make them configurable?

derekbit (approved):
In general, LGTM. Only some NIT questions.

app/recurring_job.go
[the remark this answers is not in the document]
shuo-wu: Not yet. The snapshot cleanup in this job relies on the label rather than this naming format. But it's better to have a test for this part:
1. Launch the old Longhorn system
2. Create a recurring job then wait for the recurring snapshot creation.
3. Upgrade to the test version
4. Update the recurring job so that the cron job will be recreated with the new image
5. Wait and check if the old snapshots can be cleaned up as usual.

controller/snapshot_controller.go
```diff
@@ -0,0 +1,585 @@
...
+
+	vol, err := sc.ds.GetVolumeRO(snap.Spec.Volume)
+	if err != nil {
+		sc.queue.Add(key)
+		return
+	}
+
+	if vol.Status.OwnerID != sc.controllerID {
+		return
+	}
```
shuo-wu, line 147: NIT: Should we check these here?

controller/snapshot_controller.go
```diff
@@ -0,0 +1,585 @@
...
+
+	snapshots = getSnapshotsForEnqueuingFromEngines(oldEngine, curEngine, snapshots)
+	for _, snap := range snapshots {
+		sc.enqueueSnapshot(snap)
+	}
+
+	return
+}
+
+func getSnapshotsForEnqueuingFromEngines(oldEngine, curEngine *longhorn.Engine, snapshots map[string]*longhorn.Snapshot) map[string]*longhorn.Snapshot {
```
shuo-wu, line 214: NIT: rename this. e.g. `filterSnapshotsForEngineEnqueuing`

controller/snapshot_controller.go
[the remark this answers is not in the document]
shuo-wu: I mean, I don't need to check `engine.Status.CurrentState != longhorn.InstanceStateRunning` or even get the engine multiple times in one reconcile loop. If there is no sudden change in the engine, the controller just needs to error out then retry the reconciliation

controller/snapshot_controller.go
```diff
@@ -0,0 +1,585 @@
...
+	isPurging := false
+	for _, status := range purgeStatus {
+		if status.IsPurging {
+			isPurging = true
+			break
+		}
+	}
+	if !isPurging {
+		sc.logger.Infof("start SnapshotPurge to delete snapshot %v", snapshot.Name)
+		if err := client.SnapshotPurge(); err != nil {
```
shuo-wu, line 542: We may need to add a backoff/retry interval here. (As a TODO)
Considering that some snapshots cannot be purged immediately. This will lead to lots of meaningless purge calls to the engine.

controller/controller_manager.go
shuo-wu: Same question. Are these 2 numbers large enough to satisfy the requests for the thousands of snapshot/backup CRs?

shuo-wu (dismissed):
In general LGTM.

For the throttling issue:
1. we may need to take care of the config in the webhook servers.
2. The config modification just loosens restrictions in the client side. And the API server throttling is still controlled by the parameters we mentioned last time, right?

controller/controller_manager.go
PhanLe1010: The default value for those variable are 10 and 5: [link](https://github.com/kubernetes/client-go/blob/77f63643f951f19681397a995fe0916d2d5cb992/rest/config.go#L45-L46)  and [link](https://github.com/kubernetes/client-go/blob/77f63643f951f19681397a995fe0916d2d5cb992/rest/config.go#L353-L365). This was the reason for the backup and long upgrading issue in v1.2.4

I increased them to 10x. It is tested with 1k volume and 20k snapshot. Recurring job taking/deleting 1000 snapshots every minute. It was smooth. From monitoring, the actual QPS was about 16 per second (smaller than our max 50 QPS)

I don't think we need to make them configurable. Many projects hard coded it:
1.  Prometheus set the values to 100 and 100 [link](
https://github.com/prometheus-operator/prometheus-operator/blob/9d68ecf289d711c66bef39d2f83429265abc6986/pkg/k8sutil/k8sutil.go#L96-L97)
1. Rancher is doing 10k for discovering phase [link](https://github.com/rancher/support-bundle-kit/pull/23/files)
1. ingress-nginx is doing 1M which is basically no throttling on client side (AKA: beating the API server until it yells back): [link](https://github.com/kubernetes/ingress-nginx/blob/e4d53786e771cc6bdd55f180674b79f5b692e552/pkg/ingress/controller/launch.go#L252-L259)
1. Kubectl is 300 burst and 50 QPS

controller/engine_controller.go
```diff
@@ -873,11 +877,17 @@ func (m *EngineMonitor) refresh(engine *longhorn.Engine) error {
 	}
 
 	defer func() {
+		if err != nil {
+			return
+		}
```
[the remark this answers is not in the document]
PhanLe1010: Thank you. Fixed

controller/volume_controller.go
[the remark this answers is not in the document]
PhanLe1010: I mean user might delete the volume CR using kubectl and Kubernetes uses background cascading deletion by default

datastore/longhorn.go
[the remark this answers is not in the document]
PhanLe1010: Got it. Thanks. Fixed

controller/snapshot_controller.go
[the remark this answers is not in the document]
PhanLe1010: Agree with the general logic you mentioned. However, volume not being attached is a common case and I would like to check and generate an event here to let the user know via `kubectl describe` that the snapshot is not being handled because the volume is not attached:
```
			sc.eventRecorder.Eventf(snapshot, v1.EventTypeWarning, "SnapshotDeleteError", "cannot delete snapshot because the volume engine %v is not running", engine.Name)

```

How do you think?

controller/snapshot_controller.go
PhanLe1010: I changed it to use the proxy client. Please take a look @c3y1huang

controller/snapshot_controller.go
PhanLe1010: Great catch. No, they are not being used. Removed. Thank you

controller/snapshot_controller.go
PhanLe1010: No, it is not being used. Removed. Thank you!

controller/snapshot_controller.go
PhanLe1010: It helps to reduce the workqueue depth significantly. In my test from 20k down to 2k (10 worker nodes)

controller/snapshot_controller.go
PhanLe1010: Thank you. Fixed

controller/snapshot_controller.go
PhanLe1010: Could you give more detail about the steps that could lead to lots of meaningless purge calls to the engine?

From testing I see that when we delete the snapshot next to volume-head -> snapshot controller get the snapshot -> check to see if the snapshot as been marked as removed -> do a snapshot purge call -> done progressing the snapshot. Even though the snapshot is still present, we don't repeatedly do snapshot purge as there is no resync for this controller. We only retry calling snapshot purge when there is something change to this snapshot (e.g., it has new children which is no longer volume-head)

PhanLe1010:
> For the throttling issue:
1. we may need to take care of the config in the webhook servers.
2. The config modification just loosens restrictions in the client side. And the API server throttling is still controlled by the parameters we mentioned last time, right?

@shuo-wu 
For 1, are you referring to adding labels and finalizer to snapshot CR instead of the datastore? If yes, I fixed it. Thanks
For 2, Yes, you are right. The problem was at our side, not the API server. You can see more testing in the GSK ticket where the support engineer tried to change all kind of flags on the API server https://suse.slack.com/archives/C03C0CS1ZQS/p1652227545511309?thread_ts=1652224375.697069&cid=C03C0CS1ZQS

app/recurring_job.go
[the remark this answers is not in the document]
PhanLe1010: Tested. The old snapshots were cleaned up correctly

innobead (approved):

shuo-wu:
After the discussion with Phan, we found that the webhook servers may not directly send CR creation/deletion/update requests to the API server hence there is no need to update the config QPS & BURST.

innobead:
> After the discussion with Phan, we found that the webhook servers may not directly send CR creation/deletion/update requests to the API server hence there is no need to update the config QPS & BURST.

Shuo, could you explain a bit? Do you mean we don't need to update the QPS and burst of the rest client? I thought the purpose is to avoid the aggressive throttling of K8s API access when handling tons of snapshots.

shuo-wu:
After simply walking through part of the admission webhook implementation, we realized instead of the webhook acting like a proxy that forwards the requests to the API server, it's more like responsible for answering the questions from the API server or the longhorn managers (not sure here).
https://github.com/longhorn/longhorn-manager/blob/1bc4007488cd10ccdf3c62706da11207fed5e66f/webhook/admission/admission.go#L109-L118

And the webhook servers use the listers (caches) of the kube-client or longhorn-client to get the CRs during mutation or validation. No update/creation/deletion requests get involved.

joshimoo:
CC @shuo-wu @innobead 

**Knowledge sharing:**
- The client will communicate with the API server directly, it has no knowledge of the admission webhooks.
- The api server will receive a request from client than matches it against the rules setup as part of the admission registry (i.e. path/objects/namespaces etc)
- it will determine the collection of webhooks responsible for this object, than sent `AdmissionReview` requests to the set of webhooks.
- it will wait for the `AdmissionReview` responses and determine whether the operation can proceed or fail it.
- since the webhooks are now part of the api request/response cycle it's important that they are quick and non blocking due to their impact on the api latency.

Example Diagram


**REF:**
- https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#webhook-request-and-response
- https://slack.engineering/simple-kubernetes-webhook/

controller/snapshot_controller.go
PhanLe1010: update. The 20k queue was due to the worker queue cannot take out key fast enough. Agree that we should keep this enqueue path simple. PR is created at https://github.com/longhorn/longhorn-manager/pull/1266

PR #1307, opened 2022-04-22
Update PV annotation according to volume scheduled condition

derekbit:
Currently, the Condition["Schedule"].message is still empty even if there is a scheduling failure. We can update it with a verbose error message which is a combination of the errors, e.g. insufficient storage or insufficient storage, disk not found. Then, the verbose error message is updated to PV annotation.

https://github.com/longhorn/longhorn/issues/3529

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -1040,15 +1040,18 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
 	}
 
 	scheduled := true
+	aggregatedReplicaScheduledError := util.NewMultiError()
 	for _, r := range rs {
 		// check whether the replica need to be scheduled
 		if r.Spec.NodeID != "" {
 			continue
 		}
-		scheduledReplica, err := vc.scheduler.ScheduleReplica(r, rs, v)
+		scheduledReplica, multiError, err := vc.scheduler.ScheduleReplica(r, rs, v)
 		if err != nil {
 			return err
 		}
+		aggregatedReplicaScheduledError.Append(multiError)
+
 		if scheduledReplica == nil {
 			if r.Spec.HardNodeAffinity == "" {
 				log.WithField("replica", r.Name).Error("unable to schedule replica")
@@ -1068,6 +1071,8 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
 			rs[r.Name] = scheduledReplica
 		}
 	}
+
+	failureMessage := ""
 	if scheduled {
 		v.Status.Conditions = types.SetCondition(v.Status.Conditions,
 			longhorn.VolumeConditionTypeScheduled, longhorn.ConditionStatusTrue, "", "")
@@ -1091,6 +1096,23 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
 				scheduled = true
 			}
 		}
+
+		if !scheduled {
+			if len(aggregatedReplicaScheduledError) == 0 {
+				aggregatedReplicaScheduledError.Append(util.NewMultiError(longhorn.ErrorReplicaScheduleSchedulingFailed))
+			}
+			failureMessage = aggregatedReplicaScheduledError.Join()
+			scheduledCondition := types.GetCondition(v.Status.Conditions, longhorn.VolumeConditionTypeScheduled)
+			if scheduledCondition.Status == longhorn.ConditionStatusFalse {
+				v.Status.Conditions = types.SetCondition(v.Status.Conditions,
+					longhorn.VolumeConditionTypeScheduled, longhorn.ConditionStatusFalse,
+					scheduledCondition.Reason, failureMessage)
+			}
+		}
+	}
+
+	if err := vc.updatePVAnnotation(v, types.PVAnnotationLonghornVolumeSchedulingError, failureMessage); err != nil {
+		log.Warnf("Cannot update PV annotation for volume %v", v.Name)
 	}
 
 	if err := vc.reconcileVolumeSize(v, e, rs); err != nil {
@@ -3637,4 +3659,26 @@ func (vc *VolumeController) checkVolumeNotInMigration(volume *longhorn.Volume) e
 		return fmt.Errorf("cannot operate during migration")
 	}
 	return nil
+}
+
+func (vc *VolumeController) updatePVAnnotation(volume *longhorn.Volume, annotationKey, annotationVal string) error {
+	pv, err := vc.ds.GetPersistentVolume(volume.Status.KubernetesStatus.PVName)
+	if err != nil {
+		return err
+	}
+
+	if pv.Annotations == nil {
+		pv.Annotations = map[string]string{}
+	}
+
+	val, ok := pv.Annotations[annotationKey]
+	if ok && val == annotationVal {
+		return nil
+	}
+
+	pv.Annotations[annotationKey] = annotationVal
+
+	_, err = vc.ds.UpdatePersistentVolume(pv)
+
+	return err
 }
--- a/k8s/pkg/apis/longhorn/v1beta2/node.go
+++ b/k8s/pkg/apis/longhorn/v1beta2/node.go
@@ -32,6 +32,19 @@ const (
 	DiskConditionReasonDiskNotReady          = "DiskNotReady"
 )
 
+const (
+	ErrorReplicaScheduleInsufficientStorage              = "insufficient storage"
+	ErrorReplicaScheduleDiskNotFound                     = "disk not found"
+	ErrorReplicaScheduleDiskUnavailable                  = "disks are unavailable"
+	ErrorReplicaScheduleSchedulingSettingsRetrieveFailed = "failed to retrieve scheduling settings failed to retrieve"
+	ErrorReplicaScheduleTagsNotFulfilled                 = "tags not fulfilled"
+	ErrorReplicaScheduleNodeNotFound                     = "node not found"
+	ErrorReplicaScheduleNodeUnavailable                  = "nodes are unavailable"
+	ErrorReplicaScheduleEngineImageNotReady              = "none of the node candidates contains a ready engine image"
+	ErrorReplicaScheduleHardNodeAffinityNotSatisfied     = "hard affinity cannot be satisfied"
+	ErrorReplicaScheduleSchedulingFailed                 = "replica scheduling failed"
+)
+
 type DiskSpec struct {
 	// +optional
 	Path string `json:"path"`
--- a/scheduler/replica_scheduler.go
+++ b/scheduler/replica_scheduler.go
@@ -46,23 +46,22 @@ func NewReplicaScheduler(ds *datastore.DataStore) *ReplicaScheduler {
 }
 
 // ScheduleReplica will return (nil, nil) for unschedulable replica
-func (rcs *ReplicaScheduler) ScheduleReplica(replica *longhorn.Replica, replicas map[string]*longhorn.Replica, volume *longhorn.Volume) (*longhorn.Replica, error) {
+func (rcs *ReplicaScheduler) ScheduleReplica(replica *longhorn.Replica, replicas map[string]*longhorn.Replica, volume *longhorn.Volume) (*longhorn.Replica, util.MultiError, error) {
 	// only called when replica is starting for the first time
 	if replica.Spec.NodeID != "" {
-		return nil, fmt.Errorf("BUG: Replica %v has been scheduled to node %v", replica.Name, replica.Spec.NodeID)
+		return nil, nil, fmt.Errorf("BUG: Replica %v has been scheduled to node %v", replica.Name, replica.Spec.NodeID)
 	}
 
 	// get all hosts
 	nodesInfo, err := rcs.getNodeInfo()
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 
-	nodeCandidates := rcs.getNodeCandidates(nodesInfo, replica)
-
+	nodeCandidates, multiError := rcs.getNodeCandidates(nodesInfo, replica)
 	if len(nodeCandidates) == 0 {
 		logrus.Errorf("There's no available node for replica %v, size %v", replica.ObjectMeta.Name, replica.Spec.VolumeSize)
-		return nil, nil
+		return nil, multiError, nil
 	}
 
 	nodeDisksMap := map[string]map[string]struct{}{}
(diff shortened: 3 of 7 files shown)


controller/volume_controller.go
```diff
@@ -52,6 +52,8 @@ const (
 	VolumeSnapshotsWarningThreshold   = 100
 
 	LastAppliedCronJobSpecAnnotationKeySuffix = "last-applied-cronjob-spec"
+
+	PVCAnnotationHarvesterHciVolumeStatus = "harvesterhci.io/volume-status"
```
innobead, line 56: We should use longhorn namespace instead of customizing for any downstream project.

derekbit: Agree.

types/types.go
```diff
@@ -116,6 +116,8 @@ const (
 	ControlPlaneName                 = "longhorn-manager"
 
 	DefaultRecurringJobConcurrency = 10
+
+	LonghornPVAnnotationVolumeStatus = "longhorn.io/volume-status"
```
shuo-wu, line 120: NIT: will this be more straightforward? `PVAnnotationLonghornVolumeSchedulingError = "longhorn.io/volume-scheduling-error"`

controller/volume_controller.go
```diff
@@ -1054,18 +1054,25 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
 		aggregatedVerboseErrors = util.AppendVerboseErrors(aggregatedVerboseErrors, verboseErrors)
 
 		if scheduledReplica == nil {
+			verboseErrorMessage := replicaSchedulingFailureMessage(aggregatedVerboseErrors)
+			if verboseErrorMessage == "" {
+				verboseErrorMessage = longhorn.SchedulerErrorSchedulingFailed
+			}
```
shuo-wu, line 1060: I would recommend aggregating and setting the scheduling failure message outside of the `for` loop.
The logic may be a little bit tricky. VolumeController will mark the volume scheduling condition as False **as long as** there is **one** replica failing at it. But the error message is the **summarization** of **all** scheduling failures and should be set after the iteration.

controller/volume_controller.go
```diff
@@ -1040,26 +1040,32 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
 	}
 
 	scheduled := true
+	aggregatedVerboseErrors := make(map[string]struct{}, 0)
```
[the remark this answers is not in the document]
derekbit: Yes. But what's the purpose of keeping the replica name? Would you like to show the error of each replica in the aggregated error message? e.g.
"summary of the error message: replica1: <error message>, replica2: <error message>...."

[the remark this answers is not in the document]
shuo-wu: Do you think it's necessary? It seems that most of the replica scheduling failure reasons would be similar/identical.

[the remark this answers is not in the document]
derekbit: I don't think we need the replica names and lengthy and similar/identical failure reasons.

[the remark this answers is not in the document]
shuo-wu: OK, then ignore the related review comments

scheduler/replica_scheduler.go
```diff
@@ -84,25 +83,27 @@ func (rcs *ReplicaScheduler) ScheduleReplica(replica *longhorn.Replica, replicas
...
 
 	// schedule replica to disk
 	rcs.scheduleReplicaToDisk(replica, diskCandidates)
 
-	return replica, nil
+	return replica, nil, nil
 }
 
-func (rcs *ReplicaScheduler) getNodeCandidates(nodesInfo map[string]*longhorn.Node, schedulingReplica *longhorn.Replica) map[string]*longhorn.Node {
+func (rcs *ReplicaScheduler) getNodeCandidates(nodesInfo map[string]*longhorn.Node, schedulingReplica *longhorn.Replica) (map[string]*longhorn.Node, map[string]struct{}) {
```
innobead, line 100: It seems the returned verbose error type `map[string]struct{}` frequently used here, so I would suggest having a type alias which is better for debugging/development.

innobead (approved):
In general, LGTM. After tackling Shuo's comments, we can merge this.

Need to backport to 1.2.x, so we need to have the error types in v1beta API when doing backport. Thank you!

controller/volume_controller.go
innobead: Yeah, indeed this is tricky, but I feel the intermediate messages should be fine but leave it to @derekbit to decide.

scheduler/replica_scheduler.go
derekbit: okay

scheduler/replica_scheduler.go
```diff
@@ -46,23 +46,22 @@ func NewReplicaScheduler(ds *datastore.DataStore) *ReplicaScheduler {
...
+		return nil, nil, err
 	}
 
-	nodeCandidates := rcs.getNodeCandidates(nodesInfo, replica)
-
+	nodeCandidates, verboseError := rcs.getNodeCandidates(nodesInfo, replica)
 	if len(nodeCandidates) == 0 {
 		logrus.Errorf("There's no available node for replica %v, size %v", replica.ObjectMeta.Name, replica.Spec.VolumeSize)
-		return nil, nil
+		return nil, util.VerboseErrors{verboseError: struct{}{}}, nil
```
shuo-wu, line 64: Create and use a function here `NewVerboseErrors(strList...)`

scheduler/replica_scheduler.go
```diff
@@ -84,39 +83,46 @@ func (rcs *ReplicaScheduler) ScheduleReplica(replica *longhorn.Replica, replicas
...
 
 	// schedule replica to disk
 	rcs.scheduleReplicaToDisk(replica, diskCandidates)
 
-	return replica, nil
+	return replica, nil, nil
 }
 
-func (rcs *ReplicaScheduler) getNodeCandidates(nodesInfo map[string]*longhorn.Node, schedulingReplica *longhorn.Replica) map[string]*longhorn.Node {
+func (rcs *ReplicaScheduler) getNodeCandidates(nodesInfo map[string]*longhorn.Node, schedulingReplica *longhorn.Replica) (nodeCandidates map[string]*longhorn.Node, verboseError string) {
```
shuo-wu, line 100: Always return `util.VerboseErrors` rather than a string

shuo-wu (approved):
In general LGTM

manager/kubernetes.go
```diff
@@ -61,7 +61,15 @@ func (m *VolumeManager) PVCreate(name, pvName, fsType, secretNamespace, secretNa
...
 	}
 
-	pv, err = m.ds.CreatePersistentVolume(pv)
+	scheduledCondition := types.GetCondition(v.Status.Conditions, longhorn.VolumeConditionTypeScheduled)
+	if scheduledCondition.Status == longhorn.ConditionStatusFalse {
+		if pv.Annotations == nil {
+			pv.Annotations = map[string]string{}
+		}
+		pv.Annotations[types.PVAnnotationLonghornVolumeSchedulingError] = scheduledCondition.Message
+	}
```
innobead, line 70: I would prefer this should be reconciled in the controller only, so we don't need to have sperate update logic in more than one place. The downside will be some delay for the update, but it will be finally consistent. WDYT?

k8s/pkg/apis/longhorn/v1beta2/node.go
```diff
@@ -32,6 +32,19 @@ const (
...
+const (
+	SchedulerErrorInsufficientStorage          = "insufficient storage"
+	SchedulerErrorDiskNotFound                 = "disk not found"
+	SchedulerErrorNoDisAvailable               = "no disk available"
+	SchedulerErrorSettingIOError               = "setting io error"
+	SchedulerErrorTagsNotFullfilled            = "tags not fullfilled"
+	SchedulerErrorNodeNotFound                 = "node not found"
+	SchedulerErrorNoNodeAvailable              = "none of the nodes is available"
+	SchedulerErrorEngineImageNotReady          = "none of the node candidates contains a ready engine image"
+	SchedulerErrorHardAntiAffinityNotSatisfied = "hard anti-affinity cannot be satisfied"
```
innobead, line 44: Change to alias with the spec property -> `SchedulerErrorHardNodeAffinityNotSatisfied`, the same as the description needed to adjust.

scheduler/replica_scheduler.go
```diff
@@ -84,39 +83,46 @@ func (rcs *ReplicaScheduler) ScheduleReplica(replica *longhorn.Replica, replicas
...
 	rcs.scheduleReplicaToDisk(replica, diskCandidates)
 
-	return replica, nil
+	return replica, nil, nil
 }
 
-func (rcs *ReplicaScheduler) getNodeCandidates(nodesInfo map[string]*longhorn.Node, schedulingReplica *longhorn.Replica) map[string]*longhorn.Node {
+func (rcs *ReplicaScheduler) getNodeCandidates(nodesInfo map[string]*longhorn.Node, schedulingReplica *longhorn.Replica) (nodeCandidates map[string]*longhorn.Node, verboseErrors util.VerboseErrors) {
 	if schedulingReplica.Spec.HardNodeAffinity != "" {
 		node, exist := nodesInfo[schedulingReplica.Spec.HardNodeAffinity]
```
innobead, line 102: Can we use string formatting w/ interpolation to add `schedulingReplica.Spec.HardNodeAffinity` into the error message for having much detailed info? like fmt.Sprintf(longhorn.SchedulerErrorHardNodeAffinityNotSatisfied, schedulingReplica.Spec.HardNodeAffinity).

scheduler/replica_scheduler.go
```diff
@@ -203,48 +214,77 @@ func (rcs *ReplicaScheduler) getDiskCandidates(nodeInfo map[string]*longhorn.Nod
...
 }
 
-func (rcs *ReplicaScheduler) filterNodeDisksForReplica(node *longhorn.Node, disks map[string]struct{}, replicas map[string]*longhorn.Replica, volume *longhorn.Volume, requireSchedulingCheck bool) map[string]*Disk {
-	preferredDisk := map[string]*Disk{}
+func (rcs *ReplicaScheduler) filterNodeDisksForReplica(node *longhorn.Node, disks map[string]struct{}, replicas map[string]*longhorn.Replica, volume *longhorn.Volume, requireSchedulingCheck bool) (preferredDisks map[string]*Disk, verboseErrors util.VerboseErrors) {
+	verboseErrors = util.NewVerboseErrors()
+	preferredDisks = map[string]*Disk{}
+
+	if len(disks) == 0 {
+		verboseErrors[longhorn.SchedulerErrorNoDisAvailable] = struct{}{}
```
innobead, line 284: The same, if possible, add a formatted error w/ node name, so it's better to know which node has no disks available? 

The same as below errors.

scheduler/replica_scheduler.go
```diff
@@ -261,13 +301,15 @@ func (rcs *ReplicaScheduler) filterNodeDisksForReplica(node *longhorn.Node, disk
 		}
 		if !diskFound {
 			logrus.Errorf("Cannot find the spec or the status for disk %v when scheduling replica", diskUUID)
+			verboseErrors.AppendError(longhorn.SchedulerErrorDiskNotFound)
```
innobead, line 304: Should we have a formatted message for the disk name or uuid?

scheduler/replica_scheduler.go
```diff
@@ -261,13 +301,15 @@ func (rcs *ReplicaScheduler) filterNodeDisksForReplica(node *longhorn.Node, disk
...
 			continue
 		}
 		if requireSchedulingCheck {
 			info, err := rcs.GetDiskSchedulingInfo(diskSpec, diskStatus)
 			if err != nil {
 				logrus.Errorf("Fail to get settings when scheduling replica: %v", err)
-				return preferredDisk
+				verboseErrors.AppendError(longhorn.SchedulerErrorSettingIOError)
```
innobead, line 311: Suggest changing the error much meaningful like `longhorn.SchedulerErrorSchedulingSettingsGetError`

innobead (changes requested):

k8s/pkg/apis/longhorn/v1beta2/node.go
```diff
@@ -32,6 +32,19 @@ const (
 	DiskConditionReasonDiskNotReady          = "DiskNotReady"
 )
 
+const (
+	SchedulerErrorInsufficientStorage          = "insufficient storage"
+	SchedulerErrorDiskNotFound                 = "disk not found"
+	SchedulerErrorNoDisAvailable               = "no disk available"
+	SchedulerErrorSettingIOError               = "setting io error"
```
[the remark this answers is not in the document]
derekbit: Hard to know the error. This error is caused by the failure while getting setting from etcd.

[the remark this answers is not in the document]
innobead: I mean, we need to change a meaningful error like "scheduling settings failed to retrieve".

controller/volume_controller.go
```diff
@@ -1091,6 +1097,24 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
...
+			scheduledCondition := types.GetCondition(v.Status.Conditions, longhorn.VolumeConditionTypeScheduled)
+			if scheduledCondition.Status == longhorn.ConditionStatusFalse {
+				v.Status.Conditions = types.SetCondition(v.Status.Conditions,
+					longhorn.VolumeConditionTypeScheduled, longhorn.ConditionStatusFalse,
+					scheduledCondition.Reason,
+					failureMessage)
+			}
+		}
+		if err := vc.updatePVAnnotation(v, types.PVAnnotationLonghornVolumeSchedulingError, failureMessage); err != nil {
+			log.Warnf("Cannot update PV annotation for volume %v", v.Name)
```
innobead, line 1116: This is good. We should just warn out instead of returning the error, because this is an optional update outside Longhorn resources.

scheduler/replica_scheduler.go
derekbit: If there are tens or hundreds of disks, the annotation become very lengthy. IMO, the final result is that the replica can be scheduled in each disk and on each node. Current annotation value gives enough clues and can helps the administrator to understand what happen in the system.
For more information, user can the details in longhorn logs.

innobead: Agreed that the concern is fair. Let's keep it as is, but just ensure there will be no duplicated errors in the aggregated error.

util/util.go
```diff
@@ -895,3 +895,33 @@ func GetVolumeMeta(path string) (*VolumeMeta, error) {
 	}
 	return meta, nil
 }
+
+type VerboseErrors map[string]struct{}
```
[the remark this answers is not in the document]
derekbit: After checking  https://github.com/hashicorp/go-multierror and https://github.com/uber-go/multierr, both are not good for our use case. They can append and combine the errors and store errors in a slice struct, but they cannot de-duplicate the same errors. We still need to implement our `combine` function using `errors.Is`, but the bigO is n^2, looks not a good idea...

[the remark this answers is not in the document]
innobead: Then we just keep the current way as it first.

controller/volume_controller.go
```diff
@@ -1091,6 +1097,24 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
...
 		}
+
+		failureMessage := ""
+		if !scheduled {
+			if len(aggregatedVerboseErrors) == 0 {
+				aggregatedVerboseErrors.AppendError(longhorn.SchedulerErrorSchedulingFailed)
+			}
+			failureMessage = aggregatedVerboseErrors.Join()
+			scheduledCondition := types.GetCondition(v.Status.Conditions, longhorn.VolumeConditionTypeScheduled)
+			if scheduledCondition.Status == longhorn.ConditionStatusFalse {
```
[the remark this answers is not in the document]
derekbit: Yes, I think we need `GetCondition` for getting the `Reason`. Two failure reasons in the scheduled condition:  `VolumeConditionReasonLocalReplicaSchedulingFailure` and `VolumeConditionReasonReplicaSchedulingFailure`.

controller/volume_controller.go
```diff
@@ -1091,6 +1097,24 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
...
+				v.Status.Conditions = types.SetCondition(v.Status.Conditions,
+					longhorn.VolumeConditionTypeScheduled, longhorn.ConditionStatusFalse,
+					scheduledCondition.Reason,
+					failureMessage)
+			}
+		}
+		if err := vc.updatePVAnnotation(v, types.PVAnnotationLonghornVolumeSchedulingError, failureMessage); err != nil {
+			log.Warnf("Cannot update PV annotation for volume %v", v.Name)
+		}
```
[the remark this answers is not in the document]
derekbit: Good idea, thank you.

manager/kubernetes.go
derekbit: Yes. If the delay is acceptable, I prefer your idea.

util/multierror.go
```diff
@@ -0,0 +1,31 @@
+package util
+
+import (
+	"strings"
+)
+
+type MultiError map[string]struct{}
+
+func NewMultiError(errs ...string) MultiError {
```
innobead, line 9: Usually, we will use Error interface for error, but here use string instead. This is fine for now for the specific need. However, it will be good to improve this later, to make the MultiError support Error args instead of a raw string because you can use err.Error() to get the error message.

innobead (approved):
LGTM

util/multierror.go
innobead: Also, I would prefer using error.go instead of a specific multierror.go, because it will be much extensible in the future, but this is quite minor now :slightly_smiling_face:

derekbit: Yes, the naming is a bit of weird. In order to do a fast deduplication of errors, `map[string]error` is preferred, but I'm not quiet sure `ERRORREPLICASCHEDULEXXX` should string or error for now...
Maybe we should write our own `multierror` package following errors interface for our use case and replace current impl with it. Let me think about it more.

innobead: yeah, I know you want to use map for set purpose.

PR #1308, opened 2022-04-26
Add progress status to the upgrade path v1.2.2 to v1.2.3

PhanLe1010:
To track down the progress of the upgrade path. This could help to quickly identify the root cause of long upgrading time.

longhorn/longhorn#3890


upgrade/util/util.go
```diff
@@ -16,6 +20,63 @@ import (
...
+	logrus.Infof("%v: current progress %v%% (%v/%v)", pm.description, pm.currentProgressInPercentage, pm.currentValue, pm.targetValue)
+}
+
+func (pm *ProgressMonitor) Inc() int {
+	pm.mutex.Lock()
+	defer pm.mutex.Unlock()
+
+	oldProgressInPercentage := pm.currentProgressInPercentage
+
+	pm.currentValue += 1
```
innobead, line 53: For flexibility, you probably can increase the value by the provided argument.

innobead (approved):

upgrade/util/util_test.go
```diff
@@ -0,0 +1,44 @@
...
+			}
+			for j := 0; j < count; j++ {
+				pm.Inc()
+			}
+		}(i)
+	}
+
+	wg.Wait()
+
+	currentValue, _, currentProgressInPercentage := pm.ReadCurrentProgress()
```
innobead, line 37: nit: it's better to test for all public exported interfaces, but it's minor for this case because progress is pretty straightforward.

innobead:
@PhanLe1010 needs to fix the lint error in the PR check. After that, we can merge this.

upgrade/util/util_test.go
PhanLe1010: Thank you! Fixed

PhanLe1010:
Thank you. Fixed the lint error

innobead (approved):

innobead:
We need to backport this to 1.2.x.

PR #1310, opened 2022-04-26
Fix issues found in longhorn-tests/test_orphan.py

derekbit:
Fix issues found in [longhorn-tests/test_orphan.py](url) multi-disks and successive test cases.

1. Upgrade copier to latest version
2. orphan: generate new and missing orphan lists in node monitor
    node monitor generates the new and missing orphan lists, and then simplify node controller logic.
3. Refuse node update before spec and status of disks are synced
4. api: also display disks without specs
    
https://github.com/longhorn/longhorn/issues/685


controller/node_controller.go
```diff
@@ -1067,7 +1067,11 @@ func (nc *NodeController) deleteOrphans(node *longhorn.Node, diskName string, di
 				if err := nc.ds.DeleteOrphan(orphan.Name); err != nil && !apierrors.IsNotFound(err) {
 					return errors.Wrapf(err, "failed to delete orphan %v", orphan.Name)
 				}
-				diskInfo.OrphanedReplicaDirectoryNames[orphan.Spec.Parameters[longhorn.OrphanDataName]] = longhorn.OrphanCRStatusDeleted
+
+				if _, ok := diskInfo.OrphanedReplicaDirectoryNames[orphan.Spec.Parameters[longhorn.OrphanDataName]]; ok {
+					diskInfo.OrphanedReplicaDirectoryNames[orphan.Spec.Parameters[longhorn.OrphanDataName]] = longhorn.OrphanCRStatusDeleted
```
innobead, line 1074: Do you mean we don't need to orphan record deleted records in each diskInfo? then do we still need `longhorn.OrphanCRStatusDeleted`?

controller/monitor/node_monitor.go
```diff
@@ -98,9 +98,13 @@ func (m *NodeMonitor) GetCollectedData() (interface{}, error) {
 	m.collectedDataLock.RLock()
 	defer m.collectedDataLock.RUnlock()
 
-	data := make(map[string]*CollectedDiskInfo, 0)
-	if err := copier.Copy(&data, &m.collectedData); err != nil {
-		return data, errors.Wrap(err, "failed to copy node monitor collected data")
+	v, err := copystructure.Copy(m.collectedData)
```
innobead, line 101: It's OK, but FYI. the recent version of copier has deepcopy as well, https://github.com/jinzhu/copier/blob/master/copier.go#L74.

controller/node_controller.go
derekbit: > Do you mean we don't need to orphan record deleted records in each diskInfo? 

The purpose is to avoid adding extra replica directories in `OrphanedReplicaDirectoryNames` although it does not introduce any impact. Just make the `OrphanedReplicaDirectoryNames` clear and easy to debug.

We still need the `OrphanCRStatusDeleted `. If the node monitor does not finish the collection of the data, we won't hope the node monitor recreates the orphan CRs for the deleted orphaned directories according to the old OrphanedReplicaDirectoryNames list.

innobead (approved):

controller/monitor/node_monitor.go
derekbit: Let me update to latest copier

derekbit: Addressed!

shuo-wu (commented):
One question: Does `alignDiskSpecAndStatus` in NodeController work correctly? I don't understand why the node monitor still needs to store the NodeSpec. Cannot this function filter out the mismatching items if the NodeSpec during the NodeController reconciliation is different from that during the NodeMonitor info collection?

controller/node_controller.go
derekbit: Actually, we only compare the disk path and some fields with the collected data. Thus, node controller will possibly re-do creation or deletion of orphan CRs. We need a way to avoid the re-creation of the deleted CRs.

derekbit:
> One question: Does `alignDiskSpecAndStatus` in NodeController work correctly? I don't understand why the node monitor still needs to store the NodeSpec. Cannot this function filter out the mismatching items if the NodeSpec during the NodeController reconciliation is different from that during the NodeMonitor info collection?

@shuo-wu addessed according to our discussion.

api/model.go
```diff
@@ -1477,6 +1477,23 @@ func toNodeResource(node *longhorn.Node, address string, apiContext *api.ApiCont
 		}
 		disks[name] = di
 	}
+	for name, status := range node.Status.DiskStatus {
```
innobead, line 1480: I know there will probably a delay of disk info between spec and status, but why we need this change in the API? any specific purpose? I feel this will cause some confusion.

innobead (commented):
LGTM, except one comment.

api/model.go
derekbit: Yeah, this is a change for the regression test.
For example, after doing `node.diskUpdate`, get the node and check the disk spec and status and see if the update is successful. If not, retry `node.diskUpdate`.
I think this change is confusing as well, so I'll drop this commit and see if there is better way on the regression test side.

innobead (approved):

controller/monitor/node_monitor.go
```diff
@@ -250,15 +257,42 @@ func (m *NodeMonitor) getOrphanedReplicaDirectoryNames(node *longhorn.Node, disk
 		}
 	}
 
+	// Find out the new/missing orphaned directories by checking with orphan CRs
```
shuo-wu, line 260: Not sure why you move this logic to the node monitor... I think it's NodeController's duty to filter out the mismatching orphans...

webhook/resources/node/validator.go
```diff
@@ -71,6 +71,11 @@ func (n *nodeValidator) Update(request *admission.Request, oldObj runtime.Object
 			oldNode.Name), "")
 	}
 
+	// Ensure the node controller already syncs the disk spec and status.
+	if len(oldNode.Spec.Disks) != len(oldNode.Status.DiskStatus) {
```
shuo-wu, line 75: It's better to check if the entries in `Status.DiskStatus` exists in `Spec.Disks`.
This check cannot prevent the following case:
```
Spec.Disk: Disk A, Disk B
Status.DiskStatus: Disk A, Disk C
```

controller/monitor/node_monitor.go
derekbit: I though the previous checking is hard to read. Maybe we can move it back and group them into new and missing after node controller gets the new list.

derekbit: Another consideration is that if we separate new and missing dirs in node monitor, then we don't have to do `isVolumeMetaFileExist` for all dirs blindly in the node monitor every 30 seconds.

webhook/resources/node/validator.go
derekbit: Thank you. I was not aware of the case....

innobead: I thought it should be finally consistent? Should we need to worry this case?

derekbit: Yeah. This is for the corner case when node controller failed to reconcile node before `alignDiskSpecAndStatus`.

innobead: I see, yeah we need to check the content as well not just size.

derekbit: Addressed in https://github.com/longhorn/longhorn-manager/pull/1311.

controller/monitor/node_monitor.go
shuo-wu: As I mentioned, the monitor is responsible for detecting the disk status only regardless of how NodeController reconciles the status and handles the orphans.
After renaming the related fields or functions, putting these checks in `createOrphans` or `deleteOrphans` is not as messy as you are worried about.

And I was thinking if you are trying to avoid the unnecessary `isVolumeMetaFileExist` as well. 
IMO, the cost should not be expensive. Even if there are lots of missing orphans (extra CRs) here, the CRs will be removed soon and those unnecessary metafile checks are one-time operations.

PR #1311, opened 2022-04-29
validator: reject update request before syncing disk spec and status

derekbit:
The disk spec and status should be one-to-one matching.

Longhorn 685

### DIFF
--- a/webhook/resources/node/validator.go
+++ b/webhook/resources/node/validator.go
@@ -72,7 +72,7 @@ func (n *nodeValidator) Update(request *admission.Request, oldObj runtime.Object
 	}
 
 	// Ensure the node controller already syncs the disk spec and status.
-	if len(oldNode.Spec.Disks) != len(oldNode.Status.DiskStatus) {
+	if !isNodeDiskSpecAndStatusSynced(oldNode) {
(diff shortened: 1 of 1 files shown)


webhook/resources/node/validator.go
```diff
@@ -142,3 +142,17 @@ func (n *nodeValidator) Update(request *admission.Request, oldObj runtime.Object
...
 	return nil
 }
+
+func isNodeDiskSpecAndStatusSynced(node *longhorn.Node) bool {
+	if len(node.Spec.Disks) != len(node.Status.DiskStatus) {
+		return false
+	}
+
+	for diskName := range node.Spec.Disks {
+		if _, ok := node.Status.DiskStatus[diskName]; !ok {
```
shuo-wu, line 152: Need to check the path as well

shuo-wu (changes requested):

webhook/resources/node/validator.go
derekbit: `Path`? DiskStatus does not have `Path`.

innobead (approved):
LGTM

webhook/resources/node/validator.go
shuo-wu: Oh, sorry. You are right

PR #1313, opened 2022-05-03
environment_check: fix OS detection

derekbit:
Detect the worker OS rather than host OS.

[Longhorn 3550](https://github.com/longhorn/longhorn/issues/3550)


innobead (approved):
LGTM

PR #1314, opened 2022-05-04
node controller: fix the incorrect update of disk status

derekbit:
This bug is found by test_node.py.

Before syncing with the node monitor, don't update the storage data in diskStatus. The cleanup of the old storage data should happen when the updates of ready/scheduled conditions are available.
The incorrect update leads to the wrong StorageMaximum and StorageAvailable until next successful syncWithMonitor.

Longhorn 685

### DIFF
--- a/controller/node_controller.go
+++ b/controller/node_controller.go
@@ -426,8 +426,6 @@ func (nc *NodeController) syncNode(key string) (err error) {
 		return nil
 	}
 
-	alignDiskSpecAndStatus(node)
-
 	mon, err := nc.checkMonitor(nc.controllerID)
 	if err != nil {
 		return err
@@ -561,6 +559,8 @@ func (nc *NodeController) enqueueKubernetesNode(obj interface{}) {
 }
 
 func (nc *NodeController) syncDiskStatus(node *longhorn.Node, collectedDataInfo map[string]*monitor.CollectedDiskInfo) error {
+	alignDiskSpecAndStatus(node)
+
 	notReadyDiskInfoMap, readyDiskInfoMap := nc.findNotReadyAndReadyDiskMaps(node, collectedDataInfo)
 
 	for _, diskInfoMap := range notReadyDiskInfoMap {


controller/node_controller.go
```diff
@@ -561,6 +561,8 @@ func (nc *NodeController) enqueueKubernetesNode(obj interface{}) {
 }
 
 func (nc *NodeController) syncDiskStatus(node *longhorn.Node, collectedDataInfo map[string]*monitor.CollectedDiskInfo) error {
+	cleanupDiskStatusOldStorageData(node)
```
shuo-wu, line 564: Why not directly call `alignDiskSpecAndStatus` here instead?

derekbit: Yes, we can do this.

I'm thinking whether we should return error directly if `nc.syncWithMonitor` failed no matter what reason. But this might introduce many retries while reconciling the node CR.
https://github.com/longhorn/longhorn-manager/blob/master/controller/node_controller.go#L437

Currently, we didn't return error and just align the disk spec and status if `nc.syncWithMonitor` failed. Thus, I prefer to align the disk spec and status first before exiting the reconciliation.

shuo-wu: - Yes. It's better not to return errors for the mismatching case.
- The reason I prefer to put (previous) `alignDiskSpecAndStatus` here is that the node controller can handle all disk related operations inside `syncDiskStatus`.

innobead: It seems no hurt and one place to handle that. The same suggestion, move alighDiskSpecAndStatus to the syncDiskStatus. 

But this is a good finding to avoid unintentionally updating the wrong disk status.

derekbit: > The reason I prefer to put (previous) alignDiskSpecAndStatus here is that the node controller can handle all disk related operations inside syncDiskStatus.

Sure. I agree your opinion.

BTW, looks we cannot know if diskUpdate (in test_node.py) takes effect or not… and there is no retry mechanism in diskUpdate. I have no better idea for the retry.
The mismatching disk spec and status because of the time delay leads to the false alarm in test_node.py

innobead (approved):

PR #1315, opened 2022-05-04
Fix verifyCreation input resource name

derekbit:
Use the resouce name returned by creation API as the verifyCreation
argument.

[Longhorn 3915](https://github.com/longhorn/longhorn/issues/3915)

### DIFF
--- a/datastore/longhorn.go
+++ b/datastore/longhorn.go
@@ -388,7 +388,7 @@ func (s *DataStore) CreateVolume(v *longhorn.Volume) (*longhorn.Volume, error) {
 		return ret, nil
 	}
 
-	obj, err := verifyCreation(v.Name, "volume", func(name string) (runtime.Object, error) {
+	obj, err := verifyCreation(ret.Name, "volume", func(name string) (runtime.Object, error) {
 		return s.GetVolumeRO(name)
 	})
 	if err != nil {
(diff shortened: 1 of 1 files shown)


datastore/longhorn.go
```diff
@@ -388,7 +388,7 @@ func (s *DataStore) CreateVolume(v *longhorn.Volume) (*longhorn.Volume, error) {
 		return ret, nil
 	}
 
-	obj, err := verifyCreation(v.Name, "volume", func(name string) (runtime.Object, error) {
+	obj, err := verifyCreation(ret.Name, "volume", func(name string) (runtime.Object, error) {
```
innobead, line 391: @joshimoo @shuo-wu do you think we still need this verification in each resource creation/update? I feel this is quite a blocking way for a specific purpose.

joshimoo: From a risk perspective for the moment we should keep it in, this is mostly because our controllers aren't great :/

Since the cache isn't guaranteed to be strongly consistent, our controllers should be able to deal with that and not expect a read after write to necessary immediately have the created/updated resource.

But at the moment our controller code doesn't necessary work in that way.
This was added a long time ago in https://github.com/longhorn/longhorn/issues/133 as a workaround.

innobead (approved):

PR #1316, opened 2022-05-04
Create Longhorn finalizer in the mutating webhook

derekbit:
CreateXXX, e.g. CreateBackup, adds Longhorn finalizer to new created resource.
To make the behavior consistent between API and Resource operation using CLI, we can move the longhorn finalizer creation logic to the mutating webhook.

[Longhorn 1315](https://github.com/longhorn/longhorn/issues/3918)

### DIFF
--- a/webhook/resources/backup/mutator.go
+++ b/webhook/resources/backup/mutator.go
@@ -1,12 +1,15 @@
 package backup
 
 import (
+	"github.com/pkg/errors"
 	admissionregv1 "k8s.io/api/admissionregistration/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 
 	"github.com/longhorn/longhorn-manager/datastore"
 	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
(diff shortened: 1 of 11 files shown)


webhook/resources/backup/mutator.go
```diff
@@ -33,14 +43,29 @@ func (b *backupMutator) Resource() admission.Resource {
...
-}
+	var patchOps admission.PatchOps
 
-func (b *backupMutator) Update(request *admission.Request, oldObj runtime.Object, newObj runtime.Object) (admission.PatchOps, error) {
-	return mutateBackup(newObj)
+	backup := newObj.(*longhorn.Backup)
+
+	if backup.Spec.Labels == nil {
+		patchOps = append(patchOps, `{"op": "replace", "path": "/spec/labels", "value": {}}`)
+	}
```
innobead, line 52: The code is actually duplicated, so I prefer this can be reusable as before.

innobead (changes requested):
I found some duplicated code in resource create/update. I do prefer you can keep the reusable as beofre, so that can be shared in create/update methods.

innobead (approved):
LGTM

PR #1331, opened 2022-05-13
[Backport] [v1.2.5] Fixed throttling issue with the client-go rest client inside Longhorn manager

PhanLe1010:
Items:
1. Fixed throttling issue with the client-go rest client inside Longhorn manager. This is the reason for the long upgrading time from v1.2.3 to v1.2.4 when the cluster have a lot of backups
2. Add metrics for the client-go rest client inside Longhorn manager

longhorn/longhorn#3980

ref: https://github.com/rancher/support-bundle-kit/pull/23

### DIFF
--- a/controller/controller_manager.go
+++ b/controller/controller_manager.go
@@ -46,6 +46,9 @@ func StartControllers(logger logrus.FieldLogger, stopCh chan struct{}, controlle
 		return nil, nil, errors.Wrap(err, "unable to get client config")
 	}
 
+	config.Burst = 100
+	config.QPS = 50
+
 	kubeClient, err := clientset.NewForConfig(config)
 	if err != nil {
 		return nil, nil, errors.Wrap(err, "unable to get k8s client")
--- a/upgrade/upgrade.go
+++ b/upgrade/upgrade.go
@@ -53,6 +53,11 @@ func Upgrade(kubeconfigPath, currentNodeID string) error {
 		return errors.Wrap(err, "unable to get client config")
 	}
 
+	// There is only one leading Longhorn manager that is doing modification to the CRs.
+	// Increase this value so that leading Longhorn manager can finish upgrading faster
+	config.Burst = 1000
+	config.QPS = 1000
+
 	kubeClient, err := clientset.NewForConfig(config)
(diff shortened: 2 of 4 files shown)


innobead (approved):

controller/controller_manager.go
```diff
@@ -46,6 +46,9 @@ func StartControllers(logger logrus.FieldLogger, stopCh chan struct{}, controlle
 		return nil, nil, errors.Wrap(err, "unable to get client config")
 	}
 
+	config.Burst = 100
+	config.QPS = 50
```
joshimoo, line 50: For reference the Default values were:
```
	DefaultQPS   float32 = 5.0
	DefaultBurst int     = 10
```
https://github.com/longhorn/longhorn-manager/blob/24d07f134011e8ef2ad01d2d69427a59c43ed87c/vendor/k8s.io/client-go/rest/config.go#L44-L45

upgrade/upgrade.go
```diff
@@ -53,6 +53,11 @@ func Upgrade(kubeconfigPath, currentNodeID string) error {
 		return errors.Wrap(err, "unable to get client config")
 	}
 
+	// There is only one leading Longhorn manager that is doing modification to the CRs.
+	// Increase this value so that leading Longhorn manager can finish upgrading faster
+	config.Burst = 1000
+	config.QPS = 1000
```
joshimoo, line 59: This is a good use of a comment, since it explains the why this is okay :)

joshimoo (commented):
Good change, I like that you added the metrics as well :)
Do you have a rough graph/number of how many connections we are doing as part of the controllers during normal use?

PhanLe1010:
I have this picture that show the comparison before and after changing the  rate. This cluster has 1k volumes and 20k snapshot CRs. There is a recurring job that is taking 1k snapshots every minutes and deleting another 1k snapshots every minutes.
The graph on left side was before the change, the total request per second to API server is capped at around 4. The graph on right side was after the change, the total request per second to API server is capped at around 16.

Screen Shot 2022-05-11 at 3 52 12 PM

PR #1332, opened 2022-05-15
Fix customize default settings while upgrade/update

derekbit:
Introduce ConfigMap resourceVersion into Setting
    
Record the resourceVersion of the ConfigMap, longhorn-default-setting,
in the Setting CR if the value is from ConfigMap.
    
During upgrade or longhorn-manager restart, if the resourceVersion of the
ConfigMap is different than the one in the Setting CR, update the Setting CR
using the latest value in the ConfigMap.
    
### DIFF
--- a/app/daemon.go
+++ b/app/daemon.go
@@ -120,7 +120,7 @@ func startManager(c *cli.Context) error {
 			return fmt.Errorf("cannot find customized default setting file on %v: %v", defaultSettingPath, err)
 		}
 	}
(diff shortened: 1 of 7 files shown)


innobead (approved):
LGTM

P.S. jumping out the puzzle of helm way/hook but thinking of a solution based on our current design instead is really good

innobead:
Just a note: I feel we should refactor the resource client initialization part to better reduce the duplicated code in different places in the future.

PR #1335, opened 2022-05-17
[Backport] [v1.2.5] Cache the engines and volumes object to speed up the upgrade path from v1.2.2 to v1.2.3

PhanLe1010:
longhorn/longhorn#3980
longhorn/longhorn#3996

### DIFF
--- a/upgrade/v122to123/upgrade.go
+++ b/upgrade/v122to123/upgrade.go
@@ -42,6 +42,24 @@ func upgradeBackups(namespace string, lhClient *lhclientset.Clientset) (err erro
 		return err
 	}
 
+	engines, err := lhClient.LonghornV1beta1().Engines(namespace).List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		return err
+	}
+	volumeNameToEngines := make(map[string][]*longhorn.Engine)
+	for _, e := range engines.Items {
+		volumeNameToEngines[e.Labels[types.LonghornLabelVolume]] = append(volumeNameToEngines[e.Labels[types.LonghornLabelVolume]], &e)
+	}
+
+	volumes, err := lhClient.LonghornV1beta1().Volumes(namespace).List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		return err
+	}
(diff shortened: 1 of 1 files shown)


upgrade/v122to123/upgrade.go
```diff
@@ -42,6 +42,24 @@ func upgradeBackups(namespace string, lhClient *lhclientset.Clientset) (err erro
 		return err
 	}
 
+	engines, err := lhClient.LonghornV1beta1().Engines(namespace).List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		return err
+	}
+	volumeNameToEngines := make(map[string][]longhorn.Engine)
```
shuo-wu, line 49: NIT: is it better to use pointers here? `map[string][]*longhorn.Engine`. Similar below

shuo-wu (approved):
LGTM

upgrade/v122to123/upgrade.go
PhanLe1010: Fixed it. I has been always scare of using pointer :D

innobead (approved):

upgrade/v122to123/upgrade.go
shuo-wu: It's better not to the pointers of objects init by the `for` loop. IIRC, the for loop here can be considered as:
```
var e longhorn.Engine
for idx := 0; idx < len(engines.Items); i++ {
    e = engines.Items[idx]
    p := &e  // This always points to the same variable, which may be problematic
}
```

For more detail, you can try this: https://go.dev/play/p/tjB7R-znss6

shuo-wu: This kind of syntactic sugar is tricky anyway.

PhanLe1010: Screw me! My big mistake.
 Many thanks for the alarm!

PhanLe1010: @shuo-wu 
PR created at https://github.com/longhorn/longhorn-manager/pull/1340

PR #1336, opened 2022-05-18
Cache the engines and volumes object to speed up the upgrade path from v1.2.2 to v1.2.3

PhanLe1010:
longhorn/longhorn#3980

### DIFF
--- a/upgrade/v122to123/upgrade.go
+++ b/upgrade/v122to123/upgrade.go
@@ -42,6 +42,24 @@ func upgradeBackups(namespace string, lhClient *lhclientset.Clientset) (err erro
 		return err
 	}
 
+	engines, err := lhClient.LonghornV1beta2().Engines(namespace).List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		return err
+	}
+	volumeNameToEngines := make(map[string][]*longhorn.Engine)
+	for _, e := range engines.Items {
+		volumeNameToEngines[e.Labels[types.LonghornLabelVolume]] = append(volumeNameToEngines[e.Labels[types.LonghornLabelVolume]], &e)
+	}
+
+	volumes, err := lhClient.LonghornV1beta2().Volumes(namespace).List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		return err
+	}
+	volumeMap := make(map[string]*longhorn.Volume)
+	for _, v := range volumes.Items {
+		volumeMap[v.Name] = &v
+	}
+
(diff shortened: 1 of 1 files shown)


upgrade/v122to123/upgrade.go
```diff
@@ -42,6 +42,24 @@ func upgradeBackups(namespace string, lhClient *lhclientset.Clientset) (err erro
...
+	}
+
+	volumes, err := lhClient.LonghornV1beta2().Volumes(namespace).List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		return err
+	}
+	volumeMap := make(map[string]longhorn.Volume)
+	for _, v := range volumes.Items {
+		volumeMap[v.Name] = v
+	}
```
innobead, line 61: nit: Do you think we could have this part delayed init until there is a 2-engines case happen in the default clause, because the below backup loop is not parallel.

innobead (commented):
LGTM, but some comments needed to resolve.

upgrade/v122to123/upgrade.go
```diff
@@ -42,6 +42,24 @@ func upgradeBackups(namespace string, lhClient *lhclientset.Clientset) (err erro
 		return err
 	}
 
+	engines, err := lhClient.LonghornV1beta2().Engines(namespace).List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		return err
+	}
+	volumeNameToEngines := make(map[string][]longhorn.Engine)
+	for _, e := range engines.Items {
+		volumeNameToEngines[e.Labels[types.LonghornLabelVolume]] = append(volumeNameToEngines[e.Labels[types.LonghornLabelVolume]], e)
```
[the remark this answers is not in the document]
PhanLe1010: If the label doesn't exist, the corresponding engine will be grouped into one group map[""]->[<all engines without labels>]. Golang will not panic. Then we don't care about these invalid engines when looking up the data for backup CR below.

So, I think it is ok. What do you think?

upgrade/v122to123/upgrade.go
PhanLe1010: Could you give more details about how we can delay?

On the other hand, volume CR listing should be fast and doesn't affect much to overall time.

innobead: Have a function closure to get volume-map (init once) outside the backup loop, then call it the default clause?

innobead (approved):

upgrade/v122to123/upgrade.go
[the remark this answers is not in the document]
innobead: It's implicit but for sure it will work. Thanks for the explanation.

PR #1337, opened 2022-05-18
Handle Setting CRs during installation, upgrade and update

derekbit:
1. Record the resourceVersion of the ConfigMap, longhorn-default-setting,
   in the Setting CR if the default setting value is from ConfigMap.

   During upgrade or longhorn-manager restart, if the resourceVersion of the
   ConfigMap is different than the one in the Setting CR, update the Setting CR
   using the latest value in the ConfigMap.

2. Implementation the Setting CRs update in the configmap controller. If users do
   "helm upgrade ....", but the longhorn-manager does not restart,  configmap controller
   will update the Setting CRs.

[Longhorn 2570](https://github.com/longhorn/longhorn/issues/2570)


derekbit:
Please check the commit in https://github.com/longhorn/longhorn/pull/3998 together, thanks.

deploy/install/01-prerequisite/04-default-setting.yaml
```diff
@@ -5,42 +5,3 @@ metadata:
   namespace: longhorn-system
 data:
   default-setting.yaml: |-
-    backup-target:
```
innobead, line 8: Can we comment them out instead of removing them for reference purpose? 

note: it's fine we have empty default-setting.yaml in the deploy.yaml in longhorn/longhorn, because that is automatically generated instead.

derekbit: Addressed.

controller/kubernetes_configmap_controller.go
```diff
@@ -267,3 +284,18 @@ func isLonghornStorageClass(obj interface{}) bool {
 
 	return sc.Name == types.DefaultStorageClassName && sc.Provisioner == types.LonghornDriverName
 }
+
+func (kc *KubernetesConfigMapController) UpdateBuiltInSettingsWithCustomizedValues(defaultSettingCM *v1.ConfigMap) error {
```
innobead, line 288: This function is actually the same as `types.UpdateBuiltInSettingsWithCustomizedValues`.

controller/kubernetes_configmap_controller.go
```diff
@@ -147,56 +148,72 @@ func (kc *KubernetesConfigMapController) syncHandler(key string) (err error) {
...
+			return err
+		}
+
+		if err := kc.UpdateBuiltInSettingsWithCustomizedValues(defaultSettingCM); err != nil {
+			return fmt.Errorf("failed to update built-in settings with customized values: %v", err)
+		}
+
+		if err := kc.ds.InitOrUpdateSettings(); err != nil {
+			return err
+		}
```
innobead, line 215: Just call DataStore::UpdateSettings()

innobead (changes requested):

controller/kubernetes_configmap_controller.go
derekbit: Addressed. Thank you.

controller/kubernetes_configmap_controller.go
derekbit: Addressed. Thank you

app/daemon.go
```diff
@@ -120,7 +120,7 @@ func startManager(c *cli.Context) error {
 			return fmt.Errorf("cannot find customized default setting file on %v: %v", defaultSettingPath, err)
 		}
 	}
-	if err := types.OverwriteBuiltInSettingsWithCustomizedValues(); err != nil {
+	if err := types.OverwriteBuiltInSettingsWithCustomizedValues(kubeconfigPath); err != nil {
```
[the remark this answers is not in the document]
derekbit: Addressed. Thank you.

innobead (commented):
LGTM, except one comment.

datastore/longhorn.go
```diff
@@ -41,29 +42,140 @@ var (
...
+	}
+
+	if err := overrideSettingDefinitions(overrideDefaultSettings); err != nil {
+		return err
+	}
+
+	return s.updateDefaultSettings()
+}
+
+func overrideSettingDefinitions(defaultSettings map[types.SettingName]string) error {
```
innobead, line 67: nit: Probably it can be moved to types/settings.go.

innobead (commented):
After this change, do we still need to mount default setting as a file in longhorn manager pod?

derekbit:
> After this change, do we still need to mount the default setting as a file in the longhorn manager pod?
> 
> If not, probably we need to update the manifest as well (longhorn/longhorn-manager and longhorn/longhorn).

No, we don't need the manifest mount anymore. I'll remove it from the manifest.

datastore/longhorn.go
```diff
@@ -41,29 +42,140 @@ var (
...
-				if _, err := s.CreateSetting(setting); err != nil && !apierrors.IsAlreadyExists(err) {
-					return err
-				}
-			} else {
+
+		// Won't create or update the required but empty string field
+		if definition.Required && definition.Default == "" {
+			continue
+		}
+
```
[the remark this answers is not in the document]
derekbit: We don't need this anymore. This is for the outdated design before commenting out the empty default settings.
Thank you.

controller/kubernetes_configmap_controller.go
```diff
@@ -147,56 +147,63 @@ func (kc *KubernetesConfigMapController) syncHandler(key string) (err error) {
...
+			}
+			return err
 		}
-		return err
-	}
 
-	kc.logger.Infof("Updated the default Longhorn StorageClass: %v", storageclass)
+		kc.logger.Infof("Updated the default Longhorn StorageClass: %v", storageclass)
+	case types.DefaultDefaultSettingConfigMapName:
+		if err := kc.ds.UpdateCustomizedSettings(map[types.SettingName]string{}); err != nil {
```
innobead, line 203: nit: passed setting map can be nil.

innobead (approved):
LGTM. Thanks @derekbit

innobead (approved):

PR #1338, opened 2022-05-18
grpc-proxy: refactor proxy client

c3y1huang:
https://github.com/longhorn/longhorn/issues/3967

### DIFF
--- a/controller/backup_controller.go
+++ b/controller/backup_controller.go
@@ -60,6 +60,8 @@ type BackupController struct {
 	ds *datastore.DataStore
 
 	cacheSyncs []cache.InformerSynced
+
+	engineClientProxyHandler *engineapi.EngineClientProxyHandler
 }
 
 func NewBackupController(
@@ -69,7 +71,7 @@ func NewBackupController(
 	kubeClient clientset.Interface,
 	controllerID string,
 	namespace string,
-) *BackupController {
+	engineClientProxyHandler *engineapi.EngineClientProxyHandler) *BackupController {
 	eventBroadcaster := record.NewBroadcaster()
 	eventBroadcaster.StartLogging(logrus.Infof)
 	// TODO: remove the wrapper when every clients have moved to use the clientset.
@@ -90,6 +92,8 @@ func NewBackupController(
 
 		kubeClient:    kubeClient,
 		eventRecorder: eventBroadcaster.NewRecorder(scheme, v1.EventSource{Component: "longhorn-backup-controller"}),
+
+		engineClientProxyHandler: engineClientProxyHandler,
 	}
 
 	ds.BackupInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
@@ -487,7 +491,7 @@ func (bc *BackupController) checkMonitor(backup *longhorn.Backup, volume *longho
 		return nil, err
 	}
 
-	engineClientProxy, backupTargetClient, err := getBackupTarget(bc.controllerID, backupTarget, bc.ds, bc.logger)
+	engineClientProxy, backupTargetClient, err := getBackupTarget(bc.controllerID, backupTarget, bc.engineClientProxyHandler, bc.ds, bc.logger)
 	if err != nil {
 		return nil, err
 	}
@@ -617,12 +621,12 @@ func (bc *BackupController) syncBackupStatusWithSnapshotCreationTimeAndVolumeSiz
 		return
 	}
 
-	engineClientProxy, err := engineapi.GetCompatibleClient(e, engineCliClient, bc.ds, bc.logger)
+	engineClientProxy, err := bc.engineClientProxyHandler.GetCompatibleClient(e, engineCliClient)
 	if err != nil {
 		bc.logger.Warnf("syncBackupStatusWithSnapshotCreationTimeAndVolumeSize: failed to get proxy: %v", err)
 		return
 	}
-	defer engineClientProxy.Close()
+	defer engineClientProxy.Done()
 
 	snap, err := engineClientProxy.SnapshotGet(e, backup.Spec.SnapshotName)
 	if err != nil {
--- a/controller/controller_manager.go
+++ b/controller/controller_manager.go
@@ -33,7 +33,7 @@ var (
 	longhornFinalizerKey = longhorn.SchemeGroupVersion.Group
 )
 
-func StartControllers(logger logrus.FieldLogger, stopCh chan struct{}, controllerID, serviceAccount, managerImage, kubeconfigPath, version string) (*datastore.DataStore, *WebsocketController, error) {
+func StartControllers(logger logrus.FieldLogger, stopCh chan struct{}, controllerID, serviceAccount, managerImage, kubeconfigPath, version string) (*datastore.DataStore, *engineapi.EngineClientProxyHandler, *WebsocketController, error) {
 	namespace := os.Getenv(types.EnvPodNamespace)
 	if namespace == "" {
 		logrus.Warnf("Cannot detect pod namespace, environment variable %v is missing, "+
@@ -43,7 +43,7 @@ func StartControllers(logger logrus.FieldLogger, stopCh chan struct{}, controlle
 
 	config, err := clientcmd.BuildConfigFromFlags("", kubeconfigPath)
 	if err != nil {
-		return nil, nil, errors.Wrap(err, "unable to get client config")
+		return nil, nil, nil, errors.Wrap(err, "unable to get client config")
 	}
 
 	config.Burst = 100
@@ -51,17 +51,17 @@ func StartControllers(logger logrus.FieldLogger, stopCh chan struct{}, controlle
 
 	kubeClient, err := clientset.NewForConfig(config)
 	if err != nil {
-		return nil, nil, errors.Wrap(err, "unable to get k8s client")
+		return nil, nil, nil, errors.Wrap(err, "unable to get k8s client")
 	}
 
 	lhClient, err := lhclientset.NewForConfig(config)
 	if err != nil {
-		return nil, nil, errors.Wrap(err, "unable to get clientset")
+		return nil, nil, nil, errors.Wrap(err, "unable to get clientset")
 	}
 
 	scheme := runtime.NewScheme()
 	if err := longhorn.SchemeBuilder.AddToScheme(scheme); err != nil {
-		return nil, nil, errors.Wrap(err, "unable to create scheme")
+		return nil, nil, nil, errors.Wrap(err, "unable to create scheme")
 	}
 
 	// TODO: there shouldn't be a need for a 30s resync period unless our code is buggy and our controllers aren't really
@@ -75,24 +75,26 @@ func StartControllers(logger logrus.FieldLogger, stopCh chan struct{}, controlle
 
 	ds := datastore.NewDataStore(lhInformerFactory, lhClient, kubeInformerFactory, kubeClient, namespace)
 
+	engineClientProxyHandler := engineapi.NewEngineClientProxyHandler(logger, ds, time.Second*10, stopCh)
+
 	rc := NewReplicaController(logger, ds, scheme, kubeClient, namespace, controllerID)
-	ec := NewEngineController(logger, ds, scheme, kubeClient, &engineapi.EngineCollection{}, namespace, controllerID)
-	vc := NewVolumeController(logger, ds, scheme, kubeClient, namespace, controllerID)
+	ec := NewEngineController(logger, ds, scheme, kubeClient, &engineapi.EngineCollection{}, namespace, controllerID, engineClientProxyHandler)
+	vc := NewVolumeController(logger, ds, scheme, kubeClient, namespace, controllerID, engineClientProxyHandler)
 	ic := NewEngineImageController(logger, ds, scheme, kubeClient, namespace, controllerID, serviceAccount)
 	nc := NewNodeController(logger, ds, scheme, kubeClient, namespace, controllerID)
 	ws := NewWebsocketController(logger, ds)
 	sc := NewSettingController(logger, ds, scheme, kubeClient, namespace, controllerID, version)
-	btc := NewBackupTargetController(logger, ds, scheme, kubeClient, controllerID, namespace)
-	bvc := NewBackupVolumeController(logger, ds, scheme, kubeClient, controllerID, namespace)
-	bc := NewBackupController(logger, ds, scheme, kubeClient, controllerID, namespace)
-	imc := NewInstanceManagerController(logger, ds, scheme, kubeClient, namespace, controllerID, serviceAccount)
+	btc := NewBackupTargetController(logger, ds, scheme, kubeClient, controllerID, namespace, engineClientProxyHandler)
+	bvc := NewBackupVolumeController(logger, ds, scheme, kubeClient, controllerID, namespace, engineClientProxyHandler)
+	bc := NewBackupController(logger, ds, scheme, kubeClient, controllerID, namespace, engineClientProxyHandler)
+	imc := NewInstanceManagerController(logger, ds, scheme, kubeClient, namespace, controllerID, serviceAccount, engineClientProxyHandler)
 	smc := NewShareManagerController(logger, ds, scheme, kubeClient, namespace, controllerID, serviceAccount)
 	bic := NewBackingImageController(logger, ds, scheme, kubeClient, namespace, controllerID, serviceAccount)
 	bimc := NewBackingImageManagerController(logger, ds, scheme, kubeClient, namespace, controllerID, serviceAccount)
-	bidsc := NewBackingImageDataSourceController(logger, ds, scheme, kubeClient, namespace, controllerID, serviceAccount)
+	bidsc := NewBackingImageDataSourceController(logger, ds, scheme, kubeClient, namespace, controllerID, serviceAccount, engineClientProxyHandler)
 	rjc := NewRecurringJobController(logger, ds, scheme, kubeClient, namespace, controllerID, serviceAccount, managerImage)
 	oc := NewOrphanController(logger, ds, scheme, kubeClient, controllerID, namespace)
-	snapc := NewSnapshotController(logger, ds, scheme, kubeClient, namespace, controllerID, &engineapi.EngineCollection{})
+	snapc := NewSnapshotController(logger, ds, scheme, kubeClient, namespace, controllerID, &engineapi.EngineCollection{}, engineClientProxyHandler)
 	kpvc := NewKubernetesPVController(logger, ds, scheme, kubeClient, controllerID)
 	knc := NewKubernetesNodeController(logger, ds, scheme, kubeClient, controllerID)
 	kpc := NewKubernetesPodController(logger, ds, scheme, kubeClient, controllerID)
@@ -102,8 +104,9 @@ func StartControllers(logger logrus.FieldLogger, stopCh chan struct{}, controlle
 	go kubeInformerFactory.Start(stopCh)
 	go lhInformerFactory.Start(stopCh)
 	if !ds.Sync(stopCh) {
-		return nil, nil, fmt.Errorf("datastore cache sync up failed")
+		return nil, nil, nil, fmt.Errorf("datastore cache sync up failed")
 	}
+	go engineClientProxyHandler.Run(stopCh)
 	go rc.Run(Workers, stopCh)
 	go ec.Run(Workers, stopCh)
 	go vc.Run(Workers, stopCh)
@@ -129,7 +132,7 @@ func StartControllers(logger logrus.FieldLogger, stopCh chan struct{}, controlle
 	go kcfmc.Run(Workers, stopCh)
 	go ksc.Run(Workers, stopCh)
 
-	return ds, ws, nil
+	return ds, engineClientProxyHandler, ws, nil
 }
 
 func ParseResourceRequirement(val string) (*corev1.ResourceRequirements, error) {
--- a/engineapi/engine.go
+++ b/engineapi/engine.go
@@ -250,7 +250,15 @@ func (e *EngineBinary) ReplicaRebuildVerify(engine *longhorn.Engine, url string)
 	return nil
 }
 
-// Close engine proxy client connection.
+func (e *EngineBinary) getEngineClientProxy() engineClientProxy {
+	panic(ErrNotImplement)
+}
+
+func (e *EngineBinary) use() {
+	panic(ErrNotImplement)
+}
+
+// Done decrements the number of an engine proxy client concurrent use count.
 // Do not panic this method because this is could be called by the fallback client.
-func (e *EngineBinary) Close() {
+func (e *EngineBinary) Done() {
 }
--- a/engineapi/proxy.go
+++ b/engineapi/proxy.go
@@ -1,41 +1,305 @@
 package engineapi
 
 import (
+	"fmt"
+	"sync"
+	"time"
+
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"golang.org/x/net/context"
+	"google.golang.org/grpc/connectivity"
+
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/wait"
 
 	imclient "github.com/longhorn/longhorn-instance-manager/pkg/client"
+	imtypes "github.com/longhorn/longhorn-instance-manager/pkg/types"
 	imutil "github.com/longhorn/longhorn-instance-manager/pkg/util"
 
 	"github.com/longhorn/longhorn-manager/datastore"
 	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
+	"github.com/longhorn/longhorn-manager/util"
 )
 
-func getLoggerForEngineProxyClient(logger logrus.FieldLogger, im *longhorn.InstanceManager) *logrus.Entry {
+const (
+	MaxProxyClientConcurrentInUse = 1000
+	DefaultProxyClientPerImage    = 3
+	MaxProxyClientPerImage        = 3 * DefaultProxyClientPerImage
+	ProxyPollInterval             = 5 * time.Second
+	ProxyClientMaxRetry           = 10
+	ProxyGCMaxRetry               = 3
+)
+
+type EngineClientProxyHandler struct {
+	proxyPools map[string]*engineClientProxyPool
+
+	logger logrus.FieldLogger
+	ds     *datastore.DataStore
+
+	stopCh chan struct{}
+
+	lock *sync.RWMutex
+
+	resyncPeriod time.Duration
+}
+
+type engineClientProxyPool struct {
+	clients map[string]EngineClientProxy
+
+	instanceManagerName string
+
+	lock *sync.RWMutex
+}
+
+func NewEngineClientProxyHandler(logger logrus.FieldLogger, ds *datastore.DataStore, resyncPeriod time.Duration, stopCh chan struct{}) *EngineClientProxyHandler {
+	return &EngineClientProxyHandler{
+		proxyPools:   make(map[string]*engineClientProxyPool),
+		logger:       logger,
+		ds:           ds,
+		stopCh:       stopCh,
+		lock:         &sync.RWMutex{},
+		resyncPeriod: resyncPeriod,
+	}
+}
+
+func getLoggerForProxyHandler(logger logrus.FieldLogger, im *longhorn.InstanceManager) *logrus.Entry {
 	return logger.WithFields(
 		logrus.Fields{
 			"instanceManager": im.Name,
 			"image":           im.Spec.Image,
-			"serverIP":        im.Status.IP,
 		},
 	)
 }
 
-func GetCompatibleClient(e *longhorn.Engine, fallBack interface{}, ds *datastore.DataStore, logger logrus.FieldLogger) (c EngineClientProxy, err error) {
+func (h *EngineClientProxyHandler) Run(stopCh <-chan struct{}) {
+	defer utilruntime.HandleCrash()
+
+	h.logger.Info("Starting Longhorn engine client proxy handler")
+	defer h.logger.Info("Shutting down Longhorn engine client proxy handler")
+
+	// Set the loop time the same as the gRPC client keepalive ping so there
+	// are enough time for the periodic ping and also able to recover some of
+	// the TransientFailure connections.
+	go wait.Until(h.resync, time.Second*10, stopCh)
+
+	<-stopCh
+}
+
+func (h *EngineClientProxyHandler) resync() {
+	handledPools := h.getHandledEngineClientProxyPools()
+	for instanceManagerImage, pool := range handledPools {
+		im, err := h.ds.GetInstanceManagerRO(pool.instanceManagerName)
+		if err != nil {
+			if !datastore.ErrorIsNotFound(err) {
+				h.logger.WithError(err).Warnf("Failed to get %v instance manager for engine client proxy", pool.instanceManagerName)
+				continue
+			}
+
+			h.logger.Warnf("%v instance manager no longer exist for engine client proxy", pool.instanceManagerName)
+			for _, client := range pool.clients {
+				client.getEngineClientProxy().setNextGarbageCollect()
+			}
+			continue
+		}
+
+		for _, client := range pool.clients {
+			engineClientProxy := client.getEngineClientProxy()
+			if !engineClientProxy.shouldResync() {
+				continue
+			}
+
+			log := engineClientProxy.getLogger()
+
+			// Try to stop and delete clients in garbage collect. We will
+			// force delete after some retry if the client cannot be stopped
+			// or the calls does not return by the caller.
+			h.garbageCollectEngineClientProxy(instanceManagerImage, client)
+
+			// Replace client in bad connection state with new client. Mark
+			// the bad client for garbage collect so it will not be called for
+			// any new requests.
+			if engineClientProxy.shouldGarbageCollect() {
+				engineClientProxy.setNextGarbageCollect()
+
+				if err := h.NewEngineClientProxy(im); err != nil {
+					log.WithError(err).Warn("Failed to replace with new engine client proxy")
+				}
+
+				engineClientProxy.setNextResync(h.resyncPeriod)
+
+				// gRPC is expecting the TransientFailure state to recover, but
+				// this is not guaranteed. So we replace one connection at a
+				// time leaving some time for other TransientFailure to recover.
+				break
+			}
+
+			// Start client connection for new client added to the pool.
+			if err := engineClientProxy.startNewClient(im, h.logger, h.ds, h.stopCh); err != nil {
+				log.WithError(err).Warn("Failed to start new engine client proxy")
+			}
+
+			engineClientProxy.setNextResync(h.resyncPeriod)
+		}
+
+		// When the client gets called, the calling count gets increased,
+		// and vice vesa decrease the calling count when the caller is done
+		// with the request.
+		// When exceeding the max concurrent use, we will create an extra
+		// client to handle more concurent use.
+		exceedMaxConnection := true
+		for _, client := range pool.clients {
+			engineClientProxy := client.getEngineClientProxy()
+			if engineClientProxy.getInUse() >= MaxProxyClientConcurrentInUse {
+				continue
+			}
+
+			exceedMaxConnection = false
+			break
+		}
+		if len(pool.clients) != 0 && exceedMaxConnection {
+			log := getLoggerForProxyHandler(h.logger, im)
+			log.Debugf("Exceed engine client proxy max concurrent use limit:%v", MaxProxyClientConcurrentInUse)
+
+			if err := h.NewEngineClientProxy(im); err != nil {
+				log.WithError(err).Warn("Failed to add more engine client proxy")
+			}
+		}
+	}
+}
+
+func (h *EngineClientProxyHandler) garbageCollectEngineClientProxy(instanceManagerImage string, client EngineClientProxy) {
+	engineClientProxy := client.getEngineClientProxy()
+
+	log := engineClientProxy.getLogger()
+
+	nextGC := engineClientProxy.getNextGarbageCollect()
+	now := time.Now()
+	if nextGC == nil || now.Before(*nextGC) {
+		return
+	}
+
+	retryFailedMsg := ""
+	if engineClientProxy.getInUse() != 0 {
+		retryFailedMsg = fmt.Sprintf("Cannot stop engine client proxy with %v still in use", engineClientProxy.getInUse())
+		if engineClientProxy.getGarbageCollectRetry() < ProxyGCMaxRetry {
+			log.Warnf(retryFailedMsg)
+			engineClientProxy.setNextGarbageCollect()
+			return
+		}
+	}
+
+	log.Debug("Stopping engine client proxy")
+	if err := engineClientProxy.stop(); err != nil {
+		retryFailedMsg = "Failed to stop engine client proxy"
+		if engineClientProxy.getGarbageCollectRetry() < ProxyGCMaxRetry {
+			log.WithError(err).Warnf(retryFailedMsg)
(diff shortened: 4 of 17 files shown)


c3y1huang:
Test result [#1283](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/1283/)

engineapi/proxy.go
```diff
@@ -19,77 +23,330 @@ import (
...
+	<-stopCh
+}
+
+func getLoggerForProxyManager(logger logrus.FieldLogger, imName string) *logrus.Entry {
+	return logger.WithField("instanceManager", imName)
+}
+
+func (m *EngineClientProxyManager) resync() {
+	for {
+		for imName, handle := range m.handler.Handle {
```
shuo-wu, line 75: - I don't think you can directly operate `handler.Handle` without lock
- What's the key? imImage or imName?

engineapi/proxy.go
```diff
@@ -19,77 +23,330 @@ import (
...
+				}
+
+				m.gc(imName, clientIndex)
+
+				if client.ShouldCreate() {
+					cLog.Debug("Creating proxy client")
+
+					isInstanceManagerRunning := im.Status.CurrentState == longhorn.InstanceManagerStateRunning
+					if !isInstanceManagerRunning {
+						cLog.Warnf("Cannot create new gRPC client while instance manager in %v state", im.Status.CurrentState)
```
shuo-wu, line 114:
```
continue
```

engineapi/proxy.go
```diff
@@ -19,77 +23,330 @@ import (
...
+	go wait.Until(m.resync, time.Second, stopCh)
+
+	<-stopCh
+}
+
+func getLoggerForProxyManager(logger logrus.FieldLogger, imName string) *logrus.Entry {
+	return logger.WithField("instanceManager", imName)
+}
+
+func (m *EngineClientProxyManager) resync() {
```
shuo-wu, line 73: One quick question, is this layer/abstraction necessary?
It seems that this resync can be done by the ProxyHandler as well.

engineapi/proxy.go
```diff
@@ -19,77 +23,330 @@ import (
...
+				} else {
+					log.Debugf("Add proxy client-%v to handle concurrent use", newIndex)
+					m.handler.NewHandleClient(im, newIndex)
+				}
+			}
+		}
+	}
+}
+
+func (m *EngineClientProxyManager) getNewClientIndex(clients map[int]EngineClientProxy) (int, error) {
```
shuo-wu, line 154: This looks weird... Why not directly ask the handler to launch a new client if the limit is not exceeded?

engineapi/proxy.go
```diff
@@ -172,27 +438,153 @@ type Proxy struct {
...
+func (p *Proxy) SetInUse(increment bool) {
+	p.inUseLock.Lock()
+	defer p.inUseLock.Unlock()
+
+	if increment {
+		p.inUse++
+		return
+	}
+
+	p.inUse--
```
shuo-wu, line 567:
```
if p.inUse > 0 {
    p.inUse--
}
```

shuo-wu (changes requested):
The new design looks complicated.
- It seems that you are using 2 levels of the limitation. Can you use one level limitation instead?
    1. For each IM, the client count should not exceed a number. (Maybe verbose)
    2. For each client, the caller/user count should not exceed a number. 
- Abstraction `EngineClientProxyManager` may be redundant.
- `handler.Handle` is not protected by the lock correctly

I will re-check the PR after the modification

engineapi/engine.go
```diff
@@ -268,6 +255,59 @@ func (e *EngineBinary) ReplicaRebuildVerify(engine *longhorn.Engine, url string)
...
+func (e *EngineBinary) GetLogger() logrus.FieldLogger {
+	return &logrus.Entry{}
+}
+
+func (e *EngineBinary) IsGRPC() bool {
+	return false
+}
+
+func (e *EngineBinary) GetConnectionState() connectivity.State {
+	return 0
```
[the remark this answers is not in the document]
c3y1huang: Updated.

engineapi/proxy.go
c3y1huang: > * I don't think you can directly operate `handler.Handle` without lock

Fixed.

> * What's the key? imImage or imName?

Fixed.

engineapi/proxy.go
c3y1huang: Fixed.

engineapi/proxy.go
c3y1huang: Fixed.

engineapi/proxy.go
c3y1huang: I want to ensure the client doesn't get overwritten when waiting for GC. So the new client will always use a new index while the old waits for GC.

engineapi/proxy.go
c3y1huang: Fixed.

controller/controller_manager.go
```diff
@@ -75,7 +75,7 @@ func StartControllers(logger logrus.FieldLogger, stopCh chan struct{}, controlle
 
 	ds := datastore.NewDataStore(lhInformerFactory, lhClient, kubeInformerFactory, kubeClient, namespace)
 
-	proxyHandler := engineapi.NewEngineClientProxyHandler(logger, ds)
+	proxyHandler := engineapi.NewEngineClientProxyHandler(logger, ds, time.Second*5, stopCh)
```
innobead, line 78: Can we rename to `engineClientProxyHandler`? more specific.

engineapi/engine.go
```diff
@@ -268,6 +259,59 @@ func (e *EngineBinary) ReplicaRebuildVerify(engine *longhorn.Engine, url string)
...
+func (e *EngineBinary) IsGRPC() bool {
+	return false
+}
+
+func (e *EngineBinary) GetConnectionState() connectivity.State {
+	return connectivity.Idle
+}
+
+func (e *EngineBinary) Start(int, *longhorn.InstanceManager, logrus.FieldLogger, *datastore.DataStore, chan struct{}) error {
+	return errors.Errorf(ErrNotImplement)
```
innobead, line 275: Can we just panic with "not implemented" message? because this is actually on purpose panic and callers should not call them. 

The same as others.

innobead (changes requested):

engineapi/proxy.go
```diff
@@ -19,77 +23,296 @@ import (
...
 
 	logger logrus.FieldLogger
+	ds     *datastore.DataStore
 
-	ds *datastore.DataStore
+	stopCh chan struct{}
 
 	lock *sync.RWMutex
+
+	clock        clock.Clock
```
innobead, line 44: What's clock? Suggest having a meaningful name.

engineapi/proxy.go
```diff
@@ -19,77 +23,296 @@ import (
...
+	resyncPeriod time.Duration
+}
+
+type EngineClientProxys struct {
```
innobead, line 48: Do we need to expose this structure or just in-package use? 

This seems like client pool, can we make a meaningful name instead? like `EngineClientProxyPool` or `EngineClientProxyCache`?

engineapi/proxy.go
```diff
@@ -19,77 +23,296 @@ import (
...
+	Clients map[int]EngineClientProxy
```
innobead, line 49: Do we make `Clients` public, or it's actually just used in package?

engineapi/proxy.go
```diff
@@ -19,77 +23,296 @@ import (
...
-
-	IsGRPC() bool
-	Start(*longhorn.InstanceManager, logrus.FieldLogger, *datastore.DataStore) error
-	Stop(string) error
-	Ping() error
-}
-
 type EngineClientProxyHandler struct {
-	Clients map[string]EngineClientProxy
+	handle map[string]EngineClientProxys
```
innobead, line 35: Suggest have a meaningful name like `clientPools` or `clientCaches` given my another suggestion for `EngineClientProxys` renaming`?

engineapi/proxy.go
```diff
@@ -19,77 +23,296 @@ import (
...
+			"instanceManager": im.Name,
+			"image":           im.Spec.Image,
 		},
 	)
 }
 
+func (h *EngineClientProxyHandler) Run(stopCh <-chan struct{}) {
+	defer utilruntime.HandleCrash()
+
+	h.logger.Info("Start Longhorn proxy manager")
```
innobead, line 78: "Starting engine client proxy handler"? Suggest having consistent names corresponding to the method, structure, etc.

engineapi/proxy.go
```diff
@@ -19,77 +23,296 @@ import (
...
+	defer h.logger.Info("Shutting down Longhorn proxy manager")
+
+	go wait.Until(h.resync, time.Second, stopCh)
+
+	<-stopCh
+}
+
+func (h *EngineClientProxyHandler) resync() {
+	for {
+		imImageHandled := h.GetHandledInstanceManagerImages()
```
innobead, line 88: Please use `handledImImages` instead.

engineapi/proxy.go
```diff
@@ -19,77 +23,296 @@ import (
...
+		if index <= maxIndex {
+			continue
+		}
+
+		maxIndex = index
+	}
+	return maxIndex + 1, nil
+}
+
+func (h *EngineClientProxyHandler) GetHandledInstanceManagerImages() map[string]EngineClientProxys {
```
innobead, line 223: The naming should be aligned with it's returned value. ex: `GetEngineClientProxies` instead. The same as other method names.

innobead (changes requested):

engineapi/proxy.go
```diff
@@ -19,77 +23,296 @@ import (
...
+			client.SetNextGC(h.clock)
+			return
+		}
+	}
+
+	if msg != "" {
+		log.Warn("Exceed max retry limit, force delete proxy client")
+	}
+
+	h.lock.Lock()
```
innobead, line 197: I think this lock should be for the whole scope of this method instead of only protecting the map.

engineapi/proxy.go
```diff
@@ -19,77 +23,296 @@ import (
...
+}
+
+func (h *EngineClientProxyHandler) GetHandledInstanceManagerImages() map[string]EngineClientProxys {
+	h.lock.Lock()
+	defer h.lock.Unlock()
+	return h.handle
+}
+
+func (h *EngineClientProxyHandler) GetHandledInstanceManagerImage(name string) (EngineClientProxys, bool) {
+	h.lock.Lock()
```
innobead, line 230: rlock.

engineapi/proxy.go
```diff
@@ -19,77 +23,296 @@ import (
...
+	defer h.lock.Unlock()
+
+	handle, exist := h.handle[name]
+	return handle, exist
+}
+
+func (h *EngineClientProxyHandler) GetHandleClient(instanceManagerImage string, clientIndex int) (EngineClientProxy, bool) {
+	h.lock.Lock()
```
innobead, line 238: rlock.

engineapi/proxy.go
```diff
@@ -100,69 +323,78 @@ func (h *EngineClientProxyHandler) GetClient(im *longhorn.InstanceManager) (c En
...
-	client, exist := h.Clients[im.Spec.Image]
-	h.lock.Unlock()
-	if exist {
-		err := client.Ping()
-		if err == nil {
-			return client, nil
+	var candidate EngineClientProxy
+	failCount := 0
+	for {
+		if failCount > ProxyClientMaxRetry {
```
innobead, line 336: This is tricky. I know you want to make GetClient still work as blocking to ensure it can get the correct in a best effort.

engineapi/proxy.go
```diff
@@ -100,69 +323,78 @@ func (h *EngineClientProxyHandler) GetClient(im *longhorn.InstanceManager) (c En
...
-			log.Debugf("Proxy client already created")
-			return existClient, nil
+		failCount++
+		if err := checkClient(candidate); err != nil {
+			clientLog := candidate.GetLogger()
+			clientLog.WithError(err).Debugf("Retry(%v) to get client", failCount)
+		} else {
+			candidate.SetInUse(true)
+			return candidate, nil
 		}
```
innobead, line 369: Can we create another method to handle to find out a less used client?

innobead (changes requested):

engineapi/proxy.go
```diff
@@ -172,27 +404,164 @@ type Proxy struct {
...
+
+	SetNextResync(clock.Clock, time.Duration)
+	ShouldResync(clock.Clock) bool
+
+	SetNextGC(clock.Clock)
+	GetNextGC() *time.Time
+	GetGCRetry() int
+
+	SetInUse(bool)
+	GetInUse() int
```
innobead, line 443: Suggest not adding these internal methods to the interface, so we don't need to make EngineBinary implemented these as well.

innobead (changes requested):

innobead (changes requested):

engineapi/proxy.go
```diff
@@ -172,27 +404,164 @@ type Proxy struct {
...
+	defer p.lock.Unlock()
+
+	return p.nextGC == nil && p.grpcClient == nil
+}
+
+func (p *Proxy) ShouldDelete() bool {
+	p.lock.Lock()
+	defer p.lock.Unlock()
+
+	return p.nextGC == nil && p.grpcClient != nil && (p.GetConnectionState() == connectivity.Shutdown || p.GetConnectionState() == connectivity.TransientFailure)
```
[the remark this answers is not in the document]
c3y1huang: We don't need to delete it again if it's marked for GC already.

controller/controller_manager.go
c3y1huang: Updated.

engineapi/engine.go
c3y1huang: Updated.

engineapi/proxy.go

engineapi/proxy.go

engineapi/proxy.go

engineapi/proxy.go

engineapi/proxy.go

engineapi/proxy.go

engineapi/proxy.go

engineapi/proxy.go
```diff
@@ -19,77 +23,296 @@ import (
...
+	<-stopCh
+}
+
+func (h *EngineClientProxyHandler) resync() {
+	for {
+		imImageHandled := h.GetHandledInstanceManagerImages()
+		for instanceManagerImage, handle := range imImageHandled {
+			im, err := h.ds.GetInstanceManagerRO(handle.instanceManagerName)
+			if err != nil {
+				h.logger.WithError(err).Warnf("Failed to get %v instance manager for proxy", handle.instanceManagerName)
```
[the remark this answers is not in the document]
c3y1huang: Addressed.

engineapi/proxy.go
```diff
@@ -19,77 +23,296 @@ import (
...
+				continue
+			}
+
+			log := getLoggerForProxyHandler(h.logger, im)
```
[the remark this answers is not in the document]

engineapi/proxy.go
```diff
@@ -19,77 +23,296 @@ import (
...
+
+			for clientIndex, client := range handle.Clients {
+				if !client.ShouldResync(h.clock) {
+					continue
+				}
+
+				cLog := client.GetLogger()
```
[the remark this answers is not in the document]

engineapi/proxy.go
```diff
@@ -19,77 +23,296 @@ import (
...
+			continue
+		}
+
+		maxIndex = index
+	}
+	return maxIndex + 1, nil
+}
+
+func (h *EngineClientProxyHandler) GetHandledInstanceManagerImages() map[string]EngineClientProxys {
+	h.lock.Lock()
```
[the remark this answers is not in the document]
c3y1huang: Fixed.

engineapi/proxy.go
```diff
@@ -172,27 +404,164 @@ type Proxy struct {
...
+
+func (p *Proxy) GetNextGC() *time.Time {
+	p.gcLock.Lock()
+	defer p.gcLock.Unlock()
+
+	return p.nextGC
+}
+
+func (p *Proxy) GetGCRetry() int {
+	p.gcLock.Lock()
```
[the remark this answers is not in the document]

engineapi/proxy.go
```diff
@@ -19,77 +23,296 @@ import (
...
+					cLog.Debug("Deleting proxy client")
+					client.SetNextGC(h.clock)
+
+					newIndex, err := h.getNewClientIndex(handle.Clients)
+					if err != nil {
+						cLog.WithError(err).Debug("Cannot replace proxy client")
+					} else {
+						cLog.Debug("Replacing with new proxy client")
+						h.NewHandleClient(im, newIndex)
+					}
```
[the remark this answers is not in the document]
c3y1huang: Updated.

engineapi/proxy.go
```diff
@@ -19,77 +23,296 @@ import (
...
+			log.Warnf(msg)
+			client.SetNextGC(h.clock)
+			return
+		}
+	}
+
+	log.Debug("Stopping proxy client")
+	if err := client.Stop(); err != nil {
+		msg = "Failed to stop proxy client"
+		if client.GetGCRetry() < ProxyGCMaxRetry {
```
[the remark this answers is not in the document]
c3y1huang: I prefer not to delete immediately to allow clients to complete the requests. 
We probably need time and more tests to investigate what happens if allowing immediate force delete.

engineapi/proxy.go
c3y1huang: We have Rlock protecting others in the method.

engineapi/proxy.go
```diff
@@ -19,77 +23,296 @@ import (
...
+	}
+
+	h.lock.Lock()
+	defer h.lock.Unlock()
+
+	delete(h.handle[instanceManagerImage].Clients, clientIndex)
+	log.Debug("Deleted proxy client")
+}
+
+func (h *EngineClientProxyHandler) getNewClientIndex(clients map[int]EngineClientProxy) (int, error) {
```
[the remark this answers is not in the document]
c3y1huang: Fixed.

engineapi/proxy.go
c3y1huang: Fixed.

engineapi/proxy.go
c3y1huang: Fixed.

engineapi/proxy.go
c3y1huang: Addressed.

engineapi/proxy.go
```diff
@@ -172,27 +404,164 @@ type Proxy struct {
...
+
+func (p *Proxy) ShouldCreate(im *longhorn.InstanceManager) bool {
+	p.lock.Lock()
+	defer p.lock.Unlock()
+
+	return p.nextGC == nil && p.grpcClient == nil
+}
+
+func (p *Proxy) ShouldDelete() bool {
+	p.lock.Lock()
```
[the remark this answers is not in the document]
c3y1huang: Fixed.

controller/backup_controller.go
```diff
@@ -626,6 +626,7 @@ func (bc *BackupController) syncBackupStatusWithSnapshotCreationTimeAndVolumeSiz
 		bc.logger.Warnf("syncBackupStatusWithSnapshotCreationTimeAndVolumeSize: failed to get proxy: %v", err)
 		return
 	}
+	defer engineClientProxy.SetInUse(false)
```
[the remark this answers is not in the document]
c3y1huang: Addressed.

engineapi/proxy.go
c3y1huang: Addressed.

engineapi/proxy.go
c3y1huang: Use UUID instead.

engineapi/proxy.go
```diff
@@ -1,168 +1,421 @@
...
+
+	h.logger.Info("Starting Longhorn engine client proxy handler")
+	defer h.logger.Info("Shutting down Longhorn engine client proxy handler")
+
+	go wait.Until(h.resync, time.Second, stopCh)
+
+	<-stopCh
+}
+
+func (h *EngineClientProxyHandler) resync() {
```
innobead, line 84: There are 4 parts in resync. Because there is no any private methods to compose this, could you add a comment for each part? This will help for the following debugging. 

1. Garbage collect or delete outdated client
2. Request to delete outdated client
3. Request to create a new client
4. Request to create an extra client when number of client connections not enough

engineapi/proxy.go
```diff
@@ -1,168 +1,421 @@
...
+	log := client.getLogger()
+
+	nextGC := client.getNextGC()
+	now := time.Now()
+	if nextGC == nil || now.Before(*nextGC) {
+		return
+	}
+
+	msg := ""
+	if client.getInCall() != 0 {
```
innobead, line 177: note: We need to ensure if the client has been marked GC, and after that, it should be not used anymore.

engineapi/proxy.go
```diff
@@ -1,168 +1,421 @@
...
+	if err := client.stop(); err != nil {
+		msg = "Failed to stop engine client proxy"
+		if client.getGCRetry() < ProxyGCMaxRetry {
+			log.WithError(err).Warnf(msg)
+			client.setNextGC()
+			return
+		}
+	}
+
+	if msg != "" {
```
innobead, line 196: Can we use a meaningful variable to determine the max retry case instead of a general `msg`? ex: `retryFailed'

engineapi/proxy.go
```diff
@@ -1,168 +1,421 @@
...
+		log.Warn("Exceed max retry limit, force delete engine client proxy")
+	}
+
+	h.lock.Lock()
```
innobead, line 200: note: Need to ensure all GC-marked clients will not be able to use by other requests. Otherwise, lock here is still not atomic, because a stopping client still has a chance to use.

engineapi/proxy.go
```diff
@@ -1,168 +1,421 @@
...
+
+	client, exist := h.proxyPools[instanceManagerImage].clients[uuid]
+	return client.getEngineClientProxy(), exist
+}
+
+func (h *EngineClientProxyHandler) NewEngineClientProxy(im *longhorn.InstanceManager) error {
+	imImage := im.Spec.Image
+	log := getLoggerForProxyHandler(h.logger, im)
+
+	_, handlerExist := h.getHandledEngineClientProxyPoolByImage(imImage)
```
innobead, line 247: Please use consistent variable naming here? probably `poolExist`?

engineapi/proxy.go
```diff
@@ -1,168 +1,421 @@
...
+	if !handlerExist {
+		log.Debug("Initializing engine client proxy handler")
```
innobead, line 249: "Initializing engine client proxy pool".. (please recheck other places)

engineapi/proxy.go
```diff
@@ -1,168 +1,421 @@
...
+		h.lock.Lock()
```
innobead, line 250: Can we move the lock above `h.getHandledEngineClientProxyPoolByImage(imImage)` to ensure the atomic lock here?

engineapi/proxy.go
```diff
@@ -1,168 +1,421 @@
...
+		h.lock.Unlock()
+	}
+
+	pool, _ := h.getHandledEngineClientProxyPoolByImage(imImage)
+	clientUUID, err := h.getNewClientUUID(pool.clients)
+	if err != nil {
+		return err
+	}
+
+	h.lock.Lock()
```
innobead, line 264: Maybe you can just make one lock for this method, because creating a client proxy needs atomic. Granular lock controlling here seems not beneficial a lot.

innobead (changes requested):

engineapi/proxy.go
```diff
@@ -172,62 +425,241 @@ type Proxy struct {
...
+	gcLock  sync.RWMutex
+}
+
+type EngineClientProxy interface {
+	EngineClient
+	BackupTargetBinaryClient
+
+	getEngineClientProxy() engineClientProxy
+
+	calling()
```
innobead, line 447: Suggest renaming to `use`.

engineapi/proxy.go
```diff
@@ -172,62 +425,241 @@ type Proxy struct {
...
+	Done()
+}
+
+func (p *Proxy) getEngineClientProxy() engineClientProxy {
+	p.lock.RLock()
+	defer p.lock.RUnlock()
+
+	return p
+}
```
innobead, line 456: Why we need this method? actually this just return the same object?

engineapi/proxy.go
```diff
@@ -172,62 +425,241 @@ type Proxy struct {
...
+	shouldDelete() bool
+
+	setNextResync(time.Duration)
+	shouldResync() bool
+
+	setNextGC()
+	getNextGC() *time.Time
+	getGCRetry() int
+
+	setInCall(bool)
```
innobead, line 487: Suggest renaming to `incrementUse`.

engineapi/proxy.go
```diff
@@ -172,62 +425,241 @@ type Proxy struct {
...
+	getInCall() int
```
innobead, line 488: Suggest renaming to `decrementUse`.

engineapi/proxy.go
```diff
@@ -1,168 +1,421 @@
...
+						log.WithError(err).Debug("Failed to replace with new engine client proxy")
+					}
+
+					engineClientProxy.setNextResync(h.resyncPeriod)
+					break
+				}
+
+				h.gc(instanceManagerImage, clientUUID)
+
+				if engineClientProxy.shouldCreate(im) {
```
innobead, line 128: Can we remove `shouldCreate`, but just move the logic to `start`?

engineapi/proxy.go
```diff
@@ -172,62 +425,241 @@ type Proxy struct {
...
+type engineClientProxy interface {
+	getLogger() logrus.FieldLogger
+
+	isGRPC() bool
+	getConnectionState() connectivity.State
+
+	start(*longhorn.InstanceManager, logrus.FieldLogger, *datastore.DataStore, chan struct{}) error
+	stop() error
+
+	shouldCreate(*longhorn.InstanceManager) bool
```
innobead, line 477: Probably we can remove this one as per the other comment I made

engineapi/proxy.go
```diff
@@ -172,62 +425,241 @@ type Proxy struct {
...
+	shouldDelete() bool
```
innobead, line 478: I feel you can make delete and gc consistent. You can use delete or either gc, but just choose one term, that will make the code understandable. WDYT?

innobead (changes requested):

engineapi/proxy.go
```diff
@@ -1,168 +1,421 @@
...
+				if !datastore.ErrorIsNotFound(err) {
+					h.logger.WithError(err).Warnf("Failed to get %v instance manager for engine client proxy", pool.instanceManagerName)
+					continue
+				}
+
+				h.logger.Warnf("%v instance manager no longer exist for engine client proxy", pool.instanceManagerName)
+				for _, client := range pool.clients {
+					engineClientProxy := client.getEngineClientProxy()
+					log := engineClientProxy.getLogger()
+					log.Debug("Cleanup engine client proxy")
```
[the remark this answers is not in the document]

engineapi/proxy.go
```diff
@@ -1,168 +1,421 @@
...
+			for clientUUID, client := range pool.clients {
+				engineClientProxy := client.getEngineClientProxy()
+				if !engineClientProxy.shouldResync() {
+					continue
+				}
+
+				log := engineClientProxy.getLogger()
+
+				if engineClientProxy.shouldDelete() {
+					log.Debug("Deleting engine client proxy")
```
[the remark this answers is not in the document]

engineapi/proxy.go
c3y1huang: Logic in `checkClient`.

engineapi/proxy.go
c3y1huang: https://github.com/longhorn/longhorn-manager/pull/1338#discussion_r880394409

engineapi/proxy.go

engineapi/proxy.go
c3y1huang: Moved.

engineapi/proxy.go
c3y1huang: Done.

engineapi/proxy.go
c3y1huang: Returning `engineClientProxy` interface.

engineapi/proxy.go
```diff
@@ -245,46 +689,45 @@ func (p *Proxy) Run() {
...
 	return imutil.GetURL(e.Status.StorageIP, e.Status.Port)
 }
 
-func (p *Proxy) Ping() (err error) {
-	return p.grpcClient.Ping()
-}
-
 func (p *Proxy) Name() string {
+	p.lock.RLock()
+	defer p.lock.RUnlock()
```
innobead, line 722: No need to lock here, also `InstanceManagerName` already publicly exported, so do we still need a Name method? or you can just make `instanceManagerName` private.

engineapi/proxy.go
```diff
@@ -172,62 +431,246 @@ type Proxy struct {
...
+	getNextGarbageCollect() *time.Time
+	getGarbageCollectRetry() int
+
+	incrementUse(bool)
+	getInUse() int
+}
+
+func (p *Proxy) getLogger() logrus.FieldLogger {
+	p.lock.RLock()
+	defer p.lock.RUnlock()
```
innobead, line 497: Why we need lock here? I feel this method not required, internal usage can just use logger attribute/property directly.

engineapi/proxy.go
```diff
@@ -172,62 +431,246 @@ type Proxy struct {
...
+	return p.logger
+}
+
+func (p *Proxy) isGRPC() bool {
+	return p.getGRPCClient() != nil
+}
+
+func (p *Proxy) getGRPCClient() *imclient.ProxyClient {
+	p.lock.RLock()
+	defer p.lock.RUnlock()
```
innobead, line 508: Why we need lock here? `p.grpcClient` will not be modified after the start, so probably we can just call p.grpcClient directly without another wrapper function.

engineapi/proxy.go
```diff
@@ -172,62 +431,246 @@ type Proxy struct {
...
+
+	getEngineClientProxy() engineClientProxy
+
+	use()
+	Done()
+}
+
+func (p *Proxy) getEngineClientProxy() engineClientProxy {
+	p.lock.RLock()
+	defer p.lock.RUnlock()
```
innobead, line 459: I don't think we need this lock as well.

innobead (changes requested):

engineapi/proxy.go
```diff
@@ -1,168 +1,429 @@
...
-		Clients: make(map[string]EngineClientProxy),
-		logger:  logger,
-		ds:      ds,
-		lock:    &sync.RWMutex{},
+		proxyPools:   make(map[string]engineClientProxyPool),
+		logger:       logger,
+		ds:           ds,
+		stopCh:       stopCh,
+		lock:         &sync.RWMutex{},
+		resyncPeriod: resyncPeriod,
```
shuo-wu, line 60: NIT: This actually means the resync period of a single proxy. Rename this to avoid misunderstanding.

engineapi/proxy.go
```diff
@@ -1,168 +1,429 @@
...
+					if err := h.NewEngineClientProxy(im); err != nil {
+						log.WithError(err).Debug("Failed to replace with new engine client proxy")
+					}
+
+					engineClientProxy.setNextResync(h.resyncPeriod)
+					break
+				}
+
+				// Request to create a new client
+				if err := engineClientProxy.start(im, h.logger, h.ds, h.stopCh); err != nil {
```
shuo-wu, line 131: This comment and the below `engineClientProxy.start` are confusing...
`engineClientProxy` blindly executing `start` really confuse me... it's better to rename this method. I was thinking this will be executed during init or connection retries...

engineapi/proxy.go
```diff
@@ -1,168 +1,429 @@
...
+
+	pool, exist := h.proxyPools[name]
+	return pool, exist
+}
+
+func (h *EngineClientProxyHandler) getEngineClientProxy(instanceManagerImage string, uuid string) (engineClientProxy, bool) {
+	h.lock.RLock()
+	defer h.lock.RUnlock()
+
+	client, exist := h.proxyPools[instanceManagerImage].clients[uuid]
```
shuo-wu, line 243: For this function, should it check if `h.proxyPools[instanceManagerImage] != nil` first?

engineapi/proxy.go
```diff
@@ -1,168 +1,429 @@
...
-	if err := client.Start(im, log, h.ds); err != nil {
-		return nil, err
+	clientUUID := engineClientProxy.getClientUUID()
+	if engineClientProxy.getNextGarbageCollect() != nil {
+		return errors.Errorf("engine client proxy(%v) is marked for GC", clientUUID)
 	}
 
-	log.Debug("Adding client to proxy handler")
-	h.Clients[im.Spec.Image] = client
+	if conn := engineClientProxy.getConnectionState(); conn != connectivity.Ready {
```
shuo-wu, line 413: Shouldn't `connectivity.Idle` be considered as available?

engineapi/proxy.go
shuo-wu: `GC-marked clients will not be able to use by other requests` should be guaranteed by the lock inside the client rather than `h.lock`

shuo-wu (changes requested):

engineapi/proxy.go
```diff
@@ -172,62 +425,256 @@ type Proxy struct {
...
+	log.WithField("connectionState", p.getConnectionState()).Debug("Deleting engine client proxy")
+
+	p.gcLock.Lock()
+
+	p.gcRetry++
+
+	newGC := time.Now().Add(imtypes.GRPCServiceTimeout)
+	p.nextGC = &newGC
+
+	p.gcLock.Unlock()
```
innobead, line 570: Please move this just next to `p.gcLock.Lock()` to ensure unlock correctly.

engineapi/proxy.go
```diff
@@ -245,46 +693,42 @@ func (p *Proxy) Run() {
...
 	return imutil.GetURL(e.Status.StorageIP, e.Status.Port)
 }
 
-func (p *Proxy) Ping() (err error) {
-	return p.grpcClient.Ping()
-}
-
 func (p *Proxy) Name() string {
-	return p.InstanceManagerName
+	panic(ErrNotImplement)
```
innobead, line 725: Do we still need Name method?

engineapi/proxy.go
```diff
@@ -172,62 +425,256 @@ type Proxy struct {
...
+func (p *Proxy) setNextResync(t time.Duration) {
+	nextResync := p.getNextResync()
+	now := time.Now()
+	if now.Before(nextResync) || now.Equal(nextResync) {
+		return
+	}
+
+	p.resyncLock.Lock()
+	p.nextResync = now.Add(t)
+	defer p.resyncLock.Unlock()
```
innobead, line 544: Suggest moving this unlock just next to `defer p.resyncLock.Unlock()`.

engineapi/proxy.go
```diff
@@ -172,62 +425,256 @@ type Proxy struct {
...
+
+func (p *Proxy) GetPoolID() string {
+	p.lock.RLock()
+	defer p.lock.RUnlock()
+
+	return p.InstanceManagerName
+}
+
+func (p *Proxy) getGRPCClient() *imclient.ProxyClient {
+	return p.grpcClient
```
innobead, line 514: `p.grpcClient` is *imclient.ProxyClient already, why we need an extra method wrapper?

innobead (changes requested):

c3y1huang:
Test result [#1345(coretest/upgrade)](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/1345/)

c3y1huang:
> * [#1344](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/1344/console) (full) !!!!

Investigating.

c3y1huang:
Pool:
- Full - https://ci.longhorn.io/job/private/job/longhorn-tests-regression/1353
- Full - https://ci.longhorn.io/job/private/job/longhorn-tests-regression/1354
- Core - https://ci.longhorn.io/job/private/job/longhorn-tests-regression/1356
- Core - https://ci.longhorn.io/job/private/job/longhorn-tests-regression/1359

[On-demand](https://github.com/longhorn/longhorn-manager/pull/1348):
Full - https://ci.longhorn.io/job/private/job/longhorn-tests-regression/1355/
Full - https://ci.longhorn.io/job/private/job/longhorn-tests-regression/1358/
Core - https://ci.longhorn.io/job/private/job/longhorn-tests-regression/1357/
Core - https://ci.longhorn.io/job/private/job/longhorn-tests-regression/1360/

PhanLe1010 (commented):
May I have a question? 

* Each Longhorn manager will create 1 `engineClientProxyHandler`
* `engineClientProxyHandler` has a `proxyPools` which is a map from IM image to all proxy clients to 1 single instance manager pod. This means that all grpc calls from this Longhorn manager pod will be routed through that single instance manager pod no matter the destination of the grpc calls. Will it be a problem for an HA distributed system when that single instance manager is not available? 
* In the previous design, we are distributing the grpc calls to multiple instance manager pods


Am I understanding the flow correctly here?

c3y1huang:
> * In the previous design, we are distributing the grpc calls to multiple instance manager pods
> 

Not sure if you are referring to the gRPC process communication? This is a new proxy server is to replace manager engine binary calls, Communication to the gRPC process server is still the same.

shuo-wu:
> Will it be a problem for an HA distributed system when that single instance manager is not available?

I think this should be fine. 

Assuming the IMs on a node are unavailable, for the corresponding longhorn manager pod (on the same node):
- First, the engine/replica instance starting/stopping is exactly the same as the previous version.
- Then for most of the engine calls, the longhorn manager pod has no way to handle them before or after the proxy since the IMs on this node are unavailable.
- As for some backup/file-related calls, the longhorn manager pods will directly talk with the replicas directly before introducing the proxy. But the communication will be done by the proxy server after introducing the proxy. I guess this part is the main cause of your question, right? But when diving into the previous implementations, we can find that the longhorn-manager pod has to talk with the engine before sending requests to replicas. In other words, if the IM or (engine IM) is unavailable, there is no way to handle this kind of calls before or after the proxy.

innobead:
I will keep this as draft after discussing it with @c3y1huang , and revisit this after https://github.com/longhorn/longhorn/issues/4038.

shuo-wu:
> * Before the proxy, when the instance manager on the node goes down, backup volume controller still can [read backup](https://github.com/longhorn/longhorn-manager/blob/79ce2cfde922a7e0966059f72b8aee5f05bff513/controller/backup_volume_controller.go#L342) and [delete backup](https://github.com/longhorn/longhorn-manager/blob/79ce2cfde922a7e0966059f72b8aee5f05bff513/controller/backup_controller.go#L255)
> * after the proxy, this is not the case.
> 
> I am not sure if there are other cases

Will the backup CRs be taken over by other nodes in this case? If No, this is probably one issue we need to resolve.

innobead:
> > 
> > I am not sure if there are other cases
> 
> Will the backup CRs be taken over by other nodes in this case? If No, this is probably one issue we need to resolve.

From the segregation point of view, this is the new criteria for any operations that should pass through the proxy only. 

What we should really expect is that backup-related CRs will be tackled at the end when the IM pod comes back or as @shuo-wu mentioned if they will be transferred to another manager to handle. IIRC, the transferring will only happen when the node is not ready or the manager pod on that node is unavailable.

PhanLe1010:
>. IIRC, the transferring will only happen when the node is not ready or the manager pod on that node is unavailable.

Yeah, I backup CR will not be transferred to other nodes because the manager pod is still running

innobead:
> > . IIRC, the transferring will only happen when the node is not ready or the manager pod on that node is unavailable.
> 
> Yeah, I backup CR will not be transferred to other nodes because the manager pod is still running

@shuo-wu will this be a concern for us?

c3y1huang:
Close this first as now we have no requirement for creating a proxy client pool.

PR #1339, opened 2022-05-18
Fix bug copying the pointer of the same for-loop variable

PhanLe1010:
longhorn/longhorn#3980


shuo-wu (approved):

shuo-wu:
Don't forget to backport to v1.2.5

PR #1342, opened 2022-05-20
Bump K8s, client-go, CSI sidecars, Golang version

PhanLe1010:
longhorn/longhorn#3891

* Bump K8s dependency library
From v1.18 to v1.23
* client-go
From v1.18 to v1.23
* CSI sidecars:
  * longhornio/csi-attacher:
    From v3.2.1 to v3.4.0
  * longhornio/csi-node-driver-registrar 
    From v2.3.0 to v2.5.0
* Golang version
From  v1.13 to v1.17


innobead (approved):
LGTM

innobead:
cc @yangchiu for CSI sidecars

joshimoo (changes requested):
Couple of API Refs should be updated to use V1 otherwise we would no longer be compatible with future versions >= 1.25

Double check the api deprecation guide, to see which allow for transitioning to V1 due to our increased min version.
https://kubernetes.io/docs/reference/using-api/deprecation-guide/#cronjob-v125

Some that come to mind:
- cronjob usage for the reccuring job controller.
- PDB usage for the pod distributions.

You can add these additions as additional commits on top of the existing ones.
So we have a better idea of what changed and why.

innobead:
Agree, if it's allowed to upgrade the version based on the updated libraries and also it will work for 1.18, then we should update them this time.

joshimoo:
Checked with @innobead for v1.3 we are planning to support [1.18, 1.24]
For v1.4 we can update the min version higher and see how to deal with PSP removal as well down the line.

@PhanLe1010 did we check the CR definitions to ensure that we didn't rely on any default behavior that is now removed?
https://kubernetes.io/docs/reference/using-api/deprecation-guide/#customresourcedefinition-v122

PhanLe1010:
Yeah, during the PR review meeting, we agreed that:
* Only bump the K8s library dependency 
* Not bumping the minimum supported Kubernetes version
* Longhorn v1.3.0 supports Kubernetes [v1.18, v1.21]
* Do not upgrade API versions 

> @PhanLe1010 did we check the CR definitions to ensure that we didn't rely on any default behavior that is now removed?
https://kubernetes.io/docs/reference/using-api/deprecation-guide/#customresourcedefinition-v122

Since we do not upgrade API versions, the CR definitions in our yaml should stay the same. I double-checked them. Thank you for the reminder

innobead (approved):

innobead:
@PhanLe1010 @c3y1huang https://github.com/kubernetes-csi/docs/pull/517

PR #1344, opened 2022-05-24
Protect SettingDefinitions by RWMutex

derekbit:
longhorn/longhorn#2570

### DIFF
--- a/datastore/longhorn.go
+++ b/datastore/longhorn.go
@@ -66,7 +66,7 @@ func (s *DataStore) UpdateCustomizedSettings(overrideDefaultSettings map[types.S
 
 func (s *DataStore) updateDefaultSettings() error {
 	for _, sName := range types.SettingNameList {
-		definition, ok := types.SettingDefinitions[sName]
+		definition, ok := types.GetSettingDefinition(sName)
 		if !ok {
 			return fmt.Errorf("BUG: setting %v is not defined", sName)
 		}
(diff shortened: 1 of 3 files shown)


datastore/longhorn.go
```diff
@@ -393,11 +393,13 @@ func (s *DataStore) ListSettings() (map[types.SettingName]*longhorn.Setting, err
 		// Cannot use cached object from lister
 		settingField := types.SettingName(itemRO.Name)
 		// Ignore the items that we don't recongize
-		if _, ok := types.SettingDefinitions[settingField]; ok {
+		if _, ok := types.GetSettingDefinition(settingField); ok {
 			itemMap[settingField] = itemRO.DeepCopy()
 		}
 	}
 	// fill up the missing entries
+	types.SettingDefinitionsLock.RLock()
```
innobead, line 401: Could we just have a `GetSettingDefinitions` in setting.go? so we don't need to use a specific lock in setting.go, but rather focus on using the interfaces from types package only. (it will not 100% atomic, but it should be fine)

innobead (changes requested):

types/setting.go
```diff
@@ -1120,12 +1124,25 @@ func UnmarshalNodeSelector(nodeSelectorSetting string) (map[string]string, error
...
-		SettingDefinitions[sName] = definition
+		SetSettingDefinition(sName, definition)
 	}
 	return nil
 }
+
+func GetSettingDefinition(name SettingName) (SettingDefinition, bool) {
+	SettingDefinitionsLock.RLock()
+	defer SettingDefinitionsLock.RUnlock()
+	setting, ok := SettingDefinitions[name]
```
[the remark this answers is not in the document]
derekbit: Addressed. Thank you.

datastore/longhorn.go
derekbit: Addressed. Thank you.

innobead (approved):

PR #1347, opened 2022-05-26
Update backing image related implementation & Add backing image download feature

shuo-wu:
longhorn/longhorn#3970
longhorn/longhorn#3155


shuo-wu:
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/1377/

shuo-wu:
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/1380/

controller/backing_image_controller.go
```diff
@@ -294,15 +294,15 @@ func (bic *BackingImageController) cleanupBackingImageManagers(bi *longhorn.Back
 		if bim.DeletionTimestamp != nil {
 			continue
 		}
-
+		bimLog := log.WithField("backingImageManager", bim.Name)
 		// Directly clean up old backing image managers (including incompatible managers).
 		// New backing image managers can detect and reuse the existing backing image files if necessary.
 		if bim.Spec.Image != defaultImage {
-			log.WithField("backingImageManager", bim.Name).Info("Start to delete old backing image manager")
+			bimLog.Info("Start to delete old/non-default backing image manager")
```
innobead, line 301: nit: "Start to delete old/non-default ...."

controller/backing_image_controller.go
```diff
@@ -431,6 +432,10 @@ func (bic *BackingImageController) handleBackingImageDataSource(bi *longhorn.Bac
 	}
 	existingBIDS := bids.DeepCopy()
 
+	if bids.Spec.UUID == "" {
+		bids.Spec.UUID = bi.Status.UUID
+	}
```
innobead, line 437: Why not just handle this during the upgrade migration to make this change obvious because of new change introduced from 1.3.0?

engineapi/types.go
```diff
@@ -24,6 +24,10 @@ const (
 	InstanceManagerDefaultPort      = 8500
 	InstanceManagerProxyDefaultPort = InstanceManagerDefaultPort + 1
 
+	BackingImageManagerDefaultPort    = 8000
+	BackingImageDataSourceDefaultPort = 8000
+	SyncServerDefaultPort             = 8001
```
innobead, line 29: Suggest `BackingImageSyncServerDefaultPort`

controller/backing_image_manager_controller.go
```diff
@@ -590,7 +590,7 @@ func (c *BackingImageManagerController) deleteInvalidBackingImages(bim *longhorn
 			continue
 		}
 		log.Debugf("Start to delete the file for invalid backing image %v", biName)
-		if err := cli.Delete(biName); err != nil && !types.ErrorIsNotFound(err) {
+		if err := cli.Delete(biName, bi.Status.UUID); err != nil && !types.ErrorIsNotFound(err) {
```
innobead, line 593: nit: `cli.Delete(biName, biUUID)`

controller/backing_image_manager_controller.go
```diff
@@ -339,9 +339,9 @@ func (c *BackingImageManagerController) cleanupBackingImageManager(bim *longhorn
 			log.WithError(err).Warnf("Failed to launch a gRPC client during cleanup, will skip deleting all files")
 		} else {
 			log.Infof("Start to delete all backing image files during cleanup")
-			for biName := range bim.Status.BackingImageFileMap {
-				if err := cli.Delete(biName); err != nil {
-					log.WithError(err).Warnf("Failed to launch a gRPC client during cleanup, will skip deleting the file for backing image %v", biName)
+			for biName, biFileInfo := range bim.Status.BackingImageFileMap {
+				if err := cli.Delete(biName, biFileInfo.UUID); err != nil {
```
innobead, line 343: If the UUID of the BI is unique, why not just `cli.Delete(biFileInfo.UUID)` only? keeping it for debuggin purpose?

controller/backing_image_manager_controller.go
```diff
@@ -820,6 +824,16 @@ func (c *BackingImageManagerController) generateBackingImageManagerPodManifest(b
 							MountPath: bimtypes.DiskPathInContainer,
 						},
 					},
+					Env: []v1.EnvVar{
+						{
+							Name: types.EnvPodIP,
+							ValueFrom: &v1.EnvVarSource{
+								FieldRef: &v1.ObjectFieldSelector{
+									FieldPath: "status.podIP",
```
innobead, line 832: Where is the env var used? If for data sync, why not storageIP?

engineapi/backing_image_manager.go
```diff
@@ -69,40 +67,47 @@ func (c *BackingImageManagerClient) parseBackingImageFileInfo(bi *bimapi.Backing
...
 
-func (c *BackingImageManagerClient) Delete(name string) error {
+func (c *BackingImageManagerClient) PrepareDownload(name, uuid string) (string, string, error) {
+	if err := CheckBackingImageManagerCompatibilty(c.apiMinVersion, c.apiVersion); err != nil {
+		return "", "", err
+	}
+	return c.grpcClient.PrepareDownload(name, uuid)
+}
+
+func (c *BackingImageManagerClient) Delete(name, uuid string) error {
```
innobead, line 99: I am curious why we need to use two unique identifies to do the operation? Why not just `Delete(uuid string)`? the same for Get.

controller/node_controller.go
```diff
@@ -976,13 +976,13 @@ func BackingImageDiskFileCleanup(node *longhorn.Node, bi *longhorn.BackingImage,
...
 		//  1. If there are enough ready files for a backing image, it's fine to do cleanup.
-		//  2. If there are no enough ready files, try to retain handling(state pending/starting/in-progress/unknown) files to guarantee the HA requirement.
+		//  2. If there are no enough ready files, try to retain handling(state empty/pending/starting/in-progress/unknown) files to guarantee the HA requirement.
 		//  3. If there are no enough ready & handling files, try to retain failed files to guarantee the HA requirement.
 		//  4. If there are no enough files including failed ones, skip cleanup.
 		fileStatus, exists := bi.Status.DiskFileStatusMap[diskUUID]
 		if !exists {
 			fileStatus = &longhorn.BackingImageDiskFileStatus{
-				State: longhorn.BackingImageStatePending,
+				State: "",
```
innobead, line 985: BTW, `BackingImageDiskFileCleanup` seems only used privately, so no need to export but `backingImageDiskFileCleanup` instead`.

innobead (commented):
In general, LGTM. Just a few comments to resolve.

controller/backing_image_controller.go
```diff
@@ -486,7 +491,7 @@ func (bic *BackingImageController) handleBackingImageDataSource(bi *longhorn.Bac
 	}
 
 	// Check if the data source already finished the 1st file preparing.
-	if !bids.Spec.FileTransferred && bids.Status.CurrentState == longhorn.BackingImageStateReadyForTransfer {
+	if !bids.Spec.FileTransferred && bids.Status.CurrentState == longhorn.BackingImageStateReady {
```
derekbit, line 494: Curious. Why chage it to BackingImageStateReady instead of BackingImageStateReadyForTransfer?

api/forwarder.go
```diff
@@ -128,21 +137,26 @@ func (f *Fwd) HandleProxyRequestByNodeID(targetAddress string, req *http.Request
...
 const (
 	BackingImageUpload = "upload"
 )
 
-func (f *Fwd) HandleProxyRequestForBackingImageUpload(targetAddress string, req *http.Request) (proxyRequired bool, err error) {
+func (f *Fwd) HandleProxyRequestForBackingImageUpload(parameters map[string]string, req *http.Request) (proxyRequired bool, err error) {
+	targetAddress := parameters[ParameterKeyAddress]
 	if targetAddress == req.Host {
 		return false, fmt.Errorf("backing image upload request should not be handled by current longhorn manager service")
 	}
```
derekbit, line 162: Why do we need this check here?
Will the function return error as exepceted later if `targetAddress == req.Host` and without this check?

vendor/github.com/longhorn/backing-image-manager/pkg/types/types.go
```diff
@@ -9,18 +12,17 @@ const (
...
-	DefaultDataSourceServerPort     = 8001
+	DefaultDataSourceServerPort     = 8000
+	DefaultSyncServerPort           = 8001
 	DefaultVolumeExportReceiverPort = 8002
 
-	GRPCServiceTimeout     = 3 * time.Minute
-	HTTPTimeout            = 4 * time.Second
-	FileValidationInterval = 5 * time.Second
-	FileSyncTimeout        = 120
+	GRPCServiceTimeout = 3 * time.Minute
```
derekbit, line 22: Just curious. How to determine how long the timeout should be?

controller/backing_image_controller.go
shuo-wu: I don't know if users or others will unset this field accidentally hence I prefer to leave this check & update here.

controller/backing_image_manager_controller.go
shuo-wu: The variable `biUUID` should be invalid here hence we cannot use it.

controller/backing_image_manager_controller.go
shuo-wu: We should allow the BIM to fall back to using the pod IP when the Storage Network is not launched.
https://github.com/shuo-wu/backing-image-manager/blob/6dc15c6ec60a126590ac117214122c1c1a15e5cd/pkg/util/http_util.go#L59

controller/backing_image_controller.go
shuo-wu: Right now the BackingImageDataSource will update the state to ready by itself when the BIM takes over the file:
https://github.com/shuo-wu/backing-image-manager/blob/6dc15c6ec60a126590ac117214122c1c1a15e5cd/pkg/manager/service.go#L401
https://github.com/shuo-wu/backing-image-manager/blob/6dc15c6ec60a126590ac117214122c1c1a15e5cd/pkg/datasource/service.go#L318-L319

api/forwarder.go
shuo-wu: This function is executed in the longhorn-manager pods and the `req.Host` is the manager pod IP. While the `targetAddress` should be the BIDS pod.
By explicitly erroring out here, I will know what happens...
If we didn't do anything, it will fallback to run the last function in this line. I may need to improve this by using a different function rather than `backingImageGet`: https://github.com/shuo-wu/longhorn-manager/blob/faa879b5ab35ce8f897c6b651c0c4592830de425/api/router.go#L147

vendor/github.com/longhorn/backing-image-manager/pkg/types/types.go
shuo-wu: Nice catch.
Typically it depends on if there is a blocking gRPC API and how long it can take.
But here there should be no blocking call since we should use a shorter timeout. Or we need to reuse the client/connection (similar to longhorn-engine/longhorn-instance-manager parts). I will add a TODO here.

controller/backing_image_manager_controller.go
shuo-wu: Previously, the backing image sync server uses the backing image file path as the unique key. 
To figure out the path, both biName and the biUUID are required.

Later on due to the feature requirement, the sync server has to store the UUID as well. But I didn't switch to using the UUID as the unique key. (the time is limited and the filepath is straightforward hence I am hesitating to modify this.)
If the sync server switches to using UUID only, then there is no need to input biName.

shuo-wu: Do you think the sync server need to switch to using the UUID?

innobead: Due to limited time, let's keep it as is first. We can do it as refactoring in the future, that is fine for now.

shuo-wu: Got it!

controller/backing_image_manager_controller.go
innobead: cc @c3y1huang

controller/backing_image_controller.go
derekbit: I see. Thank you.

api/forwarder.go
derekbit: Understood. Thank you.

controller/backing_image_controller.go
innobead: Generally, when making logs around an action can follow some logging conventions which can help us easily understand the context. 

- "Verb+ing ..." for before taking the action
- "Failed to Verb ..." for failing the action
- "Verb+ed" for taking the action successfully (this is optional)

However, this is just a Nit.

controller/node_controller.go
shuo-wu: I was thinking if this can be used by others later. Let me re-check it when implementing HA backing image feature.

shuo-wu:
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/1417/

innobead (approved):

PR #1348, opened 2022-05-26
grpc-proxy: refactor proxy client on demand

c3y1huang:
We want to be able to compare and see the result of a different approach to get the proxy client. 

This (`the same as the current implementation, Engine CLI, with one connection for each operation`) may also be a backup plan in place of the connection pool in https://github.com/longhorn/longhorn-manager/pull/1338


innobead (approved):
LGTM

We need to be more careful about the pros and cons of 
https://github.com/longhorn/longhorn-manager/pull/1338, and need more testing as well. 

This PR will be the baseline for us to keep the original way of one operation one connection, but still have the new proxy implementation in 1.3.0.

controller/engine_controller.go
```diff
@@ -1402,6 +1396,7 @@ func (ec *EngineController) startRebuilding(e *longhorn.Engine, replica, addr st
 	if err != nil {
 		return err
 	}
+	defer engineClientProxy.Done()
```
innobead, line 1399: This should be moved to the below go routine to ensure replica rebuilding successfully.

innobead (changes requested):

engineapi/proxy.go
```diff
@@ -1,288 +1,128 @@
...
-
-	InstanceManagerName         string
-	InstanceManagerControllerID string
-
-	// used to notify the controller that proxy has stopped
-	voluntaryStopCh chan struct{}
-	stopCh          chan struct{}
-
+	logger     logrus.FieldLogger
+	ds         *datastore.DataStore
```
shuo-wu, line 96: It seems that the proxy does not need this

engineapi/proxy.go
```diff
@@ -1,288 +1,128 @@
...
-			"instanceManagerImage": image,
+			"instanceManager": im.Name,
+			"image":           im.Spec.Image,
+			"serverIP":        im.Status.IP,
 		},
 	)
 }
 
-func (h *EngineClientProxyHandler) GetCompatibleClient(e *longhorn.Engine, fallBack interface{}) (c EngineClientProxy, err error) {
+func GetCompatibleClient(e *longhorn.Engine, fallBack interface{}, ds *datastore.DataStore, logger logrus.FieldLogger) (c EngineClientProxy, err error) {
```
shuo-wu, line 25: I prefer to input `im` rather than `ds`

innobead (dismissed):

engineapi/instance_manager.go
```diff
@@ -64,13 +66,21 @@ func CheckInstanceManagerCompatibilty(imMinVersion, imVersion int) error {
...
 	return nil
 }
 
+func CheckInstanceManagerProxySupport(im *longhorn.InstanceManager) error {
+	if UnsupportInstanceManagerProxyAPIVersion == im.Status.ProxyAPIVersion {
+		return fmt.Errorf("%v does not support proxy", im.Name)
+	}
+	return nil
+}
+
```
derekbit, line 83: nit: return a boolean value. The error message "... dost not suppport proxy" can move to caller side.

engineapi/proxy.go

controller/engine_controller.go
c3y1huang: Fixed.

engineapi/instance_manager.go
c3y1huang: I feel it should be consistent with [CheckInstanceManagerComptibility](https://github.com/longhorn/longhorn-manager/blob/master/engineapi/instance_manager.go#L58). WDYT?

engineapi/proxy.go
```diff
@@ -1,288 +1,126 @@
...
-		if err := p.Stop(im.Spec.Image); err != nil {
-			p.logger.WithError(err).Error("Failed to close proxy client")
-		}
-	}()
+type EngineClientProxy interface {
+	EngineClient
+	BackupTargetBinaryClient
 
-	return nil
+	Done()
```
derekbit, line 102: nit: how about renaming to `Close()`?

engineapi/instance_manager.go
derekbit: I see. I'm fine with it.

engineapi/proxy.go

controller/engine_controller.go
```diff
@@ -1336,13 +1329,15 @@ func (ec *EngineController) removeUnknownReplica(e *longhorn.Engine) error {
...
-	engineClientProxy, err := ec.getEngineClientProxy(e, e.Status.CurrentImage)
-	if err != nil {
-		return err
-	}
-
 	for _, url := range unknownReplicaURLs {
 		go func(url string) {
+			engineClientProxy, err := ec.getEngineClientProxy(e, e.Status.CurrentImage)
+			if err != nil {
+				ec.logger.WithError(err).Warnf("Failed to remove unknown replica %v from engine", url)
```
derekbit, line 1336: Should we use `Errorf()` here?

c3y1huang: This is a client retrieval error not related the actual `ReplicaRemove`. I feel this should be in warning because it might resolve itself on the next queue. WDYT?

derekbit: I see. Thank you.

derekbit (approved):
LGTM

controller/engine_controller.go
c3y1huang: Error this outside of goroutine so it gets enqueued again.

innobead (approved):
LGTM

innobead:
@c3y1huang Please also update longhorn/longhorn, it seems you updating crd resource part.

engineapi/proxy.go
```diff
@@ -1,288 +1,126 @@
...
 	hasIP := im.Status.IP != ""
 	if !hasIP {
-		return nil, errors.Errorf("cannot create new gRPC client while instance manager status IP is empty")
+		err = errors.Errorf("%v instance manager status IP is missing", im.Name)
+		return nil, err
 	}
 
-	client := &Proxy{}
-	if err := client.Start(im, log, h.ds); err != nil {
+	ctx, cancel := context.WithCancel(context.Background())
```
PhanLe1010, line 81: This conext doesn't timeout which mean that all future grpc has no timeout. E.g., [snapshot list](https://github.com/longhorn/longhorn-manager/blob/fb4e45e7e6db63cf4b15f0f755a1b37616d9e678/vendor/github.com/longhorn/longhorn-instance-manager/pkg/client/proxy_snapshot.go#L66)

I am thinking that we should create a new context timeout for each grpc call. I.e, create a new context with timeout from `c.ctx` [here](https://github.com/longhorn/longhorn-manager/blob/fb4e45e7e6db63cf4b15f0f755a1b37616d9e678/vendor/github.com/longhorn/longhorn-instance-manager/pkg/client/proxy_snapshot.go#L66) and use it. What do you think?

innobead: Because this is a proxy, the timeout context is now just delegated to the proxy server instead, so in the proxy server Proxy::SnapshotList, there is a context timeout when calling to the downstream engine gRPC server as usual, but there is no another timeout inf ProxyClient::SnapshotList as you mentioned.

Simply said, there is still a timeout, but it is on the proxy server-side instead of the client-side. This can be improved (nice to have) in the future, but remember to have the timeout longer than the one used in the proxy server to avoid the backend operation timeout earlier than its setting. 

```golang
func (p *Proxy) SnapshotList(ctx context.Context, req *rpc.ProxyEngineRequest) (resp *rpc.EngineSnapshotListProxyResponse, err error) {
	...
	recv, err := c.ReplicaList() //<-- timeout context inside
	....

	return resp, nil
}
```

innobead: cc @c3y1huang @shuo-wu @derekbit

shuo-wu: Make sense. But I think it's better to have a timeout between the proxy server and the longhorn managers as well. So that the caller threads in the longhorn managers won't be blocked.

c3y1huang: https://github.com/longhorn/longhorn/issues/4060

PR #1349, opened 2022-05-27
grpc-proxy: fix ctx cancelled during replica list

c3y1huang:
https://github.com/longhorn/longhorn/issues/3546

### DIFF
--- a/controller/engine_controller.go
+++ b/controller/engine_controller.go
@@ -1412,9 +1412,15 @@ func (ec *EngineController) startRebuilding(e *longhorn.Engine, replica, addr st
 
 	replicaURL := engineapi.GetBackendReplicaURL(addr)
 	go func() {
-		defer engineClientProxy.Close()
-
 		log := ec.logger.WithField("volume", e.Spec.VolumeName)
+
+		engineClientProxy, err := ec.getEngineClientProxy(e, e.Status.CurrentImage)
+		if err != nil {
+			log.WithError(err).Errorf("Failed rebuilding of replica %v", addr)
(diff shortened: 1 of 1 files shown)


PhanLe1010:
Could you explain a bit more about the issue that we are solving?

innobead:
I think @c3y1huang is trying to fix a concurrent issue that `engineClientProxy` is closed during using it to wait for the replica to be added.

@c3y1huang What do you think the below fix instead?

```golang
	engineClientProxy, err := ec.getEngineClientProxy(e, e.Status.CurrentImage)
	if err != nil {
		return err
	}
        engineClientProxy.Close()
``
, then just create another engine client in the goruntine.

**Review by innobead — CHANGES_REQUESTED** [review]
I think @c3y1huang is trying to fix a concurrent issue that `engineClientProxy` is closed during using it to wait for the replica to be added.

@c3y1huang I think we can remove `engineClientProxy.Close()` and just leave the variable to garbage collected by the go runtime instead of the graceful shutdown, so you don't need to change any other code, but need to check if this will cause the connection to leak? or the other way is to have another engine client proxy in the goroutine instead.

File: controller/engine_controller.go [hunk 1]
```diff
@@ -1412,10 +1412,16 @@ func (ec *EngineController) startRebuilding(e *longhorn.Engine, replica, addr st
 
 	replicaURL := engineapi.GetBackendReplicaURL(addr)
 	go func() {
-		defer engineClientProxy.Close()
-
 		log := ec.logger.WithField("volume", e.Spec.VolumeName)
 
+		engineClientProxyGo, err := ec.getEngineClientProxy(e, e.Status.CurrentImage)
```
innobead, line 1417: nit: `engineClientProxy` should be able to be shadowed in another scope, so basically you can use the same name as outer if you want.

innobead (approved):

PR #1351, opened 2022-05-30
Fix unexpected overwrite of default settings

derekbit:
default setting config -> this should impact default builtin setting definitions
updating setting CR -> this should not impact the default setting definitions

longhorn/longhorn#2570

### DIFF
--- a/datastore/longhorn.go
+++ b/datastore/longhorn.go
@@ -12,7 +12,6 @@ import (
 	"github.com/sirupsen/logrus"
 
 	corev1 "k8s.io/api/core/v1"
-	v1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -42,7 +41,7 @@ var (
 	VerificationRetryCounts = 20
 )
 
-func (s *DataStore) UpdateCustomizedSettings(overrideDefaultSettings map[types.SettingName]string) error {
+func (s *DataStore) UpdateCustomizedSettings(defaultImage map[types.SettingName]string) error {
 	defaultSettingCM, err := s.GetConfigMap(s.namespace, types.DefaultDefaultSettingConfigMapName)
 	if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


datastore/longhorn.go
```diff
@@ -117,17 +121,24 @@ func (s *DataStore) updateSettingDefinitions(customizedDefaultSettings map[strin
 }
 
 func (s *DataStore) createOrUpdateSetting(name types.SettingName, definition types.SettingDefinition) error {
+	// Skip the create or update to avoid the validation error from ValidateSetting(),
+	// if the required setting is not updated somewhere.
+	if definition.Required && definition.Current == "" {
+		return nil
+	}
```
innobead, line 128: Should it just be?

```
	if definition.Required{
		return nil
	}
```

innobead (commented):
LGTM except one comment.

datastore/longhorn.go
derekbit: This function is to create or update settings according the values collected from setting CRs or configmap.
In general, the required field should be created and updated with a given non-empty value.
If it is an empty value, we skip them here. Thus, we need to check `definition.Current` here.

innobead: I think I misread the code. Thanks for the explanation.

innobead:
cc @cchien816

derekbit:
Refactoring according to the discussion with @shuo-wu

innobead (dismissed):
LGTM, this looks better :slightly_smiling_face:

datastore/longhorn.go
```diff
@@ -53,108 +53,138 @@ func (s *DataStore) UpdateCustomizedSettings(overrideDefaultSettings map[types.S
...
+		setting.Annotations = map[string]string{}
+	}
+
+	setting.Annotations[types.GetLonghornLabelKey(types.ConfigMapResourceVersionKey)] = defaultSettingCMResourceVersion
+	setting.Value = value
+	_, err = s.UpdateSetting(setting)
+	return err
+}
+
+func (s *DataStore) updateSettingDefinitionsWithCustomizedDefaultSettings(customizedDefaultSettings map[string]string, defaultSettingCM *v1.ConfigMap) error {
```
shuo-wu, line 142: Remove input `defaultSettingCM`

datastore/longhorn.go
```diff
@@ -53,108 +53,138 @@ func (s *DataStore) UpdateCustomizedSettings(overrideDefaultSettings map[types.S
 		return err
 	}
 
-	if err := s.updateSettingDefinitions(customizedDefaultSettings, defaultSettingCM); err != nil {
+	if err := s.updateSettingDefinitionsWithCustomizedDefaultSettings(customizedDefaultSettings, defaultSettingCM); err != nil {
 		return err
 	}
 
-	if err := types.OverrideSettingDefinitions(overrideDefaultSettings); err != nil {
+	if err := s.syncSettingCRsFromCommandLine(overrideDefaultSettings); err != nil {
```
shuo-wu, line 60: NIT: `syncSettingsWithCMDSettingDefinitions`

datastore/longhorn.go
```diff
@@ -53,108 +53,138 @@ func (s *DataStore) UpdateCustomizedSettings(overrideDefaultSettings map[types.S
...
-	for _, sName := range types.SettingNameList {
-		definition, ok := types.GetSettingDefinition(sName)
-		if !ok {
-			return fmt.Errorf("BUG: setting %v is not defined", sName)
-		}
-
-		if err := s.createOrUpdateSetting(sName, definition); err != nil {
-			return err
-		}
+	if err := s.updateNonExistingSettingCRsWithDefaultSetting(defaultSettingCM.ResourceVersion); err != nil {
```
shuo-wu, line 64: Shouldn't the function `applyConfigMapSettingDefinitions` here?

datastore/longhorn.go
```diff
@@ -53,108 +53,138 @@ func (s *DataStore) UpdateCustomizedSettings(overrideDefaultSettings map[types.S
...
+		return err
 	}
-	return nil
+
+	return s.syncDefaultSettingsWithSettingCRs(customizedDefaultSettings, defaultSettingCM.ResourceVersion)
 }
 
-func (s *DataStore) updateSettingDefinitions(customizedDefaultSettings map[string]string, defaultSettingCM *v1.ConfigMap) error {
+func (s *DataStore) updateNonExistingSettingCRsWithDefaultSetting(configMapResourceVersion string) error {
```
shuo-wu, line 71: It seems that this can be part of `syncDefaultSettingsWithSettingCRs`

datastore/longhorn.go
```diff
@@ -53,108 +53,138 @@ func (s *DataStore) UpdateCustomizedSettings(overrideDefaultSettings map[types.S
...
+		if !ok {
+			return fmt.Errorf("BUG: setting %v is not defined", sName)
 		}
 
-		_, err = s.UpdateSetting(setting)
-		return err
+		value, exist := customizedDefaultSettings[string(sName)]
+		if !exist {
+			continue
+		}
```
shuo-wu, line 178: I don't understand why we still need the input `customizedDefaultSettings` or `definition, ok := types.GetSettingDefinition(sName)` here.

datastore/longhorn.go
```diff
@@ -53,108 +53,138 @@ func (s *DataStore) UpdateCustomizedSettings(overrideDefaultSettings map[types.S
...
-
-		if err := s.createOrUpdateSetting(sName, definition); err != nil {
-			return err
-		}
+	if err := s.updateNonExistingSettingCRsWithDefaultSetting(defaultSettingCM.ResourceVersion); err != nil {
+		return err
 	}
-	return nil
+
+	return s.syncDefaultSettingsWithSettingCRs(customizedDefaultSettings, defaultSettingCM.ResourceVersion)
```
shuo-wu, line 68: NIT: `syncSettingsWithConfigMapSettingDefinitions`

shuo-wu (changes requested):
Do we need to update the ConfigMap controller either?

datastore/longhorn.go
derekbit: Sorry. We need `customizedDefaultSettings`, because we only want to update the modified fields.

The purpose of `definition, ok := types.GetSettingDefinition(sName)` is only to know if the field is `Required` .

datastore/longhorn.go
```diff
@@ -53,108 +52,138 @@ func (s *DataStore) UpdateCustomizedSettings(overrideDefaultSettings map[types.S
...
+		definition, ok := types.GetSettingDefinition(sName)
+		if !ok {
+			return fmt.Errorf("BUG: setting %v is not defined", sName)
 		}
 
-		return nil
+		if value, ok := customizedDefaultSettings[string(sName)]; ok {
+			definition.Default = value
+			types.SetSettingDefinition(sName, definition)
+		}
```
shuo-wu, line 151: Maybe it's better to check if the field is read-only before applying the customization from ConfigMap.

shuo-wu (dismissed):
In general LGTM

datastore/longhorn.go
derekbit: Good idea. Thank you.

innobead (approved):

PR #1356, opened 2022-05-31
Fix unexpected overwrite of default settings

derekbit:
default setting config -> this should impact default builtin setting definitions
updating setting CR -> this should not impact the default setting definitions

https://github.com/longhorn/longhorn/issues/3488


innobead (approved):
LGTM

PR #1358, opened 2022-06-01
grpc-proxy: fix NPE from SnapshotBackup

c3y1huang:
https://github.com/longhorn/longhorn/issues/4061


innobead (approved):
LGTM, but do you know what case causes this?? Because I assume when calling `SnapshotBackup`, the volume should be mounted and the engine should be available, otherwise it will not go to this method actually.

note: of course, for any function/method uses passed references, all need to do nil check.

PR #1361, opened 2022-06-02
grpc-proxy: transfer backup CR ownership when IM is down

c3y1huang:
Before the proxy, the backup volume controller can access backups when the instance manager on the node is down.

Now, the manager cannot make the connection when the instance manager is down. We will actively transfer the backup CR ownership to another node with a healthy instance manager.

https://github.com/longhorn/longhorn/issues/4064

### DIFF
--- a/controller/backup_controller.go
+++ b/controller/backup_controller.go
@@ -99,6 +99,11 @@ func NewBackupController(
 	})
 	bc.cacheSyncs = append(bc.cacheSyncs, ds.BackupInformer.HasSynced)
 
+	ds.InstanceManagerInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		UpdateFunc: func(old, cur interface{}) { bc.enqueueInstanceManagerChange(cur) },
+	})
+	bc.cacheSyncs = append(bc.cacheSyncs, ds.InstanceManagerInformer.HasSynced)
+
 	return bc
 }
 
@@ -114,6 +119,50 @@ func (bc *BackupController) enqueueBackup(obj interface{}) {
 
 func (bc *BackupController) enqueueBackupForMonitor(key string) {
 	bc.queue.Add(key)
+}
+
+func (bc *BackupController) enqueueInstanceManagerChange(obj interface{}) {
+	im, isInstanceManager := obj.(*longhorn.InstanceManager)
+	if !isInstanceManager {
+		deletedState, ok := obj.(cache.DeletedFinalStateUnknown)
+		if !ok {
+			utilruntime.HandleError(fmt.Errorf("received unexpected obj: %#v", obj))
+			return
+		}
+
+		// use the last known state, to enqueue, dependent objects
+		im, ok = deletedState.Obj.(*longhorn.InstanceManager)
+		if !ok {
+			utilruntime.HandleError(fmt.Errorf("DeletedFinalStateUnknown contained invalid object: %#v", deletedState.Obj))
+			return
+		}
+	}
+
+	imType, err := datastore.CheckInstanceManagerType(im)
+	if err != nil || imType != longhorn.InstanceManagerTypeEngine {
+		return
+	}
+
+	if im.Status.OwnerID != bc.controllerID {
+		return
+	}
+
+	if im.Status.CurrentState == longhorn.InstanceManagerStateStarting || im.Status.CurrentState == longhorn.InstanceManagerStateRunning {
+		return
+	}
+
+	backups, err := bc.ds.ListBackupsRO()
+	if err != nil {
+		bc.logger.WithError(err).Warnf("Failed to list backups")
+		return
+	}
+	for _, backup := range backups {
+		if backup.Status.OwnerID != bc.controllerID {
+			continue
+		}
+
+		bc.enqueueBackup(backup)
+	}
 }
 
 func (bc *BackupController) Run(workers int, stopCh <-chan struct{}) {
@@ -221,6 +270,18 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
 		return nil
 	}
 	if !isResponsible {
+		if errIM := bc.ds.CheckDefaultEngineInstanceManagerAvailability(backup.Status.OwnerID); errIM != nil {
+			bc.logger.WithError(errIM).Infof("Transfering backup %v ownership", backup.Name)
+			backup.Status.OwnerID = ""
+			_, err = bc.ds.UpdateBackupStatus(backup)
+			if err != nil {
+				// we don't mind others coming first
+				if apierrors.IsConflict(errors.Cause(err)) {
+					return nil
+				}
+				return err
+			}
+		}
 		return nil
 	}
 	if backup.Status.OwnerID != bc.controllerID {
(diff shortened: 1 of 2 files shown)


controller/backup_controller.go
```diff
@@ -116,6 +121,49 @@ func (bc *BackupController) enqueueBackupForMonitor(key string) {
...
+		return
+	}
+
+	if im.Status.OwnerID != bc.controllerID {
+		return
+	}
+
+	if im.Status.CurrentState == longhorn.InstanceManagerStateStarting || im.Status.CurrentState == longhorn.InstanceManagerStateRunning {
+		return
+	}
```
PhanLe1010, line 152: I am a bit confused about this condition. Assume that we have the following case:
* instance-manager-e-xxx is running on node-1
* instance-manager-e-xxx is failed on node-1 and remains to be failed
* instance-manager-e-xxx still has im.Status.OwnerID is node-1
* only node-1 handles this instance manager because of the check 
  ```
  if im.Status.OwnerID != bc.controllerID {
		  return
	  }
  ```
* we enqueue all the backup with backup.Status.OwnerID is node-1
* Latter on, we reject to handle all of those backups because this condition `currentNodeInstanceManagerAvailable := bc.ds.IsDefaultEngineInstanceManagerAvailability(bc.controllerID)`. So, is it true that the backups are still not being handled by anyone?

I think we don't observer this problem yet because of the 30s resync period that dumps all backup CRs into other nodes. So, those nodes have chance to take over the backups. However, if we disable resync, we might see the problem. How do you think?

PhanLe1010 (commented):
May I have a question in the comment below

Also, I am thinking if we can simply move the backup get/list operation out of the proxy path. It would be cleaner and cheaper than ownership transfer logic.

controller/backup_controller.go
c3y1huang: You are right, thanks for pointing it out. Fixed.

c3y1huang:
Test result - https://ci.longhorn.io/job/private/job/longhorn-tests-regression/1463/

api/model.go
```diff
@@ -114,6 +114,7 @@ type Backup struct {
 	VolumeSize             string               `json:"volumeSize"`
 	VolumeCreated          string               `json:"volumeCreated"`
 	VolumeBackingImageName string               `json:"volumeBackingImageName"`
+	OwnerID                string               `json:"ownerID"`
```
innobead, line 117: Do we need this internal info? If people want to know, users can check the backup resource directly

controller/backup_controller.go
```diff
@@ -116,6 +121,50 @@ func (bc *BackupController) enqueueBackupForMonitor(key string) {
...
+
+	backups, err := bc.ds.ListBackupsRO()
```
innobead, line 154: This will be a potential concern about the cost if there are tons of backup resources needed to query. Is there a way to filter by owner instead?

controller/backup_controller.go
```diff
@@ -221,6 +270,18 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
 		return nil
 	}
 	if !isResponsible {
+		if errIM := bc.ds.CheckDefaultEngineInstanceManagerAvailability(backup.Status.OwnerID); errIM != nil {
+			bc.logger.WithError(errIM).Infof("Transfering backup %v ownership", backup.Name)
+			backup.Status.OwnerID = ""
```
innobead, line 275: Should you just simply make `backup.Status.OwnerID = ""` and update it instead of checking `CheckDefaultEngineInstanceManagerAvailability` again?

You already do the check in `isResponsibleFor`, so I think we can just simply transfer its owner to another node instead.

datastore/longhorn.go
```diff
@@ -1403,6 +1403,24 @@ func (s *DataStore) CheckEngineImageReadyOnAllVolumeReplicas(image, volumeName,
 	return s.CheckEngineImageReadiness(image, nodes...)
 }
 
+func (s *DataStore) IsDefaultEngineInstanceManagerAvailability(nodeID string) bool {
+	return s.CheckDefaultEngineInstanceManagerAvailability(nodeID) == nil
+}
+
+func (s *DataStore) CheckDefaultEngineInstanceManagerAvailability(nodeID string) error {
```
innobead, line 1410: Suggest merging this to `IsDefaultEngineInstanceManagerAvailability` if `CheckDefaultEngineInstanceManagerAvailability` is not required to export and used by others.

innobead (changes requested):

controller/backup_controller.go
innobead: cc @shuo-wu @derekbit @PhanLe1010 will this be a concern?

api/model.go
c3y1huang: I'm exposing the `OwnerID` for automation e2e.
https://github.com/longhorn/longhorn-tests/pull/986
```
    backup_volume, backup = find_backup(client, volume.name, snapshot.name)
    assert backup.ownerID != original_ownerID, \
        f'expecting {backup.name} ownerID {original_ownerID} to change'

```

datastore/longhorn.go
c3y1huang: Seems we can only filter by label selector. 
https://github.com/longhorn/longhorn-manager/blob/master/k8s/pkg/client/listers/longhorn/v1beta2/backup.go#L63

controller/backup_controller.go
c3y1huang: Seems we can only filter by label selector.
https://github.com/longhorn/longhorn-manager/blob/master/k8s/pkg/client/listers/longhorn/v1beta2/backup.go#L63

controller/backup_controller.go
c3y1huang: `isReponsibleFor` included other conditions so I am not sure if I should handle them together here. And besides, this will add some log for the transfer.

controller/backup_controller.go
shuo-wu: It seems that the lister (cache) doesn't support the field selector.
Not sure if we need to add a label for the owner ID

api/model.go
c3y1huang: Removed. https://github.com/longhorn/longhorn-tests/pull/986#discussion_r889940530

controller/backup_controller.go
shuo-wu: `isReponsibleFor` should determine if the **current** node can/need to take over the ownership. In other words, it's enough to check it in `bc.isResponsibleFor`

shuo-wu (commented):
How about the BackupVolume CRs?

innobead:
After discussing with @c3y1huang , I think we should use the old way (CLI) for backup operations instead because this will save our concerns:
- unnecessary ownership movement and it will cause extra potential heavy backup resource query
- no needs to rely on the IM proxy to help with backup operations as below because they don't really involve in the communication between engine and replicas
- no need to tackle backup credential issues anymore (https://github.com/longhorn/longhorn/issues/3977)

However, we still need to keep the engine image concept for backup commands. 

```golang
type BackupTargetBinaryClient interface {
	BackupGet(destURL string, credential map[string]string) (*Backup, error)
	BackupVolumeGet(destURL string, credential map[string]string) (volume *BackupVolume, err error)
	BackupNameList(destURL, volumeName string, credential map[string]string) (names []string, err error)
	BackupVolumeNameList(destURL string, credential map[string]string) (names []string, err error)
	BackupDelete(destURL string, credential map[string]string) (err error)
	BackupVolumeDelete(destURL, volumeName string, credential map[string]string) (err error)
	BackupConfigMetaGet(destURL string, credential map[string]string) (*ConfigMetadata, error)
}
```

@shuo-wu @joshimoo @PhanLe1010 @derekbit any questions and concerns?

shuo-wu:
OK. In term of the the potential heavy resource queries, I am fine with reverting back to the binary invocation.

c3y1huang:
https://github.com/longhorn/longhorn/issues/4076

PR #1362, opened 2022-06-06
controller: Try to recover the backing image when all records become failed

shuo-wu:
https://github.com/longhorn/longhorn/issues/4073


controller/backing_image_controller.go
```diff
@@ -448,6 +448,7 @@ func (bic *BackingImageController) handleBackingImageDataSource(bi *longhorn.Bac
 		for diskUUID := range bi.Spec.Disks {
 			fileStatus, ok := bi.Status.DiskFileStatusMap[diskUUID]
 			if !ok {
+				allFilesUnavailable = false
 				continue
```
c3y1huang, line 452: Why not break here?

innobead: @shuo-wu :point_up:

shuo-wu: Fixed. Thank you!

innobead (approved):

PR #1364, opened 2022-06-06
deploy: update backing image manager tag to v3_20220606

shuo-wu:


innobead (changes requested):
Need to update deploy/longhorn-images.txt

innobead (approved):

PR #1365, opened 2022-06-06
grpc proxy: revert backup target grpc call

c3y1huang:
https://github.com/longhorn/longhorn/issues/4076


controller/backup_controller.go
```diff
@@ -266,15 +266,15 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
 
 		if backupTarget.Spec.BackupTargetURL != "" &&
 			backupVolume != nil && backupVolume.DeletionTimestamp == nil {
-			engineClientProxy, backupTargetConfig, err := getBackupTarget(bc.controllerID, backupTarget, bc.ds, log)
+			engineClientProxy, backupTargetClient, err := getBackupTarget(bc.controllerID, backupTarget, bc.ds, log)
```
innobead, line 269: Do we still need `getBackupTarget`? or just use `getBackupTargetClient` instead. I feel we don't need the returned `engineClientProxy` any more here.

innobead (changes requested):

controller/backup_controller.go
c3y1huang: Fixed.

c3y1huang:
Test result:
- test_backup: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/1470/
- coretest: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/1471/
- full regression: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/1472/

shuo-wu (approved):

derekbit (approved):
LGTM

PR #1371, opened 2022-06-09
Relieve the constraints of cleanupFailedToScheduledReplicas()

derekbit:
In the test case https://github.com/longhorn/longhorn/issues/3220#issuecomment-1149558769,
a replica cannot be scheduled to a node, but it's spec.failedAt is set in
https://github.com/longhorn/longhorn-manager/blob/master/controller/volume_controller.go#L1317.

The strict constraints of cleanupFailedToScheduledReplicas() results in that
the failed replica cannot be cleanup up.

[Longhorn 3320](https://github.com/longhorn/longhorn/issues/3220)

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -795,7 +795,7 @@ func (vc *VolumeController) cleanupFailedToScheduledReplicas(v *longhorn.Volume,
 	if healthyCount >= v.Spec.NumberOfReplicas {
 		for _, r := range rs {
 			if !hasEvictionRequestedReplicas {
-				if r.Spec.HealthyAt == "" && r.Spec.FailedAt == "" && r.Spec.NodeID == "" &&
+				if r.Spec.HealthyAt == "" && r.Spec.NodeID == "" &&
 					(isDataLocalityDisabled(v) || r.Spec.HardNodeAffinity != v.Status.CurrentNodeID) {
 					logrus.Infof("Cleaning up failed to scheduled replica %v", r.Name)
 					if err := vc.deleteReplica(r, rs); err != nil {


innobead (approved):

shuo-wu (approved):
In general LGTM.
But it's better to test if this will affect the failed replica reusage.

innobead:
> In general LGTM. But it's better to test if this will affect the failed replica reusage.

@derekbit Could you run a regression to see if any side effects from the regression as well? also @shuo-wu comment above? Thanks.

cc @longhorn/qa

derekbit:
> In general LGTM. But it's better to test if this will affect the failed replica reusage.

No prob.
Also waiting for the e2e results
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/1520/console

derekbit:
Did the following test
1. Create a 3 nodes Longhorn cluster
2. Create a 3-replicas volume and then attach to a node
3. Shut down one node. Then, the volume becomes degraded and one replica is failed as expected.
4. Create a 4-replicas volume and then attach to a node. The volume is degraded, and two replicas are failed and unscheduled.
5. Power up the node. The 3-replicas volume becomes healthy. The 4-replicas is degraded and only one replica is failed as expected.

shuo-wu:
Not sure if the test demonstrates anything... Two questions:
1. Does the failed replica of the 3-replica volume get reused after node reboot?
2. Does this test prove that the fix would help clean up the unscheduled & failed replicas?

derekbit:
1. Does the failed replica of the 3-replica volume get reused after node reboot?

Yes. The replica is reused.

2. Does this test prove that the fix would help clean up the unscheduled & failed replicas?
Yes. 
The unscheduled and failed replicas are cleaned up in est_basic.py::test_allow_volume_creation_with_degraded_availability

PR #1372, opened 2022-06-09
Update backing-image-manager vendor & image tag

shuo-wu:


innobead (approved):

innobead:
@shuo-wu Please also update longhorn/longhorn manifests.

PR #1375, opened 2022-06-10
Wrong flag isUnusedReplicaOnCurrentNode scope.

mantissahz:
'isUnusedReplicaOnCurrentNode' is always false at outer for loop.

longhorn/longhorn#2673


c3y1huang (approved):
Good catch. Thanks!

innobead (approved):
LGTM!

derekbit (commented):
LGTM

PR #1378, opened 2022-06-13
controller: avoid using the old backing image data source pod after upgrade

shuo-wu:
https://github.com/longhorn/longhorn/issues/4106


derekbit (approved):
LGTM

innobead (approved):

PR #1379, opened 2022-06-13
monitoring: add proxy-grpc-connection

c3y1huang:
https://github.com/longhorn/longhorn/issues/4038

### DIFF
--- a/app/daemon.go
+++ b/app/daemon.go
@@ -131,14 +131,16 @@ func startManager(c *cli.Context) error {
 		return err
 	}
 
-	ds, wsc, err := controller.StartControllers(logger, done, currentNodeID, serviceAccount, managerImage, kubeconfigPath, meta.Version)
+	proxyConnCounter := util.NewAtomicCounter()
+
+	ds, wsc, err := controller.StartControllers(logger, done, currentNodeID, serviceAccount, managerImage, kubeconfigPath, meta.Version, proxyConnCounter)
 	if err != nil {
 		return err
 	}
 
-	m := manager.NewVolumeManager(currentNodeID, ds)
+	m := manager.NewVolumeManager(currentNodeID, ds, proxyConnCounter)
 
-	metricsCollector.InitMetricsCollectorSystem(logger, currentNodeID, ds, kubeconfigPath)
+	metricsCollector.InitMetricsCollectorSystem(logger, currentNodeID, ds, kubeconfigPath, proxyConnCounter)
 
 	defaultImageSettings := map[types.SettingName]string{
 		types.SettingNameDefaultEngineImage:              engineImage,
--- a/engineapi/proxy.go
+++ b/engineapi/proxy.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/longhorn/longhorn-manager/datastore"
 	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
+	"github.com/longhorn/longhorn-manager/util"
 )
 
 func getLoggerForEngineProxyClient(logger logrus.FieldLogger, im *longhorn.InstanceManager) *logrus.Entry {
@@ -22,7 +23,7 @@ func getLoggerForEngineProxyClient(logger logrus.FieldLogger, im *longhorn.Insta
 	)
 }
 
-func GetCompatibleClient(e *longhorn.Engine, fallBack interface{}, ds *datastore.DataStore, logger logrus.FieldLogger) (c EngineClientProxy, err error) {
+func GetCompatibleClient(e *longhorn.Engine, fallBack interface{}, ds *datastore.DataStore, logger logrus.FieldLogger, proxyConnCounter util.Counter) (c EngineClientProxy, err error) {
 	if e == nil {
 		return nil, errors.Errorf("BUG: failed to get engine client proxy due to missing engine")
 	}
@@ -58,10 +59,10 @@ func GetCompatibleClient(e *longhorn.Engine, fallBack interface{}, ds *datastore
 		return nil, errors.Errorf("BUG: invalid engine client proxy fallback client: %v", fallBack)
 	}
 
-	return NewEngineClientProxy(im, log)
+	return NewEngineClientProxy(im, log, proxyConnCounter)
 }
 
-func NewEngineClientProxy(im *longhorn.InstanceManager, logger logrus.FieldLogger) (c EngineClientProxy, err error) {
+func NewEngineClientProxy(im *longhorn.InstanceManager, logger logrus.FieldLogger, proxyConnCounter util.Counter) (c EngineClientProxy, err error) {
 	defer func() {
 		err = errors.Wrap(err, "failed to get engine client proxy")
 	}()
@@ -84,15 +85,20 @@ func NewEngineClientProxy(im *longhorn.InstanceManager, logger logrus.FieldLogge
 		return nil, err
 	}
 
+	proxyConnCounter.IncreaseCount()
+
 	return &Proxy{
-		logger:     logger,
-		grpcClient: client,
+		logger:           logger,
+		grpcClient:       client,
+		proxyConnCounter: proxyConnCounter,
 	}, nil
 }
 
 type Proxy struct {
 	logger     logrus.FieldLogger
 	grpcClient *imclient.ProxyClient
+
+	proxyConnCounter util.Counter
 }
 
 type EngineClientProxy interface {
@@ -103,12 +109,18 @@ type EngineClientProxy interface {
 
 func (p *Proxy) Close() {
 	if p.grpcClient == nil {
+		p.logger.WithError(errors.New("gRPC client not exist")).Debugf("cannot close engine client proxy")
 		return
 	}
 
 	if err := p.grpcClient.Close(); err != nil {
-		p.logger.Warn("failed to close engine client proxy")
+		p.logger.WithError(err).Warn("failed to close engine client proxy")
 	}
+
+	// The only potential returning error from Close() is
+	// "grpc: the client connection is closing". This means we should still
+	// decrease the connection count.
+	p.proxyConnCounter.DecreaseCount()
 }
 
 func (p *Proxy) DirectToURL(e *longhorn.Engine) string {
--- a/manager/volume.go
+++ b/manager/volume.go
@@ -28,14 +28,18 @@ type VolumeManager struct {
 
 	currentNodeID string
 	sb            *SupportBundle
+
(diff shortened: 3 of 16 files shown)


innobead:
@c3y1huang can we make it to 1.3.0?

c3y1huang:
I think we can try. It's a small PR.

datastore/datastore.go
```diff
@@ -258,3 +266,24 @@ func ErrorIsNotFound(err error) bool {
...
+
+	s.proxyConnection++
+}
+
+func (s *DataStore) ProxyConnectionDecrease() {
+	s.proxyConnectionLock.Lock()
+	defer s.proxyConnectionLock.Unlock()
+
+	s.proxyConnection--
+}
```
derekbit, line 289: Nit: can we use `sync.RWMutex`?

engineapi/proxy.go
```diff
@@ -109,6 +114,8 @@ func (p *Proxy) Close() {
 	if err := p.grpcClient.Close(); err != nil {
 		p.logger.Warn("failed to close engine client proxy")
 	}
+
+	p.ds.ProxyConnectionDecrease()
```
derekbit, line 118: Any potential issues if you decrease the ProxyConnection value regardless of the `Close` error?

c3y1huang: I think the only returned error is `"grpc: the client connection is closing"` which means we should also decrease that count. WDYT?

Ref https://github.com/longhorn/longhorn-manager/blob/master/vendor/google.golang.org/grpc/clientconn.go#L1040

derekbit: OK, looks fine here. Thank you.

innobead (approved):
LGTM

engineapi/proxy.go
shuo-wu: Good explanation. Can you make it a comment here?

datastore/datastore.go
```diff
@@ -28,6 +30,9 @@ type DataStore struct {
 
 	cacheSyncs []cache.InformerSynced
 
+	proxyConnectionLock *sync.RWMutex
+	proxyConnection     int
```
shuo-wu, line 34: Can we launch a separate counter struct in `app/daemon.go` instead?
I prefer not to ask `datastore` to record this.

Besides, we can use functions in package `atomic` to operate the int.

innobead (approved):

metrics_collector/instance_manager_collector.go
```diff
@@ -188,3 +207,34 @@ func (imc *InstanceManagerCollector) collectRequestValues(ch chan<- prometheus.M
...
+		}
+	}()
+
+	podList, err := imc.ds.ListPodsRO(imc.namespace)
+	if err != nil {
+		imc.logger.WithError(err).Warn("error during scrape")
+		return
+	}
+
+	for _, pod := range podList {
```
shuo-wu, line 224: Any reason you prefer to list pods rather than IM CRs? Or you can list both.
One benefit of listing/checking the CRs is that you can verify the counter backing to 0 when the pod is done.

engineapi/proxy.go

datastore/datastore.go
c3y1huang: I cannot put it in the pkg app because of the import loop. After talking to @shuo-wu , we will put it in pkg util.

metrics_collector/instance_manager_collector.go
c3y1huang: Updated.

c3y1huang:
Test in-progress.

engineapi/proxy.go
```diff
@@ -58,10 +59,10 @@ func GetCompatibleClient(e *longhorn.Engine, fallBack interface{}, ds *datastore
 		return nil, errors.Errorf("BUG: invalid engine client proxy fallback client: %v", fallBack)
 	}
 
-	return NewEngineClientProxy(im, log)
+	return NewEngineClientProxy(im, log, ds, proxyStore)
 }
 
-func NewEngineClientProxy(im *longhorn.InstanceManager, logger logrus.FieldLogger) (c EngineClientProxy, err error) {
+func NewEngineClientProxy(im *longhorn.InstanceManager, logger logrus.FieldLogger, ds *datastore.DataStore, proxyStore util.ProxyStore) (c EngineClientProxy, err error) {
```
shuo-wu, line 65: Remember to remove `ds *datastore.DataStore` and the comments if the tests pass

app/daemon.go
```diff
@@ -131,14 +131,16 @@ func startManager(c *cli.Context) error {
 		return err
 	}
 
-	ds, wsc, err := controller.StartControllers(logger, done, currentNodeID, serviceAccount, managerImage, kubeconfigPath, meta.Version)
+	proxyStore := util.NewProxyStore()
```
innobead, line 134: The naming probably can be improved to understand. How about `proxyConnCounter` ?

util/proxy.go
```diff
@@ -0,0 +1,44 @@
...
+import (
+	"sync"
+)
+
+type Proxy struct {
+	connectionCount     int
+	connectionCountLock *sync.RWMutex
+}
+
+type ProxyStore interface {
```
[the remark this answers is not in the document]
innobead: AtomicCounter?

app/daemon.go
c3y1huang: Updated.

c3y1huang:
> util/proxy.go should be renamed as well, ex: counter.go or other general ones.

Updated.

util/counter.go
```diff
@@ -0,0 +1,44 @@
...
+	defer ac.countLock.Unlock()
+
+	ac.count++
+}
+
+func (ac *AtomicCounter) DecreaseCount() {
+	ac.countLock.Lock()
+	defer ac.countLock.Unlock()
+
+	ac.count--
```
derekbit, line 43: nit: `atomic.AddInt32(&ac.count, -1)`

util/counter.go
```diff
@@ -0,0 +1,44 @@
...
+		countLock: &sync.RWMutex{},
+		count:     0,
+	}
+}
+
+func (ac *AtomicCounter) GetCount() int {
+	ac.countLock.RLock()
+	defer ac.countLock.RUnlock()
+
+	return ac.count
```
derekbit, line 29: nit: `atomic.LoadUint32(&ac.count)`

util/counter.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+func (ac *AtomicCounter) IncreaseCount() {
+	ac.countLock.Lock()
+	defer ac.countLock.Unlock()
+
+	ac.count++
```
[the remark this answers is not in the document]
c3y1huang: @shuo-wu , @derekbit I don't know what is the benefit of using atomic here. Can you explain?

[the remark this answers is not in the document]
derekbit: Because the counter is only an integer, we can just simply use atomic operations to reduce the CPU cycles. The benefit should not be significant, so I mark it as nit.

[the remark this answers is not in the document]
innobead: Agreed. We don't need to do the same thing again. (I forgot there are atomic things in golang natively..)

metrics_collector/manager_collector.go
```diff
@@ -12,10 +12,12 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 
 	"github.com/longhorn/longhorn-manager/datastore"
+	"github.com/longhorn/longhorn-manager/util"
 )
 
 type ManagerCollector struct {
 	*baseCollector
+	proxyConnCounter util.Counter
```
shuo-wu, line 20: Does this collector require the counter?

shuo-wu (commented):
NIT: I prefer to use functions in pkg `atomic`

c3y1huang:
> Full regression test in-progress...`261a1d15741ef4a76433631ca606a50b81c37789`

Screenshot-2022-06-15-12:00:42

Full regression completed. Seems there is 1 connection not closing.

util/counter.go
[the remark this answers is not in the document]
c3y1huang: Updated.

metrics_collector/manager_collector.go
c3y1huang: Removed.

c3y1huang:
Coretest in-progress... `5e3b86af564bc3ee3c71c350208cc797a8d3628a`

shuo-wu (approved):

derekbit (approved):

c3y1huang:
Coretest in-progress ... `03f30d6fffcf697dfbc682ac1bca4cfc9d7c8217`

manager/volume.go
```diff
@@ -28,14 +28,18 @@ type VolumeManager struct {
 
 	currentNodeID string
 	sb            *SupportBundle
+
+	proxyConnCounter util.Counter
```
innobead, line 32: Do we use it?

c3y1huang: Yes, for example, https://github.com/c3y1huang/longhorn-manager/blob/grpc-proxy-connection-metric/manager/engine.go#L38

metrics_collector/instance_manager_collector.go
```diff
@@ -188,3 +211,40 @@ func (imc *InstanceManagerCollector) collectRequestValues(ch chan<- prometheus.M
...
+		imc.logger.WithError(err).Warn("error during scrape")
+		return
+	}
+
+	for _, im := range engineInstanceManagers {
+		imPod, err := imc.ds.GetPod(im.Name)
+		if err != nil {
+			if datastore.ErrorIsNotFound(err) {
+				logrus.WithError(err).Debugf("resetting proxy gRPC connection counter for %v", im.Name)
+				imc.proxyConnCounter.ResetCount()
```
innobead, line 233: When having 2 IM-E pods after Longhorn upgrade, there will be two cases:
- No differentiated metrics for each pod, so the reported metrics actually the same for them
- If the first pod not found, the counter will be reset so it will cause the other reported metrics probably incorrect

However, this can be improved later.

manager/volume.go
innobead: Thanks! I missed that.

innobead (approved):

metrics_collector/instance_manager_collector.go
c3y1huang: https://github.com/longhorn/longhorn/issues/4112

c3y1huang:
> Coretest in-progress ... `03f30d6fffcf697dfbc682ac1bca4cfc9d7c8217`

Screenshot-2022-06-15-14:26:07

PR #1381, opened 2022-06-14
grpc-proxy: connection leak when failed to backup

c3y1huang:
https://github.com/longhorn/longhorn/issues/4110


innobead (dismissed):

innobead:
nit: You can just use `m.Close()` instead.

innobead (approved):

PR #1389, opened 2022-06-20
controller: Fix the misleading log for the replicas that have no StorageIP

shuo-wu:
Longhorn/longhorn#4153


innobead (approved):

derekbit (approved):

innobead:
Need to backport to 1.3.1.

PR #1394, opened 2022-06-21
Avoid validation failure while syncing configmap to setting CRs

derekbit:
https://github.com/longhorn/longhorn/issues/4160
https://github.com/longhorn/longhorn/issues/4163


innobead (approved):
LGTM

PR #1395, opened 2022-06-28
Make sure the VolumeAttachment has gone first before force delete the statefulset/deployment  pod

PhanLe1010:
This fix prevents the pod from hitting the backoff period with error
```
Multi-Attach error for volume "pvc-51e196b9-4f93-42a2-b362-b744fa05b07f" Volume is already exclusively attached to one node and can't be attached to another
```
Thus, speeding up the recovery time when a node is turned off.

longhorn/longhorn#2947


PhanLe1010:
Steps:
1. Deploy a sts of cale 1 that use PVC provisioned by Longhorn
2. Set `Pod Deletion Policy When Node is Down` setting to `delete-both-statefulset-and-deployment-pod`
3. Find and turn off the node where the pod is running on. 
4. Observe how long it takes for the pod to recover 


### Before the fix

| Event | Elapsed Time | Additional Info | 
|---|---|---|
| Node become note ready | 0:37  | | 
| Pod enter terminating (has the deletion timestamp) | 5:54  | | 
| Pod is force deleted | 6:32  | |
| Pod is recreate and in ContainerCreating | 6:32 | ``` Multi-Attach error for volume "pvc-51e196b9-4f93-42a2-b362-b744fa05b07f" Volume is already exclusively attached to one node and can't be attached to another ``` |
| VA is deleted | 12:34  | |
| New VA is created | 12:34  | | 
| Pod is running | 14:50 | |

### After the fix

| Event | Elapsed Time |
|---|---|
| Node become note ready | 0:37  |
| Pod enter terminating (has the deletion timestamp) | 5:44  |
| VA is deleted | 6:34  |
| Pod is force deleted | 6:38  |
| Pod is recreate and in ContainerCreating | 6:38 |
| New VA is created | 7:15  |
| Pod is running | 7:45s |

innobead (commented):
Is this related to https://github.com/longhorn/longhorn-manager/commit/25ccde1b79c62358d9f9cc2ec0c0d8ef8987fe59 (we removed the volume attachment cleanup mechanism before, but rather let kubelet to handle the lifecycle of pod deletion)?

cc @joshimoo @shuo-wu

controller/kubernetes_pod_controller.go
```diff
@@ -199,6 +200,28 @@ func (kc *KubernetesPodController) handlePodDeletionIfNodeDown(pod *v1.Pod, node
...
+			if err != nil {
+				if datastore.ErrorIsNotFound(err) {
+					continue
+				}
+				return err
+			}
+			kc.logger.Infof("%v: deleted volume attachment %v for pod %v on downed node %v", controllerAgentName, va.Name, pod.Name, nodeID)
+			return err
+		}
+		// wait the volumeattachment object to be deleted
```
innobead, line 221: Probably we can have a log here to know the pod is not deleted yet because the VAs are still under deletion.

innobead (approved):
LGTM

c3y1huang (approved):
LGTM

derekbit (dismissed):
LGTM

controller/kubernetes_pod_controller.go
PhanLe1010: Thank you. Fixed

controller/kubernetes_pod_controller.go
```diff
@@ -199,6 +200,28 @@ func (kc *KubernetesPodController) handlePodDeletionIfNodeDown(pod *v1.Pod, node
 		return nil
 	}
```
shuo-wu, line 201: Should we move the condition check `pod.DeletionTimestamp.After(time.Now())` in line 198/199 to the next of the VA deletion?

PhanLe1010: Could you explain a bit more about why?

shuo-wu: When this check is passed, it means Longhorn decides to delete or is already deleting the pod forcefully. But this PR aims to do the VA deletion before the pod force deletion.
The best timing for the VA deletion is after the pod deletion timestamp is set but before the force deletion starts.

innobead: @PhanLe1010 any concerns?

PhanLe1010: Make sense that VA deletion is after the pod deletion timestamp is set but before the force deletion starts! Thank you for the explanation! I pushed the change

shuo-wu (approved):

innobead (approved):

PR #1398, opened 2022-07-05
fix a crush bug

DaleWYZ:
If Longhorn config the Storage Network, and the  cni is use ovs-cni, in the instance-manager pod there is no ip address for the cni net. in this situation net.IPs is nil, and return net.IPs[0] will make a panic, and the longhorn-manager pod will always in the CrashLoopBackOff state, so if we add a check of the net.IPs will solve the problem.


c3y1huang (approved):
LGTM. We also need this in the master branch.

PR #1400, opened 2022-07-11
Update Longhorn client to send appVersion and extra.kubernetesVersioninstead of the deprecated LonghornVersion and KubernetesVersion

PhanLe1010:
longhorn/longhorn#3093

### DIFF
--- a/controller/setting_controller.go
+++ b/controller/setting_controller.go
@@ -83,8 +83,8 @@ type Version struct {
 }
 
 type CheckUpgradeRequest struct {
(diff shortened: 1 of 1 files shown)


innobead (commented):
I assume this change is backward compatible and will not impact the existing releases, right?

innobead (approved):

PhanLe1010:
Yes, you are correct. The server has been updated to handle both existing Long deployment and this new format

PR #1405, opened 2022-07-13
Allow update node resources if kubernetes node is already gone

derekbit:
Previously, if the kubernetes node is gone, the node validating webhook did
not allow the modification of the node resource. It hindered the remove of the
finalizers and resulted in the deletion failure of the node resource.

[Longhorn 4213](https://github.com/longhorn/longhorn/issues/4213)


innobead (dismissed):
LGTM

innobead (approved):

innobead:
@derekbit Please backport this to 1.3.x. Thanks.

PR #1409, opened 2022-07-15
replica: do not check storage IP when in deletion

c3y1huang:
When the replica controller node is down during a replica deletion, another replica controller should pick up and finish the flow. However, before removing the finalizer, the new replica controller could get blocked by the ownership change when the storage IP is not in sync.

https://github.com/longhorn/longhorn/issues/4213#issuecomment-1184955056

### DIFF
--- a/datastore/longhorn.go
+++ b/datastore/longhorn.go
@@ -1034,9 +1034,17 @@ func checkReplica(r *longhorn.Replica) error {
 	if r.Name == "" || r.Spec.VolumeName == "" {
 		return fmt.Errorf("BUG: missing required field %+v", r)
 	}
+
 	if (r.Status.CurrentState == longhorn.InstanceStateRunning) != (r.Status.IP != "") {
 		return fmt.Errorf("BUG: instance state and IP wasn't in sync %+v", r)
 	}
+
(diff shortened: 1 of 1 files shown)


derekbit (approved):
LGTM

Do you know how to trigger the mismatching storageIp and Ip error?

innobead:
@c3y1huang How frequent the replica's storage IP will be updated? So here, it's about the replica controller gone due to the node gone, so the storage IP is not updated yet. Right?

c3y1huang:
I don't know how to trigger mismatching storage IP.

Replica/engine storage IP gets updated at state change. So when the state changes to running, the replica controller will get the IP from the IM pod and then set it to the `status.StorageIP`. This is also where the `status.IP` gets set.

https://github.com/longhorn/longhorn-manager/blob/master/controller/instance_handler.go#L150-L159

c3y1huang:
@innobead has pointed out that it could be possible replica controller has not reached the `storageIP`/`IP` assignment after the upgrade. Thus there is no log in the support bundles regards to replica storage IP change.

Closing this PR in favor of https://github.com/longhorn/longhorn-manager/pull/1411

PR #1410, opened 2022-07-15
Retry restoring Backup if getting lock failed.

mantissahz:
Just give a couple times to retry restoreBackup only when getting lock failed.

longhorn/longhorn#3055

### DIFF
--- a/controller/engine_controller.go
+++ b/controller/engine_controller.go
@@ -38,7 +38,8 @@ import (
 )
 
 const (
-	unknownReplicaPrefix = "UNKNOWN-"
+	unknownReplicaPrefix    = "UNKNOWN-"
+	restoreGetLockFailedMsg = "error initiating full backup restore: failed lock"
 )
 
 var (
@@ -48,6 +49,9 @@ var (
 	EngineMonitorConflictRetryCount = 5
 
 	purgeWaitIntervalInSecond = 24 * 60 * 60
+
+	// restoreMaxInterval: deleting the backup of big size volume takes a long time for retain policy and restoring backups would be in backoff period.
+	restoreMaxInterval = 1 * time.Hour
 )
 
 const (
@@ -91,6 +95,7 @@ type EngineMonitor struct {
 	engines          engineapi.EngineClientCollection
 	stopCh           chan struct{}
 	expansionBackoff *flowcontrol.Backoff
+	restoreBackoff   *flowcontrol.Backoff
 
 	controllerID string
 	// used to notify the controller that monitoring has stopped
@@ -625,6 +630,7 @@ func (ec *EngineController) startMonitoring(e *longhorn.Engine) {
 		stopCh:                 stopCh,
 		monitorVoluntaryStopCh: monitorVoluntaryStopCh,
 		expansionBackoff:       flowcontrol.NewBackOff(time.Second*10, time.Minute*5),
+		restoreBackoff:         flowcontrol.NewBackOff(time.Second*10, restoreMaxInterval),
 		controllerID:           ec.controllerID,
 		proxyConnCounter:       ec.proxyConnCounter,
 	}
@@ -938,9 +944,15 @@ func (m *EngineMonitor) refresh(engine *longhorn.Engine) error {
 
 	// Incremental restoration will implicitly expand the DR volume once the backup volume is expanded
 	if needRestore {
-		if err = restoreBackup(m.logger, engine, rsMap, cliAPIVersion, m.ds, engineClientProxy); err != nil {
+		if m.restoreBackoff.IsInBackOffSinceUpdate(engine.Name, time.Now()) {
+			m.logger.Debugf("Cannot restore the backup for engine %v since it is still in the backoff window", engine.Name)
+			return nil
+		}
+		if err = m.restoreBackup(engine, rsMap, cliAPIVersion, engineClientProxy); err != nil {
+			m.restoreBackoff.DeleteEntry(engine.Name)
 			return err
 		}
+
 	}
 
 	var snapshotCloneStatusMap map[string]*longhorn.SnapshotCloneStatus
@@ -1191,9 +1203,9 @@ func checkSizeBeforeRestoration(log logrus.FieldLogger, engine *longhorn.Engine,
 	return true, nil
 }
 
-func restoreBackup(log logrus.FieldLogger, engine *longhorn.Engine, rsMap map[string]*longhorn.RestoreStatus, cliAPIVersion int, ds *datastore.DataStore, engineClientProxy engineapi.EngineClientProxy) error {
+func (m *EngineMonitor) restoreBackup(engine *longhorn.Engine, rsMap map[string]*longhorn.RestoreStatus, cliAPIVersion int, engineClientProxy engineapi.EngineClientProxy) error {
 	// Get default backup target
-	backupTarget, err := ds.GetBackupTargetRO(types.DefaultBackupTargetName)
+	backupTarget, err := m.ds.GetBackupTargetRO(types.DefaultBackupTargetName)
 	if err != nil {
 		if !apierrors.IsNotFound(err) {
 			return err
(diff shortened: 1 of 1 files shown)


controller/engine_controller.go
```diff
@@ -1199,22 +1204,44 @@ func restoreBackup(log logrus.FieldLogger, engine *longhorn.Engine, rsMap map[st
...
+	// Register the name with a backkofentry
+	backoff.Next(engine.Name, startTime)
 	if cliAPIVersion < engineapi.CLIVersionFour {
 		// For compatible engines, `LastRestoredBackup` is required to indicate if the restore is incremental restore
 		log.Infof("Prepare to restore backup, backup target: %v, backup volume: %v, requested restored backup name: %v, last restored backup name: %v",
 			backupTargetClient.URL, engine.Spec.BackupVolume, engine.Spec.RequestedBackupRestore, engine.Status.LastRestoredBackup)
-		if err = engineClientProxy.BackupRestore(engine, backupTargetClient.URL, engine.Spec.RequestedBackupRestore, engine.Spec.BackupVolume, engine.Status.LastRestoredBackup, backupTargetClient.Credential); err != nil {
-			if extraErr := handleRestoreErrorForCompatibleEngine(log, engine, rsMap, err); extraErr != nil {
-				return extraErr
+		for {
```
shuo-wu, line 1215: You cannot use a for loop (with sleep) here since it will block the sync between the engine CR and the engine process.
The whole monitor sync function will be executed every several seconds. You can directly use the backoff function here: If it's in the backoff window, directly skip this part.

shuo-wu (changes requested):

controller/engine_controller.go
mantissahz: OK, changed, add `restoreBackoff` into EngineMonitor to record the backoff.
Do you think `restoreBackup` could be an EngineMonitor method?

controller/engine_controller.go
```diff
@@ -1199,24 +1201,37 @@ func restoreBackup(log logrus.FieldLogger, engine *longhorn.Engine, rsMap map[st
...
 	if cliAPIVersion < engineapi.CLIVersionFour {
 		// For compatible engines, `LastRestoredBackup` is required to indicate if the restore is incremental restore
 		log.Infof("Prepare to restore backup, backup target: %v, backup volume: %v, requested restored backup name: %v, last restored backup name: %v",
 			backupTargetClient.URL, engine.Spec.BackupVolume, engine.Spec.RequestedBackupRestore, engine.Status.LastRestoredBackup)
 		if err = engineClientProxy.BackupRestore(engine, backupTargetClient.URL, engine.Spec.RequestedBackupRestore, engine.Spec.BackupVolume, engine.Status.LastRestoredBackup, backupTargetClient.Credential); err != nil {
+			// only handle the error that is "faield lock"
+			if strings.Contains(err.Error(), "failed lock") && backoff.IsInBackOffSince(engine.Name, time.Now()) {
+				log.Warnf("restoring backup %v failed: %v", engine.Spec.BackupVolume, err)
+				return nil
+			}
```
shuo-wu, line 1215: Why not put this error handling inside function `handleRestoreError`?

controller/engine_controller.go
shuo-wu: Yes you can do it.

controller/engine_controller.go
```diff
@@ -1199,24 +1201,37 @@ func restoreBackup(log logrus.FieldLogger, engine *longhorn.Engine, rsMap map[st
...
 			if extraErr := handleRestoreErrorForCompatibleEngine(log, engine, rsMap, err); extraErr != nil {
 				return extraErr
 			}
 		}
 	} else {
+		log.Infof("Prepare to restore backup, backup target: %v, backup volume: %v, requested restored backup name: %v",
+			backupTargetClient.URL, engine.Spec.BackupVolume, engine.Spec.RequestedBackupRestore)
```
shuo-wu, line 1222: Maybe we can use a single `log.Infof` before `if cliAPIVersion < engineapi.CLIVersionFour {`.

controller/engine_controller.go
mantissahz: Changed, thanks

controller/engine_controller.go
mantissahz: Yes, move it into the function `handleRestoreError`, thanks

controller/engine_controller.go
mantissahz: Moved, thanks

shuo-wu (commented):
Is this backoff designed to limit the restore retrying when there are locking failure errors?
If YES, then you need to:
1. Check the backoff window before sending the requests.
2. Add the entry only when the locking failure is hit
3. Delete the entry when there is no need to retry the restore:
    1. The restore succeeds.
    2. The error cannot be ignored.

mantissahz:
@shuo-wu Modified the flow as you mentioned. Thanks.
Please kindly review it again when you are available.

controller/engine_controller.go
```diff
@@ -1184,44 +1195,45 @@ func checkSizeBeforeRestoration(log logrus.FieldLogger, engine *longhorn.Engine,
...
-				backupTargetClient.URL, engine.Spec.BackupVolume, engine.Spec.RequestedBackupRestore)
-			if extraErr := handleRestoreError(log, engine, rsMap, err); extraErr != nil {
+			if extraErr := handleRestoreError(m.logger, engine, rsMap, m.restoreBackoff, err); extraErr != nil {
 				return extraErr
 			}
 		}
 	}
+	if err == nil || !strings.Contains(err.Error(), restoreGetLockFailedMsg) {
+		m.restoreBackoff.DeleteEntry(engine.Name)
+	}
```
shuo-wu, line 1231: NIT: Can we handle this in `handleRestoreError` or `handleRestoreErrorForCompatibleEngine`?
(The code for compatible engines will be removed in the future hence we don't need to worry about duplication)

mantissahz: Here delete the backupoff entry for successfully restoring backup, and it will not get into `handleRestoreError` or `handleRestoreErrorForCompatibleEngine` if `err == nil`, so I don't think this judgement can be moved into those two functions. `!strings.Contains(err.Error(), restoreGetLockFailedMsg` check for other two skipped error types.

shuo-wu: Got it! Make sense.

shuo-wu (approved):

controller/engine_controller.go
```diff
@@ -48,6 +49,8 @@ var (
 	EngineMonitorConflictRetryCount = 5
 
 	purgeWaitIntervalInSecond = 24 * 60 * 60
+
+	restoreMaxInterval = 1 * time.Hour
```
derekbit, line 53: Just curious. How to determine the maximum interval?

controller/engine_controller.go
```diff
@@ -1230,8 +1242,12 @@ func handleRestoreError(log logrus.FieldLogger, engine *longhorn.Engine, rsMap m
 
 	for _, re := range taskErr.ReplicaErrors {
 		if status, exists := rsMap[re.Address]; exists {
-			if strings.Contains(re.Error(), "already in progress") || strings.Contains(re.Error(), "already restored backup") {
+			if strings.Contains(re.Error(), "already in progress") || strings.Contains(re.Error(), "already restored backup") || strings.Contains(re.Error(), restoreGetLockFailedMsg) {
 				log.WithError(re).Debugf("Ignore restore error from replica %v", re.Address)
+				if strings.Contains(re.Error(), restoreGetLockFailedMsg) {
+					// Register the name with a backkofentry
```
c3y1huang, line 1248: `backoff entry`?

controller/engine_controller.go
```diff
@@ -1230,8 +1242,12 @@ func handleRestoreError(log logrus.FieldLogger, engine *longhorn.Engine, rsMap m
 
 	for _, re := range taskErr.ReplicaErrors {
 		if status, exists := rsMap[re.Address]; exists {
-			if strings.Contains(re.Error(), "already in progress") || strings.Contains(re.Error(), "already restored backup") {
+			if strings.Contains(re.Error(), "already in progress") || strings.Contains(re.Error(), "already restored backup") || strings.Contains(re.Error(), restoreGetLockFailedMsg) {
 				log.WithError(re).Debugf("Ignore restore error from replica %v", re.Address)
+				if strings.Contains(re.Error(), restoreGetLockFailedMsg) {
```
c3y1huang, line 1247: Suggest to have a separate if statement instead of wrapping in the `if strings.Contains(re.Error(), "already in progress") || strings.Contains(re.Error(), "already restored backup") || strings.Contains(re.Error(), restoreGetLockFailedMsg)`.

controller/engine_controller.go
mantissahz: yap, I think there would be some big size volume to backup for the recurring job and when it has to be deleted for the retain policy, deletion might take long time but usually deletion of backups would take a couple of minutes in my test environment (2G data).
`1 hour` is a temporary value and it could be modified if any suggestion.

controller/engine_controller.go
```diff
@@ -1184,44 +1195,45 @@ func checkSizeBeforeRestoration(log logrus.FieldLogger, engine *longhorn.Engine,
...
 		return fmt.Errorf("cannot found the %s backup target", types.DefaultBackupTargetName)
 	}
 
-	backupTargetClient, err := getBackupTargetClient(ds, backupTarget)
+	backupTargetClient, err := getBackupTargetClient(m.ds, backupTarget)
 	if err != nil {
 		return errors.Wrapf(err, "cannot get backup target config for backup restoration of engine %v", engine.Name)
 	}
 
+	m.logger.Infof("Prepare to restore backup, backup target: %v, backup volume: %v, requested restored backup name: %v, last restored backup name: %v",
```
c3y1huang, line 1213: Nit: suggest using `WithFields`.

controller/engine_controller.go
derekbit: Got it. Can you add a comment here?

mantissahz: Added. Thanks.

controller/engine_controller.go
mantissahz: Modified. More clear. Thanks.

controller/engine_controller.go
mantissahz: Modified.

c3y1huang (approved):

derekbit (approved):

shuo-wu (approved):

controller/engine_controller.go
```diff
@@ -48,6 +49,9 @@ var (
 	EngineMonitorConflictRetryCount = 5
 
 	purgeWaitIntervalInSecond = 24 * 60 * 60
+
+	// restoreMaxInterval: deleting the backup of big size volume takes a long time for retain policy and restoring backups would be in backoff period.
+	restoreMaxInterval = 1 * time.Hour
```
innobead, line 54: nit: use constant instead of var here. (it seems all variables here should be in const instead)

controller/engine_controller.go
```diff
@@ -1237,6 +1255,14 @@ func handleRestoreError(log logrus.FieldLogger, engine *longhorn.Engine, rsMap m
 
 	for _, re := range taskErr.ReplicaErrors {
 		if status, exists := rsMap[re.Address]; exists {
+			if strings.Contains(re.Error(), restoreGetLockFailedMsg) {
+				// Register the name with a restore backoff entry
+				log.WithError(re).Debugf("Ignore failed locked restore error from replica %v", re.Address)
```
innobead, line 1260: nit: for logging, suggest below patterns instead of using the general verb, because it will confuse people if the action has been taken or not. 

- `verb+ing` before the action
- `verb+ed` after the action
- `failed to verb` failed the action

controller/engine_controller.go
innobead: also, I think this kind of implicit timeout should be documented to let users better know what happens under the hood. (the same as expansion backup)

innobead:
@mantissahz Please backport to 1.3.2 and 1.2.5. Thanks.

PR #1411, opened 2022-07-15
upgrade: autofill replica storage IP

c3y1huang:
When unexpectedly removed a node during an upgrade from v1.2.x to v1.3.0. There are chances of not reaching replica storage IP assignment. In this case, other controllers will get blocked by the `checkReplica` during the ownership change.

https://github.com/longhorn/longhorn/issues/4213#issuecomment-1184955056

### DIFF
--- a/upgrade/v12xto130/upgrade.go
+++ b/upgrade/v12xto130/upgrade.go
@@ -1,0 +1,59 @@
+package v12xto130
+
+import (
+	"context"
+
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	clientset "k8s.io/client-go/kubernetes"
+
+	lhclientset "github.com/longhorn/longhorn-manager/k8s/pkg/client/clientset/versioned"
+)
+
+const (
+	upgradeLogPrefix = "upgrade from v1.2.x to v1.3.0: "
+)
+
+func UpgradeResources(namespace string, lhClient *lhclientset.Clientset, kubeClient *clientset.Clientset) error {
+	if err := upgradeReplicas(namespace, lhClient); err != nil {
+		return err
+	}
+	return nil
+}
+
+func upgradeReplicas(namespace string, lhClient *lhclientset.Clientset) (err error) {
+	defer func() {
+		err = errors.Wrapf(err, upgradeLogPrefix+"upgrade replica failed")
+	}()
+
+	replicas, err := lhClient.LonghornV1beta2().Replicas(namespace).List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil
+		}
+		return errors.Wrapf(err, "failed to list all existing Longhorn replicas during the replica upgrade")
+	}
+
+	for _, r := range replicas.Items {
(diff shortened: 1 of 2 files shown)


innobead (dismissed):
LGTM

upgrade/v12xto130/upgrade.go
```diff
@@ -0,0 +1,60 @@
...
+		if r.Status.StorageIP != "" {
+			continue
+		}
+
+		r.Status.StorageIP = r.Status.IP
+		_, err := lhClient.LonghornV1beta2().Replicas(namespace).Update(context.TODO(), &r, metav1.UpdateOptions{})
+		if err != nil {
+			if apierrors.IsConflict(errors.Cause(err)) {
+				return nil
+			}
```
shuo-wu, line 54: Why does the function directly return? Will this make the related replicas no longer updatable?

shuo-wu (commented):
Besides, will we move [this function](https://github.com/longhorn/longhorn-manager/blob/2c9fb0b9c81ae4b6d2b905e5ebde2c0d179da5a5/datastore/longhorn.go#L1033-L1044) to the validator webhook?

upgrade/v12xto130/upgrade.go
c3y1huang: Fixed, thanks!

c3y1huang:
> Besides, will we move [this function](https://github.com/longhorn/longhorn-manager/blob/2c9fb0b9c81ae4b6d2b905e5ebde2c0d179da5a5/datastore/longhorn.go#L1033-L1044) to the validator webhook?

How does it help with the issue?

shuo-wu:
> ides, will we move [this function](https://github.com/longhorn/longhorn-manager/blob/2c9fb0b9c81ae4b6d2b905e5ebde2c0d179da5a5/datastore/longhorn.go#L1033-L1044) to the validator webhook?

No. I just suddenly noticed that. We can handle it later.

upgrade/v12xto130/upgrade.go
innobead: I think it is possible. Have a retry for at least the best effort? or just have a warning log here.

c3y1huang: Updated.

upgrade/v12xto130/upgrade.go
```diff
@@ -0,0 +1,72 @@
...
+
+			if r.Status.StorageIP != "" {
+				continue
+			}
+
+			r.Status.StorageIP = r.Status.IP
+			_, err = lhClient.LonghornV1beta2().Replicas(namespace).Update(context.TODO(), &r, metav1.UpdateOptions{})
+			if err != nil {
+				retry = true
```
innobead, line 60: Thanks for the prompt update!

However, I am thinking probably we should make it simply failed, and then users can wait for the next manager pod to restart to retry the upgrade because the upgrade should be idempotent instead of using a specific retry in this upgradeReplica function. 

WDYT?

c3y1huang: We can do that but I am not sure if it is worth going through the Kubernetes pod lifecycle when caused by conflicting errors. Maybe we can still retry because this will expire quickly(1m) and return an error when failed to update?

c3y1huang: Updated.

innobead (approved):

innobead:
@c3y1huang Please backport to 1.3.1. Thanks.

PR #1412, opened 2022-07-15
We want to reuse replica on the disk that is unschedulable due to allocated space being bigger than max allocable space but the disk is not full yet

PhanLe1010:
longhorn/longhorn#3921


innobead (approved):
LGTM

PR #1420, opened 2022-07-20
Set replics.spec.failedAt when a replica failed at attaching stage before engine become running

PhanLe1010:
This allow Longhorn to detach the volume and retry attaching the volume thus fix the deadlock situation in the issue description

longhorn/longhorn#4197


innobead (approved):
LGTM. 

note: If we can add unit cases about the state transition for volume, engine, and replica, it's possible to capture more corner cases like this. (can review and do it later)

c3y1huang (approved):
LGTM

PR #1427, opened 2022-07-27
fix: replica failed to rebuild during disk evict

c3y1huang:
Skip pre-rebuild snapshot purge if the replica is unused to allow for the rebuild.

https://github.com/longhorn/longhorn/issues/4238
https://github.com/longhorn/longhorn/issues/4183

### DIFF
--- a/controller/engine_controller.go
+++ b/controller/engine_controller.go
@@ -1447,8 +1447,15 @@ func (ec *EngineController) startRebuilding(e *longhorn.Engine, replica, addr st
 		}
 		defer engineClientProxy.Close()
 
+		rep, err := ec.ds.GetReplica(replica)
+		if err != nil {
+			log.WithError(err).Errorf("Failed to get rebuilding replica %v", replica)
+			ec.eventRecorder.Eventf(e, v1.EventTypeWarning, EventReasonFailedRebuilding, "Failed to get rebuilding replica %v: %v", addr, err)
+			return
+		}
+
 		// If enabled, call and wait for SnapshotPurge to clean up system generated snapshot before rebuilding.
-		if autoCleanupSystemGeneratedSnapshot {
+		if autoCleanupSystemGeneratedSnapshot && rep.Spec.HealthyAt != "" {
 			if err := engineClientProxy.SnapshotPurge(e); err != nil {
 				log.WithError(err).Error("Failed to start snapshot purge before rebuilding")
 				ec.eventRecorder.Eventf(e, v1.EventTypeWarning, EventReasonFailedStartingSnapshotPurge, "Failed to start snapshot purge for engine %v and volume %v before rebuilding: %v", e.Name, e.Spec.VolumeName, err)


c3y1huang:
Test result: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/1712/

controller/engine_controller.go
```diff
@@ -1447,8 +1447,15 @@ func (ec *EngineController) startRebuilding(e *longhorn.Engine, replica, addr st
...
+		rep, err := ec.ds.GetReplica(replica)
+		if err != nil {
+			log.WithError(err).Errorf("Failed to get rebuilding replica %v", replica)
+			ec.eventRecorder.Eventf(e, v1.EventTypeWarning, EventReasonFailedRebuilding, "Failed to get rebuilding replica %v: %v", addr, err)
+			return
+		}
+
 		// If enabled, call and wait for SnapshotPurge to clean up system generated snapshot before rebuilding.
-		if autoCleanupSystemGeneratedSnapshot {
+		if autoCleanupSystemGeneratedSnapshot && rep.Spec.HealthyAt != "" {
```
innobead, line 1458: @shuo-wu For snapshot purge, should we handle this check on the engine side? so the manager doesn't need to do any validation for this but rather rely on the error returned from `SnapshotPurge`?

innobead:
cc @longhorn/qa this probably impacts recent failed cases.

controller/engine_controller.go
shuo-wu: Does this really resolve the issue? If Longhorn sends purge requests to the replica with empty `Spec.HealthyAt`, the cmd should fail and this engine controller will retry it. As a result, the overall rebuilding progress is just delayed a little bit.

innobead: After discussing with @shuo-wu , this should be finally reconciled, so it should not make the volume pending in attaching state. 

@cchien816 Could you clarify this? Thanks.

shuo-wu: The reason why the volume gets attaching/detaching:
1. the eviction mechanism blindly removes the healthy replica on the node the test pod resides on while the test directly removes 2 healthy replicas on other nodes.
2. the test tries to r/w data from the volume. But at this moment the volume just contains a rebuilding replica, which is invalid for IO. Then the engine marks it as ERR after the IO failure
3. Longhorn somehow deleted this only ERR replica. As a result, the volume does not contain any replica and Longhorn wrongly marks the volume as attaching or detaching state.

cchien816: Using PR code base to run `test_disk_eviction_with_node_level_soft_anti_affinity_disabled`, test case pass and yes , volume not pending in attaching state

innobead: As discussed with @shuo-wu , this seems like the day 1 issue instead of causing by the `purge-before-rebuild` introduction, but we don't know if this did not fail before, probably because of fast rebuilding before evicting the last replica?

cchien816: For referenceUsing PR code base to run some new fail cases in master [pipeline](https://ci.longhorn.io/job/public/job/master/job/sles/job/amd64/job/longhorn-tests-sles-amd64/199/#showFailuresLink)

- test_rebuild_with_restoration
- test_engine_crash_for_restore_volume
- test_engine_crash_for_dr_volume

innobead: Then, they are issues with other causes which we need to identify if they are not flaky. Please create issues for them after the analysis. Thanks, @cchien816

c3y1huang:
Close in favor of https://github.com/longhorn/longhorn-manager/pull/1433.

PR #1428, opened 2022-07-27
Backup stays in `InProgress` if volume deleted.

mantissahz:
Set the backup state `Error` if backup volume was deleted before
the backup finished.

longhorn/longhorn#3122

### DIFF
--- a/controller/backup_controller.go
+++ b/controller/backup_controller.go
@@ -330,7 +330,11 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
 			if !apierrors.IsNotFound(err) {
 				return err
 			}
-			log.WithError(err).Error("Cannot found the corresponding volume")
+			err = fmt.Errorf("Cannot find the corresponding volume: %v", err)
+			log.WithError(err).Error()
+			backup.Status.Error = err.Error()
+			backup.Status.State = longhorn.BackupStateError
+			backup.Status.LastSyncedAt = syncTime
 			return nil // Ignore error to prevent enqueue
 		}
 


derekbit (dismissed):
LGTM.

controller/backup_controller.go
```diff
@@ -331,6 +331,8 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
 				return err
 			}
 			log.WithError(err).Error("Cannot found the corresponding volume")
+			backup.Status.State = longhorn.BackupStateError
```
innobead, line 334: Suggest updating `Status.Error` as well to better know the cause. You can leverage the above log.

innobead (changes requested):

controller/backup_controller.go
mantissahz: Updated, thanks.

innobead: nit: it's better to reuse the err and customized message instead of just repeating the same message. 

```golang
err = fmt.Errorf("cannot found the corresponding volume: %v", err)
log.WithError(err).Error()
backup.Status.Error = err.Error()
```

or if you want to keep the original err + the customized message when using log
```golang
errMsg := "Cannot found the corresponding volume"
log.WithError(err).Error(errMsg)
backup.Status.Error = errMsg
```

mantissahz: Thanks, changed.

innobead (approved):

innobead:
@mantissahz Please help backport to 1.3.x and 1.2.x. Thanks.

PR #1432, opened 2022-07-27
fix 4302: avoid repeat deletion of pod

w13915984028:
Issue:
https://github.com/longhorn/longhorn/issues/4302

check `pod.DeletionTimestamp` first.

### DIFF
--- a/controller/kubernetes_pod_controller.go
+++ b/controller/kubernetes_pod_controller.go
@@ -310,6 +310,12 @@ func (kc *KubernetesPodController) handlePodDeletionIfVolumeRequestRemount(pod *
 	if pod.Status.StartTime == nil {
 		return nil
 	}
+
+	// Avoid repeat deletion
+	if pod.DeletionTimestamp != nil {
+		return nil
+	}
+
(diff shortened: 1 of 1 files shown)


joshimoo:
hold merge for @PhanLe1010 review

PhanLe1010:
This PR will fix the multiple deletion log messages. It looks good to me

However, the flow will remain the same because multiple deletion requests are collapsed into the first deletion https://github.com/kubernetes/kubernetes/issues/71494. So the behavior when volume crashes will remain the same.

joshimoo (dismissed):
LGTM, as mentioned this only skips the additional delete requests against the API server, even before though sending another delete request against a resource that is in the progress of deletion is a NOOP unless the grace period is set lower i.e. 0

PhanLe1010 (approved):

w13915984028:
local verification is mentioned in: https://github.com/longhorn/longhorn/issues/4302#issuecomment-1202118521

innobead (approved):

innobead:
@w13915984028 Please help backport to 1.3.x and 1.2.x branches.

PR #1433, opened 2022-07-28
Avoid evicting or removing the last healthy replica

shuo-wu:
https://github.com/longhorn/longhorn/issues/4238

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -848,18 +848,40 @@ func (vc *VolumeController) cleanupExtraHealthyReplicas(v *longhorn.Volume, e *l
 }
 
 func (vc *VolumeController) cleanupEvictionRequestedReplicas(v *longhorn.Volume, rs map[string]*longhorn.Replica) (bool, error) {
+	log := getLoggerForVolume(vc.logger, v)
+
+	// If there is no non-evicting healthy replica,
+	// Longhorn should retain one evicting healthy replica.
+	hasNonEvictingHealthyReplica := false
+	evictingHealthyReplica := ""
 	for _, r := range rs {
-		if r.Status.EvictionRequested {
-			if err := vc.deleteReplica(r, rs); err != nil {
-				vc.eventRecorder.Eventf(v, v1.EventTypeWarning,
-					EventReasonFailedEviction,
-					"volume %v failed to evict replica %v",
-					v.Name, r.Name)
-				return false, err
-			}
-			logrus.Debugf("Evicted replica %v", r.Name)
-			return true, nil
+		if !datastore.IsAvailableHealthyReplica(r) {
+			continue
 		}
+		if !r.Status.EvictionRequested {
+			hasNonEvictingHealthyReplica = true
+			break
+		}
+		evictingHealthyReplica = r.Name
+	}
+
+	for _, r := range rs {
+		if !r.Status.EvictionRequested {
+			continue
+		}
+		if !hasNonEvictingHealthyReplica && r.Name == evictingHealthyReplica {
+			log.Warnf("Cannot evicting replica %v for now since there is no other healthy replica", r.Name)
+			continue
+		}
+		if err := vc.deleteReplica(r, rs); err != nil {
+			vc.eventRecorder.Eventf(v, v1.EventTypeWarning,
+				EventReasonFailedEviction,
+				"volume %v failed to evict replica %v",
+				v.Name, r.Name)
+			return false, err
+		}
+		log.Debugf("Evicted replica %v in disk %v of node %v ", r.Name, r.Spec.DiskID, r.Spec.NodeID)
+		return true, nil
 	}
 	return false, nil
 }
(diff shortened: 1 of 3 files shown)


controller/volume_controller.go
```diff
@@ -848,18 +848,40 @@ func (vc *VolumeController) cleanupExtraHealthyReplicas(v *longhorn.Volume, e *l
...
 		}
+		evictingHealthyReplica = r.Name
+	}
+
+	for _, r := range rs {
+		if !r.Status.EvictionRequested {
+			continue
+		}
+		if !hasNonEvictingHealthyReplicas && r.Name == evictingHealthyReplica {
+			log.Debugf("Cannot evicting replica %v for now since there is no other healthy replica", r.Name)
```
derekbit, line 873: Should we use `log.Warnf` here?

manager/volume.go
```diff
@@ -743,18 +743,22 @@ func (m *VolumeManager) DeleteReplica(volumeName, replicaName string) error {
...
 	}
-	if !hasHealthyReplicas {
+	if healthyReplica == "" {
 		return fmt.Errorf("no other healthy replica available, cannot delete replica %v since it may still contain data for recovery", replicaName)
 	}
 	if err := m.ds.DeleteReplica(replicaName); err != nil {
 		return err
 	}
-	logrus.Debugf("Deleted replica %v of volume %v", replicaName, volumeName)
+	logrus.Debugf("Deleting replica %v of volume %v, there is still at least one available healthy replica %v", replicaName, volumeName, healthyReplica)
```
c3y1huang, line 761:
```suggestion
	logrus.Debugf("Deleted replica %v of volume %v, there is still at least one available healthy replica %v", replicaName, volumeName, healthyReplica)
```

controller/volume_controller.go
```diff
@@ -848,18 +848,40 @@ func (vc *VolumeController) cleanupExtraHealthyReplicas(v *longhorn.Volume, e *l
 }
 
 func (vc *VolumeController) cleanupEvictionRequestedReplicas(v *longhorn.Volume, rs map[string]*longhorn.Replica) (bool, error) {
+	log := getLoggerForVolume(vc.logger, v)
+
+	// If there is no non-evicting healthy replica,
+	// Longhorn should retain one evicting healthy replica.
+	hasNonEvictingHealthyReplicas := false
```
c3y1huang, line 855:
```suggestion
	hasNonEvictingHealthyReplica := false
```

controller/volume_controller.go
derekbit:
```suggestion
			log.Debugf("Cannot evicting replica %v for now since there is no other healthy replica", r.Name)
```

			log.Warnf("Cannot evict replica %v for now since there is no other healthy replica", r.Name)

shuo-wu: Sure we can use `Warnf`

c3y1huang (approved):
LGTM

derekbit (approved):
LGTM

controller/volume_controller.go
```diff
@@ -848,18 +848,40 @@ func (vc *VolumeController) cleanupExtraHealthyReplicas(v *longhorn.Volume, e *l
...
+		if err := vc.deleteReplica(r, rs); err != nil {
+			vc.eventRecorder.Eventf(v, v1.EventTypeWarning,
+				EventReasonFailedEviction,
+				"volume %v failed to evict replica %v",
+				v.Name, r.Name)
+			return false, err
+		}
+		log.Debugf("Evicted replica %v in disk %v of node %v ", r.Name, r.Spec.DiskID, r.Spec.NodeID)
+		return true, nil
 	}
```
innobead, line 885: Why not just use evictingHealthyReplica as *replica instead of replica name, and then you don't need to have another loop for the replica deletion?

```
	if !hasNonEvictingHealthyReplica {
		log.Warnf("Cannot evicting replica %v for now since there is no other healthy replica", evictingHealthyReplica.Name)
		return false, nil
	}
	  
	if err := vc.deleteReplica(evictingHealthyReplica, rs); err != nil {
		vc.eventRecorder.Eventf(
			v, 
			v1.EventTypeWarning,
			EventReasonFailedEviction,
			"volume %v failed to evict replica %v",
			v.Name, 
			evictingHealthyReplica.Name,
		)
		return false, err
	}
		
	log.Debugf("Evicted replica %v in disk %v of node %v ", r.Name, r.Spec.DiskID, r.Spec.NodeID)
	return true, nil
```

innobead (changes requested):

controller/volume_controller.go
shuo-wu: If there are multiple evicting healthy replicas, the current implement could remove some of them. But the suggested code, all eviction will be blocked, right?

innobead: ah, you are right! I missed that, LGTM 🙂

innobead (approved):

innobead:
@shuo-wu Please backport to 1.3.x and 1.2.x. Thanks.

PR #1436, opened 2022-07-29
Fix bug snapshot CR block the deletion of volume CR

PhanLe1010:
longhorn/longhorn#4278

### DIFF
--- a/controller/snapshot_controller.go
+++ b/controller/snapshot_controller.go
@@ -82,6 +82,12 @@ func NewSnapshotController(
 	ds.EngineInformer.AddEventHandlerWithResyncPeriod(cache.ResourceEventHandlerFuncs{
 		UpdateFunc: sc.enqueueEngineChange,
 	}, 0)
+	sc.cacheSyncs = append(sc.cacheSyncs, ds.EngineInformer.HasSynced)
+
+	ds.VolumeInformer.AddEventHandlerWithResyncPeriod(cache.ResourceEventHandlerFuncs{
+		DeleteFunc: sc.enqueueVolumeChange,
+	}, 0)
+	sc.cacheSyncs = append(sc.cacheSyncs, ds.VolumeInformer.HasSynced)
 
 	return sc
 }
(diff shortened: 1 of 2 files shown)


controller/volume_controller.go
```diff
@@ -361,11 +361,6 @@ func (vc *VolumeController) syncVolume(key string) (err error) {
 		}
 
 		// now snapshots, replicas, and engines have been marked for deletion
-		if snapshots, err := vc.ds.ListVolumeSnapshotsRO(volume.Name); err != nil {
```
c3y1huang, line 364: After enqueueing the volume change at deletion, the snapshots should get removed. Can you explain why we are removing this dependency for volume deletion?

PhanLe1010: Surely, there are 2 points:
1. Volume CR deletion doesn't need to wait for snapshot CR deletion
2. The DeleteFunc event handler func only gets triggered when the volume object is removed from the cache, not when the volume CR has the deletion timestamp. We might be wondering why not use the UpdateFunc instead. The reason is because UpdateFunc get triggered very frequently and waste resources watching for volume CR changes

c3y1huang: Thank you for the explanation. It's clear now.

c3y1huang (approved):
LGTM

derekbit:
IIUC, it's like a chicken-egg prob before the fix. The deleted snapshot is waiting for the deletion of volume, and vice versa?

shuo-wu (approved):

derekbit (approved):

mantissahz (approved):

controller/snapshot_controller.go
```diff
@@ -183,6 +189,35 @@ func filterSnapshotsForEngineEnqueuing(oldEngine, curEngine *longhorn.Engine, sn
...
+		}
+	}
+	if vol.DeletionTimestamp.IsZero() {
+		return
+	}
+
+	snapshots, err := sc.ds.ListVolumeSnapshotsRO(vol.Name)
+	if err != nil {
+		utilruntime.HandleError(fmt.Errorf("snapshot controller failed to list snapshots when enqueuing volume %v: %v", vol.Name, err))
+		return
```
innobead, line 214: Is it possible to cause orphaned snapshots here? if just somehow it's failed to get snapshots, and the volume has been deleted asynchronously?

innobead:
LGTM, just one question https://github.com/longhorn/longhorn-manager/pull/1436#discussion_r933057674, after resolving, we can merge this.

controller/snapshot_controller.go
innobead: After discussing with @shuo-wu , this is possible but right now it's unable to handle via the current asynchronous mechanism (controller) except by leveraging the orphan mechanism.

This can be considered later, rather low priority.

innobead (approved):

innobead:
@PhanLe1010 Please help backport to 1.3.x. Thanks.

PhanLe1010:
> IIUC, before the fix, the deleted snapshot is waiting for the deletion of volume, and vice versa?

Before the fix, the snapshot controller doesn't re-queue the snapshot CR to remove the finalizer thus snapshot CR deletion is blocked. This in turn blocks the volume CR deletion. 

Now, the flow is: when the volume CR has deletion timestamp -> volume controller just removes the finalizer for the volume CR and allows it to be deleted -> Snapshot controller catches this deletion event and enqueues all snapshot CRs of the volume -> snapshot controller removes the finalizers for those snapshot CRs -> snapshot CRs are deleted -> finish the flow

PR #1441, opened 2022-08-01
Include volumes, engines, and replicas in the engineImage refcount calculation

derekbit:
[Longhorn 4332](https://github.com/longhorn/longhorn/issues/4332)


controller/engine_image_controller_test.go
```diff
@@ -89,7 +89,7 @@ func getEngineImageControllerTestTemplate() *EngineImageControllerTestCase {
 	tc.volume.Status.CurrentNodeID = TestNode1
 	tc.volume.Status.State = longhorn.VolumeStateAttached
 	tc.volume.Status.CurrentImage = TestEngineImage
-	tc.currentEngineImage.Status.RefCount = 1
+	tc.currentEngineImage.Status.RefCount = 2
```
innobead, line 92: nit: OK, because this test only simulates volume and engine w/o replicas, it's 2 here. Probably we should cover all cases here.

innobead (approved):
LGTM

innobead:
@derekbit Please backport this to 1.2.x and 1.3.x. Thanks.

controller/engine_image_controller_test.go
derekbit: Yes, we can add test cases containing volumes, engines and replicas in e2e.

c3y1huang (approved):

controller/engine_image_controller_test.go
innobead: Let's add test cases for that, but later implementation.

joshimoo: NIT: improving the unit test coverage for a change like this, would be nice. We can do it after the release is wrapped up though.

PR #1442, opened 2022-08-01
Delete reduent manfiests and scripts in longhorn repo

innobead:

We will just keep these files in loghorn/longhorn repo instead to avoid unnecessary file sync up.


innobead:
@longhorn/qa longhorn-tests is using artifacts in longhorn-manager, so please help confirm if this deletion will not impact the current testing pipeline on CI.

PR #1443, opened 2022-08-01
vendor: update longhorn-instance-manager

c3y1huang:
Show none-JSON error from backupstore.
https://github.com/longhorn/longhorn-instance-manager/pull/129

https://github.com/longhorn/longhorn/issues/4336


derekbit (approved):

innobead (approved):
LGTM

note: this is a client-side update, so no need to release the instance-manager image.

PR #1456, opened 2022-08-05
Fix evicting error and improve the related logs

shuo-wu:
https://github.com/longhorn/longhorn/issues/4294

### DIFF
--- a/controller/engine_controller.go
+++ b/controller/engine_controller.go
@@ -865,11 +865,12 @@ func (m *EngineMonitor) refresh(engine *longhorn.Engine) error {
 
 	// Make sure the engine object is updated before engineapi calls.
 	if !reflect.DeepEqual(existingEngine.Status, engine.Status) {
-		engine, err = m.ds.UpdateEngineStatus(engine)
-		existingEngine = engine.DeepCopy()
-		if err != nil {
+		EngineOpStatusFilter(engine)
+
+		if engine, err = m.ds.UpdateEngineStatus(engine); err != nil {
 			return err
 		}
+		existingEngine = engine.DeepCopy()
 	}
 
 	if cliAPIVersion >= engineapi.MinCLIVersion {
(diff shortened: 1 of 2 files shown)


controller/engine_controller.go
```diff
@@ -1676,3 +1684,30 @@ func (ec *EngineController) isResponsibleFor(e *longhorn.Engine, defaultEngineIm
 
 	return isPreferredOwner || continueToBeOwner || requiresNewOwner, nil
 }
+
+func EngineOpStatusFilter(e *longhorn.Engine) {
+	tcpReplicaAddrMap := map[string]bool{}
```
c3y1huang, line 1689: Suggest to use `map[string]struct{}` instead.

shuo-wu: Can you explain more about this? Will Golang apply any optimizations when using `struct{}`?

innobead (commented):
LGTM, just some comments to resolve.

controller/engine_controller.go
c3y1huang: I was looking where the boolean is used, and seems we are only using it for key lookups. AFAIK Using an empty struct doesn’t require memory allocated but a bool does.

controller/engine_controller.go
```diff
@@ -863,11 +865,12 @@ func (m *EngineMonitor) refresh(engine *longhorn.Engine) error {
 
 	// Make sure the engine object is updated before engineapi calls.
 	if !reflect.DeepEqual(existingEngine.Status, engine.Status) {
-		engine, err = m.ds.UpdateEngineStatus(engine)
-		existingEngine = engine.DeepCopy()
-		if err != nil {
+		EngineOpStatusFilter(engine)
+
+		if engine, err = m.ds.UpdateEngineStatus(engine); err != nil {
```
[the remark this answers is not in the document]
shuo-wu: No. That's why we use this function in the monitor thread only.

innobead (approved):

PR #1464, opened 2022-08-08
Implement recovery-backend service for RWX volumes' NFS servers

derekbit:
https://github.com/longhorn/longhorn/issues/2293


recovery_backend/client/client.go
```diff
@@ -0,0 +1,65 @@
...
+	lhclientset "github.com/longhorn/longhorn-manager/k8s/pkg/client/clientset/versioned"
+	lhinformers "github.com/longhorn/longhorn-manager/k8s/pkg/client/informers/externalversions"
+)
+
+type Client struct {
+	clients.Clients
+	Datastore *datastore.DataStore
+}
+
+func New(ctx context.Context, config *rest.Config, namespace string) (*Client, error) {
```
innobead, line 27: nit: the code is almost the same as the webhook implementation, so would suggest abstract the same part, then reuse in both places if needed.

recovery_backend/server/helper.go
```diff
@@ -0,0 +1,18 @@
+package server
+
+import (
+	"encoding/json"
+	"net/http"
+)
+
+func RespondWithError(rw http.ResponseWriter, code int, message string) {
```
innobead, line 8: Do we have similar things from webhook? If yes, would suggest refactoring a bit then reuse.

recovery_backend/server/router.go
```diff
@@ -0,0 +1,55 @@
...
+)
+
+type HandleFuncWithError func(http.ResponseWriter, *http.Request) error
+
+const (
+	RetryCounts   = 5
+	RetryInterval = 100 * time.Millisecond
+)
+
+func HandleError(s *client.Schemas, t HandleFuncWithError) http.Handler {
```
innobead, line 23: No need to export because it is just used in NewRouter.

P.S. the code seems repeated, so suggest reusing the code via common lib.

recovery_backend/server/recovery_backend.go
```diff
@@ -0,0 +1,101 @@
...
+
+func (r *RecoveryBackendServer) RecoveryBackendCreate(rw http.ResponseWriter, req *http.Request) error {
+	var input RecoveryBackendInput
+
+	apiContext := api.GetApiContext(req)
+	if err := apiContext.Read(&input); err != nil {
+		return err
+	}
+
+	logrus.Infof("Create a recovery backend %v (version %v)", input.Hostname, input.Version)
```
innobead, line 20: `Creating a recovery backend ...`

recovery_backend/backend/configmap.go
```diff
@@ -0,0 +1,219 @@
...
+	if err != nil {
+		return errors.Wrapf(err, "failed to get configmap %v", configMapName)
+	}
+
+	dataStr := cm.Data[version]
+	if dataStr == "" {
+		return errors.Wrapf(err, "failed to get data from recovery backend %v (version %v)", configMapName, version)
+	}
+
+	data := map[string][]string{}
```
innobead, line 194: Could we use type alias to make the code easy to understand?

```golang
type clientID string
type revokeFileHandle string

data := map[clientID][]revokeFileHandle
```

innobead (commented):
In general, LGTM.

Just have some feedbacks related to code reuse, method namding and log convention.

deploy/install/02-components/02-backend.yaml
```diff
@@ -51,6 +51,23 @@ spec:
...
+  namespace: longhorn-system
+spec:
+  type: ClusterIP
+  sessionAffinity: ClientIP
+  selector:
+    app: longhorn-recovery-backend
+  ports:
+  - name: recovery-backend
+    port: 9600
+    targetPort: recbackend-wh
```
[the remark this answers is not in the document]
derekbit: Updated. Thank you.

datastore/kubernetes.go
```diff
@@ -516,6 +516,28 @@ func (s *DataStore) ListVolumeAttachmentsRO() ([]*storagev1.VolumeAttachment, er
...
+// UpdateConfigMap updates ConfigMap resource
+func (s *DataStore) UpdateConfigMap(configMap *corev1.ConfigMap) (*corev1.ConfigMap, error) {
+	return s.kubeClient.CoreV1().ConfigMaps(s.namespace).Update(context.TODO(), configMap, metav1.UpdateOptions{})
+}
+
+func (s *DataStore) ListConfigMap() ([]*corev1.ConfigMap, error) {
+	return s.cfmLister.ConfigMaps(s.namespace).List(labels.Everything())
+}
+
+func (s *DataStore) ListConfigMapLabels(labels map[string]string) ([]*corev1.ConfigMap, error) {
```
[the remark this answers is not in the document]
derekbit: We don't need the function anymore. Removed it.

app/recovery_backend.go
```diff
@@ -0,0 +1,84 @@
...
+				Usage: "Specify service account for recovery-backend service",
+			},
+			cli.StringFlag{
+				Name:  FlagKubeConfig,
+				Usage: "Specify path to kube config (optional)",
+			},
+		},
+		Action: func(c *cli.Context) {
+			if err := runRecoveryBackendServer(c); err != nil {
+				logrus.Fatalf("Error starting longhorn recovery-backend server: %v", err)
```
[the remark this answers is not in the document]
derekbit: Updated. Thank you.
There are lots of error messages like `Error V+ing....`. Do we plan to use `failed to ...` as the convention?

controller/share_manager_controller.go
```diff
@@ -299,6 +299,12 @@ func (c *ShareManagerController) syncShareManager(key string) (err error) {
 		if err := c.cleanupShareManagerPod(sm); err != nil {
 			return err
 		}
+
+		if err := c.ds.DeleteConfigMap(types.GetConfigMapNameNameFromShareManagerName(sm.Name)); err != nil {
+			log.Warnf("Unable to delete the configmap (recovery backend) for share manager %v since %v",
```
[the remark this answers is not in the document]
derekbit: Updated. Thank you.

datastore/kubernetes.go
```diff
@@ -533,6 +555,11 @@ func (s *DataStore) GetConfigMap(namespace, name string) (*corev1.ConfigMap, err
 	return resultRO.DeepCopy(), nil
 }
 
+// DeleteConfigMap deletes the ConfigMap for the given name and namespace
+func (s *DataStore) DeleteConfigMap(name string) error {
```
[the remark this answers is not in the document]
derekbit: This is weird. Actually, `DataStore` contains the namespace field, but some functions still have `namespace` argument...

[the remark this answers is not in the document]
derekbit: Updated. Thank you.

recovery_backend/backend/configmap.go
```diff
@@ -0,0 +1,219 @@
...
+			Namespace: namespace,
+			Name:      name,
+		},
+		Data: map[string]string{
+			version: "{}",
+		},
+	}
+}
+
+func (rb *RecoveryBackend) RecovCreateConfigMap(configMapName, version string) error {
```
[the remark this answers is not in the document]
derekbit: Updated. Thank you.

recovery_backend/server/router.go
derekbit: Updated. Thank you.

recovery_backend/server/recovery_backend.go
derekbit: Updated. Thank you.

recovery_backend/backend/configmap.go
derekbit: Addressed. Thank you.

recovery_backend/server/helper.go
derekbit: This is a dead code. Removed it. Thank you.

recovery_backend/client/client.go
derekbit: Addressed. Thank you.

derekbit:
Update
1. Delete recovery backend deployment during unstallation
2. Add `initialDelaySeconds: 3` and `periodSeconds: 5` to `longhorn-recovery-backend` deployment.
    sympton: `Warning Unhealthy...: connection refused` in  longhorn-recovery-backend pods due to insufficient time for the pod to start.

controller/uninstall_controller.go
```diff
@@ -296,11 +296,15 @@ func (c *UninstallController) uninstall() error {
...
 	}
 
-	if waitForUpdate, err := c.deleteWebhookDeployment(types.LonghornConversionWebhookDeploymentName); err != nil || waitForUpdate {
+	if waitForUpdate, err := c.deleteDeployment(types.LonghornAdmissionWebhookDeploymentName); err != nil || waitForUpdate {
+		return err
+	}
+
+	if waitForUpdate, err := c.deleteDeployment(types.LonghornConversionWebhookDeploymentName); err != nil || waitForUpdate {
 		return err
 	}
```
innobead, line 309: nit: this is good, but probably we can have a loop for these deletions because the thing is operation is the same.

types/types.go
```diff
@@ -44,6 +44,8 @@ const (
 	ValidatingWebhookName = "longhorn-webhook-validator"
 	MutatingWebhookName   = "longhorn-webhook-mutator"
 
+	DefaultRecoveryBackendServerPort = 9600
```
innobead, line 47: Remember to update our doc about port usage.

types/types.go
```diff
@@ -101,6 +103,7 @@ const (
 	LonghornLabelRecurringJobGroup        = "job-group"
 	LonghornLabelOrphan                   = "orphan"
 	LonghornLabelOrphanType               = "orphan-type"
+	LonghornLabelRecoveryBackendConfigMap = "recovery-backend"
```
innobead, line 106: Suggest renaming to `LonghornLabelRecoveryBackend` because configmap is implementation detail, probably no need to disclose in this const.

controller/share_manager_controller.go
```diff
@@ -299,6 +299,12 @@ func (c *ShareManagerController) syncShareManager(key string) (err error) {
 		if err := c.cleanupShareManagerPod(sm); err != nil {
 			return err
 		}
+
+		if err := c.ds.DeleteConfigMap(c.namespace, types.GetConfigMapNameNameFromShareManagerName(sm.Name)); err != nil {
+			log.Warnf("Failed to delete the configmap (recovery backend) for share manager %v since %v",
```
innobead, line 304: If can't delete, it will be left over? Should it be a transaction together by returning the error?

types/types.go
```diff
@@ -457,6 +466,10 @@ func GetShareManagerPodNameFromShareManagerName(smName string) string {
 	return shareManagerPrefix + smName
 }
 
+func GetConfigMapNameNameFromShareManagerName(smName string) string {
```
innobead, line 469: Change to `GetConfigMapNameFromShareManagerName`

deploy/install/01-prerequisite/03-crd.yaml
```diff
@@ -2710,6 +2712,7 @@ metadata:
     longhorn-manager: ""
   name: volumes.longhorn.io
 spec:
+  preserveUnknownFields: false
```
innobead, line 2715: This was missed. 

I feel we should not have a separate manifest in the manager repo but only keep one place (longhorn/longhorn) instead.

innobead (changes requested):
LGTM, just a few comments.

controller/uninstall_controller.go
derekbit: Updated. Thank you.

types/types.go
derekbit: Updated. Thank you.

controller/share_manager_controller.go
derekbit: I think we need to return error here for avoiding the left-over.

types/types.go
derekbit: Updated. Thank you.

innobead (dismissed):

derekbit:
Waiting for https://github.com/longhorn/longhorn-share-manager/pull/36

derekbit:
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/1934/

types/types.go
derekbit: Done. Thank you.

types/types.go
```diff
@@ -457,6 +466,10 @@ func GetShareManagerPodNameFromShareManagerName(smName string) string {
 	return shareManagerPrefix + smName
 }
 
+func GetConfigMapNameFromShareManagerName(smName string) string {
+	return shareManagerPrefix + smName
```
innobead, line 470: Because you actually use the share manager name, I would suggest  changing the code as below to make it explicitly.

```golang
return GetShareManagerPodNameFromShareManagerName(smName)
```

innobead (dismissed):
LGTM.

types/types.go
derekbit: No prob. Maybe add a prefix, e.g. "recovery-backend-<share manager name>". How about it?

innobead: Sounds good to know if the length is allowed.

derekbit: Updated. Thank you.

innobead (approved):

PR #1465, opened 2022-08-09
Add backing image backup restore

FrankYang0529:
https://github.com/longhorn/longhorn/issues/4165

rely on https://github.com/longhorn/backing-image-manager/pull/65


deploy/install/01-prerequisite/03-crd.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.7.0
+    controller-gen.kubebuilder.io/version: v0.8.0
```
c3y1huang, line 5: Curious, what is the reason for the version bump?

deploy/install/01-prerequisite/03-crd.yaml
```diff
@@ -350,23 +351,12 @@ apiVersion: apiextensions.k8s.io/v1
...
-    webhook:
-      clientConfig:
-        service:
-          name: longhorn-conversion-webhook
-          namespace: longhorn-system
-          path: /v1/webhook/conversion
-          port: 9443
-      conversionReviewVersions:
-      - v1beta2
-      - v1beta1
```
c3y1huang, line 369: Can you explain why are we removing these?

deploy/install/01-prerequisite/03-crd.yaml
```diff
@@ -2872,7 +2819,7 @@ spec:
               recurringJobs:
                 description: Deprecated. Replaced by a separate resource named "RecurringJob"
                 items:
-                  description: 'VolumeRecurringJobSpec is a deprecated struct. TODO: Should be removed when recurringJobs gets removed from the volume       spec.'
+                  description: 'VolumeRecurringJobSpec is a deprecated struct. TODO: Should be removed when recurringJobs gets removed from the volume spec.'
```
c3y1huang, line 2822: Thanks!

deploy/install/01-prerequisite/03-crd.yaml
FrankYang0529: I didn't notice I used different `controller-gen` version. Fixed it. Thank you.

deploy/install/01-prerequisite/03-crd.yaml
FrankYang0529: Hmm, I already checked all my command versions in [generate_code.sh](https://github.com/longhorn/longhorn-manager/blob/master/k8s/generate_code.sh). I am not sure why this is removed. Could you help to run the script again to see whether this is added back? Thank you.

deploy/install/01-prerequisite/03-crd.yaml
FrankYang0529: Looks like this is fixed in controller-gen v0.8.0. Do we want to fix this in https://github.com/longhorn/longhorn-manager/blob/97460182360e925fddfc9e586814f63b2b6a1344/k8s/pkg/apis/longhorn/v1beta2/volume.go#L104

PR #1466, opened 2022-08-10
Fix don't set volume to faulted if it is attaching the first time and one of the replica failed during attaching process

PhanLe1010:
In this case, there are still other replicas that can be started
so we don't set the volume to faulted here

longhorn/longhorn#4386

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -701,6 +701,33 @@ func (vc *VolumeController) ReconcileEngineReplicaState(v *longhorn.Volume, es m
 	return nil
 }
 
+func isAutoSalvageNeeded(rs map[string]*longhorn.Replica) bool {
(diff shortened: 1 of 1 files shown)


joshimoo (approved):
LGTM, what state will we be setting the volume to?
- all replicas failed during first attach
- one of the replicas failed during first attach
- no failure during first attach

PhanLe1010:
all replicas failed during first attach -> faulted 
one of the replicas failed during first attach -> attached and healthy
no failure during first attach -> attached and healthy

innobead (approved):
LGTM

shuo-wu (approved):

innobead:
@PhanLe1010 Please backport to 1.3 and 1.2.

PR #1467, opened 2022-08-10
Only delete the VolumeAttachment on the downed node

PhanLe1010:
longhorn/longhorn#4384


innobead (approved):
LGTM

innobead:
@PhanLe1010 Please backport this to 1.3 and 1.2. Thanks.

PhanLe1010:
> This was missed before, and suggest having unit tests coverage for this method.

Unit test is hard to catch this one since it requires interaction with the kube-controller-manager. This happens when kube-controller-manager create new VolumeAttachment on the new node for deployment and Longhorn keep fighting to delete it. This shouldn't be a problem with Statefulset

PR #1475, opened 2022-08-12
Convert volume.spec.dataLocality from empty string to disabled

derekbit:
https://github.com/longhorn/longhorn/issues/4412


joshimoo (approved):
LGTM

PR #1477, opened 2022-08-15
Failed backups cleanup

mantissahz:
Clean up failed backups when there is a backup in state `Error` or `Unknown`.

longhorn/longhorn#3898
longhorn/longhorn#4334

### DIFF
--- a/controller/backup_controller.go
+++ b/controller/backup_controller.go
@@ -299,6 +299,10 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
 		// Disable monitor regardless of backup state
 		bc.disableBackupMonitor(backup.Name)
 
+		if backup.Status.State == longhorn.BackupStateError || backup.Status.State == longhorn.BackupStateUnknown {
+			bc.eventRecorder.Eventf(backup, corev1.EventTypeWarning, string(backup.Status.State), "Failed backup %s has been deleted: %s", backup.Name, backup.Status.Error)
+		}
+
 		return bc.ds.RemoveFinalizerForBackup(backup)
 	}
 
--- a/controller/backup_volume_controller.go
+++ b/controller/backup_volume_controller.go
@@ -284,11 +284,24 @@ func (bvc *BackupVolumeController) reconcile(backupVolumeName string) (err error
 		return err
 	}
 
+	// Get `Time to live` after failed backup was marked as `Error` or `Unknown`,
+	failedBackupTTL, err := bvc.ds.GetSettingAsInt(types.SettingNameFailedBackupTTL)
+	if err != nil {
+		log.WithError(err).Warnf("Failed to get %v setting, and it will skip the auto-deletion for the failed backups", types.SettingNameFailedBackupTTL)
+	}
 	clustersSet := sets.NewString()
 	for _, b := range clusterBackups {
 		// Skip the Backup CR which is created from the local cluster and
 		// the snapshot backup hasn't be completed or pulled from the remote backup target yet
 		if b.Spec.SnapshotName != "" && b.Status.State != longhorn.BackupStateCompleted {
+			if b.Status.State == longhorn.BackupStateError || b.Status.State == longhorn.BackupStateUnknown {
+				// Failed backup `LastSyncedAt` should not be updated after it was marked as `Error` or `Unknown`
+				if failedBackupTTL > 0 && time.Now().After(b.Status.LastSyncedAt.Add(time.Duration(failedBackupTTL)*time.Minute)) {
+					if err = bvc.ds.DeleteBackup(b.Name); err != nil {
+						log.WithError(err).Errorf("failed to delete failed backup %s", b.Name)
+					}
+				}
+			}
 			continue
 		}
 		clustersSet.Insert(b.Name)
--- a/deploy/install/01-prerequisite/04-default-setting.yaml
+++ b/deploy/install/01-prerequisite/04-default-setting.yaml
@@ -44,3 +44,4 @@ data:
     #kubernetes-cluster-autoscaler-enabled:
     #orphan-auto-deletion:
     #storage-network:
+    #failed-backup-ttl:
--- a/types/setting.go
+++ b/types/setting.go
@@ -83,6 +83,7 @@ const (
 	SettingNameKubernetesClusterAutoscalerEnabled           = SettingName("kubernetes-cluster-autoscaler-enabled")
 	SettingNameOrphanAutoDeletion                           = SettingName("orphan-auto-deletion")
 	SettingNameStorageNetwork                               = SettingName("storage-network")
+	SettingNameFailedBackupTTL                              = SettingName("failed-backup-ttl")
 )
 
 var (
@@ -136,6 +137,7 @@ var (
 		SettingNameKubernetesClusterAutoscalerEnabled,
 		SettingNameOrphanAutoDeletion,
 		SettingNameStorageNetwork,
+		SettingNameFailedBackupTTL,
 	}
 )
 
@@ -213,6 +215,7 @@ var (
 		SettingNameKubernetesClusterAutoscalerEnabled:           SettingDefinitionKubernetesClusterAutoscalerEnabled,
 		SettingNameOrphanAutoDeletion:                           SettingDefinitionOrphanAutoDeletion,
 		SettingNameStorageNetwork:                               SettingDefinitionStorageNetwork,
+		SettingNameFailedBackupTTL:                              SettingDefinitionFailedBackupTTL,
 	}
 
 	SettingDefinitionBackupTarget = SettingDefinition{
@@ -253,6 +256,19 @@ var (
 		Required:    true,
 		ReadOnly:    false,
 		Default:     "300",
+	}
+
+	SettingDefinitionFailedBackupTTL = SettingDefinition{
+		DisplayName: "Failed Backup Time to Live",
+		Description: "In minutes. This setting determines how long Longhorn will keep the backup resource that was failed. Set to 0 to disable the auto-deletion.\n" +
+			"Failed backups will be checked and cleaned up during backupstore polling which is controlled by **Backupstore Poll Interval** setting.\n" +
+			"Hence this value determines the minimal wait interval of the cleanup. And the actual cleanup interval is multiple of **Backupstore Poll Interval**.\n" +
+			"Disabling **Backupstore Poll Interval** also means to disable failed backup auto-deletion.\n\n",
+		Category: SettingCategoryBackup,
+		Type:     SettingTypeInt,
+		Required: true,
+		ReadOnly: false,
+		Default:  "1440",
 	}
 
 	SettingDefinitionCreateDefaultDiskLabeledNodes = SettingDefinition{
@@ -922,6 +938,14 @@ func ValidateSetting(name, value string) (err error) {
 	case SettingNameConcurrentAutomaticEngineUpgradePerNodeLimit:
 		fallthrough
 	case SettingNameBackupstorePollInterval:
+		interval, err := strconv.Atoi(value)
+		if err != nil {
+			return fmt.Errorf("value is not int: %v", err)
+		}
+		if interval < 0 {
+			return fmt.Errorf("the value %v shouldn't be less than 0", value)
+		}
+	case SettingNameFailedBackupTTL:
 		interval, err := strconv.Atoi(value)
 		if err != nil {
 			return fmt.Errorf("value is not int: %v", err)


controller/backup_volume_controller.go
```diff
@@ -289,6 +289,27 @@ func (bvc *BackupVolumeController) reconcile(backupVolumeName string) (err error
...
+				if err != nil {
+					log.WithError(err).Errorf("Failed to get %v setting for failed backup %s", types.SettingNameFailedBackupTTL, b.Name)
+					continue
+				}
+				// Failed backup `LastSyncedAt` should not be updated after it was marked as `Error` or `Unknown`
+				if autoFailedBackupCleanup && time.Now().After(b.Status.LastSyncedAt.Add(time.Duration(interval)*time.Second)) {
+					if err = bvc.ds.DeleteBackup(b.Name); err != nil {
+						log.WithError(err).Errorf("Error deleting failed backup %s from cluster", b.Name)
+					}
+				}
```
derekbit, line 311:
```
autoFailedBackupCleanup, err := bvc.ds.GetSettingAsBool(types.SettingNameFailedBackupAutoDeleteion)
...
if autoFailedBackupCleanup {
    interval, err := bvc.ds.GetSettingAsInt(types.SettingNameFailedBackupTTL)
    ...
    if time.Now().After(b.Status.LastSyncedAt.Add(time.Duration(interval)*time.Second)) {
        DeleteBackup(b.Name)
    }

}
````

File: types/setting.go [hunk 2]
```diff
@@ -255,6 +261,33 @@ var (
...
+		Type:     SettingTypeBool,
+		Required: true,
+		ReadOnly: false,
+		Default:  "true",
+	}
+
+	SettingDefinitionFailedBackupTTL = SettingDefinition{
+		DisplayName: "Failed Backup Time to Live",
+		Description: "In seconds. This setting determines how long Longhorn will keep the backup resource that was failed.\n\n" +
+			"WARNING: \n\n" +
```
derekbit, line 281: The `WARNING` looks not necessary. The following two lines are just the usage.

derekbit (approved):

controller/backup_volume_controller.go
mantissahz: Modified, thanks.

types/setting.go
mantissahz: Agree, Use `Note` instead of `Warning`, WDYT? Thanks.

types/setting.go
```diff
@@ -255,6 +261,34 @@ var (
...
+		ReadOnly: false,
+		Default:  "true",
+	}
+
+	SettingDefinitionFailedBackupTTL = SettingDefinition{
+		DisplayName: "Failed Backup Time to Live",
+		Description: "In seconds. This setting determines how long Longhorn will keep the backup resource that was failed.\n\n" +
+			"Note: \n\n" +
+			"  - Setting 0 to disable **Backupstore Poll Interval** also means to disable failed backup time to live.\n" +
+			"  - Disabling **Failed Backup Auto-Deletion** would disable failed backup time to live as well.\n",
```
derekbit, line 284: Suggest combining `"In seconds. This setting determines how long Longhorn will keep the backup resource that was failed.` and `Disabling **Failed Backup Auto-Deletion** would disable failed backup time to live as well.` and putting it into description.

mantissahz: Combined.

types/setting.go
```diff
@@ -255,6 +261,34 @@ var (
 		Default:     "300",
 	}
 
+	SettingDefinitionFailedBackupAutoDeleteion = SettingDefinition{
+		DisplayName: "Failed Backup Auto-Deletion",
+		Description: "This setting allows Longhorn to delete the backup resource that was failed automatically.\n\n" +
```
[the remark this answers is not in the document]

types/setting.go
```diff
@@ -255,6 +261,34 @@ var (
...
+			"WARNING: \n\n" +
+			"  - The failed backup will be removed in **BackupstorePollInterval** (default: 5 mintues) and you can only check the backup failed from event log.\n" +
```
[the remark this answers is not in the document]
mantissahz: Moved. Thanks.

types/setting.go
```diff
@@ -255,6 +261,34 @@ var (
...
+			"  - Keeping lots of the failed backups might affect the performance for backup volumes, upgrading or uninstallating longhorn system.\n\n" +
```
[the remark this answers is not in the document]
mantissahz: Removed. Thanks. Agree to add to the document.

types/setting.go
```diff
@@ -255,6 +261,34 @@ var (
...
+			"Note: \n\n" +
```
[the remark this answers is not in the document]
mantissahz: Removed `Note` and combined into description.

types/setting.go
```diff
@@ -255,6 +261,34 @@ var (
...
+		Type:     SettingTypeBool,
+		Required: true,
+		ReadOnly: false,
+		Default:  "true",
+	}
+
+	SettingDefinitionFailedBackupTTL = SettingDefinition{
+		DisplayName: "Failed Backup Time to Live",
+		Description: "In seconds. This setting determines how long Longhorn will keep the backup resource that was failed.\n\n" +
+			"Note: \n\n" +
```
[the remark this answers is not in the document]
mantissahz: Removed `Note` and combined into description.

controller/backup_volume_controller.go
```diff
@@ -289,6 +289,29 @@ func (bvc *BackupVolumeController) reconcile(backupVolumeName string) (err error
 		// Skip the Backup CR which is created from the local cluster and
 		// the snapshot backup hasn't be completed or pulled from the remote backup target yet
 		if b.Spec.SnapshotName != "" && b.Status.State != longhorn.BackupStateCompleted {
+			if b.Status.State == longhorn.BackupStateError || b.Status.State == longhorn.BackupStateUnknown {
+				// Start to delete the failed backups, default: true
+				autoFailedBackupCleanup, err := bvc.ds.GetSettingAsBool(types.SettingNameFailedBackupAutoDeleteion)
+				if err != nil {
+					log.WithError(err).Errorf("Failed to get %v setting for failed backup %s", types.SettingNameFailedBackupAutoDeleteion, b.Name)
```
c3y1huang, line 296: nit: lower case the first string for all error messages. Ref https://github.com/golang/go/wiki/CodeReviewComments#error-strings.

controller/backup_volume_controller.go
```diff
@@ -289,6 +289,29 @@ func (bvc *BackupVolumeController) reconcile(backupVolumeName string) (err error
 		// Skip the Backup CR which is created from the local cluster and
 		// the snapshot backup hasn't be completed or pulled from the remote backup target yet
 		if b.Spec.SnapshotName != "" && b.Status.State != longhorn.BackupStateCompleted {
+			if b.Status.State == longhorn.BackupStateError || b.Status.State == longhorn.BackupStateUnknown {
+				// Start to delete the failed backups, default: true
+				autoFailedBackupCleanup, err := bvc.ds.GetSettingAsBool(types.SettingNameFailedBackupAutoDeleteion)
```
c3y1huang, line 294: Should this be moved outside the for loop so we don't repeatedly get the same setting value for each iteration?

c3y1huang (changes requested):

deploy/install/01-prerequisite/04-default-setting.yaml
```diff
@@ -44,3 +44,5 @@ data:
     #kubernetes-cluster-autoscaler-enabled:
     #orphan-auto-deletion:
     #storage-network:
+    #failed-backup-auto-deletion:
+    #failed-backup-ttl:
```
c3y1huang, line 48: Is it possible to have just one setting `failed-backup-ttl` instead of two settings? 
For example,
if failed-backup-ttl == 0, disable the auto-deletion. 
if failed-backup-ttl > 0, enable the auto-deletion.

controller/backup_volume_controller.go

controller/backup_volume_controller.go
mantissahz: Moved before the loop.

deploy/install/01-prerequisite/04-default-setting.yaml
mantissahz: Yes, use only `failed-backup-ttl` and disabled if `-1`, no wait deleting if `0`, wait for x seconds if x > 0. Thanks, that's more simple and clear.

mantissahz: Changed, disable if failed-backup-ttl == 0, enable if failed-backup-ttl > 0 for consistency usage.

controller/backup_volume_controller.go
```diff
@@ -284,11 +284,27 @@ func (bvc *BackupVolumeController) reconcile(backupVolumeName string) (err error
 		return err
 	}
 
+	// Get `Time to live` after failed backup was marked as `Error` or `Unknown`,
+	// default: 180s and it is shorter than `Backupstore Poll Interval` by default, so failed backups should be deleted once the backup volume is updated.
+	failedBackupTTL, err := bvc.ds.GetSettingAsInt(types.SettingNameFailedBackupTTL)
+	if err != nil {
+		log.WithError(err).Errorf("failed to get %v setting from cluster", types.SettingNameFailedBackupTTL)
```
shuo-wu, line 291: "..., will skip the auto-deletion for the failed backups"

controller/backup_volume_controller.go
```diff
@@ -284,11 +284,27 @@ func (bvc *BackupVolumeController) reconcile(backupVolumeName string) (err error
 		return err
 	}
 
+	// Get `Time to live` after failed backup was marked as `Error` or `Unknown`,
+	// default: 180s and it is shorter than `Backupstore Poll Interval` by default, so failed backups should be deleted once the backup volume is updated.
+	failedBackupTTL, err := bvc.ds.GetSettingAsInt(types.SettingNameFailedBackupTTL)
```
shuo-wu, line 289: This means, the actual TTL is controlled by the poll interval. Or more precisely, it's the multiple of the poll interval, right?

shuo-wu (changes requested):

controller/backup_volume_controller.go
mantissahz: Yes, `failed backup TTL` is more like `at least` period the longhorn would keep the failed backup.

controller/backup_volume_controller.go
mantissahz: Added. Thanks.

controller/backup_volume_controller.go
shuo-wu: Then it's better to explain more in the description part.

types/setting.go
```diff
@@ -255,6 +258,18 @@ var (
 		Default:     "300",
 	}
 
+	SettingDefinitionFailedBackupTTL = SettingDefinition{
+		DisplayName: "Failed Backup Time to Live",
+		Description: "In seconds. This setting determines how long Longhorn will keep the backup resource that was failed. Set to 0 to disable the auto-deletion.\n" +
+			"The failed backup will be removed in **BackupstorePollInterval** (default: 300 seconds.) " +
```
innobead, line 264: Suggest to `Failed backups will be checked and cleaned up during backupstore polling which is controlled by **BackupstorePollInterval** setting.`

types/setting.go
```diff
@@ -255,6 +258,18 @@ var (
...
+			"Disabling **Backupstore Poll Interval** also means to disable failed Backup auto-deletion.\n\n",
```
innobead, line 265: Suggest to `Disabling **Backupstore Poll Interval** also means to disable failed backup auto-deletion`

controller/backup_volume_controller.go
```diff
@@ -284,11 +284,27 @@ func (bvc *BackupVolumeController) reconcile(backupVolumeName string) (err error
...
+		log.WithError(err).Errorf("failed to get %v setting from cluster, and it will skip the auto-deletion for the failed backups", types.SettingNameFailedBackupTTL)
+		// set failedBackupTTL to be 0 as disable failed backup auto-deletion
+		failedBackupTTL = 0
+	}
 	clustersSet := sets.NewString()
 	for _, b := range clusterBackups {
 		// Skip the Backup CR which is created from the local cluster and
 		// the snapshot backup hasn't be completed or pulled from the remote backup target yet
 		if b.Spec.SnapshotName != "" && b.Status.State != longhorn.BackupStateCompleted {
+			if b.Status.State == longhorn.BackupStateError || b.Status.State == longhorn.BackupStateUnknown {
```
innobead, line 300: Is `longhorn.BackupStateUnknown` possibly a transient state? then cause an incorrect deletion.

controller/backup_volume_controller.go
```diff
@@ -284,11 +284,27 @@ func (bvc *BackupVolumeController) reconcile(backupVolumeName string) (err error
...
+				// Failed backup `LastSyncedAt` should not be updated after it was marked as `Error` or `Unknown`
+				if failedBackupTTL > 0 && time.Now().After(b.Status.LastSyncedAt.Add(time.Duration(failedBackupTTL)*time.Second)) {
+					if err = bvc.ds.DeleteBackup(b.Name); err != nil {
+						log.WithError(err).Errorf("delete failed backup %s from cluster", b.Name)
```
innobead, line 304: The message should

`"failed to delete failed backup %s from cluster", b.Name`

innobead:
In general, LGTM. Have some feedback, please resolve others' comments as well before the merge.

P.S. this needs longhorn/longhorn change as well because a new setting will be introduced.

controller/backup_volume_controller.go
mantissahz: Modified as David's Suggestion. Thanks.

types/setting.go

controller/backup_volume_controller.go

controller/backup_volume_controller.go
mantissahz: I think it is not a transient state.

controller/backup_volume_controller.go
```diff
@@ -284,11 +284,27 @@ func (bvc *BackupVolumeController) reconcile(backupVolumeName string) (err error
 		return err
 	}
 
+	// Get `Time to live` after failed backup was marked as `Error` or `Unknown`,
+	// default: 180s and it is shorter than `Backupstore Poll Interval` by default, so failed backups should be deleted once the backup volume is updated.
+	failedBackupTTL, err := bvc.ds.GetSettingAsInt(types.SettingNameFailedBackupTTL)
+	if err != nil {
+		log.WithError(err).Errorf("failed to get %v setting from cluster, and it will skip the auto-deletion for the failed backups", types.SettingNameFailedBackupTTL)
+		// set failedBackupTTL to be 0 as disable failed backup auto-deletion
+		failedBackupTTL = 0
```
derekbit, line 293: Looks not required? Only `failedBackupTTL > 0` can allow the deletion.

controller/backup_volume_controller.go
```diff
@@ -284,11 +284,27 @@ func (bvc *BackupVolumeController) reconcile(backupVolumeName string) (err error
...
 	clustersSet := sets.NewString()
 	for _, b := range clusterBackups {
 		// Skip the Backup CR which is created from the local cluster and
 		// the snapshot backup hasn't be completed or pulled from the remote backup target yet
 		if b.Spec.SnapshotName != "" && b.Status.State != longhorn.BackupStateCompleted {
+			if b.Status.State == longhorn.BackupStateError || b.Status.State == longhorn.BackupStateUnknown {
+				// Failed backup `LastSyncedAt` should not be updated after it was marked as `Error` or `Unknown`
+				if failedBackupTTL > 0 && time.Now().After(b.Status.LastSyncedAt.Add(time.Duration(failedBackupTTL)*time.Second)) {
+					if err = bvc.ds.DeleteBackup(b.Name); err != nil {
+						log.WithError(err).Errorf("failed to delete failed backup %s from cluster", b.Name)
```
derekbit, line 304:
```suggestion
						log.WithError(err).Errorf("failed to delete failed backup %s", b.Name)
```

controller/backup_volume_controller.go
mantissahz: yap, it will return -1 if there is an error occurred.
Modified. Thanks

controller/backup_volume_controller.go
mantissahz: Modified. Thanks

controller/backup_volume_controller.go
```diff
@@ -284,11 +284,26 @@ func (bvc *BackupVolumeController) reconcile(backupVolumeName string) (err error
 		return err
 	}
 
+	// Get `Time to live` after failed backup was marked as `Error` or `Unknown`,
+	// default: 180s and it is shorter than `Backupstore Poll Interval` by default, so failed backups should be deleted once the backup volume is updated.
+	failedBackupTTL, err := bvc.ds.GetSettingAsInt(types.SettingNameFailedBackupTTL)
+	if err != nil {
+		log.WithError(err).Errorf("failed to get %v setting from cluster, and it will skip the auto-deletion for the failed backups", types.SettingNameFailedBackupTTL)
```
derekbit, line 291:
```suggestion
		log.WithError(err).Errorf("failed to get setting %v", types.SettingNameFailedBackupTTL)
```

types/setting.go
```diff
@@ -255,6 +258,18 @@ var (
...
+	SettingDefinitionFailedBackupTTL = SettingDefinition{
+		DisplayName: "Failed Backup Time to Live",
+		Description: "In seconds. This setting determines how long Longhorn will keep the backup resource that was failed. Set to 0 to disable the auto-deletion.\n" +
+			"Failed backups will be checked and cleaned up during backupstore polling which is controlled by **BackupstorePollInterval** setting. " +
+			"Disabling **Backupstore Poll Interval** also means to disable failed backup auto-deletion.\n\n",
+		Category: SettingCategoryBackup,
+		Type:     SettingTypeInt,
+		Required: true,
+		ReadOnly: false,
+		Default:  "180",
```
derekbit, line 270: Just want to reconfirm TTL's purpose and the default value.
As you mentioned, the purpose of the TTL is to let users be aware of the failed backup, and user can check the resources and etc. According to the purpose, 180 seconds seems pretty short. Is bigger value better?

innobead: This was discussed as well. I think the longer value should be more reasonable. How about 1 day?

controller/backup_volume_controller.go

types/setting.go
mantissahz: Yeah, 1 day is reasonable if we would keep failed backups for users.
The original design is that enabling auto-deletion means deleting failed backup ASAP because failed backup is meaningless if users enabled the auto-deletion by themselves.
Now auto-deletion is on by default, and we have the TTL for failed backups.

Modified and changed the unit from `seconds` to `minutes`. Thanks.

controller/backup_volume_controller.go
```diff
@@ -284,11 +284,26 @@ func (bvc *BackupVolumeController) reconcile(backupVolumeName string) (err error
 		return err
 	}
 
+	// Get `Time to live` after failed backup was marked as `Error` or `Unknown`,
+	// default: 180s and it is shorter than `Backupstore Poll Interval` by default, so failed backups should be deleted once the backup volume is updated.
```
innobead, line 288: The default was changed not 180 seconds. This comment is not needed anymore.

controller/backup_volume_controller.go
```diff
@@ -284,11 +284,26 @@ func (bvc *BackupVolumeController) reconcile(backupVolumeName string) (err error
...
+	failedBackupTTL, err := bvc.ds.GetSettingAsInt(types.SettingNameFailedBackupTTL)
+	if err != nil {
+		log.WithError(err).Errorf("failed to get %v setting", types.SettingNameFailedBackupTTL)
+		// failedBackupTTL would be -1 if there is an error.
```
innobead, line 292: Is this comment still valid?

controller/backup_volume_controller.go
```diff
@@ -284,11 +284,26 @@ func (bvc *BackupVolumeController) reconcile(backupVolumeName string) (err error
...
+	}
 	clustersSet := sets.NewString()
 	for _, b := range clusterBackups {
 		// Skip the Backup CR which is created from the local cluster and
 		// the snapshot backup hasn't be completed or pulled from the remote backup target yet
 		if b.Spec.SnapshotName != "" && b.Status.State != longhorn.BackupStateCompleted {
+			if b.Status.State == longhorn.BackupStateError || b.Status.State == longhorn.BackupStateUnknown {
```
innobead, line 299: ref: https://github.com/longhorn/longhorn-manager/blob/c3e9b1c8c7d1043134bfbf2c8ceb25c789935bc0/controller/backup_controller.go#L359-L359

For now, the code uses `BackupStateError` and `BackupStateUnknown` as failed backup.

controller/backup_volume_controller.go
```diff
@@ -284,11 +284,26 @@ func (bvc *BackupVolumeController) reconcile(backupVolumeName string) (err error
...
+				// Failed backup `LastSyncedAt` should not be updated after it was marked as `Error` or `Unknown`
+				if failedBackupTTL > 0 && time.Now().After(b.Status.LastSyncedAt.Add(time.Duration(failedBackupTTL)*time.Minute)) {
+					if err = bvc.ds.DeleteBackup(b.Name); err != nil {
```
innobead, line 302: Can we have an event for this deletion? so, at least users will know from the event as well instead of silent deletion.

innobead (changes requested):
Some minor feedbacks.

types/setting.go
```diff
@@ -255,6 +258,18 @@ var (
 		Default:     "300",
 	}
 
+	SettingDefinitionFailedBackupTTL = SettingDefinition{
+		DisplayName: "Failed Backup Time to Live",
+		Description: "In minutes. This setting determines how long Longhorn will keep the backup resource that was failed. Set to 0 to disable the auto-deletion.\n" +
+			"Failed backups will be checked and cleaned up during backupstore polling which is controlled by **BackupstorePollInterval** setting. " +
+			"Disabling **Backupstore Poll Interval** also means to disable failed backup auto-deletion.\n\n",
```
innobead, line 265: Use **Backupstore Poll Interval** instead of **BackupstorePollInterval**, just for consistency.

controller/backup_volume_controller.go
mantissahz: Removed. Thanks.

controller/backup_volume_controller.go
mantissahz: No, removed. Thanks

controller/backup_volume_controller.go
mantissahz: Yes, it is now.

controller/backup_volume_controller.go
mantissahz: Added for kind `BackupVolume`, or just for kind `Backup`?

types/setting.go
mantissahz: Modified. Thanks.

controller/backup_volume_controller.go
mantissahz: Moved the event log of deleting failed backup to backup controller.

types/setting.go
```diff
@@ -255,6 +258,18 @@ var (
 		Default:     "300",
 	}
 
+	SettingDefinitionFailedBackupTTL = SettingDefinition{
+		DisplayName: "Failed Backup Time to Live",
+		Description: "In minutes. This setting determines how long Longhorn will keep the backup resource that was failed. Set to 0 to disable the auto-deletion.\n" +
+			"Failed backups will be checked and cleaned up during backupstore polling which is controlled by **Backupstore Poll Interval** setting. " +
```
shuo-wu, line 264: Hence this value determines the minimal wait interval of the cleanup. And the actual cleanup interval is multiple of **Backupstore Poll Interval**

controller/backup_volume_controller.go
shuo-wu: Not sure why you prefer to delete the latter part... I think we should retain it but change the log level to warning instead.

types/setting.go
mantissahz: Yes, it is.

mantissahz:
>Not sure why you prefer to delete the latter part... I think we should retain it but change the log level to warning instead.

@shuo-wu did you mean to keep this part "and it will skip the auto-deletion for the failed backups?"

types/setting.go
mantissahz: suggest to add this sentence into description, right?

shuo-wu: Yes

shuo-wu:
> @shuo-wu did you mean to keep this part "and it will skip the auto-deletion for the failed backups?"

Right. This part is what I suggest in the previous comment. But remember to modify the log level.

cc @derekbit What's the reason for removing this part?

derekbit:
> > @shuo-wu did you mean to keep this part "and it will skip the auto-deletion for the failed backups?"
> 
> Right. This part is what I suggest in the previous comment. But remember to modify the log level.
> 
> cc @derekbit What's the reason for removing this part?

Ah, sorry. I was not aware of that it is your recommendation.
Just thought the error message is enough for debugging. But yes, we can keep it here.

types/setting.go
mantissahz: Added. Thanks.

controller/backup_volume_controller.go
mantissahz: Added back and change the log level to warning. Thanks.

controller/backup_volume_controller.go
```diff
@@ -284,11 +284,24 @@ func (bvc *BackupVolumeController) reconcile(backupVolumeName string) (err error
 		return err
 	}
 
+	// Get `Time to live` after failed backup was marked as `Error` or `Unknown`,
+	failedBackupTTL, err := bvc.ds.GetSettingAsInt(types.SettingNameFailedBackupTTL)
+	if err != nil {
+		log.WithError(err).Warnf("failed to get %v setting, and it will skip the auto-deletion for the failed backups", types.SettingNameFailedBackupTTL)
```
c3y1huang, line 290:
```suggestion
		log.WithError(err).Warnf("Failed to get %v setting, and it will skip the auto-deletion for the failed backups", types.SettingNameFailedBackupTTL)
```
nit: only the error message should be lowercased.

controller/backup_controller.go
```diff
@@ -299,6 +299,10 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
 		// Disable monitor regardless of backup state
 		bc.disableBackupMonitor(backup.Name)
 
+		if backup.Status.State == longhorn.BackupStateError || backup.Status.State == longhorn.BackupStateUnknown {
+			bc.eventRecorder.Eventf(backup, corev1.EventTypeWarning, string(backup.Status.State), "Failed backup %s has been deleted: %s", backup.Name, backup.Status.Error)
```
c3y1huang, line 303: Is this event statement still true when failed-backup-ttl is disabled?

mantissahz: Yes, I think it is still true, when users delete a failed backup manually and failed-backup-ttl is disabled.
Should we record this event only when failed-backup-ttl is enabled?

shuo-wu:
> Ah, sorry. I was not aware of that is your suggestion. Just thought the simple error message is enough for debugging. But yes, we can keep it ("and it will skip the auto-deletion for the failed backups") here.

Never mind. I think it can help avoid a potential misunderstanding that this error will fail the reconciliation

innobead (approved):

derekbit (approved):

innobead:
@mantissahz Please backport this to 1.3.x and 1.2.x.

PR #1481, opened 2022-08-18
Config codecov.yml to prevent codecov status check blocking pull request.

yangchiu:
Config codecov.yml to prevent codecov status check blocking pull request.

For https://github.com/longhorn/longhorn/issues/4425

Signed-off-by: Yang Chiu <yang.chiu@suse.com>

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,7 @@
+comment: off
+coverage:
+  status:
+    project:
+      default:
+        informational: true
+    patch: off


codecov.yml
```diff
@@ -0,0 +1,18 @@
...
+        base: auto
+        flags:
+          - unit
+        paths:
+          - "src"
+       # advanced settings
+        branches:
+          - master
+        if_ci_failed: error #success, failure, error, ignore
+        informational: false
```
innobead, line 17: Should information be `true`?

yangchiu: Sorry for the confusing. The old commit is just default values. I've updated it and it should disable both project and patch status check.

innobead: Thanks 👍 

Two things here for PR.
- Can we still see the coverage in each PR status check? but don't mark the checks failed even if the coverage decreases. (I feel the informational setting of the patch status can help)
- Given the first, don't add a coverage comment in the PR

yangchiu: I've modified it and it should disable comment and let PR status check always pass.

innobead (approved):

PR #1485, opened 2022-08-29
Use volume size as the ground truth while creating the engine process

derekbit:
[Longhorn 3601](https://github.com/longhorn/longhorn/issues/3601)

**Depends on https://github.com/longhorn/longhorn-engine/pull/740**
**Depends on https://github.com/longhorn/longhorn-instance-manager/pull/133**


shuo-wu (dismissed):

innobead (commented):
LGTM for CLI part, but need to add missing proxy part change before merging. cc @c3y1huang

c3y1huang (dismissed):

derekbit:
Is there method forcing users upgrade the volume's engine image?
cc @shuo-wu @innobead

joshimoo:
We really need to fix the engine binaries so that we can add arbitrary parameters which will be ignored instead of crashing.

One big reason to switch to the GRPC communication instead of the direct binary invocation was that we are able to change/improve the api in the long run.

PhanLe1010 (dismissed):
LGTM

PhanLe1010:
> We really need to fix the engine binaries so that we can add arbitrary parameters which will be ignored instead of crashing.
Great idea

derekbit:
The backward compatibility issue is addressed. Please help review the PR. Thank you.

innobead:
@derekbit https://github.com/longhorn/longhorn-instance-manager/pull/133 merged, please update this PR and also update related manifests in longhorn as well.

innobead (approved):
LGTM

PR #1493, opened 2022-09-16
Backup and Restore recurring jobs/groups

mantissahz:
Put all recurringJobs contents into the backup volume config file in the backup target.
Restore recurring jobs/groups of the backup volume when restoring a backup with the setting `restore-volume-recurring-jobs` enabled.

longhorn/longhorn#2227


controller/volume_controller.go
```diff
@@ -2952,6 +2953,95 @@ func (vc *VolumeController) ResolveRefAndEnqueue(namespace string, ref *metav1.O
...
+	}
+
+	return nil
+}
+
+//restoreVolumeRecurringJobs create recurring jobs/groups from the backup volume if needed when restoring a backup except DR volume.
+func (vc *VolumeController) restoreVolumeRecurringJobs(v *longhorn.Volume) error {
+	log := getLoggerForVolume(vc.logger, v)
+
+	// prevent re-creating a recurring job that the job name is existing and configuration is different.
```
shuo-wu, line 3012: Avoid recurring job restoration overrides the new recurring jobs added by users.

controller/volume_controller.go
```diff
@@ -2952,6 +2953,95 @@ func (vc *VolumeController) ResolveRefAndEnqueue(namespace string, ref *metav1.O
...
+			if _, err := ds.CreateRecurringJob(&longhorn.RecurringJob{ObjectMeta: metav1.ObjectMeta{Name: jobName}, Spec: job.JobSpec}); err != nil {
+				log.WithError(err).Warnf("Create recurring job %v", jobName)
+				return err
+			}
+		} else {
+			log.WithError(err).Warnf("Get recurring job %v", jobName)
+			return err
+		}
+	} else {
+		// create teh recurring job if job name is used and configuration is different.
```
shuo-wu, line 2971: NIT: typo `teh`

controller/volume_controller.go
```diff
@@ -2952,6 +2953,95 @@ func (vc *VolumeController) ResolveRefAndEnqueue(namespace string, ref *metav1.O
 	vc.enqueueVolume(volume)
 }
 
+// createRecurringJobNotExistAndLabel add recurring jobs and group back to a restoring volume and create recurring jobs if not exist.
+func createRecurringJobNotExistAndLabel(log *logrus.Entry, ds *datastore.DataStore, v *longhorn.Volume, jobName string, job *longhorn.RecordVolumeRecurringJob) error {
+	if existJob, err := ds.GetRecurringJob(jobName); err != nil {
+		// create the recurring job if not exist
+		if apierrors.IsNotFound(err) {
```
shuo-wu, line 2960:
```
if !apierrors.IsNotFound(err) {
	log.WithError(err).Warnf("Get recurring job %v", jobName)
	return err
}
```

controller/volume_controller.go
```diff
@@ -2952,6 +2953,95 @@ func (vc *VolumeController) ResolveRefAndEnqueue(namespace string, ref *metav1.O
...
+				}
+			}
+		}
+	}
+
+	// set Volume recurring jobs/groups
+	if job.FromJob {
+		labelKey := types.GetRecurringJobLabelKey(types.LonghornLabelRecurringJob, jobName)
+		if value, exist := v.Labels[labelKey]; !exist || value == types.LonghornLabelValueEnabled {
+			// disabled recurring jobs/group and it would be enabled after the backup restoration is completed.
```
shuo-wu, line 2993: Maybe we can blindly set it then skip the job creation if the volume restoration is in-progress.

shuo-wu (changes requested):
It's better to split the implementation into multiple commits.
For example, this PR can be split into the backup part and the restore part. You can try it next time. 😄

controller/backup_controller.go
```diff
@@ -393,6 +394,9 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
 		return nil
 	}
 
+	// Remove the Backup Volume recurring jobs/groups information.
+	delete(backupInfo.Labels, types.VolumeRecurringJobInfoLabel)
```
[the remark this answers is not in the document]
mantissahz: Recurring jobs/groups information is saving in the backup volume CR and backup volume configuration on remote backup target.
I think it should not be saved in every backup the recurring job or user created.

app/recurring_job.go
```diff
@@ -875,6 +875,10 @@ func filterVolumesForJob(allowDetached bool, volumes []longhorn.Volume, filterNa
 		if util.Contains(*filterNames, volume.Name) {
 			continue
 		}
+		if volume.Status.RestoreRequired {
+			logger.Debugf("Do not create job for %v volume in restoring the backup", volume.Name)
+			continue
+		}
```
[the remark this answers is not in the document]
mantissahz: This part is trying to prevent a volume which is restoring a backup data from taking a snapshot/backup.
`volume.Status.RestoreRequired` will be `false` after a backup restoration.

controller/backup_controller.go
```diff
@@ -587,6 +591,55 @@ func (bc *BackupController) hasMonitor(backupName string) *engineapi.BackupMonit
...
+	volumeRecurringJobSpecList := make(map[string]longhorn.RecordVolumeRecurringJob)
+
+	// distinguish volume labels between job and group.
+	jobs := datastore.MarshalLabelToVolumeRecurringJob(volume.Labels)
+	for jobName, job := range jobs {
+		if job.IsGroup && !util.Contains(jobGroups, jobName) {
+			jobGroups = append(jobGroups, jobName)
+		} else {
+			// get single job information if it is not a group.
+			if recurringJob, err := bc.ds.GetRecurringJob(jobName); err == nil {
```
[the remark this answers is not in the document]
mantissahz: Will it be more efficient? 
OK, I will try it.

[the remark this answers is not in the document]
shuo-wu: At least it would avoid some repeat/unnecessary deep copied in the get function.

controller/backup_controller.go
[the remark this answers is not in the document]
shuo-wu: It seems that `backupInfo` returned by the remote backup target contains both the backup volume info and the backup info. And you want to remove the redundancy, right?

app/recurring_job.go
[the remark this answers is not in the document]
shuo-wu: OK, Got it

controller/backup_controller.go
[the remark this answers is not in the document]
mantissahz: Yes

controller/volume_controller.go
```diff
@@ -2961,6 +3051,36 @@ func (vc *VolumeController) updateRecurringJobs(v *longhorn.Volume) (err error)
 	if err = datastore.FixupRecurringJob(v); err != nil {
 		return err
 	}
+
+	// DR volume will not restore volume recurring jobs/groups configuration.
+	if v.Spec.FromBackup != "" && !v.Spec.Standby {
+		// before backup restoration.
+		if !v.Status.RestoreInitiated && !v.Status.RestoreRecurringJob {
```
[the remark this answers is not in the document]
mantissahz: I would like to make it simple and not to break the flow of the backup restoration so I chose to restore/create recurring jobs/groups between the volume created and the backup restoration.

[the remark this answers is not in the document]
shuo-wu: I think creating Longhorn jobs does not matter. We don't need to add too many limits to this creation. Everything should be fine as long as Kubernetes Jobs is not created during the restoring.

controller/backup_controller.go
```diff
@@ -587,6 +591,55 @@ func (bc *BackupController) hasMonitor(backupName string) *engineapi.BackupMonit
 	return bc.monitors[backupName]
 }
 
+//getBackupVolumeRecurringJobs get recurring jobs in the volume labels and recurring jobs of groups in the volume labels.
+func (bc *BackupController) getBackupVolumeRecurringJobs(volume *longhorn.Volume) string {
+	var jobGroups []string
```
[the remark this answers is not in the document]
mantissahz: Removed. Thanks

controller/backup_controller.go
```diff
@@ -587,6 +591,55 @@ func (bc *BackupController) hasMonitor(backupName string) *engineapi.BackupMonit
...
+				bc.logger.WithError(err).Warnf("Cannot get recurring job %v", jobName)
+			}
+		}
+	}
+
+	// collect all recurring jobs with a group name.
+	allRecurringJobs, err := bc.ds.ListRecurringJobs()
+	if err != nil {
+		bc.logger.WithError(err).Warn("Cannot list all recurring jobs")
+	}
```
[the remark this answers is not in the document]
mantissahz: Added. Thanks

controller/backup_controller.go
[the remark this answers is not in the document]
mantissahz: Addressed, reuse all recurring jobs when scanning the groups and jobs

controller/backup_controller.go
```diff
@@ -587,6 +591,55 @@ func (bc *BackupController) hasMonitor(backupName string) *engineapi.BackupMonit
...
+
+	for jobName, job := range allRecurringJobs {
+		for _, jobGroupName := range job.Spec.Groups {
+			if util.Contains(jobGroups, jobGroupName) {
```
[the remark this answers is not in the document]
mantissahz: Modified. Thanks

controller/volume_controller.go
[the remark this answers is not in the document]
mantissahz: Removed RestoreRecurringJob flag.

controller/volume_controller.go
```diff
@@ -2961,6 +3051,36 @@ func (vc *VolumeController) updateRecurringJobs(v *longhorn.Volume) (err error)
...
+				logrus.WithError(err).Warnf("Failed to get %v setting", types.SettingNameRestoreVolumeRecurringJobs)
+			}
+			if v.Spec.RestoreVolumeRecurringJob == longhorn.RestoreVolumeRecurringJobEnabled || (v.Spec.RestoreVolumeRecurringJob != longhorn.RestoreVolumeRecurringJobDisabled && RestoreRecurringJobs) {
+				v.Status.RestoreRecurringJob = true
+				vc.restoreVolumeRecurringJobs(v)
+			}
+		}
+		// after backup restoration.
+		if v.Status.RestoreInitiated && v.Status.State == longhorn.VolumeStateDetached && v.Status.RestoreRecurringJob {
+			// enable recurring jobs/groups labeled when backup restorations started.
```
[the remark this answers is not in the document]
mantissahz: Modified Thanks

controller/volume_controller.go
```diff
@@ -2961,6 +3051,36 @@ func (vc *VolumeController) updateRecurringJobs(v *longhorn.Volume) (err error)
...
+			jobs := datastore.MarshalLabelToVolumeRecurringJob(v.Labels)
+			for jobName, job := range jobs {
+				labelType := types.LonghornLabelRecurringJob
+				if job.IsGroup {
+					labelType = types.LonghornLabelRecurringJobGroup
+				}
+				key := types.GetRecurringJobLabelKey(labelType, jobName)
+				v.Labels[key] = types.LonghornLabelValueEnabled
+			}
+			v.Status.RestoreRecurringJob = false
```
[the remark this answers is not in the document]
mantissahz: Removed RestoreRecurringJob flag

controller/volume_controller.go
mantissahz: Modified. Thanks

controller/volume_controller.go
mantissahz: Modified. Thanks

controller/volume_controller.go
```diff
@@ -2952,6 +2953,95 @@ func (vc *VolumeController) ResolveRefAndEnqueue(namespace string, ref *metav1.O
...
+	// prevent re-creating a recurring job that the job name is existing and configuration is different.
+	existingJobs := datastore.MarshalLabelToVolumeRecurringJob(v.Labels)
+	for jobName, job := range existingJobs {
+		if !job.IsGroup || jobName != longhorn.RecurringJobGroupDefault {
+			log.Debugf("Restoring volume recurring jobs had done. %v", existingJobs)
+			return nil
+		}
+	}
+
+	if !v.Spec.Standby && vc.controllerID == v.Status.OwnerID {
```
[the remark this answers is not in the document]
mantissahz: Modified. Thanks

controller/volume_controller.go
mantissahz: Modified. Thanks

controller/volume_controller.go
mantissahz: Modified. Thanks

controller/backup_controller.go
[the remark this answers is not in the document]
shuo-wu: I see. Actually, I am fine with storing the recurring job info in each Backup CRs since the recurring job config can be different between these backup CRs. If you just store the recurring job info in the BackupVolume CR only, it means you will blindly apply the latest recurring job config to any backups.

controller/volume_controller.go
```diff
@@ -2999,6 +3000,92 @@ func (vc *VolumeController) ResolveRefAndEnqueue(namespace string, ref *metav1.O
...
+	// Avoid recurring job restoration overrides the new recurring jobs added by users.
+	existingJobs := datastore.MarshalLabelToVolumeRecurringJob(v.Labels)
+	for jobName, job := range existingJobs {
+		if !job.IsGroup || jobName != longhorn.RecurringJobGroupDefault {
+			log.Debugf("Restoring volume recurring jobs had done")
+			return nil
+		}
+	}
+
+	if !v.Spec.Standby {
```
shuo-wu, line 3063: NIT: You can check this and directly return at the beginning of this function.

controller/volume_controller.go
```diff
@@ -3008,6 +3095,22 @@ func (vc *VolumeController) updateRecurringJobs(v *longhorn.Volume) (err error)
 	if err = datastore.FixupRecurringJob(v); err != nil {
 		return err
 	}
+
+	// DR volume will not restore volume recurring jobs/groups configuration.
+	if v.Spec.FromBackup != "" && !v.Spec.Standby {
```
shuo-wu, line 3100: NIT: Similar here.
```
if v.Spec.FromBackup == "" || v.Spec.Standby {
    return nil
}
```

controller/backup_controller.go
```diff
@@ -587,6 +591,47 @@ func (bc *BackupController) hasMonitor(backupName string) *engineapi.BackupMonit
...
+				}
+			} else if recurringJobName == jobName {
+				// save single job information if it is not a group.
+				if recordJob, exist := volumeRecurringJobSpecList[recurringJobName]; !exist {
+					volumeRecurringJobSpecList[recurringJobName] = longhorn.RecordVolumeRecurringJob{JobSpec: recurringJob.Spec, FromJob: true, FromGroup: []string{}}
+				} else {
+					recordJob.FromJob = true
+					volumeRecurringJobSpecList[recurringJobName] = recordJob
+				}
+			}
```
shuo-wu, line 624: Add more comments or try to abstract some functions for this nested part. It's not easy-to-understand.

deploy/install/01-prerequisite/03-crd.yaml
```diff
@@ -3011,6 +3013,8 @@ spec:
                 type: boolean
               restoreRequired:
                 type: boolean
+              restoreRecurringJob:
+                type: boolean
```
shuo-wu, line 3017: Remove this.

controller/volume_controller.go
```diff
@@ -3008,6 +3095,22 @@ func (vc *VolumeController) updateRecurringJobs(v *longhorn.Volume) (err error)
...
+		if !v.Status.RestoreInitiated {
+			RestoreRecurringJobs, err := vc.ds.GetSettingAsBool(types.SettingNameRestoreVolumeRecurringJobs)
+			if err != nil {
+				vc.logger.WithError(err).Warnf("Failed to get %v setting", types.SettingNameRestoreVolumeRecurringJobs)
+			}
+			if v.Spec.RestoreVolumeRecurringJob == longhorn.RestoreVolumeRecurringJobEnabled || (v.Spec.RestoreVolumeRecurringJob != longhorn.RestoreVolumeRecurringJobDisabled && RestoreRecurringJobs) {
```
shuo-wu, line 3106: Do we still need this?

controller/volume_controller.go
```diff
@@ -3008,6 +3095,22 @@ func (vc *VolumeController) updateRecurringJobs(v *longhorn.Volume) (err error)
...
+				if err := vc.restoreVolumeRecurringJobs(v); err != nil {
```
shuo-wu, line 3107: One question: For **each** reconcile loop of each restored volume, will the volume controller unmarshal the recurring job label then compare the result with existing recurring jobs? If YES, I would prefer to add the status field back to avoid this kind of repeat & unnecessary checks.

shuo-wu (changes requested):

controller/backup_controller.go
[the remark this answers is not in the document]
mantissahz: Yes, after some discussions, now it blindly stores the latest recurring jobs of volume only.

controller/volume_controller.go
```diff
@@ -2999,6 +3000,92 @@ func (vc *VolumeController) ResolveRefAndEnqueue(namespace string, ref *metav1.O
...
+	if job.FromJob {
+		labelKey := types.GetRecurringJobLabelKey(types.LonghornLabelRecurringJob, jobName)
+		// disabled recurring jobs/group and it would be enabled after the backup restoration is completed.
+		v.Labels[labelKey] = types.LonghornLabelValueEnabled
+	}
+	for _, groupName := range job.FromGroup {
+		groupLabelKey := types.GetRecurringJobLabelKey(types.LonghornLabelRecurringJobGroup, groupName)
+		// disabled recurring jobs/group and it would be enabled after the backup restoration is completed.
+		v.Labels[groupLabelKey] = types.LonghornLabelValueEnabled
+	}
```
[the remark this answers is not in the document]
mantissahz: Sorry about the misleading of `// disabled recurring jobs/group and it would be enabled after the backup restoration is ...`, now it will be enabled at this moment.

controller/volume_controller.go
mantissahz: Yes, User could decide to restore recurring jobs or not for a specific backup restoration to override the global setting.

controller/volume_controller.go
mantissahz: No, only once, `v.Status.RestoreInitiated` would become `true` for good after restoration is initialized at `checkAndInitVolumeRestore`.

controller/volume_controller.go
shuo-wu: OK. Then it's better to use a spec field to control it, which is similar to field `volume.Spec.DataLocality`

controller/volume_controller.go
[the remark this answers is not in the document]
shuo-wu: Sorry... I am still confused about why we need this label... If enabling the recurring job restore should be already controlled by `v.Spec.RestoreVolumeRecurringJob` or the global setting...

controller/volume_controller.go
mantissahz: Yes, it is at a spec field `v.Spec.RestoreVolumeRecurringJob`.
I think you meant another flag `RestoringVolumeRecurringJob` in `volume.status` that had just been removed.

controller/volume_controller.go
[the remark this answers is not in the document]
mantissahz: `v.Spec.RestoreVolumeRecurringJob` is used to check if it needed to restore recurring jobs from remote backup target or not. And `labelKey/groupLabelKey` label is used to filter volumes that is set to the job/group when recurring job is start to work.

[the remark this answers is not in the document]
shuo-wu: Then when/where is this label used to filter volumes?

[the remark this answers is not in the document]
mantissahz: https://github.com/longhorn/longhorn-manager/blob/master/app/recurring_job.go#L136-L149
when recurring job (pod created by the cronjob) starts to do snapshot or backup.

controller/volume_controller.go
mantissahz: Modified. Thanks

controller/volume_controller.go
mantissahz: Returning `nil` directly would not update the changes `datasotre.FixupRecurringJob` above.
Modified to 
```
if v.Spec.FromBackup != "" && !v.Spec.Standby && !v.Status.RestoreInitiated {
```

controller/backup_controller.go
mantissahz: Added, and separate adding recurring jobs from group into another function.

deploy/install/01-prerequisite/03-crd.yaml
mantissahz: Removed. Thanks.

k8s/pkg/apis/longhorn/v1beta2/volume.go
```diff
@@ -99,6 +99,14 @@ const (
 	VolumeConditionReasonTooManySnapshots              = "TooManySnapshots"
 )
 
+type RestoreVolumeRecurringJobOpt string
```
shuo-wu, line 102: NIT: Just noticed that the builder validation is missing. But I don't know how to make the empty value as an enum option:
```
// +kubebuilder:validation:Enum=enabled;disabled
type RestoreVolumeRecurringJobOpt string
```

controller/volume_controller.go
```diff
@@ -3008,6 +3098,20 @@ func (vc *VolumeController) updateRecurringJobs(v *longhorn.Volume) (err error)
...
 		return err
 	}
+
+	// DR volume will not restore volume recurring jobs/groups configuration.
+	if v.Spec.FromBackup != "" && !v.Spec.Standby && !v.Status.RestoreInitiated {
+		RestoreRecurringJobs, err := vc.ds.GetSettingAsBool(types.SettingNameRestoreVolumeRecurringJobs)
+		if err != nil {
+			vc.logger.WithError(err).Warnf("Failed to get %v setting", types.SettingNameRestoreVolumeRecurringJobs)
+		}
+		if v.Spec.RestoreVolumeRecurringJob == longhorn.RestoreVolumeRecurringJobEnabled || (v.Spec.RestoreVolumeRecurringJob != longhorn.RestoreVolumeRecurringJobDisabled && RestoreRecurringJobs) {
```
shuo-wu, line 3108: You can abstract this as a separate function, e.g.,`shouldRestoreRecurringJobs()`

mantissahz: Separated. Thanks.

controller/volume_controller.go
shuo-wu: I think you can make this check as a part of `shouldRestoreRecurringJobs`

mantissahz: `v.Status.RestoreInitiated` will always be `true` when restoration is starting, and I want to restore recurring job only once.

types/setting.go
```diff
@@ -87,6 +87,7 @@ const (
 	SettingNameRecurringSuccessfulJobsHistoryLimit          = SettingName("recurring-successful-jobs-history-limit")
 	SettingNameRecurringFailedJobsHistoryLimit              = SettingName("recurring-failed-jobs-history-limit")
 	SettingNameDeletingConfirmationFlag                     = SettingName("deleting-confirmation-flag")
+	SettingNameRestoreVolumeRecurringJobs                   = SettingName("restore-volume-recurring-jobs")
```
derekbit, line 90: Remember update `deploy/install/01-prerequisite/04-default-setting.yaml` as well.

k8s/pkg/apis/longhorn/v1beta2/volume.go
```diff
@@ -99,6 +99,15 @@ const (
 	VolumeConditionReasonTooManySnapshots              = "TooManySnapshots"
 )
 
+// +kubebuilder:validation:Enum="";enabled;disabled
+type RestoreVolumeRecurringJobOpt string
+
+const (
+	RestoreVolumeRecurringJobDefault  = RestoreVolumeRecurringJobOpt("")
+	RestoreVolumeRecurringJobEnabled  = RestoreVolumeRecurringJobOpt("enabled")
+	RestoreVolumeRecurringJobDisabled = RestoreVolumeRecurringJobOpt("disabled")
```
derekbit, line 108: Should we update `""` to `disabled` during upgrade?

controller/volume_controller.go
```diff
@@ -2999,6 +3000,112 @@ func (vc *VolumeController) ResolveRefAndEnqueue(namespace string, ref *metav1.O
...
+		}
+		// create the recurring job if not exist
+		log.Debugf("Create the recurring job %v when restoring recurring jobs", jobName)
+		if _, err := ds.CreateRecurringJob(&longhorn.RecurringJob{ObjectMeta: metav1.ObjectMeta{Name: jobName}, Spec: job.JobSpec}); err != nil {
+			log.WithError(err).Warnf("Failed to create a recurring job %v", jobName)
+			return err
+		}
+	} else {
+		// create the recurring job if job name is used and configuration is different.
+		if !reflect.DeepEqual(existJob.Spec, job.JobSpec) {
```
derekbit, line 3018: nit: 5 layers of indentation. Suggest refactoring and reducing to 3 layers and make it easier to read.

engineapi/backup_monitor.go
```diff
@@ -13,6 +13,8 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/utils/clock"
 
+	"github.com/longhorn/longhorn-manager/types"
+
```
derekbit, line 17: Redundant.

controller/volume_controller.go
shuo-wu: What if the restore volume is just created and hasn't been init?

mantissahz: Moved this check into `shouldRestoreRecurringJobs`. Thanks.
Recurring jobs/groups will still be created if there is an error before `checkAndInitVolumeRestore` sets `RestoreInitiated` true, and it will skip creating recurring jobs/groups at next reconciliation

types/setting.go
mantissahz: Added. Thanks.

k8s/pkg/apis/longhorn/v1beta2/volume.go
mantissahz: Modified `""` to `"ignored"`
Will upgrading procedure do the restoration?

controller/volume_controller.go
mantissahz: Modified. Separated into a function.

engineapi/backup_monitor.go
mantissahz: `types` would be used below. It might not be removed.

k8s/pkg/apis/longhorn/v1beta1/recurringjob.go
```diff
@@ -16,12 +16,19 @@ type VolumeRecurringJob struct {
 	IsGroup bool   `json:"isGroup"`
 }
 
+// RecordVolumeRecurringJob defines the Longhorn recurring job information stroed into the backup volume configuration.
```
c3y1huang, line 19:
```suggestion
// RecordVolumeRecurringJob defines the Longhorn recurring job information stored in the backup volume configuration
```

deploy/install/01-prerequisite/03-crd.yaml
```diff
@@ -2852,6 +2852,8 @@ spec:
                 type: string
               fromBackup:
                 type: string
+              restoreVolumeRecurringJob:
+                type: string
```
c3y1huang, line 2856: Why not use boolean?

controller/volume_controller.go
```diff
@@ -2999,6 +3000,127 @@ func (vc *VolumeController) ResolveRefAndEnqueue(namespace string, ref *metav1.O
...
+	for _, groupName := range job.FromGroup {
+		groupLabelKey := types.GetRecurringJobLabelKey(types.LonghornLabelRecurringJobGroup, groupName)
+		// enable recurring groups
+		v.Labels[groupLabelKey] = types.LonghornLabelValueEnabled
+	}
+
+	return nil
+}
+
+//restoreVolumeRecurringJobs create recurring jobs/groups from the backup volume if needed when restoring a backup except DR volume.
```
c3y1huang, line 3060:
```suggestion
// restoreVolumeRecurringJobs create recurring jobs/groups from the backup volume when restoring from a backup, except for the DR volume
```

controller/volume_controller.go
```diff
@@ -2999,6 +3000,127 @@ func (vc *VolumeController) ResolveRefAndEnqueue(namespace string, ref *metav1.O
...
+func (vc *VolumeController) restoreVolumeRecurringJobs(v *longhorn.Volume) error {
+	log := getLoggerForVolume(vc.logger, v)
+
+	if v.Spec.Standby {
+		log.Debugf("DR volumes will not restore recurring jobs")
```
c3y1huang, line 3065:
```suggestion
		log.Debug("Cannot restore recurring job for DR volume")
```

controller/volume_controller.go
```diff
@@ -3008,6 +3130,15 @@ func (vc *VolumeController) updateRecurringJobs(v *longhorn.Volume) (err error)
 	if err = datastore.FixupRecurringJob(v); err != nil {
 		return err
 	}
+
+	// DR volume will not restore volume recurring jobs/groups configuration.
+	if vc.shouldRestoreRecurringJobs(v) {
+		if err := vc.restoreVolumeRecurringJobs(v); err != nil {
+			vc.logger.WithError(err).Warnf("Failed to restore the volume recurring jobs/groups")
```
c3y1huang, line 3137:
```suggestion
			vc.logger.WithError(err).Warn("Failed to restore the volume recurring jobs/groups")
```

controller/volume_controller.go
```diff
@@ -3008,6 +3130,15 @@ func (vc *VolumeController) updateRecurringJobs(v *longhorn.Volume) (err error)
...
+		}
+
```
c3y1huang, line 3139: Redundant blank line.

controller/backup_controller.go
```diff
@@ -587,6 +591,61 @@ func (bc *BackupController) hasMonitor(backupName string) *engineapi.BackupMonit
 	return bc.monitors[backupName]
 }
 
+//getBackupVolumeRecurringJobs get recurring jobs in the volume labels and recurring jobs of groups in the volume labels
```
c3y1huang, line 594:
```suggestion
// getBackupVolumeRecurringJobs get recurring jobs in the volume labels and recurring jobs of groups in the volume labels
```

controller/volume_controller.go
```diff
@@ -2999,6 +3000,127 @@ func (vc *VolumeController) ResolveRefAndEnqueue(namespace string, ref *metav1.O
 	vc.enqueueVolume(volume)
 }
 
+// createNewRecurringJobNameUsed use a new name to create a new recurring job that original name exists but configuration is different
+func createNewRecurringJobNameUsed(log *logrus.Entry, ds *datastore.DataStore, job *longhorn.RecordVolumeRecurringJob) (string, error) {
+	newJobName := util.RandomID()
+	if _, err := ds.GetRecurringJob(newJobName); err != nil {
```
c3y1huang, line 3006: Do we need to consider if the recurring job exists?

controller/volume_controller.go
```diff
@@ -2999,6 +3000,127 @@ func (vc *VolumeController) ResolveRefAndEnqueue(namespace string, ref *metav1.O
...
+
+	volumeRecurringJobsInfo := make(map[string]longhorn.RecordVolumeRecurringJob)
+	bvName, isExist := v.Labels[types.LonghornLabelBackupVolume]
+	if !isExist {
+		log.Warn("Cannot find the backup volume label")
+		return nil
+	}
+	bv, err := vc.ds.GetBackupVolumeRO(bvName)
+	if err != nil {
+		log.WithError(err).WithField("backupvolume", bvName).Warnf("Failed to get the backup volume info")
```
c3y1huang, line 3086:
```suggestion
		log.WithError(err).WithField("backupvolume", bvName).Warn("Failed to get the backup volume info")
```

c3y1huang (changes requested):

controller/volume_controller.go
```diff
@@ -2999,6 +3000,127 @@ func (vc *VolumeController) ResolveRefAndEnqueue(namespace string, ref *metav1.O
...
+			log.WithError(err).Warnf("Failed to get the recurring job %v", jobName)
+			return err
+		}
+		// create the recurring job if not exist
+		log.Debugf("Create the recurring job %v when restoring recurring jobs", jobName)
+		if _, err := ds.CreateRecurringJob(&longhorn.RecurringJob{ObjectMeta: metav1.ObjectMeta{Name: jobName}, Spec: job.JobSpec}); err != nil {
+			log.WithError(err).Warnf("Failed to create a recurring job %v", jobName)
+			return err
+		}
+	} else if !reflect.DeepEqual(existJob.Spec, job.JobSpec) {
```
c3y1huang, line 3035: What happens if the volume restores multiple times? Will this create a bunch of new recurring jobs?

deploy/install/01-prerequisite/03-crd.yaml
mantissahz: This flag is used to override the globe setting, so it might not be able to be a boolean.   Now they are `"ignored"`, `"enabled"` and `"disabled"`.

controller/volume_controller.go
mantissahz: Modified. Thanks.

controller/volume_controller.go
mantissahz: Modified. Thanks.

controller/volume_controller.go
mantissahz: Modified. Thanks.

controller/volume_controller.go
mantissahz: Removed. Thanks

controller/backup_controller.go
mantissahz: Modified. Thanks.

controller/volume_controller.go
mantissahz: Removed. Yeah, it seems that it does not need to check this, just create and return the error if it exists, and extend the re-create recurring job name to 16 chars with a prefix.

controller/volume_controller.go
mantissahz: Modified. Thanks.

controller/volume_controller.go
mantissahz: Yes, it will create a bunch of new recurring jobs.
Modified. Reuse the re-created recurring jobs restored from the backup target.

deploy/install/01-prerequisite/03-crd.yaml
c3y1huang: Got it. Thank you for the explaination.

upgrade/v12xto133/upgrade.go
```diff
@@ -0,0 +1,56 @@
...
+		}
+		return errors.Wrapf(err, "failed to list all existing Longhorn volumes during the volume upgrade")
+	}
+
+	for _, v := range volumes.Items {
+		if v.Spec.RestoreVolumeRecurringJob != "" {
+			continue
+		}
+
+		v.Spec.RestoreVolumeRecurringJob = longhorn.RestoreVolumeRecurringJobDefault
```
c3y1huang, line 46: Is it possible to put the conversion in the mutator?

controller/volume_controller.go
c3y1huang: https://github.com/longhorn/longhorn-manager/pull/1493#discussion_r1003899298

upgrade/v12xto133/upgrade.go
c3y1huang: Ok never mind, the mutator is only effective during create and update.

controller/volume_controller.go
```diff
@@ -3005,6 +3006,138 @@ func (vc *VolumeController) ResolveRefAndEnqueue(namespace string, ref *metav1.O
...
+		// create a new recurring job if job name is used and configuration is different.
+		newJobName, err := createNewRecurringJobNameUsed(log, ds, job)
```
c3y1huang, line 3054: Since you already have a `ds.CreateRecurringJob` in this function, suggest `createNewRecurringJobNameUsed` can return the RecurringJob obj and have it create here in the same function.

controller/volume_controller.go
```diff
@@ -3005,6 +3006,138 @@ func (vc *VolumeController) ResolveRefAndEnqueue(namespace string, ref *metav1.O
...
+
+	return nil
+}
+
+// restoreVolumeRecurringJobs create recurring jobs/groups from the backup volume when restoring from a backup, except for the DR volume
+func (vc *VolumeController) restoreVolumeRecurringJobs(v *longhorn.Volume) error {
+	log := getLoggerForVolume(vc.logger, v)
+
+	if v.Spec.Standby {
+		log.Debugf("Cannot restore recurring job for DR volume")
```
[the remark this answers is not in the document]
mantissahz: Modified. Thanks

upgrade/v12xto133/upgrade.go
mantissahz: Got it

types/setting.go
```diff
@@ -280,6 +283,16 @@ var (
 		Default:  "1440",
 	}
 
+	SettingDefinitionRestoreVolumeRecurringJobs = SettingDefinition{
+		DisplayName: "Restore Volume Recurring Jobs",
+		Description: "Restore recurring jobs from the backup volume on the backup target and create recurring jobs if not exist during a backup restoration.",
```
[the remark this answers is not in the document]
mantissahz: Added. Thanks.

k8s/pkg/apis/longhorn/v1beta2/recurringjob.go
```diff
@@ -17,14 +17,21 @@ type VolumeRecurringJob struct {
...
+}
+
 // RecurringJobSpec defines the desired state of the Longhorn recurring job
 type RecurringJobSpec struct {
 	// The recurring job name.
 	// +optional
 	Name string `json:"name"`
 	// The recurring job group.
 	// +optional
-	Groups []string `json:"groups,omitempty"`
```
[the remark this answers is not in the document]
mantissahz: Modified, not remove `omitempty`, set `job.JobSpec.Groups` to be `[]string{}` if `JobSpec.Groups` from backup target is `nil` 

It would get `[]string = nil` when decoding recurring job information from the backup target would be different to `[]string{}` of recurring jobs CR in the cluster judged by `reflect.DeepEqual` And then it would create a new recurring job every time because the field `groups` is not the same even they both are empty. 
It is equal when I encode the same structure and decode it locally.

controller/volume_controller.go
```diff
@@ -3005,6 +3006,138 @@ func (vc *VolumeController) ResolveRefAndEnqueue(namespace string, ref *metav1.O
...
+	if _, err := ds.CreateRecurringJob(&longhorn.RecurringJob{ObjectMeta: metav1.ObjectMeta{Name: newJobName}, Spec: job.JobSpec}); err != nil {
+		log.WithError(err).Warnf("Failed to create a new recurring job %v for restoring the job %v", newJobName, oldJobName)
+		return "", err
+	}
+
+	return newJobName, nil
+}
+
+// createRecurringJobNotExistAndLabel add recurring jobs and group back to a restoring volume and create recurring jobs if not exist
+func createRecurringJobNotExistAndLabel(log *logrus.Entry, ds *datastore.DataStore, v *longhorn.Volume, jobName string, job *longhorn.RecordVolumeRecurringJob) error {
```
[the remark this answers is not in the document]
mantissahz: Separated and renamed. Thanks

controller/volume_controller.go
mantissahz: Modified. Thanks

c3y1huang: I mean you can `ds.CreateRecurringJob` in the caller(`createRecurringJobNotExist`).

k8s/pkg/apis/longhorn/v1beta2/recurringjob.go
```diff
@@ -17,6 +17,13 @@ type VolumeRecurringJob struct {
 	IsGroup bool   `json:"isGroup"`
 }
 
+// RecordVolumeRecurringJob defines the Longhorn recurring job information stroed into the backup volume configuration.
```
c3y1huang, line 20:
```suggestion
// RecordVolumeRecurringJob defines the Longhorn recurring job information stored in the backup volume configuration
```

k8s/pkg/apis/longhorn/v1beta1/volume.go
```diff
@@ -101,6 +101,14 @@ const (
 	VolumeConditionReasonTooManySnapshots              = "TooManySnapshots"
 )
 
+type RestoreVolumeRecurringJobOpt string
```
c3y1huang, line 104: Why do we want these in v1beta1?

c3y1huang (changes requested):

k8s/pkg/apis/longhorn/v1beta1/volume.go
mantissahz: Do you mean it could be just "string" or we don't need to handle `restoreRecurringJob` for v1beta1 API?

k8s/pkg/apis/longhorn/v1beta2/recurringjob.go
mantissahz: Modified. Thanks.

controller/volume_controller.go
mantissahz: OK, `createNewRecurringJobNameUsed` is only used to create a new name or find a restored reuccuring jobs that its spec is the same to the job user want to restore.
And create recurring job only in `createRecurringJobNotExist` if needed.

k8s/pkg/apis/longhorn/v1beta1/volume.go
c3y1huang: I mean, why do we need make an update in `k8s/pkg/apis/longhorn/v1beta1/`?

mantissahz: You are right, we do not need to update `k8s/pkg/apis/longhorn/v1beta1/`.
Removed. Much appreciated.

upgrade/v12xto133/upgrade.go
```diff
@@ -0,0 +1,56 @@
...
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	clientset "k8s.io/client-go/kubernetes"
+
+	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
+	lhclientset "github.com/longhorn/longhorn-manager/k8s/pkg/client/clientset/versioned"
+)
+
+const (
+	upgradeLogPrefix = "upgrade from v1.2.x to v1.3.3: "
```
derekbit, line 18: Probably `from v1.3.2 to v1.3.3`?

controller/volume_controller.go
```diff
@@ -3005,6 +3006,146 @@ func (vc *VolumeController) ResolveRefAndEnqueue(namespace string, ref *metav1.O
...
+			log.WithError(err).Warnf("Failed to create a new recurring job %v", jobName)
+			return err
+		}
+		jobName = newJobName
+	} else {
+		return nil
+	}
+
+	// create the recurring job if not exist
+	log.Debugf("Create the recurring job %v when restoring recurring jobs", jobName)
```
derekbit, line 3069:
```suggestion
	log.Debugf("Creating the recurring job %v when restoring recurring jobs", jobName)
```

controller/backup_controller.go
```diff
@@ -587,6 +591,61 @@ func (bc *BackupController) hasMonitor(backupName string) *engineapi.BackupMonit
...
 }
 
+// getBackupVolumeRecurringJobs get recurring jobs in the volume labels and recurring jobs of groups in the volume labels
+func (bc *BackupController) getBackupVolumeRecurringJobs(volume *longhorn.Volume) string {
+	volumeRecurringJobSpecList := make(map[string]longhorn.RecordVolumeRecurringJob)
+
+	allRecurringJobs, err := bc.ds.ListRecurringJobsRO()
+	if err != nil {
+		bc.logger.WithError(err).Warn("Failed to list all recurring jobs")
+	}
```
derekbit, line 601: nit: Probably can improve the readability by
```suggestion
	allRecurringJobs, err := bc.ds.ListRecurringJobsRO()
	if err != nil {
		bc.logger.WithError(err).Warn("Failed to list all recurring jobs")
		return .......
	}
	
	volumeRecurringJobSpecList := make(map[string]longhorn.RecordVolumeRecurringJob)
	....
```

controller/backup_controller.go
```diff
@@ -587,6 +591,61 @@ func (bc *BackupController) hasMonitor(backupName string) *engineapi.BackupMonit
...
+	volumeRecurringJobInfo, err := json.Marshal(volumeRecurringJobSpecList)
+	if err != nil {
+		bc.logger.WithError(err).Warnf("Marshal volumeRecurringJobInfoMap: %v", volumeRecurringJobSpecList)
+	}
+
+	return string(volumeRecurringJobInfo)
+}
+
+// addRecurringJobFromGroup add recurring jobs in the group and add the group name into the recurring job information
+func addRecurringJobFromGroup(jobName, groupName string, jobSpec *longhorn.RecurringJobSpec, volumeRecurringJobInfo *map[string]longhorn.RecordVolumeRecurringJob) error {
```
derekbit, line 632: Prefer 
`func addRecurringJobFromGroup(jobName, groupName string, jobSpec *longhorn.RecurringJobSpec, volumeRecurringJobInfo map[string]longhorn.RecordVolumeRecurringJob) (map[string]longhorn.RecordVolumeRecurringJob, error) {
...
}`

controller/volume_controller.go
```diff
@@ -2842,6 +2842,7 @@ func (vc *VolumeController) checkForAutoDetachment(v *longhorn.Volume, e *longho
 	if (cliAPIVersion >= engineapi.CLIVersionFour && !isPurging && (v.Status.Robustness == longhorn.VolumeRobustnessHealthy || v.Status.Robustness == longhorn.VolumeRobustnessDegraded) && allScheduledReplicasIncluded) ||
 		(cliAPIVersion < engineapi.CLIVersionFour && (v.Status.Robustness == longhorn.VolumeRobustnessHealthy || v.Status.Robustness == longhorn.VolumeRobustnessDegraded)) {
 		log.Info("Prepare to do auto detachment for restore/DR volume")
+
```
derekbit, line 2845:
```suggestion
		log.Info("Preparing to do auto detachment for restore/DR volume")

```

controller/volume_controller.go
```diff
@@ -3005,6 +3006,146 @@ func (vc *VolumeController) ResolveRefAndEnqueue(namespace string, ref *metav1.O
...
+		if !strings.HasPrefix(existJobName, types.VolumeRecurringJobRestorePrefix) {
+			continue
+		}
+		job.JobSpec.Name = existJob.Spec.Name
+		if reflect.DeepEqual(existJob.Spec, job.JobSpec) {
+			return existJobName, nil
+		}
+	}
+
+	newJobName := types.VolumeRecurringJobRestorePrefix + util.RandomID() + util.RandomID()
```
derekbit, line 3026: nit: add a helper function for it
```suggestion
	newJobName := generateJobName(types.VolumeRecurringJobRestorePrefix)
```

controller/volume_controller.go
```diff
@@ -3005,6 +3006,146 @@ func (vc *VolumeController) ResolveRefAndEnqueue(namespace string, ref *metav1.O
...
+		return nil
+	}
+	bv, err := vc.ds.GetBackupVolumeRO(bvName)
+	if err != nil {
+		log.WithError(err).WithField("backupvolume", bvName).Warn("Failed to get the backup volume info")
+		return err
+	}
+
+	if volumeRecurringJobInfoStr, exist := bv.Status.Labels[types.VolumeRecurringJobInfoLabel]; exist {
+		json.Unmarshal([]byte(volumeRecurringJobInfoStr), &volumeRecurringJobsInfo)
```
derekbit, line 3109: error check is missing

engineapi/backup_monitor.go
derekbit: Got it!

engineapi/backup_monitor.go
```diff
@@ -71,6 +73,9 @@ func NewBackupMonitor(logger logrus.FieldLogger,
 
 	// Call engine API snapshot backup
 	if backup.Status.State == longhorn.BackupStateNew {
+		// put volume recurring jobs/groups information into backup labels and it would be stored in the file `volume.cfg`
+		backup.Spec.Labels[types.VolumeRecurringJobInfoLabel] = volumeRecurringJobInfo
```
derekbit, line 77: Check the nullity before adding a new lable.

upgrade/v12xto133/upgrade.go
mantissahz: Yeah, modified. Thanks

controller/volume_controller.go
mantissahz: Modified. Thanks

controller/backup_controller.go
mantissahz: Modified and returned error. Thanks

controller/backup_controller.go
mantissahz: Modified. Thanks

controller/volume_controller.go
mantissahz: Modified. Thanks

controller/volume_controller.go
mantissahz: Added. Thanks

controller/volume_controller.go
mantissahz: Modified and checked. Thanks

engineapi/backup_monitor.go
mantissahz: Modified and checked. Thanks

controller/volume_controller.go
shuo-wu: How do you guarantee the function won't repeatedly restore a recurring job of which the name is occupied?

shuo-wu (commented):
It's highly recommend to divide this PR into 2 commits: backup part and restore part

innobead:
> It's highly recommend to divide this PR into 2 commits: backup part and restore part

Or 3 parts, the other is the resource structure change and corresponding auto-generated parts.

controller/backup_controller.go
```diff
@@ -393,6 +394,9 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
 		return nil
 	}
 
+	// Remove the Backup Volume recurring jobs/groups information.
```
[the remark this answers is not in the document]
mantissahz: Added.

controller/backup_controller.go
```diff
@@ -587,6 +591,63 @@ func (bc *BackupController) hasMonitor(backupName string) *engineapi.BackupMonit
...
+
+	volumeRecurringJobSpecList := make(map[string]longhorn.RecordVolumeRecurringJob)
+	// distinguish volume labels between job and group.
+	jobs := datastore.MarshalLabelToVolumeRecurringJob(volume.Labels)
+	for recurringJobName, recurringJob := range allRecurringJobs {
+		for jobName, job := range jobs {
+			if job.IsGroup {
+				// add the recurring job from group or update its groups informtaion
+				if volumeRecurringJobSpecList, err = addRecurringJobFromGroup(recurringJobName, jobName, &recurringJob.Spec, volumeRecurringJobSpecList); err != nil {
+					bc.logger.WithError(err).WithFields(logrus.Fields{"recurringjob": recurringJobName, "group": jobName}).Warn("Failed to add the recurring job")
```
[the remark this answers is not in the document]
mantissahz: Yes, thanks.
Added

controller/volume_controller.go
mantissahz: It would search the recurring jobs having the name with the prefix `restored-recurring-job-` and compare the Spec before create a new recurring job with the prefix.

api/volume.go
```diff
@@ -127,6 +127,10 @@ func (s *Server) VolumeCreate(rw http.ResponseWriter, req *http.Request) error {
 		}
 	}
 
+	if volume.RestoreVolumeRecurringJob == "" {
+		volume.RestoreVolumeRecurringJob = longhorn.RestoreVolumeRecurringJobDefault
+	}
```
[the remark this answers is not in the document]
mantissahz: Moved. Thanks

controller/backup_controller.go
```diff
@@ -587,6 +591,63 @@ func (bc *BackupController) hasMonitor(backupName string) *engineapi.BackupMonit
...
+// addRecurringJobFromGroup add recurring jobs in the group and add the group name into the recurring job information
+func addRecurringJobFromGroup(jobName, groupName string, jobSpec *longhorn.RecurringJobSpec, volumeRecurringJobInfo map[string]longhorn.RecordVolumeRecurringJob) (map[string]longhorn.RecordVolumeRecurringJob, error) {
+	if !util.Contains(jobSpec.Groups, groupName) {
+		// this job is not in this group then do nothing
+		return volumeRecurringJobInfo, nil
+	}
+	if recordJob, exist := volumeRecurringJobInfo[jobName]; !exist {
+		// store new recurring job information in this group
+		volumeRecurringJobInfo[jobName] = longhorn.RecordVolumeRecurringJob{JobSpec: *jobSpec, FromGroup: []string{groupName}}
+	} else if !util.Contains(volumeRecurringJobInfo[jobName].FromGroup, groupName) {
```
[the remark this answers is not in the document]
mantissahz: Modified. Thanks

controller/volume_controller.go
```diff
@@ -3010,6 +3011,154 @@ func (vc *VolumeController) ResolveRefAndEnqueue(namespace string, ref *metav1.O
...
+}
+
+// restoreVolumeRecurringJobs create recurring jobs/groups from the backup volume when restoring from a backup, except for the DR volume
+func (vc *VolumeController) restoreVolumeRecurringJobs(v *longhorn.Volume) error {
+	log := getLoggerForVolume(vc.logger, v)
+
+	if v.Spec.Standby {
+		log.Debug("Cannot restore recurring job for DR volume")
+		return nil
+	}
```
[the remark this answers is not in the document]
mantissahz: Removed.

controller/volume_controller.go
```diff
@@ -3010,6 +3011,154 @@ func (vc *VolumeController) ResolveRefAndEnqueue(namespace string, ref *metav1.O
...
+
+	// Avoid recurring job restoration overrides the new recurring jobs added by users.
+	existingJobs := datastore.MarshalLabelToVolumeRecurringJob(v.Labels)
+	for jobName, job := range existingJobs {
+		if !job.IsGroup || jobName != longhorn.RecurringJobGroupDefault {
+			log.Debug("Restoring volume recurring jobs had done")
```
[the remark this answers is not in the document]
mantissahz: Changed the log and moved into `shouldRestoreRecurringJobs`

controller/volume_controller.go
```diff
@@ -3010,6 +3011,154 @@ func (vc *VolumeController) ResolveRefAndEnqueue(namespace string, ref *metav1.O
...
+			return nil
+		}
+	}
+
+	volumeRecurringJobsInfo := make(map[string]longhorn.RecordVolumeRecurringJob)
```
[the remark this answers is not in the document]
mantissahz: Changed.

controller/backup_controller.go
```diff
@@ -587,6 +591,63 @@ func (bc *BackupController) hasMonitor(backupName string) *engineapi.BackupMonit
...
+func (bc *BackupController) getBackupVolumeRecurringJobs(volume *longhorn.Volume) (string, error) {
+	allRecurringJobs, err := bc.ds.ListRecurringJobsRO()
+	if err != nil {
+		bc.logger.WithError(err).Warn("Failed to list all recurring jobs")
+		return "", err
+	}
+
+	volumeRecurringJobSpecList := make(map[string]longhorn.RecordVolumeRecurringJob)
+	// distinguish volume labels between job and group.
+	jobs := datastore.MarshalLabelToVolumeRecurringJob(volume.Labels)
```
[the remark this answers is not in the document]
shuo-wu: For me YES.

controller/backup_controller.go
```diff
@@ -601,7 +662,13 @@ func (bc *BackupController) enableBackupMonitor(backup *longhorn.Backup, volume
 		return nil, err
 	}
 
-	monitor, err = engineapi.NewBackupMonitor(bc.logger, backup, volume, backupTargetClient, biChecksum, engine, engineClientProxy, bc.enqueueBackupForMonitor)
+	// volumeRecurringJobInfo could be "".
+	volumeRecurringJobInfo, err := bc.getBackupVolumeRecurringJobs(volume)
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
mantissahz: Addressed.

controller/volume_controller.go
```diff
@@ -3022,6 +3022,153 @@ func (vc *VolumeController) ResolveRefAndEnqueue(namespace string, ref *metav1.O
...
+		log.Warn("Cannot find the backup volume label")
+		return nil
+	}
+	bv, err := vc.ds.GetBackupVolumeRO(bvName)
+	if err != nil {
+		log.WithError(err).WithField("backupvolume", bvName).Warn("Failed to get the backup volume info")
+		return err
+	}
+
+	if volumeRecurringJobInfoStr, exist := bv.Status.Labels[types.VolumeRecurringJobInfoLabel]; exist {
```
shuo-wu, line 3119: NIT:
```
if !exist {
    return nil
}
```

controller/volume_controller.go
```diff
@@ -3022,6 +3022,153 @@ func (vc *VolumeController) ResolveRefAndEnqueue(namespace string, ref *metav1.O
...
+		if err := json.Unmarshal([]byte(volumeRecurringJobInfoStr), &backupVolumeRecurringJobsInfo); err != nil {
+			log.WithError(err).WithField("backupvolume", bvName).Warn("Failed to unmarshal information of volume recurring jobs")
```
shuo-wu, line 3121: NIT: This will fail the function. Which log-level should we use, warning or error?

shuo-wu (commented):
In general LGTM

controller/volume_controller.go
mantissahz: Added.

controller/volume_controller.go
mantissahz: Modified. Thanks.

k8s/pkg/apis/longhorn/v1beta2/recurringjob.go
```diff
@@ -17,6 +17,13 @@ type VolumeRecurringJob struct {
 	IsGroup bool   `json:"isGroup"`
 }
 
+// RecordVolumeRecurringJob defines the Longhorn recurring job information stored in the backup volume configuration
+type RecordVolumeRecurringJob struct {
```
innobead, line 21: Rename to `VolumeRecurringJobInfo`

engineapi/backup_monitor.go
```diff
@@ -109,6 +123,68 @@ func NewBackupMonitor(logger logrus.FieldLogger,
 	return m, nil
 }
 
+// getBackupVolumeRecurringJobs get recurring jobs in the volume labels and recurring jobs of groups in the volume labels
+func (m *BackupMonitor) getBackupVolumeRecurringJobs(ds *datastore.DataStore, volume *longhorn.Volume) (string, error) {
```
innobead, line 127: Rename to `getVolumeRecurringJobInfos`

k8s/pkg/apis/longhorn/v1beta2/volume.go
```diff
@@ -110,6 +110,15 @@ const (
 	SnapshotDataIntegrityFastCheck = SnapshotDataIntegrity("fast-check")
 )
 
+// +kubebuilder:validation:Enum=ignored;enabled;disabled
+type RestoreVolumeRecurringJobOpt string
```
innobead, line 114: Rename to `RestoreVolumeRecurringJobType`

engineapi/backup_monitor.go
```diff
@@ -109,6 +123,68 @@ func NewBackupMonitor(logger logrus.FieldLogger,
...
+	allRecurringJobs, err := ds.ListRecurringJobsRO()
+	if err != nil {
+		m.logger.WithError(err).Warn("Failed to list all recurring jobs")
+		return "", err
+	}
+
+	volumeRecurringJobSpecList := make(map[string]longhorn.RecordVolumeRecurringJob)
```
innobead, line 134: The naming can be better, so it will be easy to connect to the method purpose.

```go
volumeRecurringJobInfos := make(map[string]longhorn.RecordVolumeRecurringInfo)
```

engineapi/backup_monitor.go
```diff
@@ -109,6 +123,68 @@ func NewBackupMonitor(logger logrus.FieldLogger,
...
+					volumeRecurringJobSpecList[recurringJobName] = longhorn.RecordVolumeRecurringJob{JobSpec: recurringJob.Spec, FromJob: true, FromGroup: []string{}}
+				} else {
+					// recurring job has been saved by groups and update it from job as well.
+					recordJob.FromJob = true
+					volumeRecurringJobSpecList[recurringJobName] = recordJob
+				}
+			}
+		}
+	}
+	volumeRecurringJobInfo, err := json.Marshal(volumeRecurringJobSpecList)
```
innobead, line 157: `volumeRecurringJobInfosBytes, err := json.Marshal(volumeRecurringJobInfos)` or even `volumeRecurringJobInfos, err = json.Marshal(volumeRecurringJobInfos)`

engineapi/backup_monitor.go
```diff
@@ -109,6 +123,68 @@ func NewBackupMonitor(logger logrus.FieldLogger,
...
+	if err != nil {
+		m.logger.WithError(err).Warnf("Marshal volumeRecurringJobInfoMap: %v", volumeRecurringJobSpecList)
+		return "", err
+	}
+
+	return string(volumeRecurringJobInfo), nil
+}
+
+// addRecurringJobFromGroup add recurring jobs in the group and add the group name into the recurring job information
+func addRecurringJobFromGroup(jobName, groupName string, jobSpec *longhorn.RecurringJobSpec, volumeRecurringJobInfo map[string]longhorn.RecordVolumeRecurringJob) (map[string]longhorn.RecordVolumeRecurringJob, error) {
```
innobead, line 167: Rename to `addVolumeRecurringJobInfosFromGroup`.

app/recurring_job.go
```diff
@@ -876,6 +876,10 @@ func filterVolumesForJob(allowDetached bool, volumes []longhorn.Volume, filterNa
 		if util.Contains(*filterNames, volume.Name) {
 			continue
 		}
+		if volume.Status.RestoreRequired {
+			logger.Debugf("Do not create job for %v volume in restoring the backup", volume.Name)
```
innobead, line 880: Change to `Bypassed to create job for %v volume during restoring from the backup`

controller/volume_controller.go
```diff
@@ -3022,6 +3022,159 @@ func (vc *VolumeController) ResolveRefAndEnqueue(namespace string, ref *metav1.O
 	vc.enqueueVolume(volume)
 }
 
+// generateJobName returns a name that contains prefix and 16 random characters
+func generateJobName(prefix string) string {
```
innobead, line 3026: This function is just used by `generateRecurringJobName`, so suggest merge into `generateRecurringJobName`

types/setting.go
```diff
@@ -301,6 +304,21 @@ var (
 		Default:  "1440",
 	}
 
+	SettingDefinitionRestoreVolumeRecurringJobs = SettingDefinition{
+		DisplayName: "Restore Volume Recurring Jobs",
+		Description: "Restore recurring jobs from the backup volume on the backup target and create recurring jobs if not exist during a backup restoration.\n\n" +
+			"Longhorn also support individual volume setting. The setting can be specified on Backup page when making a backup restoration, this overrules the global setting.\n\n" +
```
innobead, line 310: `Longhorn also supports ...`

innobead (changes requested):

types/setting.go

controller/volume_controller.go
mantissahz: Merged into the caller function `generateRecurringJobName`.

app/recurring_job.go
mantissahz: Modified. Thanks

engineapi/backup_monitor.go
mantissahz: Modified and all related codes. Thanks

engineapi/backup_monitor.go
mantissahz: Modified and all related codes. Thanks

engineapi/backup_monitor.go
mantissahz: Modified and all related codes. Thanks

k8s/pkg/apis/longhorn/v1beta2/volume.go
mantissahz: Modified and all related codes. Thanks

engineapi/backup_monitor.go
mantissahz: Modified and all related codes. Thanks

k8s/pkg/apis/longhorn/v1beta2/recurringjob.go
mantissahz: Modified and all related codes. Thanks

engineapi/backup_monitor.go
```diff
@@ -109,6 +123,68 @@ func NewBackupMonitor(logger logrus.FieldLogger,
...
+
+	volumeRecurringJobInfos := make(map[string]longhorn.VolumeRecurringJobInfo)
+	// distinguish volume labels between job and group.
+	volumeSelectedRecurringJobs := datastore.MarshalLabelToVolumeRecurringJob(volume.Labels)
+	for recurringJobName, recurringJob := range allRecurringJobs {
+		for jobName, job := range volumeSelectedRecurringJobs {
+			if job.IsGroup {
+				// add the recurring job from group or update its groups informtaion
+				if volumeRecurringJobInfos, err = addVolumeRecurringJobInfosFromGroup(recurringJobName, jobName, &recurringJob.Spec, volumeRecurringJobInfos); err != nil {
+					m.logger.WithError(err).WithFields(logrus.Fields{"recurringjob": recurringJobName, "group": jobName}).Warn("Failed to add the recurring job")
```
innobead, line 142: nit: Suggest reviewing `when we should print logs`? the basic pattern is as below (no need to do it right now, but we should review this in general later to make our logging concise)

- If the error will be returned to the caller, it should not be printed out except some extra logging
- If the error will not be returned to caller, need to print the log directly

innobead (approved):

PR #1495, opened 2022-09-20
Use 4096 bytes as default block size of the xfs and ext4 filesystem 

derekbit:
https://github.com/longhorn/longhorn/issues/4594

### DIFF
--- a/csi/node_server.go
+++ b/csi/node_server.go
@@ -32,7 +32,22 @@ const (
 	// We currently only support passphrase retrieval via direct secret values
 	CryptoKeyProvider = "CRYPTO_KEY_PROVIDER"
 	CryptoKeyValue    = "CRYPTO_KEY_VALUE"
+
+	defaultFsType = "ext4"
 )
+
+type fsParameters struct {
+	formatParameters string
+}
+
+var supportedFs = map[string]fsParameters{
+	"ext4": {
+		formatParameters: "-b4096",
+	},
+	"xfs": {
+		formatParameters: "-ssize=4096 -bsize=4096",
+	},
+}
 
 type NodeServer struct {
 	apiClient *longhornclient.RancherClient
@@ -363,11 +378,10 @@ func (ns *NodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
 		return &csi.NodeStageVolumeResponse{}, nil
 	}
 
-	// mounter uses ext4 by default
 	options := volumeCapability.GetMount().GetMountFlags()
 	fsType := volumeCapability.GetMount().GetFsType()
 	if fsType == "" {
-		fsType = "ext4"
+		fsType = defaultFsType
 	}
 
 	formatMounter, ok := mounter.(*mount.SafeFormatAndMount)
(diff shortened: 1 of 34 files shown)


weizhe0422:
LGTM

It can support xfs and ext4 fs through different length of arguments.

derekbit:
Ready for review.

cc @PhanLe1010 @shuo-wu @innobead

csi/node_server.go
```diff
@@ -615,21 +628,27 @@ func (ns *NodeServer) getMounter(volume *longhornclient.Volume, volumeCapability
...
 		}
 
 		mounter := &mount.SafeFormatAndMount{Interface: mount.New(""), Exec: utilexec.New()}
-		// we allow the user to provide additional params for ext4 filesystem creation.
-		// this allows an ext4 fs to be mounted on older kernels, see https://github.com/longhorn/longhorn/issues/1208
-		if fsType == "ext4" && userExt4Params != nil && userExt4Params.Value != "" {
-			ext4Params := userExt4Params.Value
-			logrus.Infof("volume %v using user provided ext4 fs creation params: %s", volume.Name, ext4Params)
-			cmdParamMapping := map[string]string{"mkfs." + fsType: ext4Params}
+		if fsType == "ext4" || fsType == "xfs" {
```
shuo-wu, line 649: Print out warning logs if the fs type is unrecognized.

shuo-wu (commented):
In general LGTM

csi/node_server.go
derekbit: Updated, thank you.

csi/node_server.go
```diff
@@ -615,25 +628,33 @@ func (ns *NodeServer) getMounter(volume *longhornclient.Volume, volumeCapability
...
+			fsType = DefaultFsType
+		}
+
+		// To allow users to override the default block size,
+		// put the default block size in front of other user-defined parameters.
+		params := getBlockSizeParams(fsType)
+
+		// We allow the user to provide additional params for ext4 filesystem creation.
+		// This allows an ext4 fs to be mounted on older kernels, see https://github.com/longhorn/longhorn/issues/1208
+		if fsType == "ext4" {
```
innobead, line 642: Use `DefaultFsType` instead of `ext4`

csi/node_server.go
```diff
@@ -615,25 +628,33 @@ func (ns *NodeServer) getMounter(volume *longhornclient.Volume, volumeCapability
...
 		}
 
 		mounter := &mount.SafeFormatAndMount{Interface: mount.New(""), Exec: utilexec.New()}
-		// we allow the user to provide additional params for ext4 filesystem creation.
-		// this allows an ext4 fs to be mounted on older kernels, see https://github.com/longhorn/longhorn/issues/1208
-		if fsType == "ext4" && userExt4Params != nil && userExt4Params.Value != "" {
-			ext4Params := userExt4Params.Value
-			logrus.Infof("volume %v using user provided ext4 fs creation params: %s", volume.Name, ext4Params)
-			cmdParamMapping := map[string]string{"mkfs." + fsType: ext4Params}
+		if fsType == "ext4" || fsType == "xfs" {
```
innobead, line 649: The same here to use `DefaultFsType` or another two constants for ext4 and xfs.

shuo-wu (approved):

csi/node_server.go
```diff
@@ -615,25 +628,33 @@ func (ns *NodeServer) getMounter(volume *longhornclient.Volume, volumeCapability
...
+			logrus.Infof("volume %v using user and longhorn provided %v fs creation params: %s", volume.Name, fsType, params)
+			cmdParamMapping := map[string]string{"mkfs." + fsType: params}
 			mounter = &mount.SafeFormatAndMount{
 				Interface: mount.New(""),
 				Exec:      NewForcedParamsExec(cmdParamMapping),
 			}
+		} else {
+			logrus.Infof("volume %v with unsupported filesystem %v, use default fs creation params", volume.Name, fsType)
```
innobead, line 657: Should we use `Errorf` instead? also return an error if the fs not supported?

innobead (commented):
LGTM, just minor comments.

csi/node_server.go
derekbit: Updated, thank you.

csi/node_server.go
derekbit: Updated, thank you.

csi/node_server.go
derekbit: Use `Warnf`, instead. Thank you.

innobead (approved):

innobead:
@derekbit Please help backport to 1.3 and 1.2. Thanks.

joshimoo (commented):
LGTM, good approach of reusing the forced params executor to extend the mkfs. call :)

PhanLe1010:
LGTM for this release

However, this only fix ext4 and xfs. In the future, maybe we can allow the user to specify the filesystem parameter in storageclass for any filesystem. Ref: https://github.com/longhorn/longhorn/issues/4401#issuecomment-1212565935

How do you think @innobead @derekbit @joshimoo ?

derekbit:
> LGTM for this release
> 
> However, this only fix ext4 and xfs. In the future, maybe we can allow the user to specify the filesystem parameter in storageclass for any filesystem. Ref: [longhorn/longhorn#4401 (comment)](https://github.com/longhorn/longhorn/issues/4401#issuecomment-1212565935)
> 
> How do you think @innobead @derekbit @joshimoo ?

Totally agreed! Actually, I was implementing this idea when working on this PR.
The idea is put the format (mkfs) options in `req.VolumeContext["formatOptions"]`.

In addition, we have ext4 mkfs parameters setting now. if users set the per-storage mkfs options, the options will override the settings from the global settings. How about?

joshimoo:
Good idea to add the parameters to the storage class, we can deprecate the global setting in the 1.3.x release then remove it for the 1.4 release since the parameter is only relevant for new volume creation, no point in carry it around for a long time. Let's make sure it's part of the 1.3.x and 1.4 readmes since it's an important change any way where we would want the user to specify the 4k block size.

Just for reference the storage class params can also be used to address this issue https://github.com/longhorn/longhorn/issues/4129#issuecomment-1158466458

innobead:
Sounds good to me, especially from 1.4, ideally, fs type should not be an issue anymore, so it's fair to make the fs format options general in the storage class instead.

PhanLe1010:
Ceph CSI has some example: https://github.com/ceph/ceph-csi/blob/0311eb5f447762a581194d649ea4d703da104fb8/charts/ceph-csi-cephfs/templates/storageclass.yaml#L55

PR #1496, opened 2022-09-21
crds: auto-generate preserveUnknownFields field

derekbit:
[Longhorn 4600](https://github.com/longhorn/longhorn/issues/4600)


innobead (approved):
LGTM.

Because we are planning to remove deploy folder which is duplicated with the ones in longhorn repo, but we need to take care crd generation here which will be saved in deploy/install/01-prerequisite/03-crd.yaml.

innobead:
@derekbit Please help backport to 1.3. Thanks.

PR #1497, opened 2022-09-21
Create iSCSI volumes with block size 4096 bytes

derekbit:
[Longhorn 4599](https://github.com/longhorn/longhorn/issues/4599)

TODO: update the go.mod go-iscsi-helpe, longhorn-engine and longhorn-instance-manager after review


util/util.go
```diff
@@ -895,3 +897,12 @@ func GetVolumeMeta(path string) (*VolumeMeta, error) {
 	}
 	return meta, nil
 }
+
+func GetVolumeBlockSize(blockSize int64) int64 {
+	// If blockSize is not defined, the existing volume uses 512 as its iSCSI LUN block size
+	if blockSize == 0 {
+		return 512
```
shuo-wu, line 904: NIT: Use const `LagacyBlockSize`

webhook/resources/engine/validator.go
```diff
@@ -0,0 +1,44 @@
...
+	}
+}
+
+func (e *engineValidator) Update(request *admission.Request, oldObj runtime.Object, newObj runtime.Object) error {
+	oldEngine := oldObj.(*longhorn.Engine)
+	newEngine := newObj.(*longhorn.Engine)
+
+	if oldEngine.Spec.BlockSize != newEngine.Spec.BlockSize {
+		return werror.NewInvalidError("Changing blockSize is not allowable", "")
+	}
```
shuo-wu, line 41: Can we add an engine mutator to update `spec.blockSize` for existing engines?
Then this validator that prevents the **non-zero** `e.spec.blockSize` from being modified only.

util/util.go
derekbit: Updated. Thank you.

webhook/resources/engine/validator.go
derekbit: Yes, updated. Thank you.

shuo-wu (approved):

PR #1499, opened 2022-09-26
Make cronjob jobs historic count be configurable

weizhe0422:
[Longhorn 1711](https://github.com/longhorn/longhorn/issues/1711)

## Items
1. Add two arguments for user setting cronjob successful and failed historic count: `recurring-successful-jobs-history-limit` & `recurring-failed-jobs-history-limit`
2. `recurring_job_controller` reference the argument to create cronjob.
3. Remove `CronJobSuccessfulJobsHistoryLimit` in controller/volume_controller.go

## Self-test results
- Step 1. Use the latest Longhorn [deployment file](https://raw.githubusercontent.com/longhorn/longhorn/master/deploy/longhorn.yaml), but adding `recurring-successful-jobs-history-limit` & `recurring-failed-jobs-history-limit` in `default-setting.yaml` block like below:
    image
- Step 2. Use the deployment fie to build up Longhorn
- Step 3. Create a volume and recurring job to create backup (or snapshots) for the volume
- Step 4. Type `kubectl describe cronjobs <recurring_job_name> -n longhorn-system`
    - Validate that the parameters should be the same as configured.
      image
- Step 5. After several round of recurring job execution, type `kubectl get jobs -n longhorn-system`
    - Validate the count of jobs should be same as configured.
      image


controller/recurring_job_controller.go
```diff
@@ -380,7 +381,16 @@ func (control *RecurringJobController) checkAndUpdateCronJob(cronJob, appliedCro
 
 func (control *RecurringJobController) newCronJob(recurringJob *longhorn.RecurringJob) (*batchv1beta1.CronJob, error) {
 	backoffLimit := int32(CronJobBackoffLimit)
-	successfulJobsHistoryLimit := int32(CronJobSuccessfulJobsHistoryLimit)
+	settingSuccessfulJobsHistoryLimit, err := control.ds.GetSettingAsInt(types.SettingNameRecurringSuccessfulJobsHistoryLimit)
+	if err != nil {
+		return nil, werror.NewInvalidError(fmt.Sprintf("failed to get setting of RecurringSuccessfulJobsHistoryLimit: %v", err), "")
```
shuo-wu, line 386: It's better not to use `webhook/error` in controllers. At least this is not an HTTP error...

shuo-wu (changes requested):
Others LGTM

types/setting.go
```diff
@@ -966,6 +992,15 @@ func ValidateSetting(name, value string) (err error) {
 			return fmt.Errorf("the value of %v is invalid: %v", sName, err)
 		}
 
+	case SettingNameRecurringSuccessfulJobsHistoryLimit:
+		if _, err := strconv.Atoi(value); err != nil {
+			return fmt.Errorf("value %v is not a number", value)
```
derekbit, line 997: Why not use `errors.Wrapf`?

types/setting.go
```diff
@@ -966,6 +992,15 @@ func ValidateSetting(name, value string) (err error) {
...
+		}
+	case SettingNameRecurringFailedJobsHistoryLimit:
+		if _, err := strconv.Atoi(value); err != nil {
+			return fmt.Errorf("value %v is not a number", value)
```
derekbit, line 1001: [1]

types/setting.go
innobead: Ray, need to align with others at least to have err in the message. 

I think we should refactor returned errors for each setting in this method later but use wrap instead to embed the original error if needed.

controller/recurring_job_controller.go
innobead: Ray, you can learn from other methods instead to use the consistent way like `errors.Wrapf`

weizhe0422: Okay, I use the error library to wrap the error. And I have fixed the bug.
https://github.com/longhorn/longhorn-manager/blob/a9c3853113e0d70b114260aeeaba020f3f2d9240/controller/recurring_job_controller.go#L383-L392

types/setting.go
weizhe0422: I have done to wrap the error to return.
https://github.com/longhorn/longhorn-manager/blob/a9c3853113e0d70b114260aeeaba020f3f2d9240/types/setting.go#L995-L1002

And I also modified others' return errors as wrapping errors message.

types/setting.go
```diff
@@ -875,10 +875,10 @@ func ValidateSetting(name, value string) (err error) {
 
 	definition, ok := GetSettingDefinition(sName)
 	if !ok {
-		return fmt.Errorf("setting %v is not supported", sName)
+		return errors.Wrapf(err, "setting %v is not supported", sName)
```
shuo-wu, line 878: if `err` is nil, `errors.Wrapf()` would return nil as well. In other words, do not use this function when there is no existing error.
Similar below.

shuo-wu (changes requested):

types/setting.go
weizhe0422: It is my mistake to use wrap error function even there is no error returned. I will fix it asap. Thank you.

shuo-wu (approved):

types/setting.go
```diff
@@ -815,6 +821,26 @@ var (
 		ReadOnly: false,
 		Default:  CniNetworkNone,
 	}
+
+	SettingDefinitionRecurringSuccessfulJobsHistoryLimit = SettingDefinition{
+		DisplayName: "Cronjob Successful Jobs History Limit",
+		Description: "This setting specifies how much successful backup or snapshot job history should be kept.",
```
innobead, line 827:
```suggestion
		Description: "This setting specifies how many successful backup or snapshot job histories should be retained.",
```

types/setting.go
```diff
@@ -815,6 +821,26 @@ var (
...
+		Category:    SettingCategoryBackup,
+		Type:        SettingTypeInt,
+		Required:    false,
+		ReadOnly:    false,
+		Default:     "1",
+	}
+
+	SettingDefinitionRecurringFailedJobsHistoryLimit = SettingDefinition{
+		DisplayName: "Cronjob Failed Jobs History Limit",
+		Description: "\nThis setting specifies how much failed backup or snapshot job history should be kept.",
```
innobead, line 837: Why there was a `/n` at the beginning?
```suggestion
		Description: "This setting specifies how many failed backup or snapshot job histories should be retained.",
```

types/setting.go
```diff
@@ -895,33 +921,39 @@ func ValidateSetting(name, value string) (err error) {
...
+			return errors.Wrapf(err, "failed to parse %v as size", value)
+		}
+		if value < 0 || value > 100 {
 			return fmt.Errorf("value %v should between 0 to 100", value)
 		}
 	case SettingNameDefaultReplicaCount:
 		c, err := strconv.Atoi(value)
 		if err != nil {
-			return fmt.Errorf("value %v is not int: %v", SettingNameDefaultReplicaCount, err)
+			return errors.Wrapf(err, "value %v is not number", SettingNameDefaultReplicaCount)
```
innobead, line 949: The previous was not right. 
```suggestion
			return errors.Wrapf(err, "value %v is not number", value)
```

types/setting.go
```diff
@@ -895,33 +921,39 @@ func ValidateSetting(name, value string) (err error) {
...
 		}
 		if err := ValidateReplicaCount(c); err != nil {
-			return fmt.Errorf("value %v: %v", c, err)
+			return errors.Wrapf(err, "failed to validate replica count %v", c)
 		}
 	case SettingNameReplicaAutoBalance:
 		if err := ValidateReplicaAutoBalance(longhorn.ReplicaAutoBalance(value)); err != nil {
-			return fmt.Errorf("value %v: %v", value, err)
+			return errors.Wrapf(err, "value %v: %v", value, err)
```
innobead, line 956:
```suggestion
			return errors.Wrapf(err, "failed to validate replica auto balance: %v", value)
```

innobead (changes requested):
In general, LGTM, just a few things needed to change. Please ensure the changed/refactored settings to verify as well.

Remember to update below items as well.
- deploy/install/01-prerequisite/04-default-setting.yaml
- longhorn/longhorn manifests (chart and deploy which can be generated by generate-longhorn-yaml.sh)

types/setting.go
weizhe0422: Thank you, I will follow the comment to fix.
https://github.com/longhorn/longhorn-manager/blob/de00e06ea1fcd510afee72cb8132231513953f84/types/setting.go#L827

types/setting.go
weizhe0422: Thank you, I will follow the comment to fix.
https://github.com/longhorn/longhorn-manager/blob/de00e06ea1fcd510afee72cb8132231513953f84/types/setting.go#L837

types/setting.go
weizhe0422: Thank you, I will follow the comment to fix.
https://github.com/longhorn/longhorn-manager/blob/de00e06ea1fcd510afee72cb8132231513953f84/types/setting.go#L949

types/setting.go
weizhe0422: Thank you, I will follow the comment to fix.
https://github.com/longhorn/longhorn-manager/blob/de00e06ea1fcd510afee72cb8132231513953f84/types/setting.go#L956

types/setting.go
```diff
@@ -940,30 +972,39 @@ func ValidateSetting(name, value string) (err error) {
...
+			return errors.Wrapf(err, "the value of %v is invalid", sName)
+		}
+
+	case SettingNameRecurringSuccessfulJobsHistoryLimit:
+		if _, err := strconv.Atoi(value); err != nil {
+			return errors.Wrapf(err, "value %v is not a number", value)
+		}
+	case SettingNameRecurringFailedJobsHistoryLimit:
+		if _, err := strconv.Atoi(value); err != nil {
+			return errors.Wrapf(err, "value %v is not a number", value)
```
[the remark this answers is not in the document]
weizhe0422: Thanks, I've added check logic to return an error message if the value is negative.

https://github.com/longhorn/longhorn-manager/blob/73ac8e5c29b351fea40547e84669d392007b81fb/types/setting.go#L1006-L1008
https://github.com/longhorn/longhorn-manager/blob/73ac8e5c29b351fea40547e84669d392007b81fb/types/setting.go#L1014-L1016

types/setting.go
```diff
@@ -940,30 +972,47 @@ func ValidateSetting(name, value string) (err error) {
...
+		if successJobsHistLimit < 0 {
+			return fmt.Errorf("the value %v shouldn't be less than 0", value)
+		}
+	case SettingNameRecurringFailedJobsHistoryLimit:
+		failedJobsHistLimit, err := strconv.Atoi(value)
+		if err != nil {
+			return errors.Wrapf(err, "value %v is not a number", value)
+		}
+		if failedJobsHistLimit < 0 {
+			return fmt.Errorf("the value %v shouldn't be less than 0", value)
```
derekbit, line 1015: What does 0 mean in the two settings? If it's meaningful, you can mention in the setting description.

weizhe0422: Set the limit to 0 means that it won't keep history after the job finished. And I will mention it in the description.

https://github.com/longhorn/longhorn-manager/blob/35d60930fc09d0b4d37218488aa11dcdb0e91942/types/setting.go#L827-L828
https://github.com/longhorn/longhorn-manager/blob/35d60930fc09d0b4d37218488aa11dcdb0e91942/types/setting.go#L838-L839

types/setting.go
```diff
@@ -815,6 +821,28 @@ var (
 		ReadOnly: false,
 		Default:  CniNetworkNone,
 	}
+
+	SettingDefinitionRecurringSuccessfulJobsHistoryLimit = SettingDefinition{
+		DisplayName: "Cronjob Successful Jobs History Limit",
+		Description: "This setting specifies how many successful backup or snapshot job histories should be retained. \n\n" +
+			"Set the value to 0 means that it won't keep history after the job finished.",
```
derekbit, line 828:
```suggestion
			"History will not be retained if the value is 0.",
```

same in the below

weizhe0422: Thank you, I will follow the comment to fix.

https://github.com/longhorn/longhorn-manager/blob/8511cfdf0080dbf35c1dae4e00638ed9b345984b/types/setting.go#L827-L828

https://github.com/longhorn/longhorn-manager/blob/8511cfdf0080dbf35c1dae4e00638ed9b345984b/types/setting.go#L838-L839

innobead (approved):

PR #1508, opened 2022-09-28
Unused replica failed would not be scheduled.

mantissahz:
Not to schedule unused replicas that marked as failed when the node with engine and instance-manager rebooted.

longhorn/longhorn#4546

### DIFF
--- a/scheduler/replica_scheduler.go
+++ b/scheduler/replica_scheduler.go
@@ -52,6 +52,12 @@ func (rcs *ReplicaScheduler) ScheduleReplica(replica *longhorn.Replica, replicas
 		return nil, nil, fmt.Errorf("BUG: Replica %v has been scheduled to node %v", replica.Name, replica.Spec.NodeID)
 	}
 
+	// not to schedule a replica failed and unused before.
+	if replica.Spec.HealthyAt == "" && replica.Spec.FailedAt != "" {
+		logrus.WithField("replica", replica.Name).Warn("Failed replica is not scheduled")
+		return nil, nil, nil
+	}
+
 	// get all hosts
 	nodesInfo, err := rcs.getNodeInfo()
 	if err != nil {


scheduler/replica_scheduler.go
```diff
@@ -45,13 +45,27 @@ func NewReplicaScheduler(ds *datastore.DataStore) *ReplicaScheduler {
 	return rcScheduler
 }
 
+func isFirstAttachment(rs map[string]*longhorn.Replica) bool {
```
shuo-wu, line 48: Why not directly check the Node ID or DIsk ID...?

scheduler/replica_scheduler.go
```diff
@@ -45,13 +45,27 @@ func NewReplicaScheduler(ds *datastore.DataStore) *ReplicaScheduler {
...
 func (rcs *ReplicaScheduler) ScheduleReplica(replica *longhorn.Replica, replicas map[string]*longhorn.Replica, volume *longhorn.Volume) (*longhorn.Replica, util.MultiError, error) {
 	// only called when replica is starting for the first time
 	if replica.Spec.NodeID != "" {
 		return nil, nil, fmt.Errorf("BUG: Replica %v has been scheduled to node %v", replica.Name, replica.Spec.NodeID)
 	}
 
+	if !isFirstAttachment(replicas) && replica.Spec.HealthyAt == "" && replica.Spec.FailedAt != "" {
+		logrus.WithField("replica", replica.ObjectMeta.Name).Warnf("Replica might not be scheduled.")
+		return nil, nil, nil
+	}
```
shuo-wu, line 67: It seems that somehow Longhorn wrongly re-scheduled a failed replica and ignores the affinity setting. 
But why doesn't the check `replica.Spec.NodeID != ""` prevent that from happening?

scheduler/replica_scheduler.go
mantissahz: Do you mean to check `replica.Spec.NodeID`?

scheduler/replica_scheduler.go
mantissahz: `replica.Spec.NodeID != ""` does not work because this replica was not scheduled before such as volume degraded. In the issue longhorn/longhorn#4546, there is only one worker node and  only one replica is running and the others are still in scheduling (`replica.Spec.NodeID` is empty but no available node.) Then all replica will be marked as failed after the worker node rebooted and all replica would be scheduled to this worker node because replica marked as failed does not occupy the node.

scheduler/replica_scheduler.go
shuo-wu: Yes

scheduler/replica_scheduler.go
shuo-wu: OK, got it. The key point of this PR is to delay the failed but unscheduled replicas before handling the reusable or salvageable replicas, right?

mantissahz: Yes.

scheduler/replica_scheduler.go
mantissahz: Yeah, it could check NodeId. Thanks.
Modified.

scheduler/replica_scheduler.go
shuo-wu: Then you can have a separate function with an easy-to-understand naming for this check.

mantissahz: OK, modified. 
Thanks.

scheduler/replica_scheduler.go
```diff
@@ -45,13 +45,36 @@ func NewReplicaScheduler(ds *datastore.DataStore) *ReplicaScheduler {
 	return rcScheduler
 }
 
+// checkFailedReplicaUnused will return true if replica was not used before and marked as failed
+func checkFailedReplicaUnused(replica *longhorn.Replica, replicas map[string]*longhorn.Replica) bool {
```
c3y1huang, line 49:
```suggestion
func isFailedReplicaUnused(replica *longhorn.Replica, replicas map[string]*longhorn.Replica) bool {
```

scheduler/replica_scheduler.go
```diff
@@ -45,13 +45,36 @@ func NewReplicaScheduler(ds *datastore.DataStore) *ReplicaScheduler {
...
+	isFirstAttachment := func(rs map[string]*longhorn.Replica) bool {
+		for _, r := range rs {
+			if r.Spec.NodeID != "" {
+				return false
+			}
+		}
+		return true
+	}
```
c3y1huang, line 57:
```suggestion
	isFirstAttachment := true
	for _, r := range replicas {
		if r.Spec.NodeID == "" {
			isFirstAttachment = false
			break
		}
	}
```

scheduler/replica_scheduler.go
```diff
@@ -45,13 +45,36 @@ func NewReplicaScheduler(ds *datastore.DataStore) *ReplicaScheduler {
...
 // ScheduleReplica will return (nil, nil) for unschedulable replica
 func (rcs *ReplicaScheduler) ScheduleReplica(replica *longhorn.Replica, replicas map[string]*longhorn.Replica, volume *longhorn.Volume) (*longhorn.Replica, util.MultiError, error) {
 	// only called when replica is starting for the first time
 	if replica.Spec.NodeID != "" {
 		return nil, nil, fmt.Errorf("BUG: Replica %v has been scheduled to node %v", replica.Name, replica.Spec.NodeID)
 	}
 
+	// not to schedule a replica failed and unused before.
+	if checkFailedReplicaUnused(replica, replicas) {
+		logrus.WithField("replica", replica.ObjectMeta.Name).Warnf("Replica might not be scheduled.")
```
c3y1huang, line 74:
```suggestion
		logrus.WithField("replica", replica.Name).Warn("Failed replica is not scheduled")
```

scheduler/replica_scheduler.go
mantissahz: Modified. Thanks.

scheduler/replica_scheduler.go
```diff
@@ -45,13 +45,31 @@ func NewReplicaScheduler(ds *datastore.DataStore) *ReplicaScheduler {
...
+// isFailedReplicaUnused will return true if replica was not used before and marked as failed
+func isFailedReplicaUnused(replica *longhorn.Replica, replicas map[string]*longhorn.Replica) bool {
+	isFirstAttachment := true
+	for _, r := range replicas {
+		if r.Spec.NodeID != "" {
+			isFirstAttachment = false
+			break
+		}
+	}
+	return !isFirstAttachment && replica.Spec.HealthyAt == "" && replica.Spec.FailedAt != ""
```
c3y1huang, line 57: How about check `replica.Spec.HealthyAt` and `replica.Spec.FailedAt` first? So we only go through the replicas when necessary, for ex:

```
if replica.Spec.HealthyAt != "" {
  return false
}

if replica.Spec.FailedAt == "" {
  return false
}

for _, r := range replicas {
  ...
}
```

scheduler/replica_scheduler.go
```diff
@@ -45,13 +45,31 @@ func NewReplicaScheduler(ds *datastore.DataStore) *ReplicaScheduler {
 	return rcScheduler
 }
 
+// isFailedReplicaUnused will return true if replica was not used before and marked as failed
+func isFailedReplicaUnused(replica *longhorn.Replica, replicas map[string]*longhorn.Replica) bool {
+	isFirstAttachment := true
```
derekbit, line 50: nit: the naming `isFirstAttachment` is confusing... The purpose of `isFirstAttachment` is to check if there is any scheduled replica.

mantissahz: I will remove this at next commit.

derekbit: BTW, why need to check if there is any scheduled replica here?

mantissahz: Yeah, it might not need to check this now. I will test this.
At first it just give a chance to schedule the replica if all replicas `replica.Spec.HealthyAt == ""` but now use NodeId.

scheduler/replica_scheduler.go
mantissahz: Make it more simple, just to check the replica `HealtyhAt` and `FailedAt`.

scheduler/replica_scheduler.go
mantissahz: Make it more simple, just to check the replica `HealtyhAt` and `FailedAt`.

scheduler/replica_scheduler.go
mantissahz: Make it more simple, just to check the replica `HealtyhAt` and `FailedAt`.

scheduler/replica_scheduler.go
```diff
@@ -52,6 +52,12 @@ func (rcs *ReplicaScheduler) ScheduleReplica(replica *longhorn.Replica, replicas
 		return nil, nil, fmt.Errorf("BUG: Replica %v has been scheduled to node %v", replica.Name, replica.Spec.NodeID)
 	}
 
+	// not to schedule a replica failed and unused before.
```
shuo-wu, line 55: Then will this kind of replica block the replica replenishment? or can these replicas be reused later?

mantissahz: No, it will not block the replica replenishment.

shuo-wu: Can you explain more about it? I don't remember to the details...

mantissahz: `replenishReplicas` will get Replenish Replicas Count [here](https://github.com/longhorn/longhorn-manager/blob/master/controller/volume_controller.go#L2112-L2114) (mainly according to `r.Spec.FailedAt == ""`)
and then use '[replica-scheduler.CheckAndReuseFailedReplica](https://github.com/longhorn/longhorn-manager/blob/master/scheduler/replica_scheduler.go#L434)' to get reusable replicas (it should be empty in this case.), finally check if it requires to create new replica.
I think replica replenishment would not get into the codes I add and that's what I saw when I tested the codes.

shuo-wu: Got it. Since it's not a new volume, the replenishment count should be 1. And the potentially reusable replica is the only replica that contains data. After the salvage or reuse, the volume controller will ignore those failed & unscheduled replicas then directly replenish new replicas, right?

mantissahz: Yes, that's right, only replenishment count will be `v.Spec.NumberOfReplicas - usableCount`.

shuo-wu (approved):

innobead (approved):

PR #1486, opened 2022-08-29
Save the access mode when creating backup, for restoring from the backup

weizhe0422:
[Issue #3444](https://github.com/longhorn/longhorn/issues/3444)
Signed-off-by: Ray Chang <ray.chang@suse.com>
## Enhancement
  1. When create backup, save the access mode into backup label
  2. When restore backup, if the access mode is empty then use the backup label to create new volume.

## Self-testing
  1. Create two volumes: ReadWriteOnce (RWO) * 1 + ReadWriteMany (RWX) * 1, and both are attached to one node.
      image * 1 + ReadWriteMany (RWX) * 1, and both are attached to one node.
      image
  3. Create backups for each volumes.
      image
  4. Wait all backups are complete, then delete all volumes.
  5. Switch to `Backup' page, and click the backup name and the tags filed. 
  6. **CHECK: The `LonghornVolumeAccessMode` of backup labels should be equal as the access mode.**
      image
  7. Select all backups and click `Restore Latest Backup` button
      Screenshot from 2022-08-29 20-53-17
  8. DON'T fill in any filed and click `OK` button.
  9. Switch back to `Volume` page and click the name filed.
  10. CHECK: The access mode should be same as the volume's backup.
       - ReadWriteOnece
         image
       - ReadWriteMany
          image://user-images.githubusercontent.com/17548901/187202524-568df547-f571-46d7-be32-3a58ed8dc32f.png)
  3. Create backups for each volumes.
      image
  4. Wait all backups are complete, then delete all volumes.
  5. Switch to `Backup' page, and click the backup name and the tags filed. 
  6. **CHECK: The `LonghornVolumeAccessMode` of backup labels should be equal as the access mode.**
      image
  8. Select all backups and click `Restore Latest Backup` button
      Screenshot from 2022-08-29 20-53-17
  9. DON'T fill in any filed and click `OK` button.
  10. Switch back to `Volume` page and click the name filed.
  11. CHECK: The access mode should be same as the volume's backup.
       - ReadWriteOnece
         image      
       - ReadWriteMany
          image

PR #1509, opened 2022-09-29
fix: Can not pull a backup created by another Longhorn system from the remote backup target

mantissahz:
Getting the access mode of a volume that does not exists would block creating a backup created by another Longhorn system.

longhorn/longhorn#4637

### DIFF
--- a/controller/backup_volume_controller.go
+++ b/controller/backup_volume_controller.go
@@ -314,9 +314,26 @@ func (bvc *BackupVolumeController) reconcile(backupVolumeName string) (err error
 		log.Infof("Found %d backups in the backup target that do not exist in the cluster and need to be pulled", count)
 	}
 	for backupName := range backupsToPull {
+		backupLabelMap := map[string]string{}
+
+		backupURL := backupstore.EncodeBackupURL(backupName, backupVolumeName, backupTargetClient.URL)
+		if backupInfo, err := backupTargetClient.BackupGet(backupURL, backupTargetClient.Credential); err != nil {
+			log.WithError(err).WithFields(logrus.Fields{
+				"backup":       backupName,
+				"backupvolume": backupVolumeName,
+				"backuptarget": backupURL}).Warnf("Failed to get backupInfo from remote backup target")
+		} else {
+			if accessMode, exist := backupInfo.Labels[types.GetLonghornLabelKey(types.LonghornLabelVolumeAccessMode)]; exist {
+				backupLabelMap[types.GetLonghornLabelKey(types.LonghornLabelVolumeAccessMode)] = accessMode
+			}
+		}
+
 		backup := &longhorn.Backup{
 			ObjectMeta: metav1.ObjectMeta{
 				Name: backupName,
+			},
+			Spec: longhorn.BackupSpec{
+				Labels: backupLabelMap,
 			},
 		}
 		if _, err = bvc.ds.CreateBackup(backup, backupVolumeName); err != nil && !apierrors.IsAlreadyExists(err) {
--- a/webhook/resources/backup/mutator.go
+++ b/webhook/resources/backup/mutator.go
@@ -64,16 +64,16 @@ func (b *backupMutator) Create(request *admission.Request, newObj runtime.Object
 		err := errors.Wrapf(err, "cannot find the backup volume label for backup %v", backup.Name)
 		return nil, werror.NewInvalidError(err.Error(), "")
 	}
-	volume, err := b.ds.GetVolumeRO(volumeName)
-	if err != nil {
-		err := errors.Wrapf(err, "failed to get volume %v", volumeName)
-		return nil, werror.NewInvalidError(err.Error(), "")
+
+	if _, isExist := backupLabels[types.GetLonghornLabelKey(types.LonghornLabelVolumeAccessMode)]; !isExist {
+		volumeAccessMode := longhorn.AccessModeReadWriteOnce
+		if volume, err := b.ds.GetVolumeRO(volumeName); err == nil {
+			if volume.Spec.AccessMode != "" {
+				volumeAccessMode = volume.Spec.AccessMode
+			}
+		}
+		backupLabels[types.GetLonghornLabelKey(types.LonghornLabelVolumeAccessMode)] = string(volumeAccessMode)
 	}
-	volumeAccessMode := volume.Spec.AccessMode
-	if volume.Spec.AccessMode == "" {
-		volumeAccessMode = longhorn.AccessModeReadWriteOnce
-	}
-	backupLabels[types.GetLonghornLabelKey(types.LonghornLabelVolumeAccessMode)] = string(volumeAccessMode)
 
 	valueBackupLabels, err := json.Marshal(backupLabels)
 	if err != nil {


innobead:
cc @weizhe0422 this seems a regression from the previous fix for access mode ..

webhook/resources/backup/mutator.go
```diff
@@ -64,14 +64,11 @@ func (b *backupMutator) Create(request *admission.Request, newObj runtime.Object
...
 	}
-	volume, err := b.ds.GetVolumeRO(volumeName)
-	if err != nil {
-		err := errors.Wrapf(err, "failed to get volume %v", volumeName)
-		return nil, werror.NewInvalidError(err.Error(), "")
-	}
-	volumeAccessMode := volume.Spec.AccessMode
-	if volume.Spec.AccessMode == "" {
-		volumeAccessMode = longhorn.AccessModeReadWriteOnce
+	volumeAccessMode := longhorn.AccessModeReadWriteOnce
```
derekbit, line 67: The volumeAccessMode is stored in remote backup's cfg file after this [PR](https://github.com/longhorn/longhorn-manager/pull/1486).

In the backup sync case, the correct volumeAccessMode can be read from the remote cfg file. If volumeAccessMode label does not exist, we can set to `AccessModeReadWriteOnce` by default.

mantissahz: What about `access mode` is different between backup cfg (RWX when backup) and volume spec (RWO now), I think it should still take the backup cfg, WDYT?

innobead: YES, the backup will be the source of truth if it has access mode already.

mantissahz: Addressed, update backup_controller.go as well.

webhook/resources/backup/mutator.go
```diff
@@ -64,16 +64,16 @@ func (b *backupMutator) Create(request *admission.Request, newObj runtime.Object
...
+		if volume.Spec.AccessMode != "" {
+			volumeAccessMode = volume.Spec.AccessMode
+		}
 	}
-	volumeAccessMode := volume.Spec.AccessMode
-	if volume.Spec.AccessMode == "" {
-		volumeAccessMode = longhorn.AccessModeReadWriteOnce
+	if _, isExist := backupLabels[types.GetLonghornLabelKey(types.LonghornLabelVolumeAccessMode)]; !isExist {
+		backupLabels[types.GetLonghornLabelKey(types.LonghornLabelVolumeAccessMode)] = string(volumeAccessMode)
 	}
```
innobead, line 76: Or you can do this way to avoid unnecessary volume query?

```go
    if _, isExist := backupLabels[types.GetLonghornLabelKey(types.LonghornLabelVolumeAccessMode)]; !isExist {
        volumeAccessMode := longhorn.AccessModeReadWriteOnce
        
        if volume, err := b.ds.GetVolumeRO(volumeName); err == nil {
            if volume.Spec.AccessMode != "" {
                volumeAccessMode = volume.Spec.AccessMode
            }
        }

        backupLabels[types.GetLonghornLabelKey(types.LonghornLabelVolumeAccessMode)] = string(volumeAccessMode)
    }
```

controller/backup_volume_controller.go
```diff
@@ -314,10 +314,20 @@ func (bvc *BackupVolumeController) reconcile(backupVolumeName string) (err error
 		log.Infof("Found %d backups in the backup target that do not exist in the cluster and need to be pulled", count)
 	}
 	for backupName := range backupsToPull {
+		backupLabelMap := map[string]string{types.GetLonghornLabelKey(types.LonghornLabelVolumeAccessMode): string(longhorn.AccessModeReadWriteOnce)}
+
+		backupURL := backupstore.EncodeBackupURL(backupName, backupVolumeName, backupTargetClient.URL)
+		if backupInfo, err := backupTargetClient.BackupGet(backupURL, backupTargetClient.Credential); err == nil {
```
innobead, line 320: Don't swallow error. If want to continue when encountering error, you can add a warning log to what's happening behind.

controller/backup_volume_controller.go
```diff
@@ -314,10 +314,20 @@ func (bvc *BackupVolumeController) reconcile(backupVolumeName string) (err error
...
+			backupLabelMap[types.GetLonghornLabelKey(types.LonghornLabelVolumeAccessMode)] = backupInfo.Labels[types.GetLonghornLabelKey(types.LonghornLabelVolumeAccessMode)]
```
innobead, line 321: Ideally, it should be null check, but in go, it will return empty string if the key does not exist, so it's fine.

innobead (changes requested):
LGTM, but just a few comments to resolve.

Well done to catch missing labels when creating backups from the remote backup target!

webhook/resources/backup/mutator.go
mantissahz: Modified. Thanks.

controller/backup_volume_controller.go
mantissahz: Yeah, modified, it would have a warning log.

controller/backup_volume_controller.go
mantissahz: Modified, it would check key exists first.

controller/backup_volume_controller.go
```diff
@@ -314,10 +314,27 @@ func (bvc *BackupVolumeController) reconcile(backupVolumeName string) (err error
...
 	}
 	for backupName := range backupsToPull {
+		backupLabelMap := map[string]string{types.GetLonghornLabelKey(types.LonghornLabelVolumeAccessMode): string(longhorn.AccessModeReadWriteOnce)}
+
+		backupURL := backupstore.EncodeBackupURL(backupName, backupVolumeName, backupTargetClient.URL)
+		if backupInfo, err := backupTargetClient.BackupGet(backupURL, backupTargetClient.Credential); err != nil {
+			log.WithError(err).WithFields(logrus.Fields{
+				"backup":       backupName,
+				"backupvolume": backupVolumeName,
+				"backuptarget": backupURL}).Warnf("Get backupInfo from remote backup target")
```
innobead, line 324:
```suggestion
				"backuptarget": backupURL}).Warnf("Failed to get backupInfo from remote backup target")
```

controller/backup_volume_controller.go
```diff
@@ -314,10 +314,27 @@ func (bvc *BackupVolumeController) reconcile(backupVolumeName string) (err error
 		log.Infof("Found %d backups in the backup target that do not exist in the cluster and need to be pulled", count)
 	}
 	for backupName := range backupsToPull {
+		backupLabelMap := map[string]string{types.GetLonghornLabelKey(types.LonghornLabelVolumeAccessMode): string(longhorn.AccessModeReadWriteOnce)}
```
innobead, line 317: This should be an empty map, because it will be mutated in the mutator.

innobead (changes requested):

controller/backup_volume_controller.go

controller/backup_volume_controller.go
mantissahz: Modified to empty map. Thanks.

innobead (approved):

PR #1513, opened 2022-10-01
Fix the share-manager deletion failure if the confimap is not existing

derekbit:
https://github.com/longhorn/longhorn/issues/4648

### DIFF
--- a/controller/share_manager_controller.go
+++ b/controller/share_manager_controller.go
@@ -301,8 +301,7 @@ func (c *ShareManagerController) syncShareManager(key string) (err error) {
 		}
 
 		if err := c.ds.DeleteConfigMap(c.namespace, types.GetConfigMapNameFromShareManagerName(sm.Name)); err != nil {
-			return errors.Wrapf(err, "failed to delete the configmap (recovery backend) for share manager %v",
-				sm.Name)
+			return errors.Wrapf(err, "failed to delete the configmap (recovery backend) for share manager %v", sm.Name)
 		}
 
 		return c.ds.RemoveFinalizerForShareManager(sm)
--- a/recovery_backend/backend/configmap.go
+++ b/recovery_backend/backend/configmap.go
@@ -23,21 +23,35 @@ type RecoveryBackend struct {
 	Datastore *datastore.DataStore
 }
 
-func newConfigMap(namespace, name, version string) *v1.ConfigMap {
+func (rb *RecoveryBackend) newConfigMap(namespace, name, version, smPodName string) (*v1.ConfigMap, error) {
+	smPod, err := rb.Datastore.GetPod(smPodName)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to get share manager pod %v", name)
+	}
+
+	smName, ok := smPod.Labels[types.GetLonghornLabelKey(types.LonghornLabelShareManager)]
+	if !ok {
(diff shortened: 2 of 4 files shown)


controller/share_manager_controller.go
```diff
@@ -300,9 +300,8 @@ func (c *ShareManagerController) syncShareManager(key string) (err error) {
 			return err
 		}
 
-		if err := c.ds.DeleteConfigMap(c.namespace, types.GetConfigMapNameFromShareManagerName(sm.Name)); err != nil {
-			return errors.Wrapf(err, "failed to delete the configmap (recovery backend) for share manager %v",
-				sm.Name)
+		if err := c.ds.DeleteConfigMap(c.namespace, types.GetConfigMapNameFromShareManagerName(sm.Name)); err != nil && !apierrors.IsNotFound(err) {
```
innobead, line 303: This kind of error seems appearing few times for different resource types previously. 

I would suggest to handle in datastore pkg directly, because this error doesn't mean a real error for us, or even I can say it's expected and intended. WDYT?

like: [/datastore/kubernetes.go#L138-L140](https://github.com/longhorn/longhorn-manager/blob/aa79220132ece38dbe55178104497748204e7d15/datastore/kubernetes.go#L138-L140)

innobead (changes requested):

controller/share_manager_controller.go
derekbit: Updated. Thank you.

recovery_backend/backend/configmap.go
```diff
@@ -12,9 +12,14 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/longhorn/longhorn-manager/datastore"
+	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
 	"github.com/longhorn/longhorn-manager/types"
 )
 
+var (
+	longhornFinalizerKey = longhorn.SchemeGroupVersion.Group
+)
```
innobead, line 21: WIll this be used? didn't see any reference in configmap.go

innobead (approved):

recovery_backend/backend/configmap.go
derekbit: Removed. Thank you.

innobead (approved):

recovery_backend/backend/configmap.go
```diff
@@ -23,21 +23,35 @@ type RecoveryBackend struct {
...
-func newConfigMap(namespace, name, version string) *v1.ConfigMap {
+func (rb *RecoveryBackend) newConfigMap(namespace, name, version, smPodName string) (*v1.ConfigMap, error) {
+	smPod, err := rb.Datastore.GetPod(smPodName)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to get share manager pod %v", name)
+	}
+
+	smName, ok := smPod.Labels[types.GetLonghornLabelKey(types.LonghornLabelShareManager)]
+	if !ok {
+		return nil, errors.Wrapf(err, "failed to get label %v from share manager pod %v",
```
shuo-wu, line 34: NIT: `smPod.Name` rather than `smName`. `smName` is empty here.

BTW, is the smPod name always the same as ShareManager CR name? If Yes, I think we can directly use the pod name and verify the ShareManager CR existence.

PR #1514, opened 2022-10-01
Update volume-scheduling-error failure

derekbit:
https://github.com/longhorn/longhorn/issues/4654

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -1194,7 +1194,7 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
 		}
 	}
 
-	if err := vc.updatePVAnnotation(v, types.PVAnnotationLonghornVolumeSchedulingError, failureMessage); err != nil {
+	if err := vc.ds.UpdatePVAnnotation(v, types.PVAnnotationLonghornVolumeSchedulingError, failureMessage); err != nil {
 		log.Warnf("Cannot update PV annotation for volume %v", v.Name)
 	}
 
@@ -2678,8 +2678,14 @@ func (vc *VolumeController) reconcileVolumeSize(v *longhorn.Volume, e *longhorn.
 		vc.eventRecorder.Eventf(v, v1.EventTypeNormal, EventReasonCanceledExpansion,
 			"Canceled expanding the volume %v, will automatically detach it", v.Name)
 	} else {
-		if err := vc.scheduler.CheckReplicasSizeExpansion(v, e.Spec.VolumeSize, v.Spec.Size); err != nil {
+		if diskScheduleMultiError, err := vc.scheduler.CheckReplicasSizeExpansion(v, e.Spec.VolumeSize, v.Spec.Size); err != nil {
 			log.Debugf("cannot start volume expansion: %v", err)
+			if diskScheduleMultiError != nil {
+				failureMessage := diskScheduleMultiError.Join()
+				if err := vc.ds.UpdatePVAnnotation(v, types.PVAnnotationLonghornVolumeSchedulingError, failureMessage); err != nil {
+					log.Warnf("Cannot update PV annotation for volume %v", v.Name)
+				}
+			}
 			return nil
 		}
 		log.Infof("Start volume expand from size %v to size %v", e.Spec.VolumeSize, v.Spec.Size)
@@ -3811,29 +3817,4 @@ func (vc *VolumeController) checkVolumeNotInMigration(volume *longhorn.Volume) e
 		return fmt.Errorf("cannot operate during migration")
 	}
 	return nil
-}
-
-func (vc *VolumeController) updatePVAnnotation(volume *longhorn.Volume, annotationKey, annotationVal string) error {
-	pv, err := vc.ds.GetPersistentVolume(volume.Status.KubernetesStatus.PVName)
-	if err != nil {
-		if apierrors.IsNotFound(err) {
-			return nil
-		}
-		return err
-	}
-
-	if pv.Annotations == nil {
-		pv.Annotations = map[string]string{}
-	}
-
-	val, ok := pv.Annotations[annotationKey]
-	if ok && val == annotationVal {
-		return nil
-	}
-
-	pv.Annotations[annotationKey] = annotationVal
-
-	_, err = vc.ds.UpdatePersistentVolume(pv)
-
-	return err
 }
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -1179,18 +1179,18 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
 				scheduled = true
 			}
 		}
+	}
 
-		if !scheduled {
-			if len(aggregatedReplicaScheduledError) == 0 {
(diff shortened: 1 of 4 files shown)


manager/volume.go
```diff
@@ -562,7 +561,14 @@ func (m *VolumeManager) Expand(volumeName string, size int64) (v *longhorn.Volum
...
 
-	if err := m.scheduler.CheckReplicasSizeExpansion(v, v.Spec.Size, size); err != nil {
+	multiError, err := m.scheduler.CheckReplicasSizeExpansion(v, v.Spec.Size, size)
+	if err != nil {
+		if multiError != nil {
+			failureMessage := multiError.Join()
+			if err := m.ds.UpdatePVAnnotation(v, types.PVAnnotationLonghornVolumeSchedulingError, failureMessage); err != nil {
+				logrus.Warnf("Cannot update PV annotation for volume %v", volumeName)
+			}
+		}
```
innobead, line 571: I don't think we need to update during the validation phase, but rather only do this during the volume size reconciling, because a validation failure just means the request has not been accepted and nothing changes to the volume resource change, right?

datastore/longhorn.go
```diff
@@ -3790,3 +3790,28 @@ func (s *DataStore) ListOrphansByNodeRO(name string) ([]*longhorn.Orphan, error)
 func (s *DataStore) DeleteOrphan(orphanName string) error {
 	return s.lhClient.LonghornV1beta2().Orphans(s.namespace).Delete(context.TODO(), orphanName, metav1.DeleteOptions{})
 }
+
+func (s *DataStore) UpdatePVAnnotation(volume *longhorn.Volume, annotationKey, annotationVal string) error {
```
innobead, line 3794: This is for updating a builtin resource, so will it be better to move to datastore/kubernetes.go?

cc @c3y1huang

innobead (changes requested):

manager/volume.go
derekbit: I think yes. The volume is at detached state, and PV/PVC STATUS are `Bound` in the validation failure case. So, the Havester should check the status of the volume (I mean detached`), right?

datastore/longhorn.go
derekbit: Updated, thank you.

manager/volume.go
derekbit: Should Harvester check the PVC error message (kubectl describe pvc ...) rather than volume-scheduling-error label?

scheduler/replica_scheduler.go
```diff
@@ -678,14 +677,14 @@ func (rcs *ReplicaScheduler) GetDiskSchedulingInfo(disk longhorn.DiskSpec, diskS
 	return info, nil
 }
 
-func (rcs *ReplicaScheduler) CheckReplicasSizeExpansion(v *longhorn.Volume, oldSize, newSize int64) (err error) {
+func (rcs *ReplicaScheduler) CheckReplicasSizeExpansion(v *longhorn.Volume, oldSize, newSize int64) (multiError util.MultiError, err error) {
```
innobead, line 680: Do we really need another wrap for the returned error? or using the returned error is good enough?

controller/volume_controller.go
```diff
@@ -2678,8 +2678,14 @@ func (vc *VolumeController) reconcileVolumeSize(v *longhorn.Volume, e *longhorn.
 		vc.eventRecorder.Eventf(v, v1.EventTypeNormal, EventReasonCanceledExpansion,
 			"Canceled expanding the volume %v, will automatically detach it", v.Name)
 	} else {
-		if err := vc.scheduler.CheckReplicasSizeExpansion(v, e.Spec.VolumeSize, v.Spec.Size); err != nil {
+		if multiError, err := vc.scheduler.CheckReplicasSizeExpansion(v, e.Spec.VolumeSize, v.Spec.Size); err != nil {
 			log.Debugf("cannot start volume expansion: %v", err)
+			if multiError != nil {
+				failureMessage := multiError.Join()
+				if err := vc.ds.UpdatePVAnnotation(v, types.PVAnnotationLonghornVolumeSchedulingError, failureMessage); err != nil {
```
innobead, line 2685: Update the annotation w/ the returned error instead of having another multiError? good enough?

scheduler/replica_scheduler.go
derekbit: The error contains different error messages, not sure if volume-scheduling-eerror should cover these different error messages. But I would prefer to let `multiError` focus on disk or space issue.

controller/volume_controller.go
derekbit: Same above.

manager/volume.go
innobead: If the PCV is bound but the corresponding volume replicas can't be scheduled, the downstream users will be unable to check any error messages from the PVC. 

Therefore, the only way is to rely on checking the volume status condition `scheduled` or the updated annotation ` volume-scheduling-error` of the PVC. 

Anything missing?

scheduler/replica_scheduler.go
innobead: I see, then I would suggest renaming the returned error name to `diskScheduleMultiError`. or similar to make it meaningful. WDYT?

manager/volume.go
derekbit: Yeah, if we don't set `volume-scheduling-error`, checking volume status is the only way.

scheduler/replica_scheduler.go
derekbit: Sounds good!

derekbit: Updated. Thank you.

manager/volume.go
derekbit: Updated. Thank you.

innobead (approved):
LGTM

scheduler/replica_scheduler.go
```diff
@@ -678,14 +677,14 @@ func (rcs *ReplicaScheduler) GetDiskSchedulingInfo(disk longhorn.DiskSpec, diskS
 	return info, nil
 }
 
-func (rcs *ReplicaScheduler) CheckReplicasSizeExpansion(v *longhorn.Volume, oldSize, newSize int64) (err error) {
+func (rcs *ReplicaScheduler) CheckReplicasSizeExpansion(v *longhorn.Volume, oldSize, newSize int64) (diskScheduleMultiError util.MultiError, err error) {
```
shuo-wu, line 680: NIT: returning 2 errors looks a little bit weird

shuo-wu (commented):
Actually, I have one question regarding the 1st commit: does it affect RWX volumes only? The fix seems to be applied to all volumes.

derekbit:
> Actually, I have one question regarding the 1st commit: does it affect RWX volumes only? The fix seems to be applied to all volumes.

The commit applies to all volumes.

scheduler/replica_scheduler.go
derekbit: Yeah, I'm thinking how to refactor the `MultiError, error` in replica scheduler as well as this check function.

shuo-wu:
> Yeah, the commit applies to all volumes.

Got it. I just feel a little bit confused about the title of the 1st commit.

scheduler/replica_scheduler.go
shuo-wu: Sure. We can check it later. it's not a high-priority issue.

PR #1518, opened 2022-10-08
Kubernetes 1.25 support

PhanLe1010:
1. Remove PSP and add labels to longhorn-system namespace to support Pod Security Admission
1. Support K8s 1.25 by updating deprecated resource versions
    * `batch/v1beta1` to `batch/v1`
    * `policy/v1beta1` to `policy/v1`
1. Bump K8s min version support and CSI sidecar versions
    * Min Kubernetes version from 1.18 to 1.21
    * longhornio/csi-resizer v1.2.0 -> v1.3.0
    * longhornio/csi-snapshotter: v3.0.3 -> v5.0.1
1. Update CSI snapshot examples from v1beta1 to v1
   

longhorn/longhorn#4003
longhorn/longhorn#4239


deploy/install/01-prerequisite/01-namespace.yaml
```diff
@@ -2,3 +2,10 @@ apiVersion: v1
 kind: Namespace
 metadata:
   name: longhorn-system
+  labels:
+    pod-security.kubernetes.io/enforce: privileged
+    pod-security.kubernetes.io/enforce-version: latest
+    pod-security.kubernetes.io/audit: privileged
+    pod-security.kubernetes.io/audit-version: latest
+    pod-security.kubernetes.io/warn: privileged
+    pod-security.kubernetes.io/warn-version: latest
```
shuo-wu, line 11: 1. For the pods that don't satisfy the `privileged` policy requirements, will Kubernetes block them or just generate warnings/audit annotations?
2. Why is the (minor) version `latest` rather than a specific version like v1.25?

PhanLe1010: > 1. For the pods that don't satisfy the privileged policy requirements, will Kubernetes block them or just generate warnings/audit annotations?

With this setting, Kubernetes doesn't block/warn any pods (privileged/baseline/restricted) .

* `pod-security.kubernetes.io/enforce: privileged` means blocking pods requiring more permission than `privileged`. That means don't block any pods inside `longhorn-system` namespace because there is no pod that requires more permission than the privileged level.
* `pod-security.kubernetes.io/audit: privileged` means generating audit events for pods requiring more permission than `privileged`. That means don't generate any events for pods inside `longhorn-system` namespace because there is no pod that requires more permission than the privileged level.
* `pod-security.kubernetes.io/audit: privileged` means generating warning message for pods requiring more permission than `privileged`. That means don't generate any warning for pods inside `longhorn-system` namespace because there is no pod that requires more permission than the privileged level.

> Why is the (minor) version latest rather than a specific version like v1.25?

It doesn't really matter which Pod Security Standard version we are pinning here as the setting we are doing is `don't block/warn any pod according to xyz PSA version`. This will instruct the PSA controller not to use the default value in case the default value is `blocking if missing labels on the namespace`.

shuo-wu: Got it. I was thinking `enforce` will block the unsatisfied pods while `audit` or `warn` will allow the pods, which is a conflict. But actually, the latter two are just for warning or audit rather than accepting or rejecting the pods.

As for the versioning, the doc does not explain too much about this part. If the version is `latest`, does this mean that the PSS we add applies to the Longhorn namespace under any Kubernetes version?

PhanLe1010: > As for the versioning, the doc does not explain too much about this part. If the version is latest, does this mean that the PSS we add applies to the Longhorn namespace under any Kubernetes version?

There is an example at [this link](https://kubernetes.io/docs/setup/best-practices/enforcing-pod-security-standards/#adopt-a-multi-mode-strategy)

I think the PSA controller will:
1. Get the current latest Kubernetes version from upstream and get the PSS definitions from that version
2. Use these definitions to evaluate pods

shuo-wu (commented):
In general LGTM except for 2 questions:
1. When upgrading Kubernetes version to v1.25, does a running Longhorn system need to do anything for existing PDBs and CronJobs (with API version v1beta1)?
2. The uninstaller job/pod is running in namespace `default`. It means the PSS we added for namespace `longhorn-system` cannot be applied to it. Will this lead to any uninstall issue?

deploy/longhorn-images.txt
```diff
@@ -1,7 +1,7 @@
 longhornio/csi-attacher:v3.4.0
 longhornio/csi-provisioner:v2.1.2
-longhornio/csi-resizer:v1.2.0
-longhornio/csi-snapshotter:v3.0.3
+longhornio/csi-resizer:v1.3.0
+longhornio/csi-snapshotter:v5.0.1
```
derekbit, line 4: Just curious. Why not upgrade to latest versions?

PhanLe1010:
> 1. When upgrading Kubernetes version to v1.25, does a running Longhorn system need to do anything for existing PDBs and CronJobs (with API version v1beta1)?

No, Longhorn is a client. These CRDs are shipped by Kubernetes so Kubernetes should handle conversion/storage/serve versions. Also, there are no noticeable schema changes for [cronjob](https://kubernetes.io/docs/reference/using-api/deprecation-guide/#cronjob-v125) and [PDB](https://kubernetes.io/docs/reference/using-api/deprecation-guide/#poddisruptionbudget-v125). There is a small behavior change with PDB but it doesn't applied to the our PDB as we have non-empty `spec.selector` field 

> 2. The uninstaller job/pod is running in namespace default. It means the PSS we added for namespace longhorn-system cannot be applied to it. Will this lead to any uninstall issue?

Great catch. Yeah, the installation will be stuck. Let's move them into the `longhorn-system`. What do you think? @shuo-wu @derekbit @innobead ?

deploy/longhorn-images.txt
PhanLe1010: v5.0.1 deprecated but is still serving CSI snapshot `v1beta1` while v.6.0.1 stopped serving it. This gives users some buffer to migrate their CSI snapshot manifest/script that is using `v1beta1`. What do you think?

derekbit: > This gives users some buffer to migrate their CSI snapshot manifest/script that is using v1beta1. 

Got it. Agreed the buffer. Thank you.

shuo-wu (approved):

derekbit (approved):

innobead:
LGTM, but what was the reason before we added the uninstaller in the default namespace? any possible side effect to move to the longhorn-system?

PhanLe1010:
> LGTM, but what was the reason before we added the uninstaller in the default namespace like avoid the longhorn-system deleted by accident? Any possible side effect to moving to the longhorn-system?

I don't know why we we added the uninstaller in the default namespace before. Do you have any idea @yasker @shuo-wu @joshimoo ?  I don't see any issue moving it into `longhorn-system` namespace.

> BTW, this commit will be temporary because the duplicated manifests in the deploy folder will be removed in near future but rather just keep similar ones in the longhorn/longhorn repo only.

Got it. I also copied the installer manifest to longhorn/longhorn PR https://github.com/longhorn/longhorn/pull/4695

datastore/datastore.go
```diff
@@ -5,9 +5,9 @@ import (
 	"k8s.io/client-go/informers"
 	clientset "k8s.io/client-go/kubernetes"
 	appslisters "k8s.io/client-go/listers/apps/v1"
-	batchlisters_v1beta1 "k8s.io/client-go/listers/batch/v1beta1"
+	batchlisters_v1 "k8s.io/client-go/listers/batch/v1"
 	corelisters "k8s.io/client-go/listers/core/v1"
-	policylisters "k8s.io/client-go/listers/policy/v1beta1"
+	policylisters "k8s.io/client-go/listers/policy/v1"
```
joshimoo, line 10: NIT: for group discussion any reason we include version info in the naming for some of these but not others?

deploy/install/01-prerequisite/05-podsecuritypolicy.yaml
```diff
@@ -1,61 +1,61 @@
...
-  kind: Role
-  name: longhorn-psp-role
-subjects:
-  - kind: ServiceAccount
-    name: longhorn-service-account
-    namespace: longhorn-system
-  - kind: ServiceAccount
-    name: default
-    namespace: longhorn-system
+#apiVersion: policy/v1beta1
```
joshimoo, line 1: NIT: any reason we keep the file around in the code base? I guess we can clean it up later?
How does `kubectl apply -f` react on an empty file when applied to the whole directory?

deploy/longhorn-images.txt
joshimoo: @innobead for the release note we also need to make sure that users have cleaned up their potentially previously invalid csi snapshots, since v4.1 they will no longer be served and I think there might be problems with the controller in case of invalid objects.

REF: https://github.com/kubernetes-csi/external-snapshotter#snapshot-v1-apis

deploy/uninstall/uninstall.yaml
```diff
@@ -90,13 +90,13 @@ roleRef:
...
     name: longhorn-uninstall-service-account
-    namespace: default
+    namespace: longhorn-system
 ---
 apiVersion: batch/v1
 kind: Job
 metadata:
   name: longhorn-uninstall
-  namespace: default
+  namespace: longhorn-system
```
joshimoo, line 99: I think the uninstall job was initially run outside of the `longhorn-system` namespace since it would also require cleaning up the namespace itself. And therefore there would be no way to see the status of the uninstall.

Is there any dependency on the HELM side that would require to know the status of the uninstall pod / job?

joshimoo (approved):
LGTM, uninstall change needs to be verified.

datastore/datastore.go
PhanLe1010: Good point! I agree that we should standardize it be `name+version` in other ticket

deploy/install/01-prerequisite/05-podsecuritypolicy.yaml
PhanLe1010: > NIT: any reason we keep the file around in the code base? I guess we can clean it up later?

Kubernetes < 1.25 users may still need PSP so we keep the comment there for them for convenient

> How does kubectl apply -f react on an empty file when applied to the whole directory?

It is happy from the testing

deploy/uninstall/uninstall.yaml
PhanLe1010: > I think the uninstall job was initially run outside of the longhorn-system namespace since it would also require cleaning up the namespace itself. And therefore there would be no way to see the status of the uninstall.

The installation pod doesn't seem to make attempt to clean up the `longhorn-system` namespace from testing and code

> Is there any dependency on the HELM side that would require to know the status of the uninstall pod / job?

In Helm, the installation job has been running inside the same `longhorn-system` namespace so far. Foe Helm, the flow is currently (as will stay the same):
1. installation job runs as Helm `pre-delete` hook
2. Helm watches the job to be completed and start deleting manifests
3. The namespace still exist after Helm installation

deploy/longhorn-images.txt
innobead: :point_up: let's make this info in the source ticket description as well to avoid any missing.

deploy/uninstall/uninstall.yaml
innobead: I think the uninstall job in the default namespace only exists in longhorn-manager repo, but if looking into the chart where the uninstall job there is actually just in the same namespace as the installed longhorn. 

So, it should be fine to use the same namespace to run the uninstall job. As I mentioned, the duplicated manifests will be removed from the longhorn-manager repo, also installing/uninstalling from longhorn-manager is not a formal way for users but rather for our internal development purpose only.

datastore/datastore.go
innobead: Let's create a ticket for this refactoring.

innobead (approved):

deploy/longhorn-images.txt
innobead: @PhanLe1010 Why didn't we upgrade `external-provisioner`? I saw the latest 3.3 can have min 1.20 supported. The same question for `[external-attacher](https://kubernetes-csi.github.io/docs/external-attacher.html#csi-external-attacher)`

cc @weizhe0422

PhanLe1010: I followed this steps to update the components, In steps 5, it talked about how to select csi sidecars components. 

Longhorn v1.4.0 supports min K8s v1.21. On the other hand, k8s v.1.21 is speaking CSI spec v1.3.0 version.  Therefore, Longhorn components also need to support min csi spec version 1.3.0 so that we can talk to K8s v.1.21. 

Using the min csi spec version v1.3.0, we cannot upgrade external-provisioner to newer than v2.1.2. Same thing for other csi components

PR #1519, opened 2022-10-11
feat: Snapshot checksum calculation and verification

derekbit:
longhorn/longhorn#4210
longhorn/longhorn#3198


controller/monitor/snapshot_monitor.go
```diff
@@ -0,0 +1,467 @@
...
+	for _, engine := range engines {
+		m.logger.WithField("monitor", monitorName).Infof("Populating engine %v snapshots", engine.Name)
+		m.populateEngineSnapshots(engine)
+	}
+}
+
+func (m *SnapshotMonitor) populateEngineSnapshots(engine *longhorn.Engine) {
+	snapshots := engine.Status.Snapshots
+	for _, snapshot := range snapshots {
+		if snapshot.Name == ediskutil.VolumeHeadName || !snapshot.UserCreated || snapshot.Removed {
```
shuo-wu, line 190: NIT: add a comment that explains why we skip these snapshots. (The snapshot prune issue)

controller/monitor/snapshot_monitor.go
```diff
@@ -0,0 +1,467 @@
...
+			return
+		case <-ticker.C:
+			v := m.settings[string(types.SettingNameSnapshotDataIntegrity)]
+			dataIntegrity := v.(string)
+
+			v = m.settings[string(types.SettingNameSnapshotDataIntegrityInterval)]
+			dataIntegrityInterval := v.(int64)
+
+			if dataIntegrity == string(types.SnapshotDataIntegrityDisabled) ||
+				!time.Now().After(m.lastSnapshotVerifiedAt.Time.Add(time.Duration(dataIntegrityInterval)*time.Minute)) {
```
shuo-wu, line 161: It seems that some newly created snapshots should wait for `dataIntegrityInterval` before starting the hash.
The pro is the resync/re-verification mechanism is straightforward. But the con is that the calculation for all snapshots is started at one single moment hence there will be relatively high CPU & Disk usage since that moment.

controller/node_controller.go
```diff
@@ -179,6 +195,50 @@ func (nc *NodeController) isResponsibleForReplica(obj interface{}) bool {
...
+
+	if volume.Status.OwnerID != nc.controllerID {
+		return false
+	}
+
+	return true
+
+}
+
+func (nc *NodeController) isResponsibleForEngine(obj interface{}) bool {
```
shuo-wu, line 224: NIT: Unused function

controller/node_controller.go
```diff
@@ -179,6 +195,50 @@ func (nc *NodeController) isResponsibleForReplica(obj interface{}) bool {
 	return replica.Spec.NodeID == nc.controllerID
 }
 
+func (nc *NodeController) isResponsibleForSnapshot(obj interface{}) bool {
```
shuo-wu, line 198: NIT: actually this function can directly return `false` if the setting is disabled.

shuo-wu (changes requested):

controller/monitor/snapshot_monitor.go
```diff
@@ -0,0 +1,467 @@
...
+
+	if val, ok := m.settings[string(types.SettingNameSnapshotDataIntegrity)]; ok && val == string(types.SnapshotDataIntegrityEnabled) {
+		rehash = true
+	}
+
+	if val, exist := m.settings[string(types.SettingNameSnapshotDataIntegrityPerNodeConcurrentLimit)]; exist {
+		concurrentLimit = val.(int)
+	}
+
+	return engineClientProxy.SnapshotHash(engine, snapshotName, rehash, concurrentLimit)
```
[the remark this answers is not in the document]
derekbit: You are right. The concurrent limit is wrong here. I'm thinking use a file lock method. Do you have any idea?

controller/monitor/snapshot_monitor.go
derekbit: For the newly created snapshots, they are sents to workers for hashing and don't need to wait for `dataIntegrityInterval`. The path is 

`Node controller (aware of snapshot update) -> enqueueSnapshot -> snapshotChangeEventChan -> snapshotVerifyTaskChan -> workers`

controller/node_controller.go
derekbit: Updated. Thank you.

controller/monitor/snapshot_monitor.go
derekbit: Added, thank you.

controller/monitor/snapshot_monitor.go
```diff
@@ -0,0 +1,467 @@
...
+	checksum := ""
+	maxVotes := 0
+	for c, replicas := range checksums {
+		if len(replicas) > maxVotes {
+			checksum = c
+			maxVotes = len(replicas)
+		}
+	}
+
+	if maxVotes == 1 && len(hashStatus) > 1 {
```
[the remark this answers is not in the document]
derekbit: > Shouldn't the function check the existing non-empty checksum in snapshot.status first?

Why check the snapshot.status.checksum first?
The new checksum will be determined here and be updated to snapshot.status.checksum. If the checksum is identical, the `UpdateSnapshotStatus` won't take place. WDYT?

controller/monitor/snapshot_monitor.go
shuo-wu: OK. Got it. The newly created snapshots should be fine. But the pro and con I mentioned still exist.

controller/monitor/snapshot_monitor.go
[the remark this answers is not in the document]
shuo-wu: I am not sure if using a file lock would help. As we discussed before, one snapshot monitor needs to check multiple node concurrent task counts before starting one snapshot hashing. Then how does the snapshot monitor apply the file locking on other nodes?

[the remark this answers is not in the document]
derekbit: Yes. This is a method we can do. I'm just thinking if there is any better solution.

controller/monitor/snapshot_monitor.go
derekbit: Yeah, you're right. I created a [ticket](https://github.com/longhorn/longhorn/issues/4765) to throttle the background io for reducing the performance impact. I think this will be helpful for the con.

controller/monitor/snapshot_monitor.go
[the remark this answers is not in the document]
shuo-wu: One solution I find now is adding a new field `node.Status.CurrentHashingCount`. This actually utilizes the consistency/consensus of ETCD. Before starting the hash of one snapshot, the snap monitor needs to increase the count for all node CRs. Then the bad thing is, the snap monitor on one node will update the status for other node CRs.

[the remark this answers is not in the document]
derekbit: > Then the bad thing is, the snap monitor on one node will update the status for other node CRs.

Yes.

> one snapshot monitor needs to check multiple node concurrent task counts before starting one snapshot hashing. 

I think I can work on this solution first and check if there is any outstanding issue.

controller/monitor/snapshot_monitor.go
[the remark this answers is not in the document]
derekbit: Handled all cases in the latest PR and added a unit test for `determineChecksumFromHashStatus`. Thank you.

controller/monitor/snapshot_monitor.go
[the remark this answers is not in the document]
shuo-wu: OK. Then let's see if anyone has better ideas.
cc @joshimoo @PhanLe1010 @c3y1huang @innobead

derekbit:
Please help review. Thank you.
cc @shuo-wu @c3y1huang @innobead

controller/monitor/snapshot_monitor.go
```diff
@@ -0,0 +1,588 @@
...
+)
+
+type SnapshotHashStatus string
+
+const (
+	SnapshotHashStatusNew        = SnapshotHashStatus("")
+	SnapshotHashStatusInProgress = SnapshotHashStatus("in_progress")
+	SnapshotHashStatusCompleted  = SnapshotHashStatus("complete")
+	SnapshotHashStatusError      = SnapshotHashStatus("error")
+)
```
shuo-wu, line 40: NIT: Shouldn't these be in `longhorn-manager/k8s/pkg/apis/longhorn/v1beta2/snapshot.go` ?

controller/monitor/snapshot_monitor.go
```diff
@@ -0,0 +1,588 @@
...
+	task := key.(snapshotCheckTask)
+
+	dataIntegrity, err := m.getSnapshotDataIntegrity(task.volumeName)
+	if err != nil {
+		return true
+	}
+
+	if dataIntegrity == longhorn.SnapshotDataIntegrityDisabled {
+		return true
+	}
```
shuo-wu, line 217: NIT: Do we need to check it twice? We already filter the snapshots in the change event queue.

controller/monitor/snapshot_monitor.go
```diff
@@ -0,0 +1,588 @@
...
+
+	m.logger.WithField("monitor", monitorName).Debugf("Received snapshot check task %+v", task)
```
shuo-wu, line 219: Will this flood the longhorn manager logs in a large cluster?

controller/monitor/snapshot_monitor.go
```diff
@@ -0,0 +1,588 @@
...
+	if err != nil {
+		return err
+	}
+	if v.Spec.MigrationNodeID != "" {
+		return fmt.Errorf("cannot operate during migration")
+	}
+	return nil
+}
+
+func (m *SnapshotMonitor) getEngineBinaryClient(volumeName string) (client *engineapi.EngineBinary, err error) {
```
shuo-wu, line 422: NIT: Do we have an existing function for this? It seems that I have seen this somewhere else.

controller/monitor/snapshot_monitor.go
```diff
@@ -0,0 +1,588 @@
...
+		return errors.Wrap(err, "failed to schedule snapshot check job")
+	}
+
+	m.Lock()
+	m.existingDataIntegrityCronJob = dataIntegrityCronJob
+	m.Unlock()
+
+	m.checkScheduler.StartAsync()
+
+	m.logger.WithField("monitor", monitorName).Infof("Next snapshot check job will be executed at %v", job.NextRun())
```
shuo-wu, line 290: NIT: print out the old and the new crons.

controller/monitor/snapshot_monitor.go
```diff
@@ -0,0 +1,588 @@
...
+		return nil
+	}
+	defer m.deleteFromInProgressSnapshotCheckTasks(task.snapshotName)
+
+	engine, err := m.ds.GetVolumeCurrentEngine(task.volumeName)
+	if err != nil {
+		return errors.Wrapf(err, "failed to get engine for volume %v", task.volumeName)
+	}
+
+	if err := m.checkVolumeNotInMigration(engine.Spec.VolumeName); err != nil {
```
shuo-wu, line 338: May need to check other operations like rebuilding and purge

controller/monitor/snapshot_monitor.go
```diff
@@ -0,0 +1,588 @@
...
+	}
+
+	engineClientProxy, err := engineapi.GetCompatibleClient(engine, engineCliClient, m.ds, m.logger, m.proxyConnCounter)
+	if err != nil {
+		return err
+	}
+	defer engineClientProxy.Close()
+
+	m.Lock()
+	m.SnapshotCheckState = longhorn.SnapshotCheckStateInProgress
```
shuo-wu, line 354: I don't get why the monitor needs this state

controller/monitor/snapshot_monitor.go
```diff
@@ -0,0 +1,588 @@
...
+	existingSnapshot := snapshot.DeepCopy()
+
+	checksum, err := determineChecksumFromHashStatus(snapshotName, snapshot.Status.Checksum, hashStatus)
+	if err != nil {
+		m.eventRecorder.Eventf(engine, v1.EventTypeWarning, constant.EventReasonFailedSnapshotDataIntegrityCheck,
+			"Failed to check the data integrity of snapshot %v for volume %v", snapshotName, engine.Spec.VolumeName)
+		return errors.Wrapf(err, "failed to determine checksum for snapshot %v", snapshotName)
+	}
+
+	snapshot.Status.Checksum = checksum
```
shuo-wu, line 490: Just realized that the hashing state and progress are not recorded in the Snapshot CRs. Do we need to record it?

controller/monitor/snapshot_monitor.go
```diff
@@ -0,0 +1,588 @@
...
+	}
+
+	found, checksum, maxVotes := determineChecksum(checksums)
+	if found {
+		return checksum, nil
+	}
+
+	if existingChecksum == "" {
+		return "", fmt.Errorf(prefixChecksumDetermineFailure+"since there is no existing checksum", snapshotName)
+	}
```
shuo-wu, line 541: Should we error out here? For a newly hashed snapshot, this should be empty... How about directly input `existingChecksum` as an input of `determineChecksum`? `determineChecksum` will use the vote mechanism only if `existingChecksum` is empty.

api/model.go
```diff
@@ -1184,7 +1185,7 @@ func toVolumeResource(v *longhorn.Volume, ves []*longhorn.Engine, vrs []*longhor
 	return r
 }
 
-func toSnapshotResource(s *longhorn.SnapshotInfo) *Snapshot {
+func toSnapshotResource(s *longhorn.SnapshotInfo, checksum string) *Snapshot {
```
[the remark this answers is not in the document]
derekbit: I'm not sure if there is any negative potential impact to the existing `snapshotList`. Will it be a problem when use snapshot CRs?

controller/monitor/snapshot_monitor.go
derekbit: We can remove it for avoding the potential issue, because longhorn-manager uses debug level log for now.
But I think using debug-level log in a production enviroment is weird.

controller/monitor/snapshot_monitor.go
derekbit: I think we don't need to record the transisent information. We don't plan to expose the snapshot checksum information to users explicitly in the first version.

controller/monitor/snapshot_monitor.go
```diff
@@ -0,0 +1,588 @@
...
+		retry.Delay(snapshotHashSyncStatusPeriod * time.Second),
+		retry.RetryIf(func(err error) bool {
+			if err == nil {
+				return false
+			}
+			return strings.Contains(err.Error(), string(engineapi.ProcessStateInProgress))
+		}),
+	}
+
+	if err := retry.Do(func() (err error) {
```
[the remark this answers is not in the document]
derekbit: Cannot agree with you more. :D

api/model.go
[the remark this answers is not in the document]
derekbit: cc @PhanLe1010

controller/monitor/snapshot_monitor.go
derekbit: I'm thinking if we need to show `The node is doing snapshot checksum calculation` on node ui page, and the users can expect the system might slow down.
If no, we don't need the state.
WDYT?

controller/monitor/snapshot_monitor.go
```diff
@@ -0,0 +1,588 @@
...
+		return err
+	}
+
+	return m.waitForSnapshotHashingComplete(engine, engineClientProxy, task.snapshotName)
+}
+
+func (m *SnapshotMonitor) requestSnapshotHashing(engine *longhorn.Engine, engineClientProxy engineapi.EngineClientProxy,
+	snapshotName string, changeEvent bool) error {
+	rehash := false
+	if !changeEvent {
```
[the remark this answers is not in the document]
derekbit: Yeah. Because the one snapshot CR might be updated many times in a short period. I expect it just calculates checksum once. If `refresh` is false and the checksum is existing, don't need to calculate it again if the `ctime` is not changed.

The periodc snapshot check mechanism will do the regular checks.

api/model.go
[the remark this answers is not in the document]
shuo-wu: This is caused by the legacy/deprecating field `Engine.Status. SnapshotInfo`. I am not sure either, maybe we can review it in the future.

controller/monitor/snapshot_monitor.go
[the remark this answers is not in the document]
shuo-wu: You can add a simple comment for that.

controller/monitor/snapshot_monitor.go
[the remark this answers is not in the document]
shuo-wu: OK. In other words, the full hash will be issued by the cron job only, no matter if the field is `enabled` or `fast-check`.

controller/monitor/snapshot_monitor.go
shuo-wu: Yes. You can enable this after changing the default log level to info for the longhorn managers.

controller/monitor/snapshot_monitor.go
derekbit: Yes, in `manager/engine.go`, `controller/volume_controller.go` and `controller/backup_controller.go`.
If I would like to refactor it. Which package is better for `getEngineBinaryClient`?

controller/monitor/snapshot_monitor.go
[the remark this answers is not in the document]
derekbit: Yes. You're right.

controller/monitor/snapshot_monitor.go
shuo-wu: I think it's `engineapi`

controller/monitor/snapshot_monitor.go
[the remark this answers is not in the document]
shuo-wu: OK. Please leave a comment here.

controller/node_controller.go
```diff
@@ -179,6 +198,53 @@ func (nc *NodeController) isResponsibleForReplica(obj interface{}) bool {
...
+		dataIntegrity, err := nc.ds.GetSettingValueExisted(types.SettingNameSnapshotDataIntegrity)
+		if err != nil {
+			logrus.Warnf("failed to get %v setting since %v", types.SettingNameSnapshotDataIntegrity, err)
+			return false
+		}
+
+		if longhorn.SnapshotDataIntegrity(dataIntegrity) == longhorn.SnapshotDataIntegrityDisabled {
+			return false
+		}
+	}
```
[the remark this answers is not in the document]
derekbit: Updated, thank you.

controller/monitor/snapshot_monitor.go
derekbit: Just check if users disable the snapshot hashing. If it is disabled, don't need to calculate.

controller/monitor/snapshot_monitor.go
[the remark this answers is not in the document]
derekbit: Added. Thank you.

controller/monitor/snapshot_monitor.go
```diff
@@ -0,0 +1,588 @@
...
+	m.quit()
+}
+
+func (m *SnapshotMonitor) Update(map[string]interface{}) error {
+	dataIntegrityCronJob, err := m.ds.GetSettingValueExisted(types.SettingNameSnapshotDataIntegrityCronJob)
+	if err != nil {
+		return errors.Wrapf(err, "failed to get %v setting", types.SettingNameSnapshotDataIntegrityCronJob)
+	}
+
+	if dataIntegrityCronJob == m.existingDataIntegrityCronJob && m.checkScheduler.Len() > 0 {
```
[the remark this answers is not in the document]
derekbit: Yeah, added a read lock. Thank you.

controller/monitor/snapshot_monitor.go
derekbit: > Should we error out here? For a newly hashed snapshot, this should be empty...

For the case of a newly hashed snapshot, the checksum should be determined by the previous
```
	found, checksum, maxVotes := determineChecksum(checksums)
	if found {
		return checksum, nil
	}
```

> How about directly input existingChecksum as an input of determineChecksum? determineChecksum will use the vote mechanism only if existingChecksum is empty.

The checksums from replicas might be different than prevous values, because purge, trim, corruption and etc. So, the vote mechanism is alway executed to get the latest checksum and then update the `status.checksum`.

controller/monitor/snapshot_monitor.go
derekbit: Addressed. Thank you.

controller/monitor/snapshot_monitor.go
derekbit: Addressed. Thank you.

controller/monitor/snapshot_monitor.go
shuo-wu: It's nice to have. But the concern is, the monitor on this node issuing hash requests does not mean the actual calculation is on this node. 
BTW, why not use a counter instead? There can be 5 workers executing hash tasks simultaneously. Directly updating this field may be an issue.

controller/monitor/snapshot_monitor.go
shuo-wu: OK. Got it. Please add a comment based on this `The checksums from replicas might be different than prevous values, because purge, trim, corruption and etc`

Besides, can this monitor print out a log if there is a checksum change?

controller/monitor/snapshot_monitor.go
derekbit: Yes, the concern is fair. I will probably remove it, bacause it will make users confused.

controller/monitor/snapshot_monitor.go
derekbit: Sure. Will add the log.

derekbit:
e2e tests.
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/2418/
The failures are mainly caused by upgrade tests, because the feature bumps the cli/api versions.

controller/monitor/snapshot_monitor.go
```diff
@@ -0,0 +1,623 @@
...
+	return true
+}
+
+func (m *SnapshotMonitor) handleErr(err error, key interface{}) {
+	if err == nil {
+		m.snapshotCheckTaskQueue.Forget(key)
+		return
+	}
+
+	if !strings.Contains(err.Error(), "cannot request hashing snapshot") {
```
shuo-wu, line 221: Can we have a more outstanding error message (prefix/suffix/keyword) that indicates the upper layer needs to requeue the task?

shuo-wu (commented):
Others LGTM

controller/monitor/snapshot_monitor.go
derekbit: Good idea. Will do it.

api/model.go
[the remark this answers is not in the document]
derekbit: Will handle the issue in https://github.com/longhorn/longhorn/issues/4868.

shuo-wu (dismissed):

derekbit:
Will update dependencies if longhorn-engine PR is merged.

derekbit:
The dependencies and images are updated. Ready for review.
cc @innobead

manager/snapshot.go
```diff
@@ -0,0 +1,13 @@
+package manager
+
+import (
+	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
+)
+
+func (m *VolumeManager) ListSnapshotCRsRO(volumeName string) (map[string]*longhorn.Snapshot, error) {
```
innobead, line 7: Why not just rename `ListSnapshots` to `ListSnapshotInfos` and rename this to `ListSnapshots`? the same for `GetSnapshotCRRO`, because the naming is a little confusing.

constant/events.go
```diff
@@ -44,4 +44,6 @@ const (
 
 	EventReasonFetching = "Fetching"
 	EventReasonSyncing  = "Syncing"
+
+	EventReasonFailedSnapshotDataIntegrityCheck = "FailedSnapshotDataIntegrityChecks"
```
innobead, line 48: `FailedSnapshotDataIntegrityCheck`

k8s/pkg/apis/longhorn/v1beta2/snapshot.go
```diff
@@ -4,6 +4,15 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+type SnapshotHashStatus string
+
+const (
+	SnapshotHashStatusNew        = SnapshotHashStatus("")
+	SnapshotHashStatusInProgress = SnapshotHashStatus("in_progress")
+	SnapshotHashStatusCompleted  = SnapshotHashStatus("complete")
```
innobead, line 12: nit: I know this follows the current convention, but ideally we should use adj word for status/state like`completed` except compound word or noun.

controller/monitor/snapshot_monitor.go
```diff
@@ -0,0 +1,623 @@
...
+	}
+}
+
+func (m *SnapshotMonitor) populateEngineSnapshots(engine *longhorn.Engine) {
+	snapshots := engine.Status.Snapshots
+	for _, snapshot := range snapshots {
+		// Skip volume-head because it is not a real snapshot.
+		// A system-generated snapshot is also ignored, because the prune operations the snapshots are out of
+		// sync during replica rebuilding. More investigation is in https://github.com/longhorn/longhorn/issues/4513
+		if snapshot.Name == etypes.VolumeHeadName || !snapshot.UserCreated {
```
innobead, line 179: Need to skip removed snapshots?

controller/monitor/snapshot_monitor.go
```diff
@@ -0,0 +1,623 @@
...
+	return nil
+}
+
+func (m *SnapshotMonitor) GetData() (interface{}, error) {
+	m.RLock()
+	defer m.RUnlock()
+	return m.SnapshotMonitorStatus, nil
+}
+
+func (m *SnapshotMonitor) addToInProgressSnapshotCheckTasks(snapshotName string) bool {
```
innobead, line 292: Rename to `shouldAddToInProgressSnapshotCheckTasks`

controller/monitor/snapshot_monitor_test.go
```diff
@@ -0,0 +1,548 @@
...
+					Error:    "",
+				},
+				"addr-05": {
+					State:    "Completed",
+					Checksum: "efg",
+					Error:    "",
+				},
+			},
+			existingChecksum: "cde",
+			checksum:         "cde",
```
innobead, line 532: Rename to `expectedChecksum`

controller/monitor/snapshot_monitor_test.go
```diff
@@ -0,0 +1,548 @@
...
+			err:              nil,
```
innobead, line 533: `expectedErr`

innobead (changes requested):
In general, LGTM. Some comments needed to resolve.

controller/monitor/monitor.go
```diff
@@ -12,9 +12,9 @@ import (
 type Monitor interface {
 	Start()
 	Close()
-
-	SyncCollectedData() error
-	GetCollectedData() (interface{}, error)
+	Update(map[string]interface{}) error
+	RunTask(interface{}) error
```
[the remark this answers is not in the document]
innobead: Of course, for any downstream structure, it can override all methods for it's behavior, but because we have baseMonitor, can think about how to make reusable code to the parent class.

k8s/pkg/apis/longhorn/v1beta2/snapshot.go
derekbit: Actually, there are many progress status sets in longhorn-manager, longhorn-engine and other packages. I think we can unify them in the future. However, the compatibility will be an issue...

innobead: Exactly. Let's create a ticket for the future refactoring, not a priority for now.

api/model.go
```diff
@@ -1195,13 +1196,21 @@ func toSnapshotResource(s *longhorn.SnapshotInfo) *Snapshot {
 			Type: "snapshot",
 		},
 		SnapshotInfo: *s,
+		Checksum:     checksum,
 	}
 }
 
-func toSnapshotCollection(ss map[string]*longhorn.SnapshotInfo) *client.GenericCollection {
+func toSnapshotCollection(ssList map[string]*longhorn.SnapshotInfo, ssListRO map[string]*longhorn.Snapshot) *client.GenericCollection {
```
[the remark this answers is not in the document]
derekbit: Yes, we will handle this issue in https://github.com/longhorn/longhorn/issues/4868

controller/monitor/snapshot_monitor.go
derekbit: I think we can still handle removed snapshots. Any negative issue here? cc @shuo-wu

innobead: but is it meaningful?

controller/monitor/disk_monitor.go
```diff
@@ -95,7 +95,11 @@ func (m *NodeMonitor) Close() {
 	m.quit()
 }
```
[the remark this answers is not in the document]
derekbit: It looks not easy to move `Start` and `Close` to `baseMonitor` currently. The implementation of the to monitors are quite different. We can generalize them when refractoring `backup` and `engine` monitors.

engineapi/engine.go
```diff
@@ -266,3 +266,13 @@ func (e *EngineBinary) ReplicaRebuildVerify(engine *longhorn.Engine, url string)
 // Do not panic this method because this is could be called by the fallback client.
 func (e *EngineBinary) Close() {
 }
+
+// ReplicaModeUpdate calls engine binary
+// TODO: Deprecated, replaced by gRPC proxy
+func (e *EngineBinary) ReplicaModeUpdate(engine *longhorn.Engine, url, mode string) error {
+	_, err := e.ExecuteEngineBinary("update", "--mode", mode, url)
+	if err != nil {
+		return errors.Wrapf(err, "failed to list replicas from controller '%s'", e.name)
```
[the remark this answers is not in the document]
derekbit: Thank you. Forgot updating the error message.

manager/snapshot.go
derekbit: Updated. Thank you.

api/snapshot.go
```diff
@@ -58,7 +58,10 @@ func (s *Server) SnapshotList(w http.ResponseWriter, req *http.Request) (err err
 	if err != nil {
 		return err
 	}
-	api.GetApiContext(req).Write(toSnapshotCollection(snapList))
+
+	snapListRO, _ := s.m.ListSnapshotCRsRO(volName)
+
+	api.GetApiContext(req).Write(toSnapshotCollection(snapList, snapListRO))
 	return nil
```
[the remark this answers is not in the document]
derekbit: Updated. Thank you.

controller/monitor/snapshot_monitor.go
derekbit: Updated. Thank you.

controller/monitor/snapshot_monitor_test.go
derekbit: Updated. Thank you.

controller/node_controller.go
```diff
@@ -432,19 +430,18 @@ func (nc *NodeController) syncNode(key string) (err error) {
...
 	}
 
 	if nc.controllerID != node.Name {
 		return nil
 	}
 
-	mon, err := nc.checkMonitor(nc.controllerID)
-	if err != nil {
+	// Create a monitor for collecting disk information
+	if _, err := nc.createDiskMonitor(nc.controllerID); err != nil {
```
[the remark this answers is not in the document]
derekbit: Addressed. Thank you.

For the nit part, I will handle them when refractoring other monitors.

controller/monitor/monitor.go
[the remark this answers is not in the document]
derekbit: The purpose of exporting `Run` is to do the unit test, but I also think it not good here. I think we can make it private when refractorig `backup` and `engine` monitors. Then, we can make the methods more generalized.

`GetData` and `Update` will by called by controllers for retrieving the information from monitors or upadte the monitors' configurations.

controller/monitor/snapshot_monitor.go
derekbit: I think yes. For example, the trim feature will produce a tons of removed snaphots if users enable the `UnmapMarkSnapChainRemoved`. Then, the resending the removed snapshots is still time-consuming.

innobead: then, let's keep this first.

derekbit: BTW, a `removed` snapshot still contains in-use data before merging, so checksumming it is still necessary.

controller/monitor/monitor.go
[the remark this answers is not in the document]
derekbit: Renamed the methods. Thank you.

controller/monitor/disk_monitor.go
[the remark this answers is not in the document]
derekbit: Improve in this ticket https://github.com/longhorn/longhorn/issues/4897.

manager/snapshot.go
```diff
@@ -0,0 +1,13 @@
...
+	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
+)
+
+func (m *VolumeManager) ListSnapshots(volumeName string) (map[string]*longhorn.Snapshot, error) {
+	return m.ds.ListVolumeSnapshotsRO(volumeName)
+}
+
+func (m *VolumeManager) GetSnapshotRO(snapName string) (*longhorn.Snapshot, error) {
+	return m.ds.GetSnapshotRO(snapName)
+}
```
innobead, line 13: Rename to `GetSnapshot`

innobead (changes requested):

manager/snapshot.go
derekbit: Updated. Thank you.

innobead (approved):

PR #1522, opened 2022-10-12
replica: add compatibility check and handling for ReplicaAdd

derekbit:
[Longhorn 4705](https://github.com/longhorn/longhorn/issues/4705)

### DIFF
--- a/engineapi/engine.go
+++ b/engineapi/engine.go
@@ -128,6 +128,18 @@ func (e *EngineBinary) ReplicaAdd(engine *longhorn.Engine, url string, isRestore
 	if isRestoreVolume {
 		cmd = append(cmd, "--restore")
 	}
+
+	version, err := e.VersionGet(engine, true)
+	if err != nil {
+		return err
+	}
+
+	if version.ClientVersion.CLIAPIVersion >= 6 {
+		cmd = append(cmd,
+			"--size", strconv.FormatInt(engine.Spec.VolumeSize, 10),
+			"--current-size", strconv.FormatInt(engine.Status.CurrentSize, 10))
+	}
+
 	if _, err := e.ExecuteEngineBinaryWithoutTimeout([]string{}, cmd...); err != nil {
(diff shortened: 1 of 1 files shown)


controller/engine_controller.go
```diff
@@ -1541,14 +1541,18 @@ func (ec *EngineController) startRebuilding(e *longhorn.Engine, replica, addr st
 		}
 
 		// start rebuild
+		sizeRequested, err := ec.ds.IsEngineImageCLIAPIVersionEqualToOrLargerThan(e.Spec.EngineImage, 6)
+		if err != nil {
```
c3y1huang, line 1545: Should log error.

c3y1huang (changes requested):

innobead:
> After discussion, this is not quite right, because new changes to engine operations should only be added to the proxy interface instead of the old deprecated binary call anymore.

The root cause has been clarified with @derekbit and @mantissahz . @derekbit will update the context.

controller/engine_controller.go
derekbit: Resubmit the PR. Please help review, thank you. :)

innobead (approved):

innobead:
@derekbit Please backport to 1.3.2. Thanks!

derekbit:
## Pre Ready-For-Testing Checklist
* [ ] Where is the reproduce steps/test steps documented?
The reproduce steps/test steps are at: https://github.com/longhorn/longhorn/issues/4705#issue-1405685731 

* ~~[ ] Is there a workaround for the issue? If so, where is it documented?~~
The workaround is at:

* [ ] Does the PR include the explanation for the fix or the feature?
https://github.com/longhorn/longhorn/issues/4705#issue-1405685731

* [ ] Have the backend code been merged (Manager, Engine, Instance Manager, BackupStore etc) (including `backport-needed/*`)?
The PR is at https://github.com/longhorn/longhorn-manager/pull/1523

* [ ] Which areas/issues this PR might have potential impacts on?
Area: replica add
Issues

PR #1524, opened 2022-10-13
Add deleting confirmation step to Longhorn uninstallation

PhanLe1010:
longhorn/longhorn#4509

### DIFF
--- a/controller/uninstall_controller.go
+++ b/controller/uninstall_controller.go
@@ -325,6 +325,16 @@ func (c *UninstallController) uninstall() error {
 }
 
 func (c *UninstallController) checkPreconditions() error {
+	confirmationFlag, err := c.ds.GetSettingAsBool(types.SettingNameDeletingConfirmationFlag)
+	if err != nil {
+		return errors.Wrapf(err, "failed to check deleting-confirmation-flag setting")
+	}
+	if !confirmationFlag {
+		return fmt.Errorf("cannot uninstall Longhorn because deleting-confirmation-flag is set to `false`. " +
+			"Please set it to `true` using Longhorn UI or " +
+			"kubectl -n longhorn-system edit settings.longhorn.io deleting-confirmation-flag ")
+	}
+
 	if ready, err := c.managerReady(); err != nil {
 		return err
 	} else if !ready {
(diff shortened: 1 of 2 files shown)


shuo-wu (dismissed):
May need to update default-setting.yaml as well.
In general LGTM

PhanLe1010:
> May need to update default-setting.yaml as well.

types/setting.go
```diff
@@ -843,6 +846,18 @@ var (
 		ReadOnly: false,
 		Default:  "1",
 	}
+
+	SettingDefinitionDeletingConfirmationFlag = SettingDefinition{
+		DisplayName: "Deleting Confirmation Flag",
```
innobead, line 851: `Delete Confirmation Flag` instead?

types/setting.go
```diff
@@ -920,6 +935,10 @@ func ValidateSetting(name, value string) (err error) {
 		if value != "true" && value != "false" {
 			return fmt.Errorf("value %v of setting %v should be true or false", value, sName)
 		}
+	case SettingNameDeletingConfirmationFlag:
+		if value != "true" && value != "false" {
+			return fmt.Errorf("value %v of setting %v should be true or false", value, sName)
+		}
```
innobead, line 941: The implementation is the same as other bool setting, so why not just add this case before `SettingNameUpgradeChecker` then `fallthrough`?

controller/uninstall_controller.go
```diff
@@ -325,6 +325,14 @@ func (c *UninstallController) uninstall() error {
 }
 
 func (c *UninstallController) checkPreconditions() error {
+	confirmationFlag, err := c.ds.GetSettingAsBool(types.SettingNameDeletingConfirmationFlag)
+	if err != nil {
+		return errors.Wrapf(err, "failed to check deleting-confirmation-flag setting")
+	}
+	if !confirmationFlag {
+		return fmt.Errorf("cannot uninstall Longhorn because deleting-confirmation-flag is set to `false`. Please set it to `true` using Longhorn UI or kubectl -n longhorn-system edit settings.longhorn.io deleting-confirmation-flag ")
```
innobead, line 333: The message is quite long. Could we break with a new line for better reading?

innobead (changes requested):
LGTM, some minor requests are required to resolve.

controller/uninstall_controller.go
PhanLe1010: Thank you. Fixed

types/setting.go
PhanLe1010: Ah. I see. Thank you. Fixed

innobead (approved):

PR #1526, opened 2022-10-21
fix(replica-auto-balance): data locality rebuild loop

c3y1huang:
The rebuild loop happens when the data locality is trying to rebuild a new replica, and it got cleaned up by the auto-balance because of the sorted order.

Exchanging the order will let the data locality clean up its own redundant replica first.

https://github.com/longhorn/longhorn/issues/4761

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -866,11 +866,17 @@ func (vc *VolumeController) cleanupExtraHealthyReplicas(v *longhorn.Volume, e *l
 		return err
 	}
 
-	if cleaned, err = vc.cleanupAutoBalancedReplicas(v, e, rs); err != nil || cleaned {
+	if cleaned, err = vc.cleanupDataLocalityReplicas(v, e, rs); err != nil || cleaned {
 		return err
 	}
 
-	if cleaned, err = vc.cleanupDataLocalityReplicas(v, e, rs); err != nil || cleaned {
+	// Auto-balanced replicas get cleaned based on the sorted order. Hence,
+	// cleaning up for data locality replica needs to go first. Or else, a
+	// new replica could have a name alphabetically smaller than all the
+	// existing replicas. And this causes a rebuilding loop if this new
+	// replica is for data locality.
+	// Ref: https://github.com/longhorn/longhorn/issues/4761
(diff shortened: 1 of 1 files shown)


c3y1huang:
Test results:
- [test_replica_auto_balance_with_data_locality](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/2213/)
- [test_replica_auto_balance](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/2214/)

controller/volume_controller.go
```diff
@@ -866,11 +866,11 @@ func (vc *VolumeController) cleanupExtraHealthyReplicas(v *longhorn.Volume, e *l
...
 	}
 
-	if cleaned, err = vc.cleanupAutoBalancedReplicas(v, e, rs); err != nil || cleaned {
+	if cleaned, err = vc.cleanupDataLocalityReplicas(v, e, rs); err != nil || cleaned {
 		return err
 	}
 
-	if cleaned, err = vc.cleanupDataLocalityReplicas(v, e, rs); err != nil || cleaned {
+	if cleaned, err = vc.cleanupAutoBalancedReplicas(v, e, rs); err != nil || cleaned {
 		return err
```
derekbit, line 874: Why exchanging the order can fix the issue?

c3y1huang: The issue manifested because the data locality is trying to rebuild a new replica, and it got cleaned up by the auto-balance because of the sorted order. 

Exchanging the order will let the data locality clean up its own redundant replica.

c3y1huang: Updated the description.

derekbit: Got it, thank you.
Can we add a comment here for avoiding running into the issue when others work on it in the future?

c3y1huang: Sure, updated.

derekbit (approved):

innobead (approved):
This is tricky because `cleanupAutoBalancedReplicas` is quite natural w/o considering other scheduling factors like data locality, but only nodes are schedulable. 

So, basically, `cleanupAutoBalancedReplicas` should be always at the last cleanup step in `cleanupExtraHealthyReplicas`, @c3y1huang right?

PR #1530, opened 2022-10-25
fix: Custom default settings and values ​​will not be updated if they exist

weizhe0422:
[Longhorn 4554](https://github.com/longhorn/longhorn/issues/4554)

Singed-off-by: Ray Chang <ray.chang@suse.com>

### task items
 1. Customized default settings won't be update if setting and value exist
 2. Change the Customized default settings error message to warn instead, because it's not thrown out but just recognized 
 3. Disable ConfigMap re-synchronize periodically

### self test
Step 1. Install Longhorn, e.g, via Helm
 - 1.1 Clone Helm chart to local
 - 1.2 Add tolerance value in `values.yaml` like below:
    ```yaml
    defaultSettings:
      taintToleration: "kubevirt.io/drain:NoSchedule"
    longhornManager:
      tolerations:
      - key: "kubevirt.io/drain"
        operator: "Equal"
        value: ""
        effect: "NoSchedule"
    longhornDriver:
      tolerations:
      - key: "kubevirt.io/drain"
        operator: "Equal"
        value: ""
        effect: "NoSchedule"
    longhornUI:
      tolerations:
      - key: "kubevirt.io/drain"
        operator: "Equal"
        value: ""
        effect: "NoSchedule"
     ```
 - 1.3 Start the installation: `helm install longhorn . --namespace longhorn-system --create-space`

Step 2. Check if the setting was applied success
   - setting CR: `kubectl get settings.longhorn.io taint-toleration -n longhorn-system`
   - Instance-manager: `kubectl get pod instance-manager-r-047a4c1a -n longhorn-system -o yaml` -> `spec/tolerations`

Step 3. Create a volume and attach to a node

Step 4. Validate `Item 2`
   1. type `kubectl logs longhorn-manager-wzxt6 -n longhorn-system | grep "invalid customized default setting "`
   2. The output reveals warning message like below:  
    image

Step 5. Detach all volumes

Step 6. Update the tolerations value
   - Export current value: `kubectl get cm longhorn-default-setting -n longhorn-system -o yaml > default-setting-cm.yaml`
   - Modify the value and apply it: `kubectl apply -f default-setting-cm.yaml`

Step 7. Validate `Item 1` and `Item 3`
   - type `kubectl logs longhorn-manager-wzxt6 -n longhorn-system | grep "invalid customized default setting "`
   - check the message doesn't keep appearing every 30 seconds

### DIFF
--- a/datastore/longhorn.go
+++ b/datastore/longhorn.go
@@ -99,13 +99,34 @@ func (s *DataStore) filterCustomizedDefaultSettings(customizedDefaultSettings ma
 	availableCustomizedDefaultSettings := make(map[string]string)
 
 	for name, value := range customizedDefaultSettings {
-		if err := s.ValidateSetting(string(name), value); err == nil {
-			availableCustomizedDefaultSettings[name] = value
-		} else {
-			logrus.WithError(err).Errorf("invalid customized default setting %v with value %v, will continue applying other customized settings", name, value)
+		if !s.isSettingValueChanged(types.SettingName(name), value) {
+			continue
 		}
+
+		if err := s.ValidateSetting(string(name), value); err != nil {
+			logrus.WithError(err).Warnf("Invalid customized default setting %v with value %v, will continue applying other customized settings", name, value)
+			continue
+		}
+
+		availableCustomizedDefaultSettings[name] = value
 	}
 	return availableCustomizedDefaultSettings
+}
+
+// isSettingValueChanged check if the customized default setting and value was changed
+func (s *DataStore) isSettingValueChanged(name types.SettingName, value string) bool {
+	setting, err := s.GetSettingExact(name)
+	if err != nil {
+		if !ErrorIsNotFound(err) {
+			logrus.WithError(err).Errorf("failed to get customized default setting %v value", name)
+			return false
+		}
+		return true
+	}
+	if setting.Value == value {
+		return false
+	}
+	return true
 }
 
 func (s *DataStore) syncSettingsWithDefaultImages(defaultImages map[types.SettingName]string) error {
--- a/controller/kubernetes_configmap_controller.go
+++ b/controller/kubernetes_configmap_controller.go
@@ -68,11 +68,12 @@ func NewKubernetesConfigMapController(
 		eventRecorder: eventBroadcaster.NewRecorder(scheme, v1.EventSource{Component: "longhorn-kubernetes-configmap-controller"}),
 	}
 
-	ds.ConfigMapInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
-		AddFunc:    kc.enqueueConfigMapChange,
-		UpdateFunc: func(old, cur interface{}) { kc.enqueueConfigMapChange(cur) },
-		DeleteFunc: kc.enqueueConfigMapChange,
-	})
+	ds.ConfigMapInformer.AddEventHandlerWithResyncPeriod(
+		cache.ResourceEventHandlerFuncs{
+			AddFunc:    kc.enqueueConfigMapChange,
+			UpdateFunc: func(old, cur interface{}) { kc.enqueueConfigMapChange(cur) },
+			DeleteFunc: kc.enqueueConfigMapChange,
+		}, 0)
 
 	ds.StorageClassInformer.AddEventHandlerWithResyncPeriod(
 		cache.FilteringResourceEventHandler{


datastore/longhorn.go
```diff
@@ -99,15 +99,34 @@ func (s *DataStore) filterCustomizedDefaultSettings(customizedDefaultSettings ma
...
+	setting, err := s.GetSettingExact(name)
+	if err != nil {
+		if !ErrorIsNotFound(err) {
+			logrus.WithError(err).Errorf("failed to get customized default setting %v value", name)
+			return false
+		}
+		return true
+	}
+	if setting.Value == value {
+		logrus.Warnf("the customized default setting %v with value %v is exists", name, value)
```
derekbit, line 124: This is unnecessary.

datastore/longhorn.go
```diff
@@ -99,15 +99,34 @@ func (s *DataStore) filterCustomizedDefaultSettings(customizedDefaultSettings ma
...
+			} else {
+				logrus.WithError(err).Warnf("invalid customized default setting %v with value %v, will continue applying other customized settings", name, value)
+			}
 		}
 	}
 	return availableCustomizedDefaultSettings
 }
 
+// isSettingValueNotExists check if the customized default setting and value not exists
+func (s *DataStore) isSettingValueNotExists(name types.SettingName, value string) bool {
```
derekbit, line 114: I feel a bit of confused because the the behavior of the helper function and its name is not quite matching.
Suggest removing the comparison of values from the function or renaming to a better function name.

derekbit, line 114:
```suggestion
func (s *DataStore) isSettingValueChanged(name types.SettingName, value string) bool {
```

weizhe0422: Thank you. I will follow the suggestion to rename the function to avoid confusion, and also remove the redundant warning message.

derekbit:
In general, LGTM. Suggest running e2e test for checking the change of configmap resync period is safe.

datastore/longhorn.go
```diff
@@ -99,15 +99,33 @@ func (s *DataStore) filterCustomizedDefaultSettings(customizedDefaultSettings ma
...
-			availableCustomizedDefaultSettings[name] = value
-		} else {
-			logrus.WithError(err).Errorf("invalid customized default setting %v with value %v, will continue applying other customized settings", name, value)
+		if s.isSettingValueChanged(types.SettingName(name), value) {
+			if err := s.ValidateSetting(string(name), value); err == nil {
+				availableCustomizedDefaultSettings[name] = value
+			} else {
+				logrus.WithError(err).Warnf("invalid customized default setting %v with value %v, will continue applying other customized settings", name, value)
+			}
 		}
```
c3y1huang, line 108: Nit: suggest readjusting the if statements. IMO fewer layers can improve readability, ex:
```
	for name, value := range customizedDefaultSettings {
		if !s.isSettingValueChanged(types.SettingName(name), value) {
			continue
		}

		if err := s.ValidateSetting(string(name), value); err != nil {
			logrus.WithError(err).Warnf("invalid customized default setting %v with value %v, will continue applying other customized settings", name, value)
			continue
		}

		availableCustomizedDefaultSettings[name] = value
	}
```

datastore/longhorn.go
```diff
@@ -99,15 +99,33 @@ func (s *DataStore) filterCustomizedDefaultSettings(customizedDefaultSettings ma
...
 	for name, value := range customizedDefaultSettings {
-		if err := s.ValidateSetting(string(name), value); err == nil {
-			availableCustomizedDefaultSettings[name] = value
-		} else {
-			logrus.WithError(err).Errorf("invalid customized default setting %v with value %v, will continue applying other customized settings", name, value)
+		if s.isSettingValueChanged(types.SettingName(name), value) {
+			if err := s.ValidateSetting(string(name), value); err == nil {
+				availableCustomizedDefaultSettings[name] = value
+			} else {
+				logrus.WithError(err).Warnf("invalid customized default setting %v with value %v, will continue applying other customized settings", name, value)
```
c3y1huang, line 106: Any reason we are changing from error to warning?

datastore/longhorn.go
weizhe0422: Thanks for suggestion. I will fix it to improve readability.

datastore/longhorn.go
weizhe0422: It's because this message won't be returned to handle but recognized/record instead. 

So change the log level to warning.

c3y1huang, line 106:
```suggestion
				logrus.WithError(err).Warnf("Invalid customized default setting %v with value %v, will continue applying other customized settings", name, value)
```

c3y1huang (changes requested):

datastore/longhorn.go
weizhe0422: Done to fix it.

https://github.com/longhorn/longhorn-manager/blob/9c76065dbfc0e069824afaafc187b982c73bf79e/datastore/longhorn.go#L101-L112

c3y1huang (approved):

derekbit (approved):

innobead (approved):

datastore/longhorn.go
```diff
@@ -99,15 +99,36 @@ func (s *DataStore) filterCustomizedDefaultSettings(customizedDefaultSettings ma
...
 	}
 	return availableCustomizedDefaultSettings
 }
 
+// isSettingValueChanged check if the customized default setting and value was changed
+func (s *DataStore) isSettingValueChanged(name types.SettingName, value string) bool {
+	setting, err := s.GetSettingExact(name)
+	if err != nil {
+		if !ErrorIsNotFound(err) {
+			logrus.WithError(err).Errorf("failed to get customized default setting %v value", name)
```
innobead, line 121: nit: Message convention in general. Making that will be helpful for us for the following troubleshooting and debugging. 

A log message should start with the first capitalized word like "Failed to get customized default setting %v value". But if the error instance, the message inside should be not capitalized.

PR #1531, opened 2022-10-26
Allow customization of the cipher

mantissahz:
Allow customization of the cipher used by cryptsetup in volume encryption.
Now the parameters allowing for customization are `cipher`, `hash`, and `key-size`.

longhorn/longhorn#3353

### DIFF
--- a/csi/crypto/crypto.go
+++ b/csi/crypto/crypto.go
@@ -10,7 +10,53 @@ import (
 
 const (
 	mapperFilePathPrefix = "/dev/mapper"
+
+	CryptoKeyDefaultCipher = "aes-xts-plain64"
+	CryptoKeyDefaultHash   = "sha256"
+	CryptoKeyDefaultSize   = "256"
+	CryptoDefaultPBKDF     = "argon2i"
 )
+
+// EncryptParams keeps the customized cipher options from the secret CR
+type EncryptParams struct {
+	KeyProvider string
+	KeyCipher   string
+	KeyHash     string
+	KeySize     string
+	PBKDF       string
+}
+
+func NewEncryptParams(keyProvider, keyCipher, keyHash, keySize, pbkdf string) *EncryptParams {
+	return &EncryptParams{KeyProvider: keyProvider, KeyCipher: keyCipher, KeyHash: keyHash, KeySize: keySize, PBKDF: pbkdf}
+}
+
+func (cp *EncryptParams) GetKeyCipher() string {
+	if cp.KeyCipher == "" {
+		return CryptoKeyDefaultCipher
+	}
+	return cp.KeyCipher
+}
+
+func (cp *EncryptParams) GetKeyHash() string {
+	if cp.KeyHash == "" {
+		return CryptoKeyDefaultHash
+	}
+	return cp.KeyHash
+}
+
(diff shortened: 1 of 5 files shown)


joshimoo:
Can you add 2 example secret.yamls for
- pi optimized https://github.com/longhorn/longhorn/issues/3353
- fips compliant https://github.com/longhorn/longhorn/issues/4784

mantissahz:
>Can you add 2 example secret.yamls for
>- pi optimized https://github.com/longhorn/longhorn/issues/3353
>- fips compliant https://github.com/longhorn/longhorn/issues/4784
@joshimoo like secret.yaml examples in extended LEP?

joshimoo:
@mantissahz Like in this example folder. The example folder is synced to the /longhorn/longhorn repository.
https://github.com/longhorn/longhorn-manager/tree/master/examples/crypto

joshimoo (approved):
LGTM, It would be good to have the example secrets so that you and QA can test with the different key types and we know that we covered all the necessary fields/parameters.

joshimoo (approved):
LGTM, thanks for adding the additional examples and testing with them, this will help our users of this feature out quite a bit :)

joshimoo (approved):

csi/crypto/crypto.go
```diff
@@ -10,17 +10,63 @@ import (
...
+// EncryptParams keeps the customized cipher options from the secret CR
+type EncryptParams struct {
+	KeyProvider string
+	KeyCipher   string
+	KeyHash     string
+	KeySize     string
+	PBKDF       string
+}
+
+func NewEncryptParams(keyProvider, keyCipher, keyHash, keySize, pbkdf string) *EncryptParams {
```
innobead, line 29: Suggest having a unit test to test the default values.

csi/crypto/crypto.go
```diff
@@ -10,17 +10,63 @@ import (
 
 const (
 	mapperFilePathPrefix = "/dev/mapper"
+
+	CryptoKeyDefaultCipher = "aes-xts-plain64"
+	CryptoKeyDefaultHash   = "sha256"
+	CryptoKeyDefaultSize   = "256"
+	CryptoDefaultPBKDF     = "argon2i"
```
innobead, line 17: Assume these value the same as before to ensure the backward compatibility, right?

innobead (commented):
In general, LGTM. Few comments.

csi/crypto/crypto.go
mantissahz: Yes, These are default values or compatible on most distros.
only Debian 9 dose not support option "--pbdkf" because the cryptsetup version (1.7.4,  >2.0 needed)

joshimoo: The encryption parameters are only required for initial encryption.

NIT: One could also leave the defaults blank then check whether the argument should be added to the call to luksFormat.
This way one can get rid of the Getter functions and one doesn't necessarily need the constructor function either.

But it's nit picking and the PR was good enough to get merged and testing started :)

joshimoo: @mantissahz see comment above in regard to the removal of the defaults, to delegate to the luksformat call, that should solve your debian issue :)

PR #1533, opened 2022-10-31
old IM-r Pods are not deleted

mantissahz:
Replicas are waiting for that IMs starts and the volume controller consider IMs are failed and then set replicas failed.

Not to set replicas failed when volume is upgrading.

longhorn/longhorn#4726

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -762,7 +762,8 @@ func (vc *VolumeController) cleanupReplicas(v *longhorn.Volume, es map[string]*l
 	// 	since the getHealthyReplicaCount function doesn't differentiate between replicas of different engines
 	// 	then during cleanupExtraHealthyReplicas the condition `healthyCount > v.Spec.NumberOfReplicas` will be true
 	//  which can lead to incorrect deletion of replicas.
-	if vc.isVolumeMigrating(v) || vc.isVolumeUpgrading(v) {
+	//  Allow to delete replicas in `cleanupCorruptedOrStaleReplicas` marked as failed before IM-r started during engine image update.
+	if vc.isVolumeMigrating(v) {
 		return nil
 	}
 
@@ -773,6 +774,11 @@ func (vc *VolumeController) cleanupReplicas(v *longhorn.Volume, es map[string]*l
 
 	if err := vc.cleanupCorruptedOrStaleReplicas(v, rs); err != nil {
 		return err
+	}
+
+	// give a chance to delete new replicas failed when upgrading volume and waiting for IM-r starting
+	if vc.isVolumeUpgrading(v) {
+		return nil
 	}
 
 	if err := vc.cleanupFailedToScheduledReplicas(v, rs); err != nil {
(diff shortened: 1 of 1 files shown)


shuo-wu (commented):
I prefer to fix this issue with the following methods:
1. Avoid setting the instance manager name for the (upgrading/new) replicas when the new IM is not starting or has empty status fields.
2. Adding the failed replica cleanup and retry mechanism for the upgrading replicas when marked as failed.

controller/volume_controller.go
```diff
@@ -452,13 +452,46 @@ func (vc *VolumeController) syncVolume(key string) (err error) {
 		return err
 	}
 
+	if err := vc.reuseFailedUpgradingReplica(volume, replicas); err != nil {
```
shuo-wu, line 455: It's better not to add an extra replica scheduling-related function. It makes the scheduling flow inside volume-controller become more complicated...
For the **failed** new replicas during the upgrade, I am fine with directly cleaning them up then starting new replicas...

shuo-wu (commented):
Besides, can we postpone the replica launching or the instance manager name update if the instance manager is not ready?

controller/volume_controller.go
mantissahz: OK, Got it.
Removed, and try to postpone the replica launching.

controller/volume_controller.go
```diff
@@ -762,7 +762,7 @@ func (vc *VolumeController) cleanupReplicas(v *longhorn.Volume, es map[string]*l
 	// 	since the getHealthyReplicaCount function doesn't differentiate between replicas of different engines
 	// 	then during cleanupExtraHealthyReplicas the condition `healthyCount > v.Spec.NumberOfReplicas` will be true
 	//  which can lead to incorrect deletion of replicas.
-	if vc.isVolumeMigrating(v) || vc.isVolumeUpgrading(v) {
+	if vc.isVolumeMigrating(v) {
```
shuo-wu, line 765: NIT: Need to modify the above comment a little bit. Besides, will executing `cleanupExtraHealthyReplicas` for upgrading volumes lead to any issues?

controller/volume_controller.go
```diff
@@ -1493,6 +1499,9 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
 			if r.Spec.NodeID == "" {
 				continue
 			}
+			if vc.isVolumeUpgrading(v) && r.Status.InstanceStatus.CurrentState != longhorn.InstanceStateRunning {
+				continue
+			}
```
shuo-wu, line 1504: Shouldn't it check if the instance-manager-r is starting or running instead?

controller/volume_controller.go
mantissahz: Add some comment to explain it.
I am afraid of `DataLocality` and `AutoBalanced` might lead some issues potentaily in `cleanupExtraHealthyReplicas`

controller/volume_controller.go
mantissahz: Modified, but I think `InstanceStatus.CurrentState` only `running` might be safer to schedule them to engine `replicaAddressMap`.

shuo-wu: What I mean is, the below `canInstanceManagerLaunchReplica` should return false if the related instance-manager-r is not running or starting.

mantissahz: Got it and Modified. Wait for IM-r is starting when `canInstanceManagerLaunchReplica` returns `false`.

controller/volume_controller.go
shuo-wu: Yes. Please test these features with the PR. Besides, you may need to check the live migration feature as well.

shuo-wu (approved):
LGTM. Waiting for the test result.

mantissahz:
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/2392/#showFailuresLink
Here is the regression test result.

-  `test_autosalvage_with_data_locality_enabled` is a flaky test case, QA is fixing now.

Others tested `PASSED` in my local environment.
~~~text
============================= test session starts ==============================
platform linux -- Python 3.9.14, pytest-5.3.1, py-1.11.0, pluggy-0.13.1 -- /usr/bin/python3.9
cachedir: .pytest_cache
plugins: order-1.0.1, repeat-0.9.1
collecting ... collected 6 items

test_recurring_job.py::test_recurring_job_in_storageclass[s3] PASSED     [ 16%]
test_recurring_job.py::test_recurring_job_in_storageclass[nfs] PASSED    [ 33%]
test_scheduling.py::test_replica_schedule_to_disk_with_most_usable_storage PASSED [ 50%]
test_backing_image.py::test_exporting_backing_image_from_volume PASSED   [ 66%]
test_cloning.py::test_cloning_with_detached_source_volume PASSED         [ 83%]
test_node.py::test_node_eviction_multiple_volume PASSED                  [100%]

======================== 6 passed in 1375.39s (0:22:55) ========================
~~~

innobead (approved):

controller/volume_controller.go
innobead: `cleanupExtraHealthyReplicas` is the last statement in `cleanupReplicas`, so theriotically, it should be fine, because no healthy replicas will be impacted during the volume upgrade. Need to follow up with QA later, cc @longhorn/qa

PR #1534, opened 2022-11-01
Handle unscheduled failed replica

mantissahz:
Set the replica RebuildRetryCount to FailedReplicaMaxRetryCount when the replica is considered as a rebuilding replica and marked as failed.
It would be deleted soon in func `cleanupCorruptedOrStaleReplicas'

longhorn/longhorn#4791
related: longhorn/longhorn#4724

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -1400,6 +1400,11 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
 				if r.Spec.HealthyAt == "" && r.Spec.FailedAt == "" {
 					r.Spec.FailedAt = vc.nowHandler()
 					r.Spec.DesireState = longhorn.InstanceStateStopped
+					// unscheduled replicas marked failed here when volume detached
+					// check if NodeId or DiskID is empty to avoid deleting reusableFailedReplica when replenished.
+					if r.Spec.NodeID == "" || r.Spec.DiskID == "" {
+						r.Spec.RebuildRetryCount = scheduler.FailedReplicaMaxRetryCount
+					}
 					rs[r.Name] = r
 				}
 			}


controller/volume_controller.go
```diff
@@ -1400,6 +1400,7 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
 				if r.Spec.HealthyAt == "" && r.Spec.FailedAt == "" {
 					r.Spec.FailedAt = vc.nowHandler()
 					r.Spec.DesireState = longhorn.InstanceStateStopped
+					r.Spec.RebuildRetryCount = scheduler.FailedReplicaMaxRetryCount
```
shuo-wu, line 1403: The reused failed replicas will have a similar status `r.Spec.HealthyAt == "" && r.Spec.FailedAt == ""` as well... In other words, this PR will clean up these replicas
https://github.com/longhorn/longhorn-manager/blob/cbcd5c7beab27237f2761b3f74d8fb4167e981da/controller/volume_controller.go#L1721-L1723

mantissahz: Check if disk is empty or nodeID is empty then set `RebuildRetryCount` to clean up
@shuo-wu WDYT?

derekbit: > After commit 15f850c, although node 2 scheduling enabled, the 2 failed replicas in previous step remain unscheduled (N/A), and will not be reused to restore data, the volume has to create a new replica to schedule to node 2, so the volume will have 2 scheduled replicas and 2 failed replicas from the previous step, so assertion failed.

The real issue seems how to reuse the failed replicas rather than delete failed replicas. Right?

mantissahz: These failed replicas is not used before, so I think it could be deleted and wait for replenishing when volume attached.
Added `check if NodeID or DiskId is empty`

shuo-wu: Agree with Derek's comment.
You can consider the following besides the current fix:
1. The scheduler reusage should ignore the unscheduled failed replicas.
2. The unscheduled failed replicas should be directly considered as corrupted and stale replicas.

mantissahz: >1. The scheduler reusage should ignore the unscheduled failed replicas.

Do you mean `reused` ? We did `ignore` unscheduled failed replicas (not reused and not deleted) now.

>2. The unscheduled failed replicas should be directly considered as corrupted and stale replicas.

Unscheduled replicas would be marked as failed when volume is detached and set `RebuildRetryCount` would make unscheduled failed replicas considered as corrupted and stale replicas.

shuo-wu (approved):

controller/volume_controller.go
shuo-wu: Yes. I just checked the code. This fix looks good enough.

derekbit (approved):
LGTM after discussing with @mantissahz

innobead (approved):

innobead:
@mantissahz Please backport to 1.3 and 1.2. Thanks.

PR #1538, opened 2022-11-02
Filesystem Trim Support 

shuo-wu:
https://github.com/longhorn/longhorn/issues/836


shuo-wu:
This PR is ready for review but we need to wait for the dependencies merge first.

controller/volume_controller.go
```diff
@@ -1015,6 +1019,52 @@ func (vc *VolumeController) getAutoBalancedReplicasSetting(v *longhorn.Volume) (
...
+	}
+
+	unmapMarkEnabled, err := vc.getUnmapMarkSnapChainRemovedSetting(v)
+	if err != nil {
+		return err
+	}
+
+	if es != nil {
+		for _, e := range es {
+			e.Spec.UnmapMarkSnapChainRemovedEnabled = unmapMarkEnabled
```
c3y1huang, line 1065: Break after this?

types/setting.go
```diff
@@ -858,6 +861,23 @@ var (
...
 	}
+
+	SettingDefinitionUnmapMarkSnapChainRemoved = SettingDefinition{
+		DisplayName: "Unmap Mark Snapshot Chain as Removed",
+		Description: "This setting allows Longhorn filesystem trim feature to automatically mark the latest snapshot and its ancestors as removed and stops at the snapshot containing multiple children.\n\n" +
+			"Since Longhorn filesystem trim feature can be applied to the volume head and the followed continuous removed or system snapshots only.\n\n" +
+			"Notice that trying to trim/unmap a removed files from a valid snapshot will do nothing but the filesystem will discard this kind of in-memory trimmable file info. " +
+			"Later on if you mark the snapshot as removed and want to retry the trim, you may need to unmount and remount the filesystem so that the filesystem can recollect the trimmable file info.",
+		Category: SettingCategoryGeneral,
+		Type:     SettingTypeString,
```
c3y1huang, line 872: Why not use boolean?

util/util.go
```diff
@@ -895,3 +895,29 @@ func GetVolumeMeta(path string) (*VolumeMeta, error) {
...
+		return err
+	}
+
+	mountOutput, err := nsExec.Execute("bash", []string{"-c", fmt.Sprintf("mount | grep /dev/longhorn/%s | awk '{print $3}'", name)})
+	if err != nil {
+		return fmt.Errorf("cannot find volume %v mount info on host: %v", name, err)
+	}
+
+	mountList := strings.Split(strings.TrimSpace(mountOutput), "\n")
+	logrus.Infof("trim debug1: %v", mountOutput)
```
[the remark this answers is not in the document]
shuo-wu: sorry for missing this debug log

controller/volume_controller.go
shuo-wu: We will blindly update this field for all engines and replicas.

controller/volume_controller.go
```diff
@@ -1015,6 +1019,52 @@ func (vc *VolumeController) getAutoBalancedReplicasSetting(v *longhorn.Volume) (
...
+		}
+	}
+	if rs != nil {
+		for _, r := range rs {
+			r.Spec.UnmapMarkDiskChainRemovedEnabled = unmapMarkEnabled
```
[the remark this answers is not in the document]
shuo-wu: All replicas should have the same value. Cannot break here.

types/setting.go
```diff
@@ -858,6 +861,23 @@ var (
 		ReadOnly: false,
 		Default:  "false",
 	}
+
+	SettingDefinitionUnmapMarkSnapChainRemoved = SettingDefinition{
+		DisplayName: "Unmap Mark Snapshot Chain as Removed",
```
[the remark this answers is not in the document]
shuo-wu: This determines the extra operations during the filesystem trim rather than if the trim is allowed. How about `Remove Snapshots during Filesystem Trim`?

types/setting.go
shuo-wu: Nice catch. It's better to use bool here.

deploy/install/01-prerequisite/06-storageclass.yaml
```diff
@@ -22,6 +22,7 @@ data:
     #  backingImageDataSourceType: "download"
     #  backingImageDataSourceParameters: '{"url": "https://backing-image-example.s3-region.amazonaws.com/test-backing-image"}'
     #  backingImageChecksum: "SHA512 checksum of the backing image"
+    #  unmapMarkSnapChainRemoved: ignored
```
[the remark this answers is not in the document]
shuo-wu: No. Thank you for pointing out

util/util.go
```diff
@@ -895,3 +895,28 @@ func GetVolumeMeta(path string) (*VolumeMeta, error) {
...
+	if err != nil {
+		return fmt.Errorf("cannot find volume %v mount info on host: %v", name, err)
+	}
+
+	mountList := strings.Split(strings.TrimSpace(mountOutput), "\n")
+	if len(mountList) != 1 {
+		return fmt.Errorf("zero or multiple mount points for volume %s: %+v", name, mountList)
+	}
+
+	_, err = nsExec.Execute("fstrim", mountList)
```
derekbit, line 916: Just curious. `fstrim` is a blocking call or a non-blocking call?

k8s/pkg/apis/longhorn/v1beta2/volume.go
```diff
@@ -67,6 +67,15 @@ const (
 	ReplicaAutoBalanceBestEffort  = ReplicaAutoBalance("best-effort")
 )
 
+// +kubebuilder:validation:Enum="";disabled;enabled
+type UnmapMarkSnapChainRemoved string
+
+const (
+	UnmapMarkSnapChainRemovedEmpty    = UnmapMarkSnapChainRemoved("")
```
derekbit, line 74: Should we need the empty value? How about setting the empty value to `UnmapMarkSnapChainRemovedDisabled` during upgrade?

engineapi/instance_manager.go
```diff
@@ -216,6 +220,9 @@ func (c *InstanceManagerClient) ReplicaProcessCreate(replicaName, engineImage, d
 	if revCounterDisabled {
 		args = append(args, "--disableRevCounter")
 	}
+	if enableUnmapMarkDiskChainRemoved {
+		args = append(args, "--unmap-mark-disk-chain-removed")
+	}
```
derekbit, line 225: Need to check the cli version, or the volume with an old engine cannnot be started because of the `--unmap-mark-disk-chain-removed`. missing error of.

k8s/pkg/apis/longhorn/v1beta2/volume.go
shuo-wu: This will be updated `UnmapMarkSnapChainRemoved("ignored")`.

util/util.go
shuo-wu: I think it's a blocking call. I need to have a test (by deleting tons of small files) for this part.

shuo-wu: I tried to create 50000 4MB files (200G in total) in a 500G volume then delete and trim them, `fstrim` gets stuck for around 2 seconds. I don't think there is a timeout risk.

innobead:
@shuo-wu Is this ready for review?

shuo-wu:
Yes. We need to wait for the engine im PRs before merging this.

datastore/longhorn.go
```diff
@@ -792,6 +792,35 @@ func (s *DataStore) ListVolumesByLabelSelector(selector labels.Selector) (map[st
...
+	if err != nil {
+		return nil, err
+	}
+
+	list, err := s.ListVolumesBySelectorRO(selectors)
+	if err != nil {
+		return nil, err
+	}
+	for _, itemRO := range list {
+		if itemRO.Spec.Standby {
```
innobead, line 816: Why filtering DR volumes only?

datastore/longhorn.go
```diff
@@ -792,6 +792,35 @@ func (s *DataStore) ListVolumesByLabelSelector(selector labels.Selector) (map[st
...
 }
 
+// ListVolumesFollowsGlobalSettingsRO returns an object contains all Volumes
+// that don't apply specific spec and follow the global settings
+func (s *DataStore) ListVolumesFollowsGlobalSettingsRO(followedSettingNames map[string]interface{}) (map[string]*longhorn.Volume, error) {
+	itemMap := make(map[string]*longhorn.Volume)
+
+	followedGlobalSettingsLabels := map[string]string{}
+	for sName := range followedSettingNames {
+		followedGlobalSettingsLabels[types.GetLonghornLabelKey(types.LonghornLabelFollowGlobalSetting)] = sName
```
innobead, line 802: Where/when has `LonghornLabelFollowGlobalSetting`  been updated to volumes?

innobead (commented):
In general, LGTM, but have question about `ListVolumesFollowsGlobalSettingsRO`.

datastore/longhorn.go
shuo-wu: This is typo. Removed the check.

datastore/longhorn.go
shuo-wu: Missed the changes in the mutation webhook part.

shuo-wu:
> In general, LGTM, but have question about `ListVolumesFollowsGlobalSettingsRO`.

I just update the commit `VolumeController enqueue Setting RemoveSnapshotsDuringFilesystemTrim` for this issue. PTAL.

util/util.go
```diff
@@ -906,3 +906,28 @@ func GetPodIP(pod *v1.Pod) (string, error) {
 	}
 	return pod.Status.PodIP, nil
 }
+
+func TrimFilesystem(name string) error {
```
innobead, line 910: nit: Suggest having a meaning parameter name like `volumeName`

go.mod
```diff
@@ -3,6 +3,9 @@ module github.com/longhorn/longhorn-manager
 go 1.17
 
 replace (
+	github.com/longhorn/longhorn-engine => github.com/shuo-wu/longhorn-engine v0.4.1-0.20221122040221-d44f241ef22e
+	github.com/longhorn/longhorn-instance-manager => github.com/shuo-wu/longhorn-instance-manager v0.0.0-20221122044248-bf826728b669
```
innobead, line 7: Wrong dependency

k8s/crds.yaml
```diff
@@ -1361,6 +1361,8 @@ spec:
                 type: boolean
               salvageRequested:
                 type: boolean
+              unmapMarkSnapChainRemovedEnabled:
```
innobead, line 1364: Why there is no `UnmapMarkSnapChainRemovedEnabled` in the engine status?

webhook/resources/volume/mutator.go
```diff
@@ -106,6 +106,10 @@ func (v *volumeMutator) Create(request *admission.Request, newObj runtime.Object
 		patchOps = append(patchOps, fmt.Sprintf(`{"op": "replace", "path": "/spec/restoreVolumeRecurringJob", "value": "%s"}`, longhorn.RestoreVolumeRecurringJobDefault))
 	}
 
+	if volume.Spec.UnmapMarkSnapChainRemoved == "" {
+		patchOps = append(patchOps, `{"op": "replace", "path": "/spec/unmapMarkSnapChainRemoved", "value": "ignored"}`)
```
innobead, line 110: Can use `UnmapMarkSnapChainRemovedIgnored` from v1beta2 volume instead of text `ignored` here

webhook/resources/volume/mutator.go
```diff
@@ -234,6 +238,9 @@ func (v *volumeMutator) Update(request *admission.Request, oldObj runtime.Object
 			patchOps = append(patchOps, fmt.Sprintf(`{"op": "replace", "path": "/spec/recurringJobs/%d/labels", "value": {}}`, id))
 		}
 	}
+	if volume.Spec.UnmapMarkSnapChainRemoved == "" {
+		patchOps = append(patchOps, `{"op": "replace", "path": "/spec/unmapMarkSnapChainRemoved", "value": "ignored"}`)
```
innobead, line 247: Use `UnmapMarkSnapChainRemovedIgnored` instead of `ignored`

engineapi/instance_manager.go
```diff
@@ -211,6 +215,9 @@ func (c *InstanceManagerClient) ReplicaProcessCreate(replica *longhorn.Replica,
 	if replica.Spec.RevisionCounterDisabled {
 		args = append(args, "--disableRevCounter")
 	}
+	if replica.Spec.UnmapMarkDiskChainRemovedEnabled {
```
innobead, line 218: This part needs to rebase to do backward compatibility support

innobead (changes requested):

k8s/crds.yaml
shuo-wu: Sorry. I just manually updated the CRD in repo `longhorn` and forgot to update this.

engineapi/instance_manager.go
shuo-wu: It seems that we need to check the backward compatibility for the below option `dataLocality == longhorn.DataLocalityStrictLocal` as well. I will directly include it in this commit.

k8s/crds.yaml
innobead: I see, but next time you can use https://github.com/longhorn/longhorn-manager/blob/master/k8s/generate_code.sh#L70 to update the crds.yaml directly.

controller/volume_controller.go
```diff
@@ -1037,6 +1041,34 @@ func (vc *VolumeController) updateReplicaLogRequested(e *longhorn.Engine, rs map
 	}
 }
 
+func (vc *VolumeController) getUnmapMarkSnapChainRemovedSetting(v *longhorn.Volume) (bool, error) {
```
innobead, line 1053: Suggest renaming to `isUnmapMarkSnapChainRemovedEnabled`

webhook/resources/volume/mutator.go
```diff
@@ -180,7 +184,12 @@ func (v *volumeMutator) Create(request *admission.Request, newObj runtime.Object
 		labels[types.LonghornLabelBackupVolume] = bvName
 	}
 
-	patchOp, err := common.GetLonghornLabelsPatchOp(volume, labels)
+	labelsForVolumesFollowsGlobalSettings := datastore.GetLabelsForVolumesFollowsGlobalSettings(volume)
```
innobead, line 187: `SettingNameReplicaAutoBalance` is before 1.4, so do we need to do the upgrade migration?

innobead (changes requested):
LGTM, except a few comments needed to resolve.

types/types.go
```diff
@@ -199,6 +202,14 @@ const (
 	ImageChecksumNameLength = 8
 )
 
+// SettingsRelatedToVolume should match the items in datastore.GetLabelsForVolumesFollowsGlobalSettings
+//   TODO: May need to add the data locality check
+var SettingsRelatedToVolume = map[string]string{
```
innobead, line 207: This is good.

webhook/resources/volume/mutator.go
shuo-wu: I didn't add an upgrade path for this. This mutator blindly checks the labels during update/creation.
And this labeling makes Longhorn update the volume/engine fields in time when the related settings are modified.

innobead: I see, it's OK.

innobead (approved):

PR #1539, opened 2022-11-07
fix(node): duplicated event log when node restarted

mantissahz:
Create the event log for node restart only when `UpdateNodeStatus` without then `conflict` error

longhorn/longhorn#1449

### DIFF
--- a/controller/node_controller.go
+++ b/controller/node_controller.go
@@ -286,17 +286,26 @@ func (nc *NodeController) syncNode(key string) (err error) {
 		return nc.ds.RemoveFinalizerForNode(node)
 	}
 
+	nodeIsReadyEventMsg := ""
+	nodeIsReadyEventType := ""
 	existingNode := node.DeepCopy()
 	defer func() {
 		// we're going to update volume assume things changes
 		if err == nil && !reflect.DeepEqual(existingNode.Status, node.Status) {
 			_, err = nc.ds.UpdateNodeStatus(node)
 		}
-		// requeue if it's conflict
-		if apierrors.IsConflict(errors.Cause(err)) {
-			logrus.Debugf("Requeue %v due to conflict: %v", key, err)
-			nc.enqueueNode(node)
-			err = nil
+		if err != nil {
+			// requeue if it's conflict
+			if apierrors.IsConflict(errors.Cause(err)) {
+				logrus.Debugf("Requeue %v due to conflict: %v", key, err)
+				nc.enqueueNode(node)
+				err = nil
+			} else {
+				logrus.WithError(err).Errorf("failed to update node %v status", node.Name)
+			}
+		} else if nodeIsReadyEventMsg != "" && nodeIsReadyEventType != "" {
+			// create the event when the status type `Ready` of node is changed
+			nc.eventRecorder.Event(node, nodeIsReadyEventType, longhorn.NodeConditionTypeReady, nodeIsReadyEventMsg)
 		}
 	}()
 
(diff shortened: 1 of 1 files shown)


controller/node_controller.go
```diff
@@ -313,16 +321,17 @@ func (nc *NodeController) syncNode(key string) (err error) {
 			for _, podCondition := range podConditions {
 				if podCondition.Type == v1.PodReady {
 					if podCondition.Status == v1.ConditionTrue && pod.Status.Phase == v1.PodRunning {
-						node.Status.Conditions = types.SetConditionAndRecord(node.Status.Conditions,
-							longhorn.NodeConditionTypeReady, longhorn.ConditionStatusTrue,
-							"", fmt.Sprintf("Node %v is ready", node.Name),
-							nc.eventRecorder, node, v1.EventTypeNormal)
+						// create the event after `UpdateNodeStatus` without the error
+						nodeIsReadyEventMsg = fmt.Sprintf("Node %v is ready", node.Name)
+						nodeIsReadyEventType = v1.EventTypeNormal
```
shuo-wu, line 326: NIT: You can set Msg and Type only when there is a change. Then you don't need `nodeIsReadyEventChanged`

mantissahz: Modified. Removed `nodeIsReadyEventChanged`. Thanks.

controller/node_controller.go
```diff
@@ -297,6 +299,11 @@ func (nc *NodeController) syncNode(key string) (err error) {
 			logrus.Debugf("Requeue %v due to conflict: %v", key, err)
 			nc.enqueueNode(node)
 			err = nil
+		} else {
+			if err == nil && nodeIsReadyEventMsg != "" && nodeIsReadyEventType != "" {
+				// create the event when the status type `Ready` of node is changed
+				nc.eventRecorder.Event(node, nodeIsReadyEventType, longhorn.NodeConditionTypeReady, nodeIsReadyEventMsg)
+			}
```
shuo-wu, line 306: NIT: you can refactor this defer function a little bit:
```
if err == nil {
	if !reflect.DeepEqual(existingNode.Status, node.Status) {
		_, err = nc.ds.UpdateNodeStatus(node)
	}
	if nodeIsReadyEventMsg != "" && nodeIsReadyEventType != "" {
		nc.eventRecorder.Event(node, nodeIsReadyEventType, longhorn.NodeConditionTypeReady, nodeIsReadyEventMsg)
	}
} else {
	if apierrors.IsConflict(errors.Cause(err)) {
		logrus.Debugf("Requeue %v due to conflict: %v", key, err)
		nc.enqueueNode(node)
		err = nil
	}
}
```

shuo-wu (commented):
BTW. there is a typo in the commit & PR title.

shuo-wu (dismissed):

mantissahz:
>BTW. there is a typo in the commit & PR title.

Modified the typo of the title. Thanks.

shuo-wu (dismissed):

controller/node_controller.go
```diff
@@ -286,17 +286,26 @@ func (nc *NodeController) syncNode(key string) (err error) {
...
-			err = nil
+		if err != nil {
+			// requeue if it's conflict
+			if apierrors.IsConflict(errors.Cause(err)) {
+				logrus.Debugf("Requeue %v due to conflict: %v", key, err)
+				nc.enqueueNode(node)
+				err = nil
+			}
+		} else {
+			if nodeIsReadyEventMsg != "" && nodeIsReadyEventType != "" {
```
c3y1huang, line 305:
```suggestion
		} else if nodeIsReadyEventMsg != "" && nodeIsReadyEventType != "" {
```

controller/node_controller.go
```diff
@@ -286,17 +286,26 @@ func (nc *NodeController) syncNode(key string) (err error) {
...
-			logrus.Debugf("Requeue %v due to conflict: %v", key, err)
-			nc.enqueueNode(node)
-			err = nil
+		if err != nil {
+			// requeue if it's conflict
+			if apierrors.IsConflict(errors.Cause(err)) {
+				logrus.Debugf("Requeue %v due to conflict: %v", key, err)
+				nc.enqueueNode(node)
+				err = nil
+			}
```
c3y1huang, line 303: Should we log the error when it is not a conflict error?

c3y1huang (changes requested):

controller/node_controller.go
mantissahz: Yes, added log to record other errors.

controller/node_controller.go
mantissahz: Modified. Thanks.

c3y1huang (approved):

innobead (approved):

PR #1552, opened 2022-11-16
controller: Avoid wrongly updating the node and disk ID for BIDS

shuo-wu:
Longhorn/longhorn#4887


controller/backing_image_controller.go
```diff
@@ -503,37 +485,26 @@ func (bic *BackingImageController) handleBackingImageDataSource(bi *longhorn.Bac
...
-					bids.Spec.DiskUUID = diskStatus.DiskUUID
-					bids.Spec.DiskPath = diskSpec.Path
-					bids.Spec.NodeID = node.Name
-					updated = true
-				}
-				if updated {
-					break
-				}
-			}
+		node, diskName, err := bic.ds.GetReadyDiskNode(bids.Spec.DiskUUID)
```
innobead, line 497: Why not just use `bids.Status.CurrentState`and `bids.Spec.FileTransferred` together to decide if need to update the `bids.Spec.NodeID/Disk..`?

shuo-wu: Double-checked the logic, we don't need to validate `bids.Status.CurrentState`. See the comment in the code for details.

innobead: Yeah, this can be improved later. If needed, create a ticket for this.

datastore/longhorn.go
```diff
@@ -2197,6 +2198,43 @@ func (s *DataStore) GetRandomReadyNode() (*longhorn.Node, error) {
 	return nil, fmt.Errorf("unable to get a ready node")
 }
 
+// GetRandomReadyNodeDisk a list of all Node the in the given namespace and
+// returns the first Node && the first Disk of the Node marked with condition ready and allow scheduling
+func (s *DataStore) GetRandomReadyNodeDisk() (*longhorn.Node, string, error) {
+	logrus.Debugf("Prepare to find a random ready node disk")
```
innobead, line 2204: nit: suggest changing the log to `Preparing to find a random ready node disk`

innobead (dismissed):
LGTM.

datastore/longhorn.go
innobead, line 2204:
```suggestion
	logrus.Debugf("Preparing to find a random ready node disk")
```

innobead: I updated this for you @shuo-wu

PR #1558, opened 2022-11-18
feat: make FS format options configurable in storage class

weizhe0422:
[Longhorn 4642](https://github.com/longhorn/longhorn/issues/4642)

### DIFF
--- a/csi/node_server.go
+++ b/csi/node_server.go
@@ -101,7 +101,7 @@ func (ns *NodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 		return nil, status.Errorf(codes.NotFound, "volume %s not found", volumeID)
 	}
 
-	mounter, err := ns.getMounter(volume, volumeCapability)
+	mounter, err := ns.getMounter(volume, volumeCapability, req.VolumeContext)
 	if err != nil {
 		return nil, status.Error(codes.Internal, err.Error())
 	}
@@ -324,7 +324,7 @@ func (ns *NodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
 		return nil, status.Errorf(codes.NotFound, "volume %s not found", volumeID)
 	}
 
-	mounter, err := ns.getMounter(volume, volumeCapability)
+	mounter, err := ns.getMounter(volume, volumeCapability, req.VolumeContext)
 	if err != nil {
 		return nil, status.Error(codes.Internal, err.Error())
 	}
@@ -623,7 +623,7 @@ func getNodeServiceCapabilities(cs []csi.NodeServiceCapability_RPC_Type) []*csi.
 	return nscs
 }
 
-func (ns *NodeServer) getMounter(volume *longhornclient.Volume, volumeCapability *csi.VolumeCapability) (mount.Interface, error) {
+func (ns *NodeServer) getMounter(volume *longhornclient.Volume, volumeCapability *csi.VolumeCapability, volumeContext map[string]string) (mount.Interface, error) {
 	if volumeCapability.GetBlock() != nil {
 		return mount.New(""), nil
 	}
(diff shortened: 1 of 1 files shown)


innobead (commented):
Is this ready to review?

weizhe0422:
Yes, this PR is ready for review.

I try to specify different parameters for ext4 in storage class, then deploy a pod and PVC. 

The Longhorn volume could be created with specified parameter normally.

image

csi/node_server.go
```diff
@@ -647,6 +647,11 @@ func (ns *NodeServer) getMounter(volume *longhornclient.Volume, volumeCapability
 			params += fsParams.formatParameters
 		}
 
+		// If the user specifies parameters in the storage class, the parameters are overridden.
+		if mkfsParams, ok := volumeContext["mkfsParams"]; ok && mkfsParams != "" {
+			params = mkfsParams
```
innobead, line 652: Should append instead of overwriting, because some default params of the supported fs are mandatory. 

Also, because of https://github.com/longhorn/longhorn/issues/4594 introduced, we only allow using the fixed block size (4096) which is defined in `supportedFs`, so we should filter it if it is specified in the user provided `mkfsParams`.

innobead (changes requested):
In general, LGTM. One comment needs to resolve.

csi/node_server.go
innobead: Of course, we could allow users to do that w/o any filtering or validation and they need to be responsible for that, but probably there would be some concerns later.

@derekbit @PhanLe1010 @weizhe0422  WDYT?

derekbit: I think `append` is better. Then, the later one can override the previous value.

deploy/install/01-prerequisite/06-storageclass.yaml
```diff
@@ -15,6 +15,7 @@ data:
     volumeBindingMode: Immediate
     parameters:
       fsType: "ext4"
+      mkfsParams: "-b4096"
```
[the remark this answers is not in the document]
derekbit: e.g. `mkfs.ext4 -I 512 -b 2048 -O ^metadata_csum,^64bit`

```
-I: inode size
-b: block size
-O ^metadata_csum: disable metadata checksum
-O ^64bit: disable 64 bit
```

csi/node_server.go
innobead: yeah. @derekbit do you have any concerns if users use non-4096 block size which is what we fixed/improved in https://github.com/longhorn/longhorn/issues/4594.

derekbit: No, only performance drop if use a smaller block size. We can mention it in the doc.

deploy/install/01-prerequisite/06-storageclass.yaml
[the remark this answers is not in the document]
weizhe0422: @derekbit Thank you, I will follow the suggestion to modify it.

csi/node_server.go
weizhe0422: Got it, I will modify it as 'params+=mkfsParams' to append user specified parameters.

deploy/install/01-prerequisite/06-storageclass.yaml
```diff
@@ -15,6 +15,7 @@ data:
     volumeBindingMode: Immediate
     parameters:
       fsType: "ext4"
+      mkfsParams: "-I 512 -b 2048 -O ^metadata_csum,^64bit"
```
derekbit, line 18: Should not change the `storageclass.yaml`. This is the default `longhorn` storageclass.
Add the example in the `longhorn/longhorn`'s `example/storageclass.yaml`

weizhe0422: I have fixed to remove the parameter `mkfs Params` from this yaml file.

derekbit (approved):
LGTM.

BTW, should we mark the `mkfs-ext4-parameters` setting as deprecated?

innobead:
Yeah, we should mark it deprecated by doc, and remove it from 1.5.0.

Let's create a ticket for the removal.

innobead (approved):

PR #1559, opened 2022-11-19
version: update to v1.4.0-dev

derekbit:
[Longhorn 4906](https://github.com/longhorn/longhorn/issues/4906)


innobead (changes requested):
Let's use `v1.4.0-dev` instead before 1.4.0 releases to avoid misleading and confusing metrics in longhornio metrics.

ref: https://github.com/longhorn/longhorn/blob/75222752d22abfd423f8c650f5ebf1e5c3f96ff7/chart/Chart.yaml#L3-L3

innobead (approved):

PR #1560, opened 2022-11-22
dev!: remove deploy and examples duplicated in longhorn repo

innobead:

For any development, developers should base on the manifests in the longhorn repo instead of the longhorn-manager to avoid any missing commits or inconsistent manifest changes between these two repos.  

cc @longhorn/dev


innobead:
@yangchiu Will this change impact our testing? I believe we already move to use manifests in longhorn/longhorn.

cc @longhorn/qa

yangchiu:
No, longhorn-tests didn't reference longhorn-manager anymore.

PR #1561, opened 2022-11-22
chore: generate_code.sh missing path

c3y1huang:
Side-effect from https://github.com/longhorn/longhorn-manager/pull/1560


k8s/generate_code.sh
```diff
@@ -65,5 +65,7 @@ if [ -e ${GOPATH}/src/${LH_MANAGER_DIR}/k8s/patches/crd ]; then
   find patches -type f | xargs -i sh -c 'kustomize edit add patch --path {}'
 fi
 popd
+
+mkdir -p ${GOPATH}/src/${LH_MANAGER_DIR}/deploy/install/01-prerequisite
```
innobead, line 69: We don't need the deploy folder anymore, but still need the generated CRD yaml for sure.

Suggest changing to
```bash

echo "# Generated by the CRDs from ${APIS_DIR}" > ${GOPATH}/src/${LH_MANAGER_DIR}/k8s/crd.yaml
kustomize build ${CRDS_DIR} >> ${GOPATH}/src/${LH_MANAGER_DIR}/k8s/crd.yaml
rm -r ${CRDS_DIR}
```

innobead (changes requested):

k8s/generate_code.sh
c3y1huang: Updated.

PR #1562, opened 2022-11-22
feat: support local volume

derekbit:
[Longhorn 3957](https://github.com/longhorn/longhorn/issues/3957)

### DIFF
--- a/datastore/longhorn.go
+++ b/datastore/longhorn.go
@@ -1366,16 +1366,21 @@ func (s *DataStore) CheckEngineImageReadiness(image string, nodes ...string) (is
 }
 
 // CheckEngineImageReadyOnAtLeastOneVolumeReplica checks if the IMAGE is deployed on the NODEID and on at least one of the the volume's replicas
-func (s *DataStore) CheckEngineImageReadyOnAtLeastOneVolumeReplica(image, volumeName, nodeID string) (bool, error) {
-	replicas, err := s.ListVolumeReplicas(volumeName)
+func (s *DataStore) CheckEngineImageReadyOnAtLeastOneVolumeReplica(image, volumeName, nodeID string, dataLocality longhorn.DataLocality) (bool, error) {
+	isReady, err := s.CheckEngineImageReadiness(image, nodeID)
 	if err != nil {
-		return false, fmt.Errorf("cannot get replicas for volume %v: %w", volumeName, err)
+		return false, errors.Wrapf(err, "failed to check engine image readiness of node %v", nodeID)
 	}
 
-	isReady, err := s.CheckEngineImageReadiness(image, nodeID)
-	if nodeID != "" && !isReady {
-		return isReady, err
+	if !isReady || dataLocality == longhorn.DataLocalityStrictLocal {
+		return isReady, nil
 	}
+
+	replicas, err := s.ListVolumeReplicas(volumeName)
+	if err != nil {
+		return false, errors.Wrapf(err, "cannot get replicas for volume %v", volumeName)
+	}
+
 	for _, r := range replicas {
 		isReady, err := s.CheckEngineImageReadiness(image, r.Spec.NodeID)
 		if err != nil || isReady {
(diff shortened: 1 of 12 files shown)


manager/volume.go
```diff
@@ -232,11 +232,13 @@ func (m *VolumeManager) Attach(name, nodeID string, disableFrontend bool, attach
 		return nil, err
 	}
 
-	if isReady, err := m.ds.CheckEngineImageReadyOnAtLeastOneVolumeReplica(v.Spec.EngineImage, v.Name, nodeID); !isReady {
-		if err != nil {
-			return nil, fmt.Errorf("cannot attach volume %v with image %v: %v", v.Name, v.Spec.EngineImage, err)
+	if v.Spec.DataLocality != longhorn.DataLocalityStrictLocal {
```
innobead, line 235: Why we need to add the condition here? Checking the engine binary on the node is quite general, so it should nto be filtered out?

innobead (changes requested):
In general, LGTM. One comment needed to resolve.

manager/volume.go
derekbit: Because the replica of the local volume is not really scheduled to a node after creation, the scheduling to the engine's node is done after executing attaching. So, the `replica.spec.nodeID` is empty, `CheckEngineImageReadyOnAtLeastOneVolumeReplica` will always fail.

innobead: I see, but you can't just bypass checking image on `nodeID`, otherwise the engine and replica on `nodeID` would not be able to start. 

So it's better to skip checking images on replica nodes in `CheckEngineImageReadyOnAtLeastOneVolumeReplica` if the volume is data-locality-strict but keep checking `nodeID`?

derekbit: Addressed. Thank you.

datastore/longhorn.go
```diff
@@ -1366,16 +1366,19 @@ func (s *DataStore) CheckEngineImageReadiness(image string, nodes ...string) (is
 }
 
 // CheckEngineImageReadyOnAtLeastOneVolumeReplica checks if the IMAGE is deployed on the NODEID and on at least one of the the volume's replicas
-func (s *DataStore) CheckEngineImageReadyOnAtLeastOneVolumeReplica(image, volumeName, nodeID string) (bool, error) {
+func (s *DataStore) CheckEngineImageReadyOnAtLeastOneVolumeReplica(image, volumeName, nodeID string, dataLocality longhorn.DataLocality) (bool, error) {
+	isReady, err := s.CheckEngineImageReadiness(image, nodeID)
```
innobead, line 1370: Suggest

```go
	isReady, err := s.CheckEngineImageReadiness(image, nodeID)
	if (nodeID != "" && !isReady) {
		return isReady, err
	}

	if (dataLocality == longhorn.DataLocalityStrictLocal) {
		return isReady, nil
	}

	...
```

innobead (changes requested):

datastore/longhorn.go
derekbit: Addressed. Thank you.

datastore/longhorn.go
```diff
@@ -1366,16 +1366,21 @@ func (s *DataStore) CheckEngineImageReadiness(image string, nodes ...string) (is
...
+	if nodeID != "" && !isReady {
+		return isReady, err
+	}
+
+	if dataLocality == longhorn.DataLocalityStrictLocal {
+		return isReady, nil
+	}
```
innobead, line 1377: We missed the error handling. 

```go
	isReady, err := s.CheckEngineImageReadiness(image, nodeID)
    if err != nil {
        return false, fmt.Errorf("failed to check engine image readiness of node %v: %w", nodeID, err)
    }
    
	if !isReady || dataLocality == longhorn.DataLocalityStrictLocal {
		return isReady, nil
	}

    ...
```

innobead (changes requested):

datastore/longhorn.go

innobead (approved):

PR #1563, opened 2022-11-22
Remove unused InstanceManagerProxyAPIVersion

derekbit:
CurrentInstanceManagerProxyAPIVersion and its check function are unused for now.
Remove it for avoiding the confusion.

https://github.com/longhorn/longhorn/issues/4210
https://github.com/longhorn/longhorn/issues/3198

### DIFF
--- a/engineapi/instance_manager.go
+++ b/engineapi/instance_manager.go
@@ -10,9 +10,8 @@ import (
 	imclient "github.com/longhorn/longhorn-instance-manager/pkg/client"
 	imutil "github.com/longhorn/longhorn-instance-manager/pkg/util"
 
-	"github.com/longhorn/longhorn-manager/types"
-
 	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
+	"github.com/longhorn/longhorn-manager/types"
 )
 
 const (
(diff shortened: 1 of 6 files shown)


innobead:
Why do we need to update the version?

derekbit:
The [instance-manager proxy api version](https://github.com/longhorn/longhorn-instance-manager/blob/master/pkg/meta/version.go) is 3. We update the current proxy api in longhorn-manager to 3, right?
BTW, what's the difference between instance-manager proxy api version and instance-manager api version?

innobead:
> The [instance-manager proxy api version](https://github.com/longhorn/longhorn-instance-manager/blob/master/pkg/meta/version.go?rgh-link-date=2022-11-22T16%3A37%3A35Z) is 3. We update the current proxy api in longhorn-manager to 3, right? BTW, what's the difference between instance-manager proxy api version and instance-manager api version?

I think we probably confuse the version compatibility here.

`InstanceManagerAPIVersion` is for ProcessManagerService defined in imrpc.proto
`InstanceManagerProxyAPIVersion` is for ProxyEngineService defined in proxy.proto

However, it seems we don't have any proxy compatibility check in longhorn-manager. Also, literally, if we always make gRPC and proto backward compatible, we don't really need to handle this kind of custom version compatibility check.

engineapi/instance_manager.go
```diff
@@ -16,11 +16,11 @@ import (
 )
 
 const (
-	CurrentInstanceManagerAPIVersion = 2
+	CurrentInstanceManagerAPIVersion = 3
```
innobead, line 19: Don't need to change this version because the process gRPC doesn't change?

engineapi/instance_manager.go
```diff
@@ -16,11 +16,11 @@ import (
...
 	MinInstanceManagerAPIVersion     = 1
 	UnknownInstanceManagerAPIVersion = 0
 
-	CurrentInstanceManagerProxyAPIVersion = 2
+	CurrentInstanceManagerProxyAPIVersion = 3
```
innobead, line 23: If changing the existing proxy interface method like parameter change, need to update the min as well.

innobead (commented):
Let's use the concurrent mechanism first but need to review the version compatibility check between manager and instance manager later.

@derekbit Could u help create an issue for this later review? Thanks.

innobead:
Also, the check PR failed?

derekbit:
> Also, the check PR failed?

Yeah, the unit test is not updated yet.

engineapi/instance_manager.go
derekbit: Yes, should no change `CurrentInstanceManagerAPIVersion` here.

PR #1568, opened 2022-11-25
types: setting 'Support Bundle Manager Image' should be read-only

shuo-wu:
Otherwise, the setting reset function at the beginning of each integration test will error out

### DIFF
--- a/types/setting.go
+++ b/types/setting.go
@@ -365,7 +365,7 @@ var (
 		Category:    SettingCategoryGeneral,
 		Type:        SettingTypeString,
 		Required:    true,
-		ReadOnly:    false,
+		ReadOnly:    true,
 	}
 
 	SettingDefinitionReplicaSoftAntiAffinity = SettingDefinition{


innobead:
@shuo-wu What's the error?

shuo-wu:
[This function](https://github.com/longhorn/longhorn-tests/blob/f4fbf6395e00684e738bf725b6be5f3ccc47682d/manager/integration/tests/common.py#L1504) will reset all non-read-only settings to the default values. But this new setting has no default value and it will reject the empty value (updated by the function).

c3y1huang:
The fix for regression is [here](https://github.com/longhorn/longhorn-tests/pull/1187/files#diff-a2154691ec50c6925b7f385cf30c42f33b2a3c935192c894bfe4b32fe3ee49dcR3100-R3101).

IIUC, having `ReadOnly:    true`  prohibits the user from updating the image? If true, then it's my intention to have this `false`.

shuo-wu:
Yes. This kind of read-only settings are typically determined by the new longhorn-manager flags and updated at the launching stage. Maybe I need to check your LEP... I was thinking this is similar to other default image settings...

innobead:
> Yes. This kind of read-only settings are typically determined by the new longhorn-manager flags and updated at the launching stage. Maybe I need to check your LEP... I was thinking this is similar to other default image settings...

After discussing with @c3y1huang , the version of the support bundle kit will be specified by a command option when starting longhorn-manager, but the design we want to let users able to update later by cases for troubleshooting purposes. 

We should just fix the test teardown to remove this consumed error out.

shuo-wu:
OK. Then we can close this.

PR #1569, opened 2022-11-25
fix: Check if backup exists before performing backup deletion

weizhe0422:
[Longhorn 4107](https://github.com/longhorn/longhorn/issues/4107)

If the backup not exists but performing backup deletion, then it will return error.  In the [Longhorn 4107](https://github.com/longhorn/longhorn/issues/4107) scenario, the `volumesnapshot` cannot be deleted normally.  

Test result
 - the `volumesnapshot` can be delete normally even if the backup was removed first
     image

### DIFF
--- a/api/backup.go
+++ b/api/backup.go
@@ -15,7 +15,7 @@ func (s *Server) BackupTargetList(w http.ResponseWriter, req *http.Request) erro
 
 	backupTargets, err := s.m.ListBackupTargetsSorted()
 	if err != nil {
-		return errors.Wrapf(err, "error listing backup targets")
+		return errors.Wrap(err, "failed to list backup targets")
 	}
 	apiContext.Write(toBackupTargetCollection(backupTargets))
 	return nil
@@ -35,7 +35,7 @@ func (s *Server) BackupVolumeList(w http.ResponseWriter, req *http.Request) erro
 func (s *Server) backupVolumeList(apiContext *api.ApiContext) (*client.GenericCollection, error) {
 	bvs, err := s.m.ListBackupVolumesSorted()
 	if err != nil {
-		return nil, errors.Wrap(err, "error listing backup volume")
+		return nil, errors.Wrap(err, "failed to list backup volume")
 	}
 	return toBackupVolumeCollection(bvs, apiContext), nil
 }
@@ -47,7 +47,7 @@ func (s *Server) BackupVolumeGet(w http.ResponseWriter, req *http.Request) error
 
 	bv, err := s.m.GetBackupVolume(volName)
 	if err != nil {
-		return errors.Wrapf(err, "error get backup volume '%s'", volName)
+		return errors.Wrapf(err, "failed to get backup volume '%s'", volName)
 	}
 	apiContext.Write(toBackupVolumeResource(bv, apiContext))
 	return nil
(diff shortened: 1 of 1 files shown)


api/backup.go
```diff
@@ -122,8 +122,15 @@ func (s *Server) BackupDelete(w http.ResponseWriter, req *http.Request) error {
...
 
-	if err := s.m.DeleteBackup(input.Name, volName); err != nil {
-		return errors.Wrapf(err, "error deleting backup %v of volume %v", input.Name, volName)
+	backup, err := s.m.GetBackup(input.Name, volName)
+	if err != nil {
+		logrus.Errorf("Error getting backup %v of volume %v: %v", input.Name, volName, err)
+	}
+
+	if backup != nil {
+		if err := s.m.DeleteBackup(input.Name, volName); err != nil {
```
innobead, line 131: The fix is ok, but I feel we can fix it general for all resource deletion if the target is not found. 

Do you know what error will be returned?

```go
s.lhClient.LonghornV1beta2().Backups(s.namespace).Delete(context.TODO(), backupName, metav1.DeleteOptions{})
```

api/backup.go
```diff
@@ -122,8 +122,15 @@ func (s *Server) BackupDelete(w http.ResponseWriter, req *http.Request) error {
 
 	volName := mux.Vars(req)["volName"]
 
-	if err := s.m.DeleteBackup(input.Name, volName); err != nil {
-		return errors.Wrapf(err, "error deleting backup %v of volume %v", input.Name, volName)
+	backup, err := s.m.GetBackup(input.Name, volName)
+	if err != nil {
+		logrus.Errorf("Error getting backup %v of volume %v: %v", input.Name, volName, err)
```
innobead, line 127: Because you tolerate this error on purpose, please use `warnf` and `logrus.witherror`. 

Suggest changing the log to `failed to get backup %v of volume %v`.

innobead (changes requested):

api/backup.go
weizhe0422: After discussion, because the return of the Delete function is the implementation of Error(). 

It's hard to distinguish error classes without getting the HTTP status code.

https://github.com/longhorn/longhorn-manager/blob/d4d5f8943f62cfe84c4745fd93a796c3e0789033/k8s/pkg/client/clientset/versioned/typed/longhorn/v1beta2/backup.go#L156-L164

api/backup.go
weizhe0422: I have fixed it following the comments. Thank you.

innobead (approved):

api/backup.go
```diff
@@ -117,18 +117,25 @@ func (s *Server) BackupDelete(w http.ResponseWriter, req *http.Request) error {
...
 	}
 
 	volName := mux.Vars(req)["volName"]
 
-	if err := s.m.DeleteBackup(input.Name, volName); err != nil {
-		return errors.Wrapf(err, "error deleting backup %v of volume %v", input.Name, volName)
+	backup, err := s.m.GetBackup(input.Name, volName)
+	if err != nil {
+		logrus.WithError(err).Warnf("failed to get backup '%v' of volume '%v'", input.Name, volName)
+	}
```
derekbit, line 128: The `backup` from `GetBackup` is unused. Why not directly do `DeleteBackup`?

weizhe0422: The 'GetBackup' will return error if API exception to get backup or the backup is not exists. 

Therefore, I log the error but checking if 'backup' is null or not and perform deletion if still exists.

If I execute 'DeleteBackup' directly, it will be the same as the original situation, causing volumesnapshot cannot be removed.

derekbit: I see. The code flow is a bit of confusing.
Will DeleteBackup return NOTEXIST error? We can ignore the error and continue with the remaining tasks if yes. Or, suggest adding a comment here.

weizhe0422: Yes, the error message will mention that the backup does not exist.

But the current way of judging is to parse the error string, because there is no error code or HTTP status to distinguish the error category.

I will add comment to describe the context and also create a ticket to see if the API could be enhanced to export the HTTP status code.

PR #1570, opened 2022-11-28
feat(csi-liveness-probe): longhorn-csi-plugin

c3y1huang:
https://github.com/longhorn/longhorn/issues/3907

### DIFF
--- a/csi/deployment.go
+++ b/csi/deployment.go
@@ -10,9 +10,11 @@ import (
 	v1 "k8s.io/api/core/v1"
 	storagev1 "k8s.io/api/storage/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
 	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/utils/pointer"
 
+	"github.com/longhorn/longhorn-manager/datastore"
 	"github.com/longhorn/longhorn-manager/types"
 	"github.com/longhorn/longhorn-manager/util"
 )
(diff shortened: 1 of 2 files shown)


csi/deployment.go
```diff
@@ -23,6 +25,7 @@ const (
 	DefaultCSIResizerImage             = "longhornio/csi-resizer:v1.3.0"
 	DefaultCSISnapshotterImage         = "longhornio/csi-snapshotter:v5.0.1"
 	DefaultCSINodeDriverRegistrarImage = "longhornio/csi-node-driver-registrar:v2.5.0"
+	DefaultCSILivenessProbeImage       = "registry.k8s.io/sig-storage/livenessprobe:v2.8.0"
```
c3y1huang, line 28: TODO: update after https://github.com/longhorn/longhorn/issues/4945.

csi/deployment.go
```diff
@@ -387,6 +391,20 @@ func NewPluginDeployment(namespace, serviceAccount, nodeDriverRegistrarImage, ma
 								},
 							},
 						},
+						{
+							Name:            "longhorn-liveness-probe",
+							ImagePullPolicy: imagePullPolicy,
+							Image:           livenessProbeImage,
+							Args: []string{
+								fmt.Sprintf("--csi-address=%s", GetInContainerCSISocketFilePath()),
```
innobead, line 399: No verbose? Just have a try to see what level is much suitable. 

ref: https://github.com/kubernetes-csi/livenessprobe/blob/master/deployment/kubernetes/livenessprobe-sidecar.yaml#L51

innobead (commented):
LGTM, excetp one comment.

csi/deployment.go
c3y1huang: Yes, we should set verbose, I think the default is set to `v=2`. 
`v=5` output the [probe request](https://github.com/kubernetes-csi/livenessprobe/blob/master/cmd/livenessprobe/main.go#L69) and the [connections](https://github.com/kubernetes-csi/livenessprobe/blob/master/vendor/github.com/kubernetes-csi/csi-lib-utils/connection/connection.go#L183-L187), so set `v=4` should be at the debug level and looks right to me.

innobead (approved):

csi/deployment.go
c3y1huang: Ok, I think v=4 is still too much. Need to make it less verbose.
```
[longhorn-csi-plugin-hf5vb longhorn-csi-plugin] time="2022-11-29T03:27:43Z" level=info msg="Probe: rsp: {}"
[longhorn-csi-plugin-rsvwr longhorn-csi-plugin] time="2022-11-29T03:27:43Z" level=info msg="Probe: req: {}"
```

PR #1575, opened 2022-11-29
fix(engine-binary): recovery

c3y1huang:
https://github.com/longhorn/longhorn/issues/4380


controller/engine_image_controller.go
```diff
@@ -829,6 +829,20 @@ func (ic *EngineImageController) createEngineImageDaemonSetSpec(ei *longhorn.Eng
...
+										Command: []string{
+											"sh", "-c",
+											"/data/longhorn version --client-only",
+										},
+									},
+								},
+								InitialDelaySeconds: datastore.PodProbeInitialDelay,
+								TimeoutSeconds:      datastore.PodProbeTimeoutSeconds,
+								PeriodSeconds:       datastore.PodProbePeriodSeconds,
+								FailureThreshold:    datastore.PodLivenessProbeFailureThreshold,
```
innobead, line 844: I think we can make time-related settings of `ReadinessProbe` to align with these values as well.

innobead (approved):
LGTM, it's good to use liveness.

innobead (changes requested):
https://github.com/longhorn/longhorn-manager/pull/1575/files#r1034568482

innobead (approved):

PR #1576, opened 2022-11-29
Check engine image version before creating a replica process

derekbit:
[Longhorn 4962](https://github.com/longhorn/longhorn/issues/4962)

### DIFF
--- a/engineapi/instance_manager.go
+++ b/engineapi/instance_manager.go
@@ -197,7 +197,7 @@ func (c *InstanceManagerClient) EngineProcessCreate(e *longhorn.Engine, volumeFr
 	return c.parseProcess(engineProcess), nil
 }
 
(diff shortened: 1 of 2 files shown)


engineapi/instance_manager.go
```diff
@@ -212,10 +212,12 @@ func (c *InstanceManagerClient) ReplicaProcessCreate(replica *longhorn.Replica,
 		args = append(args, "--disableRevCounter")
 	}
 
-	args = append(args, "--volume-name", replica.Spec.VolumeName)
+	if engineCLIAPIVersion >= 7 {
```
innobead, line 215: It seems we didn't do any compatibility fore replica process management before.

innobead (approved):
LGTM

PR #1582, opened 2022-12-05
Fail all replicas if there is no valid backend during the engine starting stage

derekbit:
In this PR, the `instanceCreation` condition is introduced in `engine.Spec.InstanceStatus`. The condition will be set to `false` if the manager failed to create the engine process. The volume controller will be aware of the condition with a `false` value. If the condition says the error is due to that there is no available backend, the replicas of the volume will be marked as failed.

[Longhorn 1330](https://github.com/longhorn/longhorn/issues/1330)


derekbit:
e2e
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/2641/console

controller/volume_controller.go
```diff
@@ -549,9 +549,12 @@ func (vc *VolumeController) ReconcileEngineReplicaState(v *longhorn.Volume, es m
 	if e.Status.CurrentState != longhorn.InstanceStateRunning {
 		// If a replica failed at attaching stage before engine become running,
 		// there is no record in e.Status.ReplicaModeMap
+		engineInstanceCreationCondition := types.GetCondition(e.Status.Conditions, longhorn.InstanceConditionTypeInstanceCreation)
+		isNoAvailableBackend := strings.Contains(engineInstanceCreationCondition.Message, fmt.Sprintf("exit status %v", int(syscall.ENODATA)))
```
innobead, line 554: Where to cause ` fmt.Sprintf("exit status %v", int(syscall.ENODATA)))`? Didn't see any code from engine creation (controller command)

innobead (commented):
LGTM.

controller/volume_controller.go
derekbit: Set the exit code ENODATA here.
https://github.com/longhorn/longhorn-engine/pull/780/files#diff-33f40fb2b9aa0391a4eec1808d3248f24494e63100dd5defe18370f3f2236390R172

innobead (approved):

controller/volume_controller.go
shuo-wu, line 554: Previously I tried to launch an error struct so that Longhorn can easily marshal and unmarshal error codes (even among different components). I am not sure if we can popularize it and make it a separate repo. The idea is from the gRPC error handling. WDYT?

derekbit: Sounds good. Let's work the idea in https://github.com/longhorn/longhorn/issues/4996. WDYT?

derekbit:
@Mergify backport v1.3.x

PR #1583, opened 2022-12-05
feat(concurrent-backup-restore): implement

c3y1huang:
https://github.com/longhorn/longhorn/issues/4558

### DIFF
--- a/controller/engine_controller.go
+++ b/controller/engine_controller.go
@@ -984,6 +984,16 @@ func (m *EngineMonitor) refresh(engine *longhorn.Engine) error {
 		}
 		isDRVolume := volume.Status.IsStandby
 		if !isDRVolume {
+			if isReachedLimit, err := m.isReachedConcurrentVolumeBackupRestoreLimit(); err != nil {
+				m.logger.WithError(err).Warn("Failed to check concurrent volume backup restore limit")
+				return nil
+
+			} else if isReachedLimit {
+				m.logger.Debugf("Reached the concurrent volume backup restore limit of %v, retry later", types.SettingNameConcurrentBackupRestorePerNodeLimit)
+				m.restoreBackoff.Next(engine.Name, time.Now())
+				return nil
+			}
+
 			m.aquireRestoringCounter(true)
 			defer m.aquireRestoringCounter(false)
 		}
@@ -1308,6 +1318,15 @@ func (m *EngineMonitor) restoreBackup(engine *longhorn.Engine, rsMap map[string]
 	}
 
 	return nil
+}
+
+func (m *EngineMonitor) isReachedConcurrentVolumeBackupRestoreLimit() (isUnderLimit bool, err error) {
+	limit, err := m.ds.GetSettingAsInt(types.SettingNameConcurrentBackupRestorePerNodeLimit)
+	if err != nil {
+		return false, err
+	}
+
+	return int(m.restoringCounter.GetCount()) >= int(limit), nil
 }
 
 func handleRestoreError(log logrus.FieldLogger, engine *longhorn.Engine, rsMap map[string]*longhorn.RestoreStatus, backoff *flowcontrol.Backoff, err error) error {
--- a/controller/engine_controller.go
+++ b/controller/engine_controller.go
@@ -84,6 +84,7 @@ type EngineController struct {
 	engineMonitorMap   map[string]chan struct{}
 
 	proxyConnCounter util.Counter
+	restoringCounter util.Counter
 }
 
 type EngineMonitor struct {
@@ -104,6 +105,7 @@ type EngineMonitor struct {
 	monitorVoluntaryStopCh chan struct{}
 
 	proxyConnCounter util.Counter
+	restoringCounter util.Counter
 }
 
 func NewEngineController(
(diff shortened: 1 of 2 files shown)


c3y1huang:
Test result: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/2638/testReport/tests/test_settings/

c3y1huang:
Test result https://ci.longhorn.io/job/private/job/longhorn-tests-regression/2647/testReport/tests/test_settings/

controller/engine_controller.go
```diff
@@ -1002,6 +1030,29 @@ func (m *EngineMonitor) refresh(engine *longhorn.Engine) error {
...
+		m.restoringCounter.IncreaseCount()
+		m.restoringCounterAquired = aquire
+		return
+	}
+
+	if !aquire && m.restoringCounterAquired {
+		m.restoringCounter.DecreaseCount()
+		m.restoringCounterAquired = aquire
+		return
+	}
```
derekbit, line 1053:
```suggestion
	if !aquire && m.restoringCounterAquired {
		m.restoringCounter.DecreaseCount()
		m.restoringCounterAquired = aquire
	}
```

controller/engine_controller.go
```diff
@@ -973,11 +989,23 @@ func (m *EngineMonitor) refresh(engine *longhorn.Engine) error {
 			m.logger.Debug("Cannot restore the backup for engine since it is still in the backoff window")
 			return nil
 		}
+
+		if isReachedLimit, err := m.isReachedConcurrentVolumeBackupRestoreLimit(); err != nil {
+			m.logger.WithError(err).Warn("Failed to check concurrent volume backup restore limit")
+			return nil
+
+		} else if isReachedLimit {
+			m.logger.Debugf("Reached the concurrent volume backup restore limit of %v, retry later", types.SettingNameConcurrentBackupRestorePerNodeLimit)
```
derekbit, line 998: Should it be warning level message?

innobead (changes requested):

controller/engine_controller.go
derekbit: I see. Just thought if we can easily know why the backup does not take place from log.

controller/engine_controller.go
```diff
@@ -973,11 +977,24 @@ func (m *EngineMonitor) refresh(engine *longhorn.Engine) error {
...
+			m.restoreBackoff.Next(engine.Name, time.Now())
+			return nil
+		}
```
innobead, line 989: You can add this part to `aquireRestoringCounter`, but bypass this check if the volume is DR.

controller/engine_controller.go
```diff
@@ -1002,6 +1009,24 @@ func (m *EngineMonitor) refresh(engine *longhorn.Engine) error {
...
+
+	volume, err := m.ds.GetVolumeRO(engine.Spec.VolumeName)
+	if err != nil {
+		m.logger.Warnf("Failed to get volume %v for restoring counter", engine.Spec.VolumeName)
+		return
+	}
+	isDRVolume := volume.Status.IsStandby
+	if isDRVolume {
+		return
+	}
```
innobead, line 1036: Move this to the top, because need to bypass all counting if the volume is DR.

innobead (changes requested):

innobead (dismissed):
LGTM

c3y1huang:
Test result: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/2664/

controller/engine_controller.go
c3y1huang: We need to return this, so keep this as it is.

controller/engine_controller.go
c3y1huang: Moved this to the caller so it will only be checked once.

controller/engine_controller.go
innobead: You are right!

controller/engine_controller.go
```diff
@@ -973,11 +977,31 @@ func (m *EngineMonitor) refresh(engine *longhorn.Engine) error {
...
+
+		if isReachedLimit, err := m.isReachedConcurrentVolumeBackupRestoreLimit(); err != nil {
+			m.logger.WithError(err).Warn("Failed to check concurrent volume backup restore limit")
+			return nil
+
+		} else if isReachedLimit {
+			m.logger.Debugf("Reached the concurrent volume backup restore limit of %v, retry later", types.SettingNameConcurrentBackupRestorePerNodeLimit)
+			m.restoreBackoff.Next(engine.Name, time.Now())
+			return nil
+		}
```
innobead, line 989: This part should be checked only if the volume is not DR, right?

Should move into the below `if !isDRVolume {`

c3y1huang:
Test result: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/2665/console

c3y1huang:
Test result: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/2665/

innobead (approved):

shuo-wu:
Sorry for the late comment, I think there may be one issue:
The restore actually happens inside a replica process. And IIRC, Longhorn manager does not guarantee the engine and the restoring replica are on the same node. In other words, this PR cannot limit the restoring replica count on a node.

innobead:
> Sorry for the late comment, I think there may be one issue: The restore actually happens inside a replica process. And IIRC, Longhorn manager does not guarantee the engine and the restoring replica are on the same node. In other words, this PR cannot limit the restoring replica count on a node.

You are right. Here, @c3y1huang 's implementation just limits the count of engines but the real processing unit is each replica which is likely not on the same node. However, this is still can take some effect from the cluster perspective because this concurrent limitation will be accumulated. 

This can be improved in the future to make the limitation precise for each node.

PR #1588, opened 2022-12-07
snapshot data integrity: add api

derekbit:
Longhorn/longhorn#4994


innobead (approved):
LGTM, but just a few convention feedback. See if you would like to refactor this right now, or we can tackle it later.

innobead:
After discussing with @derekbit , we will tackle the log/error format in a specific issue later on. https://github.com/longhorn/longhorn/issues/5015

PR #1590, opened 2022-12-07
fix: Check if the channel has been closed before closing

weizhe0422:
[Longhorn 4865](https://github.com/longhorn/longhorn/issues/4865)

- When the `backingImageDataSource` POD fails to respond, the monitor thread in the LH manager will consider the `backingImageDataSource` pod as invalid. Then it tries to stop and close its own channel but it has already been closed. So check to see if the channel is closed before closing it to avoid panic.

- Self test results:
  - After monitoring is stopped due to failure to connect to the data source POD, the next round of monitoring will continue.
    image

### DIFF
--- a/controller/backing_image_data_source_controller.go
+++ b/controller/backing_image_data_source_controller.go
@@ -927,10 +927,10 @@ func (c *BackingImageDataSourceController) stopMonitoring(bidsName string) {
 		log.Warn("No monitor goroutine for stopping")
 		return
 	}
-	log.Infof("Stopping monitoring")
+	log.Info("Stopping monitoring")
 	close(stopCh)
 	delete(c.monitorMap, bidsName)
-	log.Infof("Stopped monitoring")
+	log.Info("Stopped monitoring")
 
 }
 
@@ -978,7 +978,7 @@ func (m *BackingImageDataSourceMonitor) sync() {
 		if syncErr != nil {
 			m.retryCount++
 			if m.retryCount == engineapi.MaxMonitorRetryCount {
-				close(m.stopCh)
+				m.stopCh <- struct{}{}
 				m.log.Warnf("Stop monitoring since monitor %v sync reaches the max retry count %v", m.Name, engineapi.MaxMonitorRetryCount)
 				return
 			}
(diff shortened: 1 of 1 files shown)


controller/backing_image_data_source_controller.go
```diff
@@ -927,10 +927,21 @@ func (c *BackingImageDataSourceController) stopMonitoring(bidsName string) {
 		log.Warn("No monitor goroutine for stopping")
 		return
 	}
-	log.Infof("Stopping monitoring")
-	close(stopCh)
+	log.Info("Stopping monitoring")
+
+	// https://github.com/golang/go/blob/master/src/runtime/chan.go
+	// If the channel is closed and buffer is empty, the chanrecv() will return (true, false)
+	select {
```
innobead, line 934: This workaround is OK, but why not just send to a signal to the stop channel in the sync method when the data source is not responsible, then leave the stopMonitor help with closing channel instead?

The concept is `who create the resource, then she needs to handle it like pointer`, so the ownership will be clear.

innobead (changes requested):

controller/backing_image_data_source_controller.go
weizhe0422: Yes, I think you're right, it should just let `stop Monitoring()` do all about stopping the monitor. So I send the signal `struct{}{}` to the stop channel instead.

controller/backing_image_data_source_controller.go
```diff
@@ -978,7 +978,7 @@ func (m *BackingImageDataSourceMonitor) sync() {
 		if syncErr != nil {
 			m.retryCount++
 			if m.retryCount == engineapi.MaxMonitorRetryCount {
-				close(m.stopCh)
+				m.stopCh <- struct{}{}
```
innobead, line 981: Also some `close(m.stopCh)` in the sync method.

innobead (commented):
Also some `close(m.stopCh)` in the sync method.

controller/backing_image_data_source_controller.go
weizhe0422: I have replace the all of closing channel behavior with sending `struct{}{}` into channel to notify stopping monitor.

innobead: Have you reverified the fix?

weizhe0422: Yes, I verify the https://github.com/longhorn/longhorn/issues/4865 could be fixed and also running e2e test now.

innobead (approved):

controller/backing_image_data_source_controller.go
weizhe0422: Update the e2e result for backingimage is pass
image

shuo-wu:
Sorry for the late reply, will this somehow lead to panic `send on closed channel`? In the current implementation, `c.stopMonitoring` can be invoked by the main reconcile loop, which is out of the monitoring sync loop.

PR #1592, opened 2022-12-08
Enable fast replica rebuilding by default

derekbit:
longhorn/longhorn#5023


innobead (dismissed):

types/setting.go
```diff
@@ -992,12 +992,12 @@ var (
 
 	SettingDefinitionFastReplicaRebuildEnabled = SettingDefinition{
 		DisplayName: "Fast Replica Rebuild Enabled",
-		Description: "Enabling this setting will support the fast replica rebuild. The feature relies on the checksum of snapshot disk files, so the feature is setting the snapshot-data-integrity to **enable** or **fast-check** is a prerequisite.",
+		Description: "This setting enable the fast replica rebuilding feature. It relies on the checksums of snapshot disk files, so setting the snapshot-data-integrity to **enable** or **fast-check** is a prerequisite.",
```
innobead, line 995:
```suggestion
		Description: "This setting enables the fast replica rebuilding feature. It relies on the checksums of snapshot disk files, so setting the snapshot-data-integrity to **enable** or **fast-check** is a prerequisite.",
```

innobead (changes requested):

types/setting.go
derekbit: Fixed. Thank you.

innobead (approved):

PR #1593, opened 2022-12-08
feat(system-backup): sync if exist in remote

c3y1huang:
Because NFS mount cannot work without privilege, so proposing to check and sync the system backup if already exists in the backup store.

https://github.com/longhorn/longhorn/pull/4916


c3y1huang:
Tested [test(manual): create system backup exist in remote](https://github.com/longhorn/longhorn-tests/pull/1191/commits/ee94891c34318ae724d9e2d5cf669fb4935de4d2)

c3y1huang:
Test result: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/2662/

controller/system_backup_controller.go
```diff
@@ -547,6 +588,11 @@ func (c *SystemBackupController) cleanupRemoteSystemBackupFiles(systemBackup *lo
 
 	log := getLoggerForSystemBackup(c.logger, systemBackup.Name)
 
+	if systemBackup.Status.Version == "" {
+		// The backup store sync might not have finished
+		return nil
+	}
```
innobead, line 594: Why this matters?

BTW, please udpate the message of the error in the following if-block.

```
if systemBackupCfg != nil {
		return errors.New("failed to check the system backup deletion, because it's being deleted")
	}

```

innobead (commented):
LGTM, except one comment needed to clarify.

controller/system_backup_controller.go
c3y1huang: > Why this matters?
> 

If the system backup exists in the backup store then we
1. Tag the SystemBackup with the volume label -> update SystemBackup.
1. The controller remote system backup info -> update SystemBackup.status.
If the SystemBackup gets deleted somewhere in between the transmission state then it might not have the version to use to validate the deletion and cause a blocking.

So for the above case, we can just continue deleting the CR, and later the backup target sync will kick back in and the user can remove it again.

datastore/longhorn.go
```diff
@@ -4092,6 +4104,21 @@ func (s *DataStore) ListSystemBackupsRO() ([]*longhorn.SystemBackup, error) {
 	return s.sbLister.SystemBackups(s.namespace).List(labels.Everything())
 }
 
+func TagSystemBackupVersionLabel(version string, obj runtime.Object) error {
```
innobead, line 4107: This is only used for system backup controller? why not just move to it and make it as a private function.

Suggest naming `labelSystemBackupLonghornVersion`.

innobead (changes requested):

datastore/longhorn.go
innobead: The convention is not quite clear for me, but it's good now.

PR #1594, opened 2022-12-08
Fix the missing api codes for volume.spec.snapshotDataIntegrity

derekbit:
Longhorn/longhorn#4994


innobead (approved):
LGTM. This part is easy to miss.

PR #1595, opened 2022-12-09
Reject the request moving a strict-local volume to another node

derekbit:
Longhorn/longhorn#5025


innobead (approved):
LGTM

PR #1598, opened 2022-12-09
Online expansion support

shuo-wu:
longhorn/longhorn#1674

### DIFF
--- a/controller/engine_controller.go
+++ b/controller/engine_controller.go
@@ -911,18 +911,6 @@ func (m *EngineMonitor) refresh(engine *longhorn.Engine) error {
 	if cliAPIVersion >= engineapi.MinCLIVersion {
 		// Cannot continue to start restoration if expansion is not complete
 		if engine.Spec.VolumeSize > engine.Status.CurrentSize {
-			// Cannot tolerate the rebuilding replicas during the expansion
-			hasRebuildingReplicas := false
-			for name, mode := range engine.Status.ReplicaModeMap {
-				if mode == longhorn.ReplicaModeWO {
-					engine.Status.ReplicaModeMap[name] = longhorn.ReplicaModeERR
-					hasRebuildingReplicas = true
-				}
-			}
-			if hasRebuildingReplicas {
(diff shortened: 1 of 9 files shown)


innobead (approved):
LGTM

innobead:
Bypassing the code factor check on purpose, but we need to revisit our code style in the future.

controller/engine_controller.go
```diff
@@ -911,18 +911,6 @@ func (m *EngineMonitor) refresh(engine *longhorn.Engine) error {
...
-			for name, mode := range engine.Status.ReplicaModeMap {
-				if mode == longhorn.ReplicaModeWO {
-					engine.Status.ReplicaModeMap[name] = longhorn.ReplicaModeERR
-					hasRebuildingReplicas = true
-				}
-			}
-			if hasRebuildingReplicas {
-				m.logger.Warn("Cannot contain any rebuilding replicas during engine expansion, will mark all rebuilding replicas as ERROR")
-				return nil
-			}
```
PhanLe1010, line 925: Why can we remove this logic?

PhanLe1010 (commented):
LGTM 

left a small comment

controller/engine_controller.go
shuo-wu: Previously the expansion is offline and should be fast, hence waiting for the rebuilding complete is not user-friendly. Now we can have the online expansion, which does not care if rebuilding exists or not. Maybe we need to continue disabling the rebuilding for offline expansion. But we can do it after your volume controller refactoring merge.

PhanLe1010: I see. 

But will there be a problem if we try to expand a replica when it is being rebuilt?

shuo-wu: Based on the integration test results, No.
For the replicas, the expansion is like snapshotting with some extra operations. If operation snapshotting works fine with rebuilding, we don't need to worry about the expansion part.

PhanLe1010: Sounds good. Thank you for the confirmation

PR #1599, opened 2022-12-10
Prevent creating multiple default instance managers on same node

PhanLe1010:
Prevent creating multiple default instance managers by hashing (node name + image name) to create instance manager name

longhorn/longhorn#3000


PhanLe1010:
https://github.com/longhorn/longhorn/issues/3000#issuecomment-1344944033

innobead (approved):

innobead (approved):

PR #1609, opened 2022-12-12
ci: exclude s390x PR check

innobead:
ref: https://github.com/longhorn/longhorn/issues/5040


yangchiu (commented):
LGTM

PR #1610, opened 2022-12-12
ci: run CI for release and master branches only

innobead:
ref: https://github.com/longhorn/longhorn/issues/5040


yangchiu (approved):
LGTM

PR #1611, opened 2022-12-13
fix(concurrent-backup-restore): decreased counter before restored

c3y1huang:
Decrease restore when:
- Restore complete (LastRestoredBackup updated)
- Failed to restore
- Engine monitor shut down

Added lock around limit check and acquiring counter to avoid counter increase above the limit when volumes backup restores concurrently.

Ref: https://github.com/longhorn/longhorn/issues/4558#issuecomment-1345721240

### DIFF
--- a/controller/engine_controller.go
+++ b/controller/engine_controller.go
@@ -84,7 +84,9 @@ type EngineController struct {
 	engineMonitorMap   map[string]chan struct{}
 
 	proxyConnCounter util.Counter
-	restoringCounter util.Counter
+
+	restoringCounter      util.Counter
+	restoringCounterMutex *sync.Mutex
 }
 
 type EngineMonitor struct {
@@ -105,7 +107,10 @@ type EngineMonitor struct {
 	monitorVoluntaryStopCh chan struct{}
 
 	proxyConnCounter util.Counter
-	restoringCounter util.Counter
+
+	restoringCounter         util.Counter
+	restoringCounterAcquired bool
+	restoringCounterMutex    *sync.Mutex
 }
 
 func NewEngineController(
@@ -138,8 +143,9 @@ func NewEngineController(
 		engineMonitorMutex: &sync.RWMutex{},
 		engineMonitorMap:   map[string]chan struct{}{},
 
-		proxyConnCounter: proxyConnCounter,
-		restoringCounter: util.NewAtomicCounter(),
+		proxyConnCounter:      proxyConnCounter,
+		restoringCounter:      util.NewAtomicCounter(),
+		restoringCounterMutex: &sync.Mutex{},
 	}
 	ec.instanceHandler = NewInstanceHandler(ds, ec, ec.eventRecorder)
 
@@ -654,6 +660,7 @@ func (ec *EngineController) startMonitoring(e *longhorn.Engine) {
 		controllerID:           ec.controllerID,
 		proxyConnCounter:       ec.proxyConnCounter,
 		restoringCounter:       ec.restoringCounter,
+		restoringCounterMutex:  ec.restoringCounterMutex,
 	}
 
 	ec.engineMonitorMutex.Lock()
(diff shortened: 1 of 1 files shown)


c3y1huang:
Testing
```
Concurrent Volume Backup Restore Per Node Limit = 1

Number of backups  = 9

Each backup size          = 5Gi
Each backup data size = 3.17 Gi

Each restored volume replica = 3
```

controller/engine_controller.go
```diff
@@ -972,22 +987,19 @@ func (m *EngineMonitor) refresh(engine *longhorn.Engine) error {
 		}
 		isDRVolume := volume.Status.IsStandby
 		if !isDRVolume {
-			if isReachedLimit, err := m.isReachedConcurrentVolumeBackupRestoreLimit(); err != nil {
-				m.logger.WithError(err).Warn("Failed to check concurrent volume backup restore limit")
-				return nil
-
-			} else if isReachedLimit {
-				m.logger.Debugf("Reached the concurrent volume backup restore limit of %v, retry later", types.SettingNameConcurrentBackupRestorePerNodeLimit)
+			err := m.acquireRestoringCounter(true)
```
innobead, line 990: Why not just add here? no need to change others. You can handle the error returned from the acquire method to decide if need to do defer call.

```go
			m.acquireRestoringCounter(true)
			defer m.acquireRestoringCounter(false)
```

innobead (changes requested):

controller/engine_controller.go
c3y1huang: We cannot use defer here because restoration could still be in progress when this method returns.

innobead: @c3y1huang is right because the restore task is asynchronous. I missed this part :+1:

controller/engine_controller.go
```diff
@@ -951,6 +961,11 @@ func (m *EngineMonitor) refresh(engine *longhorn.Engine) error {
 				return
 			}
 			engine = e
+			if existingEngine.Status.LastRestoredBackup != engine.Status.LastRestoredBackup {
+				if err := m.acquireRestoringCounter(false); err != nil {
+					m.logger.WithError(err).Error("Failed to unacquire restoring counter")
+				}
+			}
```
innobead, line 968: As discussed, need to take care of the non-successful restore to decrease the concurrent count to avoid starvation.

c3y1huang: Updated.

innobead (changes requested):

controller/engine_controller.go
```diff
@@ -952,6 +962,13 @@ func (m *EngineMonitor) refresh(engine *longhorn.Engine) error {
 			}
 			engine = e
 		}
+
+		isBackupRestoreCompleted := existingEngine.Status.LastRestoredBackup != engine.Status.LastRestoredBackup
+		if isBackupRestoreCompleted || isBackupRestoreFailed(engine.Status.RestoreStatus) {
+			if err := m.acquireRestoringCounter(false); err != nil {
+				m.logger.WithError(err).Error("Failed to unacquire restoring counter")
```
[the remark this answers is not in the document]
c3y1huang: Updated.

controller/engine_controller.go
```diff
@@ -972,22 +989,19 @@ func (m *EngineMonitor) refresh(engine *longhorn.Engine) error {
...
 			}
-
-			m.acquireRestoringCounter(true)
-			defer m.acquireRestoringCounter(false)
 		}
 
 		if err = m.restoreBackup(engine, rsMap, cliAPIVersion, engineClientProxy); err != nil {
 			m.restoreBackoff.DeleteEntry(engine.Name)
+			if err := m.acquireRestoringCounter(false); err != nil {
+				m.logger.WithError(err).Warn("Failed to unacquire restoring counter")
```
innobead, line 1003: Sorry, this should be Error.

innobead (changes requested):
Let's merge this two commits, then its ready to merge.

innobead (approved):

innobead:

PR #1612, opened 2022-12-13
fix(system-backup): should handle empty backup target

c3y1huang:
https://github.com/longhorn/longhorn/issues/5055

### DIFF
--- a/controller/system_backup_controller.go
+++ b/controller/system_backup_controller.go
@@ -387,7 +387,11 @@ func (c *SystemBackupController) reconcile(name string, backupTargetClient engin
 		var longhornVersion string
 		longhornVersion, err = getSystemBackupVersionExistInRemoteBackupTarget(systemBackup, backupTargetClient)
 		if err != nil {
-			return err
+			c.updateSystemBackupRecord(record,
+				systemBackupRecordTypeError, longhorn.SystemBackupStateError,
+				constant.EventReasonStart, err.Error(),
+			)
+			return nil
 		}
 
 		if longhornVersion != "" {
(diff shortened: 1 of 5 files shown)


derekbit (approved):
LGTM

BTW, the validating webhook already checks the inputs, so the UI doesn't any change. right?

c3y1huang:
> BTW, the validating webhook already checks the inputs, so the UI doesn't any change. right?

We still want an error message similar to the Backup page.
Ref: https://github.com/longhorn/longhorn/issues/5055#issuecomment-1350182125

/cc @smallteeths

controller/system_backup_controller.go
```diff
@@ -422,10 +426,7 @@ func (c *SystemBackupController) reconcile(name string, backupTargetClient engin
 		cleanupLocalSystemBackupFiles(tempBackupArchivePath, tempBackupDir, log)
 
 	case longhorn.SystemBackupStateDeleting:
-		err = c.cleanupRemoteSystemBackupFiles(systemBackup, backupTargetClient)
-		if err != nil {
-			return err
-		}
+		cleanupRemoteSystemBackupFiles(systemBackup, backupTargetClient, log)
```
innobead, line 429: So we will ignore the remote deletion if there is a connection issue with the backup target. It means the SystemBackup CR will be deleted, but later if the connection is backup, the SystemBackup will be resync from the target, and be created with ready state, right?

webhook/resources/systembackup/validator.go
```diff
@@ -0,0 +1,63 @@
...
+	if err != nil {
+		return werror.NewBadRequest(err.Error())
+	}
+
+	backupTarget, err := v.ds.GetBackupTargetRO(types.DefaultBackupTargetName)
+	if err != nil {
+		return werror.NewBadRequest(err.Error())
+	}
+
+	backupType, err := util.CheckBackupType(backupTarget.Spec.BackupTargetURL)
```
innobead, line 52: This is good, at least we check the backup target setting available or not.

innobead (approved):

controller/system_backup_controller.go
c3y1huang: Correct.

innobead:
@c3y1huang Need to backport to 1.4.x.

PR #1613, opened 2022-12-14
Support encrypted volumes in block mode

joshimoo:
Add support for using encrypted volumes in block mode directly

I use the staging path as a parent folder and bind mount device mapper
mapped encrypted volume device into a file inside of that folder.

This allows us to store additional metadata and gvie us more control over
our tear down for example since we are not sharing control with kubernetes
as is the case for the staging path itself.

We should be doing the same for the volumes used in fs mode, instead of
directly mounting into the staging path.

longhorn/longhorn#4883

### DIFF
--- a/csi/node_server.go
+++ b/csi/node_server.go
@@ -130,7 +130,7 @@ func (ns *NodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 	}
 
 	if volumeCapability.GetBlock() != nil {
-		devicePath := volume.Controllers[0].Endpoint
+		devicePath := filepath.Join(stagingPath, volumeID)
 		if err := ns.nodePublishBlockVolume(volumeID, devicePath, targetPath, mounter); err != nil {
 			return nil, err
 		}
@@ -261,6 +261,13 @@ func (ns *NodeServer) nodeStageMountVolume(volumeID, devicePath, targetPath, fsT
 	return nil
 }
 
+// nodeStageBlockVolume utilizes the stagingPath to create a volumeID file to bind mount the devicePath
+// this is valid since the csi plugin is in control of the staging path
+func (ns *NodeServer) nodeStageBlockVolume(volumeID, devicePath, stagingPath string, mounter mount.Interface) error {
+	stagingPath = filepath.Join(stagingPath, volumeID)
+	return ns.nodePublishBlockVolume(volumeID, devicePath, stagingPath, mounter)
+}
+
 func (ns *NodeServer) nodePublishBlockVolume(volumeID, devicePath, targetPath string, mounter mount.Interface) error {
 	// we ensure the parent directory exists and is valid
 	if _, err := ensureMountPoint(filepath.Dir(targetPath), mounter); err != nil {
@@ -352,13 +359,6 @@ func (ns *NodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
 		return nil, status.Errorf(codes.Aborted, "volume %s is not ready for workloads", volumeID)
 	}
 
-	devicePath := volume.Controllers[0].Endpoint
-
-	// do nothing for block devices, since they are handled by publish
-	if volumeCapability.GetBlock() != nil {
-		return &csi.NodeStageVolumeResponse{}, nil
-	}
-
 	if requiresSharedAccess(volume, volumeCapability) && !volume.Migratable {
 		if volume.AccessMode != string(longhorn.AccessModeReadWriteMany) {
 			return nil, status.Errorf(codes.FailedPrecondition, "volume %s requires shared access but is not marked for shared use", volumeID)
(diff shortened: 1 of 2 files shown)


csi/node_server.go
```diff
@@ -479,15 +489,25 @@ func (ns *NodeServer) NodeUnstageVolume(ctx context.Context, req *csi.NodeUnstag
...
 	}
 
+	// for block mode we use the staging path as parent, so we have to do additional cleanup of the subfolder/files
+	// we should transition the regular fs mounts to also use the same sub folder, this allows us to store additional
+	// metadata as well as do more forcefull removals since we no longer share the control of the staging_path with kubernetes
+	deviceFilePath := filepath.Join(targetPath, volumeID)
+	if err := cleanupMountPoint(deviceFilePath, mounter); err != nil {
+		return nil, status.Error(codes.Internal, fmt.Sprintf("failed to clean up volume %s device mount point %v error %v", volumeID, deviceFilePath, err))
+	}
+
```
innobead, line 506: Suggest doing this only for block volume

Should we do this unmount operation for the staged target mounted path before the above unmounting of the published target mounted path? For the nested mount paths, it should be better to unmount the inner first, then outer. WDYT?

innobead:
@shuo-wu this change makes the encrypted volume published in `NodeStageVolume`. Will this be helpful for online volume expansion?

PhanLe1010:
Sorry. Could you please explain a bit about the flow. I am understanding that:
1. K8s calls NodeStageVolume with stagingPath
2. Inside Longhorn CSI plugin, NodeStageVolume function mounts the crypto device under `stagingPath/volumeID`
3. K8s calls NodePublishVolume with `targetPath` and `stagingPath`
4. Inside Longhorn CSI plugin,  NodePublishVolume mount `stagingPath` to `targetPath` . However, the crypto device is at `stagingPath/volumeID` instead. Am I understanding it incorrectly?

csi/node_server.go
shuo-wu: The same concern:
Shouldn't we clean up the subfolder/files for the staging directory/path first (before unmounting the staging path)?

joshimoo:
@PhanLe1010 good catch you are correct, the fact that there was no changed code as part of the `NodePublishVolume` should have made it obvious to me that something went missing. the reason it looked okay in my quick final test was that we used devPath instead of stagingPath this mapped though to the original device instead of the opened crypto device.

As my final test I did a quick check to see if the pod would come up and the device file was created, I should have done a more elaborate verification. Will fix that and push an update :)

csi/node_server.go
joshimoo: 1. I agree I would have preferred to do this only for block mode unfortunately that information is not available as part of the `NodeUnstageVolume` call. Similiar issue as with the NodeExpandVolume for the encrypted device, this is the reason why I recommed we also transition to utilizing a subfolder of the staging path for the regular fs mode, since we can store additional metadata like access type / mode etc :)

2. this is on purpose, since the unmount of the parent is a no op for block mode, this is also important for backwards compatibility of the existing block devices. 

The cleanup of the child file/folder is a no op for fs mode, we can adjust this once we have switched to using a subfolder/file instead of the staging path directly.
```
	// CO owns the staging_path so we only unmount but not remove the path
	err := unmount(targetPath, mounter);

        // subfolder/file that we have full control over
	deviceFilePath := filepath.Join(targetPath, volumeID)
	cleanupMountPoint(deviceFilePath, mounter)
```

joshimoo:
I updated the PR with the fix for the NodePublishVolume but need to retest it tomorrow.

csi/node_server.go
shuo-wu: Got it. Can you add this to the comment?
> the unmount of the parent is a no op for block mode, this is also important for backwards compatibility of the existing block devices.

shuo-wu:
> @shuo-wu this change makes the encrypted volume published in `NodeStageVolume`. Will this be helpful for online volume expansion?

Just had a try manually, and it does not help. In NodeExpandVolume we still need the secret to resize the encrypto device.

csi/node_server.go
```diff
@@ -130,7 +130,7 @@ func (ns *NodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 	}
 
 	if volumeCapability.GetBlock() != nil {
-		devicePath := volume.Controllers[0].Endpoint
+		devicePath := filepath.Join(stagingPath, volumeID)
```
PhanLe1010, line 133: I am thinking about this senerio:
1. A block volume was staged with the old logic. So the device is available at `stagingPath`
2. The pod that uses that volume is restarted on the same node so K8s only call NodeUnpublishVolume but not NodeUnstageVolume.
3. Now K8s calls NodePublishVolume to republish the volume for newly restarted pod. This call will fail because it try to mount a device at `stagingPath/volumeID` but the device is at `stagingPath`

Do you think it is a valid case?

innobead: So `Node(Un)PublichVolume ` and  `Node(Un)StageVolume ` are not atomic? then it sounds like an issue because there is no guarantee about when CSI driver pods will be updated.

innobead:
@joshimoo Would you like to finish this PR?

innobead:
@derekbit will take over this PR.

PR #1614, opened 2022-12-14
style(system-restore): job/pod name prefix

c3y1huang:
https://github.com/longhorn/longhorn/issues/5057

### DIFF
--- a/controller/system_restore_controller.go
+++ b/controller/system_restore_controller.go
@@ -148,7 +148,7 @@ func (c *SystemRestoreController) handleErr(err error, key interface{}) {
 	}
 
 	utilruntime.HandleError(err)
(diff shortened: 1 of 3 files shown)


derekbit (approved):

innobead (commented):
Do we have any resource also have this kind of prefix for name?

innobead (approved):

innobead:
nit: I feel we should have a factor function in general for creating this kind of prefix name generation for different purposes. Need to review all pod name generation to refactor in total.

c3y1huang:
> nit: I feel we should have a factor function in general for creating this kind of prefix name generation for different purposes. Need to review all pod name generation to refactor in total.

https://github.com/longhorn/longhorn/issues/5060

innobead:
@c3y1huang need to backport to 1.4.x for 1.4.0 release.

PR #1615, opened 2022-12-14
feat(replica): RebuildRetryCount for disconnection

mantissahz:
Do not count the RebuildRetryCount if rebuilding reused replicas failed caused by network disconnection.
Need to collect more failed reasons for rebuilding replicas.

longhorn/longhorn#1923

Regression test: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/2680


innobead:
@mantissahz :point_up: the PR check (codefactor) failed, need to fix that.

mantissahz:
> the PR check (codefactor) failed, need to fix that.

Yes, modified, but it was generated by k8s/generate_code.sh (zz_generated.deepcopy.go), it seems like it should not be chagned?

innobead:
> > the PR check (codefactor) failed, need to fix that.
> 
> Yes, modified, but it was generated by k8s/generate_code.sh (zz_generated.deepcopy.go), it seems like it should not be chagned?

then it's embarrassing, we can just ignore this part we can't control.

controller/engine_controller.go
```diff
@@ -1756,6 +1763,47 @@ func (ec *EngineController) startRebuilding(e *longhorn.Engine, replica, addr st
 	return nil
 }
 
+// updateReplicaRebuildFailedRebuildStatus updates the rebuild status if replica rebuilding failed
+func updateReplicaRebuildFailedRebuildStatus(log *logrus.Entry, ds *datastore.DataStore, replica, errMsg string) error {
```
shuo-wu, line 1767: NIT: Rename this to something like `updateReplicaRebuildingFailureInfo`

controller/engine_controller.go
```diff
@@ -1710,6 +1712,11 @@ func (ec *EngineController) startRebuilding(e *longhorn.Engine, replica, addr st
 			// it is past the Backoff period, we'll try to mark the Replica as Failed and increase the Backoff period
 			// for the next failure.
 			if !ec.backoff.IsInBackOffSinceUpdate(e.Name, time.Now()) {
+				if err := updateReplicaRebuildFailedRebuildStatus(log, ec.ds, replica, replicaRebuildErrMsg); err != nil {
```
shuo-wu, line 1715: I am thinking if Longhorn should record the rebuilding error immediately after the failure

k8s/pkg/apis/longhorn/v1beta2/replica.go
```diff
@@ -42,6 +58,19 @@ type ReplicaStatus struct {
 	InstanceStatus `json:""`
 	// +optional
 	EvictionRequested bool `json:"evictionRequested"`
+	// +optional
+	// +nullable
+	RebuildStatus *ReplicaRebuildStatus `json:"rebuildStatus"`
+}
+
+// ReplicaRebuildStatus transiently record the status of replica rebuilding failed
+type ReplicaRebuildStatus struct {
```
shuo-wu, line 67: I just re-checked the original ticket. I think the init idea is that we can use a new field to record multiple kinds of replica failure reasons & messages including rebuilding failure. Can we introduce a more general field like `Condition` or `ErrorInfo` instead?

controller/volume_controller.go
```diff
@@ -1831,8 +1831,13 @@ func (vc *VolumeController) replenishReplicas(v *longhorn.Volume, e *longhorn.En
...
 				reusableFailedReplica.Spec.HealthyAt = ""
-				reusableFailedReplica.Spec.RebuildRetryCount++
-				vc.backoff.Next(reusableFailedReplica.Name, time.Now())
+
+				replicaRebuildStatus := reusableFailedReplica.Status.RebuildStatus
+				if replicaRebuildStatus == nil || replicaRebuildStatus.FailedReason != longhorn.ReplicaRebuildFailedReasonDisconnection || replicaRebuildStatus.RebuildDisconnectCount >= longhorn.ReplicaRebuildDisconnectCountMax {
+					reusableFailedReplica.Spec.RebuildRetryCount++
+					vc.backoff.Next(reusableFailedReplica.Name, time.Now())
+				}
+
```
shuo-wu, line 1840: The essential of this enhancement is, a failed replica on a down node should not get stuck in the reuse retry loop and even reach the max retry limitation. 
To handle it, can we simply refresh the backoff window when the node is still down and the previous backoff window is passed, rather than using another counter `RebuildDisconnectCount`?

mantissahz:
close this PR and I will create a new one after refactoring the code with introducing a more general field like `Condition`

PR #1621, opened 2022-12-15
fix(system-restore): do not overwrite existing volume, PV, PVC

c3y1huang:
https://github.com/longhorn/longhorn/issues/5064

### DIFF
--- a/controller/system_rollout_controller.go
+++ b/controller/system_rollout_controller.go
@@ -63,7 +63,7 @@ const (
 	SystemRolloutMsgUnpackedFmt         = "Unpacked %v"
 	SystemRolloutMsgCompleted           = "System rollout completed"
 	SystemRolloutMsgCreating            = "System rollout creating"
-	SystemRolloutMsgIgnoreItem          = "System rollout ignoring item: %v"
+	SystemRolloutMsgIgnoreItemFmt       = "System rollout ignoring item: %v"
 	SystemRolloutMsgUpdating            = "System rollout updating"
 )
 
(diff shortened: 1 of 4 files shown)


controller/system_rollout_controller.go
```diff
@@ -1222,10 +1222,8 @@ func (c *SystemRolloutController) restorePersistentVolumes() (err error) {
 			return err
 		}
 
-		restore.Spec.ClaimRef = exist.Spec.ClaimRef
 		if !reflect.DeepEqual(exist.Spec, restore.Spec) {
```
innobead, line 1225: There is an existing PV already, so it means we should not to restore it anyway no matter if the existing PV is the same as the restored PV or not. Then, why still need to check `if !reflect.DeepEqual(exist.Spec, restore.Spec)`? 

Also, it seems we will still annotate the PV with the remote backup URL and last restored time, but actually it's not restored. Do you think having another annotation to mark this resource has been processed but not restored. WDYT?


The same questions above should be applied to other resources like PVC and Volume here.

innobead (changes requested):

controller/system_rollout_controller.go
c3y1huang: > There is an existing PV already, so it means we should not to restore it anyway no matter if the existing PV is the same as the restored PV or not. Then, why still need to check `if !reflect.DeepEqual(exist.Spec, restore.Spec)`?

The intention is to output some logs. But it is not needed after introducing different annotations for the skipped resources. Removed.

>
> Also, it seems we will still annotate the PV with the remote backup URL and last restored time, but actually it's not restored. Do you think having another annotation to mark this resource has been processed but not restored. WDYT?

Fair. Introducing new annotations for the skipped resources. The annotations will replace the non-skipped annotations during the update.
- `longhorn.io/last-skipped-system-restored`
- `longhorn.io/last-skipped-system-restored-at`

Ex:
```
      longhorn.io/last-skipped-system-restored: s3://c3y1huang-s3@ap-southeast-1/backupstore/system-backu
ps/v1.5.0-dev/test1
      longhorn.io/last-skipped-system-restored-at: "2022-12-15T16:04:08Z"
```

innobead (changes requested):
LGTM, just a few minor requests.

innobead (approved):
I think it's good to go :slightly_smiling_face:

innobead:
@c3y1huang Please backport to 1.4.x. Thanks.

PR #1622, opened 2022-12-15
Support metrics for IOPS and bandwidth real time

derekbit:
https://github.com/longhorn/longhorn/issues/2406

Depends on https://github.com/longhorn/longhorn-engine/pull/794 and https://github.com/longhorn/longhorn-instance-manager/pull/158

Signed-off-by: Derek Su [derek.su@suse.com](mailto:derek.su@suse.com)


derekbit:
Case 1. Run `dd if=/dev/urandom of=30G bs=1M count=30000 oflag=direct` in a pod with a longhorn volume.

- IO status got from `iostat -cx 1 sda`
image

- IO status go from longhorn metrics
image


Case 2: Copy one 30GB file in the volume to the same volume.

image

metrics_collector/metrics_collector.go
```diff
@@ -19,20 +19,22 @@ import (
 )
 
 func InitMetricsCollectorSystem(logger logrus.FieldLogger, currentNodeID string, ds *datastore.DataStore, kubeconfigPath string, proxyConnCounter util.Counter) {
+	logger.Info("Initialize metrics collector system")
```
innobead, line 22: `Initializing ....`

metrics_collector/volume_collector.go
```diff
@@ -80,26 +160,63 @@ func (vc *VolumeCollector) Describe(ch chan<- *prometheus.Desc) {
...
+				engineClientProxy, err = vc.getEngineClientProxy(e)
+				if err == nil {
+					defer engineClientProxy.Close()
+
+					metrics, err = engineClientProxy.MetricsGet(e)
+					if err != nil {
+						vc.logger.WithError(err).Warnf("Failed to get metrics from volume %v from engine %v", e.Spec.VolumeName, e.Name)
+					}
+				} else {
+					vc.logger.WithError(err).Warnf("Failed to get engine for volume %v", v.Name)
```
innobead, line 191: `Failed to get engine proxy of %v for volume %v ...`

metrics_collector/metrics_collector.go
derekbit: Updated. Thank you.

metrics_collector/volume_collector.go
derekbit: Updated. Thank you.

engineapi/proxy_metrics.go
```diff
@@ -0,0 +1,13 @@
...
+import (
+	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
+)
+
+func (p *Proxy) MetricsGet(e *longhorn.Engine) (*Metrics, error) {
+	metrics, err := p.grpcClient.MetricsGet(p.DirectToURL(e))
+	if err != nil {
+		return nil, err
+	}
+	return (*Metrics)(metrics), err
```
innobead, line 12: the metrics returned from the grpc call is a metrics type definied in`github.com/longhorn/longhorn-instance-manager/pkg/client`, but here we define another Metrics type. They can be casted to each other w/o issues? Why not just use the type from instance manager?

metrics_collector/volume_collector.go
```diff
@@ -80,26 +160,63 @@ func (vc *VolumeCollector) Describe(ch chan<- *prometheus.Desc) {
...
 }
 
+func (vc *VolumeCollector) getEngineClientProxy(engine *longhorn.Engine) (c engineapi.EngineClientProxy, err error) {
+	engineCliClient, err := controller.GetBinaryClientForEngine(engine, &engineapi.EngineCollection{}, engine.Status.CurrentImage)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to get binary client for engine %v", engine.Name)
+	}
+
+	return engineapi.GetCompatibleClient(engine, engineCliClient, vc.ds, nil, vc.proxyConnCounter)
+}
```
innobead, line 218: nit: See several places using this function, so it's time to move to a common place.

innobead (changes requested):

engineapi/proxy_metrics.go
derekbit: It's a common pattern in longhorn-manager.

> Why not just use the type from instance manager?

Yeah, using the same type is a idea way. We will refactor them in the future.

innobead (approved):

innobead:
@Mergifyio backport v1.4.x

PR #1625, opened 2022-12-16
fix(system-restore): handle non-exist system backup config file

c3y1huang:
https://github.com/longhorn/longhorn/issues/5072


innobead:

PR #1627, opened 2022-12-16
fix(system-restore): add log when volume failed to create

c3y1huang:
https://github.com/longhorn/longhorn/issues/5086


innobead:
@c3y1huang I suggest changing the message to `add log when volume is unable to create`. WDYT

c3y1huang:
> @c3y1huang I suggest changing the commit message to `add log when volume is unable to create`. WDYT

Sounds good to me!

innobead (approved):

innobead:

PR #1630, opened 2022-12-17
NFS endpoint bracketed if ClusterIP is IPv6

chalharu:
Signed-off-by: Mitsuharu Seki <mitsu1986@gmail.com>


innobead:
@chalharu Thanks for the contribution. Have you verified this? What's your network stack, dual stack? So only changing this will make your RWX volume work correctly in IPv6 env? Because longhorn has not supported IPv6 or dual stack yet, I am quite curious about your setup.

chalharu:
I have verified that the RWX volume works in an IPv6 prefered dual-stack environment by this.
My kube-apiserver is set to --service-cluster-ip-range=[ipv6 range]::/108,10.32.0.0/24.

innobead:
cc @longhorn/qa Can we verify this PR?

shuo-wu (approved):

shuo-wu:
Not sure if we need to add this to the upgrade path. Probably NO since IPv6 NFS is not supported before this fix.

innobead (approved):

PR #1633, opened 2022-12-19
fix(pod-controller): log when associated resource not found

c3y1huang:
https://github.com/longhorn/longhorn/issues/5061#issuecomment-1357085967

### DIFF
--- a/controller/kubernetes_pod_controller.go
+++ b/controller/kubernetes_pod_controller.go
@@ -128,6 +128,10 @@ func (kc *KubernetesPodController) handleErr(err error, key interface{}) {
 	utilruntime.HandleError(err)
 }
 
+func getLoggerForPod(logger logrus.FieldLogger, pod *v1.Pod) *logrus.Entry {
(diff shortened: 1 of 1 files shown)


controller/kubernetes_pod_controller.go
```diff
@@ -128,6 +128,10 @@ func (kc *KubernetesPodController) handleErr(err error, key interface{}) {
 	utilruntime.HandleError(err)
 }
 
+func getLoggerForPod(logger logrus.FieldLogger, pod *v1.Pod) *logrus.Entry {
```
innobead, line 131: nit: I suggest we can refactor this kind of getting logger in a common place, like util.

innobead (approved):

innobead:
@c3y1huang Please backport to 1.4 and 1.3.

controller/kubernetes_pod_controller.go
c3y1huang: https://github.com/longhorn/longhorn/issues/5102

PR #1641, opened 2022-12-19
Refactor error handling and Add the API version verification for expansion

shuo-wu:
Longhorn/longhorn#1674


controller/engine_controller.go
```diff
@@ -1120,6 +1122,26 @@ func (ec *EngineController) syncSnapshotCRs(engine *longhorn.Engine) error {
...
+		return false, nil
+	}
+	if engine.Spec.VolumeSize < engine.Status.CurrentSize {
+		return false, fmt.Errorf("BUG: The expected size %v of engine %v should not be smaller than the current size %v", engine.Spec.VolumeSize, engine.Name, engine.Status.CurrentSize)
+	}
+	if cliAPIVersion < engineapi.MinCLIVersion {
+		return false, nil
+	}
+	if cliAPIVersion < 7 && engineapi.IsEndpointTGTBlockDev(engine.Status.Endpoint) {
+		return false, fmt.Errorf("cannot do online expansion for the old engine")
```
innobead, line 1139: Add engine info to the formatted error, at least name.

innobead, line 1139: Suggest seperate these conditions, because we will only support engine running with block device.

```go

if !engineapi.IsEndpointTGTBlockDev(engine.Status.Endpoint) {
...
}

if cliAPIVersion < 7 {
}
```

innobead (commented):
In general, LGTM. One minor comment needed to resolve.

controller/engine_controller.go
shuo-wu: > we will only support engine running with block device.
We will support expanding block device when the cli API version >= 7

shuo-wu:
The test step:
1. Deploy Longhorn v1.3.2
2. Create and attach multiple volumes.
3. Upgrade Longhorn with the version that contains this PR.
4. Trying to expand one old volume ==> There should be expansion errors in the engine CR
5. Upgrade the engine image for the old volumes. Then retry the online expansion. ==> This should fail. But this PR cannot fix this case since we won't check the instance manager version.

innobead (dismissed):
LGTM, but because we will update the instance manager, let's update go.mod as well.

innobead (approved):

innobead:
@Mergifyio backport v1.4.x

PR #1642, opened 2022-12-20
fix(system-restore): ignore PV, PVC due to missing Volume

c3y1huang:
https://github.com/longhorn/longhorn/issues/5086#issuecomment-1358725844

### DIFF
--- a/controller/system_rollout_controller.go
+++ b/controller/system_rollout_controller.go
@@ -132,6 +132,9 @@ type SystemRolloutController struct {
 	systemRestoreName    string
 	systemRestoreVersion string
 
+	systemRestoredAt  string
+	systemRestoredURL string
+
 	downloadPath string
 	engineImage  string
 
@@ -358,6 +361,16 @@ func (c *SystemRolloutController) systemRollout() error {
 		return err
 	}
 	c.systemRestore = systemRestore
+
+	if c.systemRestoredAt == "" {
+		c.systemRestoredAt = time.Now().UTC().Format(time.RFC3339)
+	}
+	if c.systemRestoredURL == "" {
+		c.systemRestoredURL, err = systembackupstore.GetSystemBackupURL(c.systemRestore.Spec.SystemBackup, c.systemRestoreVersion, c.backupTargetURL)
+		if err != nil {
+			return err
+		}
+	}
 
 	record := &systemRolloutRecord{}
 	existingSystemRestore := c.systemRestore.DeepCopy()
@@ -801,12 +814,7 @@ func (c *SystemRolloutController) restoreClusterRoles() (err error) {
 			}
 		}
 
-		systemBackupURL, err := systembackupstore.GetSystemBackupURL(c.systemRestore.Spec.SystemBackup, c.systemRestoreVersion, c.backupTargetURL)
-		if err != nil {
-			return err
-		}
-
-		if err = tagLonghornLastSystemRestoreAnnotation(systemBackupURL, time.Now().UTC().Format(time.RFC3339), exist, false); err != nil {
+		if err = c.tagLonghornLastSystemRestoreAnnotation(exist, false); err != nil {
 			return err
 		}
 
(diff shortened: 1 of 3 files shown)


controller/system_rollout_controller.go
```diff
@@ -1719,6 +1721,9 @@ func (c *SystemRolloutController) restoreVolumes() (err error) {
 
 				reason := fmt.Sprintf(constant.EventReasonFailedCreatingFmt, types.LonghornKindVolume, restore.Name)
 				c.eventRecorder.Event(c.systemRestore, corev1.EventTypeWarning, reason, message)
+
+				c.ignorePersistenVolumeDueToMissingVolume(&restore)
+				c.ignorePersistenVolumeClaimDueToMissingVolume(&restore)
```
innobead, line 1726: Why not just apply the skipped labels to those skipped resources, then they can be bypassed during restoring PV/PVC?

```go
		annos[types.GetLastSkippedSystemRestoreLabelKey()] = systemRestoredURL
		annos[types.GetLastSkippedSystemRestoreAtLabelKey()] = systemRestoredAt
```

innobead (changes requested):

controller/system_rollout_controller.go
c3y1huang: Not sure if that is necessary.
1. The PV and PVC will not get created so the skipped annotation will not be seen by the users.
1. There will be 2 more arguments(systemRestoredURL, systemRestoredAt) to pass through to `ignorePersistenVolume*`  functions.
1. The backup PV and PVC resources might already have the annotations so we need to check to see if those are from the current rollout. Or, we need to remove them during the unpacking.

c3y1huang: After discussing with @innobead , we can still apply the skipped annotations to use as filtering for restoring resources.

controller/system_rollout_controller.go
```diff
@@ -1215,9 +1202,16 @@ func (c *SystemRolloutController) restorePersistentVolumeClaims() (err error) {
 	}
 
 	for _, restore := range c.persistentVolumeClaimList.Items {
+		isSkippedByDependency, err := c.hasSystemRestoredAnnotation(&restore, true)
```
innobead, line 1205: Suggest renaming `hasSystemRestoredAnnotation` -> `isResourceRestoreRequired` as below. (no need to pass a bool param, because this will make the code not easy to read. )

```go

restoreRequired, err := c.isResourceRestoreRequired(&restore)
```

controller/system_rollout_controller.go
```diff
@@ -1755,18 +1776,26 @@ func tagLonghornLastSystemRestoreAnnotation(systemRestoredURL, systemRestoredAt
 		annos = map[string]string{}
 	}
 
+	hasAnno, err := c.hasSystemRestoredAnnotation(obj, false)
```
innobead, line 1779: Why having this case? the resources have been restored before the system backup? Should we just override?

innobead (changes requested):

controller/system_rollout_controller.go
c3y1huang: case:
1. Resource CR-1 gets created (tagged) and CR-2 failed (un-tagged).
1. Controller tries to restore resources again on the ticker.

Before change:
1. Resource CR-1 gets tagged again with skipped annotation because it's not created and the backup CR-1 and existing CR-1 are identical.
1. Resource CR-1 has the wrong annotation.

After change:
1. Resource CR-1 will not get tagged again.
1. Resource CR-1 has the correct annotation.

innobead (changes requested):
In general, LGTM. Just two comments left.

controller/system_rollout_controller.go
```diff
@@ -1755,18 +1991,28 @@ func tagLonghornLastSystemRestoreAnnotation(systemRestoredURL, systemRestoredAt
...
 	if isSkipped {
 		delete(annos, types.GetLastSystemRestoreLabelKey())
 		delete(annos, types.GetLastSystemRestoreAtLabelKey())
 
-		annos[types.GetLastSkippedSystemRestoreLabelKey()] = systemRestoredURL
-		annos[types.GetLastSkippedSystemRestoreAtLabelKey()] = systemRestoredAt
+		annos[types.GetLastSkippedSystemRestoreLabelKey()] = c.systemRestoredURL
+		annos[types.GetLastSkippedSystemRestoreAtLabelKey()] = c.systemRestoredAt
+
+		log.Debug(message)
```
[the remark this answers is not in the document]
c3y1huang: Updated.

innobead (approved):

PR #1650, opened 2022-12-22
Support trim operation on a volume with multiple mountpoints

derekbit:
Longhorn/longhorn#836

### DIFF
--- a/util/util.go
+++ b/util/util.go
@@ -912,17 +912,20 @@ func TrimFilesystem(volumeName string) error {
 		return err
 	}
 
-	mountOutput, err := nsExec.Execute("bash", []string{"-c", fmt.Sprintf("mount | grep /dev/longhorn/%s | awk '{print $3}'", volumeName)})
+	mountOutput, err := nsExec.Execute("bash", []string{"-c", fmt.Sprintf("cat /proc/mounts | grep /dev/longhorn/%s | awk '{print $2}'", volumeName)})
 	if err != nil {
 		return fmt.Errorf("cannot find volume %v mount info on host: %v", volumeName, err)
 	}
 
 	mountList := strings.Split(strings.TrimSpace(mountOutput), "\n")
-	if len(mountList) != 1 {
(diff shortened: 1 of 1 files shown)


innobead (approved):
LGTM

innobead:
@Mergifyio backport v1.4.x

util/util.go
```diff
@@ -912,17 +912,20 @@ func TrimFilesystem(volumeName string) error {
...
 		return fmt.Errorf("cannot find volume %v mount info on host: %v", volumeName, err)
 	}
 
 	mountList := strings.Split(strings.TrimSpace(mountOutput), "\n")
-	if len(mountList) != 1 {
-		return fmt.Errorf("zero or multiple mount points for volume %s: %+v", volumeName, mountList)
+
+	var mountpoint string
+	for _, m := range mountList {
+		mountpoint = m
```
shuo-wu, line 924: Should we check if this mount point is the valid one?

derekbit: Would be better. But how to check it is a valid one?

shuo-wu: Function `ensureMountPoint` in the pkg `csi` already implements that

innobead: `ensureMountPoint` will try to clean up the corrupted mount point so it's not just a validation behavior, so probably we need to have a separate one for `mount.IsCorruptedMnt(err)` only.

https://github.com/longhorn/longhorn-manager/blob/db24c350211c524eab04afc702fdc0497aabeed6/csi/util.go#L216-L220

derekbit: I'm thinking if we need to check if the mountpoint is corrupted. If it is corrupted, the fstrim should return error.

shuo-wu: Maybe we can have a try.

shuo-wu: OK. It shows error `fstrim: vol: FITRIM ioctl failed: Input/output error`.

Here it's better to figure out which mount point is the valid one then apply `fstrim`

derekbit: Then, let me add a check.

PR #1652, opened 2022-12-26
Check mountpoint is healthy before trimming the volume

derekbit:
Longhorn/longhorn#836


innobead (approved):

innobead:
@Mergifyio backport v1.4.x

PR #1656, opened 2022-12-27
controller: Set expansion error when the info is not empty

shuo-wu:
Longhorn/longhorn#1674


derekbit (approved):

innobead (approved):

innobead:
@Mergifyio backport v1.4.x

PR #1658, opened 2022-12-27
fix(system-restore): replica crash during EngineImage update

c3y1huang:
https://github.com/longhorn/longhorn/issues/5134#issuecomment-1365604838

### DIFF
--- a/controller/system_rollout_controller.go
+++ b/controller/system_rollout_controller.go
@@ -427,6 +427,8 @@ func (c *SystemRolloutController) systemRollout() error {
 
 	case longhorn.SystemRestoreStateRestoring:
 		c.restore(types.APIExtensionsKindCustomResourceDefinitionList, c.restoreCustomResourceDefinitions, log)
+		c.restore(types.KubernetesKindDaemonSetList, c.restoreDaemonSets, log)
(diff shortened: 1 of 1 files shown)


controller/system_rollout_controller.go
```diff
@@ -427,6 +427,8 @@ func (c *SystemRolloutController) systemRollout() error {
 
 	case longhorn.SystemRestoreStateRestoring:
 		c.restore(types.APIExtensionsKindCustomResourceDefinitionList, c.restoreCustomResourceDefinitions, log)
+		c.restore(types.KubernetesKindDaemonSetList, c.restoreDaemonSets, log)
```
innobead, line 430: Do you mean we need to restore longhorn-manager DS and engine image first before other resources?

note: The resource restore order should be the same as installation.

innobead (approved):

controller/system_rollout_controller.go
innobead: After discussing with @c3y1huang , the mandatory DSs need to start first before other resources to avoid this kind of error like engine binary is not ready on nodes. Those are including longhorn manager, engine image, and CSI plugin.

c3y1huang:
@Mergifyio backport v1.4.x

PR #1660, opened 2022-12-27
Allow trimming encrypted devices via UI

shuo-wu:
Users need to enable flag `discards` for the encrypted devices manually first:
https://wiki.archlinux.org/title/Dm-crypt/Specialties#Discard/TRIM_support_for_solid_state_drives_(SSD)

Longhorn/longhorn#836

### DIFF
--- a/manager/volume.go
+++ b/manager/volume.go
@@ -669,7 +669,7 @@ func (m *VolumeManager) TrimFilesystem(name string) (v *longhorn.Volume, err err
 		return nil, fmt.Errorf("volume frontend is disabled")
 	}
 
-	return v, util.TrimFilesystem(name)
+	return v, util.TrimFilesystem(name, v.Spec.Encrypted)
 }
 
 func (m *VolumeManager) AddVolumeRecurringJob(volumeName string, name string, isGroup bool) (volumeRecurringJob map[string]*longhorn.VolumeRecurringJob, err error) {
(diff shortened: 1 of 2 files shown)


derekbit:
Doe encrypted block devices need `--allow-discards` while creating them?

shuo-wu:
> Do encrypted block devices need `--allow-discards` while creating them?

Based on [the doc](https://wiki.archlinux.org/title/Dm-crypt/Specialties#Discard/TRIM_support_for_solid_state_drives_(SSD)) mentioned in [the website PR](https://github.com/longhorn/website/pull/651), there are some security concerns hence device mapper won't enable it by default.

derekbit:
> > Do encrypted block devices need `--allow-discards` while creating them?
> 
> Based on [the doc](https://wiki.archlinux.org/title/Dm-crypt/Specialties#Discard/TRIM_support_for_solid_state_drives_(SSD)) mentioned in [the website PR](https://github.com/longhorn/website/pull/651), there are some security concerns hence device mapper won't enable it by default. Currently, we need to follow the same idea.

Make sense. Should we check if the encrypted device supports trim before executing `nsenter trim...`? Or, will the command `nsenter trim...` return error if `--allow-discard` is set?

shuo-wu:
If the flag is not enabled, `fstrim` will error out. We don't need to complicate this function.

derekbit (approved):
LGTM

innobead (approved):
LGTM

innobead:
@Mergifyio backport v1.4.x

PR #1662, opened 2022-12-28
fix(system-restore): volume name different to PV

c3y1huang:
https://github.com/longhorn/longhorn/issues/5157


innobead (approved):

innobead:
@Mergifyio backport v1.4.x

PR #1664, opened 2022-12-28
fix(system-restore): blocking uninstall

c3y1huang:
https://github.com/longhorn/longhorn/issues/5133


innobead (approved):
LGTM.

A breaking change is to remove the finalizer mechanism from SystemRestore, but only delete a SystemRestore resource until the dependent resources are deleted (like rollout job). Right?

c3y1huang:
> A breaking change is to remove the finalizer mechanism from SystemRestore, but only delete a SystemRestore resource until the dependent resources are deleted (like rollout job). Right?

Deleting the SystemRestore will also delete the dependent rollout job by `ownerReference`.

innobead:
@Mergifyio backport v1.4.x

PR #1666, opened 2022-12-29
Fix expanding volume detachment failure when node is gone

derekbit:
https://github.com/longhorn/longhorn/issues/1674

Fix the [node deletion - volume attached to down node, so the expansion stuck](https://github.com/longhorn/longhorn/issues/1674#issuecomment-1367068900)

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -2853,6 +2853,13 @@ func (vc *VolumeController) checkForAutoDetachment(v *longhorn.Volume, e *longho
 	}
 
 	if v.Status.ExpansionRequired {
+		_, err := vc.ds.GetNodeRO(v.Status.CurrentNodeID)
+		if err == nil || (err != nil && !datastore.ErrorIsNotFound(err)) {
+			return nil
+		}
+
+		log.Infof("Preparing to do auto-detachment of expanding volume %v, because the node %v is unavailable", v.Name, v.Status.CurrentNodeID)
+		v.Status.CurrentNodeID = ""
 		return nil
 	}
 


controller/volume_controller.go
```diff
@@ -2853,7 +2853,16 @@ func (vc *VolumeController) checkForAutoDetachment(v *longhorn.Volume, e *longho
...
 
 	if v.Status.ExpansionRequired {
-		return nil
+		if v.Status.CurrentNodeID == "" {
+			return nil
+		}
+
+		_, err := vc.ds.GetNodeRO(v.Status.CurrentNodeID)
+		if err == nil || (err != nil && !datastore.ErrorIsNotFound(err)) {
+			return nil
```
innobead, line 2862: Can we add a debug log here?

controller/volume_controller.go
```diff
@@ -2853,7 +2853,16 @@ func (vc *VolumeController) checkForAutoDetachment(v *longhorn.Volume, e *longho
...
+		}
+
+		v.Status.CurrentNodeID = ""
```
innobead, line 2865: Do we need to return after this setting?

derekbit: Updated. Thank you,

controller/volume_controller.go
```diff
@@ -2853,7 +2853,21 @@ func (vc *VolumeController) checkForAutoDetachment(v *longhorn.Volume, e *longho
 	}
 
 	if v.Status.ExpansionRequired {
-		return nil
+		if v.Status.CurrentNodeID != "" {
+			_, err := vc.ds.GetNodeRO(v.Status.CurrentNodeID)
+			if err == nil {
+				log.Debugf("Expansion of volume %v is required", v.Name)
```
innobead, line 2859: Let's remove this part because it would be very verbose and often, but just log the case needed to do auto detachment.

controller/volume_controller.go
```diff
@@ -2853,7 +2853,21 @@ func (vc *VolumeController) checkForAutoDetachment(v *longhorn.Volume, e *longho
...
+				return nil
+			}
+
+			if err != nil && !datastore.ErrorIsNotFound(err) {
+				log.WithError(err).Debugf("Failed to get node %v for checking volume %v detachment", v.Status.CurrentNodeID, v.Name)
```
innobead, line 2864: The same, remove this log.

controller/volume_controller.go
```diff
@@ -2853,7 +2853,21 @@ func (vc *VolumeController) checkForAutoDetachment(v *longhorn.Volume, e *longho
...
+				return nil
+			}
+
+			v.Status.CurrentNodeID = ""
```
innobead, line 2868: Add a log before setting the `CurrentNodeID` to empty, something like `Preparing to do auto detachment of volume xx, because the node xx is unavailable`.

innobead (changes requested):

controller/volume_controller.go
derekbit: Updated. Thank you.

controller/volume_controller.go
derekbit: Updated. Thank you.

controller/volume_controller.go
derekbit: Updated. Thank you.

controller/volume_controller.go
```diff
@@ -2853,6 +2853,13 @@ func (vc *VolumeController) checkForAutoDetachment(v *longhorn.Volume, e *longho
 	}
 
 	if v.Status.ExpansionRequired {
+		_, err := vc.ds.GetNodeRO(v.Status.CurrentNodeID)
+		if err == nil || (err != nil && !datastore.ErrorIsNotFound(err)) {
+			return nil
+		}
+
+		log.Infof("Preparing to do auto-detachment of volume %v, because the node %v is unavailable", v.Name, v.Status.CurrentNodeID)
```
innobead, line 2861:
```suggestion
		log.Infof("Preparing to do auto-detachment of expanding volume %v, because the node %v is unavailable", v.Name, v.Status.CurrentNodeID)
```

derekbit: Updated. Thank you.

innobead (approved):

innobead:
@Mergifyio backport v1.4.x

PR #1670, opened 2023-01-03
Improve backup performance with faster compression algorithm and multiple threads

derekbit:
longhorn/longhorn#5189


derekbit:
Dependencies are updated.

webhook/resources/volume/mutator.go
```diff
@@ -212,6 +212,14 @@ func (v *volumeMutator) Create(request *admission.Request, newObj runtime.Object
 	}
 	patchOps = append(patchOps, fmt.Sprintf(`{"op": "replace", "path": "/spec/engineImage", "value": "%s"}`, defaultEngineImage))
 
+	if volume.Spec.BackupCompressionMethod == "" {
+		defaultCompressionMethod, _ := v.ds.GetSettingValueExisted(types.SettingNameBackupCompressionMethod)
```
innobead, line 216: Do we have a validation for setting value?

api/model.go
```diff
@@ -1272,6 +1290,7 @@ func toVolumeResource(v *longhorn.Volume, ves []*longhorn.Engine, vrs []*longhor
 			actions["updateReplicaAutoBalance"] = struct{}{}
 			actions["updateUnmapMarkSnapChainRemoved"] = struct{}{}
 			actions["updateSnapshotDataIntegrity"] = struct{}{}
+			actions["updateBackupCompressionMethod"] = struct{}{}
```
innobead, line 1293: Just double confirm: Because the compression method will be saved for each backup, backups with different compression method before and after, those backups should be able to restore correctly, right?

innobead (commented):
LGTM, just a few comments.

webhook/resources/volume/mutator.go
derekbit: Added the validation. Thank you.

types/setting.go
```diff
@@ -1026,6 +1035,45 @@ var (
 		ReadOnly:    false,
 		Default:     "30",
 	}
+
+	SettingDefinitionBackupCompressionMethod = SettingDefinition{
+		DisplayName: "Backup Compression Method",
+		Description: "This setting allows users to specify backup compression method.\n\n" +
+			"Available options are: \n\n" +
+			"- **none**: Disable the compression method. Suitable for multimedia data such as encoded images and videos. \n\n" +
+			"- **lz4**: Suitable for text files. \n\n" +
```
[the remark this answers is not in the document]
derekbit: Updated. Thank you.

api/model.go
derekbit: For a volume, the backups use the same compression method in the current implementation. Changing of the volume's compression method is not supported and rejected by the validator.

https://github.com/longhorn/longhorn-manager/blob/5ab00441dd915c94f0ce10ca1a7c74212127679c/webhook/resources/volume/validator.go#L149

We can improve it if users have the request.

innobead (approved):

shuo-wu (commented):
I don't know why the review result was not submitted last time. But this PR LGTM anyway.

k8s/crds.yaml
```diff
@@ -3233,6 +3236,12 @@ spec:
                 type: string
               backingImage:
                 type: string
+              backupCompressionMethod:
+                enum:
```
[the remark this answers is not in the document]
derekbit: I planed to add `ignored` before. But, for a newly created volume, the compressionMethod is mutated using the global setting backup-compression-method. Do you think we still need `ignored`?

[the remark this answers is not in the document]
shuo-wu: Wait... Is this field still immutable after one backup creation? If YES, I think it's better to add it. For example, for all existing volumes, the compression method should be `ignored` and the default global setting is `gzip`. After the Longhorn upgrade, users can directly modify the global setting only to apply the change against all volumes.

PR #1671, opened 2023-01-03
recovery backend: get configmap from kubernetes rather than cache

derekbit:
Longhorn/longhorn#5183

### DIFF
--- a/datastore/uncached.go
+++ b/datastore/uncached.go
@@ -369,3 +369,8 @@ func (s *DataStore) GetAllLonghornCustomResourceDefinitions() (runtime.Object, e
 	crdList.Items = longhornCRDs
 	return crdList, nil
 }
+
+// GetConfigMapWithoutCache return a new ConfigMap via Kubernetes client object for the given namespace and name
+func (s *DataStore) GetConfigMapWithoutCache(namespace, name string) (*corev1.ConfigMap, error) {
+	return s.kubeClient.CoreV1().ConfigMaps(s.namespace).Get(context.TODO(), name, metav1.GetOptions{})
+}
--- a/recovery_backend/backend/configmap.go
+++ b/recovery_backend/backend/configmap.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/longhorn/longhorn-manager/datastore"
 	"github.com/longhorn/longhorn-manager/types"
+	"github.com/longhorn/longhorn-manager/util"
 )
 
 type ClientID string
@@ -58,7 +59,7 @@ func (rb *RecoveryBackend) CreateConfigMap(hostname, version string) error {
 			return errors.Wrapf(err, "failed to create a configmap %v", configMapName)
 		}
 
-		cm, err := rb.Datastore.GetConfigMap(rb.Namespace, configMapName)
+		cm, err := rb.Datastore.GetConfigMapWithoutCache(rb.Namespace, configMapName)
 		if err != nil {
 			return errors.Wrapf(err, "failed to get the configmap %v", configMapName)
 		}
@@ -74,22 +75,25 @@ func (rb *RecoveryBackend) CreateConfigMap(hostname, version string) error {
 
 func (rb *RecoveryBackend) EndGrace(hostname, version string) error {
 	configMapName := types.GetConfigMapNameFromHostname(hostname)
-	cm, err := rb.Datastore.GetConfigMap(rb.Namespace, configMapName)
-	if err != nil {
-		return errors.Wrapf(err, "failed to get the configmap %v", configMapName)
-	}
 
-	oldVersion := cm.Annotations["version"]
+	_, err := util.RetryOnConflictCause(func() (interface{}, error) {
+		cm, err := rb.Datastore.GetConfigMapWithoutCache(rb.Namespace, configMapName)
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to get the configmap %v", configMapName)
+		}
 
-	// Switch to latest version
-	annotations := map[string]string{
-		"version": version,
-	}
-	cm.ObjectMeta.Annotations = annotations
+		oldVersion := cm.Annotations["version"]
 
-	// Remove old data
-	delete(cm.Data, oldVersion)
-	_, err = rb.Datastore.UpdateConfigMap(cm)
+		// Switch to latest version
+		annotations := map[string]string{
+			"version": version,
+		}
+		cm.ObjectMeta.Annotations = annotations
+
+		// Remove old data
+		delete(cm.Data, oldVersion)
+		return rb.Datastore.UpdateConfigMap(cm)
+	})
 
 	return err
 }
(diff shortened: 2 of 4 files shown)


shuo-wu (commented):
Shouldn't we apply [this kind of verification](https://github.com/longhorn/longhorn-manager/blob/d20e1ca6e04b229b9823c1a941d865929007874c/datastore/longhorn.go#L580-L589) for functions in `datastore/kubernetes.go`?

derekbit:
> Shouldn't we apply [this kind of verification](https://github.com/longhorn/longhorn-manager/blob/d20e1ca6e04b229b9823c1a941d865929007874c/datastore/longhorn.go#L580-L589) for functions in `datastore/kubernetes.go`?

Anyway, I added the verification for avoiding any uncertainty.

shuo-wu:
> But if we directly ceate and get the configmap by the kubernetes client api, is the verification necessary?

No. But the key point is, using k8s client API for the get function is too expensive...

shuo-wu:
After adding `VerifyCreation`, do we still need flag `fromLister`?

derekbit:
> After adding `VerifyCreation`, do we still need flag `fromLister`?

Yes. the update still encounter the same issue. I'm thinking which solution is better. Retry if getting afte rupdateing failed or direct get the configmap by k8s client ranter than lister. WDYT? @shuo-wu

derekbit:
> 

Exactly. But for the client connection, only rwx volume `creation` and `attch` operations update the configmap, so the cost is not actually a concern.

derekbit:
> > But the update still encounter the same issue.
> 
> I see. `verifyUpdate` just guarantees the updated CR resource version is **no less than** the previous rather than greater than the previous version. I am not sure if we improve this. If No, then we apply the current fix.

The `update` might encounters conflict error. But the HandleError in api/route.go cannot actually retry correctly.
I add `RetryOnConflictCause` logic in the confimap operations.
https://github.com/longhorn/longhorn/issues/5183#issuecomment-1368352139

recovery_backend/backend/configmap.go
```diff
@@ -103,34 +107,36 @@ func (rb *RecoveryBackend) AddClientID(hostname, version string, clientID Client
...
+		cm, err := rb.Datastore.GetConfigMap(rb.Namespace, configMapName, false)
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to get configmap %v", configMapName)
+		}
 
-	dataStr := cm.Data[version]
-	if dataStr == "" {
-		return errors.Wrapf(err, "failed to get data from recovery backend %v (version %v)", configMapName, version)
-	}
+		dataStr := cm.Data[version]
```
shuo-wu, line 116: NIT: Why not using format `dataStr, ok := cm.Data[version]` here?

datastore/kubernetes.go
```diff
@@ -624,13 +625,23 @@ func (s *DataStore) UpdateConfigMap(configMap *corev1.ConfigMap) (*corev1.Config
...
-func (s *DataStore) GetConfigMapRO(namespace, name string) (*corev1.ConfigMap, error) {
-	return s.cfmLister.ConfigMaps(namespace).Get(name)
+func (s *DataStore) GetConfigMapRO(namespace, name string, fromLister bool) (*corev1.ConfigMap, error) {
+	if fromLister {
+		return s.cfmLister.ConfigMaps(namespace).Get(name)
+	}
+	return s.kubeClient.CoreV1().ConfigMaps(s.namespace).Get(context.TODO(), name, metav1.GetOptions{})
 }
 
 // GetConfigMap return a new ConfigMap object for the given namespace and name
```
shuo-wu, line 635: NIT: mentioning why we introduce flag `fromLister`

recovery_backend/backend/configmap.go
```diff
@@ -69,12 +70,15 @@ func (rb *RecoveryBackend) CreateConfigMap(hostname, version string) error {
 		}
 	}
 
-	return nil
+	_, err = datastore.VerifyCreation(configMapName, "configmap", func(name string) (runtime.Object, error) {
+		return rb.Datastore.GetConfigMapRO(rb.Namespace, name, false)
+	})
+	return err
```
shuo-wu, line 77: Why execute this outside of `Datastore.CreateConfigMap`?

controller/share_manager_controller.go
```diff
@@ -740,8 +741,22 @@ func (c *ShareManagerController) createShareManagerPod(sm *longhorn.ShareManager
 	if err != nil {
 		return nil, errors.Wrapf(err, "failed to create pod for share manager %v", sm.Name)
 	}
+
+	obj, err := datastore.VerifyCreation(pod.Name, "pod", func(name string) (runtime.Object, error) {
+		return c.ds.GetPodRO(pod.Namespace, name)
+	})
```
shuo-wu, line 747: Similarly, why execute this outside of `ds.CreatePod`?

shuo-wu (commented):
In the last commit, which socket buffer do you mean?

controller/share_manager_controller.go
derekbit: I updated the PR. Can you re-check it? Thank you.

recovery_backend/backend/configmap.go
derekbit: It follows the pattern in datastore/longhorn.go. Create the resource and check the resource is ready in the lister.

derekbit:
> In the last commit, which socket buffer do you mean?

Sorry. I mean the `ApiContext.r.Body` not a sock buffer.

```
func (a *ApiContext) Read(obj interface{}) error {
	decoder := json.NewDecoder(a.r.Body)
	return decoder.Decode(obj)
}
```

recovery_backend/backend/configmap.go
derekbit: `""` is also a valid value for `dataStr, ok := cm.Data[version]`.

datastore/kubernetes.go
derekbit: Added. Thank you.

recovery_backend/backend/configmap.go
shuo-wu: I mean, can we move this verification into `ds.CreateConfigMap()`?

controller/share_manager_controller.go
```diff
@@ -740,8 +742,42 @@ func (c *ShareManagerController) createShareManagerPod(sm *longhorn.ShareManager
...
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to verify creation of pod %v", pod.Name)
+	}
+
 	getLoggerForShareManager(c.logger, sm).WithField("pod", pod.Name).Info("Created pod for share manager")
-	return pod, nil
+	return ret, nil
+}
+
+func (c *ShareManagerController) verifyPodSpecNodeName(podName string) (*corev1.Pod, error) {
```
shuo-wu, line 755: Similarly, do we need to put this into `ds.CreatePod()`?

controller/share_manager_controller.go
derekbit: I remove this commit and it will be handled in https://github.com/longhorn/longhorn/issues/5233.

derekbit:
> Got it. Nice finding! The [router retry mechanism](https://github.com/longhorn/longhorn-manager/blob/25c572365ee242883c65524f5f0d1dd6af2eddae/recovery_backend/server/router.go#L14) will lead to repeatedly read from the request body, which actually does not work correctly. So you have to rely on `util.RetryOnConflictCause` instead.
> 
> It seems that this issue can be triggered by other APIs. Can you create a separate ticket to track it?

Track in https://github.com/longhorn/longhorn/issues/5259. Thank you.

controller/share_manager_controller.go
derekbit: Remove this commit and will be handled in https://github.com/longhorn/longhorn/issues/5233.
Let the PR focus on the configmap operations.

recovery_backend/backend/configmap.go
derekbit: Moving the verification in CreateConfigMap is a bit of weird.
Because the CreateConfigMap updates the k8s databese directly, but the verification is verifying the data in the lister...I perfer to verify it outside of CreateConfigMap.

datastore/longhorn.go
```diff
@@ -2936,7 +2936,7 @@ func (s *DataStore) UpdateInstanceManagerStatus(im *longhorn.InstanceManager) (*
 	return obj, nil
 }
 
-func verifyCreation(name, kind string, getMethod func(name string) (runtime.Object, error)) (runtime.Object, error) {
+func VerifyCreation(name, kind string, getMethod func(name string) (runtime.Object, error)) (runtime.Object, error) {
 	// WORKAROUND: The immedidate read after object's creation can fail.
```
innobead, line 2940: Why need to export this function? It seems only used in datastore pkg only.

datastore/kubernetes.go
```diff
@@ -624,13 +625,27 @@ func (s *DataStore) UpdateConfigMap(configMap *corev1.ConfigMap) (*corev1.Config
 // GetConfigMapRO gets ConfigMap with the given name in s.namespace
 // This function returns direct reference to the internal cache object and should not be mutated.
 // Consider using this function when you can guarantee read only access and don't want the overhead of deep copies
-func (s *DataStore) GetConfigMapRO(namespace, name string) (*corev1.ConfigMap, error) {
-	return s.cfmLister.ConfigMaps(namespace).Get(name)
+// The synchronization to lister takes a while after creating or updating a resource via Kubernetes client.
+// After creating or updating a resource, getting the resource immediately from the lister may not provide the latest result.
+func (s *DataStore) GetConfigMapRO(namespace, name string, fromLister bool) (*corev1.ConfigMap, error) {
```
innobead, line 630: This change would make the function misleading, because the original purpose for this method is to return a cached result.

For `GetConfigMap`, it's a clone from a cached result. 

To make this simple w/o over engineering, I would suggest adding two methods to `datastore/uncached.go` and make the method names differentiated. For example, `GetConfigMapWithoutCache`. We don't need to have a RO or non-RO method in uncached, because the result is from API server directly. 

Also, we can take this time to rename methods in uncached as well to make their purpose understandable. WDYT?

innobead (changes requested):

datastore/kubernetes.go
derekbit: Actually, I perfer all CRUD functions have `fromCache/toCache` parameter for letting users clearly specify the target.
But I'm fine with that adding a `GetConfigMapWithoutCache` in `datastore/uncached.go`.

innobead: This is another way to do it, but the only thing you need to take care of is no need to do an extra `DeepCopy` of the resource if it's returned from the API server directly like `GetConfigMap`, not from cache. 

Also, if we do this way, methods in uncached.go probably need to be refactored as well. This part for me looks quite confusing because it's not designed in any static way but rather a convention only.

datastore/longhorn.go
derekbit: `package backend` will use for verify the creation.

datastore/kubernetes.go
derekbit: Yeah, handle in the future. Add `GetConfigMapWithoutCache` for the current PR.

datastore/longhorn.go
derekbit: But I feel the verification is not necessary as I explained in https://github.com/longhorn/longhorn-manager/pull/1671#issuecomment-1370438033. WDYT? @innobead

datastore/kubernetes.go
innobead: ref: https://github.com/longhorn/longhorn/issues/5263

datastore/kubernetes.go
```diff
@@ -14,6 +14,7 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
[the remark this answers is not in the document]
derekbit: Removed. Thank you.

datastore/uncached.go
```diff
@@ -369,3 +369,12 @@ func (s *DataStore) GetAllLonghornCustomResourceDefinitions() (runtime.Object, e
...
 	return crdList, nil
 }
+
+// GetConfigMapWithoutCache return a new ConfigMap via Kubernetes client object for the given namespace and name
+func (s *DataStore) GetConfigMapWithoutCache(namespace, name string) (*corev1.ConfigMap, error) {
+	resultRO, err := s.kubeClient.CoreV1().ConfigMaps(s.namespace).Get(context.TODO(), name, metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+	return resultRO.DeepCopy(), nil
```
innobead, line 379: Do we need `DeepCopy`? I don't think it's necessary, because it's just created and returned back from the getter of config map.

https://github.com/longhorn/longhorn-manager/blob/60df31ab85f5f561cc012685b0f40ae7642cb29d/vendor/k8s.io/client-go/kubernetes/typed/core/v1/configmap.go#L71-L81

recovery_backend/backend/configmap.go
```diff
@@ -69,110 +71,120 @@ func (rb *RecoveryBackend) CreateConfigMap(hostname, version string) error {
...
 func (rb *RecoveryBackend) EndGrace(hostname, version string) error {
 	configMapName := types.GetConfigMapNameFromHostname(hostname)
-	cm, err := rb.Datastore.GetConfigMap(rb.Namespace, configMapName)
-	if err != nil {
-		return errors.Wrapf(err, "failed to get the configmap %v", configMapName)
-	}
 
-	oldVersion := cm.Annotations["version"]
+	_, err := util.RetryOnConflictCause(func() (interface{}, error) {
+		cm, err := rb.Datastore.GetConfigMapWithoutCache(rb.Namespace, configMapName)
```
innobead, line 84: Why we need retry here? 

If this operation fails somehow, should the recovery backend retry to call this operation again and it should succeed eventually?

Just noticed we do the same for longhorn-manager REST APIs, but why?

recovery_backend/backend/configmap.go
```diff
@@ -69,7 +70,10 @@ func (rb *RecoveryBackend) CreateConfigMap(hostname, version string) error {
 		}
 	}
 
-	return nil
+	_, err = datastore.VerifyCreation(configMapName, "configmap", func(name string) (runtime.Object, error) {
+		return rb.Datastore.GetConfigMapRO(rb.Namespace, name)
+	})
+	return err
```
innobead, line 77: Don't feel we need this. (I know in our current implementation in other places, all have this check; however, don't we just rely on the error returned by K8s API client?)

innobead (dismissed):
LGTM, but have some comments. 

(I think some places using API clients should be reviewed and improved a bit to make code simple by removing unnecessary code like retry and validate check after creation)

recovery_backend/backend/configmap.go
innobead: cc @joshimoo @shuo-wu

recovery_backend/backend/configmap.go
innobead: cc @joshimoo @shuo-wu

recovery_backend/backend/configmap.go
derekbit: > Why we need retry here?

Because we have multiple recovery-backend pod, not sure if all clients update the configmap from the smae pod.

datastore/uncached.go
derekbit: Yes, it is not necessary.

recovery_backend/backend/configmap.go
derekbit: For creation, I think we don't need the `VerifyCreation`. But not sure if there is any context.

recovery_backend/backend/configmap.go
derekbit: > If this operation fails somehow, should the recovery backend retry to call this operation again and it should succeed eventually?

3 replaces can do retry
1. NFS-ganesha: it doesn't do retry from the observation. This will make the connection failed.
2. api/route: the logic is incorrect https://github.com/longhorn/longhorn/issues/5259
3. Current implementation

3. makes sense for the update conflict error.

datastore/uncached.go
derekbit: Updated. Thank you.

recovery_backend/backend/configmap.go
shuo-wu: Compared with asking the upper caller to take care of the cache verification, I would like to handle it inside the CRUD function. I think it's reasonable to ask the creation function to make sure the object is in-sync.

derekbit: BTW, the verifyCreation is not really necessary, because the create/update/get configmap are using k8s client api. Right?

recovery_backend/backend/configmap.go
shuo-wu: Due to the buggy router retry mechanism, we should retain this kind of retry. After fixing https://github.com/longhorn/longhorn/issues/5259, we can consider removing it.

recovery_backend/backend/configmap.go
shuo-wu: We don't need this here. But as I mentioned, I am fine with having verification in the datastore Creation/Update functions.

recovery_backend/backend/configmap.go
shuo-wu: Yes. 
OK. Since we decide to implement https://github.com/longhorn/longhorn/issues/5263, we can skip the verification for the config map first.

recovery_backend/backend/configmap.go
derekbit: Agreed.

joshimoo (commented):
The verifyCreation/verifyUpdate are actual code smells from our controller/distributed system design, I mentioned this here before: https://github.com/longhorn/longhorn-manager/pull/1315#discussion_r864756917

We should over time, improve these smells and start removing some of these stalls (since what they are effectively doing is stall execution)

This will take time and refactoring, similiar to the removal of the resync period :)

innobead (approved):

innobead:
@Mergifyio backport v1.4.x

PR #1672, opened 2023-01-05
feat(recurring-job): periodically cleanup snapshots

c3y1huang:
https://github.com/longhorn/longhorn/issues/3836

### DIFF
--- a/app/recurring_job.go
+++ b/app/recurring_job.go
@@ -64,7 +64,7 @@ func RecurringJobCmd() cli.Command {
 		},
 		Action: func(c *cli.Context) {
 			if err := recurringJob(c); err != nil {
-				logrus.Fatalf("Error taking snapshot: %v", err)
+				logrus.WithError(err).Fatal("Failed to snapshot")
 			}
 		},
 	}
@@ -173,12 +173,12 @@ func recurringJob(c *cli.Context) error {
 				jobRetain,
 				doBackup)
 			if err != nil {
-				log.WithError(err).Error("failed to create new job for volume")
+				log.WithError(err).Error("Failed to create new job for volume")
 				return
 			}
 			err = job.run()
 			if err != nil {
-				log.WithError(err).Errorf("failed to run job for volume")
+				log.WithError(err).Errorf("Failed to run job for volume")
 				return
 			}
 
@@ -353,7 +353,7 @@ func (job *Job) run() (err error) {
 func (job *Job) doRecurringSnapshot() (err error) {
 	defer func() {
 		if err == nil {
-			job.logger.Info("Finish running recurring snapshot")
+			job.logger.Info("Finished recurring snapshot")
 		}
 	}()
 
@@ -431,9 +431,9 @@ func (job *Job) doSnapshotCleanup(backupDone bool) (err error) {
 			if done {
 				if len(errorList) != 0 {
 					for replica, errMsg := range errorList {
-						job.logger.Warnf("error purging snapshots on replica %v: %v", replica, errMsg)
+						job.logger.Warnf("Error purging snapshots on replica %v: %v", replica, errMsg)
 					}
-					job.logger.Warnf("encountered one or more errors while purging snapshots")
+					job.logger.Warn("Encountered one or more errors while purging snapshots")
 				}
 				return nil
 			}
@@ -475,7 +475,7 @@ func (job *Job) listSnapshotNamesForCleanup(snapshots []longhornclient.Snapshot,
 func (job *Job) doRecurringBackup() (err error) {
 	defer func() {
 		if err == nil {
-			job.logger.Info("Finish running recurring backup")
+			job.logger.Info("Finished recurring backup")
 		}
 	}()
 	backupAPI := job.api.BackupVolume
@@ -563,7 +563,7 @@ func (job *Job) doRecurringBackup() (err error) {
 
 	defer func() {
 		if err != nil {
-			job.logger.Warnf("created backup successfully but errored on cleanup for %v: %v", volumeName, err)
+			job.logger.WithError(err).Warnf("Created backup successfully but errored on cleanup for %v", volumeName)
 			err = nil
 		}
 	}()
--- a/app/recurring_job.go
+++ b/app/recurring_job.go
@@ -35,9 +35,6 @@ const (
 	DetachingWaitInterval     = 10 * time.Second
 	VolumeAttachTimeout       = 300 // 5 minutes
 	BackupProcessStartTimeout = 90  // 1.5 minutes
-
-	jobTypeSnapshot = string("snapshot")
-	jobTypeBackup   = string("backup")
 )
 
 type Job struct {
@@ -47,7 +44,7 @@ type Job struct {
 	volumeName   string
 	snapshotName string
 	retain       int
-	jobType      string
+	task         longhorn.RecurringJobType
 	labels       map[string]string
 
 	api *longhornclient.RancherClient
@@ -99,7 +96,6 @@ func recurringJob(c *cli.Context) error {
 	}
 
 	var jobGroups []string = recurringJob.Spec.Groups
-	var jobTask string = string(recurringJob.Spec.Task)
 	var jobRetain int = recurringJob.Spec.Retain
 	var jobConcurrent int = recurringJob.Spec.Concurrency
 
@@ -111,11 +107,6 @@ func recurringJob(c *cli.Context) error {
 	labelJSON, err := json.Marshal(jobLabelMap)
 	if err != nil {
 		return errors.Wrap(err, "failed to get JSON encoding for labels")
-	}
-
-	var doBackup bool = false
-	if jobTask == string(longhorn.RecurringJobTypeBackup) {
-		doBackup = true
 	}
 
 	allowDetachedSetting := types.SettingNameAllowRecurringJobWhileVolumeDetached
@@ -155,7 +146,7 @@ func recurringJob(c *cli.Context) error {
 			log := logger.WithFields(logrus.Fields{
 				"job":        jobName,
 				"volume":     volumeName,
-				"task":       jobTask,
+				"task":       recurringJob.Spec.Task,
 				"retain":     jobRetain,
 				"concurrent": jobConcurrent,
 				"groups":     strings.Join(jobGroups, ","),
@@ -171,7 +162,7 @@ func recurringJob(c *cli.Context) error {
 				snapshotName,
 				jobLabelMap,
 				jobRetain,
-				doBackup)
+				recurringJob.Spec.Task)
 			if err != nil {
 				log.WithError(err).Error("Failed to create new job for volume")
 				return
@@ -199,7 +190,7 @@ func sliceStringSafely(s string, begin, end int) string {
 	return s[begin:end]
 }
 
-func NewJob(logger logrus.FieldLogger, managerURL, volumeName, snapshotName string, labels map[string]string, retain int, backup bool) (*Job, error) {
+func NewJob(logger logrus.FieldLogger, managerURL, volumeName, snapshotName string, labels map[string]string, retain int, task longhorn.RecurringJobType) (*Job, error) {
 	namespace := os.Getenv(types.EnvPodNamespace)
 	if namespace == "" {
 		return nil, fmt.Errorf("cannot detect pod namespace, environment variable %v is missing", types.EnvPodNamespace)
@@ -228,18 +219,13 @@ func NewJob(logger logrus.FieldLogger, managerURL, volumeName, snapshotName stri
 		retain = 1
 	}
 
-	jobType := jobTypeSnapshot
-	if backup {
-		jobType = jobTypeBackup
-	}
-
 	logger = logger.WithFields(logrus.Fields{
 		"namespace":    namespace,
 		"volumeName":   volumeName,
 		"snapshotName": snapshotName,
 		"labels":       labels,
 		"retain":       retain,
-		"jobType":      jobType,
+		"task":         task,
 	})
 
 	return &Job{
@@ -250,7 +236,7 @@ func NewJob(logger logrus.FieldLogger, managerURL, volumeName, snapshotName stri
 		snapshotName: snapshotName,
 		labels:       labels,
 		retain:       retain,
-		jobType:      jobType,
+		task:         task,
 		api:          apiClient,
 	}, nil
 }
@@ -342,7 +328,7 @@ func (job *Job) run() (err error) {
 		job.logger.Infof("Volume %v is in state %v", volumeName, volume.State)
 	}
 
-	if job.jobType == jobTypeBackup {
+	if job.task == longhorn.RecurringJobTypeBackup {
 		job.logger.Infof("Running recurring backup for volume %v", volumeName)
 		return job.doRecurringBackup()
 	}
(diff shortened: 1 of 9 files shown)


c3y1huang:
Test result: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/2910/

app/recurring_job.go
```diff
@@ -403,7 +403,8 @@ func (job *Job) doSnapshotCleanup(backupDone bool) (err error) {
 		return err
 	}
 
-	if len(deleteSnapshotNames) > 0 {
+	shouldPurgeSnapshots := len(deleteSnapshotNames) > 0 || job.task == longhorn.RecurringJobTypeSnapshotCleanup
```
shuo-wu, line 406: What if `deleteSnapshotNames` (for task snapshot) is 0 but there are system snapshots? Should we help purge them?

datastore/longhorn.go
```diff
@@ -3702,7 +3702,7 @@ func (s *DataStore) DeleteRecurringJob(name string) error {
 }
 
 func ValidateRecurringJob(job longhorn.RecurringJobSpec) error {
-	if job.Cron == "" || job.Task == "" || job.Name == "" || job.Retain == 0 {
+	if job.Cron == "" || job.Task == "" || job.Name == "" {
```
shuo-wu, line 3705: `job.Retain < 0`

app/recurring_job.go
```diff
@@ -466,6 +488,10 @@ func (job *Job) listSnapshotNamesForCleanup(snapshots []longhornclient.Snapshot,
 	return snapshotsToNames(filterSnapshotsNotInTargets(snapshots, retainingSnapshots))
 }
 
+func (job *Job) listSnapshotNamesToCleanupForTaskSnapshotDelete(snapshots []longhornclient.Snapshot) []string {
+	return filterExpiredItems(snapshotsToNameWithTimestamps(snapshots), job.retain+1)
```
shuo-wu, line 496: Why `job.retain+1`?

webhook/resources/recurringjob/mutator.go
```diff
@@ -51,6 +53,23 @@ func (r *recurringJobMutator) Create(request *admission.Request, newObj runtime.
...
+		"task":         recurringjob.Spec.Task,
+	})
+	switch recurringjob.Spec.Task {
+	case longhorn.RecurringJobTypeSnapshotCleanup:
+		if recurringjob.Spec.Retain != 0 {
+			log.Debugf("Replacing ineffective retain value in RecurringJob: from %v to 0", recurringjob.Spec.Retain)
+			patchOps = append(patchOps, `{"op": "replace", "path": "/spec/retain", "value": 0}`)
+		}
+	default:
+		if recurringjob.Spec.Retain < 1 {
```
shuo-wu, line 67: For task snapshot delete, can we set retain to 0?

app/recurring_job.go
```diff
@@ -437,12 +439,32 @@ func (job *Job) doSnapshotCleanup(backupDone bool) (err error) {
...
+func (job *Job) listSnapshotNamesToCleanup(snapshots []longhornclient.Snapshot, backupDone bool) []string {
+	deleteSnapshotNames := job.listSnapshotNamesToCleanupForTaskSnapshot(snapshots, backupDone)
+	switch job.task {
+	case longhorn.RecurringJobTypeSnapshotDelete:
+		deleteSnapshotNames = job.listSnapshotNamesToCleanupForTaskSnapshotDelete(snapshots)
+	}
+	return deleteSnapshotNames
+}
+
+func (job *Job) listSnapshotNamesToCleanupForTaskSnapshot(snapshots []longhornclient.Snapshot, backupDone bool) []string {
```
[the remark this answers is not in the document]
c3y1huang: How about `filterExpiredSnapshotsOfCurrentRecurringJob`?

webhook/resources/recurringjob/mutator.go
c3y1huang: Do you mean making retain value ineffective for task snapshot delete?

app/recurring_job.go
c3y1huang: In the current implementation recurring job creates a snapshot and then does the cleanup. For the cleanup task and delete task there will be no snapshot creation. So I assume `+1` should be added for the volume-head.

app/recurring_job.go
c3y1huang: Absolutely, update!

datastore/longhorn.go
c3y1huang: We have the mutation for `recurringjob.Spec.Retain`, do we still need this check?

k8s/pkg/apis/longhorn/v1beta2/recurringjob.go
```diff
@@ -2,12 +2,14 @@ package v1beta2
...
-// +kubebuilder:validation:Enum=snapshot;backup
+// +kubebuilder:validation:Enum=snapshot;snapshot-cleanup;snapshot-delete;backup
 type RecurringJobType string
 
 const (
-	RecurringJobTypeSnapshot = RecurringJobType("snapshot")
-	RecurringJobTypeBackup   = RecurringJobType("backup")
+	RecurringJobTypeSnapshot        = RecurringJobType("snapshot")
+	RecurringJobTypeSnapshotCleanup = RecurringJobType("snapshot-cleanup")
+	RecurringJobTypeSnapshotDelete  = RecurringJobType("snapshot-delete")
```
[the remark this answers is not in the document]
shuo-wu: Yes. Having a comment would be more straightforward.

PhanLe1010:
Could we postpone this PR to wait for the AD refactoring controller to be merged first? We propally need to modify the LEP to interact with the Snapshot CRD API instead of calling the engine to create/delete snapshot directly

datastore/longhorn.go
shuo-wu: I mean, we can allow `retain == 0`, right? If the type is `snapshot-delete`, `retain == 0` means retaining the volume head only.

app/recurring_job.go
shuo-wu: Will Longhorn count the volume head as a snapshot here...?

webhook/resources/recurringjob/mutator.go
shuo-wu: As I mentioned above, 0 means retaining the volume head only.

k8s/pkg/apis/longhorn/v1beta2/recurringjob.go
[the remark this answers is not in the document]
c3y1huang: Updated.

app/recurring_job.go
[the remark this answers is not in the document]
c3y1huang: Updated.

app/recurring_job.go
```diff
@@ -437,12 +439,32 @@ func (job *Job) doSnapshotCleanup(backupDone bool) (err error) {
...
+}
+
 type NameWithTimestamp struct {
 	Name      string
 	Timestamp time.Time
 }
 
-func (job *Job) listSnapshotNamesForCleanup(snapshots []longhornclient.Snapshot, backupDone bool) []string {
+func (job *Job) listSnapshotNamesToCleanup(snapshots []longhornclient.Snapshot, backupDone bool) []string {
+	deleteSnapshotNames := job.listSnapshotNamesToCleanupForTaskSnapshot(snapshots, backupDone)
```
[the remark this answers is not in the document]
c3y1huang: Updated.

app/recurring_job.go
c3y1huang: My misunderstanding, you are right Longhorn doesn't count the volume head here. However, `filterExpiredItems` is expecting a new snapshot created before the the clean up. Therefore the `+1` is required to have the correct number of snapshots in the returning list.

datastore/longhorn.go
c3y1huang: Yes, so removing it from the validation here. And I think we don't need to validate `job.Retain < 0` since it's included in the mutation.

c3y1huang:
> Could we postpone this PR to wait for the AD refactoring controller to be merged first? We probably need to modify [the LEP](https://github.com/longhorn/longhorn/pull/5190/files) to interact with the Snapshot CRD API instead of calling the engine to create/delete snapshot directly

I think this PR should go before the AD refactoring so it can be backported to 1.3.3. Ref: https://github.com/longhorn/longhorn/issues/3836#issuecomment-1380137893

This is a small PR and focusing on the snapshot deletion only, meaning effort should be low at resolving the conflicts.

WDYT @PhanLe1010 ?

webhook/resources/recurringjob/mutator.go
c3y1huang: Got it! Thank you and updated.

app/recurring_job.go
shuo-wu: Sorry...? I do not get the idea...
If there is 1 existing snapshot only and retain is 0, this would be `filterExpiredItems(snapshotsToNameWithTimestamps([snap0]), 1)`. And the returned list is an empty one, which is unexpected...

c3y1huang: IIUC, if there is 1 existing snapshot only and retain is 0, the input snapshots will also include the volume-head. This will be
`filterExpiredItems(snapshotsToNameWithTimestamps([snap0, volume-head]), 1)`

However, I've missed the retain correction [here](https://github.com/longhorn/longhorn-manager/blob/master/app/recurring_job.go#L235-L238) which returned an empty list when retain value is 0.

Thank you for pointing this out, fixed. 

Test result: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/2977/

app/recurring_job.go
```diff
@@ -460,52 +441,34 @@ func (job *Job) doSnapshotCleanup(backupDone bool) (err error) {
...
-	if err != nil {
-		return 0, errors.Wrapf(err, "could not get volume-head for volume %v", job.volumeName)
-	}
-	volumeHeadSize, err := strconv.ParseInt(volumeHead.Size, 10, 64)
-	if err != nil {
-		return 0, err
+func (job *Job) listSnapshotNamesToCleanup(snapshots []longhornclient.Snapshot, backupDone bool) []string {
+	switch job.task {
+	case longhorn.RecurringJobTypeSnapshotDelete:
+		return job.filterExpiredSnapshotsOfAllRecurringJob(snapshots)
```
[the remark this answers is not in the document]
c3y1huang: > This naming is weird.

Renamed to `filterExpiredSnapshots`.

> We will remove user-created snapshots as well here, right?

Yes.

app/recurring_job.go
```diff
@@ -276,7 +253,7 @@ func NewJob(logger logrus.FieldLogger, managerURL, volumeName, snapshotName stri
 func (job *Job) handleVolumeDetachment() {
 	volumeAPI := job.api.Volume
 	volumeName := job.volumeName
-	jobName, _ := job.labels[types.RecurringJobLabel]
+	jobName := job.labels[types.RecurringJobLabel]
 	if jobName == "" {
 		job.logger.Warn("Missing RecurringJob label")
 		return
```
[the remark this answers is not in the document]
c3y1huang: Removed.

app/recurring_job.go
c3y1huang: I've already removed the retain correction and filtered in `snapshotsToNameWithTimestamps` by adding a new function argument `excludeVolumeHead`. I think this way will have minimal impact.

shuo-wu (dismissed):

app/recurring_job.go
```diff
@@ -514,12 +469,12 @@ func (job *Job) listSnapshotNamesForCleanup(snapshots []longhornclient.Snapshot,
 	// Only consider deleting the snapshots that were created by our current job
 	snapshots = filterSnapshotsWithLabel(snapshots, types.RecurringJobLabel, jobLabel)
 
-	if job.jobType == jobTypeSnapshot {
-		return filterExpiredItems(snapshotsToNameWithTimestamps(snapshots), job.retain)
+	if job.task == longhorn.RecurringJobTypeSnapshot {
+		return filterExpiredItems(snapshotsToNameWithTimestamps(snapshots, false), job.retain)
```
innobead, line 475: Why not exclude the volume head here?

innobead (changes requested):
In general, LGTM, but some comments needed to resolve.

app/recurring_job.go
```diff
@@ -369,17 +340,23 @@ func (job *Job) run() (err error) {
...
 
-	err = job.doSnapshot()
-	if err != nil {
-		return err
+	switch job.task {
+	case longhorn.RecurringJobTypeSnapshot:
+		if err = job.doSnapshot(); err != nil {
+			return err
+		}
+		fallthrough
```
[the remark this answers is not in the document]
c3y1huang: Keeping [the current behavior](https://github.com/longhorn/longhorn-manager/blob/master/app/recurring_job.go#L377-L382) to clean up the retain after the snapshot.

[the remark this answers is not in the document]
innobead: ah, I see.

k8s/crds.yaml
```diff
@@ -3309,6 +3311,8 @@ spec:
                     task:
                       enum:
                       - snapshot
+                      - snapshot-cleanup
```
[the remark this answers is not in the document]
c3y1huang: I prefer to have this handled in https://github.com/longhorn/longhorn/issues/2865.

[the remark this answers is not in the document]
innobead: cc @weizhe0422

app/recurring_job.go
c3y1huang: The original intention is to keep the current logic to minimize the impact. 
Having a second look at this and seems cleanup will only remove the older snapshot, so the returned list will never be the volume head.

I'm not sure about the original purpose of why the volume head isn't excluded. Likely that we did not need the extra logic to exclude it since:
1. There was enforcement for retain to be at least 1.
1. The volume head will always have a newer timestamp.

We are accepting a 0 retain value for the snapshot-cleanup/delete, so I will exclude the volume head.

c3y1huang: Test result: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/2991/

app/recurring_job.go
```diff
@@ -56,7 +46,7 @@ type Job struct {
 	volumeName   string
 	snapshotName string
 	retain       int
-	jobType      string
+	task         longhorn.RecurringJobType
 	labels       map[string]string
```
innobead, line 50: nit: not sure why changing the field name, but I would suggest the naming is based on the type, so here `jobType` is quite suitable.

app/recurring_job.go
```diff
@@ -313,7 +306,7 @@ func (job *Job) run() (err error) {
 
 	volumeAPI := job.api.Volume
 	volumeName := job.volumeName
-	jobName, _ := job.labels[types.RecurringJobLabel]
+	jobName := job.labels[types.RecurringJobLabel]
```
innobead, line 287: The below code is not required because the label is mandatory.

```go
if jobName == "" {
    return fmt.Errorf("missing RecurringJob label")
}
```

innobead (changes requested):
LGTM except two minor comments needed to resolve.

app/recurring_job.go
innobead: This is actually optional, so if decide not to change, it's also fine.

c3y1huang: Feel this should be consistent with the [RecurringJobSpec.Task](https://github.com/longhorn/longhorn-manager/blob/master/k8s/pkg/apis/longhorn/v1beta2/recurringjob.go#L38).

innobead: the spec has been defined, so this is why I said is nice to change. Better to keep it as is to make things compatible first.

innobead (approved):

innobead:
@c3y1huang Need to update the manifests in longhorn/longhorn

c3y1huang:
> @c3y1huang Need to update the manifests in longhorn/longhorn

https://github.com/longhorn/longhorn/pull/5241

PR #1679, opened 2023-01-13
Skip updating the trim related field for old engines or engines in old engine instance managers

shuo-wu:
https://github.com/longhorn/longhorn/issues/5218

### DIFF
--- a/engineapi/types.go
+++ b/engineapi/types.go
@@ -13,7 +13,7 @@ import (
 const (
 	// CurrentCLIVersion indicates the default API version manager used to talk with the
 	// engine, including `longhorn-engine` and `longhorn-instance-manager`
-	CurrentCLIVersion = 6
+	CurrentCLIVersion = 7
 	// MinCLIVersion indicates the Min API version manager used to talk with the
 	// engine.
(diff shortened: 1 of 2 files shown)


engineapi/types.go
```diff
@@ -13,7 +13,7 @@ import (
 const (
 	// CurrentCLIVersion indicates the default API version manager used to talk with the
 	// engine, including `longhorn-engine` and `longhorn-instance-manager`
-	CurrentCLIVersion = 6
+	CurrentCLIVersion = 7
```
innobead, line 16: We missed to bump this in 1.4.0?.. It seems fine, because the minVersion of engine did not change. https://github.com/longhorn/longhorn-engine/blob/67b890eb3a1ce253d109cd5546d9ce17fe27cc3e/pkg/meta/version.go#L6-L6

innobead (approved):
LGTM

innobead:
@Mergifyio backport v1.4.x

engineapi/types.go
shuo-wu: Yes... Since and min version does not change, everything is still compatible.

PR #1683, opened 2023-01-19
Force unmount the mount point of the Longhorn volume

derekbit:
UnmountWithForce will do graceful umount. If failed, do the force unmount.

[Longhorn 5296](https://github.com/longhorn/longhorn/issues/5296)

### DIFF
--- a/csi/util.go
+++ b/csi/util.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/container-storage-interface/spec/lib/go/csi"
 	"github.com/pkg/errors"
@@ -27,6 +28,8 @@ import (
 const (
 	// defaultStaleReplicaTimeout set to 48 hours (2880 minutes)
 	defaultStaleReplicaTimeout = 2880
+
+	defaultForceUmountTimeout = 30 * time.Second
 )
 
 // NewForcedParamsExec creates a osExecutor that allows for adding additional params to later occurring Run calls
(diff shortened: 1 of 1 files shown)


PhanLe1010 (approved):
LGTM

PhanLe1010:
Tested that this PR fixed the issue https://github.com/longhorn/longhorn/issues/5296#issuecomment-1397680334. Thanks @derekbit

csi/util.go
```diff
@@ -247,8 +250,16 @@ func ensureMountPoint(targetPath string, mounter mount.Interface) (bool, error)
 }
 
 func unmount(targetPath string, mounter mount.Interface) error {
-	logrus.Debugf("trying to unmount potential mount point %v", targetPath)
-	err := mounter.Unmount(targetPath)
+	var err error
+
+	forceUnmounter, ok := mounter.(mount.MounterForceUnmounter)
```
shuo-wu, line 255: When will the input mount be `MounterForceUnmounter`?

derekbit: The `UnmountWithForce` is implemented on Linux ([code](https://github.com/longhorn/longhorn-manager/blob/master/vendor/k8s.io/mount-utils/mount_linux.go#L313)), so we can assert `Mounter` as `MounterForceUnmounter` ([code](https://github.com/longhorn/longhorn-manager/blob/master/vendor/k8s.io/mount-utils/mount.go#L84)).

shuo-wu: Then by default will the unmounter be considered as the force unmounter?

derekbit: > Then by default will the unmounter be considered as the force unmounter?

Yeah, your idea is fine, because LH only supports linux for now.
Do you think we should use forceUnmounter by default? Or keep the general method?

shuo-wu: I mean, will this lead to the input unmounter be automatically converted to the force unkounter? If No, when will the input unmounter be the force one?

innobead: It just depends on how `mounter` is created. There is no any magic conversion, but rather if `mounter` implements unmounter interface.

innobead (approved):

derekbit:
@Mergify backport v1.3.x

PR #1688, opened 2023-02-06
fix(setting): UI missing asterisk in formulas

c3y1huang:
https://github.com/longhorn/longhorn/issues/5338

Result:
Screenshot-2023-02-08-16:52:21

### DIFF
--- a/types/setting.go
+++ b/types/setting.go
@@ -831,7 +831,7 @@ var (
 		DisplayName: "Guaranteed Engine Manager CPU",
 		Description: "This integer value indicates how many percentage of the total allocatable CPU on each node will be reserved for each engine manager Pod. For example, 10 means 10% of the total CPU on a node will be allocated to each engine manager pod on this node. This will help maintain engine stability during high node workload. \n\n" +
 			"In order to prevent unexpected volume engine crash as well as guarantee a relative acceptable IO performance, you can use the following formula to calculate a value for this setting: \n\n" +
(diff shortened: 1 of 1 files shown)


shuo-wu (approved):

innobead (commented):
I think this should be handled by UI part instead to avoid any escaping. 

@smallteeths Could you help with this instead?

c3y1huang:
> I think this should be handled by UI part instead to avoid any escaping.
> 
> @smallteeths Could you help with this instead?

I see. However, this PR also makes the formulas stand out for readability. We can close this PR if it is not helpful. WDYT @innobead ?

smallteeths:
> I think this should be handled by UI part instead to avoid any escaping.
> 
> @smallteeths Could you help with this instead?

ok

innobead:
> > I think this should be handled by UI part instead to avoid any escaping.
> > @smallteeths Could you help with this instead?
> 
> I see. However, this PR also makes the formulas stand out for readability (comparing to the screenshot [here](https://github.com/longhorn/longhorn/issues/5338#issuecomment-1423610792)). We can close this if is not helpful. WDYT @innobead ?

I missed the effect changed by @c3y1huang . @smallteeths to save the efforts, I think this change should be good enough for us, so probably no need to spend time customizing this on UI at the moment. Thanks.

innobead (approved):

innobead:

PR #1689, opened 2023-02-06
chore: run k8s/generate_code.sh

c3y1huang:
Same as https://github.com/longhorn/longhorn-manager/pull/1585.

Reverting the updated package in [commit](https://github.com/longhorn/longhorn-manager/commit/5092f638f4a880be17f65e60cd928fc370182306).

Ref: https://github.com/longhorn/longhorn/issues/5189


joshimoo (approved):
LGTM

PR #1691, opened 2023-02-06
Don't fail a running replica when node is no ready

derekbit:
Longhorn/longhorn#5340

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -1573,18 +1573,16 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
 			if err != nil {
 				return err
 			}
-			if !canIMLaunchReplica {
+			if canIMLaunchReplica {
+				if r.Spec.FailedAt == "" && r.Spec.EngineImage == v.Status.CurrentImage {
+					if r.Status.CurrentState == longhorn.InstanceStateStopped {
+						r.Spec.DesireState = longhorn.InstanceStateRunning
+					}
+				}
+			} else {
 				// wait for IM is starting when volume is upgrading
 				if vc.isVolumeUpgrading(v) {
 					continue
-				}
-				if r.Spec.FailedAt == "" {
-					r.Spec.FailedAt = vc.nowHandler()
-				}
-				r.Spec.DesireState = longhorn.InstanceStateStopped
-			} else if r.Spec.FailedAt == "" && r.Spec.EngineImage == v.Status.CurrentImage {
-				if r.Status.CurrentState == longhorn.InstanceStateStopped {
-					r.Spec.DesireState = longhorn.InstanceStateRunning
 				}
 			}
 			rs[r.Name] = r
--- a/controller/volume_controller_test.go
+++ b/controller/volume_controller_test.go
@@ -807,7 +807,6 @@ func (s *TestSuite) TestVolumeLifeCycle(c *C) {
 	for _, r := range tc.expectReplicas {
 		if r.Spec.NodeID == TestNode2 {
 			r.Spec.DesireState = longhorn.InstanceStateStopped
-			r.Spec.FailedAt = getTestNow()
 		} else {
 			r.Spec.DesireState = longhorn.InstanceStateRunning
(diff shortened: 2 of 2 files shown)


derekbit:
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/2971/

controller/instance_handler.go
```diff
@@ -69,6 +69,20 @@ func (h *InstanceHandler) syncStatusWithInstanceManager(im *longhorn.InstanceMan
 		return
 	}
 
+	if im.Status.CurrentState == longhorn.InstanceManagerStateUnknown {
+		// Do not set the replica status to unknown state directly when the instance manager is
```
shuo-wu, line 73: I don't think we need this change. The essential of this issue is, the volume controller shouldn't mark unknown replicas as Failed.

derekbit: Yeah, you are right. Marking the replica as unknown and don't failed it is a good idea.
Updated the PR. Thank you.

controller/volume_controller.go
```diff
@@ -1578,10 +1578,16 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
 				if vc.isVolumeUpgrading(v) {
 					continue
 				}
-				if r.Spec.FailedAt == "" {
-					r.Spec.FailedAt = vc.nowHandler()
+				notReady, err := vc.ds.IsNodeNotReady(r.Spec.NodeID)
+				if err != nil {
+					return err
+				}
```
shuo-wu, line 1584: If we plan to reply on the monitoring and feedback of the engine process, does this controller still need `canIMLaunchReplica, err := vc.canInstanceManagerLaunchReplica(r)`? If NO, then can the controller directly check if the replica is unknown or not?

derekbit: Do you mean directly checking the replica is in unknown state and don't need to rely on `canInstanceManagerLaunchReplica`?

When the replica is in unknown state, the node might be `existing but not ready` or `deleted`, we still use `canInstanceManagerLaunchReplica` to distinguish the two cases and check if we can stop the replica.

shuo-wu: If the node is unavailable, the engine process should mark the replica as ERR then report to the engine monitor thread in the engine controller. In other words, the volume controller does not need to take care of everything. This simplifies the whole control. WDYT?

derekbit: Agreed. I removed the check `faileAt` and `stopped` from here. The two fields will be updated based on the engine monitor's results.

datastore/longhorn.go
```diff
@@ -2366,6 +2366,25 @@ func (s *DataStore) IsNodeDownOrDeleted(name string) (bool, error) {
 	return false, nil
 }
 
+func (s *DataStore) IsNodeNotReady(name string) (bool, error) {
```
shuo-wu, line 2369: We don't need this anymore.

shuo-wu (changes requested):

datastore/longhorn.go
derekbit: Removed. Thank you.

controller/volume_controller.go
```diff
@@ -1573,19 +1573,17 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
 			if err != nil {
 				return err
 			}
-			if !canIMLaunchReplica {
+			if canIMLaunchReplica {
```
innobead, line 1576: In `canInstanceManagerLaunchReplica`, `ds.IsNodeDownOrDeleted` will check if node is down/gone or exists but not ready. 

Is it still valuable to rely on `node down/gone` only to determine the replica should be stopped instead of just removing the code here. This is still helpful for improving resilience if the node is really gone instead of timeout only. WDYT?

innobead: It seems already discussed at https://github.com/longhorn/longhorn-manager/pull/1691#discussion_r1099626358, then it's OK to keep the current implementation.

innobead (approved):
LGTM

PR #1698, opened 2023-02-13
Fix upgrade failures caused by CRD differences between longhorn versions

derekbit:
https://github.com/longhorn/longhorn/issues/5368

### DIFF
--- a/upgrade/util/util.go
+++ b/upgrade/util/util.go
@@ -4,8 +4,11 @@ import (
 	"context"
 	"fmt"
 	"math"
+	"reflect"
 	"sync"
 
+	"github.com/jinzhu/copier"
+	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 
 	v1 "k8s.io/api/core/v1"
@@ -14,6 +17,7 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	clientset "k8s.io/client-go/kubernetes"
 
+	"github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
 	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
 	lhclientset "github.com/longhorn/longhorn-manager/k8s/pkg/client/clientset/versioned"
 	"github.com/longhorn/longhorn-manager/meta"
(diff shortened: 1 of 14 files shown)


upgrade/util/util.go
```diff
@@ -144,3 +148,770 @@ func CreateOrUpdateLonghornVersionSetting(namespace string, lhClient *lhclientse
...
+	obj, err := lhClient.LonghornV1beta2().RecurringJobs(namespace).Create(context.TODO(), job, metav1.CreateOptions{})
+	if err != nil {
+		return obj, err
+	}
+
+	var recurringJobs map[string]*v1beta2.RecurringJob
+	if v, ok := resourceMaps[types.LonghornKindRecurringJob]; ok {
+		recurringJobs = v.(map[string]*v1beta2.RecurringJob)
+	} else {
+		recurringJobs = map[string]*v1beta2.RecurringJob{}
```
shuo-wu, line 546: For this case, do we need `resourceMaps[types.LonghornKindRecurringJob] = recurringJobs`?
Similar below

upgrade/util/util.go
```diff
@@ -144,3 +148,770 @@ func CreateOrUpdateLonghornVersionSetting(namespace string, lhClient *lhclientse
...
+	settings := map[string]*v1beta2.Setting{}
+	settingList, err := lhClient.LonghornV1beta2().Settings(namespace).List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		return nil, err
+	}
+	for _, setting := range settingList.Items {
+		settingCopy := v1beta2.Setting{}
+		if err := copier.Copy(&settingCopy, setting); err != nil {
+			return nil, err
+		}
```
shuo-wu, line 166: Any reason to copy objects?

upgrade/util/util.go
```diff
@@ -144,3 +148,770 @@ func CreateOrUpdateLonghornVersionSetting(namespace string, lhClient *lhclientse
...
+				return err
+			}
+		}
+
+		if !reflect.DeepEqual(existingNode.Spec, node.Spec) ||
+			!reflect.DeepEqual(existingNode.ObjectMeta, node.ObjectMeta) {
+			if _, err = lhClient.LonghornV1beta2().Nodes(namespace).Update(context.TODO(), node, metav1.UpdateOptions{}); err != nil && !apierrors.IsConflict(errors.Cause(err)) {
+				return err
+			}
+		}
```
shuo-wu, line 630: Sometimes the upgrade involves label/annotation/ownership/... update. Checking spec & status may be not enough.
Similar below.

shuo-wu (changes requested):
BTW, do you have a scalability test against this PR？

upgrade/util/util.go
derekbit: Do you mean directly assigne `setting`'s address to `settings[name]`?

derekbit:
> BTW, do you have a scalability test against this PR？

What's the purpose o the scalability test?

upgrade/util/util.go
derekbit: Yeah. So, I also check `ObjectMeta` here. Anything I missed?

upgrade/util/util.go
derekbit: For `recurringJobs`, it's not really necessary, because `recurringJobs` are GET and CREATE. `resourceMaps[types.LonghornKindRecurringJob]` is just for consistency. We can remove it.

upgrade/util/util.go
shuo-wu: Yes. Is there any concern missed?

upgrade/util/util.go
shuo-wu: I mean, if `resourceMaps[types.LonghornKindRecurringJob]` does not exist, should we add this newly created RecurringJob map to `resourceMaps`?

upgrade/util/util.go
shuo-wu: OK, Sorry for missing that line.

shuo-wu:
It's better to check, if there is any timeout or other unknown issues with this new upgrade flow when there are thousands of volumes/backups/snapshots

upgrade/util/util.go
derekbit: I see. From the logics related to `RecurringJob`, it should be fine and not see any issue for now.

upgrade/util/util.go
innobead: Yeah. No need to have an extra copy because the result is from the API client directly instead of cache.

innobead:
This is a fair concern, because the implemented way is straight forward but not quite scalable. 

How about iterating each resource type to walk through all upgrade paths instead? (of course, we can do even performing way by running each iteration in parallel; however, this is optional for this change)

derekbit:
> This is a fair concern, because the implemented way is straight forward but not quite scalable.
> 
> How about iterating each resource type to walk through all upgrade paths instead? (of course, we can do even performing way by running each iteration in parallel; however, this is optional for this change)

Some resources have dependencies. This might lead to extra maintainance effort.
BTW, should we handle v1beta1 api (LH version older than v1.3.0)?

innobead:
> > This is a fair concern, because the implemented way is straight forward but not quite scalable.
> > How about iterating each resource type to walk through all upgrade paths instead? (of course, we can do even performing way by running each iteration in parallel; however, this is optional for this change)
> 
> Some resources have dependencies. This might lead to extra maintainance effort. BTW, should we handle v1beta1 api (LH version older than v1.3.0)?

I don't think we need it, especially since we would like to have https://github.com/longhorn/longhorn/issues/5131, so we don't need it anymore in the master/1.5.0 code base. For this change, do we need to backport to 1.4.x?

innobead:
> Some resources have dependencies. This might lead to extra maintainance effort. BTW, should we handle v1beta1 api (LH version older than v1.3.0)?

What dependencies have you already noticed?

upgrade/util/util.go
derekbit: Originaly, I used the method I mentioned in https://github.com/longhorn/longhorn-manager/pull/1698#discussion_r1103978644 but met an issue, so I use `copy` instead.

Now, I found the root cause https://husni.dev/beware-of-implicit-memory-aliasing-in-go-foor-loop/.

innobead:
What potential dependencies mentioned by @derekbit are validation webhook, but this needs to clarify as well.

upgrade/util/util.go
shuo-wu: I see what you encountered. And I had handled some similar cases somewhere. Hence I would recommend the format I used:
```
for i := range settingList.Items {
    setting := settingList.Items[i]
    ......
}
```

upgrade/util/util.go
shuo-wu: Here all creation functions have similar formats. I would suggest adding this handling for all of them.

shuo-wu:
> What dependencies have you already noticed

We cannot guarantee there are no more dependencies in the future. Hence it's better to have a resource-handling-order-independent way for the upgrade.

innobead:
Agreed

derekbit:
I feel the avoidance of the resouces' dependencies is difficult as Shuo mentioned.
Another thought is putting `UpdateCustomizedSettings` in front to the `Upgrade`. Any concern? cc @shuo-wu @innobead

shuo-wu:
As long as the current implementation has no scalability issue, everything is good and we don't need to touch `UpdateCustomizedSettings`.

derekbit:
> As long as the current implementation has no scalability issue, everything is good and we don't need to touch `UpdateCustomizedSettings`.

Got it. Will do the scalability issue first.

innobead:
Agreed with you. The code change LGTM as well. 

Also, after upgrade path enforcement is introduced, the number of resources could be reduced as well, because the number of upgrade paths will decrease. For example, for 1.5.0, we only need 1.4.x -> 1.5.0.

derekbit:
**Scalability test**

Created a Longhorn v1.2.0 cluster on `AWS EC2 t2.2xlarge (8 vCPUs/ 32GiB / EBS)` and `Kubernetes v1.24.9+rke2r2`. Then, created the Longhorn resources.
| Resources | # |
| -- | -- |
| Nodes | 3 |
| RWO volumes | 400 |
| RWX volumes | 400 |
| Replicas | 1600 |
| Backups |  1987 |
| RecuringJobs |  400 |

Upgraded the cluster to master-head with the patch. It took 7 minutes to complete the upgrade process.

The memory usage on each node is not high and the maxima is about 90 MiB.
image

image

image


cc @shuo-wu @innobead

innobead:
@derekbit thanks for the info! Did the test include snapshot CR? Usually, the number of snapshots will be pretty large than other resources. This might be a concern if we did not check that resource or actually in our current upgrade paths, there were no snapshots involved?

derekbit:
> @derekbit thanks for the info! Did the test include snapshot CR? Usually, the number of snapshots will be pretty large than other resources. This might be a concern if we did not check that resource or actually in our current upgrade paths, there were no snapshots involved?

Forgot providing the information.
The number of snapshots is 1987.

upgrade/util/util.go
derekbit: Addressed. Thank you.

upgrade/util/util.go
derekbit: > Here all creation functions have similar formats. I would suggest adding this handling for all of them.

Sorry. What's the handling?

derekbit: cc @shuo-wu

shuo-wu: I mean, adding something like `resourceMaps[types.LonghornKindRecurringJob] = recurringJobs` for all creation functions.

shuo-wu:
It seems that the CPU consuming peak is around 3 in your test case. Is that fine? Using 400M memory at most is acceptable for me.

derekbit:
> It seems that the CPU consuming peak is around 3 in your test case. 

The upgrade happened from 23:56 ~ 00:02. The high CPU usage looks not caused by the upgrade process and happened after the it.

upgrade/util/util.go
derekbit: Updated. Thank you.

shuo-wu:
The peak is around 0:07... Can the peak be caused by controllers' force syncing against all resources. 
If the upgrade is not the cause, I am fine with this PR

shuo-wu (approved):

derekbit:
> The peak is around 0:07... Can the peak be caused by controllers' force syncing against all resources? But **all** manager pods should consume lots of resources at that moment... If the upgrade is not the cause, I am fine with this PR

Forgot mentioning that the fist node doesn't have any replicas and volumes. All volumes and replicas are on the second and third node. That's why the last two nodes's CPU became high when controllers' force syncing against all resources, right?

upgrade/util/util.go
```diff
@@ -144,3 +148,732 @@ func CreateOrUpdateLonghornVersionSetting(namespace string, lhClient *lhclientse
 	}
 	return nil
 }
+
+func ListSettings(namespace string, lhClient *lhclientset.Clientset, resourceMaps map[string]interface{}) (map[string]*v1beta2.Setting, error) {
```
innobead, line 152: Can we change to the function name to below or any similar? The same change should be applied to other functions. 

```
// ListAndUpdateSettingsInProvidedCache list all settings and save them into the provided cached `resourceMaps`. This method is not thread-safe.
func ListAndUpdateSettingsInProvidedCache(...) {
    ...
```

innobead (changes requested):

innobead (approved):
LGTM

nit: rename `resourceMaps` to `resourceMap`

PR #1699, opened 2023-02-13
feat(support-bundle): pass new parameter SUPPORT_BUNDLE_REQUEST_PURPOSE

weizhe0422:
[Longhorn 5073](https://github.com/longhorn/longhorn/issues/5073)

The [support-bundle-kit PR](https://github.com/rancher/support-bundle-kit/pull/63) will accept this new parameter and execute `collector-longhorn.sh` to collect system information and logs.

### DIFF
--- a/controller/support_bundle_controller.go
+++ b/controller/support_bundle_controller.go
@@ -679,6 +679,10 @@ func (c *SupportBundleController) newSupportBundleManager(supportBundle *longhor
 									Name:  "SUPPORT_BUNDLE_IMAGE_PULL_POLICY",
 									Value: string(api.PullAlways),
 								},
+								{
+									Name:  "SUPPORT_BUNDLE_REQUEST_PURPOSE",
+									Value: "longhorn",
+								},
 							},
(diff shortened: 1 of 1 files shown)


innobead (commented):
Why we need this?

weizhe0422:
> Why we need this?

Since the support-bundle-kit can using this parameter to identify and execute the specified collector without impact or overlap exists ones, e.g., harvester.

innobead:
> > Why we need this?
> 
> I think the support-bundle-kit can using this parameter to identify and execute the specified collector without impact or overlap exists ones, e.g., harvester.
> 
> ref. [rancher/support-bundle-kit#63 (comment)](https://github.com/rancher/support-bundle-kit/pull/63#issuecomment-1424085641)

then why need to change the code here? longhorn should be implicit here?

weizhe0422:
I think I can use the namespace where the `support-bundle-manager` resides to identify that the request is coming from the Longhorn system. So I will close this PR.

PR #1700, opened 2023-02-14
fix: incorrect router retry mechanism

mantissahz:
Keep the http request body and copy to the request body again when the retry occurred.

Handling in HandleError function is not correct when the request body was already read before, and the next retry always hit the EOF error.

longhorn/longhorn#5259

### DIFF
--- a/api/router.go
+++ b/api/router.go
@@ -2,38 +2,20 @@ package api
 
 import (
 	"net/http"
-	"time"
 
 	"github.com/gorilla/mux"
-	"github.com/pkg/errors"
 	"github.com/rancher/go-rancher/api"
 	"github.com/rancher/go-rancher/client"
 	"github.com/sirupsen/logrus"
 
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
-
 	"github.com/longhorn/longhorn-manager/metrics_collector/registry"
-)
-
-var (
-	RetryCounts   = 5
-	RetryInterval = 100 * time.Millisecond
 )
 
 type HandleFuncWithError func(http.ResponseWriter, *http.Request) error
 
 func HandleError(s *client.Schemas, t HandleFuncWithError) http.Handler {
 	return api.ApiHandler(s, http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
-		var err error
-		for i := 0; i < RetryCounts; i++ {
-			err = t(rw, req)
-			if !apierrors.IsConflict(errors.Cause(err)) {
-				break
-			}
-			logrus.WithError(err).Warn("Retry API call due to conflict")
-			time.Sleep(RetryInterval)
-		}
-		if err != nil {
+		if err := t(rw, req); err != nil {
 			logrus.Warnf("HTTP handling error %v", err)
 			apiContext := api.GetApiContext(req)
 			apiContext.WriteErr(err)


api/router.go
```diff
@@ -25,7 +27,15 @@ type HandleFuncWithError func(http.ResponseWriter, *http.Request) error
 func HandleError(s *client.Schemas, t HandleFuncWithError) http.Handler {
 	return api.ApiHandler(s, http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
 		var err error
+		copyRequestBody, err := ioutil.ReadAll(req.Body)
+		if err != nil {
+			logrus.Warnf("HTTP request copying error %v", err)
+			apiContext := api.GetApiContext(req)
+			apiContext.WriteErr(err)
```
derekbit, line 34: Should we return here?

mantissahz: Yes, modified. Thanks.

innobead (commented):
I am not sure if we still need to keep this resilience at the API level, because it looks weird to me. If it didn't work before, why do we need this?

Also, the change here looks incomplete as well, because the change is retrieving all data, but how to streaming handling here like backing image upload or download?

I would suggest just removing this retry loop if it did not work ever.

WDYT? @shuo-wu @derekbit @mantissahz

innobead:
Should we? If it did not take effect before, why do we need to add this?

mantissahz:
>>I agree to remove the retry part of HandleError then it may need to use >>util.RetryOnConflictCause() for update functions.

>Should we? If it did not take effect before, why do we need to add this?

The update might encounter conflict error as Derek mentioned in https://github.com/longhorn/longhorn-manager/pull/1671#issuecomment-1373022047.
And This is the old fix https://github.com/longhorn/longhorn/issues/132.

innobead:
> The update might encounter conflict error as Derek mentioned in [#1671 (comment)](https://github.com/longhorn/longhorn-manager/pull/1671#issuecomment-1373022047). And This is the old fix [longhorn/longhorn#132](https://github.com/longhorn/longhorn/issues/132).

so 132 was not actually fixed? then the issue stays until today. I am quite curious if nothing serious happened before, why do we need to add this?

mantissahz:
As discuss with @innobead and @derekbit, we should remove the retry part of API level/server side, and leave the retry in the backend if necessary when encountering `update` conflict error.

innobead (approved):
LGTM. 

Keeping the error handling here for later enrichment if we need something later. Suggest running the regression to see if any side effects will be caused by this change.

mantissahz:
Here is the running regression test https://ci.longhorn.io/job/private/job/longhorn-tests-regression/2992
and result almost like https://ci.longhorn.io/job/public/job/master/job/sles/job/amd64/job/longhorn-tests-sles-amd64/400/
All failed test cases PASSED except
"test_engine_upgrade.py::test_auto_upgrade_engine_to_default_version_dr_volume",
"test_ha.py::test_autosalvage_with_data_locality_enabled",
```text
Defaulted container "longhorn-test" out of: longhorn-test, longhorn-test-report
============================= test session starts ==============================
platform linux -- Python 3.9.15, pytest-5.3.1, py-1.11.0, pluggy-0.13.1 -- /usr/bin/python3.9
cachedir: .pytest_cache
plugins: order-1.0.1, repeat-0.9.1
collecting ... collected 5 items

test_cloning.py::test_cloning_with_detached_source_volume PASSED
test_node.py::test_node_eviction_multiple_volume PASSED
1536+0 records in
1536+0 records out
1610612736 bytes (1.6 GB, 1.5 GiB) copied, 20.5063 s, 78.5 MB/s
1536+0 records in
1536+0 records out
1610612736 bytes (1.6 GB, 1.5 GiB) copied, 20.6319 s, 78.1 MB/s
1536+0 records in
1536+0 records out
1610612736 bytes (1.6 GB, 1.5 GiB) copied, 21.8108 s, 73.8 MB/s
10+0 records in
10+0 records out
10 bytes copied, 4.4075e-05 s, 227 kB/s
test_snapshot.py::test_snapshot_hash_detect_corruption_in_global_enabled_mode 
snapshot 3454c79a-be39-4c7f-a52c-ebdcb5e7acea: checksum in resource=dbaaf032bb399dfc, checksum recalculated=dbaaf032bb399dfc
...
snapshot 3454c79a-be39-4c7f-a52c-ebdcb5e7acea: ctime recorded=2023-02-16 01:43:39.867991463 +0000 UTC, ctime=2023-02-16 01:43:39.867991463 +0000 UTC
PASSED
test_system_backup_restore.py::test_system_backup_and_restore_volume_with_data[s3] PASSED
test_system_backup_restore.py::test_system_backup_and_restore_volume_with_data[nfs] PASSED

=============================== warnings summary ===============================
tests/test_node.py::test_node_eviction_multiple_volume
tests/test_node.py::test_node_eviction_multiple_volume
tests/test_node.py::test_node_eviction_multiple_volume
tests/test_node.py::test_node_eviction_multiple_volume
tests/test_node.py::test_node_eviction_multiple_volume
  /usr/local/lib/python3.9/site-packages/kubernetes/client/rest.py:44: DeprecationWarning: HTTPResponse.getheaders() is deprecated and will be removed in urllib3 v2.1.0. Instead access HTTPResponse.headers directly.
    return self.urllib3_response.getheaders()

-- Docs: https://docs.pytest.org/en/latest/warnings.html
================== 5 passed, 5 warnings in 1290.71s (0:21:30) ==================
```

innobead:
> Here is the running regression test https://ci.longhorn.io/job/private/job/longhorn-tests-regression/2992 and result almost like https://ci.longhorn.io/job/public/job/master/job/sles/job/amd64/job/longhorn-tests-sles-amd64/400/ All failed test cases PASSED except "test_engine_upgrade.py::test_auto_upgrade_engine_to_default_version_dr_volume", "test_ha.py::test_autosalvage_with_data_locality_enabled",

https://github.com/longhorn/longhorn/issues/5376 
https://github.com/longhorn/longhorn/issues/4382 (flaky)

cc @longhorn/qa

mantissahz:
@Mergifyio backport v1.4.x

shuo-wu:
Sorry for the late reply.

> leave the retry in the backend if necessary when encountering update conflict error.

Do you mean asking the update functions in `datastore` to handle the conflict retry? If YES, it does not make sense, to resolve the conflict, the caller needs to re-get the CR then **re-apply the modifications** , which may be beyond the functions' responsibility.

mantissahz:
> Do you mean asking the update functions in datastore to handle the conflict retry? If YES, it does not make sense, to resolve the conflict, the caller needs to re-get the CR then re-apply the modifications , which may be beyond the functions' responsibility.

@shuo-wu what I think about is more like what Derek did in [recovery_backend/backend/configmap.go](https://github.com/longhorn/longhorn-manager/pull/1671/files#diff-004bf4ce869066e184b08bc61293b47d348edbcd9030820391a0cd5b51b46017R79) by using `util.RetryOnConflictCause`

PR #1703, opened 2023-02-16
feat(replica): RebuildRetryCount for disconnection

mantissahz:
Do not count the RebuildRetryCount if rebuilding reused replicas failed caused by network disconnection or node down.

Need to collect more failed reasons of rebuilding replicas.

longhorn/longhorn#1923

### DIFF
--- a/controller/engine_controller.go
+++ b/controller/engine_controller.go
@@ -1632,9 +1632,9 @@ func doesAddressExistInEngine(e *longhorn.Engine, addr string, engineClientProxy
 	return false, nil
 }
 
-func (ec *EngineController) startRebuilding(e *longhorn.Engine, replica, addr string) (err error) {
+func (ec *EngineController) startRebuilding(e *longhorn.Engine, replicaName, addr string) (err error) {
 	defer func() {
-		err = errors.Wrapf(err, "failed to start rebuild for %v of %v", replica, e.Name)
+		err = errors.Wrapf(err, "failed to start rebuild for %v of %v", replicaName, e.Name)
 	}()
 
 	log := ec.logger.WithFields(logrus.Fields{"volume": e.Spec.VolumeName, "engine": e.Name})
@@ -1653,7 +1653,7 @@ func (ec *EngineController) startRebuilding(e *longhorn.Engine, replica, addr st
 	}
 	// replica has already been added to the engine
 	if alreadyExists {
-		ec.logger.Debugf("Replica %v address %v has been added to the engine already", replica, addr)
+		ec.logger.Debugf("Replica %v address %v has been added to the engine already", replicaName, addr)
 		return nil
 	}
 
@@ -1735,19 +1735,34 @@ func (ec *EngineController) startRebuilding(e *longhorn.Engine, replica, addr st
 			log.Debug("Finished snapshot purge, will start rebuilding then")
 		}
 
+		replica, err := ec.ds.GetReplica(replicaName)
+		if err != nil {
+			log.WithError(err).Errorf("Failed to get replica %v unable to mark failed rebuild", replica)
+			return
+		}
+
+		// check and reset replica rebuild failed condition
+		replica, err = ec.updateReplicaRebuildFailedCondition(replica, "")
+		if err != nil {
+			log.WithError(err).Errorf("Failed to update rebuild status information on replica %v", replicaName)
+			return
+		}
+
 		// start rebuild
 		if e.Spec.RequestedBackupRestore != "" {
 			if e.Spec.NodeID != "" {
 				ec.eventRecorder.Eventf(e, v1.EventTypeNormal, constant.EventReasonRebuilding,
-					"Start rebuilding replica %v with Address %v for restore engine %v and volume %v", replica, addr, e.Name, e.Spec.VolumeName)
+					"Start rebuilding replica %v with Address %v for restore engine %v and volume %v", replicaName, addr, e.Name, e.Spec.VolumeName)
 				err = engineClientProxy.ReplicaAdd(e, replicaURL, true, fastReplicaRebuild, fileSyncHTTPClientTimeout)
 			}
 		} else {
 			ec.eventRecorder.Eventf(e, v1.EventTypeNormal, constant.EventReasonRebuilding,
-				"Start rebuilding replica %v with Address %v for normal engine %v and volume %v", replica, addr, e.Name, e.Spec.VolumeName)
+				"Start rebuilding replica %v with Address %v for normal engine %v and volume %v", replicaName, addr, e.Name, e.Spec.VolumeName)
 			err = engineClientProxy.ReplicaAdd(e, replicaURL, false, fastReplicaRebuild, fileSyncHTTPClientTimeout)
 		}
 		if err != nil {
+			replicaRebuildErrMsg := err.Error()
+
 			log.WithError(err).Errorf("Failed rebuilding of replica %v", addr)
 			ec.eventRecorder.Eventf(e, v1.EventTypeWarning, constant.EventReasonFailedRebuilding, "Failed rebuilding replica with Address %v: %v", addr, err)
 			// we've sent out event to notify user. we don't want to
@@ -1759,7 +1774,7 @@ func (ec *EngineController) startRebuilding(e *longhorn.Engine, replica, addr st
 				log.WithError(err).Errorf("Failed to remove rebuilding replica %v", addr)
 				ec.eventRecorder.Eventf(e, v1.EventTypeWarning, constant.EventReasonFailedDeleting,
 					"Failed to remove rebuilding replica %v with address %v for engine %v and volume %v due to rebuilding failure: %v",
-					replica, addr, e.Name, e.Spec.VolumeName, err)
+					replicaName, addr, e.Name, e.Spec.VolumeName, err)
 			} else {
 				log.Infof("Removed failed rebuilding replica %v", addr)
 			}
@@ -1768,30 +1783,31 @@ func (ec *EngineController) startRebuilding(e *longhorn.Engine, replica, addr st
 			// it is past the Backoff period, we'll try to mark the Replica as Failed and increase the Backoff period
 			// for the next failure.
 			if !ec.backoff.IsInBackOffSinceUpdate(e.Name, time.Now()) {
-				rep, err := ec.ds.GetReplica(replica)
+				replica, err = ec.updateReplicaRebuildFailedCondition(replica, replicaRebuildErrMsg)
 				if err != nil {
-					log.WithError(err).Errorf("Failed to get replica %v unable to mark failed rebuild", replica)
+					log.WithError(err).Errorf("Failed to update rebuild status information on replica %v", replicaName)
 					return
 				}
-				rep.Spec.FailedAt = util.Now()
-				rep.Spec.DesireState = longhorn.InstanceStateStopped
-				if _, err := ec.ds.UpdateReplica(rep); err != nil {
-					log.WithError(err).Errorf("Unable to mark failed rebuild on replica %v", replica)
+
+				replica.Spec.FailedAt = util.Now()
+				replica.Spec.DesireState = longhorn.InstanceStateStopped
+				if _, err := ec.ds.UpdateReplica(replica); err != nil {
+					log.WithError(err).Errorf("Unable to mark failed rebuild on replica %v", replicaName)
 					return
 				}
 				// Now that the Replica can actually be recreated, we can move up the Backoff.
 				ec.backoff.Next(e.Name, time.Now())
 				backoffTime := ec.backoff.Get(e.Name).Seconds()
-				log.Infof("Marked failed rebuild on replica %v, backoff period is now %v seconds", replica, backoffTime)
+				log.Infof("Marked failed rebuild on replica %v, backoff period is now %v seconds", replicaName, backoffTime)
 				return
 			}
-			log.Infof("Engine is still in backoff for replica %v rebuild failure", replica)
+			log.Infof("Engine is still in backoff for replica %v rebuild failure", replicaName)
 			return
 		}
 		// Replica rebuild succeeded, clear Backoff.
 		ec.backoff.DeleteEntry(e.Name)
 		ec.eventRecorder.Eventf(e, v1.EventTypeNormal, constant.EventReasonRebuilt,
-			"Replica %v with Address %v has been rebuilt for volume %v", replica, addr, e.Spec.VolumeName)
+			"Replica %v with Address %v has been rebuilt for volume %v", replicaName, addr, e.Spec.VolumeName)
 
 		// If enabled, call SnapshotPurge to clean up system generated snapshot after rebuilding.
 		if autoCleanupSystemGeneratedSnapshot {
@@ -1812,6 +1828,58 @@ func (ec *EngineController) startRebuilding(e *longhorn.Engine, replica, addr st
 		return err
 	}
 	return nil
+}
+
+// updateReplicaRebuildFailedCondition updates the rebuild failed condition if replica rebuilding failed
+func (ec *EngineController) updateReplicaRebuildFailedCondition(replica *longhorn.Replica, errMsg string) (*longhorn.Replica, error) {
+	replicaRebuildFailedReason, conditionStatus, err := ec.getReplicaRebuildFailedReason(replica.Spec.NodeID, errMsg)
+	if err != nil {
+		return nil, err
+	}
+
+	replica.Status.Conditions = types.SetCondition(
+		replica.Status.Conditions,
+		longhorn.ReplicaConditionTypeRebuildFailed,
+		conditionStatus,
+		replicaRebuildFailedReason,
+		errMsg)
+
+	replica, err = ec.ds.UpdateReplicaStatus(replica)
+
+	return replica, err
+}
+
+func (ec *EngineController) getReplicaRebuildFailedReason(replicaNodeID, errMsg string) (failedReason string, conditionStatus longhorn.ConditionStatus, err error) {
+	failedReason, conditionStatus, isRebuildingFailedByNetwork := getReplicaRebuildFailedReasonFromError(errMsg)
+	if isRebuildingFailedByNetwork {
+		replicaNode, err := ec.ds.GetNodeRO(replicaNodeID)
+		if err != nil {
+			return "", "", err
+		}
+
+		replicaRebuildFailedCondition := types.GetCondition(replicaNode.Status.Conditions, longhorn.NodeConditionTypeReady)
+		switch replicaRebuildFailedCondition.Reason {
+		case longhorn.NodeConditionReasonManagerPodDown, longhorn.NodeConditionReasonKubernetesNodeGone, longhorn.NodeConditionReasonKubernetesNodeNotReady:
+			failedReason = replicaRebuildFailedCondition.Reason
+		}
+	}
+
+	return failedReason, conditionStatus, nil
+}
+
+func getReplicaRebuildFailedReasonFromError(errMsg string) (string, longhorn.ConditionStatus, bool) {
+	switch {
+	case strings.Contains(errMsg, longhorn.ReplicaRebuildFailedCanceledErrorMSG):
+		fallthrough
+	case strings.Contains(errMsg, longhorn.ReplicaRebuildFailedDeadlineExceededErrorMSG):
+		fallthrough
+	case strings.Contains(errMsg, longhorn.ReplicaRebuildFailedUnavailableErrorMSG):
+		return longhorn.ReplicaConditionReasonRebuildFailedDisconnection, longhorn.ConditionStatusTrue, true
+	case errMsg == "":
+		return "", longhorn.ConditionStatusFalse, false
+	default:
+		return longhorn.ReplicaConditionReasonRebuildFailedGeneral, longhorn.ConditionStatusTrue, false
+	}
 }
 
 func (ec *EngineController) Upgrade(e *longhorn.Engine) (err error) {
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -1829,8 +1829,12 @@ func (vc *VolumeController) replenishReplicas(v *longhorn.Volume, e *longhorn.En
 				log.Debugf("Failed replica %v will be reused during rebuilding", reusableFailedReplica.Name)
 				reusableFailedReplica.Spec.FailedAt = ""
 				reusableFailedReplica.Spec.HealthyAt = ""
-				reusableFailedReplica.Spec.RebuildRetryCount++
+
+				if datastore.IsReplicaRebuildingFailed(reusableFailedReplica) {
+					reusableFailedReplica.Spec.RebuildRetryCount++
+				}
 				vc.backoff.Next(reusableFailedReplica.Name, time.Now())
+
 				rs[reusableFailedReplica.Name] = reusableFailedReplica
 				continue
 			}
--- a/datastore/longhorn.go
+++ b/datastore/longhorn.go
@@ -1257,6 +1257,22 @@ func IsAvailableHealthyReplica(r *longhorn.Replica) bool {
 	return true
 }
 
+// IsReplicaRebuildingFailed returns true if the rebuilding replica failed not caused by network issues.
+func IsReplicaRebuildingFailed(reusableFailedReplica *longhorn.Replica) bool {
+	replicaRebuildFailedCondition := types.GetCondition(reusableFailedReplica.Status.Conditions, longhorn.ReplicaConditionTypeRebuildFailed)
+
+	if replicaRebuildFailedCondition.Status != longhorn.ConditionStatusTrue {
+		return true
+	}
+
+	switch replicaRebuildFailedCondition.Reason {
+	case longhorn.ReplicaConditionReasonRebuildFailedDisconnection, longhorn.NodeConditionReasonManagerPodDown, longhorn.NodeConditionReasonKubernetesNodeGone, longhorn.NodeConditionReasonKubernetesNodeNotReady:
+		return false
+	default:
+		return true
+	}
+}
+
 // GetOwnerReferencesForEngineImage returns OwnerReference for the given
 // Longhorn EngineImage name and UID
 func GetOwnerReferencesForEngineImage(ei *longhorn.EngineImage) []metav1.OwnerReference {
--- a/k8s/pkg/apis/longhorn/v1beta2/replica.go
+++ b/k8s/pkg/apis/longhorn/v1beta2/replica.go
@@ -2,6 +2,20 @@ package v1beta2
 
 import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
+const (
+	// error messages captured from the gRPC error code
+	ReplicaRebuildFailedCanceledErrorMSG         = "rpc error: code = Canceled"
+	ReplicaRebuildFailedDeadlineExceededErrorMSG = "rpc error: code = DeadlineExceeded"
+	ReplicaRebuildFailedUnavailableErrorMSG      = "rpc error: code = Unavailable"
+)
+
+const (
+	ReplicaConditionTypeRebuildFailed = "RebuildFailed"
+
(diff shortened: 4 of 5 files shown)


controller/engine_controller.go
```diff
@@ -1814,6 +1828,58 @@ func (ec *EngineController) startRebuilding(e *longhorn.Engine, replica, addr st
...
+		for _, con := range replicaKubernetesNode.Status.Conditions {
+			if con.Type != v1.NodeReady {
+				continue
+			}
+			if con.Status != v1.ConditionTrue {
+				replicaRebuildFailedReason = longhorn.ReplicaConditionReasonRebuildFailedNodeDown
+			}
+			break
+		}
+	}
```
shuo-wu, line 1853: NIT: You can rename `getReplicaRebuildFailedReason` to `getRebuildFailedReasonFromError`, then have a new function `getReplicaRebuildFailedReason` includes `getRebuildFailedReasonFromError`and this part. You can input the node CR or `ds` for the function

shuo-wu (approved):
Others LGTM

controller/engine_controller.go
```diff
@@ -1814,6 +1828,58 @@ func (ec *EngineController) startRebuilding(e *longhorn.Engine, replica, addr st
...
+	case strings.Contains(errMsg, longhorn.ReplicaRebuildFailedConnectionTimeoutErrorMSG):
+		return longhorn.ReplicaConditionReasonRebuildFailedConnectionTimeout, longhorn.ConditionStatusTrue, true
+	case strings.Contains(errMsg, longhorn.ReplicaRebuildFailedConnectionRefusedErrorMSG):
+		return longhorn.ReplicaConditionReasonRebuildFailedConnectionRefused, longhorn.ConditionStatusTrue, true
+	case strings.Contains(errMsg, longhorn.ReplicaRebuildFailedReadingServerEOFErrorMSG):
+		return longhorn.ReplicaConditionReasonRebuildFailedReadingServerEOF, longhorn.ConditionStatusTrue, true
+	case errMsg == "":
+		return "", longhorn.ConditionStatusFalse, false
+	default:
+		return longhorn.ReplicaConditionReasonRebuildFailedUnknown, longhorn.ConditionStatusTrue, false
```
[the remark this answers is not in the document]
mantissahz: Modified. Thanks, looking for the right name as well, that's good.

controller/volume_controller.go
```diff
@@ -1870,6 +1875,25 @@ func getRebuildingReplicaCount(e *longhorn.Engine) int {
 	return rebuilding
 }
 
+func isRebuildingFailedByNetwork(rebuildFailedReason string) bool {
```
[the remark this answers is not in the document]
mantissahz: Moved. Thank you.

controller/engine_controller.go
mantissahz: Modified. Renamed to `getRebuildFailedReasonFromError` and add a new `getRebuildFailedReason` with getting kubernetes node status

shuo-wu (approved):

mantissahz:
After some testings, both node is down or network of the node is unstable,
`RebuildRetryCount` that is not increased will not make the rebuilding mechanism into an infinite loop because a replica would be replenished in `SettingNameReplicaReplenishmentWaitInterval` seconds (default is 600 sec.) if there is another available node and failed replica on the down or unstable networking node will be deleted when reaching the `volume.spec.staleReplicaTimeout` (minutes)

controller/engine_controller.go
```diff
@@ -1814,6 +1828,67 @@ func (ec *EngineController) startRebuilding(e *longhorn.Engine, replica, addr st
...
+		return longhorn.ReplicaConditionReasonRebuildFailedConnectionTimeout, longhorn.ConditionStatusTrue, true
+	case strings.Contains(errMsg, longhorn.ReplicaRebuildFailedConnectionRefusedErrorMSG):
+		return longhorn.ReplicaConditionReasonRebuildFailedConnectionRefused, longhorn.ConditionStatusTrue, true
+	case strings.Contains(errMsg, longhorn.ReplicaRebuildFailedReadingServerEOFErrorMSG):
+		return longhorn.ReplicaConditionReasonRebuildFailedReadingServerEOF, longhorn.ConditionStatusTrue, true
+	case errMsg == "":
+		return "", longhorn.ConditionStatusFalse, false
+	default:
+		return longhorn.ReplicaConditionReasonRebuildFailedGeneral, longhorn.ConditionStatusTrue, false
+	}
```
innobead, line 1889: Because the error you handled is from a gRPC call, so basically you should be able to check gRPC error code to determine what kind of error it is. 

Suggest using `status.FromError` to get the converted gRPC error then check its error code. If the code is `Unavailable`, `Canceled` or `DeadlineExceeded`, then we could guess that's caused by unstable network. You don't need to hard code message then do regex or str check to determine the error type anymore. 

If you would like to cover the node condition from node status to superseded the gRPC error code, and it should be fine. However, I would say, let's just focus on the gRPC code which will make the thing clear.

innobead (changes requested):

controller/engine_controller.go
innobead: ref: backing image manager used that as well for different code

[/pkg/util/util.go#L81-L84](https://github.com/longhorn/backing-image-manager/blob/3cb922d175822c1b9b22fc0573abf6a899468648/pkg/util/util.go#L81-L84)
```go
func IsGRPCErrorMatchingCode(err error, errCode codes.Code) bool {
	gRPCStatus, ok := status.FromError(err)
	return ok && gRPCStatus.Code() == errCode
}
```

mantissahz: I tried to get the gRPC status by `status.FromError`, it seems that we have two layers of the gRPC call from `instance-manager` to `engine`. The message would be like this:
```txt
      'proxyServer=10.42.3.40:8501 destination=10.42.3.40:10000: failed to
      add replica tcp://10.42.1.38:10000 for volume: rpc error: code = Unknown desc
      = failed to sync files [{FromFileName:volume-snap-789558f4-89af-42d6-8e1d-f615ce067d14.img
      ToFileName:volume-snap-789558f4-89af-42d6-8e1d-f615ce067d14.img ActualSize:0}
      {FromFileName:volume-snap-789558f4-89af-42d6-8e1d-f615ce067d14.img.meta ToFileName:volume-snap-789558f4-89af-42d6-8e1d-f615ce067d14.img.meta
      ActualSize:0} {FromFileName:volume-snap-9db919f3-1737-410b-84f5-00743278ee5c.img
      ToFileName:volume-snap-9db919f3-1737-410b-84f5-00743278ee5c.img ActualSize:10737442816}
      {FromFileName:volume-snap-9db919f3-1737-410b-84f5-00743278ee5c.img.meta ToFileName:volume-snap-9db919f3-1737-410b-84f5-00743278ee5c.img.meta
      ActualSize:0}] from tcp://10.42.4.41:10000: rpc error: code = Unavailable desc
      = error reading from server: read tcp 10.42.3.40:47520->10.42.1.38:10002: read:
      connection reset by peer'
```
I can get the first gRPC status from errors `rpc error: code = Unknown desc = failed to sync files ...` returned by instance-manager but it might be always `Unknown` and it seems that second error `rpc error: code = Unavailable desc = error reading from server: read tcp 10.42.3.40:47520->10.42.1.38:10002: read: connection reset by peer'` is wrapped into the message of first error and it can not be transformed back to the gRPC status (I don't have any idea).
So I scale down the number of errMsg to `code = Canceled`, `code = DeadlineExceeded` and `code = Unavailable` to check if the error is caused by network.

innobead: Exactly, this is not what I thought, because I assumed the error should be just returned back w/o any modification because the call is from the proxy. However, it's actually modified in the replica client which is used by the proxy server.

[/vendor/github.com/longhorn/longhorn-engine/pkg/replica/client/client.go#L503-L511](https://github.com/longhorn/longhorn-instance-manager/blob/cb92c147d5f8deb55cbe776d7d6f42e1f22ae2f3/vendor/github.com/longhorn/longhorn-engine/pkg/replica/client/client.go#L503-L511)
```go
	if _, err := syncAgentServiceClient.FilesSync(ctx, &ptypes.FilesSyncRequest{
		FromAddress:               fromAddress,
		ToHost:                    c.host,
		SyncFileInfoList:          syncFileInfoListToSyncAgentGRPCFormat(list),
		FastSync:                  fastSync,
		FileSyncHttpClientTimeout: int32(fileSyncHTTPClientTimeout),
	}); err != nil {
		return errors.Wrapf(err, "failed to sync files %+v from %v", list, fromAddress)
	}
```

controller/engine_controller.go
```diff
@@ -1803,6 +1810,13 @@ func (ec *EngineController) startRebuilding(e *longhorn.Engine, replica, addr st
 			}
 			logrus.Debug("Started snapshot purge after rebuilding")
 		}
+
+		// check and reset replica rebuild failed condition
+		if err := updateReplicaRebuildFailedCondition(log, ec.ds, replica, ""); err != nil {
+			log.WithError(err).Errorf("Failed to update rebuild status information on replica %v", replica)
+			return
+		}
```
innobead, line 1818: This should be moved to the beginning when starting rebuilding to avoid any leftover condition still when the replica is rebuilding.

controller/engine_controller.go
```diff
@@ -1768,6 +1770,11 @@ func (ec *EngineController) startRebuilding(e *longhorn.Engine, replica, addr st
 			// it is past the Backoff period, we'll try to mark the Replica as Failed and increase the Backoff period
 			// for the next failure.
 			if !ec.backoff.IsInBackOffSinceUpdate(e.Name, time.Now()) {
+				if err := updateReplicaRebuildFailedCondition(log, ec.ds, replica, replicaRebuildErrMsg); err != nil {
+					log.WithError(err).Errorf("Failed to update rebuild status information on replica %v", replica)
+					return
+				}
```
innobead, line 1776: Suggest

```go
if err := updateReplicaRebuildFailedCondition(log, ec.ds, rep, replicaRebuildErrMsg); err != nil {
    log.WithError(err).Errorf("Failed to update rebuild status information on replica %v", replica)
    return
}
```

and move the below code to the above and outer scope, so you don't need to repeat querying the replica CR..

```go
rep, err := ec.ds.GetReplica(replica)
if err != nil {
    log.WithError(err).Errorf("Failed to get replica %v unable to mark failed rebuild", replica)
    return
}
```

controller/engine_controller.go
```diff
@@ -1814,6 +1828,63 @@ func (ec *EngineController) startRebuilding(e *longhorn.Engine, replica, addr st
...
+	}
+
+	replica.Status.Conditions = types.SetCondition(
+		replica.Status.Conditions, longhorn.ReplicaConditionTypeRebuildFailed,
+		conditionStatus, replicaRebuildFailedReason, errMsg)
+
+	_, err = ds.UpdateReplicaStatus(replica)
+
+	return err
+}
```
innobead, line 1850: Change to the below.

```go
func (ec *EngineController) updateReplicaRebuildFailedCondition(log *logrus.Entry, replica *Replica, errMsg string) error {
    replicaRebuildFailedReason, conditionStatus, err := getReplicaRebuildFailedReason(ds, replica.Spec.NodeID, errMsg)
    if err != nil {
        return err
    }
    
    replica.Status.Conditions = types.SetCondition(
        replica.Status.Conditions, 
        longhorn.ReplicaConditionTypeRebuildFailed,
        conditionStatus, 
        replicaRebuildFailedReason, 
        errMsg
    )
    _, err = ec.ds.UpdateReplicaStatus(replica)
    
    return err
}
```

controller/engine_controller.go
```diff
@@ -1814,6 +1828,63 @@ func (ec *EngineController) startRebuilding(e *longhorn.Engine, replica, addr st
...
+
+func getReplicaRebuildFailedReason(ds *datastore.DataStore, replicaNodeID, errMsg string) (string, longhorn.ConditionStatus, error) {
+	replicaRebuildFailedReason, conditionStatus, isRebuildingFailedByNetwork := getReplicaRebuildFailedReasonFromError(errMsg)
+	if isRebuildingFailedByNetwork {
+		replicaKubernetesNode, err := ds.GetKubernetesNode(replicaNodeID)
+		if err != nil {
+			return "", "", err
+		}
+		for _, con := range replicaKubernetesNode.Status.Conditions {
+			if con.Type != v1.NodeReady {
```
innobead, line 1860: Suggest using longhorn node resource condition instead to avoid duplicated query. 

Also, NotReady doesn't mean `NodeDown`.

controller/engine_controller.go
```diff
@@ -1814,6 +1828,63 @@ func (ec *EngineController) startRebuilding(e *longhorn.Engine, replica, addr st
...
+		replica.Status.Conditions, longhorn.ReplicaConditionTypeRebuildFailed,
+		conditionStatus, replicaRebuildFailedReason, errMsg)
+
+	_, err = ds.UpdateReplicaStatus(replica)
+
+	return err
+}
+
+func getReplicaRebuildFailedReason(ds *datastore.DataStore, replicaNodeID, errMsg string) (string, longhorn.ConditionStatus, error) {
+	replicaRebuildFailedReason, conditionStatus, isRebuildingFailedByNetwork := getReplicaRebuildFailedReasonFromError(errMsg)
```
innobead, line 1853: Change to

```go
func (ec *EngineController) getReplicaRebuildFailedReason(replica *Replica, errMsg string) (reason string, condition longhorn.ConditionStatus, err error) {
...
```

If having multiple returned variables, suggest named variables.

innobead (changes requested):

k8s/pkg/apis/longhorn/v1beta2/replica.go
```diff
@@ -2,6 +2,20 @@ package v1beta2
...
 import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
+const (
+	ReplicaRebuildFailedCanceledErrorMSG         = "rpc error: code = Canceled"
+	ReplicaRebuildFailedDeadlineExceededErrorMSG = "rpc error: code = DeadlineExceeded"
+	ReplicaRebuildFailedUnavailableErrorMSG      = "rpc error: code = Unavailable"
+)
+
+const (
+	ReplicaConditionTypeRebuildFailed = "rebuildfailed"
```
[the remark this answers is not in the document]
mantissahz: Modified to `RebuildFailed` and deleted `ReplicaConditionReasonRebuildFailedNodeDown` then use longhorn node condition

k8s/pkg/apis/longhorn/v1beta2/replica.go
```diff
@@ -2,6 +2,20 @@ package v1beta2
 
 import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
+const (
+	ReplicaRebuildFailedCanceledErrorMSG         = "rpc error: code = Canceled"
+	ReplicaRebuildFailedDeadlineExceededErrorMSG = "rpc error: code = DeadlineExceeded"
+	ReplicaRebuildFailedUnavailableErrorMSG      = "rpc error: code = Unavailable"
```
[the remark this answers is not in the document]
mantissahz: Added. Thanks.

datastore/longhorn.go
```diff
@@ -1257,6 +1257,18 @@ func IsAvailableHealthyReplica(r *longhorn.Replica) bool {
...
+// IsReplicaRebuildingFailedByNetwork returns if the rebuilding replica failed caused by network issues.
+func IsReplicaRebuildingFailedByNetwork(rebuildFailedReason string) bool {
+	switch rebuildFailedReason {
+	case longhorn.ReplicaConditionReasonRebuildFailedDisconnection:
+		fallthrough
+	case longhorn.ReplicaConditionReasonRebuildFailedNodeDown:
+		return true
+	default:
+		return false
+	}
```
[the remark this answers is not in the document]
mantissahz: Modified and use longhorn node resource condition `NodeConditionReasonManagerPodDown`, `NodeConditionReasonKubernetesNodeGone`, `NodeConditionReasonKubernetesNodeNotReady` instead. Thanks.

datastore/longhorn.go
```diff
@@ -1257,6 +1257,18 @@ func IsAvailableHealthyReplica(r *longhorn.Replica) bool {
 	return true
 }
 
+// IsReplicaRebuildingFailedByNetwork returns if the rebuilding replica failed caused by network issues.
+func IsReplicaRebuildingFailedByNetwork(rebuildFailedReason string) bool {
```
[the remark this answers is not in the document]
mantissahz: Renamed. Thanks.

controller/volume_controller.go
```diff
@@ -1829,8 +1829,13 @@ func (vc *VolumeController) replenishReplicas(v *longhorn.Volume, e *longhorn.En
 				log.Debugf("Failed replica %v will be reused during rebuilding", reusableFailedReplica.Name)
 				reusableFailedReplica.Spec.FailedAt = ""
 				reusableFailedReplica.Spec.HealthyAt = ""
-				reusableFailedReplica.Spec.RebuildRetryCount++
+
+				replicaRebuildFailedCondition := types.GetCondition(reusableFailedReplica.Status.Conditions, longhorn.ReplicaConditionTypeRebuildFailed)
+				if replicaRebuildFailedCondition.Status != longhorn.ConditionStatusTrue || !datastore.IsReplicaRebuildingFailedByNetwork(replicaRebuildFailedCondition.Reason) {
+					reusableFailedReplica.Spec.RebuildRetryCount++
+				}
```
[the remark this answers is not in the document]
mantissahz: Modified and move the condition check to the method. Thanks

controller/engine_controller.go
mantissahz: Moved to the beginning when starting rebuilding. Thanks

controller/engine_controller.go
mantissahz: Modified. And Move `GetReplica` out of the goroutine function. Thanks

controller/engine_controller.go
mantissahz: Modified. Thanks

controller/engine_controller.go
mantissahz: Removed getting KubernetesNode, and use longhorn node resource condition instead. Thanks

controller/engine_controller.go
mantissahz: Modified and added named variables.
Thank you.

controller/engine_controller.go
```diff
@@ -1657,6 +1657,11 @@ func (ec *EngineController) startRebuilding(e *longhorn.Engine, replica, addr st
 		return nil
 	}
 
+	replicaObj, err := ec.ds.GetReplica(replica)
```
innobead, line 1660: Can we rename replica to replicaName and replace replicaObj with replica? Thanks.

innobead (changes requested):

controller/engine_controller.go
```diff
@@ -1768,6 +1781,11 @@ func (ec *EngineController) startRebuilding(e *longhorn.Engine, replica, addr st
 			// it is past the Backoff period, we'll try to mark the Replica as Failed and increase the Backoff period
 			// for the next failure.
 			if !ec.backoff.IsInBackOffSinceUpdate(e.Name, time.Now()) {
+				if err := ec.updateReplicaRebuildFailedCondition(replicaObj, replicaRebuildErrMsg); err != nil {
+					log.WithError(err).Errorf("Failed to update rebuild status information on replica %v", replica)
+					return
+				}
```
innobead, line 1787: The code below here is not required, because you already have queried `replica` already.

```go
rep, err := ec.ds.GetReplica(replica)
if err != nil {
`log.WithError(err).Errorf("Failed to get replica %v unable to mark failed rebuild", replica)
`return
}
```

innobead (approved):
In general, LGTM.

innobead:
@mantissahz Please update the test result, then we can merge this. Thanks.

controller/engine_controller.go
mantissahz: Renamed. Thanks

controller/engine_controller.go
mantissahz: Removed. Thanks.

controller/engine_controller.go
```diff
@@ -1768,30 +1781,30 @@ func (ec *EngineController) startRebuilding(e *longhorn.Engine, replica, addr st
...
 				}
-				rep.Spec.FailedAt = util.Now()
-				rep.Spec.DesireState = longhorn.InstanceStateStopped
-				if _, err := ec.ds.UpdateReplica(rep); err != nil {
-					log.WithError(err).Errorf("Unable to mark failed rebuild on replica %v", replica)
+
+				replica.Spec.FailedAt = util.Now()
+				replica.Spec.DesireState = longhorn.InstanceStateStopped
+				if _, err := ec.ds.UpdateReplica(replica); err != nil {
+					log.WithError(err).Errorf("Unable to mark failed rebuild on replica %v", replicaName)
```
innobead, line 1792: If somehow the replica update failed, but the previous condition has been updated, so it will cause inconsistent temporarily; however, it should be eventually consistent. 

This looks like we can' make the spec and status update at once to cause potential non-atomic translation update.

innobead (dismissed):
LGTM

controller/engine_controller.go
mantissahz: Removing 'getting the newest replica' will cause replica update failed and get into an infinite loop. Just Put it back.

mantissahz:
Regression test for rebuilding part is here https://ci.longhorn.io/job/private/job/longhorn-tests-regression/3016

Manual tests: (use [network_down.sh](https://github.com/longhorn/longhorn/files/4864127/network_down.sh.zip) to simulate network unstable)
1. Make network down and up repeatedly on node A of replica A:
   It can get `ReplicaConditionReasonRebuildFailedDisconnection` or `NodeConditionReasonManagerPodDown` failed reasons and `rebuildRetryCount` will not be increased
   Sometimes it will get `replica must be closed, cannot add in state` and `rebuildRetryCount` will be increased.
   Replica can be rebuilt normally if network is back to be stable.
2. A new replica will be replenished depended on `replica-replenishment-wait-interval` if replica A is never rebuilt completely and replica A will be deleted depended on `staleReplicaTimeout`

controller/engine_controller.go
innobead: Exactly, because our implementation is not always atomic :slightly_frowning_face: and this will require us to get the new resource update before the next update.

mantissahz:
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/3018
Rerun for failed case `test_settings.py::test_setting_concurrent_rebuild_limit` and `PASSED`

innobead:
> https://ci.longhorn.io/job/private/job/longhorn-tests-regression/3018 Rerun for failed case `test_settings.py::test_setting_concurrent_rebuild_limit` and `PASSED`

cc @longhorn/qa need to pay attention to this test to see if it's flaky.

controller/engine_controller.go
```diff
@@ -1814,6 +1832,58 @@ func (ec *EngineController) startRebuilding(e *longhorn.Engine, replica, addr st
...
+
+	replica.Status.Conditions = types.SetCondition(
+		replica.Status.Conditions,
+		longhorn.ReplicaConditionTypeRebuildFailed,
+		conditionStatus,
+		replicaRebuildFailedReason,
+		errMsg)
+
+	_, err = ec.ds.UpdateReplicaStatus(replica)
+
```
innobead, line 1850: Just noticed, we can get the updated replica from `UpdateReplicaStatus` instead of ignoring it, so we can use the returned replica later.

controller/engine_controller.go
```diff
@@ -1653,10 +1653,15 @@ func (ec *EngineController) startRebuilding(e *longhorn.Engine, replica, addr st
...
 	if alreadyExists {
-		ec.logger.Debugf("Replica %v address %v has been added to the engine already", replica, addr)
+		ec.logger.Debugf("Replica %v address %v has been added to the engine already", replicaName, addr)
 		return nil
 	}
 
+	replica, err := ec.ds.GetReplica(replicaName)
+	if err != nil {
+		return err
+	}
```
innobead, line 1663: This replica is used in the below goroutine only, so let's move into it and just before `ec.updateReplicaRebuildFailedCondition(replica, "")`.

controller/engine_controller.go
```diff
@@ -1735,19 +1740,27 @@ func (ec *EngineController) startRebuilding(e *longhorn.Engine, replica, addr st
 			log.Debug("Finished snapshot purge, will start rebuilding then")
 		}
 
+		// check and reset replica rebuild failed condition
+		if err := ec.updateReplicaRebuildFailedCondition(replica, ""); err != nil {
+			log.WithError(err).Errorf("Failed to update rebuild status information on replica %v", replicaName)
+			return
+		}
```
innobead, line 1747: Can we change to the below, so later we should not need to requery the replica.

```go
replica, err = ec.updateReplicaRebuildFailedCondition(replica, "")
if err != nil {
    log.WithError(err).Errorf("Failed to update rebuild status information on replica %v", replicaName)
    return
}
```

controller/engine_controller.go
```diff
@@ -1768,30 +1781,35 @@ func (ec *EngineController) startRebuilding(e *longhorn.Engine, replica, addr st
 			// it is past the Backoff period, we'll try to mark the Replica as Failed and increase the Backoff period
 			// for the next failure.
 			if !ec.backoff.IsInBackOffSinceUpdate(e.Name, time.Now()) {
-				rep, err := ec.ds.GetReplica(replica)
+				if err := ec.updateReplicaRebuildFailedCondition(replica, replicaRebuildErrMsg); err != nil {
+					log.WithError(err).Errorf("Failed to update rebuild status information on replica %v", replicaName)
+					return
+				}
```
innobead, line 1787: Change to below. The same reason as above. 

```go
replica, err = ec.updateReplicaRebuildFailedCondition(replica, replicaRebuildErrMsg)
if err != nil {
    log.WithError(err).Errorf("Failed to update rebuild status information on replica %v", replicaName)
    return
}
```

controller/engine_controller.go
```diff
@@ -1768,30 +1781,35 @@ func (ec *EngineController) startRebuilding(e *longhorn.Engine, replica, addr st
...
+
+				replica, err := ec.ds.GetReplica(replicaName)
```
innobead, line 1789: This replica query should not be needed any more.

controller/engine_controller.go
```diff
@@ -1814,6 +1832,58 @@ func (ec *EngineController) startRebuilding(e *longhorn.Engine, replica, addr st
 	return nil
 }
 
+// updateReplicaRebuildFailedCondition updates the rebuild failed condition if replica rebuilding failed
+func (ec *EngineController) updateReplicaRebuildFailedCondition(replica *longhorn.Replica, errMsg string) error {
```
innobead, line 1836: Change to 

```go
func (ec *EngineController) updateReplicaRebuildFailedCondition(replica *longhorn.Replica, errMsg string) (*longhorn.Replica, error) {
```

innobead (changes requested):
LGTM, but some minor suggestions we should better adjust a bit.

controller/engine_controller.go
mantissahz: Modified, it will return updated replica. Thanks

controller/engine_controller.go
mantissahz: Changed. Thanks

controller/engine_controller.go
mantissahz: Changed. Thanks

controller/engine_controller.go
mantissahz: Removed. Thanks

controller/engine_controller.go
mantissahz: Changed.
Many Thanks.

mantissahz:
Manual Tests just like https://github.com/longhorn/longhorn-manager/pull/1703#issuecomment-1441368785 works as expected.
And regression test for the rebuilding part is running: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/3019

innobead (approved):

innobead:
@Mergifyio backport v1.4.x

PR #1704, opened 2023-02-16
fix(restore): sync backup target when activate DR volume

weizhe0422:
[Longhorn 5292](https://github.com/longhorn/longhorn/issues/5292)

 - the explanation for the fix 
   - When activate the DR volume, it will trigger to synchronize the backup volume and backups. Then trigger the replicas of engine to restore from the latest backup.
- test steps
    1. with 2 clusters w/ same backup target: cluster_a & cluster_b
    2. deploy a volume w/ 5G data in cluster_a
    3. create a backup for the volume in cluster_a
    4. create a DR volume from the backup volume in cluster_b
    5. extend `Backupstore Poll Interval` setting value, e.g., 1000 in cluster_b
    6. write more data into the volume in cluster_a
    7. activate the DR volume in cluster_b
    8. show `restoring with latest backup in progress` if it is restoring w/ last backup
    9. check the activated DR volume should be with the latest data
 - e2e
   - test_basic
     image
   - test_csi
     image

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -2916,7 +2916,23 @@ func (vc *VolumeController) checkForAutoDetachment(v *longhorn.Volume, e *longho
 			break
 		}
 	}
-	if !(e.Spec.RequestedBackupRestore != "" && e.Spec.RequestedBackupRestore == e.Status.LastRestoredBackup &&
+
+	// After the volume is detached, the engine stops and does not perform recovery,
+	// so it should ensure that the backup volume is synced and updated at least once
+	// so that the engine restores with the latest backup before the volume is detached
+	if backupVolumeName, isExist := v.Labels[types.LonghornLabelBackupVolume]; isExist && backupVolumeName != "" {
+		backupVolume, err := vc.ds.GetBackupVolumeRO(backupVolumeName)
+		if err != nil {
+			return errors.Wrapf(err, "failed to get backup volume: %v", v.Name)
+		}
+		if backupVolume.Status.LastSyncedAt.Before(&backupVolume.Spec.SyncRequestedAt) {
+			return nil
+		}
+	}
+
+	// make sure engine finish restoring with the latest backup and no longer a DR volume
+	if !(e.Spec.RequestedBackupRestore != "" &&
+		e.Spec.RequestedBackupRestore == e.Status.LastRestoredBackup &&
 		!v.Spec.Standby) {
 		return nil
 	}
--- a/manager/volume.go
+++ b/manager/volume.go
@@ -3,6 +3,7 @@ package manager
 import (
 	"fmt"
 	"strconv"
+	"time"
 
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
@@ -471,23 +472,10 @@ func (m *VolumeManager) Activate(volumeName string, frontend string) (v *longhor
 		return nil, err
 	}
 
-	var engine *longhorn.Engine
-	es, err := m.ds.ListVolumeEngines(v.Name)
-	if err != nil {
-		return nil, fmt.Errorf("failed to list engines for volume %v: %v", v.Name, err)
-	}
-	if len(es) != 1 {
-		return nil, fmt.Errorf("found more than 1 engines for volume %v", v.Name)
-	}
-	for _, e := range es {
-		engine = e
-	}
-
-	if v.Status.LastBackup != engine.Status.LastRestoredBackup || engine.Spec.RequestedBackupRestore != engine.Status.LastRestoredBackup {
-		logrus.Infof("Standby volume %v will be activated after finishing restoration, "+
-			"backup volume's latest backup: %v, "+
-			"engine requested backup restore: %v, engine last restored backup: %v",
-			v.Name, v.Status.LastBackup, engine.Spec.RequestedBackupRestore, engine.Status.LastRestoredBackup)
+	// Trigger a backup volume update to get the latest backup
+	// and will confirm recovery completion in volume state reconciliation
+	if err := m.triggerBackupVolumeToSync(v); err != nil {
+		return nil, err
 	}
 
 	v.Spec.Frontend = longhorn.VolumeFrontend(frontend)
@@ -499,6 +487,25 @@ func (m *VolumeManager) Activate(volumeName string, frontend string) (v *longhor
 
 	logrus.Debugf("Activating volume %v with frontend %v", v.Name, frontend)
 	return v, nil
+}
+
+func (m *VolumeManager) triggerBackupVolumeToSync(volume *longhorn.Volume) error {
+	backupVolumeName, isExist := volume.Labels[types.LonghornLabelBackupVolume]
+	if !isExist || backupVolumeName == "" {
+		return errors.Errorf("cannot find the backup volume label for volume: %v", volume.Name)
+	}
+
+	backupVolume, err := m.ds.GetBackupVolume(backupVolumeName)
+	if err != nil {
+		return errors.Wrapf(err, "failed to get backup volume: %v", backupVolumeName)
+	}
+	requestSyncTime := metav1.Time{Time: time.Now().UTC()}
+	backupVolume.Spec.SyncRequestedAt = requestSyncTime
+	if _, err = m.ds.UpdateBackupVolume(backupVolume); err != nil {
+		return errors.Wrapf(err, "failed to update backup volume: %v", backupVolumeName)
+	}
+
+	return nil
 }
 
 func (m *VolumeManager) Expand(volumeName string, size int64) (v *longhorn.Volume, err error) {


engineapi/backups.go
```diff
@@ -20,9 +20,9 @@ import (
 )
 
 const (
-	backupStateInProgress = "in_progress"
-	backupStateComplete   = "complete"
-	backupStateError      = "error"
+	BackupStateInProgress = "in_progress"
+	BackupStateComplete   = "complete"
+	BackupStateError      = "error"
```
c3y1huang, line 25: What is the reason to export these?

weizhe0422: this constants will be used to check the replicas restoring status in engine

https://github.com/longhorn/longhorn-manager/blob/ab225bf88ccf00fb4bb7aff3e8e24245af864dcd/manager/volume.go#L535-L542

c3y1huang: I see. Doesn't `volume.Status.LastBackup != engine.Status.LastRestoredBackup || engine.Spec.RequestedBackupRestore != engine.Status.LastRestoredBackup` indicate the restore is complete?

weizhe0422: Sorry, I think I misunderstood the update logic of the originally used field during the development process; so I will remove the judgment of isRestoring and restore the original judgment method.

manager/volume.go
```diff
@@ -501,6 +499,61 @@ func (m *VolumeManager) Activate(volumeName string, frontend string) (v *longhor
 	return v, nil
 }
 
+func (m *VolumeManager) syncBackupAndRestoreWithLastBackup(volume *longhorn.Volume, engine *longhorn.Engine) (bool, error) {
+	// trigger BackupVolume synchronization
+	var backupVolumeName string
+	if volume.Status.RestoreRequired {
+		name, ok := volume.Labels[types.LonghornLabelBackupVolume]
+		if !ok {
+			return false, errors.New("Cannot find the backup volume label")
```
derekbit, line 508: Add `volume.Name` to the error message and make the first word lowercase "cannot ....".

manager/volume.go
```diff
@@ -501,6 +499,61 @@ func (m *VolumeManager) Activate(volumeName string, frontend string) (v *longhor
...
+		}
+		backupVolumeName = name
+	} else {
+		backupVolumeName = volume.Name
```
derekbit, line 512: When does the volume not have a label?

manager/volume.go
```diff
@@ -501,6 +499,61 @@ func (m *VolumeManager) Activate(volumeName string, frontend string) (v *longhor
...
+		if err != nil {
+			return false, errors.Wrapf(err, "failed to get backup volume: %v", volume.Name)
+		}
+
+		if backupVolume.Status.LastSyncedAt.After(backupVolume.Spec.SyncRequestedAt.Time) ||
+			backupVolume.Status.LastSyncedAt.Equal(backupVolume.Spec.SyncRequestedAt.DeepCopy()) {
+			break
+		}
+		time.Sleep(datastore.VerificationRetryInterval)
+	}
```
derekbit, line 537: Retrying here looks a bit weird.

derekbit (commented):
Sorry. I don't fully understand the context. Some questions before continuing the review
- Can you elaborate more on the `Activate` volume? 
- Is the real-time update necessary?

manager/volume.go
weizhe0422: Got it. I will fix the error to with adding volume name and lowercase the sentence.

manager/volume.go
weizhe0422: I think it should be my bad. 

I originally used the name of the volume as the name of the backup volume when I judged that the restore is not needed, but it is not necessary. I will fix it to take the label value of the volume.

weizhe0422:
> * Can you elaborate more on the `Activate` volume

Hi @derekbit, 
After LH v1.2.0, the latest backup is not immediately forced to sync after activating DR volume. 
[commit](https://github.com/longhorn/longhorn-manager/commit/89250d4e52593c05d466cb5586a8f1e01c03a29a) 
[PR#952](https://github.com/longhorn/longhorn-manager/pull/952)

I think activating the DR volume with the latest backup can reduce the cluster failover time without waiting for the backup poll time.

manager/volume.go
weizhe0422: Because I want to make sure I get the correct latest backup volume name after the backup volume controller reconciles, but don't want to wait a fixed amount of time. Therefore, I use a loop to repeat the attempts to check if the sync is complete.

derekbit: Does it cause any problems if the backup volume is not synchronized in the given period of time?

weizhe0422: If the backup synchronization fails and the retry time is exceeded, the activated disaster recovery volume will not have the latest data like the current version. 

So if syncing the backup volume fails, I will return an error message.

manager/volume.go
```diff
@@ -483,11 +484,8 @@ func (m *VolumeManager) Activate(volumeName string, frontend string) (v *longhor
 		engine = e
 	}
 
-	if v.Status.LastBackup != engine.Status.LastRestoredBackup || engine.Spec.RequestedBackupRestore != engine.Status.LastRestoredBackup {
-		logrus.Infof("Standby volume %v will be activated after finishing restoration, "+
-			"backup volume's latest backup: %v, "+
-			"engine requested backup restore: %v, engine last restored backup: %v",
-			v.Name, v.Status.LastBackup, engine.Spec.RequestedBackupRestore, engine.Status.LastRestoredBackup)
+	if done, err := m.syncBackupAndRestoreWithLastBackup(v, engine); !done {
```
innobead, line 487: `Activate` is used by the REST API, so this PR will make this function become from blocking/sync from non-blocking/async. This could cause some unexpected user experiences. and probably still not working if the request has not been processed completely because it's synchronous.

I would suggest checking the below conditions before attaching the activating DR volume. 
- The backup volume sync is updated
- The restored backup of the engine is updated

[/controller/volume_controller.go#L1637-L1642](https://github.com/longhorn/longhorn-manager/blob/788bb3e7c061baa90e7abc99bdb1f860b5a3efe4/controller/volume_controller.go#L1637-L1642)
```go
		e.Spec.NodeID = v.Status.CurrentNodeID
		e.Spec.ReplicaAddressMap = replicaAddressMap
		e.Spec.DesireState = longhorn.InstanceStateRunning
		// The volume may be activated
		e.Spec.DisableFrontend = v.Status.FrontendDisabled
		e.Spec.Frontend = v.Spec.Frontend
```

The thing you need to do in `Activate` is just update `backupVolume.Spec.SyncRequestedAt` as your current PR to ensure the backup volume will be reconciled to update its latest backup.

innobead (changes requested):
N

manager/volume.go
weizhe0422: I have updated to the following description:
 - `Activate` RESTful API: trigger backup volume to update to get the latest backup name
 - make sure restoring with the latest backup before detach volume when activate DR volume
   - Since the DR volume will be detached after activation, and it will also make the engine stopped. Therefore, I add the check when detach the volume.

manager/volume.go
```diff
@@ -501,6 +497,26 @@ func (m *VolumeManager) Activate(volumeName string, frontend string) (v *longhor
 	return v, nil
 }
 
+func (m *VolumeManager) triggerBackupVolumeToSync(volume *longhorn.Volume) error {
+	backupVolumeName, isExist := volume.Labels[types.LonghornLabelBackupVolume]
+	if !isExist || backupVolumeName == "" {
+		logrus.Debugf("cannot find the backup volume label: %v", volume.Name)
+		return nil
+	}
```
innobead, line 505: Shouldn't return error instead? because `Activate` method is specific for DR volume. 

nit: Also, why the label would be possible missing due to human intervention? then we probably need to improve the update mutation.

innobead (changes requested):
LGTM, but only one part needs to update.

innobead (dismissed):

innobead:
@shuo-wu Please review this.

manager/volume.go
weizhe0422: - Got it, I will return error instead of only log the label missing. 
 - ticket for improving missing labels: https://github.com/longhorn/longhorn/issues/5433

controller/volume_controller.go
```diff
@@ -2916,7 +2916,22 @@ func (vc *VolumeController) checkForAutoDetachment(v *longhorn.Volume, e *longho
...
-	if !(e.Spec.RequestedBackupRestore != "" && e.Spec.RequestedBackupRestore == e.Status.LastRestoredBackup &&
+
+	// make sure the backup volume has been synced at least once
+	if backupVolumeName, isExist := v.Labels[types.LonghornLabelBackupVolume]; isExist && backupVolumeName != "" {
+		backupVolume, err := vc.ds.GetBackupVolumeRO(backupVolumeName)
+		if err != nil {
+			return errors.Wrapf(err, "failed to get backup volume: %v", v.Name)
+		}
+		if !(backupVolume.Status.LastSyncedAt.After(backupVolume.Spec.SyncRequestedAt.Time) ||
+			backupVolume.Status.LastSyncedAt.Equal(backupVolume.Spec.SyncRequestedAt.DeepCopy())) {
```
shuo-wu, line 2927: NIT: How about `backupVolume.Status.LastSyncedAt.Before(backupVolume.Spec.SyncRequestedAt.Time)`?

manager/volume.go
```diff
@@ -471,23 +472,18 @@ func (m *VolumeManager) Activate(volumeName string, frontend string) (v *longhor
...
 	}
-	for _, e := range es {
-		engine = e
-	}
 
-	if v.Status.LastBackup != engine.Status.LastRestoredBackup || engine.Spec.RequestedBackupRestore != engine.Status.LastRestoredBackup {
-		logrus.Infof("Standby volume %v will be activated after finishing restoration, "+
-			"backup volume's latest backup: %v, "+
-			"engine requested backup restore: %v, engine last restored backup: %v",
-			v.Name, v.Status.LastBackup, engine.Spec.RequestedBackupRestore, engine.Status.LastRestoredBackup)
```
shuo-wu, line 490: NIT: Since we don't fetch the engine and check the last restored backup any more,  you can directly remove the whole engine get and check codes here.

shuo-wu (approved):
In general LGTM

controller/volume_controller.go
weizhe0422: Thanks. I think it will make it look intuitive.

controller/volume_controller.go
```diff
@@ -2916,7 +2916,21 @@ func (vc *VolumeController) checkForAutoDetachment(v *longhorn.Volume, e *longho
...
 	}
-	if !(e.Spec.RequestedBackupRestore != "" && e.Spec.RequestedBackupRestore == e.Status.LastRestoredBackup &&
+
+	// make sure the backup volume has been synced at least once
+	if backupVolumeName, isExist := v.Labels[types.LonghornLabelBackupVolume]; isExist && backupVolumeName != "" {
+		backupVolume, err := vc.ds.GetBackupVolumeRO(backupVolumeName)
+		if err != nil {
+			return errors.Wrapf(err, "failed to get backup volume: %v", v.Name)
+		}
+		if backupVolume.Status.LastSyncedAt.Before(backupVolume.Spec.SyncRequestedAt.DeepCopy()) {
```
derekbit, line 2926: Is the `DeepCopy()` necessary?

derekbit: Why do we need to check it be synced at least once? Can you add a comment here?

innobead: yeah, `DeepCopy` is unnecessary.

manager/volume.go
```diff
@@ -501,6 +489,25 @@ func (m *VolumeManager) Activate(volumeName string, frontend string) (v *longhor
 	return v, nil
 }
 
+func (m *VolumeManager) triggerBackupVolumeToSync(volume *longhorn.Volume) error {
+	backupVolumeName, isExist := volume.Labels[types.LonghornLabelBackupVolume]
+	if !isExist || backupVolumeName == "" {
+		return errors.Errorf("cannot find the backup volume label: %v", volume.Name)
```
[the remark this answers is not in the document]
weizhe0422: Thanks, I've updated to fix

controller/volume_controller.go
weizhe0422: Thanks, you are right, there is no need to pass the DeepCopy object. 

Since the parameter type of `v1.time.Before` is `*Time`, I passed the address of `SyncRequestedAt`.

innobead:

PR #1708, opened 2023-02-18
fix(recurring job): add an eventRecorder and a new job type

mantissahz:
Create an event to inform the users if it failed to clean up old snapshots.

~Add a new filed `DeleteSnapshotMode`, default is `create-first`, 
to force the recurring job to delete old snapshots before creating a new snapshot.~

Forcibly do deletion first when the snapshot count exceeds retain count.

Introduce a new recurring job type `snapshot-force-create` to take a new snapshot forcibly when cleaning up old snapshots failed and a new recurring job type `backup-force-create` to to take a new snapshot and do the backup forcibly when cleaning up old snapshots failed.

longhorn/longhorn#4898

Regression test: 
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/3223/

### DIFF
--- a/app/recurring_job.go
+++ b/app/recurring_job.go
@@ -360,6 +360,12 @@ func (job *Job) run() (err error) {
 		job.logger.Infof("Volume %v is in state %v", volumeName, volume.State)
 	}
 
+	if job.task == longhorn.RecurringJobTypeSnapshot || job.task == longhorn.RecurringJobTypeBackup {
+		if err := job.doSnapshotCleanup(false); err != nil {
+			return err
+		}
+	}
+
 	if job.task == longhorn.RecurringJobTypeBackup {
 		job.logger.Infof("Running recurring backup for volume %v", volumeName)
 		return job.doRecurringBackup()
--- a/app/recurring_job.go
+++ b/app/recurring_job.go
@@ -367,7 +367,7 @@ func (job *Job) run() (err error) {
 		}
 	}
 
-	if job.task == longhorn.RecurringJobTypeBackup {
+	if job.task == longhorn.RecurringJobTypeBackup || job.task == longhorn.RecurringJobTypeBackupForceCreate {
 		job.logger.Infof("Running recurring backup for volume %v", volumeName)
 		return job.doRecurringBackup()
 	}
@@ -571,10 +571,6 @@ func (job *Job) doRecurringBackup() (err error) {
 	}
 
 	if err := job.doSnapshot(); err != nil {
-		return err
-	}
-
-	if err := job.doSnapshotCleanup(false); err != nil {
 		return err
 	}
 
--- a/app/recurring_job.go
+++ b/app/recurring_job.go
@@ -16,12 +16,18 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/urfave/cli"
 
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	clientset "k8s.io/client-go/kubernetes"
+	typedv1core "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/record"
 
 	etypes "github.com/longhorn/longhorn-engine/pkg/types"
 
 	longhornclient "github.com/longhorn/longhorn-manager/client"
+	"github.com/longhorn/longhorn-manager/constant"
 	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
 	lhclientset "github.com/longhorn/longhorn-manager/k8s/pkg/client/clientset/versioned"
 	"github.com/longhorn/longhorn-manager/types"
@@ -48,6 +54,8 @@ type Job struct {
 	retain       int
 	task         longhorn.RecurringJobType
 	labels       map[string]string
+
+	eventRecorder record.EventRecorder
 
 	api *longhornclient.RancherClient
 }
@@ -225,6 +233,16 @@ func NewJob(logger logrus.FieldLogger, managerURL, volumeName, snapshotName stri
 		"task":         task,
 	})
 
+	scheme := runtime.NewScheme()
+	if err := longhorn.SchemeBuilder.AddToScheme(scheme); err != nil {
+		return nil, errors.Wrap(err, "unable to create scheme")
+	}
+
+	eventBroadcaster, err := createEventBroadcaster(config)
+	if err != nil {
+		return nil, err
+	}
+
 	return &Job{
 		logger:       logger,
 		lhClient:     lhClient,
@@ -235,7 +253,23 @@ func NewJob(logger logrus.FieldLogger, managerURL, volumeName, snapshotName stri
 		retain:       retain,
 		task:         task,
 		api:          apiClient,
+
+		eventRecorder: eventBroadcaster.NewRecorder(scheme, corev1.EventSource{Component: "longhorn-recurring-job"}),
 	}, nil
+}
+
+func createEventBroadcaster(config *rest.Config) (record.EventBroadcaster, error) {
+	kubeClient, err := clientset.NewForConfig(config)
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to get k8s client")
+	}
+
+	eventBroadcaster := record.NewBroadcaster()
+	eventBroadcaster.StartLogging(logrus.Infof)
+	// TODO: remove the wrapper when every clients have moved to use the clientset.
+	eventBroadcaster.StartRecordingToSink(&typedv1core.EventSinkImpl{Interface: typedv1core.New(kubeClient.CoreV1().RESTClient()).Events("")})
+
+	return eventBroadcaster, nil
 }
 
 // handleVolumeDetachment decides whether the current recurring job should detach the volume.
@@ -376,6 +410,15 @@ func (job *Job) doSnapshot() (err error) {
 }
 
 func (job *Job) doSnapshotCleanup(backupDone bool) (err error) {
+	defer func() {
+		if err != nil {
+			errMessage := errors.Wrapf(err, "failed to clean up old snapshots of volume %v", job.volumeName).Error()
+			if err := job.eventCreate(corev1.EventTypeWarning, constant.EventReasonFailedDeleting, errMessage); err != nil {
+				job.logger.WithError(err).Warn("failed to create an event log")
+			}
+		}
+	}()
+
 	volumeAPI := job.api.Volume
 	volumeName := job.volumeName
 	volume, err := volumeAPI.ById(volumeName)
@@ -428,6 +471,20 @@ func (job *Job) doSnapshotCleanup(backupDone bool) (err error) {
 			time.Sleep(SnapshotPurgeStatusInterval)
 		}
 	}
+	return nil
+}
+
+func (job *Job) eventCreate(eventType, eventReason, message string) error {
+	jobName, found := job.labels[types.RecurringJobLabel]
+	if !found || jobName == "" {
+		return fmt.Errorf("missing RecurringJob label")
+	}
+	recurringJob, err := job.lhClient.LonghornV1beta2().RecurringJobs(job.namespace).Get(context.TODO(), jobName, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+	job.eventRecorder.Eventf(recurringJob, eventType, eventReason, message)
+
 	return nil
 }
 
--- a/app/recurring_job.go
+++ b/app/recurring_job.go
@@ -360,6 +360,7 @@ func (job *Job) run() (err error) {
 		job.logger.Infof("Volume %v is in state %v", volumeName, volume.State)
 	}
 
+	// only recurring job types `snapshot` and `backup` need to check if old snapshots can be deleted or not before creating
 	if job.task == longhorn.RecurringJobTypeSnapshot || job.task == longhorn.RecurringJobTypeBackup {
 		if err := job.doSnapshotCleanup(false); err != nil {
 			return err
@@ -383,7 +384,7 @@ func (job *Job) doRecurringSnapshot() (err error) {
 	}()
 
 	switch job.task {
-	case longhorn.RecurringJobTypeSnapshot:
+	case longhorn.RecurringJobTypeSnapshot, longhorn.RecurringJobTypeSnapshotForceCreate:
 		if err = job.doSnapshot(); err != nil {
 			return err
 		}
@@ -530,7 +531,7 @@ func (job *Job) filterExpiredSnapshotsOfCurrentRecurringJob(snapshots []longhorn
 	// Only consider deleting the snapshots that were created by our current job
 	snapshots = filterSnapshotsWithLabel(snapshots, types.RecurringJobLabel, jobLabel)
 
-	if job.task == longhorn.RecurringJobTypeSnapshot {
+	if job.task == longhorn.RecurringJobTypeSnapshot || job.task == longhorn.RecurringJobTypeSnapshotForceCreate {
 		return filterExpiredItems(snapshotsToNameWithTimestamps(snapshots), job.retain)
 	}
 
--- a/k8s/pkg/apis/longhorn/v1beta2/recurringjob.go
+++ b/k8s/pkg/apis/longhorn/v1beta2/recurringjob.go
@@ -2,7 +2,7 @@ package v1beta2
 
 import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
-// +kubebuilder:validation:Enum=snapshot;snapshot-force-create;snapshot-cleanup;snapshot-delete;backup
+// +kubebuilder:validation:Enum=snapshot;snapshot-force-create;snapshot-cleanup;snapshot-delete;backup;backup-force-create
 type RecurringJobType string
 
 const (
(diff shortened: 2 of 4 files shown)


types/setting.go
```diff
@@ -930,6 +933,16 @@ var (
 		Default:  "1",
 	}
 
+	SettingDefinitionRecurringJobCreateSnapshotWhenRebuilding = SettingDefinition{
+		DisplayName: "Recurring Job Create Snapshot When Rebuilding",
+		Description: "This setting enables a recurring job to create a snapshot when replica rebuilding forcily .\n\n",
```
derekbit, line 938: Add the description for
- The potential risk if users enable the option.
- When the snapshot count reaches the maximum retain number and the option is not enabled, what will occur?

app/recurring_job.go
```diff
@@ -359,6 +393,16 @@ func (job *Job) run() (err error) {
...
 
+	if len(volume.RebuildStatus) != 0 {
+		isForceCreateSnapshot, err := getIsForceCreateSnapshot(job)
+		if err != nil {
+			return err
+		}
+		if !isForceCreateSnapshot {
+			return fmt.Errorf("recurring job will not start when replica is rebuilding")
+		}
+	}
```
derekbit, line 404: Should we always reject it when the option is enabled, or should we consider the number of snapshots and the retaintion number?

types/setting.go
mantissahz: Modified. And changed the setting to be `SettingDefinitionRecurringJobSnapshotDeleteFirst` and add the description for potential risk if this option is `false`.

app/recurring_job.go
mantissahz: Removed this part, we will do deleting first or do the recurring job like before.
Thank you.

app/recurring_job.go
```diff
@@ -375,7 +422,48 @@ func (job *Job) doSnapshot() (err error) {
...
+func (job *Job) deleteSnapshotRetainReachedOrExceeded(volume *longhornclient.Volume, isSnapshotDeleteFirst bool) error {
+	volumeAPI := job.api.Volume
+
+	collection, err := volumeAPI.ActionSnapshotList(volume)
+	if err != nil {
+		return err
+	}
+
+	jobSnapshots := filterSnapshotsWithLabel(collection.Data, types.RecurringJobLabel, job.labels[types.RecurringJobLabel])
+	if len(jobSnapshots) >= job.retain {
```
PhanLe1010, line 450: Should this one be strictly bigger:  len(jobSnapshots) > job.retain so that we don't have case in which number snapshot < expected retaining count after the deletion?

PhanLe1010 (commented):
In gerneral, the PR LGTM. We will need to refactor the logic to operate in longhorn snapshot CRD instead of calling engine API directly after the AD PR is merged.

PhanLe1010:
Actually, will we have a problem for the backup recurring job with retain count 1?

The current logic is that when a backup finish, we only keep the snapshot of the last backup. Then next time when we check deleteSnapshotRetainReachedOrExceeded we deleted the only snapshot. Then the backup will be force to be a full backup which is bad

mantissahz:
> Actually, will we have a problem for the backup recurring job with retain count 1?
The current logic is that when a backup finish, we only keep the snapshot of the last backup. Then next time when we check deleteSnapshotRetainReachedOrExceeded we deleted the only snapshot. Then the backup will be force to be a full backup which is bad

@PhanLe1010
IIUC snapshot won't be deleted for backup type before taking a new snapshot
because this filter [filterExpiredSnapshotsOfCurrentRecurringJob](https://github.com/longhorn/longhorn-manager/pull/1708/files#diff-61e6464dd7934db9eb985d4107eb81d205c39647995e386a44401980ffbfa175L470-L482)
But retain number of snapshot type which is 1 might cause this scenario, I will check this again.
Thank you.

app/recurring_job.go
mantissahz: As `MaxRecurringJobRetain = 50` in AD controller?

PhanLe1010: No, I mean if we delete the snapshot when len(jobSnapshots) == job.retain and later on if we fail to create a new snapshot, the number of snapshot will be less than expected, right? 

If that cycle repeats, we will have less and less snapshot. So, I am wondering if the condition should be len(jobSnapshots) > job.retain

PhanLe1010: Doin this would fix the issue https://github.com/longhorn/longhorn-manager/pull/1708#issuecomment-1441260982 too

mantissahz: In `listSnapshotNamesToCleanup` of `doSnapshotCleanup` to [filterExpiredItems(snapshotsToNameWithTimestamps(snapshots), job.retain)](https://github.com/longhorn/longhorn-manager/pull/1708/files#diff-61e6464dd7934db9eb985d4107eb81d205c39647995e386a44401980ffbfa175R574-R579)
IIUC the [filterExpiredItems](https://github.com/longhorn/longhorn-manager/blob/master/app/recurring_job.go#L791-L801) only list the number of old snapshots that exceeds the `job.retain` or `job.retain-1` in this PR.

mantissahz:
@shuo-wu 
Do you mean this part [upgrade path](https://github.com/longhorn/longhorn-manager/pull/1708/commits/588b2ba5eacd587b28ec57804c2a59aa91fc10e7#diff-54e77ae1d9149f7ea519019af64a71e75f8029573fbb71ee209c944508a11502R50-R69) and [mutator](https://github.com/longhorn/longhorn-manager/pull/1708/commits/588b2ba5eacd587b28ec57804c2a59aa91fc10e7#diff-54e77ae1d9149f7ea519019af64a71e75f8029573fbb71ee209c944508a11502R50-R69)?

I would do the commit separation at next commit.

shuo-wu:
I mean, if the field is empty (all existing recurring job CRs should be in this case), should we set `DeleteFirst` by default? Before this PR, we already apply `CreateFirst`, right? This is actually a default behavior change.

mantissahz:
>I mean, if the field is empty (all existing recurring job CRs should be in this case), should we set DeleteFirst by default? Before this PR, we already apply CreateFirst, right? This is actually a default behavior change.

Got it. Yes, it is should be `create-first` for all existing recurring job CRs.
Modified. Thanks.

app/recurring_job.go
```diff
@@ -375,7 +425,32 @@ func (job *Job) doSnapshot() (err error) {
...
+func (job *Job) deleteSnapshotRetainReachedOrExceeded(volume *longhornclient.Volume) error {
+	volumeAPI := job.api.Volume
+
+	collection, err := volumeAPI.ActionSnapshotList(volume)
+	if err != nil {
+		return err
+	}
+
+	jobSnapshots := filterSnapshotsWithLabel(collection.Data, types.RecurringJobLabel, job.labels[types.RecurringJobLabel])
+	if len(jobSnapshots) >= job.retain && job.retain > MinRecurringJobRetainNumber {
```
shuo-wu, line 437: IIRC, retain counting won't include the volume head. Then the min retain number should be 0 (snapshot prune will be triggered).

shuo-wu (commented):
There will be more kinds of recurring types and some of them may be unrelated to snapshot creation/deletion: https://github.com/longhorn/longhorn/pull/5520
Should we add one more enum for this field?

mantissahz:
> I still feel we should have a new RecurringJobType (ex: snapshot-delete-first) instead of introducing a new DeleteSnapshotMode field. This new field is less generic and looks a little confusing with other recurring job types.

As discussion, we might have some ways to handle old snapshots:
1. Delete `snapshot`/`backup` type jobs and use `snapshot-delete` type job to do clean-up when snapshot count exceeds the retain number for some reasons. And probably we only need to create an event when cleaning up old snapshots failed
1. Introduce a new field as a workaround to allow users to make deleting old snapshots first before taking a new snapshot possible.
1. Or we just check the number of old snapshots before taking a new snapshot:
  delete old snapshots first if the number of old snapshots exceeds the retain number forcibly
  (because it is an abnormal situation)

IMO, I might take 3.
WDYT or any other opinions? @shuo-wu @c3y1huang @innobead @PhanLe1010 @derekbit

app/recurring_job.go
mantissahz: Setting `MinRecurringJobRetainNumber` to 1 prevents that all snapshots are gone from creating a new snapshot failed as Phan mentioned here https://github.com/longhorn/longhorn-manager/pull/1708#issuecomment-1441107980

app/recurring_job.go
```diff
@@ -375,7 +413,34 @@ func (job *Job) doSnapshot() (err error) {
 	return nil
 }
 
+func (job *Job) deleteSnapIfRetainNumExceeded(volume *longhornclient.Volume) error {
+	if job.task == longhorn.RecurringJobTypeSnapshotForceCreate {
```
c3y1huang, line 417: Suggest moving this to the caller to reduce this method's complexity.

app/recurring_job.go
```diff
@@ -431,6 +496,22 @@ func (job *Job) doSnapshotCleanup(backupDone bool) (err error) {
...
+	}
+	recurringJob, err := job.lhClient.LonghornV1beta2().RecurringJobs(job.namespace).Get(context.TODO(), jobName, metav1.GetOptions{})
+	if err != nil {
+		job.logger.WithError(err).Errorf("Failed to get recurring job %v.", jobName)
+
+		return err
+	}
+	job.eventRecorder.Eventf(recurringJob, eventType, eventReason, message)
+
+	return err
```
c3y1huang, line 512:
```suggestion
	return nil
```

k8s/pkg/apis/longhorn/v1beta2/recurringjob.go
```diff
@@ -2,14 +2,15 @@ package v1beta2
...
+// +kubebuilder:validation:Enum=snapshot;snapshot-force-create;snapshot-cleanup;snapshot-delete;backup
 type RecurringJobType string
 
 const (
-	RecurringJobTypeSnapshot        = RecurringJobType("snapshot")         // periodically create snapshots
-	RecurringJobTypeSnapshotCleanup = RecurringJobType("snapshot-cleanup") // periodically purge removable snapshots and system snapshots
-	RecurringJobTypeSnapshotDelete  = RecurringJobType("snapshot-delete")  // periodically remove and purge all kinds of snapshots that exceed the retention count
-	RecurringJobTypeBackup          = RecurringJobType("backup")           // periodically create snapshots then do backups
+	RecurringJobTypeSnapshot            = RecurringJobType("snapshot")              // periodically create snapshots
+	RecurringJobTypeSnapshotForceCreate = RecurringJobType("snapshot-force-create") // periodically create snapshots forcibly
```
c3y1huang, line 10:
```suggestion
	RecurringJobTypeSnapshotForceCreate = RecurringJobType("snapshot-force-create") // periodically create snapshots even if old snapshots cleanup failed
```

c3y1huang (changes requested):

app/recurring_job.go
mantissahz: Moved the condition check to the caller. Thanks

k8s/pkg/apis/longhorn/v1beta2/recurringjob.go
mantissahz: Modified. Thanks

app/recurring_job.go
```diff
@@ -409,6 +413,23 @@ func (job *Job) doSnapshot() (err error) {
...
 
+func (job *Job) deleteSnapIfRetainNumExceeded(volume *longhornclient.Volume) error {
+	volumeAPI := job.api.Volume
+
+	collection, err := volumeAPI.ActionSnapshotList(volume)
+	if err != nil {
+		return err
+	}
+	jobSnapshots := filterSnapshotsWithLabel(collection.Data, types.RecurringJobLabel, job.labels[types.RecurringJobLabel])
+	if len(jobSnapshots) > job.retain {
```
shuo-wu, line 425: This seems to be included in `doSnapshotCleanup`. Can we directly and blindly call `job.doSnapshotCleanup(false)`?

k8s/pkg/apis/longhorn/v1beta2/recurringjob.go
```diff
@@ -2,14 +2,15 @@ package v1beta2
...
-// +kubebuilder:validation:Enum=snapshot;snapshot-cleanup;snapshot-delete;backup
+// +kubebuilder:validation:Enum=snapshot;snapshot-force-create;snapshot-cleanup;snapshot-delete;backup
 type RecurringJobType string
 
 const (
-	RecurringJobTypeSnapshot        = RecurringJobType("snapshot")         // periodically create snapshots
-	RecurringJobTypeSnapshotCleanup = RecurringJobType("snapshot-cleanup") // periodically purge removable snapshots and system snapshots
-	RecurringJobTypeSnapshotDelete  = RecurringJobType("snapshot-delete")  // periodically remove and purge all kinds of snapshots that exceed the retention count
-	RecurringJobTypeBackup          = RecurringJobType("backup")           // periodically create snapshots then do backups
+	RecurringJobTypeSnapshot            = RecurringJobType("snapshot")              // periodically create snapshots
```
shuo-wu, line 9: Add one more statement explaining the new behavior

shuo-wu (commented):
We may need to add tips in UI that explain the snapshot types.

app/recurring_job.go
```diff
@@ -397,6 +397,10 @@ func (job *Job) doSnapshot() (err error) {
 		return errors.Wrapf(err, "could not get volume %v", volumeName)
 	}
 
+	if err := job.deleteSnapIfRetainNumExceeded(volume); err != nil {
```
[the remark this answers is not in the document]
mantissahz: Moved to the `run` before doing any type of the recurring job.

app/recurring_job.go
mantissahz: Yes, I think it could do that.
Removed the function `deleteSnapIfRetainNumExceeded` and used `job.doSnapshotCleanup(false)` instead.
Thanks.

k8s/pkg/apis/longhorn/v1beta2/recurringjob.go
mantissahz: Added. Thanks

app/recurring_job.go
```diff
@@ -360,6 +360,12 @@ func (job *Job) run() (err error) {
 		job.logger.Infof("Volume %v is in state %v", volumeName, volume.State)
 	}
 
+	if job.task == longhorn.RecurringJobTypeSnapshot || job.task == longhorn.RecurringJobTypeBackup {
```
shuo-wu, line 363: NIT: Add a simple comment explaining why

shuo-wu (dismissed):
In general LGTM

app/recurring_job.go
innobead: @mantissahz It seems we can get this to 1.3.3 if this can get merged.

mantissahz: @shuo-wu Added the simple comment to explain why only these two types.

@innobead yeah, let's try to do it.

shuo-wu (approved):

app/recurring_job.go
```diff
@@ -376,6 +410,13 @@ func (job *Job) doSnapshot() (err error) {
 }
 
 func (job *Job) doSnapshotCleanup(backupDone bool) (err error) {
+	defer func() {
+		if err != nil {
+			errMessage := errors.Wrapf(err, "failed to clean up old snapshots of volume %v", job.volumeName).Error()
+			job.eventCreate(corev1.EventTypeWarning, constant.EventReasonFailedDeleting, errMessage)
```
innobead, line 423: Need to handle the returned error from `eventCreate` but it will be intended to ignore with warning log only.

app/recurring_job.go
```diff
@@ -431,6 +472,22 @@ func (job *Job) doSnapshotCleanup(backupDone bool) (err error) {
...
 }
 
+func (job *Job) eventCreate(eventType, eventReason, message string) error {
+	jobName, found := job.labels[types.RecurringJobLabel]
+	if !found || jobName == "" {
+		return fmt.Errorf("missing RecurringJob label")
+	}
+	recurringJob, err := job.lhClient.LonghornV1beta2().RecurringJobs(job.namespace).Get(context.TODO(), jobName, metav1.GetOptions{})
+	if err != nil {
+		job.logger.WithError(err).Errorf("Failed to get recurring job %v.", jobName)
```
innobead, line 489: No need to log here, but rather let the caller to deal with the log if needed.

app/recurring_job.go
```diff
@@ -326,6 +360,13 @@ func (job *Job) run() (err error) {
 		job.logger.Infof("Volume %v is in state %v", volumeName, volume.State)
 	}
 
+	// only recurring job types `snapshot` and `backup` need to check if old snapshots can be deleted or not before creating
+	if job.task == longhorn.RecurringJobTypeSnapshot || job.task == longhorn.RecurringJobTypeBackup {
```
innobead, line 364: So the `recurring back` will be behavior change as well, and it will only take effect if there is no error found during the cleanup before doing the backup? 

My question is do we need `RecurringJobTypeBackupForceCreate`?

innobead (changes requested):
In general, LGTM. Need to update longhorn/longhorn manifests as well due to CRD update.

app/recurring_job.go
mantissahz: Add `Warn` log if the error is not `nil` from `job.eventCreate`.
Thanks

app/recurring_job.go
mantissahz: Removed. Thanks

app/recurring_job.go
mantissahz: Added another new recurring job type `backup-force-create`.
Thanks.

app/recurring_job.go
```diff
@@ -574,10 +574,6 @@ func (job *Job) doRecurringBackup() (err error) {
 		return err
 	}
 
-	if err := job.doSnapshotCleanup(false); err != nil {
-		return err
-	}
```
shuo-wu, line 515: Do you think we can remove this? After the latest snapshot creation, the volume would contain 2 snapshots if it's not the first time creating snapshots.

shuo-wu (commented):
NIT: The last 2 git commits have an identical title.

mantissahz:
> NIT: The last 2 git commits have an identical title.
@shuo-wu Modified the commit title. Thanks.

shuo-wu (approved):

innobead:

Please add reference backport issue links to the following backport PRs created by @Mergifyio.

PR #1709, opened 2023-02-20
Force disable revision counter for a strict local volume

derekbit:
Longhorn/longhorn#5257


innobead (changes requested):
Should we take care of the upgrade?

derekbit:
> Should we take care of the upgrade?

It's nice to have. Will update the ticket.

shuo-wu (approved):

innobead:
@derekbit I think it should be fine to take care of the upgrade path. This only takes effect for new volumes should be OK.

derekbit:
> @derekbit I think it should be fine to take care of the upgrade path. This only takes effect for new volumes should be OK, but need to have a doc if no upgrade path for the migration.

Forgot updating this issue today.
I prefer to handle the upgrade path as well. Will check if there is any potential issue and update the PR.

innobead (approved):

PR #1712, opened 2023-02-21
fix: instance manager cpu reservation

c3y1huang:
https://github.com/longhorn/longhorn/issues/5419


c3y1huang:
Test result: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/3009/

innobead (approved):

PR #1719, opened 2023-02-24
feat(disk): add a global setting for StorageReserved of DiskSpec

mantissahz:
Add a global setting 'storage-reserved-percentage` to reserved disk storage for better user experience when adding a newly created longhorn node's disk.

Default value is 30 for `storage-reserved-percentage`.

Modify `StorageReserved` of `DiskSpec` in `CreateDefaultDisk` by the global setting `storageReservedPercentage`.

longhorn/longhorn#4773


types/setting.go
```diff
@@ -462,6 +465,16 @@ var (
 		Default:     "25",
 	}
 
+	SettingDefinitionStorageReservedPercentage = SettingDefinition{
+		DisplayName: "Storage Reserved Percentage",
+		Description: "The Reserved percentage defines how much storage will be reserved not to be used relative to the hard drive's capacity and it affects only a newly created longhorn node's disk.",
```
derekbit, line 470:
```suggestion
		Description: "The reserved percentage determines what percentage of disk space will not be used on a newly created Longhorn node.",
```

types/setting.go
```diff
@@ -1172,6 +1185,8 @@ func ValidateSetting(name, value string) (err error) {
 		if value < 0 {
 			return fmt.Errorf("value %v should be positive", value)
 		}
+	case SettingNameStorageReservedPercentage:
```
derekbit, line 1188: Error out if the value is negative.

types/setting.go

types/setting.go
mantissahz: `fallthrough` to `SettingNameStorageMinimalAvailablePercentage`
It will return error if value is negative, does this what you point out?

derekbit: Ah, yeah. Thank you.

derekbit (dismissed):
LGTM

shuo-wu:
I am considering if we really need this option. Actually, this option cannot prevent the reserved space from being consumed by users or other applications since we cannot block IO. This allows users to customize **per node** schedulable space but sometimes it confuses users even developers. Having the corresponding global setting may be a little redundant since we already have a global setting `Storage Over Provisioning Percentage` that helps control the schedulable space.
WDYT?

mantissahz:
@shuo-wu IMO it's convenient for users to changed the default disk storage reserved value for a newly added node otherwise users will need to change it manually after node is added.
Yeah, `Storage Over Provisioning Percentage` could that helps control the schedulable space but it's also not intuitive to users because we use `StorageReserved` and `OverProvisioningPercentage` for scheduling replicas, right?
@derekbit @innobead WDYT?

shuo-wu:
> it's convenient for users to changed the default disk storage reserved value for a newly added node otherwise users will need to change it manually after node is added.

With the existing per-node config, we can achieve it with this new global setting.

shuo-wu (dismissed):
So far it won’t make any feature more complicated. Let's merge it first

types/setting.go
```diff
@@ -462,6 +465,16 @@ var (
 		Default:     "25",
 	}
 
+	SettingDefinitionStorageReservedPercentage = SettingDefinition{
+		DisplayName: "Storage Reserved Percentage",
+		Description: "The reserved percentage determines what percentage of disk space will not be used on a newly created Longhorn node.",
```
innobead, line 470: Suggest `The reserved percentage specifies the percentage of disk space that will not be allocated to the default disk on each new Longhorn node.`

innobead (changes requested):
In general, LGTM.

What Shuo said is correct, because users actually can use storage over-provisioning percentage to indirectly adjust the available space. However, here it is more for setting the default storage reserved for the newly default disk, so the meaning is a little different because it will not take any effect after the disk is provisioned already.

types/setting.go
innobead: Also, can we change the setting from `storage-reserved-percentage` to `storage-reserved-percentage-for-default-disk` instead?

types/setting.go
```diff
@@ -50,6 +50,7 @@ const (
 	SettingNameReplicaAutoBalance                                       = SettingName("replica-auto-balance")
 	SettingNameStorageOverProvisioningPercentage                        = SettingName("storage-over-provisioning-percentage")
 	SettingNameStorageMinimalAvailablePercentage                        = SettingName("storage-minimal-available-percentage")
+	SettingNameStorageReservedPercentage                                = SettingName("storage-reserved-percentage-for-default-disk")
```
innobead, line 53: Please change all variables accordingly instead of updating the option.

types/setting.go
```diff
@@ -462,6 +465,16 @@ var (
 		Default:     "25",
 	}
 
+	SettingDefinitionStorageReservedPercentage = SettingDefinition{
+		DisplayName: "Storage Reserved Percentage",
```
innobead, line 469: As I said, change the name accordingly, so please review the whole PR..

innobead (changes requested):

types/setting.go
mantissahz: Updated. Thanks.

types/setting.go
mantissahz: Updated all. Thanks. Only func `CreateDefaultDisk` keeps the parameter name `storageReservedPercentage`

innobead (approved):

PR #1725, opened 2023-03-01
feat: add settings to get supported k8s versions

weizhe0422:
[Longhorn 4813](https://github.com/longhorn/longhorn/issues/4813)

Just add a new parameter: `SupportedK8sVersions` to the `ExtraInfo` field of `response.json` to respond without modifying the `Upgrade Responder`. The `SupportedK8sVersions` field should be filled with K8S information supported by the current version and the latest version of Longhorn.
image

self-test:
 - Update and add new latest version: v1.4.1
```json
{
    "Versions": [
        {
            "Name": "v1.1.3",
            "ReleaseDate": "2021-12-17T00:00:00Z",
            "Tags": [
                "stable"
            ],
            "ExtraInfo": {
                "SupportedK8sVersions": ">=v1.16.0-0 <v1.22.0-0"
            }
        },
        {
            "Name": "v1.2.4",
            "ReleaseDate": "2022-03-17T00:00:00Z",
            "Tags": [
                "stable"
            ],
            "ExtraInfo": {
                "SupportedK8sVersions": ">=1.18.0-0"
            }
        },
        {
            "Name": "v1.4.0",
            "ReleaseDate": "2022-12-30T00:00:00Z",
            "Tags": [
                "stable"
            ],
            "ExtraInfo": {
                "SupportedK8sVersions": ">=1.21.0-0"
            }
        },
        {
            "Name": "v1.4.1",
            "ReleaseDate": "2023-03-01T00:00:00Z",
            "Tags": [
                "latest"
            ],
            "ExtraInfo": {
                "SupportedK8sVersions": ">=1.21.0-0"
            }
        },
        {
            "Name": "v1.3.2",
            "ReleaseDate": "2022-10-14T00:00:00Z",
            "Tags": [
                "stable"
            ],
            "ExtraInfo": {
                    "SupportedK8sVersions": ">=1.18.0-0 <1.25.0-0"
            }
        }
    ]
}
```
image

### DIFF
--- a/controller/setting_controller.go
+++ b/controller/setting_controller.go
@@ -12,6 +12,8 @@ import (
 	"strings"
 	"time"
 
+	"golang.org/x/mod/semver"
+
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 
@@ -81,6 +83,7 @@ type Version struct {
 	Name        string // must be in semantic versioning
 	ReleaseDate string
 	Tags        []string
+	ExtraInfo   map[string]string
 }
 
 type CheckUpgradeRequest struct {
@@ -825,6 +828,14 @@ func (sc *SettingController) syncUpgradeChecker() error {
 	if err != nil {
 		return err
 	}
+	k8sLatestLonghornVersion, err := sc.ds.GetSetting(types.SettingNameLatestLonghornSupportedK8sVersions)
+	if err != nil {
+		return err
+	}
+	k8sStableLonghornVersions, err := sc.ds.GetSetting(types.SettingNameStableLonghornSupportedK8sVersions)
+	if err != nil {
+		return err
+	}
 
 	if !upgradeCheckerEnabled {
 		if latestLonghornVersion.Value != "" {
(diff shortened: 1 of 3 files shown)


innobead (changes requested):
I am quite confused.

Why let `CheckLatestAndStableLonghornVersions` leftover and not used, then create some functions looking quite confusing?

Suggest just using `CheckLatestAndStableLonghornVersions` as usual, then you can get the Kubernetes version range support from the extra info based on the latest and stable longhorn version, then return them together with longhorn versions from CheckLatestAndStableLonghornVersions.

Something like the below, but you can change the method with named returned variables for the better doc.

```go
	latestVersion := ""
	stableVersions := []string{}
	k8sLatestVersions := ""
	k8sStableVersions := ""
	for _, v := range resp.Versions {
		for _, tag := range v.Tags {
			if tag == VersionTagLatest {
				latestVersion = v.Name
				if version, ok := v.ExtraInfo["SupportedK8sVersions"]; ok {
					k8sLatestVersions = version
				}
			}
			if tag == VersionTagStable {
				stableVersions = append(stableVersions, v.Name)
				if version, ok := v.ExtraInfo["SupportedK8sVersions"]; ok {
					k8sStableVersions = version
				}
			}
		}
	}
	if latestVersion == "" {
		return "", "", "", "", fmt.Errorf("cannot find latest Longhorn version during CheckLatestAndStableLonghornVersions")
	}
	sort.Strings(stableVersions)

	return latestVersion, strings.Join(stableVersions, ","), k8sLatestVersions, k8sStableVersions, nil

```

innobead:
@weizhe0422 What you should do when updating this PR.

- Have a K8s support version range for the current longhorn version. 
   - It only needs to update when it's empty, and it will be cleaned up when doing the longhorn upgrade because the current longhorn version will be updated.
- Have a K8s support version range for the latest Longhorn version.
- Have a K8s support version range for each stable longhorn version. 
   - At the upgrade responder, we will only keep the latest stable version, for example, only have 1.3.2 for 1.3 w/o previous versions.

For parsing these versions, you should be able to do in the current `CheckLatestAndStableLonghornVersions`. KISS if this is just a minor improvement.

weizhe0422:
Hi @innobead, I've re-updated this PR with reference suggestions. 

Among them, I pulled out the part of updating the setting value and put it in another function. I refer to `CodeFactor` inspection to avoid making a single function too complicated. Thank you.

innobead:
@weizhe0422 let's close this PR first since I feel we need more thoughts on UX.

PR #1728, opened 2023-03-03
fix(system-restore): backup target overwritten

c3y1huang:
https://github.com/longhorn/longhorn/issues/5442

### DIFF
--- a/controller/system_rollout_controller.go
+++ b/controller/system_rollout_controller.go
@@ -1717,6 +1717,8 @@ func (c *SystemRolloutController) restoreServiceAccounts() (err error) {
 var systemRolloutIgnoredSettings = [...]string{
 	string(types.SettingNameConcurrentBackupRestorePerNodeLimit),
 	string(types.SettingNameConcurrentReplicaRebuildPerNodeLimit),
+	string(types.SettingNameBackupTarget),
+	string(types.SettingNameBackupTargetCredentialSecret),
 }
 
 func isSystemRolloutIgnoredSetting(name string) bool {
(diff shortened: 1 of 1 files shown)


c3y1huang:
Result:
- https://ci.longhorn.io/job/private/job/longhorn-tests-regression/3062/
- https://ci.longhorn.io/job/private/job/longhorn-tests-regression/3063/
- https://ci.longhorn.io/job/private/job/longhorn-tests-regression/3064/

controller/system_rollout_controller.go
```diff
@@ -951,6 +951,11 @@ func (c *SystemRolloutController) restoreConfigMaps() (err error) {
 	for _, restore := range c.configMapList.Items {
 		log := c.logger.WithField(types.KubernetesKindConfigMap, restore.Name)
 
+		if restore.Name == types.DefaultDefaultSettingConfigMapName {
+			log.Infof(SystemRolloutMsgIgnoreItemFmt, types.DefaultDefaultSettingConfigMapName)
+			continue
+		}
```
innobead, line 957: So basically, the default config will not be taken care during system restored? because we will just rely on setting restoring which will exclude some settings as below potentially impact the restore process, right?

If yes, we probably need to describe this behavior also mentioned what settings will not be restored in the doc. https://longhorn.io/docs/1.4.0/advanced-resources/system-backup-restore/restore-longhorn-system/#prerequisite

```go
var systemRolloutIgnoredSettings = [...]string{
	string(types.SettingNameConcurrentBackupRestorePerNodeLimit),
	string(types.SettingNameConcurrentReplicaRebuildPerNodeLimit),
	string(types.SettingNameBackupTarget),
	string(types.SettingNameBackupTargetCredentialSecret),
}

```

innobead (approved):
LGTM

innobead:
@Mergifyio backport v1.4.x

PR #1729, opened 2023-03-03
Disallow backup deletion when it is used for restoring a volume

derekbit:
Longhorn/longhorn#5458

### DIFF
--- a/controller/backup_controller.go
+++ b/controller/backup_controller.go
@@ -207,6 +207,19 @@ func getLoggerForBackup(logger logrus.FieldLogger, backup *longhorn.Backup) *log
 	)
 }
 
+func (bc *BackupController) isBackupBeingUsedForVolumeRestore(backupName, backupVolumeName string) error {
+	volumes, err := bc.ds.ListVolumesByBackupVolumeRO(backupVolumeName)
+	if err != nil {
+		return err
+	}
+	for _, v := range volumes {
+		if v.Status.RestoreRequired {
+			return fmt.Errorf("backup %v cannot be deleted due to the ongoing volume %v restoration", backupName, v.Name)
+		}
+	}
+	return nil
+}
+
 func (bc *BackupController) reconcile(backupName string) (err error) {
 	// Get Backup CR
 	backup, err := bc.ds.GetBackup(backupName)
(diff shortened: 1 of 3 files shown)


controller/backup_controller.go
```diff
@@ -207,6 +207,19 @@ func getLoggerForBackup(logger logrus.FieldLogger, backup *longhorn.Backup) *log
 	)
 }
 
+func (bc *BackupController) isBackupBeingUsedForVolumeRestore(backupName, backupVolumeName string) error {
```
innobead, line 210: Suggest `func (bc *BackupController) isBackupBeingUsedForVolumeRestore(backupName, backupVolumeName string) (bool, error)
` instead, because this is used to determine if the backup is used or not.

controller/backup_controller.go
```diff
@@ -278,6 +291,11 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
 				return nil // Ignore error to prevent enqueue
 			}
 
+			if err := bc.isBackupBeingUsedForVolumeRestore(backup.Name, backupVolumeName); err != nil {
+				log.WithError(err).Warnf("Unable to delete remote backup %v", backup.Name)
+				return err
+			}
```
innobead, line 297: Should we return err to requeue? just noticed the queue for backup is always resync in a periodical time by default. I think we should retry, but I am also curious why we have periodical resync for the queue. 

```go
	ds.BackupInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
		AddFunc:    bc.enqueueBackup,
		UpdateFunc: func(old, cur interface{}) { bc.enqueueBackup(cur) },
		DeleteFunc: bc.enqueueBackup,
	})
```

innobead (changes requested):
In general, LGTM.

controller/backup_controller.go
derekbit: Addressed. Thank you.

controller/backup_controller.go
derekbit: The resync period is 30 seconds. We can rely on the timer instead of the retries by error.

innobead: So do we still need to return error? I saw the code is the same.

innobead (approved):

innobead:
@Mergifyio backport v1.4.x

controller/backup_controller.go
derekbit: Ah..I forgot push the latest code...

innobead: Ah thanks. I also missed the check 😅

derekbit:
@Mergifyio backport v1.3.x

PR #1733, opened 2023-03-03
Retry backup deletion based on the resync mechnism

derekbit:
Longhorn/longhorn#5458

### DIFF
--- a/controller/backup_controller.go
+++ b/controller/backup_controller.go
@@ -207,17 +207,17 @@ func getLoggerForBackup(logger logrus.FieldLogger, backup *longhorn.Backup) *log
 	)
 }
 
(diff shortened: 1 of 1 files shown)


derekbit:
cc @innobead

shuo-wu (commented):
NIT: Is this necessary? There is still an error when the returned boolean is true

innobead:
It can be tricky since, even if the function returns true, it still includes a message to explain what's happening. Therefore, the message is wrapped as an error. To avoid any confusion, could rename the function to below.

```go
func (bc *BackupController) isBackupNotBeingUsedForVolumeRestore(backupName, backupVolumeName string) (bool, error)
```

PR #1737, opened 2023-03-06
Fail replica if volume is not attached and node is down

derekbit:
Whenever the engine isn't attached, if the node goes down, the replica can be marked as failed. In the attached mode, we can determine whether the replica can fail by relying on the data plane's connectivity status.

Longhorn/longhorn#5464
Longhorn/longhorn#5477


innobead (approved):

innobead (approved):

innobead:

PR #1748, opened 2023-03-06
feat: introduce failed-and-cleanup state to backingimage to handle data source tmp files

ChanYiLin:
[longhorn/longhorn#3682](https://github.com/longhorn/longhorn/issues/3682)

Depends on: backing-image-manager [PR#101](https://github.com/longhorn/backing-image-manager/pull/101)

Implementation
- Add remove data source tmp file in `Delete api` in backing-image-manager
- Introduce a new state for BackingImageDataSource: `FailedAndCleanUp`

For Case: data source Pod is deleted in the progress
- invoke `Delete` when BackingImageManager Controller finds that
    - BackingImageDataSource CurrentState is failed and Transferred is false ([ref](https://github.com/longhorn/longhorn-manager/blob/master/controller/backing_image_manager_controller.go#L678))
    - then change the BackingImageDataSource state to `FailedAndCleanUp`

For Case: BackingImage CR is deleted in the progress
In BackingImageController
- before removing the finalizer
    - if BackingImageDataSource CR is in-progress status and not transferred, update BackingImageDataSource status to `Failed`, this will trigger the above behavior to clean up the tmp files
- can remove the finalizer when BackingImageDataSource's status is changed to `FailedAndCleanUp` by BackingImageManager Controller

### DIFF
--- a/controller/backing_image_controller.go
+++ b/controller/backing_image_controller.go
@@ -212,6 +212,14 @@ func (bic *BackingImageController) syncBackingImage(key string) (err error) {
 			log.Info("Need to wait for all replicas stopping using this backing image before removing the finalizer")
 			return nil
 		}
+		log.Info("Try to clean up backing image data source tmp file before cleaning up backing image manager")
+		cleanUpTmpFileFinished, err := bic.cleanupBackingImageDataSourceTmpFile(backingImage)
+		if err != nil {
+			return err
+		}
+		if !cleanUpTmpFileFinished {
+			return nil
+		}
 		log.Info("No replica is using this backing image, will clean up the record for backing image managers and remove the finalizer then")
 		if err := bic.cleanupBackingImageManagers(backingImage); err != nil {
 			return err
@@ -277,6 +285,32 @@ func (bic *BackingImageController) syncBackingImage(key string) (err error) {
 	}
 
 	return nil
+}
+
+func (bic *BackingImageController) cleanupBackingImageDataSourceTmpFile(bi *longhorn.BackingImage) (finished bool, err error) {
+	bids, err := bic.ds.GetBackingImageDataSource(bi.Name)
+	if err != nil && !apierrors.IsNotFound(err) {
+		return false, err
+	}
+
+	// if bids is transferred or it is not in progress, there is no tmp file left on the host.
+	fileProcessingStarted :=
+		bids.Status.CurrentState == longhorn.BackingImageStateStarting ||
+			bids.Status.CurrentState == longhorn.BackingImageStateInProgress
+
+	if bids.Spec.FileTransferred || !fileProcessingStarted ||
+		bids.Status.CurrentState == longhorn.BackingImageStateFailedAndCleanUp ||
+		bids.Status.CurrentState == longhorn.BackingImageStateUnknown {
+		return true, nil
+	}
+
+	// mark the status to failed so manager can clean up the orphan tmp file
+	bids.Status.Message = "backing image is deleted, try to cleaning up the tmp file of data source"
+	bids.Status.CurrentState = longhorn.BackingImageStateFailed
+	if _, err = bic.ds.UpdateBackingImageDataSourceStatus(bids); err != nil {
+		return false, err
+	}
+	return false, nil
 }
 
 func (bic *BackingImageController) cleanupBackingImageManagers(bi *longhorn.BackingImage) (err error) {
--- a/controller/backing_image_controller.go
+++ b/controller/backing_image_controller.go
@@ -212,12 +212,12 @@ func (bic *BackingImageController) syncBackingImage(key string) (err error) {
 			log.Info("Need to wait for all replicas stopping using this backing image before removing the finalizer")
 			return nil
 		}
-		log.Info("Try to clean up backing image data source tmp file before cleaning up backing image manager")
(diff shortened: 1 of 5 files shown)


ChanYiLin:
cleanup

Delete the data source pod during uploading
- the status became failed-and-cleanup in the end (datasource CR status and UI)
- the tmp file is removed (folder list)
- `Delete API` is triggered in the the manager pod (as in the log)

ChanYiLin:
delete_finalizer

deleted the backingimage CR during uploading
- the lhbids CR status became failed by [controllerL#308](https://github.com/longhorn/longhorn-manager/pull/1748/files#diff-9cdffb3a94c7a2350738c352077504be988843d61387bd93d86ef4643d194a47R308), with message: `"backing image is deleted, try to cleaning up the tmp file of data source"`

delete
- `Delete API` is triggered in manager pod to cleanup tmp file
- And the backingImage CR is deleted

The tmp file is delete on the host

controller/backing_image_controller.go
```diff
@@ -279,6 +287,32 @@ func (bic *BackingImageController) syncBackingImage(key string) (err error) {
 	return nil
 }
 
+func (bic *BackingImageController) cleanupBackingImageDataSourceTmpFile(bi *longhorn.BackingImage) (finished bool, err error) {
```
innobead, line 290: Suggest renaming to `IsBackingImageDataSourceCleanup`.

controller/backing_image_controller.go
```diff
@@ -212,6 +212,14 @@ func (bic *BackingImageController) syncBackingImage(key string) (err error) {
 			log.Info("Need to wait for all replicas stopping using this backing image before removing the finalizer")
 			return nil
 		}
+		log.Info("Try to clean up backing image data source tmp file before cleaning up backing image manager")
+		cleanUpTmpFileFinished, err := bic.cleanupBackingImageDataSourceTmpFile(backingImage)
+		if err != nil {
+			return err
+		}
+		if !cleanUpTmpFileFinished {
+			return nil
```
innobead, line 221: This is intended, so suggest having an info log before returning nil, for examples, `Requesting to clean up Backing Image Data Source first before removing from Backing Image Manager`, you can add helpful parameters in the log.

innobead (changes requested):
In general, LGTM. Some comments needed to resovle first.

innobead (commented):
Also, this is related to CRD update, so need to update longhorn/longhorn manifests. Please consult with Derke about how to do that.

ChanYiLin:
> Also, this is related to CRD update, so need to update longhorn/longhorn manifests. Please consult with @derekbit about how to do that.

Discussed with @derekbit, since there is no crd field changed only introduce a new state, we don't need to update the crd.yaml in chart.

innobead:
> > Also, this is related to CRD update, so need to update longhorn/longhorn manifests. Please consult with @derekbit about how to do that.
> 
> Discussed with @derekbit, since there is no crd field changed only introduce a new state, we don't need to update the crd.yaml in chart.

Sounds correct. 

[/k8s/crds.yaml#L22-L26](https://github.com/longhorn/longhorn-manager/blob/b81881360c3496a8bb9e7dd05437e2805e81bb92/k8s/crds.yaml#L22-L26)
```yaml
  - additionalPrinterColumns:
    - description: The current state of the pod used to provision the backing image file from source
      jsonPath: .status.currentState
      name: State
      type: string
```

controller/backing_image_controller.go
```diff
@@ -279,6 +287,32 @@ func (bic *BackingImageController) syncBackingImage(key string) (err error) {
...
+		return true, nil
+	}
+
+	// mark the status to failed so manager can clean up the orphan tmp file
+	bids.Status.Message = "backing image is deleted, try to cleaning up the tmp file of data source"
+	bids.Status.CurrentState = longhorn.BackingImageStateFailed
+	if _, err = bic.ds.UpdateBackingImageDataSourceStatus(bids); err != nil {
+		return false, err
+	}
+	return false, nil
```
innobead, line 313: It's unusual to return false w/o error if the function/method support returning bool and error together. However, here it's quite special usage, because in fact there are two conditions here instead one, it would cause some confusion initially.

- backing image has been cleaned up 
- backing image was not cleaned up but has been requested to clean up 
  - but having error while updating the data source status
  - updated the status successfully

Ideally there should be three returned variables to make things clean. (`cleanUp`, `requested`, `error`). However, it's fine for this PR.

innobead (approved):

innobead:

controller/backing_image_controller.go
innobead: I think you misunderstood my comment. It's more about general coding patterns, when returning a false variable, it should have an error variable to explain why it's failed. This was why I said should be better to clarify the returned variables if could. 

Doing asynchronous fix is totally correct here.

ChanYiLin: yeah, I get your point now,

Maybe a better way is to remove the return err variable to make this function only return true or false,
if there is any error, we just log error and return false to wait next reconcile loop

Thanks for the advise

innobead: Yeah this is another way but as a dependent function/method it's better to return error to let callers decide what to do.

Again, this is minor stuff but a basic guide to share with you 🙂

PR #1749, opened 2023-03-06
Add log message for failing replica when the node is down and volume is at non-attached state

derekbit:
Longhorn/longhorn#5464
Longhorn/longhorn#5477

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -1589,6 +1589,7 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
 				// In the attached mode, we can determine whether the replica can fail by relying on the data plane's
 				// connectivity status.
 				if v.Status.State != longhorn.VolumeStateAttached {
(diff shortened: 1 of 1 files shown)


controller/volume_controller.go
```diff
@@ -1589,6 +1589,8 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
 				// In the attached mode, we can determine whether the replica can fail by relying on the data plane's
 				// connectivity status.
 				if v.Status.State != longhorn.VolumeStateAttached {
+					log.WithField("replica", r.Name).Warnf("Replica %v is marked as failed since the volume is in %v state and instance manager is unable to launch a replica", r.Name, v.Name)
+					vc.eventRecorder.Eventf(v, v1.EventTypeNormal, constant.EventReasonAttached, "", v.Name, v.Status.CurrentNodeID)
```
innobead, line 1593: Why the message is empty in the event? Probably can use the same log as the message.

Can we change the log to `Replica %v is marked as failed since the volume is not attached because the instance manager is unable to launch the replica`.

innobead (changes requested):

innobead (approved):

innobead:

PR #1757, opened 2023-03-09
Try the best to delete RWX volume's engine instance

derekbit:
Longhorn/longhorn#5507

### DIFF
--- a/controller/engine_controller.go
+++ b/controller/engine_controller.go
@@ -501,15 +501,38 @@ func (ec *EngineController) DeleteInstance(obj interface{}) error {
 		}
 	}
 
-	if im.Status.CurrentState != longhorn.InstanceManagerStateRunning {
-		return nil
+	v, err := ec.ds.GetVolumeRO(e.Spec.VolumeName)
+	if err != nil {
+		if !apierrors.IsNotFound(err) {
+			return err
+		}
 	}
+
+	isRWXVolume := false
+	if v != nil && v.Spec.AccessMode == longhorn.AccessModeReadWriteMany && !v.Spec.Migratable {
+		isRWXVolume = true
+	}
+
+	// For a RWX volume, the node down, for example, caused by kubelet restart, leads to share-manager pod deletion/recreation
+	// and volume detachment/attachment.
+	// Then, the newly created share-manager pod blindly mounts the longhorn volume inside /dev/longhorn/<pvc-name> and exports it.
+	// To avoid mounting a dead and orphaned volume, try to clean up the engine process as well as the orphaned iscsi device
+	// regardless of the instance-manager status.
+	if !isRWXVolume {
+		if im.Status.CurrentState != longhorn.InstanceManagerStateRunning {
+			logrus.Infof("Skip deleting engine %v since instance manager is in %v state", e.Name, im.Status.CurrentState)
+			return nil
+		}
+	}
+
+	logrus.Infof("Deleting engine instance %v", e.Name)
 
 	// For the engine process in instance manager v0.7.0, we need to use the cmdline to delete the process
 	// and stop the iscsi
 	if im.Status.APIVersion == engineapi.IncompatibleInstanceManagerAPIVersion {
 		url := imutil.GetURL(im.Status.IP, engineapi.InstanceManagerDefaultPort)
 		args := []string{"--url", url, "engine", "delete", "--name", e.Name}
+
 		if _, err := util.ExecuteWithoutTimeout([]string{}, engineapi.GetDeprecatedInstanceManagerBinary(e.Status.CurrentImage), args...); err != nil && !types.ErrorIsNotFound(err) {
 			return err
 		}
--- a/controller/system_backup_controller.go
+++ b/controller/system_backup_controller.go
@@ -46,7 +46,7 @@ const (
 
 	SystemBackupTempDir = "/tmp"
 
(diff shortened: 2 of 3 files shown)


controller/engine_controller.go
```diff
@@ -501,15 +501,42 @@ func (ec *EngineController) DeleteInstance(obj interface{}) error {
...
+	logrus.Infof("Deleting engine %v for volume %v", e.Name, v.Name)
+
+	defer func() {
+		logrus.WithError(err).Warnf("Failed to delete engine %v for volume %v", e.Name, v.Name)
+		if isRWXVolume && im.Status.CurrentState != longhorn.InstanceManagerStateRunning {
+			// Try the best to delete engine and ignore the error when volume is
+			// a RWX volume and the instance manager is not running
+			err = nil
+		}
+	}()
```
innobead, line 532: Why we need this? I assume we only need to allow the engine process is able to delete for the volume used by the share manager pod. 

If somehow it can't be deleted, this controller should periodically reconcile this to retry the deletion. It seems no need to ignore the returned error, but still respect any error happening here. WDYT?

BTW, `err` is a local variable, so if need to deal with this way, probably should use named returned variable?

innobead (changes requested):

controller/engine_controller.go
```diff
@@ -501,15 +501,42 @@ func (ec *EngineController) DeleteInstance(obj interface{}) error {
...
-		return nil
+	v, err := ec.ds.GetVolumeRO(e.Spec.VolumeName)
+	if err != nil {
+		if !apierrors.IsNotFound(err) {
+			return err
+		}
+	}
+
+	isRWXVolume := false
+	if v != nil && v.Spec.AccessMode == longhorn.AccessModeReadWriteMany {
```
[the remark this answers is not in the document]
derekbit: If the volume is somehow gone but the engine still exists, we still need the check. Or, it won't happen?

controller/engine_controller.go
derekbit: Yeah, I think reconciling the deletion is better.

controller/engine_controller.go
```diff
@@ -501,15 +501,33 @@ func (ec *EngineController) DeleteInstance(obj interface{}) error {
...
+		isRWXVolume = true
 	}
 
+	if !isRWXVolume {
```
shuo-wu, line 516: NIT: Add a comment for this check

controller/engine_controller.go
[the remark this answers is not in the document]
innobead: Good point! We should because it's not needed to enforce to delete.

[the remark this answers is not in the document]
derekbit: Addressed. Thank you.

innobead (approved):

innobead:

controller/engine_controller.go
```diff
@@ -501,15 +501,38 @@ func (ec *EngineController) DeleteInstance(obj interface{}) error {
...
+	// To avoid mounting a dead and orphaned volume, try to clean up the engine process as well as the orphaned iscsi device
+	// regardless of the instance-manager status.
+	if !isRWXVolume {
+		if im.Status.CurrentState != longhorn.InstanceManagerStateRunning {
+			logrus.Infof("Skip deleting engine %v since instance manager is in %v state", e.Name, im.Status.CurrentState)
+			return nil
+		}
+	}
+
+	logrus.Infof("Deleting engine instance %v", e.Name)
```
derekbit, line 528: One community user intentionally shut down a node where one RWX volume's engine instance is running on.
The RWX volume is stuck at detaching, because the engine instance is not able to be deleted.

Should we add a maximum tries for the stuck detaching?
cc @shuo-wu @innobead

shuo-wu: IIRC, users can detach a node attached to a down node. Why can't this engine be stopped?

derekbit: Previous, if the IM is not in running state, the `deleteInstance` is skipped directly. But in this case, we delete the engine instance for the RWX volume. If failed, retry again and again. So, the engine cannot be stopped.

shuo-wu: Got it. See the updated comment.

derekbit: > can the SM pod somehow understand if a volume is valid before mounting

No, the check logics are in share manager pod. It only see a `device`.
The `stat` can be executed and return correct information.  Then, fail to check the filesystem type (this is expected for a new volume) and do `mkfs`. It will encounter IO error when running `mkfs`.

shuo-wu: I see. Do you prefer to add a retry/backoff for this? Typically it requires a new field in engine CR... Here we may not be able to use an in-memory field.

derekbit: Do you mean retry `deleteInstance`?

derekbit: How about doing once? For kubelet restart, the underlay network and the IM pod are available. The deletion should succeed. If failed, the node can be regarded as a real shutdown.

shuo-wu: > Should we add a maximum tries for the stuck detaching?

Sorry... I was thinking you want to add a retry...

How about the network disconnection case? Will there be any leftovers/risks after the network recovery?

shuo-wu: Sounds good. This function (or the upper caller) will try its best to delete the process during cleanup but it cannot guarantee.

Is the fix PR already created?

PR #1760, opened 2023-03-10
fix(validator): misleading error

c3y1huang:
The returning error is misleading to think it is triggered by the system restore. Should return the original error.


shuo-wu (approved):

c3y1huang:
@Mergifyio backport v1.4.x

PR #1763, opened 2023-03-10
feat: enforce the upgrade path

mantissahz:
Only allow upgrading from previous minor version. ex. v1.4.x to v1.5.0

Now it returns error only when new apps/pods are created by kubernetes for upgrading and old pods will be killed.
User could only apply the old manifest to roll back the longhorn system.

longhorn/longhorn#5131

### DIFF
--- a/app/pre_uppgrade.go
+++ b/app/pre_uppgrade.go
@@ -1,0 +1,58 @@
+package app
+
+import (
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	"github.com/urfave/cli"
+	"k8s.io/client-go/tools/clientcmd"
+
+	lhclientset "github.com/longhorn/longhorn-manager/k8s/pkg/client/clientset/versioned"
+	"github.com/longhorn/longhorn-manager/types"
+	upgradeutil "github.com/longhorn/longhorn-manager/upgrade/util"
+)
+
+func PreUpgradeCmd() cli.Command {
+	return cli.Command{
+		Name: "pre-upgrade",
+		Flags: []cli.Flag{
+			cli.StringFlag{
+				Name:  FlagKubeConfig,
+				Usage: "Specify path to kube config (optional)",
+			},
+			cli.StringFlag{
+				Name:     FlagNamespace,
+				EnvVar:   types.EnvPodNamespace,
+				Required: true,
+				Usage:    "Specify Longhorn namespace",
+			},
+		},
+		Action: func(c *cli.Context) {
+			logrus.Infof("Running pre-upgrade...")
+			defer logrus.Infof("Completed pre-upgrade.")
+
+			if err := preUpgrade(c); err != nil {
+				logrus.WithError(err).Fatalf("Failed to run pre-upgrade")
+			}
+		},
+	}
+}
+
+func preUpgrade(c *cli.Context) error {
+	namespace := c.String(FlagNamespace)
+
+	config, err := clientcmd.BuildConfigFromFlags("", c.String(FlagKubeConfig))
+	if err != nil {
+		return errors.Wrap(err, "unable to get client config")
+	}
+
+	lhClient, err := lhclientset.NewForConfig(config)
+	if err != nil {
+		return errors.Wrap(err, "unable to get clientset")
+	}
+
+	if err := upgradeutil.CheckUpgradePathSupported(namespace, lhClient); err != nil {
+		return err
+	}
+
+	return nil
+}
--- a/upgrade/upgrade.go
+++ b/upgrade/upgrade.go
@@ -76,7 +76,7 @@ func Upgrade(kubeconfigPath, currentNodeID string) error {
 		return errors.Wrap(err, "unable to create scheme")
 	}
 
-	if err := upgradeLocalNode(); err != nil {
+	if err := upgradeutil.CheckUpgradePathSupported(namespace, lhClient); err != nil {
 		return err
 	}
 
@@ -216,16 +216,6 @@ func doAPIVersionUpgrade(namespace string, config *restclient.Config, lhClient *
 		return fmt.Errorf("don't support upgrade from %v to %v", crdAPIVersion, types.CurrentCRDAPIVersion)
 	}
 
-	return nil
-}
-
-func upgradeLocalNode() (err error) {
-	defer func() {
-		err = errors.Wrap(err, "upgrade local node failed")
-	}()
-	if err := v070to080.UpgradeLocalNode(); err != nil {
-		return err
-	}
 	return nil
 }
 
--- a/upgrade/util/util.go
+++ b/upgrade/util/util.go
@@ -5,7 +5,11 @@ import (
 	"fmt"
 	"math"
 	"reflect"
+	"strconv"
+	"strings"
 	"sync"
+
+	"golang.org/x/mod/semver"
 
 	"github.com/jinzhu/copier"
 	"github.com/pkg/errors"
@@ -22,6 +26,12 @@ import (
 	lhclientset "github.com/longhorn/longhorn-manager/k8s/pkg/client/clientset/versioned"
 	"github.com/longhorn/longhorn-manager/meta"
 	"github.com/longhorn/longhorn-manager/types"
+)
+
+const (
+	// LonghornV1ToV2MinorVersionNum v1 minimal minor version when the upgrade path is from v1.x to v2.0
+	// TODO: decide the v1 minimum version that could be upgraded to v2.0
+	LonghornV1ToV2MinorVersionNum = 30
 )
 
 type ProgressMonitor struct {
@@ -112,7 +122,7 @@ func MergeStringMaps(baseMap, overwriteMap map[string]string) map[string]string 
 	return result
 }
 
-func GetCurrentLonghornVersion(namespace string, lhClient *lhclientset.Clientset) (string, error) {
+func GetCurrentLonghornVersion(namespace string, lhClient lhclientset.Interface) (string, error) {
 	currentLHVersionSetting, err := lhClient.LonghornV1beta2().Settings(namespace).Get(context.TODO(), string(types.SettingNameCurrentLonghornVersion), metav1.GetOptions{})
 	if err != nil {
 		if apierrors.IsNotFound(err) {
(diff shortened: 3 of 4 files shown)


upgrade/util/util.go
```diff
@@ -149,6 +160,77 @@ func CreateOrUpdateLonghornVersionSetting(namespace string, lhClient *lhclientse
...
+	}
+
+	if (lhNewMinorVersionNum - lhMinorVersionBeforeUpgradeNum) == 1 {
+		return true
+	}
+
+	if semver.Compare(lhVersionBeforeUpgrade, meta.Version) > 0 {
+		logrus.Warnf("Downgrading is not supported, %v to %v", lhVersionBeforeUpgrade, meta.Version)
+		return false
+	}
```
shuo-wu, line 217: Is it better to check the patch version then rather than directly comparing the whole version number? Will this error out when the version is from `v1.4.0-rc1` to `v.1.4.0-rc2`?

mantissahz: Upgrading from `v1.4.0-rc1` to `v.1.4.0-rc2` will be allowed, but `v1.4.0-rc2` to `v.1.4.0-rc1` will not.
`semver.Compare` can check the `prerelease` version.
Do we allow downgrade for pre-release version?

upgrade/upgrade.go
```diff
@@ -76,7 +76,23 @@ func Upgrade(kubeconfigPath, currentNodeID string) error {
 		return errors.Wrap(err, "unable to create scheme")
 	}
 
-	if err := upgradeLocalNode(); err != nil {
+	// If the current Longhorn is already the latest version,
+	// the leader election & the whole upgrade path could be skipped.
+	lhVersionBeforeUpgrade, err := upgradeutil.GetCurrentLonghornVersion(namespace, lhClient)
```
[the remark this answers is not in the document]
mantissahz: For the new install, `lhVersionBeforeUpgrade` will get empty string and following check statement will return `true` if `lhVersionBeforeUpgrade` is an empty string

upgrade/util/util.go
```diff
@@ -149,6 +160,77 @@ func CreateOrUpdateLonghornVersionSetting(namespace string, lhClient *lhclientse
...
+		logrus.WithError(err).Warnf("failed to parse %v minor version", lhVersionBeforeUpgrade)
+		return false
+	}
+
+	if semver.Compare(lhMajorVersionBeforeUpgrade, lhNewMajorVersion) > 0 {
+		logrus.Warnf("Downgrading is not supported, %v to %v", lhVersionBeforeUpgrade, meta.Version)
+		return false
+	}
+
+	if semver.Compare(lhMajorVersionBeforeUpgrade, lhNewMajorVersion) < 0 {
```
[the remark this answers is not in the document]
mantissahz: Added the check statement. Thanks

app/recovery_backend.go
```diff
@@ -63,6 +65,14 @@ func runRecoveryBackendServer(c *cli.Context) error {
...
 	}
 
+	lhClient, err := lhclientset.NewForConfig(cfg)
+	if err != nil {
+		return errors.Wrap(err, "unable to get clientset")
+	}
+	if !upgradeutil.IsUpgradePathSupported(namespace, lhClient) {
+		return fmt.Errorf("unable to upgrade since the current upgrade path is not supported")
+	}
+
```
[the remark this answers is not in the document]
mantissahz: Removed for handling in longhorn-manager pod. Thanks

app/webhook.go
```diff
@@ -68,6 +70,14 @@ func runWebhookServer(c *cli.Context, webhookType string) error {
 		return errors.Wrap(err, "unable to get client config")
 	}
 
+	lhClient, err := lhclientset.NewForConfig(cfg)
+	if err != nil {
+		return errors.Wrap(err, "unable to get clientset")
+	}
+	if !upgradeutil.IsUpgradePathSupported(namespace, lhClient) {
+		return fmt.Errorf("unable to upgrade since the current upgrade path is not supported")
+	}
```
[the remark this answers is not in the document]
mantissahz: Removed for handling in longhorn-manager pod. Thanks

upgrade/upgrade.go
```diff
@@ -76,7 +76,23 @@ func Upgrade(kubeconfigPath, currentNodeID string) error {
...
+	if err != nil {
+		return err
+	}
+
+	if semver.IsValid(meta.Version) && semver.Compare(lhVersionBeforeUpgrade, meta.Version) >= 0 {
+		logrus.Info("Skip the upgrade since the current Longhorn system is already up to date")
+		return nil
+	}
```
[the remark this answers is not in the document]
mantissahz: Removed. Thanks

upgrade/util/util.go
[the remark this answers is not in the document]
shuo-wu: There is overlapping between this check and `(lhNewMajorVersionNum - lhMajorVersionBeforeUpgradeNum) > 1`

upgrade/util/util.go
shuo-wu: I am fine with either... If we don't need this kind of pre-release version check, we can check the patch version only instead, right?

upgrade/util/util.go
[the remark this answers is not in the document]
mantissahz: Moved the check into `if semver.Compare(lhMajorVersionBeforeUpgrade, lhNewMajorVersion) < 0 {`

upgrade/util/util.go
mantissahz: Yes, I think we could do that instead.

shuo-wu (dismissed):

upgrade/upgrade.go
```diff
@@ -219,12 +224,14 @@ func doAPIVersionUpgrade(namespace string, config *restclient.Config, lhClient *
 	return nil
 }
 
-func upgradeLocalNode() (err error) {
+func upgradeLocalNode(lhVersionBeforeUpgrade string) (err error) {
```
innobead, line 227: Do we still need this? We already check if upgrade path supported or not before. So from 1.5, the qualified version will be 1.4.x or 1.5.0. 

If yes, probably we can remove this old function.

upgrade/util/util.go
```diff
@@ -149,6 +159,97 @@ func CreateOrUpdateLonghornVersionSetting(namespace string, lhClient *lhclientse
 	return nil
 }
 
+// IsUpgradePathSupported returns if the upgrade path from lhVersionBeforeUpgrade to meta.Version is supported.
+func IsUpgradePathSupported(namespace string, lhClient *lhclientset.Clientset) bool {
+	lhVersionBeforeUpgrade, err := GetCurrentLonghornVersion(namespace, lhClient)
```
innobead, line 164: Can we add a unit test for this? so we can test the upgrade path combination. 

As I mentioned in another comment, make this function return error with message instead. 

Also, this is an exported function, so besides adding the brief intro, I would suggest briefly explaining what paths will be supported and validated.

upgrade/util/util.go
```diff
@@ -149,6 +159,97 @@ func CreateOrUpdateLonghornVersionSetting(namespace string, lhClient *lhclientse
...
+	}
+
+	if semver.Compare(lhMajorVersionBeforeUpgrade, lhNewMajorVersion) < 0 {
+		if (lhNewMajorVersionNum - lhMajorVersionBeforeUpgradeNum) > 1 {
+			return false
+		}
+		return lhMinorVersionBeforeUpgradeNum >= LonghornV1ToV2MinorVersionNum
+	}
+
+	if (lhNewMinorVersionNum-lhMinorVersionBeforeUpgradeNum) > 1 || (lhNewMinorVersionNum-lhMinorVersionBeforeUpgradeNum) < 0 {
```
innobead, line 208: `(lhNewMinorVersionNum-lhMinorVersionBeforeUpgradeNum) < 0 ` this is a downgrade case for minor version downgrade but the major is the same.

innobead (changes requested):
In general, LGTM.

However, some places need to modify. Please resolve all comments. Thanks.

upgrade/upgrade.go
```diff
@@ -76,28 +76,33 @@ func Upgrade(kubeconfigPath, currentNodeID string) error {
...
 	}
 
-	if err := upgradeLocalNode(); err != nil {
+	if !upgradeutil.IsUpgradePathSupported(namespace, lhClient) {
+		return fmt.Errorf("unable to upgrade since the current upgrade path is not supported")
+	}
+
+	// If the current Longhorn is already the latest version,
+	// the leader election & the whole upgrade path could be skipped.
+	lhVersionBeforeUpgrade, err := upgradeutil.GetCurrentLonghornVersion(namespace, lhClient)
```
innobead, line 85: For the version variable, you can make it simple as well like `lhCurrentVresion` and `lnUpgradeVersion`.

app/pre_uppgrade.go
```diff
@@ -0,0 +1,61 @@
...
+	return cli.Command{
+		Name: "pre-upgrade",
+		Flags: []cli.Flag{
+			cli.StringFlag{
+				Name:  FlagKubeConfig,
+				Usage: "Specify path to kube config (optional)",
+			},
+			cli.StringFlag{
+				Name:   FlagNamespace,
+				EnvVar: types.EnvPodNamespace,
```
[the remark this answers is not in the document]
mantissahz: Added. Thanks.

app/pre_uppgrade.go
```diff
@@ -0,0 +1,61 @@
...
+			},
+		},
+		Action: func(c *cli.Context) {
+			logrus.Infof("Running pre-upgrade...")
+			defer logrus.Infof("Completed pre-upgrade.")
+
+			if err := preUpgrade(c); err != nil {
+				logrus.Fatalf("Error during pre-upgrade: %v", err)
```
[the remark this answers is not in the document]
mantissahz: Modified. Thanks.

app/pre_uppgrade.go
```diff
@@ -0,0 +1,61 @@
...
+			}
+		},
+	}
+}
+
+func preUpgrade(c *cli.Context) error {
+	namespace := c.String(FlagNamespace)
+	if namespace == "" {
+		return errors.New("namespace is required")
+	}
```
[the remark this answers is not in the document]
mantissahz: Add `Required` for FlagNamespace StringFlag, and removed the check statement. Thanks.

app/pre_uppgrade.go
```diff
@@ -0,0 +1,61 @@
...
+		return errors.Wrap(err, "unable to get client config")
+	}
+
+	lhClient, err := lhclientset.NewForConfig(config)
+	if err != nil {
+		return errors.Wrap(err, "unable to get clientset")
+	}
+
+	if !upgradeutil.IsUpgradePathSupported(namespace, lhClient) {
+		return fmt.Errorf("unable to upgrade since the current upgrade path is not supported")
```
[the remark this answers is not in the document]
mantissahz: Modified function name to CheckUpgradePathSupport, and function returned `error`

upgrade/upgrade.go
mantissahz: Removed this old function.

upgrade/util/util.go
mantissahz: Added a unit test, and added the brief introduction for the function.
Thanks.

upgrade/util/util.go
mantissahz: Removed this duplicated check statement. Thanks.

upgrade/upgrade.go
mantissahz: Modified names for these variables. Thanks

upgrade/util/util.go
```diff
@@ -149,6 +159,99 @@ func CreateOrUpdateLonghornVersionSetting(namespace string, lhClient *lhclientse
 	return nil
 }
 
+// CheckUpgradePathSupported returns if the upgrade path from lhCurrentVersion to meta.Version is supported.
+//
+//	For example: upgrade path is from x.y.z to a.b.c,
+//	0 <= a-x <= 1 is supported, and y should be after a specific version if a-x == 1
+//	0 <= b-y <= 1 is supported when a-x == 0
+//	all downgrade is not supported
+func CheckUpgradePathSupported(namespace string, lhClient lhclientset.Interface) error {
```
innobead, line 168: Please add more one path is not allowed. Any pre-GA version upgrade is not allowed. For example, `1.4.2-rc1` to `1.5.0`.  You should be able to use semver function to achieve the comparison check. 

error log should highlight any non-GA version is not supported.

upgrade/util/util_test.go
```diff
@@ -46,3 +63,108 @@ func TestProgressMonitorInc(t *testing.T) {
 		t.Fatalf(`targetValue = %v, expectedTargetValue = %v`, targetValue, expectedTargetValue)
 	}
 }
+
+func TestCheckUpgradePathSupported(t *testing.T) {
```
innobead, line 67: This is good.

innobead (changes requested):
In general, LGTM.

Just two left comments to resolve.

upgrade/util/util.go
mantissahz: Added the check statement for non-GA version across one minor version and error log.
Do we support non-GA version upgrade e.g., from v1.5.0-rc1 to v1.5.0-rc2 and across patch versions e.g. from v1.5.0-rc3 to v1.5.0 or v1.5.1?
I assume we support these scenarios.

upgrade/util/util_test.go
mantissahz: Added more test cases for any pre-GA version upgrade is not allowed.

upgrade/util/util_test.go
```diff
@@ -46,3 +63,156 @@ func TestProgressMonitorInc(t *testing.T) {
...
+			name:           "upgradable minor version from non-GA version to non-GA version",
+			currentVersion: "v1.4.1-rc1",
+			upgradeVersion: "v1.5.0-rc2",
+			expectError:    true,
+		},
+		{
+			name:           "upgradable minor version to non-GA version",
+			currentVersion: "v1.4.2",
+			upgradeVersion: "v1.5.0-rc2",
+			expectError:    false,
```
innobead, line 144: This should be error. Can't upgrade to any non-GA version.

innobead (changes requested):

upgrade/util/util.go
innobead: Let's remove the non-GA version check first because it would impact our internal testing for the upgrade. 

I will create another ticket to tackle this in 1.6.

mantissahz: Removed the non-GA version check.

upgrade/util/util_test.go
mantissahz: Removed the non-GA version test cases.

upgrade/util/util_test.go
mantissahz: Removed the test case.

upgrade/util/util_test.go
```diff
@@ -46,3 +63,114 @@ func TestProgressMonitorInc(t *testing.T) {
...
+			name:           "new install",
+			currentVersion: "",
+			upgradeVersion: "v1.5.0",
+			expectError:    false,
+		},
+		{
+			name:           "invalid upgrade version",
+			currentVersion: "v1.5.0",
+			upgradeVersion: "v1.5.x",
+			expectError:    true,
```
innobead, line 84: This should be supported. Or, what example data you should use is `1.5.0` -> `1.5.1`. You can also add one more `1.5.1` -> `1.5.2`

1.5.0/1.4.x can be upgraded to 1.5.0

innobead (changes requested):

upgrade/util/util_test.go
mantissahz: Modified the misleading version string from `v1.5.x` to `v1.5.error`.
And added test cases from `v1.5.0` to `v1.5.1` and `v1.5.2` to `v1.5.3`

innobead: As discussed, 1.5.x is used for documentation so it's not a valid case. We don't need to test it.

mantissahz: Removed invalid version test case, and thanks for explaining.

innobead (approved):
LGTM

PR #1765, opened 2023-03-13
Make sure pod.Spec.NodeName is assigned before check the node status

derekbit:
Previously, the immediate IsNodeDownOrDeleted check could fail if the share manager pod's spec.nodeName was not set, which could lead to the detachment of the volume and the transfer of ownership.
A race condition between the deletion of old volume replicas and the attachment of new volume could lead to a faulty volumes.

Longhorn/longhorn#5537

### DIFF
--- a/controller/share_manager_controller.go
+++ b/controller/share_manager_controller.go
@@ -597,6 +597,11 @@ func (c *ShareManagerController) syncShareManagerPod(sm *longhorn.ShareManager) 
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


innobead:
@derekbit Can elaborate this a bit?

> A race condition between the deletion of old volume replicas and the attachment of new volume could lead to a faulty volumes.

derekbit:
This is not valid. Working on another PR.

PR #1772, opened 2023-03-16
feat: marked replica failed if the node is deleted

weizhe0422:
[Longhorn 5542](https://github.com/longhorn/longhorn/issues/5542)

In `ReconcileVolumeState()`, check if the node where the replica is located is deleted, and if so, directly mark the replica as failed.

Below are the test results:
 - Having a volume with one replica, then delete the VM that volume attached
   - Both the Volume and replica become `Faulted`
     volume_faulted
     replica_failedAt
     UI

 - Having a volume with two replicas
   - delete the VM that volume attached
     - The volume becomes `unknown` and the copy becomes `failed`. The volume is then reattached to another schedulable node and the failed replica is removed, and the volume returns to a `healthy` state.
       
       
   - delete the VM that volume doesn't attached
     - The volume becomes `Degraded` and the replica becomes `Failed`. After delete the failed replica, the volume back to `Healthy` state. 
     replica_failedAt_2

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -1585,11 +1585,19 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
 					continue
 				}
 
-				// Whenever the engine isn't attached, if the node goes down, the replica can be marked as failed.
-				// In the attached mode, we can determine whether the replica can fail by relying on the data plane's
-				// connectivity status.
-				if v.Status.State != longhorn.VolumeStateAttached {
-					log.WithField("replica", r.Name).Warnf("Replica %v is marked as failed since the volume %v is not attached because the instance manager is unable to launch the replica", r.Name, v.Name)
+				// If the engine isn't attached or the node goes down, the replica can be marked as failed.
+				// In the attached mode, we can determine whether the replica can fail by relying on the data plane's connectivity status.
+				nodeDeleted, err := vc.ds.IsNodeDeleted(r.Spec.NodeID)
+				if err != nil {
+					return err
+				}
+
+				if v.Status.State != longhorn.VolumeStateAttached || nodeDeleted {
+					msg := fmt.Sprintf("Replica %v is marked as failed since the volume %v is not attached because the instance manager is unable to launch the replica", r.Name, v.Name)
+					if nodeDeleted {
+						msg = fmt.Sprintf("Replica %v is marked as failed since the node %v is deleted.", r.Name, r.Spec.NodeID)
+					}
+					log.WithField("replica", r.Name).Warn(msg)
 					if r.Spec.FailedAt == "" {
 						r.Spec.FailedAt = vc.nowHandler()
 					}
--- a/datastore/longhorn.go
+++ b/datastore/longhorn.go
@@ -2415,6 +2415,25 @@ func (s *DataStore) IsNodeDownOrDeleted(name string) (bool, error) {
 	return false, nil
 }
 
+// IsNodeDeleted checks whether the node does not exist by passing in the node name
+func (s *DataStore) IsNodeDeleted(name string) (bool, error) {
+	if name == "" {
+		return false, errors.New("no node name provided to check node down or deleted")
+	}
+
+	node, err := s.GetNodeRO(name)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return true, nil
+		}
+		return false, err
+	}
+
+	cond := types.GetCondition(node.Status.Conditions, longhorn.NodeConditionTypeReady)
+
+	return cond.Status == longhorn.ConditionStatusFalse && cond.Reason == longhorn.NodeConditionReasonKubernetesNodeGone, nil
+}
+
 func (s *DataStore) IsNodeSchedulable(name string) bool {
 	node, err := s.GetNodeRO(name)
 	if err != nil {


controller/volume_controller.go
```diff
@@ -1585,6 +1585,15 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
 					continue
 				}
 
+				// The node where the replica is located is deleted, the replica can be directly marked as failed.
+				if _, err := vc.ds.GetNodeRO(r.Spec.NodeID); err != nil && apierrors.IsNotFound(err) {
```
c3y1huang, line 1589: Should also handle other errors.

c3y1huang (changes requested):

controller/volume_controller.go
weizhe0422: Updated to skip replica if a non-`NotFound` error is encountered. Thank you.

innobead (changes requested):
This is like a workaround. What we should fix is probably need to refactor the below method to make it deterimine two different, since one returend bool varaible can't achieve that. 

```go
func (s *DataStore) IsNodeDownOrDeleted(name string) (bool, error) {
	if name == "" {
		return false, errors.New("no node name provided to check node down or deleted")
	}
	node, err := s.GetNodeRO(name)
	if err != nil {
			return true, nil
		}
		return false, err
	}
	cond := types.GetCondition(node.Status.Conditions, longhorn.NodeConditionTypeReady)
	if cond.Status == longhorn.ConditionStatusFalse &&
		(cond.Reason == string(longhorn.NodeConditionReasonKubernetesNodeGone) ||
			cond.Reason == string(longhorn.NodeConditionReasonKubernetesNodeNotReady)) {
		return true, nil
	}
	return false, nil
}
```

weizhe0422:
> > @innobead I think that adding a returned string value to indicate the node state can allow the caller to accurately distinguish.
> 
> How about using two returned bool values to make a simple change like `func (s *DataStore) IsNodeDownOrDeleted(name string) (nodeDown bool, nodeDeleted bool, error)`

I split `IsNodeDownOrDeleted()` into two return parameters: `isNodeDeleted` and `isNodeDown`, like this:
```golang
isNodeDeleted, isNodeDown, err := imc.ds.IsNodeDownOrDeleted(im.Spec.NodeID)
```

shuo-wu:
> How about using two returned bool values to make a simple change like func (s *DataStore) IsNodeDownOrDeleted(name string) (nodeDown bool, nodeDeleted bool, error)

I am thinking if we can add a field `status.state` in Node, then there is no need to use function `IsNodeDownOrDeleted` outside of the node controller.

innobead:
yeah, it will be helpful, but since right now we can determine it from the existing state, so it would be nice to have.

datastore/longhorn.go
```diff
@@ -2395,24 +2395,25 @@ func (s *DataStore) IsNodeDownOrDeletedOrMissingManager(name string) (bool, erro
 
 // IsNodeDownOrDeleted gets Node for the given name and namespace and checks
 // if the Node condition is gone or not ready
-func (s *DataStore) IsNodeDownOrDeleted(name string) (bool, error) {
+func (s *DataStore) IsNodeDownOrDeleted(name string) (nodeDelete bool, nodeDown bool, err error) {
```
innobead, line 2398: No need to modify this method, because what you want to determine is to check if node is deleted or not.

Create another method `isNodeDeleted` as below, and use it in the right place. Of course, remember to add doc comment to this exported method. 

```go
func (s *DataStore) IsNodeDeleted(name string) (bool, error) {
    if name == "" {
        return false, false, errors.New("no node name provided to check node down or deleted")
    }

    node, err := s.GetNodeRO(name)
    if err != nil {
            return true, nil
	}
        return false, err
    }

    cond := types.GetCondition(node.Status.Conditions, longhorn.NodeConditionTypeReady)

    return cond.Status == longhorn.ConditionStatusFalse && cond.Reason == longhorn.NodeConditionReasonKubernetesNodeGone, nil
}    
```

innobead (changes requested):
Some comments need to resolve.

weizhe0422:
Update test result
 - Function test
   1. Having a volume with 3 replicas
   2. Delete one node
   3. The state of the replica on the deleted node will become `Failed`
      image
  - e2e test: test_basic.py
    - fail cases are related to known issues which is https://github.com/longhorn/longhorn/issues/5849
    - image

controller/volume_controller.go
```diff
@@ -1585,11 +1585,19 @@ func (vc *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[stri
...
 				}
 
-				// Whenever the engine isn't attached, if the node goes down, the replica can be marked as failed.
-				// In the attached mode, we can determine whether the replica can fail by relying on the data plane's
-				// connectivity status.
-				if v.Status.State != longhorn.VolumeStateAttached {
-					log.WithField("replica", r.Name).Warnf("Replica %v is marked as failed since the volume %v is not attached because the instance manager is unable to launch the replica", r.Name, v.Name)
+				// If the engine isn't attached or the node goes down, the replica can be marked as failed.
+				// In the attached mode, we can determine whether the replica can fail by relying on the data plane's connectivity status.
+				isNodeDeleted, err := vc.ds.IsNodeDeleted(r.Spec.NodeID)
```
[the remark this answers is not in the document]
weizhe0422: Got it, I have updated the variable name as `nodeDeleted` Thank you.

innobead:
@Mergifyio backport v1.4.x

PR #1775, opened 2023-03-17
Create a new setting so that Longhorn will allow the drain when the node contains the last healthy replica of a volume but the replica is stopped

PhanLe1010:
longhorn/longhorn#5549

### DIFF
--- a/controller/instance_manager_controller.go
+++ b/controller/instance_manager_controller.go
@@ -681,11 +681,20 @@ func (imc *InstanceManagerController) canDeleteInstanceManagerPDB(im *longhorn.I
 		return false, nil
 	}
 
+	// TODO: this setting is deprecated and will be replaced by nodeDrainingPolicy in the future
 	allowDrainingNodeWithLastReplica, err := imc.ds.GetSettingAsBool(types.SettingNameAllowNodeDrainWithLastHealthyReplica)
 	if err != nil {
 		return false, err
 	}
 	if allowDrainingNodeWithLastReplica {
+		return true, nil
+	}
+
+	nodeDrainingPolicy, err := imc.ds.GetSettingValueExisted(types.SettingNameNodeDrainPolicy)
+	if err != nil {
+		return false, err
+	}
+	if nodeDrainingPolicy == string(types.NodeDrainPolicyAlwaysAllow) {
 		return true, nil
 	}
 
@@ -697,10 +706,21 @@ func (imc *InstanceManagerController) canDeleteInstanceManagerPDB(im *longhorn.I
 		return false, err
 	}
 
-	// For each replica process in the current node,
+	targetReplicas := []*longhorn.Replica{}
+	if nodeDrainingPolicy == string(types.NodeDrainPolicyAllowIfReplicaIsStopped) {
+		for _, replica := range replicasOnCurrentNode {
+			if replica.Spec.DesireState != longhorn.InstanceStateStopped || replica.Status.CurrentState != longhorn.InstanceStateStopped {
+				targetReplicas = append(targetReplicas, replica)
+			}
+		}
+	} else {
+		targetReplicas = replicasOnCurrentNode
+	}
+
+	// For each replica in the target replica list,
 	// find out whether there is a PDB protected healthy replica of the same
 	// volume on another schedulable node.
-	for _, replica := range replicasOnCurrentNode {
+	for _, replica := range targetReplicas {
 		vol, err := imc.ds.GetVolume(replica.Spec.VolumeName)
 		if err != nil {
 			return false, err
--- a/types/setting.go
+++ b/types/setting.go
@@ -70,6 +70,7 @@ const (
 	SettingNameReplicaZoneSoftAntiAffinity                              = SettingName("replica-zone-soft-anti-affinity")
 	SettingNameNodeDownPodDeletionPolicy                                = SettingName("node-down-pod-deletion-policy")
 	SettingNameAllowNodeDrainWithLastHealthyReplica                     = SettingName("allow-node-drain-with-last-healthy-replica")
+	SettingNameNodeDrainPolicy                                          = SettingName("node-drain-policy")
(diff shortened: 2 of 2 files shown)


PhanLe1010:
TODO: I am planing to deprecate the setting `Allow Node Drain with the Last Healthy Replica` and replace it by `The Policy To Use When A Node With The Last Healthy Replica Of A Volume Is Drained`. I will need to figure out the migration path for the settings.

innobead:
@PhanLe1010 should not deprecate the setting for a patch but rather a feature release only. Basically, we should make it work as expected, but make it deprecated from 1.5.0, then remove out from 1.6.0.

types/setting.go
```diff
@@ -70,6 +70,7 @@ const (
 	SettingNameReplicaZoneSoftAntiAffinity                              = SettingName("replica-zone-soft-anti-affinity")
 	SettingNameNodeDownPodDeletionPolicy                                = SettingName("node-down-pod-deletion-policy")
 	SettingNameAllowNodeDrainWithLastHealthyReplica                     = SettingName("allow-node-drain-with-last-healthy-replica")
+	SettingNameNodeWithLastHealthyReplicaDrainPolicy                    = SettingName("node-with-last-healthy-replica-drain-policy")
```
c3y1huang, line 73: Is it possible to consider a more generic setting name, such as `node-drain-policy`? I am thinking if we can move `kubernetes-cluster-autoscaler-enabled` as a policy.

PhanLe1010: > Is it possible to consider a more generic setting name, such as node-drain-policy

Great idea! Surely 

> I am thinking if we can move `kubernetes-cluster-autoscaler-enabled` as a policy.

How does this policy work? Is it correct that this policy would be equivalent to `block` because we need to protect last replica even if it is a stopped replica?

c3y1huang: > How does this policy work? Is it correct that this policy would be equivalent to block because we need to protect last replica even if it is a stopped replica?

Sorry I've confused you earlier.  Unfortunately, the logic for `kubernetes-cluster-autoscaler-enabled` doesn't work for this case because It will protect the last replica IM even if the volume is detached. Ref: https://github.com/longhorn/longhorn-manager/blob/master/controller/instance_manager_controller.go#L764-L780

PhanLe1010: > 

Ok. I will create another policy `fall-back-to-allow-node-drain-with-last-healthy-replica-setting` as default policy for the new setting. In the future, we can drop support for that policy when the old setting is removed. What do you think?

PhanLe1010:
> @PhanLe1010 should not deprecate the setting for a patch but rather a feature release only. Basically, we should make it work as expected, but make it deprecated from 1.5.0, then remove out from 1.6.0.

Ok. I will create another policy fall-back-to-allow-node-drain-with-last-healthy-replica-setting as default policy for the new setting. In the future, we can drop support for that policy when the old setting is removed. Any better idea? @innobead

controller/instance_manager_controller.go
```diff
@@ -681,26 +681,36 @@ func (imc *InstanceManagerController) canDeleteInstanceManagerPDB(im *longhorn.I
 		return false, nil
 	}
 
-	allowDrainingNodeWithLastReplica, err := imc.ds.GetSettingAsBool(types.SettingNameAllowNodeDrainWithLastHealthyReplica)
+	nodeWithLastReplicaDrainingPolicy, err := imc.ds.GetSettingValueExisted(types.SettingNameNodeWithLastHealthyReplicaDrainPolicy)
```
innobead, line 684: As I mentioned, let's keep the original code together with this new setting to make sure it's still supported until it's removed.

P.S. let's create a ticket for that setting removal and update the description of `SettingNameAllowNodeDrainWithLastHealthyReplica` deprecated. 

```go
	allowDrainingNodeWithLastReplica, err := imc.ds.GetSettingAsBool(types.SettingNameAllowNodeDrainWithLastHealthyReplica)
	if err != nil {
		return false, err
	}

    if allowDrainingNodeWithLastReplica {
        return true, nil
    }
	
    nodeWithLastReplicaDrainingPolicy, err := imc.ds.GetSettingValueExisted(types.SettingNameNodeWithLastHealthyReplicaDrainPolicy)
    if err != nil {
        return false, err
    }

	if nodeWithLastReplicaDrainingPolicy {
		return true, nil
	}
```

innobead (changes requested):
LGTM, just need to resolve one comment.

innobead:
> Ok. I will create another policy `fall-back-to-allow-node-drain-with-last-healthy-replica-setting` as the default policy for the new setting. In the future, we can drop support for that policy when the old setting is removed. Any better idea? @innobead

Sorry to confuse that. We just need to tackle https://github.com/longhorn/longhorn-manager/pull/1775#discussion_r1139912378. 

No need to have upgrade migration for the old setting now because both of them are supported, so we can tackle the migration later when the setting is deprecated and removed. WDYT?

types/setting.go
PhanLe1010: Thank you for the confirmation. Yes, I agree that `kubernetes-cluster-autoscaler-enabled` doesn't control whether we should remove the PDB for IM when the last replica of a volume is on the node. It is just a trigger condition for the function `canDeleteInstanceManagerPDB`. Another trigger is when a node becomes unscheduled.

controller/instance_manager_controller.go
PhanLe1010: Ok. I will keep both old setting and new setting. There will be a bit of confusion for the user as which setting will take precedence in the case:
* old setting blocks, new setting allows -> allow
* old setting allows, new setting blocks -> allow
But I guess it is fine temporary until we remove the old setting the confusion will go away.

innobead (approved):

innobead:
@PhanLe1010 please update the manifests in longhorn/longhorn as well.

PhanLe1010:
> @PhanLe1010 please update the manifests in longhorn/longhorn as well.

Thank you very much

https://github.com/longhorn/longhorn/pull/5586

innobead:

PR #1784, opened 2023-03-20
Support filesystem trim on encrypted volumes

derekbit:
longhorn/longhorn#5599

### DIFF
--- a/manager/volume.go
+++ b/manager/volume.go
@@ -12,6 +12,8 @@ import (
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
+	"github.com/longhorn/longhorn-manager/csi"
+	"github.com/longhorn/longhorn-manager/csi/crypto"
 	"github.com/longhorn/longhorn-manager/datastore"
 	"github.com/longhorn/longhorn-manager/engineapi"
 	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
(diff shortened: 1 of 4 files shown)


manager/volume.go
```diff
@@ -677,6 +709,22 @@ func (m *VolumeManager) TrimFilesystem(name string) (v *longhorn.Volume, err err
...
+		if enabled {
+			passphrase := m.GetPassphraseForEnablingDiscardsOnEncryptedVolume(v)
+			logrus.Infof("Enabling discards on encrypted volume %v", v.Name)
+
+			err := crypto.EnableDiscardsOnEncryptedVolume(passphrase, v.Name)
+			if err != nil {
+				logrus.WithError(err).Errorf("Failed to enable discards on encrypted volume %v", v.Name)
+			}
+		}
+	}
```
shuo-wu, line 726: Should Longhorn handle this in volume-controller instead? This does not work if users issue `fstrim` by themselves Although doing this manually for RWX volumes is a little bit trivial and I am not sure how many users will try it.

derekbit: Because we already support RWX volume trim in https://github.com/longhorn/longhorn/issues/5143, I think we can put it here for simplicity. WDYT?

shuo-wu: Personally, I would not prefer to add this kind of reconciliation logic in this pkg. And we don't know if users really use the HTTP API to handle trim...

derekbit: OK. We can move it to volume controller. Then, we need add a new spec/status indicating the discards enabling.

BTW, why do we put `fstrim` in HTTP API layer rather than in volume controller?

shuo-wu: `fstrim` is like a single and simple sync action. And there is no reconciliation with other resources. Putting this operation in volume_controller would require a set of spec/status fields with lots of extra logic.
And actually, I don't think the majority of users will do this via API... Adding this API would inform users that Longhorn supports fstrim now.

derekbit: > `fstrim` is like a single and simple sync action. And there is no reconciliation with other resources. Putting this operation in volume_controller would require a set of spec/status fields with lots of extra logic. And actually, I don't think the majority of users will do this via API... Adding this API would inform users that Longhorn supports fstrim now.

Curious, where do users execute fstrim in your thought if not via the API?

shuo-wu: IIRC, users can issue `fstrim` inside the workload pods. I am not sure how many percentage of users would prefer to do it via HTTP API, and I am curious about that as well.

OK, one bad result of this API is, users cannot issue `fstrim` for Longhorn volumes via `kubectl`. Maybe I need to consider refactoring it.

innobead:
@derekbit Please help resolve the conflict.

manager/volume.go
derekbit: @shuo-wu
Just found `discards` cannot be disabled on a volume after allowing it.

My thought is 
- If `support-filesystem-trim-on-encrypted-volume` is not enabled, trimming volume operation will always fail
- If `support-filesystem-trim-on-encrypted-volume` is enabled, `--allow-disacrds` is enabled and allow trimming operation.
Enabling `--allow-discards` introduce a security concern, so I think we can delay it until the trim operation is issued.
WDYT?

derekbit: Any thought? cc @innobead

innobead: I prefer @shuo-wu to make the controller handle spec/status changes to ensure the only place to handle the change is via CR update. 

We can do this way:
- In the volume controller, handle the volume if the setting is enabled (global or volume level). After enabled, the setting can not be revoked/reset. (based on what you said above, the setting can not be disabled after enabling)

WDYT? 

BTW, what's the security concern?

derekbit: > I saw the below warning from the doc. This seems quite risky.

If it is risky, should we add the setting for users?

derekbit: cc @innobead

innobead: yeah, I am thinking the same question. Let's revisit this later. I think block encryption priority is higher than this due to some security concerns. WDYT?

derekbit: I'm fine with it.
We can handle block encryption first.

PR #1790, opened 2023-03-21
fix(setting): incorrect toleration key/value validation

c3y1huang:
Fix validation.

### DIFF
--- a/types/setting.go
+++ b/types/setting.go
@@ -1455,7 +1455,7 @@ func ValidateAndUnmarshalToleration(s string) (*v1.Toleration, error) {
 
 	if strings.Contains(parts[0], "=") {
 		pair := strings.Split(parts[0], "=")
-		if len(parts) != 2 {
+		if len(pair) != 2 {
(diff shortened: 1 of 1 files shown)


types/setting.go
```diff
@@ -1455,7 +1455,7 @@ func ValidateAndUnmarshalToleration(s string) (*v1.Toleration, error) {
 
 	if strings.Contains(parts[0], "=") {
 		pair := strings.Split(parts[0], "=")
-		if len(parts) != 2 {
```
innobead, line 1458: Does this mean toleration always not working before?

c3y1huang: `parts` have already been checked above.
Ref: https://github.com/longhorn/longhorn-manager/blob/master/types/setting.go#L1446-L1448

innobead: I see, so it could be just working coincidentally.

innobead (approved):
LGTM

innobead (approved):
It's better to have unit tests for this public function `ValidateAndUnmarshalToleration`.

innobead:

innobead:
This is a minor fix, the issue is optional.

PR #1791, opened 2023-03-21
fix (backing image): fix bids not found crash during clean up

ChanYiLin:
Ref: [longhorn/longhorn5626](https://github.com/longhorn/longhorn/issues/5626)

When deleting the backing image, and there is no backing image data source (missing)
The Controller crashes cause it doesn't catch not found error and use nil pointer.

We should consider the case of bids not found.

### DIFF
--- a/controller/backing_image_controller.go
+++ b/controller/backing_image_controller.go
@@ -209,16 +209,12 @@ func (bic *BackingImageController) syncBackingImage(key string) (err error) {
 			return err
 		}
 		if len(replicas) != 0 {
-			log.Info("Need to wait for all replicas stopping using this backing image before removing the finalizer")
+			log.Warn("Need to wait for all replicas stopping using this backing image before removing the finalizer")
 			return nil
 		}
-		cleanUpTmpFileFinished, err := bic.IsBackingImageDataSourceCleanup(backingImage)
-		if err != nil {
+		if _, err := bic.IsBackingImageDataSourceCleaned(backingImage); err != nil {
+			log.Warn("Need to wait until backing image data source is cleaned before removing the finalizer")
 			return err
-		}
-		if !cleanUpTmpFileFinished {
-			log.Info("Requesting to clean up Backing Image Data Source first before cleaning up Backing Image Manager")
-			return nil
 		}
 		log.Info("No replica is using this backing image, will clean up the record for backing image managers and remove the finalizer then")
 		if err := bic.cleanupBackingImageManagers(backingImage); err != nil {
@@ -287,9 +283,15 @@ func (bic *BackingImageController) syncBackingImage(key string) (err error) {
 	return nil
 }
 
-func (bic *BackingImageController) IsBackingImageDataSourceCleanup(bi *longhorn.BackingImage) (finished bool, err error) {
+func (bic *BackingImageController) IsBackingImageDataSourceCleaned(bi *longhorn.BackingImage) (cleaned bool, err error) {
+	log := getLoggerForBackingImage(bic.logger, bi)
+
 	bids, err := bic.ds.GetBackingImageDataSource(bi.Name)
-	if err != nil && !apierrors.IsNotFound(err) {
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			log.Warn("can not get backing image data source, return cleaned so the controller won't wait until bids become failed-and-cleanup")
+			return true, nil
+		}
 		return false, err
 	}
 
@@ -297,20 +299,11 @@ func (bic *BackingImageController) IsBackingImageDataSourceCleanup(bi *longhorn.
 		return true, nil
 	}
 
-	// when bids status is ready or ready for transfer, there is already no tmp file left
-	if bids.Status.CurrentState == longhorn.BackingImageStateFailedAndCleanUp ||
-		bids.Status.CurrentState == longhorn.BackingImageStateReady ||
-		bids.Status.CurrentState == longhorn.BackingImageStateReadyForTransfer {
+	if bids.Status.CurrentState == longhorn.BackingImageStateFailedAndCleanUp {
 		return true, nil
 	}
 
-	// mark the status to failed so manager can clean up the orphan tmp file
-	bids.Status.Message = "backing image is deleted, try to cleaning up the tmp file of data source"
-	bids.Status.CurrentState = longhorn.BackingImageStateFailed
-	if _, err = bic.ds.UpdateBackingImageDataSourceStatus(bids); err != nil {
-		return false, err
-	}
-	return false, nil
+	return false, fmt.Errorf("backing image data source status is %v not %v", bids.Status.CurrentState, longhorn.BackingImageStateFailedAndCleanUp)
 }
 
 func (bic *BackingImageController) cleanupBackingImageManagers(bi *longhorn.BackingImage) (err error) {
--- a/controller/backing_image_data_source_controller.go
+++ b/controller/backing_image_data_source_controller.go
@@ -275,6 +275,25 @@ func (c *BackingImageDataSourceController) syncBackingImageDataSource(key string
 		if err := c.cleanup(bids); err != nil {
 			return err
 		}
+
+		// if it is not transferred, we need to wait until it is failed-and-cleanup
+		if !bids.Spec.FileTransferred {
+			if bids.Status.CurrentState == longhorn.BackingImageStateFailedAndCleanUp {
+				return c.ds.RemoveFinalizerForBackingImageDataSource(bids)
+			}
+
+			// if bids is not transferred
+			// mark the status to failed so manager can clean up the tmp file and mark it as failed-and-cleanup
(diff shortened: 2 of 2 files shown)


controller/backing_image_controller.go
```diff
@@ -287,30 +287,34 @@ func (bic *BackingImageController) syncBackingImage(key string) (err error) {
...
-	bids.Status.Message = "backing image is deleted, try to cleaning up the tmp file of data source"
+	// mark the status to failed so manager can clean up the tmp file
+	bids.Status.Message = "backing image is deleted, requesting manager to clean up the tmp file of backing image data source"
 	bids.Status.CurrentState = longhorn.BackingImageStateFailed
 	if _, err = bic.ds.UpdateBackingImageDataSourceStatus(bids); err != nil {
-		return false, err
+		return false, false, err
 	}
-	return false, nil
+	return false, true, err
```
innobead, line 317: `return false, true, nil` if we ensure there should not be any error if `requested` is true.

controller/backing_image_controller.go
```diff
@@ -287,30 +287,34 @@ func (bic *BackingImageController) syncBackingImage(key string) (err error) {
 	return nil
 }
 
-func (bic *BackingImageController) IsBackingImageDataSourceCleanup(bi *longhorn.BackingImage) (finished bool, err error) {
+func (bic *BackingImageController) IsBackingImageDataSourceCleaned(bi *longhorn.BackingImage) (cleaned bool, requested bool, err error) {
```
innobead, line 290: Can rename the returned variables. `(cleaned bool, requested bool, err requestError)`

innobead (approved):
In general, LGTM. Just some comments needed to resolve.

controller/backing_image_controller.go
innobead: lol you are right, just typo.

ChanYiLin:
Fixed and tested

So the idea is

### Before the Transfer: should clean up the tmp file of data source
1. delete the `backingimage`, the deletion signal pass to `backing image data source`
2. we should wait until the tmp file is cleaned up so the backing image data source can be deleted
3. `backing image data source` will wait until it is in states (`failed-and-cleanup`, `ready`, `ready-for-transfer`)
4. `backing image controller` will check the status of `backing image data source`
    -  `backing image data source` is in (`failed-and-cleanup`, `ready`, `ready-for-transfer`) return cleaned and can be deleted
    - `backing image data source` is in other states, mark it as `failed`
5. `backing image manager controller` will check the `backing image data source failed state` and delete the tmp file and mark `backing image data source` as `failed-and-cleanup`
6. `backingimage` and `backing image data source`  can be deleted (finalizer can be removed)

### After Transfer: there is no tmp files
1. delete the backingimage, the deletion signal pass to backing image data source
2. `backing image data source` is transferred so it will be deleted right away (maybe even prior to `backingimage`)
3. `backingimage controller` try to get the `backing image data source`
    - `backing image data source` not found since it is deleted first, return cleaned anyway (there is no tmp file)
    - `backing image data source` found, but it is transferred, return cleaned anyway (there is no tmp file)
4. `backingimage` can be deleted and should not stuck
 

The potential issue:
- one can not delete the `backing image data source CR` during `in-progress`
    - the deletion signal will pass to the `backing image data source pod`
    - the `backing image data source CR` will stuck in `in-progress` and the deletion will stuck because the `backing image data source` will never be in states (`failed-and-cleanup`, `ready`, `ready-for-transfer`) anymore.

controller/backing_image_controller.go
```diff
@@ -212,12 +212,12 @@ func (bic *BackingImageController) syncBackingImage(key string) (err error) {
...
 		}
-		cleanUpTmpFileFinished, err := bic.IsBackingImageDataSourceCleanup(backingImage)
+		cleaned, err := bic.IsBackingImageDataSourceCleaned(backingImage)
 		if err != nil {
 			return err
 		}
-		if !cleanUpTmpFileFinished {
-			log.Info("Requesting to clean up Backing Image Data Source first before cleaning up Backing Image Manager")
+		if !cleaned {
+			log.Info("Need to wait until backing image data source is cleaned before removing the finalizer")
```
shuo-wu, line 220: Is this reachable? There is already an error when `IsBackingImageDataSourceCleaned` returns `false`

ChanYiLin: yes you are right, 
`cleaned=false` will always follow an error returned
so it will be caught already

ChanYiLin:
Hi @innobead 
I have moved the requesting logic to **bids controller** and done some refactor

When deleting
- if `bids` is transferred, it can be deleted it right away
- if `bids` is not transferred, `bids controller` will mark `bids` as `failed` no matter what
    - `bim controller` will see (!transferred && failed) -> cleanup and mark `bids` as `failed-and-cleanup`
- if `bids` is not transferred and is in `failed-and-cleanup`, it can be delete

For `bi`
- if cannot get bids -> continue to cleanup bim and delete bi
- if bids is transferred -> continue to cleanup bim and delete bi
- if bids is not transferred -> wait until it is `failed-and-cleanup` then cleanup bim and delete bi

controller/backing_image_controller.go
```diff
@@ -212,14 +212,10 @@ func (bic *BackingImageController) syncBackingImage(key string) (err error) {
 			log.Info("Need to wait for all replicas stopping using this backing image before removing the finalizer")
 			return nil
 		}
-		cleanUpTmpFileFinished, err := bic.IsBackingImageDataSourceCleanup(backingImage)
-		if err != nil {
+		if _, err := bic.IsBackingImageDataSourceCleaned(backingImage); err != nil {
+			log.Info("Need to wait until backing image data source is cleaned before removing the finalizer")
```
innobead, line 216: Suggest using warn because actually here we encountered an error, even though it will be returned to the caller.

controller/backing_image_controller.go
```diff
@@ -287,30 +283,27 @@ func (bic *BackingImageController) syncBackingImage(key string) (err error) {
 	return nil
 }
 
-func (bic *BackingImageController) IsBackingImageDataSourceCleanup(bi *longhorn.BackingImage) (finished bool, err error) {
+func (bic *BackingImageController) IsBackingImageDataSourceCleaned(bi *longhorn.BackingImage) (cleaned bool, err error) {
 	bids, err := bic.ds.GetBackingImageDataSource(bi.Name)
-	if err != nil && !apierrors.IsNotFound(err) {
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			// can not get backing image data source, return cleaned so the controller won't wait until bids become failed-and-cleanup
```
innobead, line 290: Actually you can use log trace for this comment. Ideally we should convert all comments to trace logs instead. (this can be improved later when you are working on log & error refactoring)

controller/backing_image_controller.go
```diff
@@ -287,30 +283,27 @@ func (bic *BackingImageController) syncBackingImage(key string) (err error) {
...
+			return true, nil
+		}
 		return false, err
 	}
 
+	// if bids is transferred, there is no tmp file left
```
innobead, line 296: The same, suggest using trace log.

controller/backing_image_controller.go
```diff
@@ -287,30 +283,27 @@ func (bic *BackingImageController) syncBackingImage(key string) (err error) {
...
 	if bids.Spec.FileTransferred {
 		return true, nil
 	}
 
-	// when bids status is ready or ready for transfer, there is already no tmp file left
-	if bids.Status.CurrentState == longhorn.BackingImageStateFailedAndCleanUp ||
-		bids.Status.CurrentState == longhorn.BackingImageStateReady ||
-		bids.Status.CurrentState == longhorn.BackingImageStateReadyForTransfer {
+	// if bids is not transferred, wait until it is failed-and-cleanup
```
innobead, line 301: The same, suggest using trace log.

controller/backing_image_controller.go
```diff
@@ -287,30 +283,27 @@ func (bic *BackingImageController) syncBackingImage(key string) (err error) {
...
 	}
 
-	// mark the status to failed so manager can clean up the orphan tmp file
-	bids.Status.Message = "backing image is deleted, try to cleaning up the tmp file of data source"
-	bids.Status.CurrentState = longhorn.BackingImageStateFailed
-	if _, err = bic.ds.UpdateBackingImageDataSourceStatus(bids); err != nil {
-		return false, err
-	}
-	return false, nil
+	return false, fmt.Errorf("backing image data source status is %v not failed-and-cleanup", bids.Status.CurrentState)
```
innobead, line 306: Replace `failed-and-cleanup` in log with `longhorn.BackingImageStateFailedAndCleanUp`

innobead (changes requested):
LGTM. Just some minor comments needed to fix.

shuo-wu (approved):
In general LGTM

innobead (approved):

innobead:

PR #1804, opened 2023-03-23
style: reformat backing image controller log message

ChanYiLin:
Ref: [longhorn/longhorn#5015](https://github.com/longhorn/longhorn/issues/5015)

- Logs should start with uppercase V+ed or V+ing
- For errors, need to try starting with a lowercase Verb word with different tenses (V+ing, V+ed). If there is an error, need to wrap in the error instead of string interpolation in the message directly.
- Change `Debug` logs to `Info`, `Warn` or remove them
- If the log doesn't have args, use `info` instead of `infof`


ChanYiLin:
@innobead we can use this PR as example and make sure we are on the same page of logging reformat, thanks!

controller/backing_image_controller.go
```diff
@@ -145,13 +145,13 @@ func (bic *BackingImageController) handleErr(err error, key interface{}) {
 	}
 
 	if bic.queue.NumRequeues(key) < maxRetries {
-		logrus.Warnf("Error syncing Longhorn backing image %v: %v", key, err)
+		logrus.WithError(err).Warnf("error syncing Longhorn backing image %v", key)
```
derekbit, line 148: IIRC, @innobead  we use uppercase for the log message, right?

derekbit:
IIRC, @innobead  we use uppercase for the log message regardless of the level as a convention, right?

```
logrus.WithError(err).Error("Error....")
```

```
return fmt.Errorf("error ...")
```

derekbit:
IIRC, @innobead  we use uppercase for the log message regardless of the level as a convention, right?

```
logrus.WithError(err).Error("Error....")
```

```
return fmt.Errorf("error ...")
```

controller/backing_image_controller.go
derekbit, line 148: Violate the existing convention. Please check https://github.com/longhorn/longhorn-manager/pull/1804#issuecomment-1480707966

controller/backing_image_controller.go
```diff
@@ -145,13 +145,13 @@ func (bic *BackingImageController) handleErr(err error, key interface{}) {
...
 		bic.queue.AddRateLimited(key)
 		return
 	}
 
 	utilruntime.HandleError(err)
-	logrus.Warnf("Dropping Longhorn backing image %v out of the queue: %v", key, err)
+	logrus.WithError(err).Warnf("dropping Longhorn backing image %v out of the queue", key)
```
derekbit, line 154: See 1.

controller/backing_image_controller.go
```diff
@@ -178,7 +178,7 @@ func (bic *BackingImageController) syncBackingImage(key string) (err error) {
 	backingImage, err := bic.ds.GetBackingImage(name)
 	if err != nil {
 		if !datastore.ErrorIsNotFound(err) {
-			bic.logger.WithField("backingImage", name).WithError(err).Error("Failed to retrieve backing image from datastore")
+			bic.logger.WithField("backingImage", name).WithError(err).Error("failed to retrieve backing image from datastore")
```
derekbit, line 181: See 1.

controller/backing_image_controller.go
```diff
@@ -209,11 +209,11 @@ func (bic *BackingImageController) syncBackingImage(key string) (err error) {
 			return err
 		}
 		if len(replicas) != 0 {
-			log.Warn("Need to wait for all replicas stopping using this backing image before removing the finalizer")
+			log.Warn("waiting for all replicas stopping using this backing image before removing the finalizer")
```
derekbit, line 212: See 1.

ChanYiLin:
> ~IIRC, @innobead we use uppercase for the log message regardless of the level as a convention, right?~
> 
> We use uppercase for the log message regardless of the level as a convention. [longhorn/longhorn#5015 (comment)](https://github.com/longhorn/longhorn/issues/5015#issue-1481250309)
> 
> ```
> logrus.WithError(err).Error("Error....")
> ```
> 
> ```
> return fmt.Errorf("error ...")
> ```

Oh I misunderstood the description in [#5015](https://github.com/longhorn/longhorn/issues/5015)
For logging no matter what level, start with uppercase.
Error message means message inside `error` not logging, and it should start with lowercase

controller/backing_image_data_source_controller.go
```diff
@@ -300,15 +300,15 @@ func (c *BackingImageDataSourceController) syncBackingImageDataSource(key string
 	existingBIDS := bids.DeepCopy()
 	defer func() {
 		if err != nil && strings.Contains(err.Error(), "need to wait for volume") {
-			log.Infof("Need to wait for volume attachment before handling key %v: %v", key, err)
+			log.WithError(err).Infof("Need to wait for volume attachment before handling key %v", key)
```
innobead, line 303: Change to warnf instead.

innobead (changes requested):
In general, LGTM. Just one log level needs to change. Thanks!

controller/backing_image_data_source_controller.go
ChanYiLin: fixed, thanks!

innobead (approved):

innobead:
@Mergifyio backport v1.4.x

PR #1805, opened 2023-03-24
refactor: merge webhook and recovery service into longhorn manager

ChanYiLin:
Ref: [longhorn/longhorn#5590](https://github.com/longhorn/longhorn/issues/5590)

Update Note:
- start conversion webhook service and wait until it is ready in longhorn-manager
    - change to use port 9443 (so there is no need to update crd webhook field)
- start adminssion webhook service and wait until it is ready in longhorn-manager
    - change to use port 8443
- start recovery backend service in longhorn-manager
- remove webhook and recovery deployment related code in pdb controller
- remove webhook and recovery deployment related code in uninstall controller 
- delete webhook/recovery backend deployments in `upgrade path`


app/daemon.go
```diff
@@ -218,6 +238,97 @@ func environmentCheck() error {
 	return nil
 }
 
+func startWebhook(ctx context.Context, serviceAccount, kubeconfigPath, webhookType string) error {
```
shuo-wu, line 241: NIT: Can we retain this function in `app/webhook.go`?

app/daemon.go
```diff
@@ -218,6 +238,97 @@ func environmentCheck() error {
...
+		return fmt.Errorf("unexpected webhook server type %v", webhookType)
+	}
+
+	namespace := os.Getenv(types.EnvPodNamespace)
+	if namespace == "" {
+		logrus.Warnf("Failed to detect pod namespace, environment variable %v is missing, "+
+			"using default namespace", types.EnvPodNamespace)
+		namespace = corev1.NamespaceDefault
+	}
+
```
shuo-wu, line 260: NIT: Abstract this as a util function.

types/types.go
```diff
@@ -80,8 +80,9 @@ const (
 )
 
 const (
-	DefaultAPIPort           = 9500
-	DefaultWebhookServerPort = 9443
+	DefaultAPIPort               = 9500
+	DefaultConversionWebhookPort = 9443
+	DefaultAdmissionWebhookPort  = 8443
```
shuo-wu, line 85: What's the previous admission webhook port?

shuo-wu: Besides, we need to update the doc like [this](https://github.com/longhorn/website/pull/681) after the PR merge.

app/daemon.go
```diff
@@ -132,17 +142,28 @@ func startManager(c *cli.Context) error {
 		return fmt.Errorf("BUG: failed to detect the node IP")
 	}
 
-	done := make(chan struct{})
+	ctx := signals.SetupSignalContext()
```
shuo-wu, line 145: Can you describe a little bit more about this kind of context?

types/types.go
ChanYiLin: both admission and conversion webhook ports were 9443 since they are in different deploy/svc before

but after this PR, both services are in the manager
- conversion webhook: 9443
- admission webhook: 8443

ChanYiLin: Yes, we also need to update the network policy
https://github.com/longhorn/charts/pull/101

app/daemon.go
ChanYiLin: longhorn-manager made a channel `done` and registered it as a shutdown channel to receive shut down signal by `util.RegisterShutdownChannel(done)`
it then passed it to the controllers and the manager API service

For webhook/recovery_backend, it created a context by `ctx := signals.SetupSignalContext()`
then passed it to the API service 

both actually have the same behavior and the purpose is to receive the shutdown signal.
I don't want to make a channel for manager API and context for webhook/recovery_backend.

So I decide to make a context only, and pass ctx.Done as the done channel used in manager API.

ChanYiLin: - longhorn-manager registered the channel to receive the signal: 
https://github.com/longhorn/longhorn-manager/blob/master/app/daemon.go#L203-L207
- RegisterShutdownChannel
https://github.com/longhorn/longhorn-manager/blob/master/util/util.go#L348-L356

- Same as SetupSignalContext (`vendor/github.com/rancher/wrangler/pkg/signals/signal.go`)
```
// SetupSignalContext is same as SetupSignalHandler, but a context.Context is returned.
// Only one of SetupSignalContext and SetupSignalHandler should be called, and only can
// be called once.
func SetupSignalContext() context.Context {
	close(onlyOneSignalHandler) // panics when called twice

	shutdownHandler = make(chan os.Signal, 2)

	ctx, cancel := context.WithCancel(context.Background())
	signal.Notify(shutdownHandler, shutdownSignals...)
	go func() {
		<-shutdownHandler
		cancel()
		<-shutdownHandler
		os.Exit(1) // second signal. Exit directly.
	}()

	return ctx
}
```

types/types.go
shuo-wu: I am thinking you can use port 9501, 9502, 9503... This is more straightforward when we check the port usage.

ChanYiLin: Sure, sounds good to me!
Fixed.

app/daemon.go
shuo-wu: Oh Sorry, this is already used by webhook and recovery_backend. 
Just checked this context. It registers SIGTERM and SIGINT as channel stop signals. This is what we need here.

But here another question is, could a crash of one service (conversion webhook, validation webhook, etc ) lead to channel closing or other services stop?

ChanYiLin:
```
2023-04-07T15:12:32.689034879Z time="2023-04-07T15:12:32Z" level=debug msg="2023/04/07 15:12:32 http: TLS handshake error from 10.0.2.15:52886: EOF"
2023-04-07T15:12:42.688616470Z time="2023-04-07T15:12:42Z" level=debug msg="2023/04/07 15:12:42 http: TLS handshake error from 10.0.2.15:47850: EOF"
2023-04-07T15:12:52.695363407Z time="2023-04-07T15:12:52Z" level=debug msg="2023/04/07 15:12:52 http: TLS handshake error from 10.0.2.15:35264: EOF"
2023-04-07T15:13:02.693107730Z time="2023-04-07T15:13:02Z" level=debug msg="2023/04/07 15:13:02 http: TLS handshake error from 10.0.2.15:54574: EOF"
2023-04-07T15:13:12.702013551Z time="2023-04-07T15:13:12Z" level=debug msg="2023/04/07 15:13:12 http: TLS handshake error from 10.0.2.15:34160: EOF"
```

just found out in local test that there are bunch of tls error, need to do some survey

app/daemon.go
ChanYiLin: The root **context** is created in `daemon.go` and is passed down to manager/webhook/recovery these services
**cancel** is not passed down to services so no one can close the channel.
It only `cancel()` when the shutdown signal is received in the above`SetupSignalContext()` function.

ChanYiLin:
> ```
> 2023-04-07T15:12:32.689034879Z time="2023-04-07T15:12:32Z" level=debug msg="2023/04/07 15:12:32 http: TLS handshake error from 10.0.2.15:52886: EOF"
> 2023-04-07T15:12:42.688616470Z time="2023-04-07T15:12:42Z" level=debug msg="2023/04/07 15:12:42 http: TLS handshake error from 10.0.2.15:47850: EOF"
> 2023-04-07T15:12:52.695363407Z time="2023-04-07T15:12:52Z" level=debug msg="2023/04/07 15:12:52 http: TLS handshake error from 10.0.2.15:35264: EOF"
> 2023-04-07T15:13:02.693107730Z time="2023-04-07T15:13:02Z" level=debug msg="2023/04/07 15:13:02 http: TLS handshake error from 10.0.2.15:54574: EOF"
> 2023-04-07T15:13:12.702013551Z time="2023-04-07T15:13:12Z" level=debug msg="2023/04/07 15:13:12 http: TLS handshake error from 10.0.2.15:34160: EOF"
> ```
> 
> just found out in local test that there are bunch of tls error, need to do some survey but all the functions work fine, weird...

Oh I found the reason, it is caused by the readiness probe every 10 sec, kubelet use http instead of https to probe the pod

app/daemon.go
ChanYiLin: Hi @shuo-wu 
you might be right, the I only run http server in goroutine
```
go s.ListenAndServe()
```
but the goroutine failed won't lead to other services stopped which is not expected.

Let me fix it.

ChanYiLin: I have fixed it by capturing the error returning from ListenAndServe() and fatal exit the program
WDYT?

shuo-wu: The fatal exit will lead to `ctx.Done()` close then all services/the whole pod stop, right? Then that LGTM

app/webhook.go
```diff
@@ -1,77 +1,74 @@
...
+
+	logrus.Infof("Waiting for %v webhook to become ready", webhookType)
+	cli := http.Client{
+		Timeout: time.Second,
+		Transport: &http.Transport{
+			TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+		},
+	}
+	webhookHealthEndpoint := fmt.Sprintf("https://localhost:%d/v1/healthz", webhookPort)
+	for {
```
shuo-wu, line 53: It's better to have a timeout here.

app/webhook.go
```diff
@@ -1,77 +1,74 @@
...
-	"k8s.io/client-go/tools/clientcmd"
+	"io"
+	"net/http"
+	"time"
 
 	"github.com/longhorn/longhorn-manager/types"
+	"github.com/longhorn/longhorn-manager/util"
 	"github.com/longhorn/longhorn-manager/webhook/server"
+	"github.com/sirupsen/logrus"
+	"k8s.io/client-go/tools/clientcmd"
```
shuo-wu, line 15: NIT: rearrange the imports

shuo-wu (commented):
Others LGTM

app/daemon.go
ChanYiLin: it won't lead to ctx.Done()
but the main thread would log fatal and exit(1) and the whole pod will restart.

app/webhook.go
ChanYiLin: Fixed, thanks!

derekbit:
Will the old and unused webhook and recovery-backend pods be cleaned up after upgrade?

ChanYiLin:
> Will the old and unused webhook and recovery-backend pods be cleaned up after upgrade?

Yes, it is implemented in upgrade path
https://github.com/longhorn/longhorn-manager/pull/1805/files#diff-54e77ae1d9149f7ea519019af64a71e75f8029573fbb71ee209c944508a11502R25

derekbit:
> > Will the old and unused webhook and recovery-backend pods be cleaned up after upgrade?
> 
> Yes, it is implemented in upgrade path https://github.com/longhorn/longhorn-manager/pull/1805/files#diff-54e77ae1d9149f7ea519019af64a71e75f8029573fbb71ee209c944508a11502R25

Got it. Thank you.

BTW, can you help update the port `9600` to the new value? 
https://github.com/longhorn/nfs-ganesha/blob/longhorn-ganesha-v4/src/SAL/recovery/recovery_longhorn.c#L42

derekbit:
BTW, can you help create a PR for updating the port 9600 to the new value?
https://github.com/longhorn/nfs-ganesha/blob/longhorn-ganesha-v4/src/SAL/recovery/recovery_longhorn.c#L42

ChanYiLin:
@innobead I have removed the webhook/recovery upgrade path change in 1.3.x -> 1.4.x

ChanYiLin:
> BTW, can you help create a PR for updating the port 9600 to the new value? https://github.com/longhorn/nfs-ganesha/blob/longhorn-ganesha-v4/src/SAL/recovery/recovery_longhorn.c#L42

Hi @derekbit , here is the PR: https://github.com/longhorn/nfs-ganesha/pull/6
thanks!

app/daemon.go
```diff
@@ -218,6 +238,97 @@ func environmentCheck() error {
 	return nil
 }
 
+func startWebhook(ctx context.Context, serviceAccount, kubeconfigPath, webhookType string) error {
+	logrus.Infof("Starting longhorn %s webhook server", webhookType)
+
+	var webhookPort int
+	switch webhookType {
+	case "admission":
```
innobead: Can we use `types.WebhookTypeConversion` instead? the same for `conversion`.

innobead (changes requested):
LGTM, just one comment needed to resolve, then we can merge this PR.

app/daemon.go
ChanYiLin: Fixed, thanks!

PR #1813, opened 2023-03-28
Refactor backup state handling in `backup_monitor.go` and `backup.go`

derekbit:
- Change the return type and add variable initialization in `syncBackupStatusFromEngineReplica` in `backup_monitor.go`
- Add error handling and update `currentBackupStatus` in `syncBackupStatusFromEngineReplica` in `backup_monitor.go`
- Add a new constant `backupStateStarting` in `backups. Go`
- Add a new backup state `Starting` in `backup. Go`

Depends on https://github.com/longhorn/backupstore/pull/117 and https://github.com/longhorn/longhorn-engine/pull/882
Longhorn/longhorn#5662


innobead:
@derekbit need to fix the conflict.

engineapi/backups.go
```diff
@@ -20,6 +20,7 @@ import (
 )
 
 const (
+	backupStateStarting   = "starting"
```
innobead, line 23: Can we just use these constants from https://github.com/longhorn/backupstore/commit/cc70e06e?

innobead (changes requested):
In general, LGTM.

However, please consolidate commits since some commit seems irrelevent like CRD update for share manager resource. 

After engine PR merged and the go dependencies updated, we can merge this PR.

engineapi/backups.go
derekbit: The introduction of `starting` state makes the incompatibility of the new `longhorn-manager` and the old `longhorn-engine`. So, checking the engine CLI/API version is necessary. Do we need to include the feature in v1.4.x? 
cc @innobead

innobead: Where is the incompatible part? I saw all changes are internally for backupstore lib.

innobead: For any incompatible change, as discussed before, no need, and be unsuitable for backport. 

cc @longhorn/dev

derekbit:
The purpose of the PR is to handle the side effect of `soft` mode.
In addition, the introduction of the `starting` state will break the compatibility of the new longhorn-manager and the old backupstore. So, bumping cli/api version and version check before doing a backup is necessary.
Should we backport `soft` mode to v1.4.2? Ref: https://github.com/longhorn/longhorn/issues/5628

cc @innobead

engineapi/backups.go
derekbit: https://github.com/longhorn/longhorn-manager/pull/1813#issuecomment-1531327304

innobead:
> The purpose of the PR is to handle the side effect of `soft` mode. In addition, the introduction of the `starting` state will break the compatibility of the new longhorn-manager and the old backupstore. So, bumping cli/api version and version check before doing a backup is necessary. Should we backport `soft` mode to v1.4.2? Ref: [longhorn/longhorn#5628](https://github.com/longhorn/longhorn/issues/5628)
> 
> cc @innobead

Then, we don't need to backport any incompatible change/fix.

derekbit:
> > The purpose of the PR is to handle the side effect of `soft` mode. In addition, the introduction of the `starting` state will break the compatibility of the new longhorn-manager and the old backupstore. So, bumping cli/api version and version check before doing a backup is necessary. Should we backport `soft` mode to v1.4.2? Ref: [longhorn/longhorn#5628](https://github.com/longhorn/longhorn/issues/5628)
> > cc @innobead
> 
> Then, we don't need to backport any incompatible change/fix.

OK, I will revert the `soft` mode codes in https://github.com/longhorn/backupstore/commits/v1.4.x

innobead:
Thanks @derekbit !

innobead:
@derekbit before reverting, let see the severity first.

What issue the soft mod is trying to fix? IIRC, it's about High cpu loading due to stuck nfs connection?

k8s/pkg/apis/longhorn/v1beta2/backup.go
```diff
@@ -7,6 +7,7 @@ type BackupState string
 const (
 	BackupStateNew        = BackupState("")
 	BackupStatePending    = BackupState("Pending")
+	BackupStateStarting   = BackupState("Starting")
```
innobead, line 10: This is not necessary, since we did not introduce this state from backup status anymore.

innobead (changes requested):

k8s/pkg/apis/longhorn/v1beta2/backup.go
derekbit: Removed. Thank you.

PR #1814, opened 2023-03-29
Update logs for volume spec size change

shuo-wu:
Longhorn/longhorn#5513


derekbit (dismissed):

manager/volume.go
```diff
@@ -552,14 +552,15 @@ func (m *VolumeManager) Expand(volumeName string, size int64) (v *longhorn.Volum
...
 
-	logrus.Infof("Volume %v expansion from %v to %v requested", v.Name, v.Spec.Size, size)
 	v.Spec.Size = size
 
 	v, err = m.ds.UpdateVolume(v)
 	if err != nil {
 		return nil, err
 	}
 
+	logrus.Infof("Expanding volume %v from %v to %v requested", v.Name, v.Spec.Size, size)
```
innobead, line 562: `v.Spec.Size` is the same as `size`. You should log with the original volume size instead.

innobead (changes requested):

innobead:
@Mergifyio backport v1.4.x

PR #1817, opened 2023-03-30
Fix bug create IM PDB with wrong node selector

PhanLe1010:
longhorn/longhorn#5680
longhorn/longhorn#5688


PhanLe1010:
> We should also prevent creating the PDB for an instance manager from a different node something like below. Each PDB should only be created when the corresponding IM pod exists.
> 
> [/controller/instance_manager_controller.go#L428-L430](https://github.com/longhorn/longhorn-manager/blob/79acd764ed501dc29f978d2be47a80941d10ecb4/controller/instance_manager_controller.go#L428-L430)
> 
> ```go
> 	if imc.controllerID != im.Spec.NodeID {
> 		return nil
> 	}
> ```

Discussed with David. We agreed with this statement `Each PDB should only be created when the corresponding IM pod exists`. Together with the fact that sometimes IM is handled by the different longhorn-manager node. In that case, we should still create IM PDBs when corresponding IM pods are running (i.e., still create IM PDBs when corresponding IM pods are running even if the `imc.controllerID != im.Spec.NodeID`)

PhanLe1010:
cc @c3y1huang Could you take a look at this PR?

innobead (approved):
LGTM. Well done.

innobead:
We need to backport this to 1.4.x and 1.3.x.

innobead:
https://github.com/Mergifyio backport v1.4.x v1.3.x

PR #1827, opened 2023-04-06
Increasing MaxRecurringJobRetain to 100

cedenilla:
https://github.com/longhorn/longhorn/discussions/5710

We need to increase the MaxRecurringJobRetain to at least 100 since 50 isn't sufficient in some cases. I understand that this change won't have a significant impact on performance. Please let me know if we should also make this change in other repositories, especially in the Longhorn repository where I couldn't find any reference to it.

Thanks,
Daniel


innobead:
I am fine with this change. 

@c3y1huang WDYT?

c3y1huang:
[Here is the original issue](https://github.com/longhorn/longhorn/issues/1016) on the decision regarding the maximum retaining number.

Given that we have fixed the [snapshot creation issue](https://github.com/longhorn/longhorn/issues/4898), and also introduced other recurring job types, like snapshot-delete/purge, that can help to resolve the problem, I am fine with this change.

c3y1huang:
It may be worth considering making this customizable.
Ref: https://github.com/longhorn/longhorn/issues/5713

cedenilla:
> It may be worth considering making this customizable. Ref: [longhorn/longhorn#5713](https://github.com/longhorn/longhorn/issues/5713)

I completely agree. In fact, I've been reviewing the Helm values.yaml file to check if there's already an option available for setting the value. It would be nice to have this functionality in place rather than relying on a static value

innobead (approved):
LGTM. 

The 'retain value' will restrict the number of snapshots or backups. Having a hard limit on the number of snapshots is reasonable since it ensures space efficiency in the cluster. However, for backups, which are out-of-cluster backups, using the same limitation could be a challenge for users.

For backup, we should respect the retain value. If not specified, should not have any upper bound limitation instead or even having a setting for that, like default-backup-retain.  

cc @shuo-wu @c3y1huang @derekbit @longhorn/qa

innobead:

PR #1828, opened 2023-04-06
Fix bug: volume is stuck in live engine upgrading

PhanLe1010:
1. During the live engine upgrade, if the volume becomes degraded before the engine.Spec.EngineImage is set to new engine image, Longhorn doesn't try to rebuild the volume. Volume never becomes healthy which block the engine live upgrade process
2. Due to ETCD update conflict, sometime the UpgradeEngineProcess() is called multiple times. However, if the first call already succeeded, the subsequent calls will always fail. We shouldn't call the function UpgradeEngineProcess() if the existing process already has the newer engine image

More details are at https://github.com/longhorn/longhorn/issues/5684#issuecomment-1496777410

longhorn/longhorn#5684

### DIFF
--- a/controller/engine_controller.go
+++ b/controller/engine_controller.go
@@ -1997,7 +1997,7 @@ func (ec *EngineController) UpgradeEngineProcess(e *longhorn.Engine, log *logrus
 
 	processBinary, err := c.ProcessGetBinary(e.Name)
 	if err != nil {
-		return errors.Wrapf(err, "fail to get the binary of the current engine process")
+		return errors.Wrapf(err, "failed to get the binary of the current engine process")
 	}
 	if strings.Contains(processBinary, types.GetImageCanonicalName(e.Spec.EngineImage)) {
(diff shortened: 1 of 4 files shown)


innobead (approved):
LGTM

controller/engine_controller.go
```diff
@@ -1997,6 +1995,15 @@ func (ec *EngineController) UpgradeEngineProcess(e *longhorn.Engine) error {
 		return err
 	}
 
+	processBinary, err := c.ProcessGetBinary(e.Name)
+	if err != nil {
+		return errors.Wrapf(err, "fail to get the binary of the current engine process")
```
innobead, line 2000:
```suggestion
		return errors.Wrapf(err, "failed to get the binary of the current engine process")
```

innobead (approved):

innobead:

innobead:
cc @shuo-wu @derekbit

PR #1829, opened 2023-04-07
Improve manageability of share manager pod behaviors

derekbit:
In the current implementation, the share manager pod is created and blindly mounts a volume regardless of its healthy state. The blind mounting behavior can lead to unexpected volume faulted if the volume is dead or orphaned. This ticket aims to share manager pod's behaviors more manageable and predictable.

With the improvement, the start flow will be as follows:

1. Share manager controller creates a share manager pod and bootstraps an NFS server without any filesystem exported
2. Share manager pod mounts the volume and exports it after receiving a FilesystemMount request
3. Share manager controller checks the mount status by FilesystemMountStatus. If the volume is mounted, transition the share manager state to "running"

Shutdown flow will be as follows:

1. Gracefully shut down the NFS server
2. Gracefully or forcefully unmount the volume
3. Tear down the device.

Depends on https://github.com/longhorn/longhorn-share-manager/pull/69

https://github.com/longhorn/longhorn/issues/5717


derekbit:
Revisit it after merging AD controller

PR #1832, opened 2023-04-07
chore: run k8s/generate_code.sh

c3y1huang:
Same reason as https://github.com/longhorn/longhorn-manager/pull/1585.


innobead (approved):

innobead:
cc @derekbit @shuo-wu

PR #1837, opened 2023-04-07
volume mutator: update restoreVolumeRecurringJob and snapshotDataIntegrity

derekbit:
The upgrade path can patch the values and directly update to the API server. However, vefore the volume lister is updated, volume.spec will be upated with outdated values. To mitigate the issue, we need to patch the empty values in the update as well.

Longhorn/longhorn#5485


innobead (approved):

innobead:
@Mergifyio backport v1.4.x

PR #1839, opened 2023-04-07
Bump CSI sidecars v1.5

ejweber:
[longhorn/longhorn#5672](https://github.com/longhorn/longhorn/issues/5672)

We can't move to these versions of external-provisioner and external-snapshotter in Longhorn < v1.5, as they no longer support the v1beta1 VolumeSnapshot API. We deprecated its use in Longhorn v1.4 (https://github.com/longhorn/longhorn/blob/master/CHANGELOG/CHANGELOG-1.4.0.md#deprecation--incompatibilities). I will create separate PRs to bump sidecar versions in the v1.4.x and v1.3.x branches without these components.

EDIT: This comment was previously (incorrectly) missing the "<" sign.

### DIFF
--- a/csi/deployment.go
+++ b/csi/deployment.go
@@ -20,12 +20,12 @@ import (
 )
 
 const (
-	DefaultCSIAttacherImage            = "longhornio/csi-attacher:v3.4.0"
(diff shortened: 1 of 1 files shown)


PhanLe1010 (approved):
LGTM

PR #1847, opened 2023-04-10
fix(expand): shrink volume is not allowed.

mantissahz:
Check if new volume size is smaller than old one and PVC size is expanded in the validator before expanding the volume.

Expansion is not allowed if storage class does not exist

Ref: longhorn/longhorn#5513

### DIFF
--- a/manager/volume.go
+++ b/manager/volume.go
@@ -9,7 +9,6 @@ import (
 	"github.com/sirupsen/logrus"
 
 	corev1 "k8s.io/api/core/v1"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
@@ -570,21 +569,6 @@ func (m *VolumeManager) checkAndExpandPVC(namespace string, pvcName string, size
 	pvc, err := m.ds.GetPersistentVolumeClaim(namespace, pvcName)
 	if err != nil {
 		return false, -1, err
-	}
-
-	longhornStaticStorageClass, err := m.ds.GetSettingValueExisted(types.SettingNameDefaultLonghornStaticStorageClass)
-	if err != nil {
-		return false, -1, err
-	}
-
-	pvcSCName := *pvc.Spec.StorageClassName
-	if pvcSCName == longhornStaticStorageClass {
-		if _, err := m.ds.GetStorageClassRO(pvcSCName); err != nil {
-			if !apierrors.IsNotFound(err) {
-				return false, -1, err
-			}
-			return false, size, nil
-		}
 	}
 
 	// TODO: Should check for pvc.Spec.Resources.Requests.Storage() here, once upgrade API to v0.18.x.
--- a/webhook/resources/volume/validator.go
+++ b/webhook/resources/volume/validator.go
@@ -2,9 +2,12 @@ package volume
 
 import (
 	"fmt"
+	"strconv"
 
 	"github.com/pkg/errors"
 	admissionregv1 "k8s.io/api/admissionregistration/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/runtime"
 
 	"github.com/longhorn/longhorn-manager/datastore"
@@ -110,6 +113,10 @@ func (v *volumeValidator) Update(request *admission.Request, oldObj runtime.Obje
 	oldVolume := oldObj.(*longhorn.Volume)
 	newVolume := newObj.(*longhorn.Volume)
 
+	if err := v.validateExpansionSize(oldVolume, newVolume); err != nil {
+		return werror.NewInvalidError(err.Error(), "")
+	}
+
 	if err := validateDataLocalityUpdate(oldVolume, newVolume); err != nil {
 		return werror.NewInvalidError(err.Error(), "")
 	}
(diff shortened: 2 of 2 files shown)


webhook/resources/volume/validator.go
```diff
@@ -156,6 +164,55 @@ func (v *volumeValidator) Update(request *admission.Request, oldObj runtime.Obje
...
+		return fmt.Errorf("cannot get request storage")
+	}
+
+	requestedSize := resource.MustParse(strconv.FormatInt(newSize, 10))
+	if pvcSpecValue.Cmp(requestedSize) < 0 {
+		return fmt.Errorf("waiting for pvc %v expansion from %v to %v", pvcName, pvcSpecValue.Value(), requestedSize.Value())
+	}
+	if pvcSpecValue.Cmp(requestedSize) > 0 && !newVolume.Status.ExpansionRequired {
+		return fmt.Errorf("shrinking pvc %v size from %v to %v is not allowed", pvcName, pvcSpecValue.Value(), requestedSize.Value())
+	}
```
shuo-wu, line 212: I am thinking if this is a little bit risky:
Somehow users expand the PVC spec size while the volume is handling other requests and it cannot start expansion immediately. Then the volume probably cannot continue handling anything since the update for other requests will be failed by this check.

BTW, IIRC, Kubernetes does not support shrink PVC spec size.

manager/volume.go
```diff
@@ -572,19 +572,16 @@ func (m *VolumeManager) checkAndExpandPVC(namespace string, pvcName string, size
...
-			}
-			return false, size, nil
+	pvcStorageClass, err := m.ds.GetStorageClassRO(pvcSCName)
+	if err != nil {
+		if !apierrors.IsNotFound(err) {
+			return false, -1, err
 		}
+		return false, size, err
+	}
+	if !*pvcStorageClass.AllowVolumeExpansion {
```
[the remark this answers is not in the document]
shuo-wu: Besides, is this a duplicate? You add this check in the validator now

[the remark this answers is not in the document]
mantissahz: Yes, you are right.
Added this check into the validator and Removed this part in the manager/volume.go.

webhook/resources/volume/validator.go
mantissahz: Yes, Kubernetes does not support shrink PVC spec size.
Removed this check statement. Thanks

shuo-wu (dismissed):

webhook/resources/volume/validator.go
```diff
@@ -156,6 +164,53 @@ func (v *volumeValidator) Update(request *admission.Request, oldObj runtime.Obje
...
 }
 
+func (v *volumeValidator) validateExpansionSize(oldVolume *longhorn.Volume, newVolume *longhorn.Volume) error {
+	oldSize := oldVolume.Spec.Size
+	newSize := newVolume.Spec.Size
+	if newSize == oldSize {
+		return nil
+	}
+	if newSize < oldSize && !newVolume.Status.ExpansionRequired {
+		return fmt.Errorf("shrinking volume size from %v/%v to %v/%v is not supported", oldVolume.Name, oldSize, newVolume.Name, newSize)
```
innobead, line 174: nit: volume name is the same, so the error message can be simplified.

webhook/resources/volume/validator.go
```diff
@@ -156,6 +164,53 @@ func (v *volumeValidator) Update(request *admission.Request, oldObj runtime.Obje
...
+	pvcSCName := *pvc.Spec.StorageClassName
+	pvcStorageClass, err := v.ds.GetStorageClassRO(pvcSCName)
+	if err != nil {
+		if !apierrors.IsNotFound(err) {
+			return errors.Wrapf(err, "cannot find storage class %v", pvcSCName)
+		}
+		return err
+	}
+	if pvcStorageClass.AllowVolumeExpansion != nil && !*pvcStorageClass.AllowVolumeExpansion {
+		return fmt.Errorf("storage class does not allow the volume expansion")
```
innobead, line 198: nit: should add volume info like name and storage class name in the error message like what you did above.

webhook/resources/volume/validator.go
```diff
@@ -156,6 +164,53 @@ func (v *volumeValidator) Update(request *admission.Request, oldObj runtime.Obje
...
+	}
+
+	pvcSpecValue, ok := pvc.Spec.Resources.Requests[corev1.ResourceStorage]
+	if !ok {
+		return fmt.Errorf("cannot get request storage")
```
innobead, line 203: nit: the same, need to add more context to the message.

innobead:
@mantissahz Please remember to resolve the review conversation if they are all resolved.

innobead (changes requested):
In genreal, LGTM.

However, could you improve the error message before merging this PR.

webhook/resources/volume/validator.go
mantissahz: Modified. just concerned about that old and new volume not the same one.
Thanks

webhook/resources/volume/validator.go
mantissahz: Added the pvcSCName and pvc.Name into this error log. Thanks

webhook/resources/volume/validator.go
mantissahz: Added `pvc.Name` into this error log. Thanks.

webhook/resources/volume/validator.go
```diff
@@ -156,6 +164,53 @@ func (v *volumeValidator) Update(request *admission.Request, oldObj runtime.Obje
...
 }
 
+func (v *volumeValidator) validateExpansionSize(oldVolume *longhorn.Volume, newVolume *longhorn.Volume) error {
+	oldSize := oldVolume.Spec.Size
+	newSize := newVolume.Spec.Size
+	if newSize == oldSize {
+		return nil
+	}
+	if newSize < oldSize && !newVolume.Status.ExpansionRequired {
+		return fmt.Errorf("shrinking volume size from %v to %v is not supported", oldSize, newSize)
```
innobead, line 174: Should have volume name in the message.

innobead (changes requested):
Just one left needed to improve.

innobead:
@Mergifyio backport v1.4.x

webhook/resources/volume/validator.go
mantissahz: Added volume name, thanks

webhook/resources/volume/validator.go
```diff
@@ -156,6 +164,53 @@ func (v *volumeValidator) Update(request *admission.Request, oldObj runtime.Obje
...
+	}
+
+	pvcSCName := *pvc.Spec.StorageClassName
+	pvcStorageClass, err := v.ds.GetStorageClassRO(pvcSCName)
+	if err != nil {
+		if !apierrors.IsNotFound(err) {
+			return errors.Wrapf(err, "cannot find storage class %v", pvcSCName)
+		}
+		return err
+	}
```
innobead, line 196: Just missed this..

This is not right. Just return err, good enough. 
```go
	if err != nil {
		return err
	}
```

Also, `!apierrors.IsNotFound(err)` means `resource found` instead..

webhook/resources/volume/validator.go
```diff
@@ -156,6 +164,53 @@ func (v *volumeValidator) Update(request *admission.Request, oldObj runtime.Obje
...
+
+	pvcSpecValue, ok := pvc.Spec.Resources.Requests[corev1.ResourceStorage]
+	if !ok {
+		return fmt.Errorf("cannot get request storage of pvc %v", pvc.Name)
+	}
+
+	requestedSize := resource.MustParse(strconv.FormatInt(newSize, 10))
+	if pvcSpecValue.Cmp(requestedSize) < 0 {
+		return fmt.Errorf("waiting for pvc %v expansion from %v to %v", pvcName, pvcSpecValue.Value(), requestedSize.Value())
+	}
```
innobead, line 209: Just revisited.

What does this mean actually? What cases will trigger this? updating volume directly instead of PVC?

The message should be `PVC %v can't be shrinked from %v to %v`. Isn't it?

innobead (changes requested):

webhook/resources/volume/validator.go
mantissahz: This means PVC size is smaller than requested size now and we need to wait for the PVC expansion first then we can update the volume size to requested size.
When users update volume size of `Volume CR` directly, this will be triggered.

webhook/resources/volume/validator.go
mantissahz: Removed.
Yes, you are right. Need to be more careful.
Many thanks.

webhook/resources/volume/validator.go
mantissahz: Modified the error log. The PVC size needs to be expanded first.

innobead (approved):
LGTM

PR #1848, opened 2023-04-11
feat: clean up all mount points when target is unset

ChanYiLin:
ref: [longhorn/longhorn#5655](https://github.com/longhorn/longhorn/issues/5655)

- for each manager on each node, execute `cleanup-all-mounts` when the backup store target is unset


controller/backup_target_controller.go
```diff
@@ -272,6 +272,15 @@ func (btc *BackupTargetController) reconcile(name string) (err error) {
 		return nil
 	}
 
+	log := getLoggerForBackupTarget(btc.logger, backupTarget)
+
+	if backupTarget.Spec.BackupTargetURL == "" {
+		if err := btc.cleanUpAllMounts(backupTarget); err != nil {
+			log.WithError(err).Error("Failed to clean up all mount points")
+			return err
+		}
```
innobead, line 281: Can you move this part to below code. Also, check `backupTarget.Status.Available` first to do the cleanup only if needed.

Also, don't return error to fail the reconciling, the cleanup operation should be best efforts instead. Log with warning log.

[/controller/backup_target_controller.go#L329-L347](https://github.com/longhorn/longhorn-manager/blob/d0b0a1222d40ebaf471361de3f24fee8c2b23a75/controller/backup_target_controller.go#L329-L347)
```go
	if backupTarget.Spec.BackupTargetURL == "" {
		backupTarget.Status.Available = false
		backupTarget.Status.Conditions = types.SetCondition(backupTarget.Status.Conditions,
			longhorn.BackupTargetConditionTypeUnavailable, longhorn.ConditionStatusTrue,
			longhorn.BackupTargetConditionReasonUnavailable, "backup target URL is empty")
		// Clean up all BackupVolume CRs
		if err := btc.cleanupBackupVolumes(); err != nil {
			log.WithError(err).Error("failed to clean up BackupVolumes")
			return err
		}

		// Clean up all SystemBackup CRs
		if err := btc.cleanupSystemBackups(); err != nil {
			log.WithError(err).Error("failed to clean up SystemBackups")
			return err
		}

		return nil
	}
```

innobead (changes requested):

controller/backup_target_controller.go
ChanYiLin: Hi @innobead the reason to put the code independently is because
all manager pods have chance to have backup target mounted not only the one who is responsible for it. 

So we have to do the cleanup before checking the ownerID, so all the controller will try to do the cleanup

Also, the cleanup will check if there is mount point, if not, it will return without error.

innobead (approved):

innobead:
@Mergifyio backport v1.4.x

PR #1852, opened 2023-04-15
Change mutating webhook's failure policy to Fail

derekbit:
Validating webhook's  failure policy is `Fail`. Change mutating webhook's failure policy to `Fail` as well.

https://github.com/longhorn/longhorn/issues/5762


innobead (approved):

innobead:
It seems the default is fail instead, but why we made that ignore, it didn't make sense.

Good finding @derekbit!

innobead:

derekbit:
I'm thinking if it is really an issue if using Fail in mutating webhook. If the Longhorn resources runs into issues caused by invalid values in the mutating webhook, stopping and rejecting the resources creation or update looks valid. Isn't it?

PhanLe1010:
> for mutating purpose, will this change impact other pods (example, a pod without PVC) , when the mutating webhook controller pod crashes or network problem, all other pods will be "stopped"

I am not sure why this Longhorn webhook can have affect on the pods don't use Longhorn PVC. These pod should not be affected by this webhook by any means. Could you elaborate more @panpan0000

PR #1857, opened 2023-04-19
chore(api): regenerated clientset

derekbit:


innobead (approved):
LGTM. 

@derekbit I am a little confused why the code needed to regenerate afterward. I thought the code should be always regenerated when resources have some change.

PR #1858, opened 2023-04-21
feat(csi): extend csi snapshot to support backing image

ChanYiLin:
ref: [longhorn/longhorn#5005](https://github.com/longhorn/longhorn/issues/5005)

LEP: https://github.com/longhorn/longhorn/pull/5771

Summary:
- Users can use VolumeSnapshot with type `bi` to create/delete BackingImage
- Users can restore a Volume with a VolumeSnapshot (create a Volume with BackingImage)
- Users can restore a Volume with an existing BackingImage (pre-provision)
    - Create the VolumeSnapshot/VolumeSnapshotContent to associate to the existing BackingImage
- Users can restore a Volume with non-existing BackingImage (on-demand provision)
    - Create the VolumeSnapshot/VolumeSnapshotContent to provide BackingImage parameters
    - Create BackingImage based on the parameters when the users create PVC from the VolumeSnapshot


csi/controller_server.go
```diff
@@ -578,6 +596,62 @@ func (cs *ControllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
 	return rsp, err
 }
 
+func (cs *ControllerServer) createCSISnapshotTypeLonghornBackingImage(req *csi.CreateSnapshotRequest) (*csi.CreateSnapshotResponse, error) {
+	csiSnapshotName := req.GetName()
+	csiVolumeName := req.GetSourceVolumeId()
+	if len(csiVolumeName) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "Volume name must be provided")
```
innobead, line 603: If creating backing image via snapshot, why volume name is always required? three cases are here.

- create a backing image from a volume
- create a backing image from different sources (like download)
- create a backing image from a staticaly created snapshot content

Here, we only cover the first case? or please clarify this.

innobead (changes requested):
Some comments and questions need to resolve.

csi/controller_server.go
ChanYiLin: csi `createSnapsot` will only be triggered when `VolumeSnapshot` source pointing to a volume
```
apiVersion: snapshot.storage.k8s.io/v1beta1
kind: VolumeSnapshot
metadata:
  name: test-snapshot-backing
spec:
  volumeSnapshotClassName: longhorn-snapshot-vsc
  source:
    persistentVolumeClaimName: test-vol
```
csi will then trigger the function to create snapshot with the given volume


On the other hand, if `VolumeSnapshot` source pointing to a `VolumeSnapshotContent`, csi won't trigger the `createSnapshot()` function
```
apiVersion: snapshot.storage.k8s.io/v1beta1
kind: VolumeSnapshot
metadata:
  name: test-snapshot-on-demand-backing
spec:
  volumeSnapshotClassName: longhorn-snapshot-vsc
  source:
    volumeSnapshotContentName: test-on-demand-backing
```

We can only wait until user create PVC and using the `VolumeSnapshot` and the csi trigger `createVolume()` then we can provision the BackingImage

csi/controller_server.go
```diff
@@ -282,20 +294,24 @@ func (cs *ControllerServer) checkAndPrepareBackingImage(volumeName, backingImage
...
+			return fmt.Errorf("volume %s backing image type %v is not supported via CSI", volumeName, bidsType)
+		case longhorn.BackingImageDataSourceTypeDownload:
+			if bidsParameters[longhorn.DataSourceTypeDownloadParameterURL] == "" {
+				return fmt.Errorf("volume %s missing parameters %v for preparing backing image",
+					volumeName, longhorn.DataSourceTypeDownloadParameterURL)
 			}
-			bidsType = string(longhorn.BackingImageDataSourceTypeDownload)
-			bidsParameters[longhorn.DataSourceTypeDownloadParameterURL] = volumeParameters["backingImageURL"]
+		case longhorn.BackingImageDataSourceTypeExportFromVolume:
+			if bidsParameters[longhorn.DataSourceTypeExportParameterExportType] == "" || bidsParameters[longhorn.DataSourceTypeExportParameterVolumeName] == "" {
```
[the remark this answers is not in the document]
ChanYiLin: Yes, this enables user to use storageclass to provision backingimage with `export-from-volume` type

ChanYiLin:
Since the modifications are related to original `snapshot`, `backup` volumesnapshot
let me run the e2e test to make sure they work fine

ChanYiLin:
e2e test for csi snapshotter passed: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/3714

innobead (changes requested):
Just a few minor comments needed to fix.

csi/controller_server.go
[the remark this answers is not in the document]
ChanYiLin: got it will update in this PR https://github.com/longhorn/website/pull/693

innobead (approved):
LGTM

csi/controller_server.go
```diff
@@ -780,39 +881,75 @@ func (cs *ControllerServer) DeleteSnapshot(ctx context.Context, req *csi.DeleteS
 	}
 
 	csiSnapshotType, sourceVolumeName, id := decodeSnapshotID(snapshotID)
-	if id == "" {
-		return nil, status.Errorf(codes.NotFound, "volume source snapshot %v is not found", snapshotID)
-	}
```
PhanLe1010, line 785: Could you please explain why we should delete this check?

ChanYiLin: the backing image has different snapshotID format
but we share the same decodeSnapshotID() function
- For BackingImage, the `sourceVolumeName`, `id` will be "", so we skip the check
- For Snapshot/Backup, we check in the below code https://github.com/longhorn/longhorn-manager/pull/1858/files#diff-4dbe40287acb148c5d0499c0adf977582db1e7b841f0d5d0943bd7e5e77b0cc8R890

PR #1865, opened 2023-04-25
Don't wait a minute after starting instance manager to update instance status

ejweber:
[longhorn/longhorn#5809](https://github.com/longhorn/longhorn/issues/5809)

Set updateNotification to true when starting the InstanceManagerMonitor so that it immediately attempts to update status.

Instance manager receives requests to create processes within a few seconds because its monitor starts reporting processes are not running almost immediately.

```
time="2023-04-25T17:15:19Z" level=info msg="Storing process logs at path: /var/log/instances"
[longhorn-instance-manager] time="2023-04-25T17:15:19Z" level=warning msg="Failed to add TLS key pair from /tls-files/" error="open /tls-files/tls.crt: no such file or directory"
[longhorn-instance-manager] time="2023-04-25T17:15:19Z" level=info msg="Creating gRPC server with no auth"
[longhorn-instance-manager] time="2023-04-25T17:15:19Z" level=info msg="Instance Manager proxy gRPC server listening to 0.0.0.0:8501"
[longhorn-instance-manager] time="2023-04-25T17:15:19Z" level=info msg="Instance Manager process gRPC server listening to 0.0.0.0:8500"
[longhorn-instance-manager] time="2023-04-25T17:15:20Z" level=info msg="Started new process manager update watch"
[longhorn-instance-manager] time="2023-04-25T17:15:20Z" level=info msg="Process Manager: prepare to delete process 1gb-r-8075cd99"
[longhorn-instance-manager] time="2023-04-25T17:15:20Z" level=info msg="Process Manager: prepare to delete process 2gb-r-fed8f2e5"
[longhorn-instance-manager] time="2023-04-25T17:15:20Z" level=info msg="Process Manager: prepare to delete process 3gb-r-e554a4c7"
[longhorn-instance-manager] time="2023-04-25T17:15:20Z" level=info msg="Process Manager: prepare to delete process 5gb-r-b11ba1f7"
[longhorn-instance-manager] time="2023-04-25T17:15:20Z" level=info msg="Process Manager: prepare to delete process 4gb-r-44abd906"
[longhorn-instance-manager] time="2023-04-25T17:15:20Z" level=info msg="Process Manager: prepare to delete process 5gb-r-b11ba1f7"
[longhorn-instance-manager] time="2023-04-25T17:15:20Z" level=info msg="Process Manager: prepare to delete process 1gb-r-8075cd99"
[longhorn-instance-manager] time="2023-04-25T17:15:20Z" level=info msg="Process Manager: prepare to delete process 3gb-r-e554a4c7"
[longhorn-instance-manager] time="2023-04-25T17:15:20Z" level=info msg="Process Manager: prepare to delete process 4gb-r-44abd906"
[longhorn-instance-manager] time="2023-04-25T17:15:20Z" level=info msg="Process Manager: prepare to delete process 2gb-r-fed8f2e5"
[longhorn-instance-manager] time="2023-04-25T17:15:20Z" level=info msg="Process Manager: prepare to delete process 4gb-r-44abd906"
[longhorn-instance-manager] time="2023-04-25T17:15:21Z" level=info msg="Process Manager: prepare to create process 1gb-r-8075cd99"
[longhorn-instance-manager] time="2023-04-25T17:15:21Z" level=info msg="Process Manager: prepare to create process 2gb-r-fed8f2e5"
```


controller/instance_manager_controller.go
```diff
@@ -1276,7 +1276,7 @@ func (imc *InstanceManagerController) startMonitoring(im *longhorn.InstanceManag
 		done:                   false,
 		monitorVoluntaryStopCh: monitorVoluntaryStopCh,
 		// notify monitor to update the instance map
-		updateNotification: false,
+		updateNotification: true,
```
ejweber, line 1279: I suspect this was always intended to be true, given the comment.

shuo-wu: Not sure why this was always intended to be true... When the system is stable, most of the time this should be false.

ejweber: What I mean to say is that it seems like it is always desirable to set this to true when starting the monitor (so we don't have to wait engineapi.MaxPollCount for the first invocation of pollAndUpdateInstanceMap). We usually don't actually wait that long, because some process creation event causes a change in process state and updateNotification flips to true on its own. However, in the situation observed in https://github.com/longhorn/longhorn/issues/5809, this is not the case, delaying stats updates.

I read the comment as meaning we want to set it to true on monitor startup. Certainly I agree that updateNotification should spend most of its life as false. Sorry for the confusion!

shuo-wu: Thank you for the clarification! Yes, this should be true during monitor startup.

innobead (approved):

innobead:

PR #1867, opened 2023-04-26
fix: fix create backup in engine image not fully deployed cluster

ChanYiLin:
ref: [longhorn/longhorn#5248](https://github.com/longhorn/longhorn/issues/5248)

The root cause of the issue is when the engine-image in the **deploying** state, none of the controllers can take responsibility because of `isResponsibleFor()` function

When selecting nodes with engine-image
we should use `ListReadyNodesWithEngineImage()` since it tolerates engine-image in deploying state
so the controller with the available engine-image can reconcile the `backup` and create `backupvolume`

This bug was introduced when fixing this issue 
- https://github.com/longhorn/longhorn/issues/5248: PR: https://github.com/longhorn/longhorn-manager/pull/1038

I have tested the test case, the target would be reconciled when the poll interval is 0 and the engine-image is redeployed.

cc @shuo-wu


controller/backup_controller.go
```diff
@@ -452,7 +452,7 @@ func (bc *BackupController) isResponsibleFor(b *longhorn.Backup, defaultEngineIm
 
 	isResponsible := isControllerResponsibleFor(bc.controllerID, bc.ds, b.Name, "", b.Status.OwnerID)
 
-	readyNodesWithReadyEI, err := bc.ds.ListReadyNodesWithReadyEngineImage(defaultEngineImage)
+	readyNodesWithReadyEI, err := bc.ds.ListReadyNodesWithEngineImage(defaultEngineImage)
```
innobead, line 455: We can remove `ListReadyNodesWithReadyEngineImage` if no ones use it.

Also, the below code seems unnecessary. I don't see anything it is trying to achieve. 

```go
	// No node in the system has the default engine image in ready state
	if len(readyNodesWithReadyEI) == 0 {
		return false, nil
	}
```

innobead (changes requested):

ChanYiLin:
e2e test pass: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/3596/

controller/backup_controller.go
ChanYiLin: Yes, you are right
I am just thinking why do we add this the first time
it is added in this PR: https://github.com/longhorn/longhorn-manager/pull/1038/files
but after review that PR, this seems unnecessary
cause the following code will check if the current node has ready image `currentNodeEngineAvailable`
and if `currentNodeEngineAvailable = False`, the function will always return False 
There is no need to check if "all nodes" have ready image or not
```
	currentOwnerEngineAvailable, err := bc.ds.CheckEngineImageReadiness(defaultEngineImage, b.Status.OwnerID)
	if err != nil {
		return false, err
	}
	currentNodeEngineAvailable, err := bc.ds.CheckEngineImageReadiness(defaultEngineImage, bc.controllerID)
	if err != nil {
		return false, err
	}

	isPreferredOwner := currentNodeEngineAvailable && isResponsible
	continueToBeOwner := currentNodeEngineAvailable && bc.controllerID == b.Status.OwnerID
	requiresNewOwner := currentNodeEngineAvailable && !currentOwnerEngineAvailable
        return isPreferredOwner || continueToBeOwner || requiresNewOwner, nil
```

cc @shuo-wu

ChanYiLin: Fixed, let me retest the previous issue https://github.com/longhorn/longhorn/issues/2960 (Fix backup target status.available always false) as well

ChanYiLin: both cases have passed locally
- cannot create backup when engine is not fully deployed
- target won't be reconcile

innobead (approved):
LGTM

innobead:

datastore/longhorn.go
```diff
@@ -2283,23 +2283,6 @@ func (s *DataStore) ListReadyNodesWithEngineImage(image string) (map[string]*lon
 	return readyNodes, nil
 }
 
-// ListReadyNodesWithReadyEngineImage returns list of ready nodes that have the corresponding engine image deployed
-func (s *DataStore) ListReadyNodesWithReadyEngineImage(image string) (map[string]*longhorn.Node, error) {
```
shuo-wu, line 2287: Good fix. This function does not make sense. It directly returns false once there are other nodes that do not contain the engine image and don't return the nodes that have the engine image. This contradicts the description and makes the function useless.

On the other hand, `bc.ds.CheckEngineImageReadiness` already did what we want.

PR #1873, opened 2023-04-28
Remove deprecated mkfs-ext4-parameters setting

ejweber:
https://github.com/longhorn/longhorn/issues/4914

Remove deprecated mkfs-ext4-parameters setting.


innobead:
@ejweber PR failed, need to check.

innobead (approved):
LGTM.

Please also update the doc (longhorn/website) and add an important note for this removal and reqeust usres to use the storage class way instead by pointing to the corresponding doc.

PR #1878, opened 2023-05-03
fix(backup): wait until restoration finished before deleting the backup volume

ChanYiLin:
Users have two operations
1. delete backup -> delete backup CR
2. delete all backup -> delete backup volume CR -> delete all backup

when deleting backup we check if it is used for restoration and wait until it finished [code](https://github.com/longhorn/longhorn-manager/blob/master/controller/backup_controller.go#L305)
but when `deleting all backup -> delete backup volume` we didn't check if there is any backup of the backup volume is used for restoration 
- so the volume in restoration will  stuck in progress because remote backupstore is deleted already
- and volume wii be in attached state and Ready for workload = Not Ready

we should wait until there is no any backup of the backup volume is used for restoration 

ref: [longhorn/longhorn#5506](https://github.com/longhorn/longhorn/issues/5506)

### DIFF
--- a/controller/backup_volume_controller.go
+++ b/controller/backup_volume_controller.go
@@ -219,6 +219,12 @@ func (bvc *BackupVolumeController) reconcile(backupVolumeName string) (err error
 
 	// Examine DeletionTimestamp to determine if object is under deletion
 	if !backupVolume.DeletionTimestamp.IsZero() {
+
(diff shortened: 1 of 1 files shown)


ChanYiLin:
e2e test passed: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/3691/
(error is not related to this PR it is because we remove the setting `mkfs-ext4-parameters` and the test is not updated yet)

innobead (approved):
LGTM

innobead:

PR #1891, opened 2023-05-09
feat(upgrade-checker): collect extra info

c3y1huang:
https://github.com/longhorn/longhorn/issues/5235


controller/setting_controller.go
```diff
@@ -1068,3 +1079,367 @@ func (sc *SettingController) cleanupFailedSupportBundles() error {
...
+		structFields:  util.StructFields{},
+		controllerID:  sc.controllerID,
+		namespace:     sc.namespace,
+	}
+
+	kubeVersion, err := sc.kubeClient.Discovery().ServerVersion()
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get Kubernetes server version")
+	}
+	extraInfo.structFields.Append(ExtraInfoKubernetesVersion, kubeVersion.GitVersion, extraInfo.structFields.ConvertTag(ExtraInfoKubernetesVersion))
```
shuo-wu, line 1098: NIT: Should not this be a part of `ClusterScope` info?

controller/setting_controller.go
```diff
@@ -1068,3 +1079,367 @@ func (sc *SettingController) cleanupFailedSupportBundles() error {
...
+func (info *ExtraInfo) collectHostOSDistro() error {
+	osDistro, err := util.GetHostOSDistro()
+	if err == nil {
+		info.structFields.Append(ExtraInfoHostOsDistro, osDistro, info.structFields.ConvertTag(ExtraInfoHostOsDistro))
+	}
+	return err
+}
+
+func (info *ExtraInfo) collectKubernetesNodeProvider() error {
+	node, err := info.kubeClient.CoreV1().Nodes().Get(context.TODO(), info.controllerID, metav1.GetOptions{})
```
shuo-wu, line 1417: NIT: Any special reason for not using cache?

controller/setting_controller.go
```diff
@@ -1068,3 +1079,367 @@ func (sc *SettingController) cleanupFailedSupportBundles() error {
...
+	allowCollectingUsage, err := sc.ds.GetSettingAsBool(types.SettingNameAllowCollectingLonghornUsage)
+	if err != nil {
+		return nil, err
+	}
+
+	if !allowCollectingUsage {
+		return extraInfo.structFields.NewStruct(), nil
+	}
+
+	extraInfo.collectNodeScope()
```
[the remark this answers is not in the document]
c3y1huang: This only collects information from the host node of the manager pod. To collect information from all nodes, each manager will send its respective node information to the influxDB.

controller/setting_controller.go
c3y1huang: While it could be included in the ClusterScope, I've separated it from other usage information so it won't be excluded by the `Allow Collecting Longhorn Usage Metrics` setting, considering we are already collecting the Kubernetes versions.

types/device.go
```diff
@@ -0,0 +1,112 @@
...
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/pkg/errors"
+
+	"github.com/longhorn/longhorn-manager/util"
+)
+
+func GetDeviceTypeOf(mountPath string) (string, error) {
```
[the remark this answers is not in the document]
c3y1huang: This is for the current disk implementation. Do we need to do anything for the SPDK now, or can it be done later?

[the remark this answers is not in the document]
innobead: For any SPDK part, can be improved in 1.6 (experimental). @c3y1huang go create an issue to follow up with any topic related to SPDK here if needed.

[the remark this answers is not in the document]
c3y1huang: https://github.com/longhorn/longhorn/issues/6033

util/os.go
```diff
@@ -0,0 +1,17 @@
+package util
+
+import (
+	"fmt"
+
+	iscsiutil "github.com/longhorn/go-iscsi-helper/util"
+)
+
+func GetHostKernelRelease() (string, error) {
```
innobead, line 17: Add doc.

util/util.go
```diff
@@ -1014,3 +1018,7 @@ func VerifySnapshotLabels(labels map[string]string) error {
 	}
 	return nil
 }
+
+func RemoveNewlines(input string) string {
+	return strings.Replace(input, "\n", "", -1)
```
innobead, line 1034: How about `strings.TrimSuffix` instead?

util/os.go
```diff
@@ -0,0 +1,43 @@
...
+	if err != nil {
+		return "", err
+	}
+	return RemoveNewlines(output), nil
+}
+
+func GetHostOSDistro() (string, error) {
+	initiatorNSPath := iscsiutil.GetHostNamespacePath(HostProcPath)
+	mountPath := fmt.Sprintf("--mount=%s/mnt", initiatorNSPath)
+	output, err := Execute([]string{}, "nsenter", mountPath, "cat", OsReleasePath)
```
innobead, line 30: Can make this reusable, since it is repeated to use in different host info collection.

util/os.go
```diff
@@ -0,0 +1,43 @@
...
+	initiatorNSPath := iscsiutil.GetHostNamespacePath(HostProcPath)
+	mountPath := fmt.Sprintf("--mount=%s/mnt", initiatorNSPath)
+	output, err := Execute([]string{}, "nsenter", mountPath, "uname", "-r")
+	if err != nil {
+		return "", err
+	}
+	return RemoveNewlines(output), nil
+}
+
+func GetHostOSDistro() (string, error) {
```
innobead, line 27: Suggest adding some unit tests if could.

util/os.go
```diff
@@ -0,0 +1,43 @@
...
+	}
+
+	scanner := bufio.NewScanner(strings.NewReader(string(output)))
+	for scanner.Scan() {
+		line := scanner.Text()
+		if strings.HasPrefix(line, "ID=") {
+			return RemoveNewlines(strings.TrimPrefix(line, "ID=")), nil
+		}
+	}
+	return "", fmt.Errorf("cannot find ID field in %v", OsReleasePath)
```
innobead, line 42: `failed to ID field...`

innobead (changes requested):

types/types.go
```diff
@@ -1001,3 +1001,15 @@ func IsSelectorsInTags(tags, selectors []string) bool {
 
 	return true
 }
+
+func GetKubernetesNodeProviderFrom(providerURL string) string {
```
innobead, line 999: Can we add cluster info related function to one file instead? This one seems being not used in other places. 

Also, can we change the name to `GetKubernetesProviderNameFromURL` instead?

https://github.com/longhorn/longhorn-manager/blob/b260399001bcf7bd40796eabf91889fbc17ff003/vendor/k8s.io/kubernetes/pkg/apis/core/types.go#L4378-L4378

controller/setting_controller.go
```diff
@@ -1062,3 +1071,151 @@ func (sc *SettingController) cleanupFailedSupportBundles() error {
...
+
+const (
+	ExtraInfoKubernetesVersion      = util.StructName("KubernetesVersion")
+	ExtraInfoKubernetesNodeProvider = util.StructName("KubernetesNodeProvider")
+
+	ExtraInfoHostKernelRelease = util.StructName("HostKernelRelease")
+	ExtraInfoHostOsDistro      = util.StructName("HostOsDistro")
+
+	ExtraInfoNamespaceUID = util.StructName("LonghornNamespaceUid")
+	ExtraInfoNodeCount    = util.StructName("LonghornNodeCount")
```
innobead, line 1124: Can we have a comment to differentiate which are cluster or node level info? I know you have collectNode and collectCluster, but it would be helpful to organize the const as well.


```go
// Cluster info 
...
// Node info
...
```

For example, longhorn node count is cluster level info.

controller/setting_controller.go
```diff
@@ -1062,3 +1072,212 @@ func (sc *SettingController) cleanupFailedSupportBundles() error {
...
+		}
+	}
+	info.structFields.AppendCounted(accessModeCountStruct)
+	info.structFields.AppendCounted(dataLocalityCountStruct)
+	info.structFields.AppendCounted(frontendCountStruct)
+
+	var avgVolumeSnapshotCount int
+	var avgVolumeSize int
+	if len(volumesRO) > 0 {
+		avgVolumeSize = totalVolumeSize / len(volumesRO)
```
innobead, line 1355: use a var for `len(volumesRO)`, since it's frequently reused here.

controller/setting_controller.go
```diff
@@ -1062,3 +1073,350 @@ func (sc *SettingController) cleanupFailedSupportBundles() error {
...
+	settingMap := make(map[string]string)
+	for _, setting := range settings {
+		settingName := types.SettingName(setting.Name)
+
+		switch {
+		case excludes[settingName]:
+			continue
+		case includeAsBoolean[settingName]:
+			settingMap[setting.Name] = fmt.Sprint(setting.Value != "")
+		case includeAsCustom[settingName] && settingName == types.SettingNameBackupTarget:
```
innobead, line 1292: Why not just `case settingName == types.SettingNameBackupTarget`? no need to strictly follow the group pattern (I know you try to group settings for different types)

controller/setting_controller.go
```diff
@@ -1062,3 +1073,350 @@ func (sc *SettingController) cleanupFailedSupportBundles() error {
...
+			settingMap[setting.Name] = types.GetBackupTargetSchemeFrom(setting.Value)
+		case includeAsCustom[settingName]:
+			logrus.Debugf("Found %v as invalid setting for custom extra info collection", settingName)
+		default:
+			settingMap[setting.Name] = setting.Value
+		}
```
innobead, line 1298: Ignore instead? Suggest whitelist to prevent collecting the info which are not acknowledged first.

types/types.go
```diff
@@ -1001,3 +995,29 @@ func IsSelectorsInTags(tags, selectors []string) bool {
...
+	}
+
+	scheme := util.GetSchemeFromURL(providerURL)
+	if scheme == "" {
+		scheme = ValueUnknown
+	}
+	return scheme
+}
+
+func GetBackupTargetSchemeFrom(backupTargetURL string) string {
```
innobead, line 1011: `GetbackupTargetSchemeFromURL`

innobead (changes requested):

util/struct.go
```diff
@@ -0,0 +1,63 @@
...
+type StructFields []StructField
+
+// StructField contains information about a single field in a struct.
+type StructField struct {
+	Name  string
+	Value string
+	Tag   string
+}
+
+func (sf *StructFields) Append(name StructName, value, tag string) {
```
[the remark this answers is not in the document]
c3y1huang: > Suggest making tag implicitly generated based on `name`? so the caller just need to provide `name` and `value`.

Updated.

> Make sure the tag is `camelCase` not just the first letter lower case. There should be some lib able to reuse directly.

I wasn't able to find a suitable library. However, since the name is defined by us so I think it should be ok for now, WDYT?

types/types.go
c3y1huang: > Can we add cluster info related function to one file instead? This one seems being not used in other places.
> 

To be honest, I am not sure where to move this to.
I think it will be better to reorganize the common packages as a whole.

https://github.com/longhorn/longhorn/issues/6039

> Also, can we change the name to `GetKubernetesProviderNameFromURL` instead?
> 

Updated.

innobead: Sounds good

util/struct.go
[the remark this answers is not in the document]
innobead: It's OK.

innobead (approved):

PhanLe1010:
I notice one quick issue with the current PR is that we cannot easily calculate the average (CPU/MEM) globally across all clusters.

Also, the graphs  [here](https://user-images.githubusercontent.com/22583541/237286543-59c07881-7255-4c6c-be6a-8067dbd6b76f.png)  will be very crowded (a.k.a there are a lot of lines) because each CPU value is on tag in the database and the CPU value range from 1 to 50000)

c3y1huang:
> I notice one quick issue with the current PR is that we cannot easily calculate the average (CPU/MEM) globally across all clusters. Harvester is having the same issue, they cannot easily calculate the total nodes from the node_count tag that each cluster leader sends.

Can you explain a bit more about what the challenge is?

> 
> Also, the graphs [here](https://user-images.githubusercontent.com/22583541/237286543-59c07881-7255-4c6c-be6a-8067dbd6b76f.png) will be very crowded (a.k.a there are a lot of lines) because each CPU value is one tag in the database and the CPU value range from 1 to 50000)
> 

Currently, we are showing in ranges (0-50, 50-100.. etc)
- CPU per 50 milli value
- Memory per 50 mib

PhanLe1010:
Hi @c3y1huang 

> Not sure where we need to calculate the total nodes from the node_count tag for the average CPU/memory. Can you explain a bit more about what the challenge is?

Harvester sends POST request only from the leader (per cluster), and they add node_counts as an extra field to the request body. When the data is written to InfluxDB, the node_count will be a tag of string type. The problem is node_count is a tag field in string, they had a hard time finding a way to sum it to find the total_node_count globally.


We will have the same problem for our cpu_* and memory_* metrics here when we want to find out the average usage across globally. What we are representing very well now is what CPU value is the most popular one (the highest line in the graph). What we are having an issue with is the simple query: what is the average CPU/Mem that a Longhorn cluster has.

> Currently, we are showing in ranges (0-50, 50-100.. etc)

This means that there could be 1000 lines if the max CPU value is 50000

What do you think?

innobead:
> This means that there could be 1000 lines if the max CPU value is 50000

Can we just make the range bigger, for example, 0-1000, 1000-2000? 
The average metric for CPU/Memory can be node level first.

BTW, another question is can influxdb do the aggregation, so we can just send the raw value instead? however, if yes, it would cause extra costs on the server side.

innobead:
After discussing with @c3y1huang , I would suggest the below change.

- Just send raw data w/o any local aggregation like CPU/Memory/Disk size range.
- Consider different visualization methods to effectively display numerical metrics, such as histograms or heatmaps

PhanLe1010:
Thanks @innobead for the update

> Can we just make the range bigger, for example, 0-1000, 1000-2000?

We still cannot answer the question:  what is the average CPU/Mem that a Longhorn cluster has

> The average metric for CPU/Memory can be node level first.

Your suggested approach is working well for CPU/MEM because these are per-node data. But we will have issues with other metrics such as number_of_volume_* 

> BTW, another question is can influxdb do the aggregation something like builtin functions by Prometheus, so we can just send the raw value instead, also tag int instead of string type? however, if yes, it would cause extra costs on the server side.

In influxQL, tag value is always string https://docs.influxdata.com/influxdb/v1.3/concepts/glossary/#tag-value and aggregation function such as MEAN(), MEDIAN() doesn't work on tag

**Proposal:**

From the discussion so far, I think we are hitting the limitation that comes with using [tag](https://docs.influxdata.com/influxdb/v1.3/concepts/glossary/#tag) for numerical values (tag must be string type and it will be indexed for faster lookup. I don't think we need to fast lookup which cluster has for example 70m CPU). My proposal is to use [field](https://docs.influxdata.com/influxdb/v1.3/concepts/glossary/#field) for these values. Field can be string, float, etc.. and we can easily calculate the MEAN(), MEDIAN(), TOTAL and draw the histogram, and pie charts. From there we can visualize a lot of things:
1. The histogram of the average CPU/MEM globally and TOTAL/MEAN() value for a typical cluster
1. The ratio of different volume types and total number of Longhorn volumes

IMG_1030

innobead:
@PhanLe1010 I like the proposal for especially we can better leverage the number data. 

The only thing we should think about is to make it extensible w/o changes. For example, if we want to send extra data in later versions, no need to change the upgrade responder. 

Go discuss with @c3y1huang about the proposal. 👍

PR #1893, opened 2023-05-09
fix(backupvolume): Ignore BackupVolume not found error in autodetachment

ChanYiLin:
ref: [longhorn/longhorn#5506](https://github.com/longhorn/longhorn/issues/5506)

There is a timing issue

- when user click Delete all Backups right before volume is attached and start restoration, the condition of volume restore is still false, and the BackupVolume can be deleted. In this case, we should block the situation by checking engine status instead, because its `isRestoring` will be **true** at the moment
- But if we only check engine status, then when engine restore progress 100%, it then turns into `isRestoring=False`. But the volume condition restore will still be true until it is **autodetached**. We should wait until condition `restore=False` then we can delete BackupVolume

### DIFF
--- a/controller/backup_volume_controller.go
+++ b/controller/backup_volume_controller.go
@@ -220,11 +220,6 @@ func (bvc *BackupVolumeController) reconcile(backupVolumeName string) (err error
 	// Examine DeletionTimestamp to determine if object is under deletion
 	if !backupVolume.DeletionTimestamp.IsZero() {
 
-		if unused, err := bvc.isBackupNotBeingUsedForVolumeRestore(backupVolumeName); !unused {
-			log.WithError(err).Warnf("Unable to delete backups of the backup volume")
-			return nil
-		}
-
 		if err := bvc.ds.DeleteAllBackupsForBackupVolume(backupVolumeName); err != nil {
 			log.WithError(err).Error("Error deleting backups")
 			return err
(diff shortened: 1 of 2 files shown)


innobead (commented):
If we don't wait for the volume to be detached but the backup is deleted, what will happen? 

When detaching the volume, if the backup does not exist, does it matter?

innobead (approved):
LGTM. Waiting for further testing.

ChanYiLin:
@innobead I have tested it locally, 
when the volume is restoring and in `in-progress`
the volume will restore successfully and detach as expected without error

I will demo it tomorrow.

ChanYiLin:
@innobead I have created a issue for adding the ownerReferece to Backup/BackupVolume: https://github.com/longhorn/longhorn/issues/5896

controller/volume_controller.go
```diff
@@ -2948,10 +2948,14 @@ func (vc *VolumeController) checkForAutoDetachment(v *longhorn.Volume, e *longho
 	if backupVolumeName, isExist := v.Labels[types.LonghornLabelBackupVolume]; isExist && backupVolumeName != "" {
 		backupVolume, err := vc.ds.GetBackupVolumeRO(backupVolumeName)
 		if err != nil {
-			return errors.Wrapf(err, "failed to get backup volume: %v", v.Name)
+			if !datastore.ErrorIsNotFound(err) {
+				return errors.Wrapf(err, "failed to get backup volume: %v", v.Name)
```
innobead, line 2952: Prevent nested if clause.

do this instead `if err != nil && !datastore.ErrorIsNotFound(err) {`

innobead (changes requested):

innobead (approved):

innobead:
@mergify backport v1.4.x

innobead:
@mergify backport v1.3.x

controller/backup_volume_controller.go
```diff
@@ -220,11 +220,6 @@ func (bvc *BackupVolumeController) reconcile(backupVolumeName string) (err error
 	// Examine DeletionTimestamp to determine if object is under deletion
 	if !backupVolume.DeletionTimestamp.IsZero() {
 
-		if unused, err := bvc.isBackupNotBeingUsedForVolumeRestore(backupVolumeName); !unused {
```
PhanLe1010, line 223: Could you please explain why we can remove this check but still ensure that the backup is not deleted while there is volume using it for restoration? @ChanYiLin

ChanYiLin: because below L#228
```
Delete the backup volume from the remote backup target
```
https://github.com/longhorn/longhorn-manager/pull/1893/files#diff-4967e3520dbda6bc373b2fea83e911dd65aac0ab4f48a53f977413620ce4ba7dR228

This will fail because of the lock of the file in the remote backup store
so even we don't have this check
the CR wouldn't be delete when the file is still used for restoration

ChanYiLin: because below L#228
```
Delete the backup volume from the remote backup target
```
https://github.com/longhorn/longhorn-manager/pull/1893/files#diff-4967e3520dbda6bc373b2fea83e911dd65aac0ab4f48a53f977413620ce4ba7dR228

This will fail because of the lock of the file in the remote backup store
so even we don't have this check
the CR wouldn't be delete when the file is still used for restoration

PR #1899, opened 2023-05-11
Remove disable-replica-rebuild setting

ejweber:
longhorn/longhorn#4997

This would look cleaner if we followed my suggestion in https://github.com/longhorn/longhorn/issues/5131#issuecomment-1544541401, but there may good reason not to do so. For now, I left a local copy of the setting in `upgrade/v120to121/upgrade.go` so the project will build.

### DIFF
--- a/types/setting.go
+++ b/types/setting.go
@@ -73,7 +73,6 @@ const (
 	SettingNameNodeDrainPolicy                                          = SettingName("node-drain-policy")
 	SettingNamePriorityClass                                            = SettingName("priority-class")
 	SettingNameDisableRevisionCounter                                   = SettingName("disable-revision-counter")
-	SettingNameDisableReplicaRebuild                                    = SettingName("disable-replica-rebuild")
 	SettingNameReplicaReplenishmentWaitInterval                         = SettingName("replica-replenishment-wait-interval")
 	SettingNameConcurrentReplicaRebuildPerNodeLimit                     = SettingName("concurrent-replica-rebuild-per-node-limit")
 	SettingNameConcurrentBackupRestorePerNodeLimit                      = SettingName("concurrent-volume-backup-restore-per-node-limit")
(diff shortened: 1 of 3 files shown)


innobead (changes requested):
We should remove unsupported paths from 1.5.0, @mantissahz will help with that. 

However, we also need to delete this setting after users upgrade to 1.5.0, so need to cover in 1.4.x -> 1.5 upgrade path. The same we need to do for mkfs-ext4, https://github.com/longhorn/longhorn/issues/4914.

ejweber:
> However, we also need to delete this setting after users upgrade to 1.5.0, so need to cover in 1.4.x -> 1.5 upgrade path. The same we need to do for mkfs-ext4, [longhorn/longhorn#4914](https://github.com/longhorn/longhorn/issues/4914).

I thought it was a little more maintainable to remove all settings that longhorn-manager no longer has definitions for. I tried the latest commit out on an upgrade from `v1.4.2` -> `master-head` and it removed the two settings we expect (`disable-replica-rebuild` and `mkfs-ext4-parameters`). Obviously the upgrade added and changed some settings as well.

```
eweber@laptop:~/longhorn> diff 1.4.2_before_upgrade master-head_after_upgrade 
9a10,11
> backup-compression-method                                         lz4
> backup-concurrent-limit                                           5
18,19c20,21
< current-longhorn-version                                          v1.4.2
< default-backing-image-manager-image                               longhornio/backing-image-manager:v1.4.2
---
> current-longhorn-version                                          v1.5.0-dev
> default-backing-image-manager-image                               longhornio/backing-image-manager:master-head
22,23c24,25
< default-engine-image                                              longhornio/longhorn-engine:v1.4.2
< default-instance-manager-image                                    longhornio/longhorn-instance-manager:v1.4.2
---
> default-engine-image                                              longhornio/longhorn-engine:master-head
> default-instance-manager-image                                    longhornio/longhorn-instance-manager:master-head
26c28
< default-share-manager-image                                       longhornio/longhorn-share-manager:v1.4.2
---
> default-share-manager-image                                       longhornio/longhorn-share-manager:master-head
28d29
< disable-replica-rebuild                                           false
35a37
> guaranteed-instance-manager-cpu                                   12
39d40
< mkfs-ext4-parameters
52a54
> restore-concurrent-limit                                          5
60a63
> storage-reserved-percentage-for-default-disk                      30
```

I can certainly do it instead in the `1.4.x` -> `1.5.0` upgrade path if you prefer. Copying @derekbit, since he most recently refactored the upgrading.

innobead (approved):
LGTM.

PR #1900, opened 2023-05-12
fix(csi): no need to check if volume is attached when creating backing image type snapshot

ChanYiLin:
ref: [longhorn/longhorn#5005](https://github.com/longhorn/longhorn/issues/5005)

Volume will be auto-attached for exporting the BackingImage

### DIFF
--- a/csi/controller_server.go
+++ b/csi/controller_server.go
@@ -614,9 +614,6 @@ func (cs *ControllerServer) createCSISnapshotTypeLonghornBackingImage(req *csi.C
 	if vol == nil {
 		return nil, status.Errorf(codes.NotFound, "volume %s not found", csiVolumeName)
 	}
-	if vol.State != string(longhorn.VolumeStateAttached) {
-		return nil, status.Errorf(codes.FailedPrecondition, "volume %s invalid state %v for taking snapshot. Volume must be in %v state to take snapshot", vol.Name, vol.State, longhorn.VolumeStateAttached)
-	}
 
 	var backingImage *longhornclient.BackingImage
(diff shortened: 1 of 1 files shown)


innobead (approved):
LGTM.

ref:
[/controller/volume_controller.go#L2845-L2866](https://github.com/longhorn/longhorn-manager/blob/784f5af8a2add1a54c204d69037ada5d59701593/controller/volume_controller.go#L2845-L2866)
```go
	// Do auto attachment for:
	//   1. restoring/DR volumes.
	//   2. expansion.
	//   3. Eviction requested on this volume.
	//   4. The target volume of a cloning
	//   5. The source volume of a cloning
	//   6, Export data as a backing image
	isRestoringDRVol := v.Status.RestoreRequired || v.Status.IsStandby
	isExpansionVol := v.Status.ExpansionRequired
	isEvictionRequestedOnVol := vc.hasReplicaEvictionRequested(rs)
	isTargetVolOfCloning := isTargetVolumeOfCloning(v)
	sourceVolumeOfCloning, err := vc.isSourceVolumeOfCloning(v)
	isExportingBackingImage := len(exportingBackingImageDataSources) != 0
	if err != nil {
		return err
	}
	if isRestoringDRVol || isExpansionVol || isEvictionRequestedOnVol ||
		isTargetVolOfCloning || sourceVolumeOfCloning || isExportingBackingImage {
		// Should use vc.controllerID or v.Status.OwnerID as CurrentNodeID,
		// otherwise they may be not equal
		v.Status.CurrentNodeID = v.Status.OwnerID
	}
```

PR #1902, opened 2023-05-15
feat(upgrade): remove dead codes for upgrade

mantissahz:
After introducing upgrade path enforcement from v1.5.0, we only upgrade from previous stable version, so it is not necessary to keep upgrade patch codes across two or more major/minor versions.
And engine.status.cliAPIVersion will not be 1 after we introduce the upgrade path enforcement so we remove the check statement codes of `isCLIAPIVersionOne`.

Ref: longhorn/longhorn#5131


ejweber (dismissed):
LGTM. Suggested the addition of one minor comment.

ejweber (commented):
Thanks for the comment addition!

With the discussion going on [here](https://github.com/longhorn/longhorn/issues/5131#issuecomment-1547910532), I'm not sure we should merge the `isCLIAPIVersionOne` commit. Are you confident we are safe to do so, @mantissahz?

ejweber (approved):
LGTM. Thanks for working through the churn!

innobead (approved):

PR #1903, opened 2023-05-15
fix(support-bundle): orphaned ready SupportBundle blocking new creation

c3y1huang:
ref: https://github.com/longhorn/longhorn/issues/5882


c3y1huang:
Test result: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/3826/

innobead (approved):
LGTM

innobead:
The current support bundle mechanism can't support concurrent operations like create, download, etc.

@c3y1huang Probably we need to mention this in the doc.

innobead:
@mergify backport v1.4.x

PR #1907, opened 2023-05-16
fix: log

c3y1huang:
Fix log


manager/volume.go
```diff
@@ -694,7 +694,7 @@ func (m *VolumeManager) AddVolumeRecurringJob(volumeName string, name string, is
 		if err != nil {
 			return nil, err
 		}
-		logrus.Debugf("Updated volume %v recurring jobs to %+v", v.Name, volumeRecurringJob)
+		logrus.Debugf("Updated volume %v recurring jobs to %+v", v.Name, v.Labels)
```
innobead, line 697: `Lables` does not just have recurring job related labels but also irrelevant labels.

I feel there is no need to have the log, but if you think it's helpful, probably you should add the log after `ListVolumeRecurringJob` then log with `volumeRecurringJob` instead?

innobead (changes requested):

manager/volume.go
c3y1huang: > `Lables` does not just have recurring job related labels but also irrelevant labels.

Update to use `key` instead.

> I feel there is no need to have the log, but if you think it's helpful, probably you should add the log after `ListVolumeRecurringJob` then log with `volumeRecurringJob` instead?

I prefer to have the log in its current place (after `v, err = s.UpdateVolume(v)`), as it can still show if `m.ListVolumeRecurringJob` returns error.

innobead (approved):
LGTM

PR #1908, opened 2023-05-17
fix: remove default image settings

mantissahz:
Remove the global default image settings during upgrade to v1.5.0 including `default-instance-manager-image`, `default-share-manager-image` and `default-backing-image-manager-image`.

Ref: longhorn/longhorn#5028

Regression Test:
- RWX: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/3873
- Backing image: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/3872

- Full run: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/3875

### DIFF
--- a/controller/backing_image_data_source_controller.go
+++ b/controller/backing_image_data_source_controller.go
@@ -51,6 +51,7 @@ type BackingImageDataSourceController struct {
 	namespace      string
 	controllerID   string
 	serviceAccount string
+	bimImageName   string
 
 	kubeClient    clientset.Interface
 	eventRecorder record.EventRecorder
@@ -84,7 +85,7 @@ func NewBackingImageDataSourceController(
 	ds *datastore.DataStore,
 	scheme *runtime.Scheme,
 	kubeClient clientset.Interface,
-	namespace, controllerID, serviceAccount string,
+	namespace, controllerID, serviceAccount, imageManagerImage string,
 	proxyConnCounter util.Counter,
 ) *BackingImageDataSourceController {
 
@@ -98,6 +99,7 @@ func NewBackingImageDataSourceController(
 		namespace:      namespace,
 		controllerID:   controllerID,
 		serviceAccount: serviceAccount,
+		bimImageName:   imageManagerImage,
 
 		kubeClient:    kubeClient,
 		eventRecorder: eventBroadcaster.NewRecorder(scheme, v1.EventSource{Component: "longhorn-backing-image-data-source-controller"}),
@@ -406,11 +408,6 @@ func (c *BackingImageDataSourceController) syncBackingImageDataSourcePod(bids *l
 	}()
 	log := getLoggerForBackingImageDataSource(c.logger, bids)
 
-	defaultImage, err := c.ds.GetSettingValueExisted(types.SettingNameDefaultBackingImageManagerImage)
-	if err != nil {
-		return err
-	}
-
 	newBackingImageDataSource := bids.Status.CurrentState == ""
 
 	podName := types.GetBackingImageDataSourcePodName(bids.Name)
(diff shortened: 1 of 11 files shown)


upgrade/v14xto150/upgrade.go
ejweber, line 1: This change (https://github.com/longhorn/longhorn-manager/pull/1899/commits/5984418caaabb344118a58752befd2dc84b2c5e8) was merged yesterday. It should handle cleaning up all settings that Longhorn no longer knows about after all upgrade paths have been walked, making this change unnecessary.

mantissahz: Thanks, that's great. I would removed the upgrade patch for the settings.

controller/backing_image_data_source_controller.go
```diff
@@ -427,8 +424,6 @@ func (c *BackingImageDataSourceController) syncBackingImageDataSourcePod(bids *l
 		podNotReadyMessage = fmt.Sprintf("pod spec node ID %v doesn't match the desired node ID %v", pod.Spec.NodeName, bids.Spec.NodeID)
 	} else if pod.DeletionTimestamp != nil {
 		podNotReadyMessage = "the pod dedicated to prepare the first backing image file is being deleted"
-	} else if pod.Spec.Containers[0].Image != defaultImage {
-		podNotReadyMessage = "the pod image is not the default one"
 	} else {
```
innobead, line 427: Should still keep the below code in `syncBackingImageDataSourcePod` to make sure the the existing BIDS pod handling correctly.

```go
	} else if pod.Spec.Containers[0].Image != c.bimImageName {
		podNotReadyMessage = "the pod image is not the default one"
	} 
```

manager/backingimage.go
```diff
@@ -67,20 +67,13 @@ func (m *VolumeManager) GetBackingImageDataSourcePod(name string) (*v1.Pod, erro
...
+func (m *VolumeManager) GetBackingImageManagersByDiskUUID(diskUUID string) (*longhorn.BackingImageManager, error) {
 	bims, err := m.ds.ListBackingImageManagersByDiskUUID(diskUUID)
 	if err != nil {
 		return nil, err
 	}
 	for _, bim := range bims {
-		if bim.Spec.Image == defaultImage {
-			return bim, nil
-		}
+		return bim, nil
```
innobead, line 76: Do we still need the below code? It think it's still required to prevent getting the old bim before it gets deleted.

```go
		if bim.Spec.Image == defaultImage {
			return bim, nil
		}
```

datastore/longhorn.go
```diff
@@ -225,6 +225,34 @@ func (s *DataStore) syncSettingCRsWithCustomizedDefaultSettings(customizedDefaul
 	return nil
 }
 
+func (s *DataStore) UpdateDefaultSettingConfigMapInstanceManagerImage(imImage string) error {
```
innobead, line 228: ConfigMap is used for setting purpose, so if the setting is removed, should not use the config map for the persistence.

I would suggest making the default instance manager setting as hidden instead of removing it to achieve the same goal w/o touching the setting config map.

innobead (changes requested):

innobead:
@mantissahz remember to run the related e2e test cases to ensure all working fine.

datastore/longhorn.go
mantissahz: @innobead
Do we have any example to make a setting as hidden? Just need a reference because I have no idea how to hide a Longhorn setting

innobead:
```go
		Type:     SettingTypeDeprecated,
```

This type is misleading, but it will actually be hidden and you can make it readyonly together.

manager/backingimage.go
mantissahz: Kept the check statement. Thanks

controller/backing_image_data_source_controller.go
mantissahz: Kept the check statement. Thanks

datastore/longhorn.go
mantissahz: Store the `backing image manager image` and `instance manager image` into the `Datastore` when initializing longhorn-manager daemon

innobead (changes requested):
@mantissahz I just noticed datastore is used by different apps, so the way here is not a good solution. Let's keep the original way to make the setting deprecated first, at least hidden from users from UI first. 

Tackle this in a later version (1.6) when developing a read-only setting.

mantissahz:
Modified to make the setting deprecated only.

innobead (approved):

PR #1909, opened 2023-05-17
feat(log): add global setting for log level

ChanYiLin:

add global setting for log level used in longhorn manager, default is warning

### DIFF
--- a/types/setting.go
+++ b/types/setting.go
@@ -104,6 +104,7 @@ const (
 	SettingNameBackupCompressionMethod                                  = SettingName("backup-compression-method")
 	SettingNameBackupConcurrentLimit                                    = SettingName("backup-concurrent-limit")
 	SettingNameRestoreConcurrentLimit                                   = SettingName("restore-concurrent-limit")
+	SettingNameLogLevel                                                 = SettingName("log-level")
 )
 
 var (
@@ -176,6 +177,7 @@ var (
 		SettingNameBackupCompressionMethod,
 		SettingNameBackupConcurrentLimit,
 		SettingNameRestoreConcurrentLimit,
+		SettingNameLogLevel,
 	}
 )
 
@@ -273,6 +275,7 @@ var (
 		SettingNameBackupCompressionMethod:                                  SettingDefinitionBackupCompressionMethod,
 		SettingNameBackupConcurrentLimit:                                    SettingDefinitionBackupConcurrentLimit,
 		SettingNameRestoreConcurrentLimit:                                   SettingDefinitionRestoreConcurrentLimit,
+		SettingNameLogLevel:                                                 SettingDefinitionLogLevel,
 	}
 
 	SettingDefinitionBackupTarget = SettingDefinition{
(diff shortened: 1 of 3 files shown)


ChanYiLin:
Screenshot from 2023-05-18 17-03-13

update log level by setting, there is no info msg after update log level to warning

types/setting.go
```diff
@@ -1112,6 +1115,16 @@ var (
 		ReadOnly:    false,
 		Default:     "5",
 	}
+
+	SettingDefinitionLogLevel = SettingDefinition{
+		DisplayName: "Log Level",
+		Description: "The log level 0 (Panic) ~ 6 (Trace) used in longhorn manager. By default 3 (Warning).",
```
shuo-wu, line 1121: Should the default level be `warning` rather than `info`?

shuo-wu (commented):
How do longhorn manager pods load this setting during startup?

types/setting.go
innobead: `info` is lower than `warning`, so `warning` will have less logs than info level.

innobead, line 1121: Let's use level constant instead of number. You can use `ParseLevel` to achieve that.

innobead (changes requested):
In general, LGTM. 

But, please use log string intead of log number to specify log level.

innobead:
> How do longhorn manager pods load this setting during startup?

@ChanYiLin I think we can do something during the start?

types/setting.go
shuo-wu: `info` logs contain some important info like object state change. Suppressing it by default would make the support bundle debugging more difficult.

innobead: @shuo-wu do you mean it will be more tedious? however, right now it's `debug` level which is even lower than `info`.

ChanYiLin: I agree with @shuo-wu 's comment
Most of the people use default setting
So if they generate support bundle when encountering some issue, there will be no info message in the log and make us more difficult to debug

ChanYiLin:
> > How do longhorn manager pods load this setting during startup?
> 
> @ChanYiLin I think we can do something during the start?

Longhorn-manager setup the setting at the very beginning
https://github.com/longhorn/longhorn-manager/blob/master/app/daemon.go#LL173C1-L173C27

It generates the setting CR and trigger the setting controller
https://github.com/longhorn/longhorn-manager/blob/master/datastore/longhorn.go#L74

There might have some debug level message before controller starts, but there is a PR to disable debug args
https://github.com/longhorn/longhorn/pull/5858

Plus we are using Info as default, so I think we don't need to do anything here

types/setting.go
ChanYiLin: Fixed, thanks

ChanYiLin: Change to default info level

innobead: I thought it was info. Yes we should use info.

types/setting.go
```diff
@@ -1112,6 +1115,16 @@ var (
...
 	}
+
+	SettingDefinitionLogLevel = SettingDefinition{
+		DisplayName: "Log Level",
+		Description: "The log level Panic, Fatal, Error, Warn, Info, Debug, Trace used in longhorn manager. By default Info.",
+		Category:    SettingCategoryGeneral,
+		Type:        SettingTypeString,
+		Required:    true,
+		ReadOnly:    false,
+		Default:     "Info",
```
innobead, line 1126: This is case insensitive.

innobead (approved):
LGTM. Remember to update longhorn/longhorn manifests.

types/setting.go
ChanYiLin: Yes, ParseLevel will lower the string

PR #1911, opened 2023-05-19
feat(recurring-job): set recurring job to pvc

c3y1huang:
https://github.com/longhorn/longhorn/discussions/5791

### DIFF
--- a/controller/recurring_job_controller.go
+++ b/controller/recurring_job_controller.go
@@ -217,7 +217,7 @@ func (c *RecurringJobController) syncRecurringJob(key string) (err error) {
 	}
 
 	if recurringJob.DeletionTimestamp != nil {
-		return c.cleanupVolumeRecurringJob(recurringJob)
+		return c.cleanupRecurringJobLabelInVolumesAndPVCs(recurringJob)
 	}
 
 	existingRecurringJob := recurringJob.DeepCopy()
@@ -244,32 +244,21 @@ func (c *RecurringJobController) syncRecurringJob(key string) (err error) {
 	return nil
 }
 
-func (c *RecurringJobController) cleanupVolumeRecurringJob(recurringJob *longhorn.RecurringJob) error {
+func (c *RecurringJobController) cleanupRecurringJobLabelInVolumesAndPVCs(recurringJob *longhorn.RecurringJob) (err error) {
+	defer func() {
+		err = errors.Wrapf(err, "failed to cleanup recurring job label %v in Volumes and PVCs", recurringJob.Name)
+	}()
+
 	// Check if each group of the recurring job contains other recurring jobs.
 	// If No, it means the recurring job is the last job of the group then
 	// Longhorn will clean up this group labels for all volumes.
-	checkRecurringJobs, err := c.ds.ListRecurringJobs()
+	recurringJobs, err := c.ds.ListRecurringJobs()
 	if err != nil {
 		return err
 	}
-	rmGroups := []string{}
-	for _, group := range recurringJob.Spec.Groups {
-		inUse := false
-		for _, checkRecurringJob := range checkRecurringJobs {
-			if checkRecurringJob.Name == recurringJob.Name {
-				continue
-			}
-			if util.Contains(checkRecurringJob.Spec.Groups, group) {
-				inUse = true
-				break
-			}
-		}
-		if !inUse {
-			rmGroups = append(rmGroups, group)
-		}
-	}
+	unusedGroups := c.getUnusedRecurringJobGroupInfo(recurringJob, recurringJobs)
 
-	// delete volume labels
+	// delete PVC or Volume labels
 	volumes, err := c.ds.ListVolumes()
 	if err != nil {
 		return err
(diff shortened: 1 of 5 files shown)


c3y1huang:
Test result: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/3860/

c3y1huang:
[Test PVC recurring job sync from Volume after upgrade](https://github.com/longhorn/longhorn-tests/blob/2d0aa0d45e0c2ea3810f2f3e95e5df0d65361f03/docs/content/manual/release-specific/v1.5.0/test-pvc-recurring-job-sync-after-upgrade.md): PASS

innobead (commented):
Can we make the implementation simpler as below w/ fewer resources (w/o a new PVC controller), WDYT?

1. in volume_controller, when resync, check if the volume has PVC and it has recurring jobs labels
2. If yes, override the volume labels, because PVC labels have higher priority.
3. If no PVC, all things are the same as usual

Just need to define the priority for which resource will decide the labels, no need to do sync efforts which will require another PVC controller in the implementation and seem not really necessary cost. Still, we can achieve the case for gitops use case.

c3y1huang:
Test result: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/3883/

datastore/longhorn.go
```diff
@@ -724,27 +724,28 @@ func MarshalLabelToVolumeRecurringJob(labels map[string]string) map[string]*long
...
+		}
+		logrus.Debugf("Added volume %v recurring job label %v", volume.Name, labelKey)
 	}
-	if _, exist := v.Labels[key]; exist {
-		delete(v.Labels, key)
-		v, err = s.UpdateVolume(v)
+	return nil
+}
+
+func (s *DataStore) DeleteRecurringJobLabelOn(volume *longhorn.Volume, labelKey string) (*longhorn.Volume, error) {
```
innobead, line 739: `DeleteRecurringJobLabelFromVolume`

types/types.go
```diff
@@ -500,6 +500,13 @@ func GetVolumeLabels(volumeName string) map[string]string {
 	}
 }
 
+func GetRecurringJobLabelKeyBy(name string, isGroup bool) string {
```
innobead, line 504: `GetRecurringJobLabelKeyByType`

controller/recurring_job_controller.go
```diff
@@ -497,3 +500,52 @@ func (control *RecurringJobController) newCronJob(recurringJob *longhorn.Recurri
 
 	return cronJob, nil
 }
+
+func syncRecurringJobLabelsTo(targetKind string, targetObj, sourceObj runtime.Object, log logrus.FieldLogger) error {
```
innobead, line 538: `syncRecurringJobLabelsFromSourceResource`

controller/volume_controller.go
```diff
@@ -3455,22 +3455,49 @@ func (vc *VolumeController) updateRecurringJobs(v *longhorn.Volume) (err error)
...
 	if err == nil && !reflect.DeepEqual(existingVolume.Labels, v.Labels) {
 		_, err = vc.ds.UpdateVolume(v)
 	}
 
 	return nil
 }
 
+func (vc *VolumeController) syncPVCRecurringJobLabels(volume *longhorn.Volume) error {
+	kubeStatus := volume.Status.KubernetesStatus
+	if kubeStatus.PVCName == "" {
```
innobead, line 3483: Do we need to check if the PVC still bound to the volume via `LastPVCRefAt`?

types/types.go
```diff
@@ -511,6 +518,12 @@ func GetRecurringJobLabelValueMap(labelType, recurringJobName string) map[string
 	}
 }
 
+func IsRecurringJobLabel(key string) bool {
```
innobead, line 528: Add doc.

manager/volume.go
```diff
@@ -717,11 +708,21 @@ func (m *VolumeManager) ListVolumeRecurringJob(volumeName string) (map[string]*l
...
 		return nil, err
 	}
-	return m.ds.DeleteVolumeRecurringJob(name, isGroup, volume)
+
+	key := types.GetRecurringJobLabelKeyBy(name, isGroup)
+	volume, err = m.ds.DeleteRecurringJobLabelOn(volume, key)
+	if err != nil {
+		return nil, err
+	}
+	return m.ds.UpdateVolume(volume)
```
innobead, line 725: Why not just updating the volume in datastore's `DeleteRecurringJobLabelOn`?

controller/recurring_job_controller.go
```diff
@@ -275,20 +261,71 @@ func (control *RecurringJobController) cleanupVolumeRecurringJob(recurringJob *l
...
+func (control *RecurringJobController) getIsLastGroupsExistsIn(recurringJob *longhorn.RecurringJob, existingRecurringJobs map[string]*longhorn.RecurringJob) map[string]bool {
+	groups := make(map[string]bool, len(recurringJob.Spec.Groups))
+	for _, group := range recurringJob.Spec.Groups {
+		groups[group] = true
+		for _, existingRecurringJob := range existingRecurringJobs {
+			if existingRecurringJob.Name == recurringJob.Name {
+				continue
+			}
+			if util.Contains(existingRecurringJob.Spec.Groups, group) {
+				groups[group] = false
```
innobead, line 305: break?

c3y1huang:
Test result: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/3911/

controller/recurring_job_controller.go
```diff
@@ -226,116 +229,171 @@ func (control *RecurringJobController) syncRecurringJob(key string) (err error)
...
 		log.WithError(err).Error("failed to reconcile recurring job")
 	}
 
 	return nil
 }
 
-func (control *RecurringJobController) cleanupVolumeRecurringJob(recurringJob *longhorn.RecurringJob) error {
+func (c *RecurringJobController) cleanupRecurringJobLabelFromVolumeAndPVC(recurringJob *longhorn.RecurringJob) (err error) {
+	defer func() {
+		err = errors.Wrap(err, "failed to cleanup recurring job labels from Volumes and PVCs")
```
innobead, line 249: `"failed to cleanup recurring job labels %v from Volumes and PVCs"`

controller/recurring_job_controller.go
```diff
@@ -226,116 +229,171 @@ func (control *RecurringJobController) syncRecurringJob(key string) (err error)
...
+	if err != nil {
+		return err
+	}
+
+	defer func() {
+		err = errors.Wrapf(err, "failed to update recurring job labels from Volume %v to PVC %v", volume.Name, pvc.Name)
+	}()
+
+	if err := syncRecurringJobLabelsToTargetResource(types.KubernetesKindPersistentVolumeClaim, pvc, volume, c.logger); err != nil {
+		return errors.Wrap(err, "failed to sync recurring job labels to PVC")
```
innobead, line 345: `"failed to sync recurring job labels to PVC %v"`

controller/volume_controller.go
```diff
@@ -3448,51 +3450,88 @@ func (vc *VolumeController) shouldRestoreRecurringJobs(v *longhorn.Volume) bool
...
-func (vc *VolumeController) updateRecurringJobs(v *longhorn.Volume) (err error) {
+func (c *VolumeController) updateRecurringJobs(v *longhorn.Volume) (err error) {
 	defer func() {
 		err = errors.Wrapf(err, "failed to update recurring jobs for %v", v.Name)
 	}()
 
 	existingVolume := v.DeepCopy()
+
+	if err := c.syncPVCRecurringJobLabels(v); err != nil {
+		return errors.Wrap(err, "failed attempting to sync PVC recurring job labels")
```
innobead, line 3474: `failed attempting to sync PVC recurring job labels from volume %v`

controller/recurring_job_controller.go
```diff
@@ -226,116 +229,171 @@ func (control *RecurringJobController) syncRecurringJob(key string) (err error)
...
+			if util.Contains(existingRecurringJob.Spec.Groups, group) {
+				unusedGroupInfo[group] = false
+				break
 			}
 		}
 	}
+	return unusedGroupInfo
+}
+
+func (c *RecurringJobController) updatePVCRecurringJobLabelsFrom(volume *longhorn.Volume) (err error) {
```
innobead, line 329: Rename `updatePVCRecurringJobLabelsFromVolume`

innobead (changes requested):
Some comments need to resolve.

I forgot why we couldn't remove a group label from a volume if the group is being used. What does that matter?

c3y1huang:
> I forgot why we couldn't remove a group label from a volume if the group is being used. What does that matter?

When a RecurringJob CR gets deleted, we also cleans the labels in the Volumes label. So we need to ensure the group label has no-op (no recurring jobs).

innobead:
> > I forgot why we couldn't remove a group label from a volume if the group is being used. What does that matter?
> 
> When a RecurringJob CR gets deleted, we also cleans the labels in the Volumes label. So we need to ensure the group label has no-op (no recurring jobs).

I see, so it's about removing the unused group label from other volumes.

innobead (approved):

PR #1920, opened 2023-05-22
[PART-2] Attachment/Detachment Controller and Volume Controller Refactoring

PhanLe1010:
- [x] Implement interrupt mechanism for workload attachment tickets
- [x]  UI modification to work with new AD mechanism (Longhorn UI -> Longhorn API) (the GitHub ticket and logic is at https://github.com/longhorn/longhorn/issues/6004)
- [x] Fix unit test of AD and volume controller (This will be handled separately at https://github.com/longhorn/longhorn/issues/6005)
- [x] Consider how to upgrade from the previous Longhorn version which doesn't have VA resource yet 
- [x] Attachment creation and deletion functions are implemented in some controllers. We can probably make it reused by different controllers
- [x] TODO: hadndle cases in which NodeID is empty. Should we detach the volume from all nodes???
- [x] `func (m *VolumeManager) CreateSnapshotCR` is checking condition before creating the snapshot CR, perhaps this should be handle by webhook

longhorn/longhorn#3715

### DIFF
--- a/controller/volume_attachment_controller.go
+++ b/controller/volume_attachment_controller.go
@@ -310,7 +310,7 @@ func (vac *VolumeAttachmentController) handleVolumeMigrationStart(va *longhorn.V
 
 	hasCSIAttachmentTicket := false
 	for _, attachmentTicket := range va.Spec.AttachmentTickets {
-		if attachmentTicket.Type != longhorn.AttacherTypeCSIAttacher {
+		if attachmentTicket.Type != longhorn.AttacherTypeCSIAttacher && attachmentTicket.Type != longhorn.AttacherTypeLonghornUpgrader {
 			continue
 		}
 		// Found one csi attachmentTicket that is requesting volume to attach to the current node
@@ -324,7 +324,7 @@ func (vac *VolumeAttachmentController) handleVolumeMigrationStart(va *longhorn.V
 	}
 
 	for _, attachmentTicket := range va.Spec.AttachmentTickets {
-		if attachmentTicket.Type != longhorn.AttacherTypeCSIAttacher {
+		if attachmentTicket.Type != longhorn.AttacherTypeCSIAttacher && attachmentTicket.Type != longhorn.AttacherTypeLonghornUpgrader {
 			continue
 		}
 		// Found one csi attachmentTicket that is requesting volume to attach to a different node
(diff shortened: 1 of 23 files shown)


PhanLe1010:
With this PR, the behavior of the upgrade path is described at https://github.com/longhorn/longhorn/issues/3715#issuecomment-1562305097

There is some limitation for auto-attached volumes. What do you think? @innobead @shuo-wu @derekbit

PhanLe1010:
ping @innobead @shuo-wu @derekbit Could you help to review this one? It is urgent for rc1. Thank you very much

manager/volume.go
```diff
@@ -263,9 +263,7 @@ func (m *VolumeManager) Attach(name, nodeID string, disableFrontend bool, attach
 	return v, nil
 }
 
-// Detach will handle regular detachment as well as volume migration confirmation/rollback
-// if nodeID is not specified, the volume will be detached from all nodes
-func (m *VolumeManager) Detach(name, attachmentID string, forceDetach bool) (v *longhorn.Volume, err error) {
+func (m *VolumeManager) Detach(name, attachmentID, hostID string, forceDetach bool) (v *longhorn.Volume, err error) {
```
innobead, line 266: In general, please remember to add a doc comment for any exported interface/method/function.

innobead:
> The limitation is auto attached volumes are being detached then reattached after the upgrade. It is because the old auto attachment logic doesn't set volume.Spec.NodeID while the new volume controller considers volume.Spec.NodeID as the source of true (AKA it will detach the volume when it see volume.Spec.NodeID is empty). Then later on other controllers create attachment tickets which help to reattach the volumes. I think this limitation is acceptable and we should add a release note. WDYT?

I am fine with this, but we need to emphasize this behavior impact to prevent any wrong expectations from users. Need to add a doc to an important note related to the upgrade.

innobead (approved):
In general, LGTM.

Please fix the conflict first, then we can merge this.

manager/volume.go
PhanLe1010: Thank you. I added the correct comment

PhanLe1010:
> I am fine with this, but we need to emphasize this behavior impact to prevent any wrong expectations from users. Need to add a doc to an important note related to the upgrade.
What operations will be impacted? recurring jobs, one-time operations like volume snapshot, clone, etc?

Yes, I will add the note. 

Operation like:
1. Clone
2. Backing Image exporting 

will be failed and cannot be recovered because there is no auto-retry .

Operation like:
1. DR/Restoring volume 
2. Replica eviction 

Will be able to retry and finish

innobead (approved):

controller/volume_attachment_controller.go
```diff
@@ -469,12 +471,44 @@ func shouldDoDetach(va *longhorn.VolumeAttachment, vol *longhorn.Volume) bool {
...
 	}
-	return true
+
+	if len(currentAttachmentTickets) == 0 {
+		return true
+	}
+	if !hasUninterruptibleTicket(currentAttachmentTickets) && hasWorkloadTicket(attachmentTicketsOnOtherNodes) {
+		log.Debugf("Workload attachment ticket interrupted snapshot-controller/backup-controller attachment tickets")
+		return true
+	}
```
derekbit, line 488: @PhanLe1010 Why does `hasWorkloadTicket` only check  `attachmentTicketsOnOtherNodes` other than all tickets?

PR #1921, opened 2023-05-24
fix(replica-auto-balance): filter node-selector

c3y1huang:
https://github.com/longhorn/longhorn/issues/5971

### DIFF
--- a/types/types.go
+++ b/types/types.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"path/filepath"
 	"regexp"
+	"sort"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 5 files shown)


c3y1huang:
Test result: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/3884/

util/util.go
```diff
@@ -1014,3 +1014,19 @@ func VerifySnapshotLabels(labels map[string]string) error {
 	}
 	return nil
 }
+
+func IsTagsEqual(tags1, tags2 []string) bool {
```
innobead, line 1018: Can we have a doc comment for this function? Suggest taking any change to improve this basic coding practice. (also, unit tests will be good to have for this exporeted function)

innobead (approved):
In general, LGTM.

util/util.go
```diff
@@ -1014,3 +1014,19 @@ func VerifySnapshotLabels(labels map[string]string) error {
...
+	if !sort.StringsAreSorted(tags1) {
+		logrus.Warn("BUG: Tags are not sorted, sorting now")
```
innobead: Make this as a debug log. (we are changing the log level to info, so this kind of log can be in debug, so it will not be shown by default until the level is changed to debug.)

util/util.go
innobead: Please add a doc comment. This is only really testing the tags are equal, but determine if tags2 is subset of tags1 instead.

types/types.go
```diff
@@ -950,3 +951,22 @@ func ConsolidateInstanceManagers(instanceManagerMaps ...map[string]*longhorn.Ins
 func GetLHVolumeAttachmentNameFromVolumeName(volName string) string {
 	return volName
 }
+
+// IsSelectorsInTags checks if all the selectors are present in the tags slice.
+// It returns true if all selectors are found, false otherwise.
+func IsSelectorsInTags(tags, selectors []string) bool {
```
innobead, line 957: This is better.

innobead (approved):

c3y1huang:

PR #1922, opened 2023-05-26
fix(snapshot): live migration deletes snapshots

mantissahz:
New created engine for migration will have empty snapshot list in `Status`, and running `syncSnapshotCRs` will delete all snapshot CRs of this volume before engine monitor synchronize the snapshots into `Status.Snapshots`.

Ref: longhorn/longhorn#5992

### DIFF
--- a/controller/engine_controller.go
+++ b/controller/engine_controller.go
@@ -1148,6 +1148,16 @@ func (m *EngineMonitor) acquireRestoringCounter(acquire bool) error {
 }
 
 func (ec *EngineController) syncSnapshotCRs(engine *longhorn.Engine) error {
+	log := ec.logger.WithField("engine", engine.Name)
+
+	vol, err := ec.ds.GetVolumeRO(engine.Spec.VolumeName)
+	if err != nil {
+		return err
+	}
+	if isVolumeMigrating(vol) {
+		return nil
+	}
+
 	snapshotCRs, err := ec.ds.ListVolumeSnapshotsRO(engine.Spec.VolumeName)
 	if err != nil {
 		return err
@@ -1160,7 +1170,7 @@ func (ec *EngineController) syncSnapshotCRs(engine *longhorn.Engine) error {
 		requestCreateNewSnapshot := snapCR.Spec.CreateSnapshot
 		alreadyCreatedBefore := snapCR.Status.CreationTime != ""
 		if _, ok := engine.Status.Snapshots[snapName]; !ok && (!requestCreateNewSnapshot || alreadyCreatedBefore) && snapCR.DeletionTimestamp == nil {
-			ec.logger.Infof("Deleting snapshot CR for the snapshot %v", snapName)
+			log.Infof("Deleting snapshot CR for the snapshot %v", snapName)
 			if err := ec.ds.DeleteSnapshot(snapName); err != nil && !apierrors.IsNotFound(err) {
 				utilruntime.HandleError(fmt.Errorf("syncSnapshotCRs: failed to delete snapshot CR %v: %v", snapCR.Name, err))
 			}
@@ -1189,7 +1199,7 @@ func (ec *EngineController) syncSnapshotCRs(engine *longhorn.Engine) error {
 					CreateSnapshot: false,
 				},
 			}
-			ec.logger.Infof("Creating snapshot CR for the snapshot %v", snapName)
+			log.Infof("Creating snapshot CR for the snapshot %v", snapName)
 			if _, err := ec.ds.CreateSnapshot(snapCR); err != nil && !apierrors.IsAlreadyExists(err) {
 				utilruntime.HandleError(fmt.Errorf("syncSnapshotCRs: failed to create snapshot CR %v: %v", snapCR.Name, err))
 			}
--- a/controller/volume_attachment_controller.go
+++ b/controller/volume_attachment_controller.go
@@ -101,6 +101,16 @@ func (vac *VolumeAttachmentController) enqueueVolumeAttachment(obj interface{}) 
 	vac.queue.Add(key)
 }
 
+func (vac *VolumeAttachmentController) enqueueVolumeAttachmentAfter(obj interface{}, duration time.Duration) {
+	key, err := controller.KeyFunc(obj)
+	if err != nil {
+		utilruntime.HandleError(fmt.Errorf("enqueueVolumeAttachmentAfter: couldn't get key for object %#v: %v", obj, err))
+		return
+	}
+
+	vac.queue.AddAfter(key, duration)
+}
+
 func (vac *VolumeAttachmentController) enqueueForLonghornVolume(obj interface{}) {
 	vol, ok := obj.(*longhorn.Volume)
 	if !ok {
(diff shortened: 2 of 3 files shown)


controller/volume_controller.go
```diff
@@ -3198,6 +3198,11 @@ func (c *VolumeController) createEngine(v *longhorn.Volume, isNewEngine bool) (*
 
 	if isNewEngine {
 		engine.Spec.Active = true
+	} else {
+		engine.Status.Snapshots = *snapshots
```
shuo-wu, line 3202: 1. Will a CR `Create` API include the status info?
2. I prefer to somehow skip the snapshot syncing before an engine becomes the main/active one.

mantissahz: Yes, you are right, it is weird to have the status info when creating a CR.
Removed it, and it will still wait util the engine is running and engine monitor starts to synchronize snapshots.

controller/volume_controller.go
```diff
@@ -3198,6 +3198,10 @@ func (c *VolumeController) createEngine(v *longhorn.Volume, isNewEngine bool) (*
 
 	if isNewEngine {
 		engine.Spec.Active = true
+	} else {
+		if v.Spec.MigrationNodeID != "" {
+			engine.Spec.NodeID = v.Spec.MigrationNodeID
+		}
```
innobead, line 3204: Why need to do this? this has been handled in https://github.com/longhorn/longhorn-manager/blob/023dd3cbd5009f390608ea9cdd9f31bd397cf8a2/controller/volume_controller.go#L3888?

innobead (changes requested):

controller/engine_controller.go
```diff
@@ -1148,6 +1148,14 @@ func (m *EngineMonitor) acquireRestoringCounter(acquire bool) error {
 }
 
 func (ec *EngineController) syncSnapshotCRs(engine *longhorn.Engine) error {
+	log := ec.logger.WithField("engine", engine.Name)
+
+	// it should have the `volume-head` snapshot at least for synchronizing the snapshot list by the engine monitor
+	// new migration engine might have an empty snapshot list once its state changed to be active and running
+	if len(engine.Status.Snapshots) == 0 {
```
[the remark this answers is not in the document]
mantissahz: No.
When migration finished and the migration engine is running, it could still has an empty snapshot list here if the engine monitor does not finish synchronizing the snapshots.

controller/volume_controller.go
mantissahz: Because I found that updating engine spec for the migrationEngine will be failed sometimes and it would never run this part of codes again for some reasons such as `v.Spec.MigrationNodeID` had been set "" or old engine had been deleted.
Therefore I set the engine nodeID as v.Spec.MigrationNodeID when creating the migrating engine.

controller/engine_controller.go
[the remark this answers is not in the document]
innobead: Should we make the migration ready until the snapshot sync instead? probably something needed to update in the monitor? or at least to sync the snapshot list when determining whether the migration is really finished or not?

controller/volume_controller.go
shuo-wu: Wait... Is this engine the old one? Can we directly change the running old engine spec node ID?

shuo-wu (commented):
We can block the migration confirmation if the new/target engine does not hold the same snapshot map as the old engine.

controller/volume_controller.go
mantissahz: The migration engine is the new created one.

shuo-wu: Here you are operating `engine`

mantissahz: The migrationEngine is created here for `isNewEngine` is `false`

controller/engine_controller.go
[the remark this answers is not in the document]
mantissahz: Check if snapshots sync before volume attachment controller set `MigrationNodeID` ""

controller/volume_controller.go
mantissahz: Removed this assignment, it will have been set again when synchronizing the snapshots if there is a conflict for updating the migration engine.
Thanks.

mantissahz:
>We can block the migration confirmation if the new/target engine does not hold the same snapshot map as the old engine.

Modified. Check if snapshots sync before volume attachment controller set MigrationNodeID ""
Thanks.

controller/volume_attachment_controller.go
```diff
@@ -342,13 +352,64 @@ func (vac *VolumeAttachmentController) handleVolumeMigrationConfirmation(va *lon
...
+	if oldEngine == nil {
+		return false, fmt.Errorf("failed to find the active engine for volume %v", vol.Name)
+	}
+
+	if migratingEngine == nil {
+		return false, fmt.Errorf("failed to find the migrating engine for volume %v", vol.Name)
+	}
+
+	if !reflect.DeepEqual(oldEngine.Status.Snapshots, migratingEngine.Status.Snapshots) {
+		vac.logger.Debugf("Volume migration is in progress for synchronizing snapshots from old snapshot list %v to new %v", oldEngine.Status.Snapshots, migratingEngine.Status.Snapshots)
```
innobead, line 394: This log would be quite tedious if the number of snapshots is a lot. Suggest log like this. `Volume migration (%v) is in progress for synchronizing snapshots` with volume name should be good enough.

controller/volume_controller.go
```diff
@@ -3633,17 +3633,21 @@ func (c *VolumeController) processMigration(v *longhorn.Volume, es map[string]*l
...
-		}
-
 		// in the case of a confirmation we need to switch the v.Status.CurrentNodeID to v.Spec.NodeID
 		// so that currentEngine becomes the migration engine
 		if v.Status.CurrentNodeID != v.Spec.NodeID {
 			log.Infof("volume migration complete switching current node id from %v to %v", v.Status.CurrentNodeID, v.Spec.NodeID)
 			v.Status.CurrentNodeID = v.Spec.NodeID
 		}
 
+		if len(es) < 2 {
```
innobead, line 3643: Why moving this condition after the above `if v.Status.CurrentNodeID != v.Spec.NodeID {`? This is part of conditions to do any migration check, so it should be at the beginning, right?

Suggest just changing `if len(es) < 2 {` to below.

```go
if len(es) < 2 {
    if v.Status.CurrentMigrationNodeID == "" {
        return nil
	}
    
    return errors.New(....) // return an intended error to let processMigration rerun next time. 
}
```

innobead (changes requested):

controller/volume_controller.go
shuo-wu: if `v.Status.CurrentMigrationNodeID != ""` but somehow `len(es) < 2`, the new engine or the old engine may crash and Longhorn should be able to handle this rather than returning an error.

innobead: Exactly, so probably just continue the following logic to ensure https://github.com/longhorn/longhorn-manager/blob/023dd3cbd5009f390608ea9cdd9f31bd397cf8a2/controller/volume_controller.go#L3677 is reached eventually?

controller/volume_attachment_controller.go
mantissahz: Modified. Thanks.

controller/volume_controller.go
mantissahz: Modified 
`if len(es) < 2 {` to be `if len(es) < 2 && v.Status.CurrentMigrationNodeID == "" {`
to handle old engine is deleted then volume status updated failed or new engine or the old engine may crash.
The following code flow could update the correct volume status information again.
https://github.com/longhorn/longhorn-manager/blob/023dd3cbd5009f390608ea9cdd9f31bd397cf8a2/controller/volume_controller.go#L3677  Yes, it could be reached eventually if the engine in the es (engine list) is the new one.

innobead:

PR #1923, opened 2023-05-26
fix(logging): review and adjust log message and level in backing image and backup controllers

ChanYiLin:
ref: [longhorn/longhorn 5888](https://github.com/longhorn/longhorn/issues/5888)

Main Idea:
- use `Failed to + V` instead of `Error Ving`
- when doing operation, log before with `.Info("Ving + ...")` and no need to log after operation is successful since if there is any error, error will be returned and reported in upper function.
- ~~In monitor goroutine, use `Debug` instead of `Warn` and `Info`.~~
- In monitor goroutine, replace trace `Info` with `Debug`

UPDATE:
- if an error will be consumed w/o being returned back to the caller, using warning, since it has been handled.
- If an error will be returned/bubbled up, suggest no extra logging but can be enriched with another error wrapper to prevent a duplicated error log. This needs to check if there would be a chance to have duplicated error message.


ChanYiLin:
Hi @innobead 
Lets do the refactor 1~2 controllers at a time
So the scope can be smaller and we can adjust the standard during the process more easily.
Thanks!

controller/backing_image_data_source_controller.go
```diff
@@ -1033,14 +1031,11 @@ func (c *BackingImageDataSourceController) stopMonitoring(bidsName string) {
 	log := c.logger.WithField("backingImageDataSource", bidsName)
 	stopCh, ok := c.monitorMap[bidsName]
 	if !ok {
-		log.Warn("No monitor goroutine for stopping")
 		return
 	}
-	log.Info("Stopping monitoring")
+	log.Infof("Stopping monitoring %v", bidsName)
```
shuo-wu, line 1036: This `log` is `c.logger.WithField("backingImageDataSource", bidsName)`

controller/backing_image_data_source_controller.go
```diff
@@ -1071,7 +1065,7 @@ func (c *BackingImageDataSourceController) startMonitoring(bids *longhorn.Backin
 	}
 	c.monitorMap[bids.Name] = stopCh
 
-	log.Info("Starting monitoring")
+	log.Infof("Starting monitoring backing image data source %v", bids.Name)
```
shuo-wu, line 1068: Similarly, no need to add `bids.Name` here

controller/backing_image_data_source_controller.go
```diff
@@ -1088,7 +1082,7 @@ func (m *BackingImageDataSourceMonitor) sync() {
 			m.retryCount++
 			if m.retryCount == engineapi.MaxMonitorRetryCount {
 				m.stopCh <- struct{}{}
-				m.log.Warnf("Stopped monitoring since monitor %v sync reaches the max retry count %v", m.Name, engineapi.MaxMonitorRetryCount)
+				m.log.Debugf("Stopped monitoring since monitor %v sync reaches the max retry count %v", m.Name, engineapi.MaxMonitorRetryCount)
```
shuo-wu, line 1085: Should this be considered as a normal case?

controller/backing_image_data_source_controller.go
```diff
@@ -1110,11 +1104,11 @@ func (m *BackingImageDataSourceMonitor) sync() {
 	}
 	if bids.Status.OwnerID != m.controllerID {
 		m.stopCh <- struct{}{}
-		m.log.Warnf("Stopped monitoring since backing image data source %v owner %v is not the same as monitor current controller %v", m.Name, bids.Status.OwnerID, m.controllerID)
+		m.log.Debugf("Stopped monitoring since backing image data source %v owner %v is not the same as monitor current controller %v", m.Name, bids.Status.OwnerID, m.controllerID)
 		return
 	}
 	if bids.Status.IP == "" {
-		m.log.Warnf("Stopped monitoring since backing image data source %v current IP is empty", m.Name)
+		m.log.Debugf("Stopped monitoring since backing image data source %v current IP is empty", m.Name)
```
shuo-wu, line 1111: Prefer to keep log level `warn` since these 2 cases are not normal ones.

controller/backing_image_manager_controller.go
```diff
@@ -698,25 +696,23 @@ func (c *BackingImageManagerController) prepareBackingImageFiles(currentBIM *lon
 				continue
 			}
 			if bids.Status.StorageIP == "" {
-				log.Warnf("Failed to get backing image data source %v storage IP, cannot start transfer the file to the backing image manager", bids.Name)
+				log.Debugf("Failed to get backing image data source %v storage IP, cannot start transfer the file to the backing image manager", bids.Name)
```
shuo-wu, line 699: Same here. Prefer to keep log level `warn`

controller/backup_controller.go
```diff
@@ -282,18 +282,18 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
...
 			backupTargetClient, err := newBackupTargetClientFromDefaultEngineImage(bc.ds, backupTarget)
 			if err != nil {
-				log.WithError(err).Error("Error init backup target clients")
+				log.WithError(err).Warn("Failed to init backup target clients")
 				return nil // Ignore error to prevent enqueue
 			}
 
 			if unused, err := bc.isBackupNotBeingUsedForVolumeRestore(backup.Name, backupVolumeName); !unused {
-				log.WithError(err).Warnf("Unable to delete remote backup")
+				log.WithError(err).Warn("Unable to delete remote backup")
```
shuo-wu, line 290: `Failed to check volume restore usage of the backup`

controller/backup_controller.go
```diff
@@ -777,34 +777,36 @@ func (bc *BackupController) disableBackupMonitor(backupName string) {
 }
 
 func (bc *BackupController) syncBackupStatusWithSnapshotCreationTimeAndVolumeSize(volume *longhorn.Volume, backup *longhorn.Backup) {
+	bc.logger.Info("Syncing backup status with snapshot creation time and volume size.")
```
shuo-wu, line 780: Do we need this?
If users try to a backupstore contains thousands of backup to a new cluster, this message will flood the log.
You can add a defer function to add a prefix for the error messages.

controller/backing_image_manager_controller.go
```diff
@@ -1151,9 +1138,7 @@ func (c *BackingImageManagerController) isMonitoring(bimName string) bool {
 }
 
 func (m *BackingImageManagerMonitor) Run() {
-	m.log.Info("Starting monitoring")
```
shuo-wu, line 1154: There is no need to remove the log or change the log level for bids monitor starting/stopping since you are still using info log for the bim monitor

controller/backup_controller.go
ChanYiLin: The function return `unused == false`
Not only when the backup is used but if there is any error during checking
So the message here needs to keep general

controller/backup_controller.go
ChanYiLin: You are right I will add this to every error message instead

controller/backing_image_manager_controller.go
ChanYiLin: This change here is to move this info log outside of goroutine 
it is moved to here: https://github.com/longhorn/longhorn-manager/pull/1923/files#diff-50cf5d5b10c83f46186aef3f483f4c18054476a8370822b5f8f3ac692f64b66fR1104

shuo-wu (approved):

innobead (changes requested):
In genreal, LGTM. Two primary things need to check before merging.

1. if the error will be consumped w/o being returned back to the caller, suggest using `warning`.
2. If the error will be returned/bubbled up/backup, suggest no extra logging but can be enriched with another error wrapper to prevent a duplicated error log. Please help confirm this.

ChanYiLin:
> In general, LGTM. Good job.
> 
> Two primary things need to check before merging.
> 
> 1. if an error will be consumed w/o being returned back to the caller, suggest using warning, since it has been handled. WDYT?
> 2. If an error will be returned/bubbled up, suggest no extra logging but can be enriched with another error wrapper to prevent a duplicated error log. This needs to check if there would be a chance to have duplicated error message.

Sounds good to me!

controller/backing_image_controller.go
```diff
@@ -506,18 +505,18 @@ func (bic *BackingImageController) handleBackingImageDataSource(bi *longhorn.Bac
 		fileStatus, exists := bi.Status.DiskFileStatusMap[bids.Spec.DiskUUID]
 		if exists && fileStatus.State == longhorn.BackingImageStateReady {
 			bids.Spec.FileTransferred = true
-			log.Infof("Default backing image manager successfully took over the file prepared by the backing image data source, will mark the data source as file transferred")
+			log.Infof("Default backing image manager successfully took over the file, will mark the data source as file transferred")
```
derekbit, line 508:
```suggestion
			log.Info("Default backing image manager successfully took over the file, will mark the data source as file transferred")
```

controller/backing_image_manager_controller.go
```diff
@@ -1229,15 +1214,15 @@ func (m *BackingImageManagerMonitor) pollAndUpdateBackingImageFileMap() (needSto
 	bim, err := m.ds.GetBackingImageManager(m.Name)
 	if err != nil {
 		if apierrors.IsNotFound(err) {
-			m.log.Info("Stopping monitoring because the backing image manager no longer exists")
+			m.log.Debug("Stopping monitoring because the backing image manager no longer exists")
```
derekbit, line 1217: Will it only show once? If yes, using `Info` is good to know the image does not exist anymore.

controller/backing_image_manager_controller.go
```diff
@@ -1229,15 +1214,15 @@ func (m *BackingImageManagerMonitor) pollAndUpdateBackingImageFileMap() (needSto
...
 			return true
 		}
 		monitorErr = err
 		return false
 	}
 
 	if bim.Status.OwnerID != m.controllerID {
-		m.log.Infof("Stopping monitoring because the backing image manager owner ID becomes %v", bim.Status.OwnerID)
+		m.log.Warn("Stopping monitoring because the backing image manager owner ID becomes %v", bim.Status.OwnerID)
```
derekbit, line 1225:
```suggestion
		m.log.Warnf("Stopping monitoring because the backing image manager owner ID becomes %v", bim.Status.OwnerID)
```

controller/backup_controller.go
```diff
@@ -780,31 +780,31 @@ func (bc *BackupController) syncBackupStatusWithSnapshotCreationTimeAndVolumeSiz
...
 	if err != nil {
-		bc.logger.Warnf("syncBackupStatusWithSnapshotCreationTimeAndVolumeSize: failed to get engine client: %v", err)
+		bc.logger.WithError(err).Warn("failed to get engine client when syncing backup status")
 		return
 	}
 
 	e, err := bc.ds.GetVolumeCurrentEngine(volume.Name)
 	if err != nil {
-		bc.logger.Warnf("syncBackupStatusWithSnapshotCreationTimeAndVolumeSize: failed to get engine: %v", err)
+		bc.logger.WithError(err).Warn("failed to get engine when syncing backup status")
```
derekbit, line 789:
```suggestion
		bc.logger.WithError(err).Warn("Failed to get engine when syncing backup status")
```

controller/backup_controller.go
```diff
@@ -780,31 +780,31 @@ func (bc *BackupController) syncBackupStatusWithSnapshotCreationTimeAndVolumeSiz
...
 		return
 	}
 
 	engineClientProxy, err := engineapi.GetCompatibleClient(e, engineCliClient, bc.ds, bc.logger, bc.proxyConnCounter)
 	if err != nil {
-		bc.logger.Warnf("syncBackupStatusWithSnapshotCreationTimeAndVolumeSize: failed to get proxy: %v", err)
+		bc.logger.WithError(err).Warn("failed to get proxy when syncing backup status")
```
derekbit, line 795:
```suggestion
		bc.logger.WithError(err).Warn("Failed to get proxy when syncing backup status")
```

controller/backup_controller.go
```diff
@@ -780,31 +780,31 @@ func (bc *BackupController) syncBackupStatusWithSnapshotCreationTimeAndVolumeSiz
...
 		return
 	}
 	defer engineClientProxy.Close()
 
 	snap, err := engineClientProxy.SnapshotGet(e, backup.Spec.SnapshotName)
 	if err != nil {
-		bc.logger.Warnf("syncBackupStatusWithSnapshotCreationTimeAndVolumeSize: failed to get snapshot %v: %v", backup.Spec.SnapshotName, err)
+		bc.logger.WithError(err).Warnf("failed to get snapshot %v when syncing backup status", backup.Spec.SnapshotName)
```
derekbit, line 802:
```suggestion
		bc.logger.WithError(err).Warnf("Failed to get snapshot %v when syncing backup status", backup.Spec.SnapshotName)
```

controller/backup_controller.go
```diff
@@ -780,31 +780,31 @@ func (bc *BackupController) syncBackupStatusWithSnapshotCreationTimeAndVolumeSiz
...
 		return
 	}
 
 	if snap == nil {
-		bc.logger.Warnf("syncBackupStatusWithSnapshotCreationTimeAndVolumeSize: couldn't find the snapshot %v in volume %v ", backup.Spec.SnapshotName, volume.Name)
+		bc.logger.WithError(err).Warnf("failed to get the snapshot %v in volume %v when syncing backup status", backup.Spec.SnapshotName, volume.Name)
```
derekbit, line 807:
```suggestion
		bc.logger.WithError(err).Warnf("Failed to get the snapshot %v in volume %v when syncing backup status", backup.Spec.SnapshotName, volume.Name)
```

controller/backing_image_manager_controller.go
```diff
@@ -1029,7 +1018,7 @@ func (c *BackingImageManagerController) enqueueForLonghornNode(obj interface{})
 	bims, err := c.ds.ListBackingImageManagersByNode(node.Name)
 	if err != nil {
 		if apierrors.IsNotFound(err) {
-			c.logger.WithField("node", node.Name).Warnf("Can't list backing image managers for a node, may be deleted")
+			c.logger.WithField("node", node.Name).Warn("Can't list backing image managers for a node, may be deleted")
```
derekbit, line 1021:
```suggestion
			c.logger.WithField("node", node.Name).Warn("Failed to list backing image managers for a node, may be deleted")
```

controller/backup_controller.go
```diff
@@ -780,31 +780,31 @@ func (bc *BackupController) syncBackupStatusWithSnapshotCreationTimeAndVolumeSiz
 	backup.Status.VolumeSize = strconv.FormatInt(volume.Spec.Size, 10)
 	engineCliClient, err := bc.getEngineBinaryClient(volume.Name)
 	if err != nil {
-		bc.logger.Warnf("syncBackupStatusWithSnapshotCreationTimeAndVolumeSize: failed to get engine client: %v", err)
+		bc.logger.WithError(err).Warn("failed to get engine client when syncing backup status")
```
derekbit, line 783:
```suggestion
		bc.logger.WithError(err).Warn("Failed to get engine client when syncing backup status")
```

controller/backup_controller.go
```diff
@@ -716,18 +716,18 @@ func (bc *BackupController) syncBackupVolume(volumeName string) error {
 		// Request backup_volume_controller to reconcile BackupVolume immediately.
 		backupVolume.Spec.SyncRequestedAt = syncTime
 		if _, err = bc.ds.UpdateBackupVolume(backupVolume); err != nil && !apierrors.IsConflict(errors.Cause(err)) {
-			bc.logger.WithError(err).Errorf("Error updating backup volume %s spec", volumeName)
+			bc.logger.WithError(err).Errorf("Failed to update backup volume %v spec", volumeName)
```
derekbit, line 719: Why not error out here?

controller/backup_controller.go
```diff
@@ -716,18 +716,18 @@ func (bc *BackupController) syncBackupVolume(volumeName string) error {
...
 		backupTarget, err := bc.ds.GetBackupTarget(types.DefaultBackupTargetName)
 		if err != nil {
-			bc.logger.WithError(err).Warn("Failed to get backup target")
+			bc.logger.WithError(err).Error("Failed to get backup target")
 			return err
 		}
 		backupTarget.Spec.SyncRequestedAt = syncTime
 		if _, err = bc.ds.UpdateBackupTarget(backupTarget); err != nil && !apierrors.IsConflict(errors.Cause(err)) {
-			bc.logger.WithError(err).Warn("Failed to update backup target")
+			bc.logger.WithError(err).Error("Failed to update backup target")
```
derekbit, line 730: It seems not to hinder the overall logic, so using `Warn` might be better.

controller/backup_controller.go
derekbit, line 290:
```suggestion
				log.WithError(err).Warn("Failed to delete remote backup")
```

controller/backup_controller.go
ChanYiLin: I am also confused with this
In Backup/BackupVolume/BackupTarget controller, there are many place only log error but don't return 
some with comment 
```
// Ignore error to prevent enqueue
```

controller/backing_image_manager_controller.go
ChanYiLin: We have lots of `apierrors.IsNotFound(err)` 
Do you think we should info it or warn it or debug it? 

From my point of view
Most of the time we try to get the resource because we want to do something
But since it is gone we don't need to do anything, so we can skip it.

If our goal here is to reduce the log size, I think we don't have to log this kind of message.
Moreover, If the resource is gone we should be aware by the missing of the CR in the cluster and no need to check the log.

ChanYiLin: After discussion,
For this case, since it will stop the monitor because of this condition, so we should use `info` here
Thanks @derekbit

controller/backup_controller.go
ChanYiLin: Use `Warn` since it doesn't return err

ChanYiLin:
Hi @innobead @derekbit
thanks for the feedback
Fixed, thanks!

controller/backing_image_manager_controller.go
```diff
@@ -541,7 +538,7 @@ func (c *BackingImageManagerController) handleBackingImageFiles(bim *longhorn.Ba
 	}
 
 	if err := engineapi.CheckBackingImageManagerCompatibility(bim.Status.APIMinVersion, bim.Status.APIVersion); err != nil {
-		log.Debug("BackingImageManagerController will skip handling files for incompatible backing image manager")
+		log.Warn("Skipping handling files for incompatible backing image manager")
```
innobead, line 541: log.withError

controller/backup_controller.go
```diff
@@ -263,8 +262,7 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
 		if types.ErrorIsNotFound(err) {
 			return nil // Ignore error to prevent enqueue
 		}
-		log.WithError(err).Warning("Cannot find backup volume name")
-		return err
+		return errors.Wrapf(err, "Failed to get backup volume name")
```
innobead, line 265: `failed to ...`

controller/backup_controller.go
```diff
@@ -282,18 +280,18 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
...
 			if unused, err := bc.isBackupNotBeingUsedForVolumeRestore(backup.Name, backupVolumeName); !unused {
-				log.WithError(err).Warnf("Unable to delete remote backup")
+				log.WithError(err).Warn("Failed to delete remote backup")
 				return nil
 			}
 
 			backupURL := backupstore.EncodeBackupURL(backup.Name, backupVolumeName, backupTargetClient.URL)
 			if err := backupTargetClient.BackupDelete(backupURL, backupTargetClient.Credential); err != nil {
-				log.WithError(err).Error("Error deleting remote backup")
+				log.WithError(err).Error("Failed to delete remote backup")
```
innobead, line 294: return a wrappered error instead?

controller/backup_controller.go
```diff
@@ -716,13 +714,13 @@ func (bc *BackupController) syncBackupVolume(volumeName string) error {
...
 		if _, err = bc.ds.UpdateBackupVolume(backupVolume); err != nil && !apierrors.IsConflict(errors.Cause(err)) {
-			bc.logger.WithError(err).Errorf("Error updating backup volume %s spec", volumeName)
+			bc.logger.WithError(err).Warnf("Failed to update backup volume %v spec", volumeName)
 		}
 	} else if err != nil && apierrors.IsNotFound(err) {
 		// Request backup_target_controller to reconcile BackupTarget immediately.
 		backupTarget, err := bc.ds.GetBackupTarget(types.DefaultBackupTargetName)
 		if err != nil {
-			bc.logger.WithError(err).Warn("Failed to get backup target")
+			bc.logger.WithError(err).Error("Failed to get backup target")
```
innobead, line 723: return an wrapperd error instead?

controller/backup_target_controller.go
```diff
@@ -276,9 +276,7 @@ func (btc *BackupTargetController) reconcile(name string) (err error) {
 
 	// Every controller should do the clean up even it is not responsible for the CR
 	if backupTarget.Spec.BackupTargetURL == "" {
-		if err := btc.cleanUpAllMounts(backupTarget); err != nil {
-			log.WithError(err).Warn("Failed to clean up all mount points")
-		}
+		btc.cleanUpAllMounts(backupTarget)
```
innobead, line 279: Don't skip the returned error. I think the original warning is ok?

controller/backup_target_controller.go
```diff
@@ -338,15 +336,13 @@ func (btc *BackupTargetController) reconcile(name string) (err error) {
 		backupTarget.Status.Conditions = types.SetCondition(backupTarget.Status.Conditions,
 			longhorn.BackupTargetConditionTypeUnavailable, longhorn.ConditionStatusTrue,
 			longhorn.BackupTargetConditionReasonUnavailable, "backup target URL is empty")
-		// Clean up all BackupVolume CRs
 		if err := btc.cleanupBackupVolumes(); err != nil {
-			log.WithError(err).Error("failed to clean up BackupVolumes")
+			log.WithError(err).Error("Failed to clean up BackupVolumes")
```
innobead, line 340: Wrapped error instead?

controller/backup_target_controller.go
```diff
@@ -338,15 +336,13 @@ func (btc *BackupTargetController) reconcile(name string) (err error) {
...
 			return err
 		}
 
-		// Clean up all SystemBackup CRs
 		if err := btc.cleanupSystemBackups(); err != nil {
-			log.WithError(err).Error("failed to clean up SystemBackups")
+			log.WithError(err).Error("Failed to clean up SystemBackups")
```
innobead, line 345: Wrapped error instead?

controller/backup_volume_controller.go
```diff
@@ -210,32 +210,32 @@ func (bvc *BackupVolumeController) reconcile(backupVolumeName string) (err error
...
 	backupTarget, err := bvc.ds.GetBackupTargetRO(types.DefaultBackupTargetName)
 	if err != nil {
-		if !apierrors.IsNotFound(err) {
-			return err
+		if apierrors.IsNotFound(err) {
+			return nil
 		}
-		log.Warnf("Cannot found the %s backup target", types.DefaultBackupTargetName)
-		return nil
+		log.WithError(err).Errorf("Failed to get %s backup target", types.DefaultBackupTargetName)
```
innobead, line 216: Wrapped error instead? `failed to ...`

controller/backup_volume_controller.go
```diff
@@ -210,32 +210,32 @@ func (bvc *BackupVolumeController) reconcile(backupVolumeName string) (err error
...
+		return err
 	}
 
 	// Examine DeletionTimestamp to determine if object is under deletion
 	if !backupVolume.DeletionTimestamp.IsZero() {
 
 		if err := bvc.ds.DeleteAllBackupsForBackupVolume(backupVolumeName); err != nil {
-			log.WithError(err).Error("Error deleting backups")
+			log.WithError(err).Error("Failed to delete backups")
```
innobead, line 224: Wrapped error instead?

controller/backup_volume_controller.go
```diff
@@ -210,32 +210,32 @@ func (bvc *BackupVolumeController) reconcile(backupVolumeName string) (err error
...
 			if err != nil || engineClientProxy == nil {
-				log.WithError(err).Error("Error init backup target clients")
+				log.WithError(err).Error("Failed to init backup target clients")
 				return nil // Ignore error to prevent enqueue
 			}
 			defer engineClientProxy.Close()
 
 			if err := backupTargetClient.BackupVolumeDelete(backupTargetClient.URL, backupVolumeName, backupTargetClient.Credential); err != nil {
-				log.WithError(err).Error("Error clean up remote backup volume")
+				log.WithError(err).Error("Failed to delete remote backup volume")
```
innobead, line 238: Wrapped error instead?

controller/backup_volume_controller.go
```diff
@@ -351,15 +351,15 @@ func (bvc *BackupVolumeController) reconcile(backupVolumeName string) (err error
 	}
 	for backupName := range backupsToDelete {
 		if err = bvc.ds.DeleteBackup(backupName); err != nil {
-			log.WithError(err).Errorf("Error deleting backup %s from cluster", backupName)
+			log.WithError(err).Errorf("Failed to delete backup %s from cluster", backupName)
```
innobead, line 354: Wrapped error instead?

innobead (changes requested):

controller/backing_image_data_source_controller.go
```diff
@@ -245,12 +245,10 @@ func (c *BackingImageDataSourceController) syncBackingImageDataSource(key string
...
 	if err != nil {
-		if !datastore.ErrorIsNotFound(err) {
-			c.logger.WithField("backingImageDataSource", name).WithError(err).Error("Failed to retrieve backing image data source from datastore")
-			return err
+		if datastore.ErrorIsNotFound(err) {
+			return nil
 		}
-		c.logger.WithField("backingImageDataSource", name).Warn("Failed to find backing image data source, may have been deleted")
-		return nil
+		return errors.Wrapf(err, "Failed to get backing image data source")
```
innobead, line 251:
```suggestion
		return errors.Wrap(err, "failed to get backing image data source")
```

controller/backing_image_manager_controller.go
```diff
@@ -237,12 +237,10 @@ func (c *BackingImageManagerController) syncBackingImageManager(key string) (err
...
 	if err != nil {
-		if !datastore.ErrorIsNotFound(err) {
-			c.logger.WithField("backingImageManager", name).WithError(err).Error("Failed to retrieve backing image manager from datastore")
-			return err
+		if datastore.ErrorIsNotFound(err) {
+			return nil
 		}
-		c.logger.WithField("backingImageManager", name).Warn("Can't find backing image manager, may have been deleted")
-		return nil
+		return errors.Wrapf(err, "failed to get backing image manager")
```
innobead, line 243:
```suggestion
		return errors.Wrap(err, "failed to get backing image manager")
```

controller/backup_controller.go
```diff
@@ -282,27 +280,26 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
...
-				log.WithError(err).Warnf("Unable to delete remote backup")
+				log.WithError(err).Warn("Failed to delete remote backup")
 				return nil
 			}
 
 			backupURL := backupstore.EncodeBackupURL(backup.Name, backupVolumeName, backupTargetClient.URL)
 			if err := backupTargetClient.BackupDelete(backupURL, backupTargetClient.Credential); err != nil {
-				log.WithError(err).Error("Error deleting remote backup")
-				return err
+				return errors.Wrapf(err, "failed to delete remote backup")
```
innobead, line 294:
```suggestion
				return errors.Wrap(err, "failed to delete remote backup")
```

controller/backup_target_controller.go
```diff
@@ -338,16 +338,12 @@ func (btc *BackupTargetController) reconcile(name string) (err error) {
 		backupTarget.Status.Conditions = types.SetCondition(backupTarget.Status.Conditions,
 			longhorn.BackupTargetConditionTypeUnavailable, longhorn.ConditionStatusTrue,
 			longhorn.BackupTargetConditionReasonUnavailable, "backup target URL is empty")
-		// Clean up all BackupVolume CRs
 		if err := btc.cleanupBackupVolumes(); err != nil {
-			log.WithError(err).Error("failed to clean up BackupVolumes")
-			return err
+			return errors.Wrapf(err, "failed to clean up BackupVolumes")
```
innobead, line 342:
```suggestion
			return errors.Wrap(err, "failed to clean up BackupVolumes")
```

controller/backup_volume_controller.go
```diff
@@ -210,33 +210,30 @@ func (bvc *BackupVolumeController) reconcile(backupVolumeName string) (err error
...
+		return errors.Wrapf(err, "failed to get %s backup target", types.DefaultBackupTargetName)
 	}
 
 	// Examine DeletionTimestamp to determine if object is under deletion
 	if !backupVolume.DeletionTimestamp.IsZero() {
 
 		if err := bvc.ds.DeleteAllBackupsForBackupVolume(backupVolumeName); err != nil {
-			log.WithError(err).Error("Error deleting backups")
-			return err
+			return errors.Wrapf(err, "failed to delete backups")
```
innobead, line 223:
```suggestion
			return errors.Wrap(err, "failed to delete backups")
```

controller/backup_volume_controller.go
```diff
@@ -210,33 +210,30 @@ func (bvc *BackupVolumeController) reconcile(backupVolumeName string) (err error
...
-				log.WithError(err).Error("Error init backup target clients")
+				log.WithError(err).Error("Failed to init backup target clients")
 				return nil // Ignore error to prevent enqueue
 			}
 			defer engineClientProxy.Close()
 
 			if err := backupTargetClient.BackupVolumeDelete(backupTargetClient.URL, backupVolumeName, backupTargetClient.Credential); err != nil {
-				log.WithError(err).Error("Error clean up remote backup volume")
-				return err
+				return errors.Wrapf(err, "failed to delete remote backup volume")
```
innobead, line 236:
```suggestion
				return errors.Wrap(err, "failed to delete remote backup volume")
```

innobead (approved):
LGTM.

We need to create a coding convention guide in https://github.com/longhorn/longhorn/wiki later, which can be after the 1.5.0 release.

PR #1924, opened 2023-05-29
feat(system-backup): volume backup policy

c3y1huang:
https://github.com/longhorn/longhorn/issues/5011


c3y1huang:
Test result: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/3915/

controller/system_backup_controller.go
```diff
@@ -683,6 +700,151 @@ func (c *SystemBackupController) GenerateSystemBackup(systemBackup *longhorn.Sys
...
+
+func (c *SystemBackupController) backupVolumesIfNotPresent(systemBackup *longhorn.SystemBackup) (map[string]*longhorn.Backup, error) {
+	volumes, err := c.ds.ListVolumesRO()
+	if err != nil {
+		return nil, err
+	}
+
+	volumeBackups := make(map[string]*longhorn.Backup, len(volumes))
+	for _, volume := range volumes {
+		if volume.Status.LastBackup != "" {
```
shuo-wu, line 740: Will this field be unset when the last backup is removed?

shuo-wu (commented):
In general LGTM

controller/system_backup_controller.go
innobead: https://github.com/longhorn/longhorn-manager/blob/023dd3cbd5009f390608ea9cdd9f31bd397cf8a2/controller/volume_controller.go#L4270-L4271?

PR #1925, opened 2023-05-29
chore: remove deprecated fields

c3y1huang:
https://github.com/longhorn/longhorn/issues/3289


c3y1huang:
Test result: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/3916/

innobead (approved):

PR #1928, opened 2023-05-29
 feat(dr-volume): forcibly activate a DR volume

mantissahz:
Allow to forcibly activate a DR volume as long as there is a ready replica.
Modified `activate` API to receive the force parameter.
Add a `volume.Status` parameter `ForciblyActivating`

Ref: longhorn/longhorn#1512


shuo-wu (commented):
I am thinking if adding a new field in volume spec is necessary.
How about just blindly activating the DR volume if users set `v.Spec.Standby` to `false` via UI or `kubectl`?

mantissahz:
>I am thinking if adding a new field in volume spec is necessary.

or a global setting? but IMO it should handle each DR volume separately.

>How about just blindly activating the DR volume if users set v.Spec.Standby to false via UI or kubectl?

Would it be dangerous if there is only one ready replica?
I think we could blindly activating the DR volume if users who knew the DR volume is not healthy set `v.Spec.Standby` to `false`

@innobead WDYT? and Does this ticket https://github.com/longhorn/longhorn/issues/1512 resemble more of a bug fixing or an improvement?

shuo-wu:
> Would it be dangerous if there is only one ready replica?

No, since the backup is still there. 
And for the UI action, we can still retain the flag `force`. For the `kubectl` operation, users should understand the risk.

controller/volume_controller.go
```diff
@@ -2878,12 +2878,19 @@ func (c *VolumeController) checkAndFinishVolumeRestore(v *longhorn.Volume, e *lo
...
 
-	allScheduledReplicasIncluded, err := c.checkAllScheduledReplicasIncluded(v, e, rs)
+	allScheduledReplicasIncluded, healthReplicaCount, err := c.checkAllScheduledReplicasIncluded(v, e, rs)
 	if err != nil {
 		return err
 	}
 
-	if !isPurging && (v.Status.Robustness == longhorn.VolumeRobustnessHealthy || v.Status.Robustness == longhorn.VolumeRobustnessDegraded) && allScheduledReplicasIncluded {
+	allowCreateDegraded, err := c.ds.GetSettingAsBool(types.SettingNameAllowVolumeCreationWithDegradedAvailability)
+	if err != nil {
```
innobead, line 2887: `allowCreatDegraded` -> `degradedVolumeSupported`

controller/volume_controller.go
```diff
@@ -2878,12 +2878,19 @@ func (c *VolumeController) checkAndFinishVolumeRestore(v *longhorn.Volume, e *lo
...
+		return err
+	}
+
+	if !isPurging &&
+		(v.Status.Robustness == longhorn.VolumeRobustnessHealthy || v.Status.Robustness == longhorn.VolumeRobustnessDegraded) &&
+		(allScheduledReplicasIncluded || (allowCreateDegraded && healthReplicaCount > 0)) {
```
innobead, line 2893: I just noticed the original condition already having `v.Status.Robustness == longhorn.VolumeRobustnessDegraded`, so should we still check `allowCreateDegraded`? 

or just below, good enough?

```go
if !isPurging &&
		(v.Status.Robustness == longhorn.VolumeRobustnessHealthy || v.Status.Robustness == longhorn.VolumeRobustnessDegraded) {
```

innobead (changes requested):

controller/volume_controller.go
mantissahz: Modified. Thanks

controller/volume_controller.go
mantissahz: We still need to check `allScheduledReplicasIncluded` and `degradedVolumeSupported` if users disable the global setting `allow-volume-creation-with-degraded-availability`.

controller/volume_controller.go
```diff
@@ -2883,7 +2883,14 @@ func (c *VolumeController) checkAndFinishVolumeRestore(v *longhorn.Volume, e *lo
...
 
-	if !isPurging && (v.Status.Robustness == longhorn.VolumeRobustnessHealthy || v.Status.Robustness == longhorn.VolumeRobustnessDegraded) && allScheduledReplicasIncluded {
+	degradedVolumeSupported, err := c.ds.GetSettingAsBool(types.SettingNameAllowVolumeCreationWithDegradedAvailability)
+	if err != nil {
+		return err
+	}
+
+	if !isPurging &&
+		(v.Status.Robustness == longhorn.VolumeRobustnessHealthy || v.Status.Robustness == longhorn.VolumeRobustnessDegraded) &&
+		(allScheduledReplicasIncluded || degradedVolumeSupported) {
```
innobead, line 2893: Sounds good to me. (nit: I feel `allScheduledReplicasIncluded` is redundant if we already know the volume is healthy)

However, suggest adjusting the condition to make it clear like below. WDYT? 

```go
	if !isPurging &&
		(v.Status.Robustness == longhorn.VolumeRobustnessHealthy && allScheduledReplicasIncluded) || (v.Status.Robustness == longhorn.VolumeRobustnessDegraded && degradedVolumeSupported) {
```

Also, we need to update DR doc to clarify the conditions for the activation.

innobead (changes requested):

controller/volume_controller.go
mantissahz: Modified. Thanks

innobead (approved):

PR #1929, opened 2023-05-30
fix(logging): review and adjust log message and level in engine, relica and instance controllers

ChanYiLin:

Main Idea:
- use `Failed to + V` instead of `Error Ving`
- when doing operation, log before with `.Info("Ving + ...")` and no need to log after operation is successful since if there is any error, error will be returned and reported in upper function.
- if an error will be consumed w/o being returned back to the caller, using warning, since it has been handled.
- If an error will be returned/bubbled up, suggest no extra logging but can be enriched with another error wrapper to prevent a duplicated error log. This needs to check if there would be a chance to have duplicated error message.


ChanYiLin:
> This pull request is now in conflicts. Could you fix it @ChanYiLin? 🙏

conflicts resolved

ChanYiLin:
Hi @innobead
I have fixed the 
- `Info` instead of `Infof` if there is no argument (`Warn`, `Error` as well)
- Error logging in the `handleErr` function

innobead (approved):

PR #1932, opened 2023-05-31
Remove webhook PDBs in the upgrade path from v1.4.x to v1.5.0

PhanLe1010:
longhorn/longhorn#6026


types/deploy.go
```diff
@@ -1,8 +1,10 @@
 package types
 
 const (
-	LonghornManagerDaemonSetName = "longhorn-manager"
-	LonghornUIDeploymentName     = "longhorn-ui"
+	LonghornManagerDaemonSetName            = "longhorn-manager"
+	LonghornUIDeploymentName                = "longhorn-ui"
+	LonghornAdmissionWebhookDeploymentName  = "longhorn-admission-webhook"
+	LonghornConversionWebhookDeploymentName = "longhorn-conversion-webhook"
```
innobead, line 7: There are no deployments for webhooks anymore in 1.5, so adding these const here are misleading. 

Suggest just adding them to the upgrade.go, so they can be specific there.

innobead (changes requested):

types/deploy.go
PhanLe1010: Great point. Thank you! Fixed

innobead (approved):

PR #1951, opened 2023-06-05
fix(logging): review and adjust log message and level in api and app

ChanYiLin:

Also update
- skip the log if log level is the same

### DIFF
--- a/controller/backing_image_controller.go
+++ b/controller/backing_image_controller.go
@@ -107,8 +107,8 @@ func (bic *BackingImageController) Run(workers int, stopCh <-chan struct{}) {
 	defer utilruntime.HandleCrash()
 	defer bic.queue.ShutDown()
 
-	logrus.Infof("Starting Longhorn Backing Image controller")
-	defer logrus.Infof("Shut down Longhorn Backing Image controller")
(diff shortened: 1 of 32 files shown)


controller/backing_image_controller.go
```diff
@@ -147,13 +147,13 @@ func (bic *BackingImageController) handleErr(err error, key interface{}) {
 	}
 
 	if bic.queue.NumRequeues(key) < maxRetries {
-		logrus.WithError(err).Warnf("Error syncing Longhorn backing image %v", key)
+		logrus.WithError(err).Errorf("Error syncing Longhorn backing image %v", key)
```
innobead, line 150: Suggest keeping warn, since we handle the error by adding it to the queue for retry.

innobead (changes requested):
LGTM. Just one comment.

innobead:
@mergify backport v1.5.x

controller/backing_image_controller.go
ChanYiLin: As discussed, this is the top caller function, so we should use `Error`
Since adding to queue to retry is not longhorn mechanism bus k8s which is outside of our control

PR #1952, opened 2023-06-06
spdk: validate nodes running instance manager pod rather than all nodes

derekbit:
Longhorn/longhorn#5739


innobead (approved):
LGTM

innobead:
@mergify backport v1.5.x

PR #1957, opened 2023-06-07
fix(backup uninstall): delete backup having no backup volume manually when uninstall

ChanYiLin:
ref: [longhorn/longhorn#5868](https://github.com/longhorn/longhorn/issues/5868)

### DIFF
--- a/controller/uninstall_controller.go
+++ b/controller/uninstall_controller.go
@@ -432,7 +432,12 @@ func (c *UninstallController) deleteCRDs() (bool, error) {
 	if backups, err := c.ds.ListBackups(); err != nil {
 		return true, err
 	} else if len(backups) > 0 {
-		c.logger.Infof("Found %d backups remaining", len(backups))
+		c.logger.Infof("Found %d backups remaining, deleting if they don't have backup volume", len(backups))
+		for _, backup := range backups {
+			if err := c.deleteLeftBackups(backup); err != nil {
+				return true, err
+			}
+		}
 		return true, nil
 	}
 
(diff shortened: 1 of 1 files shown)


controller/uninstall_controller.go
```diff
@@ -617,6 +622,23 @@ func (c *UninstallController) deleteReplicas(replicas map[string]*longhorn.Repli
...
+func (c *UninstallController) deleteLeftBackups(backup *longhorn.Backup) (err error) {
+	backupVolumeName, ok := backup.Labels[types.LonghornLabelBackupVolume]
+	if !ok {
+		// directly delete it if there is even no backup volume label
+		c.ds.DeleteBackup(backup.Name)
+	}
+	_, err = c.ds.GetBackupVolume(backupVolumeName)
+	if err != nil && apierrors.IsNotFound(err) {
+		if err = c.ds.DeleteBackup(backup.Name); err != nil {
+			return errors.Wrap(err, "failed to mark for deletion")
```
innobead, line 635: Suggest enriching the message, like `failed to delete backup %s`.

innobead (changes requested):
LGTM, just one comment.

controller/uninstall_controller.go
```diff
@@ -617,6 +622,23 @@ func (c *UninstallController) deleteReplicas(replicas map[string]*longhorn.Repli
 	return
 }
 
+// deleteLeftBackups deletes the backup having no backup volume
+func (c *UninstallController) deleteLeftBackups(backup *longhorn.Backup) (err error) {
+	backupVolumeName, ok := backup.Labels[types.LonghornLabelBackupVolume]
+	if !ok {
+		// directly delete it if there is even no backup volume label
+		c.ds.DeleteBackup(backup.Name)
```
derekbit, line 630: What happens if this operation fails?

controller/uninstall_controller.go
```diff
@@ -617,6 +622,23 @@ func (c *UninstallController) deleteReplicas(replicas map[string]*longhorn.Repli
...
+	}
+	_, err = c.ds.GetBackupVolume(backupVolumeName)
+	if err != nil && apierrors.IsNotFound(err) {
+		if err = c.ds.DeleteBackup(backup.Name); err != nil {
+			return errors.Wrap(err, "failed to mark for deletion")
+		}
+		return nil
```
derekbit, line 637: no need to error out when running into a `NotFound` error when deleting a backup.

controller/uninstall_controller.go
ChanYiLin: Also need to catch the error here, thanks for the remind!

controller/uninstall_controller.go
ChanYiLin: yeah, I have the same idea, but the whole uninstall process use the same wording...
I will change others to the same wording as well
thanks!

ChanYiLin: Oh, I got it, they all have defer function to add more message
Then I will just need to fix this one

ChanYiLin: Fixed

controller/uninstall_controller.go
ChanYiLin: Fixed, thanks

controller/uninstall_controller.go
ChanYiLin: Fixed.

innobead (approved):

innobead:

PR #1959, opened 2023-06-08
spdk: disallow disabling spdk if there are block-type disks on nodes

derekbit:
Longhorn/longhorn#5778


innobead:
@mergify backport v1.5.x

PR #1965, opened 2023-06-08
fix(recurringjob): keep snapshots for the backup

mantissahz:
Add a new global setting that allows users to keep the retain number of snapshots when the recurring job does a backup.

Ref: longhorn/longhorn#2997


derekbit (approved):
LGTM

app/recurring_job.go
```diff
@@ -552,7 +557,7 @@ func (job *Job) filterExpiredSnapshotsOfCurrentRecurringJob(snapshotCRs []longho
 	// Only consider deleting the snapshots that were created by our current job
 	snapshotCRs = filterSnapshotCRsWithLabel(snapshotCRs, types.RecurringJobLabel, jobLabel)
 
-	if job.task == longhorn.RecurringJobTypeSnapshot || job.task == longhorn.RecurringJobTypeSnapshotForceCreate {
+	if job.task == longhorn.RecurringJobTypeSnapshot || job.task == longhorn.RecurringJobTypeSnapshotForceCreate || !allowBackupSnapshotDeleted {
 		return filterExpiredItems(snapshotCRsToNameWithTimestamps(snapshotCRs), job.retain)
 	}
```
shuo-wu, line 563: NIT: Add new comments or update the below comments

shuo-wu (dismissed):
LGTM

shuo-wu (dismissed):

mantissahz:
@mergify backport v1.5.x

app/recurring_job.go
```diff
@@ -785,6 +791,20 @@ func (job *Job) UpdateVolumeStatus(v *longhorn.Volume) (*longhorn.Volume, error)
 	return job.lhClient.LonghornV1beta2().Volumes(job.namespace).UpdateStatus(context.TODO(), v, metav1.UpdateOptions{})
 }
 
+// GetSettingAsBool returns boolean of the setting value searching by name.
+func (job *Job) GetSettingAsBool(name string) (bool, error) {
+	obj, err := job.lhClient.LonghornV1beta2().Settings(job.namespace).Get(context.TODO(), name, metav1.GetOptions{})
```
c3y1huang, line 796: nit: Consider using `types.SettingName` to prevent rewriting the conversion from the caller each time this method is reused.
```suggestion
func (job *Job) GetSettingAsBool(name types.SettingName) (bool, error) {
	obj, err := job.lhClient.LonghornV1beta2().Settings(job.namespace).Get(context.TODO(), string(name), metav1.GetOptions{})
```

app/recurring_job.go
```diff
@@ -445,7 +445,12 @@ func (job *Job) doSnapshotCleanup(backupDone bool) (err error) {
 		return err
 	}
 
-	cleanupSnapshotNames := job.listSnapshotNamesToCleanup(collection.Data, backupDone)
+	allowBackupSnapshotDeleted, err := job.GetSettingAsBool(string(types.SettingNameAutoCleanupRecurringJobBackupSnapshot))
```
c3y1huang, line 448: Any reason for parsing through as argument instead of getting this setting directly within the  `filterExpiredSnapshotsOfCurrentRecurringJob` method?

app/recurring_job.go
mantissahz: Modified. Thanks

app/recurring_job.go
mantissahz: IIRC it can not return an error in then `filterExpiredSnapshotsOfCurrentRecurringJob` method.
Moved into the `filterExpiredSnapshotsOfCurrentRecurringJob` method and changed to leave a message if an error occurred from the `job.GetSettingAsBool` method.
Thanks.

c3y1huang (approved):

innobead (approved):
LGTM.

PR #1966, opened 2023-06-08
Require file check if the bids in failed state

WebberHuang1118:
**Problem:**
backingimage and backingimagedatasource deletion got stuck, if the backing image is deleted during upload process and bids just changes to `ready-for-transfer` state

**Solution:**
For bim controller, it requires file check as long as the bids in failed state

**Related Issue:**
[https://github.com/longhorn/longhorn/issues/6086](https://github.com/longhorn/longhorn/issues/6086)

**Test Plan:**
- Via Longhorn GUI, upload backing image from local bi_upload
- Delete backingimage as bids just turns into `ready-for-transfer` state 
  - Delete backingimage by `$ kubectl -n longhorn-system delete backingimage img1 --cascade=foreground` bi_delete_ready_for_transfer
- CR backingimage, backingimagedatasource and backingimagemanager should be cleared


innobead:
@WebberHuang1118 Please help fix this for the master instead. After being merged, we will backport to the supported branches.

WebberHuang1118:
@innobead, this PR is updated for master, thanks

ChanYiLin:
LGTM, the PR is a quick solution to solve the issue and has no side effect. 
But might get confused about why backing-image-manager require files when bids failed. 

We can merge it first
Thanks for the contribution :+1:

ChanYiLin (dismissed):

controller/backing_image_manager_controller.go
```diff
@@ -649,12 +643,20 @@ func (c *BackingImageManagerController) prepareBackingImageFiles(currentBIM *lon
...
+
 		bids, err := c.ds.GetBackingImageDataSource(biName)
 		if err != nil {
 			log.WithError(err).Warn("Failed to get backing image data source before preparing files, will skip handling this backing image")
 			continue
 		}
 
+		currentInfo, exists := currentBIM.Status.BackingImageFileMap[biName]
+		requireFile := !exists || currentInfo.State == longhorn.BackingImageStateFailed
+		requireFile = requireFile || bids.Status.CurrentState == longhorn.BackingImageStateFailed
```
shuo-wu, line 655: NIT: As mentioned above, this is confusing without the explanation in [this comment](https://github.com/longhorn/longhorn/issues/6086#issuecomment-1590662594)

shuo-wu (approved):
LGTM but it's better to add some comments. We can merge this first

innobead (commented):
LGTM. 

After adding a comment, then we can merge this.

controller/backing_image_manager_controller.go
innobead: @WebberHuang1118 you should resolve the comment until you fix the comment.

controller/backing_image_manager_controller.go
```diff
@@ -649,12 +643,20 @@ func (c *BackingImageManagerController) prepareBackingImageFiles(currentBIM *lon
...
+		if !requireFile {
+			continue
+		}
```
innobead, line 658:
```suggestion
		currentInfo, exists := currentBIM.Status.BackingImageFileMap[biName]
		requireFile := !exists || currentInfo.State == longhorn.BackingImageStateFailed
		// Ensure the bids can be deleted instead of being stuck in the `BackingImageStateFailed` state. 
		// ref: https://github.com/longhorn/longhorn/issues/6086#issuecomment-1590662594
		requireFile = requireFile || bids.Status.CurrentState == longhorn.BackingImageStateFailed
		if !requireFile {
			continue
		}
```

controller/backing_image_manager_controller.go
WebberHuang1118: Got it, thanks

innobead:

PR #1970, opened 2023-06-13
Fix invalid memory address or nil pointer dereference

derekbit:
Longhorn/longhorn#6110


innobead (approved):
LGTM. It seems the volume is allowed to delete before engine resources.

innobead:

PR #1974, opened 2023-06-13
Add metric for last backup of a volume

koeberlue:
Add metric `last_backup_at` for each volume.

It will publish the Unix timestamp (in seconds) of the last backup for each volume. 

Because Prometheus does not have any concept of `nil` values, if no such backup exists, `0` will be returned, which corresponds to `1970-01-01 00:00:00`.

Closes https://github.com/longhorn/longhorn/issues/6049


innobead:
`last_backup_at` seems not a metric but it's more about info. @koeberlue Can you explain why you think this new metric is helpful?

koeberlue:
> `last_backup_at` seems not a metric but it's more about info. @koeberlue Can you explain why you think this new metric is helpful?

I am running the [`kube-prometheus-stack`](https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack) to monitor some kubernetes clusters. I have set up some alerts according to the [documentation](https://longhorn.io/docs/1.4.2/monitoring/alert-rules-example/). 

I want to configure an additional alert for any volume, where the backup is older than one week. In the UI I can see `Last Backup At` for every volume, but I can't scrape this value from the metrics. 

So I guess the main purpose of this metric would be to get alerts whenever a volume is missing backups.

Although I must admit that it probably does not make _that_ much sense to keep track of this metric. Users would probably only ever care for the latest value.  On the other hand when I do search my prometheus instance, there seem to be some other apps that do expose similar metrics holding just a `timestamp` of the last occurence of an event.

innobead:
> I want to configure an additional alert for any volume, where the backup is older than one week. In the UI I can see `Last Backup At` for every volume, but I can't scrape this value from the metrics.
> 

Sounds reasonable to me, even the timestamp is not really a qualified metric type. @PhanLe1010 @c3y1huang  WDYT?

c3y1huang:
Proactively mitigating the risk of data loss from potential backup gaps sounds fair to me.

innobead (approved):

innobead:
@c3y1huang Please help review this before merging it.

metrics_collector/volume_collector.go
```diff
@@ -251,6 +265,14 @@ func getVolumeRobustnessValue(v *longhorn.Volume) int {
 	return robustnessValue
 }
 
+func getLastBackupAtValue(v *longhorn.Volume) int64 {
+	lastBackupAt, err := time.Parse(time.DateTime, v.Status.LastBackupAt)
+	if err != nil {
+		return 0
+	}
```
c3y1huang, line 272: Suggest to inputting just the lastBackupAt rather than the entire Volume object since there is no use for other fields here.
1. If the value is "", return 0
1. If the time pasering fails, print a trace log so the error is not swollowed, and then then return 0.

c3y1huang (changes requested):

metrics_collector/volume_collector.go
```diff
@@ -251,6 +265,14 @@ func getVolumeRobustnessValue(v *longhorn.Volume) int {
 	return robustnessValue
 }
 
+func getLastBackupAtValue(v *longhorn.Volume) int64 {
+	lastBackupAt, err := time.Parse(time.DateTime, v.Status.LastBackupAt)
```
c3y1huang, line 269: Doesn't look like it will parse successfully, probably need to use a different layout.

c3y1huang (changes requested):

PR #1982, opened 2023-06-14
Fix unexpected attachment and detachment behaviors of RWX volumes

derekbit:
- Do not detach attached RWX volumes while upgrading Longhorn system 
- RWX volume is not required if there is no active workload using it

longhorn/longhorn#6120

### DIFF
--- a/controller/share_manager_controller.go
+++ b/controller/share_manager_controller.go
@@ -395,10 +395,12 @@ func (c *ShareManagerController) isShareManagerRequiredForVolume(volume *longhor
 
 	for _, attachmentTicket := range va.Spec.AttachmentTickets {
 		if isRegularRWXVolume(volume) && (isCSIAttacherTicket(attachmentTicket) || isUpgraderTicket(attachmentTicket)) {
-			return true
+			// no active workload, there is no need to keep the share manager around
+			if hasActiveWorkload(volume) {
+				return true
+			}
 		}
 	}
-
 	return false
 }
 
--- a/controller/share_manager_controller.go
+++ b/controller/share_manager_controller.go
@@ -394,7 +394,7 @@ func (c *ShareManagerController) isShareManagerRequiredForVolume(volume *longhor
 	}
 
 	for _, attachmentTicket := range va.Spec.AttachmentTickets {
-		if isCSIAttacherTicketOfRegularRWXVolume(attachmentTicket, volume) {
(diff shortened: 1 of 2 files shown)


controller/share_manager_controller.go
```diff
@@ -394,11 +394,13 @@ func (c *ShareManagerController) isShareManagerRequiredForVolume(volume *longhor
 	}
 
 	for _, attachmentTicket := range va.Spec.AttachmentTickets {
-		if isCSIAttacherTicketOfRegularRWXVolume(attachmentTicket, volume) {
-			return true
+		if isCSIAttacherOrUpgraderTicketOfRegularRWXVolume(attachmentTicket, volume) {
```
shuo-wu, line 397: NIT: How about `isRegularRWXVolume(...) && (isCSIAttacherTicket(...) || isUpgraderTicket(...))`? 
This makes each check function more straightforward and easy-to-use

shuo-wu (approved):
LGTM

innobead (approved):
LGTM

innobead:
@mergify backport v1.5.x

controller/share_manager_controller.go
```diff
@@ -394,11 +394,13 @@ func (c *ShareManagerController) isShareManagerRequiredForVolume(volume *longhor
 	}
 
 	for _, attachmentTicket := range va.Spec.AttachmentTickets {
-		if isCSIAttacherTicketOfRegularRWXVolume(attachmentTicket, volume) {
-			return true
+		if isRegularRWXVolume(volume) && (isCSIAttacherTicket(attachmentTicket) || isUpgraderTicket(attachmentTicket)) {
+			// no active workload, there is no need to keep the share manager around
+			if hasActiveWorkload(volume) {
```
PhanLe1010, line 399: I thought that the existence of `isCSIAttacherTicket(attachmentTicket) || isUpgraderTicket(attachmentTicket)` already indicate that `hasActiveWorkload(volume)` is true? If so, do we need the nested:
```
if hasActiveWorkload(volume) {
    return true
}
```

PhanLe1010: ping @derekbit

derekbit: Yeah, the check is redundant. Remove it https://github.com/longhorn/longhorn-manager/pull/1993. @PhanLe1010  Can you help review it? Thank you.

PR #1983, opened 2023-06-14
SPDK volume does not support engine binary client

derekbit:
Longhorn/longhorn#6123


innobead (approved):
LGTM

innobead:
@mergify backport v1.5.x

shuo-wu:
Should we allow the UI to list SPDK (rebuilding) snapshot info? If NO, then this simple fix should be enough.

derekbit:
> Should we allow the UI to list SPDK (rebuilding) snapshot info? If NO, then this simple fix should be enough.

I think no. We can support snapshot list when snapshotting is supported.

PR #1993, opened 2023-06-15
rwx: remove redundant `hasActiveWorkload` check

derekbit:
The check is unnecessary, and upgrader ticker is enough. When detaching a volume, the upgrader ticker is deleted.

Longhorn/longhorn#6120


innobead (approved):

innobead:
@mergify backport v1.5.x

PR #2000, opened 2023-06-17
Set correct sync-agent-port-count when creating a replica process

derekbit:
The default value of sync-agent-port-count is 10 ([code](https://github.com/longhorn/longhorn-engine/blob/master/app/cmd/replica.go#L58)), and it is fine when DefaultReplicaPortCount is 15. Currently, DefaultReplicaPortCount is set to 10, 3 ports are used by replica server, data server and syncagent server. Hence, sync-agent-port-count should be `DefaultReplicaPortCount - 3`.

Longhorn/longhorn#6145

### DIFF
--- a/engineapi/instance_manager.go
+++ b/engineapi/instance_manager.go
@@ -344,7 +344,7 @@ func getBinaryAndArgsForEngineProcessCreation(e *longhorn.Engine,
 }
 
 func getBinaryAndArgsForReplicaProcessCreation(r *longhorn.Replica,
-	dataPath, backingImagePath string, dataLocality longhorn.DataLocality, engineCLIAPIVersion int) (string, []string) {
+	dataPath, backingImagePath string, dataLocality longhorn.DataLocality, portCount, engineCLIAPIVersion int) (string, []string) {
 
 	args := []string{
 		"replica", types.GetReplicaMountedDataPath(dataPath),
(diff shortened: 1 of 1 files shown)


derekbit:
> sync-agent-port-count should be DefaultReplicaPortCount - 3

Should we use up the remaining port for `sync-agent-port-count`? Any concern?
cc @innobead @shuo-wu

derekbit:
e2e https://ci.longhorn.io/job/private/job/longhorn-tests-regression/4226/

shuo-wu:
> Should we use up the remaining port for sync-agent-port-count? Any concern?

I am fine with using all of the remaining ports.

But should we directly update https://github.com/longhorn/longhorn-engine/blob/0896a6f78315a4fd09e5afa74c85fb3fd5747b56/app/cmd/replica.go#L58 instead?

derekbit:
> > Should we use up the remaining port for sync-agent-port-count? Any concern?
> 
> I am fine with using all of the remaining ports.
> 
> But should we directly update https://github.com/longhorn/longhorn-engine/blob/0896a6f78315a4fd09e5afa74c85fb3fd5747b56/app/cmd/replica.go#L58 instead?

Because default replica port count is set/defined in longhorn-manager, so I changed the value in longhorn-manager. Make them consistent.

shuo-wu:
OK... The default replica port count for spdk engine is defined in `longhorn-spdk-engine`... Do we need to move this default value to `longhorn-engine` then update the default value for this option?

derekbit:
> OK... The default replica port count for spdk engine is defined in `longhorn-spdk-engine`... Do we need to move this default value to `longhorn-engine` then update the default value for this option?

I'm fine with the current implementation. I'm thinking if we can move the port count of spdk volumes to longorn-manager instead.

shuo-wu:
OK. I am OK with either.

shuo-wu (approved):

innobead (approved):

innobead:
> > OK... The default replica port count for spdk engine is defined in `longhorn-spdk-engine`... Do we need to move this default value to `longhorn-engine` then update the default value for this option?
> 
> I'm fine with the current implementation. I'm thinking if we can move the port count of spdk volumes to longorn-manager instead.

@derekbit Let's create an issue for this improvement.

innobead:
@mergify backport v1.5.x

PR #2005, opened 2023-06-19
fix(backup): backup a snapshot twice

mantissahz:
Backing up a small size of a snapshot will cause the backup volume tries to synchronize the `backup.Spec.SyncRequestedAt` and it will make a conflict when this new creating backup tries to update its status and then re-queue it to make a second time snapshot backup at engine side.

Regression test:
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/4235

Re-run Regression test:
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/4245
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/4248
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/4250

Ref: longhorn/longhorn#6130

### DIFF
--- a/controller/backup_controller.go
+++ b/controller/backup_controller.go
@@ -318,6 +318,7 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
 
 	syncTime := metav1.Time{Time: time.Now().UTC()}
 	existingBackup := backup.DeepCopy()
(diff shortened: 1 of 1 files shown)


innobead (approved):
LGTM. Well done.

innobead:

PR #2010, opened 2023-06-20
Highlight CPU usage in v2-data-engine setting

derekbit:
Longhorn/longhorn#6126


innobead (approved):
LGTM.

Update the longhorn manifest as well.

innobead:
@mergify backport v1.5.x

PR #2017, opened 2023-06-20
Fix attaching/detaching loop when a degraded SPDK volume is in maintenance mode

derekbit:
If a degraded volume is in maintenance mode, the replica rebuilding can directly be executed and no need to detach it again.

Longhorn/longhorn#6166

### DIFF
--- a/controller/volume_attachment_controller.go
+++ b/controller/volume_attachment_controller.go
@@ -482,19 +482,33 @@ func (vac *VolumeAttachmentController) shouldDoDetach(va *longhorn.VolumeAttachm
 	if len(currentAttachmentTickets) == 0 {
 		return true
 	}
-	if !hasUninterruptibleTicket(currentAttachmentTickets) && hasWorkloadTicket(attachmentTicketsOnOtherNodes) {
-		log.Info("Workload attachment ticket interrupted snapshot-controller/backup-controller attachment tickets")
+
+	// Check if there is any workload ticket regardless of frontend on other nodes
+	// If exist, detach and interrupt the current ticket.
+	if !hasUninterruptibleTicket(currentAttachmentTickets) && hasWorkloadTicket(attachmentTicketsOnOtherNodes, longhorn.AnyValue) {
+		log.Info("Workload attachment ticket interrupted snapshot/backup/rebuilding-controller attachment tickets")
 		return true
 	}
 
-	// Currently, the only ticket that is interruptible and frontend disabled is the rebuilding-controller ticket
-	// Offline replica rebuilding has the potential to fail and repeated attempts.
-	// Users can give up (interrupt) the replica rebuilding and attach the degraded volume to a node.
-	if hasInterruptibleAndFrontendDisabledTicket(currentAttachmentTickets) && hasWorkloadTicket(va.Spec.AttachmentTickets) {
-		log.Debugf("Workload attachment ticket interrupted rebuilding-controller attachment tickets")
+	// If there is an interruptible ticket and frontend disabled ticket on the current node (currently, only offline rebuilding ticket)
+	// need to check if there is any workload ticket with frontend enabled (disableFrontend=false) on the same node.
+	// If exist, detach and interrupt the rebuilding ticket.
+	if hasInterruptibleAndFrontendDisabledTicket(currentAttachmentTickets) && hasWorkloadTicket(currentAttachmentTickets, longhorn.FalseValue) {
+		log.Info("Workload attachment ticket interrupted rebuilding-controller attachment tickets")
 		return true
 	}
 
+	return false
+}
+
+func hasUninterruptibleTicket(attachmentTickets map[string]*longhorn.AttachmentTicket) bool {
+	for _, ticket := range attachmentTickets {
+		if ticket.Type != longhorn.AttacherTypeSnapshotController &&
+			ticket.Type != longhorn.AttacherTypeBackupController &&
+			ticket.Type != longhorn.AttacherTypeVolumeRebuildingController {
+			return true
+		}
+	}
 	return false
 }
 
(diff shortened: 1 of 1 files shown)


derekbit:
Ready for review
cc @innobead @shuo-wu

controller/volume_attachment_controller.go
```diff
@@ -490,7 +490,7 @@ func (vac *VolumeAttachmentController) shouldDoDetach(va *longhorn.VolumeAttachm
 	// Currently, the only ticket that is interruptible and frontend disabled is the rebuilding-controller ticket
 	// Offline replica rebuilding has the potential to fail and repeated attempts.
 	// Users can give up (interrupt) the replica rebuilding and attach the degraded volume to a node.
-	if hasInterruptibleAndFrontendDisabledTicket(currentAttachmentTickets) && hasWorkloadTicket(va.Spec.AttachmentTickets) {
+	if hasInterruptibleAndFrontendDisabledTicket(currentAttachmentTickets) && !vol.Status.FrontendDisabled && hasWorkloadTicket(va.Spec.AttachmentTickets) {
```
shuo-wu, line 493: Why do you need this line? Can we directly change `hasUninterruptibleTicket` and remove this line instead?
```
func hasUninterruptibleTicket(attachmentTickets map[string]*longhorn.AttachmentTicket) bool {
	for _, ticket := range attachmentTickets {
		if ticket.Type != longhorn.AttacherTypeSnapshotController &&
			ticket.Type != longhorn.AttacherTypeBackupController &&
			ticket.Type != longhorn.AttacherTypeVolumeRebuildingController {
			return true
		}
	}
	return false
}
```

innobead: It's better to update the log `log.Info("Workload attachment ticket interrupted snapshot-controller/backup-controller attachment tickets")` as well.

derekbit: `hasUninterruptibleTicket` can handle both `FrontendDisabled=true` and `FrontendDisabled=false` tickets.
`hasInterruptibleAndFrontendDisabledTicket` only focus on `FrontendDisabled=false` tickets.

Yeah, we can merge them together. However, for readability, I prefer to separate them.

derekbit: Merged them together, but separate the checks.

controller/volume_attachment_controller.go
```diff
@@ -501,6 +501,7 @@ func (vac *VolumeAttachmentController) shouldDoDetach(va *longhorn.VolumeAttachm
 func hasInterruptibleAndFrontendDisabledTicket(attachmentTickets map[string]*longhorn.AttachmentTicket) bool {
 	for _, ticket := range attachmentTickets {
 		if ticket.Type == longhorn.AttacherTypeVolumeRebuildingController {
+			logrus.Infof("Debug ====> hasInterruptibleAndFrontendDisabledTicket: %v", ticket.Type)
```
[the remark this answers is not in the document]
derekbit: Removed. Thank you.

controller/volume_attachment_controller.go
```diff
@@ -482,39 +482,34 @@ func (vac *VolumeAttachmentController) shouldDoDetach(va *longhorn.VolumeAttachm
...
+	if hasInterruptibleTicket(currentAttachmentTickets, vol) && hasWorkloadTicket(attachmentTicketsOnOtherNodes) {
+		log.Info("Workload attachment ticket interrupted snapshot-controller/backup-controller/rebuilding-controller attachment tickets")
 		return true
 	}
 
 	return false
 }
 
-func hasInterruptibleAndFrontendDisabledTicket(attachmentTickets map[string]*longhorn.AttachmentTicket) bool {
+func hasInterruptibleTicket(attachmentTickets map[string]*longhorn.AttachmentTicket, vol *longhorn.Volume) bool {
```
shuo-wu, line 494: Wait... The naming is wrong... When the list contains tickets for auto snapshot/backup/rebuilding, aka interruptible tickets, this function will return `false`... 
I prefer the previous logic: We should do detachment when there is no uninterruptible ticket on the current node but important tickets on other nodes. And with this logic, do we need to check `vol.Status.FrontendDisabled` still?

derekbit: Oh, yes. You're right. Don't need to check `vol.Status.FrontendDisabled` if we consider the tickets on other nodes.

`!hasUninterruptibleTicket` racked my brain...:(

innobead (commented):
After this change, does it mean when dogin offline replica rebuilding, it will not be allowed to interrupt if there is a workload requesting the volume?

shuo-wu (approved):

shuo-wu:
After the change, if the volume is attached by users, the auto detachment for offline rebuilding won't take effect since there is an a ticket created by users. For regular offline rebuilding, it works as usual.

derekbit:
`hasWorkloadTicket` only handles workload tickets on other nodes.
I think we should handle the workload tickets on the local node as well.

innobead:
@derekbit Will update the PR as per the discussion, need to make it interruptable if there is a local workload requesting the attachment which is occupied by the offline rebuilding ticket.

derekbit:
Original `if !hasUninterruptibleTicket(currentAttachmentTickets) && hasWorkloadTicket(attachmentTicketsOnOtherNodes)` only handles workload tickets on other nodes.

To avoid the workload attachment stuck on the current node due to offline rebuilding,  add a check 
```
if hasInterruptibleAndFrontendDisabledTicket(currentAttachmentTickets) && hasWorkloadTicket(currentAttachmentTickets, longhorn.FalseValue)
```
So, if there is a workload ticket with `frontendDisable==false`,  we should do detachment.

derekbit:
Ready for review
cc @innobead @shuo-wu

innobead (approved):

innobead:
@mergify backport v1.5.x

controller/volume_attachment_controller.go
PhanLe1010: > !hasUninterruptibleTicket() racked my brain...:(

LOL. My brain is racked too but I couldn't find better name :D

PR #2023, opened 2023-06-21
Set v.Status.CurrentNodeID before successfully attaching to a node

derekbit:
Set v.Status.CurrentNodeID before successfully attaching to a node. There's a chance to detach the volume if the volume engine fails in volume attaching stage, so it won't be stuck at attaching stage.

Longhorn/longhorn#6155


derekbit:
The solution is not valid, I need to rethink it.

PhanLe1010:
This change would be the revert of the new AD logic back to old logic. Agree that this solution is not valid

PR #2026, opened 2023-06-24
Try to reattach volume if volume is detached or engine is dead unexpectedly

derekbit:
Reattach volume if
- volume is detached unexpectedly and there are still healthy replicas
- engine dead unexpectedly and there are still healthy replicas when the volume is not attached

Longhorn/longhorn#6155


innobead (approved):

innobead:
@mergify backport v1.5.x

PR #2042, opened 2023-06-29
6217 engine rebuilds deleted replica

ejweber:
longhorn/longhorn#6217

There are two fixes here. I will test each independently to confirm it fixes the issue, but would like to see both merged:

1. Don't delete rebuilding replicas until the engine is stopped when a volume detaches. This completely eliminates the recreate and avoids any awkward fail outs during engine reconciliation.
2. Check if an engine's `spec.desireState = stopped` before proceeding from snapshot purge to rebuild. This is kind of awkward, but it provides broader protection. There may be a way to trigger the issue different from the existing recreate, and this should help there.

### DIFF
--- a/controller/engine_controller.go
+++ b/controller/engine_controller.go
@@ -1759,19 +1759,16 @@ func (ec *EngineController) startRebuilding(e *longhorn.Engine, replicaName, add
 			for !purgeDone && time.Now().Before(endTime) {
 				<-ticker.C
 
+				// It may have been a long time since we started purging. Should we proceed?
 				e, err := ec.ds.GetEngineRO(e.Name)
 				if err != nil {
-					if apierrors.IsNotFound(err) {
-						log.Error("Failed to continue waiting for the purge before rebuilding since engine is not found")
-						return
-					}
-					log.WithError(err).Warn("Failed to get engine and wait for the purge before rebuilding")
-					continue
-				}
-				if e.Status.CurrentState != longhorn.InstanceStateRunning {
-					log.Errorf("Failed to continue waiting for the purge before rebuilding since engine is state %v", e.Status.CurrentState)
+					log.WithError(err).Error("Failed to get engine and wait for the purge before rebuilding")
 					return
 				}
+				if !shouldProceedToWaitAndRebuild(e, replicaName, addr, log) {
+					return
+				}
+
 				// Wait for purge complete
 				purgeDone = true
 				for _, purgeStatus := range e.Status.PurgeStatus {
@@ -1788,28 +1785,6 @@ func (ec *EngineController) startRebuilding(e *longhorn.Engine, replicaName, add
 					e.Spec.VolumeName, purgeWaitIntervalInSecond)
 				return
 			}
-		}
-
-		// It has been at least 10 seconds (2 * EnginePollInterval) since we were asked to rebuild.
-		// Should we still communicate with addr?
-		// TODO: Remove this check in versions that include a completed longhorn/longhorn#5845.
-		updatedEngine, err := ec.ds.GetEngineRO(e.Name)
-		if err != nil {
-			if apierrors.IsNotFound(err) {
-				log.Error("Failed to proceed to rebuild since engine is not found")
-				return
-			}
-			log.WithError(err).Error("Failed to get latest engine spec before rebuilding")
-			return
-		}
-		updatedAddr, ok := updatedEngine.Spec.ReplicaAddressMap[replicaName]
-		if !ok {
-			log.Warnf("Refusing to rebuild since replica %v is no longer in the replicaAddressMap", replicaName)
-			return
-		}
-		if addr != updatedAddr {
-			log.Warnf("Refusing to rebuild since the address for replica %v has been updated from %v to %v", replicaName, addr, updatedAddr)
-			return
 		}
 
 		replica, err := ec.ds.GetReplica(replicaName)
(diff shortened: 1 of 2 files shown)


controller/engine_controller.go
```diff
@@ -1790,28 +1787,6 @@ func (ec *EngineController) startRebuilding(e *longhorn.Engine, replicaName, add
...
-		updatedAddr, ok := updatedEngine.Spec.ReplicaAddressMap[replicaName]
-		if !ok {
-			log.Warnf("Refusing to rebuild since replica %v is no longer in the replicaAddressMap", replicaName)
-			return
-		}
-		if addr != updatedAddr {
-			log.Warnf("Refusing to rebuild since the address for replica %v has been updated from %v to %v", replicaName, addr, updatedAddr)
-			return
-		}
-
```
derekbit, line 1814: The commit moves the them to the earlier place and only checks the conditions when `autoCleanupSystemGeneratedSnapshot` is enabled. Should we check them regardless of the value of `autoCleanupSystemGeneratedSnapshot`?

controller/volume_controller.go
```diff
@@ -1773,30 +1773,6 @@ func (c *VolumeController) closeVolumeDependentResources(v *longhorn.Volume, e *
 		}
 	}
 
-	// check if any replica has been RW yet
```
shuo-wu, line 1776: 1. This function will be invoked when the volume will be updated to `detaching` or `detached`. Does this make sense?
2. Even if we ignore the context, the function name `closeXXXX` means we have to try our best to stop all volume-related resources including rebuilding replicas. Leaving such kind of rebuilding replica is not expected in terms of callers' perspective.

controller/engine_controller.go
shuo-wu: I am fine with it since **rebuilding** always requires that.

controller/volume_controller.go
ejweber: We still stop the rebuilding replica on 1816. The difference here is that we do so AFTER the engine stops, at the same time we stop the other replicas.

I am not 100% sure why we had logic to stop the rebuilding replica sooner. I assume it is because we thought that it was safe to do so (since it cannot lead to corruption like the comment on 1789 indicates). However, my analysis indicates that it can still cause this other issue, so IMO moving it makes sense.

WDYT?

controller/engine_controller.go
ejweber: My thinking is that the it is the 10 second snapshot purge wait that causes us the largest danger. If `autoCleanupSystemGenerateSnapshot = false`, it should be relatively little time between the start of an engine reconciliation and the moment we kick of a rebuild. Even if we do check, we can never be 100% sure the situation hasn't changed in the milliseconds between to lines of execution, so I confined the check to the situation we know can cause us problems.

I don't feel particularly strongly that we shouldn't also check when `autoCleanupGenerateSnapshot = false` if that's what the team wants. I just think the code is cleaner this way.

controller/volume_controller.go
shuo-wu: OK. Got it. Stopping replicas including rebuilding ones **after the engine stop** makes sense and it's supposed to be. In this perspective, we cannot stop rebuilding replicas first during detachment or cleanup.

shuo-wu (approved):

ejweber:
Fix 1 by itself survived 20 iterations of the test script before I stopped it.

Fix 2 by itself survived 25 iterations of the test script before I stopped it. I modified the script slightly to look for messages from shouldProceedToWaitAndRebuild and saw a decent number of the following log, indicating that it might have prevented an incorrect expansion.

```
time="2023-06-30T15:23:31Z" level=warning msg="Failed to proceed to rebuild since engine state should be stopped" controller=longhorn-engine engine=pvc-cc6872d7-1f32-4797-8dbe-1d2c1c8f5cde-e-005080ff node=eweber-v124-worker-1ae51dbb-4pngn volume=pvc-cc6872d7-1f32-4797-8dbe-1d2c1c8f5cde
time="2023-06-30T15:28:45Z" level=warning msg="Failed to proceed to rebuild since engine state should be stopped" controller=longhorn-engine engine=pvc-0821317e-9660-4848-bf36-96c3b835d198-e-860ac589 node=eweber-v124-worker-1ae51dbb-4pngn volume=pvc-0821317e-9660-4848-bf36-96c3b835d198
time="2023-06-30T15:34:09Z" level=warning msg="Failed to proceed to rebuild since engine state should be stopped" controller=longhorn-engine engine=pvc-b64adb3b-7483-467f-807f-34d6ec47007b-e-b9203198 node=eweber-v124-worker-1ae51dbb-4pngn volume=pvc-b64adb3b-7483-467f-807f-34d6ec47007b
```

ejweber:
Passing core test run: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/4387/.

ejweber:
Adding one more fix to this that I don't think will be particularly controversial (apologies to @shuo-wu, since you already reviewed). While running through more iterations of the script, I hit the following:

```
2023-06-29T20:48:24.810594355Z time="2023-06-29T20:48:24Z" level=error msg="Error syncing Longhorn engine" controller=longhorn-engine engine=longhorn-system/pvc-7cb92e9c-7b6f-430e-b3d9-544c42857335-e-6fa1620a error="failed to sync engine for longhorn-system/pvc-7cb92e9c-7b6f-430e-b3d9-544c42857335-e-6fa1620a: failed to start rebuild for pvc-7cb92e9c-7b6f-430e-b3d9-544c42857335-r-57896af0 of pvc-7cb92e9c-7b6f-430e-b3d9-544c42857335-e-6fa1620a: proxyServer=10.42.150.127:8501 destination=10.42.150.127:10035: failed to list replicas for volume: rpc error: code = Unknown desc = failed to list replicas for volume 10.42.150.127:10035: rpc error: code = FailedPrecondition desc = incorrect volume name pvc-7cb92e9c-7b6f-430e-b3d9-544c42857335; check controller address" node=eweber-v124-worker-1ae51dbb-4pngn
```

This is a related but slightly different issue to the one the previous fixes are targeting. It is even more rare, and it occurs when:

- The engine controller starts a reconciliation with `engine.Spec.DesireState = stopped` and a changed `engine.Spec.ReplicaAddressMap`.
- We stop the engine [here](https://github.com/longhorn/longhorn-manager/blob/4a57a7222edec14a2e7fcd84e2a0fa9e3fe141da/controller/engine_controller.go#L334-L336).
- We immediately proceed to rebuild a new replica [here](https://github.com/longhorn/longhorn-manager/blob/4a57a7222edec14a2e7fcd84e2a0fa9e3fe141da/controller/engine_controller.go#L352-L354).
- We don't actually rebuild the replica thanks to the newly introduced checks.
- While waiting, a different engine starts.
- We still check whether the replica was rebuilt [here](https://github.com/longhorn/longhorn-manager/blob/4a57a7222edec14a2e7fcd84e2a0fa9e3fe141da/controller/engine_controller.go#L1903), except we communicate with the wrong engine.

If the timing isn't exactly right, something else will happen. Additionally, this log indicates an erroneous ReplicaList with no actual action taken (i.e. no accidental expansion), so it's not particularly dangerous. Still, I think the right approach is to not call [ReconcileEngineState](https://github.com/longhorn/longhorn-manager/blob/4a57a7222edec14a2e7fcd84e2a0fa9e3fe141da/controller/engine_controller.go#L352-L354) if we know we are killing the engine.

ejweber:
@mergifyio backport v1.5.x

PhanLe1010 (approved):
LGTM!

Thanks for the great fix

innobead:
@derekbit @shuo-wu please help review this. I will check this later. Thanks.

innobead (approved):
In general, LGTM.

It seems best efforts to try to prevent rebuilding with the a wrong replica. Surely, it should reduce chances to rebuild with a wrong replica. Good prevention. 

Howecer, I am still confused of why a wrong replica would happen here when the engine is still around.

derekbit (approved):
LGTM

innobead:

ejweber:
> Howecer, I am still confused of why a wrong replica would happen here when the engine is still around.

@innobead 

Before these changes, our detach operation deletes rebuilding replicas and the engine in the same volume controller reconcile loop. Replica deletion is generally pretty much instantaneous, but engine deletion may take 30 seconds or more under rare circumstances (because we wait 30 seconds within an engine reconcile loop for the snapshot purge and engine rebuild).

Bad timing:
1. We just started to purge snapshots and are waiting to instruct the engine to rebuild.
2. We want to detach, so we delete the rebuilding replica and the engine.
3. The rebuilding replica stops.
4. A completely different replica takes its "place" (port).
5. The engine does not stop, because the previous reconciliation is ongoing.
6. 10 seconds after starting the snapshot purge, we rebuild the replica, not knowing that our information is out-of-date.

On the next reconcile we stop the engine as instructed, but the damage is done.

PR #2043, opened 2023-07-03
fix(csi): Invalid mount point with volume detached

mantissahz:
Volume detached because only one replica failed and not remount the `globalmount`. We need to handle it at NodePublishVolume because os.ReadDir will be passed when it becomes a read-only file system after an unexpected volume detaching.

Ref: longhorn/longhorn#4814, longhorn/longhorn#5801

Test case test:
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/4403/
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/4425/

### DIFF
--- a/csi/node_server.go
+++ b/csi/node_server.go
@@ -126,7 +126,7 @@ func (ns *NodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 	if volume.State != string(longhorn.VolumeStateAttached) || volume.Controllers[0].Endpoint == "" {
 		log.Infof("Volume %v hasn't been attached yet, unmounting potential mount point %v", volumeID, targetPath)
 		if err := unmount(targetPath, mounter); err != nil {
-			log.WithError(err).Warn("Failed to unmount")
+			log.WithError(err).Warnf("Failed to unmount: %v", targetPath)
 		}
 		return nil, status.Errorf(codes.InvalidArgument, "volume %s hasn't been attached yet", volumeID)
 	}
--- a/csi/util.go
+++ b/csi/util.go
@@ -4,7 +4,9 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"io"
 	"os"
+	"path"
 	"strconv"
 	"strings"
 	"time"
@@ -29,6 +31,8 @@ const (
 	defaultStaleReplicaTimeout = 2880
 
 	defaultForceUmountTimeout = 30 * time.Second
+
+	tempTestMountPointValidStatusFile = ".longhorn-volume-mount-point-test.tmp"
 )
 
 // NewForcedParamsExec creates a osExecutor that allows for adding additional params to later occurring Run calls
@@ -239,6 +243,42 @@ func parseJSONRecurringJobs(jsonRecurringJobs string) ([]longhornclient.Recurrin
 	return recurringJobs, nil
 }
 
+func syncMountPointDirectory(targetPath string) error {
+	d, err := os.OpenFile(targetPath, os.O_SYNC, 0750)
+	if err != nil {
+		return err
+	}
+	defer d.Close()
+
+	if _, err := d.Readdirnames(1); err != nil {
+		if !errors.Is(err, io.EOF) {
+			return err
+		}
+	}
+
+	// it would not always return `Input/Output Error` or `read-only file system` errors if we only use ReadDir() or Readdirnames() without an I/O operation
+	// an I/O operation will make the targetPath mount point invalid immediately
+	tempFile := path.Join(targetPath, tempTestMountPointValidStatusFile)
+	f, err := os.OpenFile(tempFile, os.O_CREATE|os.O_RDWR|os.O_SYNC|os.O_TRUNC, 0666)
+	if err != nil {
+		return err
+	}
+	defer func() {
+		if err := f.Close(); err != nil {
+			logrus.WithError(err).Warnf("Failed to close %v", tempFile)
+		}
+		if err := os.Remove(tempFile); err != nil {
+			logrus.WithError(err).Warnf("Failed to remove %v", tempFile)
+		}
+	}()
+
+	if _, err := f.WriteString(tempFile); err != nil {
+		return err
+	}
+
+	return nil
+}
+
 // ensureMountPoint evaluates whether a path is a valid mountPoint
 // in case the targetPath does not exists it will create a path and return false
 // in case where the mount point exists but is corrupt, the mount point will be cleaned up and a error is returned
@@ -252,9 +292,9 @@ func ensureMountPoint(targetPath string, mounter mount.Interface) (bool, error) 
 
 	IsCorruptedMnt := mount.IsCorruptedMnt(err)
 	if !IsCorruptedMnt {
-		logrus.Debugf("Mount point %v try reading dir to make sure it's healthy", targetPath)
-		if _, err := os.ReadDir(targetPath); err != nil {
-			logrus.Debugf("Mount point %v was identified as corrupt by ReadDir", targetPath)
+		logrus.Debugf("Mount point %v try opening and syncing dir to make sure it's healthy", targetPath)
+		if err := syncMountPointDirectory(targetPath); err != nil {
+			logrus.WithError(err).Warnf("Mount point %v was identified as corrupt by opening and syncing", targetPath)
 			IsCorruptedMnt = true
 		}
 	}


csi/util.go
```diff
@@ -252,9 +270,9 @@ func ensureMountPoint(targetPath string, mounter mount.Interface) (bool, error)
 
 	IsCorruptedMnt := mount.IsCorruptedMnt(err)
 	if !IsCorruptedMnt {
-		logrus.Debugf("mount point %v try reading dir to make sure it's healthy", targetPath)
-		if _, err := os.ReadDir(targetPath); err != nil {
```
shuo-wu, line 256: Can't this detect a corrupted fs?

shuo-wu (approved):
Others LGTM

csi/util.go
mantissahz: When volume detached and re-attached and pod restarts quickly, it still can read the directory as normal.
OS also considers the mount point is valid at that moment.

mantissahz: But yes, it could detect a corrupted fs after mount point became invalid.

csi/util.go
```diff
@@ -239,6 +239,24 @@ func parseJSONRecurringJobs(jsonRecurringJobs string) ([]longhornclient.Recurrin
...
+	}
+	defer d.Close()
+
+	if _, err := d.Readdirnames(-1); err != nil {
+		return err
+	}
+
+	if err := d.Sync(); err != nil {
+		return err
+	}
```
derekbit, line 255: Why need the `Sync` operation?

mantissahz: It needs some I/O to trigger the mount point to be invalid.
Almost `I/O error` errors return from this `Sync()` method.

innobead: Quite confusing about the operations here. Also, need some clarification.

csi/util.go
```diff
@@ -252,9 +270,9 @@ func ensureMountPoint(targetPath string, mounter mount.Interface) (bool, error)
 
 	IsCorruptedMnt := mount.IsCorruptedMnt(err)
 	if !IsCorruptedMnt {
-		logrus.Debugf("Mount point %v try reading dir to make sure it's healthy", targetPath)
-		if _, err := os.ReadDir(targetPath); err != nil {
-			logrus.Debugf("Mount point %v was identified as corrupt by ReadDir", targetPath)
+		logrus.Debugf("Mount point %v try opening and syncing dir  to make sure it's healthy", targetPath)
+		if err := syncMountPointDirectory(targetPath); err != nil {
+			logrus.WithError(err).Debugf("Mount point %v was identified as corrupt by opening and syncing", targetPath)
```
derekbit, line 275: Is `Warnf` better?

```suggestion
			logrus.WithError(err).Warnf("Mount point %v was identified as corrupt by opening and syncing", targetPath)
```

csi/util.go
derekbit: > It needs some I/O to trigger the mount point to be invalid. Almost `I/O error` errors return from this `Sync()` method.

Can't `Readdirnames` return IO error if the mountpoint is broken?

mantissahz: It needs some I/Os to make the mount point invalid when volume detached unexpectedly.
1. Readdirnames(-1) would make mount point invalid. (it would return `I/O error` if there are other I/Os before this metohd.)
2. Sync() would return `I/O error`

derekbit: I see. You probably can refer to `ensureMountpoint()`. The function checks the availability of a mountpoint by `os.ReadDir`. Using the same method can make the codes more consistent.

derekbit: I see. You probably can refer to ensureMountpoint(). The function checks the availability of a mountpoint by os.ReadDir. Using the same method can make the codes more consistent.

derekbit: Another question:

> Readdirnames(-1) would make mount point invalid.

What error will be returned if mount point is invalid or become invalid when executing `Readdirnames`?

innobead: Besides @derekbit questions, I am just concerned if any side effects here. Or should we worry about that?

mantissahz: @derekbit IIRC it would return `Input/output error`, Sync() would returns `Input/output error` or `read-only file system`
@innobead We might be able to keep `os.RedaDir` as `Readdirnames`, and do `Open/Sync` after `os.ReadDir`. I am trying to test there operations.

csi/util.go
shuo-wu: OK. Got it. You need to trigger IO write via `sync` later hence it's better to directly open the path then read the dir via the fd.

csi/util.go
derekbit: @mantissahz
I'm fine with the implementation but still confused about why need extra `Sync()` after `Readdirnames`. Making mount point invalid sounds a bit weird.

shuo-wu (approved):
LGTM. After addressing Derek's comments, we can merge it

csi/util.go
shuo-wu: I guess it's for triggering read (`Readdirnames`) and write `Sync` ops. `Readdirnames` may not be able to trigger write op.

csi/util.go
mantissahz: Modified to be `Warnf`. Thanks

csi/util.go
derekbit: After discussing @mantissahz, the `os. Readdirnames` still works after detachment, and it is might be due to the cache effect. `Sync()`. Afterwards, using `Sync()` can trigger IO error.
Theoretically, using single `Sync()` without `os. Readdirnames` should be okay. However, according to @mantissahz's test, using single `Sync()` doesn't work. We can investigate it later and add a comment and TODO here.

derekbit (approved):
LGTM

csi/util.go
mantissahz: It would not return an error using only `Sync` without `ReadDir` or `Readdirnames` in my testings. It is a weird thing for me.
@derekbit Added comments for using two system calls.
Thanks

csi/node_server.go
```diff
@@ -126,7 +126,11 @@ func (ns *NodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 	if volume.State != string(longhorn.VolumeStateAttached) || volume.Controllers[0].Endpoint == "" {
 		log.Infof("Volume %v hasn't been attached yet, unmounting potential mount point %v", volumeID, targetPath)
 		if err := unmount(targetPath, mounter); err != nil {
-			log.WithError(err).Warn("Failed to unmount")
+			log.WithError(err).Warnf("Failed to unmount: %v", targetPath)
+		}
+		log.Infof("Volume %v hasn't been attached yet, unmounting potential mount point %v", volumeID, stagingPath)
+		if err := unmount(stagingPath, mounter); err != nil {
+			log.WithError(err).Warnf("Failed to unmount %v", stagingPath)
```
PhanLe1010, line 133: Do we really need to unmount here? If the stagingPath is corrupted, I think the NodePublishVolume() call already handle it, right?

csi/util.go
```diff
@@ -239,6 +239,26 @@ func parseJSONRecurringJobs(jsonRecurringJobs string) ([]longhornclient.Recurrin
...
 }
 
+func syncMountPointDirectory(targetPath string) error {
+	d, err := os.Open(targetPath)
+	if err != nil {
+		return err
+	}
+	defer d.Close()
+
+	if _, err := d.Readdirnames(-1); err != nil {
```
PhanLe1010, line 249: `-1` means Readdirnames returns all file names all the names from the directory targetPath. It would be time-consuming if the targetPath directory contains a lot of file under it. What about using `d.Readdirnames(1)` or the old `os.ReadDir(targetPath)` ?

csi/util.go
```diff
@@ -239,6 +239,26 @@ func parseJSONRecurringJobs(jsonRecurringJobs string) ([]longhornclient.Recurrin
...
+		return err
+	}
+
+	// it would not always return `Input/Output Error` or `read-only file system` errors if we only use Sync() without the previous I/O operation
+	// and it needs more investigations why only Sync() would not return an error
+	if err := d.Sync(); err != nil {
+		return err
+	}
+
+	return d.Close()
```
c3y1huang, line 259: Isn't this already handled by the [defer d.Close()](https://github.com/longhorn/longhorn-manager/blob/aae8aac5e8d1321670b461ede38436f470418532/csi/util.go#L247)?

csi/util.go
```diff
@@ -252,9 +272,9 @@ func ensureMountPoint(targetPath string, mounter mount.Interface) (bool, error)
 
 	IsCorruptedMnt := mount.IsCorruptedMnt(err)
 	if !IsCorruptedMnt {
-		logrus.Debugf("Mount point %v try reading dir to make sure it's healthy", targetPath)
-		if _, err := os.ReadDir(targetPath); err != nil {
-			logrus.Debugf("Mount point %v was identified as corrupt by ReadDir", targetPath)
+		logrus.Debugf("Mount point %v try opening and syncing dir  to make sure it's healthy", targetPath)
```
c3y1huang, line 275:
```suggestion
		logrus.Debugf("Mount point %v try opening and syncing dir to make sure it's healthy", targetPath)
```

c3y1huang (changes requested):

innobead:
@mantissahz Please help check the remaining reviews from Chinya and Phan.

csi/util.go
mantissahz: Modified. Thanks

csi/util.go
mantissahz: Yes, you are right. Modified. Thanks

csi/util.go
mantissahz: Modified to "1", thanks

csi/node_server.go
mantissahz: Yes, removed it. Thanks

csi/util.go
```diff
@@ -239,6 +239,26 @@ func parseJSONRecurringJobs(jsonRecurringJobs string) ([]longhornclient.Recurrin
...
+
+	if _, err := d.Readdirnames(-1); err != nil {
+		return err
+	}
+
+	// it would not always return `Input/Output Error` or `read-only file system` errors if we only use Sync() without the previous I/O operation
+	// and it needs more investigations why only Sync() would not return an error
+	if err := d.Sync(); err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
mantissahz: Have the same concern as well. Modified to create an empty file and remove it to check if the mount point is valid.
Thanks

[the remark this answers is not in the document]
PhanLe1010: The `return f.Sync()` function below is still sync the host filesystem, right? Can we remove the sync and still be able to detect the IO error?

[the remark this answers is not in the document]
mantissahz: No, it only synchronize the temp file just created.
I don't have any other idea that how could it detect the IO error.
Any good suggestion?

csi/util.go
```diff
@@ -239,6 +242,33 @@ func parseJSONRecurringJobs(jsonRecurringJobs string) ([]longhornclient.Recurrin
...
+	if err != nil {
+		return err
+	}
+	if _, err := d.Readdirnames(1); err != nil {
+		if !errors.Is(err, io.EOF) {
+			d.Close()
+			return err
+		}
+	}
+	if err := d.Close(); err != nil {
```
innobead, line 256: nit: you can actually do the defer d.close after it has been successfully opened, but return error in the defer function if you want.

csi/util.go
```diff
@@ -239,6 +242,33 @@ func parseJSONRecurringJobs(jsonRecurringJobs string) ([]longhornclient.Recurrin
...
+
+	// it would not always return `Input/Output Error` or `read-only file system` errors if we only use ReadDir() or Readdirnames() without an I/O operation
+	// an I/O operation will make the targetPath mount point invalid immediately
+	tempFile := targetPath + tempTestMountPointValidStatusFile
+	f, err := os.OpenFile(tempFile, os.O_CREATE|os.O_RDWR|os.O_SYNC|os.O_TRUNC, 0666)
+	if err != nil {
+		return err
+	}
+	defer f.Close()
+	defer os.Remove(tempFile)
```
innobead, line 268: so here the `os.Remove(tempFile) will be first to deter, does it matter?` or just do

defer fun() {
    _ = f.Close()
    os.Remove(tempFile)
}

csi/util.go
```diff
@@ -239,6 +242,33 @@ func parseJSONRecurringJobs(jsonRecurringJobs string) ([]longhornclient.Recurrin
...
+		}
+	}
+	if err := d.Close(); err != nil {
+		return err
+	}
+
+	// it would not always return `Input/Output Error` or `read-only file system` errors if we only use ReadDir() or Readdirnames() without an I/O operation
+	// an I/O operation will make the targetPath mount point invalid immediately
+	tempFile := targetPath + tempTestMountPointValidStatusFile
+	f, err := os.OpenFile(tempFile, os.O_CREATE|os.O_RDWR|os.O_SYNC|os.O_TRUNC, 0666)
```
innobead, line 263: For the option, although the temp file is invalid, can we just use `O_SYNC` to do the same effect? so don't need to really do a dummy creation?

innobead (commented):
In general, LGTM.

James, well done, but a few comments need you help to check.

csi/util.go
mantissahz: Without `O_CREATE` it will return `no such file or directory` error and following operations will not be done.

csi/util.go
mantissahz: Testing now

csi/util.go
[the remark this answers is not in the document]
shuo-wu: What if there are lots of caches that haven't been flushed before the volume crash? Will this late sync lead to any unexpected issues?

As I mentioned before, can we try to use the direct IO libs in `sparse-tools` to modify the disk meta-file instead?

csi/util.go
innobead: Thanks to clarify that.

innobead:
@mergify backport v1.5.x

csi/util.go
mantissahz: Modified. Thanks

csi/util.go
mantissahz: Modified. Thanks

innobead:
It seems the PR is not forcedly updated yet, so after it's updated, I will approve this. 

@derekbit @shuo-wu double check this after @mantissahz updates the PR.

csi/util.go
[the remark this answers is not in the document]
mantissahz: Do you mean access the volume directly instead of accessing the file system?

innobead (dismissed):

csi/util.go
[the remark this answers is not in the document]
derekbit: If the filesystem is somehow in `readonly` state, will the error lead to any side effect?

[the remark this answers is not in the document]
mantissahz: It might get into the `NodeStageVolume/NodeUnStageVolume` loop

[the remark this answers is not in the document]
derekbit: I see. Should we ignore the error and make it as best-effort instead?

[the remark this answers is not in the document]
mantissahz: ignore `read-only file system`? let me do some testings for ignoring this error.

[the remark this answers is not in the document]
derekbit: @mantissahz 
After going through the usage of `ensureMountPoint()`, I think returning error is fine and no need to ignore it.

derekbit (approved):
LGTM

csi/util.go
```diff
@@ -239,6 +242,34 @@ func parseJSONRecurringJobs(jsonRecurringJobs string) ([]longhornclient.Recurrin
...
+	if err != nil {
+		return err
+	}
+	defer func() {
+		f.Close()
```
c3y1huang, line 266: Should output the error.

csi/util.go
```diff
@@ -239,6 +242,34 @@ func parseJSONRecurringJobs(jsonRecurringJobs string) ([]longhornclient.Recurrin
...
+		os.Remove(tempFile)
```
c3y1huang, line 267: Suggest outputting the error.

csi/util.go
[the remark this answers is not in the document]
mantissahz: @derekbit files in the directory of the mount point `globalmount` with the `read-only file system` error for opening a file would still return `Input/output error` when accessing.

csi/util.go
mantissahz: Modified to log the error. Thanks

csi/util.go
mantissahz: Modified to log the error. Thanks

c3y1huang (dismissed):

derekbit (dismissed):

csi/util.go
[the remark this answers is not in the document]
shuo-wu: Do you mean that there is no need to write anything or run `sync` to trigger the IO error if the fs is corrupted?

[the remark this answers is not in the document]
mantissahz: No. It still needs some I/Os to trigger I/O error. What I meant is files might not be accessible if they are in a read-only file system for the situation of this issue.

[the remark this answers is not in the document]
shuo-wu: Then what I mentioned above may be applicable:
You can directly open/create the disk meta file `<disk mount point or data path>/longhorn-disk.cfg` directly launch a new file like `<disk mount point or data path>/longhorn-disk.tmp`.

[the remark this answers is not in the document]
mantissahz: Yes, now I create a temp empty file in the mount point directory to test the file system and remove it after testing without `sparse-tools` libs but `OpenFile` with the `O_SYNC` flag instead of synchronizing the whole file system.
@shuo-wu Does it look like what you mentioned?

csi/util.go
```diff
@@ -239,6 +242,38 @@ func parseJSONRecurringJobs(jsonRecurringJobs string) ([]longhornclient.Recurrin
...
+
+	if _, err := d.Readdirnames(1); err != nil {
+		if !errors.Is(err, io.EOF) {
+			return err
+		}
+	}
+
+	// it would not always return `Input/Output Error` or `read-only file system` errors if we only use ReadDir() or Readdirnames() without an I/O operation
+	// an I/O operation will make the targetPath mount point invalid immediately
+	tempFile := targetPath + tempTestMountPointValidStatusFile
```
shuo-wu, line 260: NIT: `path.Join()`

csi/util.go
```diff
@@ -239,6 +242,38 @@ func parseJSONRecurringJobs(jsonRecurringJobs string) ([]longhornclient.Recurrin
...
+	defer func() {
+		if err := f.Close(); err != nil {
+			logrus.WithError(err).Warnf("Failed to close %v", tempFile)
+		}
+		if err := os.Remove(tempFile); err != nil {
+			logrus.WithError(err).Warnf("Failed to remove %v", tempFile)
+		}
+	}()
+
+	return f.Sync()
```
shuo-wu, line 274: NIT: Do we still need this after using `O_SYNC`?

csi/util.go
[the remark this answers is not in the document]
shuo-wu: I think it's good enough to use `O_SYNC`.

csi/util.go
mantissahz: Modified. Thanks

csi/util.go
mantissahz: After some testing, it could not still detect the corrupted fs if there is no `f.Sync()` with the flag `O_SYNC`

PhanLe1010: Looks like to detect a bad mount point, we cannot avoid `Sync()` so we will have to live with the fact that `Sync()` might take a long time when there is a lot of cache. So let's use it directly on the directory `d` instead. WDYT @shuo-wu @derekbit @mantissahz ? 

```
func syncMountPointDirectory(targetPath string) error {
	d, err := os.OpenFile(targetPath, os.O_SYNC, 0750)
	if err != nil {
		return err
	}
	defer d.Close()

	if _, err := d.Readdirnames(1); err != nil {
		if !errors.Is(err, io.EOF) {
			return err
		}
	}

	// it would not always return `Input/Output Error` or `read-only file system` errors if we only use ReadDir() or Readdirnames() without an I/O operation
	// an I/O operation will make the targetPath mount point invalid immediately
	return d.Sync()
```

mantissahz: Testing that use `NewDirectFileIoProcessor` from `sparse-tools` libs without `Sync()` to open a file.

mantissahz: It looks good to use `NewDirectFileIoProcessor` from sparse-tools libs to open/close/remove a file without `Sync()`.
A low chance will run to `WriteString` to trigger `Input/Output error`, almost it returns the `Input/Output error/read-only file system` errors at `Open`

mantissahz: Use `WriteString` instead of `Sync()`

innobead (approved):

innobead:

mantissahz:

PR #2048, opened 2023-07-04
Support encrypted volumes in block mode

derekbit:
Longhorn/longhorn#4883

### DIFF
--- a/csi/node_server.go
+++ b/csi/node_server.go
@@ -56,6 +56,7 @@ type NodeServer struct {
 	apiClient *longhornclient.RancherClient
 	nodeID    string
 	caps      []*csi.NodeServiceCapability
+	log       *logrus.Entry
 }
 
 func NewNodeServer(apiClient *longhornclient.RancherClient, nodeID string) *NodeServer {
@@ -68,17 +69,13 @@ func NewNodeServer(apiClient *longhornclient.RancherClient, nodeID string) *Node
 				csi.NodeServiceCapability_RPC_STAGE_UNSTAGE_VOLUME,
 				csi.NodeServiceCapability_RPC_EXPAND_VOLUME,
 			}),
+		log: logrus.StandardLogger().WithField("component", "csi-node-server"),
 	}
-}
-
-func getLoggerForCSINodeServer() *logrus.Entry {
-	return logrus.StandardLogger().WithField("component", "csi-node-server")
 }
 
 // NodePublishVolume will mount the volume /dev/longhorn/<volume_name> to target_path
 func (ns *NodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublishVolumeRequest) (*csi.NodePublishVolumeResponse, error) {
-	log := getLoggerForCSINodeServer()
-	log = log.WithFields(logrus.Fields{"function": "NodePublishVolume"})
+	log := ns.log.WithFields(logrus.Fields{"function": "NodePublishVolume"})
 
 	targetPath := req.GetTargetPath()
 	if targetPath == "" {
@@ -140,7 +137,17 @@ func (ns *NodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 	}
 
 	if volumeCapability.GetBlock() != nil {
-		devicePath := volume.Controllers[0].Endpoint
+		devicePath := getStageBlockVolumePath(stagingTargetPath, volumeID)
+		_, err := os.Stat(devicePath)
+		if err != nil {
+			if !os.IsNotExist(err) {
+				return nil, status.Errorf(codes.Internal, errors.Wrapf(err, "failed to stat device %s", devicePath).Error())
+			}
+			// Fall back to the controller endpoint if the device path under the stagingTargetPath doesn't exist
+			log.Infof("Device path %s doesn't exist, falling back to controller endpoint %s", devicePath, volume.Controllers[0].Endpoint)
+			devicePath = volume.Controllers[0].Endpoint
+		}
+
 		if err := ns.nodePublishBlockVolume(volumeID, devicePath, targetPath, mounter); err != nil {
 			log.WithError(err).Errorf("Failed to publish BlockVolume %s", volumeID)
 			return nil, err
@@ -152,7 +159,7 @@ func (ns *NodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 
 	// we validate the staging path to make sure the global mount is still valid
 	if isMnt, err := ensureMountPoint(stagingTargetPath, mounter); err != nil || !isMnt {
-		msg := fmt.Sprintf("Staging path %v is no longer valid for volume %v", stagingTargetPath, volumeID)
+		msg := fmt.Sprintf("Staging target path %v is no longer valid for volume %v", stagingTargetPath, volumeID)
 		log.WithError(err).Error(msg)
 
 		// HACK: normally when we return FailedPrecondition below kubelet should call NodeStageVolume again
@@ -249,12 +256,11 @@ func (ns *NodeServer) nodeStageSharedVolume(volumeID, shareEndpoint, targetPath 
 }
 
 func (ns *NodeServer) nodeStageMountVolume(volumeID, devicePath, stagingTargetPath, fsType string, mountFlags []string, mounter *mount.SafeFormatAndMount) error {
-	log := getLoggerForCSINodeServer()
-	log = log.WithFields(logrus.Fields{"function": "NodePublishVolume"})
+	log := ns.log.WithFields(logrus.Fields{"function": "NodePublishVolume"})
 
 	isMnt, err := ensureMountPoint(stagingTargetPath, mounter)
 	if err != nil {
-		return status.Errorf(codes.Internal, errors.Wrapf(err, "failed to prepare mount point for volume %v", volumeID).Error())
+		return status.Errorf(codes.Internal, errors.Wrapf(err, "failed to prepare mount point %v for volume %v", stagingTargetPath, volumeID).Error())
 	}
 	if isMnt {
 		return nil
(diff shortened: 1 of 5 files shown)


csi/node_server.go
```diff
@@ -456,6 +456,26 @@ func (ns *NodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
 		devicePath = cryptoDevice
 	}
 
+	if volumeCapability.GetBlock() != nil {
+		if err := ns.nodeStageBlockVolume(volumeID, devicePath, targetPath, mounter); err != nil {
```
shuo-wu, line 460: NIT: Just found that the staging path is named `targetPath` in this function. I am thinking if it's better to rename it

csi/node_server.go
```diff
@@ -500,16 +520,29 @@ func (ns *NodeServer) NodeUnstageVolume(ctx context.Context, req *csi.NodeUnstag
 		return nil, status.Error(codes.InvalidArgument, "volume id missing in request")
 	}
 
+	mounter := mount.New("")
+
 	// CO owns the staging_path so we only unmount but not remove the path
-	if err := unmount(targetPath, mount.New("")); err != nil {
-		return nil, status.Error(codes.Internal, fmt.Sprintf("failed to unmount volume %s mount point %v error %v", volumeID, targetPath, err))
+	if err := unmount(targetPath, mounter); err != nil {
+		return nil, status.Error(codes.Internal, errors.Wrapf(err, "failed to unmount volume %s mount point %v", volumeID, targetPath).Error())
```
shuo-wu, line 527: For block volumes, `targetPath` is the parent dir instead then umounting the dir will error out. Hence, should we try the below umounting when `err != nil`?

derekbit: If `targetPath` is not a mount point, the `umount` doesn't work and the error `nil` will be returned.
https://github.com/longhorn/longhorn-manager/blob/master/csi/util.go#L290

csi/node_server.go
derekbit: Rename to `stagingTargetPath` instead. Thank you.

shuo-wu (dismissed):

innobead:
@james-munson @mantissahz @ChanYiLin please help release this, so you can learn more about the current encryption mechanism. (@mantissahz has done some work for that before for extra encryption options)

ChanYiLin (dismissed):

csi/node_server.go
```diff
@@ -246,44 +258,58 @@ func (ns *NodeServer) nodeStageSharedVolume(volumeID, shareEndpoint, targetPath
...
 	}
 	return nil
 }
 
-func (ns *NodeServer) nodePublishBlockVolume(volumeID, devicePath, targetPath string, mounter mount.Interface) error {
+// nodeStageBlockVolume utilizes the stagingTargetPath to create a volumeID file to bind mount the devicePath
+// this is valid since the csi plugin is in control of the staging path
+func (ns *NodeServer) nodeStageBlockVolume(volumeID, devicePath, stagingTargetPath string, mounter mount.Interface) error {
+	stagingTargetPath = filepath.Join(stagingTargetPath, volumeID)
+	return ns.nodePublishBlockVolume(volumeID, devicePath, stagingTargetPath, mounter)
```
innobead, line 284: Have `getStageBlockVolumePath` instead?

```go
func getStageBlockVolumePath(stagingTargetPath, volumeID string) string {
    filepath.Join(stagingTargetPath, volumeID)
}
```

csi/node_server.go
```diff
@@ -136,8 +140,19 @@ func (ns *NodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 	}
 
 	if volumeCapability.GetBlock() != nil {
-		devicePath := volume.Controllers[0].Endpoint
+		devicePath := filepath.Join(stagingTargetPath, volumeID)
+		_, err := os.Stat(devicePath)
```
innobead, line 144: Why use `stagingTargetPath` + `volumeID` instead of `stagingTargetPath` only? Shouldn't the stage path be still only for the volume?

innobead (changes requested):
The log improvement is great. 

In general, LGTM. However, some feedback and comments need to resolve first.

csi/node_server.go
```diff
@@ -246,44 +248,48 @@ func (ns *NodeServer) nodeStageSharedVolume(volumeID, shareEndpoint, targetPath
...
 		return status.Error(codes.Internal, err.Error())
 	}
 	return nil
 }
 
 func (ns *NodeServer) nodePublishBlockVolume(volumeID, devicePath, targetPath string, mounter mount.Interface) error {
 	// we ensure the parent directory exists and is valid
-	if _, err := ensureMountPoint(filepath.Dir(targetPath), mounter); err != nil {
-		return status.Errorf(codes.Internal, "Failed to prepare mount point for block device %v error %v", devicePath, err)
+	if _, err := ensureMountPoint(filepath.Dir(stagingTargetPath), mounter); err != nil {
```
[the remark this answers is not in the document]
derekbit: Hmm, hard to say it is incorrect, because the path is sometimes a targetPath and sometimes stagingTargetPath. Anyway, will revert to `targetPath`.

csi/node_server.go
```diff
@@ -526,7 +536,7 @@ func (ns *NodeServer) NodeGetVolumeStats(ctx context.Context, req *csi.NodeGetVo
 
 	existVol, err := ns.apiClient.Volume.ById(volumeID)
 	if err != nil {
-		return nil, status.Error(codes.Internal, err.Error())
+		return nil, status.Error(codes.Internal, errors.Wrapf(err, "failed to get volume %s for volume statistics", volumeID).Error())
```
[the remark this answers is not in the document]
derekbit: Addressed. Thank you.

csi/node_server.go
```diff
@@ -336,20 +342,24 @@ func (ns *NodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
...
+		return nil, status.Errorf(codes.InvalidArgument, "volume %s has invalid frontend type %v", volumeID, volume.Frontend)
 	}
 
 	// Check volume attachment status
 	if volume.State != string(longhorn.VolumeStateAttached) || volume.Controllers[0].Endpoint == "" {
-		log.Infof("Volume %v hasn't been attached yet, try unmounting potential mount point %v", volumeID, targetPath)
-		if err := unmount(targetPath, mounter); err != nil {
-			log.Warnf("Failed to unmount error: %v", err)
+		log.Infof("Volume %v hasn't been attached yet, unmounting potential mount point %v", volumeID, stagingTargetPath)
+		if err := unmount(stagingTargetPath, mounter); err != nil {
```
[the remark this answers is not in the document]
derekbit: Addressed. Thank you.

csi/node_server.go
```diff
@@ -246,44 +248,48 @@ func (ns *NodeServer) nodeStageSharedVolume(volumeID, shareEndpoint, targetPath
 	return nil
 }
 
-func (ns *NodeServer) nodeStageMountVolume(volumeID, devicePath, targetPath, fsType string, mountFlags []string, mounter *mount.SafeFormatAndMount) error {
-	isMnt, err := ensureMountPoint(targetPath, mounter)
+func (ns *NodeServer) nodeStageMountVolume(volumeID, devicePath, stagingTargetPath, fsType string, mountFlags []string, mounter *mount.SafeFormatAndMount) error {
+	log := getLoggerForCSINodeServer()
+	log = log.WithFields(logrus.Fields{"function": "NodePublishVolume"})
```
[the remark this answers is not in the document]
derekbit: Will handled in a separate ticket.

csi/node_server.go
```diff
@@ -246,44 +248,48 @@ func (ns *NodeServer) nodeStageSharedVolume(volumeID, shareEndpoint, targetPath
 	return nil
 }
 
-func (ns *NodeServer) nodeStageMountVolume(volumeID, devicePath, targetPath, fsType string, mountFlags []string, mounter *mount.SafeFormatAndMount) error {
-	isMnt, err := ensureMountPoint(targetPath, mounter)
+func (ns *NodeServer) nodeStageMountVolume(volumeID, devicePath, stagingTargetPath, fsType string, mountFlags []string, mounter *mount.SafeFormatAndMount) error {
+	log := getLoggerForCSINodeServer()
```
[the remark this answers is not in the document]
derekbit: Updated. Thank you.

csi/node_server.go
```diff
@@ -474,29 +507,42 @@ func (ns *NodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
...
+		return nil, status.Error(codes.Internal, errors.Wrapf(err, "failed to clean up volume %s device mount point %v", volumeID, deviceFilePath).Error())
 	}
 
 	// optionally try to retrieve the volume and check if it's an RWX volume
 	// if it is we let the share-manager clean up the crypto device
 	volume, _ := ns.apiClient.Volume.ById(volumeID)
-	cleanupCryptoDevice := !requiresSharedAccess(volume, nil)
 
+	// Currently, only "RWO volumes" and "block device with volume.Migratable is true" supports encryption.
+	cleanupCryptoDevice := !requiresSharedAccess(volume, nil) || (volume != nil && volume.Migratable)
```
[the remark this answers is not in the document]
derekbit: Oh, yes. The check is redundant. Thank you.

innobead:
@derekbit after resolving the conflict, I will do a quick review before merging.

derekbit:
> @derekbit after resolving the conflict, I will do a quick review before merging.

Rebased. Thank you.

csi/node_server.go
```diff
@@ -474,29 +500,41 @@ func (ns *NodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
...
 	if len(volumeID) == 0 {
 		return nil, status.Error(codes.InvalidArgument, "volume id missing in request")
 	}
 
+	mounter := mount.New("")
+
 	// CO owns the staging_path so we only unmount but not remove the path
-	if err := unmount(targetPath, mount.New("")); err != nil {
-		return nil, status.Error(codes.Internal, fmt.Sprintf("failed to unmount volume %s mount point %v error %v", volumeID, targetPath, err))
+	if err := unmount(stagingTargetPath, mounter); err != nil {
```
innobead, line 518: If it's an encrypted block volume, do we still need to unmount `stagingTargetPath`? or just `getStageBlockVolumePath(stagingTargetPath, volumeID)` only?

derekbit: `unmount` is a no op for a block mode volume.
https://github.com/longhorn/longhorn-manager/blob/master/csi/util.go#L338

innobead (approved):

PR #2049, opened 2023-07-06
Fix strict-local volume restore stuck in attaching state

derekbit:
For a volume being restored, the spec.nodeID is empyty but the status.ownerID is set. Thus, set r.Spec.HardNodeAffinity to volume.status.ownerID instead.

Longhorn/longhorn#6239

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -1564,11 +1564,14 @@ func (c *VolumeController) reconcileVolumeCondition(v *longhorn.Volume, e *longh
 			continue
 		}
 		if v.Spec.DataLocality == longhorn.DataLocalityStrictLocal {
-			if v.Spec.NodeID == "" {
-				continue
+			if v.Status.RestoreRequired {
+				r.Spec.HardNodeAffinity = v.Status.OwnerID
+			} else {
+				if v.Spec.NodeID == "" {
+					continue
+				}
+				r.Spec.HardNodeAffinity = v.Spec.NodeID
 			}
-
-			r.Spec.HardNodeAffinity = v.Spec.NodeID
 		}
 		scheduledReplica, multiError, err := c.scheduler.ScheduleReplica(r, rs, v)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


ejweber (approved):
LGTM

derekbit:
@Mergify backport v1.5.x

derekbit:
@mergify backport v1.4.x

controller/volume_controller.go
```diff
@@ -1564,11 +1564,14 @@ func (c *VolumeController) reconcileVolumeCondition(v *longhorn.Volume, e *longh
 			continue
 		}
 		if v.Spec.DataLocality == longhorn.DataLocalityStrictLocal {
-			if v.Spec.NodeID == "" {
-				continue
+			if v.Status.RestoreRequired {
+				r.Spec.HardNodeAffinity = v.Status.OwnerID
```
innobead, line 1568: Can you explain where `v.Status.OwnerID` will be set when the volume is restoring? so it can be used here.

derekbit: For a volume being restored, the `volume.spec.nodeID` and `volume.status.ownerID` are empty, and the later one is determined in https://github.com/longhorn/longhorn-manager/blob/master/controller/volume_controller.go#L295.

innobead (changes requested):
I think this has been fixed in the master and 1.5.x code base already as below.
[/controller/volume_restore_controller.go#L190-L194](https://github.com/longhorn/longhorn-manager/blob/fedfefc273086b8f0425d07c8fe26b510683284a/controller/volume_restore_controller.go#L190-L194)
```go
	if vol.Status.RestoreRequired {
		createOrUpdateAttachmentTicket(va, restoringAttachmentTicketID, vol.Status.OwnerID, longhorn.TrueValue, longhorn.AttacherTypeVolumeRestoreController)
	} else {
		delete(va.Spec.AttachmentTickets, restoringAttachmentTicketID)
	}
```

The fix here is more for 1.4.x instead?

derekbit:
Ah, yes. You're right. I didn't notice the fix in v1.5.x and master. So, the fix targets v1.4.x.

innobead:
then, let's change the target branch is 1.4.x instead. :slightly_smiling_face:

derekbit:
Target for v1.4.x only

PR #2053, opened 2023-07-07
Implement/fix the unit tests of Volume Attachment and volume controller

PhanLe1010:
longhorn/longhorn#6005

### DIFF
--- a/controller/volume_attachment_controller_test.go
+++ b/controller/volume_attachment_controller_test.go
@@ -1,400 +1,500 @@
 package controller
 
-//
-//import (
-//	"context"
-//	"fmt"
-//	"github.com/longhorn/longhorn-manager/datastore"
-//	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
-//	lhfake "github.com/longhorn/longhorn-manager/k8s/pkg/client/clientset/versioned/fake"
-//	lhinformerfactory "github.com/longhorn/longhorn-manager/k8s/pkg/client/informers/externalversions"
-//	"github.com/sirupsen/logrus"
-//	. "gopkg.in/check.v1"
-//	apiextensionsfake "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset/fake"
-//	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-//	"k8s.io/client-go/informers"
-//	"k8s.io/client-go/kubernetes/fake"
-//	"k8s.io/client-go/kubernetes/scheme"
-//	"k8s.io/client-go/tools/record"
-//	utilpointer "k8s.io/utils/pointer"
-//)
-//
-//type volumeAttachmentTestCase struct {
-//	volAttachment *longhorn.VolumeAttachment
-//	vol           *longhorn.Volume
-//
-//	expectedVolAttachment *longhorn.VolumeAttachment
-//	expectedVol           *longhorn.Volume
-//}
-//
-//func (tc *volumeAttachmentTestCase) copyCurrentToExpect() {
-//	tc.expectedVolAttachment = tc.volAttachment.DeepCopy()
-//	tc.expectedVol = tc.vol.DeepCopy()
-//}
-//
-//func (s *TestSuite) TestVolumeAttachmentLifeCycle(c *C) {
-//	var tc *volumeAttachmentTestCase
-//	testCases := map[string]*volumeAttachmentTestCase{}
-//
-//	///////////////////////////////////////////////////////////////////
-//	tc = generateVolumeAttachmentTestCaseTemplate(TestVolumeName)
-//	tc.volAttachment.Spec.AttachmentSpecs = map[string]*longhorn.AttachmentSpec{
-//		"attachment-01": &longhorn.AttachmentSpec{
-//			ID:         "attachment-01",
-//			Type:       longhorn.AttacherTypeCSIAttacher,
-//			NodeID:     TestNode1,
-//			Parameters: map[string]string{},
-//			Generation: 0,
-//		},
-//	}
-//	tc.vol.Status.OwnerID = TestNode1
-//	tc.vol.Status.State = longhorn.VolumeStateDetached
-//	tc.copyCurrentToExpect()
-//	tc.expectedVolAttachment.Status.AttachmentStatuses = map[string]*longhorn.AttachmentStatus{
-//		"attachment-01": &longhorn.AttachmentStatus{
-//			ID:         "attachment-01",
-//			Satisfied:  false,
-//			Generation: 0,
-//		},
-//	}
-//	tc.expectedVol.Spec.NodeID = TestNode1
-//	testCases["test case 1: attach: basic"] = tc
-//	///////////////////////////////////////////////////////////////////
-//
-//	///////////////////////////////////////////////////////////////////
-//	tc = generateVolumeAttachmentTestCaseTemplate(TestVolumeName)
-//	tc.volAttachment.Spec.Attachments = map[string]*longhorn.Attachment{
-//		"attachment-01": &longhorn.Attachment{
-//			ID:         "attachment-01",
-//			Type:       longhorn.AttacherTypeCSIAttacher,
-//			NodeID:     TestNode1,
-//			Parameters: map[string]string{},
-//		},
-//		"attachment-02": &longhorn.Attachment{
-//			ID:         "attachment-02",
-//			Type:       longhorn.AttacherTypeSnapshotController,
-//			NodeID:     TestNode2,
-//			Parameters: map[string]string{},
-//		},
-//	}
-//	tc.vol.Status.OwnerID = TestNode1
-//	tc.vol.Status.State = longhorn.VolumeStateDetached
-//	tc.copyCurrentToExpect()
-//	tc.expectedVolAttachment.Status.Attachments = map[string]*longhorn.Attachment{
-//		"attachment-01": &longhorn.Attachment{
-//			ID:         "attachment-01",
-//			Type:       longhorn.AttacherTypeCSIAttacher,
-//			NodeID:     TestNode1,
-//			Parameters: map[string]string{},
-//			Attached:   utilpointer.Bool(false),
-//		},
-//		"attachment-02": &longhorn.Attachment{
-//			ID:         "attachment-02",
-//			Type:       longhorn.AttacherTypeSnapshotController,
-//			NodeID:     TestNode2,
-//			Parameters: map[string]string{},
-//			Attached:   utilpointer.Bool(false),
-//			AttachError: &longhorn.VolumeError{
-//				Message: fmt.Sprintf("cannot attach the volume to node %v because volume has already desired to be attached to node %v", TestNode2, TestNode1),
-//			},
-//		},
-//	}
-//	tc.expectedVol.Spec.NodeID = TestNode1
-//	testCases["test case 2: attach: multiple attachments"] = tc
-//	///////////////////////////////////////////////////////////////////
-//
-//	///////////////////////////////////////////////////////////////////
-//	tc = generateVolumeAttachmentTestCaseTemplate(TestVolumeName)
-//	tc.volAttachment.Spec.Attachments = map[string]*longhorn.Attachment{
-//		"attachment-01": &longhorn.Attachment{
-//			ID:         "attachment-01",
-//			Type:       longhorn.AttacherTypeCSIAttacher,
-//			NodeID:     TestNode1,
-//			Parameters: map[string]string{},
-//		},
-//		"attachment-02": &longhorn.Attachment{
-//			ID:         "attachment-02",
-//			Type:       longhorn.AttacherTypeCSIAttacher,
-//			NodeID:     TestNode2,
-//			Parameters: map[string]string{},
-//		},
-//	}
-//	tc.vol.Status.OwnerID = TestNode1
-//	tc.vol.Status.State = longhorn.VolumeStateDetached
-//	tc.copyCurrentToExpect()
-//	tc.expectedVolAttachment.Status.Attachments = map[string]*longhorn.Attachment{
-//		"attachment-01": &longhorn.Attachment{
-//			ID:         "attachment-01",
-//			Type:       longhorn.AttacherTypeCSIAttacher,
-//			NodeID:     TestNode1,
-//			Parameters: map[string]string{},
-//			Attached:   utilpointer.Bool(false),
-//		},
-//		"attachment-02": &longhorn.Attachment{
-//			ID:         "attachment-02",
-//			Type:       longhorn.AttacherTypeCSIAttacher,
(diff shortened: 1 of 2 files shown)


PhanLe1010:
@mergify backport v1.5.x

controller/volume_attachment_controller_test.go
```diff
@@ -1,400 +1,408 @@
...
+			Satisfied: false,
+			Conditions: types.SetConditionWithoutTimestamp([]longhorn.Condition{},
+				longhorn.AttachmentStatusConditionTypeSatisfied, longhorn.ConditionStatusFalse, "", ""),
+			Generation: 0,
+		},
+	}
+	tc.expectedVol.Spec.NodeID = TestNode2
+	testCases["test case 2: attach: multiple attachments"] = tc
+	///////////////////////////////////////////////////////////////////
+
```
james-munson, line 105: When you take out the comments, this is the same set of tests with Attachment replaced by AttachmentTicketStatus.  The field mappings look reasonable.
Tests 2 and 3 formerly had error message text to check, and that's not possible any more?  (Test 5 has a ReasonAttached string, so it must be possible sometimes.)

james-munson (approved):
Looks reasonable to me.

controller/volume_attachment_controller_test.go
PhanLe1010: Since the last time I wrote this unit test, the AD controller have been evolve quite a bit so the error message and struct field need to be modified.

> Tests 2 and 3 formerly had error message text to check, and that's not possible any more?

Yeah, you are correct

> Test 5 has a ReasonAttached string, so it must be possible sometimes.

This is also correct

controller/volume_controller_test.go
```diff
@@ -913,7 +925,7 @@ func (s *TestSuite) TestVolumeLifeCycle(c *C) {
 	}
 	tc.expectReplicas = expectRs
 	testCases["volume attaching - start replicas - manager down"] = tc
-	//s.runTestCases(c, testCases)
+	s.runTestCases(c, testCases)
```
ejweber, line 928: It looks like we are running most volume controller test cases twice due to this line duplicating 1058. Confirmed by running `go test -v ./controller/...` on this branch and searching for "volume attaching - start replicas - manager down", which runs twice and "replica rebuilding - delay replica replenishment", which runs once. I immediately see a reason for the duplication.

ejweber (changes requested):
One potential blocker, but a quick fix if one is needed.

I have some questions about changes in volume_controller_test.go, but they are more for my understanding than anything. I will sync up with you offline.

controller/volume_attachment_controller_test.go
```diff
@@ -1,400 +1,408 @@
...
+				longhorn.AttachmentStatusConditionTypeSatisfied, longhorn.ConditionStatusTrue, "", ""),
+			Generation: 0,
+		},
+	}
+	tc.vol.Status.OwnerID = TestNode1
+	tc.vol.Spec.NodeID = TestNode1
+	tc.vol.Status.CurrentNodeID = TestNode1
+	tc.vol.Status.State = longhorn.VolumeStateAttached
+	tc.copyCurrentToExpect()
+	tc.expectedVolAttachment.Status.AttachmentTicketStatuses = map[string]*longhorn.AttachmentTicketStatus{
```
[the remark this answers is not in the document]
PhanLe1010: Thank you. Fixed

controller/volume_controller_test.go
PhanLe1010: Nice catch! Thank you. Fixed

controller/volume_attachment_controller_test.go
```diff
@@ -1,400 +1,400 @@
...
+		},
+		"attachment-02": &longhorn.AttachmentTicketStatus{
+			ID:        "attachment-02",
+			Satisfied: false,
+			Conditions: types.SetConditionWithoutTimestamp([]longhorn.Condition{},
+				longhorn.AttachmentStatusConditionTypeSatisfied, longhorn.ConditionStatusFalse, "", ""),
+			Generation: 0,
+		},
+	}
+	tc.expectedVol.Spec.NodeID = TestNode1
```
shuo-wu, line 143: NBIT: Explain why the expected NodeID is `TestNode1` when the priority is identical. (The candidates are sort by name)

controller/volume_attachment_controller_test.go
```diff
@@ -1,400 +1,400 @@
...
+		},
+		"attachment-02": &longhorn.AttachmentTicketStatus{
+			ID:        "attachment-02",
+			Satisfied: true,
+			Conditions: types.SetConditionWithoutTimestamp([]longhorn.Condition{},
+				longhorn.AttachmentStatusConditionTypeSatisfied, longhorn.ConditionStatusTrue, "", ""),
+			Generation: 0,
+		},
+	}
+	testCases["test case 5: attach: fail to attach because the volume is already attached with incompatible parameters"] = tc
```
shuo-wu, line 218: Question: Will the VA controller re-attach the volume for CSIAttacher ticket here?

controller/volume_attachment_controller_test.go
```diff
@@ -1,400 +1,400 @@
...
+	tc.vol.Spec.DisableFrontend = true
+	tc.vol.Status.CurrentNodeID = TestNode1
+	tc.vol.Status.State = longhorn.VolumeStateAttached
+	tc.copyCurrentToExpect()
+	tc.expectedVolAttachment.Status.AttachmentTicketStatuses = map[string]*longhorn.AttachmentTicketStatus{
+		"attachment-01": &longhorn.AttachmentTicketStatus{
+			ID:        "attachment-01",
+			Satisfied: true,
+			Conditions: types.SetConditionWithoutTimestamp([]longhorn.Condition{},
+				longhorn.AttachmentStatusConditionTypeSatisfied, longhorn.ConditionStatusTrue, "", ""),
```
shuo-wu, line 301: Why is this true? Isn't the volume still state detaching?

shuo-wu (commented):
I think it's better to add 2 more kinds of tests for the VA controller:
1. `Generation` related tests
2. Attachment/detachment cases caused by higher priority tickets suddenly introducing.

controller/volume_attachment_controller_test.go
```diff
@@ -1,400 +1,400 @@
...
-//	}
-//}
-//
-//func generateVolumeAttachmentTestCaseTemplate(name string) *volumeAttachmentTestCase {
-//	return &volumeAttachmentTestCase{
-//		volAttachment: newVolumeAttachment(name),
-//		vol:           newVolume(name, 1),
-//	}
-//}
+import (
```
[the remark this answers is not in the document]
PhanLe1010: Thank you. Fixed

controller/volume_attachment_controller_test.go
PhanLe1010: Thank you. Fixed

controller/volume_attachment_controller_test.go
PhanLe1010: When the volume CR has some changes, VA controller will retry

controller/volume_attachment_controller_test.go
PhanLe1010: The volume start detaching because the volume.Spec.NodeID is just reset. However, the vol.Status is still attached at this moment so the AD ticket refect that fact.

PhanLe1010:
> I think it's better to add 2 more kinds of tests for the VA controller:
> 
> 1. `Generation` related tests
> 2. Attachment/detachment cases caused by higher priority tickets suddenly introducing.

Thank you. Added

controller/volume_attachment_controller_test.go
shuo-wu: Sorry... Are there any changes in the volume CR...?

PhanLe1010: Sorry for the confusion.

So the CSI ticket cannot be satisfied yet because the volume is already attached with a different parameter. When the volume is detached, the VA will watch that event and retry to fulfill the CSI ticket. Does it answer the question?

shuo-wu: So in this case the VA controller will do nothing for the CSI Attacher ticket even if it holds a higher priority?

controller/volume_attachment_controller_test.go
```diff
@@ -1,400 +1,500 @@
...
+	retVol, err := lhClient.LonghornV1beta2().Volumes(TestNamespace).Get(context.TODO(), tc.vol.Name, metav1.GetOptions{})
+	c.Assert(err, IsNil)
+	c.Assert(retVol.Spec, DeepEquals, tc.expectedVol.Spec)
+
+	retVolAttachment, err := lhClient.LonghornV1beta2().VolumeAttachments(TestNamespace).Get(context.TODO(), tc.volAttachment.Name, metav1.GetOptions{})
+	c.Assert(err, IsNil)
+	// mask timestamps
+	for _, ticketStatus := range retVolAttachment.Status.AttachmentTicketStatuses {
+		for ctype, condition := range ticketStatus.Conditions {
+			condition.LastTransitionTime = ""
```
shuo-wu, line 468: NIT: You can check `nowHandler` in other controllers and see how they handle this timestamp issue

controller/volume_controller_test.go
```diff
@@ -128,20 +128,22 @@ func (s *TestSuite) TestVolumeLifeCycle(c *C) {
...
 	tc.expectVolume.Status.State = longhorn.VolumeStateCreating
 	tc.expectVolume.Status.CurrentImage = tc.volume.Spec.EngineImage
-	tc.expectVolume.Status.Robustness = longhorn.VolumeRobustnessUnknown
 	tc.expectVolume.Status.Conditions = setVolumeConditionWithoutTimestamp(tc.expectVolume.Status.Conditions,
 		longhorn.VolumeConditionTypeScheduled, longhorn.ConditionStatusFalse, longhorn.VolumeConditionReasonReplicaSchedulingFailure, longhorn.ErrorReplicaScheduleNodeUnavailable)
 	testCases["volume create - replica scheduling failure"] = tc
 
 	// after creation, volume in detached state
 	tc = generateVolumeTestCaseTemplate()
+	tc.volume.Status.State = longhorn.VolumeStateDetaching
```
shuo-wu, line 137: NIT: Do we need to add a case that explain how the volume state is updated from `creating` to `detaching`?

controller/volume_controller_test.go
```diff
@@ -102,7 +103,6 @@ func (s *TestSuite) TestVolumeLifeCycle(c *C) {
 	// default replica and engine objects will be copied by copyCurrentToExpect
 	tc.copyCurrentToExpect()
 	tc.expectVolume.Status.State = longhorn.VolumeStateCreating
-	tc.expectVolume.Status.Robustness = longhorn.VolumeRobustnessUnknown
```
shuo-wu, line 105: Why removing this?

controller/volume_controller_test.go
```diff
@@ -849,7 +860,7 @@ func (s *TestSuite) TestVolumeLifeCycle(c *C) {
 
 	// Salvage Requested
 	tc = generateVolumeTestCaseTemplate()
-	tc.volume.Spec.NodeID = TestNode1
+	tc.volume.Spec.NodeID = ""
```
shuo-wu, line 863: One question, which controller/part is responsible for unsettling this after `v.Status.Robustness = longhorn.VolumeRobustnessFaulted`?

ejweber (commented):
The changes I asked for were made. Curious on a couple of Shuo's questions.

controller/volume_attachment_controller_test.go
PhanLe1010: Oh. it will. VA controller will set vol.Spec.NodeId to "" to detach the volume

PhanLe1010: Yes, you are right. This seems to be an unexpected behavior when the low priority ticket already attached and the higher priority ticket is created with the same node and different parameter

controller/volume_attachment_controller_test.go
PhanLe1010: Thanks for the hint. I looked at this `nowHandler` before. I checked it again now. However, it doesn't seems to solve the timestamp of the conditions field. 

If we look into the [setCondition() function](https://github.com/longhorn/longhorn-manager/blob/a5b67e5ec57ec38be0d50cdbe77428d924f1b2fb/types/condition.go#L62), we can see that it directly `util.Now()` to set LastTransitionTime, it doesn't allow the caller to influence the behavior. This is why inside volume_controller_test.go, we also have to mask the LastTransitionTime of conditions [here](https://github.com/longhorn/longhorn-manager/blob/a5b67e5ec57ec38be0d50cdbe77428d924f1b2fb/controller/volume_controller_test.go#L1490) even though volume_controller has the `nowHandler`

I know perhaps, we can modify it so that setCondition accepts a `nowHandler` as an argument and use it to set time, but it would be outside scope of this unit test PR. In addition, I don't see much benefit from having a `nowHandler` function to set the condition LastTransitionTime to a fix a value comparing to just clearing them directly

controller/volume_controller_test.go
PhanLe1010: Behavior of volume_controller changed, now when the volume is just created we create replicas and engine, set `vol.Status.State = longhorn.VolumeStateCreating` and returns: [link](https://github.com/longhorn/longhorn-manager/blob/a5b67e5ec57ec38be0d50cdbe77428d924f1b2fb/controller/volume_controller.go#L1381C2-L1381C2). We set volume reobusnees until the next sync.

controller/volume_controller_test.go
PhanLe1010: The VA controller https://github.com/longhorn/longhorn-manager/blob/a5b67e5ec57ec38be0d50cdbe77428d924f1b2fb/controller/volume_attachment_controller.go#L461

controller/volume_controller_test.go
PhanLe1010: Good idea. Added. Thank you!

controller/volume_controller_test.go
shuo-wu: I see. When the volume is updated from `creating` to `detaching` (in the next test case), this volume robustness field is set to `longhorn.VolumeRobustnessUnknown`

controller/volume_controller_test.go
shuo-wu: NIT: Then you can add a comment that explains this is at the mid of the salvage flow. (volume is Faulted and the VA controller already unset spec.NodeID for this volume.) Or if possible, you can add one more case that explains how the volume becomes Faulted.

shuo-wu (approved):
In general LGTM

ejweber (dismissed):

controller/volume_controller_test.go
PhanLe1010: I added the comment. Thank you!

james-munson (approved):

ejweber (approved):
LGTM

PR #2054, opened 2023-07-07
Fix migration stuck due to snapshot chains in old engine and new engine mismatch

PhanLe1010:
The symptom is similar to https://github.com/longhorn/longhorn/issues/6215#issuecomment-1612222566. But this one is at API level instead of inside the AD controller

longhorn/longhorn#6238

### DIFF
--- a/manager/volume.go
+++ b/manager/volume.go
@@ -2,7 +2,6 @@ package manager
 
 import (
 	"fmt"
-	"reflect"
 	"strconv"
 	"time"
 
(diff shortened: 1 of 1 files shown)


PhanLe1010:
Test result https://ci.longhorn.io/job/private/job/longhorn-tests-regression/4503/

james-munson (approved):
Looks good.

manager/volume.go
```diff
@@ -409,14 +408,28 @@ func (m *VolumeManager) checkMigratingEngineSyncSnapshots(vol *longhorn.Volume)
...
-	if !reflect.DeepEqual(oldEngine.Status.Snapshots, migratingEngine.Status.Snapshots) {
+	if !hasSameKeys(oldEngine.Status.Snapshots, migratingEngine.Status.Snapshots) {
 		logrus.Debugf("Volume migration (%v) is in progress for synchronizing snapshots", vol.Name)
 		return false, nil
 	}
 
 	return true, nil
 }
 
+func hasSameKeys(map1, map2 map[string]*longhorn.SnapshotInfo) bool {
```
ejweber, line 419: NIT: I would prefer to see a comment here explaining why this function is needed (e.g. map[string]*longhorn.SnapshotInfo can't be compared directly because created field might be different).

ejweber (approved):
LGTM with one nit.

manager/volume.go
PhanLe1010: Added the comment in the above function. Thank you!

ejweber (approved):

PhanLe1010:
@mergify backport v1.3.x

derekbit:
LGTM as well

innobead:
cc @mantissahz

PR #2055, opened 2023-07-07
Remove orphapn UPDATE validation

derekbit:
Longhorn/longhorn#6246

### DIFF
--- a/webhook/resources/orphan/validator.go
+++ b/webhook/resources/orphan/validator.go
@@ -2,7 +2,6 @@ package orphan
 
 import (
 	"fmt"
-	"reflect"
 
 	admissionregv1 "k8s.io/api/admissionregistration/v1"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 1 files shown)


innobead (changes requested):
You probably can still keep the validation, but only remove the below part. WDYT?

```go
	if !reflect.DeepEqual(oldOrphan.Spec, newOrphan.Spec) {
		return werror.NewInvalidError(fmt.Sprintf("orphan %v spec fields are immutable", oldOrphan.Name), "")
	}
```

derekbit:
> You probably can still keep the validation, but only remove the below part. WDYT?
> 
> ```go
> 	if !reflect.DeepEqual(oldOrphan.Spec, newOrphan.Spec) {
> 		return werror.NewInvalidError(fmt.Sprintf("orphan %v spec fields are immutable", oldOrphan.Name), "")
> 	}
> ```

Yeah, good idea. Updated. Thank you.

innobead:
@mergify backport v1.5.x

PR #2058, opened 2023-07-10
fix(recurring-job): flooded source label log

c3y1huang:
https://github.com/longhorn/longhorn/issues/6257

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -3647,7 +3647,8 @@ func (c *VolumeController) syncPVCRecurringJobLabels(volume *longhorn.Volume) er
 	}
 
 	if !hasSourceLabel {
-		c.logger.Warnf("Ignoring recurring job labels on Volume %v PVC %v due to missing source label", volume.Name, pvc.Name)
+		c.logger.Debugf("Ignoring recurring job labels on Volume %v PVC %v due to missing source label", volume.Name, pvc.Name)
+
 		return nil
 	}
 
--- a/types/setting.go
+++ b/types/setting.go
@@ -1064,7 +1064,7 @@ var (
 		Type:        SettingTypeString,
 		Required:    true,
 		ReadOnly:    false,
-		Default:     "Debug",
+		Default:     "Info",
 	}
 
 	SettingDefinitionOfflineReplicaRebuilding = SettingDefinition{


controller/volume_controller.go
```diff
@@ -3647,7 +3647,7 @@ func (c *VolumeController) syncPVCRecurringJobLabels(volume *longhorn.Volume) er
 	}
 
 	if !hasSourceLabel {
-		c.logger.Warnf("Ignoring recurring job labels on Volume %v PVC %v due to missing source label", volume.Name, pvc.Name)
+		logrus.Tracef("Ignoring recurring job labels on Volume %v PVC %v due to missing source label", volume.Name, pvc.Name)
```
innobead, line 3650: Why not using the existing c.logger?

innobead (changes requested):

controller/volume_controller.go
c3y1huang: [There is no trace in c.logger.](https://github.com/longhorn/longhorn-manager/blob/7a5ca17682b43e6da430d8e459566c713eb4a29f/vendor/github.com/sirupsen/logrus/logrus.go#L139-L177)

derekbit: Most log messages' levels are corrected. I think we use INFO as default log level. Then, use `c.logger.Debugf` here. WDYT? @c3y1huang @innobead

innobead: IIRC we still use debug, confirm?

innobead: @c3y1huang do u know why it doesn't have trace? Duo to a different interface?

c3y1huang: > @c3y1huang do u know why it doesn't have trace? Duo to a different interface?

Yes, [it got removed due to backward compatibility concerns](https://github.com/sirupsen/logrus/commit/b54cafe5cee93a35bf6b061e261e8265ea0a628c).

derekbit: > IIRC we still use debug, confirm? @derekbit

How about changing it to INFO from v1.5.1?
I've adjusted most of log messages' levels.

c3y1huang: Updated.

derekbit (approved):
LGTM.

BTW, I adjust some logs in https://github.com/longhorn/longhorn-manager/pull/2059.

innobead (approved):

innobead:
@Mergifyio backport v1.5.x

innobead:
@c3y1huang need to backport this to 1.4 and 1.3 with a different implementation since their default log level is debug.

c3y1huang:

c3y1huang:
> @c3y1huang need to backport this to 1.4 and 1.3 with a different implementation since their default log level is debug.

I think we don't need to backport this since the source label is introduced in v1.5.0.

innobead:
@c3y1huang please go ahead to remove backup labels. Thanks.

PR #2063, opened 2023-07-10
fix(backup): pvc storage class could be in annotation

ChanYiLin:
ref: [longhorn/longhorn#6264](https://github.com/longhorn/longhorn/issues/6264)

### DIFF
--- a/controller/backup_controller.go
+++ b/controller/backup_controller.go
@@ -636,7 +636,17 @@ func (bc *BackupController) checkMonitor(backup *longhorn.Backup, volume *longho
 	if kubernetesStatus.PVCName != "" && kubernetesStatus.LastPVCRefAt == "" {
 		pvc, _ := bc.ds.GetPersistentVolumeClaim(kubernetesStatus.Namespace, kubernetesStatus.PVCName)
 		if pvc != nil {
-			storageClassName = *pvc.Spec.StorageClassName
+			if pvc.Spec.StorageClassName != nil {
+				storageClassName = *pvc.Spec.StorageClassName
+			}
+			if storageClassName == "" {
+				if v, exist := pvc.Annotations[corev1.BetaStorageClassAnnotation]; exist {
+					storageClassName = v
+				}
+			}
+			if storageClassName == "" {
+				bc.logger.Warnf("Failed to find the StorageClassName from the pvc %v", pvc.Name)
+			}
 		}
 	}
 


controller/backup_controller.go
```diff
@@ -636,7 +636,19 @@ func (bc *BackupController) checkMonitor(backup *longhorn.Backup, volume *longho
 	if kubernetesStatus.PVCName != "" && kubernetesStatus.LastPVCRefAt == "" {
 		pvc, _ := bc.ds.GetPersistentVolumeClaim(kubernetesStatus.Namespace, kubernetesStatus.PVCName)
 		if pvc != nil {
-			storageClassName = *pvc.Spec.StorageClassName
+			if pvc.Spec.StorageClassName != nil {
+				storageClassName = *pvc.Spec.StorageClassName
+			}
+			if storageClassName == "" {
+				if v, exist := pvc.Annotations["volume.beta.kubernetes.io/storage-class"]; exist {
```
derekbit, line 643: use `corev1.BetaStorageClassAnnotation` instead.

ChanYiLin: thanks for reminding the const name!

ChanYiLin: Do you think we need to support alpha
I didn't find alpha any k8s doc
but from users comment
```
---
kind: "PersistentVolumeClaim"
apiVersion: "v1"
metadata:
  name: {{ template "portainer.fullname" . }}
  namespace: {{ .Release.Namespace }}
  annotations:
  {{- if .Values.persistence.storageClass }}
    volume.beta.kubernetes.io/storage-class: {{ .Values.persistence.storageClass | quote }}
  {{- else }}
    volume.alpha.kubernetes.io/storage-class: "generic"
  {{- end }}
  ....
  ....
```
They have alpha ?

derekbit (dismissed):
LGTM

controller/backup_controller.go
derekbit: I'm fine with getting from `alpha`, because the PVC manifest might follow old rules or from created from old cluster.

ChanYiLin: Got it, but for alpha, there is no const defined for it now.

ChanYiLin: Fixed.

derekbit: Yeah, just keep it as it is.

shuo-wu: There is no `volume.alpha.kubernetes.io/storage-class` or something similar defined in `k8s.io/api/core/v1/types.go`. Do we need to consider it? Or how can we reproduce this...?

ChanYiLin: I have tested the annotation
`volume.alpha.kubernetes.io/storage-class` won't work
the pvc will use the default storageClass in the cluster instead, and k8s will add the `storageClassName` to the pvc.Spec automatically in this case

I think we don't have to handle alpha annotation
cc @derekbit @shuo-wu

shuo-wu: I am fine with ignoring the alpha annotation as it's probably not used by Kubernetes.

derekbit: Sure. logging it here is fine as well, e.g. `failed to find storageclass from the pvc....`

ChanYiLin: Fixed.

controller/backup_controller.go
```diff
@@ -636,7 +636,15 @@ func (bc *BackupController) checkMonitor(backup *longhorn.Backup, volume *longho
...
-			storageClassName = *pvc.Spec.StorageClassName
+			if pvc.Spec.StorageClassName != nil {
+				storageClassName = *pvc.Spec.StorageClassName
+			}
+			if storageClassName == "" {
+				if v, exist := pvc.Annotations[corev1.BetaStorageClassAnnotation]; exist {
+					storageClassName = v
+				}
+			}
+			bc.logger.Warnf("Failed to find the StorageClassName from the pvc %v", pvc.Name)
```
derekbit, line 647:
```suggestion
                if storageClassName == "" {
			bc.logger.Warnf("Failed to find the StorageClassName from the pvc %v", pvc.Name)
	        }
```

ChanYiLin: thanks!

ChanYiLin: Fixed!

derekbit (approved):

innobead (approved):
What we had is a nil exception issue instead of if need to support the upstream deprecated annotation.

However, I am fine with adding this support, because it seems the upstream has not really tackled it yet. 

We made a wrong assumption to assume `Spec.StorageClassName` should not nil, but we should pay more attention to this when dealing with any reference variable, and should always do nil check w/o any assumption. 

ref: https://github.com/kubernetes/kubernetes/issues/116336

innobead:
@mergify backport v1.5.x

innobead:
No need to backport the change was only applied to 1.5.x.

derekbit:
> What we had is a nil exception issue instead of if need to support the upstream deprecated annotation.
> 
> However, I am fine with adding this support, because it seems the upstream has not really tackled it yet.
> 
> We made a wrong assumption to assume `Spec.StorageClassName` should not nil, but we should pay more attention to this when dealing with any reference variable, and should always do nil check w/o any assumption.
> 
> ref: [kubernetes/kubernetes#116336](https://github.com/kubernetes/kubernetes/issues/116336)

Another question is that why we use `storageClassName = *pvc.Spec.StorageClassName` rather than `storageClassName = pvc.Spec.StorageClassName`? The latter one looks enough?

innobead:
I thought `pvc.Spec.StorageClassName` is a reference?

derekbit:
> pvc.Spec.StorageClassName

Yeah, you're right. Didn't notice it https://github.com/kubernetes/kubernetes/blob/0ae9aaacfa217b18652f34d6ef29f08a15851b0a/pkg/apis/core/types.go#L445.

ChanYiLin:
> No need to backport the change was only applied to 1.5.x. @ChanYiLin right?

Yes, this feature only in v1.5.0 cc @innobead

PR #2065, opened 2023-07-10
Fix strict-local volume restore stuck in attaching state

derekbit:
For a volume being restored, the spec.nodeID is empyty but the status.ownerID is set. Thus, set r.Spec.HardNodeAffinity to volume.status.ownerID instead.

Longhorn/longhorn#6239

### DIFF
--- a/types/types.go
+++ b/types/types.go
@@ -239,7 +239,7 @@ const (
 )
 
 // SettingsRelatedToVolume should match the items in datastore.GetLabelsForVolumesFollowsGlobalSettings
-//   TODO: May need to add the data locality check
+// TODO: May need to add the data locality check
 var SettingsRelatedToVolume = map[string]string{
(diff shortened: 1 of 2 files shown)


types/types.go
```diff
@@ -807,7 +807,8 @@ type DiskSpecWithName struct {
 
 // UnmarshalToDisks input format should be:
 // `[{"path":"/mnt/disk1","allowScheduling":false},
-//   {"path":"/mnt/disk2","allowScheduling":false,"storageReserved":1024,"tags":["ssd","fast"]}]`
+//
+//	{"path":"/mnt/disk2","allowScheduling":false,"storageReserved":1024,"tags":["ssd","fast"]}]`
```
innobead, line 811: Why need to change this?

innobead (changes requested):

types/types.go
derekbit: My edit changes this by itself....didn't notice it.

PR #2066, opened 2023-07-10
fix(support-bundle): download timed out

c3y1huang:
https://github.com/longhorn/longhorn/issues/6256


derekbit (approved):
LGTM

c3y1huang:
[Test results](https://github.com/longhorn/longhorn/issues/6256)

innobead (approved):
LGTM.

note: probably we don't need any client timeout but let the server side decide instead.

PR #2067, opened 2023-07-10
fix(server): return 404 if get backing image and backup not found

ChanYiLin:
ref: [longhorn/longhorn#6266](https://github.com/longhorn/longhorn/issues/6266)

### DIFF
--- a/api/router.go
+++ b/api/router.go
@@ -8,6 +8,7 @@ import (
 	"github.com/rancher/go-rancher/client"
 	"github.com/sirupsen/logrus"
 
+	"github.com/longhorn/longhorn-manager/datastore"
 	"github.com/longhorn/longhorn-manager/metrics_collector/registry"
 )
 
@@ -17,6 +18,12 @@ func HandleError(s *client.Schemas, t HandleFuncWithError) http.Handler {
 	return api.ApiHandler(s, http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
 		if err := t(rw, req); err != nil {
 			logrus.WithError(err).Warnf("HTTP handling error")
+
+			if datastore.ErrorIsNotFound(err) {
+				rw.WriteHeader(http.StatusNotFound)
+				return
+			}
+
 			apiContext := api.GetApiContext(req)
 			apiContext.WriteErr(err)
 		}


api/backingimage.go
```diff
@@ -5,6 +5,7 @@ import (
 	"net/http"
 
 	"github.com/gorilla/mux"
+	"github.com/longhorn/longhorn-manager/datastore"
```
shuo-wu, line 8: NIT: re-arrange this.

api/backup.go
```diff
@@ -4,6 +4,7 @@ import (
 	"net/http"
 
 	"github.com/gorilla/mux"
+	"github.com/longhorn/longhorn-manager/datastore"
```
shuo-wu, line 7: NIT: re-arrange this.

shuo-wu (approved):
Do we need to handle volume/backupvolume for CSI API `DeleteSnapshot`?
The idea LGTM.

api/backup.go
ChanYiLin: Fixed, thanks!

api/backingimage.go
ChanYiLin: Fixed.

ChanYiLin:
VolumeSnapshot has 3 types
- backing image
- snapshot
- backup

When deleting
- csi: try to get the CR first, it it doesn't exist, then pass
- client: check if 404, return nil
- server: return 404, if it cannot get the CR from datastore

innobead (dismissed):
LGTM.

innobead:
@shuo-wu Please help review this.

ChanYiLin:
@innobead 
I am running the e2e test, and it failed at backup snapshot
Let me check for it

innobead:
NP. Passing the test is the criteria before merging.

ChanYiLin:
weird, it passed the test in my environment
need some time to investigate

ChanYiLin:
I know the reason why
Because I use `/v1/backups/XXX` in our generated client to try to get the backup before deleting it 
Turns out, we don't provide such API in our server
So csi thought there was no such backup and pass the deletion, and failed when checking if the backup still existed in the python test.

ChanYiLin:
We don't have to check by get the backup or snapshot before deletion
because in the delete api we already pass the deletion if it is not found

https://github.com/longhorn/longhorn-manager/blob/75d2e77c539bd8a635140b87e81779b7d049787b/api/backup.go#L125-L134

so removed the check

ChanYiLin:
e2e running
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/4550/console

shuo-wu:
> We don't have to check by getting the backup or snapshot before deletion
because in the delete api we already pass the deletion if it is not found
[longhorn-manager/api/backup.go](https://github.com/longhorn/longhorn-manager/blob/75d2e77c539bd8a635140b87e81779b7d049787b/api/backup.go#L125-L134)
so removed the check

Sorry... Does the PR contain the removal (of getting backing image/backup volume before requesting deletion)...? BTW, for the snapshot deletion, getting volume is unavoidable. We still need to make sure getting a nil volume should be handled as expected.

ChanYiLin:
Hi @shuo-wu 
No, this PR maintain the same as before
And only add 404 not found header in server side
For csi part, I remove all the change I made since the deletion api for `backup` and `snapshot` already check if it exists

ChanYiLin:
e2e tests passed
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/4550/

shuo-wu:
OK. I was thinking you directly remove the backup volume/backing image get calls from the CSI `SnapshotDelete` implementation. But this is an alternative solution for this issue while making the implementation simpler. WDTY?

ChanYiLin:
I think if from server side it already check if it exists before delete it
We don't need to do it in client side again 
and if we do it in client side, we have to make sure every client call has checked if it exists? 
I would say it adds the burden to the client side.

innobead:
So do we need to do any other code change here?

ChanYiLin:
> So do we need to do any other code change here?

No, I have tested it manually and through e2e
Three types of volumesnapshot can be successfully deleted in any case

shuo-wu:
I mean, for backing image cleanup, since there is no parent resource, there is no need to invoke `cs.apiClient.BackingImage.ById(backingImageParameters[longhorn.BackingImageParameterName])` before deletion. We can directly call `cs.apiClient.BackingImage.Delete(backingImage)`

ChanYiLin:
@shuo-wu
But in csi we don't have longhorn CR clientset only have longhorn api client
so we manipulate the resources through api client
and for BackingImage delete api, it needs BackingImage object

Even we delete the BackingImage directly, if it doesn't exist, we still need to handle not found error
I don't think the process will be different
- get -> if found -> delete
- delete -> if not found error then pass

And besides this issue, we still need to add 404 not found in the server api for other cases, maybe there are some places in our code also use 404 to determine if api has not found error.

shuo-wu (approved):

innobead:
@mergify backport v1.5.x

PR #2068, opened 2023-07-10
Mutate the resources that need update during upgrade

derekbit:
The patch is only for improving resilience.

Longhorn/longhorn#6259

### DIFF
--- a/upgrade/v14xto150/upgrade.go
+++ b/upgrade/v14xto150/upgrade.go
@@ -5,28 +5,29 @@ import (
 	"strconv"
 
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 
 	storagev1 "k8s.io/api/storage/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 	clientset "k8s.io/client-go/kubernetes"
 
 	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
 	lhclientset "github.com/longhorn/longhorn-manager/k8s/pkg/client/clientset/versioned"
 	"github.com/longhorn/longhorn-manager/types"
 	upgradeutil "github.com/longhorn/longhorn-manager/upgrade/util"
+	"github.com/longhorn/longhorn-manager/util"
 )
 
 const (
 	upgradeLogPrefix = "upgrade from v1.4.x to v1.5.0: "
+
+	maxRetryForDeploymentDeletion = 300
 )
 
 func UpgradeResources(namespace string, lhClient *lhclientset.Clientset, kubeClient *clientset.Clientset, resourceMaps map[string]interface{}) error {
 	if err := upgradeCSIPlugin(namespace, kubeClient); err != nil {
-		return err
-	}
-
-	if err := upgradeVolumes(namespace, lhClient, resourceMaps); err != nil {
 		return err
 	}
 
@@ -35,6 +36,10 @@ func UpgradeResources(namespace string, lhClient *lhclientset.Clientset, kubeCli
 	}
 
 	if err := upgradeWebhookPDB(namespace, kubeClient); err != nil {
+		return err
+	}
+
+	if err := upgradeVolumes(namespace, lhClient, resourceMaps); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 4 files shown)


upgrade/v14xto150/upgrade.go
```diff
@@ -46,6 +42,10 @@ func UpgradeResources(namespace string, lhClient *lhclientset.Clientset, kubeCli
 		return err
 	}
 
+	if err := upgradeVolumes(namespace, lhClient, resourceMaps); err != nil {
```
PhanLe1010, line 45: Maybe `upgradeVolumeAttachments` should be after `upgradeVolumes` ?

upgrade/v14xto150/upgrade.go
```diff
@@ -26,10 +26,6 @@ func UpgradeResources(namespace string, lhClient *lhclientset.Clientset, kubeCli
 		return err
 	}
 
-	if err := upgradeVolumes(namespace, lhClient, resourceMaps); err != nil {
-		return err
-	}
-
 	if err := upgradeWebhookAndRecoveryService(namespace, kubeClient); err != nil {
```
PhanLe1010, line 29: Good fix! Should we wait for the old webhook pods to be completely removed first? If not request might be sending to old webhooks and mess up the states

PhanLe1010 (changes requested):

upgrade/v14xto150/upgrade.go
derekbit: Good idea. Should we wait for a period, e.g. up to 300 seconds?

PhanLe1010: 300 sounds good. Then if timeout while waiting, we crash the longhorn-manager so that it retries webhook pod deletion. As long as we not move forward with CR upgrade while the old webhook pods exist, we are good.

innobead:
@ChanYiLin Please also help review this since some code was introduced by you, so it's better to review by you as well.

upgrade/v14xto150/upgrade.go
innobead: Sounds good, also it's better to have logs.

webhook/resources/volume/mutator.go
```diff
@@ -303,6 +303,12 @@ func (v *volumeMutator) Update(request *admission.Request, oldObj runtime.Object
 	if string(volume.Spec.BackendStoreDriver) == "" {
 		patchOps = append(patchOps, fmt.Sprintf(`{"op": "replace", "path": "/spec/backendStoreDriver", "value": "%s"}`, longhorn.BackendStoreDriverTypeV1))
 	}
+	if string(volume.Spec.BackupCompressionMethod) == "" {
+		patchOps = append(patchOps, fmt.Sprintf(`{"op": "replace", "path": "/spec/backupCompressionMethod", "value": "%s"}`, longhorn.BackupCompressionMethodGzip))
```
innobead, line 307: If not set, isn't it `BackupCompressionMethodNone` instead?

innobead (changes requested):
Good prompt improvement. These changes make sense, but just one feedback from my side and need to resolve the comments from Phan.

upgrade/v14xto150/upgrade.go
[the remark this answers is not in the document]
derekbit: Adjuested. Thank you.

upgrade/v14xto150/upgrade.go
derekbit: Addressed. Thank you.

webhook/resources/volume/mutator.go
derekbit: No, this is for the case that an existing volume with empty BackupCompressionMethod. The compression method is gzip.

innobead: OH I see, thanks for the clarification.

derekbit: Actually, these mutations are redundant because we already have upgrade path. However, it seems the upgrade path might not work properly in the rare cases, so the mutations are more for resilience.

innobead: I feel it seems like an atomic problem which means something is not finished in the middle but it has been regarded as upgrade finished somehow. Do you think we need to create an issue to follow up to make the upgrade more solid?

derekbit: I read the upgrade path codes but cannot have any idea.
Yes, I can create a ticket for it and see how to make the path more solid.

innobead: Sounds good. We can let another member help with that later.

upgrade/v14xto150/upgrade.go
```diff
@@ -230,16 +235,37 @@ func upgradeWebhookAndRecoveryService(namespace string, kubeClient *clientset.Cl
...
+			if err != nil {
+				return errors.Wrapf(err, upgradeLogPrefix+"failed to wait for the deployment with label %v to be deleted during the upgrade", selector)
+			}
+			logrus.Infof("Deleted deployment %v with label %v during the upgrade", deployment.Name, selector)
 		}
 	}
 	return nil
 }
 
+func waitForDeploymentDeleted(namespace, name string, kubeClient *clientset.Clientset) error {
```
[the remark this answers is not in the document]
innobead: If it's not easy to reuse, it's fine to keep the code here.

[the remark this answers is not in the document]
derekbit: Handling.
Working on using `waitForDeletion`. As you said, need to move it to util and this makes the layout _**ugly**_.
TODO: `delopyment_util.go` has many general functions that can move to an individual lib.

util/util.go
```diff
@@ -1006,3 +1006,16 @@ func VerifySnapshotLabels(labels map[string]string) error {
 func RemoveNewlines(input string) string {
 	return strings.Replace(input, "\n", "", -1)
 }
+
+type ResourceGetFunc func(kubeClient *clientset.Clientset, name, namespace string) (runtime.Object, error)
+
+func WaitForDeletion(kubeClient *clientset.Clientset, name, namespace, resource string, maxRetryForDeletion int, getFunc ResourceGetFunc) error {
```
innobead, line 1012: Can we rename it to `WaitForResourceDeletion`?

innobead (changes requested):
LGTM, just one minor suggestion.

innobead:
@mergify backport v1.5.x

util/util.go
derekbit: Renamed. Thank you.

innobead (approved):
LGTM

PhanLe1010:
From my testing, waiting for the deployment to be removed is not enough. The pod of the deployment might still take some time to be terminated. Maybe we have to wait for the pod to be removed @derekbit?

PR #2075, opened 2023-07-11
Ensure deletion of snapshot CR when it is gone from engine

ejweber:
longhorn/longhorn#6298

### DIFF
--- a/controller/snapshot_controller.go
+++ b/controller/snapshot_controller.go
@@ -318,6 +318,7 @@ func (sc *SnapshotController) reconcile(snapshotName string) (err error) {
 			return sc.ds.RemoveFinalizerForSnapshot(snapshot)
 		}
 
+		// Wait to get engine until now. If isVolDeletedOrBeingDeleted, the engine may not exist.
 		engine, err := sc.getTheOnlyEngineCRforSnapshot(snapshot)
 		if err != nil {
 			return err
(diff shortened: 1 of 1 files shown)


controller/snapshot_controller.go
```diff
@@ -335,6 +336,14 @@ func (sc *SnapshotController) reconcile(snapshotName string) (err error) {
...
 		}()
 
+		// If the snapshot no longer exists in the engine, allow its CR to be deleted.
+		if _, ok := engine.Status.Snapshots[snapshot.Name]; !ok {
+			if err = sc.handleAttachmentTicketDeletion(snapshot); err != nil {
+				return err
+			}
+			return sc.ds.RemoveFinalizerForSnapshot(snapshot)
+		}
+
```
PhanLe1010, line 346: This is a bit dangerous. Sometimes the `engine.Status.Snapshot` info might be clear (live migration, live upgrade) but the snapshot still exists on disk. We should not check with the engine process before removing the finalizer

ejweber: I think I agree in principle, but I'm not sure I see where we were previously doing so. Do you mean on 357 when we verify the engine is running?

PhanLe1010: Previously, we make sure engine is running, issue the snapshotdelete request to engine, then check the engine.status.snapshot

This make sure that at least we issue the snapshot delete command to engine process successfully 

But this PR might not attempt that snapshotdelete request

ejweber: We discussed this offline and I am now in agreement. Closing this PR.

PR #2078, opened 2023-07-13
feat(setting): add tolerate node disk empty selector volume setting

ChanYiLin:
ref: [longhorn/longhorn#4826](https://github.com/longhorn/longhorn/issues/4826)


types/types.go
```diff
@@ -995,12 +995,18 @@ func GetLHVolumeAttachmentNameFromVolumeName(volName string) string {
...
 	if !sort.StringsAreSorted(tags) {
 		logrus.Debug("BUG: Tags are not sorted, sorting now")
 		sort.Strings(tags)
 	}
 
+	if !tolerateEmptySelector {
+		if len(selectors) == 0 && len(tags) != 0 {
+			return false
+		}
+	}
```
shuo-wu, line 1008: NIT: I would prefer the following format. Since `len(selectors) == 0` is the only exception of the below code `for _, selector := range selectors`. Listing this special case handling together with the general handling makes the code more easy-to-understand.
```
	if len(selectors) == 0 {
		if !tolerateEmptySelector && len(tags) != 0 {
			return false
		}
	}
```

types/setting.go
```diff
@@ -1103,6 +1109,26 @@ var (
 		ReadOnly:    true,
 		Default:     "1024",
 	}
+
+	SettingDefinitionTolerateEmptyNodeSelectorVolume = SettingDefinition{
+		DisplayName: "Tolerate Empty Node Selector Volume",
```
shuo-wu, line 1114: NIT: How about renaming this to `Allow scheduling empty selector volumes to any nodes`? The first glance of `Tolerate` mislead me that it is about the Kubernetes taint & toleration feature LOL

shuo-wu (dismissed):
In general LGTM

types/setting.go
ChanYiLin: Yes, I got this naming from the `taint & toleration`
but your are right it might confuse users

innobead: @ChanYiLin Also need to change the setting to `SettingAllowEmptyNodeSelectorVolume`. Please re-check the places needed to update.

controller/volume_controller.go
```diff
@@ -2304,10 +2304,15 @@ func (c *VolumeController) listReadySchedulableAndScheduledNodes(volume *longhor
 		return nil, err
 	}
 
+	tolerateEmptyNodeSelectorVolume, err := c.ds.GetSettingAsBool(types.SettingNameTolerateEmptyNodeSelectorVolume)
+	if err != nil {
+		logrus.Errorf("Failed to get tolerate empty node selector volume setting: %v", err)
```
innobead, line 2309: Can we use `log` instead? even though this is a general message.

scheduler/replica_scheduler.go
```diff
@@ -147,7 +147,7 @@ func (rcs *ReplicaScheduler) getDiskCandidates(nodeInfo map[string]*longhorn.Nod
 
 	nodeSoftAntiAffinity, err := rcs.ds.GetSettingAsBool(types.SettingNameReplicaSoftAntiAffinity)
 	if err != nil {
-		logrus.Errorf("error getting replica soft anti-affinity setting: %v", err)
+		logrus.Errorf("Failed to get replica soft anti-affinity setting: %v", err)
```
innobead, line 150: `withErr` instead?

scheduler/replica_scheduler.go
```diff
@@ -157,7 +157,7 @@ func (rcs *ReplicaScheduler) getDiskCandidates(nodeInfo map[string]*longhorn.Nod
 
 	zoneSoftAntiAffinity, err := rcs.ds.GetSettingAsBool(types.SettingNameReplicaZoneSoftAntiAffinity)
 	if err != nil {
-		logrus.Errorf("Error getting replica zone soft anti-affinity setting: %v", err)
+		logrus.Errorf("Failed to get replica zone soft anti-affinity setting: %v", err)
```
innobead, line 160: `withErr` instead?

scheduler/replica_scheduler.go
```diff
@@ -291,6 +296,11 @@ func (rcs *ReplicaScheduler) filterNodeDisksForReplica(node *longhorn.Node, disk
 	multiError = util.NewMultiError()
 	preferredDisks = map[string]*Disk{}
 
+	tolerateEmptyDiskSelectorVolume, err := rcs.ds.GetSettingAsBool(types.SettingNameTolerateEmptyDiskSelectorVolume)
+	if err != nil {
+		logrus.Errorf("Failed to get tolerate empty tag volume setting: %v", err)
```
innobead, line 301: Make the message exact. Need to return err here? (the same for other places)

`"Failed to get % setting", types.SettingNameTolerateEmptyDiskSelectorVolume`

scheduler/replica_scheduler.go
```diff
@@ -559,6 +569,11 @@ func (rcs *ReplicaScheduler) isFailedReplicaReusable(r *longhorn.Replica, v *lon
 		return false
 	}
 
+	tolerateEmptyDiskSelectorVolume, err := rcs.ds.GetSettingAsBool(types.SettingNameTolerateEmptyDiskSelectorVolume)
+	if err != nil {
+		logrus.Errorf("Failed to get tolerate empty tag volume setting: %v", err)
```
innobead, line 574: Should we change `isFailedReplicaReusable` to return bool and error instead? so here, if encountering an error when getting the setting, then we can just return the error back, and no need to print log here.

Or this is intended to ignore the error? if yes, why?

innobead (changes requested):

scheduler/replica_scheduler.go
ChanYiLin: originally, I plan to just log error and use the default `false` returned from GetSettingAsBool  like what we did before for `zoneSoftAntiAffinity` and `nodeSoftAntiAffinity` above
```
	zoneSoftAntiAffinity, err := rcs.ds.GetSettingAsBool(types.SettingNameReplicaZoneSoftAntiAffinity)
	if err != nil {
		logrus.Errorf("Error getting replica zone soft anti-affinity setting: %v", err)
        }
```
but you are right, we should return error so we don't get confused behavior when setting is not found
change `zoneSoftAntiAffinity` and `nodeSoftAntiAffinity`  as well

types/types.go
```diff
@@ -995,12 +995,18 @@ func GetLHVolumeAttachmentNameFromVolumeName(volName string) string {
...
 	if !sort.StringsAreSorted(tags) {
 		logrus.Debug("BUG: Tags are not sorted, sorting now")
 		sort.Strings(tags)
 	}
 
+	if len(selectors) == 0 {
+		if !tolerateEmptySelector && len(tags) != 0 {
+			return false
+		}
+	}
```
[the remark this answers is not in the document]
ChanYiLin: because this condition may miss the following case
```
tags = ["A"], selector = [], emptySelectorAllowed=false
```
and since the selector is empy so it also skips the following iteration of selector list
then return true directly, which should be falise since tags = ["A"] is not empy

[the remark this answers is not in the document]
ChanYiLin: when selectors is empty
- if not allow empty selector
    - if tags = [] => return true
    - if tags = ['A'] => return false

There are still two condition when not allow empty selector

types/setting.go
```diff
@@ -107,6 +107,8 @@ const (
 	SettingNameV2DataEngine                                             = SettingName("v2-data-engine")
 	SettingNameV2DataEngineHugepageLimit                                = SettingName("v2-data-engine-hugepage-limit")
 	SettingNameOfflineReplicaRebuilding                                 = SettingName("offline-replica-rebuilding")
+	SettingNameTolerateEmptyNodeSelectorVolume                          = SettingName("tolerate-empty-node-selector-volume")
+	SettingNameTolerateEmptyDiskSelectorVolume                          = SettingName("tolerate-empty-disk-selector-volume")
```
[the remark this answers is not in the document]
ChanYiLin: Fixed

scheduler/replica_scheduler.go
ChanYiLin: Fixed and return error directly

scheduler/replica_scheduler.go
ChanYiLin: Fixed and return error directly

scheduler/replica_scheduler.go
ChanYiLin: Fixed.

scheduler/replica_scheduler.go
ChanYiLin: Fixed, and return the error when setting is not found

scheduler/replica_scheduler.go
```diff
@@ -291,6 +302,11 @@ func (rcs *ReplicaScheduler) filterNodeDisksForReplica(node *longhorn.Node, disk
 	multiError = util.NewMultiError()
 	preferredDisks = map[string]*Disk{}
 
+	allowEmptyDiskSelectorVolume, err := rcs.ds.GetSettingAsBool(types.SettingNameAllowEmptyDiskSelectorVolume)
+	if err != nil {
+		logrus.Errorf("Failed to get allow empty tag volume setting: %v", err)
```
innobead, line 307: return error here and no need to log error here, because the caller will handle the returned error.

innobead (changes requested):

innobead:
@ChanYiLin Also, the chart manifests need to update.

innobead (approved):

PR #2079, opened 2023-07-14
api: generalize writeErr function

derekbit:
Longhorn/longhorn#6319

### DIFF
--- a/api/router.go
+++ b/api/router.go
@@ -19,15 +19,29 @@ func HandleError(s *client.Schemas, t HandleFuncWithError) http.Handler {
 		if err := t(rw, req); err != nil {
 			logrus.WithError(err).Warnf("HTTP handling error")
 
+			statusCode := http.StatusInternalServerError
 			if datastore.ErrorIsNotFound(err) {
-				rw.WriteHeader(http.StatusNotFound)
-				return
+				statusCode = http.StatusNotFound
 			}
-
-			apiContext := api.GetApiContext(req)
-			apiContext.WriteErr(err)
+			writeErr(rw, req, err, statusCode)
 		}
 	}))
+}
+
+func writeErr(rw http.ResponseWriter, req *http.Request, err error, statusCode int) {
+	apiContext := api.GetApiContext(req)
+	rw.WriteHeader(statusCode)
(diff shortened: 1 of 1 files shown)


derekbit:
Waiting for https://ci.longhorn.io/job/private/job/longhorn-tests-regression/4586/console

shuo-wu (dismissed):

api/router.go
```diff
@@ -19,17 +19,33 @@ func HandleError(s *client.Schemas, t HandleFuncWithError) http.Handler {
...
-
-			apiContext := api.GetApiContext(req)
-			apiContext.WriteErr(err)
+			writeErr(rw, req, err, statusCode)
 		}
 	}))
 }
 
+func writeErr(rw http.ResponseWriter, req *http.Request, err error, statusCode int) {
+	logrus.WithError(err).Errorf("Error in request")
```
innobead, line 32: The error has been logged by the caller, so here it will cause a redundant log.

innobead (changes requested):
LGTM, just one comment. 

P.S. this is good to improve this by duplicating the implementation from the dependent lib.

api/router.go
derekbit: Update the log message, should be 
```
	if writeErr != nil {
		logrus.WithError(writeErr).Error("Failed to write error %v", err)
	}
```t
The second error is printed only when failing to write error

ChanYiLin:
LGTM! Thanks!

ChanYiLin (approved):

api/router.go
innobead: What I mean is `logrus.WithError(err).Errorf("Error in request")` in `writeErr` is not required, since it has been logged in the caller `logrus.WithError(err).Warnf("HTTP handling error")`.

innobead (changes requested):
Only one left comment https://github.com/longhorn/longhorn-manager/pull/2079#discussion_r1264659264? then we can merge this PR.

innobead:
@mergify backport v1.5.x

api/router.go
derekbit: Oh, yes. It's redundant. Removed, thank you.

innobead (approved):

PR #2080, opened 2023-07-14
Add finalizer with webhook instead of controllers

ejweber:
longhorn/longhorn#4872

This work was started in https://github.com/longhorn/longhorn-manager/pull/1535. In that PR, we ensured that the webhook only added the finalizer during CREATE operations and not during UPDATE operations to avoid resources that cannot be deleted (e.g. we attempt to remove a resource's finalizer, but the finalizer is added back when we do so). After reading some discussion on the internet and thinking on it a bit, I decided a better approach would be for the webhook to always add the finalizer on both CREATE and UPDATE except when the deletion timestamp is set. After reviewing our code, every time longhorn-manager removes the finalizer, the deletion timestamp has already been set on an object. This approach has two benefits:

- The finalizer cannot be accidentally deleted by a user unless the object is already being deleted. Previously, a user could update an object so that it did not have a finalizer. This could have been problematic later when the object was deleted.
- Consistency/code cleanliness. We always do the same thing on CREATE and UPDATE and can share logic between the two.

Otherwise, I also tried to clean up the webhook resources slightly. Code was organized differently between resources and some code was duplicated. Since I was already doing a webhook refactor, I spent some time (hopefully) improving this.

Keeping this as a draft until regression tests pass: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/4596/console.

### DIFF
--- a/webhook/resources/backingimagedatasource/mutator.go
+++ b/webhook/resources/backingimagedatasource/mutator.go
@@ -3,12 +3,16 @@ package backingimagedatasource
 import (
 	"fmt"
 
+	"github.com/pkg/errors"
+
 	admissionregv1 "k8s.io/api/admissionregistration/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 
 	"github.com/longhorn/longhorn-manager/datastore"
 	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
 	"github.com/longhorn/longhorn-manager/webhook/admission"
+	"github.com/longhorn/longhorn-manager/webhook/common"
+	werror "github.com/longhorn/longhorn-manager/webhook/error"
 )
 
 type backingImageDataSourceMutator struct {
@@ -35,24 +39,34 @@ func (b *backingImageDataSourceMutator) Resource() admission.Resource {
 }
 
 func (b *backingImageDataSourceMutator) Create(request *admission.Request, newObj runtime.Object) (admission.PatchOps, error) {
-	return mutateBackingImageDatasource(newObj)
+	return mutate(newObj)
 }
 
 func (b *backingImageDataSourceMutator) Update(request *admission.Request, oldObj runtime.Object, newObj runtime.Object) (admission.PatchOps, error) {
-	return mutateBackingImageDatasource(newObj)
+	return mutate(newObj)
 }
 
-func mutateBackingImageDatasource(newObj runtime.Object) (admission.PatchOps, error) {
+// mutate contains functionality shared by Create and Update.
+func mutate(newObj runtime.Object) (admission.PatchOps, error) {
 	var patchOps admission.PatchOps
 
-	backingimagedatasource := newObj.(*longhorn.BackingImageDataSource)
+	backingImageDataSource := newObj.(*longhorn.BackingImageDataSource)
 
-	if backingimagedatasource.Spec.SourceType == "" {
+	if backingImageDataSource.Spec.SourceType == "" {
 		patchOps = append(patchOps, fmt.Sprintf(`{"op": "replace", "path": "/spec/sourceType", "value": "%s"}`, longhorn.BackingImageDataSourceTypeDownload))
 	}
 
-	if backingimagedatasource.Spec.Parameters == nil {
+	if backingImageDataSource.Spec.Parameters == nil {
 		patchOps = append(patchOps, `{"op": "replace", "path": "/spec/parameters", "value": {}}`)
+	}
+
+	patchOp, err := common.GetLonghornFinalizerPatchOpIfNeeded(backingImageDataSource)
+	if err != nil {
+		err := errors.Wrapf(err, "failed to get finalizer patch for backingImageDataSource %v", backingImageDataSource.Name)
+		return nil, werror.NewInvalidError(err.Error(), "")
+	}
+	if patchOp != "" {
+		patchOps = append(patchOps, patchOp)
 	}
 
 	return patchOps, nil
--- a/webhook/resources/volume/mutator.go
+++ b/webhook/resources/volume/mutator.go
@@ -48,26 +48,9 @@ func (v *volumeMutator) Create(request *admission.Request, newObj runtime.Object
 	volume := newObj.(*longhorn.Volume)
 	var patchOps admission.PatchOps
 
-	var err error
-	if patchOps, err = mutate(newObj); err != nil {
-		return nil, err
-	}
-
 	name := util.AutoCorrectName(volume.Name, datastore.NameMaximumLength)
 	if name != volume.Name {
 		patchOps = append(patchOps, fmt.Sprintf(`{"op": "replace", "path": "/metadata/name", "value": "%s"}`, name))
-	}
-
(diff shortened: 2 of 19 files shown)


ejweber:
I tested this PR by deploying an updated longhorn-manager and creating/editing/deleting resources. Generally, each resource should have a finalizer added on creation. It should be impossible to remove the finalizer while the resource's deletion timestamp is not set, and the resource should still clean up as normal. See specific results below:

```
backingimage
(created through UI with url: https://longhorn-backing-image.s3-us-west-1.amazonaws.com/parrot.qcow2 and checksum: bd79ab9e6d45abf4f3f0adf552a868074dd235c4698ce7258d521160e0ad79ffe555b94e7d4007add6e1a25f4526885eb25c53ce38f7d344dd4925b9f2cb5d3b)
has finalizer x
doesn't allow finalizer removal while not deleting x
cleans up when directly deleted x

backingimagedatasource
has finalizer x
doesn't allow finalizer removal while not deleting x
cleans up when backing image is deleted x

backingimagemanager
has finalizer x
doesn't allow finalizer removal while not deleting x
cleans up when backing image is deleted x

backup
has finalizer x
doesn't allow finalizer removal while not deleting x
cleans up when directly deleted x

backupvolume
has finalizer x
doesn't allow finalizer removal while not deleting x
cleans up when directly deleted x

engine
has finalizer x
doesn't allow finalizer removal while not deleting x
cleans up when volume deleted x

engineimage
has finalizer x
doesn't allow finalizer removal while not deleting x
cleans up during uninstall ?

node
has finalizer x
doesn't allow finalizer removal while not deleting x
cleans up when directly deleted x

replica
has finalizer x
doesn't allow finalizer removal while not deleting x
cleans up when volume deleted x

sharemanager
has finalizer x
doesn't allow finalizer removal while not deleting x
cleans up when rwx volume deleted x

snapshot
has finalizer x
doesn't allow finalizer removal while not deleting x

systembackup
has finalizer x
doesn't allow finalizer removal while not deleting x
cleans up when directly deleted x

volume
has finalizer x
doesn't allow finalizer removal while not deleting x
cleans up when directly deleted x

volumeattachment
has finalizer x
doesn't allow finalizer removal while not deleting x
cleans up when volume deleted x
```

Note: any time we edit a resource to remove its finalizer but don't delete it, we should see something like the following in the longhorn-manager logs:

```
time="2023-07-13T20:42:05Z" level=info msg="Request (user: user-gkb8z, longhorn.io/v1beta2, Kind=BackingImage, namespace: longhorn-system, name: test-bi, operation: UPDATE) patchOps: [{\"op\": \"replace\", \"path\": \"/metadata/finalizers\", \"value\": [\"longhorn.io\"]}]" service=admissionWebhook
```

ejweber:
These tests were originally failing due to a mistake I made in the volume controller webhook:

- [tests.test_basic.test_restore_inc[s3]](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/4596/testReport/junit/tests/test_basic/test_restore_inc_s3_/)
- [tests.test_basic.test_restore_inc[nfs]](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/4596/testReport/junit/tests/test_basic/test_restore_inc_nfs_/)
- [tests.test_basic.test_restore_inc_with_offline_expansion[s3]](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/4596/testReport/junit/tests/test_basic/test_restore_inc_with_offline_expansion_s3_/)
- [tests.test_basic.test_restore_inc_with_offline_expansion[nfs]](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/4596/testReport/junit/tests/test_basic/test_restore_inc_with_offline_expansion_nfs_/)

Fixed and passing at:

- https://ci.longhorn.io/job/private/job/longhorn-tests-regression/4610/
- https://ci.longhorn.io/job/private/job/longhorn-tests-regression/4611/

webhook/resources/backingimagedatasource/mutator.go
```diff
@@ -3,12 +3,14 @@ package backingimagedatasource
...
-	"k8s.io/apimachinery/pkg/runtime"
-
 	"github.com/longhorn/longhorn-manager/datastore"
 	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
 	"github.com/longhorn/longhorn-manager/webhook/admission"
+	"github.com/longhorn/longhorn-manager/webhook/common"
+	werror "github.com/longhorn/longhorn-manager/webhook/error"
+	"github.com/pkg/errors"
+	admissionregv1 "k8s.io/api/admissionregistration/v1"
+	"k8s.io/apimachinery/pkg/runtime"
```
james-munson, line 13: Is there a convention in longhorn about clumping the longhorn and non-longhorn import paths separately?  Most of the files in this PR are that way and it is left untouched, but a couple of these have them glommed together like this one -  perhaps by a tool?  (I prefer the separation, myself.)

derekbit: Here is an example
https://github.com/longhorn/longhorn-manager/blob/master/controller/volume_controller.go#L3-L42

ejweber: I see. Yes, on past projects I worked on all team members were using the same Go linter (whatever VSCode uses by default) and we agreed to let it handle import clumping/ordering. It looks like I defaulted to that here instead of following our convention. Fixing.

shuo-wu (dismissed):
LGTM

webhook/resources/volume/mutator.go
```diff
@@ -260,40 +198,79 @@ func (v *volumeMutator) Create(request *admission.Request, newObj runtime.Object
...
 
 func (v *volumeMutator) Update(request *admission.Request, oldObj runtime.Object, newObj runtime.Object) (admission.PatchOps, error) {
+	volume := newObj.(*longhorn.Volume)
 	var patchOps admission.PatchOps
 
+	if volume.Spec.AccessMode == "" {
+		patchOps = append(patchOps, `{"op": "replace", "path": "/spec/accessMode", "value": "rwo"}`)
+	}
+
+	size := util.RoundUpSize(volume.Spec.Size)
```
PhanLe1010, line 221: Looks like this one is duplicated with https://github.com/longhorn/longhorn-manager/blob/f87d716b8b23b3b0463b588df1a7e4eab34f8f4a/webhook/resources/volume/mutator.go#L221

PhanLe1010 (commented):
In general LGTM.

Can we make sure add this uninstallation test plan:
1. Install Longhorn with all possible CRDs (volumes, snapshot, backups, backing image, recurring jobs, ...)
2. Make sure Uninstallation is ok

james-munson (approved):
An idle question / whine, but LGTM.

webhook/resources/volume/mutator.go
ejweber: You are right. Since create and update have different logic around this (with create potentially writing a new size for the volume and update simply rounding it up) I have removed this action from the shared logic section in an amend.

PhanLe1010 (approved):
LGTM

PR #2086, opened 2023-07-17
fix(backup): set backup error if backup target invalid

mantissahz:
When starting a backup monitor, it will start to take a snapshot backup procedure and it will be failed if the s3 backup target is invalid (nfs backup target will be not be failed because of timeout and test the backup when it is in progress.)
Set the backup state as `Error` if starting the backup procedure returns an error.

Ref: longhorn/longhorn#1249

### DIFF
--- a/controller/backup_controller.go
+++ b/controller/backup_controller.go
@@ -375,6 +375,10 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
 
 		monitor, err := bc.checkMonitor(backup, volume, backupTarget)
 		if err != nil {
+			if backup.Status.State == longhorn.BackupStateError {
+				log.WithError(err).Warnf("Failed to enable the backup monitor for backup %v", backup.Name)
+				return nil
+			}
 			return err
 		}
 


ejweber:
Related to (and possible fixes) https://github.com/longhorn/longhorn/issues/6358.

cc @PhanLe1010

PhanLe1010:
Yeah, this is one way to fix it. I am fine with either this approach or https://github.com/longhorn/longhorn/issues/6358#issuecomment-1644382633. 

Leave it to @mantissahz and @ejweber to make the call :X

ejweber (commented):
I think if we do this, the backup can never complete, even if the backup target becomes valid later. 

https://github.com/longhorn/longhorn-manager/blob/ce99ddf4c6fa8ef3a50bb623c23d17df826e4250/controller/backup_controller.go#L347-L354
That was fine for me in [#6358](https://github.com/longhorn/longhorn/issues/6358#issuecomment-1644382633) because the snapshot didn't exist and the situation was not recoverable. 

Do we need to distinguish between recoverable and unrecoverable situations?

PhanLe1010:
> I think if we do this, the backup can never complete, even if the backup target becomes valid later.

It is fine to me

PhanLe1010:
I am taking it back. May be this backup target not setup error should recoverable

mantissahz:
@PhanLe1010 Do you mean that we should have some retries for the backup when encountering a backup target is invalid to check if the backup target is recoverable?

shuo-wu (approved):
The PR itself LGTM (without considering https://github.com/longhorn/longhorn/issues/6358#issuecomment-1644382633). 

Without this fix, the backup CR can be never handled. The reconcile loop will directly error out rather than updating the CR state to `Error`.

ejweber (approved):
Approving this and will test it as the fix to https://github.com/longhorn/longhorn/issues/6358. When it merges.

I am still thinking it would be better to have the ability to retry if the backup target becomes valid. But to @shuo-wu's point, without making this change, we can never know a backup is in error and we will keep silently trying over and over to do an impossible thing (this is the underlying complaint in https://github.com/longhorn/longhorn/issues/6358). Maybe in the future we can consider an addition to backup monitor reconcile loop that can recognize if a backup's error is related to an invalid backup target and retrying if the backup target is now valid.

innobead (approved):

innobead:
> Approving this and will test it as the fix to [longhorn/longhorn#6358](https://github.com/longhorn/longhorn/issues/6358) when it merges.
> 
> I am still thinking it would be better to have the ability to retry if the backup target becomes valid. But to @shuo-wu's point, without making this change, we can never know a backup is in error and we will keep silently trying over and over to do an impossible thing (this is the underlying complaint in [longhorn/longhorn#6358](https://github.com/longhorn/longhorn/issues/6358)). Maybe in the future we can consider an addition to backup monitor reconcile loop that can recognize if a backup's error is related to an invalid backup target and retrying if the backup target is now valid.

Agreed with this point. @ejweber please create a ticket for this. This should be part of how resilient the system/function we want to make. Any operation should be considered with retry, backoff, etc if suitable/could. Surely, the audit should be clear during the resilience period.

PR #2089, opened 2023-07-27
Improve log and minor cleanup

Vicente-Cheng:
related: https://github.com/longhorn/longhorn/issues/6396

I tried to modify some log outputs and added some logs for more clarity for investigation.

And also, do some minor changes as below:
1.  Reorder the checking order on `isResponsibleFor()`. If the `currentNodeEngineAvailable` is false we could conclude that it would be false.
2. Do not wait for all replicas to be running before we change the engine state to running. We could change the engine's desired state to running if we have enough replicas instead of all replicas.

Please help to review. Thanks!

### DIFF
--- a/controller/engine_controller.go
+++ b/controller/engine_controller.go
@@ -2073,15 +2073,19 @@ func (ec *EngineController) isResponsibleFor(e *longhorn.Engine, defaultEngineIm
 		return isResponsible, nil
 	}
 
+	currentNodeEngineAvailable, err := ec.ds.CheckEngineImageReadiness(e.Status.CurrentImage, ec.controllerID)
+	if err != nil {
+		return false, err
+	}
+	if !currentNodeEngineAvailable {
+		// avoid redundant checking if the current node is not available
+		return false, nil
+	}
 	preferredOwnerEngineAvailable, err := ec.ds.CheckEngineImageReadiness(e.Status.CurrentImage, e.Spec.NodeID)
 	if err != nil {
 		return false, err
 	}
 	currentOwnerEngineAvailable, err := ec.ds.CheckEngineImageReadiness(e.Status.CurrentImage, e.Status.OwnerID)
-	if err != nil {
-		return false, err
-	}
-	currentNodeEngineAvailable, err := ec.ds.CheckEngineImageReadiness(e.Status.CurrentImage, ec.controllerID)
 	if err != nil {
 		return false, err
 	}
--- a/controller/engine_controller.go
+++ b/controller/engine_controller.go
@@ -328,7 +328,7 @@ func (ec *EngineController) syncEngine(key string) (err error) {
 		syncReplicaAddressMap = true
 	}
 	if syncReplicaAddressMap && !reflect.DeepEqual(engine.Status.CurrentReplicaAddressMap, engine.Spec.ReplicaAddressMap) {
-		log.Infof("Updating engine current replica address map to %+v", engine.Spec.ReplicaAddressMap)
+		log.Infof("Updating engine current replica address map to %+v. We need to wait for all replicas running then engine will be running.", engine.Spec.ReplicaAddressMap)
 		engine.Status.CurrentReplicaAddressMap = engine.Spec.ReplicaAddressMap
 		// Make sure the CurrentReplicaAddressMap persist in the etcd before continue
 		return nil
--- a/controller/node_controller.go
+++ b/controller/node_controller.go
@@ -1014,7 +1014,7 @@ func (nc *NodeController) syncInstanceManagers(node *longhorn.Node) error {
 				}
 			}
 			if cleanupRequired {
-				log.Infof("Cleaning up the redundant instance manager %v when there is no running/starting instance", im.Name)
+				log.Infof("Cleaning up the redundant instance manager %v when there is no running/starting instance: %v", im.Name, dumpInstanceState(im.Status.Instances))
 				if err := nc.ds.DeleteInstanceManager(im.Name); err != nil {
 					return err
 				}
@@ -1564,4 +1564,12 @@ func (nc *NodeController) shouldEvictReplica(node *longhorn.Node, kubeNode *core
 	}
 
 	return false, constant.EventReasonEvictionCanceled, nil
+}
+
+func dumpInstanceState(instanceMap map[string]longhorn.InstanceProcess) map[string]string {
+	output := map[string]string{}
+	for _, instance := range instanceMap {
+		output[instance.Spec.Name] = string(instance.Status.State)
+	}
+	return output
 }
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -1297,6 +1297,7 @@ func (c *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[strin
 					log.WithField("replica", r.Name).WithError(err).Warnf("Failed to check if node %v is still running for failed replica", r.Spec.NodeID)
 					continue
 				} else if isDownOrDeleted {
+					log.WithField("replica", r.Name).Warnf("Node %v is NotReady or Deleted for failed replica", r.Spec.NodeID)
 					continue
 				}
 				node, err := c.ds.GetNodeRO(r.Spec.NodeID)
@@ -1313,6 +1314,7 @@ func (c *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[strin
 					}
 				}
 				if !diskSchedulable {
+					log.WithField("replica", r.Name).Warnf("Disk %v is not schedulable for failed replica", r.Spec.DiskID)
 					continue
 				}
 				failedAt, err := util.ParseTime(r.Spec.FailedAt)
@@ -1921,11 +1923,13 @@ func (c *VolumeController) reconcileVolumeSize(v *longhorn.Volume, e *longhorn.E
 }
 
 func (c *VolumeController) canInstanceManagerLaunchReplica(r *longhorn.Replica) (bool, error) {
+	log := c.logger.WithField("replica", r.Name)
 	nodeDown, err := c.ds.IsNodeDownOrDeleted(r.Spec.NodeID)
 	if err != nil {
 		return false, errors.Wrapf(err, "fail to check IsNodeDownOrDeleted %v", r.Spec.NodeID)
 	}
 	if nodeDown {
+		log.Warnf("Cannot launch replica due to node %v is NotReady or Deleted", r.Spec.NodeID)
 		return false, nil
 	}
 	// Replica already had IM
@@ -4074,12 +4078,14 @@ func (c *VolumeController) IsReplicaUnavailable(r *longhorn.Replica) (bool, erro
 	if r.Spec.NodeID == "" || r.Spec.DiskID == "" || r.Spec.DiskPath == "" || r.Spec.DataDirectoryName == "" {
 		return true, nil
 	}
+	log := c.logger.WithField("replica", r.Name)
 
 	isDownOrDeleted, err := c.ds.IsNodeDownOrDeleted(r.Spec.NodeID)
 	if err != nil {
 		return true, errors.Wrapf(err, "failed  to check if node %v is still running for failed replica %v", r.Spec.NodeID, r.Name)
 	}
 	if isDownOrDeleted {
+		log.Warnf("Replica is unavailable due to Node %v is NotReady or Deleted", r.Spec.NodeID)
 		return true, nil
 	}
 


controller/node_controller.go
```diff
@@ -985,7 +984,7 @@ func (nc *NodeController) syncInstanceManagers(node *longhorn.Node) error {
 				}
 			}
 			if cleanupRequired {
-				logrus.Infof("Cleaning up the redundant instance manager %v when there is no running/starting instance", im.Name)
+				logrus.Infof("Cleaning up the redundant instance manager %v when there is no running/starting instance: %+v", im.Name, im.Status.Instances)
```
innobead, line 987: `im.Status.Instances` could be a lot, so it would overwhelm logs. What exact things you would like to look into like instances with some specific state? probably intentionally log them instead and use debug level log.

datastore/longhorn.go
```diff
@@ -2430,6 +2429,7 @@ func (s *DataStore) IsNodeDownOrDeleted(name string) (bool, error) {
 	if cond.Status == longhorn.ConditionStatusFalse &&
 		(cond.Reason == string(longhorn.NodeConditionReasonKubernetesNodeGone) ||
 			cond.Reason == string(longhorn.NodeConditionReasonKubernetesNodeNotReady)) {
+		logrus.Debugf("Node is marked down due to %s", cond.Reason)
```
innobead, line 2432: Should not log here but the caller instead, because this is just a query method.

controller/volume_controller.go
```diff
@@ -1730,6 +1729,9 @@ func (c *VolumeController) openVolumeDependentResources(v *longhorn.Volume, e *l
 		}
 		replicaAddressMap[r.Name] = imutil.GetURL(r.Status.StorageIP, r.Status.Port)
 	}
+	if !anyRunningReplica {
```
innobead, line 1732: Don't need this variable, because you can just rely on the below condition. If `len(replicaAddressMap) != 0` already means there is at least one healthy inside?

innobead (changes requested):

controller/volume_controller.go
```diff
@@ -1712,9 +1708,12 @@ func (c *VolumeController) openVolumeDependentResources(v *longhorn.Volume, e *l
 		if r.Status.CurrentState == longhorn.InstanceStateError {
 			continue
 		}
-		// wait for all potentially healthy replicas become running
-		if r.Status.CurrentState != longhorn.InstanceStateRunning {
-			return nil
+		// wait at least one replica is running
```
innobead, line 1711: @shuo-wu @derekbit @PhanLe1010 WDYT?

controller/volume_controller.go
Vicente-Cheng: Yap, I am also thinking about reusing this condition.
But I felt that checking is error handling instead retry. Then I choose to add a new checking variable.

The original flow is more like:
1. wait for whole replica running
2. if the whole replica running we should not have the situation that replicaAddressMap is empty. So if we meet this situation, just return the error.

If I use the `replicaAddressMap` for checking, the new flow would like
1. keep checking until at least one replica is healthy.
Then we could remove the running checks.

I thought maybe the second would be enough, but originally I would like to keep more code flow as before.
So what do you think about the second change (means only relying on `replicaAddressMap`)?

controller/node_controller.go
Vicente-Cheng: That's a good point. Thanks for it.
I will try to get the fields I need!

controller/volume_controller.go
ejweber: I think this may be dangerous. If all replicas are failed, we may auto-salvage with only one replica. We set `r.Spec.FailedAt = ""` for all replicas we want to autosalvage [here](https://github.com/longhorn/longhorn-manager/blob/6b7d12b397738a6ae759f2414e53f7f1b4ba8f30/controller/volume_controller.go#L1305). [Later](https://github.com/longhorn/longhorn-manager/blob/6b7d12b397738a6ae759f2414e53f7f1b4ba8f30/controller/volume_controller.go#L1353) we call `reconcileAttachDetachStateMachine` which [calls](https://github.com/longhorn/longhorn-manager/blob/6b7d12b397738a6ae759f2414e53f7f1b4ba8f30/controller/volume_controller.go#L1433) `openVolumeDependentResources`, where this change lives. If the most up-to-date replica isn't running, we will auto-salvage without it and lose data. WDYT?

innobead: Good point. I think you could be right.

Q: When auto salvage happens, how are replicas getting rebuilt? Will the volume start with the salvaged replicas first? after those ready, replenish the remaining unscheduled replicas? If yes, the concern still exists?

ejweber: @Vicente-Cheng, it looks like you are referencing [this line of code](https://github.com/longhorn/longhorn-manager/blob/6b7d12b397738a6ae759f2414e53f7f1b4ba8f30/controller/volume_controller.go#L1291), which we use to determine when the last replica failed. However, there is no break from the loop. We continue to check other replicas and add all of them to `failedUsableReplicas`.

[Later](https://github.com/longhorn/longhorn-manager/blob/6b7d12b397738a6ae759f2414e53f7f1b4ba8f30/controller/volume_controller.go#L1303-L1311), we set `r.spec.FailedAt = ""` for ALL replicas that have failed within AutoSalvageTimeLimit (1 minute).

```golang
// Bring up the replicas for auto-salvage
for _, r := range failedUsableReplicas {
	if util.TimestampWithinLimit(lastFailedAt, r.Spec.FailedAt, AutoSalvageTimeLimit) {
		r.Spec.FailedAt = ""
		log.WithField("replica", r.Name).Warn("Automatically salvaging volume replica")
		msg := fmt.Sprintf("Replica %v of volume %v will be automatically salvaged", r.Name, v.Name)
		c.eventRecorder.Event(v, corev1.EventTypeWarning, constant.EventReasonAutoSalvaged, msg)
		salvaged = true
	}
}
```

If I understand correctly, we start the engine with all qualifying replicas, and then let the engine [decide ](https://github.com/longhorn/longhorn-engine/blob/e05a119a9cfab1928b138e036cd1dcd0e97eaf91/pkg/controller/control.go#L580-L610) which replica to use as the source of truth based on `revisionCounter`. Longhorn-manager narrows down the list to only those replicas that failed within a minute of each other, but longhorn-engine makes the final call:

```golang
// checkReplicasRevisionCounter will check if any replica has unmatched
// revision counter, and mark unmatched replica as 'ERR' state.
func (c *Controller) checkReplicasRevisionCounter() error {
	var expectedRevision int64

	revisionCounters := make(map[string]int64)
	for _, r := range c.replicas {
		// The related backend is nil if the mode is ERR
		if r.Mode == types.ERR {
			continue
		}
		counter, err := c.backend.GetRevisionCounter(r.Address)
		if err != nil {
			return err
		}
		if counter > expectedRevision {
			expectedRevision = counter
		}
		revisionCounters[r.Address] = counter
	}

	for address, counter := range revisionCounters {
		if counter != expectedRevision {
			logrus.Errorf("Revision conflict detected! Expect %v, got %v in replica %v. Mark as ERR",
				expectedRevision, counter, address)
			c.setReplicaModeNoLock(address, types.ERR)
		}
	}

	return nil
}

If we start the engine with only one replica, there may be another with a higher revision counter. Do you see it differently?
```

Vicente-Cheng: Hi @ejweber, great analyst. You are correct about it!
I do not take care about the case that all replicas are failed in one minute. Thanks for catching up!

Yes, with the whole code flow.  If all replicas fail in one minute, they will all be the salvage candidates.
Then if we are just waiting for one replica, we cannot use the latest revision to recover.

The original issue I tried to introduce this change for the first-time creation because we sometimes got stuck on the engine with one of the replicas not ready, and the volume could not be active.

Could we still improve the creation and ensure the auto-salvage stage will wait for all replicas?
Like add checking with `e.Spec.SalvageRequested`? WDUT?
@ejweber @innobead @derekbit @shuo-wu @PhanLe1010

ejweber: Seems reasonable to me. Curious what others think. Thanks for looking into this!

shuo-wu: The issue here is not only about the salvage part. It's more about the attachment flow instead.
Starting the engine with at least one replica means that:
1. Other healthy replicas may not connect to the engine when the engine is up and handling IO. Then those replicas should be considered as staled and Longhorn needs to pay more resources to do rebuilding. Besides, skipping waiting for all healthy replicas means the HA is hurt/not guaranteed at the beginning.
2. As Eric mentioned, without the wait, we may miss the correct replica candidates and lead to data loss.

Here, I think the correct way is, if a replica cannot be started and it blocks the volume attachment, we should recognize the case and **mark this kind of replica as `Failed` in time** (after several retries). 
I remembered that there was a ticket for this but I cannot find it. And I am not sure if there is a better way to improve this part

controller/volume_controller.go
```diff
@@ -1714,6 +1709,7 @@ func (c *VolumeController) openVolumeDependentResources(v *longhorn.Volume, e *l
 		}
 		// wait for all potentially healthy replicas become running
 		if r.Status.CurrentState != longhorn.InstanceStateRunning {
+			log.WithField("replica", r.Name).Debug("Replica is not running yet, just skip this round reconcile.")
```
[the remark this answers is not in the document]
Vicente-Cheng: OK, that's also a good suggestion. Indeed, that would flood the logs when the replica stuck could be a potential risk for log rotation. So as the @shuo-wu suggested, we could describe more when the engine updates the replica map, and the user would quickly check the replica status if the engine is not running.

controller/engine_controller.go
```diff
@@ -9,9 +9,12 @@ import (
 	"sync"
 	"time"
 
+	etypes "github.com/longhorn/longhorn-engine/pkg/types"
+	imapi "github.com/longhorn/longhorn-instance-manager/pkg/api"
+	imclient "github.com/longhorn/longhorn-instance-manager/pkg/client"
+	imutil "github.com/longhorn/longhorn-instance-manager/pkg/util"
```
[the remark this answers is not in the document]
Vicente-Cheng: OK, thanks for the reminder! Updated

controller/volume_controller.go
Vicente-Cheng: Agree that. I removed this commit seems there are various points we need to take care of for when the engine could be set as running. It would be better to discuss and improve after that.

I will open an enhancement issue to record it. Thanks for your opinions!

controller/node_controller.go
```diff
@@ -1014,7 +1014,7 @@ func (nc *NodeController) syncInstanceManagers(node *longhorn.Node) error {
 				}
 			}
 			if cleanupRequired {
-				log.Infof("Cleaning up the redundant instance manager %v when there is no running/starting instance", im.Name)
+				log.Infof("Cleaning up the redundant instance manager %v when there is no running/starting instance: %v", im.Name, dumpInstanceState(im.Status.Instances))
```
innobead, line 1017: What's the purpose of dumping all instances' state here? Are you looking to determine if any running or starting instances are still present in rare cases?

This could generate a numerous log if there are many instances inside, although this should rarely happen, only when we have multiple IM pods

controller/volume_controller.go
```diff
@@ -1921,11 +1923,13 @@ func (c *VolumeController) reconcileVolumeSize(v *longhorn.Volume, e *longhorn.E
 }
 
 func (c *VolumeController) canInstanceManagerLaunchReplica(r *longhorn.Replica) (bool, error) {
+	log := c.logger.WithField("replica", r.Name)
 	nodeDown, err := c.ds.IsNodeDownOrDeleted(r.Spec.NodeID)
 	if err != nil {
 		return false, errors.Wrapf(err, "fail to check IsNodeDownOrDeleted %v", r.Spec.NodeID)
 	}
 	if nodeDown {
+		log.Warnf("Cannot launch replica due to node %v is NotReady or Deleted", r.Spec.NodeID)
```
innobead, line 1932: This log might be misleading because it's not an action but rather a query. Typically, for query functions/methods (such as is/get/check/...), I would suggest avoiding logging inside them except using lower log levels (ex: debug, trace..). Instead, it's better to utilize returned variables including error, allowing the caller to determine the reason. This approach could also prevent unexpected log flooding in case it will be used frequently.

Since the only caller already dumps the warning log as below, do you still this we need this log here in this method?

[/controller/volume_controller.go#L1691-L1725](https://github.com/longhorn/longhorn-manager/blob/7496347232f4f98ad51a1c5b35c2599db896499a/controller/volume_controller.go#L1691-L1725)
```go
		canIMLaunchReplica, err := c.canInstanceManagerLaunchReplica(r)
		if err != nil {
			return err
		}
		if canIMLaunchReplica {
			if r.Spec.FailedAt == "" && r.Spec.Image == v.Status.CurrentImage {
				if r.Status.CurrentState == longhorn.InstanceStateStopped {
					r.Spec.DesireState = longhorn.InstanceStateRunning
				}
			}
		} else {
			// wait for IM is starting when volume is upgrading
			if c.isVolumeUpgrading(v) {
				continue
			}

			// If the engine isn't attached or the node goes down, the replica can be marked as failed.
			// In the attached mode, we can determine whether the replica can fail by relying on the data plane's connectivity status.
			nodeDeleted, err := c.ds.IsNodeDeleted(r.Spec.NodeID)
			if err != nil {
				return err
			}

			if v.Status.State != longhorn.VolumeStateAttached || nodeDeleted {
				msg := fmt.Sprintf("Replica %v is marked as failed since the volume %v is not attached because the instance manager is unable to launch the replica", r.Name, v.Name)
				if nodeDeleted {
					msg = fmt.Sprintf("Replica %v is marked as failed since the node %v is deleted.", r.Name, r.Spec.NodeID)
				}
				log.WithField("replica", r.Name).Warn(msg)
				if r.Spec.FailedAt == "" {
					r.Spec.FailedAt = c.nowHandler()
				}
				r.Spec.DesireState = longhorn.InstanceStateStopped
			}
		}
```

controller/volume_controller.go
```diff
@@ -4074,12 +4078,14 @@ func (c *VolumeController) IsReplicaUnavailable(r *longhorn.Replica) (bool, erro
...
 		return true, nil
 	}
+	log := c.logger.WithField("replica", r.Name)
 
 	isDownOrDeleted, err := c.ds.IsNodeDownOrDeleted(r.Spec.NodeID)
 	if err != nil {
 		return true, errors.Wrapf(err, "failed  to check if node %v is still running for failed replica %v", r.Spec.NodeID, r.Name)
 	}
 	if isDownOrDeleted {
+		log.Warnf("Replica is unavailable due to Node %v is NotReady or Deleted", r.Spec.NodeID)
```
innobead, line 4088: The same, suggest logging at the caller. If you need a specific reason when the bool var is true, leverage the returned error or even a third reason returned variable.

innobead (changes requested):

controller/volume_controller.go
Vicente-Cheng: Hi @innobead, 

I checked again with the whole logic, and the above error message is only for the node deleted.
Could we use the `c.ds.IsNodeDownOrDeleted(r.Spec.NodeID)` instead `c.ds.IsNodeDeleted(r.Spec.NodeID)` and make minor logging improvements here?

controller/volume_controller.go
Vicente-Cheng: Thanks for the reminder, I would like to wrap the error and return it.

PR #2093, opened 2023-07-28
fix(telemetric): collect SettingNameSnapshotDataIntegrityCronJob setting as boolean

c3y1huang:
https://github.com/longhorn/longhorn/issues/6410

### DIFF
--- a/controller/setting_controller.go
+++ b/controller/setting_controller.go
@@ -1318,6 +1318,7 @@ func (info *ClusterInfo) collectSettings() error {
 		types.SettingNameSystemManagedComponentsNodeSelector: true,
 		types.SettingNameRegistrySecret:                      true,
 		types.SettingNamePriorityClass:                       true,
+		types.SettingNameSnapshotDataIntegrityCronJob:        true,
 		types.SettingNameStorageNetwork:                      true,
 	}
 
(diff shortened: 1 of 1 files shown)


controller/setting_controller.go
```diff
@@ -1387,7 +1387,12 @@ func (info *ClusterInfo) collectSettings() error {
 
 		// Setting that should be collected as boolean (true if configured, false if not)
 		case includeAsBoolean[settingName]:
-			settingMap[setting.Name] = setting.Value != ""
+			definition, ok := types.GetSettingDefinition(types.SettingName(setting.Name))
+			if !ok {
+				logrus.WithError(err).Warnf("Failed to get Setting %v definition", setting.Name)
+				continue
```
PhanLe1010, line 1393: A quick question. Can this case ever happen? I think the setting is already in `includeAsBoolean` which means it should already be in `GetSettingDefinition`, shouldn't it?

c3y1huang: I don't think having the setting in the includeAsBoolean guaranteed its presence in the GetSettingDefinition.  Although this case is unlikely to occur, we should still do a null check to avoid any potential NPE.

shuo-wu (approved):

controller/setting_controller.go
shuo-wu: I think at least it can avoid the NPE issue when we somehow typo or miss the entries of `settingDefinitions`.

PhanLe1010: Sounds good. Thanks all

PhanLe1010 (approved):
LGTM

PhanLe1010:
Could we backport this to v1.5.2?

c3y1huang:
@Mergifyio backport v1.5.x

PR #2094, opened 2023-07-31
Add ReplicaDiskSoftAntiAffinity setting

ejweber:
longhorn/longhorn#3823


ejweber (commented):
Please look carefully at this commit. I raised this in an internal discussion, and I continue to be of the opinion that `unusedNodesInNewZones` and `nodesAndUnusedZones` are the same set. Combining them simplifies scheduling logic somewhat.

scheduler/replica_scheduler.go
```diff
@@ -261,24 +260,8 @@ func (rcs *ReplicaScheduler) getDiskCandidates(nodeInfo map[string]*longhorn.Nod
...
-		if len(diskCandidates) > 0 {
-			return diskCandidates, nil
-		}
-		multiError.Append(errors)
-		diskCandidates, errors = getDiskCandidatesFromNodes(nodesInUnusedZones)
-		if len(diskCandidates) > 0 {
-			return diskCandidates, nil
-		}
-		multiError.Append(errors)
-		diskCandidates, errors = getDiskCandidatesFromNodes(nodesWithEvictingReplicas)
```
ejweber, line 265: We can change this to `getDiskCandidatesFromNodes(filterNodesWithLessThanTwoReplicas(nodesWithEvictingReplicas))` as in the `!zoneSoftAntiAffinity && nodeSoftAntiAffinity` case. It is not legal to schedule a replica to a node with more than one (evicting) replica, because doing so would put two replicas in the same zone.

ejweber (commented):
Please look carefully at this commit. This is the first PR to add v15x to v16x upgrade logic. It is working for my needs, but I am curious if there is any reason I don't know to take this approach.

https://github.com/longhorn/longhorn-manager/pull/2094/commits/832131ca1bebb1d442e14c3d3589764da4e2bc45

scheduler/replica_scheduler.go
```diff
@@ -365,6 +360,41 @@ func (rcs *ReplicaScheduler) filterNodeDisksForReplica(node *longhorn.Node, disk
 	return preferredDisks, multiError
 }
 
+// filterDisksWithMatchingReplicas filters the input disks map and returns only the disks that have the fewest matching
+// replicas. If allowDuplicates is false, it only returns disks that have no matching replicas.
```
james-munson, line 364: Did the parameter name change from allowDuplicates to allowMatches?

scheduler/replica_scheduler.go
```diff
@@ -214,18 +225,18 @@ func (rcs *ReplicaScheduler) getDiskCandidates(nodeInfo map[string]*longhorn.Nod
...
-				unusedNodesInNewZones[nodeName] = node
-			}
 		}
 		if _, ok := usedZones[node.Status.Zone]; !ok {
-			nodesInUnusedZones[nodeName] = node
+			unusedNodesInUnusedZones[nodeName] = node
 		}
 	}
 
 	switch {
```
james-munson, line 234: NIT:  I still claim that since getDiskCandidatesFromNodes(unusedNodesInUnusedZones) is the first try in every case, it could be factored out and done before the switch, which might not even need to be executed.  Seems simpler to me, but I concede it's subjective.

james-munson (commented):
A couple of small things, and I'm not yet qualified to review upgrade code, but otherwise LGTM.

scheduler/replica_scheduler.go
ejweber: Fixing.

scheduler/replica_scheduler.go
ejweber: I forgot you said that, but I see what you mean. I will do as you suggest and see how it looks to other reviewers.

scheduler/replica_scheduler.go
```diff
@@ -224,33 +223,31 @@ func (rcs *ReplicaScheduler) getDiskCandidates(nodeInfo map[string]*longhorn.Nod
...
+
 	switch {
 	case !zoneSoftAntiAffinity && !nodeSoftAntiAffinity:
-		diskCandidates, errors := getDiskCandidatesFromNodes(unusedNodesInNewZones)
-		if len(diskCandidates) > 0 {
-			return diskCandidates, nil
-		}
-		multiError.Append(errors)
+		fallthrough
+	// Same as the above. If we cannot schedule two replicas in the same zone, we cannot schedule them on the same node.
```
shuo-wu, line 245: If we cannot find a candidate from `unusedNodesInUnusedZones` above, should not this be `break`?

scheduler/replica_scheduler.go
```diff
@@ -224,33 +223,31 @@ func (rcs *ReplicaScheduler) getDiskCandidates(nodeInfo map[string]*longhorn.Nod
...
+	case !zoneSoftAntiAffinity && nodeSoftAntiAffinity:
```
shuo-wu, line 246: Sorry... I don't get how you guarantee `!zoneSoftAntiAffinity` with the below code...

ejweber: Interesting. This is not a behavioral change. We previously used the same code in both blocks that had `!zoneSoftAntiAffinity`. I am guessing the thinking was that we are just replacing an evicting replica and the replica already was scheduled to this node, so rescheduling it to this node is fine.

The only way I can think of this being a problem is if the zone a node is in has changed since we originally scheduled the evicting replica. I think this is possible (maybe the user redefined zones or originally did not define zones), so we probably need to make changes.

This might have made sense to handle in a separate ticket, but since we are already refactoring the scheduler in this one, I'll see what I can do.

- [ ] Ensure we only replace an evicting replica if doing so would still respect `!zoneSoftAntiAffinity`.

scheduler/replica_scheduler.go
```diff
@@ -351,6 +353,41 @@ func (rcs *ReplicaScheduler) filterNodeDisksForReplica(node *longhorn.Node, disk
 	return preferredDisks, multiError
 }
 
+// filterDisksWithMatchingReplicas filters the input disks map and returns only the disks that have the fewest matching
+// replicas. If allowMatches is false, it only returns disks that have no matching replicas.
+func filterDisksWithMatchingReplicas(disks map[string]*Disk, replicas map[string]*longhorn.Replica,
+	allowMatches bool) map[string]*Disk {
```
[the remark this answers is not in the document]
ejweber: This just felt natural to me. I'm fine changing it.

- [ ] Change `allowMatches` to `diskSoftAntiAffinity`.

upgrade/v15xto160/upgrade.go
```diff
@@ -0,0 +1,179 @@
...
+	volumeMap, err := upgradeutil.ListAndUpdateVolumesInProvidedCache(namespace, lhClient, resourceMaps)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil
+		}
+		return errors.Wrapf(err, "failed to list all existing Longhorn volumes during the volume upgrade")
+	}
+
+	for _, v := range volumeMap {
+		_ = v // Nothing to do.
```
james-munson, line 79: Working by analogy with the 1.4->1.5 upgrade, perhaps we should check v.Spec.ReplicaDiskSoftAntiAffinity against "" and replace it with longhorn.ReplicaDiskSoftAntiAffinityDefault.

ejweber: Yes! That was literally the entire point of that commit originally! Not sure what happened to it, but I think I decided to rework more of the file than I originally intended and somehow lost the important change.

- [ ] Fix upgrade issue.

scheduler/replica_scheduler.go
shuo-wu: > I am guessing the thinking was that we are just replacing an evicting replica and the replica already was scheduled to this node, so rescheduling it to this node is fine.

OK. This is why the function has `getDiskCandidatesFromNodes(filterNodesWithLessThanTwoReplicas(nodesWithEvictingReplicas))`. 
But I don't think it always respects the affinity rules as the affinities may change before the eviction happens. In other words, a node containing an existing/evicting replica may no longer match the rules. Probably we need to re-check rather than blindly use the nodes containing one evicting replica.

Besides, with the modification you make for case `!zoneSoftAntiAffinity && nodeSoftAntiAffinity`, there is no way to schedule a replica to a **used** node in unused zones.

ejweber: > Besides, with the modification you make for case !zoneSoftAntiAffinity && nodeSoftAntiAffinity, there is no way to schedule a replica to a used node in unused zones.

This was the discussion that lead to the simplification in the first place (https://github.com/longhorn/longhorn-manager/pull/2094#pullrequestreview-1555771391 and internal). What is a "used node in an unused zone"? If a node is used, then the zone it is in is also used. Our criteria for determining whether a zone is used is whether it contains a used node.

I tried a couple of times to diagram this, but struggled to produce something I liked. Here's one attempt at it:

image

Is there is a counterexample to this I'm missing?

ejweber: > But I don't think it always respects the affinity rules as the affinities may change before the eviction happens. In other words, a node containing an existing/evicting replica may no longer match the rules. Probably we need to re-check rather than blindly use the nodes containing one evicting replica.

This is also something interesting to think about, though I would argue for doing it in a separate ticket if needed. While this PR introduces minor changes to scheduling behavior, it is mostly just a logic refactor. We already are scheduling replicas to nodes that are evicting them, and reconsidering this makes it difficult to focus on the purpose of this PR (disk anti-affinity).

To kick off that discussion though, can you clarify this with an example? When you say affinity, or you referring to node, zone, and disk anti-affinities? These are already covered. For example in the !zoneSoftAntiAffinity && !nodeSoftAntiAffinity case, we know we cannot schedule to a used node or used zone. We try the obvious (schedule to an unused node in an unused zone) first, then we schedule to a node which:

- Is in a zone where all replicas are evicting (this means soon the zone will be unused).
- Has only one evicting replica (this means soon the node will have no replica).

I am not sure what else we need to check regarding affinities.

If you are referring to other scheduling constraints (e.g., is a disk actually schedulable, are the tags appropriate, etc.), we handle that after this logic in `filterNodeDisksForReplica`.

shuo-wu: > If a node is used, then the zone it is in is also used. Our criteria for determining whether a zone is used is whether it contains a used node.

Got it. Great explanation!

> To kick off that discussion though, can you clarify this with an example?

Sure. Here is an example: Let's say there are 3 nodes in the cluster. node1 and node2 belongs to zoneA while node3 belongs to zoneB
1. When `zoneSoftAntiAffinity && !nodeSoftAntiAffinity`. We create a 3-replica volume. Each replica is on a separate replica
2. Later on we update the affinity rule to `!zoneSoftAntiAffinity && !nodeSoftAntiAffinity`
3. Somehow we evict the replica on node1.
4. Ideally, the volume cannot do scheduling for the new replica. But here `nodesWithEvictingReplicas()` returns node1 and `filterNodesWithLessThanTwoReplicas(filterNodesInOnlyEvictingZones())` cannot filter it out. As a result, the new replica will be scheduled to node1, which breaks the new rule. Is this expected?

james-munson (dismissed):
From my perspective, this is ready to go.

scheduler/replica_scheduler.go
ejweber: I went ahead and implemented a new test similar to your example. The final test as described below only uses one zone, but I also ran it with two zones as described.

1. Create three nodes in the same zone. Each node has a replica, but the replica on node1 is evicting.
2. Try to schedule another replica.
3. Verify that replica scheduling fails.

The test passes.

In your exact example, we fail to schedule a replica because `filterNodesInOnlyEvictingZones` actually does filter out `node1`. `zoneA` is NOT an only evicting zone, so we are not allowed to schedule to it.

image

ejweber: While looking into the example and writing the new test case, I further simplified/clarified the scheduling for the evicting replica case. I do not see any behavioral changes, but I think it makes the code a bit easier to read.

controller/setting_controller.go
```diff
@@ -1501,6 +1504,9 @@ func (info *ClusterInfo) collectVolumesInfo() error {
 		replicaZoneSoftAntiAffinity := info.collectSettingInVolume(string(volume.Spec.ReplicaZoneSoftAntiAffinity), string(longhorn.ReplicaZoneSoftAntiAffinityDefault), types.SettingNameReplicaZoneSoftAntiAffinity)
 		replicaZoneSoftAntiAffinityCountStruct[util.StructName(fmt.Sprintf(ClusterInfoVolumeReplicaZoneSoftAntiAffinityCountFmt, util.ConvertToCamel(string(replicaZoneSoftAntiAffinity), "-")))]++
 
+		replicaDiskSoftAntiAffinity := info.collectSettingInVolume(string(volume.Spec.ReplicaDiskSoftAntiAffinity), string(longhorn.ReplicaDiskSoftAntiAffinityDefault), types.SettingNameReplicaDiskSoftAntiAffinity)
+		replicaDiskSoftAntiAffinityCountStruct[util.StructName(fmt.Sprintf(ClusterInfoVolumeReplicaDiskSoftAntiAffinityCountFmt, util.ConvertToCamel(string(replicaDiskSoftAntiAffinity), "-")))]++
+
```
PhanLe1010, line 1509: Looks like we have a bug here @c3y1huang 

This setting `replicaDiskSoftAntiAffinity` only has 3 possible value: `ignore`, `enable`, `disable`. However, the upgrade responder might received 5 possible value returns by the function `collectSettingInVolume` which are: `ignore`, `enable`, `disable`, `true`, and `false`

upgrade/upgrade.go
```diff
@@ -219,23 +219,23 @@ func doResourceUpgrade(namespace string, lhClient *lhclientset.Clientset, kubeCl
...
 	}
 
-	// When lhVersionBeforeUpgrade < v1.5.0, it is v1.4.x. The `CheckUpgradePathSupported` method would have failed us out earlier if it was not v1.4.x.
+	// When lhVersionBeforeUpgrade < v1.6.0, it is v1.5.x. The `CheckUpgradePathSupported` method would have failed us out earlier if it was not v1.5.x.
 	resourceMaps = map[string]interface{}{}
 	if semver.Compare(lhVersionBeforeUpgrade, "v1.5.0") < 0 {
-		logrus.Info("Walking through the resource status upgrade path v1.4.x to v1.5.0")
-		if err := v14xto150.UpgradeResourcesStatus(namespace, lhClient, kubeClient, resourceMaps); err != nil {
+		logrus.Info("Walking through the resource status upgrade path v1.5.x to v1.6.0")
+		if err := v15xto160.UpgradeResourcesStatus(namespace, lhClient, kubeClient, resourceMaps); err != nil {
```
PhanLe1010, line 238: Should it be `if semver.Compare(lhVersionBeforeUpgrade, "v1.6.0") < 0` instead?

upgrade/v15xto160/upgrade.go
```diff
@@ -0,0 +1,179 @@
...
+)
+
+const (
+	upgradeLogPrefix = "upgrade from v1.5.x to v1.6.0: "
+)
+
+func UpgradeResources(namespace string, lhClient *lhclientset.Clientset, kubeClient *clientset.Clientset, resourceMaps map[string]interface{}) error {
+	if err := upgradeCSIPlugin(namespace, kubeClient); err != nil {
+		return err
+	}
```
PhanLe1010, line 25: Do we need this one? I think we probably don't need this one. It was needed when we switch to the new AD mechanism the first time

upgrade/v15xto160/upgrade.go
```diff
@@ -0,0 +1,179 @@
...
+
+	if err := upgradeEngines(namespace, lhClient, resourceMaps); err != nil {
+		return err
+	}
+
+	if err := upgradeReplicas(namespace, lhClient, resourceMaps); err != nil {
+		return err
+	}
+
+	return upgradeOrphans(namespace, lhClient, resourceMaps)
```
PhanLe1010, line 43: Should we remove these functions since we don't upgrade any thing inside these functions?

upgrade/v15xto160/upgrade.go
```diff
@@ -0,0 +1,179 @@
...
+		}
+		return errors.Wrapf(err, "failed to list all existing Longhorn nodes during the node upgrade")
+	}
+
+	for _, n := range nodeMap {
+		_ = n // Nothing to do.
+	}
+
+	return nil
+}
```
PhanLe1010, line 174: Should we remove this since we don't upgrade any thing inside this function?

client/README.md
```diff
@@ -1,27 +1,58 @@
...
-But in order to fix some compatibility issues, you need to use this [repo](https://github.com/niusmallnan/go-rancher), and you should checkout the `longhorn-v1` branch.
+The code in this package is generated by [rancher/go-rancher](https://github.com/rancher/go-rancher). But in order to
+fix some compatibility issues, you must use the `longhorn-v1` branch of this
+[repo](https://github.com/niusmallnan/go-rancher) instead.
 
 ## Generate code in go-rancher
 
-If you have a longhorn api URL, you can use this script:
+The version of go-rancher we use does not have Go module support. It must be cloned onto the correct location on the
+GOPATH and the Go tools must be executed with `GO111MODULE=off`. This might be unusual for developers used to working
```
PhanLe1010, line 10: Looks like I am able to run it without `GO111MODULE=off`. The path need to be under GOPATH indeed

upgrade/v15xto160/upgrade.go
james-munson: I assumed these were all placeholders to add steps to as various items crop up in 1.6.0 development.  It seems harmless to leave them as NOPs.

upgrade/upgrade.go
james-munson: Good catch!

controller/setting_controller.go
james-munson: This pattern shows up in other storage class parameters.  The global setting is either true or false; the storage class flavor is one of "ignore" (use the global setting), "enable" (force to true), or "disable" (force to false).
I have been supposing that the override-or-not check is made when a storage class is evaluated for use in a volume creation, and any such parameter would be resolved to either true or false in the volume itself, and those would be the only legal values for a volume.

james-munson: Phan explained to me that the parameter like "ignore" is not evaluated at volume creation.  It is retained in the volume and evaluated when needed, which is to say any time a new replica needs to be placed.  At that point, the controller will use whatever the current global setting is.  So it is possible for a volume that doesn't express a preference to change its behavior if the global setting is changed between attempts to place another replica.

client/README.md
ejweber: Sorry @PhanLe1010, can you clarify whether you are agreeing with this addition, or if changes need to me made? Thanks.

webhook/resources/volume/mutator.go
```diff
@@ -195,6 +195,9 @@ func (v *volumeMutator) Create(request *admission.Request, newObj runtime.Object
 		if volume.Spec.ReplicaZoneSoftAntiAffinity != longhorn.ReplicaZoneSoftAntiAffinityDisabled {
 			patchOps = append(patchOps, fmt.Sprintf(`{"op": "replace", "path": "/spec/replicaZoneSoftAntiAffinity", "value": "%s"}`, longhorn.ReplicaZoneSoftAntiAffinityDefault))
 		}
+		if volume.Spec.ReplicaDiskSoftAntiAffinity != longhorn.ReplicaDiskSoftAntiAffinityDisabled {
+			patchOps = append(patchOps, fmt.Sprintf(`{"op": "replace", "path": "/spec/replicaDiskSoftAntiAffinity", "value": "%s"}`, longhorn.ReplicaDiskSoftAntiAffinityDefault))
+		}
```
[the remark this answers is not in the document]
ejweber: Per line 171, for now, we allow `enabled` for many SPDK volume settings. To be honest, I'm not too sure why, but ReplicaDiskSoftAntiAffinity seems similar enough to the other affinity settings to mimc this behavior.

controller/setting_controller.go
```diff
@@ -1358,6 +1359,7 @@ func (info *ClusterInfo) collectSettings() error {
 		types.SettingNameReplicaReplenishmentWaitInterval:                         true,
 		types.SettingNameReplicaSoftAntiAffinity:                                  true,
 		types.SettingNameReplicaZoneSoftAntiAffinity:                              true,
+		types.SettingNameReplicaDiskSoftAntiAffinity:                              true,
```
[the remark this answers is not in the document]
ejweber: - [ ] Create and test upgrade responder PR.

webhook/resources/volume/mutator.go
```diff
@@ -277,6 +280,9 @@ func mutate(newObj runtime.Object, moreLabels map[string]string) (admission.Patc
 	if string(volume.Spec.ReplicaZoneSoftAntiAffinity) == "" {
 		patchOps = append(patchOps, fmt.Sprintf(`{"op": "replace", "path": "/spec/replicaZoneSoftAntiAffinity", "value": "%s"}`, longhorn.ReplicaZoneSoftAntiAffinityDefault))
 	}
+	if string(volume.Spec.ReplicaDiskSoftAntiAffinity) == "" {
+		patchOps = append(patchOps, fmt.Sprintf(`{"op": "replace", "path": "/spec/replicaDiskSoftAntiAffinity", "value": "%s"}`, longhorn.ReplicaDiskSoftAntiAffinityDefault))
```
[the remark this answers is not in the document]
ejweber: This applies to all volumes and ensures we an empty `ReplicaDiskSoftAntiAffinity` setting is always replaced with `ignored`. v2 

The one in `Create` only applies to v2 volumes and ensures they cannot be created with affinity settings enabled. As above, I am unsure why we should not attempt to apply affinity settings to v2 volumes, but it seems to me that we should handle all v2 affinity settings separately.

upgrade/v15xto160/upgrade.go
ejweber: Discussed with @PhanLe1010 and @james-munson offline and chose to remove all "placeholders" but leave a couple of comments suggesting where developers looking to add more upgrade behavior can go for examples.

upgrade/v15xto160/upgrade.go
ejweber: Discussed with @PhanLe1010 and @james-munson offline and chose to remove all "placeholders" but leave a couple of comments suggesting where developers looking to add more upgrade behavior can go for examples.

client/README.md
ejweber: Ah, I misread your comment. I don't remember that working for me. I will check again. What version is your Go binary?

- [ ] Verify whether or not we need GO111MODULE=off.

controller/setting_controller.go
ejweber: I did not understand this comment until I started trying to test my changes with the upgrade responder.

The standard configuration for the upgrade responder is to store true or false counts.

```
> show field keys
...
longhorn_volume_replica_auto_balance_disabled_count                 float
longhorn_volume_replica_disk_soft_anti_affinity_true_count          float
longhorn_volume_replica_soft_anti_affinity_false_count              float
```

When I was experimenting, I got the following rows:

```
> select app_version, value, longhorn_setting_replica_disk_soft_anti_affinity, longhorn_volume_replica_disk_soft_anti_affinity_true_count from upgrade_request
name: upgrade_request
time                app_version value longhorn_setting_replica_disk_soft_anti_affinity longhorn_volume_replica_disk_soft_anti_affinity_true_count
----                ----------- ----- ------------------------------------------------ ----------------------------------------------------------
1692199982764244311 v1.6.0-dev  1                                                      
1692199984374879541 v1.6.0-dev  1     true                                             
1692199984932156839 v1.6.0-dev  1                                                      
1692201702247618751 v0.8.1      1                                                      
1692207182563500432 v1.6.0-dev  1                                                      
1692207184123366093 v1.6.0-dev  1     true                                             2
1692207184676124001 v1.6.0-dev  1                                                      
1692210782632842772 v1.6.0-dev  1                                                      
1692210784275448343 v1.6.0-dev  1     false                                            
1692210784788222835 v1.6.0-dev  1                                                      
1692211523658344337 v1.6.0-dev  1                                                      
1692211528320057508 v1.6.0-dev  1                                                      
1692211534450305082 v1.6.0-dev  1     false        
```

I got the row with the 2 in it when the `ReplicaDiskSoftAntiAffinity` setting was `true` and both volumes had `ignored`.

The [collectSettingInVolume function](https://github.com/longhorn/longhorn-manager/blob/c144fc16d8b30f1eda63005bae83e6604586554a/controller/setting_controller.go#L1560-L1569) counted both volumes as `true` (which comes from the `ReplicaDiskSoftAntiAffinity` setting itself.

However, when I set the `ReplicaDiskSoftAntiAffinity` setting to `false` and one volume to `enabled` before the last line, presumably one volume was counted as `enabled` and one as `false`. Since neither are `true` they are not represented in the database.

@PhanLe1010, I think there are only four possible values: `true`, `false`, `enabled`, and `disabled` right now. If a volume has ignored, it will register as `true` or `false`.

Not sure exactly what we want to see here. IMO, we actually want to reduce the data to a boolean, whether we represent it as `enabled` vs. `disabled` or `true` vs. `false`. We want to know whether or not the volume is actually using `ReplicaDiskAntiAffinity` at the moment, not necessarily the exact value of this field setting.

controller/setting_controller.go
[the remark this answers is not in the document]
ejweber: You are right. I made the change in my open longhorn/longhorn PR: https://github.com/longhorn/longhorn/pull/6432/commits/7458355578d725cdae12b0d695dc1a8d3202fa72.

See https://github.com/longhorn/longhorn-manager/pull/2094#discussion_r1296320755 for some basic testing. As far as I can tell, the new dashboard isn't available to test with, so I used the InfluxDB CLI instead.

client/README.md
ejweber: These are my results:

```
eweber@laptop:~/go/src/github.com/niusmallnan/go-rancher> go version
go version go1.20.2 linux/amd64

eweber@laptop:~/go/src/github.com/niusmallnan/go-rancher> go get
go: go.mod file not found in current directory or any parent directory.
        'go get' is no longer supported outside a module.
        To build and install a command, use 'go install' with a version,
        like 'go install example.com/cmd@latest'
        For more information, see https://golang.org/doc/go-get-install-deprecation
        or run 'go help get' or 'go help install'.

eweber@laptop:~/go/src/github.com/niusmallnan/go-rancher> ./scripts/generate-longhorn-schemas.sh http://164.92.95.242:30186
++ dirname ./scripts/generate-longhorn-schemas.sh
+ cd ./scripts/../generator
+ URL_BASE=http://localhost:8080
+ '[' 1 -gt 1 ']'
+ '[' http://164.92.95.242:30186 '!=' '' ']'
+ URL_BASE=http://164.92.95.242:30186
+ source ../scripts/common_functions
++ set -e
+ gen http://164.92.95.242:30186/v1 longhorn rename
+ BASE=http://164.92.95.242:30186/v1
+ curl -s 'http://164.92.95.242:30186/v1/schemas?_role=service'
+ jq .
+ echo Saved schemas.json
Saved schemas.json
+ echo -n Generating go code...
Generating go code...+ rm -rf '../output/generated_*'
+ go run generator.go
generator.go:13:2: no required module provides package github.com/rancher/go-rancher/client: go.mod file not found in current directory or any parent directory; see 'go help modules'
```

Can't get dependencies or generate code due to modules related errors.

With GO111MODULE=off:

```
eweber@laptop:~/go/src/github.com/niusmallnan/go-rancher> GO111MODULE=off go get

eweber@laptop:~/go/src/github.com/niusmallnan/go-rancher> GO111MODULE=off ./scripts/generate-longhorn-schemas.sh http://164.92.95.242:30186
++ dirname ./scripts/generate-longhorn-schemas.sh
+ cd ./scripts/../generator
+ URL_BASE=http://localhost:8080
+ '[' 1 -gt 1 ']'
+ '[' http://164.92.95.242:30186 '!=' '' ']'
+ URL_BASE=http://164.92.95.242:30186
+ source ../scripts/common_functions
++ set -e
+ gen http://164.92.95.242:30186/v1 longhorn rename
+ BASE=http://164.92.95.242:30186/v1
+ curl -s 'http://164.92.95.242:30186/v1/schemas?_role=service'
+ jq .
+ echo Saved schemas.json
Saved schemas.json
+ echo -n Generating go code...
Generating go code...+ rm -rf '../output/generated_*'
+ go run generator.go
+ echo ' Done'
 Done
+ gofmt -w ../output/generated_activate_input.go ../output/generated_api_version.go ../output/generated_attach_input.go ../output/generated_attachment.go ../output/generated_backing_image_cleanup_input.go ../output/generated_backing_image_disk_file_status.go ../output/generated_backing_image.go ../output/generated_backup.go ../output/generated_backup_input.go ../output/generated_backup_list_output.go ../output/generated_backup_status.go ../output/generated_backup_target.go ../output/generated_backup_volume.go ../output/generated_client.go ../output/generated_clone_status.go ../output/generated_controller.go ../output/generated_detach_input.go ../output/generated_disk_condition.go ../output/generated_disk_info.go ../output/generated_disk_update.go ../output/generated_disk_update_input.go ../output/generated_empty.go ../output/generated_engine_image.go ../output/generated_engine_upgrade_input.go ../output/generated_error.go ../output/generated_event.go ../output/generated_expand_input.go ../output/generated_instance_manager.go ../output/generated_instance_process.go ../output/generated_kubernetes_status.go ../output/generated_longhorn_condition.go ../output/generated_node_condition.go ../output/generated_node.go ../output/generated_orphan.go ../output/generated_purge_status.go ../output/generated_pvccreate_input.go ../output/generated_pvcreate_input.go ../output/generated_rebuild_status.go ../output/generated_recurring_job.go ../output/generated_replica.go ../output/generated_replica_remove_input.go ../output/generated_restore_status.go ../output/generated_salvage_input.go ../output/generated_setting_definition.go ../output/generated_setting.go ../output/generated_snapshot_cr.go ../output/generated_snapshot_crinput.go ../output/generated_snapshot_crlist_output.go ../output/generated_snapshot.go ../output/generated_snapshot_input.go ../output/generated_snapshot_list_output.go ../output/generated_support_bundle.go ../output/generated_support_bundle_initate_input.go ../output/generated_system_backup.go ../output/generated_system_restore.go ../output/generated_tag.go ../output/generated_update_access_mode_input.go ../output/generated_update_backup_compression_input.go ../output/generated_update_data_locality_input.go ../output/generated_update_offline_replica_rebuilding_input.go ../output/generated_update_replica_auto_balance_input.go ../output/generated_update_replica_count_input.go ../output/generated_update_replica_disk_soft_anti_affinity_input.go ../output/generated_update_replica_soft_anti_affinity_input.go ../output/generated_update_replica_zone_soft_anti_affinity_input.go ../output/generated_update_snapshot_data_integrity_input.go ../output/generated_update_unmap_mark_snap_chain_removed_input.go ../output/generated_volume_attachment.go ../output/generated_volume_condition.go ../output/generated_volume.go ../output/generated_volume_recurring_job.go ../output/generated_volume_recurring_job_input.go ../output/generated_workload_status.go
+ echo Formatted code
Formatted code
+ '[' -n rename ']'
+ rm -rf ../longhorn/generated_activate_input.go ../longhorn/generated_api_version.go ../longhorn/generated_attach_input.go ../longhorn/generated_attachment.go ../longhorn/generated_backing_image_cleanup_input.go ../longhorn/generated_backing_image_disk_file_status.go ../longhorn/generated_backing_image.go ../longhorn/generated_backup.go ../longhorn/generated_backup_input.go ../longhorn/generated_backup_list_output.go ../longhorn/generated_backup_status.go ../longhorn/generated_backup_target.go ../longhorn/generated_backup_volume.go ../longhorn/generated_client.go ../longhorn/generated_clone_status.go ../longhorn/generated_controller.go ../longhorn/generated_detach_input.go ../longhorn/generated_disk_condition.go ../longhorn/generated_disk_info.go ../longhorn/generated_disk_update.go ../longhorn/generated_disk_update_input.go ../longhorn/generated_empty.go ../longhorn/generated_engine_image.go ../longhorn/generated_engine_upgrade_input.go ../longhorn/generated_error.go ../longhorn/generated_event.go ../longhorn/generated_expand_input.go ../longhorn/generated_instance_manager.go ../longhorn/generated_instance_process.go ../longhorn/generated_kubernetes_status.go ../longhorn/generated_longhorn_condition.go ../longhorn/generated_node_condition.go ../longhorn/generated_node.go ../longhorn/generated_orphan.go ../longhorn/generated_purge_status.go ../longhorn/generated_pvccreate_input.go ../longhorn/generated_pvcreate_input.go ../longhorn/generated_rebuild_status.go ../longhorn/generated_recurring_job.go ../longhorn/generated_replica.go ../longhorn/generated_replica_remove_input.go ../longhorn/generated_restore_status.go ../longhorn/generated_salvage_input.go ../longhorn/generated_setting_definition.go ../longhorn/generated_setting.go ../longhorn/generated_snapshot_cr.go ../longhorn/generated_snapshot_crinput.go ../longhorn/generated_snapshot_crlist_output.go ../longhorn/generated_snapshot.go ../longhorn/generated_snapshot_input.go ../longhorn/generated_snapshot_list_output.go ../longhorn/generated_support_bundle.go ../longhorn/generated_support_bundle_initate_input.go ../longhorn/generated_system_backup.go ../longhorn/generated_system_restore.go ../longhorn/generated_tag.go ../longhorn/generated_update_access_mode_input.go ../longhorn/generated_update_backup_compression_input.go ../longhorn/generated_update_data_locality_input.go ../longhorn/generated_update_offline_replica_rebuilding_input.go ../longhorn/generated_update_replica_auto_balance_input.go ../longhorn/generated_update_replica_count_input.go ../longhorn/generated_update_replica_disk_soft_anti_affinity_input.go ../longhorn/generated_update_replica_soft_anti_affinity_input.go ../longhorn/generated_update_replica_zone_soft_anti_affinity_input.go ../longhorn/generated_update_snapshot_data_integrity_input.go ../longhorn/generated_update_unmap_mark_snap_chain_removed_input.go ../longhorn/generated_volume_attachment.go ../longhorn/generated_volume_condition.go ../longhorn/generated_volume.go ../longhorn/generated_volume_recurring_job.go ../longhorn/generated_volume_recurring_job_input.go ../longhorn/generated_workload_status.go
+ mv ../output/generated_activate_input.go ../output/generated_api_version.go ../output/generated_attach_input.go ../output/generated_attachment.go ../output/generated_backing_image_cleanup_input.go ../output/generated_backing_image_disk_file_status.go ../output/generated_backing_image.go ../output/generated_backup.go ../output/generated_backup_input.go ../output/generated_backup_list_output.go ../output/generated_backup_status.go ../output/generated_backup_target.go ../output/generated_backup_volume.go ../output/generated_client.go ../output/generated_clone_status.go ../output/generated_controller.go ../output/generated_detach_input.go ../output/generated_disk_condition.go ../output/generated_disk_info.go ../output/generated_disk_update.go ../output/generated_disk_update_input.go ../output/generated_empty.go ../output/generated_engine_image.go ../output/generated_engine_upgrade_input.go ../output/generated_error.go ../output/generated_event.go ../output/generated_expand_input.go ../output/generated_instance_manager.go ../output/generated_instance_process.go ../output/generated_kubernetes_status.go ../output/generated_longhorn_condition.go ../output/generated_node_condition.go ../output/generated_node.go ../output/generated_orphan.go ../output/generated_purge_status.go ../output/generated_pvccreate_input.go ../output/generated_pvcreate_input.go ../output/generated_rebuild_status.go ../output/generated_recurring_job.go ../output/generated_replica.go ../output/generated_replica_remove_input.go ../output/generated_restore_status.go ../output/generated_salvage_input.go ../output/generated_setting_definition.go ../output/generated_setting.go ../output/generated_snapshot_cr.go ../output/generated_snapshot_crinput.go ../output/generated_snapshot_crlist_output.go ../output/generated_snapshot.go ../output/generated_snapshot_input.go ../output/generated_snapshot_list_output.go ../output/generated_support_bundle.go ../output/generated_support_bundle_initate_input.go ../output/generated_system_backup.go ../output/generated_system_restore.go ../output/generated_tag.go ../output/generated_update_access_mode_input.go ../output/generated_update_backup_compression_input.go ../output/generated_update_data_locality_input.go ../output/generated_update_offline_replica_rebuilding_input.go ../output/generated_update_replica_auto_balance_input.go ../output/generated_update_replica_count_input.go ../output/generated_update_replica_disk_soft_anti_affinity_input.go ../output/generated_update_replica_soft_anti_affinity_input.go ../output/generated_update_replica_zone_soft_anti_affinity_input.go ../output/generated_update_snapshot_data_integrity_input.go ../output/generated_update_unmap_mark_snap_chain_removed_input.go ../output/generated_volume_attachment.go ../output/generated_volume_condition.go ../output/generated_volume.go ../output/generated_volume_recurring_job.go ../output/generated_volume_recurring_job_input.go ../output/generated_workload_status.go ../longhorn
+ sed -i -e 's/package client/package longhorn/g' ../longhorn/client.go ../longhorn/common.go ../longhorn/generated_activate_input.go ../longhorn/generated_api_version.go ../longhorn/generated_attach_input.go ../longhorn/generated_attachment.go ../longhorn/generated_backing_image_cleanup_input.go ../longhorn/generated_backing_image_disk_file_status.go ../longhorn/generated_backing_image.go ../longhorn/generated_backup.go ../longhorn/generated_backup_input.go ../longhorn/generated_backup_list_output.go ../longhorn/generated_backup_status.go ../longhorn/generated_backup_target.go ../longhorn/generated_backup_volume.go ../longhorn/generated_client.go ../longhorn/generated_clone_status.go ../longhorn/generated_controller.go ../longhorn/generated_detach_input.go ../longhorn/generated_disk_condition.go ../longhorn/generated_disk_info.go ../longhorn/generated_disk_update.go ../longhorn/generated_disk_update_input.go ../longhorn/generated_empty.go ../longhorn/generated_engine_image.go ../longhorn/generated_engine_upgrade_input.go ../longhorn/generated_error.go ../longhorn/generated_event.go ../longhorn/generated_expand_input.go ../longhorn/generated_instance_manager.go ../longhorn/generated_instance_process.go ../longhorn/generated_kubernetes_status.go ../longhorn/generated_longhorn_condition.go ../longhorn/generated_node_condition.go ../longhorn/generated_node.go ../longhorn/generated_orphan.go ../longhorn/generated_purge_status.go ../longhorn/generated_pvccreate_input.go ../longhorn/generated_pvcreate_input.go ../longhorn/generated_rebuild_status.go ../longhorn/generated_recurring_job.go ../longhorn/generated_replica.go ../longhorn/generated_replica_remove_input.go ../longhorn/generated_restore_status.go ../longhorn/generated_salvage_input.go ../longhorn/generated_setting_definition.go ../longhorn/generated_setting.go ../longhorn/generated_snapshot_cr.go ../longhorn/generated_snapshot_crinput.go ../longhorn/generated_snapshot_crlist_output.go ../longhorn/generated_snapshot.go ../longhorn/generated_snapshot_input.go ../longhorn/generated_snapshot_list_output.go ../longhorn/generated_support_bundle.go ../longhorn/generated_support_bundle_initate_input.go ../longhorn/generated_system_backup.go ../longhorn/generated_system_restore.go ../longhorn/generated_tag.go ../longhorn/generated_update_access_mode_input.go ../longhorn/generated_update_backup_compression_input.go ../longhorn/generated_update_data_locality_input.go ../longhorn/generated_update_offline_replica_rebuilding_input.go ../longhorn/generated_update_replica_auto_balance_input.go ../longhorn/generated_update_replica_count_input.go ../longhorn/generated_update_replica_disk_soft_anti_affinity_input.go ../longhorn/generated_update_replica_soft_anti_affinity_input.go ../longhorn/generated_update_replica_zone_soft_anti_affinity_input.go ../longhorn/generated_update_snapshot_data_integrity_input.go ../longhorn/generated_update_unmap_mark_snap_chain_removed_input.go ../longhorn/generated_volume_attachment.go ../longhorn/generated_volume_condition.go ../longhorn/generated_volume.go ../longhorn/generated_volume_recurring_job.go ../longhorn/generated_volume_recurring_job_input.go ../longhorn/generated_workload_status.go ../longhorn/schemas.go ../longhorn/types.go
+ rm -rf ../output
+ rm schemas.json
+ rm -rf '../longhorn/*.go-e'
+ echo Success
Success
```

Is it possible you are running an older version of Go than me (`v1.20.2`) or have `GO111MODULE=off` permanently set in your Go environment?

ejweber:
@phan and @shuo-wu, I believe I have addressed all comments (either with code or with a comment of my own).

controller/setting_controller.go
PhanLe1010: > Not sure exactly what we want to see here. IMO, we actually want to reduce the data to a boolean, whether we represent it as enabled vs. disabled or true vs. false. We want to know whether or not the volume is actually using ReplicaDiskAntiAffinity at the moment, not necessarily the exact value of this field setting.

Agree with this one. Could we create a ticket for this since it is happening with other settings also?

webhook/resources/volume/mutator.go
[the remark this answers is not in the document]
PhanLe1010: Ok. Let's keep this disabled for now. We can also enable it later if needed

controller/setting_controller.go
ejweber: https://github.com/longhorn/longhorn/issues/6593

client/README.md
PhanLe1010: It is possible that I was running old version of Go. Thank you

scheduler/replica_scheduler.go
PhanLe1010: > I'm assuming this is because the replica will soon no longer exist on the node, and the newly scheduled replica is a replacement. I am not aware of any reason to change this behavior. 

This is related to this ticket https://github.com/longhorn/longhorn/issues/3076. The expectation was there might be a different disk, disk-b, on the node-1 when user is trying to evict replicas from disk-a on node-1. It is not because that the replica will soon no longer exist on the node. Understand this usecase, maybe we want to rename some of the concepts `unusedNodesAfterEviction` and  `unusedNodesInUnusedZonesAfterEviction`

ejweber:
@PhanLe1010 brought up the following concern out-of-band:

> I am having a bit difficulty understand some replica scheduling refactoring logic.
First one is the concept of onlyEvictingNodes and  onlyEvictingZones The way they are calculated is https://github.dev/longhorn/longhorn-manager/blob/472498aa9d73a0e2022b2af1a86d9911d7e15048/scheduler/replica_scheduler.go#L209-L214.
Assume that there is a 2 replica on the same zone-a. One of the 2 replicas has r.Status.EvictionRequested and the other one doesn't. When we loop over the list of the replica. Depending on the looping order, onlyEvictingZones might have different value?

This was a correct criticism. I did not notice before, but the new test case "fail scheduling when doing so would reuse an invalid evicting node" was flaking (rarely) due to this issue. In addition, I moved the flawed logic to its own function (`getCurrentNodesAndZones`) and added a new unit test (`testGetCurrentNodesAndZones`) to verify the issue before fixing it in https://github.com/longhorn/longhorn-manager/pull/2094/commits/d7dba1327293cef48e4a32959c0fa1b6f0bd0a15.

Before the fix, both "fail scheduling when doing so would reuse an invalid evicting node" and `testGetCurrentNodesAndZones` were sure to fail when run enough times. After the fix, there are no order related flakes:

```
eweber@laptop:~/longhorn-manager> go test ./scheduler/... -count=1000
ok      github.com/longhorn/longhorn-manager/scheduler  3.591s
```

Thanks @PhanLe1010!

scheduler/replica_scheduler.go
ejweber: I see. Thanks for the clarification @PhanLe1010! However, I'm not sure there's a better name for these concepts, despite my initial confusion. I thought about something like `unusedNodesConsideringEviction`, etc., but that just seems wordier.

ejweber: Verified https://github.com/longhorn/longhorn/issues/3076#issuecomment-961507326 after changes per @PhanLe1010's request.

PhanLe1010:
cc @c3y1huang Could you help to co-review this PR? You have a lot of knowledge on the replica scheduler part so it would be great if you could double-check. Thank you

scheduler/replica_scheduler.go
```diff
@@ -168,43 +168,33 @@ func (rcs *ReplicaScheduler) getDiskCandidates(nodeInfo map[string]*longhorn.Nod
...
+		diskCandidates = map[string]*Disk{}
 		multiError = util.NewMultiError()
 		for _, node := range nodes {
-			diskCandidates, errors := rcs.filterNodeDisksForReplica(node, nodeDisksMap[node.Name], replicas, volume, requireSchedulingCheck)
-			if len(diskCandidates) > 0 {
-				return diskCandidates, nil
+			diskCandidatesFromNode, errors := rcs.filterNodeDisksForReplica(node, nodeDisksMap[node.Name], replicas,
+				volume, requireSchedulingCheck)
+			for k, v := range diskCandidatesFromNode {
+				diskCandidates[k] = v
```
c3y1huang, line 189: Could two separate nodes potentially have the same disk UUIDs?

c3y1huang (commented):
Just one question. The rest lgtm.

shuo-wu (approved):
LGTM

PhanLe1010 (approved):
LGTM

c3y1huang (approved):

PR #2102, opened 2023-08-08
Update unknown instance manager handling logic

shuo-wu:
https://github.com/longhorn/longhorn/issues/6479


shuo-wu:
For the 2nd commit, what I am worried about is, if it's risky to re-attach the volumes on the problematic nodes after the IM pod eviction is triggered. Since the pod eviction may not succeed, the engines and workloads on the node can be still running.

controller/instance_manager_controller.go
```diff
@@ -412,15 +405,30 @@ func (imc *InstanceManagerController) syncStatusWithPod(im *longhorn.InstanceMan
 		} else {
 			im.Status.CurrentState = longhorn.InstanceManagerStateStarting
 		}
```
innobead, line 408: Why we remove the below code? If the pod is unknown, and it also means it could be still running somehow. 

```go
	case corev1.PodUnknown:
		im.Status.CurrentState = longhorn.InstanceManagerStateUnknown
```

derekbit: Same question. Other parts LGTM.

mantissahz: logic moved into `syncStatusWithNode(im)`?

shuo-wu: It was already deprecated a long time ago:
```
// Deprecated: It isn't being set since 2015 (74da3b14b0c0f658b3bb8d2def5094686d0e9095)
```

innobead: Oh then I am ok with this.

innobead (approved):

innobead:

PR #2103, opened 2023-08-09
feat: talos support

c3y1huang:
https://github.com/longhorn/longhorn/issues/3161


go.mod
```diff
@@ -2,6 +2,14 @@ module github.com/longhorn/longhorn-manager
 
 go 1.20
 
+replace github.com/longhorn/go-common-libs => github.com/c3y1huang/go-common-libs v0.0.0-20230818102433-680b5db81371
```
c3y1huang, line 5: TODO: update after https://github.com/longhorn/go-common-libs/pull/1

go.mod
```diff
@@ -2,6 +2,14 @@ module github.com/longhorn/longhorn-manager
...
+
+replace github.com/longhorn/go-iscsi-helper => github.com/c3y1huang/go-iscsi-helper v0.0.0-20230818102652-d94028f9f0e8
```
c3y1huang, line 7: TODO: update after https://github.com/longhorn/go-iscsi-helper/pull/69

go.mod
```diff
@@ -2,6 +2,14 @@ module github.com/longhorn/longhorn-manager
...
+
+replace github.com/longhorn/longhorn-engine => github.com/c3y1huang/longhorn-engine v1.4.0-rc1.0.20230818103126-3ba57dbd5d39
```
c3y1huang, line 9: TODO: update after https://github.com/longhorn/longhorn-engine/pull/919

go.mod
```diff
@@ -2,6 +2,14 @@ module github.com/longhorn/longhorn-manager
...
+
+replace github.com/longhorn/longhorn-share-manager => github.com/c3y1huang/longhorn-share-manager v0.0.0-20230821032331-a5abe0d98fd7
```
c3y1huang, line 11: TODO: update after https://github.com/longhorn/longhorn-share-manager/pull/80

shuo-wu (approved):
In general LGTM

csi/crypto/crypto.go
```diff
@@ -72,8 +75,20 @@ func VolumeMapper(volume string) string {
 
 // EncryptVolume encrypts provided device with LUKS.
 func EncryptVolume(devicePath, passphrase string, cryptoParams *EncryptParams) error {
+	namespaces := []lhtypes.Namespace{lhtypes.NamespaceMnt, lhtypes.NamespaceIpc}
```
innobead, line 78: I would suggest creating a few namespace groups for our default usage in common-libs and exporting them, so we need to redeclare these namespaces again. Like which cases require mnt and ipc or even net. 

Create a ticket for that to improve. This will make our code better maintainable.

csi/crypto/crypto.go
```diff
@@ -86,8 +101,14 @@ func OpenVolume(volume, devicePath, passphrase string) error {
 		return nil
 	}
 
+	namespaces := []lhtypes.Namespace{lhtypes.NamespaceMnt, lhtypes.NamespaceIpc}
+	nsexec, err := lhns.NewNamespaceExecutor(lhtypes.ProcessNone, lhtypes.ProcDirectory, namespaces)
+	if err != nil {
+		return err
+	}
+
```
innobead, line 109: We can transform the following code into `LuksOpen`, reducing the need for the caller to go through tedious steps to build the executor. We can open another ticket to accomplish this.

Having a general-purpose library is beneficial, but for specific purposes and one-step calls, simplifying it would be great.
```go
	namespaces := []lhtypes.Namespace{lhtypes.NamespaceMnt, lhtypes.NamespaceIpc}
	nsexec, err := lhns.NewNamespaceExecutor(lhtypes.ProcessNone, lhtypes.ProcDirectory, namespaces)
	if err != nil {
		return err
	}

```

csi/crypto/crypto.go
```diff
@@ -96,8 +117,14 @@ func OpenVolume(volume, devicePath, passphrase string) error {
...
 	logrus.Infof("Closing LUKS device %s", volume)
-	_, err := luksClose(volume)
+	_, err = nsexec.LuksClose(volume, lhtypes.LuksTimeout)
 	return err
```
innobead, line 128: If the code like below, it will be perfect.

```go
	logrus.Infof("Closing LUKS device %s", volume)
	_, err := luksClose(volume)
	_, err = nsexec.LuksClose(volume, lhtypes.LuksTimeout)
	return err
```

util/util.go
```diff
@@ -878,28 +886,38 @@ func TrimFilesystem(volumeName string, encryptedDevice bool) error {
...
-		return fmt.Errorf("cannot find a valid mount point for volume %v", volumeName)
+
+	namespaces := []lhtypes.Namespace{lhtypes.NamespaceMnt, lhtypes.NamespaceNet}
+	nsexec, err := lhns.NewNamespaceExecutor(lhns.GetDefaultProcessName(), lhtypes.HostProcDirectory, namespaces)
+	if err != nil {
+		return err
 	}
 
-	_, err = nsExec.Execute("fstrim", []string{mountpoint})
+	_, err = nsexec.Execute(lhtypes.BinaryFstrim, []string{validMountpoint}, lhtypes.ExecuteDefaultTimeout)
```
innobead, line 737: Suggest having a unit test for this mount point parsing from the content of mounts

innobead (changes requested):
In general, LGTM. Just some comments need to resolve. Also, needs to update the deps.

csi/crypto/crypto.go
c3y1huang: Why do we want to call the operation twice?

https://github.com/longhorn/longhorn-manager/blob/v1.5.3/csi/crypto/luks.go#L24
```go
func luksClose(volume string) (stdout string, err error) {
	return cryptSetup("luksClose", volume)
}
```

https://github.com/longhorn/go-common-libs/blob/main/ns/crypto.go#L18
```go
func (nsexec *Executor) LuksClose(volume string, timeout time.Duration) (stdout string, err error) {
	args := []string{"luksClose", volume}
	return nsexec.Cryptsetup(args, timeout)
}
```

util/util.go
c3y1huang: We can have unit tests to cover the validation of the mount point instead of just the content. This will give us a bit more coverage.

csi/crypto/crypto.go
innobead: > Why do we want to call the operation twice?
Typos ..

csi/crypto/crypto.go
c3y1huang: https://github.com/longhorn/longhorn/issues/7406

csi/crypto/crypto.go
c3y1huang: https://github.com/longhorn/longhorn/issues/7408

util/util.go
innobead: Sounds good.

innobead (approved):

PR #2104, opened 2023-08-10
Refactor CreateSnapshot to return error when backup cannot complete

ejweber:
longhorn/longhorn#4979

https://github.com/kubernetes-csi/external-snapshotter/issues/880 seems to be the main driver for the observed behavior. When a VolumeSnapshot is deleted, the csi-snapshotter sidecar attempts a SnapshotCreate instead of a SnapshotDelete. It does not attempt a SnapshotDelete until its `--resync-period` (up to 15 minutes by default).

That being said, we aren't correctly adhering to the CSI spec. If we do so, we get slightly improved behavior as a side effect.

> A gRPC error code SHALL be returned if an error occurs during any stage of the snapshotting process. A CO SHOULD explicitly delete snapshots when an error occurs.

Before this PR, we return an error on the first CSI CreateSnapshot call (the backup never initializes because there is no backup target set). In subsequent CSI CreateSnapshot calls, we do not return an error, even though we know one has occurred. After this PR, we correctly return an error on every CSI CreateSnapshot call. In addition the error is actually meaningful.

Before, VolumeSnapshotContent had:

```
status:
  creationTime: 1691438945000000000
  error:
    message: 'Failed to create snapshot: failed to take snapshot of the volume pvc-9c1ed502-4fda-48c9-ac43-bc32c61f2cf2:
      "rpc error: code = DeadlineExceeded desc = waitForBackupInitiation: timeout
      while waiting for backup initiation for volume pvc-9c1ed502-4fda-48c9-ac43-bc32c61f2cf2
      for snapshot snapshot-bdf730c6-d130-44c5-867e-3aebef26582c"'
    time: "2023-08-07T20:10:07Z"
  readyToUse: false
  restoreSize: 5368709120
  snapshotHandle: bak://pvc-9c1ed502-4fda-48c9-ac43-bc32c61f2cf2/backup-92eeb8bde2244e2d
```

Now, it has the true cause of the problem:

```
 status:
    creationTime: 1691701725000000000
    error:
      message: 'Failed to check and update snapshot content: failed to take snapshot
        of the volume pvc-4301b9b8-ce2b-4dc0-9df5-550dbdbc9648: "rpc error: code =
        Internal desc = proxyServer=10.42.27.16:8501 destination=10.42.59.137:10010:
        failed to backup snapshot snapshot-e2fa0e40-7bf5-43af-a756-7f33e9c5b620 to
        backup-813a5068031b4204: rpc error: code = Unknown desc = failed to create
        backup to  for volume pvc-4301b9b8-ce2b-4dc0-9df5-550dbdbc9648: rpc error:
        code = Unknown desc = missing input parameter"'
      time: "2023-08-10T21:08:49Z"
    readyToUse: false
    restoreSize: 5368709120
    snapshotHandle: bak://pvc-4301b9b8-ce2b-4dc0-9df5-550dbdbc9648/backup-813a5068031b4204
```

Inside the csi-snapshotter, calls with errors are treated differently from calls without. This means, when a VolumeSnapshot is deleted, we get a DeleteSnapshot faster (though we are still subject to rate limiting, so in the worst case it can take `--retry-interval-max`, which defaults to five minutes.

NOTE: This change makes https://github.com/kubernetes-csi/external-snapshotter/issues/778 more noticeable in this scenario. I have tested the fix on external-provisioner's master branch and it works, so we should update to the next patch version when possible.


ejweber:
Made this a draft because I realized while typing up the description that some of these changes may not be necessary. Will strip them out and retest. Feel free to not review for now.

ejweber:
Multiple small fixes:

1. Return quickly and successfully on the initial CSI CreateSnapshot call because we look for a backup with spec.SnapshotName (which exists immediately) instead of status.SnapshotName (which, in the case of a failed backup, may never exist).
2. Return an error on subsequent CSI CreateSnapshot calls if a backup error has occured. This is the correct action to take and moderately improves deletion performance (though we need changes to external-snapshotter to further improve).

ejweber:

innobead (approved):
LGTM. Good analysis.

PR #2107, opened 2023-08-17
feat: engine version enforcement

c3y1huang:
https://github.com/longhorn/longhorn/issues/5842

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -49,7 +49,7 @@ require (
 	github.com/longhorn/go-common-libs v0.0.0-20231207171057-cd5fff99e604
 	github.com/longhorn/go-iscsi-helper v0.0.0-20231220091133-38e296e68ad2
 	github.com/longhorn/go-spdk-helper v0.0.0-20231222092128-dd4eb991eecb
-	github.com/longhorn/longhorn-engine v1.6.0-dev-20231217.0.20231221124647-1a5cdbf02a5a
+	github.com/longhorn/longhorn-engine v1.6.0-dev-20231217.0.20231226052201-1c1b8312a30d
 	github.com/longhorn/longhorn-instance-manager v1.6.0-dev-20231217.0.20231226062554-00268adaa271
 	github.com/longhorn/longhorn-share-manager v1.6.0-dev-20231217.0.20231221124415-c1ae736675af
 	github.com/pkg/errors v0.9.1
--- a/upgrade/util/util.go
+++ b/upgrade/util/util.go
@@ -22,6 +22,8 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	clientset "k8s.io/client-go/kubernetes"
 
+	emeta "github.com/longhorn/longhorn-engine/pkg/meta"
+
 	"github.com/longhorn/longhorn-manager/meta"
 	"github.com/longhorn/longhorn-manager/types"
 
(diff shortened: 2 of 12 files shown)


go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/longhorn/longhorn-manager
 
 go 1.20
 
+replace github.com/longhorn/longhorn-engine => github.com/c3y1huang/longhorn-engine v1.4.0-rc1.0.20230817031518-40783211d476
```
c3y1huang, line 5: TODO: update after https://github.com/longhorn/longhorn-engine/pull/922 merged.

innobead (changes requested):

upgrade/util/util.go
```diff
@@ -243,6 +277,55 @@ func CheckUpgradePathSupported(namespace string, lhClient lhclientset.Interface)
...
+		"newEngineControllerAPIVersion":    newEngineVersion.ControllerAPIVersion,
+		"newEngineControllerAPIMinVersion": newEngineVersion.ControllerAPIMinVersion,
+		"newEngineClientAPIVersion":        newEngineVersion.CLIAPIVersion,
+		"newEngineClientAPIMinVersion":     newEngineVersion.CLIAPIMinVersion,
+	}).Infof("Checking if the engine upgrade path from %+v is supported", engineImageVersions)
+
+	checkSupportVersion := func(currentVersion, newVersion, newMinVersion int) error {
+		if currentVersion > newVersion {
+			return fmt.Errorf("downgrading from %v to %v is not supported", currentVersion, newVersion)
+		}
```
innobead, line 306: note: this just means we did not do for forward compatibility for longhorn manager if we have a newer engine version to commute with.

upgrade/util/util.go
```diff
@@ -243,6 +277,55 @@ func CheckUpgradePathSupported(namespace string, lhClient lhclientset.Interface)
 	return nil
 }
 
+// checkEngineUpgradePathSupported returns error if the upgrade path from
+// current EngineImage version to emeta.ControllerAPIVersion is not supported.
+func checkEngineUpgradePathSupported(namespace string, lhClient lhclientset.Interface, newEngineVersion emeta.VersionOutput) (err error) {
```
innobead, line 282: Just noticed in the current implementation, we will periodically check if engine images are compatible or not.

[/controller/engine_image_controller.go#L327-L331](https://github.com/longhorn/longhorn-manager/blob/88d986ee8420fcecc0903e27627063f7518ac69b/controller/engine_image_controller.go#L327-L331)
```go
	if err := engineapi.CheckCLICompatibility(engineImage.Status.CLIAPIVersion, engineImage.Status.CLIAPIMinVersion); err != nil {
		engineImage.Status.Conditions = types.SetCondition(engineImage.Status.Conditions, longhorn.EngineImageConditionTypeReady, longhorn.ConditionStatusFalse, longhorn.EngineImageConditionTypeReadyReasonBinary, "incompatible")
		engineImage.Status.State = longhorn.EngineImageStateIncompatible
		return nil
	}

```

Therefore, the code here is more about runtime check during the upgrade instead of reusing EngineImage::Status, which will be periodically updated by controller reconciling loop.

innobead (changes requested):
LGTM.

One more suggestion:
- improve webhook to prevent volumes (engine/replica) to create/update by using any imcompatible engine image. You can rely on EngineImageStatus.

innobead (approved):
LGTM

PR #2108, opened 2023-08-17
Add replicaName argument to ReplicaRebuildVerify

ejweber:
longhorn/longhorn#6522

I added the replicaName argument to instance-manager's ReplicaRebuildVerify client method (and the associated gRPC request) as a part of longhorn/longhorn#5845, but I did not create a way for longhorn-manager to provide it. This PR fixes the issue and allows the tests mentioned in longhorn/longhorn#6522 to pass.


PhanLe1010 (approved):
LGTM

PR #2109, opened 2023-08-18
feat: remove non-existing source directory log

c3y1huang:
https://github.com/longhorn/longhorn/issues/3786


innobead (changes requested):
`CopyHostDirectoryContent` is not used anymore. We could just remove the whole function.

innobead (approved):

innobead:

PR #2110, opened 2023-08-18
feat: add pvc name to volume metrics

c3y1huang:
https://github.com/longhorn/longhorn/issues/5297


innobead (approved):
LGTM.

I assume the label will be empty when the collected volume is not being bound to a PVC.

innobead:

PR #2127, opened 2023-08-28
feat(backingimage): add backing image backup support

ChanYiLin:
ref: [longhorn/longhorn#4165](https://github.com/longhorn/longhorn/issues/4165)


ChanYiLin:
conflict fixed

manager/volume.go
```diff
@@ -149,6 +150,11 @@ func (m *VolumeManager) Create(name string, spec *longhorn.VolumeSpec, recurring
 		}
 	}
 
+	// restore backing image if needed
```
ChanYiLin, line 153: restore volume is actually create volume with `FromBackup` in the spec
in create volume function we check 
1. if backing image exists
2. if back up backing image exists
3. create backing image by `restore` if backup backing image exists

ChanYiLin:
conflicts fixed.

ChanYiLin:
Create a new one

PR #2133, opened 2023-08-30
refactor: replace resource lister and informer abbreviations

c3y1huang:
As Longhorn has expanded with additional Custom Resource Definitions (CRDs), the use of abbreviations may lose its clarity. Therefore, this PR proposes substituting the resource lister and informer abbreviations with their full names to improve readability.


ejweber (approved):
LGTM.

I'm don't feel particularly strongly that this change is necessary, but having recently joined the project, any time I saw a "bi", and "sm", an "rj", etc., it certainly slowed me down as I tried to figure out what type of thing was even being referred to.

innobead:

PR #2139, opened 2023-09-12
Replace spec.engineImage with spec.image in volume, replica and engine resources

derekbit:
Longhorn/longhorn#6647

### DIFF
--- a/api/model.go
+++ b/api/model.go
@@ -40,7 +40,7 @@ type Volume struct {
 	StaleReplicaTimeout              int                                    `json:"staleReplicaTimeout"`
 	State                            longhorn.VolumeState                   `json:"state"`
 	Robustness                       longhorn.VolumeRobustness              `json:"robustness"`
-	EngineImage                      string                                 `json:"engineImage"`
+	Image                            string                                 `json:"image"`
 	CurrentImage                     string                                 `json:"currentImage"`
 	BackingImage                     string                                 `json:"backingImage"`
 	Created                          string                                 `json:"created"`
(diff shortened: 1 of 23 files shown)


k8s/pkg/apis/longhorn/v1beta2/volume.go
```diff
@@ -224,9 +224,12 @@ type VolumeSpec struct {
 	NodeID string `json:"nodeID"`
 	// +optional
 	MigrationNodeID string `json:"migrationNodeID"`
+	// Deprecated: Replaced by field `image`.
```
innobead, line 227: Create an issue to remove this deprecated field in 1.7, the next feature release.

api/model.go
```diff
@@ -41,6 +41,7 @@ type Volume struct {
 	State                            longhorn.VolumeState                   `json:"state"`
 	Robustness                       longhorn.VolumeRobustness              `json:"robustness"`
 	EngineImage                      string                                 `json:"engineImage"`
+	Image                            string                                 `json:"image"`
```
innobead, line 44: Can we just remove `EngineImage` file here? It seems it will be invalid after upgrade, right?

Do we need UI part to do the corresponding change?

innobead (changes requested):

api/model.go
derekbit: Removed the `EngineImage`, thank you.

> Remove deprecated fields from CRDs in v1.7.0

Yes, tracked in https://github.com/longhorn/longhorn/issues/6685

k8s/pkg/apis/longhorn/v1beta2/volume.go
derekbit: Tracked in https://github.com/longhorn/longhorn/issues/6684

innobead (approved):

innobead:
@mergify backport v1.5.x

derekbit:
@innobead 
No need to backport to v1.5.x.
It is for the SPDK volume upgrade in v1.6.0

innobead:
@derekbit thanks for clarifying this.

PR #2140, opened 2023-09-12
feat: add snapshot max count and size

FrankYang0529:
ref: https://github.com/longhorn/longhorn/issues/6563

rely on:
* https://github.com/longhorn/longhorn-instance-manager/pull/286
* https://github.com/longhorn/longhorn-engine/pull/933

### `snapshot-max-count` setting

* Test the default value is 250.
* Test its value should be integer and between 2 to 250.
* Update the setting to 5 and create a Volume without `snapshotMaxCount`. The default `snapshotMaxCount` on the Volume will be 5.

### `snapshotMaxCount` in Volume &  `updateSnapshotMaxCount` action

* Test we can't create a volume with `snapshotMaxCount` which is lower than `2` or more than `250`.
* Create a volume with `snapshotMaxCount=3` and `snapshotMaxSize=0`.
* Attach the volume with maintenance mode.
* Create two snapshots.
* Revert to the first snapshot. (`volume-head` will be on the second branch)
* Create the third snapshot.
* Revert to the first snapshot. (`volume-head` will be on the third branch)
* Should not be able to create a new snapshot, because we already have 3 snapshots.
* Use `updateSnapshotMaxCount` action to update `snapshotMaxCount` to `4`.
* Create the fourth snapshot.
* Should not be able to use `updateSnapshotMaxCount` action to update `snapshotMaxCount` to `3`, because we can't make `snapshotMaxCount` be smaller than current usage.
* Delete a snapshot.
* Use `updateSnapshotMaxCount` action to update `snapshotMaxCount` to `3`.

### `snapshotMaxSize` in Volume &  `updateSnapshotMaxSize` action

* Test we can't create a volume with `snapshotMaxSize` which is more than `0` and lower than `2 * volume.spec.size`.
* Create a volume with `size=1073741824` (1G), `snapshotMaxCount=250` and `snapshotMaxSize=2357198848` (2.2G).
* Create a related PV/PVC.
* Use a pod to mount volume.
* Create `file1` with `950M` in the volume.
* Create the first snapshot.
* Remove the `file1` and create `file2` with `950M` in the volume.
* Create the second snapshot.
* Remove the `file2` and create `file3` with `950M` in the volume.
* Should not be able to create a new snapshot, because the total snapshot size will be more than `2.2G`.
* Use `updateSnapshotMaxSize` action to update `snapshotMaxSize` to `3430940672` (3.2G).
* Create the third snapshot.
* Should not be able to use `updateSnapshotMaxSize` action to update `snapshotMaxSize` to `2357198848` (2.2G), because we can't make `snapshotMaxSize` be smaller than current usage.
* Delete a snapshot.
* Use `updateSnapshotMaxSize` action to update `snapshotMaxSize` to `2357198848` (2.2G).


k8s/pkg/apis/longhorn/v1beta2/engine.go
```diff
@@ -138,6 +138,11 @@ type EngineSpec struct {
 	UnmapMarkSnapChainRemovedEnabled bool `json:"unmapMarkSnapChainRemovedEnabled"`
 	// +optional
 	Active bool `json:"active"`
+	// +optional
+	SnapshotMaxCount int `json:"snapshotMaxCount"`
+	// +kubebuilder:validation:Type=string
+	// +optional
+	SnapshotMaxSize int64 `json:"snapshotMaxSize,string"`
```
innobead, line 145: Why `kubebuilder:validation:Type=string` instead of `int`?

controller/engine_controller.go
```diff
@@ -973,6 +973,24 @@ func (m *EngineMonitor) refresh(engine *longhorn.Engine) error {
 						volumeInfo.UnmapMarkSnapChainRemoved, engine.Spec.UnmapMarkSnapChainRemovedEnabled)
 				}
 			}
+
+			engine.Status.SnapshotMaxCount = volumeInfo.MaximumSnapshotCount
```
innobead, line 977: Why we need to do this update? Shouldn't `engine.Status.SnapshotMaxCount` be synced eventually from the engine? The same question for `engine.Status.SnapshotMaxSize`.

engineapi/instance_manager.go
```diff
@@ -338,6 +338,8 @@ func getBinaryAndArgsForEngineProcessCreation(e *longhorn.Engine,
 
 	if engineCLIAPIVersion >= 9 {
 		args = append([]string{"--engine-instance-name", e.Name}, args...)
+		args = append(args, "--snapshot-max-count", strconv.Itoa(e.Spec.SnapshotMaxCount))
+		args = append(args, "--snapshot-max-size", strconv.FormatInt(e.Spec.SnapshotMaxSize, 10))
```
innobead, line 342: ~This should be in a new API version, right? as the command interface is changed.~

`9` is actually updated to 1.6, so it's OK here.

webhook/resources/volume/validator.go
```diff
@@ -410,3 +431,54 @@ func (v *volumeValidator) canDisableRevisionCounter(engineImage string) (bool, e
...
+func validateSnapshotMaxCount(snapshotMaxCount int) error {
+	if snapshotMaxCount < 2 || snapshotMaxCount > 250 {
+		return fmt.Errorf("snapshot max count should be between 2 to 250")
+	}
+	return nil
+}
+
+func validateSnapshotMaxSize(size, snapshotMaxSize int64) error {
+	if snapshotMaxSize != 0 && snapshotMaxSize < size*2 {
+		return fmt.Errorf("snapshot max size can be 0 or at least twice of volume size")
```
innobead, line 444: `snapshot max size can not be 0 and at least twice of volume size`

webhook/resources/volume/validator.go
```diff
@@ -410,3 +431,54 @@ func (v *volumeValidator) canDisableRevisionCounter(engineImage string) (bool, e
...
+	var (
+		currentSnapshotCount     int
+		currentTotalSnapshotSize int64
+		engine                   *longhorn.Engine
+	)
+	engines, err := v.ds.ListVolumeEngines(newVolume.Name)
+	if err != nil && !datastore.ErrorIsNotFound(err) {
+		return werror.NewInternalError(fmt.Sprintf("can't list engines for volume %s, err %v", newVolume.Name, err))
+	} else if len(engines) >= 2 {
+		return werror.NewInvalidError("can't update snapshotMaxCount or snapshotMaxSize during migration", "")
```
innobead, line 459: Suggest checking `Status.CurrentMigrationNodeID` beforehand. However, ideally, we should allow users to set even it's during the migration, because it should eventually take effect. 

It's OK for this implementation now.

webhook/resources/volume/validator.go
```diff
@@ -410,3 +431,54 @@ func (v *volumeValidator) canDisableRevisionCounter(engineImage string) (bool, e
...
+	} else if len(engines) == 0 {
+		return nil
+	}
+
+	for _, e := range engines {
+		engine = e
```
innobead, line 465: We should check the number of engines just one, then use that engine to continue. 

If it's not one, throw an error w/ potential reasons.

webhook/resources/volume/validator.go
```diff
@@ -410,3 +431,54 @@ func (v *volumeValidator) canDisableRevisionCounter(engineImage string) (bool, e
...
+		}
+		currentSnapshotCount++
+		snapshotSize, err := strconv.ParseInt(snapshotInfo.Size, 10, 64)
+		if err != nil {
+			return werror.NewInternalError(fmt.Sprintf("can't parse size %s from snapshot %s in volume %s, err %v", snapshotInfo.Size, snapshotInfo.Name, newVolume.Name, err))
+		}
+		currentTotalSnapshotSize += snapshotSize
+	}
+
+	if currentSnapshotCount > newVolume.Spec.SnapshotMaxCount || currentTotalSnapshotSize > newVolume.Spec.SnapshotMaxSize {
```
innobead, line 480: If `SnapshotMaxSize` not set, we don't need that validation, right?

P.S. `SnapshotMaxCount` will be mutated, so it should have a value.

innobead (changes requested):

k8s/pkg/apis/longhorn/v1beta2/engine.go
FrankYang0529: I followed `VolumeSpec.Size`. @derekbit, could you share the idea to use string validation? This was added from https://github.com/longhorn/longhorn-manager/pull/1136. Thank you.

controller/engine_controller.go
FrankYang0529: The `engine.Status.SnapshotMaxCount` means the current value in longhorn-engine. If it's not equal to `engine.Spec.SnapshotMaxCount`, we send a request to update it.

webhook/resources/volume/validator.go
FrankYang0529: Yes, good catch. Thank you!

webhook/resources/volume/mutator.go
```diff
@@ -239,6 +253,20 @@ func (v *volumeMutator) Update(request *admission.Request, oldObj runtime.Object
...
+		snapshotMaxCount, err := v.getSnapshotMaxCount()
+		if err != nil {
+			err = errors.Wrap(err, "BUG: cannot get valid number for setting snapshot max count")
+			return nil, werror.NewInvalidError(err.Error(), "")
+		}
+		logrus.Infof("Use the default snapshot max count %v", snapshotMaxCount)
+		patchOps = append(patchOps, fmt.Sprintf(`{"op": "replace", "path": "/spec/snapshotMaxCount", "value": %v}`, snapshotMaxCount))
+	}
+
+	if oldVolume.Spec.Size != volume.Spec.Size && volume.Spec.SnapshotMaxSize != 0 && volume.Spec.Size*2 > volume.Spec.SnapshotMaxSize {
```
innobead, line 266: Why we need to check `oldVolume.Spec.Size != volume.Spec.Size`

webhook/resources/volume/validator.go
```diff
@@ -452,3 +473,54 @@ func (v *volumeValidator) canDisableRevisionCounter(image string, dataEngine lon
...
+		return werror.NewInternalError(fmt.Sprintf("can't list engines for volume %s, err %v", newVolume.Name, err))
+	} else if len(engines) >= 2 {
+		return werror.NewInvalidError("can't update snapshotMaxCount or snapshotMaxSize during migration", "")
+	} else if len(engines) == 0 {
+		return nil
+	}
+
+	for _, e := range engines {
+		engine = e
+	}
```
innobead, line 508: engine := engines[0]

webhook/resources/volume/mutator.go
```diff
@@ -180,6 +180,16 @@ func (v *volumeMutator) Create(request *admission.Request, newObj runtime.Object
 		patchOps = append(patchOps, fmt.Sprintf(`{"op": "replace", "path": "/spec/backupCompressionMethod", "value": "%s"}`, defaultCompressionMethod))
 	}
 
+	if volume.Spec.SnapshotMaxCount == 0 {
+		snapshotMaxCount, err := v.getSnapshotMaxCount()
+		if err != nil {
+			err = errors.Wrap(err, "BUG: cannot get valid number for setting snapshot max count")
+			return nil, werror.NewInvalidError(err.Error(), "")
+		}
```
innobead, line 188: We should do the upgrade path as well to make the existing resources to update during the upgrade. Go check upgrade package.

innobead (changes requested):
A few comments.

controller/volume_controller.go
```diff
@@ -3191,6 +3212,8 @@ func (c *VolumeController) createEngine(v *longhorn.Volume, currentEngineName st
 			ReplicaAddressMap:         map[string]string{},
 			UpgradedReplicaAddressMap: map[string]string{},
 			RevisionCounterDisabled:   v.Spec.RevisionCounterDisabled,
+			SnapshotMaxCount:          v.Spec.SnapshotMaxCount,
+			SnapshotMaxSize:           v.Spec.Size,
```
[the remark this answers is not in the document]
FrankYang0529: Yes, thanks for catching this.

webhook/resources/volume/mutator.go
FrankYang0529: This check is to address this one https://github.com/longhorn/longhorn/pull/6634#discussion_r1438053336.

webhook/resources/volume/validator.go
FrankYang0529: The `engines` is map here, so we can't use `engines[0]`.

webhook/resources/volume/mutator.go
FrankYang0529: Yes, I modify `SnapshotMaxCount` to `250` in upgrade.go file, so we don't hit this one.

FrankYang0529: I remove the change currently, because we can't get latest setting in `Upgrade` function when we just started a new longhorn-manager. We call `Upgrade` function before we create default setting (`UpdateCustomizedSettings `). I think we can just leave mutator to handle it.
Ref: https://github.com/longhorn/longhorn-manager/blob/420b7798b3c5b2d747e391aa6065b3c2695d07fe/app/daemon.go#L162-L185

innobead: For backward compatibility, we need to make sure if no snapshot max count is set for any existing volumes, the behavior should be the same, with the default limit (internal, it should be 250). 

Then, your original code in the upgrade should be OK. Right?

FrankYang0529: Yeah, let me set a global variable to make sure it's consistent with the default setting. Thank you.

FrankYang0529: Updated it. Thanks.

innobead (approved):

PR #2145, opened 2023-09-14
offline rebuilding: check spc.backendStoreDriver using a defensive way

derekbit:
This is inspired by https://github.com/longhorn/longhorn/issues/6688.
Currently, we haven't figured out the root cause of the issue yet, and the volume spec fields are correct. 
However, we can adopt a defensive way when checking the spec fields.

Longhorn/longhorn#6688

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -681,15 +681,13 @@ func (c *VolumeController) ReconcileEngineReplicaState(v *longhorn.Volume, es ma
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


shuo-wu (approved):

james-munson (approved):
LGTM, too.

ejweber (approved):
LGTM. Should give us an extra layer of protection.

innobead (approved):

innobead:
@mergify backport v1.5.x

PR #2146, opened 2023-09-14
feat(volume): add condition for wait backing image

ChanYiLin:
ref: [longhorn/longhorn#6615](https://github.com/longhorn/longhorn/issues/6615)

- add condition in replica to show the status of BackingImage
- add condition in volume to aggregate the status of BackingImage from replicas


k8s/pkg/apis/longhorn/v1beta2/replica.go
```diff
@@ -10,7 +10,10 @@ const (
 )
 
 const (
-	ReplicaConditionTypeRebuildFailed = "RebuildFailed"
+	ReplicaConditionTypeRebuildFailed                = "RebuildFailed"
+	ReplicaConditionTypeWaitForBackingImage          = "WaitForBackingImage"
+	ReplicaConditionReasonWaitForBackingImageFailed  = "GetBackingImageFailed"
+	ReplicaConditionReasonWaitForBackingImageWaiting = "Waiting"
```
innobead, line 17: Whats the difference between `ReplicaConditionReasonWaitForBackingImageWaiting` and `ReplicaConditionTypeWaitForBackingImage`

k8s/pkg/apis/longhorn/v1beta2/volume.go
```diff
@@ -99,9 +99,10 @@ type VolumeCloneStatus struct {
...
 
 const (
-	VolumeConditionTypeScheduled        = "scheduled"
-	VolumeConditionTypeRestore          = "restore"
-	VolumeConditionTypeTooManySnapshots = "toomanysnapshots"
+	VolumeConditionTypeScheduled           = "scheduled"
+	VolumeConditionTypeRestore             = "restore"
+	VolumeConditionTypeTooManySnapshots    = "toomanysnapshots"
+	VolumeConditionTypeWaitForBackingImage = "WaitForBackingImage"
 )
```
innobead, line 106: Do we have any convention of condition value? Just noticed some existing values are just lower case? Should we correct them for consistency?

k8s/pkg/apis/longhorn/v1beta2/replica.go
ChanYiLin: One is `Reason`, the other one is `Type`
A condition is made of 
```
type Condition struct {
	// Type is the type of the condition.
	Type string `json:"type"`
	// Status is the status of the condition.
	// Can be True, False, Unknown.
	Status ConditionStatus `json:"status"`
	// Last time we probed the condition.
	LastProbeTime string `json:"lastProbeTime"`
	// Last time the condition transitioned from one status to another.
	LastTransitionTime string `json:"lastTransitionTime"`
	// Unique, one-word, CamelCase reason for the condition's last transition.
	Reason string `json:"reason"`
	// Human-readable message indicating details about last transition.
	Message string `json:"message"`
}
```

k8s/pkg/apis/longhorn/v1beta2/volume.go
ChanYiLin: From Kubernetes community, https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties
It should be CamelCase

> Condition type names should make sense for humans; neither positive nor negative polarity can be recommended as a general rule. A negative condition like "MemoryExhausted" may be easier for humans to understand than "SufficientMemory". Conversely, "Ready" or "Succeeded" may be easier to understand than "Failed", because "Failed=Unknown" or "Failed=False" may cause double-negative confusion.

k8s/pkg/apis/longhorn/v1beta2/volume.go
```diff
@@ -110,6 +111,8 @@ const (
 	VolumeConditionReasonRestoreInProgress             = "RestoreInProgress"
 	VolumeConditionReasonRestoreFailure                = "RestoreFailure"
 	VolumeConditionReasonTooManySnapshots              = "TooManySnapshots"
+	VolumeConditionReasonWaitForBackingImageFailed     = "GetBackingImageFailed"
+	VolumeConditionReasonWaitForBackingImageWaiting    = "Waiting"
 )
```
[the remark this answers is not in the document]
ChanYiLin: From Volume/Replicas perspective, they have no knowledge why backing image is not ready, it only blindly waiting
https://github.com/longhorn/longhorn-manager/pull/2146/files#diff-3e9e243d2398a7a25a0b80f42d1a9118c4804d3fbc9f1f2f7e48296b9bfcfdbcR377

Users can then check from `BackingImage` UI to see why backing image is not ready yet

k8s/pkg/apis/longhorn/v1beta2/volume.go
ChanYiLin: Let me fix them in this PR.

ChanYiLin: Fixed.

innobead (approved):
LGTM

k8s/pkg/apis/longhorn/v1beta2/volume.go
innobead: cc @longhorn/qa this change could impact our existing e2e test cases if we use condition for assertion.

innobead:

PR #2147, opened 2023-09-14
Prevent orphan snapshot attachment tickets

ejweber:
https://github.com/longhorn/longhorn/issues/6652

It is possible for a snapshot to be reconciled again even after its finalizer is removed and it is deleted. Because of this, it is dangerous to create an `attachmentTicket` for a snapshot after its `deletionTimestamp` has been set. (If this is the last time we ever reconcile it and we fail to clean up the `attachmentTicket` in this reconcile loop, the `attachmentTicket` will exist forever).

This PR introduces a snapshot annotation. The FIRST time we successfully create an `attachmentTicket` after a snapshot's `deletionTimestamp` is set, we update the snapshot with this annotation. We NEVER create an `attachmentTicket` for this snapshot again (although we can update the `nodeID` for the existing one as necessary). This ensures we do not create an `attachmentTicket` on the final reconciliation.

This PR also adds upgrade logic to clean up existing orphan snapshot `attachmentTickets`.

### DIFF
--- a/controller/snapshot_controller.go
+++ b/controller/snapshot_controller.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/longhorn/longhorn-manager/datastore"
 	"github.com/longhorn/longhorn-manager/engineapi"
+	"github.com/longhorn/longhorn-manager/types"
 	"github.com/longhorn/longhorn-manager/util"
 
 	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
@@ -299,15 +300,17 @@ func (sc *SnapshotController) reconcile(snapshotName string) (err error) {
 		if err != nil && !shouldUpdateObject(err) {
 			return
 		}
-		if reflect.DeepEqual(existingSnapshot.Status, snapshot.Status) {
-			return
+		if !reflect.DeepEqual(existingSnapshot.Annotations, snapshot.Annotations) {
+			// We need to annotate this deleting snapshot.
+			if snapshot, err = sc.ds.UpdateSnapshot(snapshot); err != nil {
+				return
+			}
+		} else if !reflect.DeepEqual(existingSnapshot.Status, snapshot.Status) {
+			if snapshot, err = sc.ds.UpdateSnapshotStatus(snapshot); err != nil {
+				return
+			}
+			sc.generatingEventsForSnapshot(existingSnapshot, snapshot)
 		}
-
-		snapshot, err = sc.ds.UpdateSnapshotStatus(snapshot)
-		if err != nil {
-			return
-		}
-		sc.generatingEventsForSnapshot(existingSnapshot, snapshot)
 	}()
 
 	// deleting snapshotCR
@@ -344,8 +347,25 @@ func (sc *SnapshotController) reconcile(snapshotName string) (err error) {
 			return sc.handleAttachmentTicketDeletion(snapshot)
 		}
 
-		if err := sc.handleAttachmentTicketCreation(snapshot); err != nil {
-			return err
+		// It is possible for us to be reconciling an out-of-date snapshot that has had its finalizer removed and has
+		// been deleted from the cluster, so we must take special care around attachment tickets.
+		if val, ok := snapshot.Annotations[types.AttachingForDeletionAnnotationKey]; ok && val == "true" {
+			// It is fine to update an attachment ticket more than once after DeletionTimestamp is set. If we attempt
+			// to update an out-of-date volume attachment, we will simply fail.
+			if err := sc.handleAttachmentTicketModification(snapshot, false); err != nil {
+				return err
+			}
+		} else {
+			// Only create an attachment ticket once after DeletionTimestamp is set and then immediately return. We must
+			// not create an attachment ticket again in case this is the last time we reconcile.
+			if err := sc.handleAttachmentTicketModification(snapshot, true); err != nil {
+				return err
+			}
+			if snapshot.Annotations == nil {
+				snapshot.Annotations = make(map[string]string)
+			}
+			snapshot.Annotations[types.AttachingForDeletionAnnotationKey] = "true"
+			return nil
 		}
 
 		if engine.Status.CurrentState != longhorn.InstanceStateRunning {
@@ -387,7 +407,7 @@ func (sc *SnapshotController) reconcile(snapshotName string) (err error) {
 
 	// newly created snapshotCR by user
 	if requestCreateNewSnapshot && !alreadyCreatedBefore {
-		if err := sc.handleAttachmentTicketCreation(snapshot); err != nil {
+		if err := sc.handleAttachmentTicketModification(snapshot, true); err != nil {
 			return err
 		}
 		if engine.Status.CurrentState != longhorn.InstanceStateRunning {
@@ -450,8 +470,9 @@ func (sc *SnapshotController) handleAttachmentTicketDeletion(snap *longhorn.Snap
 	return nil
 }
 
-// handleAttachmentTicketCreation check and create attachment so that the source volume is attached if needed
-func (sc *SnapshotController) handleAttachmentTicketCreation(snap *longhorn.Snapshot) (err error) {
+// handleAttachmentTicketModification creates or updates an attachment ticket if allowCreation is true. If allowCreation
+// is false, it only updates an existing attachment ticket.
+func (sc *SnapshotController) handleAttachmentTicketModification(snap *longhorn.Snapshot, allowCreation bool) (err error) {
 	defer func() {
 		err = errors.Wrap(err, "handleAttachmentTicketCreation: failed to create/update attachment")
 	}()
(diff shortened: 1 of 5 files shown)


upgrade/v15xto160/upgrade.go
```diff
@@ -43,6 +45,48 @@ func upgradeVolumes(namespace string, lhClient *lhclientset.Clientset, resourceM
...
+		return errors.Wrapf(err, "failed to list all existing Snapshots during the Longhorn VolumeAttachment a upgrade")
+	}
+
+	ticketIDsForExistingSnapshotsMap := map[string]interface{}{}
+	for snapshotName := range snapshotMap {
+		ticketID := longhorn.GetAttachmentTicketID(longhorn.AttacherTypeSnapshotController, snapshotName)
+		ticketIDsForExistingSnapshotsMap[ticketID] = nil
+	}
+
+	// Previous Longhorn versions may have created attachmentTickets for snapshots that no longer exist. Clean these up.
```
james-munson, line 75: Would you suppose that it would be appropriate to clean up orphaned backup VATs similarly?

upgrade/util/util.go
```diff
@@ -911,23 +931,21 @@ func updateOrphans(namespace string, lhClient *lhclientset.Clientset, orphans ma
...
+		volumeAttachment, ok := volumeAttachments[existingVolumeAttachment.Name]
+		if !ok {
 			continue
 		}
-		if _, err = lhClient.LonghornV1beta2().VolumeAttachments(namespace).Create(context.TODO(), va, metav1.CreateOptions{}); err != nil {
-			return err
+		if !reflect.DeepEqual(existingVolumeAttachment.Spec, volumeAttachment.Spec) ||
+			!reflect.DeepEqual(existingVolumeAttachment.ObjectMeta, volumeAttachment.ObjectMeta) {
+			if _, err = lhClient.LonghornV1beta2().VolumeAttachments(namespace).Update(context.TODO(), volumeAttachment, metav1.UpdateOptions{}); err != nil {
+				return err
```
james-munson, line 947: I know this is a common idiom in the code, but it gives me the willies to punt halfway through a `for` loop and abandon later items that might well have worked.  Seems like we could accumulate a list of the errors while completing the list.

controller/snapshot_controller.go
```diff
@@ -299,15 +300,17 @@ func (sc *SnapshotController) reconcile(snapshotName string) (err error) {
...
-		}
-
-		snapshot, err = sc.ds.UpdateSnapshotStatus(snapshot)
-		if err != nil {
-			return
+		if !reflect.DeepEqual(existingSnapshot.Annotations, snapshot.Annotations) {
+			// We need to annotate this deleting snapshot.
+			if snapshot, err = sc.ds.UpdateSnapshot(snapshot); err != nil {
+				return
+			}
```
shuo-wu, line 307: NIT: Can we update the CR spec/meta inside its own controller?

controller/snapshot_controller.go
```diff
@@ -344,8 +347,25 @@ func (sc *SnapshotController) reconcile(snapshotName string) (err error) {
...
+			// Only create an attachment ticket once after DeletionTimestamp is set and then immediately return. We must
+			// not create an attachment ticket again in case this is the last time we reconcile.
+			if err := sc.handleAttachmentTicketModification(snapshot, true); err != nil {
+				return err
+			}
+			if snapshot.Annotations == nil {
+				snapshot.Annotations = make(map[string]string)
+			}
+			snapshot.Annotations[types.AttachingForDeletionAnnotationKey] = "true"
+			return nil
```
shuo-wu, line 368:
```suggest
requireVATicket := snapshot.Annotations[types.AttachingForDeletionAnnotationKey] == "true"
if err := sc.handleAttachmentTicketModification(snapshot, requireVATicket); err != nil {
	return err
}
snapshot.Annotations[types.AttachingForDeletionAnnotationKey] = "true"
```

shuo-wu: BTW, we can do `snapshot.Annotations = make(map[string]string)` in the mutation webhook.

controller/snapshot_controller.go
ejweber: I discussed this with @PhanLe1010 and @james-munson as well, but am waiting for their review to see what the consensus is. The other option is a `snapshot.status.attachingForDeletion` field. This more correctly adheres to our standard (a controller can only update its own status and other resource specs). I feel that the annotation is more natural than a status field, but will switch if that's what the team wants.

controller/snapshot_controller.go
ejweber: Good suggestions. They clean things up a bit since:

- Checking for `ok` and `"true"` is redundant. If `!ok`, then we get `""`.
- In both cases, we take the same basic action. The only differente is the boolean we pass.
- It is fine to set the annotation to true in either case (even if it's been done before).
- We do not actually need to automatically return.

> BTW, we can do snapshot.Annotations = make(map[string]string) in the mutation webhook.

Yes, this can be done, but I prefer we do a nil check every time we start writing to a map, just in case.

upgrade/util/util.go
ejweber: I get your point, though I think changing all of these occurrences is outside the scope of the issue.

upgrade/v15xto160/upgrade.go
ejweber: Yes, though if I understand https://github.com/longhorn/longhorn/issues/6654 correctly, the criteria would be slightly different. We would be looking for completed backups (and not non-existent ones) right? I think we should add and test this logic there.

upgrade/util/util.go
james-munson: Not asking to change the world, just this little part of it.  But I get your point that this is suspenders-and-a-belt code that we don't really expect to find work, so perhaps simplest is best.

upgrade/v15xto160/upgrade.go
PhanLe1010: For the backup, I think we don't need the upgrade logic because when longhorn-manager starts the backup controller it will loop over all backup CRs and try to sync them. The new backup controller logic will kick in and remove the bad VA tickets

controller/snapshot_controller.go
```diff
@@ -344,9 +347,19 @@ func (sc *SnapshotController) reconcile(snapshotName string) (err error) {
...
+		// attachment will simply fail), but we must only create one once (to avoid recreating one the list time we
+		// reconcile).
+		requireVATicket := snapshot.Annotations[types.AttachingForDeletionAnnotationKey] == "true"
+		if err := sc.handleAttachmentTicketModification(snapshot, requireVATicket); err != nil {
 			return err
 		}
+		if snapshot.Annotations == nil {
+			snapshot.Annotations = make(map[string]string)
+		}
+		snapshot.Annotations[types.AttachingForDeletionAnnotationKey] = "true"
```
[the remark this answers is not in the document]
ejweber: Interesting. Yes, I think you are right. My original logic included a return if the annotation was not set. We would NEVER proceed to delete the finalizer if we weren't operating on an annotated snapshot, so there was no way of this happening.

https://github.com/longhorn/longhorn-manager/pull/2147/commits/10b343eced295276fe3467d287c6ee89a719417d

I removed the immediate return while simplifying things during the review. IIRC, the new code still passed my testing, but we need to put the immediate return back. Unfortunately, I'm not going to get this done before I leave for vacation, but WDYT of the flow in that linked commit?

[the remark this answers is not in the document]
PhanLe1010: Sounds good. Let's continue on this one after you are back from the vacation.
The logic in the liked comment LGTM.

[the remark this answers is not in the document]
PhanLe1010: Btw, I see another potential root cause that can trigger the issue in one sync loop:
1. snapshot controller create VA ticket https://github.dev/longhorn/longhorn-manager/blob/714bc966d5e271e9f1159b9266578da7f8a27600/controller/snapshot_controller.go#L347-L349
1. The VA object in ETCD is updated but not in the cache
1. snapshot controller checks the VA object in the cache and doesn't delete/update the VA because it doesn't see the ticket 
1. snapshot controller removes finalizer.
1. Assume that cach is updated fast enough so that in the next sync loop, snapshot controller doesn't see snapshot CR and never sync it again
1. Now we have orphan VA ticket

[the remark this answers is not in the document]
innobead: It seems we have a concern to use outdated cache to do the resource reconciling and following update. Should we just bypass the cache to do resource query and update? Only use the cache informer to watch resource change. 

Can we check the best practice of controller implementation as well? like K8s builtin controllers, have we checked them?

ejweber:
Closing this in favor of https://github.com/longhorn/longhorn-manager/pull/2197 based on the discussion following https://github.com/longhorn/longhorn-manager/pull/2147#discussion_r1329303571, especially https://github.com/longhorn/longhorn-manager/pull/2147#discussion_r1329341605. It seems that no matter what we require additional communication with the API server to ensure we don't make a mistake. It's more straightforward to handle the root cause by doing an uncached GET of a deleting snapshot to confirm it hasn't already been garbage collected before creating an attachment ticket.

PR #2148, opened 2023-09-18
Move common functions and types to backupstore lib

c3y1huang:
https://github.com/longhorn/longhorn/issues/6514


c3y1huang:
cc @derekbit

innobead (approved):

PR #2150, opened 2023-09-18
refactor(logging): change scheduling related message to debug

ChanYiLin:
ref: [longhorn/longhorn#6019](https://github.com/longhorn/longhorn/issues/6019)

To prevent log flooding issue here https://github.com/longhorn/longhorn/issues/4069
We could move most of scheduling log messages to debug level and only keep those important decision message

And only do `getIsSchedulableToDiskNodes` check when there is `adjustCount` so we can reduce the log from checking function

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -2131,7 +2131,7 @@ func (c *VolumeController) getReplicaCountForAutoBalanceLeastEffort(v *longhorn.
 	if err != nil {
 		return 0
 	}
-	log.Infof("Found %v replica candidate for auto-balance", adjustCount)
+	log.Debugf("Found %v replica candidate for auto-balance", adjustCount)
 	return adjustCount
 }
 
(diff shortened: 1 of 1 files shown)


controller/volume_controller.go
```diff
@@ -2281,7 +2281,7 @@ func (c *VolumeController) getReplicaCountForAutoBalanceZone(v *longhorn.Volume,
 		unusedZone[node.Status.Zone] = append(unusedZone[node.Status.Zone], nodeName)
 	}
 	if len(unusedZone) == 0 {
-		log.Info("Balanced, all ready zones are used by this volume")
+		log.Debugf("Balanced, all ready zones are used by this volume")
```
innobead, line 2284: How about the below log?

```go
log.Infof("Found %v zone available for auto-balance duplicates in %v", adjustCount, zoneExtraRs)
```

controller/volume_controller.go
```diff
@@ -2416,7 +2416,7 @@ func (c *VolumeController) getReplicaCountForAutoBalanceNode(v *longhorn.Volume,
 		}
 	}
 	if len(nodeExtraRs) == len(readyNodes) {
-		log.Info("Balanced, all ready nodes are used by this volume")
+		log.Debugf("Balanced, all ready nodes are used by this volume")
 		return 0, nodeExtraRs, nil
```
innobead, line 2420: How about the below code?

```go
	log.Infof("Found %v node available for auto-balance duplicates in %v", adjustCount, nodeExtraRs)

```

innobead (changes requested):
LGTM. Just a few comments need to resolve.

innobead (approved):
@ChanYiLin answered the comments on Slack.

innobead:

controller/volume_controller.go
ChanYiLin: This message is important scheduling decision message, need to keep in `Info`

ChanYiLin: This message is important scheduling decision message, need to keep in `Info`

controller/volume_controller.go
ChanYiLin: This message is important scheduling decision message, need to keep in Info

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/6712

PR #2154, opened 2023-09-18
Refuse to create a duplicate engine

ejweber:
longhorn/longhorn#6682

I have tested this PR and it accomplishes its objective in the "normal" case. However, I think it may be just as susceptible to corner cases as the volume controller is without it. If the webhook's engine cache is out-of-date when it is listing engines (e.g. if one volume controller or two volume controllers decided to create duplicate engines nearly simultaneously), it will still allow the creation of duplicates. I'll discuss this further in the linked issue.

### DIFF
--- a/webhook/resources/engine/validator.go
+++ b/webhook/resources/engine/validator.go
@@ -41,6 +41,15 @@ func (e *engineValidator) Resource() admission.Resource {
 func (e *engineValidator) Create(request *admission.Request, newObj runtime.Object) error {
 	engine := newObj.(*longhorn.Engine)
 
+	volume, err := e.ds.GetVolume(engine.Spec.VolumeName)
+	if err != nil {
+		if datastore.ErrorIsNotFound(err) {
+			return werror.NewInvalidError("volume does not exist for engine", "spec.volumeName")
+		}
+		err = errors.Wrap(err, "failed to get volume for engine")
+		return werror.NewInternalError(err.Error())
+	}
+
 	if engine.Spec.BackendStoreDriver == longhorn.BackendStoreDriverTypeV2 {
 		v2DataEngineEnabled, err := e.ds.GetSettingAsBool(types.SettingNameV2DataEngine)
 		if err != nil {
@@ -52,7 +61,7 @@ func (e *engineValidator) Create(request *admission.Request, newObj runtime.Obje
 		}
 	}
 
-	return nil
+	return e.validateNumberOfEngines(engine, volume)
 }
 
 func (e *engineValidator) Update(request *admission.Request, oldObj runtime.Object, newObj runtime.Object) error {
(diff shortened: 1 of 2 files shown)


ejweber:
Tests so far:

Create a duplicate engine for a non-migratable volume:

1. Create a volume in the UI.
2. Copy the yaml for the engine that is automatically created.
3. Edit the yaml to have a different name.
4. Attempt to apply the yaml.
5. `The request is invalid: spec.volumeName: engine creation would result in 2 engines for non-migratable volume`

Create two duplicate engines for a migratable volume:

1. Create a migratable storage class.
2. Create a PVC that references the storage class.
3. Copy the yaml for the engine that is automatically created TWICE.
4. Edit BOTH yamls to have different names.
5. Apply the first yaml.
6. Attempt to apply the second yaml.
7. `The request is invalid: spec.volumeName: engine creation would result in 3 engines for migratable volume`

Create a volume and two referencing engines simultaneously many times. The below script never exits after detecting multiple engines.

```bash
for i in {1..100}; do kubectl apply -f duplicate-test.yaml; if [[ $(kubectl get engine --no-headers -n longhorn-system | wc -l) > 1 ]]; then break; fi; kubectl delete volume -n longhorn-system test; sleep 15; done`
```

webhook/resources/engine/validator.go
```diff
@@ -41,6 +41,15 @@ func (e *engineValidator) Resource() admission.Resource {
 func (e *engineValidator) Create(request *admission.Request, newObj runtime.Object) error {
 	engine := newObj.(*longhorn.Engine)
 
+	volume, err := e.ds.GetVolume(engine.Spec.VolumeName)
+	if err != nil {
+		if datastore.ErrorIsNotFound(err) {
+			return werror.NewInvalidError("volume does not exist for engine", "spec.volumeName")
+		}
+		errors.Wrap(err, "failed to get volume for engine")
+		return werror.NewInternalError(err.Error())
```
innobead, line 50: This should be

```go
		err = errors.Wrap(err, "failed to get volume for engine")
		return werror.NewInternalError(err.Error())
```

webhook/resources/engine/validator.go
```diff
@@ -68,3 +77,23 @@ func (e *engineValidator) Update(request *admission.Request, oldObj runtime.Obje
 
 	return nil
 }
+
+func (e *engineValidator) validateNumberOfEngines(newEngine *longhorn.Engine, volume *longhorn.Volume) error {
+	volumeEngines, err := e.ds.ListVolumeEngines(newEngine.Spec.VolumeName)
+	if err != nil {
+		errors.Wrap(err, "failed to list engines for volume")
+		return werror.NewInternalError(err.Error())
```
innobead, line 85:
```go
		err = errors.Wrap(err, "failed to list engines for volume")
		return werror.NewInternalError(err.Error())
```

innobead (changes requested):
LGTM, just two comments need to resolve. 

This fix is just focusing on the correct mutation behavior for engine creation in admission webhook only.

innobead:
Maintainer is allowed to edit this PR. @PhanLe1010 could you help take over this implementation, since Eric will come back after 10/4, and we need to get this to 1.5.2 and 1.4.4. Thanks.

PhanLe1010:
I changed getting engine from cache to getting them from uncached kubeAPI server cc @innobead

webhook/resources/engine/validator.go
```diff
@@ -68,3 +77,23 @@ func (e *engineValidator) Update(request *admission.Request, oldObj runtime.Obje
...
+	}
+
+	newNumVolumeEngines := len(volumeEngines) + 1
+	if volume.Spec.Migratable && newNumVolumeEngines > 2 {
+		message := fmt.Sprintf("engine creation would result in %d engines for migratable volume", newNumVolumeEngines)
+		return werror.NewInvalidError(message, "spec.volumeName")
+	}
+	if !volume.Spec.Migratable && newNumVolumeEngines > 1 {
+		message := fmt.Sprintf("engine creation would result in %d engines for non-migratable volume", newNumVolumeEngines)
+		return werror.NewInvalidError(message, "spec.volumeName")
```
innobead, line 95: It seems this validation is to check specific conditions instead of a field, so `"spec.volumeName"` should not be required.

innobead (changes requested):

innobead:

webhook/resources/engine/validator.go
PhanLe1010: Good point. Fixed. Thank you

innobead (approved):

ejweber:
Thanks for moving this along while I was out @PhanLe1010 and @innobead.

PR #2155, opened 2023-09-18
Fix bug two active engine when volume migrating

PhanLe1010:
Only switch the new engine to active if we finish cleaning up the extra engines

longhorn/longhorn#6642

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -3807,6 +3807,8 @@ func (c *VolumeController) processMigration(v *longhorn.Volume, es map[string]*l
 			}
 		}
 
+		currentEngine.Spec.Active = true
+
 		// cleanupCorruptedOrStaleReplicas() will take care of old replicas
 		if err := c.switchActiveReplicas(rs, func(r *longhorn.Replica, engineName string) bool {
 			return r.Spec.EngineName == engineName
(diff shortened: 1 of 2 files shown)


ejweber (commented):
If I understand correctly, the primary fix is to wait to set `currentEngine.Spec.Active == true` until AFTER we clean up the old engine. Otherwise, if engine updating fails in this reconcile, we could permanently have two active engines. The change to `GetCurrentEngineAndExtras` was required because it previously relied on `GetNewCurrentEngineAndExtras` to set  `currentEngine.Spec.Active == true`.

LGTM in general.

Since we may rereconcile again before the old active engine is actually cleaned up (i.e. the process stops and the finalizer is removed), might we log "BUG: found the second new active engine %v besides %v" still, temporarily?

PhanLe1010:
> Since we may rereconcile again before the old active engine is actually cleaned up (i.e. the process stops and the finalizer is removed), might we log "BUG: found the second new active engine %v besides %v" still, temporarily?

I think no, we will not see this error temporarily because [this check ](https://github.com/longhorn/longhorn-manager/blob/04170c613098f0a3bb333e08468552dca0601e48/datastore/longhorn.go#L989-L991) only matches the engine on the new node

But it might hit this error temporarily https://github.com/longhorn/longhorn-manager/blob/04170c613098f0a3bb333e08468552dca0601e48/datastore/longhorn.go#L961

ejweber:
> But it might hit this error temporarily
> 
> https://github.com/longhorn/longhorn-manager/blob/04170c613098f0a3bb333e08468552dca0601e48/datastore/longhorn.go#L961

I'd consider this a NIT then. Generally, we reserve "BUG" for things we don't expect ever to happen. Maybe we can rewrite that message?

Otherwise, LGTM. I'll leave it to others to approve since I'm not going to get a chance to reevaluate before leaving for vacation.

PhanLe1010:
migration e2e tests result: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/4926/

innobead (approved):

innobead:

Vicente-Cheng (commented):
Sorry for missing this one 
LGTM for fixing this race. Thanks @PhanLe1010

PR #2167, opened 2023-09-20
rwx volume: use soft mode with long timeout for nfs client

derekbit:
When a RWX volume is attached, a share-manager pod embedded with a userspace NFS server is created and the volume is exported. A remote exported share is hard mounted by Longhorn, and it is then provided to the workload. When the share-manager pod or embedded NFS server is somehow crashed or unreachable, the 'hard mount' option keeps the client trying to connect to NFS server and prevents data loss.

The root cause is the Linux kernel is trying to maintain filesystem stability. Linux kernel will not allow a filesystem to be unmounted until all its pending IO is written back to storage, and the system can't shut down until all file systems are unmounted. Currently, the bug/issue is not resolved.

Longhorn/longhorn#6655


csi/node_server.go
```diff
@@ -240,11 +240,11 @@ func (ns *NodeServer) nodeStageSharedVolume(volumeID, shareEndpoint, targetPath
...
 			// "sync", // sync mode is prohibitively expensive on the client, so we allow for host defaults
-			"intr",
-			"hard",
-			//"soft", // for this release we use soft mode, so we can always cleanup mount points
-			//"timeo=30",  // This is tenths of a second, so a 3 second timeout, each retrans the timeout will be linearly increased, 3s, 6s, 9s
-			//"retrans=3", // We try the io operation for a total of 3 times, before failing, max runtime of 18s
+			// "intr",
+			//"hard",
+			"soft",      // for this release we use soft mode, so we can always cleanup mount points
+			"timeo=150", // This is tenths of a second, so a 15 second timeout, each retrans the timeout will be linearly increased, 15s, 30s, 45s
```
innobead, line 246: It seems the share manager pod is controlled by the Longhorn manager, so we don't need to consider pod eviction timeout, right? 150 seconds could be good enough based on the reconciling period?

so it will be `timeo * retrans`, it will be 450 seconds before the nfs client fails.

innobead:

csi/node_server.go
derekbit: > It seems the share manager pod is controlled by the Longhorn manager, so we don't need to consider pod eviction timeout, right? 

You're right.

derekbit: The maximum retry period is 15 + 30 + 60 = 105 seconds. It is sometimes not enough and fails the IO.
I will increase the time to a bigger value.

In addition, there is a new mount option `softerr`.
According to the description, I would use `softerr` instead. WDYT? cc @innobead 
```
A new optional additional softerr mount option to return ETIMEDOUT (instead of EIO) to the application after a 
timeout, so that applications written to be aware of this may better differentiate between the the timeout case, e.g. 
to drive a failover response, and other I/O errors, so that the client application may choose a different recovery 
action for those cases. Mounts using only the soft mount option will see the other improvements, but timeout errors 
will still be returned to the application with the EIO error.
``` 
https://blogs.oracle.com/linux/post/whats-new-for-nfs-in-unbreakable-enterprise-kernel-release-6

innobead: oh, I missed the comment. Need to adjust a bit or increase the number of retrans. 
> "timeo=150", // This is tenths of a second, so a 15 second timeout, each retrans the timeout will be linearly increased, 15s, 30s, 45s

derekbit: > "timeo=150", // This is tenths of a second, so a 15 second timeout, each retrans the timeout will be linearly increased, 15s, 30s, 45s

Yeah, need to check the version.

PR #2170, opened 2023-09-20
RWX volume: use bigger timeo, retrans and softerr for nfs client

derekbit:
- biggger timeo and retrans
- softerr: return ETIMEDOUT rather than EIO. Fall back to soft if the option is not supported

Longhorn/longhorn#6655


innobead (approved):
LGTM. 

This has a fallback when using softerr, but do we know what versions support softerr?

innobead:

derekbit:
> This has a fallback when using softerr, but do we know what versions support softerr?

Yes. From [linux kernel 5.2](https://kernelnewbies.org/Linux_5.2).

innobead:
In QA, need to test this fallback.

PR #2180, opened 2023-09-26
Make the new engine name deterministic based on the existing engine

PhanLe1010:
* If there is no engine exists yet, the name of the new engine will be `<VolumeName>-e-0`
* if there is existing engine, increase the engine suffix by one for the new engine

By doing this, we don't create extra new engines even if the volume controller run multiple sync loops with stale cache data

longhorn/longhorn#6682

### DIFF
--- a/types/types.go
+++ b/types/types.go
@@ -292,8 +292,16 @@ const (
 	replicaManagerPrefix  = instanceManagerPrefix + "r-"
 )
 
-func GenerateEngineNameForVolume(vName string) string {
-	return vName + engineSuffix + "-" + util.RandomID()
+func GenerateEngineNameForVolume(vName, currentEngineName string) string {
+	if currentEngineName == "" {
+		return vName + engineSuffix + "-" + "0"
+	}
+	parts := strings.Split(currentEngineName, "-")
+	suffix, err := strconv.Atoi(parts[len(parts)-1])
+	if err != nil {
+		return vName + engineSuffix + "-" + "1"
+	}
+	return vName + engineSuffix + "-" + strconv.Itoa(suffix+1)
 }
 
 func GenerateReplicaNameForVolume(vName string) string {
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -1513,7 +1513,7 @@ func (c *VolumeController) reconcileVolumeCreation(v *longhorn.Volume, e *longho
 
 	if len(es) == 0 {
 		// first time creation
-		e, err = c.createEngine(v, true)
+		e, err = c.createEngine(v, "")
 		if err != nil {
 			return false, e, err
 		}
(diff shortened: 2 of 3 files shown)


PhanLe1010:
e2e tests (core tests and migration tests): https://ci.longhorn.io/job/private/job/longhorn-tests-regression/4964/

james-munson (approved):
LGTM

types/types.go
```diff
@@ -292,8 +292,19 @@ const (
...
-func GenerateEngineNameForVolume(vName string) string {
-	return vName + engineSuffix + "-" + util.RandomID()
+func GenerateEngineNameForVolume(vName, currentEngineName string) string {
+	if currentEngineName == "" {
+		return vName + engineSuffix + "-" + "0"
+	}
+	parts := strings.Split(currentEngineName, "-")
+	if len(parts) < 3 {
+		return vName + engineSuffix + "-" + "1"
+	}
```
innobead, line 302: When will we have this case? After the first engine created, isn't the `parts` always 3?

innobead, line 302: nit: have a unit test for this function.

innobead (changes requested):
In general, LGTM. Just a few comments.

types/types.go
PhanLe1010: I think we should never have the case that `len(parts) < 3`. But since we are using `parts[len(parts)-1]` below, I think we should check to make sure that len(parts) nonempty to avoid panic. Since we need to check for check for len(parts) > 0 and we know that it should be >= 3, I decided to check for `NOT len(parts) < 3`. I can change it back to:
```
	if len(parts) < 1 {
		return vName + engineSuffix + "-" + "1"
	}
```
if preferred. WDYT?

innobead: Why not just throw an error? because it's not a valid case, it would be better to throw an error?

Since we only expect `vName + engineSuffix + "-" + "1"` if the current volume is `vName + engineSuffix + "-" + "0"`?

innobead:
After resolving all comments, we can merge this PR. Well done @PhanLe1010

types/types.go
PhanLe1010: We also expect `vName + engineSuffix + "-" + "1"` if the current engine is `vName + engineSuffix + "-" + "xxxxxxxxx"` (which is the older format)

Btw, I think you are right that we don't need to check for `len(parts) < 3` or `len(parts) < 1`. The length of `parts` slice is always >= 1 because the `currentEngineName` is != "" at this step. Therefore, I remove the `if len(parts) < 3 {` check. 

WDYT?

PhanLe1010: Added. Thank you

shuo-wu (approved):
In general LGTM

innobead (approved):

types/types.go
innobead: Sounds good. However, I would like to clarify that checking index out of range is always necessary to prevent any unexpected panic but if it's an invalid case, we should better raise the error explicitly instead.

innobead:

PR #2181, opened 2023-09-27
fix(backup): ensure volume attached before taking backup

ChanYiLin:
ref: [longhorn/longhorn#6699](https://github.com/longhorn/longhorn/issues/6699)

need to ensure engine status and desired statue are the same before taking backup

### DIFF
--- a/controller/backup_controller.go
+++ b/controller/backup_controller.go
@@ -659,7 +659,9 @@ func (bc *BackupController) checkMonitor(backup *longhorn.Backup, volume *longho
 		return nil, err
 	}
 
-	if engine.Status.CurrentState != longhorn.InstanceStateRunning {
+	if engine.Status.CurrentState != longhorn.InstanceStateRunning ||
+		engine.Spec.DesireState != longhorn.InstanceStateRunning ||
+		volume.Status.State != longhorn.VolumeStateAttached {
 		return nil, fmt.Errorf("waiting for engine %v to be running before enabling backup monitor", engine.Name)
 	}
 


controller/backup_controller.go
```diff
@@ -659,7 +659,9 @@ func (bc *BackupController) checkMonitor(backup *longhorn.Backup, volume *longho
 		return nil, err
 	}
 
-	if engine.Status.CurrentState != longhorn.InstanceStateRunning {
+	if engine.Status.CurrentState != longhorn.InstanceStateRunning ||
+		engine.Spec.DesireState != longhorn.InstanceStateRunning ||
+		volume.Status.State != longhorn.VolumeStateAttached {
```
james-munson, line 664: Perhaps split this into two separate checks, so that the error returned can accurately reflect that it is engine state or volume state that is the problem.

james-munson (changes requested):

innobead (approved):
LGTM.

It would be better to improve `VerifyAttachment` to ensure the ticket is really satisfied. 

[/controller/backup_controller.go#L624-L627](https://github.com/longhorn/longhorn-manager/blob/428281c75380d0cab58a2f4da6a3215e64c3aa2c/controller/backup_controller.go#L624-L627)
```go
	ok, err := bc.VerifyAttachment(backup, volume.Name)
	if err != nil {
		return nil, err
	}
```

controller/backup_controller.go
innobead: Agreed, but in this case, it could be optional because all conditions here are mandatory to just ensure the volume is really attached.

innobead:
@mergify backport v1.5.x

PR #2183, opened 2023-09-28
fix(restore): dr volume failed by delete lock

mantissahz:
DR volume will fail to start an incremental restoration in target cluster, when the source cluster is deleting the eliminated backup by deleting lock in the backup store.

We handle the full restoration at the issue longhorn/longhorn#3055 and now add a regular expression pattern to handle similar failed message to have a backoff mechanism to retry the full/incremental restoration.

Ref: longhorn/longhorn#6750

### DIFF
--- a/controller/engine_controller.go
+++ b/controller/engine_controller.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"regexp"
 	"strconv"
 	"strings"
 	"sync"
@@ -42,7 +43,7 @@ import (
 
 const (
 	unknownReplicaPrefix            = "UNKNOWN-"
-	restoreGetLockFailedMsg         = "error initiating full backup restore: failed lock"
+	restoreGetLockFailedPatternMsg  = "error initiating (full|incremental) backup restore: failed lock"
 	restoreAlreadyInProgressMsg     = "already in progress"
 	restoreAlreadyRestoredBackupMsg = "already restored backup"
 )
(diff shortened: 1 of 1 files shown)


controller/engine_controller.go
```diff
@@ -1504,6 +1505,11 @@ func handleRestoreError(log logrus.FieldLogger, engine *longhorn.Engine, rsMap m
 	return nil
 }
 
+func checkFailedLock(testString string) bool {
```
innobead, line 1508: Rename to `isReplicaRestoreFailedLockError(err *client.ReplicaError)`

innobead (changes requested):
In general, LGTM. Some code suggetions need to resolve.

innobead:

controller/engine_controller.go
```diff
@@ -42,7 +43,7 @@ import (
 
 const (
 	unknownReplicaPrefix            = "UNKNOWN-"
-	restoreGetLockFailedMsg         = "error initiating full backup restore: failed lock"
+	restoreGetLockFailedPatternMsg  = "error initiating [a-zA-Z]+ backup restore: failed lock"
```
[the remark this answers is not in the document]
mantissahz: Modified. Thanks.

controller/engine_controller.go
mantissahz: Modified. Thanks.

innobead (dismissed):

innobead (approved):

PR #2198, opened 2023-10-05
fix(setting): put settings' ReadOnly into effect

mantissahz:
Add an annotation to mark that the updating is from Longhorn and it should be able to update successfully even if setting definition is set to be read-only.

Ref: longhorn/longhorn#5989

### DIFF
--- a/types/setting.go
+++ b/types/setting.go
@@ -1576,6 +1576,7 @@ func UnmarshalNodeSelector(nodeSelectorSetting string) (map[string]string, error
 	return nodeSelector, nil
 }
 
+// GetSettingDefinition gets the setting definition in `settingDefinitions` by the parameter `name`
 func GetSettingDefinition(name SettingName) (SettingDefinition, bool) {
 	settingDefinitionsLock.RLock()
 	defer settingDefinitionsLock.RUnlock()
@@ -1583,6 +1584,7 @@ func GetSettingDefinition(name SettingName) (SettingDefinition, bool) {
 	return setting, ok
 }
 
+// SetSettingDefinition sets the setting definition in `settingDefinitions` by the parameter `name` and `definition`
 func SetSettingDefinition(name SettingName, definition SettingDefinition) {
 	settingDefinitionsLock.Lock()
 	defer settingDefinitionsLock.Unlock()
--- a/upgrade/util/util.go
+++ b/upgrade/util/util.go
@@ -164,6 +164,12 @@ func CreateOrUpdateLonghornVersionSetting(namespace string, lhClient *lhclientse
 
 	if s.Value != meta.Version {
 		s.Value = meta.Version
+		s.Annotations[types.GetLonghornLabelKey(types.UpdateSettingFromLonghorn)] = ""
+		s, err = lhClient.LonghornV1beta2().Settings(namespace).Update(context.TODO(), s, metav1.UpdateOptions{})
+		if err != nil {
+			return err
+		}
+		delete(s.Annotations, types.GetLonghornLabelKey(types.UpdateSettingFromLonghorn))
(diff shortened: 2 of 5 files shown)


innobead (changes requested):
Why not just use `GetSettingDefinition` and `SettingDefinition::ReadOnly` to determine whether the being updated setting is ready only or not? 

`annotation` is best effort. Also, you need to add annoation to different settings in different places, so it would make the code a little complicated.

mantissahz:
>Why not just use GetSettingDefinition and SettingDefinition::ReadOnly to determine whether the being updated setting is ready only or not?

There is an check statement here for [SettingDefinition::ReadOnly](https://github.com/longhorn/longhorn-manager/pull/2198/files#diff-8e6acb1a95cd72f122c6311ec6693f94330894a558ab93356c029a9f4e57f740R56)
I tried to temporarily change `SettingDefinition::ReadOnly` if it is updated by controllers or somewhere by Longhorn,
but when the object is updating and going through the validator,
it seems that `SettingDefinition` will be read again from types/settings.go by `GetSettingDefinition`

innobead:
Let me recheck this PR later.

types/setting.go
```diff
@@ -29,6 +29,12 @@ const (
 	ValueUnknown = "unknown"
 )
 
+const (
+	SettingUpdateFromAPI        = "api"
+	SettingUpdateFromController = "controller"
+	SettingUpdateFromUpgrade    = "upgrade"
+)
```
innobead, line 36: Probably no need, if we set the empty value to the annotation.

upgrade/util/util.go
```diff
@@ -164,6 +164,12 @@ func CreateOrUpdateLonghornVersionSetting(namespace string, lhClient *lhclientse
 
 	if s.Value != meta.Version {
 		s.Value = meta.Version
+		s.Annotations[types.GetLonghornLabelKey(types.UpdateSettingFromLonghorn)] = types.SettingUpdateFromUpgrade
+		s, err = lhClient.LonghornV1beta2().Settings(namespace).Update(context.TODO(), s, metav1.UpdateOptions{})
+		if err != nil {
+			return err
+		}
+		delete(s.Annotations, types.GetLonghornLabelKey(types.UpdateSettingFromLonghorn))
```
innobead, line 172: What's the purpose here?

webhook/resources/setting/validator.go
```diff
@@ -44,6 +44,19 @@ func (v *settingValidator) Create(request *admission.Request, newObj runtime.Obj
...
+	setting := newObj.(*longhorn.Setting)
+
+	settingDef, isExist := types.GetSettingDefinition(types.SettingName(setting.Name))
+	if !isExist {
+		return werror.NewInvalidError(fmt.Sprintf("setting %s does not exist", setting.Name), "metadata.name")
+	}
+	existingSetting := oldObj.(*longhorn.Setting)
+	_, isFromLHOld := existingSetting.Annotations[types.GetLonghornLabelKey(types.UpdateSettingFromLonghorn)]
+	_, isFromLH := setting.Annotations[types.GetLonghornLabelKey(types.UpdateSettingFromLonghorn)]
+	if settingDef.ReadOnly && !isFromLHOld && !isFromLH {
```
innobead, line 56: Why do we need to check the annotation of the old setting? If the new update has an annotation, it simply means we need to update it, right?

types/setting.go
mantissahz: Removed. Thanks.

mantissahz:
@mergify backport v1.5.x

mantissahz:
@mergify backport v1.4.x

innobead (approved):
LGTM

PR #2211, opened 2023-10-10
Improve log level for reconcilidation error

PhanLe1010:
If the error is API conflict error, the level should be debug

longhorn/longhorn#6848


james-munson (approved):
That looks fine.

innobead:

PR #2221, opened 2023-10-12
6581 remove dummy services

ejweber:
longhorn/longhorn#6581

We used to deploy the CSI components in StatefulSets, which require a service (even though in our case the service was nonfunctional). Since we are now using Deployments, do not deploy dummy services and clean up the dummy services that exist.

I thought about handling the cleanup of erroneous services in the driver deployer instead, but it's cleaner to remove all references from csi package in this PR and delete the services one time while walking the upgrade path.

There are some reports of these services not cleaning up nicely during driver upgrades in the issues linked to longhorn/longhorn#6581. Since the services are completely nonfunctional, I wrote the upgrade path to do a best effort background deletion. In the worst case, a dummy service will just stick around doing nothing until it is manually deleted or deleted with the longhorn-system namespace.

### DIFF
--- a/controller/uninstall_controller.go
+++ b/controller/uninstall_controller.go
@@ -1017,36 +1017,6 @@ func (c *UninstallController) deleteDriver() (bool, error) {
 		wait = true
 	}
 
-	servicesToClean := []string{
-		types.CSIAttacherName,
-		types.CSIProvisionerName,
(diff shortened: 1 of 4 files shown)


shuo-wu (approved):

innobead (approved):

innobead:

PR #2228, opened 2023-10-17
Suppress nagging error logging.

james-munson:
https://github.com/longhorn/longhorn/issues/6812

Removes all the "level=error" logging from longhorn-manager, but leaves the one kubernetes-layer log.  The only thing logged every hour will be 
```
[longhorn-manager-rjzn8] E1017 21:53:38.219480       1 setting_controller.go:205] failed to sync setting for longhorn-system/v2-data-engine: Current state prevents this: cannot apply v2-data-engine setting to Longhorn workloads when there are attached volumes
[longhorn-manager-rjzn8] E1017 21:53:38.219766       1 setting_controller.go:205] failed to sync setting for longhorn-system/storage-network: Current state prevents this: failed to apply storage-network setting to Longhorn workloads when there are attached volumes
```

### DIFF
--- a/controller/setting_controller.go
+++ b/controller/setting_controller.go
@@ -665,7 +665,7 @@ func (sc *SettingController) updateCNI() error {
 	}
 
 	if !volumesDetached {
-		return errors.Errorf("failed to apply %v setting to Longhorn workloads when there are attached volumes", types.SettingNameStorageNetwork)
+		return &types.ErrorInvalidState{Reason: fmt.Sprintf("failed to apply %v setting to Longhorn workloads when there are attached volumes", types.SettingNameStorageNetwork)}
 	}
 
 	nadAnnot := string(types.CNIAnnotationNetworks)
--- a/controller/utils.go
+++ b/controller/utils.go
@@ -62,6 +62,7 @@ func createOrUpdateAttachmentTicket(va *longhorn.VolumeAttachment, ticketID, nod
 
 func handleReconcileErrorLogging(logger logrus.FieldLogger, err error, mesg string) {
 	if types.ErrorIsInvalidState(err) {
+		logger.WithError(err).Trace(mesg)
 		return
 	}
 
--- a/controller/utils.go
+++ b/controller/utils.go
@@ -61,6 +61,10 @@ func createOrUpdateAttachmentTicket(va *longhorn.VolumeAttachment, ticketID, nod
 }
 
 func handleReconcileErrorLogging(logger logrus.FieldLogger, err error, mesg string) {
+	if types.ErrorIsInvalidState(err) {
+		return
+	}
+
 	if apierrors.IsConflict(err) {
 		logger.WithError(err).Debug(mesg)
 	} else {
--- a/datastore/longhorn.go
+++ b/datastore/longhorn.go
@@ -305,24 +305,20 @@ func (s *DataStore) ValidateSetting(name, value string) (err error) {
 			}
 		}
 	case types.SettingNameTaintToleration:
-		list, err := s.ListVolumesRO()
+		volumesDetached, err := s.AreAllVolumesDetachedState()
 		if err != nil {
 			return errors.Wrapf(err, "failed to list volumes before modifying toleration setting")
 		}
-		for _, v := range list {
-			if v.Status.State != longhorn.VolumeStateDetached {
-				return fmt.Errorf("cannot modify toleration setting before all volumes are detached")
-			}
+		if !volumesDetached {
+			return &types.ErrorInvalidState{Reason: fmt.Sprintf("cannot modify toleration setting before all volumes are detached")}
 		}
 	case types.SettingNameSystemManagedComponentsNodeSelector:
-		list, err := s.ListVolumesRO()
+		volumesDetached, err := s.AreAllVolumesDetachedState()
 		if err != nil {
 			return errors.Wrapf(err, "failed to list volumes before modifying node selector for managed components setting")
 		}
-		for _, v := range list {
-			if v.Status.State != longhorn.VolumeStateDetached {
-				return fmt.Errorf("cannot modify node selector for managed components setting before all volumes are detached")
-			}
+		if !volumesDetached {
+			return &types.ErrorInvalidState{Reason: fmt.Sprintf("cannot modify node selector for managed components setting before all volumes are detached")}
 		}
 	case types.SettingNamePriorityClass:
 		if value != "" {
@@ -330,14 +326,12 @@ func (s *DataStore) ValidateSetting(name, value string) (err error) {
 				return errors.Wrapf(err, "failed to get priority class %v before modifying priority class setting", value)
 			}
 		}
-		list, err := s.ListVolumesRO()
+		volumesDetached, err := s.AreAllVolumesDetachedState()
 		if err != nil {
 			return errors.Wrapf(err, "failed to list volumes before modifying priority class setting")
 		}
-		for _, v := range list {
-			if v.Status.State != longhorn.VolumeStateDetached {
-				return fmt.Errorf("cannot modify priority class setting before all volumes are detached")
-			}
+		if !volumesDetached {
+			return &types.ErrorInvalidState{Reason: fmt.Sprintf("cannot modify priority class setting before all volumes are detached")}
 		}
 	case types.SettingNameGuaranteedInstanceManagerCPU:
 		guaranteedInstanceManagerCPU, err := s.GetSetting(types.SettingNameGuaranteedInstanceManagerCPU)
@@ -356,7 +350,7 @@ func (s *DataStore) ValidateSetting(name, value string) (err error) {
 			return errors.Wrapf(err, "failed to check volume detachment for %v setting update", name)
 		}
 		if !volumesDetached {
-			return errors.Errorf("cannot apply %v setting to Longhorn workloads when there are attached volumes", name)
+			return &types.ErrorInvalidState{Reason: fmt.Sprintf("cannot apply %v setting to Longhorn workloads when there are attached volumes", name)}
 		}
 	case types.SettingNameV2DataEngine:
 		old, err := s.GetSetting(types.SettingNameV2DataEngine)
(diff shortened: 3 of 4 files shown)


types/types.go
```diff
@@ -273,6 +273,14 @@ func (e *NotFoundError) Error() string {
 	return fmt.Sprintf("cannot find %v", e.Name)
 }
 
+type ErrorNotNow struct {
+	Reason string
+}
+
+func (e *ErrorNotNow) Error() string {
+	return fmt.Sprintf("Current state prevents this: %v", e.Reason)
```
c3y1huang, line 281: nit:
1. `Error` might be a bit broad as looks like it also wraps state errors around the reason. Maybe we can use a more specific method name like `ErrorInvalidState`. 
1. Consider using lowercase text for better readability.
```
[longhorn-manager-rjzn8] E1017 21:53:38.219480       1 setting_controller.go:205] failed to sync setting for longhorn-system/v2-data-engine: current state prevents this: cannot apply v2-data-engine setting to Longhorn workloads when there are attached volumes
```

types/types.go
```diff
@@ -273,6 +273,14 @@ func (e *NotFoundError) Error() string {
 	return fmt.Sprintf("cannot find %v", e.Name)
 }
 
+type ErrorNotNow struct {
```
c3y1huang, line 276: Suggest adding some comments to this.

c3y1huang (changes requested):
Just some minor suggestions. Also, might want to consider squashing the commits together. The one saying `Remove debug logging` doesn't seem to make meaningful changes.

Besides that, everything else lgtm.

types/types.go
```diff
@@ -706,6 +714,11 @@ func ErrorAlreadyExists(err error) bool {
 	return strings.Contains(err.Error(), "already exists")
 }
 
+func ErrorIsNotNow(err error) bool {
+	var dummy *ErrorNotNow
+	return errors.As(err, &dummy)
+}
+
```
PhanLe1010, line 721: This casting logic is cool. I am wondering if you have any documents to learn more about this?

types/types.go
PhanLe1010: Is this correct that this method is a dummy method exists only to make the ErrorNotNow an interface?

controller/setting_controller.go
```diff
@@ -665,7 +665,7 @@ func (sc *SettingController) updateCNI() error {
 	}
 
 	if !volumesDetached {
-		return errors.Errorf("failed to apply %v setting to Longhorn workloads when there are attached volumes", types.SettingNameStorageNetwork)
+		return &types.ErrorNotNow{Reason: fmt.Sprintf("failed to apply %v setting to Longhorn workloads when there are attached volumes", types.SettingNameStorageNetwork)}
```
PhanLe1010, line 668: This means that this error will never be printed so the users might not understand why the new setting doesn't get applied?

datastore/longhorn.go
```diff
@@ -383,7 +377,7 @@ func (s *DataStore) ValidateV2DataEngine(v2DataEngineEnabled bool) error {
 		return errors.Wrapf(err, "failed to check volume detachment for %v setting update", types.SettingNameV2DataEngine)
 	}
 	if !volumesDetached {
-		return errors.Errorf("cannot apply %v setting to Longhorn workloads when there are attached volumes", types.SettingNameV2DataEngine)
+		return &types.ErrorNotNow{Reason: fmt.Sprintf("cannot apply %v setting to Longhorn workloads when there are attached volumes", types.SettingNameV2DataEngine)}
```
PhanLe1010, line 380: Looks like there is an inconsistency between this case (returned `&types.ErrorNotNow`) and other case (return `errors.Wrapf`). Could you elaborate a bit more?

types/types.go
ejweber: This method exists to satisfy the error interface. The only requirement for Go to treat a type as if it in an error (i.e. allow a function to return that type when an error return is called for) is for that type to implement the Error() method.

It is not a dummy method. Any time string formatting is used to print the error object, Error() will be invoked.

https://pkg.go.dev/fmt

> 4. If an operand implements the error interface, the Error method will be invoked to convert the object to a string, which will then be formatted as required by the verb (if any).

controller/setting_controller.go
james-munson: First, the setting is not necessarily new.  Second, this is just making the error type distinguishable from any other error, so we can decide to handle it differently when it is returned.  But here we are just instantiating it.

types/types.go
james-munson: Right, this is just to make an error type that can be distinguished from other errors.  I would rather not call it ErrorInvalidState, because that's a bit ambiguous.  It's not the state that is invalid, it is the action we are thinking about taking in the current state.  I could go for ErrorNotInCurrentState, for instance.

types/types.go
james-munson: This came from the docs for github.com/pkg/errors (not to be confused with the golang standard library errors package, which has superseded it, but is still compatible.)
Yeah, it's pretty cool, and better than the first thing I tried, which was a straight cast: err.(*ErrorNotNow).  That doesn't work for wrapped errors, but errors.As() will unwrap and check at each level.

types/types.go
james-munson: OK.

datastore/longhorn.go
james-munson: I agree.  The case for setting TaintToleration is similar, for instance.  But it doesn't litter the logs like the two I changed.

james-munson:
As I remarked in the defect, every hour there is a pile of error-level logging about inability to set storage-network and v2 engine.  This happens on just any quiet cluster, even if *no* setting has been changed, much less either of those two.  The logging comes from this method, which appears in every controller:
```go
func (sc *SettingController) handleErr(err error, key interface{}) {
    if err == nil {
        sc.queue.Forget(key)
        return
    }

    log := sc.logger.WithField("Setting", key)
    if sc.queue.NumRequeues(key) < maxRetries {
        handleReconcileErrorLogging(log, err, "Failed to sync Longhorn setting")
        sc.queue.AddRateLimited(key)
        return
    }

    utilruntime.HandleError(err)
    handleReconcileErrorLogging(log, err, "Dropping Longhorn setting out of the queue")
    sc.queue.Forget(key)
}
```
So it will log until maxRetries and then logs again that it is giving up.

So this change makes the particular error recognizable, then just suppresses its logging in handleReconcileErrorLogging (just as "isConflict" errors are logged at debug rather than error.)  Note that it is still logged by the call to utilruntime.HandleError.

That improves things, but there are still several things I don't understand:
- Why is there enqueued a validation of the current storage-network and v2 engine setting once an hour, whether the setting has been changed or not?
- If that is necessary, why is there no such check for other settings that cannot be changed with attached volumes, such as TaintToleration?
- Why are there two methods for checking all volumes detached in datastore/longhorn.go?  (One of which I pulled into its own function, AreAllVolumesDetachedState).

ejweber:
> Why is there enqueued a validation of the current storage-network and v2 engine setting once an hour, whether the setting has been changed or not?

This is the only one I can answer off the top of my head. The setting controller uses a resync period of one hour. No matter what, all settings will be reconciled at that interval. In a perfectly written controller, this would probably not be necessary, but most Kubernetes adjacent projects use a resync period in case there is some broken queuing logic introduced that would prevent a controller from ever syncing an object again.

https://github.com/longhorn/longhorn-manager/blob/286e6e600ec9e47fec3be0e527f60f0ecc4cb4af/controller/setting_controller.go#L51

types/types.go
PhanLe1010: Thanks James and Eric

types/types.go
PhanLe1010: Cool. Thank you

controller/setting_controller.go
PhanLe1010: Talked to @james-munson , Looks like the function `utilruntime.HandleError(err)` still print out the error so it is good to me

PhanLe1010:
> Why is there enqueued a validation of the current storage-network and v2 engine setting once an hour, whether the setting has been changed or not?

Yeah, the resync period of the setting informer is 1h

> If that is necessary, why is there no such check for other settings that cannot be changed with attached volumes, such as TaintToleration?


Looks like an inconsistency between the way to handle new settings (SettingNameStorageNetwork, SettingNameV2DataEngine) which check for whether volumes are detached and old settings (SettingNameTaintToleration,...) which doesn't perform the check 


> Why are there two methods for checking all volumes detached in datastore/longhorn.go? (One of which I pulled into its own function, AreAllVolumesDetachedState).


Good question. the `AreAllVolumesDetached` one is getting information from the instance manager. Perhaps this one is more accurate when determining whether we can allow drains. The other one, `AreAllVolumesDetachedState` fetches info directly from volume CRs. Maybe we should investigate if we can consolidate these 2 functions.

datastore/longhorn.go
james-munson: Made all "not with attached volumes" cases return this specialized error type.

types/types.go
james-munson: Renamed error type to ErrorNotWithAttached

types/types.go
```diff
@@ -273,6 +273,14 @@ func (e *NotFoundError) Error() string {
 	return fmt.Sprintf("cannot find %v", e.Name)
 }
 
+type ErrorNotWithAttached struct {
+	Reason string
+}
+
+func (e *ErrorNotWithAttached) Error() string {
+	return fmt.Sprintf("current attached state prevents this: %v", e.Reason)
+}
```
c3y1huang, line 282: Suggest making this more generic, as there is already a specific reason provided.

```suggestion
func (e *ErrorInvalidState) Error() string {
	return fmt.Sprintf("invalid state: %v", e.Reason)
}
```

james-munson:
Name changed.  Let's commit this.

controller/utils.go
```diff
@@ -61,6 +61,10 @@ func createOrUpdateAttachmentTicket(va *longhorn.VolumeAttachment, ticketID, nod
 }
 
 func handleReconcileErrorLogging(logger logrus.FieldLogger, err error, mesg string) {
+	if types.ErrorIsInvalidState(err) {
+		return
```
innobead, line 65: Suggest still having a log but with a lower level like trace instead of info or debug. This could give us a chance to know why updating those settings, but they haven't taken effect in any error situation.

innobead (changes requested):
LGTM, just one comment.

innobead:

controller/utils.go
james-munson: As you wish.  As mentioned above, there is still logging, from the call to utilruntime.HandleError(err) in SettingController.handleErr, so it is not entirely silent.
We will be back to the 10 lines of logging instead of just two every hour, but at least they will be trace.

innobead: yes, trace, so it will not be printed out by default. This is more for troubleshooting purposes.

james-munson: Sure.

innobead: Thanks, @c3y1huang already updated this PR with this change.

james-munson (commented):
That looks fine, thanks.

innobead (approved):

PR #2229, opened 2023-10-18
fix: CVE issues

c3y1huang:
https://github.com/longhorn/security/issues/16


innobead (approved):
There are no any deprecated or removed APIs from 1.28, so updating to use 0.28 related Kubernetes libs is fine. 

ref: https://kubernetes.io/docs/reference/using-api/deprecation-guide/

P.S. it seems we upgraded Go version, so please also upgrade the Go version of other components. 

These changes will not be backported, but rather only in 1.6.0. Technically, we only backport patch version of depenednecies to previous supported release branches. If any minor dependencies need to upgrade, need to have more detailed justification and analysis.

c3y1huang:
ref https://github.com/longhorn/longhorn/issues/7001

c3y1huang:

PR #2231, opened 2023-10-19
feat(log): add function name and line number in log message

ChanYiLin:
ref: [longhorn/longhorn#5509](https://github.com/longhorn/longhorn/issues/5509)

result:
Screenshot from 2023-10-19 16-12-30

### DIFF
--- a/main.go
+++ b/main.go
@@ -3,6 +3,8 @@ package main
 import (
 	"fmt"
 	"os"
+	"path"
+	"runtime"
 
 	"github.com/sirupsen/logrus"
 	"github.com/urfave/cli"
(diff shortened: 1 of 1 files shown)


innobead:
This is good, but having the line number seems a little tedious after reviewing. I think having package and func/method name could be good enough. 

P.S. Having line info is more helpful when hitting a panic and it is already the default behavior in Go, so we don't need to do anything on this.

innobead (changes requested):
For my above comment, WDYT?

PhanLe1010:
I think having line number may be useful in case we have similar logs in the same function. Also, this makes looking up the code more accurately. Kubernetes upstream is also doing this.

PhanLe1010 (approved):
LGTM

shuo-wu (approved):

innobead:
This is helpful info!

> Also, this makes looking up the code more accurately. Kubernetes upstream is also doing this.

innobead (approved):

innobead:
Let's backport this.

ChanYiLin:

PR #2232, opened 2023-10-19
Verify replica is closed by checking both spec.DesireState and status.CurrentState

derekbit:
To avoid running into a case that status.CurrentState is stopped but spec.DesireState is not stopped. verifyVolumeDependentResourcesClosed() needs to check status.CurrentState as well as spec.DesireState.

Longhorn/longhorn#6776


james-munson:
I agree that this seems safer, to conclude that a volume's status of stopped (engine and replicas) is not just transient, but matches the desired state.  I'm not sure yet how that helps address https://github.com/longhorn/longhorn/issues/6776, though.

Also, why wouldn't the same logic apply in areVolumeDependentResourcesOpened(), say?  Shouldn't we check that the engine and the (at least one) running replica are desired to be in state "running"?

derekbit:
@james-munson 
The draft might be incorrect for https://github.com/longhorn/longhorn/issues/6776. Please ignore it.

PR #2237, opened 2023-10-23
Eliminate multiple client creations

derekbit:
Longhorn/longhorn#6866
Longhorn/longhorn#6936

### DIFF
--- a/util/client/client.go
+++ b/util/client/client.go
@@ -1,75 +1,111 @@
 package client
 
 import (
-	"context"
 	"fmt"
+	"os"
 	"time"
 
 	"github.com/pkg/errors"
-	"github.com/rancher/wrangler/pkg/clients"
-	"github.com/rancher/wrangler/pkg/schemes"
+	wranglerClients "github.com/rancher/wrangler/pkg/clients"
+	wranglerSchemes "github.com/rancher/wrangler/pkg/schemes"
+	"github.com/sirupsen/logrus"
 
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/informers"
-	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
 
 	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
 	apiextensionsclientset "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
-	clientset "k8s.io/client-go/kubernetes"
+	metricsclientset "k8s.io/metrics/pkg/client/clientset/versioned"
 
 	"github.com/longhorn/longhorn-manager/datastore"
-
+	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
 	lhclientset "github.com/longhorn/longhorn-manager/k8s/pkg/client/clientset/versioned"
 	lhinformers "github.com/longhorn/longhorn-manager/k8s/pkg/client/informers/externalversions"
+	"github.com/longhorn/longhorn-manager/types"
 )
 
-type Client struct {
-	clients.Clients
-	Datastore *datastore.DataStore
+type Clients struct {
+	wranglerClients.Clients
+	MetricsClient *metricsclientset.Clientset
+	Scheme        *runtime.Scheme
+	Namespace     string
+	Datastore     *datastore.DataStore
+	StopCh        <-chan struct{}
 }
 
-func NewClient(ctx context.Context, config *rest.Config, namespace string, needDataStore bool) (*Client, error) {
-	if err := schemes.Register(appsv1.AddToScheme); err != nil {
-		return nil, err
+func NewClients(kubeconfigPath string, stopCh <-chan struct{}) (*Clients, error) {
+	namespace := os.Getenv(types.EnvPodNamespace)
+	if namespace == "" {
+		logrus.Warnf("Cannot detect pod namespace, environment variable %v is missing, using default namespace", types.EnvPodNamespace)
+		namespace = corev1.NamespaceDefault
 	}
 
-	clients, err := clients.NewFromConfig(config, nil)
+	config, err := clientcmd.BuildConfigFromFlags("", kubeconfigPath)
 	if err != nil {
+		return nil, errors.Wrap(err, "unable to get client config")
+	}
+
+	config.Burst = 100
+	config.QPS = 50
+
+	if err := wranglerSchemes.Register(appsv1.AddToScheme); err != nil {
 		return nil, err
 	}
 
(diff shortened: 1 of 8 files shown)


derekbit:
The memory usage is reduced by 50% (800MiB->400MiB) in the 1000 RWO volumes test.

innobead (approved):
LGTM.

util/client/client.go
```diff
@@ -1,75 +1,111 @@
...
+	lhClient, err := lhclientset.NewForConfig(config)
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to get clientset")
+	}
 
-		kubeInformerFactory := informers.NewSharedInformerFactory(kubeClient, time.Second*30)
-		lhInformerFactory := lhinformers.NewSharedInformerFactory(lhClient, time.Second*30)
+	scheme := runtime.NewScheme()
+	if err := longhorn.SchemeBuilder.AddToScheme(scheme); err != nil {
+		return nil, errors.Wrap(err, "unable to create scheme")
```
derekbit, line 71: One question
Is it ok that call `AddToScheme` twice here?
cc @innobead @shuo-wu

webhook/server/server.go
```diff
@@ -47,31 +45,24 @@ var (
 
 type WebhookServer struct {
 	context     context.Context
-	cfg         *rest.Config
 	namespace   string
 	webhookType string
 }
 
-func New(ctx context.Context, cfg *rest.Config, namespace, webhookType string) *WebhookServer {
+func New(ctx context.Context, namespace, webhookType string) *WebhookServer {
```
shuo-wu, line 52: NIT: Why not cache `clients` in this `WebhookServer`? Then the below methods do not need input `clients *client.Clients`

shuo-wu (approved):
LGTM

webhook/server/server.go
derekbit: Updated. Thank you.

util/client/client.go
ChanYiLin: I think is okay to call it twice
AddToScheme function should be idempotent

util/client/client.go
```diff
@@ -1,75 +1,111 @@
...
+	wranglerSchemes "github.com/rancher/wrangler/pkg/schemes"
+	"github.com/sirupsen/logrus"
 
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/informers"
-	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
 
-	corev1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
```
innobead, line 17: Just noticed the previous is `corev1 "k8s.io/api/apps/v1"`, so is there any typo before?

innobead: cc @c3y1huang is this caused by a wrong change by a previous format update?

ChanYiLin: I think both 
"k8s.io/api/core/v1", "k8s.io/api/apps/v1" have `AddToScheme` function and they are actually pointing to the same place.
So both works fine, but in this PR, we need to use `corev1.NamespaceDefault` which  only in "k8s.io/api/core/v1" library

ChanYiLin (approved):
LGTM

webhook/server/server.go
```diff
@@ -121,9 +109,9 @@ func (s *WebhookServer) ListenAndServe() error {
 	}
 }
 
-func (s *WebhookServer) runAdmissionWebhookListenAndServe(client *client.Client, handler http.Handler, validationResources []admission.Resource, mutationResources []admission.Resource) error {
-	apply := client.Apply.WithDynamicLookup()
-	client.Core.Secret().OnChange(s.context, "secrets", func(key string, secret *corev1.Secret) (*corev1.Secret, error) {
+func (s *WebhookServer) runAdmissionWebhookListenAndServe(clients *client.Clients, handler http.Handler, validationResources []admission.Resource, mutationResources []admission.Resource) error {
```
shuo-wu, line 112: NIT: No need to input `clients *client.Clients`. Same below

util/client/client.go
innobead: It's different. One is for `app` workload related resource scheme but `core` is for basic resources like the pod below. 

[/vendor/k8s.io/api/core/v1/register.go#L36-L49](https://github.com/longhorn/longhorn-manager/blob/a6a15ae9f4546f796bd4ebdc31851536d11a8c8b/vendor/k8s.io/api/core/v1/register.go#L36-L49)
```go
var (
	// We only register manually written functions here. The registration of the
	// generated functions takes place in the generated files. The separation
	// makes the code compile even when the generated files are missing.
	SchemeBuilder = runtime.NewSchemeBuilder(addKnownTypes)
	AddToScheme   = SchemeBuilder.AddToScheme
)

// Adds the list of known types to the given scheme.
func addKnownTypes(scheme *runtime.Scheme) error {
	scheme.AddKnownTypes(SchemeGroupVersion,
		&Pod{},
		&PodList{},
		&PodStatusResult{},
```

c3y1huang: > cc @c3y1huang is this caused by a wrong change by a previous format update?

It was previously [v1 k8s.io/api/apps/v1](https://github.com/longhorn/longhorn-manager/commit/c93aea9de7ed5e02feafcd70d3888d75742571ba#diff-7b3f4aabf038dc0bcb882dac26677c633b8cbb05ee4b57cfcbff61ea5051ec56L12).

innobead: I see. Let's fix this if it has been backported.

c3y1huang: https://github.com/longhorn/longhorn-manager/pull/2238

innobead: It's confusing, that we were using is core resource (secret below) but used the app scheme.

[/webhook/server/server.go#L126-L131](https://github.com/longhorn/longhorn-manager/blob/a6a15ae9f4546f796bd4ebdc31851536d11a8c8b/webhook/server/server.go#L126-L131)
```go
	client.Core.Secret().OnChange(s.context, "secrets", func(key string, secret *corev1.Secret) (*corev1.Secret, error) {
		if secret == nil || secret.Name != caName || secret.Namespace != s.namespace || len(secret.Data[corev1.TLSCertKey]) == 0 {
			return nil, nil
		}

		port := int32(types.DefaultWebhookServerPort)

```

webhook/server/server.go
derekbit: Updated. Thank you.

util/client/client.go
derekbit: Fixed. Thank you.

innobead (approved):
LGTM.

innobead:
@mergify backport v1.5.x

PR #2242, opened 2023-10-24
Avoid unnecessary object deep copy by using read-only get/list functions in controllers

shuo-wu:
https://github.com/longhorn/longhorn/issues/6866

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -1923,7 +1923,7 @@ func (c *VolumeController) canInstanceManagerLaunchReplica(r *longhorn.Replica) 
 	if r.Status.InstanceManagerName != "" {
 		return true, nil
 	}
-	defaultIM, err := c.ds.GetInstanceManagerByInstance(r)
+	defaultIM, err := c.ds.GetInstanceManagerByInstanceRO(r)
 	if err != nil {
 		return false, errors.Wrapf(err, "failed to find instance manager for replica %v", r.Name)
 	}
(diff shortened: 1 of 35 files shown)


innobead:
@shuo-wu PR check failed, please resolve that.

innobead:
@shuo-wu is this ready to review?

shuo-wu:
Yes it's ready to review. But I need to do some tests before the merge.

innobead:
> Yes it's ready to review. But I need to do some tests before the merge.

Sounds good. See if we have a way to see the improvement.

controller/volume_controller.go
```diff
@@ -2600,12 +2604,12 @@ func (c *VolumeController) getNodeCandidatesForAutoBalanceZone(v *longhorn.Volum
 		return candidateNames
 	}
 
-	readyNodes, err := c.ds.ListReadyAndSchedulableNodes()
+	readyNodes, err := c.ds.ListReadyAndSchedulableNodesRO()
```
derekbit, line 2607: What's the difference between `ListReadyAndSchedulableNodesRO` and `ListReadyAndSchedulableNodes` for `delete(readyNodes, nName)`?

shuo-wu: Here, the operations like `delete(readyNodes, nName)` do not include object updating over `readyNodes`. Hence, there is no need to do deep copy during the node list

derekbit: Got it. Thank you.

derekbit (dismissed):

shuo-wu:
Core tests passed
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/5200/

innobead:
This will be only introduced to master and 1.5.4+.

innobead (approved):
In general, LGTM.

innobead:
@mergify backport v1.5.x

PR #2243, opened 2023-10-24
Filter Kubernetes resources by longhorn-system namespace

derekbit:
When a Kubernetes cluster contains numerous secrets, configmaps, and other Kubernetes resources, the synchronization of the cache can lead to longhorn-manager pod's high memory consumption. This ticket focuses on optimizing the resource cache to prevent high memory usage caused by the cache synchronization. For instance, secret is used for the backup target and share manager, so caching all secrets doesn't make sense. We need to avoid caching them.

Longhorn/longhorn#6954

### DIFF
--- a/app/system_rollout.go
+++ b/app/system_rollout.go
@@ -88,9 +88,10 @@ func systemRollout(c *cli.Context) error {
 	}
 
 	kubeInformerFactory := informers.NewSharedInformerFactory(kubeClient, time.Second*30)
+	kubeFilteredInformerFactory := informers.NewSharedInformerFactoryWithOptions(kubeClient, time.Second*30, informers.WithNamespace(namespace))
 	lhInformerFactory := lhinformers.NewSharedInformerFactory(lhClient, time.Second*30)
 
-	ds := datastore.NewDataStore(lhInformerFactory, lhClient, kubeInformerFactory, kubeClient, extensionsClient, namespace)
+	ds := datastore.NewDataStore(lhInformerFactory, lhClient, kubeInformerFactory, kubeFilteredInformerFactory, kubeClient, extensionsClient, namespace)
 
 	logger := logrus.StandardLogger()
 	logrus.SetLevel(logrus.DebugLevel)
@@ -98,6 +99,7 @@ func systemRollout(c *cli.Context) error {
 	doneCh := make(chan struct{})
 	go lhInformerFactory.Start(doneCh)
 	go kubeInformerFactory.Start(doneCh)
+	go kubeFilteredInformerFactory.Start(doneCh)
 
 	ctrl := controller.NewSystemRolloutController(
 		systemRestoreName,
--- a/app/uninstall.go
+++ b/app/uninstall.go
@@ -79,9 +79,10 @@ func uninstall(c *cli.Context) error {
 	}
 
 	kubeInformerFactory := informers.NewSharedInformerFactory(kubeClient, time.Second*30)
+	kubeFilteredInformerFactory := informers.NewSharedInformerFactoryWithOptions(kubeClient, time.Second*30, informers.WithNamespace(namespace))
 	lhInformerFactory := lhinformers.NewSharedInformerFactory(lhClient, time.Second*30)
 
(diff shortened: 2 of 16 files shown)


innobead:
PR check failed :thinking:

innobead (approved):
LGTM, but need to figure out why the PR check failed.

derekbit:
> PR check failed 🤔

Some parameters in unit tests were not updated.

innobead (approved):

app/system_rollout.go
```diff
@@ -88,9 +88,10 @@ func systemRollout(c *cli.Context) error {
 	}
 
 	kubeInformerFactory := informers.NewSharedInformerFactory(kubeClient, time.Second*30)
+	kubeFilteredInformerFactory := informers.NewFilteredSharedInformerFactory(kubeClient, time.Second*30, namespace, nil)
```
PhanLe1010, line 91: Looks like we forgot to start kubeInformerFactory
```
go kubeInformerFactory.Start(doneCh)
```

PhanLe1010: Also, looks like `NewFilteredSharedInformerFactory` is deprecated https://github.com/longhorn/longhorn-manager/blob/cf7b3e97346de9e677cf66081bf56cfcbb36d302/k8s/pkg/client/informers/externalversions/factory.go#L83-L89 Should we use [NewSharedInformerFactoryWithOptions](https://github.com/longhorn/longhorn-manager/blob/cf7b3e97346de9e677cf66081bf56cfcbb36d302/k8s/pkg/client/informers/externalversions/factory.go#L92-L93) instead

app/uninstall.go
```diff
@@ -79,9 +79,10 @@ func uninstall(c *cli.Context) error {
 	}
 
 	kubeInformerFactory := informers.NewSharedInformerFactory(kubeClient, time.Second*30)
+	kubeFilteredInformerFactory := informers.NewFilteredSharedInformerFactory(kubeClient, time.Second*30, namespace, nil)
 	lhInformerFactory := lhinformers.NewSharedInformerFactory(lhClient, time.Second*30)
 
-	ds := datastore.NewDataStore(lhInformerFactory, lhClient, kubeInformerFactory, kubeClient, extensionsClient, namespace)
+	ds := datastore.NewDataStore(lhInformerFactory, lhClient, kubeInformerFactory, kubeFilteredInformerFactory, kubeClient, extensionsClient, namespace)
```
PhanLe1010, line 86: Same as above:
* looks like NewFilteredSharedInformerFactory is deprecated
* Looks like we forgot to start kubeInformerFactory

app/system_rollout.go
derekbit: Yes, updated. Thank you.

app/uninstall.go
derekbit: Updated. Thank you.

util/client/client.go
```diff
@@ -90,11 +90,13 @@ func NewClients(kubeconfigPath string, stopCh <-chan struct{}) (*Clients, error)
 	//  if a specific controller requires a periodic resync, one enable it only for that informer, add a resync to the event handler, go routine, etc.
 	//  some refs to look at: https://github.com/kubernetes-sigs/controller-runtime/issues/521
 	kubeInformerFactory := informers.NewSharedInformerFactory(clients.K8s, time.Second*30)
+	kubeFilteredInformerFactory := informers.NewFilteredSharedInformerFactory(clients.K8s, time.Second*30, namespace, nil)
```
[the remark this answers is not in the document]
derekbit: Updated. Thank you.

innobead (approved):
LGTM. 

Is it possible to apply this filterd informer to longhorn resources as well? I think our resources are all namespaced?

app/uninstall.go
innobead: Good catch @PhanLe1010

PhanLe1010 (approved):
LGTM

innobead:
@mergify backport v1.5.x

innobead:
Backport to v1.4.x after 1.4.4 out.

PR #2248, opened 2023-10-25
fix(backup): AWS IAM Role Annotation

mantissahz:
AWS IAM Role Annotation will not be added to longhorn-manager pods if the secret is created first and then set the `backup-target` and `backup-target-credential-secret`.

Ref: longhorn/longhorn#6947

### DIFF
--- a/controller/setting_controller.go
+++ b/controller/setting_controller.go
@@ -359,6 +359,9 @@ func (sc *SettingController) syncBackupTarget() (err error) {
 			if _, err = sc.ds.UpdateBackupTarget(backupTarget); err != nil && !apierrors.IsConflict(errors.Cause(err)) {
 				sc.logger.WithError(err).Warn("Failed to update backup target")
 			}
+			if err = sc.handleSecretsForAWSIAMRoleAnnotation(backupTarget.Spec.BackupTargetURL, existingBackupTarget.Spec.CredentialSecret, secretSetting.Value, existingBackupTarget.Spec.BackupTargetURL != targetSetting.Value); err != nil {
+				sc.logger.WithError(err).Warn("Failed to update secrets for AWSIAMRoleAnnotation")
+			}
 		}
 	}()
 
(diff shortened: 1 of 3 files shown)


mantissahz:
@mergify backport v1.5.x

ChanYiLin (dismissed):
LGTM

controller/setting_controller.go
```diff
@@ -390,6 +395,85 @@ func (sc *SettingController) syncBackupTarget() (err error) {
 	return nil
 }
 
+func (sc *SettingController) handleSecretsForAWSIAMRoleAnnotation(backupTargetURL, oldSecretName, newSecretName string, isBackupTargetURLChanged bool) (err error) {
+	isSameSecretName := oldSecretName == newSecretName
+	if isSameSecretName && !isBackupTargetURLChanged {
+		return nil
+	}
+
+	isHavingAWSIAMRoleArn := false
```
c3y1huang, line 404: nit: consider using `isArnExists` instead to improve readability.

controller/setting_controller.go
```diff
@@ -390,6 +395,85 @@ func (sc *SettingController) syncBackupTarget() (err error) {
...
+	if !isSameSecretName {
+		isHavingAWSIAMRoleArn, _, err = sc.updateSecretForAWSIAMRoleAnnotation(backupTargetURL, oldSecretName, true)
+		if err != nil {
+			return err
+		}
+	}
+	_, isNewSecretNotExist, err := sc.updateSecretForAWSIAMRoleAnnotation(backupTargetURL, newSecretName, false)
```
c3y1huang, line 411: nit: consider using `isNewSecret` instead to improve readability.

controller/setting_controller.go
```diff
@@ -390,6 +395,85 @@ func (sc *SettingController) syncBackupTarget() (err error) {
...
+	// kubernetes_secret_controller will not reconcile the secret that does not exist such as named "", so we remove AWS IAM role annotation of pods if new secret name is "".
+	if isNewSecretNotExist && isHavingAWSIAMRoleArn {
+		if err = sc.removePodsAWSIAMRoleAnnotation(); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (sc *SettingController) updateSecretForAWSIAMRoleAnnotation(backupTargetURL, secretName string, isOldSecret bool) (bool, bool, error) {
```
c3y1huang, line 424: nit: `(bool, bool, error)` can be difficult to comprehend at first glance, consider assigning variables or adding a comment.

controller/setting_controller.go
```diff
@@ -390,6 +395,85 @@ func (sc *SettingController) syncBackupTarget() (err error) {
...
+	}
+	pods := append(managerPods, instanceManagerPods...)
+
+	for _, pod := range pods {
+		_, exist := pod.Annotations[types.AWSIAMRoleAnnotation]
+		if !exist {
+			continue
+		}
+		delete(pod.Annotations, types.AWSIAMRoleAnnotation)
+		sc.logger.Infof("Removing AWS IAM role for pod %v/%v", pod.Namespace, pod.Name)
```
c3y1huang, line 469: nit: do we need to log the pod.Namespace?

controller/setting_controller.go
```diff
@@ -390,6 +395,85 @@ func (sc *SettingController) syncBackupTarget() (err error) {
...
+		if _, err = sc.kubeClient.CoreV1().Pods(pod.Namespace).Update(context.TODO(), pod, metav1.UpdateOptions{}); err != nil {
```
c3y1huang, line 470: Should make this a reusable function.

controller/setting_controller.go
```diff
@@ -390,6 +395,85 @@ func (sc *SettingController) syncBackupTarget() (err error) {
...
+			return err
```
c3y1huang, line 471: Ignore when not found?

controller/setting_controller.go
mantissahz: Modified to `isArnExists`.Thanks

controller/setting_controller.go
mantissahz: Modified to `isValidSecret`. Thanks

controller/setting_controller.go
mantissahz: Added a comment and assigned variables. Thanks

controller/setting_controller.go
mantissahz: Removed to log the pod.Namespace.

controller/setting_controller.go
mantissahz: Made a reusable function. Thanks.

controller/setting_controller.go
mantissahz: Do you mean the pod is not found when updating this pod?
Ignore the not found error when updating the pod.
Thanks.

controller/setting_controller.go
```diff
@@ -390,6 +395,92 @@ func (sc *SettingController) syncBackupTarget() (err error) {
...
+	}
+
+	if isOldSecret {
+		delete(secret.Annotations, types.GetLonghornLabelKey(string(types.SettingNameBackupTarget)))
+		isArnExists = secret.Data[types.AWSIAMRoleArn] != nil
+	} else {
+		secret.Annotations[types.GetLonghornLabelKey(string(types.SettingNameBackupTarget))] = backupTargetURL
+	}
+
+	if _, err = sc.ds.UpdateSecret(sc.namespace, secret); err != nil {
```
c3y1huang, line 449: It appears that the secret could potentially be updated twice. What do you think about moving this out to the caller?

mantissahz: I believe you miss the `isSameSecretName` flag and it would not call `sc.updateSecretForAWSIAMRoleAnnotation` twice for the same secret name.

c3y1huang (dismissed):

mantissahz:
@mergify backport v1.4.x

controller/setting_controller.go
```diff
@@ -350,7 +350,9 @@ func (sc *SettingController) syncBackupTarget() (err error) {
 
 	existingBackupTarget := backupTarget.DeepCopy()
 	defer func() {
+		oldBackupTargetURL := backupTarget.Spec.BackupTargetURL
```
innobead, line 353: Use `existingBackupTarget.Spec.BackupTargetURL` instead. 

Also, there is no need to create another variable, because you can do below. 

```go
if err = sc.handleSecretsForAWSIAMRoleAnnotation(backupTarget.Spec.BackupTargetURL, existingBackupTarget.Spec.CredentialSecret, secretSetting.Value, existingBackupTarget.Spec.BackupTargetURL != targetSetting.Value); err != nil {
				sc.logger.WithError(err).Warn("Failed to update secrets for AWSIAMRoleAnnotation")
			}
```

innobead (changes requested):
In general, LGTM. 

Just one comment.

controller/setting_controller.go
mantissahz: Modified. Thanks.

innobead (approved):

innobead:

PR #2252, opened 2023-10-26
Refactor informer factories creation and initialization

derekbit:
- Refactor informer factories creation and initialization
- Filter pdb, daemonset and deployment by namespace

longhorn/longhorn#6954


innobead (approved):

innobead:
@mergify backport v1.5.x

shuo-wu (commented):
I am thinking if DaemonSet and Deployment can be filtered by namespace as well? The main blocker may be the rollout usage

innobead:
> I am thinking if DaemonSet and Deployment can be filtered by namespace as well? The main blocker may be the rollout usage

Rollout should only care about longhorn-system namespace only, so it should be fine. @c3y1huang ?

datastore/datastore.go
```diff
@@ -17,8 +16,8 @@ import (
 	storagelisters_v1 "k8s.io/client-go/listers/storage/v1"
 
 	lhclientset "github.com/longhorn/longhorn-manager/k8s/pkg/client/clientset/versioned"
-	lhinformers "github.com/longhorn/longhorn-manager/k8s/pkg/client/informers/externalversions"
 	lhlisters "github.com/longhorn/longhorn-manager/k8s/pkg/client/listers/longhorn/v1beta2"
+	"github.com/longhorn/longhorn-manager/util"
```
c3y1huang, line 20:
```suggestion
	"github.com/longhorn/longhorn-manager/util"

    lhclientset "github.com/longhorn/longhorn-manager/k8s/pkg/client/clientset/versioned"
	lhlisters "github.com/longhorn/longhorn-manager/k8s/pkg/client/listers/longhorn/v1beta2"
```

util/client/client.go
```diff
@@ -22,8 +21,12 @@ import (
 	"github.com/longhorn/longhorn-manager/datastore"
 	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
 	lhclientset "github.com/longhorn/longhorn-manager/k8s/pkg/client/clientset/versioned"
-	lhinformers "github.com/longhorn/longhorn-manager/k8s/pkg/client/informers/externalversions"
 	"github.com/longhorn/longhorn-manager/types"
+	"github.com/longhorn/longhorn-manager/util"
+)
+
+const (
+	defaultResyncPeriod = 30 * time.Second
```
c3y1huang, line 29: Where is this used?

shuo-wu:
OK we won't record the workload deployment/daemonset in the system backup. Namespace filtering seems fine. Let's wait for Chin-Ya's final confirmation

c3y1huang (commented):
System backup/restore doesn't handle the application, should be fine.

util/client/client.go
derekbit: Forgot to remove it. Thank you.

derekbit:
@shuo-wu @c3y1huang @shuo-wu 
Just checked datastore's pod CRUD. We can filter pod by longhorn-system namespace? Right?

shuo-wu:
> Just checked datastore's pod CRUD. We can filter pod by longhorn-system namespace? Right?

No. We will retrieve the workload status for volumes. And the workloads can be in any namespaces.

derekbit:
> > Just checked datastore's pod CRUD. We can filter pod by longhorn-system namespace? Right?
> 
> No. We will retrieve the workload status for volumes. And the workloads can be in any namespaces.

Got it. Thank you.

Ref:
the namespace is from `pv.Spec.ClaimRef.Namespace`.
https://github.com/longhorn/longhorn-manager/blob/master/controller/kubernetes_pv_controller.go#L427

datastore/datastore.go
derekbit: Updated. Thank you.

innobead (approved):

PR #2264, opened 2023-10-27
fix(setting): checking setting annotation is nil or not

mantissahz:
Checking the annotation of the `Settings` object is not nil before adding a new annotation for a Setting object.

Ref: longhorn/longhorn#6987, longhorn/longhorn#6947


mantissahz:

innobead (approved):

PR #2270, opened 2023-10-30
feat: pass support bundle issueURL and description

Yu-Jack:
## Problem

There is a issue https://github.com/harvester/harvester/issues/4630 in harvester. The description and issue url are empty from downloaded support bundle. And LH manager has the same issue here.

[Related downstream PR](https://github.com/rancher/support-bundle-kit/pull/81)

### DIFF
--- a/controller/support_bundle_controller.go
+++ b/controller/support_bundle_controller.go
@@ -699,6 +699,14 @@ func (c *SupportBundleController) newSupportBundleManager(supportBundle *longhor
 									Value: supportBundle.Name,
 								},
 								{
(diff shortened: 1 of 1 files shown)


innobead (approved):

innobead:
@c3y1huang see if we need to backport this, but need to update the bundle image version first.

c3y1huang:
> @c3y1huang see if we need to backport this, but need to update the bundle image version first.

Created issue for tracking: https://github.com/longhorn/longhorn/issues/6997

Yu-Jack:
Add tested result

Pasted_Image_2023_10_30__4_14 PM
image

PR #2275, opened 2023-10-30
Fix engine migration crash

ejweber:
longhorn/longhorn#6961

Just https://github.com/longhorn/longhorn-manager/commit/be6aca60ae375d212256d9bc3363f70bce86a767 is sufficient to prevent the engine from crashing under the circumstances from https://github.com/longhorn/longhorn/issues/6961#issuecomment-1785634706. However, this commit alone leads to orphaned data once the migration is complete (as the replica we refuse to add to the engine becomes inactive without an active counterpart).

I experimented with a path to making use of this newly inactive replica after the migration (to avoid unnecessary rebuilding), but I ultimately decided that the migration logic refactor this required was outside the scope of the ticket.

The other commits are an attempt to ensure we don't orphan the newly inactive replica's directory by providing a path to cleaning it up.

I need more time to test this because I'm running into issues with backing image download checksums. For some reason, the download of the backing image to the uncordoned node is consistently failing with checksum issues.

### DIFF
--- a/controller/replica_controller.go
+++ b/controller/replica_controller.go
@@ -303,11 +303,18 @@ func (rc *ReplicaController) syncReplica(key string) (err error) {
 			return errors.Wrapf(err, "failed to cleanup the related replica instance before deleting replica %v", replica.Name)
 		}
 
+		rs, err := rc.ds.ListReplicasByNodeRO(replica.Spec.NodeID)
+		if err != nil {
+			return errors.Wrapf(err, "failed to list replicas by node before deleting replica %v", replica.Name)
+		}
+
 		if replica.Spec.NodeID != "" && replica.Spec.NodeID != rc.controllerID {
 			log.Warn("Failed to cleanup replica's data because the replica's data is not on this node")
 		} else if replica.Spec.NodeID != "" {
 			if replica.Spec.BackendStoreDriver == longhorn.BackendStoreDriverTypeV1 {
-				if replica.Spec.Active && dataPath != "" {
+				// Clean up the data directory if this is the active replica or if this inactive replica is the only one
+				// using it.
+				if (replica.Spec.Active || !hasMatchingReplica(replica, rs)) && dataPath != "" {
 					// prevent accidentally deletion
 					if !strings.Contains(filepath.Base(filepath.Clean(dataPath)), "-") {
 						return fmt.Errorf("%v doesn't look like a replica data path", dataPath)
@@ -936,4 +943,13 @@ func (rc *ReplicaController) enqueueAllRebuildingReplicaOnCurrentNode() {
 
 func (rc *ReplicaController) isResponsibleFor(r *longhorn.Replica) bool {
 	return isControllerResponsibleFor(rc.controllerID, rc.ds, r.Name, r.Spec.NodeID, r.Status.OwnerID)
+}
+
+func hasMatchingReplica(replica *longhorn.Replica, replicas []*longhorn.Replica) bool {
+	for _, r := range replicas {
+		if r.Name != replica.Name && r.Spec.DataDirectoryName == replica.Spec.DataDirectoryName {
+			return true
+		}
+	}
+	return false
 }
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -1757,6 +1757,12 @@ func (c *VolumeController) openVolumeDependentResources(v *longhorn.Volume, e *l
 			log.WithField("replica", r.Name).Warn("Replica is running but Port is empty")
 			continue
 		}
+		if _, ok := e.Spec.ReplicaAddressMap[r.Name]; !ok && isVolumeMigrating(v) && e.Spec.NodeID == v.Spec.NodeID {
+			// The volume is migrating from this engine. Don't allow new replicas to be added until migration is
+			// complete per https://github.com/longhorn/longhorn/issues/6961.
+			log.WithField("replica", r.Name).Warn("Replica is running, but can't be added while migration is ongoing")
+			continue
+		}
 		replicaAddressMap[r.Name] = imutil.GetURL(r.Status.StorageIP, r.Status.Port)
 	}
 	if len(replicaAddressMap) == 0 {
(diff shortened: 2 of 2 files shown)


ejweber:
With this PR, the following modifications in steps 12 and 13 from https://github.com/longhorn/longhorn/issues/6961#issuecomment-1785634706 are observed:

```
# The pod never experiences an I/O error.
eweber@laptop:~/longhorn> k get pod
NAME       READY   STATUS    RESTARTS   AGE
test-pod   1/1     Running   0          10m

eweber@laptop:~/longhorn> k logs test-pod
1+0 records in
1+0 records out
...

# The migration process ISN'T confused by the state of the waiting replica.
[longhorn-manager-jrzpp] time="2023-10-31T18:27:35Z" level=warning msg="Running replica pvc-3428a4f1-463c-4f27-a536-06050ef6e828-r-fd71a95d wasn't added to engine, will ignore it and continue migration" func="controller.(*VolumeController).prepareReplicasAndEngineForMigration" file="volume_controller.go:4000" accessMode=rwx controller=longhorn-volume frontend=blockdev migratable=true migrationEngine=pvc-3428a4f1-463c-4f27-a536-06050ef6e828-e-1 migrationNodeID=eweber-v125-worker-e472db53-xjmzr node=eweber-v125-worker-e472db53-9kz5b owner=eweber-v125-worker-e472db53-9kz5b shareEndpoint= shareState= state=attached volume=pvc-3428a4f1-463c-4f27-a536-06050ef6e828
[longhorn-manager-jrzpp] time="2023-10-31T18:27:35Z" level=warning msg="Replica is running, but can't be added while migration is ongoing" func="controller.(*VolumeController).openVolumeDependentResources" file="volume_controller.go:1763" accessMode=rwx controller=longhorn-volume frontend=blockdev migratable=true node=eweber-v125-worker-e472db53-9kz5b owner=eweber-v125-worker-e472db53-9kz5b replica=pvc-3428a4f1-463c-4f27-a536-06050ef6e828-r-fd71a95d shareEndpoint= shareState= state=attached volume=pvc-3428a4f1-463c-4f27-a536-06050ef6e828

# The waiting replica is cleaned up when migration is complete.
[longhorn-manager-jrzpp] time="2023-10-31T18:28:30Z" level=info msg="Cleaning up replica" func="controller.(*ReplicaController).syncReplica" file="replica_controller.go:321" controller=longhorn-replica node=eweber-v125-worker-e472db53-9kz5b nodeID=eweber-v125-worker-e472db53-9kz5b ownerID=eweber-v125-worker-e472db53-9kz5b replica=pvc-3428a4f1-463c-4f27-a536-06050ef6e828-r-fd71a95d
```

The good:

- No I/O errors.
- No confusing logs.
- No orphaned replica directories. (Previous versions of the PR left them behind.)

The bad:

- A replica must be completely rebuild on the affected node. (Previously, the migration engine would crash, but we would not lose that replica.

ejweber:
I also backported and tested in Harvester. I was not able to reproduce I/O errors with the fix (though the stuck migration still persists).

ejweber:
Passed integration tests: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/5195/.

controller/volume_controller.go
```diff
@@ -3995,6 +4001,8 @@ func (c *VolumeController) prepareReplicasAndEngineForMigration(v *longhorn.Volu
 				return false, true, nil
 			case longhorn.ReplicaModeRW:
 				currentAvailableReplicas[dataPath] = r
+			case "":
+				log.Warnf("Running replica %v wasn't added to engine, will ignore it and continue migration", r.Name)
```
shuo-wu, line 4000: Should we clean up or fail this kind of replica?

controller/replica_controller.go
```diff
@@ -303,11 +303,17 @@ func (rc *ReplicaController) syncReplica(key string) (err error) {
...
+		if err != nil {
+			return errors.Wrapf(err, "failed to list replicas by node before deleting replica %v", replica.Name)
+		}
+
 		if replica.Spec.NodeID != "" && replica.Spec.NodeID != rc.controllerID {
 			log.Warn("Failed to cleanup replica's data because the replica's data is not on this node")
 		} else if replica.Spec.NodeID != "" {
 			if replica.Spec.BackendStoreDriver == longhorn.BackendStoreDriverTypeV1 {
-				if replica.Spec.Active && dataPath != "" {
+				// Clean up the data directory if this is the active replica or if there is no matching replica.
```
shuo-wu, line 315: NIT: 
```suggest
or the inactive replica is the only replica using the directory 
```

shuo-wu (approved):
In general LGTM

controller/volume_controller.go
ejweber: I do not think we should clean it up. If the migration is reverted, it will be good to still have this replica around, as it will theoretically be much faster to rebuild it then it will be to build a new one.

I'm a bit less sure about whether or not we should fail it. Ultimately, I chose not to, as there is nothing wrong with it and it's nice to have it ready to immediately rebuild if the migration is reverted. In addition, if it remains failed for long enough, I think we will clean it up automatically, which I would prefer not to do for the reason above. (Obviously we expect migrations to be fast, but this issue was only raised in the first place due to very long running migrations.) The downside is that there is an extra replica process running for the length of the migration.

If you feel there is a strong reason to fail it instead, please let me know. We can wait to merge until we are agreed.

controller/replica_controller.go
ejweber: Fixed in an amend.

controller/replica_controller.go
```diff
@@ -303,11 +303,18 @@ func (rc *ReplicaController) syncReplica(key string) (err error) {
 			return errors.Wrapf(err, "failed to cleanup the related replica instance before deleting replica %v", replica.Name)
 		}
 
+		rs, err := rc.ds.ListReplicasByNodeRO(replica.Spec.NodeID)
```
innobead, line 306: nit: rs can be initialized later in the following `else if replica.Spec.NodeID != ""` condition.

innobead (approved):

innobead:

PR #2276, opened 2023-10-31
fix(backup): backupInfo might be nil

mantissahz:
The `backupInfo` would be nil when we try to pull the backups from the remote backup target.

Ref: longhorn/longhorn#6998


mantissahz:

innobead (approved):

PR #2281, opened 2023-11-05
Fix resource not found error

derekbit:
1. Get the secret/configmap directly from API server if it is not in longhorn-system namespace.
2. PriorityClass is non-namespaces, so moving it to KubeInformerFactory is much clear.

Longhorn/longhorn#7045


innobead (approved):
LGTM.

innobead:
@mergify backport v1.5.x

PR #2283, opened 2023-11-10
fix(backing): disable block type disk for v1 type backing iamge

ChanYiLin:
ref: [longhorn/longhorn#7062](https://github.com/longhorn/longhorn/issues/7062)

When selecting disk/node for backingimage
we should skip block type disk

In the future we will support backing image in spdk and we should distinguish them when selecting disk type
ref: https://github.com/longhorn/longhorn/issues/6341


innobead (approved):
LGTM

innobead:
@mergify backport v1.5.x

PR #2292, opened 2023-11-13
Support instance-manager pod for v2 volumes on selected nodes

derekbit:
When a kubernetes node is labeled with `node.longhorn.io/disable-v2-volume: "true"`, the v2 volume feature on the node is not supported.

The PR is based on https://github.com/longhorn/longhorn-manager/pull/2272
    
Longhorn/longhorn#7015


innobead (approved):
LGTM

PR #2293, opened 2023-11-14
rwx: fix mount option typo

derekbit:
Should be `softerr`.

Longhorn/longhorn#7104


innobead (approved):

innobead:
This should be fixed in the master and 1.4.x?

derekbit:
Just in v1.5.x. Other branches are correct.

PR #2294, opened 2023-11-15
rwx: remount and reexport volume if share manager is in running state

derekbit:
longhorn/longhorn#7106

Deps on https://github.com/longhorn/longhorn-share-manager/pull/95/files


controller/share_manager_controller.go
```diff
@@ -257,6 +257,9 @@ func (c *ShareManagerController) handleErr(err error, key interface{}) {
 func (c *ShareManagerController) syncShareManager(key string) (err error) {
 	defer func() {
 		err = errors.Wrapf(err, "failed to sync %v", key)
+		if err != nil {
+			c.logger.WithError(err).Errorf("Failed to sync share manager %v", key)
+		}
```
innobead, line 262: Doesn't the error be logged in the below code already?

[/controller/utils.go#L72](https://github.com/longhorn/longhorn-manager/blob/b5e478db9114a9b47f6f4b0bff916a3e4568b0f3/controller/utils.go#L72)
```go
		logger.WithError(err).Error(mesg)
```

innobead (changes requested):
LGTM, but just one comment needs to resolve.

controller/share_manager_controller.go
derekbit: Ah... you're right.

BTW, 
```
	if apierrors.IsConflict(err) {
		logger.WithError(err).Debug(mesg)
```
Does using `Debug` for conflict make sense?

innobead: After discussing with @derekbit , we decided to change the log level to warn, so it would be still helpful for us to know if there is any failed resource update to cause the resource status not updated successfully.

controller/share_manager_controller.go
```diff
@@ -458,6 +458,34 @@ func (c *ShareManagerController) unmountShareManagerVolume(sm *longhorn.ShareMan
...
+	}
+
+	client, err := engineapi.NewShareManagerClient(sm, pod)
+	if err != nil {
+		return errors.Wrapf(err, "failed to create share manager client for pod %v", podName)
+	}
+	defer client.Close()
+
+	if err := client.Mount(); err != nil {
+		log.WithError(err).Warnf("Failed to mount share manager pod %v", podName)
```
innobead, line 482: You will return the error to the caller, so should we have this error log? or do you want to wrap another error by enriching the error message instead?

When you return the error to the caller, it will eventually be handled in the reconciler error handler.

innobead (changes requested):
Just one feedback.

innobead (approved):

derekbit:

PR #2299, opened 2023-11-16
build: reduce iamge size by removing unused packages

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/5911

PR #2302, opened 2023-11-18
Fix bug cannot mount xfs PV

PhanLe1010:
Add the `xfsprogs` back to longhorn-manger image so that longhorn-csi-plugin can run `mkfs.xfs` to mount the xfs PV

longhorn/longhorn#7140


innobead (approved):

innobead:
@PhanLe1010 See if we need to backport this.

PhanLe1010:
Luckily that we don't need to backport since the change https://github.com/longhorn/longhorn-manager/pull/2299 was no backported

PR #2303, opened 2023-11-21
feat(mount): request remount when volume mount becomes read only

ChanYiLin:

~~- VolumeController periodically use engineProxyClient to request instance-manager to check the globalMountPoint of the Volume~~
~~- If it is read only, we request remount of the pod by adding requestRemountAt~~

- Instance-Manager will periodically monitor the volume mount point and update the engine status
- Engine/Replica will sync the errorMsg from Instance-Manager
- Volume then check the Engine CRD to get the errorMsg

Pros:
using errorMsg to get the information from process allow us to extend the feature for more situations.

Blocked by: https://github.com/longhorn/longhorn-instance-manager/pull/316


innobead:
Is this ready for review? What's does `[BLOCKED]` mean? If not ready, just mark this PR draft.

ChanYiLin:
> Is this ready for review? What's does `[BLOCKED]` mean? If not ready, just mark this PR draft.

It is blocked by the PR: https://github.com/longhorn/longhorn-instance-manager/pull/316
instance-manager PR needs to merge first so this can update the vendor
I will change it to draft, thanks

controller/volume_controller.go
```diff
@@ -1402,6 +1407,19 @@ func (c *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[strin
 	return c.checkAndFinishVolumeRestore(v, e, rs)
 }
 
+func (c *VolumeController) checkVolumeMountStatus(v *longhorn.Volume, e *longhorn.Engine) (err error) {
+	log := getLoggerForVolume(c.logger, v)
+	if v.Status.State == longhorn.VolumeStateAttached && e.Status.CurrentState == longhorn.InstanceStateRunning {
+		if strings.Contains(e.Status.ErrorMsg, "READONLY") {
```
derekbit, line 1413: Can we introduce `e.Status.FilesystemReadOnly` for preventing string comparison?

ChanYiLin: Fixed, but I think it has room to discuss if we should put the value to engine status
or keep it in engine.Status.InstanceStatus.XXX, so we don't have to copy/pass the value

controller/volume_controller.go
```diff
@@ -1405,6 +1410,19 @@ func (c *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[strin
 	return c.checkAndFinishVolumeRestore(v, e, rs)
 }
 
+func (c *VolumeController) checkVolumeMountFileSystemStatus(v *longhorn.Volume, e *longhorn.Engine) (err error) {
+	log := getLoggerForVolume(c.logger, v)
+	if v.Status.State == longhorn.VolumeStateAttached && e.Status.CurrentState == longhorn.InstanceStateRunning {
+		if e.Status.FilesystemReadOnly {
```
c3y1huang, line 1416: Why not just check `e.Status.HealthConditions[imtypes.EngineConditionFilesystemReadOnly]`?

c3y1huang (commented):
Also, I'd suggest splitting the code-generation enhancement into a seperate commit.

controller/volume_controller.go
```diff
@@ -1405,6 +1410,19 @@ func (c *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[strin
...
+			v.Status.RemountRequestedAt = c.nowHandler()
+			log.Infof("Volume request remount at %v", v.Status.RemountRequestedAt)
+			msg := fmt.Sprintf("Volume %v requested remount at %v", v.Name, v.Status.RemountRequestedAt)
+			c.eventRecorder.Eventf(v, corev1.EventTypeNormal, constant.EventReasonRemount, msg)
+		}
+	}
+	return nil
+}
```
c3y1huang, line 1424: Consider removing the error return from this method since it will always return nil.

k8s/generate_code.sh
```diff
@@ -15,9 +15,11 @@ CRDS_DIR="crds"
 CONTROLLER_TOOLS_VERSION="v0.7.0"
 KUSTOMIZE_VERSION="kustomize/v3.10.0"
 
-if [[ -z "${GOPATH}" ]]; then
-  GOPATH=~/go
-fi
+# if [[ -z "${GOPATH}" ]]; then
+#   GOPATH=~/go
+# fi
```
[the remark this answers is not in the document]
ChanYiLin: Sorry, changes in `generate_code` should not be included in this PR
That's only my local change, thanks

controller/volume_controller.go
ChanYiLin: I agree with this change so we don't have to copy the status from `e.Status.HealthConditions` to other status field
This is originally from @derekbit 's advice 
cc @derekbit WDYT?

derekbit: Yeah, we don't need ` e.Status.FilesystemReadOnly` because you've already introduced `e.Status.HealthConditions`.

BTW, as discussion with @ChanYiLin, we can add FilesystemReadOnly condition in the `conditions` and no need to have a new `healthConditions`.

controller/volume_controller.go
```diff
@@ -1405,6 +1410,19 @@ func (c *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[strin
 	return c.checkAndFinishVolumeRestore(v, e, rs)
 }
 
+func (c *VolumeController) checkVolumeMountFileSystemStatus(v *longhorn.Volume, e *longhorn.Engine) (err error) {
```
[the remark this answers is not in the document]
ChanYiLin: sounds good to me, thanks!

controller/volume_controller.go
```diff
@@ -1405,6 +1410,19 @@ func (c *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[strin
...
+	log := getLoggerForVolume(c.logger, v)
+	if v.Status.State == longhorn.VolumeStateAttached && e.Status.CurrentState == longhorn.InstanceStateRunning {
+		if e.Status.FilesystemReadOnly {
+			v.Status.RemountRequestedAt = c.nowHandler()
+			log.Infof("Volume request remount at %v", v.Status.RemountRequestedAt)
+			msg := fmt.Sprintf("Volume %v requested remount at %v", v.Name, v.Status.RemountRequestedAt)
```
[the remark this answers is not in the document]
ChanYiLin: sure, thanks!

controller/volume_controller.go
ChanYiLin: fixed, thanks!

ChanYiLin:
Hi @c3y1huang @derekbit 
This PR has been fixed, PTAL, thanks!

c3y1huang (approved):
lgtm

shuo-wu (dismissed):

innobead (approved):
LGTM

PR #2304, opened 2023-11-21
feat: add GOCOVERDIR for deployment

FrankYang0529:
https://github.com/longhorn/longhorn/issues/5681

### DIFF
--- a/controller/backing_image_manager_controller.go
+++ b/controller/backing_image_manager_controller.go
@@ -789,6 +789,7 @@ func (c *BackingImageManagerController) createBackingImageManagerPod(bim *longho
 	if err != nil {
 		return err
 	}
+
 	if _, err := c.ds.CreatePod(podManifest); err != nil && !apierrors.IsAlreadyExists(err) {
 		return err
 	}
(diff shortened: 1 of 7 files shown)


types/deploy.go
```diff
@@ -11,3 +18,69 @@ const (
 	CSISnapshotterName = "csi-snapshotter"
 	CSIPluginName      = "longhorn-csi-plugin"
 )
+
+func AddGoCoverDirToPod(pod *corev1.Pod) {
```
innobead, line 22: nit: add a doc comment.

controller/backing_image_manager_controller.go
```diff
@@ -789,6 +789,8 @@ func (c *BackingImageManagerController) createBackingImageManagerPod(bim *longho
 	if err != nil {
 		return err
 	}
+	types.AddGoCoverDirToPod(podManifest)
```
innobead, line 792: Suggest moving this to `generateBackingImageManagerPodManifest` consistent with other changes here.

FrankYang0529: Good catch. Updated it. Thank you.

types/deploy.go
FrankYang0529: Added doc comment for `AddGoCoverDirToPod` and `AddGoCoverDirToDaemonSet`.

innobead (approved):

PR #2307, opened 2023-11-21
Remove engine image dependency of v2 volumes

derekbit:
https://github.com/longhorn/longhorn/issues/7157


derekbit:
Closed and reviewed in https://github.com/longhorn/longhorn-manager/pull/2371/commits

PR #2311, opened 2023-11-22
fix(crypto): minimal encrypted volume size

mantissahz:
The minimal encrypted volume size should be bigger than 16MiB because the default LUKS2 header size is 16MiB.

Ref: longhorn/longhorn#7033

### DIFF
--- a/csi/crypto/crypto.go
+++ b/csi/crypto/crypto.go
@@ -16,6 +16,12 @@ const (
 	CryptoKeyDefaultHash   = "sha256"
 	CryptoKeyDefaultSize   = "256"
 	CryptoDefaultPBKDF     = "argon2i"
+
+	// Luks2MinimalVolumeSize the minimal volume size for the LUKS2format encryption.
+	//  https://gitlab.com/cryptsetup/cryptsetup/-/wikis/FrequentlyAskedQuestions
+	//  Section 10.10 What about the size of the LUKS2 header
+	//  The default size is 16MB
+	Luks2MinimalVolumeSize = 16 * 1024 * 1024
 )
 
 // EncryptParams keeps the customized cipher options from the secret CR
--- a/datastore/longhorn.go
+++ b/datastore/longhorn.go
@@ -22,6 +22,7 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
+	"github.com/longhorn/longhorn-manager/csi/crypto"
 	"github.com/longhorn/longhorn-manager/types"
 	"github.com/longhorn/longhorn-manager/util"
 
@@ -628,6 +629,11 @@ func CheckVolume(v *longhorn.Volume) error {
 	if v.Name == "" || size == 0 || v.Spec.NumberOfReplicas == 0 {
 		return fmt.Errorf("BUG: missing required field %+v", v)
 	}
+
+	if v.Spec.Encrypted && size <= crypto.Luks2MinimalVolumeSize {
+		return fmt.Errorf("invalid volume size %v, need to be bigger than 16MiB, default LUKS2 header size for encryption", v.Spec.Size)
+	}
+
 	errs := validation.IsDNS1123Label(v.Name)
 	if len(errs) != 0 {
 		return fmt.Errorf("invalid volume name: %+v", errs)


csi/crypto/crypto.go
```diff
@@ -16,6 +16,8 @@ const (
 	CryptoKeyDefaultHash   = "sha256"
 	CryptoKeyDefaultSize   = "256"
 	CryptoDefaultPBKDF     = "argon2i"
+
+	Luks2MinimalVolumeSize = 16 * 1024 * 1024
```
derekbit, line 20: Can you add a reference link for the size?

csi/node_server.go
```diff
@@ -442,6 +443,14 @@ func (ns *NodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
 			return nil, status.Errorf(codes.InvalidArgument, "unsupported disk encryption format %v", diskFormat)
 		}
 
+		volumeSize, err := util.ConvertSize(volume.Size)
+		if err != nil {
+			return nil, status.Error(codes.Internal, err.Error())
+		}
+		if volumeSize <= crypto.Luks2MinimalVolumeSize {
```
derekbit, line 450: "<=" or "<" ?

csi/node_server.go
```diff
@@ -442,6 +443,14 @@ func (ns *NodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
...
+			return nil, status.Errorf(codes.Aborted, "unsupported volume size %v, need to be bigger than 16MiB default LUKS2 header size for encryption", volume.Size)
```
derekbit, line 451: Using `codes.InvalidArgument` instead.

Ref:
**InvalidArgument** indicates client specified an invalid argument.
**Aborted** indicates the operation was aborted, typically due to a concurrency issue like sequencer check failures, transaction aborts, etc

csi/node_server.go
```diff
@@ -442,6 +443,14 @@ func (ns *NodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
 			return nil, status.Errorf(codes.InvalidArgument, "unsupported disk encryption format %v", diskFormat)
 		}
 
+		volumeSize, err := util.ConvertSize(volume.Size)
+		if err != nil {
+			return nil, status.Error(codes.Internal, err.Error())
```
derekbit, line 448:
```suggestion
			return nil, status.Error(codes. InvalidArgument, err.Error())
```

csi/node_server.go
mantissahz: "<=", because I tested 16MiB volume and it still could not be activated.

csi/node_server.go
mantissahz: Modified. Thanks for the explanation.

csi/node_server.go
mantissahz: Modified. Thanks

derekbit:
Should we add the same check for RWX volumes?
https://github.com/longhorn/longhorn-share-manager/tree/master/pkg/crypto

csi/node_server.go
derekbit: Thank you for the clarification.

csi/node_server.go
```diff
@@ -442,6 +443,14 @@ func (ns *NodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
...
 		}
 
+		volumeSize, err := util.ConvertSize(volume.Size)
+		if err != nil {
+			return nil, status.Error(codes.InvalidArgument, err.Error())
+		}
+		if volumeSize <= crypto.Luks2MinimalVolumeSize {
+			return nil, status.Errorf(codes.InvalidArgument, "unsupported volume size %v, need to be bigger than 16MiB default LUKS2 header size for encryption", volume.Size)
+		}
+
```
derekbit, line 453: Can we move the check to volume validator?

mantissahz: Yes, moved it to the volume validator but it would only show the error message one of `longhorn-csi-plugin` pods.

derekbit (approved):
LGTM

datastore/longhorn.go
```diff
@@ -628,6 +629,11 @@ func CheckVolume(v *longhorn.Volume) error {
 	if v.Name == "" || size == 0 || v.Spec.NumberOfReplicas == 0 {
 		return fmt.Errorf("BUG: missing required field %+v", v)
 	}
+
+	if v.Spec.Encrypted && size <= crypto.Luks2MinimalVolumeSize {
+		return fmt.Errorf("invalid volume size %v, need to be bigger than 16MiB, default LUKS2 header size for encryption", v.Spec.Size)
```
innobead, line 634: nit: you can use `crypto.Luks2MinimalVolumeSize` in log instead of hard-coding `16 MiB`. 

ref: you can use https://pkg.go.dev/github.com/ylin610/units for unit formatting.

innobead (approved):
LGTM.

PR #2312, opened 2023-11-23
share manager: update status to error if unable to re-mount volume

derekbit:
Longhorn/longhorn#7183


innobead (approved):
LGTM

derekbit:
After rethinking the solution, we can apply the fix.
The network unreachable error is rare, because share-manager resource and pod are on the same node. A network error is usually caused by an unrecoverable gRPC error, so setting the share-manager to error and triggering share-manager restart should be fine.

derekbit:
@Mergify backport v1.4.x v1.5.x

PR #2315, opened 2023-11-23
build: add zip/unzip package for backup and restore

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/7143


ChanYiLin:
So originally (before started reducing image size), we had
```
RUN zypper -n install curl nfs-client iproute2 bind-utils iputils telnet \
              zip unzip e2fsprogs e2fsprogs-devel xfsprogs xfsprogs-devel cifs-utils && \
```

packages can be removed are
```
curl, telnet, e2fsprogs-devel, xfsprogs-devel
```

innobead (approved):

PR #2316, opened 2023-11-24
k8s: fix generate_code.sh

derekbit:
Fix "unbound variable" error if GOPATH is not set correctly.

### DIFF
--- a/k8s/generate_code.sh
+++ b/k8s/generate_code.sh
@@ -14,9 +14,16 @@ CODE_GENERATOR_VERSION="v0.18.0"
 CRDS_DIR="crds"
 CONTROLLER_TOOLS_VERSION="v0.7.0"
 KUSTOMIZE_VERSION="kustomize/v3.10.0"
+GOPATH="${GOPATH:-}"
 
-if [[ -z "${GOPATH}" ]]; then
-  GOPATH=~/go
+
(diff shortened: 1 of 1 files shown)


k8s/generate_code.sh
```diff
@@ -15,8 +15,12 @@ CRDS_DIR="crds"
 CONTROLLER_TOOLS_VERSION="v0.7.0"
 KUSTOMIZE_VERSION="kustomize/v3.10.0"
 
-if [[ -z "${GOPATH}" ]]; then
-  GOPATH=~/go
+if [[ -z "${GOPATH:-}" ]]; then
```
innobead, line 18: Suggest moving `GOPATH` declaration to the variable declaration above, so you don't need to do this repeatedly in each condition expression. 

```shell
GOPATH="${GOPATH:-}"

if [[ -z "$GOPATH" ]]; then
  GOPATH="$(go env GOPATH)"
  
  if [[ -z "$GOPATH" ]]; then
    echo "GOPATH is not set"
    exit 1
  fi
fi
```

innobead (changes requested):

k8s/generate_code.sh
derekbit: Updated. Thank you.

innobead (approved):

PR #2317, opened 2023-11-27
6776 - Simplify share-manager volume attachment path.

james-munson:
Issues:  https://github.com/longhorn/longhorn/issues/6776 and https://github.com/longhorn/longhorn/issues/7112

### DIFF
--- a/controller/share_manager_controller.go
+++ b/controller/share_manager_controller.go
@@ -405,27 +405,51 @@ func (c *ShareManagerController) isShareManagerRequiredForVolume(sm *longhorn.Sh
 	return false
 }
 
-func (c *ShareManagerController) createShareManagerAttachmentTicket(sm *longhorn.ShareManager, va *longhorn.VolumeAttachment) {
+func (c *ShareManagerController) createShareManagerAttachmentTicket(sm *longhorn.ShareManager, va *longhorn.VolumeAttachment) error {
 	log := getLoggerForShareManager(c.logger, sm)
+	podName := types.GetShareManagerPodNameFromShareManagerName(sm.Name)
+	pod, err := c.ds.GetPodRO(c.namespace, podName)
+	if err != nil && !apierrors.IsNotFound(err) {
+		return errors.Wrapf(err, "failed to retrieve pod %v for share manager from datastore", podName)
+	}
+
+	// This is not a fatal error, just wait for the pod.
+	if pod == nil {
+		log.Infof("No share-manager pod yet to attach the volume %v", va.Name)
+		return nil
+	}
+	if pod.Spec.NodeName == "" {
+		log.Infof("Share-manager pod %v for volume %v reports no owner node", pod.Name, va.Name)
+		return nil
+	}
+	if pod.Status.Phase != corev1.PodPending && pod.Status.Phase != corev1.PodRunning {
+		log.Infof("Share-manager pod %v for volume %v reports unusable phase %v", pod.Name, va.Name, pod.Status.Phase)
+		return nil
+	}
+	nodeID := pod.Spec.NodeName
+
 	shareManagerAttachmentTicketID := longhorn.GetAttachmentTicketID(longhorn.AttacherTypeShareManagerController, sm.Name)
 	shareManagerAttachmentTicket, ok := va.Spec.AttachmentTickets[shareManagerAttachmentTicketID]
 	if !ok {
-		// create new one
+		//create new one
 		shareManagerAttachmentTicket = &longhorn.AttachmentTicket{
 			ID:     shareManagerAttachmentTicketID,
 			Type:   longhorn.AttacherTypeShareManagerController,
-			NodeID: sm.Status.OwnerID,
+			NodeID: nodeID,
 			Parameters: map[string]string{
 				longhorn.AttachmentParameterDisableFrontend: longhorn.FalseValue,
 			},
 		}
+		log.Infof("Adding volume attachment ticket: %v to attach the volume %v to node %v", shareManagerAttachmentTicketID, va.Name, nodeID)
 	}
-	if shareManagerAttachmentTicket.NodeID != sm.Status.OwnerID {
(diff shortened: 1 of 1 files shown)


controller/share_manager_controller.go
```diff
@@ -407,22 +407,32 @@ func (c *ShareManagerController) isShareManagerRequiredForVolume(sm *longhorn.Sh
 
 func (c *ShareManagerController) createShareManagerAttachmentTicket(sm *longhorn.ShareManager, va *longhorn.VolumeAttachment) {
 	log := getLoggerForShareManager(c.logger, sm)
+	nodeID := ""
+	pod, err := c.ds.GetPod(types.GetShareManagerPodNameFromShareManagerName(sm.Name))
```
innobead, line 411: Use `GetPodRO` instead.

controller/share_manager_controller.go
```diff
@@ -407,22 +407,32 @@ func (c *ShareManagerController) isShareManagerRequiredForVolume(sm *longhorn.Sh
...
+	if err != nil || pod == nil {
+		log.Infof("No share-manager pod yet to attach the volume %v to node %v", va.Name, nodeID)
+		return
+	}
+
```
innobead, line 416: Remove this empty line to make the logic together.

controller/share_manager_controller.go
```diff
@@ -407,22 +407,32 @@ func (c *ShareManagerController) isShareManagerRequiredForVolume(sm *longhorn.Sh
 
 func (c *ShareManagerController) createShareManagerAttachmentTicket(sm *longhorn.ShareManager, va *longhorn.VolumeAttachment) {
 	log := getLoggerForShareManager(c.logger, sm)
+	nodeID := ""
+	pod, err := c.ds.GetPod(types.GetShareManagerPodNameFromShareManagerName(sm.Name))
+	if err != nil || pod == nil {
```
innobead, line 412: Use `ErrorIsNotFound` instead. If it's not a not-found error, still return the error  instead of swallowing it.

innobead (changes requested):
In general, LGTM. Just a few comments need to resolve.

controller/share_manager_controller.go
james-munson: Done.  I also switched the call in isResponsibleFor() because it also uses the pod as just a read-only object.

shuo-wu (commented):
Besides, as I mentioned in [the comment](https://github.com/longhorn/longhorn/issues/6776#issuecomment-1772056218), should we improve the detachment part?

controller/share_manager_controller.go
```diff
@@ -405,27 +405,43 @@ func (c *ShareManagerController) isShareManagerRequiredForVolume(sm *longhorn.Sh
...
+	if err != nil && !apierrors.IsNotFound(err) {
+		return errors.Wrapf(err, "failed to retrieve pod %v for share manager from datastore", podName)
+	}
+
+	// This is not a fatal error, just wait for the pod.
+	if pod == nil {
+		log.Infof("No share-manager pod yet to attach the volume %v", va.Name)
+		return nil
+	}
+	nodeID := pod.Spec.NodeName
```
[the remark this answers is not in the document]
james-munson: Yes, found the case of pod.Spec.Nodename == "" while testing.

james-munson:
Rebased and squashed.

controller/share_manager_controller.go
```diff
@@ -405,27 +405,51 @@ func (c *ShareManagerController) isShareManagerRequiredForVolume(sm *longhorn.Sh
...
+		return nil
+	}
+	if pod.Spec.NodeName == "" {
+		log.Infof("Share-manager pod %v for volume %v reports no owner node", pod.Name, va.Name)
+		return nil
+	}
+	if pod.Status.Phase != corev1.PodPending && pod.Status.Phase != corev1.PodRunning {
+		log.Infof("Share-manager pod %v for volume %v reports unusable phase %v", pod.Name, va.Name, pod.Status.Phase)
+		return nil
+	}
```
innobead, line 428: nit: Suggest adding the code about pod (pre-)readness in a common place for reuse.

controller/share_manager_controller.go
```diff
@@ -405,27 +405,51 @@ func (c *ShareManagerController) isShareManagerRequiredForVolume(sm *longhorn.Sh
 	return false
 }
 
-func (c *ShareManagerController) createShareManagerAttachmentTicket(sm *longhorn.ShareManager, va *longhorn.VolumeAttachment) {
+func (c *ShareManagerController) createShareManagerAttachmentTicket(sm *longhorn.ShareManager, va *longhorn.VolumeAttachment) error {
```
innobead, line 408: You add `error` returned value, but forgot the update the caller needs to check the error.  

[/controller/share_manager_controller.go#L584-L586](https://github.com/longhorn/longhorn-manager/blob/a09908dab732d514a8750dd8ead776129c92838e/controller/share_manager_controller.go#L584-L586)
```go
	c.createShareManagerAttachmentTicket(sm, va)

	return nil
```

innobead (changes requested):
Almost there. After resolving the comments, we can merge this PR.

controller/share_manager_controller.go
james-munson: Actually, that was kind of on purpose.  If there is an error from createShareManagerAttachmentTicket(), and it is returned back up by syncShareManagerVolume(), the current code in syncShareManager() will just skip the calls to syncShareManagerPod() and syncShareManagerEndpoint().  Is that really what we want?  It seems to open up deadlock possibilities for when successfully doing the Pod sync would clear up the situation causing an error in Volume sync, but we never get the chance to do that.

innobead: Agreed, that is intended. However, suggest updating the code as below to make it explicit. 

```go
_ = c.createShareManagerAttachmentTicket(sm, va)
```

controller/share_manager_controller.go
james-munson: Will do.  I'll save this for the next set of changes.

controller/share_manager_controller.go
james-munson: Done.

innobead (approved):

james-munson:

PR #2322, opened 2023-12-01
Global setting default-data-path supports block device

derekbit:
Longhorn/longhorn#7234

### DIFF
--- a/types/types.go
+++ b/types/types.go
@@ -956,6 +956,26 @@ func UnmarshalToNodeTags(s string) ([]string, error) {
 }
 
 func CreateDefaultDisk(dataPath string, storageReservedPercentage int64) (map[string]longhorn.DiskSpec, error) {
+	fileInfo, err := os.Stat(dataPath)
+	if err != nil {
+		return nil, err
+	}
+
+	// block-type disk
+	if (fileInfo.Mode() & os.ModeDevice) == os.ModeDevice {
+		return map[string]longhorn.DiskSpec{
+			DefaultDiskPrefix + util.RandomID(): {
+				Type:              longhorn.DiskTypeBlock,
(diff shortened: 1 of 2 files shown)


types/types.go
```diff
@@ -956,6 +956,27 @@ func UnmarshalToNodeTags(s string) ([]string, error) {
...
+
+	// block-type disk
+	if (fileInfo.Mode() & os.ModeDevice) == os.ModeDevice {
+		return map[string]longhorn.DiskSpec{
+			DefaultDiskPrefix + util.RandomID(): {
+				Type:              longhorn.DiskTypeBlock,
+				Path:              dataPath,
+				AllowScheduling:   true,
+				EvictionRequested: false,
+				// TODO: should we support storageReserved for block device?
```
innobead, line 972: Don't think so, because it should be fully managed and used in lvolstore, right?

innobead (dismissed):
LGTM. Afer removing the comment, we can merget this.

webhook/resources/node/validator.go
```diff
@@ -61,6 +61,12 @@ func (n *nodeValidator) Create(request *admission.Request, newObj runtime.Object
 				return werror.NewInvalidError(fmt.Sprintf("disk %v type %v is not supported since v2 data engine is disabled", name, disk.Type), "")
 			}
 		}
+
+		if disk.Type == longhorn.DiskTypeBlock {
+			if disk.StorageReserved != 0 {
+				return werror.NewInvalidError(fmt.Sprintf("disk %v type %v is not supported to reserve storage", name, disk.Type), "")
```
innobead, line 67: Good update!

innobead (approved):

PR #2324, opened 2023-12-01
feat(node drain): add DetachManuallyAttachedVolumesDuringDrain setting

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/6978

iterate the volumeattachment ticket
if there is `AttacherTypeLonghornAPI` ticket and the target node is being drained, then delete the ticket
in the next function `handleVolumeDetachment` will detach the volume if the volume is currently attached by this ticket.

### DIFF
--- a/controller/volume_attachment_controller.go
+++ b/controller/volume_attachment_controller.go
@@ -20,6 +20,7 @@ import (
 	clientset "k8s.io/client-go/kubernetes"
 	v1core "k8s.io/client-go/kubernetes/typed/core/v1"
 
+	"github.com/longhorn/longhorn-manager/constant"
 	"github.com/longhorn/longhorn-manager/datastore"
 	"github.com/longhorn/longhorn-manager/types"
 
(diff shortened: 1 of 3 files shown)


controller/volume_attachment_controller.go
```diff
@@ -289,6 +291,41 @@ func (vac *VolumeAttachmentController) reconcile(vaName string) (err error) {
...
+	var manualAttachmentTicket *longhorn.AttachmentTicket
+	for _, attachmentTicket := range va.Spec.AttachmentTickets {
+		if attachmentTicket.Type == longhorn.AttacherTypeLonghornAPI {
+			manualAttachmentTicket = attachmentTicket
+			break
+		}
+	}
+
+	if manualAttachmentTicket != nil {
+		unschedulable, err := vac.ds.IsKubeNodeUnschedulable(manualAttachmentTicket.NodeID)
```
innobead, line 317: Is it possible to watch the eviction resource events of Longhorn pods to further determine if the drain operation is running, instead of only checking if the node is unscheduled?

P.S. that node is unscheduled does not mean it is being drained.

innobead (changes requested):

controller/volume_attachment_controller.go
ChanYiLin: The goal is to detach the UI attached volume from the node when it is drained
so if we can be assured that `node.spec.unschedulable` is only set to true when the node is cordoned/drained
than we don't need to add more conditions

innobead:
Sounds good to me as per both updates.

controller/volume_attachment_controller.go
```diff
@@ -91,6 +92,12 @@ func NewLonghornVolumeAttachmentController(
 	}, 0)
 	vac.cacheSyncs = append(vac.cacheSyncs, ds.EngineInformer.HasSynced)
 
+	ds.KubeNodeInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
```
innobead, line 95: Why need to watch delete event?

Why not using `AddEventHandlerWithResyncPeriod`?

innobead (changes requested):
In general, just one comment.

controller/volume_attachment_controller.go
ChanYiLin: Fixed, thanks!

innobead (approved):

PR #2332, opened 2023-12-05
feat: default priority class

mantissahz:
For an upgrade, not update the setting `priority-class` if there
is an value of the the setting `priority-class` or one or more
volumes attached.

Ref: longhorn/longhorn#6528

### DIFF
--- a/datastore/longhorn.go
+++ b/datastore/longhorn.go
@@ -198,13 +198,25 @@ func (s *DataStore) applyCustomizedDefaultSettingsToDefinitions(customizedDefaul
 func (s *DataStore) syncSettingCRsWithCustomizedDefaultSettings(customizedDefaultSettings map[string]string, defaultSettingCMResourceVersion string) error {
 	for _, sName := range types.SettingNameList {
 		configMapResourceVersion := ""
-		if s, err := s.GetSettingExactRO(sName); err != nil {
+		setting, err := s.GetSettingExactRO(sName)
+		if err != nil {
 			if !apierrors.IsNotFound(err) {
 				return err
 			}
 		} else {
-			if s.Annotations != nil {
-				configMapResourceVersion = s.Annotations[types.GetLonghornLabelKey(types.ConfigMapResourceVersionKey)]
+			if setting.Annotations != nil {
+				configMapResourceVersion = setting.Annotations[types.GetLonghornLabelKey(types.ConfigMapResourceVersionKey)]
+			}
+		}
+
+		if sName == types.SettingNamePriorityClass {
+			volumesDetached, err := s.AreAllVolumesDetachedState()
+			if err != nil {
+				return errors.Wrapf(err, "failed to list volumes to check if all volumes are detached")
+			}
+			if !volumesDetached {
+				logrus.Warnf("the setting %v will not be updated as one or more volumes are attached, or the setting has been customized by users", types.SettingNamePriorityClass)
+				continue
 			}
 		}
 
--- a/types/setting.go
+++ b/types/setting.go
@@ -1515,7 +1515,7 @@ func isValidChoice(choices []string, value string) bool {
 func GetCustomizedDefaultSettings(defaultSettingCM *corev1.ConfigMap) (defaultSettings map[string]string, err error) {
 	defaultSettingYAMLData := []byte(defaultSettingCM.Data[DefaultSettingYAMLFileName])
 
-	defaultSettings, err = getDefaultSettingFromYAML(defaultSettingYAMLData)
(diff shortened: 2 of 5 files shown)


types/setting.go
```diff
@@ -729,6 +730,7 @@ var (
 		Category: SettingCategoryDangerZone,
 		Required: false,
 		ReadOnly: false,
+		Default:  DefaultPriorityClassName,
```
innobead, line 733: No need to have a default value here, because this value will be controlled by chart values as we discussed.

upgrade/v15xto160/upgrade.go
```diff
@@ -160,6 +164,55 @@ func upgradeVolumeAttachments(namespace string, lhClient *lhclientset.Clientset,
 	return nil
 }
 
+func upgradeDefaultPriorityClass(namespace string, lhClient *lhclientset.Clientset, resourceMaps map[string]interface{}) (err error) {
```
innobead, line 167: We don't need to do this instead of just relying on the chart values (or manifest)

innobead (changes requested):
James, as we discussed, this PR should not be necessary anymore. WDYT?

mantissahz:
> James, as we discussed, this PR should not be necessary anymore. WDYT?

Agreed and closed this PR.

upgrade/v15xto160/upgrade.go
mantissahz: Removed the upgrade path for the default Priority Class.

datastore/longhorn.go
```diff
@@ -197,44 +197,55 @@
...
+			}
+		}
+
+		if sName == types.SettingNamePriorityClass {
+			volumesDetached, err := s.AreAllVolumesDetachedState()
+			if err != nil {
+				return errors.Wrapf(err, "failed to list volumes to check if all volumes are detached")
+			}
+			if !volumesDetached || (setting != nil && setting.Value != "") {
+				continue
```
innobead, line 220: Can we create a warning log to inform users that the setting will not be updated because there are volumes attached?

I assume the controller will not trigger this function except when there is a new change from the config map of Longhorn's default settings. Please confirm this to prevent any unexpected log flooding.

innobead (changes requested):
In general, LGTM. One comment needs to answer before merging.

datastore/longhorn.go
mantissahz: Yes, it would be triggered only when the daemon is starting or the config map of Longhorn's default settings has been changed.
Modified. Thanks

datastore/longhorn.go
```diff
@@ -200,13 +200,25 @@ func (s *DataStore) applyCustomizedDefaultSettingsToDefinitions(customizedDefaul
...
+				configMapResourceVersion = setting.Annotations[types.GetLonghornLabelKey(types.ConfigMapResourceVersionKey)]
+			}
+		}
+
+		if sName == types.SettingNamePriorityClass {
+			volumesDetached, err := s.AreAllVolumesDetachedState()
+			if err != nil {
+				return errors.Wrapf(err, "failed to list volumes to check if all volumes are detached")
+			}
+			if !volumesDetached || (setting != nil && setting.Value != "") {
```
innobead, line 219: Then, later if users update the config map directly, the setting will not be updated anymore. The source of truth should be from the config map, right?

```go
(setting != nil && setting.Value != "")
```

It seems you need to remove this condition and also still prevent this change during the upgrade.

innobead (changes requested):

datastore/longhorn.go
innobead: Sounds good, but suggest the parameter named `allowOverride bool`.

mantissahz: Modified to that change the ConfigMap `longhorn-default-setting` for the setting `priority-class` if there are one or more volumes attached or the setting `priority-class` has been customized by users when doing the upgrade.

innobead: Can't understand. Can you explain the change again?

mantissahz: When the Longhorn daemon starts and walks through the upgrade path, it will check the ConfigMap CR `longhorn-default-setting`.
The setting `priority-class` in ConfigMap CR `longhorn-default-setting` will be set as the value of the setting `priority-class` in the cluster if volumes are attached or users have customized the setting `priority-class` of the cluster.
The setting 'priority-class' of the default settings in ConfigMap CR `longhorn-default-setting` will be consistent with the setting 'priority-class' of the cluster.

mantissahz:
The PR will be invalid after the PR https://github.com/longhorn/longhorn-manager/pull/2411 of https://github.com/longhorn/longhorn/issues/7173

PR #2333, opened 2023-12-06
fix: scope of err variable while migrating volume

m-ildefons:
Fix the scope of the err variable while migrating a volume.

The root of this problem is that in case a migration revert is attempted, `err` is re-defined in a defer'ed anonymous function. This shadows the definition of the variable `err` in the outer scope, which causes the original error during the migration to be lost. The simple fix is to define a different error and wrap the original error with a message containing the new error. This avoids redefining `err`, making it available with its original value in the defer'ed function.

related: longhorn/longhorn#7009


m-ildefons:
@mergify backport v1.5.x

PR #2355, opened 2023-12-12
refactor: move non-entry app functions away from app pkg

innobead:


app/daemon.go
```diff
@@ -3,9 +3,8 @@ package app
 import (
 	"fmt"
 	"net/http"
-	"os"
-
 	_ "net/http/pprof" // for runtime profiling
+	"os"
```
c3y1huang, line 7: Any reason we are moving this here?

innobead: just the default formatter by goimports, nothing special.

c3y1huang (approved):

innobead:

PR #2357, opened 2023-12-12
Allow kubelet to be down for 10 seconds before responding

ejweber:
longhorn/longhorn#7302

When kubelet goes down, we generally aren't able to update the node.longhorn.io for 40 seconds.

- https://github.com/kubernetes/kubernetes/blob/0c645922edcc06adff43c70c02fb56751364bbb5/pkg/controller/nodelifecycle/config/v1alpha1/defaults.go#L41
- https://github.com/kubernetes/kubernetes/blob/0c645922edcc06adff43c70c02fb56751364bbb5/pkg/controller/nodelifecycle/node_lifecycle_controller.go#L926

However, when kubelet is restarted, it is able to report an interim unhealthy status immediately. In my testing, this interim unhealthy status may only last for less than a second, but it leads directly to the killing of migration engines.

This PR allows us to wait 10 seconds before reacting to an unhealthy kubelet.

### DIFF
--- a/controller/node_controller.go
+++ b/controller/node_controller.go
@@ -38,6 +38,7 @@ import (
 
 const (
 	nodeControllerResyncPeriod = 30 * time.Second
+	ignoreKubeletNotReadyTime  = 10 * time.Second
 
 	unknownDiskID = "UNKNOWN_DISKID"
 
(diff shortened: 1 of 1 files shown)


ejweber:
Thanks to @PhanLe1010 and @james-munson for suggesting this fairly straightforward approach. I experimented with more difficult (and potentially more dangerous) approached first.

james-munson (approved):
Sweet.

PhanLe1010 (approved):
LGTM

c3y1huang (approved):

innobead (approved):

ejweber:

PR #2360, opened 2023-12-13
ci: import test script

m-ildefons:
- By default run only tests on affected modules, if there are uncomitted changes in the git worktree. This speeds up local unit test runtime during development by only running tests on affected modules. In the CI pipeline or when changes have been committed, all unit tests are run for complete coverage.

- Make it possible to run particular tests using the TESTS environment variable. This is helpful when writing tests, as it helps both with turnaround times as well as reduce the output to only include the test under development.

### DIFF
--- a/scripts/test
+++ b/scripts/test
@@ -1,15 +1,25 @@
 #!/bin/bash
 set -e
 
-cd $(dirname $0)/..
+TESTS=${TESTS:-""}
 
-args=$1
+cd "$(dirname "$0")/.."
 
-echo Running tests
+args=("$@")
 
-PACKAGES="$(find . -name '*.go' | xargs -I{} dirname {} | sort -u | grep -Ev '(.git|.trash-cache|vendor|bin|k8s)')"
+echo "Running tests ${TESTS}"
 
-echo Packages: ${PACKAGES}
+if ! git diff --name-only | grep -Ev '(.git|.trash-cache|vendor|bin|k8s)' | grep -E '.*\.go' ; then
+  PACKAGES="$(find . -name '*.go' -exec dirname {} \; | sort -u | grep -Ev '(.git|.trash-cache|vendor|bin|k8s)')"
+else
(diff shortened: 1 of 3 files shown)


PhanLe1010 (commented):
Great idea! 

I have a small concern: if dev wants to run all tests and if there are uncommitted changes, it is impossible after this PR. The workaround would be to commit the changes and run the tests. Maybe we can document this point in README.md?

PhanLe1010 (approved):
LGTM

scripts/test
```diff
@@ -1,15 +1,25 @@
...
+fi
+
+echo Packages: "${PACKAGES}"
 
 [ "${ARCH}" == "amd64" ] && RACE=-race
-go test ${RACE} ${args} ${PACKAGES}
+if [ -n "${TESTS}" ]; then
+  go test -v "${RACE}" "${args[@]}" ${PACKAGES} -check.f "${TESTS}"
+else
+  go test -v "${RACE}" "${args[@]}" ${PACKAGES}
```
james-munson, line 24: Did you mean to also wrap "${PACKAGES}" in quotes as well, for the same reasons?  It make shellcheck happy.

james-munson (changes requested):
Just one nit.

scripts/test
m-ildefons: It's deliberately not quoted to let the shell re-split the words. The "correct" way would be to generate an array of package names in `$PACKAGES` and use `"${PACKAGES[@]}"` with quotes to pass the package names individually without expanding globs. However this is way more involved and I can pretty safely assume that we're not going to introduce package names with globs in them in the near future.
If this was quoted as you suggest, the shell would pass a single string of all package names to `go test` instead of individual strings with each package name individually. This would not work.

innobead (approved):

innobead:
It seems we have this in other first-class repos (ex: longhorn-engine, instance manager, share manager, backing image, spdk engine). **It's better to sync them up. No need to backport, since we are primarily working on the master branch.**

PR #2361, opened 2023-12-13
Don't use updated golang.org/x/tools dependency for golint

ejweber:
longhorn/longhorn#7330

cc @c3y1huang, since I think you have done the most around Go and dependency versions.


ejweber:
There is an ongoing discussion in https://github.com/longhorn/longhorn/issues/7330 about whether this is the right approach.

c3y1huang:
https://github.com/longhorn/longhorn/issues/7330#issuecomment-1857089149
Let's have another PR to fix the out-of-sync Go versions.

ejweber:
Closed in favor of the more correct fixes in https://github.com/longhorn/longhorn/issues/7341.

PR #2363, opened 2023-12-14
Fix bug volume stuck in live engine upgrading forever if it was degraded during live engine upgrade

PhanLe1010:
Quick summary of [the live engine upgrade logic](https://github.com/longhorn/longhorn-manager/blob/ae6b6aa99d0e2eb12fd0d868f5e67105bb6b685c/controller/volume_controller.go#L2631):
1. Volume controller check for the volume is attached and healthy
1. Check the old engine image and new engine image is ready
1. Create new inactive replicas set from old active replica set. Note that new inactive replicas will use new engine image and the old set is still using the old engine image
1. Start the new inactive replica set
1. Set the `e.Spec.UpgradedReplicaAddressMap` to the new inactive replica set and `e.Spec.Image` to the new engine image to start the upgrading inside engine process
1. Once the engine controller finish the image upgrade for engine, it will set `e.Status.CurrentImage` to new image. At this point, volume controller can:
    1. Set the inactive replica set to active and set the old active replica to inactive
    1. Clear  `e.Spec.UpgradedReplicaAddressMap` and set `v.Status.CurrentImage` to new image to indicate that the live upgrade finished 

This fix considers 2 cases:

1. If volume becomes degraded before engine starts the upgrading (i.e. before step 5 above), we will clean up the inactive replica which is corresponding to the non-existing active replica. This way, the volume can starts rebuilding a new active replica using old engine image. Then the volume will become healthy again and the live upgrade logic can finish.

    To demonstrate this case, let's consider the following test before the fix:

    1. Create/attach a volume using engine image v1.5.3
    1. Start a live upgrade by upgrade the engine for the volume to master-head
    1. Immediately delete a replica before the engine starts upgrading (i.e. before `engine.Spec.Image` is set to `master-head`)
    1. At this moment, `v.Spec.Image` is  `master-head`, `v.Status.CurrentImage` is  `1.5.3`, `engine.Spec.Image`  is  `1.5.3`, and `engine.Status.CurrentImage`  is  `1.5.3`. 
    1. The volume becomes degraded because it is missing an active replica. However, a new replica cannot be rebuilt as there already exists an inactive replica on the same disk
        ```
        [longhorn-manager-8254x] time="2023-12-14T02:57:52Z" level=error msg="There's no available disk for replica testvol-r-14cee1bd, size 10737418240" func="scheduler.(*ReplicaScheduler).ScheduleReplica" file="replica_scheduler.go:95"
        ```
    1. The live upgrade also cannot proceed as the volume is not healthy https://github.com/longhorn/longhorn-manager/blob/ae6b6aa99d0e2eb12fd0d868f5e67105bb6b685c/controller/volume_controller.go#L2692
    1. The result is volume stuck in degraded and live upgrade forever 
        Screenshot from 2023-12-13 19-00-41

     This PR fixes it so that we cleanup the inactive replica which doesn't have a corresponding active replica. This allow the replica scheduler schedules the active stopped replica in the above picture. So the volume becomes healthy again and be able to start and finish the live upgrade

1. If volume becomes degraded after engine starts the upgrading  (i.e. after step 5 above), the engine already uses the newly inactive replicas. We cannot rebuild a new active replica because that replica will use old engine image. It will be problematic as the engine will issue IO to some inactive replica (with new engine image) and the newly rebuilt active replica (with old engine image). Instead, we should continue and finish the live upgrade. Then the volume can start rebuilding new replica with new engine image

    To demonstrate this case, let's consider the following test before the fix:

    1. Create/attach a volume using engine image v1.5.3
    1. Start a live upgrade by upgrade the engine for the volume to master-head
    1. Wait for `engine.Spec.Image`  is  `master-head` (engine started to upgrade) then delete an old active replica 
    1. At this moment, `v.Spec.Image` is  `master-head`, `v.Status.CurrentImage` is  `1.5.3`, `engine.Spec.Image`  is  `master-head`, and `engine.Status.CurrentImage`  is  `1.5.3`. 
    1. Eventually, the engine controller performed the upgrade and set `engine.Status.CurrentImage` to  `master-head`. 
    1. The volume becomes degraded because it is missing an active replica. However, a new replica cannot be rebuilt as there already exists an inactive replica on the same disk
        ```
        [longhorn-manager-8254x] time="2023-12-14T02:57:52Z" level=error msg="There's no available disk for replica testvol-r-14cee1bd, size 10737418240" func="scheduler.(*ReplicaScheduler).ScheduleReplica" file="replica_scheduler.go:95"
        ```
     1. Even with the fix # 1 (which is deleting the inactive replica corresponding to the non-existing active replica), the volume still stuck in degrading and live upgrade. The reason is that at this moment, the engine already performed the upgrade and set `replicaModeMap: null` [link](https://github.com/longhorn/longhorn-manager/blob/ae6b6aa99d0e2eb12fd0d868f5e67105bb6b685c/controller/engine_controller.go#L1982). Engine controller is waiting for volume controller to mark the volume as finished live upgrading and clear `e.Spec.UpgradedReplicaAddressMap` the [link](https://github.com/longhorn/longhorn-manager/blob/ae6b6aa99d0e2eb12fd0d868f5e67105bb6b685c/controller/engine_controller.go#L824)
     1. However, volume controller is waiting for engine controller to refresh `e.Status.replicaModeMap` first [link](https://github.com/longhorn/longhorn-manager/blob/ae6b6aa99d0e2eb12fd0d868f5e67105bb6b685c/controller/volume_controller.go#L637) 
     1. The result is deadlock
        Screenshot from 2023-12-13 19-35-08

     This PR fixes it so that we when the volume is degraded, even though we don't start the live engine upgrade, try to mark it as finished if it already started. Thus solving the deadlock above

longhorn/longhorn#7012

### DIFF
--- a/controller/replica_controller.go
+++ b/controller/replica_controller.go
@@ -240,9 +240,9 @@ func (rc *ReplicaController) syncReplica(key string) (err error) {
 			return errors.Wrapf(err, "failed to cleanup the related replica instance before deleting replica %v", replica.Name)
 		}
 
-		rs, err := rc.ds.ListReplicasByNodeRO(replica.Spec.NodeID)
+		rs, err := rc.ds.ListVolumeReplicasRO(replica.Spec.VolumeName)
 		if err != nil {
-			return errors.Wrapf(err, "failed to list replicas by node before deleting replica %v", replica.Name)
+			return errors.Wrapf(err, "failed to list replicas of the volume before deleting replica %v", replica.Name)
 		}
 
 		if replica.Spec.NodeID != "" && replica.Spec.NodeID != rc.controllerID {
@@ -251,7 +251,7 @@ func (rc *ReplicaController) syncReplica(key string) (err error) {
 			if replica.Spec.BackendStoreDriver == longhorn.BackendStoreDriverTypeV1 {
 				// Clean up the data directory if this is the active replica or if this inactive replica is the only one
 				// using it.
-				if (replica.Spec.Active || !hasMatchingReplica(replica, rs)) && dataPath != "" {
+				if (replica.Spec.Active || !hasMatchingActiveReplica(replica, rs)) && dataPath != "" {
 					// prevent accidentally deletion
 					if !strings.Contains(filepath.Base(filepath.Clean(dataPath)), "-") {
 						return fmt.Errorf("%v doesn't look like a replica data path", dataPath)
@@ -882,9 +882,14 @@ func (rc *ReplicaController) isResponsibleFor(r *longhorn.Replica) bool {
 	return isControllerResponsibleFor(rc.controllerID, rc.ds, r.Name, r.Spec.NodeID, r.Status.OwnerID)
 }
 
-func hasMatchingReplica(replica *longhorn.Replica, replicas []*longhorn.Replica) bool {
+func hasMatchingActiveReplica(replica *longhorn.Replica, replicas map[string]*longhorn.Replica) bool {
 	for _, r := range replicas {
-		if r.Name != replica.Name && r.Spec.DataDirectoryName == replica.Spec.DataDirectoryName {
+		if r.Spec.Active &&
+			r.Name != replica.Name &&
+			r.Spec.NodeID == replica.Spec.NodeID &&
+			r.Spec.DiskID == replica.Spec.DiskID &&
+			r.Spec.DiskPath == replica.Spec.DiskPath &&
+			r.Spec.DataDirectoryName == replica.Spec.DataDirectoryName {
 			return true
 		}
 	}
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -2688,8 +2688,42 @@ func (c *VolumeController) upgradeEngineForVolume(v *longhorn.Volume, es map[str
 		return nil
 	}
 
-	// only start live upgrade if volume is healthy
+	// Only start live upgrade if volume is healthy
 	if v.Status.State != longhorn.VolumeStateAttached || v.Status.Robustness != longhorn.VolumeRobustnessHealthy {
+		if v.Status.State != longhorn.VolumeStateAttached || v.Status.Robustness != longhorn.VolumeRobustnessDegraded {
+			return nil
+		}
+		// Clean up inactive replica corresponding to non-existing active replica.
+		// This will allow volume controller to schedule and rebuild a new active replica
+		// and make the volume healthy again for the live upgrade to be able to continue.
+		// https://github.com/longhorn/longhorn/issues/7012
+
+		// sort the replica by name so that we don't select multiple replicas when running this logic repeatedly quickly
+		var sortedReplicaNames []string
+		for rName := range rs {
+			sortedReplicaNames = append(sortedReplicaNames, rName)
+		}
+		sort.Strings(sortedReplicaNames)
+
+		var inactiveReplicaToCleanup *longhorn.Replica
+		for _, rName := range sortedReplicaNames {
+			r := rs[rName]
+			if !r.Spec.Active && !hasMatchingActiveReplica(r, rs) {
+				inactiveReplicaToCleanup = r
+				break
+			}
+		}
+		if inactiveReplicaToCleanup != nil {
+			log.Infof("Cleaning up inactive replica %v which doesn't have corresponding active replica", inactiveReplicaToCleanup.Name)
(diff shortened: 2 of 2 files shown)


PhanLe1010:
e2e test for engine upgrade tests https://ci.longhorn.io/job/private/job/longhorn-tests-regression/5383/

controller/volume_controller.go
```diff
@@ -2688,11 +2688,59 @@ func (c *VolumeController) upgradeEngineForVolume(v *longhorn.Volume, es map[str
...
+					r.Spec.DiskID == replica.Spec.DiskID &&
+					r.Spec.DiskPath == replica.Spec.DiskPath &&
+					r.Spec.DataDirectoryName == replica.Spec.DataDirectoryName {
+					return true
+				}
+			}
+			return false
+		}
+
+		// sort the replica by name so that we don't select multiple replicas when running this logic repeatedly quickly
```
ejweber, line 2713: Can you clarify this? I wasn't able to fully understand why the sorting was necessary.

controller/volume_controller.go
```diff
@@ -2688,11 +2688,59 @@ func (c *VolumeController) upgradeEngineForVolume(v *longhorn.Volume, es map[str
...
+		var inactiveReplicaToCleanup *longhorn.Replica
+		for _, rName := range sortedReplicaNames {
+			r := rs[rName]
+			if !r.Spec.Active && !hasCorrespondingActiveReplica(r) {
+				inactiveReplicaToCleanup = r
+				break
+			}
+		}
+		if inactiveReplicaToCleanup != nil {
+			log.Infof("cleaning up inactive replica %v which doesn't have corresponding active replica", inactiveReplicaToCleanup.Name)
```
ejweber, line 2729: NIT: I think our convention is to start log messages capitalized: https://github.com/longhorn/longhorn/wiki/coding-convention#wording.

controller/volume_controller.go
```diff
@@ -2688,11 +2688,59 @@ func (c *VolumeController) upgradeEngineForVolume(v *longhorn.Volume, es map[str
...
+			return nil
+		}
+
+		// When the volume is degraded, even though we don't start the live engine upgrade, try to mark it as finished if it already started
+		c.finishLiveEngineUpgrade(v, e, rs, log)
+
 		return nil
 	}
 
+	// Only start live upgrade if volume is healthy
```
ejweber, line 2742: NIT: This looks out of place.

ejweber (commented):
LGTM in general. Just a couple of NITs and a non-blocking question.

controller/volume_controller.go
james-munson: I think it is the same reason as the other uses of sort in this controller - if we run this code twice with the same set of replicas, we want it to be idempotent, so we better choose the same inactive replica to delete.

james-munson (commented):
Beyond Eric's notes, LGTM.

controller/volume_controller.go
PhanLe1010: Yeah, if there are multiple inactive replicas to be deleted, I would want to delete the same one, one by one, at a time. Effectively slowing down the replica deletion.

controller/volume_controller.go

controller/volume_controller.go

james-munson (approved):

ejweber (dismissed):
LGTM.

PhanLe1010:
@shuo-wu @c3y1huang Could you also take a look before merging?

controller/replica_controller.go
```diff
@@ -240,9 +240,9 @@ func (rc *ReplicaController) syncReplica(key string) (err error) {
 			return errors.Wrapf(err, "failed to cleanup the related replica instance before deleting replica %v", replica.Name)
 		}
 
-		rs, err := rc.ds.ListReplicasByNodeRO(replica.Spec.NodeID)
+		rs, err := rc.ds.ListVolumeReplicas(replica.Spec.VolumeName)
```
ejweber, line 243: I think the change here is fine in general, but it looks like we can use `ListVolumeReplicasRO` for efficiency.

ejweber (changes requested):
I like the generalized use of the new `hasMatchingActiveReplicas` function. One thing to check.

controller/replica_controller.go
PhanLe1010: Good idea! Thank you. Fixed

innobead (approved):
LGTM

PhanLe1010:
@mergify backport v1.5.x

PR #2377, opened 2023-12-18
feat(metrics): collect snapshot size

c3y1huang:
https://github.com/longhorn/longhorn/issues/5869

### DIFF
--- a/metrics_collector/snapshot_collector.go
+++ b/metrics_collector/snapshot_collector.go
@@ -1,0 +1,80 @@
+package metricscollector
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/sirupsen/logrus"
+
+	"k8s.io/apimachinery/pkg/labels"
+
+	"github.com/longhorn/longhorn-manager/datastore"
+)
+
+type SnapshotCollector struct {
+	*baseCollector
+
+	sizeMetric metricInfo
+}
+
+func NewSnapshotCollector(
+	logger logrus.FieldLogger,
+	nodeID string,
+	ds *datastore.DataStore) *SnapshotCollector {
+
+	snapshotCollector := &SnapshotCollector{
+		baseCollector: newBaseCollector(subsystemSnapshot, logger, nodeID, ds),
+	}
+
+	snapshotCollector.sizeMetric = metricInfo{
+		Desc: prometheus.NewDesc(
+			prometheus.BuildFQName(longhornName, subsystemSnapshot, "actual_size_bytes"),
+			"Actual size of this snapshot",
+			[]string{volumeLabel, userCreatedLabel},
+			nil,
+		),
+		Type: prometheus.GaugeValue,
+	}
+
+	return snapshotCollector
+}
+
+func (c *SnapshotCollector) Describe(ch chan<- *prometheus.Desc) {
+	ch <- c.sizeMetric.Desc
+}
+
+func (c *SnapshotCollector) Collect(ch chan<- prometheus.Metric) {
+	defer func() {
+		if err := recover(); err != nil {
+			c.logger.WithField("error", err).Warn("Panic during collecting metrics")
+		}
+	}()
+
+	snapshotLists, err := c.ds.ListSnapshotsRO(labels.Everything())
+	if err != nil {
+		c.logger.WithError(err).Warn("Error during scrape")
+		return
+	}
+
+	for _, snapshot := range snapshotLists {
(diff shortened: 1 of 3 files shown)


metrics_collector/backup_collector.go
```diff
@@ -38,6 +39,16 @@ func NewBackupCollector(
 		Type: prometheus.GaugeValue,
 	}
 
+	bc.sizeTotalMetric = metricInfo{
+		Desc: prometheus.NewDesc(
+			prometheus.BuildFQName(longhornName, subsystemBackup, "actual_size_total_bytes"),
+			"Sum of actual size of backups owned by this node",
```
innobead, line 45: Feel this metric is meaningless for us because there is no actual data stored on the node. 

Let's remove this metrics. WDYT?

innobead (changes requested):

PhanLe1010:
Could we check this issue together with this PR https://github.com/longhorn/longhorn-manager/pull/2346#issuecomment-1858649541? @c3y1huang

metrics_collector/snapshot_collector.go
```diff
@@ -0,0 +1,93 @@
...
+			snapshotSizeTotal.user.count++
+		} else {
+			snapshotSizeTotal.system.size += snapshot.Status.Size
+			snapshotSizeTotal.system.count++
+		}
+	}
+
+	ch <- prometheus.MustNewConstMetric(c.sizeTotalMetric.Desc, c.sizeTotalMetric.Type, float64(snapshotSizeTotal.user.size), strconv.Itoa(snapshotSizeTotal.user.count), "true")
+	ch <- prometheus.MustNewConstMetric(c.sizeTotalMetric.Desc, c.sizeTotalMetric.Type, float64(snapshotSizeTotal.system.size), strconv.Itoa(snapshotSizeTotal.system.count), "false")
+}
```
PhanLe1010, line 93: I think that we don't need to sump up the size in longhorn-manager because user can write Prometheus query to sume the data later. Proposing the metric definition like:
```go


type SnapshotCollector struct {
	*baseCollector

	sizeMetric metricInfo
}

func NewSnapshotCollector(
	logger logrus.FieldLogger,
	nodeID string,
	ds *datastore.DataStore) *SnapshotCollector {

	snapshotCollector := &SnapshotCollector{
		baseCollector: newBaseCollector(subsystemSnapshot, logger, nodeID, ds),
	}

	snapshotCollector.sizeMetric = metricInfo{
		Desc: prometheus.NewDesc(
			prometheus.BuildFQName(longhornName, subsystemSnapshot, "size_bytes"),
			"Actual space used by the snapshot",
			[]string{nodeLabel, volumeLabel, userCreatedLabel},
			nil,
		),
		Type: prometheus.GaugeValue,
	}

	return snapshotCollector
}


```

c3y1huang:
> Could we check this issue together with this PR [#2346 (comment)](https://github.com/longhorn/longhorn-manager/pull/2346#issuecomment-1858649541)? @c3y1huang

This is a good catch. Considering this could use some QA testing, so let's handle this at https://github.com/longhorn/longhorn/issues/7380.

PhanLe1010 (dismissed):
~~LGTM~~

metrics_collector/snapshot_collector.go
```diff
@@ -0,0 +1,80 @@
...
+		if snapshot.Name == "volume-head" {
+			continue
+		}
+
+		isUserCreated := "false"
+		if snapshot.Status.UserCreated {
+			isUserCreated = "true"
+		}
+
+		ch <- prometheus.MustNewConstMetric(c.sizeMetric.Desc, c.sizeMetric.Type, float64(snapshot.Status.Size), c.currentNodeID, volume.Name, isUserCreated)
```
PhanLe1010, line 78: Sorry, is the c.currentNodeID (nodelabel) useful to derive any conclusion? It seems that it is the longhorn-manager that is handling the volume CR but doesn't have any relation to this snapshot

c3y1huang: I am thinking we should just remove the `nodeLabel`, there is `snapshot.Status.OwnerID`, but doesn't look like Longhorn is using it. The user should be able to sum up the sizes with the volume label. What do you think?

PhanLe1010: Sounds good to me

PhanLe1010 (approved):
LGTM

c3y1huang:
Test result: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/5403/

innobead (approved):

c3y1huang:

PR #2378, opened 2023-12-18
Introduce guaranteed instance manager CPU setting for v2 volume

derekbit:
Longhorn/longhorn#7361

### DIFF
--- a/controller/controller_manager.go
+++ b/controller/controller_manager.go
@@ -148,18 +148,32 @@ func GetInstanceManagerCPURequirement(ds *datastore.DataStore, imName string) (*
 	}
 
 	cpuRequest := lhNode.Spec.InstanceManagerCPURequest
-	guaranteedCPUSettingName := types.SettingNameGuaranteedInstanceManagerCPU
 	if cpuRequest == 0 {
-		guaranteedCPUSetting, err := ds.GetSettingWithAutoFillingRO(guaranteedCPUSettingName)
-		if err != nil {
-			return nil, err
+		switch im.Spec.BackendStoreDriver {
+		case longhorn.BackendStoreDriverTypeV1:
+			guaranteedCPUSetting, err := ds.GetSettingWithAutoFillingRO(types.SettingNameGuaranteedInstanceManagerCPU)
+			if err != nil {
+				return nil, err
+			}
+			guaranteedCPUPercentage, err := strconv.ParseFloat(guaranteedCPUSetting.Value, 64)
+			if err != nil {
+				return nil, err
+			}
+			allocatableMilliCPU := float64(kubeNode.Status.Allocatable.Cpu().MilliValue())
+			cpuRequest = int(math.Round(allocatableMilliCPU * guaranteedCPUPercentage / 100.0))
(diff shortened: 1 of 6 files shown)


controller/controller_manager.go
```diff
@@ -148,18 +148,32 @@ func GetInstanceManagerCPURequirement(ds *datastore.DataStore, imName string) (*
...
 
 	cpuRequest := lhNode.Spec.InstanceManagerCPURequest
-	guaranteedCPUSettingName := types.SettingNameGuaranteedInstanceManagerCPU
 	if cpuRequest == 0 {
-		guaranteedCPUSetting, err := ds.GetSettingWithAutoFillingRO(guaranteedCPUSettingName)
-		if err != nil {
-			return nil, err
+		switch im.Spec.BackendStoreDriver {
+		case longhorn.BackendStoreDriverTypeV1:
+			guaranteedCPUSetting, err := ds.GetSettingWithAutoFillingRO(types.SettingNameGuaranteedInstanceManagerCPU)
```
innobead, line 154: todo: Need to update the doc to mention this is for v1 only.

types/types.go
```diff
@@ -1008,10 +1008,20 @@ func ValidateCPUReservationValues(instanceManagerCPUStr string) error {
...
+		isOverLimit := instanceManagerCPU > 40
+		if isUnderLimit || isOverLimit {
+			return fmt.Errorf("invalid requested v1 data engine instance manager CPUs. Valid instance manager CPU range between 0%% - 40%%")
+		}
+	case SettingDefinitionV2DataEngineGuaranteedInstanceManagerCPU.Default:
+		isUnderLimit := instanceManagerCPU < 0
+		isOverLimit := instanceManagerCPU > 2000
+		if isUnderLimit || isOverLimit {
+			return fmt.Errorf("invalid requested v2 data engine instance manager CPUs. Valid instance manager CPU range between 0 - 2000 millicpu")
+		}
```
innobead, line 1024: Why passing `instanceManagerCPUStr`? Don't we need to pass the setting `instanceManagerCpuSetting` instead? Then, we are able to do a different range check for it.

In the current code, if `instanceManagerCPUStr` is `20`, there is no `case` able to handle it?

innobead (changes requested):

types/types.go
derekbit: > Why passing instanceManagerCPUStr? Don't we need to pass the setting instanceManagerCpuSetting instead?

Yes, passing `instanceManagerCpuSetting` is also feasible.  I reuse the existing function here, but we can improve it in another ticket later.

> In the current code, if instanceManagerCPUStr is 20, there is no case able to handle it?

For an acceptable value, the function returns nil error and the value will be updated in https://github.com/longhorn/longhorn-manager/blob/master/datastore/longhorn.go#L112

controller/controller_manager.go
derekbit: Doc is added. Thank you.
https://github.com/longhorn/website/pull/825

derekbit:
Handled in https://github.com/longhorn/longhorn-manager/pull/2386

PR #2379, opened 2023-12-18
feat: customize MaxRecurringJobRetain

mantissahz:
Add a global setting to allow users to customize the maximum retain number of the snapshot or backup for a recurring job.

This number should not exceed the maximum number of snapshots (250).

Ref: longhorn/longhorn#5713

### DIFF
--- a/types/setting.go
+++ b/types/setting.go
@@ -27,6 +27,9 @@ const (
 
 	ValueEmpty   = "none"
 	ValueUnknown = "unknown"
+
+	// From `maximumChainLength` in longhorn-engine/pkg/replica/replica.go
+	maxSnapshotNum = 250
 )
 
 type SettingType string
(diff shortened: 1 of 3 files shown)


types/setting.go
```diff
@@ -94,6 +97,7 @@ const (
 	SettingNameFailedBackupTTL                                          = SettingName("failed-backup-ttl")
 	SettingNameRecurringSuccessfulJobsHistoryLimit                      = SettingName("recurring-successful-jobs-history-limit")
 	SettingNameRecurringFailedJobsHistoryLimit                          = SettingName("recurring-failed-jobs-history-limit")
+	SettingNameRecurringJobMaxRetain                                    = SettingName("recurring-job-max-retain")
```
innobead, line 100: Retain is verb, let's use `retention` instead. Please change the var name and description.

types/setting.go
```diff
@@ -956,6 +962,16 @@ var (
 		Default:  "1",
 	}
 
+	SettingDefinitionRecurringJobMaxRetain = SettingDefinition{
+		DisplayName: "Recurring Job Maximum Retain Number",
```
innobead, line 966: Change the display name to `Maximum Retention Number for Recurring Job`.

innobead (changes requested):
In general, LGTM. Just two comments need to resolve first.

types/setting.go

innobead (approved):

innobead:
@mantissahz Make sure we have a PR to update longhorn/longhorn repo as well.

PR #2380, opened 2023-12-18
ci: refactor node controller unit tests

m-ildefons:
Refactor node controller unit tests:
- Break up single test function into multiple test functions
- Create a clear separation between test case input data and expected result data
- Implement existing tests with new structure as individual test functions
- Avoid checking things that aren't subject matter of the test case at hand

related-to: longhorn/longhorn#7332

### DIFF
--- a/controller/node_controller_test.go
+++ b/controller/node_controller_test.go
@@ -2,52 +2,1337 @@ package controller
 
 import (
 	"context"
-	"fmt"
 
+	monitor "github.com/longhorn/longhorn-manager/controller/monitor"
+	"github.com/longhorn/longhorn-manager/datastore"
+	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
+	lhfake "github.com/longhorn/longhorn-manager/k8s/pkg/client/clientset/versioned/fake"
+	"github.com/longhorn/longhorn-manager/types"
+	"github.com/longhorn/longhorn-manager/util"
 	"github.com/sirupsen/logrus"
-
-	"k8s.io/client-go/kubernetes/fake"
-	"k8s.io/client-go/kubernetes/scheme"
-	"k8s.io/client-go/tools/record"
-	"k8s.io/kubernetes/pkg/controller"
-
 	corev1 "k8s.io/api/core/v1"
 	apiextensionsfake "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset/fake"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	clientset "k8s.io/client-go/kubernetes"
-
-	"github.com/longhorn/longhorn-manager/datastore"
-	"github.com/longhorn/longhorn-manager/types"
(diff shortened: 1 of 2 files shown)


controller/node_controller_test.go
```diff
@@ -2,132 +2,169 @@ package controller
...
+	nodeIndexer cache.Indexer
+
+	controller *NodeController
 }
 
-func fakeTopologyLabelsChecker(kubeClient clientset.Interface, vers string) (bool, error) {
-	return false, nil
+// This data type contains resource that exist in the cluster environment, like
+// nodes and pods
+type TestCase struct {
```
PhanLe1010, line 78: Should we name this struct to `NodeTestCase`? I think if we adopt this new test patent for other controllers in this package, the generic `TestCase` naming will be conflicted later on

controller/node_controller_test.go
```diff
@@ -621,233 +1017,349 @@ func (s *TestSuite) TestSyncNode(c *C) {
...
+		condition.LastTransitionTime = ""
+		condition.Message = ""
+		node.Status.Conditions[idx] = condition
+	}
+	c.Assert(node.Status.Conditions, DeepEquals, exp.nodeStatus[node.Name].Conditions)
+}
 
-		nIndexer := informerFactories.LhInformerFactory.Longhorn().V1beta2().Nodes().Informer().GetIndexer()
-		pIndexer := informerFactories.KubeInformerFactory.Core().V1().Pods().Informer().GetIndexer()
+func (s *NodeControllerSuite) checkDiskConditions(c *C, tc *TestCase, exp *Expectation, node *longhorn.Node) {
```
PhanLe1010, line 1225: Looks like we don't call this function in any the test cases. Is it expected?

controller/node_controller_test.go
```diff
@@ -2,132 +2,169 @@ package controller
...
-	return map[string]*corev1.Node{
-		TestNode1: kubeNode1,
-		TestNode2: kubeNode2,
-	}
+// This data type contains expected results in the form of resources. Each test
+// will set up fixtures, input resources and expected results. Then it will
+// initialize the mock API with initTest, execute a function of the controller
+// and finally execute a set of assertions, which compare the actual contents of
+// the mock API to the expected results
+type Expectation struct {
```
PhanLe1010, line 93: Same as above, should we chane the name to `NodeControllerTestExpectation`?

PhanLe1010 (approved):
I very like the new test design pattern! 
In general LGTM expect for some small nits

controller/node_controller_test.go
m-ildefons: I'll rename it to `NodeControllerFixture` and `NodeControllerExpectation` respectively. If the other test suites also match this structure, we don't need to maintain separate fixture and expectation data structures for each controller, but until I've looked at other tests this makes a lot of sense.

controller/node_controller_test.go
m-ildefons: Good spot, this needs to be called for all tests that have non-empty disk conditions in their expectations.

PhanLe1010 (approved):
LGTM

PR #2386, opened 2023-12-19
Support v1 and v2 data engines enablement and disablement independently

derekbit:
longhorn/longhorn#7095


webhook/resources/replica/validator.go
```diff
@@ -42,19 +42,20 @@ func (r *replicaValidator) Resource() admission.Resource {
 }
 
 func (r *replicaValidator) Create(request *admission.Request, newObj runtime.Object) error {
-	replica, ok := newObj.(*longhorn.Replica)
-	if !ok {
-		return werror.NewInvalidError(fmt.Sprintf("%v is not a *longhorn.Replica", newObj), "")
-	}
+	replica := newObj.(*longhorn.Replica)
```
innobead, line 45: Add nil check.

datastore/longhorn.go
```diff
@@ -402,21 +415,72 @@ func (s *DataStore) ValidateSetting(name, value string) (err error) {
...
+		}
+
+		if !dataEngineEnabled {
+			for _, disk := range lhNode.Spec.Disks {
+				if disk.Type == longhorn.DiskTypeFilesystem {
+					return fmt.Errorf("failed to disable %v setting because there are filesystem-type disks on node %v", types.SettingNameV2DataEngine, node.Name)
+				}
+			}
+		}
+	}
```
innobead, line 464: Why do we need to check instance manager pods?
Why do users need to remove all longhorn FS disks before disabling the v1 engine?

Don't we just let everything intact instead? All disks and data inside still persist even though the engine has been disabled.

The same questions for v2 engine.

controller/setting_controller.go
```diff
@@ -804,36 +804,48 @@ func (sc *SettingController) updateLogLevel() error {
...
+	if !enabled {
+		return sc.cleanupInstanceManager(backednStoreDriver)
 	}
 
 	return nil
 }
 
-func (sc *SettingController) cleanupInstanceManagerForV2DataEngine() error {
-	imMap, err := sc.ds.ListInstanceManagersBySelectorRO("", "", longhorn.InstanceManagerTypeAllInOne, longhorn.BackendStoreDriverTypeV2)
+func (sc *SettingController) cleanupInstanceManager(backendStoreDriver longhorn.BackendStoreDriverType) error {
```
innobead, line 836: Let's add an info log here to highlight this operation.

innobead (changes requested):

datastore/longhorn.go
derekbit: Why do we need to check instance manager pods?
> Why do users need to remove all longhorn FS disks before disabling the v1 engine?

The check for v1 engine is not required, but it is necessary for v2 engine because the disk is managed by spdk_tgt.

controller/setting_controller.go
derekbit: Added. Thank you.

datastore/longhorn.go
derekbit: Removed the unnecessary checks. Thank you.

webhook/resources/replica/validator.go
derekbit: Added. Thank you.

datastore/longhorn.go
```diff
@@ -402,21 +415,35 @@ func (s *DataStore) ValidateSetting(name, value string) (err error) {
 	return nil
 }
 
-func (s *DataStore) ValidateV2DataEngine(v2DataEngineEnabled bool) error {
-	if !v2DataEngineEnabled {
-		allV2VolumesDetached, err := s.AreAllV2VolumesDetached()
+func (s *DataStore) ValidateV1DataEngine(dataEngineEnabled bool) error {
```
innobead, line 418: nit: `ValidateV1DataEngineEnabled`

the same for v2.

datastore/longhorn.go
```diff
@@ -499,7 +526,7 @@ func (s *DataStore) AreAllVolumesDetached() (bool, error) {
 	return true, nil
 }
 
-func (s *DataStore) AreAllV2VolumesDetached() (bool, error) {
+func (s *DataStore) AreAllVolumesDetached() (bool, error) {
```
innobead, line 529: In `AreAllVolumesDetached`, it filters the pods backed by `longhorn.BackendStoreDriverTypeV2` only?

webhook/resources/volume/validator.go
```diff
@@ -126,19 +126,17 @@ func (v *volumeValidator) Create(request *admission.Request, newObj runtime.Obje
...
+		dataEngineSetting = types.SettingNameV2DataEngine
+	}
+	enabled, err := v.ds.GetSettingAsBool(dataEngineSetting)
+	if err != nil {
+		err = errors.Wrapf(err, "failed to get %v setting", dataEngineSetting)
+		return werror.NewInvalidError(err.Error(), "")
+	}
+	if !enabled {
+		return werror.NewInvalidError(fmt.Sprintf("%v data engine is not enabled", dataEngineSetting), "")
 	}
```
innobead, line 140: nit: The validation in volume/engine/replica is almost the same. Suggest extracting that part into `webhook/resources` package instead.

innobead (changes requested):

datastore/longhorn.go
derekbit: It's wrong. Refactor AreAllVolumesDetached and remove AreAllVolumesDetachedByBackendStoreDriver. Thank you.

datastore/longhorn.go
derekbit: Renamed. Thank you.

webhook/resources/volume/validator.go
derekbit: Addressed. Thank you.

datastore/longhorn.go
```diff
@@ -348,36 +348,49 @@ func (s *DataStore) ValidateSetting(name, value string) (err error) {
...
-			guaranteedInstanceManagerCPU.Value = value
-		}
-		if err := types.ValidateCPUReservationValues(guaranteedInstanceManagerCPU.Value); err != nil {
+		guaranteedInstanceManagerCPU.Value = value
+		if err := types.ValidateCPUReservationValues(sName, guaranteedInstanceManagerCPU.Value); err != nil {
 			return err
 		}
 	case types.SettingNameStorageNetwork:
-		volumesDetached, err := s.AreAllVolumesDetached()
+		volumesDetached, err := s.AreAllVolumesDetached("")
```
innobead, line 361: nit: Using `"""` not quite intuitive. I would suggest having a specific type for that, since unfortunately, golang doesn't support default value ..

```go
	BackendStoreDriverTypeAll = BackendStoreDriverType("all")
```

Or even have a dedicated method for it `AreAllVolumesDetached` and `AreAllVolumesDetachedByDataEngine`. BTW, as discussed before, we should make a consistent name here `DataEngine` or `BackendStore`. IIRC, you decided to use `DataEngine`

datastore/longhorn.go
```diff
@@ -471,19 +498,18 @@ func (s *DataStore) ValidateV2DataEngine(v2DataEngineEnabled bool) error {
 	return nil
 }
 
-func (s *DataStore) AreAllVolumesDetached() (bool, error) {
+func (s *DataStore) AreAllVolumesDetached(backendStoreDriver longhorn.BackendStoreDriverType) (bool, error) {
```
innobead, line 501: nit: change `longhorn.BackendStoreDriverType` to `longhorn.DataEngineType`.

webhook/common/common.go
```diff
@@ -89,3 +89,19 @@ func GetLonghornLabelsPatchOp(obj runtime.Object, requiredLabels, removingLabels
 
 	return fmt.Sprintf(`{"op": "replace", "path": "/metadata/labels", "value": %v}`, string(bytes)), nil
 }
+
+func IsRequiredDataEngineEnabled(ds *datastore.DataStore, backendStoreDriver longhorn.BackendStoreDriverType) error {
```
innobead, line 93: `Is` function but did not return bool, so I would suggest renaming it to `ValidateRequiredDataEngineEnabled` or `CheckRequiredDataEngineEnabled`

innobead (changes requested):
LGTM. a few feedback. 

for nit comments, we can improve them later.

datastore/longhorn.go
derekbit: > BTW, as discussed before, we should make a consistent name here DataEngine or BackendStore. IIRC, you decided to use DataEngine

Rename BackendStoreDriver to DataEngine seems not quite easy... I prabaly continue using BackendStoreDriver

datastore/longhorn.go
derekbit: There are CRD and proto fields using BackendStoreDriver. How about still using BackendStoreDriver?

datastore/longhorn.go
innobead: That's OK. However, in the future, it's better to align the naming.

datastore/longhorn.go
innobead: oh, then it's not easy for now. Let's improve this later.

innobead: cc @c3y1huang

innobead (approved):

innobead:
Waiting for @derekbit ready first before merging.

derekbit:
> Waiting for @derekbit ready first before merging.

I introduced some new commits in the PR

1. Rename `backendStoreDriver` to `dataEngine` in longhorn-manager side.
2. Fix that v2 volumes are unable to attach caused by enforce-engine-upgrade feature (cc @c3y1huang )
3. Remove the requirement of the removal block-type disks before disabling v2-data-engine

cc @innobead

innobead (changes requested):

webhook/resources/volume/validator.go
```diff
@@ -231,90 +222,86 @@ func (v *volumeValidator) Update(request *admission.Request, oldObj runtime.Obje
...
 
 		if oldVolume.Spec.ReplicaZoneSoftAntiAffinity != newVolume.Spec.ReplicaZoneSoftAntiAffinity {
-			err := fmt.Errorf("changing replica zone soft anti-affinity for volume %v is not supported for backend store driver %v",
-				newVolume.Name, newVolume.Spec.BackendStoreDriver)
+			err := fmt.Errorf("changing replica zone soft anti-affinity for volume %v is not supported for data engine %v",
+				newVolume.Name, newVolume.Spec.DataEngine)
 			return werror.NewInvalidError(err.Error(), "")
 		}
 
 		if oldVolume.Spec.ReplicaDiskSoftAntiAffinity != newVolume.Spec.ReplicaDiskSoftAntiAffinity {
```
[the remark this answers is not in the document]
derekbit: Second `if` (`if oldVolume.Spec.ReplicaDiskSoftAntiAffinity != "" && newVolume.Spec.ReplicaDiskSoftAntiAffinity != longhorn.ReplicaDiskSoftAntiAffinityDefault`) is for recent change in the upgrade path (https://github.com/longhorn/longhorn-manager/blob/master/upgrade/v15xto160/upgrade.go#L65).

controller/setting_controller.go
```diff
@@ -804,36 +804,43 @@ func (sc *SettingController) updateLogLevel() error {
...
-	imMap, err := sc.ds.ListInstanceManagersBySelectorRO("", "", longhorn.InstanceManagerTypeAllInOne, longhorn.BackendStoreDriverTypeV2)
+func (sc *SettingController) cleanupInstanceManager(dataEngine longhorn.DataEngineType) error {
+	sc.logger.Infof("Cleaning up the instance manager for %v data engine", dataEngine)
+	imMap, err := sc.ds.ListInstanceManagersBySelectorRO("", "", longhorn.InstanceManagerTypeAllInOne, dataEngine)
 	if err != nil {
-		return err
+		return errors.Wrapf(err, "failed to list instance managers for cleaning up %v data engine", dataEngine)
 	}
 
 	for _, im := range imMap {
```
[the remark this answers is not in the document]
derekbit: Yes. Before calling the function, all volumes are detached and validated in `updateDataEngine()`.

controller/monitor/disk_monitor.go
```diff
@@ -178,31 +185,43 @@ func (m *NodeMonitor) getRunningInstanceManagerRO(backendStoreDriver longhorn.Ba
...
-	backendStoreDrivers := m.getBackendStoreDrivers()
+	dataEngines := m.getDataEngines()
 
-	for _, backendStoreDriver := range backendStoreDrivers {
-		im, err := m.getRunningInstanceManagerRO(backendStoreDriver)
-		if err != nil {
-			return clients, errors.Wrapf(err, "failed to get running instance manager for node %v", m.nodeName)
-		}
+	for _, dataEngine := range dataEngines {
+		var client *engineapi.DiskService
```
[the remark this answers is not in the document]
derekbit: Yes, removed it and enrich the error messages in a new commit. Thank you.

k8s/crds.yaml
```diff
@@ -1660,7 +1663,7 @@ spec:
           spec:
             description: InstanceManagerSpec defines the desired state of the Longhorn instancer manager
             properties:
-              backendStoreDriver:
```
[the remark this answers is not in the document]
derekbit: No. This is a newly introduced field in v1.6.0.

client/generated_replica.go
```diff
@@ -9,10 +9,10 @@ type Replica struct {
 
 	Address string `json:"address,omitempty" yaml:"address,omitempty"`
 
-	BackendStoreDriver string `json:"backendStoreDriver,omitempty" yaml:"backend_store_driver,omitempty"`
-
 	CurrentImage string `json:"currentImage,omitempty" yaml:"current_image,omitempty"`
 
+	DataEngine string `json:"dataEngine,omitempty" yaml:"data_engine,omitempty"`
```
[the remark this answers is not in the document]
derekbit: Yes, I've prepared a PR for it.
https://github.com/longhorn/longhorn-ui/pull/671

k8s/pkg/apis/longhorn/v1beta2/instancemanager.go
```diff
@@ -142,7 +148,7 @@ type InstanceManagerSpec struct {
 	// +optional
 	Type InstanceManagerType `json:"type"`
 	// +optional
-	BackendStoreDriver BackendStoreDriverType `json:"backendStoreDriver"`
```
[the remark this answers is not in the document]
derekbit: No, it is a newly introduced field in the v1.6.0 development, so we can remove it directly.

derekbit:
All comments are addressed cc @innobead

innobead (approved):

PR #2388, opened 2023-12-19
Add a startup probe to the longhorn-csi-plugin container

ejweber:
longhorn/longhorn#7116

If we can't get upstream changes in a reasonable amount of time, this works in the interim. It prevents a liveness probe from killing the `longhorn-csi-plugin` container for five minutes so we can get a successful liveness probe when the `livenessprobe` container comes back.

### DIFF
--- a/csi/deployment.go
+++ b/csi/deployment.go
@@ -344,6 +344,21 @@ func NewPluginDeployment(namespace, serviceAccount, nodeDriverRegistrarImage, li
 									Protocol:      corev1.ProtocolTCP,
 								},
 							},
+							StartupProbe: &corev1.Probe{
+								ProbeHandler: corev1.ProbeHandler{
+									HTTPGet: &corev1.HTTPGetAction{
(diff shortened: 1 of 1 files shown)


ejweber:
The downside here is that we prevent liveness probes from failing the csi-plugin container for five minutes (or until the first successful probe) even in the case that the livenessprobe container is healthy and really can't communicate with the csi-plugin container over the socket. I don't know what real world scenarios could cause that (I think it's a lot more likely that the csi-plugin container crashes on its own during startup), but we will be slow to respond if it happens.

ejweber:
Test steps:

Repeat https://github.com/longhorn/longhorn/issues/7116#issuecomment-1861814093.

When the service is restored, it takes a maximum of five minutes for the `longhorn-csi-plugin` container to run stable again. (If we are unlucky, `node-driver-registrar` may take longer.)

innobead (approved):

innobead:

PR #2389, opened 2023-12-20
fix(node): the deploy driver failed with linkerd

mantissahz:
Linkerd will inject a container `linkerd-proxy` in the longhorn-manger pod by adding the annotation `linkerd.io/inject: enabled`. It will fail when we check the `MountPropagation` for the first container that is `linkerd-proxy`. What we want to check is the `longhorn-manager` container.

Ref: longhorn/longhorn#3809

### DIFF
--- a/app/post_upgrade.go
+++ b/app/post_upgrade.go
@@ -3,6 +3,7 @@ package app
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/pkg/errors"
@@ -13,6 +14,7 @@ import (
 	"k8s.io/client-go/tools/clientcmd"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 
 	"github.com/longhorn/longhorn-manager/types"
 )
@@ -98,19 +100,25 @@ func (u *postUpgrader) waitManagerUpgradeComplete() error {
 			continue
 		}
 
-		podList, err := u.kubeClient.CoreV1().Pods(u.namespace).List(context.TODO(), metav1.ListOptions{})
+		podList, err := u.kubeClient.CoreV1().Pods(u.namespace).List(context.TODO(), metav1.ListOptions{LabelSelector: labels.Set(types.GetManagerLabels()).String()})
 		if err != nil {
 			logrus.Warnf("failed to list pods: %v", err)
 			continue
 		}
 		complete = true
+	PodLookupLoop:
 		for _, pod := range podList.Items {
-			if app, ok := pod.Labels["app"]; !ok || app != types.LonghornManagerDaemonSetName {
-				continue
-			}
-			if len(pod.Spec.Containers) != 1 || pod.Spec.Containers[0].Image != ds.Spec.Template.Spec.Containers[0].Image {
-				complete = false
-				break
+			newImage := ds.Spec.Template.Spec.Containers[0].Image
+			if len(pod.Spec.Containers) != 1 || pod.Spec.Containers[0].Image != newImage {
+				// we should disregard linkerd service or other sidecar containers that are not created by Longhorn
+				// https://github.com/longhorn/longhorn/issues/3809
+				for _, container := range pod.Spec.Containers {
+					if !strings.HasPrefix(container.Name, "longhorn") {
+						continue
+					}
+					complete = container.Image == newImage
+					break PodLookupLoop
+				}
 			}
 		}
 		if complete {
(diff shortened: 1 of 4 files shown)


controller/node_controller.go
```diff
@@ -931,7 +931,17 @@ func (nc *NodeController) updateDiskStatusSchedulableCondition(node *longhorn.No
 
 func (nc *NodeController) syncNodeStatus(pod *corev1.Pod, node *longhorn.Node) error {
 	// sync bidirectional mount propagation for node status to check whether the node could deploy CSI driver
-	for _, mount := range pod.Spec.Containers[0].VolumeMounts {
+	var longhornContainer *corev1.Container
```
innobead, line 934: How about?

```go
	var mgrContainer *corev1.Container
	for _, container := range pod.Spec.Containers {
		// TODO please add `type LonghornManagerContainerName = LonghornManagerDaemonSetName`
		if container.Name == types.LonghornManagerContainerName {
			mgrContainer = &container
			break
		}
	}
	
	if mgrContainer == nil {
		return fmt.Errorf("failed to find the %v container in the pod %v", types.LonghornManagerContainerName, pod.Name)
	}

	for _, mount := range mgrContainer.VolumeMounts {
		if mount.Name == types.LonghornSystemKey {
			mountPropagationStr := ""
			if mount.MountPropagation != nil {
				mountPropagationStr = string(*mount.MountPropagation)
			}

			if mountPropagationStr != corev1.MountPropagationBidirectional {
				node.Status.Conditions = types.SetCondition(node.Status.Conditions, longhorn.NodeConditionTypeMountPropagation, longhorn.ConditionStatusFalse,
					string(longhorn.NodeConditionReasonNoMountPropagationSupport),
					fmt.Sprintf("The MountPropagation value %s is not detected from pod %s, node %s", mountPropagationStr, pod.Name, pod.Spec.NodeName))
			} else {
				node.Status.Conditions = types.SetCondition(node.Status.Conditions, longhorn.NodeConditionTypeMountPropagation, longhorn.ConditionStatusTrue, "", "")
			}

			break
		}
	}
```

controller/volume_controller_test.go
```diff
@@ -1173,7 +1173,7 @@ func newDaemonPod(phase corev1.PodPhase, name, namespace, nodeID, podIP string,
 			NodeName: nodeID,
 			Containers: []corev1.Container{
 				{
-					Name:  "test-container",
+					Name:  "longhorn-manager",
```
innobead, line 1176: Does this mean the test did not work before? but why we did not notice any failed cases in PR checks?

innobead (changes requested):
In general, LGTM. However, the code can be improved.

controller/volume_controller_test.go
mantissahz: The test worked before because we always got the first container and did not check the name.
Because I modified the implementation to get the container named "longhorn-manager", it will fail if testing with other names.

app/post_upgrade.go
```diff
@@ -108,7 +109,18 @@ func (u *postUpgrader) waitManagerUpgradeComplete() error {
...
+				}
+				containerCount++
+				if container.Image != ds.Spec.Template.Spec.Containers[0].Image {
+					isOldImage = true
+				}
+			}
+			if containerCount != 1 || isOldImage {
 				complete = false
 				break
 			}
```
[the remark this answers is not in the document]
mantissahz: Modified. Thanks

controller/node_controller.go
mantissahz: Modified. Thanks

app/post_upgrade.go
```diff
@@ -104,13 +105,22 @@ func (u *postUpgrader) waitManagerUpgradeComplete() error {
...
+			if len(pod.Spec.Containers) != 1 || pod.Spec.Containers[0].Image != newImage {
+				for _, container := range pod.Spec.Containers {
+					if !strings.HasPrefix(container.Name, "longhorn") {
+						continue
+					}
+					if container.Image != newImage {
+						complete = false
+						break PodLookupLoop
+					}
+				}
```
c3y1huang, line 123: Consider adding comments to describe the reason for this condition check, and including an issue link would provide better context and traceability.

mantissahz: Added a comment to describe the reason and an issue link. Thanks

c3y1huang (dismissed):

app/post_upgrade.go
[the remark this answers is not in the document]
mantissahz: Modified getting Pod list as well and removed the check the pod labels.

innobead (approved):

mantissahz:
@Mergifyio backport v1.5.x

PR #2390, opened 2023-12-20
fix(telemetry): wrong average volume sizes due to V2 volume counts

c3y1huang:
https://github.com/longhorn/longhorn/issues/7380

### DIFF
--- a/controller/setting_controller.go
+++ b/controller/setting_controller.go
@@ -1541,6 +1541,12 @@ func (info *ClusterInfo) collectVolumesInfo() error {
 		return errors.Wrapf(err, "failed to list Longhorn Volumes")
 	}
 	volumeCount := len(volumesRO)
+	volumeCountV1 := 0
+	for _, volume := range volumesRO {
+		if volume.Spec.BackendStoreDriver == longhorn.BackendStoreDriverTypeV1 {
+			volumeCountV1++
+		}
+	}
 
 	var totalVolumeSize int
 	var totalVolumeActualSize int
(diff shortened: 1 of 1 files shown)


controller/setting_controller.go
```diff
@@ -1541,6 +1541,7 @@ func (info *ClusterInfo) collectVolumesInfo() error {
 		return errors.Wrapf(err, "failed to list Longhorn Volumes")
 	}
 	volumeCount := len(volumesRO)
+	volumeCountV1 := types.GetVolumeCountByBackendStoreDriver(longhorn.BackendStoreDriverTypeV1, volumesRO)
```
innobead, line 1544: I think `GetVolumeCountByBackendStoreDriver` is very specific for this usage and would not be used later, suggest adding here as a private function instead.

P.S. If needed, we can move it to a shared package later.

Or add `FilterVolumesByBackendStoreDriver` instead, which would be potentially reusable.. 

```go
volumeCountV1 := len(types.FilterVolumesByBackendStoreDriver(volumes, longhorn.BackendStoreDriverTypeV1))
```

innobead (changes requested):
LGTM, but just one comment needed to resolve.

controller/setting_controller.go
c3y1huang: I prefer removing the function as there are no other uses for either one at this point, and the logic is straightforward.

innobead (approved):

c3y1huang:
@mergify backport v1.5.x

PR #2393, opened 2023-12-21
Reject the creation of encrypted v2 volume in validating webhook

derekbit:
Longhorn/longhorn#7404


innobead (approved):

innobead (changes requested):
Actually there is a missing change causing the PR check failed.

derekbit:
> Actually there is a missing change causing the PR check failed.

Fixed. The error is due to the API in unmerged PRs.

innobead (approved):

PR #2395, opened 2023-12-21
Reject the last healthy replica deletion

derekbit:
To prevent accidentally from deleting the last replica regardless of that it is healthy or faulty, reject the last replica deletion if spec.deletionTimestamp of its volume is not set in the validating webhook.

Longhorn/longhorn#7372

### DIFF
--- a/webhook/resources/replica/validator.go
+++ b/webhook/resources/replica/validator.go
@@ -36,6 +36,7 @@ func (r *replicaValidator) Resource() admission.Resource {
 		OperationTypes: []admissionregv1.OperationType{
 			admissionregv1.Create,
 			admissionregv1.Update,
+			admissionregv1.Delete,
 		},
 	}
 }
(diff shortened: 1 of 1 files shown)


derekbit:
e2e
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/5446/

webhook/resources/replica/validator.go
```diff
@@ -70,3 +71,51 @@ func (r *replicaValidator) Update(request *admission.Request, oldObj runtime.Obj
...
+		}
+	}
+	if len(healthyReplicas) == 1 {
+		if _, ok := healthyReplicas[replica.Name]; !ok {
+			return nil
+		}
+		return fmt.Errorf("cannot delete replica %v because volume %v only has one available replica", replica.Name, volume.Name)
+	}
+
+	return nil
```
innobead, line 120: `To prevent accidentally from deleting the last replica regardless of it is healthy or faulty`

It seems we did not handle if the last replica is failed here?

innobead (changes requested):

webhook/resources/replica/validator.go
```diff
@@ -70,3 +71,51 @@ func (r *replicaValidator) Update(request *admission.Request, oldObj runtime.Obj
...
+	}
+
+	replicas, err := r.ds.ListVolumeReplicasRO(volume.Name)
+	if err != nil {
+		return errors.Wrapf(err, "failed to list replicas for volume %v before deleting replica", volume.Name)
+	}
+
+	healthyReplicas := map[string]struct{}{}
+	for _, r := range replicas {
+		if r.Spec.HealthyAt != "" {
```
james-munson, line 109: Why not use datastore.IsAvailableHealthreplica() here?

james-munson (commented):
One question.

webhook/resources/replica/validator.go
derekbit: If a volume only has one available replica whose `healthyAt != ""` but `failedAt != ""` , then we cannot allow the deletion. The replica might be the last hope of the volume.

webhook/resources/replica/validator.go
derekbit: `IsAvailableHealthyReplica()` checks failed `failedAt`, but the check is not required in the implementation. explained in https://github.com/longhorn/longhorn-manager/pull/2395#discussion_r1434597274.

webhook/resources/replica/validator.go
```diff
@@ -70,3 +71,51 @@ func (r *replicaValidator) Update(request *admission.Request, oldObj runtime.Obj
...
+		return nil
+	}
+
+	replicas, err := r.ds.ListVolumeReplicasRO(volume.Name)
+	if err != nil {
+		return errors.Wrapf(err, "failed to list replicas for volume %v before deleting replica", volume.Name)
+	}
+
+	healthyReplicas := map[string]struct{}{}
+	for _, r := range replicas {
```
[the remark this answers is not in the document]
derekbit: Good catch. Yes, need to check it as well. Thank you.

innobead (approved):
LGTM

derekbit:
@mergify backport v1.5.x

derekbit:
@mergify backport v1.4.x

PR #2397, opened 2023-12-21
Increase ignoreKubeletNotReadyTime to 15 seconds

ejweber:
longhorn/longhorn#7302

https://github.com/longhorn/longhorn-tests/pull/1629#issuecomment-1866933096


ejweber:

PR #2398, opened 2023-12-22
fix(metrics): snapshot size was collected before with the same name and label values

c3y1huang:
https://github.com/longhorn/longhorn/issues/5869#issuecomment-1865823767


c3y1huang:
Test result: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/5450/

innobead (approved):

PR #2400, opened 2023-12-22
[CI]: static analysis checks

m-ildefons:
- Introduce static analysis into the validation step of the build process, using the tool `staticcheck`

- Fix problem areas found by `staticcheck`. Most of the fixes are dead code removals and style issues, but there are also some potential nil dereferences and code duplication issues.
The fixes are broken up into many small commits making it easier to review and validate them.

Related to: https://github.com/longhorn/longhorn/issues/7300


PhanLe1010 (commented):
From the doc https://staticcheck.dev/docs/running-staticcheck/cli/

> At its core, the staticcheck command works a lot like go vet or go build

Could you give a big explanation of why we need the staticcheck instead of using go vet or go build?

controller/controller_test.go
```diff
@@ -626,12 +626,12 @@ func (s *TestSuite) TestIsSameGuaranteedCPURequirement(c *C) {
 	c.Assert(err, IsNil)
 	c.Assert(IsSameGuaranteedCPURequirement(a, b), Equals, true)
 
-	b.Requests[corev1.ResourceCPU], err = resource.ParseQuantity("250m")
+	b.Requests[corev1.ResourceCPU], _ = resource.ParseQuantity("250m")
```
c3y1huang, line 629: Suggest to `c.Assert(err, IsNil)` instead of overlooking it. The same applies to the code below.

datastore/longhorn.go
```diff
@@ -3171,15 +3174,13 @@ func (s *DataStore) GetInstanceManagerByInstanceRO(obj interface{}) (*longhorn.I
...
-		name = engine.Name
-		nodeID = engine.Spec.NodeID
+		name = obj.Name
+		nodeID = obj.Spec.NodeID
 	case *longhorn.Replica:
-		replica := obj.(*longhorn.Replica)
-		name = replica.Name
-		nodeID = replica.Spec.NodeID
+		name = obj.Name
+		nodeID = obj.Spec.NodeID
```
c3y1huang, line 3183: I am not sure if it's a good idea to redeclare `obj`, perhaps consider using a different name.

c3y1huang (changes requested):

c3y1huang (commented):
Also for context referencing, please add the issue number in the commit message. Thank you.

datastore/longhorn.go
m-ildefons: Good spot. Fixed it.

controller/controller_test.go
m-ildefons: Arguably, we should be able to assume that the `resource.ParseQuantity` can correctly parse the values given. It's not the point of our unit tests to make sure of that.
However I see the point that we should also make sure we pass parseable values to `resource.ParseQuantity`.

m-ildefons:
@PhanLe1010  the documentation you quoted is talking about how the command line interface is similar to `go vet` or `go build`.

`go build` is executing the actual compiler on the source file which is different that executing a linter. The compiler may accept inputs simply because they are legal within the language spec - even though there is a high likelyhood that they are buggy. A static analysis tool, like `go vet` or `staticcheck` is meant to catch such problems and alert on them.
However different static analysis tools may alert on different problems. `go vet` for example specializes in checking the usage of standard library functions like the `sync` or `atomic` libraries or the sanity of `printf` arguments.
`staticcheck` is the successor of `go lint` which was in use by Longhorn previously, but it's no longer maintained, but `staticcheck` also looks for quite a few more errors than `go lint` did, most notably unused code, scope of variables and [about 150 other potential problems](https://staticcheck.dev/docs/checks).

I hope this answers your question.

c3y1huang:
> Also for context referencing, please add the issue number in the commit message. Thank you.

@m-ildefons will you be including the issue numbers in each commit message? I suggest not using `longhorn/longhorn#7300` since it might flood the issue with commit notification messages.

derekbit (approved):
LGTM

derekbit:
@m-ildefons
Just reviewed the changes.
The commits remove the dead codes, check the unhandled variables, and fix incorrect formatting errors. Good job!
Can you help resolve the conflicts and address @c3y1huang's comment before merging the PR? 
Thank you.

derekbit:
@FrankYang0529 Can you help review this PR and see if we can close the PR? I think most of the fixes have been included in your golint fix PR. 

cc @innobead

derekbit:
The PR has been pending for a long time and without update. Most of the warnings and errors are fixed by @FrankYang0529. Let's close it. 

cc @innobead

PR #2401, opened 2023-12-22
ci: add tests for unknown node conditions

m-ildefons:
Add test ensuring that an unknown node condition does not emit a K8s event.

related-to: longhorn/longhorn#7290

### DIFF
--- a/controller/node_controller_test.go
+++ b/controller/node_controller_test.go
@@ -891,8 +891,7 @@ func (s *NodeControllerSuite) TestCreateDefaultInstanceManager(c *C) {
 			Conditions: []longhorn.Condition{
 				newNodeCondition(longhorn.DiskConditionTypeSchedulable, longhorn.ConditionStatusTrue, ""),
 			},
-			Type:   longhorn.DiskTypeFilesystem,
-			FSType: TestDiskPathFSType,
+			Type: longhorn.DiskTypeFilesystem,
 		},
 	}
 
@@ -1007,8 +1006,7 @@ func (s *NodeControllerSuite) TestCleanupRedundantInstanceManagers(c *C) {
 			Conditions: []longhorn.Condition{
 				newNodeCondition(longhorn.DiskConditionTypeSchedulable, longhorn.ConditionStatusTrue, ""),
 			},
-			Type:   longhorn.DiskTypeFilesystem,
-			FSType: TestDiskPathFSType,
+			Type: longhorn.DiskTypeFilesystem,
 		},
 	}
 
@@ -1290,7 +1288,7 @@ func (s *NodeControllerSuite) TestEventOnNotReady(c *C) {
 			"": {
 				Type:    "Warning",
 				Reason:  "Schedulable",
-				Message: "the disk fsid(/var/lib/longhorn) on the node test-node-name-1 has 0 available, but requires reserved 0, minimal 25% to schedule more replicas",
+				Message: "Disk fsid (/var/lib/longhorn) on the node test-node-name-1 has 0 available, but requires reserved 0, minimal 25% to schedule more replicas",
 			},
 			"node1-not-ready": {
 				Type:    "Warning",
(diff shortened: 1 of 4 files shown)


controller/node_controller_test.go
```diff
@@ -1213,6 +1223,101 @@ func (s *NodeControllerSuite) TestCleanupAllInstanceManagers(c *C) {
...
+	node2 := newKubernetesNode(
+		TestNode2,
+		corev1.ConditionTrue,
+		corev1.ConditionFalse,
+		corev1.ConditionFalse,
+		corev1.ConditionFalse,
+		corev1.ConditionFalse,
+		corev1.ConditionFalse,
+		corev1.ConditionTrue,
+	)
```
innobead, line 1255: Suggest covering all checked conditions.

[/controller/node_controller.go#L437-L464](https://github.com/longhorn/longhorn-manager/blob/c46a5456aacf3441f40eab70d95fc7d58ab230de/controller/node_controller.go#L437-L464)
```go
			case corev1.NodeReady:
				if con.Status != corev1.ConditionTrue {
					if con.Status == corev1.ConditionFalse &&
						time.Since(con.LastTransitionTime.Time) < ignoreKubeletNotReadyTime {
						// When kubelet restarts, it briefly reports Ready == False. Responding too quickly can cause
						// undesirable churn. See https://github.com/longhorn/longhorn/issues/7302 for an example.
						nc.logger.Warnf("Ignoring %v == %v condition due to %v until %v", corev1.NodeReady, con.Status,
							con.Reason, con.LastTransitionTime.Add(ignoreKubeletNotReadyTime))
					} else {
						node.Status.Conditions = types.SetConditionAndRecord(node.Status.Conditions,
							longhorn.NodeConditionTypeReady, longhorn.ConditionStatusFalse,
							string(longhorn.NodeConditionReasonKubernetesNodeNotReady),
							fmt.Sprintf("Kubernetes node %v not ready: %v", node.Name, con.Reason),
							nc.eventRecorder, node, corev1.EventTypeWarning)
					}
				}
			case corev1.NodeDiskPressure,
				corev1.NodePIDPressure,
				corev1.NodeMemoryPressure,
				corev1.NodeNetworkUnavailable:
				if con.Status == corev1.ConditionTrue {
					node.Status.Conditions = types.SetConditionAndRecord(node.Status.Conditions,
						longhorn.NodeConditionTypeReady, longhorn.ConditionStatusFalse,
						string(longhorn.NodeConditionReasonKubernetesNodePressure),
						fmt.Sprintf("Kubernetes node %v has pressure: %v, %v", node.Name, con.Reason, con.Message),
						nc.eventRecorder, node, corev1.EventTypeWarning)
				}
			}
```

innobead (changes requested):
In general, LGTM. Just one comment.

controller/node_controller_test.go
PhanLe1010: Looks like they are all already covered? https://github.com/m-ildefons/longhorn-manager/blob/b810121b33789d145f220bfd0e41102a7801a354/controller/volume_controller_test.go#L1250-L1280

PhanLe1010 (dismissed):
LGTM

controller/node_controller_test.go
m-ildefons: I think what David is referring to here is that we should have tests for each known node condition to make sure if events are emitted, they are exactly as expected. I.e. if there is a `MemoryPressure`, `DiskPressure`, etc. condition that is true, that we do see the corresponding "Kubernetest node X has pressure" event emitted by the Longhorn Manager. Likewise check that the manager emits events for the `NodeReady` status condition as expected as well.
This is not yet covered by the tests and there are other controllers that also emit events.

controller/node_controller_test.go
```diff
@@ -1233,6 +1221,559 @@ func (s *NodeControllerSuite) TestCleanupAllInstanceManagers(c *C) {
...
+		events: map[string]*corev1.Event{
+			"node1-ready": {
+				Type:    "Normal",
+				Reason:  "Ready",
+				Message: "Node test-node-name-1 is ready",
+			},
+			"node2-ready": {
+				Type:    "Normal",
+				Reason:  "Ready",
+				Message: "Node test-node-name-2 is ready",
```
innobead, line 1281: todo: it's better to improve the message in node controller as well, `Node .. is ready because longhorn manager pod is running`. The original message `Node is reday` is somehow confusing when the `kube node is not ready`.

innobead (approved):
LGTM

innobead:
Need to backport.

PR #2407, opened 2023-12-24
disk: only reject automatic creation of default block-type disk

derekbit:
Longhorn/longhorn#7385


types/types.go
```diff
@@ -966,11 +966,18 @@ func UnmarshalToNodeTags(s string) ([]string, error) {
...
 	if err != nil {
-		return nil, err
+		if !os.IsNotExist(err) {
+			return nil, errors.Wrapf(err, "failed to stat %v for creating default disk", dataPath)
+		}
+
+		// Longhorn is unable to create block-type disk automatically
+		if strings.HasPrefix(dataPath, "/dev/") {
+			return nil, errors.Wrapf(err, "creating default block-type disk %v is not supported", dataPath)
+		}
```
innobead, line 976: Will this block be reached? Or this is just in case?

innobead (approved):
LGTM

PR #2399, opened 2023-12-22
feat: replace golint with golangci-lint

FrankYang0529:
issue: https://github.com/longhorn/longhorn/issues/7366

Using `golangci-lint run --new-from-rev=HEAD~` because there are too many warning for current code.

Ref: https://golangci-lint.run/usage/faq/#how-to-integrate-golangci-lint-into-large-project-with-thousands-of-issues

PR #2408, opened 2023-12-25
[Test CI] replace golint with golangci-lint

FrankYang0529:


FrankYang0529:
Close this PR because we don't need it.

https://github.com/longhorn/longhorn-manager/pull/2399

PR #2411, opened 2023-12-26
feat(setting): lazy update for volume-related settings

mantissahz:
For settings `taint-toleration`, `guaranteed-instance-manager-cpu`, `system-managed-components-node-selector`, and `priority-class`, users can customize the settings, and the settings will only take effect when all volumes are detached.

Ref: longhorn/longhorn#7173

### DIFF
--- a/controller/instance_manager_controller.go
+++ b/controller/instance_manager_controller.go
@@ -453,7 +453,15 @@ func (imc *InstanceManagerController) handlePod(im *longhorn.InstanceManager) er
 		return err
 	}
 
-	if im.Status.CurrentState != longhorn.InstanceManagerStateError && im.Status.CurrentState != longhorn.InstanceManagerStateStopped {
+	isSettingSynced, isPodDeletedOrNotRunning, areInstancesRunningInPod, err := imc.areDangerZoneSettingsSyncedToIMPod(im)
+	if err != nil {
+		return err
+	}
+
+	isPodDeletionNotRequired := isSettingSynced || areInstancesRunningInPod || isPodDeletedOrNotRunning
+	if im.Status.CurrentState != longhorn.InstanceManagerStateError &&
+		im.Status.CurrentState != longhorn.InstanceManagerStateStopped &&
+		isPodDeletionNotRequired {
 		return nil
 	}
 
@@ -466,15 +474,14 @@ func (imc *InstanceManagerController) handlePod(im *longhorn.InstanceManager) er
 	}
 
 	// Since `spec.nodeName` is specified during the pod creation,
-	// the node cordon can not prevent the pod being launched.
-	if unschedulable, err := imc.ds.IsKubeNodeUnschedulable(im.Spec.NodeID); unschedulable || err != nil {
+	// the node cordon cannot prevent the pod being launched.
+	if unscheduled, err := imc.ds.IsKubeNodeUnschedulable(im.Spec.NodeID); unscheduled || err != nil {
 		return err
 	}
 
 	if err := imc.createInstanceManagerPod(im); err != nil {
 		return err
 	}
-	// The instance manager state will be updated in the next reconcile loop.
 
 	return nil
 }
@@ -513,6 +520,128 @@ func (imc *InstanceManagerController) annotateCASafeToEvict(im *longhorn.Instanc
 	}
 
 	return nil
+}
+
+func (imc *InstanceManagerController) areDangerZoneSettingsSyncedToIMPod(im *longhorn.InstanceManager) (isSynced, isPodDeletedOrNotRunning, areInstancesRunningInPod bool, err error) {
+	if im.Status.CurrentState != longhorn.InstanceManagerStateRunning {
+		return false, true, false, nil
+	}
+
+	// nolint:all
+	for _, instance := range types.ConsolidateInstances(im.Status.InstanceEngines, im.Status.InstanceReplicas, im.Status.Instances) {
+		if instance.Status.State == longhorn.InstanceStateRunning || instance.Status.State == longhorn.InstanceStateStarting {
+			return false, false, true, nil
+		}
+	}
+
+	pod, err := imc.ds.GetPod(im.Name)
+	if err != nil {
+		return false, false, false, errors.Wrapf(err, "cannot get pod for instance manager %v", im.Name)
+	}
+	if pod == nil {
+		return false, true, false, nil
+	}
+
+	for settingName := range types.GetDangerZoneSettings() {
+		isSettingSynced := true
+		setting, err := imc.ds.GetSettingWithAutoFillingRO(settingName)
+		if err != nil {
+			return false, false, false, err
+		}
+		switch settingName {
+		case types.SettingNameTaintToleration:
+			isSettingSynced, err = imc.isSettingTaintTolerationSynced(setting, pod)
+		case types.SettingNameSystemManagedComponentsNodeSelector:
+			isSettingSynced, err = imc.isSettingNodeSelectorSynced(setting, pod)
+		case types.SettingNameGuaranteedInstanceManagerCPU, types.SettingNameV2DataEngineGuaranteedInstanceManagerCPU:
+			isSettingSynced, err = imc.isSettingGuaranteedInstanceManagerCPUSynced(setting, pod)
+		case types.SettingNamePriorityClass:
+			isSettingSynced, err = imc.isSettingPriorityClassSynced(setting, pod)
+		case types.SettingNameStorageNetwork:
+			isSettingSynced, err = imc.isSettingStorageNetworkSynced(setting, pod)
+		case types.SettingNameV1DataEngine, types.SettingNameV2DataEngine:
+			isSettingSynced, err = imc.isSettingDataEngineSynced(settingName, im)
+		}
+		if err != nil {
+			return false, false, false, err
+		}
+		if !isSettingSynced {
+			return false, false, false, nil
+		}
+	}
+
+	return true, false, false, nil
+}
+
+func (imc *InstanceManagerController) isSettingTaintTolerationSynced(setting *longhorn.Setting, pod *corev1.Pod) (bool, error) {
+	newTolerationsList, err := types.UnmarshalTolerations(setting.Value)
+	if err != nil {
+		return false, err
+	}
+	newTolerationsMap := util.TolerationListToMap(newTolerationsList)
+	lastAppliedTolerations, err := getLastAppliedTolerationsList(pod)
+	if err != nil {
+		return false, err
+	}
+
+	return reflect.DeepEqual(util.TolerationListToMap(lastAppliedTolerations), newTolerationsMap), nil
+}
+
+func (imc *InstanceManagerController) isSettingNodeSelectorSynced(setting *longhorn.Setting, pod *corev1.Pod) (bool, error) {
+	newNodeSelector, err := types.UnmarshalNodeSelector(setting.Value)
+	if err != nil {
+		return false, err
+	}
+	if pod.Spec.NodeSelector == nil && len(newNodeSelector) == 0 {
+		return true, nil
+	}
+
+	return reflect.DeepEqual(pod.Spec.NodeSelector, newNodeSelector), nil
+}
+
+func (imc *InstanceManagerController) isSettingGuaranteedInstanceManagerCPUSynced(setting *longhorn.Setting, pod *corev1.Pod) (bool, error) {
+	lhNode, err := imc.ds.GetNode(pod.Spec.NodeName)
+	if err != nil {
+		return false, err
+	}
+	if types.GetCondition(lhNode.Status.Conditions, longhorn.NodeConditionTypeReady).Status != longhorn.ConditionStatusTrue {
+		return true, nil
+	}
+
+	resourceReq, err := GetInstanceManagerCPURequirement(imc.ds, pod.Name)
+	if err != nil {
+		return false, err
+	}
+	podResourceReq := pod.Spec.Containers[0].Resources
+	return IsSameGuaranteedCPURequirement(resourceReq, &podResourceReq), nil
+}
+
+func (imc *InstanceManagerController) isSettingPriorityClassSynced(setting *longhorn.Setting, pod *corev1.Pod) (bool, error) {
+	return pod.Spec.PriorityClassName == setting.Value, nil
+}
+
+func (imc *InstanceManagerController) isSettingStorageNetworkSynced(setting *longhorn.Setting, pod *corev1.Pod) (bool, error) {
+	nadAnnot := string(types.CNIAnnotationNetworks)
+
+	return pod.Annotations[nadAnnot] == setting.Value, nil
+}
+
+func (imc *InstanceManagerController) isSettingDataEngineSynced(settingName types.SettingName, im *longhorn.InstanceManager) (bool, error) {
+	enabled, err := imc.ds.GetSettingAsBool(settingName)
+	if err != nil {
+		return false, errors.Wrapf(err, "failed to get %v setting for updating data engine", settingName)
+	}
+	var dataEngine longhorn.DataEngineType
+	switch settingName {
+	case types.SettingNameV1DataEngine:
+		dataEngine = longhorn.DataEngineTypeV1
+	case types.SettingNameV2DataEngine:
+		dataEngine = longhorn.DataEngineTypeV2
+	}
+	if !enabled && im.Spec.DataEngine == dataEngine {
+		return false, nil
+	}
+	return true, nil
 }
 
 func (imc *InstanceManagerController) syncInstanceManagerAPIVersion(im *longhorn.InstanceManager) error {
--- a/controller/setting_controller.go
+++ b/controller/setting_controller.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"net/http"
 	"reflect"
+	"slices"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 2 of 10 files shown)


controller/setting_controller.go
```diff
@@ -223,10 +223,44 @@ func (sc *SettingController) syncSetting(key string) (err error) {
...
+		return nil
+	case string(types.SettingNameV2DataEngine):
+		if err := sc.updateV2DataEngine(); err != nil {
+			return err
+		}
+		return nil
+	}
+
+	// SettingCategoryDangerZone and they will try to delete pods to update the settings.
+	volumesDetached, err := sc.ds.AreAllVolumesDetached()
```
innobead, line 255: Please have a new `syncSettingForSystemManagedComponents`. 

Besides remaining settings here, also move `sc.updateV2DataEngine` there, but probably you need to refactor it a bit.

P.S. need to wait for https://github.com/longhorn/longhorn-manager/pull/2386 to merge first.

controller/setting_controller.go
```diff
@@ -223,10 +223,44 @@ func (sc *SettingController) syncSetting(key string) (err error) {
...
+	if err != nil {
+		return errors.Wrapf(err, "failed to check volume detachment for %v setting update", name)
+	}
+	if !volumesDetached {
+		return &types.ErrorInvalidState{Reason: fmt.Sprintf("failed to apply %v setting to Longhorn workloads when there are attached volumes", name)}
```
innobead, line 260: Don't throw error in this case, because it will be just requeued three times.

Have a debug log as well. 

[/controller/setting_controller.go#L194-L209](https://github.com/longhorn/longhorn-manager/blob/b517601153190d762c88e38442e8037ff6b94391/controller/setting_controller.go#L194-L209)
```go
func (sc *SettingController) handleErr(err error, key interface{}) {
	if err == nil {
		sc.queue.Forget(key)
		return
	}

	log := sc.logger.WithField("Setting", key)
	if sc.queue.NumRequeues(key) < maxRetries {
		handleReconcileErrorLogging(log, err, "Failed to sync Longhorn setting")
		sc.queue.AddRateLimited(key)
		return
	}

	utilruntime.HandleError(err)
	handleReconcileErrorLogging(log, err, "Dropping Longhorn setting out of the queue")
	sc.queue.Forget(key)
```

Instead, do AddRateLimited directly, similar to below. It will be eventually updated, but it could be a delay to take effect depending on the rate, even though all volumes have been detached by users. Therefore, we need to update the description of impacted settings to explain this eventual effect behavior. Also, the doc needs to be updated.

[/controller/node_controller.go#L580-L588](https://github.com/longhorn/longhorn-manager/blob/b517601153190d762c88e38442e8037ff6b94391/controller/node_controller.go#L580-L588)
```go
func (nc *NodeController) enqueueNodeRateLimited(obj interface{}) {
	key, err := controller.KeyFunc(obj)
	if err != nil {
		utilruntime.HandleError(fmt.Errorf("couldn't get key for object %#v: %v", obj, err))
		return
	}

	nc.queue.AddRateLimited(key)
}
```

innobead (changes requested):
The idea of this change is good. 

Some comments need to resolve first.

controller/setting_controller.go
mantissahz: Modified to have a new function `syncSettingForSystemManagedComponents`

controller/setting_controller.go
mantissahz: If we call the function e.g. `enqueueSettingRateLimited` to `AddRateLimited` and return `nil` back to `handleErr`.

```golang
func (sc *SettingController) handleErr(err error, key interface{}) {
	if err == nil {
		sc.queue.Forget(key)
		return
	}
```
It will forget the key, and another controller will get the item and then enqueue by `AddRateLimited` but it is not delayed actually.

Modified to get the error message and enqueue it by `sc.queue.AddRateLimited(key)` in `handleErr`.

mantissahz: >Did you mean the code in [/controller/node_controller.go#L580-L588](https://github.com/longhorn/longhorn-manager/blob/b517601153190d762c88e38442e8037ff6b94391/controller/node_controller.go#L580-L588) not working as expected?

It will retry to reconcile the node CR (test the replica eviction when updating replicas fails), but it will not delay the item enqueued as expected.
That's what I found when testing the codes by adding `enqueueSettingRateLimited' with  `queue.AddRateLimited(key)`.

innobead: > It will retry to reconcile the node CR (test the replica eviction when updating replicas fails), but it will not delay the item enqueued as expected.
> 

To be clear, didn't u mean the item enqueued Will immediately be forgotten during handling the nil error in the last reply?

What "delay" u meant? 

> That's what I found when testing the codes for retrying the Setting CR by adding `enqueueSettingRateLimited` with `queue.AddRateLimited(key)`.

So u found it on as requeued immediately but without rate limitation?

innobead:
This PR only partially tackle immediate setting update before waiting for all volumes detached first.

However, it missed a primary part:
> When changing any volume-related Pod configuration (ex: IM pod, share manager pod, upcoming object storage/s3gw pod) like priority class, the change will only take effect when doing volume upgrade or volume reattachement.

Let's continue this in LEP.

controller/setting_controller.go
mantissahz: >To be clear, didn't u mean the item enqueued immediately be forgotten (forget) during handling the nil error in the last reply?

Yes, IIUC, it will be forgotten if no other controller starts to handle the event.

>What "delay" did u mean?
So u found the event can be requeued immediately but without rate limitation?

Yes, and "delay" means `the event can be requeued immediately but without rate limitation`

controller/setting_controller.go
```diff
@@ -227,6 +234,49 @@ func (sc *SettingController) syncSetting(key string) (err error) {
...
+
+	return nil
+}
+
+func isDangerZoneSetting(settingName string) bool {
+	return settingName == string(types.SettingNameTaintToleration) || settingName == string(types.SettingNameSystemManagedComponentsNodeSelector) ||
+		settingName == string(types.SettingNameGuaranteedInstanceManagerCPU) || settingName == string(types.SettingNameV2DataEngineGuaranteedInstanceManagerCPU) ||
+		settingName == string(types.SettingNamePriorityClass) || settingName == string(types.SettingNameStorageNetwork) ||
+		settingName == string(types.SettingNameV1DataEngine) || settingName == string(types.SettingNameV2DataEngine)
+}
```
c3y1huang, line 268: We might be able to do:
```go
	settingDefinition, ok := types.GetSettingDefinition(types.SettingName(settingName))
	if !ok {
		// do something
	}
	return settingDefinition.Category == types.SettingCategoryDangerZone
```

controller/setting_controller.go
```diff
@@ -186,6 +188,11 @@ func (sc *SettingController) processNextWorkItem() bool {
 	defer sc.queue.Done(key)
 
 	err := sc.syncSetting(key.(string))
+	if err != nil && strings.Contains(err.Error(), lhWorkloadAttachedVolumeMsg) {
+		sc.logger.WithError(err).Debugf("Retrying to update setting %v", key)
+		sc.queue.AddRateLimited(key)
```
c3y1huang, line 193: What happens if the setting has been changed several times without ever detaching all volumes?

controller/setting_controller.go
```diff
@@ -227,6 +234,49 @@ func (sc *SettingController) syncSetting(key string) (err error) {
...
+		settingName == string(types.SettingNameV1DataEngine) || settingName == string(types.SettingNameV2DataEngine)
+}
+
+func (sc *SettingController) syncSettingForSystemManagedComponents(settingName string) error {
+	volumesDetached, err := sc.ds.AreAllVolumesDetached(longhorn.DataEngineTypeAll)
+	if err != nil {
+		return errors.Wrapf(err, "failed to check volume detachment for %v setting update", settingName)
+	}
+	if !volumesDetached {
+		return &types.ErrorInvalidState{Reason: fmt.Sprintf("failed to apply %v setting to %v", settingName, lhWorkloadAttachedVolumeMsg)}
```
c3y1huang, line 276: For the change to take effect, the user still needs to detach all volumes. So I am confused about the benefit of this feature. I originally expected this feature would enable the setting change to apply separately to each volume. For example:

```gherkin
Given a cluster with 2 volumes.
When a user updates the setting.
And the user re-attaches volume-1.
Then the setting should take effect on volume-1.
When the user re-attaches volume-2.
Then the setting should take effect on volume-2.
```

innobead: @c3y1huang That was the original expectation, but I doubt its feasibility in the current design.

The benefits should be two for the current implementation. @mantissahz Please correct this.
- During the upgrade, the setting can be applied and the new IM pod will start with the new setting, even though there are running volumes in the current IM pod. 
- After the upgrade, updating the setting will be allowed, even though there are running volumes, but it will only take effect until all volumes are detached. This would be a benefit for automation. However, the behavior is similar to the current behavior.

controller/setting_controller.go
mantissahz: Removed that returning errors to retry updating danger zone settings when there are volumes attached.

controller/setting_controller.go
mantissahz: Modified to compare with `settingDefinition.Category`.
Thanks.

controller/setting_controller.go
mantissahz: Without ever detaching all volumes, it should always retry to update with the latest changed value.

Removed this part. Thanks.

controller/volume_controller.go
```diff
@@ -1574,6 +1589,48 @@ func (c *VolumeController) reconcileAttachDetachStateMachine(v *longhorn.Volume,
 	return nil
 }
 
+func (c *VolumeController) syncIMSystemComponentSettingStatus(v *longhorn.Volume) error {
```
c3y1huang, line 1592: Why not let instance-manager controller handle this?

controller/setting_controller.go
```diff
@@ -227,40 +228,70 @@ func (sc *SettingController) syncSetting(key string) (err error) {
...
+func (sc *SettingController) isDangerZoneSetting(settingName string) bool {
+	settingDefinition, exists := types.GetSettingDefinition(types.SettingName(settingName))
+	if !exists {
+		sc.logger.Debugf("Setting %s does not exist", settingName)
+		return false
+	}
+	return settingDefinition.Category == types.SettingCategoryDangerZone || string(types.SettingNameV1DataEngine) == settingName || string(types.SettingNameV2DataEngine) == settingName
+}
+
+func (sc *SettingController) syncSettingForSystemManagedComponents(settingName string) error {
```
c3y1huang, line 266: Is there a way to identify the specific volume or component that is currently awaiting detachment for setting sync?

c3y1huang (changes requested):

controller/setting_controller.go
```diff
@@ -21,6 +21,7 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/sets"
```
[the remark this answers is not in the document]
mantissahz: Removed. It would not be used.

controller/setting_controller.go
```diff
@@ -849,6 +888,25 @@ func (sc *SettingController) cleanupInstanceManager(dataEngine longhorn.DataEngi
 	return nil
 }
 
+func (sc *SettingController) listActivatingIMNodes() (sets.Set[string], error) {
```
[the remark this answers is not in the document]
mantissahz: Renamed to `getIMsWithInstanceRunningMap` because changed to get instance manager with running processes.
Thanks.

controller/volume_controller.go
```diff
@@ -1574,6 +1589,48 @@ func (c *VolumeController) reconcileAttachDetachStateMachine(v *longhorn.Volume,
...
+	dangerZoneList := []types.SettingName{
+		types.SettingNameTaintToleration, types.SettingNameSystemManagedComponentsNodeSelector,
+		types.SettingNameGuaranteedInstanceManagerCPU, types.SettingNameV2DataEngineGuaranteedInstanceManagerCPU,
+		types.SettingNamePriorityClass, types.SettingNameStorageNetwork,
+		types.SettingNameV1DataEngine, types.SettingNameV2DataEngine,
+	}
```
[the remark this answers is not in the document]
mantissahz: Have a new function named `GetSettingForSystemManagedComponents` in types/setting.go to get the settings' name from the danger zone with the settings related to system components.
Thanks.

controller/volume_controller.go
mantissahz: Yes, that's a good idea.
Moved the setting synchronization in the instance manager monitor.
Thanks.

controller/setting_controller.go
mantissahz: Introduced a new field, `syncingIMsMap` to record the instance managers with running processes when the setting related to system components is modified.
Is this close to what you mentioned?

innobead:
@mantissahz Let's just need to do in 1.6.0.

https://github.com/longhorn/longhorn-manager/pull/2411#discussion_r1440604306

controller/setting_controller.go
```diff
@@ -849,6 +888,25 @@ func (sc *SettingController) cleanupInstanceManager(dataEngine longhorn.DataEngi
...
+	imList, err := sc.ds.ListInstanceManagersRO()
+	if err != nil {
+		return sets.Set[string]{}, errors.Wrap(err, "failed to list instance managers for toleration update")
```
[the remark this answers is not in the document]
mantissahz: Modified the message. Thanks

innobead:
I mean keeping your original implementation instead of introducing extra change to setting CR. Let's sync up later.

controller/setting_controller.go
mantissahz: Removed the new field.

controller/setting_controller.go
```diff
@@ -227,6 +227,43 @@ func (sc *SettingController) syncSetting(key string) (err error) {
 		if err := sc.syncBackupTarget(); err != nil {
 			return err
 		}
+	case string(types.SettingNameKubernetesClusterAutoscalerEnabled):
```
innobead, line 230: Suggest doing this in `syncSetting`:

```go
	defer func() {
		err = errors.Wrapf(err, "failed to sync setting for %v", key)
	}()

	_, name, err := cache.SplitMetaNamespaceKey(key)
	if err != nil {
		return err
	}

	err := sc.syncNonDangerZoneSettingsForSystemManagedComponents(name)
	if err != nil {
		return err
	}

	// You can return nil directly if the setting belongs to danger zone category 
	err := sc.syncDangerZoneSettingsForSystemManagedComponents(name)
	if err != nil {
		return err
	}

	return nil
```

controller/instance_manager_controller.go
```diff
@@ -515,6 +520,130 @@ func (imc *InstanceManagerController) annotateCASafeToEvict(im *longhorn.Instanc
...
+func (imc *InstanceManagerController) isSettingsSyncedToPod(im *longhorn.InstanceManager) (bool, error) {
+	if !im.DeletionTimestamp.IsZero() || im.Status.CurrentState != longhorn.InstanceManagerStateRunning {
+		return true, nil
+	}
+
+	for _, instance := range types.ConsolidateInstances(im.Status.InstanceEngines, im.Status.InstanceReplicas, nil) {
+		if instance.Status.State == longhorn.InstanceStateRunning || instance.Status.State == longhorn.InstanceStateStarting {
+			return true, nil
+		}
+	}
```
innobead, line 532: If `!im.DeletionTimestamp.IsZero() || im.Status.CurrentState != longhorn.InstanceManagerStateRunning`, do we still need to check whether there are instances running in the pod?

types/setting.go
```diff
@@ -1726,3 +1728,15 @@ func SetSettingDefinition(name SettingName, definition SettingDefinition) {
...
 }
+
+func GetSettingForSystemManagedComponents() sets.Set[SettingName] {
+	settingList := sets.New[SettingName]()
+	for settingName, setting := range settingDefinitions {
+		if setting.Category == SettingCategoryDangerZone {
+			settingList = settingList.Insert(settingName)
+		}
+	}
+	settingList = settingList.Insert(SettingNameV1DataEngine, SettingNameV2DataEngine)
```
innobead, line 1739: Move `SettingNameV1DataEngine, SettingNameV2DataEngine` to danger zone category.

innobead (changes requested):
In general, the concept LGTM. Some comments need to resolve first. 

Please reflect the change to your LEP.

controller/controller_test.go
```diff
@@ -236,6 +236,10 @@ func newPod(status *corev1.PodStatus, name, namespace, nodeID string) *corev1.Po
 	if status == nil {
 		return nil
 	}
+	var containers []corev1.Container
+	containers = append(containers, corev1.Container{
+		Name:      "instance-manager",
+		Resources: corev1.ResourceRequirements{Requests: corev1.ResourceList{"cpu": resource.MustParse("480m")}}})
```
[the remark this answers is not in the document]
mantissahz: Moved the container setting of pods to instance_manager_controller_test.go.
It is testing the setting `guaranteed-instance-manager-cpu`.

controller/setting_controller.go
mantissahz: Modified to separate the settings into two parts. Thanks.

controller/instance_manager_controller.go
mantissahz: `!im.DeletionTimestamp.IsZero()` will be handled at the beginning of `syncInstanceManager`, so it can be removed here.
No, we don't need to check running instances if `CurrentState != longhorn.InstanceManagerStateRunnin` and here we will return to the caller immediately.

types/setting.go
mantissahz: Moved `SettingNameV1DataEngine, SettingNameV2DataEngine, and SettingNameV2DataEngineGuaranteedInstanceManagerCPU` to the danger zone category.
Thanks.

innobead:
@mantissahz Please verify the e2e, the expected behavior should be.

- When a danger zone setting is updated, it will persist, even though there are attached volumes.
- The updated setting will not take effect if three are attached volumes. 3 cases here to apply the settings to user-deployed/system-managed pods.
  - It will be retried until being removed from the queue. 
  - Each setting will be revisited every 60 minutes (in the current setting controller resync time), so it will be automatically retried again. 
  - Otherwise, the setting will be only applied when the pods (user & system managed pods) are deleted and restarted. Users can also re-update the setting to trigger the process again.
- For an IM pod especially, if there are no running instances on a running IM pod, it will be deleted and recreated to apply the new setting.

controller/instance_manager_controller.go
```diff
@@ -515,6 +520,130 @@ func (imc *InstanceManagerController) annotateCASafeToEvict(im *longhorn.Instanc
...
+		dataEngine = longhorn.DataEngineTypeV1
+	case types.SettingNameV2DataEngine:
+		dataEngine = longhorn.DataEngineTypeV2
+	}
+	if !enabled && im.Spec.DataEngine == dataEngine {
+		imc.logger.Infof("Cleaning up the instance manager %v for %v data engine", im.Name, dataEngine)
+		if err := imc.ds.DeleteInstanceManager(im.Name); err != nil {
+			return false, err
+		}
+	}
```
innobead, line 643: We should not delete instance managers in `isSettingDataEngineSynced`. This method is used for checking if the setting synced or not instead fo doing any action.

controller/setting_controller.go
```diff
@@ -218,7 +218,28 @@ func (sc *SettingController) syncSetting(key string) (err error) {
 	if err != nil {
 		return err
 	}
-	switch name {
+
+	if err := sc.syncNonDangerZoneSettingsForManagedComponents(name); err != nil {
+		return err
+	}
+	// SettingCategoryDangerZone and they will try to delete pods to update the settings.
```
innobead, line 225: Remove this comment or make it understandable, or you can make a comment to `syncDangerZoneSettingsForManagedComponents`.

innobead (changes requested):
As I mentioned, some code was still your previous change.

Please review https://github.com/longhorn/longhorn-manager/commit/5057a56c418066d78bf5f204ce8404a87e468f88#diff-be348ed45296f49335ad1afc4a80cdcb18350cbcbbf701a740c2f400e6c31407 again, to see how to improve the PR.

controller/setting_controller.go
mantissahz: Removed the comment. Thanks.

controller/instance_manager_controller.go
mantissahz: Removed
```golang
imc.logger.Infof("Cleaning up the instance manager %v for %v data engine", im.Name, dataEngine)
if err := imc.ds.DeleteInstanceManager(im.Name); err != nil {
    return false, err
}
```
and used instead
```golang
return false, nil
```

innobead (dismissed):
LGTM.

innobead (approved):

innobead:
cc @derekbit , this is also related to v1/v2 data engine settings.

k8s/crds.yaml
```diff
@@ -2801,6 +2801,7 @@ spec:
           metadata:
             type: object
           value:
+            description: The value of the setting.
```
innobead, line 2804: Have you updated the longhorn repo?

PR #2414, opened 2023-12-27
Handling of 16TiB+ volumes with ext4 as the underlying file system

mantissahz:
Get the storage class of the PVC to check if the new volume size is bigger than the maximum size of the file system in the storage class for the volume expansion.

If the file system format of the disk is `ext4` and the volume size of this replica is bigger than 16TiB, it will not schedule replicas to this disk.

Ref: longhorn/longhorn#7423

### DIFF
--- a/controller/node_controller.go
+++ b/controller/node_controller.go
@@ -720,7 +720,7 @@ func (nc *NodeController) syncDiskStatus(node *longhorn.Node, collectedDataInfo 
 
 	for _, diskInfoMap := range readyDiskInfoMap {
 		nc.updateReadyDiskStatusReadyCondition(node, diskInfoMap)
-		nc.updateReadyDiskStatusFileSystemType(node, diskInfoMap)
+		nc.updateDiskStatusFileSystemType(node, diskInfoMap)
 	}
 
 	return nc.updateDiskStatusSchedulableCondition(node)
@@ -843,7 +843,7 @@ func (nc *NodeController) updateReadyDiskStatusReadyCondition(node *longhorn.Nod
 	}
 }
 
-func (nc *NodeController) updateReadyDiskStatusFileSystemType(node *longhorn.Node, diskInfoMap map[string]*monitor.CollectedDiskInfo) {
+func (nc *NodeController) updateDiskStatusFileSystemType(node *longhorn.Node, diskInfoMap map[string]*monitor.CollectedDiskInfo) {
 	diskStatusMap := node.Status.DiskStatus
 	for diskName, info := range diskInfoMap {
 		diskStatus := diskStatusMap[diskName]
--- a/controller/node_controller.go
+++ b/controller/node_controller.go
@@ -720,6 +720,7 @@ func (nc *NodeController) syncDiskStatus(node *longhorn.Node, collectedDataInfo 
 
 	for _, diskInfoMap := range readyDiskInfoMap {
 		nc.updateReadyDiskStatusReadyCondition(node, diskInfoMap)
+		nc.updateReadyDiskStatusFileSystemType(node, diskInfoMap)
 	}
 
 	return nc.updateDiskStatusSchedulableCondition(node)
@@ -837,6 +838,17 @@ func (nc *NodeController) updateReadyDiskStatusReadyCondition(node *longhorn.Nod
 				longhorn.DiskConditionTypeReady, longhorn.ConditionStatusTrue,
 				"", fmt.Sprintf("Disk %v(%v) on node %v is ready", diskName, diskInfoMap[diskName].Path, node.Name),
 				nc.eventRecorder, node, corev1.EventTypeNormal)
+		}
+		diskStatusMap[diskName] = diskStatus
+	}
+}
+
+func (nc *NodeController) updateReadyDiskStatusFileSystemType(node *longhorn.Node, diskInfoMap map[string]*monitor.CollectedDiskInfo) {
+	diskStatusMap := node.Status.DiskStatus
+	for diskName, info := range diskInfoMap {
+		diskStatus := diskStatusMap[diskName]
+		if diskStatus.DiskUUID == info.DiskUUID && diskStatus.Type == longhorn.DiskTypeFilesystem {
+			diskStatus.FSType = info.DiskStat.Type
 		}
 		diskStatusMap[diskName] = diskStatus
 	}
--- a/datastore/longhorn.go
+++ b/datastore/longhorn.go
@@ -4881,3 +4881,15 @@ func IsBackendStoreDriverV1(backendstoreDriver longhorn.BackendStoreDriverType) 
 func IsBackendStoreDriverV2(backendstoreDriver longhorn.BackendStoreDriverType) bool {
 	return backendstoreDriver == longhorn.BackendStoreDriverTypeV2
 }
+
+// IsSupportedVolumeSize returns turn if the v1 volume size is supported by the given fsType file system.
+func IsSupportedVolumeSize(backendstoreDriver longhorn.BackendStoreDriverType, fsType string, volumeSize int64) bool {
+	// TODO: check the logical volume maximum size limit
+	if IsBackendStoreDriverV1(backendstoreDriver) {
+		// unix.Statfs can not differentiate the ext2/ext3/ext4 file systems.
+		if (strings.HasPrefix(fsType, "ext") && volumeSize >= util.MaxExt4VolumeSize) || (fsType == "xfs" && volumeSize >= util.MaxXfsVolumeSize) {
+			return false
+		}
+	}
+	return true
+}
--- a/k8s/crds.yaml
+++ b/k8s/crds.yaml
@@ -2019,6 +2019,8 @@ spec:
                       type: string
                     diskUUID:
                       type: string
+                    filesystemType:
+                      type: string
                     scheduledReplica:
                       additionalProperties:
                         format: int64
--- a/scheduler/replica_scheduler.go
+++ b/scheduler/replica_scheduler.go
@@ -2,6 +2,7 @@ package scheduler
 
 import (
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/pkg/errors"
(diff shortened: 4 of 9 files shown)


scheduler/replica_scheduler.go
```diff
@@ -320,6 +321,13 @@ func (rcs *ReplicaScheduler) filterNodeDisksForReplica(node *longhorn.Node, disk
 			continue
 		}
 
+		volumeSize := volume.Spec.Size
+		// unix.Statfs can not differentiate the ext2/ext3/ext4 file systems.
+		if (strings.HasPrefix(string(diskStatus.Type), "ext") && volumeSize >= util.MaxExt4VolumeSize) || (string(diskStatus.Type) == "xfs" && volumeSize >= util.MaxXfsVolumeSize) {
+			logrus.Debugf("Volume %v size %v is not compatible with the file system %v of the disk %v", volume.Name, volume.Spec.Size, diskStatus.Type, diskName)
+			continue
+		}
+
```
derekbit, line 330: volume validator already rejects volumes with the conditions. Do you need to add the check here?

webhook/resources/volume/validator.go
```diff
@@ -362,6 +362,10 @@ func (v *volumeValidator) validateExpansionSize(oldVolume *longhorn.Volume, newV
 		return fmt.Errorf("storage class %v of PVC %v does not allow the volume expansion", pvcSCName, pvcName)
 	}
 
+	if (pvcStorageClass.Parameters["fsType"] == "ext4" && newSize >= util.MaxExt4VolumeSize) || (pvcStorageClass.Parameters["fsType"] == "xfs" && newSize >= util.MaxXfsVolumeSize) {
```
derekbit, line 365: Should we add a check in Create() function as well?

scheduler/replica_scheduler.go
mantissahz: The volume validator will not reject that users create a volume over 16TiB because we don't know if there are disks that are formatted to "ext4", "xfs" or others until we create the volume and it starts to schedule the replicas to disks.

@derekbit can you point out where the volume validator rejects creating volumes?

webhook/resources/volume/validator.go
mantissahz: There might be no PV/PVC we can get if we just create a volume from Longhorn UI, so I took the suggestion from Eric https://github.com/longhorn/longhorn/issues/7423#issue-2054290621
```txt
Do not schedule replicas for 16TiB+ volumes to ext4 disks
```

webhook/resources/volume/validator.go
```diff
@@ -341,6 +342,25 @@ func (v *volumeValidator) validateExpansionSize(oldVolume *longhorn.Volume, newV
...
+
+	for _, replica := range replicas {
+		if datastore.IsBackendStoreDriverV1(replica.Spec.BackendStoreDriver) {
+			diskUUID := replica.Spec.DiskID
+			node, diskName, err := v.ds.GetReadyDiskNode(diskUUID)
+			if err != nil {
+				return err
+			}
+			diskStatus := node.Status.DiskStatus[diskName]
+			if (strings.HasPrefix(string(diskStatus.Type), "ext") && newSize >= util.MaxExt4VolumeSize) || (string(diskStatus.Type) == "xfs" && newSize >= util.MaxXfsVolumeSize) {
```
ejweber, line 358:
```suggestion
			if (string(diskStatus.Type) == "ext4" && newSize >= util.MaxExt4VolumeSize) || (string(diskStatus.Type) == "xfs" && newSize >= util.MaxXfsVolumeSize) {
```

I think we should do an exact match check here. I'm honestly not sure whether it's possible to run Longhorn on top of ext3, but if it is, ext3 has a different maximum file size.

webhook/resources/volume/validator.go
ejweber: I don't think this check is necessary. It is ok to create an ext4 file system that is larger than 16 TiB. We just can't create a file system larger than 16 TiB on top of an ext4 file system (within a 16 TiB file).

webhook/resources/volume/validator.go
ejweber: Ah, I see your above comment that we cannot know which ext it is. So what do we see here? `ext2/ext3` as in my output below?

```
eweber@laptop:~/longhorn-manager> stat -f .
  File: "."
    ID: 23a4862bca551670 Namelen: 255     Type: ext2/ext3
Block size: 4096       Fundamental block size: 4096
Blocks: Total: 263940717  Free: 249947342  Available: 236521474
Inodes: Total: 67108864   Free: 66501058
```

If this is the case, I guess for now the `HasPrefix` makes sense. (Maybe we can put a similar comment here so it is clear why we do not do the exact match.

controller/node_controller.go
```diff
@@ -837,6 +837,9 @@ func (nc *NodeController) updateReadyDiskStatusReadyCondition(node *longhorn.Nod
 				longhorn.DiskConditionTypeReady, longhorn.ConditionStatusTrue,
 				"", fmt.Sprintf("Disk %v(%v) on node %v is ready", diskName, diskInfoMap[diskName].Path, node.Name),
 				nc.eventRecorder, node, corev1.EventTypeNormal)
+			if diskStatus.Type == longhorn.DiskTypeFilesystem {
+				diskStatus.Type = longhorn.DiskType(info.DiskStat.Type)
+			}
```
ejweber, line 842: I do not think it is appropriate to set this here (since this function is for updating the ready condition. We should probably uplevel it to `syncDiskStatus` (or a new function within `syncDiskStatus`.

I don't think I like changing the meaning of the `diskStatus.Type` field, which differentiates between v1 and v2 disks. I prefer a new field like `diskStatus.filesystemType` (which can be ignored for v2 disks).

scheduler/replica_scheduler.go
ejweber: I think we do. We should not prevent users from creating a 16TiB volume in case there are at least some XFS replicas in the cluster. So we need to decide during replica scheduling which disks we can use.

ejweber (changes requested):
Looks good in principle, but I think we need some implementation changes.

webhook/resources/volume/validator.go
mantissahz: Removed checking the PVC when expanding volume.

webhook/resources/volume/validator.go
mantissahz: >Ah, I see your above comment that we cannot know which ext it is. So what do we see here? ext2/ext3 as in my output below?

Yes, it is as in your output. The `stat -f` shows `ext2/ext3` for `ext4` file system.
Added the comment to the check statement.
Thanks.

controller/node_controller.go
mantissahz: I agree to add a new field to keep the file system type as well.
Added a new field `FSType` in `DiskStatus` to record the file system type.
Thanks.

webhook/resources/volume/validator.go
innobead: What Eric said, it's longhorn FS disk.

k8s/crds.yaml
```diff
@@ -2019,6 +2019,8 @@ spec:
                       type: string
                     diskUUID:
                       type: string
+                    filesystemType:
```
innobead, line 2022: note: this is only for v1.

controller/node_controller.go
```diff
@@ -720,6 +720,7 @@ func (nc *NodeController) syncDiskStatus(node *longhorn.Node, collectedDataInfo
 
 	for _, diskInfoMap := range readyDiskInfoMap {
 		nc.updateReadyDiskStatusReadyCondition(node, diskInfoMap)
+		nc.updateReadyDiskStatusFileSystemType(node, diskInfoMap)
```
innobead, line 723: `updateDiskStatusFileSystemType`

scheduler/replica_scheduler.go
```diff
@@ -320,6 +321,13 @@ func (rcs *ReplicaScheduler) filterNodeDisksForReplica(node *longhorn.Node, disk
 			continue
 		}
 
+		volumeSize := volume.Spec.Size
+		// unix.Statfs can not differentiate the ext2/ext3/ext4 file systems.
+		if (strings.HasPrefix(diskStatus.FSType, "ext") && volumeSize >= util.MaxExt4VolumeSize) || (diskStatus.FSType == "xfs" && volumeSize >= util.MaxXfsVolumeSize) {
+			logrus.Debugf("Volume %v size %v is not compatible with the file system %v of the disk %v", volume.Name, volume.Spec.Size, diskStatus.Type, diskName)
+			continue
+		}
```
innobead, line 329: Do this only for v1.

webhook/resources/volume/validator.go
```diff
@@ -331,50 +332,71 @@
...
 		return fmt.Errorf("shrinking volume %v size from %v to %v is not supported", newVolume.Name, oldSize, newSize)
 	}
 
+	replicas, err := v.ds.ListVolumeReplicasRO(newVolume.Name)
+	if err != nil {
+		return err
+	}
+
+	for _, replica := range replicas {
+		if datastore.IsBackendStoreDriverV1(replica.Spec.BackendStoreDriver) {
```
innobead, line 351: Suggest a common function for this? the similar code used in replica scheduling.

innobead (changes requested):

controller/node_controller.go
mantissahz: Modified the name. Thanks.

webhook/resources/volume/validator.go
mantissahz: Added a common function. Thanks.

scheduler/replica_scheduler.go
mantissahz: Added a common function and check if it is v1 volume. Thanks.

controller/node_controller.go
mantissahz: Also added a new function to update the `FSType` in `syncDiskStatus`

scheduler/replica_scheduler.go
derekbit: Got it. Thanks @mantissahz and @ejweber

datastore/longhorn.go
```diff
@@ -4881,3 +4881,14 @@ func IsBackendStoreDriverV1(backendstoreDriver longhorn.BackendStoreDriverType)
 func IsBackendStoreDriverV2(backendstoreDriver longhorn.BackendStoreDriverType) bool {
 	return backendstoreDriver == longhorn.BackendStoreDriverTypeV2
 }
+
+// IsFileSystemSupportedVolumeSize returns turn if the volume size is supported by the given fsType file system.
+func IsFileSystemSupportedVolumeSize(backendstoreDriver longhorn.BackendStoreDriverType, fsType string, volumeSize int64) bool {
+	if IsBackendStoreDriverV1(backendstoreDriver) {
```
derekbit, line 4887: This can be applied to v2 volumes, right?

derekbit (commented):
In general, LGTM.

datastore/longhorn.go
```diff
@@ -4881,3 +4881,14 @@ func IsBackendStoreDriverV1(backendstoreDriver longhorn.BackendStoreDriverType)
...
+		// unix.Statfs can not differentiate the ext2/ext3/ext4 file systems.
+		if (strings.HasPrefix(fsType, "ext") && volumeSize >= util.MaxExt4VolumeSize) || (fsType == "xfs" && volumeSize >= util.MaxXfsVolumeSize) {
+			return false
+		}
+	}
```
derekbit, line 4892: Two points

1. Rename `IsFileSystemSupportedVolumeSize` to `IsSupportedVolumeSize`
2.  Can you add 
```
TODO: check the logical volume maximum size limit
```

innobead (approved):

datastore/longhorn.go
mantissahz: Renamed the function and added the TODO comment.
Thanks.

innobead (approved):

derekbit (approved):

PR #2415, opened 2023-12-28
fix(log): mismatched args

c3y1huang:
`...Infof call needs 2 args but has 3 args`


innobead (approved):

c3y1huang:

PR #2416, opened 2023-12-28
feat(backingimage): backup backing image support

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/4165


manager/volume.go
```diff
@@ -149,6 +150,11 @@ func (m *VolumeManager) Create(name string, spec *longhorn.VolumeSpec, recurring
 		}
 	}
 
+	// restore backing image if needed
+	if err := m.restoreBackingImage(spec.BackingImage); err != nil {
+		return nil, errors.Wrapf(err, "failed to restore backing image %v when create volume %v", spec.BackingImage, v.Name)
+	}
+
```
ChanYiLin, line 157: restore volume is actually create volume with FromBackup in the spec
in create volume function we check

1. if backing image exists
2. if backing image doesn't exist, check if backup backing image exists
3. create backing image by restore if backup backing image exists

ChanYiLin:
Conflicts resolved

ChanYiLin:
Conflict resolved

ChanYiLin:
Conflicts resolved

ChanYiLin:
Conflicts resolved

webhook/resources/backupbackingimage/mutator.go
```diff
@@ -0,0 +1,62 @@
...
+	return mutate(newObj)
+}
+
+func (b *backupBackingImageMutator) Update(request *admission.Request, oldObj runtime.Object, newObj runtime.Object) (admission.PatchOps, error) {
+	return mutate(newObj)
+}
+
+// mutate contains functionality shared by Create and Update.
+func mutate(newObj runtime.Object) (admission.PatchOps, error) {
+	backupBackingImage := newObj.(*longhorn.BackupBackingImage)
```
innobead, line 49: Check if `backupBackingImage` is nil.

engineapi/backup_backing_image.go
```diff
@@ -0,0 +1,260 @@
...
+			}
+
+			m.logger.WithError(err).Warnf("backup backing image request timeout")
+			bbi.Status.State = longhorn.BackupStatePending
+		}
+
+		m.backupBackingImageStatus = bbi.Status
+	}
+
+	go m.monitorBackupBackingImage()
```
innobead, line 132: If the monitor has been deleted from the backup backing image controller, will this goroutine be deleted immediately or until the timeout?

We should inject the ctx to let the goroutine be deleted immediately.

controller/backup_backing_image_controller.go
```diff
@@ -0,0 +1,478 @@
...
+		eventRecorder: eventBroadcaster.NewRecorder(scheme, v1.EventSource{Component: "longhorn-backup-backing-image-controller"}),
+
+		proxyConnCounter: proxyConnCounter,
+	}
+
+	ds.BackupBackingImageInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{ //nolint:errcheck
+		AddFunc:    bc.enqueueBackupBackingImage,
+		UpdateFunc: func(old, cur interface{}) { bc.enqueueBackupBackingImage(cur) },
+		DeleteFunc: bc.enqueueBackupBackingImage,
+	})
```
innobead, line 95: ~Why using periodic sync? Use AddEventHandlerWithResyncPeriod with 0 resync instead.~

Ok, you are relying on the periodic sync to update the status from the monitor.

innobead (changes requested):
In general, LGTM. A few comments need to resolve first.

engineapi/backup_backing_image.go
innobead: Sinc we have other monitor mechanisms, are they having this issue?

ChanYiLin: Yes, the go routine `m.monitorBackupBackingImage()`
m->monitor has context and a quit() function in it
```
	m := &BackupBackingImageMonitor{
		logger: logger.WithFields(logrus.Fields{"backupBackingImage": bbi.Name}),

		backupBackingImageName: bbi.Name,
		client:                 bimClient,

		backupStatusLock: sync.RWMutex{},

		syncCallback: syncCallback,
		callbackKey:  bbi.Namespace + "/" + bbi.Name,

		ctx:  ctx,
		quit: quit,
	}
```

inside the goroutine, it will return when the `m.ctx` is done
- https://github.com/longhorn/longhorn-manager/pull/2416/files#diff-241fdec810f0d17f4f8d4857a52ea99eb1cf318e00992cd0a530c64738ca7676R148
- https://github.com/longhorn/longhorn-manager/pull/2416/files#diff-241fdec810f0d17f4f8d4857a52ea99eb1cf318e00992cd0a530c64738ca7676R207

ChanYiLin:
Conflicts resolved

innobead (approved):

engineapi/backup_backing_image.go
innobead: yeah, it will be called by `quite`.

PR #2419, opened 2023-12-28
webhook: Disable the engine image check for v2 volume

shuo-wu:
Otherwise, creating, updating, and deleting v2 volumes will get stuck

### DIFF
--- a/webhook/resources/engine/validator.go
+++ b/webhook/resources/engine/validator.go
@@ -66,8 +66,10 @@ func (e *engineValidator) Create(request *admission.Request, newObj runtime.Obje
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


derekbit:
I have resolved it in https://github.com/longhorn/longhorn-manager/pull/2386

shuo-wu:
@derekbit Is that PR ready to go? This bug blocks the debugging

derekbit:
> @derekbit Is that PR ready to go? This bug blocks the debugging

Yeah, waiting for @innobead's approval.

innobead:
Merged

derekbit:
Thank you @innobead 
Then, we can close the PR. cc @shuo-wu

PR #2420, opened 2023-12-29
Rename backendStoreDriver to dataEngin in instance-manager and associated components

derekbit:
https://github.com/longhorn/longhorn/issues/7095
https://github.com/longhorn/longhorn/issues/7480

deps on https://github.com/longhorn/backupstore/pull/153 and https://github.com/longhorn/longhorn-instance-manager/pull/348


engineapi/instance_manager.go
```diff
@@ -240,10 +240,22 @@ func parseProcess(p *imapi.Process) *longhorn.InstanceProcess {
 			Conditions: p.ProcessStatus.Conditions,
 			PortStart:  p.ProcessStatus.PortStart,
 			PortEnd:    p.ProcessStatus.PortEnd,
+
+			// These fields are not used, maybe we can deprecate them later.
```
innobead, line 244: Sounds good. 

// FIXME These fields are not used, maybe we can deprecate them later.

innobead (approved):
Merge this PR after the dep gets updated.

engineapi/instance_manager.go
derekbit: Updated. Thank you!

innobead (approved):

engineapi/instance_manager.go
```diff
@@ -240,10 +240,22 @@ func parseProcess(p *imapi.Process) *longhorn.InstanceProcess {
...
 		},
 	}
 }
 
+func getDataEngineFromInstanceProcess(p *imapi.Instance) longhorn.DataEngineType {
+	if p.DataEngine != "" {
+		return longhorn.DataEngineType(p.DataEngine)
+	}
+
+	return longhorn.DataEngineType(p.BackendStoreDriver)
```
innobead, line 256: Caught by the PR check.

derekbit: Fixed. Thank you.

engineapi/instance_manager.go
```diff
@@ -240,10 +240,22 @@ func parseProcess(p *imapi.Process) *longhorn.InstanceProcess {
...
 		},
 	}
 }
 
+func getDataEngineFromInstanceProcess(p *imapi.Instance) longhorn.DataEngineType {
+	if p.DataEngine != "" {
+		return longhorn.DataEngineType(p.DataEngine)
+	}
+
+	return longhorn.DataEngineType(p.DataEngine)
```
c3y1huang, line 256: This returns the same value as the condition above?

innobead: Good :eye:

derekbit: The original one is correct. That is falling back to backendStoreDriver.
static checker does not like reading data from deprecated field...

derekbit: Any solution to the golinter's deprecation check?
@innobead @c3y1huang

c3y1huang: I think we could ignore `SA1019`? 
https://github.com/longhorn/longhorn-manager/pull/2400/files#diff-4c0e1e708c59d9a06da4c01a46c2d30739fce6b57e17eff33b2930da8d294692R25

innobead: Can we ignore a specific line? I feel it should be feasible.

innobead: https://golangci-lint.run/usage/false-positives/#nolint-directive

derekbit: Not sure why `// lint:ignore SA1019` doesn't work. Let me try `nolint`

derekbit: `nolint: all` works

innobead (approved):

PR #2423, opened 2023-12-29
feat(instance status): reset instance error status when it is not running

ChanYiLin:

When the pod is recreated and the engine is relaunched
There might be a time gap that the error condition is not sync from the instance-manager and remains `error=true`
And keep requesting remount.

Should reset error condition when the instance is not running


innobead (dismissed):

innobead:
@ChanYiLin please check the failed PR check.

ChanYiLin:
Hi @innobead , PR check fixed, thanks!

innobead (approved):

PR #2425, opened 2024-01-03
Allow a Longhorn node to be deleted after its corresponding Kubernetes node

ejweber:
longhorn/longhorn#7475

This PR fixes the two main issues from https://github.com/longhorn/longhorn/issues/7475#issuecomment-1876060097:

1. The nil pointer dereference is no longer possible. The manager pod on the removed node no longer panics and crashes immediately after the node is removed.
2. Even if the manager pod DOES restart and recreate the Longhorn node object in a broken (deleting) state, we can delete it right away instead of getting stuck.


innobead (approved):
LGTM

PR #2426, opened 2024-01-04
fix(cve): update dependencies

c3y1huang:
https://github.com/longhorn/longhorn/issues/7523


c3y1huang:
Trivy identified CVE vulnerability in [otel](https://github.com/longhorn/longhorn-manager/blob/970ba447c112ee5300f7cfc61b07b42c29fabaa5/go.mod#L170-L171). This is fixed Kubernetes v1.29, however, upstream has decided not to backport the fix to v1.28. See [discussion](https://github.com/kubernetes/kubernetes/pull/121559#issuecomment-1782870871).

cc @innobead

innobead (approved):

PR #2427, opened 2024-01-05
fix(util): resource name should only contain lower case

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/7543

> K8s resource name should follow the following constraints
> - contain at most 63 characters
> - contain only lowercase alphanumeric characters or '-'
> - start with an alphanumeric character
> - end with an alphanumeric character

But our regex allows uppper case which is invalid.
Also add the validation to the BackupVolume vame list, so it won't create BackupVolume with invalid name.

This util function is mostly used in webhook validation so it won't effect those functions
The only part effected is the group name of the recurring job. It is okay to have uppercase, so I create another util function `ValidateString()` for it.

Screenshot from 2024-01-05 11-55-43

### DIFF
--- a/datastore/longhorn.go
+++ b/datastore/longhorn.go
@@ -4137,7 +4137,7 @@ func ValidateRecurringJob(job longhorn.RecurringJobSpec) error {
 		return fmt.Errorf("job name %v must be %v characters or less", job.Name, NameMaximumLength)
 	}
 	for _, group := range job.Groups {
-		if !util.ValidateName(group) {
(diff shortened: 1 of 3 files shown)


ChanYiLin:
Test is now running: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/5863/

ChanYiLin:
Test passed
cc @innobead

derekbit (approved):
LGTM

c3y1huang (approved):

PR #2428, opened 2024-01-05
instance manager: update liveness probe

derekbit:

Issue longhorn/longhorn#7551

- instancce manager pod for v1 data engine
  - 8500: process manager grpc service
  - 8501: proxy grpc service
  - 8502: disk grpc service
  - 8503: instance grpc service

- instancce manager pod for v2 data engine
  - 8500-8503
  - 8504: spdk grpc service
  - spdk_tgt daemon process

https://github.com/longhorn/longhorn-spdk-engine/pull/87

### DIFF
--- a/controller/instance_manager_controller.go
+++ b/controller/instance_manager_controller.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"reflect"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


innobead (approved):
LGTM.

nit: The other way is we can prepare a script to handle this, then the liveness probe only call it to prevent complicated command aggregration. The script can be added to the build image.

PR #2429, opened 2024-01-05
refactor(webhook): break for early exit in finalizer check

c3y1huang:
Add break for early exit from the loop once the finalizer is found.


innobead (approved):

innobead:
ref: longhorn/longhorn#7475

cc @ejweber

ejweber:
Good find. Thanks @c3y1huang!

PR #2432, opened 2024-01-05
Be more careful while cleaning up stale replicas

ejweber:

longhorn/longhorn#7425

- The `replica.Spec.HealthyAt` field tells us if a replica has ever been healthy.
- The `replica.Spec.FailedAt` field is set every time we detect a replica failure.
- Before we rebuild a replica, we set `replica.Spec.HealthyAt == ""` and `replica.Spec.FailedAt == ""`. The `replica.Spec.HealthyAt == ""` should be a solid indicator that we cannot trust the data on the replica and should never try to use it.
- If all replicas have `replica.Spec.FailedAt != ""`, we select replicas with `replica.Spec.HealthyAt != ""` to salvage (by removing clearing its `replica.Spec.FailedAt`.
- In some situations, we have seen a replica that was previously corrupted by a failed rebuild have a `replica.Spec.HealthyAt != ""` and be used by the engine.
- Longhorn tries to use this replica to rebuild other failed replicas, but it is impossible. The engine cannot start with this replica.
- Eventually, other replicas become stale and are deleted by Longhorn, leaving only this replica.

It is currently very difficult for Longhorn to automatically understand this situation. It knows not to clean up other replicas if there is no `replica.Spec.HealthyAt != ""` replica. But there IS a replica with `replica.Spec.HealthyAt != ""`. There is no path for us to decide that some replica with `replica.Spec.HealthyAt != ""` actually is NOT healthy.

This PR introduces a `replica.Spec.LastHealthyAt` field and a `replica.Spec.LastFailedAt` field. We set `replica.Spec.LastHealthyAt` every time a replica goes RW instead of just the first time and never clear it. Now we remember the most recent time this replica was working. We set `replica.Spec.LastFailedAt` every time we set `replica.Spec.FailedAt` and never clear it. Now we remember the most recent time this replica failed (even if it is currently being used for a rebuild or salvage).

Now we can look at a replica with `replica.Spec.HealthyAt` and ask, "Has it been healthy recently?" More specifically, we can ask, "Has it become healthy (RW) since the last time it failed?" A replica in an autosalvage loop cannot become healthy, so the answer to this question is, "no." If "no", we should not delete any replicas that have ever been healthy. In the worst case scenario where a user/customer has a volume that cannot attach due to its only "healthy" replica being corrupt, hopefully we will still have these other replicas around and can potentially extract data from them.

TBH, this is best-effort. We do not know exactly how a replica can be rebuilding (and then fail to be rebuilt) while its `replica.Spec.HealthyAt != ""`. I have demonstrated one (artificial) situation in https://github.com/longhorn/longhorn-manager/pull/2432#issuecomment-1894675916 that this fix can be beneficial in, but I do not know if it would help us in the real world or not.

~@innobead, this still needs another work day of additional testing and upgrade logic. I'm opening this draft so you can get a quick idea of what I'm thinking and make a decision whether we should try to push for v1.6.0 or not.~

### DIFF
--- a/controller/engine_controller.go
+++ b/controller/engine_controller.go
@@ -1893,9 +1893,7 @@ func (ec *EngineController) startRebuilding(e *longhorn.Engine, replicaName, add
 					return
 				}
 
-				now := util.Now()
-				replica.Spec.FailedAt = now
-				replica.Spec.LastFailedAt = now
+				setReplicaFailedAt(replica, util.Now())
 				replica.Spec.DesireState = longhorn.InstanceStateStopped
 				if _, err := ec.ds.UpdateReplica(replica); err != nil {
 					log.WithError(err).Errorf("Unable to mark failed rebuild on replica %v", replicaName)
--- a/controller/engine_controller.go
+++ b/controller/engine_controller.go
@@ -1893,7 +1893,9 @@ func (ec *EngineController) startRebuilding(e *longhorn.Engine, replicaName, add
 					return
 				}
 
-				replica.Spec.FailedAt = util.Now()
+				now := util.Now()
+				replica.Spec.FailedAt = now
+				replica.Spec.LastFailedAt = now
 				replica.Spec.DesireState = longhorn.InstanceStateStopped
 				if _, err := ec.ds.UpdateReplica(replica); err != nil {
 					log.WithError(err).Errorf("Unable to mark failed rebuild on replica %v", replicaName)
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -916,6 +916,7 @@ func isDefinitelyHealthyAndActiveReplica(r *longhorn.Replica) bool {
 	// An empty r.Spec.FailedAt doesn't necessarily indicate a healthy replica. The replica could be caught in an
 	// autosalvage loop. If it is, its r.Spec.FailedAt is repeatedly transitioning between empty and some time. Ensure
 	// the replica has become healthyAt since it last became failedAt.
+	// We know r.Spec.LastHealthyAt != "" because r.Spec.HealthyAt != "" from isHealthyAndActiveReplica.
 	if r.Spec.LastFailedAt != "" && !util.TimestampAfterTimestamp(r.Spec.LastHealthyAt, r.Spec.LastFailedAt) {
 		return false
 	}
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -654,7 +654,9 @@ func (c *VolumeController) ReconcileEngineReplicaState(v *longhorn.Volume, es ma
 					r.Name, r.Status.CurrentState, r.Spec.EngineName, r.Spec.Active, isNoAvailableBackend)
 				e.Spec.LogRequested = true
 				r.Spec.LogRequested = true
-				r.Spec.FailedAt = c.nowHandler()
+				now := c.nowHandler()
+				r.Spec.FailedAt = now
+				r.Spec.LastFailedAt = now
 				r.Spec.DesireState = longhorn.InstanceStateStopped
 			}
 		}
@@ -708,7 +710,9 @@ func (c *VolumeController) ReconcileEngineReplicaState(v *longhorn.Volume, es ma
 			}
 			if r.Spec.FailedAt == "" {
 				log.Warnf("Replica %v is marked as failed, current state %v, mode %v, engine name %v, active %v", r.Name, r.Status.CurrentState, mode, r.Spec.EngineName, r.Spec.Active)
-				r.Spec.FailedAt = c.nowHandler()
+				now := c.nowHandler()
+				r.Spec.FailedAt = now
+				r.Spec.LastFailedAt = now
 				e.Spec.LogRequested = true
 				r.Spec.LogRequested = true
 			}
@@ -719,7 +723,9 @@ func (c *VolumeController) ReconcileEngineReplicaState(v *longhorn.Volume, es ma
 			// from replica failed during rebuilding
 			if r.Spec.HealthyAt == "" {
 				c.backoff.DeleteEntry(r.Name)
-				r.Spec.HealthyAt = c.nowHandler()
+				now := c.nowHandler()
+				r.Spec.HealthyAt = now
+				r.Spec.LastHealthyAt = now
 				r.Spec.RebuildRetryCount = 0
 			}
 			healthyCount++
@@ -733,7 +739,9 @@ func (c *VolumeController) ReconcileEngineReplicaState(v *longhorn.Volume, es ma
 				r.Name, r.Status.CurrentState, r.Spec.EngineName, r.Spec.Active)
 			e.Spec.LogRequested = true
 			r.Spec.LogRequested = true
-			r.Spec.FailedAt = c.nowHandler()
+			now := c.nowHandler()
+			r.Spec.FailedAt = now
+			r.Spec.LastFailedAt = now
 			r.Spec.DesireState = longhorn.InstanceStateStopped
 		}
 	}
@@ -888,10 +896,46 @@ func isFirstAttachment(rs map[string]*longhorn.Replica) bool {
 	return true
 }
 
+func isHealthyAndActiveReplica(r *longhorn.Replica) bool {
+	if r.Spec.FailedAt != "" {
+		return false
+	}
+	if r.Spec.HealthyAt == "" {
+		return false
+	}
+	if !r.Spec.Active {
+		return false
+	}
+	return true
+}
+
+func isDefinitelyHealthyAndActiveReplica(r *longhorn.Replica) bool {
+	if !isHealthyAndActiveReplica(r) {
+		return false
+	}
+	// An empty r.Spec.FailedAt doesn't necessarily indicate a healthy replica. The replica could be caught in an
+	// autosalvage loop. If it is, its r.Spec.FailedAt is repeatedly transitioning between empty and some time. Ensure
+	// the replica has become healthyAt since it last became failedAt.
+	if r.Spec.LastFailedAt != "" && !util.TimestampAfterTimestamp(r.Spec.LastHealthyAt, r.Spec.LastFailedAt) {
+		return false
+	}
+	return true
+}
+
 func getHealthyAndActiveReplicaCount(rs map[string]*longhorn.Replica) int {
 	count := 0
 	for _, r := range rs {
-		if r.Spec.FailedAt == "" && r.Spec.HealthyAt != "" && r.Spec.Active {
+		if isHealthyAndActiveReplica(r) {
+			count++
+		}
+	}
+	return count
+}
+
+func getDefinitelyHealthyAndActiveReplicaCount(rs map[string]*longhorn.Replica) int {
+	count := 0
+	for _, r := range rs {
+		if isDefinitelyHealthyAndActiveReplica(r) {
 			count++
 		}
 	}
@@ -944,7 +988,7 @@ func (c *VolumeController) cleanupReplicas(v *longhorn.Volume, es map[string]*lo
 }
 
 func (c *VolumeController) cleanupCorruptedOrStaleReplicas(v *longhorn.Volume, rs map[string]*longhorn.Replica) error {
-	healthyCount := getHealthyAndActiveReplicaCount(rs)
+	healthyCount := getDefinitelyHealthyAndActiveReplicaCount(rs)
 	cleanupLeftoverReplicas := !c.isVolumeUpgrading(v) && !isVolumeMigrating(v)
 	log := getLoggerForVolume(c.logger, v)
 
@@ -972,19 +1016,8 @@ func (c *VolumeController) cleanupCorruptedOrStaleReplicas(v *longhorn.Volume, r
 			continue
 		}
 
-		staled := false
-		if v.Spec.StaleReplicaTimeout > 0 &&
-			util.TimestampAfterTimeout(r.Spec.FailedAt, time.Duration(int64(v.Spec.StaleReplicaTimeout*60))*time.Second) {
-
-			staled = true
-		}
-
 		if datastore.IsDataEngineV1(v.Spec.DataEngine) {
-			// 1. failed for multiple times or failed at rebuilding (`Spec.RebuildRetryCount` of a newly created rebuilding replica
-			//    is `FailedReplicaMaxRetryCount`) before ever became healthy/ mode RW,
-			// 2. failed too long ago, became stale and unnecessary to keep around, unless we don't have any healthy replicas
-			// 3. failed for race condition at upgrading when waiting IM-r to start and it would never became healty
-			if (r.Spec.RebuildRetryCount >= scheduler.FailedReplicaMaxRetryCount) || (healthyCount != 0 && staled) || (r.Spec.Image != v.Status.CurrentImage) {
+			if shouldCleanUpFailedReplicaV1(r, v.Spec.StaleReplicaTimeout, healthyCount, v.Spec.Image) {
 				log.WithField("replica", r.Name).Info("Cleaning up corrupted, staled replica")
 				if err := c.deleteReplica(r, rs); err != nil {
 					return errors.Wrapf(err, "cannot clean up staled replica %v", r.Name)
@@ -1004,6 +1037,8 @@ func (c *VolumeController) cleanupCorruptedOrStaleReplicas(v *longhorn.Volume, r
 }
 
 func (c *VolumeController) cleanupFailedToScheduledReplicas(v *longhorn.Volume, rs map[string]*longhorn.Replica) (err error) {
+	// We don't need the more rigorous getDefinitelyHealthyAndActiveReplicaCount here, because the replicas we will
+	// potentially delete are definitely worthless (contain no data).
 	healthyCount := getHealthyAndActiveReplicaCount(rs)
 	hasEvictionRequestedReplicas := hasReplicaEvictionRequested(rs)
 
@@ -1025,7 +1060,7 @@ func (c *VolumeController) cleanupFailedToScheduledReplicas(v *longhorn.Volume, 
 }
 
 func (c *VolumeController) cleanupExtraHealthyReplicas(v *longhorn.Volume, e *longhorn.Engine, rs map[string]*longhorn.Replica) (err error) {
-	healthyCount := getHealthyAndActiveReplicaCount(rs)
+	healthyCount := getDefinitelyHealthyAndActiveReplicaCount(rs)
 	if healthyCount <= v.Spec.NumberOfReplicas {
 		return nil
 	}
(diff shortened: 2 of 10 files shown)


ejweber:
We should not merge this for now. It works in theory, but I tried today to come up with even a partially real world scenario in which it would outperform our existing logic in protecting replicas and could not. We need a situation in which one replica is corrupted but still `healthyAt`. The other replicas must be `failedAt`. If not, we won't try to clean them up anyway. This also means we can never have tried to rebuild them (which removes their `failedAt` and their `healthyAt`.

Even in the false scenario where I intentionally corrupt a running replica (so that we still consider it `healthyAt`, despite our inability to use it), I can't get the other replicas into this situation without doing something very drastic, like killing multiple nodes and intentionally choosing the worst one to restart.

ejweber:
## Attempts to prove usefulness of behavior:

### Trial 1:

1. Create a volume with two replicas and `staleReplicaTimeout = 5`.
2. Shut down a node with ReplicaA on it (immediate power off).
3. Intentionally corrupt ReplicaB while it is running (delete `xxx.img.meta`).
4. Crash ReplicaB while it is running (kill process inside instance manager).
5. Wait 5 minutes.

#### Results before fix in `v1.3.2`:

1. Longhorn immediately tries to rebuild ReplicaA as soon as communication between the engine and it fails. ReplicaA now has `healthyAt = ""` and `failedAt = ""`.
2. The rebuild fails, so ReplicaA has `failedAt = ""`. This makes it a potential target of stale replica cleanup.

```
    name: ReplicaA
    failedAt: "2024-01-10T15:35:40Z"
    healthyAt: ""
    rebuildRetryCount: 1

    name: ReplicaB
    failedAt: ""
    healthyAt: "2024-01-10T15:23:05Z"
    rebuildRetryCount: 0
```

3. The volume goes into an attach loop with the corrupted ReplicaB.
4. ReplicaA is cleaned up as stale.
5. Data cannot be recovered.

#### Results before fix in `v1.6.0-dev`:

1. Longhorn immediately tries to rebuild ReplicaA as soon as communication between the engine and it fails. ReplicaA now has `healthyAt = ""` and `failedAt = ""`.
2. The rebuild fails, so ReplicaA has `failedAt = ""`. This makes it a potential target of stale replica cleanup.
3. The volume goes into an attach loop with the corrupted ReplicaB.
4. Due to https://github.com/longhorn/longhorn-manager/pull/1582, ReplicaA's `failedAt` is continuously updated to the same as ReplicaB's, even though ReplicaA's node is offline.
5. ReplicaA is never accidentally cleaned up. Maybe we can manually make use of it if we find the volume in this condition.

ejweber:
## Useful test

### Steps:

1. Create a volume with:
    `numberOfReplicas = 2`
    `staleReplicaTimeout = 1`
2. Attach the volume to a node.
3. Write data to the volume.
4. Silently corrupt ReplicaB by `rm -rf <replica_dir_path>/*.img.meta`. This corruption is artificial, but we have not yet determined how it has actually happened in the real world.
5. Cordon the node with ReplicaA.
6. Crash ReplicaA's instance manager.
7. Crash ReplicaB's instance manager.
8. Observe an "autosalvage loop". The volume cannot start with ReplicaB, but there are no other replicas, so it continually tries and fails. Replica status bounce back and forth between the below.

```
eweber@laptop:~/longhorn> kl get replica -oyaml | grep -i -e healthyat -e failedat -e " name:"
    name: test-r-ReplicaB
      name: test
    failedAt: "2024-01-16T21:26:38Z"
    healthyAt: "2024-01-16T21:24:55Z"
    lastFailedAt: "2024-01-16T21:26:38Z"
    lastHealthyAt: "2024-01-16T21:24:55Z"
    name: test-r ReplicaA
      name: test
    failedAt: "2024-01-16T21:26:36Z"
    healthyAt: "2024-01-16T21:12:38Z"
    lastFailedAt: "2024-01-16T21:26:36Z"
    lastHealthyAt: "2024-01-16T21:12:38Z"

eweber@laptop:~/longhorn> kl get replica -oyaml | grep -i -e healthyat -e failedat -e " name:"
    name: test-r-ReplicaB
      name: test
    failedAt: ""
    healthyAt: "2024-01-16T21:24:55Z"
    lastFailedAt: "2024-01-16T21:26:39Z"
    lastHealthyAt: "2024-01-16T21:24:55Z"
    name: test-r-ReplicaA
      name: test
    failedAt: "2024-01-16T21:26:39Z"
    healthyAt: "2024-01-16T21:12:38Z"
    lastFailedAt: "2024-01-16T21:26:39Z"
    lastHealthyAt: "2024-01-16T21:12:38Z"
```

9. Detach the volume (e.g. by modifying the volumeattachment.longhorn.io).

### Before fix:

1. Wait one minute.
2. ReplicaA is cleaned up. Only ReplicaB remains, but ReplicaB is corrupt.

### After fix:

1. Wait any length of time.
2. ReplicaA is never cleaned up.
3. Uncordon the node with ReplicaA.
4. Reattach the volume.
5. Delete ReplicaB.
6. Wait for the volume to rebuild from ReplicaA.
7. Verify the data.

ejweber:
Need to fix volume lifecycle unit tests. They are failing because the volume controller is updating the new `lastHealthyAt` and `lastFailedAt` fields, but the test comparisons don't expect it.

ejweber:
- https://ci.longhorn.io/job/private/job/longhorn-tests-regression/6005/
- https://ci.longhorn.io/job/private/job/longhorn-tests-regression/6009/

ejweber:
There is a philosophical debate amongst US team members regarding this PR. No one is definitely opposed, but there are reservations. To paraphrase the concerns of @PhanLe1010 and @james-munson:

The previous design (using only the `HealthyAt` field to determine health) was intended to protect us from deleting replicas when there are no available healthy replicas. This PR is an additional system that tries to protect us in the rare case that the `HealthyAt` field is unreliable. Instead of adding additional protections, we should fix the unreliability of the `HealthyAt` field.

I agree with this sentiment in general, but I simply cannot determine the root cause of https://github.com/longhorn/longhorn/issues/7425 or of https://github.com/longhorn/longhorn/issues/4212#issuecomment-1643456399 after a long time looking for it. I propose including this PR (or a variation of it) as an additional precaution. WDYT @longhorn/dev-control-plane?

controller/engine_controller.go
```diff
@@ -1890,7 +1890,9 @@ func (ec *EngineController) startRebuilding(e *longhorn.Engine, replicaName, add
 					return
 				}
 
-				replica.Spec.FailedAt = util.Now()
+				now := util.Now()
+				replica.Spec.FailedAt = now
+				replica.Spec.LastFailedAt = now
```
james-munson, line 1895: Let's start here.  With a correlated set of attributes, it's risky to set individual ones directly, allowing for possible contradictory situations if one is overlooked.  Better to have a "setter" method that does all together, perhaps in utils.  So setHealthyAt(time) would set not only Spec.HealthyAt and Spec.LastHealthyAt, but also clear FailedAt, if that's what we decide it should do.   Likewise for setFailedAt(time).

(With these changes in this improvement, non-empty FailedAt should mean "known to be failed right now" and should never be set at the same time that HealthyAt is, I claim. HealthyAt means "known to be healthy right now in a similar way.  It's a contradiction to have both non-empty, even if one could try to reason about the meaning of comparing the timestamps.  LastHealthAt and LastFailedAt are never cleared, so they make this crystal clear.

If I understand this right, it is not a contradiction for both HealthyAt and FailedAt to be empty.  That would be the case for a brand new replica (and LastHealthyAt and LastFailedAt would also be empty.  It would also be the case for a replica that is in the middle of rebuilding and has neither finished nor failed that process.  We could even put that test into a method isRebuilding(), and use it in place of the inline logic, which is used in (c *VolumeController) closeVolumeDependentResources() below, for instance.

controller/volume_controller.go
```diff
@@ -888,10 +896,47 @@ func isFirstAttachment(rs map[string]*longhorn.Replica) bool {
...
+	if r.Spec.HealthyAt == "" {
+		return false
+	}
+	if !r.Spec.Active {
+		return false
+	}
+	return true
+}
+
+func isDefinitelyHealthyAndActiveReplica(r *longhorn.Replica) bool {
```
james-munson, line 912: Perhaps the difference between this and the previous method is clearer if we name it something like "isUsableAsLastReplica(r)"  Or we can just that in a function comment.

controller/volume_controller.go
```diff
@@ -4450,3 +4490,27 @@ func (c *VolumeController) ReconcilePersistentVolume(volume *longhorn.Volume) er
 	}
 	return nil
 }
+
+func shouldCleanUpFailedReplicaV1(r *longhorn.Replica, staleReplicaTimeout, definitelyHealthyCount int,
```
james-munson, line 4494: This is cleaner.

upgrade/v16xto170/upgrade.go
```diff
@@ -0,0 +1,56 @@
+package v16xto170
```
james-munson, line 1: Congratulations on being the first into the 1.7 upgrade pool.

upgrade/v16xto170/upgrade.go
```diff
@@ -0,0 +1,56 @@
...
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil
+		}
+		return errors.Wrapf(err, "failed to list all existing Longhorn replicas during the replica upgrade")
+	}
+
+	for _, r := range replicaMap {
+		if r.Spec.LastHealthyAt == "" {
+			// We could attempt to figure out if the replica is currently RW in an engine and set its
```
james-munson, line 44: Agreed.

util/util.go
```diff
@@ -261,6 +261,20 @@ func TimestampWithinLimit(latest time.Time, ts string, limit time.Duration) bool
 	return deadline.After(latest)
 }
 
+func TimestampAfterTimestamp(after string, before string) bool {
```
james-munson, line 264: I just think backwards, I suppose, but I would have intuitively ordered the arguments as (before, after).

controller/volume_controller.go
```diff
@@ -888,10 +896,47 @@ func isFirstAttachment(rs map[string]*longhorn.Replica) bool {
...
+
+func isDefinitelyHealthyAndActiveReplica(r *longhorn.Replica) bool {
+	if !isHealthyAndActiveReplica(r) {
+		return false
+	}
+	// An empty r.Spec.FailedAt doesn't necessarily indicate a healthy replica. The replica could be caught in an
+	// autosalvage loop. If it is, its r.Spec.FailedAt is repeatedly transitioning between empty and some time. Ensure
+	// the replica has become healthyAt since it last became failedAt.
+	// We know r.Spec.LastHealthyAt != "" because r.Spec.HealthyAt != "" from isHealthyAndActiveReplica.
+	if r.Spec.LastFailedAt != "" && !util.TimestampAfterTimestamp(r.Spec.LastHealthyAt, r.Spec.LastFailedAt) {
```
james-munson, line 920: You could save the ! by swapping the time arguments.  Unless you suspect that they might be equal, I suppose.  (But see my note below on argument order.)

util/util_test.go
```diff
@@ -182,3 +183,25 @@ func (s *TestSuite) TestGetValidMountPoint(c *C) {
 		}
 	}
 }
+
+func TestTimestampAfterTimestamp(t *testing.T) {
+	tests := map[string]struct {
+		before string
+		after  string
```
james-munson, line 190: Hmm.  Perhaps you agree with me, because you order the structure as before, after.  And so are the arguments to the function under test (tc.before, tc.after).  But then the actuallyAfter case has "before" after "after"... Oy.

james-munson (changes requested):
Overall I like this proposal.  It makes the currently ambiguous HealthyAt and FailedAt fields less so.  It may be useful for troubleshooting, too.

controller/engine_controller.go
james-munson: For an example of my confusion, see https://github.com/longhorn/longhorn-manager/blob/8a8414a150ca284996ac434160901f11e5f812fb/webhook/resources/replica/validator.go#L114

controller/volume_controller.go
ejweber: > This statement is right: An empty r.Spec.FailedAt doesn't necessarily indicate a healthy replica. But isHealthyAndActiveReplica(r) == true means r.Spec.HealthyAt != "" && r.Spec.FailedAt == "" ==> This should prove a replica is running/available.

I would like to agree, but I created this PR because this is NOT always the case. Sometimes, we have a replica with `r.Spec.HealthyAt != ""` that is completely corrupted. If it is selected for autosalvage, we clear `r.Spec.FailedAt` (`r.Spec.FailedAt == ""`). This fits our previous definition of "healthy", but we cannot rely on this definition when cleaning up stale replicas! If we do, we may clean up all other replicas in favor of a completely broken one.

> r.Spec.LastHealthyAt being earlier than r.Spec.LastFailedAt != "" ==> Replica is crashed and hasn't been successfully salvaged/reused.

Yes. In an "autosalvage loop", some replica is repeatedly crashed -> salvaged -> crashed -> salvaged -> crashed. If the replica has EVER failed (`r.Spec.LastFailedAt != ""`) and it hasn't "become healthy" (appeared as RW in an engine) since the last time it failed (`!util.TimestampAfterTimestamp(r.Spec.LastHealthyAt, r.Spec.LastFailedAt)`, we cannot rely on it. We must wait to clean up other replicas.

controller/volume_controller.go
```diff
@@ -4450,3 +4490,27 @@ func (c *VolumeController) ReconcilePersistentVolume(volume *longhorn.Volume) er
...
 }
+
+func shouldCleanUpFailedReplicaV1(r *longhorn.Replica, staleReplicaTimeout, definitelyHealthyCount int,
+	volumeCurrentImage string) bool {
+	// Even if healthyAt == "", lastHealthyAt != "" indicates this replica has some (potentially invalid) data. We MUST
+	// NOT delete it until we're sure the engine can start with another replica. In the worst case scenario, maybe we
+	// can recover data from this replica.
+	if r.Spec.LastHealthyAt != "" && definitelyHealthyCount == 0 {
+		return false
+	}
```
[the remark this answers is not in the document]
ejweber: The way I see it, there are two essential fixes:

1. This (as you mentioned).
2. Don't trust `healthyAt != ""` alone to guarantee that there is some other healthy replica.

util/util.go
ejweber: Not sure what the "right" approach is here. I tried both at one point or another.

I read it as `TimestampAfterTimestamp` should answer the question, "Is the first TIMESTAMP I provide AFTER the second TIMESTAMP I provide?" Maybe it adds the the confusion that I name the variables `after` and `before` (presupposing that the first time will be after the second time until proven otherwise). I will change the names to `time1` and `time2` in the next commit.

util/util_test.go
ejweber: Hopefully I clarified this in the same commit I used for https://github.com/longhorn/longhorn-manager/pull/2432/files#r1462198405.

controller/engine_controller.go
ejweber: I'm still working through this feedback a bit, but to start: I thought about the utility method to combine setting (for example) `HealthyAt` and `LastHealthyAt`. I ultimately decided against it because:

1. I don't think it really provides that much protection. Sure a developer could look and see if there is a utility function that they can use, but they may also just set whatever field they want independently.
2. (This is the primary reason I didn't.) We SHOULDN'T actually always set them together if we want to maintain the meaning of `HealthyAt`. I originally recognized that, but it looks like a bug snuck into this PR by which I accidentally set them both together in https://github.com/longhorn/longhorn-manager/blob/792a7e9f09c79760bc86409a66395b895dbcaa2a/controller/volume_controller.go#L721-L730. This was unintentional and incorrect. I intend to set `LastHealthyAt` EVERY time we see a RW replica and `HealthyAt` ONLY when we complete a rebuild.

ejweber: I did end up adding a utility function for `FailedAt` and `LastFailedAt` since these occur often and are atomic when setting to a time value.

ejweber:
Please check `k8s/pkg/apis/longhorn/v1beta2/replica.go` in case it is not clear what is intended by the new fields.

controller/volume_controller.go
shuo-wu: I see. The salvaged replicas satisfy `r.Spec.HealthyAt != "" && r.Spec.FailedAt == ""`.  They may be invalid and can intervene in the subsequent replica cleanup executed.
And this function can filter out all failed replicas and salvaged but unverified replicas.

shuo-wu: > Sometimes, we have a replica with r.Spec.HealthyAt != "" that is completely corrupted. If it is selected for autosalvage, we clear r.Spec.FailedAt (r.Spec.FailedAt == ""). This fits our previous definition of "healthy", but we cannot rely on this definition when cleaning up stale replicas!

> The salvaged replicas satisfy r.Spec.HealthyAt != "" && r.Spec.FailedAt == "". They may be invalid and can intervene in the subsequent replica cleanup executed.
And this function can filter out all failed replicas and salvaged but unverified replicas.

Can you add more comments based on these discussions?

shuo-wu (approved):
In general LGTM

PhanLe1010 (dismissed):
LGTM

PhanLe1010:
> I initially tried to suppress these new descriptions from appearing in the CRD, but:

I don't see the harmfulness of putting the comment into CRD object (including the comment for internal field). So agree to keep it inside CRD definition

controller/engine_controller.go
james-munson: So, should the names be "LastRWAt" and "RebuiltAt"?

util/util.go
```diff
@@ -261,6 +261,22 @@ func TimestampWithinLimit(latest time.Time, ts string, limit time.Duration) bool
 	return deadline.After(latest)
 }
 
+// TimestampAfterTimestamp returns true if timestamp1 is after timestamp2. It returns false otherwise or if either
+// timestamp cannot be parsed.
```
james-munson, line 265: Sorry for the late comment, but this bothers me.  If you ask the question with one or two invalid timestamps, you get an answer as if they were valid.  If you ask the question with the (invalid) timestamps reversed, you get an apparently valid answer with the opposite conclusion, which might cause you to take a whole different path in the code.  Neither is right, maybe, but having such a "random" outcome seems like a good source of unexpected behavior.  Perhaps this should return (bool, error)?

ejweber: My soft preference is for the way I had it (mostly because it is convenient and mirrors the existing `TimestampAfterTimeout` and `TimestampWithinLimit` functions, but I can respect the complaint.

Changed the function to return an error and updated its caller and unit tests.

controller/volume_controller.go
ejweber: I added an additional explanation to `isSafeAsLastReplica`. In addition, I put additional field explanations in the CRD.

ejweber:
Squashed and rebased in preparation for final approval.

james-munson (dismissed):
Good to go.  Thanks.

james-munson (approved):

shuo-wu (approved):

shuo-wu:
@mergify backport v1.6.x

shuo-wu:
@mergify backport v1.5.x

ejweber:
@mergify backport v1.6.x

PR #2433, opened 2024-01-07
Prevent from complains in spdk_tgt when deleting a v2 volume

derekbit:

Issue longhorn/longhorn#7568

spdk_tgt complains the improper sequence of deleting engines and replicas.
Replicas should be deleted before deleting its engine.

### DIFF
--- a/controller/setting_controller.go
+++ b/controller/setting_controller.go
@@ -834,13 +834,13 @@ func (sc *SettingController) updateDataEngine(setting types.SettingName) error {
 }
 
 func (sc *SettingController) cleanupInstanceManager(dataEngine longhorn.DataEngineType) error {
-	sc.logger.Infof("Cleaning up the instance manager for %v data engine", dataEngine)
 	imMap, err := sc.ds.ListInstanceManagersBySelectorRO("", "", longhorn.InstanceManagerTypeAllInOne, dataEngine)
 	if err != nil {
 		return errors.Wrapf(err, "failed to list instance managers for cleaning up %v data engine", dataEngine)
 	}
 
 	for _, im := range imMap {
+		sc.logger.Infof("Cleaning up the instance manager for %v data engine", dataEngine)
 		if err := sc.ds.DeleteInstanceManager(im.Name); err != nil {
 			return err
 		}
--- a/controller/setting_controller.go
+++ b/controller/setting_controller.go
@@ -840,7 +840,7 @@ func (sc *SettingController) cleanupInstanceManager(dataEngine longhorn.DataEngi
 	}
 
 	for _, im := range imMap {
-		sc.logger.Infof("Cleaning up the instance manager for %v data engine", dataEngine)
+		sc.logger.Infof("Cleaning up the instance manager %v for %v data engine", im.Name, dataEngine)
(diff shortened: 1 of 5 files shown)


controller/setting_controller.go
```diff
@@ -834,13 +834,13 @@ func (sc *SettingController) updateDataEngine(setting types.SettingName) error {
...
 
 func (sc *SettingController) cleanupInstanceManager(dataEngine longhorn.DataEngineType) error {
-	sc.logger.Infof("Cleaning up the instance manager for %v data engine", dataEngine)
 	imMap, err := sc.ds.ListInstanceManagersBySelectorRO("", "", longhorn.InstanceManagerTypeAllInOne, dataEngine)
 	if err != nil {
 		return errors.Wrapf(err, "failed to list instance managers for cleaning up %v data engine", dataEngine)
 	}
 
 	for _, im := range imMap {
+		sc.logger.Infof("Cleaning up the instance manager for %v data engine", dataEngine)
```
c3y1huang, line 843:
```suggestion
		sc.logger.Infof("Cleaning up the instance manager %v for %v data engine", im.Name, dataEngine)
```

controller/volume_controller.go
```diff
@@ -350,6 +350,28 @@ func (c *VolumeController) syncVolume(key string) (err error) {
 				}
 			}
 		}
+		if datastore.IsDataEngineV2(volume.Spec.DataEngine) {
+			// To prevent from the "no such device" error in spdk_tgt,
+			// remove the raid bdev before tearing down the replicas.
+			engines, err := c.ds.ListVolumeEnginesRO(volume.Name)
```
c3y1huang, line 356: Should we consider the engine's DeletionTimestamp?

derekbit: Although `engine.Spec.DeletionTimestamp` is set, it does not mean the engine (raid bdev) is already tear down. Hence, I only check the existence of engines here.
Do you think there is any side effect?

controller/setting_controller.go
derekbit: Updated. Thank you.

controller/volume_controller.go
c3y1huang: I see. No concerns.

c3y1huang (approved):

controller/volume_controller.go
```diff
@@ -350,6 +350,28 @@ func (c *VolumeController) syncVolume(key string) (err error) {
...
+			if len(engines) > 0 {
+				replicaDeleted := false
+				for _, r := range replicas {
+					if r.DeletionTimestamp != nil {
+						replicaDeleted = true
+						break
+					}
+				}
+				if !replicaDeleted {
+					c.logger.Infof("Volume %v still has engines %v", volume.Name, engines)
```
innobead, line 369:
```go
Infof("Volume (%s) %v still has engines %v, so skip deleting its replicas until the engines have been deleted", volume.Spec.DataEngine, volume.Name, engines)
```

innobead (changes requested):
LGTM, just one comment.

controller/volume_controller.go
derekbit: Updated. Thank you.

innobead (approved):

PR #2435, opened 2024-01-08
fix(upgrade): skip upgrade path check

mantissahz:

Issue longhorn/longhorn#7539

- Add v140-to-v15x path back
-  Add a new flag `FlagUpgradeVersionCheck: "upgrade-version-check"`
    to accept the command to skip the upgrade path check.

### DIFF
--- a/app/daemon.go
+++ b/app/daemon.go
@@ -43,6 +43,7 @@ const (
 	FlagSupportBundleManagerImage = "support-bundle-manager-image"
 	FlagServiceAccount            = "service-account"
 	FlagKubeConfig                = "kube-config"
+	FlagUpgradeVersionCheck       = "upgrade-version-check"
 )
 
 func DaemonCmd() cli.Command {
@@ -80,6 +81,10 @@ func DaemonCmd() cli.Command {
 			cli.StringFlag{
 				Name:  FlagKubeConfig,
 				Usage: "Specify path to kube config (optional)",
+			},
+			cli.BoolFlag{
+				Name:  FlagUpgradeVersionCheck,
+				Usage: "Enforce version checking for upgrades. If disabled, there will be no requirement for the necessary upgrade source version",
 			},
 		},
 		Action: func(c *cli.Context) {
@@ -159,7 +164,7 @@ func startManager(c *cli.Context) error {
 		return err
 	}
 
-	if err := upgrade.Upgrade(kubeconfigPath, currentNodeID, managerImage); err != nil {
+	if err := upgrade.Upgrade(kubeconfigPath, currentNodeID, managerImage, c.Bool(FlagUpgradeVersionCheck)); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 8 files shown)


app/daemon.go
```diff
@@ -81,6 +82,10 @@ func DaemonCmd() cli.Command {
 				Name:  FlagKubeConfig,
 				Usage: "Specify path to kube config (optional)",
 			},
+			cli.BoolFlag{
+				Name:  FlagUpgradeVersionCheck,
+				Usage: "Have a warning log and an event for unrecommended upgrade path",
```
innobead, line 87: Usage should be `Enforce version checking for upgrades. If disabled, there will be no requirement for the necessary upgrade source version`

upgrade/upgrade.go
```diff
@@ -35,7 +39,7 @@
 	LeaseLockName = "longhorn-manager-upgrade-lock"
 )
 
-func Upgrade(kubeconfigPath, currentNodeID, managerImage string) error {
+func Upgrade(kubeconfigPath, currentNodeID, managerImage string, upgradeVersionCheck bool) error {
```
innobead, line 42: `enableUpgradeVersionCheck bool`

upgrade/upgrade.go
```diff
@@ -67,7 +71,13 @@
...
 	}
 
-	if err := upgradeutil.CheckUpgradePathSupported(namespace, lhClient); err != nil {
+	eventBroadcaster := record.NewBroadcaster()
+	eventBroadcaster.StartLogging(logrus.Infof)
+	// TODO: remove the wrapper when every clients have moved to use the clientset.
+	eventBroadcaster.StartRecordingToSink(&v1core.EventSinkImpl{Interface: v1core.New(kubeClient.CoreV1().RESTClient()).Events("")})
+	eventRecorder := eventBroadcaster.NewRecorder(scheme, corev1.EventSource{Component: "longhorn-upgrade"})
+
+	if err := upgradeutil.CheckUpgradePathSupported(namespace, lhClient, eventRecorder, upgradeVersionCheck); err != nil {
```
innobead, line 80: Rename `upgradeutil.CheckUpgradePath`

upgrade/util/util.go
```diff
@@ -235,6 +239,15 @@ func checkLHUpgradePathSupported(namespace string, lhClient lhclientset.Interfac
 		return fmt.Errorf("failed to upgrade since upgrading version %v is not valid", meta.Version)
 	}
 
+	if !upgradeVersionCheck {
+		if semver.Compare(lhCurrentVersion, "v1.4.0") < 0 {
+			return fmt.Errorf("failed to upgrade since downgrading from %v to %v is not supported", lhCurrentVersion, meta.Version)
```
innobead, line 244:
```go
fmt.Errorf("failed to skip the upgrade check for %s to upgrade to %s, as it only supports the current version larger than v1.4.0", lhCurrentVersion, meta.Version)
```

innobead (changes requested):
In general, LGTM. Some comments need to fix.

app/daemon.go
mantissahz: Modified. Thanks

upgrade/util/util.go
mantissahz: Modified. Thanks

upgrade/upgrade.go
mantissahz: Modified. Thanks

innobead (approved):

PR #2438, opened 2024-01-09
fix(upgrade): skip to check engine upgrade path

mantissahz:

Issue longhorn/longhorn#7539

Skip to check the engine upgrade path if the upgrade version check flag is disabled.

### DIFF
--- a/k8s/pkg/apis/longhorn/v1beta2/engineimage.go
+++ b/k8s/pkg/apis/longhorn/v1beta2/engineimage.go
@@ -5,10 +5,9 @@ import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 type EngineImageState string
 
 const (
-	EngineImageStateDeploying    = EngineImageState("deploying")
-	EngineImageStateDeployed     = EngineImageState("deployed")
-	EngineImageStateIncompatible = EngineImageState("incompatible")
-	EngineImageStateError        = EngineImageState("error")
+	EngineImageStateDeploying = EngineImageState("deploying")
+	EngineImageStateDeployed  = EngineImageState("deployed")
+	EngineImageStateError     = EngineImageState("error")
 )
 
 const (
@@ -55,6 +54,8 @@ type EngineImageStatus struct {
 	RefCount int `json:"refCount"`
 	// +optional
 	NoRefSince string `json:"noRefSince"`
+	// +optional
+	Incompatible bool `json:"incompatible"`
 	// +optional
 	// +nullable
 	Conditions []Condition `json:"conditions"`
--- a/k8s/pkg/apis/longhorn/v1beta2/engineimage.go
+++ b/k8s/pkg/apis/longhorn/v1beta2/engineimage.go
@@ -7,7 +7,9 @@ type EngineImageState string
 const (
 	EngineImageStateDeploying = EngineImageState("deploying")
 	EngineImageStateDeployed  = EngineImageState("deployed")
-	EngineImageStateError     = EngineImageState("error")
+	// EngineImageStateIncompatible is deprecated from 1.6.0. It has been replaced with `incompatible` of EngineImageStatus
+	EngineImageStateIncompatible = EngineImageState("incompatible")
+	EngineImageStateError        = EngineImageState("error")
 )
 
 const (
--- a/upgrade/util/util.go
+++ b/upgrade/util/util.go
@@ -210,7 +210,20 @@ func CreateOrUpdateLonghornVersionSetting(namespace string, lhClient *lhclientse
 }
 
 func CheckUpgradePath(namespace string, lhClient lhclientset.Interface, eventRecorder record.EventRecorder, enableUpgradeVersionCheck bool) error {
-	if err := checkLHUpgradePath(namespace, lhClient, eventRecorder, enableUpgradeVersionCheck); err != nil {
+	lhCurrentVersion, err := GetCurrentLonghornVersion(namespace, lhClient)
+	if err != nil {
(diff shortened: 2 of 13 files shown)


upgrade/util/util.go
```diff
@@ -302,7 +302,7 @@ func checkEngineUpgradePath(namespace string, lhClient lhclientset.Interface, ne
 		return err
 	}
 
-	if len(engineImageVersions) == 0 {
+	if len(engineImageVersions) == 0 || !enableUpgradeVersionCheck {
```
innobead, line 305: You can skip `!enableUpgradeVersionCheck` at the beginning of the method, but do it after querying the engine images.

innobead (changes requested):
Can we do this?

Move the following code from `checkLHUpgradePath` to the beginning of `CheckUpgradePath`? so we don't need to pass the variable to the downstream methods.  

```go
	if !enableUpgradeVersionCheck {
		if semver.Compare(lhCurrentVersion, "v1.4.0") < 0 {
			return fmt.Errorf("failed to skip the upgrade check for %s to upgrade to %s, as it only supports the current version larger than v1.4.0", lhCurrentVersion, meta.Version)
		}
		logrus.Warnf(skipUpgradeVersionCheckMsg, lhCurrentVersion, meta.Version)
		eventRecorder.Eventf(&corev1.ObjectReference{Namespace: namespace, Name: "longhorn-upgrade"}, corev1.EventTypeWarning, constant.EventReasonUpgrade, skipUpgradeVersionCheckMsg, lhCurrentVersion, meta.Version)
		return nil
	}
```

upgrade/util/util.go
mantissahz: Moved codes
```golang
if !enableUpgradeVersionCheck {
	if semver.Compare(lhCurrentVersion, "v1.4.0") < 0 {
		return fmt.Errorf("failed to skip the upgrade check for %s to upgrade to %s, as it only supports the current version larger than v1.4.0", lhCurrentVersion, meta.Version)
	}
	logrus.Warnf(skipUpgradeVersionCheckMsg, lhCurrentVersion, meta.Version)
	eventRecorder.Eventf(&corev1.ObjectReference{Namespace: namespace, Name: "longhorn-upgrade"}, corev1.EventTypeWarning, constant.EventReasonUpgrade, skipUpgradeVersionCheckMsg, lhCurrentVersion, meta.Version)
	return nil
}
```
from checkLHUpgradePath to the beginning of CheckUpgradePath.
And removed passing variables to downstream methods.
Thanks.

innobead:
There are a few things to change after the engine upgrade version enforcement was introduced, which increased `	CLIAPIMinVersion` to 8 from 3, which is the version used in 1.5, 1.4, and 1.3.

- Add back the VA creation in the upgrade path like what we did in 1.4 -> 1.5
- Add a new field `incompatible` to engineer image status and remove the original [`incompatible` state](https://github.com/longhorn/longhorn-manager/blob/2931e5b7bbc88849f67e9eb9cdab32b3fa531da4/k8s/pkg/apis/longhorn/v1beta2/engineimage.go#L10), which will make `CheckEngineImageReadiness` incorrectly check the image readiness.
- Update the code using the `EngineImageStateIncompatible1` to use the new status filed `incompatible`.

[/datastore/longhorn.go#L1746-L1761](https://github.com/longhorn/longhorn-manager/blob/2931e5b7bbc88849f67e9eb9cdab32b3fa531da4/datastore/longhorn.go#L1746-L1761)
```go
func (s *DataStore) CheckEngineImageReadiness(image string, nodes ...string) (isReady bool, err error) {
	if len(nodes) == 0 {
		return false, nil
	}
	if len(nodes) == 1 && nodes[0] == "" {
		return false, nil
	}
	ei, err := s.GetEngineImageRO(types.GetEngineImageChecksumName(image))
	if err != nil {
		return false, errors.Wrapf(err, "failed to get engine image %v", image)
	}
	if ei.Status.State != longhorn.EngineImageStateDeployed &&
		ei.Status.State != longhorn.EngineImageStateDeploying {
		logrus.Warnf("CheckEngineImageReadiness: engine image %v is in state %v", image, ei.Status.State)
		return false, nil
	}
```

@mantissahz after fixing this, update https://github.com/longhorn/longhorn/blob/master/enhancements/20230815-engine-upgrade-enforcement.md. Also, use separate commits to make the implementation clear if needed.

mantissahz:
Tested with the latest implementation:

After upgrading v1.4.x volume attaching to a workload to 1.6.x, we can live/offline upgrade the volume engine to the master-head engine image.

One warning operation:
- After the upgrade, detach the v1.4.x volume and then it should not re-attach v1.4.x volume without upgrading the engine.

innobead:
What's the reason for this warning operation?

k8s/pkg/apis/longhorn/v1beta2/engineimage.go
```diff
@@ -5,10 +5,9 @@ import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 type EngineImageState string
 
 const (
-	EngineImageStateDeploying    = EngineImageState("deploying")
-	EngineImageStateDeployed     = EngineImageState("deployed")
-	EngineImageStateIncompatible = EngineImageState("incompatible")
-	EngineImageStateError        = EngineImageState("error")
```
innobead, line 11: Let's don't remove it, but mark it as deprecated. 

Also, make an upgrade path for updating the `incompatible` field by `status::state::EngineImageStateIncompatible`.

innobead (approved):

PR #2440, opened 2024-01-10
6654 - Move deletion of backup volume attachment ticket.

james-munson:

Issue https://github.com/longhorn/longhorn/issues/6654

Ensures that the temporary volume attachment used to create a backup is detached when the backup creation is complete.

### DIFF
--- a/controller/backup_controller.go
+++ b/controller/backup_controller.go
@@ -324,6 +324,10 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
 		if err != nil {
 			return
 		}
+
+		if bc.backupInFinalState(backup) && (!backup.Status.LastSyncedAt.IsZero() || backup.Spec.SnapshotName == "") {
+			err = bc.handleAttachmentTicketDeletion(backup, backupVolumeName)
+		}
 		if reflect.DeepEqual(existingBackup.Status, backup.Status) {
 			return
 		}
@@ -339,10 +343,6 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
 				return
 			}
 		}
-		if !backup.Status.LastSyncedAt.IsZero() || backup.Spec.SnapshotName == "" {
-			err = bc.handleAttachmentTicketDeletion(backup, backupVolumeName)
-			return
-		}
 	}()
 
 	// Perform backup snapshot to the remote backup target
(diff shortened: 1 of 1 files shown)


controller/backup_controller.go
```diff
@@ -324,6 +324,10 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
 		if err != nil {
 			return
 		}
+
+		if bc.backupInFinalState(backup) && (!backup.Status.LastSyncedAt.IsZero() || backup.Spec.SnapshotName == "") {
```
PhanLe1010, line 328: Good catch that we need to check `bc.backupInFinalState(backup)` before doing the clean up. Otherwise, we might cleanup the ticket and detach the volume before the backup CR is updated to be completed.

PhanLe1010 (approved):
LGTM

PhanLe1010 (commented):
Sorry, I think we need to visit the upgrade path to cleanup the existing orphan VA tickets

controller/backup_controller.go
```diff
@@ -856,8 +856,8 @@ func (bc *BackupController) syncBackupStatusWithSnapshotCreationTimeAndVolumeSiz
 	backup.Status.SnapshotCreatedAt = snap.Created
 }
 
-func (bc *BackupController) backupNotInFinalState(backup *longhorn.Backup) bool {
-	return backup.Status.State != longhorn.BackupStateCompleted &&
-		backup.Status.State != longhorn.BackupStateError &&
-		backup.Status.State != longhorn.BackupStateUnknown
+func (bc *BackupController) backupInFinalState(backup *longhorn.Backup) bool {
```
innobead, line 859: Good change to use positive naming.

innobead:
> Sorry, I think we need to visit the upgrade path to cleanup the existing orphan VA tickets

@james-munson Can you have this in another PR?

innobead (approved):

innobead:
@mergify backport v1.5.x

controller/backup_controller.go
james-munson: Right - we did exactly that.

PhanLe1010:
Thanks @james-munson for double check and sorry for the false alarm

PR #2443, opened 2024-01-10
Incease default hugepage size for v2 data engine to 2GiB

derekbit:

Issue https://github.com/longhorn/longhorn/issues/7606


innobead (approved):

innobead:
See if we need to update longhorn/longhorn accordingly, also doc.

PR #2444, opened 2024-01-10
feat: add snapshotMaxCount and snapshotMaxSize to Volume API

FrankYang0529:

https://github.com/longhorn/longhorn/issues/7522

Support `snapshotMaxCount` and `snapshotMaxSize` in LH Volume API.

### DIFF
--- a/api/model.go
+++ b/api/model.go
@@ -60,6 +60,8 @@ type Volume struct {
 	DataEngine                       longhorn.DataEngineType                `json:"dataEngine"`
 	OfflineReplicaRebuilding         longhorn.OfflineReplicaRebuilding      `json:"offlineReplicaRebuilding"`
 	OfflineReplicaRebuildingRequired bool                                   `json:"offlineReplicaRebuildingRequired"`
+	SnapshotMaxCount                 int                                    `json:"snapshotMaxCount"`
+	SnapshotMaxSize                  string                                 `json:"snapshotMaxSize"`
 
 	DiskSelector         []string                      `json:"diskSelector"`
 	NodeSelector         []string                      `json:"nodeSelector"`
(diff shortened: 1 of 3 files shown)


derekbit (dismissed):
LGTM

api/volume.go
```diff
@@ -167,6 +167,11 @@ func (s *Server) VolumeCreate(rw http.ResponseWriter, req *http.Request) error {
 		}
 	}
 
+	snapshotMaxSize, err := util.ConvertSize(volume.SnapshotMaxSize)
+	if err != nil {
+		return fmt.Errorf("failed to parse snapshot max size %v", err)
```
derekbit, line 172: nit: 
```suggestion
		return errors.Wrap(err, "failed to parse snapshot max size")
```

api/model.go
```diff
@@ -60,6 +60,8 @@ type Volume struct {
 	DataEngine                       longhorn.DataEngineType                `json:"dataEngine"`
 	OfflineReplicaRebuilding         longhorn.OfflineReplicaRebuilding      `json:"offlineReplicaRebuilding"`
 	OfflineReplicaRebuildingRequired bool                                   `json:"offlineReplicaRebuildingRequired"`
+	SnapshotMaxCount                 int                                    `json:"snapshotMaxCount"`
+	SnapshotMaxSize                  string                                 `json:"snapshotMaxSize"`
```
innobead, line 64: Why the type is string?

innobead (commented):
Just one question.

api/volume.go
FrankYang0529: Updated it. Thanks.

api/model.go
FrankYang0529: It's like `Size`. We use string for it.

https://github.com/longhorn/longhorn-manager/blob/591946024a333421dc446afeea62618525fdb42a/api/model.go#L33

innobead: I see. Should revisit why we did that way later.

innobead (approved):

FrankYang0529:
> Should this also return the actions for updating the values? It seems that they are currently missing

Yeah, let me add it. Thanks.

PR #2445, opened 2024-01-10
vendor: update dependencies

derekbit:
Longhorn 7581

Issue Longhorn/longhorn#7581


Dockerfile.dapper1806645496
```diff
@@ -0,0 +1,2 @@
+FROM longhorn-manager:snapshot-backup
+COPY . /go/src/github.com/longhorn/longhorn-manager/
```
innobead, line 2: Remove temp dapper files.

innobead (changes requested):

Dockerfile.dapper1806645496
derekbit: Removed. Thanks.

innobead (approved):

PR #2447, opened 2024-01-11
feat: add updateSnapshotMaxCount and updateSnapshotMaxSize actions

FrankYang0529:
#### Which issue(s) this PR fixes:

https://github.com/longhorn/longhorn/issues/7522

#### What this PR does / why we need it:

add updateSnapshotMaxCount and updateSnapshotMaxSize to Volume actions

PR #2451, opened 2024-01-12
feat: add updateSnapshotMaxCount and updateSnapshotMaxSize to schema

FrankYang0529:

https://github.com/longhorn/longhorn/issues/7522

Addressed comment in https://github.com/longhorn/longhorn-manager/pull/2447#issuecomment-1887221111.

### DIFF
--- a/client/generated_update_snapshot_max_count_input.go
+++ b/client/generated_update_snapshot_max_count_input.go
@@ -7,7 +7,7 @@ const (
 type UpdateSnapshotMaxCountInput struct {
 	Resource `yaml:"-"`
 
-	SnapshotMaxCount int `json:"snapshotMaxCount,omitempty" yaml:"snapshot_max_count,omitempty"`
+	SnapshotMaxCount int64 `json:"snapshotMaxCount,omitempty" yaml:"snapshot_max_count,omitempty"`
 }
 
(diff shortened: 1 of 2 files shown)


innobead (approved):

innobead:
@mergify backport v1.6.x

client/generated_update_snapshot_max_count_input.go
```diff
@@ -7,7 +7,7 @@ const (
 type UpdateSnapshotMaxCountInput struct {
 	Resource `yaml:"-"`
 
-	SnapshotMaxCount int `json:"snapshotMaxCount,omitempty" yaml:"snapshot_max_count,omitempty"`
+	SnapshotMaxCount int64 `json:"snapshotMaxCount,omitempty" yaml:"snapshot_max_count,omitempty"`
```
FrankYang0529, line 10: It looks like go-rancher use `int64` instead of `int`. `NumberOfReplicas` also has similar behavior.

https://github.com/longhorn/longhorn-manager/blob/98189bfad382f206c0a6ba13e1dbde6283707795/k8s/pkg/apis/longhorn/v1beta2/volume.go#L277

https://github.com/longhorn/longhorn-manager/blob/98189bfad382f206c0a6ba13e1dbde6283707795/client/generated_volume.go#L60

PR #2454, opened 2024-01-12
fix: use engineCLIAPIVersion for snapshot max count & size

FrankYang0529:

https://github.com/longhorn/longhorn/issues/7642

Only append snapshot max count and size if `engineCLIAPIVersion` is larger than `10`.


derekbit (approved):

innobead (approved):

innobead:
Does this mean we missed incompatible engine image e2e tests?

innobead:
@mergify backport v1.6.x

PR #2457, opened 2024-01-12
fix: use engineCLIAPIVersion for snapshot max count & size

FrankYang0529:

https://github.com/longhorn/longhorn/issues/7649

We updated engine process in last change, but not replica. Added it in this PR.


innobead (approved):

innobead:
@mergify backport v1.6.x

PR #2460, opened 2024-01-12
Update the description of v2-data-engine setting

derekbit:

Issue Longhorn/longhorn#7655

v1 and v2 date engines are separate, so the warning `When applying the setting, Longhorn will restart all instance-manager pods.` is not required.


innobead (approved):

innobead:
@mergify backport v1.6.x

PR #2463, opened 2024-01-13
Fix bug uninstallation stuck if the MutatingWebhookConfigurations or ValidatingWebhookConfigurations already deleted

PhanLe1010:
The old code assumes that validatingCfg is nil when ValidatingWebhookConfigurations().Get() return error NotFound. This is an incorrect assumption, the validatingCfg is actually an empty non-nil object in this case and the old logic attempted to delete a non-exist object https://github.com/longhorn/longhorn-manager/blob/41d50c237d2f91b9acfc57cedb6fa0c18970d7e6/controller/uninstall_controller.go#L946C1-L951

longhorn/longhorn#7657


innobead (approved):

innobead:

PR #2473, opened 2024-01-14
fix(setting): flooding error message

mantissahz:

Issue longhorn/longhorn#7654

https://github.com/longhorn/longhorn/issues/7649#issuecomment-1889282614

Move returning logs from `errors.Wrapf` to `types.ErrorInvalidState` when synchronizing the danger zone setting with volumes attached.

Move the log level from `Error` to `Trace`.


innobead (approved):

innobead:
@mergify backport v1.6.x

PR #2475, opened 2024-01-14
Prevent flooding warning messages

derekbit:

Issue Longhorn/longhorn#7619

Prevent flooding warning messages "Found more than 1 xxx instance manager ...". It is a side effect of the solution to Longhorn 7619. To mitigate the issue, change the level to debug level.


innobead (approved):

innobead:

PR #2479, opened 2024-01-15
Crypto device mapper of RWX volume with migratable=true is not cleaned up

derekbit:

Issue  Longhorn/longhorn#7678


c3y1huang (approved):

innobead (approved):

derekbit:
Backport to v1.5.x and v1.4.x is not required, because it is an encrypted RWX (migratable=true) block device is a new feature in v1.6.0.

PR #2480, opened 2024-01-15
fix(engineimage): engine image incompatible

mantissahz:

Issue longhorn/longhorn#7631

Allow to modify a volume, create/update an engine, or create/update a replica with its engine image incompatible when the volume is attached.

Not affect to the engine image `master-head`:
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/5985

### DIFF
--- a/webhook/resources/engine/validator.go
+++ b/webhook/resources/engine/validator.go
@@ -60,10 +60,6 @@ func (e *engineValidator) Create(request *admission.Request, newObj runtime.Obje
 		return err
 	}
 
-	if err := e.ds.CheckDataEngineImageCompatiblityByImage(engine.Spec.Image, engine.Spec.DataEngine); err != nil {
-		return werror.NewInvalidError(err.Error(), "engine.spec.image")
-	}
-
 	return e.validateNumberOfEngines(engine, volume)
 }
 
@@ -81,12 +77,6 @@ func (e *engineValidator) Update(request *admission.Request, oldObj runtime.Obje
 		if oldEngine.Spec.DataEngine != newEngine.Spec.DataEngine {
 			err := fmt.Errorf("changing data engine for engine %v is not supported", oldEngine.Name)
 			return werror.NewInvalidError(err.Error(), "")
-		}
-	}
-
-	if oldEngine.Spec.Image != newEngine.Spec.Image {
-		if err := e.ds.CheckDataEngineImageCompatiblityByImage(newEngine.Spec.Image, newEngine.Spec.DataEngine); err != nil {
-			return werror.NewInvalidError(err.Error(), "engine.spec.image")
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


datastore/longhorn.go
```diff
@@ -1723,7 +1723,7 @@ func (s *DataStore) CheckDataEngineImageCompatiblityByImage(image string, dataEn
 		return err
 	}
 
-	if engineImage.Status.Incompatible {
+	if engineImage.Status.Incompatible && volume.Status.State != longhorn.VolumeStateAttached {
```
innobead, line 1726: Why do you check whether the volume is attached?

innobead (changes requested):
As discussed, I think what you need to do is just remove validation from the engine & replica, as we only need to prevent the following two cases.

- new volumes with incompatible image should not be created
- existing volumes updated with incompatible image should not be updated.

datastore/longhorn.go
mantissahz: Because the volume can be attached, we should allow the rebuilding or expanding actions.

>As discussed, I think what you need to do is just remove validation from the engine & replica, as we only need to prevent the following two cases.
>- new volumes with incompatible image should not be created
>- existing volumes updated with incompatible image should not be updated.

Yes, I will update the commit as we discussed.

innobead: > Because the volume can be attached, we should allow the rebuilding or expanding actions.

This is more like a workaround .. it would make `CheckDataEngineImageCompatiblityByImage` not work as expected.

innobead (changes requested):
Please also remove the update validation for the engine and replica as well. If an existing volume with an incompatible image undergoes an engine image upgrade, and if the upgrade fails, the images of the engine and replicas would be rolled back to the original incompatible image. In this case, the rollback will fail due to a validation error.

[/controller/volume_controller.go#L2735-L2741](https://github.com/longhorn/longhorn-manager/blob/98189bfad382f206c0a6ba13e1dbde6283707795/controller/volume_controller.go#L2735-L2741)
```go
	// If volume is detached accidentally during the live upgrade,
	// the live upgrade info and the inactive replicas are meaningless.
	if v.Status.State == longhorn.VolumeStateDetached {
		if e.Spec.Image != v.Spec.Image {
			e.Spec.Image = v.Spec.Image
			e.Spec.UpgradedReplicaAddressMap = map[string]string{}
		}
```

mantissahz:
>Please also remove the update validation for the engine and replica as well. If an existing volume with an incompatible image undergoes an engine image upgrade, and if the upgrade fails, the images of the engine and replicas would be rolled back to the original incompatible image. In this case, the rollback will fail due to a validation error.

Removed. Thanks.

innobead (approved):

innobead:
@mergify backport v1.6.x

PR #2484, opened 2024-01-15
fix(dataengine): separately update the setting for IMs CPU.

mantissahz:

Issue longhorn/longhorn#7676

Check if volumes are attached for v1 and v2 dataEngine separately when updating the setting `GuaranteedInstanceManagerCPU` or `V2DataEngineGuaranteedInstanceManagerCPU`.

### DIFF
--- a/controller/setting_controller.go
+++ b/controller/setting_controller.go
@@ -272,8 +272,6 @@ func (sc *SettingController) syncDangerZoneSettingsForManagedComponents(settingN
 	dangerSettingsRequiringAllVolumesDetached := []types.SettingName{
 		types.SettingNameTaintToleration,
 		types.SettingNameSystemManagedComponentsNodeSelector,
-		types.SettingNameGuaranteedInstanceManagerCPU,
-		types.SettingNameV2DataEngineGuaranteedInstanceManagerCPU,
 		types.SettingNamePriorityClass,
 		types.SettingNameStorageNetwork,
 	}
@@ -297,10 +295,6 @@ func (sc *SettingController) syncDangerZoneSettingsForManagedComponents(settingN
 			if err := sc.updateNodeSelector(); err != nil {
 				return err
 			}
-		case types.SettingNameGuaranteedInstanceManagerCPU, types.SettingNameV2DataEngineGuaranteedInstanceManagerCPU:
-			if err := sc.updateInstanceManagerCPURequest(); err != nil {
-				return err
-			}
 		case types.SettingNamePriorityClass:
 			if err := sc.updatePriorityClass(); err != nil {
 				return err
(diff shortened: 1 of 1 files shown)


controller/setting_controller.go
```diff
@@ -319,12 +313,35 @@ func (sc *SettingController) syncDangerZoneSettingsForManagedComponents(settingN
...
+			}
+
+			detached, _, err := sc.ds.AreAllVolumesDetached(dataEngine)
+			if err != nil {
+				return errors.Wrapf(err, "failed to check volume detachment for %v setting update", settingName)
+			}
+
+			if !detached {
+				return &types.ErrorInvalidState{Reason: fmt.Sprintf("failed to apply %v setting to Longhorn components when there are attached volumes. It will be eventually applied", settingName)}
+			}
```
derekbit, line 340: In general, LGTM.
Jus one question: when to use `return errors.Wrapf`  or `return &types.ErrorInvalidState`

mantissahz: When returning the error to `syncSetting` and `handleErr` will receive and handle the error, and `errors.Wrapf` will be `error` log level and `types.ErrorInvalidState` will be `trace` log level. That's the method we prevent logs from flooding.
Any concerns?

derekbit (dismissed):

controller/setting_controller.go
```diff
@@ -319,12 +313,35 @@ func (sc *SettingController) syncDangerZoneSettingsForManagedComponents(settingN
...
+
+			if err := sc.updateInstanceManagerCPURequest(); err != nil {
```
innobead, line 342: The `updateInstanceManagerCPURequest` will trigger the recreation of all IM pods regardless of whether it's v1 or v2. For example, if the v1 setting is changed but v2 is also enabled, then the v2 IM pods will be restarted as well.

innobead (changes requested):

controller/setting_controller.go
mantissahz: No, it will not restart v2 IM pods.
The changing resource request is handled by [GetInstanceManagerCPURequirement](https://github.com/longhorn/longhorn-manager/blob/master/controller/setting_controller.go#L1238C74-L1238C74) in `updateInstanceManagerCPURequest` and it will handle v1 and v2 volumes separately

innobead: I meant all IM pods will be walked through instead of filtering the data engine type initially.

[/controller/setting_controller.go#L1208-L1233](https://github.com/longhorn/longhorn-manager/blob/98189bfad382f206c0a6ba13e1dbde6283707795/controller/setting_controller.go#L1208-L1233)
```go
func (sc *SettingController) updateInstanceManagerCPURequest() error {
	imPodList, err := sc.ds.ListInstanceManagerPods()
	if err != nil {
		return errors.Wrap(err, "failed to list instance manager pods for toleration update")
	}
	imMap, err := sc.ds.ListInstanceManagersRO()
	if err != nil {
		return err
	}

	for _, imPod := range imPodList {
		if _, exists := imMap[imPod.Name]; !exists {
			continue
		}
		lhNode, err := sc.ds.GetNode(imPod.Spec.NodeName)
		if err != nil {
			return err
		}
		if types.GetCondition(lhNode.Status.Conditions, longhorn.NodeConditionTypeReady).Status != longhorn.ConditionStatusTrue {
			continue
		}

		resourceReq, err := GetInstanceManagerCPURequirement(sc.ds, imPod.Name)
		if err != nil {
			return err
		}
```

mantissahz: Modified to pass the `dataEngine` to `updateInstanceManagerCPURequest` and then List the pods and the instance managers by the `dataEgnine`.
Thanks.

controller/setting_controller.go
```diff
@@ -1213,12 +1230,12 @@ func (sc *SettingController) enqueueSettingForBackupTarget(obj interface{}) {
 }
 
 // updateInstanceManagerCPURequest deletes all instance manager pods immediately with the updated CPU request.
-func (sc *SettingController) updateInstanceManagerCPURequest() error {
-	imPodList, err := sc.ds.ListInstanceManagerPods()
+func (sc *SettingController) updateInstanceManagerCPURequest(dataEngine longhorn.DataEngineType) error {
+	imPodList, err := sc.ds.ListInstanceManagerPodsBy("", "", longhorn.InstanceManagerTypeAllInOne, dataEngine)
```
innobead, line 1234: For v1, should we query all instance types including engine and replica IM pos in case there could be some old IM pods running? 

```go
ListInstanceManagerPodsBy("", "", dataEngine, longhorn.InstanceManagerTypeAllInOne, InstanceManagerTypeEngine, InstanceManagerTypeReplica)
```

Like

[/controller/kubernetes_secret_controller.go#L214-L224](https://github.com/longhorn/longhorn-manager/blob/98189bfad382f206c0a6ba13e1dbde6283707795/controller/kubernetes_secret_controller.go#L214-L224)
```go
	replicaInstanceManagerPods, err := ks.ds.ListInstanceManagerPodsBy(ks.controllerID, "", longhorn.InstanceManagerTypeReplica, "")
	if err != nil {
		return err
	}
	pods := append(managerPods, replicaInstanceManagerPods...)

	aioInstanceManagerPods, err := ks.ds.ListInstanceManagerPodsBy(ks.controllerID, "", longhorn.InstanceManagerTypeAllInOne, "")
	if err != nil {
		return err
	}
	pods = append(pods, aioInstanceManagerPods...)
```

innobead (changes requested):

controller/setting_controller.go
mantissahz: >For v1, should we query all instance types including engine and replica IM pos in case there could be some old IM pods running?

IIUC, 
- If the upgrade is from v1.4.4 or before and volumes are still attached (it needs to disable the upgrade 
 version check), the setting can not be applied to them (im-e or im-r).
- If volumes are detached, there are no im-e or im-r.

innobead (approved):

PR #2493, opened 2024-01-17
feat: shift NFS default version to 4.2 for RWX mount.

james-munson:

Issue https://github.com/longhorn/longhorn/issues/7638

NFS v4.2 is a better default option, with more capabilities than the current v4.1.  The user can still force v4.1 if necessary.

Passes regression rwx tests.

### DIFF
--- a/csi/node_server.go
+++ b/csi/node_server.go
@@ -251,7 +251,7 @@ func (ns *NodeServer) nodeStageSharedVolume(volumeID, shareEndpoint, targetPath 
 	export := fmt.Sprintf("%s:%s", server, exportPath)
 
 	defaultMountOptions := []string{
-		"vers=4.1",
+		"vers=4.2",
 		"noresvport",
 		//"sync",    // sync mode is prohibitively expensive on the client, so we allow for host defaults
 		//"intr",


ejweber (approved):
LGTM.

innobead (approved):
4.2 has been supported since Linux kernel 3.10.

innobead:

innobead:
@james-munson Let's resolve this comment from @derekbit https://github.com/longhorn/longhorn/issues/7638#issuecomment-1898054075. If test cases pass, then we can merge this PR. 

Besides, @derekbit has some suggestions about adding test cases with specific NFS version option, so please check with him later. That can be improved later.

james-munson:
Not sure what is being asked for in the issue [comment](https://github.com/longhorn/longhorn/issues/7638#issuecomment-1898054075).  If it is to run the existing rwx cases, that has already been done.  If it is to add new cases with storageclasses that have nfsOptions overrides, I'm on it.

derekbit:
Sorry, I was not aware of the PR.
v4.1 is still popular in most environment from previous feedback.
We can mention Longhorn can support v4.2 by changing mount option but still keep 4.1 as a default value.
WDYT?
cc @innobead @james-munson

innobead:
I think it's ok to revert back if this is too aggressive but what's your original idea? Isn't it to have 4.2 by default or actually have it as optional but 4.1 is still the default?

BTW, IIRC, 4.2 has been supported in most recent distro kernels, or still have a concern for older kernels?

innobead:
Also don't think we should backport this, even we decide to change. That was a wrong backport as well..

derekbit:
> BTW, IIRC, 4.2 has been supported in most recent distro kernels, or still have a concern for older kernels?

Ideally, yes. But according to the previous experience, NFS v4.2 in kernel config is not always enabled by users. The concrete reason is unknown.

> Isn't it to have 4.2 by default or actually have it as optional but 4.1 is still the default?

No. My original idea is not using v4.2 as a default value. Just make them both supported.
https://github.com/longhorn/longhorn/issues/7638#issue-2077968428

Sorry for that I didn't mention the compatibility issue. Any change should address the compatibility rather than an aggressive change.

innobead:
Sounds good. It's a timely raise. If having concerns, we should clarify it, even it has been merged. 

Let's handle the revert next week.

PR #2494, opened 2024-01-18
fix(csi): volume manager failed to find vol_data.json after kubelet restart

c3y1huang:

Issue longhorn/longhorn#7694

See the *So, What Happened?* section in https://github.com/longhorn/longhorn/issues/7694#issuecomment-1893094315.

`None`

`None`

### DIFF
--- a/csi/node_server.go
+++ b/csi/node_server.go
@@ -18,6 +18,7 @@ import (
 
 	"k8s.io/mount-utils"
 
+	corev1 "k8s.io/api/core/v1"
 	utilexec "k8s.io/utils/exec"
 
 	"github.com/longhorn/longhorn-manager/csi/crypto"
@@ -129,7 +130,7 @@ func (ns *NodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 	// Check volume attachment status
 	if datastore.IsDataEngineV1(longhorn.DataEngineType(volume.DataEngine)) {
 		if volume.State != string(longhorn.VolumeStateAttached) || volume.Controllers[0].Endpoint == "" {
-			log.Infof("Volume %v hasn't been attached yet, unmounting potential mount point %v", volumeID, targetPath)
+			log.WithField("state", volume.State).Infof("Volume %v hasn't been attached yet, unmounting potential mount point %v", volumeID, targetPath)
 			if err := unmount(targetPath, mounter); err != nil {
 				log.WithError(err).Warnf("Failed to unmount targetPath %v", targetPath)
 			}
@@ -139,6 +140,11 @@ func (ns *NodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 
 	if !volume.Ready {
 		return nil, status.Errorf(codes.Aborted, "volume %s is not ready for workloads", volumeID)
+	}
+
+	podsStatus := ns.collectWorkloadPodsStatus(volume, log)
+	if len(podsStatus[corev1.PodPending]) == 0 {
+		return nil, status.Errorf(codes.Aborted, "no %v workload pods for volume %v to be mounted: %+v", corev1.PodPending, volumeID, podsStatus)
 	}
 
 	if volumeCapability.GetBlock() != nil {
@@ -222,6 +228,17 @@ func (ns *NodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 	}
 
 	return &csi.NodePublishVolumeResponse{}, nil
+}
+
+func (ns *NodeServer) collectWorkloadPodsStatus(volume *longhornclient.Volume, log *logrus.Entry) map[corev1.PodPhase][]string {
+	podsStatus := map[corev1.PodPhase][]string{}
+
+	for _, workload := range volume.KubernetesStatus.WorkloadsStatus {
+		phase := corev1.PodPhase(workload.PodStatus)
+		podsStatus[phase] = append(podsStatus[phase], workload.PodName)
+	}
+
+	return podsStatus
 }
 
 func (ns *NodeServer) nodeStageSharedVolume(volumeID, shareEndpoint, targetPath string, mounter mount.Interface, customMountOptions []string) error {
--- a/csi/util.go
+++ b/csi/util.go
@@ -294,7 +294,7 @@ func syncMountPointDirectory(targetPath string) error {
 // the underlying implementation utilizes mounter.IsLikelyNotMountPoint so it cannot detect bind mounts
 func ensureMountPoint(path string, mounter mount.Interface) (bool, error) {
 	logrus.Infof("Trying to ensure mount point %v", path)
-	notMnt, err := mount.IsNotMountPoint(mounter, path)
+	isMnt, err := mounter.IsMountPoint(path)
 	if os.IsNotExist(err) {
 		return false, os.MkdirAll(path, 0750)
 	}
@@ -318,7 +318,7 @@ func ensureMountPoint(path string, mounter mount.Interface) (bool, error) {
 		return false, fmt.Errorf("unmounted existing corrupt mount point %v", path)
 	}
 
-	return !notMnt, err
+	return isMnt, err
 }
 
 func unmount(path string, mounter mount.Interface) (err error) {


csi/node_server.go
```diff
@@ -224,6 +230,17 @@ func (ns *NodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
...
+func (ns *NodeServer) collectWorkloadPodsStatus(volume *longhornclient.Volume, log *logrus.Entry) map[corev1.PodPhase][]string {
+	podsStatus := map[corev1.PodPhase][]string{}
+
+	for _, workload := range volume.KubernetesStatus.WorkloadsStatus {
+		phase := corev1.PodPhase(workload.PodStatus)
+		podsStatus[phase] = append(podsStatus[phase], workload.PodName)
+	}
+
+	return podsStatus
+}
```
shuo-wu, line 242: Do you test the RWX volume with multiple pods (part of them are running)?

shuo-wu:
>   3. The volume manager initiated the [reconcile loop starting with the unmounting process](https://github.com/kubernetes/kubernetes/blob/v1.27.1/pkg/kubelet/volumemanager/reconciler/reconciler.go#L47-L52) ensuring [volumes that should be unmounted are unmounted by comparing the actual and desired state of the world](https://github.com/kubernetes/kubernetes/blob/v1.27.1/pkg/kubelet/volumemanager/reconciler/reconciler_common.go#L162-L163). This step is blocking because the taint manager marked the mounted pod for deletion but Longhorn already unmounted the target mount point due to the [detection of a corrupt mount point](https://github.com/longhorn/longhorn-manager/blob/v1.6.0-rc1/csi/node_server.go#L204-L208).

>As a result, the affected pod hangs in the terminating state with the volume still attached but the target mount point unmounted, and Kubelet is unable to move past the unmounting stage of the reconcile loop.

Could you explain why kubelet unmounting will be block and why the CSI `NodePublishVolume` erroring out added by this PR can help kubelet volume manager `move past the unmounting stage of the reconcile loop`? Inside [the unmount](https://github.com/kubernetes/kubernetes/blob/eb1ae05cf040346bdb197490ef74ed929fdf60b7/pkg/volume/util/operationexecutor/operation_executor.go#L985) function, it seems that the [`TearDown()`](https://github.com/kubernetes/kubernetes/blob/eb1ae05cf040346bdb197490ef74ed929fdf60b7/pkg/volume/util/operationexecutor/operation_generator.go#L864C33-L864C43) will call [`csi.NodeUnpublishVolume`](https://github.com/kubernetes/kubernetes/blob/eb1ae05cf040346bdb197490ef74ed929fdf60b7/pkg/volume/csi/csi_mounter.go#L433), which should succeed.

c3y1huang:
> Could you explain why kubelet unmounting will be block and why the CSI `NodePublishVolume` erroring out added by this PR can help kubelet volume manager `move past the unmounting stage of the reconcile loop`? Inside [the unmount](https://github.com/kubernetes/kubernetes/blob/eb1ae05cf040346bdb197490ef74ed929fdf60b7/pkg/volume/util/operationexecutor/operation_executor.go#L985) function, it seems that the [`TearDown()`](https://github.com/kubernetes/kubernetes/blob/eb1ae05cf040346bdb197490ef74ed929fdf60b7/pkg/volume/util/operationexecutor/operation_generator.go#L864C33-L864C43) will call [`csi.NodeUnpublishVolume`](https://github.com/kubernetes/kubernetes/blob/eb1ae05cf040346bdb197490ef74ed929fdf60b7/pkg/volume/csi/csi_mounter.go#L433), which should succeed.

The [error](https://github.com/kubernetes/kubernetes/blob/v1.27.1/pkg/volume/csi/csi_plugin.go#L445) happened at [volumePlugin.NewUnmounter](https://github.com/kubernetes/kubernetes/blob/v1.27.1/pkg/volume/util/operationexecutor/operation_generator.go#L836) before it had the chance to [TearDown](https://github.com/kubernetes/kubernetes/blob/v1.27.1/pkg/volume/util/operationexecutor/operation_generator.go#L855).

The calls to `NodePublishVolume` and `NodeUnpublishVolume` rely on the actual and desired states of the world. However, in the scenario of the Kublet restart, it seems that the etcd might not have been recorded correctly, leading to incorrect calls while attempting to correct it.

In the support-bundle of the volume state, I noticed that the pod status is already marked as `Failed`, or sometimes `Running` during other test runs. This indicates that the `NodePulishVolume` has already been invoked and processed. In such scenarios, even if the mount point is corrupted, we should not rely on the unmounting in `NodePublishVolume`, instead, we should allow the volume manager to handle the volume lifecycle through its reconcile (`Teardown`) process. Therefore, this fix ensures some level of idempotence when `NodePublishVolume` is somehow called again.

csi/node_server.go
c3y1huang: The `NodePublishVolume` should only proceed with the target mounting when the pod is being initiated (`Pending` state), AFAIK, the volume attachment doesn't occur in any other state. So this condition should also apply to RWX volume workloads.

For confirmation, I've tested the following scenario:
1. Deploy [rwx-nginx-deployment.yaml](https://github.com/longhorn/longhorn/blob/master/examples/rwx/rwx-nginx-deployment.yaml).
1. Wait until all workload pods are in the running state.
1. Delete one of the workload pods.
1. Wait for the deleted workload pod to restart and reach the running state.

c3y1huang:
Checking to see if [this](https://github.com/kubernetes/kubernetes/issues/116970) is related to the behavior of the pod after being deleted by the taint manager.

shuo-wu:
YES, [this](https://github.com/kubernetes/kubernetes/blob/4c9411232e10168d7b050c49a1b59f6df9d7ea4b/pkg/volume/csi/csi_plugin.go#L443) is where the error is printed out. But later on somehow kubelet called `NodePublishVolume` and resolved the corrupted mount point. Then how can the taint manager deleting the pod lead to the recovery blocking?

csi/node_server.go
shuo-wu: I mean, if this fix the handle the below RWX case:
1. Have a RWX volume with multiple running pods 
2. Kubelet on one node (volume attached node, or a normal node containing one running pod) is stopped for a while (in or out of the pod eviction time)
3. Kubelet get back and try to recover the RWX volume

c3y1huang:
> YES, [this](https://github.com/kubernetes/kubernetes/blob/4c9411232e10168d7b050c49a1b59f6df9d7ea4b/pkg/volume/csi/csi_plugin.go#L443) is where the error is printed out. But later on somehow kubelet called `NodePublishVolume` and resolved the corrupted mount point.

The error was flooding the kubelet log which is an indication that it was not resolved. Please share what leads you to think that the `NodePublishVolume` was called and the issue has been addressed.

> Then how can the taint manager deleting the pod lead to the recovery blocking?

I am not sure about this part. The issue wasn't seen in Kubernetes 1.26 but has been observed since 1.27. I think it might be related to the actual and desired state of the pod.

shuo-wu:
> Please share what leads you to think that the NodePublishVolume was called and the issue has been addressed.

Oh sorry. I thought clearing up the corrupted mount point would help eliminate the error but NO. `NewUnmounter.loadVolumeData` still returns errors even when there is nothing mounted on the volume device path.

csi/node_server.go
c3y1huang: 1. Create [rwx-nginx-deployment.yaml](https://github.com/longhorn/longhorn/blob/master/examples/rwx/rwx-nginx-deployment.yaml)
    ```bash
    curl https://raw.githubusercontent.com/longhorn/longhorn/master/examples/rwx/rwx-nginx-deployment.yaml | kubectl apply -f -
    ```
1. Wait for deployments running 
    ```bash
    > k get deployments.apps 
    NAME       READY   UP-TO-DATE   AVAILABLE   AGE
    rwx-test   3/3     3            3           36s
  
    > k -n longhorn-system get volume
    NAME                                       DATA ENGINE   STATE      ROBUSTNESS   SCHEDULED   SIZE         NODE            AGE
    pvc-8893d93f-f4dd-43ac-ade1-035a08b647a6   v1            attached   healthy                  1073741824   ip-10-0-2-121   40s
    ```
1. Stop kubelet for the volume node for 360 seconds
    ```bash
    > ssh ec2-user@10.0.2.121
    > systemctl stop k3s-agent && sleep 360 && systemctl start k3s-agent
    ```
1. Check deployment pods running
    ```bash
    > k get pod
    NAME                        READY   STATUS    RESTARTS   AGE
    rwx-test-79d5494fd8-9wpdc   2/2     Running   0          10m
    rwx-test-79d5494fd8-vpfsz   2/2     Running   0          10m
    rwx-test-79d5494fd8-zhvqd   2/2     Running   0          72s
    ```

c3y1huang: Is there any reason that you are concerned that this change might not work for RWX volume and its workloads?

shuo-wu:
After a discussion with Chin-Ya, I realized that not cleaning up the corrupted mount point in advance is probably the key point:
After kubelet restart, we can roughly consider the recovery as 2 phases:
1. Phase 1, incorrect reconcile phase. Due to the expired/wrongly recorded actual & desired states of the world, kubelet calls some CSI calls like `NodePublishVolume` incorrectly.
2. Phase 2, correct cleanup stage for the terminating pod. Here kubelet will try to read `vol_data.json` from the mount point then unmount the volume.

Without the fix, `NodePublishVolume` in phase 1 will help clean up the corrupted mount point. Then at the beginning of phase 2, kubelet cannot find file `vol_data.json` then keeps failing at the unmounting handling. 
With the fix, phase 1 will do nothing, hence the corrupted mount point is still there at the beginning of phase 2. Then kubelet **is probably able to read file `vol_data.json` from the corrupted mount point** during the unmounting.

shuo-wu (approved):

innobead (approved):

innobead:

csi/node_server.go
```diff
@@ -141,6 +142,11 @@ func (ns *NodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 		return nil, status.Errorf(codes.Aborted, "volume %s is not ready for workloads", volumeID)
 	}
 
+	podsStatus := ns.collectWorkloadPodsStatus(volume, log)
+	if len(podsStatus[corev1.PodPending]) == 0 {
+		return nil, status.Errorf(codes.Aborted, "no %v workload pods for volume %v to be mounted: %+v", corev1.PodPending, volumeID, podsStatus)
+	}
+
```
derekbit, line 149: @c3y1huang 
Should it error out here?

PR #2500, opened 2024-01-19
fix(volume): only restart pod when it is ro and the option is not ro

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/7728

StorageClass can specify the mountOption.
User may intend to mount the volume as read-only by giving "ro" in the mountOptions in StorageClass
We should only restart the Pod when it is read-only and there is no "ro" in mountOption

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -1456,7 +1456,28 @@ func (c *VolumeController) requestRemountIfFileSystemReadOnly(v *longhorn.Volume
 	if v.Status.State == longhorn.VolumeStateAttached && e.Status.CurrentState == longhorn.InstanceStateRunning {
 		fileSystemReadOnlyCondition := types.GetCondition(e.Status.Conditions, imtypes.EngineConditionFilesystemReadOnly)
 
-		if fileSystemReadOnlyCondition.Status == longhorn.ConditionStatusTrue {
+		isPVMountOptionReadOnly := false
+		kubeStatus := v.Status.KubernetesStatus
+		if kubeStatus.PVName != "" {
+			pv, err := c.ds.GetPersistentVolumeRO(kubeStatus.PVName)
+			if err != nil {
+				if apierrors.IsNotFound(err) {
+					return
+				}
+				log.WithError(err).Warnf("Failed to get PV when checking the mount option of the volume")
+				return
+			}
+			if pv != nil {
+				for _, opt := range pv.Spec.MountOptions {
+					if opt == "ro" {
+						isPVMountOptionReadOnly = true
+						break
+					}
+				}
+			}
+		}
+
+		if fileSystemReadOnlyCondition.Status == longhorn.ConditionStatusTrue && !isPVMountOptionReadOnly {
 			v.Status.RemountRequestedAt = c.nowHandler()
 			log.Infof("Volume request remount at %v due to engine detected read-only filesystem", v.Status.RemountRequestedAt)
 			msg := fmt.Sprintf("Volume %s requested remount at %v due to engine detected read-only filesystem", v.Name, v.Status.RemountRequestedAt)


ChanYiLin:
Testing

controller/volume_controller.go
```diff
@@ -1456,7 +1456,23 @@ func (c *VolumeController) requestRemountIfFileSystemReadOnly(v *longhorn.Volume
...
+			}
+			for _, opt := range pv.Spec.MountOptions {
+				if opt == "ro" {
+					isPVMountOptionReadOnly = true
+					break
+				}
+			}
+		}
+
+		if fileSystemReadOnlyCondition.Status == longhorn.ConditionStatusTrue && !isPVMountOptionReadOnly {
```
innobead, line 1475: When a mount point is ready-only, it can be:
- mount with `ro`
- volume engine is unresponsive or error

What we assumed is the second case, but how about that the mount point is `ro` and the second case happens, how to handle this mixed case?

innobead: or do we try to check if the read operation is working? if yes, then we can differentiate that.

ChanYiLin: volume engine crashing will be handled in other place
https://github.com/longhorn/longhorn-manager/blob/e355c6364a2abbd30b3652fbf47f1ad0ab113797/controller/volume_controller.go#L1398-L1424
For that case we will request remount no matter what, read-only is only a side effect of engine crashing

Here we catch other unexpected "ro" cases

ChanYiLin: you can check my HackWeek Note for how we handle engine crash
https://docs.google.com/presentation/d/1aFtPX5yY53TGbndy3D2IRs5SrBxZigBRCiuLt3fMaW4/edit#slide=id.g26073da28eb_0_530

ChanYiLin:
With this PR, even the Engine detect that the filesystem is "ro"
Volume won't get remount request
```
$ k get lhe -n longhorn-system -oyaml | grep -A 15 "conditions:"
    conditions:
    - lastProbeTime: ""
      lastTransitionTime: "2024-01-19T08:25:06Z"
      message: ""
      reason: ""
      status: "True"
      type: InstanceCreation
    - lastProbeTime: ""
      lastTransitionTime: "2024-01-19T08:27:11Z"
      message: ""
      reason: ""
      status: "True"
      type: FilesystemReadOnly
```

```
$ k get lhv -n longhorn-system -oyaml | grep "remountRequestedAt:"
    remountRequestedAt: ""
```

controller/volume_controller.go
```diff
@@ -1456,7 +1456,23 @@ func (c *VolumeController) requestRemountIfFileSystemReadOnly(v *longhorn.Volume
 	if v.Status.State == longhorn.VolumeStateAttached && e.Status.CurrentState == longhorn.InstanceStateRunning {
 		fileSystemReadOnlyCondition := types.GetCondition(e.Status.Conditions, imtypes.EngineConditionFilesystemReadOnly)
 
-		if fileSystemReadOnlyCondition.Status == longhorn.ConditionStatusTrue {
+		isPVMountOptionReadOnly := false
+		kubeStatus := v.Status.KubernetesStatus
+		if kubeStatus.PVName != "" {
+			pv, err := c.ds.GetPersistentVolume(kubeStatus.PVName)
```
innobead, line 1462: Use `GetPersistentVolumeRO`

innobead (changes requested):

controller/volume_controller.go
ChanYiLin: Fixed, thanks!

innobead (dismissed):

controller/volume_controller.go
```diff
@@ -1456,7 +1456,23 @@ func (c *VolumeController) requestRemountIfFileSystemReadOnly(v *longhorn.Volume
 	if v.Status.State == longhorn.VolumeStateAttached && e.Status.CurrentState == longhorn.InstanceStateRunning {
 		fileSystemReadOnlyCondition := types.GetCondition(e.Status.Conditions, imtypes.EngineConditionFilesystemReadOnly)
 
-		if fileSystemReadOnlyCondition.Status == longhorn.ConditionStatusTrue {
+		isPVMountOptionReadOnly := false
+		kubeStatus := v.Status.KubernetesStatus
+		if kubeStatus.PVName != "" {
+			pv, err := c.ds.GetPersistentVolumeRO(kubeStatus.PVName)
+			if err != nil {
+				log.Warnf("Failed to get PV when checking the mount option of the volume")
```
derekbit, line 1464:
```suggestion
				log.WithError(err).Warn("Failed to get PV when checking the mount option of the volume")
```

derekbit: If err is `NotFound`, we can skip the check. WDYT?

ChanYiLin: Yes the current code return no matter what err it is

ChanYiLin: Fixed, thanks

controller/volume_controller.go
```diff
@@ -1456,7 +1456,26 @@ func (c *VolumeController) requestRemountIfFileSystemReadOnly(v *longhorn.Volume
...
+		if kubeStatus.PVName != "" {
+			pv, err := c.ds.GetPersistentVolumeRO(kubeStatus.PVName)
+			if err != nil {
+				if apierrors.IsNotFound(err) {
+					return
+				}
+				log.WithError(err).Warnf("Failed to get PV when checking the mount option of the volume")
+				return
+			}
+			for _, opt := range pv.Spec.MountOptions {
```
innobead, line 1470: Check `pv != nil`

innobead (dismissed):

derekbit (approved):
LGTM

controller/volume_controller.go
ChanYiLin: Fixed, thanks!

innobead (approved):

innobead:

derekbit:
@mergify backport v1.4.x

PR #2506, opened 2024-01-22
fix(rwx): revert default NFS version to 4.1 for RWX volume mounts 

james-munson:

Issue https://github.com/longhorn/longhorn/issues/7741

This reverts the change in default RWX mount options to use NFS v4.2, and sets it back to v4.1.  The default can still be overridden using parameter nfsOptions in the volume's storage class, and that will remain the preferred way to use v4.2.

Passed `rwx` regression tests: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/6055/

This will need to be backported to 1.4.5, 1.5.4, and 1.60, since the original change was.

### DIFF
--- a/csi/node_server.go
+++ b/csi/node_server.go
@@ -251,7 +251,7 @@ func (ns *NodeServer) nodeStageSharedVolume(volumeID, shareEndpoint, targetPath 
 	export := fmt.Sprintf("%s:%s", server, exportPath)
 
 	defaultMountOptions := []string{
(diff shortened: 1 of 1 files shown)


derekbit (approved):
LGTM

derekbit:
@mergify backport v1.4.x

derekbit:
@mergify backport v1.5.x

derekbit:
@mergify backport v1.6.x

PR #2510, opened 2024-01-23
feat(mount): process manager periodically check mount status and update engine status

ChanYiLin:

1. periodically update engine status based on instance status
2. request remount when volume is ro
3. check PV mount options to see it user intentionally mount with ro
4. reset engine condition when engine is not running


ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/7500

innobead (approved):

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/7729

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/7500

PR #2517, opened 2024-01-24
fix(volume): request remount every 1 min when read only

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/7751

Pod controller reconcile pod every 30 secs.
We should not update request remount time too frequently or it might not delete the pod.
Because it checks if it is 5 secs after request remount time before deleting the pod.

### DIFF
--- a/controller/kubernetes_pod_controller.go
+++ b/controller/kubernetes_pod_controller.go
@@ -29,6 +29,8 @@ import (
 
 const (
 	controllerAgentName = "longhorn-kubernetes-pod-controller"
+
+	remountRequestDelayDuration = 5 * time.Second
 )
 
 type KubernetesPodController struct {
@@ -336,9 +338,13 @@ func (kc *KubernetesPodController) handlePodDeletionIfVolumeRequestRemount(pod *
 		}
 
 		timeNow := time.Now()
-		delayDuration := time.Duration(int64(5)) * time.Second
+		if podStartTime.Before(remountRequestedAt) {
+			if !timeNow.After(remountRequestedAt.Add(remountRequestDelayDuration)) {
+				kc.logger.Infof("Current time is not %v seconds after request remount, requeue the pod %v to handle it later", remountRequestDelayDuration.Seconds(), pod.GetName())
+				kc.enqueuePod(pod)
+				return nil
+			}
 
-		if podStartTime.Before(remountRequestedAt) && timeNow.After(remountRequestedAt.Add(delayDuration)) {
 			gracePeriod := int64(30)
 			err := kc.kubeClient.CoreV1().Pods(pod.Namespace).Delete(context.TODO(), pod.GetName(), metav1.DeleteOptions{
 				GracePeriodSeconds: &gracePeriod,
@@ -468,4 +474,14 @@ func (kc *KubernetesPodController) getAssociatedVolumes(pod *corev1.Pod) ([]*lon
 	}
 
 	return volumeList, nil
+}
+
+func (kc *KubernetesPodController) enqueuePod(obj interface{}) {
+	key, err := controller.KeyFunc(obj)
+	if err != nil {
+		utilruntime.HandleError(fmt.Errorf("couldn't get key for object %#v: %v", obj, err))
+		return
+	}
+
+	kc.queue.Add(key)
 }


controller/volume_controller.go
```diff
@@ -1477,8 +1477,19 @@ func (c *VolumeController) requestRemountIfFileSystemReadOnly(v *longhorn.Volume
...
 
-		if fileSystemReadOnlyCondition.Status == longhorn.ConditionStatusTrue && !isPVMountOptionReadOnly {
-			v.Status.RemountRequestedAt = c.nowHandler()
+		timeNow := time.Now()
+		delayDuration := time.Duration(int64(60)) * time.Second
+		remountRequestedAt, err := time.Parse(time.RFC3339, v.Status.RemountRequestedAt)
+		if err != nil {
+			log.WithError(err).Warnf("Failed to parse remount request time when handling read only condition")
+			return
+		}
```
shuo-wu, line 1486: What if `v.Status.RemountRequestedAt` is empty?

derekbit:
> Pod controller reconcile pod every 30 secs. We should not update request remount time too frequently or it might not delete the pod. Because it checks if it is 5 secs after request remount time before deleting the pod.

Changing the period is fine, but, IMO, it is pretty ad-hoc. We can improve it in the future.

ChanYiLin:
Agree with the ideas,
I changed the fix to re-queue the pod if podStartTime.Before(remountRequestedAt) but !timeNow.After(remountRequestedAt.Add(delayDuration))

derekbit (dismissed):
LGTM

controller/kubernetes_pod_controller.go
```diff
@@ -29,6 +29,8 @@ import (
 
 const (
 	controllerAgentName = "longhorn-kubernetes-pod-controller"
+
+	remountDelayDuration = 5
```
innobead, line 33:
```go
remountDelayDuration = 5 * time.Second
```

controller/kubernetes_pod_controller.go
```diff
@@ -336,9 +338,14 @@ func (kc *KubernetesPodController) handlePodDeletionIfVolumeRequestRemount(pod *
 		}
 
 		timeNow := time.Now()
-		delayDuration := time.Duration(int64(5)) * time.Second
+		delayDuration := time.Duration(int64(remountDelayDuration)) * time.Second
+		if podStartTime.Before(remountRequestedAt) {
+			if !timeNow.After(remountRequestedAt.Add(delayDuration)) {
+				kc.logger.Infof("Current time is not %v secs after request remount, requeue the pod %v to handle it later", remountDelayDuration, pod.GetName())
```
innobead, line 344: `secs` -> `seconds`

innobead (changes requested):
LGTM, just a few feedback.

controller/kubernetes_pod_controller.go
ChanYiLin: The time will be handled here
`delayDuration := time.Duration(int64(remountDelayDuration)) * time.Second`

time.Duration accepts a int64

innobead:

shuo-wu:
> Agree with the ideas,
I changed the fix to re-queue the pod if podStartTime.Before(remountRequestedAt) but !timeNow.After(remountRequestedAt.Add(delayDuration))

Could you create a ticket to track it?

ChanYiLin:
> > Agree with the ideas,
> > I changed the fix to re-queue the pod if podStartTime.Before(remountRequestedAt) but !timeNow.After(remountRequestedAt.Add(delayDuration))
> 
> Could you create a ticket to track it?

Hi @shuo-wu The current PR code is already using re-queue mechanism you provided

controller/kubernetes_pod_controller.go
```diff
@@ -29,6 +29,8 @@ import (
 
 const (
 	controllerAgentName = "longhorn-kubernetes-pod-controller"
+
+	delayDuration = 5 * time.Second
```
innobead, line 33: nit: `remountRequestDelayDuration` 

Suggest naming should be clear.

innobead (changes requested):

shuo-wu:
Oh OK. You mean requeuing the pod object is the final improvement rather than an ad-hoc fix?

controller/kubernetes_pod_controller.go
ChanYiLin: Fixed

ChanYiLin:
> Changing the period is fine, but, IMO, it is pretty ad-hoc. We can improve it in the future.

Yes, according to @derekbit , changing the request remount update time to every 1 mins is ad-hoc fix.
So I abandoned that approach
> Changing the period is fine, but, IMO, it is pretty ad-hoc. We can improve it in the future.

innobead (approved):

PR #2523, opened 2024-01-25
Remove unused process manager connection in longhorn-manager

derekbit:

Issue Longhorn/longhorn#7783

The longhorn-manager communicates with the process manager gRPC service in an older instance-manager (apiVersion < 4). For a newer instance-manager, however, the longhorn-manager only needs to interact with the instance gRPC service. As a result, there is no requirement for the longhorn-manager to establish a connection to the process manager.

### DIFF
--- a/engineapi/instance_manager.go
+++ b/engineapi/instance_manager.go
@@ -138,26 +138,30 @@ func NewInstanceManagerClient(im *longhorn.InstanceManager) (*InstanceManagerCli
 
 	// Create a new process manager client
 	// HACK: TODO: fix me
+	var err error
+	var processManagerClient *imclient.ProcessManagerClient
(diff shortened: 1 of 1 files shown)


innobead (approved):

innobead:
@mergify backport v1.6.x

derekbit:
@mergify backport v1.5.x

ejweber:
I'm a bit concerned about the error log in this change. While I was testing mTLS yesterday, I noticed a flood of error logs (originating from somewhere inside the gRPC stack) while falling back to non-TLS in an expected scenario (longhorn-manager updated with TLS secret but old instance-manager still running). We support this configuration (at least temporarily), but we couldn't get anything else useful from the logs while we were in it. I'll evaluate as I test further today.

innobead:
The log type has been changed at https://github.com/longhorn/longhorn-manager/pull/2527 by @derekbit

PR #2525, opened 2024-01-25
csi: use ensureFolderPath rather than ensureMount for checking folder path

derekbit:

Issue longhorn/longhorn#7784

### DIFF
--- a/csi/util.go
+++ b/csi/util.go
@@ -321,6 +321,29 @@ func ensureMountPoint(path string, mounter mount.Interface) (bool, error) {
 	return isMnt, err
 }
 
+// ensureDirectory checks if a folder exists at the specified path.
+// If not, it creates the folder and returns true, otherwise returns false.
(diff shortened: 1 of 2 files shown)


csi/util.go
```diff
@@ -321,6 +321,29 @@ func ensureMountPoint(path string, mounter mount.Interface) (bool, error) {
 	return isMnt, err
 }
 
+// ensureFolderPath checks if a folder exists at the specified path.
+// If not, it creates the folder and returns true, otherwise returns false.
+// If the path exists but is not a folder, it returns an error.
+func ensureFolderPath(path string) (bool, error) {
```
c3y1huang, line 327: WDYT?
```suggestion
func ensureDirectory(path string) (bool, error) {
```

derekbit: Done! Thank you.

innobead (approved):

innobead:
@mergify backport v1.6.x

PR #2527, opened 2024-01-25
engineapi: proxy client supports TLS

derekbit:

Issue https://github.com/longhorn/longhorn/issues/7040

Deps on https://github.com/longhorn/longhorn-instance-manager/pull/382

### DIFF
--- a/engineapi/proxy.go
+++ b/engineapi/proxy.go
@@ -1,6 +1,8 @@
 package engineapi
 
 import (
+	"path/filepath"
+
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"golang.org/x/net/context"
@@ -9,6 +11,7 @@ import (
 	imutil "github.com/longhorn/longhorn-instance-manager/pkg/util"
 
 	"github.com/longhorn/longhorn-manager/datastore"
+	"github.com/longhorn/longhorn-manager/types"
 	"github.com/longhorn/longhorn-manager/util"
 
 	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
@@ -81,9 +84,19 @@ func NewEngineClientProxy(im *longhorn.InstanceManager, logger logrus.FieldLogge
 	}
 
 	ctx, cancel := context.WithCancel(context.Background())
-	client, err := imclient.NewProxyClient(ctx, cancel, im.Status.IP, InstanceManagerProxyServiceDefaultPort)
+	client, err := imclient.NewProxyClientWithTLS(ctx, cancel, im.Status.IP, InstanceManagerProxyServiceDefaultPort,
+		filepath.Join(types.TLSDirectoryInContainer, types.TLSCAFile),
+		filepath.Join(types.TLSDirectoryInContainer, types.TLSCertFile),
+		filepath.Join(types.TLSDirectoryInContainer, types.TLSKeyFile),
+		"longhorn-backend.longhorn-system")
 	if err != nil {
-		return nil, err
+		logrus.WithError(err).Tracef("Falling back to non-tls client for Proxy Service Client for %v IP %v",
+			im.Name, im.Status.IP)
+		client, err = imclient.NewProxyClient(ctx, cancel, im.Status.IP, InstanceManagerProxyServiceDefaultPort, nil)
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to initialize Proxy Service Client for %v IP %v",
+				im.Name, im.Status.IP)
+		}
 	}
 
 	proxyConnCounter.IncreaseCount()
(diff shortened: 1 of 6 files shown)


derekbit:
Updated proxy client in engineapi while testing mTLS.

cc @sfackler

engineapi/proxy.go
```diff
@@ -81,9 +84,19 @@ func NewEngineClientProxy(im *longhorn.InstanceManager, logger logrus.FieldLogge
 	}
 
 	ctx, cancel := context.WithCancel(context.Background())
-	client, err := imclient.NewProxyClient(ctx, cancel, im.Status.IP, InstanceManagerProxyServiceDefaultPort)
+	client, err := imclient.NewProxyClientWithTLS(ctx, cancel, im.Status.IP, InstanceManagerProxyServiceDefaultPort,
+		filepath.Join(types.TLSDirectoryInContainer, types.TLSCAFile),
+		filepath.Join(types.TLSDirectoryInContainer, types.TLSCertFile),
+		filepath.Join(types.TLSDirectoryInContainer, types.TLSKeyFile),
+		"longhorn-backend.longhorn-system")
```
c3y1huang, line 91: Why `longhorn-backend.longhorn-system`?

engineapi/proxy.go
```diff
@@ -81,9 +84,19 @@ func NewEngineClientProxy(im *longhorn.InstanceManager, logger logrus.FieldLogge
...
 	if err != nil {
-		return nil, err
+		logrus.WithError(err).Warnf("Falling back to non-tls client for Instance Manager Instance Service Client for %v, state: %v, IP: %v",
```
c3y1huang, line 93: Probably not a good idea to log as warnings, suggest use at the trace level.

c3y1huang: Any reason we are logging the current state?

engineapi/proxy.go
```diff
@@ -81,9 +84,19 @@ func NewEngineClientProxy(im *longhorn.InstanceManager, logger logrus.FieldLogge
...
+			im.Name, im.Status.CurrentState, im.Status.IP)
+		client, err = imclient.NewProxyClient(ctx, cancel, im.Status.IP, InstanceManagerProxyServiceDefaultPort, nil)
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to initialize Instance Manager Instance Service Client for %v, state: %v, IP: %v, TLS: %v",
```
c3y1huang, line 97: Proxy service client

c3y1huang: Is it necessary to have the state, IP, and TLS in the error? They've already been logged above. WDYT?

engineapi/proxy.go
derekbit: [Server uses `longhorn-backend.longhorn-system` as the peer name.](https://github.com/longhorn/longhorn-instance-manager/blob/master/app/cmd/start.go#L195)

engineapi/proxy.go
derekbit: Removed. Thanks.

engineapi/proxy.go
derekbit: Change the above log level to Trace, so I think recording the IP is still necessary.

innobead (approved):
LGTM. Only merge this until the the dependency gets updated.

ejweber (changes requested):
This PR does not work as intended because it is missing the call to `CheckConnection` like @sfackler's https://github.com/sfackler/longhorn-manager/blob/248b0c622495b35f2b22e765fd78dd4fabb9ffeb/engineapi/proxy.go#L101-L103. As a result, we do not recognize the TLS connection has not been initialized and cannot fall back if TLS is not supported on the instance-manager side.

When we have a longhorn-manager using TLS and an instance-manager not using TLS:

```
[longhorn-manager-spntg] W0125 17:09:11.336376       1 logging.go:59] [core] [Channel #59 SubChannel #60] grpc: addrConn.createTransport failed to connect to {Addr: "10.42.1.29:8501", ServerName: "longhorn-backend.longhorn-system", }. Err: connection error: desc = "transport: authentication handshake failed: tls: first record does not look like a TLS handshake"
[longhorn-manager-spntg] E0125 17:09:11.336579       1 engine_controller.go:838] failed to update status for engine test-after-im-switch-e-0: proxyServer=10.42.1.29:8501 destination=10.42.1.29:10010: failed to list replicas for volume: rpc error: code = Unavailable desc = connection error: desc = "transport: authentication handshake failed: tls: first record does not look like a TLS handshake"
```

We should have recognized the broken TLS handshake when creating the client, but we did not.

It is somewhat surprising the `Dial` fails to return a TLS related error, but by default `Dial` does not block for a connection to be established, so we don't know until we first try to use the connection that there is a TLS related problem. This is why @sfackler had to implement the connection check and make use of it in his longhorn-manager code. I think there are some potential improvements we can make (e.g. use a blocking Dial?) in the future, but for `v1.6.0`, we can focus on just getting back to functional.

I'm going to put together a PR that combines @sfackler's method and the logging changes from this one since it is currently late for @derekbit.

derekbit:
Use https://github.com/longhorn/longhorn-manager/pull/2533#issuecomment-1911173568 instead

PR #2535, opened 2024-01-26
Fix `rpc error: code = Canceled desc = context canceled` when falling back to non-TLS client

derekbit:

Issue longhorn/longhorn#7040

If TLS connection is broken, cancel() of ctx will be called. Then, it causes the fallback to non-TLS runs into the error `rpc error: code = Canceled desc = context canceled`.

### DIFF
--- a/engineapi/instance_manager.go
+++ b/engineapi/instance_manager.go
@@ -92,7 +92,7 @@ func CheckInstanceManagerProxySupport(im *longhorn.InstanceManager) error {
 }
 
 // NewInstanceManagerClient creates a new instance manager client
-func NewInstanceManagerClient(ctx context.Context, ctxCancel context.CancelFunc, im *longhorn.InstanceManager) (*InstanceManagerClient, error) {
+func NewInstanceManagerClient(im *longhorn.InstanceManager) (*InstanceManagerClient, error) {
 	// Do not check the major version here. Since IM cannot get the major version without using this client to call VersionGet().
 	if im.Status.CurrentState != longhorn.InstanceManagerStateRunning || im.Status.IP == "" {
 		return nil, fmt.Errorf("invalid Instance Manager %v, state: %v, IP: %v", im.Name, im.Status.CurrentState, im.Status.IP)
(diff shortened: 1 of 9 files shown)


engineapi/instance_manager.go
```diff
@@ -189,7 +197,7 @@ func NewInstanceManagerClient(ctx context.Context, ctxCancel context.CancelFunc,
 
 			version, err := processManagerClient.VersionGet()
 			if err != nil {
-				return nil, errors.Wrapf(err, "failed to check version of Instance Manager Process Manager Service Client without TLS for %v IP %v",
+				return nil, errors.Wrapf(err, "failed to check version of Instance Manager Process Manager Service Client without for %v IP %v",
```
c3y1huang, line 200:
```suggestion
				return nil, errors.Wrapf(err, "failed to check version of Instance Manager Process Manager Service Client for %v IP %v",
```

engineapi/instance_manager.go
```diff
@@ -228,7 +238,7 @@ func NewInstanceManagerClient(ctx context.Context, ctxCancel context.CancelFunc,
 
 		version, err := instanceServiceClient.VersionGet()
 		if err != nil {
-			return nil, errors.Wrapf(err, "failed to check version of Instance Manager Instance Service Client without TLS for %v IP %v",
+			return nil, errors.Wrapf(err, "failed to check version of Instance Manager Instance Service Client without for %v IP %v",
```
c3y1huang, line 241:
```suggestion
			return nil, errors.Wrapf(err, "failed to check version of Instance Manager Instance Service Client for %v IP %v",
```

c3y1huang (changes requested):

engineapi/instance_manager.go
derekbit: Fixed. Thank you.

engineapi/instance_manager.go
derekbit: Fixed. Thank you.

c3y1huang (approved):
LGTM

derekbit:
@mergify backport v1.6.x

engineapi/instance_manager.go
```diff
@@ -104,11 +104,14 @@ func NewInstanceManagerClient(ctx context.Context, ctxCancel context.CancelFunc,
 		defer func() {
 			if err != nil && processManagerClient != nil {
 				_ = processManagerClient.Close()
+				processManagerClient = nil
 			}
```
innobead, line 108: Why not call close?

innobead: Misread. U just prevent it was used after it is closed with an error caught.

ejweber (commented):
LGTM.

PR #2539, opened 2024-01-29
fix(setting): check if the settings are synchronized first.

mantissahz:

Issue longhorn/longhorn#7797

Check if the settings are synchronized first instead of checking
if volumes are attached first to eliminate unnecessary error logs.

### DIFF
--- a/controller/setting_controller.go
+++ b/controller/setting_controller.go
@@ -277,15 +277,6 @@ func (sc *SettingController) syncDangerZoneSettingsForManagedComponents(settingN
 	}
 
 	if slices.Contains(dangerSettingsRequiringAllVolumesDetached, settingName) {
-		detached, _, err := sc.ds.AreAllVolumesDetached(longhorn.DataEngineTypeAll)
-		if err != nil {
-			return errors.Wrapf(err, "failed to check volume detachment for %v setting update", settingName)
-		}
-
-		if !detached {
-			return &types.ErrorInvalidState{Reason: fmt.Sprintf("failed to apply %v setting to Longhorn components when there are attached volumes. It will be eventually applied", settingName)}
-		}
-
 		switch settingName {
 		case types.SettingNameTaintToleration:
 			if err := sc.updateTaintToleration(); err != nil {
@@ -328,15 +319,6 @@ func (sc *SettingController) syncDangerZoneSettingsForManagedComponents(settingN
 			dataEngine := longhorn.DataEngineTypeV1
 			if settingName == types.SettingNameV2DataEngineGuaranteedInstanceManagerCPU {
 				dataEngine = longhorn.DataEngineTypeV2
-			}
-
-			detached, _, err := sc.ds.AreAllVolumesDetached(dataEngine)
-			if err != nil {
-				return errors.Wrapf(err, "failed to check volume detachment for %v setting update", settingName)
-			}
-
-			if !detached {
-				return &types.ErrorInvalidState{Reason: fmt.Sprintf("failed to apply %v setting to Longhorn components when there are attached volumes. It will be eventually applied", settingName)}
 			}
 
 			if err := sc.updateInstanceManagerCPURequest(dataEngine); err != nil {
@@ -576,74 +558,110 @@ func (sc *SettingController) updateTaintToleration() error {
 	}
 	newTolerationsMap := util.TolerationListToMap(newTolerationsList)
 
-	daemonsetList, err := sc.ds.ListDaemonSetWithLabels(types.GetBaseLabelsForSystemManagedComponent())
+	updatingRuntimeObjects, err := sc.collectRuntimeObjects()
 	if err != nil {
-		return errors.Wrap(err, "failed to list Longhorn daemonsets for toleration update")
+		return errors.Wrap(err, "failed to collect runtime objects for toleration update")
+	}
+	notUpdatedTolerationObjs, err := getNotUpdatedTolerationList(newTolerationsMap, updatingRuntimeObjects...)
+	if err != nil {
+		return err
+	}
+	if len(notUpdatedTolerationObjs) == 0 {
+		return nil
 	}
 
-	deploymentList, err := sc.ds.ListDeploymentWithLabels(types.GetBaseLabelsForSystemManagedComponent())
+	detached, _, err := sc.ds.AreAllVolumesDetached(longhorn.DataEngineTypeAll)
 	if err != nil {
-		return errors.Wrap(err, "failed to list Longhorn deployments for toleration update")
+		return errors.Wrapf(err, "failed to check volume detachment for %v setting update", types.SettingNameTaintToleration)
+	}
+	if !detached {
+		return &types.ErrorInvalidState{Reason: fmt.Sprintf("failed to apply %v setting to Longhorn components when there are attached volumes. It will be eventually applied", types.SettingNameTaintToleration)}
 	}
 
-	imPodList, err := sc.ds.ListInstanceManagerPods()
-	if err != nil {
-		return errors.Wrap(err, "failed to list instance manager pods for toleration update")
-	}
-
-	smPodList, err := sc.ds.ListShareManagerPods()
-	if err != nil {
-		return errors.Wrap(err, "failed to list share manager pods for toleration update")
-	}
-
-	bimPodList, err := sc.ds.ListBackingImageManagerPods()
-	if err != nil {
-		return errors.Wrap(err, "failed to list backing image manager pods for toleration update")
-	}
-
-	for _, dp := range deploymentList {
-		lastAppliedTolerationsList, err := getLastAppliedTolerationsList(dp)
+	for _, obj := range notUpdatedTolerationObjs {
+		lastAppliedTolerationsList, err := getLastAppliedTolerationsList(obj)
 		if err != nil {
 			return err
 		}
-		if reflect.DeepEqual(util.TolerationListToMap(lastAppliedTolerationsList), newTolerationsMap) {
-			continue
-		}
-		if err := sc.updateTolerationForDeployment(dp, lastAppliedTolerationsList, newTolerationsList); err != nil {
-			return err
-		}
-	}
-
-	for _, ds := range daemonsetList {
-		lastAppliedTolerationsList, err := getLastAppliedTolerationsList(ds)
-		if err != nil {
-			return err
-		}
-		if reflect.DeepEqual(util.TolerationListToMap(lastAppliedTolerationsList), newTolerationsMap) {
-			continue
-		}
-		if err := sc.updateTolerationForDaemonset(ds, lastAppliedTolerationsList, newTolerationsList); err != nil {
-			return err
-		}
-	}
-
-	pods := append(imPodList, smPodList...)
-	pods = append(pods, bimPodList...)
-	for _, pod := range pods {
-		lastAppliedTolerations, err := getLastAppliedTolerationsList(pod)
-		if err != nil {
-			return err
-		}
-		if reflect.DeepEqual(util.TolerationListToMap(lastAppliedTolerations), newTolerationsMap) {
-			continue
-		}
-		sc.logger.Infof("Deleting pod %v to update tolerations from %v to %v", pod.Name, util.TolerationListToMap(lastAppliedTolerations), newTolerationsMap)
-		if err := sc.ds.DeletePod(pod.Name); err != nil {
-			return err
+		switch objType := obj.(type) {
+		case *appsv1.DaemonSet:
+			ds := obj.(*appsv1.DaemonSet)
+			sc.logger.Infof("Deleting daemonset %v to update tolerations from %v to %v", ds.Name, util.TolerationListToMap(lastAppliedTolerationsList), newTolerationsMap)
+			if err := sc.updateTolerationForDaemonset(ds, lastAppliedTolerationsList, newTolerationsList); err != nil {
+				return err
+			}
+		case *appsv1.Deployment:
+			dp := obj.(*appsv1.Deployment)
+			sc.logger.Infof("Updating deployment %v to update tolerations from %v to %v", dp.Name, util.TolerationListToMap(lastAppliedTolerationsList), newTolerationsMap)
+			if err := sc.updateTolerationForDeployment(dp, lastAppliedTolerationsList, newTolerationsList); err != nil {
+				return err
+			}
+		case *corev1.Pod:
+			pod := obj.(*corev1.Pod)
+			sc.logger.Infof("Deleting pod %v to update tolerations from %v to %v", pod.Name, util.TolerationListToMap(lastAppliedTolerationsList), newTolerationsMap)
+			if err := sc.ds.DeletePod(pod.Name); err != nil {
+				return err
+			}
+		default:
+			return fmt.Errorf("unknown object type %v when updating %v setting", objType, types.SettingNameTaintToleration)
 		}
 	}
 
 	return nil
+}
+
+func (sc *SettingController) collectRuntimeObjects() (returnCollectRuntimeObjects []runtime.Object, err error) {
+	dsList, err := sc.ds.ListDeploymentWithLabels(types.GetBaseLabelsForSystemManagedComponent())
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to list Longhorn deployments for collecting runtime objects")
+	}
+
+	dpList, err := sc.ds.ListDaemonSetWithLabels(types.GetBaseLabelsForSystemManagedComponent())
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to list Longhorn daemonsets for collecting runtime objects")
+	}
+
+	imPodList, err := sc.ds.ListInstanceManagerPods()
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to list instance manager pods for collecting runtime objects")
+	}
+	smPodList, err := sc.ds.ListShareManagerPods()
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to list share manager pods for collecting runtime objects")
+	}
+	bimPodList, err := sc.ds.ListBackingImageManagerPods()
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to list backing image manager pods for collecting runtime objects")
+	}
+	podList := append(imPodList, smPodList...)
+	podList = append(podList, bimPodList...)
+
+	for _, ds := range dsList {
+		returnCollectRuntimeObjects = append(returnCollectRuntimeObjects, runtime.Object(ds))
+	}
+	for _, dp := range dpList {
+		returnCollectRuntimeObjects = append(returnCollectRuntimeObjects, runtime.Object(dp))
+	}
+	for _, pod := range podList {
+		returnCollectRuntimeObjects = append(returnCollectRuntimeObjects, runtime.Object(pod))
+	}
+
+	return returnCollectRuntimeObjects, nil
+}
+
+func getNotUpdatedTolerationList(newTolerationsMap map[string]corev1.Toleration, objs ...runtime.Object) ([]runtime.Object, error) {
+	notUpdatedObjsList := []runtime.Object{}
+	for _, obj := range objs {
+		lastAppliedTolerationsList, err := getLastAppliedTolerationsList(obj)
+		if err != nil {
+			return notUpdatedObjsList, err
+		}
+		if !reflect.DeepEqual(util.TolerationListToMap(lastAppliedTolerationsList), newTolerationsMap) {
+			notUpdatedObjsList = append(notUpdatedObjsList, obj)
+		}
+	}
+
+	return notUpdatedObjsList, nil
 }
 
 func (sc *SettingController) updateTolerationForDeployment(dp *appsv1.Deployment, lastAppliedTolerations, newTolerations []corev1.Toleration) error {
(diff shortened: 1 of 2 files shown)


innobead:
@PhanLe1010 @derekbit Please help review this.

controller/setting_controller.go
```diff
@@ -606,9 +588,19 @@ func (sc *SettingController) updateTaintToleration() error {
...
 		}
+
+		detached, _, err := sc.ds.AreAllVolumesDetached(longhorn.DataEngineTypeAll)
+		if err != nil {
+			return errors.Wrapf(err, "failed to check volume detachment for %v setting update", types.SettingNameTaintToleration)
+		}
+		if !detached {
+			return &types.ErrorInvalidState{Reason: fmt.Sprintf("failed to apply %v setting to Longhorn components when there are attached volumes. It will be eventually applied", types.SettingNameTaintToleration)}
+		}
+
```
derekbit, line 603: Can we relocate the check just before the for loop rather than inside the loop?

controller/setting_controller.go
```diff
@@ -619,9 +611,19 @@ func (sc *SettingController) updateTaintToleration() error {
...
```
derekbit, line 626: If https://github.com/longhorn/longhorn-manager/pull/2539/files#r1469726286 works, the check can be removed.
.

controller/setting_controller.go
```diff
@@ -634,9 +636,19 @@ func (sc *SettingController) updateTaintToleration() error {
...
 			continue
 		}
+
+		detached, _, err := sc.ds.AreAllVolumesDetached(longhorn.DataEngineTypeAll)
+		if err != nil {
+			return errors.Wrapf(err, "failed to check volume detachment for %v setting update", types.SettingNameTaintToleration)
+		}
+		if !detached {
+			return &types.ErrorInvalidState{Reason: fmt.Sprintf("failed to apply %v setting to Longhorn components when there are attached volumes. It will be eventually applied", types.SettingNameTaintToleration)}
+		}
```
derekbit, line 650: Same as above

controller/setting_controller.go
```diff
@@ -740,6 +752,15 @@ func (sc *SettingController) updatePriorityClass() error {
...
 		}
+
+		detached, _, err := sc.ds.AreAllVolumesDetached(longhorn.DataEngineTypeAll)
+		if err != nil {
+			return errors.Wrapf(err, "failed to check volume detachment for %v setting update", types.SettingNamePriorityClass)
+		}
+		if !detached {
+			return &types.ErrorInvalidState{Reason: fmt.Sprintf("failed to apply %v setting to Longhorn components when there are attached volumes. It will be eventually applied", types.SettingNamePriorityClass)}
+		}
+
```
derekbit, line 763: same question as https://github.com/longhorn/longhorn-manager/pull/2539/files#r1469726286

controller/setting_controller.go
```diff
@@ -855,6 +894,14 @@ func (sc *SettingController) updateCNI() error {
...
 		}
 
+		detached, _, err := sc.ds.AreAllVolumesDetached(longhorn.DataEngineTypeAll)
+		if err != nil {
+			return errors.Wrapf(err, "failed to check volume detachment for %v setting update", types.SettingNameStorageNetwork)
+		}
+		if !detached {
+			return &types.ErrorInvalidState{Reason: fmt.Sprintf("failed to apply %v setting to Longhorn components when there are attached volumes. It will be eventually applied", types.SettingNameStorageNetwork)}
+		}
+
```
derekbit, line 904: Same question as https://github.com/longhorn/longhorn-manager/pull/2539/files#r1469726286

controller/setting_controller.go
```diff
@@ -991,15 +1039,26 @@ func (sc *SettingController) updateNodeSelector() error {
...
 		}
+
+		detached, _, err := sc.ds.AreAllVolumesDetached(longhorn.DataEngineTypeAll)
+		if err != nil {
+			return errors.Wrapf(err, "failed to check volume detachment for %v setting update", types.SettingNameSystemManagedComponentsNodeSelector)
+		}
+		if !detached {
+			return &types.ErrorInvalidState{Reason: fmt.Sprintf("failed to apply %v setting to Longhorn components when there are attached volumes. It will be eventually applied", types.SettingNameSystemManagedComponentsNodeSelector)}
+		}
+
```
derekbit, line 1061: same question as https://github.com/longhorn/longhorn-manager/pull/2539/files#r1469726286

controller/setting_controller.go
```diff
@@ -1012,9 +1071,19 @@ func (sc *SettingController) updateNodeSelector() error {
...
```
derekbit, line 1086: Same question as https://github.com/longhorn/longhorn-manager/pull/2539/files#r1469729061

controller/setting_controller.go
```diff
@@ -1029,9 +1098,19 @@ func (sc *SettingController) updateNodeSelector() error {
...
 			continue
 		}
+
+		detached, _, err := sc.ds.AreAllVolumesDetached(longhorn.DataEngineTypeAll)
+		if err != nil {
+			return errors.Wrapf(err, "failed to check volume detachment for %v setting update", types.SettingNameSystemManagedComponentsNodeSelector)
+		}
+		if !detached {
+			return &types.ErrorInvalidState{Reason: fmt.Sprintf("failed to apply %v setting to Longhorn components when there are attached volumes. It will be eventually applied", types.SettingNameSystemManagedComponentsNodeSelector)}
+		}
```
derekbit, line 1112: Same question as https://github.com/longhorn/longhorn-manager/pull/2539/files#r1469729061

controller/setting_controller.go
PhanLe1010: Agree that we should check it once before the loop. WDYT about something like this:
```go
func hasUpdatedToleration(newTolerationsMap map[string]corev1.Toleration, objs ...runtime.Object) (bool, error) {
	for _, obj := range objs {
		lastAppliedTolerationsList, err := getLastAppliedTolerationsList(obj)
		if err != nil {
			return false, err
		}
		if !reflect.DeepEqual(util.TolerationListToMap(lastAppliedTolerationsList), newTolerationsMap) {
			return true, nil
		}
	}

	return false, nil
}

// updateTaintToleration deletes all user-deployed and system-managed components immediately with the updated taint toleration.
func (sc *SettingController) updateTaintToleration() error {
	setting, err := sc.ds.GetSettingWithAutoFillingRO(types.SettingNameTaintToleration)
	if err != nil {
		return err
	}
	newTolerations := setting.Value
	newTolerationsList, err := types.UnmarshalTolerations(newTolerations)
	if err != nil {
		return err
	}
	newTolerationsMap := util.TolerationListToMap(newTolerationsList)
	runtimeObjects := []runtime.Object

	daemonsetList, err := sc.ds.ListDaemonSetWithLabels(types.GetBaseLabelsForSystemManagedComponent())
	if err != nil {
		return errors.Wrap(err, "failed to list Longhorn daemonsets for toleration update")
	}
	for _, obj := range daemonsetList {
		runtimeObjects = append(runtimeObjects, runtime.Object(obj))
	}

	deploymentList, err := sc.ds.ListDeploymentWithLabels(types.GetBaseLabelsForSystemManagedComponent())
	if err != nil {
		return errors.Wrap(err, "failed to list Longhorn deployments for toleration update")
	}
	for _, obj := range deploymentList {
		runtimeObjects = append(runtimeObjects, runtime.Object(obj))
	}

	imPodList, err := sc.ds.ListInstanceManagerPods()
	if err != nil {
		return errors.Wrap(err, "failed to list instance manager pods for toleration update")
	}
	for _, obj := range imPodList {
		runtimeObjects = append(runtimeObjects, runtime.Object(obj))
	}

	smPodList, err := sc.ds.ListShareManagerPods()
	if err != nil {
		return errors.Wrap(err, "failed to list share manager pods for toleration update")
	}
	for _, obj := range smPodList {
		runtimeObjects = append(runtimeObjects, runtime.Object(obj))
	}

	bimPodList, err := sc.ds.ListBackingImageManagerPods()
	if err != nil {
		return errors.Wrap(err, "failed to list backing image manager pods for toleration update")
	}
	for _, obj := range bimPodList {
		runtimeObjects = append(runtimeObjects, runtime.Object(obj))
	}


	hasUpdated, err := hasUpdatedToleration(newTolerationsMap, runtimeObjects...)
	if err != nil {
		return err
	}
	if hasUpdated {
		return nil
	}
```

mantissahz: @derekbit 
Moved the whole section out of the loops and relocated the check volume attached section after checking the setting synchronized.

@PhanLe1010
```golang
func hasUpdatedToleration(newTolerationsMap map[string]corev1.Toleration, objs ...runtime.Object) (bool, error) {
	for _, obj := range objs {
		lastAppliedTolerationsList, err := getLastAppliedTolerationsList(obj)
		if err != nil {
			return false, err
		}
		if !reflect.DeepEqual(util.TolerationListToMap(lastAppliedTolerationsList), newTolerationsMap) {
			return true, nil
		}
	}

	return false, nil
}
...
```
There is a scenario: 
1. If we have three setting controllers on three worker nodes,
- controller A starts updating deployment A >B >C
- controller B starts updating deployment B >A >C
- controller C starts updating deployment B >A >C

1. When controller A tries to update B, it will get a conflict error and then be re-queued and reconcile the setting again (the situation as controller B and C.)
1. When controllers A, B, and C start to reconcile the setting again, it will get deployment A, and B updated and return nil as the setting synchronized but deployment C does not apply the setting.

If we check which objects are not synchronized and return an unsynchronized list at the end of  `hasUpdatedToleration`, it should be able to handle the scenario above.

The implementation was updated as you suggested.

controller/setting_controller.go
mantissahz: Modified. Thanks

controller/setting_controller.go
mantissahz: Modified. Thanks

controller/setting_controller.go
mantissahz: Modified. Thanks

controller/setting_controller.go
mantissahz: Modified. Thanks

controller/setting_controller.go
mantissahz: Modified. Thanks

controller/setting_controller.go
mantissahz: Modified. Thanks

controller/setting_controller.go
mantissahz: Modified. Thanks

PhanLe1010 (commented):
Looks like we have not handle the setting `updateDataEngine`. Right now, it will still log even if a particular data engine is disabled even though we already cleanup the corresponding IM and there is no changes in the system. Could you double check. Thank you

mantissahz:
>Right now, it will still log even if a particular data engine is disabled even though we already cleanup the corresponding IM and there is no changes in the system.

@PhanLe1010,

Could you describe the Longhorn version, the steps to reproduce it, and show the logs you found?
Is the log like this one?

```txt
E0127 15:17:07.376434       1 setting_controller.go:208] failed to sync setting for longhorn-system/guaranteed-instance-manager-cpu: current state prevents this: failed to apply guaranteed-instance-manager-cpu setting to Longhorn components when there are attached volumes. It will be eventually applied
```

Because we depend on the function `AreAllVolumesDetached`([here](https://github.com/longhorn/longhorn-manager/blob/master/datastore/longhorn.go#L472-L502)) to check if all volumes are detached, it should not return an error (and log) if there is no instance manager for a particular data engine.

PhanLe1010:
> > Right now, it will still log even if a particular data engine is disabled even though we already cleanup the corresponding IM and there is no changes in the system.
> 
> @PhanLe1010,
> 
> Could you describe the Longhorn version, the steps to reproduce it, and show the logs you found? Is the log like this one?
> 
> ```
> E0127 15:17:07.376434       1 setting_controller.go:208] failed to sync setting for longhorn-system/guaranteed-instance-manager-cpu: current state prevents this: failed to apply guaranteed-instance-manager-cpu setting to Longhorn components when there are attached volumes. It will be eventually applied
> ```
> 
> Because we depend on the function `AreAllVolumesDetached`([here](https://github.com/longhorn/longhorn-manager/blob/master/datastore/longhorn.go#L472-L502)) to check if all volumes are detached, it should not return an error (and log) if there is no instance manager for a particular data engine.


Sorry for the late response. So with this PR, we have:
1. When we sync the setting `SettingNameV2DataEngine`, we enter this line of code https://github.com/mantissahz/longhorn-manager/blob/a9bff9d1a3ddc496808cc13d7989bc6efed839cf/controller/setting_controller.go#L985-L986
1. The above line of code check if the `dataEngineEnabled` is enable or not. Assume that the setting is never changed and it is always is `false`. In this case, we are still printing out the log https://github.com/mantissahz/longhorn-manager/blob/a9bff9d1a3ddc496808cc13d7989bc6efed839cf/datastore/longhorn.go#L419-L420 if there are attached volumes.
1. The end result is that we will print out that log every 30s even though there is nothing change in the system/setting


WDYT about this concern?

mantissahz:
@PhanLe1010,
>The above line of code check if the dataEngineEnabled is enable or not. Assume that the setting is never changed and it is always is false.

1. The setting is always false after a fresh install, and it would have any volumes attached.
   I think it will not print the logs or do you think there will be an error returned from https://github.com/mantissahz/longhorn-manager/blob/a9bff9d1a3ddc496808cc13d7989bc6efed839cf/datastore/longhorn.go#L419-L420?
2. The setting is enabled and some volumes are attached and then disable the setting.
   It should be printing logs.

>In this case, we are still printing out the log https://github.com/mantissahz/longhorn-manager/blob/a9bff9d1a3ddc496808cc13d7989bc6efed839cf/datastore/longhorn.go#L419-L420 if there are attached volumes.

Do you mean we are printing out the log https://github.com/mantissahz/longhorn-manager/blob/a9bff9d1a3ddc496808cc13d7989bc6efed839cf/datastore/longhorn.go#L423-L425 when volumes are attached with `dataEngineEnabled` false?

or Do you mean that when `dataEngineEnabled` is false and https://github.com/mantissahz/longhorn-manager/blob/a9bff9d1a3ddc496808cc13d7989bc6efed839cf/datastore/longhorn.go#L417 will return an error?

1. I think we should have the log when volumes are attached with `dataEngineEnabled` false, right?
4. The log level is `Trace` (ErrorIsInvalidState) like other danger zone settings.

PhanLe1010:
> Do you mean we are printing out the log https://github.com/mantissahz/longhorn-manager/blob/a9bff9d1a3ddc496808cc13d7989bc6efed839cf/datastore/longhorn.go#L423-L425 when volumes are attached with dataEngineEnabled false?

Hi @mantissahz , this is what I meant. We are printing out the logs when there is nothing change in the Longhorn system. Specifically, we are printing out the log when:
1. There are attached volumes and they have not changed. They are still attached
2. Setting dataEngineEnabled is false and it has not changed. It is still dataEngineEnabled.

WDYT?

PhanLe1010:
After discussing with @mantissahz , we think that it is difficult to achieve the behavior at https://github.com/longhorn/longhorn-manager/pull/2539#issuecomment-2043939546. We can keep the current implementation for now

PhanLe1010 (dismissed):
LGTM

controller/setting_controller.go
```diff
@@ -600,50 +589,73 @@ func (sc *SettingController) updateTaintToleration() error {
 	if err != nil {
 		return errors.Wrap(err, "failed to list backing image manager pods for toleration update")
 	}
+	pods := append(imPodList, smPodList...)
+	pods = append(pods, bimPodList...)
+	for _, pod := range pods {
+		updatingRuntimeObjects = append(updatingRuntimeObjects, runtime.Object(pod))
+	}
```
derekbit, line 596: Can we move the updatingRuntimeObjects collection codes to a helper function, for example
```
updatingRuntimeObjects, err := collectRuntimeObjects(...)
```

Same for other functions.

mantissahz: >Can we move the updatingRuntimeObjects collection codes to a helper function

Do you mean that move similar codes to `utils` or `go-common-libs`?

mantissahz: Modified. Thanks

controller/setting_controller.go
```diff
@@ -600,50 +582,71 @@ func (sc *SettingController) updateTaintToleration() error {
...
+				return err
+			}
+		case *corev1.Pod:
+			pod := obj.(*corev1.Pod)
+			sc.logger.Infof("Deleting pod %v to update tolerations from %v to %v", pod.Name, util.TolerationListToMap(lastAppliedTolerationsList), newTolerationsMap)
+			if err := sc.ds.DeletePod(pod.Name); err != nil {
+				return err
+			}
+		default:
+			return fmt.Errorf("Unknown object type %v when updating %v setting", objType, types.SettingNameTaintToleration)
```
derekbit, line 630:
```suggestion
			return fmt.Errorf("unknown object type %v when updating %v setting", objType, types.SettingNameTaintToleration)
```

controller/setting_controller.go
```diff
@@ -600,50 +582,71 @@ func (sc *SettingController) updateTaintToleration() error {
...
-		if err != nil {
-			return err
-		}
-		if reflect.DeepEqual(util.TolerationListToMap(lastAppliedTolerationsList), newTolerationsMap) {
-			continue
-		}
-		if err := sc.updateTolerationForDeployment(dp, lastAppliedTolerationsList, newTolerationsList); err != nil {
-			return err
-		}
+	updatingRuntimeObjects := util.CollectRuntimeObjects(daemonsetList, deploymentList, pods)
```
derekbit, line 588: Why not moving
```
daemonsetList, err := sc.ds.ListDaemonSetWithLabels 
....
bimPodList, err := sc.ds.ListBackingImageManagerPods
```
to util.CollectRuntimeObjects?

controller/setting_controller.go
```diff
@@ -735,41 +738,81 @@ func (sc *SettingController) updatePriorityClass() error {
...
+				return err
+			}
+		case *corev1.Pod:
+			pod := obj.(*corev1.Pod)
+			sc.logger.Infof("Deleting pod %v to update the priority class from %v to %v", pod.Name, pod.Spec.PriorityClassName, newPriorityClass)
+			if err := sc.ds.DeletePod(pod.Name); err != nil {
+				return err
+			}
+		default:
+			return fmt.Errorf("Unknown object type %v when updating %v setting", objType, types.SettingNamePriorityClass)
```
derekbit, line 784:
```suggestion
			return fmt.Errorf("unknown object type %v when updating %v setting", objType, types.SettingNamePriorityClass)
```

controller/setting_controller.go
```diff
@@ -991,55 +1052,86 @@ func (sc *SettingController) updateNodeSelector() error {
...
+		case *corev1.Pod:
+			pod := obj.(*corev1.Pod)
+			if pod.DeletionTimestamp == nil {
+				sc.logger.Infof("Deleting pod %v to update the node selector from %v to %v", pod.Name, pod.Spec.NodeSelector, newNodeSelector)
+				if err := sc.ds.DeletePod(pod.Name); err != nil {
+					return err
+				}
+			}
+		default:
+			return fmt.Errorf("Unknown object type %v when updating %v setting", objType, types.SettingNamePriorityClass)
```
derekbit, line 1100:
```suggestion
			return fmt.Errorf("unknown object type %v when updating %v setting", objType, types.SettingNamePriorityClass)
```

controller/setting_controller.go
```diff
@@ -991,55 +1052,86 @@ func (sc *SettingController) updateNodeSelector() error {
...
+			ds := obj.(*appsv1.DaemonSet)
+			oldNodeSelector = ds.Spec.Template.Spec.NodeSelector
+		case *appsv1.Deployment:
+			dp := obj.(*appsv1.Deployment)
+			oldNodeSelector = dp.Spec.Template.Spec.NodeSelector
+		case *corev1.Pod:
+			pod := obj.(*corev1.Pod)
+			oldNodeSelector = pod.Spec.NodeSelector
+		default:
+			return nil, fmt.Errorf("Unknown object type %v when updating %v setting", objType, types.SettingNameSystemManagedComponentsNodeSelector)
```
derekbit, line 1122:
```suggestion
			return nil, fmt.Errorf("unknown object type %v when updating %v setting", objType, types.SettingNameSystemManagedComponentsNodeSelector)
```

controller/setting_controller.go
```diff
@@ -735,41 +738,81 @@ func (sc *SettingController) updatePriorityClass() error {
...
+			ds := obj.(*appsv1.DaemonSet)
+			oldPriorityClassName = ds.Spec.Template.Spec.PriorityClassName
+		case *appsv1.Deployment:
+			dp := obj.(*appsv1.Deployment)
+			oldPriorityClassName = dp.Spec.Template.Spec.PriorityClassName
+		case *corev1.Pod:
+			pod := obj.(*corev1.Pod)
+			oldPriorityClassName = pod.Spec.PriorityClassName
+		default:
+			return nil, fmt.Errorf("Unknown object type %v when updating %v setting", objType, types.SettingNamePriorityClass)
```
derekbit, line 806:
```suggestion
			return nil, fmt.Errorf("unknown object type %v when updating %v setting", objType, types.SettingNamePriorityClass)
```

derekbit (commented):
In general, LGTM except for the code refactoring.

controller/setting_controller.go
mantissahz: Modified. Thanks

controller/setting_controller.go
mantissahz: Moved. And move the `CollectRuntimeObject` back to `setting_controller` because we need `*datastore.Datastore`.

controller/setting_controller.go
mantissahz: Modified. Thanks

controller/setting_controller.go
mantissahz: Modified. Thanks

controller/setting_controller.go
mantissahz: Modified. Thanks

controller/setting_controller.go
mantissahz: Modified. Thanks

derekbit (approved):
LGTM

mantissahz:
@Mergify backport v1.6.x

PR #2542, opened 2024-01-30
Revert v1.6.0 backport PR "Be more careful while cleaning up stale replicas"

shuo-wu:

[This](https://github.com/longhorn/longhorn-manager/pull/2537) should be backported to v1.6.1 rather than v1.6.0


derekbit (approved):
LGTM

derekbit:
Will apply the original PR to v1.6.x after v1.6.0 is out.
cc @shuo-wu @ejweber @innobead

PR #2545, opened 2024-02-01
fix(instance): add conditions when parsing instance

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/7501


ChanYiLin:
Hi @shuo-wu 
this also needs your help, thanks!

shuo-wu (approved):

PR #2548, opened 2024-02-02
[MANUALLY BACKPORT v1.4.x ] fix: mount the whole host to engine im

ChanYiLin:
ref:
https://github.com/longhorn/longhorn/issues/7501
https://github.com/longhorn/longhorn/issues/7730

To allow engine instance manager to inspect mount point, we need to mount the whole host to the pod
After v1.5.x, engine/replica instance manager become one, and it mount the whole host to the pod as well.


derekbit (approved):
LGTM

innobead:
Won't fix in 1.4, as it's EOL. If users encounter this issue, suggest them to upgrade to 1.5.x.

PR #2550, opened 2024-02-07
feat(volume): remount volume if it becomes read only

ChanYiLin:


derekbit:
The solution looks good to me.
I'm thinking whether an application returns to a normal state or functions as expected after the RO volume switches back to RW without a restart, but this is primarily the responsibility of the application. We can see users' feedback.

controller/volume_controller.go
```diff
@@ -1521,10 +1521,25 @@ func (c *VolumeController) requestRemountIfFileSystemReadOnly(v *longhorn.Volume
...
+
+			engineClientProxy, err := engineapi.GetCompatibleClient(e, engineCliClient, c.ds, c.logger, c.proxyConnCounter)
+			if err != nil {
+				log.WithError(err).Warnf("Failed to get engineClientProxy when remounting read only volume")
+				return
+			}
+			defer engineClientProxy.Close()
+
+			if err := engineClientProxy.RemountReadOnlyVolume(e); err != nil {
+				log.WithError(err).Warnf("Failed to get remount read only volume")
```
derekbit, line 1539:
```suggestion
				log.WithError(err).Warnf("Failed to remount read only volume")
```

ChanYiLin: Fixed, thanks!

ChanYiLin:
Wait for instance-manager PR merged then fix the conflicts

ChanYiLin:
> The solution looks good to me. I'm thinking whether an application returns to a normal state or functions as expected after the RO volume switches back to RW without a restart, but this is primarily the responsibility of the application. We can see users' feedback.
> 
> Suggest clearly describing the behavior in the official document.

Hi @derekbit I have tested that after remounting the mount point to `rw`
The IO works fine.
But you are right, there are lots of cases not in our control, I will update the doc

innobead (approved):

ChanYiLin:
@mergify backport v1.6.1

ChanYiLin:
@mergify backport v1.6.x

PR #2551, opened 2024-02-07
[BACKPORT v1.5.x] feat(volume): remount volume if it becomes read only

ChanYiLin:


ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/7844

innobead (approved):
LGTM. Waiting for dep update.

shuo-wu (approved):

PR #2533, opened 2024-01-25
Enable mTLS in for the instance manager proxy service

ejweber:
#### Which issue(s) this PR fixes:

longhorn/longhorn#7040

#### What this PR does / why we need it:

This PR is an alternative to https://github.com/longhorn/longhorn-manager/pull/2527. In my testing, that one was not effective for the reason highlighted in: https://github.com/longhorn/longhorn-manager/pull/2527#pullrequestreview-1844281313.

#### Special notes for your reviewer:

I added https://github.com/longhorn/longhorn-manager/commit/3d54be9fabf35a2feaa5866d42417f9f49e3351a because of concerning behavior I was seeing without it. It appears that somehow the connections we create while attempting TLS remain open (perhaps indefinitely), even after we fall back to non-TLS. Without that commit, attempting to run with longhorn-managers that know the TLS secret and instance-managers that do not starts to flood the logs with hundreds of messages like the ones below per second. This is far more than the number of attempts longhorn-managers are actually making to connect to instance-managers.

```
[longhorn-manager-t78p7] W0125 22:18:28.327574       1 logging.go:59] [core] [Channel #797 SubChannel #798] grpc: addrConn.createTransport failed to connect to {Addr: "10.42.2.73:8501", ServerName: "longhorn-backend.longhorn-system", }. Err: connection error: desc = "transport: authentication handshake failed: tls: first record does not look like a TLS handshake"
[longhorn-manager-t78p7] W0125 22:18:28.328654       1 logging.go:59] [core] [Channel #1077 SubChannel #1079] grpc: addrConn.createTransport failed to connect to {Addr: "10.42.2.73:8501", ServerName: "longhorn-backend.longhorn-system", }. Err: connection error: desc = "transport: authentication handshake failed: tls: first record does not look like a TLS handshake"
[longhorn-manager-t78p7] W0125 22:18:28.334602       1 logging.go:59] [core] [Channel #493 SubChannel #494] grpc: addrConn.createTransport failed to connect to {Addr: "10.42.2.73:8501", ServerName: "longhorn-backend.longhorn-system", }. Err: connection error: desc = "transport: authentication handshake failed: tls: first record does not look like a TLS handshake"
[...]

PR #2552, opened 2024-02-07
[BACKPORT v1.4.x] feat(volume): remount volume if it becomes read only

ChanYiLin:


ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/7845

ChanYiLin:
We backport this change to instance-manager v1.4.x
https://github.com/longhorn/longhorn-instance-manager/pull/382#issuecomment-1911186398

but we didn't backport it in manager
https://github.com/longhorn/longhorn-manager/pull/2533

cc @derekbit @innobead

innobead:
@ChanYiLin the PR check failed.

innobead:
Won't fix in 1.4, as it's EOL. If users encounter this issue, suggest them to upgrade to 1.5.x.

PR #2558, opened 2024-02-07
Don't automatically upgrade engines while migrating

ejweber:

longhorn/longhorn#7833

The key change here is that we prevent automatic live engine upgrade while migrating. (We already checked for migration for user-directed live engine upgrades.)

In the process of checking the user-directed flow, I noticed the check for migration there was different than the one for the automatic flow (which uses a nice new utility function). I moved the utility function so it could be used in both locations.


ejweber:
To test:

Follow steps 1-7 in the reproduce (https://github.com/longhorn/longhorn/issues/7833#issue-2115652385). Then:

8. Notice that the volume is NOT automatically upgraded.

image

9. Detach the volume from NodeB. The kubectl command completes successfully and the VolumeAttachment goes away.

```
eweber@laptop:~/issue-files/7833> k delete -f va-2.yaml 
volumeattachment.storage.k8s.io "test-va-2" deleted
eweber@laptop:~/issue-files/7833> k get volumeattachment
NAME        ATTACHER             PV                                         NODE                                ATTACHED   AGE
test-va-1   driver.longhorn.io   pvc-991a9849-aebe-4a3a-ae6a-81207a073a15   eweber-v126-worker-9c1451b4-kgxdq   true       15m
```

10. Wait a few seconds and notice that the volume is successfully automatically upgraded. (Since it was detached from NodeB, the migration was cancelled, and the automatic upgrade proceeded easily.)

image


11. Detach the volume from NodeA. The kubectl command completes successfully and the VolumeAttachment goes away.

```
eweber@laptop:~/issue-files/7833> kl delete -f va-1.yaml 
Warning: deleting cluster-scoped resources, not scoped to the provided namespace
volumeattachment.storage.k8s.io "test-va-1" deleted
eweber@laptop:~/issue-files/7833> k get volumeattachment
No resources found
```

james-munson (approved):
Looks good.

ejweber:

PR #2563, opened 2024-02-07
Add the AD state machine graph

PhanLe1010:
Fix a TODO: add the AD state machine graph to the code for better readability


ejweber (approved):
LGTM.

PR #2573, opened 2024-02-08
fix(setting): check both setting and definition default for possible …

james-munson:
…change.

Issue https://github.com/longhorn/longhorn/issues/7755

When checking a potential configmap change for default settings, check not only the current setting value but also the setting definition default value to decide whether there is a change to apply.

### DIFF
--- a/datastore/longhorn.go
+++ b/datastore/longhorn.go
@@ -54,7 +54,7 @@ func (s *DataStore) UpdateCustomizedSettings(defaultImages map[types.SettingName
 		return err
 	}
 
-	availableCustomizedDefaultSettings := s.filterCustomizedDefaultSettings(customizedDefaultSettings)
+	availableCustomizedDefaultSettings := s.filterCustomizedDefaultSettings(customizedDefaultSettings, defaultSettingCM.ResourceVersion)
 
 	if err := s.applyCustomizedDefaultSettingsToDefinitions(availableCustomizedDefaultSettings); err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


datastore/longhorn.go
```diff
@@ -124,10 +124,15 @@ func (s *DataStore) isSettingValueChanged(name types.SettingName, value string)
...
+
+	// Also check the setting definition.
+	definition, ok := types.GetSettingDefinition(name)
+	if !ok {
+		logrus.Errorf("customized default setting %v is not defined", name)
+		return true
 	}
-	return true
+
+	return (setting.Value != value || definition.Default != value)
```
shuo-wu, line 135: This function checks if a customized value can be set as the definition default value. Then I am thinking if `setting.Value != value`. BTW, we may need to change the function name as well.

james-munson: The function checks whether a change would have any effect or can be skipped.  The logic for changing the configmap default definition also propagates the change to the global setting, regardless of its current value.  To decide that there is a change to be made, it needs to compare to both, because the global setting might be different from the default definition.

I agree that the function could be better named, so I've pushed that improvement.

james-munson:
Rebased and re-pushed.

james-munson:
Rebased and re-pushed.  What is holding up this PR?

ejweber (approved):
LGTM.

datastore/longhorn.go
shuo-wu: NIT: The new naming is confusing as well... I am thinking if we can move the config map resource version check into this function and rename this function to `shouldApplyCustomizedSettingValue`

james-munson: `shouldApplyCustomizedSettingValue` it shall be.
For the other, are you asking for the check of config map resource version in the setting's Annotations that is currently done in `syncSettingCRsWithCustomizedDefaultSettings` to be moved to `shouldApplyCustomizedSettingValue`?
Done.

james-munson: I think I'm caught up with the feedback items.  Can you check and approve, @shuo-wu?  Thanks!

shuo-wu (approved):

james-munson:

PR #2579, opened 2024-02-09
Delete default StorageClass if ConfigMap doesn't exist

ejweber:

longhorn/longhorn#4648

Deletes the default StorageClass if the longhorn-storageclass ConfigMap doesn't exist.

If a user doesn't want Longhorn to manage the default StorageClass, they can remove the ConfigMap from the cluster (or instruct Helm to never create it).

I ended up with this approach because we shouldn't use the finalizer pattern (in which we maintain a finalizer on the ConfigMap until we delete the StorageClass) because we don't want to set a finalizer on the user-created ConfigMap. However, it occurs to me that an owner reference pattern may be appropriate. (Set the owner reference on the StorageClass so that it is owned by the ConfigMap. When the ConfigMap is deleted, so is the StorageClass.) It's a little weird on the first upgrade if the user deletes the ConfigMap before the StorageClass gets an owner reference, but I'll experiment with it on Monday.

### DIFF
--- a/controller/kubernetes_configmap_controller.go
+++ b/controller/kubernetes_configmap_controller.go
@@ -107,6 +107,9 @@ func (kc *KubernetesConfigMapController) Run(workers int, stopCh <-chan struct{}
 }
 
 func (kc *KubernetesConfigMapController) worker() {
+	// Since we don't apply a finalizer to the longhorn-storageclass ConfigMap, we may miss its deletion while we are
+	// not running. Try to reconcile it once at startup.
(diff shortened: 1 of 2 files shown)


james-munson (approved):
That approach (deleting the default storage class if the configmap defining it has been deleted) seems reasonable.  LGTM

ejweber:
We decided to close https://github.com/longhorn/longhorn/issues/4648 instead of continuing.

PR #2580, opened 2024-02-12
Explicitly set preserveUnknownFields to false

ejweber:
Longhorn 7887

longhorn/longhorn#7887

If Longhorn and Kubernetes are upgraded in a particular order from Longhorn `v1.0.2-` to the latest, `preserveUnknownFields` may remain `true`, even though the default (and intended value) is `false`. As a result, Kubernetes does not consider affected CRDs to have structural schemas, potentially resulting in unexpected behavior (including https://github.com/longhorn/longhorn/issues/7887).


james-munson (approved):

ejweber:

innobead:
cc @derekbit

PR #2584, opened 2024-02-16
Enable spdk_tgt debug log if the global setting log-level

derekbit:
To make debugging process easier, enable spdk_tgt debug log if the global setting log-level is Debug.

Longhorn 7939

Issue https://github.com/longhorn/longhorn/issues/7939

### DIFF
--- a/controller/instance_manager_controller.go
+++ b/controller/instance_manager_controller.go
@@ -1201,9 +1201,20 @@ func (imc *InstanceManagerController) createInstanceManagerPodSpec(im *longhorn.
 	podSpec.Spec.Containers[0].Name = "instance-manager"
 
 	if datastore.IsDataEngineV2(dataEngine) {
-		podSpec.Spec.Containers[0].Args = []string{
-			"instance-manager", "--enable-spdk", "--debug", "daemon", "--spdk-enabled", "--listen", fmt.Sprintf("0.0.0.0:%d", engineapi.InstanceManagerProcessManagerServiceDefaultPort),
+		setting, err := imc.ds.GetSettingWithAutoFillingRO(types.SettingNameLogLevel)
+		if err != nil {
+			return nil, err
 		}
+
+		logLevel := strings.ToLower(setting.Value)
+
+		args := []string{"instance-manager"}
+		if logLevel == "debug" || logLevel == "trace" {
+			args = append(args, "--spdk-log", "all")
+		}
+		args = append(args, "--enable-spdk", "--debug", "daemon", "--spdk-enabled", "--listen", fmt.Sprintf("0.0.0.0:%d", engineapi.InstanceManagerProcessManagerServiceDefaultPort))
+
+		podSpec.Spec.Containers[0].Args = args
 
 		hugepage, err := imc.ds.GetSettingAsInt(types.SettingNameV2DataEngineHugepageLimit)
 		if err != nil {


controller/instance_manager_controller.go
```diff
@@ -1201,10 +1201,21 @@ func (imc *InstanceManagerController) createInstanceManagerPodSpec(im *longhorn.
...
+		setting, err := imc.ds.GetSettingWithAutoFillingRO(types.SettingNameLogLevel)
+		if err != nil {
+			return nil, err
 		}
 
+		logLevel := strings.ToLower(setting.Value)
+
+		args := []string{"instance-manager"}
+		if logLevel == "debug" || logLevel == "trace" {
+			args = append(args, "--spdk-log", "all")
```
innobead, line 1213: Do we have this support in the instance manager already? Can we just pass the general log levels which SPDK supports? so we don't need to do only for debug or trace level. WDYT?

[/lib/event/app.c#L680-L692](https://github.com/longhorn/spdk/blob/6925ca6543e4ad5a5e3a5db09217ae16bf939831/lib/event/app.c#L680-L692)
```c
	tty = ttyname(STDERR_FILENO);
	if (opts->print_level > SPDK_LOG_WARN &&
	    isatty(STDERR_FILENO) &&
	    tty &&
	    !strncmp(tty, "/dev/tty", strlen("/dev/tty"))) {
		printf("Warning: printing stderr to console terminal without -q option specified.\n");
		printf("Suggest using --silence-noticelog to disable logging to stderr and\n");
		printf("monitor syslog, or redirect stderr to a file.\n");
		printf("(Delaying for 10 seconds...)\n");
		sleep(10);
	}

	spdk_log_set_print_level(opts->print_level);
```

derekbit: > Do we have this support in the instance manager already?

Yes

> Can we just pass the general log levels which SPDK supports? so we don't need to do only for debug or trace level. WDYT?

Sure, but this needs to introduce a new setting or spec field. WDYT?

derekbit: BTW, the log level and flag in SPDK is very complicated.
```
 -L, --logflag <flag>    enable log flag (all, accel, accel_ioat, aio, app_config, app_rpc, bdev, bdev_concat, bdev_ftl, bdev_malloc, bdev_null, bdev_nvme, bdev_raid, bdev_raid0, bdev_raid1, bdev_raid_sb, blob, blob_esnap, blob_rw, blobfs, blobfs_bdev, blobfs_bdev_rpc, blobfs_rw, ftl_core, ftl_init, gpt_parse, ioat, iscsi, json_util, log, log_rpc, lvol, lvol_rpc, nbd, notify_rpc, nvme, nvme_vfio, nvmf, nvmf_tcp, opal, reactor, rpc, rpc_client, scsi, sock, sock_posix, thread, trace, vbdev_delay, vbdev_gpt, vbdev_lvol, vbdev_opal, vbdev_passthru, vbdev_split, vbdev_zone_block, vfio_pci, vfio_user, vhost, vhost_blk, vhost_blk_data, vhost_ring, vhost_rpc, vhost_scsi, vhost_scsi_data, vhost_scsi_queue, virtio, virtio_blk, virtio_dev, virtio_pci, virtio_user, virtio_vfio_user, vmd)
```
I'm thinking we can only support `Debug` and `Trace` which are just used for enabling `all` logs.

derekbit:
Wait... I think I can close the PR. We can control the flag and level in runtime env by spdk's rpc.py

PR #2586, opened 2024-02-16
Clean up defunct patches and scripts

ejweber:

longhorn/longhorn#5945

General code cleanup.


james-munson (approved):
If I understand you right, 
- The patches are long since obsolete since upstream caught up to them multiple releases ago.  
- The fact that they can't really be applied implies that the scripts that try to apply them are probably also unnecessary as well.
I buy that.  LGTM.

innobead (approved):

PR #2611, opened 2024-02-20
controller: DR volume should wait for the latest backup restored before activation

shuo-wu:

Issue https://github.com/longhorn/longhorn/issues/7945

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -3158,6 +3158,9 @@ func (c *VolumeController) checkAndFinishVolumeRestore(v *longhorn.Volume, e *lo
 	// 3) The restore/DR volume is
 	//   3.1) it's state `Healthy`;
 	//   3.2) or it's state `Degraded` with all the scheduled replica included in the engine
+	if v.Spec.FromBackup == "" || !v.Status.IsStandby {
+		return nil
+	}
 	isPurging := false
 	for _, status := range e.Status.PurgeStatus {
 		if status.IsPurging {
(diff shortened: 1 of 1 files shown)


shuo-wu:
The v1.5.x regression test: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/6302/

controller/volume_controller.go
```diff
@@ -3146,49 +3146,72 @@
...
~		!v.Spec.Standby) {
 		return nil
 	}
 
+	// Make sure the backup volume is up-to-date and the latest backup is restored
+	_, bvName, _, err := backupstore.DecodeBackupURL(v.Spec.FromBackup)
+	if err != nil {
+		return errors.Wrapf(err, "failed to get backup name from volume %s backup URL %v", v.Name, v.Spec.FromBackup)
+	}
+	bv, err := c.ds.GetBackupVolume(bvName)
```
innobead, line 3181: `GetBackupVolumeRO`

controller/volume_controller.go
```diff
@@ -3170,6 +3173,26 @@ func (c *VolumeController) checkAndFinishVolumeRestore(v *longhorn.Volume, e *lo
...
+	if err != nil {
+		return err
+	}
+	if !bv.Status.LastSyncedAt.IsZero() &&
+		bv.Spec.SyncRequestedAt.After(bv.Status.LastSyncedAt.Time) {
+		log.Infof("Restore/DR volume needs to wait for backup volume %s update", bvName)
+		return nil
```
innobead, line 3188: Should we requeue the volume?

innobead (changes requested):
In general, LGTM. Just a few comments.

controller/volume_controller.go
shuo-wu: Here the volume controller should wait for the backup volume controller to sync up this `bv`.

shuo-wu:

PR #2624, opened 2024-02-22
fix(supportbundle): update support bundle image pull policy to if not present

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/7998

Current image pull policy of support bundle is `Always` which is not necessary.
Change to `IfNotPresent`


innobead (approved):

innobead:

c3y1huang (changes requested):
And [here](https://github.com/longhorn/longhorn-manager/blob/v1.6.0/controller/support_bundle_controller.go#L727).

ChanYiLin:
Hi @c3y1huang fixed, thanks!
I have tested in my local env, and it is now using `IfNotPresent`

innobead (approved):

PR #2628, opened 2024-02-22
controller: Allow DR volume activation if the latest backup is not found

shuo-wu:
Longhorn 7997

https://github.com/longhorn/longhorn/issues/7997

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -3120,15 +3120,19 @@ func (c *VolumeController) checkAndFinishVolumeRestore(v *longhorn.Volume, e *lo
 			return err
 		}
 		if bv != nil {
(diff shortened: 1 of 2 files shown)


innobead:
@shuo-wu don't we need to fix this to master, 1.6 as well?

shuo-wu:
Yes. I will create a PR for the master and backport v1.6.x later.

shuo-wu:
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/6370/
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/6371/

shuo-wu:
The only failed test case `test_migration_with_restore_volume` works fine in my local env...

innobead:
@shuo-wu We need to fix the master and 1.6.

PR #2627, opened 2024-02-22
Fix a race condition in volumeattachment controller

PhanLe1010:
The race condition is that we sometimes mark the volumeattachment ticket as satisfied even though the volume has empty spec.NodeID and has started the detaching process.

More context is at https://github.com/longhorn/longhorn/issues/7937#issuecomment-1958604698

longhorn/longhorn#7937

PR #2629, opened 2024-02-22
fix(recurring-job): avoid intermediate volume detachment during backup

c3y1huang:

Issue longhorn/longhorn#7937

This was initially created to resolve the issue in longhorn/longhorn#7937, and @PhanLe1010 proposed an alternative [solution](https://github.com/longhorn/longhorn-manager/pull/2627) to tackle the problem.

However, Longhorn could still benefit by avoiding the unnecessary volume detachment during the recurring job backup. Currently, the recurring job runs the snapshot followed by the backup, resulting in an intermediate volume detachment.

`None`

### DIFF
--- a/app/recurring_job.go
+++ b/app/recurring_job.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"reflect"
 	"sort"
 	"strconv"
 	"strings"
@@ -20,6 +21,7 @@ import (
 	"k8s.io/client-go/tools/record"
 
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	clientset "k8s.io/client-go/kubernetes"
 	typedv1core "k8s.io/client-go/kubernetes/typed/core/v1"
@@ -48,6 +50,8 @@ const (
 )
 
 type Job struct {
+	name string
+
 	logger       logrus.FieldLogger
 	lhClient     lhclientset.Interface
 	namespace    string
@@ -184,15 +188,18 @@ func startVolumeJob(
 	job, err := newJob(
 		logger,
 		managerURL,
+		jobName,
 		volumeName,
 		snapshotName,
 		jobLabelMap,
 		jobRetain,
-		jobTask)
+		jobTask,
+	)
 	if err != nil {
 		log.WithError(err).Error("Failed to create new job for volume")
 		return err
 	}
+
 	err = job.run()
 	if err != nil {
 		log.WithError(err).Errorf("Failed to run job for volume")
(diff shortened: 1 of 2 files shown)


PhanLe1010:
Thank you for the idea! I have some feedbacks as:

1. We identified the root cause and fix the root cause. The test is passing 100/100 times now. So the benefit of this approach has become limited as the root cause has been fixed.
1. This approach might hide the potential races that should be been caught by e2e tests. Then other clients other than the recurring job might still hit the races.
1. If  we are creating the VA ticket for recurring jobs there are a few concerns:
    1. This PR currently set the VA ticket type as `longhorn.AttacherTypeBackupController`. It is clearly that the VA ticket is not created by backup controller.
    1. If we change the PR to use a different type `AttacherTypeRecurringJobApp` we will encounter other issues:
        1. What if the recurring job created the VA ticket then it is deleted. Now, no one is going to cleanup the VA ticket and volume stuck in attached state forever. 
        3. Even worse, the `AttacherTypeRecurringJobApp` is not interuptable so the workload pod cannot attach the volume. We can certainly make it interruptable but the logic will become more complicated.

Given the small benefit and the potential downside, I suggest that we should not go with this approach. However, if the team strongly agrees with this approach, it would be understandable for me.

c3y1huang:
> We identified the root cause and fix the root cause. The test is passing 100/100 times now. So the benefit of this approach has become limited as the root cause has been fixed./

The intention of this PR is not to overtake your approach, but to avoid intermediate detachment when recurring job runs backup.

> This PR currently set the VA ticket type as longhorn.AttacherTypeBackupController. The backup controller does not create the VA ticket.

I was aiming for `AttacherTypeRecurringJobApp`.

> If we change the PR to use a different type AttacherTypeRecurringJobApp we will encounter other issues:
If the recurring job created the VA ticket then it is deleted. Now, no one is going to cleanup the VA ticket and volume stuck in attached state forever.

The recurring job controller should handle the case (missed in this PR). https://github.com/longhorn/longhorn-manager/blob/master/controller/recurring_job_controller.go#L355 

> Even worse, the AttacherTypeRecurringJobApp is not interuptable so the workload pod cannot attach the volume. We can certainly make it interruptable but the logic will become more complicated.

Could you elaborate?

> This approach might hide the potential races that should be been caught by e2e tests. Then other clients other than the recurring job might still hit the races.

I agree with your point here even though this case was specifically for the recurring job. So I've closed this PR unless others feel this is beneficial.

PhanLe1010:
> Even worse, the AttacherTypeRecurringJobApp is not interuptable so the workload pod cannot attach the volume. We can certainly make it interruptable but the logic will become more complicated.
> 
> Could you elaborate?

Hi @c3y1huang , the situation I was mentioning is that while the recurring job attaches the volume to take a backup, if the workload pod is scaled up and the pod is scheduled to a new node, the pod cannot attach the volume because the recurring job is still holding the volume on different node. If backup take 1 hour, the pod will not be able to start for 1 hour.


Thank you for the explanation about other points! Yeah, we can see the feedback from others. If the team strongly agrees with this approach, it would be understandable for me.

PR #2635, opened 2024-02-23
Fix bug we initiate a backup when the volume has started detaching

PhanLe1010:
When the volume.Spec.NodeID is different than the node ID of the backup VA ticket, we should not initiate a backup as the volume is going to detach soon

longhorn/longhorn#7937

This PR is going to replace the approach at the PR https://github.com/longhorn/longhorn-manager/pull/2627. After discussed with @ejweber and @james-munson , we think that this approach is better because:
1. The other approach only touches the logic for the VA ticket of RWO non-migratible volume. It still leaves the logic for the RWX and mitigrable volume the same as we cannot compare the volume.Spec.NodeID with the node ID of these VA tickets. They are most likely to be different
1. It is fundamentally still correct that when the volume start detaching (but not finish detaching yet) the VA ticket can set to satisfied because at this exact moment the volume is still attached. VA controller can set reset the VA ticket to satisfied:false when the volume finish detaching. It is the responsibility of the client (backup controller in this case) to check and make sure that the volume.Spec.NodeID is still the desired one

### DIFF
--- a/controller/backup_controller.go
+++ b/controller/backup_controller.go
@@ -531,8 +531,9 @@ func (bc *BackupController) VerifyAttachment(backup *longhorn.Backup, volumeName
 	}
 
 	attachmentTicketID := longhorn.GetAttachmentTicketID(longhorn.AttacherTypeBackupController, backup.Name)
+	isVolumeStillDesiredToAttachToSameNodeAsAttachmentTicket := vol.Spec.NodeID == longhorn.GetNodeIdOfAttachmentTicket(attachmentTicketID, va)
 
-	return longhorn.IsAttachmentTicketSatisfied(attachmentTicketID, va), nil
+	return isVolumeStillDesiredToAttachToSameNodeAsAttachmentTicket && longhorn.IsAttachmentTicketSatisfied(attachmentTicketID, va), nil
 }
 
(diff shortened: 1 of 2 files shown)


c3y1huang (approved):

PhanLe1010:
Test run:
* Master: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/6379/
* v1.5.x: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/6385/

innobead (approved):

innobead:
Still need to follow up with the test results later, even though the PR has been merged.

innobead:

innobead:
The fix is general, so it should be expected to happen in 1.6 and 1.5 after VA was introduced. Do you know why this issue can reproduce after 1.5.4-RC2? @PhanLe1010 @c3y1huang

PhanLe1010:
Hi @innobead my currenty theory is that we change the code flow somehow and make it longer between volume.Spec.NodeID being cleanup and the time that volume start detaching by setting the engine.spec.desirestate to stop. But that is just an idea and I couldn't find proof in from the `git diff` of recent commit. I belive @c3y1huang and @mantissahz already search the commit history too. We couldn't pin down the exact commit that make the race become more visible

innobead:
That's reasonable and should be fine, as we already clarified the fix here.

PR #1029, opened 2021-09-03
Fix bug: Longhorn wrongly restart downloading for the ready backing image 

shuo-wu:
longhorn/longhorn#2957

PR #2641, opened 2024-02-23
refactor(setting): add value range in setting definition and use it in validation

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/7441

Some ideas:
1. **use map to present max and min value** instead of use int directly.
    - Because we still want to `omitempty` and `distinguish 0 or nil`. We then need to use `*int` in the structure so it can show 0 correctly and omitempy when nil.
    - We then need to declare a value first and assign it to structure as pointer for every definition which might make code not clean.
       ```
       val := 0
       difinition := SettingDefinition {
           Minimum: &val
       }
       ``` 
    - Use map, we can avoid the above problem.
2. For some multi-choice settings, we used another util function to validate it which is not correct
    - `SettingNameOfflineReplicaRebuilding`, `SettingNameSnapshotDataIntegrity`, `SettingNameBackupCompressionMethod`, `SettingNameReplicaAutoBalance` cc @derekbit @c3y1huang 
    - In the setting here, they should be global setting, so they are not allowed to have setting like `Ignored`.
    - `Ignored` should be used in volume level setting for it to follow global setting.
    - Thus, we shouldn't use util validation function like `ValidateReplicaAutoBalance()` to validate it because it allows `Ignored` setting.
    - I moved these settings to `multi-choice` category, so they follow the choices inside the `definition.Choice`.
3. I separated `< 0` and `< 1`, they should be the same, but I think it make developers more clear about the setting and no need to look up the code.

### DIFF
--- a/types/setting.go
+++ b/types/setting.go
@@ -19,7 +19,6 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 
 	"github.com/longhorn/longhorn-manager/meta"
-	"github.com/longhorn/longhorn-manager/util"
 
 	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
 )
@@ -41,6 +40,9 @@ const (
 	SettingTypeInt        = SettingType("int")
 	SettingTypeBool       = SettingType("bool")
 	SettingTypeDeprecated = SettingType("deprecated")
+
+	ValueIntRangeMinimum = "minimum"
+	ValueIntRangeMaximum = "maximum"
 )
 
 type SettingName string
@@ -234,6 +236,8 @@ type SettingDefinition struct {
 	ReadOnly    bool            `json:"readOnly"`
 	Default     string          `json:"default"`
 	Choices     []string        `json:"options,omitempty"` // +optional
+	// Use map to present minimum and maximum value instead of using int directly, so we can omitempy and distinguish 0 or nil at the same time.
+	ValueIntRange map[string]int `json:"range,omitempty"` // +optional
 }
 
 var settingDefinitionsLock sync.RWMutex
@@ -360,6 +364,9 @@ var (
 		Required:    true,
 		ReadOnly:    false,
 		Default:     "300",
+		ValueIntRange: map[string]int{
+			ValueIntRangeMinimum: 0,
+		},
 	}
 
 	SettingDefinitionFailedBackupTTL = SettingDefinition{
@@ -373,6 +380,9 @@ var (
 		Required: true,
 		ReadOnly: false,
 		Default:  "1440",
+		ValueIntRange: map[string]int{
+			ValueIntRangeMinimum: 0,
+		},
 	}
 
 	SettingDefinitionRestoreVolumeRecurringJobs = SettingDefinition{
@@ -491,6 +501,9 @@ var (
 		Required:    true,
 		ReadOnly:    false,
 		Default:     "100",
+		ValueIntRange: map[string]int{
+			ValueIntRangeMinimum: 0,
+		},
 	}
 
 	SettingDefinitionStorageMinimalAvailablePercentage = SettingDefinition{
@@ -501,6 +514,10 @@ var (
 		Required:    true,
 		ReadOnly:    false,
 		Default:     "25",
+		ValueIntRange: map[string]int{
+			ValueIntRangeMinimum: 0,
+			ValueIntRangeMaximum: 100,
+		},
 	}
 
 	SettingDefinitionStorageReservedPercentageForDefaultDisk = SettingDefinition{
@@ -511,6 +528,10 @@ var (
 		Required:    true,
 		ReadOnly:    false,
 		Default:     "30",
+		ValueIntRange: map[string]int{
+			ValueIntRangeMinimum: 0,
+			ValueIntRangeMaximum: 100,
+		},
 	}
 
 	SettingDefinitionUpgradeChecker = SettingDefinition{
@@ -570,6 +591,10 @@ var (
 		Required:    true,
 		ReadOnly:    false,
 		Default:     "3",
+		ValueIntRange: map[string]int{
+			ValueIntRangeMinimum: 1,
+			ValueIntRangeMaximum: 20,
+		},
 	}
 
 	SettingDefinitionDefaultDataLocality = SettingDefinition{
@@ -783,6 +808,9 @@ var (
 		Required: true,
 		ReadOnly: false,
 		Default:  "600",
+		ValueIntRange: map[string]int{
+			ValueIntRangeMinimum: 0,
+		},
 	}
 
 	SettingDefinitionConcurrentReplicaRebuildPerNodeLimit = SettingDefinition{
@@ -798,6 +826,9 @@ var (
 		Required: true,
 		ReadOnly: false,
 		Default:  "5",
+		ValueIntRange: map[string]int{
+			ValueIntRangeMinimum: 0,
+		},
 	}
 
 	SettingDefinitionConcurrentVolumeBackupRestorePerNodeLimit = SettingDefinition{
@@ -810,6 +841,9 @@ var (
 		Required: true,
 		ReadOnly: false,
 		Default:  "5",
+		ValueIntRange: map[string]int{
+			ValueIntRangeMinimum: 0,
+		},
 	}
 
 	SettingDefinitionSystemManagedPodsImagePullPolicy = SettingDefinition{
@@ -868,6 +902,9 @@ var (
 		Required: true,
 		ReadOnly: false,
 		Default:  "0",
+		ValueIntRange: map[string]int{
+			ValueIntRangeMinimum: 0,
+		},
 	}
 
 	SettingDefinitionBackingImageCleanupWaitInterval = SettingDefinition{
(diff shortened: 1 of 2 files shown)


ChanYiLin:
E2E test: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/6460/console

ChanYiLin:
`-m coretest`
E2E test: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/6528/

types/setting.go
```diff
@@ -354,6 +359,9 @@ var (
 		Required:    true,
 		ReadOnly:    false,
 		Default:     "300",
+		ValueRange: map[string]int{
```
innobead, line 362: `ValueIntRange`, as this is for `SettingTypeInt`

types/setting.go
```diff
@@ -1371,44 +1454,7 @@ func ValidateSetting(name, value string) (err error) {
 			return fmt.Errorf("value %v of setting %v should be true or false", value, sName)
 		}
 
-	case SettingNameStorageOverProvisioningPercentage:
```
innobead, line 1374: Suggest having a general `validateInt` and run it for each setting at the beginning. If the updated setting is not int type, then bypass. 

In `validateInt`, it will check if the setting in the range no matter if it just has min or max, and provide a consistent validation error message. This will simplify the code.

P.S. If we have more validation later, then it can be a validation chain like middle layers for validation.

innobead (changes requested):

types/setting.go
ChanYiLin: Fixed, thanks

types/setting.go
ChanYiLin: Fixed, thanks

ChanYiLin:

Conflicts resolved.

types/setting.go
```diff
@@ -1768,3 +1608,129 @@ func GetDangerZoneSettings() sets.Set[SettingName] {
...
+	case SettingNameReplicaAutoBalance:
+		fallthrough
+	case SettingNameLogLevel:
+		fallthrough
+	case SettingNameSystemManagedPodsImagePullPolicy:
+		definition, _ := GetSettingDefinition(sName)
+		choices := definition.Choices
+		if !isValidChoice(choices, value) {
+			return fmt.Errorf("value %v is not a valid choice, available choices %v", value, choices)
+		}
```
shuo-wu, line 1688: We can blindly check `isValidChoice` as long as `definition.Choices != nil && len(definition.Choices) > 0`. No need to list all settings

types/types.go
```diff
@@ -1038,18 +1035,21 @@ func ValidateCPUReservationValues(settingName SettingName, instanceManagerCPUStr
...
+		isUnderLimit := instanceManagerCPU < valueIntRange[ValueIntRangeMinimumKey]
+		isOverLimit := instanceManagerCPU > valueIntRange[ValueIntRangeMaximumKey]
 		if isUnderLimit || isOverLimit {
-			return fmt.Errorf("invalid requested v1 data engine instance manager CPUs. Valid instance manager CPU range between 0%% - 40%%")
+			return fmt.Errorf("invalid requested v1 data engine instance manager CPUs. Valid instance manager CPU range between %v - %v millicpu", valueIntRange[ValueIntRangeMinimumKey], valueIntRange[ValueIntRangeMaximumKey])
 		}
 	case SettingNameV2DataEngineGuaranteedInstanceManagerCPU:
-		isUnderLimit := instanceManagerCPU < 1000
-		isOverLimit := instanceManagerCPU > 8000
+		isUnderLimit := instanceManagerCPU < valueIntRange[ValueIntRangeMinimumKey]
```
shuo-wu, line 1049:
```
case SettingNameGuaranteedInstanceManagerCPU, SettingNameV2DataEngineGuaranteedInstanceManagerCPU:
    ......
```

types/setting.go
```diff
@@ -41,6 +40,9 @@ const (
 	SettingTypeInt        = SettingType("int")
 	SettingTypeBool       = SettingType("bool")
 	SettingTypeDeprecated = SettingType("deprecated")
+
+	ValueIntRangeMinimumKey = "minimum"
+	ValueIntRangeMaximumKey = "maximum"
```
derekbit, line 45: Why not just `ValueIntRangeMinima` and `ValueIntRangeMaxima`?

types/setting.go
```diff
@@ -491,6 +501,9 @@ var (
 		Required:    true,
 		ReadOnly:    false,
 		Default:     "100",
+		ValueIntRange: map[string]int{
+			ValueIntRangeMinimumKey: 0,
```
derekbit, line 505: Can the Default be 0? If so, what will happen when Default is 0?

types/setting.go
derekbit: Do we need to handle `close bracket` and `open bracket` such as `(min, max)`, `[min, max]`, `(min, max]` and `[min, max)`

ChanYiLin: I think maximum, minimum is the correct spelling
Yeah I can remove the "key" suffix, maybe more easy to use when coding.

ChanYiLin: I don't think we need to handle `close bracket` and `open bracket`
That is too complicated.
Besides, when we say `maximum value`, `minimum` value
We usually include the value, and we do it in our original code.

For example:
- Our original Code:
```
	case SettingNameStorageMinimalAvailablePercentage:
		if _, err := strconv.Atoi(value); err != nil {
			return errors.Wrapf(err, "value %v is not a number", value)
		}
		// additional check whether minimal available percentage is between 0 to 100
		value, err := util.ConvertSize(value)
		if err != nil {
			return errors.Wrapf(err, "failed to parse %v as size", value)
		}
		if value < 0 || value > 100 {
			return fmt.Errorf("value %v should between 0 to 100", value)
		}
```
The minimum value, maximum value are included: [0, 100]
We check if the `value < minimum || value > max`
I think this convention is common to user and developer

types/setting.go
ChanYiLin: Yes it can be "0"
I define the value range by referring our original code
```
	case SettingNameStorageOverProvisioningPercentage:
		if _, err := strconv.Atoi(value); err != nil {
			return errors.Wrapf(err, "value %v is not a number", value)
		}
		// additional check whether over provisioning percentage is positive
		value, err := util.ConvertSize(value)
		if err != nil {
			return errors.Wrapf(err, "failed to parse %v as size", value)
		}
		if value < 0 {
			return fmt.Errorf("value %v should be positive", value)
		}
```

For this case, we only invalid the setting when it is < 0 , thats why I define the minimum as "0"

Other case would be the same.
But yes, it is a chance for us to review if the value range is correct.

types/setting.go
derekbit: > I think maximum, minimum is the correct spelling Yeah I can remove the "key" suffix, maybe more easy to use when coding.

Yeah, I mean `key` seems not needed.

derekbit: > I don't think we need to handle `close bracket` and `open bracket` That is too complicated. Besides, when we say `maximum value`, `minimum` value We usually include the value, and we do it in our original code.

OK. Just make sure if the `close bracket` is required. If not, I'm fine with the implementation.

types/setting.go
```diff
@@ -891,6 +931,9 @@ var (
 		Required: true,
 		ReadOnly: false,
 		Default:  "300",
+		ValueIntRange: map[string]int{
+			ValueIntRangeMinimumKey: 0,
```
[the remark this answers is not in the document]
ChanYiLin: It will re-download the BackingImage immediately when all the BackingImage on the disk is in error state  

ref: https://github.com/longhorn/longhorn-manager/pull/1029/files#diff-59782221a3f8c591d66ab9f73614211f67aba44b4ad4a31329f875f3ff0eeb25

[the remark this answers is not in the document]
ChanYiLin: For all these values, I use the same definition `minimum: 0` as original check

```
	case SettingNameBackingImageCleanupWaitInterval:
		fallthrough
	case SettingNameBackingImageRecoveryWaitInterval:
		fallthrough
	case SettingNameReplicaReplenishmentWaitInterval:
		fallthrough
	case SettingNameConcurrentReplicaRebuildPerNodeLimit:
		fallthrough
	case SettingNameConcurrentBackupRestorePerNodeLimit:
		fallthrough
	case SettingNameConcurrentAutomaticEngineUpgradePerNodeLimit:
		fallthrough
	case SettingNameSupportBundleFailedHistoryLimit:
		fallthrough
	case SettingNameBackupstorePollInterval:
		fallthrough
	case SettingNameRecurringSuccessfulJobsHistoryLimit:
		fallthrough
	case SettingNameRecurringFailedJobsHistoryLimit:
		fallthrough
	case SettingNameFailedBackupTTL:
		fallthrough
	case SettingNameV2DataEngineHugepageLimit:
		value, err := strconv.Atoi(value)
		if err != nil {
			errors.Wrapf(err, "value %v is not a number", value)
		}
		if value < 0 {
			return fmt.Errorf("the value %v shouldn't be less than 0", value)
		}
```

types/setting.go
ChanYiLin: Fixed, thanks

types/types.go
ChanYiLin: Fixed, thanks

types/setting.go
ChanYiLin: Fixed, thanks

shuo-wu (approved):

derekbit (approved):

innobead (approved):

PR #2645, opened 2024-02-23
feat(node): add node condition for environment check, starting with k…

james-munson:
…ernel release

Issue https://github.com/longhorn/longhorn/issues/7931

Add a Node Condition for environment check.

### DIFF
--- a/controller/node_controller.go
+++ b/controller/node_controller.go
@@ -2,6 +2,7 @@ package controller
 
 import (
 	"fmt"
+	"os/exec"
 	"reflect"
 	"strings"
 	"sync"
@@ -32,6 +33,7 @@ import (
 	"github.com/longhorn/longhorn-manager/types"
 	"github.com/longhorn/longhorn-manager/util"
 
+	lhns "github.com/longhorn/go-common-libs/ns"
 	"github.com/longhorn/longhorn-manager/controller/monitor"
 	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
 )
@@ -72,6 +74,18 @@ type NodeController struct {
 }
 
 type TopologyLabelsChecker func(kubeClient clientset.Interface, vers string) (bool, error)
+
+// Ref: https://github.com/longhorn/longhorn/issues/7931
+type defectiveKernelRange struct {
+	broken string
+	fixed  string
+}
+
+// TODO: Keep an up-to-date bad kernel list in upgrade-responder and pull the latest info from there.
+var knownDefectiveKernels = []defectiveKernelRange{
+	{"5.15.0-94", "5.15.0-100"},
+	{"6.5.6", "6.5.9"},
+}
 
 func NewNodeController(
 	logger logrus.FieldLogger,
@@ -539,7 +553,7 @@ func (nc *NodeController) syncNode(key string) (err error) {
 		}
 	}
 
-	// sync mount propagation status on current node
+	// sync status Conditions on current node.
 	for _, pod := range managerPods {
 		if pod.Spec.NodeName == node.Name {
 			if err := nc.syncNodeStatus(pod, node); err != nil {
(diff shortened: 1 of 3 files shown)


controller/node_controller.go
```diff
@@ -73,6 +75,17 @@ type NodeController struct {
...
+// Ref: https://github.com/longhorn/longhorn/issues/7931
+type defectiveKernelRange struct {
+	broken string
+	fixed  string
+}
+
+var knownDefectiveKernels = []defectiveKernelRange{
+	{"5.15.0-94", "5.15.0-100"},
+	{"6.5.6", "6.5.7"},
+}
```
PhanLe1010, line 87: I am wondering if we should dynamically fetch this list instead of hard-coding. We can use our upgrade responder server for this purpose. For airgap env, we can use a hard-coded list. WDYT?

k8s/pkg/apis/longhorn/v1beta2/node.go
```diff
@@ -6,6 +6,7 @@ const (
 	NodeConditionTypeReady            = "Ready"
 	NodeConditionTypeMountPropagation = "MountPropagation"
 	NodeConditionTypeSchedulable      = "Schedulable"
+	NodeConditionTypeEnvironmentCheck = "EnvironmentCheck"
```
PhanLe1010, line 9: Should we use the name `EnvironmentCheckPassed` instead of `EnvironmentCheck` so that the condition will read like this:
```
    - lastProbeTime: ""
      lastTransitionTime: "2024-02-23T20:44:36Z"
      message: Node jbm-u22-pool2-ca5a5aa4-5fcq7 running kernel 5.15.0-67-generic
      reason: ""
      status: "True"
      type: EnvironmentCheckPassed
```

controller/node_controller.go
```diff
@@ -972,9 +985,52 @@ func (nc *NodeController) syncNodeStatus(pod *corev1.Pod, node *longhorn.Node) e
...
+		}
+	}
+
 	return nil
 }
 
+// isDefectiveKernel duplicates the logic of longhorn/scripts/environment_check.sh because
+// * It should get the same answer, and
+// * "sort -V" is more accurate than "semver", because Linux kernels are not semver.
+func (nc *NodeController) isDefectiveKernel(kernelRelease string) bool {
```
ejweber, line 1012: Can we have a unit test of this? Something table driven, preferably that includes good and bad kernels from major distros (e.g. SLES, RHEL, Ubuntu).

I phrase it like "can we" because I think the decision to use `exec.Command` makes it difficult/impossible.

In general, I don't like calling the shell to do a sort, but I do get the logic for wanting to ensure we get the same answer here and in the environment check script.

k8s/pkg/apis/longhorn/v1beta2/node.go
james-munson: Done.

controller/node_controller_test.go
```diff
@@ -1973,3 +1985,36 @@ func newTestNodeController(lhClient *lhfake.Clientset, kubeClient *fake.Clientse
...
+		want          bool
+	}{
+		"Before both ranges":              {"5.15.0-67-generic", false},
+		"First broken version":            {"5.15.0-94", true},
+		"First broken version plus label": {"5.15.0-94-generic", true}, // semver gets this wrong.
+		"First broken version plus build": {"5.15.0-94+5ac3", true},    // semver gets this wrong.
+		"First fixed version":             {"5.15.0-100", false},
+		"Not in either range":             {"6.1.0-1029-oem", false},
+		"Second minor version":            {"6.5.0-18-generic", false},
+		"Before seconde broken version":   {"6.5.5", false},
```
ejweber, line 2006:
```suggestion
		"Before second broken version":   {"6.5.5", false},
```

controller/node_controller.go
```diff
@@ -73,6 +75,18 @@ type NodeController struct {
...
+type defectiveKernelRange struct {
+	broken string
+	fixed  string
+}
+
+// TODO: Keep an up-to-date bad kernel list in upgrade-responder and pull the latest info from there.
+var knownDefectiveKernels = []defectiveKernelRange{
+	{"5.15.0-94", "5.15.0-100"},
+	{"6.5.6", "6.5.9"},
+}
```
ejweber, line 88: I am wondering if, in addition to @PhanLe1010's suggestion to have somewhere we can pull these from dynamically (which can be handled in another ticket), we should have this behavior dynamically configurable on the cluster.

- Let's say we get it wrong. Maybe some distro is built on `v6.5.6` but it backports a fix such that `v6.5.6-1-fixedit` works.
- Let's say a user is using a broken kernel, but definitely does not intend on ever using NFSv4 or RWX in general.

Now the user sees perpetual red lights for a situation they don't care about.

Should we have a setting like:

- `disable-kernel-version-check`, or
- `ignore-kernel-version-check-for` which can include a list of kernels to ignore, or
- `known-defective-kernels` which includes the above list by default, but can be edited by the user as desired?

ejweber (commented):
This is shaping up nicely, but I still have some questions about our overall approach (the code itself is looking good to me).

- Does this work as expected in the UI, or do we need to add a UI label to the parent issue?
- Do we need to do more testing around vulnerable kernel versions as suggested by https://github.com/longhorn/longhorn/issues/7931#issuecomment-1968375788?

controller/node_controller.go
james-munson: Yes, I think an override setting would be useful (or maybe more than one, since it might be handy to have both "disable" and "override-list" capabilities.)  The latter, if we publish it somewhere that is easily copy/paste-able (the KB article itself, perhaps) also gives a handy solution to the air-gapped case that doesn't require a software upgrade to get new hard-coded values.

james-munson:
> This is shaping up nicely, but I still have some questions about our overall approach (the code itself is looking good to me).
> 
>     * Does this work as expected in the UI, or do we need to add a UI label to the parent issue?
> 
>     * Do we need to do more testing around vulnerable kernel versions as suggested by [[IMPROVEMENT] Improve environment_check script for NFS protocol bug and the host system self diagnosis  longhorn#7931 (comment)](https://github.com/longhorn/longhorn/issues/7931#issuecomment-1968375788)?

* We probably do need a UI component.  The failed condition does not show up in the UI currently, and it likely does need a specific call-out, because Conditions are not necessarily written so that any with a value of "False" can be assumed to be a Bad Thing.
* I would imagine that testing, verifying, and updating the list of known problem kernels will likely be an ongoing thing, given the apparent frequency with which we seem to stumble across yet another.

controller/node_controller.go
james-munson: That being said, at the moment, this gives us a handy troubleshooting item.  Would it be appropriate to push the other aspects into a follow-on ticket?

innobead (changes requested):
After reading this PR, the kernel versions will be dynamic factors to determine if the condition info is accurate.

The implementation is straightforward, but it's possibly over-engineering, as when the error happens due to failed mount, the log is outstanding and propagated to the pod event as well. The missing part is just the kernel version on that node. 

Suggest just adding the node identifier (ex: node name) and kernel version which should be good enough for users to do the next action. Surely, they can run the env check script to do a best-effort check. 

P.S. We should start thinking of moving operations for runtime check to our [CLI](https://github.com/longhorn/cli) in the future, which is currently planned for 1.7.

controller/node_controller.go
```diff
@@ -972,9 +986,52 @@ func (nc *NodeController) syncNodeStatus(pod *corev1.Pod, node *longhorn.Node) e
...
 	return nil
 }
 
+// isDefectiveKernel duplicates the logic of longhorn/scripts/environment_check.sh because
+// * It should get the same answer, and
+// * "sort -V" is more accurate than "semver", because Linux kernels are not semver.
+func isDefectiveKernel(kernelRelease string) bool {
+	verlte := func(v1, v2 string) bool {
+		in := v1 + "\n" + v2
+		cmd := exec.Command("bash", "-c", "sort -C -V")
```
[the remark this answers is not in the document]
james-munson: Sadly, Linux kernel versions are not semver, and using a library such as our go semver will give misleading answers on version comparison.

innobead:
@james-munson @derekbit WDYT?

controller/node_controller.go
shuo-wu: > have somewhere we can pull these from dynamically (which can be handled in another ticket), we should have this behavior dynamically configurable on the cluster.

Do you mean making it become a setting like `Stable Longhorn Versions`? Then the upgrade checker can help pull the info from the server, in which we can modify the kernel versions dynamically.

james-munson:
> Suggest just enriching the error info below with the node identifier (ex: node name) and kernel version which should be good enough for users to do the next action. Surely, they can run the env check script to do a best-effort check.
Agreed.  The desired direction here is to ditch the Condition and just augment the mount error logging with the node name and kernel release.

james-munson:
I will close this pull request, and open a new one with the changes to logging on mount error.

PR #2649, opened 2024-02-26
fix(daemon): validate webhook by its domain

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/8005

The root cause is that we now have conversion-webhook(`:9501`), admission-webhook(`:9502`) in longhorn-manager
We set the readiness port to `9501`
The order of starting is as follow
1. conversion-webhook
2. admission-webhook
3. upgrade
4. ....

So there is a chance that the pod is not ready yet (readiness on `9501` hasn't succeeded) when the longhorn-manager starts to do the upgrade 
It then fails to call the webhook by endpoint because there is no longhorn-manager pod ready. 
```
Post \"https://longhorn-admission-webhook.longhorn-system.svc:9502/v1/webhook/validaton?timeout=10s\"
````

This PR fix it by validating the endpoint works by testing the domain instead of local address before doing the upgrade.

### DIFF
--- a/webhook/webhook.go
+++ b/webhook/webhook.go
@@ -11,6 +11,7 @@ import (
 	"github.com/sirupsen/logrus"
 
 	"github.com/longhorn/longhorn-manager/types"
+	"github.com/longhorn/longhorn-manager/util"
 	"github.com/longhorn/longhorn-manager/util/client"
 	"github.com/longhorn/longhorn-manager/webhook/server"
 )
@@ -22,12 +23,12 @@ var (
 func StartWebhook(ctx context.Context, webhookType string, clients *client.Clients) error {
 	logrus.Infof("Starting longhorn %s webhook server", webhookType)
 
-	var webhookPort int
+	var webhookLocalEndpoint string
 	switch webhookType {
 	case types.WebhookTypeAdmission:
-		webhookPort = types.DefaultAdmissionWebhookPort
+		webhookLocalEndpoint = fmt.Sprintf("https://localhost:%d/v1/healthz", types.DefaultAdmissionWebhookPort)
 	case types.WebhookTypeConversion:
-		webhookPort = types.DefaultConversionWebhookPort
+		webhookLocalEndpoint = fmt.Sprintf("https://localhost:%d/v1/healthz", types.DefaultConversionWebhookPort)
 	default:
 		return fmt.Errorf("unexpected webhook server type %v", webhookType)
 	}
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by shuo-wu** [comment]
> So there is a chance that the pod is not ready yet (readiness on 9501 hasn't succeeded) when the longhorn-manager starts to do the upgrade
It then fails to call the webhook by endpoint because there is no longhorn-manager pod ready.

I didn't get the point of why this PR would resolve this case... And before a longhorn-manager pod starts to do the upgrade, it should already start the webhook services by itself.

**Review by ejweber — CHANGES_REQUESTED** [review]
So the problem is that the webhook IS ready on `localhost`, allowing us to proceed to the upgrade, but the pod is NOT added to the list of endpoints for `longhorn-admission-webhook` and `longhorn-conversion-webhook`, because the liveness probe hasn't yet succeeded. Is that correct?

I can see how this PR is a solution to that problem. Now we cannot proceed to the upgrade step until we are sure the endpoint is added and we can expect Kubernetes API operations to succeed during the upgrade.

However, it looks like we are no longer checking whether the webhook in THIS pod is available. The `localhost` check previously accomplished this. In a multi-node cluster, we may proceed past `Waiting for %v webhook to become ready` if there are ANY successfully running `longhorn-manager`s with endpoints added to `longhorn-admission-webhook` and `longhorn-conversion-webhook`.

I think we should retain the `localhost` check and simply add the `longhorn-admission-webhook` and `longhorn-conversion-webhook` checks. WDYT?

**Comment by ChanYiLin** [comment]
Hi @shuo-wu 
> And before a longhorn-manager pod starts to do the upgrade, it should already start the webhook services by itself.
Yes, correct, on the localhost.
However, because the k8s readiness check hasn't succeeded yet, K8s hasn't marked longhorn-manager pod as ready and added it to the endpoint of the k8s svc. So when doing the upgrade, it accessed the webhook by the endpoint `longhorn-admission-webhook.longhorn-system.svc:9502`, it then failed.


Hi @ejweber 
Yes you are right, we should still check the status of local webhook services in the current pod. Thanks!

**Comment by ChanYiLin** [comment]
Hi @ejweber 
I have added another check for the service availability, please take a look, thanks!

File: webhook/webhook.go [hunk 1]
```diff
@@ -40,37 +40,57 @@ func StartWebhook(ctx context.Context, webhookType string, clients *client.Clien
...
+	}
+	logrus.Infof("Started longhorn %s webhook server on localhost", webhookType)
+	return nil
+}
+
+func CheckWebhookServiceAvailability(webhookType string) error {
+	var webhookServiceEndpoint string
+	switch webhookType {
+	case types.WebhookTypeAdmission:
+		webhookServiceEndpoint = fmt.Sprintf("https://longhorn-admission-webhook.longhorn-system.svc:%d/v1/healthz", types.DefaultAdmissionWebhookPort)
```
innobead, line 54: Don't hard code the namespace here, even though the default is `longhorn-system` and we don't commit other namespaces will be supported.

You can use the follow function to get the namespace.

[/util/util.go#L581-L589](https://github.com/longhorn/longhorn-manager/blob/1956c63e2c5d55f03c2afba8e2bbf44a61218e59/util/util.go#L581-L589)
```go
func GetNamespace(key string) string {
	namespace := os.Getenv(key)
	if namespace == "" {
		logrus.Warnf("Failed to detect pod namespace, environment variable %v is missing, "+
			"using default namespace", key)
		namespace = corev1.NamespaceDefault
	}
	return namespace
}
```

Besides, use the existing constants (like `AdmissionWebhookServiceName`) instead of raw string here. Please review this in other places as well.

webhook/webhook.go
```diff
@@ -40,37 +40,57 @@ func StartWebhook(ctx context.Context, webhookType string, clients *client.Clien
...
+	case types.WebhookTypeConversion:
+		webhookServiceEndpoint = fmt.Sprintf("https://longhorn-conversion-webhook.longhorn-system.svc:%d/v1/healthz", types.DefaultConversionWebhookPort)
+	default:
+		return fmt.Errorf("unexpected webhook server type %v", webhookType)
+	}
+
+	if !isServiceAvailable(webhookServiceEndpoint) {
+		return fmt.Errorf("%v webhook service is not accessible after %v sec", webhookType, defaultStartTimeout)
```
innobead, line 62: Suggest refactor `isServiceAvailable` with another parameter for `timeout` and pass `defaultStartTimeout` for it. This way will make the following log with `defaultStartTimeout` making more sense.

innobead (changes requested):

webhook/webhook.go
ChanYiLin: Fixed, thanks!

webhook/webhook.go
ChanYiLin: Fixed.

innobead (approved):

innobead:

PR #2658, opened 2024-03-01
feat(backupbackingimage): add backend api

ChanYiLin:
ref: 
https://github.com/longhorn/longhorn/issues/8068
https://github.com/longhorn/longhorn/issues/7541

1. Get `BackupBackingImages (list)`
2. Get `BackupBackingImage`
3. Delete `BackupBackingImage`
4. For each BackupBackingImage, we have action `backupBackingImageRestore`
5. For each BackingImage, we have new action `backupBackingImageCreate`

### DIFF
--- a/api/model.go
+++ b/api/model.go
@@ -163,6 +163,20 @@ type Backup struct {
 	CompressionMethod      string               `json:"compressionMethod"`
 }
 
+type BackupBackingImage struct {
+	client.Resource
+	Name              string               `json:"name"`
+	State             longhorn.BackupState `json:"state"`
+	Progress          int                  `json:"progress"`
+	Error             string               `json:"error"`
+	URL               string               `json:"url"`
+	Created           string               `json:"created"`
+	Size              int64                `json:"size"`
+	Labels            map[string]string    `json:"labels"`
+	Messages          map[string]string    `json:"messages"`
+	CompressionMethod string               `json:"compressionMethod"`
+}
+
 type Setting struct {
 	client.Resource
 	Name       string                  `json:"name"`
@@ -631,6 +645,7 @@ func NewSchema() *client.Schemas {
 	snapshotSchema(schemas.AddType("snapshot", Snapshot{}))
 	snapshotCRSchema(schemas.AddType("snapshotCR", SnapshotCR{}))
 	backupVolumeSchema(schemas.AddType("backupVolume", BackupVolume{}))
+	backupBackingImageSchema(schemas.AddType("backupBackingImage", BackupBackingImage{}))
 	settingSchema(schemas.AddType("setting", Setting{}))
 	recurringJobSchema(schemas.AddType("recurringJob", RecurringJob{}))
 	engineImageSchema(schemas.AddType("engineImage", EngineImage{}))
@@ -715,7 +730,8 @@ func backingImageSchema(backingImage *client.Schema) {
 			Input:  "backingImageCleanupInput",
 			Output: "backingImage",
 		},
-		BackingImageUpload: {},
+		BackingImageUpload:         {},
+		"backupBackingImageCreate": {},
 	}
 
 	name := backingImage.ResourceFields["name"]
@@ -742,6 +758,15 @@ func backingImageSchema(backingImage *client.Schema) {
 	diskFileStatusMap := backingImage.ResourceFields["diskFileStatusMap"]
 	diskFileStatusMap.Type = "map[backingImageDiskFileStatus]"
 	backingImage.ResourceFields["diskFileStatusMap"] = diskFileStatusMap
+}
+
+func backupBackingImageSchema(backupBackingImage *client.Schema) {
+	backupBackingImage.CollectionMethods = []string{"GET"}
+	backupBackingImage.ResourceMethods = []string{"GET", "DELETE"}
+
+	backupBackingImage.ResourceActions = map[string]client.Action{
+		"backupBackingImageRestore": {},
+	}
 }
 
 func recurringJobSchema(job *client.Schema) {
@@ -1742,6 +1767,45 @@ func toBackupVolumeCollection(bvs []*longhorn.BackupVolume, apiContext *api.ApiC
 	return &client.GenericCollection{Data: data, Collection: client.Collection{ResourceType: "backupVolume"}}
 }
 
+func toBackupBackingImageResource(bbi *longhorn.BackupBackingImage, apiContext *api.ApiContext) *BackupBackingImage {
+	if bbi == nil {
+		return nil
+	}
+
+	backupBackingImage := &BackupBackingImage{
+		Resource: client.Resource{
+			Id:    bbi.Name,
+			Type:  "backupBackingImage",
+			Links: map[string]string{},
+		},
+		Name:              bbi.Name,
+		State:             bbi.Status.State,
+		Progress:          bbi.Status.Progress,
+		Error:             bbi.Status.Error,
+		URL:               bbi.Status.URL,
+		Created:           bbi.Status.BackupCreatedAt,
+		Size:              bbi.Status.Size,
+		Labels:            bbi.Status.Labels,
+		Messages:          bbi.Status.Messages,
+		CompressionMethod: string(bbi.Status.CompressionMethod),
+	}
+
+	backupBackingImage.Actions = map[string]string{
+		"backupBackingImageRestore": apiContext.UrlBuilder.ActionLink(backupBackingImage.Resource, "backupBackingImageRestore"),
+	}
+
+	return backupBackingImage
+}
+
+func toBackupBackingImageCollection(bbis []*longhorn.BackupBackingImage, apiContext *api.ApiContext) *client.GenericCollection {
+	data := []interface{}{}
+	for _, bbi := range bbis {
+		data = append(data, toBackupBackingImageResource(bbi, apiContext))
+	}
+
+	return &client.GenericCollection{Data: data, Collection: client.Collection{ResourceType: "backupBackingImage"}}
+}
+
 func toBackupResource(b *longhorn.Backup) *Backup {
 	if b == nil {
 		return nil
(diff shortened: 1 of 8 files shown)


ChanYiLin:
Test Steps:

1. Create BackingImage
```
name: parrot
downloadURL: https://longhorn-backing-image.s3-us-west-1.amazonaws.com/parrot.raw 
```
Note: keep the `checksum`

2. Test Create BackupBackingImage
```
curl -X POST http://localhost:8080/v1/backingimages/parrot?action=backupBackingImageCreate
```

3. Test Get BackupBackingImage
```
curl -X GET http://localhost:8080/v1/backupbackingimages/parrot
{"actions":{"backupBackingImageRestore":"http://localhost:8080/v1/backupbackingimages/parrot?action=backupBackingImageRestore"},"compressionMethod":"lz4","created":"2024-03-01T05:28:35Z","error":"","id":"parrot","labels":null,"links":{"self":"http://localhost:8080/v1/backupbackingimages/parrot"},"messages":null,"name":"parrot","progress":100,"size":33554432,"state":"Completed","type":"backupBackingImage","url":"nfs://longhorn-test-nfs-svc.default:/opt/backupstore?backingImage=parrot"}
```

4. Delete original BackingImage
```
$ kubectl delete lhbi/parrot -n longhorn-system
```

5. Restore BackingImage
```
curl -X POST http://localhost:8080/v1/backupbackingimages/parrot?action=backupBackingImageRestore
```

6. check the checksum is the same as original

api/model.go
```diff
@@ -1860,8 +1924,9 @@ func toBackingImageResource(bi *longhorn.BackingImage, apiContext *api.ApiContex
 		DeletionTimestamp: deletionTimestamp,
 	}
 	res.Actions = map[string]string{
-		"backingImageCleanup": apiContext.UrlBuilder.ActionLink(res.Resource, "backingImageCleanup"),
-		BackingImageUpload:    apiContext.UrlBuilder.ActionLink(res.Resource, BackingImageUpload),
+		"backingImageCleanup":      apiContext.UrlBuilder.ActionLink(res.Resource, "backingImageCleanup"),
+		BackingImageUpload:         apiContext.UrlBuilder.ActionLink(res.Resource, BackingImageUpload),
+		"backupBackingImageCreate": apiContext.UrlBuilder.ActionLink(res.Resource, "backupBackingImageCreate"),
```
innobead, line 1929: `BackingImageUpload` is `upload` which is different from others. Is this intended? It seems the prefix already exists, should we just name `cleanup`, `upload` and `create` instead?

api/router.go
```diff
@@ -162,14 +162,24 @@ func NewRouter(s *Server) *mux.Router {
 	r.Methods("POST").Path("/v1/backingimages").Handler(f(schemas, s.BackingImageCreate))
 	r.Methods("DELETE").Path("/v1/backingimages/{name}").Handler(f(schemas, s.BackingImageDelete))
 	backingImageActions := map[string]func(http.ResponseWriter, *http.Request) error{
-		"backingImageCleanup": s.BackingImageCleanup,
-
-		BackingImageUpload: s.fwd.Handler(s.fwd.HandleProxyRequestForBackingImageUpload, UploadParametersForBackingImage(s.m), s.BackingImageGet),
+		"backingImageCleanup":      s.BackingImageCleanup,
+		BackingImageUpload:         s.fwd.Handler(s.fwd.HandleProxyRequestForBackingImageUpload, UploadParametersForBackingImage(s.m), s.BackingImageGet),
+		"backupBackingImageCreate": s.BackupBackingImageCreate,
```
innobead, line 167: It's about creation, suggest using `POST` for consistency with functions.

manager/backupbackingimage.go
```diff
@@ -0,0 +1,70 @@
...
+import (
+	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
+	"github.com/longhorn/longhorn-manager/util"
+	"github.com/pkg/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func (m *VolumeManager) ListBackupBackingImagesSorted() ([]*longhorn.BackupBackingImage, error) {
+	backupBackingImageMap, err := m.ds.ListBackupBackingImages()
```
innobead, line 12: `ListBackupBackingImagesRO`

manager/backupbackingimage.go
```diff
@@ -0,0 +1,70 @@
...
+func (m *VolumeManager) DeleteBackupBackingImage(name string) error {
+	return m.ds.DeleteBackupBackingImage(name)
+}
+
+func (m *VolumeManager) RestoreBackupBackingImage(name string) error {
+	return m.restoreBackingImage(name)
+}
+
+func (m *VolumeManager) CreateBackupBackingImage(name string) error {
+	_, err := m.ds.GetBackingImage(name)
```
innobead, line 43: `GetBackingImageRO`

innobead (changes requested):

api/model.go
ChanYiLin: I can fix that and notice UI to adjust as well

api/router.go
ChanYiLin: For these three actions, they are `"POST"`, it is defined the next line.

```
	backingImageActions := map[string]func(http.ResponseWriter, *http.Request) error{
		"backingImageCleanup":      s.BackingImageCleanup,
		BackingImageUpload:         s.fwd.Handler(s.fwd.HandleProxyRequestForBackingImageUpload, UploadParametersForBackingImage(s.m), s.BackingImageGet),
		"backupBackingImageCreate": s.BackupBackingImageCreate,
	}
	for name, action := range backingImageActions {
		r.Methods("POST").Path("/v1/backingimages/{name}").Queries("action", name).Handler(f(schemas, action))
	}
```

manager/backupbackingimage.go
ChanYiLin: I need the map, because we need to sort the list.
Using RO will only return list and I still need to deep copy the item when sorting.

The implementation is the same as other APIs
https://github.com/longhorn/longhorn-manager/blob/8cdf6820620161a0e2b1e71f2994a3c4f763917c/manager/volume.go#L66-L81

```
func (m *VolumeManager) ListSorted() ([]*longhorn.Volume, error) {
	volumeMap, err := m.List()
	if err != nil {
		return []*longhorn.Volume{}, err
	}

	volumes := make([]*longhorn.Volume, len(volumeMap))
	volumeNames, err := util.SortKeys(volumeMap)
	if err != nil {
		return []*longhorn.Volume{}, err
	}
	for i, volumeName := range volumeNames {
		volumes[i] = volumeMap[volumeName]
	}
	return volumes, nil
}
```

manager/backupbackingimage.go
ChanYiLin: Fixed, thanks!

api/model.go
ChanYiLin: by the way, this is under backingimage
but for create, it is for creating BackupBackingImage, I am not sure if it will confuse others if we only use "create"

ChanYiLin: I think adding prefix is more appropriate.
We put all the resource and their actions in the same file `model.go`, for each action they also have Input
Adding prefix can distinguish them

e.g.

```
type BackingImageCleanupInput struct {
	Disks []string `json:"disks"`
}

type PVCreateInput struct {
	PVName string `json:"pvName"`
	FSType string `json:"fsType"`

	SecretName      string `json:"secretName"`
	SecretNamespace string `json:"secretNamespace"`

	StorageClassName string `json:"storageClassName"`
}

type PVCCreateInput struct {
	Namespace string `json:"namespace"`
	PVCName   string `json:"pvcName"`
}
```

ChanYiLin: > `BackingImageUpload` is `upload` which is different from others. Is this intended? It seems the prefix already exists, should we just name `cleanup`, `upload` and `create` instead?

I think we just forgot to make it consistent to have prefix when implementing
https://github.com/ChanYiLin/longhorn-manager/commit/f271ef470f706e58cbd0d30948e1f1189b028d8d#diff-cb740c237beef9090d8b3a15ca8b76aac83a3af2d52deadc382b4309916a715cL141

I can fix that, but need UI/backing-image-manager changes or it will break the API
And also need to check harvester side.

ChanYiLin: I see why do we use "upload" without prefix

1. in `model.go` we should use the prefix because we put all the resources in the same places, so we have "BackingImageCleanup"
2. When implementing "upload", since it is a proxy from `longhorn-manager` -> `backing-image-manager/datasource` -> `backing-image-manager/sync-file`

https://github.com/longhorn/backing-image-manager/blob/5c9b6e5c6f3ad1d24e84a344462d6a434dd48f98/pkg/datasource/router.go#L8-L19

https://github.com/longhorn/backing-image-manager/blob/5c9b6e5c6f3ad1d24e84a344462d6a434dd48f98/pkg/sync/router.go#L22-L27

3. we proxy the "action" which is "upload" all the way down to the `sync-file server`. Since we added the `proxy` after `datasource` and `sync-file` so we keep using the "upload".

To fix the wording we have to change
- datasource in backing-image-manager, but that make the action name there not consistent to others which have no prefix.
- sync-file in backing-image-manager, but that make the action name there not consistent.
- UI
- Harvester side

ChanYiLin: I prefer not to change it this time,
- "upload" is for proxy, and we remain the same action name all the way down the proxy path
- "BackingImageCleanup", "BackupBackingImageCreate" follow the convention in "model.go" to have prefix, besides, we also use the same name for their handler.
```
	backingImageActions := map[string]func(http.ResponseWriter, *http.Request) error{
		"backingImageCleanup":      s.BackingImageCleanup,
		BackingImageUpload:         s.fwd.Handler(s.fwd.HandleProxyRequestForBackingImageUpload, UploadParametersForBackingImage(s.m), s.BackingImageGet),
		"backupBackingImageCreate": s.BackupBackingImageCreate, 
        }
```

https://github.com/longhorn/longhorn-manager/pull/2658/files#diff-cb740c237beef9090d8b3a15ca8b76aac83a3af2d52deadc382b4309916a715cR165-R167

innobead: Discussed. Go with your current implementation.

api/router.go
innobead: This has been clarified. 

`backupBackupImageCreate` is not for backing image, but rather for backing image backup. Ideally, CRUD operations suggested to be mapped to the corresponding HTTP action; however, this is only for the same domain resource.

innobead (approved):

PR #2659, opened 2024-03-01
fix(csi): recurring event (no Pending workload ...) after kubelet restart

c3y1huang:

Issue longhorn/longhorn#8072

Continue with the NodePublishVolume if all workload pods are running.


csi/node_server.go
```diff
@@ -143,7 +143,7 @@ func (ns *NodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 	}
 
 	podsStatus := ns.collectWorkloadPodsStatus(volume, log)
-	if len(podsStatus[corev1.PodPending]) == 0 {
+	if len(podsStatus[corev1.PodPending]) == 0 && len(podsStatus[corev1.PodRunning]) != len(volume.KubernetesStatus.WorkloadsStatus) {
```
derekbit, line 146: Can you elaborate more on what will happen if an error is returned here?

derekbit (approved):
LGTM

innobead (approved):
LGTM. 

This just means `NodePublishVolume` should be `idempotent`. We should be more casual about this in the future.

innobead:

c3y1huang:
@innobead mergify seems to be pending in `Backport to branch v1.4.x in progress` for a while now.

innobead:
@mergify backport v1.4.x

PR #2660, opened 2024-03-01
fix(backup): manually synchronizing backup volumes.

mantissahz:

Issue # longhorn/longhorn#7982

When users make a requst to get backup volumes information by Backup Volume Get/List APIs, it will also trigger the synchronization for these backup volumes once.

### DIFF
--- a/api/backup.go
+++ b/api/backup.go
@@ -1,6 +1,7 @@
 package api
 
 import (
+	"fmt"
 	"net/http"
 
 	"github.com/gorilla/mux"
@@ -18,7 +19,61 @@ func (s *Server) BackupTargetList(w http.ResponseWriter, req *http.Request) erro
 	if err != nil {
 		return errors.Wrap(err, "failed to list backup targets")
 	}
-	apiContext.Write(toBackupTargetCollection(backupTargets))
+	apiContext.Write(toBackupTargetCollection(backupTargets, apiContext))
+	return nil
+}
+
+func (s *Server) BackupTargetSyncAll(w http.ResponseWriter, req *http.Request) error {
+	var input SyncBackupResource
+
+	apiContext := api.GetApiContext(req)
+	if err := apiContext.Read(&input); err != nil {
+		return err
+	}
+
+	bts, err := s.m.ListBackupTargetsSorted()
+	if err != nil {
+		return errors.Wrap(err, "failed to list backup targets")
+	}
+
+	if input.SyncAllBackupTargets {
+		for _, bt := range bts {
+			if _, err := s.m.SyncBackupTarget(bt); err != nil {
+				logrus.WithError(err).Warnf("Failed to synchronize backup target %v", bt.Name)
+			}
+		}
+	}
+
+	apiContext.Write(toBackupTargetCollection(bts, apiContext))
+	return nil
+}
+
+func (s *Server) BackupTargetSync(w http.ResponseWriter, req *http.Request) error {
+	var input SyncBackupResource
+
+	apiContext := api.GetApiContext(req)
+	if err := apiContext.Read(&input); err != nil {
+		return err
+	}
+
+	backupTargetName := mux.Vars(req)["backupTargetName"]
+	if backupTargetName == "" {
+		return fmt.Errorf("backup target name is required")
+	}
+
+	bt, err := s.m.GetBackupTarget(backupTargetName)
+	if err != nil {
+		return errors.Wrapf(err, "failed to get backup target %v", backupTargetName)
+	}
+
+	if input.SyncBackupTarget {
+		bt, err = s.m.SyncBackupTarget(bt)
+		if err != nil {
+			return errors.Wrapf(err, "failed to synchronize backup target %v", backupTargetName)
+		}
+	}
+
+	apiContext.Write(toBackupTargetResource(bt, apiContext))
 	return nil
 }
 
@@ -50,6 +105,60 @@ func (s *Server) BackupVolumeGet(w http.ResponseWriter, req *http.Request) error
 	if err != nil {
 		return errors.Wrapf(err, "failed to get backup volume '%s'", volName)
 	}
+	apiContext.Write(toBackupVolumeResource(bv, apiContext))
+	return nil
+}
+
+func (s *Server) BackupVolumeSyncAll(w http.ResponseWriter, req *http.Request) error {
+	var input SyncBackupResource
+
+	apiContext := api.GetApiContext(req)
+	if err := apiContext.Read(&input); err != nil {
+		return err
+	}
+
+	bvs, err := s.m.ListBackupVolumesSorted()
+	if err != nil {
+		return errors.Wrap(err, "failed to list backup volumes")
+	}
+
+	if input.SyncAllBackupVolumes {
+		for _, bv := range bvs {
+			if _, err := s.m.SyncBackupVolume(bv); err != nil {
+				logrus.WithError(err).Warnf("Failed to synchronize backup volume %v", bv.Name)
+			}
+		}
+	}
+
+	apiContext.Write(toBackupVolumeCollection(bvs, apiContext))
+	return nil
+}
+
+func (s *Server) SyncBackupVolume(w http.ResponseWriter, req *http.Request) error {
+	var input SyncBackupResource
+
+	apiContext := api.GetApiContext(req)
+	if err := apiContext.Read(&input); err != nil {
+		return err
+	}
+
+	volName := mux.Vars(req)["volName"]
+	if volName == "" {
+		return fmt.Errorf("backup volume name is required")
+	}
+
+	bv, err := s.m.GetBackupVolume(volName)
+	if err != nil {
+		return errors.Wrapf(err, "failed to get backup volume '%s'", volName)
+	}
+
+	if input.SyncBackupVolume {
+		bv, err = s.m.SyncBackupVolume(bv)
+		if err != nil {
+			return errors.Wrapf(err, "failed to synchronize backup volume %v", volName)
+		}
+	}
+
 	apiContext.Write(toBackupVolumeResource(bv, apiContext))
 	return nil
 }
--- a/api/model.go
+++ b/api/model.go
@@ -142,6 +142,14 @@ type BackupVolume struct {
 	StorageClassName     string            `json:"storageClassName"`
 }
 
+// SyncBackupResource is used for the Backup*Sync* actions
+type SyncBackupResource struct {
+	SyncAllBackupTargets bool `json:"syncAllBackupTargets"`
+	SyncBackupTarget     bool `json:"syncBackupTarget"`
+	SyncAllBackupVolumes bool `json:"syncAllBackupVolumes"`
+	SyncBackupVolume     bool `json:"syncBackupVolume"`
+}
+
 type Backup struct {
 	client.Resource
 
@@ -556,6 +564,16 @@ type VolumeRecurringJobInput struct {
 	longhorn.VolumeRecurringJob
 }
 
+type BackupTargetListOutput struct {
+	Data []BackupTarget `json:"data"`
+	Type string         `json:"type"`
+}
+
+type BackupVolumeListOutput struct {
+	Data []BackupVolume `json:"data"`
+	Type string         `json:"type"`
+}
+
 type BackupListOutput struct {
 	Data []Backup `json:"data"`
 	Type string   `json:"type"`
@@ -581,10 +599,10 @@ func NewSchema() *client.Schemas {
 	schemas.AddType("detachInput", DetachInput{})
 	schemas.AddType("snapshotInput", SnapshotInput{})
 	schemas.AddType("snapshotCRInput", SnapshotCRInput{})
-	schemas.AddType("backupTarget", BackupTarget{})
 	schemas.AddType("backup", Backup{})
 	schemas.AddType("backupInput", BackupInput{})
 	schemas.AddType("backupStatus", BackupStatus{})
+	schemas.AddType("syncBackupResource", SyncBackupResource{})
 	schemas.AddType("orphan", Orphan{})
 	schemas.AddType("restoreStatus", RestoreStatus{})
 	schemas.AddType("purgeStatus", PurgeStatus{})
@@ -644,6 +662,7 @@ func NewSchema() *client.Schemas {
 	volumeSchema(schemas.AddType("volume", Volume{}))
 	snapshotSchema(schemas.AddType("snapshot", Snapshot{}))
 	snapshotCRSchema(schemas.AddType("snapshotCR", SnapshotCR{}))
+	backupTargetSchema(schemas.AddType("backupTarget", BackupTarget{}))
 	backupVolumeSchema(schemas.AddType("backupVolume", BackupVolume{}))
 	backupBackingImageSchema(schemas.AddType("backupBackingImage", BackupBackingImage{}))
 	settingSchema(schemas.AddType("setting", Setting{}))
@@ -654,6 +673,8 @@ func NewSchema() *client.Schemas {
 	diskSchema(schemas.AddType("diskUpdateInput", DiskUpdateInput{}))
 	diskInfoSchema(schemas.AddType("diskInfo", DiskInfo{}))
 	kubernetesStatusSchema(schemas.AddType("kubernetesStatus", longhorn.KubernetesStatus{}))
+	backupTargetListOutputSchema(schemas.AddType("backupTargetListOutput", BackupTargetListOutput{}))
+	backupVolumeListOutputSchema(schemas.AddType("backupVolumeListOutput", BackupVolumeListOutput{}))
 	backupListOutputSchema(schemas.AddType("backupListOutput", BackupListOutput{}))
 	snapshotListOutputSchema(schemas.AddType("snapshotListOutput", SnapshotListOutput{}))
 	systemBackupSchema(schemas.AddType("systemBackup", SystemBackup{}))
@@ -814,9 +835,21 @@ func kubernetesStatusSchema(status *client.Schema) {
 	status.ResourceFields["workloadsStatus"] = workloadsStatus
 }
 
+func backupTargetSchema(backupTarget *client.Schema) {
+	backupTarget.CollectionMethods = []string{"GET"}
+	backupTarget.ResourceMethods = []string{"GET", "PUT"}
+
+	backupTarget.ResourceActions = map[string]client.Action{
+		"backupTargetSync": {
+			Input:  "syncBackupResource",
+			Output: "backupTargetListOutput",
+		},
+	}
+}
+
 func backupVolumeSchema(backupVolume *client.Schema) {
 	backupVolume.CollectionMethods = []string{"GET"}
-	backupVolume.ResourceMethods = []string{"GET", "DELETE"}
+	backupVolume.ResourceMethods = []string{"GET", "PUT", "DELETE"}
 	backupVolume.ResourceActions = map[string]client.Action{
 		"backupList": {
 			Output: "backupListOutput",
@@ -828,6 +861,10 @@ func backupVolumeSchema(backupVolume *client.Schema) {
 		"backupDelete": {
 			Input:  "backupInput",
 			Output: "backupVolume",
+		},
+		"backupVolumeSync": {
+			Input:  "syncBackupResource",
+			Output: "backupVolumeListOutput",
 		},
 	}
 }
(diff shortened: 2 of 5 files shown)


derekbit:
When users make a `BackupVolumeGet` or `BackupVolumeList` HTTP call, they can trigger a backup volume refresh.

Is it possible to synchronize backup volumes by simply editing customer resources? Or do we need to a different method that doesn't involve HTTP calls?

mantissahz:
>Is it possible to synchronize backup volumes by simply editing customer resources? Or do we need to a different method that doesn't involve HTTP calls?

Users can modify the field `spec.syncRequestedAt` of the BackupVolume CR  for synchronizing the backup volume.
And the time of the filed `spec.syncRequestedAt` should be after the time of the filed `status.lastSyncedAt`

shuo-wu:
Can we just simply update `backuptarget.spec.syncRequestedAt` instead? Updating the spec for all backup volumes is probably not enough since there may be backup volume CR creation/deletion.

mantissahz:
>Can we just simply update backuptarget.spec.syncRequestedAt instead? Updating the spec for all backup volumes is probably not enough since there may be backup volume CR creation/deletion.

Yes, we can and modified. Thanks.

shuo-wu (approved):
The implementation LGTM.

Currently, the main concern is, will the websocket keep calling BackupVolumeList then Longhorn generates tons of requests to the remote? If users are using public cloud services like S3, it will lead to a higher bill.

innobead (changes requested):
Should we consider just updating `SyncRequestedAt` to trigger the sync? We could have a constant value representing that the sync will be just triggered, and the mutator will update it with a qualified time value instead. WDYT?

In this way, for UI or even manifest, users can trigger the sync immediately with the same experience.

```go
backupTarget.Spec.SyncRequestedAt = "now" // metav1.Time{Time: time.Now().UTC()}
```

P.S. I assume we don't have schema validation for that property.

shuo-wu:
That's a good idea. And you don't need to worry about the schema validation.
k8s-webhook-1

mantissahz:
>backupTarget.Spec.SyncRequestedAt = "now" // metav1.Time{Time: time.Now().UTC()}

I think it is difficult to have a constant value representing to trigger the sync , and the mutator will update it with a qualified time value instead because of https://github.com/longhorn/longhorn-manager/blob/master/webhook/admission/request.go#L43.

derekbit:
Hello @mantissahz Is the PR ready?

mantissahz:
Hi @derekbit,
Yes, it's ready. Thanks

api/router.go
```diff
@@ -120,8 +120,15 @@ func NewRouter(s *Server) *mux.Router {
...
 	r.Methods("GET").Path("/v1/backuptargets").Handler(f(schemas, s.BackupTargetList))
+	backupTargetActions := map[string]func(http.ResponseWriter, *http.Request) error{
+		"syncBackupTarget": s.fwd.Handler(s.fwd.HandleProxyRequestByNodeID, s.fwd.GetHTTPAddressByNodeID(NodeHasDefaultEngineImage(s.m)), s.SyncBackupTarget),
+	}
+	for name, action := range backupTargetActions {
+		r.Methods("POST").Path("/v1/backuptargets/{btName}").Queries("action", name).Handler(f(schemas, action))
+	}
 	r.Methods("GET").Path("/v1/backupvolumes").Handler(f(schemas, s.fwd.Handler(s.fwd.HandleProxyRequestByNodeID, s.fwd.GetHTTPAddressByNodeID(NodeHasDefaultEngineImage(s.m)), s.BackupVolumeList)))
 	r.Methods("GET").Path("/v1/backupvolumes/{volName}").Handler(f(schemas, s.fwd.Handler(s.fwd.HandleProxyRequestByNodeID, s.fwd.GetHTTPAddressByNodeID(NodeHasDefaultEngineImage(s.m)), s.BackupVolumeGet)))
+	r.Methods("PUT").Path("/v1/backupvolumes/{volName}").Handler(f(schemas, s.fwd.Handler(s.fwd.HandleProxyRequestByNodeID, s.fwd.GetHTTPAddressByNodeID(NodeHasDefaultEngineImage(s.m)), s.UpdateBackupVolume)))
```
shuo-wu, line 131: Why do you choose `PUT` rather than adding a new API `syncBackupVolume` here?

shuo-wu (commented):
Currently, will Longhorn UI show all backup volumes for **all backup targets**? If NO, it's better to add an input `BackupTargetName` for API `SyncBackupTarget`.

api/router.go
mantissahz: Choose `PUT` to update a corresponding backup volume to synchronize the backups.

Do you mean we can add a new API to sync all backup volumes simultaneously?

api/backup.go
```diff
@@ -18,7 +18,38 @@ func (s *Server) BackupTargetList(w http.ResponseWriter, req *http.Request) erro
...
+
+func (s *Server) SyncBackupTarget(w http.ResponseWriter, req *http.Request) error {
+	var input SyncBackupTarget
+
+	apiContext := api.GetApiContext(req)
+	if err := apiContext.Read(&input); err != nil {
+		return err
+	}
+
+	btName := mux.Vars(req)["btName"]
```
derekbit, line 33: nit: replace `btName` with `backupTargetName`

derekbit: I assume this PR will apply to multiple backup targets feature.
Should we
1. synchronize all backup targets if `btName` is empty
2. set `btName` to `default` is if `btName` is empty
3. error out if `btName` is empty
Which one is better?

If we choose 2 or 3, can we sync all backup targes?

api/backup.go
```diff
@@ -18,7 +18,38 @@ func (s *Server) BackupTargetList(w http.ResponseWriter, req *http.Request) erro
...
+
+	bt, err := s.m.GetBackupTarget(btName)
+	if err != nil {
+		return errors.Wrap(err, "failed to list backup targets")
```
derekbit, line 37:
```suggestion
		return errors.Wrapf(err, "failed to get backup target %v", btName)
```

manager/engine.go
```diff
@@ -363,6 +378,15 @@ func (m *VolumeManager) GetBackupVolume(volumeName string) (*longhorn.BackupVolu
 	return backupVolume, err
 }
 
+func (m *VolumeManager) SyncBackupVolume(backupVolume *longhorn.BackupVolume) (*longhorn.BackupVolume, error) {
+	now := metav1.Time{Time: time.Now().UTC()}
+	if now.Sub(backupVolume.Spec.SyncRequestedAt.Time).Seconds() < minSyncBackupTargetIntervalSec {
+		return nil, errors.Errorf("cannot synchronize backup volume '%v' in %v seconds", backupVolume.Name, minSyncBackupTargetIntervalSec)
```
derekbit, line 384:
```suggestion
		return nil, errors.Errorf("failed to synchronize backup volume '%v' in %v seconds", backupVolume.Name, minSyncBackupTargetIntervalSec)
```

manager/engine.go
```diff
@@ -18,6 +18,8 @@ import (
 
 const (
 	BackupStatusQueryInterval = 2 * time.Second
+
+	minSyncBackupTargetIntervalSec = 30
```
derekbit, line 22: How to determine value? Suggest adding a comment here.

api/router.go
mantissahz: Separate the APIs with synchronizing all backup volumes and corresponding one.

api/backup.go
mantissahz: Modified to `backupTargetName`.
I took 2 and 3 because python clients, it is not easy to update backup targets/volumes with an empty name.

I added two APIs, one is to synchronize all backup targets and another is to synchronize all backup volumes.

manager/engine.go
mantissahz: Modified. Thanks

api/backup.go
mantissahz: Modified. Thanks.

api/backup.go
```diff
@@ -54,6 +109,61 @@ func (s *Server) BackupVolumeGet(w http.ResponseWriter, req *http.Request) error
...
+
+	bvs, err := s.m.ListBackupVolumesSorted()
+	if err != nil {
+		return errors.Wrap(err, "failed to list backup volumes")
+	}
+
+	if input.SyncBackupVolume {
+		for _, bv := range bvs {
+			if _, err := s.m.SyncBackupVolume(bv); err != nil {
+				return errors.Wrapf(err, "failed to update backup volume '%s'", bv.Name)
```
derekbit, line 128: Shall we error out here or just do best effort?

mantissahz: We could have a warning log if there is an error with synchronizing the backup volume and continue synchronizing the next backup volume.
WDYT?

api/backup.go
```diff
@@ -54,6 +109,61 @@ func (s *Server) BackupVolumeGet(w http.ResponseWriter, req *http.Request) error
...
+		return fmt.Errorf("backup volume name is required")
+	}
+
+	bv, err := s.m.GetBackupVolume(volName)
+	if err != nil {
+		return errors.Wrapf(err, "failed to get backup volume '%s'", volName)
+	}
+
+	if input.SyncBackupVolume {
+
```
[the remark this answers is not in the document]
mantissahz: Removed the redundant line. Thanks

api/backup.go
mantissahz: Modified to leave a warning log instead of returning an error and modified the part of backup target as well. Thanks

api/backup.go
```diff
@@ -18,7 +19,64 @@ func (s *Server) BackupTargetList(w http.ResponseWriter, req *http.Request) erro
...
+	bts, err := s.m.ListBackupTargetsSorted()
+	if err != nil {
+		return errors.Wrap(err, "failed to list backup targets")
+	}
+
+	if input.SyncBackupTarget {
+		for _, bt := range bts {
+			if _, err := s.m.SyncBackupTarget(bt); err != nil {
+				logrus.Warnf("failed to synchronize backup target %v", bt.Name)
```
derekbit, line 42:
```suggestion
				logrus.WithError(err).Warnf("Failed to synchronize backup target %v", bt.Name)
```

api/backup.go
```diff
@@ -18,7 +19,64 @@ func (s *Server) BackupTargetList(w http.ResponseWriter, req *http.Request) erro
...
+
+	bt, err := s.m.GetBackupTarget(backupTargetName)
+	if err != nil {
+		return errors.Wrapf(err, "failed to get backup target %v", backupTargetName)
+	}
+
+	if input.SyncBackupTarget {
+		updatedBackupTarget, err := s.m.SyncBackupTarget(bt)
+		if err != nil {
+			logrus.Warnf("failed to synchronize backup target %v", backupTargetName)
```
derekbit, line 72: For single backuptarget, shall we error out?

api/backup.go
```diff
@@ -54,6 +112,63 @@ func (s *Server) BackupVolumeGet(w http.ResponseWriter, req *http.Request) error
...
+
+	bvs, err := s.m.ListBackupVolumesSorted()
+	if err != nil {
+		return errors.Wrap(err, "failed to list backup volumes")
+	}
+
+	if input.SyncBackupVolume {
+		for _, bv := range bvs {
+			if _, err := s.m.SyncBackupVolume(bv); err != nil {
+				logrus.Warnf("failed to synchronize backup volume %v", bv.Name)
```
derekbit, line 131:
```suggestion
				logrus.WithError(err).Warnf("Failed to synchronize backup volume %v", bv.Name)
```

api/backup.go
```diff
@@ -54,6 +112,63 @@ func (s *Server) BackupVolumeGet(w http.ResponseWriter, req *http.Request) error
...
+
+	bv, err := s.m.GetBackupVolume(volName)
+	if err != nil {
+		return errors.Wrapf(err, "failed to get backup volume '%s'", volName)
+	}
+
+	if input.SyncBackupVolume {
+		updatedBackupVolume, err := s.m.SyncBackupVolume(bv)
+		if err != nil {
+			logrus.Warnf("failed to synchronize backup volume %v", volName)
```
derekbit, line 161: Same question as above

api/backup.go
mantissahz: Modified. Thanks.

api/backup.go
mantissahz: Modified. Thanks.

api/backup.go
mantissahz: Yes, let's return the error.

api/backup.go
mantissahz: Yes, let's return the error.

derekbit (dismissed):
LGTM

innobead:
> > backupTarget.Spec.SyncRequestedAt = "now" // metav1.Time{Time: time.Now().UTC()}
> 
> I think it isn't easy to have a constant value representing to trigger the sync , and the mutator will update it with a qualified time value instead because of [`master`/webhook/admission/request.go#L43](https://github.com/longhorn/longhorn-manager/blob/master/webhook/admission/request.go?rgh-link-date=2024-03-25T07%3A35%3A43Z#L43).

I meant, you can rely on the mutator to update `now` to a qualified time.

api/backup.go
```diff
@@ -18,7 +19,61 @@ func (s *Server) BackupTargetList(w http.ResponseWriter, req *http.Request) erro
 	if err != nil {
 		return errors.Wrap(err, "failed to list backup targets")
 	}
-	apiContext.Write(toBackupTargetCollection(backupTargets))
+	apiContext.Write(toBackupTargetCollection(backupTargets, apiContext))
+	return nil
+}
+
+func (s *Server) BackupTargetUpdateAll(w http.ResponseWriter, req *http.Request) error {
```
innobead, line 26: Rename `SyncAllBackupTargets` if this is what you want to achieve.

api/backup.go
```diff
@@ -18,7 +19,61 @@ func (s *Server) BackupTargetList(w http.ResponseWriter, req *http.Request) erro
...
+	if err := apiContext.Read(&input); err != nil {
+		return err
+	}
+
+	bts, err := s.m.ListBackupTargetsSorted()
+	if err != nil {
+		return errors.Wrap(err, "failed to list backup targets")
+	}
+
+	if input.SyncBackupTarget {
```
innobead, line 39: rename `input.SyncAllBackupTargets`

api/backup.go
```diff
@@ -18,7 +19,61 @@ func (s *Server) BackupTargetList(w http.ResponseWriter, req *http.Request) erro
...
+			}
+		}
+	}
+
+	apiContext.Write(toBackupTargetCollection(bts, apiContext))
+	return nil
+}
+
+func (s *Server) SyncBackupTarget(w http.ResponseWriter, req *http.Request) error {
+	var input SyncBackupTarget
```
innobead, line 52: Why not just use `BackupTarget` model directly? just want the `name`?

api/backup.go
```diff
@@ -54,6 +109,60 @@ func (s *Server) BackupVolumeGet(w http.ResponseWriter, req *http.Request) error
 	return nil
 }
 
+func (s *Server) BackupVolumeUpdateAll(w http.ResponseWriter, req *http.Request) error {
```
innobead, line 112: Rename `SyncAllBackupVolumes`

api/backup.go
```diff
@@ -54,6 +109,60 @@ func (s *Server) BackupVolumeGet(w http.ResponseWriter, req *http.Request) error
...
+	if err := apiContext.Read(&input); err != nil {
+		return err
+	}
+
+	bvs, err := s.m.ListBackupVolumesSorted()
+	if err != nil {
+		return errors.Wrap(err, "failed to list backup volumes")
+	}
+
+	if input.SyncBackupVolume {
```
innobead, line 125: Rename `input.SyncAllBackupVolumes`

api/backup.go
```diff
@@ -54,6 +109,60 @@ func (s *Server) BackupVolumeGet(w http.ResponseWriter, req *http.Request) error
...
+			}
+		}
+	}
+
+	apiContext.Write(toBackupVolumeCollection(bvs, apiContext))
+	return nil
+}
+
+func (s *Server) SyncBackupVolume(w http.ResponseWriter, req *http.Request) error {
+	var input SyncBackupVolume
```
innobead, line 138: Why not just use `BackupVolume` model to get the volume name instead of creating a new model. The functions related to backup volume should be able to reuse the same request model?

api/model.go
```diff
@@ -124,6 +124,11 @@ type SnapshotCR struct {
 type BackupTarget struct {
 	client.Resource
 	engineapi.BackupTarget
+	SyncBackupTarget bool `json:"syncBackupTarget"`
```
innobead, line 127: This is more like a workaround to hijack to the `BackupTarget` resource. Or you want to have a Sync resource for this purpose instead?

for example:
```go
type SyncBackupResource struct {
    SyncAllBackupTargets bool
    SyncAllBackupVolumes bool
}
```

api/model.go
```diff
@@ -124,6 +124,11 @@ type SnapshotCR struct {
...
+}
+
+type SyncBackupTarget struct {
+	SyncBackupTarget bool `json:"syncBackupTarget"`
```
innobead, line 131: It's weird to have the same name as the struct, or we already did this pattern in the current code?

manager/engine.go
```diff
@@ -328,6 +331,19 @@ func (m *VolumeManager) ListBackupTargetsSorted() ([]*longhorn.BackupTarget, err
...
+func (m *VolumeManager) GetBackupTarget(backupTargetName string) (*longhorn.BackupTarget, error) {
+	return m.ds.GetBackupTarget(backupTargetName)
+}
+
+func (m *VolumeManager) SyncBackupTarget(backupTarget *longhorn.BackupTarget) (*longhorn.BackupTarget, error) {
+	now := metav1.Time{Time: time.Now().UTC()}
+	if now.Sub(backupTarget.Spec.SyncRequestedAt.Time).Seconds() < minSyncBackupTargetIntervalSec {
+		return nil, errors.Errorf("cannot synchronize backup target '%v' in %v seconds", backupTarget.Name, minSyncBackupTargetIntervalSec)
+	}
+	backupTarget.Spec.SyncRequestedAt = metav1.Time{Time: time.Now().UTC()}
```
innobead, line 343: `backupTarget.Spec.SyncRequestedAt = now`?

manager/engine.go
```diff
@@ -363,6 +379,15 @@ func (m *VolumeManager) GetBackupVolume(volumeName string) (*longhorn.BackupVolu
 	return backupVolume, err
 }
 
+func (m *VolumeManager) SyncBackupVolume(backupVolume *longhorn.BackupVolume) (*longhorn.BackupVolume, error) {
+	now := metav1.Time{Time: time.Now().UTC()}
+	if now.Sub(backupVolume.Spec.SyncRequestedAt.Time).Seconds() < minSyncBackupTargetIntervalSec {
+		return nil, errors.Errorf("failed to synchronize backup volume '%v' in %v seconds", backupVolume.Name, minSyncBackupTargetIntervalSec)
+	}
+	backupVolume.Spec.SyncRequestedAt = metav1.Time{Time: time.Now().UTC()}
```
innobead, line 387: `= now`?

innobead (changes requested):
In general, LGTM. however, still have some feedback required to resolve.

P.S. In the current implementation, it's specific to REST API. How do users trigger it by updating the CR? need to exactly update the right time in the syncat properties on their own? I still recommend a way that can benefit both interfaces (UI and CLI kubectl)

innobead:
> In general, LGTM. however, still have some feedback required to resolve.
> 
> P.S. In the current implementation, it's specific for REST API. How do users trigger it by updating the CR? need to exactly update the right time in the syncat field of backup target or backup volume on their own? I still recommend a way that can benefit both interfaces (UI and CLI kubectl)

After a quick discussion with @mantissahz , we added type validation for the sync_at field, so it's infeasible to use a non-time value to trigger the sync.  so for non-UI users, we can create a CLI command for it. James will create a ticket for the following improvement. 

[/chart/templates/crds.yaml#L961-L965](https://github.com/longhorn/longhorn/blob/9ef72a18bf39317d94e8636c4fadc12b4edaa433/chart/templates/crds.yaml#L961-L965)
```yaml
              syncRequestedAt:
                description: The time to request run sync the remote backup target.
                format: date-time
                nullable: true
                type: string
```

api/backup.go
mantissahz: Used `BackupTargetSyncAll` to keep the same prefix for group methods in the `api` level.
Thanks.

api/backup.go
mantissahz: Renamed. Thanks

api/backup.go
mantissahz: Moved to struct `SyncBackupResource`.
Thanks.

api/backup.go
mantissahz: Renamed. Thanks

api/model.go
mantissahz: Used the struct `SyncBackupResource` instead. Thanks.

manager/engine.go
mantissahz: https://github.com/longhorn/longhorn-manager/pull/2660#issuecomment-2098143628

manager/engine.go
mantissahz: https://github.com/longhorn/longhorn-manager/pull/2660#issuecomment-2098143628

api/backup.go
mantissahz: Renamed. Thanks.

api/model.go
mantissahz: Moved to struct SyncBackupResource.
Thanks.

derekbit:
@mantissahz Ready to review?

mantissahz:
>@mantissahz Ready to review?

@derekbit, Yes, it is ready.

api/backup.go
mantissahz: Moved to struct SyncBackupResource.
Thanks.

derekbit (approved):

mantissahz:
@Mergifyio backport v1.6.x

PR #2675, opened 2024-03-06
Allow maxSnapshotSize to change from 0 to 250 while upgrading

ejweber:

longhorn/longhorn#7833

@roger-ryao discovered that https://github.com/longhorn/longhorn-manager/pull/2564 was ineffective when upgrading from `v1.5.x` to `v1.6.x+`. In fact, the inclusion of https://github.com/longhorn/longhorn/issues/6563 in `v1.6.x` makes it impossible to upgrade Longhorn from `v1.5.x` to `v1.6.x+` while a volume is migrating at all. This is something I didn't notice in my original testing. Thanks @roger-ryao!

This PR:

1. Allows the `volume.spec.snapshotMaxCount` to be changed on a migrating volume ONLY if is from 0 (coming from `v1.5.x`) to `250` (as the upgrade logic does).
2. Waits to propagate that change to engines and replicas until migration is complete.

I updated the test steps at https://github.com/longhorn/longhorn-manager/pull/2558#issuecomment-1932296348 to account for the change.

### DIFF
--- a/webhook/resources/volume/validator.go
+++ b/webhook/resources/volume/validator.go
@@ -5,6 +5,7 @@ import (
 	"strconv"
 
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 
 	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -339,8 +340,9 @@ func (v *volumeValidator) Update(request *admission.Request, oldObj runtime.Obje
 		return werror.NewInvalidError(err.Error(), "spec.snapshotMaxSize")
 	}
 
-	if (oldVolume.Spec.SnapshotMaxCount != newVolume.Spec.SnapshotMaxCount) || (oldVolume.Spec.SnapshotMaxSize != newVolume.Spec.SnapshotMaxSize) {
-		if err := v.validateUpdatingSnapshotMaxCountAndSize(newVolume); err != nil {
+	if (oldVolume.Spec.SnapshotMaxCount != newVolume.Spec.SnapshotMaxCount) ||
+		(oldVolume.Spec.SnapshotMaxSize != newVolume.Spec.SnapshotMaxSize) {
+		if err := v.validateUpdatingSnapshotMaxCountAndSize(oldVolume, newVolume); err != nil {
 			return err
 		}
 	}
@@ -496,7 +498,7 @@ func validateSnapshotMaxSize(size, snapshotMaxSize int64) error {
 	return nil
 }
 
-func (v *volumeValidator) validateUpdatingSnapshotMaxCountAndSize(newVolume *longhorn.Volume) error {
+func (v *volumeValidator) validateUpdatingSnapshotMaxCountAndSize(oldVolume, newVolume *longhorn.Volume) error {
 	var (
 		currentSnapshotCount     int
 		currentTotalSnapshotSize int64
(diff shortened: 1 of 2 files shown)


ejweber:
@mergify backport v1.6.x

FrankYang0529 (approved):
LGTM. Thanks @ejweber for fixing this. 👍

webhook/resources/volume/validator.go
```diff
@@ -496,39 +498,51 @@
...
 		return nil
 	}
 
+	// It is dangerous to update snapshotMaxCount and snapshotMaxSize while  migrating. However, when upgrading to a
+	// Longhorn version that includes these fields from one that does not, we automatically set snapshotMaxCount = 250,
+	// and we do not want a validation failure here to stop the upgrade. We accept the change, but do not propagate it
+	// to engines or replicas until the migration is complete.
+	if len(engines) >= 2 {
+		if oldVolume.Spec.SnapshotMaxCount == 0 &&
+			newVolume.Spec.SnapshotMaxCount == types.MaxSnapshotNum &&
```
innobead, line 520: Could the default max snapshot count differ from the default value `types.MaxSnapshotNum` if users set a different value during the upgrade?

[/webhook/resources/volume/mutator.go#L256-L264](https://github.com/longhorn/longhorn-manager/blob/5d12108e27cb4e5b7ef892650d985c2d50fe3c2b/webhook/resources/volume/mutator.go#L256-L264)
```go
	if volume.Spec.SnapshotMaxCount == 0 {
		snapshotMaxCount, err := v.getSnapshotMaxCount()
		if err != nil {
			err = errors.Wrap(err, "BUG: cannot get valid number for setting snapshot max count")
			return nil, werror.NewInvalidError(err.Error(), "")
		}
		logrus.Infof("Use the default snapshot max count %v", snapshotMaxCount)
		patchOps = append(patchOps, fmt.Sprintf(`{"op": "replace", "path": "/spec/snapshotMaxCount", "value": %v}`, snapshotMaxCount))
	}
```

ejweber: The upgrade logic explicitly sets `volume.Spec.SnapshotMaxCount == 250` (`types.MaxSnapshotNum`). So, during the upgrade, this value cannot be different.

https://github.com/longhorn/longhorn-manager/blob/0ef33e907c70142760196caa6f551aa8b11d42ed/upgrade/v15xto160/upgrade.go#L89

After the upgrade, newly created volumes may have `volume.Spec.SnapshotMaxCount == 0`. Then the mutator code you linked will take effect. (`volume.Spec.SnapshotMaxCount` is NOT affected by the `snapshot-max-count` setting if `volume.Spec.SnapshotMaxCount != 0`).

I tested this by setting the following in the Helm chart and upgrading from `v1.5.x` to `v1.6.x + my changes`.

```
  # -- Maximum snapshot count for a volume. The value should be between 2 to 250
  snapshotMaxCount: 120
```

I did not observe a crash loop in `longhorn-manager`. I did observe the expected log.

```
[longhorn-manager-2kznf] time="2024-03-07T16:22:18Z" level=debug msg="Allowing snapshotMaxCount of a migrating volume to change from 0 to 250 during upgrade" func="volume.(*volumeValidator).validateUpdatingSnapshotMaxCountAndSize" file="validator.go:522" volumeName=pvc-4f6b7fb6-5489-458a-a275-2d73f3d03989
```

The setting value did change, but, as expected, the volume field remained how it was set by the upgrade logic.

```
eweber@laptop:~/longhorn> kl get setting | grep snapshot-max-count
snapshot-max-count                                                120                                           15m

eweber@laptop:~/longhorn> kl get volume -oyaml | grep snapshotMaxCount
    snapshotMaxCount: 250
```

innobead: Well explained!

PR #2686, opened 2024-03-09
fix(csi): add node info to failed mount logging.

james-munson:

Issue # https://github.com/longhorn/longhorn/issues/7931

Log with kernel release and OS info when an NFS mount fails, to aid in troubleshooting.

### DIFF
--- a/csi/node_server.go
+++ b/csi/node_server.go
@@ -24,6 +24,8 @@ import (
 	"github.com/longhorn/longhorn-manager/csi/crypto"
 	"github.com/longhorn/longhorn-manager/datastore"
 
+	lhns "github.com/longhorn/go-common-libs/ns"
+
 	longhornclient "github.com/longhorn/longhorn-manager/client"
 	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
 )
@@ -280,6 +282,16 @@ func (ns *NodeServer) nodeStageSharedVolume(volumeID, shareEndpoint, targetPath 
 			if err == nil {
 				return nil
 			}
+			// Still failed.  Log with mounting node and kernel version for possible troubleshooting.  Don't step on actual mount error.
+			kernelRelease, err1 := lhns.GetKernelRelease()
+			if err1 != nil {
+				kernelRelease = err1.Error()
+			}
+			osDistro, err2 := lhns.GetOSDistro()
+			if err2 != nil {
+				osDistro = err2.Error()
+			}
+			log.WithError(err).Warnf("Failed to mount volume %v on node %s with kernel release %s, os distro %s", volumeID, ns.nodeID, kernelRelease, osDistro)
 		}
 		return status.Error(codes.Internal, err.Error())
 	}
--- a/csi/deployment.go
+++ b/csi/deployment.go
@@ -414,7 +414,7 @@ func NewPluginDeployment(namespace, serviceAccount, nodeDriverRegistrarImage, li
 								},
 								{
 									Name:             "host",
-									MountPath:        "/rootfs", // path is required for namespaced mounter
+									MountPath:        "/host",
 									MountPropagation: &MountPropagationBidirectional,
 								},
 								{


csi/util.go
```diff
@@ -460,3 +461,28 @@ func requiresSharedAccess(vol *longhornclient.Volume, cap *csi.VolumeCapability)
...
+// We use this utility function because csi-plugin has no datastore object to ask about kubernetes node,
+// and the go-common-lib/ns functions won't work because the /proc mount path is different.
+func getOSInfo() string {
+	utsname := &syscall.Utsname{}
+	if err := syscall.Uname(utsname); err != nil {
+		logrus.WithError(err).Warn("Failed making syscall Uname.")
+		return "<unknown OS>"
+	}
+
+	extract := func(field [65]int8) string {
```
derekbit, line 474: Why 65 here?

james-munson: From the golang syscall package, the Utsname struct is
```go
type Utsname struct {
	Sysname    [65][int8](https://pkg.go.dev/builtin#int8)
	Nodename   [65][int8](https://pkg.go.dev/builtin#int8)
	Release    [65][int8](https://pkg.go.dev/builtin#int8)
	Version    [65][int8](https://pkg.go.dev/builtin#int8)
	Machine    [65][int8](https://pkg.go.dev/builtin#int8)
	Domainname [65][int8](https://pkg.go.dev/builtin#int8)
```

innobead (changes requested):
The log looks good to me, but we already have similar util functions in go-common-libs. Please use them directly.

[/sys/sys.go#L18-L55](https://github.com/longhorn/go-common-libs/blob/0f980544d04d95afc3325d320562198f7a95d5ab/sys/sys.go#L18-L55)
```go
func GetKernelRelease() (string, error) {
	utsname := &syscall.Utsname{}
	if err := syscall.Uname(utsname); err != nil {
		logrus.WithError(err).Warn("Failed to get kernel release")
		return "", err
	}

	// Extract the kernel release from the Utsname structure
	release := make([]byte, 0, len(utsname.Release))
	for _, b := range utsname.Release {
		if b == 0x00 {
			logrus.Trace("Found end of kernel release string [0x00]")
			break
		}
		release = append(release, byte(b))
	}
	return string(release), nil
}

// GetOSDistro reads the /etc/os-release file and returns the ID field.
func GetOSDistro(osReleaseContent string) (string, error) {
	var err error
	defer func() {
		err = errors.Wrapf(err, "failed to get host OS distro")
	}()

	lines := strings.Split(osReleaseContent, "\n")
	for _, line := range lines {
		if strings.HasPrefix(line, "ID=") {
			id := strings.TrimPrefix(line, "ID=")
			id = strings.Trim(id, `"`)
			logrus.Tracef("Found OS distro: %v", id)
			return id, nil
		}
	}

	return "", errors.Errorf("failed to find ID field in %v", types.OsReleaseFilePath)
}

```

james-munson:
Right, the first draft of this PR used `sys.GetKernelRelease()` from `go-common-libs` for the kernel release.
But to get the OS distro was a little trickier.  `ns.GetOSDistro()` does an `nsenter` using the path `/host/proc/ns/1/mnt` which exists in a "normal" pod, but not in the longhorn-csi-plugin.  There the right path would be `/rootfs/proc/ns/1/mnt`.  But `sys.GetOSDistro(osReleaseContent string)` is just a utiilty method for `ns.GetOsDistro()` which feeds it the contents of the `/etc/os-release` file to parse.

I figured I would just use the result from `uname -v` which would be something like
```
 #74-Ubuntu SMP Wed Feb 22 14:14:39 UTC 2023.
```
and so I borrowed the parsing code from sys.GetKernelRelease and did the equivalent of `uname -rv`

However, I can revert to using `sys.GetKernelRelease()` directly and then read the contents of the local `/etc/os-release` without doing the namespace stuff, and use that to call `sys.GetOSDistro()`.  Should be more like what is expected.

james-munson:
OK, so revised to use the go-common-libs/sys utilities.  I had to make a local copy of RunFunc to use the CSI plugin's different `/proc` location, and some slightly modified flavors of the common-libs/ns utilities.

james-munson:
One other option:  Mount the host's `/` on `/host` for the CSI deployment(s).  The path `/rootfs` is set in this code:
https://github.com/longhorn/longhorn-manager/blame/103bcaaae5e080dc93af66bf0f8355c31d2f1b0e/csi/deployment.go#L417
and I'm not sure whether the comment is asserting that this specific path is required.  It would be handy if it could follow the same pattern as other host root mounts.

james-munson:
That's right.  Manual and regression testing are happy with the change to mount host's `/` on `/host`.  I have pushed up the change.

csi/node_server.go
```diff
@@ -280,6 +282,16 @@ func (ns *NodeServer) nodeStageSharedVolume(volumeID, shareEndpoint, targetPath
...
+			// Still failed.  Log with mounting node and kernel version for possible troubleshooting.  Don't step on actual mount error.
+			kernelRelease, err1 := lhns.GetKernelRelease()
+			if err1 != nil {
+				kernelRelease = err1.Error()
+			}
+			osDistro, err2 := lhns.GetOSDistro()
+			if err2 != nil {
+				osDistro = err2.Error()
+			}
+			log.WithError(err).Warnf("Failed to mount volume %v on node %s with kernel release %s, os distro %s.", volumeID, ns.nodeID, kernelRelease, osDistro)
```
ejweber, line 294: NIT: I guess it isn't called out in https://github.com/longhorn/longhorn/wiki/coding-convention#wording, but our log messages virtually never end with punctuation.

ejweber (approved):
LGTM except for an inconsequential nit.

I like this a lot better than the previous implementation. It's good to standardize our mounting conventions.

csi/node_server.go
ejweber: I ended up just changing https://github.com/longhorn/longhorn/wiki/Coding-Convention#wording to reflect my perception of our convention. Feel free to change it back if you disagree.

james-munson: Ok, thanks for the heads-up.  I pushed it, and that cleared your approval, but thanks for taking a look.

ejweber (approved):

innobead (approved):

innobead:

PR #2688, opened 2024-03-10
setting: set v2-data-engine-log-flags to "all" if not assigned

derekbit:

Issue Longhorn/longhorn#7939

Set v2-data-engine-log-flags to "all" (enable all component logs) if it is not assigned and v2-data-engine-log-level is "Debug".


innobead (commented):
Why is `SettingNameV2DataEngineLogLevel` not configured in instance-manager command options?

derekbit:
> Why is `SettingNameV2DataEngineLogLevel` not configured in instance-manager command options?

This is tricky. spdk_tgt doesn't support log level option, so the log level is updated in instance manager controller ([codes](https://github.com/longhorn/longhorn-manager/blob/master/controller/instance_manager_controller.go#L460-L486)).

innobead:
> This is tricky. spdk_tgt doesn't support log level option, so the log level is updated in instance manager controller ([codes](https://github.com/longhorn/longhorn-manager/blob/master/controller/instance_manager_controller.go?rgh-link-date=2024-03-11T12%3A47%3A20Z#L460-L486)).

Do you mean the log level is only able to be set at runtime after `spdk_tgt` is started? If yes, suggest adding a doc comment to the code change in this PR. 

[/vendor/github.com/longhorn/go-spdk-helper/pkg/spdk/client/basic.go#L993-L1019](https://github.com/longhorn/longhorn-spdk-engine/blob/8ac41e98550678b0654d5fc70ace454cb020c57e/vendor/github.com/longhorn/go-spdk-helper/pkg/spdk/client/basic.go#L993-L1019)
```go
// LogSetLevel sets the log level.
//
// "level": Required. Supported values are "disabled", "error", "warn", "notice", "info", "debug". Default is "notice".
func (c *Client) LogSetLevel(level string) (result bool, err error) {
	req := spdktypes.LogSetLevelRequest{
		Level: level,
	}

	cmdOutput, err := c.jsonCli.SendCommand("log_set_level", req)
	if err != nil {
		return false, err
	}

	return result, json.Unmarshal(cmdOutput, &result)
}

// LogGetLevel gets the log level.
func (c *Client) LogGetLevel() (string, error) {
	req := spdktypes.LogGetLevelRequest{}

	level, err := c.jsonCli.SendCommand("log_get_level", req)
	if err != nil {
		return "", err
	}

	return strings.Trim(string(level), "\"\n"), nil
}
```

derekbit:
Added. Thanks.

PR #2689, opened 2024-03-11
chore(vendor): update go-common-libs/go-iscsi-helper and resolve breaking changes

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/7960

Update dependencis and resolve the breaking changes.
- go-iscsi-helper
- go-common-libs

The PR is mainly based on this PR https://github.com/longhorn/longhorn-manager/pull/2103/files
Talos support is not backported to v1.5.x. But we did some refactors on the common libs at the same time
So those refactors were not backported as well.

I am thinking maybe in the future, refactors can be in separated PR so we can backport those refactors immediately.


ChanYiLin:
Note:
- This PR depends on `longhron-engine`: `longhorn-engine` in the vendor also needs to be refactored and updated here, or the image build will still fail.

ChanYiLin:

Fixed.

innobead:
> I am thinking maybe in the future, refactors can be in separated PR so we can backport those refactors immediately.

Sounds good.

PR #2691, opened 2024-03-12
fix(metrics): longhorn_volume_robustness not collected for detached volume

c3y1huang:

Issue longhorn/longhorn#8139

Do not return early for non-engine-related metric collections.


c3y1huang:
test result: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/6619/

innobead (approved):

innobead:
@mergify backport v1.6.x

PR #2701, opened 2024-03-18
Recognize pods that use Longhorn generic ephemeral volumes

ejweber:

Fixes the root cause of longhorn/longhorn#8198.

https://github.com/longhorn/longhorn-manager/pull/2494 only allowed NodePublishVolume calls to proceed if a Longhorn volume's `status.kubernetesStatus.workloadsStatus` contained a pending workload. However, Longhorn's Kubernetes PV controller does not recognize workloads that use generic ephemeral volumes. As a result, NodePublishVolume calls fail for generic ephemeral volumes. https://github.com/longhorn/longhorn-manager/pull/2659 happened to mitigate this behavior, but the root cause remains.

This PR modifies Longhorn's Kubernetes PV controller so that it can detect workloads that use generic ephemeral volumes.

### DIFF
--- a/controller/kubernetes_pv_controller.go
+++ b/controller/kubernetes_pv_controller.go
@@ -434,6 +434,12 @@ func (kc *KubernetesPVController) getAssociatedPods(ks *longhorn.KubernetesStatu
 			if v.PersistentVolumeClaim != nil && v.PersistentVolumeClaim.ClaimName == ks.PVCName {
 				return true
 			}
+			// Generic ephemeral volumes have an Ephemeral section instead of a PersistentVolumeClaim section. The name
+			// of the PersistentVolumeClaim associated with a generic ephemeral volume is deterministic:
+			// https://kubernetes.io/docs/concepts/storage/ephemeral-volumes/#persistentvolumeclaim-naming.
+			if v.Ephemeral != nil && fmt.Sprintf("%s-%s", pod.Name, v.Name) == ks.PVCName {
+				return true
+			}
 		}
 		return false
 	})


ejweber:
## Test steps

Deploy `examples/pod_with_gev.yaml` from https://github.com/longhorn/longhorn/pull/8200/files.

### Broken in v1.6.0 and v1.5.4

Notice that the pod never becomes running and fails with the below error.

```
eweber@laptop:~/longhorn> k describe pod volume-test | grep FailedMount
  Warning  FailedMount             18s (x8 over 82s)  kubelet                  MountVolume.SetUp failed for volume "pvc-fa7faf7b-3540-43e8-9389-ddfa5b7db532" : rpc error: code = Aborted desc = no Pending workload pods for volume pvc-fa7faf7b-3540-43e8-9389-ddfa5b7db532 to be mounted: map[]
```

### Broken in all active branches before this PR

Notice that the volume's `kubernetesStatus` never mentions the associated workload. This is true even on Longhorn versions that can successfully run the volume.

```
eweber@laptop:~/longhorn> kl get volume -oyaml | grep kubernetesStatus -A 7
    kubernetesStatus:
      lastPVCRefAt: ""
      lastPodRefAt: ""
      namespace: default
      pvName: pvc-fa7faf7b-3540-43e8-9389-ddfa5b7db532
      pvStatus: Bound
      pvcName: volume-test-volv
      workloadsStatus: null
```

### Fixed with this PR

After this PR, the pod runs successfully AND the `kubernetesStatus` reflects the running workload.

```
eweber@laptop:~/longhorn> k get pod
NAME          READY   STATUS    RESTARTS   AGE
volume-test   1/1     Running   0          2m44s

eweber@laptop:~/longhorn> kl get volume -oyaml | grep kubernetesStatus -A 11
    kubernetesStatus:
      lastPVCRefAt: ""
      lastPodRefAt: ""
      namespace: default
      pvName: pvc-3b179574-67df-4a17-befa-142886cf5742
      pvStatus: Bound
      pvcName: volume-test-volv
      workloadsStatus:
      - podName: volume-test
        podStatus: Running
        workloadName: ""  # Empty because we are using a bare pod.
        workloadType: ""  # Empty because we are using a bare pod.
```

ejweber:
Core regression tests running at: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/6649/.

james-munson (approved):
Looks reasonable.  Thanks for the detailed explanation.

innobead (approved):

innobead:

innobead:
@mergify backport v1.4.x

c3y1huang:
Thank you for the fix!

PR #2702, opened 2024-03-19
feat(backup): add option validation for backup creation

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/7070

backupstore: If the backup has the option `backupmodea: full`, it will pretend there is no last backup, and backup all the blocks of the current volume.

longhorn-manager: We need to validate the option of the Backup to check if there is any typo in the option.


ChanYiLin:
This is created from the wrong branch

PR #2710, opened 2024-03-26
feat(system-backup): support backup backing image

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/5085

1. Backup all BackingImages after backing up Volumes and their yaml when taking system-backup
    - **`After backing up Volumes`: Because now we already backup BackingImage when backing up Volumes**, In this stage, we only backup those BackingImages not being used.
3. Restore BackingImages before restoring Volumes, so the Volumes with BackingImages can restore successfully.
    - **`Before restoring Volumes`: Because Volume with BackingImage creation will fail if there is no BackingImage in the cluster**
    - If the BackingImage already exists, update the tag `last-skipped-system-restored` and `last-skipped-system-restored-at`
    - If the BackingImage does not exist, restore it by creating `restore` type BackingImage
        - use `Checksum:   restore.Status.Checksum,` to verify the BackingImage (checksum from the yaml system-backup stored)
        -  update the tag `last-system-restored` and `last-system-restored-at`

### DIFF
--- a/controller/system_backup_controller.go
+++ b/controller/system_backup_controller.go
@@ -435,6 +435,18 @@ func (c *SystemBackupController) reconcile(name string, backupTargetClient engin
 		// TODO: handle error check
 		go c.WaitForVolumeBackupToComplete(backups, systemBackup) // nolint: errcheck
 
+	case longhorn.SystemBackupStateBackingImageBackup:
+		backupBackingImages, err := c.BackupBackingImage()
+		if err != nil {
+			c.updateSystemBackupRecord(record,
+				systemBackupRecordTypeError, longhorn.SystemBackupStateError,
+				constant.EventReasonStart, err.Error(),
+			)
+			return nil
+		}
+
+		go c.WaitForBackingImageBackupToComplete(backupBackingImages, systemBackup)
+
 	case longhorn.SystemBackupStateGenerating:
 		go c.GenerateSystemBackup(systemBackup, tempBackupArchivePath, tempBackupDir)
 
@@ -764,12 +776,12 @@ func (c *SystemBackupController) WaitForVolumeBackupToComplete(backups map[strin
 				systemBackupRecordTypeError, longhorn.SystemBackupStateError,
 				constant.EventReasonStart, err.Error(),
 			)
+		} else {
+			c.updateSystemBackupRecord(record,
+				systemBackupRecordTypeNormal, longhorn.SystemBackupStateBackingImageBackup,
+				constant.EventReasonStart, SystemBackupMsgStarting,
+			)
 		}
-
-		c.updateSystemBackupRecord(record,
-			systemBackupRecordTypeNormal, longhorn.SystemBackupStateGenerating,
-			constant.EventReasonStart, SystemBackupMsgStarting,
-		)
 
 		c.handleStatusUpdate(record, systemBackup, existingSystemBackup, err, log)
 	}()
@@ -785,6 +797,7 @@ func (c *SystemBackupController) WaitForVolumeBackupToComplete(backups map[strin
 			backup, err = c.ds.GetBackup(name)
 			if err != nil {
 				if apierrors.IsNotFound(err) {
+					log.Warnf("Backup %v not found when checking the volume backup status in system backup", name)
 					continue
 				}
 				break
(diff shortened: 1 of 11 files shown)


ChanYiLin:
## Manual Test

1. Create a BackingImage and a volume using it
2. Create a system-backup
3. `kubectl get lhbbi, lhb -n longhorn-system` will see the backup of the Volume and the Backingimage
4. Delete the Volume and the BackingImage
5. Restore the system-backup
6. The BackingImage and the Volume will be restored. The BackingImage source type will be `restored` and the checksum will be verified.

c3y1huang:
> ref: [longhorn/longhorn#5085](https://github.com/longhorn/longhorn/issues/5085)
> 
> 1. Backup all BackingImages after backing up Volumes and their yaml when taking system-backup
>    
>    * **`After backing up Volumes`: Because now we already backup BackingImage when backing up Volumes**, In this stage, we only backup those BackingImages not being used.
> 2. Restore BackingImages before restoring Volumes, so the Volumes with BackingImages can restore successfully.
>    
>    * **`Before restoring Volumes`: Because Volume with BackingImage creation will fail if there is no BackingImage in the cluster**
>    * If the BackingImage already exists, update the tag `last-skipped-system-restored` and `last-skipped-system-restored-at`
>    * If the BackingImage does not exist, restore it by creating `restore` type BackingImage
>      
>      * use `Checksum:   restore.Status.Checksum,` to verify the BackingImage (checksum from the yaml system-backup stored)
>      * update the tag `last-system-restored` and `last-system-restored-at`

Could you check if we should also update the [LEP](https://github.com/longhorn/longhorn/blob/master/enhancements/20220913-longhorn-system-backup-restore.md)?

c3y1huang:
> ## Manual Test
> 1. Create a BackingImage and a volume using it
> 2. Create a system-backup
> 3. `kubectl get lhbbi, lhb -n longhorn-system` will see the backup of the Volume and the Backingimage
> 4. Delete the Volume and the BackingImage
> 5. Restore the system-backup
> 6. The BackingImage and the Volume will be restored. The BackingImage source type will be `restored` and the checksum will be verified.

Can this be automated? 
https://github.com/longhorn/longhorn-tests/blob/master/manager/integration/tests/test_system_backup_restore.py

ChanYiLin:
Hi @c3y1huang 
Sure, I will update the LEP and the e2e tests.
Thanks for the reminder.

ChanYiLin:
Hi @c3y1huang 
Here is the update of the lep: https://github.com/longhorn/longhorn/pull/8389
PTAL, thanks!

ChanYiLin:
Hi @c3y1huang 
Here is the e2e test: https://github.com/longhorn/longhorn-tests/pull/1864
Thanks!

ChanYiLin:

Fixed.

controller/system_rollout_controller.go
```diff
@@ -1843,6 +1851,77 @@ func (c *SystemRolloutController) restoreStorageClasses() (err error) {
...
+					},
+				},
+			}
+
+			log.Info(SystemRolloutMsgCreating)
+
+			fnCreate := func(backingImage runtime.Object) (runtime.Object, error) {
+				obj, ok := backingImage.(*longhorn.BackingImage)
+				if !ok {
+					return nil, fmt.Errorf(SystemRolloutErrFailedConvertToObjectFmt, restore.GetObjectKind(), types.LonghornKindVolume)
```
c3y1huang, line 1893:
```suggestion
					return nil, fmt.Errorf(SystemRolloutErrFailedConvertToObjectFmt, restore.GetObjectKind(), types.LonghornKindBackingImage)
```

controller/system_rollout_controller_test.go
```diff
@@ -902,6 +905,57 @@ func (s *TestSuite) TestSystemRollout(c *C) {
 				},
 			},
 		},
+		"system rollout BackingImage exist in cluster": {
+			state:        longhorn.SystemRestoreStateRestoring,
+			isInProgress: true,
+			expectState:  longhorn.SystemRestoreStateCompleted,
+
+			existBackingImages: map[SystemRolloutCRName]*longhorn.BackingImage{
+				SystemRolloutCRName(TestVolumeName): {
```
c3y1huang, line 914:
```suggestion
				SystemRolloutCRName(TestBackingImageName): {
```

c3y1huang (approved):

controller/system_backup_controller.go
```diff
@@ -845,6 +857,112 @@ func (c *SystemBackupController) createVolumeBackup(volume *longhorn.Volume, sys
...
+	defer ticker.Stop()
+
+	for range ticker.C {
+		for name := range backupBackingImages {
+			// Retrieve the latest backup
+			var backupBackingImage *longhorn.BackupBackingImage
+			backupBackingImage, err = c.ds.GetBackupBackingImageRO(name)
+			if err != nil {
+				if apierrors.IsNotFound(err) {
+					continue
```
derekbit, line 913: Should we do `delete(backupBackingImages, name)` if not found?

controller/system_backup_controller.go
```diff
@@ -845,6 +857,112 @@ func (c *SystemBackupController) createVolumeBackup(volume *longhorn.Volume, sys
...
+		}
+
+		// Return when BackingImage backup exceeds timeout
+		if time.Since(startTime) > datastore.BackingImageBackupTimeout {
+			log.Warn("timed out waiting for BackingImage backups to complete")
+			return
+		}
+	}
+	// This should never be reached.
+	log.Warn("unexpected error: stopped waiting for BackingImage backups without completing, failing or timing out")
```
derekbit, line 938:
```suggestion
	log.Warn("Unexpected error: stopped waiting for BackingImage backups without completing, failing or timing out")
```

controller/system_backup_controller.go
```diff
@@ -845,6 +857,112 @@ func (c *SystemBackupController) createVolumeBackup(volume *longhorn.Volume, sys
...
+			c.updateSystemBackupRecord(record,
+				systemBackupRecordTypeError, longhorn.SystemBackupStateError,
+				constant.EventReasonStart, err.Error(),
+			)
+		} else {
+			c.updateSystemBackupRecord(record,
+				systemBackupRecordTypeNormal, longhorn.SystemBackupStateGenerating,
+				constant.EventReasonStart, SystemBackupMsgStarting,
+			)
+
```
derekbit, line 896: redundant line

controller/system_backup_controller.go
ChanYiLin: We don't delete the the backup name `WaitForVolumeBackupToComplete` 
I think we can follow the same behavior here

The result of `not deleting the name when not found` is that it would definitely timeout in the end
because `if len(backupBackingImages) == 0` will never be zero

@c3y1huang 
WDYT?

ChanYiLin: Is it possible that it is not found for a while, for example like the situation of etcd not synced?

ChanYiLin: I think we can add a log message for the not found case
So user can debug more easily.

derekbit: > I think we can add a log message for the not found case So user can debug more easily.

Sounds good

derekbit (approved):
LGTM

PR #2726, opened 2024-04-12
fix(expand): check volume size first

mantissahz:

Issue #

https://github.com/longhorn/longhorn/issues/7841

Check if volume size can be scheduled by the replica scheduler before expading the pv/pvc size.

### DIFF
--- a/manager/volume.go
+++ b/manager/volume.go
@@ -468,6 +468,15 @@ func (m *VolumeManager) Expand(volumeName string, size int64) (v *longhorn.Volum
 
 	size = util.RoundUpSize(size)
 
+	if v.Spec.Size >= size {
+		logrus.Infof("Volume %v expansion is not allowable since current size %v >= %v", v.Name, v.Spec.Size, size)
+		return v, nil
+	}
+
+	if _, err := m.scheduler.CheckReplicasSizeExpansion(v, v.Spec.Size, size); err != nil {
+		return nil, err
+	}
+
 	kubernetesStatus := &v.Status.KubernetesStatus
 	if kubernetesStatus.PVCName != "" && kubernetesStatus.LastPVCRefAt == "" {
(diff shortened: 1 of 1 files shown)


manager/volume.go
```diff
@@ -468,6 +468,15 @@ func (m *VolumeManager) Expand(volumeName string, size int64) (v *longhorn.Volum
 
 	size = util.RoundUpSize(size)
 
+	if v.Spec.Size >= size {
+		logrus.Infof("Volume %v expansion is not necessary since current size %v >= %v", v.Name, v.Spec.Size, size)
```
derekbit, line 472: How's the modification?
```suggestion
		logrus.Infof("Volume %v expansion is not allowable since current size %v >= %v", v.Name, v.Spec.Size, size)
```

derekbit:
According the description in https://github.com/longhorn/longhorn/issues/7841#issue-2117626490, is it possible to avoid `Online expand a volume to a unavailable value(ex: bigger than node storage available size).`?

manager/volume.go
mantissahz: It is good like denying a shrink operation.
Modified. Thanks

mantissahz:
>According the description in https://github.com/longhorn/longhorn/issues/7841#issue-2117626490, is it possible to avoid Online expand a volume to an unavailable value(ex: bigger than node storage available size).?

This PR will block this operation `expand a volume to an unavailable value` before updating the PVC and the volume size so it should include the online and offline expansions.
WDYT @derekbit?

derekbit:
> bigger than node storage available size

I mean ` bigger than node storage available size`, but it is out of the scope of the PR.
We can create a ticket for addressing the issue in the future.

derekbit (approved):
LGTM

shuo-wu (approved):

PR #2728, opened 2024-04-17
fix(setting): UX of updating danger zone settings

mantissahz:

Issue # longhorn/longhorn#8070

Introduce a new field `Applied` with the `Setting` CRD. It will be true when the setting is applied.


controller/setting_controller.go
```diff
@@ -211,15 +211,27 @@ func (sc *SettingController) handleErr(err error, key interface{}) {
...
 
+	defer func() {
+		err = errors.Wrapf(err, "failed to sync setting for %v", key)
+
+		setting, dsErr := sc.ds.GetSettingExact(types.SettingName(name))
+		if dsErr != nil {
+			sc.logger.WithError(dsErr).Warnf("Failed to get setting: %v", name)
+			return
+		}
+		if setting.Status.Applied != (err == nil) {
```
innobead, line 227: Suggest the following statement for readable code. 

```go
if !setting.Status.Applied && err == nil {
    setting.Status.Applied = true
    // ...
}
```

controller/setting_controller.go
```diff
@@ -211,15 +211,27 @@ func (sc *SettingController) handleErr(err error, key interface{}) {
...
-		err = errors.Wrapf(err, "failed to sync setting for %v", key)
-	}()
-
 	_, name, err := cache.SplitMetaNamespaceKey(key)
 	if err != nil {
 		return err
 	}
 
+	defer func() {
+		err = errors.Wrapf(err, "failed to sync setting for %v", key)
```
innobead, line 220: This err should be still in the original defer closure, as it will be eventually executed.

controller/setting_controller.go
```diff
@@ -1154,6 +1166,7 @@ func (sc *SettingController) syncUpgradeChecker() error {
 
 	if !upgradeCheckerEnabled {
 		if latestLonghornVersion.Value != "" {
+
```
innobead, line 1169: A unnecessary line.

innobead (changes requested):

k8s/pkg/apis/longhorn/v1beta2/setting.go
```diff
@@ -17,6 +18,14 @@ type Setting struct {
 
 	// The value of the setting.
 	Value string `json:"value"`
+
+	Status SettingStatus `json:"status,omitempty"`
```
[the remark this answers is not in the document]
mantissahz: Added. Thanks

controller/setting_controller.go
mantissahz: Modified. Thanks.

controller/setting_controller.go
mantissahz: Removed. Thanks

controller/setting_controller.go
mantissahz: Moved back to the original defer closure.

innobead (approved):

PR #2737, opened 2024-04-23
Prevent unnecessary updates of instanceManager status

derekbit:

Issue Longhorn/longhorn#8420

reflect.DeepEqual treats the two maps `var m1 map[string]process` and `m2 := map[string]process` as different maps. Therefore, to prevent unnecessary updates, we must check both that the length of the maps is zero and that the maps are identical.


innobead (approved):
LGTM

derekbit:
@Mergify backport v1.5.x v1.6.x

PR #2742, opened 2024-04-25
feat(backingimage): backingimage ha eviction enhancement

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/2856

1. Add `minNumberOfCopies`
2. Add `nodeSelector`, `diskSelector`
3. Change the `Spec.Disks` to `Spec.DiskFileSpecMap` and Add `evictionRequested` 
4. Periodically replenish the copies if the number of healthy copies less than the  `minNumberOfCopies` or eviction situation
    - `replenishBackingImageCopies()`
        - if `nonFailedCopies >= MinNumberOfCopies`, we check if we need to replenish one copy for **eviction**
            - replenish one if `NonEvictingCount < MinNumberOfCopies`
        - if `nonFailedCopies < MinNumberOfCopies`
            - replenish one copy to meet the MinNumberOfCopies requirement.
5. Periodically delete evicted copies
    - `cleanupEvictionRequestedBackingImageCopies()`
        - If there is no non evicted healthy copy, don't delete the copy.
        - Otherwise, delete the evicted copy. 
6. Add `concurrent-backing-image-replenish-per-node-limit` to limit the number of BackingImage being synced on the same node at the same time
    - if the number = 0, longhorn won't replenish the BackingImage to new node/disk
7. Adjust replica scheduler, replica won't be scheduled to the node where BackingImage can not be stored.


ChanYiLin:

fixed.

ChanYiLin:
Hi @shuo-wu 
I don't know the context why we always keep the first prepared file when cleanup backing image file on the disk in node controller
I think maybe there is a case when a user tries to evict the file on that node.
Besides, the first prepared file has nothing special
If all the files are failed, we still cleanup the `bids.Spec.NodeID` and `bids.Spec.DiskUUID` to restart the prepare process.

So I changed the logic to
- only starts to cleanup backing image file on the disk  when the file is transferred
- it is okay to clean up the file on the first prepared file

WDYT?
Thansk

ChanYiLin:

fixed.

ChanYiLin:

fixed.

client/generated_backing_image.go
```diff
@@ -11,12 +11,18 @@ type BackingImage struct {
 
 	DeletionTimestamp string `json:"deletionTimestamp,omitempty" yaml:"deletion_timestamp,omitempty"`
 
-	DiskFileStatusMap map[string]BackingImageDiskFileStatus `json:"diskFileStatusMap,omitempty" yaml:"disk_file_status_map,omitempty"`
+	DiskFileStatusMap map[string]interface{} `json:"diskFileStatusMap,omitempty" yaml:"disk_file_status_map,omitempty"`
```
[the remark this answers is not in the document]
ChanYiLin: Fixed, thanks

manager/backingimage.go
```diff
@@ -226,3 +228,23 @@ func (m *VolumeManager) CleanUpBackingImageDiskFiles(name string, diskFileList [
...
+	defer func() {
+		if err == nil {
+			if !reflect.DeepEqual(bi.Spec, existingBI.Spec) {
+				bi, err = m.ds.UpdateBackingImage(bi)
+				return
+			}
+		}
+	}()
+
+	bi.Spec.MinNumberOfCopies = minNumberOfCopies
```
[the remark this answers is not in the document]
ChanYiLin: Fix, directly update the field.

types/setting.go
```diff
@@ -1369,6 +1389,19 @@ var (
...
 	}
+
+	SettingDefinitionDefaultMinNumberOfBackingImageCopies = SettingDefinition{
+		DisplayName: "Default Minimum Number of BackingImage Copies",
+		Description: "The default minimum number of backing image copies Longhorn maintains",
+		Category:    SettingCategoryGeneral,
+		Type:        SettingTypeInt,
+		Required:    true,
+		ReadOnly:    false,
+		Default:     "1",
```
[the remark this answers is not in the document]
ChanYiLin: Sure

controller/replica_controller.go
```diff
@@ -417,6 +417,14 @@ func (rc *ReplicaController) GetBackingImagePathForReplicaStarting(r *longhorn.R
...
 	}
 	if _, exists := bi.Spec.Disks[r.Spec.DiskID]; !exists {
+		res, err := rc.ds.CanPutBackingImageOnDisk(bi, r.Spec.DiskID)
+		if err != nil {
+			log.Warnf("Failed to check if backing image %v can be put on disk %v", bi.Name, r.Spec.DiskID)
+		}
+		if !res {
+			log.Warnf("The backing image %v can not be put on the disk %v", bi.Name, r.Spec.DiskID)
+			return "", nil
+		}
```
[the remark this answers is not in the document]
ChanYiLin: Yes you are right.
It should be an error since we already make it impossible to schedule replica to the node/disk where Backing Image cannot be stored

datastore/longhorn.go
```diff
@@ -4975,6 +5003,22 @@ func (s *DataStore) IsV2DataEngineDisabledForNode(nodeName string) (bool, error)
 	return false, nil
 }
 
+func (s *DataStore) GetDiskBackingImageMap(node *longhorn.Node) (map[string][]*longhorn.BackingImage, error) {
```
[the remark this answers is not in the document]
ChanYiLin: Fixed, thanks

datastore/longhorn.go
```diff
@@ -2664,18 +2664,32 @@ func (s *DataStore) ListReadyNodesContainingEngineImageRO(image string) (map[str
 	return readyNodes, nil
 }
 
-// GetRandomReadyNodeDisk a list of all Node the in the given namespace and
+// GetReadyNodeDiskForBackingImage a list of all Node the in the given namespace and
 // returns the first Node && the first Disk of the Node marked with condition ready and allow scheduling
-func (s *DataStore) GetRandomReadyNodeDisk() (*longhorn.Node, string, error) {
+func (s *DataStore) GetReadyNodeDiskForBackingImage(backingImage *longhorn.BackingImage, usedDisks map[string]bool) (*longhorn.Node, string, error) {
```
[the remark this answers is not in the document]
ChanYiLin: Yes you are right
and the `usedDisks` can be directly referenced to the `backingImage.Spec.DiskFileSpecMap`

controller/node_controller.go
```diff
@@ -1588,6 +1588,70 @@ func (nc *NodeController) createSnapshotMonitor() (mon monitor.Monitor, err erro
...
+	backingImagesToSync := []backingImageToSync{}
+
+	for diskName, diskSpec := range node.Spec.Disks {
+		diskStatus := node.Status.DiskStatus[diskName]
+		diskUUID := diskStatus.DiskUUID
+
+		if diskSpec.EvictionRequested || node.Spec.EvictionRequested {
+			for _, backingImage := range diskBackingImageMap[diskUUID] {
+				// trigger eviction request
+				backingImage.Status.DiskFileStatusMap[diskUUID].EvictionRequested = true
```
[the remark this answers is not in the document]
ChanYiLin: Fixed, thanks

webhook/resources/backingimage/mutator.go
```diff
@@ -102,12 +103,28 @@ func (b *backingImageMutator) Create(request *admission.Request, newObj runtime.
 		patchOps = append(patchOps, `{"op": "replace", "path": "/spec/disks", "value": {}}`)
```
shuo-wu, line 103: I am thinking, should Longhorn reject the requests if users want to update the deprecated fields to a non-empty value?

shuo-wu (approved):
LGTM

webhook/resources/backingimage/mutator.go
ChanYiLin: users can only update the `spec.disks` and `spec.diskFileSpecMap` through command line `kubectl edit ...`
Besides, `spec.disks` is not used in any code.
I think it is not a must but only better have.

ChanYiLin:
### E2E Test PASSED

https://ci.longhorn.io/job/private/job/longhorn-tests-regression/7009/

controller/backing_image_manager_controller.go
```diff
@@ -61,6 +61,9 @@ type BackingImageManagerController struct {
 	backoffMap sync.Map
 
 	versionUpdater func(*longhorn.BackingImageManager) error
+
+	replenishLock             *sync.Mutex
+	inProgressReplenishingMap map[string]struct{}
```
shuo-wu, line 66: Does the controller need to do cleanup when somehow `bim.deletionTimestamp != nil`?

controller/backing_image_controller.go
```diff
@@ -291,9 +304,133 @@ func (bic *BackingImageController) syncBackingImage(key string) (err error) {
...
+		if !fileSpec.EvictionRequested {
+			continue
+		}
+		if !hasNonEvictingHealthyBackingImageCopy && diskUUID == evictingHealthyBackingImageCopyDiskUUID {
+			msg := fmt.Sprintf("Failed to evict backing image copy on disk %v for now since there is no other healthy backing image copy", diskUUID)
+			log.Warn(msg)
+			bic.eventRecorder.Event(bi, corev1.EventTypeNormal, constant.EventReasonFailedDeleting, msg)
+			continue
+		}
+		delete(bi.Spec.DiskFileSpecMap, diskUUID)
```
shuo-wu, line 429: Wait... Isn't `NodeController.syncBackingImageEvictionRequested` responsible for this?

In general, the resource controller cannot update its CR spec.

ChanYiLin: sure, I think we can move the logic to the node controller.

controller/backing_image_manager_controller.go
ChanYiLin: This logic is copied from replica
https://github.com/longhorn/longhorn-manager/blob/master/controller/replica_controller.go#L59

We don't cleanup the `inProgressRebuildingMap ` there
what would be the possible consequence if we don't clean it up?

controller/backing_image_controller.go
```diff
@@ -291,9 +304,133 @@ func (bic *BackingImageController) syncBackingImage(key string) (err error) {
...
+			nonFailedCopies += 1
+		}
+	}
+
+	if nonFailedCopies == 0 {
+		return nil
+	} else if nonFailedCopies >= bi.Spec.MinNumberOfCopies {
+		if err := bic.handleBackingImageCopiesEvictions(nonFailedCopies, bi); err != nil {
+			return nil
+		}
```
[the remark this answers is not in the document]
ChanYiLin: Do you mean
we count SatisfiedCopies here and then we can remove the logic of `handleBackingImageCopiesEvictions`?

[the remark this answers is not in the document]
ChanYiLin: fixed.

controller/backing_image_manager_controller.go
ChanYiLin: fixed.

controller/backing_image_controller.go
ChanYiLin: Hi @shuo-wu 
I think keep the delete copy logic in the backing image controller is better.

> Isn't NodeController.syncBackingImageEvictionRequested responsible for this?
inside the NodeController, it only update the eviction request to true for the copy on the node or disks
So the BackingImageController can then replenish the copy to other nodes/disks

> In general, the resource controller cannot update its CR spec.

I know the reason is to prevent unlimited updating loop. But
- we actually already put the **replenish copy logic in the BackingImageController** as well which will add the disk to the `bi.Spec`.
- It is much better to put it here then the node controller. Here is the reason
     - to decide if we can evict the copy on one disk needs the knowledge of all the copies of the backing image
     - NodeController only consider its own node and disks. But backing image copies can be on two nodes
     - It could possible that two node controller all decide to delete one of the evicted the backing image on their node at the same time based on the status it gets at the moment. Thus, the race condition can happen, perhaps that is the last healthy one because the other NodeController already delete the other one.

Put the eviction logic and replenish logic in the BackingImageController is better.
Because it handle the backing image one at the time and has the whole picture of the status.
Thus it actually won't get into unlimited update loop because it will get balanced on the minimal HA number in limited reconcile times.

controller/backing_image_controller.go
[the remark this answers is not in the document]
ChanYiLin: Wait this is not correct.
Because, if we count the SatisfiedCopies
- Non-failed &&
- Not eviction requested &&
- All Selector satisfied
Then we cannot distinguish the situation of SatisfiedCopies == 0
It can be 
1. there is no healthy copy, all the copies are dead, we should re-prepare the copy (for download and export-from-volume)
2. there is one copy, but it get eviction request.
    - In this case, we should first replenish one as backup before really delete the evicted copy

Thus, we can't mixed it together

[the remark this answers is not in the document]
shuo-wu: > there is one copy, but it get eviction request.

Nice catch. This is what I haven't considered.

NIT: And the next question is, does the controller need to count `if nonFailedCopies >= bi.Spec.MinNumberOfCopies`? I think it can directly do replenishment when `if NonEvictingCount < bi.Spec.MinNumberOfCopies`. I mean:
```
	if nonFailedCopies == 0 {
		return nil
	} else {
		// Get NonEvictingCount
		......
		if NonEvictingCount < bi.Spec.MinNumberOfCopies {
			log.Infof("Creating one more backing image copy for eviction")
			readyNode, readyDiskName, err := bic.ds.GetReadyNodeDiskForBackingImage(bi)
			if err != nil {
				logrus.WithError(err).Warnf("failed to create the backing image copy")
				return nil
			}
			// BackingImageManager will then sync the BackingImage to the disk
			bi.Spec.DiskFileSpecMap[readyNode.Status.DiskStatus[readyDiskName].DiskUUID] = &longhorn.BackingImageDiskFileSpec{}
		}
	}
```

controller/backing_image_controller.go
shuo-wu: Got it, the explanation makes sense. 

Adding or deleting entries for `bi.Spec.DiskFileSpecMap` involves multiple nodes. Asking a specific NodeController to check other nodes for that field is not a good idea.

In other words, the BackingImage is similar to Volume CRD while file copies of a BackingImage are similar to replicas of a volume. By contrast, it's BackingImageController's responsibility to handle the BI copies. Currently, I don't want to introduce a new CRD for the copies hence the controller updating its own CR spec looks weird but somehow is reasonable.

@derekbit Are you fine with breaking the controller convention here?

derekbit: >  I don't want to introduce a new CRD for the copies hence the controller updating its own CR spec looks weird but somehow is reasonable.

A bit weird but I was struggling with it as well.
To avoid potential deadlock issues, it would be better to move this to another controller. However, I'm fine with the current implementation and we can optimize it later if it requires significant effort.


> the BackingImage is similar to Volume CRD while file copies of a BackingImage are similar to replicas of a volume.

Yes, we might need a high-level CRD (like a volume) for `BackingImage` (like a replica) in the future.

ChanYiLin: As discussed, for current version we still implement the logic in the BackingImageController
We will refactor this part in the future with this issue: https://github.com/longhorn/longhorn/issues/8783

controller/backing_image_controller.go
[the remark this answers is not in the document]
ChanYiLin: sure this could work

k8s/pkg/apis/longhorn/v1beta2/backingimage.go
```diff
@@ -36,16 +39,30 @@ type BackingImageDiskFileStatus struct {
...
 	// +optional
+	DiskFileSpecMap map[string]*BackingImageDiskFileSpec `json:"diskFileSpecMap"`
+	// +optional
 	Checksum string `json:"checksum"`
 	// +optional
 	SourceType BackingImageDataSourceType `json:"sourceType"`
 	// +optional
 	SourceParameters map[string]string `json:"sourceParameters"`
+	// +optional
+	MinNumberOfCopies int `json:"minNumberOfCopies"`
```
derekbit, line 61: nit: using `MinNumberOfReplicas` might be better.

types/setting.go
```diff
@@ -290,6 +294,7 @@ var (
 		SettingNameDisableRevisionCounter:                                   SettingDefinitionDisableRevisionCounter,
 		SettingNameReplicaReplenishmentWaitInterval:                         SettingDefinitionReplicaReplenishmentWaitInterval,
 		SettingNameConcurrentReplicaRebuildPerNodeLimit:                     SettingDefinitionConcurrentReplicaRebuildPerNodeLimit,
+		SettingNameConcurrentBackingImageReplenishPerNodeLimit:              SettingDefinitionConcurrentBackingImageReplenishPerNodeLimit,
```
derekbit, line 297: nit: use `SettingNameConcurrentBackingImageCopyReplenishPerNodeLimit`

types/types.go
```diff
@@ -741,6 +741,17 @@ func ValidateReplicaCount(count int) error {
 	return nil
 }
 
+func ValidateMinNumberOfBackingIamgeCopies(number int) error {
+
```
derekbit, line 745: Remove redundant line

types/types.go
```diff
@@ -741,6 +741,17 @@ func ValidateReplicaCount(count int) error {
...
+	definition, _ := GetSettingDefinition(SettingNameDefaultMinNumberOfBackingImageCopies)
```
derekbit, line 746: Check the returned boolean value before accessing definition.

webhook/resources/backingimage/mutator.go
```diff
@@ -135,6 +152,12 @@ func (b *backingImageMutator) Update(request *admission.Request, oldObj runtime.
 	if backingImage.Spec.Disks == nil {
 		patchOps = append(patchOps, `{"op": "replace", "path": "/spec/disks", "value": {}}`)
 	}
+
+	// Handle Spec.DiskFileSpecMap
+	if backingImage.Spec.DiskFileSpecMap == nil {
+		patchOps = append(patchOps, `{"op": "replace", "path": "/spec/disks", "value": {}}`)
```
derekbit, line 158: Isn't it `/spec/diskFileSpecMap`?

webhook/resources/backingimage/mutator.go
```diff
@@ -102,12 +103,28 @@ func (b *backingImageMutator) Create(request *admission.Request, newObj runtime.
 		patchOps = append(patchOps, `{"op": "replace", "path": "/spec/disks", "value": {}}`)
 	}
 
+	// Handle Spec.DiskFileSpecMap
+	if backingImage.Spec.DiskFileSpecMap == nil {
+		patchOps = append(patchOps, `{"op": "replace", "path": "/spec/disks", "value": {}}`)
```
derekbit, line 108: Isn't it `/spec/diskFileSpecMap`?

webhook/resources/backingimage/mutator.go
```diff
@@ -102,12 +103,28 @@ func (b *backingImageMutator) Create(request *admission.Request, newObj runtime.
...
 		return nil, werror.NewInvalidError(err.Error(), "")
 	}
+
+	if backingImage.Spec.MinNumberOfCopies == 0 {
+		minNumberOfCopies, err := b.getDefaultMinNumberOfBackingImageCopies()
+		if err != nil {
+			err = errors.Wrap(err, "BUG: cannot get valid number for setting default min number of backing image copies")
+			return nil, werror.NewInvalidError(err.Error(), "")
+		}
+		logrus.Infof("Use the default number of minimum number of copies %v", minNumberOfCopies)
```
derekbit, line 124:
```suggestion
		logrus.Infof("Use the default minimum number of copies %v for backing image %v", minNumberOfCopies, backingImage.Name)
```

webhook/resources/backingimage/mutator.go
```diff
@@ -102,12 +103,28 @@ func (b *backingImageMutator) Create(request *admission.Request, newObj runtime.
...
 	patchOp, err := common.GetLonghornLabelsPatchOp(backingImage, longhornLabels, nil)
 	if err != nil {
 		err := errors.Wrapf(err, "failed to get label patch for backingImage %v", backingImage.Name)
 		return nil, werror.NewInvalidError(err.Error(), "")
 	}
+
+	if backingImage.Spec.MinNumberOfCopies == 0 {
+		minNumberOfCopies, err := b.getDefaultMinNumberOfBackingImageCopies()
+		if err != nil {
+			err = errors.Wrap(err, "BUG: cannot get valid number for setting default min number of backing image copies")
```
derekbit, line 121: Should we need "BUG"?

```suggestion
			err = errors.Wrap(err, "failed to get valid number for setting default min number of backing image copies")
```

derekbit (commented):
I will continue reviewing the controllers and scheduler tomorrow morning.

controller/backing_image_controller.go
```diff
@@ -291,9 +304,117 @@ func (bic *BackingImageController) syncBackingImage(key string) (err error) {
...
+		if !fileSpec.EvictionRequested {
+			continue
+		}
+		if !hasNonEvictingHealthyBackingImageCopy && diskUUID == evictingHealthyBackingImageCopyDiskUUID {
+			msg := fmt.Sprintf("Failed to evict backing image copy on disk %v for now since there is no other healthy backing image copy", diskUUID)
+			log.Warn(msg)
+			bic.eventRecorder.Event(bi, corev1.EventTypeNormal, constant.EventReasonFailedDeleting, msg)
+			continue
+		}
+		delete(bi.Spec.DiskFileSpecMap, diskUUID)
```
shuo-wu, line 413: NIT: Add a comment to explain this special spec update

shuo-wu (commented):
In general LGTM except for the some comments from Derek

controller/backing_image_manager_controller.go
```diff
@@ -713,8 +726,22 @@ func (c *BackingImageManagerController) prepareBackingImageFiles(currentBIM *lon
 			continue
 		}
 
+		// Before syncing the backing image copy to this backing image manager,
+		// check if there are more than ReplenishPerNodeLimit number of backing image copies are being synced on this node.
+		canSync, err := c.CanSyncCopy(currentBIM, biName, currentBIM.Spec.NodeID, currentBIM.Spec.DiskUUID)
```
derekbit, line 731: nit:
I thought CanSyncCopy only checks if syncing copy is available, but I found it also delete or add entry to `c.inProgressReplenishingMap`?
I would suggest separating this functionality into two distinct functions: one for checking availability and another for modifying the map.

k8s/pkg/apis/longhorn/v1beta2/backingimage.go
ChanYiLin: prefer copy to avoid get confused with the volume replica

types/setting.go
ChanYiLin: fixed.

types/types.go
ChanYiLin: fixed.

types/types.go
ChanYiLin: fixed.

webhook/resources/backingimage/mutator.go
ChanYiLin: fixed.

webhook/resources/backingimage/mutator.go
ChanYiLin: fixed.

webhook/resources/backingimage/mutator.go
ChanYiLin: fixed.

webhook/resources/backingimage/mutator.go
ChanYiLin: fixed.

controller/backing_image_controller.go
ChanYiLin: added

controller/node_controller.go
```diff
@@ -1186,8 +1190,11 @@ func (nc *NodeController) cleanUpBackingImagesInDisks(node *longhorn.Node) error
 			log.WithError(err).Warn("Failed to get the backing image data source when cleaning up the images in disks")
 			continue
 		}
+		if bids == nil {
+			continue
+		}
```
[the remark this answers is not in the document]
ChanYiLin: no need
original version bids is only used here and it checks if bids is nil as well
``` 
isFirstFile := bids != nil && !bids.Spec.FileTransferred && diskUUID == bids.Spec.DiskUUID
```

controller/node_controller.go
```diff
@@ -1201,17 +1208,17 @@ func (nc *NodeController) cleanUpBackingImagesInDisks(node *longhorn.Node) error
 	return nil
 }
 
-func BackingImageDiskFileCleanup(node *longhorn.Node, bi *longhorn.BackingImage, bids *longhorn.BackingImageDataSource, waitInterval time.Duration, haRequirement int) {
-	if bi.Spec.Disks == nil || bi.Status.DiskLastRefAtMap == nil {
+func BackingImageDiskFileCleanup(node *longhorn.Node, bi *longhorn.BackingImage, bids *longhorn.BackingImageDataSource, waitInterval time.Duration, minNumberOfCopies int) {
+	if bi.Spec.DiskFileSpecMap == nil || bi.Status.DiskLastRefAtMap == nil || !bids.Spec.FileTransferred {
```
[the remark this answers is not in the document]
ChanYiLin: bids is to prepare the first backing file
when it is ready it will be transferred to backing image manager

backing image can only be used when it is transferred
so we should only check if we need to cleanup the backing image file after it is transferred

original code did the same thing, I just moved it to the very beginning.
```
		isFirstFile := bids != nil && !bids.Spec.FileTransferred && diskUUID == bids.Spec.DiskUUID
		if isFirstFile {
			continue
		}
```

controller/node_controller.go
```diff
@@ -1584,6 +1587,70 @@ func (nc *NodeController) createSnapshotMonitor() (mon monitor.Monitor, err erro
...
+
+	diskBackingImageMap, err := nc.ds.GetDiskBackingImageMap()
+	if err != nil {
+		return err
+	}
+
+	type backingImageToSync struct {
+		*longhorn.BackingImage
+		diskUUID string
+		evict    bool
```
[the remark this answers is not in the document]
ChanYiLin: fixed.

controller/node_controller.go
```diff
@@ -1584,6 +1587,70 @@ func (nc *NodeController) createSnapshotMonitor() (mon monitor.Monitor, err erro
...
+			}
+		}
+	}
+
+	for _, backingImageToSync := range backingImagesToSync {
+		backingImageLog := log.WithField("backingimage", backingImageToSync.Name).WithField("disk", backingImageToSync.diskUUID)
+		if backingImageToSync.evict {
+			backingImageLog.Infof("Requesting backing image copy eviction")
+			if _, err := nc.ds.UpdateBackingImage(backingImageToSync.BackingImage); err != nil {
+				backingImageLog.Warn("Failed to request backing image copy eviction, will enqueue then resync the node")
```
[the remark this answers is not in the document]
ChanYiLin: fixed.

controller/node_controller.go
```diff
@@ -1584,6 +1587,70 @@ func (nc *NodeController) createSnapshotMonitor() (mon monitor.Monitor, err erro
...
+				nc.enqueueNodeRateLimited(node)
+				continue
+			}
+			nc.eventRecorder.Eventf(backingImageToSync.BackingImage, corev1.EventTypeNormal, constant.EventReasonEvictionUserRequested, "Requesting backing image %v eviction from node %v and disk %v", backingImageToSync.Name, node.Spec.Name, backingImageToSync.diskUUID)
+		} else {
+			backingImageLog.Infof("Cancelling backing image copy eviction")
+			if _, err := nc.ds.UpdateBackingImage(backingImageToSync.BackingImage); err != nil {
+				backingImageLog.Warn("Failed to cancel backing image copy eviction, will enqueue then resync the node")
```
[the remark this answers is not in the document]
ChanYiLin: fixed.

controller/node_controller.go
```diff
@@ -1584,6 +1587,70 @@ func (nc *NodeController) createSnapshotMonitor() (mon monitor.Monitor, err erro
...
+				nc.enqueueNodeRateLimited(node)
+				continue
+			}
+			nc.eventRecorder.Eventf(backingImageToSync.BackingImage, corev1.EventTypeNormal, constant.EventReasonEvictionCanceled, "Requesting backing image %v eviction from node %v and disk %v", backingImageToSync.Name, node.Spec.Name, backingImageToSync.diskUUID)
+		}
+	}
```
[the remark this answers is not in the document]
ChanYiLin: fixed, thanks

controller/replica_controller.go
```diff
@@ -412,12 +412,19 @@ func (rc *ReplicaController) GetBackingImagePathForReplicaStarting(r *longhorn.R
...
+	if bi.Spec.DiskFileSpecMap == nil {
 		log.Warnf("The requested backing image %v has not started disk file handling", bi.Name)
 		return "", nil
 	}
-	if _, exists := bi.Spec.Disks[r.Spec.DiskID]; !exists {
-		bi.Spec.Disks[r.Spec.DiskID] = ""
+	if _, exists := bi.Spec.DiskFileSpecMap[r.Spec.DiskID]; !exists {
+		res, err := rc.ds.CanPutBackingImageOnDisk(bi, r.Spec.DiskID)
+		if err != nil {
+			log.Warnf("Failed to check if backing image %v can be put on disk %v", bi.Name, r.Spec.DiskID)
```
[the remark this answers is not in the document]
ChanYiLin: fixed.

controller/backing_image_manager_controller.go
```diff
@@ -354,6 +361,16 @@ func (c *BackingImageManagerController) cleanupBackingImageManager(bim *longhorn
 			}
 		}
 	}
+
+	// delete in progress record of this manager from the inProgressReplenishingMap
+	// each manager only controls only one disk
+	diskUUID := bim.Spec.DiskUUID
+	for biNameDiskID := range c.inProgressReplenishingMap {
+		if strings.Contains(biNameDiskID, diskUUID) {
```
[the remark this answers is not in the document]
ChanYiLin: fixed.

controller/backing_image_manager_controller.go
```diff
@@ -1235,3 +1265,74 @@ func (m *BackingImageManagerMonitor) pollAndUpdateBackingImageFileMap() (needSto
 func (c *BackingImageManagerController) isResponsibleFor(bim *longhorn.BackingImageManager) bool {
 	return isControllerResponsibleFor(c.controllerID, c.ds, bim.Name, bim.Spec.NodeID, bim.Status.OwnerID)
 }
+
+func (c *BackingImageManagerController) CanSyncCopy(bim *longhorn.BackingImageManager, biName, nodeID, diskUUID string) (bool, error) {
```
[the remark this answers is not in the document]
ChanYiLin: fixed.

datastore/longhorn.go
```diff
@@ -5045,6 +5073,22 @@ func (s *DataStore) IsV2DataEngineDisabledForNode(nodeName string) (bool, error)
...
+	}
+
+	for _, bi := range backingImages {
+		for diskUIID := range bi.Status.DiskFileStatusMap {
+			diskBackingImageMap[diskUIID] = append(diskBackingImageMap[diskUIID], bi)
+		}
+	}
+
+	return diskBackingImageMap, nil
+}
```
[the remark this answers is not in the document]
ChanYiLin: will be resolved when we abstract the copy CRD from the backing image CRD

controller/backing_image_manager_controller.go
ChanYiLin: The logic here is copied from replica: https://github.com/longhorn/longhorn-manager/blob/v1.6.2/controller/replica_controller.go#L436

the map is a in-memory map, so we iterate all the entries every time so it can rebuild if the controller or the pod crashes
we need to add and delete the map  so we can finally decide if it can sync the copy now

derekbit (approved):
LGTM

PR #2743, opened 2024-04-25
 Remove v2 volume offline rebuilding

derekbit:

Issue longhorn/longhorn#8442

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -1054,19 +1054,10 @@ func (c *VolumeController) cleanupCorruptedOrStaleReplicas(v *longhorn.Volume, r
 			continue
 		}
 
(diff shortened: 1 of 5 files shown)


derekbit:
image

stuck? cc @FrankYang0529

innobead:
> image
> 
> stuck? cc @FrankYang0529

@derekbit branch protection needs to be updated. Derek, please help update each repo after @FrankYang0529 replaces drone with github action. 

image

innobead (dismissed):
LGTM. 

Do we need to remove the setting resource as well? or just leave it as w/o any effect?

derekbit:
> LGTM.
> 
> Do we need to remove the setting resource as well? or just leave it as w/o any effect?

Retaining the setting is fine. Let me add it back.

innobead (approved):

PR #2744, opened 2024-04-25
Add a global FreezeFS setting and a volume FreezeFS field

ejweber:

longhorn/longhorn#2187

- Depending on the global and volume-specific setting, either pass `freezeFS == true` or `freezeFS == false` in the proxy snapshot request to the engine.
- Make the settings configurable via the REST API.

Depends on:

- https://github.com/longhorn/types/pull/8
- https://github.com/longhorn/longhorn-engine/pull/1083
- https://github.com/longhorn/longhorn-instance-manager/pull/477
- https://github.com/longhorn/longhorn/pull/8447

Need to resolve imports before merging.


controller/backing_image_data_source_controller.go
```diff
@@ -836,14 +836,19 @@ func (c *BackingImageDataSourceController) prepareRunningParameters(bids *longho
 		}
 	}
 	if newSnapshotRequired {
+		freezeFilesystem, err := c.ds.GetFreezeFilesystemForSnapshotSetting(e)
+		if err != nil {
+			return err
+		}
```
james-munson, line 842: ~If for some reason we can't get a true or false, would be better to default to "sync, don't freeze" rather than failing the operation?~
Never mind, I have my answer.  The datastore method only returns error on serious conditions like "can't find that volume."  Simple mechanical lookup errors will default to "false".

james-munson (commented):
Looks reasonable to me, although I would probably not catch omissions.

ejweber:
I ran out of time to retest this now that the dependencies are resolved and coming from the lower-level merged PRs, but I don't anticipate any change in functionality. It is ready for review (and I will retest the full stack before merge).

ejweber:
I retested the full stack and everything is looking good. Please review when you can.

engineapi/snapshot.go
```diff
@@ -15,13 +15,17 @@ import (
...
+func (e *EngineBinary) SnapshotCreate(engine *longhorn.Engine, name string, labels map[string]string,
+	freezeFilesystem bool) (string, error) {
 	args := []string{"snapshot", "create"}
 	for k, v := range labels {
 		args = append(args, "--label", k+"="+v)
 	}
 	args = append(args, name)
 
+	// Ignore freezeFilesystem argument. v1.7.0 engines understand the corresponding freeze-fs CLI flag, but older
+	// engines do not.
```
PhanLe1010, line 27: Are we expecting that newer engine will not be called by this CLI command?

controller/volume_controller.go
```diff
@@ -4434,7 +4439,7 @@ func (c *VolumeController) createSnapshot(snapshotName string, labels map[string
 		}
 	}
 
-	snapshotName, err = engineClientProxy.SnapshotCreate(e, snapshotName, labels)
+	snapshotName, err = engineClientProxy.SnapshotCreate(e, snapshotName, labels, freezeFilesystem)
```
PhanLe1010, line 4442: Looks good to me because looks like this function is used for volume cloning so we still need to freeze filesystem

controller/system_rollout_controller.go
```diff
@@ -457,11 +459,8 @@ func (c *SystemRolloutController) systemRollout() error {
 		}
 		wg.Wait()
 
-		// Need to wait until backing images are restored, so the volumes with backingimages won't be rejected by webhook when restoring.
-		// PV/PVC restoration depends on Volume, so move them after volume restoration.
+		// Need to wait until backingimages are restored, so the volumes with backingimages can be restored.
 		c.restore(types.LonghornKindVolumeList, c.restoreVolumes, log)
-		c.restore(types.KubernetesKindPersistentVolumeList, c.restorePersistentVolumes, log)
-		c.restore(types.KubernetesKindPersistentVolumeClaimList, c.restorePersistentVolumeClaims, log)
```
PhanLe1010, line 464: Any reason that we are touching this logic in this PR?

controller/setting_controller.go
```diff
@@ -1858,6 +1863,7 @@ func (info *ClusterInfo) collectVolumesInfo() error {
 	info.structFields.fields.AppendCounted(restoreVolumeRecurringJobCountStruct)
 	info.structFields.fields.AppendCounted(snapshotDataIntegrityCountStruct)
 	info.structFields.fields.AppendCounted(unmapMarkSnapChainRemovedCountStruct)
+	info.structFields.fields.AppendCounted(freezeFilesystemForSnapshotCountStruct)
```
PhanLe1010, line 1866: We also need to update the upgrade responder server as well https://github.com/longhorn/longhorn/blob/master/deploy/upgrade_responder_server/chart-values.yaml

controller/system_rollout_controller.go
ejweber: Good catch. Looks like a consequence of a bad rebase on https://github.com/longhorn/longhorn-manager/pull/2819. Fixing.

controller/setting_controller.go
ejweber: Got it. Added https://github.com/longhorn/longhorn/pull/8447/commits/2b7f9282015a693ae63651cfa9233444b26e8a7b.

engineapi/snapshot.go
ejweber: I think the answer to your question is generally yes, we do not expect it. Of course, it is still possible if you are running a very old instance-manager with the latest engine. The EngineBinary interface was deprecated in `v1.3.0` and is getting more difficult to test with (must run a `v1.3.0` instance-manager with a `v1.7.0` engine). IMO, we should not attempt to support newer CLI API calls with this interface unless they are critical to stability. WDYT?

PhanLe1010: > IMO, we should not attempt to support newer CLI API calls with this interface unless they are critical to stability
I agree!

PhanLe1010 (approved):
LGTM

PR #2761, opened 2024-04-28
Record instance-manager name for a block disksin `node.spec.diskStatus`

derekbit:

Issue longhorn/longhorn#8458

A block-type disk can only be managed by a specific spdk_tgt within an instance manager. The commit is intended to record the name of the instance manager associated with that disk.
On the other hand, for a filesystem-type disk, any instance manager can access it. Therefore, the recorded instance manager for a filesystem-type disk is always the default instance manager.

### DIFF
--- a/controller/monitor/disk_monitor.go
+++ b/controller/monitor/disk_monitor.go
@@ -61,6 +61,7 @@ type CollectedDiskInfo struct {
 	DiskDriver                    longhorn.DiskDriver
 	Condition                     *longhorn.Condition
 	OrphanedReplicaDirectoryNames map[string]string
+	InstanceManagerName           string
 }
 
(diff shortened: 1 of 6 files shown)


DamiaSan (approved):

controller/monitor/disk_monitor.go
```diff
@@ -206,6 +207,7 @@ func (m *NodeMonitor) closeDiskServiceClients(clients map[longhorn.DataEngineTyp
 	for _, client := range clients {
 		if client.c != nil {
 			client.c.Close()
+			client.c = nil
```
james-munson, line 210: And some good cleanup changes, as well.

james-munson (approved):
LGTM

PR #2762, opened 2024-04-29
fix: remove parameters for checkout action

FrankYang0529:
https://github.com/longhorn/longhorn/issues/8039

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,8 +14,6 @@ jobs:
     steps:
     - name: Checkout code
       uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


innobead (approved):

innobead (commented):
The default should be just the PR ref impliclity, but I am curious if keeping the removed code, can the job still run correctly?

FrankYang0529:
> The default should be just the PR ref impliclity, but I am curious if keeping the removed code, can the job still run correctly?

I think it's ok for pull requests, but not sure for branch/tag.

PR #2763, opened 2024-04-29
fix: check if environment settings are achieved

mantissahz:

Issue # longhorn/longhorn#7571

Check if the environment settings are achieved, ref: [environment_check.sh](https://github.com/longhorn/longhorn/blob/master/scripts/environment_check.sh)

### DIFF
--- a/app/daemon.go
+++ b/app/daemon.go
@@ -24,7 +24,6 @@ import (
 	"github.com/longhorn/longhorn-manager/datastore"
 	"github.com/longhorn/longhorn-manager/manager"
 	"github.com/longhorn/longhorn-manager/meta"
-	recoverybackend "github.com/longhorn/longhorn-manager/recovery_backend"
 	"github.com/longhorn/longhorn-manager/types"
 	"github.com/longhorn/longhorn-manager/upgrade"
 	"github.com/longhorn/longhorn-manager/util"
@@ -32,6 +31,7 @@ import (
 	"github.com/longhorn/longhorn-manager/webhook"
 
 	metricscollector "github.com/longhorn/longhorn-manager/metrics_collector"
+	recoverybackend "github.com/longhorn/longhorn-manager/recovery_backend"
 )
 
 const (
@@ -131,7 +131,7 @@ func startManager(c *cli.Context) error {
 	kubeconfigPath := c.String(FlagKubeConfig)
 
 	if err := environmentCheck(); err != nil {
-		return errors.Wrap(err, "Failed environment check, please make sure you have iscsiadm/open-iscsi installed on the host")
+		return errors.Wrap(err, "failed to check environment, please make sure you have iscsiadm/open-iscsi installed on the host")
 	}
 
 	currentNodeID, err := util.GetRequiredEnv(types.EnvNodeName)
@@ -253,6 +253,8 @@ func startManager(c *cli.Context) error {
 }
 
 func environmentCheck() error {
+	// Here we only check if the necessary tool the iscsiadm is installed when Longhorn starts up.
+	// Others tools and settings like kernel versions, multipathd, nfs client, etc. are checked in the node controller (every 30 sec).
 	namespaces := []lhtypes.Namespace{lhtypes.NamespaceMnt, lhtypes.NamespaceNet}
 	nsexec, err := lhns.NewNamespaceExecutor(iscsiutil.ISCSIdProcess, lhtypes.HostProcDirectory, namespaces)
 	if err != nil {
--- a/app/pre_upgrade.go
+++ b/app/pre_upgrade.go
@@ -56,5 +56,9 @@ func preUpgrade(c *cli.Context) error {
 		return err
 	}
 
+	if err := environmentCheck(); err != nil {
+		return errors.Wrap(err, "failed to check environment, please make sure you have iscsiadm/open-iscsi installed on the host")
+	}
+
 	return nil
 }
--- a/controller/node_controller.go
+++ b/controller/node_controller.go
@@ -25,6 +25,10 @@ import (
 	clientset "k8s.io/client-go/kubernetes"
 	v1core "k8s.io/client-go/kubernetes/typed/core/v1"
 
+	lhexec "github.com/longhorn/go-common-libs/exec"
+	lhns "github.com/longhorn/go-common-libs/ns"
+	lhtypes "github.com/longhorn/go-common-libs/types"
+
 	"github.com/longhorn/longhorn-manager/constant"
 	"github.com/longhorn/longhorn-manager/datastore"
 	"github.com/longhorn/longhorn-manager/engineapi"
@@ -558,6 +562,9 @@ func (nc *NodeController) syncNode(key string) (err error) {
 		}
 	}
 
+	// check if environment settings meet the requirements on current node
+	nc.environmentCheck(kubeNode, node)
+
 	if err := nc.syncInstanceManagers(node); err != nil {
 		return err
 	}
@@ -993,6 +1000,152 @@ func (nc *NodeController) syncNodeStatus(pod *corev1.Pod, node *longhorn.Node) e
 	}
 
 	return nil
+}
+
+func (nc *NodeController) environmentCheck(kubeNode *corev1.Node, node *longhorn.Node) {
+	// Need to find the better way to check if various kernel versions are supported
+	namespaces := []lhtypes.Namespace{lhtypes.NamespaceMnt, lhtypes.NamespaceNet}
+	nc.syncPackagesInstalled(kubeNode, node, namespaces)
+	nc.syncMultipathd(node, namespaces)
+	nc.syncNFSClientVersion(kubeNode, node, namespaces)
+}
+
+func (nc *NodeController) syncPackagesInstalled(kubeNode *corev1.Node, node *longhorn.Node, namespaces []lhtypes.Namespace) {
+	osImage := strings.ToLower(kubeNode.Status.NodeInfo.OSImage)
+	queryPackagesCmd := ""
+	packages := []string{}
+	switch {
+	case strings.Contains(osImage, "ubuntu"):
+	case strings.Contains(osImage, "debian"):
+		queryPackagesCmd = "dpkg -l | grep -w"
+		packages = []string{"nfs-common", "open-iscsi", "cryptsetup", "dmsetup"}
+	case strings.Contains(osImage, "centos"):
+	case strings.Contains(osImage, "fedora"):
+	case strings.Contains(osImage, "rocky"):
+	case strings.Contains(osImage, "ol"):
+		queryPackagesCmd = "rpm -q"
+		packages = []string{"nfs-utils", "iscsi-initiator-utils", "cryptsetup", "device-mapper"}
+	case strings.Contains(osImage, "suse"):
+		queryPackagesCmd = "rpm -q"
+		packages = []string{"nfs-client", "open-iscsi", "cryptsetup", "device-mapper"}
+	case strings.Contains(osImage, "arch"):
+		queryPackagesCmd = "pacman -Q"
+		packages = []string{"nfs-utils", "open-iscsi", "cryptsetup", "device-mapper"}
+	case strings.Contains(osImage, "gentoo"):
+		queryPackagesCmd = "qlist -I"
+		packages = []string{"net-fs/nfs-utils", "sys-block/open-iscsi", "sys-fs/cryptsetup", "sys-fs/lvm2"}
+	default:
+		node.Status.Conditions = types.SetCondition(node.Status.Conditions, longhorn.NodeConditionTypeRequiredPackages, longhorn.ConditionStatusFalse,
+			string(longhorn.NodeConditionReasonUnknownOS),
+			fmt.Sprintf("Unable to verify the required packages because the OS image '%v' is unknown to the Longhorn system. Please ensure the required packages are installed.", osImage))
+		return
+	}
+
+	nsexec, err := lhns.NewNamespaceExecutor(lhtypes.ProcessNone, lhtypes.HostProcDirectory, namespaces)
+	if err != nil {
+		node.Status.Conditions = types.SetCondition(node.Status.Conditions, longhorn.NodeConditionTypeRequiredPackages, longhorn.ConditionStatusFalse,
+			string(longhorn.NodeConditionReasonNamespaceExecutorErr),
+			fmt.Sprintf("Failed to get namespace executor: %v", err.Error()))
+		return
+	}
+	notFoundPkgs := []string{}
+	for _, pkg := range packages {
+		args := []string{pkg}
+		if _, err := nsexec.Execute(nil, queryPackagesCmd, args, lhtypes.ExecuteDefaultTimeout); err != nil {
+			nc.logger.WithError(err).Debugf("Package %v is not found in node %v", pkg, node.Name)
+			notFoundPkgs = append(notFoundPkgs, pkg)
+		}
+	}
+
+	if len(notFoundPkgs) > 0 {
+		node.Status.Conditions = types.SetCondition(node.Status.Conditions, longhorn.NodeConditionTypeRequiredPackages, longhorn.ConditionStatusFalse,
+			string(longhorn.NodeConditionReasonPackagesNotInstalled),
+			fmt.Sprintf("Missing packages: %v", notFoundPkgs))
+		return
+	}
+
+	node.Status.Conditions = types.SetCondition(node.Status.Conditions, longhorn.NodeConditionTypeRequiredPackages, longhorn.ConditionStatusTrue, "", "")
+}
+
+func (nc *NodeController) syncMultipathd(node *longhorn.Node, namespaces []lhtypes.Namespace) {
+	nsexec, err := lhns.NewNamespaceExecutor(lhtypes.ProcessNone, lhtypes.HostProcDirectory, namespaces)
+	if err != nil {
+		node.Status.Conditions = types.SetCondition(node.Status.Conditions, longhorn.NodeConditionTypeMultipathd, longhorn.ConditionStatusFalse,
+			string(longhorn.NodeConditionReasonNamespaceExecutorErr),
+			fmt.Sprintf("Failed to get namespace executor: %v", err.Error()))
+		return
+	}
+	args := []string{"show", "status"}
+	if result, _ := nsexec.Execute(nil, "multipathd", args, lhtypes.ExecuteDefaultTimeout); result != "" {
+		node.Status.Conditions = types.SetCondition(node.Status.Conditions, longhorn.NodeConditionTypeMultipathd, longhorn.ConditionStatusFalse,
+			string(longhorn.NodeConditionReasonMultipathdIsRunning),
+			"multipathd is running with a known issue that affects Longhorn. See description and solution at https://longhorn.io/kb/troubleshooting-volume-with-multipath")
+		return
+	}
+
+	node.Status.Conditions = types.SetCondition(node.Status.Conditions, longhorn.NodeConditionTypeMultipathd, longhorn.ConditionStatusTrue, "", "")
+}
+
+func (nc *NodeController) syncNFSClientVersion(kubeNode *corev1.Node, node *longhorn.Node, namespaces []lhtypes.Namespace) {
+	kernelVersion := kubeNode.Status.NodeInfo.KernelVersion
+	nfsClientVersions := []string{"CONFIG_NFS_V4_2", "CONFIG_NFS_V4_1", "CONFIG_NFS_V4"}
+
+	nsexec, err := lhns.NewNamespaceExecutor(lhtypes.ProcessNone, lhtypes.HostProcDirectory, namespaces)
+	if err != nil {
+		node.Status.Conditions = types.SetCondition(node.Status.Conditions, longhorn.NodeConditionTypeNFSClientInstalled, longhorn.ConditionStatusFalse,
+			string(longhorn.NodeConditionReasonNamespaceExecutorErr),
+			fmt.Sprintf("Failed to get namespace executor: %v", err.Error()))
+		return
+	}
+
+	kernelConfigPath := "/boot/config-" + kernelVersion
+	args := []string{kernelConfigPath}
+	if _, err := nsexec.Execute(nil, "ls", args, lhtypes.ExecuteDefaultTimeout); err != nil {
+		node.Status.Conditions = types.SetCondition(node.Status.Conditions, longhorn.NodeConditionTypeNFSClientInstalled, longhorn.ConditionStatusFalse,
+			string(longhorn.NodeConditionReasonKernelConfigIsNotFound),
+			fmt.Sprintf("Unable to find %v for checking %v: %v", kernelConfigPath, nfsClientVersions, err.Error()))
+		return
+	}
+
+	for _, ver := range nfsClientVersions {
+		args := []string{ver + "=", kernelConfigPath}
+		result, err := nsexec.Execute(nil, "grep", args, lhtypes.ExecuteDefaultTimeout)
+		if err != nil {
+			nc.logger.WithError(err).Debugf("Failed to find kernel config %v on node %v", ver, node.Name)
+			continue
+		}
+		enabled := strings.TrimSpace(strings.Split(result, "=")[1])
+		switch enabled {
+		case "y":
+			node.Status.Conditions = types.SetCondition(node.Status.Conditions, longhorn.NodeConditionTypeNFSClientInstalled, longhorn.ConditionStatusTrue, "", "")
+			return
+		case "m":
+			kmodResult, err := lhexec.NewExecutor().Execute(nil, "kmod", []string{"list"}, lhtypes.ExecuteDefaultTimeout)
+			if err != nil {
+				node.Status.Conditions = types.SetCondition(node.Status.Conditions, longhorn.NodeConditionTypeNFSClientInstalled, longhorn.ConditionStatusFalse,
+					string(longhorn.NodeConditionReasonNFSClientIsNotFound),
+					fmt.Sprintf("Failed to execute command `kmod`: %v", err.Error()))
+				return
+			}
+			res, err := lhexec.NewExecutor().ExecuteWithStdinPipe("grep", []string{"nfs"}, kmodResult, lhtypes.ExecuteDefaultTimeout)
+			if err != nil {
+				node.Status.Conditions = types.SetCondition(node.Status.Conditions, longhorn.NodeConditionTypeNFSClientInstalled, longhorn.ConditionStatusFalse,
+					string(longhorn.NodeConditionReasonNFSClientIsNotFound),
+					fmt.Sprintf("Failed to execute command `grep`: %v", err.Error()))
+				return
+			}
+			if res != "" {
+				node.Status.Conditions = types.SetCondition(node.Status.Conditions, longhorn.NodeConditionTypeNFSClientInstalled, longhorn.ConditionStatusTrue, "", "")
+				return
+			}
+		default:
+			nc.logger.Debugf("Unknown kernel config value for %v: %v", ver, enabled)
+		}
+	}
+
+	node.Status.Conditions = types.SetCondition(node.Status.Conditions, longhorn.NodeConditionTypeNFSClientInstalled, longhorn.ConditionStatusFalse,
+		string(longhorn.NodeConditionReasonNFSClientIsNotFound),
+		fmt.Sprintf("NFS clients %v not found. At least one should be enabled", nfsClientVersions))
 }
 
 func (nc *NodeController) getImTypeDataEngines(node *longhorn.Node) map[longhorn.InstanceManagerType][]longhorn.DataEngineType {
--- a/controller/node_controller_test.go
+++ b/controller/node_controller_test.go
@@ -184,6 +184,9 @@ func (s *NodeControllerSuite) TestManagerPodUp(c *C) {
 					newNodeCondition(longhorn.NodeConditionTypeSchedulable, longhorn.ConditionStatusTrue, ""),
 					newNodeCondition(longhorn.NodeConditionTypeReady, longhorn.ConditionStatusTrue, ""),
 					newNodeCondition(longhorn.NodeConditionTypeMountPropagation, longhorn.ConditionStatusTrue, ""),
+					newNodeCondition(longhorn.NodeConditionTypeRequiredPackages, longhorn.ConditionStatusFalse, longhorn.NodeConditionReasonUnknownOS),
+					newNodeCondition(longhorn.NodeConditionTypeMultipathd, longhorn.ConditionStatusTrue, ""),
+					newNodeCondition(longhorn.NodeConditionTypeNFSClientInstalled, longhorn.ConditionStatusFalse, longhorn.NodeConditionReasonKernelConfigIsNotFound),
 				},
 			},
 			TestNode2: {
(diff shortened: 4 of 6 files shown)


app/daemon.go
```diff
@@ -254,6 +309,158 @@ func environmentCheck() error {
...
+	nfsClientVersions := []string{"CONFIG_NFS_V4_2", "CONFIG_NFS_V4_1", "CONFIG_NFS_V4"}
+
+	nsexec, err := lhns.NewNamespaceExecutor(lhtypes.ProcessNone, lhtypes.HostProcDirectory, namespaces)
+	if err != nil {
+		return err
+	}
+
+	kernelConfigPath := "/boot/config-" + kernelVersion
+	args := []string{kernelConfigPath}
+	if _, err := nsexec.Execute(nil, "ls", args, lhtypes.ExecuteDefaultTimeout); err != nil {
```
c3y1huang, line 437: Need to be careful about invoking binaries with `nsenter`, this might be blocking for Talos Linux.

c3y1huang: Consider utilizing [ReadDirectory](https://github.com/longhorn/go-common-libs/blob/main/ns/file.go#L103).

mantissahz: It is great, thank you.
That is a file, I would use [GetFileInfo](https://github.com/longhorn/go-common-libs/blob/70d1933bfa908d89411ff553ce43d3b051db6003/ns/file.go#L166C6-L166C17).

app/daemon.go
```diff
@@ -130,15 +139,15 @@ func startManager(c *cli.Context) error {
...
-		return errors.Wrap(err, "Failed environment check, please make sure you have iscsiadm/open-iscsi installed on the host")
-	}
-
 	currentNodeID, err := util.GetRequiredEnv(types.EnvNodeName)
 	if err != nil {
 		return fmt.Errorf("failed to detect the node name")
 	}
 
+	if err := environmentCheck(kubeconfigPath, currentNodeID); err != nil {
+		return errors.Wrap(err, "Failed environment check, please make sure you have iscsiadm/open-iscsi installed on the host")
```
james-munson, line 148: Since environmentCheck() checks everything, not just iscsi, this error message may be misleading.

james-munson (changes requested):
Just one nit.

app/daemon.go
derekbit: Agree with @james-munson's comment.
You can combine the errors in `environmentCheck` and just do
```
return errors.Wrap(err, "Failed environment check")
```
here.

app/daemon.go
```diff
@@ -244,8 +253,54 @@ func startManager(c *cli.Context) error {
 	return nil
 }
 
-func environmentCheck() error {
+func environmentCheck(kubeconfigPath, currentNodeID string) error {
```
derekbit, line 256: nit:
```suggestion
func environmentCheck(kubeconfigPath, nodeID string) error {
```

app/daemon.go
```diff
@@ -244,8 +253,54 @@ func startManager(c *cli.Context) error {
...
+		return err
+	}
+
+	kubeNode, err := kubeClient.CoreV1().Nodes().Get(context.Background(), currentNodeID, metav1.GetOptions{})
+	if err != nil {
+		return errors.Wrap(err, "failed to get node when checking kernel version")
+	}
+
+	// Check if kernel version is recommended
+	if err := checkKernelVersion(kubeNode, currentNodeID); err != nil {
```
derekbit, line 284: `kubeNode` already contains the `nodeID`. Can we remove the parameter `currentNodeID`?

```suggestion
	if err := checkKernelVersion(kubeNode); err != nil {
```

app/daemon.go
```diff
@@ -244,8 +253,54 @@ func startManager(c *cli.Context) error {
...
+		return err
+	}
+
 	namespaces := []lhtypes.Namespace{lhtypes.NamespaceMnt, lhtypes.NamespaceNet}
+	// Check if necessary packages are installed
+	if err := checkPackagesInstalled(kubeNode, namespaces, currentNodeID); err != nil {
```
derekbit, line 290: Same as above

app/daemon.go
```diff
@@ -244,8 +253,54 @@ func startManager(c *cli.Context) error {
...
+		return err
+	}
+
+	// Check if multipath is supported
+	if err := checkMultipathd(namespaces, currentNodeID); err != nil {
+		return err
+	}
+
+	// Check if nfs client versions are supported
+	if err := checkNFSClientVersion(kubeNode, namespaces, currentNodeID); err != nil {
```
derekbit, line 300: Same as above

app/daemon.go
```diff
@@ -254,6 +309,158 @@ func environmentCheck() error {
...
+			opts := []string{"|grep", "nfs"}
+			if _, err := nsexec.Execute(nil, "lsmod", opts, lhtypes.ExecuteDefaultTimeout); err != nil {
+				logrus.WithError(err).Warnf("Failed to check %v enabled on node %v", ver, currentNodeID)
+			}
+		default:
+			logrus.Warnf("Unknown kernel config value for %v: %v", ver, enabled)
+		}
+	}
+
+	return fmt.Errorf("NFS clients %v not found. At least one should be enabled", nfsClientVersions)
```
derekbit, line 461: If a cluster doesn't require RWX volumes, then NFS server and client are not necessary. Should we error out in this use case?

app/daemon.go
mantissahz: Modified. Thanks.

app/daemon.go
mantissahz: Modified. Thanks.

app/daemon.go
mantissahz: Modified. Thanks.

app/daemon.go
mantissahz: Modified. Thanks.

app/daemon.go
mantissahz: Modified. Thanks.

app/daemon.go
mantissahz: Modified. Thanks.

app/daemon.go
mantissahz: Modified. Thanks.

derekbit:
The check is enforced when starting the Longhorn Manager daemon. If any condition fails, the Longhorn Manager will enter a loop of Starting and CrashLoopBackOff.

For users who only require RWO volumes, the NFS client is unnecessary. Then, the failure doesn't make sense.

Instead of performing checks only during startup, we can introduce conditions in `nodes.longhorn.io` to record the status of necessary packages, utilities, daemons. These conditions can be periodically checked by the node controller. This approach will help users stay informed and diagnose the Longhorn system effectively.

WDYT @shuo-wu @PhanLe1010 @innobead

derekbit:
The check is enforced when starting the Longhorn Manager daemon. If any condition fails, the Longhorn Manager will enter a loop of Starting and CrashLoopBackOff.

For users who only require RWO volumes, the NFS client is unnecessary. Then, the failure doesn't make sense.

Instead of performing checks only during startup, we can introduce conditions in nodes.longhorn.io to record the status of necessary packages, utilities, and demons. These conditions can be periodically checked by the node controller. This approach will help users stay informed and diagnose the Longhorn system effectively.

WDYT @c3y1huang @shuo-wu @PhanLe1010 @innobead

PhanLe1010:
> Instead of performing checks only during startup, we can introduce conditions in nodes.longhorn.io to record the status of necessary packages, utilities, and demons. These conditions can be periodically checked by the node controller. This approach will help users stay informed and diagnose the Longhorn system effectively.


I agree with this design! We already check the mountProagation and record it inside the node condition:
```
  status:
    autoEvicting: false
    conditions:
    - lastProbeTime: ""
      lastTransitionTime: "2024-06-13T02:22:19Z"
      message: ""
      reason: ""
      status: "True"
      type: MountPropagation
```

We can do something similar for this new `EnviromentCheck` condition

mantissahz:
Hi @derekbit, @PhanLe1010, 
I have updated the PR to use the node conditions to record the environment setting.
Could you review the PR when you are available?

derekbit:
cc @ChanYiLin for helping review it if having time. Thank you.

controller/node_controller.go
```diff
@@ -558,6 +561,11 @@ func (nc *NodeController) syncNode(key string) (err error) {
 		}
 	}
 
+	// check if environment settings meet the requirements on current node
+	if err := nc.environmentCheck(kubeNode, node); err != nil {
+		return err
+	}
+
```
derekbit, line 568: The error will hinder the remaining tasks. If we upate the error to conditions, we don't need to error out.

controller/node_controller.go
```diff
@@ -995,6 +1003,178 @@ func (nc *NodeController) syncNodeStatus(pod *corev1.Pod, node *longhorn.Node) e
...
+	node.Status.Conditions = types.SetCondition(node.Status.Conditions, longhorn.NodeConditionTypeRequiredPackages, longhorn.ConditionStatusTrue, "", "")
+	return nil
+}
+
+func (nc *NodeController) syncMultipathd(node *longhorn.Node, namespaces []lhtypes.Namespace) error {
+	nsexec, err := lhns.NewNamespaceExecutor(lhtypes.ProcessNone, lhtypes.HostProcDirectory, namespaces)
+	if err != nil {
+		return err
+	}
```
derekbit, line 1119: We can log the error to NodeConditionTypeMultipathd condition.

controller/node_controller.go
```diff
@@ -995,6 +1003,178 @@ func (nc *NodeController) syncNodeStatus(pod *corev1.Pod, node *longhorn.Node) e
...
+	return nil
+}
+
+func (nc *NodeController) syncNFSClientVersion(kubeNode *corev1.Node, node *longhorn.Node, namespaces []lhtypes.Namespace) error {
+	kernelVersion := kubeNode.Status.NodeInfo.KernelVersion
+	nfsClientVersions := []string{"CONFIG_NFS_V4_2", "CONFIG_NFS_V4_1", "CONFIG_NFS_V4"}
+
+	nsexec, err := lhns.NewNamespaceExecutor(lhtypes.ProcessNone, lhtypes.HostProcDirectory, namespaces)
+	if err != nil {
+		return err
```
derekbit, line 1138: We can log the error to NodeConditionTypeNFSClientInstalled condition.

controller/node_controller.go
```diff
@@ -995,6 +1003,178 @@ func (nc *NodeController) syncNodeStatus(pod *corev1.Pod, node *longhorn.Node) e
...
+	default:
+		node.Status.Conditions = types.SetCondition(node.Status.Conditions, longhorn.NodeConditionTypeRequiredPackages, longhorn.ConditionStatusFalse,
+			string(longhorn.NodeConditionReasonUnknownOS),
+			fmt.Sprintf("Node %v has an unknown OS image %v, please make sure the necessary packages are installed", node.Name, osImage))
+		return nil
+	}
+
+	nsexec, err := lhns.NewNamespaceExecutor(lhtypes.ProcessNone, lhtypes.HostProcDirectory, namespaces)
+	if err != nil {
+		return err
```
derekbit, line 1093: We can log the error to NodeConditionTypeRequiredPackages condition.

controller/node_controller.go
```diff
@@ -995,6 +1003,178 @@ func (nc *NodeController) syncNodeStatus(pod *corev1.Pod, node *longhorn.Node) e
...
+
+	brokenFlag := false
+	for _, version := range brokenKernelVersion {
+		if !strings.Contains(kernelVersion, version) {
+			continue
+		}
+		var kernel, major, minor, patch int
+		if _, err := fmt.Sscanf(kernelVersion, "%d.%d.%d-%d", &kernel, &major, &minor, &patch); err != nil {
+			return errors.Wrap(err, "failed to parse kernel version "+kernelVersion)
+		}
```
derekbit, line 1044: log the error in the condition. WDYT?

controller/node_controller.go
```diff
@@ -995,6 +1001,169 @@ func (nc *NodeController) syncNodeStatus(pod *corev1.Pod, node *longhorn.Node) e
...
+	brokenKernel5150PatchVersion := 94
+	brokenKernel656PatchVersion := 6
+	fixedKernel5150PatchVersion := 100
+
+	for _, version := range brokenKernelVersion {
+		if !strings.Contains(kernelVersion, version) {
+			continue
+		}
+		var kernel, major, minor, patch int
+		if _, err := fmt.Sscanf(kernelVersion, "%d.%d.%d-%d", &kernel, &major, &minor, &patch); err != nil {
```
derekbit, line 1026: Some distros don't have `-%d`. Suggest using regexp.
An example https://goplay.tools/snippet/EyotP7UfLSy

controller/node_controller.go
```diff
@@ -995,6 +1001,169 @@ func (nc *NodeController) syncNodeStatus(pod *corev1.Pod, node *longhorn.Node) e
...
+			string(longhorn.NodeConditionReasonKernelConfigIsNotFound),
+			fmt.Sprintf("Failed to check %v, because %v does not exist on node %v: %v", nfsClientVersions, kernelConfigPath, node.Name, err))
+		return
+	}
+
+	for _, ver := range nfsClientVersions {
+		args := []string{ver + "=", kernelConfigPath}
+		result, err := nsexec.Execute(nil, "grep", args, lhtypes.ExecuteDefaultTimeout)
+		if err != nil {
+			logrus.WithError(err).Warnf("Failed to find kernel config %v on node %v", ver, node.Name)
```
derekbit, line 1144: Change `Warnf` to `Trancef` or `Debugf`, or it will lead to flooding warning messages

controller/node_controller.go
```diff
@@ -995,6 +1001,169 @@ func (nc *NodeController) syncNodeStatus(pod *corev1.Pod, node *longhorn.Node) e
...
+			node.Status.Conditions = types.SetCondition(node.Status.Conditions, longhorn.NodeConditionTypeNFSClientInstalled, longhorn.ConditionStatusTrue, "", "")
+			return
+		case "m":
+			opts := []string{"|grep", "nfs"}
+			if _, err := nsexec.Execute(nil, "lsmod", opts, lhtypes.ExecuteDefaultTimeout); err == nil {
+				node.Status.Conditions = types.SetCondition(node.Status.Conditions, longhorn.NodeConditionTypeNFSClientInstalled, longhorn.ConditionStatusTrue, "", "")
+				return
+			}
+		default:
+			logrus.Warnf("Unknown kernel config value for %v: %v", ver, enabled)
```
derekbit, line 1158: Change Warnf to Trancef or Debugf, or it will lead to flooding warning messages

controller/node_controller.go
```diff
@@ -995,6 +1001,169 @@ func (nc *NodeController) syncNodeStatus(pod *corev1.Pod, node *longhorn.Node) e
...
+		node.Status.Conditions = types.SetCondition(node.Status.Conditions, longhorn.NodeConditionTypeRequiredPackages, longhorn.ConditionStatusFalse,
+			string(longhorn.NodeConditionReasonNamespaceExecutorErr),
+			fmt.Sprintf("Get namespace executor failed: %v", err))
+		return
+	}
+	var NotFoundPkgs []string
+	for _, pkg := range packages {
+		args := []string{pkg}
+		if _, err := nsexec.Execute(nil, queryPackagesCmd, args, lhtypes.ExecuteDefaultTimeout); err != nil {
+			logrus.WithError(err).Warnf("Package %v is not found in node %v", pkg, node.Name)
```
derekbit, line 1085: Change Warnf to Trancef or Debugf, or it will lead to flooding warning messages

controller/node_controller.go
mantissahz: Removed returning errors. Thanks

controller/node_controller.go
mantissahz: Logged the error to NodeConditionTypeNFSClientInstalled condition. Thanks

controller/node_controller.go
mantissahz: Logged the error to NodeConditionTypeRequiredPackages condition. Thanks

controller/node_controller.go
mantissahz: Logged the error to NodeConditionTypeSupportedKernel condition. Thanks

controller/node_controller.go
mantissahz: Logged the error to the condition. Thanks

derekbit:
There is an exsting environment check for open-iscsi in the app/daemon.go. We already have a condition in node.longhorn.io for it. Do we still need to error out if open-iscsi is missing while staring up the daemon?

cc @PhanLe1010 @shuo-wu @innobead WDYT?

controller/node_controller.go
mantissahz: Used `regex` to parse the kernel version. Thanks.

controller/node_controller.go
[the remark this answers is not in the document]
mantissahz: Yes, it is. Added `continue`. Thanks

mantissahz: Modified to `Debugf`. Thanks.

controller/node_controller.go
mantissahz: Modified to `Debugf`. Thanks.

controller/node_controller.go
mantissahz: Modified to `Debugf`. Thanks.

app/pre_upgrade.go
```diff
@@ -56,5 +56,9 @@ func preUpgrade(c *cli.Context) error {
 		return err
 	}
 
+	if err := environmentCheck(); err != nil {
+		return errors.Wrap(err, "Failed environment check, please make sure you have iscsiadm/open-iscsi installed on the host")
```
derekbit, line 60:
```suggestion
		return errors.Wrap(err, "failed to check environment, please make sure you have iscsiadm/open-iscsi installed on the host")
```

mantissahz: Modified. Thanks.

controller/node_controller.go
```diff
@@ -46,6 +51,13 @@ const (
 	snapshotChangeEventQueueMax = 1048576
 )
 
+type kernelVersion struct {
```
derekbit, line 54: What's the patch version if version is `5.4.2-123-aws`?

controller/node_controller.go
```diff
@@ -995,6 +1010,199 @@ func (nc *NodeController) syncNodeStatus(pod *corev1.Pod, node *longhorn.Node) e
...
+		if !strings.Contains(kernelVersion, version) {
+			continue
+		}
+		kVersion, err := parseKernelVersion(kernelVersion)
+		if err != nil {
+			node.Status.Conditions = types.SetCondition(node.Status.Conditions, longhorn.NodeConditionTypeSupportedKernel, longhorn.ConditionStatusFalse,
+				string(longhorn.NodeConditionReasonKernelNotSupport),
+				fmt.Sprintf("Failed to parse kernel version %v: %v", kernelVersion, err))
+			return
+		}
```
derekbit, line 1040: Move the parsing logic outside of the for loop.
It only needs only be parsed once.

derekbit:
Ther kernel check function is not easy to read and maintain.
What we need to check is. the four kernels

Distro | Broken Version
-- | --
Vanilla kernel | 6.5.6
Ubuntu | 5.15.0-94
Ubuntu | 6.5.0-21
Ubuntu | 6.5.0-1014-aws

Can we check it use string compare?

controller/node_controller.go
mantissahz: That would be `kernel=5, major=4, minor=2, patch=123`

mantissahz: Removed the kernel check.

controller/node_controller.go
mantissahz: Removed the kernel check.

derekbit:
As discussed with @mantissahz, we decide to remove kernel version check because
- The current implementation is not easy to maintain. We need to figure out a better way to do it.
- We are unable to include all problematic kernel versions in the check.

controller/node_controller.go
```diff
@@ -995,6 +1001,139 @@ func (nc *NodeController) syncNodeStatus(pod *corev1.Pod, node *longhorn.Node) e
...
+	case strings.Contains(osImage, "arch"):
+		queryPackagesCmd = "pacman -Q"
+		packages = []string{"nfs-utils", "open-iscsi", "cryptsetup", "device-mapper"}
+	case strings.Contains(osImage, "gentoo"):
+		queryPackagesCmd = "qlist -I"
+		packages = []string{"net-fs/nfs-utils", "sys-block/open-iscsi", "sys-fs/cryptsetup", "sys-fs/lvm2"}
+	default:
+		node.Status.Conditions = types.SetCondition(node.Status.Conditions, longhorn.NodeConditionTypeRequiredPackages, longhorn.ConditionStatusFalse,
+			string(longhorn.NodeConditionReasonUnknownOS),
+			fmt.Sprintf("Node %v has an unknown OS image %v, please make sure the necessary packages are installed", node.Name, osImage))
```
derekbit, line 1039:
```suggestion
			fmt.Sprintf("Unable to verify the required packages because the OS image '%v' is unknown to the Longhorn system. Please ensure the required packages are installed.", osImage))
```

controller/node_controller.go
```diff
@@ -995,6 +1001,139 @@ func (nc *NodeController) syncNodeStatus(pod *corev1.Pod, node *longhorn.Node) e
...
+		return
+	}
+
+	nsexec, err := lhns.NewNamespaceExecutor(lhtypes.ProcessNone, lhtypes.HostProcDirectory, namespaces)
+	if err != nil {
+		node.Status.Conditions = types.SetCondition(node.Status.Conditions, longhorn.NodeConditionTypeRequiredPackages, longhorn.ConditionStatusFalse,
+			string(longhorn.NodeConditionReasonNamespaceExecutorErr),
+			fmt.Sprintf("Get namespace executor failed: %v", err))
```
derekbit, line 1047:
```suggestion
			fmt.Sprintf("Failed to get namespace executor: %v", err.Error()))
```
Same for other messages.

controller/node_controller.go
```diff
@@ -995,6 +1001,139 @@ func (nc *NodeController) syncNodeStatus(pod *corev1.Pod, node *longhorn.Node) e
...
+		if _, err := nsexec.Execute(nil, queryPackagesCmd, args, lhtypes.ExecuteDefaultTimeout); err != nil {
+			nc.logger.WithError(err).Debugf("Package %v is not found in node %v", pkg, node.Name)
+			NotFoundPkgs = append(NotFoundPkgs, pkg)
+		}
+	}
+
+	if len(NotFoundPkgs) > 0 {
+		node.Status.Conditions = types.SetCondition(node.Status.Conditions, longhorn.NodeConditionTypeRequiredPackages, longhorn.ConditionStatusFalse,
+			string(longhorn.NodeConditionReasonPackagesNotInstalled),
+			fmt.Sprintf("Node %v is missing the following packages: %v, please make sure they are installed", node.Name, NotFoundPkgs))
```
derekbit, line 1062: Node name is unnecessary because the condition is under node.status.

```suggestion
			fmt.Sprintf("Missing packages: %v", NotFoundPkgs))
```

controller/node_controller.go
```diff
@@ -995,6 +1001,139 @@ func (nc *NodeController) syncNodeStatus(pod *corev1.Pod, node *longhorn.Node) e
...
+
+	node.Status.Conditions = types.SetCondition(node.Status.Conditions, longhorn.NodeConditionTypeRequiredPackages, longhorn.ConditionStatusTrue, "", "")
+}
+
+func (nc *NodeController) syncMultipathd(node *longhorn.Node, namespaces []lhtypes.Namespace) {
+	nsexec, err := lhns.NewNamespaceExecutor(lhtypes.ProcessNone, lhtypes.HostProcDirectory, namespaces)
+	if err != nil {
+		node.Status.Conditions = types.SetCondition(node.Status.Conditions, longhorn.NodeConditionTypeMultipathd, longhorn.ConditionStatusFalse,
+			string(longhorn.NodeConditionReasonNamespaceExecutorErr),
+			fmt.Sprintf("Get namespace executor failed: %v", err))
```
derekbit, line 1074:
```suggestion
			fmt.Sprintf("Failed to get namespaced executor: %v", err.Error()))
```

controller/node_controller.go
```diff
@@ -995,6 +1001,139 @@ func (nc *NodeController) syncNodeStatus(pod *corev1.Pod, node *longhorn.Node) e
...
+		return
+	}
+	args := []string{"show", "status"}
+	if result, _ := nsexec.Execute(nil, "multipathd", args, lhtypes.ExecuteDefaultTimeout); result != "" {
+		node.Status.Conditions = types.SetCondition(node.Status.Conditions, longhorn.NodeConditionTypeMultipathd, longhorn.ConditionStatusFalse,
+			string(longhorn.NodeConditionReasonMultipathdIsRunning),
+			fmt.Sprintf("Multipathd is running on %v known to have a breakage that affects Longhorn. See description and solution at https://longhorn.io/kb/troubleshooting-volume-with-multipath", node.Name))
```
derekbit, line 1081:
```suggestion
			fmt.Sprintf("multipathd is running with a known issue that affects Longhorn. See description and solution at https://longhorn.io/kb/troubleshooting-volume-with-multipath", node.Name))
```

derekbit: If a user applies the workaround, the condition is still `false`. We can improve it in the future. Can you create a ticket for the improvement?

controller/node_controller.go
mantissahz: Modified. Thanks

controller/node_controller.go
mantissahz: Modified. Thanks

controller/node_controller.go
mantissahz: Modified. Thanks

controller/node_controller.go
mantissahz: Modified. Thanks

controller/node_controller.go
```diff
@@ -995,6 +1001,139 @@ func (nc *NodeController) syncNodeStatus(pod *corev1.Pod, node *longhorn.Node) e
...
+	}
+
+	nsexec, err := lhns.NewNamespaceExecutor(lhtypes.ProcessNone, lhtypes.HostProcDirectory, namespaces)
+	if err != nil {
+		node.Status.Conditions = types.SetCondition(node.Status.Conditions, longhorn.NodeConditionTypeRequiredPackages, longhorn.ConditionStatusFalse,
+			string(longhorn.NodeConditionReasonNamespaceExecutorErr),
+			fmt.Sprintf("Failed to get namespace executor: %v", err.Error()))
+		return
+	}
+	var NotFoundPkgs []string
```
derekbit, line 1050: Require initialization?
```
NotFoundPkgs := []string{}
```

controller/node_controller.go
mantissahz: Could we handle it with this ticket https://github.com/longhorn/longhorn/issues/6755

controller/node_controller.go
```diff
@@ -995,6 +1001,139 @@ func (nc *NodeController) syncNodeStatus(pod *corev1.Pod, node *longhorn.Node) e
...
+func (nc *NodeController) syncMultipathd(node *longhorn.Node, namespaces []lhtypes.Namespace) {
+	nsexec, err := lhns.NewNamespaceExecutor(lhtypes.ProcessNone, lhtypes.HostProcDirectory, namespaces)
+	if err != nil {
+		node.Status.Conditions = types.SetCondition(node.Status.Conditions, longhorn.NodeConditionTypeMultipathd, longhorn.ConditionStatusFalse,
+			string(longhorn.NodeConditionReasonNamespaceExecutorErr),
+			fmt.Sprintf("Failed to get namespace executor: %v", err.Error()))
+		return
+	}
+	args := []string{"show", "status"}
+	if result, _ := nsexec.Execute(nil, "multipathd", args, lhtypes.ExecuteDefaultTimeout); result != "" {
```
derekbit, line 1078: What does `result` mean? 
Do we need to check `err` of ns.exec.Execute()?

controller/node_controller.go
```diff
@@ -995,6 +1001,139 @@ func (nc *NodeController) syncNodeStatus(pod *corev1.Pod, node *longhorn.Node) e
...
+
+func (nc *NodeController) syncNFSClientVersion(kubeNode *corev1.Node, node *longhorn.Node, namespaces []lhtypes.Namespace) {
+	kernelVersion := kubeNode.Status.NodeInfo.KernelVersion
+	nfsClientVersions := []string{"CONFIG_NFS_V4_2", "CONFIG_NFS_V4_1", "CONFIG_NFS_V4"}
+
+	nsexec, err := lhns.NewNamespaceExecutor(lhtypes.ProcessNone, lhtypes.HostProcDirectory, namespaces)
+	if err != nil {
+		node.Status.Conditions = types.SetCondition(node.Status.Conditions, longhorn.NodeConditionTypeNFSClientInstalled, longhorn.ConditionStatusFalse,
+			string(longhorn.NodeConditionReasonNamespaceExecutorErr),
+			fmt.Sprintf("Get namespace executor failed: %v", err))
```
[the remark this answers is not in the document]
mantissahz: Modified. Thanks

controller/node_controller.go
```diff
@@ -995,6 +1001,139 @@ func (nc *NodeController) syncNodeStatus(pod *corev1.Pod, node *longhorn.Node) e
...
+		return
+	}
+
+	kernelConfigPath := "/boot/config-" + kernelVersion
+	args := []string{kernelConfigPath}
+	if _, err := nsexec.Execute(nil, "ls", args, lhtypes.ExecuteDefaultTimeout); err != nil {
+		node.Status.Conditions = types.SetCondition(node.Status.Conditions, longhorn.NodeConditionTypeNFSClientInstalled, longhorn.ConditionStatusFalse,
+			string(longhorn.NodeConditionReasonKernelConfigIsNotFound),
+			fmt.Sprintf("Failed to check %v, because %v does not exist on node %v: %v", nfsClientVersions, kernelConfigPath, node.Name, err))
```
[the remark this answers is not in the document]
mantissahz: Modified. Thanks

controller/node_controller.go
```diff
@@ -995,6 +1001,139 @@ func (nc *NodeController) syncNodeStatus(pod *corev1.Pod, node *longhorn.Node) e
...
+			fmt.Sprintf("Failed to get namespace executor: %v", err.Error()))
+		return
+	}
+
+	kernelConfigPath := "/boot/config-" + kernelVersion
+	args := []string{kernelConfigPath}
+	if _, err := nsexec.Execute(nil, "ls", args, lhtypes.ExecuteDefaultTimeout); err != nil {
+		node.Status.Conditions = types.SetCondition(node.Status.Conditions, longhorn.NodeConditionTypeNFSClientInstalled, longhorn.ConditionStatusFalse,
+			string(longhorn.NodeConditionReasonKernelConfigIsNotFound),
+			fmt.Sprintf("Failed to check %v, because %v does not exist: %v", nfsClientVersions, kernelConfigPath, err.Error()))
```
derekbit, line 1105:
```suggestion
			fmt.Sprintf("Unable to find %v for checking %v: %v", kernelConfigPath, nfsClientVersions, err.Error()))
```

controller/node_controller.go
```diff
@@ -995,6 +1001,139 @@ func (nc *NodeController) syncNodeStatus(pod *corev1.Pod, node *longhorn.Node) e
...
+			continue
+		}
+		enabled := strings.TrimSpace(strings.Split(result, "=")[1])
+		switch enabled {
+		case "y":
+			node.Status.Conditions = types.SetCondition(node.Status.Conditions, longhorn.NodeConditionTypeNFSClientInstalled, longhorn.ConditionStatusTrue, "", "")
+			return
+		case "m":
+			opts := []string{"|grep", "nfs"}
+			if _, err := nsexec.Execute(nil, "lsmod", opts, lhtypes.ExecuteDefaultTimeout); err == nil {
```
derekbit, line 1123: What happens if `lsmod` is no found in the host side?
Can we install `kmod` in longhorn-manager and check the modules inside the pod?

controller/node_controller.go
derekbit: Sure. Please a comment https://github.com/longhorn/longhorn/issues/6755.

controller/node_controller.go
mantissahz: >What happens if lsmod is no found in the host side?

An error string "No such file or directory" or "Command 'lsmod' not found" in returning `err`.

>nit: can we install kmod in longhorn-manager and check the modules inside the pod? we can improve it later.

Yes, `kmod` can get the module status of the worker node in the longhorn-manager pod.

controller/node_controller.go
mantissahz: Modified. Thanks.

controller/node_controller.go
mantissahz: Modified. Thanks.

controller/node_controller.go
mantissahz: Modified to use 'kmod' command in the longhorn-manager pod.

derekbit (approved):

derekbit:
In general, LGTM.

PR #2764, opened 2024-04-29
Fall back to a running instance-manager if a default is not available

derekbit:

Issue Longhorn/longhorn#6001

Some operations in controllers, such as deleting disks and reconciling backup targets, initially fetch the default instance-manager. However, these operations are not restricted to the default instance-manager; they can use any running instance-manager. To enhance resilience, if the default instance-manager is unable to start before an upgrade of the v2 data engine or for any other reason for both v1 and v2 data engines, the controller can switch to any running instance-manager as a fallback.


datastore/longhorn.go
```diff
@@ -5061,3 +5061,28 @@ func (s *DataStore) ListBackupBackingImages() (map[string]*longhorn.BackupBackin
...
+	ims, err := s.ListInstanceManagersByNodeRO(node, longhorn.InstanceManagerTypeAllInOne, dataEngine)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to list instance managers for node %v", node)
+	}
+
+	for _, im := range ims {
+		if im.Status.CurrentState == longhorn.InstanceManagerStateRunning {
+			return im, nil
+		}
+	}
```
c3y1huang, line 5085: Would it be beneficial to include a log to keep a trace of whether Longhorn is using the default or falling back to any running IM?

derekbit: Adding a log message sounds good. Added, thank you.

c3y1huang (approved):

innobead (approved):
LGTM. 

Would like to know how QA tests this.

PR #2770, opened 2024-05-03
fix(webhook): disable tsl 1.0 and 1.1 on webhook service

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/8387

Follow the same implementation in [Harvester](https://github.com/harvester/harvester/blob/68c52fa87b44906b7915e81fb674fb73c1d41951/pkg/server/types.go#L165-L175) to disable tls 1.0 and 1.1 on webhook service

### DIFF
--- a/webhook/server/server.go
+++ b/webhook/server/server.go
@@ -2,6 +2,7 @@ package server
 
 import (
 	"context"
+	"crypto/tls"
 	"fmt"
 	"net/http"
 	"reflect"
@@ -31,6 +32,17 @@ var (
 	matchPolicyExact = admissionregv1.Exact // nolint: unused
 
 	sideEffectClassNone = admissionregv1.SideEffectClassNone
+)
+
+var (
+	whiteListedCiphers = []uint16{
+		tls.TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,
+		tls.TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,
+		tls.TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,
+		tls.TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,
+		tls.TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305,
+		tls.TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305,
+	}
 )
 
 type WebhookServer struct {
@@ -178,6 +190,10 @@ func (s *WebhookServer) runAdmissionWebhookListenAndServe(handler http.Handler, 
 				tlsName,
 			},
 			FilterCN: dynamiclistener.OnlyAllow(tlsName),
+			TLSConfig: &tls.Config{
+				MinVersion:   tls.VersionTLS12,
+				CipherSuites: whiteListedCiphers,
+			},
 		},
 	})
 }
@@ -237,6 +253,10 @@ func (s *WebhookServer) runConversionWebhookListenAndServe(handler http.Handler,
 				tlsName,
 			},
 			FilterCN: dynamiclistener.OnlyAllow(tlsName),
+			TLSConfig: &tls.Config{
+				MinVersion:   tls.VersionTLS12,
+				CipherSuites: whiteListedCiphers,
+			},
 		},
 	})
 }


c3y1huang (commented):
I've run the `sslscan` on both v1.6.1 and the master branch clusters.

Interestingly, there is a difference in the results without applying the fix in this PR.

In v1.6.1, as described, TLSv1.0 and TLSv1.1 shows enabled:
```
> k -n longhorn-system get daemonsets.apps longhorn-manager -o yaml | grep image:
        image: longhornio/longhorn-manager:v1.6.1
   
> k get service -n longhorn-system | grep admission
longhorn-admission-webhook    ClusterIP   10.43.18.107    <none>        9502/TCP       2m17s

> k -n longhorn-system exec -it longhorn-manager-j9pdl -- sslscan 10.43.18.107:9502
Version: 2.0.16
OpenSSL 1.1.1l-fips  24 Aug 2021 SUSE release 150500.17.25.1

Connected to 10.43.18.107

Testing SSL server 10.43.18.107 on port 9502 using SNI name 10.43.18.107

  SSL/TLS Protocols:
SSLv2     disabled
SSLv3     disabled
TLSv1.0   enabled
TLSv1.1   enabled
TLSv1.2   enabled
TLSv1.3   enabled
```
However, on the master branch, TLSv1.0 and TLSv1.1 shows disabled (without this PR):
```
> k -n longhorn-system get daemonsets.apps longhorn-manager -o yaml | grep image:
        image: longhornio/longhorn-manager:master-head
    
> k get service -n longhorn-system | grep admission
longhorn-admission-webhook    ClusterIP   10.43.213.78    <none>        9502/TCP       39s

> k -n longhorn-system exec -it longhorn-manager-8ltwk -- sslscan 10.43.213.78:9502
Version: 2.0.16
OpenSSL 1.1.1w-fips  11 Sep 2023 SUSE release 150600.2.11

Connected to 10.43.213.78

Testing SSL server 10.43.213.78 on port 9502 using SNI name 10.43.213.78

  SSL/TLS Protocols:
SSLv2     disabled
SSLv3     disabled
TLSv1.0   disabled
TLSv1.1   disabled
TLSv1.2   enabled
TLSv1.3   enabled
```

So, I have 2 questions:
1. What is the cause of the different outcomes when the cluster is running with the master-head?
1. Is the purpose of this fix to address versions before v1.7? If so, any benefit for it to be implemented on the master branch?

c3y1huang (approved):
Apart from the [questions](https://github.com/longhorn/longhorn-manager/pull/2770#pullrequestreview-2037679066), the fix LGTM if the intention is to backport and introduce a whitelist approach.

derekbit (approved):

ChanYiLin:
Hi @c3y1huang 
I found the reason.
Because in v1.6.1, we use `go 1.21`, and the default `MinVersion=1.0`
Golang community update the `MinVersion to 1.2` in `go 1.22` [here](https://github.com/golang/go/issues/62459)

So for the second question,
we have to backport the fix by specifying the minVersion in the config.

c3y1huang:
> Hi @c3y1huang I found the reason. Because in v1.6.1, we use `go 1.21`, and the default `MinVersion=1.0` Golang community update the `MinVersion to 1.2` in `go 1.22` [here](https://github.com/golang/go/issues/62459)
> 
> So for the second question, we have to backport the fix by specifying the minVersion in the config.

If I understand correctly, updating to Go version 1.22 should resolve this issue. Is there any reason we are implementing a whitelist and the minVersion instead of updating the go version?

c3y1huang:
Hmm. mergify already merged this. @ChanYiLin please check to see if golang version upgrade is sufficient. If it is, I think we don't need this comment.

cc @derekbit

derekbit:
> Hmm. mergify already merged this. @ChanYiLin please check to see if golang version upgrade is sufficient. If it is, I think we don't need this PR.
> 
> cc @derekbit

@ChanYiLin Feel free to revert the PR if @c3y1huang's comment is valid. Thank you.

derekbit:
> > Hi @c3y1huang I found the reason. Because in v1.6.1, we use `go 1.21`, and the default `MinVersion=1.0` Golang community update the `MinVersion to 1.2` in `go 1.22` [here](https://github.com/golang/go/issues/62459)
> > So for the second question, we have to backport the fix by specifying the minVersion in the config.
> 
> If I understand correctly, updating to Go version 1.22 should resolve this issue. Is there any reason we are implementing a whitelist and the minVersion instead of updating the go version? Seems we've already done so in the master and 1.5.x branches.

BTW, we need to update v1.6.x's golang version.
https://github.com/longhorn/longhorn-manager/blob/v1.6.x/go.mod#L3

c3y1huang:
> BTW, we need to update v1.6.x's golang version. https://github.com/longhorn/longhorn-manager/blob/v1.6.x/go.mod#L3

https://github.com/longhorn/longhorn-manager/pull/2774

PR #2783, opened 2024-05-08
feat: remove related snapshot when removing backup

FrankYang0529:
https://github.com/longhorn/longhorn/issues/8365

When users create a Bakcup, LH automatically creates a related Snapshot. However, when users delete the Bakcup, LH doesn't cleanup the Snapshot. This occupies snapshot quota. We have snapshot space management now, so it's better to clean up related resources.

#### Test steps

Case 1: set `auto-cleanup-when-delete-backup` as `false`
1. Create a volume and setup backup target.
2. Mount the volume.
3. Create multiple backups.
4. Delete a backup and check related snapshot is not removed.

Case 2: set `auto-cleanup-when-delete-backup` as `true`
1. Create a volume and setup backup target.
2. Mount the volume.
3. Create multiple backups.
4. Delete a backup and check related snapshot is removed too.

### DIFF
--- a/controller/backup_controller.go
+++ b/controller/backup_controller.go
@@ -317,6 +317,31 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
 			bc.eventRecorder.Eventf(backup, corev1.EventTypeWarning, string(backup.Status.State), "Failed backup %s has been deleted: %s", backup.Name, backup.Status.Error)
 		}
 
+		autocleanup, err := bc.ds.GetSettingAsBool(types.SettingNameAutoCleanupSnapshotWhenDeleteBackup)
+		if err != nil {
+			logrus.WithError(err).WithFields(logrus.Fields{
+				"name": types.SettingNameAutoCleanupSnapshotWhenDeleteBackup,
+			}).Warn("Failed to get the setting")
+		}
+		if autocleanup {
+			// do the best effort to delete the snapshot
+			snapshot, err := bc.ds.GetSnapshotRO(backup.Spec.SnapshotName)
+			if err != nil {
+				if !apierrors.IsNotFound(err) {
+					logrus.WithError(err).WithFields(logrus.Fields{
+						"backup":   backup.Name,
+						"snapshot": snapshot.Name,
+					}).Warn("Failed to get snapshot")
+				}
+				return nil
+			}
+			if err = bc.ds.DeleteSnapshot(snapshot.Name); err != nil {
+				logrus.WithError(err).WithFields(logrus.Fields{
+					"backup":   backup.Name,
+					"snapshot": snapshot.Name,
+				}).Warn("Failed to delete snapshot")
+			}
+		}
 		return bc.ds.RemoveFinalizerForBackup(backup)
 	}
 
--- a/types/setting.go
+++ b/types/setting.go
@@ -131,6 +131,7 @@ const (
 	SettingNameV2DataEngineLogLevel                                     = SettingName("v2-data-engine-log-level")
 	SettingNameV2DataEngineLogFlags                                     = SettingName("v2-data-engine-log-flags")
 	SettingNameFreezeFilesystemForSnapshot                              = SettingName("freeze-filesystem-for-snapshot")
+	SettingNameAutoCleanupSnapshotWhenDeleteBackup                      = SettingName("auto-cleanup-when-delete-backup")
 )
 
 var (
@@ -218,6 +219,7 @@ var (
 		SettingNameAllowEmptyDiskSelectorVolume,
 		SettingNameDisableSnapshotPurge,
 		SettingNameFreezeFilesystemForSnapshot,
+		SettingNameAutoCleanupSnapshotWhenDeleteBackup,
 	}
 )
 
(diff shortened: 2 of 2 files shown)


controller/backup_controller.go
```diff
@@ -307,6 +307,20 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
 			bc.eventRecorder.Eventf(backup, corev1.EventTypeWarning, string(backup.Status.State), "Failed backup %s has been deleted: %s", backup.Name, backup.Status.Error)
 		}
 
+		snapshot, err := bc.ds.GetSnapshot(backup.Spec.SnapshotName)
+		if err != nil {
+			if !apierrors.IsNotFound(err) {
+				return err
+			}
```
derekbit, line 314: Does it need to error out or just do best effort?

FrankYang0529: Yes, we can do best effort for it. Thanks.

mantissahz:
Hi @FrankYang0529,

A question for this scenario:

1. A user created a snapshot A for reverting in the future.
2. A user created a backup A with this snapshot A for a restoration in the future or creating a DR volume on another cluster.
3. This PR will delete the snapshot A when the user switches/empties the backup target URL.

I'm not sure that users would like to delete the snapshot A in this situation.

FrankYang0529:
> Hi @FrankYang0529,
> 
> A question for this scenario:
> 
> 1. A user created a snapshot A for reverting in the future.
> 2. A user created a backup A with this snapshot A for a restoration in the future or creating a DR volume on another cluster.
> 3. This PR will delete the snapshot A when the user switches/empties the backup target URL.
> 
> I'm not sure that users would like to delete the snapshot A in this situation.

Hi @mantissahz, thanks for the feedback. IIRC, this PR doesn't break backup function, because we already put data to external storage. However, I'm not sure for DR volume. Do you know whether LH copy data to another cluster when users setup a DR volume?

Another question is that do we have a way to know whether a snapshot is created by a backup or by a user? Thanks.

mantissahz:
>However, I'm not sure for DR volume. Do you know whether LH copy data to another cluster when users setup a DR volume?

@FrankYang0529, the LH of another cluster will detect the new backup coming and synchronize the data from this new backup so it's fine if the data is not removed on the remote backup server.

>Another question is that do we have a way to know whether a snapshot is created by a backup or by a user? Thanks.

IIRC, No.

FrankYang0529:
> > However, I'm not sure for DR volume. Do you know whether LH copy data to another cluster when users setup a DR volume?
> 
> @FrankYang0529, the LH of another cluster will detect if there is a new backup and synchronize the data from this new backup so it's fine if the data is not removed on the remote backup server.
> 
> > Another question is that do we have a way to know whether a snapshot is created by a backup or by a user? Thanks.
> 
> IIRC, No.

Got it. For functionality, we don't break anything. However, for UX, we don't respect whether users want to keep the snapshot or not. I think we can add a setting for users to decide whether to remove snapshot when removing the backup. WDYT?

mantissahz:
>Got it. For functionality, we don't break anything. However, for UX, we don't respect whether users want to keep the snapshot or not. I think we can add a setting for users to decide whether to remove snapshot when removing the backup. WDYT?

Sure, would you like to have a global setting or a field in backup/snapshot resources?

innobead:
@mantissahz raised a good point about the dependency between snapshot and backup. They could be independent indeed. 

I agreed that we can at least have a setting to let users control whether they want to delete the snapshot used by a backup when the backup is deleted instead of blindly deleting it.

controller/backup_controller.go
```diff
@@ -307,6 +307,23 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
...
+		if err != nil {
+			if !apierrors.IsNotFound(err) {
+				logrus.WithError(err).WithFields(logrus.Fields{
+					"backup":   backup.Name,
+					"snapshot": snapshot.Name,
+				}).Warn("Failed to get snapshot")
+			}
+			return nil
+		}
+		if err = bc.ds.DeleteSnapshot(snapshot.Name); err != nil {
```
innobead, line 321: This is to trigger the snapshot deletion asynchronously, so when the backup is deleted, it doesn't mean the snapshot has been deleted, but it will eventually be deleted. We need to describe this clearly in doc.

innobead (changes requested):
The PR is good for space efficiency, but need to consider https://github.com/longhorn/longhorn-manager/pull/2783#issuecomment-2143494694 first.

derekbit:
ping @FrankYang0529 Ready for review?

FrankYang0529:
@derekbit, sorry for late. I update the code to add `auto-cleanup-when-delete-backup` setting and it's ready for review now.

controller/backup_controller.go
```diff
@@ -317,6 +317,31 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
...
 
+		autocleanup, err := bc.ds.GetSettingAsBool(types.SettingNameAutoCleanupSnapshotWhenDeleteBackup)
+		if err != nil {
+			logrus.WithError(err).WithFields(logrus.Fields{
+				"name": types.SettingNameAutoCleanupSnapshotWhenDeleteBackup,
+			}).Warn("Failed to get the setting")
+		}
+		if autocleanup {
+			// do the best effort to delete the snapshot
+			snapshot, err := bc.ds.GetSnapshot(backup.Spec.SnapshotName)
```
innobead, line 328: Use `GetSnapshotRO` instead.

innobead (changes requested):
LGTM, only one comment.

innobead (approved):

PR #2784, opened 2024-05-08
fix(backup): appropriate warning messages for credential data

mantissahz:

Issue # longhorn/longhorn#7159

### DIFF
--- a/datastore/longhorn.go
+++ b/datastore/longhorn.go
@@ -313,7 +313,17 @@ func (s *DataStore) ValidateSetting(name, value string) (err error) {
 		for _, checkKey := range checkKeyList {
 			if value, ok := secret.Data[checkKey]; ok {
 				if strings.TrimSpace(string(value)) != string(value) {
-					return fmt.Errorf("there is space or new line in %s", checkKey)
+					switch {
+					case strings.TrimLeft(string(value), " ") != string(value):
+						return fmt.Errorf("invalid leading white space in %s", checkKey)
+					case strings.TrimRight(string(value), " ") != string(value):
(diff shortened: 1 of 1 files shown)


derekbit (approved):
LGTM

derekbit:
@c3y1huang Can you help review? Thank you.

datastore/longhorn.go
```diff
@@ -313,6 +313,16 @@ func (s *DataStore) ValidateSetting(name, value string) (err error) {
 		for _, checkKey := range checkKeyList {
 			if value, ok := secret.Data[checkKey]; ok {
 				if strings.TrimSpace(string(value)) != string(value) {
+					switch {
+					case strings.TrimLeft(string(value), " ") != string(value):
+						return fmt.Errorf("there is leading white space in %s", checkKey)
```
c3y1huang, line 318:
```suggestion
						return fmt.Errorf("invalid leading white space in %s", checkKey)
```
Could you also fix the others? Thank you.

c3y1huang (changes requested):

datastore/longhorn.go
mantissahz: Sure. Modified. Thanks

c3y1huang (approved):
The value could contain mixed invalid conditions, so it's better to check all conditions and aggregate the errors with [MultiError](https://github.com/longhorn/longhorn-manager/blob/master/util/multierror.go). However, I am also fine with the current implementation, we can improve this later.

mantissahz:
@Mergifyio backport v1.6.x

mantissahz:
@Mergifyio backport v1.5.x

PR #2792, opened 2024-05-10
Always clean up FailedToScheduleReplica with wrong HardNodeAffinity

ejweber:

longhorn/longhorn#8522

Delete a failed to schedule replica with HardNodeAffinity if DataLocality is disabled.

Previously, we would only do this if there were enough healthy replicas, but it led to the "deadlock" in longhorn/longhorn#8522 where we would not schedule more replicas until we deleted the failed to schedule one, but we would not delete the failed to schedule one until there were enough healthy ones.

I experimented with changing the behavior of this function even more, but there were some weird side effects. I think it is better to keep it working almost exactly as it did before. I did, however, rearrange a bit.

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -975,7 +975,7 @@ func (c *VolumeController) cleanupReplicas(v *longhorn.Volume, es map[string]*lo
 		return nil
 	}
 
-	if err := c.cleanupFailedToScheduledReplicas(v, rs); err != nil {
+	if err := c.cleanupFailedToScheduleReplicas(v, rs); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


ejweber:
## Test steps

Follow the reproduce steps in https://github.com/longhorn/longhorn/issues/8522#issuecomment-2105011487.

6. Observe that the failed to schedule replica is deleted and a new one (that can be scheduled) replaces it.

ejweber:
Regression testing in: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/6924/.

controller/volume_controller.go
```diff
@@ -1028,24 +1028,38 @@ func (c *VolumeController) cleanupCorruptedOrStaleReplicas(v *longhorn.Volume, r
 	return nil
 }
 
-func (c *VolumeController) cleanupFailedToScheduledReplicas(v *longhorn.Volume, rs map[string]*longhorn.Replica) (err error) {
+func (c *VolumeController) cleanupFailedToScheduleReplicas(v *longhorn.Volume, rs map[string]*longhorn.Replica) (err error) {
```
james-munson, line 1031: While you're at it, the name could change to cleanupFailedToScheduleReplica, since it used to do multiple replicas, and now it does the first such it finds.

james-munson: (Personally, I found the name hard to parse, until I mentally translated it as cleanupUnscheduledReplica.)

james-munson (commented):
One question.

controller/volume_controller.go
ejweber: TBH, this is a bit of an oversight on my part. When I read the original code, I thought it deleted one replica at a time as well. Reading it back again, I'm not sure why. I'm inclined to return to the original behavior and allow multiple deletions in a single reconcile. WDYT?

james-munson: I agree, I would stick with the original behavior unless there is a pressing reason to change, and I don't see one.

controller/volume_controller.go
```diff
@@ -1028,24 +1028,38 @@ func (c *VolumeController) cleanupCorruptedOrStaleReplicas(v *longhorn.Volume, r
...
+				replicaToCleanUp = r
+				break
 			}
 		}
 	}
 
+	if replicaToCleanUp != nil {
+		logrus.Infof("Cleaning up failed to scheduled replica %v", replicaToCleanUp.Name)
+		if err := c.deleteReplica(replicaToCleanUp, rs); err != nil {
+			return errors.Wrapf(err, "failed to cleanup failed to scheduled replica %v", replicaToCleanUp.Name)
```
james-munson, line 1059: Really tiny nit, but could you make this error text "failed to cleanup failed-to-schedule replica %v".

controller/volume_controller.go
james-munson: Looks good.

james-munson (approved):
LGTM.

shuo-wu (approved):

ejweber:
@mergify backport v1.5.x

ejweber:
@mergify backport v1.6.x

PR #2812, opened 2024-05-21
Fix bug: we should start conversion webhook before initiating Longhorn datastore

PhanLe1010:
longhorn/longhorn#8578
More details are at https://github.com/longhorn/longhorn/issues/8578#issuecomment-2121405795

### DIFF
--- a/app/daemon.go
+++ b/app/daemon.go
@@ -148,22 +148,30 @@ func startManager(c *cli.Context) error {
 
 	logger := logrus.StandardLogger().WithField("node", currentNodeID)
 
-	clients, err := client.NewClients(kubeconfigPath, ctx.Done())
-	if err != nil {
-		return err
-	}
-
 	// Conversion webhook needs to be started first since we use its port 9501 as readiness port.
 	// longhorn-manager pod becomes ready only when conversion webhook is running.
 	// The services in the longhorn-manager can then start to receive the requests.
-	webhookTypes := []string{types.WebhookTypeConversion, types.WebhookTypeAdmission}
-	for _, webhookType := range webhookTypes {
-		if err := webhook.StartWebhook(ctx, webhookType, clients); err != nil {
-			return err
-		}
-		if err := webhook.CheckWebhookServiceAvailability(webhookType); err != nil {
-			return err
-		}
+	// Conversion webhook does not longhorn datastore.
+	clientsWithoutDatastore, err := client.NewClients(kubeconfigPath, false, ctx.Done())
+	if err != nil {
+		return err
+	}
+	if err := webhook.StartWebhook(ctx, types.WebhookTypeConversion, clientsWithoutDatastore); err != nil {
+		return err
+	}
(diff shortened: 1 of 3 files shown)


PhanLe1010:
Putting this PR in draft until doing some more testing and have better code style

derekbit (commented):
LGTM

PhanLe1010:
@derekbit Can you review the PR again? It actually still needs a k8s client. Just do not need the Longhorn datastore. 

The new modification will create 2 clients:
1. One without Longhorn datastore
2. One with Longhorn datastore.

I think it should not create a high memory usage issue since the one without Longhorn datastore doesn't have any cache. However, we should still double-check the test  related to multiple clients in Longhorn manager. I forgot the ticket number, do you remember it?

derekbit:
> @derekbit Can you review the PR again? It actually still needs a k8s client. Just do not need the Longhorn datastore.
> 
> The new modification will create 2 clients:
> 
> 1. One without Longhorn datastore
> 2. One with Longhorn datastore.
> 
> I think it should not create a high memory usage issue since the one without Longhorn datastore doesn't have any cache. However, we should still double-check the test related to multiple clients in Longhorn manager. I forgot the ticket number, do you remember it?

longhorn#6866?

ChanYiLin:
LGTM
@PhanLe1010 do you mean this one? https://github.com/longhorn/longhorn/issues/6936

PhanLe1010:
Yes, it is the test plan at https://github.com/longhorn/longhorn/issues/6936#issuecomment-1775396141

Thank you @ChanYiLin and @derekbit !

derekbit (commented):
The separate clients LGTM.
I think there is no memory usage issue as @PhanLe1010 said.

PhanLe1010:
Testing looks good. It is ready to be merged after the reviewers approve

ChanYiLin (dismissed):

PhanLe1010:
Looks like there is an new race condition arise after this PR. Please hold on merging

innobead:
@PhanLe1010 is this still draft?

PhanLe1010:
It turned out that we not only have problem with upgrade stuck but also the problem with uninstall stuck because we delete the longhorn-manager daemonset (AKA the conversion webhook server) then list and the the BackupTarget which is still in v1beta1 version. This flow will fail. To solve this problem:
1. A quick hack is https://github.com/longhorn/longhorn-manager/pull/2812/commits/ad321772848e34cc3074af69157bbe1f3eb15260
2. A better solution would be to migrate the API versions for all Longhorn resources in the upgrade path (thus eliminate the need for conversion webhook for future Longhorn upgrade) OR separating the conversion webhook from longhorn-manager 

Due to time pressure, I think we can go with the first approach for now and do the 2nd approach in v1.7.0 . WYDT cc @derekbit @innobead @ChanYiLin @ejweber

innobead:
Sounds good to me.

innobead (approved):

innobead:

PR #2817, opened 2024-05-23
Update engine and volume size less frequently

ejweber:

longhorn/longhorn#8076

Update the size of the volume-head snapshot in the engine CR less frequently. As a result, also update the actualSize in the volume CR less frequently.

I basically stole the method for "truncating" from the below.

https://github.com/longhorn/longhorn-manager/blob/4e8fb0ab239559f5bfa6a711ca9b82ceba6b33f1/controller/node_controller.go#L850-L852

I didn't want to use a fixed size of `100 MiB` for the truncation because I want the user to still be able to see the size of small volumes increasing under light write activity. See the code comments and tests for the heuristic I chose. I'm certainly open to discussion on whether it is the right one.

### DIFF
--- a/controller/engine_controller.go
+++ b/controller/engine_controller.go
@@ -936,7 +936,7 @@ func (m *EngineMonitor) refresh(engine *longhorn.Engine) error {
 	if err != nil {
 		engine.Status.SnapshotsError = err.Error()
 	} else {
-		engine.Status.Snapshots = snapshots
+		engine.Status.Snapshots = m.prepareSnapshotsForStatus(snapshots, engine.Spec.VolumeSize)
 		engine.Status.SnapshotsError = ""
 	}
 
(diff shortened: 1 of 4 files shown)


ejweber:
To test:

- Deploy Longhorn without this change (e.g. `master-head` before it merged).
- Deploy the following kbench job. (Note: it uses a custom image.) The job writes to a `60 GiB` volume at `1 MiB/s`.

```
kind: PersistentVolumeClaim
apiVersion: v1
metadata:
  name: kbench-pvc
spec:
  volumeMode: Filesystem
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 60Gi
---
apiVersion: batch/v1
kind: Job
metadata:
  name: kbench
spec:
  template:
    metadata:
      labels:
        kbench: fio
    spec:
      containers:
      - name: kbench
        image: webericjames/kbench:run-indefinitely
        imagePullPolicy: Always
        env:
        - name: MODE
          value: "sequential-write-bandwidth"
        - name: FILE_NAME
          value: "/volume/test"
        - name: SIZE
          value: "55G"
        - name: CPU_IDLE_PROF
          value: "disabled"
        - name: RATE
          value: 1m,1m
        volumeMounts:
        - name: vol
          mountPath: /volume/
      restartPolicy: Never
      volumes:
      - name: vol
        persistentVolumeClaim:
          claimName: kbench-pvc
  backoffLimit: 0
```

- Use the following Prometheus queries to verify that there are twelve updates per minute to both engine and volume CRs even though there is not much writing.

```
sum (rate(apiserver_request_total{resource="engines", verb="PUT", group="longhorn.io", job="apiserver"}[2m])) * 60
sum (rate(apiserver_request_total{resource="volumes", verb="PUT", group="longhorn.io", job="apiserver"}[2m])) * 60
```

- Use the following kubectl command to see the `size` of the `volume-head` snapshot in the engine CR changing by `5 MiB/s` every five seconds.

```
kl get engine -oyaml -w | grep size
```

- While the kbench job is running, swap in a longhorn-manager image with this change.
- Use the same Prometheus queries to verify that there is only one update per minute to both engine and volume CRs.
- Use the same kubectl command to see the `size` of the `volume-head` snapshot in the engine CR changing by `60 MiB` every sixty seconds.

```
eweber@laptop:~/longhorn> kl get engine -oyaml -w | grep size
      size: "3208642560"
      size: "3271557120"
      size: "3334471680"
      size: "3397386240"
      size: "3460300800"
```

- Repeat with a smaller volume. A `30 GiB` volume should change by `30MiB` every thirty seconds based on the heuristic I chose.
- Repeat with a larger volume. Any volume `100 GiB` or above should change by `100 MiB` every 100 seconds. We do not every decrease the granularity of our updating lower than `100 MiB`.

Here are graphs from the `60 GiB` volume case.

image
image

ejweber:
Running regression tests to check on @PhanLe1010's suggestion in https://github.com/longhorn/longhorn/issues/8076#issuecomment-2098944658.

https://ci.longhorn.io/job/private/job/longhorn-tests-regression/6971/

james-munson (approved):
That seems like a thoughtful and reasonable heuristic to me. I might have opted for a 1% rule, but I can seen how that might be too big a jump on a sizeable volume. The results look like what was hoped for.

ejweber:
Converted to a draft. After discussing with @james-munson, @shuo-wu, and @PhanLe1010, the decision is that sizes in engine and volume CRs are, in principle, different than the size in the node CR. It is not acceptable for them to not have eventual byte accuracy. (In this PR, we truncate engine and volume sizes in a similar way to disk sizes, so that the size remains an approximation indefinitely.)

The main approach discussed is to use rate limiting on the size field instead. For example, allow frequent size changes at first, but if only the size is changing and not other fields, back off over time. There are two suggested approaches:

1. Keep the time we last updated in the engine CR and use that as part of the exponential backoff calculation.
2. Maintain an exponential backoff in memory within the engine CR. This is the approach I will experiment with.

ejweber:
Closed in favor of https://github.com/longhorn/longhorn-manager/pull/2837.

PR #2819, opened 2024-05-24
fix: fix system restore PV and PVC stuck beause of backing image and volume

ChanYiLin:
https://github.com/longhorn/longhorn/issues/8601

During the system restore
Volume depends on BackingImage, our webhook rejects the Volume creation if the BackingImage is not in the cluster
=> move the volume restoration out of the parallel restoration

And since PV depends on Volume
=> So we also need to move PV,PVC restoration after Volume restoration

cc @c3y1huang

### DIFF
--- a/controller/system_rollout_controller.go
+++ b/controller/system_rollout_controller.go
@@ -435,19 +435,17 @@ func (c *SystemRolloutController) systemRollout() error {
 
 		wg := &sync.WaitGroup{}
 		restoreFns := map[string]func() error{
-			types.KubernetesKindServiceList:               c.restoreService,
-			types.KubernetesKindServiceAccountList:        c.restoreServiceAccounts,
-			types.KubernetesKindClusterRoleList:           c.restoreClusterRoles,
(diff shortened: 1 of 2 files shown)


ChanYiLin:
Hi @innobead 
Can we disable the check for non touched files.
That will make the PR complicated.

c3y1huang (approved):

innobead (approved):

innobead:
This is interesting, as the resource creation will not be independent in this case.

innobead:
Yeah, I see, especially related to runtime validation. 

[/webhook/resources/volume/validator.go#L97-L101](https://github.com/longhorn/longhorn-manager/blob/23ccdd17f9ef3000038f66ef5b0380134fd7c335/webhook/resources/volume/validator.go#L97-L101)
```go
	if volume.Spec.BackingImage != "" {
		if _, err := v.ds.GetBackingImage(volume.Spec.BackingImage); err != nil {
		}
	}
```

PR #2823, opened 2024-05-27
fix: need to wait until snapshot is ready before taking backup in system backup

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/8601

In system backup e2e tests
Some tests failed because the backup creation failed due to snapshot missing.
We should wait until snapshot is ready before create backup, cause backup creation is one time shot.

ref: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/6979/console

in the support bundle, the backup creation failed because of snapshot missing
```
  spec:
    labels:
      longhorn.io/volume-access-mode: rwo
    snapshotName: system-backup-49d05cf04e624d85
    syncRequestedAt: null
  status:
    backupCreatedAt: "null"
    compressionMethod: "null"
    error: could not find snapshot 'system-backup-49d05cf04e624d85' to backup, engine
      'longhorn-testvol-jway75-e-0'
    labels: "null"
```

### DIFF
--- a/controller/system_backup_controller.go
+++ b/controller/system_backup_controller.go
@@ -79,6 +79,10 @@ const (
 	systemBackupRecordTypeNormal = systemBackupRecordType("normal")
 )
 
+const (
+	SnapshotReadyTimeout = 300
+)
+
 type systemBackupRecord struct {
 	nextState longhorn.SystemBackupState
 
@@ -842,6 +846,10 @@ func (c *SystemBackupController) createVolumeBackup(volume *longhorn.Volume, sys
 		return nil, errors.Wrapf(err, "failed to create Volume %v snapshot %s", volume.Name, snapshot.Name)
 	}
 
+	if err = c.waitForSnaphotReady(volumeBackupName, SnapshotReadyTimeout); err != nil {
+		return nil, errors.Wrapf(err, "failed to wait for snapshot %v to be ready", volumeBackupName)
+	}
+
 	backup = &longhorn.Backup{
 		ObjectMeta: metav1.ObjectMeta{
 			Name: volumeBackupName,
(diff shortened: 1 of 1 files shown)


c3y1huang (commented):
Having a blocker in the system backup controller to wait for the snapshot to be ready doesn't look like a good idea. Could we handle this within the backup controller before invoking `engineClientProxy.SnapshotBackup`?

ChanYiLin:
Hi @c3y1huang 
```
```

ChanYiLin:
@c3y1huang 
May I know why **"Having a blocker in the system backup controller"** is not a good idea?
Cause we also wait for Volume to be backed up with a timer as well
We even return the error if the backup timeout

c3y1huang:
> Oh I see your point we check if snapshot exists before invoking `engineClientProxy.SnapshotBackup` if snapshot doesn't exist, we then wait for the next round.
> 
> I think we can do it



> Oh I see your point we check if snapshot exists before invoking `engineClientProxy.SnapshotBackup` if snapshot doesn't exist, we then wait for the next round.
> 
> I think we can do it

Exactly, I think it would make more sense for the backup controller to handle the missing snapshot rather than leaving the responsibility to the system backup controller.

ChanYiLin:
@c3y1huang I am just not sure, if we have any test or code logic that assumes the backup should be failed in a period of time if snapshot is missing.

Because in this case, the backup CR will be hanging there without any status forever if snapshot is missing

c3y1huang:
> @c3y1huang I am just not sure, if we have any test or code logic that assumes the backup should be failed in a period of time if snapshot is missing.
> 
> Because in this case, the backup CR will be hanging there without any status forever if snapshot is missing

Can we introduce a new state for the backup? Perhaps `Pending`?

ChanYiLin:
It already have all the states from `New`->`Pending`->...->`Completed`
But those are syned from the backup state from the engine.

Maybe we can just assign `Pending` to the CR before it starts,
it will then be overridden by state from engine once it starts.

But it will be `Pending` forever, if the snapshot is never created.

c3y1huang:
> It already have all the states from `New`->`Pending`->...->`Completed` But those are syned from the backup state from the engine.
> 
> Maybe we can just assign `Pending` to the CR before it starts, it will then be overridden by state from engine once it starts.
> 
> But it will be `Pending` forever, if the snapshot is never created.



> It already have all the states from `New`->`Pending`->...->`Completed` But those are syned from the backup state from the engine.
> 
> Maybe we can just assign `Pending` to the CR before it starts, it will then be overridden by state from engine once it starts.
> 
> But it will be `Pending` forever, if the snapshot is never created.

I see. Any indication on the Snapshot CR if the actual snapshot never gets created?

PR #2825, opened 2024-05-28
feat(support-bundle): add support-bundle-node-collection-timeout setting

c3y1huang:

Issue longhorn/longhorn#8623

Propose adding a setting that allows users to configure the node bundle collection timeout during support bundle generation.

`None`

https://github.com/rancher/support-bundle-kit/pull/109

### DIFF
--- a/types/setting.go
+++ b/types/setting.go
@@ -103,6 +103,7 @@ const (
 	SettingNameRecurringFailedJobsHistoryLimit                          = SettingName("recurring-failed-jobs-history-limit")
 	SettingNameRecurringJobMaxRetention                                 = SettingName("recurring-job-max-retention")
 	SettingNameSupportBundleFailedHistoryLimit                          = SettingName("support-bundle-failed-history-limit")
+	SettingNameSupportBundleNodeCollectionTimeout                       = SettingName("support-bundle-node-collection-timeout")
 	SettingNameDeletingConfirmationFlag                                 = SettingName("deleting-confirmation-flag")
 	SettingNameEngineReplicaTimeout                                     = SettingName("engine-replica-timeout")
 	SettingNameSnapshotDataIntegrity                                    = SettingName("snapshot-data-integrity")
@@ -187,6 +188,7 @@ var (
 		SettingNameRecurringFailedJobsHistoryLimit,
 		SettingNameRecurringJobMaxRetention,
 		SettingNameSupportBundleFailedHistoryLimit,
+		SettingNameSupportBundleNodeCollectionTimeout,
 		SettingNameDeletingConfirmationFlag,
 		SettingNameEngineReplicaTimeout,
 		SettingNameSnapshotDataIntegrity,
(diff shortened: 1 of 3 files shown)


c3y1huang:
cc @Yu-Jack

types/setting.go
```diff
@@ -1057,6 +1060,20 @@ var (
 		},
 	}
 
+	SettingDefinitionSupportBundleNodeCollectionTimeout = SettingDefinition{
+		DisplayName: "Timeout for Support Bundle Node Collection",
+		Description: "In minutes. The timeout for collecting node bundles for support bundle generation. The default value is 1440 (24 hours).\n\n" +
```
innobead, line 1065: The time out is quite long. Is it possible to have any unexpected hanging?

Are users able to see what's happening until timeout? For example, check the support manager  log.

c3y1huang: The node collection involves gathering and compressing host logs and host info, which are then transferred to the support-bundle manager. However, some clusters might have larger log files, which I assume might prolong the process. Users can also check the support bundle manager logs for errors.

I am not sure what would be a safe value for the timeout, given that the support bundle generation will proceed even if the node collection times out. So the 24-hour timeout is just my concern that if the support bundle might be generated without complete node collection the issue is caused by a slow environment.

I've raised the question in the original issue [here](https://github.com/longhorn/longhorn/issues/8623#issuecomment-2126278720), but it might be overlooked.

@PhanLe1010 @ejweber @james-munson @shuo-wu @derekbit , do we have an idea of what a reasonable timeout for the node bundle collection would be?

innobead: What's the original default value? We can just keep it the same if we don't know, as it is being used.

c3y1huang: In the previous version of the support bundle kit, there was no timeout specified for the node collection process.
It's recently been added in https://github.com/rancher/support-bundle-kit/pull/109, given a default of 30 minutes. And this change has not yet been tested in real-world scenarios.

c3y1huang: I think it's also fine to keep the default at 30 minutes. If the initial support bundle doesn't include the node collection, we can ask the user to regenerate it with a longer timeout. Over time, we might be able to work out an appropriate volume for the timeout. What do you think?

Yu-Jack: In [original ideal](https://github.com/rancher/support-bundle-kit/issues/73#issuecomment-2024768040), I think it's better rewrite the current download flow. For example, show each node progress on GUI, then users could decide to terminate or not. 

Otherwise, it's too hard to decide the appropriate value, and specifically we don't have any metrics to record how long node spends. However, rewriting flow is too much effort.  

So, like @c3y1huang mentioned, I think it's good approach to figure out the appropriate value for the timeout in this case.

innobead: For the default value of the timeout, it should be fine to use the current default one, as users can rerun it with a longer value if needed.

c3y1huang: Updated. Thanks!

innobead:
Please pay attention to users feedback in the future, so we can continue improving UX.

c3y1huang:

PR #2826, opened 2024-05-28
feat(image): pre-pull instance/share manager images

mantissahz:

Issue # longhorn/longhorn#8376

- add a mutator to handle the `Image` resources.
- add the image controller to handle the `Image` resources.
- modify the uninstall controller for cleaning up `Image`.


mantissahz:
we do not need to change codes now.

PR #2831, opened 2024-05-28
unit test: fix `newInstanceManager`

derekbit:

Issue https://github.com/longhorn/longhorn/issues/8658

The TestCleanupRedundantInstanceManagers test fails due to a hard-coded and incorrect instance manager image. This issue surfaces after the introduction of the feature in commit 7b3521b6b9086d0a0b14b8da2621cc7b626f02d6.


controller/controller_test.go
```diff
@@ -255,14 +255,15 @@ func newInstanceManager(
...
+	instanceManagerImage string,
 	isDeleting bool) *longhorn.InstanceManager {
 
 	im := &longhorn.InstanceManager{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      name,
 			Namespace: TestNamespace,
 			UID:       uuid.NewUUID(),
-			Labels:    types.GetInstanceManagerLabels(nodeID, TestInstanceManagerImage, longhorn.InstanceManagerTypeAllInOne, dataEngine),
+			Labels:    types.GetInstanceManagerLabels(nodeID, instanceManagerImage, longhorn.InstanceManagerTypeAllInOne, dataEngine),
```
innobead, line 266: I don't think the result will be different after passing a name instead of using the hard-coded value. It seems you are still passing `TestInstanceManagerImage` for each function call. Or I missed something?

derekbit: Check this one

image

innobead (approved):

mantissahz (approved):
LGTM
Just one question: we still have two instance managers when doing the `s.controller.diskMonitor.RunOnce()`, right? And I assume in the real case we won't have two instance managers with the same instance manager image.

derekbit:
> LGTM Just one question: we still have two instance managers when doing the `s.controller.diskMonitor.RunOnce()`, right? And I assume in the real case we won't have two instance managers with the same instance manager image.

The two IMs have different images.

derekbit:
@Mergify backport v1.6.x

PR #2837, opened 2024-05-31
Update engine and volume size less frequently

ejweber:

longhorn/longhorn#8076

This is meant to be an improvement over the rejected https://github.com/longhorn/longhorn-manager/pull/2817. It adds a rate limiter to provide the following behavior:

- The rate limiter generally only allows "size-only updates" (updates to `engine.status.snapshots[volume-head].size` once every thirty seconds.
- For responsiveness to bursty writing, the rate limiter allows up to three "size-only updates" in quick succession. (The underlying implementation is a "token bucket", so this burstability can be rebuilt over time if no writes happen.)
- Size changes that exceed a heuristic threshold ignore rate limiting (otherwise, it might feel weird).
- Any status changes that aren't "size-only" ignore rate limiting.
- Size fields have "eventual" byte accuracy (they are not truncated).

If this approach is generally acceptable, we can tweak `sizeUpdateLimit`, `sizeUpdateBurst`, or the heuristic now or in the future to provide different tradeoffs between ETCD load and responsiveness.

### DIFF
--- a/controller/engine_controller.go
+++ b/controller/engine_controller.go
@@ -13,6 +13,8 @@ import (
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 
+	"golang.org/x/time/rate"
+
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/tools/cache"
@@ -61,6 +63,11 @@ var (
 
 	// restoreMaxInterval: deleting the backup of big size volume takes a long time for retain policy and restoring backups would be in backoff period.
 	restoreMaxInterval = 1 * time.Hour
+
+	// amount of time between actual size updates that do not exceed threshold during periods with stable writes
+	sizeUpdateLimit = 30 * time.Second
+	// number of consecutive actual size updates allowed during bursts
+	sizeUpdateBurst = 3
 )
 
 const (
@@ -120,6 +127,8 @@ type EngineMonitor struct {
 	restoringCounter         util.Counter
 	restoringCounterAcquired bool
 	restoringCounterMutex    *sync.Mutex
+
+	sizeUpdateLimiter *rate.Limiter
 }
 
 func NewEngineController(
@@ -744,6 +753,7 @@ func (ec *EngineController) startMonitoring(e *longhorn.Engine) {
 		proxyConnCounter:       ec.proxyConnCounter,
 		restoringCounter:       ec.restoringCounter,
 		restoringCounterMutex:  ec.restoringCounterMutex,
+		sizeUpdateLimiter:      rate.NewLimiter(rate.Every(sizeUpdateLimit), sizeUpdateBurst),
 	}
 
 	ec.engineMonitorMutex.Lock()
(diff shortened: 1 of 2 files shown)


ejweber:
### To test:

#### Rate limited

- Deploy Longhorn without this change (e.g. `master-head` before it merged).
- Deploy the following kbench job. (Note: it uses a custom image.) The job writes to a `60 GiB` volume at `1 MiB/s`.

```
kind: PersistentVolumeClaim
apiVersion: v1
metadata:
  name: kbench-pvc
spec:
  volumeMode: Filesystem
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 60Gi
---
apiVersion: batch/v1
kind: Job
metadata:
  name: kbench
spec:
  template:
    metadata:
      labels:
        kbench: fio
    spec:
      containers:
      - name: kbench
        image: webericjames/kbench:run-indefinitely
        imagePullPolicy: Always
        env:
        - name: MODE
          value: "sequential-write-bandwidth"
        - name: FILE_NAME
          value: "/volume/test"
        - name: SIZE
          value: "55G"
        - name: CPU_IDLE_PROF
          value: "disabled"
        - name: RATE
          value: 1m,1m
        volumeMounts:
        - name: vol
          mountPath: /volume/
      restartPolicy: Never
      volumes:
      - name: vol
        persistentVolumeClaim:
          claimName: kbench-pvc
  backoffLimit: 0
```

- Use the following Prometheus queries to verify that there are twelve updates per minute to both engine and volume CRs even though there is not much writing.

```
sum (rate(apiserver_request_total{resource="engines", verb="PUT", group="longhorn.io", job="apiserver"}[2m])) * 60
sum (rate(apiserver_request_total{resource="volumes", verb="PUT", group="longhorn.io", job="apiserver"}[2m])) * 60
```

- Use the following kubectl command to see the `size` of the `volume-head` snapshot in the engine CR changing by `5 MiB/s` every five seconds.

```
kl get engine -oyaml -w | grep size
```

- While the kbench job is running, swap in a longhorn-manager image with this change.
- Use the same Prometheus queries to verify that there are two updates per minute to both engine and volume CRs (after things have calmed a bit).
- Use the same kubectl command to see the `size` of the `volume-head` snapshot in the engine CR changing by `~30 MiB` every thirty seconds.
- The updates are once every thirty seconds because that is `sizeUpdateLimit` and we are not writing fast enough to ignore rate limiting.

```
eweber@laptop:~/longhorn> kl get engine -oyaml -w | grep size
      size: "5989253120"
      size: "6020710400"
      size: "6046924800"
      size: "6083624960"
```

image
image

---

#### Too much writing for rate limiting

- Repeat with a kbench pod that writes faster. Because it writes at `60 MiB/s`, it is never rate limited. It causes `12 PUT/s` (like in the unmodified case). 

```
        - name: RATE
          value: 12m,12m
```

---

#### Burst

- Deploy `examples/pod_with_pvc.yaml`.
- Execute the following multiline command.

```
kl get engine -oyaml | grep size && \
k exec volume-test -- dd if=/dev/random bs=4k count=1 of=/data/file1 && \
sleep 6 && \
kl get engine -oyaml | grep size && \
k exec volume-test -- dd if=/dev/random bs=4k count=1 of=/data/file2 && \
sleep 6 && \
kl get engine -oyaml | grep size && \
k exec volume-test -- dd if=/dev/random bs=4k count=1 of=/data/file3 && \
sleep 6 && \
kl get engine -oyaml | grep size && \
k exec volume-test -- dd if=/dev/random bs=4k count=1 of=/data/file4 && \
sleep 6 && \
kl get engine -oyaml | grep size
```

```

```

- Even though the writes are super small (`4 kiB`) the first three writes cause a size update anyway. The fourth write does not because we have maxed out our potential to burst (governed by `sizeUpdateBurst`).
image

controller/engine_controller.go
```diff
@@ -2215,3 +2238,75 @@ func shouldProceedToWaitAndRebuild(e *longhorn.Engine, replicaName, originalRepl
...
+	newSnapshot, newSnapshotOK := new.Snapshots[etypes.VolumeHeadName]
+	if !existingSnapshotOK || !newSnapshotOK {
+		return !reflect.DeepEqual(existing, new)
+	}
+
+	// Otherwise, compare without the size of volume-head.
+	existingSize := existingSnapshot.Size
+	existingSnapshot.Size = newSnapshot.Size
+	needUpdate := !reflect.DeepEqual(existing, new)
+	existingSnapshot.Size = existingSize
```
james-munson, line 2286: A safer and much more laborious way would be to make copies of new and existing snapshot and zero out the respective Size fields for the DeepEqual check.  That has the advantage of not doctoring the input arguments, which one would expect to be `const` in some sense.  We just hope that nobody makes a copy or refers to the Size field for the brief moment that it is being hijacked.  You probably considered that and decided that the risk was tiny and the overhead of the full copy just for comparison was pretty heavyweight for such a mild exposure.

james-munson (commented):
LGTM.  Thanks for the explanatory comments on Tokens() before Allow().  Just one small question.

controller/engine_controller.go
ejweber: I agree with your analysis here, and you are correct about my thought process. Since `existingEngine` itself is a deep copy created just for this refresh and has no practical way of being operated simultaneously by another actor, the exposure is small. I didn't want to deep copy the entire `EngineStatus` struct (I think there's no way around comparing the whole thing) additionally every five seconds to mitigate the risk.

james-munson (dismissed):

PhanLe1010 (approved):
LGTM

Thank you for the great implementation using token bucket rate limit :rocket:

innobead (approved):
LGTM. Very well considered design for different conditions.

ejweber:

PR #2843, opened 2024-06-04
Cancel migration when the old node is down

ejweber:

longhorn/longhorn#3401

According to my analysis at https://github.com/longhorn/longhorn/issues/3401#issuecomment-2148015598, we cannot roll back a migration from `oldNode` to `newNode` while `oldNode` is down primarily because NO `processMigration` code can run while a volume's robustness is unknown. In other words, typically, we must wait for a volume to be attached and running before taking migration steps.

In this issue, however, `oldNode` is down. It may take a long time to come back, or it may never come back. We want to allow the migration from `oldNode` to be canceled so that the migration engine and replicas can be cleaned up. (Long running migrations can cause unexpected issues, so it's better to end them if we can.)

In this PR, we delay checking the volume's robustness until AFTER we check for the special case that `oldNode` is down. If `oldNode` is down, we cancel the migration and clean up the extra resources.

### DIFF
--- a/controller/utils.go
+++ b/controller/utils.go
@@ -18,7 +18,7 @@ func hasReplicaEvictionRequested(rs map[string]*longhorn.Replica) bool {
 	return false
 }
 
-func (vc *VolumeController) isVolumeUpgrading(v *longhorn.Volume) bool {
+func isVolumeUpgrading(v *longhorn.Volume) bool {
 	return v.Status.CurrentImage != v.Spec.Image
 }
 
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -971,7 +971,7 @@ func (c *VolumeController) cleanupReplicas(v *longhorn.Volume, es map[string]*lo
 	}
 
 	// give a chance to delete new replicas failed when upgrading volume and waiting for IM-r starting
-	if c.isVolumeUpgrading(v) {
+	if isVolumeUpgrading(v) {
 		return nil
 	}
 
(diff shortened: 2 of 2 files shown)


ejweber:
Expected behavior:

After this change, when the reproduce steps from https://github.com/longhorn/longhorn/issues/3401#issue-1079076046 are followed:

- The volume still remains attached to `oldNode` with `robustness == unknown`. This is NOT the target of the PR (or the current target of the issue) to resolve.
- The migration engine is cleaned up.
- The migration replicas are cleaned up.
- There is no evidence that the volume is attempting to migrate (as it no longer is). `status.currentMigrationNodeID == ""` and `spec.migrationNodeID == ""`.

ejweber:
Running regression tests at https://ci.longhorn.io/job/private/job/longhorn-tests-regression/7003/.

shuo-wu:
Generally speaking, there are 4 cases regarding this kind of node-down-during-migration issue:
1. The current/old node is down while users want to detach the volume from the old node. (Migration confirmation)
2. The migration/new node is down while users want to detach the volume from the old node. (Migration confirmation)
3. The current/old node is down while users want to detach the volume from the new node. (Migration cancellation)
4. The migration/new node is down while users want to detach the volume from the new node. (Migration cancellation)

What the original issue describes should be the 3rd case. Can Longhorn handle all of them correctly?
For the 1st and 4th cases, I think Longhorn should handle the migration completely.
For the 2nd and 3rd cases, I think Longhorn should be able to clean up the correct engine then leave the engine on the down node there, and the volume remains `unknown`.

ejweber:
Converted to draft while handling https://github.com/longhorn/longhorn/issues/8735.

ejweber:
This one is made unnecessary by https://github.com/longhorn/longhorn-manager/pull/2894 for longhorn/longhorn#8735. Closing.

PR #2844, opened 2024-06-04
Increase the client-go rest client rate limit inside the csi sidecar components

PhanLe1010:

Issue longhorn/longhorn#8699

Unblock the Longhorn control plane when there is a surge of volume creation/attach/detach/snapshotting/resizer. The CSI component will process the events faster.

See more details about reducing steps in the issue description longhorn/longhorn#8699


csi/deployment.go
```diff
@@ -67,6 +67,8 @@ func NewAttacherDeployment(namespace, serviceAccount, attacherImage, rootDir str
 			"--timeout=1m50s",
 			"--leader-election",
 			"--leader-election-namespace=$(POD_NAMESPACE)",
+			"--kube-api-qps=50",
+			"--kube-api-burst=100",
```
shuo-wu, line 71: NIT: Use constants rather than hard-coded values

shuo-wu (dismissed):
LGTM

csi/deployment.go
PhanLe1010: Thank you. Fixed

shuo-wu (approved):

types/types.go
```diff
@@ -204,6 +204,9 @@ const (
 
 	CniNetworkNone          = ""
 	StorageNetworkInterface = "lhnet1"
+
+	KubeApiQps   = 50
+	KubeApiBurst = 100
```
derekbit, line 209: Just curious what will happen if using bigger numbers.

ejweber (approved):
LGTM.

TBH, I'm not sure how we can decide what the "right" values are here. Presumably, the default values set in the sidecar are intended to prevent the volume provisioning/attaching infrastructure from unintentionally denying service to other Kubernetes API clients. It seems fine to increase them to the values we are already using for other Longhorn components (as you have done here), but there is probably some limit to how high they should go. If we decide we want to increase them further in the future, we will need a plan to figure that limit out.

types/types.go
PhanLe1010: I don't see benefit from increasing it to bigger values when testing scaling yet. These values seems big enough to eliminate the throttling

PhanLe1010:
> TBH, I'm not sure how we can decide what the "right" values are here. Presumably, the default values set in the sidecar are intended to prevent the volume provisioning/attaching infrastructure from unintentionally denying service to other Kubernetes API clients. It seems fine to increase them to the values we are already using for other Longhorn components (as you have done here), but there is probably some limit to how high they should go. If we decide we want to increase them further in the future, we will need a plan to figure that limit out.

As we discussed in the US sync. There isn't a concert rule for determine how big is acceptable. I just pick the value base on:
1. Scale test: these values seems big enough to avoid throttling 
2. Other projects are using similar value: [kubectl](https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/#:~:text=%2D%2Devent%2Dqps%20int32%C2%A0%C2%A0%C2%A0%C2%A0%C2%A0Default%3A%2050)

PhanLe1010:

PR #2845, opened 2024-06-04
Fix TooManySnapshots error message

ejweber:

longhorn/longhorn#8668

See the original issue for context.

This is one option for fixing the issue @roger-ryao reported. With this one, the user only hits the `TooManySnapshots` condition after 100 snapshots BESIDES `volume-head` are created. If we do it this way, we probably have to change [test_volume_toomanysnapshots_condition](https://github.com/longhorn/longhorn-tests/blob/4e5d43a272db8b87aacf744d5ef5ec782f9f166e/manager/integration/tests/test_basic.py#L3965-L3976), as creating 100 snapshots new snapshots will no longer trigger the condition.

WDYT @roger-ryao? Is this the appropriate fix?


roger-ryao:
> #### Which issue(s) this PR fixes:
> [longhorn/longhorn#8668](https://github.com/longhorn/longhorn/issues/8668)
> 
> #### What this PR does / why we need it:
> See the original issue for context.
> 
> #### Additional documentation or context
> This is one option for fixing the issue @roger-ryao reported. With this one, the user only hits the `TooManySnapshots` condition after 100 snapshots BESIDES `volume-head` are created. If we do it this way, we probably have to change [test_volume_toomanysnapshots_condition](https://github.com/longhorn/longhorn-tests/blob/4e5d43a272db8b87aacf744d5ef5ec782f9f166e/manager/integration/tests/test_basic.py#L3965-L3976), as creating 100 snapshots new snapshots will no longer trigger the condition.
> 
> WDYT @roger-ryao? Is this the appropriate fix?

Hi @ejweber 
    I built the private `longhorn-manager` from your repo and have tested it. It looks good. 
     We also need to update the test case [test_volume_toomanysnapshots_condition](https://github.com/longhorn/longhorn-tests/blob/4e5d43a272db8b87aacf744d5ef5ec782f9f166e/manager/integration/tests/test_basic.py#L3983) so that the `max_count` is `101`.

derekbit (approved):

derekbit (approved):
LGTM

innobead (approved):

innobead:

PR #2847, opened 2024-06-06
metrics: update warning messages to debug level

derekbit:

Issue Longhorn/longhorn#8266

Metric collector emits warning messages if the engine is not found during the transition state. Update these messages to debug level to prevent flooding in longhorn-manager.


innobead (approved):

innobead:

PR #2865, opened 2024-06-11
Minor improvement for orphan creation

shuo-wu:

https://github.com/longhorn/longhorn/issues/8642

### DIFF
--- a/controller/monitor/disk_monitor.go
+++ b/controller/monitor/disk_monitor.go
@@ -406,7 +406,7 @@ func (m *NodeMonitor) getOrphanedReplicaLvolNames(replicaDataStores map[string]s
 
 	for name := range replicaDataStores {
 		_, err := m.ds.GetReplica(name)
-		if err == nil || !datastore.ErrorIsNotFound(err) {
+		if err == nil || !datastore.ErrorIsNotFound(err) || name == "" {
 			delete(replicaDataStores, name)
 		}
 	}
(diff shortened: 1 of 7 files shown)


controller/monitor/disk_monitor.go
```diff
@@ -388,34 +388,34 @@ func NewDiskInfo(diskName, diskUUID, diskPath string, diskDriver longhorn.DiskDr
...
+	if len(replicaDataStores) == 0 {
 		return map[string]string{}, nil
 	}
 
-	for name := range replicaDirectoryNames {
+	for name := range replicaDataStores {
 		_, err := m.ds.GetReplica(name)
-		if err == nil || !datastore.ErrorIsNotFound(err) {
-			delete(replicaDirectoryNames, name)
+		if err == nil || !datastore.ErrorIsNotFound(err) || name == "" {
```
innobead, line 409: Why will we have this case? `name` is empty?

innobead (commented):
In general, LGTM.

Just one question.

controller/monitor/disk_monitor.go
shuo-wu: As reported in the issue description, somehow a v2 orphan `Spec.Parameters.DataName` is empty, I don't know why an empty string will be in the candidate list but I think at least this function should filter it out here. 
Without this change, getting a replica with an empty `name` will get `ErrorIsNotFound` then the empty string will be an orphaned data store, which shouldn't be.

innobead (approved):

PR #2871, opened 2024-06-11
Make long gRPC timeout configurable

PhanLe1010:
longhorn/longhorn#2765

Note: we need to update go.mod to have this changes: before merging:
1. https://github.com/longhorn/types/pull/14
2. https://github.com/longhorn/longhorn-instance-manager/pull/512

### DIFF
--- a/engineapi/proxy_replica.go
+++ b/engineapi/proxy_replica.go
@@ -4,10 +4,10 @@ import (
 	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
 )
 
-func (p *Proxy) ReplicaAdd(e *longhorn.Engine, replicaName, replicaAddress string, restore, fastSync bool, replicaFileSyncHTTPClientTimeout int64) (err error) {
+func (p *Proxy) ReplicaAdd(e *longhorn.Engine, replicaName, replicaAddress string, restore, fastSync bool, replicaFileSyncHTTPClientTimeout, grpcTimeoutSeconds int64) (err error) {
 	return p.grpcClient.ReplicaAdd(string(e.Spec.DataEngine), e.Name, e.Spec.VolumeName, p.DirectToURL(e),
 		replicaName, replicaAddress, restore, e.Spec.VolumeSize, e.Status.CurrentSize,
-		int(replicaFileSyncHTTPClientTimeout), fastSync)
+		int(replicaFileSyncHTTPClientTimeout), fastSync, grpcTimeoutSeconds)
 }
 
 func (p *Proxy) ReplicaRemove(e *longhorn.Engine, address string) (err error) {
--- a/go.mod
+++ b/go.mod
@@ -63,8 +63,8 @@ require (
 	github.com/longhorn/go-common-libs v0.0.0-20240616051056-103c7d62a0d5
 	github.com/longhorn/go-iscsi-helper v0.0.0-20240610031841-3deb441d256f
 	github.com/longhorn/go-spdk-helper v0.0.0-20240616091844-5afd86a1d892
-	github.com/longhorn/longhorn-engine v1.7.0-dev.0.20240614034019-c1370ebfd1f9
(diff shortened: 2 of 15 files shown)


ejweber (approved):
LGTM.

@jillian-maroket to look at the setting definition in `types/setting.go` if she has time. Usually that definition makes its way into the Helm `values.yaml` file and the website, so it might be easier if she can weigh in sooner rather than later.

jillian-maroket (commented):
PTAL at the reworded the setting description. Thank you.

types/setting.go
```diff
@@ -1199,6 +1202,20 @@ var (
 		},
 	}
 
+	SettingDefinitionLongGPRCTimeOut = SettingDefinition{
+		DisplayName: "Long gRPC Timeout",
+		Description: "In seconds. The setting specifies the timeout for replica rebuilding and snapshot clone operation",
```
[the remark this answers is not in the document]
PhanLe1010: Thank you. Fixed

go.mod
```diff
@@ -48,6 +48,11 @@ replace (
 	k8s.io/sample-controller => k8s.io/sample-controller v0.30.1
 )
 
+replace (
+	github.com/longhorn/longhorn-instance-manager => ../longhorn-instance-manager
+	github.com/longhorn/types => ../types
+)
+
```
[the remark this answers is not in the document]
PhanLe1010: Fixed. Thank you

jillian-maroket (approved):

engineapi/proxy_replica.go
```diff
@@ -4,10 +4,10 @@ import (
 	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
 )
 
-func (p *Proxy) ReplicaAdd(e *longhorn.Engine, replicaName, replicaAddress string, restore, fastSync bool, replicaFileSyncHTTPClientTimeout int64) (err error) {
+func (p *Proxy) ReplicaAdd(e *longhorn.Engine, replicaName, replicaAddress string, restore, fastSync bool, replicaFileSyncHTTPClientTimeout, grpcTimeoutSeconds int64) (err error) {
 	return p.grpcClient.ReplicaAdd(string(e.Spec.DataEngine), e.Name, e.Spec.VolumeName, p.DirectToURL(e),
 		replicaName, replicaAddress, restore, e.Spec.VolumeSize, e.Status.CurrentSize,
-		int(replicaFileSyncHTTPClientTimeout), fastSync)
+		int(replicaFileSyncHTTPClientTimeout), fastSync, grpcTimeoutSeconds)
```
derekbit, line 10: Should we handle the compatibility of talking to an old instance manager?

PhanLe1010: I think it should work since the old IM will just ignore the new gRPC field. WDYT?

ejweber: I agree with @PhanLe1010. According to https://kubernetes.io/blog/2023/08/25/native-sidecar-containers/#how-did-users-get-sidecar-behavior-before-1-28 (and my experience), it is fine to send the new field to the old service. It will simply be ignored.

For this change, we also have the following scenario if the old instance-manager communicates with a new engine (running inside it) or replica (running elsewhere):

- Old instance-manager ignores `grpcTimeoutSeconds` from longhorn-manager.
- Old instance-manager does not send `grpcTimeoutSeconds` to new engine or replica.
- New engine or replica uses default value (0).

IIRC, this scenario is also fine, since @PhanLe1010's updates handle the default value by using the old `grpcLongTimeout`. Correct me if I'm wrong though.

ejweber (approved):
LGTM.

innobead (approved):

engineapi/proxy_replica.go
PhanLe1010: > IIRC, this scenario is also fine, since @PhanLe1010's updates handle the default value by using the old grpcLongTimeout. Correct me if I'm wrong though.

Yeah, you are correct

PR #2893, opened 2024-06-18
feat(docker-volume): remove volume from dockerfile, no compliant with CIS benchmark

lenglet-k:
 
[Issue#8780](https://github.com/longhorn/longhorn/issues/8780)

This remove volume /usr/local/bin define in dockerfile. In my case, longhorn-manager gets the permission denied error because the noexec option is propagated to the container automatically. in fact, my /var/ volume, used by containerd, has `noexec` option, with this option longhorn-manager is not allow to start because this option not allow the execution of executable binaries in the mounted filesystem

is this volume really used ?

### DIFF
--- a/package/Dockerfile
+++ b/package/Dockerfile
@@ -20,6 +20,5 @@ RUN zypper -n ref && \
 RUN zypper -n install iputils iproute2 nfs-client cifs-utils bind-utils e2fsprogs xfsprogs zip unzip && \
     rm -rf /var/cache/zypp/*
 
-VOLUME /usr/local/sbin
 EXPOSE 9500
 CMD ["launch-manager"]


innobead (approved):

innobead:
@lenglet-k BTW, can you share more about your CIS setup? Longhorn should work if your cluster is hareded somehow like https://docs.rke2.io/security/hardening_guide.

lenglet-k:
Here's my background:

I have a lvm configuration on /var mounted with the noexec option. Containerd is installed in /var/lib/containerd; when containerd goes to mount a container and the latter needs to create a volume, then this volume inherits the mount options, namely noexec.

below aboutthe cis chapter
image

innobead:
> @innobead we following CIS K8S hardening and Linux Hardening

@lenglet-k thanks, very clear.

lenglet-k:
I did the same thing on longhorn-instance-manager: https://github.com/longhorn/longhorn-instance-manager/pull/526

derekbit:
Sorry, I'm confused. Why is the removed volume `/usr/local/sbin` related to the `/var` you mentioned?

derekbit:
> inerd is installed in /var (which has the noexec option) and longhorn-manager has the VOLUME instruction. Then containerd prepares a volume as source /var/lib/containerd/io.containerd.grpc.v1. cri/containers/5080a92e3451e295a7ffc2d4bf10ab18d352124d27eb351ff9c2063b4dbe0d3d/volumes/eca09d7cf90ee824fa5477ef8350c457890c5f09ed622a85de597dbdc60456e6 on host et comme destination /usr/local/bin.
> 
> Below is a return from

Thanks for the explanation. It is clear!

shuo-wu:
`/usr/local/sbin` is used in the following instructions only:
```
COPY package/bin/longhorn-manager-${ARCH} /usr/local/sbin/longhorn-manager

COPY package/launch-manager package/nsmounter /usr/local/sbin/
```

I don't think we need to persist these scripts on the host hence I am fine with this fix.

shuo-wu (approved):

PR #2894, opened 2024-06-19
Improve interaction between volume attachment controller and migrations

ejweber:

longhorn/longhorn#8735

See context at https://github.com/longhorn/longhorn/issues/8735#issue-2347139063 and implementation plan at https://github.com/longhorn/longhorn/issues/8735#issuecomment-2177309258.

~This is a draft for now as I work on testing it in all situations.~

### DIFF
--- a/controller/volume_attachment_controller.go
+++ b/controller/volume_attachment_controller.go
@@ -657,6 +657,17 @@ func selectAttachmentTicketToAttach(va *longhorn.VolumeAttachment, vol *longhorn
 		}
 	}
 
+	// If a volume was migrating and became detached:
+	// - We no longer know which node it was migrating from or to.
+	// - We cannot do an "online" migration anyways, because the volume already crashed.
+	// Now, we refuse to attach the volume again until the caller resolves the situation by reducing the number of
+	// attachment tickets to one.
+	if util.IsMigratableVolume(vol) &&
+		maxAttacherPriorityLevel == longhorn.AttacherPriorityLevelCSIAttacher &&
+		len(highPriorityTicketCandidates) > 1 {
+		return nil
+	}
+
 	// TODO: sort by time
 
 	// sort by name
--- a/controller/volume_attachment_controller.go
+++ b/controller/volume_attachment_controller.go
@@ -398,31 +398,15 @@ func (vac *VolumeAttachmentController) handleVolumeMigrationStart(va *longhorn.V
 		return
 	}
 
-	hasCSIAttachmentTicket := false
-	for _, attachmentTicket := range va.Spec.AttachmentTickets {
-		if attachmentTicket.Type != longhorn.AttacherTypeCSIAttacher {
-			continue
-		}
-		// Found one csi attachmentTicket that is requesting volume to attach to the current node
-		if attachmentTicket.NodeID == vol.Spec.NodeID && verifyAttachmentParameters(attachmentTicket.Parameters, vol) {
-			hasCSIAttachmentTicket = true
-		}
-	}
-
-	if !hasCSIAttachmentTicket {
+	if !hasCSIAttachmentTicketRequestingNode(vol.Spec.NodeID, va, vol) {
 		return
 	}
+	// Found one csi attachmentTicket that is requesting volume to attach to the current node
 
-	for _, attachmentTicket := range va.Spec.AttachmentTickets {
-		if attachmentTicket.Type != longhorn.AttacherTypeCSIAttacher {
-			continue
-		}
+	if attachmentTicket := getCSIAttachmentTicketNotRequestingNode(vol.Spec.NodeID, va, vol); attachmentTicket != nil {
 		// Found one csi attachmentTicket that is requesting volume to attach to a different node
-		if attachmentTicket.NodeID != vol.Spec.NodeID {
-			vol.Spec.MigrationNodeID = attachmentTicket.NodeID
-		}
+		vol.Spec.MigrationNodeID = attachmentTicket.NodeID
 	}
-
 }
 
 func (vac *VolumeAttachmentController) handleVolumeMigrationConfirmation(va *longhorn.VolumeAttachment, vol *longhorn.Volume) {
@@ -431,27 +415,20 @@ func (vac *VolumeAttachmentController) handleVolumeMigrationConfirmation(va *lon
 		return
 	}
 
-	hasCSIAttachmentTicketRequestingPrevNode := false
-	for _, attachmentTicket := range va.Spec.AttachmentTickets {
-		if attachmentTicket.Type != longhorn.AttacherTypeCSIAttacher {
-			continue
-		}
-		// Found one csi attachmentTicket that is requesting volume to attach to the current node
-		if attachmentTicket.NodeID == vol.Spec.NodeID && verifyAttachmentParameters(attachmentTicket.Parameters, vol) {
-			hasCSIAttachmentTicketRequestingPrevNode = true
-			break
-		}
+	if hasCSIAttachmentTicketRequestingNode(vol.Spec.NodeID, va, vol) {
+		return
 	}
+
 	migratingEngineSnapSynced, err := vac.checkMigratingEngineSyncSnapshots(va, vol)
 	if err != nil {
 		vac.logger.WithError(err).Warn("Failed to check migrating engine snapshot status")
+	}
+	if !migratingEngineSnapSynced {
 		return
 	}
-	if !hasCSIAttachmentTicketRequestingPrevNode && migratingEngineSnapSynced {
-		// TODO: Do we need check if the volume is available on the currentMigrationIDNode?
-		vol.Spec.NodeID = vol.Status.CurrentMigrationNodeID
-		vol.Spec.MigrationNodeID = ""
-	}
+
+	vol.Spec.NodeID = vol.Status.CurrentMigrationNodeID
+	vol.Spec.MigrationNodeID = ""
 }
 
 func (vac *VolumeAttachmentController) checkMigratingEngineSyncSnapshots(va *longhorn.VolumeAttachment, vol *longhorn.Volume) (bool, error) {
@@ -511,17 +488,7 @@ func (vac *VolumeAttachmentController) handleVolumeMigrationRollback(va *longhor
 		return
 	}
 
-	hasCSIAttachmentTicketRequestingMigratingNode := false
-	for _, attachmentTicket := range va.Spec.AttachmentTickets {
-		if attachmentTicket.Type != longhorn.AttacherTypeCSIAttacher {
-			continue
-		}
-		// Found one csi attachmentTicket that is requesting volume to attach to the current node
-		if attachmentTicket.NodeID == vol.Spec.MigrationNodeID {
-			hasCSIAttachmentTicketRequestingMigratingNode = true
-		}
-	}
-	if !hasCSIAttachmentTicketRequestingMigratingNode {
+	if !hasCSIAttachmentTicketRequestingNode(vol.Spec.MigrationNodeID, va, vol) {
 		vol.Spec.MigrationNodeID = ""
 	}
 }
@@ -949,4 +916,28 @@ func (vac *VolumeAttachmentController) isVolumeAvailableOnNode(volumeName, node 
 	}
 
 	return false
+}
+
+func hasCSIAttachmentTicketRequestingNode(nodeID string, va *longhorn.VolumeAttachment, vol *longhorn.Volume) bool {
+	for _, attachmentTicket := range va.Spec.AttachmentTickets {
+		if attachmentTicket.Type != longhorn.AttacherTypeCSIAttacher {
+			continue
+		}
+		if attachmentTicket.NodeID == nodeID && verifyAttachmentParameters(attachmentTicket.Parameters, vol) {
+			return true
+		}
+	}
+	return false
+}
+
+func getCSIAttachmentTicketNotRequestingNode(nodeID string, va *longhorn.VolumeAttachment, vol *longhorn.Volume) *longhorn.AttachmentTicket {
+	for _, attachmentTicket := range va.Spec.AttachmentTickets {
+		if attachmentTicket.Type != longhorn.AttacherTypeCSIAttacher {
+			continue
+		}
+		if attachmentTicket.NodeID != nodeID && verifyAttachmentParameters(attachmentTicket.Parameters, vol) {
+			return attachmentTicket
+		}
+	}
+	return nil
 }
--- a/controller/volume_attachment_controller.go
+++ b/controller/volume_attachment_controller.go
@@ -436,6 +436,26 @@ func (vac *VolumeAttachmentController) handleVolumeMigrationConfirmation(va *lon
 		return
 	}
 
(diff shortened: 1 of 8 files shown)


ejweber:
The behaviors in this PR now match the test cases from https://github.com/longhorn/longhorn-tests/pull/1948. It is ready for review.

PhanLe1010:
I will review this today

controller/volume_controller.go
```diff
@@ -3734,50 +3734,49 @@ func (c *VolumeController) processMigration(v *longhorn.Volume, es map[string]*l
...
 		return nil
 	}
 
-	// the only time there should be more then 1 engines is when we are migrating or upgrading
-	// if there are more then 1 and we no longer have a migration id set we can cleanup the extra engine
 	if v.Spec.MigrationNodeID == "" {
+		// The volume attachment controller has stopped the migration (if one was ever started). We must clean up any
+		// extra engines/replicas and leave the volume in a "good" state.
+
+		// The only time there should be more then one engines is when we are migrating or upgrading. If there are more
```
PhanLe1010, line 3753: NIT: Looks like volume should have 1 engine during upgrading. 2 engines only during migrating

controller/volume_controller.go
```diff
@@ -3734,50 +3734,49 @@ func (c *VolumeController) processMigration(v *longhorn.Volume, es map[string]*l
...
 		if len(es) < 2 && v.Status.CurrentMigrationNodeID == "" {
-			return nil
+			return nil // There is nothing to do.
+		}
+
+		// The volume is no longer attached or should no longer be attached. We will clean up the migration below by
+		// removing the extra engine and replicas. Warn the user.
+		if v.Spec.NodeID == "" || v.Status.CurrentNodeID == "" {
+			msg := ("Volume migration failed unexpectedly; detach volume from extra node to resume")
+			c.eventRecorder.Event(v, corev1.EventTypeWarning, constant.EventReasonMigrationFailed, msg)
```
PhanLe1010, line 3763: Conceptual question: is there a rate limit for event emitting?

datastore/longhorn.go
```diff
@@ -1109,6 +1112,29 @@ func GetNewCurrentEngineAndExtras(v *longhorn.Volume, es map[string]*longhorn.En
 		}
 	}
 
+	// Corner case:
+	// During a migration confirmation, we make the following requests to the API server:
+	// 1. Delete the active engine (guaranteed to succeed or we don't do anything else).
+	// 2. Switch the new current replicas to active (might fail, preventing everything below).
+	// 3. Set the new current engine to active (might fail, preventing everything below).
+	// 4. Set the volume.Spec.CurrentNodeID (might fail).
```
PhanLe1010, line 1120: `volume.Status.CurrentNodeID ` instead?

webhook/resources/volumeattachment/validator.go
```diff
@@ -57,6 +60,15 @@ func (v *volumeAttachmentValidator) Update(request *admission.Request, oldObj ru
...
 	}
+	isRemovingLonghornFinalizer, err := common.IsRemovingLonghornFinalizer(oldObj, newObj)
+	if err != nil {
+		err = errors.Wrap(err, "failed to check if removing longhorn.io finalizer from deleted object")
+		return werror.NewInvalidError(err.Error(), "")
+	} else if isRemovingLonghornFinalizer {
+		// We always allow the removal of the longhorn.io finalizer while an object is being deleted. It is the
+		// controller's responsibility to wait for the correct conditions to attempt to remove it.
+		return nil
+	}
```
PhanLe1010, line 71: I am wondering the purpose of this code block. Looks like we check but always allow?

PhanLe1010 (commented):
In general LGTM

controller/volume_controller.go
ejweber: Interesting! Looks like you are correct.

webhook/resources/volumeattachment/validator.go
ejweber: The goal is to always allow the deletion of the Longhorn finalizer, regardless of any other rules in place. The context can be found in https://github.com/longhorn/longhorn/issues/7553 and https://github.com/longhorn/longhorn-manager/pull/2425. The basic idea is that sometimes:

- A controller may choose not to meaningfully continue to reconcile a resource after its deletionTimestamp is set.
- Instead, the controller wants to delete the resource.
- But some webhook check fails and disallows the deletion.

I did not test it, but I thought this may be a possible situation introduced by these changes.

- The volume attachment controller sees the volume is gone and the volume attachment is being deleted. It decids to remove the finalizer for the volume attachment. 
  https://github.com/longhorn/longhorn-manager/blob/ffe4a02e240cee2299967c38ab3ab37a6d13abf0/controller/volume_attachment_controller.go#L293-L302
- At this moment the new validation logic I wrote disallows updating the volume attachment.
- The situation never changes (since the volume attachment controller no longer meaningfully reconciles).

Since the goal in https://github.com/longhorn/longhorn/issues/7553 is to add this logic to all resources in the webhook eventually, I thought it was safest to add it here now while adding logic to avoid any issues. WDYT?

controller/volume_controller.go
ejweber: Good question. I looked into this a bit.

- The volume attachment controller creates an EventBroadcaster with the default options.
  https://github.com/longhorn/longhorn-manager/blob/e5701524c9610cb780ef06960d8ede62f25fa7ec/controller/volume_attachment_controller.go#L55
- This leads to the use of an EventCorrelator with default options.
  https://github.com/longhorn/longhorn-manager/blob/e5701524c9610cb780ef06960d8ede62f25fa7ec/vendor/k8s.io/client-go/tools/record/events_cache.go#L477-L507
- The default options allow a source to send a burst of 25 events, but the "bucket" used only refills at a rate of 1 event per 300 seconds.
  https://github.com/longhorn/longhorn-manager/blob/e5701524c9610cb780ef06960d8ede62f25fa7ec/vendor/k8s.io/client-go/tools/record/events_cache.go#L44-L49
- Some more details about how this works.
  https://github.com/longhorn/longhorn-manager/blob/e5701524c9610cb780ef06960d8ede62f25fa7ec/vendor/k8s.io/client-go/tools/record/events_cache.go#L424-L438

TBH, I'm not 100% sure of the effects in relation to your question. It looks to me like the EventBroadcaster will eventually stop broadcasting events from this source (volume controller) for this object (volume) if we try to spam them. And identical events will not appear twice, but instead have their count updated.

In practice, I think the only way we can really hit the rate limit is if we repeatedly fail to pick the new current engine or repeatedly fail to delete the old engine.

ejweber: I inserted a test error just after emitting the event to see what happened. (So, after purposely causing a MigrationFailed event, Longhorn could never recover.)

- Only one event was ever created, with its `count` increasing over time.
- On the Longhorn side, we try (and fail) to `processMigration` four times every 30 seconds. (Every volume queues every thirty seconds and there are a maximum of three retries.)
- On the event side, after a short while, we only update the event once every five minutes. When we do, though, we increase the count by the number of times we called `Event()` in the interim.

```
# Before an interval "rollover"

eweber@laptop:~/longhorn> kl get event pvc-82ebabc8-777d-4939-9198-af33320a379e.17de7c648a6d2cd6
LAST SEEN   TYPE      REASON            OBJECT                                            MESSAGE
4m55s       Warning   MigrationFailed   volume/pvc-82ebabc8-777d-4939-9198-af33320a379e   Volume migration failed unexpectedly; detach volume from extra node to resume

eweber@laptop:~/longhorn> kl get event pvc-82ebabc8-777d-4939-9198-af33320a379e.17de7c648a6d2cd6 -oyaml
apiVersion: v1
count: 245
eventTime: null
firstTimestamp: "2024-07-02T19:34:38Z"
involvedObject:
  kind: Volume
  name: pvc-82ebabc8-777d-4939-9198-af33320a379e
  namespace: longhorn-system
  resourceVersion: "120809419"
  uid: 57f11de5-f73c-42a8-88bd-3e72efe0a4a1
kind: Event
lastTimestamp: "2024-07-02T20:05:06Z"
message: Volume migration failed unexpectedly; detach volume from extra node to resume
metadata:
  creationTimestamp: "2024-07-02T19:34:38Z"
  name: pvc-82ebabc8-777d-4939-9198-af33320a379e.17de7c648a6d2cd6
  namespace: longhorn-system
  resourceVersion: "120818848"
  uid: d3fc0f66-b4c4-47c6-835a-ef1fbb9f1933
reason: MigrationFailed
reportingComponent: longhorn-volume-controller
reportingInstance: ""
source:
  component: longhorn-volume-controller
type: Warning

# After an interval "rollover"

eweber@laptop:~/longhorn> kl get event pvc-82ebabc8-777d-4939-9198-af33320a379e.17de7c648a6d2cd6
LAST SEEN   TYPE      REASON            OBJECT                                            MESSAGE
1s          Warning   MigrationFailed   volume/pvc-82ebabc8-777d-4939-9198-af33320a379e   Volume migration failed unexpectedly; detach volume from extra node to resume

eweber@laptop:~/longhorn> kl get event pvc-82ebabc8-777d-4939-9198-af33320a379e.17de7c648a6d2cd6 -oyaml
apiVersion: v1
count: 285
eventTime: null
firstTimestamp: "2024-07-02T19:34:38Z"
involvedObject:
  kind: Volume
  name: pvc-82ebabc8-777d-4939-9198-af33320a379e
  namespace: longhorn-system
  resourceVersion: "120809419"
  uid: 57f11de5-f73c-42a8-88bd-3e72efe0a4a1
kind: Event
lastTimestamp: "2024-07-02T20:10:06Z"
message: Volume migration failed unexpectedly; detach volume from extra node to resume
metadata:
  creationTimestamp: "2024-07-02T19:34:38Z"
  name: pvc-82ebabc8-777d-4939-9198-af33320a379e.17de7c648a6d2cd6
  namespace: longhorn-system
  resourceVersion: "120820386"
  uid: d3fc0f66-b4c4-47c6-835a-ef1fbb9f1933
reason: MigrationFailed
reportingComponent: longhorn-volume-controller
reportingInstance: ""
source:
  component: longhorn-volume-controller
type: Warning
```

I am not inclined to make changes here because:

- We should not be able to remain in this state indefinitely.
- Even if we do, the above client-side rate limiting will keep us from hammering the API server (eventually).
- It seems there are other locations in the volume controller at which we would also generate events similarly if some unknown bug surfaced.
- The solution (I think) would have to be either an additional in-memory backoff in addition to the mechanism built into the event client.

webhook/resources/volumeattachment/validator.go
PhanLe1010: Sounds good to me. Thank you for the explanation

PhanLe1010 (approved):
LGTM

controller/volume_attachment_controller.go
```diff
@@ -436,6 +436,26 @@ func (vac *VolumeAttachmentController) handleVolumeMigrationConfirmation(va *lon
...
+	// under these conditions.
+	// - Kubernetes does not want it attached to the old node.
+	// - We can't get a migration engine up on the new down node.
+	// Stop the migration and detach from the old node. If the upper layer changes nothing and the new node comes back,
+	// the volume can exclusively attach to it.
+	if downOrDeleted, err := vac.ds.IsNodeDownOrDeleted(vol.Status.CurrentMigrationNodeID); err != nil {
+		log.WithError(err).Warn("Failed to check if node is down")
+	} else if downOrDeleted {
+		vol.Spec.NodeID = ""
+		vol.Spec.MigrationNodeID = ""
```
shuo-wu, line 449: I was thinking the migration would be confirmed as usual, and the volume would become detached when the new engine (the related IM pod) becomes terminating. 
In other words, should this controller be responsible for detaching the volumes? Can we still rely on the engine crash logic inside VolumeController?

datastore/longhorn.go
```diff
@@ -1109,6 +1112,29 @@ func GetNewCurrentEngineAndExtras(v *longhorn.Volume, es map[string]*longhorn.En
...
+	// So we might delete the active engine, fail to set the new current engine to active, and fail to set the
+	// volume.Spec.CurrentNodeID. Then, the volume attachment controller might try to do a full detachment, setting
+	// volume.Spec.NodeID == "". At this point, there is no engine that can ever be considered active again by the rules
+	// above, but we want to use the new current engine. It's engine.Spec.NodeID ==
+	// volume.Status.CurrentMigrationNodeID.
```
shuo-wu, line 1125: > So we might delete the active engine, fail to set the volume.Spec.CurrentNodeID.

> At this point, there is no engine that can ever be considered active again by the rules above, but we want to use the new current engine

Will Longhorn retry marking this new engine as active later?

shuo-wu (approved):
In general LGTM

controller/volume_attachment_controller.go
ejweber: Good suggestion. I tried this out with the following changes (essentially allowing a normal migration confirmation instead of a detachment):

```
        if downOrDeleted, err := vac.ds.IsNodeDownOrDeleted(vol.Status.CurrentMigrationNodeID); err != nil {
                log.WithError(err).Warn("Failed to check if node is down")
        } else if downOrDeleted {
-               vol.Spec.NodeID = ""
+               vol.Spec.NodeID = vol.Status.CurrentMigrationNodeID
                vol.Spec.MigrationNodeID = ""
                log = getLoggerForMigratingLHVolumeAttachment(vac.logger, va, vol)
-               log.Warn("Detaching volume for attempted migration to down node")
+               log.Warn("Doing Shuo's suggested behavior")
                return
        }
```

In the couple of tests I did, everything worked fine. The `currentNodeID` switched as expected and the usual reattachment logic started trying to reattach to the new node (which was down).

However, this behavior makes me nervous. In the past, we have never allowed a migration confirmation when the new engine is not available. I think doing so leaves us open to potential races that I have not accounted for.

> In other words, should this controller be responsible for detaching the volumes?

This is the only controller that is responsibly for detaching volumes. While the usual logic that sets `vol.Spec.NodeID = ""` is in a different function, it is a different function within this controller.

I think it is safest to maintain the already implemented/tested behavior. In practice, it is very similar to your suggestion, but it does not switch the active engine and replicas (which is the part that is a bit dicey when the migration engine is actually not running IMO).

datastore/longhorn.go
ejweber: No. We will only set an engine to active if it is returned from `GetNewCurrentEngineAndExtras`.

https://github.com/longhorn/longhorn-manager/blob/e5701524c9610cb780ef06960d8ede62f25fa7ec/controller/volume_controller.go#L3849

This function disallows engines with deletion timestamps.

https://github.com/longhorn/longhorn-manager/blob/e5701524c9610cb780ef06960d8ede62f25fa7ec/datastore/longhorn.go#L1101-L1103

ejweber: Discussed this with @shuo-wu offline. I actually misunderstood his question. The answer is yes. Steps 2-4 modify the associated objects mid-reconcile, but don't attempt to make changes on the API server until the end. If some error occurs (e.g. a conflict), we will rereconcile and repeat the actions.

controller/volume_attachment_controller.go
ejweber: Discussed this with @shuo-wu offline and we are fine to proceed with the current implementation.

PR #2895, opened 2024-06-19
Volume clone enhancement 

PhanLe1010:
1. Update the checkVolumeCreated condition for CSI CreateVolume call.
Sometime the volume is auto-attached quickly after being created. For
example, volume which require backup restore or volume require data
cloning. We shouldn't wait for the volume to be detached before return
OK for CreateVolume call because it could take a very long time to
finish the data transfer.
1. Implement the retry logic with exponential backoff for volume cloning

longhorn/longhorn#2907


util/util.go
```diff
@@ -236,6 +236,10 @@ func Now() string {
 	return time.Now().UTC().Format(time.RFC3339)
 }
 
+func DurationAfterNow(d time.Duration) string {
```
ejweber, line 239: I don't think this name makes sense. `DurationAfterNow` seems like it would return a duration instead of a UTC formatted timestamp.

Should it be `TimestampAfterDuration`?

csi/controller_server.go
```diff
@@ -440,6 +443,8 @@ func (cs *ControllerServer) ControllerPublishVolume(ctx context.Context, req *cs
 		return nil, status.Errorf(codes.Aborted, "volume %s is not ready for workloads", volumeID)
 	}
 
+	// TODO: make sure that volume is not ready to be attached if the clone hasn't finished
```
ejweber, line 446: This is resolved in resolved in `toVolumeResource`, isn't it?

ejweber (commented):
LGMT in general. A couple of nits.

Do we need to make the number of retries and initial time user configurable? (I think I can pretty easily be convinced now, but it would be good to write down a justification.)

PhanLe1010:
> Do we need to make the number of retries and initial time user configurable? (I think I can pretty easily be convinced now, but it would be good to write down a justification.)

I would like to avoid introducing more settings and API changes (for updating per volume setting). Maybe we can hard-code it for now and add customization later if there are users/custumers' requests for it. WDYT?

util/util.go
PhanLe1010: Thank you. Fixed it

csi/controller_server.go
PhanLe1010: You are right. I forgot to remove this comment. Thank you. Fixed

ejweber (approved):
LGTM.

PhanLe1010:
Ping @ChanYiLin Would you be able to take a look to see if there is any concern before merging?

PhanLe1010:
Merging. I will push a new PR if there are additional review comes in

PR #2898, opened 2024-06-20
Only need to force ignore xfs uuid in NodeStageVolume

PhanLe1010:
Like @shuo-wu mentioned in the US sync meeting, we don't need to do this NodePublishVolume because bind mount will preserve the mount options from NodeStageVolume

longhorn/longhorn#8796


shuo-wu (approved):

PhanLe1010:

PhanLe1010:
@mergify backport v1.6.x

PhanLe1010:
@mergify backport v1.5.x

PR #2913, opened 2024-06-24
fix(setting): create a real SC for the setting "default-longhorn-static-storage-class"

mantissahz:

Issue # longhorn/longhorn#6446

### DIFF
--- a/controller/setting_controller.go
+++ b/controller/setting_controller.go
@@ -19,6 +19,7 @@ import (
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	storagev1 "k8s.io/api/storage/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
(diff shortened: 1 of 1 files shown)


controller/setting_controller.go
```diff
@@ -965,6 +970,34 @@ func (sc *SettingController) updateLogLevel() error {
...
+	_, err = sc.ds.GetStorageClassRO(defaultStaticStorageClassName)
+	if err != nil && apierrors.IsNotFound(err) {
+		allowVolumeExpansion := true
+		storageClass := &storagev1.StorageClass{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: defaultStaticStorageClassName,
+			},
+			Provisioner:          types.LonghornDriverName,
+			AllowVolumeExpansion: &allowVolumeExpansion,
+			Parameters:           map[string]string{"staleReplicaTimeout": "30"},
```
derekbit, line 989: Only 30 seconds?

mantissahz: That's 30 minutes. https://github.com/longhorn/longhorn/issues/6446#issuecomment-1861816858

derekbit (approved):
LGTM.

Also need to write a KB about a workaround for users with PVC with longhorn-static SC in older version.

innobead:

PR #2918, opened 2024-06-28
fix(replica-scheduling): power off replica node should not rebuild new replica on same node

c3y1huang:

Issue longhorn/longhorn#1992

- Before creating a new Replica CR, verify the availability of suitable nodes with sufficient disk to accommodate the replica. This prevents creating Replicas that cannot be scheduled.
- During node disk candidate selection, exclude nodes where the InstanceManager is not in a running state. This prevents creating redundant Replica CRs on nodes that haven't fully recovered after a power down.

`None`

`None`

### DIFF
--- a/controller/volume_controller_test.go
+++ b/controller/volume_controller_test.go
@@ -119,8 +119,9 @@ func (s *TestSuite) TestVolumeLifeCycle(c *C) {
 		tc.nodes[i].Spec.AllowScheduling = false
 	}
 	tc.copyCurrentToExpect()
-	// replicas and engine object would still be created
+	// replica object would not be created
 	tc.replicas = nil
+	// engine object would still be created
 	tc.engines = nil
 	for _, r := range tc.expectReplicas {
 		r.Spec.NodeID = ""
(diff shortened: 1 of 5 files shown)


c3y1huang:
Testing WIP.

c3y1huang:
Test result: https://ci.longhorn.io/job/private/job/longhorn-e2e-test/669/

controller/volume_controller_test.go
```diff
@@ -119,15 +119,38 @@ func (s *TestSuite) TestVolumeLifeCycle(c *C) {
...
 		r.Spec.DiskID = ""
 		r.Spec.DiskPath = ""
 		r.Spec.DataDirectoryName = ""
 	}
+	tc.expectVolume.Status.State = longhorn.VolumeStateCreating
+	tc.expectVolume.Status.CurrentImage = tc.volume.Spec.Image
+	tc.expectVolume.Status.Robustness = longhorn.VolumeRobustnessFaulted
+	testCases["volume create - replica creation failure"] = tc
+
+	// unable to create volume because no node to schedule
```
ejweber, line 137: Can you help me to understand this test case? If I understand correctly, the one above (`volume create - replica creation failure`) is now expected. (We will try not to create replicas that cannot be scheduled.) In this test case, the replicas are already created but cannot be scheduled (as was the old behavior). Is this a state we expect to be in? Or is it just to cover a corner case that probably won't be triggered?

ejweber (approved):
LGTM.

I do have one question about a test case, but it is for my understanding and not a blocker.

innobead (approved):

c3y1huang:
@mergify backport v1.5.x v1.6.x

controller/volume_controller_test.go
c3y1huang: This is likely a corner case. In my opinion, the unit test will give us a better understanding of the behavior.

PR #2920, opened 2024-06-28
feat(webhook selector): add longhorn-manager labels when webhook ready.

james-munson:

Issue longhorn/longhorn#8803 and longhorn/longhorn#8612

Add the longhorn-manager pod labels to be used as webhook service selectors dynamically after the webhooks have been started.

Note that this PR **must** be accompanied by the manifest changes in PR https://github.com/longhorn/longhorn/pull/8804.  They depend on each other.

### DIFF
--- a/app/daemon.go
+++ b/app/daemon.go
@@ -1,6 +1,7 @@
 package app
 
 import (
+	"context"
 	"fmt"
 	"net/http"
 	_ "net/http/pprof" // for runtime profiling
@@ -11,6 +12,7 @@ import (
 	"github.com/rancher/wrangler/pkg/signals"
 	"github.com/sirupsen/logrus"
 	"github.com/urfave/cli"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/longhorn/go-iscsi-helper/iscsi"
 
@@ -144,6 +146,16 @@ func startManager(c *cli.Context) error {
 		return fmt.Errorf("failed to detect the node IP")
 	}
 
+	podName, err := util.GetRequiredEnv(types.EnvPodName)
+	if err != nil {
+		return fmt.Errorf("failed to detect the manager pod name")
+	}
+
+	podNamespace, err := util.GetRequiredEnv(types.EnvPodNamespace)
+	if err != nil {
+		return fmt.Errorf("failed to detect the manager pod namespace")
+	}
+
 	ctx := signals.SetupSignalContext()
 
 	logger := logrus.StandardLogger().WithField("node", currentNodeID)
@@ -151,12 +163,24 @@ func startManager(c *cli.Context) error {
 	// Conversion webhook needs to be started first since we use its port 9501 as readiness port.
 	// longhorn-manager pod becomes ready only when conversion webhook is running.
 	// The services in the longhorn-manager can then start to receive the requests.
-	// Conversion webhook does not longhorn datastore.
+	// Conversion webhook does not use datastore, since it is a prerequisite for
+	// datastore operation.
 	clientsWithoutDatastore, err := client.NewClients(kubeconfigPath, false, ctx.Done())
 	if err != nil {
 		return err
 	}
 	if err := webhook.StartWebhook(ctx, types.WebhookTypeConversion, clientsWithoutDatastore); err != nil {
+		return err
+	}
+
+	// This adds the label for the conversion webhook's selector.  We do it the hard way without datastore to avoid chicken-and-egg.
+	pod, _ := clientsWithoutDatastore.Clients.K8s.CoreV1().Pods(podNamespace).Get(context.Background(), podName, v1.GetOptions{})
+	labels := types.GetConversionWebhookLabel()
+	for key, value := range labels {
+		pod.Labels[key] = value
+	}
+	_, err = clientsWithoutDatastore.Clients.K8s.CoreV1().Pods(podNamespace).Update(context.Background(), pod, v1.UpdateOptions{})
+	if err != nil {
 		return err
 	}
 	if err := webhook.CheckWebhookServiceAvailability(types.WebhookTypeConversion); err != nil {
@@ -167,7 +191,11 @@ func startManager(c *cli.Context) error {
 	if err != nil {
 		return err
 	}
+
 	if err := webhook.StartWebhook(ctx, types.WebhookTypeAdmission, clients); err != nil {
+		return err
+	}
+	if err := clients.Datastore.AddLabelToManagerPod(currentNodeID, types.GetAdmissionWebhookLabel()); err != nil {
 		return err
 	}
 	if err := webhook.CheckWebhookServiceAvailability(types.WebhookTypeAdmission); err != nil {
@@ -209,7 +237,7 @@ func startManager(c *cli.Context) error {
 		return err
 	}
 
-	if err := initDaemonNode(clients.Datastore); err != nil {
+	if err := initDaemonNode(clients.Datastore, currentNodeID); err != nil {
 		return err
 	}
 
@@ -284,8 +312,7 @@ func updateRegistrySecretName(m *manager.VolumeManager) error {
 	return nil
 }
 
-func initDaemonNode(ds *datastore.DataStore) error {
-	nodeName := os.Getenv("NODE_NAME")
+func initDaemonNode(ds *datastore.DataStore, nodeName string) error {
 	if _, err := ds.GetNode(nodeName); err != nil {
 		// init default disk on node when starting longhorn-manager
 		if datastore.ErrorIsNotFound(err) {
--- a/datastore/kubernetes.go
+++ b/datastore/kubernetes.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/sirupsen/logrus"
 
+	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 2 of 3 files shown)


derekbit:
As I mentioned in https://github.com/longhorn/longhorn/issues/8803#issuecomment-2192856444, the PR only addes the labels to webhook but doesn't handle the deletion. Could you help clear up my confusion?

james-munson:
I hope so.  This PR (and 8805 in longhorn) just add the new selector labels.  Should make no difference to normal operation.  After it is committed, then I can pull the changes into my RWX HA branch, because the logic for deletion and restoration depends on conditions and checks that only exist in that branch.
I did it this way, rather than just introducing it all in that branch, because it does offer a fix for the situation described by @PhanLe1010 in #8612

derekbit:
> I hope so. This PR (and [longhorn/longhorn#8804](https://github.com/longhorn/longhorn/pull/8804) in longhorn) just add the new selector labels. Should make no difference to normal operation. After it is committed, then I can pull the changes into my RWX HA branch, because the logic for deletion and restoration depends on conditions and checks that only exist in that branch. I did it this way, rather than just introducing it all in that branch, because it does also offer a fix for the situation described by @PhanLe1010 in [longhorn/longhorn#8612](https://github.com/longhorn/longhorn/issues/8612).

I see. Then, let's merge the PR after RWX HA branch is ready for review. Then, we can see how webhook selector works in the implementation. WDYT?

app/daemon.go
```diff
@@ -151,23 +151,30 @@ func startManager(c *cli.Context) error {
...
+	// Conversion webhook does not need or use longhorn datastore.
 	clientsWithoutDatastore, err := client.NewClients(kubeconfigPath, false, ctx.Done())
 	if err != nil {
 		return err
 	}
+	clients, err := client.NewClients(kubeconfigPath, true, ctx.Done())
+	if err != nil {
+		return err
+	}
+
```
PhanLe1010, line 163: I think this change might be problematic. We are everting the fix https://github.com/longhorn/longhorn-manager/pull/2812. AKA, we are trying to start dastore before the conversion webhook which might create a deadlock since the datastore may have to wait for the conversion webhook to be running first 

WDYT?

app/daemon.go
```diff
@@ -151,23 +151,30 @@ func startManager(c *cli.Context) error {
...
 	if err := webhook.StartWebhook(ctx, types.WebhookTypeConversion, clientsWithoutDatastore); err != nil {
 		return err
 	}
```
PhanLe1010, line 166: I think this change might be problematic. We are everting the fix https://github.com/longhorn/longhorn-manager/pull/2812. AKA, we are trying to start dastore before the conversion webhook which might create a deadlock since the datastore may have to wait for the conversion webhook to be running first 

WDYT?

ejweber: Good catch!

I think we can preserve this approach pretty easily though by changing how it works. We already make a number of details about the pod available to its running process using the downwardAPI. `metadata.name` is not one of the fields we use, but we can.

```
env:
    - name: POD_NAMESPACE
      valueFrom:
        fieldRef:
          apiVersion: v1
          fieldPath: metadata.namespace
    - name: POD_IP
      valueFrom:
        fieldRef:
          apiVersion: v1
          fieldPath: status.podIP
    - name: NODE_NAME
      valueFrom:
        fieldRef:
          apiVersion: v1
          fieldPath: spec.nodeName
```

https://kubernetes.io/docs/concepts/workloads/pods/downward-api/

It would look something like this:

```go
pod, _ := clientsWithoutDatastore.Clients.K8s.CoreV1().Pods(os.Getenv(types.EnvPodNamespace)).Get(context.Background(), os.Getenv("POD_NAME"), v1.GetOptions{})
pod.Labels["key"] = "value"
clientsWithoutDatastore.Clients.K8s.CoreV1().Pods(os.Getenv(types.EnvPodNamespace)).Update(context.Background(), pod, v1.UpdateOptions{})
```

This strategy is not quite as nice as using the datastore function you are planning to use for other interactions with the webhook labels, but it is (I think) a solid way for a longhorn-manager pod to add the necessary label TO ITSELF without initializing the datastore.

james-munson: I'll buy that.  I did wonder a bit, but it was happy in my testing.  I can see where it might set up a race with a possible deadlock, though.  Thanks, @ejweber for the suggestion.

james-munson: For what it is worth, here are the logs from a run with the previous method.  As you can see, the localhost test passes even before the call to datastore to modify the p2024-06-28T22:34:06.978508405Z I0628 22:34:06.978322       1 shared_informer.go:320] Caches are synced for longhorn datastore
2024-06-28T22:34:06.978630751Z time="2024-06-28T22:34:06Z" level=info msg="Starting longhorn conversion webhook server" func=webhook.StartWebhook file="webhook.go:24"
2024-06-28T22:34:06.978639002Z time="2024-06-28T22:34:06Z" level=info msg="Waiting for conversion webhook to become ready" func=webhook.StartWebhook file="webhook.go:43"
2024-06-28T22:34:06.981596794Z time="2024-06-28T22:34:06Z" level=warning msg="Failed to check endpoint https://localhost:9501/v1/healthz" func=webhook.isServiceAvailable file="webhook.go:78" error="Get \"https://localhost:9501/v1/healthz\": dial tcp [::1]:9501: connect: connection refused"
2024-06-28T22:34:06.988202655Z time="2024-06-28T22:34:06Z" level=info msg="Active TLS secret longhorn-system/longhorn-webhook-tls (ver=51640559) (count 2): map[listener.cattle.io/cn-longhorn-admission-webhook.longhor-59584d:longhorn-admission-webhook.longhorn-system.svc listener.cattle.io/cn-longhorn-conversion-webhook.longho-6a0089:longhorn-conversion-webhook.longhorn-system.svc listener.cattle.io/fingerprint:SHA1=A22EFDA2A84D7A4EFD53D3A7926D1ACA54A52804]" func="memory.(*memory).Update" file="memory.go:42"
2024-06-28T22:34:06.988244638Z time="2024-06-28T22:34:06Z" level=info msg="Listening on :9501" func=server.ListenAndServe.func2 file="server.go:77"
2024-06-28T22:34:07.315294352Z time="2024-06-28T22:34:07Z" level=info msg="Starting /v1, Kind=Secret controller" func="controller.(*controller).run" file="controller.go:148"
2024-06-28T22:34:07.315566430Z time="2024-06-28T22:34:07Z" level=info msg="Starting apiextensions.k8s.io/v1, Kind=CustomResourceDefinition controller" func="controller.(*controller).run" file="controller.go:148"
2024-06-28T22:34:07.315814195Z time="2024-06-28T22:34:07Z" level=info msg="Building conversion rules..." func="server.(*WebhookServer).runConversionWebhookListenAndServe.func1" file="server.go:193"
2024-06-28T22:34:07.316297058Z time="2024-06-28T22:34:07Z" level=info msg="Starting apiregistration.k8s.io/v1, Kind=APIService controller" func="controller.(*controller).run" file="controller.go:148"
2024-06-28T22:34:07.321634322Z time="2024-06-28T22:34:07Z" level=info msg="Updating TLS secret for longhorn-system/longhorn-webhook-tls (count: 2): map[listener.cattle.io/cn-longhorn-admission-webhook.longhor-59584d:longhorn-admission-webhook.longhorn-system.svc listener.cattle.io/cn-longhorn-conversion-webhook.longho-6a0089:longhorn-conversion-webhook.longhorn-system.svc listener.cattle.io/fingerprint:SHA1=A22EFDA2A84D7A4EFD53D3A7926D1ACA54A52804]" func="kubernetes.(*storage).saveInK8s" file="controller.go:225"
2024-06-28T22:34:08.982742691Z time="2024-06-28T22:34:08Z" level=info msg="Started longhorn conversion webhook server on localhost" func=webhook.StartWebhook file="webhook.go:47"
2024-06-28T22:34:09.024365655Z time="2024-06-28T22:34:09Z" level=info msg="conversion webhook service is now accessible" func=webhook.CheckWebhookServiceAvailability file="webhook.go:63"

```

james-munson: Updated as suggested to avoid datastore use for conversion webhook setup.

ejweber: > For what it is worth, here are the logs from a run with the previous method. As you can see, the localhost test passes even before the call to datastore to modify the pod

IIUC, it is only an issue in very specific cases (which are hopefully no longer relevant). In the case @PhanLe1010 fixed:

- Datastore initialization failed due to a BackupTarget using an outdated version (`v1alpha1`) and no conversion webhook.
- Since datastore initialization failed, we couldn't get a conversion webhook up.

Realistically, any cluster you use that isn't super old or intentionally broken has no chicken/egg issue.

PhanLe1010: Agree with Eric that the cluster needs to be old to trigger the issue (start using Longhorn since v1.2.6).

Though some NITs: 
* the outdated version can be `v1beta1`, not necessarily `v1alpha1`
* Not only backup target CR, it can also happen with engine image CR, ...
* Arguably non trivial chance of happening since it only require user to install Longhorn < 1.2.x and upgrade to Longhorn >= 1.4.x

PhanLe1010 (commented):
Looks reasonable to me

james-munson:
The change in this PR has been pulled into RWX HA PR https://github.com/longhorn/longhorn-manager/pull/2811.  I will close this one, since it is unlikely to be adopted on its own.

PR #2921, opened 2024-06-29
Remove headless services for instance manager pods

PhanLe1010:
See explanation at https://github.com/longhorn/longhorn/issues/8844#issuecomment-2197769331

longhorn/longhorn#8844

### DIFF
--- a/upgrade/v16xto170/upgrade.go
+++ b/upgrade/v16xto170/upgrade.go
@@ -1,9 +1,12 @@
 package v16xto170
 
 import (
+	"context"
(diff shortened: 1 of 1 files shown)


innobead (approved):
LGTM.

innobead:
Yeah. Those temporary patched services were mistakenly merged for 1.2.4 and continued to be used in the following feature and patch versions. This is a good fix, even though the orphaned headless services should not hurt.

innobead:

PR #2938, opened 2024-07-02
Fix bug: don't fail post upgrade check if longhorn-manager pod has multiple containers

PhanLe1010:
It is expected that Longhorn manager pod has multiple contains (some of them are image pre-pull containers)

longhorn/longhorn#8866


derekbit:
> It is expected that Longhorn manager pod has multiple contains (some of them are image pre-pull containers)
> 
> [longhorn/longhorn#8866](https://github.com/longhorn/longhorn/issues/8866)

Thanks @PhanLe1010.
So fast. I should assign it to @mantissahz ...

derekbit (approved):

PR #2939, opened 2024-07-02
fix(volume): wrong scheduled condition

c3y1huang:

Issue longhorn/longhorn#8867

- Set volume condition scheduled to false when failed at replica creation precheck.
- Add a condition to check if the Replica count equals or is greater then volume.Spec.NumberOfReplicas before set volume condition scheduled to true.

`None`

- https://github.com/longhorn/longhorn-manager/pull/2918

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -1773,7 +1773,7 @@ func (c *VolumeController) reconcileVolumeCondition(v *longhorn.Volume, e *longh
 	}
 
 	failureMessage := ""
-	if scheduled {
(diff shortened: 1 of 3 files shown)


ejweber (approved):
I thought there might be a weird interaction with `allow-volume-creation-with-degraded-availability`, but a quick test showed this was not the case.

LGTM.

c3y1huang:
Test result: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/7092/

derekbit (approved):
LGTM

PR #2941, opened 2024-07-02
fix(volume-controller): replica CR not created when data-locality set best-effort

c3y1huang:

Issue longhorn/longhorn#8869

Bypass the precheck when hardNodeAffinity is provided, as users would expect a new Replica CR to be created in this scenario.

`None`

https://github.com/longhorn/longhorn-manager/pull/2918

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -2247,12 +2247,16 @@ func (c *VolumeController) replenishReplicas(v *longhorn.Volume, e *longhorn.Eng
 		if checkBackDuration := c.scheduler.RequireNewReplica(rs, v, hardNodeAffinity); checkBackDuration == 0 {
 			newReplica := c.newReplica(v, e, hardNodeAffinity)
 
-			if err := c.precheckCreateReplica(newReplica, rs, v); err != nil {
-				log.WithError(err).Warnf("Unable to create new replica %v", newReplica.Name)
-				v.Status.Conditions = types.SetCondition(v.Status.Conditions,
-					longhorn.VolumeConditionTypeScheduled, longhorn.ConditionStatusFalse,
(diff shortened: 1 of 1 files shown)


c3y1huang:
Test result: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/7094/

ejweber (commented):
If I understand correctly, the failure observed in https://github.com/longhorn/longhorn/issues/8869 is caused by:

- We cannot clean up an extra healthy replica automatically until a replica with HardNodeAffinity set to the engine's node exists.
- Since the relevant node is unschedulable, the precheck refuses to allow the creation of this replica.

If I am correct, I have no issues with the implementation. Can you provide a comment with some context here? I think it will be difficult to understand why precheck is disabled for best effort data locality when making changes in the future.

c3y1huang:
> If I understand correctly, the failure observed in [longhorn/longhorn#8869](https://github.com/longhorn/longhorn/issues/8869) is caused by:
> 
> * We cannot clean up an extra healthy replica automatically until a replica with HardNodeAffinity set to the engine's node exists.
> * Since the relevant node is unschedulable, the precheck refuses to allow the creation of this replica.
> 
> If I am correct, I have no issues with the implementation. Can you provide a comment with some context here? I think it will be difficult to understand why precheck is disabled for best effort data locality when making changes in the future.

Yes, you are correct. Also, your comment reminded me that we should probably bypass replica creation when hardNodeAffinity is provided, rather than relying on data-locality best effort.

c3y1huang:
Test WIP: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/7108/

derekbit (approved):
LGTM

PR #2942, opened 2024-07-03
fix(volume-controller): mark volume as scheduled when no more replicas to replenish

c3y1huang:

Issue longhorn/longhorn#8872

Before marking the volume as scheduled, we need to check the replica replenish count. Otherwise, the controller might accidentally mark it as scheduled when the replica count is correct, but new replicas are actually expected and are being blocked by the pre-check.

`None`

`None`


c3y1huang:
Test WIP: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/7123/

derekbit (approved):

PR #2943, opened 2024-07-03
fix(recurring): fix validation typo

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/8868

fix typo
should validate on parameters instead of lables


derekbit (approved):
LGTM

PR #2944, opened 2024-07-03
Bump the version to v1.6.3-dev so that upgrade path v162to163 is executed

PhanLe1010:
More details at https://github.com/longhorn/longhorn/issues/8858#issuecomment-2207497386


derekbit (approved):
LGTM

PR #2945, opened 2024-07-03
Bump the version to v1.5.6-dev so that upgrade path v155to156 is executed

PhanLe1010:
More details at https://github.com/longhorn/longhorn/issues/8858#issuecomment-2207497386


derekbit (approved):
LGTM

PR #2946, opened 2024-07-04
fix(recurring): fix update recurring job parameters does not work issue

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/8291

fix update recurring job parameters doesn't work issue

### DIFF
--- a/manager/recurringjob.go
+++ b/manager/recurringjob.go
@@ -65,11 +65,7 @@ func (m *VolumeManager) UpdateRecurringJob(spec longhorn.RecurringJobSpec) (*lon
 	}
 	sort.Strings(recurringJob.Spec.Groups)
 	sort.Strings(spec.Groups)
-	if recurringJob.Spec.Cron == spec.Cron &&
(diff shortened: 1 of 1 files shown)


derekbit (approved):
LGTM

manager/recurringjob.go
```diff
@@ -69,14 +69,16 @@ func (m *VolumeManager) UpdateRecurringJob(spec longhorn.RecurringJobSpec) (*lon
 		reflect.DeepEqual(recurringJob.Spec.Groups, spec.Groups) &&
 		recurringJob.Spec.Retain == spec.Retain &&
 		recurringJob.Spec.Concurrency == spec.Concurrency &&
-		reflect.DeepEqual(recurringJob.Spec.Labels, spec.Labels) {
+		reflect.DeepEqual(recurringJob.Spec.Labels, spec.Labels) &&
+		reflect.DeepEqual(recurringJob.Spec.Parameters, spec.Parameters) {
```
derekbit, line 73: nit: why not directly compare the entire spec?

ChanYiLin: fixed.

PR #2948, opened 2024-07-04
chore(k8s): update controller-gen and kustomize

derekbit:

Issue longhorn/longhorn#8901

- update controller-gen to `v0.15.0`
- update kustomize to `kustomize/v5.4.2`

### DIFF
--- a/k8s/crds.yaml
+++ b/k8s/crds.yaml
@@ -3,8 +3,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.7.0
-  creationTimestamp: null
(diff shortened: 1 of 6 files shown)


shuo-wu (commented):
Why do we need to change the webhook port? At least in this PR.

BTW, the commit said that `update webhook port to 8501` while the ports are actually changed to 9501

derekbit:
> Why do we need to change the webhook port? At least in this PR.
> 
> BTW, the commit said that `update webhook port to 8501` while the ports are actually changed to 9501

webhook is already changed to 8501 since v1.6.0.
We only updated longhorn/longhorn but forgot updating longhorn/longhorn-manager.
cc @ChanYiLin

innobead (approved):
LGTM

PR #2956, opened 2024-07-08
fix(backing image clone): use ignore as default value when encryption is not set

ChanYiLin:
in the backing iamge webhook mutator
use ignore as default value when encryption is not set

### DIFF
--- a/webhook/resources/backingimage/mutator.go
+++ b/webhook/resources/backingimage/mutator.go
@@ -94,7 +94,12 @@ func (b *backingImageMutator) Create(request *admission.Request, newObj runtime.
 	}
 
 	if longhorn.BackingImageDataSourceType(backingImage.Spec.SourceType) == longhorn.BackingImageDataSourceTypeClone {
-		// inherit the secret and secretNamespace when cloning from another backing image and the encryption is ignore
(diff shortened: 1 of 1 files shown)


webhook/resources/backingimage/mutator.go
```diff
@@ -94,7 +94,12 @@ func (b *backingImageMutator) Create(request *admission.Request, newObj runtime.
 	}
 
 	if longhorn.BackingImageDataSourceType(backingImage.Spec.SourceType) == longhorn.BackingImageDataSourceTypeClone {
-		// inherit the secret and secretNamespace when cloning from another backing image and the encryption is ignore
+		// Use ignore as default value when encryption is not set
+		if parameters[longhorn.DataSourceTypeCloneParameterEncryption] == "" {
+			parameters[longhorn.DataSourceTypeCloneParameterEncryption] = "ignore"
```
derekbit, line 99:
```suggestion
			parameters[longhorn.DataSourceTypeCloneParameterEncryption] = string(bimtypes.EncryptionTypeIgnore)
```

ChanYiLin: fixed. thanks

derekbit (approved):
LGTM

PR #2957, opened 2024-07-08
Fix bug engine.status.clonestatus is unnecessarily repeatedly updated

PhanLe1010:
longhorn/longhorn#8952

The monitor update engine.Status.CloneStatus to a non-nil value and the PR https://github.com/longhorn/longhorn-manager/pull/2895 set it back to nil. They are fighting each other leading to unnecessarily repeatedly engine CR updates


ejweber (approved):
LGTM.

PR #2958, opened 2024-07-09
fix(websocket): add backup backing image websocket to websocket controller

ChanYiLin:
Need to add the informer to websocket controller so the websocket can work

- Have tested it in local env
- Have tested it with UI

cc @a110605


derekbit (approved):
LGTM

PR #2960, opened 2024-07-09
Fix cloning stuck due to having empty entries in e.Status.CloneStatus

PhanLe1010:
longhorn/longhorn#3692

Don't block cloning if `e.Status.CloneStatus` has empty entries


PhanLe1010:
CI tests https://ci.longhorn.io/job/private/job/longhorn-tests-regression/7198/

ejweber (approved):
Discussed with @PhanLe1010 offline.

LGTM.

PR #2961, opened 2024-07-10
fix(datastore): CheckInstanceManagersReadiness should find running instance managers for v2 data engine

derekbit:

Issue Longhorn/longhorn#8967

There is one running instance manager for v2 data engine on each node, so the function find one running instance manager instead.

### DIFF
--- a/datastore/longhorn.go
+++ b/datastore/longhorn.go
@@ -3400,8 +3400,10 @@ func CheckInstanceManagerType(im *longhorn.InstanceManager) (longhorn.InstanceMa
 	return longhorn.InstanceManagerType(""), fmt.Errorf("unknown type %v for instance manager %v", imType, im.Name)
 }
 
-// CheckInstanceManagerReadiness checks if the InstanceManager is running on the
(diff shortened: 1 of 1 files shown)


datastore/longhorn.go
```diff
@@ -3427,7 +3435,7 @@ func (s *DataStore) CheckInstanceManagersReadiness(dataEngine longhorn.DataEngin
 					"instanceManager": instanceManager.Name,
 					"node":            node,
 				},
-			).Errorf("CheckInstanceManagerReadiness: instance manager is not running")
+			).Errorf("CheckInstanceManagersReadiness: instance manager is not running")
```
c3y1huang, line 3438:
```suggestion
			).Error("CheckInstanceManagersReadiness: instance manager is not running")
```

c3y1huang (changes requested):

datastore/longhorn.go
derekbit: Fixed. Thank you.

c3y1huang (approved):

innobead (approved):

PR #2962, opened 2024-07-10
fix(pre-upgrade): don't call environmentCheck without necessary bind mounts

ejweber:

https://github.com/longhorn/longhorn/issues/8974

I think this is the right fix, but I may be missing some context. Feel free to open a different PR if the fix should be different @mantissahz.


derekbit:
The issue https://github.com/longhorn/longhorn/issues/8974 in is fixed by https://github.com/longhorn/longhorn/issues/8974#issuecomment-2221727128. Let's close this one.

PR #2963, opened 2024-07-11
fix: fix recurring job recreate issue

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/8874

fix recurring job recreating even the job already exists issue
`reflect.DeepEqual` considers `nil` and `map[]` are different, need to init the map struct before comparing


ChanYiLin:
### Wait for testing
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/7207/

derekbit (approved):
LGTM

PR #2972, opened 2024-07-16
Fix(RWX):  long remount time when share manager pod restart

c3y1huang:

Issue longhorn/longhorn#8999

This PR addressed a regression where remounting NFS volume hangs for an extended period before succeeding. The regression is a side effect from  https://github.com/longhorn/longhorn-manager/pull/2815, and the way the share manager's Service is cleaned up caused the remounting to hang after the share manager Pod restarts.

This PR proposes moving the responsibility of cleaning up the Service and Endpoint to the setting controller. This is a more logical fit since the setting controller handles the storage network change, which aligns better with the purpose of this cleanup process. This change should eliminate the race condition.

`None`

- https://github.com/longhorn/longhorn/issues/8999#issuecomment-2228183416

### DIFF
--- a/controller/setting_controller.go
+++ b/controller/setting_controller.go
@@ -344,6 +344,14 @@ func (sc *SettingController) syncDangerZoneSettingsForManagedComponents(settingN
 			if err := sc.updateCNI(sc.ds.AreAllRWXVolumesDetached); err != nil {
 				return err
 			}
+
+			// Perform cleanup of the share manager Service
+			// This is to allow the creation of the correct Service
+			// and Endpoint when switching between cluster network
+			// and storage network.
+			if err := sc.cleanupShareManagerServiceAndEndpoints(); err != nil {
+				return err
+			}
 		}
 		return nil
 	}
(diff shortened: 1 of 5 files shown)


c3y1huang:
The failure from `CodeFactor` is non-related to the changes proposed in this PR.

c3y1huang:
Test WIP.

derekbit (dismissed):
LGTM

controller/setting_controller.go
```diff
@@ -914,6 +922,38 @@ func (sc *SettingController) updateKubernetesClusterAutoscalerEnabled() error {
...
+		}
+	}()
+
+	shareManagers, err := sc.ds.ListShareManagers()
+	if err != nil {
+		return err
+	}
+
+	for _, shareManager := range shareManagers {
+		_log := sc.logger.WithField("shareManager", shareManager.Name)
```
derekbit, line 939: Why not just use `log`?

controller/setting_controller.go
```diff
@@ -914,6 +922,38 @@ func (sc *SettingController) updateKubernetesClusterAutoscalerEnabled() error {
...
+
+		_log.WithField("service", shareManager.Name).Infof("Deleting Service for %v setting update", types.SettingNameStorageNetworkForRWXVolumeEnabled)
+		err := sc.ds.DeleteService(shareManager.Namespace, shareManager.Name)
+		if err != nil && !apierrors.IsNotFound(err) {
+			return err
+		}
+
+		_log.WithField("endpoint", shareManager.Name).Infof("Deleting Endpoint for %v setting update", types.SettingNameStorageNetworkForRWXVolumeEnabled)
+		err = sc.ds.DeleteKubernetesEndpoint(shareManager.Namespace, shareManager.Name)
```
ejweber, line 948: Was it an oversight from before that we did not delete the Endpoints? Or does this PR somehow require us to do so now when the last one did not?

ejweber:
The regression is a side effect from https://github.com/longhorn/longhorn-manager/pull/2815, and the way the share manager's Service is cleaned up caused the remounting to hang after the share manager Pod restarts.

Is it because we reference the service using a cluster IP, and now the cluster IP has changed?

ejweber (approved):
I was confused why we never had to delete the share manager service before the changes from https://github.com/longhorn/longhorn-manager/pull/2815. I see now that it was because the service was owned by the share manager and automatically deleted when the share manager was deleted. https://github.com/longhorn/longhorn-manager/pull/2815 caused it to be deleted every time the share manager POD was deleted. Now we only delete it when changing the storage network setting, which seems to make sense.

LGTM, though I think we should address @derekbit's `_log` comment before merging.

controller/setting_controller.go
c3y1huang: The Endpoint is owned by the ShareManager, so it gets cleaned up when the ShareManager is removed. Because it gets cleaned up during my test, so I overlooked that we should also clean up the Endpoint when the Service is gone. It's a minor, non-blocking bug.

controller/setting_controller.go
c3y1huang: It is to improve readability indicating that this is used within the loop. However, I am also fine using `log`.

Updated.

derekbit (approved):

c3y1huang:
@mergify backport v1.7.x

PR #2977, opened 2024-07-18
fix(rwx): workload pod hangs after node reboot

c3y1huang:

Issue longhorn/longhorn#9014, longhorn/longhorn#9022

Fix AreAllRWXVolumesDetached to check if there are no more Longhorn VolumeAttachment for the RWX volumes.

`None`

- https://github.com/longhorn/longhorn/issues/9022#issuecomment-2234423678

### DIFF
--- a/controller/setting_controller.go
+++ b/controller/setting_controller.go
@@ -323,11 +323,20 @@ func (sc *SettingController) syncDangerZoneSettingsForManagedComponents(settingN
 				return err
 			}
 		case types.SettingNameStorageNetwork:
-			fnAreAllVolumeDetached := func() (bool, error) {
-				isDetached, _, err := sc.ds.AreAllVolumesDetached(longhorn.DataEngineTypeAll)
-				return isDetached, err
+			funcPreupdate := func() error {
+				detached, _, err := sc.ds.AreAllVolumesDetached(longhorn.DataEngineTypeAll)
+				if err != nil {
+					return errors.Wrapf(err, "failed to check volume detachment for %v setting update", types.SettingNameStorageNetwork)
+				}
+
+				if !detached {
+					return &types.ErrorInvalidState{Reason: fmt.Sprintf("failed to apply %v setting to Longhorn components when there are attached volumes. It will be eventually applied", types.SettingNameStorageNetwork)}
+				}
+
+				return nil
 			}
-			if err := sc.updateCNI(fnAreAllVolumeDetached); err != nil {
+
+			if err := sc.updateCNI(funcPreupdate); err != nil {
 				return err
 			}
 		}
@@ -341,15 +350,27 @@ func (sc *SettingController) syncDangerZoneSettingsForManagedComponents(settingN
 	if slices.Contains(dangerSettingRequiringRWXVolumesDetached, settingName) {
 		switch settingName {
 		case types.SettingNameStorageNetworkForRWXVolumeEnabled:
-			if err := sc.updateCNI(sc.ds.AreAllRWXVolumesDetached); err != nil {
-				return err
+			funcPreupdate := func() error {
+				detached, err := sc.ds.AreAllRWXVolumesDetached()
+				if err != nil {
+					return errors.Wrapf(err, "failed to check volume detachment for %v setting update", types.SettingNameStorageNetworkForRWXVolumeEnabled)
+				}
+				if !detached {
+					return &types.ErrorInvalidState{Reason: fmt.Sprintf("failed to apply %v setting to Longhorn components when there are attached volumes. It will be eventually applied", types.SettingNameStorageNetworkForRWXVolumeEnabled)}
+				}
+
+				// Perform cleanup of the share manager Service
+				// This is to allow the creation of the correct Service
+				// and Endpoint when switching between cluster network
+				// and storage network.
+				if err := sc.cleanupShareManagerServiceAndEndpoints(); err != nil {
+					return err
+				}
+
+				return nil
 			}
 
-			// Perform cleanup of the share manager Service
-			// This is to allow the creation of the correct Service
-			// and Endpoint when switching between cluster network
-			// and storage network.
-			if err := sc.cleanupShareManagerServiceAndEndpoints(); err != nil {
+			if err := sc.updateCNI(funcPreupdate); err != nil {
 				return err
 			}
 		}
@@ -955,7 +976,7 @@ func (sc *SettingController) cleanupShareManagerServiceAndEndpoints() error {
 }
 
 // updateCNI deletes all system-managed data plane components immediately with the updated CNI annotation.
-func (sc *SettingController) updateCNI(fnCheckVolumeDetached func() (bool, error)) error {
+func (sc *SettingController) updateCNI(funcPreupdate func() error) error {
 	storageNetwork, err := sc.ds.GetSettingWithAutoFillingRO(types.SettingNameStorageNetwork)
 	if err != nil {
 		return err
@@ -985,12 +1006,9 @@ func (sc *SettingController) updateCNI(fnCheckVolumeDetached func() (bool, error
 		return nil
 	}
 
-	detached, err := fnCheckVolumeDetached()
+	err = funcPreupdate()
 	if err != nil {
-		return errors.Wrapf(err, "failed to check volume detachment for %v setting update", types.SettingNameStorageNetwork)
-	}
-	if !detached {
-		return &types.ErrorInvalidState{Reason: fmt.Sprintf("failed to apply %v setting to Longhorn components when there are attached volumes. It will be eventually applied", types.SettingNameStorageNetwork)}
+		return err
 	}
 
 	for _, daemonSet := range incorrectCNIDaemonSets {
--- a/datastore/longhorn.go
+++ b/datastore/longhorn.go
@@ -490,11 +490,26 @@ func (s *DataStore) ValidateV2DataEngineEnabled(dataEngineEnabled bool) (ims []*
 }
 
 func (s *DataStore) AreAllRWXVolumesDetached() (bool, error) {
-	pods, err := s.ListShareManagerPodsRO("")
+	volumes, err := s.ListVolumesRO()
 	if err != nil {
 		return false, err
 	}
-	return len(pods) == 0, nil
+	for _, volume := range volumes {
+		if volume.Spec.AccessMode != longhorn.AccessModeReadWriteMany {
+			continue
+		}
+
+		volumeAttachment, err := s.GetLHVolumeAttachmentByVolumeName(volume.Name)
+		if err != nil && !apierrors.IsNotFound(err) {
+			return false, err
+		}
+
+		if volumeAttachment != nil && len(volumeAttachment.Spec.AttachmentTickets) > 0 {
+			return false, nil
+		}
+	}
+
+	return true, nil
 }
 
 func (s *DataStore) AreAllVolumesDetached(dataEngine longhorn.DataEngineType) (bool, []*longhorn.InstanceManager, error) {


datastore/longhorn.go
```diff
@@ -490,11 +490,26 @@ func (s *DataStore) ValidateV2DataEngineEnabled(dataEngineEnabled bool) (ims []*
...
+		}
+
+		volumeAttachment, err := s.GetLHVolumeAttachmentByVolumeName(volume.Name)
+		if err != nil && !apierrors.IsNotFound(err) {
+			return false, err
+		}
+
+		if volumeAttachment != nil {
+			return false, nil
+		}
```
PhanLe1010, line 509: We always has VA object for every volume. Perhaps you was intended to check if the VA.Spec.AttachmentTickets empty or not?

PhanLe1010 (commented):
Can we also use the `setting.Status.Applied` field to check if the setting has changed? Then we only attempt the service deletion logic if we the setting has changed?

Ref:
```
kind: Setting
metadata:
  annotations:
    longhorn.io/configmap-resource-version: "4374"
  creationTimestamp: "2024-07-17T00:40:23Z"
  generation: 1
  name: storage-network-for-rwx-volume-enabled
  namespace: longhorn-system
  resourceVersion: "8963"
  uid: 61da6942-5d8e-4817-b671-9cae0ed48700
status:
  applied: true
value: "false"
```

datastore/longhorn.go
```diff
@@ -490,11 +490,26 @@ func (s *DataStore) ValidateV2DataEngineEnabled(dataEngineEnabled bool) (ims []*
...
+
+		volumeAttachment, err := s.GetLHVolumeAttachmentByVolumeName(volume.Name)
+		if err != nil && !apierrors.IsNotFound(err) {
+			return false, err
+		}
+
+		if volumeAttachment != nil &&volumeAttachment.Spec.AttachmentTickets != nil {
+			return false, nil
+		}
```
PhanLe1010, line 509:
```suggestion
		if volumeAttachment != nil && len(volumeAttachment.Spec.AttachmentTickets) > 0 {
			return false, nil
		}
```

PhanLe1010 (commented):
I am not fully understand the full logic for setting.Status.Applied. Do you have full understand? Like user switch the setting value, where do we clear the setting.Status.Applied ?

mantissahz:
>I am not fully understand the full logic for setting.Status.Applied. Do you have full understand? Like user switch the setting value, where do we clear the setting.Status.Applied ?

`setting.Status.Applied` will be true only when the setting controller walks through syncSetting without errors. (Modifying Danger Zone settings with volumes attaching will return an error.)

https://github.com/longhorn/longhorn-manager/blob/b615732bc75d5513a7f3289a667af1f06a55af1d/controller/setting_controller.go#L238-L248
Therefore `setting.Status.Applied` will always be `False` if there are volumes attached for danger zone settings that have been modified.
And users need to reapply the changes to Danger Zone settings modified (or wait an hour for the default reconciling period of the setting controller) after all volumes are detached to reconcile the setting status.

PhanLe1010:
I am thinking that the current logic looks a bit off to me:
1. Initially `setting.Status.Applied` is false
1. It is updated to `true` after the first sync
1. When it is synced again, if there is an ETCD error, it is updated back to false (eventhough we already applied the setting?)

I am thinking about this design:

1. When someone update the setting CR, webhook update the `setting.Status.Applied` to false because setting has new value
1. Setting controller try to take some action (in your case it is deleting the RWX service) only if the `setting.Status.Applied`  is `false`
1. Setting controller set `setting.Status.Applied` to `true` if that action was successful
1. Setting controller nerver set `setting.Status.Applied` to `false`. It is not the responsibiliity of setting controller to set it to `false`

WDYT @mantissahz @c3y1huang ?

c3y1huang:
https://github.com/longhorn/longhorn-manager/blob/v1.7.0-rc1/controller/setting_controller.go#L239-L240

It will be updated to true when the sync returned error is nil.
So for the storage-network settings, after the CNI is updated, the setting should be applied (first reconcile). and the setting will be queued again due to the change and it will run through the same logic, and then the Service and Endpoint should be cleaned up in the second reconcile queue.

> I am thinking that the current logic looks a bit off to me:
> 
> 1. Initially `setting.Status.Applied` is false
> 2. It is updated to `true` after the first sync
> 3. When it is synced again, if there is an ETCD error, it is updated back to false (eventhough we already applied the setting?)
> 
> I am thinking about this design:
> 
> 1. When someone update the setting CR, webhook update the `setting.Status.Applied` to false because setting has new value
> 2. Setting controller try to take some action (in your case it is deleting the RWX service) only if the `setting.Status.Applied`  is `false`
> 3. Setting controller set `setting.Status.Applied` to `true` if that action was successful
> 4. Setting controller nerver set `setting.Status.Applied` to `false`. It is not the responsibiliity of setting controller to set it to `false`
> 
> WDYT @mantissahz @c3y1huang ?

How will this resolve the case when the CNI is already updated but the Service and Endpoint cleanup failed case?
1. `setting.Status.Applied` is false.
1. CNI updated successful.
1. Service and Endpoint cleaning up failed.
1. `setting.Status.Applied` remains false and it doesn't get queued again during the default reconciling period.

For the PR I think we can requeue when the cleanup failed instead of returning error. This way the `setting.Status.Applied` won't get updated to false.

What do you think @PhanLe1010 @mantissahz ?

mantissahz:
>I am thinking about this design:
>1. When someone update the setting CR, webhook update the `setting.Status.Applied` to `false` >because setting has new value
>2. Setting controller try to take some action (in your case it is deleting the RWX service) only if the `setting.Status.Applied` is `false`
>3. Setting controller set `setting.Status.Applied` to true if that action was successful
>4. Setting controller nerver set `setting.Status.Applied` to `false`. It is not the responsibiliity of setting controller to set it to `false`

It is more clear and simple than the current implementation and I think we can have this improvement but does it still have the same issue you mentioned (When it is synced again, if there is an ETCD error, it is updated back to false (eventhough we already applied the setting?))?

> move the cleanup logic before CNI update so it gets handled first. (preferred)
> - Check are all RWX volumes are detached. If not, return an error. (Applied = false)
> - Cleanup Service and Endpoint. If failed, return an error (Applied = false)
> - Update CNI. If failed, return the error (Applied = false)

I prefer this one because `Applied` should be `false` if the cleanup of Service and Endpoint is a part of reconciling the `SettingNameStorageNetworkForRWXVolumeEnabled` setting and it failed.

c3y1huang:
Test WIP.

controller/setting_controller.go
```diff
@@ -323,11 +323,16 @@ func (sc *SettingController) syncDangerZoneSettingsForManagedComponents(settingN
...
-			fnAreAllVolumeDetached := func() (bool, error) {
-				isDetached, _, err := sc.ds.AreAllVolumesDetached(longhorn.DataEngineTypeAll)
-				return isDetached, err
+			detached, _, err := sc.ds.AreAllVolumesDetached(longhorn.DataEngineTypeAll)
+			if err != nil {
+				return errors.Wrapf(err, "failed to check volume detachment for %v setting update", types.SettingNameStorageNetwork)
+			}
+
+			if !detached {
+				return &types.ErrorInvalidState{Reason: fmt.Sprintf("failed to apply %v setting to Longhorn components when there are attached volumes. It will be eventually applied", types.SettingNameStorageNetwork)}
```
ejweber, line 332: NIT: It looks like we will log this error every time the setting rereconciles, even if no changes are required. The previous implementation (and the other similar ones) checked whether there was anything in need of updating BEFORE checking if volumes were detached.

datastore/longhorn.go
```diff
@@ -490,11 +490,26 @@ func (s *DataStore) ValidateV2DataEngineEnabled(dataEngineEnabled bool) (ims []*
...
 	if err != nil {
 		return false, err
 	}
-	return len(pods) == 0, nil
+	for _, volume := range volumes {
+		if volume.Spec.AccessMode != longhorn.AccessModeReadWriteMany {
+			continue
+		}
+
+		volumeAttachment, err := s.GetLHVolumeAttachmentByVolumeName(volume.Name)
```
ejweber, line 502: As far as I can tell, this change will not help with https://github.com/longhorn/longhorn/issues/9022. The issue @PhanLe1010 and @james-munson are having (if I understand it properly) is that NFS clients still exist on the cluster trying to use the old ClusterIP. The `volumeattachment.longhorn.io` does not know anything about them. It is only concerned with the underlying Longhorn volume. In order to ensure all NFS clients are gone, I think we must look at `volumeattachment.storage.k8s.io` objects. Alternatively, do we trust the `volume.status.KubernetesStatus`? Reading through the code, I think it is the case that all RWX clients are gone if `lastPodRefAt != ""`. It indicates that we can no longer find any pods on the cluster that reference the volume. So either they have been successfully torn down or force deleted (in which case, all bets are off).

https://github.com/longhorn/longhorn-manager/blob/b615732bc75d5513a7f3289a667af1f06a55af1d/controller/kubernetes_pv_controller.go#L460-L467

ejweber (changes requested):
~It does not look to me like this set of changes addresses https://github.com/longhorn/longhorn/issues/9022. See my comments for details.~

datastore/longhorn.go
ejweber: Sorry, we can disregard this. I was confused, but @PhanLe1010 set me straight offline.

We actually DO create a `volumeattachment.longhorn.io` for every node attempting to use an RWX volume. In ADDITION, we have the one responsible for attaching the underlying volume to the share manager's node.

https://github.com/longhorn/longhorn-manager/blob/b615732bc75d5513a7f3289a667af1f06a55af1d/controller/volume_attachment_controller.go#L579-L583

If there are no attachment tickets, it means Kubernetes has issued ControllerUnpublishVolume RPCs for every node that had the volume RWX volume attached. It should only do this if NodeUnpublishVolume and NodeUnstageVolume RPCs have succeeded. These RPCs should clean up all active NFS clients.

PhanLe1010:
> How will this resolve the case when the CNI is already updated but the Service and Endpoint cleanup failed case?
> 1. `setting.Status.Applied` is false.
> 1. CNI updated successful.
> 1. Service and Endpoint cleaning up failed.
> 1. `setting.Status.Applied` remains false and it doesn't get queued again during the default reconciling period.

@c3y1huang when there is an error in the setting sync loop, we will re-queue setting CR automatically here https://github.com/longhorn/longhorn-manager/blob/b615732bc75d5513a7f3289a667af1f06a55af1d/controller/setting_controller.go#L210-L215 

> It is more clear and simple than the current implementation and I think we can have this improvement but does it still have the same issue you mentioned (When it is synced again, if there is an ETCD error, it is updated back to false (eventhough we already applied the setting?))?

@mantissahz In my proposal, setting controller should never update the `setting.Status.Applied` to `false`

However, my proposal at https://github.com/longhorn/longhorn-manager/pull/2977#issuecomment-2235117001 is violating the Kubernetes controller design pattern (the CR.Status should reflect the actual state of the world) because I propose that setting controller never update `setting.Status.Applied` to `false`. This was pointed out by @ejweber in the US syncup. So we are thinking of a new design like:
1. Create an interface function like: `func isSettingTakeEffect(settingName string, settingValue string) bool`. This function will check the actual state of the world (for example for the toleration setting, it check to see if all Longhorn components has correct toleration). Return true if the actual state of the world agree with the setting value 
2. Keep `setting.Status.Applied` meaning to reflect the current state of the world. I.E, set it to the value of `isSettingTakeEffect(settingName, settingValue)` 
3. Only try to apply the setting if `isSettingTakeEffect(settingName, settingValue)` is `false`

Also, I agree with @c3y1huang and @mantissahz that this isApplied logic should be handled in a different ticket. WDYT @ejweber @c3y1huang @mantissahz ?

For now the current PR LGTM

mantissahz:
>Also, I agree with @c3y1huang and @mantissahz that this `isApplied` logic should be handled in a different ticket. WDYT @ejweber @c3y1huang @mantissahz ?

Yes, I agree that we should discuss `isApplied` logic should be handled in a different ticket.

controller/setting_controller.go
c3y1huang: Good catch. Thank you.

derekbit:
> > Also, I agree with @c3y1huang and @mantissahz that this `isApplied` logic should be handled in a different ticket. WDYT @ejweber @c3y1huang @mantissahz ?
> 
> Yes, I agree that we should discuss `isApplied` logic in a different ticket.

@mantissahz Could you create a ticket for the improvement? Thank you.

derekbit (approved):

controller/setting_controller.go
```diff
@@ -323,11 +323,20 @@ func (sc *SettingController) syncDangerZoneSettingsForManagedComponents(settingN
...
-				isDetached, _, err := sc.ds.AreAllVolumesDetached(longhorn.DataEngineTypeAll)
-				return isDetached, err
+			funcPreupdate := func() error {
+				detached, _, err := sc.ds.AreAllVolumesDetached(longhorn.DataEngineTypeAll)
+				if err != nil {
+					return errors.Wrapf(err, "failed to check volume detachment for %v setting update", types.SettingNameStorageNetwork)
+				}
+
+				if !detached {
+					return &types.ErrorInvalidState{Reason: fmt.Sprintf("failed to apply %v setting to Longhorn components when there are attached volumes. It will be eventually applied", types.SettingNameStorageNetwork)}
```
derekbit, line 333: nit: We can generalize a package for errors with different type such InvalidState, NotFound and so on.

c3y1huang:
@merfigy backport v1.7.x

PR #2978, opened 2024-07-18
feat(csi metrics): enable csi sidecar metrics

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/7938

enable csi sidecar metrics on port 8000
- csi-attacher
- csi-provisioner
- csi-resizer
- csi-snapshotter

cc @PhanLe1010


derekbit (approved):
LGTM

PR #2982, opened 2024-07-19
Fix(trim): extend the timeout

c3y1huang:

Issue longhorn/longhorn#6868

Propose extending the execution timeout for the `fstrim` command to 1 hour for v1.7.0. In the future release, we can work on enhancements such as running sync or making the timeout configurable.

I've not been able to reproduce the timeout issue as mentioned in the issue. In my environment, the `fstrim` completes quickly, However, my test data size is relatively small (100G). cc @longhorn/qa 

`None`

### DIFF
--- a/util/util.go
+++ b/util/util.go
@@ -749,7 +749,7 @@ func TrimFilesystem(volumeName string, encryptedDevice bool) error {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


derekbit (approved):

derekbit:

PR #2996, opened 2024-07-24
fix(disk): non-existing block device results in longhorn-manager to be in Crashloopbackoff state

derekbit:

Issue Longhorn/longhorn#9073

Non-existing block device results in longhorn-manager pods to be in Crashloopbackoff state. When longhorn-manager pods are in Crashloopbackoff, users are unable to update the default-data-path setting.

### DIFF
--- a/datastore/longhorn.go
+++ b/datastore/longhorn.go
@@ -2584,16 +2584,7 @@ func (s *DataStore) needDefaultDiskCreation(dataPath string) bool {
 	}
 
 	// Do not create default block-type disk if v2 data engine is disabled
-	ok, err := types.IsBlockDisk(dataPath)
-	if err != nil {
-		logrus.WithError(err).Errorf("Failed to check if the data path %v is block-type", dataPath)
-		return false
-	}
-	if ok {
-		return false
-	}
-
-	return true
+	return !types.IsPotentialBlockDisk(dataPath)
 }
 
 func (s *DataStore) GetNodeRO(name string) (*longhorn.Node, error) {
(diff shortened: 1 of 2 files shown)


shuo-wu:
The solution in the PR should work. But the users may confused why there is even no disk after they wrongly change the default disk setting.
I prefer to blindly create default disks, then let the node controller detect if it's a valid one

innobead (approved):

innobead:
Can we determine whether the default path is for  v1 or v2 now? The current implementation is more like a fallback. If it's just fallback, probably @shuo-wu suggestion is good enough. However, still need a log/event to notice users.

shuo-wu:
Discussed it with Derek. 
Now `CreateDefaultDisk` relies on `IsBlockDisk` to guess the type of the default disk path and this function should not error out. Hence `IsBlockDisk` just tries its best, no need to have an accurate judgment. We will leave the concrete validations and handling to the node controller and users themselves.

derekbit:
> Discussed it with Derek. Now `CreateDefaultDisk` relies on `IsBlockDisk` to guess the type of the default disk path and this function should not error out. Hence `IsBlockDisk` just tries its best, no need to have an accurate judgment. We will leave the concrete validations and handling to the node controller and users themselves.

Updated the PR. Ready for review @shuo-wu @innobead

shuo-wu (approved):

innobead (approved):

derekbit:
@mergify backport v1.7.x

derekbit:
@mergify backport v1.6.x

PR #3000, opened 2024-07-24
chore(backup): clarify deletion log message

ejweber:

No particular linked issue. I saw the following logged when I configured a backup target in my cluster and I was quite confused.

```
[longhorn-manager-qwx8h longhorn-manager] time="2024-07-24T14:47:28Z" level=info msg="Found 1 backup volumes in the backup target that do not exist in the backup target and need to be deleted" func="controller.
```

The repeated mention of "backup target" doesn't make sense, and I was worried that backups were being deleted from the target itself, which I would not expect. (That is not the case, so the log can be clarified.)

### DIFF
--- a/controller/backup_target_controller.go
+++ b/controller/backup_target_controller.go
@@ -496,7 +496,7 @@ func (btc *BackupTargetController) syncBackupVolume(backupStoreBackupVolumeNames
 	// and delete the BackupVolume CR in the cluster
 	backupVolumesToDelete := clusterBackupVolumesSet.Difference(backupStoreBackupVolumes)
 	if count := backupVolumesToDelete.Len(); count > 0 {
-		log.Infof("Found %d backup volumes in the backup target that do not exist in the backup target and need to be deleted", count)
+		log.Infof("Found %d backup volumes in the backup target that do not exist in the cluster and need to be deleted from the cluster", count)
 	}
 	for backupVolumeName := range backupVolumesToDelete {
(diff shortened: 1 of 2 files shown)


ejweber:
@mergify backport v1.7.x v1.6.x v1.5.x

controller/backup_target_controller.go
```diff
@@ -496,7 +496,7 @@ func (btc *BackupTargetController) syncBackupVolume(backupStoreBackupVolumeNames
 	// and delete the BackupVolume CR in the cluster
 	backupVolumesToDelete := clusterBackupVolumesSet.Difference(backupStoreBackupVolumes)
 	if count := backupVolumesToDelete.Len(); count > 0 {
-		log.Infof("Found %d backup volumes in the backup target that do not exist in the backup target and need to be deleted", count)
+		log.Infof("Found %d backup volumes in the backup target that do not exist in the cluster and need to be deleted from the cluster", count)
```
james-munson, line 499: Still not entirely non-confusing.  I might leave it at "need to be deleted."  I suppose it is really from the cluster's stored resources.  Anyway, better than it was.

james-munson (approved):
I'll buy that.

controller/backup_target_controller.go
ejweber: Agree that it's still a bit confusing, but I think it's really important to distinguish here that we are NOT deleting it from the backup target. That was the thing that scared me when I first saw it. It seemed to me that I hooked up a new cluster to an existing backup store and it started deleting my backups!

PR #3004, opened 2024-07-24
If volume is delinquent, switch owner of volume/engine/replica to share manager CR's owner

PhanLe1010:
Without the fix, volume/engine/replica continue to wait for the share manager pod to be scheduled (i.e., pod.Spec.NodeName is non empty) to set ownerID to the same pod's node. However, because we don't want to use pod's imformers, when the share manager pod is scheduled, volume/engine/controller might not catch that event and continue to wait. This introduce up to 30s delay and behavioral inconsistency

Also, the > 30s delay in share manager pod recreation is destroying the RWX fast failover's original goal

longhorn/longhorn#6205

Some testing results:
* Before the fix, it was taking from 15s to 70s for the new share manager pod to become running after shutting down the node of the old share manager pod
* After the fix, it is taking from 15s to 17s for the new share manager pod to become running after shutting down the node of the old share manager pod

### DIFF
--- a/controller/engine_controller.go
+++ b/controller/engine_controller.go
@@ -2296,22 +2296,27 @@ func (ec *EngineController) isResponsibleFor(e *longhorn.Engine, defaultEngineIm
 		err = errors.Wrap(err, "error while checking isResponsibleFor")
 	}()
 
-	// If a regular RWX is delinquent, try to switch ownership quickly to the node of the newly created share-manager pod
-	isDelinquent, err := ec.ds.IsNodeDelinquent(e.Status.OwnerID, e.Spec.VolumeName)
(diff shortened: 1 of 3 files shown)


ejweber (approved):
Discussed this with @PhanLe1010 and @james-munson offline. I think the reason for the fix makes sense and the implementation looks good.

I had a bit of trouble understanding exactly why the fix was effective, but @PhanLe1010 helped me understand that:

- There are enough triggering events with the existing informers to ensure fast failover, but
- If we wait to transfer ownership until the new share manager pod is scheduled, we may not have an active owner when the trigger events occur
- Then, we have to wait the resync period.

james-munson (approved):

PhanLe1010:
@mergify backport v1.7.x

PR #3008, opened 2024-07-25
Maintain the previous state machine of share manager CR

PhanLe1010:
1. Switch the share manager controller sync loop to sync volume first then sync share manager pod after
2. Skip calling gprc to mount the share manager volume if the share manager is currently delinquent. This helps to avoid the 20s delay when attemping to make the grpc call to dead share-manager pod

This PR helps:
1. Fulfill the requirement to the investigation at https://github.com/longhorn/longhorn-manager/pull/2811#issuecomment-2244142848
2. Resolve the fail test https://github.com/longhorn/longhorn/issues/9081

longhorn/longhorn#6205

### DIFF
--- a/controller/share_manager_controller.go
+++ b/controller/share_manager_controller.go
@@ -407,11 +407,11 @@ func (c *ShareManagerController) syncShareManager(key string) (err error) {
 		}
 	}()
 
-	if err = c.syncShareManagerPod(sm); err != nil {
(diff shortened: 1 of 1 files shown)


ejweber (approved):
LGTM. Thanks for the fix!

james-munson (approved):
As discussed.

PhanLe1010:
@mergify backport v1.7.x

PR #3013, opened 2024-07-26
If the volume is faulted, we don't need to have RWX fast failover.

PhanLe1010:
If the volume is faulted, we don't need to have RWX fast failover.
If shareManager is delinquent, clear both delinquent and stale state.
If we don't do that, the volume will get stuck in the auto-savage loop.

See https://github.com/longhorn/longhorn/issues/9089

New statemachine

lease_cr_statemachine

### DIFF
--- a/datastore/kubernetes.go
+++ b/datastore/kubernetes.go
@@ -282,6 +282,39 @@ func (s *DataStore) UpdateLease(lease *coordinationv1.Lease) (*coordinationv1.Le
 	return s.kubeClient.CoordinationV1().Leases(s.namespace).Update(context.TODO(), lease, metav1.UpdateOptions{})
 }
 
+func (s *DataStore) ClearDelinquentAndStaleStateIfVolumeIsDelinquent(volumeName string) (err error) {
+	defer func() {
+		err = errors.Wrapf(err, "failed to ClearDelinquentAndStaleStateIfVolumeIsDelinquent")
+	}()
+
+	lease, err := s.GetLeaseRO(volumeName)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil
+		}
+		return err
+	}
+
+	holder := *lease.Spec.HolderIdentity
+	if holder == "" {
+		// Empty holder means not delinquent.
+		// Ref: IsRWXVolumeDelinquent() function
+		return nil
+	}
+	if !(lease.Spec.AcquireTime).IsZero() {
+		// Non Zero lease.Spec.AcquireTime means not delinquent.
+		// Ref: IsRWXVolumeDelinquent() function
+		return nil
+	}
+
+	oneSecondAfterZero := time.Time{}.Add(time.Second)
+	lease.Spec.AcquireTime = &metav1.MicroTime{Time: oneSecondAfterZero}
+	lease.Spec.RenewTime = &metav1.MicroTime{Time: time.Time{}}
+
+	_, err = s.UpdateLease(lease)
+	return err
+}
+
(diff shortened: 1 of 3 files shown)


PhanLe1010:
The PR is ready to take review/feedback but not ready to merge yet. I will need to do more testing

derekbit (approved):
Make sense. LGTM

datastore/kubernetes.go
```diff
@@ -282,6 +282,38 @@ func (s *DataStore) UpdateLease(lease *coordinationv1.Lease) (*coordinationv1.Le
 	return s.kubeClient.CoordinationV1().Leases(s.namespace).Update(context.TODO(), lease, metav1.UpdateOptions{})
 }
 
+func (s *DataStore) ClearDelinquentState(volumeName string) (err error) {
+	defer func() {
+		err = errors.Wrapf(err, "failed to check ClearDelinquentState")
```
james-munson, line 287: "failed to check ClearDelinquentState" -> "failed to ClearDelinquentState".

james-munson (approved):
A nit in the logging, but this looks good.

ejweber (dismissed):
LGTM pending successful testing. Please update the PR comments (and the LEP if you will @james-munson) with the modified state machine we were discussing when you get the chance.

datastore/kubernetes.go
PhanLe1010: Fixed. Thank you

PhanLe1010:
I have some name changes and fix one more logic related to zeroTime variable @james-munson @ejweber https://github.com/longhorn/longhorn-manager/pull/3013/commits/db8e03b9de9c9a60192d780d1710d81e1671c2fd

PhanLe1010:
Testing looks good:

Following the reproducing step at https://github.com/longhorn/longhorn/issues/9089,

1. Install Longhorn master-head on a 3 worker nodes
1. Enable RWX fast failover setting 
1. Deploy a workload which use a RWX PVC provisioned by Longhorn. For example, https://github.com/longhorn/longhorn/blob/master/examples/rwx/rwx-nginx-deployment.yaml
1. Reduce the number of replicas to 1. Delete 2 replicas not on the same node as the share manager pod
1. Turn off the node of the share manager pod
1. Volume stay in faulted, lease CR is not delinquent or stale. Share manager pod is recreated but then error and deleted. Cluster is stable, no error log storm. CPU usage is small `46m`
1. Turn on the node of the share-manager pod
1. Volume is salvaged. Share manager pod is recreated and become running. Lease CR is updated normal renewal time and acquired time. Cluster is stable, no error log storm. CPU usage is small:
    ```
    longhorn-manager-r94s7                                   17m          79Mi            
    longhorn-manager-wz7p7                                   35m          69Mi    
    ```

ejweber (approved):
LGTM.

datastore/kubernetes.go
```diff
@@ -282,6 +282,39 @@ func (s *DataStore) UpdateLease(lease *coordinationv1.Lease) (*coordinationv1.Le
...
+		// Ref: IsRWXVolumeDelinquent() function
+		return nil
+	}
+	if !(lease.Spec.AcquireTime).IsZero() {
+		// Non Zero lease.Spec.AcquireTime means not delinquent.
+		// Ref: IsRWXVolumeDelinquent() function
+		return nil
+	}
+
+	oneSecondAfterZero := time.Time{}.Add(time.Second)
```
james-munson, line 310: As noted above, we could also have done this by clearing the holder identity, rather than using another special time.  But I can live with this.

PhanLe1010:
@mergify backport v1.7.x

PR #3030, opened 2024-07-29
chore(ci): add Conventional Commits GitHub Action

derekbit:

Issue Longhorn/longhorn#8824

### DIFF
--- a/.github/workflows/conventional_commits.yml
+++ b/.github/workflows/conventional_commits.yml
@@ -1,0 +1,41 @@
+name: Conventional Commits
+
+on:
+  pull_request_target:
+    types:
+    - opened
+    - edited
+    - synchronize
+    - reopened
+
+permissions:
+  pull-requests: read
+
+jobs:
+  commit-lint:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
+    - name: Lint Commits
+      uses: wagoid/commitlint-github-action@v6
+    - name: Lint Pull Request
+      uses: amannn/action-semantic-pull-request@v5.5.3
+      env:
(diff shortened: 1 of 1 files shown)


c3y1huang (approved):

.github/workflows/conventional_commits.yml
```diff
@@ -0,0 +1,41 @@
...
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
+    - name: Lint Commits
+      uses: wagoid/commitlint-github-action@v6
+    - name: Lint Pull Request
+      uses: amannn/action-semantic-pull-request@v5.5.3
+      env:
+        GITHUB_TOKEN: ${{ secrets.CUSTOM_GITHUB_TOKEN }}
```
innobead, line 26: `CUSTOM_GITHUB_TOKEN` has more permissions than the built-in one. Can we just use the built-in token (each action will automatically create a token for the action and delete it after the action is finished) instead?

ref: https://github.com/wagoid/commitlint-github-action#token

.github/workflows/conventional_commits.yml
```diff
@@ -0,0 +1,41 @@
...
+          style
+          refactor
+          perf
+          test
+          chore
+          vendor
+          build
+          ci
+          revert
+          BREAKING
```
innobead, line 41: For breaking change, we can just use the second format via `!`.
> BREAKING CHANGE: a commit that has a footer BREAKING CHANGE:, or appends a ! after the type/scope, introduces a breaking API change (correlating with [MAJOR](http://semver.org/#summary) in Semantic Versioning). A BREAKING CHANGE can be part of commits of any type.

innobead: After thinking, we can keep this as it's also valid.

PR #3036, opened 2024-07-31
fix(disk): no need to check disk UUID while deleting a disk 

derekbit:

Issue  Longhorn/longhron#9126


c3y1huang (approved):

derekbit:
@mergify backport v1.7.x

PR #3039, opened 2024-08-02
fix(backup): customize the timeout of backup binary execution

mantissahz:

Issue # longhorn/longhorn#8954 longhorn/longhorn#8319

Add a default setting `backupExecutionTimeout`

### DIFF
--- a/types/setting.go
+++ b/types/setting.go
@@ -138,6 +138,7 @@ const (
 	SettingNameFreezeFilesystemForSnapshot                              = SettingName("freeze-filesystem-for-snapshot")
 	SettingNameAutoCleanupSnapshotWhenDeleteBackup                      = SettingName("auto-cleanup-when-delete-backup")
 	SettingNameDefaultMinNumberOfBackingImageCopies                     = SettingName("default-min-number-of-backing-image-copies")
+	SettingNameBackupExecutionTimeout                                   = SettingName("backup-execution-timeout")
 	SettingNameRWXVolumeFastFailover                                    = SettingName("rwx-volume-fast-failover")
 )
 
@@ -231,6 +232,7 @@ var (
 		SettingNameFreezeFilesystemForSnapshot,
 		SettingNameAutoCleanupSnapshotWhenDeleteBackup,
 		SettingNameDefaultMinNumberOfBackingImageCopies,
+		SettingNameBackupExecutionTimeout,
 		SettingNameRWXVolumeFastFailover,
 	}
 )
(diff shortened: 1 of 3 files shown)


types/setting.go
```diff
@@ -409,6 +412,19 @@ var (
 		},
 	}
 
+	SettingDefinitionBackupEngineBinaryTimeout = SettingDefinition{
+		DisplayName: "Backup Engine Binary Timeout",
```
derekbit, line 416: Can we replace the name with a better one without `Binary`?

mantissahz: Sure, any suggestion?

mantissahz: Modified to `BackupExecuteTimeout` for `backup-execute-timeout`. Thanks

derekbit: No for now. My thought is using a generic name.

mantissahz: Modified to `BackupExecutionTimeout`: `backup-execute-timeout`,

types/setting.go
```diff
@@ -409,6 +412,19 @@ var (
 		},
 	}
 
+	SettingDefinitionBackupExecutionTimeout = SettingDefinition{
+		DisplayName: "Backup Execution Timeout",
+		Description: "In minutes. This setting determines the timeout for the backup engine execution. Set to 0 to disable the timeout.",
```
derekbit, line 417: Suggest rewording the `Set to 0 to disable the timeout.` as I suggested in longhorn/longhorn.

derekbit (approved):
LGTM

derekbit:
As we discussed, the timeout `0` (infinite wait) is not allowed.

types/setting.go
```diff
@@ -411,6 +414,19 @@ var (
 		},
 	}
 
+	SettingDefinitionBackupExecutionTimeout = SettingDefinition{
+		DisplayName: "Backup Execution Timeout",
+		Description: "In minutes. This setting determines the timeout for the backup engine execution. The default value is 1.",
```
derekbit, line 419: Waiting for the review of the description.

derekbit (dismissed):
In general, LGTM.

types/setting.go
mantissahz: Modified to `Number of minutes that Longhorn allows for backup execution. The default value is "1"`

derekbit (approved):

mantissahz:
@Mergifyio backport v1.7.x

PR #2970, opened 2024-07-15
fix(crypto): execute cryptsetup in the host namespace again

ejweber:
#### Which issue(s) this PR fixes:

longhorn/longhorn#9000

#### What this PR does / why we need it:

https://github.com/longhorn/longhorn/issues/9000#issuecomment-2228720946

#### Special notes for your reviewer:

- `examples/block/crypto` works with this PR (from https://github.com/longhorn/longhorn/issues/9000#issuecomment-2227888212).
- https://ci.longhorn.io/job/private/job/longhorn-tests-regression/7224 - PASSED
- https://ci.longhorn.io/job/private/job/longhorn-tests-regression/7225 - PASSED
- https://ci.longhorn.io/job/private/job/longhorn-tests-regression/7226 - PASSED
- https://ci.longhorn.io/job/private/job/longhorn-tests-regression/7228/ - PASSED
- I also set up a storage network and tried some basic attach/detach operations and everything worked.

@c3y1huang, can you foresee any issues with Talos / container optimized distros?

PR #3064, opened 2024-08-13
fix(networking): networking between longhorn-csi-plugin and longhorn-manager is broken after upgrading Longhorn to 1.7.0-rc3

PhanLe1010:
See the explanation at:
1. RCA https://github.com/longhorn/longhorn/issues/9223#issuecomment-2285019408
2. Proposal https://github.com/longhorn/longhorn/issues/9223#issuecomment-2285033777

longhorn/longhorn#9223

### DIFF
--- a/csi/deployment.go
+++ b/csi/deployment.go
@@ -447,12 +447,6 @@ func NewPluginDeployment(namespace, serviceAccount, nodeDriverRegistrarImage, li
 									MountPath: "/dev",
 								},
 								{
-									// The plugin must be able to switch to the host's namespaces in order to execute
-									// cryptsetup commands for encrypted devices.
-									Name:      "host-proc",
-									MountPath: "/host/proc",
-								},
-								{
 									Name:      "host-sys",
 									MountPath: "/sys",
 								},
@@ -511,14 +505,6 @@ func NewPluginDeployment(namespace, serviceAccount, nodeDriverRegistrarImage, li
 							VolumeSource: corev1.VolumeSource{
 								HostPath: &corev1.HostPathVolumeSource{
 									Path: "/dev",
-								},
-							},
-						},
-						{
-							Name: "host-proc",
-							VolumeSource: corev1.VolumeSource{
-								HostPath: &corev1.HostPathVolumeSource{
-									Path: "/proc",
 								},
 							},
 						},
--- a/csi/deployment.go
+++ b/csi/deployment.go
@@ -307,8 +307,7 @@ func NewPluginDeployment(namespace, serviceAccount, nodeDriverRegistrarImage, li
 					Tolerations:        tolerations,
 					NodeSelector:       nodeSelector,
 					PriorityClassName:  priorityClass,
-					HostNetwork:        true,
-					DNSPolicy:          corev1.DNSClusterFirstWithHostNet,
+					HostPID:            true,
 					Containers: []corev1.Container{
 						{
 							Name:  "node-driver-registrar",
@@ -458,6 +457,11 @@ func NewPluginDeployment(namespace, serviceAccount, nodeDriverRegistrarImage, li
 									MountPath: "/sys",
 								},
 								{
+									Name:             "host",
+									MountPath:        "/host",
+									MountPropagation: &MountPropagationBidirectional,
+								},
+								{
 									Name:      "lib-modules",
 									MountPath: "/lib/modules",
 									ReadOnly:  true,
@@ -523,6 +527,14 @@ func NewPluginDeployment(namespace, serviceAccount, nodeDriverRegistrarImage, li
 							VolumeSource: corev1.VolumeSource{
 								HostPath: &corev1.HostPathVolumeSource{
 									Path: "/sys",
+								},
+							},
+						},
+						{
+							Name: "host",
+							VolumeSource: corev1.VolumeSource{
+								HostPath: &corev1.HostPathVolumeSource{
+									Path: "/",
 								},
 							},
 						},
--- a/csi/deployment.go
+++ b/csi/deployment.go
@@ -307,7 +307,6 @@ func NewPluginDeployment(namespace, serviceAccount, nodeDriverRegistrarImage, li
 					Tolerations:        tolerations,
 					NodeSelector:       nodeSelector,
 					PriorityClassName:  priorityClass,
-					HostPID:            true,
 					Containers: []corev1.Container{
 						{
 							Name:  "node-driver-registrar",
@@ -447,13 +446,15 @@ func NewPluginDeployment(namespace, serviceAccount, nodeDriverRegistrarImage, li
 									MountPath: "/dev",
 								},
 								{
-									Name:      "host-sys",
-									MountPath: "/sys",
+									// The plugin must be able to switch to the host's namespaces in order to execute
+									// cryptsetup commands for encrypted devices to mount RWX volumes when not using a
+									// storage network.
+									Name:      "host-proc",
+									MountPath: "/host/proc",
 								},
 								{
-									Name:             "host",
-									MountPath:        "/host",
-									MountPropagation: &MountPropagationBidirectional,
+									Name:      "host-sys",
+									MountPath: "/sys",
 								},
 								{
 									Name:      "lib-modules",
@@ -509,18 +510,18 @@ func NewPluginDeployment(namespace, serviceAccount, nodeDriverRegistrarImage, li
 							},
 						},
 						{
-							Name: "host-sys",
+							Name: "host-proc",
 							VolumeSource: corev1.VolumeSource{
 								HostPath: &corev1.HostPathVolumeSource{
-									Path: "/sys",
+									Path: "/proc",
 								},
 							},
 						},
 						{
-							Name: "host",
+							Name: "host-sys",
 							VolumeSource: corev1.VolumeSource{
 								HostPath: &corev1.HostPathVolumeSource{
-									Path: "/",
+									Path: "/sys",
 								},
 							},
 						},
--- a/csi/node_server.go
+++ b/csi/node_server.go
@@ -812,9 +812,47 @@ func getNodeServiceCapabilities(cs []csi.NodeServiceCapability_RPC_Type) []*csi.
 	return nscs
 }
 
+func (ns *NodeServer) requireHostNamespaceMounter(volume *longhornclient.Volume, volumeCapability *csi.VolumeCapability) (bool, error) {
+	if !requiresSharedAccess(volume, volumeCapability) {
+		return false, nil
+	}
+	if volume.Migratable {
+		return false, nil
+	}
+
+	storageNetworkSetting, err := ns.apiClient.Setting.ById(string(types.SettingNameStorageNetwork))
+	if err != nil {
+		return false, err
+	}
+	storageNetworkForRWXVolumeEnabledSetting, err := ns.apiClient.Setting.ById(string(types.SettingNameStorageNetworkForRWXVolumeEnabled))
+	if err != nil {
+		return false, err
+	}
+	storageNetworkForRWXVolumeEnabled, err := strconv.ParseBool(storageNetworkForRWXVolumeEnabledSetting.Value)
+	if err != nil {
+		return false, err
+	}
+
+	if storageNetworkSetting.Value != "" && storageNetworkForRWXVolumeEnabled {
+		return false, nil
+	}
+
+	return true, nil
+}
+
 func (ns *NodeServer) getMounter(volume *longhornclient.Volume, volumeCapability *csi.VolumeCapability, volumeContext map[string]string) (mount.Interface, error) {
 	if volumeCapability.GetBlock() != nil {
 		return mount.New(""), nil
+	}
+
+	// HACK: to nsenter host namespaces for the nfs mounts to stay available after csi plugin dies.
+	// Only do this for regular RWX volume with false storageNetworkForRWXVolumeEnabledSetting
+	requireHostNamespaceMounter, err := ns.requireHostNamespaceMounter(volume, volumeCapability)
+	if err != nil {
+		return nil, err
+	}
+	if requireHostNamespaceMounter {
+		return mount.New("/usr/local/sbin/nsmounter"), nil
 	}
 
 	// mounter that can format and use hard coded filesystem params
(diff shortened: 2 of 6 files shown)


PhanLe1010:
I am leaving this as draft PR for @c3y1huang or @ejweber review the idea first

shuo-wu:
Do you know why we switched to using the host network before?
```
HostNetwork:        true,
DNSPolicy:          corev1.DNSClusterFirstWithHostNet,
```

PhanLe1010:
> Do you know why we switched to using the host network before?
> 
> ```
> HostNetwork:        true,
> DNSPolicy:          corev1.DNSClusterFirstWithHostNet,
> ```

@shuo-wu I think because we want to refactor it. Ref https://github.com/longhorn/longhorn/issues/9223#issuecomment-2285033777

csi/node_server.go
```diff
@@ -812,11 +812,51 @@ func getNodeServiceCapabilities(cs []csi.NodeServiceCapability_RPC_Type) []*csi.
 	return nscs
 }
 
+func (ns *NodeServer) requireHostNamespaceMounter(volume *longhornclient.Volume, volumeCapability *csi.VolumeCapability) (bool, error) {
+	if !requiresSharedAccess(volume, volumeCapability) {
+		return false, nil
```
c3y1huang, line 817: return true?

csi/node_server.go
```diff
@@ -812,11 +812,51 @@ func getNodeServiceCapabilities(cs []csi.NodeServiceCapability_RPC_Type) []*csi.
...
+	}
+	if volume.Migratable {
```
c3y1huang, line 819: Is this covered in requiresSharedAccess?

types/deploy.go
```diff
@@ -137,6 +128,5 @@ func UpdateDaemonSetTemplateBasedOnStorageNetwork(daemonSet *appsv1.DaemonSet, s
 		}
 	}
 
-	updateHostNetwork()
```
c3y1huang, line 140: Also need to remove https://github.com/longhorn/longhorn-manager/blob/v1.7.0-rc3/types/deploy.go#L136.

csi/deployment.go
```diff
@@ -307,8 +307,7 @@ func NewPluginDeployment(namespace, serviceAccount, nodeDriverRegistrarImage, li
 					Tolerations:        tolerations,
 					NodeSelector:       nodeSelector,
 					PriorityClassName:  priorityClass,
-					HostNetwork:        true,
-					DNSPolicy:          corev1.DNSClusterFirstWithHostNet,
+					HostPID:            true,
```
c3y1huang, line 310: The drawback of reverting to the previous implementation is that it would allow the pod to view all host processes and also the ability to gain control of the host.

What about removing longhorn-csi-plugin from the NetworkPolicy? What are the security risks here?

csi/node_server.go
PhanLe1010: I am intending to return `false` because I only want to enter host name space for regular RWX when RWX storage network is not enable.

For RWO volume, I think we don't need to enter host namespace

This is similar to the logic previously https://github.com/longhorn/longhorn-manager/blob/af348f2432967cd2981582cddb8129cab3826919/csi/node_server.go#L823-L827 

WDYT?

csi/node_server.go
PhanLe1010: I think no, ther are 2 types of RWX:
1. Regular (might need to enter host namespace if RWX storage network is not enabled)
2. Migratable volume for Havester (no need to enter host namespace ever) 

WDYT?

types/deploy.go
PhanLe1010: Thank you. Fixed

csi/deployment.go
PhanLe1010: > The drawback of reverting to the previous implementation is that it would allow the pod to view all host processes and also the ability to gain control of the host.

I see. Yeah.

> What about removing longhorn-csi-plugin from the NetworkPolicy? What are the security risks here?

But I don't know how to do this. The network policy is to add more pod into the allow list. If we don't add longhorn-csi-plugin into the policy it will be blocked by default. Any idea?

csi/node_server.go
c3y1huang: I see. Thank you.

csi/node_server.go
c3y1huang: I see. Thank you.

csi/deployment.go
c3y1huang: > But I don't know how to do this. The network policy is to add more pod into the allow list. If we don't add longhorn-csi-plugin into the policy it will be blocked. Any idea?

Does CIDR notation work?

PhanLe1010: To whitelist the host IPs? 

I think host IPs are dynamic depending on each cluser setup? 

Also, this mean we allow any host processes to hit the longhorn-manager pods which sounds a bit off to me

WDYT?

c3y1huang: Yea, true.

c3y1huang (changes requested):
A small change request; everything else looks good to me. And I've run some basic tests with `storage-network-for-rwx-volume-enabled: true` with this PR.

csi/deployment.go
Vicente-Cheng: >And, how do we handle the scenario where both the network policy and storage network for RWX volume are enabled? Could this be a use case for Harvester?

Sorry @c3y1huang, I am confused with that.

It looks like we will get hostPID back, right? If yes, hostPID will break the storage network. (from the comment https://github.com/longhorn/longhorn/issues/8184#issuecomment-2124179154)
Does it mean we need to find another way to support the storage network on the RWX volume?

csi/node_server.go
```diff
@@ -812,11 +812,51 @@ func getNodeServiceCapabilities(cs []csi.NodeServiceCapability_RPC_Type) []*csi.
...
+		return false, nil
+	}
+
+	storageNetworkSetting, err := ns.apiClient.Setting.ById(string(types.SettingNameStorageNetwork))
+	if err != nil {
+		return false, err
+	}
+	if storageNetworkSetting.Value == "" {
```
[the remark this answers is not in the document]
PhanLe1010: Good catch! Thank you very much. Fixed

PhanLe1010:
Doing more testing. Please hold on merging

PhanLe1010:
Test case 2 at https://github.com/longhorn/longhorn/issues/9223#issuecomment-2287014612 is failing . Not sure why I cannot mount the NFS within the host namespace either from longhorn-csi-plugin OR ssh into the host directly. Weird

PhanLe1010:
OK. So looks like in v1.7.0-rc3, we have switched from using IP of the share manager service to using FQN. I.e., switching from something like:
`10.42.68.91`
to something like
`pvc-0c47b0cd-d2af-4641-8f0b-7da85b99cc93.longhorn-system.svc.cluster.local`

Ref the commit https://github.com/longhorn/longhorn-manager/commit/27e02192978fac47fac23305749e54acd047c511

The problem is that now when we enter the host namespace and doing NFS mount like this:
`mount -t nfs -o vers=4.1,noresvport,timeo=600,retrans=5,softerr pvc-0c47b0cd-d2af-4641-8f0b-7da85b99cc93.longhorn-system.svc.cluster.local:/pvc-0c47b0cd-d2af-4641-8f0b-7da85b99cc93 /var/lib/kubelet/plugins/kubernetes.io/csi/driver.longhorn.io/0af2eb7eee4f66591c49921b5832b8b0ddf608fd9303f6d242180273061655a1/globalmount`

It is failing because it cannot resolve the domain name `pvc-0c47b0cd-d2af-4641-8f0b-7da85b99cc93.longhorn-system.svc.cluster.local` from the host namespace. Leading to this error:
```
[longhorn-csi-plugin-knlq7 longhorn-csi-plugin] E0813 23:16:14.992960 1417708 mount_linux.go:236] Mount failed: exit status 32
[longhorn-csi-plugin-knlq7 longhorn-csi-plugin] Mounting command: /usr/local/sbin/nsmounter
[longhorn-csi-plugin-knlq7 longhorn-csi-plugin] Mounting arguments: mount -t nfs -o vers=4.1,noresvport,timeo=600,retrans=5,softerr pvc-0c47b0cd-d2af-4641-8f0b-7da85b99cc93.longhorn-system.svc.cluster.local:/pvc-0c47b0cd-d2af-4641-8f0b-7da85b99cc93 /var/lib/kubelet/plugins/kubernetes.io/csi/driver.longhorn.io/0af2eb7eee4f66591c49921b5832b8b0ddf608fd9303f6d242180273061655a1/globalmount
[longhorn-csi-plugin-knlq7 longhorn-csi-plugin] Output: mount.nfs: Resource temporarily unavailable
```

cc @c3y1huang

c3y1huang:
> OK. So looks like in v1.7.0-rc3, we have switched from using IP of the share manager service to using FQN. I.e., switching from something like:
> `10.42.68.91`
> to something like
> `pvc-0c47b0cd-d2af-4641-8f0b-7da85b99cc93.longhorn-system.svc.cluster.local`

Sorry for missing this in the review. The service FQDN won't resolve in the host namespace, so we need to use the cluster IP for regular RWX volume. However, for storage network enabled RWX volumes, we still need to use the service FQDN due to the usage of headless service.

PhanLe1010:
No worries at all @c3y1huang I think it is my bad not to testing this quick enough :D 

I am thinking to remove the `-u` over here https://github.com/longhorn/longhorn-manager/blob/8bfc78b23a957dc48a6f624fafb11a127687b7e8/package/nsmounter#L46 so that we don't enter the  UTS namespace of the host process. I am currently understand that it would allow use to lookup the FQN from the container namespace. WDYT?


But it is still falling ... Maybe I am missing something?

csi/deployment.go
c3y1huang: We will get the `hostPID` and the custom `mounter` back. The custom mounter is a wrapper of `nsenter` around the mounter commands to execute in the host namespace, we will only apply this to regular RWX volumes. For storage network enabled RWX volumes, the default mounter will be used, this means the mounting will be done within the CSI plugin pod container. Therefore, this change shouldn't impact storage network support for RWX volumes.

Vicente-Cheng: Hi @c3y1huang,
That's clear.
- Regular RWX -> use nsenter for custom mounter
- Storage Network RWX -> use normal mounted

Thanks for the explanation for it!

c3y1huang:
> No worries at all @c3y1huang I think it is my bad not to testing this quick enough :D
> 
> I am thinking to remove the `-u` over here
> 
> https://github.com/longhorn/longhorn-manager/blob/8bfc78b23a957dc48a6f624fafb11a127687b7e8/package/nsmounter#L46
> 
> so that we don't enter the UTS namespace of the host process. I am currently understand that it would allow use to lookup the FQN from the container namespace. WDYT?
> But it is still falling ... Maybe I am missing something?

Afaik, DNS resolution within Kubernetes is typically managed by CoreDNS, which manages custom DNS entries and service discoveries that are different from the host's DNS configuration. So, resolving Kubernetes service FQDN from the host would require additional configuration on the host.  This is likely why the original implementation uses the cluster IP instead.

PhanLe1010:
Got it. Thanks @c3y1huang pushed the fix at https://github.com/longhorn/longhorn-manager/pull/3064/commits/7892c33151c361a131da66056b5597343dfdc77e Could you take a look to see if it is ok?

I also remove the `/host/proc` mount because we already reverted back to mount the whole `/host` https://github.com/longhorn/longhorn-manager/pull/3064/commits/57d6d2f5c3df15edf7443de5aabb7a611395411b cc @ejweber Any concern?

c3y1huang (approved):

ejweber (commented):
I'm just now catching up here, but I have no major concerns. It looks like this should work as discussed and tested.

Regarding https://github.com/longhorn/longhorn-manager/pull/3064#issuecomment-2287561351 and the changes I previously made in https://github.com/longhorn/longhorn-manager/pull/2970, I am a bit skeptical of the need to do BOTH of the following:

- Add back `HostPID`, and
- Have the `/host/` mount.

It looks like we only need `HostPID` because the `package/nsmounter` script looks in `/proc` to find a namespace to enter. Since we have the `/host/proc` mount, can't we just look there instead? (The other sneaky issue with losing `HostPID` was with encrypted volumes, but that is already solved by https://github.com/longhorn/longhorn-manager/pull/2970 and isn't affected by these proposed changes.)

ejweber:
> It looks like we only need `HostPID` because the `package/nsmounter` script looks in `/proc` to find a namespace to enter. Since we have the `/host/proc` mount, can't we just look there instead? (The other sneaky issue with losing `HostPID` was with encrypted volumes, but that is already solved by #2970 and isn't affected by these proposed changes.)

It wasn't quite that simple because `package/nsmounter` uses the `-t` option instead of referencing the files in `/proc/host` to do the namespace switching. However, there's an easy fix. WDYT about https://github.com/longhorn/longhorn-manager/commit/390f26c72e4857c37be2d883abfadcbe1efbd731, which removes `HostPID` and reverts to only doing the `/host/proc` mount, but modifies `package/nsmounter` so it all works fine @PhanLe1010? It has passed all the tests from your test plan that I have tried so far, but I don't have the right environment for the robot tests currently.

PhanLe1010:
Thanks @ejweber for digging into finding even more better refactoring!

I think your approach looks good. However, since the time for 1.7.0 is tight, I am a bit hesitant to push more new logic. Is it ok to put this refactor to a new ticket for 1.8.0 milestone?

ejweber:
> Thanks @ejweber for digging into finding even better refactoring!
> 
> I think your approach looks good. However, since the time for 1.7.0 is tight, I am a bit hesitant to push more new logic unless it is bug fix. Is it ok to put this refactor to a new ticket for 1.8.0 milestone?

I discussed this a bit with @PhanLe1010 offline. I can understand the desire to wait, but I feel pretty strongly that it is better resolve the question of whether we need `HostPID` and what we need to mount to the CSI plugin pod in this version if possible, while we are messing with the mechanisms in general. I have pushed the referenced commit to this PR, and we are planning on merging it.

ejweber (approved):

PhanLe1010:
@mergify backport v1.7.x

PhanLe1010:
Thanks @ejweber for the great refinement

PR #3060, opened 2024-08-11
refactor: remove isCLIAPIVersionOne related codes

derekbit:
#### Which issue(s) this PR fixes:

Issue Longhorn/longhorn#7191

PR #3045, opened 2024-08-05
fix(manager): add detailed message when deleting pdb is not allowed

w13915984028:
#### Which issue(s) this PR fixes:

Issue # https://github.com/longhorn/longhorn/issues/9183
btw: https://github.com/longhorn/longhorn/issues/5910

#### What this PR does / why we need it:

Log more detailed message to help troubleshooting

#### Additional documentation or context

Test log of manual attached volume when drain node

>time="2024-08-05T18:48:14Z" level=info msg="Node instance-manager-68cd2514dd3f6d59b95cbd865d5b08f7 is marked unschedulable but removing harv2 PDB is blocked: some volumes are still attached " func="controller.(*InstanceManagerController).syncInstanceManagerPDB" file="instance_manager_controller.go:730" controller=longhorn-instance-manager node=harv2

PR #3069, opened 2024-08-16
fix(lint): avoid some lint errors about deprecated fields

james-munson:

No tracking issue.

### DIFF
--- a/controller/instance_manager_controller.go
+++ b/controller/instance_manager_controller.go
@@ -940,14 +940,14 @@ func formatInstanceMessage(im *longhorn.InstanceManager) string {
 		}
 	}
 
-	if len(im.Status.Instances) > 0 {
+	if len(im.Status.Instances) > 0 { // nolint: staticcheck
 		if ieFormated {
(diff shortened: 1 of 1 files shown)


PhanLe1010:
Looks like it is overlapping with this PR https://github.com/longhorn/longhorn-manager/pull/3060

ejweber:
This one is caused by me.

https://github.com/longhorn/longhorn-manager/pull/3045#issuecomment-2290020232

I chose not to apply this fix because it entirely disables `staticcheck` for these lines instead of just `SA1019`. I thought it would not trigger again if that code wasn't being modified, since we reference the deprecated field in other places in the file. I guess that was incorrect.

ejweber (approved):
LGTM. Sorry for the inconvenience.

ejweber:
I spoke with @james-munson and @PhanLe1010 offline and we decided to merge this to clean up master in case https://github.com/longhorn/longhorn-manager/pull/3060 doesn't merge for a while.

PR #3077, opened 2024-08-19
fix(manager): fix logic for when RWX workload is restarted after node…

james-munson:
… failure

https://github.com/longhorn/longhorn/issues/9095

### DIFF
--- a/controller/kubernetes_pod_controller.go
+++ b/controller/kubernetes_pod_controller.go
@@ -209,19 +209,12 @@ func (kc *KubernetesPodController) handleWorkloadPodDeletionIfCSIPluginPodIsDown
 		return nil
 	}
 
-	storageNetworkSetting, err := kc.ds.GetSettingWithAutoFillingRO(types.SettingNameStorageNetwork)
+	isStorageNetworkForRWXVolume, err := kc.ds.IsStorageNetworkForRWXVolume()
 	if err != nil {
-		log.WithError(err).Warnf("%s. Failed to get setting %v", logAbort, types.SettingNameStorageNetwork)
+		log.WithError(err).Warnf("%s. Failed to check isStorageNetwork", logAbort)
 		return nil
 	}
-
-	storageNetworkForRWXVolumeEnabled, err := kc.ds.GetSettingAsBool(types.SettingNameStorageNetworkForRWXVolumeEnabled)
-	if err != nil {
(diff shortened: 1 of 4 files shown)


controller/kubernetes_pod_controller.go
```diff
@@ -329,6 +322,20 @@ func (kc *KubernetesPodController) handleWorkloadPodDeletionIfCSIPluginPodIsDown
 	return nil
 }
 
+func (kc *KubernetesPodController) isStorageNetworkForRWXVolume() (bool, error) {
```
ejweber, line 325: NIT: This is an exact replica of a function in the share-manager controller. Can we make it a datastore method instead? Or a utility function that takes a datastore?

controller/kubernetes_pod_controller.go
```diff
@@ -493,6 +505,11 @@ func (kc *KubernetesPodController) handlePodDeletionIfVolumeRequestRemount(pod *
 		if vol.Status.RemountRequestedAt == "" {
 			continue
 		}
+
+		if isRegularRWXVolume(vol) && !isStorageNetworkForRWXVolume {
```
ejweber, line 509: NIT: It is obvious why in context, but can we add a comment here explaining the logic? The following as an example, or something better?

```suggestion
		// NFS clients can generally recover without a remount when the NFS server restarts using the same ClusterIP. 
		// A remount is required when a storage network is in use because the new NFS server has a different IP.
		if isRegularRWXVolume(vol) && !isStorageNetworkForRWXVolume {
```

ejweber (commented):
Should we remove the code at https://github.com/longhorn/longhorn-manager/blob/69f8cdb6c201e9d02ecd3d60027773ea9e121557/controller/volume_controller.go#L4524-L4531? If no, should we modify the comment?

controller/kubernetes_pod_controller.go
james-munson: Done.

controller/kubernetes_pod_controller.go
james-munson: Done.

james-munson:
I thought about removing the kick from volume_controller, but it is still necessary in some cases.  Comment revised to be more accurate, though.

ejweber (approved):
LGTM. Thanks!

PhanLe1010 (dismissed):
LGTM

Could you please help to resolve the conflict for merging @james-munson ?

PhanLe1010 (approved):

james-munson:
@mergify backport v1.7.x

PR #3078, opened 2024-08-21
fix(networking): cleanup service/endpoint if needed

Vicente-Cheng:

Issue https://github.com/longhorn/longhorn/issues/9272

We meet a corner case that the service/endpoint would not be cleanup. That will cause the service keep the ClusterIP `None`. With this config, the endpoint of sharemanager would not be correct, so the CSI driver cannot perform the mountpoint well.

We would like to have a checking mechanism to know if the service/endpoint did not cleanup. Then we will cleanup the service/endpoint to ensure the correct endpoint.

### DIFF
--- a/controller/share_manager_controller.go
+++ b/controller/share_manager_controller.go
@@ -451,17 +451,12 @@ func (c *ShareManagerController) syncShareManagerEndpoint(sm *longhorn.ShareMana
 		return nil
 	}
 
-	storageNetwork, err := c.ds.GetSettingWithAutoFillingRO(types.SettingNameStorageNetwork)
+	storageNetworkForRWXVolume, err := c.isStorageNetworkForRWXVolume()
 	if err != nil {
 		return err
 	}
 
-	storageNetworkForRWXVolumeEnabled, err := c.ds.GetSettingAsBool(types.SettingNameStorageNetworkForRWXVolumeEnabled)
-	if err != nil {
-		return err
-	}
-
-	if types.IsStorageNetworkForRWXVolume(storageNetwork, storageNetworkForRWXVolumeEnabled) {
+	if storageNetworkForRWXVolume {
 		serviceFqdn := fmt.Sprintf("%v.%v.svc.cluster.local", sm.Name, sm.Namespace)
 		sm.Status.Endpoint = fmt.Sprintf("nfs://%v/%v", serviceFqdn, sm.Name)
 	} else {
@@ -1050,6 +1045,92 @@ func (c *ShareManagerController) getShareManagerTolerationsFromStorageClass(sc *
 	return tolerations
 }
 
+func (c *ShareManagerController) isStorageNetworkForRWXVolume() (bool, error) {
+	storageNetwork, err := c.ds.GetSettingWithAutoFillingRO(types.SettingNameStorageNetwork)
+	if err != nil {
+		return false, errors.Wrapf(err, "failed to get setting value %v", types.SettingNameStorageNetwork)
+	}
+
+	storageNetworkForRWXVolumeEnabled, err := c.ds.GetSettingAsBool(types.SettingNameStorageNetworkForRWXVolumeEnabled)
+	if err != nil {
+		return false, errors.Wrapf(err, "failed to get setting value %v", types.SettingNameStorageNetworkForRWXVolumeEnabled)
+	}
+
+	return types.IsStorageNetworkForRWXVolume(storageNetwork, storageNetworkForRWXVolumeEnabled), nil
+}
+
+func (c *ShareManagerController) checkStorageNetworkApplied() (bool, error) {
+	targetSettings := []types.SettingName{types.SettingNameStorageNetwork, types.SettingNameStorageNetworkForRWXVolumeEnabled}
+	for _, item := range targetSettings {
+		if applied, err := c.ds.GetSettingApplied(item); err != nil || !applied {
+			return applied, err
+		}
+	}
+	return true, nil
+}
+
+func (c *ShareManagerController) canCleanupService(shareManagerName string) (bool, error) {
+	service, err := c.ds.GetService(c.namespace, shareManagerName)
+	if err != nil {
+		// if NotFound, means the service/endpoint is already cleaned up
+		// The service and endpoint are related with the kubernetes endpoint controller.
+		// It means once the service is deleted, the corresponding endpoint will be deleted automatically.
+		if apierrors.IsNotFound(err) {
+			return false, nil
+		}
+		return false, errors.Wrap(err, "failed to get service")
+	}
+
+	// check the settings status of storage network and storage network for RWX volume
+	settingsApplied, err := c.checkStorageNetworkApplied()
+	if err != nil {
+		return false, errors.Wrap(err, "failed to check if the storage network settings are applied")
+	}
+	if !settingsApplied {
+		c.logger.Warn("Storage network settings are not applied, do nothing")
+		return false, nil
+	}
+
+	storageNetworkForRWXVolume, err := c.isStorageNetworkForRWXVolume()
+	if err != nil {
+		return false, err
+	}
+
+	// no need to cleanup because looks the service file is correct
+	if storageNetworkForRWXVolume {
+		if service.Spec.ClusterIP == core.ClusterIPNone {
+			return false, nil
+		}
+	} else {
+		if service.Spec.ClusterIP != core.ClusterIPNone {
+			return false, nil
+		}
+	}
+	return true, nil
+}
+
+func (c *ShareManagerController) cleanupService(shareManager *longhorn.ShareManager) error {
+	if ok, err := c.canCleanupService(shareManager.Name); !ok || err != nil {
+		if err != nil {
+			return errors.Wrapf(err, "failed to check if we can cleanup service and endpoint for share manager %v", shareManager.Name)
+		}
+		return nil
+	}
+
+	// let's cleanup
+	c.logger.Infof("Deleting Service for share manager %v", shareManager.Name)
+	err := c.ds.DeleteService(c.namespace, shareManager.Name)
+	if err != nil && !apierrors.IsNotFound(err) {
+		return errors.Wrapf(err, "failed to delete service for share manager %v", shareManager.Name)
+	}
+
+	// we don't need to cleanup the endpoint because the kubernetes endpoints_controller
+	// will sync the service then clean up the corresponding endpoint.
+	// https://github.com/kubernetes/kubernetes/blob/v1.31.0/pkg/controller/endpoint/endpoints_controller.go#L374-L392
+
+	return nil
+}
+
 func (c *ShareManagerController) createServiceAndEndpoint(shareManager *longhorn.ShareManager) error {
 	// check if we need to create the service
 	_, err := c.ds.GetService(c.namespace, shareManager.Name)
@@ -1121,6 +1202,11 @@ func (c *ShareManagerController) createShareManagerPod(sm *longhorn.ShareManager
 		return nil, errors.Wrap(err, "failed to get priority class setting before creating share manager pod")
 	}
 	priorityClass := setting.Value
+
+	err = c.cleanupService(sm)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to cleanup service for share manager %v", sm.Name)
+	}
 
 	err = c.createServiceAndEndpoint(sm)
 	if err != nil {
@@ -1276,17 +1362,12 @@ func (c *ShareManagerController) createServiceManifest(sm *longhorn.ShareManager
 
 	log := getLoggerForShareManager(c.logger, sm)
 
-	storageNetwork, err := c.ds.GetSettingWithAutoFillingRO(types.SettingNameStorageNetwork)
+	storageNetworkForRWXVolume, err := c.isStorageNetworkForRWXVolume()
 	if err != nil {
-		log.WithError(err).Warnf("Failed to get %v setting, fallback to cluster IP", types.SettingNameStorageNetwork)
+		log.WithError(err).Warnf("Failed to check storage network for RWX volume")
 	}
 
-	storageNetworkForRWXVolumeEnabled, err := c.ds.GetSettingAsBool(types.SettingNameStorageNetworkForRWXVolumeEnabled)
-	if err != nil {
-		log.WithError(err).Warnf("Failed to get %v setting, fallback to cluster IP", types.SettingNameStorageNetworkForRWXVolumeEnabled)
-	}
-
-	if types.IsStorageNetworkForRWXVolume(storageNetwork, storageNetworkForRWXVolumeEnabled) {
+	if storageNetworkForRWXVolume {
 		// Create a headless service do it doesn't use a cluster IP. This allows
 		// directly reaching the share manager pods using their individual
 		// IP address.
--- a/controller/setting_controller.go
+++ b/controller/setting_controller.go
@@ -359,14 +359,6 @@ func (sc *SettingController) syncDangerZoneSettingsForManagedComponents(settingN
 					return &types.ErrorInvalidState{Reason: fmt.Sprintf("failed to apply %v setting to Longhorn components when there are attached volumes. It will be eventually applied", types.SettingNameStorageNetworkForRWXVolumeEnabled)}
 				}
 
-				// Perform cleanup of the share manager Service
-				// This is to allow the creation of the correct Service
-				// and Endpoint when switching between cluster network
-				// and storage network.
-				if err := sc.cleanupShareManagerServiceAndEndpoints(); err != nil {
-					return err
-				}
-
 				return nil
 			}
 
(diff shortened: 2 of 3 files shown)


controller/share_manager_controller.go
```diff
@@ -1050,6 +1050,50 @@ func (c *ShareManagerController) getShareManagerTolerationsFromStorageClass(sc *
 	return tolerations
 }
 
+func (c *ShareManagerController) cleanupServiceAndEndpoint(shareManager *longhorn.ShareManager) error {
+	service, err := c.ds.GetService(c.namespace, shareManager.Name)
+	if err != nil {
+		// if NotFound, means the service/endpoint is already cleaned up
```
c3y1huang, line 1056: I'm thinking of removing the cleanup from the setting controller so it can be handled just here instead. What do you think?

Vicente-Cheng: After discussing this offline with @c3y1huang, we will add the status check here and remove the cleanup from the setting controller.

I will mark it ready once I complete the above changes. Thanks!

Vicente-Cheng: Updated! Please help to review. Thanks!

Vicente-Cheng:
codeFactor checking failure is not related to this PR.

james-munson (approved):
This looks reasonable to me.

controller/share_manager_controller.go
```diff
@@ -1050,6 +1050,69 @@ func (c *ShareManagerController) getShareManagerTolerationsFromStorageClass(sc *
...
+		if apierrors.IsNotFound(err) {
+			return nil
+		}
+		return errors.Wrapf(err, "failed to get service for share manager %v", shareManager.Name)
+	}
+
+	applied, err := c.checkStorageNetworkApplied()
+	if err != nil {
+		return errors.Wrapf(err, "failed to check if the storage network setting is applied")
```
c3y1huang, line 1075:
```suggestion
		return errors.Wrap(err, "failed to check if the storage network settings are applied")
```

controller/share_manager_controller.go
```diff
@@ -1050,6 +1050,69 @@ func (c *ShareManagerController) getShareManagerTolerationsFromStorageClass(sc *
...
+	}
+	if !applied {
+		c.logger.Warnf("`StorageNetwork` related settings are not applied, do nothing.")
```
c3y1huang, line 1078:
```suggestion
		c.logger.Warn("storage network settings are not applied, do nothing")
```

controller/share_manager_controller.go
```diff
@@ -1050,6 +1050,69 @@ func (c *ShareManagerController) getShareManagerTolerationsFromStorageClass(sc *
...
+		return nil
+	}
+
+	storageNetwork, err := c.ds.GetSettingWithAutoFillingRO(types.SettingNameStorageNetwork)
```
c3y1huang, line 1082: Nit: I am thinking if we should move the service and endpoint check validation to a separate method like what you've done for `c.checkStorageNetworkApplied()`. I am also fine with the current, what do you think?

c3y1huang (changes requested):

controller/share_manager_controller.go
derekbit: Capitalized "Storage network ...."

controller/share_manager_controller.go
```diff
@@ -1050,6 +1045,81 @@ func (c *ShareManagerController) getShareManagerTolerationsFromStorageClass(sc *
...
+	} else {
+		if service.Spec.ClusterIP != core.ClusterIPNone {
+			return nil
+		}
+	}
+
+	// let's cleanup
+	c.logger.Infof("Deleting Service for sharemanager %v", shareManager.Name)
+	err = c.ds.DeleteService(c.namespace, shareManager.Name)
+	if err != nil && !apierrors.IsNotFound(err) {
```
derekbit, line 1110:
```suggestion
		return errors.Wrapf(err, "failed to delete endpoint for share manager %v", shareManager.Name)
```

controller/share_manager_controller.go
```diff
@@ -1050,6 +1045,81 @@ func (c *ShareManagerController) getShareManagerTolerationsFromStorageClass(sc *
...
+	if storageNetworkForRWXVolumeValidate {
+		if service.Spec.ClusterIP == core.ClusterIPNone {
+			return nil
+		}
+	} else {
+		if service.Spec.ClusterIP != core.ClusterIPNone {
+			return nil
+		}
+	}
+
```
derekbit, line 1106: Suggest abstracting the codes to `canCleanupServiceAndEndpoint()`?

controller/share_manager_controller.go
```diff
@@ -1050,6 +1045,81 @@ func (c *ShareManagerController) getShareManagerTolerationsFromStorageClass(sc *
...
+	// let's cleanup
+	c.logger.Infof("Deleting Service for sharemanager %v", shareManager.Name)
```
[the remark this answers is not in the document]
Vicente-Cheng: updated!

controller/share_manager_controller.go
Vicente-Cheng: Hi @derekbit, I am not sure of this.
We are trying to delete the service in this section, so the error msg should be `failed to delete service` instead of `failed to delete endpoint`. Or did I get some misunderstanding?

controller/share_manager_controller.go
Vicente-Cheng: updated!

controller/share_manager_controller.go
Vicente-Cheng: Hi @c3y1huang, I wrapped some helpful functions like `validateStorageNetworkForRWXVolume` and `canCleanupServiceAndEndpoint`.

Please help to recheck it. Thanks!

controller/share_manager_controller.go
derekbit: oh, typo. should be `service`

controller/share_manager_controller.go
```diff
@@ -1050,6 +1045,92 @@ func (c *ShareManagerController) getShareManagerTolerationsFromStorageClass(sc *
...
+	c.logger.Infof("Deleting Service for share manager %v", shareManager.Name)
+	err := c.ds.DeleteService(c.namespace, shareManager.Name)
+	if err != nil && !apierrors.IsNotFound(err) {
+		return errors.Wrapf(err, "failed to delete service for share manager %v", shareManager.Name)
+	}
+
+	c.logger.Infof("Deleting Endpoint for share manager %v", shareManager.Name)
+	err = c.ds.DeleteKubernetesEndpoint(c.namespace, shareManager.Name)
+	if err != nil && !apierrors.IsNotFound(err) {
+		return errors.Wrapf(err, "failed to delete Endpoint for share manager %v", shareManager.Name)
```
derekbit, line 1128:
```suggestion
		return errors.Wrapf(err, "failed to delete endpoint for share manager %v", shareManager.Name)
```

controller/share_manager_controller.go
```diff
@@ -1050,6 +1045,92 @@ func (c *ShareManagerController) getShareManagerTolerationsFromStorageClass(sc *
...
+func (c *ShareManagerController) cleanupServiceAndEndpoint(shareManager *longhorn.ShareManager) error {
+	if ok, err := c.canCleanupServiceAndEndpoint(c.namespace, shareManager.Name); !ok || err != nil {
+		if err != nil {
+			return errors.Wrapf(err, "failed to check if we can cleanup service and endpoint for share manager %v", shareManager.Name)
+		}
+		return nil
+	}
+
+	// let's cleanup
+	c.logger.Infof("Deleting Service for share manager %v", shareManager.Name)
```
derekbit, line 1119:
```suggestion
	c.logger.Infof("Deleting service for share manager %v", shareManager.Name)
```

controller/share_manager_controller.go
```diff
@@ -1050,6 +1045,92 @@ func (c *ShareManagerController) getShareManagerTolerationsFromStorageClass(sc *
...
+	if storageNetworkForRWXVolumeValidate {
+		if service.Spec.ClusterIP == core.ClusterIPNone {
+			return false, nil
+		}
+	} else {
+		if service.Spec.ClusterIP != core.ClusterIPNone {
+			return false, nil
+		}
+	}
+	return true, nil
```
c3y1huang, line 1107: Suggest moving this condition check into `c.validateStorageNetworkForRWXVolume()`.

controller/share_manager_controller.go
```diff
@@ -1050,6 +1045,92 @@ func (c *ShareManagerController) getShareManagerTolerationsFromStorageClass(sc *
...
+
+func (c *ShareManagerController) canCleanupServiceAndEndpoint(ns, shareManagerName string) (bool, error) {
+	service, err := c.ds.GetService(c.namespace, shareManagerName)
+	if err != nil {
+		// if NotFound, means the service/endpoint is already cleaned up
+		if apierrors.IsNotFound(err) {
+			return false, nil
+		}
+		return false, errors.Wrap(err, "failed to get service")
+	}
```
derekbit, line 1080: Is it possible the service is gone but the endpoint is still present?

controller/share_manager_controller.go
```diff
@@ -1050,6 +1045,92 @@ func (c *ShareManagerController) getShareManagerTolerationsFromStorageClass(sc *
 	return tolerations
 }
 
+func (c *ShareManagerController) validateStorageNetworkForRWXVolume() (bool, error) {
```
c3y1huang, line 1048: Should using a similar naming convention to `checkStorageNetworkApplied`, for example:
```suggestion
func (c *ShareManagerController) checkStorageNetworkForRWXVolumeIsValid() (bool, error) {
```

c3y1huang (changes requested):

controller/share_manager_controller.go
Vicente-Cheng: Honestly, the endpoint doesn't matter.
The main factor affecting the `status.endpoint` of ShareManager (w/o storage network) is the service.
The endpoint will be updated no matter if it is recreated or not.

But if we do not clean up the service file, the ClusterIP will already be set to `None`. This will block the next update for the ClusterIP (when we disable the storage network).

So the service is the main thing we need to take care of here.

cc @c3y1huang (I am afraid I got something misunderstanding)

derekbit: But you error out in `err = c.ds.DeleteKubernetesEndpoint(c.namespace, shareManager.Name)` if the deletion failed? Does it mean we don't need to error out and can just leave a warning message instead?

Vicente-Cheng: Hmm, yeah, that's true.

I thought it would not be harmful if we deleted the endpoint with failure and did nothing next time.
But if we would like to keep the consistency, we could add extra checking for endpoint like a service.

WDYT?

controller/share_manager_controller.go
c3y1huang: After discussing with @Vicente-Cheng , we will keep the current implementation because it's reused by other callers. However, we will rename `c.validateStorageNetworkForRWXVolume()` to better reflect its purpose.

controller/share_manager_controller.go
Vicente-Cheng: I discussed this with @c3y1huang and did some tests.

It looks like Kubernetes will clean up the `endpoint` once the corresponding `service` is removed.
So the case `service` deleted and 'endpoint` existing will be handled on the Kubernetes side.
It will not happen, I will also add the comment on the code.

c3y1huang: https://github.com/kubernetes/kubernetes/blob/v1.31.0/pkg/controller/endpoint/endpoints_controller.go#L374-L392

controller/share_manager_controller.go
```diff
@@ -1050,6 +1045,92 @@ func (c *ShareManagerController) getShareManagerTolerationsFromStorageClass(sc *
...
+	// let's cleanup
+	c.logger.Infof("Deleting Service for share manager %v", shareManager.Name)
+	err := c.ds.DeleteService(c.namespace, shareManager.Name)
+	if err != nil && !apierrors.IsNotFound(err) {
+		return errors.Wrapf(err, "failed to delete service for share manager %v", shareManager.Name)
+	}
+
+	c.logger.Infof("Deleting Endpoint for share manager %v", shareManager.Name)
+	err = c.ds.DeleteKubernetesEndpoint(c.namespace, shareManager.Name)
+	if err != nil && !apierrors.IsNotFound(err) {
```
derekbit, line 1127: According to https://github.com/longhorn/longhorn-manager/pull/3078#discussion_r1728200155, the error out is not necessary, so we can just add a warning message here.

Vicente-Cheng: updated and also added comments for the reason

controller/share_manager_controller.go
Vicente-Cheng: Renamed the `validateStorageNetworkForRWXVolume` -> `isStorageNetworkForRWXVOlume` for better readability. Please help to recheck!

controller/share_manager_controller.go
Vicente-Cheng: Thanks for the code path reference. I updated it in the comment also!

controller/share_manager_controller.go
Vicente-Cheng: updated!

controller/share_manager_controller.go
Vicente-Cheng: updated!

controller/share_manager_controller.go
c3y1huang: Do we still need to handle the Endpoint deletion? or should we let the Kubernetes controller handle it entirely? I vote for the latter.

Vicente-Cheng: +1, if the Kubernetes controller can handle it well. We could skip the deletion from ourselves. WDYT?

derekbit: Yes, same thought. We don't really need to delete it.

Vicente-Cheng: Updated! BTW, I renamed the `cleanupServiceAndEndpoing` to `cleanupService` and added the comment explaining why we did not delete the endpoint.
Please help to check. Thanks!

controller/share_manager_controller.go
```diff
@@ -1050,6 +1045,92 @@ func (c *ShareManagerController) getShareManagerTolerationsFromStorageClass(sc *
...
+	// let's cleanup
+	c.logger.Infof("Deleting Service for share manager %v", shareManager.Name)
+	err := c.ds.DeleteService(c.namespace, shareManager.Name)
+	if err != nil && !apierrors.IsNotFound(err) {
+		return errors.Wrapf(err, "failed to delete service for share manager %v", shareManager.Name)
+	}
+
+	// we don't need to cleanup the endpoint because the kubernetes endpoints_controller
+	// will sync the service then clean up the corresponding endpoint.
+	// https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/endpoint/endpoints_controller.go#L374-L392
```
c3y1huang, line 1129: Should avoid using the master branch as the targeted lines may change over time.

derekbit (approved):
LGTM.

derekbit:
Need @c3y1huang's final check and approval. Thank you.

controller/share_manager_controller.go
```diff
@@ -1122,6 +1203,11 @@ func (c *ShareManagerController) createShareManagerPod(sm *longhorn.ShareManager
 	}
 	priorityClass := setting.Value
 
+	err = c.cleanupService(sm)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to cleanup service and endpoint for share manager %v", sm.Name)
```
c3y1huang, line 1208:
```suggestion
		return nil, errors.Wrapf(err, "failed to cleanup service for share manager %v", sm.Name)
```

c3y1huang (changes requested):

controller/share_manager_controller.go
Vicente-Cheng: Nice suggestion! Updated!

controller/share_manager_controller.go
Vicente-Cheng: updated!

controller/share_manager_controller.go
```diff
@@ -1050,6 +1045,92 @@ func (c *ShareManagerController) getShareManagerTolerationsFromStorageClass(sc *
...
+	targetSettings := []types.SettingName{types.SettingNameStorageNetwork, types.SettingNameStorageNetworkForRWXVolumeEnabled}
+	for _, item := range targetSettings {
+		if applied, err := c.ds.GetSettingApplied(item); err != nil || !applied {
+			return applied, err
+		}
+	}
+	return true, nil
+}
+
+func (c *ShareManagerController) canCleanupService(ns, shareManagerName string) (bool, error) {
```
c3y1huang, line 1072:
```suggestion
func (c *ShareManagerController) canCleanupService(shareManagerName string) (bool, error) {
```

derekbit (dismissed):
LGTM

c3y1huang (changes requested):

controller/share_manager_controller.go
Vicente-Cheng: Nice catch! updated

c3y1huang (approved):
LGTM, thank you for the fix.

c3y1huang:
@mergify backport v1.7.x

PR #3079, opened 2024-08-22
fix: update engine naming regex to allow non digit enging

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/9267

Before v1.4.x, users might have engine name like "XXXX-e-12avc" or "XXXX-e-abcd1"
We change to use "-0", "-1" starting from v1.4.4, v1.5.2
https://github.com/longhorn/longhorn-manager/pull/2180

The previous regex didn't allow "XXXX-e-12avc" which  has non digit ending
We should fix it


ChanYiLin:
Will update the vendor after following PR is merged.
- longhorn-instance-manager: https://github.com/longhorn/longhorn-instance-manager/pull/631

derekbit:
> Will update the vendor after following PR is merged.
> 
> * longhorn-instance-manager: [fix: update engine naming regex to allow non digit enging longhorn-instance-manager#631](https://github.com/longhorn/longhorn-instance-manager/pull/631)

Merged.

derekbit:
@mergify backport v1.7.x

derekbit (approved):
LGTM

PR #3100, opened 2024-08-26
fix(uninstall): ignore not found error when deleting CR

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/9303

We should ignore not found error when deleting resources

### DIFF
--- a/controller/uninstall_controller.go
+++ b/controller/uninstall_controller.go
@@ -648,16 +648,26 @@ func (c *UninstallController) deleteVolumes(vols map[string]*longhorn.Volume) (e
 		timeout := metav1.NewTime(time.Now().Add(-gracePeriod))
 		if vol.DeletionTimestamp == nil {
 			if err = c.ds.DeleteVolume(vol.Name); err != nil {
-				err = errors.Wrap(err, "failed to mark for deletion")
-				return
+				if datastore.ErrorIsNotFound(err) {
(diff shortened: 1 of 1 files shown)


controller/uninstall_controller.go
```diff
@@ -647,13 +647,13 @@ func (c *UninstallController) deleteVolumes(vols map[string]*longhorn.Volume) (e
 
 		timeout := metav1.NewTime(time.Now().Add(-gracePeriod))
 		if vol.DeletionTimestamp == nil {
-			if err = c.ds.DeleteVolume(vol.Name); err != nil {
+			if err = c.ds.DeleteVolume(vol.Name); err != nil && !datastore.ErrorIsNotFound(err) {
 				err = errors.Wrap(err, "failed to mark for deletion")
 				return
 			}
 			log.Info("Marked for deletion")
```
derekbit, line 654: Suggest 
```
			if err = c.ds.DeleteVolume(vol.Name); err != nil {
					log.Info("Volume is not found")
				} else {
					err = errors.Wrap(err, "failed to mark for deletion")
					return
				}
			} else {
				log.Info("Marked for deletion")
			}
```

ChanYiLin: fixed

derekbit (approved):
LGTM

derekbit:
@mergify backport v1.7.x

PR #3103, opened 2024-08-26
fix(manager): add detailed message when deleting pdb is not allowed (backport #3045)

w13915984028:
Issue # https://github.com/longhorn/longhorn/issues/9183
btw: https://github.com/longhorn/longhorn/issues/5910

Test log of manual attached volume when drain node
```
time="2024-08-05T18:48:14Z" level=info msg="Node instance-manager-68cd2514dd3f6d59b95cbd865d5b08f7 is marked unschedulable but removing harv2 PDB is blocked: some volumes are still attached " func="controller.(*InstanceManagerController).syncInstanceManagerPDB" file="instance_manager_controller.go:730" controller=longhorn-instance-manager node=harv2
```

This is a manual backport to replace https://github.com/longhorn/longhorn-manager/pull/3098, which has conflicts and I have no rights to push to/on-top-of it.

### DIFF
--- a/controller/instance_manager_controller.go
+++ b/controller/instance_manager_controller.go
@@ -940,14 +940,14 @@ func formatInstanceMessage(im *longhorn.InstanceManager) string {
 		}
 	}
 
-	if len(im.Status.Instances) > 0 {
+	if len(im.Status.Instances) > 0 { // nolint: staticcheck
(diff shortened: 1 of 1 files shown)


derekbit (approved):
LGTM

PR #3107, opened 2024-08-27
fix(uninstall): don't return error when error is not found

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/9303

should not return error when it is not found


derekbit (approved):
LGTM

derekbit:
@mergify backport v1.7.x

PR #3109, opened 2024-08-27
fix(datastore): skip disabled nodes in ValidateV2DataEngineEnabled

tserong:
Issue https://github.com/longhorn/longhorn/issues/9319

So we can enable the v2 data engine without needing to allocate hugepages on disabled nodes


derekbit (approved):
LGTM

derekbit:
@mergify backport v1.7.x

PR #3111, opened 2024-08-27
feat(v2 volume/configurable cpu mask): support configurable CPU cores

derekbit:

Issue longhorn/longhorn#8835


innobead:
Will this part of dynamic scheduler?

derekbit:
> Will this be part of dynamic scheduler?

No. This is only for the configurable dedicated cpu cores.

datastore/longhorn.go
```diff
@@ -15,9 +17,10 @@ import (
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/labels"
-	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/validation"
 
+	k8sRuntime "k8s.io/apimachinery/pkg/runtime"
```
shuo-wu, line 22: NIT: IIRC, we won't use camel naming for import alias

types/setting.go
```diff
@@ -1385,6 +1388,16 @@ var (
 		},
 	}
 
+	SettingDefinitionV2DataEngineCPUMask = SettingDefinition{
```
shuo-wu, line 1391: Can there be a conflict between this and `V2DataEngineGuaranteedInstanceManagerCPU`? For example, `V2DataEngineGuaranteedInstanceManagerCPU` requests 1 vCPU while the mask tries to use 2 or more specific CPU cores.
Then, the next question is, do we need to add a webhook validation for this?

derekbit: Yeah, I'm also thinking about the two settings.
`V2DataEngineGuaranteedInstanceManagerCPU` is better to be larger than `V2DataEngineCPUMask`, and we can validate it in the validating webhook. WDYT?

controller/instance_manager_controller.go
```diff
@@ -498,6 +499,27 @@ func (imc *InstanceManagerController) syncInstanceStatus(im *longhorn.InstanceMa
 	return nil
 }
 
+func (imc *InstanceManagerController) syncDateEngineCPUMask(im *longhorn.InstanceManager) (bool, error) {
```
DamiaSan, line 502: If I understand well the purpose of this functions, it doesn't actually _sync_ masks, it only make some checks. If this is right, maybe the name of the function is a bit confusing.

types/setting.go
```diff
@@ -1369,7 +1372,7 @@ var (
 
 	SettingDefinitionV2DataEngineGuaranteedInstanceManagerCPU = SettingDefinition{
 		DisplayName: "Guaranteed Instance Manager CPU for V2 Data Engine",
-		Description: "Number of millicpus on each node to be reserved for each instance manager pod when the V2 Data Engine is enabled. By default, the Storage Performance Development Kit (SPDK) target daemon within each instance manager pod uses 1 CPU core. Configuring a minimum CPU usage value is essential for maintaining engine and replica stability, especially during periods of high node workload. \n\n" +
+		Description: "Number of millicpus on each node to be reserved for each instance manager pod when the V2 Data Engine is enabled. By default, the Storage Performance Development Kit (SPDK) target daemon within each instance manager pod uses 1 or multiple CPU cores. Configuring a minimum CPU usage value is essential for maintaining engine and replica stability, especially during periods of high node workload. \n\n" +
```
DamiaSan, line 1375: _By default, the Storage Performance Development Kit (SPDK) target daemon within each instance manager pod uses 1 or multiple CPU cores_. How can the default have multiple values?

derekbit: Removed "By default" which is for the old description. Thank you.

controller/instance_manager_controller.go
derekbit: Renamed it to `isDateEngineCPUMaskApplied`. Thank you.

DamiaSan (approved):

derekbit:
@shuo-wu Could you review it before merging it if you have time? Thank you.

derekbit:
cc @c3y1huang

datastore/longhorn.go
```diff
@@ -494,6 +504,49 @@ func (s *DataStore) ValidateV2DataEngineEnabled(dataEngineEnabled bool) (ims []*
...
+	maskValue, err := strconv.ParseUint(value[2:], 16, 64) // skip 0x prefix
+	if err != nil {
+		return fmt.Errorf("failed to parse CPU mask: %s", value)
+	}
+
+	// Validate the mask value is not larger than the number of available CPUs
+	numCPUs := runtime.NumCPU()
+	maxCPUMaskValue := (1 << numCPUs) - 1
+	if maskValue > uint64(maxCPUMaskValue) {
+		return fmt.Errorf("CPU mask exceeds the maximum allowed value for the current system: %s", value)
```
innobead, line 523: nit: add the max value info in the error message.

types/setting.go
```diff
@@ -1387,6 +1390,16 @@ var (
 		},
 	}
 
+	SettingDefinitionV2DataEngineCPUMask = SettingDefinition{
+		DisplayName: "CPU Mask for V2 Data Engine",
+		Description: "The setting is used to specify which CPU cores the Storage Performance Development Kit (SPDK) target daemon within each instance manager pod should run on. The mask is a case-insensitive hexadecimal string, which may optionally include the '0x' prefix. For example, 0x1, 0x1f, etc. \n\n",
+		Category:    SettingCategoryDangerZone,
```
innobead, line 1396: This may need to be updated when reviewed by a doc in longhorn/longhorn repo.

innobead (approved):
LGTM

datastore/longhorn.go

innobead (approved):
LGTM. Merge this until the description of the setting is updated after the longhorn/longhorn PR is merged.

DamiaSan (approved):

innobead (approved):

PR #3112, opened 2024-08-27
fix: don't log NodeReady is true and false in same reconcile loop

ejweber:

longhorn/longhorn#7738

Key change:

- A defer statement and a `nodeReady` conditional variable now ensure we don't record ready (i.e. log and emit an event) during the check for a manager pod just to later record not ready during the Kubernetes node status check.

Additional changes:

- A nil pointer exception used to be possible in this controller, though the severity and potential to hit it was extremely low. (It probably required a longhorn-manager pod running on a node that was actively being removed from the cluster. In this situation, a crash is probably fine.) The exception should no longer be possible.
- The logic for setting `Ready` and `Schedulable` is now broken out into a separate function.

### DIFF
--- a/controller/node_controller.go
+++ b/controller/node_controller.go
@@ -420,6 +420,8 @@ func (nc *NodeController) syncNode(key string) (err error) {
 		return err
 	}
 
+	node.Status.Region, node.Status.Zone = types.GetRegionAndZone(kubeNode.Labels)
+
 	if nc.controllerID != node.Name {
 		return nil
 	}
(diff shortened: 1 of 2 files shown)


james-munson (approved):
This has bugged me in RWX HA testing, too.  Glad to see it addressed.  LGTM.

ejweber:
It looks like the CI failure is legitimate. I will investigate.

ejweber:
> (I assume that when kubelet is restored, or the node restarts, it does return to Ready.)

I does, but it is good to add this to the test plan above. I will do that.

james-munson (approved):

controller/node_controller.go
```diff
@@ -1954,3 +1859,117 @@ func isNodeOrDisksEvictionRequested(node *longhorn.Node) bool {
...
+				node.Status.Conditions = types.SetConditionAndRecord(node.Status.Conditions,
+					longhorn.NodeConditionTypeReady, longhorn.ConditionStatusFalse,
+					string(longhorn.NodeConditionReasonKubernetesNodePressure),
+					fmt.Sprintf("Kubernetes node %v has pressure: %v, %v", node.Name, con.Reason, con.Message),
+					nc.eventRecorder, node, corev1.EventTypeWarning)
+			}
+		}
+	}
+
+	DisableSchedulingOnCordonedNode, err :=
```
derekbit, line 1941: nit:
```suggestion
	disableSchedulingOnCordonedNode, err :=
```

controller/node_controller.go
```diff
@@ -1954,3 +1859,117 @@ func isNodeOrDisksEvictionRequested(node *longhorn.Node) bool {
...
+							string(longhorn.NodeConditionReasonManagerPodDown),
+							fmt.Sprintf("Node %v is down: the manager pod %v is not running", node.Name, pod.Name),
+							nc.eventRecorder, node, corev1.EventTypeWarning)
+					}
+					break
+				}
+			}
+			break
+		}
+	}
```
derekbit, line 1896: nit: abstract the codes to `checkNodeManagerPodStatus` function

controller/node_controller.go
```diff
@@ -1954,3 +1859,117 @@ func isNodeOrDisksEvictionRequested(node *longhorn.Node) bool {
...
+			if con.Status == corev1.ConditionTrue {
+				nodeReady = false
+				node.Status.Conditions = types.SetConditionAndRecord(node.Status.Conditions,
+					longhorn.NodeConditionTypeReady, longhorn.ConditionStatusFalse,
+					string(longhorn.NodeConditionReasonKubernetesNodePressure),
+					fmt.Sprintf("Kubernetes node %v has pressure: %v, %v", node.Name, con.Reason, con.Message),
+					nc.eventRecorder, node, corev1.EventTypeWarning)
+			}
+		}
+	}
```
derekbit, line 1939: nit: abstract the codes

derekbit (dismissed):
LGTM

controller/node_controller.go
ejweber: Please check the latest commit to see if it meets your request.

controller/node_controller.go
ejweber: Please check the latest commit to see if it meets your request.

derekbit (approved):
LGTM

ejweber:

PR #3114, opened 2024-08-28
fix(live upgrade): newly live upgraded engine doesn't respect the disableRevCounter setting

PhanLe1010:
We forget to set disableRevCounter for the newly live upgraded engine. As a result, the newly live upgraded engine always has disableRevCounter as `false` even if the volume has disableRevCounter as `true`.

This mismatch prevents the volume from successfully rebuilding a new replica.

longhorn/longhorn#9331


PhanLe1010:
@mergify backport v1.7.x v1.6.x v1.5.x

derekbit (approved):
LGTM

PR #3118, opened 2024-08-29
fix(system-backup/restore): DR volume

c3y1huang:

Issue longhorn/longhorn#9330

A DR volume doesn't need to have its volume data backed up and should retain and use the `FromBackup` field in the system backup.

`None`

`None`

### DIFF
--- a/controller/system_backup_controller.go
+++ b/controller/system_backup_controller.go
@@ -731,6 +731,13 @@ func (c *SystemBackupController) backupVolumesAlways(systemBackup *longhorn.Syst
 
 	volumeBackups := make(map[string]*longhorn.Backup, len(volumes))
 	for _, volume := range volumes {
+		// Don't need to create volume data backup for DR volumes since it will
+		// be restored from the source volume's backup.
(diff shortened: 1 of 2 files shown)


controller/system_backup_controller.go
```diff
@@ -731,6 +731,11 @@ func (c *SystemBackupController) backupVolumesAlways(systemBackup *longhorn.Syst
 
 	volumeBackups := make(map[string]*longhorn.Backup, len(volumes))
 	for _, volume := range volumes {
+		if volume.Status.IsStandby {
+			c.logger.Infof("Skip backup for standby volume %v", volume.Name)
+			continue
+		}
```
derekbit, line 737: Does it mean user need to create standby volumes by themselves after restoring the system backup?

c3y1huang: No, the system backup will skip creating a backup for the DR volume. But system restore will still create the DR volume with the URL in the `FromBackup`.

derekbit (approved):
LGTM

c3y1huang:

PR #3122, opened 2024-08-30
fix(backingimage): set default min number of copies when upgrading

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/9352

When upgrading Longhorn to v1.7.x
The BackingImage will have minNumberOfCopies = 0
Need to set the default value in upgrade path.


c3y1huang (approved):

ChanYiLin:
@mergify backport v1.7.x

PR #3127, opened 2024-08-30
fix: attempt to delete processes from instance-managers in unknown state

ejweber:

longhorn/longhorn#6552

See https://github.com/longhorn/longhorn/issues/6552#issuecomment-2319177183 for context. It may be possible to delete engine and replica processes from an instance-manager even when the instance-manager's state is unknown. Doing so prevents the processes from becoming orphans if the instance-manager eventually recovers from the unknown state.

This PR is looking good in some local testing, but I want to put it through a few more paces before review.

### DIFF
--- a/engineapi/instance_manager.go
+++ b/engineapi/instance_manager.go
@@ -92,11 +92,22 @@ func CheckInstanceManagerProxySupport(im *longhorn.InstanceManager) error {
 	return nil
 }
 
-// NewInstanceManagerClient creates a new instance manager client
-func NewInstanceManagerClient(im *longhorn.InstanceManager) (*InstanceManagerClient, error) {
-	// Do not check the major version here. Since IM cannot get the major version without using this client to call VersionGet().
-	if im.Status.CurrentState != longhorn.InstanceManagerStateRunning || im.Status.IP == "" {
-		return nil, fmt.Errorf("invalid Instance Manager %v, state: %v, IP: %v", im.Name, im.Status.CurrentState, im.Status.IP)
+// NewInstanceManagerClient creates a new instance manager client. Usually, we only want to attempt to communicate with
+// an instance manager in state running. However, sometimes it makes sense to make a best effort attempt to communicate
+// with an instance manager in state unknown. It should be safe to do so, since Kubernetes should not reassign the pod's
+// IP address until it is at least terminating, which puts the instance manager in state error. However, there is an
+// increased chance of failure.
+func NewInstanceManagerClient(im *longhorn.InstanceManager, allowUnknown bool) (*InstanceManagerClient, error) {
+	// Do not check the major version here since IM cannot get the major version without using this client to call
+	// VersionGet().
+
+	if im.Status.IP == "" {
+		return nil, fmt.Errorf("invalid instance manager %v, state %v, IP %v", im.Name, im.Status.CurrentState, im.Status.IP)
+	}
+	if im.Status.CurrentState == longhorn.InstanceManagerStateUnknown && allowUnknown {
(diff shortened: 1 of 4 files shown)


derekbit:
CI failures:
```
controller/engine_controller.go:2430:16: S1039: unnecessary use of fmt.Sprintf (gosimple)
		return true, fmt.Sprintf("the RWX volume is delinquent")
		             ^
controller/replica_controller.go:910:16: S1039: unnecessary use of fmt.Sprintf (gosimple)
		return true, fmt.Sprintf("the RWX volume is delinquent")
```

ejweber:
To test (in contrast with https://github.com/longhorn/longhorn/issues/6552#issuecomment-2319177183):

1. Deploy Longhorn with these changes.
2. Create and attach a volume that has the same number of replicas as nodes (e.g. with the UI).
3. SSH to the node the volume is attached to.
4. Stop the kubelet.
5. Wait ~30s for Kubernetes to notice and for the instance-manager state to transition to unknown.
6. Detach the volume.
7. Verify that no related engine or replica processes exist in any instance-manager CR.

james-munson (dismissed):
Looks good.  The re-factor makes the logic clearer, too.

ejweber:
@james-munson raised a concern about the effect this might have on RWX fast failover. I will test it a bit before merging (and modify the test plan so QA also tests it eventually).

ejweber:
@james-munson, I ran case 1 from https://github.com/longhorn/longhorn/issues/6205#issuecomment-2255821468 a few times. The results were:

- ShareManager up in 10 seconds, workload pods writing in 52 seconds and 57 seconds respectively.
- ShareManager up in 9 seconds, workload pods writing in 53 and 58 seconds respectively.
- ShareManager up in 7 seconds, workload pods writing in 52 seconds and 52 seconds respectively.

I think RWX fast failover is still working as expected.

engineapi/instance_manager.go
```diff
@@ -92,11 +92,24 @@ func CheckInstanceManagerProxySupport(im *longhorn.InstanceManager) error {
...
+
+	if im.Status.IP == "" {
+		return nil, fmt.Errorf("invalid instance manager %v, state %v, IP %v", im.Name, im.Status.CurrentState, im.Status.IP)
+	}
+	if im.Status.CurrentState == longhorn.InstanceManagerStateRunning {
+		// Nothing to do.
+	} else if im.Status.CurrentState == longhorn.InstanceManagerStateUnknown && allowUnknown {
+		logrus.Warnf("Communicating with instance manager %v, state %v, IP %v", im.Name, im.Status.CurrentState, im.Status.IP)
+	} else {
+		return nil, fmt.Errorf("invalid instance manager %v, state %v, IP %v", im.Name, im.Status.CurrentState, im.Status.IP)
```
PhanLe1010, line 112:
```suggestion
	if im.Status.CurrentState == longhorn.InstanceManagerStateUnknown && allowUnknown {
		logrus.Warnf("Communicating with instance manager %v, state %v, IP %v", im.Name, im.Status.CurrentState, im.Status.IP)
	} else if im.Status.CurrentState != longhorn.InstanceManagerStateRunning {
		return nil, fmt.Errorf("invalid instance manager %v, state %v, IP %v", im.Name, im.Status.CurrentState, im.Status.IP)
```

PhanLe1010 (commented):
LGTM. Leave a styling comment but feel free to ignore it if you prefer the current implementation

engineapi/instance_manager.go
ejweber: Makes sense to me. Thanks for the suggestion. Fixed in an amend while rebasing.

PhanLe1010 (approved):

derekbit:
@mergify backport v1.7.x

PR #3133, opened 2024-09-02
fix: The datalocality of engine v2 volumes is always set to disable by the mutator.

hookak:

https://github.com/longhorn/longhorn/issues/9280

when a Longhorn engine v2 volume is created, the `dataLocality` field is always set to `disable` by the volume mutator in the webhook. To allow the use of different `dataLocality` settings for engine v2 volumes, this behavior should be removed.


derekbit:
@hookak 
Nice catch! Thanks for your contribution.
We just implemented the online replica rebuilding feature in v1.7.0 and haven't verified the data locality feature. It is time to verify and support it in v1.8.0.

BTW, are you using v2 volumes in your environment?

hookak:
Thanks, the data locality feature is important for our purposes.

We are looking for high-performance block storage, and the v2 volume is one of the options we're considering. We are testing its functionality and performance. The most important thing is whether we can operate this system reliably at a production level, but we haven't found a good example of that yet.

We are still not very familiar with Longhorn, especially the v2 volumes, so if there are any considerations we should be aware of, it would be great if you could let us know. @derekbit

hookak:
Hello, @derekbit 

I have also removed the logic from the volume validator(https://github.com/longhorn/longhorn-manager/blob/master/webhook/resources/volume/validator.go#L269-L273)

innobead:
@hookak thanks for the contribution!

PR #3140, opened 2024-09-05
fix(encrypt): close encrypted volume if it is opened

mantissahz:

Issue # longhorn/longhorn#9385

In the normal process of attaching a volume via CSI, the encrypted volume should be closed or inactivated before Longhorn attempts to open it.

Full regression test:
~https://ci.longhorn.io/job/private/job/longhorn-tests-regression/7529~
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/7538/

### DIFF
--- a/csi/crypto/crypto.go
+++ b/csi/crypto/crypto.go
@@ -145,6 +145,16 @@ func ResizeEncryptoDevice(volume, passphrase string) error {
 	return err
 }
 
+// IsDeviceMappedToNullPath determines if encrypted device is already open at a null path. The command 'cryptsetup status [crypted_device]' show "device:  (null)"
+func IsDeviceMappedToNullPath(device string) (bool, error) {
+	devPath, mappedFile, err := DeviceEncryptionStatus(device)
+	if err != nil {
+		return false, err
+	}
+
+	return mappedFile != "" && strings.Compare(devPath, "(null)") == 0, nil
+}
+
 // IsDeviceOpen determines if encrypted device is already open.
 func IsDeviceOpen(device string) (bool, error) {
 	_, mappedFile, err := DeviceEncryptionStatus(device)
--- a/csi/node_server.go
+++ b/csi/node_server.go
@@ -487,6 +487,18 @@ func (ns *NodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
 		cryptoDevice := crypto.VolumeMapper(volumeID)
 		log.Infof("Volume %s requires crypto device %s", volumeID, cryptoDevice)
 
+		// check if the crypto device is open at the null path.
+		// this will happen if the crypto device is not closed properly and a new attaching request is made on the same node.
+		// reference issue: https://github.com/longhorn/longhorn/issues/9385
+		if mappedToNullPath, err := crypto.IsDeviceMappedToNullPath(cryptoDevice); err != nil {
+			return nil, status.Errorf(codes.Internal, "failed to check if the crypto device %s for volume %s is mapped to the null path: %v", cryptoDevice, volumeID, err.Error())
+		} else if mappedToNullPath {
+			log.Warnf("Closing active crypto device %s for volume %s since the volume is not closed properly before", cryptoDevice, volumeID)
+			if err := crypto.CloseVolume(volumeID); err != nil {
+				return nil, status.Errorf(codes.Internal, "failed to close active crypto device %s for volume %s: %v ", cryptoDevice, volumeID, err.Error())
+			}
+		}
+
 		if err := crypto.OpenVolume(volumeID, devicePath, passphrase); err != nil {
 			return nil, status.Error(codes.Internal, err.Error())
 		}


csi/crypto/crypto.go
```diff
@@ -97,8 +97,10 @@ func EncryptVolume(devicePath, passphrase string, cryptoParams *EncryptParams) e
 // OpenVolume opens volume so that it can be used by the client.
 func OpenVolume(volume, devicePath, passphrase string) error {
 	if isOpen, _ := IsDeviceOpen(VolumeMapper(volume)); isOpen {
-		logrus.Infof("Device %s is already opened at %s", devicePath, VolumeMapper(volume))
-		return nil
+		logrus.Infof("Device %s is already opened at %s, and close it first", devicePath, VolumeMapper(volume))
+		if err := CloseVolume(volume); err != nil {
+			return err
+		}
```
derekbit, line 103: Why do we need to close the volume here?

mantissahz: In a normal process of attaching a volume via CSI, the volume on the worker node should be inactivated. (New deploying workload with PVC or scale the workload down and then scale the workload up)

In an abnormal process, the [NodeUnStageVolume](https://github.com/longhorn/longhorn-manager/blob/master/csi/node_server.go#L553) is unreachable and the encrypted volume is not closed, and then the new attaching request is coming.

mantissahz: Or we can check if the encrypted volume is opened and then close it here.
https://github.com/longhorn/longhorn-manager/blob/master/csi/node_server.go#L126-L131

derekbit: I'm confused. The device is opened based on line 99. Isn't the volume open? Is it necessary to close it?

mantissahz: Line 99 `IsDeviceOpen` only checks whether the device is opened.
The opening operation is the line 113 `nsexec.LuksOpen(...)`

c3y1huang: The original implementation continues staging the volume when the crypto device is already opened. This means the code will not reach `nsexec.LuksOpen()`, and the CSI then attempts to reuse the already opened crypto device. So I guess the question is, what is blocking the CSI from reusing the crypto device?

mantissahz: Modified to check if the device path is null from the `cryptsetup status [device]` information before opening the device. If so, close the device before opening it.

csi/crypto/crypto.go
```diff
@@ -145,6 +145,18 @@ func ResizeEncryptoDevice(volume, passphrase string) error {
...
+// IsDeviceOpenAtNullPath determines if encrypted device is already open at unknown device path. The command 'cryptsetup status [device]' show "device:  (null)"
+func IsDeviceOpenAtNullPath(device string) (bool, error) {
+	devPath, mappedFile, err := DeviceEncryptionStatus(device)
+	if err != nil {
+		return false, err
+	}
+	if mappedFile != "" && strings.Contains(devPath, "null") {
+		return true, nil
+	}
+	return false, nil
```
c3y1huang, line 157:
```suggestion
	return mappedFile != "" && strings.Contains(devPath, "null"), nil
```

csi/node_server.go
```diff
@@ -487,6 +487,15 @@ func (ns *NodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
 		cryptoDevice := crypto.VolumeMapper(volumeID)
 		log.Infof("Volume %s requires crypto device %s", volumeID, cryptoDevice)
 
+		if isOpenAtNullPath, err := crypto.IsDeviceOpenAtNullPath(cryptoDevice); err != nil {
+			return nil, status.Error(codes.Internal, err.Error())
+		} else if isOpenAtNullPath {
+			log.Infof("Volume %s closing active crypto device %s", volumeID, cryptoDevice)
```
c3y1huang, line 493:
```suggestion
			log.Infof("Closing active crypto device %s for volume %s", cryptoDevice, volumeID)
```

csi/node_server.go
```diff
@@ -487,6 +487,15 @@ func (ns *NodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
 		cryptoDevice := crypto.VolumeMapper(volumeID)
 		log.Infof("Volume %s requires crypto device %s", volumeID, cryptoDevice)
 
+		if isOpenAtNullPath, err := crypto.IsDeviceOpenAtNullPath(cryptoDevice); err != nil {
```
c3y1huang, line 490: Nit: consider adding a brief description the scenario and the possible cause , or a link to the issue.

c3y1huang (changes requested):

csi/crypto/crypto.go
mantissahz: Modified. Thanks.

csi/node_server.go
mantissahz: Modified. Thanks.

csi/node_server.go
mantissahz: Added. Thanks.

c3y1huang (approved):
LGTM.

csi/crypto/crypto.go
```diff
@@ -145,6 +145,16 @@ func ResizeEncryptoDevice(volume, passphrase string) error {
...
 }
 
+// IsDeviceOpenAtNullPath determines if encrypted device is already open at unknown device path. The command 'cryptsetup status [device]' show "device:  (null)"
+func IsDeviceOpenAtNullPath(device string) (bool, error) {
+	devPath, mappedFile, err := DeviceEncryptionStatus(device)
+	if err != nil {
+		return false, err
+	}
+
+	return mappedFile != "" && strings.Contains(devPath, "null"), nil
```
derekbit, line 155: `strings.Contains(devPath, "null")` might be a bit dangerous if a valid path contains `null` substring. 
What is `devPath` when a device is with null path?

mantissahz: It will show `(null)` from `device:  (null)`.
```shell
/dev/mapper/pvc-a35cf6d0-2fbf-4f7a-bc0c-90520fcd3581 is active and is in use.
  type:    n/a
  cipher:  aes-xts-plain64
  keysize: 256 bits
  key location: keyring
  device:  (null)
  sector size:  512
  offset:  32768 sectors
  size:    10452992 sectors
  mode:    read/write
```

We can check if it is `(null)` for the full path instead of the `null` sub-string comparison.

derekbit: Sounds good.

csi/crypto/crypto.go
```diff
@@ -145,6 +145,16 @@ func ResizeEncryptoDevice(volume, passphrase string) error {
 	return err
 }
 
+// IsDeviceOpenAtNullPath determines if encrypted device is already open at unknown device path. The command 'cryptsetup status [device]' show "device:  (null)"
+func IsDeviceOpenAtNullPath(device string) (bool, error) {
```
derekbit, line 149:
```suggestion
func IsDeviceMappedToNullPath(device string) (bool, error) {
```

In addition, the function only checks `null`. What does `unknown device path` mean in the function description?

csi/crypto/crypto.go
mantissahz: Modified. Thanks

csi/crypto/crypto.go
mantissahz: Modified the method name and description from `unknown device path` to `null device path`. Thanks

csi/node_server.go
```diff
@@ -487,6 +487,18 @@ func (ns *NodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
 		cryptoDevice := crypto.VolumeMapper(volumeID)
 		log.Infof("Volume %s requires crypto device %s", volumeID, cryptoDevice)
 
+		// check if the crypto device is open at the null path.
+		// this will happen if the crypto device is not closed properly and a new attaching request is made on the same node.
+		// reference issue: https://github.com/longhorn/longhorn/issues/9385
+		if isOpenAtNullPath, err := crypto.IsDeviceMappedToNullPath(cryptoDevice); err != nil {
```
derekbit, line 493:
```suggestion
		if mappedToNullPath, err := crypto.IsDeviceMappedToNullPath(cryptoDevice); err != nil {
```

csi/node_server.go
```diff
@@ -487,6 +487,18 @@ func (ns *NodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
...
+			return nil, status.Error(codes.Internal, err.Error())
+		} else if isOpenAtNullPath {
+			log.Infof("Closing active crypto device %s for volume %s", cryptoDevice, volumeID)
```
derekbit, line 496:
```suggestion
			log.Wanf("Closing active crypto device %s for volume %s since .... 
```

csi/node_server.go
derekbit: Is it possible that the device is not found? If YES, need to handle it?

csi/node_server.go
```diff
@@ -487,6 +487,18 @@ func (ns *NodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
...
+			if err := crypto.CloseVolume(volumeID); err != nil {
```
derekbit, line 497: Is it possible that the device is not found? If YES, need to handle it?

csi/node_server.go
mantissahz: Modified. Thanks, and `mappedToNullPath` will be `false` and `err` will be `nil` if the device is not found because `mappedFile` is empty returned from the method `DeviceEncryptionStatus`

csi/node_server.go
mantissahz: Modified. Thanks

csi/node_server.go
mantissahz: Modified. Thanks, and `mappedToNullPath` will be `false` and `err` will be `nil` if the device is not found because `mappedFile` is empty returned from the method `DeviceEncryptionStatus`.

csi/node_server.go
```diff
@@ -487,6 +487,18 @@ func (ns *NodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
...
 
+		// check if the crypto device is open at the null path.
+		// this will happen if the crypto device is not closed properly and a new attaching request is made on the same node.
+		// reference issue: https://github.com/longhorn/longhorn/issues/9385
+		if mappedToNullPath, err := crypto.IsDeviceMappedToNullPath(cryptoDevice); err != nil {
+			return nil, status.Error(codes.Internal, err.Error())
+		} else if mappedToNullPath {
+			log.Warnf("Closing active crypto device %s for volume %s since the volume is not closed properly before", cryptoDevice, volumeID)
+			if err := crypto.CloseVolume(volumeID); err != nil {
+				return nil, status.Error(codes.Internal, err.Error())
```
derekbit, line 498: The error message will help troubleshooting.

```suggestion
				return nil, status.Error(codes.Internal, "failed to close active ...... since ...")
```

csi/node_server.go
```diff
@@ -487,6 +487,18 @@ func (ns *NodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
 		cryptoDevice := crypto.VolumeMapper(volumeID)
 		log.Infof("Volume %s requires crypto device %s", volumeID, cryptoDevice)
 
+		// check if the crypto device is open at the null path.
+		// this will happen if the crypto device is not closed properly and a new attaching request is made on the same node.
+		// reference issue: https://github.com/longhorn/longhorn/issues/9385
+		if mappedToNullPath, err := crypto.IsDeviceMappedToNullPath(cryptoDevice); err != nil {
+			return nil, status.Error(codes.Internal, err.Error())
```
derekbit, line 494: The error message will help troubleshooting.
```suggestion
			return nil, status.Error(codes.Internal, "failed to .... since ...")
```

derekbit (commented):
In general, LGTM

1. Update the error messages
2. QA test cases need to include
    -  encrypted volume attachment and detachment
    - validate the case when path is `(null)`

csi/node_server.go
mantissahz: Modified. Thanks

csi/node_server.go
mantissahz: Modified. Thanks

derekbit (approved):
LGTM

c3y1huang (approved):

mantissahz:
@Mergifyio backport v1.7.x v1.6.x

PR #3156, opened 2024-09-11
fix(crd): update conversion webhook port to 9501

derekbit:

Issue Longhorn/longhorn#9438

longhorn/longhorn crds.yaml is already updated.
Correct the port number to avoid any confusion.

### DIFF
--- a/k8s/crds.yaml
+++ b/k8s/crds.yaml
@@ -1357,7 +1357,7 @@ spec:
           name: longhorn-conversion-webhook
           namespace: longhorn-system
           path: /v1/webhook/conversion
-          port: 9443
+          port: 9501
       conversionReviewVersions:
       - v1beta2
       - v1beta1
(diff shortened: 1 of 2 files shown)


PhanLe1010 (approved):
LGTM

Just want to double check that we don't have an issue even though the port here is wrong because the lCRDs in the onghorn/longhorn repo still have correct port. Most of the time, we use the yaml in longhorn/longhorn instead. But this fix is good to make them consistent. Am I missing anything?

derekbit:
> LGTM
> 
> Just want to double check that we don't have an issue even though the port here is wrong because the lCRDs in the onghorn/longhorn repo still have correct port. Most of the time, we use the yaml in longhorn/longhorn instead. But this fix is good to make them consistent. Am I missing anything?

You're right. This is only for consistency, we've updated longhorn/longhorn chart and manifest before.

PR #3157, opened 2024-09-11
fix(crd): update conversion webhook port to 9501

derekbit:

Issue Longhorn/longhorn#9438

### DIFF
--- a/k8s/patches/crd/webhook/volume.yaml
+++ b/k8s/patches/crd/webhook/volume.yaml
@@ -12,5 +12,5 @@ spec:
           namespace: longhorn-system
           name: longhorn-conversion-webhook
           path: /v1/webhook/conversion
-          port: 9443
+          port: 9501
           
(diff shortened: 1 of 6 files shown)


k8s/patches/crd/webhook/volume.yaml
```diff
@@ -12,5 +12,5 @@ spec:
           namespace: longhorn-system
           name: longhorn-conversion-webhook
           path: /v1/webhook/conversion
-          port: 9443
+          port: 9501
```
innobead, line 15: Why do we change this? Is this missed to be backported? Can we have the description for this PR?

ChanYiLin: I think this is the backport of this PR: https://github.com/longhorn/longhorn-manager/pull/2948

ChanYiLin (approved):

derekbit:
Should be this one https://github.com/longhorn/longhorn/issues/5590.
We changed the port from 9443 to 9501 in that ticket, but didn't update longhorn-manager crd patch files.

shuo-wu:
Just curious, when do we update [this network policy ref doc](https://longhorn.io/docs/1.7.1/references/networking/#longhorn-manager)? The doc already shows longhorn-conversion-webhook port 9501 several minor versions ago.

PR #3165, opened 2024-09-16
fix: failed to cancel volume expansion after engine is already expanded successfully

derekbit:

Issue Longhorn/longhorn#9466

When the volume has already been successfully expanded, if the cancellation is issued after completing the volume expansion, and the spec.size and status.currentSize of the volume and engine are as follows:

- v.spec.size: 21474836480
- e.spec.size: 21474836480
- e.status.currentSize: 22548578304

The check
```
if e.Spec.VolumeSize == v.Spec.Size {
    return nil
}
```
prevents v.Status.ExpansionRequired from becoming true, which hinders the cancellation of the expansion.

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -2026,7 +2026,7 @@ func (c *VolumeController) reconcileVolumeSize(v *longhorn.Volume, e *longhorn.E
 	if e == nil || rs == nil {
 		return nil
 	}
-	if e.Spec.VolumeSize == v.Spec.Size {
+	if e.Spec.VolumeSize == v.Spec.Size && e.Status.CurrentSize == v.Spec.Size {
 		return nil
 	}
 


controller/volume_controller.go
```diff
@@ -2026,7 +2026,7 @@ func (c *VolumeController) reconcileVolumeSize(v *longhorn.Volume, e *longhorn.E
 	if e == nil || rs == nil {
 		return nil
 	}
-	if e.Spec.VolumeSize == v.Spec.Size {
+	if e.Spec.VolumeSize == v.Spec.Size && e.Status.CurrentSize == v.Spec.Size {
```
c3y1huang, line 2029: This fix looks ok to me as far as resolving the issue. It seems this fix will allow `v.Status.ExpansionRequired == true` to be set again, which enables the user to cancel expansion, correcting the volume size to the post-expansion size  [v.Spec.Size = engine.Status.CurrentSize](https://github.com/longhorn/longhorn-manager/blob/b393140a1bea3c3d198295ac5d143a26232cffc7/manager/volume.go#L585).
 
However, I am a bit confused about how `v.Status.ExpansionRequired`  ended up being set to `False` during cancellation, as described in the issue.

> Expanding volume through UI and then Cancelling expansion through UI results in volume stuck in expansion error:

IIRC, we only set `v.Status.ExpansionRequired = False` when `e.Status.CurrentSize == v.Spec.Size`. 

But from the https://github.com/longhorn/longhorn/issues/9466#issuecomment-2352006846, there seems to be an inconsistency between the sizes in spec and status:
> - v.spec.size: 21474836480
> - e.spec.size: 21474836480
> - e.status.currentSize: 22548578304

c3y1huang: I see, so this means the cancel volume expansion is ineffective once `engineClientProxy.VolumeExpand()` is invoked.

PR #3168, opened 2024-09-16
fix(expansion): add an error message to guide users in resolving the stuck volume expansion cancellation

derekbit:

Issue Longhorn/longhorn#9466

## Summary by CodeRabbit

- **Bug Fixes**
	- Enhanced error handling for volume expansion cancellation, providing clearer feedback to users when a cancellation is not possible.

- **Improvements**
	- Updated error messages for better clarity regarding volume state and cancellation issues.

### DIFF
--- a/manager/volume.go
+++ b/manager/volume.go
@@ -555,12 +555,6 @@ func (m *VolumeManager) CancelExpansion(volumeName string) (v *longhorn.Volume, 
 	if err != nil {
 		return nil, err
 	}
-	if !v.Status.ExpansionRequired {
-		return nil, fmt.Errorf("volume expansion is not started")
-	}
-	if v.Status.IsStandby {
-		return nil, fmt.Errorf("canceling expansion for standby volume is not supported")
-	}
 
 	var engine *longhorn.Engine
 	es, err := m.ds.ListVolumeEngines(v.Name)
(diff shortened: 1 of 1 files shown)


manager/volume.go
```diff
@@ -574,6 +568,17 @@ func (m *VolumeManager) CancelExpansion(volumeName string) (v *longhorn.Volume,
 		engine = e
 	}
 
+	if !v.Status.ExpansionRequired {
+		if v.Spec.Size == engine.Spec.VolumeSize && v.Spec.Size != engine.Status.CurrentSize {
```
shuo-wu, line 572: Is this possible?

The intermediate state that causes the bug is:
1. The expansion is already started and finished in the engine process level, which means `v.Spec.Size == engine.Spec.VolumeSize`. 
2. Longhorn has not updated engine/volume CR status. In other words, `v.Spec.Size != engine.Status.CurrentSize`  and **`v.Status.ExpansionRequired == true`**. (I am not sure if `engine.Status.IsExpanding` can be false here)

To be honest, it is hard to capture this state then reject the requests in API layer.

shuo-wu: The error message itself is reasonable. But my question is, how can `v.Status.ExpansionRequired` become false (or can the new condition be satisfied) before `VolumeManager.CancelExpansion` execution complete? 
The only place of adding the hint is probably [here](https://github.com/longhorn/longhorn-manager/blob/86a4416bf8c2a7def43b2dccf7d0433647099537/controller/engine_controller.go#L1222-L1224). And currently, Longhorn has no way to figure out if there is an cancellation before this error is trigger, the error hint can be only like:
```
the expected size %v of engine %v should not be smaller than the current size %v. It can be caused by a cancellation request after the expansion is complete, please re-do the expansion to resolve this error.
```

innobead: This is a general problem for canceling any async operations relying on the resource status which is regularly updated but not in realtime. 

Is it possible to get the runtime info from the engine directly when calling the cancellation to decide whether the volume is in expansion or not?

shuo-wu: > get the runtime info from the engine directly when calling the cancellation to decide whether the volume is in expansion

It is not an atomic operation hence there can be a race anyway. For now, I think adding warnings/notifications are enough.

PR #3177, opened 2024-09-25
fix: aggregate "*-dev-*" versions into single "dev" version

PhanLe1010:
This will prevent the upgrade responder from having too many tags which can be a performance issue

longhorn/longhorn#9522


PhanLe1010:
@mergify backport v1.7.x, v1.6.x, v1.5.x

innobead (approved):

PhanLe1010:

PR #3191, opened 2024-10-07
fix: disable clone for v2 volume 

shuo-wu:

longhorn/longhorn##9106


shuo-wu:
@mergify backport to v1.7.x

PR #3198, opened 2024-10-08
feat: allow v2 volume to attach on the an old and running instance-manager pod

derekbit:

Issue Longhorn/longhorn#9383


derekbit:
cc @c3y1huang @ChanYiLin @shuo-wu @innobead ready to go.

c3y1huang (approved):

PR #3211, opened 2024-10-14
fix: add events when pre- or post-upgrade check fails

james-munson:

https://github.com/longhorn/longhorn/issues/9569

Add an event when upgrade pre- or post-check job completes, either with error or success message.

## Summary by CodeRabbit

- **New Features**
	- Enhanced event broadcasting and recording capabilities during the upgrade process.
	- Improved logging for pre-upgrade checks and outcomes.

- **Bug Fixes**
	- Simplified event handling by streamlining the event broadcaster functionality.

- **Documentation**
	- Updated comments for better clarity on upgrade path checks.

### DIFF
--- a/app/post_upgrade.go
+++ b/app/post_upgrade.go
@@ -10,16 +10,24 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/urfave/cli"
 
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/tools/record"
 
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/labels"
 
+	"github.com/longhorn/longhorn-manager/constant"
 	"github.com/longhorn/longhorn-manager/types"
+
+	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
 )
 
 const (
+	PostUpgradeEventer = "longhorn-post-upgrade"
+
 	RetryCounts   = 360
 	RetryInterval = 5 * time.Second
 )
@@ -33,8 +41,10 @@ func PostUpgradeCmd() cli.Command {
 				Usage: "Specify path to kube config (optional)",
 			},
 			cli.StringFlag{
-				Name:   FlagNamespace,
-				EnvVar: types.EnvPodNamespace,
+				Name:     FlagNamespace,
+				EnvVar:   types.EnvPodNamespace,
+				Required: true,
+				Usage:    "Specify Longhorn namespace",
 			},
 		},
 		Action: func(c *cli.Context) {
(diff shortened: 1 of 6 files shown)


app/post_upgrade.go
```diff
@@ -50,34 +60,62 @@ func PostUpgradeCmd() cli.Command {
...
 		return errors.Wrap(err, "failed to get k8s client")
 	}
 
-	return newPostUpgrader(namespace, kubeClient).Run()
+	err = newPostUpgrader(namespace, kubeClient, eventRecorder).Run()
+	if err != nil {
+		// hang around so logs cam be collected.
+		// TODO - make this a --ttl argument.
+		time.Sleep(1 * time.Hour)
+	}
```
james-munson, line 91: Note that with the event captured, this is not as necessary, but it might be useful.  I'm not sure what to use for the time to wait.  The pre-upgrade job itself has `spec.activeDeadlineSeconds: 900` so the pod will be killed after 15 minutes anyway, and perhaps that is a reasonable value to use.

constant/events.go
```diff
@@ -62,7 +62,10 @@ const (
 	EventReasonReady    = "Ready"
 	EventReasonUploaded = "Uploaded"
 
-	EventReasonUpgrade = "Upgrade"
+	EventReasonUpgrade                = "Upgrade"
+	EventReasonFailedUpgradePreCheck  = "FailedUpgradePreCheck"
+	EventReasonFailedUpgradePostCheck = "FailedUpgradePostCheck"
+	EventReasonPassedUpgradeCheck     = "UpgradeCheckPassed"
```
derekbit, line 68: How about "`UpgradePreCheckFailed` and `UpgradeCheckPassed`" or "`FailedUpgradePreCheck` and `PassedUpgradeCheck`"

app/post_upgrade.go
derekbit: If the event is emitted, does it need to sleep for minutes?

james-munson: Not really.  The event will last for an hour, so if a support bundle is collected in that time, the event should be there.  
It would be the way to accomplish the goal in https://github.com/longhorn/longhorn/issues/9448.
I can certainly take it out if preferred.

constant/events.go
james-munson: I'll make it more consistent.

app/post_upgrade.go
```diff
@@ -50,34 +60,57 @@ func PostUpgradeCmd() cli.Command {
...
+		if err != nil {
+			u.eventRecorder.Event(&corev1.ObjectReference{Namespace: u.namespace, Name: PostUpgradeEventer},
+				corev1.EventTypeWarning, constant.EventReasonFailedUpgradePostCheck, err.Error())
+		} else {
+			u.eventRecorder.Event(&corev1.ObjectReference{Namespace: u.namespace, Name: PostUpgradeEventer},
+				corev1.EventTypeNormal, constant.EventReasonPassedUpgradeCheck, "post-upgrade check passed")
+		}
+	}()
+
+	if err = u.waitManagerUpgradeComplete(); err != nil {
```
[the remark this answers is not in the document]
james-munson: This is nonsense.  If modified as suggested, it won't compile because the anonymous defer function references `err`, which is undeclared.
If the declaration is kept, but the other change is made, that `err` will shadow the first one, and the defer will event a success, even though `waitManagerUpgradeComplete` failed.

app/post_upgrade.go
james-munson: I removed it.

james-munson: I put it back.  Without it, sometimes the panic from the "fatal" error means the event does not get created.

derekbit: > I put it back. Without it, sometimes the panic from the "fatal" error means the event does not get created.

Can you elaborate more on the statement?

app/pre_upgrade.go
```diff
@@ -47,17 +59,62 @@ func preUpgrade(c *cli.Context) error {
...
 		return errors.Wrap(err, "failed to get clientset")
 	}
 
-	if err := upgradeutil.CheckUpgradePath(namespace, lhClient, nil, true); err != nil {
+	err = newPreUpgrader(namespace, lhClient, eventRecorder).Run()
+	if err != nil {
+		logrus.Warnf("Done with Run() ... err is %v", err)
+		// Hang around so that event can happen before fatal exit.
+		time.Sleep(5 * time.Minute)
+	}
```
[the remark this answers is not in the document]
PhanLe1010: I think the suggesting from @coderrabbitai is a good one. Good work! @derekbit

app/post_upgrade.go
PhanLe1010: > I put it back. Without it, sometimes the panic from the "fatal" error means the event does not get created.

Looks like the AI suggestion can solve this one. WDYT @james-munson ?

https://github.com/longhorn/longhorn-manager/pull/3211/files#r1803865276

james-munson: The events are queued by Event/Eventf, but not necessarily propagated to the sink before the Event() call returns.  They can be lost when the next thing that happens is an os.Exit as part of the log.Fatal.
However, the AI suggestion is a good one.  I tested it, and it looks like eventBroadcaster.Shutdown() forces a flush of the queued events, so they show up even without a sleep to delay the exit.  I have pushed up the change.

app/pre_upgrade.go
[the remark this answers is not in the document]
james-munson: Agreed.  I have taken it.

PhanLe1010 (approved):
LGTM

derekbit (approved):
LGTM

derekbit:
@mergify backport v1.6.x v1.7.x

PR #3226, opened 2024-10-19
fix: don't build longhorn-manager twice before packaging

derekbit:

Issue Longhorn/longhorn#8744


innobead (approved):

derekbit:

PR #3230, opened 2024-10-22
chore(dockerfile): reduce packaging time

derekbit:

Issue Longhorn/longhorn#8744

The packages aren't updated frequently, so the content of the image layer usually remains unchanged.
To avoid unnecessary rebuilding of this layer, copy the longhorn-manager binaries after installing the packages.


innobead (approved):

derekbit:

PR #3236, opened 2024-10-24
chore: add support for ReadWriteOncePod access mode

shikanime:
Signed-off-by: William Phetsinorath <william.phetsinorath@shikanime.studio>

### DIFF
--- a/csi/controller_server.go
+++ b/csi/controller_server.go
@@ -69,6 +69,8 @@ func NewControllerServer(apiClient *longhornclient.RancherClient, nodeID string)
 		accessModes: getVolumeCapabilityAccessModes(
 			[]csi.VolumeCapability_AccessMode_Mode{
 				csi.VolumeCapability_AccessMode_SINGLE_NODE_WRITER,
+				csi.VolumeCapability_AccessMode_SINGLE_NODE_SINGLE_WRITER,
+				csi.VolumeCapability_AccessMode_SINGLE_NODE_MULTI_WRITER,
 				csi.VolumeCapability_AccessMode_MULTI_NODE_MULTI_WRITER,
 			}),
 		log: logrus.StandardLogger().WithField("component", "csi-controller-server"),
@@ -218,6 +220,10 @@ func (cs *ControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 	for _, cap := range volumeCaps {
 		if requiresSharedAccess(nil, cap) {
 			volumeParameters["share"] = "true"
+			break
+		}
+		if requireExclusiveAccess(nil, cap) {
+			volumeParameters["exclusive"] = "true"
 			break
 		}
 	}
@@ -449,6 +455,13 @@ func (cs *ControllerServer) ControllerPublishVolume(ctx context.Context, req *cs
 
 	if requiresSharedAccess(volume, volumeCapability) {
 		volume, err = cs.updateVolumeAccessMode(volume, longhorn.AccessModeReadWriteMany)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	if requireExclusiveAccess(volume, volumeCapability) {
+		volume, err = cs.updateVolumeAccessMode(volume, longhorn.AccessModeReadWriteOncePod)
 		if err != nil {
 			return nil, err
 		}
--- a/csi/util.go
+++ b/csi/util.go
@@ -115,9 +115,24 @@ func getVolumeOptions(volumeID string, volOptions map[string]string) (*longhornc
 
 		if isShared {
 			vol.AccessMode = string(longhorn.AccessModeReadWriteMany)
-		} else {
-			vol.AccessMode = string(longhorn.AccessModeReadWriteOnce)
 		}
+	}
+
+	if exclusive, ok := volOptions["exclusive"]; ok {
+		isExclusive, err := strconv.ParseBool(exclusive)
+		if err != nil {
+			return nil, errors.Wrap(err, "invalid parameter exclusive")
+		}
+		if isExclusive && vol.AccessMode == string(longhorn.AccessModeReadWriteMany) {
+			return nil, errors.New("cannot set both share and exclusive to true")
+		}
+		if isExclusive {
+			vol.AccessMode = string(longhorn.AccessModeReadWriteOncePod)
+		}
+	}
+
+	if vol.AccessMode == "" {
+		vol.AccessMode = string(longhorn.AccessModeReadWriteOnce)
 	}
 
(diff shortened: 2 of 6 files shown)


derekbit:
@shikanime Awesome! Thanks for your contribution.
@james-munson Please help review the PR if you have time. Thank you.

james-munson (commented):
LGTM.  I would ask that the PR get a corresponding Github issue, and that it have a note about how the feature was tested.  Also, that will allow it to be tracked into internal QA feature test as well.

derekbit:
@james-munson Could you create Feature ticket in longhorn/longhorn for this PR and help follow up it? Thank you.

james-munson:
Associated tracking issue is https://github.com/longhorn/longhorn/issues/9727

csi/controller_server.go
```diff
@@ -69,6 +69,8 @@ func NewControllerServer(apiClient *longhornclient.RancherClient, nodeID string)
 		accessModes: getVolumeCapabilityAccessModes(
 			[]csi.VolumeCapability_AccessMode_Mode{
 				csi.VolumeCapability_AccessMode_SINGLE_NODE_WRITER,
+				csi.VolumeCapability_AccessMode_SINGLE_NODE_SINGLE_WRITER,
+				csi.VolumeCapability_AccessMode_SINGLE_NODE_MULTI_WRITER,
```
innobead, line 73: Do we want to support `VolumeCapability_AccessMode_SINGLE_NODE_MULTI_WRITER`?? 

> 	// Can be published as read/write at multiple workloads on a
	// single node simultaneously. SHOULD be used instead of
	// SINGLE_NODE_WRITER for COs using the experimental
	// SINGLE_NODE_MULTI_WRITER capability.
	VolumeCapability_AccessMode_SINGLE_NODE_MULTI_WRITER VolumeCapability_AccessMode_Mode = 7

https://github.com/kubernetes/enhancements/blob/master/keps/sig-storage/2485-read-write-once-pod-pv-access-mode/README.md#csi-specification-changes-volume-capabilities

`VolumeCapability_AccessMode_SINGLE_NODE_WRITER` might be sufficient in this case. Use `VolumeCapability_AccessMode_SINGLE_NODE_SINGLE_WRITER` until we support `VolumeCapability_AccessMode_SINGLE_NODE_MULTI_WRITER`.

cc @derekbit @james-munson

innobead:
Thanks, @shikanime contribution. The commit message can start with `feat:` as this is a new feature. 

For reviewers. besides reviewing, please help verify if this works as expected. The issue owner should be @james-munson .

csi/controller_server.go
shikanime: Technically, we already support it at https://github.com/longhorn/longhorn-manager/blob/1e95ce6aab3aabc4f8acc4c0e3dbde1f95023c3c/csi/util.go#L454-L469, but we haven't really cared since then.

shikanime: Technically, we already support it at https://github.com/longhorn/longhorn-manager/blob/235b1ae23f6ad4d26dc56063c95358145e937c09/csi/util.go#L439-L454, but we haven't really cared until now.

shikanime: Oh nvm, this isn't the same the same capability

innobead:
@shikanime will you continue working on this PR? Thanks for your contribution.

shikanime:
Yes for sure, afaik the implementation is right. I just need to fix the conflict as this PR is quite old.

derekbit:
Hello @shikanime 

CI failed. Could you fix it? Thanks.

```
# Warning: 'commonLabels' is deprecated. Please use 'labels' instead. Run 'kustomize edit fix' to update your Kustomization automatically.
Replacing labels and namespace in crds.yaml...
diff --git a/k8s/crds.yaml b/k8s/crds.yaml
index 5f87b4a..d1d66fc 100644
--- a/k8s/crds.yaml
+++ b/k8s/crds.yaml
@@ -3426,6 +3426,7 @@ spec:
               accessMode:
                 enum:
                 - rwo
+                - rwop
                 - rwx
                 type: string
               backingImage:
crds.yaml is not up to date. Please run 'make generate' and commit the changes.
time="2025-11-18T12:38:41Z" level=fatal msg="exit status 1"
```

csi/util.go
```diff
@@ -118,11 +118,26 @@ func getVolumeOptions(volumeID string, volOptions map[string]string) (*longhornc
...
 		}
 	}
 
+	if exclusive, ok := volOptions["exclusive"]; ok {
+		isExclusive, err := strconv.ParseBool(exclusive)
+		if err != nil {
+			return nil, errors.Wrap(err, "invalid parameter exclusive")
+		}
+		if isExclusive && vol.AccessMode == string(longhorn.AccessModeReadWriteMany) {
+			return nil, errors.New("cannot set both share and exclusive to true")
```
COLDTURNIP, line 130: Similar to migratable RWX volume (around [line 147](https://github.com/shikanime/longhorn-manager/blob/pr3236/csi/util.go#L147-L152)), we should just ignore the exclusive flag and log the issue here, rather than return an error to break the volume creation.

csi/controller_server.go
```diff
@@ -258,6 +259,10 @@ func (cs *ControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 			volumeParameters["share"] = "true"
 			break
 		}
+		if requireExclusiveAccess(nil, cap) {
+			volumeParameters["exclusive"] = "true"
```
COLDTURNIP: Please change to `strconv.FormatBool` (also at above line 259)

```
volumeParameters["exclusive"] = strconv.FormatBool(true)
```

csi/util.go
derekbit: Hmm, I prefer the current implementation, an explicit error if the flag is mistakenly set, because `exclusive` is for `ReadWriteOncePod` and not similar to migratable RWX volume.

COLDTURNIP: The purpose of `getVolumeOptions` is to create a Volume CR from the options. If we need to reject the volume creation for unreasonable options, a better place to validate the access-mode-related capability might be `CreateVolume` ([line 255-266](https://github.com/shikanime/longhorn-manager/blob/pr3236/csi/controller_server.go#L255-L266)). The options "shared" and "exclusive" are mutually conflicting.

derekbit: I think we can introduce  a `storageclass` validation in https://github.com/longhorn/longhorn-manager/tree/master/webhook/resources as well as validate the two values in volume validating logic. Then, we don't need a complicated check here.

COLDTURNIP (approved):

PR #3267, opened 2024-11-11
fix(resize): delegate RWX filesystem resize to share-manager

james-munson:

Issue https://github.com/longhorn/longhorn/issues/9736 aka https://github.com/longhorn/longhorn/issues/8118


csi/node_server.go
```diff
@@ -690,11 +695,66 @@ func (ns *NodeServer) NodeGetVolumeStats(ctx context.Context, req *csi.NodeGetVo
...
+	podName := types.GetShareManagerPodNameFromShareManagerName(sm.Name)
+	pod, err := kubeClient.CoreV1().Pods(lhNamespace).Get(context.TODO(), podName, metav1.GetOptions{})
+	if err != nil {
+		return errors.Wrap(err, "failed to get ShareManager pod")
+	}
+
+	client, err := engineapi.NewShareManagerClient(sm, pod)
+	if err != nil {
+		return errors.Wrapf(err, "failed to launch gRPC client for share manager before resizing volume %v", volumeName)
+	}
```
PhanLe1010, line 737: I think we can get IP of share manager service by accessing `volume.ShareEndpoint` without having to look up the share manager pod?

csi/deployment.go
```diff
@@ -418,6 +418,14 @@ func NewPluginDeployment(namespace, serviceAccount, nodeDriverRegistrarImage, li
...
 									Value: GetCSIEndpoint(),
 								},
+								{
+									Name: "POD_NAMESPACE",
+									ValueFrom: &corev1.EnvVarSource{
+										FieldRef: &corev1.ObjectFieldSelector{
+											FieldPath: "metadata.namespace",
+										},
+									},
+								},
```
PhanLe1010, line 428: Maybe we don't need to add namespace if we don't need to look up the share manager pod as mentioned below

PhanLe1010 (commented):
I have some concerns:
1. As we mentioned in the sync up, we should double check network policy for the new connection from Longhorn CSI plugin to share manager 
2. With this design, we can only finish resizing when the RWX PVC is attached, right? Because only when RWX PVC is attached and used by workload then we have share manager pod. What about offline RWX PVC expansion. Do you have any idea?

james-munson:
> I have some concerns:
> 
>     1. As we mentioned in the sync up, we should double check network policy for the new connection from Longhorn CSI plugin to share manager
> 
>     2. With this design, we can only finish resizing when the RWX PVC is attached, right? Because only when RWX PVC is attached and used by workload then we have share manager pod. What about offline RWX PVC expansion. Do you have any idea?

Checking on the first.  My test cluster has no network policies, except for a cattle-fleet-system one that allows all in its namespace.

For the second, offline expansion works like it formerly did.  The resize will happen when the share-manager pod is started and mounts the volume.  That's why the code for resize was already there, https://github.com/longhorn/longhorn-share-manager/blob/5b2d717e3ca35a727fbaa628af478af4da535fcf/pkg/server/share_manager.go#L143

PhanLe1010:
> Checking on the first. My test cluster has no network policies, except for a cattle-fleet-system one that allows all in its namespace.

Thank you

> For the second, offline expansion works like it formerly did. The resize will happen when the share-manager pod is started and mounts the volume. That's why the code for resize was already there, https://github.com/longhorn/longhorn-share-manager/blob/5b2d717e3ca35a727fbaa628af478af4da535fcf/pkg/server/share_manager.go#L143

So in the meantime, the PVC will stuck in pending expansion until the workload start? Is it same behavior as RWO PVC?
Could we add a test case for this scenario? Thank you

james-munson:
For the point about network policy, we agreed that it does not look to be necessary for this functionality.  Better to keep things simple.

csi/node_server.go
PhanLe1010: Yes, `volume.ShareEndpoint` is the IP of the service which should eventually forward to the pod

csi/node_server.go
```diff
@@ -16,18 +16,23 @@ import (
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/rest"
 	"k8s.io/mount-utils"
 
 	corev1 "k8s.io/api/core/v1"
+	clientset "k8s.io/client-go/kubernetes"
 	utilexec "k8s.io/utils/exec"
```
derekbit, line 25:
```suggestion
	"k8s.io/client-go/rest"
	"k8s.io/mount-utils"

	corev1 "k8s.io/api/core/v1"
	clientset "k8s.io/client-go/kubernetes"
	utilexec "k8s.io/utils/exec"
	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```

csi/node_server.go
```diff
@@ -729,6 +789,20 @@ func (ns *NodeServer) NodeExpandVolume(ctx context.Context, req *csi.NodeExpandV
...
 		return nil, status.Errorf(codes.FailedPrecondition, "invalid state %v for volume %v node expansion", volume.State, volumeID)
 	}
+
+	if requiresSharedAccess(volume, volumeCapability) && !volume.Migratable {
+		if volume.AccessMode != string(longhorn.AccessModeReadWriteMany) {
+			return nil, status.Errorf(codes.FailedPrecondition, "volume %s requires shared access but is not marked for shared use", volumeID)
+		}
+
+		if err := ns.NodeExpandSharedVolume(volumeID); err != nil {
+			log.WithError(err).Info("NodeExpandSharedVolume returned error.")
```
derekbit, line 799:
```suggestion
			log.WithError(err).Errorf("Failed to expand shared volume %v", volumeID)
```

csi/node_server.go
```diff
@@ -690,11 +695,66 @@ func (ns *NodeServer) NodeGetVolumeStats(ctx context.Context, req *csi.NodeGetVo
...
+	err = client.FilesystemResize()
+	if status.Code(err) == codes.Unimplemented {
+		// This is a downrev longhorn-share-manager image.  It will be necessary either to kill the share-manager pod
+		// and let it restart with the new image, or scale the workload down and back up to accomplish the same thing.
+		// It might be tempting to delete the pod here, but there will be one of these calls for each workload pod,
+		// and it would be messy to have multiple kill requests at once.  So the kill will need to be done by the user.
+		log.WithError(err).Warn("Share Manager image is down-rev, does not implement RPC FilesystemResize.  Share Manager pod must be restarted with current image.")
+	}
+
+	return err
```
derekbit, line 750:
```suggestion
	return errors.Wrapf(err, "Failed to .....
```

csi/node_server.go
james-munson: Doesn't seem to work.  Here is a bit of the share-manager pod
```
 podIP: 10.42.215.169
```
And the volume
```
    shareEndpoint: nfs://10.43.184.173/pvc-eb61e66c-e54d-473d-8e14-b98c5bccae2f
```
Extracting the IP address and using it as the gRPC target,
```
[longhorn-csi-plugin-rmz9g longhorn-csi-plugin] W1127 21:26:31.201924       1 logging.go:55] [core] [Channel #119 SubChannel #120]grpc: addrConn.createTransport failed to connect to {Addr: "10.43.184.173:9600", ServerName: "10.43.184.173:9600", }. Err: connection error: desc = "transport: Error while dialing: dial tcp 10.43.184.173:9600: i/o timeout" 
[longhorn-csi-plugin-rmz9g longhorn-csi-plugin] time="2024-11-27T21:26:31Z" level=info msg="NodeExpandSharedVolume returned error." func="csi.(*NodeServer).NodeExpandVolume" file="node_server.go:806" component=csi-node-server error="rpc error: code = Unavailable desc = connection error: desc = \"transport: Error while dialing: dial tcp 10.43.184.173:9600: i/o timeout\"" function=NodeExpandVolume 
[longhorn-csi-plugin-rmz9g longhorn-csi-plugin] time="2024-11-27T21:26:31Z" level=error msg="NodeExpandVolume: err: rpc error: code = Unavailable desc = connection error: desc = \"transport: Error while dialing: dial tcp 10.43.184.173:9600: i/o timeout\"" func=csi.logGRPC file="server.go:136" 
```

csi/node_server.go
```diff
@@ -690,111 +695,180 @@
...
+	pod, err := kubeClient.CoreV1().Pods(lhNamespace).Get(context.TODO(), podName, metav1.GetOptions{})
+	if err != nil {
+		return errors.Wrap(err, "failed to get ShareManager pod")
+	}
+
+	client, err := engineapi.NewShareManagerClient(sm, pod)
+	if err != nil {
+		return errors.Wrapf(err, "failed to launch gRPC client for share manager before resizing volume %v", volumeName)
+	}
+	defer client.Close()
```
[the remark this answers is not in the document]
james-munson: Not necessary.  At bottom, this calls grpc.NewClient, which always returns a nil client if there is a non-nil error.

csi/node_server.go
```diff
@@ -690,111 +695,180 @@
...
+
+	// Each node with a workload pod will send an RPC request.  The first will win, and the others are no-ops.
+	err = client.FilesystemResize()
+	if status.Code(err) == codes.Unimplemented {
+		// This is a downrev longhorn-share-manager image.  It will be necessary either to kill the share-manager pod
+		// and let it restart with the new image, or scale the workload down and back up to accomplish the same thing.
+		// It might be tempting to delete the pod here, but there will be one of these calls for each workload pod,
+		// and it would be messy to have multiple kill requests at once.  So the kill will need to be done by the user.
+		log.WithError(err).Warn("Share Manager image is down-rev, does not implement RPC FilesystemResize.  Share Manager pod must be restarted with current image.")
+	}
```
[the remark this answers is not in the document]
james-munson: There is no fallback.

csi/node_server.go
james-munson: Done.

csi/node_server.go
james-munson: Done.

csi/node_server.go
james-munson: It would be possible to patch the share-manager service that exports the NFS port to add the gRPC port as well.  But that would mean mixing data-plane and control plane.  It is preferred not to do that.  We will stick with the pod lookup, although it is requested to cache the kubeclient and lhclient.

james-munson:
Also, check that in the offline case, there are not repeated errors from failed attempts to look up the share-manager pod.

csi/node_server.go
james-munson: Done.

PhanLe1010 (approved):
LGTM

PR #3280, opened 2024-11-19
fix(logging): make RPC call to instance-manager to set log level

james-munson:

Issue https://github.com/longhorn/longhorn/issues/6702

Sync the log level with instance-managers.

### DIFF
--- a/controller/instance_manager_controller.go
+++ b/controller/instance_manager_controller.go
@@ -521,10 +521,6 @@ func (imc *InstanceManagerController) isDateEngineCPUMaskApplied(im *longhorn.In
 }
 
 func (imc *InstanceManagerController) syncLogSettingsToInstanceManagerPod(im *longhorn.InstanceManager) error {
-	if types.IsDataEngineV1(im.Spec.DataEngine) {
-		return nil
-	}
-
 	if im.Status.CurrentState != longhorn.InstanceManagerStateRunning {
 		return nil
 	}
@@ -536,6 +532,7 @@ func (imc *InstanceManagerController) syncLogSettingsToInstanceManagerPod(im *lo
 	defer client.Close()
 
 	settingNames := []types.SettingName{
+		types.SettingNameLogLevel,
 		types.SettingNameV2DataEngineLogLevel,
 		types.SettingNameV2DataEngineLogFlags,
 	}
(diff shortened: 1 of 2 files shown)


controller/instance_manager_controller.go
```diff
@@ -535,6 +531,21 @@ func (imc *InstanceManagerController) syncLogSettingsToInstanceManagerPod(im *lo
...
+			return err
+		}
+
+		err = client.LogSetLevel(longhorn.DataEngineTypeV1, "", setting.Value)
+		if err != nil {
+			return errors.Wrapf(err, "failed to set instance-manager log level from %v", settingName)
+		}
+
+		return nil
+	}
```
derekbit, line 547: Why can this setting only be applied to v1 data engine? v2 data engine should be applied as well.

james-munson: Ah, thank you.

controller/instance_manager_controller.go
```diff
@@ -547,15 +543,21 @@ func (imc *InstanceManagerController) syncLogSettingsToInstanceManagerPod(im *lo
...
 			}
 		case types.SettingNameV2DataEngineLogFlags:
-			err = client.LogSetFlags(longhorn.DataEngineTypeV2, "spdk_tgt", setting.Value)
-			if err != nil {
-				return errors.Wrapf(err, "failed to set log flags for %v", settingName)
+			if types.IsDataEngineV2(im.Spec.DataEngine) {
+				err = client.LogSetFlags(longhorn.DataEngineTypeV2, "spdk_tgt", setting.Value)
+				if err != nil {
+					return errors.Wrapf(err, "failed to set spdk_tgt log flags to %v, using setting %v", setting.Value, settingName)
+				}
```
PhanLe1010, line 560: Are we deprecating this setting?

controller/instance_manager_controller.go
```diff
@@ -547,15 +543,21 @@ func (imc *InstanceManagerController) syncLogSettingsToInstanceManagerPod(im *lo
...
 
 		switch settingName {
-		case types.SettingNameV2DataEngineLogLevel:
-			err = client.LogSetLevel(longhorn.DataEngineTypeV2, "spdk_tgt", setting.Value)
+		case types.SettingNameLogLevel:
+			if types.IsDataEngineV1(im.Spec.DataEngine) {
+				err = client.LogSetLevel(longhorn.DataEngineTypeV1, "", setting.Value)
+			} else {
+				err = client.LogSetLevel(longhorn.DataEngineTypeV2, "", setting.Value)
+			}
```
PhanLe1010, line 551: What about other components like UI, csi plugin, csi side car pod as mentioned at https://github.com/longhorn/longhorn/issues/6702#issuecomment-2445030144

controller/instance_manager_controller.go
james-munson: Yes, that's the conclusion of the discussion yesterday.  No need for granularity to set it separately for instance-manager and spdk_tgt.  That will need to be updated in the docs.

controller/instance_manager_controller.go
james-munson: That would be a separate PR, although less critical, since they don't log as heavily.  It shouldn't gate this change.

controller/instance_manager_controller.go
james-munson: Oh, wait.  We deprecated V2DataEngineLogLevel.  LogFlags needs to remain.  That is a separate feature of SPDK logging.

controller/instance_manager_controller.go
james-munson: Remedied and re-pushed.

controller/instance_manager_controller.go
```diff
@@ -536,7 +532,7 @@ func (imc *InstanceManagerController) syncLogSettingsToInstanceManagerPod(im *lo
 	defer client.Close()
 
 	settingNames := []types.SettingName{
-		types.SettingNameV2DataEngineLogLevel,
+		types.SettingNameLogLevel,
 		types.SettingNameV2DataEngineLogFlags,
```
derekbit, line 536: Rename `SettingNameV2DataEngineLogFlags` to `SettingNameLogFlags`. Make it generalized although v1 data engine doesn't need it.

derekbit:
@james-munson need to handle https://github.com/longhorn/longhorn/issues/6702#issuecomment-2499674384

james-munson:
The current PR handles it, using setting `log-level` to control instance-manager logrus logs, and `v2-data-engine-log-level` to control spdk_tgt logs.
Note that it requires the instance-manager changes in https://github.com/longhorn/longhorn-instance-manager/pull/726

controller/instance_manager_controller.go
```diff
@@ -547,15 +544,26 @@ func (imc *InstanceManagerController) syncLogSettingsToInstanceManagerPod(im *lo
 		}
 
 		switch settingName {
-		case types.SettingNameV2DataEngineLogLevel:
-			err = client.LogSetLevel(longhorn.DataEngineTypeV2, "spdk_tgt", setting.Value)
+		case types.SettingNameLogLevel:
+			// We use this to set the instance-manager log level, for either engine type.
+			err = client.LogSetLevel(longhorn.DataEngineTypeV1, "", setting.Value)
```
derekbit, line 549: This is quite confusing. Can we replace `longhorn.DataEngineTypeV1` with `""`?

james-munson: Yes, that should work.  `getDataEngine` in longhorn-instance-manager/pkg/client should translate `""` that way.

james-munson: Tested and re-pushed.

derekbit:
LGTM @james-munson ready to merge?

derekbit (approved):

james-munson:

PR #3293, opened 2024-11-27
fix: detached volume replicas not evicted

c3y1huang:

Issue longhorn/longhorn#9781, longhorn/longhorn#9857

We need to allow the volume eviction controller to handle detached volumes. This proposes another approach to address longhorn/longhorn#9781: proceed with volume attachment for replica eviction only when suitable node/disk candidates are available.


c3y1huang:
@coderabbitai review

derekbit (approved):
LGTM. Let @PhanLe1010 and @shuo-wu review as well.

derekbit:

PR #3294, opened 2024-11-29
feat: add automatic storage_reserved calculation for block type disks

jangseon-ryu:

Issue # https://github.com/longhorn/longhorn/issues/9871 https://github.com/longhorn/longhorn/discussions/9797

This PR addresses a bug where the Storage Reserved value is incorrectly set to 0 during automatic disk registration when using Engine v2 block type.

To retrieve the total capacity of the block device, the implementation uses unix.Syscall.
If there is a more efficient or reliable approach to gather block device capacity, please provide your suggestions or feedback.

### DIFF
--- a/controller/kubernetes_node_controller.go
+++ b/controller/kubernetes_node_controller.go
@@ -304,10 +304,6 @@ func (knc *KubernetesNodeController) syncDefaultDisks(node *longhorn.Node) (err 
 		if err != nil {
 			return err
 		}
-		storageReservedPercentageForDefaultDisk, err := knc.ds.GetSettingAsInt(types.SettingNameStorageReservedPercentageForDefaultDisk)
-		if err != nil {
-			return err
-		}
 		disks, err = types.CreateDefaultDisk(dataPath, storageReservedPercentageForDefaultDisk)
 		if err != nil {
 			return err
--- a/controller/kubernetes_node_controller.go
+++ b/controller/kubernetes_node_controller.go
@@ -293,7 +293,10 @@ func (knc *KubernetesNodeController) syncDefaultDisks(node *longhorn.Node) (err 
 		return nil
 	}
 	val = strings.ToLower(val)
-
+	storageReservedPercentageForDefaultDisk, err := knc.ds.GetSettingAsInt(types.SettingNameStorageReservedPercentageForDefaultDisk)
+	if err != nil {
+		return err
+	}
 	disks := map[string]longhorn.DiskSpec{} // nolint: ineffassign,staticcheck
 	switch val {
 	case types.NodeCreateDefaultDiskLabelValueTrue:
@@ -314,7 +317,7 @@ func (knc *KubernetesNodeController) syncDefaultDisks(node *longhorn.Node) (err 
 		if !ok {
 			return nil
 		}
-		disks, err = types.CreateDisksFromAnnotation(annotation)
+		disks, err = types.CreateDisksFromAnnotation(annotation, storageReservedPercentageForDefaultDisk)
 		if err != nil {
 			knc.logger.WithError(err).Warnf("Failed to create disk from annotation, invalid annotation %v: %v", types.KubeNodeDefaultDiskConfigAnnotationKey, val)
 			return nil
--- a/types/types.go
+++ b/types/types.go
@@ -14,6 +14,8 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
+	"golang.org/x/sys/unix"
+	"unsafe"
 
 	lhns "github.com/longhorn/go-common-libs/ns"
 
@@ -996,7 +998,7 @@ func LabelsToString(labels map[string]string) string {
 	return res
 }
 
-func CreateDisksFromAnnotation(annotation string) (map[string]longhorn.DiskSpec, error) {
+func CreateDisksFromAnnotation(annotation string, storageReservedPercentage int64) (map[string]longhorn.DiskSpec, error) {
 	validDisks := map[string]longhorn.DiskSpec{}
 	existDiskID := map[string]string{}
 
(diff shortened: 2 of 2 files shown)


controller/kubernetes_node_controller.go
```diff
@@ -272,67 +272,70 @@
...
~	val, ok := kubeNode.Labels[types.NodeCreateDefaultDiskLabelKey]
~	if !ok {
 		return nil
 	}
 	val = strings.ToLower(val)
-
+	storageReservedPercentageForDefaultDisk, err := knc.ds.GetSettingAsInt(types.SettingNameStorageReservedPercentageForDefaultDisk)
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
jangseon-ryu: Done.

types/types.go
```diff
@@ -952,61 +954,78 @@
...
+				if err != nil {
+					return nil, fmt.Errorf("failed to open block device at %s: %w", disk.Path, err)
+				}
+				var size uint64
+				_, _, errno := unix.Syscall(unix.SYS_IOCTL, file.Fd(), 0x80081272, uintptr(unsafe.Pointer(&size)))
+				file.Close()
+				if errno != 0 {
+					return nil, fmt.Errorf("failed to get block device size for %s: errno=%v", disk.Path, errno)
+				}
+				disk.StorageReserved = int64(size) * storageReservedPercentage / 100
```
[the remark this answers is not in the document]
jangseon-ryu: IoctlGetUint64 is not currently available in the golang.org/x/sys/unix package
As a result, I had to rely on unix.Syscall with the appropriate ioctl number to retrieve the block device size.
If IoctlGetUint64 becomes available in the future I would be happy to refactor the implementation

DamiaSan:
> To retrieve the total capacity of the block device, the implementation uses unix.Syscall.
If there is a more efficient or reliable approach to gather block device capacity, please provide your suggestions or feedback.

With v2, you should have these information in the `bdevMap`, like for example here: https://github.com/longhorn/longhorn-spdk-engine/blob/main/pkg/spdk/engine.go#L662

jangseon-ryu:
> > To retrieve the total capacity of the block device, the implementation uses unix.Syscall.
> > If there is a more efficient or reliable approach to gather block device capacity, please provide your suggestions or feedback.
> 
> With v2, you should have these information in the `bdevMap`, like for example here: https://github.com/longhorn/longhorn-spdk-engine/blob/main/pkg/spdk/engine.go#L662

Thank you for the review. 
This code runs during the initial disk registration phase, before the `lvstore` is configured. Therefore, I didn’t use the [getDiskStat](https://github.com/longhorn/longhorn-manager/blob/master/controller/monitor/disk_utils.go#L28C6-L28C17) function and focused on simply collecting the disk capacity at this stage.
Let me know if further adjustments are needed.

types/types.go
```diff
@@ -992,6 +994,23 @@ func CreateDisksFromAnnotation(annotation string) (map[string]longhorn.DiskSpec,
...
 		}
+		if disk.StorageReserved == 0 {
+			if disk.Type == longhorn.DiskTypeBlock {
+				file, err := os.Open(ReplicaHostPrefix + disk.Path)
+				if err != nil {
+					return nil, fmt.Errorf("failed to open block device at %s: %w", disk.Path, err)
+				}
+				defer file.Close()
+				var size uint64
+				_, _, errno := unix.Syscall(unix.SYS_IOCTL, file.Fd(), 0x80081272, uintptr(unsafe.Pointer(&size)))
```
derekbit, line 1005: Does the IOCTL command code (0x80081272) work for both SATA and NVMe SSDs?

jangseon-ryu: This has been tested in a SATA SSD environment only. I will verify whether it is supported on NVMe as well and will share the findings once confirmed

derekbit: Thanks @jangseon-ryu

derekbit (dismissed):
LGTM

c3y1huang (dismissed):

types/types.go
```diff
@@ -14,6 +14,8 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
+	"golang.org/x/sys/unix"
+	"unsafe"
```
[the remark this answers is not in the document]
jangseon-ryu: done

controller/kubernetes_node_controller.go
```diff
@@ -293,18 +293,17 @@ func (knc *KubernetesNodeController) syncDefaultDisks(node *longhorn.Node) (err
...
 	case types.NodeCreateDefaultDiskLabelValueTrue:
 		dataPath, err := knc.ds.GetSettingValueExisted(types.SettingNameDefaultDataPath)
 		if err != nil {
 			return err
 		}
-		storageReservedPercentageForDefaultDisk, err := knc.ds.GetSettingAsInt(types.SettingNameStorageReservedPercentageForDefaultDisk)
-		if err != nil {
-			return err
-		}
 		disks, err = types.CreateDefaultDisk(dataPath, storageReservedPercentageForDefaultDisk)
```
derekbit, line 307: `CreateDefaultDisk()` also needs to update the block-type disk'`StorageReserved`. Right?

types/types.go
```diff
@@ -1040,6 +1042,23 @@ func CreateDisksFromAnnotation(annotation string) (map[string]longhorn.DiskSpec,
...
+				}
+				defer file.Close()
+				var size uint64
+				_, _, errno := unix.Syscall(unix.SYS_IOCTL, file.Fd(), 0x80081272, uintptr(unsafe.Pointer(&size)))
+				if errno != 0 {
+					return nil, fmt.Errorf("failed to get block device size for %s: errno=%v", disk.Path, errno)
+				}
+				disk.StorageReserved = int64(size) * storageReservedPercentage / 100
+			} else {
+				disk.StorageReserved = diskStat.StorageMaximum * storageReservedPercentage / 100
```
derekbit, line 1059: Good catch. I just found Longhorn hadn't handled the `StorageReserved` for filesystem-type disk as well...
cc @shuo-wu @PhanLe1010 @c3y1huang @innobead

controller/kubernetes_node_controller.go
jangseon-ryu: CreateDefaultDisk() only supports filesystem type. If node.longhorn.io/create-default-disk is set to true, it uses the settings.default-data-path to create a default disk, which is exclusively filesystem-based. Therefore, StorageReserved for block-type disks is not required in this context.

derekbit: Doesn't CreateDefaultDisk() support filesystem type as well?
https://github.com/longhorn/longhorn-manager/blob/601b04bebbf77ad0e98a80e3d0b82b745b994fee/types/types.go#L1122-L1134

jangseon-ryu: Yes, that's correct.  It seems that adding StorageReserved configuration to CreateDefaultDisk is necessary. I will work on modifying it to support this functionality.

derekbit:
Hello @jangseon-ryu is the PR ready?

jangseon-ryu:
> Hello @jangseon-ryu is the PR ready?

Yes, it has been completed. https://github.com/longhorn/longhorn-manager/pull/3294/commits/e820258837641855ca479cefbcad3c0a07930253

derekbit (approved):
LGTM

c3y1huang (approved):

DamiaSan (approved):

derekbit:
@jangseon-ryu Thanks for your contribution! Great work!

PR #3306, opened 2024-12-05
feat: log the deletion of isntance manager pod

derekbit:

- log the deletion of isntance manager pod
- rename cleanupInstanceManager() to cleanupInstanceManagerPod()

Issue longhorn/longhorn#9886

### DIFF
--- a/controller/instance_manager_controller.go
+++ b/controller/instance_manager_controller.go
@@ -295,7 +295,8 @@ func (imc *InstanceManagerController) syncInstanceManager(key string) (err error
 	im, err := imc.ds.GetInstanceManager(name)
 	if err != nil {
 		if datastore.ErrorIsNotFound(err) {
-			return imc.cleanupInstanceManager(name)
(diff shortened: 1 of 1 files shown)


controller/instance_manager_controller.go
```diff
@@ -1248,7 +1254,7 @@ func (imc *InstanceManagerController) enqueueSettingChange(obj interface{}) {
 	imc.enqueueInstanceManagersForNode(imc.controllerID)
 }
 
-func (imc *InstanceManagerController) cleanupInstanceManager(imName string) error {
+func (imc *InstanceManagerController) cleanupInstanceManagerPod(imName string) error {
```
mantissahz, line 1257: This function deletes the InstanceManager objects directly, can we just remain the original name?

derekbit: Yes, removed. Thanks.

mantissahz (approved):
LGTM

PR #3315, opened 2024-12-06
feat: add kernel module check for data engine v2

jangseon-ryu:

Issue # https://github.com/longhorn/longhorn/issues/9915

Currently, the kernel module check only verifies dm_crypt. It would be beneficial to also check the kernel modules used by Longhorn Engine v2 (vfio_pci, uio_pci_generic, nvme_tcp) to ensure accurate node status reporting.

### DIFF
--- a/controller/node_controller.go
+++ b/controller/node_controller.go
@@ -59,6 +59,7 @@ const (
 
 var (
 	kernelModules     = map[string]string{"CONFIG_DM_CRYPT": "dm_crypt"}
+	kernelModulesV2   = map[string]string{"CONFIG_VFIO_PCI": "vfio_pci", "CONFIG_UIO_PCI_GENERIC": "uio_pci_generic", "CONFIG_NVME_TCP": "nvme_tcp"}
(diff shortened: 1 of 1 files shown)


c3y1huang (approved):

derekbit (approved):
LGTM

controller/node_controller.go
```diff
@@ -1137,7 +1138,24 @@ func (nc *NodeController) syncMultipathd(node *longhorn.Node, namespaces []lhtyp
 }
 
 func (nc *NodeController) checkKernelModulesLoaded(kubeNode *corev1.Node, node *longhorn.Node, namespaces []lhtypes.Namespace) {
-	notFoundModulesUsingkmod, err := checkModulesLoadedUsingkmod(kernelModules)
+	isV2DataEngine, err := nc.ds.GetSettingAsBool(types.SettingNameV2DataEngine)
+	if err != nil {
+		nc.logger.Errorf("Failed to fetch v2-data-engine setting: %v", err)
```
COLDTURNIP, line 1143: Would be better to attach node information in the logger, and using structured logging.

```suggestion
	log := getLoggerForNode(nc.logger, node)
	isV2DataEngine, err := nc.ds.GetSettingAsBool(types.SettingNameV2DataEngine)
	if err != nil {
		log.WithError(err).Error("Failed to fetch v2-data-engine setting")
```

PR #3334, opened 2024-12-09
feat(webhook): reject strict-local RWX volume creation

COLDTURNIP:

Issue longhorn/longhorn#6735

When the server node is down, the RWX NFS server is unable to create on another node because of the strict-local volume. Also, strict-local is for distributed application only. Distributed application means it has replication capability, so they will only RWO.

### DIFF
--- a/types/types.go
+++ b/types/types.go
@@ -824,6 +824,13 @@ func ValidateDataLocalityAndReplicaCount(mode longhorn.DataLocality, count int) 
 	return nil
 }
 
+func ValidateDataLocalityAndAccessMode(locality longhorn.DataLocality, migratable bool, mode longhorn.AccessMode) error {
+	if mode == longhorn.AccessModeReadWriteMany && !migratable && locality == longhorn.DataLocalityStrictLocal {
+		return fmt.Errorf("access mode %v (migratable: %v) is incompatible with data locality %v mode", mode, migratable, longhorn.DataLocalityStrictLocal)
+	}
+	return nil
+}
+
 func ValidateReplicaAutoBalance(option longhorn.ReplicaAutoBalance) error {
 	switch option {
 	case longhorn.ReplicaAutoBalanceIgnored,
--- a/webhook/resources/volume/validator.go
+++ b/webhook/resources/volume/validator.go
@@ -74,6 +74,10 @@ func (v *volumeValidator) Create(request *admission.Request, newObj runtime.Obje
 		return werror.NewInvalidError(err.Error(), "")
 	}
 
+	if err := types.ValidateDataLocalityAndAccessMode(volume.Spec.DataLocality, volume.Spec.Migratable, volume.Spec.AccessMode); err != nil {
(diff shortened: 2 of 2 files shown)


types/types.go
```diff
@@ -824,6 +824,13 @@ func ValidateDataLocalityAndReplicaCount(mode longhorn.DataLocality, count int)
 	return nil
 }
 
+func ValidateDataLocalityAndAccessMode(locality longhorn.DataLocality, mode longhorn.AccessMode) error {
+	if mode == longhorn.AccessModeReadWriteMany && locality == longhorn.DataLocalityStrictLocal {
+		return fmt.Errorf("access mode %v is incompatible with data locality %v mode", longhorn.AccessModeReadWriteMany, longhorn.DataLocalityStrictLocal)
```
innobead, line 829:
```suggestion
		return fmt.Errorf("access mode %v is incompatible with data locality %v mode", mode, longhorn.DataLocalityStrictLocal)
```

innobead (changes requested):
LGTM, just one minor comment.

innobead (approved):

types/types.go
```diff
@@ -824,6 +824,13 @@ func ValidateDataLocalityAndReplicaCount(mode longhorn.DataLocality, count int)
 	return nil
 }
 
+func ValidateDataLocalityAndAccessMode(locality longhorn.DataLocality, mode longhorn.AccessMode) error {
+	if mode == longhorn.AccessModeReadWriteMany && locality == longhorn.DataLocalityStrictLocal {
```
derekbit, line 828: Also need to check `v.spec.migratable`. We only check the real RWX volume whose volume.spec.migratable is `false`.
 
```suggestion
	if mode == longhorn.AccessModeReadWriteMany && migratable == false, locality == longhorn.DataLocalityStrictLocal {
```

innobead: Good catch! cc @PhanLe1010

derekbit (approved):

types/types.go
```diff
@@ -824,6 +824,13 @@ func ValidateDataLocalityAndReplicaCount(mode longhorn.DataLocality, count int)
 	return nil
 }
 
+func ValidateDataLocalityAndAccessMode(locality longhorn.DataLocality, migratable bool, mode longhorn.AccessMode) error {
+	if mode == longhorn.AccessModeReadWriteMany && !migratable && locality == longhorn.DataLocalityStrictLocal {
+		return fmt.Errorf("access mode %v is incompatible with data locality %v mode", mode, longhorn.DataLocalityStrictLocal)
```
innobead, line 829:
```suggestion
		return fmt.Errorf("access mode %v (migratable: %v) is incompatible with data locality %v mode", mode, migratable, longhorn.DataLocalityStrictLocal)
```

innobead (changes requested):
One minor suggestion for providing more context in log.

innobead (approved):

derekbit:
@mergify backport v1.6.x v1.7.x

PR #3339, opened 2024-12-10
fix: extend im pod liveness probe failure timeout

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/9851

Extend the liveness probe failure timeout.

Period: every 30 seconds
Timeout: every 29 seconds if command not response
FailureThreshold: 3 times failure in a row.

### DIFF
--- a/datastore/kubernetes.go
+++ b/datastore/kubernetes.go
@@ -43,6 +43,11 @@ const (
 	PodProbeTimeoutSeconds           = PodProbePeriodSeconds - 1
 	PodProbePeriodSeconds            = 5
 	PodLivenessProbeFailureThreshold = 3
+
+	IMPodProbeInitialDelay             = 3
+	IMPodProbeTimeoutSeconds           = IMPodProbePeriodSeconds - 1
+	IMPodProbePeriodSeconds            = 5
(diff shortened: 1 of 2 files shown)


datastore/kubernetes.go
```diff
@@ -43,6 +43,11 @@ const (
 	PodProbeTimeoutSeconds           = PodProbePeriodSeconds - 1
 	PodProbePeriodSeconds            = 5
 	PodLivenessProbeFailureThreshold = 3
+
+	IMPodProbeInitialDelay             = 3
+	IMPodProbeTimeoutSeconds           = IMPodProbePeriodSeconds - 1
+	IMPodProbePeriodSeconds            = 30
+	IMPodLivenessProbeFailureThreshold = 3
```
derekbit, line 50: How about
```
IMPodProbePeriodSeconds = 5
IMPodLivenessProbeFailureThreshold = 6
```
cc @WebberHuang1118

ChanYiLin: Sounds good to me
It seems better

derekbit (approved):
LGTM!

derekbit:

PR #3340, opened 2024-12-10
fix: clear fast-failover status when node goes not ready

james-munson:

Issue https://github.com/longhorn/longhorn/issues/9093

Do a fail-safe clearing of RWX fast-failover status still pending on a node that is NotReady.  If it hasn't already happened, it is too late for fast failover, so let the normal IsNodeDownOrDeleted logic take over.

### DIFF
--- a/controller/node_controller.go
+++ b/controller/node_controller.go
@@ -434,6 +434,13 @@ func (nc *NodeController) syncNode(key string) (err error) {
 		return err
 	}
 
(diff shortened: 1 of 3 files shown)


james-munson:
Address some rabbit review notes and rebase.

derekbit (approved):
LGTM.

derekbit:
@PhanLe1010 Could you take a look and see if there is any concern? Thank you.

PhanLe1010:
The idea sound good to me

PR #3351, opened 2024-12-17
fix(backup): modify backup volume and backup backing image name

mantissahz:

Issue # longhorn/longhorn#5411

### DIFF
--- a/types/types.go
+++ b/types/types.go
@@ -738,15 +738,11 @@ func GetShareManagerNameFromShareManagerPodName(podName string) string {
 }
 
 func GetBackupVolumeNameFromVolumeName(volumeName string) string {
-	return generateBackupResourceName(volumeName)
-}
-
-func generateBackupResourceName(resourceName string) string {
-	return util.GetStringChecksum(strings.TrimSpace(resourceName))[:util.RandomIDLength] + "-" + util.RandomID()
+	return volumeName + "-" + util.RandomID()
 }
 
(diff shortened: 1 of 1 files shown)


types/types.go
```diff
@@ -738,15 +741,11 @@ func GetShareManagerNameFromShareManagerPodName(podName string) string {
 }
 
 func GetBackupVolumeNameFromVolumeName(volumeName string) string {
-	return generateBackupResourceName(volumeName)
-}
-
-func generateBackupResourceName(resourceName string) string {
-	return util.GetStringChecksum(strings.TrimSpace(resourceName))[:util.RandomIDLength] + "-" + util.RandomID()
+	return volumeName + backupVolumeSuffix + "-" + util.RandomID()
```
derekbit, line 744: `backupVolumeSuffix` seems not really required?

types/types.go
```diff
@@ -738,15 +741,11 @@ func GetShareManagerNameFromShareManagerPodName(podName string) string {
...
 }
 
 func GetBackupBackingImageNameFromBIName(backingImageName string) string {
-	return generateBackupResourceName(backingImageName)
+	return backingImageName + backupBackingImageSuffix + "-" + util.RandomID()
```
derekbit, line 748: `backupBackingImageSuffix` seems not really required?

types/types.go
mantissahz: Removed. Thanks

derekbit (approved):
LGTM

PR #3367, opened 2024-12-19
fix: use backup backing image spec backing image name when getting backing image CR

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/10020


derekbit (approved):

derekbit:
@mergify backport v1.8.x

PR #3372, opened 2024-12-20
fix(backup): do not get backing image custom resource

mantissahz:

Issue #ref longhorn/longhorn#10026, longhorn/longhorn#6341

Do not get the backing image custom resource if the backup backing image is synchronized from the remote backup target because the backing image might not exist.

### DIFF
--- a/webhook/resources/backupbackingimage/validator.go
+++ b/webhook/resources/backupbackingimage/validator.go
@@ -45,11 +45,11 @@ func (bbi *backupBackingImageValidator) Create(request *admission.Request, newOb
 	backingImageName := backupBackingImage.Spec.BackingImage
 
 	backingImage, err := bbi.ds.GetBackingImageRO(backingImageName)
-	if err != nil {
+	if err != nil && !datastore.ErrorIsNotFound(err) {
 		return werror.NewInvalidError(fmt.Sprintf("failed to get the backing image %v for backup: %v", backingImageName, err), "")
 	}
 	// TODO: support backup for v2 data engine in the future
-	if types.IsDataEngineV2(backingImage.Spec.DataEngine) {
+	if backingImage != nil && types.IsDataEngineV2(backingImage.Spec.DataEngine) {
 		return werror.NewInvalidError(fmt.Sprintf("backing image %v uses v2 data engine which doesn't support backup operations", backingImageName), "")
 	}
 


webhook/resources/backupbackingimage/validator.go
```diff
@@ -44,13 +44,21 @@ func (bbi *backupBackingImageValidator) Create(request *admission.Request, newOb
...
-		return werror.NewInvalidError(fmt.Sprintf("backing image %v uses v2 data engine which doesn't support backup operations", backingImageName), "")
+	if !exists {
+		backingImage, err := bbi.ds.GetBackingImageRO(backingImageName)
+		if err != nil {
+			return werror.NewInvalidError(fmt.Sprintf("failed to get the backing image %v for backup: %v", backingImageName, err), "")
+		}
+		// TODO: support backup for v2 data engine in the future
+		if types.IsDataEngineV2(backingImage.Spec.DataEngine) {
+			return werror.NewInvalidError(fmt.Sprintf("backing image %v uses v2 data engine which doesn't support backup operations", backingImageName), "")
+		}
```
derekbit, line 61: I think we don't need to validate whether a backing image exists. We can allow users to create v2 backing image but always failed the backup (backupBacking.status.State=error, backupBackingImage.status.Error = "backing up v2 backing image is not supported). Then, we don't need the special label. WDYT?

cc @ChanYiLin

mantissahz: Removed the label, and ignored the error `ErrorIsNotFound`.

ChanYiLin: Sounds good to me

ChanYiLin: After a second thought, I think using webhook is still more simple.
The current change in this PR is also good enough
If the backing image exists in the cluster and it is v2 backing image
then the webhook blocks the backup creation.

If the backing image doesn't exist in the cluster, then the backupbackingimage is pulled from the remote, we allow it.

derekbit: There are two different behaviors for v2 backup backingimage. Is making them consistent better?

derekbit (approved):
LGTM

mantissahz:
@mergify backport v1.8.x

PR #3387, opened 2024-12-23
fix(backup): modify logic to list backup volume for recurring job and CSI snapshotter

mantissahz:

Issue # longhorn/longhorn#5411

### DIFF
--- a/csi/controller_server.go
+++ b/csi/controller_server.go
@@ -270,29 +270,26 @@ func (cs *ControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 
 func (cs *ControllerServer) getBackupVolume(volumeName string) (*longhornclient.BackupVolume, error) {
 	log := cs.log.WithFields(logrus.Fields{"function": "getBackupVolume"})
+	list, err := cs.apiClient.BackupVolume.List(&longhornclient.ListOpts{})
+	if err != nil {
+		return nil, err
+	}
+	if len(list.Data) == 0 {
+		return nil, nil
+	}
+
 	vol, err := cs.apiClient.Volume.ById(volumeName)
 	if err != nil {
 		return nil, errors.Wrapf(err, "getBackupVolume: fail to get source volume %v", volumeName)
 	}
 
-	list, err := cs.apiClient.BackupVolume.List(&longhornclient.ListOpts{
-		Filters: map[string]interface{}{
-			types.LonghornLabelBackupTarget: vol.BackupTargetName,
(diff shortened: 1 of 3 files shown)


csi/controller_server.go
```diff
@@ -270,29 +270,26 @@ func (cs *ControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
...
-		return nil, nil
+	for _, bv := range list.Data {
+		if bv.BackupTargetName == vol.BackupTargetName && bv.VolumeName == volumeName {
+			return &bv, nil
+		}
 	}
-
-	return &list.Data[0], nil
+	log.Debugf("cannot find backup volume with backup target %s and volume %s", vol.BackupTargetName, volumeName)
+	return nil, nil
```
derekbit, line 292: Why not `return nil, fmt.Errorf(....)"

mantissahz: As we discussed before, we return the error as nil when we can not find a backup volume as the backup is not completed. https://github.com/longhorn/longhorn-manager/blob/master/csi/controller_server.go#L1247-L1258

mantissahz:
>Can you run the regression test cases involving the changes and paste the links here?

Regression test cases:
- Recurring=Job: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/7933/
- CSI: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/7934/

csi/controller_server.go
derekbit: Oh, I got it. Can you add a comment here.

derekbit:
@mantissahz 
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/7933/ is failed

mantissahz:
>@mantissahz
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/7933/ is failed

@derekbit,
The PR ([fix(backup): recurring job test case failed](https://github.com/longhorn/longhorn-tests/pull/2220/commits/1e8f9bae174b68d1b835a08ae5772b0fbc248662)) tries to fix`test_recurring_job.py::test_recurring_jobs_allow_detached_volume` is merged, I will re-run it to check if it still failed.

https://ci.longhorn.io/job/private/job/longhorn-tests-regression/7940/

csi/controller_server.go
mantissahz: Added. Thanks

derekbit (approved):

derekbit:
@mergify backport v1.8.x

PR #3390, opened 2024-12-23
fix: ignore backing image real size inconsistency between nodes

ChanYiLin:
ref: 
- https://github.com/longhorn/longhorn/issues/9976
- https://github.com/longhorn/longhorn/issues/9882

We found that the real size may change when syncing with other nodes.
And it may also changes when transferring to different OS.


derekbit (approved):

derekbit:
@mergify backport v1.8.x

PR #3392, opened 2024-12-23
fix: add data engine option to restore backing image api

ChanYiLin:
ref: 
- https://github.com/longhorn/longhorn/issues/6341
- https://github.com/longhorn/longhorn/issues/10023

Fix:
- generate UUID for backupBackingImage when backing up from UI  (cc @mantissahz )
    - To fix: https://github.com/longhorn/longhorn/issues/10023
- add `DataEngine` option in restoring backup backing image
- restore volume will check the existing backing image dataEngine or restore backupBackingImage with correct dataEngine

cc @derekbit @mantissahz @houhoucoop

### DIFF
--- a/api/backupbackingimage.go
+++ b/api/backupbackingimage.go
@@ -5,9 +5,10 @@ import (
 
 	"github.com/gorilla/mux"
 	"github.com/pkg/errors"
-
 	"github.com/rancher/go-rancher/api"
 	"github.com/rancher/go-rancher/client"
+
+	"github.com/longhorn/longhorn-manager/types"
 )
 
 func (s *Server) backupBackingImageList(apiContext *api.ApiContext) (*client.GenericCollection, error) {
@@ -56,7 +57,7 @@ func (s *Server) BackupBackingImageRestore(w http.ResponseWriter, req *http.Requ
 	}
 
 	backupBackingImageName := mux.Vars(req)["name"]
-	if err := s.m.RestoreBackupBackingImage(backupBackingImageName, input.Secret, input.SecretNamespace); err != nil {
+	if err := s.m.RestoreBackupBackingImage(backupBackingImageName, input.Secret, input.SecretNamespace, input.DataEngine); err != nil {
 		return errors.Wrapf(err, "failed to restore backup backing image '%s'", backupBackingImageName)
 	}
 	return nil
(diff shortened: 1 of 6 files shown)


api/backupbackingimage.go
```diff
@@ -4,6 +4,7 @@ import (
 	"net/http"
 
 	"github.com/gorilla/mux"
+	"github.com/longhorn/longhorn-manager/types"
 	"github.com/pkg/errors"
```
derekbit, line 8: Reorg

manager/backupbackingimage.go
```diff
@@ -41,7 +41,7 @@ func (m *VolumeManager) DeleteBackupBackingImage(name string) error {
 	return m.ds.DeleteBackupBackingImage(name)
 }
 
-func (m *VolumeManager) RestoreBackupBackingImage(name string, secret, secretNamespace string) error {
+func (m *VolumeManager) RestoreBackupBackingImage(name string, secret, secretNamespace string, dataEngine string) error {
```
derekbit, line 44:
```suggestion
func (m *VolumeManager) RestoreBackupBackingImage(name string, secret, secretNamespace, dataEngine string) error {
```

derekbit (commented):
In general, LGTM. Few comments need to be addressed.

derekbit:
@mergify backport v1.8.x

api/backupbackingimage.go
```diff
@@ -71,8 +72,9 @@ func (s *Server) BackupBackingImageCreate(w http.ResponseWriter, req *http.Reque
 	}
 
 	backingImageName := mux.Vars(req)["name"]
-	if err := s.m.CreateBackupBackingImage(input.Name, backingImageName, input.BackupTargetName); err != nil {
-		return errors.Wrapf(err, "failed to create backup backing image '%s'", input.Name)
+	backupBackingImageName := types.GetBackupBackingImageNameFromBIName(backingImageName)
```
mantissahz, line 75: - Can we check if `input.Name` is equal to `backingImageName` or an empty string, and then generate a random UUID name?
- Or we won't allow users to customize the backup backing image name from the API when creating it.

manager/backupbackingimage.go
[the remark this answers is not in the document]
ChanYiLin: fixed.

api/backupbackingimage.go
ChanYiLin: fixed.

manager/backupbackingimage.go
ChanYiLin: fixed.

api/backupbackingimage.go
ChanYiLin: cc @mantissahz @derekbit 
We don't allow customizing the backup backing image name

The name is directly from the URL
`HTTP/1.1 POST /v1/backingimages/parrot?action=backupBackingImageCreate`
```
	r.Methods("DELETE").Path("/v1/backingimages/{name}").Handler(f(schemas, s.BackingImageDelete))
	backingImageActions := map[string]func(http.ResponseWriter, *http.Request) error{
		"backupBackingImageCreate": s.BackupBackingImageCreate,
	}
```

Just like we don't allow customizing the backup volume name

mantissahz: Got it. It's reasonable.

derekbit (dismissed):

derekbit (approved):

PR #3397, opened 2024-12-23
fix(monitor): remove the close of monitors

derekbit:

Issue Longhorn/longhorn#10035

No need to close the monitors after deleting a node resource. If a node is gone, monitor will return error and won't continue the following tasks.

e2e
https://ci.longhorn.io/job/private/job/longhorn-e2e-test/2295/console


derekbit:
@mergify backport v1.8.x

c3y1huang (approved):

PR #3400, opened 2024-12-24
fix: reject creating the same backing image to the same backup target

ChanYiLin:
ref: longhorn/longhorn#10023

reject creating the same backing image to the same backup target

- Create a backup backing image to BackupTarget: `nfs`
- Do the same operation again, it should be reject because it already exists.

Since now the BackupBackingImage name has suffix random UUID
So this is not blocked naturally and need to use webhook to do the validation.


ChanYiLin:
cc @mantissahz @derekbit

mantissahz (approved):
LGTM

derekbit (approved):
LGTM

derekbit:
@mergify backport v1.8.x

PR #3401, opened 2024-12-24
feat: v2 fast rebuilding

shuo-wu:

https://github.com/longhorn/longhorn/issues/9488


controller/volume_controller.go
```diff
@@ -4983,9 +4983,17 @@ func (c *VolumeController) shouldCleanUpFailedReplica(v *longhorn.Volume, r *lon
...
 	}
-	// TODO: Remove it once we can reuse failed replicas during v2 rebuilding
+
 	if types.IsDataEngineV2(v.Spec.DataEngine) {
-		return true
+		V2DataEngineFastReplicaRebuilding, err := c.ds.GetSettingAsBool(types.SettingNameV2DataEngineFastReplicaRebuilding)
+		if err != nil {
+			log.WithError(err).Warnf("Failed to get the setting %v, will consider it as false", types.SettingDefinitionV2DataEngineFastReplicaRebuilding)
+			V2DataEngineFastReplicaRebuilding = false
+		}
```
derekbit, line 4992: Should we make it optional?

derekbit (commented):
LGTM.
Only one question and need to update the dependencies.

types/setting.go
```diff
@@ -1443,6 +1446,16 @@ var (
 		Default:     "",
 	}
 
+	SettingDefinitionV2DataEngineFastReplicaRebuilding = SettingDefinition{
+		DisplayName: "V2 Data Engine Fast Replica Rebuilding",
```
innobead, line 1450: I assume this setting is temporary before v2 becomes GA, as delta replica rebuilding should be eventually enabled always. Right?

innobead (commented):
LGTM, just one question.

go.mod
```diff
@@ -4,6 +4,11 @@ go 1.23.1
 
 toolchain go1.23.4
 
+replace (
+	github.com/longhorn/longhorn-instance-manager => github.com/shuo-wu/longhorn-instance-manager v0.0.0-20241221010736-e78d54cf1bdf
+	github.com/longhorn/longhorn-spdk-engine => github.com/shuo-wu/longhorn-spdk-engine v0.0.0-20241220193646-793fe7001495
+)
+
```
derekbit, line 11: Update the deps

types/setting.go
shuo-wu: YES! Next, I will check when we can launch a setting similar to v1 `Snapshot Data Integrity` instead. If we can do that before v1.8 official release, `V2 Data Engine Fast Replica Rebuilding` won't be delivered to users at all.

derekbit (approved):
LGTM

derekbit:
@mergify backport v1.8.x

PR #3405, opened 2024-12-24
fix(webhook): endless secret update retries

c3y1huang:

Issue longhorn/longhorn#10054

Introduce leader election to sequentially start webhook servers, one node at a time.

`None`

`None`

### DIFF
--- a/app/daemon.go
+++ b/app/daemon.go
@@ -12,7 +12,8 @@ import (
 	"github.com/rancher/wrangler/v3/pkg/signals"
 	"github.com/sirupsen/logrus"
 	"github.com/urfave/cli"
-	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/longhorn/go-iscsi-helper/iscsi"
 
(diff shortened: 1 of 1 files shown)


c3y1huang:
@coderabbitai review

derekbit (approved):
LGTM

derekbit:
@mergify backport v1.8.x

derekbit:
@mergify backport 1.7.x v1.6.x

ChanYiLin (approved):
LGTM

innobead:
> #### What this PR does / why we need it:
> Introduce leader election to sequentially start webhook servers, one node at a time.

Is there any side effect for this improvement? like pods might be late to start? @c3y1huang

innobead:
Suggest moving the code below to `startWebhooksByLeaderElection` as well, as it is part of webhook initialization. Additionally, pass `clients` to `startWebhooksByLeaderElection` instead of creating a new one in the method. 

[/app/daemon.go#L285-L299](https://github.com/longhorn/longhorn-manager/blob/7114b699543d268b3a42a3529b569284e69eeb14/app/daemon.go#L285-L299)
```go
	clients, err := client.NewClients(kubeconfigPath, true, ctx.Done())
	if err != nil {
		return err
	}

	if err := clients.Datastore.AddLabelToManagerPod(currentNodeID, types.GetAdmissionWebhookLabel()); err != nil {
		return err
	}
	if err := webhook.CheckWebhookServiceAvailability(types.WebhookTypeAdmission); err != nil {
		return err
	}

	if err := recoverybackend.StartRecoveryBackend(clients); err != nil {
		return err
	}

```

c3y1huang:
> Is there any side effect for this improvement? like longhorn-manager app might be late to start? @c3y1huang

The manager will start sequentially, which may impact smaller clusters with slower restarts. However, in my view, this impact is not significant.

innobead:
> > Is there any side effect for this improvement? like longhorn-manager app might be late to start? @c3y1huang
> 
> The manager will start sequentially, which may impact smaller clusters with slower restarts. However, in my view, this impact is not significant.

Sounds good. However, as we use this in nodes running longhorn-manager pods and it differs from traditional Kubernetes control plane usage in terms of scale, see whether this needs any tests in scale if feasible. 

BTW, do we use any lease elsewhere?

c3y1huang:
@mergify backport v1.7.x

PR #3406, opened 2024-12-24
feat(env-check): remove dependency of system package manager from node env checker

COLDTURNIP:

Issue longhorn/longhorn#10019

Instead of reading the system package database, check the executables needed by the system, to eliminate the system management tool chain issue.

### DIFF
--- a/controller/monitor/environment_check_monitor.go
+++ b/controller/monitor/environment_check_monitor.go
@@ -221,9 +221,7 @@ func (m *EnvironmentCheckMonitor) syncPackagesInstalled(kubeNode *corev1.Node, n
 		packageProbeExecutables["sys-fs/cryptsetup"] = "cryptsetup"
 		packageProbeExecutables["sys-fs/lvm2"] = "dmsetup"
 	default:
-		collectedData.conditions = types.SetCondition(collectedData.conditions, longhorn.NodeConditionTypeRequiredPackages, longhorn.ConditionStatusFalse,
-			string(longhorn.NodeConditionReasonUnknownOS),
-			fmt.Sprintf("Unable to verify the required packages because the OS image '%v' is unknown to the Longhorn system. Please ensure the required packages are installed.", osImage))
+		// unsupported host platform, skip environment check condition
 		return
 	}
 
--- a/controller/monitor/environment_check_monitor.go
+++ b/controller/monitor/environment_check_monitor.go
@@ -178,10 +178,8 @@ func (m *EnvironmentCheckMonitor) environmentCheck(kubeNode *corev1.Node) *Colle
 
 func (m *EnvironmentCheckMonitor) syncPackagesInstalled(kubeNode *corev1.Node, namespaces []lhtypes.Namespace, collectedData *CollectedEnvironmentCheckInfo) {
 	osImage := strings.ToLower(kubeNode.Status.NodeInfo.OSImage)
-	queryPackagesCmd := ""
-	options := []string{}
-	packages := []string{}
-	pipeFlag := false
+
+	packageProbeExecutables := make(map[string]string)
 
 	switch {
 	case strings.Contains(osImage, "talos"):
@@ -190,10 +188,10 @@ func (m *EnvironmentCheckMonitor) syncPackagesInstalled(kubeNode *corev1.Node, n
 	case strings.Contains(osImage, "ubuntu"):
 		fallthrough
 	case strings.Contains(osImage, "debian"):
-		queryPackagesCmd = "dpkg"
-		options = append(options, "-l")
-		packages = append(packages, "nfs-common", "open-iscsi", "cryptsetup", "dmsetup")
-		pipeFlag = true
+		packageProbeExecutables["nfs-common"] = "mount.nfs"
+		packageProbeExecutables["open-iscsi"] = "iscsiadm"
+		packageProbeExecutables["cryptsetup"] = "cryptsetup"
+		packageProbeExecutables["dmsetup"] = "dmsetup"
 	case strings.Contains(osImage, "centos"):
 		fallthrough
 	case strings.Contains(osImage, "fedora"):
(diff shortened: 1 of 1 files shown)


controller/monitor/environment_check_monitor.go
```diff
@@ -203,21 +201,25 @@ func (m *EnvironmentCheckMonitor) syncPackagesInstalled(kubeNode *corev1.Node, n
...
+		packageProbeExecutables["cryptsetup"] = "cryptsetup"
+		packageProbeExecutables["device-mapper"] = "dmsetup"
 	case strings.Contains(osImage, "gentoo"):
-		queryPackagesCmd = "qlist"
-		options = append(options, "-I")
-		packages = append(packages, "net-fs/nfs-utils", "sys-block/open-iscsi", "sys-fs/cryptsetup", "sys-fs/lvm2")
+		packageProbeExecutables["net-fs/nfs-utils"] = "mount.nfs"
+		packageProbeExecutables["sys-block/open-iscsi"] = "iscsiadm"
+		packageProbeExecutables["sys-fs/cryptsetup"] = "cryptsetup"
+		packageProbeExecutables["sys-fs/lvm2"] = "dmsetup"
```
derekbit, line 222: Excellent!

controller/monitor/environment_check_monitor.go
```diff
@@ -190,10 +188,10 @@ func (m *EnvironmentCheckMonitor) syncPackagesInstalled(kubeNode *corev1.Node, n
 	case strings.Contains(osImage, "ubuntu"):
 		fallthrough
 	case strings.Contains(osImage, "debian"):
-		queryPackagesCmd = "dpkg"
-		options = append(options, "-l")
-		packages = append(packages, "nfs-common", "open-iscsi", "cryptsetup", "dmsetup")
-		pipeFlag = true
+		packageProbeExecutables["nfs-common"] = "mount.nfs"
```
COLDTURNIP, line 191: We don't directly use mount.nfs in Longhorn. Just ensure the system NFS capability is fulfilled.

controller/monitor/environment_check_monitor.go
```diff
@@ -233,36 +235,27 @@ func (m *EnvironmentCheckMonitor) syncPackagesInstalled(kubeNode *corev1.Node, n
...
-		if !pipeFlag {
-			args = append(args, pkg)
-		}
-		queryResult, err := nsexec.Execute(nil, queryPackagesCmd, args, lhtypes.ExecuteDefaultTimeout)
-		if err != nil {
+	installedPackages := []string{}
+	notInstalledPackages := []string{}
+	for pkg, executable := range packageProbeExecutables {
+		shellCmd := fmt.Sprintf("command -v %s", executable)
+		if _, err := nsexec.Execute(nil, "sh", []string{"-c", shellCmd}, lhtypes.ExecuteDefaultTimeout); err != nil {
```
COLDTURNIP, line 242: `command` is a shell function, not an independent executable, hence we have to wrap with `sh -c "..."`. The reason we don't use `which` is to minimize dependency. `which` is not installed on some Linux distro by default.

derekbit (approved):
LGTM

derekbit:
@mergify backport v1.7.x

derekbit:
@mergify backport v1.8.x

controller/monitor/environment_check_monitor.go
```diff
@@ -203,66 +201,49 @@ func (m *EnvironmentCheckMonitor) syncPackagesInstalled(kubeNode *corev1.Node, n
...
-		packages = append(packages, "net-fs/nfs-utils", "sys-block/open-iscsi", "sys-fs/cryptsetup", "sys-fs/lvm2")
+		packageProbeExecutables["net-fs/nfs-utils"] = "mount.nfs"
+		packageProbeExecutables["sys-block/open-iscsi"] = "iscsiadm"
+		packageProbeExecutables["sys-fs/cryptsetup"] = "cryptsetup"
+		packageProbeExecutables["sys-fs/lvm2"] = "dmsetup"
 	default:
 		collectedData.conditions = types.SetCondition(collectedData.conditions, longhorn.NodeConditionTypeRequiredPackages, longhorn.ConditionStatusFalse,
 			string(longhorn.NodeConditionReasonUnknownOS),
 			fmt.Sprintf("Unable to verify the required packages because the OS image '%v' is unknown to the Longhorn system. Please ensure the required packages are installed.", osImage))
 		return
```
derekbit, line 227: I'm wondering if we can remove the message if Longhorn doesn't support the OS.

mantissahz (approved):
LGTM

controller/monitor/environment_check_monitor.go
derekbit: WDYT? @COLDTURNIP @mantissahz

mantissahz: Do we have another place/information to show users that the OS is unsupported except for official documents?
If users are aware of the unsupported OS and responsible for it when installing Longhorn, this message is annoying.

In general, I'm fine to remove it.

derekbit: > Do we have another place/information to show users that the OS is unsupported except for official documents?
https://longhorn.io/docs/1.8.0/best-practices/#operating-system

Yes, we don't need to handle the OSes that Longhorn doesn't support or handle.

COLDTURNIP: Agree to remove. Maybe we can show such warning on the UI, or we already have CLI preflight checker at least. Node condition is too deep for user to get notice.

derekbit: > Node condition is too deep for user to get notice.

Node conditions are good for troubleshooting and self-diagnosis. However, there are too many OSes in the market, so we only need to take care of the OSes mentioned in the official doc.

derekbit (dismissed):

controller/monitor/environment_check_monitor.go
COLDTURNIP: Will append a commit to remove this condition for unsupported OSes.

derekbit: Isn't it removed?

COLDTURNIP: Yeah, we're going to remove the condition completely on unsupported OSes.

derekbit (approved):
LGTM

mantissahz (approved):

PR #3416, opened 2024-12-27
fix: backing image should follow the same data engine as volume in strorageclass

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/10084

backing image should be created with the same data engine as volume in storageclass


derekbit (approved):
LGTM

derekbit:
@mergify backport v1.8.x

PR #3425, opened 2024-12-30
feat(backup): separate backup settings from default settings

mantissahz:

Issue # longhorn/longhorn#5411, longhorn/longhorn#10089

Separate the default backup related settings from default settings into a new yaml file `default-backupstore-setting.yaml`. Check if ConfigMap `longhorn-default-backupstore-settings` changed to update the default backup target.

### DIFF
--- a/controller/kubernetes_configmap_controller.go
+++ b/controller/kubernetes_configmap_controller.go
@@ -211,6 +211,11 @@ func (kc *KubernetesConfigMapController) reconcile(namespace, cfmName string) er
 		if err := kc.ds.UpdateCustomizedSettings(nil); err != nil {
 			return errors.Wrap(err, "failed to update built-in settings with customized values")
 		}
+	// Users tries to update the default backup target with customized values using Helm at runtime.
+	case types.DefaultDefaultResourceConfigMapName:
+		if err := kc.ds.CreateOrUpdateDefaultBackupTarget(); err != nil {
+			return errors.Wrap(err, "failed to create or update default backup target with customized values")
+		}
 	}
 
 	return nil
--- a/datastore/longhorn.go
+++ b/datastore/longhorn.go
@@ -56,7 +56,7 @@ func (s *DataStore) UpdateCustomizedSettings(defaultImages map[types.SettingName
 		return err
 	}
 
-	customizedDefaultSettings, customizedDefaultBackupTargetSettings, err := types.GetCustomizedDefaultSettings(defaultSettingCM)
+	customizedDefaultSettings, err := types.GetCustomizedDefaultSettings(defaultSettingCM)
 	if err != nil {
 		return err
 	}
@@ -72,10 +72,6 @@ func (s *DataStore) UpdateCustomizedSettings(defaultImages map[types.SettingName
 	}
 
 	if err := s.createNonExistingSettingCRsWithDefaultSetting(defaultSettingCM.ResourceVersion); err != nil {
-		return err
-	}
-
-	if err := s.syncDefaultBackupTargetResourceWithCustomizedSettings(customizedDefaultBackupTargetSettings); err != nil {
 		return err
 	}
 
@@ -213,69 +209,6 @@ func (s *DataStore) applyCustomizedDefaultSettingsToDefinitions(customizedDefaul
 		}
 	}
 	return nil
-}
-
-func (s *DataStore) syncDefaultBackupTargetResourceWithCustomizedSettings(customizedDefaultBackupTargetSettings map[string]string) error {
-	backupTarget := &longhorn.BackupTarget{
-		ObjectMeta: metav1.ObjectMeta{
-			Name: types.DefaultBackupTargetName,
-		},
-		Spec: longhorn.BackupTargetSpec{
-			PollInterval: metav1.Duration{Duration: types.DefaultBackupstorePollInterval},
-		},
-	}
-	backupTargetURL, urlExists := customizedDefaultBackupTargetSettings[string(types.SettingNameBackupTarget)]
-	backupTargetCredentialSecret, secretExists := customizedDefaultBackupTargetSettings[string(types.SettingNameBackupTargetCredentialSecret)]
-	pollIntervalStr, pollInterExists := customizedDefaultBackupTargetSettings[string(types.SettingNameBackupstorePollInterval)]
-
-	if backupTargetURL != "" {
-		backupTarget.Spec.BackupTargetURL = backupTargetURL
-	}
-	if backupTargetCredentialSecret != "" {
-		backupTarget.Spec.CredentialSecret = backupTargetCredentialSecret
-	}
-
-	if pollIntervalStr != "" {
-		pollIntervalAsInt, err := strconv.ParseInt(pollIntervalStr, 10, 64)
-		if err != nil {
-			return err
-		}
-		backupTarget.Spec.PollInterval = metav1.Duration{Duration: time.Duration(pollIntervalAsInt) * time.Second}
-	}
-
-	existingBackupTarget, err := s.GetBackupTarget(types.DefaultBackupTargetName)
-	if err != nil {
-		if !ErrorIsNotFound(err) {
-			return err
-		}
-		_, err = s.CreateBackupTarget(backupTarget)
-		if apierrors.IsAlreadyExists(err) {
-			return nil // Already exists, no need to return error
-		}
-		return err
-	}
-
-	// If the settings are not in the longhorn-default-setting ConfigMap, use the existing values.
-	if !urlExists {
-		backupTarget.Spec.BackupTargetURL = existingBackupTarget.Spec.BackupTargetURL
-	}
-	if !secretExists {
-		backupTarget.Spec.CredentialSecret = existingBackupTarget.Spec.CredentialSecret
-	}
-	if !pollInterExists {
-		backupTarget.Spec.PollInterval = existingBackupTarget.Spec.PollInterval
-	}
-	syncTime := metav1.Time{Time: time.Now().UTC()}
-	backupTarget.Spec.SyncRequestedAt = syncTime
-	existingBackupTarget.Spec.SyncRequestedAt = syncTime
-
-	if reflect.DeepEqual(existingBackupTarget.Spec, backupTarget.Spec) {
-		return nil // No changes, no need to update
-	}
-
-	existingBackupTarget.Spec = backupTarget.Spec
-	_, err = s.UpdateBackupTarget(existingBackupTarget)
-	return err
 }
 
 func (s *DataStore) syncSettingCRsWithCustomizedDefaultSettings(customizedDefaultSettings map[string]string, defaultSettingCMResourceVersion string) error {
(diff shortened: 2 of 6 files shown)


mantissahz:
Regression test:
- Settings: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/8091/
- Backup Related: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/8092/

types/types.go
```diff
@@ -208,12 +208,13 @@ const (
...
-	DepracatedDriverName               = "io.rancher.longhorn"
-	DefaultStorageClassConfigMapName   = "longhorn-storageclass"
-	DefaultDefaultSettingConfigMapName = "longhorn-default-setting"
-	DefaultStorageClassName            = "longhorn"
-	ControlPlaneName                   = "longhorn-manager"
+	DeprecatedProvisionerName                      = "rancher.io/longhorn"
+	DepracatedDriverName                           = "io.rancher.longhorn"
+	DefaultStorageClassConfigMapName               = "longhorn-storageclass"
+	DefaultDefaultSettingConfigMapName             = "longhorn-default-setting"
+	DefaultDefaultBackupStoreSettingsConfigMapName = "longhorn-default-backupstore-settings"
```
derekbit, line 215:
```suggestion
	DefaultDefaultBackupStoreConfigMapName = "longhorn-default-backupstore"
```

datastore/longhorn.go
```diff
@@ -4168,6 +4101,89 @@ func (s *DataStore) ListShareManagersRO() ([]*longhorn.ShareManager, error) {
 	return s.shareManagerLister.ShareManagers(s.namespace).List(labels.Everything())
 }
 
+// UpdateDefaultBackupStoreSettings updates the default backup store settings from the ConfigMap longhorn-default-backupstore-settings
+func (s *DataStore) UpdateDefaultBackupStoreSettings() error {
```
derekbit, line 4105:
```suggestion
// CreateOrUpdateDefaultBackupTarget creates or updates the default backup target from the ConfigMap longhorn-default-backupstore
func (s *DataStore) CreateOrUpdateDefaultBackupTarget() error {
```

controller/kubernetes_configmap_controller.go
```diff
@@ -211,6 +211,10 @@ func (kc *KubernetesConfigMapController) reconcile(namespace, cfmName string) er
 		if err := kc.ds.UpdateCustomizedSettings(nil); err != nil {
 			return errors.Wrap(err, "failed to update built-in settings with customized values")
 		}
+	case types.DefaultDefaultBackupStoreSettingsConfigMapName:
+		if err := kc.ds.UpdateDefaultBackupStoreSettings(); err != nil {
+			return errors.Wrap(err, "failed to update default backup store settings with customized values")
+		}
 	}
```
derekbit, line 218:
```suggestion
	case types.DefaultDefaultBackupStoreConfigMapName:
		if err := kc.ds.CreateOrUpdateDefaultBackupTarget(); err != nil {
			return errors.Wrap(err, "failed to create or update default backup target with customized values")
		}
	}
```

datastore/longhorn.go
```diff
@@ -4168,6 +4101,89 @@ func (s *DataStore) ListShareManagersRO() ([]*longhorn.ShareManager, error) {
...
+	defaultBackupStoreCM, err := s.GetConfigMapRO(s.namespace, types.DefaultDefaultBackupStoreSettingsConfigMapName)
```
derekbit, line 4106:
```suggestion
	defaultBackupStoreCM, err := s.GetConfigMapRO(s.namespace, types.DefaultDefaultBackupStoreConfigMapName)
```

datastore/longhorn.go
```diff
@@ -4168,6 +4101,89 @@ func (s *DataStore) ListShareManagersRO() ([]*longhorn.ShareManager, error) {
...
+	if err != nil {
+		return err
+	}
+
+	defaultBackupStoreSettingsYAMLData := []byte(defaultBackupStoreCM.Data[types.DefaultBackupStoreSettingsYAMLFileName])
```
derekbit, line 4111:
```suggestion
	defaultBackupStoreYAMLData := []byte(defaultBackupStoreCM.Data[types.DefaultBackupStoreYAMLFileName])
```

datastore/longhorn.go
```diff
@@ -4168,6 +4101,89 @@ func (s *DataStore) ListShareManagersRO() ([]*longhorn.ShareManager, error) {
...
+
+	defaultBackupStoreSettings, err := util.GetSettingFromYAML(defaultBackupStoreSettingsYAMLData)
+	if err != nil {
+		return err
+	}
+
+	if err := s.syncDefaultBackupTargetResourceWithCustomizedSettings(defaultBackupStoreSettings); err != nil {
+		return err
+	}
+	return nil
```
derekbit, line 4121:
```suggestion
	defaultBackupStoreParameters, err := util.GetSettingFromYAML(defaultBackupStoreSettingsYAMLData)
	if err != nil {
		return err
	}

	if err := s.syncDefaultBackupTargetResourceWithCustomizedParameters(defaultBackupStoreParameters); err != nil {
		return err
	}
	return nil
```

types/setting.go
```diff
@@ -10,19 +10,20 @@ import (
...
 	"github.com/longhorn/longhorn-manager/meta"
+	"github.com/longhorn/longhorn-manager/util"
 
 	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
 )
 
 const (
-	DefaultSettingYAMLFileName = "default-setting.yaml"
+	DefaultSettingYAMLFileName             = "default-setting.yaml"
+	DefaultBackupStoreSettingsYAMLFileName = "default-backupstore-settings.yaml"
```
derekbit, line 26:
```suggestion
	DefaultBackupStoreYAMLFileName = "default-backupstore.yaml"
```

types/setting.go
```diff
@@ -138,7 +139,7 @@ const (
 	SettingNameDefaultMinNumberOfBackingImageCopies                     = SettingName("default-min-number-of-backing-image-copies")
 	SettingNameBackupExecutionTimeout                                   = SettingName("backup-execution-timeout")
 	SettingNameRWXVolumeFastFailover                                    = SettingName("rwx-volume-fast-failover")
-	// These three backup target settings are used in the "longhorn-default-setting" ConfigMap
+	// These three backup target settings are used in the "longhorn-default-backupstore-settings" ConfigMap
```
derekbit, line 142:
```suggestion
	// These three backup target parameters are used in the "longhorn-default-backupstore" ConfigMap
```

derekbit (commented):
1. To prevent any confusion, suggest removing `settings` as the comments.
2. Need to remove the three default backup store settings while upgrading to v1.8.x.
3. Create or update default backup store after https://github.com/longhorn/longhorn-manager/blob/master/app/daemon.go#L330-L332.

controller/kubernetes_configmap_controller.go
```diff
@@ -211,6 +211,10 @@ func (kc *KubernetesConfigMapController) reconcile(namespace, cfmName string) er
 		if err := kc.ds.UpdateCustomizedSettings(nil); err != nil {
 			return errors.Wrap(err, "failed to update built-in settings with customized values")
 		}
+	case types.DefaultDefaultBackupStoreSettingsConfigMapName:
+		if err := kc.ds.UpdateDefaultBackupStoreSettings(); err != nil {
+			return errors.Wrap(err, "failed to update default backup store settings with customized values")
+		}
```
derekbit, line 217: Do we need to handle the configmap in each reconciliation? Or, we only need to read and create (or update) the default configmap while starting the deamon?

types/types.go
mantissahz: Modified. Thanks.

datastore/longhorn.go
mantissahz: Modified. Thanks.

controller/kubernetes_configmap_controller.go
mantissahz: Modified. Thanks.

datastore/longhorn.go
mantissahz: Modified. Thanks.

datastore/longhorn.go
mantissahz: Modified. Thanks.

types/setting.go

controller/kubernetes_configmap_controller.go
mantissahz: >Why do we need to handle the default backupstore configmap in each reconciliation?

The controller will only start to reconcile the ConfigMap `longhorn-default-backupstore` when it is modified.

>Isn't it a one-time job?

No. We need to monitor the ConfigMap when it is modified for these three parameters to update the default backup target as we did in default settings before.
Or we will not allow users to update the default backup target parameters at runtime.

mantissahz:
>Create or update default backup store after https://github.com/longhorn/longhorn-manager/blob/master/app/daemon.go#L330-L332. 

Added, Thanks @derekbit

datastore/longhorn.go
mantissahz: Modified. Thanks.

types/setting.go

controller/kubernetes_configmap_controller.go
derekbit: I don't think we need to handle this for the default backupstore
1. Fresh installation
    - User sets the parameters for the default backupstore in chart/values.yaml
    - app/daemon.go creates the default backupstore based on the parameters
    - If user updates the configmap, we don't need to sync them
      - Any benefit that we sync the configmap to the default backupstore?
 2. Upgrade
    - Case 1
      -  User updates the parameters for the default backupstore in chart/values.yaml
      - app/daemon.go updates the default backupstore based on the parameters
    - Case 2
      - User doesn't set the parameters for the default backupstore in chart/values.yaml
      - app/daemon.go doesn't need to update the default backupstore

mantissahz: >If user updates the configmap, we don't need to sync them
Any benefit that we sync the configmap to the default backupstore?

It's a bit weird if users update the values using Helm at runtime, but nothing changes.

mantissahz: Added a comment to kubernete configmap controller for updating the `longhorn-default-resource` configmap using Helm.

util/util.go
```diff
@@ -877,3 +877,15 @@ func GetDataEngineForDiskType(diskType longhorn.DiskType) longhorn.DataEngineTyp
...
 }
+
+// GetSettingFromYAML unmarshals the setting YAML data into a map
+func GetSettingFromYAML(settingsYAMLData []byte) (map[string]string, error) {
+	customizedSettings := map[string]string{}
+
+	if err := yaml.Unmarshal(settingsYAMLData, &customizedSettings); err != nil {
+		logrus.WithError(err).Errorf("Failed to unmarshal customized settings from yaml data %v, will give up using them", string(settingsYAMLData))
+		customizedSettings = map[string]string{}
+	}
```
derekbit, line 888: No need to return error?

derekbit (commented):
In general, LGTM

util/util.go
mantissahz: Yes, it needs.
Modified. Thanks

derekbit (approved):
LGTM.

The test cases need to include
- Fresh installation 
  - empty configmap
  -  configmap with values
- Upgrade
-

mantissahz:
@Mergifyio backport v1.8.x

derekbit:
@mergify backport v1.8.x

PR #3426, opened 2024-12-30
fix(env-check): remove monitor initial pending to fix disk mismatch problem

COLDTURNIP:
EDIT: change to draft since the initial waiting is expected for while disk mismatch really happens.

Issue longhorn/longhorn#10098

Trigger the initial disk status collecting right after the disk monitor get created, to eliminate disk mismatching errors, which blocks the rest node sync logic.

### DIFF
--- a/controller/monitor/environment_check_monitor.go
+++ b/controller/monitor/environment_check_monitor.go
@@ -45,11 +45,16 @@ var (
 	nfsProtocolVersions = map[string]bool{"4.0": true, "4.1": true, "4.2": true}
 )
 
-type EnvironmentCheckMonitor struct {
+type EnvironmentCheckMonitor Monitor[[]longhorn.Condition]
+
+var _ EnvironmentCheckMonitor = &environmentCheckMonitorImpl{}
+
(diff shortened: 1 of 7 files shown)


controller/monitor/environment_check_monitor.go
```diff
@@ -74,39 +79,44 @@ func NewEnvironmentCheckMonitor(logger logrus.FieldLogger, ds *datastore.DataSto
...
-		return false, nil
-	}); err != nil {
-		if errors.Is(err, context.Canceled) {
-			m.logger.WithError(err).Warning("Environment check monitor is stopped")
-		} else {
-			m.logger.WithError(err).Error("Failed to start environment check monitor")
-		}
-	}
+		go func() {
+			if err := wait.PollUntilContextCancel(m.ctx, m.syncPeriod, false, func(context.Context) (bool, error) {
```
COLDTURNIP, line 91: Explicitly invoke `m.run` to eliminate the chance that `m.GetCollectedData` get invoked before the first collecting. The `wait.PollUntilContextCancel` is scheduled in another goroutine.

PR #3429, opened 2025-01-02
feat: no need to truncate block-type disk's StorageAvailable value

derekbit:

Issue Longhorn/longhorn#10121

Unlink a filesyste-type disk that contains log, etc and other files, there is no need to truncate block-type disk's StorageAvailable value.


innobead (dismissed):

innobead:
@mergify backport v1.8.x

innobead:
@derekbit there is a format error; probably there is a missing indent.

derekbit:
> @derekbit there is a format error; probably there is a missing indent.

Fixed. Thank you.

innobead (approved):

PR #3436, opened 2025-01-08
fix: use `experimental` for alpha features

derekbit:

Issue Longhorn/longhorn#9794

Experimental: An early stage "Preview," these features truly are experiments and might never reach GA.


innobead (approved):

derekbit:
@mergify backport v1.8.x

PR #3246, opened 2024-10-30
feat(system-backup): check outdated backup in if-not-present volume backup policy

c3y1huang:
#### Which issue(s) this PR fixes:

Issue longhorn/longhorn#6027

#### What this PR does / why we need it:

When the system backup volume backup policy is set to `if-not-present`, check the last backup and create a backup when it's outdated. 

https://github.com/longhorn/longhorn/issues/6027#issuecomment-2445884734

#### Special notes for your reviewer:

`None`

#### Additional documentation or context

`None`

PR #3437, opened 2025-01-08
fix: handle backup with non-existing snapshot and fix recurring job to correctly handle pending backup 

PhanLe1010:

https://github.com/longhorn/longhorn/issues/10090

This PR has 2 commits:
1. fix: recurring job should not fail on pending backup
2. fix: backup with non-existing snapshot should be marked as errored
    Right now, backup with non-existing snapshot is stuck in pending forever.
    This is not good because it will keep the volume in attached state
    forever.
    Moreover, snapshot doesn't exist is a non-recoverable error so
    the backup should be marked as errored

### Demo:

**Before the fix**

Deploying a backup with non-existing snapshot 
```yaml
kind: Backup
metadata:
  name: test-backup
  namespace: longhorn-system
  labels:
    backup-volume: pvc-34e83a41-bb6a-4406-b9d9-91fe3a8fcfbc
spec:
  backupMode: incremental
  snapshotName: non-exist-snapshot-name
```
The backup will stuck in pending state forever 
```yaml
status:
  backupCreatedAt: ""
  compressionMethod: ""
  labels: null
  lastSyncedAt: null
  messages:
    info: Failed to get the Snapshot non-exist-snapshot-name
  newlyUploadDataSize: ""
  ownerID: phan-v710-pool2-sbk7f-7dc5x
  progress: 0
  reUploadedDataSize: ""
  replicaAddress: ""
  size: ""
  snapshotCreatedAt: ""
  snapshotName: ""
  state: Pending
```
and keep the volume in attached state:
```yaml
spec:
  attachmentTickets:
    backup-controller-test-backup:
      generation: 0
      id: backup-controller-test-backup
      nodeID: phan-v710-pool2-sbk7f-9jmsr
      parameters:
        disableFrontend: any
      type: backup-controller
  volume: pvc-34e83a41-bb6a-4406-b9d9-91fe3a8fcfbc
status:
  attachmentTicketStatuses:
    backup-controller-test-backup:
      conditions:
      - lastProbeTime: ""
        lastTransitionTime: "2025-01-08T06:52:10Z"
        message: ""
        reason: ""
        status: "True"
        type: Satisfied
      generation: 0
      id: backup-controller-test-backup
      satisfied: true
```

**After the fix**

The backup is marked as error and the volume is detached

### DIFF
--- a/app/recurring_job.go
+++ b/app/recurring_job.go
@@ -656,7 +656,7 @@ func (job *Job) doRecurringBackup() (err error) {
 		case string(longhorn.BackupStateCompleted):
 			complete = true
 			job.logger.Infof("Completed creating backup %v", info.Id)
-		case string(longhorn.BackupStateNew), string(longhorn.BackupStateInProgress):
+		case string(longhorn.BackupStateNew), string(longhorn.BackupStatePending), string(longhorn.BackupStateInProgress):
 			job.logger.Infof("Creating backup %v, current progress %v", info.Id, info.Progress)
 		case string(longhorn.BackupStateError), string(longhorn.BackupStateUnknown):
 			return fmt.Errorf("failed to create backup %v: %v", info.Id, info.Error)
--- a/controller/backup_controller.go
+++ b/controller/backup_controller.go
@@ -820,8 +820,14 @@ func (bc *BackupController) checkMonitor(backup *longhorn.Backup, volume *longho
 
 	snapshot, err := bc.ds.GetSnapshotRO(backup.Spec.SnapshotName)
 	if err != nil {
-		backup.Status.State = longhorn.BackupStatePending
-		backup.Status.Messages[MessageTypeReconcileInfo] = fmt.Sprintf(FailedToGetSnapshotMessage, backup.Spec.SnapshotName)
+		if apierrors.IsNotFound(err) {
+			backup.Status.Error = fmt.Sprintf("Snapshot %v not found", backup.Spec.SnapshotName)
+			backup.Status.State = longhorn.BackupStateError
+			backup.Status.LastSyncedAt = metav1.Time{Time: time.Now().UTC()}
+		} else {
+			backup.Status.State = longhorn.BackupStatePending
+			backup.Status.Messages[MessageTypeReconcileInfo] = fmt.Sprintf(FailedToGetSnapshotMessage, backup.Spec.SnapshotName)
+		}
 		return nil, errors.Wrapf(err, "failed to get the snapshot %v before enabling backup monitor", backup.Spec.SnapshotName)
 	}
 	if snapshot != nil {


derekbit:
> fix: backup with non-existing snapshot should be marked as errored

Is it possible the snapshot shows up later, but we mark it as Error?

ChanYiLin:
Hi @PhanLe1010 
I think the `Pending` state is still necessary
The reason I introduce this state at the time was because he e2e tests failed frequently before due to creating snapshot and backup almost at the same time.
The backup failed because snapshot CR was not created.  

That is, originally, the backup will already fail if snapshot is not found because backup grpc call will fail when engine can't find the snapshot. 

I just discussed with @derekbit earlier, we think adding a max retry count to the backup creation is reasonable. WDYT?

PhanLe1010:
> Is it possible the snapshot shows up later, but we mark it as Error?

Hi @derekbit 
The recurring job creates snapshot CR first ([link](https://github.com/longhorn/longhorn-manager/blob/fd2ef9c5ee891fdbc787e475359a48b3f6b0e094/app/recurring_job.go#L620-L622)) then create backup CR ([link](https://github.com/longhorn/longhorn-manager/blob/fd2ef9c5ee891fdbc787e475359a48b3f6b0e094/app/recurring_job.go#L638-L644)). How can snapshot CR shows up after the backup CR?

PhanLe1010:
> The reason I introduce this state at the time was because he e2e tests failed frequently before due to creating snapshot and backup almost at the same time.
> The backup failed because snapshot CR was not created.


Regarding to the e2e test, looks like the reason that snapshot CR appear after backup CR is because e2e test is using deprecated API [SnapshotCreate](https://github.com/longhorn/longhorn-manager/blob/fd2ef9c5ee891fdbc787e475359a48b3f6b0e094/api/snapshot.go#L24-L52) instead of the new API [SnapshotCRCreate](https://github.com/longhorn/longhorn-manager/blob/fd2ef9c5ee891fdbc787e475359a48b3f6b0e094/api/snapshot.go#L218-L247) (note that recurring job is already using new API) . Ref [an e2e test](https://github.com/longhorn/longhorn-tests/blob/cd2fb66e5d0c89d920f8accb64a99d469fb99f97/manager/integration/tests/test_basic.py#L2040-L2042)

To this point, I am ok with either:
1. Keep this PR and fixing the e2e test to use new API SnapshotCRCreate 
2. OR have the retry logic like @ChanYiLin mentioned. 

Which approach do you prefer @derekbit @ChanYiLin ?

---

> That is, originally, the backup would already fail if snapshot is not found because backup grpc call would fail when engine couldn't find the snapshot.

I don't understand this point  @ChanYiLin . I don't see how snapshot info is missing in the engine now that we always wait for the snapshot to be ready before taking backup. This PR still keeps this behavior.

derekbit:
> > Is it possible the snapshot shows up later, but we mark it as Error?
> 
> Hi @derekbit The recurring job creates snapshot CR first ([link](https://github.com/longhorn/longhorn-manager/blob/fd2ef9c5ee891fdbc787e475359a48b3f6b0e094/app/recurring_job.go#L620-L622)) then create backup CR ([link](https://github.com/longhorn/longhorn-manager/blob/fd2ef9c5ee891fdbc787e475359a48b3f6b0e094/app/recurring_job.go#L638-L644)). How can snapshot CR shows up after the backup CR?

Got it. I thought it creates a snapshot instance first and the snapshot CR is created later. After checking the recurring job flow, we don't need to retry.

derekbit:

derekbit:
@ChanYiLin Can you help review?

ChanYiLin (approved):
LGTM

ChanYiLin:
> I don't understand this point @ChanYiLin . I don't see how snapshot info is missing in the engine now that we always wait for the snapshot to be ready before taking backup. This PR still keeps this behavior.

Hi @PhanLe1010 
I am just saying that before my fix [Lh8650 make backup wait for snapshot ](https://github.com/longhorn/longhorn-manager/pull/2835/files)
The backup would fail since we didn't check if the snapshot was ready to use before
In that case, even if the snapshot was found, but the engine didn't complete the snapshot and didn't have the latest snapshot info
The backup grpc call to the engine would fail because the snapshot was not found in the engine with the message.
```
could not find snapshot 'system-backup-49d05cf04e624d85' to backup, engine 'longhorn-testvol-jway75-e-0'
```
And my fix was not for e2e test but for system backup. (Sorry for the misleading.)
System backup once had code like this
```
func (c *SystemBackupController) createVolumeBackup(...)...{
    snapshotCR := &longhorn.Snapshot{...}
    snapshot, err := c.ds.CreateSnapshot(snapshotCR)
    if err != nil {...}
    backup = &longhorn.Backup{....}
    backup, err = c.ds.CreateBackup(backup, volume.Name)
    if err != nil {...}
}
```
In this case, even the CR was created first, but the snapshot was not created in the engine yet. The backup would still fail because of snapshot not found in the engine as described above.
This system backup code has been fixed as well to make sure snapshot is ready to use before creating backup [here](https://github.com/longhorn/longhorn-manager/pull/3246)

ChanYiLin:
After discussed with @derekbit today, follow the context [above](https://github.com/longhorn/longhorn-manager/pull/3437#issuecomment-2579402485) 

There is still a chance that the backup will be stuck forever if the snapshot CR is never readyToUse
```
	snapshot, err := bc.ds.GetSnapshotRO(backup.Spec.SnapshotName)
	if err != nil {
			backup.Status.Error = fmt.Sprintf("Snapshot %v not found", backup.Spec.SnapshotName)
			backup.Status.State = longhorn.BackupStateError
			backup.Status.LastSyncedAt = metav1.Time{Time: time.Now().UTC()}
		} else {
			backup.Status.State = longhorn.BackupStatePending
			backup.Status.Messages[MessageTypeReconcileInfo] = fmt.Sprintf(FailedToGetSnapshotMessage, backup.Spec.SnapshotName)
		}
		return nil, errors.Wrapf(err, "failed to get the snapshot %v before enabling backup monitor", backup.Spec.SnapshotName)
	}
	if snapshot != nil {
		if !snapshot.Status.ReadyToUse {
			backup.Status.State = longhorn.BackupStatePending
			backup.Status.Messages[MessageTypeReconcileInfo] = fmt.Sprintf(WaitForSnapshotMessage, backup.Spec.SnapshotName)
			return nil, fmt.Errorf("waiting for the snapshot %v to be ready before enabling backup monitor", backup.Spec.SnapshotName)
		}
	}
``` 

We believe that we still need to have some max retry mechanism for waiting the snapshot to be ready to use. WDYT?

cc @derekbit

PhanLe1010:
> There is still a chance that the backup will be stuck forever if the snapshot CR is never readyToUse
> We believe that we still need to have some max retry mechanism for waiting the snapshot to be ready to use. WDYT?

Hi @ChanYiLin 

Thank you for the additional context. Yeah, there is a chance that this might happen:
1. Snapshot CR is created
2. Snapshot is taken
3. Snapshot CR becomes readyToUse
4. Somehow something deletes the snapshot using old API which remove the actual snapshot but doesn't remove the snapshot CR
5. Snapshot CR become not readyToUse
6. Backup CR stuck in pending

Do you have any other senerios that this can happen? 

Although this sequence should be rare, it might happen. I agree that we should have max retry. Let's me create a PR for max retry.

PhanLe1010:
Implementing the retry mechanism. Pushing PR soon

PhanLe1010:
Pushed draft PR https://github.com/longhorn/longhorn-manager/pull/3442 @derekbit @ChanYiLin

PR #3442, opened 2025-01-10
fix: add retry max retry mechanism for backup creation

PhanLe1010:
Prevent the backup CR from being stuck in the pending state forever. Backup creation will only be retried up until max retry

longhorn/longhorn#10090

### DIFF
--- a/controller/backup_controller.go
+++ b/controller/backup_controller.go
@@ -41,7 +41,9 @@ const (
 
 const (
 	WaitForSnapshotMessage                 = "Waiting for the snapshot %v to be ready"
+	FailedWaitingForSnapshotMessage        = "Failed waiting for the snapshot %v to be ready"
 	WaitForEngineMessage                   = "Waiting for the engine %v to be ready"
+	FailedWaitingForEngineMessage          = "Failed waiting for the engine %v to be ready"
 	WaitForBackupDeletionIsCompleteMessage = "Wait for backup %v to be deleted"
 	FailedToGetSnapshotMessage             = "Failed to get the Snapshot %v"
 	FailedToDeleteBackupMessage            = "Failed to delete the backup %v in the backupstore, err %v"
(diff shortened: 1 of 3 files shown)


derekbit:

controller/backup_controller.go
```diff
@@ -832,6 +854,14 @@ func (bc *BackupController) checkMonitor(backup *longhorn.Backup, volume *longho
 	}
 	if snapshot != nil {
 		if !snapshot.Status.ReadyToUse {
+			bc.creationRetryCounter.IncreaseCount(backup.Name)
+			if bc.creationRetryCounter.GetCount(backup.Name) >= maxCreationRetry {
+				backup.Status.Error = fmt.Sprintf(FailedWaitingForSnapshotMessage, backup.Spec.SnapshotName)
+				backup.Status.State = longhorn.BackupStateError
```
derekbit, line 860: If it is marked as Error, when will the volumeattachment be deleted?

PhanLe1010: Yes, the `error` state is [a final state](https://github.com/longhorn/longhorn-manager/blob/6ab807644f15ec6c5941eeebe3a93780613e238b/controller/backup_controller.go#L999-L1004) so VA ticket will be deleted here https://github.com/longhorn/longhorn-manager/blob/6ab807644f15ec6c5941eeebe3a93780613e238b/controller/backup_controller.go#L380-L382

derekbit: Thank you.

derekbit (approved):
LGTM. Thanks @PhanLe1010

derekbit (approved):

controller/backup_controller.go
```diff
@@ -813,17 +824,28 @@ func (bc *BackupController) checkMonitor(backup *longhorn.Backup, volume *longho
 	if engine.Status.CurrentState != longhorn.InstanceStateRunning ||
 		engine.Spec.DesireState != longhorn.InstanceStateRunning ||
 		volume.Status.State != longhorn.VolumeStateAttached {
+		bc.creationRetryCounter.IncreaseCount(backup.Name)
+		if bc.creationRetryCounter.GetCount(backup.Name) >= maxCreationRetry {
```
james-munson, line 828: It would have simplified things a little if IncreaseCount just returned the current count, which can be checked.  Generally don't need the extra lock, map traversal, and unlock to turn right around and get it after increment.

PhanLe1010: Good suggestion. But PR is already merged so I will keep it as it is for now. Is it ok?

james-munson: Sure.  We can improve it if we have reason to touch it again.

PR #3445, opened 2025-01-10
fix(upgrade): directly add backup target label `default` to old backups

mantissahz:

Issue # longhorn/longhorn#10181

### DIFF
--- a/upgrade/v17xto180/upgrade.go
+++ b/upgrade/v17xto180/upgrade.go
@@ -133,18 +133,8 @@ func upgradeBackups(namespace string, lhClient *lhclientset.Clientset, resourceM
 	}
 
 	for _, b := range backupMap {
-		backupTargetName := types.DefaultBackupTargetName
-		vol, err := lhClient.LonghornV1beta2().Volumes(namespace).Get(context.TODO(), b.Status.VolumeName, metav1.GetOptions{})
-		if err != nil {
-			if !apierrors.IsNotFound(err) {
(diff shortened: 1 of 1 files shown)


upgrade/v17xto180/upgrade.go
```diff
@@ -133,18 +133,8 @@ func upgradeBackups(namespace string, lhClient *lhclientset.Clientset, resourceM
...
-		if err != nil {
-			if !apierrors.IsNotFound(err) {
-				return errors.Wrapf(err, "failed to get volume %v of backup %v", b.Status.VolumeName, b.Name)
-			}
-		} else {
-			if vol.Spec.BackupTargetName != "" {
-				backupTargetName = vol.Spec.BackupTargetName
-			}
-		}
-		b.Labels = addLabel(b.Labels, types.LonghornLabelBackupTarget, backupTargetName)
```
derekbit, line 147: Why doesn't the original solution work?

mantissahz: Because the 'b.Status.VolumeName' of the failed backup is an empty string and it is not allowed to get a volume CR by an empty string name.

derekbit: What’s the returned error of the Volume GET function if the volume name is empty?

mantissahz: > What’s the returned error of the Volume GET function if the volume name is empty?

'resource may not be empty'

derekbit:
@mergify backport v1.8.x

derekbit (approved):

PR #3463, opened 2025-01-12
fix: backupTargetController.bsTimerMap should be protected by lock

derekbit:

Issue Longhorn/longhorn#10190

There are 5 workers in each controller, so the map bsTimerMap should be protected by lock.


derekbit:
@mergify backport v1.8.x

innobead (approved):
LGTM

Good catch!

PR #3471, opened 2025-01-13
fix(backup): modify backup APIs for backup target and volume name

mantissahz:

Issue # longhorn/longhorn#10182

Ensure backups have backup target name and volume name in the
returning information of API `BackupList` and `BackupGet`.
Backups in state `pending`, `in-progress`, and `error`
will not have the backup target name and volume name in `b.Status`.

### DIFF
--- a/api/model.go
+++ b/api/model.go
@@ -1951,6 +1951,19 @@ func toBackupResource(b *longhorn.Backup) *Backup {
 		return nil
 	}
 
+	// For frontend, it will use the backup target name and volume name to tell this backup is belong to which backup volume.
+	// The backup not in `completed` state will not have the backup target name and volume name in `Status`.
+	backupTargetName := b.Status.BackupTargetName
+	volumeName := b.Status.VolumeName
+	if b.Labels != nil {
+		if backupTargetName == "" {
+			backupTargetName = b.Labels[types.LonghornLabelBackupTarget]
+		}
+		if volumeName == "" {
+			volumeName = b.Labels[types.LonghornLabelBackupVolume]
+		}
+	}
+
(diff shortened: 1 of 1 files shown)


derekbit:
Title and body are broken

api/model.go
```diff
@@ -1951,6 +1951,15 @@ func toBackupResource(b *longhorn.Backup) *Backup {
...
 	}
 
+	backupTargetName := b.Status.BackupTargetName
+	if backupTargetName == "" {
+		backupTargetName = b.Labels[types.LonghornLabelBackupTarget]
+	}
+	volumeName := b.Status.VolumeName
+	if volumeName == "" {
+		volumeName = b.Labels[types.LonghornLabelBackupVolume]
+	}
```
derekbit, line 1961: Why can `b.Status.BackupTargetName` and `b.Status.VolumeName` be empty?

a110605:
Thanks fixing in backend. 

Test locally looks good.
Screenshot 2025-01-13 at 11 43 45 AM

api/model.go
derekbit, line 1961: Check if `b.Labels` is empty before accessing it.

derekbit: Discussed with @mantissahz. A failed backup might have empty values because the information cannot be retrieved from backend.

mantissahz:
>Title and body are broken

Modified. Thanks.

api/model.go
mantissahz: Modified. Thanks

mantissahz:
@Mergifyio backport v1.8.x

derekbit (approved):
LGTM

derekbit:
@mergify backport v1.8.x

PR #3474, opened 2025-01-14
fix(backup): return null if backup volume is not found

mantissahz:

Issue # longhorn/longhorn#5411, longhorn/longhorn#9987, longhorn/longhorn#10024


derekbit (approved):
LGTM

PR #3476, opened 2025-01-14
feat(recurring-job): support system backup task

c3y1huang:

Issue longhorn/longhorn#6534

Support system backup task in the recurring job.

`None`

`None`


c3y1huang:
@coderabbitai review

c3y1huang:
Regression test result: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/8271/

k8s/crds.yaml
```diff
@@ -19,8 +19,7 @@ spec:
   scope: Namespaced
   versions:
   - additionalPrinterColumns:
-    - description: The current state of the pod used to provision the backing image
-        file from source
+    - description: The current state of the pod used to provision the backing image file from source
       jsonPath: .status.currentState
```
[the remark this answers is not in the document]
c3y1huang: I am not sure, but this is the output I got after running [k8s/generate_code.sh](https://github.com/longhorn/longhorn-manager/blob/master/k8s/generate_code.sh). Could you try running the script on your end and let me know if it comes out differently?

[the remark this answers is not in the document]
james-munson: Interesting.  I get this output:
```
jmunson@JMunson:~/repos/Longhorn/longhorn-manager$ ./k8s/generate_code.sh 
controller-gen is missing
Prepare to install controller-gen
go: sigs.k8s.io/controller-tools@v0.15.0 requires go >= 1.22.0; switching to go1.22.10
kustomize is missing
Prepare to install kustomize
~/go/src/github.com/kubernetes-sigs ~/repos/Longhorn/longhorn-manager
~/repos/Longhorn/longhorn-manager
Generating deepcopy funcs
Generating clientset for longhorn:v1beta1,v1beta2 at github.com/longhorn/longhorn-manager/k8s/pkg/client/clientset
Generating listers for longhorn:v1beta1,v1beta2 at github.com/longhorn/longhorn-manager/k8s/pkg/client/listers
Generating informers for longhorn:v1beta1,v1beta2 at github.com/longhorn/longhorn-manager/k8s/pkg/client/informers
Generating CRD
./k8s/generate_code.sh: line 68: controller-gen: command not found
```
And it doesn't touch crds.yaml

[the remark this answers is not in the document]
james-munson:
```
jmunson@JMunson:~/repos/Longhorn/longhorn-manager$ go version
go version go1.23.4 linux/amd64
```

app/recurringjob/systembackup.go
```diff
@@ -0,0 +1,147 @@
...
+		if err != nil {
+			return err
+		}
+
+		if lhutils.Contains(expectedStates, systemBackup.Status.State) {
+			return nil
+		}
+
+		job.logger.Infof("Waiting for system backup to reach state %v, current state is %v", expectedStates, systemBackup.Status.State)
+		time.Sleep(5 * time.Second)
```
james-munson, line 145: Would it be reasonable to use the named constant WaitInterval here?

james-munson (approved):
LGTM.  A clean refactoring.  One other small nit.

k8s/crds.yaml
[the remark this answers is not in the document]
james-munson: Seems like this whole question will be resolved by https://github.com/longhorn/longhorn/issues/10193.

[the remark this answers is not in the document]
c3y1huang: > When I run it, I get the multi-line descriptions

The reason I am getting the descriptions in a single line seems to be caused by my `kustomize` version.

app/recurringjob/systembackup.go
c3y1huang: Updated, thank you.

app/recurringjob/job.go
```diff
@@ -0,0 +1,131 @@
...
+
+	config, err := rest.InClusterConfig()
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get client config")
+	}
+	lhClient, err := lhclientset.NewForConfig(config)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get clientset")
+	}
+
```
mantissahz, line 40: Could we use `GetLonghornClientset` here?

NIT:
Before the func `NewJob` creates a new job, the func `recurringJob` will also use `GetLonghornClientset` to update the `RecurringJob`.
Is there any way to reuse the `lhClient` or update the `RecurringJob` in the func `NewJob`?

mantissahz (approved):
In general, LGTM.
Only one nit question.

mantissahz (approved):

app/recurringjob/constant.go
```diff
@@ -0,0 +1,17 @@
...
+	SnapshotPurgeStatusInterval = 5 * time.Second
+	// SnapshotPurgeStatusTimeout is set to 24 hours because we don't know the appropriate value.
+	SnapshotPurgeStatusTimeout = 24 * time.Hour
+
+	WaitInterval              = 5 * time.Second
+	DetachingWaitInterval     = 10 * time.Second
+	VolumeAttachTimeout       = 300 // 5 minutes
+	BackupProcessStartTimeout = 90  // 1.5 minutes
+	SnapshotReadyTimeout      = 390 // 6.5 minutes
+)
```
derekbit, line 17: Looks good! Thanks for the refactoring.

app/recurringjob/job.go
```diff
@@ -0,0 +1,126 @@
...
+	label := fmt.Sprintf("%s=%s", labelKey, job.name)
+
+	job.logger.Infof("Getting SystemBackups by label %v", label)
+	systemBackupList, err := job.lhClient.LonghornV1beta2().SystemBackups(job.namespace).List(context.TODO(), metav1.ListOptions{
+		LabelSelector: label,
+	})
+	if err != nil {
+		return nil, err
+	}
+	return systemBackupList, nil
```
derekbit, line 125: Why not 

```suggestion
	return job.lhClient.LonghornV1beta2().SystemBackups(job.namespace).List(context.TODO(), metav1.ListOptions{
		LabelSelector: label,
	})
```

app/recurringjob/systembackup.go
```diff
@@ -0,0 +1,147 @@
...
+	finalStates := []longhorn.SystemBackupState{
+		longhorn.SystemBackupStateReady,
+		longhorn.SystemBackupStateError,
+	}
+	err = job.waitForSystemBackupToStates(finalStates)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
derekbit, line 101:
```suggestion
	return job.waitForSystemBackupToStates(finalStates)
```

app/recurringjob/systembackup.go
```diff
@@ -0,0 +1,147 @@
...
+		}
+
+		if lhutils.Contains(expectedStates, systemBackup.Status.State) {
+			return nil
+		}
+
+		job.logger.Infof("Waiting for system backup to reach state %v, current state is %v", expectedStates, systemBackup.Status.State)
+		time.Sleep(WaitInterval)
+	}
+}
```
derekbit, line 147: Is it possible an infinite loop? Should we consider adding a maximum loop count?

app/recurringjob/type.go
```diff
@@ -0,0 +1,60 @@
...
+
+	eventRecorder record.EventRecorder // Used to record events related to the job.
+	logger        logrus.FieldLogger   // Log messages related to the job.
+
+	name           string                    // Name for the RecurringJob.
+	namespace      string                    // Kubernetes namespace in which the RecurringJob is running.
+	retain         int                       // Number of task CRs to retain.
+	task           longhorn.RecurringJobType // Type of task to be executed.
+	parameters     map[string]string         // Additional parameters for the task.
+	executionCount int                       // Number of times the job has been executed.
```
derekbit, line 28: Comments for the fields look good.

app/recurringjob/volume.go
```diff
@@ -0,0 +1,651 @@
...
+		return job.doRecurringSnapshot()
+	}
+}
+
+func (job *VolumeJob) doRecurringSnapshot() (err error) {
+	defer func() {
+		err = errors.Wrap(err, "failed recurring snapshot")
+		if err == nil {
+			job.logger.Info("Finished recurring snapshot")
+		}
```
derekbit, line 183: Can we log the error here?

webhook/resources/recurringjob/mutator.go
```diff
@@ -52,6 +52,9 @@ func (r *recurringJobMutator) Create(request *admission.Request, newObj runtime.
 	if name != recurringjob.Name {
 		patchOps = append(patchOps, fmt.Sprintf(`{"op": "replace", "path": "/metadata/name", "value": "%s"}`, name))
 	}
+	if recurringjob.Spec.Name == "" {
+		patchOps = append(patchOps, fmt.Sprintf(`{"op": "replace", "path": "/spec/name", "value": "%s"}`, recurringjob.Name))
+	}
```
derekbit, line 57: Why do we need the spec.Name?

c3y1huang: IIRC it's a placeholder for the full name before auto-correction.

app/recurringjob/volume.go
c3y1huang: Isn't this already being logged by the caller?
```
	err = volumeJob.run()
	if err != nil {
		volumeJob.logger.WithError(err).Error("Failed to run volume job")
		return err
	}
```

app/recurringjob/systembackup.go
c3y1huang: I intentionally designed it this way because the time required for a system backup depends on the cluster, making it challenging to determine a reasonable maximum duration - especially if volume backups are involved.

Having this as a blocking operation shouldn't harm the recurring system backup. Even if it gets stuck, users can remove it and let the cronjob recreate it.

I prefer to wait and see if there's any feedback from the community. What do you think?

app/recurringjob/volume.go
derekbit: Oh, yes. The error and info are logged in different places..

app/recurringjob/systembackup.go
derekbit: Sure.

webhook/resources/recurringjob/mutator.go
derekbit, line 57: Last question: should we update the `/spec/name` in the upgrade path for the existing jobs?

c3y1huang: Is there any reason it might be empty during the upgrade?

derekbit (approved):

PR #3477, opened 2025-01-15
fix: volume backup is regarded as not up to date if snapshot is not found

derekbit:

Issue Longhorn/longhorn#10215

### DIFF
--- a/controller/system_backup_controller.go
+++ b/controller/system_backup_controller.go
@@ -920,6 +920,10 @@ func (c *SystemBackupController) isVolumeBackupUpToDate(volume *longhorn.Volume,
 
 	lastBackupSnapshot, err := c.ds.GetSnapshot(lastBackup.Status.SnapshotName)
 	if err != nil {
+		if apierrors.IsNotFound(err) {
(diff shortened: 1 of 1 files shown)


controller/system_backup_controller.go
```diff
@@ -920,6 +920,10 @@ func (c *SystemBackupController) isVolumeBackupUpToDate(volume *longhorn.Volume,
 
 	lastBackupSnapshot, err := c.ds.GetSnapshot(lastBackup.Status.SnapshotName)
 	if err != nil {
+		if apierrors.IsNotFound(err) {
+			log.Warnf("Snapshot %v not found for backup %v", lastBackup.Status.SnapshotName, lastBackup.Name)
+			return false, nil
+		}
```
mantissahz, line 926: can we add the IsNotFound Error check below for the last backup?

```diff
lastBackup, err := c.ds.GetBackup(volume.Status.LastBackup)
if err != nil {
+	if apierrors.IsNotFound(err) {
+			log.Warnf("Last Backup %v not found", lastBackup.Name)
+			return false, nil
+		}
		return false, err
}
```

PR #3490, opened 2025-01-17
fix: use volume state to check all volumes detached when changing some settings

james-munson:

Issue https://github.com/longhorn/longhorn/issues/10233

Revert to checking volume state rather than engine instances when changing some settings.
Rename the engine instance check to reflect what it actually does.

Regression tests for changing settings passed:  https://ci.longhorn.io/job/private/job/longhorn-tests-regression/8332/

### DIFF
--- a/controller/setting_controller.go
+++ b/controller/setting_controller.go
@@ -300,7 +300,7 @@ func (sc *SettingController) syncDangerZoneSettingsForManagedComponents(settingN
 			}
 		case types.SettingNameStorageNetwork:
 			funcPreupdate := func() error {
-				detached, _, err := sc.ds.AreAllVolumesDetached(longhorn.DataEngineTypeAll)
+				detached, err := sc.ds.AreAllVolumesDetachedState()
 				if err != nil {
 					return errors.Wrapf(err, "failed to check volume detachment for %v setting update", types.SettingNameStorageNetwork)
 				}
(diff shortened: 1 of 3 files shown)


controller/setting_controller.go
```diff
@@ -1258,12 +1258,12 @@ func (sc *SettingController) updateInstanceManagerCPURequest(dataEngine longhorn
...
+	stopped, _, err := sc.ds.AreAllEngineInstancesStopped(dataEngine)
 	if err != nil {
-		return errors.Wrapf(err, "failed to check volume detachment for %v setting update", settingName)
+		return errors.Wrapf(err, "failed to check engine instances for %v setting update", settingName)
 	}
-	if !detached {
-		return &types.ErrorInvalidState{Reason: fmt.Sprintf("failed to apply %v setting to Longhorn components when there are attached volumes. It will be eventually applied", settingName)}
+	if !stopped {
+		return &types.ErrorInvalidState{Reason: fmt.Sprintf("failed to apply %v setting to Longhorn components when there are running engine instances. It will be eventually applied", settingName)}
 	}
```
PhanLe1010, line 1267: Any idea why we use `AreAllEngineInstancesStopped` instead of `AreAllVolumesDetachedState` ? Any convention around this choice?

PhanLe1010 (commented):
In general LGTM.

However, I think that this change should be carefully monitored in the e2e test and negative test. It possible this might introduce regression bug

controller/setting_controller.go
shuo-wu: Same question.

PhanLe1010: Thanks @james-munson

shuo-wu (approved):

PhanLe1010:
Will defer to @derekbit to take a look first

derekbit:
@james-munson LGTM. Could you run full regression and e2e test before merging it? Thanks.

james-munson:
Full regression run going in https://ci.longhorn.io/job/private/job/longhorn-tests-regression/8383/  Looks good so far.

james-munson:
Tests passed.  I think we're good to go.

derekbit (approved):

derekbit:

PR #3505, opened 2025-01-22
feat(backup): delete snapshot immediately  after backup completed

mantissahz:

Issue # longhorn/longhorn#9213

- Introduce a new boolean field `Spec.CleanupBackupSnasphot` to check if it needs to delete the snapshot after the backup is completed when doing a recurring job.
- Introduce a new field `Spec.CleanupSnapshot` to specify the backup will delete the snapshot after the backup is completed.

### DIFF
--- a/controller/backup_controller.go
+++ b/controller/backup_controller.go
@@ -499,6 +499,27 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
 		return nil
 	}
 
+	defer func() {
+		// If the backup is in the final state, delete the snapshot if needed
+		if _, ok := backupInfo.Labels[types.RecurringJobLabel]; ok {
+			// leave the snapshot management for recurring jobs to the `SettingNameAutoCleanupRecurringJobBackupSnapshot` setting.
+			return
+		}
+
+		cleanupSnap, err := bc.ds.GetSettingAsBool(types.SettingNameAutoCleanupSnapshotAfterOnDemandBackupCompleted)
+		if err != nil {
+			log.WithError(err).Errorf("Failed to get the setting %v", types.SettingNameAutoCleanupSnapshotAfterOnDemandBackupCompleted)
+			return
+		}
+		if cleanupSnap && backup.Spec.SnapshotName != "" {
+			if errDel := bc.ds.DeleteSnapshot(backup.Spec.SnapshotName); errDel != nil && !apierrors.IsNotFound(errDel) {
+				err = errDel
+				log.WithError(err).Error("Failed to delete the snapshot")
+				return
+			}
+		}
+	}()
+
 	// Remove the Backup Volume recurring jobs/groups information.
 	// Only record the latest recurring jobs/groups information in backup volume CR and volume.cfg on remote backup target.
 	delete(backupInfo.Labels, types.VolumeRecurringJobInfoLabel)
@@ -521,7 +542,7 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
 	backup.Status.LastSyncedAt = syncTime
 	backup.Status.NewlyUploadedDataSize = backupInfo.NewlyUploadedDataSize
 	backup.Status.ReUploadedDataSize = backupInfo.ReUploadedDataSize
-	return nil
+	return err
 }
 
 // handleAttachmentTicketDeletion check and delete attachment so that the source volume is detached if needed
--- a/types/setting.go
+++ b/types/setting.go
@@ -137,6 +137,7 @@ const (
 	SettingNameV2DataEngineFastReplicaRebuilding                        = SettingName("v2-data-engine-fast-replica-rebuilding")
 	SettingNameFreezeFilesystemForSnapshot                              = SettingName("freeze-filesystem-for-snapshot")
 	SettingNameAutoCleanupSnapshotWhenDeleteBackup                      = SettingName("auto-cleanup-when-delete-backup")
+	SettingNameAutoCleanupSnapshotAfterOnDemandBackupCompleted          = SettingName("auto-cleanup-snapshot-after-on-demand-backup-completed")
 	SettingNameDefaultMinNumberOfBackingImageCopies                     = SettingName("default-min-number-of-backing-image-copies")
 	SettingNameBackupExecutionTimeout                                   = SettingName("backup-execution-timeout")
 	SettingNameRWXVolumeFastFailover                                    = SettingName("rwx-volume-fast-failover")
@@ -236,6 +237,7 @@ var (
 		SettingNameDisableSnapshotPurge,
 		SettingNameFreezeFilesystemForSnapshot,
 		SettingNameAutoCleanupSnapshotWhenDeleteBackup,
+		SettingNameAutoCleanupSnapshotAfterOnDemandBackupCompleted,
 		SettingNameDefaultMinNumberOfBackingImageCopies,
 		SettingNameBackupExecutionTimeout,
 		SettingNameRWXVolumeFastFailover,
@@ -357,6 +359,7 @@ var (
 		SettingNameDisableSnapshotPurge:                                     SettingDefinitionDisableSnapshotPurge,
 		SettingNameFreezeFilesystemForSnapshot:                              SettingDefinitionFreezeFilesystemForSnapshot,
 		SettingNameAutoCleanupSnapshotWhenDeleteBackup:                      SettingDefinitionAutoCleanupSnapshotWhenDeleteBackup,
+		SettingNameAutoCleanupSnapshotAfterOnDemandBackupCompleted:          SettingDefinitionAutoCleanupSnapshotAfterOnDemandBackupCompleted,
 		SettingNameDefaultMinNumberOfBackingImageCopies:                     SettingDefinitionDefaultMinNumberOfBackingImageCopies,
 		SettingNameBackupExecutionTimeout:                                   SettingDefinitionBackupExecutionTimeout,
 		SettingNameRWXVolumeFastFailover:                                    SettingDefinitionRWXVolumeFastFailover,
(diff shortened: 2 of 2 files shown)


derekbit:
cc @c3y1huang @PhanLe1010 @shuo-wu

k8s/pkg/apis/longhorn/v1beta2/backup.go
```diff
@@ -51,6 +51,9 @@ type BackupSpec struct {
 	// Can be "full" or "incremental"
 	// +optional
 	BackupMode BackupMode `json:"backupMode"`
+	// Whether to cleanup the snapshot after the backup is done.
```
derekbit, line 54:
```suggestion
	// Whether to cleanup the snapshot after the backup is completed.
```

controller/backup_controller.go
```diff
@@ -499,6 +499,14 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
 		return nil
 	}
 
+	defer func() {
+		if backup.Spec.CleanupSnapshot {
```
derekbit, line 503: Can we introduce a global setting for the snapshot cleanup?

k8s/pkg/apis/longhorn/v1beta2/backup.go
mantissahz: Removed the new field. Thanks.

controller/backup_controller.go
mantissahz: Introduced a global setting and simplified the flow. Thanks.

types/setting.go
```diff
@@ -1480,6 +1483,16 @@ var (
 		Default:     "false",
 	}
 
+	SettingDefinitionAutoCleanupSnapshotAfterBackupCompleted = SettingDefinition{
+		DisplayName: "Automatically Cleanup Snapshot After Backup Completed",
+		Description: "This setting allows users to trigger automatically delete the backup snapshot after the backup is completed.",
```
derekbit, line 1488: It seems the setting is applied for all backup regardless of an on-demand or recurring job backup. Might it conflict with `Automatically Cleanup Recurring Job Backup Snapshot` setting?

mantissahz: I think it will not conflict with the `Automatically Cleanup Recurring Job Backup Snapshot` setting.

| BackupCompleted | RecurringJob | Result |
| --- | --- | --- |
| True | True | All snapshots for backups (after `AutoCleanupSnapshotAfterBackupCompleted` is set to true ) will be deleted. |
| True | False | All snapshots for new backups will be deleted, and (retain number -1) snapshots created by the recurring job (before `AutoCleanupSnapshotAfterBackupCompleted` is set to true) will be kept. |
| False | True | Only the snapshot for the latest backup will be kept. |
| False | False | The retain number of snapshots for backups will be kept. |

The controversial part is when this setting `AutoCleanupSnapshotAfterBackupCompleted` is set to true.

| BackupCompleted | RecurringJob | Explaination |
| --- | --- | --- |
| True | False | Users want to delete the snapshot after the backup is completed, and the recurring job is not responsible for deleting the snapshot for the backup. It is reasonable that it keeps only old snapshots for previous backups created by the recurring job after the `AutoCleanupSnapshotAfterBackupCompleted` setting is set to true and new snapshots for backups will be removed.

mantissahz: Updated the setting `AutoCleanupSnapshotAfterBackupCompleted` to `AutoCleanupSnapshotAfterOnDemandBackupCompleted` and it is only used for the on-demand backups.

controller/backup_controller.go
```diff
@@ -499,6 +499,24 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
...
 	}
 
+	defer func() {
+		// If the backup is in the final state, delete the snapshot if needed
+		if _, ok := backupInfo.Labels[types.RecurringJobLabel]; ok {
+			// leave the snapshot management for recurring jobs to the `SettingNameAutoCleanupRecurringJobBackupSnapshot` setting.
+			return
+		}
+
+		cleaupupSnap, err := bc.ds.GetSettingAsBool(types.SettingNameAutoCleanupSnapshotAfterOnDemandBackupCompleted)
```
derekbit, line 509: typo
```suggestion
		cleanupSnap, err := bc.ds.GetSettingAsBool(types.SettingNameAutoCleanupSnapshotAfterOnDemandBackupCompleted)
```

controller/backup_controller.go
```diff
@@ -499,6 +499,24 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
...
+		if err != nil {
+			log.WithError(err).Errorf("Failed to get the setting %v", types.SettingNameAutoCleanupSnapshotAfterOnDemandBackupCompleted)
```
derekbit, line 511: Can we explicitly return here? Or, is the deletion best-effort?

controller/backup_controller.go
```diff
@@ -499,6 +499,24 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
...
+		}
+		if cleaupupSnap && backup.Spec.SnapshotName != "" {
+			if err := bc.ds.DeleteSnapshot(backup.Spec.SnapshotName); err != nil {
```
derekbit, line 514: Same question as above.
If we want to return an error
```suggestion
			err = bc.ds.DeleteSnapshot(backup.Spec.SnapshotName)
            if err != nil {
                ...
            }
```

types/setting.go
```diff
@@ -137,6 +137,7 @@ const (
 	SettingNameV2DataEngineFastReplicaRebuilding                        = SettingName("v2-data-engine-fast-replica-rebuilding")
 	SettingNameFreezeFilesystemForSnapshot                              = SettingName("freeze-filesystem-for-snapshot")
 	SettingNameAutoCleanupSnapshotWhenDeleteBackup                      = SettingName("auto-cleanup-when-delete-backup")
+	SettingNameAutoCleanupSnapshotAfterOnDemandBackupCompleted          = SettingName("auto-cleanup-after-on-demand-backup-completed")
```
derekbit, line 140: Just realized `auto-cleanup-when-delete-backup` misses `snapshot`...

```suggestion
	SettingNameAutoCleanupSnapshotAfterOnDemandBackupCompleted          = SettingName("auto-cleanup-snapshot-after-on-demand-backup-completed")
```

controller/backup_controller.go
mantissahz: Modified. Thanks.

controller/backup_controller.go
mantissahz: Yes, we should do `return` here.
Modified. Thanks.

controller/backup_controller.go
mantissahz: Yes, we should do `return` here.
Modified. Thanks.

types/setting.go
mantissahz: >Just realized auto-cleanup-when-delete-backup misses snapshot...

Yes, it is.

```suggestion
	SettingNameAutoCleanupSnapshotAfterOnDemandBackupCompleted          = SettingName("auto-cleanup-snapshot-after-on-demand-backup-completed")
```
Modified. Thanks.

controller/backup_controller.go
```diff
@@ -499,6 +499,26 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
...
+		cleanupSnap, err := bc.ds.GetSettingAsBool(types.SettingNameAutoCleanupSnapshotAfterOnDemandBackupCompleted)
+		if err != nil {
+			log.WithError(err).Errorf("Failed to get the setting %v", types.SettingNameAutoCleanupSnapshotAfterOnDemandBackupCompleted)
+			return
+		}
+		if cleanupSnap && backup.Spec.SnapshotName != "" {
+			if err := bc.ds.DeleteSnapshot(backup.Spec.SnapshotName); err != nil {
+				log.WithError(err).Error("Failed to delete the snapshot")
+				return
+			}
```
derekbit, line 518: The `err` will not pass to the returned value (err) of the function. You can use 

```
			err = bc.ds.DeleteSnapshot(backup.Spec.SnapshotName)
                         if err != nil {
mantissahz marked this conversation as resolved.
				log.WithError(err).Error("Failed to delete the snapshot")
				return
			}
```

mantissahz: Modified. Thanks.

derekbit (dismissed):
LGTM

ChanYiLin (changes requested):
One comment

controller/backup_controller.go
```diff
@@ -499,6 +499,27 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
...
+		}
+
+		cleanupSnap, err := bc.ds.GetSettingAsBool(types.SettingNameAutoCleanupSnapshotAfterOnDemandBackupCompleted)
+		if err != nil {
+			log.WithError(err).Errorf("Failed to get the setting %v", types.SettingNameAutoCleanupSnapshotAfterOnDemandBackupCompleted)
+			return
+		}
+		if cleanupSnap && backup.Spec.SnapshotName != "" {
+			err = bc.ds.DeleteSnapshot(backup.Spec.SnapshotName)
+			if err != nil {
```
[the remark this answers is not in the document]
mantissahz: Yes, we should ignore the `NotFound` error.
Modified. thanks.

ChanYiLin (approved):
LGTM

derekbit (dismissed):
LGTM

controller/backup_controller.go
```diff
@@ -499,6 +499,31 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
...
+			errDel := bc.ds.DeleteSnapshot(backup.Spec.SnapshotName)
+			if errDel != nil {
+				if apierrors.IsNotFound(errDel) {
+					return
+				}
+				err = errDel
+				log.WithError(err).Error("Failed to delete the snapshot")
+				return
+			}
+		}
```
ChanYiLin, line 524: how about change to the following structure, more readable
```
			if errDel := bc.ds.DeleteSnapshot(backup.Spec.SnapshotName); errDel != nil && apierrors.IsNotFound(errDel)  {
				err = errDel
				log.WithError(err).Error("Failed to delete the snapshot")
				return
			}
```

mantissahz: Sure. More simple. Modified. Thanks.

ChanYiLin (approved):
LGTM

derekbit (approved):

derekbit:
It is auto-merged, but @PhanLe1010 will help review the PR.

PhanLe1010:
The PR LGTM.

However, the PR does not seem to implement the design at the issue description

> Introduce a new boolean field Spec.CleanupBackupSnasphot to check if it needs to delete the snapshot after the backup is completed when doing a recurring job.
> Introduce a new field Spec.CleanupSnapshot to specify the backup will delete the snapshot after the backup is completed.

Can you confirm that it is intended? @mantissahz

mantissahz:
Hi @PhanLe1010, that was the outdated description for this PR, and I had updated it. The global setting is instead of the Backup.Spec field.

PhanLe1010:
Sounds good to me. Thanks @mantissahz 

> Hi @PhanLe1010, that was the outdated description for this PR, and I had updated it. The global setting is instead of the Backup.Spec field.

PR #3506, opened 2025-01-22
fix(upgrade): Handle Mutated Images

Khalid-Nowaf:

Issue https://github.com/longhorn/longhorn/issues/9638

The upgrade has a bug that prevents it from supporting K8s webhook mutation (which is widely adaptable for security reasons), which could alter the image's name.

The new mutated image can not be known before deployment, so there is no way to configure the helm char to adapt it.

Longhorn is not deployable now using the helm in such cases. So we need this fix to make it compatible. 

https://github.com/zarf-dev/zarf/blob/main/adr/0005-mutating-webhook.md


Khalid-Nowaf:
@coderabbitai full review

Khalid-Nowaf:
To test it, I have built a container `Khalid/longhorn-manager:v1.8.0-patch1`. Currently, it works with no issues.

upgrade/upgrade.go
```diff
@@ -352,12 +353,18 @@ func waitForOldLonghornManagersToBeFullyRemoved(namespace, managerImage string,
...
 func isOldManagerPod(pod corev1.Pod, managerImage string) (bool, string) {
+	runningPodVersion, exists := pod.GetLabels()["app.kubernetes.io/version"]
+
 	for _, container := range pod.Spec.Containers {
 		if container.Name == "longhorn-manager" {
-			if container.Image != managerImage {
+			// if the pod has a version label and the label does not match
+			// the image version, then the running pod (Manger) is old
+			// we are doing this since images can be mutated by webhooks
+			if exists && !strings.Contains(managerImage, runningPodVersion) || runningPodVersion == "" {
```
derekbit, line 363: Some thoughts

- This original condition check should be kept
   ```
      if container.Image != managerImage {
         return return true, container.Image
      } else {
          // Handle zarf case in this block
      }
```

- `managerImage` consists of image name and image tag. The tag might not have version string. Using `strings.Contains` is not a good solution.

derekbit:
@Khalid-Nowaf 
Feel free to reopen the ticket if you want to continue and revise the PR. Thanks.

PR #3508, opened 2025-01-23
fix(workflow): fix create-crd-update-pr-in-longhorn-repo.yml

derekbit:

Issue Longhorn/longhorn#10193

- Use CUSTOM_GITHUB_ACTION_TOKEN that has permissions to create PR in longhorn/longhorn.
- Fix committer for passing DOC check.
- Create PR after opening, editing or synchronizing the longhorn-manager PR.

### DIFF
--- a/.github/workflows/create-crd-update-pr-in-longhorn-repo.yml
+++ b/.github/workflows/create-crd-update-pr-in-longhorn-repo.yml
@@ -10,6 +10,8 @@ on:
 jobs:
   create-pr:
     runs-on: ubuntu-latest
+    env:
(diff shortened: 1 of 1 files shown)


derekbit:
```
With the exception of GITHUB_TOKEN, secrets are not passed to the runner when a workflow is triggered from a forked repository. The GITHUB_TOKEN has read-only permissions in pull requests from forked repositories. For more information, see "Automatic token authentication."
```


The workflow is triggered by the forked repo, so the secrets of the repo cannot pass to the workflow.

PR #3512, opened 2025-01-24
fix(workflow): fix DCO check

derekbit:

Issue Longhorn/longhorn#10193

Use head commit's author name and email as committer's information.

### DIFF
--- a/.github/workflows/create-crd-update-pr-in-longhorn-repo.yml
+++ b/.github/workflows/create-crd-update-pr-in-longhorn-repo.yml
@@ -37,6 +37,14 @@ jobs:
         bash scripts/generate-longhorn-yaml.sh
         bash scripts/helm-docs.sh
 
+    - name: Get Head Commit Name
+      id: get_head_commit_name
+      run: echo "::set-output name=commit_name::$(git log -1 --pretty=format:'%an')"
+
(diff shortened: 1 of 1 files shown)


.github/workflows/create-crd-update-pr-in-longhorn-repo.yml
```diff
@@ -46,9 +54,8 @@ jobs:
         delete-branch: true
         sign-commits: true
         signoff: true
-        author: ${{ github.actor }} <${{ github.actor }}@users.noreply.github.com>
-        committer: Longhorn GitHub Bot <67932897+longhorn-io-github-bot@users.noreply.github.com>
```
innobead, line 50: Do we know why the bot account with sign is not valid for DCO check?

derekbit: I have no idea for now. Although I set the committer and author to Longhorn bot, the longhorn/longhorn PR will be created by the **author** of the longhorn/longhorn-manager PR rather than the Longhorn bot when triggering the workflow by `pull_request_target`. However, using the **author** of the longhorn/longhorn-manager PR might make sense.

innobead (approved):

PR #3519, opened 2025-01-31
fix: do not panic when looking up BackupVolume of non-existing Volume

PhanLe1010:
1. One Volume can have multiple BackupVolumes, need to check all BackupVolumes when finding backups
2. It is totally valid to look up BackupVolumes of non-existing Volume. A deleted Volume can still have Backups in the cluster. Do not panic

Issue # https://github.com/longhorn/longhorn/issues/10303

Putting this PR as draft. Need @mantissahz to help to review the multiple backup target logic

### DIFF
--- a/api/backup.go
+++ b/api/backup.go
@@ -317,7 +317,7 @@ func (s *Server) BackupGet(w http.ResponseWriter, req *http.Request) error {
 	}
 	backupVolumeName := mux.Vars(req)["backupVolumeName"]
 
-	backup, err := s.m.GetBackup(input.Name, backupVolumeName)
+	backup, err := s.m.GetBackup(input.Name)
 	if err != nil {
 		return errors.Wrapf(err, "failed to get backup '%v' of volume '%v'", input.Name, backupVolumeName)
 	}
@@ -344,13 +344,13 @@ func (s *Server) BackupDelete(w http.ResponseWriter, req *http.Request) error {
 
 	backupVolumeName := mux.Vars(req)["backupVolumeName"]
 
-	backup, err := s.m.GetBackup(input.Name, backupVolumeName)
+	backup, err := s.m.GetBackup(input.Name)
 	if err != nil {
 		logrus.WithError(err).Warnf("failed to get backup '%v' of volume '%v'", input.Name, backupVolumeName)
 	}
 
 	if backup != nil {
-		if err := s.m.DeleteBackup(input.Name, backupVolumeName); err != nil {
+		if err := s.m.DeleteBackup(input.Name); err != nil {
 			return errors.Wrapf(err, "failed to delete backup '%v' of volume '%v'", input.Name, backupVolumeName)
 		}
 	}
--- a/csi/controller_server.go
+++ b/csi/controller_server.go
@@ -126,17 +126,25 @@ func (cs *ControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 					if id == "" {
 						return nil, status.Errorf(codes.NotFound, "volume source snapshot %v is not found", snapshot.SnapshotId)
 					}
-					backupVolume, backupName := sourceVolumeName, id
-					bv, err := cs.getBackupVolume(backupVolume)
+					backupName := id
+					bvs, err := cs.getBackupVolumes(sourceVolumeName)
 					if err != nil {
-						return nil, status.Errorf(codes.Internal, "failed to retrieve backup volume %s: %v", backupVolume, err)
+						return nil, status.Errorf(codes.Internal, "failed to retrieve backup volumes of volume %v: %v", sourceVolumeName, err)
 					}
-					if bv == nil {
-						return nil, status.Errorf(codes.NotFound, "failed to restore CSI snapshot %s backup volume %s unavailable", snapshot.SnapshotId, backupVolume)
+					if len(bvs) == 0 {
+						return nil, status.Errorf(codes.NotFound, "failed to restore CSI snapshot %v of volume %v: there is no backup volume", snapshot.SnapshotId, sourceVolumeName)
 					}
-
-					backup, err := cs.apiClient.BackupVolume.ActionBackupGet(bv, &longhornclient.BackupInput{Name: backupName})
-					if err != nil {
+					var backup *longhornclient.Backup
+					for _, bv := range bvs {
+						backup, err = cs.apiClient.BackupVolume.ActionBackupGet(bv, &longhornclient.BackupInput{Name: backupName})
+						if err != nil {
+							return nil, status.Errorf(codes.NotFound, "failed to restore CSI snapshot %v : failed to get backup %v: %v", snapshot.SnapshotId, backupName, err)
+						}
+						if backup != nil {
+							break
+						}
+					}
+					if backup == nil {
 						return nil, status.Errorf(codes.NotFound, "failed to restore CSI snapshot %v backup %s unavailable", snapshot.SnapshotId, backupName)
 					}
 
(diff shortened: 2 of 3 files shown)


csi/controller_server.go
```diff
@@ -268,7 +273,8 @@ func (cs *ControllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 	}, nil
 }
 
-func (cs *ControllerServer) getBackupVolume(volumeName string) (*longhornclient.BackupVolume, error) {
+func (cs *ControllerServer) getBackupVolume(volumeName string) ([]*longhornclient.BackupVolume, error) {
```
james-munson, line 276: Even with the return signature changed, I think the function name should also change to `getBackupVolumes`

derekbit: Agree with @james-munson. The function name is confusing.

PhanLe1010: I agree. Fixed. Thank you

derekbit:
@PhanLe1010 It is still a draft. Ready for review?

csi/controller_server.go
```diff
@@ -1071,17 +1077,20 @@ func (cs *ControllerServer) cleanupSnapshot(sourceVolumeName, id string) error {
...
 		return err
 	}
-	if backupVolume != nil && backupVolume.Name != "" {
-		if _, err = cs.apiClient.BackupVolume.ActionBackupDelete(backupVolume, &longhornclient.BackupInput{
-			Name: backupName,
-		}); err != nil {
-			return err
+	for _, bv := range backupVolumes {
+		if bv.Name != "" {
+			if _, err = cs.apiClient.BackupVolume.ActionBackupDelete(bv, &longhornclient.BackupInput{
```
mantissahz, line 1086: Assume `backupVolumes: {"bv1", "b1", "bv2", "b2"}`, and `bv1` and `bv2` are from `volume1`.
Would it always return an error if we want to delete the backup `b2` with the first backup volume `bv1` and the backup `b2` can not be deleted?

PhanLe1010:
>  @PhanLe1010 It is still a draft. Ready for review?


I am waiting for @mantissahz to review the multiple backup logic first

csi/controller_server.go
PhanLe1010: I see. Thanks for the pointer. 
So the name `cleanupBackupVolume` is misleading because it is deleting `backup` instead of `BackupVolume` ? 

cc @ChanYiLin Could you take a look too see if it is ok to rename is to deletebackup?

mantissahz:
>@derekbit I am waiting for @mantissahz to review the multiple backup logic first

@PhanLe1010, Which part did you mean or all multiple backup target logic?

csi/controller_server.go
```diff
@@ -1287,24 +1296,22 @@ func (cs *ControllerServer) waitForBackupControllerSync(volumeName, snapshotName
 // (and in particular, its name) as quickly as possible and in any state.
 func (cs *ControllerServer) getBackup(volumeName, snapshotName string) (*longhornclient.Backup, error) {
```
shuo-wu, line 1297: As we discussed yesterday, this function will return `nil, nil` if the backup is not found. I think if this behavior may be unexpected for some developers then leads to some NPE issues. Do we need to change this or at least add a comment for this behavior?

csi/controller_server.go
PhanLe1010: @mantissahz @shuo-wu 

> This function name is indeed misleading

Ok. I changed the function name from `cleanupBackupVolume` to `cleanupBackup`

>  And it seems that we do need to figure out if the current bv contains the targeting backup before sending ActionBackupDelete

I think we don't have to care whether a bv has the backup as explained in the commented section. WDYT?
```go
	for _, bv := range backupVolumes {
		// The BackupDelete API actually doesn't care about bv when deleting a backup.
		// The bv is there for backward compatibility.
		// Any bv will work
		if bv.Name != "" {
			_, err = cs.apiClient.BackupVolume.ActionBackupDelete(bv, &longhornclient.BackupInput{
				Name: backupName,
			})
			if err != nil {
				return err
			} else {
				return nil
			}
		}
	}
```

csi/controller_server.go
PhanLe1010: Thank you. I added the comment. 

Looks like we check for return backup to be different than `nil` everywhere we call this fuction so I would prefer to keep the current logic. Ref:
1. https://github.com/longhorn/longhorn-manager/blob/469a446b29588bdaf9104f343c585c8c3c1cf9ee/csi/controller_server.go#L811-L832
2. https://github.com/longhorn/longhorn-manager/blob/469a446b29588bdaf9104f343c585c8c3c1cf9ee/csi/controller_server.go#L1247-L1256
3. https://github.com/longhorn/longhorn-manager/blob/469a446b29588bdaf9104f343c585c8c3c1cf9ee/csi/controller_server.go#L1273-L1280

PhanLe1010:
> > @derekbit I am waiting for @mantissahz to review the multiple backup logic first
> 
> @PhanLe1010, Which part did you mean (CSI part?) or all multiple backup target logic?
> 
> IMO, This PR is good enough to handle the volumes that do not exist.

I mean this PR @mantissahz Thank you for the reviewing.

PhanLe1010:
Updated the PR to addressed all feebacks @mantissahz @shuo-wu @derekbit Could you check again?

csi/controller_server.go
```diff
@@ -1069,19 +1070,28 @@ func (cs *ControllerServer) cleanupSnapshot(sourceVolumeName, id string) error {
...
+	backupVolumes, err := cs.getBackupVolumes(backupVolumeName)
 	if err != nil {
 		return err
 	}
-	if backupVolume != nil && backupVolume.Name != "" {
-		if _, err = cs.apiClient.BackupVolume.ActionBackupDelete(backupVolume, &longhornclient.BackupInput{
-			Name: backupName,
-		}); err != nil {
-			return err
+	for _, bv := range backupVolumes {
```
mantissahz, line 1079: Can we leave a log that the backup `id` is not deleted if the list `backupVolumes` is empty?

PhanLe1010: If the backvolume is empty, Longhorn would have already cleanup all the backup. The comment might be misleading?

mantissahz: >If the backvolume is empty, Longhorn would have already cleanup all the backup

Agree.

csi/controller_server.go
```diff
@@ -1069,19 +1070,28 @@ func (cs *ControllerServer) cleanupSnapshot(sourceVolumeName, id string) error {
...
+		// Any bv will work
+		if bv.Name != "" {
+			_, err = cs.apiClient.BackupVolume.ActionBackupDelete(bv, &longhornclient.BackupInput{
+				Name: backupName,
+			})
+			if err != nil {
+				return err
+			} else {
+				return nil
+			}
```
mantissahz, line 1091:
```diff
-    } else {
-       return nil
-    }
+    }
+
+    break
}
```

I think we don't need `else` here as the robot mentioned.
or we can return `err` directly, WDYT?

PhanLe1010: Surely. Fixed. Thanks

csi/controller_server.go
innobead: Let's improve this later, need to tackle it eventually to prevent similar issues. 

cc @derekbit

innobead (commented):
LGTM, just one comment.

csi/controller_server.go
ChanYiLin: Yes this function is to delete the backup not the backupVolume
the name was misleading indeed, thanks for the fix.

api/backup.go
```diff
@@ -317,7 +317,7 @@ func (s *Server) BackupGet(w http.ResponseWriter, req *http.Request) error {
 	}
 	backupVolumeName := mux.Vars(req)["backupVolumeName"]
 
-	backup, err := s.m.GetBackup(input.Name, backupVolumeName)
+	backup, err := s.m.GetBackup(input.Name)
```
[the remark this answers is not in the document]
PhanLe1010: Hi @innobead It is still used for some warning log below to know which volume the backup belong to:
```go
	backup, err := s.m.GetBackup(input.Name)
	if err != nil {
		return errors.Wrapf(err, "failed to get backup '%v' of volume '%v'", input.Name, backupVolumeName)
	}
	if backup == nil {
		logrus.Warnf("cannot find backup '%v' of volume '%v'", input.Name, backupVolumeName)
		w.WriteHeader(http.StatusNotFound)
		return nil
	}
```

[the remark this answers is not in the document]
shuo-wu: I think it's fine to remove the bv name from these logs as well.

csi/controller_server.go
```diff
@@ -1065,19 +1066,27 @@ func (cs *ControllerServer) cleanupSnapshot(sourceVolumeName, id string) error {
...
-		if _, err = cs.apiClient.BackupVolume.ActionBackupDelete(backupVolume, &longhornclient.BackupInput{
-			Name: backupName,
-		}); err != nil {
-			return err
+	for _, bv := range backupVolumes {
+		// The BackupDelete API actually doesn't care about bv when deleting a backup.
+		// The bv is there for backward compatibility.
+		// Any bv will work
+		if bv.Name != "" {
+			_, err = cs.apiClient.BackupVolume.ActionBackupDelete(bv, &longhornclient.BackupInput{
```
shuo-wu, line 1080: Will there be an error if the bv does not contain the specified backup?

PhanLe1010: No, please see comment https://github.com/longhorn/longhorn-manager/pull/3519#discussion_r1941987374

api/backup.go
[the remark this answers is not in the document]
PhanLe1010: Please note that `backupVolumeName` is part of the API URL https://github.com/longhorn/longhorn-manager/blob/4e93392f152ea2c268a40d736529c3e441ef85c4/api/router.go#L147-L148 . So even if we don't use it here, client still has to send it to from the URL. 

Now, since backupVolumeName is always available as part of the URL, I think it would be better to print out as more info as possible for easier debugging as opposed to not print it out. WDYT @shuo-wu ?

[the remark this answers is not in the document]
shuo-wu: OK. Considering the debugging purpose. Let's retain it.

shuo-wu (approved):

PhanLe1010:
@mergify backport v1.8.x

PR #3523, opened 2025-02-04
fix: should find the volume name of a pv from pv.Spec.CSI.VolumeHandle

PhanLe1010:
pv is not always the same as pv.Spec.CSI.VolumeHandle and pv.Spec.CSI.VolumeHandle is always the same as longhorn volume name

longhorn/longhorn#10314

More details is at https://github.com/longhorn/longhorn/issues/10314#issuecomment-2632648534


derekbit (approved):
LGTM

derekbit (approved):

derekbit:
@mergify backport v1.6.x v1.7.x v1.8.x

PR #3529, opened 2025-02-06
fix: workload pod will be able to move to new node when backup operation is taking a long time

PhanLe1010:
Reverting the commit
https://github.com/longhorn/longhorn-manager/commit/473b6d4b29c16e42ddb5019769396a33d76d9d2c because I cannot see the reason why we need to refuse the csi detachment when the volume is currently attached for operation such as backup/snapshot

Futher more, reading the design decision at
https://github.com/longhorn/longhorn/issues/8735#issuecomment-2177309258, I think we don't really need that commit
(https://github.com/longhorn/longhorn-manager/commit/473b6d4b29c16e42ddb5019769396a33d76d9d2c)

longhorn/longhorn#10171


derekbit (approved):
Agree with the revert. LGTM.

derekbit:

PR #3532, opened 2025-02-06
fix: empty string of backup.status.size

derekbit:

Issue Longhorn/longhorn#10358

The data type of backup.status.size is string. If backup is failed and the size is not set to "0". The metrics collect will fail to parse it.

- The default value of backup.status.size should be "0"
- Update the existing backups with backup.status.size == "" to "0"

### DIFF
--- a/k8s/crds.yaml
+++ b/k8s/crds.yaml
@@ -970,6 +970,7 @@ spec:
                 description: The address of the replica that runs snapshot backup.
                 type: string
               size:
+                default: "0"
                 description: The snapshot size.
                 type: string
               snapshotCreatedAt:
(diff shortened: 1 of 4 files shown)


mantissahz (approved):
LGTM in general.

NIT:
  Do we need to set the default values for all values related to the size (VolumeSize, NewlyUploadedDataSize, ...) of the `b.Status`?

derekbit:
> LGTM in general.
> 
> NIT: Do we need to set the default values for all values related to the size (VolumeSize, NewlyUploadedDataSize, ...) of the `b.Status`?

Yes, we can. However, it is fine to keep it as it is, because the values are not parsed in other places.

derekbit:
@mergify backport v1.8.x v1.7.x v1.6.x

PR #3538, opened 2025-02-07
fix: update k8s patch yamls

derekbit:

Issue #


innobead:
Do we have an issue for this?

derekbit:
> Do we have an issue for this?

No, we just forgot to update it before. No issues for chart and manifest.

innobead (approved):

PR #3544, opened 2025-02-07
feat: self-contained container image build

COLDTURNIP:

Supports security build.

* Merge binary building process into container build as a separate building stage.
* Prepare for potential secure release

The secrets provided by EIO MUST NOT present in any visible output.

Guidance on achieving SLSA Level 3

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,25 +11,57 @@ jobs:
   build:
     name: Build binaries
     runs-on: ubuntu-latest
+    outputs:
+      version_major: ${{ steps.build_info.outputs.version_major }}
+      version_minor: ${{ steps.build_info.outputs.version_minor }}
+      version_patch: ${{ steps.build_info.outputs.version_patch }}
+      image_tag: ${{ steps.build_info.outputs.image_tag }}
+
     steps:
     - name: Checkout code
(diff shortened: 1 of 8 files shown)


.github/workflows/build.yml
```diff
@@ -11,26 +11,60 @@ jobs:
...
+        version_major=${version_major}
+        version_minor=${version_minor}
+        version_build=${version_build}
+        image_tag=${image_tag}
+        EOF
+
     # Build binaries
     - name: Run ci
-      run: make ci
+      run: make ci-workflow
```
COLDTURNIP, line 60: This new makefile target excludes the image building to avoid out-of-space error in the Github builder VM.

scripts/ci-workflow
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+set -e
+
+cd $(dirname $0)
+
+./build
+./validate
+./test -cover
```
derekbit, line 8: Can we reuse the scripts/ci and add a new parameter like `workflow`?
If `workflow` is `true`, then no need to run package.

derekbit: @COLDTURNIP Let's remove this file. Thank you.

derekbit (approved):

PR #3554, opened 2025-02-11
fix(metrics): don't parse backup.status.size with an empty string

derekbit:

Issue longhorn/longhorn#10358


derekbit:
@mergify backport v1.8.x v1.7.x v1.6.x

PR #3566, opened 2025-02-14
fix: data lost caused by Longhorn CSI plugin doing a wrong re-encryption of volume in rare race condition

PhanLe1010:
longhorn/longhorn#10416

## RCA 
This happens with encrypted volume only

While Longhorn CSI plugin is doing `NodeStageVolume`, if the Longhorn engine crash right before [GetDiskFormat](https://github.com/longhorn/longhorn-manager/blob/5f9ec86d18612afaa9f97c78cee3e34f7b653ad6/csi/node_server.go#L486) and recover quickly after that, the following race condition can happen:
1. When engine is crash, the engine removes the Longhorn device at `/dev/longhorn/volume-name`
2. Because the device path is gone, `getDiskFormat` returns empty value for `diskFormat` and `nil` error. Ref https://github.com/longhorn/longhorn-manager/blob/5f9ec86d18612afaa9f97c78cee3e34f7b653ad6/vendor/k8s.io/mount-utils/mount_linux.go#L687-L693
3. Now engine is recovered, so it recreates the device at `/dev/longhorn/volume-name`
4. However, due to `diskFormat` having empty value, Longhorn re-encrypts the volume which wipes out the all data. Ref https://github.com/longhorn/longhorn-manager/blob/5f9ec86d18612afaa9f97c78cee3e34f7b653ad6/csi/node_server.go#L513-L517
5. Finally, Longhorn cannot detect a filesystem for the volume any more so it reformat the filesystem
6. The root cause of the issue is that `getDiskFormat` is using `blkid`. The `blkid` command cannot differentiate 2 cases: the device doesn't exist VS the device doesn't have filesystem inside it

## Proposal

Use cryptsetup isLuks instead. It can differentiate between:
1. The device is already encrypted -> return exit code 0
2. The device is not encrypted -> return exit code 1
3. All other errors -> return exit code 4

Ref: https://gitlab.com/cryptsetup/cryptsetup/-/blob/main/FAQ.md?plain=1#L2848

With this proposal, we can safely make decision of only doing encryption in the 2nd case

We were discussing of the below ideas with @derekbit and @shuo-wu but it seems cannot remove the race condition 100% 

Idea 1: Check if the device path exist before doing GetDiskFormat
```go
_, err := os.stat(devicePath)
if err != nil {
    return err
}

formatMounter.GetDiskFormat(devicePath)
```
-> Cannot avoid the race condition as the device can still disappear before GetDiskFormat and re-appear after GetDiskFormat

Idea 2: Check if the device path exist after doing GetDiskFormat
```go
formatMounter.GetDiskFormat(devicePath)

_, err := os.stat(devicePath)
if err != nil {
    return err
}
```
-> Cannot avoid the race condition as the device can still disappear before GetDiskFormat and re-appear after GetDiskFormat

### DIFF
--- a/csi/crypto/crypto.go
+++ b/csi/crypto/crypto.go
@@ -8,11 +8,11 @@ import (
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 
-	"github.com/longhorn/longhorn-manager/types"
-
 	lhns "github.com/longhorn/go-common-libs/ns"
 	lhtypes "github.com/longhorn/go-common-libs/types"
 	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
+
+	"github.com/longhorn/longhorn-manager/types"
 )
 
 const (
@@ -85,12 +85,21 @@ func VolumeMapper(volume, dataEngine string) string {
 
 // EncryptVolume encrypts provided device with LUKS.
 func EncryptVolume(devicePath, passphrase string, cryptoParams *EncryptParams) error {
+	isEncrypted, err := isDeviceEncrypted(devicePath)
+	if err != nil {
+		logrus.WithError(err).Warnf("Failed to check IsDeviceEncrypted before encrypting volume %v", devicePath)
+		return err
+	}
+	if isEncrypted {
+		logrus.Infof("The device %v is already encrypted. Skipping the encryption to avoid data lost", devicePath)
+		return nil
+	}
+
 	namespaces := []lhtypes.Namespace{lhtypes.NamespaceMnt, lhtypes.NamespaceIpc}
 	nsexec, err := lhns.NewNamespaceExecutor(lhtypes.ProcessNone, lhtypes.HostProcDirectory, namespaces)
 	if err != nil {
 		return err
 	}
-
 	logrus.Infof("Encrypting device %s with LUKS", devicePath)
 	if _, err := nsexec.LuksFormat(
 		devicePath, passphrase,
@@ -180,6 +189,15 @@ func IsDeviceMappedToNullPath(device string) (bool, error) {
 func IsDeviceOpen(device string) (bool, error) {
 	_, mappedFile, err := DeviceEncryptionStatus(device)
 	return mappedFile != "", err
+}
+
+func isDeviceEncrypted(devicePath string) (bool, error) {
+	namespaces := []lhtypes.Namespace{lhtypes.NamespaceMnt, lhtypes.NamespaceIpc}
+	nsexec, err := lhns.NewNamespaceExecutor(lhtypes.ProcessNone, lhtypes.HostProcDirectory, namespaces)
+	if err != nil {
+		return false, err
+	}
+	return nsexec.IsLuks(devicePath, lhtypes.LuksTimeout)
 }
 
 // DeviceEncryptionStatus looks to identify if the passed device is a LUKS mapping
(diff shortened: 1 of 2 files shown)


csi/crypto/crypto.go
```diff
@@ -85,12 +84,21 @@ func VolumeMapper(volume, dataEngine string) string {
 
 // EncryptVolume encrypts provided device with LUKS.
 func EncryptVolume(devicePath, passphrase string, cryptoParams *EncryptParams) error {
+	isEncrypted, err := IsDeviceEncrypted(devicePath)
+	if err != nil {
```
shuo-wu, line 88: Based on your commit description, there will be an error if the device exists but is not encrypted
```
2. The device is not encrypted -> return exit code 1
```

Hence here do we need to figure out this error by ourselves (and continue the encryption)?

PhanLe1010: Yes, we figure it out and continue here https://github.com/longhorn/longhorn-manager/pull/3566/files#diff-54750e8b890e240ba5a42e35654428da7c3d67323eb4d8961f162b186460bb62R60-R64

derekbit: If not encrypted, IsDeviceEncrypted will return "false, nil", right?

csi/crypto/crypto.go
```diff
@@ -182,6 +190,15 @@ func IsDeviceOpen(device string) (bool, error) {
 	return mappedFile != "", err
 }
 
+func IsDeviceEncrypted(devicePath string) (bool, error) {
```
derekbit, line 193:
```suggestion
func isDeviceEncrypted(devicePath string) (bool, error) {
```

csi/crypto/crypto.go
```diff
@@ -7,12 +7,11 @@ import (
...
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
-
-	"github.com/longhorn/longhorn-manager/types"
-
+	
 	lhns "github.com/longhorn/go-common-libs/ns"
 	lhtypes "github.com/longhorn/go-common-libs/types"
 	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
+	"github.com/longhorn/longhorn-manager/types"
```
derekbit, line 14:
```suggestion

	"github.com/longhorn/longhorn-manager/types"
```

csi/crypto/crypto.go
PhanLe1010: > If not encrypted, IsDeviceEncrypted will return "false, nil", right?
Yes, you are correct. Then we proceed to encrypt the volume by `luksFormat` which will wipe out the data

csi/crypto/crypto.go
PhanLe1010: Thank you. Fixed

csi/crypto/crypto.go
PhanLe1010: Thank you. Fixed

derekbit (approved):
LGTM

PhanLe1010:
Btw, please let me know if you have better idea than the `cryptsetup isLuks` solution @derekbit @shuo-wu @c3y1huang @mantissahz

c3y1huang (dismissed):

derekbit:
@PhanLe1010 go fmt failed

PhanLe1010:
> @PhanLe1010 go fmt failed

Fixed

derekbit (approved):

derekbit:

PR #3596, opened 2025-02-20
fix: fix nil pointer if the record in the status but not in the spec

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/10464

There is a chance that the disk in the spec is deleted first, but the status is not reflected yet
Then when the node/disk is eviction requested, the lookup map is build from the status
And since we are going to check and update the spec.evictionRequested, It will hit the nil pointer issue.

disk is deleted from the spec means it is going to be deleted, we don't need to update/check the evictionRequested in this case.
skip it if it is not in the spec.


derekbit (approved):
LGTM

derekbit:

PR #3607, opened 2025-03-01
feat: pass backup mode through csi volume snapshot

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/9785

pass backup mode through csi volume snapshot

### DIFF
--- a/csi/controller_server.go
+++ b/csi/controller_server.go
@@ -705,9 +705,12 @@ func (cs *ControllerServer) createCSISnapshotTypeLonghornBackingImage(req *csi.C
 		}
 	}
 
-	exportType, exist := req.Parameters["export-type"]
+	exportType, exist := req.Parameters["exportType"]
 	if !exist {
-		exportType = "raw"
+		exportType, exist = req.Parameters["export-type"]
+		if !exist {
+			exportType = "raw"
+		}
 	}
 
 	if backingImage == nil {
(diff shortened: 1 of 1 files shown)


csi/controller_server.go
```diff
@@ -805,6 +805,11 @@ func (cs *ControllerServer) createCSISnapshotTypeLonghornBackup(req *csi.CreateS
 		return nil, status.Error(codes.InvalidArgument, "Snapshot name must be provided")
 	}
 
+	backupMode, exist := req.Parameters["backup-mode"]
```
derekbit, line 808: is it better to use `backupMode`?

ChanYiLin: Yes you are right.
Fixed.
Thanks

innobead (dismissed):

derekbit:
Onee more question: 
```
exportType, exist := req.Parameters["export-type"]
```

Should it be updated to `exportType`?
If yes, need to handle the compatibility.

ChanYiLin:
> Onee more question:
> 
> ```
> exportType, exist := req.Parameters["export-type"]
> ```
> 
> Should it be updated to `exportType`? If yes, need to handle the compatibility.

Hi @derekbit 
Yes, it should be `exportType`, fixed.

derekbit:
> >   
> > Should it be updated to `exportType`? If yes, need to handle the compatibility.
> 
> Hi @derekbit Yes, it should be `exportType`, fixed.

To address the compatibility, for an existing storageClass with `export-type`, should we allow to fall back to `export-type`?
Or, just mention the issue in important node and have users recreate a storageClass with `exportType`?

ChanYiLin:
I think we can allow to fall back in the next version 
And then have a note to notify users to recreate it
Then remove the fall back in the following version

derekbit:
> I think we can allow to fall back in the next version And then have a note to notify users to recreate it Then remove the fall back in the following version

Sounds good to me.
But it is existing for 2 years, no users mentioned this 😅

ChanYiLin:
Probably they all directly copy the example 😅

derekbit (approved):
LGTM.

derekbit:
Merged. Thanks @ChanYiLin

derekbit:
@ChanYiLin 
Merged, thanks. Let's move the ticket to the next stage.

PR #3613, opened 2025-03-04
feat: move SettingNameV2DataEngineHugepageLimit to danger zone settings

derekbit:

Issue Longhorn/longhorn#7746

The huge page setting for v2 IM should belong to danger zone settings as CPU requests, as it's related to resource change. It should be eventually applied in the same as other dangerous settings.


c3y1huang (approved):

derekbit:

PR #3615, opened 2025-03-05
feat(snapshot): skip snapshot calculation if volume's healthy replica count is less than 2

derekbit:

Issue Longhorn/longhorn#10518

Snapshot checksum is used for fast replica rebuilding and identify corrupted replicas. If the healthy replica count is less than 2, the checksum is useless.

### DIFF
--- a/controller/monitor/snapshot_monitor.go
+++ b/controller/monitor/snapshot_monitor.go
@@ -316,6 +316,16 @@ func (m *SnapshotMonitor) deleteFromInProgressSnapshotCheckTasks(snapshotName st
 	delete(m.inProgressSnapshotCheckTasks, snapshotName)
 }
 
+func getNumberOfHealthyReplicas(e *longhorn.Engine) int {
+	numOfHealthyReplicas := 0
+	for _, mode := range e.Status.ReplicaModeMap {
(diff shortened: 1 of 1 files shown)


controller/monitor/snapshot_monitor.go
```diff
@@ -332,6 +332,17 @@ func (m *SnapshotMonitor) run(arg interface{}) error {
 		return errors.Wrapf(err, "failed to get engine for volume %v", task.volumeName)
 	}
 
+	// Skip snapshot hashing if the volume has less than 2 replicas
+	volume, err := m.ds.GetVolumeRO(engine.Spec.VolumeName)
+	if err != nil {
+		m.logger.WithField("monitor", monitorName).WithError(err).Debugf("failed to get volume %v for snapshot calculation", engine.Spec.VolumeName)
+		return nil
+	}
+	if volume.Spec.NumberOfReplicas <= 1 {
```
shuo-wu, line 341: Can we check the healthy replica count instead?

derekbit: Addressed. Thank you.

c3y1huang (approved):

PR #3616, opened 2025-03-06
feat(k8s): deprecates v1beta1

COLDTURNIP:

Issue longhorn/longhorn#10250

We're going to retire the v1beta1 CRDs (#10249), and the migration in this PR is the prerequisite.

Since we marked v1beta2 as the storage version, and we already have [K8s builtin trivial converter](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/) and our customized conversion webhooks, we migrate the all existed resources to the new version by simply read in v1beta2 then write back WITHOUT modification.

- Mark all v1beta1 CRDs as deprecated
- Implement the migration logic from 1.8.x to 1.9.0

The following CRDs are included in this PR:

- BackingImage
- BackingImageDataSource
- BackingImageManager
- Backup
- BackupTarget
- BackupVolume
- Engine
- EngineImage
- InstanceManager
- Node
- RecurringJob
- Replica
- Setting
- ShareManager
- Volume

And the following CRDs are not included since they are new from v1beta2:

- BackupBackingImage
- Orphan
- Snapshot
- SupportBundle
- SystemBackup
- SystemRestore
- VolumeAttachment

But, in this PR, it will write all resource back from cache, include v1beta2 resources.

### DIFF
--- a/upgrade/util/util.go
+++ b/upgrade/util/util.go
@@ -832,9 +832,9 @@ func CreateAndUpdateBackingImageInProvidedCache(namespace string, lhClient *lhcl
 	return obj, nil
 }
 
-type untypedResourceUpdater func(namespace string, lhClient *lhclientset.Clientset, resourceMap any, forceWrite bool) error
+type untypedUpdateResourceFunc func(namespace string, lhClient *lhclientset.Clientset, resourceMap any, forceWrite bool) error
 
-func toUntypedResourceUpdater[K any](kind string, updater func(namespace string, lhClient *lhclientset.Clientset, resourceMap map[string]K, forceWrite bool) error) untypedResourceUpdater {
+func toUntypedUpdateResourceFunc[K any](kind string, updater func(namespace string, lhClient *lhclientset.Clientset, resourceMap map[string]K, forceWrite bool) error) untypedUpdateResourceFunc {
 	return func(namespace string, lhClient *lhclientset.Clientset, resourceMap any, forceWrite bool) error {
 		typedResourceMap, ok := resourceMap.(map[string]K)
 		if !ok {
@@ -844,27 +844,27 @@ func toUntypedResourceUpdater[K any](kind string, updater func(namespace string,
 	}
 }
 
-var resourceUpdaters = map[string]untypedResourceUpdater{
-	types.LonghornKindNode:                   toUntypedResourceUpdater(types.LonghornKindNode, updateNodes),
-	types.LonghornKindVolume:                 toUntypedResourceUpdater(types.LonghornKindVolume, updateVolumes),
-	types.LonghornKindEngine:                 toUntypedResourceUpdater(types.LonghornKindEngine, updateEngines),
-	types.LonghornKindReplica:                toUntypedResourceUpdater(types.LonghornKindReplica, updateReplicas),
-	types.LonghornKindBackupTarget:           toUntypedResourceUpdater(types.LonghornKindBackupTarget, updateBackupTargets),
-	types.LonghornKindBackupVolume:           toUntypedResourceUpdater(types.LonghornKindBackupVolume, updateBackupVolumes),
-	types.LonghornKindBackup:                 toUntypedResourceUpdater(types.LonghornKindBackup, updateBackups),
-	types.LonghornKindBackupBackingImage:     toUntypedResourceUpdater(types.LonghornKindBackupBackingImage, updateBackupBackingImages),
-	types.LonghornKindBackingImageDataSource: toUntypedResourceUpdater(types.LonghornKindBackingImageDataSource, updateBackingImageDataSources),
-	types.LonghornKindEngineImage:            toUntypedResourceUpdater(types.LonghornKindEngineImage, updateEngineImages),
-	types.LonghornKindInstanceManager:        toUntypedResourceUpdater(types.LonghornKindInstanceManager, updateInstanceManagers),
-	types.LonghornKindShareManager:           toUntypedResourceUpdater(types.LonghornKindShareManager, updateShareManagers),
-	types.LonghornKindBackingImage:           toUntypedResourceUpdater(types.LonghornKindBackingImage, updateBackingImages),
-	types.LonghornKindBackingImageManager:    toUntypedResourceUpdater(types.LonghornKindBackingImageManager, updateBackingImagesManager),
-	types.LonghornKindRecurringJob:           toUntypedResourceUpdater(types.LonghornKindRecurringJob, updateRecurringJobs),
-	types.LonghornKindSetting:                toUntypedResourceUpdater(types.LonghornKindSetting, updateSettings),
-	types.LonghornKindVolumeAttachment:       toUntypedResourceUpdater(types.LonghornKindVolumeAttachment, createOrUpdateVolumeAttachments),
-	types.LonghornKindSnapshot:               toUntypedResourceUpdater(types.LonghornKindSnapshot, updateSnapshots),
-	types.LonghornKindOrphan:                 toUntypedResourceUpdater(types.LonghornKindOrphan, updateOrphans),
-	types.LonghornKindSystemBackup:           toUntypedResourceUpdater(types.LonghornKindSystemBackup, updateSystemBackups),
+var resourceUpdaters = map[string]untypedUpdateResourceFunc{
+	types.LonghornKindNode:                   toUntypedUpdateResourceFunc(types.LonghornKindNode, updateNodes),
+	types.LonghornKindVolume:                 toUntypedUpdateResourceFunc(types.LonghornKindVolume, updateVolumes),
+	types.LonghornKindEngine:                 toUntypedUpdateResourceFunc(types.LonghornKindEngine, updateEngines),
+	types.LonghornKindReplica:                toUntypedUpdateResourceFunc(types.LonghornKindReplica, updateReplicas),
+	types.LonghornKindBackupTarget:           toUntypedUpdateResourceFunc(types.LonghornKindBackupTarget, updateBackupTargets),
+	types.LonghornKindBackupVolume:           toUntypedUpdateResourceFunc(types.LonghornKindBackupVolume, updateBackupVolumes),
+	types.LonghornKindBackup:                 toUntypedUpdateResourceFunc(types.LonghornKindBackup, updateBackups),
+	types.LonghornKindBackupBackingImage:     toUntypedUpdateResourceFunc(types.LonghornKindBackupBackingImage, updateBackupBackingImages),
+	types.LonghornKindBackingImageDataSource: toUntypedUpdateResourceFunc(types.LonghornKindBackingImageDataSource, updateBackingImageDataSources),
+	types.LonghornKindEngineImage:            toUntypedUpdateResourceFunc(types.LonghornKindEngineImage, updateEngineImages),
+	types.LonghornKindInstanceManager:        toUntypedUpdateResourceFunc(types.LonghornKindInstanceManager, updateInstanceManagers),
+	types.LonghornKindShareManager:           toUntypedUpdateResourceFunc(types.LonghornKindShareManager, updateShareManagers),
+	types.LonghornKindBackingImage:           toUntypedUpdateResourceFunc(types.LonghornKindBackingImage, updateBackingImages),
+	types.LonghornKindBackingImageManager:    toUntypedUpdateResourceFunc(types.LonghornKindBackingImageManager, updateBackingImagesManager),
+	types.LonghornKindRecurringJob:           toUntypedUpdateResourceFunc(types.LonghornKindRecurringJob, updateRecurringJobs),
+	types.LonghornKindSetting:                toUntypedUpdateResourceFunc(types.LonghornKindSetting, updateSettings),
+	types.LonghornKindVolumeAttachment:       toUntypedUpdateResourceFunc(types.LonghornKindVolumeAttachment, createOrUpdateVolumeAttachments),
+	types.LonghornKindSnapshot:               toUntypedUpdateResourceFunc(types.LonghornKindSnapshot, updateSnapshots),
+	types.LonghornKindOrphan:                 toUntypedUpdateResourceFunc(types.LonghornKindOrphan, updateOrphans),
+	types.LonghornKindSystemBackup:           toUntypedUpdateResourceFunc(types.LonghornKindSystemBackup, updateSystemBackups),
 }
 
 // UpdateResources persists all the resources' spec changes in provided cached `resourceMap`. This method is not thread-safe.
--- a/upgrade/util/util.go
+++ b/upgrade/util/util.go
@@ -712,6 +712,26 @@ func ListAndUpdateBackingImageDataSourcesInProvidedCache(namespace string, lhCli
 	return bidss, nil
 }
 
+// ListAndUpdateBackingImageManagersInProvidedCache list all backingImageManagers and save them into the provided cached `resourceMap`. This method is not thread-safe.
+func ListAndUpdateBackingImageManagersInProvidedCache(namespace string, lhClient *lhclientset.Clientset, resourceMaps map[string]interface{}) (map[string]*longhorn.BackingImageManager, error) {
+	if v, ok := resourceMaps[types.LonghornKindBackingImageManager]; ok {
+		return v.(map[string]*longhorn.BackingImageManager), nil
+	}
+
+	bims := map[string]*longhorn.BackingImageManager{}
+	bimList, err := lhClient.LonghornV1beta2().BackingImageManagers(namespace).List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		return nil, err
+	}
+	for i, bids := range bimList.Items {
+		bims[bids.Name] = &bimList.Items[i]
+	}
+
+	resourceMaps[types.LonghornKindBackingImageManager] = bims
+
+	return bims, nil
+}
+
 // ListAndUpdateRecurringJobsInProvidedCache list all recurringJobs and save them into the provided cached `resourceMap`. This method is not thread-safe.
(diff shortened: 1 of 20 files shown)


upgrade/util/util.go
```diff
@@ -812,63 +832,54 @@ func CreateAndUpdateBackingImageInProvidedCache(namespace string, lhClient *lhcl
...
+	types.LonghornKindRecurringJob:           toUntypedResourceUpdater(types.LonghornKindRecurringJob, updateRecurringJobs),
+	types.LonghornKindSetting:                toUntypedResourceUpdater(types.LonghornKindSetting, updateSettings),
+	types.LonghornKindVolumeAttachment:       toUntypedResourceUpdater(types.LonghornKindVolumeAttachment, createOrUpdateVolumeAttachments),
+	types.LonghornKindSnapshot:               toUntypedResourceUpdater(types.LonghornKindSnapshot, updateSnapshots),
+	types.LonghornKindOrphan:                 toUntypedResourceUpdater(types.LonghornKindOrphan, updateOrphans),
+	types.LonghornKindSystemBackup:           toUntypedResourceUpdater(types.LonghornKindSystemBackup, updateSystemBackups),
+}
+
+// UpdateResources persists all the resources' spec changes in provided cached `resourceMap`. This method is not thread-safe.
+func UpdateResources(namespace string, lhClient *lhclientset.Clientset, resourceMaps map[string]interface{}, forceWrite bool) error {
```
COLDTURNIP, line 871: Refactor to reduce single function complexity complained by CodeFactor

upgrade/util/util.go
```diff
@@ -1009,7 +1027,7 @@ func updateBackups(namespace string, lhClient *lhclientset.Clientset, backups ma
 	return nil
 }
 
-func updateBackupBackingImages(namespace string, lhClient *lhclientset.Clientset, backupBackingImages map[string]*longhorn.BackupBackingImage) error {
+func updateBackupBackingImages(namespace string, lhClient *lhclientset.Clientset, backupBackingImages map[string]*longhorn.BackupBackingImage, forceWrite bool) error {
```
COLDTURNIP, line 1030: Actually we don't need to force update the following resources during 1.8.x to 1.9.0 upgrade, but still add `forceWrite` here. It should have not side effect during runtime because we don't update the content of these CRs:

- BackupBackingImage
- Orphan
- Snapshot
- SupportBundle
- SystemBackup
- SystemRestore
- VolumeAttachment

derekbit:
Should set v1beta1 CRD's `served` to `false`. Please check https://github.com/longhorn/longhorn/issues/10250#issuecomment-2706026786.

upgrade/v18xto190/upgrade.go
```diff
@@ -0,0 +1,151 @@
...
+		upgradeEngineImages,
+		upgradeReplicas,
+		upgradeVolumes,
+		upgradeBackupVolumes,
+		upgradeBackups,
+		upgradeBackupTargets,
+		upgradeBackingImageManagers,
+		upgradeBackingImageDataSources,
+		upgradeBackingImages,
+		upgradeRecurringJobs,
```
derekbit, line 38: nit:
`upgradexxx` functions seem unnecessary wappers.

Can we use 
```
upgradeutil.ListAndUpdateInstanceManagersInProvidedCache,
upgradeutil.ListAndUpdateVolumesInProvidedCache,
...
...
...
```

derekbit:
One question, do we need to force update the status of the resources?

COLDTURNIP:
> One question, do we need to force update the status of the resources?

I don't think so. Status is part of the whole version schema. Earlier version migration and the K8s version converters should be able to cover this part.

upgrade/v18xto190/upgrade.go
```diff
@@ -6,36 +6,45 @@ import (
 	clientset "k8s.io/client-go/kubernetes"
 
 	lhclientset "github.com/longhorn/longhorn-manager/k8s/pkg/client/clientset/versioned"
+	"github.com/longhorn/longhorn-manager/types"
```
derekbit, line 9: Reorganize the imports.

upgrade/v18xto190/upgrade.go
```diff
@@ -6,36 +6,45 @@ import (
...
 	upgradeutil "github.com/longhorn/longhorn-manager/upgrade/util"
 )
 
-const (
-	upgradeLogPrefix = "upgrade from v1.8.x to v1.9.0: "
-)
+type resourceMapLoader func(namespace string, lhClient *lhclientset.Clientset, resourceMaps map[string]interface{}) error
```
derekbit, line 13: The name of the type is a bit weird and confusing and is unable easily realize what it does.

Suggest
```
listAndUpdateResourcesInProvidedCacheFunc
```

upgrade/v18xto190/upgrade.go
```diff
@@ -0,0 +1,55 @@
...
+
+	clientset "k8s.io/client-go/kubernetes"
+
+	lhclientset "github.com/longhorn/longhorn-manager/k8s/pkg/client/clientset/versioned"
+	"github.com/longhorn/longhorn-manager/types"
+	upgradeutil "github.com/longhorn/longhorn-manager/upgrade/util"
+)
+
+type resourceMapLoader func(namespace string, lhClient *lhclientset.Clientset, resourceMaps map[string]interface{}) error
+type typedResourceMapLoader[K any] func(namespace string, lhClient *lhclientset.Clientset, resourceMaps map[string]interface{}) (map[string]*K, error)
```
derekbit, line 14:
```suggestion
type typedListAndUpdateResourcesInProvidedCacheFunc[K any] func(namespace string, lhClient *lhclientset.Clientset, resourceMaps map[string]interface{}) (map[string]*K, error)
```

upgrade/v18xto190/upgrade.go
```diff
@@ -0,0 +1,55 @@
...
+
+func toUntypedResourceMapLoader[K any](kind string, listFunc typedResourceMapLoader[K]) resourceMapLoader {
```
derekbit, line 16: naming

upgrade/v18xto190/upgrade.go
```diff
@@ -0,0 +1,55 @@
...
+		toUntypedResourceMapLoader(types.LonghornKindBackingImageManager, upgradeutil.ListAndUpdateBackingImageManagersInProvidedCache),
+		toUntypedResourceMapLoader(types.LonghornKindBackingImageDataSource, upgradeutil.ListAndUpdateBackingImageDataSourcesInProvidedCache),
+		toUntypedResourceMapLoader(types.LonghornKindBackingImage, upgradeutil.ListAndUpdateBackingImagesInProvidedCache),
+		toUntypedResourceMapLoader(types.LonghornKindRecurringJob, upgradeutil.ListAndUpdateRecurringJobsInProvidedCache),
+	}
+	for _, updateResource := range updates {
+		if err := updateResource(namespace, lhClient, resourceMaps); err != nil {
+			return err
+		}
+	}
```
derekbit, line 53: naming

upgrade/util/util.go
```diff
@@ -812,63 +832,54 @@ func CreateAndUpdateBackingImageInProvidedCache(namespace string, lhClient *lhcl
 	return obj, nil
 }
 
-// UpdateResources persists all the resources' spec changes in provided cached `resourceMap`. This method is not thread-safe.
-func UpdateResources(namespace string, lhClient *lhclientset.Clientset, resourceMaps map[string]interface{}) error {
-	var err error
+type untypedResourceUpdater func(namespace string, lhClient *lhclientset.Clientset, resourceMap any, forceWrite bool) error
```
derekbit, line 835:
```suggestion
type untypedUpdateResourceFunc func(namespace string, lhClient *lhclientset.Clientset, resourceMap any, forceWrite bool) error
```

upgrade/util/util.go
```diff
@@ -812,63 +832,54 @@ func CreateAndUpdateBackingImageInProvidedCache(namespace string, lhClient *lhcl
...
 
+func toUntypedResourceUpdater[K any](kind string, updater func(namespace string, lhClient *lhclientset.Clientset, resourceMap map[string]K, forceWrite bool) error) untypedResourceUpdater {
```
derekbit, line 837:
```suggestion
func toUntypedUpdateResourceFunc[K any](kind string, updater func(namespace string, lhClient *lhclientset.Clientset, resourceMap map[string]K, forceWrite bool) error) untypedUpdateResourceFunc {
```

derekbit (commented):
The implementation is quite straightforward and simple, but the naming is a bit confusing and anti-pattern. Need to refine this part.

derekbit:
Remember to update https://github.com/longhorn/longhorn-manager/pull/3616#issue-2899218033
Update
```
Issue #10250
```
to 
```
Issue longhorn/longhorn#10250
```

COLDTURNIP:
Renamed and updated. Please help review again, thanks.

derekbit (dismissed):
LGTM

upgrade/v18xto190/upgrade.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/longhorn/longhorn-manager/types"
+
+	lhclientset "github.com/longhorn/longhorn-manager/k8s/pkg/client/clientset/versioned"
+	upgradeutil "github.com/longhorn/longhorn-manager/upgrade/util"
+)
+
+type listAndUpdateResourcesInProvidedCacheFunc func(namespace string, lhClient *lhclientset.Clientset, resourceMaps map[string]interface{}) error
+type typedListAndUpdateResourcesInProvidedCacheFunc[K any] func(namespace string, lhClient *lhclientset.Clientset, resourceMaps map[string]interface{}) (map[string]*K, error)
+
+func toListAndUpdateResourcesInProvidedCacheFunc[K any](listUpdateFunc typedListAndUpdateResourcesInProvidedCacheFunc[K]) listAndUpdateResourcesInProvidedCacheFunc {
```
c3y1huang, line 17: NIT: Long names can sometimes be hard to read. Consider keeping them short and simple:
```suggestion
type listAndUpdateFunc func(namespace string, lhClient *lhclientset.Clientset, resourceMaps map[string]interface{}) error
type typedListAndUpdateFunc[K any] func(namespace string, lhClient *lhclientset.Clientset, resourceMaps map[string]interface{}) (map[string]*K, error)

func listAndUpdateResources[K any](listUpdateFunc typedListAndUpdateResourcesInProvidedCacheFunc[K]) listAndUpdateResourcesInProvidedCacheFunc {
```
That said, I am also fine with the current naming.

derekbit: Looks better!

derekbit:
@COLDTURNIP, after resolving the comment https://github.com/longhorn/longhorn-manager/pull/3616#discussion_r1998344422, we can merge it.

COLDTURNIP:
@derekbit renaming is applied and now ready for merge.

COLDTURNIP:
@derekbit resolved. The PR is ready to merge now.

derekbit (approved):
LGTM

upgrade/v18xto190/upgrade.go
```diff
@@ -0,0 +1,151 @@
...
+const (
+	upgradeLogPrefix = "upgrade from v1.8.x to v1.9.0: "
+)
+
+func UpgradeResources(namespace string, lhClient *lhclientset.Clientset, kubeClient *clientset.Clientset, resourceMaps map[string]interface{}) error {
+	if resourceMaps == nil {
+		return errors.New("resourceMaps cannot be nil")
+	}
+
+	// From v1.9.0, the v1beta1 API is deprecated. Load all resource and write back into v1beta2.
```
innobead: Why do we need this step? As we already make v1beta2 as the storage version, so the resources persist in API server should be v1beta2 already.

innobead (changes requested):
In general, LGTM. Just one question needs to be clarified.

upgrade/v18xto190/upgrade.go
COLDTURNIP: The storage version of an object is determined at its [last update](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#writing-reading-and-updating-versioned-customresourcedefinition-objects), and not get upgraded automatically while changing the storage version on CRD:

> When an object is written, it is stored at the version designated as the storage version at the time of the write. If the storage version changes, existing objects are never converted automatically. However, newly-created or updated objects are written at the new storage version. It is possible for an object to have been written at a version that is no longer served.

Hence in this PR, we load all the existing resource then write back again, to ensure that all resources are written in our current storage version `v1beta2`.

shuo-wu: Is there a way to filter out the v1beta1 resources only? For a large cluster, this kind of blind list & upgrade may temporarily crash the API server.

innobead: It seems we are performing sequential updates, so the resource impact might be acceptable. However, slow performance may occur due to the inefficiency of sequential updates.

[/upgrade/util/util.go#L816-L869](https://github.com/longhorn/longhorn-manager/blob/f5ad885cc32edd66dc92cf525b9e2a233494ed34/upgrade/util/util.go#L816-L869)
```go
func UpdateResources(namespace string, lhClient *lhclientset.Clientset, resourceMaps map[string]interface{}) error {
	var err error

	for resourceKind, resourceMap := range resourceMaps {
		switch resourceKind {
		case types.LonghornKindNode:
			err = updateNodes(namespace, lhClient, resourceMap.(map[string]*longhorn.Node))
		case types.LonghornKindVolume:
			err = updateVolumes(namespace, lhClient, resourceMap.(map[string]*longhorn.Volume))
		case types.LonghornKindEngine:
			err = updateEngines(namespace, lhClient, resourceMap.(map[string]*longhorn.Engine))
		case types.LonghornKindReplica:
			err = updateReplicas(namespace, lhClient, resourceMap.(map[string]*longhorn.Replica))
		case types.LonghornKindBackupTarget:
			err = updateBackupTargets(namespace, lhClient, resourceMap.(map[string]*longhorn.BackupTarget))
		case types.LonghornKindBackupVolume:
			err = updateBackupVolumes(namespace, lhClient, resourceMap.(map[string]*longhorn.BackupVolume))
		case types.LonghornKindBackup:
			err = updateBackups(namespace, lhClient, resourceMap.(map[string]*longhorn.Backup))
		case types.LonghornKindBackupBackingImage:
			err = updateBackupBackingImages(namespace, lhClient, resourceMap.(map[string]*longhorn.BackupBackingImage))
		case types.LonghornKindBackingImageDataSource:
			err = updateBackingImageDataSources(namespace, lhClient, resourceMap.(map[string]*longhorn.BackingImageDataSource))
		case types.LonghornKindEngineImage:
			err = updateEngineImages(namespace, lhClient, resourceMap.(map[string]*longhorn.EngineImage))
		case types.LonghornKindInstanceManager:
			err = updateInstanceManagers(namespace, lhClient, resourceMap.(map[string]*longhorn.InstanceManager))
		case types.LonghornKindShareManager:
			err = updateShareManagers(namespace, lhClient, resourceMap.(map[string]*longhorn.ShareManager))
		case types.LonghornKindBackingImage:
			err = updateBackingImages(namespace, lhClient, resourceMap.(map[string]*longhorn.BackingImage))
		case types.LonghornKindRecurringJob:
			err = updateRecurringJobs(namespace, lhClient, resourceMap.(map[string]*longhorn.RecurringJob))
		case types.LonghornKindSetting:
			err = updateSettings(namespace, lhClient, resourceMap.(map[string]*longhorn.Setting))
		case types.LonghornKindVolumeAttachment:
			err = createOrUpdateVolumeAttachments(namespace, lhClient, resourceMap.(map[string]*longhorn.VolumeAttachment))
		case types.LonghornKindSnapshot:
			err = updateSnapshots(namespace, lhClient, resourceMap.(map[string]*longhorn.Snapshot))
		case types.LonghornKindOrphan:
			err = updateOrphans(namespace, lhClient, resourceMap.(map[string]*longhorn.Orphan))
		case types.LonghornKindSystemBackup:
			err = updateSystemBackups(namespace, lhClient, resourceMap.(map[string]*longhorn.SystemBackup))
		default:
			return fmt.Errorf("resource kind %v is not able to updated", resourceKind)
		}

		if err != nil {
			return err
		}
	}

	return nil
}
```

innobead (approved):
In general, LGTM.

upgrade/v18xto190/upgrade.go
derekbit: Do we need to benchmark upgrade progress while there are thousands of volumes?

COLDTURNIP: > Is there a way to filter out the v1beta1 resources only? For a large cluster, this kind of blind list & upgrade may temporarily crash the API server.

To a Kubernetes API client, include Kubectl and libs we are using, it is not possible to filter the resource by stored CRD version, and all the resources are collected then converted to a client desired version ([explain](https://github.com/longhorn/longhorn/issues/10250#issuecomment-2705508434), [doc](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#writing-reading-and-updating-versioned-customresourcedefinition-objects)). The way to verify the migration is to access the underlayer database directly (for example, Kine SQLite for K3s longhorn/longhorn-tests#2378).

COLDTURNIP:
Rebased and re-signed again.

derekbit:
Build failed

upgrade/v18xto190/upgrade.go
derekbit: Do we need to benchmark the upgrade process for thousands of volumes in cluster? @COLDTURNIP @shuo-wu @innobead

COLDTURNIP:
Build broken because of linter upgrade. Fixing.

COLDTURNIP:
Lint errors are fixed in another PR: longhorn/longhorn-manager#3649

COLDTURNIP:
Rebased to resolve linting problems.

derekbit (approved):

PR #3619, opened 2025-03-07
fix: reject snapshot purge when volume is being live migration

derekbit:

Issue Longhorn/longhorn#10495

Reject snapshot purge when volume is being live migration. If snapshot purge is executed while live migrating a volume, the disk files will be changed. The new engine and replicas are unable to be aware of the change and cannot list the new disk files, resulting in ERR replicas.

### DIFF
--- a/controller/backing_image_controller.go
+++ b/controller/backing_image_controller.go
@@ -1618,7 +1618,7 @@ func (bic *BackingImageController) getEngineClientProxy(nodeName string, dataEng
 	if err != nil {
 		return nil, errors.Wrapf(err, "failed to get the default instance manager for node %v", nodeName)
 	}
-	engineClientProxy, err := engineapi.NewEngineClientProxy(instanceManager, log, bic.proxyConnCounter)
+	engineClientProxy, err := engineapi.NewEngineClientProxy(instanceManager, log, bic.proxyConnCounter, bic.ds)
 	if err != nil {
 		return nil, errors.Wrapf(err, "failed to get the engine client proxy for instance manager %v", instanceManager.Name)
 	}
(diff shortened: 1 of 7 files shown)


derekbit:
Running regression test
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/8305/

datastore/longhorn.go
```diff
@@ -6140,3 +6140,12 @@ func (s *DataStore) IsStorageNetworkForRWXVolume() (bool, error) {
 
 	return types.IsStorageNetworkForRWXVolume(storageNetworkSetting, storageNetworkForRWXVolumeEnabled), nil
 }
+
+func (s *DataStore) IsMigrationInProgress(e *longhorn.Engine) (bool, error) {
```
shuo-wu, line 6144: There is already such a function `util.IsVolumeMigrating`

shuo-wu (commented):
NIT: I am thinking if it's better to add the migration check inside `engineClientProxy.SnapshotPurge` implementation

derekbit:
> NIT: I am thinking if it's better to add the migration check inside `engineClientProxy.SnapshotPurge` implementation

How about add a function `purgeSnapshot()` that checks the migration status and then execute `engineClientProxy.SnapshotPurge`?

derekbit:
After rethinking the implementation, adding the migration check in `engineClientProxy.SnapshotPurge implementation` is much simpler.

derekbit:
> Running regression test https://ci.longhorn.io/job/private/job/longhorn-tests-regression/8305/

Rerun in https://ci.longhorn.io/job/private/job/longhorn-tests-regression/8308/

c3y1huang (approved):

derekbit:
@mergify backport v1.7.x v1.8.x

PR #3620, opened 2025-03-07
chore(k8s): generalize K8s code generator

COLDTURNIP:

Issue #10556

Currently, the code generator works only when the longhorn-manager's source is placed under conventional Go source, i.e. `$HOME/go/src/github.com/longhorn/longhorn-manager`. This change make it possible to execute even when the source part of Go workspace.

### DIFF
--- a/k8s/generate_code.sh
+++ b/k8s/generate_code.sh
@@ -6,10 +6,11 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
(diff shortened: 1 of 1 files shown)


derekbit (approved):
LGTM

derekbit:
Rememeber to update 
```
Which issue(s) this PR fixes:
Issue #10556
```
to
```
Which issue(s) this PR fixes:
Issue longhorn/longhorn#10556
```

PR #3633, opened 2025-03-12
fix: fix reject message when deleting disk

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/10522

Backing images can also have disk storage allocated. We should mention it when reject deleting disk.


derekbit (approved):
LGTM

ChanYiLin:
@mergify backport v1.8.x

derekbit:
@mergify backport v1.8.x

PR #3648, opened 2025-03-25
feat(volume): offline replica rebuilding

mantissahz:

Issue # longhorn/longhorn#8443

- Introduce a global setting `offline-replica-rebuilding`
- Introduce the field `Volume.Spec.OfflineRebuild`
- Add APIs for the offline replica rebuilding
- Modify the setting and volume rebuild controllers to support volume offline rebuilding.

### DIFF
--- a/api/model.go
+++ b/api/model.go
@@ -75,9 +75,10 @@ type Volume struct {
 	CloneStatus      longhorn.VolumeCloneStatus    `json:"cloneStatus"`
 	Ready            bool                          `json:"ready"`
 
-	AccessMode    longhorn.AccessMode        `json:"accessMode"`
-	ShareEndpoint string                     `json:"shareEndpoint"`
-	ShareState    longhorn.ShareManagerState `json:"shareState"`
+	AccessMode        longhorn.AccessMode              `json:"accessMode"`
+	ShareEndpoint     string                           `json:"shareEndpoint"`
+	ShareState        longhorn.ShareManagerState       `json:"shareState"`
+	OfflineRebuilding longhorn.VolumeOfflineRebuilding `json:"offlineRebuilding"`
 
 	Migratable bool `json:"migratable"`
 
@@ -403,6 +404,10 @@ type UpdateBackupTargetInput struct {
 	BackupTargetName string `json:"backupTargetName"`
 }
 
+type UpdateOfflineRebuildingInput struct {
+	OfflineRebuilding string `json:"offlineRebuilding"`
+}
+
 type PVCreateInput struct {
 	PVName string `json:"pvName"`
 	FSType string `json:"fsType"`
@@ -664,6 +669,7 @@ func NewSchema() *client.Schemas {
 	schemas.AddType("UpdateReplicaDiskSoftAntiAffinityInput", UpdateReplicaDiskSoftAntiAffinityInput{})
 	schemas.AddType("UpdateFreezeFilesystemForSnapshotInput", UpdateFreezeFilesystemForSnapshotInput{})
 	schemas.AddType("UpdateBackupTargetInput", UpdateBackupTargetInput{})
+	schemas.AddType("UpdateOfflineRebuildingInput", UpdateOfflineRebuildingInput{})
 	schemas.AddType("workloadStatus", longhorn.WorkloadStatus{})
 	schemas.AddType("cloneStatus", longhorn.VolumeCloneStatus{})
 	schemas.AddType("empty", Empty{})
@@ -994,10 +1000,13 @@ func volumeSchema(volume *client.Schema) {
 		"cancelExpansion": {
 			Output: "volume",
 		},
+		"offlineReplicaRebuilding": {
+			Input:  "UpdateOfflineRebuildingInput",
+			Output: "volume",
+		},
 		"trimFilesystem": {
 			Output: "volume",
 		},
-
 		"snapshotPurge": {
 			Output: "volume",
 		},
@@ -1634,9 +1643,10 @@ func toVolumeResource(v *longhorn.Volume, ves []*longhorn.Engine, vrs []*longhor
 		DataEngine:                  v.Spec.DataEngine,
 		Ready:                       ready,
 
-		AccessMode:    v.Spec.AccessMode,
-		ShareEndpoint: v.Status.ShareEndpoint,
-		ShareState:    v.Status.ShareState,
+		AccessMode:        v.Spec.AccessMode,
+		ShareEndpoint:     v.Status.ShareEndpoint,
+		ShareState:        v.Status.ShareState,
+		OfflineRebuilding: v.Spec.OfflineRebuilding,
 
 		Migratable: v.Spec.Migratable,
 
@@ -1677,6 +1687,7 @@ func toVolumeResource(v *longhorn.Volume, ves []*longhorn.Engine, vrs []*longhor
 			actions["activate"] = struct{}{}
 			actions["expand"] = struct{}{}
 			actions["cancelExpansion"] = struct{}{}
+			actions["offlineReplicaRebuilding"] = struct{}{}
 			actions["replicaRemove"] = struct{}{}
 			actions["engineUpgrade"] = struct{}{}
 			actions["pvCreate"] = struct{}{}
@@ -1699,6 +1710,7 @@ func toVolumeResource(v *longhorn.Volume, ves []*longhorn.Engine, vrs []*longhor
 			actions["recurringJobList"] = struct{}{}
 		case longhorn.VolumeStateAttaching:
 			actions["cancelExpansion"] = struct{}{}
+			actions["offlineReplicaRebuilding"] = struct{}{}
 			actions["recurringJobAdd"] = struct{}{}
 			actions["recurringJobDelete"] = struct{}{}
 			actions["recurringJobList"] = struct{}{}
@@ -1729,6 +1741,7 @@ func toVolumeResource(v *longhorn.Volume, ves []*longhorn.Engine, vrs []*longhor
 			actions["pvCreate"] = struct{}{}
 			actions["pvcCreate"] = struct{}{}
 			actions["cancelExpansion"] = struct{}{}
+			actions["offlineReplicaRebuilding"] = struct{}{}
 			actions["trimFilesystem"] = struct{}{}
 			actions["recurringJobAdd"] = struct{}{}
 			actions["recurringJobDelete"] = struct{}{}
--- a/controller/volume_attachment_controller.go
+++ b/controller/volume_attachment_controller.go
@@ -600,15 +600,7 @@ func (vac *VolumeAttachmentController) shouldDoDetach(va *longhorn.VolumeAttachm
 	// Check if there is any workload ticket regardless of frontend on other nodes
 	// If exist, detach and interrupt the current ticket.
 	if !hasUninterruptibleTicket(currentAttachmentTickets) && hasWorkloadTicket(attachmentTicketsOnOtherNodes, longhorn.AnyValue) {
-		log.Info("Workload attachment ticket interrupted snapshot/backup/rebuilding-controller attachment tickets")
-		return true
-	}
-
-	// If there is an interruptible ticket and frontend disabled ticket on the current node (currently, only offline rebuilding ticket)
-	// need to check if there is any workload ticket with frontend enabled (disableFrontend=false) on the same node.
-	// If exist, detach and interrupt the rebuilding ticket.
-	if hasInterruptibleAndFrontendDisabledTicket(currentAttachmentTickets) && hasWorkloadTicket(currentAttachmentTickets, longhorn.FalseValue) {
-		log.Info("Workload attachment ticket interrupted rebuilding-controller attachment tickets")
+		log.Info("Workload attachment ticket interrupted snapshot/backup attachment tickets")
 		return true
 	}
 
@@ -618,17 +610,7 @@ func (vac *VolumeAttachmentController) shouldDoDetach(va *longhorn.VolumeAttachm
 func hasUninterruptibleTicket(attachmentTickets map[string]*longhorn.AttachmentTicket) bool {
 	for _, ticket := range attachmentTickets {
 		if ticket.Type != longhorn.AttacherTypeSnapshotController &&
-			ticket.Type != longhorn.AttacherTypeBackupController &&
-			ticket.Type != longhorn.AttacherTypeVolumeRebuildingController {
-			return true
-		}
-	}
-	return false
-}
-
-func hasInterruptibleAndFrontendDisabledTicket(attachmentTickets map[string]*longhorn.AttachmentTicket) bool {
-	for _, ticket := range attachmentTickets {
-		if ticket.Type == longhorn.AttacherTypeVolumeRebuildingController {
+			ticket.Type != longhorn.AttacherTypeBackupController {
 			return true
 		}
 	}
@@ -673,7 +655,7 @@ func (vac *VolumeAttachmentController) handleVolumeAttachment(va *longhorn.Volum
 		return
 	}
 
-	log.Infof("Volume %v is selected to attach to node %v, ticket +%v", vol.Name, attachmentTicket.NodeID, attachmentTicket)
+	log.Infof("Volume %v is selected to attach to node %v, ticket %+v", vol.Name, attachmentTicket.NodeID, attachmentTicket)
 
 	vol.Spec.NodeID = attachmentTicket.NodeID
 	setAttachmentParameter(attachmentTicket.Parameters, vol)
--- a/controller/volume_rebuilding_controller.go
+++ b/controller/volume_rebuilding_controller.go
@@ -21,6 +21,8 @@ import (
 
 	"github.com/longhorn/longhorn-manager/constant"
 	"github.com/longhorn/longhorn-manager/datastore"
+	"github.com/longhorn/longhorn-manager/types"
+	"github.com/longhorn/longhorn-manager/util"
 
 	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
 )
@@ -73,6 +75,28 @@ func NewVolumeRebuildingController(
 	}
 	vbc.cacheSyncs = append(vbc.cacheSyncs, ds.VolumeInformer.HasSynced)
 
+	if _, err = ds.SettingInformer.AddEventHandlerWithResyncPeriod(cache.FilteringResourceEventHandler{
+		FilterFunc: isSettingOfflineReplicaRebuilding,
+		Handler: cache.ResourceEventHandlerFuncs{
+			AddFunc:    vbc.enqueueSetting,
+			UpdateFunc: func(old, cur interface{}) { vbc.enqueueSetting(cur) },
+		},
+	}, 0); err != nil {
+		return nil, err
+	}
+	vbc.cacheSyncs = append(vbc.cacheSyncs, ds.SettingInformer.HasSynced)
+
+	if _, err = ds.ReplicaInformer.AddEventHandlerWithResyncPeriod(cache.FilteringResourceEventHandler{
+		FilterFunc: isReplicaOwnerChanged,
+		Handler: cache.ResourceEventHandlerFuncs{
+			AddFunc:    vbc.enqueueVolumeForReplica,
+			UpdateFunc: func(old, cur interface{}) { vbc.enqueueVolumeForReplica(cur) },
+		},
+	}, 0); err != nil {
+		return nil, err
+	}
+	vbc.cacheSyncs = append(vbc.cacheSyncs, ds.ReplicaInformer.HasSynced)
+
 	return vbc, nil
 }
 
@@ -94,6 +118,69 @@ func (vbc *VolumeRebuildingController) enqueueVolumeAfter(obj interface{}, durat
 	}
 
 	vbc.queue.AddAfter(key, duration)
+}
+
+func isSettingOfflineReplicaRebuilding(obj interface{}) bool {
+	setting, ok := obj.(*longhorn.Setting)
+	if !ok {
+		return false
+	}
+	return setting.Name == string(types.SettingNameOfflineReplicaRebuilding)
+}
+
+func (vbc *VolumeRebuildingController) enqueueSetting(obj interface{}) {
+	_, ok := obj.(*longhorn.Setting)
+	if !ok {
+		utilruntime.HandleError(fmt.Errorf("received unexpected obj: %#v", obj))
+		return
+	}
+
+	vs, err := vbc.ds.ListVolumesRO()
+	if err != nil {
+		utilruntime.HandleError(fmt.Errorf("failed to list volumes: %v", err))
+		return
+	}
+
+	for _, v := range vs {
+		if v.Spec.OfflineRebuilding == longhorn.VolumeOfflineRebuildingDisabled {
+			continue
+		}
+		vbc.enqueueVolume(v)
+	}
+}
+
+func isReplicaOwnerChanged(obj interface{}) bool {
+	r, ok := obj.(*longhorn.Replica)
+	if !ok {
+		return false
+	}
+	if !r.DeletionTimestamp.IsZero() {
+		return true
+	}
+	if r.Spec.NodeID == "" || r.Status.OwnerID == "" {
+		return false
+	}
+
+	return r.Spec.NodeID != r.Status.OwnerID
+}
+
+func (vbc *VolumeRebuildingController) enqueueVolumeForReplica(obj interface{}) {
+	r, ok := obj.(*longhorn.Replica)
+	if !ok {
+		utilruntime.HandleError(fmt.Errorf("received unexpected obj: %#v", obj))
+		return
+	}
+
+	v, err := vbc.ds.GetVolumeRO(r.Spec.VolumeName)
+	if err != nil {
+		utilruntime.HandleError(fmt.Errorf("failed to get volume: %v", err))
+		return
+	}
+	if v.Status.State != longhorn.VolumeStateDetached {
+		return
+	}
+
+	vbc.enqueueVolume(v)
 }
 
 func (vbc *VolumeRebuildingController) Run(workers int, stopCh <-chan struct{}) {
@@ -157,7 +244,7 @@ func (vbc *VolumeRebuildingController) syncHandler(key string) (err error) {
 }
 
 func (vbc *VolumeRebuildingController) reconcile(volName string) (err error) {
-	vol, err := vbc.ds.GetVolume(volName)
+	vol, err := vbc.ds.GetVolumeRO(volName)
 	if err != nil {
 		if !apierrors.IsNotFound(err) {
 			return err
@@ -192,10 +279,132 @@ func (vbc *VolumeRebuildingController) reconcile(volName string) (err error) {
 	}()
 
 	rebuildingAttachmentTicketID := longhorn.GetAttachmentTicketID(longhorn.AttacherTypeVolumeRebuildingController, volName)
+	deleteVATicketRequired := true
+	defer func() {
+		if deleteVATicketRequired {
+			delete(va.Spec.AttachmentTickets, rebuildingAttachmentTicketID)
+		}
+	}()
 
-	delete(va.Spec.AttachmentTickets, rebuildingAttachmentTicketID)
+	if !vol.DeletionTimestamp.IsZero() {
+		vbc.logger.Infof("Volume %v is deleting, skip offline rebuilding", volName)
+		return nil
+	}
+	if vol.Status.Robustness == longhorn.VolumeRobustnessFaulted {
+		vbc.logger.Warnf("Volume %v is faulted, skip offline rebuilding", volName)
+		return nil
+	}
+
+	isOfflineRebuildEnabled, err := vbc.isVolumeOfflineRebuildEnabled(vol.Spec.OfflineRebuilding)
+	if err != nil {
+		return err
+	}
+	if isOfflineRebuildEnabled && !util.IsHigherPriorityVATicketExisting(va, longhorn.AttacherTypeVolumeRebuildingController) {
+		if vol.Status.State == longhorn.VolumeStateDetached {
+			_, err = vbc.syncLHVolumeAttachmentForOfflineRebuild(vol, va, rebuildingAttachmentTicketID)
+			if err != nil {
+				return err
+			}
+			deleteVATicketRequired = false
+			return nil
+		}
+		if vol.Status.State == longhorn.VolumeStateAttaching {
+			deleteVATicketRequired = false
+			return nil
+		}
+
+		engine, err := vbc.getVolumeEngine(vol)
+		if err != nil {
+			return err
+		}
+		if engine == nil {
+			vbc.logger.Warnf("Volume %v engine not found, skip offline rebuilding", volName)
+			return nil
+		}
+
+		if engine.Status.ReplicaModeMap == nil {
+			// wait for engine status synced
+			deleteVATicketRequired = false
+			return nil
+		}
+		if vbc.isVolumeReplicasRebuilding(vol, engine) {
+			deleteVATicketRequired = false
+			return nil
+		}
+	}
 
 	return nil
+}
+
+func (vbc *VolumeRebuildingController) isVolumeOfflineRebuildEnabled(offlineRebuilding longhorn.VolumeOfflineRebuilding) (bool, error) {
+	if offlineRebuilding == longhorn.VolumeOfflineRebuildingEnabled {
+		return true, nil
+	}
+
+	globalOfflineRebuildingEnabled, err := vbc.ds.GetSettingAsBool(types.SettingNameOfflineReplicaRebuilding)
+	if err != nil {
+		return false, err
+	}
+	return globalOfflineRebuildingEnabled && offlineRebuilding != longhorn.VolumeOfflineRebuildingDisabled, nil
+}
+
+func (vbc *VolumeRebuildingController) syncLHVolumeAttachmentForOfflineRebuild(vol *longhorn.Volume, va *longhorn.VolumeAttachment, attachmentID string) (*longhorn.VolumeAttachment, error) {
+	replicas, err := vbc.ds.ListVolumeReplicasRO(vol.Name)
+	if err != nil {
+		return va, err
+	}
+
+	if !vbc.isVolumeReplicasHealthy(vol.Spec.NumberOfReplicas, replicas) {
+		if va.Spec.AttachmentTickets == nil {
+			va.Spec.AttachmentTickets = map[string]*longhorn.AttachmentTicket{}
+		}
+		createOrUpdateAttachmentTicket(va, attachmentID, vol.Status.OwnerID, longhorn.AnyValue, longhorn.AttacherTypeVolumeRebuildingController)
+	}
+	return va, nil
+}
+
+func (vbc *VolumeRebuildingController) isVolumeReplicasHealthy(numberOfReplicas int, replicas map[string]*longhorn.Replica) bool {
+	if len(replicas) < numberOfReplicas {
+		return false
+	}
+
+	healthyCount := 0
+	for _, replica := range replicas {
+		if replica.Spec.NodeID != replica.Status.OwnerID {
+			continue
+		}
+		if replica.Spec.FailedAt != "" {
+			continue
+		}
+		// the volume might be newly created and not attached yet.
+		if replica.Spec.HealthyAt != "" || (replica.Spec.LastFailedAt == "" && replica.Spec.LastHealthyAt == "") {
+			healthyCount++
+		}
+	}
+	return healthyCount >= numberOfReplicas
+}
+
+func (vbc *VolumeRebuildingController) getVolumeEngine(vol *longhorn.Volume) (*longhorn.Engine, error) {
+	engines, err := vbc.ds.ListVolumeEngines(vol.Name)
+	if err != nil {
+		return nil, err
+	}
+
+	return vbc.ds.PickVolumeCurrentEngine(vol, engines)
+}
+
+func (vbc *VolumeRebuildingController) isVolumeReplicasRebuilding(vol *longhorn.Volume, engine *longhorn.Engine) bool {
+	healthyCount := 0
+	for _, mode := range engine.Status.ReplicaModeMap {
+		if mode == longhorn.ReplicaModeWO {
+			return true
+		}
+		if mode == longhorn.ReplicaModeRW {
+			healthyCount++
+		}
+	}
+
+	return healthyCount < vol.Spec.NumberOfReplicas
 }
 
 func (vbc *VolumeRebuildingController) isResponsibleFor(vol *longhorn.Volume) bool {
--- a/k8s/crds.yaml
+++ b/k8s/crds.yaml
@@ -4364,6 +4364,17 @@ spec:
                 type: array
               numberOfReplicas:
                 type: integer
+              offlineRebuilding:
+                description: |-
+                  Specifies whether Longhorn should rebuild replicas while the detached volume is degraded.
+                  - ignored: Use the global setting for offline replica rebuilding.
+                  - enabled: Enable offline rebuilding for this volume, regardless of the global setting.
+                  - disabled: Disable offline rebuilding for this volume, regardless of the global setting
+                enum:
+                - ignored
+                - disabled
+                - enabled
+                type: string
               replicaAutoBalance:
                 enum:
                 - ignored
--- a/k8s/pkg/apis/longhorn/v1beta2/volume.go
+++ b/k8s/pkg/apis/longhorn/v1beta2/volume.go
@@ -81,6 +81,14 @@ const (
 	UnmapMarkSnapChainRemovedEnabled  = UnmapMarkSnapChainRemoved("enabled")
 )
 
+type VolumeOfflineRebuilding string
+
+const (
+	VolumeOfflineRebuildingEnabled  = VolumeOfflineRebuilding("enabled")
+	VolumeOfflineRebuildingDisabled = VolumeOfflineRebuilding("disabled")
+	VolumeOfflineRebuildingIgnored  = VolumeOfflineRebuilding("ignored")
+)
+
 type VolumeCloneState string
 
 const (
@@ -291,6 +299,13 @@ type VolumeSpec struct {
 	// The backup target name that the volume will be backed up to or is synced.
 	// +optional
 	BackupTargetName string `json:"backupTargetName"`
+	// +kubebuilder:validation:Enum=ignored;disabled;enabled
+	// Specifies whether Longhorn should rebuild replicas while the detached volume is degraded.
+	// - ignored: Use the global setting for offline replica rebuilding.
+	// - enabled: Enable offline rebuilding for this volume, regardless of the global setting.
+	// - disabled: Disable offline rebuilding for this volume, regardless of the global setting
+	// +optional
+	OfflineRebuilding VolumeOfflineRebuilding `json:"offlineRebuilding"`
 }
 
 // VolumeStatus defines the observed state of the Longhorn volume
--- a/manager/volume.go
+++ b/manager/volume.go
@@ -198,6 +198,7 @@ func (m *VolumeManager) Create(name string, spec *longhorn.VolumeSpec, recurring
 			DataEngine:                  spec.DataEngine,
 			FreezeFilesystemForSnapshot: spec.FreezeFilesystemForSnapshot,
 			BackupTargetName:            backupTargetName,
+			OfflineRebuilding:           spec.OfflineRebuilding,
 		},
 	}
 
@@ -600,6 +601,30 @@ func (m *VolumeManager) CancelExpansion(volumeName string) (v *longhorn.Volume, 
 	}
 
 	logrus.Infof("Canceling volume %v expansion from %v to %v requested", v.Name, previousSize, v.Spec.Size)
+	return v, nil
+}
+
+func (m *VolumeManager) UpdateOfflineRebuilding(volumeName string, offlineRebuildMode longhorn.VolumeOfflineRebuilding) (v *longhorn.Volume, err error) {
+	defer func() {
+		err = errors.Wrapf(err, "unable to rebuild the offline volume %v", volumeName)
+	}()
+
+	v, err = m.ds.GetVolume(volumeName)
+	if err != nil {
+		return nil, err
+	}
+
+	if v.Spec.OfflineRebuilding == offlineRebuildMode {
+		return v, nil
+	}
+
+	v.Spec.OfflineRebuilding = offlineRebuildMode
+	v, err = m.ds.UpdateVolume(v)
+	if err != nil {
+		return nil, err
+	}
+
+	logrus.Infof("Updated volume %v offline rebuilding to %v", volumeName, v.Spec.OfflineRebuilding)
 	return v, nil
 }
 
--- a/types/setting.go
+++ b/types/setting.go
@@ -142,6 +142,7 @@ const (
 	SettingNameDefaultMinNumberOfBackingImageCopies                     = SettingName("default-min-number-of-backing-image-copies")
 	SettingNameBackupExecutionTimeout                                   = SettingName("backup-execution-timeout")
 	SettingNameRWXVolumeFastFailover                                    = SettingName("rwx-volume-fast-failover")
+	SettingNameOfflineReplicaRebuilding                                 = SettingName("offline-replica-rebuilding")
 	// These three backup target parameters are used in the "longhorn-default-resource" ConfigMap
 	// to update the default BackupTarget resource.
 	// Longhorn won't create the Setting resources for these three parameters.
@@ -243,6 +244,7 @@ var (
 		SettingNameDefaultMinNumberOfBackingImageCopies,
 		SettingNameBackupExecutionTimeout,
 		SettingNameRWXVolumeFastFailover,
+		SettingNameOfflineReplicaRebuilding,
 	}
 )
 
@@ -366,6 +368,7 @@ var (
 		SettingNameDefaultMinNumberOfBackingImageCopies:                     SettingDefinitionDefaultMinNumberOfBackingImageCopies,
 		SettingNameBackupExecutionTimeout:                                   SettingDefinitionBackupExecutionTimeout,
 		SettingNameRWXVolumeFastFailover:                                    SettingDefinitionRWXVolumeFastFailover,
+		SettingNameOfflineReplicaRebuilding:                                 SettingDefinitionOfflineReplicaRebuilding,
 	}
 
 	SettingDefinitionAllowRecurringJobWhileVolumeDetached = SettingDefinition{
@@ -1533,6 +1536,20 @@ var (
 		Required:    true,
 		ReadOnly:    false,
 		Default:     "false",
+	}
+
+	SettingDefinitionOfflineReplicaRebuilding = SettingDefinition{
+		DisplayName: "Offline Replica Rebuilding",
+		Description: "Enables automatic rebuilding of degraded replicas while the volume is detached. This setting only takes effect if the individual volume setting is set to `ignored` or `enabled`. \n\n" +
+			"Available options: \n\n" +
+			"- **true**: Enables offline replica rebuilding for all detached volumes, unless overridden by individual volume settings. \n\n" +
+			"- **false**: Disables offline replica rebuilding globally, unless overridden by individual volume settings. \n\n" +
+			"**Note:** Offline rebuilding applies only when a volume is detached. Volumes in a faulted state will not trigger offline rebuilding.",
+		Category: SettingCategoryGeneral,
+		Type:     SettingTypeBool,
+		Required: true,
+		ReadOnly: false,
+		Default:  "false",
 	}
 )
 
--- a/upgrade/v18xto190/upgrade.go
+++ b/upgrade/v18xto190/upgrade.go
@@ -3,12 +3,18 @@ package v18xto190
 import (
 	"github.com/pkg/errors"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	clientset "k8s.io/client-go/kubernetes"
 
 	"github.com/longhorn/longhorn-manager/types"
 
+	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
 	lhclientset "github.com/longhorn/longhorn-manager/k8s/pkg/client/clientset/versioned"
 	upgradeutil "github.com/longhorn/longhorn-manager/upgrade/util"
+)
+
+const (
+	upgradeLogPrefix = "upgrade from v1.8.x to v1.9.0: "
 )
 
 type listAndUpdateFunc func(namespace string, lhClient *lhclientset.Clientset, resourceMaps map[string]interface{}) error
@@ -50,5 +56,32 @@ func UpgradeResources(namespace string, lhClient *lhclientset.Clientset, kubeCli
 			return errors.Wrapf(err, "upgrade from v1.8.x to v1.9.0: failed to list all existing Longhorn %s during the %s upgrade", resourceKind, resourceKind)
 		}
 	}
+
+	if err := upgradeVolumes(namespace, lhClient, resourceMaps); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func upgradeVolumes(namespace string, lhClient *lhclientset.Clientset, resourceMaps map[string]interface{}) (err error) {
+	defer func() {
+		err = errors.Wrapf(err, upgradeLogPrefix+"upgrade volume failed")
+	}()
+
+	volumesMap, err := upgradeutil.ListAndUpdateVolumesInProvidedCache(namespace, lhClient, resourceMaps)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil
+		}
+		return errors.Wrapf(err, "failed to list all existing Longhorn volumes during the volume upgrade")
+	}
+
+	for _, v := range volumesMap {
+		if v.Spec.OfflineRebuilding == "" {
+			v.Spec.OfflineRebuilding = longhorn.VolumeOfflineRebuildingIgnored
+		}
+	}
+
 	return nil
 }
--- a/api/router.go
+++ b/api/router.go
@@ -83,6 +83,7 @@ func NewRouter(s *Server) *mux.Router {
 		"activate":                          s.VolumeActivate,
 		"expand":                            s.VolumeExpand,
 		"cancelExpansion":                   s.VolumeCancelExpansion,
+		"offlineReplicaRebuilding":          s.VolumeOfflineRebuilding,
 
 		"updateReplicaCount":                s.VolumeUpdateReplicaCount,
 		"updateReplicaAutoBalance":          s.VolumeUpdateReplicaAutoBalance,
--- a/api/volume.go
+++ b/api/volume.go
@@ -201,6 +201,7 @@ func (s *Server) VolumeCreate(rw http.ResponseWriter, req *http.Request) error {
 		DataEngine:                  volume.DataEngine,
 		FreezeFilesystemForSnapshot: volume.FreezeFilesystemForSnapshot,
 		BackupTargetName:            volume.BackupTargetName,
+		OfflineRebuilding:           volume.OfflineRebuilding,
 	}, volume.RecurringJobSelector)
 	if err != nil {
 		return errors.Wrap(err, "failed to create volume")
@@ -641,6 +642,30 @@ func (s *Server) VolumeCancelExpansion(rw http.ResponseWriter, req *http.Request
 
 	obj, err := util.RetryOnConflictCause(func() (interface{}, error) {
 		return s.m.CancelExpansion(id)
+	})
+	if err != nil {
+		return err
+	}
+	v, ok := obj.(*longhorn.Volume)
+	if !ok {
+		return fmt.Errorf("failed to convert to volume %v object", id)
+	}
+
+	return s.responseWithVolume(rw, req, "", v)
+}
+
+func (s *Server) VolumeOfflineRebuilding(rw http.ResponseWriter, req *http.Request) error {
+	var input UpdateOfflineRebuildingInput
+
+	apiContext := api.GetApiContext(req)
+	if err := apiContext.Read(&input); err != nil {
+		return err
+	}
+
+	id := mux.Vars(req)["name"]
+
+	obj, err := util.RetryOnConflictCause(func() (interface{}, error) {
+		return s.m.UpdateOfflineRebuilding(id, longhorn.VolumeOfflineRebuilding(input.OfflineRebuilding))
 	})
 	if err != nil {
 		return err
--- a/client/generated_volume.go
+++ b/client/generated_volume.go
@@ -63,6 +63,8 @@ type Volume struct {
 
 	NumberOfReplicas int64 `json:"numberOfReplicas,omitempty" yaml:"number_of_replicas,omitempty"`
 
+	OfflineRebuilding string `json:"offlineRebuilding,omitempty" yaml:"offline_rebuilding,omitempty"`
+
 	PurgeStatus []PurgeStatus `json:"purgeStatus,omitempty" yaml:"purge_status,omitempty"`
 
 	Ready bool `json:"ready,omitempty" yaml:"ready,omitempty"`
@@ -138,6 +140,8 @@ type VolumeOperations interface {
 	ActionAttach(*Volume, *AttachInput) (*Volume, error)
 
 	ActionCancelExpansion(*Volume) (*Volume, error)
+
+	ActionOfflineReplicaRebuilding(*Volume) (*Volume, error)
 
 	ActionDetach(*Volume, *DetachInput) (*Volume, error)
 
@@ -257,6 +261,15 @@ func (c *VolumeClient) ActionCancelExpansion(resource *Volume) (*Volume, error) 
 	resp := &Volume{}
 
 	err := c.rancherClient.doAction(VOLUME_TYPE, "cancelExpansion", &resource.Resource, nil, resp)
+
+	return resp, err
+}
+
+func (c *VolumeClient) ActionOfflineReplicaRebuilding(resource *Volume) (*Volume, error) {
+
+	resp := &Volume{}
+
+	err := c.rancherClient.doAction(VOLUME_TYPE, "offlineReplicaRebuilding", &resource.Resource, nil, resp)
 
 	return resp, err
 }
--- a/k8s/pkg/client/applyconfiguration/longhorn/v1beta2/volumespec.go
+++ b/k8s/pkg/client/applyconfiguration/longhorn/v1beta2/volumespec.go
@@ -58,6 +58,7 @@ type VolumeSpecApplyConfiguration struct {
 	SnapshotMaxSize             *int64                                         `json:"snapshotMaxSize,omitempty"`
 	FreezeFilesystemForSnapshot *longhornv1beta2.FreezeFilesystemForSnapshot   `json:"freezeFilesystemForSnapshot,omitempty"`
 	BackupTargetName            *string                                        `json:"backupTargetName,omitempty"`
+	OfflineRebuilding           *longhornv1beta2.VolumeOfflineRebuilding       `json:"offlineRebuilding,omitempty"`
 }
 
 // VolumeSpecApplyConfiguration constructs a declarative configuration of the VolumeSpec type for use with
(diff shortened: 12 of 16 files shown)


derekbit:
Just discussed with @mantissahz the design, @mantissahz will check the feasibility and update the LEP soon.

derekbit:
cc @c3y1huang @PhanLe1010 @shuo-wu for review

controller/volume_rebuilding_controller.go
```diff
@@ -96,6 +108,106 @@ func (vbc *VolumeRebuildingController) enqueueVolumeAfter(obj interface{}, durat
...
+	for _, v := range vs {
+		if v.Spec.OfflineRebuild != longhorn.VolumeOfflineRebuildIgnored {
+			continue
+		}
+		if offlineRebuildFlag && v.Status.State != longhorn.VolumeStateDetached {
+			continue
+		}
+		if !offlineRebuildFlag && v.Status.State != longhorn.VolumeStateAttached {
+			continue
+		}
```
c3y1huang, line 168: Could you clarify the conditions for volume enqueue? Why not just skip:
- `v.Spec.OfflineRebuild != longhorn.VolumeOfflineRebuildIgnored`
- `!offlineRebuildFlag`
- `v.Status.State != longhorn.VolumeStateDetached`

controller/volume_rebuilding_controller.go
```diff
@@ -73,6 +74,17 @@ func NewVolumeRebuildingController(
 	}
 	vbc.cacheSyncs = append(vbc.cacheSyncs, ds.VolumeInformer.HasSynced)
 
+	if _, err = ds.SettingInformer.AddEventHandlerWithResyncPeriod(cache.FilteringResourceEventHandler{
+		FilterFunc: isSettingRelatedToOffLineRebuilding,
+		Handler: cache.ResourceEventHandlerFuncs{
+			AddFunc:    vbc.enqueueSetting,
+			UpdateFunc: vbc.enqueueSettingUpdated,
```
c3y1huang, line 81: Why not use `func(old, cur interface{}) { sc.enqueueSetting(cur) }`?

controller/volume_rebuilding_controller.go
```diff
@@ -190,14 +302,128 @@ func (vbc *VolumeRebuildingController) reconcile(volName string) (err error) {
...
+	for _, replica := range replicas {
+		if replica.Spec.FailedAt == "" && replica.Spec.HealthyAt != "" {
+			healthyCount++
+		}
+	}
+
+	return healthyCount < vol.Spec.NumberOfReplicas
+}
+
+func (vbc *VolumeRebuildingController) checkifVolumeInRebuilding(vol *longhorn.Volume) (bool, error) {
```
c3y1huang, line 388:
```suggestion
func (vbc *VolumeRebuildingController) isVolumeReplicasRebuilding(vol *longhorn.Volume) (bool, error) {
```

controller/volume_rebuilding_controller.go
```diff
@@ -190,14 +302,128 @@ func (vbc *VolumeRebuildingController) reconcile(volName string) (err error) {
...
+	}
+
+	if healthyCount < vol.Spec.NumberOfReplicas {
+		return true, nil
+	}
+
+	return false, nil
+}
+
+func checkIfVATicketExists(va *longhorn.VolumeAttachment, ticketType longhorn.AttacherType) bool {
```
c3y1huang, line 415:
```suggestion
func isVATicketExists(va *longhorn.VolumeAttachment, ticketType longhorn.AttacherType) bool {
```

types/setting.go
```diff
@@ -1514,6 +1517,16 @@ var (
 		ReadOnly:    false,
 		Default:     "false",
 	}
+
+	SettingDefinitionOfflineReplicaRebuilding = SettingDefinition{
+		DisplayName: "Offline Replica Rebuilding",
+		Description: "Rebuild replicas of detached volumes which need to be rebuilt",
```
c3y1huang, line 1523: This is user-facing content; consider requesting a review from the doc team.

c3y1huang (changes requested):

controller/volume_rebuilding_controller.go
```diff
@@ -190,14 +302,128 @@ func (vbc *VolumeRebuildingController) reconcile(volName string) (err error) {
...
+		}
+	}()
 
 	rebuildingAttachmentTicketID := longhorn.GetAttachmentTicketID(longhorn.AttacherTypeVolumeRebuildingController, volName)
+	csiTicketExists := checkIfVATicketExists(va, longhorn.AttacherTypeCSIAttacher)
+	offlineRebuildRebuildingValue, err := vbc.ds.GetSettingAsBool(types.SettingNameOfflineReplicaRebuilding)
+	if err != nil {
+		return err
+	}
+	offlineRebuildFlag := (vol.Spec.OfflineRebuild == longhorn.VolumeOfflineRebuildEnabled) || (vol.Spec.OfflineRebuild == longhorn.VolumeOfflineRebuildIgnored && offlineRebuildRebuildingValue)
```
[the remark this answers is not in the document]
mantissahz: Extracted the logic into a helper function `isVolumeOfflineRebuildEnabled`. Thanks.

controller/volume_rebuilding_controller.go
```diff
@@ -96,6 +108,106 @@ func (vbc *VolumeRebuildingController) enqueueVolumeAfter(obj interface{}, durat
 	vbc.queue.AddAfter(key, duration)
 }
 
+func isSettingRelatedToOffLineRebuilding(obj interface{}) bool {
```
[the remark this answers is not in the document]
mantissahz: Renamed. Thanks

controller/volume_rebuilding_controller.go
```diff
@@ -96,6 +108,106 @@ func (vbc *VolumeRebuildingController) enqueueVolumeAfter(obj interface{}, durat
...
+		if !ok {
+			return false
+		}
+	}
+
+	return setting.Name == string(types.SettingNameOfflineReplicaRebuilding)
+}
+
+func (vbc *VolumeRebuildingController) enqueueSetting(obj interface{}) {
+	vbc.logger.Infof("[James_DBG] enqueueSetting Setting updated: %v in VRBController", obj)
```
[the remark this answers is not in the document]
mantissahz: Removed. Thanks.

controller/volume_rebuilding_controller.go
```diff
@@ -96,6 +108,106 @@ func (vbc *VolumeRebuildingController) enqueueVolumeAfter(obj interface{}, durat
...
+
+		// use the last known state, to requeue the claimed volumes
+		_, ok = deletedState.Obj.(*longhorn.Setting)
+		if !ok {
+			utilruntime.HandleError(fmt.Errorf("DeletedFinalStateUnknown contained non Setting object: %#v", deletedState.Obj))
+			return
+		}
+	}
+
+	offlineRebuildFlag, err := vbc.ds.GetSettingAsBool(types.SettingNameOfflineReplicaRebuilding)
```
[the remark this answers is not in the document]
mantissahz: Removed. Thanks

controller/volume_rebuilding_controller.go
mantissahz: Only skip `v.Spec.OfflineRebuild != longhorn.VolumeOfflineRebuildIgnored`.
Others will be handled during the reconciliation.
Thanks

controller/volume_rebuilding_controller.go
mantissahz: Modified to use `func(old, cur interface{}) { sc.enqueueSetting(cur) }`
Thanks.

controller/volume_rebuilding_controller.go
mantissahz: Renamed. Thanks.

controller/volume_rebuilding_controller.go
mantissahz: Renamed. Thanks

types/setting.go
mantissahz: Yes, please refine the description @longhorn/doc. Thanks

manager/volume.go
```diff
@@ -603,6 +603,30 @@ func (m *VolumeManager) CancelExpansion(volumeName string) (v *longhorn.Volume,
 	return v, nil
 }
 
+func (m *VolumeManager) OfflineRebuild(volumeName string, rebuildState longhorn.VolumeOfflineRebuild) (v *longhorn.Volume, err error) {
```
derekbit, line 606: The naming of parameter `rebuildState` is a bit weird, but I don't have a better suggestion.

```suggestion
func (m *VolumeManager) UpdateOfflineRebuild(volumeName string, rebuildState longhorn.VolumeOfflineRebuild) (v *longhorn.Volume, err error) {
```

controller/volume_rebuilding_controller.go
```diff
@@ -190,14 +252,124 @@ func (vbc *VolumeRebuildingController) reconcile(volName string) (err error) {
...
+func (vbc *VolumeRebuildingController) syncLHVolumeAttachementForOfflineRebuild(vol *longhorn.Volume, va *longhorn.VolumeAttachment, attachmentID string) (*longhorn.VolumeAttachment, error) {
+	replicas, err := vbc.ds.ListVolumeReplicasRO(vol.Name)
+	if err != nil {
+		return va, err
+	}
+	if !util.IsVolumeReplicasHealthy(vol.Spec.NumberOfReplicas, replicas) {
+		createOrUpdateAttachmentTicket(va, attachmentID, vol.Status.OwnerID, longhorn.TrueValue, longhorn.AttacherTypeVolumeRebuildingController)
+		vol.Spec.DisableFrontend = true
+	}
+	return va, nil
```
derekbit, line 331: Why not just check volume.status.robutness?

mantissahz: because there is no value for `volume.status.robutness` when the volume is detached.

manager/volume.go
mantissahz: Renamed the API name and modified the name `rebuildState` to `offlineRebuildMode`, WDYT?
Thanks.

controller/volume_attachment_controller.go
```diff
@@ -618,17 +610,7 @@ func (vac *VolumeAttachmentController) shouldDoDetach(va *longhorn.VolumeAttachm
...
-			ticket.Type != longhorn.AttacherTypeVolumeRebuildingController {
-			return true
-		}
-	}
-	return false
-}
-
-func hasInterruptibleAndFrontendDisabledTicket(attachmentTickets map[string]*longhorn.AttachmentTicket) bool {
-	for _, ticket := range attachmentTickets {
-		if ticket.Type == longhorn.AttacherTypeVolumeRebuildingController {
```
PhanLe1010, line 631: Is it the cleanup from the previous `rebuilding-controller` feature?

controller/volume_rebuilding_controller.go
```diff
@@ -190,14 +252,124 @@ func (vbc *VolumeRebuildingController) reconcile(volName string) (err error) {
...
+}
+
+func (vbc *VolumeRebuildingController) syncLHVolumeAttachementForOfflineRebuild(vol *longhorn.Volume, va *longhorn.VolumeAttachment, attachmentID string) (*longhorn.VolumeAttachment, error) {
+	replicas, err := vbc.ds.ListVolumeReplicasRO(vol.Name)
+	if err != nil {
+		return va, err
+	}
+	if !util.IsVolumeReplicasHealthy(vol.Spec.NumberOfReplicas, replicas) {
+		createOrUpdateAttachmentTicket(va, attachmentID, vol.Status.OwnerID, longhorn.TrueValue, longhorn.AttacherTypeVolumeRebuildingController)
+		vol.Spec.DisableFrontend = true
```
PhanLe1010, line 329: Why do we need to disable frontend?

controller/volume_rebuilding_controller.go
```diff
@@ -190,14 +252,124 @@ func (vbc *VolumeRebuildingController) reconcile(volName string) (err error) {
...
 
 	rebuildingAttachmentTicketID := longhorn.GetAttachmentTicketID(longhorn.AttacherTypeVolumeRebuildingController, volName)
+	csiTicketExists := isVATicketExists(va, longhorn.AttacherTypeCSIAttacher)
+	isOfflineRebuildEnabled, err := vbc.isVolumeOfflineRebuildEnabled(vol.Spec.OfflineRebuild)
+	if err != nil {
+		return err
+	}
 
-	delete(va.Spec.AttachmentTickets, rebuildingAttachmentTicketID)
+	if isOfflineRebuildEnabled && vol.Status.Robustness != longhorn.VolumeRobustnessFaulted && !csiTicketExists {
```
PhanLe1010, line 276: What about if other important tickets such as longhorn-ui, restore controller, clone controller exist? CSI ticket is not only the highest priority ticket. Maybe we can check if there are no other ticket here?

controller/volume_attachment_controller.go
mantissahz: >Is it the cleanup from the previous rebuilding-controller feature?

Did you mean the volume rebuild VA ticket will be deleted by `volume-rebuild-controler`?

controller/volume_rebuilding_controller.go
mantissahz: Let it get into the `maintenence` mode.

controller/volume_attachment_controller.go
mantissahz: The volume rebuild VA ticket should be able to be preempted by the higher-priority VA ticket.

controller/volume_rebuilding_controller.go
mantissahz: Yes, it's a good idea.
Modified to be that higher priority VA ticket can preempt the volume rebuild VA ticket.
Thanks

controller/volume_rebuilding_controller.go
PhanLe1010: I see. 
Now thinking about it again, I think we can do something similar to https://github.com/longhorn/longhorn-manager/blob/763ac9117eb243349d7d91827a1023eaa7980f35/controller/snapshot_controller.go#L563-L564 by setting `longhorn.AttachmentParameterDisableFrontend` to  `longhorn.AnyValue` instead of hardcode the `vol.Spec.DisableFrontend = true`

I think for rebuilding, front-end does not matter. 

WDYT?

controller/volume_attachment_controller.go
PhanLe1010: Sorry for the confusion. I meant we removed this logic because it is no longer needed and it is the leftover from the original v2 rebuilding feature?

mantissahz: >because it is no longer needed and it is the leftover from the original v2 rebuilding feature?

I think Yes.

controller/volume_rebuilding_controller.go
mantissahz: >by setting longhorn.AttachmentParameterDisableFrontend to longhorn.AnyValue instead of hardcode the vol.Spec.DisableFrontend = true

Do you mean to set `longhorn.AttachmentParameterDisableFrontend` to `longhorn.TrueValue` because we want volumes in the `maintenance` mode?

PhanLe1010: > Do you mean to set longhorn.AttachmentParameterDisableFrontend to longhorn.TrueValue because we want volumes in the maintenance mode?

I mean to set longhorn.AttachmentParameterDisableFrontend to longhorn.TrueValue because we don't care whether the volume has frontenddisable or not. In either case, the rebuilding operation can proceed

mantissahz: Modified. Thanks.

types/setting.go
```diff
@@ -1534,6 +1537,16 @@ var (
...
+		Category:    SettingCategoryGeneral,
```
derekbit, line 1544: Hello @sushant-suse,
Please help review the description. Thanks.

types/setting.go
derekbit: IMO, mention "offline rebuilding" in the description. Does the setting apply to both v1 and v2 data engine?

manager/volume.go
```diff
@@ -603,6 +604,30 @@ func (m *VolumeManager) CancelExpansion(volumeName string) (v *longhorn.Volume,
...
+		return v, nil
+	}
+
+	v.Spec.OfflineRebuild = offlineRebuildMode
+	v, err = m.ds.UpdateVolume(v)
+	if err != nil {
+		return nil, err
+	}
+
+	logrus.Infof("Updated volume %v offline rebuild spec to %v", volumeName, v.Spec.OfflineRebuild)
```
derekbit, line 627:
```suggestion
	logrus.Infof("Updated volume %v offline rebuilding to %v", volumeName, v.Spec.OfflineRebuild)
```

k8s/pkg/apis/longhorn/v1beta2/volume.go
```diff
@@ -80,6 +80,14 @@ const (
 	UnmapMarkSnapChainRemovedEnabled  = UnmapMarkSnapChainRemoved("enabled")
 )
 
+type VolumeOfflineRebuild string
```
derekbit, line 83: The global setting is `OfflineReplicaRebuilding`. Make the naming consistent.

```suggestion
type VolumeOfflineRebuilding string
```

k8s/pkg/apis/longhorn/v1beta2/volume.go
```diff
@@ -290,6 +298,9 @@ type VolumeSpec struct {
 	// The backup target name that the volume will be backed up to or is synced.
 	// +optional
 	BackupTargetName string `json:"backupTargetName"`
+	// +kubebuilder:validation:Enum=ignored;disabled;enabled
+	// +optional
+	OfflineRebuild VolumeOfflineRebuild `json:"offlineRebuild"`
```
derekbit, line 303:
```suggestion
	OfflineRebuilding VolumeOfflineRebuilding `json:"offlineRebuilding"`
```

manager/volume.go
```diff
@@ -198,6 +198,7 @@ func (m *VolumeManager) Create(name string, spec *longhorn.VolumeSpec, recurring
 			DataEngine:                  spec.DataEngine,
 			FreezeFilesystemForSnapshot: spec.FreezeFilesystemForSnapshot,
 			BackupTargetName:            backupTargetName,
+			OfflineRebuild:              spec.OfflineRebuild,
```
derekbit, line 201:
```suggestion
			OfflineRebuilding:              spec.OfflineRebuilding,
```

manager/volume.go
```diff
@@ -603,6 +604,30 @@ func (m *VolumeManager) CancelExpansion(volumeName string) (v *longhorn.Volume,
 	return v, nil
 }
 
+func (m *VolumeManager) UpdateOfflineRebuild(volumeName string, offlineRebuildMode longhorn.VolumeOfflineRebuild) (v *longhorn.Volume, err error) {
```
derekbit, line 607:
```suggestion
func (m *VolumeManager) UpdateOfflineRebuilding(volumeName string, offlineRebuildMode longhorn.VolumeOfflineRebuild) (v *longhorn.Volume, err error) {
```

constant/events.go
```diff
@@ -33,6 +33,8 @@ const (
 	EventReasonSucceededExpansion = "SucceededExpansion"
 	EventReasonCanceledExpansion  = "CanceledExpansion"
 
+	EventReasonCanceledOfflineRebuild = "CanceledOfflineRebuild"
```
derekbit, line 36:
```suggestion
	EventReasonCanceledOfflineRebuilding = "CanceledOfflineRebuilding"
```

derekbit: EventReasonCanceledOfflineRebuilding = "CanceledOfflineRebuilding"

derekbit (commented):
Suggest making the naming consistent.

controller/volume_rebuilding_controller.go
```diff
@@ -73,6 +75,17 @@ func NewVolumeRebuildingController(
...
+	if _, err = ds.SettingInformer.AddEventHandlerWithResyncPeriod(cache.FilteringResourceEventHandler{
+		FilterFunc: isSettingOfflineReplicaRebuilding,
+		Handler: cache.ResourceEventHandlerFuncs{
+			AddFunc:    vbc.enqueueSetting,
+			UpdateFunc: func(old, cur interface{}) { vbc.enqueueSetting(cur) },
+		},
+	}, 0); err != nil {
+		return nil, err
+	}
+	vbc.cacheSyncs = append(vbc.cacheSyncs, ds.SettingInformer.HasSynced)
```
derekbit, line 87: Move to setting controller.

derekbit (commented):
Need to test more scenarios.

types/setting.go
mantissahz: >Does the setting apply to both v1 and v2 data engine?

Yes, it's applied to v1 and v2 data engines.

k8s/pkg/apis/longhorn/v1beta2/volume.go
mantissahz: Modified. Thanks

k8s/pkg/apis/longhorn/v1beta2/volume.go
mantissahz: Modified. Thanks

manager/volume.go
mantissahz: Modified. Thanks

constant/events.go
mantissahz: Modified. Thanks

api/model.go
```diff
@@ -75,9 +75,10 @@ type Volume struct {
 	CloneStatus      longhorn.VolumeCloneStatus    `json:"cloneStatus"`
 	Ready            bool                          `json:"ready"`
 
-	AccessMode    longhorn.AccessMode        `json:"accessMode"`
-	ShareEndpoint string                     `json:"shareEndpoint"`
-	ShareState    longhorn.ShareManagerState `json:"shareState"`
+	AccessMode     longhorn.AccessMode           `json:"accessMode"`
+	ShareEndpoint  string                        `json:"shareEndpoint"`
+	ShareState     longhorn.ShareManagerState    `json:"shareState"`
+	OfflineRebuild longhorn.VolumeOfflineRebuild `json:"offlineRebuild"`
```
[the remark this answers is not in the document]
mantissahz: Modified. Thanks

controller/volume_rebuilding_controller.go
mantissahz: won't move it. Thanks

manager/volume.go
mantissahz: Modified. Thanks

controller/volume_rebuilding_controller.go
```diff
@@ -191,13 +318,122 @@ func (vbc *VolumeRebuildingController) reconcile(volName string) (err error) {
...
 	return nil
 }
 
+func (vbc *VolumeRebuildingController) isVolumeOfflineRebuildEnabled(offlineRebuilding longhorn.VolumeOfflineRebuilding) (bool, error) {
+	offlineReplicaRebuildingValue, err := vbc.ds.GetSettingAsBool(types.SettingNameOfflineReplicaRebuilding)
+	if err != nil {
+		return false, err
+	}
+	return (offlineRebuilding == longhorn.VolumeOfflineRebuildingEnabled) || (offlineRebuilding == longhorn.VolumeOfflineRebuildingIgnored && offlineReplicaRebuildingValue), nil
+}
```
derekbit, line 371: Check if this makes sense.

```suggestion
func (vbc *VolumeRebuildingController) isVolumeOfflineRebuildEnabled(offlineRebuilding longhorn.VolumeOfflineRebuilding) (bool, error) {
    // Per-volume setting has the highest priority
    if offlineRebuilding == longhorn.VolumeOfflineRebuildingEnabled {
        return true, nil
    }

	offlineReplicaRebuildingValue, err := vbc.ds.GetSettingAsBool(types.SettingNameOfflineReplicaRebuilding)
	if err != nil {
		return false, err
	}

    return offlineReplicaRebuildingValue == longhorn.VolumeOfflineRebuildingEnabled && offlineRebuilding != longhorn.VolumeOfflineRebuildingDisabled
	
}
```

controller/volume_rebuilding_controller.go
```diff
@@ -191,13 +318,122 @@ func (vbc *VolumeRebuildingController) reconcile(volName string) (err error) {
...
+			healthyCount++
+		}
+	}
+
+	if healthyCount < vol.Spec.NumberOfReplicas {
+		return true, nil
+	}
+
+	return false, nil
+}
```
derekbit, line 435: If a replica is being rebuilt, the ReplicaMode is WO. Could we check by this mode?

controller/volume_rebuilding_controller.go
```diff
@@ -191,13 +318,122 @@ func (vbc *VolumeRebuildingController) reconcile(volName string) (err error) {
...
+			if apierrors.IsNotFound(err) {
+				continue
+			}
+			return false, err
+		}
+		isNodeReady := false
+		if node.Status.Conditions != nil {
+			nodeReadyCondition := types.GetCondition(node.Status.Conditions, longhorn.NodeConditionTypeReady)
+			isNodeReady = nodeReadyCondition.Status == longhorn.ConditionStatusTrue
+		}
```
derekbit, line 402: I think we don't need to check the node condition and can focus solely on the volume replicas.
If there's an issue with the node, it will be reflected through the replicas status.

types/setting.go
sushant-suse: Hi @derekbit, it looks good. Clear and concise.

types/setting.go
c3y1huang: Consider providing more context about the purpose and usage of the setting. A more detailed explanation will help users understand what the setting does and when to use it without needing to refer to the Longhorn doc. You can also refer to the other setting descriptions.

---
Enable this setting to allow automatic rebuilding of degraded replicas while the volume is detached.

This global setting only takes effect if the volume's individual setting is set to  `ignored`.

Global options:
- `true`: Enable offline replica rebuilding for all detached volumes (unless overridden by individual volume settings).
- `false`: Disable offline replica rebuilding globally (unless overridden by individual volume settings).

Longhorn also supports per-volume setting. You can configure each volume individually on the **Volume** page to override the global behavior.

Available per-volume setting options:
- `ignored`: Follows the volume of the global setting.
- `enabled`: Enable offline rebuilding for the volume, regardless of the global setting.
- `disable`: Disable offline rebuilding of the volume, regardless of the global setting. 

Note: Offline rebuilding is only applicable when a volume is detached. Volume that are faulted will not trigger offline rebuilding.

k8s/pkg/apis/longhorn/v1beta2/volume.go
```diff
@@ -290,6 +298,9 @@ type VolumeSpec struct {
 	// The backup target name that the volume will be backed up to or is synced.
 	// +optional
 	BackupTargetName string `json:"backupTargetName"`
+	// +kubebuilder:validation:Enum=ignored;disabled;enabled
+	// +optional
+	OfflineRebuilding VolumeOfflineRebuilding `json:"offlineRebuilding"`
```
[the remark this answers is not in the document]
mantissahz: Added. Thanks.

controller/volume_rebuilding_controller.go
```diff
@@ -96,6 +120,109 @@ func (vbc *VolumeRebuildingController) enqueueVolumeAfter(obj interface{}, durat
...
+		if !ok {
+			return false
+		}
+
+		// use the last known state, to enqueue, dependent objects
+		node, ok = deletedState.Obj.(*longhorn.Node)
+		if !ok {
+			return false
+		}
+	}
```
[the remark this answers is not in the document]
mantissahz: Modified. Thanks

controller/volume_rebuilding_controller.go
```diff
@@ -96,6 +120,109 @@ func (vbc *VolumeRebuildingController) enqueueVolumeAfter(obj interface{}, durat
...
+	isNodeReady := false
+	if node.Status.Conditions != nil {
+		nodeReadyCondition := types.GetCondition(node.Status.Conditions, longhorn.NodeConditionTypeReady)
+		isNodeReady = nodeReadyCondition.Status == longhorn.ConditionStatusTrue
+	}
+
+	return !node.DeletionTimestamp.IsZero() || !isNodeReady
```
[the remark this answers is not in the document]
mantissahz: Modified. Thanks

controller/volume_rebuilding_controller.go
```diff
@@ -96,6 +120,109 @@ func (vbc *VolumeRebuildingController) enqueueVolumeAfter(obj interface{}, durat
...
+			return
+		}
+
+		// use the last known state, to requeue the claimed volumes
+		_, ok = deletedState.Obj.(*longhorn.Node)
+		if !ok {
+			utilruntime.HandleError(fmt.Errorf("DeletedFinalStateUnknown contained non SetNodeting object: %#v", deletedState.Obj))
+			return
+		}
+	}
```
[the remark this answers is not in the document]
mantissahz: Modified. Thanks

controller/volume_rebuilding_controller.go
```diff
@@ -191,13 +318,122 @@ func (vbc *VolumeRebuildingController) reconcile(volName string) (err error) {
 		}
 	}()
 
+	isOfflineRebuildEnabled, err := vbc.isVolumeOfflineRebuildEnabled(vol.Spec.OfflineRebuilding)
+	if err != nil {
+		return err
+	}
 	rebuildingAttachmentTicketID := longhorn.GetAttachmentTicketID(longhorn.AttacherTypeVolumeRebuildingController, volName)
+	higherPrioritypVATicketExists := util.IsHigherPriorityVATicketExisting(va, longhorn.AttacherTypeVolumeRebuildingController)
```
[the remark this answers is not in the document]
mantissahz: Modified. Thanks

controller/volume_rebuilding_controller.go
```diff
@@ -191,13 +318,122 @@ func (vbc *VolumeRebuildingController) reconcile(volName string) (err error) {
...
+
+	if isOfflineRebuildEnabled && vol.Status.Robustness != longhorn.VolumeRobustnessFaulted && !higherPrioritypVATicketExists {
+		if vol.Status.State == longhorn.VolumeStateDetached {
+			va, err = vbc.syncLHVolumeAttachementForOfflineRebuild(vol, va, rebuildingAttachmentTicketID)
```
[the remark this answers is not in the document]
mantissahz: Modified. Thanks

controller/volume_rebuilding_controller.go
```diff
@@ -191,13 +318,122 @@ func (vbc *VolumeRebuildingController) reconcile(volName string) (err error) {
 		}
 	}()
```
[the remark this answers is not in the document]
mantissahz: Modified. Thanks

controller/volume_rebuilding_controller.go
```diff
@@ -73,6 +75,28 @@ func NewVolumeRebuildingController(
...
+		Handler: cache.ResourceEventHandlerFuncs{
+			AddFunc:    vbc.enqueueSetting,
+			UpdateFunc: func(old, cur interface{}) { vbc.enqueueSetting(cur) },
+		},
+	}, 0); err != nil {
+		return nil, err
+	}
+	vbc.cacheSyncs = append(vbc.cacheSyncs, ds.SettingInformer.HasSynced)
+
+	if _, err = ds.NodeInformer.AddEventHandlerWithResyncPeriod(cache.FilteringResourceEventHandler{
```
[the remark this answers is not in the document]
mantissahz: Removed event-related codes. Thanks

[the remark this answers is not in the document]
mantissahz: >When a node is offline, the replicas on the node will be deleted. Then, the volumes will be degraded. The rebuilding controller can be aware of the degraded volumes.

After some tests, the replica is not removed if a volume is detached, and a worker node is shut down. (waited around 6 hours)
Therefore, I would keep the node informer to check if the node is down.

controller/volume_rebuilding_controller.go
```diff
@@ -191,13 +318,122 @@ func (vbc *VolumeRebuildingController) reconcile(volName string) (err error) {
...
+		warnMsg := fmt.Sprintf("Volume %v is faulted, skip rebuilding", volName)
+		vbc.logger.Info(warnMsg)
+		vbc.eventRecorder.Event(vol, corev1.EventTypeWarning, constant.EventReasonCanceledOfflineRebuilding, warnMsg)
+	}
+
+	if !isOfflineRebuildEnabled && util.IsHigherPriorityVATicketExisting(va, longhorn.AttacherTypeVolumeRebuildingController) {
+		warnMsg := fmt.Sprintf("Canceling volume %v offline rebuilding", volName)
+		vbc.logger.Info(warnMsg)
+		vbc.eventRecorder.Event(vol, corev1.EventTypeWarning, constant.EventReasonCanceledOfflineRebuilding, warnMsg)
+	}
```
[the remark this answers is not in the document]
mantissahz: Removed. Thanks.

controller/volume_rebuilding_controller.go
mantissahz: Modified. Thanks

controller/volume_rebuilding_controller.go
mantissahz: Yes, modified to check if a replica is in `WO` mode, we can return `true` directly. Thanks.

controller/volume_rebuilding_controller.go
mantissahz: >If there's an issue with the node, it will be reflected through the replicas status.

After some tests, the replica status just switched the `OwnerID` to another worker node when a worker node shut down was detected.

types/setting.go
```diff
@@ -1534,6 +1537,20 @@ var (
...
+			"Available options are: \n\n" +
+			"- **true**: Enable offline replica rebuilding for all detached volumes (unless overridden by individual volume settings). \n\n" +
+			"- **false**: Disable offline replica rebuilding globally (unless overridden by individual volume settings). \n\n" +
+			"Note: Offline rebuilding is only applicable when a volume is detached. Volume that are faulted will not trigger offline rebuilding",
+		Category: SettingCategoryGeneral,
+		Type:     SettingTypeBool,
+		Required: true,
+		ReadOnly: false,
+		Default:  "false",
+	}
```
derekbit, line 1553: @sushant-suse Please help review the description. Thank you.

types/setting.go
```diff
@@ -1534,6 +1537,20 @@ var (
 		ReadOnly:    false,
 		Default:     "false",
 	}
+
+	SettingDefinitionOfflineReplicaRebuilding = SettingDefinition{
+		DisplayName: "Offline Replica Rebuilding",
+		Description: "This setting enables automatic rebuilding of degraded replicas while the volume is detached, and it only takes effect if the volume's individual setting is set to `ignored`. \n\n" +
```
sushant-suse, line 1543:
```suggestion
		Description: "Enables automatic rebuilding of degraded replicas while the volume is detached. This setting only takes effect if the individual volume setting is set to `ignored`.\n\n" +
```

types/setting.go
```diff
@@ -1534,6 +1537,20 @@ var (
...
+			"Available options are: \n\n" +
```
sushant-suse, line 1544:
```suggestion
			"Available options:\n\n" +
```

types/setting.go
```diff
@@ -1534,6 +1537,20 @@ var (
...
+			"- **true**: Enable offline replica rebuilding for all detached volumes (unless overridden by individual volume settings). \n\n" +
```
sushant-suse, line 1545:
```suggestion
			"- **true**: Enables offline replica rebuilding for all detached volumes, unless overridden by individual volume settings.\n" +
```

types/setting.go
```diff
@@ -1534,6 +1537,20 @@ var (
...
+			"- **false**: Disable offline replica rebuilding globally (unless overridden by individual volume settings). \n\n" +
```
sushant-suse, line 1546:
```suggestion
			"- **false**: Disables offline replica rebuilding globally, unless overridden by individual volume settings.\n\n" +
```

types/setting.go
```diff
@@ -1534,6 +1537,20 @@ var (
...
+			"Note: Offline rebuilding is only applicable when a volume is detached. Volume that are faulted will not trigger offline rebuilding",
```
sushant-suse, line 1547:
```suggestion
			"Note: Offline rebuilding applies only when a volume is detached. Volumes in a faulted state will not trigger offline rebuilding.",
```

sushant-suse (changes requested):
Thanks.

types/setting.go
mantissahz: Updated. Thanks

types/setting.go
mantissahz: Updated. Thanks

types/setting.go
mantissahz: Updated. Thanks

types/setting.go
mantissahz: Updated. Thanks

types/setting.go
mantissahz: Updated. Thanks

k8s/crds.yaml
```diff
@@ -4347,6 +4347,17 @@ spec:
                 type: array
               numberOfReplicas:
                 type: integer
+              offlineRebuilding:
+                description: |-
```
sushant-suse, line 4351: Can I suggest some enhancement here? @derekbit  @mantissahz

mantissahz: Sure, please.

k8s/crds.yaml
```diff
@@ -4347,6 +4347,17 @@ spec:
...
+                  Setting that allows Longhorn to rebuild replicas while the volume is offline.
```
sushant-suse, line 4352:
```suggestion
                  Controls whether Longhorn should rebuild replicas while the volume is offline.
```

k8s/crds.yaml
```diff
@@ -4347,6 +4347,17 @@ spec:
...
+                  - ignored: Follows the volume of the global setting.
```
sushant-suse, line 4353:
```suggestion
                  - ignored: Use the global setting for offline replica rebuilding.
```

k8s/crds.yaml
```diff
@@ -4347,6 +4347,17 @@ spec:
...
+                  - enabled: Enable offline rebuilding for the volume, regardless of the global setting.
```
sushant-suse, line 4354:
```suggestion
                  - enabled: Enable offline rebuilding for this volume, regardless of the global setting.
```

k8s/crds.yaml
```diff
@@ -4347,6 +4347,17 @@ spec:
...
+                  - disable: Disable offline rebuilding of the volume, regardless of the global setting
```
sushant-suse, line 4355:
```suggestion
                  - disable: Disable offline rebuilding for this volume, regardless of the global setting.
```

sushant-suse (changes requested):

k8s/crds.yaml
mantissahz: Modified. Thanks

k8s/crds.yaml
mantissahz: Modified. Thanks

k8s/crds.yaml
mantissahz: Modified. Thanks

k8s/crds.yaml
mantissahz: Modified. Thanks

controller/volume_rebuilding_controller.go
mantissahz: We switched to monitoring the replica status instead of the node status.

controller/volume_rebuilding_controller.go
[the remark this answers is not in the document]
mantissahz: We switched to monitoring the replica status instead of the node status.

controller/volume_rebuilding_controller.go
```diff
@@ -73,6 +75,28 @@ func NewVolumeRebuildingController(
...
+	}, 0); err != nil {
+		return nil, err
+	}
+	vbc.cacheSyncs = append(vbc.cacheSyncs, ds.SettingInformer.HasSynced)
+
+	if _, err = ds.NodeInformer.AddEventHandlerWithResyncPeriod(cache.FilteringResourceEventHandler{
+		FilterFunc: isNodeReadyForOfflineRebuilding,
+		Handler: cache.ResourceEventHandlerFuncs{
+			AddFunc:    vbc.enqueueNode,
+			UpdateFunc: func(old, cur interface{}) { vbc.enqueueNode(cur) },
```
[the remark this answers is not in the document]
mantissahz: We switched to monitoring the replica status instead of the node status.

sushant-suse (approved):
LGTM

controller/volume_rebuilding_controller.go
```diff
@@ -96,6 +120,82 @@ func (vbc *VolumeRebuildingController) enqueueVolumeAfter(obj interface{}, durat
...
+		}
+		utilruntime.HandleError(fmt.Errorf("failed to get node: %v", err))
+		return
+	}
+	if n.Status.Conditions != nil {
+		nodeReadyCondition := types.GetCondition(n.Status.Conditions, longhorn.NodeConditionTypeReady)
+		if nodeReadyCondition.Status == longhorn.ConditionStatusTrue {
+			return
+		}
+	}
```
derekbit, line 194: Do we need to check node resource?
Can we check whether each replicas' spec.NodeID is equal to stataus.OwnerID?
If No, the volume should be enqueued.

controller/volume_rebuilding_controller.go
```diff
@@ -192,12 +292,117 @@ func (vbc *VolumeRebuildingController) reconcile(volName string) (err error) {
 	}()
 
 	rebuildingAttachmentTicketID := longhorn.GetAttachmentTicketID(longhorn.AttacherTypeVolumeRebuildingController, volName)
+	if vol.Status.Robustness == longhorn.VolumeRobustnessFaulted {
+		vbc.logger.Infof("Volume %v is faulted, skip rebuilding", volName)
```
derekbit, line 296:
```suggestion
		vbc.logger.Warnf("Volume %v is faulted, skip offline rebuilding", volName)
```

types/setting.go
```diff
@@ -1534,6 +1537,20 @@ var (
...
 		Default:     "false",
 	}
+
+	SettingDefinitionOfflineReplicaRebuilding = SettingDefinition{
+		DisplayName: "Offline Replica Rebuilding",
+		Description: "Enables automatic rebuilding of degraded replicas while the volume is detached. This setting only takes effect if the individual volume setting is set to `ignored`. \n\n" +
+			"Available options: \n\n" +
+			"- **true**: Enables offline replica rebuilding for all detached volumes, unless overridden by individual volume settings. \n\n" +
+			"- **false**: Disables offline replica rebuilding globally, unless overridden by individual volume settings. \n\n" +
+			"Note: Offline rebuilding applies only when a volume is detached. Volumes in a faulted state will not trigger offline rebuilding.",
```
c3y1huang, line 1547:
```suggestion
			**Note:** Offline rebuilding applies only when a volume is detached. Volumes in a faulted state will not trigger offline rebuilding.",
```

controller/volume_rebuilding_controller.go
```diff
@@ -96,6 +120,82 @@ func (vbc *VolumeRebuildingController) enqueueVolumeAfter(obj interface{}, durat
...
+	if !ok {
+		return false
+	}
+	return setting.Name == string(types.SettingNameOfflineReplicaRebuilding)
+}
+
+func (vbc *VolumeRebuildingController) enqueueSetting(obj interface{}) {
+	_, ok := obj.(*longhorn.Setting)
+	if !ok {
+		utilruntime.HandleError(fmt.Errorf("DeletedFinalStateUnknown contained non Setting object: %#v", obj))
```
c3y1huang, line 134: The error message needs to be fixed.

controller/volume_rebuilding_controller.go
```diff
@@ -96,6 +120,82 @@ func (vbc *VolumeRebuildingController) enqueueVolumeAfter(obj interface{}, durat
...
+		return true
+	}
+
+	return r.Spec.NodeID != r.Status.OwnerID
+}
+
+func (vbc *VolumeRebuildingController) enqueueVolumeForReplica(obj interface{}) {
+	r, ok := obj.(*longhorn.Replica)
+	if !ok {
+		utilruntime.HandleError(fmt.Errorf("DeletedFinalStateUnknown contained non SetNodeting object: %#v", obj))
```
c3y1huang, line 167: The error message needs to be fixed.

controller/volume_rebuilding_controller.go
```diff
@@ -96,6 +120,82 @@ func (vbc *VolumeRebuildingController) enqueueVolumeAfter(obj interface{}, durat
...
+		return
+	}
+	if v.Status.State != longhorn.VolumeStateDetached {
+		return
+	}
+
+	n, err := vbc.ds.GetNodeRO(r.Spec.NodeID)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			vbc.enqueueVolume(v)
```
c3y1huang, line 183: Why enqueue the volume when the node is not found?

controller/volume_rebuilding_controller.go
```diff
@@ -192,12 +292,117 @@ func (vbc *VolumeRebuildingController) reconcile(volName string) (err error) {
...
+
+func (vbc *VolumeRebuildingController) isVolumeReplicasHealthy(numberOfReplicas int, replicas map[string]*longhorn.Replica) (bool, error) {
+	healthyCount := 0
+	for _, replica := range replicas {
+		node, err := vbc.ds.GetNodeRO(replica.Spec.NodeID)
+		if err != nil {
+			if apierrors.IsNotFound(err) {
+				continue
+			}
+			return false, err
```
c3y1huang, line 366: Is the intention is to requeue when failing to get the node? An inline comment here might help clarify, as the different error handling may not be immediately obvious.

controller/volume_rebuilding_controller.go
```diff
@@ -73,6 +75,28 @@ func NewVolumeRebuildingController(
...
+		},
+	}, 0); err != nil {
+		return nil, err
+	}
+	vbc.cacheSyncs = append(vbc.cacheSyncs, ds.SettingInformer.HasSynced)
+
+	if _, err = ds.ReplicaInformer.AddEventHandlerWithResyncPeriod(cache.FilteringResourceEventHandler{
+		FilterFunc: isReplicaOwnerChanged,
+		Handler: cache.ResourceEventHandlerFuncs{
+			AddFunc:    vbc.enqueueVolumeForReplica,
```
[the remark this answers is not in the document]
mantissahz: Yes, we need it. I'm not sure why the informer uses `addFunc` when I shut down a node, and it updated the ownerID of the replica.

controller/volume_rebuilding_controller.go
mantissahz: Removed. Thanks.

controller/volume_rebuilding_controller.go
mantissahz: Modified. Thanks

types/setting.go

controller/volume_rebuilding_controller.go
mantissahz: Modified. Thanks

controller/volume_rebuilding_controller.go
mantissahz: Modified. Thanks

controller/volume_rebuilding_controller.go
mantissahz: Removed. We will check this in the controller. Thanks.

controller/volume_rebuilding_controller.go
mantissahz: Yes. Added an inline comment. Please review it again

controller/volume_rebuilding_controller.go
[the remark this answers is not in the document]
derekbit: Okay. This is due to the replica resource first appears in the informer.

controller/volume_rebuilding_controller.go
```diff
@@ -192,12 +279,134 @@ func (vbc *VolumeRebuildingController) reconcile(volName string) (err error) {
...
+			return nil
+		}
+		if vol.Status.State == longhorn.VolumeStateAttaching {
+			return nil
+		}
+
+		isVolumeInRebuilding, err := vbc.isVolumeReplicasRebuilding(vol)
+		if err != nil {
+			return err
+		}
```
derekbit, line 312: Should we delete the volumeattachment ticket and return `nil` if the volume or engine is not found?

controller/volume_rebuilding_controller.go
```diff
@@ -192,12 +279,134 @@ func (vbc *VolumeRebuildingController) reconcile(volName string) (err error) {
...
+func (vbc *VolumeRebuildingController) isVolumeReplicasRebuilding(vol *longhorn.Volume) (bool, error) {
+	engines, err := vbc.ds.ListVolumeEngines(vol.Name)
+	if err != nil {
+		return false, err
+	}
+	engine, err := vbc.ds.PickVolumeCurrentEngine(vol, engines)
+	if err != nil {
+		return false, err
+	}
+	if engine == nil || engine.Status.ReplicaModeMap == nil {
```
derekbit, line 393: Why return true while engine == nil?

controller/volume_rebuilding_controller.go
mantissahz: >Should we delete the volumeattachment ticket and return nil if the volume or engine is not found?

Yes, if engine is `nil`, we could delete the va ticket as it occurred an unknown error.
Modified. Thanks.

controller/volume_rebuilding_controller.go
mantissahz: Moved out the logic. And yes, it should not be `nil` unless an unexpected error.
Thanks.

controller/volume_rebuilding_controller.go
```diff
@@ -192,12 +279,134 @@ func (vbc *VolumeRebuildingController) reconcile(volName string) (err error) {
...
+		return nil, err
+	}
+	engine, err := vbc.ds.PickVolumeCurrentEngine(vol, engines)
+	if err != nil {
+		return nil, err
+	}
+	return engine, nil
+}
+
+func (vbc *VolumeRebuildingController) isVolumeReplicasRebuilding(vol *longhorn.Volume, engine *longhorn.Engine) (bool, error) {
```
c3y1huang, line 396: The `error` in return seems redundant.

controller/volume_rebuilding_controller.go
```diff
@@ -192,12 +279,134 @@ func (vbc *VolumeRebuildingController) reconcile(volName string) (err error) {
...
+func (vbc *VolumeRebuildingController) getVolumeEngine(vol *longhorn.Volume) (*longhorn.Engine, error) {
+	engines, err := vbc.ds.ListVolumeEngines(vol.Name)
+	if err != nil {
+		return nil, err
+	}
+	engine, err := vbc.ds.PickVolumeCurrentEngine(vol, engines)
+	if err != nil {
+		return nil, err
+	}
+	return engine, nil
```
c3y1huang, line 393:
```suggestion
	return vbc.ds.PickVolumeCurrentEngine(vol, engines)
```

controller/volume_rebuilding_controller.go
```diff
@@ -192,12 +279,134 @@ func (vbc *VolumeRebuildingController) reconcile(volName string) (err error) {
...
+	if err != nil {
+		return va, err
+	}
+	if !areAllReplicasHealthy {
+		createOrUpdateAttachmentTicket(va, attachmentID, vol.Status.OwnerID, longhorn.AnyValue, longhorn.AttacherTypeVolumeRebuildingController)
+	}
+	return va, nil
+}
+
+func (vbc *VolumeRebuildingController) isVolumeReplicasHealthy(numberOfReplicas int, replicas map[string]*longhorn.Replica) (bool, error) {
```
c3y1huang, line 363: The `error` in return seems redundant.

controller/volume_rebuilding_controller.go
```diff
@@ -192,12 +279,134 @@ func (vbc *VolumeRebuildingController) reconcile(volName string) (err error) {
...
+}
+
+func (vbc *VolumeRebuildingController) getVolumeEngine(vol *longhorn.Volume) (*longhorn.Engine, error) {
+	engines, err := vbc.ds.ListVolumeEngines(vol.Name)
+	if err != nil {
+		return nil, err
+	}
+	engine, err := vbc.ds.PickVolumeCurrentEngine(vol, engines)
+	if err != nil {
+		return nil, err
```
derekbit, line 391: Can the error be `not found`?

controller/volume_rebuilding_controller.go
```diff
@@ -192,12 +279,134 @@ func (vbc *VolumeRebuildingController) reconcile(volName string) (err error) {
...
+			return nil
+		}
+		if engine.Status.ReplicaModeMap == nil {
+			// wait for engine status synced
+			return nil
+		}
+
+		isVolumeInRebuilding, err := vbc.isVolumeReplicasRebuilding(vol, engine)
+		if err != nil {
+			return err
```
derekbit, line 325: Do we need to remove the volumeattachment ticket if encountering an error?

controller/volume_rebuilding_controller.go
```diff
@@ -192,12 +279,134 @@ func (vbc *VolumeRebuildingController) reconcile(volName string) (err error) {
...
+			}
+			return nil
+		}
+		if vol.Status.State == longhorn.VolumeStateAttaching {
+			return nil
+		}
+
+		engine, err := vbc.getVolumeEngine(vol)
+		if err != nil {
+			return err
```
derekbit, line 311: Do we need to remove the volumeattachment ticket if encountering an error?

controller/volume_rebuilding_controller.go
```diff
@@ -192,12 +279,134 @@ func (vbc *VolumeRebuildingController) reconcile(volName string) (err error) {
...
-	delete(va.Spec.AttachmentTickets, rebuildingAttachmentTicketID)
+	isOfflineRebuildEnabled, err := vbc.isVolumeOfflineRebuildEnabled(vol.Spec.OfflineRebuilding)
+	if err != nil {
+		return err
+	}
+	if isOfflineRebuildEnabled && !util.IsHigherPriorityVATicketExisting(va, longhorn.AttacherTypeVolumeRebuildingController) {
+		if vol.Status.State == longhorn.VolumeStateDetached {
+			va, err = vbc.syncLHVolumeAttachmentForOfflineRebuild(vol, va, rebuildingAttachmentTicketID)
+			if err != nil {
+				return err
```
derekbit, line 301: Do we need to remove the volumeattachment ticket if encountering an error?

controller/volume_rebuilding_controller.go
mantissahz: Removed. Thanks.

controller/volume_rebuilding_controller.go
mantissahz: Modified. Thanks

controller/volume_rebuilding_controller.go
mantissahz: Removed. Thanks.

constant/events.go
```diff
@@ -33,6 +33,8 @@ const (
 	EventReasonSucceededExpansion = "SucceededExpansion"
 	EventReasonCanceledExpansion  = "CanceledExpansion"
 
+	EventReasonCanceledOfflineRebuilding = "CanceledOfflineRebuilding"
```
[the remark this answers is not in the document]
mantissahz: Removed. Thanks.

controller/volume_rebuilding_controller.go
```diff
@@ -192,12 +279,134 @@ func (vbc *VolumeRebuildingController) reconcile(volName string) (err error) {
...
+	delete(va.Spec.AttachmentTickets, rebuildingAttachmentTicketID)
 	return nil
 }
 
+func (vbc *VolumeRebuildingController) isVolumeOfflineRebuildEnabled(offlineRebuilding longhorn.VolumeOfflineRebuilding) (bool, error) {
+	if offlineRebuilding == longhorn.VolumeOfflineRebuildingEnabled {
+		return true, nil
+	}
+
+	offlineReplicaRebuildingValue, err := vbc.ds.GetSettingAsBool(types.SettingNameOfflineReplicaRebuilding)
```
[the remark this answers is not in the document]
mantissahz: Renamed. Thanks

controller/volume_rebuilding_controller.go
```diff
@@ -192,12 +279,134 @@ func (vbc *VolumeRebuildingController) reconcile(volName string) (err error) {
...
+	}
 
-	delete(va.Spec.AttachmentTickets, rebuildingAttachmentTicketID)
+	isOfflineRebuildEnabled, err := vbc.isVolumeOfflineRebuildEnabled(vol.Spec.OfflineRebuilding)
+	if err != nil {
+		return err
+	}
+	if isOfflineRebuildEnabled && !util.IsHigherPriorityVATicketExisting(va, longhorn.AttacherTypeVolumeRebuildingController) {
+		if vol.Status.State == longhorn.VolumeStateDetached {
+			va, err = vbc.syncLHVolumeAttachmentForOfflineRebuild(vol, va, rebuildingAttachmentTicketID)
```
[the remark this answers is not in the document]
mantissahz: We don't need to return it, and it was replaced by `-`. Thanks.

controller/volume_rebuilding_controller.go
mantissahz: No, it picks the engine from `engines`, and if it is unable to find the engine, it returns a customized error, https://github.com/longhorn/longhorn-manager/blob/c313e6d7666f58a09222075b906e34348938b4cb/datastore/longhorn.go#L1285-L1287

controller/volume_rebuilding_controller.go
mantissahz: Added the defer function to delete the VA ticket if the following code flow returns an error.

controller/volume_rebuilding_controller.go
mantissahz: Added the defer function to delete the VA ticket if the following code flow returns an error. Thanks

controller/volume_rebuilding_controller.go
mantissahz: Added the defer function to delete the VA ticket if the following code flow returns an error. Thanks

derekbit:
@coderabbitai review

controller/volume_rebuilding_controller.go
```diff
@@ -96,6 +120,69 @@ func (vbc *VolumeRebuildingController) enqueueVolumeAfter(obj interface{}, durat
...
+	vs, err := vbc.ds.ListVolumesRO()
+	if err != nil {
+		utilruntime.HandleError(fmt.Errorf("failed to list volumes: %v", err))
+		return
+	}
+
+	for _, v := range vs {
+		if v.Spec.OfflineRebuilding != longhorn.VolumeOfflineRebuildingIgnored {
+			continue
+		}
```
derekbit, line 147: Should enqueue `enabled` and `ignored`. Right?

```suggestion
		if v.Spec.OfflineRebuilding == longhorn.VolumeOfflineRebuildingDisabled {
			continue
		}
```

controller/volume_rebuilding_controller.go
```diff
@@ -192,12 +279,130 @@ func (vbc *VolumeRebuildingController) reconcile(volName string) (err error) {
 	}()
 
 	rebuildingAttachmentTicketID := longhorn.GetAttachmentTicketID(longhorn.AttacherTypeVolumeRebuildingController, volName)
+	defer func() {
+		if err != nil {
+			delete(va.Spec.AttachmentTickets, rebuildingAttachmentTicketID)
+			return
+		}
+	}()
```
derekbit, line 287: deleteVATicketRequired := true
```suggestion
	defer func() {
        if deleteVATicketRequired {
			delete(va.Spec.AttachmentTickets, rebuildingAttachmentTicketID)
		}
	}()
```

controller/volume_rebuilding_controller.go
```diff
@@ -192,12 +279,130 @@ func (vbc *VolumeRebuildingController) reconcile(volName string) (err error) {
...
+	if !vol.DeletionTimestamp.IsZero() {
+		vbc.logger.Infof("Volume %v is deleting, skip offline rebuilding", volName)
+		delete(va.Spec.AttachmentTickets, rebuildingAttachmentTicketID)
+		return nil
+	}
+	if vol.Status.Robustness == longhorn.VolumeRobustnessFaulted {
+		vbc.logger.Warnf("Volume %v is faulted, skip offline rebuilding", volName)
+		delete(va.Spec.AttachmentTickets, rebuildingAttachmentTicketID)
+		return nil
+	}
```
derekbit, line 297:
```suggestion
	if !vol.DeletionTimestamp.IsZero() {
		vbc.logger.Infof("Volume %v is deleting, skip offline rebuilding", volName)
		return nil
	}
	if vol.Status.Robustness == longhorn.VolumeRobustnessFaulted {
		vbc.logger.Warnf("Volume %v is faulted, skip offline rebuilding", volName)
		return nil
	}
```

controller/volume_rebuilding_controller.go
```diff
@@ -192,12 +279,130 @@ func (vbc *VolumeRebuildingController) reconcile(volName string) (err error) {
...
 
-	delete(va.Spec.AttachmentTickets, rebuildingAttachmentTicketID)
+	isOfflineRebuildEnabled, err := vbc.isVolumeOfflineRebuildEnabled(vol.Spec.OfflineRebuilding)
+	if err != nil {
+		return err
```
derekbit, line 301: Should delete the VA ticket if err != nil. Right?

controller/volume_rebuilding_controller.go
```diff
@@ -192,12 +279,130 @@ func (vbc *VolumeRebuildingController) reconcile(volName string) (err error) {
...
+	if isOfflineRebuildEnabled && !util.IsHigherPriorityVATicketExisting(va, longhorn.AttacherTypeVolumeRebuildingController) {
+		if vol.Status.State == longhorn.VolumeStateDetached {
+			_, err = vbc.syncLHVolumeAttachmentForOfflineRebuild(vol, va, rebuildingAttachmentTicketID)
+			if err != nil {
+				return err
+			}
+			return nil
+		}
+		if vol.Status.State == longhorn.VolumeStateAttaching {
+			return nil
```
derekbit, line 312:
```suggestion
            deleteVATicketRequired = false
			return nil
```

controller/volume_rebuilding_controller.go
```diff
@@ -192,12 +279,130 @@ func (vbc *VolumeRebuildingController) reconcile(volName string) (err error) {
...
+	if err != nil {
+		return err
+	}
+	if isOfflineRebuildEnabled && !util.IsHigherPriorityVATicketExisting(va, longhorn.AttacherTypeVolumeRebuildingController) {
+		if vol.Status.State == longhorn.VolumeStateDetached {
+			_, err = vbc.syncLHVolumeAttachmentForOfflineRebuild(vol, va, rebuildingAttachmentTicketID)
+			if err != nil {
+				return err
+			}
+			return nil
```
derekbit, line 309:
```suggestion
            deleteVATicketRequired = false
			return nil
```

controller/volume_rebuilding_controller.go
```diff
@@ -192,12 +279,130 @@ func (vbc *VolumeRebuildingController) reconcile(volName string) (err error) {
...
 
+		engine, err := vbc.getVolumeEngine(vol)
+		if err != nil {
+			return err
+		}
+		if engine == nil {
+			vbc.logger.Warnf("Volume %v engine not found, skip offline rebuilding", volName)
+			delete(va.Spec.AttachmentTickets, rebuildingAttachmentTicketID)
+			return nil
+		}
```
derekbit, line 323:
```suggestion
		if engine == nil {
			vbc.logger.Warnf("Volume %v engine not found, skip offline rebuilding", volName)
			return nil
		}
```

controller/volume_rebuilding_controller.go
```diff
@@ -192,12 +279,130 @@ func (vbc *VolumeRebuildingController) reconcile(volName string) (err error) {
...
+		if engine.Status.ReplicaModeMap == nil {
+			// wait for engine status synced
+			return nil
+		}
+
+		if vbc.isVolumeReplicasRebuilding(vol, engine) {
+			return nil
+		}
```
derekbit, line 331:
```suggestion
        deleteVATicketRequired = false

		if engine.Status.ReplicaModeMap == nil {
			// wait for engine status synced
			return nil
		}

		if vbc.isVolumeReplicasRebuilding(vol, engine) {
			return nil
		}
```

k8s/crds.yaml
```diff
@@ -4364,6 +4364,17 @@ spec:
...
+                description: |-
+                  Controls whether Longhorn should rebuild replicas while the volume is offline.
+                  - ignored: Use the global setting for offline replica rebuilding.
+                  - enabled: Enable offline rebuilding for this volume, regardless of the global setting.
+                  - disable: Disable offline rebuilding for this volume, regardless of the global setting
+                enum:
+                - ignored
+                - disabled
+                - enabled
+                type: string
```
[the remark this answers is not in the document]
mantissahz: good catch, modified. Thanks.

api/model.go
```diff
@@ -403,6 +404,10 @@ type UpdateBackupTargetInput struct {
 	BackupTargetName string `json:"backupTargetName"`
 }
 
+type UpdateOfflineRebuildingInput struct {
+	OfflineRebuilding string `json:"offlineRebuilding"`
+}
+
```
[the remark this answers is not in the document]
mantissahz: good catch, modified. Thanks.

upgrade/v18xto190/upgrade.go
```diff
@@ -50,5 +56,32 @@ func UpgradeResources(namespace string, lhClient *lhclientset.Clientset, kubeCli
...
+		return errors.Wrapf(err, "failed to list all existing Longhorn volumes during the volume upgrade")
+	}
+
+	for _, v := range volumesMap {
+		if v.Spec.OfflineRebuilding == "" {
+			v.Spec.OfflineRebuilding = longhorn.VolumeOfflineRebuildingIgnored
+		}
+	}
+
 	return nil
```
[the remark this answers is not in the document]
mantissahz: It would be updated by the caller.

controller/volume_rebuilding_controller.go
```diff
@@ -192,12 +279,130 @@ func (vbc *VolumeRebuildingController) reconcile(volName string) (err error) {
...
+	replicas, err := vbc.ds.ListVolumeReplicasRO(vol.Name)
+	if err != nil {
+		return va, err
+	}
+
+	if !vbc.isVolumeReplicasHealthy(vol.Spec.NumberOfReplicas, replicas) {
+		createOrUpdateAttachmentTicket(va, attachmentID, vol.Status.OwnerID, longhorn.AnyValue, longhorn.AttacherTypeVolumeRebuildingController)
+	}
+	return va, nil
+}
```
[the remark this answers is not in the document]
mantissahz: modified. Thanks.

k8s/pkg/apis/longhorn/v1beta2/volume.go
```diff
@@ -291,6 +299,13 @@ type VolumeSpec struct {
 	// The backup target name that the volume will be backed up to or is synced.
 	// +optional
 	BackupTargetName string `json:"backupTargetName"`
+	// +kubebuilder:validation:Enum=ignored;disabled;enabled
+	// Controls whether Longhorn should rebuild replicas while the volume is offline.
```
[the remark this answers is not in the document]
mantissahz: Modified. Thanks.

types/setting.go
```diff
@@ -1534,6 +1537,20 @@ var (
 		ReadOnly:    false,
 		Default:     "false",
 	}
+
+	SettingDefinitionOfflineReplicaRebuilding = SettingDefinition{
+		DisplayName: "Offline Replica Rebuilding",
+		Description: "Enables automatic rebuilding of degraded replicas while the volume is detached. This setting only takes effect if the individual volume setting is set to `ignored`. \n\n" +
```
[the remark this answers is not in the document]

controller/volume_rebuilding_controller.go
mantissahz: Modified. Thanks.

controller/volume_rebuilding_controller.go
mantissahz: Removed. Thanks.

controller/volume_rebuilding_controller.go
mantissahz: Yes.

controller/volume_rebuilding_controller.go
mantissahz: Added. Thanks.

controller/volume_rebuilding_controller.go
mantissahz: Added. Thanks.

controller/volume_rebuilding_controller.go
mantissahz: Removed. Thanks

controller/volume_rebuilding_controller.go
mantissahz: Added. Thanks

k8s/crds.yaml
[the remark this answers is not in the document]
derekbit: Cool. coderabbitai can use emoji.

controller/volume_rebuilding_controller.go
```diff
@@ -192,12 +279,134 @@ func (vbc *VolumeRebuildingController) reconcile(volName string) (err error) {
...
+		if engine.Status.ReplicaModeMap == nil {
+			// wait for engine status synced
+			return nil
+		}
+		if vbc.isVolumeReplicasRebuilding(vol, engine) {
+			return nil
+		}
+	}
+
+	deleteVATicketRequired = true
```
derekbit, line 335: Is it redundant?

mantissahz: No, because theline 325, we set the flag to false.

derekbit, line 335:
```suggestion
		if engine.Status.ReplicaModeMap == nil {
			// wait for engine status synced
			deleteVATicketRequired = false
			return nil
		}
		if vbc.isVolumeReplicasRebuilding(vol, engine) {
			deleteVATicketRequired = false
			return nil
		}
	}
```

mantissahz: Modified the implementation. Thanks

derekbit (approved):
LGTM

PhanLe1010:
Sorry for the delay.
The updated PR LGTM

PR #3649, opened 2025-03-26
chore: fix lint

COLDTURNIP:

Issue longhorn/longhorn#10639

Fix lint errors and pass all validations in CI.

### DIFF
--- a/client/common.go
+++ b/client/common.go
@@ -165,7 +165,11 @@ func setupRancherBaseClient(rancherClient *RancherBaseClientImpl, opts *ClientOp
 		return err
 	}
 
-	defer resp.Body.Close()
+	defer func(body io.Closer) {
+		if closeErr := body.Close(); closeErr != nil && debug {
+			fmt.Printf("Warning: failed to close schema response body: %v\n", closeErr)
+		}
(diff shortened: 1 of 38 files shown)


COLDTURNIP:
Added logs for deferred resource closing.

client/common.go
```diff
@@ -165,7 +165,11 @@ func setupRancherBaseClient(rancherClient *RancherBaseClientImpl, opts *ClientOp
 		return err
 	}
 
-	defer resp.Body.Close()
+	defer func(body io.Closer) {
+		if closeErr := body.Close(); closeErr != nil && debug {
+			fmt.Printf("Warning: failed to close schema response body: %v\n", closeErr)
```
derekbit, line 170:
```suggestion
			logrus.WithError(closeErr).Warn("Failed to close schema response body")
```

derekbit: Apply to other changes.

COLDTURNIP: Agree to use logrus instead, but, do we have any concern remove debug flag and change to logrus in the rest part of client/common.go?

derekbit (approved):
LGTM

PR #3651, opened 2025-03-28
feat: add strict field validation to the upgrade operaiton in upgrade path

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/10644

The upgrade path will fail if the CRD field is unknown during the upgrade, which can occur when the new CRD is applied after the new Longhorn manager starts.


derekbit (approved):
LGTM

COLDTURNIP (approved):
LGTM

PR #3660, opened 2025-03-31
feat(setting): remove ValueIntRangeMaximum of v2-data-engine-guaranteed-instance-manager-cpu

derekbit:

Issue Longhorn/longhorn#10662


mantissahz:
https://github.com/longhorn/longhorn-manager/blob/c396e03839a20f8d10a3fbd85c93b1b3c7a83afa/types/types.go#L1208-L1214

Is it (line1211) still needed here?

derekbit:
> https://github.com/longhorn/longhorn-manager/blob/c396e03839a20f8d10a3fbd85c93b1b3c7a83afa/types/types.go#L1208-L1214
> 
> Is it (line1211) still needed here?

Good catch. Fixed. Thank you.

mantissahz (approved):
LGTM

PR #3669, opened 2025-04-07
fix: wrapping a nil error is misleading

derekbit:

Issue Longhorn/longhorn#10495


mantissahz (approved):

derekbit:

PR #3684, opened 2025-04-14
fix(backup): delete snapshot after backup completed

mantissahz:

Issue # longhorn/longhorn#9213

Delete the backup snapshot after the backup is completed and the backup status is updated correctly.

### DIFF
--- a/controller/backup_controller.go
+++ b/controller/backup_controller.go
@@ -405,6 +405,10 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
 				log.Warnf("failed to sync backup volume %v for backup target %v", canonicalBackupVolumeName, backupTargetName)
 				return
 			}
+			if err := bc.deleteSnapshotAfterBackupCompleted(backup); err != nil {
+				log.WithError(err).Warn("Failed to delete snapshot after backup completed")
+				return
+			}
 		}
 	}()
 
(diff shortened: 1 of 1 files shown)


controller/backup_controller.go
```diff
@@ -405,6 +405,10 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
 				log.Warnf("failed to sync backup volume %v for backup target %v", canonicalBackupVolumeName, backupTargetName)
 				return
 			}
+			if err := bc.deleteSnasphotAfterBackupCompleted(backup); err != nil {
```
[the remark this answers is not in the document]
mantissahz: good catch. Modified. Thanks

controller/backup_controller.go
```diff
@@ -1174,3 +1157,25 @@ func (bc *BackupController) setInprogressDeletionMap(backupURL string, state lon
...
+	if err != nil {
+		bc.logger.WithError(err).Errorf("Failed to get the setting %v", types.SettingNameAutoCleanupSnapshotAfterOnDemandBackupCompleted)
+		return err
+	}
+	if cleanupSnap && backup.Spec.SnapshotName != "" && backup.Status.State == longhorn.BackupStateCompleted {
+		if errDel := bc.ds.DeleteSnapshot(backup.Spec.SnapshotName); errDel != nil && !apierrors.IsNotFound(errDel) {
+			err = errDel
+			bc.logger.WithError(err).Error("Failed to delete the snapshot")
+			return err
+		}
```
derekbit, line 1178: Why do we need `errDel`?

mantissahz: Yes, we don't need it. Removed. Thanks.

PR #3685, opened 2025-04-14
fix: fix naming collision when creating bim based on the prefix of the disk uuid

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/10616

When generating the disk uuid, we first get all existing disk uuid to check and prevent it from having the same first-4-char prefix.

### DIFF
--- a/controller/monitor/disk_utils.go
+++ b/controller/monitor/disk_utils.go
@@ -14,6 +14,7 @@ import (
 	lhns "github.com/longhorn/go-common-libs/ns"
 	lhtypes "github.com/longhorn/go-common-libs/types"
 
+	"github.com/longhorn/longhorn-manager/datastore"
 	"github.com/longhorn/longhorn-manager/engineapi"
 	"github.com/longhorn/longhorn-manager/util"
 
@@ -21,7 +22,8 @@ import (
 )
 
 const (
-	defaultBlockSize = 512
+	defaultBlockSize      = 512
+	uuidGenerationRetries = 20
 )
 
 // GetDiskStat returns the disk stat of the given directory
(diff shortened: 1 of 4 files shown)


datastore/longhorn.go
```diff
@@ -6140,3 +6140,19 @@ func (s *DataStore) IsStorageNetworkForRWXVolume() (bool, error) {
...
+	firstFourCharSet := make(map[string]bool)
+	nodes, err := s.ListNodesRO()
+	if err != nil {
+		return nil, err
+	}
+
+	for _, node := range nodes {
+		for _, diskStatus := range node.Status.DiskStatus {
+			firstFourCharSet[diskStatus.DiskUUID[:4]] = true
+		}
```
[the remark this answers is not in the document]
ChanYiLin: It will always larger than 4

controller/monitor/disk_utils.go
```diff
@@ -224,3 +235,14 @@ func getSpdkReplicaInstanceNames(client *DiskServiceClient, diskType, diskName,
 
 	return instanceNames, nil
 }
+
+func generateUniqueFirstFourCharUUID(allDiskUUIDFirstFourCharSet map[string]bool) (string, error) {
+	for i := 0; i < 20; i++ {
```
derekbit, line 240:
```suggestion
	for i := 0; i < uuidGenerationRetries; i++ {
```

controller/monitor/disk_utils.go
```diff
@@ -224,3 +235,14 @@ func getSpdkReplicaInstanceNames(client *DiskServiceClient, diskType, diskName,
...
+		uuid := util.UUID()
+		prefix := uuid[:4]
```
derekbit, line 242: Checked the `[google/uuid](https://github.com/google/uuid)`, `uuid.New()` is based UUID v4. The chance of a prefix collision between two consecutive UUIDs is very low, so I'm okay with relying on retries.

controller/monitor/disk_utils.go
```diff
@@ -224,3 +235,14 @@ func getSpdkReplicaInstanceNames(client *DiskServiceClient, diskType, diskName,
...
+		if !allDiskUUIDFirstFourCharSet[prefix] {
+			return uuid, nil
+		}
+	}
+	return "", errors.New("failed to generate UUID with unique prefix after 20 attempts")
```
derekbit, line 247:
```suggestion
	return "", errors.New("failed to generate UUID with unique prefix after %v attempts", uuidGenerationRetries)
```

derekbit:
One question: how to handle the existing disks?

ChanYiLin:
@derekbit 
For the existing disks we can only request users to do the workaround
Because if we change the diskUUID internally, there will be lots of things to do, like evicting the replicas
And I think asking users to remove the disks manually and re-add it back is more safe.
WDYT?

derekbit:
> @derekbit For the existing disks we can only request users to do the workaround Because if we change the diskUUID internally, there will be lots of things to do, like evicting the replicas And I think asking users to remove the disks manually and re-add it back is more safe. WDYT?

WDYT? @WebberHuang1118

controller/monitor/disk_utils.go
ChanYiLin: fixed.

controller/monitor/disk_utils.go
ChanYiLin: fixed.

derekbit (approved):

shuo-wu (approved):
LGTM

derekbit:

PR #3686, opened 2025-04-15
fix: safely delete VolumeAttachment for force-deleted pods

c3y1huang:

Issue longhorn/longhorn#10689

Add logic to determine whether a VolumeAttachment associated with a force-deleted pod should be cleaned up. This ensures that stale attachments can be removed safely.

`None`

https://github.com/longhorn/longhorn/issues/9248#issuecomment-2800206182

### DIFF
--- a/controller/kubernetes_pod_controller.go
+++ b/controller/kubernetes_pod_controller.go
@@ -329,6 +329,10 @@ func (kc *KubernetesPodController) handleWorkloadPodDeletionIfCSIPluginPodIsDown
 // cleanupForceDeletedPodResources removes stale resources left behind when a pod
 // is force-deleted (i.e., deletion grace period is zero).
 func (kc *KubernetesPodController) cleanupForceDeletedPodResources(pod *corev1.Pod) error {
+	if !isControllerResponsibleFor(kc.controllerID, kc.ds, pod.Name, "", pod.Spec.NodeName) {
+		return nil
+	}
+
 	if pod.DeletionTimestamp == nil {
 		return nil
 	}
--- a/controller/kubernetes_pod_controller.go
+++ b/controller/kubernetes_pod_controller.go
@@ -3,6 +3,7 @@ package controller
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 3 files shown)


c3y1huang:
@coderabbitai review

c3y1huang:
Test Results:
- Restart Cluster While Workload Heavy Writing
  - https://ci.longhorn.io/job/private/job/longhorn-e2e-test/2600/
  - https://ci.longhorn.io/job/private/job/longhorn-e2e-test/2601/
  - https://ci.longhorn.io/job/private/job/longhorn-e2e-test/2608/
- Reboot Node One By One While Workload Heavy Writing
  - https://ci.longhorn.io/job/private/job/longhorn-e2e-test/2602/
  - https://ci.longhorn.io/job/private/job/longhorn-e2e-test/2603/
  - https://ci.longhorn.io/job/private/job/longhorn-e2e-test/2609/
- Node Down And Longhorn Not Force Delete Terminating Deployment Pod
  - https://ci.longhorn.io/job/private/job/longhorn-e2e-test/2604/
  - https://ci.longhorn.io/job/private/job/longhorn-e2e-test/2605/

c3y1huang:
https://github.com/coderabbitai review

PhanLe1010 (commented):
In gerneral LGTM. 

Quick questions:
1. How do we know which pod is being forced delete?
1. Is it correct that the new improvement here only benerfit when `node-down-pod-deletion-policy` is `do-nothing` (or a value which does not apply to thw workload type) AND user forcefully delete a pod?

c3y1huang:
> * How do we know which pod is being forced delete?

We identify a stale K8s VA that is safe to clean up based on two key indicators:
1. The deleted pod has `deletionGracePeriodSeconds == 0`, meaning it was forcefully deleted.
1. Another pod using the same PVC triggers an event with a `Multi-Attach error`, indicating that multiple pods are attempting to use the same RWO volume, but the VA still exists and has not been properly cleaned up by Kubernetes.

> * Is it correct that the new improvement here only benerfit when `node-down-pod-deletion-policy` is `do-nothing` (or a value which does not apply to thw workload type) AND user forcefully delete a pod?

This improvement applies regardless of the `node-down-pod-deletion-policy` setting. The goal is to address scenarios where the pod hangs during creation due to a multi-attached volume caused by a stale VA.

mantissahz (approved):

c3y1huang:
@mergify backport v1.7.x v1.8.x

PR #3694, opened 2025-04-17
feat: orphaned runtime instance cleanup

COLDTURNIP:

Issue longhorn/longhorn#6764

Implement orphaned runtime instance cleanup.
LEP: longhorn/longhorn#10675

### DIFF
--- a/controller/instance_manager_controller.go
+++ b/controller/instance_manager_controller.go
@@ -87,6 +87,8 @@ type InstanceManagerMonitor struct {
 	proxyClient engineapi.EngineClientProxy
 }
 
+type instanceProcessMap map[string]longhorn.InstanceProcess
+
 func updateInstanceManagerVersion(im *longhorn.InstanceManager) error {
 	cli, err := engineapi.NewInstanceManagerClient(im, false)
 	if err != nil {
@@ -170,6 +172,13 @@ func NewInstanceManagerController(
 	}
 	imc.cacheSyncs = append(imc.cacheSyncs, ds.KubeNodeInformer.HasSynced)
 
+	if _, err = ds.OrphanInformer.AddEventHandlerWithResyncPeriod(cache.ResourceEventHandlerFuncs{
+		AddFunc: imc.enqueueInstanceManagerOrphan,
+	}, 0); err != nil {
+		return nil, err
+	}
+	imc.cacheSyncs = append(imc.cacheSyncs, ds.OrphanInformer.HasSynced)
+
 	if _, err = ds.SettingInformer.AddEventHandlerWithResyncPeriod(
 		cache.FilteringResourceEventHandler{
 			FilterFunc: imc.isResponsibleForSetting,
@@ -200,7 +209,8 @@ func (imc *InstanceManagerController) isResponsibleForSetting(obj interface{}) b
 	}
 
 	return types.SettingName(setting.Name) == types.SettingNameKubernetesClusterAutoscalerEnabled ||
-		types.SettingName(setting.Name) == types.SettingNameV2DataEngineCPUMask
+		types.SettingName(setting.Name) == types.SettingNameV2DataEngineCPUMask ||
+		types.SettingName(setting.Name) == types.SettingNameOrphanResourceAutoDeletion
 }
 
 func isInstanceManagerPod(obj interface{}) bool {
@@ -333,6 +343,7 @@ func (imc *InstanceManagerController) syncInstanceManager(key string) (err error
 
 	if im.DeletionTimestamp != nil {
 		log.Warnf("Deleting instance manager pod %v since the instance manager is being deleted", name)
+		// the orphan instance CRs will be also deleted because of owner reference
 		return imc.cleanupInstanceManagerPod(im.Name)
 	}
 
@@ -386,6 +397,10 @@ func (imc *InstanceManagerController) syncInstanceManager(key string) (err error
 	}
 
 	if err := imc.syncMonitor(im); err != nil {
+		return err
+	}
+
+	if err := imc.syncOrphans(im); err != nil {
 		return err
 	}
 
@@ -988,6 +1003,32 @@ func (imc *InstanceManagerController) deleteInstanceManagerPDB(im *longhorn.Inst
 	return nil
 }
 
+func (imc *InstanceManagerController) syncOrphans(im *longhorn.InstanceManager) error {
+	// Instances are live inside the instance manager pod.
+	// Remove the instance orphan CRs when the instance manager pod is deleted or not running.
+	var isInstanceManagerPodDeletedOrNotRunning bool
+	if im.Status.CurrentState != longhorn.InstanceManagerStateRunning {
+		isInstanceManagerPodDeletedOrNotRunning = true
+	} else {
+		switch pod, err := imc.ds.GetPodRO(im.Namespace, im.Name); {
+		case datastore.ErrorIsNotFound(err):
+			isInstanceManagerPodDeletedOrNotRunning = true
+		case err != nil:
+			return errors.Wrapf(err, "failed to check instance manager %v pod state while sync orphans", im.Name)
+		default:
+			isInstanceManagerPodDeletedOrNotRunning = pod == nil
+		}
+	}
+
+	// Once the original owner controller lost the ownership of the instance manager CR, it is no longer able to monitor the instance status.
+	// This may be caused by network outage or Longhorn manager failure.
+	// Remove the corresponding orphan CRs, and they will be resync after the owner instance manager gets back.
+	instanceManagerOwnershipChanged := im.Spec.NodeID != imc.controllerID
+
+	isInstanceManagerTerminating := isInstanceManagerPodDeletedOrNotRunning || instanceManagerOwnershipChanged
+	return imc.deleteOrphans(im, isInstanceManagerTerminating)
+}
+
 const ITERATE_NAME_LIMIT = 5
 
 func formatInstanceMessage(im *longhorn.InstanceManager) string {
@@ -1245,6 +1286,15 @@ func (imc *InstanceManagerController) enqueueKubernetesNode(obj interface{}) {
 	}
 
 	imc.enqueueInstanceManagersForNode(kubernetesNode.Name)
+}
+
+func (imc *InstanceManagerController) enqueueInstanceManagerOrphan(obj interface{}) {
+	orphan, ok := obj.(*longhorn.Orphan)
+	if !ok {
+		return
+	}
+
+	imc.enqueueInstanceManagersForNode(orphan.Spec.NodeID)
 }
 
 func (imc *InstanceManagerController) enqueueInstanceManagersForNode(nodeName string) {
@@ -1610,6 +1660,123 @@ func (imc *InstanceManagerController) createInstanceManagerPodSpec(im *longhorn.
 	return podSpec, nil
 }
 
+func (imc *InstanceManagerController) deleteOrphans(im *longhorn.InstanceManager, isInstanceManagerTerminating bool) error {
+	autoDeletionTypes, err := imc.ds.GetSettingOrphanResourceAutoDeletion()
+	var autoDeleteEngine = false
+	var autoDeleteReplica = false
+	if err != nil {
+		imc.logger.WithError(err).Warnf("Failed to fetch orphan auto deletion setting, disabled by default")
+	} else {
+		autoDeleteEngine = autoDeletionTypes[types.OrphanResourceTypeEngineInstance]
+		autoDeleteReplica = autoDeletionTypes[types.OrphanResourceTypeReplicaInstance]
+	}
+
+	orphanList, err := imc.ds.ListOrphansForEngineAndReplicaInstancesRO(im.Spec.NodeID)
+	if err != nil {
+		return err
+	}
+	multiError := util.NewMultiError()
+	for _, orphan := range orphanList {
+		if !orphan.DeletionTimestamp.IsZero() {
+			continue
+		}
+
+		var isDeletable = false
+		switch orphan.Spec.Type {
+		case longhorn.OrphanTypeEngineInstance:
+			isDeletable, err = imc.isOrphanDeletable(orphan, isInstanceManagerTerminating || autoDeleteEngine, im.Status.InstanceEngines, imc.isEngineScheduledOnCurrentNode)
+		case longhorn.OrphanTypeReplicaInstance:
+			isDeletable, err = imc.isOrphanDeletable(orphan, isInstanceManagerTerminating || autoDeleteReplica, im.Status.InstanceReplicas, imc.isReplicaScheduledOnCurrentNode)
+		}
+		if err != nil {
+			multiError.Append(util.NewMultiError(fmt.Sprintf("%v: %v", orphan.Name, err)))
+		} else if isDeletable {
+			if err := imc.ds.DeleteOrphan(orphan.Name); err != nil && !apierrors.IsNotFound(err) {
+				multiError.Append(util.NewMultiError(fmt.Sprintf("%v: %v", orphan.Name, err)))
+			}
+		}
+	}
+	if len(multiError) > 0 {
+		return fmt.Errorf("instance manager failed to delete orphan CR: %v", multiError.Join())
+	}
+	return nil
+}
+
+func (imc *InstanceManagerController) isOrphanDeletable(orphan *longhorn.Orphan, autoDelete bool, instanceMap instanceProcessMap, isInstanceScheduledOnNode func(node, instance string) (bool, error)) (isDeletable bool, err error) {
+	if autoDelete {
+		return true, nil
+	}
+
+	instanceName := orphan.Spec.Parameters[longhorn.OrphanInstanceName]
+	instanceProc, instanceExist := instanceMap[instanceName]
+	instanceCRScheduledBack, err := isInstanceScheduledOnNode(orphan.Spec.NodeID, instanceName)
+	if err != nil {
+		return false, errors.Wrapf(err, "failed to check CR scheduled node for %v orphan %v", orphan.Spec.Type, instanceName)
+	}
+
+	var instanceState longhorn.InstanceState
+	if instanceExist {
+		instanceState = instanceProc.Status.State
+	} else {
+		instanceState = longhorn.InstanceStateTerminated
+	}
+
+	imc.logger.WithFields(logrus.Fields{
+		"instanceExist":           instanceExist,
+		"instanceState":           instanceState,
+		"instanceCRScheduledBack": instanceCRScheduledBack,
+	}).Debugf("Checked deletable for %s orphan instance %s", orphan.Spec.Type, orphan.Name)
+	if !instanceExist || instanceCRScheduledBack {
+		return true, nil
+	}
+	return false, nil
+}
+
+func (imc *InstanceManagerController) isEngineScheduledOnCurrentNode(node string, instance string) (bool, error) {
+	existEngine, err := imc.ds.GetEngineRO(instance)
+	switch {
+	case err == nil:
+		// Engine CR still exists - check the ownership
+		return imc.isInstanceScheduledOnCurrentNode(node, &existEngine.Spec.InstanceSpec, &existEngine.Status.InstanceStatus), nil
+	case apierrors.IsNotFound(err):
+		// Engine CR not found - instance is orphaned
+		return false, nil
+	default:
+		// Unexpected error - unable to check if engine instance is orphaned or not
+		return false, err
+	}
+}
+
+func (imc *InstanceManagerController) isReplicaScheduledOnCurrentNode(node string, instance string) (bool, error) {
+	existReplica, err := imc.ds.GetReplicaRO(instance)
+	switch {
+	case err == nil:
+		// Replica CR still exists - check the ownership
+		return imc.isInstanceScheduledOnCurrentNode(node, &existReplica.Spec.InstanceSpec, &existReplica.Status.InstanceStatus), nil
+	case apierrors.IsNotFound(err):
+		// Replica CR not found - instance is orphaned
+		return false, nil
+	default:
+		// Unexpected error - unable to check if replica instance is orphaned or not
+		return false, err
+	}
+}
+
+// isInstanceScheduledOnCurrentNode returns true only when it is very certain that an instance is scheduled on this node
+func (imc *InstanceManagerController) isInstanceScheduledOnCurrentNode(node string, spec *longhorn.InstanceSpec, status *longhorn.InstanceStatus) bool {
+	if status.CurrentState != spec.DesireState {
+		return false
+	}
+	switch status.CurrentState {
+	case longhorn.InstanceStateRunning:
+		return status.OwnerID == spec.NodeID && spec.NodeID == node
+	case longhorn.InstanceStateStopped:
+		return spec.NodeID == node
+	default:
+		return false
+	}
+}
+
 func (imc *InstanceManagerController) startBackingImageMonitoring(im *longhorn.InstanceManager) {
 	log := imc.logger.WithField("instance manager", im.Name)
 
@@ -1942,8 +2109,13 @@ func (m *InstanceManagerMonitor) Run() {
 			if !needUpdate {
 				continue
 			}
-			if needStop := m.pollAndUpdateInstanceMap(); needStop {
+			instanceManager, instanceMap, needStop := m.pollInstanceMap()
+			if needStop {
 				return
+			}
+			if instanceManager != nil && instanceMap != nil {
+				m.syncInstances(instanceManager, instanceMap)
+				m.syncOrphans(instanceManager, instanceMap)
 			}
 		case <-m.stopCh:
 			return
@@ -1951,39 +2123,43 @@ func (m *InstanceManagerMonitor) Run() {
 	}
 }
 
-func (m *InstanceManagerMonitor) pollAndUpdateInstanceMap() (needStop bool) {
+func (m *InstanceManagerMonitor) pollInstanceMap() (im *longhorn.InstanceManager, instanceMap instanceProcessMap, needStop bool) {
 	im, err := m.ds.GetInstanceManager(m.Name)
 	if err != nil {
 		if datastore.ErrorIsNotFound(err) {
 			m.logger.Warn("Stop monitoring because the instance manager no longer exists")
-			return true
+			return nil, nil, true
 		}
 		utilruntime.HandleError(errors.Wrapf(err, "failed to get instance manager %v for monitoring", m.Name))
-		return false
+		return nil, nil, false
 	}
 
 	if im.Status.OwnerID != m.controllerID {
 		m.logger.Warnf("Stop monitoring the instance manager on this node (%v) because the instance manager has new ownerID %v", m.controllerID, im.Status.OwnerID)
-		return true
+		return nil, nil, true
 	}
 
 	resp, err := m.client.InstanceList()
 	if err != nil {
 		utilruntime.HandleError(errors.Wrapf(err, "failed to poll instance info to update instance manager %v", m.Name))
-		return false
+		return nil, nil, false
 	}
-	if !m.updateInstanceMap(im, resp) {
-		return false
+	return im, resp, false
+}
+
+func (m *InstanceManagerMonitor) syncInstances(im *longhorn.InstanceManager, instanceMap instanceProcessMap) {
+	if !m.updateInstanceMap(im, instanceMap) {
+		return
 	}
 	if _, err := m.ds.UpdateInstanceManagerStatus(im); err != nil {
 		utilruntime.HandleError(errors.Wrapf(err, "failed to update instance map for instance manager %v", m.Name))
-		return false
+		return
 	}
 
 	clusterAutoscalerEnabled, err := m.ds.GetSettingAsBool(types.SettingNameKubernetesClusterAutoscalerEnabled)
 	if err != nil {
 		utilruntime.HandleError(errors.Wrapf(err, "failed to get %v setting for instance manager %v", types.SettingNameKubernetesClusterAutoscalerEnabled, m.Name))
-		return false
+		return
 	}
 
 	// During volume attaching/detaching, it is likely both the engine and replica
@@ -1995,39 +2171,28 @@ func (m *InstanceManagerMonitor) pollAndUpdateInstanceMap() (needStop bool) {
 	if clusterAutoscalerEnabled && im.Spec.Type == longhorn.InstanceManagerTypeEngine {
 		m.nodeCallback(m.controllerID)
 	}
-
-	return false
 }
 
-func (m *InstanceManagerMonitor) updateInstanceMap(im *longhorn.InstanceManager, resp map[string]longhorn.InstanceProcess) bool {
+func (m *InstanceManagerMonitor) updateInstanceMap(im *longhorn.InstanceManager, instanceMap instanceProcessMap) bool {
 	switch {
 	case im.Status.APIVersion < 4:
-		if reflect.DeepEqual(im.Status.Instances, resp) { // nolint: staticcheck
+		if reflect.DeepEqual(im.Status.Instances, instanceMap) { // nolint: staticcheck
 			return false
 		}
 
-		im.Status.Instances = resp // nolint: staticcheck
+		im.Status.Instances = instanceMap // nolint: staticcheck
 	default:
-		engineProcess := map[string]longhorn.InstanceProcess{}
-		replicaProcess := map[string]longhorn.InstanceProcess{}
-		for name, process := range resp {
-			switch process.Status.Type {
-			case longhorn.InstanceTypeEngine:
-				engineProcess[name] = process
-			case longhorn.InstanceTypeReplica:
-				replicaProcess[name] = process
-			}
-		}
+		engineProcesses, replicaProcesses := m.categorizeProcesses(instanceMap)
 
 		// reflect.DeepEqual treats the two maps `var m1 map[string]process` and `m2 := map[string]process` as different maps.
 		// Therefore, to prevent unnecessary updates, we must check both that the length of the maps is zero and that the maps are identical.
-		if ((len(im.Status.InstanceEngines) == 0 && len(engineProcess) == 0) || reflect.DeepEqual(im.Status.InstanceEngines, engineProcess)) &&
-			((len(im.Status.InstanceReplicas) == 0 && len(replicaProcess) == 0) || reflect.DeepEqual(im.Status.InstanceReplicas, replicaProcess)) {
+		if ((len(im.Status.InstanceEngines) == 0 && len(engineProcesses) == 0) || reflect.DeepEqual(im.Status.InstanceEngines, engineProcesses)) &&
+			((len(im.Status.InstanceReplicas) == 0 && len(replicaProcesses) == 0) || reflect.DeepEqual(im.Status.InstanceReplicas, replicaProcesses)) {
 			return false
 		}
 
-		im.Status.InstanceEngines = engineProcess
-		im.Status.InstanceReplicas = replicaProcess
+		im.Status.InstanceEngines = engineProcesses
+		im.Status.InstanceReplicas = replicaProcesses
 	}
 	return true
 }
@@ -2042,6 +2207,127 @@ func (m *InstanceManagerMonitor) StopMonitorWithLock() {
 	m.lock.Lock()
 	defer m.lock.Unlock()
 	m.done = true
+}
+
+func (m *InstanceManagerMonitor) syncOrphans(im *longhorn.InstanceManager, instanceMap instanceProcessMap) {
+	engineProcesses, replicaProcesses := m.categorizeProcesses(instanceMap)
+	existOrphansList, err := m.ds.ListOrphansForEngineAndReplicaInstancesRO(im.Spec.NodeID)
+	if err != nil {
+		m.logger.WithError(err).Errorf("Failed to list orphans on node %s", im.Spec.NodeID)
+		return
+	}
+	existOrphans := make(map[string]bool)
+	for _, orphan := range existOrphansList {
+		existOrphans[orphan.Name] = true
+	}
+
+	// exam instances and create orphan CRs
+	m.createOrphanForInstances(existOrphans, im, engineProcesses, longhorn.OrphanTypeEngineInstance, m.isEngineOrphaned)
+	m.createOrphanForInstances(existOrphans, im, replicaProcesses, longhorn.OrphanTypeReplicaInstance, m.isReplicaOrphaned)
+}
+
+func (m *InstanceManagerMonitor) isEngineOrphaned(instanceName, node string) (bool, error) {
+	existEngine, err := m.ds.GetEngineRO(instanceName)
+	switch {
+	case err == nil:
+		// Engine CR still exists - check the ownership
+		return m.isInstanceOrphanedOnNode(&existEngine.Spec.InstanceSpec, &existEngine.Status.InstanceStatus, node), nil
+	case apierrors.IsNotFound(err):
+		// Engine CR not found - instance is orphaned
+		return true, nil
+	default:
+		// Unexpected error - unable to check if engine instance is orphaned or not
+		return false, err
+	}
+}
+
+func (m *InstanceManagerMonitor) isReplicaOrphaned(instanceName, node string) (bool, error) {
+	existReplica, err := m.ds.GetReplicaRO(instanceName)
+	switch {
+	case err == nil:
+		// Replica CR still exists - check the ownership
+		return m.isInstanceOrphanedOnNode(&existReplica.Spec.InstanceSpec, &existReplica.Status.InstanceStatus, node), nil
+	case apierrors.IsNotFound(err):
+		// Replica CR not found - instance is orphaned
+		return true, nil
+	default:
+		// Unexpected error - unable to check if replica instance is orphaned or not
+		return false, err
+	}
+}
+
+// isInstanceOrphanedOnNode returns true only when it is very certain that an instance is scheduled on another node
+func (m *InstanceManagerMonitor) isInstanceOrphanedOnNode(spec *longhorn.InstanceSpec, status *longhorn.InstanceStatus, node string) bool {
+	if status.CurrentState != spec.DesireState {
+		return false
+	}
+	switch status.CurrentState {
+	case longhorn.InstanceStateRunning:
+		return status.OwnerID == spec.NodeID && spec.NodeID != node
+	case longhorn.InstanceStateStopped:
+		return spec.NodeID != node
+	default:
+		return false
+	}
+}
+
+func (m *InstanceManagerMonitor) createOrphanForInstances(existOrphans map[string]bool, im *longhorn.InstanceManager, instanceMap instanceProcessMap, orphanType longhorn.OrphanType, orphanFilter func(instanceName, node string) (bool, error)) {
+	for instanceName, instance := range instanceMap {
+		orphanName := types.GetOrphanChecksumNameForOrphanedInstance(instanceName, m.controllerID, string(instance.Spec.DataEngine))
+		if _, isExist := existOrphans[orphanName]; isExist {
+			continue
+		}
+		if isOrphaned, err := orphanFilter(instanceName, im.Spec.NodeID); err != nil {
+			m.logger.WithError(err).Errorf("Failed to check %v orphan for instance %v", orphanType, instanceName)
+		} else if isOrphaned {
+			m.logger.Infof("Instance %v is orphaned", instanceName)
+			newOrphan, err := m.createOrphan(orphanName, im, instanceName, orphanType, instance.Spec.DataEngine)
+			if err != nil {
+				m.logger.WithError(err).Errorf("Failed to create %v orphan for instance %v", orphanType, instanceName)
+			} else if newOrphan != nil {
+				existOrphans[newOrphan.Name] = true
+			}
+		}
+	}
+}
+
+func (m *InstanceManagerMonitor) createOrphan(name string, im *longhorn.InstanceManager, instanceName string, orphanType longhorn.OrphanType, dataEngineType longhorn.DataEngineType) (*longhorn.Orphan, error) {
+	if _, err := m.ds.GetOrphanRO(name); err == nil || !apierrors.IsNotFound(err) {
+		return nil, err
+	}
+
+	// labels will be attached by mutator webhook
+	orphan := &longhorn.Orphan{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:            name,
+			OwnerReferences: datastore.GetOwnerReferencesForInstanceManager(im),
+		},
+		Spec: longhorn.OrphanSpec{
+			NodeID:     m.controllerID,
+			Type:       orphanType,
+			DataEngine: dataEngineType,
+			Parameters: map[string]string{
+				longhorn.OrphanInstanceName:    instanceName,
+				longhorn.OrphanInstanceManager: im.Name,
+			},
+		},
+	}
+
+	return m.ds.CreateOrphan(orphan)
+}
+
+func (m *InstanceManagerMonitor) categorizeProcesses(instanceMap instanceProcessMap) (instanceProcessMap, instanceProcessMap) {
+	engineProcesses := make(instanceProcessMap)
+	replicaProcesses := make(instanceProcessMap)
+	for name, process := range instanceMap {
+		switch process.Status.Type {
+		case longhorn.InstanceTypeEngine:
+			engineProcesses[name] = process
+		case longhorn.InstanceTypeReplica:
+			replicaProcesses[name] = process
+		}
+	}
+	return engineProcesses, replicaProcesses
 }
 
 func (imc *InstanceManagerController) isResponsibleFor(im *longhorn.InstanceManager) bool {
--- a/controller/node_controller.go
+++ b/controller/node_controller.go
@@ -194,7 +194,7 @@ func (nc *NodeController) isResponsibleForSetting(obj interface{}) bool {
 
 	return types.SettingName(setting.Name) == types.SettingNameStorageMinimalAvailablePercentage ||
 		types.SettingName(setting.Name) == types.SettingNameBackingImageCleanupWaitInterval ||
-		types.SettingName(setting.Name) == types.SettingNameOrphanAutoDeletion ||
+		types.SettingName(setting.Name) == types.SettingNameOrphanResourceAutoDeletion ||
 		types.SettingName(setting.Name) == types.SettingNameNodeDrainPolicy
 }
 
@@ -1328,7 +1328,9 @@ func (nc *NodeController) getNewAndMissingOrphanedReplicaDataStores(diskName, di
 	}
 	orphanMap := make(map[string]*longhorn.Orphan, len(orphanList))
 	for _, o := range orphanList {
-		orphanMap[o.Name] = o
+		if o.Spec.Type == longhorn.OrphanTypeReplicaData {
+			orphanMap[o.Name] = o
+		}
 	}
 
 	for dataStore := range replicaDataStores {
@@ -1355,10 +1357,11 @@ func (nc *NodeController) getNewAndMissingOrphanedReplicaDataStores(diskName, di
 }
 
 func (nc *NodeController) deleteOrphans(node *longhorn.Node, diskName string, diskInfo *monitor.CollectedDiskInfo, missingOrphanedReplicaDataStores map[string]string) error {
-	autoDeletionEnabled, err := nc.ds.GetSettingAsBool(types.SettingNameOrphanAutoDeletion)
+	autoDeletionResourceTypes, err := nc.ds.GetSettingOrphanResourceAutoDeletion()
 	if err != nil {
-		return errors.Wrapf(err, "failed to get %v setting", types.SettingNameOrphanAutoDeletion)
+		return errors.Wrapf(err, "failed to get %v setting", types.SettingNameOrphanResourceAutoDeletion)
 	}
+	autoDeletionEnabled := autoDeletionResourceTypes[types.OrphanResourceTypeReplicaData]
 
 	for dataStore := range missingOrphanedReplicaDataStores {
 		orphanName := types.GetOrphanChecksumNameForOrphanedDataStore(node.Name, diskName, diskInfo.Path, diskInfo.DiskUUID, dataStore)
@@ -1415,7 +1418,7 @@ func (nc *NodeController) createOrphan(node *longhorn.Node, diskName, replicaDat
 		},
 		Spec: longhorn.OrphanSpec{
 			NodeID: node.Name,
-			Type:   longhorn.OrphanTypeReplica,
+			Type:   longhorn.OrphanTypeReplicaData,
 			Parameters: map[string]string{
 				longhorn.OrphanDataName: replicaDataStore,
 				longhorn.OrphanDiskName: diskName,
--- a/controller/orphan_controller.go
+++ b/controller/orphan_controller.go
@@ -17,7 +17,6 @@ import (
 	"k8s.io/kubernetes/pkg/controller"
 
 	corev1 "k8s.io/api/core/v1"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	clientset "k8s.io/client-go/kubernetes"
 	v1core "k8s.io/client-go/kubernetes/typed/core/v1"
@@ -93,6 +92,15 @@ func NewOrphanController(
 	}
 	oc.cacheSyncs = append(oc.cacheSyncs, ds.NodeInformer.HasSynced)
 
+	if _, err = ds.InstanceManagerInformer.AddEventHandlerWithResyncPeriod(cache.ResourceEventHandlerFuncs{
+		AddFunc:    func(cur interface{}) { oc.enqueueForInstanceManager(cur) },
+		UpdateFunc: func(old, cur interface{}) { oc.enqueueForInstanceManager(cur) },
+		DeleteFunc: func(cur interface{}) { oc.enqueueForInstanceManager(cur) },
+	}, 0); err != nil {
+		return nil, err
+	}
+	oc.cacheSyncs = append(oc.cacheSyncs, ds.InstanceManagerInformer.HasSynced)
+
 	return oc, nil
 }
 
@@ -125,6 +133,33 @@ func (oc *OrphanController) enqueueForLonghornNode(obj interface{}) {
 	orphans, err := oc.ds.ListOrphansByNodeRO(node.Name)
 	if err != nil {
 		utilruntime.HandleError(fmt.Errorf("failed to list orphans on node %v since %v", node.Name, err))
+		return
+	}
+
+	for _, orphan := range orphans {
+		oc.enqueueOrphan(orphan)
+	}
+}
+
+func (oc *OrphanController) enqueueForInstanceManager(obj interface{}) {
+	im, ok := obj.(*longhorn.InstanceManager)
+	if !ok {
+		deletedState, ok := obj.(cache.DeletedFinalStateUnknown)
+		if !ok {
+			utilruntime.HandleError(fmt.Errorf("received unexpected obj: %#v", obj))
+			return
+		}
+		// use the last known state, to enqueue, dependent objects
+		im, ok = deletedState.Obj.(*longhorn.InstanceManager)
+		if !ok {
+			utilruntime.HandleError(fmt.Errorf("DeletedFinalStateUnknown contained invalid object: %#v", deletedState.Obj))
+			return
+		}
+	}
+
+	orphans, err := oc.ds.ListOrphansByNodeRO(im.Spec.NodeID)
+	if err != nil {
+		utilruntime.HandleError(fmt.Errorf("failed to list orphans on instance manager %v since %v", im.Name, err))
 		return
 	}
 
@@ -201,7 +236,7 @@ func (oc *OrphanController) syncOrphan(key string) (err error) {
 func (oc *OrphanController) reconcile(orphanName string) (err error) {
 	orphan, err := oc.ds.GetOrphan(orphanName)
 	if err != nil {
-		if !apierrors.IsNotFound(err) {
+		if !datastore.ErrorIsNotFound(err) {
 			return err
 		}
 		return nil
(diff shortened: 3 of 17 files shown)


k8s/pkg/apis/longhorn/v1beta2/orphan.go
```diff
@@ -5,12 +5,15 @@ import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
...
+	OrphanTypeEngineInstance   = OrphanType("engine-instance")
+	OrphanTypeReplicaInstance  = OrphanType("replica-instance")
 )
 
 const (
-	OrphanConditionTypeError         = "Error"
-	OrphanConditionTypeDataCleanable = "DataCleanable"
+	OrphanConditionTypeError                = "Error"
+	OrphanConditionTypeDataCleanable        = "DataCleanable"
+	OrphanConditionTypeRuntimeInstanceState = "InstanceState"
```
derekbit, line 16: I think we can remove `Runtime`. `Instance` should be enough and is already used in longhorn-manager and other componenets.

```suggestion
	OrphanConditionTypeInstanceState = "InstanceState"
```

k8s/pkg/apis/longhorn/v1beta2/orphan.go
```diff
@@ -20,6 +23,9 @@ const (
 )
 
 const (
+	OrphanInstanceName   = "RuntimeInstanceName"
```
derekbit, line 26:
```suggestion
	OrphanInstanceName   = "InstanceName"
```

controller/orphan_controller.go
```diff
@@ -269,48 +303,187 @@ func (oc *OrphanController) isResponsibleFor(orphan *longhorn.Orphan) bool {
...
+}
+
+func (oc *OrphanController) cleanupOrphanedEngineInstance(orphan *longhorn.Orphan) (isCleanupComplete bool, err error) {
+	defer func() {
+		if err != nil {
+			err = errors.Wrapf(err, "cannot cleanup engine instance orphan %v", orphan.Name)
+		}
+	}()
+
+	instance, deType, err := oc.extractOrphanedRuntimeInfo(orphan)
```
derekbit, line 358: Apply to other similar changes.
```suggestion
	instance, deType, err := oc.extractOrphanedInstanceInfo(orphan)
```

types/setting.go
```diff
@@ -1052,14 +1055,30 @@ var (
 	SettingDefinitionOrphanAutoDeletion = SettingDefinition{
 		DisplayName: "Orphan Auto-Deletion",
 		Description: "This setting allows Longhorn to delete the orphan resource and its corresponding orphaned data automatically. \n\n" +
-			"Orphan resources on down or unknown nodes will not be cleaned up automatically. \n\n",
+			"Orphan resources on down or unknown nodes will not be cleaned up automatically. \n\n" +
```
derekbit, line 1058: Mention the setting is deprecated and replaced by the new setting in the description.

cc @sushant-suse

controller/instance_manager_controller.go
```diff
@@ -612,6 +617,12 @@ func (imc *InstanceManagerController) handlePod(im *longhorn.InstanceManager) er
 		return err
 	}
 
+	if isPodDeletedOrNotRunning {
+		if err := imc.cleanupInstanceManagerOrphanRuntimeInstances(im); err != nil {
+			log.WithError(err).Warnf("Instance manager %v failed to cleanup runtime instance orphans for stopped instance manager pod", im.Name)
```
derekbit, line 622: nit:
```suggestion
			log.WithError(err).Warnf("Instance manager %v failed to cleanup orphaned instance for stopped instance manager pod", im.Name)
```

derekbit (commented):
I will continue the review later.

controller/orphan_controller.go
```diff
@@ -269,48 +303,187 @@ func (oc *OrphanController) isResponsibleFor(orphan *longhorn.Orphan) bool {
...
 	}()
 
 	// Make sure if the orphan nodeID and controller ID are same.
 	// If NO, just delete the orphan resource object and don't touch the data.
 	if orphan.Spec.NodeID != oc.controllerID {
-		log.Infof("Orphan nodeID %v is different from controllerID %v, so just delete the orphan resource object",
-			orphan.Name, oc.controllerID)
-		return nil
+		log.WithField("orphanType", orphan.Spec.Type).Infof("Orphan nodeID %v is on %v. Delete the orphan resource object.",
+			orphan.Name, orphan.Spec.NodeID)
```
c3y1huang, line 323:
```suggestion
	// Make sure if the orphan nodeID and controller ID are same.
	// If NO, just delete the orphan resource object and don't touch the data.
	if orphan.Spec.NodeID != oc.controllerID {
		log.WithFields(logrus.Fields{
			"orphanType": orphan.Spec.Type,
			"orphanName": orphan.Name,
			"orphanNode": orphan.Spec.NodeID,
		}).Infof("Orphan does not belong to this controller. Skipping resource cleanup")
```

controller/orphan_controller.go
```diff
@@ -269,48 +303,187 @@ func (oc *OrphanController) isResponsibleFor(orphan *longhorn.Orphan) bool {
...
+	dataEngineTypeStr, ok := orphan.Spec.Parameters[longhorn.OrphanDataEngineType]
+	if !ok {
+		return "", "", fmt.Errorf("failed to get data engine type for runtime instance orphan %v", orphan.Name)
+	}
+	switch dataEngineType = longhorn.DataEngineType(dataEngineTypeStr); dataEngineType {
+	case longhorn.DataEngineTypeV1, longhorn.DataEngineTypeV2:
+		// supported data engine type
 	default:
-		err = fmt.Errorf("unknown orphan type %v", orphan.Spec.Type)
+		return "", "", fmt.Errorf("unknown data engine type %v for runtime instance orphan %v", dataEngineTypeStr, orphan.Name)
```
c3y1huang, line 444: Is it possible to validate via webhook?

controller/orphan_controller.go
```diff
@@ -269,48 +303,187 @@ func (oc *OrphanController) isResponsibleFor(orphan *longhorn.Orphan) bool {
...
+		}
+	}()
+
+	_, err = imc.InstanceGet(engineType, instanceName, string(imType))
+	isCleanupComplete = false
+	if err != nil && types.ErrorIsNotFound(err) {
+		isCleanupComplete = true
+		err = nil
+	}
+	return isCleanupComplete, err
```
c3y1huang, line 483: NIT: consider tweaking this for better readabiliy and maintainablity.

```suggestion
	_, err = imc.InstanceGet(engineType, instanceName, string(imType))
	switch {
	case err == nil:
		return false, nil // instance still exists - cleanup not complete.
	case types.ErrorIsNotFound(err):
		return true, nil // instance not found - cleanup completed.
	default:
		return false, err // Unexpected error - cleanup status unknown.
	}
```

controller/orphan_controller.go
```diff
@@ -360,8 +535,13 @@ func (oc *OrphanController) DeleteV2ReplicaInstance(diskName, diskUUID, diskDriv
 }
 
 func (oc *OrphanController) updateConditions(orphan *longhorn.Orphan) error {
-	if err := oc.updateDataCleanableCondition(orphan); err != nil {
-		return err
+	switch orphan.Spec.Type {
```
c3y1huang, line 538: Consider handling the default case and logging unexpected type.

types/setting.go
```diff
@@ -1052,14 +1055,30 @@ var (
...
 	}
 
+	SettingDefinitionOrphanResourceAutoDeletion = SettingDefinition{
+		DisplayName: "Orphan Resource Auto-Deletion",
+		Description: "This setting allows Longhorn to delete the orphan resource and its corresponding orphaned resource automatically. \n\n" +
+			"Orphan resources on down or unknown nodes will not be cleaned up automatically. \n\n" +
+			"List the enabled resource types in a semicolon-separate list. Available items are: \n\n" +
+			"  - **replicaData**: replica data store \n\n" +
+			"  - **engineInstance**: engine runtime instance \n\n" +
+			"  - **replicaInstance**: replica runtime instance \n\n",
```
c3y1huang, line 1074: This is user-facing content. Consider aligning with other setting descriptions for consistency.
```suggestion
			"List the enabled resource types in a semicolon-separate list. \n\n" +
			"Available items are: \n\n" +
			"- **replicaData**: replica data store \n\n" +
			"- **engineInstance**: engine runtime instance \n\n" +
			"- **replicaInstance**: replica runtime instance \n\n",
```

controller/orphan_controller.go
COLDTURNIP: Good idea, will add this.

c3y1huang: Where is the validation of `dataEngineType` in webhook?

COLDTURNIP: Sorry for misleading. We have checked orphan type in webhook, but not the data engine type.

Though the type is checked in the webhook, printing error log for unknown type here would prevent developers forgot check this in the webhook in the future. At least we'll get errors, instead of leaving it silently.

COLDTURNIP: Just add checks for `dataEngineType`. Thanks for reminding.

types/setting.go
PhanLe1010: Is it just setting rename?

k8s/pkg/apis/longhorn/v1beta2/instancemanager.go
```diff
@@ -55,12 +55,13 @@ type InstanceProcessSpec struct {
...
-	InstanceStateStarting = InstanceState("starting")
-	InstanceStateStopping = InstanceState("stopping")
-	InstanceStateUnknown  = InstanceState("unknown")
+	InstanceStateRunning    = InstanceState("running")
+	InstanceStateStopped    = InstanceState("stopped")
+	InstanceStateError      = InstanceState("error")
+	InstanceStateStarting   = InstanceState("starting")
+	InstanceStateStopping   = InstanceState("stopping")
+	InstanceStateUnknown    = InstanceState("unknown")
+	InstanceStateTerminated = InstanceState("terminated")
```
PhanLe1010, line 64: Could you elaborate more about the use of the InstanceStateTerminated state?

types/setting.go
PhanLe1010: Oh. I see. New setting can control multiple orphan types

controller/setting_controller.go
```diff
@@ -896,6 +900,32 @@ func (sc *SettingController) syncDefaultLonghornStaticStorageClass() error {
...
+	deprecatedSetting, err := sc.ds.GetSetting(types.SettingNameOrphanAutoDeletion)
+	if err != nil {
+		return err
+	}
+	if replicaDataCleanupValue != deprecatedSetting.Value {
+		deprecatedSetting.Value = replicaDataCleanupValue
+		if _, err := sc.ds.UpdateSetting(deprecatedSetting); err != nil {
+			return err
+		}
+	}
```
PhanLe1010, line 924: Do we need to update the deprecatedSetting? Will the deprecated setting value be reset the next time user apply the Longhorn deployment yaml?

controller/node_controller.go
```diff
@@ -1402,7 +1405,7 @@ func (nc *NodeController) deleteOrphans(node *longhorn.Node, diskName string, di
 	return nil
 }
 
-func (nc *NodeController) createOrphans(node *longhorn.Node, diskName string, diskInfo *monitor.CollectedDiskInfo, newOrphanedReplicaDataStores map[string]string) error {
+func (nc *NodeController) createDataStoreOrphans(node *longhorn.Node, diskName string, diskInfo *monitor.CollectedDiskInfo, newOrphanedReplicaDataStores map[string]string) error {
```
PhanLe1010, line 1408: Can you elaborate more on why we rename to `createDataStoreOrphans`? What does DataStore mean in this context?

controller/instance_manager_controller.go
```diff
@@ -2044,6 +2087,284 @@ func (m *InstanceManagerMonitor) StopMonitorWithLock() {
...
+				longhorn.OrphanDataEngineType: string(dataEngineType),
+			},
+		},
+	}
+
+	_, err := m.ds.CreateOrphan(orphan)
+	return err
+}
+
+func (m *InstanceManagerMonitor) updateOrphanInstanceStatus(orphan *longhorn.Orphan, instanceState longhorn.InstanceState) error {
```
PhanLe1010, line 2325: Updating orphan CR status inside instance manager controller doesn't appropriate according to the general rule of k8s controller. AK.A only update the status of the CR inside its controller 

WDYT @derekbit

derekbit: Agree. We should prevent the anti-pattern behavior.

COLDTURNIP: Thanks, will orphan CR instance status update to orphan controller when it reconciling instance manager status update.

k8s/pkg/apis/longhorn/v1beta2/orphan.go
```diff
@@ -20,6 +23,9 @@ const (
 )
 
 const (
+	OrphanInstanceName   = "InstanceName"
+	OrphanDataEngineType = "DataEngineType"
```
derekbit, line 27: Could we add a field `DataEngine` in OrphanSpec instead?

controller/node_controller.go
COLDTURNIP: In this PR, besides the existing data store orphan, we introduce new types of orphans for runtime instance. The node controller observes the disk status change, and has no knowledge of engine/replica instances. Rename the method here to emphasize that the node controller will create only the data store orphans.

types/setting.go
```diff
@@ -1754,6 +1782,32 @@ func UnmarshalNodeSelector(nodeSelectorSetting string) (map[string]string, error
...
+			} else {
+				invalidItems = append(invalidItems, item)
+			}
+		}
+	}
+	if len(invalidItems) > 0 {
+		return nil, fmt.Errorf("invalid orphan resource types: %s", strings.Join(invalidItems, ", "))
+	}
+	return resourceTypes, nil
+}
```
derekbit, line 1809: nit: move to go-common-libs in the future.

controller/node_controller.go
```diff
@@ -1316,10 +1316,10 @@ func (nc *NodeController) syncOrphans(node *longhorn.Node, collectedDataInfo map
 		newOrphanedReplicaDataStores, missingOrphanedReplicaDataStores :=
 			nc.getNewAndMissingOrphanedReplicaDataStores(diskName, diskInfo.DiskUUID, diskInfo.Path, diskInfo.OrphanedReplicaDataStores)
 
-		if err := nc.createOrphans(node, diskName, diskInfo, newOrphanedReplicaDataStores); err != nil {
+		if err := nc.createDataStoreOrphans(node, diskName, diskInfo, newOrphanedReplicaDataStores); err != nil {
```
derekbit, line 1319: IMO for the convention, we should use the original naming `createOrphans`.

controller/node_controller.go
```diff
@@ -1316,10 +1316,10 @@ func (nc *NodeController) syncOrphans(node *longhorn.Node, collectedDataInfo map
...
 			return errors.Wrapf(err, "failed to create orphans for disk %v", diskName)
 		}
-		if err := nc.deleteOrphans(node, diskName, diskInfo, missingOrphanedReplicaDataStores); err != nil {
+		if err := nc.deleteDataStoreOrphans(node, diskName, diskInfo, missingOrphanedReplicaDataStores); err != nil {
```
derekbit, line 1322: Same as above

controller/node_controller.go
derekbit: I think we should use `createOrphans` and `createOrphan` that are similar to `createVolume` and so on. It is a convention and developers already know it is for resource creation.

controller/setting_controller.go
COLDTURNIP: 1. syncing the new config to deprecated setting would eliminate the user confusion about whether the data store auto cleanup is enabled or not.
2. the new setting will first be synced from old setting in LH manager's upgrade stage, then keep syncing the new setting back to the old one after the setting controller launched.

controller/orphan_controller.go
```diff
@@ -327,22 +508,24 @@ func (oc *OrphanController) deleteOrphanedReplica(orphan *longhorn.Orphan) error
 		err := lhns.DeletePath(filepath.Join(diskPath, "replicas", replicaDirectoryName))
 		return errors.Wrapf(err, "failed to delete orphan replica directory %v in disk %v", replicaDirectoryName, diskPath)
 	case longhorn.DiskTypeBlock:
-		return oc.DeleteV2ReplicaInstance(orphan.Spec.Parameters[longhorn.OrphanDiskName], orphan.Spec.Parameters[longhorn.OrphanDiskUUID], "", orphan.Spec.Parameters[longhorn.OrphanDataName])
+		return oc.DeleteV2ReplicaStore(orphan.Spec.Parameters[longhorn.OrphanDiskName], orphan.Spec.Parameters[longhorn.OrphanDiskUUID], "", orphan.Spec.Parameters[longhorn.OrphanDataName])
```
derekbit, line 511: The new name is a bit confusing. Could we still use `DeleteV2ReplicaInstance` because the instance means the SPDK lvol. `Store` sounds weird here.

controller/orphan_controller.go
```diff
@@ -269,48 +303,195 @@ func (oc *OrphanController) isResponsibleFor(orphan *longhorn.Orphan) bool {
 	return isControllerResponsibleFor(oc.controllerID, oc.ds, orphan.Name, orphan.Spec.NodeID, orphan.Status.OwnerID)
 }
 
-func (oc *OrphanController) cleanupOrphanedData(orphan *longhorn.Orphan) (err error) {
+func (oc *OrphanController) cleanupOrphanedResource(orphan *longhorn.Orphan) (isCleanupComplete bool, err error) {
```
derekbit, line 306: I'm fine with the change. 
Resource can mean replica directory, replica instance, engine instance in the data plane.

k8s/pkg/apis/longhorn/v1beta2/orphan.go
```diff
@@ -5,12 +5,15 @@ import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 type OrphanType string
 
 const (
-	OrphanTypeReplica = OrphanType("replica")
+	OrphanTypeReplicaDataStore = OrphanType("replica")
```
derekbit, line 8: If we are unable to find a better name, we can continue using `OrphanTypeReplicaData`.
`DataStore` is already used for the dataStore in the control plane. The naming is quite confusing. 
```suggestion
	OrphanTypeReplicaDataStore = OrphanType("replica")
```

k8s/pkg/apis/longhorn/v1beta2/instancemanager.go
COLDTURNIP: The `terminated` state represents that an instance was detected by the instance manager, but disappeared from the instance manager later. This is the trigger for orphan controller get notice that an orphan complete the cleanup. Will remove this because we're going to move the orphaned instance status update to orphan controller (see [below](https://github.com/longhorn/longhorn-manager/pull/3694#discussion_r2051284164))

k8s/pkg/apis/longhorn/v1beta2/orphan.go
COLDTURNIP: Add new field `DataEngine` in `OrphanSpec`, and the webhook validate when the orphan type is instance orphans.

k8s/pkg/apis/longhorn/v1beta2/instancemanager.go
PhanLe1010: I see. Thank you

controller/instance_manager_controller.go
```diff
@@ -2044,6 +2087,265 @@ func (m *InstanceManagerMonitor) StopMonitorWithLock() {
...
+		} else if isOrphaned {
+			m.logger.Infof("Replica instance %v is orphaned", replicaInstance.Spec.Name)
+			err := m.createOrphan(im, replicaInstance.Spec.Name, longhorn.OrphanTypeReplicaInstance, replicaInstance.Spec.DataEngine)
+			if err != nil {
+				m.logger.WithError(err).Errorf("Failed to create orphan for replica instance %v", replicaInstance.Spec.Name)
+			}
+		}
+	}
+
+	// update instance state on orphan CRs
```
PhanLe1010, line 2132: Looks like this commend is no longer needed since we don't update the orphan status inside IM controller now?

controller/instance_manager_controller.go
```diff
@@ -988,6 +999,15 @@ func (imc *InstanceManagerController) deleteInstanceManagerPDB(im *longhorn.Inst
 	return nil
 }
 
+func (imc *InstanceManagerController) syncOrphanedInstance(im *longhorn.InstanceManager) error {
+	if im.Spec.NodeID != imc.controllerID {
+		// The orphan owner node is no longer able to monitor instance status.
+		// Clear the orphan CRs, and they will be resync after the owner instance manager get back.
+		return imc.cleanupInstanceManagerOrphanInstances(im)
+	}
```
PhanLe1010, line 1007: Just a small note. We may need to be careful with this logic to avoid triggering actual data deletion

k8s/pkg/apis/longhorn/v1beta2/orphan.go
```diff
@@ -5,12 +5,15 @@ import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
...
-	OrphanTypeReplica = OrphanType("replica")
+	OrphanTypeReplicaData     = OrphanType("replica")
+	OrphanTypeEngineInstance  = OrphanType("engine-instance")
+	OrphanTypeReplicaInstance = OrphanType("replica-instance")
 )
 
 const (
 	OrphanConditionTypeError         = "Error"
 	OrphanConditionTypeDataCleanable = "DataCleanable"
+	OrphanConditionTypeInstanceState = "InstanceState"
```
PhanLe1010, line 16: Look like this condition represent InstanceStateTerminated. Should we change the name?

PhanLe1010 (commented):
In general, LGTM except for some NITs above

k8s/pkg/apis/longhorn/v1beta2/orphan.go
COLDTURNIP: Sure, will change to `InstanceExist`, positive if instance manager contains this instance, and record the instance state in reason.

types/types.go
```diff
@@ -726,6 +746,10 @@ func GetOrphanChecksumNameForOrphanedDataStore(nodeID, diskName, diskPath, diskU
 	return orphanPrefix + util.GetStringChecksumSHA256(strings.TrimSpace(fmt.Sprintf("%s-%s-%s-%s-%s", nodeID, diskName, diskPath, diskUUID, dataStore)))
 }
 
+func GetOrphanChecksumNameForOrphanedInstance(instanceName, nodeID, dataEngineType string) string {
```
derekbit, line 749: nit:
```suggestion
func GetOrphanChecksumNameForOrphanedInstance(instanceName, nodeID, dataEngine string) string {
```

types/setting.go
```diff
@@ -1055,14 +1058,31 @@ var (
...
 
+	SettingDefinitionOrphanResourceAutoDeletion = SettingDefinition{
+		DisplayName: "Orphan Resource Auto-Deletion",
+		Description: "This setting allows Longhorn to delete the orphan resource and its corresponding orphaned resource automatically. \n\n" +
+			"Orphan resources on down or unknown nodes will not be cleaned up automatically. \n\n" +
+			"List the enabled resource types in a semicolon-separate list. \n\n" +
+			"Available items are: \n\n" +
+			"- **replicaData**: replica data store \n\n" +
+			"- **engineInstance**: engine runtime instance \n\n" +
+			"- **replicaInstance**: replica runtime instance \n\n",
```
derekbit, line 1078: @sushant-suse Could you help review the description? Thanks.

types/setting.go
```diff
@@ -1771,6 +1799,32 @@ func UnmarshalNodeSelector(nodeSelectorSetting string) (map[string]string, error
 	return nodeSelector, nil
 }
 
+func UnmarshalOrphanResourceTypes(resourceTypesSetting string) (map[OrphanResourceType]bool, error) {
```
derekbit, line 1802: Add a unittest for the function.

controller/instance_manager_controller.go
```diff
@@ -1610,6 +1630,22 @@ func (imc *InstanceManagerController) createInstanceManagerPodSpec(im *longhorn.
...
+	orphans, err := imc.ds.ListOrphansByNodeRO(im.Spec.NodeID)
+	if err != nil {
+		return err
+	}
+	for _, orphan := range orphans {
+		switch orphan.Spec.Type {
+		case longhorn.OrphanTypeEngineInstance, longhorn.OrphanTypeReplicaInstance:
+			if err := imc.ds.DeleteOrphan(orphan.Name); err != nil && !apierrors.IsNotFound(err) {
+				return err
+			}
```
derekbit, line 1643: Use `multierr` to record the err and continue the next deletions.

controller/orphan_controller.go
```diff
@@ -227,13 +263,11 @@ func (oc *OrphanController) reconcile(orphanName string) (err error) {
...
 
 	if !orphan.DeletionTimestamp.IsZero() {
-		defer func() {
-			if err == nil {
-				err = oc.ds.RemoveFinalizerForOrphan(orphan)
-			}
-		}()
-
-		return oc.cleanupOrphanedData(orphan)
+		isCleanupComplete, err := oc.cleanupOrphanedResource(orphan)
```
derekbit, line 266: Is it possible `isCleanupComplete` is `false` but `err` is `nil`?

controller/orphan_controller.go
```diff
@@ -269,48 +303,200 @@ func (oc *OrphanController) isResponsibleFor(orphan *longhorn.Orphan) bool {
...
+			log.Infof("Only delete orphan %v resource object and do not delete the orphaned data store", orphan.Name)
+			return true, nil
+		}
+		err = oc.deleteOrphanedReplicaDataStore(orphan)
+		if err == nil || (err != nil && apierrors.IsNotFound(err)) {
+			isCleanupComplete = true
+			err = nil
+		}
+	default:
+		return false, fmt.Errorf("unknown orphan type %v to clear orphan %v", orphan.Spec.Type, orphan.Name)
```
derekbit, line 347:
```suggestion
		return false, fmt.Errorf("unknown orphan type %v to clean up orphaned resource for %v", orphan.Spec.Type, orphan.Name)
```

controller/orphan_controller.go
```diff
@@ -334,15 +520,17 @@ func (oc *OrphanController) deleteOrphanedReplica(orphan *longhorn.Orphan) error
 }
 
 func (oc *OrphanController) DeleteV2ReplicaInstance(diskName, diskUUID, diskDriver, replicaInstanceName string) (err error) {
-	logrus.Infof("Deleting SPDK replica instance %v on disk %v on node %v", replicaInstanceName, diskUUID, oc.controllerID)
+	logrus.Infof("Deleting SPDK replica store %v on disk %v on node %v", replicaInstanceName, diskUUID, oc.controllerID)
 
 	defer func() {
-		err = errors.Wrapf(err, "cannot delete v2 replica instance %v", replicaInstanceName)
+		if err != nil {
+			err = errors.Wrapf(err, "cannot delete v2 replica store %v", replicaInstanceName)
```
derekbit, line 527:
```suggestion
			err = errors.Wrapf(err, "cannot delete v2 replica instance %v", replicaInstanceName)
```

controller/orphan_controller.go
```diff
@@ -334,15 +520,17 @@ func (oc *OrphanController) deleteOrphanedReplica(orphan *longhorn.Orphan) error
 }
 
 func (oc *OrphanController) DeleteV2ReplicaInstance(diskName, diskUUID, diskDriver, replicaInstanceName string) (err error) {
-	logrus.Infof("Deleting SPDK replica instance %v on disk %v on node %v", replicaInstanceName, diskUUID, oc.controllerID)
+	logrus.Infof("Deleting SPDK replica store %v on disk %v on node %v", replicaInstanceName, diskUUID, oc.controllerID)
```
derekbit, line 523:
```suggestion
	logrus.Infof("Deleting SPDK replica instance %v on disk %v on node %v", replicaInstanceName, diskUUID, oc.controllerID)
```

controller/orphan_controller.go
```diff
@@ -334,15 +520,17 @@ func (oc *OrphanController) deleteOrphanedReplica(orphan *longhorn.Orphan) error
...
-		err = errors.Wrapf(err, "cannot delete v2 replica instance %v", replicaInstanceName)
+		if err != nil {
+			err = errors.Wrapf(err, "cannot delete v2 replica store %v", replicaInstanceName)
+		}
 	}()
 
 	im, err := oc.ds.GetRunningInstanceManagerByNodeRO(oc.controllerID, longhorn.DataEngineTypeV2)
 	if err != nil {
-		return errors.Wrapf(err, "failed to get running instance manager for node %v for deleting v2 replica instance %v", oc.controllerID, replicaInstanceName)
+		return errors.Wrapf(err, "failed to get running instance manager for node %v for deleting v2 replica store %v", oc.controllerID, replicaInstanceName)
```
derekbit, line 533:
```suggestion
		return errors.Wrapf(err, "failed to get running instance manager for node %v for deleting v2 replica instance %v", oc.controllerID, replicaInstanceName)
```

controller/orphan_controller.go
```diff
@@ -371,6 +572,55 @@ func (oc *OrphanController) updateConditions(orphan *longhorn.Orphan) error {
...
+
+	instanceProc, err := imc.InstanceGet(orphan.Spec.DataEngine, instanceName, string(instanceType))
+	switch {
+	case err != nil:
+		return errors.Wrapf(err, "failed to get instance %v", instanceName)
+	case instanceProc == nil:
+		instanceState = longhorn.InstanceStateTerminated
+	default:
+		instanceState = instanceProc.Status.State
+	}
```
derekbit, line 619: What's the purpose of reading the instance state?

controller/instance_manager_controller.go
```diff
@@ -612,6 +617,12 @@ func (imc *InstanceManagerController) handlePod(im *longhorn.InstanceManager) er
 		return err
 	}
 
+	if isPodDeletedOrNotRunning {
+		if err := imc.cleanupInstanceManagerOrphanInstances(im); err != nil {
```
derekbit, line 621: The function `cleanupInstanceManagerOrphanInstances` actually deletes Orphan resources rather than instances. Suggest renaming to `deleteOrphans(im)`.

```suggestion
		if err := imc.deleteOrphans(im); err != nil {
```

controller/instance_manager_controller.go
```diff
@@ -612,6 +617,12 @@ func (imc *InstanceManagerController) handlePod(im *longhorn.InstanceManager) er
...
+			log.WithError(err).Warnf("Instance manager %v failed to cleanup orphan instance for stopped instance manager pod", im.Name)
```
derekbit, line 622: Update the log message.

controller/instance_manager_controller.go
```diff
@@ -2044,6 +2087,265 @@ func (m *InstanceManagerMonitor) StopMonitorWithLock() {
...
+	engineProcess := map[string]longhorn.InstanceProcess{}
+	replicaProcess := map[string]longhorn.InstanceProcess{}
+	for name, process := range instanceProcess {
+		switch process.Status.Type {
+		case longhorn.InstanceTypeEngine:
+			engineProcess[name] = process
+		case longhorn.InstanceTypeReplica:
+			replicaProcess[name] = process
+		}
+	}
```
derekbit, line 2100: nit: wrap the codes into a helper function like `categorizeProcesses()`

controller/instance_manager_controller.go
```diff
@@ -2044,6 +2087,265 @@ func (m *InstanceManagerMonitor) StopMonitorWithLock() {
...
+	if err != nil {
+		m.logger.WithError(err).Warnf("Failed to fetch orphan auto deletion setting, disable by default")
+	} else {
+		autoDeleteEngine = autoDeletionTypes[types.OrphanResourceTypeEngineInstance]
+		autoDeleteReplica = autoDeletionTypes[types.OrphanResourceTypeReplicaInstance]
+	}
+	m.cleanupOrphans(existOrphans, deleteOrphans, autoDeleteEngine, autoDeleteReplica)
+}
+
+func (m *InstanceManagerMonitor) getNodeRuntimeOrphanMap() (map[string]*longhorn.Orphan, error) {
```
derekbit, line 2190: It is not easy to quickly understand the purpose of the function from the function name.
Suggest `listOrphansForEngineAndReplicaInstances(nodeName string)` or similar one.

controller/instance_manager_controller.go
```diff
@@ -2044,6 +2087,265 @@ func (m *InstanceManagerMonitor) StopMonitorWithLock() {
...
+		if isOrphaned, err := m.isReplicaOrphaned(replicaInstance.Spec.Name); err != nil {
+			m.logger.WithError(err).Errorf("Failed to check orphan for replica instance %v", replicaInstance.Spec.Name)
+		} else if isOrphaned {
+			m.logger.Infof("Replica instance %v is orphaned", replicaInstance.Spec.Name)
+			err := m.createOrphan(im, replicaInstance.Spec.Name, longhorn.OrphanTypeReplicaInstance, replicaInstance.Spec.DataEngine)
+			if err != nil {
+				m.logger.WithError(err).Errorf("Failed to create orphan for replica instance %v", replicaInstance.Spec.Name)
+			}
+		}
+	}
```
derekbit, line 2130: Wrap the codes into a helper function like `createOrphansForInstances`

controller/instance_manager_controller.go
```diff
@@ -2044,6 +2087,265 @@ func (m *InstanceManagerMonitor) StopMonitorWithLock() {
...
+
+	// update instance state on orphan CRs
+	deleteOrphans := make(map[string]bool)
```
derekbit, line 2133:
```suggestion
	deletableOrphans := make(map[string]bool)
```

controller/instance_manager_controller.go
```diff
@@ -2044,6 +2087,265 @@ func (m *InstanceManagerMonitor) StopMonitorWithLock() {
...
+			if !instanceExist || instanceCRScheduledBack {
+				m.logger.WithFields(logrus.Fields{
+					"instanceExist":           instanceExist,
+					"instanceState":           instanceState,
+					"instanceCRScheduledBack": instanceCRScheduledBack,
+				}).Debugf("Orphan instance %s is deletable", orphanName)
+				deleteOrphans[orphanName] = true
+			}
+		}
+	}
```
derekbit, line 2175: Wrap the codes into a helper function

controller/instance_manager_controller.go
```diff
@@ -2044,6 +2087,265 @@ func (m *InstanceManagerMonitor) StopMonitorWithLock() {
...
+
+	// delete orphan CRs
+	autoDeletionTypes, err := m.ds.GetSettingOrphanResourceAutoDeletion()
+	var autoDeleteEngine = false
+	var autoDeleteReplica = false
+	if err != nil {
+		m.logger.WithError(err).Warnf("Failed to fetch orphan auto deletion setting, disable by default")
```
derekbit, line 2182: should we directly return if run into an error?

controller/instance_manager_controller.go
```diff
@@ -2044,6 +2087,265 @@ func (m *InstanceManagerMonitor) StopMonitorWithLock() {
...
+			}
+		case longhorn.OrphanTypeReplicaInstance:
+			if err := m.cleanupOrphan(orphanName, deleteOrphans, autoDeleteReplica); err != nil {
+				m.logger.WithError(err).WithField("instance", instanceName).Errorf("Failed to cleanup replica orphan %s", orphanName)
+			}
+		}
+	}
+}
+
+func (m *InstanceManagerMonitor) cleanupOrphan(orphanName string, deleteOrphans map[string]bool, autoDelete bool) error {
```
derekbit, line 2339: nit: why not `deleteOrphan`?
```suggestion
func (m *InstanceManagerMonitor) deleteOrphan(orphanName string, deleteOrphans map[string]bool, autoDelete bool) error {
```

controller/instance_manager_controller.go
```diff
@@ -2044,6 +2087,265 @@ func (m *InstanceManagerMonitor) StopMonitorWithLock() {
...
+				longhorn.OrphanInstanceManager: im.Name,
+			},
+		},
+	}
+
+	_, err := m.ds.CreateOrphan(orphan)
+	return err
+}
+
+func (m *InstanceManagerMonitor) cleanupOrphans(existOrphans map[string]*longhorn.Orphan, deleteOrphans map[string]bool, autoDeleteEngine, autoDeleteReplica bool) {
```
derekbit, line 2323: nit: why not `deleteOrphans`?
```suggestion
func (m *InstanceManagerMonitor) deleteOrphans(existOrphans map[string]*longhorn.Orphan, deleteOrphans map[string]bool, autoDeleteEngine, autoDeleteReplica bool) {
```

controller/instance_manager_controller.go
```diff
@@ -2044,6 +2087,265 @@ func (m *InstanceManagerMonitor) StopMonitorWithLock() {
...
+	}
+}
+
+func (m *InstanceManagerMonitor) cleanupOrphan(orphanName string, deleteOrphans map[string]bool, autoDelete bool) error {
+	if autoDelete {
+		return m.ds.DeleteOrphan(orphanName)
+	}
+	if toDelete := deleteOrphans[orphanName]; toDelete {
+		return m.ds.DeleteOrphan(orphanName)
+	}
```
derekbit, line 2345: How about simplifying the codes

```
if autoDelete || deleteOrphans[orphanName] {
    return m.ds.DeleteOrphan(orphanName)
}
```

derekbit (commented):
In general, LGTM. Suggest improving the functions and variables naming and breaking down the longer functions into smaller ones for better readability.

upgrade/v18xto190/upgrade.go
```diff
@@ -85,3 +101,49 @@ func upgradeVolumes(namespace string, lhClient *lhclientset.Clientset, resourceM
...
+		err = errors.Wrapf(err, upgradeLogPrefix+"upgrade settings failed")
+	}()
+
+	settingsMap, err := upgradeutil.ListAndUpdateSettingsInProvidedCache(namespace, lhClient, resourceMaps)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil
+		}
+		return errors.Wrapf(err, "failed to list all existing Longhorn settings during the settings upgrade")
+	}
```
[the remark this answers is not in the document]
COLDTURNIP: Will remove it.

controller/orphan_controller.go
COLDTURNIP: Instance manager may take time to wait for instance terminating. Orphan controller will confirm the instance state with instance manager after sending the deletion request. `isCleanupComplete` will be false if the instance is still exist (terminating or some other state), and the deletion will continue after the instance state changes, triggered by instance manager status update event.

controller/orphan_controller.go
COLDTURNIP: To track the state of corresponding instance, and make trouble shooting easier. User don't need to compare the orphan object with the instance list in im CR.

upgrade/v18xto190/upgrade.go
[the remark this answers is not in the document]
COLDTURNIP: We have `upgradeutil.ListAndUpdateVolumesInProvidedCache` in `upgradeVolumes`, and actually `upgradeutil.ListAndUpdate*InProvidedCache` returns the existing loaded CRs if presents in cache. Should we avoid invoking these loader multiple times?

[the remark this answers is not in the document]
derekbit: Okay. Got your point. Then, this should be fine.

upgrade/v18xto190/upgrade.go
```diff
@@ -32,6 +36,18 @@ func UpgradeResources(namespace string, lhClient *lhclientset.Clientset, kubeCli
 		return errors.New("resourceMaps cannot be nil")
 	}
 
+	if err := updateCRs(namespace, lhClient, kubeClient, resourceMaps); err != nil {
+		return err
+	}
+
+	if err := updateSettings(namespace, lhClient, kubeClient, resourceMaps); err != nil {
```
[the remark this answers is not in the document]
COLDTURNIP: Changed to `upgradeSettings` as a part of `updateCRs`

controller/instance_manager_controller.go
COLDTURNIP: Refined. Please help review this part again, thanks.

controller/instance_manager_controller.go
COLDTURNIP: Extracted to `createOrphanForInstancesAndUpdateCache`

types/setting.go
COLDTURNIP: Will cover this.

upgrade/v18xto190/upgrade.go
```diff
@@ -85,3 +101,49 @@ func upgradeVolumes(namespace string, lhClient *lhclientset.Clientset, resourceM
...
+	resourceTypes[types.OrphanResourceTypeReplicaData] = oldSetting.Value == "true"
+	enabledResourceType := make([]string, 0, len(resourceTypes))
+	for rt, enabled := range resourceTypes {
+		if enabled {
+			enabledResourceType = append(enabledResourceType, string(rt))
+		}
+	}
+	newSetting.Value = strings.Join(enabledResourceType, ";")
+	return nil
+}
```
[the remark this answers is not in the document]
COLDTURNIP: Will improve the test coverage.

k8s/pkg/apis/longhorn/v1beta2/orphan.go
```diff
@@ -36,6 +42,10 @@ type OrphanSpec struct {
 	// Can be "replica".
 	// +optional
 	Type OrphanType `json:"orphanType"`
+	// The type of data engine for instance orphan.
+	// Can be "v1", "v2".
+	// +optional
```
derekbit, line 47:
```suggestion
	// +kubebuilder:validation:Enum=v1;v2
```

controller/orphan_controller.go
```diff
@@ -269,48 +303,200 @@ func (oc *OrphanController) isResponsibleFor(orphan *longhorn.Orphan) bool {
...
+
+	im, err := oc.ds.GetRunningInstanceManagerByNodeRO(oc.controllerID, deType)
+	if err != nil {
+		oc.logger.WithError(err).Infof("No running instance manager for node %v for deleting orphan instance %v", oc.controllerID, orphan.Name)
+		return true, nil
+	}
+	if im.Name != imName {
+		oc.logger.WithField("instanceManager", imName).Infof("Orphan instance %v is not managed by current running instance manager %v", orphan.Name, im.Name)
+		return true, nil
+	}
```
derekbit, line 416: nit: wrap the codes into a helper function like `isCleanupOrphanedInstanceRequired`?

controller/orphan_controller.go
```diff
@@ -269,48 +303,200 @@ func (oc *OrphanController) isResponsibleFor(orphan *longhorn.Orphan) bool {
...
+	var spec *longhorn.InstanceSpec
+	if replicaCR, err := oc.ds.GetReplicaRO(instance); err != nil {
+		if !apierrors.IsNotFound(err) {
+			return false, err
+		}
+		spec = nil
+	} else {
+		spec = &replicaCR.Spec.InstanceSpec
+	}
+	return oc.cleanupOrphanedInstance(orphan, instance, imName, longhorn.InstanceManagerTypeReplica, orphan.Spec.DataEngine, spec)
```
derekbit, line 398: `orphan` is one of the parameters. Should we need `orphan.Spec.DataEngine` here?

controller/orphan_controller.go
```diff
@@ -269,48 +303,200 @@ func (oc *OrphanController) isResponsibleFor(orphan *longhorn.Orphan) bool {
...
+	var spec *longhorn.InstanceSpec
+	if engineCR, err := oc.ds.GetEngineRO(instance); err != nil {
+		if !apierrors.IsNotFound(err) {
+			return false, err
+		}
+		spec = nil
+	} else {
+		spec = &engineCR.Spec.InstanceSpec
+	}
+	return oc.cleanupOrphanedInstance(orphan, instance, imName, longhorn.InstanceManagerTypeEngine, orphan.Spec.DataEngine, spec)
```
derekbit, line 374: `orphan` is one of the parameters. Should we need `orphan.Spec.DataEngine` here?

controller/orphan_controller.go
```diff
@@ -337,7 +523,9 @@ func (oc *OrphanController) DeleteV2ReplicaInstance(diskName, diskUUID, diskDriv
 	logrus.Infof("Deleting SPDK replica instance %v on disk %v on node %v", replicaInstanceName, diskUUID, oc.controllerID)
 
 	defer func() {
-		err = errors.Wrapf(err, "cannot delete v2 replica instance %v", replicaInstanceName)
+		if err != nil {
+			err = errors.Wrapf(err, "cannot delete v2 replica instance %v", replicaInstanceName)
+		}
```
derekbit, line 528: `if err != nil {` is not required. If `err` is nil, the result is nil as well.

controller/orphan_controller.go
```diff
@@ -269,48 +303,200 @@ func (oc *OrphanController) isResponsibleFor(orphan *longhorn.Orphan) bool {
...
+
+	instance, imName, err := oc.extractOrphanedInstanceInfo(orphan)
+	if err != nil {
+		return false, err
+	}
+
+	var spec *longhorn.InstanceSpec
+	if engineCR, err := oc.ds.GetEngineRO(instance); err != nil {
+		if !apierrors.IsNotFound(err) {
+			return false, err
```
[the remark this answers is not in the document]
COLDTURNIP: Yes. When the engine CR is absent later in `cleanupOrphanedInstance`, it skip the CR node location examination and continue on instance deletion.

controller/orphan_controller.go
COLDTURNIP: im is needed to create instance manager client later. Will extract to something like `getRunningInstanceManagerClientForOrphan`, and skip deletion if there's no available client without error.

derekbit: Okay. Not easy to abstract this part. Let's keep it as it is. Thank you.

controller/instance_manager_controller.go
```diff
@@ -2044,6 +2084,268 @@ func (m *InstanceManagerMonitor) StopMonitorWithLock() {
...
+	existOrphans, err := m.listOrphansForEngineAndReplicaInstances(im.Spec.NodeID)
+	if err != nil {
+		m.logger.WithError(err).Errorf("Failed to list orphans on node %s", im.Spec.NodeID)
+		return
+	}
+
+	// exam instances and create orphan CRs
+	existOrphans = m.createOrphanForInstancesAndUpdateCache(existOrphans, im, engineProcess, longhorn.OrphanTypeEngineInstance, m.isEngineOrphaned)
+	existOrphans = m.createOrphanForInstancesAndUpdateCache(existOrphans, im, replicaProcess, longhorn.OrphanTypeReplicaInstance, m.isReplicaOrphaned)
+
```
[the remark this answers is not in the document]
COLDTURNIP: Refactor `getDeletableOrphans` to `isOrphanDeletable` to simplify `deleteOrphans`

COLDTURNIP:
After discussion, will move the orphan deletion from im monitor to im controller, to reflect the settings change.

controller/instance_manager_controller.go
```diff
@@ -988,6 +1009,16 @@ func (imc *InstanceManagerController) deleteInstanceManagerPDB(im *longhorn.Inst
 	return nil
 }
 
+func (imc *InstanceManagerController) syncOrphan(im *longhorn.InstanceManager) error {
```
derekbit, line 1012:
```suggestion
func (imc *InstanceManagerController) syncOrphans(im *longhorn.InstanceManager) error {
```

types/setting.go
```diff
@@ -1055,14 +1058,31 @@ var (
...
 		Type:     SettingTypeBool,
 		Required: true,
-		ReadOnly: false,
+		ReadOnly: true,
 		Default:  "false",
 	}
 
+	SettingDefinitionOrphanResourceAutoDeletion = SettingDefinition{
+		DisplayName: "Orphan Resource Auto-Deletion",
+		Description: "This setting allows Longhorn to delete the orphan resource and its corresponding orphaned resource automatically. \n\n" +
```
sushant-suse, line 1072:
```suggestion
		Description: "This setting allows Longhorn to automatically delete orphan resources and their corresponding orphaned resources. \n\n" +
```

types/setting.go
```diff
@@ -1055,14 +1058,31 @@ var (
...
+			"Orphan resources on down or unknown nodes will not be cleaned up automatically. \n\n" +
+			"List the enabled resource types in a semicolon-separate list. \n\n" +
```
sushant-suse, line 1074:
```suggestion
			"List the enabled resource types in a semicolon-separated list. \n\n" +
```

sushant-suse (changes requested):
Thanks.

sushant-suse (approved):
LGTM.
Thanks.

controller/orphan_controller.go
```diff
@@ -269,48 +303,205 @@ func (oc *OrphanController) isResponsibleFor(orphan *longhorn.Orphan) bool {
...
+}
+
+func (oc *OrphanController) deleteInstance(imc *engineapi.InstanceManagerClient, instanceName string, instanceKind longhorn.InstanceManagerType, engineType longhorn.DataEngineType) (err error) {
+	// There is a delay between deletion initiation and state/InstanceManager update,
+	// this function may be called multiple times before given instance exits.
+
+	oc.logger.Infof("Orphan controller deleting instance %v", instanceName)
+
+	defer func() {
+		err = errors.Wrapf(err, "cannot delete instance %v", instanceName)
```
derekbit, line 471:
```suggestion
		err = errors.Wrapf(err, "failed to delete instance %v", instanceName)
```

controller/orphan_controller.go
```diff
@@ -371,6 +575,55 @@ func (oc *OrphanController) updateConditions(orphan *longhorn.Orphan) error {
...
+			status = longhorn.ConditionStatusFalse
+		}
+
+		orphan.Status.Conditions = types.SetCondition(orphan.Status.Conditions, longhorn.OrphanConditionTypeInstanceExist, status, string(instanceState), "")
+	}()
+
+	instanceName, instanceManager, err := oc.extractOrphanedInstanceInfo(orphan)
+	im, err := oc.ds.GetInstanceManager(instanceManager)
+	if err != nil {
+		if !apierrors.IsNotFound(err) {
```
derekbit, line 597: Shoud it be `if apierrors.IsNotFound(err) {`?

```suggestion
		if apierrors.IsNotFound(err) {
```

controller/orphan_controller.go
```diff
@@ -371,6 +575,55 @@ func (oc *OrphanController) updateConditions(orphan *longhorn.Orphan) error {
...
+	if err != nil {
+		return errors.Wrapf(err, "failed to check node down or missing manager for node %v", orphan.Spec.NodeID)
+	}
+	defer func() {
+		if closeErr := imc.Close(); closeErr != nil {
+			oc.logger.WithError(closeErr).Error("Failed to close instance manager client")
+		}
+	}()
+
+	instanceProc, err := imc.InstanceGet(orphan.Spec.DataEngine, instanceName, string(instanceType))
```
derekbit, line 614:
```suggestion
	instance, err := imc.InstanceGet(orphan.Spec.DataEngine, instanceName, string(instanceType))
```

derekbit:
@coderabbitai review

controller/orphan_controller.go
COLDTURNIP: Sure, it should be `apierrors.IsNotFound(err)`

controller/orphan_controller.go
```diff
@@ -269,48 +303,205 @@ func (oc *OrphanController) isResponsibleFor(orphan *longhorn.Orphan) bool {
...
-	switch orphan.Spec.Type {
-	case longhorn.OrphanTypeReplica:
-		err = oc.deleteOrphanedReplica(orphan)
+	return isCleanupComplete, err
+}
+
+func (oc *OrphanController) cleanupOrphanedEngineInstance(orphan *longhorn.Orphan) (isCleanupComplete bool, err error) {
+	defer func() {
+		if err != nil {
+			err = errors.Wrapf(err, "cannot cleanup orphan engine instance %v", orphan.Name)
```
derekbit, line 356:
```suggestion
			err = errors.Wrapf(err, "failed to cleanup orphan engine instance %v", orphan.Name)
```

controller/orphan_controller.go
```diff
@@ -269,48 +303,205 @@ func (oc *OrphanController) isResponsibleFor(orphan *longhorn.Orphan) bool {
...
+	} else {
+		spec = &engineCR.Spec.InstanceSpec
+	}
+	return oc.cleanupOrphanedInstance(orphan, instance, imName, longhorn.InstanceManagerTypeEngine, spec)
+}
+
+func (oc *OrphanController) cleanupOrphanedReplicaInstance(orphan *longhorn.Orphan) (isCleanupComplete bool, err error) {
+	defer func() {
+		if err != nil {
+			err = errors.Wrapf(err, "cannot cleanup orphan replica instance %v", orphan.Name)
```
derekbit, line 380:
```suggestion
			err = errors.Wrapf(err, "failed to cleanup orphan replica instance %v", orphan.Name)
```

controller/orphan_controller.go
```diff
@@ -269,48 +303,205 @@ func (oc *OrphanController) isResponsibleFor(orphan *longhorn.Orphan) bool {
...
+func (oc *OrphanController) cleanupOrphanedInstance(orphan *longhorn.Orphan, instance, imName string, imType longhorn.InstanceManagerType, instanceCRSpec *longhorn.InstanceSpec) (isCleanupComplete bool, err error) {
+	if instanceCRSpec != nil && instanceCRSpec.NodeID == orphan.Spec.NodeID {
+		oc.logger.Infof("Orphan instance %v is scheduled back to current node %v. Skip cleaning up the instance resource and finalize the orphan CR.", instance, orphan.Spec.NodeID)
+		return true, nil
 	}
 
-	return err
+	imc, err := oc.getRunningInstanceManagerClientForOrphan(orphan, imName)
+	if err != nil {
+		return false, errors.Wrapf(err, "cannot get running instance manager client for orphan %v", orphan.Name)
```
derekbit, line 430:
```suggestion
		return false, errors.Wrapf(err, "failed to get running instance manager client for orphan %v", orphan.Name)
```

controller/orphan_controller.go
```diff
@@ -269,48 +303,205 @@ func (oc *OrphanController) isResponsibleFor(orphan *longhorn.Orphan) bool {
...
+	}
+	return isCleanupComplete, cleanupErr
 }
 
-func (oc *OrphanController) deleteOrphanedReplica(orphan *longhorn.Orphan) error {
+func (oc *OrphanController) getRunningInstanceManagerClientForOrphan(orphan *longhorn.Orphan, imName string) (*engineapi.InstanceManagerClient, error) {
+	im, err := oc.ds.GetRunningInstanceManagerByNodeRO(orphan.Spec.NodeID, orphan.Spec.DataEngine)
+	if err != nil {
+		oc.logger.WithError(err).Warnf("No running instance manager for node %v for deleting orphan instance %v", oc.controllerID, orphan.Name)
+		return nil, nil
```
derekbit, line 455: Do we need to handle different errors? For example, returns  `nil, nil`  only when the error is `not found`.

datastore/longhorn.go
```diff
@@ -3523,6 +3523,18 @@ func (s *DataStore) GetSettingSystemManagedComponentsNodeSelector() (map[string]
 	return nodeSelector, nil
 }
 
+func (s *DataStore) GetSettingOrphanResourceAutoDeletion() (map[types.OrphanResourceType]bool, error) {
```
derekbit, line 3526: Add a description for the new function.

k8s/pkg/apis/longhorn/v1beta2/instancemanager.go
derekbit, line 64: Add a comment here.  IIUC, `terminated` means the instance is deleted and not found.

```suggestion
	InstanceStateTerminated = InstanceState("terminated")
```

controller/instance_manager_controller.go
```diff
@@ -988,6 +1009,16 @@ func (imc *InstanceManagerController) deleteInstanceManagerPDB(im *longhorn.Inst
 	return nil
 }
 
+func (imc *InstanceManagerController) syncOrphans(im *longhorn.InstanceManager) error {
+	var isInstanceManagerTerminating = false
```
derekbit, line 1013: More clear. WDYT?
```suggestion
	// Add a comment explaining the ownership change might be due to blabla...
	var instanceManagerOwnershipChanged = false
```

controller/instance_manager_controller.go
```diff
@@ -170,6 +172,13 @@
 	}
 	imc.cacheSyncs = append(imc.cacheSyncs, ds.KubeNodeInformer.HasSynced)
 
+	if _, err = ds.OrphanInformer.AddEventHandlerWithResyncPeriod(cache.ResourceEventHandlerFuncs{
+		AddFunc: imc.enqueueInstanceManagerOrphan,
+	}, 0); err != nil {
+		return nil, err
+	}
+	imc.cacheSyncs = append(imc.cacheSyncs, ds.OrphanInformer.HasSynced)
+
```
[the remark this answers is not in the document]
COLDTURNIP: Instance manager handles only the orphan CR creation since it is responsible to trigger orphan CR deletion. The orphan CR status update does not effects instance manager.

controller/instance_manager_controller.go
```diff
@@ -1610,6 +1650,123 @@
...
+	orphanList, err := imc.ds.ListOrphansForEngineAndReplicaInstancesRO(im.Spec.NodeID)
+	if err != nil {
+		return err
+	}
+	multiError := util.NewMultiError()
+	for _, orphan := range orphanList {
+		if !orphan.DeletionTimestamp.IsZero() {
+			continue
+		}
```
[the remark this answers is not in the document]
COLDTURNIP: `DeletionTimestamp.IsZero()` handles `nil` internally.

upgrade/v18xto190/upgrade.go
```diff
@@ -85,3 +101,53 @@ func upgradeVolumes(namespace string, lhClient *lhclientset.Clientset, resourceM
...
+	}
+
+	resourceTypes[types.OrphanResourceTypeReplicaData] = oldSetting.Value == "true"
+	enabledResourceType := make([]string, 0, len(resourceTypes))
+	for rt, enabled := range resourceTypes {
+		if enabled {
+			enabledResourceType = append(enabledResourceType, string(rt))
+		}
+	}
+	newSetting.Value = strings.Join(enabledResourceType, ";")
```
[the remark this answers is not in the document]
COLDTURNIP: For the new setting, the value would be blank string, and the unmarshaled result would be off for all resource types:

```
resourceTypes := map[OrphanResourceType]bool{
		OrphanResourceTypeReplicaData:     false,
		OrphanResourceTypeEngineInstance:  false,
		OrphanResourceTypeReplicaInstance: false,
	}
```

`UnmarshalOrphanResourceTypes` must not returns `nil` if there's no error, and at least it should returns a non-nil map. Unexpected `nil` should crash it as early as possible.

controller/instance_manager_controller.go
```diff
@@ -612,6 +627,12 @@ func (imc *InstanceManagerController) handlePod(im *longhorn.InstanceManager) er
 		return err
 	}
 
+	if isPodDeletedOrNotRunning {
+		if err := imc.deleteOrphans(im, true); err != nil {
+			log.WithError(err).Warnf("Failed to delete orphan CRs for stopped instance manager pod %v", im.Name)
+		}
+	}
+
```
derekbit, line 635: Move it to `imc.syncOrphans()`

types/types.go
```diff
@@ -726,6 +746,10 @@ func GetOrphanChecksumNameForOrphanedDataStore(nodeID, diskName, diskPath, diskU
 	return orphanPrefix + util.GetStringChecksumSHA256(strings.TrimSpace(fmt.Sprintf("%s-%s-%s-%s-%s", nodeID, diskName, diskPath, diskUUID, dataStore)))
 }
 
+func GetOrphanChecksumNameForOrphanedInstance(instanceName, nodeID, dataEngine string) string {
+	return orphanPrefix + util.GetStringChecksumSHA256(strings.TrimSpace(fmt.Sprintf("%s-%s-%s", instanceName, nodeID, dataEngine)))
+}
```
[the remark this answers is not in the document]
COLDTURNIP: CR object name should conform the DNS subdomain name (RFC 1123):

- contain no more than 253 characters
- contain only lowercase alphanumeric characters, '-' or '.'
- start with an alphanumeric character
- end with an alphanumeric character

The 63 chars limitation is for labels (RFC 1123 and RFC 1035).

> The name of the CRD object itself must be a valid [DNS subdomain name](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names) derived from the defined resource name and its API group; see [how to create a CRD](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#create-a-customresourcedefinition) for more details. Further, the name of an object whose kind/resource is defined by a CRD must also be a valid DNS subdomain name.

controller/instance_manager_controller.go
```diff
@@ -988,6 +1003,32 @@ func (imc *InstanceManagerController) deleteInstanceManagerPDB(im *longhorn.Inst
...
+	} else {
+		switch pod, err := imc.ds.GetPodRO(im.Namespace, im.Name); {
+		case datastore.ErrorIsNotFound(err):
+			isInstanceManagerPodDeletedOrNotRunning = true
+		case err != nil:
+			return errors.Wrapf(err, "failed to check instance manager %v pod state while sync orphans", im.Name)
+		default:
+			isInstanceManagerPodDeletedOrNotRunning = pod == nil
+		}
+	}
```
COLDTURNIP, line 1021: Partially extracted from `areDangerZoneSettingsSyncedToIMPod`, but not share as a method since `areDangerZoneSettingsSyncedToIMPod` needs the pod object for further checks.

derekbit (approved):
LGTM

derekbit:
@COLDTURNIP CI still failed

PR #3695, opened 2025-04-18
feat: implement ublk frontend for v2 data engine - part 3

PhanLe1010:
Support create ublk frontend volume from CSI flow

longhorn/longhorn#9456


derekbit (approved):
LGTM

PR #3715, opened 2025-04-23
fix(deps): add replace directive for `k8s.io/externaljwt`

bachmanity1:

Issue #

Listing all modules fails because there is no replace directive for `k8s.io/externaljwt`
``` bash
> go list -m -mod=mod all
go: k8s.io/externaljwt@v0.0.0: invalid version: unknown revision v0.0.0
```
Comments in the `go.mod` file recommends to add replace directives for all k8s.io dependencies of `k8s.io/kubernetes` that `k8s.io/kubernetes` itself replaces in its go.mod file

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -37,6 +37,7 @@ replace (
 	k8s.io/csi-translation-lib => k8s.io/csi-translation-lib v0.32.3
 	k8s.io/dynamic-resource-allocation => k8s.io/dynamic-resource-allocation v0.32.3
 	k8s.io/endpointslice => k8s.io/endpointslice v0.32.3
+	k8s.io/externaljwt => k8s.io/externaljwt v0.32.3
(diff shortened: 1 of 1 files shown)


derekbit:
@bachmanity1 Why need the dependency?

innobead (approved):

innobead:
> > @bachmanity1 Why need the dependency?
> 
> Actually, it's not really needed since it's not being used. I just noticed that my IDE reported this issue, and I also saw the comment below in the `go.mod`, so I thought it might be helpful to add this replace directive.
> 
> https://github.com/longhorn/longhorn-manager/blob/84262a085e4fab2dd34a2ec32b8a2d2795da529d/go.mod#L18-L20

I guess this might actually cause some problems for your IDE, possibly making it unable to resolve dependencies correctly.

PR #3716, opened 2025-04-24
fix: update vendor for k8s.io/externaljwt

COLDTURNIP:

PR longhorn/longhorn-manager#3715

Fix vendor to resolve build break in CI pipe: https://github.com/longhorn/longhorn-manager/actions/runs/14621878387/job/41023739095?pr=3694#step:4:586


derekbit (approved):

bachmanity1:
I'm really sorry for causing this issue. I'm curious—why do you run build and test pipelines only after a PR is merged, instead of running them before to avoid situations like this?

derekbit:
> I'm really sorry for causing this issue. I'm curious—why do you run build and test pipelines only after a PR is merged, instead of running them before to avoid situations like this?

Oh, it is due to that I merged it too fast. The CI actually failed https://github.com/longhorn/longhorn-manager/actions/runs/14612666514/job/41019021238.

PR #3718, opened 2025-04-24
feat: cleanup orphaned instances while evicting node

COLDTURNIP:

Issue longhorn/longhorn#6764

Implement orphaned runtime instance cleanup for evicting node.
LEP: longhorn/longhorn#10675


controller/node_controller.go
```diff
@@ -1356,7 +1356,33 @@ func (nc *NodeController) getNewAndMissingOrphanedReplicaDataStores(diskName, di
 	return newOrphanedReplicaDataStores, missingOrphanedReplicaDataStores
 }
 
-func (nc *NodeController) deleteOrphans(node *longhorn.Node, diskName string, diskInfo *monitor.CollectedDiskInfo, missingOrphanedReplicaDataStores map[string]string) error {
+func (nc *NodeController) deleteInstanceOrphans(node *longhorn.Node) error {
```
derekbit, line 1359:
```suggestion
func (nc *NodeController) deleteOrphansForEngineAndReplicaInstances(node *longhorn.Node) error {
```

controller/node_controller.go
```diff
@@ -1356,7 +1356,33 @@ func (nc *NodeController) getNewAndMissingOrphanedReplicaDataStores(diskName, di
...
+	if !node.Spec.EvictionRequested {
+		return nil
+	}
```
derekbit, line 1362: Move out of the function.
This will be better

```
if node.Spec.EvictionRequested {
    err = deleteOrphansForEngineAndReplicaInstances()
     ....
}
```

controller/node_controller.go
```diff
@@ -1305,15 +1305,15 @@ func (nc *NodeController) syncOrphans(node *longhorn.Node, collectedDataInfo map
...
 			return errors.Wrapf(err, "failed to create orphans for disk %v", diskName)
 		}
-		if err := nc.deleteOrphans(node, diskName, diskInfo, missingOrphanedReplicaDataStores); err != nil {
+		if err := nc.deleteDataOrphans(node, diskName, diskInfo, missingOrphanedReplicaDataStores); err != nil {
 			return errors.Wrapf(err, "failed to delete orphans for disk %v", diskName)
 		}
 	}
 
-	return nil
+	return nc.deleteInstanceOrphans(node)
```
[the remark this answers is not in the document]
COLDTURNIP: Moved the execution condition out from `deleteOrphansForEngineAndReplicaInstances`.

controller/node_controller.go
```diff
@@ -1356,7 +1360,30 @@ func (nc *NodeController) getNewAndMissingOrphanedReplicaDataStores(diskName, di
 	return newOrphanedReplicaDataStores, missingOrphanedReplicaDataStores
 }
 
-func (nc *NodeController) deleteOrphans(node *longhorn.Node, diskName string, diskInfo *monitor.CollectedDiskInfo, missingOrphanedReplicaDataStores map[string]string) error {
+func (nc *NodeController) deleteOrphansForEngineAndReplicaInstances(node *longhorn.Node) error {
+	nc.logger.Debugf("Deleting orphaned instances to evict node %v", node.Name)
```
derekbit, line 1364: The function actually deletes orphan CRs.

```suggestion
	nc.logger.Infof("Deleting orphans on evicted node %v", node.Name)
```

controller/node_controller.go
```diff
@@ -1356,7 +1360,30 @@ func (nc *NodeController) getNewAndMissingOrphanedReplicaDataStores(diskName, di
...
+		return errors.Wrapf(err, "failed to list orphans to evict node %v", node.Name)
+	}
+
+	multiError := util.NewMultiError()
+	for _, orphan := range orphans {
+		switch orphan.Spec.Type {
+		case longhorn.OrphanTypeEngineInstance, longhorn.OrphanTypeReplicaInstance:
+			if err := nc.ds.DeleteOrphan(orphan.Name); err != nil {
+				multiError.Append(util.NewMultiError(fmt.Sprintf("%v: %v", orphan.Name, err)))
+			}
```
derekbit, line 1377:
```suggestion
			if err := nc.ds.DeleteOrphan(orphan.Name); ; err != nil && !apierrors.IsNotFound(err) {
				multiError.Append(util.NewMultiError(fmt.Sprintf("%v: %v", orphan.Name, err)))
			}
```

controller/node_controller.go
```diff
@@ -1356,7 +1360,30 @@ func (nc *NodeController) getNewAndMissingOrphanedReplicaDataStores(diskName, di
...
+		}
+	}
+	if len(multiError) > 0 {
+		return fmt.Errorf("node controller failed to delete instance orphan CR: %v", multiError.Join())
```
derekbit, line 1381:
```suggestion
		return fmt.Errorf("node controller failed to delete instance orphans: %v", multiError.Join())
```

controller/instance_manager_controller.go
```diff
@@ -1004,6 +1004,17 @@ func (imc *InstanceManagerController) deleteInstanceManagerPDB(im *longhorn.Inst
...
+	node, err := imc.ds.GetNode(im.Spec.NodeID)
+	switch {
+	case datastore.ErrorIsNotFound(err):
+		isNodeDeletedOrEvictionRequested = true
+	case err != nil:
+		return errors.Wrapf(err, "failed to check node eviction status on %v", im.Spec.NodeID)
+	default:
+		isNodeDeletedOrEvictionRequested = node.Spec.EvictionRequested
+	}
+
```
derekbit, line 1017: node eviction is already handled in the node controller. Why do we need the new logic?

COLDTURNIP: Node controler handles lhn update event, and have no knowledge about orphan creation. When a lhn is set to evict then node comes back later, instance manager delete the new detected orphaned instaces here.

derekbit: Got your point but a bit confused. If node is evicted, the node controller will try to list and delete the orphans in every reconciliation. Why does instance manager need the extra deletion?

COLDTURNIP: Or, node controller also register the orphan creation event, WDYT?

derekbit: Why is it related to orphan creation? Aren't we discussing the orphan deletion?

COLDTURNIP: There are two scenarios.

## A. Orphans existed, and evict the nodes

This is simple: node controller reconcile the node update event, delete all orphans because it notice the eviction required.

## B. Node is marked evicted, then new orphaned engine is detected

This may caused by IM monitor scan period delay, or network outage.

There is no node update event in this case to trigger LHN controller reconcilation. We need a way to delete such instance orphan CRs.

**Solution 1.** IM controller is already responsible to delete orphan CRs. Add some check here to delete new orphan CR for instances if it is in eviction.

**Solution 2.** LHN controller register the orphan CR creation events, then the orphaned instances will be cleanup dring node eviction.

derekbit: > There is no node update event in this case to trigger LHN controller reconcilation. We need a way to delete such instance orphan CRs.

https://github.com/longhorn/longhorn-manager/blob/master/controller/node_controller.go#L107-L116
node resync period is 30 seconds, so we don't need to delete orphans in the IM controller, right?

mantissahz: Can we use `func (s *DataStore) IsNodeDeleted(name string) (bool, error) {}` instead?

derekbit (approved):
LGTM. Thanks for the quick improvment.

PR #3719, opened 2025-04-25
fix(orphan): finalize orphan instance CR when instance manager unavai…

COLDTURNIP:
…lable

Issue longhorn/longhorn#10791

When instance manager is down, and the instance manager CR is deleted and waiting for finalizing, the owned instance orphan CRs has to be finalized without instance manager client.

### DIFF
--- a/controller/orphan_controller.go
+++ b/controller/orphan_controller.go
@@ -315,7 +315,7 @@ func (oc *OrphanController) cleanupOrphanedResource(orphan *longhorn.Orphan) (is
 			longhorn.OrphanConditionTypeError, longhorn.ConditionStatusTrue, "", err.Error())
 	}()
 
-	// Make sure if the orphan nodeID and controller ID are same.
+	// Make sure if the orphan nodeID and controller ID are the same.
 	// If NO, just delete the orphan resource object and don't touch the data.
 	if orphan.Spec.NodeID != oc.controllerID {
 		log.WithFields(logrus.Fields{
@@ -370,7 +370,8 @@ func (oc *OrphanController) cleanupOrphanedEngineInstance(orphan *longhorn.Orpha
 	} else {
 		spec = &engineCR.Spec.InstanceSpec
 	}
-	return oc.cleanupOrphanedInstance(orphan, instance, imName, longhorn.InstanceManagerTypeEngine, spec)
+	oc.cleanupOrphanedInstance(orphan, instance, imName, longhorn.InstanceManagerTypeEngine, spec)
+	return true, nil
 }
 
 func (oc *OrphanController) cleanupOrphanedReplicaInstance(orphan *longhorn.Orphan) (isCleanupComplete bool, err error) {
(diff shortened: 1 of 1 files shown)


controller/orphan_controller.go
```diff
@@ -438,13 +441,21 @@ func (oc *OrphanController) cleanupOrphanedInstance(orphan *longhorn.Orphan, ins
 	}()
 
 	if err := oc.deleteInstance(imc, instance, imType, orphan.Spec.DataEngine); err != nil {
-		return false, err
+		oc.logger.WithError(err).Warnf("Failed to delete orphan instance %v due to instance manager client error. Continue to finalize orphan %v", instance, orphan.Name)
```
derekbit, line 444:
```suggestion
		oc.logger.WithError(err).Warnf("Failed to delete orphan instance %v. Continue to finalize orphan %v", instance, orphan.Name)
```

controller/orphan_controller.go
```diff
@@ -438,13 +441,21 @@ func (oc *OrphanController) cleanupOrphanedInstance(orphan *longhorn.Orphan, ins
...
 	}
 	isCleanupComplete, cleanupErr := oc.confirmOrphanInstanceCleanup(imc, instance, imType, orphan.Spec.DataEngine)
-	if cleanupErr == nil && !isCleanupComplete {
-		oc.logger.Infof("Orphan instance %v cleanup in progress", instance)
+	switch {
+	case cleanupErr != nil:
+		oc.logger.WithError(err).Warnf("Failed to confirm deletion of orphan instance %v due to instance manager client error. Continue to finalize orphan %v", instance, orphan.Name)
+		return true
+	case !isCleanupComplete:
+		oc.logger.Infof("Orphan instance %v cleanup in progress, waiting for instance state update", instance)
```
derekbit, line 453: After rethinking `isCleanupComplete`, I think we don't need to make sure the instance is deleted or not because
- The deletion is blocking call
- If the orphan instance is unable to be deleted. Instance-manager monitor will create a new one.

To prevent the potential blocking issue, we can change

```
	if err := oc.deleteInstance(imc, instance, imType, orphan.Spec.DataEngine); err != nil {
		return false, err
	}
	isCleanupComplete, cleanupErr := oc.confirmOrphanInstanceCleanup(imc, instance, imType, orphan.Spec.DataEngine)
	if cleanupErr == nil && !isCleanupComplete {
		oc.logger.Infof("Orphan instance %v cleanup in progress", instance)
	}
	return isCleanupComplete, cleanupErr
```
to
```
	err = imc.InstanceDelete(engineType, instanceName, string(instanceKind), "", false)
	if err != nil && !types.ErrorIsNotFound(err) {
		only logging error ..... and not need to return error
	}
```

COLDTURNIP: Agree, the orphaned instance eventually get deleted. Note that there's still little chance:

1. Delete the orphan CR, and orphan controller ask IM to delete the instance, then finalize the orphan CR.
2. Before the instance delete complete, IM monitor detects the instance existance, and create the orphan CR for it again.
3. IM monitor later observes the instance elimination, delete the orphan CR, and orphan controller finalize it.

As the result, orphan CR disappear -> appear again -> disappear again.

We may still need some annotation on instances inside IM in the future.

derekbit (approved):
LGTM

derekbit:
@coderabbitai review

COLDTURNIP:
Waiting for e2e negative test.

yangchiu:
> Waiting for e2e negative test.

Test results: 
https://ci.longhorn.io/job/private/job/longhorn-e2e-test/2923/
https://ci.longhorn.io/job/private/job/longhorn-e2e-test/2925/

Didn't encounter the issue using the custom longhorn-manager image based on this PR.

derekbit:
Thank you @yangchiu

PR #3729, opened 2025-04-28
feat(csi): enable storage capacity-aware pod scheduling using `CSIStorageCapacity`

bachmanity1:

Issue https://github.com/longhorn/longhorn/issues/10685

Enable storage capacity-aware pod scheduling using `CSIStorageCapacity`. 

Changes:
- set `StorageCapacity` field of csi driver to `true`
- added `GET_CAPACITY` capability and implemented `GetCapacity` rpc
- added `VOLUME_ACCESSIBILITY_CONSTRAINTS` plugin capability and slightly changed `NodeGetInfo` rpc implementation to return accessible topologies. This is needed because kubelet uses response from `NodeGetInfo` rpc to populate `topologyKeys` field of the `CSINode` objects, which is then consumed by the external-provisioner to build the `AccessibleTopolgies` field of the `GetCapacityRequest`.
- added flags to the external-provisioner deployment to make it automatically produce `CSIStorageCapacity` objects.

These changes won't affect existing users unless they use storage class with `volumeBindingMode` set to the `WaitForFirstConsumer`.

I've tested this PR manually in my local k8s cluster. It seems to be working well. I'm planning to add integrations tests in the https://github.com/longhorn/longhorn-tests/tree/master/manager repository once/if this PR is accecpted.

### DIFF
--- a/csi/controller_server.go
+++ b/csi/controller_server.go
@@ -691,9 +691,10 @@ func (cs *ControllerServer) GetCapacity(ctx context.Context, req *csi.GetCapacit
 		return nil, status.Errorf(codes.InvalidArgument, "failed to parse node id: %v", err)
 	}
 	node, err := cs.lhClient.LonghornV1beta2().Nodes(cs.lhNamespace).Get(ctx, nodeID, metav1.GetOptions{})
-	if apierrors.IsNotFound(err) {
-		return nil, status.Errorf(codes.NotFound, "node %s not found", nodeID)
-	} else if err != nil {
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil, status.Errorf(codes.NotFound, "node %s not found", nodeID)
+		}
 		return nil, status.Errorf(codes.Internal, "unexpected error: %v", err)
 	}
 	if types.GetCondition(node.Status.Conditions, longhorn.NodeConditionTypeReady).Status != longhorn.ConditionStatusTrue {
@@ -708,7 +709,7 @@ func (cs *ControllerServer) GetCapacity(ctx context.Context, req *csi.GetCapacit
 
 	allowEmptyNodeSelectorVolume, err := cs.getSettingAsBoolean(types.SettingNameAllowEmptyNodeSelectorVolume)
 	if err != nil {
-		return nil, status.Errorf(codes.Internal, "failed to get setting, err: %v", err)
+		return nil, status.Errorf(codes.Internal, "failed to get setting %v: %v", types.SettingNameAllowEmptyNodeSelectorVolume, err)
 	}
 	var nodeSelector []string
 	if nodeSelectorRaw, ok := scParameters["nodeSelector"]; ok && len(nodeSelectorRaw) > 0 {
@@ -724,7 +725,7 @@ func (cs *ControllerServer) GetCapacity(ctx context.Context, req *csi.GetCapacit
 	}
 	allowEmptyDiskSelectorVolume, err := cs.getSettingAsBoolean(types.SettingNameAllowEmptyDiskSelectorVolume)
 	if err != nil {
-		return nil, status.Errorf(codes.Internal, "failed to get setting, err: %v", err)
+		return nil, status.Errorf(codes.Internal, "failed to get setting %v: %v", types.SettingNameAllowEmptyDiskSelectorVolume, err)
 	}
 
 	var v1AvailableCapacity int64 = 0
--- a/csi/controller_server.go
+++ b/csi/controller_server.go
@@ -20,7 +20,6 @@ import (
 	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/types/known/timestamppb"
 
-	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/client-go/rest"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -682,55 +681,78 @@ func (cs *ControllerServer) GetCapacity(ctx context.Context, req *csi.GetCapacit
 		}
 	}()
 
-	dataEngine, err := parseDataEngine(req.GetParameters())
-	if err != nil {
-		return nil, status.Errorf(codes.InvalidArgument, "failed to parse engine type: %v", err)
-	}
 	nodeID, err := parseNodeID(req.GetAccessibleTopology())
 	if err != nil {
 		return nil, status.Errorf(codes.InvalidArgument, "failed to parse node id: %v", err)
 	}
-
 	node, err := cs.lhClient.LonghornV1beta2().Nodes(cs.lhNamespace).Get(ctx, nodeID, metav1.GetOptions{})
 	if apierrors.IsNotFound(err) {
 		return nil, status.Errorf(codes.NotFound, "node %s not found", nodeID)
 	} else if err != nil {
 		return nil, status.Errorf(codes.Internal, "unexpected error: %v", err)
 	}
-	v1CapacitySize := resource.NewQuantity(0, resource.BinarySI)
-	v2CapacitySize := resource.NewQuantity(0, resource.BinarySI)
-	for _, diskStatus := range node.Status.DiskStatus {
+
+	scParameters := req.GetParameters()
+	if scParameters == nil {
+		scParameters = map[string]string{}
+	}
+	var diskSelector []string
+	if diskSelectorRaw, ok := scParameters["diskSelector"]; ok && len(diskSelectorRaw) > 0 {
+		diskSelector = strings.Split(diskSelectorRaw, ",")
+	}
+	allowEmptyDiskSelectorVolume, err := cs.getSettingAsBoolean(types.SettingNameAllowEmptyDiskSelectorVolume)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to get %v setting, err: %v", types.SettingNameAllowEmptyDiskSelectorVolume, err)
+	}
+
+	var v1AvailableCapacity int64 = 0
+	var v2AvailableCapacity int64 = 0
+	for diskName, diskStatus := range node.Status.DiskStatus {
+		diskSpec, exists := node.Spec.Disks[diskName]
+		if !exists {
+			continue
+		}
+		if !diskSpec.AllowScheduling || diskSpec.EvictionRequested {
+			continue
+		}
+		if types.GetCondition(diskStatus.Conditions, longhorn.DiskConditionTypeSchedulable).Status != longhorn.ConditionStatusTrue {
+			continue
+		}
+		if !types.IsSelectorsInTags(diskSpec.Tags, diskSelector, allowEmptyDiskSelectorVolume) {
+			continue
+		}
+		storageSchedulable := diskStatus.StorageAvailable - diskSpec.StorageReserved
 		if diskStatus.Type == longhorn.DiskTypeFilesystem {
-			v1CapacitySize.Add(*resource.NewQuantity(diskStatus.StorageAvailable, resource.BinarySI))
+			v1AvailableCapacity = max(v1AvailableCapacity, storageSchedulable)
 		}
 		if diskStatus.Type == longhorn.DiskTypeBlock {
-			v2CapacitySize.Add(*resource.NewQuantity(diskStatus.StorageAvailable, resource.BinarySI))
+			v2AvailableCapacity = max(v2AvailableCapacity, storageSchedulable)
 		}
 	}
 
+	dataEngine, ok := scParameters["dataEngine"]
+	if !ok {
+		return nil, status.Error(codes.InvalidArgument, "storage class parameters missing 'dataEngine' key")
+	}
 	rsp := &csi.GetCapacityResponse{}
-	switch dataEngine {
+	switch longhorn.DataEngineType(dataEngine) {
 	case longhorn.DataEngineTypeV1:
-		rsp.AvailableCapacity = v1CapacitySize.Value()
+		rsp.AvailableCapacity = v1AvailableCapacity
 	case longhorn.DataEngineTypeV2:
-		rsp.AvailableCapacity = v2CapacitySize.Value()
+		rsp.AvailableCapacity = v2AvailableCapacity
 	default:
 		return nil, status.Errorf(codes.InvalidArgument, "unknown data engine type %v", dataEngine)
 	}
 
-	log.Infof("Node: %s, DataEngine: %s, v1CapacitySize: %s, v2CapacitySize: %s", nodeID, dataEngine, v1CapacitySize, v2CapacitySize)
+	log.Infof("Node: %s, DataEngine: %s, v1AvailableCapacity: %d, v2AvailableCapacity: %d", nodeID, dataEngine, v1AvailableCapacity, v2AvailableCapacity)
 	return rsp, nil
 }
 
-func parseDataEngine(parameters map[string]string) (longhorn.DataEngineType, error) {
-	if parameters == nil {
-		return "", fmt.Errorf("missing storage class parameters")
+func max(x, y int64) int64 {
+	if x > y {
+		return x
 	}
-	dataEngine, ok := parameters["dataEngine"]
-	if !ok {
-		return "", fmt.Errorf("storage class parameters missing data engine key")
-	}
-	return longhorn.DataEngineType(dataEngine), nil
+	return y
 }
 
 func parseNodeID(topology *csi.Topology) (string, error) {
@@ -742,6 +764,18 @@ func parseNodeID(topology *csi.Topology) (string, error) {
 		return "", fmt.Errorf("accessible topology request parameter is missing %s key", nodeTopologyKey)
 	}
 	return nodeId, nil
+}
+
+func (cs *ControllerServer) getSettingAsBoolean(name types.SettingName) (bool, error) {
+	obj, err := cs.lhClient.LonghornV1beta2().Settings(cs.lhNamespace).Get(context.TODO(), string(name), metav1.GetOptions{})
+	if err != nil {
+		return false, err
+	}
+	value, err := strconv.ParseBool(obj.Value)
+	if err != nil {
+		return false, err
+	}
+	return value, nil
 }
 
 func (cs *ControllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateSnapshotRequest) (*csi.CreateSnapshotResponse, error) {
--- a/csi/controller_server.go
+++ b/csi/controller_server.go
@@ -5,11 +5,6 @@ import (
 	"crypto/sha256"
 	"encoding/json"
 	"fmt"
-	lhclientset "github.com/longhorn/longhorn-manager/k8s/pkg/client/clientset/versioned"
-	"k8s.io/apimachinery/pkg/api/resource"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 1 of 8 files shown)


csi/controller_server.go
```diff
@@ -642,8 +670,78 @@ func (cs *ControllerServer) ListVolumes(context.Context, *csi.ListVolumesRequest
...
+	dataEngine, err := parseDataEngine(req.GetParameters())
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "failed to parse engine type: %v", err)
+	}
+	nodeID, err := parseNodeID(req.GetAccessibleTopology())
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "failed to parse node id: %v", err)
+	}
+
+	node, err := cs.lhClient.LonghornV1beta2().Nodes(cs.lhNamespace).Get(context.TODO(), nodeID, metav1.GetOptions{})
```
ahnuyh, line 694:
```suggestion
	node, err := cs.lhClient.LonghornV1beta2().Nodes(cs.lhNamespace).Get(ctx, nodeID, metav1.GetOptions{})
```

bachmanity1:
Hi @derekbit @c3y1huang, could you please take a look at this PR?

derekbit:
Hello @hookak
Could you help review the PR when you are available? Thanks.

hookak (approved):
Everything looks good! @derekbit 
Could you squash the commits into meaningful ones? @bachmanity1

derekbit:
> Everything looks good! @derekbit Could you squash the commits into meaningful ones? @bachmanity1

Thanks @derekbit @ahnuyh for your reviews. I'm reviewing the PR as well.

csi/manager.go
```diff
@@ -41,7 +41,11 @@ func (m *Manager) Run(driverName, nodeID, endpoint, identityVersion, managerURL
 		return errors.Wrap(err, "Failed to create CSI node server ")
 	}
 
-	m.cs = NewControllerServer(apiClient, nodeID)
+	m.cs, err = NewControllerServer(apiClient, nodeID)
+	if err != nil {
+		return errors.Wrap(err, "Failed to create CSI controller server ")
```
derekbit:
```suggestion
		return errors.Wrap(err, "failed to create CSI controller server")
```

csi/controller_server.go
```diff
@@ -642,8 +669,135 @@ func (cs *ControllerServer) ListVolumes(context.Context, *csi.ListVolumesRequest
...
+	nodeID, err := parseNodeID(req.GetAccessibleTopology())
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "failed to parse node id: %v", err)
+	}
+	node, err := cs.lhClient.LonghornV1beta2().Nodes(cs.lhNamespace).Get(ctx, nodeID, metav1.GetOptions{})
+	if apierrors.IsNotFound(err) {
+		return nil, status.Errorf(codes.NotFound, "node %s not found", nodeID)
+	} else if err != nil {
+		return nil, status.Errorf(codes.Internal, "unexpected error: %v", err)
+	}
```
derekbit:
```suggestion
    if err != nil {
	    if apierrors.IsNotFound(err) {
		    return nil, status.Errorf(codes.NotFound, "node %s not found", nodeID)
        }
		return nil, status.Errorf(codes.Internal, "unexpected error: %v", err)
	}
```

csi/controller_server.go
```diff
@@ -642,8 +669,135 @@ func (cs *ControllerServer) ListVolumes(context.Context, *csi.ListVolumesRequest
...
+	if types.GetCondition(node.Status.Conditions, longhorn.NodeConditionTypeSchedulable).Status != longhorn.ConditionStatusTrue {
+		return &csi.GetCapacityResponse{}, nil
+	}
+	if !node.Spec.AllowScheduling || node.Spec.EvictionRequested {
+		return &csi.GetCapacityResponse{}, nil
+	}
+
+	allowEmptyNodeSelectorVolume, err := cs.getSettingAsBoolean(types.SettingNameAllowEmptyNodeSelectorVolume)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to get setting, err: %v", err)
```
derekbit:
```suggestion
		return nil, status.Errorf(codes.Internal, "failed to get setting %v: %v", types.SettingNameAllowEmptyNodeSelectorVolume, err)
```

csi/controller_server.go
```diff
@@ -642,8 +669,135 @@ func (cs *ControllerServer) ListVolumes(context.Context, *csi.ListVolumesRequest
...
+	}
+	var nodeSelector []string
+	if nodeSelectorRaw, ok := scParameters["nodeSelector"]; ok && len(nodeSelectorRaw) > 0 {
+		nodeSelector = strings.Split(nodeSelectorRaw, ",")
+	}
+	if !types.IsSelectorsInTags(node.Spec.Tags, nodeSelector, allowEmptyNodeSelectorVolume) {
```
derekbit, line 718: Awesome! You've considered the setting `allowEmptyNodeSelectorVolume`.

csi/controller_server.go
```diff
@@ -642,8 +669,135 @@ func (cs *ControllerServer) ListVolumes(context.Context, *csi.ListVolumesRequest
...
+	dataEngine, ok := scParameters["dataEngine"]
+	if !ok {
+		return nil, status.Error(codes.InvalidArgument, "storage class parameters missing 'dataEngine' key")
+	}
+	rsp := &csi.GetCapacityResponse{}
+	switch longhorn.DataEngineType(dataEngine) {
+	case longhorn.DataEngineTypeV1:
+		rsp.AvailableCapacity = v1AvailableCapacity
+	case longhorn.DataEngineTypeV2:
+		rsp.AvailableCapacity = v2AvailableCapacity
```
derekbit, line 765: I just realized that `GetCapacity` retrieves the maximum available capacity among the schedulable disks on a node, but it doesn't return the corresponding disk name. Due to the current scheduling strategies, the replica scheduler might place the replica on a different disk instead of the one with the maximum available capacity. This is something we can consider improving in the scheduler in the future.

csi/deployment.go
```diff
@@ -117,6 +117,8 @@ func NewProvisionerDeployment(namespace, serviceAccount, provisionerImage, rootD
 			"--leader-election",
 			"--leader-election-namespace=$(POD_NAMESPACE)",
 			"--default-fstype=ext4",
+			"--enable-capacity",
+			"--capacity-ownerref-level=2",
```
derekbit, line 121: Could you elaborate on why `capacity-ownerref-level` is set to 2?

csi/deployment.go
```diff
@@ -591,13 +593,14 @@ type DriverObjectDeployment struct {
 }
 
 func NewCSIDriverObject() *DriverObjectDeployment {
-	falseFlag := true
+	trueFlag := true
```
derekbit: Good catch!

csi/deployment.go
derekbit: Figured out the answer.

When the external-provisioner creates a `CSIStorageCapacity` object with `--capacity-ownerref-level=2`, it will add an OwnerReference pointing to its Deployment. This ensures that if the external-provisioner's Deployment is deleted, the associated capacity information will also be automatically cleaned up, preventing resource leakage.

derekbit: Figured out the answer.

When the external-provisioner creates a `CSIStorageCapacity` object with `--capacity-ownerref-level=2`, it will add an OwnerReference pointing to its Deployment. This ensures that if the external-provisioner's Deployment is deleted, the associated capacity information will also be automatically cleaned up, preventing resource leakage.

Correct?

bachmanity1: yes, right.

csi/controller_server.go
bachmanity1: I think the return value of `GetCapacity` only indicates the maximum size of a replica that can be placed on the node. For example, if it returns 100Gi, it means there's at least one disk that can support a replica of that size, but there could be multiple disks that meet this requirement. It's then the replica scheduler's job to choose the actual disk.

derekbit: cc @c3y1huang

bachmanity1:
Hi @davidcheng0922, when you have a moment, could you kindly review this PR so we can get it merged this week?

derekbit:
cc @shuo-wu @PhanLe1010 @c3y1huang  for review

csi/controller_server.go
```diff
@@ -642,8 +669,136 @@ func (cs *ControllerServer) ListVolumes(context.Context, *csi.ListVolumesRequest
...
+	}
+	if types.GetCondition(node.Status.Conditions, longhorn.NodeConditionTypeReady).Status != longhorn.ConditionStatusTrue {
+		return &csi.GetCapacityResponse{}, nil
+	}
+	if types.GetCondition(node.Status.Conditions, longhorn.NodeConditionTypeSchedulable).Status != longhorn.ConditionStatusTrue {
+		return &csi.GetCapacityResponse{}, nil
+	}
+	if !node.Spec.AllowScheduling || node.Spec.EvictionRequested {
+		return &csi.GetCapacityResponse{}, nil
+	}
```
shuo-wu, line 708: I am not sure whether we can return an empty response rather than an error when the node is not ready. [The csi spec doc](https://github.com/container-storage-interface/spec/blob/master/spec.md#getcapacity) does not mention this issue explicitly. Probably returning an empty/zero response here is enough and means not available in the selected Topology/Node

csi/controlller_server_test.go
```diff
@@ -0,0 +1,318 @@
+package csi
+
+import (
+	"context"
+	"fmt"
+	"github.com/longhorn/longhorn-manager/types"
```
shuo-wu: NIT: Re-arrange imports

shuo-wu (dismissed):
In general LGTM

csi/controller_server.go
bachmanity1: Returning empty response would prevent kube-scheduler from scheduling any workload on the given node.

csi/controller_server.go
```diff
@@ -642,8 +669,136 @@ func (cs *ControllerServer) ListVolumes(context.Context, *csi.ListVolumesRequest
...
+		rsp.AvailableCapacity = v2AvailableCapacity
+	default:
+		return nil, status.Errorf(codes.InvalidArgument, "unknown data engine type %v", dataEngine)
+	}
+
+	log.Infof("Node: %s, DataEngine: %s, v1AvailableCapacity: %d, v2AvailableCapacity: %d", nodeID, dataEngine, v1AvailableCapacity, v2AvailableCapacity)
+	return rsp, nil
+}
+
+func max(x, y int64) int64 {
```
shuo-wu: NIT: Move the below new functions to `csi/util.go`

shuo-wu (approved):
LGTM

derekbit:
@c3y1huang Could you do a final review? Thank you.

csi/util.go
```diff
@@ -466,3 +468,21 @@ func requiresSharedAccess(vol *longhornclient.Volume, cap *csi.VolumeCapability)
 func getStageBlockVolumePath(stagingTargetPath, volumeID string) string {
 	return filepath.Join(stagingTargetPath, volumeID)
 }
+
+func max(x, y int64) int64 {
+	if x > y {
+		return x
+	}
+	return y
+}
```
c3y1huang: Can we use [built-in#max](https://pkg.go.dev/builtin#max)?

csi/deployment_util.go
```diff
@@ -105,6 +105,22 @@ func getCommonDeployment(commonName, namespace, serviceAccount, image, rootDir s
...
 									},
 								},
+								{
+									Name: "NAMESPACE",
+									ValueFrom: &corev1.EnvVarSource{
+										FieldRef: &corev1.ObjectFieldSelector{
+											FieldPath: "metadata.namespace",
+										},
+									},
+								},
```
c3y1huang: What is the purpose of this env variable? And is there a specific reason for duplicating the namespace value into a different env variable (already defined in `Name: "POD_NAMESPACE"`)?

csi/deployment_util.go
```diff
@@ -105,6 +105,22 @@ func getCommonDeployment(commonName, namespace, serviceAccount, image, rootDir s
...
+								{
+									Name: "POD_NAME",
```
c3y1huang, line 119: Same question here; what is this env used for?

csi/deployment.go
```diff
@@ -591,13 +593,14 @@ type DriverObjectDeployment struct {
...
 	obj := &storagev1.CSIDriver{
 		ObjectMeta: metav1.ObjectMeta{
 			Name: types.LonghornDriverName,
 		},
 		Spec: storagev1.CSIDriverSpec{
-			PodInfoOnMount: &falseFlag,
+			PodInfoOnMount:  &trueFlag,
+			StorageCapacity: &trueFlag,
```
c3y1huang:
```suggestion
			PodInfoOnMount:  ptr.To(true),
			StorageCapacity: ptr.To(true),
```

csi/controller_server.go
```diff
@@ -642,8 +669,118 @@ func (cs *ControllerServer) ListVolumes(context.Context, *csi.ListVolumesRequest
...
+			continue
+		}
+		storageSchedulable := diskStatus.StorageAvailable - diskSpec.StorageReserved
+		if diskStatus.Type == longhorn.DiskTypeFilesystem {
+			v1AvailableCapacity = max(v1AvailableCapacity, storageSchedulable)
+		}
+		if diskStatus.Type == longhorn.DiskTypeBlock {
+			v2AvailableCapacity = max(v2AvailableCapacity, storageSchedulable)
+		}
+	}
```
c3y1huang, line 754: NIT: Consider extracting this logic into a helper function, e.g., `func getAvailableCapacityForNode(...) (int64)`
```go
rsp.AvailableCapacity = getAvailableCapacityForNode(...)
```

csi/controlller_server_test.go
```diff
@@ -0,0 +1,319 @@
...
+		skipNodeCreation        bool
+		skipNodeSettingCreation bool
+		skipDiskSettingCreation bool
+		dataEngine              string
+		diskSelector            string
+		nodeSelector            string
+		availableCapacity       int64
+		disks                   []*disk
+		err                     error
+	}{
```
c3y1huang: The test output doesn't show what scenarios were covered. This makes it difficult to know which csae passed or failed.
```bash
--- PASS: TestGetCapacity (0.00s)
PASS
ok      github.com/longhorn/longhorn-manager/csi        0.011s
```

Suggest to use `t.Run()` to improve readability.

```go
for testName, testCase := range tests {
	t.Run(testName, func(t *testing.T) {
		// test logic here
	})
}
```
Example:
```go
	for testName, testCase := range map[string]struct {
		node                    *longhorn.Node
		skipNodeCreation        bool
		skipNodeSettingCreation bool
		skipDiskSettingCreation bool
		dataEngine              string
		diskSelector            string
		nodeSelector            string
		availableCapacity       int64
		disks                   []*disk
		err                     error
	}{
        "Node not found": {
			skipNodeCreation: true,
			node:             newNode("node-0", "storage", true, true, true	, false),
			err:              status.Errorf(codes.NotFound, "node node-0 not found"),
		},
    } {
		t.Run(testName, func(t *testing.T) {
            // test logic here
	    })
    }
```
Example output after change:
```bash
--- PASS: TestGetCapacity (0.00s)
    --- PASS: TestGetCapacity/Node_not_found (0.00s)
    --- PASS: TestGetCapacity/Node_setting_not_found (0.00s)
    --- PASS: TestGetCapacity/Disk_setting_not_found (0.00s)
    --- PASS: TestGetCapacity/Missing_data_engine_type (0.00s)
    --- PASS: TestGetCapacity/Unknown_data_engine_type (0.00s)
    --- PASS: TestGetCapacity/Data_engine_v1_with_no_disks (0.00s)
    --- PASS: TestGetCapacity/Data_engine_v2_with_no_disks (0.00s)
PASS
ok      github.com/longhorn/longhorn-manager/csi        0.011s
```

csi/deployment_util.go
bachmanity1: `NAMESPACE` and `POD_NAME` env variables are needed to correctly [configure owner references](https://github.com/kubernetes-csi/external-provisioner#:~:text=The%20level%20indicates%20the%20number%20of%20objects%20that%20need%20to%20be%20traversed%20starting%20from%20the%20pod%20identified%20by%20the%20POD_NAME%20and%20NAMESPACE%20environment%20variables%20to%20reach%20the%20owning%20object%20for%20CSIStorageCapacity%20objects) for the `CSIStorageCapacity` objects. Since the external-provisioner expects these specific variable names and doesn't support alternatives like `POD_NAMESPACE`, we need to explicitly define them.

csi/controller_server.go
bachmanity1: I'm not sure this is a good idea because `GetCapacity` rpc itself is about getting available capacity for the given node. I don't see how we can logically separate the part shown above into a helper function.

csi/deployment_util.go
bachmanity1: replied above.

csi/controlller_server_test.go
bachmanity1:
```
--- PASS: TestGetCapacity (0.01s)
    --- PASS: TestGetCapacity/Node_not_found (0.00s)
    --- PASS: TestGetCapacity/Node_setting_not_found (0.00s)
    --- PASS: TestGetCapacity/Disk_setting_not_found (0.00s)
    --- PASS: TestGetCapacity/Missing_data_engine_type (0.00s)
    --- PASS: TestGetCapacity/Unknown_data_engine_type (0.00s)
    --- PASS: TestGetCapacity/v1_engine_with_no_disks (0.00s)
    --- PASS: TestGetCapacity/v2_engine_with_no_disks (0.00s)
    --- PASS: TestGetCapacity/Node_condition_is_not_ready (0.00s)
    --- PASS: TestGetCapacity/Node_condition_is_not_schedulable (0.00s)
    --- PASS: TestGetCapacity/Scheduling_not_allowed_on_a_node (0.00s)
    --- PASS: TestGetCapacity/Node_eviction_is_requested (0.00s)
    --- PASS: TestGetCapacity/Node_tags_don't_match_node_selector (0.00s)
    --- PASS: TestGetCapacity/v1_engine_with_two_valid_disks (0.00s)
    --- PASS: TestGetCapacity/v1_engine_with_two_valid_disks_and_one_with_mismatched_engine_type (0.00s)
    --- PASS: TestGetCapacity/v2_engine_with_two_valid_disks_and_one_with_mismatched_engine_type (0.00s)
    --- PASS: TestGetCapacity/v2_engine_with_one_valid_disk_and_two_with_unmatched_tags (0.00s)
    --- PASS: TestGetCapacity/v2_engine_with_one_valid_disk_and_one_with_unhealthy_condition (0.00s)
    --- PASS: TestGetCapacity/v2_engine_with_one_valid_disk_and_one_with_scheduling_disabled (0.00s)
    --- PASS: TestGetCapacity/v2_engine_with_one_valid_disk_and_one_marked_for_eviction (0.00s)

```

csi/controller_server.go
c3y1huang: No problem. Thanks for explaining.

csi/deployment_util.go
c3y1huang: Got it. Could you add the explanation as a comment? It will help others understand why those env var names are required.

shuo-wu (approved):

c3y1huang (approved):

PR #3730, opened 2025-04-28
fix(upgrade): prevent override existing orphan data deletion setting

COLDTURNIP:
#### Which issue(s) this PR fixes:

Issue longhorn/longhorn#6764
https://github.com/longhorn/longhorn/issues/6764#issuecomment-2834461677

replaced by #3731

#### What this PR does / why we need it:

- Upgrade stage does not create new objects. It update only the existing objects.
- New settings are created later from setting definition after upgrade.
- Need to update the default value of `orphan-resource-auto-deletion` from old setting `orphan-auto-deletion` to prevent accidentally overriding.

PR #3731, opened 2025-04-28
feat(setting): replace orphan-auto-deletion with orphan-resource-auto…

COLDTURNIP:

Issue longhorn/longhorn#6764
https://github.com/longhorn/longhorn/issues/6764#issuecomment-2834461677

replaces #3730

- Upgrade stage does not create new objects. It update only the existing objects, and remove settings without definition.
- Leave settings to be replaced in settings setup stage, replacing the old settings, and remove the replaced settings at the end of this stage.
- Since the old `orphan-auto-deletion` settings is replaced and removed, we don't need to sync it in setting controller.

### DIFF
--- a/upgrade/util/util.go
+++ b/upgrade/util/util.go
@@ -339,7 +339,8 @@ func DeleteRemovedSettings(namespace string, lhClient *lhclientset.Clientset) er
 				return true
 			}
 		}
-		return false
+		// replaced settings will be deleted after replaced
+		return types.IsSettingReplaced(name)
 	}
 
 	existingSettingList, err := lhClient.LonghornV1beta2().Settings(namespace).List(context.TODO(), metav1.ListOptions{})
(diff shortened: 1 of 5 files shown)


COLDTURNIP:
If the improvement proposal longhorn/longhorn#10816 is acceptable, I'd like to postpone the entire settings deletion lifecycle to settings setup stage, to make settings replaceable in the future.

Not sure if it is too risky just before the coming release cycle? @derekbit @innobead

derekbit (dismissed):

upgrade/util/util.go
```diff
@@ -400,6 +401,10 @@ func ListAndUpdateSettingsInProvidedCache(namespace string, lhClient *lhclientse
 		return nil, err
 	}
 	for _, setting := range settingList.Items {
+		if _, exist := types.GetSettingDefinition(types.SettingName(setting.Name)); !exist {
+			logrus.Infof("Found unknown setting %v, skipping", setting.Name)
```
derekbit, line 405:
```suggestion
			logrus.Warnf("Found unknown setting %v, skipping", setting.Name)
```

derekbit:
In general, LGTM. Only some minor comments.

mantissahz (dismissed):
In general, LGTM

COLDTURNIP:
Minor update, and this PR is ready to merge.

derekbit (approved):
LGTM

COLDTURNIP:
@mergify backport v1.9.x

derekbit:
@mergify backport v1.9.x

PR #3736, opened 2025-04-30
feat(rebuild): implement per-volume and global QoS control for replic…

hookak:
…a rebuilding (v2 engine)

- Adds  in EngineMonitor to apply QoS during rebuilding
- Supports per-volume  and global setting fallback
- Adds  to  to track actual QoS applied
- Updates CRD, types, and applyconfig for the new fields

Issue https://github.com/longhorn/longhorn/issues/10770

- Adds support for applying replica rebuilding QoS on v2 (SPDK) engine replicas
- Introduces a global setting `v2-data-engine-rebuilding-mbytes-per-second` to control default bandwidth limit
- Adds a per-volume override field `rebuildingMbytesPerSecond` in VolumeSpec
- Ensures actual QoS application is tracked via `AppliedRebuildingMbps` in Engine.Status.RebuildStatus
- Updates CRD schema and generated applyconfiguration files

- If both global and per-volume values are 0, no QoS will be applied (unlimited)
- Rebuild QoS is only supported for `DataEngine: v2` volumes

### DIFF
--- a/controller/engine_controller.go
+++ b/controller/engine_controller.go
@@ -944,6 +944,10 @@ func (m *EngineMonitor) refresh(engine *longhorn.Engine) error {
 		if err != nil {
 			return err
 		}
+
+		if err := m.checkAndApplyRebuildQoS(engine, engineClientProxy, rebuildStatus); err != nil {
+			return err
+		}
 		engine.Status.RebuildStatus = rebuildStatus
 
 		// It's meaningless to sync the trim related field for old engines or engines in old engine instance managers
@@ -1127,6 +1131,62 @@ func (m *EngineMonitor) refresh(engine *longhorn.Engine) error {
 	}
 
 	return nil
+}
+
+func (m *EngineMonitor) checkAndApplyRebuildQoS(engine *longhorn.Engine, engineClientProxy engineapi.EngineClientProxy, rebuildStatus map[string]*longhorn.RebuildStatus) error {
+	if !types.IsDataEngineV2(engine.Spec.DataEngine) {
+		return nil
+	}
+
+	expectedQoSValue, err := m.getEffectiveRebuildQoS(engine)
+	if err != nil {
+		return err
+	}
+
+	for replica, newStatus := range rebuildStatus {
+		if newStatus == nil {
+			continue
+		}
+
+		var appliedQoS int64
+		if oldStatus, exists := engine.Status.RebuildStatus[replica]; exists && oldStatus != nil {
+			appliedQoS = oldStatus.AppliedRebuildingMBps
+		}
+
+		if appliedQoS == expectedQoSValue {
+			newStatus.AppliedRebuildingMBps = appliedQoS
+			continue
+		}
+
+		if !newStatus.IsRebuilding {
+			continue
+		}
+
+		if err := engineClientProxy.ReplicaRebuildQosSet(engine, expectedQoSValue); err != nil {
+			m.logger.WithError(err).Warnf("[qos] Failed to set QoS for volume %s, replica %s", engine.Spec.VolumeName, replica)
+			continue
+		}
+		newStatus.AppliedRebuildingMBps = expectedQoSValue
+	}
+	return nil
+}
+
+func (m *EngineMonitor) getEffectiveRebuildQoS(engine *longhorn.Engine) (int64, error) {
+	globalQoS, err := m.ds.GetSettingAsInt(types.SettingNameV2DataEngineRebuildingMbytesPerSecond)
+	if err != nil {
+		return 0, err
+	}
+
+	volume, err := m.ds.GetVolumeRO(engine.Spec.VolumeName)
+	if err != nil {
+		return 0, err
+	}
+
+	if volume.Spec.RebuildingMbytesPerSecond > 0 {
+		return volume.Spec.RebuildingMbytesPerSecond, nil
+	}
+
+	return globalQoS, nil
 }
 
 func isBackupRestoreFailed(rsMap map[string]*longhorn.RestoreStatus) bool {
@@ -1616,7 +1676,6 @@ func (ec *EngineController) ReconcileEngineState(e *longhorn.Engine) error {
 	if err := ec.rebuildNewReplica(e); err != nil {
 		return err
 	}
-
 	return nil
 }
 
--- a/k8s/pkg/apis/longhorn/v1beta2/engine.go
+++ b/k8s/pkg/apis/longhorn/v1beta2/engine.go
@@ -77,6 +77,8 @@ type RebuildStatus struct {
 	State string `json:"state"`
 	// +optional
 	FromReplicaAddress string `json:"fromReplicaAddress"`
+	// +optional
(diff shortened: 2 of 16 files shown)


derekbit:
Hello @hookak 
Thanks for the contribution. Sorry for late reply because of the upcoming v1.9.0 release. We will review the PRs soon and put it in the v1.10.

hookak:
We're also very excited about the features coming in v1.9.0.
When you have time, we'd really appreciate a review. 

For the overall design of the feature, please refer to this comment:
https://github.com/longhorn/longhorn/issues/10770#issuecomment-2841700145 @derekbit

derekbit:
Hello @hookak 
Could you help update the go.mod? Thank you.

hookak:
I've updated it. Thank you in advance for taking the time to review it

controller/engine_controller.go
```diff
@@ -1129,6 +1133,62 @@ func (m *EngineMonitor) refresh(engine *longhorn.Engine) error {
...
+		var appliedQoS int64
+		if oldStatus, exists := engine.Status.RebuildStatus[replica]; exists && oldStatus != nil {
+			appliedQoS = oldStatus.AppliedRebuildingMbps
+		}
+
+		if appliedQoS == expectedQoSValue {
+			newStatus.AppliedRebuildingMbps = appliedQoS
+			continue
+		}
+
```
shuo-wu: NIT: I am thinking if the spdk server needs to directly return this. This means adding a new field in [ReplicaRebuildStatusResponse](https://github.com/longhorn/types/blob/b79d2bcff04b0614f18c0b41414888793f681c1d/protobuf/ptypes/syncagent.proto#L164) and [ReplicaRebuildingDstShallowCopyCheckResponse](https://github.com/longhorn/types/blob/b79d2bcff04b0614f18c0b41414888793f681c1d/protobuf/spdkrpc/spdk.proto#L217)

shuo-wu (dismissed):
In general LGTM

controller/engine_controller.go
hookak: I think before calling the `checkAndApplyRebuildQoS` function, we already invoke `rebuildStatus, err := engineClientProxy.ReplicaRebuildStatus(engine)`.
This means we're retrieving the latest rebuild status at that point, so I don't think it's necessary to request it again from the SPDK server within the function.
What do you think?

derekbit: WDYT @shuo-wu ?

shuo-wu: I mean, we can integrate qos info into `rebuildStatus`. Anyway the current implementation is good enough.

k8s/pkg/apis/longhorn/v1beta2/volume.go
```diff
@@ -306,6 +306,10 @@ type VolumeSpec struct {
 	// - disabled: Disable offline rebuilding for this volume, regardless of the global setting
 	// +optional
 	OfflineRebuilding VolumeOfflineRebuilding `json:"offlineRebuilding"`
+	// RebuildingMBytesPerSecond limits the write bandwidth (in megabytes per second) on the destination replica during rebuilding.
+	// +kubebuilder:validation:Minimum=0
```
derekbit, line 310: Could you mention what `0` means in the description?

k8s/pkg/apis/longhorn/v1beta2/engine.go
```diff
@@ -77,6 +77,8 @@ type RebuildStatus struct {
 	State string `json:"state"`
 	// +optional
 	FromReplicaAddress string `json:"fromReplicaAddress"`
+	// +optional
+	AppliedRebuildingMbps int64 `json:"appliedRebuildingMbps"`
```
derekbit: Is AppliedRebuildingMbytesPerSecond better?

controller/engine_controller.go
hookak: Ah, now I understand what you meant. Integrating the QoS info directly into `rebuildStatus` would indeed allow us to retrieve more accurate information. However, it looks like doing so would require a number of additional changes. It might be a good idea to revisit this in a future issue—especially when we work on applying QoS not only during rebuild but also more generally to bdevs.

k8s/pkg/apis/longhorn/v1beta2/engine.go
hookak: I also think `AppliedRebuildingMbytesPerSecond` is a clearer and more descriptive field name. I originally shortened it to `AppliedRebuildingMbps` because it felt too long compared to the other fields in `rebuildStatus`

k8s/pkg/apis/longhorn/v1beta2/volume.go
hookak: Sure. Clarified what 0 means.

hookak:
```
ec00b5a95 (HEAD -> LH10770_v2_set_qos, hookak/LH10770_v2_set_qos) <-- HEAD
8e37ec207 feat(rebuild)
```
I have a quick question. When addressing the review comments, should I update the changes directly in commit 8e37ec207, or is it acceptable to add new commits on top of ec00b5a95?

k8s/pkg/apis/longhorn/v1beta2/engine.go
derekbit: Agree with you. Then, to prevent confusion, how about rename it to `AppliedRebuildingMBps`. `Mbps` is for megabits per second.

hookak: I totally agree with you

k8s/pkg/apis/longhorn/v1beta2/volume.go
derekbit: > is it acceptable to add new commits on top of https://github.com/longhorn/longhorn-manager/commit/ec00b5a95365c114d036fba679326d82c6d0f065

This is acceptable. No worries. Thank you.

derekbit:
@hookak 
Could you help to resolve the conflicts? The PR is ready to merge. Thank you.

hookak:
Of course! ~~I've resolved the conflicts and updated the PR~~. Thanks!

derekbit:
Hello @hookak, build failed, can you help check it? Thanks.

hookak:
Could you review this one first?  https://github.com/longhorn/longhorn-instance-manager/pull/1020
The instance manager needs to update the field name as well due to the rename of `AppliedRebuildingMBps`.

derekbit:
> Could you review this one first? [longhorn/longhorn-instance-manager#1020](https://github.com/longhorn/longhorn-instance-manager/pull/1020) The instance manager needs to update the field name as well due to the rename of `AppliedRebuildingMBps`.

@hookak 
LGTM and merged. Thank you.

hookak:
I have resolved the conflicts and updated the PR. Please review it at your convenience. @derekbit

derekbit (approved):
LGTM. Thank you @hookak!

PR #3774, opened 2025-05-13
fix(orphan): prevent creating orphans from a terminating instance

COLDTURNIP:

Issue longhorn/longhorn#10888

While deleting the instance (engine or replica) CR, the CR may be removed before the instance state update event, because of 1 second delay in instance monitor. The instance monitor should not create an orphan CR for such terminating instances.


COLDTURNIP:
Verifying in [e2e regression](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/8535/)

controller/instance_manager_controller.go
```diff
@@ -2257,27 +2257,43 @@ func (m *InstanceManagerMonitor) isReplicaOrphaned(instanceName, node string) (b
...
-func (m *InstanceManagerMonitor) createOrphanForInstances(existOrphans map[string]bool, im *longhorn.InstanceManager, instanceMap instanceProcessMap, orphanType longhorn.OrphanType, orphanFilter func(instanceName, node string) (bool, error)) {
+func (m *InstanceManagerMonitor) createOrphanForInstances(existOrphans map[string]bool, im *longhorn.InstanceManager, instanceMap instanceProcessMap, orphanType longhorn.OrphanType, orphanFilter func(process *longhorn.InstanceProcess, instanceName, node string) (bool, error)) {
 	for instanceName, instance := range instanceMap {
+		switch instance.Status.State {
+		case longhorn.InstanceStateStarting:
+			// Status transitioning. Will handle this after running.
+			continue
+		case longhorn.InstanceStateStopping, longhorn.InstanceStateStopped:
+			// Terminating. No orphan CR needed, and the orphaned instances will be cleanup by instance manager after stopped.
+			continue
```
derekbit, line 2289: Should ignore `stopped`, isn't it?

COLDTURNIP: Yeah, so we list both `stopping` and `stopped` here.

COLDTURNIP:
Even with this change, there still a [very rare chance](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/8548/) to create a short-living orphan CR:

1. engine controller handles CR deletion event, ask instance manager to delete the process (blocking call)
2. process stopping -> stopped state change events generated by instance manager, queued into gRPC notification stream, and return the process deletion request.
3. engine controller receives the deletion response, finalize the engine CR
4. in instance controller, the monitor receives the process state change event from gRPC stream, checks the corresponding engine CR which was finalized before, then consider the process is orphaned because of the absence of engine CR.

controller/instance_manager_controller.go
```diff
@@ -1682,16 +1683,17 @@ func (imc *InstanceManagerController) deleteOrphans(im *longhorn.InstanceManager
 		}
 
 		var isDeletable = false
+		var isResourceDeletable = false
 		switch orphan.Spec.Type {
 		case longhorn.OrphanTypeEngineInstance:
-			isDeletable, err = imc.isOrphanDeletable(orphan, isInstanceManagerTerminating || autoDeleteEngine, im.Status.InstanceEngines, imc.isEngineScheduledOnCurrentNode)
+			isDeletable, isResourceDeletable, err = imc.isOrphanDeletable(orphan, isInstanceManagerTerminating || autoDeleteEngine, im.Status.InstanceEngines, imc.isEngineScheduledOnCurrentNode)
```
derekbit, line 1689:
```suggestion
			isOrphanDeletable, isResourceDeletable, err = imc.isOrphanDeletable(orphan, isInstanceManagerTerminating || autoDeleteEngine, im.Status.InstanceEngines, imc.isEngineScheduledOnCurrentNode)
```

controller/instance_manager_controller.go
```diff
@@ -1682,16 +1683,17 @@ func (imc *InstanceManagerController) deleteOrphans(im *longhorn.InstanceManager
...
 		case longhorn.OrphanTypeReplicaInstance:
-			isDeletable, err = imc.isOrphanDeletable(orphan, isInstanceManagerTerminating || autoDeleteReplica, im.Status.InstanceReplicas, imc.isReplicaScheduledOnCurrentNode)
+			isDeletable, isResourceDeletable, err = imc.isOrphanDeletable(orphan, isInstanceManagerTerminating || autoDeleteReplica, im.Status.InstanceReplicas, imc.isReplicaScheduledOnCurrentNode)
```
derekbit, line 1691:
```suggestion
			isOrphanDeletable, isResourceDeletable, err = imc.isOrphanDeletable(orphan, isInstanceManagerTerminating || autoDeleteReplica, im.Status.InstanceReplicas, imc.isReplicaScheduledOnCurrentNode)
```

controller/instance_manager_controller.go
```diff
@@ -1702,16 +1704,39 @@ func (imc *InstanceManagerController) deleteOrphans(im *longhorn.InstanceManager
...
+	return nil
+}
+
+func (imc *InstanceManagerController) isOrphanDeletable(orphan *longhorn.Orphan, autoDelete bool, instanceMap instanceProcessMap, isInstanceScheduledOnNode func(node, instance string) (bool, error)) (isDeletable, isResourceDeletable bool, err error) {
 	instanceName := orphan.Spec.Parameters[longhorn.OrphanInstanceName]
 	instanceProc, instanceExist := instanceMap[instanceName]
+	isResourceDeletable = instanceExist
+	if autoDelete {
+		return true, isResourceDeletable, nil
+	}
```
derekbit, line 1735: Why not continue the following checks for autoDelete?

controller/instance_manager_controller.go
```diff
@@ -2257,30 +2282,46 @@ func (m *InstanceManagerMonitor) isReplicaOrphaned(instanceName, node string) (b
 }
 
 // isInstanceOrphanedOnNode returns true only when it is very certain that an instance is scheduled on another node
-func (m *InstanceManagerMonitor) isInstanceOrphanedOnNode(spec *longhorn.InstanceSpec, status *longhorn.InstanceStatus, node string) bool {
-	if status.CurrentState != spec.DesireState {
+func (m *InstanceManagerMonitor) isInstanceOrphanedOnNode(crMeta *metav1.ObjectMeta, crSpec *longhorn.InstanceSpec, crStatus *longhorn.InstanceStatus, node string) bool {
```
derekbit, line 2285:
```suggestion
func (m *InstanceManagerMonitor) isInstanceOrphanedOnNode(meta *metav1.ObjectMeta, spec *longhorn.InstanceSpec, status *longhorn.InstanceStatus, node string) bool {
```

controller/instance_manager_controller.go
```diff
@@ -2231,7 +2256,7 @@ func (m *InstanceManagerMonitor) isEngineOrphaned(instanceName, node string) (bo
 	switch {
 	case err == nil:
 		// Engine CR still exists - check the ownership
-		return m.isInstanceOrphanedOnNode(&existEngine.Spec.InstanceSpec, &existEngine.Status.InstanceStatus, node), nil
+		return m.isInstanceOrphanedOnNode(&existEngine.ObjectMeta, &existEngine.Spec.InstanceSpec, &existEngine.Status.InstanceStatus, node), nil
```
derekbit, line 2259: Wht not directly pass engine to `isInstanceOrphanedOnNode`?

controller/instance_manager_controller.go
COLDTURNIP: `autoDelete` means that we can initiate the orphan CR deletion directly, without checking the other CRs. Will rename the method `isOrphanDeletable` to `isOrphanCrDeletable`.

controller/instance_manager_controller.go
COLDTURNIP: `isInstanceOrphanedOnNode` is shared between `v1beta2.Engine` and `v1beta2.Replica`. There's no common interface between these two to retrieve these three embedded structure.

controller/instance_manager_controller.go
```diff
@@ -1702,16 +1704,39 @@ func (imc *InstanceManagerController) deleteOrphans(im *longhorn.InstanceManager
...
+	switch {
+	case datastore.ErrorIsNotFound(err):
+		return nil
+	case err != nil:
+		return errors.Wrapf(err, "failed to delete orphan %v", orphan.Name)
+	}
+	return nil
+}
+
+func (imc *InstanceManagerController) isOrphanDeletable(orphan *longhorn.Orphan, autoDelete bool, instanceMap instanceProcessMap, isInstanceScheduledOnNode func(node, instance string) (bool, error)) (isDeletable, isResourceDeletable bool, err error) {
```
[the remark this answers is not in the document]
COLDTURNIP: Will rename `isDeletable` to `isCrDeletable`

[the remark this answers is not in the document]
derekbit: How about `isOrphanDeletable`?

controller/instance_manager_controller.go
derekbit: But why not check the other CRs?

derekbit:
Let me know if the PR is ready for review.

controller/instance_manager_controller.go
COLDTURNIP: The `autoDelete` parameter is enabled when:

1. instance manager is terminating, or
2. the `orphan-resource-auto-deletion` is enabled

In this two situation, we can initiate the orphan CR deletion directly, and orphan controller will take care about the resource clearning. After this flag check, it checks whether the corresponding instance is scheduled back to the node so that it is no longer considered to be orphaned.

derekbit (commented):
After rethinking the `Orphan Resource Auto-Deletion` setting, separating `engineInstance` and `replicaInstance` is not necessary. Can we merge them into `instance`?

controller/instance_manager_controller.go
COLDTURNIP: Got your point. the resource deletable should also consider the resource relocation even when auto delete is enabled.

controller/instance_manager_controller.go
```diff
@@ -1681,17 +1681,18 @@ func (imc *InstanceManagerController) deleteOrphans(im *longhorn.InstanceManager
...
-		var isDeletable = false
+		var isCrDeletable = false
+		var isResourceDeletable = false
 		switch orphan.Spec.Type {
 		case longhorn.OrphanTypeEngineInstance:
-			isDeletable, err = imc.isOrphanDeletable(orphan, isInstanceManagerTerminating || autoDeleteEngine, im.Status.InstanceEngines, imc.isEngineScheduledOnCurrentNode)
+			isCrDeletable, isResourceDeletable, err = imc.isOrphanCrDeletable(orphan, isInstanceManagerTerminating || autoDeleteEngine, im.Status.InstanceEngines, imc.isEngineScheduledOnCurrentNode)
 		case longhorn.OrphanTypeReplicaInstance:
-			isDeletable, err = imc.isOrphanDeletable(orphan, isInstanceManagerTerminating || autoDeleteReplica, im.Status.InstanceReplicas, imc.isReplicaScheduledOnCurrentNode)
+			isCrDeletable, isResourceDeletable, err = imc.isOrphanCrDeletable(orphan, isInstanceManagerTerminating || autoDeleteReplica, im.Status.InstanceReplicas, imc.isReplicaScheduledOnCurrentNode)
```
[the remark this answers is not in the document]
COLDTURNIP: The deletion of Orphan CR needs to be initiated unconditionally if instance manager is terminating, or the setting `orphan-resource-auto-deletion` is enabled. Will separate into two parameters.

controller/orphan_controller.go
```diff
@@ -261,15 +261,30 @@ func (oc *OrphanController) reconcile(orphanName string) (err error) {
...
+		defer func() {
+			if reflect.DeepEqual(existingOrphan.Status, orphan.Status) {
+				return
+			}
+			_, updateStatusErr := oc.ds.UpdateOrphanStatus(orphan)
+			if updateStatusErr != nil {
+				log.WithError(updateStatusErr).Errorf("Failed to update condition while cleaning up %v orphan %v", orphan.Spec.Type, orphan.Name)
+				if err == nil {
+					err = updateStatusErr
+				}
```
derekbit, line 276: Might not need to pass `updateStatusErr` to `err`. The condition update is best-effort. Logging the error is good enough. WDYT?

controller/orphan_controller.go
```diff
@@ -360,6 +376,10 @@ func (oc *OrphanController) cleanupOrphanedEngineInstance(orphan *longhorn.Orpha
 	if err != nil {
 		return false, err
 	}
+	if !orphan.Spec.ResourceDeletable {
+		oc.logger.Infof("Engine instance %v is not deletable. Skip cleaning up the instance resource and finalize the orphan CR.", instance)
```
derekbit, line 380:
```suggestion
		oc.logger.Infof("Engine instance %v is not deletable. Skipping cleaning up the instance resource and finalize the orphan CR.", instance)
```

controller/orphan_controller.go
```diff
@@ -385,6 +405,10 @@ func (oc *OrphanController) cleanupOrphanedReplicaInstance(orphan *longhorn.Orph
 	if err != nil {
 		return false, err
 	}
+	if !orphan.Spec.ResourceDeletable {
+		oc.logger.Infof("Replica instance %v is not deletable. Skip cleaning up the instance resource and finalize the orphan CR.", instance)
```
derekbit, line 409:
```suggestion
		oc.logger.Infof("Replica instance %v is not deletable. Skipping cleaning up the instance resource and finalize the orphan CR.", instance)
```

controller/instance_manager_controller.go
```diff
@@ -1662,13 +1662,11 @@ func (imc *InstanceManagerController) createInstanceManagerPodSpec(im *longhorn.
 
 func (imc *InstanceManagerController) deleteOrphans(im *longhorn.InstanceManager, isInstanceManagerTerminating bool) error {
 	autoDeletionTypes, err := imc.ds.GetSettingOrphanResourceAutoDeletion()
-	var autoDeleteEngine = false
-	var autoDeleteReplica = false
+	var autoDelete = false
```
derekbit, line 1665: Why not move `autoDelete` into `areOrphanAndResourceDeletable`?

k8s/pkg/apis/longhorn/v1beta2/orphan.go
```diff
@@ -47,6 +47,9 @@ type OrphanSpec struct {
 	// +optional
 	// +kubebuilder:validation:Enum=v1;v2
 	DataEngine DataEngineType `json:"dataEngine,omitempty"`
+	// Indicates whether the underlying resource can be deleted.
+	// +optional
+	ResourceDeletable bool `json:"resourceDeletable"`
```
c3y1huang, line 52: `ResourceDeletable` sounds like a state. Since this represents an action to be performed, consider renaming it to `DeleteResource` for clarity.

controller/instance_manager_controller.go
```diff
@@ -1702,34 +1701,52 @@ func (imc *InstanceManagerController) deleteOrphans(im *longhorn.InstanceManager
...
+	if orphan.Spec.ResourceDeletable != resourceDeletable {
+		orphan.Spec.ResourceDeletable = resourceDeletable
+		_, err := imc.ds.UpdateOrphan(orphan)
+		switch {
+		case datastore.ErrorIsNotFound(err):
+			return nil
+		case err != nil:
+			return errors.Wrapf(err, "failed to update resource deletable on orphan %v before deletion", orphan.Name)
+		}
 	}
```
c3y1huang, line 1714: Suggest moving the update logic into a separate method so the method name reflects its purpose.

controller/instance_manager_controller.go
COLDTURNIP: If there's some errors on the `orphan-resource-auto-deletion` setting, it will keep logging over the orphan iteration loop.

controller/instance_manager_controller.go
```diff
@@ -1702,34 +1701,52 @@ func (imc *InstanceManagerController) deleteOrphans(im *longhorn.InstanceManager
...
 
+	err := imc.ds.DeleteOrphan(orphan.Name)
+	switch {
+	case datastore.ErrorIsNotFound(err):
+		return nil
+	case err != nil:
+		return errors.Wrapf(err, "failed to delete orphan %v", orphan.Name)
+	}
+	return nil
+}
```
c3y1huang, line 1724: Not sure about deleting the orphan CR blindly. Maybe consider adding a condition such as `resourceDeleted` to orphan.State.Conditions to make the deletion more interpretable.

controller/instance_manager_controller.go
```diff
@@ -1702,34 +1701,52 @@ func (imc *InstanceManagerController) deleteOrphans(im *longhorn.InstanceManager
...
+
+func (imc *InstanceManagerController) areOrphanAndResourceDeletable(orphan *longhorn.Orphan, imTerminating, crAutoDelete bool, instanceMap instanceProcessMap, isInstanceScheduledOnNode func(node, instance string) (bool, error)) (isOrphanCRDeletable, isOrphanResourceDeletable bool, err error) {
```
c3y1huang, line 1726: Suggest separating this into two methods, `canDeleteOrphan` and `canDeleteOrphanedInstance`, to improve maintainability.

controller/instance_manager_controller.go
COLDTURNIP: This is what `spec.resourceDeletable` means, and another way is to reuse `DataCleanable` condition on data orphans. But, wouldn't it violate the convention to update status outside of orphan controller?

k8s/pkg/apis/longhorn/v1beta2/orphan.go
COLDTURNIP: Per discussion, this flag is no longer needed. Will be replaced in next PR  by instance process UUID (#3778)

controller/instance_manager_controller.go
COLDTURNIP: We'll keep blind cleanup with UUID in the next PR, and let instance manager protect the processes instead.

controller/instance_manager_controller.go
COLDTURNIP: `canDeleteOrphanedInstance` is removed.

controller/instance_manager_controller.go
```diff
@@ -1702,48 +1713,33 @@ func (imc *InstanceManagerController) deleteOrphans(im *longhorn.InstanceManager
...
-	}
-	return false, nil
+func (imc *InstanceManagerController) isOrphanCRDeletable(imTerminating, crAutoDelete, instanceExist, instanceCRScheduledBack bool) bool {
+	return imTerminating || crAutoDelete || !instanceExist || instanceCRScheduledBack
 }
 
-func (imc *InstanceManagerController) isEngineScheduledOnCurrentNode(node string, instance string) (bool, error) {
-	existEngine, err := imc.ds.GetEngineRO(instance)
+func (imc *InstanceManagerController) isEngineScheduledOnCurrentNode(node string, instance string) (isScheduledOnCurrentNode bool, err error) {
+	existEngine, errGetInstance := imc.ds.GetEngineRO(instance)
```
derekbit, line 1732: Why not use `err`?

controller/instance_manager_controller.go
```diff
@@ -1681,17 +1679,30 @@ func (imc *InstanceManagerController) deleteOrphans(im *longhorn.InstanceManager
...
-			if err := imc.ds.DeleteOrphan(orphan.Name); err != nil && !apierrors.IsNotFound(err) {
+			continue
+		}
+
+		imc.logger.WithFields(logrus.Fields{
+			"isInstanceManagerTerminating": isInstanceManagerTerminating,
+			"autoDelete":                   autoDelete,
+			"instanceExist":                instanceExist,
+			"instanceCRScheduledBack":      instanceCRScheduledBack,
+		}).Debugf("Checked deletable for %s orphan instance %s", orphan.Spec.Type, orphan.Name)
```
c3y1huang, line 1703: I don't think we need to log this.

controller/instance_manager_controller.go
c3y1huang, line 1665:
```suggestion
	var autoDeletionEnabled = false
```

derekbit (commented):
In general, LGTM. After resolving @c3y1huang's comments. The PR can be merged.

c3y1huang (commented):
Additionally, include the issue number (e.g., `longhorn/longhorn-10888`) in the commit messages for reference.

c3y1huang:
> Also, remember to include the issue number (i.e., `longhorn/longhorn-10888`) in the commit messages for reference.

Should also include it in a706bba8b4ccfea002da9e69bd2962c004c5a3de if it's relevant to the issue.

controller/instance_manager_controller.go
```diff
@@ -1662,13 +1662,11 @@ func (imc *InstanceManagerController) createInstanceManagerPodSpec(im *longhorn.
 
 func (imc *InstanceManagerController) deleteOrphans(im *longhorn.InstanceManager, isInstanceManagerTerminating bool) error {
 	autoDeletionTypes, err := imc.ds.GetSettingOrphanResourceAutoDeletion()
-	var autoDeleteEngine = false
-	var autoDeleteReplica = false
+	var autoDeleteEnabled = false
 	if err != nil {
 		imc.logger.WithError(err).Warnf("Failed to fetch orphan auto deletion setting, disabled by default")
```
c3y1huang, line 1667: Is there a reason we're not returning the error here?

controller/instance_manager_controller.go
```diff
@@ -1662,13 +1662,11 @@ func (imc *InstanceManagerController) createInstanceManagerPodSpec(im *longhorn.
...
 	} else {
-		autoDeleteEngine = autoDeletionTypes[types.OrphanResourceTypeEngineInstance]
-		autoDeleteReplica = autoDeletionTypes[types.OrphanResourceTypeReplicaInstance]
+		autoDeleteEnabled = autoDeletionTypes[types.OrphanResourceTypeInstance]
```
c3y1huang, line 1669: Would it be better to return directly if auto-delete is disabled?

controller/instance_manager_controller.go
```diff
@@ -2257,10 +2248,20 @@ func (m *InstanceManagerMonitor) isReplicaOrphaned(instanceName, node string) (b
 }
 
 // isInstanceOrphanedOnNode returns true only when it is very certain that an instance is scheduled on another node
-func (m *InstanceManagerMonitor) isInstanceOrphanedOnNode(spec *longhorn.InstanceSpec, status *longhorn.InstanceStatus, node string) bool {
+func (m *InstanceManagerMonitor) isInstanceOrphanedOnNode(meta *metav1.ObjectMeta, spec *longhorn.InstanceSpec, status *longhorn.InstanceStatus, node string) bool {
+	if !meta.DeletionTimestamp.IsZero() {
+		m.logger.Debugf("Instance %s in deletion, judge orphan later", meta.Name)
```
c3y1huang, line 2253:
```suggestion
		m.logger.Debugf("Skipping orphan check; Instance %s is marked for deletion", meta.Name)
```

controller/instance_manager_controller.go
```diff
@@ -2257,10 +2248,20 @@ func (m *InstanceManagerMonitor) isReplicaOrphaned(instanceName, node string) (b
...
+		return false
+	}
+
 	if status.CurrentState != spec.DesireState {
+		m.logger.WithFields(logrus.Fields{
+			"currentState": status.CurrentState,
+			"desiredState": spec.DesireState,
+		}).Debugf("Instance %s in state transition, judge orphan later", meta.Name)
```
c3y1huang, line 2261:
```suggestion
		}).Debugf("Skipping orphan check; Instance %s is in state transition", meta.Name)
```

c3y1huang:
@coderabbitai review

controller/instance_manager_controller.go
COLDTURNIP: No matter the auto deletion is configured, there's still some situation that orphan CR deletion should be initiated:

- instance manager is terminating or terminated
- instance disappears from the instance manager
- the instance is scheduled back to the node, and is no longer considered as orphaned

controller/instance_manager_controller.go
COLDTURNIP: (see the reason at https://github.com/longhorn/longhorn-manager/pull/3774#discussion_r2092188262)

types/setting.go
```diff
@@ -1605,9 +1604,8 @@
 type OrphanResourceType string
 
 const (
-	OrphanResourceTypeReplicaData     = OrphanResourceType("replicaData")
-	OrphanResourceTypeEngineInstance  = OrphanResourceType("engineInstance")
-	OrphanResourceTypeReplicaInstance = OrphanResourceType("replicaInstance")
+	OrphanResourceTypeReplicaData = OrphanResourceType("replicaData")
+	OrphanResourceTypeInstance    = OrphanResourceType("instance")
 )
```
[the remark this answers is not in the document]
COLDTURNIP: `engineInstance` and `replicaInstance` are newly introduced in the coming v1.9.0. There's no compatibility issue to change the types before the release.

types/setting.go
```diff
@@ -1791,9 +1789,8 @@
 
 func UnmarshalOrphanResourceTypes(resourceTypesSetting string) (map[OrphanResourceType]bool, error) {
 	resourceTypes := map[OrphanResourceType]bool{
-		OrphanResourceTypeReplicaData:     false,
-		OrphanResourceTypeEngineInstance:  false,
-		OrphanResourceTypeReplicaInstance: false,
+		OrphanResourceTypeReplicaData: false,
+		OrphanResourceTypeInstance:    false,
 	}
```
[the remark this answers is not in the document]
COLDTURNIP: `engineInstance` and `replicaInstance` are newly introduced in the coming v1.9.0. There's no compatibility issue to change the types before the release.

controller/orphan_controller.go
```diff
@@ -575,6 +588,8 @@
 
 	instance, err := imc.InstanceGet(orphan.Spec.DataEngine, instanceName, string(instanceType))
 	switch {
+	case err != nil && types.ErrorIsNotFound(err):
+		instanceState = longhorn.InstanceStateTerminated
 	case err != nil:
```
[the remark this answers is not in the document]
COLDTURNIP: `imc.InstanceGet` fetch the instance information from the instance manager server, not the cached K8s resource in datastore. It is not reasonable to check the error with `datastore.ErrorIsNotFound`

controller/instance_manager_controller.go
derekbit: Could you add a comment here?

controller/instance_manager_controller.go
```diff
@@ -2273,14 +2274,23 @@
...
 func (m *InstanceManagerMonitor) createOrphanForInstances(existOrphans map[string]bool, im *longhorn.InstanceManager, instanceMap instanceProcessMap, orphanType longhorn.OrphanType, orphanFilter func(instanceName, node string) (bool, error)) {
 	for instanceName, instance := range instanceMap {
+		switch instance.Status.State {
+		case longhorn.InstanceStateStarting:
+			// Status transitioning. Will handle this after running.
+			continue
+		case longhorn.InstanceStateStopping, longhorn.InstanceStateStopped:
+			// Terminating. No orphan CR needed, and the orphaned instances will be cleanup by instance manager after stopped.
+			continue
+		}
```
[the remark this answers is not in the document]
COLDTURNIP: The `error` state of an instance is expected to create the corresponding orphan CR as a handle. This is a stable state inside the instance manager.

controller/orphan_controller.go
```diff
@@ -575,6 +588,8 @@ func (oc *OrphanController) updateInstanceStateCondition(orphan *longhorn.Orphan
 
 	instance, err := imc.InstanceGet(orphan.Spec.DataEngine, instanceName, string(instanceType))
 	switch {
+	case err != nil && types.ErrorIsNotFound(err):
+		instanceState = longhorn.InstanceStateTerminated
```
derekbit, line 592: The switch here is not easy to read. Suggest
```
instance, err := imc.InstanceGet(orphan.Spec.DataEngine, instanceName, string(instanceType))
if err != nil {
	if types.ErrorIsNotFound(err) {
		instanceState = longhorn.InstanceStateTerminated
	} else {
		return errors.Wrapf(err, "failed to get instance %v", instanceName)
	}
} else if instance == nil {
	instanceState = longhorn.InstanceStateTerminated
} else {
	instanceState = instance.Status.State
}
```

controller/orphan_controller.go
```diff
@@ -261,15 +261,27 @@ func (oc *OrphanController) reconcile(orphanName string) (err error) {
...
+			}
+
+			if _, updateStatusErr := oc.ds.UpdateOrphanStatus(orphan); updateStatusErr != nil {
+				log.WithError(updateStatusErr).Errorf("Failed to update condition while cleaning up %v orphan %v", orphan.Spec.Type, orphan.Name)
+			}
+		}()
+
 		isCleanupComplete, err := oc.cleanupOrphanedResource(orphan)
 		if isCleanupComplete {
+			oc.eventRecorder.Eventf(orphan, corev1.EventTypeNormal, "OrphanCleanupComplete", "Orphan %v cleanup completed", orphan.Name)
```
derekbit, line 279: Should define the event in `constant/events.go`.

EventReasonOrphanCleanupCompleted = "OrphanCleanupCompleted"

```suggestion
			oc.eventRecorder.Eventf(orphan, corev1.EventTypeNormal, EventReasonOrphanCleanupCompleted, "Orphan %v cleanup completed", orphan.Name)
```

controller/instance_manager_controller.go
COLDTURNIP: After reviewing again over this function, we should handle the list of orphan CRs belongs to the given instance, not the node of this instance. Will refine this part.

derekbit: > After reviewing again over this function, we should handle the list of orphan CRs belongs to the given instance, not the node of this instance. Will refine this part.

Can you elaborate more on what the handling means?

controller/instance_manager_controller.go
```diff
@@ -1702,34 +1707,20 @@ func (imc *InstanceManagerController) deleteOrphans(im *longhorn.InstanceManager
...
-		instanceState = longhorn.InstanceStateTerminated
+func (imc *InstanceManagerController) deleteOrphan(orphan *longhorn.Orphan) error {
+	imc.logger.Infof("Deleting Orphan %q", orphan.Name)
+	err := imc.ds.DeleteOrphan(orphan.Name)
+	switch {
+	case datastore.ErrorIsNotFound(err):
+		return nil
+	case err != nil:
+		return errors.Wrapf(err, "failed to delete orphan %v", orphan.Name)
 	}
```
c3y1huang, line 1718: Use if statement for readability.
```suggestion
	if err := imc.ds.DeleteOrphan(orphan.Name); err != nil {
		if datastore.ErrorIsNotFound(err) {
			return nil
		}
		return errors.Wrapf(err, "failed to delete orphan %q", orphan.Name)
	}
```

controller/orphan_controller.go
```diff
@@ -261,15 +261,27 @@ func (oc *OrphanController) reconcile(orphanName string) (err error) {
...
 
+	existingOrphan := orphan.DeepCopy()
+
 	if !orphan.DeletionTimestamp.IsZero() {
+		defer func() {
+			if reflect.DeepEqual(existingOrphan.Status, orphan.Status) {
+				return
+			}
+
+			if _, updateStatusErr := oc.ds.UpdateOrphanStatus(orphan); updateStatusErr != nil {
```
c3y1huang, line 272: What happens if the update is conflicted?

COLDTURNIP: The status update will be dropped. The cleanup errors will be logged but won't be not recorded in the conditions, and there is no other status change during the finalization. WDYT @derekbit ?

derekbit: After rethinking the logic, we can return `err` and retry. If it always fails to update the status, it means the API server is problematic. WDYT @c3y1huang @COLDTURNIP

c3y1huang: Sounds good to me.

COLDTURNIP: Agree.

COLDTURNIP:
Per rethinking the design (https://github.com/longhorn/longhorn-manager/pull/3774#discussion_r2092209944), we change to compare the instance with the instance manager, instead of the node, since the instances are belong to instance managers, and it is possible to have multiple instance manager on a node. Please help review, thank. @derekbit @c3y1huang

controller/instance_manager_controller.go
```diff
@@ -2256,31 +2267,53 @@ func (m *InstanceManagerMonitor) isReplicaOrphaned(instanceName, node string) (b
...
-		orphanName := types.GetOrphanChecksumNameForOrphanedInstance(instanceName, m.controllerID, string(instance.Spec.DataEngine))
+		switch instance.Status.State {
+		case longhorn.InstanceStateStarting:
+			// Status transitioning. Will handle this after running.
+			continue
+		case longhorn.InstanceStateStopping, longhorn.InstanceStateStopped:
+			// Terminating. No orphan CR needed, and the orphaned instances will be cleanup by instance manager after stopped.
+			continue
+		}
+
```
derekbit, line 2308:
```suggestion
        if instance.Status.State == longhorn.InstanceStateStarting ||
           instance.Status.State == longhorn.InstanceStateStopping ||
           instance.Status.State == longhorn.InstanceStateStopped {
			// Starting: Status transitioning. Will handle this after running.
			// Stoppin, Stopped: Terminating. No orphan CR needed, and the orphaned instances will be cleanup by instance manager after stopped.
			continue
		}
```

COLDTURNIP:
> Per rethinking the design ([#3774 (comment)](https://github.com/longhorn/longhorn-manager/pull/3774#discussion_r2092209944)), we change to compare the instance with the instance manager, instead of the node, since the instances are belong to instance managers, and it is possible to have multiple instance manager on a node. Please help review, thank. @derekbit @c3y1huang

More detail about the last change. The main changes are located in instance manager controller. There are two methods:
- `InstanceManagerMonitor.isInstanceOrphanedInInstanceManager`: this method is to check if an instance is really orphaned before creating the orphan CR. While corresponding engine/replica CR enters a stable state (running on desired node), then we can ensure it is orphaned if the engine/replica CR's instance manager is NOT the one that the instance presents. (Before the change: compare the node ID instead)
- `InstanceManagerController.isInstanceScheduledToCurrentInstanceManager`: similar to the above but reversed. this method is to check if an instance is really NOT orphaned and need to initiate the orphan CR deletion. Compare the instance manager to check if the engine/replica is scheduled back.
- Also, to identify the instances between instance managers, we calculate the orphan CR name from instance name, instance manager, data engine type, instead of the node. See `types.GetOrphanChecksumNameForOrphanedInstance`

controller/orphan_controller.go
```diff
@@ -386,16 +403,16 @@ func (oc *OrphanController) cleanupOrphanedReplicaInstance(orphan *longhorn.Orph
...
 			return false, err
 		}
-		spec = nil
+		status = nil
 	} else {
-		spec = &replicaCR.Spec.InstanceSpec
+		status = &replicaCR.Status.InstanceStatus
 	}
-	oc.cleanupOrphanedInstance(orphan, instance, imName, longhorn.InstanceManagerTypeReplica, spec)
+	oc.cleanupOrphanedInstance(orphan, instance, imName, longhorn.InstanceManagerTypeReplica, status)
```
derekbit, line 415: What's the purpose of the change?

COLDTURNIP: Inside `cleanupOrphanedInstance`, we change the instance ownership check logic, compares the `Status.InstanceManagerName` instead of `Spec.NodeID`.

COLDTURNIP: When the `instance.Status.InstanceManagerName` is the same with the one inside orphan CR, it means the instance is scheduled back to the instance manager, the instance process is no longer considered orphaned, and the engine/replica controller will manage this process. The orphan controller should finalize the orphan CR without touching the instance process.

controller/instance_manager_controller.go
```diff
@@ -1702,76 +1715,74 @@ func (imc *InstanceManagerController) deleteOrphans(im *longhorn.InstanceManager
...
-	if !instanceExist || instanceCRScheduledBack {
-		return true, nil
-	}
-	return false, nil
+func (imc *InstanceManagerController) canDeleteOrphanCR(imTerminating, autoDeleteEnabled, instanceExist, instanceCRScheduledBack bool) bool {
+	return imTerminating || autoDeleteEnabled || !instanceExist || instanceCRScheduledBack
 }
 
-func (imc *InstanceManagerController) isEngineScheduledOnCurrentNode(node string, instance string) (bool, error) {
+func (imc *InstanceManagerController) isEngineScheduledInCurrentInstanceManager(instanceManager string, instance string) (bool, error) {
```
c3y1huang, line 1733:
```suggestion
func (imc *InstanceManagerController) isEngineOnInstanceManager(instanceManager string, instance string) (bool, error) {
```

derekbit (dismissed):

controller/instance_manager_controller.go
```diff
@@ -1702,76 +1715,74 @@ func (imc *InstanceManagerController) deleteOrphans(im *longhorn.InstanceManager
...
 		return false, nil
 	default:
 		// Unexpected error - unable to check if engine instance is orphaned or not
-		return false, err
+		return false, errors.Wrapf(err, "failed to check CR scheduled node for engine orphan %v", instance)
 	}
 }
 
-func (imc *InstanceManagerController) isReplicaScheduledOnCurrentNode(node string, instance string) (bool, error) {
+func (imc *InstanceManagerController) isReplicaScheduledInCurrentInstanceManager(instanceManager string, instance string) (bool, error) {
```
c3y1huang, line 1748:
```suggestion
func (imc *InstanceManagerController) isReplicaOnInstanceManager(instanceManager string, instance string) (bool, error) {
```

c3y1huang, line 1748: Improve reability and fix log. Suggestion also applies for `isEngineScheduledInCurrentInstanceManager`.

```suggestion
	existReplica, err := imc.ds.GetReplicaRO(instance)
	if err != nil {
		if apierrors.IsNotFound(err) {
			// Replica CR not found - instance is orphaned
			return false, nil
		}
		return false, errors.Wrapf(err, "failed to check if instance %q is scheduled on instance manager %q", instance, instanceManager)
	}

	return imc.isInstanceOnInstanceManager(instanceManager, &existReplica.ObjectMeta, &existReplica.Spec.InstanceSpec, &existReplica.Status.InstanceStatus), nil

```

controller/instance_manager_controller.go
```diff
@@ -1702,76 +1715,74 @@ func (imc *InstanceManagerController) deleteOrphans(im *longhorn.InstanceManager
...
-		return imc.isInstanceScheduledOnCurrentNode(node, &existReplica.Spec.InstanceSpec, &existReplica.Status.InstanceStatus), nil
+		return imc.isInstanceOnInstanceManager(instanceManager, &existReplica.ObjectMeta, &existReplica.Spec.InstanceSpec, &existReplica.Status.InstanceStatus), nil
 	case apierrors.IsNotFound(err):
 		// Replica CR not found - instance is orphaned
 		return false, nil
 	default:
 		// Unexpected error - unable to check if replica instance is orphaned or not
-		return false, err
+		return false, errors.Wrapf(err, "failed to check CR scheduled node for replica orphan %v", instance)
 	}
```
c3y1huang, line 1760: Improve readablity and fix log. Suggestion also applies for `isEngineScheduledInCurrentInstanceManager`.

```suggestion
	existReplica, err := imc.ds.GetReplicaRO(instance)
	if err != nil {
		if apierrors.IsNotFound(err) {
			// Replica CR not found - instance is orphaned
			return false, nil
		}
		return false, errors.Wrapf(err, "failed to check if instance %q is scheduled on instance manager %q", instance, instanceManager)
	}

	return imc.isInstanceOnInstanceManager(instanceManager, &existReplica.ObjectMeta, &existReplica.Spec.InstanceSpec, &existReplica.Status.InstanceStatus), nil
```

controller/orphan_controller.go
```diff
@@ -433,7 +450,7 @@ func (oc *OrphanController) cleanupOrphanedInstance(orphan *longhorn.Orphan, ins
 		oc.logger.WithError(err).Warnf("Failed to delete orphan instance %v due to instance manager client initialization failure. Continue to finalize orphan %v", instance, orphan.Name)
 		return
 	} else if imc == nil {
-		oc.logger.WithField("orphanInstanceNode", orphan.Spec.NodeID).Warnf("No running instance manager for deleting orphan instance %v", orphan.Name)
+		oc.logger.WithField("orphanInstanceManager", imName).Warnf("No running instance manager for deleting orphan instance %v", orphan.Name)
```
derekbit, line 453: nit:
```suggestion
		oc.logger.WithField("instanceManager", imName).Warnf("No running instance manager for deleting orphan instance %v", orphan.Name)
```

c3y1huang (approved):

derekbit (dismissed):

derekbit:
@mergify backport v1.9.x

types/setting.go
```diff
@@ -1064,8 +1064,7 @@ var (
 			"List the enabled resource types in a semicolon-separated list. \n\n" +
 			"Available items are: \n\n" +
 			"- **replicaData**: replica data store \n\n" +
```
c3y1huang, line 1066: BTW, IIRC Longhorn doesn't use camel casing for the setting values, we should fix the `replicaData` for consistency.

derekbit: which one is better "replica-data", "replica-data" or "replica data"?

COLDTURNIP: Thanks. Will change to `replica-data`.

derekbit (approved):

c3y1huang (approved):

derekbit:
@mergify backport v1.9.x

PR #3778, opened 2025-05-14
feat: record UUID of engine and replica instance

derekbit:

Issue Longhorn/longhorn#10888


controller/orphan_controller.go
```diff
@@ -412,32 +412,37 @@ func (oc *OrphanController) cleanupOrphanedReplicaInstance(orphan *longhorn.Orph
...
 	default:
-		return "", "", fmt.Errorf("unknown data engine type %v for instance orphan %v", orphan.Spec.DataEngine, orphan.Name)
+		return "", "", "", fmt.Errorf("unknown data engine type %v for instance orphan %v", orphan.Spec.DataEngine, orphan.Name)
 	}
 
-	return name, instanceManager, nil
+	uuid, exist := orphan.Spec.Parameters[longhorn.OrphanInstanceUUID]
+	if !exist || uuid == "" {
+		return "", "", "", fmt.Errorf("failed to get instance UUID for instance orphan %v", orphan.Name)
+	}
```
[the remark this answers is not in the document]
COLDTURNIP: Agree. When UUID is unavailable, the orphan CR is illegal, and we should remove it instead triggering a retry.

controller/instance_manager_controller.go
```diff
@@ -2316,16 +2316,27 @@ func (m *InstanceManagerMonitor) createOrphanForInstances(existOrphans map[strin
 			// Stopping, Stopped: Terminating. No orphan CR needed, and the orphaned instances will be cleanup by instance manager after stopped.
 			continue
 		}
+		if instance.Spec.DataEngine != longhorn.DataEngineTypeV1 {
+			m.logger.Debugf("Skipping orphan check; Instance %s is not data engine v1", instanceName)
```
derekbit, line 2320:
```suggestion
			m.logger.Debugf("Skipping orphan creation, instance %s is not data engine v1", instanceName)
```

controller/orphan_controller.go
```diff
@@ -398,46 +399,52 @@ func (oc *OrphanController) cleanupOrphanedReplicaInstance(orphan *longhorn.Orph
...
+		return "", "", "", fmt.Errorf("failed to get instance manager for instance orphan %v", orphan.Name)
 	}
 
 	switch orphan.Spec.DataEngine {
 	case longhorn.DataEngineTypeV1, longhorn.DataEngineTypeV2:
 		// supported data engine type
 	default:
-		return "", "", fmt.Errorf("unknown data engine type %v for instance orphan %v", orphan.Spec.DataEngine, orphan.Name)
+		return "", "", "", fmt.Errorf("unknown data engine type %v for instance orphan %v", orphan.Spec.DataEngine, orphan.Name)
+	}
```
c3y1huang, line 437: Seems redundant, as it should have already been validated by the webhook.

COLDTURNIP: Could we assume all the content form a CR is valid? Indeed there's no chance to modify the orphan CR spec in our current design.

c3y1huang: > Could we assume all the content form a CR is valid? Indeed there's no chance to modify the orphan CR spec in our current design.

Could we assume `orphan.Spec.DataEngine` is valid?

derekbit: Removing it should be fine.
Can you add a validating logic in the webhook to make the field immutable?

COLDTURNIP: Immutable would be great. Let me include this.

derekbit:
Does the orphan resource name take the UUID into account?

version
```diff
@@ -1 +1 @@
-v1.10.0-dev
+v1.9.0-dev
```
derekbit, line 1: Why change the version?

COLDTURNIP: Sorry for mistake. Reverting it.

derekbit:
@mergify backport v1.9.x

COLDTURNIP:
> Does the orphan resource name take the UUID into account?

Yes: https://github.com/derekbit/longhorn-manager/blob/issue-10888/controller/instance_manager_controller.go#L2328

derekbit (commented):
In general, LGTM.

The test cases need to test fresh installation and upgrade to check the compatibility.

derekbit:
@coderabbitai review

controller/orphan_controller.go
```diff
@@ -398,56 +401,40 @@ func (oc *OrphanController) cleanupOrphanedReplicaInstance(orphan *longhorn.Orph
...
-func (oc *OrphanController) extractOrphanedInstanceInfo(orphan *longhorn.Orphan) (name, instanceManager string, err error) {
-	name, ok := orphan.Spec.Parameters[longhorn.OrphanInstanceName]
-	if !ok {
-		return "", "", fmt.Errorf("failed to get instance name for instance orphan %v", orphan.Name)
-	}
-
-	instanceManager, ok = orphan.Spec.Parameters[longhorn.OrphanInstanceManager]
-	if !ok {
-		return "", "", fmt.Errorf("failed to get instance manager for instance orphan %v", orphan.Name)
+func (oc *OrphanController) extractOrphanedInstanceInfo(orphan *longhorn.Orphan) orphanedInstanceParameters {
```
c3y1huang, line 419:
```suggestion
func (oc *OrphanController) getOrphanedInstanceParameters(orphan *longhorn.Orphan) orphanedInstanceParameters {
```

controller/orphan_controller.go
```diff
@@ -676,3 +655,11 @@ func (oc *OrphanController) checkOrphanedReplicaDataCleanable(node *longhorn.Nod
 
 	return ""
 }
+
+func getOrphanedInstanceParameters(orphan *longhorn.Orphan) orphanedInstanceParameters {
```
c3y1huang, line 659: NIT: Move `orphanedInstanceParameters struct` above this function.

c3y1huang (approved):

controller/orphan_controller.go
```diff
@@ -373,21 +373,18 @@ func (oc *OrphanController) cleanupOrphanedEngineInstance(orphan *longhorn.Orpha
...
 		if !datastore.ErrorIsNotFound(err) {
 			return false, err
 		}
 		status = nil
 	} else {
-		status = &engineCR.Status.InstanceStatus
+		status = &engine.Status.InstanceStatus
 	}
-	oc.cleanupOrphanedInstance(orphan, instance, imName, longhorn.InstanceManagerTypeEngine, status)
+	oc.cleanupOrphanedInstance(orphan, instanceParameters.name, instanceParameters.uuid, instanceParameters.instanceManager, longhorn.InstanceManagerTypeEngine, status)
```
c3y1huang, line 387:
```suggestion
	oc.cleanupOrphanedInstance(orphan, instanceParameters, longhorn.InstanceManagerTypeEngine, status)
```

c3y1huang (approved):

PR #3790, opened 2025-05-19
feat(orphan): introduce orphan-resource-auto-deletion-grace-period setting

derekbit:

Issue Longhorn/longhorn#10904

1. switch-case works better for conditions based on a single number or string, whereas if-else is more suitable for decisions involving multiple variables.
2. In the current implementation, with auto orphan deletion enabled, the orphan CR and its associated process are deleted almost immediately after the orphan CR is created. This behavior might be too abrupt. We could introduce a configurable setting, such as a wait interval (deletion grace period), to delay this automatic deletion. For instance, if the interval is set to 60 seconds, the orphan CR and its process would be deleted 60 seconds after the creation timestamp.
Manual deletion of the orphan CR by the user would bypass this delay and proceed immediately.

### DIFF
--- a/controller/instance_manager_controller.go
+++ b/controller/instance_manager_controller.go
@@ -1669,11 +1669,17 @@ func (imc *InstanceManagerController) createInstanceManagerPodSpec(im *longhorn.
 //   - Automatic deletion of orphan resources is enabled.
 func (imc *InstanceManagerController) deleteOrphans(im *longhorn.InstanceManager, isInstanceManagerTerminating bool) error {
 	autoDeletionTypes, err := imc.ds.GetSettingOrphanResourceAutoDeletion()
-	var autoDeleteEnabled = false
 	if err != nil {
-		imc.logger.WithError(err).Warnf("Failed to fetch orphan auto deletion setting, disabled by default")
-	} else {
-		autoDeleteEnabled = autoDeletionTypes[types.OrphanResourceTypeInstance]
+		return errors.Wrapf(err, "failed to get setting %v", types.SettingNameOrphanResourceAutoDeletion)
+	}
+	autoDeleteEnabled, ok := autoDeletionTypes[types.OrphanResourceTypeInstance]
+	if !ok {
+		autoDeleteEnabled = false
+	}
+
+	autoDeleteGracePeriod, err := imc.ds.GetSettingAsInt(types.SettingNameOrphanResourceAutoDeletionGracePeriod)
+	if err != nil {
+		return errors.Wrapf(err, "failed to get setting %v", types.SettingNameOrphanResourceAutoDeletionGracePeriod)
 	}
 
 	orphanList, err := imc.ds.ListInstanceOrphansByInstanceManagerRO(im.Name)
@@ -1703,20 +1709,20 @@ func (imc *InstanceManagerController) deleteOrphans(im *longhorn.InstanceManager
 			continue
 		}
 
-		if imc.canDeleteOrphanCR(isInstanceManagerTerminating, autoDeleteEnabled, instanceExist, instanceCRScheduledBack) {
+		if imc.canDeleteOrphan(orphan, isInstanceManagerTerminating, autoDeleteEnabled, instanceExist, instanceCRScheduledBack, autoDeleteGracePeriod) {
 			if err := imc.deleteOrphan(orphan); err != nil {
 				multiError.Append(util.NewMultiError(fmt.Sprintf("%v: %v", orphan.Name, err)))
 			}
 		}
 	}
 	if len(multiError) > 0 {
-		return fmt.Errorf("instance manager failed to delete orphan CR: %v", multiError.Join())
+		return fmt.Errorf("failed to delete orphans: %v", multiError.Join())
 	}
 	return nil
 }
 
 func (imc *InstanceManagerController) deleteOrphan(orphan *longhorn.Orphan) error {
-	imc.logger.Infof("Deleting Orphan %q", orphan.Name)
+	imc.logger.Infof("Deleting Orphan %v", orphan.Name)
 	if err := imc.ds.DeleteOrphan(orphan.Name); err != nil {
 		if datastore.ErrorIsNotFound(err) {
 			return nil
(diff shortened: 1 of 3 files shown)


controller/instance_manager_controller.go
```diff
@@ -1726,8 +1732,19 @@ func (imc *InstanceManagerController) deleteOrphan(orphan *longhorn.Orphan) erro
 	return nil
 }
 
-func (imc *InstanceManagerController) canDeleteOrphanCR(imTerminating, autoDeleteEnabled, instanceExist, instanceCRScheduledBack bool) bool {
-	return imTerminating || autoDeleteEnabled || !instanceExist || instanceCRScheduledBack
+func (imc *InstanceManagerController) canDeleteOrphan(orphan *longhorn.Orphan, imTerminating, autoDeleteEnabled, instanceExist, instanceCRScheduledBack bool, autoDeleteGracePeriod int64) bool {
+	autoDeleteAllowed := false
+	if autoDeleteEnabled {
+		elapsedTime := time.Since(orphan.CreationTimestamp.Time)
+		if elapsedTime > gracePeriod {
```
c3y1huang, line 1739: `gracePeriod` is uninitialized.

controller/instance_manager_controller.go
```diff
@@ -1726,8 +1732,19 @@ func (imc *InstanceManagerController) deleteOrphan(orphan *longhorn.Orphan) erro
...
+			autoDeleteAllowed = true
+		} else {
+			remainingTime := gracePeriod - elapsedTime
+			imc.logger.Debugf("Orphan %v is created %v seconds ago, will be deleted after %v seconds", orphan.Name, elapsedTime.Seconds(), remainingTime.Seconds())
```
c3y1huang, line 1743: The log message could be misleading since all conditions are evaluated at the end of the function. Consider evaluating and logging the input conditions earlier for clarity.

types/setting.go
```diff
@@ -1072,6 +1075,20 @@ var (
 		Default:  "",
 	}
 
+	SettingDefinitionOrphanResourceAutoDeletionGracePeriod = SettingDefinition{
+		DisplayName: "Orphan Resource Auto-Deletion Grace Period",
+		Description: "In seconds. The interval determines how long Longhorn will wait before automatically deleting an orphan CR and its associated resource. \n\n" +
+			"**Note:** Manual deletion of an orphan CR by an user will bypass this delay and proceed immediately. \n\n",
```
c3y1huang, line 1081: This is user-facing content, cc @sushant-suse for review

types/setting.go
```diff
@@ -1072,6 +1075,20 @@ var (
 		Default:  "",
 	}
 
+	SettingDefinitionOrphanResourceAutoDeletionGracePeriod = SettingDefinition{
```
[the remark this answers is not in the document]
COLDTURNIP: IMHO only the instance orphans need the grace period because of changeable state. Data keeps orphaned after judgement. WDYT @derekbit

controller/instance_manager_controller.go
COLDTURNIP: Should be `autoDeleteGracePeriod`

types/setting.go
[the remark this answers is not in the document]
derekbit: I'm fine with applying the grace period to all resources to maintain the consistency. WDYT? @c3y1huang @COLDTURNIP

controller/instance_manager_controller.go
derekbit: Fixed. Thank you.

controller/instance_manager_controller.go
derekbit: Updated. Thanks.

types/setting.go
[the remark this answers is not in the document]
COLDTURNIP: The consistency between different resource types is acceptable to me.

controller/instance_manager_controller.go
```diff
@@ -1669,11 +1669,17 @@ func (imc *InstanceManagerController) createInstanceManagerPodSpec(im *longhorn.
...
+	}
+	autoDeleteEnabled, ok := autoDeletionTypes[types.OrphanResourceTypeInstance]
+	if !ok {
+		autoDeleteEnabled = false
+	}
+
+	autoDeleteGracePeriod, err := imc.ds.GetSettingAsInt(types.SettingNameOrphanResourceAutoDeletionGracePeriod)
+	if err != nil {
+		return errors.Wrapf(err, "failed to get setting %v", types.SettingNameOrphanResourceAutoDeletionGracePeriod)
 	}
```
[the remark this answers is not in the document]
derekbit: Validating webhook will reject a negative number.

controller/node_controller.go
```diff
@@ -1420,6 +1419,32 @@ func (nc *NodeController) deleteOrphansForReplicaDataStore(node *longhorn.Node,
...
+
+	autoDeleteAllowed := false
+	if autoDeleteEnabled {
+		elapsedTime := time.Since(orphan.CreationTimestamp.Time).Seconds()
+		if elapsedTime > float64(autoDeleteGracePeriod) {
+			autoDeleteAllowed = true
+		}
+	}
+
+	canDelete := autoDeleteAllowed || dataCleanableCondition.Status == longhorn.ConditionStatusFalse
```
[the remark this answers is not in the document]
derekbit: Added a comment.

c3y1huang (approved):

derekbit:
@mergify backport v1.9.x

types/setting.go
sushant-suse, line 1081:
```suggestion
		Description: "Specifies the wait time, in seconds, before Longhorn automatically deletes an orphaned Custom Resource (CR) and its associated resources. \n\n" +
			"**Note:** If a user manually deletes an orphaned CR, the deletion occurs immediately and does not respect this grace period. \n\n",
```

sushant-suse (changes requested):
Thanks.

COLDTURNIP (approved):
LGTM

sushant-suse (approved):

derekbit:
@mergify backport v1.9.x

PR #3805, opened 2025-05-29
feat: move validation and mutation logics to the admission webhook

derekbit:

Issue Longhorn/longhorn#5156

### DIFF
--- a/webhook/resources/backupbackingimage/mutator.go
+++ b/webhook/resources/backupbackingimage/mutator.go
@@ -56,8 +56,15 @@ func (b *backupBackingImageMutator) Create(request *admission.Request, newObj ru
 	}
 	patchOps = append(patchOps, mutatePatchOps...)
 
-	backupTarget, err := b.ds.GetBackupTargetRO(backupBackingImage.Spec.BackupTargetName)
-	if apierrors.IsNotFound(err) {
+	backupTargetName := backupBackingImage.Spec.BackupTargetName
+	if backupTargetName == "" {
+		backupTargetName = types.DefaultBackupTargetName
+	}
+	patchOps = append(patchOps, fmt.Sprintf(`{"op": "replace", "path": "/spec/backupTargetName", "value": "%s"}`, backupTargetName))
+
+	backupTarget, err := b.ds.GetBackupTargetRO(backupTargetName)
+	if apierrors.IsNotFound(err) && backupTargetName != types.DefaultBackupTargetName {
+		// If the backup target is not found, we will try to use the default backup target.
 		backupTarget, err = b.ds.GetDefaultBackupTargetRO()
 		if err != nil {
 			return nil, werror.NewInvalidError(errors.Wrapf(err, "failed to get default backup target").Error(), "")
--- a/webhook/resources/replica/validator.go
+++ b/webhook/resources/replica/validator.go
@@ -74,10 +74,10 @@ func (r *replicaValidator) Update(request *admission.Request, oldObj runtime.Obj
 	return nil
 }
 
-func (r *replicaValidator) Delete(request *admission.Request, oldObj runtime.Object) error {
-	replica, ok := oldObj.(*longhorn.Replica)
+func (r *replicaValidator) Delete(request *admission.Request, obj runtime.Object) error {
+	replica, ok := obj.(*longhorn.Replica)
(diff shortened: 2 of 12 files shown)


derekbit:
Running regression test
~~https://ci.longhorn.io/job/private/job/longhorn-tests-regression/8762/~~
~~https://ci.longhorn.io/job/private/job/longhorn-tests-regression/8776/~~
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/8811/

webhook/resources/backupbackingimage/mutator.go
```diff
@@ -56,8 +56,15 @@ func (b *backupBackingImageMutator) Create(request *admission.Request, newObj ru
 	}
 	patchOps = append(patchOps, mutatePatchOps...)
 
-	backupTarget, err := b.ds.GetBackupTargetRO(backupBackingImage.Spec.BackupTargetName)
-	if apierrors.IsNotFound(err) {
+	backupTargetName := backupBackingImage.Spec.BackupTargetName
+	if backupTargetName == "" {
+		backupTargetName = types.DefaultBackupTargetName
+	}
```
shuo-wu, line 62: NIT: redundant

derekbit: Redundant?

webhook/resources/replica/validator.go
```diff
@@ -106,21 +106,26 @@ func (r *replicaValidator) validateReplicaDeletion(replica *longhorn.Replica) er
...
+		if r.Spec.EvictionRequested {
+			continue
 		}
+
+		availableReplicas[r.Name] = struct{}{}
 	}
 	if len(availableReplicas) == 1 {
 		if _, ok := availableReplicas[replica.Name]; ok {
-			return fmt.Errorf("cannot delete replica %v because volume %v only has one available replica", replica.Name, volume.Name)
+			return fmt.Errorf("no other healthy replica available, cannot delete replica %v since it may still contain data for recovery", replica.Name, volume.Name)
```
c3y1huang: > fmt.Errorf call needs 1 arg but has 2 args

derekbit: Fixed. Thank you.

webhook/resources/replica/validator.go
```diff
@@ -106,21 +106,26 @@ func (r *replicaValidator) validateReplicaDeletion(replica *longhorn.Replica) er
...
+		if r.Spec.EvictionRequested {
+			continue
 		}
+
+		availableReplicas[r.Name] = struct{}{}
 	}
 	if len(availableReplicas) == 1 {
 		if _, ok := availableReplicas[replica.Name]; ok {
-			return fmt.Errorf("cannot delete replica %v because volume %v only has one available replica", replica.Name, volume.Name)
+			return fmt.Errorf("no other healthy replica available, cannot delete replica %v since it may still contain data for recovery", replica.Name)
```
c3y1huang: Suggest keeping the volume name to provide better context for debugging.
```suggestion
			return fmt.Errorf("no other healthy replica available for volume %v, cannot delete replica %v since it may still contain data for recovery", volume.Name, replica.Name)
```

derekbit: Good idea. Thank you.

c3y1huang (approved):

webhook/resources/backupbackingimage/mutator.go
shuo-wu: The code below would cover and handle this case:
```
backupTarget, err := b.ds.GetBackupTargetRO(backupTargetName)
	if apierrors.IsNotFound(err) && backupTargetName != types.DefaultBackupTargetName {
		// If the backup target is not found, we will try to use the default backup target.
		backupTarget, err = b.ds.GetDefaultBackupTargetRO()
		if err != nil {
			return nil, werror.NewInvalidError(errors.Wrapf(err, "failed to get default backup target").Error(), "")
		}
		patchOps = append(patchOps, fmt.Sprintf(`{"op": "replace", "path": "/spec/backupTargetName", "value": "%s"}`, backupTarget.Name))
	} else if err != nil {
		return nil, werror.NewInvalidError(errors.Wrapf(err, "failed to get backup target of backup backing image").Error(), "")
	}
```

PR #3806, opened 2025-05-29
feat(scheduler): delay replica scheduling until volume is attached for `best-effort` locality

bachmanity1:

Issue https://github.com/longhorn/longhorn/issues/11007

Changed scheduler logic for `best-effort` locality to wait until volume is attached to a node and then attempt to schedule at least one replica to the node where volume is attached.

### DIFF
--- a/datastore/longhorn.go
+++ b/datastore/longhorn.go
@@ -2100,8 +2100,8 @@ func (s *DataStore) CheckDataEngineImageReadiness(image string, dataEngine longh
 	return s.CheckEngineImageReadiness(image, nodes...)
 }
 
-// CheckDataEngineImageReadyOnAtLeastOneVolumeReplica checks if the IMAGE is deployed on the NODEID and on at least one of the the volume's replicas
-func (s *DataStore) CheckDataEngineImageReadyOnAtLeastOneVolumeReplica(image, volumeName, nodeID string, dataLocality longhorn.DataLocality, dataEngine longhorn.DataEngineType) (bool, error) {
+// IsDataEngineImageReady checks if the IMAGE is deployed on the NODEID and, if data locality is disabled, also on at least one replica node of the volume.
+func (s *DataStore) IsDataEngineImageReady(image, volumeName, nodeID string, dataLocality longhorn.DataLocality, dataEngine longhorn.DataEngineType) (bool, error) {
 	isReady, err := s.CheckDataEngineImageReadiness(image, dataEngine, nodeID)
 	if err != nil {
 		return false, errors.Wrapf(err, "failed to check data engine image readiness of node %v", nodeID)
@@ -2111,6 +2111,11 @@ func (s *DataStore) CheckDataEngineImageReadyOnAtLeastOneVolumeReplica(image, vo
 		return isReady, nil
 	}
 
+	return s.checkDataEngineImageReadyOnAtLeastOneVolumeReplica(image, volumeName)
+}
+
+// checkDataEngineImageReadyOnAtLeastOneVolumeReplica checks if the IMAGE is deployed on at least one replica node of the volume.
+func (s *DataStore) checkDataEngineImageReadyOnAtLeastOneVolumeReplica(image, volumeName string) (bool, error) {
 	replicas, err := s.ListVolumeReplicas(volumeName)
 	if err != nil {
 		return false, errors.Wrapf(err, "failed to get replicas for volume %v", volumeName)
@@ -2129,6 +2134,7 @@ func (s *DataStore) CheckDataEngineImageReadyOnAtLeastOneVolumeReplica(image, vo
 	if !hasScheduledReplica {
 		return false, errors.Errorf("volume %v has no scheduled replicas", volumeName)
 	}
+
 	return false, nil
 }
 
--- a/scheduler/replica_scheduler_test.go
+++ b/scheduler/replica_scheduler_test.go
@@ -1694,9 +1694,7 @@ func (s *TestSuite) TestScheduleReplicaToDiskOnLocalNode(c *C) {
 	rs := NewReplicaScheduler(nil)
 	volume := newVolume(TestVolumeName, 2)
 	replica1 := newReplicaForVolume(volume)
-	replica1.Spec.HealthyAt = getTestNow().String()
 	replica2 := newReplicaForVolume(volume)
-	replica2.Spec.HealthyAt = getTestNow().String()
 	replicas := map[string]*longhorn.Replica{}
 	replicas[replica1.Name] = replica1
 	replicas[replica2.Name] = replica2
--- a/manager/volume.go
+++ b/manager/volume.go
@@ -237,7 +237,7 @@ func (m *VolumeManager) Attach(name, nodeID string, disableFrontend bool, attach
 		return nil, err
 	}
 
-	if isReady, err := m.ds.CheckDataEngineImageReadyOnAtLeastOneVolumeReplica(v.Spec.Image, v.Name, node.Name, v.Spec.DataLocality, v.Spec.DataEngine); !isReady {
+	if isReady, err := m.ds.IsDataEngineImageReady(v.Spec.Image, v.Name, node.Name, v.Spec.DataLocality, v.Spec.DataEngine); !isReady {
 		if err != nil {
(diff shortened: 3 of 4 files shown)


controller/volume_controller.go
```diff
@@ -1749,6 +1749,10 @@ func (c *VolumeController) reconcileVolumeCondition(v *longhorn.Volume, e *longh
 
 			r.Spec.HardNodeAffinity = v.Spec.NodeID
 		}
+		// For best-effort locality, wait until the volume is attached to a node before scheduling the replica.
+		if v.Spec.DataLocality == longhorn.DataLocalityBestEffort && v.Spec.NodeID == "" {
+			continue
+		}
```
COLDTURNIP: Nit: organize the code blocks for `DataLocality`
```suggestion
		switch v.Spec.DataLocality {
		case longhorn.DataLocalityStrictLocal:
			if v.Spec.NodeID == "" {
				continue
			}
			r.Spec.HardNodeAffinity = v.Spec.NodeID
		case longhorn.DataLocalityBestEffort:
			// For best-effort locality, wait until the volume is attached to a node before scheduling the replica.
			if v.Spec.NodeID == "" {
				continue
			}
		}
```

COLDTURNIP (commented):
Overall, LGTM.

COLDTURNIP:
@coderabbitai review

COLDTURNIP:
Also, please follow the commit comment linting rules: https://github.com/longhorn/longhorn-manager/actions/runs/15556110001/job/43797125257?pr=3806

shuo-wu (approved):
In general LGTM
Is it possible to add a unit test or modify an existing unit test for this case?

scheduler/replica_scheduler_test.go
```diff
@@ -1689,3 +1689,43 @@ func getTestNow() time.Time {
 	now, _ := time.Parse(time.RFC3339, TestTimeNow)
 	return now
 }
+
+func (s *TestSuite) TestScheduleReplicaToDiskOnLocalNode(c *C) {
+	rs := NewReplicaScheduler(nil)
+	volume := newVolume(TestVolumeName, 2)
+	replica1 := newReplicaForVolume(volume)
+	replica1.Spec.HealthyAt = getTestNow().String()
```
shuo-wu: IIRC, we don't need to set `Spec.HealthyAt` for this kind of newly created replica.

bachmanity1: `newReplicaForVolume` function doesn't set `HealthyAt` field, but it is required to pass the following if statement inside the `scheduleReplicaToDiskOnLocalNode` function:
```golang
		if r.Spec.NodeID == localNodeID && r.Spec.HealthyAt != "" && r.Spec.FailedAt == "" {
			return
		}
```
The purpose of this if statement is to find a healthy replica that is already assigned to the local node. Maybe, we could change this if statement to look like this?
```golang
		if r.Spec.NodeID == localNodeID && r.Spec.FailedAt == "" {
			return
		}
```

shuo-wu: Typically, `r.Spec.HealthyAt` will be set when the replica is started as a healthy one of an attached volume for the first time.

> The purpose of this if statement is to find a healthy replica that is already assigned to the local node. Maybe, we could change this if statement to look like this?

Yes. This change should make more sense.

bachmanity1:
Hi @COLDTURNIP @shuo-wu, I've addressed your comments. Could you please take another look?

derekbit:
Hello @hookak 
Could you help review the PR when you are available? Thanks.

datastore/longhorn.go
```diff
@@ -2098,7 +2098,7 @@ func (s *DataStore) CheckDataEngineImageReadyOnAtLeastOneVolumeReplica(image, vo
 		return false, errors.Wrapf(err, "failed to check data engine image readiness of node %v", nodeID)
 	}
 
-	if !isReady || dataLocality == longhorn.DataLocalityStrictLocal {
+	if !isReady || dataLocality == longhorn.DataLocalityStrictLocal || dataLocality == longhorn.DataLocalityBestEffort {
 		return isReady, nil
 	}
```
derekbit, line 2112: Should we skip the following checks if the best-effort volume's replicas are already scheduled and created?

bachmanity1: Yes, I think we can skip because if replicas are already created and scheduled, then one of the replicas is running on the attached node, so it's enough to check engine readiness only on the attached node. If the engine is not ready on the attached node, we'd need to return false anyway, but if it's ready, then it means that the condition `the IMAGE is deployed on the NODEID and on at least one of the volume's replicas` is satisfied, so we can return true.

hookak: For best-effort volumes, the attach operation must complete first in order to allow replica scheduling, as implemented in lines 1751–1755.
However, in order to proceed with attach, the function CheckDataEngineImageReadyOnAtLeastOneVolumeReplica must return true.
Despite its name, this function should return true even if it cannot verify the engineImage of any replica in the case of a best-effort volume.
Therefore, the behavior of the function is inconsistent with its name, which can cause confusion.

That said, the current implementation should still work as intended, so this inconsistency is unlikely to cause functional issues.

derekbit: Agreed wtih @bachmanity1 and @hookak.
BTW, to avoid confusion, we can move `dataLocality == longhorn.DataLocalityStrictLocal || dataLocality == longhorn.DataLocalityBestEffort` out of `CheckDataEngineImageReadyOnAtLeastOneVolumeReplica`. WDYT?

hookak: > BTW, to avoid confusion, we can move dataLocality == longhorn.DataLocalityStrictLocal || dataLocality == longhorn.DataLocalityBestEffort out of CheckDataEngineImageReadyOnAtLeastOneVolumeReplica. WDYT?

Yes, I think that could be a good way to reduce confusion.
For `strict-local`, the condition is always satisfied since a replica must exist on the attached node.
However, for `best-effort`, while the scheduler tries to place a replica on the attached node, it may fall back to scheduling on a remote node. In that case, the condition required by `CheckDataEngineImageReadyOnAtLeastOneVolumeReplica` may not be guaranteed.

Fundamentally, I believe that the `delayed replica scheduling` behavior of `best-effort` and the logic in `CheckDataEngineImageReadyOnAtLeastOneVolumeReplica` cannot both be satisfied at the same time.

bachmanity1: Since both `strict-local` and `best-effort` still require the engine image to be ready on the attached node, how about simply renaming this function to `IsDataEngineImageReady`? We could also add a comment noting that if data locality is disabled, the engine image must also be ready on at least one of the nodes where the replicas are scheduled.

derekbit: I'm fine with the proposal that renames the original `CheckDataEngineImageReadyOnAtLeastOneVolumeReplica` to `IsDataEngineImageReady`.

In addition, I think we can wrap the codes
```
	replicas, err := s.ListVolumeReplicas(volumeName)
	if err != nil {
		return false, errors.Wrapf(err, "failed to get replicas for volume %v", volumeName)
	}

	hasScheduledReplica := false
	for _, r := range replicas {
		isReady, err := s.CheckDataEngineImageReadiness(image, r.Spec.DataEngine, r.Spec.NodeID)
		if err != nil || isReady {
			return isReady, err
		}
		if r.Spec.NodeID != "" {
			hasScheduledReplica = true
		}
	}
	if !hasScheduledReplica {
		return false, errors.Errorf("volume %v has no scheduled replicas", volumeName)
	}
	return false, nil
```
to `CheckDataEngineImageReadyOnAtLeastOneVolumeReplica`.

WDYT @bachmanity1 @hookak @shuo-wu?

derekbit:
@bachmanity1 Can you consolidate the commits into few commits in the PR? Thank you.

shuo-wu (approved):
LGTM

bachmanity1:
> @bachmanity1 Can you consolidate the commits into few commits in the PR? Thank you.

can we just squash merge?

shuo-wu:
You can squash some of your commits by `git rebase -i origin/master`. This makes each of the commits perfect.

COLDTURNIP (approved):

COLDTURNIP:
LGTM

PR #3815, opened 2025-06-03
feat: improve starting time of longhorn-csi-plugin

PhanLe1010:
### Add longhorn-csi-plugin retry logic
Retry the Longhorn client initialization instead of failing immediately after
the first try

### Add StartupProbe for Longhorn CSI plugin
StartupProbe is needed because Longhorn CSI plugin might need more time
at the begining to establish connection to Longhorn manager API. Without
StartupProbe, we would have to rely on the LivenessProbe which will not
wait for the long enough, unnessary crash the Longhorn CSI plugin container,
and causing more delay.

longhorn/longhorn#9482


derekbit:

PR #3816, opened 2025-06-03
fix(backing_image): reconcile backing image disk evict event by spec disks

COLDTURNIP:

Issue longhorn/longhorn#11034

There is some chance that the status does not reflect spec due to reconciliation timing gap. The `DataStore.GetDiskBackingImageMap` lists the backing images by disk, and is used by the following two controllers:

- node controller: to attach node/disk eviction flag onto backing image spec
- backing image controller: when node changed, list the backing image CRs on evicted disks

The CR's status should be reconciled from the spec instead of the status.

longhorn/longhorn#10464

### DIFF
--- a/datastore/longhorn.go
+++ b/datastore/longhorn.go
@@ -5966,7 +5966,7 @@ func (s *DataStore) IsV2DataEngineDisabledForNode(nodeName string) (bool, error)
 	return false, nil
 }
 
-func (s *DataStore) GetDiskBackingImageMap() (map[string][]*longhorn.BackingImage, error) {
+func (s *DataStore) GetCurrentDiskBackingImageMap() (map[string][]*longhorn.BackingImage, error) {
 	diskBackingImageMap := map[string][]*longhorn.BackingImage{}
 	backingImages, err := s.ListBackingImages()
 	if err != nil {
--- a/controller/backing_image_controller.go
+++ b/controller/backing_image_controller.go
@@ -1527,7 +1527,7 @@ func (bic *BackingImageController) enqueueBackingImageForNodeUpdate(oldObj, curr
 		}
 	}
 
-	diskBackingImageMap, err := bic.ds.GetDiskBackingImageMap()
+	diskBackingImageMap, err := bic.ds.GetCurrentDiskBackingImageMap()
 	if err != nil {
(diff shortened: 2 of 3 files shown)


COLDTURNIP:
@coderabbitai review

datastore/longhorn.go
```diff
@@ -5974,7 +5974,7 @@ func (s *DataStore) GetDiskBackingImageMap() (map[string][]*longhorn.BackingImag
 	}
 
 	for _, bi := range backingImages {
-		for diskUUID := range bi.Status.DiskFileStatusMap {
+		for diskUUID := range bi.Spec.DiskFileSpecMap {
```
mantissahz: Hi @COLDTURNIP, 

Could you explain more about the difference between `bi.Status.DiskFileStatusMap` and `bi.Spec.DiskFileSpecMap`? Why do we check the `bi.Status.DiskFileStatusMap` before?

COLDTURNIP: I think the original intention is to fetch the CRs base on up to date status.

- To node controller, it is asking the backing image controller to proceed disk eviction, hence it modifies the desired spec
- To backing image controller, it is responsible to update the disk map in status to reflect the node disk status.

After rethinking, we should list the BI CR in different way for these two controllers. WDYT @derekbit @shuo-wu ?

derekbit:

c3y1huang (commented):
Could there be a potential mismatch in the counts [here](https://github.com/longhorn/longhorn-manager/blob/604a88fbb69b40727e2470c2e63b99842a748ea8/controller/backing_image_controller.go#L824-L847)?

If `backingImage.Spec.DiskFileSpecMap[diskUUID]` is missing, wouldn't it be cleaner to just requeue and let the node controller handle it in the next reconcile loop?

COLDTURNIP:
Understood. Let me refine the change.

- Will still rename `GetDiskBackingImageMap` to `GetCurrentDiskBackingImageMap` to reflect the intention of the map.
- We may improve the backing image replenishing to eliminate the node re-reconciliation in the future.

c3y1huang (approved):

derekbit (approved):
LGTM

PR #3820, opened 2025-06-04
fix(snapshot): set OwnerID and get notification when volume is updated

mantissahz:

Issue # longhorn/longhorn#10874

### DIFF
--- a/controller/snapshot_controller.go
+++ b/controller/snapshot_controller.go
@@ -100,7 +100,8 @@ func NewSnapshotController(
 	sc.cacheSyncs = append(sc.cacheSyncs, ds.EngineInformer.HasSynced)
 
 	if _, err = ds.VolumeInformer.AddEventHandlerWithResyncPeriod(cache.ResourceEventHandlerFuncs{
-		DeleteFunc: sc.enqueueVolumeChange,
+		UpdateFunc: sc.enqueueVolumeChange,
+		DeleteFunc: sc.enqueueVolumeDeleted,
 	}, 0); err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


controller/snapshot_controller.go
```diff
@@ -100,7 +100,9 @@ func NewSnapshotController(
 	sc.cacheSyncs = append(sc.cacheSyncs, ds.EngineInformer.HasSynced)
 
 	if _, err = ds.VolumeInformer.AddEventHandlerWithResyncPeriod(cache.ResourceEventHandlerFuncs{
-		DeleteFunc: sc.enqueueVolumeChange,
+		UpdateFunc: sc.enqueueVolumeChange,
+		DeleteFunc: sc.enqueueVolumeDeleted,
+		//DeleteFunc: sc.enqueueVolumeChange,
```
shuo-wu: NIT: Remove this line

controller/snapshot_controller.go
```diff
@@ -211,7 +213,51 @@ func filterSnapshotsForEngineEnqueuing(oldEngine, curEngine *longhorn.Engine, sn
...
 }
 
-func (sc *SnapshotController) enqueueVolumeChange(obj interface{}) {
+func (sc *SnapshotController) enqueueVolumeChange(old, new interface{}) {
+	oldVol, ok := old.(*longhorn.Volume)
+	if !ok {
+		utilruntime.HandleError(fmt.Errorf("received unexpected obj: %#v", old))
+		return
+	}
+	newVol, ok := new.(*longhorn.Volume)
```
shuo-wu, line 224: NIT: Would you add a comment that explains when newVol should be handled by SnapshotController?

controller/snapshot_controller.go
```diff
@@ -211,7 +213,51 @@ func filterSnapshotsForEngineEnqueuing(oldEngine, curEngine *longhorn.Engine, sn
...
+	}
+
+	snapshots, err := sc.ds.ListVolumeSnapshotsRO(newVol.Name)
+	if err != nil {
+		utilruntime.HandleError(fmt.Errorf("snapshot controller failed to list snapshots when enqueuing volume %v: %v", newVol.Name, err))
+		return
+	}
+	for _, snap := range snapshots {
+		// Longhorn#10874:
+		// Requeueue the snapshot if there is an attachment ticket for it to ensure
```
shuo-wu: NIT: typo `Requeueue`

controller/snapshot_controller.go
```diff
@@ -760,7 +818,27 @@ func (sc *SnapshotController) isResponsibleFor(snap *longhorn.Snapshot) (bool, e
 		}
 		return false, err
 	}
-	return sc.controllerID == volume.Status.OwnerID, nil
+
+	defaultEngineImage, err := sc.ds.GetSettingValueExisted(types.SettingNameDefaultEngineImage)
```
shuo-wu: May need to skip this check for v2 volumes

controller/snapshot_controller.go
mantissahz: Removed. Thanks

controller/snapshot_controller.go
mantissahz: Added. Thanks. Please review it again to ensure that nothing needs to be modified.

controller/snapshot_controller.go
mantissahz: Corrected. Thanks.

controller/snapshot_controller.go
mantissahz: Modified. Thanks.

controller/snapshot_controller.go
```diff
@@ -760,7 +819,32 @@ func (sc *SnapshotController) isResponsibleFor(snap *longhorn.Snapshot) (bool, e
...
+	if err != nil {
+		return false, err
+	}
+
+	isResponsible := isControllerResponsibleFor(sc.controllerID, sc.ds, snap.Name, volume.Status.OwnerID, snap.Status.OwnerID)
+
+	if types.IsDataEngineV2(volume.Spec.DataEngine) {
+		return isResponsible, nil
+	}
```
shuo-wu: NIT: Move this to the top of `defaultEngineImage, err := sc.ds.GetSettingValueExisted(types.SettingNameDefaultEngineImage)`

derekbit:
@mantissahz Could you provide the test results from 20 runs to verify the fix? Thank you.

mantissahz:
>@mantissahz Could you provide the test results from 20 runs to verify the fix? Thank you.

Sure. [Testing is Passed](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/8859/)

cc @PhanLe1010, @derekbit

controller/snapshot_controller.go
mantissahz: Modified. Thanks

PhanLe1010:
> Based on the observation, there is a transition state that volume.status.ownerID is not updated after volume got a new owner. If this situation (described in the observation) occurs, the snapshot won't be reconciled, and the volumeattachment ticket is unable to be deleted. As a result, the volume is unable to be detached.

@derekbit In this case, I think we only need to enqueue the snapshot CR when the volume change. We don't need to set the `snapshot.Status.OwnerID` WDYT?

cc @mantissahz

derekbit:
> > Based on the observation, there is a transition state that volume.status.ownerID is not updated after volume got a new owner. If this situation (described in the observation) occurs, the snapshot won't be reconciled, and the volumeattachment ticket is unable to be deleted. As a result, the volume is unable to be detached.
> 
> @derekbit In this case, I think we only need to enqueue the snapshot CR when the volume change. We don't need to set the `snapshot.Status.OwnerID` WDYT?
> 
> cc @mantissahz

This should be feasible. @mantissahz Can you try the solution and update your testing result?

mantissahz:
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/8860
Here is the test as suggested by Phan.
Over 20 successful runs.

PhanLe1010 (approved):
LGTM.

Thank you for the fix!

derekbit (approved):
LGTM. Well done @mantissahz!

derekbit:

PR #3833, opened 2025-06-04
fix(backingimage): prevent eviction of in-used copies

COLDTURNIP:

Issue longhorn/longhorn#11053

Backing image copy should be deleted only when all replicas are removed from a disk. If the copy is deleted, the engine monitor will mark the replica as ERR, which prevent the removal by the volume controller.

### DIFF
--- a/controller/backing_image_controller.go
+++ b/controller/backing_image_controller.go
@@ -281,7 +281,7 @@ func (bic *BackingImageController) syncBackingImage(key string) (err error) {
 	}()
 
 	if backingImage.DeletionTimestamp != nil {
-		replicas, err := bic.ds.ListReplicasByBackingImage(backingImage.Name)
+		replicas, err := bic.ds.ListReplicasByBackingImage(backingImage.Name, "")
 		if err != nil {
 			return err
 		}
@@ -345,7 +345,9 @@ func (bic *BackingImageController) syncBackingImage(key string) (err error) {
 		return err
 	}
 
-	bic.cleanupEvictionRequestedBackingImageCopies(backingImage)
+	if err := bic.cleanupEvictionRequestedBackingImageCopies(backingImage); err != nil {
+		return err
+	}
 
 	if types.IsDataEngineV2(backingImage.Spec.DataEngine) {
 		return bic.handleV2BackingImage(backingImage)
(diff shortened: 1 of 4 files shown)


COLDTURNIP:
This PR is verified locally using the steps in longhorn/longhorn#11053.

COLDTURNIP:
@coderabbitai review

controller/backing_image_controller.go
```diff
@@ -901,11 +901,36 @@ func (bic *BackingImageController) cleanupEvictionRequestedBackingImageCopies(bi
 		// only this controller can gather all the information of all the copies of this backing image at once.
 		// By deleting the disk from the spec, backing image manager controller will delete the copy on that disk.
 		// TODO: introduce a new CRD for the backing image copy so we can delete the copy like volume controller deletes replicas.
+		used, err := bic.isBIDiskFileUsedByReplicas(bi.Name, diskUUID)
+		if err != nil {
+			log.WithError(err).Warnf("Failed to check if the backing image copy on disk %v is used by replicas", diskUUID)
```
derekbit: If failed and is being used, how long will the controller continue retrying the deletion?

controller/backing_image_controller.go
```diff
@@ -901,11 +901,36 @@ func (bic *BackingImageController) cleanupEvictionRequestedBackingImageCopies(bi
...
+			continue
+		}
+		if used {
+			log.Infof("Backing image copy on disk %v is used by replicas. Copy eviction is blocked", diskUUID)
```
derekbit:
```suggestion
			log.Debugf("Backing image copy on disk %v is used by replicas. Copy eviction is blocked", diskUUID)
```

controller/backing_image_controller.go
COLDTURNIP: The reconciliation will be triggered each time when replica get updated (deleted or finalized) or cleaned by K8s.

COLDTURNIP: Should we raise the error here to trigger another reconciliation?

COLDTURNIP:
@coderabbitai review

controller/backing_image_controller.go
derekbit: Seems yes, WDYT? Any side effects?

controller/backing_image_controller.go
```diff
@@ -901,11 +901,31 @@ func (bic *BackingImageController) cleanupEvictionRequestedBackingImageCopies(bi
 		// only this controller can gather all the information of all the copies of this backing image at once.
 		// By deleting the disk from the spec, backing image manager controller will delete the copy on that disk.
 		// TODO: introduce a new CRD for the backing image copy so we can delete the copy like volume controller deletes replicas.
+		used, err := bic.isBIDiskFileUsedByReplicas(bi.Name, diskUUID)
```
c3y1huang: NIT
```suggestion
		isUsed, err := bic.isBIDiskFileUsedByReplicas(bi.Name, diskUUID)
```

c3y1huang (approved):

derekbit (approved):
LGTM

derekbit:

PR #3846, opened 2025-06-11
Draft: Pass mkfs parameters from Storage class to share-manager

Florianisme:
Issue #11107

Passes the required mkfs Parameters as an argument to the share-manager if it needs to format the volume's filesystem

Requires a change in the share-manager implementation


derekbit:
@coderabbitai review

controller/share_manager_controller.go
```diff
@@ -1335,6 +1341,30 @@ func (c *ShareManagerController) createShareManagerPod(sm *longhorn.ShareManager
 	return pod, nil
 }
 
+func (c *ShareManagerController) splitFormatOptions(sc *storagev1.StorageClass) []string {
+	var formatOptions []string
+
+	if mkfsParams, ok := sc.Parameters["mkfsParams"]; ok {
+		regex, err := regexp.Compile("-\\w+[\\s=]?[\"']?\\w*[\"']?")
```
derekbit: Can it convert `-I 256 -b 4096 -O ^metadata_csum,^64bit` to `[]string{"-I", "256", "-b", "4096", "-O", "^metadata_csum,^64bit"}`?

Florianisme: It currently does not but I can adjust the regex. Note that it would actually return 
`[]string{"-I 256", "-b 4096", "-O ^metadata_csum,^64bit"}` instead.

Florianisme: I have actually found a few other variations which currently do not work. I will adjust the regex and push again with new test cases added

controller/share_manager_controller.go
```diff
@@ -1493,6 +1520,15 @@ func (c *ShareManagerController) createPodManifest(sm *longhorn.ShareManager, da
 		},
 	}
 
+	if len(formatOptions) > 0 {
+		podSpec.Spec.Containers[0].Env = append(podSpec.Spec.Containers[0].Env, []corev1.EnvVar{
+			{
+				Name:  "FS_FORMAT_OPTIONS",
+				Value: fmt.Sprint(strings.Join(formatOptions, ",")),
```
derekbit: Can we use a space (`" "`) instead, since the parameters for -O are comma-separated?
If yes, need to update https://github.com/longhorn/longhorn-share-manager/blob/master/pkg/server/share_manager.go#L265 as well.

Florianisme: Parameters such as Labels can also contain spaces, what do you think about separation by pipe `|` or semicolon?

derekbit: Using `":"` might be good. WDYT?

Florianisme: Sure, will open another MR in the longhorn-share-manager Repo and link it under the original issue

controller/share_manager_controller.go
```diff
@@ -1271,6 +1274,9 @@ func (c *ShareManagerController) createShareManagerPod(sm *longhorn.ShareManager
 			}
 			tolerationsFromStorageClass := c.getShareManagerTolerationsFromStorageClass(sc)
 			tolerations = append(tolerations, tolerationsFromStorageClass...)
+
+			// A storage class can override mkfs parameters which need to be passed to the share manager
+			formatOptions = c.splitFormatOptions(sc)
```
derekbit, line 1279: Why not move the parsing function to longhorn-share-manager? We can simply assign `sc.Parameters["mkfsParams"]` to the env `FS_FORMAT_OPTIONS`?

Florianisme: Yes could do that. My idea was that we at least made sure the parameters somewhat look like we expect with the regex. If there is something wrong with the format, we could fail early.

Also might be good for debugging as you could see, which individual options were passed into the share-manager. I see your point though - why split, join and then split again?

derekbit: >  I see your point though - why split, join and then split again?

Yes, you are right.

Thanks for the explanation. I'm fine with the current implementation.
Let me know if the PRs are ready for review. Thank you.

Florianisme: If I should change it thats fine for me. Otherwise PRs are ready for review now.
I'll fill out the ready-for-testing checklist in parallel

derekbit (approved):
LGTM

derekbit:
Thanks @Florianisme for the contribution!

PR #3859, opened 2025-06-16
Add a new validation on validator.go of a node

Felipalds:

[IMPROVEMENT] Validate node.longhorn.io resource spec fields
[#11079](https://github.com/longhorn/longhorn/issues/11079)

This PR adds a new validation to node validator.go. It ensures that a Node never gets the Spec.name deleted.


webhook/resources/node/validator.go
```diff
@@ -143,6 +143,11 @@ func (n *nodeValidator) Update(request *admission.Request, oldObj runtime.Object
 		}
 	}
 
+	// We need to ensure that the name is not empty because it can lead to errors in the Longhorn
+	if newNode.Spec.Name == "" {
+		return werror.NewInvalidError("Node name is invalid. You can't have a Spec.Name empty.", "")
```
derekbit: It is a convention.

```suggestion
		return werror.NewInvalidError("node name is invalid. You can't have a Spec.Name empty", "")
```

derekbit (dismissed):
Thanks @Felipalds

derekbit:
image
Hello @Felipalds 
You can merge the three commits to single one and rename the commit title to `feat: add a new validation on validator. Go of a node`.

More info for conventional commit: https://www.conventionalcommits.org/en/v1.0.0/

derekbit (approved):

Felipalds:
How to test this PR:
1. Create a Longhorn cluster
2. Try to remove the node Spec.name: `kubectl -n longhorn-system edit node.longhorn.io <node name>`. You can remove the entire property or just put the Spec.Name as ""
3. The edit should be validated as invalid, because the name cannot be empty or null.

PR #3861, opened 2025-06-18
fix(node): allow deleting node during uninstallation

COLDTURNIP:

Issue longhorn/longhorn#11131

Webhook prevents deleting Longhorn nodes when the corresponding Kubernetes node is healthy. Adding a special annotation, `delete-node-from-longhorn`, to allow deleting during Longhorn uninstallation.

#3860

### DIFF
--- a/webhook/resources/node/validator.go
+++ b/webhook/resources/node/validator.go
@@ -228,7 +228,18 @@ func (n *nodeValidator) Delete(request *admission.Request, obj runtime.Object) e
 		return werror.NewInvalidError(fmt.Sprintf("%v is not a *longhorn.Node", obj), "")
 	}
 
-	// only remove node from longhorn without any volumes on it
+	// Annotations `DeleteNodeFromLonghorn` is used to note that deleting node is by Longhorn during uninstalling.
+	// When `isUninstalling` is true, the node is deleted by Longhorn, allows the deletion;
+	// otherwise, continue to validate the instance and the node status.
+	isUninstalling := false
+	if node.Annotations != nil {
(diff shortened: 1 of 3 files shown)


COLDTURNIP:
@coderabbitai review

webhook/resources/node/validator.go
```diff
@@ -228,7 +228,17 @@ func (n *nodeValidator) Delete(request *admission.Request, obj runtime.Object) e
...
-	// only remove node from longhorn without any volumes on it
+	// Annotations `DeleteNodeFromLonghorn` is used to note that deleting node is by Longhorn during uninstalling.
+	// When `isUninstalled` is true, the node is deleted by Longhorn.
+	isUninstalled := false
+	if node.Annotations != nil {
+		_, isUninstalled = node.Annotations[types.GetLonghornLabelKey(types.DeleteNodeFromLonghorn)]
+	}
+	if isUninstalled {
+		return nil
+	}
```
derekbit: Shouldn't it be
```suggestion
	if !isUninstalled {
		return werror.NewInvalidError(...)
	}
```

COLDTURNIP: If the node is annotated to be deleted, the webhook should allow the deletion directly; otherwise, it should perform the following Kubernetes node status check. Maybe we should rename to `isUninstalling`.

derekbit: okay. can you add a comment here?

derekbit: cc @COLDTURNIP

PR #3868, opened 2025-06-23
fix: do not silent the error returned by ReconcileBackupVolumeState

PhanLe1010:
Right now when there are multiple backup volumes of the same volume on the same backup target, we silently ignore the error, return nil, and abort the volume syncing: https://github.com/longhorn/longhorn-manager/blob/7b9d2aa08a572b2b8bce35b3ff0e90dfaa1b0eaf/controller/volume_controller.go#L530-L532

User would have no idea of why the volume is stuck.

longhorn/longhorn#11152

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -528,7 +528,7 @@ func (c *VolumeController) syncVolume(key string) (err error) {
 	}
 
 	if err := c.ReconcileBackupVolumeState(volume); err != nil {
-		return nil
+		return err
 	}
 
 	if err := c.ReconcileVolumeState(volume, engines, replicas); err != nil {


controller/volume_controller.go
```diff
@@ -528,7 +528,7 @@ func (c *VolumeController) syncVolume(key string) (err error) {
 	}
 
 	if err := c.ReconcileBackupVolumeState(volume); err != nil {
-		return nil
+		return err
```
shuo-wu, line 531: Actually, I am thinking if the controller can continue the reconciliation after encountering this issue: This function mainly updates the backup-related status fields for volume CRs. Would you leave them with the previous values and continue updating other fields if there is a backup volume error?

derekbit: WDYT? @PhanLe1010

PhanLe1010: Sorry for the delay. With the PR https://github.com/longhorn/longhorn-manager/pull/3893/files already trying to clean up duplicated backupvolume CR, I think it is best to simply return err here and retry later. The douplicated backup volume CR would be cleaned up and the normal volume sync follow would happen. WDYT @derekbit @shuo-wu ?

derekbit: Then, @PhanLe1010's impl LGTM.

derekbit: cc @mantissahz

shuo-wu: I am OK with Phan's idea as well

derekbit (approved):

mantissahz (approved):

derekbit:

PR #3871, opened 2025-06-25
fix(support-bundle): panic NPE after update conflicted

c3y1huang:

Issue longhorn/longhorn#11169

Remove from the queue if the update fails to avoid accessing a nil object.

`None`

`None`


c3y1huang:
@coderabbitai review

c3y1huang:
@coderabbitai review

controller/support_bundle_controller.go
```diff
@@ -234,11 +233,19 @@ func (c *SupportBundleController) handleStatusUpdate(record *supportBundleRecord
...
 			log.WithError(err).Warnf(SupportBundleMsgRequeueOnConflictFmt, supportBundle.Name)
 			c.enqueue(supportBundle)
 		}
+		// Avoid requeue if the object is nil to avoid unnecessary requeue loops.
+		// This can happen if the update failed, for example, when the object was
+		// deleted during reconciliation.
+		if _supportBundle == nil {
+			log.WithError(err).Warnf("SupportBundle %v is nil after update, skipping requeue", supportBundle.Name)
+			return
+		}
```
derekbit, line 247: If `_supportBundle` is nil, why isn't a not found error returned?

c3y1huang: If the object is not found, an error will be returned. However, other kinds of errors may also result in a nil object. Since the SupportBundle controller depends on status updates to make progress, a failure to update could potentially lead to unexpected behavior (not tested, and I assume this would be a rare case). So, to prevent such scenarios, it will be removed from the queue. Users can recreate the SupportBundle manually if this issue happens.
https://github.com/longhorn/longhorn-manager/blob/548229725656086ef4e6387038ed340616ed7810/datastore/longhorn.go#L5480-L5483

derekbit (approved):
LGTM

derekbit:
@mergify backport v1.8.x 1.9.x

c3y1huang:
@mergify backport v1.9.x

PR #3874, opened 2025-06-25
fix(storage-network): skip migratable RWX volume workloads restarted

c3y1huang:

Issue longhorn/longhorn#11158

For the storage network scenario, RWX volumes in Longhorn are not true NFS-based RWX volumes, so the workload should not be restarted when the CSI plugin pod restarts.

`None`

`None`


derekbit:

PR #3877, opened 2025-06-26
feat: implement exponential backoff retry for re-creating pods on failure

bachmanity1:

Issue https://github.com/longhorn/longhorn/issues/10263

### DIFF
--- a/controller/instance_manager_controller.go
+++ b/controller/instance_manager_controller.go
@@ -19,6 +19,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/record"
+	"k8s.io/client-go/util/flowcontrol"
 	"k8s.io/kubernetes/pkg/controller"
 
 	corev1 "k8s.io/api/core/v1"
@@ -61,6 +62,8 @@ type InstanceManagerController struct {
 	instanceManagerMonitorMap   map[string]chan struct{}
 
 	proxyConnCounter util.Counter
+
+	backoff *flowcontrol.Backoff
 
 	// for unit test
 	versionUpdater func(*longhorn.InstanceManager) error
@@ -140,6 +143,8 @@ func NewInstanceManagerController(
 		proxyConnCounter: proxyConnCounter,
 
 		versionUpdater: updateInstanceManagerVersion,
+
+		backoff: newBackoff(),
 	}
 
 	var err error
@@ -655,8 +660,16 @@ func (imc *InstanceManagerController) handlePod(im *longhorn.InstanceManager) er
 		return err
 	}
 
-	if err := imc.createInstanceManagerPod(im); err != nil {
-		return err
+	backoffID := im.Name
+	if imc.backoff.IsInBackOffSinceUpdate(backoffID, time.Now()) {
+		log.Infof("Skipping pod creation for instance manager %s, will retry after backoff of %s", im.Name, imc.backoff.Get(backoffID))
+	} else {
+		log.Infof("Creating pod for instance manager %s", im.Name)
+		imc.backoff.Next(backoffID, time.Now())
+
+		if err := imc.createInstanceManagerPod(im); err != nil {
+			return errors.Wrap(err, "failed to create pod for instance manager")
+		}
 	}
 
 	return nil
--- a/controller/utils.go
+++ b/controller/utils.go
@@ -1,9 +1,12 @@
 package controller
 
 import (
+	"time"
+
 	"github.com/sirupsen/logrus"
 
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/util/flowcontrol"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 
@@ -12,6 +15,28 @@ import (
 
 	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
 )
+
+const (
+	podRecreateInitBackoff = 1 * time.Second
+	podRecreateMaxBackoff  = 120 * time.Second
+	backoffGCPeriod        = 12 * time.Hour
+)
+
+// newBackoff returns a flowcontrol.Backoff and starts a background GC loop.
+func newBackoff() *flowcontrol.Backoff {
+	backoff := flowcontrol.NewBackOff(podRecreateInitBackoff, podRecreateMaxBackoff)
+
+	// Start a background GC loop.
+	go func() {
+		ticker := time.NewTicker(backoffGCPeriod)
+		defer ticker.Stop()
+		for range ticker.C {
+			backoff.GC()
+		}
+	}()
+
+	return backoff
+}
 
 func hasReplicaEvictionRequested(rs map[string]*longhorn.Replica) bool {
 	for _, r := range rs {
--- a/controller/backing_image_manager_controller.go
+++ b/controller/backing_image_manager_controller.go
@@ -66,6 +66,8 @@ type BackingImageManagerController struct {
 
 	replenishLock             *sync.Mutex
 	inProgressReplenishingMap map[string]string
+
+	podRecreateBackoff *flowcontrol.Backoff
 }
 
 type BackingImageManagerMonitor struct {
@@ -135,6 +137,8 @@ func NewBackingImageManagerController(
 
 		replenishLock:             &sync.Mutex{},
 		inProgressReplenishingMap: map[string]string{},
+
+		podRecreateBackoff: newBackoff(),
 	}
 
 	var err error
@@ -546,9 +550,16 @@ func (c *BackingImageManagerController) syncBackingImageManagerPod(bim *longhorn
 			// Similar to InstanceManagerController.
 			// Longhorn shouldn't create the pod when users set taints with NoExecute effect on a node the bim is preferred.
 			if c.controllerID == bim.Spec.NodeID {
-				log.Info("Creating backing image manager pod")
-				if err := c.createBackingImageManagerPod(bim); err != nil {
-					return err
+				backoffID := bim.Name
+				if c.podRecreateBackoff.IsInBackOffSinceUpdate(backoffID, time.Now()) {
+					log.Infof("Skipping pod creation for backing image manager %s, will retry after backoff of %s", bim.Name, c.podRecreateBackoff.Get(backoffID))
+				} else {
+					log.Infof("Creating pod for backing image manager %s", bim.Name)
+					c.podRecreateBackoff.Next(backoffID, time.Now())
+
+					if err := c.createBackingImageManagerPod(bim); err != nil {
+						return errors.Wrap(err, "failed to create pod for backing image manager")
+					}
 				}
 				bim.Status.CurrentState = longhorn.BackingImageManagerStateStarting
 				c.eventRecorder.Eventf(bim, corev1.EventTypeNormal, constant.EventReasonCreate, "Creating backing image manager pod %v for disk %v on node %v. Backing image manager state will become %v", bim.Name, bim.Spec.DiskUUID, bim.Spec.NodeID, longhorn.BackingImageManagerStateStarting)
--- a/controller/share_manager_controller.go
+++ b/controller/share_manager_controller.go
@@ -17,6 +17,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/record"
+	"k8s.io/client-go/util/flowcontrol"
 	"k8s.io/kubernetes/pkg/apis/core"
 	"k8s.io/kubernetes/pkg/controller"
 
@@ -59,6 +60,8 @@ type ShareManagerController struct {
 	ds *datastore.DataStore
 
 	cacheSyncs []cache.InformerSynced
+
+	backoff *flowcontrol.Backoff
 }
 
 func NewShareManagerController(
@@ -88,6 +91,8 @@ func NewShareManagerController(
 		eventRecorder: eventBroadcaster.NewRecorder(scheme, corev1.EventSource{Component: "longhorn-share-manager-controller"}),
 
 		ds: ds,
+
+		backoff: newBackoff(),
 	}
 
 	var err error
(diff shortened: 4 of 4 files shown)


controller/instance_manager_controller.go
```diff
@@ -652,8 +656,14 @@ func (imc *InstanceManagerController) handlePod(im *longhorn.InstanceManager) er
 		return err
 	}
 
-	if err := imc.createInstanceManagerPod(im); err != nil {
-		return err
+	backoffKey := im.Name
+	if canRun, retryAfter := imc.exponentialBackoff.CanRun(backoffKey); canRun {
+		log.Infof("Creating pod for instance manager %s", im.Name)
+		if err := imc.createInstanceManagerPod(im); err != nil {
```
bachmanity1, line 670: Even if the pod is created successfully and `err` is nil, it might still fail to start due to issues like insufficient hugepage resources, which causes the reconcile loop to recreate it endlessly. One way to handle this is to monitor the pod’s state for a while to make sure it’s actually running before resetting the retry interval—but it’s tricky to decide how long is “long enough.” Instead, I added a cleanup goroutine that runs periodically and resets the backoff only if the process using it has been inactive for a while.

derekbit:
Hello @@jangseon-ryu @hookak 
Could you help review the PR? Thanks.

jangseon-ryu:
@derekbit 
Sure, I’ll take a look and leave my comments soon

jangseon-ryu:
@derekbit 
I don’t have review approval permission, so I'm leaving a message instead.
This PR has already been reviewed within our team — we’re part of the same team within the company — and it was submitted with some revisions.
I’ve taken another look and didn’t find any issues that stand out, though please note that my expertise in Longhorn is limited, so I may have missed something.
( cc @hookak )

controller/utils.go
```diff
@@ -13,6 +15,93 @@ import (
 	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
 )
 
+const (
+	defaultInitBackoffInterval = 1 * time.Second
+	defaultMaxBackoffInterval  = 300 * time.Second
```
derekbit: The `defaultMaxBackoffInterval` is quite large. If the Instance Manager Pod fails to start due to insufficient hugepages on the host, even after the user fixes the issue, they might have to wait up to 5 minutes for the IM pod to restart in the worst-case scenario. Does the case make sense?
If YSS, to address this issue, could we make this interval configurable?

derekbit: WDYT @bachmanity1 @jangseon-ryu @hookak

jangseon-ryu: @derekbit 
You're right — selecting an appropriate value is not straightforward, so we referred to [Kubernetes](https://github.com/kubernetes/kubernetes/blob/ec80dcaf3892d0cf89e4fd7b1ea60300fb417d8a/pkg/kubelet/apis/config/v1beta1/defaults.go#L46)' default behavior as a guideline.
```go
	MaxContainerBackOff = 300 * time.Second
```
As you mentioned, if the user resolves the hugepage issue, they might still have to wait up to 5 minutes for the Instance Manager Pod to restart. Since this delay would occur before the volume is actively used, it’s unlikely to impact service availability in most cases.

However, we can't guarantee that it will be safe in all scenarios. To reduce potential risks, I agree that making the interval configurable is the better approach, with a default value of 60 or 120 seconds so users can adjust it based on their environment.

( cc @bachmanity1 )

jangseon-ryu (approved):

types/setting.go
```diff
@@ -1592,6 +1595,16 @@ var (
 		ReadOnly: false,
 		Default:  "false",
 	}
+
+	SettingDefinitionMaxPodRecreateBackoff = SettingDefinition{
+		DisplayName: "Max Pod Recreate Backoff",
+		Description: "Maximum backoff duration (in seconds) between pod recreation attempts. The interval doubles after each failure, up to this limit. Default is 120 seconds.",
```
derekbit: Hello @sushant-suse (doc engineer), could you review the `DisplayName` and `Description`? Thank you.

controller/utils.go
```diff
@@ -13,6 +15,73 @@ import (
 	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
 )
 
+type ExponentialBackoff struct {
+	interval           map[string]time.Duration
+	lastAttempt        map[string]time.Time
+	maxBackoffInterval time.Duration
+	mu                 sync.Mutex
+}
```
derekbit: @bachmanity1 To simplify the implemation and improve the maintainability, is it possible to replace the `ExponentialBackoff` implementation with client-go's `flowcontrol.Backoff`?

bachmanity1: As I mentioned earlier, it’s hard to decide when to reset the backoff because even when the pod creation request succeeds, the pod may still fail to start. We need to monitor the pod for some time to determine if it’s running, so that we can reset the backoff. The question is, how long should we monitor the pod?

types/setting.go
```suggestion
		DisplayName: "Maximum Pod Recreate Backoff",
		Description: "Specifies the maximum duration (in seconds) to wait between pod recreation attempts. After each failed attempt, the backoff interval doubles until it reaches this limit. The default value is 120 seconds.",
```

controller/utils.go
derekbit: @bachmanity1 Got your point. Then, the implementation makes sense.

controller/instance_manager_controller.go
```diff
@@ -190,6 +192,12 @@ func NewInstanceManagerController(
 	}
 	imc.cacheSyncs = append(imc.cacheSyncs, ds.SettingInformer.HasSynced)
 
+	maxPodRecreateBackoff, err := ds.GetSettingAsInt(types.SettingNameMaxPodRecreateBackoff)
+	if err != nil {
+		return nil, err
+	}
+	imc.exponentialBackoff = NewExponentialBackoff(maxPodRecreateBackoff)
```
derekbit: The `maxPodRecreateBackoff` in a controller's `exponentialBackoff` is configured when the controller is created. If the user updates this setting later, the `maxPodRecreateBackoff` in the `exponentialBackoff` will not be updated accordingly. The only method to apply the new setting value is deleting and recreating longhorn-manager pods . Can we improve this?

bachmanity1: Yes, good point. I'll work on it.

types/setting.go
```diff
@@ -1592,6 +1595,16 @@ var (
...
 	}
+
+	SettingDefinitionMaxPodRecreateBackoff = SettingDefinition{
+		DisplayName: "Maximum Pod Recreate Backoff",
+		Description: "Specifies the maximum duration (in seconds) to wait between pod recreation attempts. After each failed attempt, the backoff interval doubles until it reaches this limit. The default value is 120 seconds.",
+		Category:    SettingCategoryGeneral,
+		Type:        SettingTypeInt,
+		Required:    false,
+		ReadOnly:    false,
+		Default:     "120",
```
mantissahz:
```diff
+        ValueIntRange: map[string]int{
+			ValueIntRangeMinimum: 1,
+		},
```
I believe we do not allow a negative value here, right?

bachmanity1: yes, right. I've applied your suggestion.

controller/utils.go
```diff
@@ -13,6 +16,104 @@ import (
...
+
+type ExponentialBackoff struct {
+	interval    map[string]time.Duration
+	lastAttempt map[string]time.Time
+	mu          sync.Mutex
+}
+
+func init() {
+	maxBackoffMutex.RLock()
+	defer maxBackoffMutex.RUnlock()
```
c3y1huang:
```suggestion
	maxBackoffMutex.Lock()
	defer maxBackoffMutex.Unlock()
```

controller/utils.go
```diff
@@ -13,6 +16,104 @@ import (
 	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
 )
 
+var (
+	maxBackoff      int64
+	maxBackoffMutex sync.RWMutex
+)
+
+type ExponentialBackoff struct {
```
c3y1huang: NIT: This and the methods seems like good candidate to move into [go-common-libs](https://github.com/longhorn/go-common-libs)

controller/utils.go
```diff
@@ -2,6 +2,9 @@ package controller
 
 import (
 	"github.com/sirupsen/logrus"
+	"strconv"
+	"sync"
+	"time"
```
c3y1huang:
```suggestion
	"strconv"
	"sync"
	"time"

	"github.com/sirupsen/logrus"
```

controller/utils_test.go
```diff
@@ -0,0 +1,45 @@
+package controller
+
+import (
+	"github.com/stretchr/testify/assert"
+	"testing"
+	"time"
```
c3y1huang:
```suggestion
	"testing"
	"time"

	"github.com/stretchr/testify/assert"
```

c3y1huang (changes requested):

controller/utils.go
bachmanity1: I can move it to go-common-libs, but I think it will take some time for it to be merged and for the dependencies to be updated so it can be used here.

c3y1huang: That makes sense. I understand the extra steps involved. While it may take a bit more time upfront, moving it to go-common-libs would help simplify things in the long run and avoid duplication. 

@bachmanity1 If you are open to giving it a try, we can help review that PR first so it can be updated here. 

That said, no worries if you'd prefer not to.

controller/utils.go
bachmanity1: @derekbit  Reading the `flowcontrol.Backoff` docs again, I realized it actually does have the functionality I was talking about. We don't need to explicitly reset the backoff, we can just call [GC](https://github.com/kubernetes/client-go/blob/v0.33.2/util/flowcontrol/backoff.go#L149) periodically to clean up expired records. So with a small util function, we can get `flowcontrol.Backoff` to behave the way we want.

The only issue is that `flowcontrol.Backoff` doesn't support dynamically updating the max backoff. But I don't think many people would need to change the default anyway, so how about just making the max backoff a constant (like 2 minutes?) and just using `flowcontrol.Backoff` like you originally suggested?

bachmanity1: @derekbit should we just use `flowcontrol.Backoff` with fixed max backoff?

derekbit: > The only issue is that flowcontrol.Backoff doesn't support dynamically updating the max backoff. But I don't think many people would need to change the default anyway, so how about just making the max backoff a constant (like 2 minutes?) and just using flowcontrol.Backoff like you originally suggested?

Same thought. WDYT? @c3y1huang @innobead 

> @derekbit should we just use flowcontrol.Backoff with fixed max backoff?

Sounds good to me.

c3y1huang: Sounds good to me. Thanks @bachmanity1 .

controller/utils.go
```diff
@@ -2,6 +2,8 @@ package controller
 
 import (
 	"github.com/sirupsen/logrus"
+	"k8s.io/client-go/util/flowcontrol"
```
shuo-wu: NIT: Re-arrange the imports

shuo-wu (dismissed):
In general LGTM

controller/utils.go
```diff
@@ -2,8 +2,10 @@ package controller
 
 import (
 	"github.com/sirupsen/logrus"
+	"time"
 
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/util/flowcontrol"
```
derekbit:
```suggestion
	"time"

	"github.com/sirupsen/logrus"

	"k8s.io/apimachinery/pkg/runtime"
	"k8s.io/client-go/util/flowcontrol"
```

controller/utils.go
```diff
@@ -13,6 +16,28 @@ import (
 	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
 )
 
+const (
+	podRecreateInitBackoff = 1 * time.Second
+	podRecreateMaxBackoff  = 120 * time.Second
+	backoffGCPeriod        = 12 * time.Hour
```
derekbit, line 22: Is it too long for GC?

bachmanity1: GC is needed to remove expired entries, but it’s not critical to remove them right away. If an expired entry is updated, it’s treated as a new one and the backoff starts from 1 second. So GC mainly helps keep the number of entries from growing too much. I think running it once every 12 hours is reasonable.

shuo-wu: There should not be too many different IM/SM/BIM pod failures accumulated hence this GC interval is enough. 
I mean, Longhorn may retry pod creation multiple times for some **specific** pods rather than using **different namings** in each retry.

derekbit: It is clear. Thanks @bachmanity1 and @shuo-wu.
@shuo-wu Can you approve the PR if there is no other issue? Thank you.

shuo-wu (approved):

shuo-wu:
@derekbit It seems that @c3y1huang and @sushant-suse need to approve this PR before merging.

controller/utils.go
```diff
@@ -13,6 +16,28 @@ import (
...
+)
+
+// newBackoff returns a flowcontrol.Backoff and starts a background GC loop.
```
c3y1huang, line 25: NIT: It might be worth mentioning the potential risk if this function is ever reused in a different context later. 
```suggestion
// newBackoff returns a flowcontrol.Backoff and starts a background GC loop.
//
// This function starts a background goroutine that cannot be gracefully stopped,
// making it suitable only for processes that run for the entire application's
// lifetime. Using it in short-lived or frequently called contexts will lead to
// goroutine leaks. For such cases, refactor to accept a context.Context for proper
// cleanup.
```

c3y1huang (approved):

controller/utils.go
```diff
@@ -13,6 +16,28 @@ import (
...
+	go func() {
+		ticker := time.NewTicker(backoffGCPeriod)
+		defer ticker.Stop()
+		for range ticker.C {
+			backoff.GC()
+		}
+	}()
+
+	return backoff
+}
```
c3y1huang, line 39: For improvement.

NIT: It might be worth mentioning the potential risk if this function is ever reused in a different context later. 
```go
// newBackoff returns a flowcontrol.Backoff and starts a background GC loop.
//
// This function starts a background goroutine that cannot be gracefully stopped,
// making it suitable only for processes that run for the entire application's
// lifetime. Using it in short-lived or frequently called contexts will lead to
// goroutine leaks. For such cases, refactor to accept a context.Context for proper
// cleanup.
```

c3y1huang: Or refactor it. Example:
```go
func newBackoff(ctx context.Context) *flowcontrol.Backoff {
	backoff := flowcontrol.NewBackOff(podRecreateInitBackoff, podRecreateMaxBackoff)

	go func() {
		ticker := time.NewTicker(backoffGCPeriod)
		defer ticker.Stop()
		for {
			select {
			case <-ctx.Done():
				return
			case <-ticker.C:
				backoff.GC()
			}
		}
	}()

	return backoff
}
```

derekbit: The PR is merged by Mergify.
@bachmanity1 Could you help create a PR to address @c3y1huang's comment? Thank you.

c3y1huang: @bachmanity1 
For now, I think adding a comment should be sufficient. Since Mergify already merged this automatically :sweat_smile: , I suggest opening another PR for improvement.

cc @derekbit

PR #3884, opened 2025-06-30
feat: add error in the engine image condition

derekbit:

Issue Longhorn/longhorn#7022

To improve troubleshooting, add error in the engine image condition.


derekbit:
Closed. Wrong branch.

PR #3892, opened 2025-07-03
feat(types): add support for floating point values for guaranteed-ins…

gigabyte132:

Issue #11179

https://github.com/longhorn/longhorn/issues/11179

This PR adds support for setting floating point values for guaranteed-instance-manager-cpu.


derekbit:
@gigabyte132 Thanks for the contribution.
This PR updates the data type of the global setting guaranteed-instance-manager-cpu. The change is valid and should not cause any side effects.

However, Longhorn also supports a per-node InstanceManagerCPURequest ([definition](https://github.com/longhorn/longhorn-manager/blob/master/k8s/pkg/apis/longhorn/v1beta2/node.go#L152)). For consistency and simplicity, should we keep this as an integer?
If we decide to change it to a float, we'll need to consider compatibility handling.

IMO, we can improve the global setting first and update the `node.spec.InstanceManagerCPURequest` in the future.

WDYT? @gigabyte132 @shuo-wu @PhanLe1010 @c3y1huang @innobead

types/types.go
```diff
@@ -1243,19 +1243,25 @@ func CreateDefaultDisk(dataPath string, storageReservedPercentage int64) (map[st
...
-		if isUnderLimit {
-			return fmt.Errorf("invalid requested instance manager CPUs. Valid instance manager CPU range is larger than %v millicpu", valueIntRange[ValueIntRangeMinimum])
+	case SettingNameV2DataEngineGuaranteedInstanceManagerCPU:
+		instanceManagerCPU, err := strconv.Atoi(instanceManagerCPUStr)
+		if err != nil {
+			return errors.Wrapf(err, "invalid requested instance manager CPU millicpu value (%v)", instanceManagerCPUStr)
+		}
+
+		if instanceManagerCPU != 0 && instanceManagerCPU < 1000 {
+			return fmt.Errorf("invalid requested instance manager CPU millicpu. Valid range is 0 or >= 1000, got %v", instanceManagerCPU)
```
derekbit: Can we use the sample implementation for `SettingNameV2DataEngineGuaranteedInstanceManagerCPU`?

```
		isUnderLimit := instanceManagerCPU < valueIntRange[ValueIntRangeMinimum]
		if isUnderLimit {
			return fmt.Errorf("invalid requested instance manager CPUs. Valid instance manager CPU range is larger than %v millicpu", valueIntRange[ValueIntRangeMinimum])
```

gigabyte132:
> @gigabyte132 Thanks for the contribution. This PR updates the data type of the global setting guaranteed-instance-manager-cpu. The change is valid and should not cause any side effects.
> 
> However, Longhorn also supports a per-node InstanceManagerCPURequest ([definition](https://github.com/longhorn/longhorn-manager/blob/master/k8s/pkg/apis/longhorn/v1beta2/node.go#L152)). For consistency and simplicity, should we keep this as an integer? If we decide to change it to a float, we'll need to consider compatibility handling.
> 
> IMO, we can improve the global setting first and update the `node.spec.InstanceManagerCPURequest` in the future.
> 
> WDYT? @gigabyte132 @shuo-wu @PhanLe1010 @c3y1huang @innobead

I would still like to be able to set float values as a general configuration for the whole cluster but it is your decision.

types/types.go
gigabyte132: Done, should look more similar to the previous implementation

derekbit:
> > @gigabyte132 Thanks for the contribution. This PR updates the data type of the global setting guaranteed-instance-manager-cpu. The change is valid and should not cause any side effects.
> > However, Longhorn also supports a per-node InstanceManagerCPURequest ([definition](https://github.com/longhorn/longhorn-manager/blob/master/k8s/pkg/apis/longhorn/v1beta2/node.go#L152)). For consistency and simplicity, should we keep this as an integer? If we decide to change it to a float, we'll need to consider compatibility handling.
> > IMO, we can improve the global setting first and update the `node.spec.InstanceManagerCPURequest` in the future.
> > WDYT? @gigabyte132 @shuo-wu @PhanLe1010 @c3y1huang @innobead
> 
> I would still like to be able to set float values as a general configuration for the whole cluster but it is your decision.

@gigabyte132 
Do you mean you'd like to make per-node `node.spec.InstanceManagerCPURequest` as float value as well?

gigabyte132:
> > > @gigabyte132 Thanks for the contribution. This PR updates the data type of the global setting guaranteed-instance-manager-cpu. The change is valid and should not cause any side effects.
> > > However, Longhorn also supports a per-node InstanceManagerCPURequest ([definition](https://github.com/longhorn/longhorn-manager/blob/master/k8s/pkg/apis/longhorn/v1beta2/node.go#L152)). For consistency and simplicity, should we keep this as an integer? If we decide to change it to a float, we'll need to consider compatibility handling.
> > > IMO, we can improve the global setting first and update the `node.spec.InstanceManagerCPURequest` in the future.
> > > WDYT? @gigabyte132 @shuo-wu @PhanLe1010 @c3y1huang @innobead
> > 
> > 
> > I would still like to be able to set float values as a general configuration for the whole cluster but it is your decision.
> 
> @gigabyte132 Do you mean you'd like to make per-node `node.spec.InstanceManagerCPURequest` as float value as well?

@derekbit Sorry I think I misread your first comment about this, at least for now I don't think that's necessary for us at least to set `node.spec.InstanceManagerCPURequest` as a float at least. Sorry again for the confusion

derekbit:
> > > > @gigabyte132 Thanks for the contribution. This PR updates the data type of the global setting guaranteed-instance-manager-cpu. The change is valid and should not cause any side effects.
> > > > However, Longhorn also supports a per-node InstanceManagerCPURequest ([definition](https://github.com/longhorn/longhorn-manager/blob/master/k8s/pkg/apis/longhorn/v1beta2/node.go#L152)). For consistency and simplicity, should we keep this as an integer? If we decide to change it to a float, we'll need to consider compatibility handling.
> > > > IMO, we can improve the global setting first and update the `node.spec.InstanceManagerCPURequest` in the future.
> > > > WDYT? @gigabyte132 @shuo-wu @PhanLe1010 @c3y1huang @innobead
> > > 
> > > 
> > > I would still like to be able to set float values as a general configuration for the whole cluster but it is your decision.
> > 
> > 
> > @gigabyte132 Do you mean you'd like to make per-node `node.spec.InstanceManagerCPURequest` as float value as well?
> 
> @derekbit Sorry I think I misread your first comment about this, at least for now I don't think that's necessary for us at least to set `node.spec.InstanceManagerCPURequest` as a float at least. Sorry again for the confusion

Thanks for the clarification! We have the same thought. 
For the `node.spec.InstanceManagerCPURequest`, we can improve it in the future.

The PR LGTM. Thanks for the contribution!

PR #3893, opened 2025-07-03
fix(backup): delete duplicated backup volumes

mantissahz:

Issue # longhorn/longhorn#11154

1. Delete duplicated backup volumes if the volume name or backup target name is empty in `BackupVolume.Spec`.
2. Add a backup volume validator to prevent backup volume CRs from the empty volume name or empty backup target name in `BackupVolume.Spec`.


webhook/resources/backupvolume/validator.go
```diff
@@ -18,13 +23,48 @@ func NewValidator(ds *datastore.DataStore) admission.Validator {
...
+	return nil
+}
+
+func checkEmptyBackupTargetVolumeName(backupVolume *longhorn.BackupVolume) error {
+	if backupVolume.Spec.BackupTargetName == "" {
+		return fmt.Errorf("backup target name can not be empty")
+	}
+
+	if backupVolume.Spec.VolumeName == "" {
+		return fmt.Errorf("volume name can not be empty")
```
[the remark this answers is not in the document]
mantissahz: Modified. Thanks

controller/backup_target_controller.go
```diff
@@ -633,14 +656,50 @@ func (btc *BackupTargetController) cleanupBackupVolumeNotExistOnBackupTarget(clu
...
+	return nil
+}
+
+func (btc *BackupTargetController) deleteOnlyBackupVolumeCR(backupVolumeName string, log logrus.FieldLogger) (err error) {
+
+	if err = datastore.AddBackupVolumeDeleteCustomResourceOnlyLabel(btc.ds, backupVolumeName); err != nil {
+		return errors.Wrapf(err, "failed to add label delete-custom-resource-only to Backupvolume %s", backupVolumeName)
+	}
+	if err = btc.ds.DeleteBackupVolume(backupVolumeName); err != nil {
+		return err
```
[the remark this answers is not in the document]
mantissahz: Modified. Thanks

webhook/resources/backupvolume/validator.go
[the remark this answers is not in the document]
mantissahz: Modified. Thanks

derekbit:
@mantissahz Could you elaborate more on the root cause of the issue (RCA)?

mantissahz:
>@mantissahz Could you elaborate more on the root cause of the issue (RCA)?

Actually, I did not find out when the invalid backup volume CRs are created, and the root cause should be similar to what @shuo-wu mentioned.
1. The split-brain occurred
2. Two backup target controllers were responsible for reconciling the backup target
3. One (B) of the controllers got the empty backup volume list
4. B tried to create backup volume CRs with new convention names to synchronize the backup information
5. And we got the duplicated backup volumes

I can not figure out why the "BackupVolume.Spec.VolumeName" and "BackupVolume.Spec.BackupTargetName" are empty, because no logs tell how these invalid backup volume CRs are created.

types/types.go
```diff
@@ -778,8 +778,8 @@ func GetShareManagerNameFromShareManagerPodName(podName string) string {
 	return strings.TrimPrefix(podName, shareManagerPrefix)
 }
 
-func GetBackupVolumeNameFromVolumeName(volumeName string) string {
-	return volumeName + "-" + util.RandomID()
+func GetBackupVolumeNameFromVolumeName(volumeName, backupTargetName string) string {
+	return volumeName + "-" + util.GetStringChecksumSHA256(strings.TrimSpace(fmt.Sprintf("%s-%s", volumeName, backupTargetName)))[:util.RandomIDLength]
```
PhanLe1010, line 783: Oh. I like this change!

Does it mean the previous implementation does not hash the backuptarget name?

PhanLe1010 (approved):
LGTM

types/types.go
mantissahz: Yeah, we concatenated the volume name and an 8-bit random ID.

shuo-wu (commented):
The modification of this PR is mainly for `syncBackupVolume`. But the issue is, when the **backup target is unavailable** (line 436 - 447), does this controller really have a chance to execute `syncBackupVolume` (line 466)?

If the backup target is reachable, the syncup between remote BV and cluster CRs will eliminate all duplications without this PR

mantissahz:
>The modification of this PR is mainly for syncBackupVolume. But the issue is, when the backup target is unavailable (line 436 - 447), does this controller really have a chance to execute syncBackupVolume (line 466)?

If `backupTarget.Spec.BackupTargetURL` is empty (line 436-447), it will start to delete backup volume CRs and backup backing image CRs in method `cleanUpAllBackupRelatedResources`.

We might need to be concerned that if we also need to clean up them when getting backup information fails (line 450-458, it will mark the backup target available as false, like line 436-477)

WDYT @shuo-wu, @derekbit, @PhanLe1010 

>If the backup target is reachable, the syncup between remote backup metadata and cluster CRs will eliminate all duplications without this PR

Yes, but it might be a long process because we only delete a backup volume CR where `Spec.VolumeName` is empty in a reconciliation (backup poll interval).

shuo-wu:
> > If backupTarget.Spec.BackupTargetURL is empty (line 436-447), it will start to delete backup volume CRs and backup backing image CRs in method cleanUpAllBackupRelatedResources.

> We might need to be concerned that if we also need to clean up them when getting backup information fails (line 450-458, it will mark the backup target available as false, like line 436-477)

YES. An empty URL is not the only way to invalidate a backup target. The essence of this fix is, Longhorn is always able to clean up invalid/duplicate BackupVolume CRs, no matter what status the backup target is in.

> > If the backup target is reachable, the syncup between remote backup metadata and cluster CRs will eliminate all duplications without this PR

> Yes, but it might be a long process because we only delete a backup volume CR where Spec.VolumeName is empty in a reconciliation (backup poll interval).

Right. It depends on the sync-up interval setting. But this auto sync-up mechanism can resolve the invalid BackupVolume anyway. The cleanup you added makes things happen quicker, which is nice but does not hit the critical.

mantissahz:
>YES. An empty URL is not the only way to invalidate a backup target. The essence of this fix is, Longhorn is always able to clean up invalid/duplicate BackupVolume CRs, no matter what status the backup target is in.

I can try to handle the duplicated backup volumes before line 436, and that might be the fix, right?
WDYT @shuo-wu?

derekbit:

shuo-wu:
> I can try to handle the duplicated backup volumes before line 436, and that might be the fix, right?

I am OK with any implementations as long as they resolve the issue I mentioned above.

derekbit:
> Delete duplicated backup volumes if the volume name or backup target name is empty in BackupVolume.Spec.
> Add a backup volume validator to prevent backup volume CRs from the empty volume name or empty backup target name in BackupVolume.Spec.

Adding a validating webhook or removing duplicated backup volumes would be nice to have, as they serve as mitigation and prevention measures.
Theoretically, backupVolume's volume name or backupTarget name shouldn't be empty, isn't it? Could you help point me to the logic that creates a backupVolume with an empty volume name or backupTarget name?

mantissahz:
>Theoretically, backupVolume's volume name or backupTarget name shouldn't be empty, isn't it? Could you help point me to the logic that creates a backupVolume with an empty volume name or backupTarget name?

Yes, they should not be empty.

For creating a BackupVolume by Longhorn:
https://github.com/longhorn/longhorn-manager/blob/a5b6285f66182d7fb9c2aa76e28bb9f78b930d40/controller/backup_target_controller.go#L598-L617
We only create BackupVolumes at the backup target controller, and we will set backupTarget and volume labels to be correct values as we set `Spec.BackupTargetName` and `Spec.VolumeName`, and in the support bundle, we can see the BackupVolume with an empty `Spec.VolumeName` has a valid volume label.

For upgrading from 1.7.x:
We will only upgrade the old backup volume CRs, and we can see that almost all backup volumes with an empty `Spec.VolumeName` have a suffix in the BackupVolume name.

For updating a BackupVolume by Longhorn:
1. In the backup target controller, we only modify the `Spec.SyncRequestedAt`
https://github.com/longhorn/longhorn-manager/blob/a5b6285f66182d7fb9c2aa76e28bb9f78b930d40/controller/backup_target_controller.go#L583-L588
1. In the backup controller, we only modify the `Spec.SyncRequestedAt`
https://github.com/longhorn/longhorn-manager/blob/a5b6285f66182d7fb9c2aa76e28bb9f78b930d40/controller/backup_controller.go#L323-L354
https://github.com/longhorn/longhorn-manager/blob/a5b6285f66182d7fb9c2aa76e28bb9f78b930d40/controller/backup_controller.go#L945-L951
1. To trigger synchronizing a BackupVolume, we only modify the `Spec.SyncRequestedAt`
https://github.com/longhorn/longhorn-manager/blob/a5b6285f66182d7fb9c2aa76e28bb9f78b930d40/manager/engine.go#L452-L459
1. To trigger synchronizing a BackupVolume, we only modify the `Spec.SyncRequestedAt`
https://github.com/longhorn/longhorn-manager/blob/a5b6285f66182d7fb9c2aa76e28bb9f78b930d40/manager/volume.go#L445-L459

webhook/resources/backupvolume/validator.go
```diff
@@ -18,13 +23,48 @@ func NewValidator(ds *datastore.DataStore) admission.Validator {
...
+		APIVersion: longhorn.SchemeGroupVersion.Version,
+		ObjectType: &longhorn.BackupVolume{},
+		OperationTypes: []admissionregv1.OperationType{
+			admissionregv1.Create,
+			admissionregv1.Update,
+		},
+	}
+}
+
+func (bv *backupVolumeValidator) Create(request *admission.Request, newObj runtime.Object) error {
```
derekbit: nit:
```suggestion
func (bv *backupVolumeValidator) Create(request *admission.Request, obj runtime.Object) error {
```

webhook/resources/backupvolume/validator.go
```diff
@@ -18,13 +23,48 @@ func NewValidator(ds *datastore.DataStore) admission.Validator {
...
+	backupVolume := newObj.(*longhorn.BackupVolume)
+
+	if err := checkEmptyBackupTargetVolumeName(backupVolume); err != nil {
+		return werror.NewInvalidError(err.Error(), "")
+	}
+
+	return nil
+}
+
+func checkEmptyBackupTargetVolumeName(backupVolume *longhorn.BackupVolume) error {
```
derekbit:
```suggestion
func validateRequiredFields(backupVolume *longhorn.BackupVolume) error {
```

webhook/resources/backupvolume/validator.go
```diff
@@ -18,13 +23,48 @@ func NewValidator(ds *datastore.DataStore) admission.Validator {
...
+
+	if backupVolume.Spec.VolumeName == "" {
+		return fmt.Errorf("volume name cannot be empty")
 	}
+
+	return nil
+}
+
+func (bv *backupVolumeValidator) Update(request *admission.Request, oldObj runtime.Object, newObj runtime.Object) error {
+	backupVolume := newObj.(*longhorn.BackupVolume)
```
derekbit:
```suggestion
	newBackupVolume := newObj.(*longhorn.BackupVolume)
```

types/types.go
```diff
@@ -778,8 +778,8 @@ func GetShareManagerNameFromShareManagerPodName(podName string) string {
 	return strings.TrimPrefix(podName, shareManagerPrefix)
 }
 
-func GetBackupVolumeNameFromVolumeName(volumeName string) string {
-	return volumeName + "-" + util.RandomID()
+func GetBackupVolumeNameFromVolumeName(volumeName, backupTargetName string) string {
```
derekbit, line 781: What's the purpose of the change?

controller/backup_target_controller.go
```diff
@@ -633,14 +670,50 @@ func (btc *BackupTargetController) cleanupBackupVolumeNotExistOnBackupTarget(clu
...
+		if err := btc.deleteOnlyBackupVolumeCR(bvName, log); err != nil {
+			if apierrors.IsNotFound(err) {
+				continue
+			}
+			multiError.Append(util.NewMultiError(fmt.Sprintf("%v: %v", bvName, err)))
 		}
 	}
 
+	if len(multiError) > 0 {
+		return fmt.Errorf("failed to delete backup volumes from cluster: %v", multiError.Join())
```
derekbit:
```suggestion
		return fmt.Errorf("failed to delete backup volumes: %v", multiError.Join())
```

controller/backup_target_controller.go
```diff
@@ -546,22 +553,48 @@ func (btc *BackupTargetController) getInfoFromBackupStore(backupTarget *longhorn
...
-		return err
+		return nil, err
 	}
 
+	deletingDuplicateBackupVolumeSet := sets.New[string]()
 	clusterVolumeBVMap := make(map[string]*longhorn.BackupVolume, len(clusterBackupVolumes))
-	clusterBackupVolumesSet := sets.New[string]()
 	for _, bv := range clusterBackupVolumes {
-		clusterBackupVolumesSet.Insert(bv.Spec.VolumeName)
+		if bv.Spec.VolumeName == "" || bv.Spec.BackupTargetName == "" || bv.Spec.BackupTargetName != backupTargetName {
```
derekbit: Is it possible that all the backupVolumes' `spec.VolumeName` or `spec.BackupTargetName` is empty?
What will happen if you delete all the backupVolumes?

controller/backup_target_controller.go
```diff
@@ -546,22 +553,48 @@ func (btc *BackupTargetController) getInfoFromBackupStore(backupTarget *longhorn
...
+			log.WithField("backupVolume", bv.Name).Debugf("Volume name or backup target name is empty or different backup target name %v in BackupVolume, deleting it", bv.Spec.BackupTargetName)
```
derekbit: Need to refactor the if codes and make the log messages more granular to improve the troubleshooting capability.
```suggestion
		if bv.Spec.VolumeName == "" || bv.Spec.BackupTargetName == "" || bv.Spec.BackupTargetName != backupTargetName {
			log.WithField("backupVolume", bv.Name).Warn("Found volume name is empty") bv.Spec.BackupTargetName)
		} elseif bv.Spec.BackupTargetName == "" {
			log.WithField("backupVolume", bv.Name).Warn("Found volume name is empty") 
		} else if bv.Spec.BackupTargetName != backupTargetName {
			log....
		}
```

controller/backup_target_controller.go
```diff
@@ -546,22 +553,48 @@ func (btc *BackupTargetController) getInfoFromBackupStore(backupTarget *longhorn
 	return info, nil
 }
 
-func (btc *BackupTargetController) syncBackupVolume(backupTarget *longhorn.BackupTarget, backupStoreBackupVolumeNames []string, syncTime metav1.Time, log logrus.FieldLogger) error {
-	backupStoreBackupVolumes := sets.New[string](backupStoreBackupVolumeNames...)
-
+func (btc *BackupTargetController) cleanupDuplicateBackupVolumes(backupTargetName string, log logrus.FieldLogger) (map[string]*longhorn.BackupVolume, error) {
```
derekbit:
```suggestion
func (btc *BackupTargetController) cleanupDuplicateBackupVolumes(backupTargetName string) (map[string]*longhorn.BackupVolume, error) {
```

controller/backup_target_controller.go
```diff
@@ -633,14 +670,50 @@ func (btc *BackupTargetController) cleanupBackupVolumeNotExistOnBackupTarget(clu
...
+	if err = datastore.AddBackupVolumeDeleteCustomResourceOnlyLabel(btc.ds, backupVolumeName); err != nil {
+		return errors.Wrapf(err, "failed to add label delete-custom-resource-only to Backupvolume %s", backupVolumeName)
+	}
+	if err = btc.ds.DeleteBackupVolume(backupVolumeName); err != nil {
+		return errors.Wrapf(err, "failed to delete BackupVolume %s", backupVolumeName)
+	}
+	return nil
+}
+
+func (btc *BackupTargetController) deleteDuplicateBackupVolumeForBackupTarget(duplicateBackupVolumesSet sets.Set[string], log logrus.FieldLogger) (err error) {
```
derekbit:
```suggestion
func (btc *BackupTargetController) deleteDuplicateBackupVolumeForBackupTarget(duplicateBackupVolumesSet sets. Set[string]) (err error) {
```

controller/backup_target_controller.go
```diff
@@ -546,22 +553,48 @@ func (btc *BackupTargetController) getInfoFromBackupStore(backupTarget *longhorn
...
 	// Get a list of all the backup volumes that exist as custom resources in the cluster
-	clusterBackupVolumes, err := btc.ds.ListBackupVolumesWithBackupTargetNameRO(backupTarget.Name)
+	clusterBackupVolumes, err := btc.ds.ListBackupVolumesWithBackupTargetNameRO(backupTargetName)
 	if err != nil {
-		return err
+		return nil, err
 	}
 
+	deletingDuplicateBackupVolumeSet := sets.New[string]()
 	clusterVolumeBVMap := make(map[string]*longhorn.BackupVolume, len(clusterBackupVolumes))
```
derekbit: Why do we need to prefix `cluster`?
```suggestion
	clusterVolumeBVMap := make(map[string]*longhorn.BackupVolume, len(clusterBackupVolumes))
```

controller/backup_target_controller.go
```diff
@@ -546,22 +553,48 @@ func (btc *BackupTargetController) getInfoFromBackupStore(backupTarget *longhorn
...
 	for _, bv := range clusterBackupVolumes {
-		clusterBackupVolumesSet.Insert(bv.Spec.VolumeName)
+		if bv.Spec.VolumeName == "" || bv.Spec.BackupTargetName == "" || bv.Spec.BackupTargetName != backupTargetName {
+			log.WithField("backupVolume", bv.Name).Debugf("Volume name or backup target name is empty or different backup target name %v in BackupVolume, deleting it", bv.Spec.BackupTargetName)
+			deletingDuplicateBackupVolumeSet.Insert(bv.Name)
+			continue
+		}
+		if existingBV, exists := clusterVolumeBVMap[bv.Spec.VolumeName]; exists {
+			if existingBV.CreationTimestamp.Before(&bv.CreationTimestamp) {
+				log.WithField("backupVolume", bv.Name).Debugf("Found duplicated BackupVolume with volume name %s, deleting it", bv.Spec.VolumeName)
```
derekbit:
```suggestion
			if existingBV.CreationTimestamp.Before(&bv.CreationTimestamp) {
				log.WithField("backupVolume", bv.Name).Warnf("Found duplicated BackupVolume with volume name %s", bv.Spec.VolumeName)
```

controller/backup_target_controller.go
```diff
@@ -546,22 +553,48 @@ func (btc *BackupTargetController) getInfoFromBackupStore(backupTarget *longhorn
...
+				deletingDuplicateBackupVolumeSet.Insert(bv.Name)
+				continue
+			}
+			log.WithField("backupVolume", existingBV.Name).Debugf("Found duplicated BackupVolume with volume name %s, deleting it", existingBV.Spec.VolumeName)
```
derekbit:
```suggestion
			log.WithField("backupVolume", existingBV.Name).Warnf("Found duplicated BackupVolume with volume name %s", existingBV.Spec.VolumeName)
```

controller/backup_target_controller.go
```diff
@@ -546,22 +553,48 @@ func (btc *BackupTargetController) getInfoFromBackupStore(backupTarget *longhorn
...
+			deletingDuplicateBackupVolumeSet.Insert(existingBV.Name)
+		}
 		clusterVolumeBVMap[bv.Spec.VolumeName] = bv
 	}
 
+	if err := btc.deleteDuplicateBackupVolumeForBackupTarget(deletingDuplicateBackupVolumeSet, log); err != nil {
+		log.WithError(err).Error("Failed to clean up backup volumes that have duplicate volume name in cluster")
```
derekbit:
```suggestion
		log.WithError(err).Error("Failed to clean up duplicated backup volumes")
```

controller/backup_target_controller.go
```diff
@@ -546,22 +553,48 @@ func (btc *BackupTargetController) getInfoFromBackupStore(backupTarget *longhorn
 	return info, nil
 }
 
-func (btc *BackupTargetController) syncBackupVolume(backupTarget *longhorn.BackupTarget, backupStoreBackupVolumeNames []string, syncTime metav1.Time, log logrus.FieldLogger) error {
-	backupStoreBackupVolumes := sets.New[string](backupStoreBackupVolumeNames...)
-
+func (btc *BackupTargetController) cleanupDuplicateBackupVolumes(backupTargetName string, log logrus.FieldLogger) (map[string]*longhorn.BackupVolume, error) {
 	// Get a list of all the backup volumes that exist as custom resources in the cluster
-	clusterBackupVolumes, err := btc.ds.ListBackupVolumesWithBackupTargetNameRO(backupTarget.Name)
+	clusterBackupVolumes, err := btc.ds.ListBackupVolumesWithBackupTargetNameRO(backupTargetName)
```
derekbit: What do we need the prefix `cluster`?
```suggestion
	clusterBackupVolumes, err := btc.ds.ListBackupVolumesWithBackupTargetNameRO(backupTargetName)
```

controller/backup_target_controller.go
```diff
@@ -433,6 +433,13 @@ func (btc *BackupTargetController) reconcile(name string) (err error) {
 		}
 	}()
 
+	// clean up invalid backup volumes that are created during split-brain
+	// https://github.com/longhorn/longhorn/issues/11154
+	clusterVolumeBVMap, err := btc.cleanupDuplicateBackupVolumes(backupTarget.Name, log)
```
derekbit: What's the period of the cleanupDuplicateBackupVolumes? Will it lead to high system load?

controller/backup_target_controller.go
```diff
@@ -546,22 +553,48 @@ func (btc *BackupTargetController) getInfoFromBackupStore(backupTarget *longhorn
...
+func (btc *BackupTargetController) cleanupDuplicateBackupVolumes(backupTargetName string, log logrus.FieldLogger) (map[string]*longhorn.BackupVolume, error) {
 	// Get a list of all the backup volumes that exist as custom resources in the cluster
-	clusterBackupVolumes, err := btc.ds.ListBackupVolumesWithBackupTargetNameRO(backupTarget.Name)
+	clusterBackupVolumes, err := btc.ds.ListBackupVolumesWithBackupTargetNameRO(backupTargetName)
 	if err != nil {
-		return err
+		return nil, err
 	}
 
+	deletingDuplicateBackupVolumeSet := sets.New[string]()
```
derekbit: Can we name `deletingDuplicateBackupVolumeSet` to `duplicateBackupVolumeSet`?

derekbit (commented):
I will review the PR again after addressing the comments.

controller/backup_target_controller.go
mantissahz: Modified. Thanks.

controller/backup_target_controller.go
mantissahz: >What's the period of the cleanupDuplicateBackupVolumes?

As the `backupstore-poll-interval` setting and the default value is 300 seconds, 
or customized value in the `BackupTarget.Spec.PollInterval`.

controller/backup_target_controller.go
mantissahz: Yes, we don't need it. Modified. Thanks.

controller/backup_target_controller.go
mantissahz: Modified. Thanks.

controller/backup_target_controller.go
mantissahz: Modified. Thanks.

controller/backup_target_controller.go
mantissahz: Modified. Thanks.

controller/backup_target_controller.go
mantissahz: Yes, we don't need it. Modified. Thanks.

controller/backup_target_controller.go
mantissahz: Modified. Thanks.

controller/backup_target_controller.go
mantissahz: Modified. Thanks.

controller/backup_target_controller.go
mantissahz: Modified. Thanks.

controller/backup_target_controller.go
mantissahz: Modified. Thanks.

webhook/resources/backupvolume/validator.go
mantissahz: Modified. Thanks.

webhook/resources/backupvolume/validator.go
mantissahz: Modified. Thanks.

webhook/resources/backupvolume/validator.go
mantissahz: Modified. Thanks.

types/types.go
mantissahz: It should be safer, as we might create multiple BVs for the same volume and backup target if we encounter issues such as a split-brain.

controller/backup_target_controller.go
mantissahz: >Is it possible that all the backupVolumes' spec.VolumeName or spec.BackupTargetName is empty?

I think it might be possible because we still cannot figure out why invalid BackupVolumes had an empty volume name and an empty backup target name

>What will happen if you delete all the backupVolumes?

We only delete the backup volume CRs, and data will still be stored in the backup store.
We can still synchronize the backup information at the next poll interval.

types/types.go
derekbit: Can you add a comment?

derekbit:
@mantissahz Please help resolve the comments from copilot. Thanks.

controller/backup_target_controller.go
```diff
@@ -546,20 +553,59 @@ func (btc *BackupTargetController) getInfoFromBackupStore(backupTarget *longhorn
...
+			log.WithField("backupVolume", bv.Name).Debugf("Backup target name is empty in BackupVolume")
+			duplicateBackupVolumeSet.Insert(bv.Name)
+			continue
+		}
+		if bv.Spec.VolumeName == "" {
+			log.WithField("backupVolume", bv.Name).Debugf("Volume name is empty in BackupVolume")
+			duplicateBackupVolumeSet.Insert(bv.Name)
+			continue
+		}
+		if bv.Spec.BackupTargetName != backupTargetName {
```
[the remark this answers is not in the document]
mantissahz: The method `ListBackupVolumesWithBackupTargetNameRO` returns the backup volume list with labels that match the backup target name. As `bv.Spec.BackupTargetName` might be empty, it might be different from the backup target we are reconciling.

controller/backup_target_controller.go
```diff
@@ -546,20 +553,59 @@ func (btc *BackupTargetController) getInfoFromBackupStore(backupTarget *longhorn
...
 
-func (btc *BackupTargetController) syncBackupVolume(backupTarget *longhorn.BackupTarget, backupStoreBackupVolumeNames []string, syncTime metav1.Time, log logrus.FieldLogger) error {
-	backupStoreBackupVolumes := sets.New[string](backupStoreBackupVolumeNames...)
+func (btc *BackupTargetController) cleanupDuplicateBackupVolumes(backupTarget *longhorn.BackupTarget) (map[string]*longhorn.BackupVolume, error) {
+	log := getLoggerForBackupTarget(btc.logger, backupTarget)
+	backupTargetName := backupTarget.Name
 
 	// Get a list of all the backup volumes that exist as custom resources in the cluster
-	clusterBackupVolumes, err := btc.ds.ListBackupVolumesWithBackupTargetNameRO(backupTarget.Name)
+	backupVolumeList, err := btc.ds.ListBackupVolumesWithBackupTargetNameRO(backupTargetName)
```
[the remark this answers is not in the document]
mantissahz: We only handle backup volumes for the backup target that is reconciling.

controller/backup_target_controller.go
```diff
@@ -633,14 +683,51 @@ func (btc *BackupTargetController) cleanupBackupVolumeNotExistOnBackupTarget(clu
...
+	if len(multiError) > 0 {
+		return fmt.Errorf("failed to delete backup volumes from cluster: %v", multiError.Join())
+	}
+	return nil
+}
+
+func (btc *BackupTargetController) deleteOnlyBackupVolumeCR(backupVolumeName string, log logrus.FieldLogger) (err error) {
+
+	if err = datastore.AddBackupVolumeDeleteCustomResourceOnlyLabel(btc.ds, backupVolumeName); err != nil {
+		return errors.Wrapf(err, "failed to add label delete-custom-resource-only to Backupvolume %s", backupVolumeName)
```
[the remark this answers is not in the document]
mantissahz: Modified. Thanks

controller/backup_target_controller.go
derekbit: Will it lead to high system load?

controller/backup_target_controller.go
derekbit: Please add a test case in https://github.com/longhorn/longhorn/issues/11154#issuecomment-3031699450 for QA to verify the deletion is safe.

cc @longhorn/qa

controller/backup_target_controller.go
```diff
@@ -433,6 +433,13 @@ func (btc *BackupTargetController) reconcile(name string) (err error) {
 		}
 	}()
 
+	// clean up invalid backup volumes that are created during split-brain
+	// https://github.com/longhorn/longhorn/issues/11154
+	clusterVolumeBVMap, err := btc.cleanupDuplicateBackupVolumes(backupTarget)
```
derekbit: Why do we need the prefix `cluster`?
```suggestion
	clusterVolumeBVMap, err := btc.cleanupDuplicateBackupVolumes(backupTarget)
```

controller/backup_target_controller.go
mantissahz: >Will it lead to high system load?

I beleive No.

It should work as the previous backup target reconciliation because I move the listing cluster backup volume CRs into `cleanupDuplicateBackupVolumes`.

controller/backup_target_controller.go
```diff
@@ -546,20 +553,59 @@ func (btc *BackupTargetController) getInfoFromBackupStore(backupTarget *longhorn
...
+	}
+
+	return volumeBackupVolumeMap, nil
+}
+
+func (btc *BackupTargetController) syncBackupVolume(backupTarget *longhorn.BackupTarget, backupStoreBackupVolumeNames []string, clusterVolumeBVMap map[string]*longhorn.BackupVolume, syncTime metav1.Time, log logrus.FieldLogger) error {
+	backupStoreBackupVolumes := sets.New[string](backupStoreBackupVolumeNames...)
 	clusterBackupVolumesSet := sets.New[string]()
-	for _, bv := range clusterBackupVolumes {
+	for _, bv := range clusterVolumeBVMap {
```
derekbit, line 605: Any reason of the prefix `cluster`?

controller/backup_target_controller.go
```diff
@@ -633,14 +683,51 @@ func (btc *BackupTargetController) cleanupBackupVolumeNotExistOnBackupTarget(clu
...
+func (btc *BackupTargetController) deleteOnlyBackupVolumeCR(backupVolumeName string, log logrus.FieldLogger) (err error) {
+
+	if err = datastore.AddBackupVolumeDeleteCustomResourceOnlyLabel(btc.ds, backupVolumeName); err != nil {
+		return errors.Wrapf(err, "failed to add label delete-custom-resource-only to BackupVolume %s", backupVolumeName)
+	}
+	if err = btc.ds.DeleteBackupVolume(backupVolumeName); err != nil {
+		return errors.Wrapf(err, "failed to delete BackupVolume %s", backupVolumeName)
+	}
+	return nil
+}
```
derekbit:
```suggestion
func (btc *BackupTargetController) deleteOnlyBackupVolumeCR(backupVolumeName string, log logrus.FieldLogger) error {

	if err := datastore.AddBackupVolumeDeleteCustomResourceOnlyLabel(btc.ds, backupVolumeName); err != nil {
		return errors.Wrapf(err, "failed to add label delete-custom-resource-only to BackupVolume %s", backupVolumeName)
	}
	if err := btc.ds.DeleteBackupVolume(backupVolumeName); err != nil {
		return errors.Wrapf(err, "failed to delete BackupVolume %s", backupVolumeName)
	}
	return nil
}
```

controller/backup_target_controller.go
mantissahz: Modified. Thanks

controller/backup_target_controller.go
mantissahz: Modified. Thanks.

controller/backup_target_controller.go
mantissahz: Modified. Thanks.

types/types.go
mantissahz: Added. Thanks

controller/backup_target_controller.go
```diff
@@ -546,20 +553,59 @@ func (btc *BackupTargetController) getInfoFromBackupStore(backupTarget *longhorn
 	return info, nil
 }
 
-func (btc *BackupTargetController) syncBackupVolume(backupTarget *longhorn.BackupTarget, backupStoreBackupVolumeNames []string, syncTime metav1.Time, log logrus.FieldLogger) error {
-	backupStoreBackupVolumes := sets.New[string](backupStoreBackupVolumeNames...)
+func (btc *BackupTargetController) cleanupDuplicateBackupVolumes(backupTarget *longhorn.BackupTarget) (map[string]*longhorn.BackupVolume, error) {
```
derekbit: The function is quite confusing. Refactor the function.

controller/backup_target_controller.go
```diff
@@ -433,6 +433,13 @@ func (btc *BackupTargetController) reconcile(name string) (err error) {
 		}
 	}()
 
+	// clean up invalid backup volumes that are created during split-brain
+	// https://github.com/longhorn/longhorn/issues/11154
+	volumeBackupVolumeMap, err := btc.cleanupDuplicateBackupVolumes(backupTarget)
```
derekbit: Separate the `cleanupDuplicateBackupVolumes` into two functions, one is collecting `volumeBackupVolumeMap` and the other one is cleaning up the duplicates.

derekbit: Rename `volumeBackupVolumeMap` to `volumeBVMap`

controller/backup_target_controller.go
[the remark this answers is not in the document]
derekbit: Rename `volumeBackupVolumeMap` to `volumeBVMap` to simplicity

controller/backup_target_controller.go
```diff
@@ -633,14 +683,51 @@ func (btc *BackupTargetController) cleanupBackupVolumeNotExistOnBackupTarget(clu
 
 		backupVolumeName := bv.Name
 		log.WithField("backupVolume", backupVolumeName).Info("Deleting BackupVolume not exist in backupstore")
-		if err = datastore.AddBackupVolumeDeleteCustomResourceOnlyLabel(btc.ds, backupVolumeName); err != nil {
-			return errors.Wrapf(err, "failed to add label delete-custom-resource-only to Backupvolume %s", backupVolumeName)
+		if err := btc.deleteOnlyBackupVolumeCR(backupVolumeName, log); err != nil {
```
derekbit:
```suggestion
		if err := btc.deleteBackupVolumeCROnly(backupVolumeName, log); err != nil {
```

controller/backup_target_controller.go
```diff
@@ -633,14 +683,51 @@ func (btc *BackupTargetController) cleanupBackupVolumeNotExistOnBackupTarget(clu
 
 		backupVolumeName := bv.Name
 		log.WithField("backupVolume", backupVolumeName).Info("Deleting BackupVolume not exist in backupstore")
```
derekbit, line 688: The info message is misleading. Please improve it. `deleteOnlyBackupVolumeCR` also deletes BV found in backupstore.

derekbit: BTW, rename 
`clusterVolumeBVMap` to `volumeBVMap`
`clusterBackupVolumesSet` to `backupVolumesSet`


cluster seems meaningless here.

controller/backup_target_controller.go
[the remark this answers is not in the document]
mantissahz: Modified. Thanks

controller/backup_target_controller.go
mantissahz: Separated to `getClusterBackupVolumes` and `cleanupDuplicateBackupVolumeForBackupTarget`

controller/backup_target_controller.go
mantissahz: Modified. Thanks.

controller/backup_target_controller.go
mantissahz: Modified. Thanks.

controller/backup_target_controller.go
mantissahz: >The info message is misleading. Please improve it. deleteOnlyBackupVolumeCR also deletes BV found in backupstore.

This log is in `cleanupBackupVolumeNotExistOnBackupTarget` to delete BackupVolume CRs not in the backup store and not in `deleteOnlyBackupVolumeCR `.

As we discussed, keep the prefix `cluster` in the name with other names in the methods.

controller/backup_target_controller.go
```diff
@@ -546,20 +556,54 @@ func (btc *BackupTargetController) getInfoFromBackupStore(backupTarget *longhorn
...
-		return err
+		return nil, nil, err
 	}
 
-	clusterVolumeBVMap := make(map[string]*longhorn.BackupVolume, len(clusterBackupVolumes))
+	duplicateBackupVolumeSet := sets.New[string]()
+	volumeBVMap := make(map[string]*longhorn.BackupVolume, len(backupVolumeList))
+	for _, bv := range backupVolumeList {
+		if bv.Spec.BackupTargetName == "" {
+			log.WithField("backupVolume", bv.Name).Debugf("Backup target name is empty in BackupVolume")
```
derekbit:
```suggestion
			log.WithField("backupVolume", bv.Name).Debug("spec.backupTargetName is empty")
```

controller/backup_target_controller.go
```diff
@@ -546,20 +556,54 @@ func (btc *BackupTargetController) getInfoFromBackupStore(backupTarget *longhorn
...
+			duplicateBackupVolumeSet.Insert(bv.Name)
+			continue
+		}
+		if bv.Spec.VolumeName == "" {
+			log.WithField("backupVolume", bv.Name).Debugf("Volume name is empty in BackupVolume")
```
derekbit:
```suggestion
			log.WithField("backupVolume", bv.Name).Debug("spec.volumeName is empty")
```

controller/backup_target_controller.go
```diff
@@ -546,20 +556,54 @@ func (btc *BackupTargetController) getInfoFromBackupStore(backupTarget *longhorn
...
+			continue
+		}
+		if existingBV, exists := volumeBVMap[bv.Spec.VolumeName]; exists {
+			if existingBV.CreationTimestamp.Before(&bv.CreationTimestamp) {
+				log.WithField("backupVolume", bv.Name).Warnf("Found duplicated BackupVolume with volume name %s", bv.Spec.VolumeName)
+				duplicateBackupVolumeSet.Insert(bv.Name)
+				continue
+			}
+			log.WithField("backupVolume", existingBV.Name).Warnf("Found duplicated BackupVolume with volume name %s", existingBV.Spec.VolumeName)
+			duplicateBackupVolumeSet.Insert(existingBV.Name)
```
derekbit, line 594: No need `continue`?

controller/backup_target_controller.go
```diff
@@ -546,20 +556,54 @@ func (btc *BackupTargetController) getInfoFromBackupStore(backupTarget *longhorn
...
+		}
+		volumeBVMap[bv.Spec.VolumeName] = bv
+	}
+
+	return volumeBVMap, duplicateBackupVolumeSet, nil
```
derekbit, line 599: Does it mean each volume in volumeBVMap can only have one BV?

controller/backup_target_controller.go
```diff
@@ -633,14 +681,51 @@ func (btc *BackupTargetController) cleanupBackupVolumeNotExistOnBackupTarget(clu
...
+	}
+
+	if len(multiError) > 0 {
+		return fmt.Errorf("failed to delete backup volumes from cluster: %v", multiError.Join())
+	}
+	return nil
+}
+
+func (btc *BackupTargetController) deleteBackupVolumeCROnly(backupVolumeName string, log logrus.FieldLogger) error {
+
```
derekbit: Redundant line

controller/backup_target_controller.go
```diff
@@ -580,10 +624,13 @@ func (btc *BackupTargetController) syncBackupVolume(backupTarget *longhorn.Backu
...
+	for volumeName, backupVolume := range clusterVolumeBVMap {
+		if !backupStoreBackupVolumes.Has(volumeName) {
+			continue
+		}
 		backupVolume.Spec.SyncRequestedAt = syncTime
-		if _, err = btc.ds.UpdateBackupVolume(backupVolume); err != nil && !apierrors.IsConflict(errors.Cause(err)) {
-			log.WithError(err).Errorf("Failed to update backup volume %s spec", backupVolumeName)
+		if _, err := btc.ds.UpdateBackupVolume(backupVolume); err != nil && !apierrors.IsConflict(errors.Cause(err)) {
+			log.WithError(err).Errorf("Failed to update backup volume %s spec", backupVolume.Name)
 		}
```
derekbit: Don't we need to return error?

controller/backup_target_controller.go
```diff
@@ -580,10 +624,13 @@ func (btc *BackupTargetController) syncBackupVolume(backupTarget *longhorn.Backu
...
-	for backupVolumeName, backupVolume := range clusterBackupVolumes {
+	for volumeName, backupVolume := range clusterVolumeBVMap {
+		if !backupStoreBackupVolumes.Has(volumeName) {
+			continue
+		}
 		backupVolume.Spec.SyncRequestedAt = syncTime
-		if _, err = btc.ds.UpdateBackupVolume(backupVolume); err != nil && !apierrors.IsConflict(errors.Cause(err)) {
-			log.WithError(err).Errorf("Failed to update backup volume %s spec", backupVolumeName)
+		if _, err := btc.ds.UpdateBackupVolume(backupVolume); err != nil && !apierrors.IsConflict(errors.Cause(err)) {
+			log.WithError(err).Errorf("Failed to update backup volume %s spec", backupVolume.Name)
```
derekbit:
```suggestion
			log.WithError(err).Errorf("Failed to update backup volume %s", backupVolume.Name)
```

controller/backup_target_controller.go
mantissahz: Modified. Thanks

controller/backup_target_controller.go
mantissahz: Modified. Thanks

controller/backup_target_controller.go
mantissahz: Modified. Thanks

controller/backup_target_controller.go
mantissahz: Modified. Thanks

controller/backup_target_controller.go
mantissahz: Removed. Thanks

controller/backup_target_controller.go
mantissahz: Yes, there is only one volume to one backup volume for a backup target.

controller/backup_target_controller.go
mantissahz: No, it does not need it.
We insert the old existing BV into duplicatedBVSets and wait for the deletion.
Therefore, we need to replace the BackupVolume `existingBV` with this BackupVolume `bv` at the next line.

derekbit (approved):
LGTM

controller/backup_target_controller.go
```diff
@@ -546,20 +556,54 @@ func (btc *BackupTargetController) getInfoFromBackupStore(backupTarget *longhorn
 	return info, nil
 }
 
-func (btc *BackupTargetController) syncBackupVolume(backupTarget *longhorn.BackupTarget, backupStoreBackupVolumeNames []string, syncTime metav1.Time, log logrus.FieldLogger) error {
-	backupStoreBackupVolumes := sets.New[string](backupStoreBackupVolumeNames...)
+func (btc *BackupTargetController) getClusterBackupVolumes(backupTarget *longhorn.BackupTarget) (map[string]*longhorn.BackupVolume, sets.Set[string], error) {
```
shuo-wu: NIT: naming the outputs for this function.

shuo-wu (approved):
In general LGTM

PhanLe1010 (dismissed):
LGTM

controller/backup_target_controller.go
mantissahz: Modified the method name `getClusterBackupVolumes` to `getClusterBVsDuplicatedBVs`
Thanks

derekbit (approved):

PR #3899, opened 2025-07-07
fix(replica): enable offline replica rebuilding by default

mantissahz:

Issue # longhorn/longhorn#11204

Enable offline replica rebuilding by default.

### DIFF
--- a/types/setting.go
+++ b/types/setting.go
@@ -1590,7 +1590,7 @@ var (
 		Type:     SettingTypeBool,
 		Required: true,
 		ReadOnly: false,
-		Default:  "false",
+		Default:  "true",
 	}
 )
 


upgrade/v19xto1100/upgrade.go
```diff
@@ -0,0 +1,55 @@
...
+		return errors.Wrapf(err, "failed to list all existing Longhorn Settings during the Longhorn Settings upgrade")
+	}
+
+	for _, setting := range settingMap {
+		if setting.Name == string(types.SettingNameOfflineReplicaRebuilding) {
+			// The default value of OfflineReplicaRebuilding is changed from false to true in v1.10.0.
+			setting.Value = "true"
+			break
+		}
+	}
```
derekbit: No need to use for loop

mantissahz: Modified. Thanks.

innobead (dismissed):

derekbit:
> Do we need to blindly and silently update the existing setting value to `true` during the upgrade?

@shuo-wu Are you fine with the blind update?

mantissahz:
>Do we need to blindly and silently update the existing setting value to true during the upgrade?

Removed the upgrade path as we discussed.

shuo-wu (approved):

PR #3905, opened 2025-07-10
feat(disk): add disk performance metrics monitoring support

hookak:

Issue https://github.com/longhorn/longhorn/issues/11223

* The existing disk status is reflected in DiskStatus, and DiskStatus was already being exported to Prometheus.
Therefore, in order to include the new disk metrics, it was necessary to modify the DiskStatus field in the CRD.
* Since the resync interval of the controller is 30 seconds, it is difficult to detect very short-term changes in disk performance.


hookak:
The commit was pushed directly to the `longhorn-manager` repository instead of my own fork.
I'll push it to the `hookak` repo and then close this PR.

PR #3906, opened 2025-07-10
feat(disk): add disk performance metrics monitoring support

hookak:

Issue https://github.com/longhorn/longhorn/issues/11223

* The existing disk status is reflected in DiskStatus, and DiskStatus was already being exported to Prometheus.
Therefore, in order to include the new disk metrics, it was necessary to modify the DiskStatus field in the CRD.
* Since the resync interval of the controller is 30 seconds, it is difficult to detect very short-term changes in disk performance.


controller/node_controller.go
```diff
@@ -767,6 +767,19 @@ func (nc *NodeController) findNotReadyAndReadyDiskMaps(node *longhorn.Node, coll
...
+			if diskInfo.DiskMetrics != nil {
+				node.Status.DiskStatus[diskName].DiskMetrics = &longhorn.DiskMetrics{
+					ReadThroughput:  diskInfo.DiskMetrics.ReadThroughput,
+					WriteThroughput: diskInfo.DiskMetrics.WriteThroughput,
+					ReadLatency:     diskInfo.DiskMetrics.ReadLatency,
+					WriteLatency:    diskInfo.DiskMetrics.WriteLatency,
+					ReadIOPS:        diskInfo.DiskMetrics.ReadIOPS,
+					WriteIOPS:       diskInfo.DiskMetrics.WriteIOPS,
+				}
+			}
```
derekbit, line 781: Ah, one more question. Should we have the metrics in `node.diskStatus`? Or just using monitoring tool to fetch the data is enough?

hookak: Actually, I also wanted to avoid adding metrics to `node.diskStatus`, since it would require changing the CRD. But other disk info metrics already follow the flow of `disk monitor -> disk status -> Prometheus`, so I implemented it that way to say consistent.
Also, I couldn't find a proper way to call `diskService` from the `metric_collector` instead of the `disk monitor`.

hookak: If there's a good way to fetch the metrics in monitoring tools without putting them into `diskStatus`, I'd really appreciate it if you could share it.

derekbit: Got your point. Let me think if there is any solution to this.

derekbit: Hello @hookak 

I've implemented a PoC without node resource's `diskMetrics`.
https://github.com/derekbit/longhorn-manager/commit/f23fb593584b6d2a1afc3b8bf0f40d56c6e221aa
Could you help check the implementation? If it works for you, feel free to submit a PR to improve and remove node resource's diskMetrics. Thank you ;)

hookak: Thanks, I'll take a look and submit a PR! @derekbit

PR #3907, opened 2025-07-10
feat: persist instance manager and spdk_tgt log messages

derekbit:

Issue Longhorn/longhorn#5948

### DIFF
--- a/controller/instance_manager_controller.go
+++ b/controller/instance_manager_controller.go
@@ -5,6 +5,8 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"os"
+	"path/filepath"
 	"reflect"
 	"strconv"
 	"strings"
@@ -22,6 +24,7 @@ import (
 	"k8s.io/client-go/tools/record"
 	"k8s.io/client-go/util/flowcontrol"
 	"k8s.io/kubernetes/pkg/controller"
+	"k8s.io/utils/ptr"
 
 	corev1 "k8s.io/api/core/v1"
 	policyv1 "k8s.io/api/policy/v1"
@@ -760,6 +763,8 @@ func (imc *InstanceManagerController) areDangerZoneSettingsSyncedToIMPod(im *lon
 			isSettingSynced, err = imc.isSettingDataEngineSynced(settingName, im)
 		case types.SettingNameInstanceManagerPodLivenessProbeTimeout:
 			isSettingSynced, err = imc.isSettingInstanceManagerPodLivenessProbeTimeoutSynced(setting, pod)
+		case types.SettingNameLogPath:
+			isSettingSynced, err = imc.isSettingLogPathSynced(setting, pod)
 		}
 		if err != nil {
 			return false, false, false, err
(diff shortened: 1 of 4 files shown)


controller/instance_manager_controller.go
```diff
@@ -1461,15 +1464,37 @@ func (imc *InstanceManagerController) createGenericManagerPodSpec(im *longhorn.I
...
+
+		processProbe := "[ $(ps aux | grep 'spdk_tgt' | grep -v 'grep' | grep -v 'tee' | wc -l) != 0 ]"
+		livenessProbes = append(livenessProbes, processProbe)
+	}
+	return fmt.Sprintf("test $(%s; echo $?) -eq 0", strings.Join(livenessProbes, " && "))
+}
+
 func (imc *InstanceManagerController) createInstanceManagerPodSpec(im *longhorn.InstanceManager, tolerations []corev1.Toleration, registrySecret string, nodeSelector map[string]string, dataEngine longhorn.DataEngineType) (*corev1.Pod, error) {
-	podSpec, err := imc.createGenericManagerPodSpec(im, tolerations, registrySecret, nodeSelector)
+	logDirectory, err := imc.ds.GetSettingValueExisted(types.SettingNameLogDirectory)
```
shuo-wu: Do we need to:
1. Check if the path format is valid
2. Have a fallback default value if the setting is invalid

shuo-wu (commented):
In general LGTM

controller/instance_manager_controller.go
derekbit: Addressed. Thank you.

controller/instance_manager_controller.go
```diff
@@ -819,6 +824,33 @@ func (imc *InstanceManagerController) isSettingPriorityClassSynced(setting *long
...
+	if logPath == "" {
+		logPath = types.DefaultLogDirectoryOnHost
+	}
+
+	normalizedLogPath := filepath.Clean(logPath)
+	for _, v := range pod.Spec.Volumes {
+		if v.Name == "log" {
+			if filepath.Clean(v.HostPath.Path) == normalizedLogPath {
+				return true, nil
+			} else {
```
mantissahz: As CodeFactor reviewed, we don't need ` else {}` here and move out the log and `return`.

mantissahz (approved):
In general, LGTM

controller/instance_manager_controller.go
derekbit: Updated. Thank you!

mantissahz (approved):

controller/instance_manager_controller.go
```diff
@@ -1503,215 +1536,264 @@
...
+					Optional:   ptr.To(true),
+				},
+			},
+		},
+		{
+			Name: "log",
+			VolumeSource: corev1.VolumeSource{
+				HostPath: &corev1.HostPathVolumeSource{
+					Path: logPath,
+					Type: &[]corev1.HostPathType{corev1.HostPathDirectoryOrCreate}[0],
```
innobead, line 1772: Simplify as below instead, as the type of const is type alias already? Or it doesn't work?

```go
&corev1.HostPathDirectoryOrCreate
```

types/setting.go
```diff
@@ -1702,6 +1705,17 @@ var (
...
+
+	SettingDefinitionLogPath = SettingDefinition{
+		DisplayName:        "Log Path",
+		Description:        "Specifies the directory on the host where Longhorn stores log files for the instance manager pod. Currently, it is only used for instance manager pods in the v2 data engine.",
+		Category:           SettingCategoryDangerZone,
+		Type:               SettingTypeString,
+		Required:           true,
+		ReadOnly:           false,
+		DataEngineSpecific: false,
+		Default:            "/var/lib/longhorn/logs/",
```
innobead, line 1717: Can just use const `DefaultLogDirectoryOnHost` from types.go.

controller/instance_manager_controller.go
```diff
@@ -1503,15 +1536,64 @@ func (imc *InstanceManagerController) createGenericManagerPodSpec(im *longhorn.I
...
+	}
+
+	validatePathAndFallback := func(p string) string {
+		parent := filepath.Dir(p)
+		if parent == "." || parent == string(filepath.Separator) {
+			imc.logger.Warnf("Log path %q is not a valid directory, using default log directory", p)
+			return types.DefaultLogDirectoryOnHost
+		}
+
+		st, err := os.Stat(parent)
```
innobead, line 1578: Longhorn Manager pods only mount the following host paths. The validation of the provided non-default log path may be misleading, as it is not mapped to the path on the host.

May just make the path fixed, but users can mount that accordingly on host to simplify the operation?

https://github.com/longhorn/longhorn/blob/600801b5c35869ec0d05cf5190daf01c110b1eca/chart/templates/daemonset-sa.yaml#L63-L79

innobead:
@derekbit Please check the comments after the merge. See if anything needs to be fixed.

controller/instance_manager_controller.go
derekbit: This doesn't work because `corev1.HostPathTypeDirectoryOrCreate` is a constant of type `HostPathType`, but `Type` needs `*HostPathType`. In golang, it doesn't get the address of a constant.

derekbit:
@innobead 
The comments are addressed in 
 - https://github.com/longhorn/go-common-libs/pull/129
 - https://github.com/longhorn/longhorn-manager/pull/4001
Thank you.

PR #3910, opened 2025-07-11
feat(crd): remove preserveUnknownFields patch

derekbit:

Issue Longhorn/longhorn#11263

- Kubernetes enforces the rule: If `preserveUnknownFields: true`, then `conversion.Strategy` must be `None`, because
    - When `preserveUnknownFields: true` is set, Kubernetes skips strict schema validation and structural typing.
    - Webhook conversion requires a well-defined schema and structural CRs, so the two settings are incompatible. Starting from Longhorn v1.10, the v1beta1 API has been removed, and the conversion webhook is no longer used.

- In addition, preserveUnknownFields is deprecated in `apiextensions.k8s.io/v1` since Kubernetes v1.22, and the default value is false.


derekbit:

PR #3931, opened 2025-07-18
fix(backup): not check backup target available if

mantissahz:

Issue # longhorn/longhorn#11337, https://github.com/longhorn/longhorn/issues/10085

There is no need to check the backup target availability if Backups, SystemBackups, and BackupBackingImages are created by Longhorn (the backup target controller).
The backup target is unavailable before we set up a valid backup target, and if it has existing backups, system backups, or backup backing images, setting the backup target to be available will be blocked

### DIFF
--- a/controller/backup_target_controller.go
+++ b/controller/backup_target_controller.go
@@ -653,8 +653,9 @@ func (btc *BackupTargetController) pullBackupVolumeFromBackupTarget(backupTarget
 			ObjectMeta: metav1.ObjectMeta{
 				Name: backupVolumeName,
 				Labels: map[string]string{
-					types.LonghornLabelBackupTarget: backupTarget.Name,
-					types.LonghornLabelBackupVolume: remoteVolumeName,
+					types.LonghornLabelBackupTarget:                                   backupTarget.Name,
+					types.LonghornLabelBackupVolume:                                   remoteVolumeName,
+					types.GetLonghornLabelKey(types.CreateCustomResourceFromLonghorn): "true",
 				},
 				OwnerReferences: datastore.GetOwnerReferencesForBackupTarget(backupTarget),
 			},
@@ -759,8 +760,9 @@ func (btc *BackupTargetController) syncBackupBackingImage(backupTarget *longhorn
 			ObjectMeta: metav1.ObjectMeta{
 				Name: backupBackingImageName,
 				Labels: map[string]string{
-					types.LonghornLabelBackupTarget: backupTarget.Name,
-					types.LonghornLabelBackingImage: canonicalBackingImageName,
+					types.LonghornLabelBackupTarget:                                   backupTarget.Name,
+					types.LonghornLabelBackingImage:                                   canonicalBackingImageName,
+					types.GetLonghornLabelKey(types.CreateCustomResourceFromLonghorn): "true",
 				},
 				OwnerReferences: datastore.GetOwnerReferencesForBackupTarget(backupTarget),
 			},
@@ -831,7 +833,8 @@ func (btc *BackupTargetController) syncSystemBackup(backupTarget *longhorn.Backu
 				Labels: map[string]string{
 					// Label with the version to be used by the system-backup controller
 					// to get the config from the backup target.
-					types.GetVersionLabelKey(): longhornVersion,
+					types.GetVersionLabelKey():                                        longhornVersion,
+					types.GetLonghornLabelKey(types.CreateCustomResourceFromLonghorn): "true",
 				},
 				OwnerReferences: datastore.GetOwnerReferencesForBackupTarget(backupTarget),
 			},
--- a/controller/backup_volume_controller.go
+++ b/controller/backup_volume_controller.go
@@ -338,7 +338,8 @@ func (bvc *BackupVolumeController) reconcile(backupVolumeName string) (err error
 			ObjectMeta: metav1.ObjectMeta{
 				Name: backupName,
 				Labels: map[string]string{
-					types.LonghornLabelBackupTarget: backupTarget.Name,
+					types.LonghornLabelBackupTarget:                                   backupTarget.Name,
+					types.GetLonghornLabelKey(types.CreateCustomResourceFromLonghorn): "true",
 				},
 				OwnerReferences: datastore.GetOwnerReferencesForBackupVolume(backupVolume),
 			},
--- a/datastore/longhorn.go
+++ b/datastore/longhorn.go
@@ -3346,6 +3346,21 @@ func labelBackupVolume(backupVolumeName string, obj k8sruntime.Object) error {
 	return nil
 }
 
+// IsLabelLonghornCreateCustomResourceFromLonghornExisting check if the object label `longhorn.io/create-custom-resource-from-longhorn` exists
+func IsLabelLonghornCreateCustomResourceFromLonghornExisting(obj k8sruntime.Object) (bool, error) {
+	metadata, err := meta.Accessor(obj)
+	if err != nil {
+		return false, err
+	}
+
+	labels := metadata.GetLabels()
+	if labels == nil {
+		return false, nil
+	}
+	_, ok := labels[types.GetLonghornLabelKey(types.CreateCustomResourceFromLonghorn)]
+	return ok, nil
+}
+
 // labelLonghornDeleteCustomResourceOnly labels the object with the label `longhorn.io/delete-custom-resource-only: true`
 func labelLonghornDeleteCustomResourceOnly(obj k8sruntime.Object) error {
 	metadata, err := meta.Accessor(obj)
--- a/types/types.go
+++ b/types/types.go
@@ -138,7 +138,8 @@ const (
 	ConfigMapResourceVersionKey = "configmap-resource-version"
 	UpdateSettingFromLonghorn   = "update-setting-from-longhorn"
 
-	DeleteCustomResourceOnly = "delete-custom-resource-only"
+	CreateCustomResourceFromLonghorn = "create-custom-resource-from-longhorn"
+	DeleteCustomResourceOnly         = "delete-custom-resource-only"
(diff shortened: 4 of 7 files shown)


mantissahz:
@Mergifyio backport v1.8.x v1.9.x

PhanLe1010:
> We don't need to check the backup target availability if Backups, SystemBackups, and BackupBackingImages are created by Longhorn (the backup target/backup volume controller).
The backup target is unavailable before we set up a valid backup target, and if it has existing system backups or backup backing images, setting the backup target to be available will be blocked.

Hi @mantissahz Is it the reproduce steps?

> 1. The backup target is not available when the backup target URL is empty.
> 2. We set up the correct backup target
> 3. Reconcile the backup target to get the backup information from the backup store.
> 4. At this moment, the backup target is still unavailable, and the backup target controller tries to create Backups, BackupVolumes, BackupBackingImages and SystemBackups.
> 5. The validator will deny the CREATE operations.
> 6. The backup target cannot be available.

I am not clear why The backup target cannot be available at step 6? Why backup target setting is blocked?

mantissahz:
>Hi @mantissahz Is it the reproduce steps?
>1. The backup target is not available when the backup target URL is empty.
>2. We set up the correct backup target
>3. Reconcile the backup target to get the backup information from the backup store.
>4. At this moment, the backup target is still unavailable, and the backup target controller tries to create Backups, BackupVolumes, BackupBackingImages and SystemBackups.
>6. The validator will deny the CREATE operations.
>7. The backup target cannot be available.

>I am not clear why The backup target cannot be available at step 6? Why backup target setting is blocked?

Hi @PhanLe1010,

https://github.com/longhorn/longhorn-manager/blob/c933a9ea1442475c24a56b551600eead81fbbb7b/controller/backup_target_controller.go#L480-L489

Here, we will try to create BBIs and SystemBackups pulled from the backup store.
If the backup target is not available, the validators will deny the CREATE operation and return the error.
If there is an error, the reconciliation will not update the backup target status.
https://github.com/longhorn/longhorn-manager/blob/c933a9ea1442475c24a56b551600eead81fbbb7b/controller/backup_target_controller.go#L417-L434

PhanLe1010:
I see. Thanks @mantissahz for the explaination. I see that:
* BackupVolume is always created by Longhorn. So maybe we can always skip checking for this
* BackupBackingImageSpec already has `UserCreated:      false,` which indicate that it does not need to be created.

So I am thinking if we should follow the same convention as BackupBackingImageSpec, Longhorn Snapshot CR. Which means we should add `UserCreated:      false,` to BackupBackingImageSpec and SystemBackups CR to indicate whether they are user created or not. Instead of using the new label as in the PR.

WDYT about the proposal?

shuo-wu (approved):
LGTM but the label naming is a little bit confusing...

PhanLe1010:
> BTW, I don't think we can use field UserCreated since it's mainly for distinguishing whether the backup creation request is generated from users or Longhorn. The case here is more like if a backup is originated from this cluster request (and requiring actual data transfer) or from the remote backupstore pulling result.

* `UserCreated:      true` mean backup is created by user -> require data transfer 
* `UserCreated:      false` mean backup is created by some controller by pulling from backup target

Seems working ok to me. 

The label approach is ok too if all prefer it

mantissahz:
The `BackupTarget.Status.Available` field means the backup store is valid and reachable. Then the backup target controller can start to get the backup information.
After discussing with @derekbit, we should update the backup target status to `available` before the backup target controller begins pulling backup-related information to create or delete CRs.
Therefore, there's no need to add additional labels or fields.
WDYT? @PhanLe1010, @shuo-wu 

I ~will test~ tested this way to check if it works: Works
Backup-related regression test:
https://10.115.5.5/job/private/job/longhorn-tests-regression/82/

derekbit:
> The `BackupTarget.Status.Available` field means the backup store is valid and reachable. Then the backup target controller can start to get the backup information. After discussing with @derekbit, we should update the backup target status to `available` before the backup target controller begins pulling backup-related information to create or delete CRs. Therefore, there's no need to add additional labels or fields. WDYT? @PhanLe1010, @shuo-wu
> 
> I ~will test~ tested this way to check if it works: Works Backup-related regression test: [10.115.5.5/job/private/job/longhorn-tests-regression/82](https://10.115.5.5/job/private/job/longhorn-tests-regression/82/)

One case failed. Do you know why?

mantissahz:
>One case failed. Do you know why?
Because the test case `test_backuptarget_invalid` only checks the failed messages below:
```python
BACKUP_TARGET_MESSAGES_INVALID = ["failed to init backup target clients",
                                  "failed to list backup volumes in",
                                  "error listing backup volume names"]
```
And I moved the SystemBackups listing before the BackupVolumes listing, so we need to add a failed sub-message `failed to list system backups in`.
```python
BACKUP_TARGET_MESSAGES_INVALID = ["failed to init backup target clients",
                                  "failed to list backup volumes in",
                                  "failed to list system backups in",
                                  "error listing backup volume names"]
```

controller/backup_target_controller.go
```diff
@@ -540,6 +543,14 @@ func (btc *BackupTargetController) getInfoFromBackupStore(backupTarget *longhorn
...
 
 	// Get required information using backup target client.
+	// Get SystemBackups first to update the backup target to `available` while minimizing requests to S3.
+	info.backupStoreSystemBackups, err = backupTargetClient.ListSystemBackup()
+	if err != nil {
+		return backupStoreInfo{}, errors.Wrapf(err, "failed to list system backups in %v", backupTargetClient.URL)
+	}
+	if !backupTarget.Status.Available {
+		return info, nil
+	}
```
shuo-wu, line 556: Do we have other kinds of methods to verify if the backup target is available/reachable or not?

controller/backup_target_controller.go
```diff
@@ -466,12 +466,15 @@ func (btc *BackupTargetController) reconcile(name string) (err error) {
...
-	backupTarget.Status.Available = true
-	backupTarget.Status.Conditions = types.SetCondition(backupTarget.Status.Conditions,
-		longhorn.BackupTargetConditionTypeUnavailable, longhorn.ConditionStatusFalse,
-		"", "")
+	if !backupTarget.Status.Available {
+		backupTarget.Status.Available = true
+		backupTarget.Status.Conditions = types.SetCondition(backupTarget.Status.Conditions,
+			longhorn.BackupTargetConditionTypeUnavailable, longhorn.ConditionStatusFalse,
+			"", "")
+		return nil
```
shuo-wu, line 478: Add a comment explaining why we need to return here

controller/backup_target_controller.go
mantissahz: No. IIUC, the only way to check if the backup target is available or reachable from the' longhorn-manager' is to execute the engine binary command.

controller/backup_target_controller.go
mantissahz: Added. Thanks.

controller/backup_target_controller.go
shuo-wu: OK. NTW, the comment is a little bit confusing as well. How about:
```
If the controller can communicate with the remote backup target while "backupTarget.Status.Available" is "false", Longhorn should update the field to "true" first rather than continuing to fetch info from the target.
```

mantissahz: Modified the comment. Thanks.

derekbit:
@mantissahz Can you run full regression on Jenkins before merging the PR? Thanks.

mantissahz:
>@mantissahz Can you run full regression on Jenkins before merging the PR? Thanks.

Full regression is running on our HAL Jenkins servers [longhorn-tests-regression-104](https://10.115.5.5/job/private/job/longhorn-tests-regression/104/)

derekbit:
@coderabbitai review

derekbit:
> > @mantissahz Can you run full regression on Jenkins before merging the PR? Thanks.
> 
> Full regression is running on our HAL Jenkins servers [longhorn-tests-regression-104](https://10.115.5.5/job/private/job/longhorn-tests-regression/104/)

@mantissahz 
Many test cases failed. Could you check them?

mantissahz:
> > > @mantissahz Can you run full regression on Jenkins before merging the PR? Thanks.
> >
> > Full regression is running on our HAL Jenkins servers [longhorn-tests-regression-104](https://10.115.5.5/job/private/job/longhorn-tests-regression/104/)
>
>@mantissahz
Many test cases failed. Could you check them?

Checking now.

Update:
Test cases failed because
```txt
msg="Failed to run list system backup command" func=cmd.SystemBackupCmd.SystemBackupListCmd.func4 file="system_backup.go:72" error="failed to list objects with param: {\\n  Bucket: \\"backupbucket\\",\\n  Delimiter: \\"/\\",\\n  Prefix: \\"backupstore/\\"\\n} error: AWS Error:  RequestError send request failed Get \\"https://minio-service.default:9000/backupbucket?delimiter=%2F&prefix=backupstore%2F\\": dial tcp 10.43.42.38:9000: connect: connection refused\\n"\n: exit status 1'
```
Don't know why it can not connect to the local Minio S3 server. cc @chriscchien, @longhorn/qa 
Replay one:
https://10.115.5.5/job/private/job/longhorn-tests-regression/108/
Start to run the full regression on AWS Jenkins:
https://ci.longhorn.io/job/private/job/longhorn-tests-regression/9139
Test cases have been PASSED three times in my local environment.

mantissahz:
> Replay one on HAL: https://10.115.5.5/job/private/job/longhorn-tests-regression/108/

Looks okay.

> AWS Jenkins: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/9139

Looks okay, failed cases are not replated to the backup target available status.

PhanLe1010 (approved):
LGTM

PR #3936, opened 2025-07-21
feat: consolidate settings for both v1 and v2 data engines

derekbit:

Issue Longhorn/longhorn#10926

This PR refactors Longhorn's settings system to support consolidated data engine specific settings, replacing the previous approach of maintaining separate V1 and V2 data engine settings with a unified system that can specify different values per data engine type.

- Key Features
  - Consolidated Setting Architecture
    - Introduces DataEngineSpecific field in SettingDefinition to enable unified setting management
    - Settings can now support both traditional single values and JSON-formatted data engine specific values
    - Eliminates the need for separate V1/V2 setting definitions
  - Enhanced Setting Value Formats
    - Non-JSON format: Single value applied to all applicable data engines
    - JSON format: Data engine specific values
      - `{"v1": "value1", "v2": "value2"}`: The applicable engines are v1 and v2
      - `{"v1": "value1"}`: The applicable engines are v1
      - `{"v2": "value1"}`: The applicable engines are v2
    - Automatic format detection and validation
  - New DataStore Methods
    - `GetSettingValueExistedByDataEngine()` - Get setting values for specific data engines
    - `GetSettingAsFloatByDataEngine()` - Retrieve float64 values by data engine
    - `GetSettingAsIntByDataEngine()` - Retrieve int64 values by data engine
    - `GetSettingAsBoolByDataEngine()` - Retrieve Boolean values by data engine
  - Robust Validation System
    - Enhanced validation functions support both single and JSON-formatted values
    - Data engine specific value parsing and validation
    - Comprehensive error handling with detailed error messages
  - Webhook Security
    - New setting mutation webhook ensures value validation during Create/Update operations
    - Automatic value normalization using datastore validation logic
    - Prevents invalid configurations from being applied
- Migration & Compatibility
  - The system automatically migrates existing V2-specific settings to consolidated equivalents:
		- `v2-data-engine-hugepage-limit` -> `hugepage-limit`
		- `v2-data-engine-cpu-mask` -> `cpu-mask`
		- `v2-data-engine-log-level` -> `spdk-tgt-log-level`
	    - `v2-data-engine-log-flags` -> `spdk-tgt-log-flags`
  - Upgrade Path
    - Added v1.9.x to v1.10.0 upgrade support with automatic setting value conversion
    - Existing setting values are preserved during migration
    - Backward compatibility maintained for existing deployments


derekbit:
Running regression test: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/8998/

k8s/crds.yaml
```diff
@@ -3555,6 +3548,13 @@ spec:
                 - enabled
                 - disabled
                 type: string
+              replicaRebuildBandwidthLimit:
```
c3y1huang: Do we need to add this to the update path?

k8s/pkg/apis/longhorn/v1beta2/setting.go
```diff
@@ -15,12 +15,16 @@ import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
...
 	// The value of the setting.
 	Value string `json:"value"`
-
 	// The status of the setting.
 	Status SettingStatus `json:"status,omitempty"`
+	// Default values for the setting based on the data engine type.
+	// If the value for a specific data engine type is not set, the default value will be used.
+	DefaultsByEngine map[DataEngineType]string `json:"defaultsByEngine,omitempty"`
+	// ApplicableEngines defines which data engines the setting is applicable to.
+	// If empty, the setting is applicable to all data engines.
```
c3y1huang: What is the benefit of allowing `ApplicableEngines` to be empty? Wouldn't it be more readable to explicitly set all applicable engines to `true`, so users don't have to be familiar with the "empty means all" convention?

types/setting.go
```diff
@@ -135,13 +135,10 @@ const (
 	SettingNameDisableSnapshotPurge                                     = SettingName("disable-snapshot-purge")
 	SettingNameV1DataEngine                                             = SettingName("v1-data-engine")
 	SettingNameV2DataEngine                                             = SettingName("v2-data-engine")
-	SettingNameV2DataEngineHugepageLimit                                = SettingName("v2-data-engine-hugepage-limit")
-	SettingNameV2DataEngineGuaranteedInstanceManagerCPU                 = SettingName("v2-data-engine-guaranteed-instance-manager-cpu")
-	SettingNameV2DataEngineCPUMask                                      = SettingName("v2-data-engine-cpu-mask")
-	SettingNameV2DataEngineLogLevel                                     = SettingName("v2-data-engine-log-level")
-	SettingNameV2DataEngineLogFlags                                     = SettingName("v2-data-engine-log-flags")
-	SettingNameV2DataEngineFastReplicaRebuilding                        = SettingName("v2-data-engine-fast-replica-rebuilding")
-	SettingNameV2DataEngineSnapshotDataIntegrity                        = SettingName("v2-data-engine-snapshot-data-integrity")
```
c3y1huang, line 144: Should Longhorn handle these change in the upgrade path?

types/setting.go
```diff
@@ -392,11 +385,13 @@ var (
...
-		Required: true,
-		ReadOnly: false,
-		Default:  "false",
+		Category:          SettingCategoryBackup,
+		Type:              SettingTypeBool,
+		Required:          true,
+		ReadOnly:          false,
+		Default:           "false",
+		DefaultsByEngine:  map[longhorn.DataEngineType]string{},
+		ApplicableEngines: map[longhorn.DataEngineType]bool{},
```
c3y1huang: Can we leave out  `DefaultsByEngine` and `ApplicableEngines` entirely for settings that are not engine-specific, to avoid confusion?

datastore/longhorn.go
```diff
@@ -736,6 +736,31 @@ func (s *DataStore) GetSettingValueExisted(sName types.SettingName) (string, err
 	return setting.Value, nil
 }
 
+func (s *DataStore) GetSettingValueExistedByDataEngine(sName types.SettingName, dataEngine longhorn.DataEngineType) (string, error) {
```
c3y1huang: Suggest adding description to exported methods.

datastore/longhorn.go
```diff
@@ -736,6 +736,31 @@ func (s *DataStore) GetSettingValueExisted(sName types.SettingName) (string, err
...
+		dataEngine = longhorn.DataEngineTypeV1
+	}
+
+	setting, err := s.GetSettingWithAutoFillingRO(sName)
+	if err != nil {
+		return "", err
+	}
+
+	value := setting.Value
+	if setting.DefaultsByEngine != nil {
```
c3y1huang: I think we can simplify the logic by removing the fallback to `setting.Value`. Could the engine-specific setting rely only on `setting.DefaultsByEngine`?

controller/engine_controller.go
```diff
@@ -1492,10 +1492,10 @@ func checkSizeBeforeRestoration(log logrus.FieldLogger, engine *longhorn.Engine,
 	return true, nil
 }
 
-func (m *EngineMonitor) restoreBackup(engine *longhorn.Engine, rsMap map[string]*longhorn.RestoreStatus, cliAPIVersion int, engineClientProxy engineapi.EngineClientProxy) error {
-	backupVolume, err := m.ds.GetBackupVolumeRO(engine.Spec.BackupVolume)
+func (m *EngineMonitor) restoreBackup(e *longhorn.Engine, rsMap map[string]*longhorn.RestoreStatus, cliAPIVersion int, engineClientProxy engineapi.EngineClientProxy) error {
+	backupVolume, err := m.ds.GetBackupVolumeRO(e.Spec.BackupVolume)
 	if err != nil {
-		return errors.Wrapf(err, "failed to get backup volume %v for backup restoration of engine %v", engine.Spec.BackupVolume, engine.Name)
```
c3y1huang: I suggest we stick with `engine`, as it improves readability and is more self-explanatory for new contributors. Readability should always be preferred unless the variable is used in a tight loop.

webhook/resources/setting/validator.go
```diff
@@ -44,20 +48,42 @@ func (v *settingValidator) Create(request *admission.Request, newObj runtime.Obj
 }
 
 func (v *settingValidator) Update(request *admission.Request, oldObj runtime.Object, newObj runtime.Object) error {
-	setting, ok := newObj.(*longhorn.Setting)
+	oldSetting, ok := oldObj.(*longhorn.Setting)
 	if !ok {
-		return werror.NewInvalidError(fmt.Sprintf("%v is not a *longhorn.Setting", newObj), "")
+		return werror.NewInvalidError(fmt.Sprintf("oldOjb %v is not a *longhorn.Setting", oldObj), "")
```
c3y1huang:
```suggestion
		return werror.NewInvalidError(fmt.Sprintf("oldObj %v is not a *longhorn.Setting", oldObj), "")
```

webhook/resources/setting/validator.go
```diff
@@ -44,20 +48,42 @@ func (v *settingValidator) Create(request *admission.Request, newObj runtime.Obj
...
+	}
+	newSetting, ok := newObj.(*longhorn.Setting)
+	if !ok {
+		return werror.NewInvalidError(fmt.Sprintf("newObj %v is not a *longhorn.Setting", newObj), "")
+	}
+
+	if !cmp.Equal(oldSetting.ApplicableEngines, newSetting.ApplicableEngines) {
+		return werror.NewInvalidError("applicableEngines cannot be modified", "applicableEngines")
+	}
```
c3y1huang: Could this potentially break future upgrades where engine support changes?

webhook/resources/setting/validator.go
```diff
@@ -100,3 +139,13 @@ func (v *settingValidator) validateSetting(newObj runtime.Object) error {
...
+
+func getNumOfApplicableEngines(applicableEngines map[longhorn.DataEngineType]bool) int {
+	numApplicable := 0
+	for _, applicable := range applicableEngines {
+		if applicable {
+			numApplicable++
+		}
+	}
+	return numApplicable
+}
```
c3y1huang:
```suggestion
func getNumOfApplicableEngines(applicableEngines map[longhorn.DataEngineType]bool) int {
	count := 0
	for _, applicable := range applicableEngines {
		if applicable {
			count++
		}
	}
	return count
}
```

go.mod
```diff
@@ -139,7 +139,7 @@ require (
 	github.com/go-openapi/jsonreference v0.21.0 // indirect
 	github.com/go-openapi/swag v0.23.0 // indirect
 	github.com/gogo/protobuf v1.3.2 // indirect
-	github.com/google/go-cmp v0.7.0 // indirect
+	github.com/google/go-cmp v0.7.0
```
c3y1huang: Can we move this to the direct `require` block?

types/setting.go
```diff
@@ -153,6 +153,15 @@ const (
...
 	SettingNameBackupstorePollInterval      = SettingName("backupstore-poll-interval")
+
+	// Longhorn won't create Setting Resources for these parameters.
+	SettingNameV2DataEngineHugepageLimit                = SettingName("v2-data-engine-hugepage-limit")
+	SettingNameV2DataEngineGuaranteedInstanceManagerCPU = SettingName("v2-data-engine-guaranteed-instance-manager-cpu")
+	SettingNameV2DataEngineCPUMask                      = SettingName("v2-data-engine-cpu-mask")
+	SettingNameV2DataEngineLogLevel                     = SettingName("v2-data-engine-log-level")
+	SettingNameV2DataEngineLogFlags                     = SettingName("v2-data-engine-log-flags")
+	SettingNameV2DataEngineFastReplicaRebuilding        = SettingName("v2-data-engine-fast-replica-rebuilding")
+	SettingNameV2DataEngineSnapshotDataIntegrity        = SettingName("v2-data-engine-snapshot-data-integrity")
```
c3y1huang: Should we add a TODO to mark these for deprecation?

datastore/longhorn.go
```diff
@@ -615,6 +618,24 @@ func (s *DataStore) ValidateCPUMask(kubeNode *corev1.Node, value string) error {
 	return nil
 }
 
+func (s *DataStore) getMinNumCPUs() (int64, error) {
+	kubeNodes, err := s.ListKubeNodesRO()
+	if err != nil {
+		return -1, errors.Wrapf(err, "failed to list Kubernetes nodes for CPU mask validation")
```
c3y1huang:
```suggestion
		return -1, errors.Wrapf(err, "failed to list Kubernetes nodes")
```

c3y1huang:
@coderabbitai review

k8s/pkg/apis/longhorn/v1beta2/setting.go
derekbit: Good question.
`ApplicableEngines := map[...]{}` means the default value is applied to all data engines and users cannot configure defaults for different engines.

`ApplicableEngines := map[...]{v1: true, v2: true}` means user can configure defaults for different engines.

types/setting.go
derekbit: They are migrated to new settings during longhorn-manager daemon startup rather than in the upgrade path to avoid the confusion caused by the existence of old settings.

types/setting.go
derekbit: Addressed. Thank you.

datastore/longhorn.go
derekbit: ~~Directly removing the fallback to `setting.Value` will directly break our setting configmap, because the configmap is currently supporting `setting.Value`. I think we can keep the `setting.Value` and fallback. WDYT?~~

k8s/pkg/apis/longhorn/v1beta2/setting.go
derekbit: After discussing with @c3y1huang, we will use `ApplicableEngines := map[...]{ all: true} for indicating  the default value is applied to all data engines and users cannot configure defaults for different engines.

controller/engine_controller.go
derekbit: Addressed. Thank you.

go.mod
derekbit: Addressed. Thank you.

datastore/longhorn.go
derekbit: Addressed. Thank you.

types/setting.go
derekbit: Added. Thank you.

types/setting.go
```diff
@@ -1983,7 +2180,7 @@ func validateString(sName SettingName, definition SettingDefinition, value strin
 			return errors.Wrapf(err, "the value of %v is invalid", sName)
 		}
 
-	case SettingNameV2DataEngineLogFlags:
+	case SettingNameDataEngineLogFlags:
 		if err := ValidateV2DataEngineLogFlags(value); err != nil {
 			return errors.Wrapf(err, "failed to validate v2 data engine log flags %v", value)
 		}
```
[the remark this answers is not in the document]
derekbit: Addressed. Thanks.

datastore/longhorn.go
```diff
@@ -394,13 +435,24 @@ func (s *DataStore) ValidateSetting(name, value string) (err error) {
...
+		for _, lhnode := range lhNodes {
+			kubeNode, err := s.GetKubernetesNodeRO(lhnode.Name)
+			if err != nil && !apierrors.IsNotFound(err) {
+				return errors.Wrapf(err, "failed to get Kubernetes node %s for %v setting validation", lhnode.Name, types.SettingNameCPUMask)
+			}
+
+			if err := s.ValidateCPUMask(kubeNode, value); err != nil {
+				return err
+			}
 		}
```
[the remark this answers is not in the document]
derekbit: Good catch! Addressed. Thanks.

webhook/resources/setting/validator.go
derekbit: Good catch! We shouldn't validate it.

webhook/resources/setting/validator.go
derekbit: Addressed. Thank you.

webhook/resources/setting/validator.go
derekbit: Addressed. Thank you.

datastore/longhorn.go
```diff
@@ -3515,6 +3611,91 @@
...
+	if definition.Type == types.SettingTypeFloat {
+		result, err := strconv.ParseFloat(value, 64)
+		if err != nil {
+			return -1, err
+		}
+		return result, nil
+	}
+
+	return -1, fmt.Errorf("the %v setting value couldn't change to float, value is %v ", string(settingName), value)
+}
```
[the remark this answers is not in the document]
derekbit: Good catch! Thanks.

k8s/crds.yaml
derekbit: No, because the setting is newly added since v1.10.

k8s/crds.yaml
```diff
@@ -2612,6 +2612,20 @@ spec:
...
+          defaultsByEngine:
+            additionalProperties:
+              type: string
+            description: |-
+              Default values for the setting based on the data engine type.
+              If the value for a specific data engine type is not set, the default value will be used.
+            type: object
           kind:
             description: |-
               Kind is a string value representing the REST resource this object represents.
```
sushant-suse:
```suggestion
              A string that identifies the REST resource type represented by this object.
```

k8s/crds.yaml
```diff
@@ -2612,6 +2612,20 @@ spec:
               may reject unrecognized values.
               More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources
             type: string
+          applicableEngines:
+            additionalProperties:
+              type: boolean
+            description: |-
+              ApplicableEngines defines which data engines the setting is applicable to.
+              If empty, the setting is applicable to all data engines.
```
sushant-suse:
```suggestion
              Defines the data engines that this setting applies to. 
              The setting can be mapped to specific engines with a boolean value indicating applicability. 
              If no engines are specified, the setting applies to all data engines by default.
```

k8s/crds.yaml
```diff
@@ -2612,6 +2612,20 @@ spec:
...
+            type: object
+          defaultsByEngine:
+            additionalProperties:
+              type: string
+            description: |-
+              Default values for the setting based on the data engine type.
+              If the value for a specific data engine type is not set, the default value will be used.
```
sushant-suse:
```suggestion
              Defines default values for the setting based on specific data engine types. 
              If no value is specified for a particular engine, the global default will be used.
```

k8s/crds.yaml
```diff
@@ -3555,6 +3562,13 @@ spec:
                 - enabled
                 - disabled
                 type: string
+              replicaRebuildBandwidthLimit:
+                description: ReplicaRebuildBandwidthLimit limits the write bandwidth
+                  (in megabytes per second) on the destination replica during rebuilding.
+                  Set to 0 to disable bandwidth limiting.
```
sushant-suse:
```suggestion
                description: ReplicaRebuildBandwidthLimit controls the maximum write bandwidth (in megabytes per second) allowed on the destination replica during the rebuilding process. 
                  Set this value to 0 to disable bandwidth limiting.
```

sushant-suse (changes requested):
Thanks.

datastore/longhorn.go
derekbit: After rethinking @c3y1huang, this should be available.
For example, if the user sets `default-replica-count: 3` (legacy format) in the `values.yaml`, we apply this value to all applicable data engines. If the user sets `default-replica-count: {"v1": "3", "v2": "2"}`, we apply the respective values to v1 and v2 respectively.

c3y1huang: Make sense. Thank you!

derekbit:
@coderabbitai review

upgrade/upgrade.go
```diff
@@ -252,6 +253,17 @@ func doResourceUpgrade(namespace string, lhClient *lhclientset.Clientset, kubeCl
...
+	if semver.Compare(lhVersionBeforeUpgrade, "v1.10.0") < 0 {
+		logrus.Info("Walking through the resource upgrade path v1.9.x to v1.10.0")
+		if err := v19xto1100.UpgradeResources(namespace, lhClient, kubeClient, resourceMaps); err != nil {
+			return err
+		}
+		forceResourceUpdate = true
+	}
+	if err := upgradeutil.UpdateResources(namespace, lhClient, resourceMaps, forceResourceUpdate); err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
derekbit: Good catch! Thanks.

derekbit:
@coderabbitai review

derekbit:
- V1 regression test
  - https://ci.longhorn.io/job/private/job/longhorn-tests-regression/9072/
  - https://ci.longhorn.io/job/private/job/longhorn-tests-regression/9074/
- V2 regression test
  - https://ci.longhorn.io/job/private/job/longhorn-tests-regression/9073/
  - https://ci.longhorn.io/job/private/job/longhorn-tests-regression/9075

datastore/longhorn.go
```diff
@@ -189,6 +194,36 @@ func (s *DataStore) syncSettingOrphanResourceAutoDeletionSettings() error {
...
+func (s *DataStore) syncConsolidatedV2DataEngineSettings() error {
+	settings := map[types.SettingName]types.SettingName{
+		types.SettingNameV2DataEngineHugepageLimit: types.SettingNameHugepageLimit,
+		types.SettingNameV2DataEngineCPUMask:       types.SettingNameCPUMask,
+		types.SettingNameV2DataEngineLogLevel:      types.SettingNameSpdkTgtLogLevel,
+		types.SettingNameV2DataEngineLogFlags:      types.SettingNameSpdkTgtLogFlags,
+	}
+
+	for oldSettingName, newSettingName := range settings {
+		if err := s.syncConsolidatedV2DataEngineSetting(oldSettingName, newSettingName); err != nil {
```
c3y1huang, line 219: Should we also consider deleting the old setting CRs after syncing and consolidation?

types/setting.go
```diff
@@ -280,8 +292,11 @@ type SettingDefinition struct {
 	Type        SettingType     `json:"type"`
 	Required    bool            `json:"required"`
 	ReadOnly    bool            `json:"readOnly"`
-	Default     string          `json:"default"`
-	Choices     []string        `json:"options,omitempty"` // +optional
+	// If DataEngineSpecific is false, the setting is applicable to both V1 and V2 Data Engine, and Default is a non-JSON string.
+	// If DataEngineSpecific is true, the setting is only applicable to V1 or V2 Data Engine, and Default is a JSON string.
+	DataEngineSpecific bool     `json:"dataEngineSpecific,omitempty"`
```
c3y1huang:
```suggestion
	DataEngineSpecific bool     `json:"dataEngineSpecific,omitempty"`  //+optional
```

c3y1huang: Suggest also reorganize for clarity.
```
type SettingDefinition struct {
	DisplayName string          `json:"displayName"`
	Description string          `json:"description"`
	Category    SettingCategory `json:"category"`
	Type        SettingType     `json:"type"`
	Required    bool            `json:"required"`
	ReadOnly    bool            `json:"readOnly"`
	Default     string          `json:"default"`
	Choices     []string        `json:"options,omitempty"` // +optional

	// Use map to present minimum and maximum value instead of using int directly, so we can omitempy and distinguish 0 or nil at the same time.
	ValueIntRange   map[string]int     `json:"range,omitempty"`      // +optional
	ValueFloatRange map[string]float64 `json:"floatRange,omitempty"` // +optional

	// If DataEngineSpecific is false, the setting is applicable to both V1 and V2 Data Engine, and Default is a non-JSON string.
	// If DataEngineSpecific is true, the setting is only applicable to V1 or V2 Data Engine, and Default is a JSON string.
	DataEngineSpecific bool `json:"dataEngineSpecific,omitempty"` // +optional
}
```

controller/controller_manager.go
```diff
@@ -262,28 +261,25 @@ func GetInstanceManagerCPURequirement(ds *datastore.DataStore, imName string) (*
...
-			guaranteedCPUSetting, err := ds.GetSettingWithAutoFillingRO(types.SettingNameGuaranteedInstanceManagerCPU)
+			guaranteedCPUPercentage, err := ds.GetSettingAsFloatByDataEngine(types.SettingNameGuaranteedInstanceManagerCPU, im.Spec.DataEngine)
 			if err != nil {
 				return nil, err
 			}
-			guaranteedCPUPercentage, err := strconv.ParseFloat(guaranteedCPUSetting.Value, 64)
+			allocatableMilliCPU := float64(kubeNode.Status.Allocatable.Cpu().MilliValue())
+			cpuRequest = int(math.Round(allocatableMilliCPU * guaranteedCPUPercentage / 100.0))
+		}
+	case longhorn.DataEngineTypeV2:
```
c3y1huang: The logic seems to be identical. Can we combine them into a single case?
```
```

datastore/longhorn.go
```diff
@@ -271,14 +272,88 @@ func (s *DataStore) applyCustomizedDefaultSettingsToDefinitions(customizedDefaul
...
+	for dataEngine := range values {
+		if _, ok := defaultValues[dataEngine]; !ok {
+			delete(values, dataEngine)
+		}
+	}
+
+	return convertValuesToJSONString(values)
+}
+
+func convertValuesToJSONString(values map[longhorn.DataEngineType]any) (string, error) {
```
c3y1huang: The function name feels too generic. How about something more specific to `DataEngineType`, like `convertDataEngineValuesToJSONString`?

types/setting.go
```diff
@@ -1942,116 +1921,389 @@ func GetDangerZoneSettings() sets.Set[SettingName] {
...
+		_, err = parseSingleBool(definition, value)
 	}
 
-	if value != "true" && value != "false" {
-		return fmt.Errorf("value %v should be true or false", value)
+	return err
+}
+
+// ParseBoolsInJSONFormat processes JSON-formatted input for data-engine-specific boolean settings.
+func ParseBoolsInJSONFormat(definition SettingDefinition, value string) (map[longhorn.DataEngineType]any, error) {
```
c3y1huang: The function name feels a bit too generic. Consider renaming it to reflect that it’s parsing data-engine-specific values. The same applies to the other new functions as well.

datastore/longhorn.go
derekbit: Yes, the settings will be removed in the following function `deleteReplacedSettings`.

types/setting.go
derekbit: Thank you. It is more clear!

datastore/longhorn.go

types/setting.go
derekbit: Good idea. Updated to `ParseDataEngineSpecificSettingStrings{Ints,Bools, ...}`. Thank you.

controller/controller_manager.go
derekbit: Addressed. Thank you.

derekbit:
@c3y1huang All comments are addressed. Thank you.

upgrade/v19xto1100/upgrade.go
```diff
@@ -0,0 +1,111 @@
+package v19xto1100
+
+import (
+	"github.com/cockroachdb/errors"
```
c3y1huang, line 4: Any specific reason we're using `github.com/cockroachdb/errors` here instead of `github.com/pkg/errors`?

derekbit: pkg/errors is deprecated
github.com/cockroachdb/errors  is 100% compatible with pkg/errors and is still actively maintained.

c3y1huang: ~https://github.com/longhorn/longhorn/issues/11438~
https://github.com/longhorn/longhorn/issues/11413

derekbit:
- Updated the regression test cases in https://github.com/longhorn/longhorn-tests/pull/2654
- e2e test cases will be updated in other PRs. 

cc @c3y1huang

derekbit:
> * V1 regression test
>   
>   * [ci.longhorn.io/job/private/job/longhorn-tests-regression/9072](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/9072/)
>   * [ci.longhorn.io/job/private/job/longhorn-tests-regression/9074](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/9074/)
> * V2 regression test
>   
>   * [ci.longhorn.io/job/private/job/longhorn-tests-regression/9073](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/9073/)
>   * [ci.longhorn.io/job/private/job/longhorn-tests-regression/9075](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/9075)

Rerunning after updating the regression test cases
- v1 full regression
  - https://10.115.5.5/job/private/job/longhorn-tests-regression/91/
- v2 full regression
  - https://10.115.5.5/job/private/job/longhorn-tests-regression/

derekbit:
Hello @c3y1huang 
All issues are resolved. Can you review the PR again? Thank you.

c3y1huang (dismissed):

datastore/longhorn.go
```diff
@@ -190,6 +194,36 @@ func (s *DataStore) syncSettingOrphanResourceAutoDeletionSettings() error {
...
+	}
+
+	for oldSettingName, newSettingName := range settings {
+		if err := s.syncConsolidatedV2DataEngineSetting(oldSettingName, newSettingName); err != nil {
+			return errors.Wrapf(err, "failed to sync consolidated v2 data engine setting %v to %v", oldSettingName, newSettingName)
+		}
+	}
+
+	return nil
+}
```
shuo-wu: NIT: Do we need to move these one-time functions to pkg/folder `upgrade`? The caller `UpdateCustomizedSettings` will be executed for every upgrade while this consolidation is required only in v1.10 upgrade

types/setting.go
```diff
@@ -294,9 +293,14 @@ type SettingDefinition struct {
...
 	Default     string          `json:"default"`
 	Choices     []string        `json:"options,omitempty"` // +optional
+
 	// Use map to present minimum and maximum value instead of using int directly, so we can omitempy and distinguish 0 or nil at the same time.
 	ValueIntRange   map[string]int     `json:"range,omitempty"`      // +optional
 	ValueFloatRange map[string]float64 `json:"floatRange,omitempty"` // +optional
+
+	// If DataEngineSpecific is false, the setting is applicable to both V1 and V2 Data Engine, and Default is a non-JSON string.
+	// If DataEngineSpecific is true, the setting is only applicable to V1 or V2 Data Engine, and Default is a JSON string.
+	DataEngineSpecific bool `json:"dataEngineSpecific,omitempty"` // +optional
```
shuo-wu, line 304: I am thinking if relying on this new field is really enough since there can be 3 cases:
1. For some settings, they can apply different values to both v1 and v2. For example, Instance manager guaranteed CPU settings
2. For some settings, they always apply the same value to both v1 and v2. For example, backup-related settings
3. For some settings, they can be applied to v1 or v2 engine only. For example, `Default Engine Image`, `Default Data Path`, `CPU Mask`

This option cannot distinguish case 2 from case 3. As a result, for case 3, sometimes you will use json string like `fmt.Sprintf("{%q:\"false\"}", longhorn.DataEngineTypeV1)`, sometimes you will directly use a non-json string

types/setting.go
```diff
@@ -1351,246 +1428,220 @@ var (
...
+	}
+
+	SettingDefinitionSpdkTgtLogLevel = SettingDefinition{
+		DisplayName:        "SPDK Target Daemon Log Level",
+		Description:        "The log level used in SPDK target daemon (spdk_tgt) of V2 Data Engine. Supported values are: Error, Warning, Notice, Info and Debug. By default Notice.",
+		Category:           SettingCategoryGeneral,
+		Type:               SettingTypeString,
+		Required:           true,
+		ReadOnly:           false,
+		DataEngineSpecific: true,
```
shuo-wu, line 1573: Should not this be false?

types/setting.go
```diff
@@ -1351,246 +1428,220 @@ var (
...
+	}
+
+	SettingDefinitionSpdkTgtLogFlags = SettingDefinition{
+		DisplayName:        "SPDK Target Daemon Log Flags",
+		Description:        "The log flags used in SPDK target daemon (spdk_tgt) of V2 Data Engine.",
+		Category:           SettingCategoryGeneral,
+		Type:               SettingTypeString,
+		Required:           false,
+		ReadOnly:           false,
+		DataEngineSpecific: true,
```
shuo-wu, line 1585: Should not this be false?

types/setting.go
```diff
@@ -1351,246 +1428,220 @@ var (
...
-		Category: SettingCategoryV2DataEngine,
-		Type:     SettingTypeInt,
-		Required: false,
-		ReadOnly: false,
-		Default:  "0",
+		Category:           SettingCategoryGeneral,
+		Type:               SettingTypeInt,
+		Required:           false,
+		ReadOnly:           false,
+		DataEngineSpecific: true,
```
shuo-wu, line 1598: Should not this be false? at least for now

types/setting.go
```diff
@@ -872,30 +913,33 @@ var (
...
-		ReadOnly:    false,
-		Default:     "true",
+		DisplayName:        "Disable Revision Counter",
+		Description:        "This setting is only for volumes created by UI. By default, this is true meaning Longhorn will not have revision counter file to track every write to the volume. During the salvage recovering, Longhorn will use the 'volume-head-xxx.img' file last modification time and file size to pick the replica candidate to recover the whole volume. If this setting is false, there will be a revision counter file to track every write to the volume. During salvage recovering Longhorn will pick the replica with largest revision counter as candidate to recover the whole volume.",
+		Category:           SettingCategoryGeneral,
+		Type:               SettingTypeBool,
+		Required:           true,
+		ReadOnly:           false,
+		DataEngineSpecific: true,
+		Default:            fmt.Sprintf("{%q:\"true\",%q:\"true\"}", longhorn.DataEngineTypeV1, longhorn.DataEngineTypeV2),
```
shuo-wu: There is no revision counter mechanism for v2 data engine

types/setting.go
```diff
@@ -1944,116 +1923,471 @@ func GetDangerZoneSettings() sets.Set[SettingName] {
 	return settingList
 }
 
-func validateBool(definition SettingDefinition, value string) error {
-	if definition.Type != SettingTypeBool {
-		return nil
+// IsJSONFormat checks if the input string starts with '{' indicating JSON format.
+func IsJSONFormat(value string) bool {
+	return strings.HasPrefix(strings.TrimSpace(value), "{")
+}
```
shuo-wu: NIT: Can we use std library like `json.Valid([]byte(validJSONString))` instead?

types/setting.go
```diff
@@ -1944,116 +1923,471 @@ func GetDangerZoneSettings() sets.Set[SettingName] {
...
+		}
+	}
+
+	return err
 }
 
-func validateInt(definition SettingDefinition, value string) error {
-	if definition.Type != SettingTypeInt {
-		return nil
+// ParseDataEngineSpecificSettingBools processes JSON-formatted input for data-engine-specific boolean settings.
```
shuo-wu: I am thinking if we can have a generic parsing function:
```
func ParseSettingValue(definition SettingDefinition, value string) (map[longhorn.DataEngineType]any, error) {
	trimmedValue := strings.TrimSpace(value)

	values := make(map[longhorn.DataEngineType]any)
	var jsonValues map[longhorn.DataEngineType]string

	if json.Valid([]byte(value)) {
		if !definition.DataEngineSpecific {
			return values, fmt.Errorf("JSON-formatted value is not supported for non-data-engine-specific setting")
		}
		if err := json.Unmarshal([]byte(trimmedValue), &jsonValues); err != nil {
			return values, errors.Wrapf(err, "failed to parse JSON-formatted value %v", trimmedValue)
		}
	} else {
		if definition.DataEngineSpecific {
			return values, fmt.Errorf("Non JSON-formatted value is not supported for data-engine-specific setting")
		}
		jsonValues[longhorn.DataEngineTypeAll] = trimmedValue
	}

	var errs []error
	for dataEngine, valueStr := range jsonValues {
		valueStr = strings.TrimSpace(valueStr)
		switch definition.Type {
		case SettingTypeBool:
			boolValue, err := strconv.ParseBool(valueStr)
			if err != nil {
				errs = append(errs, errors.Wrapf(err, "failed to parse bool value %s for data engine %s", valueStr, dataEngine))
				continue
			}
			values[dataEngine] = boolValue
		case SettingTypeInt:
			intValue, err := strconv.ParseInt(valueStr, 10, 64)
			if err != nil {
				errs = append(errs, errors.Wrapf(err, "failed to parse int value %s for data engine %s", valueStr, dataEngine))
				continue
			}
			values[dataEngine] = intValue
		case SettingTypeFloat:
			floatValue, err := strconv.ParseFloat(valueStr, 64)
			if err != nil {
				errs = append(errs, errors.Wrapf(err, "failed to parse float value %s for data engine %s", valueStr, dataEngine))
				continue
			}
			values[dataEngine] = floatValue
		case SettingTypeString:
		// ......
		default:
			return values, errors.Errorf("unknown setting type %v", definition.Type)
		}
	}

	return values, errors.Join(errs...)
}
```

Then the caller just needs to parse `map[longhorn.DataEngineType]any` to the correct format and validate the value

types/setting.go
derekbit: Good catch! Thank you.

types/setting.go
derekbit: Sure. Thank you.

types/setting.go
derekbit: V1 data engine doesn't support it, so I set it to data-engine-specific and only for v2 data engine.

types/setting.go
derekbit: It is only for v2 data engine, so it is data-engine-specific.

types/setting.go
derekbit: It is only for v2 data engine, so it is data-engine-specific.

types/setting.go
derekbit: Yes. That's why I add `DataEngineSpecific` field. If `DataEngineSpecific` is `true`, the value is a json-formatted string. Then, it can help distinguish 2 and 3.

datastore/longhorn.go
derekbit: IIRC, there is a technical concern. 
@COLDTURNIP Do you remember the context?

COLDTURNIP: @derekbit the related issue for replacing the setting from the old one: longhorn/longhorn#10816

types/setting.go
derekbit: Addressed. Thank you.

datastore/longhorn.go
derekbit: Thanks @COLDTURNIP. I think we can improve it in 10816.

types/setting.go
derekbit: `DataEngineSpecific: true` and `Default: {"v1": "value1", "v2": "value2"}`: The applicable engines are v1 and v2
`DataEngineSpecific: true` and `Default: {"v1": "value1"}`: The applicable engine is only v1
`DataEngineSpecific: true` and `Default: {"v2": "value1"}`: The applicable engine is only v2

`DataEngineSpecific: false` and `Default: any string`: The value is applied to all engines and is non-data-engine-specific.

shuo-wu: Then, for settings like `Default Engine Image` and `Default Data Path`, we should set `DataEngineSpecific` to `true` as they fit v1 data engine only.

derekbit: I was thinking these setting before. 

Default Engine Image is like a system level setting. To keep it simple, I made it non-engine-specific.

For Default Data Path, because we only support one default data path for now and the logic of adding disk can distinguish filesystem and block type disk, so I keep it   non-engine-specific for simplicity.

derekbit: Updated the descriptions. Thanks @shuo-wu.

datastore/longhorn.go
```diff
@@ -770,6 +775,44 @@ func (s *DataStore) GetSettingValueExisted(sName types.SettingName) (string, err
...
+// Returns error if the setting does not have a value for the given data engine.
+func (s *DataStore) GetSettingValueExistedByDataEngine(settingName types.SettingName, dataEngine longhorn.DataEngineType) (string, error) {
+	definition, ok := types.GetSettingDefinition(settingName)
+	if !ok {
+		return "", fmt.Errorf("setting %v is not supported", settingName)
+	}
+
+	if !definition.DataEngineSpecific {
+		return s.GetSettingValueExisted(settingName)
+	}
```
shuo-wu, line 950: This looks redundant. The majority of func `GetSettingValueExisted` is `GetSettingWithAutoFillingRO` invocation.

controller/instance_manager_controller.go
```diff
@@ -574,36 +574,41 @@ func (imc *InstanceManagerController) syncLogSettingsToInstanceManagerPod(im *lo
 
 	settingNames := []types.SettingName{
 		types.SettingNameLogLevel,
-		types.SettingNameV2DataEngineLogLevel,
-		types.SettingNameV2DataEngineLogFlags,
+		types.SettingNameSpdkTgtLogLevel,
+		types.SettingNameSpdkTgtLogFlags,
```
shuo-wu: Any specific reason for changing the setting name? IIRC, when we decide this kind of v2 setting naming, we don't want to expose too much about the implementation details, hence we call it `V2DataEngineXXXX` rather than `SPDKXXXX`

controller/node_controller.go
```diff
@@ -238,9 +238,13 @@ func (nc *NodeController) isResponsibleForSnapshot(obj interface{}) bool {
 }
 
 func (nc *NodeController) snapshotHashRequired(volume *longhorn.Volume) bool {
-	dataIntegrityImmediateChecking, err := nc.ds.GetSettingAsBool(types.SettingNameSnapshotDataIntegrityImmediateCheckAfterSnapshotCreation)
+	if volume.Spec.DataEngine == longhorn.DataEngineTypeV2 {
+		return false
+	}
```
shuo-wu: Hmmm... Why disable this for v2 data engine? IIRC, this is one prerequisite of v2 fast rebuilding feature

controller/volume_controller.go
```diff
@@ -5007,13 +5007,15 @@ func (c *VolumeController) shouldCleanUpFailedReplica(v *longhorn.Volume, r *lon
...
+				types.SettingNameFastReplicaRebuildEnabled, v.Spec.DataEngine)
+			fastReplicaRebuilding = false
 		}
-		if !V2DataEngineFastReplicaRebuilding {
-			log.Infof("Failed replica %v should be cleaned up blindly since setting %v is not enabled", r.Name, types.SettingNameV2DataEngineFastReplicaRebuilding)
+		if !fastReplicaRebuilding {
+			log.Infof("Failed replica %v should be cleaned up blindly since setting %v for data engine %v is not enabled",
+				r.Name, types.SettingNameFastReplicaRebuildEnabled, v.Spec.DataEngine)
 			return true
 		}
```
shuo-wu: Should we check this for v1 data engine either?

datastore/longhorn.go
```diff
@@ -437,13 +427,28 @@ func (s *DataStore) ValidateSetting(name, value string) (err error) {
...
+				if apierrors.IsNotFound(err) {
+					logrus.Warnf("Kubernetes node %s not found, skipping CPU mask validation for this node", lhnode.Name)
+					continue
+				}
+				return errors.Wrapf(err, "failed to get Kubernetes node %s for %v setting validation", lhnode.Name, types.SettingNameCPUMask)
+			}
+
+			if err := s.ValidateCPUMask(kubeNode, value); err != nil {
+				return err
+			}
```
shuo-wu: I am thinking whether Longhorn should error out when only a few nodes do not have enough CPU counts to apply the CPU mask value. What if those nodes are tainted or disable v2 data engine?

datastore/longhorn.go
```diff
@@ -519,53 +524,62 @@ func (s *DataStore) ValidateV2DataEngineEnabled(dataEngineEnabled bool) (ims []*
...
 
-	{
-		hugepageRequested := resource.MustParse(hugepageRequestedInMiB.Value + "Mi")
+	// hugepageRequestedInMiB is integer
+	hugepageRequested, err := resource.ParseQuantity(fmt.Sprintf("%dMi", hugepageRequestedInMiB))
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to parse hugepage value %qMi", hugepageRequestedInMiB)
+	}
+
+	_ims, err := s.ListInstanceManagersRO()
```
shuo-wu, line 637: It's better to filter out v1 IMs first

scheduler/replica_scheduler.go
```diff
@@ -615,13 +615,15 @@ func filterActiveReplicas(replicas map[string]*longhorn.Replica) map[string]*lon
 
 func (rcs *ReplicaScheduler) CheckAndReuseFailedReplica(replicas map[string]*longhorn.Replica, volume *longhorn.Volume, hardNodeAffinity string) (*longhorn.Replica, error) {
 	if types.IsDataEngineV2(volume.Spec.DataEngine) {
```
shuo-wu, line 619: As I mentioned in another comment, is it better to check the fast rebuilding setting for both data engines?

scheduler/replica_scheduler.go
```diff
@@ -715,13 +717,15 @@ func (rcs *ReplicaScheduler) RequireNewReplica(replicas map[string]*longhorn.Rep
 	}
 
 	if types.IsDataEngineV2(volume.Spec.DataEngine) {
```
shuo-wu, line 723: Same. Is it better to check the fast rebuilding setting for both data engines?

webhook/resources/node/validator.go
```diff
@@ -125,19 +125,19 @@ func (n *nodeValidator) Update(request *admission.Request, oldObj runtime.Object
 			if !datastore.ErrorIsNotFound(err) {
 				return werror.NewInvalidError(err.Error(), "")
 			}
-			logrus.Warnf("Kubernetes node %v has been deleted", oldNode.Name)
+			logrus.WithError(err).Warnf("Kubernetes node %v has been deleted", oldNode.Name)
 		} else {
-			allocatableCPU := float64(kubeNode.Status.Allocatable.Cpu().MilliValue())
-			instanceManagerCPUSetting, err := n.ds.GetSettingWithAutoFillingRO(types.SettingNameGuaranteedInstanceManagerCPU)
+			// TODO: Support v2 data engine
```
shuo-wu, line 130: Do you mean that `newNode.Spec.InstanceManagerCPURequest` is for v1 data engine only hence we don't need to check and compare it with the v2 setting?

upgrade/v19xto1100/upgrade.go
```diff
@@ -0,0 +1,111 @@
...
+		}
+		return errors.Wrapf(err, "failed to list all existing Longhorn settings during the setting upgrade")
+	}
+
+	// Update Setting CRs
+	var errs []error
+	for _, s := range settingMap {
+		definition, ok := types.GetSettingDefinition(types.SettingName(s.Name))
+		if !ok {
+			logrus.Warnf("Unknown setting %v found during upgrade, skipping", s.Name)
```
shuo-wu, line 94: Do we need to delete the deprecated v2 settings?

derekbit: Yes, we will delete them in https://github.com/longhorn/longhorn-manager/blob/master/datastore/longhorn.go#L272-L287

webhook/resources/node/validator.go
derekbit: Yes

controller/node_controller.go
derekbit: One question: Does v2 support `SettingNameSnapshotDataIntegrityImmediateCheckAfterSnapshotCreation`?

controller/instance_manager_controller.go
derekbit: I'm struggling with the renaming.
`SettingNameV2DataEngineLogLevel` is only for `spdk_tgt`. I was thinking we can make it more specific, or customer might think it is also applied to IM pod.

datastore/longhorn.go
derekbit: The main purpose is to avoid misusing longhorn datastore APIs. We can fall back to the legacy behavior. WDYT?

datastore/longhorn.go
derekbit: Good idea. We can skip the `tainted` or `v2 data engine disabled` nodes.

derekbit: Updated `getMinNumCPUs`. Can you check it again?

controller/volume_controller.go
derekbit: I remember you added the logic.
We don't need a check for v1 data engine. If the fast rebuilding is not enabled for v1 data engine, the failed repica can still be used for legacy rebuilding rebuilding.

scheduler/replica_scheduler.go
derekbit: Same reason as above.

scheduler/replica_scheduler.go
derekbit: Same reason as above.

datastore/longhorn.go
derekbit: Added a check. Thank you.

controller/node_controller.go
derekbit: After going through the code flow, v2 also supports `SettingNameSnapshotDataIntegrityImmediateCheckAfterSnapshotCreation`. Removed the `volume.Spec.DataEngine == longhorn.DataEngineTypeV2` check. Thank you.

controller/instance_manager_controller.go
derekbit: How about `dataEngineLogLevel`? It might be much clearer and won't introduce confusion.

shuo-wu: `DataEngineLogLevel ` looks good!

derekbit: Updated to DataEngineLogLevel! Thank you.

controller/volume_controller.go
shuo-wu: Just recalled the reason why we need this implementation: This is added when Longhorn cannot reuse failed replicas for the v2 data engine. Now I prefer to remove it directly. It's better not to aggressively clean up a failed replica even if the v2 data engine fast rebuilding is not enabled at this moment.

scheduler/replica_scheduler.go
shuo-wu: This is different from the above cleanup logic. Suppose Longhorn does not check the v1 data engine fast rebuilding setting in this failed replica CR pickup/schedule logic. In that case, Longhorn will blindly operate some critical fields like `replica.spec.failedAt` or `replica.spec.healthyAt` for those failed replicas even if the setting is disabled. 

Same for the below comment.

derekbit: A bit confused. For v1 data engine, it doesn't need to check fast rebuilding setting, since the failed resuable replica check works for both legacy replica rebuilding and fast rebulding.

scheduler/replica_scheduler.go
derekbit: Addressed. Thank you.

scheduler/replica_scheduler.go
derekbit: Added a comment as discussion. Thank you.

controller/volume_controller.go
derekbit: Removed. Thank you.

shuo-wu (approved):
LGTM

PR #3937, opened 2025-07-21
5215 configurable bkup block size

COLDTURNIP:

Issue longhorn/longhorn#5215

Implement configurable backup block size.

- Manager API accept a new volume field `backupBlockSize`, which is a string of quantity
- For CSI volume provisioning, backup block size can be specified in the storage class `parameters.backupBlockSize`
- Websocket change for Volume and Backup
  - Volume and Backup mutator set the `BackupBlockSize` to default 2MB
  - Volume and Backup validator prevent `BackupBlockSize` mutation after the valid value is set
- While creating backup CR from remote backup target, if the backup block size is absent, fallback to default 2MB

This PR depends on the following PRs

- longhorn/go-common-libs#122
- longhorn/backupstore#250
- longhorn/longhorn-engine#1533

The dependency commits must be resolved before merging.


COLDTURNIP:
@coderabbitai review

engineapi/backup_monitor.go
```diff
@@ -334,5 +334,6 @@ func (m *BackupMonitor) Close() {
 func getBackupParameters(backup *longhorn.Backup) map[string]string {
 	parameters := map[string]string{}
 	parameters[lhbackup.LonghornBackupParameterBackupMode] = string(backup.Spec.BackupMode)
+	parameters[lhbackup.LonghornBackupParameterBackupBlockSize] = string(backup.Spec.BackupBlockSize)
 	return parameters
```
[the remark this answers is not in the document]
COLDTURNIP: The `backup` here is a CR, and `backup.Spec.BackupBlockSize` is guaranteed to be a non-empty valid quantity string because of the mutator and validator webhooks. We don't need to check it again here in the CR controller.

derekbit:
Hello @COLDTURNIP CI failed.

derekbit:
Can we introduce a global setting for the backup block size as well?

webhook/resources/backup/mutator.go
```diff
@@ -3,8 +3,10 @@ package backup
 import (
 	"encoding/json"
 	"fmt"
-
+	"github.com/longhorn/backupstore"
```
shuo-wu: NIT: rearrange the imports. Same for other modified files

webhook/resources/backup/validator.go
```diff
@@ -63,9 +63,36 @@ func (b *backupValidator) Create(request *admission.Request, newObj runtime.Obje
...
+		return werror.NewInvalidError(fmt.Sprintf("%v is not a *longhorn.Backup", newObj), "")
+	}
+
+	// Allow backup block size mutation only when the existing obj is not set, or correcting the existed invalid value
+	isValidOldBackupBlockSize := types.ValidateBackupBlockSize(-1, oldBackup.Spec.BackupBlockSize) == nil
+	if isValidOldBackupBlockSize && oldBackup.Spec.BackupBlockSize != newBackup.Spec.BackupBlockSize {
+		err := fmt.Errorf("changing backup block size for backup %v is not supported", oldBackup.Name)
+		return werror.NewInvalidError(err.Error(), "")
+	}
+	if err := types.ValidateBackupBlockSize(-1, newBackup.Spec.BackupBlockSize); err != nil {
```
shuo-wu: Should we fetch the corresponding volume spec size (if exists) rather than using -1 here?

shuo-wu (commented):
In general LGTM

COLDTURNIP:
Global setting added. The default block size is now configurable.

k8s/pkg/apis/longhorn/v1beta2/volume.go
```diff
@@ -285,6 +285,9 @@ type VolumeSpec struct {
 	// +kubebuilder:validation:Enum=none;lz4;gzip
 	// +optional
 	BackupCompressionMethod BackupCompressionMethod `json:"backupCompressionMethod"`
+	// +kubebuilder:validation:Enum="2Mi";"16Mi"
+	// +optional
+	BackupBlockSize BackupBlockSize `json:"backupBlockSize"`
```
derekbit: Can we use `int64` instead? `BackupBlockSize` can be `2097152` and `16777216`. The unit conversion can be done in UI or client API side.
```suggestion
	BackupBlockSize int64 `json:"backupBlockSize"`
```

k8s/pkg/apis/longhorn/v1beta2/backup.go
```diff
@@ -51,6 +58,9 @@ type BackupSpec struct {
 	// Can be "full" or "incremental"
 	// +optional
 	BackupMode BackupMode `json:"backupMode"`
+	// The backup block size
+	// +optional
+	BackupBlockSize BackupBlockSize `json:"backupBlockSize"`
```
derekbit: Use int64 instead.
```suggestion
	BackupBlockSize int64 `json:"backupBlockSize"`
```

k8s/pkg/apis/longhorn/v1beta2/volume.go
COLDTURNIP: Sure. To align the `volume.spec.size`?

derekbit: Yes, it will be more generic.

COLDTURNIP: To go this approach, to the backup metadata stored inside backup volume, we don't need a duplicated [backup.BlockSize](https://github.com/longhorn/backupstore/blob/master/backupstore.go#L52) since it is identical with the one in the `backup.Parameters`. Will submit a backupstore PR to remove this.

webhook/resources/backup/validator.go
```diff
@@ -90,3 +96,46 @@ func (b *backupValidator) Create(request *admission.Request, newObj runtime.Obje
...
+		return err
+	}
+
+	var volumeSize int64 = -1
+	if backupVolumeName, exist := backup.Labels[types.LonghornLabelBackupVolume]; exist {
+		volume, err := b.ds.GetVolumeRO(backupVolumeName)
+		if err != nil {
+			if !datastore.ErrorIsNotFound(err) {
+				return fmt.Errorf("fail to get volume size of %s from datastore: %w", backupVolumeName, err)
+			}
```
shuo-wu: I am thinking if this should be best-effort rather than erroring out, since the associated volume of a backup may not be in this cluster.

webhook/resources/volume/mutator.go
```diff
@@ -2,6 +2,7 @@ package volume
 
 import (
 	"fmt"
+	"k8s.io/apimachinery/pkg/api/resource"
```
shuo-wu: NIT: rearrange import

webhook/resources/volume/validator.go
```diff
@@ -264,6 +268,16 @@ func (v *volumeValidator) Update(request *admission.Request, oldObj runtime.Obje
 		}
 	}
 
+	// Allow backup block size mutation only when the existing obj is not set, or correcting the existed invalid value
+	isValidOldBackupBlockSize := types.ValidateBackupBlockSize(oldVolume.Spec.Size, oldVolume.Spec.BackupBlockSize) == nil
+	if isValidOldBackupBlockSize && oldVolume.Spec.BackupBlockSize != newVolume.Spec.BackupBlockSize {
+		err := fmt.Errorf("changing backup block size for volume %v is not supported", oldVolume.Name)
+		return werror.NewInvalidError(err.Error(), "")
```
shuo-wu, line 275: I am thinking how Longhorn can handle the case containing the following conditions:
1. The default block size is set to 16Mi
2. The backup CR contains an empty `Spec.BackupBlockSize`, and the actual size it uses is 2Mi
3. There is no corresponding volume in the cluster, or `volume.Spec.BackupBlockSize` is not 2Mi

COLDTURNIP: @shuo-wu The configurable default block size effects only on volume creating. To create a new backup:

1. Backup is created from a volume in the cluster: the `backup.Spec.BackupBlockSize` is copied from the corresponding volume.
2. Backup is created from the backup target: the `backup.Spec.BackupBlockSize` is recovered from the `.Parameters["backup-block-size"]` of the backup metadata `backup_*.cfg`. If the `.Parameters["backup-block-size"]` is empty or absent, it means that the backup contains legacy 2Mi block size.
    ```json
    {
      "Name": "backup-bcca5b4023f04b7d",
      "VolumeName": "vol-16mb",
      "SnapshotName": "cb694d55-9d18-49df-ae9f-25063be19521",
      "SnapshotCreatedAt": "2025-08-01T08:54:32Z",
      "CreatedTime": "2025-08-01T08:54:34Z",
      "Size": "0",
      "Labels": {
        "VolumeRecurringJobInfo": "{}",
        "longhorn.io/volume-access-mode": "rwo"
      },
      "Parameters": {
        "backup-block-size": "16777216",
        "backup-mode": "incremental"
      },
      "IsIncremental": false,
      "CompressionMethod": "lz4",
      "NewlyUploadedDataSize": "0",
      "ReUploadedDataSize": "0",
      "ProcessingBlocks": {},
      "SingleFile": {
        "FilePath": ""
      }
    }
    ```

In addition, to the existing volume & backup CRs, the `*.Spec.BackupBlockSize` is always set to 2Mi while upgrading to v1.10.x.

derekbit:
@COLDTURNIP Could you help update the PR?

derekbit:
@coderabbitai review

webhook/resources/volume/mutator.go
derekbit: Not organized yet
https://github.com/longhorn/longhorn/wiki/coding-convention#organize-package-imports

types/setting.go
```diff
@@ -1367,6 +1373,17 @@ var (
...
+	SettingDefinitionDefaultBackupBlockSize = SettingDefinition{
+		DisplayName: "Default Backup Block Size",
+		Description: "This setting controls the default backup block size to create the new volume.",
+		Category:    SettingCategoryBackup,
+		Type:        SettingTypeString,
+		Required:    true,
+		ReadOnly:    false,
+		Default:     "2Mi",
+		Choices:     []string{"2Mi", "16Mi"},
+	}
```
```suggestion
	SettingDefinitionDefaultBackupBlockSize = SettingDefinition{
		DisplayName: "Default Backup Block Size",
		Description: "In MiB. This setting controls the default backup block size to create the new volume.",
		Category:    SettingCategoryBackup,
		Type:        SettingTypeString,
		Required:    true,
		ReadOnly:    false,
		Default:     "2",
		Choices:     []string{"2", "16"},
	}
```

api/volume.go
```diff
@@ -172,6 +172,11 @@ func (s *Server) VolumeCreate(rw http.ResponseWriter, req *http.Request) error {
 		return errors.Wrap(err, "failed to parse snapshot max size")
 	}
 
+	backupBlockSize, err := util.ConvertSize(volume.BackupBlockSize)
+	if err != nil {
+		return errors.Wrap(err, "failed to parse backup block size %v")
```
derekbit:
```suggestion
		return errors.Wrapf(err, "failed to parse backup block size %v", backupBlockSize)
```

client/generated_volume.go
```diff
@@ -141,8 +143,6 @@ type VolumeOperations interface {
 
 	ActionCancelExpansion(*Volume) (*Volume, error)
 
-	ActionOfflineReplicaRebuilding(*Volume) (*Volume, error)
-
```
derekbit: Shouldn't remove this method.

client/generated_volume.go
```diff
@@ -265,15 +265,6 @@ func (c *VolumeClient) ActionCancelExpansion(resource *Volume) (*Volume, error)
...
 
-func (c *VolumeClient) ActionOfflineReplicaRebuilding(resource *Volume) (*Volume, error) {
-
-	resp := &Volume{}
-
-	err := c.rancherClient.doAction(VOLUME_TYPE, "offlineReplicaRebuilding", &resource.Resource, nil, resp)
-
-	return resp, err
-}
-
```
derekbit: Shouldn't remove this method.

controller/backup_volume_controller.go
```diff
@@ -331,6 +335,16 @@ func (bvc *BackupVolumeController) reconcile(backupVolumeName string) (err error
 				if accessMode, exist := backupInfo.Labels[types.GetLonghornLabelKey(types.LonghornLabelVolumeAccessMode)]; exist {
 					backupLabelMap[types.GetLonghornLabelKey(types.LonghornLabelVolumeAccessMode)] = accessMode
 				}
+				backupBlockSizeParam := backupInfo.Parameters[lhbackup.LonghornBackupParameterBackupBlockSize]
+				if blockSizeBytes, convertErr := util.ConvertSize(backupBlockSizeParam); convertErr != nil {
+					log.WithError(convertErr).Warnf("Invalid backup block size string from remote backup %v: %v", backupName, backupBlockSizeParam)
```
derekbit: Should we set the state of the backup CR to ERROR state?

upgrade/v19xto1100/upgrade.go
```diff
@@ -0,0 +1,82 @@
+package v19xto1100
+
+import (
+	"github.com/pkg/errors"
+	"strconv"
```
derekbit:
```suggestion
	"strconv"
	
	"github.com/pkg/errors"
```

k8s/pkg/apis/longhorn/v1beta2/backup.go
```diff
@@ -51,6 +51,9 @@ type BackupSpec struct {
 	// Can be "full" or "incremental"
 	// +optional
 	BackupMode BackupMode `json:"backupMode"`
+	// The backup block size
+	// +optional
+	BackupBlockSize string `json:"backupBlockSize"`
```
derekbit: Why not use `int64`?

COLDTURNIP: To align the `BackupStatus.Size`. Both `string` or `int64` are OK to me.

controller/backup_volume_controller.go
COLDTURNIP: Good idea. Let me add this.

client/generated_volume.go
COLDTURNIP: Sorry for accidentally removal.

k8s/pkg/apis/longhorn/v1beta2/backup.go
derekbit: Got it. Not sure why `status. Size` is a string, but I prefer the `int64` instead.

client/generated_volume.go
derekbit: No worries.

k8s/pkg/apis/longhorn/v1beta2/backup.go
COLDTURNIP: Thanks. Will change to `int64`.

controller/backup_volume_controller.go
COLDTURNIP: After think twice, this means to create a new backup CR, set the block size to some default value to pass the validator, and update the status by backupVolume controller. Maybe it would be better to record in the event recorder without creating the backup CR. WDYT? @derekbit @shuo-wu

derekbit: >  Maybe it would be better to record in the event recorder without creating the backup CR. 

Will it result in flooding events?

derekbit: Checked
```
if backupInfo, err := backupTargetClient.BackupGet(backupURL, backupTargetClient.Credential); err != nil && !types.ErrorIsNotFound(err)
```

The backup CR is blindly created even though failed to execute `BackupGet()`.
Current implementation might be fine.

k8s/pkg/apis/longhorn/v1beta2/volume.go
```diff
@@ -285,6 +285,10 @@ type VolumeSpec struct {
 	// +kubebuilder:validation:Enum=none;lz4;gzip
 	// +optional
 	BackupCompressionMethod BackupCompressionMethod `json:"backupCompressionMethod"`
+	// +kubebuilder:validation:Type=string
+	// +kubebuilder:validation:Enum="2097152";"16777216"
+	// +optional
+	BackupBlockSize int64 `json:"backupBlockSize,string"`
```
derekbit, line 292: Why need `string` in the JSON?

k8s/pkg/apis/longhorn/v1beta2/backup.go
```diff
@@ -51,6 +51,9 @@ type BackupSpec struct {
 	// Can be "full" or "incremental"
 	// +optional
 	BackupMode BackupMode `json:"backupMode"`
+	// The backup block size
+	// +optional
+	BackupBlockSize int64 `json:"backupBlockSize,string"`
```
derekbit, line 58: Why need `string` in the JSON?

types/setting.go
```diff
@@ -1367,6 +1374,17 @@ var (
 		},
 	}
 
+	SettingDefinitionDefaultBackupBlockSize = SettingDefinition{
+		DisplayName: "Default Backup Block Size",
+		Description: "In MiB. This setting controls the default backup block size to create the new volume.",
+		Category:    SettingCategoryBackup,
+		Type:        SettingTypeString,
```
```suggestion
		Type:        SettingTypeInt,
```

controller/backup_volume_controller.go
COLDTURNIP: Per discussion, we still need a way to indicate the backup is corrupted, to prevent creating a broken volume from such problematic backup. Will create a new backup with block size `-1`.

k8s/pkg/apis/longhorn/v1beta2/volume.go
COLDTURNIP: Just align the `Spec.Size`.

k8s/pkg/apis/longhorn/v1beta2/backup.go
COLDTURNIP: Just align the other size-like fields.

derekbit:
Hello @COLDTURNIP CI failed.

k8s/pkg/apis/longhorn/v1beta2/backup.go
derekbit: Sould we add
```
	// +kubebuilder:validation:Type=string
	// +kubebuilder:validation:Enum="2097152";"16777216"
```

webhook/resources/backup/validator.go
```diff
@@ -92,3 +98,41 @@ func (b *backupValidator) Create(request *admission.Request, newObj runtime.Obje
 
 	return nil
 }
+
+func (b *backupValidator) Update(request *admission.Request, oldObj runtime.Object, newObj runtime.Object) error {
+	oldBackup, ok := oldObj.(*longhorn.Backup)
+	if !ok {
+		return werror.NewInvalidError(fmt.Sprintf("%v is not a *longhorn.Backup", oldObj), "")
```
derekbit:
```suggestion
		return werror.NewInvalidError(fmt.Sprintf("oldOjb %v is not a longhorn. Backupp", oldObj), "")
```

webhook/resources/backup/validator.go
```diff
@@ -92,3 +98,41 @@ func (b *backupValidator) Create(request *admission.Request, newObj runtime.Obje
...
+	}
+	newBackup, ok := newObj.(*longhorn.Backup)
+	if !ok {
+		return werror.NewInvalidError(fmt.Sprintf("%v is not a *longhorn.Backup", newObj), "")
```
derekbit:
```suggestion
		return werror.NewInvalidError(fmt.Sprintf("newObj %v is not a longhorn.Backup", newObj), "")
```

webhook/resources/backup/validator.go
```diff
@@ -92,3 +98,41 @@ func (b *backupValidator) Create(request *admission.Request, newObj runtime.Obje
...
+	}
+
+	// Allow backup block size mutation only when the existing obj is not set, or correcting the existed invalid value
+	isValidOldBackupBlockSize := b.validateBackupBlockSize(oldBackup, true) == nil
+	if isValidOldBackupBlockSize && oldBackup.Spec.BackupBlockSize != newBackup.Spec.BackupBlockSize {
+		err := fmt.Errorf("changing backup block size for backup %v is not supported", oldBackup.Name)
+		return werror.NewInvalidError(err.Error(), "")
+	}
```
derekbit: If `oldBackup.Spec.BackupBlockSize` is `-1` and `newBackup.Spec.BackupBlockSize` is `2MiB`, the `isValidOldBackupBlockSize` will be true, but the request will be rejected. Isn't it? Then, user cannot correct the value.

types/types.go
```diff
@@ -1031,6 +1031,17 @@ func ValidateOfflineRebuild(value longhorn.VolumeOfflineRebuilding) error {
 	return nil
 }
 
+// ValidateBackupBlockSize skips the volume size check if volSize set to negative.
+func ValidateBackupBlockSize(volSize int64, backupBlockSize int64) error {
+	if backupBlockSize != BackupBlockSize2Mi && backupBlockSize != BackupBlockSize16Mi {
+		return fmt.Errorf("unrecognized of BackupBlockSize setting: %v", backupBlockSize)
```
derekbit:
```suggestion
		return fmt.Errorf("unsupported BackupBlockSize %v", backupBlockSize)
```

webhook/resources/backup/mutator.go
```diff
@@ -156,5 +154,10 @@ func mutate(newObj runtime.Object) (admission.PatchOps, error) {
 		patchOps = append(patchOps, patchOp)
 	}
 
+	// Mutate the backup block size to the legacy default one if not set
+	if backup.Spec.BackupBlockSize == 0 {
+		patchOps = append(patchOps, fmt.Sprintf(`{"op": "replace", "path": "/spec/backupBlockSize", "value": "%d"}`, types.BackupBlockSize2Mi))
```
mantissahz: Should it be set as `Volume.Spec.BackupBlockSize` or `types.SettingNameDefaultBackupBlockSize` when creating a backup?

COLDTURNIP: While creating a backup from volume snapshot, it would be better to set the block size from `Volume.Spec.BackupBlockSize`, and let the validator rejects the other cases. I don't think the backup CR creation should follow the `types.SettingNameDefaultBackupBlockSize` to ensure the consistency between backups that belong to the same volume.

webhook/resources/backup/mutator.go
```diff
@@ -109,6 +109,12 @@ func (b *backupMutator) Create(request *admission.Request, newObj runtime.Object
 		patchOps = append(patchOps, fmt.Sprintf(`{"op": "replace", "path": "/spec/backupMode", "value": "%s"}`, string(longhorn.BackupModeIncremental)))
 	}
 
+	if backup.Spec.BackupBlockSize == 0 {
+		if volume != nil {
```
mantissahz: What if `volume == nil`? For example, A backup pulled from a backup store does not have its volume in this cluster.
Then the create operation will fail at the validator?

k8s/pkg/apis/longhorn/v1beta2/backup.go
```diff
@@ -51,6 +51,11 @@ type BackupSpec struct {
 	// Can be "full" or "incremental"
 	// +optional
 	BackupMode BackupMode `json:"backupMode"`
+	// The backup block size
```
derekbit: nit: add an description for the value `-1`

webhook/resources/backup/validator.go
```diff
@@ -92,3 +98,41 @@ func (b *backupValidator) Create(request *admission.Request, newObj runtime.Obje
...
+
+	return nil
+}
+
+func (b *backupValidator) validateBackupBlockSize(backup *longhorn.Backup, allowInvalid bool) error {
+	if allowInvalid && backup.Spec.BackupBlockSize == types.BackupBlockSizeInvalid {
+		return nil
+	}
```
derekbit, line 129: Add a comment for explaining when the backup block size is BackupBlockSizeInvalid and why we blindly return nil here.

types/types.go
```diff
@@ -1031,6 +1031,17 @@ func ValidateOfflineRebuild(value longhorn.VolumeOfflineRebuilding) error {
 	return nil
 }
 
+// ValidateBackupBlockSize skips the volume size check if volSize set to negative.
+func ValidateBackupBlockSize(volSize int64, backupBlockSize int64) error {
+	if backupBlockSize != BackupBlockSize2Mi && backupBlockSize != BackupBlockSize16Mi {
+		return fmt.Errorf("unsupported BackupBlockSize: %v", backupBlockSize)
+	}
+	if volSize >= 0 && volSize%backupBlockSize != 0 {
+		return fmt.Errorf("invalid value of Size setting: size %v must be an integer multiple of the backup block size %v", volSize, backupBlockSize)
```
derekbit: nit: make the error message more clear.
```suggestion
		return fmt.Errorf("volume size %v must be an integer multiple of the backup block size %v", volSize, backupBlockSize)
```

k8s/pkg/apis/longhorn/v1beta2/volume.go
```diff
@@ -285,6 +285,10 @@ type VolumeSpec struct {
 	// +kubebuilder:validation:Enum=none;lz4;gzip
 	// +optional
 	BackupCompressionMethod BackupCompressionMethod `json:"backupCompressionMethod"`
+	// +kubebuilder:validation:Type=string
+	// +kubebuilder:validation:Enum="2097152";"16777216"
+	// +optional
```
derekbit, line 291: Let's add a description for the newly introduced `BackupBlockSize`.

webhook/resources/backup/mutator.go
```diff
@@ -65,28 +64,29 @@ func (b *backupMutator) Create(request *admission.Request, newObj runtime.Object
...
+	//check snapshot name, if not exist, return error
+
+	volumeName, isExist := backup.Labels[types.LonghornLabelBackupVolume]
+	if !isExist {
+		if volume == nil {
+			err := errors.Wrapf(err, "cannot find the backup volume label for backup %v", backup.Name)
+			return nil, werror.NewInvalidError(err.Error(), "")
+		}
+		metaLabels[types.LonghornLabelBackupVolume] = volume.Name
 	}
```
derekbit: Why do we need to refactor this part?
After the change, it always tries to get Volume and Snapshot although `LonghornLabelBackupVolume` label is already existing and introduces necessary GET operations.

webhook/resources/backup/mutator.go
COLDTURNIP: To the backup created from the remote backup store, the backup volume controller will set the block size from the backup's metadata. If the block size is absent in the metadata, [the legacy `2MiB` size will be set to the backup CR](https://github.com/longhorn/longhorn-manager/pull/3937/files#diff-4967e3520dbda6bc373b2fea83e911dd65aac0ab4f48a53f977413620ce4ba7dR347-R348); otherwise, the block size should be a valid or invalid (`-1`) value, and be validated by the verifier.

webhook/resources/backup/mutator.go
COLDTURNIP: Reorganized the logic and using a reusable volume getter.

derekbit:
@COLDTURNIP Please help fix the conflicts. Thank you.

datastore/datastore.go
```diff
@@ -328,3 +329,18 @@ func ErrorIsNotFound(err error) bool {
...
+	var ret R
+	var err error
+	once := new(sync.Once)
+	return func() (R, error) {
+		once.Do(func() {
+			ret, err = getter()
+		})
+		return ret, err
+	}
+}
```
derekbit: Why do we need this?

webhook/resources/backup/mutator.go
```diff
@@ -65,13 +64,19 @@ func (b *backupMutator) Create(request *admission.Request, newObj runtime.Object
 		metaLabels = map[string]string{}
 	}
 
-	//check snapshot name, if not exist, return error
+	getVolume := datastore.LazyGetter(func() (*longhorn.Volume, error) {
```
derekbit: What's the purpose of the LazyGetter?

COLDTURNIP: To reuse the volume CR we've got. With this cache, the mutator only need one GET operation for the volume CR if there are multiple mutation needed, including the backup volume, the backup target, the access mode, and the block size.

derekbit: What does `multiple mutation` mean here? The volume CR is got from local cache. It seems there is no too much benefit from the lazy get.

COLDTURNIP: I see. Let me remove it.

derekbit:
@COLDTURNIP CI failed

manager/engine.go
```diff
@@ -271,12 +276,13 @@ func (m *VolumeManager) BackupSnapshot(backupName, backupTargetName, volumeName,
 			},
 		},
 		Spec: longhorn.BackupSpec{
-			SnapshotName: snapshotName,
-			Labels:       labels,
-			BackupMode:   longhorn.BackupMode(backupMode),
+			SnapshotName:    snapshotName,
+			Labels:          labels,
+			BackupMode:      longhorn.BackupMode(backupMode),
+			BackupBlockSize: blockSize,
```
derekbit: Can we move the `BackupBlockSize` assignment to the mutating webhook?

manager/engine.go
```diff
@@ -274,6 +274,8 @@ func (m *VolumeManager) BackupSnapshot(backupName, backupTargetName, volumeName,
 			SnapshotName: snapshotName,
 			Labels:       labels,
 			BackupMode:   longhorn.BackupMode(backupMode),
+
+			// BackupBlockSize will be copied from the corresponding volume in mutator webhook.
```
derekbit: nit: the comment seems not required.

derekbit (approved):
LGTM

types/setting.go
```diff
@@ -1457,6 +1466,17 @@ var (
...
+	SettingDefinitionDefaultBackupBlockSize = SettingDefinition{
+		DisplayName: "Default Backup Block Size",
+		Description: "In MiB. This setting controls the default backup block size to create the new volume.",
+		Category:    SettingCategoryBackup,
+		Type:        SettingTypeInt,
+		Required:    true,
+		ReadOnly:    false,
+		Default:     "2",
+		Choices:     []any{int64(2), int64(16)},
+	}
```
derekbit: nit:
```suggestion
	SettingDefinitionDefaultBackupBlockSize = SettingDefinition{
		DisplayName: "Default Backup Block Size",
		Description: "In MiB. This setting controls the default backup block size to create the new volume.",
		Category:    SettingCategoryBackup,
		Type:        SettingTypeInt,
		Required:    true,
		ReadOnly:    false,
		DataEngineSpecific: false,
		Choices:     []any{int64(2), int64(16)},
		Default:     "2",
	}
```

types/setting.go
```diff
@@ -1457,6 +1466,18 @@ var (
 		},
 	}
 
+	SettingDefinitionDefaultBackupBlockSize = SettingDefinition{
+		DisplayName:        "Default Backup Block Size",
+		Description:        "In MiB. This setting controls the default backup block size to create the new volume.",
```
derekbit: Need to mention it only supports 2 and 16.

types/setting.go
```diff
@@ -1458,6 +1466,18 @@ var (
 		},
 	}
 
+	SettingDefinitionDefaultBackupBlockSize = SettingDefinition{
+		DisplayName:        "Default Backup Block Size",
+		Description:        "2 or 16 in MiB. This setting controls the default backup block size to create the new volume.",
```
derekbit: cc @sushant-suse Please help review it. Thank you.

sushant-suse: Can we write it as:
```
Description: Specifies the default backup block size, in MiB, used when creating a new volume. Supported values are 2 or 16.
```

I don't know why but I am not able to suggest.

COLDTURNIP: Description reworded. Thanks for the suggestion.

derekbit (approved):

PR #3939, opened 2025-07-22
feat(v2): support interrupt mode

c3y1huang:

Issue longhorn/longhorn#9834, longhorn/longhorn#9419

Support interrupt mode for V2 volumes.


c3y1huang:
@coderabbitai review

types/setting.go
```diff
@@ -1441,6 +1444,19 @@ var (
 		Default:     "0x1",
 	}
 
+	SettingDefinitionV2DataEngineInterruptModeEnabled = SettingDefinition{
+		DisplayName: "Enable Interrupt Mode for V2 Data Engine",
+		Description: "Specifies whether the Storage Performance Development Kit (SPDK) target daemon should run in interrupt mode. " +
+			"This setting is applicable only when the V2 Data Engine is enabled. \n\n" +
+			"  - `true`: Enables interrupt mode, which may reduce CPU usage. \n\n" +
+			"  - `false`: Uses polling mode for maximum performance. \n\n",
+		Category: SettingCategoryDangerZone,
```
shuo-wu: Should Longhorn delete the IM pods when this setting is out of sync?

c3y1huang: Yes, Longhorn will delete the instance manager pod if the setting is out of sync, and there are no V2 volumes attached.

https://github.com/longhorn/longhorn-manager/blob/e54c383996882eea4a9aef9b3da876c0becdcbc7/controller/instance_manager_controller.go#L633-L643

derekbit (commented):
Is it possible some nodes use interruption mode and some use polling mode?

c3y1huang:
> Is it possible some nodes use interruption mode and some use polling mode?

Yes, I think it’s possible for SPDK to have some nodes using interrupt mode while others use polling mode. As far as I know, this is controlled per reactor/controller at runtime, so mixing modes across nodes should be feasible. However, this PR doesn’t support it, and I haven’t verified it yet. We can consider improving this in the future.

derekbit:
> > Is it possible some nodes use interruption mode and some use polling mode?
> 
> Yes, I think it’s possible for SPDK to have some nodes using interrupt mode while others use polling mode. As far as I know, this is controlled per reactor/controller at runtime, so mixing modes across nodes should be feasible. However, this PR doesn’t support it, and I haven’t verified it yet. We can consider improving this in the future.

Sounds good. Could you create an improvement ticket? We can improve it later. Thank you.

c3y1huang:
> > > Is it possible some nodes use interruption mode and some use polling mode?
> > 
> > 
> > Yes, I think it’s possible for SPDK to have some nodes using interrupt mode while others use polling mode. As far as I know, this is controlled per reactor/controller at runtime, so mixing modes across nodes should be feasible. However, this PR doesn’t support it, and I haven’t verified it yet. We can consider improving this in the future.
> 
> Sounds good. Could you create an improvement ticket? We can improve it later. Thank you.

https://github.com/longhorn/longhorn/issues/11677

types/setting.go
```diff
@@ -1571,6 +1574,21 @@ var (
 		Default:            fmt.Sprintf("{%q:\"0x1\"}", longhorn.DataEngineTypeV2),
 	}
 
+	SettingDefinitionDataEngineInterruptModeEnabled = SettingDefinition{
+		DisplayName: "Enable Interrupt Mode for Data Engine",
+		Description: "Specifies whether the Storage Performance Development Kit (SPDK) target daemon should run in interrupt mode. " +
+			"This setting is applicable only when the V2 Data Engine is enabled. \n\n" +
+			"  - DO NOT CHANGE THIS SETTING WITH ATTACHED VOLUMES. Longhorn will block this setting update when there are attached v2 volumes. \n\n" +
+			"  - `true`: Enables interrupt mode, which may reduce CPU usage. \n\n" +
+			"  - `false`: Uses polling mode for maximum performance. \n\n",
```
derekbit, line 1583: @sushant-suse Please help review the description. Thanks.

derekbit (approved):
LGTM.

derekbit:
@mergify backport v1.10.x

types/setting.go
sushant-suse: LGTM, Thanks.

PR #3945, opened 2025-07-28
fix: shouldn't count replica marked as deleted as healthy and active replicas

derekbit:

Issue longhorn/longhorn#11388

After a replica’s `spec.deletionTimestamp` is set but before it is actually deleted, it is still considered a healthy replica. During the cleanup of extra replicas, the volume controller counts this replica as healthy and may proceed to delete additional replicas, potentially resulting in the removal of too many replicas.

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -890,7 +890,8 @@ func isAutoSalvageNeeded(rs map[string]*longhorn.Replica) bool {
 	if isFirstAttachment(rs) {
 		return areAllReplicasFailed(rs)
 	}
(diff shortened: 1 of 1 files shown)


derekbit:
- v1 volume full regression
  - https://ci.longhorn.io/job/private/job/longhorn-tests-regression/9030/
  - All are passed
- v2 volume: -vs --count 5 test_scheduling.py::test_replica_auto_balance_node_least_effort
  - https://ci.longhorn.io/job/private/job/longhorn-tests-regression/9031/
  - Passed

c3y1huang (approved):

derekbit:

PR #3950, opened 2025-08-01
feat(setting): rename `V2DataEngineRebuildingMbytesPerSecond` to `ReplicuildBandwidthLimit`

derekbit:

Issue Longhorn/longhorn#11403  Longhorn/longhorn#10926


mantissahz:
https://github.com/longhorn/longhorn-manager/blob/c933a9ea1442475c24a56b551600eead81fbbb7b/controller/engine_controller.go#L1185-L1186
Here we need to modify the spec field because the field name has changed.

derekbit:
> https://github.com/longhorn/longhorn-manager/blob/c933a9ea1442475c24a56b551600eead81fbbb7b/controller/engine_controller.go#L1185-L1186
> 
> 
> Here we need to modify the spec field because the field name has changed.

Yes, updated. Thanks.

mantissahz (approved):
LGTM

PR #3954, opened 2025-08-04
fix(webhook): handle backup validation when snapshot name is empty

derekbit:

Issue Longhorn/longhorn#11416

Improve backup webhook validation and mutation to handle cases where backup.Spec.SnapshotName might be empty, which can occur during certain backup operations.

### DIFF
--- a/webhook/resources/backup/validator.go
+++ b/webhook/resources/backup/validator.go
@@ -66,26 +66,28 @@ func (b *backupValidator) Create(request *admission.Request, newObj runtime.Obje
 		return werror.NewInvalidError(fmt.Sprintf("backup target %s is not available", backupTargetName), "")
 	}
 
-	//check if label volume name matches snapshot volume name
-	labelVolumeName := backup.Labels[types.LonghornLabelBackupVolume]
-	snapshot, err := b.ds.GetSnapshotRO(backup.Spec.SnapshotName)
-	if err != nil {
-		return werror.NewInvalidError(fmt.Sprintf("snapshot %v is invalid", snapshot), "")
-	}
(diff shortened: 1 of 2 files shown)


mantissahz (approved):
In general, LGTM

webhook/resources/backup/validator.go
```diff
@@ -66,26 +66,28 @@ func (b *backupValidator) Create(request *admission.Request, newObj runtime.Obje
...
-
-	if labelVolumeName != snapshotVolumeName {
-		return werror.NewInvalidError(fmt.Sprintf("snapshot volume name %s and label volume name %s does not match", snapshotVolumeName, labelVolumeName), "")
-	}
-
-	//check if volume backup target matches labelbackup target
-	volumeBackupTargetName := volume.Spec.BackupTargetName
-	if volumeBackupTargetName != backupTargetName {
-		return werror.NewInvalidError(fmt.Sprintf("volume backup target %s and label backup target %s does not match", volumeBackupTargetName, backupTargetName), "")
+	if backup.Spec.SnapshotName != "" {
```
mantissahz, line 69: NIT:
```suggestion
    if backup.Spec.SnapshotName == "" {
        return nil
    }
```

derekbit: What do you mean?

derekbit: Got your point...

mantissahz: We can use this check statement at line 68:
```golang
if backup.Spec.SnapshotName == "" {
   return nil
}
```
Then we don't need to modify the following code.

PR #3955, opened 2025-08-04
fix(controller): avoid variable shadowing in waitForV2EngineRebuild

derekbit:

Issue Longhorn/longhorn#11420

Fix variable shadowing issue where the parameter 'e' was being reused within the loop, potentially causing confusion and masking the original engine parameter.


shuo-wu (approved):

shuo-wu:

PR #3961, opened 2025-08-04
fix: Ensure Orphan CR Retention in Longhorn When orphan-auto-deletion is Set to False 

nzhan126:

Issue #7795

Delete orphan CR only when node is deleted, evicting, or the actual data of the orphan CR no longer exist


controller/orphan_controller.go
```diff
@@ -596,7 +596,7 @@ func (oc *OrphanController) updateDataCleanableCondition(orphan *longhorn.Orphan
 		orphan.Status.Conditions = types.SetCondition(orphan.Status.Conditions, longhorn.OrphanConditionTypeDataCleanable, status, reason, "")
 	}()
 
-	isUnavailable, err := oc.ds.IsNodeDownOrDeletedOrMissingManager(orphan.Spec.NodeID)
+	isUnavailable, err := oc.ds.IsNodeDownOrMissingManager(orphan.Spec.NodeID)
```
shuo-wu: When the node is deleted, this function will error out. Then the below `reason = longhorn.OrphanConditionTypeDataCleanableReasonNodeDeleted` is unreachable.

Actually, the change in line 613 should be enough.

shuo-wu (commented):
The suggested solution includes 2 parts:
```
1. We shouldn't clean up orphan CR when its node is down, missing manager, or Longhorn manager cannot [get disk stat](https://github.com/longhorn/longhorn-manager/blob/c4e7942684cc1f8ece900854d09126a7b1f8c0b6/util/util.go#L434-L435)
2. We still should delete orphan CR when node is deleted, evicting, or the actual data of the orphan CR no longer exist
```
How does this PR achieve `the actual data of the orphan CR no longer exist
 `?

nzhan126:
> The suggested solution includes 2 parts:
> 
> ```
> 1. We shouldn't clean up orphan CR when its node is down, missing manager, or Longhorn manager cannot [get disk stat](https://github.com/longhorn/longhorn-manager/blob/c4e7942684cc1f8ece900854d09126a7b1f8c0b6/util/util.go#L434-L435)
> 2. We still should delete orphan CR when node is deleted, evicting, or the actual data of the orphan CR no longer exist
> ```
> 
> How does this PR achieve `the actual data of the orphan CR no longer exist `?
@shuo-wu 
I suppose it is done here:
https://github.com/longhorn/longhorn-manager/blob/dd2415bf01f6a7788c9bbee02a74d4e3ce412152/controller/node_controller.go#L1356-L1359

controller/orphan_controller.go
```diff
@@ -596,20 +596,20 @@ func (oc *OrphanController) updateDataCleanableCondition(orphan *longhorn.Orphan
...
 		}
+		reason = longhorn.OrphanConditionTypeDataCleanableReasonNodeDeleted
+		return nil
+	}
+
+	isUnavailable, err := oc.ds.IsNodeDownOrMissingManager(orphan.Spec.NodeID)
+	if err != nil {
+		return errors.Wrapf(err, "failed to check node down or missing manager for node %v", orphan.Spec.NodeID)
+	}
+	if isUnavailable {
```
shuo-wu: Using `IsNodeDownOrDeletedOrMissingManager` should be enough since the node delete case is already excluded

PhanLe1010:
Hi @derekbit @COLDTURNIP Could I have a question about orphan delete setting:

The current code is
```go
		Description: "This setting allows Longhorn to automatically delete orphan resources and their corresponding orphaned resources. \n\n" +
			"Orphan resources located on nodes that are in down or unknown state will not be cleaned up automatically. \n\n" +
			"Available items are: \n\n" +
			"- **replica-data**: replica data store \n\n" +
			"- **instance**: engine and replica runtime instance \n\n",
```

From the design perceptive. What would be the expectation?
1. When node down/missing manager pod, we do not delete orphan CRs at all
2. OR When node down/missing manager pod, we do delete orphan CRs but does not actually delete data

PhanLe1010:
This PR is related to https://github.com/longhorn/longhorn/issues/7795

COLDTURNIP:
> Hi @derekbit @COLDTURNIP Could I have a question about orphan delete setting:
> 
> The current code is
> 
> ```go
> 		Description: "This setting allows Longhorn to automatically delete orphan resources and their corresponding orphaned resources. \n\n" +
> 			"Orphan resources located on nodes that are in down or unknown state will not be cleaned up automatically. \n\n" +
> 			"List the enabled resource types in a semicolon-separated list. \n\n" +
> 			"Available items are: \n\n" +
> 			"- **replica-data**: replica data store \n\n" +
> 			"- **instance**: engine and replica runtime instance \n\n",
> ```
> 
> From the design perceptive. What would be the expectation?
> 
>     1. When node down/missing manager pod, we do not delete orphan CRs at all
>     2. OR When node down/missing manager pod, we do delete orphan CRs but does not actually delete data

Leaving aside the impact of this setting, I think the 2nd one is expected. The CR reflects the status of the resource in the cluster. When a resource is not reachable from the cluster (e.g. node leave), the corresponding CR should be deleted. Once a node is unreachable by Longhorn, the underlying resource, including the landed data, are not controlled by Longhorn, except the eviction is required. Back to the auto deletion, when this setting is enabled, we expects the controller clear the orphaned resource once it is identified. Without the auto deletion enabled, the CRs are expected to reflect the detectability of the resource, and delete the them only when user asks to.

When auto deletion enabled:

- Resource monitor detects that a resource becomes orphaned; creates an orphan CR for it.
- Then clear the orphan CR immediately to clean up the orphaned resource.

No matter the auto deletion is enabled or not, when a node/disk eviction requested:

- Clear the orphaned resources on the node/disk by deleting the orphan CR

No matter the auto deletion is enabled or not, when a node leaves Longhorn without eviction, and there are some resources on the leaving node:

- Since the resource is no longer reachable by Longhorn, delete the orphan CR directly without touching the disk resource
- Runtime instances live in instance manager, and instance manager would be terminated along with the node, hence the orphaned instance will also be cleared. This removal is not caused by the auto deletion setting.

derekbit:
> > Hi @derekbit @COLDTURNIP Could I have a question about orphan delete setting:
> > The current code is
> > ```go
> > 		Description: "This setting allows Longhorn to automatically delete orphan resources and their corresponding orphaned resources. \n\n" +
> > 			"Orphan resources located on nodes that are in down or unknown state will not be cleaned up automatically. \n\n" +
> > 			"List the enabled resource types in a semicolon-separated list. \n\n" +
> > 			"Available items are: \n\n" +
> > 			"- **replica-data**: replica data store \n\n" +
> > 			"- **instance**: engine and replica runtime instance \n\n",
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > From the design perceptive. What would be the expectation?
> > ```
> > 1. When node down/missing manager pod, we do not delete orphan CRs at all
> > 2. OR When node down/missing manager pod, we do delete orphan CRs but does not actually delete data
> > ```
> 
> Leaving aside the impact of this setting, I think the 2nd one is expected. The CR reflects the status of the resource in the cluster. When a resource is not reachable from the cluster (e.g. node leave), the corresponding CR should be deleted. Once a node is unreachable by Longhorn, the underlying resource, including the landed data, are not controlled by Longhorn, except the eviction is required. Back to the auto deletion, when this setting is enabled, we expects the controller clear the orphaned resource once it is identified. Without the auto deletion enabled, the CRs are expected to reflect the detectability of the resource, and delete the them only when user asks to.
> 
> When auto deletion enabled:
> 
> * Resource monitor detects that a resource becomes orphaned; creates an orphan CR for it.
> * Then clear the orphan CR immediately to clean up the orphaned resource.
> 
> No matter the auto deletion is enabled or not, when a node/disk eviction requested:
> 
> * Clear the orphaned resources on the node/disk by deleting the orphan CR
> 
> No matter the auto deletion is enabled or not, when a node leaves Longhorn without eviction, and there are some resources on the leaving node:
> 
> * Since the resource is no longer reachable by Longhorn, delete the orphan CR directly without touching the disk resource
> * Runtime instances live in instance manager, and instance manager would be terminated along with the node, hence the orphaned instance will also be cleared. This removal is not caused by the auto deletion setting.

Second that 2 is expected.

shuo-wu:
> No matter the auto deletion is enabled or not, when a node leaves Longhorn without eviction, and there are some resources on the leaving node:
> 
> - Since the resource is no longer reachable by Longhorn, delete the orphan CR directly without touching the disk resource
> - Runtime instances live in instance manager, and instance manager would be terminated along with the node, hence the orphaned instance will also be cleared. This removal is not caused by the auto deletion setting.

This makes sense.

> No matter the auto deletion is enabled or not, when a node/disk eviction requested:
>
>Clear the orphaned resources on the node/disk by deleting the orphan CR

For the eviction case, I prefer to retain the related data resource and clean up the orphan CR only. One typical eviction case is node upgrade. After that, Longhorn may be able to continue using it.

Besides, we need to be careful about one corner case for the node down/manager missing scenario. If the node is suddenly back to healthy when deleting orphan CRs, Longhorn cannot accidently clean up the related data.

PR #3963, opened 2025-08-07
feat(backingimage): evict backing image manager for missing disks

COLDTURNIP:

Issue longhorn/longhorn#10983

While a broken disk is force deleted from a node without eviction, trigger the eviction on corresponding backing image manager.


innobead (approved):
In general, LGTM. A good feedback fro codefactor bot. Suggest fixing it.

controller/backing_image_manager_controller.go
```diff
@@ -388,6 +391,40 @@ func (c *BackingImageManagerController) cleanupBackingImageManager(bim *longhorn
 	return nil
 }
 
+// evictMissingDiskBackingImageManager trigger image manager eviction for missing disks
+func (c *BackingImageManagerController) evictMissingDiskBackingImageManager(bim *longhorn.BackingImageManager) error {
+	node, _ := c.ds.GetNode(bim.Spec.NodeID)
+	if node == nil {
+		return nil
+	}
```
derekbit: Should we check the returned error to handle the situation that both `node` and `err` are not `nil`?

controller/backing_image_manager_controller.go
```diff
@@ -388,6 +391,40 @@ func (c *BackingImageManagerController) cleanupBackingImageManager(bim *longhorn
...
+	for _, diskStatus := range node.Status.DiskStatus {
+		if diskStatus.DiskUUID == bim.Spec.DiskUUID {
+			return nil
+		}
+	}
```
derekbit: Suggest refactoring this part into a separate function `checkDiskExisting`.

controller/backing_image_manager_controller.go
```diff
@@ -388,6 +391,40 @@ func (c *BackingImageManagerController) cleanupBackingImageManager(bim *longhorn
...
+
+	// Backing image manager is serving on the disk that no longer belongs to any node. Trigger the manager eviction.
+	for imageName := range bim.Spec.BackingImages {
+		bi, getImageErr := c.ds.GetBackingImageRO(imageName)
+		if getImageErr != nil {
+			if datastore.ErrorIsNotFound(getImageErr) {
+				c.logger.Warnf("No corresponding backing image %v for missing disk backing image manager %v", imageName, bim.Name)
+				continue
+			} else {
+				return errors.Wrapf(getImageErr, "Failed to get backing image %v for missing disk backing image manager %v", bi.Name, bim.Name)
```
derekbit: convention
```suggestion
				return errors.Wrapf(getImageErr, "failed to get backing image %v for missing disk backing image manager %v", bi.Name, bim.Name)
```

controller/backing_image_manager_controller.go
```diff
@@ -388,6 +391,40 @@ func (c *BackingImageManagerController) cleanupBackingImageManager(bim *longhorn
...
+			}
+		}
+		if bimDiskFileSpec, exist := bi.Spec.DiskFileSpecMap[bim.Spec.DiskUUID]; exist {
+			c.logger.Infof("Evicting backing image manager %v because of missing disk %v", bim.Name, bim.Spec.DiskUUID)
+			bimDiskFileSpec.EvictionRequested = true
+			if _, updateErr := c.ds.UpdateBackingImage(bi); updateErr != nil {
+				return errors.Wrapf(updateErr, "Failed to evict missing disk backing image manager %v from backing image %v", bim.Name, bi.Name)
```
derekbit:
```suggestion
				return errors.Wrapf(updateErr, "failed to evict missing disk backing image manager %v from backing image %v", bim.Name, bi.Name)
```

controller/backing_image_manager_controller.go
```diff
@@ -388,6 +391,40 @@ func (c *BackingImageManagerController) cleanupBackingImageManager(bim *longhorn
...
+		bi, getImageErr := c.ds.GetBackingImageRO(imageName)
+		if getImageErr != nil {
+			if datastore.ErrorIsNotFound(getImageErr) {
+				c.logger.Warnf("No corresponding backing image %v for missing disk backing image manager %v", imageName, bim.Name)
+				continue
+			} else {
+				return errors.Wrapf(getImageErr, "Failed to get backing image %v for missing disk backing image manager %v", bi.Name, bim.Name)
+			}
+		}
+		if bimDiskFileSpec, exist := bi.Spec.DiskFileSpecMap[bim.Spec.DiskUUID]; exist {
```
derekbit: Is it better to check if `DiskFileSpecMap` is not `nil` before accessing `bi.Spec.DiskFileSpecMap[bim.Spec.DiskUUID]`?

controller/backing_image_manager_controller.go
COLDTURNIP: Not sure if there's a chance to have both `node` and `err` are not `nil`. But will change to return any non-nil error.

COLDTURNIP:
@coderabbitai review

controller/backing_image_manager_controller.go
```diff
@@ -388,6 +391,38 @@ func (c *BackingImageManagerController) cleanupBackingImageManager(bim *longhorn
...
+				c.logger.Warnf("No corresponding backing image %v for missing disk backing image manager %v", imageName, bim.Name)
+				continue
+			}
+			return errors.Wrapf(getImageErr, "failed to get backing image %v for missing disk backing image manager %v", bi.Name, bim.Name)
+		}
+		if bi.Spec.DiskFileSpecMap != nil {
+			if bimDiskFileSpec, exist := bi.Spec.DiskFileSpecMap[bim.Spec.DiskUUID]; exist {
+				c.logger.Infof("Evicting backing image manager %v because of missing disk %v", bim.Name, bim.Spec.DiskUUID)
+				bimDiskFileSpec.EvictionRequested = true
+				if _, updateErr := c.ds.UpdateBackingImage(bi); updateErr != nil {
```
innobead, line 417: Should `bim.Spec.DiskUUID` be removed from `bi.Spec.DiskFileSpecMap` before calling `UpdateBackingImage`? The current PR calls `UpdateBackingImage` without changes to `bi`.

Additionally, should `bimDiskFileSpec.EvictionRequested = true` until all `Spec.DiskFileSpecMap` entries are cleaned up by removing the problematic BIM?

COLDTURNIP: The `bimDiskFileSpec` is the corresponding item in `bi.Spec.DiskFileSpecMap`. Once an eviction is initiated by `bi.Spec.DiskFileSpecMap[diskUUID].EvictionRequested = true`, BI controller will update the corresponding BIM spec, to delete the image from BIM or shutdown the BIM. We initiate the BIM eviction in the BIM controller rather than the BI controller since only BIM contains both the knowledge of node and disk IDs. Maybe it would be more clear to initiate the eviction in BI controller.

innobead (dismissed):

datastore/longhorn.go
```diff
@@ -3167,6 +3167,19 @@ func (s *DataStore) IsNodeSchedulable(name string) bool {
 	return nodeSchedulableCondition.Status == longhorn.ConditionStatusTrue
 }
 
+func (s *DataStore) IsNodeHasDiskUUID(nodeName, diskUUID string) (bool, error) {
+	node, err := s.GetNodeRO(nodeName)
+	if err != nil {
+		return false, err
```
derekbit, line 3173: What will happen if the node is not found?

datastore/longhorn.go
```diff
@@ -3404,6 +3404,19 @@ func (s *DataStore) IsNodeSchedulable(name string) bool {
...
+	}
```
derekbit: What will happen if the node is not found? If the node is not found, does it mean the disks are gone, the it need to return `false, nil`.

derekbit:
@COLDTURNIP After addressing the comments, we can merge the PR.

derekbit (approved):
LGTM

COLDTURNIP:

PR #3971, opened 2025-08-12
fix(crd): remove omitempty from ReplicaRebuildingBandwidthLimit

derekbit:

Issue Longhorn/longhorn#11403


mantissahz (approved):
LGTM
But do we need to patch old volumes during upgrading?

derekbit:
> LGTM But do we need to patch old volumes during upgrading? And is the default value `0`?

No need. Default is 0.

PR #3972, opened 2025-08-12
feat(crd): update api for volume.spec.ReplicaRebuildingBandwidthLimit

derekbit:

Issue Longhorn/longhorn#11403


api/model.go
```diff
@@ -396,6 +401,10 @@ type UpdateSnapshotMaxSize struct {
 	SnapshotMaxSize string `json:"snapshotMaxSize"`
 }
 
+type UpdateReplicaRebuildingBandwidthLimit struct {
```
mantissahz, line 404: Why do we have two structures that have the same content?
`UpdateReplicaRebuildingBandwidthLimitInput`

derekbit: `UpdateReplicaRebuildingBandwidthLimitInput` is for the schema  in `func NewSchema()client. Schemasas`

PR #3975, opened 2025-08-14
feat: support v2 cloning (full-copy and linked-clone mode)

PhanLe1010:
State machine for volume cloning 
cloning_state_machine

https://github.com/longhorn/longhorn/issues/7794


PhanLe1010:
~~There is a small work on storageclass for this one~~

Done

PhanLe1010:
ping @derekbit @c3y1huang

controller/volume_controller.go
```diff
@@ -859,18 +859,34 @@ func (c *VolumeController) ReconcileEngineReplicaState(v *longhorn.Volume, es ma
 			continue
 		}
 
-		if status.State == engineapi.ProcessStateComplete && v.Status.CloneStatus.State != longhorn.VolumeCloneStateCompleted {
-			v.Status.CloneStatus.State = longhorn.VolumeCloneStateCompleted
-			c.eventRecorder.Eventf(v, corev1.EventTypeNormal, constant.EventReasonVolumeCloneCompleted,
-				"finished cloning snapshot %v from source volume %v",
+		if !isCloneTargetCopyInProgress(v) {
+			// No longer need to sync up with the engine
```
c3y1huang: > // No longer need to sync up with the engine

This comment is a bit brief. Could you add some detail to explain why we don’t need to sync with the engine anymore?

controller/volume_controller.go
```diff
@@ -859,18 +859,34 @@ func (c *VolumeController) ReconcileEngineReplicaState(v *longhorn.Volume, es ma
...
+		case engineapi.ProcessStateError:
 			v.Status.CloneStatus.State = longhorn.VolumeCloneStateFailed
-			c.eventRecorder.Eventf(v, corev1.EventTypeWarning, constant.EventReasonVolumeCloneFailed,
+			c.eventRecorder.Eventf(
+				v, corev1.EventTypeWarning, constant.EventReasonVolumeCloneFailed,
 				"failed to clone snapshot %v from source volume %v: %v",
-				v.Status.CloneStatus.Snapshot, v.Status.CloneStatus.SourceVolume, status.Error)
+				v.Status.CloneStatus.Snapshot, v.Status.CloneStatus.SourceVolume, status.Error,
+			)
 		}
```
c3y1huang, line 882: Since we’re looping over `e.Status.CloneStatus` but only updating `v.Status.CloneStatus.State`, if there are multiple entries, won’t the last one just overwrite the earlier updates and cause weird state transitions?

app/recurringjob/util.go
```diff
@@ -96,6 +96,11 @@ func filterVolumesForJob(allowDetached bool, volumes []longhorn.Volume, filterNa
 			continue
 		}
 
+		if volume.Spec.CloneMode == longhorn.CloneModeLinkedClone {
+			logger.Infof("Skipping linked-clone volume %v", volume.Name)
+			continue
+		}
```
c3y1huang: Can we leverage the webhook validator to prevent adding recurring jobs to linked-clone volumes?

controller/node_controller.go
```diff
@@ -1747,8 +1747,16 @@ func (nc *NodeController) syncReplicaEvictionRequested(node *longhorn.Node, kube
 				return err
 			}
 			if replica.Spec.EvictionRequested != shouldEvictReplica {
-				replica.Spec.EvictionRequested = shouldEvictReplica
-				replicasToSync = append(replicasToSync, replicaToSync{replica, reason})
+				isLinkedClone, err := nc.ds.IsVolumeLinkedCloneVolume(replica.Spec.VolumeName)
+				if err != nil {
+					log.Infof("Cannot evict replica %v: %v", replicaName, err.Error())
```
c3y1huang:
```suggestion
					log.WithError(err).Warnf("Cannot evict replica %v", replica.Name)
```

controller/node_controller.go
```diff
@@ -1747,8 +1747,16 @@ func (nc *NodeController) syncReplicaEvictionRequested(node *longhorn.Node, kube
...
+				} else if isLinkedClone {
+					log.Infof("Cannot evict replica %v because the volume %v is a linked-clone volume. "+
+						"Please delete this volume to unblock the eviction", replica.Name, replica.Spec.VolumeName)
```
c3y1huang:
```suggestion
					log.Warnf("Cannot evict replica %v because the volume %v is a linked-clone volume. "+
						"Please delete this volume to unblock the eviction", replica.Name, replica.Spec.VolumeName)
```

c3y1huang (changes requested):

controller/utils.go
```diff
@@ -57,19 +57,30 @@ func isVolumeUpgrading(v *longhorn.Volume) bool {
...
-		v.Status.CloneStatus.State == longhorn.VolumeCloneStateFailed
-	return isCloningDesired && !isCloningCompletedOrFailed
+// The flow is: VolumeCloneStateEmpty => VolumeCloneStateInitiated => VolumeCloneStateCopyCompletedAwaitingHealthy => VolumeCloneStateFailed or VolumeCloneStateCompleted
+// isCloneTargetCopyInProgress returns true if the volume is a clone target and has not reached copy-complete yet
+func isCloneTargetCopyInProgress(v *longhorn.Volume) bool {
+	isCloneTarget := types.IsDataFromVolume(v.Spec.DataSource)
+	reachedOrAfterCopyComplete := v.Status.CloneStatus.State == longhorn.VolumeCloneStateCopyCompletedAwaitingHealthy ||
+		v.Status.CloneStatus.State == longhorn.VolumeCloneStateFailed ||
+		v.Status.CloneStatus.State == longhorn.VolumeCloneStateCompleted
+	return isCloneTarget && !reachedOrAfterCopyComplete
```
c3y1huang, line 68:
```suggestion
	return types.IsDataFromVolume(v.Spec.DataSource) && !reachedOrAfterCopyComplete
```

controller/utils.go
```diff
@@ -57,19 +57,30 @@ func isVolumeUpgrading(v *longhorn.Volume) bool {
...
-func isCloningRequiredAndNotCompleted(v *longhorn.Volume) bool {
-	isCloningDesired := types.IsDataFromVolume(v.Spec.DataSource)
-	isCloningCompleted := v.Status.CloneStatus.State == longhorn.VolumeCloneStateCompleted
-	return isCloningDesired && !isCloningCompleted
+// isCloneTargetActive returns true if the volume is a clone target and has not reached a terminal state
+func isCloneTargetActive(v *longhorn.Volume) bool {
+	isCloneTarget := types.IsDataFromVolume(v.Spec.DataSource)
+	isTerminal := v.Status.CloneStatus.State == longhorn.VolumeCloneStateFailed ||
+		v.Status.CloneStatus.State == longhorn.VolumeCloneStateCompleted
+	return isCloneTarget && !isTerminal
```
c3y1huang, line 76:
```suggestion
	return types.IsDataFromVolume(v.Spec.DataSource) && !isTerminal
```

controller/utils.go
```diff
@@ -57,19 +57,30 @@ func isVolumeUpgrading(v *longhorn.Volume) bool {
...
+}
+
+// isCloneTargetNotCompletedAndNotCopyCompleted returns true if v is a clone target and not in copy-complete or completed state
+func isCloneTargetNotCompletedAndNotCopyCompleted(v *longhorn.Volume) bool {
+	isCloneTarget := types.IsDataFromVolume(v.Spec.DataSource)
+	completedOrCopyCompleted := v.Status.CloneStatus.State == longhorn.VolumeCloneStateCopyCompletedAwaitingHealthy ||
+		v.Status.CloneStatus.State == longhorn.VolumeCloneStateCompleted
+	return isCloneTarget && !completedOrCopyCompleted
```
c3y1huang, line 84:
```suggestion
	return types.IsDataFromVolume(v.Spec.DataSource) && !completedOrCopyCompleted
```

c3y1huang (changes requested):

k8s/pkg/apis/longhorn/v1beta2/volume.go
```diff
@@ -45,6 +45,15 @@ const (
 	VolumeDataSourceTypeVolume   = VolumeDataSourceType("volume")
 )
 
+// +kubebuilder:validation:Enum="";full-copy;linked-clone
+type CloneMode string
+
+const (
+	CloneModeEmpty       = CloneMode("")
```
derekbit: How abount
```suggestion
	CloneModeNone       = CloneMode("")
```

k8s/pkg/apis/longhorn/v1beta2/volume.go
```diff
@@ -92,10 +101,11 @@ const (
 type VolumeCloneState string
 
 const (
-	VolumeCloneStateEmpty     = VolumeCloneState("")
-	VolumeCloneStateInitiated = VolumeCloneState("initiated")
-	VolumeCloneStateCompleted = VolumeCloneState("completed")
-	VolumeCloneStateFailed    = VolumeCloneState("failed")
+	VolumeCloneStateEmpty                        = VolumeCloneState("")
```
derekbit, line 104:
```suggestion
	VolumeCloneStateNone                        = VolumeCloneState("")
```

manager/engine.go
```diff
@@ -91,6 +91,14 @@ func (m *VolumeManager) CreateSnapshot(snapshotName string, labels map[string]st
 		return nil, err
 	}
 
+	vol, err := m.ds.GetVolumeRO(volumeName)
+	if err != nil {
+		return nil, err
+	}
+	if vol.Spec.CloneMode == longhorn.CloneModeLinkedClone {
+		return nil, fmt.Errorf("cannot create snapshot for linked-clone volume %v", volumeName)
+	}
```
derekbit, line 100: Can we move the check to snapshot/validator.go?

webhook/resources/backup/validator.go
```diff
@@ -72,9 +72,10 @@ func (b *backupValidator) Create(request *admission.Request, newObj runtime.Obje
 		return werror.NewInvalidError(fmt.Sprintf("backup target %s is not available", backupTargetName), "")
 	}
 
+	labelVolumeName := backup.Labels[types.LonghornLabelBackupVolume]
+
```
derekbit, line 76: Should we check if the label `types.LonghornLabelBackupVolume` exists?

webhook/resources/volume/validator.go
```diff
@@ -7,6 +7,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 
+	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/runtime"
```
derekbit:
```suggestion
"k8s.io/apimachinery/pkg/api/resource"
"k8s.io/apimachinery/pkg/runtime"

apiequality "k8s.io/apimachinery/pkg/api/equality"
```

webhook/resources/volume/validator.go
```diff
@@ -491,6 +504,11 @@ func validateReplicaCount(dataLocality longhorn.DataLocality, replicaCount int)
 			return werror.NewInvalidError(fmt.Sprintf("number of replica count should be 1 when data locality is %v", longhorn.DataLocalityStrictLocal), "")
 		}
 	}
+	if cloneMode == longhorn.CloneModeLinkedClone {
+		if replicaCount != 1 {
+			return werror.NewInvalidError(fmt.Sprintf("number of replica count must be 1 when clone mode %v", longhorn.CloneModeLinkedClone), "")
+		}
+	}
```
derekbit, line 487: Is it possible to remove the limitation in the future?

controller/volume_controller.go

controller/volume_controller.go
PhanLe1010: No because there would be max 1 replica when doing cloning

app/recurringjob/util.go
PhanLe1010: Good idea. Updated. Thank you

controller/node_controller.go
PhanLe1010: Thank you. Fixed

controller/node_controller.go
PhanLe1010: Thank you. Fixed

controller/utils.go
PhanLe1010: I prefer to keep the variable `isCloneTarget` to make it a bit easier to read the function. WDYT?

controller/utils.go
PhanLe1010: I prefer to keep the variable `isCloneTarget` to make it a bit easier to read the function. WDYT?

controller/utils.go
PhanLe1010: I prefer to keep the variable `isCloneTarget` to make it a bit easier to read the function. WDYT?

k8s/pkg/apis/longhorn/v1beta2/volume.go
PhanLe1010: Thank you. Fixed

manager/engine.go
PhanLe1010: Webhook already has the logic.

This is still required as this API talk to engine directly by passing webhook. This is the legacy API

webhook/resources/backup/validator.go
PhanLe1010: Maybe not. It is already added by the mutator https://github.com/longhorn/longhorn-manager/blob/8a666918a9eba325d836c67cc1bc480da1a28ce5/webhook/resources/backup/mutator.go#L69-L89

webhook/resources/volume/validator.go
PhanLe1010: They are from the same module `k8s.io/apimachinery`. Should they stay in the same block as currently?

webhook/resources/volume/validator.go
PhanLe1010: I think we should not remove it. It is by design that linked-clone volume is a short-live, no replication, for backup solution to quickly read data and then destroy it.

PhanLe1010:
All review comments are addressed. Can we move forward to merging today? I would like to have this in rc2 for QA to test it

cc @derekbit @c3y1huang

webhook/resources/volume/validator.go
derekbit: We have a convention like https://github.com/longhorn/longhorn/wiki/coding-convention#organize-package-imports

PhanLe1010: Ok. Fixed.

Btw, [the convention](https://github.com/longhorn/longhorn/wiki/coding-convention#organize-package-imports) has a typo at 
```go
		// third-party packages without alias names related to k8s
		corev1 "k8s.io/api/core/v1"

		// third-party packages with alias names related to k8s
		"k8s.io/kubernetes/pkg/controller"

```

controller/volume_controller.go
c3y1huang: Can we validate this condition first, or clarify it with a comment?

PhanLe1010: I specified it at replica replenish already https://github.com/PhanLe1010/longhorn-manager/blob/7ff444697ffb0fb2dfcd8950b8eaa52ef92bd4c2/controller/volume_controller.go#L2946-L2952

Do you still prefer to copy the comment here?

PhanLe1010: Ok. I added the comment

c3y1huang: Yes, I think validating the condition or having a comment closer to this loop would make the code a bit more self-explanatory.

derekbit:
@mergify backport v1.10.x

PR #3977, opened 2025-08-14
feat: getBackupTarget() should iterate all available data engines

derekbit:

Issue Longhorn/longhorn#11478


mantissahz (approved):
That means when `v2-data-engine` is enabled and there is no running instance manager for v2 data engine.

LGTM

derekbit:
> That means when `v2-data-engine` is enabled and there is no running instance manager for v2 data engine.
> 
> LGTM

Yes, v2 IM pod might fail to start because of the insufficient hugepage.

derekbit:

PR #3983, opened 2025-08-14
fix(webhook): correct backup block size updating validation logic

COLDTURNIP:

Issue longhorn/longhorn#11487

While validating the backup update, the backup block size should be only validated if the value is changed.

### DIFF
--- a/webhook/resources/backup/validator.go
+++ b/webhook/resources/backup/validator.go
@@ -110,13 +110,15 @@ func (b *backupValidator) Update(request *admission.Request, oldObj runtime.Obje
 	}
 
 	// Allow backup block size mutation only when the existing obj is not set, or correcting the existed invalid value
-	isValidOldBackupBlockSize := b.validateBackupBlockSize(oldBackup, false) == nil
-	if isValidOldBackupBlockSize && oldBackup.Spec.BackupBlockSize != newBackup.Spec.BackupBlockSize {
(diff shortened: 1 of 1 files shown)


COLDTURNIP:
@coderabbitai review

webhook/resources/backup/validator.go
```diff
@@ -110,13 +110,15 @@ func (b *backupValidator) Update(request *admission.Request, oldObj runtime.Obje
...
-	isValidOldBackupBlockSize := b.validateBackupBlockSize(oldBackup, false) == nil
-	if isValidOldBackupBlockSize && oldBackup.Spec.BackupBlockSize != newBackup.Spec.BackupBlockSize {
-		err := fmt.Errorf("changing backup block size for backup %v is not supported", oldBackup.Name)
-		return werror.NewInvalidError(err.Error(), "")
-	}
-	if err := b.validateBackupBlockSize(newBackup, false); err != nil {
-		return werror.NewInvalidError(err.Error(), "")
+	if oldBackup.Spec.BackupBlockSize != newBackup.Spec.BackupBlockSize {
+		if b.validateBackupBlockSize(oldBackup, false) == nil {
+			err := fmt.Errorf("changing backup block size for backup %v is not supported", oldBackup.Name)
```
derekbit: Make it clear.
```suggestion
			err := fmt.Errorf("changing backup block size for backup %v is not supported. Only changing invalid backup block size is supported.", oldBackup.Name)
```

PR #3985, opened 2025-08-16
feat(setting): move instance-manager-pod-liveness-probe-timeout to danger zone settings

derekbit:
When applying the setting, Longhorn will try to restart all instance-manager pods if all volumes are detached and eventually restart the instance manager pod without instances running on the instance manager.

Issue Longhorn/longhorn#10788

When applying the setting, Longhorn will try to restart all instance-manager pods if all volumes are detached and eventually restart the instance manager pod without instances running on the instance manager.

### DIFF
--- a/controller/instance_manager_controller.go
+++ b/controller/instance_manager_controller.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"reflect"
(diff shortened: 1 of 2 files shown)


innobead (approved):

derekbit:

PR #3989, opened 2025-08-17
feat: remove ErrorReplicaScheduleEvictReplicaFailed

derekbit:

Issue Longhorn/longhorn#11460

ErrorReplicaScheduleEvictReplicaFailed is not one failed reason of replica scheduling.


derekbit:
> LGTM, only one question. It looks like the error message is the same to 'ErrorReplicaSchedulePrecheckNewReplicaFailed', why do we need to remove it?

It is not a real schedule failure reason and just a hit message, so I remove it from the reasons.

PR #3994, opened 2025-08-18
fix: precheckCreateReplica() should return error if there is no available disk

derekbit:

Issue Longhorn/longhorn#11512


derekbit:
regression test
- test_scheduling.py: https://10.115.5.5/job/private/job/longhorn-tests-regression/167/
- test_basic.py: https://10.115.5.5/job/private/job/longhorn-tests-regression/168/
- full regression: https://10.115.5.5/job/private/job/longhorn-tests-regression/169/ and https://10.115.5.5/job/private/job/longhorn-tests-regression/170/

PR #4000, opened 2025-08-19
fix: validate hugepage while enabling v2 data engine

derekbit:

Issue Longhorn/longhorn#11519

### DIFF
--- a/datastore/longhorn.go
+++ b/datastore/longhorn.go
@@ -529,14 +529,20 @@ func (s *DataStore) ValidateSetting(name, value string) (err error) {
 			}
 
 			// Ensure if the CPU mask can be satisfied on each node
-			for _, lhnode := range lhNodes {
-				kubeNode, err := s.GetKubernetesNodeRO(lhnode.Name)
+			for _, lhNode := range lhNodes {
+				if isUnavailable, err := s.IsNodeDownOrDeletedOrMissingManager(lhNode.Name); err != nil {
+					return errors.Wrapf(err, "failed to check if node %v is down or deleted", lhNode.Name)
+				} else if isUnavailable {
+					continue
+				}
+
+				kubeNode, err := s.GetKubernetesNodeRO(lhNode.Name)
 				if err != nil {
 					if apierrors.IsNotFound(err) {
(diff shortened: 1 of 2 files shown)


datastore/longhorn.go
```diff
@@ -640,9 +640,6 @@ func (s *DataStore) ValidateV2DataEngineEnabled(dataEngineEnabled bool) (ims []*
 	}
 
 	for _, im := range _ims {
-		if types.IsDataEngineV1(im.Spec.DataEngine) {
-			continue
-		}
```
mantissahz, line 645: Why do we process ims of the data engine v1 here?

derekbit: While enabling v2 data engine, the v2 IM pods don't exist. We can only leverage the existing v1 IM pods.

mantissahz (approved):
In general, LGTM

datastore/longhorn.go
derekbit: Good question. The previous validation in older versions is a bit weird. Iterate the nodes instead, but 
- Skip down or unavailable node
- Skip node that doesn't support v2 data engine
Thank you.

datastore/longhorn.go
```diff
@@ -634,33 +640,36 @@ func (s *DataStore) ValidateV2DataEngineEnabled(dataEngineEnabled bool) (ims []*
...
-	_ims, err := s.ListInstanceManagersRO()
+	lhNodes, err := s.ListNodes()
 	if err != nil {
 		return nil, errors.Wrapf(err, "failed to list instance managers for %v setting update", types.SettingNameV2DataEngine)
 	}
 
-	for _, im := range _ims {
-		if types.IsDataEngineV1(im.Spec.DataEngine) {
+	for _, lhNode := range lhNodes {
+		if isUnavailable, err := s.IsNodeDownOrDeletedOrMissingManager(lhNode.Name); err != nil {
```
innobead, line 649: Is this runtime check necessary? as the current implementation just uses node CR for the following validation.

derekbit: Yes, this is necessary.
The kubeNode and lhNode might be mismatching. If lhNode is gone and manager is gone due to cordoned node, these cases shouldn't hinder enabling v2 data engine.

c3y1huang (approved):

PR #4003, opened 2025-08-19
fix(node): support zero storage reservation from configured annotation

COLDTURNIP:

Issue longhorn/longhorn#11444

To the default disk configuration annotated on a Kubernetes node:

1. If the `StorageReserved` is set to zero explicitly, do not reserve the storage to this default disk.
2. If the `StorageReserved` is not set in the configuration object, reserve the storage by the global percentage setting.


COLDTURNIP:
@coderabbitai review

COLDTURNIP:
After discussion, [we may have other approach to improve the default disk configuration](https://github.com/longhorn/longhorn/issues/11444#issuecomment-3204132837). Drop this PR for now.

PR #4004, opened 2025-08-20
fix: collect settings ReplicaRebuildingBandwidthLimit and DefaultBackupBlockSize

derekbit:

Issue Longhorn/longhorn#5948

Fix the missing collecting settings.


derekbit:
cc @COLDTURNIP

PR #4005, opened 2025-08-21
feat(support-bundle): collect from log-path

c3y1huang:

Issue longhorn/longhorn#11522

Enhances the support bundle to collect logs from the directory specified by the `log-path` setting.

`None`

`None`


derekbit (approved):
The env method is smart.

PR #4007, opened 2025-08-22
fix(backup): use 'snapshotCRCreate' action to create a snapshot for UI

mantissahz:

Issue # longhorn/longhorn#11451

We should create a snapshot CR instead of making a snapshot directly on the engine side.

After we modify the code in longhorn-ui below, the issue is fixed.

https://github.com/longhorn/longhorn-ui/blob/74e6ba0c98872c5048dc4325255479ffa4d34c9a/src/routes/volume/detail/Snapshots.js#L159

@houhoucoop will fix it.


derekbit (commented):
Can we keep `actions["snapshotCreate"] = struct{}{}` as well?

derekbit:
I will merge it after `snapshotCRCreate` works for UI

mantissahz:
>Can we keep actions["snapshotCreate"] = struct{}{} as well and set it deprecated?

Sure. Modified to be deprecated.

PR #4008, opened 2025-08-22
fix(volume): add ReplicaRebuildingBandwidthLimit in volume spec

davidcheng0922:

Issue # https://github.com/longhorn/longhorn/issues/11306#issuecomment-3195441258

Add column 'ReplicaRebuildingBandwidthLimit' in spec

Screenshot from 2025-08-22 17-26-35
Screenshot from 2025-08-22 17-26-11


derekbit:
Thank you @davidcheng0922 !

PR #4013, opened 2025-08-26
feat(controller): cleanup SPDK disks before node deletion

davidcheng0922:

https://github.com/longhorn/longhorn/issues/9127

Ensure the SPDK driver is properly unbound from the disk during Longhorn v2 volume uninstallation.
If not handled, the disk may remain bound to SPDK, making it invisible to the host and unusable for reuse.

To avoid a race condition when removing the finalizer between the uninstall and node controllers, I delegate full control of the node lifecycle to the node controller.
This ensures deletion completes cleanly before the pod is removed.

[Screencast from 2025-08-26 10-02-30.webm](https://github.com/user-attachments/assets/10c92fe2-2170-433e-9330-4def9f4f2d81)

### DIFF
--- a/controller/node_controller.go
+++ b/controller/node_controller.go
@@ -364,6 +364,12 @@ func (nc *NodeController) syncNode(key string) (err error) {
 	log := getLoggerForNode(nc.logger, node)
 
 	if node.DeletionTimestamp != nil {
+		// If V2 engine is enabled, we must clean up SPDK-related drivers
+		// before deleting the node. This ensures that SPDK devices can be reused on future node joins.
+		if err := nc.cleanupDisksBeforeNodeDeletion(node); err != nil {
+			nc.logger.WithError(err).Warn("Failed to clean up disks during node deletion")
+		}
+
 		nc.eventRecorder.Eventf(node, corev1.EventTypeWarning, constant.EventReasonDelete, "Deleting node %v", node.Name)
 		return nc.ds.RemoveFinalizerForNode(node)
 	}
@@ -1605,6 +1611,53 @@ func (nc *NodeController) deleteDisk(diskType longhorn.DiskType, diskName, diskU
 
 	if err := monitor.DeleteDisk(diskType, diskName, diskUUID, diskPath, diskDriver, diskServiceClient); err != nil {
 		return errors.Wrapf(err, "failed to delete disk %v", diskName)
+	}
+
+	return nil
+}
+
+func (nc *NodeController) cleanupDisksBeforeNodeDeletion(node *longhorn.Node) error {
+	nc.logger.Info("Cleaning up disks before node deletion")
+
+	if _, settingExist := types.GetSettingDefinition(types.SettingNameV2DataEngine); !settingExist {
+		nc.logger.Info("Skipping disk cleanup as v2 data engine setting not exist")
+		return nil
+	}
+
+	v2DataEngineEnabled, err := nc.ds.GetSettingAsBool(types.SettingNameV2DataEngine)
+	if err != nil {
+		return errors.Wrapf(err, "failed to get v2 data engine setting")
+	}
+
+	if !v2DataEngineEnabled {
+		nc.logger.Info("Skipping disk cleanup as v2 data engine is not enabled")
+		return nil
+	}
+
+	errs := multierr.NewMultiError()
+	for diskName, diskStatus := range node.Status.DiskStatus {
+		nc.logger.Infof("Cleaning up disk %s", diskName)
+		// Skip non-SPDK disks
+		if diskStatus.Type != longhorn.DiskTypeBlock {
+			continue
+		}
+
+		if diskStatus.DiskDriver == longhorn.DiskDriverNone {
+			continue
+		}
+
+		// Note: Use diskName instead of diskStatus.Name since they are different:
+		//   diskName:        default-disk-block-1
+		//   diskStatus.Name: default-disk-block-1n1
+		if err := nc.deleteDisk(diskStatus.Type, diskName, diskStatus.DiskUUID, diskStatus.DiskPath, string(diskStatus.DiskDriver)); err != nil {
+			nc.logger.WithError(err).Errorf("Failed to delete SPDK disk %v during node cleanup", diskStatus.DiskName)
+			errs.Append("errors", errors.Wrapf(err, "failed to delete SPDK disk %v during node cleanup", diskStatus.DiskName))
+			continue
+		}
+	}
+
+	if len(errs) > 0 {
+		return fmt.Errorf("cleanupDisksBeforeNodeDeletion: %v", errs.ErrorByReason("errors"))
 	}
 
 	return nil
--- a/controller/uninstall_controller.go
+++ b/controller/uninstall_controller.go
@@ -898,6 +898,14 @@ func (c *UninstallController) deleteEngineImages(engineImages map[string]*longho
 }
 
 func (c *UninstallController) deleteNodes(nodes map[string]*longhorn.Node) (err error) {
+	// For node deletion, if there are v2 volumes, we need to unbind the NVMe device first.
+	// To allow more time for the controller to handle NVMe unbinding,
+	// we add a grace period before removing the finalizer.
+
+	// If we directly update the node, delete it in one step, and remove the finalizer
+	// the instance manager controller may not be able to process the node deletion,
+	// since the instance manager would already be deleted.
+
 	defer func() {
 		err = errors.Wrapf(err, "failed to delete nodes")
 	}()
@@ -908,6 +916,7 @@ func (c *UninstallController) deleteNodes(nodes map[string]*longhorn.Node) (err 
 			node.Annotations = make(map[string]string)
 		}
 
+		timeout := metav1.NewTime(time.Now().Add(-gracePeriod))
 		if node.DeletionTimestamp == nil {
 			log.Infof("Adding annotation %v to node %s to mark for deletion", types.GetLonghornLabelKey(types.DeleteNodeFromLonghorn), node.Name)
 			node.Annotations[types.GetLonghornLabelKey(types.DeleteNodeFromLonghorn)] = ""
@@ -924,7 +933,8 @@ func (c *UninstallController) deleteNodes(nodes map[string]*longhorn.Node) (err 
 			} else {
 				log.Info("Marked for deletion")
 			}
-		} else {
+		} else if node.DeletionTimestamp.Before(&timeout) {
+			log.Warn("Node deletion did not finish within timeout, removing finalizer forcibly")
 			if errRemove := c.ds.RemoveFinalizerForNode(node); errRemove != nil {
 				if datastore.ErrorIsNotFound(errRemove) {
 					log.Info("Node is not found")


controller/node_controller.go
```diff
@@ -364,6 +364,13 @@ func (nc *NodeController) syncNode(key string) (err error) {
 	log := getLoggerForNode(nc.logger, node)
 
 	if node.DeletionTimestamp != nil {
+		// If V2 engine is enabled, we must clean up SPDK-related drivers
+		// before deleting the node. This ensures that SPDK devices can be reused on future node joins.
+		if err := nc.cleanupDisksBeforeNodeDeletion(node); err != nil {
+			nc.logger.WithError(err).Errorf("Cleanup spdk disk driver")
+			return err
+		}
```
derekbit, line 371: Best-effort cleanup and don't hinder the node deletion. WDYT?

```suggestion
		if errCleanup := nc.cleanupDisksBeforeNodeDeletion(node); errCleanup != nil {
			nc.logger.WithError(errCleanup).Errorf("Cleanup spdk disk driver")
		}
```

controller/node_controller.go
```diff
@@ -1607,6 +1614,46 @@ func (nc *NodeController) deleteDisk(diskType longhorn.DiskType, diskName, diskU
...
+		nc.logger.Info("Skipping disk cleanup as v2 data engine is not enabled")
+		return nil
+	}
+
+	for diskName, diskInfo := range node.Spec.Disks {
+		nc.logger.Infof("Cleaning up disk %s", diskName)
+		// Skip non-SPDK disks
+		if diskInfo.DiskDriver == "" {
+			continue
+		}
```
```suggestion
		if diskInfo.Type != DiskTypeBlock {
			continue
		}
```

controller/uninstall_controller.go
```diff
@@ -925,16 +925,7 @@ func (c *UninstallController) deleteNodes(nodes map[string]*longhorn.Node) (err
...
-			if errRemove := c.ds.RemoveFinalizerForNode(node); errRemove != nil {
-				if datastore.ErrorIsNotFound(errRemove) {
-					log.Info("Node is not found")
-				} else {
-					err = errors.Wrap(errRemove, "failed to remove finalizer")
-					return
-				}
-			} else {
-				log.Info("Removed finalizer")
-			}
```
derekbit, line 937: If the cleanup of disks is the best effort, do we need to remove the logic?

controller/node_controller.go
davidcheng0922: Sure, even though it fail, we have work around to handle it

controller/uninstall_controller.go
davidcheng0922: It will have race condition,  there is the case uninstall_controller remove the finalizer first, and move forward to delete the instance-manager, but the disk deletion is not done yet.

davidcheng0922: Add grace period timeout for solving this issue.

derekbit: Can we move the teardown logic inside the uninstall job?

davidcheng0922: It may not be necessary. 
Aligning the logic with other deletion operations by adding a grace period is a straightforward approach with no additional risk.

davidcheng0922:
@coderabbitai review

controller/node_controller.go
```diff
@@ -364,6 +364,12 @@ func (nc *NodeController) syncNode(key string) (err error) {
 	log := getLoggerForNode(nc.logger, node)
 
 	if node.DeletionTimestamp != nil {
+		// If V2 engine is enabled, we must clean up SPDK-related drivers
+		// before deleting the node. This ensures that SPDK devices can be reused on future node joins.
+		if err := nc.cleanupDisksBeforeNodeDeletion(node); err != nil {
+			nc.logger.WithError(err).Errorf("Cleanup spdk disk driver")
```
innobead: Suggest a warning log instead, since error logs generally mean a real failure. Besides, suggest message change to `Failed to clean up disks during node deletion`

controller/node_controller.go
```diff
@@ -1607,6 +1613,46 @@ func (nc *NodeController) deleteDisk(diskType longhorn.DiskType, diskName, diskU
...
+		}
+
+		// Retrieve disk status
+		diskStatus, ok := node.Status.DiskStatus[diskName]
+		if !ok {
+			nc.logger.Infof("Disk %s exists in spec but is missing from status. Skipping SPDK cleanup.", diskName)
+			continue
+		}
+
+		diskInstanceName := diskStatus.DiskName
```
innobead: Don't we expect the value should be identical? If yes, why not just use `diskName` or `diskStatus.DiskName` directly w/o this fallback design?

controller/uninstall_controller.go
innobead: I thought uninstall controller should only run in uninstall job already?

innobead: nit: If best efforts are not guaranteed, you can also make a timeout setting explicit or at least log out This way, users will understand that cleanup is a best-effort process. If teardown is not finished on time, users should know how to handle the cleanup.

innobead (changes requested):
A few feedback. Additionally, suggest a CLI command in longhornctl to cleanup leftover resources after uninstallattion. In this case, it's managed v2 longhorn disks.

controller/node_controller.go
```diff
@@ -1607,6 +1613,46 @@ func (nc *NodeController) deleteDisk(diskType longhorn.DiskType, diskName, diskU
...
+		nc.logger.Infof("Cleaning up disk %s", diskName)
+		// Skip non-SPDK disks
+		if diskInfo.Type != longhorn.DiskTypeBlock {
+			continue
+		}
+
+		// Retrieve disk status
+		diskStatus, ok := node.Status.DiskStatus[diskName]
+		if !ok {
+			nc.logger.Infof("Disk %s exists in spec but is missing from status. Skipping SPDK cleanup.", diskName)
```
innobead: Consider using `V2 data engine` instead of `SPDK` directly. 

In this case, can be "Disk %s exists in spec but is missing from status. Skipping longhorn block disk cleanup for v2 data engine"

controller/uninstall_controller.go
davidcheng0922: Yes, uninstall controller will only run in uninstall job. Normal longhorn manager will not activate it.

innobead:
> A few feedback. Additionally, suggest a CLI command in longhornctl to cleanup leftover resources after uninstallattion. In this case, it's managed v2 longhorn disks.

@davidcheng0922 can you help create an issue for this CLI command? leftover resource clean-up for uninstalled clusters. Thanks. 

cc @c3y1huang @derekbit

davidcheng0922:
> leftover resource clean-up for uninstalled clusters.

https://github.com/longhorn/longhorn/issues/11707

controller/uninstall_controller.go
```diff
@@ -924,7 +933,8 @@ func (c *UninstallController) deleteNodes(nodes map[string]*longhorn.Node) (err
 			} else {
 				log.Info("Marked for deletion")
 			}
-		} else {
+		} else if node.DeletionTimestamp.Before(&timeout) {
+			log.Infof("Node deletion did not finish within timeout, removing finalizer forcibly")
```
c3y1huang:
```suggestion
			log.Warn("Node deletion did not finish within timeout, removing finalizer forcibly")
```

controller/node_controller.go
```diff
@@ -1617,6 +1623,41 @@ func (nc *NodeController) deleteDisk(diskType longhorn.DiskType, diskName, diskU
...
+
+		// Retrieve disk status
+		diskStatus, ok := node.Status.DiskStatus[diskName]
+		if !ok {
+			nc.logger.Infof("Disk %s exists in spec but is missing from status. Skipping longhorn block disk cleanup for v2 data engine.", diskName)
+			continue
+		}
+
+		if err := nc.deleteDisk(diskStatus.Type, diskStatus.DiskName, diskStatus.DiskUUID, diskStatus.DiskPath, string(diskStatus.DiskDriver)); err != nil {
+			return errors.Wrapf(err, "failed to delete SPDK disk %v during node cleanup", diskStatus.DiskName)
```
c3y1huang: Should the controller log errors per disk and continue deleting other disks, then aggregate and return errors at the end?

controller/node_controller.go
```diff
@@ -1617,6 +1623,41 @@ func (nc *NodeController) deleteDisk(diskType longhorn.DiskType, diskName, diskU
...
+	if err != nil {
+		return errors.Wrapf(err, "failed to get v2 data engine setting")
+	}
+
+	if !isV2DataEngine {
+		nc.logger.Info("Skipping disk cleanup as v2 data engine is not enabled")
+		return nil
+	}
+
+	for diskName, diskInfo := range node.Spec.Disks {
```
c3y1huang: Why not iterate over `node.Status.DiskStatus` instead?

controller/node_controller.go
davidcheng0922: Thank you! Your suggestion is better — I’ll update it.

controller/node_controller.go
```diff
@@ -1610,6 +1616,48 @@ func (nc *NodeController) deleteDisk(diskType longhorn.DiskType, diskName, diskU
 	return nil
 }
 
+func (nc *NodeController) cleanupDisksBeforeNodeDeletion(node *longhorn.Node) error {
+	nc.logger.Info("Cleaning up disks before node deletion")
+
+	isV2DataEngine, err := nc.ds.GetSettingAsBool(types.SettingNameV2DataEngine)
```
derekbit: nit:
```suggestion
	v2DataEngineEnabled, err := nc.ds.GetSettingAsBool(types.SettingNameV2DataEngine)
```

controller/node_controller.go
```diff
@@ -1610,6 +1616,48 @@ func (nc *NodeController) deleteDisk(diskType longhorn.DiskType, diskName, diskU
...
+	if err != nil {
+		return errors.Wrapf(err, "failed to get v2 data engine setting")
+	}
```
derekbit, line 1630: if the setting is not found, we can return nil and skip the deletion.

controller/node_controller.go
```diff
@@ -1610,6 +1616,48 @@ func (nc *NodeController) deleteDisk(diskType longhorn.DiskType, diskName, diskU
...
+		}
+
+		if diskStatus.DiskDriver == longhorn.DiskDriverNone {
+			continue
+		}
+
+		// Note: Use diskName instead of diskStatus.Name since they are different:
+		//   diskName:        default-disk-block-1
+		//   diskStatus.Name: default-disk-block-1n1
+		if err := nc.deleteDisk(diskStatus.Type, diskName, diskStatus.DiskUUID, diskStatus.DiskPath, string(diskStatus.DiskDriver)); err != nil {
```
derekbit, line 1652: is the deletion a sync operation?

davidcheng0922: Yes, it is sync operation

controller/node_controller.go
```diff
@@ -1610,6 +1616,53 @@ func (nc *NodeController) deleteDisk(diskType longhorn.DiskType, diskName, diskU
...
+			nc.logger.WithError(err).Errorf("Failed to delete SPDK disk %v during node cleanup", diskStatus.DiskName)
+			errs.Append("errors", errors.Wrapf(err, "failed to delete SPDK disk %v during node cleanup", diskStatus.DiskName))
+			continue
+		}
+	}
+
+	if len(errs) > 0 {
+		return fmt.Errorf("cleanupDisksBeforeNodeDeletion: %v", errs.ErrorByReason("errors"))
+	}
```
derekbit, line 1661: If the disks are somehow unable to be deleted, the error will block the deletion of the node resource.
Can we just log the error message and no need to return error? Any potential issue?

davidcheng0922: It will not block the deletion even failed, the caller only log and then remove the finalizer 

```go
if node.DeletionTimestamp != nil {
	// If V2 engine is enabled, we must clean up SPDK-related drivers
	// before deleting the node. This ensures that SPDK devices can be reused on future node joins.
	if err := nc.cleanupDisksBeforeNodeDeletion(node); err != nil {
		nc.logger.WithError(err).Warn("Failed to clean up disks during node deletion")
	}

	nc.eventRecorder.Eventf(node, corev1.EventTypeWarning, constant.EventReasonDelete, "Deleting node %v", node.Name)
	return nc.ds.RemoveFinalizerForNode(node)
}
```

Or you think it is better to handle it here and make it not to return any error?

derekbit: You are right! I missed it. The implementation LGTM. Thank you.

derekbit (approved):
LGTM

PR #4016, opened 2025-08-27
fix: volume stuck in detaching/attaching loop due to migration cannot finish

PhanLe1010:
This PR fix 2 issues:
1. If volume only has 1 active engine left and it does not have spec.nodeID, select it as the current engine. Do not stuck here
2. Do not attempt a new attach if the volume is currently under migration with vol.Status.CurrentMigrationNodeID != ""

longhorn/longhorn#11479

FYI, this is the flow of all engine picking related logic:

Blank diagram


derekbit (approved):
LGTM

shuo-wu (approved):

derekbit:

PR #4019, opened 2025-08-28
fix(webhook): dr volume always set backup block size from source backup

COLDTURNIP:

Issue longhorn/longhorn#11580

Always copy the backup block size form the source backup while creating a new DR volume.


COLDTURNIP:
@coderabbit review

webhook/resources/volume/mutator.go
```diff
@@ -389,13 +389,28 @@ func (v *volumeMutator) mutate(newObj runtime.Object, moreLabels map[string]stri
...
+		// The backup block size of a DR volume is always copied from the latest backup
+		bName, _, _, err := backupstore.DecodeBackupURL(volume.Spec.FromBackup)
+		if err != nil {
+			return nil, werror.NewInvalidError(fmt.Sprintf("failed to get backup name from backup URL %s: %s", volume.Spec.FromBackup, err.Error()), "spec.fromBackup")
+		}
+		backup, err := v.ds.GetBackupRO(bName)
+		if err != nil {
+			return nil, werror.NewInvalidError(fmt.Sprintf("failed to inspect the backup config %s: %s", volume.Spec.FromBackup, err.Error()), "spec.fromBackup")
+		}
+		backupBlockSize = backup.Spec.BackupBlockSize
```
derekbit, line 403: Is it possible `backup.Spec.BackupBlockSize` 0?

COLDTURNIP: Good catch. Should fall back to 2MiB in this case.

webhook/resources/volume/mutator.go
```diff
@@ -389,13 +389,32 @@ func (v *volumeMutator) mutate(newObj runtime.Object, moreLabels map[string]stri
...
+			backupBlockSize = types.BackupBlockSize2Mi
+		}
+	} else if backupBlockSize == 0 {
+		// For non-DR volumes, mutate the backup block size to the default one if not set
 		defaultBackupBlockSize, settingErr := v.getDefaultBackupBlockSize()
 		if settingErr != nil {
 			return nil, werror.NewInvalidError(fmt.Sprintf("invalid default backup block size setting: %s", settingErr.Error()), "")
 		}
-		patchOps = append(patchOps, fmt.Sprintf(`{"op": "replace", "path": "/spec/backupBlockSize", "value": "%d"}`, defaultBackupBlockSize))
+		backupBlockSize = defaultBackupBlockSize
```
davidcheng0922, line 414: What's the difference between `defaultBackupBlockSize` and `types.BackupBlockSize2Mi` ?

COLDTURNIP: `defaultBackupBlockSize` is the configurable global setting for creating new normal volumes. While creating a new volume without specifying the backup block size, it will be filled as `defaultBackupBlockSize`. This global setting `default-backup-block-size` is newly introduced in v1.10.0, and set to `types.BackupBlockSize2Mi` by default. The `types.BackupBlockSize2Mi` is the legacy default block size used in all backups. In other word, when a backup is fetch from some remote backup store, and the backup block size is not set or zero, it implies that the block size is `types.BackupBlockSize2Mi`.

PR #4024, opened 2025-08-31
feat: update volume validation logics for v2 data engine

derekbit:

Issue Longhorn/longhorn#11600

### DIFF
--- a/webhook/resources/volume/validator.go
+++ b/webhook/resources/volume/validator.go
@@ -312,38 +312,6 @@ func (v *volumeValidator) Update(request *admission.Request, oldObj runtime.Obje
 				newVolume.Name, newVolume.Spec.DataEngine)
 			return werror.NewInvalidError(err.Error(), "")
 		}
-
-		if oldVolume.Spec.ReplicaAutoBalance != newVolume.Spec.ReplicaAutoBalance {
-			err := fmt.Errorf("changing replica auto balance for volume %v is not supported for data engine %v",
-				newVolume.Name, newVolume.Spec.DataEngine)
-			return werror.NewInvalidError(err.Error(), "")
-		}
-
-		if oldVolume.Spec.RestoreVolumeRecurringJob != newVolume.Spec.RestoreVolumeRecurringJob {
-			err := fmt.Errorf("changing restore volume recurring job for volume %v is not supported for data engine %v",
-				newVolume.Name, newVolume.Spec.DataEngine)
-			return werror.NewInvalidError(err.Error(), "")
-		}
-
-		if oldVolume.Spec.ReplicaSoftAntiAffinity != newVolume.Spec.ReplicaSoftAntiAffinity {
-			err := fmt.Errorf("changing replica soft anti-affinity for volume %v is not supported for data engine %v",
-				newVolume.Name, newVolume.Spec.DataEngine)
-			return werror.NewInvalidError(err.Error(), "")
-		}
-
-		if oldVolume.Spec.ReplicaZoneSoftAntiAffinity != newVolume.Spec.ReplicaZoneSoftAntiAffinity {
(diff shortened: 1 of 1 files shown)


webhook/resources/volume/validator.go
```diff
@@ -312,38 +312,38 @@ func (v *volumeValidator) Update(request *admission.Request, oldObj runtime.Obje
...
 			return werror.NewInvalidError(err.Error(), "")
 		}
+	}
 
-		if oldVolume.Spec.ReplicaAutoBalance != newVolume.Spec.ReplicaAutoBalance {
-			err := fmt.Errorf("changing replica auto balance for volume %v is not supported for data engine %v",
-				newVolume.Name, newVolume.Spec.DataEngine)
-			return werror.NewInvalidError(err.Error(), "")
-		}
+	if oldVolume.Spec.ReplicaAutoBalance != newVolume.Spec.ReplicaAutoBalance {
```
c3y1huang: I'm confused. Does this mean Longhorn will block spec changes for all data engine types?

c3y1huang (approved):

innobead (commented):
Should we also remove the following from v2 check? since backing image or encryption of a volume should not be changed after creation? 

```go
		if oldVolume.Spec.BackingImage != newVolume.Spec.BackingImage {
			err := fmt.Errorf("changing backing image for volume %v is not supported for data engine %v",
				newVolume.Name, newVolume.Spec.DataEngine)
		}

		if oldVolume.Spec.Encrypted != newVolume.Spec.Encrypted {
			err := fmt.Errorf("changing encryption for volume %v is not supported for data engine %v",
				newVolume.Name, newVolume.Spec.DataEngine)
		}
```

innobead:
For the removed validation, does it mean that v2 now supports scheduling work as expected? including auto balance * affinity? If yes, this should be part of the v2 feature party to v1.

derekbit:
> For the removed validation, does it mean that v2 now supports scheduling work as expected? including auto balance * affinity? If yes, this should be part of the v2 feature party to v1.

Yes, they are supported.

innobead:
> Should we also move the following to v1 as well instead of v2 check only? since backing image or encryption of a volume should not be changed after creation?
> 
> ```go
> 		if oldVolume.Spec.BackingImage != newVolume.Spec.BackingImage {
> 			err := fmt.Errorf("changing backing image for volume %v is not supported for data engine %v",
> 				newVolume.Name, newVolume.Spec.DataEngine)
> 			return werror.NewInvalidError(err.Error(), "")
> 		}
> 
> 		if oldVolume.Spec.Encrypted != newVolume.Spec.Encrypted {
> 			err := fmt.Errorf("changing encryption for volume %v is not supported for data engine %v",
> 				newVolume.Name, newVolume.Spec.DataEngine)
> 			return werror.NewInvalidError(err.Error(), "")
> 		}
> ```

@derekbit as discussed today, we may need to make the validation for both v1 and v2 engines.

derekbit:
> > Should we also move the following to v1 as well instead of v2 check only? since backing image or encryption of a volume should not be changed after creation?
> > ```go
> > 		if oldVolume.Spec.BackingImage != newVolume.Spec.BackingImage {
> > 			err := fmt.Errorf("changing backing image for volume %v is not supported for data engine %v",
> > 				newVolume.Name, newVolume.Spec.DataEngine)
> > 			return werror.NewInvalidError(err.Error(), "")
> > 		}
> > 
> > 		if oldVolume.Spec.Encrypted != newVolume.Spec.Encrypted {
> > 			err := fmt.Errorf("changing encryption for volume %v is not supported for data engine %v",
> > 				newVolume.Name, newVolume.Spec.DataEngine)
> > 			return werror.NewInvalidError(err.Error(), "")
> > 		}
> > ```
> 
> @derekbit as discussed today, we may need to make the validation for both v1 and v2 engines.

Improved in https://github.com/longhorn/longhorn/issues/11615

PR #4032, opened 2025-09-01
fix(orphan): detect running unscheduled instances

COLDTURNIP:
longhorn-11605

Issue longhorn/longhorn#11605

When an instance is not scheduled to any node, e.g. an engine of a detached volume, the `instance.spec.NodeID == ""` but `instance.status.OwnerID != ""`. Orphan controller should not consider the instance is in ownership transaction.

### DIFF
--- a/controller/instance_manager_controller.go
+++ b/controller/instance_manager_controller.go
@@ -1895,6 +1895,7 @@ func (imc *InstanceManagerController) canDeleteOrphan(orphan *longhorn.Orphan, i
 	return canDelete
 }
 
+// isEngineOnInstanceManager returns true only when it is very certain that an engine is scheduled in a given instance manager
 func (imc *InstanceManagerController) isEngineOnInstanceManager(instanceManager string, instance string) (bool, error) {
 	existEngine, err := imc.ds.GetEngineRO(instance)
 	if err != nil {
@@ -1904,9 +1905,20 @@ func (imc *InstanceManagerController) isEngineOnInstanceManager(instanceManager 
 		}
 		return false, errors.Wrapf(err, "failed to check if engine instance %q is scheduled on instance manager %q", instance, instanceManager)
 	}
-	return imc.isInstanceOnInstanceManager(instanceManager, &existEngine.ObjectMeta, &existEngine.Spec.InstanceSpec, &existEngine.Status.InstanceStatus), nil
+	if isEngineInTransitionState(existEngine) {
+		imc.logger.WithFields(logrus.Fields{
+			"isDeleted":    !existEngine.DeletionTimestamp.IsZero(),
+			"currentState": existEngine.Status.CurrentState,
+			"desiredState": existEngine.Spec.DesireState,
+			"currentNode":  existEngine.Status.OwnerID,
+			"desiredNode":  existEngine.Spec.NodeID,
+		}).Debugf("Skipping check if engine %q is scheduled on instance manager %q; instance is in state transition", existEngine.Name, instanceManager)
+		return false, nil
+	}
+	return imc.isInstanceOnInstanceManager(instanceManager, &existEngine.Status.InstanceStatus), nil
 }
 
+// isReplicaOnInstanceManager returns true only when it is very certain that a replica is scheduled in a given instance manager
 func (imc *InstanceManagerController) isReplicaOnInstanceManager(instanceManager string, instance string) (bool, error) {
 	existReplica, err := imc.ds.GetReplicaRO(instance)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


COLDTURNIP:
@coderabbitai review

controller/instance_manager_controller.go
```diff
@@ -2432,7 +2432,7 @@ func (m *InstanceManagerMonitor) isInstanceOrphanedInInstanceManager(meta *metav
 		return false
 	}
 
-	if status.CurrentState != spec.DesireState || status.OwnerID != spec.NodeID {
+	if status.CurrentState != spec.DesireState || (spec.NodeID != "" && status.OwnerID != spec.NodeID) {
```
WebberHuang1118: How about wrapping it as a utility function for better readability?

innobead (dismissed):

controller/instance_manager_controller.go
derekbit: Can you add a comment to explain the condidition?

controller/instance_manager_controller.go
```diff
@@ -2442,6 +2442,11 @@ func (m *InstanceManagerMonitor) isInstanceOrphanedInInstanceManager(meta *metav
 		return false
 	}
 
+	if spec.NodeID == "" {
+		// instance is in desired state and is not scheduled to any node, but instance process presents in the instance manager
+		return true
+	}
```
derekbit, line 2452: This is not true for a replica of v2 volume, because v2 replica always exists regardless of detachment or attachment.

Currently, v2 data engine doesn't support orphan instance cleanup.  Can you add TODO here?

WebberHuang1118:
Hi @COLDTURNIP

I have a question: If a volume is attached after being immediately detached, meaning the sequence is as follows:

1. Volume attach
2. Immediate volume detach
3. Volume attach again

If I understand correctly, the orphan CR is created after step 2. Will this CR be automatically removed after step 3?

COLDTURNIP:
> Hi @COLDTURNIP
> 
> I have a question: If a volume is attached after being immediately detached, meaning the sequence is as follows:
> 
>     1. Volume attach
> 
>     2. Immediate volume detach
> 
>     3. Volume attach again
> 
> 
> If I understand correctly, the orphan CR is created after step 2. Will this CR be automatically removed after step 3?

When an instance [is scheduled back](https://github.com/longhorn/longhorn-manager/blob/master/controller/instance_manager_controller.go#L1845) to an instance manager, the corresponding orphan CR deletion will be initiated [without deleting the instance process](https://github.com/longhorn/longhorn-manager/blob/6a4d4a9586869a57981bd4ab739a53e6039e0f48/controller/orphan_controller.go#L414-L417).

controller/instance_manager_controller.go
```diff
@@ -2534,3 +2540,21 @@ func (m *InstanceManagerMonitor) categorizeProcesses(instanceMap instanceProcess
...
+
+	isInDesiredState := status.CurrentState == spec.DesireState
+
+	isOnDesiredNode := false
+	switch spec.DataEngine {
+	case longhorn.DataEngineTypeV1:
+		isOnDesiredNode = spec.NodeID == "" || spec.NodeID == status.OwnerID
+	case longhorn.DataEngineTypeV2:
+		isOnDesiredNode = spec.NodeID == status.OwnerID
```
derekbit: Explicitly add a comment to explain why v1 and v2 differ and skip the check of `|| spec.NodeID == status.OwnerID`.

COLDTURNIP: Why don't we need to check the node ownership?

derekbit: I mean adding a comment to explain why you skip the check `|| spec.NodeID == status.OwnerID` for v2 data engine.

COLDTURNIP: Got it, thanks.

COLDTURNIP:
Full E2E build 4038 is triggered.

COLDTURNIP:
@coderabbitai review

controller/instance_manager_controller.go
```diff
@@ -2534,3 +2546,55 @@ func (m *InstanceManagerMonitor) categorizeProcesses(instanceMap instanceProcess
...
+		return true
+	}
+
+	if spec.NodeID != status.OwnerID {
+		// not on desired node
+		return true
+	}
+
+	return false
+}
```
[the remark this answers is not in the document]
COLDTURNIP: Replicas is always assigned to a node once it get scheduled. We ensure the replica CR is reconciled before checking the node ownership at line 2584. After replica CR is reconciled by its owner controller, the `status.OwnerID` is not empty, and the `spec.NodeID` is expected to equal to `status.OwnerID` regardless the volume is attached.

controller/instance_manager_controller.go
```diff
@@ -2534,3 +2546,55 @@ func (m *InstanceManagerMonitor) categorizeProcesses(instanceMap instanceProcess
...
+		// status has not been reconciled yet
+		return true
+	}
+
+	if status.CurrentState != spec.DesireState {
+		// not in desired state
+		return true
+	}
+
+	isAttached := spec.NodeID != ""
```
shuo-wu: NIT: `requireAttach := spec.NodeID != ""`

controller/instance_manager_controller.go
```diff
@@ -2534,3 +2546,55 @@ func (m *InstanceManagerMonitor) categorizeProcesses(instanceMap instanceProcess
...
+	isScheduledToOwner := spec.NodeID == status.OwnerID
+	if isAttached && !isScheduledToOwner {
+		// attached but not on desired node
```
shuo-wu: It's better to add a comment explaining in which scenario the engine will fall into this status

c3y1huang (approved):

derekbit:
@mergify backport v1.10.x

derekbit:
@mergify backport v1.9.x

PR #4033, opened 2025-09-01
fix(env-check): remove outdated unsupported os conditions

COLDTURNIP:

Issue longhorn/longhorn#11612

### DIFF
--- a/upgrade/v19xto1100/upgrade.go
+++ b/upgrade/v19xto1100/upgrade.go
@@ -10,6 +10,7 @@ import (
 	"github.com/longhorn/longhorn-manager/datastore"
 	"github.com/longhorn/longhorn-manager/types"
 
+	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
 	lhclientset "github.com/longhorn/longhorn-manager/k8s/pkg/client/clientset/versioned"
 	upgradeutil "github.com/longhorn/longhorn-manager/upgrade/util"
 )
@@ -35,6 +36,9 @@ func updateCRs(namespace string, lhClient *lhclientset.Clientset, kubeClient *cl
 		return errors.New("resourceMaps cannot be nil")
 	}
 
+	if err := upgradeNodes(namespace, lhClient, resourceMaps); err != nil {
+		return err
+	}
 	if err := upgradeSettings(namespace, lhClient, resourceMaps); err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


COLDTURNIP:
@coderabbitai review

controller/monitor/environment_check_monitor.go
```diff
@@ -226,6 +226,8 @@ func (m *EnvironmentCheckMonitor) syncPackagesInstalled(kubeNode *corev1.Node, n
 		// Skip environment check condition when
 		// 1.The OS is unknown to Longhorn
 		// 2.The OS that already has all packages installed such as Harvester
+		// And the outdated condition attached by older Longhorn version must be removed.
+		collectedData.conditions = types.RemoveCondition(collectedData.conditions, longhorn.NodeConditionTypeRequiredPackages)
```
innobead: Why not just address this during the upgrade as a one-time effort?

upgrade/v19xto1100/upgrade.go
```diff
@@ -48,6 +56,57 @@ func updateCRs(namespace string, lhClient *lhclientset.Clientset, kubeClient *cl
...
+		// 1.The OS is unknown to Longhorn
+		// 2.The OS that already has all packages installed such as Harvester
+		osImage := strings.ToLower(kubeNode.Status.NodeInfo.OSImage)
+		for _, kw := range knownOSKeywords {
+			if strings.Contains(osImage, kw) {
+				continue NodeLoop
+			}
+		}
+		n.Status.Conditions = types.RemoveCondition(n.Status.Conditions, longhorn.NodeConditionTypeRequiredPackages)
+	}
```
derekbit: Not a good method. Can we avoid using `NodeLoop`?

derekbit (commented):
Can we blindly clean up the condition in the upgrade path because the OS package condition will be set after node controller detects the errors?

upgrade/v19xto1100/upgrade.go
```diff
@@ -48,6 +52,28 @@ func updateCRs(namespace string, lhClient *lhclientset.Clientset, kubeClient *cl
...
+
+	nodeMap, err := upgradeutil.ListAndUpdateNodesInProvidedCache(namespace, lhClient, resourceMaps)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil
+		}
+		return errors.Wrapf(err, "failed to list all existing Longhorn nodes during the node upgrade")
+	}
+
+	var errs []error
```
derekbit: `errs` is not required

derekbit (approved):
LGTM

derekbit:

PR #4041, opened 2025-09-04
feat(setting): introduce System Info category

derekbit:

Issue Longhorn/longhorn#11656

Some settings such as default engine image are system info and read-only. We can move them to System Info category rather than put them in General category.


derekbit:
@mergify backport v1.10.x

PR #4046, opened 2025-09-05
fix: make sure the BIM has a ready backing image replica before creating a backing image backup

WebberHuang1118:
Make sure the BIM has a ready backing image replica before creating a backing image backup.

Issue longhorn/longhorn#11675

When Longhorn creates a BackupBackingImage (BBI), it may [select a BackingImageManager (BIM) that doesn’t have a ready BackingImage (BI) replica](https://github.com/longhorn/longhorn-manager/blob/979b8d559b4fe2cb087e39a32d08ce1e329a0cef/controller/backup_backing_image_controller.go#L366-L382). As a result, the BIM pod might try to read a file that isn’t ready during BBI creation.

Reproduction steps:
* Set up a cluster with Longhorn (I used Harvester).
* Run this [[script](https://github.com/WebberHuang1118/misc-tools/blob/main/backingimage/create_bi_bbi_cycle.sh)](https://github.com/WebberHuang1118/misc-tools/blob/main/backingimage/create_bi_bbi_cycle.sh) to repeatedly create BIs and BBIs.
* The script is able to pass more than 10 rounds.

### DIFF
--- a/controller/backup_backing_image_controller.go
+++ b/controller/backup_backing_image_controller.go
@@ -362,23 +362,9 @@ func (bc *BackupBackingImageController) checkMonitor(bbi *longhorn.BackupBacking
 	}
 
 	// pick one backing image manager to do backup
-	var targetBim *longhorn.BackingImageManager
-	for diskUUID := range backingImage.Spec.DiskFileSpecMap {
-		bimMap, err := bc.ds.ListBackingImageManagersByDiskUUID(diskUUID)
-		if err != nil {
-			return nil, err
-		}
-		for _, bim := range bimMap {
-			if bim.DeletionTimestamp == nil {
-				if uuidInManager, exists := bim.Status.BackingImageFileMap[backingImage.Name]; exists && uuidInManager.UUID == backingImage.Status.UUID {
-					targetBim = bim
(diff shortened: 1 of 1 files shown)


derekbit:
@mergify backport v1.8.x v1.9.x v1.10.x

WebberHuang1118:
> LGTM.
> 
> BTW, the enclosed method name is a bit confusing. Suggest renaming it to `startNewBackupWithMonitor(` instead.
> 
> ```
> func (bc *BackupBackingImageController) checkMonitor(bbi *longhorn.B
> ```

This part is not covered by this PR. @derekbit, will the Longhorn team create another ticket for @innobead’s suggestion?

derekbit:
> > LGTM.
> > BTW, the enclosed method name is a bit confusing. Suggest renaming it to `startNewBackupWithMonitor(` instead.
> > ```
> > func (bc *BackupBackingImageController) checkMonitor(bbi *longhorn.B
> > ```
> 
> This part is not covered by this PR. @derekbit, will the Longhorn team create another ticket for @innobead’s suggestion?

Sure. We can handle it in another refactor ticket.

PR #4065, opened 2025-09-09
fix(webhook): retry webhook configuration apply

derekbit:

Issue longhorn/longhorn#11654

Retry webhook configuration apply if failed.


derekbit:
@coderabbitai review

webhook/server/server.go
```diff
@@ -163,7 +165,28 @@ func (s *WebhookServer) runAdmissionWebhookListenAndServe(handler http.Handler,
 			},
 		}
 
-		return secret, apply.WithOwner(secret).ApplyObjects(validatingWebhookConfiguration, mutatingWebhookConfiguration)
+		// Retry apply
```
innobead, line 168: The change can improve resilience, though the current implementation failing soon may also benefit users. However, I am curious if this resolution addresses the original issue, which still seems unclear from the conversation in the thread. 

In the current implementation, if the workloads fail to apply, there should be an error returned, and then longhorn-manager should fail to start. I assume that, eventually, the longhorn-manager pods should be started along with the webhook servers. 

[/webhook/webhook.go#L38-L40](https://github.com/longhorn/longhorn-manager/blob/a315536a02db2c7c2688ba77f4429b8564a4f3a2/webhook/webhook.go#L38-L40)
```go
		if err := s.ListenAndServe(); err != nil {
			logrus.Fatalf("Error %v webhook server failed: %v", webhookType, err)
		}
```

derekbit: Yes, I have the same thought, so the issue is addressed and mitigated in the second commit.

c3y1huang (approved):

derekbit:
@mergify backport v1.9.x v1.10.x

webhook/webhook.go
```diff
@@ -90,7 +119,11 @@ func isServiceAvailable(endpoint string, timeout time.Duration) bool {
 		time.Sleep(2 * time.Second)
 	}
```
davidcheng0922, line 120: Question, will we standardize the timeout retry pattern with `retry.Do`?

derekbit: Do you mean other controllers and components?
If yes, it is good to standardize them.

derekbit: cc @longhorn/dev

PR #4066, opened 2025-09-09
fix(scheduler): exclude nodes where the data engine is disabled

derekbit:

Issue Longhorn/longhorn#11695


innobead:
@mergify backport v1.10.x

PR #4074, opened 2025-09-11
fix: remove  for backupMode CRD

nzhan126:
longhorn/longhorn#9718

Issue #

### DIFF
--- a/webhook/resources/backup/mutator.go
+++ b/webhook/resources/backup/mutator.go
@@ -104,10 +104,9 @@ func (b *backupMutator) Create(request *admission.Request, newObj runtime.Object
 	}
 	patchOps = append(patchOps, fmt.Sprintf(`{"op": "replace", "path": "/spec/labels", "value": %s}`, string(valueBackupLabels)))
 
-	if backup.Spec.BackupMode == longhorn.BackupModeIncrementalNone {
+	if backup.Spec.BackupMode == "" {
 		patchOps = append(patchOps, fmt.Sprintf(`{"op": "replace", "path": "/spec/backupMode", "value": "%s"}`, string(longhorn.BackupModeIncremental)))
 	}
-
 	if backup.Spec.SnapshotName != "" {
 		volume, err := b.ds.GetVolumeRO(volumeName)
 		if err != nil {
--- a/k8s/crds.yaml
+++ b/k8s/crds.yaml
@@ -700,7 +700,6 @@ spec:
                 enum:
                 - full
                 - incremental
-                - ""
                 type: string
               labels:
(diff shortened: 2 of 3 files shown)


webhook/resources/backup/mutator.go
```diff
@@ -104,10 +104,6 @@ func (b *backupMutator) Create(request *admission.Request, newObj runtime.Object
 	}
 	patchOps = append(patchOps, fmt.Sprintf(`{"op": "replace", "path": "/spec/labels", "value": %s}`, string(valueBackupLabels)))
 
-	if backup.Spec.BackupMode == longhorn.BackupModeIncrementalNone {
-		patchOps = append(patchOps, fmt.Sprintf(`{"op": "replace", "path": "/spec/backupMode", "value": "%s"}`, string(longhorn.BackupModeIncremental)))
-	}
```
shuo-wu, line 109: We cannot remove this. This guarantees that the default value will be applied automatically for this field.

Would you try to retain `BackupModeIncrementalNone` while modifying `// +kubebuilder:validation:Enum=full;incremental` only?

webhook/resources/backup/mutator.go
```diff
@@ -104,10 +104,9 @@ func (b *backupMutator) Create(request *admission.Request, newObj runtime.Object
 	}
 	patchOps = append(patchOps, fmt.Sprintf(`{"op": "replace", "path": "/spec/labels", "value": %s}`, string(valueBackupLabels)))
 
-	if backup.Spec.BackupMode == longhorn.BackupModeIncrementalNone {
+	if backup.Spec.BackupMode == "" {
```
derekbit, line 107: If a Backup CR is created from an older backup that was made before Longhorn supported full backups, what will be the value of `backup.Spec.BackupMode`?

shuo-wu: If the CR is from an old backup, the backup metafile probably does not contain field `BackupMode` hence we always apply default value `BackupModeIncremental`. 
This modification does not change any logic. Any concerns?

derekbit: Because the PR limits the value of `Spec.BackupMode` to either `incremental" or "full",  I'm thinking if we can create a backup with `backup.Spec.BackupMode=""`. (I know the value will be mutated by the mutation webhook later).

nzhan126: in this case then we can only update the documentation to  _The backup mode of this backup.
    Can be "full" or "incremental" or ""_ ?

derekbit:
Thanks @nzhan126 for the contribution!

PR #4085, opened 2025-09-11
fix(validator): add volume size check for ublk frontend

davidcheng0922:

[[BUG] Missing size validation for ublk frontend expansion #11742
](https://github.com/longhorn/longhorn/issues/11742)

add volume size check for ublk frontend


derekbit:
@mergify backport v1.10.x

PR #4099, opened 2025-09-15
feat: gracefully shut down spdk_tgt and retry retrieving disk status if it is not ready

derekbit:

Issue Longhorn/longhorn#11760


controller/monitor/disk_utils.go
```diff
@@ -106,6 +115,10 @@ func getBlockTypeDiskConfig(client *DiskServiceClient, diskName, diskPath string
 		return nil, err
 	}
 
+	if info.State != string(DiskStateReady) {
+		return nil, fmt.Errorf("disk %v is not ready: %v", diskName, info.State)
+	}
```
shuo-wu: Can we have a more graceful handling? This stat get function will be used by other functions like `CreateDisksFromAnnotation`. Then this failure will lead to all its upper callers' failures, including `syncKubernetesNode`.

derekbit: Addressed. Use `diskconfig.State` instead. Thank you.

controller/monitor/disk_utils.go
```diff
@@ -26,6 +26,15 @@ const (
 	uuidGenerationRetries = 20
 )
 
+type DiskState string
+
+const (
+	DirkStateUnknown  = DiskState("")
```
c3y1huang:
```suggestion
	DiskStateUnknown  = DiskState("")
```

controller/monitor/disk_utils.go
```diff
@@ -26,6 +26,15 @@ const (
...
+	DiskStateError    = DiskState("error")
+	DiskStateReady    = DiskState("ready")
+	DiskStateCreating = DiskState("creating")
+)
```
c3y1huang: Are the other states used?

derekbit: Use the consts in longhorn-spdk-engine instead.

controller/monitor/disk_utils.go
derekbit: Use the consts in longhorn-spdk-engine instead.

controller/monitor/disk_utils.go
```diff
@@ -26,6 +28,8 @@ const (
 	uuidGenerationRetries = 20
 )
 
+type DiskState string
```
c3y1huang: Still needed?

derekbit: Removed. Thank you.

c3y1huang (approved):

derekbit:
@mergify backport v1.10.x

PR #4100, opened 2025-09-15
fix(engine_controller): v2 volume also require checkSizeBeforeRestoration

davidcheng0922:

https://github.com/longhorn/longhorn/issues/11767

V2 volume require checkSizeBeforeRestoration() for updating the size before restoration

### DIFF
--- a/controller/engine_controller.go
+++ b/controller/engine_controller.go
@@ -1372,7 +1372,8 @@ func preRestoreCheckAndSync(log logrus.FieldLogger, engine *longhorn.Engine,
 		return false, fmt.Errorf("backup volume is empty for backup restoration of engine %v", engine.Name)
 	}
 
-	if cliAPIVersion >= engineapi.CLIAPIMinVersionForExistingEngineBeforeUpgrade {
+	if (types.IsDataEngineV1(engine.Spec.DataEngine) && cliAPIVersion >= engineapi.CLIAPIMinVersionForExistingEngineBeforeUpgrade) ||
+		types.IsDataEngineV2(engine.Spec.DataEngine) {
 		return checkSizeBeforeRestoration(log, engine, ds)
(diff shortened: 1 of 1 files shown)


controller/engine_controller.go
```diff
@@ -1372,7 +1372,8 @@ func preRestoreCheckAndSync(log logrus.FieldLogger, engine *longhorn.Engine,
 		return false, fmt.Errorf("backup volume is empty for backup restoration of engine %v", engine.Name)
 	}
 
-	if cliAPIVersion >= engineapi.CLIAPIMinVersionForExistingEngineBeforeUpgrade {
+	if (types.IsDataEngineV1(engine.Spec.DataEngine) && cliAPIVersion >= engineapi.CLIAPIMinVersionForExistingEngineBeforeUpgrade) ||
```
derekbit, line 1375: What's the value of `cliAPIVersion` for v2 data engine?

davidcheng0922: Always 0, so it skip to expand the volume before restoration

```go
func (s *DataStore) GetDataEngineImageCLIAPIVersion(imageName string, dataEngine longhorn.DataEngineType) (int, error) {
	if imageName == "" {
		return -1, fmt.Errorf("cannot check the CLI API Version based on empty image name")
	}

	if types.IsDataEngineV2(dataEngine) {
		return 0, nil
	}

	ei, err := s.GetEngineImageRO(types.GetEngineImageChecksumName(imageName))
	if err != nil {
		return -1, errors.Wrapf(err, "failed to get engine image object based on image name %v", imageName)
	}

	return ei.Status.CLIAPIVersion, nil
}
```

c3y1huang (approved):

derekbit:
@mergify backport v1.10.x

PR #4101, opened 2025-09-15
feat: deploy engine image only if v1 engine is enabled

bachmanity1:

Issue https://github.com/longhorn/longhorn/issues/11769

I've tested this in my local cluster and it seems to be working well.

### DIFF
--- a/controller/setting_controller.go
+++ b/controller/setting_controller.go
@@ -354,7 +354,15 @@ func (sc *SettingController) syncDangerZoneSettingsForManagedComponents(settingN
 
 	if slices.Contains(dangerSettingsRequiringSpecificDataEngineVolumesDetached, settingName) {
 		switch settingName {
-		case types.SettingNameV1DataEngine, types.SettingNameV2DataEngine:
+		case types.SettingNameV1DataEngine:
+			if err := sc.updateDataEngine(settingName); err != nil {
+				return errors.Wrapf(err, "failed to apply %v setting to Longhorn instance managers when there are attached volumes. "+
+					"It will be eventually applied", settingName)
+			}
+			if err := sc.syncEngineImage(); err != nil {
+				return errors.Wrapf(err, "failed to sync engine image for v1 data engine")
+			}
+		case types.SettingNameV2DataEngine:
 			if err := sc.updateDataEngine(settingName); err != nil {
 				return errors.Wrapf(err, "failed to apply %v setting to Longhorn instance managers when there are attached volumes. "+
 					"It will be eventually applied", settingName)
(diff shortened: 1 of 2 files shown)


controller/setting_controller.go
```diff
@@ -926,6 +934,59 @@ func (sc *SettingController) updateDataEngine(setting types.SettingName) error {
...
+				},
+				Spec: longhorn.EngineImageSpec{
+					Image: defaultEngineImage,
+				},
+			}
+			if _, err = sc.ds.CreateEngineImage(ei); err != nil {
+				return errors.Wrapf(err, "cannot create engine image for %v", engineImageName)
+			}
+		}
+	} else {
```
hookak: The conditions for deleting the engine-image ds are quite complicated. It's not just about disabling the v1 engine - we also need to check v1 engine volumes and related services. WDYT @bachmanity1?

bachmanity1: I think that if there are active v1 volumes, the webhook won’t allow disabling the `v1-data-engine` setting, so the engine image won’t be deleted. https://github.com/longhorn/longhorn-manager/blob/45c02175a7db225168b394b04bd54a4031427e47/datastore/longhorn.go#L611

hookak: Thanks — I appreciate it.

controller/setting_controller.go
```diff
@@ -926,6 +934,59 @@ func (sc *SettingController) updateDataEngine(setting types.SettingName) error {
...
+		}
+		if ei.Annotations == nil {
+			ei.Annotations = make(map[string]string)
+		}
+		ei.Annotations[types.GetLonghornLabelKey(types.DeleteEngineImageFromLonghorn)] = ""
+		if _, err := sc.ds.UpdateEngineImage(ei); err != nil {
+			return errors.Wrap(err, "failed to update engine image annotations to mark for deletion")
+		}
+		if err := sc.ds.DeleteEngineImage(engineImageName); err != nil {
+			return err
```
derekbit: Can we return `nil` if the engine image is not found?

controller/setting_controller.go
```diff
@@ -926,6 +934,59 @@ func (sc *SettingController) updateDataEngine(setting types.SettingName) error {
...
+		return errors.Wrapf(err, "failed to get %v setting for syncing engine image", types.SettingNameV1DataEngine)
+	}
+
+	defaultEngineImage, err := sc.ds.GetSettingValueExisted(types.SettingNameDefaultEngineImage)
+	if err != nil {
+		return err
+	}
+	engineImageName := types.GetEngineImageChecksumName(defaultEngineImage)
+	if v1EngineEnabled {
+		if _, err := sc.ds.GetEngineImage(engineImageName); err != nil {
```
derekbit:
```suggestion
		if _, err := sc.ds.GetEngineImageRO(engineImageName); err != nil {
```

controller/setting_controller.go
bachmanity1: Done.

controller/setting_controller.go
bachmanity1: Done.

controller/setting_controller.go
```diff
@@ -926,6 +934,62 @@ func (sc *SettingController) updateDataEngine(setting types.SettingName) error {
...
+			return errors.Wrapf(err, "cannot get engine image %v", engineImageName)
+		}
+		if ei.Annotations == nil {
+			ei.Annotations = make(map[string]string)
+		}
+		ei.Annotations[types.GetLonghornLabelKey(types.DeleteEngineImageFromLonghorn)] = ""
+		if _, err := sc.ds.UpdateEngineImage(ei); err != nil {
+			return errors.Wrap(err, "failed to update engine image annotations to mark for deletion")
+		}
+		if err := sc.ds.DeleteEngineImage(engineImageName); err != nil {
```
shuo-wu: I am thinking whether we can retain all engine image CRs and delete the pods only.

If we decide to delete CRs, should we remove the user-created/non-default engine image CRs besides the default CRs?

bachmanity1: > I am thinking whether we can retain all engine image CRs and delete the pods only.

Can I ask why we would want to retain engine image CRs when there are no pods? I think it would be cleaner to delete the engine image CRs and let the controller handle the pods. Otherwise, the controller logic could become unnecessarily complicated.

> If we decide to delete CRs, should we remove the user-created/non-default engine image CRs besides the default CRs?

I think we should only care about the default CR that is auto-generated when the v1 engine is enabled, and let users handle any other CRs they create manually.

derekbit: cc @shuo-wu

shuo-wu: Sorry for the late reply. I am thinking if it's too aggressive to delete all user-created engine image CRs once v1 data engine is disabled. Will there be any delay for the cleanup?

Anyway, the implementation in general LGTM

derekbit: Same questions. We have a mechanism to delete expired engine images, so we don't need to delete them aggressively and immediately.

bachmanity1: > I am thinking if it's too aggressive to delete all user-created engine image CRs once v1 data engine is disabled. Will there be any delay for the cleanup?

I’m not sure I understand what you mean. As I mentioned, this PR only handles the deletion of the default engine image CR; other engine image CRs are not handled here.

derekbit: Can we delay the deletion and let engine image controller handle the deletion?
If the v1 data engine is disabled, we can skip the check https://github.com/longhorn/longhorn-manager/blob/master/controller/engine_image_controller.go#L675-L678?

bachmanity1: I've moved default engine image deletion logic to engine image controller. Settings controller makes sure that default engine image is deployed when v1 engine is enabled.

derekbit (approved):
LGTM

derekbit:
Thanks @bachmanity1's contribution and the reviews from @shuo-wu and @hookak.

PR #4112, opened 2025-09-16
fix: get the correct owner for v2 volume

mantissahz:

Issue # longhorn/longhorn#11774

Some nodes will label a disabled v2 data engine, and then these nodes should not be the owner for the v2 volumes.

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -4539,6 +4539,12 @@ func (c *VolumeController) isResponsibleFor(v *longhorn.Volume, defaultEngineIma
 		err = errors.Wrap(err, "error while checking isResponsibleFor")
 	}()
 
+	if types.IsDataEngineV2(v.Spec.DataEngine) {
+		if isV2DisabledForNode, err := c.ds.IsV2DataEngineDisabledForNode(c.controllerID); err != nil || isV2DisabledForNode {
+			return false, err
+		}
+	}
+
 	// If a regular RWX is delinquent, try to switch ownership quickly to the owner node of the share manager CR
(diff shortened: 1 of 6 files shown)


innobead (approved):
ref: https://github.com/longhorn/longhorn/issues/7015

controller/volume_controller.go
```diff
@@ -4539,6 +4539,10 @@ func (c *VolumeController) isResponsibleFor(v *longhorn.Volume, defaultEngineIma
 		err = errors.Wrap(err, "error while checking isResponsibleFor")
 	}()
 
+	if compatible, err := c.ds.IsVolumeCompatibleWithNodeEngine(v.Name, c.controllerID); err != nil || !compatible {
+		return false, err
+	}
```
derekbit: `IsVolumeCompatibleWithNodeEngine()` will get the Volume CR. Can we avoid the extra retrieval?

derekbit:
@mergify backport v1.10.x

controller/volume_controller.go
mantissahz: Yes, we can do it.
Use the function `IsV2DataEngineDisabledForNode` instead of `IsVolumeCompatibleWithNodeEngine`
Thanks.

mantissahz:
>IsVolumeCompatibleWithNodeEngine is confusing. Suggest renaming the function to isNodeSupportingDataEngine()

Modified. Thanks.

derekbit:
@mantissahz CI failed.

mantissahz:
>@mantissahz CI failed.

Fixed. Thanks.

PR #4132, opened 2025-09-19
fix: add a Kubernetes node validator

mantissahz:

Issue # longhorn/longhorn#11799

Check if v2 engines or replicas instances are running on the node when labeling the node `node.longhorn.io/disable-v2-data-engine=true`


webhook/server/validation.go
```diff
@@ -17,6 +17,7 @@ import (
 	"github.com/longhorn/longhorn-manager/webhook/resources/engine"
 	"github.com/longhorn/longhorn-manager/webhook/resources/engineimage"
 	"github.com/longhorn/longhorn-manager/webhook/resources/instancemanager"
+	"github.com/longhorn/longhorn-manager/webhook/resources/kubernetesnode"
```
derekbit, line 20: Good naming!

derekbit (approved):
LGTM

derekbit (approved):

PR #4150, opened 2025-09-25
fix(setting): sync V2DataEngineGuaranteedInstanceManagerCPU to GuaranteedInstanceManagerCPU

derekbit:

Issue Longhorn/longhorn#11864

### DIFF
--- a/datastore/longhorn.go
+++ b/datastore/longhorn.go
@@ -240,6 +240,11 @@ func (s *DataStore) syncV2DataEngineGuaranteedInstanceManagerCPU() error {
 		return nil
 	}
 
+	if v2DataEngineGuaranteedInstanceManagerCPU > int(numCPUs*1000) {
+		logrus.Warnf("Skipping the migration for guaranteed-instance-manager-cpu for v2 data engine since old setting %v value %v milliCPUs is larger than the minimum number of CPUs %v among all nodes", types.SettingNameV2DataEngineGuaranteedInstanceManagerCPU, oldV2Setting.Value, numCPUs)
+		return nil
+	}
+
 	guaranteedInstanceManagerCPUInPercentage := int64(v2DataEngineGuaranteedInstanceManagerCPU*100) / (numCPUs * 1000)
 
 	// Get setting guaranteed-instance-manager-cpu


derekbit:
@mergify backport v1.10.x

derekbit:
@coderabbitai review

datastore/longhorn.go
```diff
@@ -207,6 +207,61 @@ func (s *DataStore) syncConsolidatedV2DataEngineSetting(oldSettingName, newSetti
...
+		logrus.WithError(err).Warn("Failed to get the minimum number of CPUs among all nodes, skipping the migration for guaranteed-instance-manager-cpu for v2 data engine")
+		return nil
+	}
+
+	guaranteedInstanceManagerCPUInPercentage := int64(v2DataEngineGuaranteedInstanceManagerCPU*100) / (numCPUs * 1000)
+
+	// Get setting guaranteed-instance-manager-cpu
+	setting, err := s.GetSettingExact(types.SettingNameGuaranteedInstanceManagerCPU)
+	if err != nil {
+		return errors.Wrapf(err, "failed to get old setting %v", types.SettingNameGuaranteedInstanceManagerCPU)
```
c3y1huang:
```suggestion
		return errors.Wrapf(err, "failed to get setting %v", types.SettingNameGuaranteedInstanceManagerCPU)
```

derekbit:
> 1. What if an old setting value is an invalid one? It will fail `syncConsolidatedV2DataEngineSettings` -> `UpdateCustomizedSettings` infinitely. Then the longhorn manager pods cannot be started.

If the old setting is not found, the function `syncConsolidatedV2DataEngineSetting` won't error out to prevent the infinite restarts.
```
func (s *DataStore) syncConsolidatedV2DataEngineSetting(oldSettingName, newSettingName types.SettingName) error {
	oldSetting, err := s.getSettingRO(string(oldSettingName))
	if err != nil {
		if ErrorIsNotFound(err) {
			logrus.Debugf("No old setting %v to be replaced.", oldSettingName)
			return nil
		}
		return errors.Wrapf(err, "failed to get old setting %v", oldSettingName)
	}

	return s.createOrUpdateSetting(newSettingName, oldSetting.Value, "")
}
```


> Should we set a kind of special value or flag for the old settings after the sync? Later on, if users change a new setting to a different value and somehow restart the longhorn manager pod, will the logic change the new setting back to the old setting value?

Good catch! The current order is `syncConsolidatedV2DataEngineSettings`, `syncSettingCRsWithCustomizedDefaultSettings` and then `deleteReplacedSettings`.

I think we can change it to `syncConsolidatedV2DataEngineSettings`, `deleteReplacedSettings` and `syncSettingCRsWithCustomizedDefaultSettings` to prevent the the new setting is changed back to the old setting value. WDYT?

shuo-wu:
> If the old setting is not found, the function syncConsolidatedV2DataEngineSetting won't error out to prevent the infinite restarts.

What if somehow the min cpu number is smaller than the old setting value. I am thinking whether we can ignore the old setting if there is an error.

The order change looks good.

derekbit:
> What if somehow the min cpu number is smaller than the old setting value. I am thinking whether we can ignore the old setting if there is an error.

 ignore the old setting => sounds good. I will make the migration best-effort.

datastore/longhorn.go
derekbit: Fixed. Thank you.

derekbit:
@coderabbitai review

derekbit:
@coderabbitai review

derekbit:
Running testing
https://10.115.5.5/job/private/job/longhorn-tests-regression/274

mantissahz (approved):

PR #4152, opened 2025-09-26
fix: clean up extra new replicas and avoid creating more duplicates during migration

shuo-wu:

Issue https://github.com/longhorn/longhorn/issues/11397

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -4146,22 +4146,84 @@ func (c *VolumeController) createAndStartMatchingReplicas(v *longhorn.Volume,
 	return nil
 }
 
-func (c *VolumeController) deleteInvalidMigrationReplicas(rs, pathToOldRs, pathToNewRs map[string]*longhorn.Replica, v *longhorn.Volume) error {
-	for path, r := range pathToNewRs {
-		matchTheOldReplica := pathToOldRs[path] != nil && r.Spec.DesireState == pathToOldRs[path].Spec.DesireState
-		newReplicaIsAvailable := r.DeletionTimestamp == nil && r.Spec.DesireState == longhorn.InstanceStateRunning &&
-			r.Status.CurrentState != longhorn.InstanceStateError && r.Status.CurrentState != longhorn.InstanceStateUnknown && r.Status.CurrentState != longhorn.InstanceStateStopping
-		if matchTheOldReplica && newReplicaIsAvailable {
-			continue
-		}
-		delete(pathToNewRs, path)
-		if types.IsDataEngineV1(v.Spec.DataEngine) {
-			if err := c.deleteReplica(r, rs); err != nil {
-				return errors.Wrapf(err, "failed to delete the new replica %v when there is no matching old replica in path %v", r.Name, path)
+// deleteInvalidMigrationReplicas selects one replica for each path and delete the rests.
+//
+//	Notice that the selected replica may not be valid.
+//	Returning invalid replicas for some paths helps inform the caller that it should wait for the existing invalid replica disappears before creating a new one.
+func (c *VolumeController) deleteInvalidMigrationReplicas(rs, pathToOldRs map[string]*longhorn.Replica, pathToNewRLists map[string][]*longhorn.Replica, v *longhorn.Volume) (map[string]*longhorn.Replica, error) {
+	migrationReplicas := make(map[string]*longhorn.Replica)
+	for path, rList := range pathToNewRLists {
+		sort.Slice(rList, func(i, j int) bool {
+			return rList[i].Name < rList[j].Name
+		})
+		var chosenReplica *longhorn.Replica
+		isChosenReplicaValid, isChosenReplicaRunning := false, false
+		for _, r := range rList {
+			isReplicaValid := isReplicaValidForMigration(pathToOldRs[path], r)
+			if chosenReplica == nil {
+				chosenReplica = r
+				isChosenReplicaValid = isReplicaValid
+				isChosenReplicaRunning = r.Status.CurrentState == longhorn.InstanceStateRunning
 			}
-		} else {
-			r.Spec.MigrationEngineName = ""
+
+			if isReplicaValid {
+				if !isChosenReplicaValid { // prefer a valid new replica
+					chosenReplica = r
+					isChosenReplicaValid = true
+					isChosenReplicaRunning = r.Status.CurrentState == longhorn.InstanceStateRunning
+				} else {
+					if !isChosenReplicaRunning && r.Status.CurrentState == longhorn.InstanceStateRunning { // when both are valid, prefer a running replica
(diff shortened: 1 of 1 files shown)


controller/volume_controller.go
```diff
@@ -4146,23 +4146,85 @@ func (c *VolumeController) createAndStartMatchingReplicas(v *longhorn.Volume,
...
+		newR.Status.CurrentState != longhorn.InstanceStateError && newR.Status.CurrentState != longhorn.InstanceStateUnknown && newR.Status.CurrentState != longhorn.InstanceStateStopping &&
+		newR.Spec.LastFailedAt == ""
+}
+
+func (c *VolumeController) migrationReplicaCleanup(r *longhorn.Replica, rs map[string]*longhorn.Replica, dataEngineType longhorn.DataEngineType) error {
+	if types.IsDataEngineV1(dataEngineType) {
+		if err := c.deleteReplica(r, rs); err != nil {
+			return errors.Wrapf(err, "failed to delete migration replica %v", r.Name)
+		}
+	} else {
```
derekbit, line 4227: Should we delete the v2 replica as well?

controller/volume_controller.go
```diff
@@ -4146,23 +4146,85 @@ func (c *VolumeController) createAndStartMatchingReplicas(v *longhorn.Volume,
...
+		for _, r := range rList {
+			if r.Name != chosenReplica.Name {
+				c.logger.Infof("Cleaning up the extra migration replica %v (data engine %v) since another migration replica %v is chosen for path %v", r.Name, v.Spec.DataEngine, chosenReplica.Name, path)
+				if err := c.migrationReplicaCleanup(r, rs, v.Spec.DataEngine); err != nil {
+					return nil, err
+				}
 			}
-		} else {
-			r.Spec.MigrationEngineName = ""
+		}
```
davidcheng0922, line 4200: It looks like invalid replicas are being cleaned up twice — once when detected as invalid and again during the final cleanup loop. Should we avoid the duplicate cleanup?

controller/volume_controller.go
```diff
@@ -4146,23 +4146,85 @@ func (c *VolumeController) createAndStartMatchingReplicas(v *longhorn.Volume,
...
+					chosenReplica = r
+					isChosenReplicaValid = true
+					isChosenReplicaRunning = r.Status.CurrentState == longhorn.InstanceStateRunning
+				} else {
+					if !isChosenReplicaRunning && r.Status.CurrentState == longhorn.InstanceStateRunning { // when both are valid, prefer a running replica
+						chosenReplica = r
+						isChosenReplicaValid = true
+						isChosenReplicaRunning = true
+					}
+				}
```
davidcheng0922, line 4181: I'm not sure if we should combine these two conditions
Both of them are
```
chosenReplica = r
isChosenReplicaValid = true
isChosenReplicaRunning = r.Status.CurrentState == longhorn.InstanceStateRunning
```

to 
```go
	if !isChosenReplicaValid || (!isChosenReplicaRunning && r.Status.CurrentState == longhorn.InstanceStateRunning) {
		chosenReplica = r
		isChosenReplicaValid = true
		isChosenReplicaRunning = r.Status.CurrentState == longhorn.InstanceStateRunning
	}
```

WDYT?

shuo-wu: A current running replica may be invalid. In your implementation, we may replace a valid but stopped replica with an invalid running replica. Besides, it is not as readable as the current implementation.

controller/volume_controller.go
shuo-wu: Yes. We can avoid duplicate cleanup!

controller/volume_controller.go
shuo-wu: For v2 live migration, IIRC, Longhorn always uses the same set of replicas. Hence, it's better not to issue any deletion for v2 here.

derekbit: Got it. Thank you.

controller/volume_controller.go
davidcheng0922: yeah, you're right

controller/volume_controller.go
```diff
@@ -4146,23 +4146,85 @@ func (c *VolumeController) createAndStartMatchingReplicas(v *longhorn.Volume,
...
+				if err := c.migrationReplicaCleanup(r, rs, v.Spec.DataEngine); err != nil {
+					return nil, err
+				}
 			}
-		} else {
-			r.Spec.MigrationEngineName = ""
+		}
+
+		if chosenReplica != nil {
+			migrationReplicas[path] = chosenReplica
```
c3y1huang, line 4203: Would this be a valid case where an invalid replica ends up being added to `migrationReplicas`? Would this cause any issues, or is this expected?
1. `rList` has only one replica, and `isReplicaValid == false`.
1. The invalid replica is selected as `chosenReplica`.
1. It is immediately deleted because it’s invalid.
1. However, since `chosenReplica != nil`, it still gets added to `migrationReplicas`.

controller/volume_controller.go
```diff
@@ -4146,23 +4146,85 @@ func (c *VolumeController) createAndStartMatchingReplicas(v *longhorn.Volume,
...
+					if !isChosenReplicaRunning && r.Status.CurrentState == longhorn.InstanceStateRunning { // when both are valid, prefer a running replica
+						chosenReplica = r
+						isChosenReplicaValid = true
+						isChosenReplicaRunning = true
+					}
+				}
+			} else { // for invalid replicas, do cleanup immediately
+				c.logger.Infof("Cleaning up the invalid migration replica %v (data engine %v) since it is state %s, or it is last failed at %s, or there is no matching old replica in path %v", r.Name, v.Spec.DataEngine, r.Status.CurrentState, r.Spec.LastFailedAt, path)
+				if err := c.migrationReplicaCleanup(r, rs, v.Spec.DataEngine); err != nil {
+					return nil, err
```
c3y1huang, line 4185: Should errors accumulate here and continue (with best-effort cleanup) for other invalid replicas?

controller/volume_controller.go
shuo-wu: Yes, it's possible. As I mentioned in the function comment, we may return an invalid replica for a path, and this invalid replica has probably `deletionTimestamp` set. The purpose of returning this kind of special replica is to inform the caller that it should wait for the existing invalid replica to disappear before creating a new one.

controller/volume_controller.go
shuo-wu: The purposes of this PR are:
1. Remove all invalid or extra replicas
2. Do not create a new migration replica or continue the migration if the invalid or extra replicas are not completely removed.

Hence, I prefer not to continue when there is an error here.

derekbit:
@mergify backport v1.10.x v1.9.x v1.8.x

PR #4161, opened 2025-09-30
feat(controller): support endpointslices to replace endpoints

davidcheng0922:
https://github.com/longhorn/longhorn/issues/10796

In Kubernetes v1.33, the `endpoints` resource is deprecated.
This PR updates Longhorn to replace endpoints with `endpointslices`.

For each ShareManager, there will now be two EndpointSlices:

1. Kubernetes-managed EndpointSlice

-  Automatically generated by the `kubernetes controller`
- Has OwnerReference pointing to the `Service`

2. Longhorn-managed EndpointSlice

- Explicitly created by `longhorn-manager`
- Has OwnerReference pointing to the `ShareManager`

We cannot directly modify the Kubernetes-managed EndpointSlice, since changes will be reverted during reconciliation.
Instead, longhorn-manager creates its own EndpointSlice to support use cases such as Multus.

As a result, a ShareManager Service may have two EndpointSlices. Consumers (e.g., kube-proxy, CNI) will aggregate them into a unified set of endpoints.

https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/


derekbit:
@davidcheng0922 Could you resolve the conflict? Thank you.

davidcheng0922:
Pending

PR #4162, opened 2025-09-30
Add support for a new Pod Deletion Policy

FireDrunk:
Issue (https://github.com/longhorn/longhorn/issues/11876)

This PR adds an extra PodDeletionPolicy called: `delete-both-statefulset-and-deployment-pod-and-annotated-pods` (for lack of a better name...)
It adds an additional option to set an Annotation on a Pod to mark it for explicit deletion.

This helps in cases where the Pod is not managed by a StatefulSet, Deployment or DaemonSet.
This is useful in situations where an Operator or Controller manages the Pod manually.

This requires Administrators to explicitly configure those annotations on the pod before longhorn engages in the deletion of the pod(s).

It's quite hard to test, we will have to setup a 3 node cluster, with a pod that has a pvc that is stuck in Terminating, and see if the longhorn-manager now correctly deletes the pod after it's grace period has expired.

Docs PR here: https://github.com/longhorn/website/pull/1210

### DIFF
--- a/controller/kubernetes_pod_controller.go
+++ b/controller/kubernetes_pod_controller.go
@@ -35,6 +35,8 @@ const (
 	controllerAgentName = "longhorn-kubernetes-pod-controller"
 
 	remountRequestDelayDuration = 5 * time.Second
+
+	podDeletionAnnotation = "longhorn.io/allow-deletion"
 )
 
 type KubernetesPodController struct {
@@ -472,7 +474,7 @@ func (kc *KubernetesPodController) getPodWithConflictedAttachment(pods []*corev1
 //
 // Force delete a pod when all of the below conditions are meet:
 // 1. NodeDownPodDeletionPolicy is different than DoNothing
-// 2. pod belongs to a StatefulSet/Deployment depend on NodeDownPodDeletionPolicy
+// 2. pod belongs to a StatefulSet/Deployment depend on NodeDownPodDeletionPolicy OR is explicitly marked for deletion.
 // 3. node containing the pod is down
 // 4. the pod is terminating and the DeletionTimestamp has passed.
 // 5. pod has a PV with provisioner driver.longhorn.io
@@ -484,7 +486,8 @@ func (kc *KubernetesPodController) handlePodDeletionIfNodeDown(pod *corev1.Pod, 
 
 	shouldDelete := (deletionPolicy == types.NodeDownPodDeletionPolicyDeleteStatefulSetPod && isOwnedByStatefulSet(pod)) ||
 		(deletionPolicy == types.NodeDownPodDeletionPolicyDeleteDeploymentPod && isOwnedByDeployment(pod)) ||
-		(deletionPolicy == types.NodeDownPodDeletionPolicyDeleteBothStatefulsetAndDeploymentPod && (isOwnedByStatefulSet(pod) || isOwnedByDeployment(pod)))
+		(deletionPolicy == types.NodeDownPodDeletionPolicyDeleteBothStatefulsetAndDeploymentPod && (isOwnedByStatefulSet(pod) || isOwnedByDeployment(pod))) ||
+		(deletionPolicy == types.NodeDownPodDeletionPolicyDeleteBothStatefulsetAndDeploymentPodAndAnnotatedPods && (isOwnedByStatefulSet(pod) || isOwnedByDeployment(pod) || isAnnotatedWithDeletionAnnotation(pod)))
 
 	if !shouldDelete {
 		return nil
@@ -679,6 +682,15 @@ func isOwnedByDeployment(pod *corev1.Pod) bool {
 		return ownerRef.Kind == types.KubernetesReplicaSet
 	}
 	return false
+}
+
+func isAnnotatedWithDeletionAnnotation(pod *corev1.Pod) bool {
+	if val, ok := pod.ObjectMeta.Annotations[podDeletionAnnotation]; ok {
+		if strings.ToLower(val) == "true" {
+			return true
+		}
+	}
+	return false // We did not find the annotation, or the annotation is something else than "true" or "True"
 }
 
 // enqueuePodChange determines if the pod requires processing based on whether the pod has a PV created by us (driver.longhorn.io)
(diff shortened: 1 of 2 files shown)


derekbit:
@FireDrunk 
CI failed.
1. The commits need sign-off
2. The titles of the PR and the commits need to follow the rules of https://www.conventionalcommits.org/en/v1.0.0/#summary

controller/kubernetes_pod_controller.go
```diff
@@ -35,6 +35,8 @@ const (
 	controllerAgentName = "longhorn-kubernetes-pod-controller"
 
 	remountRequestDelayDuration = 5 * time.Second
+
+	podDeletionAnnotation = "longhorn.io/allow-deletion"
```
derekbit: To prevent breaking the default behavior, can we rename the annotation to 
```
podDisallowDeletionAnnotation = "longhorn.io/disallow-deletion"
```

derekbit:
@FireDrunk Can you sign off your commits? Thanks.

derekbit:
CI failed. Waiting for the fix by @FireDrunk

controller/kubernetes_pod_controller.go
innobead, line 39:
```suggestion
	podDeletionAnnotation = "longhorn.io/allow-pod-deletion-if-node-down"
```

types/setting.go
```diff
@@ -1781,10 +1781,11 @@ var (
...
-	NodeDownPodDeletionPolicyDeleteStatefulSetPod                  = NodeDownPodDeletionPolicy("delete-statefulset-pod")
-	NodeDownPodDeletionPolicyDeleteDeploymentPod                   = NodeDownPodDeletionPolicy("delete-deployment-pod")
-	NodeDownPodDeletionPolicyDeleteBothStatefulsetAndDeploymentPod = NodeDownPodDeletionPolicy("delete-both-statefulset-and-deployment-pod")
+	NodeDownPodDeletionPolicyDoNothing                                             = NodeDownPodDeletionPolicy("do-nothing") // Kubernetes default behavior
+	NodeDownPodDeletionPolicyDeleteStatefulSetPod                                  = NodeDownPodDeletionPolicy("delete-statefulset-pod")
+	NodeDownPodDeletionPolicyDeleteDeploymentPod                                   = NodeDownPodDeletionPolicy("delete-deployment-pod")
+	NodeDownPodDeletionPolicyDeleteBothStatefulsetAndDeploymentPod                 = NodeDownPodDeletionPolicy("delete-both-statefulset-and-deployment-pod")
+	NodeDownPodDeletionPolicyDeleteBothStatefulsetAndDeploymentPodAndAnnotatedPods = NodeDownPodDeletionPolicy("delete-both-statefulset-and-deployment-pod-and-annotated-pods")
```
innobead, line 1788: Should we retain one? Given that the new policy is essentially a superset of the current one, `delete-both-statefulset-and-deployment-pod`. 

Additionally, we might consider implementing this resilience for all pods managed by controllers, rather than exclusively for SS and DS workloads, as we currently do for the setting `auto-delete-pod-when-volume-detached-unexpectedly` handling mentioned below. @derekbit @longhorn/maintainer  WDYT?

[/controller/kubernetes_pod_controller.go#L634-L636](https://github.com/longhorn/longhorn-manager/blob/e6647b81bf21f5337dcb710a65e25ad7b941ef0a/controller/kubernetes_pod_controller.go#L634-L636)
```go
	if ownerRef := metav1.GetControllerOf(pod); ownerRef == nil || kc.isControllerInBlacklist(ownerRef) {
		return nil
	}
```

derekbit: > Additionally, we might consider implementing this resilience for all pods managed by controllers, rather than exclusively for SS and DS workloads, as we currently do for the setting auto-delete-pod-when-volume-detached-unexpectedly handling mentioned below. @derekbit @longhorn/maintainer WDYT?

Agree with the proposal!

FireDrunk: I've checked, and both our CNPG Operator pods and our Strimzi PodSet pods both have that ownerReference correctly set, which makes me believe that using that GetControllerOf function would work properly.

Does that mean it's no longer useful to have the annotation for pods for which the ownerReference is not explicitly set, and only support _properly_ owned pods?

I agree, but it's a bit of a reduction in functionality.

FireDrunk: Would this code change be OK?

```golang
SettingNameEnableDeletionOfControllerOwnedPods = SettingName("enable-deletion-of-controller-owned-pods")
```

```golang
// Check if we are allowed to delete the pod regardless of its type because it's owned by a Controller.
deletePodsOwnedByControllers, err := kc.ds.GetSettingAsBool(types.SettingNameEnableDeletionOfControllerOwnedPods)
if err != nil {
  return err
}

	// Check if the pod has an owning Controller reference
shouldDeleteBecauseOwnedByController := false
if deletePodsOwnedByControllers {
  if ownerRef := metav1.GetControllerOf(pod); ownerRef != nil && !kc.isControllerInBlacklist(ownerRef) {
	shouldDeleteBecauseOwnedByController = true
  }
}

// Check if the pod would be delted by a Policy
shouldDeleteBecauseOfPolicy := (deletionPolicy == types.NodeDownPodDeletionPolicyDeleteStatefulSetPod && isOwnedByStatefulSet(pod)) ||
		(deletionPolicy == types.NodeDownPodDeletionPolicyDeleteDeploymentPod && isOwnedByDeployment(pod)) ||
		(deletionPolicy == types.NodeDownPodDeletionPolicyDeleteBothStatefulsetAndDeploymentPod && (isOwnedByStatefulSet(pod) || isOwnedByDeployment(pod))) ||
		(deletionPolicy == types.NodeDownPodDeletionPolicyDeleteBothStatefulsetAndDeploymentPodAndAnnotatedPods && (isOwnedByStatefulSet(pod) || isOwnedByDeployment(pod) || isAnnotatedWithDeletionAnnotation(pod)))

// If both checks fail to decide to delete the pod, don't do anything.
if !shouldDeleteBecauseOwnedByController && !shouldDeleteBecauseOfPolicy {
  return nil
}
```

PR #4166, opened 2025-10-03
fix(csi): GetCapacity rpc must default to v1 engine

bachmanity1:
…ey is missing

Issue #https://github.com/longhorn/longhorn/issues/11906


derekbit:
@mergify backport v1.10.x

PR #4175, opened 2025-10-08
feat(replica_scheduler): introduce balance schedule algorithm

davidcheng0922:

https://github.com/longhorn/longhorn/issues/10512

Introduce a balanced replica schedule method, detail in https://github.com/longhorn/longhorn/pull/11922

Currently, `rebalanceWeightAlpha` is set to `1`, which balances replicas purely based on usable storage. To make this behavior configurable, we can introduce a new setting.

`Absolute size formula`: 
   - When disk capacities differ greatly, replicas are more likely to be scheduled on larger disks.
   
`Ratio size formula`: 
   - Replicas are distributed more evenly among all disks, though smaller disks may reach full capacity sooner.
   
`Hybrid formula`: 
   - Depends on the `rebalanceWeightAlpha` 
   - `hybrid = alpha*ratioImbalance + (1-alpha)*absoluteImbalance`


scheduler/replica_scheduler.go
```diff
@@ -691,6 +709,236 @@ func (rcs *ReplicaScheduler) getDiskWithMostUsableStorage(disks map[string]*Disk
...
+		// Compute usable storage for this disk
+		usable := disk.StorageAvailable - disk.StorageReserved - disk.StorageScheduled
+		if usable < 0 {
+			usable = 0
+		}
+
+		// Compute effective total capacity for ratio
+		total := disk.StorageMaximum - disk.StorageReserved
+		if total <= 0 {
+			total = 1
```
derekbit: Is this to avoid a divide-by-zero error?
If yes, can we set it to 0 and handle the divide-by-zero while compting the ratio?

COLDTURNIP:
@coderabbitai review

scheduler/replica_scheduler.go
```diff
@@ -691,6 +709,236 @@ func (rcs *ReplicaScheduler) getDiskWithMostUsableStorage(disks map[string]*Disk
...
+	// Ratio = usable / total for each entry.
+	// Formula: (max(ratio) - min(ratio)) / mean(ratio)
+
+	if len(usableStorage) == 0 || len(totalStorage) == 0 {
+		return math.MaxFloat64
+	}
+
+	sumRatio := 0.0
+	maxRatio := -math.MaxFloat64
+	minRatio := math.MaxFloat64
```
COLDTURNIP: Can change to IEEE754 `Inf` so that we don't need to worry about overflow, and is easy to check with `math.IsInf`.
```suggestion
	maxRatio := math.Inf(-1)
	minRatio := math.Inf(1)
```

scheduler/replica_scheduler.go
```diff
@@ -691,6 +709,236 @@ func (rcs *ReplicaScheduler) getDiskWithMostUsableStorage(disks map[string]*Disk
...
+			bestDisk = disk
+		}
+	}
+
+	if bestDisk == nil || bestDiskScore == math.MaxFloat64 {
+		return nil, fmt.Errorf("no candidate disk found for node %v", nodeID)
+	}
+
+	return bestDisk, nil
+}
```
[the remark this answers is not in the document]
davidcheng0922: Already do it in `FindDiskCandidates()`, all candidates have enough size

scheduler/replica_scheduler.go
davidcheng0922: Thanks, it's better

derekbit:
@COLDTURNIP @c3y1huang Please help review the PR. Thanks.

derekbit:
Weird. My review comments are somehow missing.

scheduler/replica_scheduler_test.go
```diff
@@ -1728,3 +1729,139 @@ func (s *TestSuite) TestScheduleReplicaToDiskOnLocalNode(c *C) {
 	rs.scheduleReplicaToDiskOnLocalNode(replica2, replicas, volume, diskCandidates)
 	c.Assert(replica2.Spec.NodeID, Equals, TestNode1)
 }
+
+func (s *TestSuite) TestGetDiskWithMostBalanceScore(c *C) {
```
derekbit, line 1735: Can you add a test case with storage available 0? Thanks.

scheduler/replica_scheduler.go
```diff
@@ -657,7 +666,16 @@ func (rcs *ReplicaScheduler) ListSchedulableNodes(dataEngine longhorn.DataEngine
...
 func (rcs *ReplicaScheduler) scheduleReplicaToDisk(replica *longhorn.Replica, diskCandidates map[string]*Disk) {
-	disk := rcs.getDiskWithMostUsableStorage(diskCandidates)
+	disk, err := rcs.getDiskWithMostBalanceScore(diskCandidates, replica.Spec.VolumeSize)
+	if err != nil {
+		logrus.WithError(err).Warnf("Failed to select disk for replica %v using balance score, falling back to selecting disk with most usable storage", replica.Name)
+	}
+
+	if disk == nil {
+		// should never happen
+		disk = rcs.getDiskWithMostUsableStorage(diskCandidates)
```
c3y1huang: Why not remove `getDiskWithMostUsableStorage` and return an error instead, since this case should never happen? If we decide to keep it, consider adding a log here.

scheduler/replica_scheduler.go
```diff
@@ -18,6 +20,13 @@ import (
 
 const (
 	FailedReplicaMaxRetryCount = 5
+
+	// rebalanceWeightAlpha controls the weighting between absolute and ratio imbalance
+	// when computing the hybrid balance score for replica scheduling.
+	// - 1.0 means only absolute imbalance is considered (difference in usable bytes).
+	// - 0.0 means only ratio imbalance is considered (difference in usable/total ratio).
+	// - Values between 0 and 1 blend both metrics.
+	rebalanceWeightAlpha = 1.0
```
c3y1huang, line 29: Looks like Longhorn will only uses absolute imbalance for picking disks. I guess the other strategies are just there in case we want to switch later?

davidcheng0922: Yeah, currently only use absolute imbalance method, for other strategies just in case for future use if we need

derekbit:
@davidcheng0922 CI failed.

scheduler/replica_scheduler_test.go
```diff
@@ -1728,3 +1729,153 @@ func (s *TestSuite) TestScheduleReplicaToDiskOnLocalNode(c *C) {
...
+	c.Assert(node, Equals, "")
+	c.Assert(err, NotNil)
+
+	// Normal: pick node with best balance
+	nodeUsable = map[string]int64{"n1": 100, "n2": 200}
+	nodeTotal = map[string]int64{"n1": 100, "n2": 200}
+	node, err = selectBestNode(nodeUsable, nodeTotal, 50)
+	c.Assert(err, IsNil)
+	c.Assert(node, Not(Equals), "")
+}
```
c3y1huang: Consider make the tests table-driven to reduce repetitive code and make the test easier to read and maintain.
Example:
```go
func (s *TestSuite) TestSelectBestNode(c *C) {
	tests := []struct {
		name        string
		nodeUsable  map[string]int64
		nodeTotal   map[string]int64
		replicaSize int64
		expectNode  string
		expectErr   bool
	}{
		{
			name:        "no available nodes",
			nodeUsable:  map[string]int64{},
			nodeTotal:   map[string]int64{},
			replicaSize: 10,
			expectNode:  "",
			expectErr:   true,
		},
		// ...
	}

	for _, tt := range tests {
		c.Logf("Running test: %s", tt.name)
		// test logic here
	}
}
```

scheduler/replica_scheduler.go
```diff
@@ -671,26 +685,243 @@ func (rcs *ReplicaScheduler) scheduleReplicaToDisk(replica *longhorn.Replica, di
...
 }
 
-// Investigate
-func (rcs *ReplicaScheduler) getDiskWithMostUsableStorage(disks map[string]*Disk) *Disk {
-	diskWithMostUsableStorage := &Disk{}
-	for _, disk := range disks {
-		diskWithMostUsableStorage = disk
-		break
+// getDiskWithMostBalanceScore selects a disk for a replica by minimizing imbalance.
+func (rcs *ReplicaScheduler) getDiskWithMostBalanceScore(candidateDisks map[string]*Disk, replicaSize int64) (*Disk, error) {
```
derekbit: Just noticed the previous implementation `getDiskWithMostUsableStorage` always picks the only one disk when there is only disk. Can we keep the same behavior in the new implementation?

davidcheng0922:
```go

// getDiskWithMostBalanceScore selects a disk for a replica by minimizing imbalance.
func (rcs *ReplicaScheduler) getDiskWithMostBalanceScore(candidateDisks map[string]*Disk, replicaSize int64) (*Disk, error) {
	// It works in two stages:
	//   1. Find the best node: simulate placing the replica on each candidate node,
	//      and pick the one that leads to the most balanced usable storage distribution across nodes.
	//   2. Within that node, find the best disk: simulate placing the replica on each candidate disk,
	//      and pick the one that leads to the most balanced usable storage distribution among disks on that node.

	if len(candidateDisks) == 1 {
		for _, disk := range candidateDisks {
			usable := disk.StorageAvailable - disk.StorageReserved - disk.StorageScheduled
			if usable < replicaSize {
				return nil, fmt.Errorf("no candidate disk has enough usable storage")
			}
			return disk, nil
		}
	}
```

this already check the candidate disks, if only one disk, will not do the following computing

derekbit: So, do we need to return `err`?

davidcheng0922: after meeting, fix it

scheduler/replica_scheduler.go
```diff
@@ -671,26 +681,245 @@ func (rcs *ReplicaScheduler) scheduleReplicaToDisk(replica *longhorn.Replica, di
...
+	for _, disk := range candidateDisks {
+		// Compute usable storage for this disk
+		usable := disk.StorageAvailable - disk.StorageReserved - disk.StorageScheduled
+		if usable < 0 {
+			usable = 0
+		}
+
+		// Compute effective total capacity for ratio
+		total := disk.StorageMaximum - disk.StorageReserved
+		if total <= 0 {
```
derekbit: nit: 
```suggestion
		if total < 0 {
```

derekbit (commented):
LGTM

derekbit (approved):

derekbit (approved):

types/types.go
```diff
@@ -841,7 +841,7 @@ func ErrorRecordNotFoundButLvolFound(err error) bool {
 }
 
 func ErrorIsNotFound(err error) bool {
-	return strings.Contains(err.Error(), "cannot find")
+	return strings.Contains(err.Error(), "cannot find") || strings.Contains(err.Error(), "NotFound")
```
derekbit: Do we need the change in this PR?

davidcheng0922: Ah sorry, it is another ticket changes. I'll remove it

davidcheng0922: Thank you for catching!

PR #4203, opened 2025-10-16
fix: fix NPE issue when recurring jobs handles a deleted volume

shuo-wu:

Since API `volumeAPI.ById(volumeName)` will not return any error when the volume is gone, the job needs to handle nil volume case separately.

https://github.com/longhorn/longhorn/issues/11925

### DIFF
--- a/app/recurringjob/volume.go
+++ b/app/recurringjob/volume.go
@@ -144,7 +144,11 @@ func (job *VolumeJob) run() (err error) {
 	volumeName := job.volumeName
 	volume, err := volumeAPI.ById(volumeName)
 	if err != nil {
-		return errors.Wrapf(err, "could not get volume %v", volumeName)
+		return errors.Wrapf(err, "failed to get volume %v", volumeName)
+	}
+	if volume == nil {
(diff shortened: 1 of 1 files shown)


PhanLe1010 (approved):

app/recurringjob/volume.go
```diff
@@ -144,7 +144,11 @@ func (job *VolumeJob) run() (err error) {
 	volumeName := job.volumeName
 	volume, err := volumeAPI.ById(volumeName)
 	if err != nil {
-		return errors.Wrapf(err, "could not get volume %v", volumeName)
+		return errors.Wrapf(err, "failed to get volume %v", volumeName)
+	}
+	if volume == nil {
+		job.logger.Infof("Volume %v not found, skipping", volumeName)
+		return nil
 	}
```
derekbit, line 152: Thanks @shuo-wu.
IIRC, we hit a similar issue before. Quite annoying.
cc @mantissahz

derekbit (approved):

derekbit:
@mergify backport v1.10.x v1.9.x v1.8.x

PR #4366, opened 2025-12-16
feat(disk): synchronized disk scheduling

COLDTURNIP:
longhorn-10422

Issue longhorn/longhorn#10422

Reconcile disk scheduling using a single-thread disk schedule controller.


derekbit:
One question
Why do we need to introduce a new CRD rather than using the original node CRD? Any advantage?

COLDTURNIP:
@coderabbitai review

k8s/pkg/apis/longhorn/v1beta2/diskSchedule.go
```diff
@@ -0,0 +1,150 @@
...
+	// Scheduled status of backing images
+	// +optional
+	ScheduledBackingImageStatus map[string]*DiskScheduledResourcesStatus `json:"scheduledBackingImageStatus"`
+	// Sum of successfully scheduled storage
+	// +optional
+	StorageScheduled int64 `json:"storageScheduled"`
+	// +optional
+	// +nullable
+	Conditions []Condition `json:"conditions"`
+}
```
derekbit, line 55: Shouldn't we add `OwnerID`? The diskSchedule might be handled by a controller on another node?

controller/disk_schedule_controller.go
```diff
@@ -0,0 +1,463 @@
...
+	}
+
+	var err error
+	if _, err = ds.DiskScheduleInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    c.enqueueDiskSchedule,
+		UpdateFunc: func(old, cur any) { c.enqueueDiskSchedule(cur) },
+		DeleteFunc: c.enqueueDiskSchedule,
+	}); err != nil {
+		return nil, err
+	}
```
derekbit, line 84: If a Longhorn Node is somehow created, updated or deleted, do we need to enqueue the DiskSchedule CRs?

upgrade/v110xto1110/upgrade.go
```diff
@@ -0,0 +1,46 @@
+package v110xto1110
+
+import (
+	"github.com/cockroachdb/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	clientset "k8s.io/client-go/kubernetes"
+
+	lhclientset "github.com/longhorn/longhorn-manager/k8s/pkg/client/clientset/versioned"
+	upgradeutil "github.com/longhorn/longhorn-manager/upgrade/util"
+)
```
derekbit, line 18: Reorg

controller/node_controller.go
```diff
@@ -687,6 +690,117 @@ func (nc *NodeController) syncDiskStatus(node *longhorn.Node, collectedDataInfo
...
+		diskSchedule = &longhorn.DiskSchedule{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: diskUUID,
+			},
+			Spec: longhorn.DiskScheduleSpec{
+				Name:               diskStatus.DiskName,
+				NodeID:             node.Name,
+				SchedulableMaximum: schedulableMaximum,
+				AllowScheduling:    diskAllowScheduling,
+			},
```
derekbit: Should we set owner reference?

controller/node_controller.go
```diff
@@ -687,6 +690,117 @@ func (nc *NodeController) syncDiskStatus(node *longhorn.Node, collectedDataInfo
...
+	}
+	return nil
+}
+
+func (nc *NodeController) createOrUpdateDiskSchedule(node *longhorn.Node, diskSpec *longhorn.DiskSpec, diskStatus *longhorn.DiskStatus) error {
+	diskUUID := diskStatus.DiskUUID
+	diskAllowScheduling := node.Spec.AllowScheduling && diskSpec.AllowScheduling
+	schedulableMaximum := diskStatus.StorageMaximum - diskSpec.StorageReserved
+
+	diskSchedule, getErr := nc.ds.GetDiskSchedule(diskUUID)
```
derekbit: There is no variable shadowing here. Reusing `err` is fine, and introducing too many `xxxErr` variables actually makes the code harder to read.


```suggestion
	diskSchedule, err := nc.ds.GetDiskSchedule(diskUUID)
```

controller/node_controller.go
```diff
@@ -687,6 +690,117 @@ func (nc *NodeController) syncDiskStatus(node *longhorn.Node, collectedDataInfo
...
+				Name: diskUUID,
+			},
+			Spec: longhorn.DiskScheduleSpec{
+				Name:               diskStatus.DiskName,
+				NodeID:             node.Name,
+				SchedulableMaximum: schedulableMaximum,
+				AllowScheduling:    diskAllowScheduling,
+			},
+		}
+		_, createErr := nc.ds.CreateDiskSchedule(diskSchedule)
```
```suggestion
		_, err := nc.ds.CreateDiskSchedule(diskSchedule)
```

controller/node_controller.go
```diff
@@ -687,6 +690,117 @@ func (nc *NodeController) syncDiskStatus(node *longhorn.Node, collectedDataInfo
...
+		return errors.Wrapf(createErr, "failed to update disk schedule")
```
```suggestion
		return errors.Wrapf(err, "failed to create disk schedule %v", ...)
```

controller/node_controller.go
```diff
@@ -687,6 +690,117 @@ func (nc *NodeController) syncDiskStatus(node *longhorn.Node, collectedDataInfo
...
+	}
+
+	diskSchedule.Spec.AllowScheduling = diskSpec.AllowScheduling
```
derekbit: Why do we need this assignment? Isn't it set above?

controller/node_controller.go
```diff
@@ -687,6 +690,117 @@ func (nc *NodeController) syncDiskStatus(node *longhorn.Node, collectedDataInfo
...
+	diskSchedule.Spec.SchedulableMaximum = schedulableMaximum
+
+	_, updateErr := nc.ds.UpdateDiskSchedule(diskSchedule)
```
```suggestion
	_, err := nc.ds.UpdateDiskSchedule(diskSchedule)
```

controller/node_controller.go
```diff
@@ -687,6 +690,117 @@ func (nc *NodeController) syncDiskStatus(node *longhorn.Node, collectedDataInfo
 	return nc.updateDiskStatusSchedulableCondition(node)
 }
 
+func (nc *NodeController) syncDiskSchedule(node *longhorn.Node) error {
+	if err := nc.cleanupDiskSchedule(node); err != nil {
+		return err
+	}
+
+	if err := nc.syncDiskScheduleSpec(node); err != nil {
```
derekbit, line 760: We always need sync Node and DiskSchedule and introduce extra efforts. Why not directly store the disk information in the node CRs?

k8s/pkg/apis/longhorn/v1beta2/diskSchedule.go
COLDTURNIP: Similar to `Node` CRs, `DiskSchedule` CRs are always owned by the corresponding node. Other node reconcile this CR only when deletion.

controller/disk_schedule_controller.go
COLDTURNIP: - **Creation**: disk monitor confirms the disk creation and the UUID is generated, then the node controller creates a DiskSchedule CR
- **Update**: node controller reconcile the disk status, then sync the status, includes the capacity and disablement, to the corresponding DiskSchedule CR
- **Delete**: once node controller decide to remove a disk from its disk list, it will also trigger deletion on the corresponding DiskSchedule CR

controller/node_controller.go
COLDTURNIP: Good idea. DiskSchedule CR could be owned by Node CR.

controller/node_controller.go
COLDTURNIP: `diskSpec.AllowScheduling` is a mestake. Should be `diskAllowScheduling` defined at line 770.

controller/disk_schedule_controller.go
COLDTURNIP: Without syncing the the capacity and disablement, then node controller doesn't need to update the DiskSchedule CR.

COLDTURNIP: > If a Longhorn Node is somehow created, updated or deleted, do we need to enqueue the DiskSchedule CRs?

Absolutely. Volume and Backing Image controllers submit allocation requirement in the DiskSchedule CR, and the DiskSchedule controller reconciles then update the allocation result in the CR status.

controller/node_controller.go
COLDTURNIP: The original idea is to sync the maximum storage to DiskSchedule. Now we keep this information in the Node CR, and the ReplicaScheduler will calculate the scheduling score from [both Node and DiskSchedule](https://github.com/COLDTURNIP/longhorn-manager/blob/10422-disk_schedule_controller/scheduler/replica_scheduler.go#L1406) CRs.

k8s/pkg/apis/longhorn/v1beta2/diskSchedule.go
derekbit: But how can we avoid multiple nodes competing for a single DiskSchedule CR?

controller/node_controller.go
```diff
@@ -459,6 +459,9 @@ func (nc *NodeController) syncNode(key string) (err error) {
 	if err := nc.syncDiskStatus(node, collectedDiskInfo); err != nil {
 		return err
 	}
+	if err := nc.syncDiskSchedule(node); err != nil {
+		return err
+	}
```
derekbit, line 510: When a node is deleted and the deletionTimestamp is set, will the function be called? Who will delete the DiskSchedule CRs for the deleted node?

derekbit (commented):
Do we need to update the uninstall controller?

k8s/pkg/apis/longhorn/v1beta2/diskSchedule.go
COLDTURNIP: When a DiskSchedule update event occurs, [only the node that hosts the disk can reconcile the DiskSchedule CR](https://github.com/COLDTURNIP/longhorn-manager/blob/10422-disk_schedule_controller/controller/disk_schedule_controller.go#L181-L184). When the system is deleting the node or disk, no further allocation is needed, and [any node is able to blindly remove the finalizer from DiskSchedule CR](https://github.com/COLDTURNIP/longhorn-manager/blob/10422-disk_schedule_controller/controller/disk_schedule_controller.go#L177-L179).

controller/node_controller.go
COLDTURNIP: [DiskSchedule CR has an owner reference points to the corresponding Node CR](https://github.com/COLDTURNIP/longhorn-manager/blob/10422-disk_schedule_controller/controller/node_controller.go#L787). Kubernetes will trigger DiskSchedule CR once the Node is deleted.

In addition, when a disk is removed from the Node spec, [the deletion of DiskSchedule CR will be triggered by the Node controller](https://github.com/COLDTURNIP/longhorn-manager/blob/10422-disk_schedule_controller/controller/node_controller.go#L1702-L1706).

COLDTURNIP:
> Do we need to update the uninstall controller?

Sure. Just added [this part](https://github.com/COLDTURNIP/longhorn-manager/blob/10422-disk_schedule_controller/controller/uninstall_controller.go#L622-L627).

COLDTURNIP:
Some CSI-related unit test cases are still broken. Fixing.

COLDTURNIP:
Passing all unit test cases. But there is some endless event queuing problem for backing image. The replica allocation is ready to review.

derekbit:
> Passing all unit test cases. But there is some endless event queuing problem for backing image. The replica allocation is ready to review.

To save time, I will review the PR when all issues are resolved.

In addtion. please add the Jenkins
- regression testing
- e2e testing
- negative testing

links to make sure all test cases are passed.
Thank you.

cc @longhorn/qa

controller/controller_test.go
```diff
@@ -565,7 +565,6 @@ func newNode(name, namespace string, allowScheduling bool, status longhorn.Condi
 			DiskStatus: map[string]*longhorn.DiskStatus{
 				TestDiskID1: {
 					StorageAvailable: TestDiskAvailableSize,
-					StorageScheduled: 0,
```
derekbit, line 573: Why do we need to remove the initialization?

controller/disk_schedule_controller.go
```diff
@@ -0,0 +1,579 @@
...
+import (
+	"fmt"
+	"math/rand"
+	"reflect"
+	"sync"
+	"time"
+
+	"github.com/cockroachdb/errors"
+	"github.com/sirupsen/logrus"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
```
derekbit: Reorganize the imports

controller/disk_schedule_controller.go
```diff
@@ -0,0 +1,579 @@
...
+
+	var err error
+	if _, err = ds.DiskScheduleInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    c.enqueueDiskSchedule,
+		UpdateFunc: func(old, cur any) { c.enqueueDiskSchedule(cur) },
+		DeleteFunc: c.enqueueDiskSchedule,
+	}); err != nil {
+		return nil, err
+	}
+	c.cacheSyncs = append(c.cacheSyncs, ds.DiskScheduleInformer.HasSynced)
```
derekbit, line 85: Should we add Longhorn Node EventHandler?

controller/disk_schedule_controller.go
```diff
@@ -0,0 +1,579 @@
...
+	if !cache.WaitForNamedCacheSync(dsc.name, stopCh, dsc.cacheSyncs...) {
+		return
+	}
+
+	go wait.Until(dsc.worker, time.Second, stopCh)
+
+	<-stopCh
+}
+
+func (dsc *DiskScheduleController) worker() {
```
derekbit, line 122: How many workers are in the disk schedule controller?

webhook/resources/diskschedule/validator.go
```diff
@@ -0,0 +1,131 @@
+package diskschedule
+
+import (
+	"fmt"
+
+	"github.com/cockroachdb/errors"
+	"github.com/longhorn/longhorn-manager/types"
+	"go.uber.org/multierr"
```
derekbit, line 7: Reorg the imports

controller/node_controller.go
```diff
@@ -687,6 +690,111 @@ func (nc *NodeController) syncDiskStatus(node *longhorn.Node, collectedDataInfo
...
+
+	diskScheduleMap, err := nc.ds.ListDiskSchedulesOnNode(node.Name)
+	if err != nil && !datastore.ErrorIsNotFound(err) {
+		return errors.Wrapf(err, "failed to list disk schedule on node %v for clean up", node.Name)
+	}
+
+	deleteErrs := multierr.NewMultiError()
+	for _, diskSchedule := range diskScheduleMap {
+		if !nodeDisks[diskSchedule.Name] {
+			nc.logger.Infof("Clearning disk schedule %v (%q) on node %v", diskSchedule.Name, diskSchedule.Spec.Name, node.Name)
```
```suggestion
			nc.logger.Infof("Deleting DiskDchedule %v (%q) from Node %v", diskSchedule.Name, diskSchedule.Spec.Name, node.Name)
```

controller/node_controller.go
```diff
@@ -687,6 +690,111 @@ func (nc *NodeController) syncDiskStatus(node *longhorn.Node, collectedDataInfo
...
+			if deleteErr := nc.ds.DeleteDiskSchedule(diskSchedule.Name); deleteErr != nil {
```
derekbit: nit:
Simplify the variable naming. Apply to other functions.

```suggestion
			if err := nc.ds.DeleteDiskSchedule(diskSchedule.Name); err != nil {
```

controller/node_controller.go
```diff
@@ -687,6 +690,111 @@ func (nc *NodeController) syncDiskStatus(node *longhorn.Node, collectedDataInfo
...
+				deleteErrs.Append(fmt.Sprintf("%v (%v)", diskSchedule.Name, diskSchedule.Spec.Name), deleteErr)
+			}
+		}
+	}
+
+	if len(deleteErrs) > 0 {
+		return fmt.Errorf("failed to delete disk schedules on node %v: %v", node.Name, deleteErrs.ErrorByReason("errors"))
```
```suggestion
		return fmt.Errorf("failed to delete DiskSchedules on node %v: %v", node.Name, deleteErrs.ErrorByReason("errors"))
```

controller/node_controller.go
```diff
@@ -687,6 +690,111 @@ func (nc *NodeController) syncDiskStatus(node *longhorn.Node, collectedDataInfo
...
+	return nil
+}
+
+func (nc *NodeController) cleanupDiskSchedule(node *longhorn.Node) error {
+	nodeDisks := map[string]bool{}
+	if node.DeletionTimestamp.IsZero() && node.Status.DiskStatus != nil {
+		for _, diskStatus := range node.Status.DiskStatus {
+			nodeDisks[diskStatus.DiskUUID] = true
+		}
+	}
```
derekbit: Why do we need to extra copy? 
Can't we directly check `node.Status.DiskStatus`? Another question, why not check `node.spec.disks`?

controller/node_controller.go
```diff
@@ -687,6 +690,111 @@ func (nc *NodeController) syncDiskStatus(node *longhorn.Node, collectedDataInfo
...
+
+	diskScheduleMap, err := nc.ds.ListDiskSchedulesOnNode(node.Name)
+	if err != nil && !datastore.ErrorIsNotFound(err) {
+		return errors.Wrapf(err, "failed to list disk schedule on node %v for clean up", node.Name)
```
```suggestion
		return errors.Wrapf(err, "failed to list DiskSchedules on node %v for cleanup", node.Name)
```

derekbit (commented):
Besides

controller/controller_test.go
COLDTURNIP: `Node.Status.DiskStatus[*].StorageScheduled` is now replaced by `DiskSchedule.Status.StorageScheduled`, and will be used by the `ReplicaScheduler`.

controller/disk_schedule_controller.go
COLDTURNIP: After the DiskSchedule CR is created by the Node controller, there is nothing to sync to the DiskSchedule spec. The disk information, including `AllowScheduling` and `StorageMaximum`, are looked up each time when DiskSchedule reconciles the DiskSchedule CR.

controller/disk_schedule_controller.go
COLDTURNIP: There is always only one controller worker thread on each node. Commented inside the `DiskScheduleController.worker`

controller/node_controller.go
COLDTURNIP: `NodeController.cleanupDiskSchedule` deletes the DiskSchedule CRs if the disk is no longer considered to be hosted by the node. DiskSchedule CRs are named using the disk UUID. Hence in this method:

1. Iterate over `Node.Status.DiskStatus`, which is a "disk name to status" map, and convert the map into the set of disk UUID
2. List all existing DiskSchedule CRs in the cluster
3. Remove DiskSchedule CRs that not list in the disk UUID set in the first step

COLDTURNIP: After re-think, yes we can compare existing DiskSchedule CRs with the `node.Status.DiskStatus`. The reason we compare only with `node.Status.DiskStatus` is to align the disk UUID, which is not included in the disk spec.

COLDTURNIP:
> Passing all unit test cases. But there is some endless event queuing problem for backing image. The replica allocation is ready to review.

The above Backing Image (BI) disk allocation conflict problem is resolved.

- Ideally, when BI controller decide to create a copy to a disk, it should submit the disk allocation request on DiskSchedule CR, waits for the DiskSchedule controller accept the allocation, then update the copy to the Backing Image Manager (BIM) CR
- The problem is, when BI controller update the DiskSchedule CR, the new image copy has not been updated to the BI spec; DiskSchedule receives the allocation request, and check with the BI spec, then the DiskSchedule cannot decide whether the allocation is still needed or not, then remove it from the DiskSchedule.

To resolve the problem,

- DiskSchedule CR removes allocation requirement of a Backing Image copy only when the BI CR is deleted.
- BI controller is responsible to remove the allocation requirement on DiskSchedule CR once a copy is no longer needed on a disk.

controller/controller_test.go
derekbit: Please deprecate the fields.

k8s/pkg/apis/longhorn/v1beta2/node.go
```diff
@@ -173,13 +173,16 @@ type DiskStatus struct {
...
+	// StorageScheduled is moved to DiskSchedule
 	// +optional
 	StorageScheduled int64 `json:"storageScheduled"`
 	// +optional
 	StorageMaximum int64 `json:"storageMaximum"`
+	// ScheduledReplica is moved to DiskSchedule
 	// +optional
 	// +nullable
 	ScheduledReplica map[string]int64 `json:"scheduledReplica"`
+	// ScheduledBackingImage is moved to DiskSchedule
```
derekbit, line 194: Please deprecate the fields.

controller/volume_controller.go
```diff
@@ -622,6 +624,43 @@ func (c *VolumeController) ReconcileEngineReplicaState(v *longhorn.Volume, es ma
 		}
 	}()
 
+	// Aggregate replica wait for disk allocation condition
+	aggregatedReplicaDiskAllocationError := multierr.NewMultiError()
```
derekbit: Using a shorter name is easy to read. Apply to other codes.

```suggestion
	errs := multierr.NewMultiError()
```

controller/volume_controller.go
```diff
@@ -622,6 +624,43 @@ func (c *VolumeController) ReconcileEngineReplicaState(v *longhorn.Volume, es ma
...
+	waitForDiskAllocation := false
+	for _, r := range rs {
+		diskAllocationCondition := types.GetCondition(r.Status.Conditions, longhorn.ReplicaConditionTypeDiskAllocation)
```
derekbit: nit:
Same as above. 

```suggestion
		condition := types.GetCondition(r.Status.Conditions, longhorn.ReplicaConditionTypeDiskAllocation)
```

k8s/pkg/apis/longhorn/v1beta2/node.go
COLDTURNIP: Marking "Deprecated: .." makes the go vet reject variable reset by node controller. We have to attach deprecation at the next version.

controller/controller_test.go
COLDTURNIP: Same with [comment](https://github.com/longhorn/longhorn-manager/pull/4366#discussion_r2650023851)

k8s/pkg/apis/longhorn/v1beta2/node.go
derekbit: Then, please add `TODO`

k8s/pkg/apis/longhorn/v1beta2/replica.go
```diff
@@ -10,13 +10,21 @@ const (
...
 	ReplicaConditionReasonWaitForBackingImageWaiting = "Waiting"
 
 	ReplicaConditionReasonRebuildFailedDisconnection = "Disconnection"
 	ReplicaConditionReasonRebuildFailedGeneral       = "General"
+
+	ReplicaConditionReasonDiskAllocationError    = "Error"
+	ReplicaConditionReasonDiskAllocationPending  = "Pending"
+	ReplicaConditionReasonDiskAllocationWaiting  = "Waiting"
+	ReplicaConditionReasonDiskAllocationAccepted = "Accepted"
+	ReplicaConditionReasonDiskAllocationRejected = "Rejected"
```
derekbit: As discussed today, the disk schedule only judges if a replica can be scheduled. If not, the replica should be marked as failed. Therefore, some of the reasons can be removed.

controller/disk_schedule_controller.go
```diff
@@ -0,0 +1,570 @@
...
+	replicas, replicaErr := dsc.ds.ListReplicasByDiskUUID(diskSchedule.Name)
+	if replicaErr != nil {
+		return errors.Wrapf(replicaErr, "failed to list existing replicas for disk schedule %v", diskSchedule.Name)
+	}
+	for _, replica := range replicas {
+		if !isReplicaAllocatedToDisk(replica, diskSchedule) {
+			continue
+		}
+		logger.Infof("sync from allocated replica %v for disk schedule %v", replica.Name, diskSchedule.Name)
+		updateAllocation(&diskSchedule.Spec.SchedulingReplicas, &diskSchedule.Status.ScheduledReplicaStatus, replica.Name, replica.Spec.VolumeSize)
```
derekbit: Suggest directly passing diskSchedule to the function `updateAllocation`.

controller/disk_schedule_controller.go
```diff
@@ -0,0 +1,570 @@
...
+
+	// sync from existing replicas
+	replicas, replicaErr := dsc.ds.ListReplicasByDiskUUID(diskSchedule.Name)
+	if replicaErr != nil {
+		return errors.Wrapf(replicaErr, "failed to list existing replicas for disk schedule %v", diskSchedule.Name)
+	}
+	for _, replica := range replicas {
+		if !isReplicaAllocatedToDisk(replica, diskSchedule) {
+			continue
+		}
```
derekbit: If a replica is unable to allocate space, we need to mark it failed. Where to fail it?

controller/disk_schedule_controller.go
```diff
@@ -0,0 +1,570 @@
...
+			RequiredSize:  size,
+			ScheduledSize: size,
+		}
+	}
+
+	// sync from existing replicas
+	replicas, replicaErr := dsc.ds.ListReplicasByDiskUUID(diskSchedule.Name)
+	if replicaErr != nil {
+		return errors.Wrapf(replicaErr, "failed to list existing replicas for disk schedule %v", diskSchedule.Name)
+	}
```
derekbit: nit: 
Simplify the vairiable naming. Apply to other variables.

```suggestion
	replicas, err := dsc.ds.ListReplicasByDiskUUID(diskSchedule.Name)
	if err != nil {
		return errors.Wrapf(err, "failed to list existing replicas for disk schedule %v", diskSchedule.Name)
	}
```

controller/disk_schedule_controller.go
```diff
@@ -0,0 +1,570 @@
...
+
+	// sync from existing backing images
+	bisMap, biErr := dsc.ds.GetCurrentDiskBackingImageMap()
+	if biErr != nil {
+		return errors.Wrapf(biErr, "failed to list existing backing image for disk schedule %v", diskSchedule.Name)
+	}
+	bims, bimErr := dsc.ds.ListBackingImageManagersByDiskUUID(diskSchedule.Name)
+	if bimErr != nil {
+		return errors.Wrapf(bimErr, "failed to list existing backing image manager for disk schedule %v", diskSchedule.Name)
+	}
```
derekbit:
```suggestion
	bisMap, err := dsc.ds.GetCurrentDiskBackingImageMap()
	if biErr != nil {
		return errors.Wrapf(biErr, "failed to list existing backing image for disk schedule %v", diskSchedule.Name)
	}
	bims, err := dsc.ds.ListBackingImageManagersByDiskUUID(diskSchedule.Name)
	if bimErr != nil {
		return errors.Wrapf(bimErr, "failed to list existing backing image manager for disk schedule %v", diskSchedule.Name)
	}
```

controller/disk_schedule_controller.go
```diff
@@ -0,0 +1,570 @@
...
+	}()
+
+	if dsc.isDiskScheduleInitialing(diskSchedule) {
+		// blindly record the allocation for existing resource
+		if err := dsc.syncFromAllocatedResources(diskSchedule); err != nil {
+			return err
+		}
+	}
+
+	if err := dsc.syncResourceDeallocation(diskSchedule); err != nil {
```
derekbit, line 237: Same as above

controller/disk_schedule_controller.go
```diff
@@ -0,0 +1,570 @@
...
+		return err
+	}
+
+	if err := dsc.syncResourceAllocation(diskSchedule); err != nil {
```
derekbit, line 241: Same as above.

controller/disk_schedule_controller.go
```diff
@@ -0,0 +1,570 @@
...
+		}
+		if _, updateStatusErr := dsc.ds.UpdateDiskScheduleStatus(diskSchedule); updateStatusErr != nil && datastore.ErrorIsConflict(errors.Cause(updateStatusErr)) {
+			log.WithError(updateStatusErr).Debugf("Requeue %v due to status conflict", diskScheduleName)
+			dsc.enqueueDiskSchedule(diskSchedule)
+		}
+	}()
+
+	if dsc.isDiskScheduleInitialing(diskSchedule) {
+		// blindly record the allocation for existing resource
+		if err := dsc.syncFromAllocatedResources(diskSchedule); err != nil {
```
derekbit: What does Resource mean? If it is space, sugguest renaming it to syncFromAllocatedSpace to make the function easy to understand.

derekbit (commented):
@COLDTURNIP What's the status of the PR? Ready to review?

controller/backing_image_controller.go
```diff
@@ -1305,6 +1327,31 @@ func (bic *BackingImageController) syncBackingImageFileInfo(bi *longhorn.Backing
...
+		diskScheduleMap = nil
+	}
+	for _, diskSchedule := range diskScheduleMap {
+		allocationStatus, exist := diskSchedule.Status.ScheduledBackingImageStatus[bi.Name]
+		if !exist {
+			continue
+		}
+		if allocationStatus.RequiredSize != bi.Status.Size {
+			continue
+		}
```
shuo-wu: Do we need to handle this size mismatch?

COLDTURNIP:
After rethink and discussed, to eliminate logic divergence, to replica disk allocation, we're going to move the DiskSchedule update & check logic to Volume controller, rather than check DiskSchedule CR just before creating instance.

PR #4410, opened 2026-01-09
fix(v2): recover DR volume replicas after instance manager restart

c3y1huang:

Issue longhorn/longhorn#12412

https://github.com/longhorn/longhorn/issues/12412#issuecomment-3722260785

`None`

`None`


derekbit:
The issue is gone.

PR #4571, opened 2026-03-02
fix(orphan): cr-level data orphan deletion for node or disk eviction

COLDTURNIP:

Issue longhorn/longhorn#7795

### DIFF
--- a/controller/node_controller.go
+++ b/controller/node_controller.go
@@ -1497,11 +1497,6 @@ func (nc *NodeController) deleteOrphansForReplicaDataStore(node *longhorn.Node, 
 	if err != nil {
 		return errors.Wrapf(err, "failed to get %v setting", types.SettingNameOrphanResourceAutoDeletion)
 	}
-	autoDeleteEnabled, ok := autoDeletionTypes[types.OrphanResourceTypeReplicaData]
-	if !ok {
-		autoDeleteEnabled = false
-	}
-
 	for dataStore := range missingOrphanedReplicaDataStores {
 		orphanName := types.GetOrphanChecksumNameForOrphanedDataStore(node.Name, diskName, diskInfo.Path, diskInfo.DiskUUID, dataStore)
 		if err := nc.ds.DeleteOrphan(orphanName); err != nil && !datastore.ErrorIsNotFound(err) {
@@ -1515,7 +1510,7 @@ func (nc *NodeController) deleteOrphansForReplicaDataStore(node *longhorn.Node, 
 	}
 
 	for _, orphan := range orphans {
-		if nc.canDeleteOrphan(orphan, autoDeleteEnabled, autoDeleteGracePeriod) {
+		if nc.canDeleteOrphan(orphan, node, autoDeletionTypes[types.OrphanResourceTypeReplicaData], autoDeleteGracePeriod) {
 			if err := nc.ds.DeleteOrphan(orphan.Name); err != nil && !datastore.ErrorIsNotFound(err) {
 				return errors.Wrapf(err, "failed to delete orphan %v", orphan.Name)
 			}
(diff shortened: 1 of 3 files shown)


controller/node_controller.go
```diff
@@ -1502,16 +1497,62 @@ func (nc *NodeController) canDeleteOrphan(orphan *longhorn.Orphan, autoDeleteEna
...
 	return canDelete
 }
 
+func (nc *NodeController) canDeleteOrphanDataFromNodeState(orphan *longhorn.Orphan, node *longhorn.Node) (bool, string) {
+	if orphan.Spec.Type != longhorn.OrphanTypeReplicaData {
+		return false, ""
+	}
+
+	if orphan.Spec.NodeID != node.Name {
+		return false, ""
```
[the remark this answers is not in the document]
COLDTURNIP: This change request is invalid. A data Orphan CR is always bind to a Node since it is the files stored in a disk. In this function, which is triggered by the corresponding Node change event, we check the removal criteria with the node we're reconciling.

COLDTURNIP:
Convert to draft: the Orphan CR deletion event arrived before the node status update. Need to refine the event order.

COLDTURNIP:
The PR is now ready for review.

controller/node_controller.go
```diff
@@ -1524,34 +1533,73 @@ func (nc *NodeController) deleteOrphansForReplicaDataStore(node *longhorn.Node,
...
 			autoDeleteAllowed = true
 		}
 	}
 
-	// When dataCleanableCondition is false, it means the associated node is not ready, missing or evicted (check updateDataCleanableCondition()).
-	// In this case, we can delete the orphan directly because the data is not reachable and no need to keep the orphan resource.
-	canDelete := autoDeleteAllowed || dataCleanableCondition.Status == longhorn.ConditionStatusFalse
+	canDeleteWithoutAutoDeletion := false
+	if orphan.Spec.NodeID == node.Name && orphan.Spec.Type == longhorn.OrphanTypeReplicaData {
+		canDeleteCR, reason := nc.canDeleteOrphanDataFromNodeState(orphan, node)
```
derekbit: `canDeleteOrphanDataFromNodeState` is a bit confusing. Is it better to rename it to   `canDeleteOrphanCRFromNodeState`?

controller/node_controller.go
```diff
@@ -1524,34 +1533,73 @@ func (nc *NodeController) deleteOrphansForReplicaDataStore(node *longhorn.Node,
...
+		if canDeleteCR {
+			nc.logger.Infof("Deletable orphan data CR %v: %v", orphan.Name, reason)
```
```
nc.logger.Infof("Deletable orphan data CR %v: %v", orphan.Name, reason)
```

Is it orphan data or orphan CR?

controller/node_controller.go
```diff
@@ -1497,9 +1497,18 @@ func (nc *NodeController) deleteOrphansForReplicaDataStore(node *longhorn.Node,
...
+	for deletionType, enabled := range autoDeletionTypes {
+		if !enabled {
+			continue
+		}
+		switch deletionType {
+		case types.OrphanResourceTypeReplicaData:
+			autoDeleteEnabledTypes[longhorn.OrphanTypeReplicaData] = true
+		case types.OrphanResourceTypeInstance:
+			autoDeleteEnabledTypes[longhorn.OrphanTypeEngineInstance] = true
+			autoDeleteEnabledTypes[longhorn.OrphanTypeReplicaInstance] = true
```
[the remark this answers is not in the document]
COLDTURNIP: To eliminate the confusion, and the Node controller handles only the data Orphans, we remove the type check and focus on one orphan type.

derekbit:
cc @carterli0407-cell for reviewing the PR.

controller/node_controller.go
```diff
@@ -1524,34 +1519,73 @@ func (nc *NodeController) deleteOrphansForReplicaDataStore(node *longhorn.Node,
...
-func (nc *NodeController) canDeleteOrphan(orphan *longhorn.Orphan, autoDeleteEnabled bool, autoDeleteGracePeriod int64) bool {
+func (nc *NodeController) canDeleteOrphan(orphan *longhorn.Orphan, node *longhorn.Node, isAutoDeleteDataOrphanEnabled bool, autoDeleteGracePeriod int64) bool {
 	if orphan.Status.OwnerID != nc.controllerID {
 		return false
 	}
 
-	dataCleanableCondition := types.GetCondition(orphan.Status.Conditions, longhorn.OrphanConditionTypeDataCleanable)
-	if dataCleanableCondition.Status == longhorn.ConditionStatusUnknown {
-		return false
-	}
```
derekbit, line 1535: Why do we remove this logic?

COLDTURNIP: Accidentally removed. Just added them back.

controller/orphan_controller.go
```diff
@@ -278,6 +278,23 @@ func (oc *OrphanController) reconcile(orphanName string) (err error) {
...
+		}
+
+		// Persist the refreshed status now, before any further Update or finalizer
+		// removal. RemoveFinalizerForOrphan issues a metadata Update that bumps
+		// resourceVersion, so the deferred UpdateOrphanStatus would conflict if it
+		// still held the old resourceVersion. Persisting first and refreshing both
+		// orphan and existingOrphan makes the deferred diff a no-op on the
+		// happy path and ensures RemoveFinalizerForOrphan uses a fresh object.
+		updatedOrphan, err := oc.ds.UpdateOrphanStatus(orphan)
+		if err != nil {
```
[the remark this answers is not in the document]
COLDTURNIP: This suggestion is invalid. When we failed to update the status of a CR, it is likely that the API server is broken, and we won't have a chance to revert the status. We should raise the error immediately to trigger another reconcilation.

controller/node_controller_test.go
```diff
@@ -913,9 +912,7 @@ func (s *NodeControllerSuite) TestDisableDiskOnFilesystemChange(c *C) {
 				},
 			},
 		},
-		orphans: map[string]*longhorn.Orphan{
-			DefaultOrphanTestNode1.Name: DefaultOrphanTestNode1,
-		},
+		orphans: map[string]*longhorn.Orphan{},
```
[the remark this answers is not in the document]
COLDTURNIP: Invalid suggestion. In this PR, we have to trigger Orphan CR deletion when the node or disk is leaving the cluster. Deleting an Orphan CR doesn't mean to remove the underlying resources.

controller/node_controller.go
```diff
@@ -1535,23 +1530,67 @@ func (nc *NodeController) canDeleteOrphan(orphan *longhorn.Orphan, autoDeleteEna
...
 	}
 
 	return canDelete
 }
 
+func (nc *NodeController) canDeleteOrphanCRFromNodeState(orphan *longhorn.Orphan, node *longhorn.Node) (bool, string) {
+	nodeReadyCondition := types.GetCondition(node.Status.Conditions, longhorn.NodeConditionTypeReady)
+	switch nodeReadyCondition.Reason {
+	case longhorn.NodeConditionReasonKubernetesNodeGone, longhorn.NodeConditionReasonKubernetesNodeNotReady, longhorn.NodeConditionReasonManagerPodMissing:
+		return true, nodeReadyCondition.Message
```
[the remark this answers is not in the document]
COLDTURNIP: The condition reason is always synced with the status.

derekbit:
IMHO, the title of the PR, `fix(orphan): cr-level data orphan deletion for node or disk eviction`, is not easy to understand. The purpose of the PR is to make orphaned data not deleted and only orphan CRs are cleaned up when `orphan-auto-deletion` is set to `false` and node or disk is deleted, correct? Could we change it to a clear one?

derekbit:
Some thoughts 
https://github.com/longhorn/longhorn/issues/7795#issuecomment-4222864538

controller/node_controller.go
```diff
@@ -1524,34 +1519,81 @@ func (nc *NodeController) deleteOrphansForReplicaDataStore(node *longhorn.Node,
...
+	}
```
davidcheng0922, line 1564: Question here, according to the [ticket 7795](https://github.com/longhorn/longhorn/issues/7795), 
> We shouldn't clean up orphan CR when its node is down, missing manager, or Longhorn manager cannot [get disk stat](https://github.com/longhorn/longhorn-manager/blob/c4e7942684cc1f8ece900854d09126a7b1f8c0b6/util/util.go#L434-L435)

when `orphan-auto-deletion=false`, we should retain the orphan CR when the node is down, the manager pod is missing, or Longhorn cannot get disk stats. 

But it returns true for `NodeConditionReasonKubernetesNodeNotReady` and `NodeConditionReasonManagerPodMissing`. Am I missing something here?

controller/node_controller.go
```diff
@@ -1524,34 +1519,81 @@ func (nc *NodeController) deleteOrphansForReplicaDataStore(node *longhorn.Node,
...
-	canDelete := autoDeleteAllowed || dataCleanableCondition.Status == longhorn.ConditionStatusFalse
+	canDeleteWithoutAutoDeletion := false
+	if orphan.Spec.NodeID == node.Name && orphan.Spec.Type == longhorn.OrphanTypeReplicaData {
+		canDeleteCR, reason := nc.canDeleteOrphanCRFromNodeState(orphan, node)
+		if canDeleteCR {
+			nc.logger.Infof("Deletable orphan CR %v: %v", orphan.Name, reason)
+			canDeleteWithoutAutoDeletion = true
+		}
+	}
+	canDelete := autoDeleteAllowed || canDeleteWithoutAutoDeletion
```
davidcheng0922, line 1551: Orphan CR deletion is only allowed when `orphan.Spec.NodeID == node.Name`. IIRC, from the current controller flow, once the original node becomes deleted or unavailable, orphan ownership can be transferred to another controller. 

The new controller only runs `syncOrphans()` for its own node, not for the original node that owned the orphan. In that case, this condition seems to prevent the orphan CR from ever being deleted. If so, it seems the orphan CR would never be deleted.

PR #4586, opened 2026-03-11
feat: add distro in check upgrade request

davidcheng0922:

- https://github.com/longhorn/longhorn/issues/12778
- https://github.com/rancher/suse-storage-mgmt/issues/116

- Add `distro` information in `CheckUpgradeRequest` and forward to `upgrade-responder` server

### DIFF
--- a/controller/setting_controller.go
+++ b/controller/setting_controller.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"os"
 	"reflect"
 	"slices"
 	"sort"
@@ -48,6 +49,8 @@ import (
 const (
 	VersionTagLatest = "latest"
 	VersionTagStable = "stable"
+
+	DefaultEnvDistroUnknown = "unknown"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


controller/setting_controller.go
```diff
@@ -50,6 +51,15 @@ const (
 	VersionTagStable = "stable"
 )
 
+type DistroType string
+
+const (
```
l-mb: Similar to the comment on the upgrade-responder side: this requires a code change whenever someone wants to build a "new" flavor (e.g., "rgs" or another partner build).

Any other will be silently swallowed to an empty string.

Perhaps just sanitize the string from the environment variable and pass it along as-is?

derekbit:
@davidcheng0922 Could you create a ticket for the feature in https://github.com/longhorn/longhorn/issues? Thank you.

controller/setting_controller.go
```diff
@@ -2000,6 +2009,15 @@ func (info *ClusterInfo) collectNodeScope() {
 	}
 }
 
+func (info *ClusterInfo) collectLonghornDistro() {
+	distro := strings.TrimSpace(os.Getenv(types.EnvDistro))
+	if distro == "" {
```
l-mb, line 2012: If the environment variable isn't set, this is likely a `dev` build - maybe we should use that as the sentinel variable, or "unknown"?

l-mb (changes requested):

controller/setting_controller.go
davidcheng0922: default to `unknown`

derekbit (approved):
LGTM

derekbit:
@mergify backport v1.11.x v1.10.x v1.9.x

PR #4595, opened 2026-03-18
feat: make storage capacity tracking configurable

bachmanity1:

Issue https://github.com/longhorn/longhorn/issues/12807

The new setting `csi-storage-capacity-tracking` controls the granularity of CSI storage capacity tracking.
                                                      
- `disabled`: Capacity tracking is turned off. `StorageCapacity: false` on the CSIDriver object.                              
- `node`: Capacity is reported per node. This is the default and matches the current behavior.                                                   
- `zone`: Capacity is reported per zone.

### DIFF
--- a/app/driver.go
+++ b/app/driver.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"strconv"
 
 	"github.com/cockroachdb/errors"
 	"github.com/sirupsen/logrus"
@@ -287,6 +288,15 @@ func deployCSIDriver(kubeClient *clientset.Clientset, lhClient *lhclientset.Clie
 		return err
 	}
 
+	storageCapacityTrackingSetting, err := lhClient.LonghornV1beta2().Settings(namespace).Get(context.TODO(), string(types.SettingNameCSIStorageCapacityTracking), metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+	storageCapacityTracking, err := strconv.ParseBool(storageCapacityTrackingSetting.Value)
+	if err != nil {
+		return err
+	}
+
 	var imagePullPolicy corev1.PullPolicy
 	switch imagePullPolicySetting.Value {
 	case string(types.SystemManagedPodsImagePullPolicyNever):
@@ -319,7 +329,7 @@ func deployCSIDriver(kubeClient *clientset.Clientset, lhClient *lhclientset.Clie
 		return err
 	}
 
-	csiDriverObjectDeployment := csi.NewCSIDriverObject()
+	csiDriverObjectDeployment := csi.NewCSIDriverObject(storageCapacityTracking)
 	if err := csiDriverObjectDeployment.Deploy(kubeClient); err != nil {
 		return err
 	}
--- a/app/driver.go
+++ b/app/driver.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"strconv"
 
 	"github.com/cockroachdb/errors"
 	"github.com/sirupsen/logrus"
@@ -287,11 +288,13 @@ func deployCSIDriver(kubeClient *clientset.Clientset, lhClient *lhclientset.Clie
 		return err
 	}
 
+	storageCapacityEnabled := false
 	storageCapacityTrackingSetting, err := lhClient.LonghornV1beta2().Settings(namespace).Get(context.TODO(), string(types.SettingNameCSIStorageCapacityTracking), metav1.GetOptions{})
 	if err != nil {
+		logrus.WithError(err).Warnf("failed to get storage capacity tracking setting, defaulting to false")
+	} else if storageCapacityEnabled, err = strconv.ParseBool(storageCapacityTrackingSetting.Value); err != nil {
 		return err
 	}
-	storageCapacityEnabled := types.IsCSIStorageCapacityEnabled(storageCapacityTrackingSetting.Value)
 
 	var imagePullPolicy corev1.PullPolicy
 	switch imagePullPolicySetting.Value {
--- a/app/driver.go
+++ b/app/driver.go
@@ -5,7 +5,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
-	"strconv"
 
 	"github.com/cockroachdb/errors"
 	"github.com/sirupsen/logrus"
@@ -292,10 +291,7 @@ func deployCSIDriver(kubeClient *clientset.Clientset, lhClient *lhclientset.Clie
 	if err != nil {
 		return err
 	}
-	storageCapacityTracking, err := strconv.ParseBool(storageCapacityTrackingSetting.Value)
-	if err != nil {
-		return err
-	}
+	storageCapacityEnabled := types.IsCSIStorageCapacityEnabled(storageCapacityTrackingSetting.Value)
 
 	var imagePullPolicy corev1.PullPolicy
 	switch imagePullPolicySetting.Value {
@@ -329,7 +325,7 @@ func deployCSIDriver(kubeClient *clientset.Clientset, lhClient *lhclientset.Clie
 		return err
 	}
 
-	csiDriverObjectDeployment := csi.NewCSIDriverObject(storageCapacityTracking)
+	csiDriverObjectDeployment := csi.NewCSIDriverObject(storageCapacityEnabled)
 	if err := csiDriverObjectDeployment.Deploy(kubeClient); err != nil {
 		return err
 	}
--- a/controller/setting_controller.go
+++ b/controller/setting_controller.go
@@ -23,6 +23,7 @@ import (
 	storagev1 "k8s.io/api/storage/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	k8stypes "k8s.io/apimachinery/pkg/types"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	clientset "k8s.io/client-go/kubernetes"
 	v1core "k8s.io/client-go/kubernetes/typed/core/v1"
@@ -261,7 +262,7 @@ func (sc *SettingController) syncNonDangerZoneSettingsForManagedComponents(setti
 			return err
 		}
 	case types.SettingNameCSIStorageCapacityTracking:
-		if err := sc.updateCSIStorageCapacityTracking(); err != nil {
+		if err := sc.syncCSIDriverStorageCapacity(); err != nil {
 			return err
 		}
 	}
@@ -1086,10 +1087,32 @@ func (sc *SettingController) updateNodeSelector() error {
 	return nil
 }
 
-func (sc *SettingController) updateCSIStorageCapacityTracking() error {
-	storageCapacity, err := sc.ds.GetSettingAsBool(types.SettingNameCSIStorageCapacityTracking)
+// syncCSIDriverStorageCapacity syncs the CSIDriver StorageCapacity field and restarts
+// longhorn-csi-plugin based on the capacity tracking setting.
+func (sc *SettingController) syncCSIDriverStorageCapacity() error {
+	setting, err := sc.ds.GetSetting(types.SettingNameCSIStorageCapacityTracking)
 	if err != nil {
 		return err
+	}
+	storageCapacityEnabled := types.IsCSIStorageCapacityEnabled(setting.Value)
+	// Restart longhorn-csi-plugin to update CSINode topology keys, which must reflect the new
+	// capacity tracking mode for capacity reporting to function correctly.
+	// Use the setting value as the annotation to avoid restart on every setting resync.
+	if storageCapacityEnabled {
+		ds, err := sc.ds.GetDaemonSet(types.CSIPluginName)
+		if err != nil {
+			return errors.Wrapf(err, "failed to get %v daemonset", types.CSIPluginName)
+		}
+		if ds.Spec.Template.Annotations[types.AnnotationCSIStorageCapacityTracking] != setting.Value {
+			restartPatch := fmt.Sprintf(`{"spec":{"template":{"metadata":{"annotations":{%q:%q}}}}}`,
+				types.AnnotationCSIStorageCapacityTracking, setting.Value)
+			if _, err := sc.kubeClient.AppsV1().DaemonSets(sc.namespace).Patch(
+				context.TODO(), types.CSIPluginName, k8stypes.MergePatchType, []byte(restartPatch), metav1.PatchOptions{},
+			); err != nil {
+				return errors.Wrapf(err, "failed to restart %v daemonset", types.CSIPluginName)
+			}
+			sc.logger.Infof("Restarted %v daemonset to apply new CSI topology keys (mode: %s)", types.CSIPluginName, setting.Value)
+		}
 	}
 
 	csiDriver, err := sc.ds.GetCSIDriver(types.LonghornDriverName)
@@ -1097,17 +1120,17 @@ func (sc *SettingController) updateCSIStorageCapacityTracking() error {
 		return err
 	}
 
-	if csiDriver.Spec.StorageCapacity != nil && *csiDriver.Spec.StorageCapacity == storageCapacity {
+	if csiDriver.Spec.StorageCapacity != nil && *csiDriver.Spec.StorageCapacity == storageCapacityEnabled {
 		return nil
 	}
 
-	csiDriver.Spec.StorageCapacity = ptr.To(storageCapacity)
+	csiDriver.Spec.StorageCapacity = ptr.To(storageCapacityEnabled)
 	// NOTE: StorageCapacity became mutable in Kubernetes 1.23+. On older versions, this update will fail.
 	if _, err := sc.ds.UpdateCSIDriver(csiDriver); err != nil {
 		return err
 	}
+	sc.logger.Infof("Updated CSI driver StorageCapacity to %v (value: %s)", storageCapacityEnabled, setting.Value)
 
-	sc.logger.Infof("Updated CSI driver StorageCapacity to %v", storageCapacity)
 	return nil
 }
 
--- a/controller/setting_controller.go
+++ b/controller/setting_controller.go
@@ -23,7 +23,6 @@ import (
 	storagev1 "k8s.io/api/storage/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	k8stypes "k8s.io/apimachinery/pkg/types"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	clientset "k8s.io/client-go/kubernetes"
 	v1core "k8s.io/client-go/kubernetes/typed/core/v1"
@@ -1087,49 +1086,26 @@ func (sc *SettingController) updateNodeSelector() error {
 	return nil
 }
 
-// syncCSIDriverStorageCapacity syncs the CSIDriver StorageCapacity field and restarts
-// longhorn-csi-plugin based on the capacity tracking setting.
+// syncCSIDriverStorageCapacity syncs the CSIDriver StorageCapacity field
 func (sc *SettingController) syncCSIDriverStorageCapacity() error {
-	setting, err := sc.ds.GetSetting(types.SettingNameCSIStorageCapacityTracking)
+	storageCapacityEnabled, err := sc.ds.GetSettingAsBool(types.SettingNameCSIStorageCapacityTracking)
 	if err != nil {
 		return err
-	}
-	storageCapacityEnabled := types.IsCSIStorageCapacityEnabled(setting.Value)
-	// Restart longhorn-csi-plugin to update CSINode topology keys, which must reflect the new
-	// capacity tracking mode for capacity reporting to function correctly.
-	// Use the setting value as the annotation to avoid restart on every setting resync.
-	if storageCapacityEnabled {
-		ds, err := sc.ds.GetDaemonSet(types.CSIPluginName)
-		if err != nil {
-			return errors.Wrapf(err, "failed to get %v daemonset", types.CSIPluginName)
-		}
-		if ds.Spec.Template.Annotations[types.AnnotationCSIStorageCapacityTracking] != setting.Value {
-			restartPatch := fmt.Sprintf(`{"spec":{"template":{"metadata":{"annotations":{%q:%q}}}}}`,
-				types.AnnotationCSIStorageCapacityTracking, setting.Value)
-			if _, err := sc.kubeClient.AppsV1().DaemonSets(sc.namespace).Patch(
-				context.TODO(), types.CSIPluginName, k8stypes.MergePatchType, []byte(restartPatch), metav1.PatchOptions{},
-			); err != nil {
-				return errors.Wrapf(err, "failed to restart %v daemonset", types.CSIPluginName)
-			}
-			sc.logger.Infof("Restarted %v daemonset to apply new CSI topology keys (mode: %s)", types.CSIPluginName, setting.Value)
-		}
 	}
 
 	csiDriver, err := sc.ds.GetCSIDriver(types.LonghornDriverName)
 	if err != nil {
 		return err
 	}
-
 	if csiDriver.Spec.StorageCapacity != nil && *csiDriver.Spec.StorageCapacity == storageCapacityEnabled {
 		return nil
 	}
 
 	csiDriver.Spec.StorageCapacity = ptr.To(storageCapacityEnabled)
-	// NOTE: StorageCapacity became mutable in Kubernetes 1.23+. On older versions, this update will fail.
 	if _, err := sc.ds.UpdateCSIDriver(csiDriver); err != nil {
 		return err
 	}
-	sc.logger.Infof("Updated CSI driver StorageCapacity to %v (value: %s)", storageCapacityEnabled, setting.Value)
+	sc.logger.Infof("Updated CSI driver StorageCapacity to %v", storageCapacityEnabled)
 
 	return nil
 }
--- a/controller/setting_controller.go
+++ b/controller/setting_controller.go
@@ -35,6 +35,7 @@ import (
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/kubernetes/pkg/controller"
+	"k8s.io/utils/ptr"
 
 	lhns "github.com/longhorn/go-common-libs/ns"
 
@@ -257,6 +258,10 @@ func (sc *SettingController) syncNonDangerZoneSettingsForManagedComponents(setti
 		}
 	case types.SettingNameDefaultLonghornStaticStorageClass:
 		if err := sc.syncDefaultLonghornStaticStorageClass(); err != nil {
+			return err
+		}
+	case types.SettingNameCSIStorageCapacityTracking:
+		if err := sc.updateCSIStorageCapacityTracking(); err != nil {
 			return err
 		}
 	}
@@ -1078,6 +1083,31 @@ func (sc *SettingController) updateNodeSelector() error {
 		}
 	}
 
+	return nil
+}
+
+func (sc *SettingController) updateCSIStorageCapacityTracking() error {
+	storageCapacity, err := sc.ds.GetSettingAsBool(types.SettingNameCSIStorageCapacityTracking)
+	if err != nil {
+		return err
+	}
+
+	csiDriver, err := sc.ds.GetCSIDriver(types.LonghornDriverName)
+	if err != nil {
+		return err
+	}
+
+	if csiDriver.Spec.StorageCapacity != nil && *csiDriver.Spec.StorageCapacity == storageCapacity {
+		return nil
+	}
+
+	csiDriver.Spec.StorageCapacity = ptr.To(storageCapacity)
+	// NOTE: StorageCapacity became mutable in Kubernetes 1.23+. On older versions, this update will fail.
+	if _, err := sc.ds.UpdateCSIDriver(csiDriver); err != nil {
+		return err
+	}
+
+	sc.logger.Infof("Updated CSI driver StorageCapacity to %v", storageCapacity)
 	return nil
 }
 
--- a/csi/controller_server.go
+++ b/csi/controller_server.go
@@ -23,10 +23,9 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
-	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/client-go/rest"
-	"k8s.io/client-go/tools/cache"
 
 	"github.com/longhorn/longhorn-manager/datastore"
 	"github.com/longhorn/longhorn-manager/types"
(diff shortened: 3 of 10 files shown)


csi/node_server.go
```diff
@@ -935,10 +935,10 @@ func (ns *NodeServer) NodeGetInfo(ctx context.Context, req *csi.NodeGetInfoReque
...
 	kubeNode, err := ns.kubeClient.CoreV1().Nodes().Get(ctx, ns.nodeID, metav1.GetOptions{})
 	if err != nil {
-		ns.log.WithError(err).Warnf("Failed to get Kubernetes node %s for topology labels, using hostname only", ns.nodeID)
+		ns.log.WithError(err).Warnf("Failed to get Kubernetes node %s for topology labels", ns.nodeID)
 	} else {
 		for key, value := range kubeNode.Labels {
-			if key == nodeTopologyKey || allowedKeys[key] {
+			if allowedKeys[key] {
 				topologySegments[key] = value
 			}
```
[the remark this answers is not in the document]
bachmanity1: Validating webhook doesn't allow to update to `zone` mode when `csi-allowed-topology-keys` contains hostname key.

datastore/longhorn.go
```diff
@@ -672,6 +672,20 @@ func (s *DataStore) ValidateSetting(name, value string) (err error) {
...
+				return err
+			}
+			allowedKeys := types.ParseCSIAllowedTopologyKeys(allowedKeysSetting.Value)
+			if !allowedKeys[corev1.LabelTopologyZone] {
+				return fmt.Errorf("zone-level capacity tracking requires %q to include %q", types.SettingNameCSIAllowedTopologyKeys, corev1.LabelTopologyZone)
+			}
+			if allowedKeys[corev1.LabelHostname] {
+				return fmt.Errorf("zone-level capacity tracking requires %q to exclude %q", types.SettingNameCSIAllowedTopologyKeys, corev1.LabelHostname)
+			}
+		}
```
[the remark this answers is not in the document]
bachmanity1: Updates to `csi-allowed-topology-keys` don't take effect until `longhorn-csi-plugin` daemonset is restarted. So even if this setting is updated it shouldn't affect capacity tracking unless plugin daemonset is restarted.

csi/controller_server.go
```diff
@@ -700,111 +702,134 @@ func (cs *ControllerServer) ListVolumes(context.Context, *csi.ListVolumesRequest
...
-	allowEmptyDiskSelectorVolume, err := cs.getSettingAsBoolean(ctx, types.SettingNameAllowEmptyDiskSelectorVolume)
-	if err != nil {
-		return nil, status.Errorf(codes.Internal, "failed to get setting %v: %v", types.SettingNameAllowEmptyDiskSelectorVolume, err)
-	}
-	overProvisioningPercentage, err := cs.getSettingAsInt(ctx, types.SettingNameStorageOverProvisioningPercentage)
-	if err != nil {
-		return nil, status.Errorf(codes.Internal, "failed to get setting %v: %v", types.SettingNameStorageOverProvisioningPercentage, err)
+	dataEngine := longhorn.DataEngineTypeV1
+	if dataEngineType, ok := scParameters["dataEngine"]; ok {
+		dataEngine = longhorn.DataEngineType(dataEngineType)
```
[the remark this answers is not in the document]
bachmanity1: I think it's ok to return 0 when data engine value is not valid, this would result in pods using this storage class not being able to get scheduled.

csi/controller_server_test.go
```diff
@@ -276,65 +278,118 @@ func TestGetCapacity(t *testing.T) {
 			} else if expectedStatus.Message() != actualStatus.Message() {
 				t.Errorf("expected error message: '%s', but got: '%s'", expectedStatus.Message(), actualStatus.Message())
 			}
-			if res != nil && res.AvailableCapacity != test.availableCapacity {
-				t.Errorf("expected available capacity: %d, but got: %d", test.availableCapacity, res.AvailableCapacity)
+			if res != nil && res.MaximumVolumeSize.Value != test.maximumVolumeSize {
+				t.Errorf("expected maximum volume size: %d, but got: %d", test.maximumVolumeSize, res.AvailableCapacity)
```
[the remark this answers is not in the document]
bachmanity1: Fixed.

csi/node_server.go
```diff
@@ -953,24 +956,41 @@ func (ns *NodeServer) NodeGetInfo(ctx context.Context, req *csi.NodeGetInfoReque
...
+// ensureTopologyKey adds the given label key to topologySegments if not already present, and warns if the label is missing.
+func (ns *NodeServer) ensureTopologyKey(topologySegments, nodeLabels map[string]string, key string) {
+	labelVal, hasLabel := nodeLabels[key]
+	if !hasLabel {
+		ns.log.Errorf("Node is missing label %q, capacity tracking may not work correctly", key)
+	}
+	_, inTopology := topologySegments[key]
+	if !inTopology {
+		topologySegments[key] = labelVal
+		ns.log.Infof("Added label %q to CSINode topology keys", key)
```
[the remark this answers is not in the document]
bachmanity1: fixed.

csi/controller_server.go
```diff
@@ -700,111 +702,146 @@ func (cs *ControllerServer) ListVolumes(context.Context, *csi.ListVolumesRequest
...
+		if err != nil {
+			return nil, status.Errorf(codes.Internal, "failed to list nodes: %v", err)
+		}
+		for _, node := range nodeList.Items {
+			if node.Status.Zone == zone {
+				nodeMaximumVolumeSize, nodeAvailableCapacity := nodeCapacity(&node)
+				maximumVolumeSize = max(maximumVolumeSize, nodeMaximumVolumeSize)
+				availableCapacity += nodeAvailableCapacity
+			}
+		}
```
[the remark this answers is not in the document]
bachmanity1: To reduce load on api-server refactored code to use informers.

csi/controller_server.go
```diff
@@ -700,111 +702,146 @@ func (cs *ControllerServer) ListVolumes(context.Context, *csi.ListVolumesRequest
...
-	case longhorn.DataEngineTypeV1:
-		rsp.AvailableCapacity = v1AvailableCapacity
-	case longhorn.DataEngineTypeV2:
-		rsp.AvailableCapacity = v2AvailableCapacity
+	switch types.CSIStorageCapacityTracking(storageCapacityTrackingSetting) {
+	case types.CSIStorageCapacityTrackingNode:
+		nodeID, ok := topology.Segments[corev1.LabelHostname]
+		if !ok {
+			return nil, status.Errorf(codes.InvalidArgument, "missing %q topology segment", corev1.LabelHostname)
+		}
```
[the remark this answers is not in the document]
bachmanity1: When mode is `disabled` csi-provisioner won't call GetCapacity rpc because csidriver's storagecapacity field is set to false. Because of some timing issue this may occur in some very rare situations in which case controller server will respond with error.

csi/controller_server_test.go
```diff
@@ -276,65 +282,119 @@ func TestGetCapacity(t *testing.T) {
 			} else if expectedStatus.Message() != actualStatus.Message() {
 				t.Errorf("expected error message: '%s', but got: '%s'", expectedStatus.Message(), actualStatus.Message())
 			}
-			if res != nil && res.AvailableCapacity != test.availableCapacity {
-				t.Errorf("expected available capacity: %d, but got: %d", test.availableCapacity, res.AvailableCapacity)
+			if res != nil && res.MaximumVolumeSize.Value != test.maximumVolumeSize {
+				t.Errorf("expected maximum volume size: %d, but got: %d", test.maximumVolumeSize, res.MaximumVolumeSize.Value)
```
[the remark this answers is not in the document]
bachmanity1: I've refactored the code so that GetCapacity now returns MaximumVolumeSize, which is preferred over AvailableCapacity according to the KEP-1472 spec.

AvailableCapacity now returns the total capacity on a node by summing schedulable storage across all disks on the node, but this field is not really used by the kube-scheduler unless the user opts in to use KEP-4049, which sorts nodes by remaining capacity.

bachmanity1:
Hi @derekbit. This PR is now ready for review, please have a look when you have time. Thanks!

hookak:
I’m not looking at this from a detailed code review perspective, but rather from a broader design point of view.

If Longhorn is configured to expose topology at either the node level or the zone level globally, it may be difficult to support both of these cases within a single Longhorn cluster:

* using StorageClass A for node-level local replicas, and
* using StorageClass B for zone-level remote replicas.

It seems like the key part here is the label exposed through `NodeGetInfo`.
So I was wondering whether it would make sense to implement this in a way that allows the administrator to decide, based on node labels or similar configuration, whether a given node should be exposed at the zone level or at the host level.

With that approach, it seems like both use cases could be handled more flexibly. What do you think?

bachmanity1:
> I’m not looking at this from a detailed code review perspective, but rather from a broader design point of view.
> 
> If Longhorn is configured to expose topology at either the node level or the zone level globally, it may be difficult to support both of these cases within a single Longhorn cluster:
> 
> * using StorageClass A for node-level local replicas, and
> * using StorageClass B for zone-level remote replicas.
> 
> It seems like the key part here is the label exposed through `NodeGetInfo`. So I was wondering whether it would make sense to implement this in a way that allows the administrator to decide, based on node labels or similar configuration, whether a given node should be exposed at the zone level or at the host level.
> 
> With that approach, it seems like both use cases could be handled more flexibly. What do you think?

I've also considered this, but I felt it might be too complex for users if we try to support both node-level and zone-level capacity reporting within the same cluster.

If this is really needed, I think we could simply expose both zone and node topology keys on all CSINodes. This would allow us to support both node-level and zone-level capacity reporting within the same cluster.

We can introduce a new storage class parameter, something like `storageCapacity`, with possible values like `node` and `zone`. If it’s not specified, we can assume it defaults to `node`. Then, based on this new parameter `GetCapacity` can decide which level of capacity reporting the given storage class requires.

In the current implementation, when zone-level reporting is enabled, each CSINode exposes only the zone key. As a result, `GetCapacity` is called only once per zone, and the number of `CSIStorageCapacity` objects equals the number of zones. However, if each CSINode exposes both node and zone keys, then the number of `CSIStorageCapacity` objects will equal the number of nodes, because `GetCapacity` is called for each node. This is also somewhat inefficient because nodes within the same zone perform the same computation and return the same value.

derekbit:
Thanks @bachmanity1. Sorry for late reply. We will start reviewing LEP.

COLDTURNIP (approved):

types/setting.go
```diff
@@ -1921,14 +1924,33 @@ var (
...
+		Type:               SettingTypeString,
+		Required:           true,
+		ReadOnly:           false,
+		DataEngineSpecific: false,
+		Default:            string(CSIStorageCapacityTrackingNode),
+		Choices: []any{
+			string(CSIStorageCapacityTrackingDisabled),
+			string(CSIStorageCapacityTrackingNode),
+			string(CSIStorageCapacityTrackingZone),
+		},
```
derekbit, line 2032: @sushant-suse Please review the description. Thank you.

controller/setting_controller.go
```diff
@@ -1021,6 +1027,53 @@ func (sc *SettingController) updateNodeSelector() error {
...
+
+	if csiDriver.Spec.StorageCapacity != nil && *csiDriver.Spec.StorageCapacity == storageCapacityEnabled {
+		return nil
+	}
+
+	csiDriver.Spec.StorageCapacity = ptr.To(storageCapacityEnabled)
+	// NOTE: StorageCapacity became mutable in Kubernetes 1.23+. On older versions, this update will fail.
+	if _, err := sc.ds.UpdateCSIDriver(csiDriver); err != nil {
+		return err
+	}
```
derekbit, line 1131: This is fine, because the requirement of kubernetes for the newer Longhorn is v1.25+.
https://longhorn.io/docs/1.11.1/important-notes/#kubernetes-version-requirement

csi/controller_server.go
```diff
@@ -54,13 +59,13 @@ const (
...
-	log         *logrus.Entry
-	lhClient    lhclientset.Interface
-	lhNamespace string
+	apiClient     *longhornclient.RancherClient
+	nodeID        string
+	caps          []*csi.ControllerServiceCapability
+	accessModes   []*csi.VolumeCapability_AccessMode
+	log           *logrus.Entry
+	settingLister lhlisters.SettingNamespaceLister
+	nodeLister    lhlisters.NodeNamespaceLister
```
derekbit, line 68: Why not use the original `lhClient` and `lhNamespace`?

types/setting.go
```diff
@@ -1921,14 +1924,33 @@ var (
...
 		ReadOnly:           false,
 		DataEngineSpecific: false,
 		Default:            "",
 	}
+
+	SettingDefinitionCSIStorageCapacityTracking = SettingDefinition{
+		DisplayName: "CSI Storage Capacity Tracking",
+		Description: "Controls CSI storage capacity tracking (KEP-1472), which allows the kube-scheduler to filter " +
+			"nodes that cannot fit the requested volume. " +
+			"Possible values are \"disabled\" (tracking off), \"node\" (per-node capacity), and \"zone\" (per-zone capacity).",
```
sushant-suse, line 2021:
```suggestion
		Description: "Enables CSI storage capacity tracking (KEP-1472) to improve pod scheduling by filtering out nodes with insufficient storage. " +
			"Options: \"disabled\" (tracking is turned off), \"node\" (tracks capacity on a per-node basis), and \"zone\" (tracks capacity across availability zones).",
```

sushant-suse (changes requested):

types/setting.go
sushant-suse: Done thanks.

csi/controller_server.go
```diff
@@ -79,6 +84,19 @@ func NewControllerServer(apiClient *longhornclient.RancherClient, nodeID string)
...
+	informerFactory := lhinformers.NewSharedInformerFactoryWithOptions(lhClient, 0, lhinformers.WithNamespace(lhNamespace))
+	settingInformer := informerFactory.Longhorn().V1beta2().Settings()
+	_ = settingInformer.Informer()
+	nodeInformer := informerFactory.Longhorn().V1beta2().Nodes()
+	_ = nodeInformer.Informer()
+
+	// Start informers and wait for cache sync
+	stopCh := make(chan struct{})
+	informerFactory.Start(stopCh)
+	if !cache.WaitForCacheSync(stopCh, settingInformer.Informer().HasSynced, nodeInformer.Informer().HasSynced) {
```
davidcheng0922, line 96: Will this stopCh be closed?

csi/node_server.go
```diff
@@ -923,27 +923,30 @@ func (ns *NodeServer) getEncryptionPassphrase(secrets map[string]string, volumeI
...
 			}
 		}
 	}
 
+	switch ns.getStorageCapacityTracking(ctx) {
+	case types.CSIStorageCapacityTrackingNode:
+		ns.ensureTopologyKey(topologySegments, kubeNode.Labels, corev1.LabelHostname)
+	case types.CSIStorageCapacityTrackingZone:
+		ns.ensureTopologyKey(topologySegments, kubeNode.Labels, corev1.LabelTopologyZone)
+	}
```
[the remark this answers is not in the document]
davidcheng0922: Same question: is there a risk of a panic here, or is kubeNode guaranteed to never be nil?

csi/controller_server.go
```diff
@@ -700,140 +718,167 @@ func (cs *ControllerServer) ListVolumes(context.Context, *csi.ListVolumesRequest
...
+	topology := req.GetAccessibleTopology()
+	if topology == nil || len(topology.Segments) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "missing or empty accessible topology request parameter")
 	}
-	switch dataEngine {
-	case longhorn.DataEngineTypeV1:
-		rsp.AvailableCapacity = v1AvailableCapacity
-	case longhorn.DataEngineTypeV2:
-		rsp.AvailableCapacity = v2AvailableCapacity
+	switch types.CSIStorageCapacityTracking(storageCapacityTrackingSetting) {
```
davidcheng0922, line 814: Missing handle the case `CSIStorageCapacityTrackingDisabled`, it will  be error "unexpected value..."

controller/setting_controller.go
```diff
@@ -1021,6 +1027,53 @@ func (sc *SettingController) updateNodeSelector() error {
...
+			restartPatch := fmt.Sprintf(`{"spec":{"template":{"metadata":{"annotations":{%q:%q}}}}}`,
+				types.AnnotationCSIStorageCapacityTracking, setting.Value)
+			if _, err := sc.kubeClient.AppsV1().DaemonSets(sc.namespace).Patch(
+				context.TODO(), types.CSIPluginName, k8stypes.MergePatchType, []byte(restartPatch), metav1.PatchOptions{},
+			); err != nil {
+				return errors.Wrapf(err, "failed to restart %v daemonset", types.CSIPluginName)
+			}
+			sc.logger.Infof("Restarted %v daemonset to apply new CSI topology keys (mode: %s)", types.CSIPluginName, setting.Value)
+		}
+	}
```
COLDTURNIP, line 1116: We may modify the value of daemonset annotation when the setting is turned off.

COLDTURNIP (changes requested):

controller/setting_controller.go
COLDTURNIP: In addition, in this implementation, the annotation is not attached during daemonset creation. Does it means the CSI plugin daemonset always restarts after initialization?

derekbit:
Hello @bachmanity1 
Feel free to ping me if the PR is ready. If possible, we can backport the setting to the upcoming v1.11.2. Thank you.

derekbit:
cc @bachmanity1

bachmanity1:
Hi @derekbit. Sorry for the late response.                                                                                                      
                                                                 
I've made some changes to the PR. Storage capacity mode (node or zone) is now controlled per StorageClass via StorageClass parameters, allowing granularity to be configured per StorageClass, similar to what @hookak suggested in https://github.com/longhorn/longhorn-manager/pull/4595#issuecomment-4113804222.

csi/controller_server.go
bachmanity1: I think using informers would put less pressure on the kube-apiserver, especially in large clusters with many nodes and StorageClasses.

csi/node_server.go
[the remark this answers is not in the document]
bachmanity1: I included the nil check, but I'm not sure if kubeNode can actually be nil. If it is, there is probably a bigger issue anyway.

controller/setting_controller.go
bachmanity1: This code is removed now. Daemonset is not restarted or touched in any other way.

csi/controller_server.go
bachmanity1: This code has changed now.

types/setting.go
```diff
@@ -2001,14 +2004,27 @@ var (
...
 		Type:               SettingTypeString,
 		Required:           false,
 		ReadOnly:           false,
 		DataEngineSpecific: false,
 		Default:            "",
 	}
+
+	SettingDefinitionCSIStorageCapacityTracking = SettingDefinition{
+		DisplayName: "CSI Storage Capacity Tracking",
+		Description: "Controls CSI storage capacity tracking (KEP-1472), which allows the kube-scheduler to filter " +
```
```suggestion
		Description: "Controls CSI storage capacity tracking, which allows the kube-scheduler to filter " +
```

types/types.go
```diff
@@ -232,6 +232,8 @@ const (
 
 	PVAnnotationLonghornVolumeSchedulingError = "longhorn.io/volume-scheduling-error"
 
+	AnnotationCSIStorageCapacityTracking = "longhorn.io/csi-storage-capacity-tracking"
```
derekbit: Do we need the annotation? It seems unused.

csi/node_server.go
```diff
@@ -923,27 +923,28 @@ func (ns *NodeServer) getEncryptionPassphrase(secrets map[string]string, volumeI
...
-			if key == nodeTopologyKey || allowedKeys[key] {
+			if allowedKeys[key] {
 				topologySegments[key] = value
 			}
 		}
 	}
 
+	if kubeNode != nil {
+		ns.ensureTopologyKey(kubeNode.Name, topologySegments, kubeNode.Labels, corev1.LabelHostname)
+		ns.ensureTopologyKey(kubeNode.Name, topologySegments, kubeNode.Labels, corev1.LabelTopologyZone)
```
derekbit, line 948: `ensureTopologyKey` ensures that important topology keys (zone and hostname) are included in the CSI topology output if they exist on the node. Can we add a comment here to explain the purpose of the function?

csi/deployment.go
```diff
@@ -636,14 +636,14 @@ type DriverObjectDeployment struct {
...
-func NewCSIDriverObject() *DriverObjectDeployment {
+func NewCSIDriverObject(storageCapacity bool) *DriverObjectDeployment {
 	obj := &storagev1.CSIDriver{
 		ObjectMeta: metav1.ObjectMeta{
 			Name: types.LonghornDriverName,
 		},
 		Spec: storagev1.CSIDriverSpec{
 			PodInfoOnMount:  ptr.To(true),
-			StorageCapacity: ptr.To(true),
+			StorageCapacity: ptr.To(storageCapacity),
```
derekbit: nit:
```suggestion
func NewCSIDriverObject(storageCapacityEnabled bool) *DriverObjectDeployment {
	obj := &storagev1.CSIDriver{
		ObjectMeta: metav1.ObjectMeta{
			Name: types.LonghornDriverName,
		},
		Spec: storagev1.CSIDriverSpec{
			PodInfoOnMount:  ptr.To(true),
			StorageCapacity: ptr.To(storageCapacityEnabled),
```

csi/controller_server.go
```diff
@@ -700,140 +718,167 @@ func (cs *ControllerServer) ListVolumes(context.Context, *csi.ListVolumesRequest
...
+			if types.GetCondition(diskStatus.Conditions, longhorn.DiskConditionTypeSchedulable).Status != longhorn.ConditionStatusTrue {
+				continue
+			}
+			if !types.IsSelectorsInTags(diskSpec.Tags, diskSelector, allowEmptyDiskSelectorVolume) {
+				continue
+			}
+
+			overProvisionLimit := ((diskStatus.StorageMaximum - diskSpec.StorageReserved) * overProvisioningPercentage) / 100
+			storageSchedulable := overProvisionLimit - diskStatus.StorageScheduled
+			storageSchedulable = max(storageSchedulable, 0)
```
derekbit, line 794: Is it possible that storageSchedulable is a negative value?

csi/controller_server.go
```diff
@@ -700,140 +718,167 @@ func (cs *ControllerServer) ListVolumes(context.Context, *csi.ListVolumesRequest
...
-	if err != nil {
-		return nil, status.Errorf(codes.Internal, "failed to get setting %v: %v", types.SettingNameAllowEmptyDiskSelectorVolume, err)
+	dataEngine := longhorn.DataEngineTypeV1
+	if dataEngineType, ok := scParameters["dataEngine"]; ok {
+		dataEngine = longhorn.DataEngineType(dataEngineType)
 	}
-	overProvisioningPercentage, err := cs.getSettingAsInt(ctx, types.SettingNameStorageOverProvisioningPercentage)
-	if err != nil {
-		return nil, status.Errorf(codes.Internal, "failed to get setting %v: %v", types.SettingNameStorageOverProvisioningPercentage, err)
+	storageCapacityMode := "node"
```
derekbit: Could we introduce two constants like
StorageCapacityModeNode
StorageCapacityModeZone

csi/controller_server.go
```diff
@@ -700,140 +718,167 @@ func (cs *ControllerServer) ListVolumes(context.Context, *csi.ListVolumesRequest
...
+			overProvisionLimit := ((diskStatus.StorageMaximum - diskSpec.StorageReserved) * overProvisioningPercentage) / 100
+			storageSchedulable := overProvisionLimit - diskStatus.StorageScheduled
+			storageSchedulable = max(storageSchedulable, 0)
+			if dataEngine == longhorn.DataEngineTypeV1 && diskStatus.Type == longhorn.DiskTypeFilesystem {
+				maximumVolumeSize = max(maximumVolumeSize, storageSchedulable)
+				availableCapacity += storageSchedulable
+			} else if dataEngine == longhorn.DataEngineTypeV2 && diskStatus.Type == longhorn.DiskTypeBlock {
+				maximumVolumeSize = max(maximumVolumeSize, storageSchedulable)
+				availableCapacity += storageSchedulable
+			}
```
derekbit: How about?
```suggestion
                validDisk := (dataEngine == longhorn.DataEngineTypeV1 && diskStatus.Type == longhorn.DiskTypeFilesystem) ||
                            (dataEngine == longhorn.DataEngineTypeV2 && diskStatus.Type == longhorn.DiskTypeBlock)

                if validDisk {
                    maximumVolumeSize = max(maximumVolumeSize, storageSchedulable)
                    availableCapacity += storageSchedulable
                }
```

csi/controller_server.go
```diff
@@ -700,140 +718,167 @@ func (cs *ControllerServer) ListVolumes(context.Context, *csi.ListVolumesRequest
...
+			return nil, status.Errorf(codes.InvalidArgument, "missing %q topology segment", corev1.LabelTopologyZone)
+		}
+		nodes, err := cs.nodeLister.List(labels.Everything())
+		if err != nil {
+			return nil, status.Errorf(codes.Internal, "failed to list nodes: %v", err)
+		}
+		for _, node := range nodes {
+			if node.Status.Zone == zone {
+				nodeMaximumVolumeSize, nodeAvailableCapacity := nodeCapacity(node)
+				maximumVolumeSize = max(maximumVolumeSize, nodeMaximumVolumeSize)
```
derekbit, line 840: I'm a bit confused. If `maximumVolumeSize` is 1G and `nodeMaximumVolumeSize` is 5G, the maximum available volume is actually 1G.
Why do we need `maximumVolumeSize = max(maximumVolumeSize, nodeMaximumVolumeSize)`? The `maximumVolumeSize` will be set to 5G.

derekbit:
cc @davidcheng0922 @COLDTURNIP for review. Thank you.

csi/node_server.go
bachmanity1: Added comment.

csi/deployment.go
bachmanity1: Applied your suggestion.

csi/controller_server.go
bachmanity1: It's possible to set `diskSpec.StorageReserved` to be larger than `diskStatus.StorageMaximum` in which case `storageSchedulable` can be negative.

csi/controller_server.go
bachmanity1: Added new constants.

csi/controller_server.go
bachmanity1: In this case, `maximumVolumeSize` is the maximum size of a volume that can be scheduled in the given zone. Say `maximumVolumeSize` is 1Gi before processing node-1. node-1 has disks with capacities of 1Gi, 2Gi, and 5Gi, so the maximum volume size that can be scheduled on node-1 is 5Gi, and `nodeMaximumVolumeSize` will be set to 5Gi. We then update `maximumVolumeSize` to 5Gi because we now know that a volume of that size can be scheduled in this zone.

types/types.go
bachmanity1: We don't need it anymore. I removed it.

app/driver.go
```diff
@@ -287,6 +288,15 @@ func deployCSIDriver(kubeClient *clientset.Clientset, lhClient *lhclientset.Clie
...
 	}
 
+	storageCapacityTrackingSetting, err := lhClient.LonghornV1beta2().Settings(namespace).Get(context.TODO(), string(types.SettingNameCSIStorageCapacityTracking), metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+	storageCapacityEnabled, err := strconv.ParseBool(storageCapacityTrackingSetting.Value)
+	if err != nil {
+		return err
+	}
```
derekbit: How about using `false` by default


```suggestion
	storageCapacityEnabled, err := strconv.ParseBool(storageCapacityTrackingSetting.Value)
	if err != nil {
		logrus.WithError(err).Warnf(...., fall back to ...)
        storageCapacityEnabled = false
	}
```

csi/controller_server.go
derekbit: Got it! Thanks for the explanation!

types/setting.go
```diff
@@ -2001,14 +2004,34 @@ var (
...
+	SettingDefinitionCSIStorageCapacityTracking = SettingDefinition{
+		DisplayName: "CSI Storage Capacity Tracking",
+		Description: "Controls CSI storage capacity tracking, which allows the kube-scheduler to filter " +
+			"nodes that cannot fit the requested volume.",
+		Category:           SettingCategoryGeneral,
+		Type:               SettingTypeBool,
+		Required:           true,
+		ReadOnly:           false,
+		DataEngineSpecific: false,
+		Default:            "true",
```
derekbit: Can we set it to `false` by default?

csi/controller_server.go
COLDTURNIP: I think the performance improvement is another topic. Before this change, we already have client for these two resources, and there is no evidence to indicate that the uncached client is the bottleneck.
If we're replacing the client with the lister, suggest to maintain the lister instance at the manager layer (csi/manager.go), injects into the ControllerServer during initialization, so that we have better scope to handle the lifecycle of this cached client, for example, to wait for cache sync, or the context handling that is addressed by @davidcheng0922 at https://github.com/longhorn/longhorn-manager/pull/4595/changes#r3071576292

derekbit: Agree with @COLDTURNIP. How about creating a new ticket for the local cache and revert back to the original implementation?

app/driver.go
bachmanity1: Applied this suggestion.

types/setting.go
bachmanity1: changed to false by default.

derekbit:
Hello @bachmanity1 
Can you check https://github.com/longhorn/longhorn-manager/pull/4595#discussion_r3122357346? Thank you.

csi/controller_server.go
bachmanity1: I've removed informers and replaced it with lhclient.

derekbit:
The PR is almost there. @bachmanity1 Could you consolidate the commits? We will do the final review. Thanks.

csi/controller_server.go
```diff
@@ -700,111 +702,147 @@ func (cs *ControllerServer) ListVolumes(context.Context, *csi.ListVolumesRequest
...
+		zone, ok := topology.Segments[corev1.LabelTopologyZone]
+		if !ok {
+			return nil, status.Errorf(codes.InvalidArgument, "missing %q topology segment", corev1.LabelTopologyZone)
+		}
+		nodeList, err := cs.lhClient.LonghornV1beta2().Nodes(cs.lhNamespace).List(ctx, metav1.ListOptions{})
+		if err != nil {
+			return nil, status.Errorf(codes.Internal, "failed to list nodes: %v", err)
+		}
+		for i := range nodeList.Items {
+			node := &nodeList.Items[i]
```
COLDTURNIP: nit
```suggestion
		for _, node := range nodeList.Items {
```

COLDTURNIP (dismissed):
I'm good for this change, and there are some nits.

derekbit:
@mergify backport v1.11.x v1.10.x

derekbit (approved):
LGTM. Thanks @bachmanity1 for the contribution!

derekbit:
@mergify backport v1.11.x v1.10.x

PR #4627, opened 2026-04-01
fix(instancemanager,engineimage): honor node selector setting while setup the pod

COLDTURNIP:

Issue longhorn/longhorn#12834

### DIFF
--- a/controller/engine_image_controller.go
+++ b/controller/engine_image_controller.go
@@ -11,6 +11,7 @@ import (
 	"github.com/sirupsen/logrus"
 
 	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/apimachinery/pkg/util/wait"
@@ -303,10 +304,14 @@ func (ic *EngineImageController) syncEngineImage(key string) (err error) {
 	// TODO: Will remove this reference kind correcting after all Longhorn components having used the new kinds
 	if len(ds.OwnerReferences) < 1 || ds.OwnerReferences[0].Kind != types.LonghornKindEngineImage {
 		ds.OwnerReferences = datastore.GetOwnerReferencesForEngineImage(engineImage)
-		_, err = ic.kubeClient.AppsV1().DaemonSets(ic.namespace).Update(context.TODO(), ds, metav1.UpdateOptions{})
+		ds, err = ic.kubeClient.AppsV1().DaemonSets(ic.namespace).Update(context.TODO(), ds, metav1.UpdateOptions{})
 		if err != nil {
 			return err
 		}
+	}
+
+	if err := ic.syncEngineImageDaemonSetNodeSelector(ds); err != nil {
+		return errors.Wrapf(err, "failed to sync node selector for engine image daemonset %v", ds.Name)
 	}
 
 	if err := ic.updateEngineImageRefCount(engineImage); err != nil {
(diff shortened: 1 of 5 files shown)


derekbit:
@COLDTURNIP  Please address the comments from Copilot. Thanks.

COLDTURNIP:
@coderabbitai review

controller/node_controller.go
```diff
@@ -1140,7 +1162,14 @@ func (nc *NodeController) syncInstanceManagers(node *longhorn.Node) error {
...
 				cleanupRequired := true
 
-				if (im.Spec.Image == defaultInstanceManagerImage || im.Spec.Image == nc.instanceManagerImage) && im.Spec.DataEngine == dataEngine {
+				if !isNodeSelectorMatching {
+					if runningOrStartingInstanceFound {
+						cleanupRequired = false
+						log.Infof("Keeping instance manager %v on selector-excluded node %v because it still has running/starting instances", im.Name, node.Name)
+					} else {
+						log.Infof("Cleaning up instance manager %v because node %v does not match %v", im.Name, node.Name, types.SettingNameSystemManagedComponentsNodeSelector)
+					}
```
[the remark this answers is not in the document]
COLDTURNIP: We should still remove the `Unknown` instance manager from the node that is not selected. The `Unknown` state instance manager may not always repairable.

controller/node_controller.go
```diff
@@ -1169,7 +1191,14 @@ func (nc *NodeController) syncInstanceManagers(node *longhorn.Node) error {
...
```
derekbit, line 1204: In the new `!isNodeSelectorMatching` branch, an IM with no running or starting instances is deleted immediately. An IM in InstanceManagerStateUnknown (node unreachable / network partition) gets cleaned up.

When a node is both excluded by the selector and `Unknown`, we can't confirm whether its engine/replica instances are actually still running. It is better to skip the cleanup of an Unknown-state instance manager on a selector-excluded node, because while the node is unreachable we cannot confirm whether its engine/replica instances are still running.

WDYT?


```suggestion
					if !isNodeSelectorMatching {
					    if runningOrStartingInstanceFound {
					        cleanupRequired = false
					        log.Infof("Keeping instance manager %v on selector-excluded node %v because it still has running/starting instances", im.Name, node.Name)
					    } else if im.Status.CurrentState == longhorn.InstanceManagerStateUnknown && im.DeletionTimestamp == nil {
					        // The node is unreachable, so the real instance state is unknown. Defer cleanup until the node is observable again.
					        cleanupRequired = false
					        log.Infof("Skipping cleanup of instance manager %v on selector-excluded node %v because it is in unknown state", im.Name, node.Name)
					    } else {
					        log.Infof("Cleaning up instance manager %v because node %v does not match %v", im.Name, node.Name, types.SettingNameSystemManagedComponentsNodeSelector)
					    }
					}
```

controller/engine_image_controller.go
```diff
@@ -338,21 +343,37 @@ func (ic *EngineImageController) syncEngineImage(key string) (err error) {
...
+				continue
+			}
+		}
+		targetCount++
+		if engineImage.Status.NodeDeploymentMap[nodeName] {
+			deployedCount++
+		}
+	}
+
+	if deployedCount < targetCount {
```
derekbit: When the `system-managed-components-node-selector` is misconfigured so it matches 0 ready nodes, `targetCount` stays 0. The condition `deployedCount < targetCount` evaluates to` 0 < 0` (false), so execution falls into the `else` branch and the EngineImage is marked Ready=True / State=Deployed.

```
if targetCount == 0 {
    // The node selector matches no ready node, so the engine image is not deployed anywhere and cannot serve any volume.
    message := fmt.Sprintf("Engine image %v (%v) is not deployed on any ready node because no ready node matches the system managed components node selector", engineImage.Name, engineImage.Spec.Image)
    log.Warn(message)
    engineImage.Status.Conditions = types.SetCondition(engineImage.Status.Conditions, longhorn.EngineImageConditionTypeReady, longhorn.ConditionStatusFalse,
        longhorn.EngineImageConditionTypeReadyReasonDaemonSet, message)
    engineImage.Status.State = longhorn.EngineImageStateDeploying
} else if deployedCount < targetCount {
    ...
} else {
    ...
}
```

controller/node_controller.go
COLDTURNIP: But the `system-managed-components-node-selector` is a danger zone settings, which implies all the volumes must be detached at this moment. Do we still need to concern about the `unknown` IMs in this case?

controller/engine_image_controller.go
derekbit: @COLDTURNIP Could you address the comment?

controller/node_controller.go
derekbit: The "all volumes detached" precondition isn't actually enforced on this path. The setting value is accepted immediately. `syncInstanceManagers` reads the selector directly and cleans up non-matching IMs with no detachment gate.

controller/engine_image_controller.go
```diff
@@ -338,21 +343,41 @@ func (ic *EngineImageController) syncEngineImage(key string) (err error) {
...
-		}
-	}
-
 	readyNodes, err := ic.ds.ListReadyNodesRO()
 	if err != nil {
 		return err
 	}
 
-	if deployedNodeCount < len(readyNodes) {
+	nodeSelector := ds.Spec.Template.Spec.NodeSelector
```
COLDTURNIP, line 362: `ds.Spec.Template.Spec.NodeSelector` is now updated after invoking `ic.syncEngineImageDaemonSetNodeSelector` earlier.

derekbit (commented):
Should engine image controller listen to the changes of the settings?

controller/engine_image_controller.go
```diff
@@ -774,6 +799,48 @@ func (ic *EngineImageController) ResolveRefAndEnqueue(namespace string, ref *met
...
+	// setting. Updating the Engine Image DaemonSet pod template restarts Engine
+	// Image pods; if volumes are attached, the restart can make engine images
+	// unavailable and trigger image deletion. Apply the DaemonSet selector
+	// change only after all volumes are detached.
+	detached, err := ic.ds.AreAllVolumesDetachedState()
+	if err != nil {
+		return errors.Wrapf(err, "failed to check volume detachment for %v setting update", types.SettingNameSystemManagedComponentsNodeSelector)
+	}
+	if !detached {
+		ic.logger.Infof("Skipping node selector update for engine image daemonset %v because some volumes are still attached", ds.Name)
```
derekbit: Might be result in too many noise.
```suggestion
		ic.logger.Debugf("Skipping node selector update for engine image daemonset %v because some volumes are still attached", ds.Name)
```

derekbit (dismissed):
In general, LGTM.

controller/engine_image_controller.go
```diff
@@ -338,21 +354,41 @@ func (ic *EngineImageController) syncEngineImage(key string) (err error) {
...
+	nodeSelector := ds.Spec.Template.Spec.NodeSelector
+
+	targetCount, deployedCount := 0, 0
+	selector := labels.SelectorFromSet(nodeSelector)
+	for nodeName := range readyNodes {
+		if len(nodeSelector) > 0 {
+			kubeNode, err := ic.ds.GetKubernetesNodeRO(nodeName)
+			if err != nil {
+				return errors.Wrapf(err, "failed to get Kubernetes node %v while checking system managed components node selector", nodeName)
+			}
```
derekbit, line 376:
```suggestion
			if err != nil {
				if !datastore.ErrorIsNotFound(err) {
					return errors.Wrapf(err, "failed to get Kubernetes node %v while checking system managed components node selector", nodeName)
				}
				// The Kubernetes node is gone (e.g. the node is being removed), so it
				// cannot host an engine image pod. Skip it instead of failing the whole
				// readiness calculation; a later reconcile re-evaluates.
				log.WithError(err).Debugf("Skipping missing Kubernetes node %v while checking system managed components node selector", nodeName)
				continue
			}
```

derekbit (approved):

PR #4642, opened 2026-04-16
feat(backuptarget): enforce NFS soft-mount options in webhook mutator

COLDTURNIP:

Issue longhorn/longhorn#12896

Always inject `soft`, `timeo=300`, and `retry=2` into `nfsOptions` for NFS BackupTarget URLs. This ensures that longhorn-backup-store NFS mounts time out instead of blocking indefinitely when the NFS server is slow or unreachable, preventing the process accumulation described in longhorn/longhorn#12896.

Merge rules applied in backupTargetMutator.mutate via admission webhook:
- `hard` removed when present
- `soft` added if absent
- `timeo=N` kept if already specified, otherwise `timeo=300` added
- `retry=N` kept if already specified, otherwise `retry=2` added

### DIFF
--- a/webhook/resources/backuptarget/mutator.go
+++ b/webhook/resources/backuptarget/mutator.go
@@ -1,7 +1,11 @@
 package backuptarget
 
 import (
+	"encoding/json"
 	"fmt"
+	"net/url"
+	"slices"
+	"strings"
 
 	"github.com/cockroachdb/errors"
 
@@ -10,6 +14,7 @@ import (
 	admissionregv1 "k8s.io/api/admissionregistration/v1"
 
 	"github.com/longhorn/longhorn-manager/datastore"
+	"github.com/longhorn/longhorn-manager/types"
 	"github.com/longhorn/longhorn-manager/webhook/admission"
 
 	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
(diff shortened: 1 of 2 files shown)


webhook/resources/backuptarget/mutator.go
```diff
@@ -66,5 +70,114 @@ func mutate(newObj runtime.Object) (admission.PatchOps, error) {
...
+		return nil, werror.NewInvalidError(err.Error(), "")
+	}
+	if newURL != backupTarget.Spec.BackupTargetURL {
+		marshaledURL, err := json.Marshal(newURL)
+		if err != nil {
+			err = errors.Wrapf(err, "failed to marshal backupTargetURL for backupTarget %v", backupTarget.Name)
+			return nil, werror.NewInvalidError(err.Error(), "")
+		}
+		patchOps = append(patchOps, fmt.Sprintf(`{"op": "replace", "path": "/spec/backupTargetURL", "value": %s}`, string(marshaledURL)))
+	}
```
derekbit, line 85: Could you compare the option list rather than the string?

webhook/resources/backuptarget/mutator.go
```diff
@@ -66,5 +71,132 @@ func mutate(newObj runtime.Object) (admission.PatchOps, error) {
...
+		// nfsOptions key is present but all values are blank -- strip it.
+		q.Del(NFSOptionsKey)
+		u.RawQuery = q.Encode()
+		return u.String(), true, nil
+	}
+
+	merged := mergeNFSSoftMountOptions(opts)
+	if slices.Equal(opts, merged) {
+		return rawURL, false, nil
+	}
```
COLDTURNIP, line 145: Introduce a flag to indicate if the content is changed by item level (item removal or append).

webhook/resources/backuptarget/mutator.go
COLDTURNIP: Updated. See https://github.com/longhorn/longhorn-manager/pull/4642#discussion_r3097898006

derekbit (approved):
LGTM

derekbit:
@mergify backport v1.11.x

derekbit:
@mergify backport v1.10.x

PR #4646, opened 2026-04-17
fix(system-backup): clean associated backup/snapshot CRs on retention…

carterli0407-cell:
## Which issue(s) this PR fixes:
Issue [longhorn/longhorn#12915](https://github.com/longhorn/longhorn/issues/12915)

## What this PR does / why we need it:
SystemBackup recurring jobs accumulate system-backup-created volume snapshots and backups beyond the configured retain count.
This PR implements three fixes:
1. **Label on creation**: Stamp label on Snapshot and Backup CRs when created by SystemBackup controller, linking them to the parent SystemBackup CR.
2. **Respect auto-cleanup setting**: Delete intermediate snapshots after their volume backup completes, when `auto-cleanup-recurring-job-backup-snapshot` is enabled.
3. **Cascade cleanup on deletion**: When a SystemBackup enters Deleting state, list and delete all associated Backup and Snapshot CRs before removing the finalizer.
## Expected behavior after fix:
- When recurring job retention logic deletes an expired SystemBackup CR, all associated volume snapshot and backup CRs are also cleaned up.
- With retain=7, there are at most 7 system-backup-associated snapshot and backup CRs per volume.
- When auto-cleanup-recurring-job-backup-snapshot is enabled, snapshots are removed immediately after the backup completes.

## Testing performed:
- Unit tests added in `controller/system_backup_controller_test.go`:
  - `TestSystemBackupCreatedChildrenCarrySystemBackupLabel` - verifies labels stamped on creation
  - `TestSystemBackupWaitForVolumeBackupToCompleteAutoCleanup` - verifies auto-cleanup behavior
  - `TestSystemBackupDeletingStateCleansAssociatedBackupAndSnapshotCRs` - verifies cascade cleanup
- Verified label selectors correctly identify associated CRs
- Verified no orphaned CRs after SystemBackup deletion

## Special notes for your reviewer:
- This is forward-safe: labels are only stamped on newly created resources
- Existing unlabeled legacy resources are not affected by the cascade cleanup logic
- The cascade cleanup respects NotFound errors gracefully to handle race conditions
- Setting `auto-cleanup-recurring-job-backup-snapshot` already existed; this PR applies it to SystemBackup backups

## Files changed:
- `types/types.go` - Added `LonghornLabelSystemBackup` constant and `GetSystemBackupLabelKey()` helper
- `controller/system_backup_controller.go` - Updated snapshot/backup creation, added auto-cleanup logic, added cascade cleanup in Deleting state
- `controller/system_backup_controller_test.go` - Added 3 new test cases

- Related setting: `SettingNameAutoCleanupRecurringJobBackupSnapshot`
- Datastore helper methods used: `ListSnapshotsRO()`, `ListBackupsRO()`, `DeleteSnapshot()`, `DeleteBackup()`
- Issue: https://github.com/longhorn/longhorn/issues/12915

## Summary by CodeRabbit

- **New Features**
  - Added automatic cleanup of related backups and snapshots when a system backup is deleted.
  - Added labels to associate created backups and snapshots with their system backup.
  - Added an option to automatically remove snapshots after volume backups complete.
  - Added cleanup for temporary snapshots when an existing backup is already up to date.

- **Tests**
  - Added coverage for label propagation, snapshot auto-cleanup, and deletion workflows.


controller/system_backup_controller.go
```diff
@@ -864,13 +929,19 @@ func (c *SystemBackupController) WaitForVolumeBackupToComplete(backups map[strin
...
+		autoCleanupSnapshot = false
+	}
+
 	startTime := time.Now()
 	ticker := time.NewTicker(time.Second * 5)
 	defer ticker.Stop()
 
 	for range ticker.C {
 		for name := range backups {
-			// Retrieve the latest backup
```
c3y1huang, line 873: Is there a reason for removing the comments?

controller/system_backup_controller.go
```diff
@@ -485,6 +494,62 @@ func (c *SystemBackupController) reconcile(name string, backupTargetClient engin
...
+		}
+		if err := c.ds.DeleteBackup(backup.Name); err != nil && !apierrors.IsNotFound(err) {
+			return false, errors.Wrapf(err, "failed to delete associated backup %v", backup.Name)
+		}
+	}
+
+	if remaining > 0 {
+		log.Infof("Waiting for associated backup/snapshot CRs to be removed for system backup %v", systemBackup.Name)
+		return false, nil
+	}
```
c3y1huang: This feels a bit over-engineered. Does the controller really need to requeue just to check what's remaining? A fire-and-forget(best-effort) might be sufficient here.

controller/system_backup_controller.go
```diff
@@ -472,6 +473,14 @@ func (c *SystemBackupController) reconcile(name string, backupTargetClient engin
 		cleanupLocalSystemBackupFiles(tempBackupArchivePath, tempBackupDir, log)
 
 	case longhorn.SystemBackupStateDeleting:
+		allDeleted, cleanupErr := c.cleanupAssociatedBackupAndSnapshotCRs(systemBackup, log)
```
c3y1huang: Is it possible that the controller deleted a volume backup while the system backup file is still in the backup store? Would that cause restore failures when another cluster tries to use it?

controller/system_backup_controller.go
```diff
@@ -485,6 +494,62 @@ func (c *SystemBackupController) reconcile(name string, backupTargetClient engin
...
+		if snapshot.DeletionTimestamp != nil {
+			continue
+		}
+		if err := c.ds.DeleteSnapshot(name); err != nil && !apierrors.IsNotFound(err) {
+			return false, errors.Wrapf(err, "failed to delete associated snapshot %v", name)
+		}
+	}
+
+	// Backups filtered by label
+	backupList, err := c.ds.ListBackupsRO()
```
c3y1huang: Suggest improving ListBackupsRO to support input selectors.

controller/system_backup_controller.go
```diff
@@ -864,13 +929,19 @@ func (c *SystemBackupController) WaitForVolumeBackupToComplete(backups map[strin
 		c.handleStatusUpdate(record, systemBackup, existingSystemBackup, err, log)
 	}()
 
+	autoCleanupSnapshot := false
```
c3y1huang: The initialization looks redundant.

c3y1huang (commented):
Remember to use your SUSE email for the sign-off.

controller/system_backup_controller.go
```diff
@@ -818,11 +883,6 @@ func (c *SystemBackupController) backupVolumesIfNotPresent(systemBackup *longhor
 
 		volumeBackupName := bsutil.GenerateName("system-backup")
 
-		snapshot, err := c.createVolumeSnapshot(ctx, volume, volumeBackupName)
```
c3y1huang, line 821: Without creating a new volume snapshot, how can the controller validate if the backup is up to date?

controller/system_backup_controller.go
```diff
@@ -485,6 +494,62 @@ func (c *SystemBackupController) reconcile(name string, backupTargetClient engin
...
+		}
+		if backup.Labels[types.GetSystemBackupLabelKey()] != systemBackup.Name {
+			continue
+		}
+
+		remaining++
+		if backup.DeletionTimestamp != nil {
+			continue
+		}
+		if err := c.ds.DeleteBackup(backup.Name); err != nil && !apierrors.IsNotFound(err) {
```
c3y1huang: What happens when the volume backup policy is `if-not-present`? Could a volume in the system backup end up referencing a nonexisting backup?

c3y1huang (changes requested):

c3y1huang (commented):
Suggest running the relevant Longhorn tests to make sure we don't introduce regressions.

derekbit:
> Suggest running the relevant Longhorn tests to make sure we don't introduce regressions.

https://ci.longhorn.io/
@carterli0407-cell  You can check with @longhorn/qa.

controller/system_backup_controller.go
```diff
@@ -485,6 +494,62 @@ func (c *SystemBackupController) reconcile(name string, backupTargetClient engin
...
+			return false, errors.Wrapf(err, "failed to delete associated backup %v", backup.Name)
+		}
+	}
```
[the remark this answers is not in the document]
carterli0407-cell: I added selector-based backup listing in datastore and switched system backup cleanup to use label selectors for both Backup and Snapshot CRs.

controller/system_backup_controller.go
```diff
@@ -485,6 +494,62 @@ func (c *SystemBackupController) reconcile(name string, backupTargetClient engin
...
+			continue
+		}
+		remaining++
+		if snapshot.DeletionTimestamp != nil {
+			continue
+		}
+		if err := c.ds.DeleteSnapshot(name); err != nil && !apierrors.IsNotFound(err) {
+			return false, errors.Wrapf(err, "failed to delete associated snapshot %v", name)
+		}
+	}
```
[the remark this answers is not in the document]
carterli0407-cell: I reordered cleanup to delete Backup CRs first, then Snapshot CRs.

controller/system_backup_controller.go
```diff
@@ -472,6 +473,14 @@ func (c *SystemBackupController) reconcile(name string, backupTargetClient engin
 		cleanupLocalSystemBackupFiles(tempBackupArchivePath, tempBackupDir, log)
 
 	case longhorn.SystemBackupStateDeleting:
+		allDeleted, cleanupErr := c.cleanupAssociatedBackupAndSnapshotCRs(systemBackup, log)
+		if cleanupErr != nil {
+			return cleanupErr
+		}
+		if !allDeleted {
+			c.enqueue(systemBackup)
+			return nil
```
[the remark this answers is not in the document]
carterli0407-cell: I replaced immediate requeue with delayed requeue to avoid hot-looping while child deletions propagate.

controller/system_backup_controller.go
carterli0407-cell: I restored snapshot-first behavior for if-not-present validation and delete the intermediate snapshot immediately when no backup is needed.

controller/system_backup_controller_test.go
```diff
@@ -485,3 +485,308 @@ func fakeSystemRolloutNamespace(c *C, informerFactory informers.SharedInformerFa
...
+		checkFinalizer := !util.FinalizerExists(longhornFinalizerKey, systemBackup) == tc.expectRemove
+		c.Assert(checkFinalizer, Equals, true)
+
+		_, backupErr := lhClient.LonghornV1beta2().Backups(TestNamespace).Get(context.TODO(), backup.Name, metav1.GetOptions{})
+		_, snapshotErr := lhClient.LonghornV1beta2().Snapshots(TestNamespace).Get(context.TODO(), snapshot.Name, metav1.GetOptions{})
+
+		// After the first reconcile the children should already have been deleted (or be deleting),
+		// but the SystemBackup finalizer should remain until the next reconcile.
+		c.Assert(backupErr, NotNil)
+		c.Assert(snapshotErr, NotNil)
```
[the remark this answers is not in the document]
carterli0407-cell: Updated tests to assert apierrors.IsNotFound where deletion is expected and simplified finalizer assertions for readability.

controller/system_backup_controller_test.go
```diff
@@ -485,3 +485,308 @@ func fakeSystemRolloutNamespace(c *C, informerFactory informers.SharedInformerFa
...
+		err = systemBackupController.WaitForVolumeBackupToComplete(backups, systemBackup)
+		c.Assert(err, IsNil)
+
+		for _, backup := range backups {
+			_, err := lhClient.LonghornV1beta2().Snapshots(TestNamespace).Get(context.TODO(), backup.Spec.SnapshotName, metav1.GetOptions{})
+			if tc.expectSnapshotDelete {
+				c.Assert(err, NotNil)
+			} else {
+				c.Assert(err, IsNil)
+			}
```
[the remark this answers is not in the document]
carterli0407-cell: Updated tests to assert apierrors.IsNotFound where deletion is expected and simplified finalizer assertions for readability.

derekbit:
@carterli0407-cell Ready for review?

carterli0407-cell:
> @carterli0407-cell Ready for review?

Yes

derekbit:
@carterli0407-cell Could you add the link to the regression/e2e testing?
https://github.com/longhorn/longhorn/issues/12932

derekbit:
Hello @carterli0407-cell
Please make sure to address and respond to all comments, and don't resolve them without handling. If you believe a comment doesn't need to be addressed, please provide an explanation.

controller/system_backup_controller_test.go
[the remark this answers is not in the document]
derekbit: Seems not updated?

controller/system_backup_controller.go
c3y1huang: Could a volume (with `if-not-present` policy, and no data change) in the newer system backup end up referencing a nonexistent backup (previously created) due to the cleanup logic?

controller/system_backup_controller_test.go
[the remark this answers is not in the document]
carterli0407-cell: I've updated.

controller/system_backup_controller.go
carterli0407-cell: No strong reason on my side. I restored the important comments.

controller/system_backup_controller.go
carterli0407-cell: I changed it to delayed requeue to avoid hot-looping while still guaranteeing deterministic cleanup before finalizer removal.

controller/system_backup_controller.go
carterli0407-cell: I implemented selector-based backup listing in datastore and switched cleanup to use selector-based query instead of scanning all backups in-memory.

controller/system_backup_controller.go
carterli0407-cell: @c3y1huang , my understanding of the code is:
In "if-not-present mode", no new backup is created when data is unchanged.
Association labels are typically stamped only when a backup is created.
Reused backups are not automatically re-associated with the newer SystemBackup.
Deleting the older SystemBackup can delete that backup data, even though a newer SystemBackup still relies on it.
I would like to hear your opinion.

controller/system_backup_controller.go
carterli0407-cell: I restored snapshot-first behavior for if-not-present validation. If backup is determined up-to-date, the intermediate snapshot is cleaned up immediately. This keeps correctness while preventing snapshot accumulation.

controller/system_backup_controller.go
carterli0407-cell: I have updated logic/discussion accordingly and will keep this protected by reference/association checks.

controller/system_backup_controller_test.go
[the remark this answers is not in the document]
derekbit: > I've updated.

But the code is not changed?

controller/system_backup_controller.go
```diff
@@ -485,6 +496,57 @@ func (c *SystemBackupController) reconcile(name string, backupTargetClient engin
...
+	selector := labels.SelectorFromSet(labels.Set{
+		types.GetSystemBackupLabelKey(): systemBackup.Name,
+	})
+
+	remaining := 0
+
+	// Delete backups first.
+	backups, err := c.ds.ListBackupsWithSelectorRO(selector)
+	if err != nil {
+		return false, errors.Wrap(err, "failed to list associated backups")
```
derekbit:
```suggestion
	if err != nil {
		return false, errors.Wrapf(err, "failed to list associated backups with selector ....
```

controller/system_backup_controller.go
```diff
@@ -485,6 +496,57 @@ func (c *SystemBackupController) reconcile(name string, backupTargetClient engin
...
+			continue
+		}
+		remaining++
+		if backup.DeletionTimestamp != nil {
+			continue
+		}
+		if err := c.ds.DeleteBackup(backup.Name); err != nil && !apierrors.IsNotFound(err) {
+			return false, errors.Wrapf(err, "failed to delete associated backup %v", backup.Name)
+		}
+	}
```
derekbit, line 525: Create a helper function for the deletion

controller/system_backup_controller.go
```diff
@@ -472,6 +473,16 @@ func (c *SystemBackupController) reconcile(name string, backupTargetClient engin
 		cleanupLocalSystemBackupFiles(tempBackupArchivePath, tempBackupDir, log)
 
 	case longhorn.SystemBackupStateDeleting:
+		allDeleted, cleanupErr := c.cleanupAssociatedBackupAndSnapshotCRs(systemBackup, log)
+		if cleanupErr != nil {
+			return cleanupErr
+		}
+		if !allDeleted {
+			if err := enqueueAfterDelay(c.queue, systemBackup, 2*time.Second); err != nil {
```
derekbit: Why do we need to enqueue? Backup deletion might take time because it involves the data (block files) deletion. If backup CRs are marked as deleted, we don't need to care about them.

controller/system_backup_controller.go
```diff
@@ -485,6 +496,57 @@ func (c *SystemBackupController) reconcile(name string, backupTargetClient engin
...
+			return false, errors.Wrapf(err, "failed to delete associated snapshot %v", name)
+		}
+	}
+
+	if remaining > 0 {
+		log.Infof("Waiting for associated backup/snapshot CRs to be removed for system backup %v", systemBackup.Name)
+		return false, nil
+	}
+
+	return true, nil
```
derekbit: IMO, we don't need the `remaining` and enqueue as I mentioned.

controller/system_backup_controller.go
```diff
@@ -485,6 +496,57 @@ func (c *SystemBackupController) reconcile(name string, backupTargetClient engin
...
+			continue
+		}
+		remaining++
+		if snapshot.DeletionTimestamp != nil {
+			continue
+		}
+		if err := c.ds.DeleteSnapshot(name); err != nil && !apierrors.IsNotFound(err) {
+			return false, errors.Wrapf(err, "failed to delete associated snapshot %v", name)
+		}
+	}
```
derekbit, line 545: Add a help function handling the deletions

controller/system_backup_controller.go
```diff
@@ -485,6 +496,57 @@ func (c *SystemBackupController) reconcile(name string, backupTargetClient engin
...
+		if backup == nil {
+			continue
+		}
+		remaining++
+		if backup.DeletionTimestamp != nil {
+			continue
+		}
+		if err := c.ds.DeleteBackup(backup.Name); err != nil && !apierrors.IsNotFound(err) {
+			return false, errors.Wrapf(err, "failed to delete associated backup %v", backup.Name)
+		}
```
derekbit, line 524: Can we continue the remaining deletions after hitting an error?

controller/system_backup_controller.go
```diff
@@ -485,6 +496,57 @@ func (c *SystemBackupController) reconcile(name string, backupTargetClient engin
 	return nil
 }
 
+func (c *SystemBackupController) cleanupAssociatedBackupAndSnapshotCRs(systemBackup *longhorn.SystemBackup, log logrus.FieldLogger) (allDeleted bool, err error) {
```
derekbit: Suggest 
```
func (c *SystemBackupController) cleanupAssociatedBackupAndSnapshotCRs(systemBackup *longhorn.SystemBackup, log logrus.FieldLogger) (err error) {

foreach backup {
    delete backup
     record error to errs if hitting an error
}

foreach snapshot {
    delete snapshot
     record error to errs if hitting an error
}

return errs
}
```

derekbit (commented):
One question: Is it possible to handle the existing and accumulated snapshots and backups?

carterli0407-cell:
> One question: Is it possible to handle the existing and accumulated snapshots and backups?

I tested it on my side, but it does not seem to work at the moment.

derekbit:
> > One question: Is it possible to handle the existing and accumulated snapshots and backups?
> 
> I tested it on my side, but it does not seem to work at the moment.

How did you test? Can you elaborate more on why it doesn't work from the codes perspective?

carterli0407-cell:
> How did you test? Can you elaborate more on why it doesn't work from the codes perspective?

I just tested it in an environment where the issue was already reproducible, and confirmed the behavior only from the observed result.

I have not investigated it deeply from the code perspective yet, but I can check that further if needed.

derekbit:
> > How did you test? Can you elaborate more on why it doesn't work from the codes perspective?
> 
> I just tested it in an environment where the issue was already reproducible, and confirmed the behavior only from the observed result.
> 
> I have not investigated it deeply from the code perspective yet, but I can check that further if needed.

Please go review the system backup feature in detail and check your implementation. Always evaluate whether the improvement can be applied to an existing cluster. If it cannot, provide a clear explanation of the constraints and reasoning.

controller/system_backup_controller.go
```diff
@@ -864,6 +929,12 @@ func (c *SystemBackupController) WaitForVolumeBackupToComplete(backups map[strin
 		c.handleStatusUpdate(record, systemBackup, existingSystemBackup, err, log)
 	}()
 
+	autoCleanupSnapshot, settingErr := c.ds.GetSettingAsBool(types.SettingNameAutoCleanupRecurringJobBackupSnapshot)
+	if settingErr != nil {
+		log.WithError(settingErr).Warnf("Failed to get setting %v, fallback to disabled", types.SettingNameAutoCleanupRecurringJobBackupSnapshot)
+		autoCleanupSnapshot = false
+	}
```
derekbit: Why do we get the setting value early? Can we retrieve the value right before we need it?

controller/system_backup_controller.go
carterli0407-cell: @derekbit , I agree that once Backup/Snapshot CRs are already marked for deletion (DeletionTimestamp set), we should not keep waiting for their underlying data cleanup.
I will updat the cleanup logic to count only non-terminating associated CRs as remaining.

But I would like to keep enqueue/requeue in Deleting state for safety. I would like to the feedback for my opinion.

controller/system_backup_controller.go
```diff
@@ -896,12 +975,10 @@ func (c *SystemBackupController) WaitForVolumeBackupToComplete(backups map[strin
 			return
 		}
 
-		// Return error when Volume backup exceeds timeout
 		if time.Since(startTime) > datastore.VolumeBackupTimeout {
 			return fmt.Errorf("timed out waiting for Volume backups to complete")
 		}
 	}
-	// This should never be reached, return this error just in case.
 	return fmt.Errorf("unexpected error: stopped waiting for Volume backups without completing, failing or timing out")
```
derekbit, line 905: Why are the comments removed?

controller/system_backup_controller.go
carterli0407-cell: @derekbit , I adjusted logic so terminating Backup/Snapshot CRs are not counted as remaining. I’d prefer to keep one requeue safeguard for reconciliation consistency before finalizer removal. If you still prefer full fire-and-forget, I can simplify further by removing.

controller/system_backup_controller.go
carterli0407-cell: I extracted the associated backup deletion logic into a dedicated helper

controller/system_backup_controller.go
derekbit: > I will updat the cleanup logic to count only non-terminating associated CRs as remaining.

Please check https://github.com/longhorn/longhorn-manager/pull/4646#discussion_r3115006403.

> But I would like to keep enqueue/requeue in Deleting state for safety. I would like to the feedback for my opinion.

If it errors out, the request will be enqueued. You don't need to requeue it here. Please check 
- https://github.com/longhorn/longhorn-manager/blob/master/controller/system_backup_controller.go#L187-L198
- https://github.com/longhorn/longhorn-manager/blob/master/controller/system_backup_controller.go#L200C1-L200C71

controller/system_backup_controller.go
derekbit: See the comment https://github.com/longhorn/longhorn-manager/pull/4646#discussion_r3115262389

controller/system_backup_controller.go
carterli0407-cell: I extracted the associated backup deletion logic into a dedicated helper

controller/system_backup_controller.go
derekbit: ~~Codes are not updated yet.~~

controller/system_backup_controller.go
carterli0407-cell: I restored the comment.

controller/system_backup_controller.go
```diff
@@ -485,6 +489,68 @@ func (c *SystemBackupController) reconcile(name string, backupTargetClient engin
...
+		if backup == nil {
+			continue
+		}
+		if backup.DeletionTimestamp != nil {
+			continue
+		}
+		if err := c.deleteAssociatedBackup(backup); err != nil {
+			return err
+		}
+	}
```
derekbit, line 525: Why not moving the codes to `deleteAssociatedBackups`

controller/system_backup_controller.go
```diff
@@ -485,6 +489,68 @@ func (c *SystemBackupController) reconcile(name string, backupTargetClient engin
...
+		if snapshot == nil {
+			continue
+		}
+		if snapshot.DeletionTimestamp != nil {
+			continue
+		}
+		if err := c.deleteAssociatedSnapshot(name); err != nil {
+			return err
+		}
+	}
```
derekbit: Why not moving the codes to `deleteAssociatedSnapshots`

controller/system_backup_controller.go
```diff
@@ -485,6 +489,68 @@ func (c *SystemBackupController) reconcile(name string, backupTargetClient engin
...
+	for _, backup := range backups {
+		if backup == nil {
+			continue
+		}
+		if backup.DeletionTimestamp != nil {
+			continue
+		}
+		if err := c.deleteAssociatedBackup(backup); err != nil {
+			return err
```
derekbit: Why not continue the remaining deletions and collect the backups/snapshos that failed to delete?

controller/system_backup_controller.go
```diff
@@ -485,6 +489,68 @@ func (c *SystemBackupController) reconcile(name string, backupTargetClient engin
...
+	}
+
+	if err := c.ds.DeleteBackup(backup.Name); err != nil && !apierrors.IsNotFound(err) {
+		return errors.Wrapf(err, "failed to delete associated backup %v", backup.Name)
+	}
+
+	return nil
+}
+
+func (c *SystemBackupController) deleteAssociatedSnapshot(snapshotName string) error {
```
derekbit:
```suggestion
func (c *SystemBackupController) deleteAssociatedSnapshost(snapshotName string) error {
```

controller/system_backup_controller.go
```diff
@@ -485,6 +489,68 @@ func (c *SystemBackupController) reconcile(name string, backupTargetClient engin
...
+		}
+		if err := c.deleteAssociatedSnapshot(name); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (c *SystemBackupController) deleteAssociatedBackup(backup *longhorn.Backup) error {
```
derekbit:
```suggestion
func (c *SystemBackupController) deleteAssociatedBackups(backup *longhorn.Backup) error {
```

controller/system_backup_controller.go
```diff
@@ -485,6 +489,68 @@ func (c *SystemBackupController) reconcile(name string, backupTargetClient engin
...
+	if backup == nil || backup.DeletionTimestamp != nil {
+		return nil
+	}
+
+	if err := c.ds.DeleteBackup(backup.Name); err != nil && !apierrors.IsNotFound(err) {
+		return errors.Wrapf(err, "failed to delete associated backup %v", backup.Name)
```
derekbit: IMO, the error message is not informative.

controller/system_backup_controller.go
```diff
@@ -485,6 +489,68 @@ func (c *SystemBackupController) reconcile(name string, backupTargetClient engin
...
+	}
+
+	return nil
+}
+
+func (c *SystemBackupController) deleteAssociatedSnapshot(snapshotName string) error {
+	if err := c.ds.DeleteSnapshot(snapshotName); err != nil && !apierrors.IsNotFound(err) {
+		return errors.Wrapf(err, "failed to delete associated snapshot %v", snapshotName)
```
derekbit: Same as above.

controller/system_backup_controller.go
carterli0407-cell: I can retrieve the value right before but I would like to avoid repeated setting lookups in the loop.

derekbit:
I will continue the 2nd-round review after addressing all the comments from @c3y1huang and me

controller/system_backup_controller.go
carterli0407-cell: what does it mean?
Earlier name is wrong?

derekbit: Sorry, typo. Should be `deleteAssociatedSnapshots(...)`

derekbit: It is better to move the for loop into `deleteAssociatedSnapshots` and `deleteAssociatedBackups` to make the `cleanupAssociatedBackupAndSnapshotCRs` easy to read.

controller/system_backup_controller.go
derekbit: @carterli0407-cell Can you add the code links and elaborate more on how you handle the question raised by @c3y1huang?

controller/system_backup_controller.go
derekbit: @carterli0407-cell The answer and implementation don't address the concern from @c3y1huang

controller/system_backup_controller.go
carterli0407-cell: I updated the code to make those errors more informative

controller/system_backup_controller.go
carterli0407-cell: I updated the code to make those errors more informative

controller/system_backup_controller.go
carterli0407-cell: Updated.

controller/system_backup_controller.go
carterli0407-cell: I updated

controller/system_backup_controller.go
carterli0407-cell: The delete operations are already extracted into helper functions, but the current flow is still fail-fast.

controller/system_backup_controller.go
carterli0407-cell: The delete operations are already extracted into helper functions, but the current flow is still fail-fast.

controller/system_backup_controller.go
derekbit: Hmm, the helper function aims to simplify `cleanupAssociatedBackupAndSnapshotCRs` to make it more readable. If you leave the for loops in the `cleanupAssociatedBackupAndSnapshotCRs`, then the helper funtions are meaningless.

> the current flow is still fail-fast

IMO, this is not an issue, you can handle errors inside `deleteAssociatedBackups` and `deleteAssociatedSnapshots` and return them. Isn't it?

controller/system_backup_controller.go
```diff
@@ -864,6 +933,12 @@ func (c *SystemBackupController) WaitForVolumeBackupToComplete(backups map[strin
 		c.handleStatusUpdate(record, systemBackup, existingSystemBackup, err, log)
 	}()
 
+	autoCleanupSnapshot, settingErr := c.ds.GetSettingAsBool(types.SettingNameAutoCleanupRecurringJobBackupSnapshot)
```
derekbit: nit: why not using `err`?
```suggestion
	autoCleanupSnapshot, err := c.ds.GetSettingAsBool(types.SettingNameAutoCleanupRecurringJobBackupSnapshot)
```

controller/system_backup_controller.go
derekbit: Please check https://github.com/longhorn/longhorn-manager/pull/4646#discussion_r3115470345

derekbit:
CI failed

image

controller/system_backup_controller.go
carterli0407-cell: Yes, agree with that. I've updated as you mentioned

controller/system_backup_controller.go
carterli0407-cell: Makes sense. I've updated.

controller/system_backup_controller.go
carterli0407-cell: I've updated.

controller/system_backup_controller.go
carterli0407-cell: We can switch this cleanup path to best-effort: continue deleting the remaining Backup/Snapshot CRs even after one deletion fails, ignore NotFound as usual, and collect non-NotFound errors to return as one aggregated error at the end.

controller/system_backup_controller.go
carterli0407-cell: I moved the code into helper function.

controller/system_backup_controller.go
carterli0407-cell: I changed.

controller/system_backup_controller.go
derekbit: The current implementation doesn't aggregate errors. If you prefer to return immediately upon encountering an error to avoid producing an overly large error message, that's fine. However, please ensure your response aligns with the implementation to prevent any confusion.


In addtion, the wrapped error message doesn't introduce any extra information, why not 
```
		if err := c.ds.DeleteBackup(backup.Name); err != nil && !apierrors.IsNotFound(err) {
			return err
		}
```

controller/system_backup_controller.go
```diff
@@ -485,6 +489,64 @@ func (c *SystemBackupController) reconcile(name string, backupTargetClient engin
...
+	for name, snapshot := range snapshots {
+		if snapshot == nil {
+			continue
+		}
+		if snapshot.DeletionTimestamp != nil {
+			continue
+		}
+		if err := c.ds.DeleteSnapshot(name); err != nil && !apierrors.IsNotFound(err) {
+			return errors.Wrapf(err, "failed to delete associated snapshot %v", name)
+		}
```
derekbit, line 544: Same question as above.

controller/system_backup_controller.go
```diff
@@ -485,6 +489,64 @@ func (c *SystemBackupController) reconcile(name string, backupTargetClient engin
...
+
+	// Delete backups first.
+	if err := c.deleteBackups(selector); err != nil {
+		return err
+	}
+
+	// Then delete snapshots.
+	if err := c.DeleteSnapshots(selector); err != nil {
+		return err
+	}
```
derekbit, line 505: Does the order matter? If yes, can you add a comment here to explain it?

controller/system_backup_controller.go
carterli0407-cell: The current implementation is fail-fast and does not aggregate errors. I'll make the response and implementation consistent to avoid confusion.

controller/system_backup_controller.go
carterli0407-cell: Yes, the order matters here.
We delete Backup CRs before Snapshot CRs because the backup is created from snapshot, so removing the Backup CR first makes the cleanup relationship clearer and avoids deleting the source snapshot before handling the corresponding backup cleanup. I'll add a comment here.

controller/system_backup_controller.go
c3y1huang: Discussed with @derekbit , consider introducing `status.owners.[]string` in the Backup & Snapshot CRDs to support reference tracking for shared resource ownership.

carterli0407-cell: @c3y1huang , I think this direction is reasonable, but it would require a CRD/status design update and a clear backward-compatibility plan for existing backups/snapshots that do not already have this ownership metadata.

derekbit: > @c3y1huang , I think this direction is reasonable, but it would require a CRD/status design update and a clear backward-compatibility plan for existing backups/snapshots that do not already have this ownership metadata.

@carterli0407-cell  Yes, we need to introduce a new field, handle the compatibility and handle the upgrade in the upgrade path. Please design and evaluate the tasks to see if we can complete it in v1.12.0. If you need more time, please let me know.

controller/system_backup_controller_test.go
```diff
@@ -485,3 +486,326 @@ func fakeSystemRolloutNamespace(c *C, informerFactory informers.SharedInformerFa
...
+		c.Assert(err, IsNil)
+
+		for _, backup := range backups {
+			_, err := lhClient.LonghornV1beta2().Snapshots(TestNamespace).Get(context.TODO(), backup.Spec.SnapshotName, metav1.GetOptions{})
+			if tc.expectSnapshotDelete {
+				c.Assert(apierrors.IsNotFound(err), Equals, true)
+			} else {
+				c.Assert(err, IsNil)
+			}
+		}
```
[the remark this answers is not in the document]
carterli0407-cell: Updated

controller/system_backup_controller.go
```diff
@@ -864,6 +929,12 @@ func (c *SystemBackupController) WaitForVolumeBackupToComplete(backups map[strin
 		c.handleStatusUpdate(record, systemBackup, existingSystemBackup, err, log)
 	}()
 
+	autoCleanupSnapshot, err := c.ds.GetSettingAsBool(types.SettingNameAutoCleanupRecurringJobBackupSnapshot)
+	if err != nil {
+		log.WithError(err).Warnf("Failed to get setting %v, fallback to disabled", types.SettingNameAutoCleanupRecurringJobBackupSnapshot)
+		autoCleanupSnapshot = false
+	}
```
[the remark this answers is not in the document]
carterli0407-cell: Updated.

PR #4647, opened 2026-04-19
fix: fix auto balance bugs

shuo-wu:

Issue https://github.com/longhorn/longhorn/issues/12926

# Root Cause Analysis: Healthy Volume Stuck in Rebuild-Cleanup Loop with Replica Auto Balance

## Loop Scenario:

1. Auto-balance detects imbalance → creates a new replica (without targeting a specific node)
2. Replica rebuilds successfully → volume has N+1 healthy replicas
3. Cleanup deletes a replica — but potentially the **wrong** one (including the newly added one)
4. Volume returns to the original imbalanced state → goto step 1

---

## The Bug1: `getReplicaCountForAutoBalanceLeastEffort` does not provide node affinity

```go
// volume_controller.go:3253-3258
case v.Spec.NumberOfReplicas == usableCount:
    if adjustCount := c.getReplicaCountForAutoBalanceLeastEffort(v, e, rs, c.getReplicaCountForAutoBalanceZone); adjustCount != 0 {
        return adjustCount, ""   // ← BUG: returns hardNodeAffinity = ""
    }
    if adjustCount := c.getReplicaCountForAutoBalanceLeastEffort(v, e, rs, c.getReplicaCountForAutoBalanceNode); adjustCount != 0 {
        return adjustCount, ""   // ← BUG: returns hardNodeAffinity = ""
    }
```

When `least-effort` detects an imbalance (e.g., 2 replicas on node A, 1 on node B, node C is empty), it returns `adjustCount=1` but **`hardNodeAffinity=""`** — meaning the new replica has no node placement constraint. The scheduler may place it on any node, including the already-overcrowded node A.

### Contrast with `best-effort` (line 3262–3274):

```go
// volume_controller.go:3262-3274
adjustCount, _, nCandidates := c.getReplicaCountForAutoBalanceBestEffort(v, e, rs, c.getReplicaCountForAutoBalanceNode)
if adjustCount == 0 {
    adjustCount, _, zCandidates := c.getReplicaCountForAutoBalanceBestEffort(v, e, rs, c.getReplicaCountForAutoBalanceZone)
    if adjustCount != 0 {
        nCandidates = c.getNodeCandidatesForAutoBalanceZone(v, e, rs, zCandidates)
    }
}
if adjustCount != 0 && len(nCandidates) != 0 {
    return adjustCount, nCandidates[0]   // ← Correctly uses node affinity
}

return adjustCount, ""   // ← BUG: Same problem when nCandidates is empty
```

`best-effort` **tries** to provide a target node via `nCandidates[0]`, but when `nCandidates` is empty (line 3274), it falls back to returning `adjustCount > 0` with `hardNodeAffinity = ""` — creating the same loop potential.

---

### Add side: `getReplicaCountForAutoBalanceNode` (line 3141–3158)
Only counts **`Running`** replicas:
```go
    continue
}
```

### Cleanup side: `getPreferredReplicaCandidatesForDeletion` (line 2400–2406)  
Counts **ALL** replicas (including non-running, rebuilding, failed):
```go
for _, r := range rs {
    diskToReplicaMap[r.Spec.NodeID+r.Spec.DiskID] = append(...)
    nodeToReplicaMap[r.Spec.NodeID] = append(...)
    ...
}
```

This means the "add" side may see node A as having 1 Running replica (imbalanced), while the "cleanup" side sees node A as having 2 replicas (including a non-running one), leading to different decisions about which node is overcrowded.

---

## The Bug3: `cleanupReplicaInUnstableEnv` cannot handle one node with multiple replicas

```go
    nodeReplicaMap := make(map[string]*longhorn.Replica, len(rs))
    for _, r := range rs {
        if isHealthyAndActiveReplica(r, false) {
            nodeReplicaMap[r.Spec.NodeID] = r
        }
    }
```

### DIFF
--- a/controller/volume_autobalance_test.go
+++ b/controller/volume_autobalance_test.go
@@ -3,6 +3,7 @@ package controller
 import (
 	"context"
 	"fmt"
+	"time"
 
 	. "gopkg.in/check.v1"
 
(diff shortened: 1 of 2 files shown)


shuo-wu:
All newly added unit tests pass. But I still need more time to check whether we can reproduce the bugs manually in my dev cluster

c3y1huang:
> The Bug1: getReplicaCountForAutoBalanceLeastEffort does not provide node affinity

This is intentional. The scheduler is responsible for handling placement decisions, so this function doesn't need to worry about node affinity. In the best-effort case, it returns the `hardNodeAffinity` because it handles the redistribution across nodes that already have replicas for a volume to bypass the precheck of the anti-affinity rules.

> The Bug2: Counting Mismatch Between Add and Cleanup

Likely a false alarm. The cleanups are ordered, and `cleanupExtraHealthyReplicas` is responsible for picking the extra replicas to remove, so it doesn’t need to match the add-side count exactly. Before that, `cleanupCorruptedOrStaleReplicas`, `cleanupFailedToScheduleReplicas` will handle unhealthy replicas first.
https://github.com/longhorn/longhorn-manager/blob/981638876f85c2c9210352b1d7b0676ae2039e20/controller/volume_controller.go#L1026

derekbit:
Agree with @c3y1huang's comment. Could you only address the BUG 3 in this PR?

controller/volume_autobalance_test.go
```diff
@@ -0,0 +1,1336 @@
...
+// loop (issue #11730 and #12926) observed in production:
+//
+//  1. Zone/Node auto-balance creates a replica on a node whose kube Ready transition
+//     time is 30+ minutes later than other nodes (the "unstable" node).
+//  2. The replica rebuilds successfully → volume has N+1 healthy replicas.
+//  3. cleanupExtraHealthyReplicas → cleanupAutoBalancedReplicas is called.
+//     - Before the fix, the cleanup logic deleted the sole replica in zone-b
+//     on the "unstable" node FIRST, causing auto-balance to immediately recreate it → infinite loop.
+//     - After the fix, cleanupReplicaInUnstableEnv deleted the extra replicas from
+//     overcrowded nodes or zones (the zone-a replicas, since zone-a has 2 while zone-b has 1) FIRST,
```
derekbit, line 529: Do we need to consider the disk level?
https://longhorn.io/docs/1.11.1/high-availability/auto-balance-replicas/#replica-auto-balance-disk-pressure-threshold-

controller/volume_controller.go
```diff
@@ -1279,115 +1320,162 @@ func (c *VolumeController) cleanupReplicaInUnstableEnv(v *longhorn.Volume, rs ma
...
-		rNames, err := c.getPreferredReplicaCandidatesForDeletion(rs)
-		if err != nil {
-			return false, err
+	if isDataLocalityDisabled(v) || !hasLocalReplicaOnSameNodeAsEngine(e, rs) {
+		return false, nil
+	}
+
+	// Only consider non-local replicas for deletion. The overcrowding
+	// analysis must exclude the local (engine-node) replicas, otherwise
+	// the engine node itself may be the most overcrowded group — yielding
```
derekbit, line 1452: Does the `group` mean `zone`?

derekbit (commented):
Hello @shuo-wu 
Could you elaborate on the root cause of the creation-deletion loop?

controller/volume_controller.go
```diff
@@ -2397,40 +2485,44 @@ func (c *VolumeController) getPreferredReplicaCandidatesForDeletion(rs map[strin
...
-	deletionCandidates = findValueWithBiggestLength(nodeToReplicaMap)
-	if len(deletionCandidates) > 1 {
-		return deletionCandidates, nil
-	}
-
-	// if all replicas are on different nodes, prefer to delete replicas on the same zone
-	deletionCandidates = findValueWithBiggestLength(zoneToReplicaMap)
-	if len(deletionCandidates) > 1 {
-		return deletionCandidates, nil
+	// Try disk > node > zone in order of priority.
```
derekbit, line 2503: Not fully understand the logic, why not `zone-> node->disk`?

controller/volume_controller.go
```diff
@@ -1279,115 +1320,162 @@ func (c *VolumeController) cleanupReplicaInUnstableEnv(v *longhorn.Volume, rs ma
...
+		return false, err
 	}
-
-	var rNames []string
-	if setting == longhorn.ReplicaAutoBalanceBestEffort {
-		_, rNames, _ = c.getReplicaCountForAutoBalanceBestEffort(v, e, rs, c.getReplicaCountForAutoBalanceNode)
-		if len(rNames) == 0 {
-			_, rNames, _ = c.getReplicaCountForAutoBalanceBestEffort(v, e, rs, c.getReplicaCountForAutoBalanceZone)
-		}
+	if len(preferredReplicaNames) == 0 {
```
derekbit, line 1408: Is it possible?
Should we error out if hitting this case?

controller/volume_controller.go
```diff
@@ -1279,115 +1320,162 @@ func (c *VolumeController) cleanupReplicaInUnstableEnv(v *longhorn.Volume, rs ma
...
+	// Step 4: If there is no replica on unstable nodes,
+	// fallback to delete the replica on the disk with the least available storage.
+	preferredReplicaNames, err = c.getSortedReplicasByAscendingStorageAvailable(preferredReplicaNames, rs)
 	if err != nil {
 		return false, err
 	}
-
-	r := rs[rNames[0]]
-	log.Infof("Deleting replica %v", r.Name)
+	r := rs[preferredReplicaNames[0]]
```
derekbit, line 1436: nit: check the length of `preferredReplicaNames` before accessing it.

c3y1huang (commented):
> The Bug3: cleanupReplicaInUnstableEnv cannot handle one node with multiple replicas

IIUC (Ref: https://suse.slack.com/archives/C0AR33VAF9N), the loop seems to be that Longhorn attempted to auto-balance the replica to another zone, and then that replica gets deleted because the node is evaluated unstable. So Bug3 doesn't seem to be the cause of the loop. 

For the fix, would it make sense to skip cleaning up the replica in `cleanupReplicaInUnstableEnv` if it becomes healthy (or created) after the node's last ready transition time? This is likely to be a good indication that the replica was rebuilt from a good source after the node is stable.

shuo-wu:
The manual reproducing step is available:
https://github.com/longhorn/longhorn/issues/12926#issuecomment-4284808262

controller/volume_controller.go
shuo-wu: It should be impossible but it's a safeguard. This is why I prefer to error out if it is triggered

controller/volume_controller.go
shuo-wu: `group` can stand for `zone`, `node`, or `disk`. Here, it means `node`

controller/volume_controller.go
shuo-wu: Since `disk` level redundancy:
- is typically meaningless. IMO, Multiple replicas on the same disk are a kind of space waste. It will greatly degrade the write performance since all replicas belonging to the same volume will fight for the disk bandwidth. Meanwhile, it cannot speed up the read performance (by asking multiple disks to read data simultaneously).
- provides the minimal availability/robustness: All replicas in a disk will be affected, no matter whether the disk is down, or the corresponding node or zone is down.

---
Let's take an example:
```
zone-a has 3 replicas: disk-11 on node-1 has **2 replicas**, disk-12 on node-1 has 1 replica;  disk-21 on node-2 has 0 replica
zone-b has 5 replica: disk-31 on node-3 has 1 replica; disk-41 on node-4 has 1 replica; disk-51 on node-5 has 1 replica; disk-61 on node-6 has 1 replica; disk-71 on node-7 has 1 replica
```
The 2 replicas in disk-11 are **in the biggest group** that is affected by the **whole 3-layer factors**: disk-11, node-1, and zone-a. 
In this example, if we handle `zone` first, we should move one replica from zone-b to zone-a (probably to disk-21 on node-2). I don't think this kind of movement is a good choice.

---

Sometimes, there are some vague cases in the node-level (for `best-effort`):
```
node-a has 2 replicas: disk-11 has 2 replicas; disk-12 has 0 replica
node-b has 4 replicas: disk-21 has 1 replica; disk-22 has 1 replica; disk-23 has 1 replica; disk-24 has 1 replica
```
If we handle the node-level candidates first, we probably remove one replica from node-b and reschedule it to disk-12 on node-a
If we handle the node-level candidates first, we probably remove one replica from disk-11 of node-a and reschedule it to disk-12 on node-a

I may need to double-check this case and see if there are any conflicts as well

shuo-wu:
After the discussion with @c3y1huang , we plan to close this refactor PR and launch the mininal fix first:

> skip cleaning up the replica in cleanupReplicaInUnstableEnv if it becomes healthy (or created) after the node's last ready transition time?

If `r.Spec.HealthyAt` is later than node `readyCondition.LastTransitionTime`, we will skip `cleanupReplicaInUnstableEnv` for it.

As for the refactor, this auto balance scheduling & cleanup is pretty complicated. It's better not to do so before the upcoming release v1.11.2. And the refactored solution itself should consider scheduling & cleanup in the same place before making the final decision, rather than keeping scheduling and cleanup logic separate.
Anyway, we will use a separate ticket for that.

PR #4658, opened 2026-04-20
fix(engine): verify completed rebuild to recover WO replica stuck

mschneider82:

Issue #9626
https://github.com/longhorn/longhorn/issues/9626

When a network interruption occurs during replica rebuilding, the gRPC connection to the sync agent can break while the file transfer (ssync) continues independently. In that case,reloadAndVerify is never called by startRebuilding and the replica stays in WO mode indefinitely after the transfer completes.                                                              
                                                                                                                                                                                            
This PR adds verifyCompletedRebuild to the engine monitor refresh loop, which detects replicas with state == "complete" and isRebuilding == false that are still in WO mode, and calls ReplicaRebuildVerify to transition them to RW. Verify errors are only logged (not returned) so the next poll cycle retries instead of setting the replica to ERR while the connection may still be flaky.

The fix mirrors the existing ReplicaRebuildVerify logic in syncWithRestoreStatus (restore path) but for the normal rebuild path, which was missing this recovery mechanism.               
                                                                                                                                                                                            
ReplicaRebuildVerify is idempotent-- if the replica is already RW (e.g. due to a concurrent call from startRebuilding), it returns immediately without side effects.


Related sparse-tools retry fix: https://github.com/longhorn/sparse-tools/pull/120 (the HTTP-level retry in ssync allows the file transfer to complete despite the network interruption,making this WO-stuck scenario reproducible)

### DIFF
--- a/controller/engine_controller.go
+++ b/controller/engine_controller.go
@@ -972,6 +972,11 @@ func (m *EngineMonitor) refresh(engine *longhorn.Engine) error {
 		}
 		engine.Status.RebuildStatus = rebuildStatus
 
+		// If a rebuild completed while the gRPC connection to the sync agent was interrupted,
+		// reloadAndVerify is never called and the replica stays in WO mode indefinitely.
+		// Detect this case and call ReplicaRebuildVerify to transition the replica to RW.
+		m.verifyCompletedRebuild(engine, addressReplicaMap, rebuildStatus, engineClientProxy)
+
 		// It's meaningless to sync the trim related field for old engines or engines in old engine instance managers
 		if cliAPIVersion >= 7 && im.Status.APIVersion >= 3 {
 			// Check and correct flag UnmapMarkSnapChainRemoved for the engine and replicas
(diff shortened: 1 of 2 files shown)


controller/engine_controller.go
```diff
@@ -1179,6 +1184,33 @@ func (m *EngineMonitor) refresh(engine *longhorn.Engine) error {
...
+
+		mode, exists := engine.Status.ReplicaModeMap[replicaName]
+		if !exists || mode != longhorn.ReplicaModeWO {
+			continue
+		}
+
+		m.logger.Infof("Replica %v completed rebuilding but is still in WO mode, calling ReplicaRebuildVerify", replicaName)
+		if err := engineClientProxy.ReplicaRebuildVerify(engine, replicaName, url); err != nil {
+			m.logger.WithError(err).Warnf("Failed to verify completed rebuild for replica %v, will retry on next poll", replicaName)
+		}
```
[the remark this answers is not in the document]
mschneider82: i think its better to see it, then silently going into debug log

controller/engine_controller.go
```diff
@@ -972,6 +972,11 @@ func (m *EngineMonitor) refresh(engine *longhorn.Engine) error {
 		}
 		engine.Status.RebuildStatus = rebuildStatus
 
+		// If a rebuild completed while the gRPC connection to the sync agent was interrupted,
+		// reloadAndVerify is never called and the replica stays in WO mode indefinitely.
+		// Detect this case and call ReplicaRebuildVerify to transition the replica to RW.
+		m.verifyCompletedRebuild(engine, addressReplicaMap, rebuildStatus, engineClientProxy)
```
derekbit, line 978: @mschneider82 
It sounds a very rare case. Did you hit the issue in your environment?

mschneider82: No it was discovered in the failed qa tests

derekbit:
@mschneider82 

go fmt error. Could you udpate it? Thanks.

```
148.8 Running: go vet
274.3 Running: golangci-lint
332.5 0 issues.
332.6 Running: go fmt
333.1 controller/engine_controller_test.go
```

davidcheng0922 (commented):
After addressing Derek's comment, overall LGTM.

mschneider82:
done

derekbit (approved):
LGTM. Thanks @mschneider82's contribution!

c3y1huang (approved):

PR #4659, opened 2026-04-21
fix: delete replica on unstable node only when its healthy time is earlier than node ready time

shuo-wu:

Issue https://github.com/longhorn/longhorn/issues/12926

When users enable `best-effort` auto balance, and the only node of a zone is an "unstable" node, volumes may get stuck in an infinite loop due to a bug in `cleanupReplicaInUnstableEnv`

The reproducing step: https://github.com/longhorn/longhorn/issues/12926#issuecomment-4284808262

The fix is from the suggestion: https://github.com/longhorn/longhorn-manager/pull/4647#pullrequestreview-4138707938

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -1250,12 +1250,17 @@ func (c *VolumeController) cleanupReplicaInUnstableEnv(v *longhorn.Volume, rs ma
 	// Pick up the replicas on the node that becomes ready most recently.
 	// Here we use kube node rather than Longhorn node because Longhorn node's ready condition will be affected by
 	// the corresponding longhorn-manager pod.
-	var chosenReplica, lastReadyReplica *longhorn.Replica
+	var chosenReplica *longhorn.Replica
+	var lastReadyReplicas []*longhorn.Replica
 	var lastReadyTransitionTime, secondLastReadyTransitionTime metav1.Time
-	nodeReplicaMap := make(map[string]*longhorn.Replica, len(rs))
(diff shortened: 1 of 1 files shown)


c3y1huang (commented):
Suggest improving the log message in https://github.com/longhorn/longhorn-manager/pull/4659#discussion_r3120276336. Otherwise, LGTM.

controller/volume_controller.go
```diff
@@ -1281,12 +1286,15 @@ func (c *VolumeController) cleanupReplicaInUnstableEnv(v *longhorn.Volume, rs ma
 		}
 		// In case of a replica being on not-ready node, chose it first.
 		if readyCondition.Status != corev1.ConditionTrue {
-			chosenReplica = nodeReplicaMap[node.Name]
-			break
+			if len(nodeReplicasMap[node.Name]) > 0 {
+				chosenReplica = nodeReplicasMap[node.Name][0]
```
derekbit, line 1290: We always delete the first replica in the map and don't care about the disk pressure. Will it fall into a deletion-creation loop if we enable `auto-balance with disk pressure feature` and `replica-soft-anti-affinity`?

shuo-wu: Here, the corresponding node is not ready. I don't think Longhorn will/should try to rebuild a new replica on any disks of the node.

c3y1huang (approved):

derekbit (approved):

derekbit:
@mergify backport v1.10.x v1.11.x

PR #4661, opened 2026-04-22
feat(setting): make staleReplicaTimeout globally configurable

sushant-suse:

[Related Issue](https://github.com/longhorn/longhorn/issues/4542)

It makes the `staleReplicaTimeout` configurable at a global level and introduces a configuration hierarchy for volume cleanup.

**Key Changes**:

- Added `stale-replica-timeout` as a registered setting in `longhorn-manager`.
- Updated the `VolumeController` to prioritize `Volume.Spec.StaleReplicaTimeout` if set; otherwise, it falls back to the new global setting.
- Modified the cleanup logic to ensure the `staleReplicaTimeout` respects the `Replica Replenishment Wait Interval`. The effective timeout is now the maximum of the two, preventing the premature deletion of replicas that the system is still attempting to reuse.
- Removed the hardcoded default (2880 minutes) from the CSI layer, allowing the controller to manage the default value through the global setting.

N/A

N/A

### DIFF
--- a/api/model.go
+++ b/api/model.go
@@ -1265,7 +1265,7 @@ func volumeSchema(volume *client.Schema) {
 
 	volumeStaleReplicaTimeout := volume.ResourceFields["staleReplicaTimeout"]
 	volumeStaleReplicaTimeout.Create = true
-	volumeStaleReplicaTimeout.Default = 2880
+	// volumeStaleReplicaTimeout.Default = 2880
 	volume.ResourceFields["staleReplicaTimeout"] = volumeStaleReplicaTimeout
 
 	volumeBackingImage := volume.ResourceFields["backingImage"]
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -1238,11 +1238,34 @@ func (c *VolumeController) cleanupReplicas(v *longhorn.Volume, es map[string]*lo
 
 // cleanupEngineFrontends cleans up EngineFrontends during volume deletion
 func (c *VolumeController) cleanupCorruptedOrStaleReplicas(v *longhorn.Volume, rs map[string]*longhorn.Replica) error {
-	// See comments for isSafeAsLastReplica for an explanation of why we call getSafeAsLastReplicaCount instead of
-	// getHealthyAndActiveReplicaCount here.
+	log := getLoggerForVolume(c.logger, v)
+
+	// 1. Convert our string constant to an int using strconv.Atoi
+	staleReplicaTimeoutSetting, err := strconv.Atoi(types.DefaultStaleReplicaTimeout)
+	if err != nil {
+		staleReplicaTimeoutSetting = 2880 // Absolute hardcoded fallback if parsing fails
+	}
+
+	// For the replenishment wait interval fallback, we use the standard default value (600 seconds)
+	replenishmentWaitIntervalSetting := 600
+
+	// 2. Try to fetch dynamic value from datastore, log warning and fallback if it fails
+	if dynamicTimeout, err := c.ds.GetSettingAsInt(types.SettingNameStaleReplicaTimeout); err != nil {
+		log.WithError(err).Warnf("Failed to get setting %v, falling back to default %v minutes",
+			types.SettingNameStaleReplicaTimeout, staleReplicaTimeoutSetting)
+	} else {
+		staleReplicaTimeoutSetting = int(dynamicTimeout) // Explicitly cast int64 to int
+	}
+
+	if dynamicInterval, err := c.ds.GetSettingAsInt(types.SettingNameReplicaReplenishmentWaitInterval); err != nil {
+		log.WithError(err).Warnf("Failed to get setting %v, falling back to default %v seconds",
+			types.SettingNameReplicaReplenishmentWaitInterval, replenishmentWaitIntervalSetting)
+	} else {
+		replenishmentWaitIntervalSetting = int(dynamicInterval) // Explicitly cast int64 to int
+	}
+
 	safeAsLastReplicaCount := getSafeAsLastReplicaCount(rs)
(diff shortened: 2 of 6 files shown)


derekbit:
Thanks @sushant-suse. The PR will be reviewed after creating v1.12.x branch.

derekbit:
Hello @sushant-suse CI failed.

image

api/model.go
```diff
@@ -1265,7 +1265,7 @@ func volumeSchema(volume *client.Schema) {
 
 	volumeStaleReplicaTimeout := volume.ResourceFields["staleReplicaTimeout"]
 	volumeStaleReplicaTimeout.Create = true
-	volumeStaleReplicaTimeout.Default = 2880
+	// volumeStaleReplicaTimeout.Default = 2880
```
derekbit, line 1268: Why do we comment out this line?

controller/volume_controller_test.go
```diff
@@ -19,18 +21,22 @@ import (
...
 	imutil "github.com/longhorn/longhorn-instance-manager/pkg/util"
 
+	// Category 9: this_repo_no_alias
 	"github.com/longhorn/longhorn-manager/datastore"
 	"github.com/longhorn/longhorn-manager/types"
 	"github.com/longhorn/longhorn-manager/util"
 
+	// Category 10: this_repo_alias
 	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
 	lhfake "github.com/longhorn/longhorn-manager/k8s/pkg/client/clientset/versioned/fake"
```
derekbit: Remove `// Category ....`

types/setting.go
```diff
@@ -1101,6 +1104,23 @@ var (
...
+		Description: "In minutes. The timeout determines how long Longhorn will wait before cleaning up a stale replica " +
+			"that is no longer part of the volume engine.\n" +
+			"The effective cleanup delay is the greater of this setting and the Replica Replenishment Wait Interval " +
+			"(converted to minutes and rounded up). Setting this to 0 allows cleanup as soon as that effective delay is reached.",
+		Category:           SettingCategoryGeneral,
+		Type:               SettingTypeInt,
+		Required:           true,
+		ReadOnly:           false,
+		DataEngineSpecific: false,
+		Default:            "2880",
```
derekbit: Define a const `defaultStaleReplicaTimeout` for 2880

```suggestion
		Default:            defaultStaleReplicaTimeout,
```

controller/volume_controller.go
```diff
@@ -1189,6 +1189,17 @@ func (c *VolumeController) cleanupReplicas(v *longhorn.Volume, es map[string]*lo
...
+	// Fetch settings once here to pass into shouldCleanUpFailedReplica (O(1) vs O(N) optimization)
+	staleReplicaTimeoutSetting, err := c.ds.GetSettingAsInt(types.SettingNameStaleReplicaTimeout)
+	if err != nil {
+		return errors.Wrapf(err, "failed to get setting %v for cleaning up corrupted or stale replicas", types.SettingNameStaleReplicaTimeout)
+	}
+
+	replenishmentWaitIntervalSetting, err := c.ds.GetSettingAsInt(types.SettingNameReplicaReplenishmentWaitInterval)
+	if err != nil {
+		return errors.Wrapf(err, "failed to get setting %v for cleaning up corrupted or stale replicas", types.SettingNameReplicaReplenishmentWaitInterval)
+	}
```
derekbit: How about 

```suggestion
    staleReplicaTimeout := strconv.Atoi(defaultStaleReplicaTimeout)
	staleReplicaTimeoutSetting, err := c.ds.GetSettingAsInt(types.SettingNameStaleReplicaTimeout)
	if err != nil {
		c.WithError(err).Warnf(.....)
	}

     replenishmentWaitInterval := strconv.Atoi(defaultReplenishmentWaitInterval)
	replenishmentWaitIntervalSetting, err := c.ds.GetSettingAsInt(types.SettingNameReplicaReplenishmentWaitInterval)
	if err != nil {
		c.WithError(err).Warnf(.....)
	}
```

We can fall back to the default values rather than error out.

controller/volume_controller.go
```diff
@@ -6465,12 +6477,30 @@ func (c *VolumeController) shouldCleanUpFailedReplica(v *longhorn.Volume, r *lon
...
+	// This ensures we never clean up a replica that the replenishment logic is still considering.
+	replenishmentWaitInMinutes := replenishmentWaitIntervalSetting / 60
+	if replenishmentWaitIntervalSetting > 0 && replenishmentWaitIntervalSetting%60 != 0 {
+		replenishmentWaitInMinutes++
+	}
+
+	if timeout < replenishmentWaitInMinutes {
+		timeout = replenishmentWaitInMinutes
+	}
+
```
derekbit, line 6514: Could you explain why we handle replenishment here?

api/model.go
sushant-suse: We commented this out because leaving `volumeStaleReplicaTimeout.Default = 2880` at the API schema layer completely prevents the new Global Setting from taking effect on newly created volumes.

controller/volume_controller.go
sushant-suse: We handle it here as a safety guardrail to prevent a race condition between the cleanup logic and the replenishment logic.

If a user sets a very low `StaleReplicaTimeout` (or `0` for immediate cleanup), the controller could delete a failed replica while the replenishment logic is still actively waiting out its `Replica Replenishment Wait Interval` to decide on a rebuild.

Deleting the replica prematurely destroys its state and `FailedAt` timestamp, which disrupts the replenishment tracking. Taking the `max()` of both values ensures we never clean up a replica that the replenishment subsystem is still actively considering.

PR #4662, opened 2026-04-23
fix: check if BackupVolume.Status.LastBackupName is updated correctly

mantissahz:

Issue # longhorn/longhorn#12542

Run the e2e test case 101 times:
https://ci.longhorn.io/job/private/job/longhorn-e2e-test/7535/
https://ci.longhorn.io/job/private/job/longhorn-e2e-test/7547/

### DIFF
--- a/controller/backup_volume_controller.go
+++ b/controller/backup_volume_controller.go
@@ -399,10 +399,12 @@ func (bvc *BackupVolumeController) reconcile(backupVolumeName string) (err error
 		return nil
 	}
 
-	// If there is no backup CR creation/deletion and the backup volume config metadata not changed
+	// If there is no backup CR creation/deletion, the correct backup in the cluster is presented at the last backup of the backup volume,
+	// and the backup volume config metadata not changed
 	// skip read the backup volume config
 	if len(backupsToPull) == 0 && len(backupsToDelete) == 0 &&
-		backupVolume.Status.LastModificationTime.Time.Equal(configMetadata.ModificationTime) {
+		backupVolume.Status.LastModificationTime.Time.Equal(configMetadata.ModificationTime) &&
+		bvc.isLastBackupUpdateSafeToSkip(clusterBackups, backupStoreBackups, backupVolume) {
 		backupVolume.Status.LastSyncedAt = syncTime
 		return nil
 	}
@@ -455,6 +457,60 @@ func (bvc *BackupVolumeController) reconcile(backupVolumeName string) (err error
 	backupVolume.Status.StorageClassName = backupVolumeInfo.StorageClassName
 	backupVolume.Status.LastSyncedAt = syncTime
 	return nil
+}
+
+// isLastBackupUpdateSafeToSkip checks if it's safe to skip updating the last backup related status of the backup volume
+// when the backup volume config metadata is not changed.
+//
+//	clusterBackups: the backup CRs in the cluster with the same backup target and backup volume, which are expected to be consistent with the backup list in the backup target.
+//	backupStoreBackups: the backup names in the backup target, which are expected to be consistent with the backup CRs in the cluster.
+//
+//nolint:staticcheck
+func (bvc *BackupVolumeController) isLastBackupUpdateSafeToSkip(clusterBackups map[string]*longhorn.Backup, backupStoreBackups sets.String, bv *longhorn.BackupVolume) bool {
+	if backupStoreBackups.Len() == 0 {
+		// If there is no backup in the backup target, the last backup name should be empty.
+		return bv.Status.LastBackupName == ""
+	}
+
+	// backupStoreBackups is not empty and the last backup name should not be empty.
+	if bv.Status.LastBackupName == "" {
+		return false
+	}
+
+	// Last backup of the backup volume is deleted.
+	if !backupStoreBackups.Has(bv.Status.LastBackupName) {
+		return false
+	}
+
+	// check if there is any backup which is created after the last backup of the backup volume, and the backup CR is not updated with the backup created time yet.
+	lastBackupCreateAt := time.Time{}
+	lastCreatedAndCompletedBackupName := ""
+	for backupName, backup := range clusterBackups {
+		if !backupStoreBackups.Has(backupName) || backup.Status.State != longhorn.BackupStateCompleted {
+			continue
+		}
+
+		if bv.Status.LastBackupAt == "" {
+			return false
+		}
+
+		backupCreatedAt, err := util.ParseTimeZ(backup.Status.BackupCreatedAt)
+		if backup.Status.BackupCreatedAt == "" || err != nil {
+			bvc.logger.WithError(err).Debugf("Failed to parse backup created time for backup %s", backup.Name)
+			return false
+		}
+
+		if backupCreatedAt.After(lastBackupCreateAt) {
+			lastBackupCreateAt = backupCreatedAt
+			lastCreatedAndCompletedBackupName = backupName
+		}
+	}
+
+	if lastCreatedAndCompletedBackupName != bv.Status.LastBackupName {
+		return false
+	}
+
+	return true
 }
 
 func (bvc *BackupVolumeController) isResponsibleFor(bv *longhorn.BackupVolume, defaultEngineImage string) (bool, error) {


controller/backup_volume_controller.go
```diff
@@ -401,7 +401,9 @@ func (bvc *BackupVolumeController) reconcile(backupVolumeName string) (err error
 
 	// If there is no backup CR creation/deletion and the backup volume config metadata not changed
 	// skip read the backup volume config
-	if len(backupsToPull) == 0 && len(backupsToDelete) == 0 &&
+	isLastBackupUpdated := ((len(clustersSet) != 0 && len(backupsToDelete) == 0 && backupVolume.Status.LastBackupName != "") ||
+		(len(clustersSet) == 0 && len(backupsToPull) == 0 && backupVolume.Status.LastBackupName == ""))
```
derekbit: It is quite complicated. Can we create a function for handling the conditions?

mantissahz: Modified to create a separate function. Thanks.

derekbit:
@mantissahz Before reviewing the PR, could you elaborate on the root cause of the ticket and the solution?

mantissahz:
> Before reviewing the PR, could you elaborate on the root cause of the ticket and the solution?

@derekbit, here is the RCA: https://github.com/longhorn/longhorn/issues/12542#issuecomment-4310360476

derekbit:
cc @carterli0407-cell Please help review the PR. Thanks.

controller/backup_volume_controller.go
```diff
@@ -457,6 +458,20 @@ func (bvc *BackupVolumeController) reconcile(backupVolumeName string) (err error
 	return nil
 }
 
+//nolint:staticcheck
```
derekbit, line 468: Why do we need the check?

derekbit (commented):
The fix is applied to v1 and v2 data engine simultaneously. Does the issue happen in v1 volume? cc @yangchiu

controller/backup_volume_controller.go
mantissahz: 'set.String' is marked as deprecated

controller/backup_volume_controller.go
```diff
@@ -457,6 +458,20 @@ func (bvc *BackupVolumeController) reconcile(backupVolumeName string) (err error
...
+func isLastBackupUpdated(clustersSet, backupsToDelete, backupsToPull sets.String, backupVolume *longhorn.BackupVolume) bool {
+	// If there are some backups in the cluster and no backup is deleted, the last backup name should not be empty.
+	if len(clustersSet) != 0 && len(backupsToDelete) == 0 && backupVolume.Status.LastBackupName != "" {
+		return true
+	}
+	// If there is no backup in the cluster and no backup is pulled, the last backup name should be empty.
+	if len(clustersSet) == 0 && len(backupsToPull) == 0 && backupVolume.Status.LastBackupName == "" {
+		return true
```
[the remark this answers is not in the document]
derekbit: @mantissahz Could you check the comment? I think it is a valid case.

controller/backup_volume_controller.go
```diff
@@ -457,6 +458,20 @@ func (bvc *BackupVolumeController) reconcile(backupVolumeName string) (err error
...
+	}
+
+	return false
+}
```
derekbit, line 514: The original solution might be problematic when the condition is true
```
	if len(clustersSet) != 0 && len(backupsToDelete) == 0 && backupVolume.Status.LastBackupName != "" {
		return true
	}
```

It might result in that the `bv.Status.LastBackupName` has a outdated value. For example
1. backup-A exists，`BackupVolume.Status.LastBackupName` = backup-A
2. New backup-B is created and completed
3. backup_controller updates `BackupVolume.Spec.SyncRequestedAt`
4. backup_volume_controller reconcile
5. backup list already contains backup-B, no pull/delete operations, metadata timestamp is not changed
6. Because `LastBackupName != ""`, the solution will skip `BackupVolumeGet()`
7. `BackupVolume.Status.LastBackupName` won't be updated to backup-A


What we need is to ensure `Volume.status.lastBackup` records the latest backup.

How about replacing the function with 
```
func isLastBackupStatusSafeToSkip(clusterBackups map[string]*longhorn.Backup, backupStoreBackups sets.String, bv *longhorn.BackupVolume) bool {
    if backupStoreBackups.Len() == 0 {
        return bv.Status.LastBackupName == ""
    }

    if bv.Status.LastBackupName == "" || !backupStoreBackups.Has(bv.Status.LastBackupName) {
        return false
    }

    lastBackupAt, err := time.Parse(time.RFC3339, bv.Status.LastBackupAt)
    if bv.Status.LastBackupAt != "" && err != nil {
        return false
    }

    for backupName, backup := range clusterBackups {
        if backupName == bv.Status.LastBackupName ||
            !backupStoreBackups.Has(backupName) ||
            backup.Status.State != longhorn.BackupStateCompleted {
            continue
        }

        // Newly completed local backup; BackupVolume must inspect volume.cfg.
        if backup.Status.LastSyncedAt.IsZero() {
            return false
        }

        backupCreatedAt, err := time.Parse(time.RFC3339, backup.Status.BackupCreatedAt)
        if backup.Status.BackupCreatedAt == "" || err != nil {
            return false
        }

        if bv.Status.LastBackupAt == "" || backupCreatedAt.After(lastBackupAt) {
            return false
        }
    }

    return true
}
```

controller/backup_volume_controller.go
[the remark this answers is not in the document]
mantissahz: Modified.

controller/backup_volume_controller.go
mantissahz: Modified as suggested.
Thanks.

controller/backup_volume_controller.go
```diff
@@ -402,6 +402,7 @@ func (bvc *BackupVolumeController) reconcile(backupVolumeName string) (err error
 	// If there is no backup CR creation/deletion and the backup volume config metadata not changed
 	// skip read the backup volume config
```
derekbit: Please update the outdated comment.

controller/backup_volume_controller.go
```diff
@@ -457,6 +458,45 @@ func (bvc *BackupVolumeController) reconcile(backupVolumeName string) (err error
 	return nil
 }
 
+//nolint:staticcheck
```
derekbit, line 468: Add a comment to explain the function.

controller/backup_volume_controller.go
```diff
@@ -457,6 +458,45 @@ func (bvc *BackupVolumeController) reconcile(backupVolumeName string) (err error
...
+func isLastBackupUpdateSafeToSkip(clusterBackups map[string]*longhorn.Backup, backupStoreBackups sets.String, bv *longhorn.BackupVolume) bool {
+	if backupStoreBackups.Len() == 0 {
+		// If there is no backup in the backup target, the last backup name should be empty.
+		return bv.Status.LastBackupName == ""
+	}
+
+	// backupStoreBackups is not empty.
+	if bv.Status.LastBackupName == "" || !backupStoreBackups.Has(bv.Status.LastBackupName) {
+		return false
+	}
```
derekbit: nit: suggest separating the two conditions. 

```suggestion
	if bv.Status.LastBackupName == "" {
		return false
	}

	if !backupStoreBackups.Has(bv.Status.LastBackupName) {
	    return false
	}
```

controller/backup_volume_controller.go
```diff
@@ -457,6 +458,45 @@ func (bvc *BackupVolumeController) reconcile(backupVolumeName string) (err error
...
+
+	bvLastBackupAt, err := util.ParseTimeZ(bv.Status.LastBackupAt)
+	if bv.Status.LastBackupAt != "" && err != nil {
+		return false
```
derekbit, line 482: Can we add a debug log message here if "err != nil"?

controller/backup_volume_controller.go
```diff
@@ -457,6 +458,45 @@ func (bvc *BackupVolumeController) reconcile(backupVolumeName string) (err error
 	return nil
 }
 
+//nolint:staticcheck
+func isLastBackupUpdateSafeToSkip(clusterBackups map[string]*longhorn.Backup, backupStoreBackups sets.String, bv *longhorn.BackupVolume) bool {
```
derekbit: What's the difference between `clusterBackups` and `backupStoreBackups`? It is better to explain in the function mentioned above.

controller/backup_volume_controller.go
```diff
@@ -457,6 +458,45 @@ func (bvc *BackupVolumeController) reconcile(backupVolumeName string) (err error
...
+	}
+
+	for backupName, backup := range clusterBackups {
+		if backupName == bv.Status.LastBackupName || !backupStoreBackups.Has(backupName) || backup.Status.State != longhorn.BackupStateCompleted {
+			continue
+		}
+		// Newly completed local backup; BackupVolume must inspect volume.cfg.
+		if backup.Status.LastSyncedAt.IsZero() {
+			return false
+		}
```
derekbit: Why do we return `false` when `backup.Status.LastSyncedAt` is zero?

controller/backup_volume_controller.go
```diff
@@ -457,6 +458,45 @@ func (bvc *BackupVolumeController) reconcile(backupVolumeName string) (err error
...
+
+		backupCreatedAt, err := util.ParseTimeZ(backup.Status.BackupCreatedAt)
+		if backup.Status.BackupCreatedAt == "" || err != nil {
```
derekbit, line 498: Add a debug message if err is not nil

controller/backup_volume_controller.go
mantissahz: Updated. Thanks.

controller/backup_volume_controller.go
mantissahz: Added. Thanks.

controller/backup_volume_controller.go

controller/backup_volume_controller.go
mantissahz: Added. Thanks.

controller/backup_volume_controller.go
mantissahz: Removed this check and refactored the loop logic here.
Thanks.

controller/backup_volume_controller.go
mantissahz: Added. Thanks.

controller/backup_volume_controller.go
mantissahz: Removed to get `bvLastBackupAt . Thanks

derekbit: > Removed to get `bvLastBackupAt . Thanks

Don't we need it?

mantissahz: `BackupVolume.Status.bvLastBackupAt` might always be before `Backup.Status.BackupCreateAt`, and new logic will check all backups' `BackupCreateAt` to get the latest one and compare it with `BackupVolume.Status.LastBackupName`.
Therefore, we don't use this field.

derekbit (approved):

PR #4676, opened 2026-04-28
feat: v2 live upgrade

davidcheng0922:

https://github.com/longhorn/longhorn/issues/9104

Implement detail in LEP: https://github.com/longhorn/longhorn/pull/13214


controller/node_controller.go
```diff
@@ -1098,6 +1144,64 @@ func (nc *NodeController) cleanupAllReplicaManagers(node *longhorn.Node) error {
...
+	}
+
+	for _, vol := range volumes {
+		if !types.IsDataEngineV2(vol.Spec.DataEngine) {
+			continue
+		}
+		if vol.Status.CurrentEngineNodeID == node.Name {
+			return true, nil
+		}
+	}
```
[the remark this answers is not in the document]
davidcheng0922: We intentionally use `volume.Status.CurrentEngineNodeID` here because this decision is about the volume-level current-engine owner, not just whether a node happens to have running Engine CRs. 

During live-upgrade and switchover there can be transient states where engines exist on multiple nodes, but only one node is the authoritative current engine for the volume. `CurrentEngineNodeID` is the stronger correctness signal for deciding whether the old v2 IM must still be preserved, so we prefer it over the more efficient node-scoped engine lookup here.

types/types.go
```diff
@@ -1471,3 +1479,15 @@ func GetBackingImageMonitorName(imName string) string {
...
+func IsActiveInstanceManagerUpgradeState(state longhorn.InstanceManagerUpgradeState) bool {
+	switch state {
+	case longhorn.InstanceManagerUpgradeStateRelocatingEngines,
+		longhorn.InstanceManagerUpgradeStateWaitingForSourceIM,
+		longhorn.InstanceManagerUpgradeStateRestoringEngines,
+		longhorn.InstanceManagerUpgradeStateWaitingForHealthyVolumes:
+		return true
+	default:
+		return false
+	}
```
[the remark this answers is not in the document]
davidcheng0922: We intentionally keep Pending out of `IsActiveInstanceManagerUpgradeState()`.

We do have `Pending` cases that are already in a timed wait (`StartedAt` is set), but we do not want every `Pending` IMU to be treated as an externally active live-upgrade window. Some `Pending` IMUs have not started any data-path-affecting work yet, and widening `hasActiveInstanceManagerUpgradeOnNode()` to include all `Pending` states would start attach/detach and EF toleration logic too early.

So the current handling is:

- `Pending` is considered in-progress only for IMU internal timeout/requeue handling once `StartedAt` is set
- external live-upgrade checks still begin only after the IMU reaches the relocation/restore phases

This keeps the waiting `Pending` cases covered without making every pre-start Pending IMU look like an active live upgrade.

davidcheng0922:
e2e CI: https://github.com/longhorn/longhorn/issues/9104#issuecomment-4665980779

controller/instance_manager_controller.go
```diff
@@ -200,6 +200,7 @@ func NewInstanceManagerController(
 		return nil, err
 	}
 	imc.cacheSyncs = append(imc.cacheSyncs, ds.SettingInformer.HasSynced)
+	imc.cacheSyncs = append(imc.cacheSyncs, ds.InstanceManagerUpgradeControlInformer.HasSynced)
```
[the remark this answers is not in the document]
davidcheng0922: Fixed

controller/instance_manager_upgrade_control_controller.go
```diff
@@ -0,0 +1,814 @@
...
+}
+
+func (c *InstanceManagerUpgradeControlController) syncIMUC(key string) (err error) {
+	namespace, name, err := cache.SplitMetaNamespaceKey(key)
+	if err != nil {
+		return err
+	}
+	if namespace != c.namespace {
+		return nil
+	}
```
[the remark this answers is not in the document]
davidcheng0922: Fixed

davidcheng0922:
SPDK changed the type of `DmaDeviceType` from `int32` to `string`, which is a breaking change in the data structure:
https://github.com/longhorn/go-spdk-helper/commit/9b45c1d11863b56e557aeb1db0c369a0c743136e

As a result, an old engine (`v1.12.1`) can't rebuild a replica running on a new-version instance manager. The engine has to access the new replica during rebuilding, and the type mismatch makes the JSON unmarshal fail.

So live upgrade from `v1.12.1` currently doesn't work — rebuilding can't complete correctly.


---------------------------------------------------------------------------------------------

**Alternative approach:**

- Allow live upgrade to proceed for degraded volumes.
- Prevent an old-version engine from rebuilding a replica on a new-version instance manager.

Scenario:

There are two nodes, `nodeA` and `nodeB`, and two volumes, `volumeA` and `volumeB`.
Each volume has two replicas.

After `nodeA` is upgraded, `volumeB` may remain degraded because the old-version
engine on nodeB cannot rebuild a replica on nodeA's new-version instance manager.

Before upgrading nodeB:

1. Switch `volumeB's engine` to `nodeA`.
2. Wait until the engine is running on `nodeA`.
3. For a degraded volume, rebuild the replica on `nodeA` using the new-version
   engine, and wait until rebuilding completes.
4. Upgrade `nodeB`.
5. Switch `volumeB's engine` back to `nodeB`.
6. Wait until `volumeB` becomes healthy.

For a healthy volume:

1. Switch the volume's engine to `nodeA`.
2. Wait until the engine switch completes and the volume remains healthy.
3. Upgrade `nodeB`.
4. Switch the volume's engine back to `nodeB`.
5. Wait until the volume becomes healthy.

scheduler/replica_scheduler.go
```diff
@@ -246,6 +246,23 @@ func (rcs *ReplicaScheduler) buildLinkedCloneSrcNodeDiskMap(replica *longhorn.Re
 func (rcs *ReplicaScheduler) getNodeCandidates(nodes map[string]*longhorn.Node, schedulingReplica *longhorn.Replica, volume *longhorn.Volume) (nodeCandidates map[string]*longhorn.Node, errs multierr.MultiError) {
 	errs = multierr.NewMultiError()
 
+	nodesUnderUpgrade := map[string]bool{}
+	if types.IsDataEngineV2(schedulingReplica.Spec.DataEngine) {
+		imuList, err := rcs.ds.ListInstanceManagerUpgradesRO()
+		if err != nil {
+			logrus.WithError(err).Warnf("Failed to list instance manager upgrades while scheduling replica %v, continuing without upgrade-node exclusion", schedulingReplica.Name)
+		} else {
```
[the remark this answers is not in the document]
davidcheng0922: Fix it

controller/node_controller.go
```diff
@@ -1202,6 +1208,8 @@ func (nc *NodeController) syncInstanceManagers(node *longhorn.Node) error {
 					} else {
 						log.Infof("Cleaning up instance manager %v because node %v does not match %v", im.Name, node.Name, types.SettingNameSystemManagedComponentsNodeSelector)
 					}
+				} else if preserveV2AIOInstanceManagers {
+					cleanupRequired = false
```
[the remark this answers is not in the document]
davidcheng0922: Fix it

controller/instance_manager_upgrade_controller.go
```diff
@@ -0,0 +1,1657 @@
...
+}
+
+func (imuc *InstanceManagerUpgradeController) enforceUpgradeTimeout(
+	imu *longhorn.InstanceManagerUpgrade,
+	log *logrus.Entry,
+) (bool, error) {
+	if (imu.Status.State != longhorn.InstanceManagerUpgradeStatePending &&
+		!types.IsActiveInstanceManagerUpgradeState(imu.Status.State)) || imu.Status.StartedAt == "" {
+		return false, nil
+	}
```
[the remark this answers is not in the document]
davidcheng0922: Fix it

controller/instance_manager_upgrade_control_controller.go
```diff
@@ -0,0 +1,823 @@
...
+			Name: imuName,
+		},
+		Spec: longhorn.InstanceManagerUpgradeSpec{
+			NodeID:      nodeID,
+			TargetImage: targetImage,
+		},
+	}
+	if _, err := c.ds.CreateInstanceManagerUpgrade(imu); err != nil {
+		if apierrors.IsAlreadyExists(err) {
+			return imuName, nil
```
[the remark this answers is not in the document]
davidcheng0922: Fix it

controller/instance_manager_controller.go
```diff
@@ -2780,6 +2896,98 @@ func (m *InstanceManagerMonitor) syncOrphans(im *longhorn.InstanceManager, insta
...
+		return nil
+	}
+
+	log := getLoggerForInstanceManager(imc.logger, im)
+
+	allowed, err := imc.ds.GetSettingAsBool(types.SettingNameAllowV2InstanceManagerAutomaticUpgrade)
+	if err != nil {
+		return err
+	}
+	if !allowed {
```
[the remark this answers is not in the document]
davidcheng0922: Fix it

scheduler/replica_scheduler.go
```diff
@@ -246,6 +246,23 @@ func (rcs *ReplicaScheduler) buildLinkedCloneSrcNodeDiskMap(replica *longhorn.Re
...
+			for _, imu := range imuList {
+				if imu.Status.State == longhorn.InstanceManagerUpgradeStateRelocatingEngines ||
+					imu.Status.State == longhorn.InstanceManagerUpgradeStateWaitingForSourceIM ||
+					imu.Status.State == longhorn.InstanceManagerUpgradeStateRestoringEngines {
```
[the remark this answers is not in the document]
davidcheng0922: Fix it

controller/volume_controller.go
```diff
@@ -1230,6 +1263,43 @@ func (c *VolumeController) cleanupReplicas(v *longhorn.Volume, es map[string]*lo
...
+	}
+	for _, imu := range imus {
+		switch imu.Status.State {
+		case longhorn.InstanceManagerUpgradeStatePending:
+			if len(imu.Status.Engines) != 0 {
+				continue
+			}
+		case longhorn.InstanceManagerUpgradeStateRelocatingEngines:
+			if _, ok := imu.Status.Engines[v.Name]; !ok {
+				continue
```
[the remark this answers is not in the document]
davidcheng0922: Fix it

controller/instance_manager_upgrade_control_controller.go
```diff
@@ -0,0 +1,823 @@
...
+	ims, err := c.ds.ListInstanceManagersRO()
+	if err != nil {
+		c.logger.WithError(err).Warn("Failed to list instance managers while picking next node")
+	} else {
+		for _, im := range ims {
+			if !types.IsDataEngineV2(im.Spec.DataEngine) || im.Spec.Type != longhorn.InstanceManagerTypeAllInOne {
+				continue
+			}
+			if im.Spec.Image == imuc.Spec.TargetImage {
+				continue
```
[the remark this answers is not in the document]
davidcheng0922: Fix it

controller/instance_manager_upgrade_controller.go
```diff
@@ -0,0 +1,1657 @@
...
+		if err == nil && !reflect.DeepEqual(existingStatus, &imu.Status) {
+			if _, updateErr := imuc.ds.UpdateInstanceManagerUpgradeStatus(imu); updateErr != nil {
+				err = updateErr
+			}
+		}
+	}()
+
+	if imu.DeletionTimestamp != nil {
+		imuc.eventRecorder.Eventf(imu, corev1.EventTypeWarning, constant.EventReasonDelete, "Deleting instance manager upgrade %v", imu.Name)
+		return imuc.ds.RemoveFinalizerForInstanceManagerUpgrade(imu)
```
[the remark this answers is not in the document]
davidcheng0922: Fix it

controller/instance_manager_upgrade_controller.go
```diff
@@ -0,0 +1,1774 @@
...
+
+		diskStatus, ok := node.Status.DiskStatus[diskName]
+		if !ok {
+			continue
+		}
+
+		condition := types.GetCondition(diskStatus.Conditions, longhorn.DiskConditionTypeSchedulable)
+		if condition.Status == longhorn.ConditionStatusTrue {
+			return true, nil
+		}
```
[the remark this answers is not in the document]
davidcheng0922: No, `DiskConditionTypeSchedulable` is correct.

controller/volume_controller.go
```diff
@@ -1230,6 +1266,40 @@ func (c *VolumeController) cleanupReplicas(v *longhorn.Volume, es map[string]*lo
...
+	if err != nil {
+		return nil, err
+	}
+	for _, imu := range imus {
+		switch imu.Status.State {
+		case longhorn.InstanceManagerUpgradeStatePending:
+			if len(imu.Status.Engines) != 0 {
+				continue
+			}
+		case longhorn.InstanceManagerUpgradeStateRelocatingEngines:
```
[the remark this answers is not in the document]
davidcheng0922: This is intentional. RelocatingEngines also handles replica-only detach plans. We must suppress all planned detached replicas to prevent volume reconciliation from rebuilding them before the source IM restart.

controller/instance_manager_upgrade_controller.go
```diff
@@ -0,0 +1,1798 @@
...
+				delete(imu.Status.PlannedDetachedReplicas, volumeName)
+				continue
+			}
+			return err
+		}
+
+		if volume.Status.Robustness != longhorn.VolumeRobustnessHealthy {
+			log.Debugf("Volume %v not yet healthy after planned replica detach (robustness: %v), waiting",
+				volumeName, volume.Status.Robustness)
+			allHealthy = false
```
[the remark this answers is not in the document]
davidcheng0922: This is intentional. The live upgrade flow assumes that the old engine can rebuild replicas on the upgraded instance manager. If that compatibility requirement is not met, the upgrade path is unsupported and should fail rather than bypass the healthy-volume gate.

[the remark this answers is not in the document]
davidcheng0922: Fully supporting this scenario requires a separate compatibility strategy that allows degraded volumes to be upgraded safely.

[Analysis](https://github.com/longhorn/longhorn-manager/pull/4676#issuecomment-5519729754)

controller/instance_manager_upgrade_controller.go
```diff
@@ -0,0 +1,1956 @@
...
+					"Volume %v deleted during relocation, removing from upgrade plan", volumeName)
+				delete(imu.Status.Engines, volumeName)
+				continue
+			}
+			return err
+		}
+
+		// Already running on the temporary node — done for this engine.
+		if volume.Status.CurrentEngineNodeID == reloc.TemporaryNodeID {
+			continue
```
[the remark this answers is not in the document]
davidcheng0922: Temporary node failure is expected to cause application downtime. Live upgrade doesn't change this: the volume depends on the instance manager that hosts its engine, so if that instance manager fails, the volume becoming unavailable is the expected behavior.

controller/instance_manager_upgrade_controller.go
```diff
@@ -0,0 +1,1965 @@
...
+	for _, volume := range volumes {
+		if volume.DeletionTimestamp != nil || volume.Status.State != longhorn.VolumeStateAttached {
+			continue
+		}
+		if volume.Status.CurrentNodeID != nodeID && volume.Status.CurrentEngineNodeID != nodeID {
+			continue
+		}
+		if volume.Spec.Frontend == longhorn.VolumeFrontendUblk {
+			return fmt.Errorf("%w: node %v has ublk frontend volume %v, which is not supported for live upgrade",
+				errUpgradeUnsupported, nodeID, volume.Name)
```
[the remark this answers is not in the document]
davidcheng0922: I'd rather not. Extending the guard to replica placement means a single UBLK volume can block the upgrade on every node that holds one of its replicas, which is too broad a blast radius.

controller/instance_manager_upgrade_controller.go
```diff
@@ -0,0 +1,1965 @@
...
+		if volume.Status.CurrentEngineNodeID == reloc.OriginalNodeID {
+			continue
+		}
+		sourceReady, err := imuc.isSourceIMReadyForRestore(imu.Spec.NodeID)
+		if err != nil {
+			return err
+		}
+		if !sourceReady {
+			log.Debugf("Waiting for source IM on node %v to become ready before restoring volume %v", imu.Spec.NodeID, volumeName)
+			allRestored = false
```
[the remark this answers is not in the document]
davidcheng0922: For this case, we intentionally take the conservative approach: keep the IMU in the restoring state and wait for the source IM to become ready. We do not roll the source IM back because the engine frontend remains on the source node and restoring engines elsewhere could introduce additional risk.

The only exception is when the source node is confirmed deleted. In that case, restoration cannot succeed, so we terminalize the IMU, skip retries for that node, and allow the rolling upgrade to continue with the next node.

davidcheng0922:
https://ci.longhorn.io/job/private/job/longhorn-e2e-test/9488/

datastore/longhorn.go
```diff
@@ -665,6 +666,23 @@ func (s *DataStore) ValidateSetting(name, value string) (err error) {
...
+		}
+
+		currentVersionSetting, err := s.GetSettingExactRO(types.SettingNameCurrentLonghornVersion)
+		if err != nil {
+			if ErrorIsNotFound(err) {
+				break
+			}
+			return errors.Wrap(err, "failed to get current Longhorn version")
+		}
+		if currentVersionSetting.Value != "" && semver.Compare(currentVersionSetting.Value, types.MinimumLonghornVersionForV2InstanceManagerLiveUpgrade) < 0 {
```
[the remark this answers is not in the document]
davidcheng0922: Make sense, fix it

controller/instance_manager_upgrade_control_controller.go
```diff
@@ -0,0 +1,976 @@
...
+			} else if info.State == longhorn.NodeUpgradeStateFailed {
+				// A failed IMU is retained as the retry record. Do not reset a node
+				// for the same target image here: retryFailedNode must resume this
+				// IMU and enforce the retry limit. A different target starts a new
+				// upgrade cycle with a fresh IMU.
+				failedIMU, err := c.ds.GetInstanceManagerUpgradeRO(info.IMUName)
+				if err != nil {
+					if !datastore.ErrorIsNotFound(err) && !types.ErrorIsNotFound(err) {
+						c.logger.WithError(err).Warnf("Failed to get failed IMU %v for node %v", info.IMUName, nodeID)
+					}
```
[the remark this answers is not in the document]
davidcheng0922: Fix

controller/instance_manager_upgrade_control_controller.go
```diff
@@ -0,0 +1,987 @@
...
+			if err := c.markNodeCompleted(imuc, nodeID); err != nil {
+				return false, err
+			}
+			continue
+		}
+
+		if imu.Status.State != longhorn.InstanceManagerUpgradeStateFailed {
+			imu.Status.State = longhorn.InstanceManagerUpgradeStateFailed
+			imu.Status.ErrorMsg = "instance manager upgrade became orphaned"
+			if _, err := c.ds.UpdateInstanceManagerUpgradeStatus(imu); err != nil {
```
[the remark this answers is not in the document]
davidcheng0922: We intentionally fail orphaned IMUs directly. Once IMUC ownership is lost, the persisted plan may no longer match actual engine placement; abort/restore could conflict with current state or block the rolling upgrade.

The engine remains where it is, this node is skipped for the current cycle, and a later upgrade cycle handles it from observed state.

k8s/pkg/apis/longhorn/v1beta2/instancemanagerupgradecontrol.go
```diff
@@ -0,0 +1,102 @@
...
+// InstanceManagerUpgradeControlSpec defines the desired state of the
+// InstanceManagerUpgradeControl singleton.
+type InstanceManagerUpgradeControlSpec struct {
+	// TargetImage is the desired instance manager image for all nodes.
+	// +optional
+	TargetImage string `json:"targetImage,omitempty"`
+
+	// StartAt is the RFC3339 timestamp at which the upgrade cycle should begin.
+	// +optional
+	StartAt string `json:"startAt,omitempty"`
```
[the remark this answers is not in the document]
davidcheng0922: `StartAt` is populated only from `v2-instance-manager-upgrade-start-time`, which is already validated as `RFC3339` by the setting validator. We do not support direct user edits to this internal IMUC CR, so an additional validation layer is unnecessary.

types/setting.go
```diff
@@ -1758,6 +1767,53 @@ var (
...
+			"During the live upgrade, Longhorn may temporarily relocate engines, detach replicas from engines, and trigger replica rebuilding. " +
+			"When disabled, Longhorn does not automatically upgrade V2 instance managers, and existing V2 instance managers remain on the current image. " +
+			"If this setting is disabled while an automatic V2 instance manager upgrade is in progress, Longhorn allows the current node upgrade to finish but does not start upgrades on additional nodes.",
+		Category:           SettingCategoryGeneral,
+		Type:               SettingTypeBool,
+		Required:           true,
+		ReadOnly:           false,
+		DataEngineSpecific: false,
+		Default:            "false",
+	}
```
derekbit: We probably can extend the mechanism to v1 data engine. WDYT?

BTW, we can add a new category `SettingCategoryUpgrade`. cc @shuo-wu@innobead
```suggestion
	SettingDefinitionAllowInstanceManagerAutomaticUpgrade = SettingDefinition{
		DisplayName: "Allow Instance Manager Automatic Upgrade",
        Description: "This setting allows Longhorn to automatically upgrade instance managers after Longhorn manager is upgraded. " +
        "During the live upgrade, Longhorn may temporarily relocate engines, detach replicas from engines, and trigger replica rebuilding. " +
        "When disabled, Longhorn does not automatically upgrade instance managers, and existing instance managers remain on the current image. " +
        "If this setting is disabled while an automatic instance manager upgrade is in progress, Longhorn allows the current node upgrade to finish but does not start upgrades on additional nodes. " +
        "This setting currently only supports the V2 Data Engine."
		Category:           SettingCategoryUpgrade,
		Type:               SettingTypeBool,
		Type:               SettingTypeBool,
		Required:           true,
		ReadOnly:           false,
		DataEngineSpecific: true,
		Default:            fmt.Sprintf("{%q:\"false\"}", longhorn.DataEngineTypeV2),
	}
```

types/setting.go
```diff
@@ -1758,6 +1767,53 @@ var (
...
+			"This provides flexibility for scheduling the upgrade at a preferred time. If empty, the upgrade starts immediately. " +
+			"Updates to this setting are rejected while an upgrade is actively in progress.\n\n" +
+			"Example: 2026-04-20T15:00:00Z\n\n",
+		Category:           SettingCategoryGeneral,
+		Type:               SettingTypeString,
+		Required:           false,
+		ReadOnly:           false,
+		DataEngineSpecific: false,
+		Default:            "",
+	}
```
```suggestion
	SettingDefinitionInstanceManagerUpgradeStartTime = SettingDefinition{
		DisplayName: "Instance Manager Upgrade Start Time",
		Description: "Specifies when the rolling upgrade of instance managers should begin, in RFC3339 format. " +
		"This setting allows the upgrade to be scheduled at a preferred time. If empty, the upgrade starts immediately. " +
		"Updates to this setting are rejected while an upgrade is in progress. " +
		"This setting currently only applies to the V2 Data Engine.\n\n" +
		"Example: 2026-04-20T15:00:00Z",
		Category:           SettingCategoryUpgrade,
		Type:               SettingTypeString,
		Required:           false,
		ReadOnly:           false,
		DataEngineSpecific: false,
		Default:            fmt.Sprintf("{%q:\"\"}", longhorn.DataEngineTypeV2),
	}
```

types/setting.go
```diff
@@ -1758,6 +1767,53 @@ var (
...
+		Category:           SettingCategoryGeneral,
+		Type:               SettingTypeInt,
+		Required:           true,
+		ReadOnly:           false,
+		DataEngineSpecific: false,
+		Default:            "60",
+		ValueIntRange: map[string]int{
+			ValueIntRangeMinimum: 1,
+		},
+	}
```
```suggestion
	SettingDefinitionInstanceManagerUpgradeTimeout = SettingDefinition{
		DisplayName: "Instance Manager Upgrade Timeout",
        Description: "In minutes. The instance manager is upgraded one node at a time, so an unexpected issue on one node could block upgrades on the remaining nodes. " +
        "This timeout specifies how long a single-node upgrade can remain in the upgrade phases before it is aborted, allowing the upgrade process to continue with other nodes. " +
        "It applies while the upgrade is pending, relocating engines, waiting for the source instance manager, or restoring engines, but does not apply while waiting for post-restore volume health. " +
        "The default value is 60 minutes. " +
        "This setting currently only applies to the V2 Data Engine.\n\n" +
        "Important: Changes to this setting take effect immediately and apply to all in-flight upgrade operations. " +
        "Increasing the timeout gives struggling upgrades more time to complete; decreasing it may cause currently running upgrades to abort if they exceed the new timeout.\n\n",
		Category:           SettingCategoryUpgrade,
		Type:               SettingTypeInt,
		Required:           true,
		ReadOnly:           false,
		DataEngineSpecific: false,
		Default:            fmt.Sprintf("{%q:\"60\"}", longhorn.DataEngineTypeV2),,
		ValueIntRange: map[string]int{
			ValueIntRangeMinimum: 1,
		},
	}
```

types/setting.go
```diff
@@ -489,6 +495,9 @@ var (
 		SettingNameEngineImagePodLivenessProbeTimeout:                       SettingDefinitionEngineImagePodLivenessProbeTimeout,
 		SettingNameEngineImagePodLivenessProbeFailureThreshold:              SettingDefinitionEngineImagePodLivenessProbeFailureThreshold,
 		SettingNameInstanceManagerPodLivenessProbeTimeout:                   SettingDefinitionInstanceManagerPodLivenessProbeTimeout,
+		SettingNameAllowV2InstanceManagerAutomaticUpgrade:                   SettingDefinitionAllowV2InstanceManagerAutomaticUpgrade,
+		SettingNameV2InstanceManagerUpgradeStartTime:                        SettingDefinitionV2InstanceManagerUpgradeStartTime,
+		SettingNameV2InstanceManagerUpgradeTimeout:                          SettingDefinitionV2InstanceManagerUpgradeTimeout,
```
```suggestion
		SettingNameAllowInstanceManagerAutomaticUpgrade:                   SettingDefinitionAllowInstanceManagerAutomaticUpgrade,
		SettingNameInstanceManagerUpgradeStartTime:                        SettingDefinitionInstanceManagerUpgradeStartTime,
		SettingNameInstanceManagerUpgradeTimeout:                          SettingDefinitionInstanceManagerUpgradeTimeout,
```

types/setting.go
```diff
@@ -322,6 +325,9 @@ var (
 		SettingNameEngineImagePodLivenessProbeTimeout,
 		SettingNameEngineImagePodLivenessProbeFailureThreshold,
 		SettingNameInstanceManagerPodLivenessProbeTimeout,
+		SettingNameAllowV2InstanceManagerAutomaticUpgrade,
+		SettingNameV2InstanceManagerUpgradeStartTime,
+		SettingNameV2InstanceManagerUpgradeTimeout,
```
```suggestion
		SettingNameAllowInstanceManagerAutomaticUpgrade,
		SettingNameInstanceManagerUpgradeStartTime,
		SettingNameInstanceManagerUpgradeTimeout,
```

types/setting.go
```diff
@@ -189,6 +189,9 @@ const (
 	SettingNameEngineImagePodLivenessProbeTimeout                       = SettingName("engine-image-pod-liveness-probe-timeout")
 	SettingNameEngineImagePodLivenessProbeFailureThreshold              = SettingName("engine-image-pod-liveness-probe-failure-threshold")
 	SettingNameInstanceManagerPodLivenessProbeTimeout                   = SettingName("instance-manager-pod-liveness-probe-timeout")
+	SettingNameAllowV2InstanceManagerAutomaticUpgrade                   = SettingName("allow-v2-instance-manager-automatic-upgrade")
+	SettingNameV2InstanceManagerUpgradeStartTime                        = SettingName("v2-instance-manager-upgrade-start-time")
+	SettingNameV2InstanceManagerUpgradeTimeout                          = SettingName("v2-instance-manager-upgrade-timeout")
```
```suggestion
	SettingNameAllowInstanceManagerAutomaticUpgrade                   = SettingName("allow-instance-manager-automatic-upgrade")
	SettingNameInstanceManagerUpgradeStartTime                        = SettingName("instance-manager-upgrade-start-time")
	SettingNameInstanceManagerUpgradeTimeout                          = SettingName("instance-manager-upgrade-timeout")
```

datastore/longhorn.go
```diff
@@ -665,6 +665,23 @@ func (s *DataStore) ValidateSetting(name, value string) (err error) {
 			}
 		}
 
+	case types.SettingNameAllowV2InstanceManagerAutomaticUpgrade:
+		if value != "true" {
```
derekbit: nit:

```suggestion
		if value != longhorn.TrueValue {
```

derekbit: Can we support changing the setting to `false` while upgrading v2 instance managers?

datastore/longhorn.go
```diff
@@ -665,6 +665,23 @@ func (s *DataStore) ValidateSetting(name, value string) (err error) {
...
+			break
+		}
+
+		currentVersionSetting, err := s.GetSettingExactRO(types.SettingNameCurrentLonghornVersion)
+		if err != nil {
+			if ErrorIsNotFound(err) {
+				break
+			}
```
derekbit, line 691: If the setting is not found, should we reject the value change?
IMO, we can reject the request and require users to fix the missing setting before updating the value. WDYT?

datastore/longhorn.go
```diff
@@ -665,6 +665,23 @@ func (s *DataStore) ValidateSetting(name, value string) (err error) {
...
+			return errors.Wrap(err, "failed to get current Longhorn version")
+		}
+		if currentVersionSetting.Annotations[types.GetLonghornLabelKey(types.V2InstanceManagerLiveUpgradeUnsupported)] == "true" {
```
derekbit:
```suggestion
		if currentVersionSetting.Annotations[types.GetLonghornLabelKey(types.V2InstanceManagerLiveUpgradeUnsupported)] == longhorn.TrueValue {
```

datastore/longhorn.go
```diff
@@ -665,6 +665,23 @@ func (s *DataStore) ValidateSetting(name, value string) (err error) {
...
+			return fmt.Errorf("cannot enable %v: v2 instance manager live upgrade is only supported from %s",
+				name, types.MinimumLonghornVersionForV2InstanceManagerLiveUpgrade)
```
derekbit:
```suggestion
			return fmt.Errorf("cannot enable %v: v2 instance manager live upgrade is not supported before Longhorn %s", name, types.MinimumLonghornVersionForV2InstanceManagerLiveUpgrade)
```

types/types.go
```diff
@@ -99,6 +101,13 @@ const (
 	CRDAPIVersionV1beta1  = "longhorn.io/v1beta1"
 	CRDAPIVersionV1beta2  = "longhorn.io/v1beta2"
 	CurrentCRDAPIVersion  = CRDAPIVersionV1beta2
+
+	InstanceManagerUpgradeControlName = "longhorn-instance-manager-upgrade-control"
+
+	// MinimumLonghornVersionForV2InstanceManagerLiveUpgrade is the oldest
+	// Longhorn version that supports upgrading into the V2 instance manager
+	// live upgrade feature.
+	MinimumLonghornVersionForV2InstanceManagerLiveUpgrade = "v1.12.2"
```
derekbit, line 110: cc @c3y1huang @shuo-wu @innobead

upgrade/util/util.go
```diff
@@ -183,6 +183,18 @@ func CreateOrUpdateLonghornVersionSetting(namespace string, lhClient *lhclientse
...
+			// Reaching a supported release through an offline upgrade makes subsequent
+			// live upgrades safe again.
+			delete(s.Annotations, types.GetLonghornLabelKey(types.V2InstanceManagerLiveUpgradeUnsupported))
+		} else if s.Value != "" && semver.Compare(s.Value, types.MinimumLonghornVersionForV2InstanceManagerLiveUpgrade) < 0 {
+			// Preserve the unsupported source marker until the cluster reaches a
+			// release that supports v2 instance manager live upgrade.
+			if s.Annotations == nil {
+				s.Annotations = make(map[string]string)
+			}
+			s.Annotations[types.GetLonghornLabelKey(types.V2InstanceManagerLiveUpgradeUnsupported)] = "true"
```
derekbit, line 192:
```suggestion
			s.Annotations[types.GetLonghornLabelKey(types.V2InstanceManagerLiveUpgradeUnsupported)] = longhorn.TrueValue
```

derekbit: The validation approach is quite clever.

controller/instance_manager_controller.go
```diff
@@ -2792,6 +2922,115 @@ func (m *InstanceManagerMonitor) syncOrphans(im *longhorn.InstanceManager, insta
 	m.createOrphanForInstances(existOrphans, im, shardGroupProcesses, longhorn.OrphanTypeShardGroupInstance, m.isShardGroupInstanceOrphaned, longhorn.DataEngineTypeV2)
 }
 
+// syncInstanceManagerUpgrade ensures an InstanceManagerUpgradeControl CR exists
+// and tracks the current default instance manager image for v2 AllInOne IMs.
+func (imc *InstanceManagerController) syncInstanceManagerUpgrade(im *longhorn.InstanceManager) error {
+	if !types.IsDataEngineV2(im.Spec.DataEngine) || im.Spec.Type != longhorn.InstanceManagerTypeAllInOne {
```
derekbit, line 2928: We can remove
```
	InstanceManagerTypeEngine  = InstanceManagerType("engine")
	InstanceManagerTypeReplica = InstanceManagerType("replica")
```
in the following versions.
cc @c3y1huang

controller/instance_manager_controller.go
```diff
@@ -2792,6 +2922,115 @@ func (m *InstanceManagerMonitor) syncOrphans(im *longhorn.InstanceManager, insta
...
+	if datastore.ErrorIsNotFound(existingErr) {
+		newIMUC := &longhorn.InstanceManagerUpgradeControl{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: types.InstanceManagerUpgradeControlName,
+			},
+			Spec: longhorn.InstanceManagerUpgradeControlSpec{
+				TargetImage: defaultIMImage,
+				StartAt:     startTime,
+			},
+		}
```
derekbit, line 2974: As convention, we need mutation logics for instancemanagerupgrade and instancemanagerupgradecontrol CRs to patch appropriate labels. 
Example: https://github.com/longhorn/longhorn-manager/blob/master/webhook/resources/orphan/mutator.go#L58

controller/instance_manager_upgrade_controller.go
```diff
@@ -0,0 +1,2031 @@
...
+	imu = imu.DeepCopy()
+	if imu.DeletionTimestamp == nil && !util.FinalizerExists(longhornFinalizerKey, imu) {
+		if err := util.AddFinalizer(longhornFinalizerKey, imu); err != nil {
+			return err
+		}
+		imu, err = imuc.ds.UpdateInstanceManagerUpgrade(imu)
+		if err != nil {
+			return err
+		}
+	}
```
derekbit: The finalizer should be added by mutation logics.
Example: https://github.com/longhorn/longhorn-manager/blob/master/webhook/resources/orphan/mutator.go#L98-L105

controller/instance_manager_controller.go
```diff
@@ -2792,6 +2922,115 @@ func (m *InstanceManagerMonitor) syncOrphans(im *longhorn.InstanceManager, insta
...
+
+	if needsUpdate {
+		if _, err := imc.ds.UpdateInstanceManagerUpgradeControl(imuc); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (imc *InstanceManagerController) hasStartedNodeUpgradeForTarget(imuc *longhorn.InstanceManagerUpgradeControl, targetImage string) (bool, error) {
```
derekbit, line 3015:
```suggestion
// hasStartedNodeUpgradeForTarget reports whether the upgrade cycle for targetImage has already
// begun, which freezes Spec.StartAt so a scheduling change cannot disturb the active cycle.
// NodeUpgradeInfo does not record the target image, so a StartedAt entry is only conclusive
// after its IMU confirms the entry belongs to this cycle rather than an earlier one.
func (imc *InstanceManagerController) hasStartedNodeUpgradeForTarget(imuc *longhorn.InstanceManagerUpgradeControl, targetImage string) (bool, error) {
```

controller/instance_manager_controller.go
```diff
@@ -2792,6 +2922,115 @@ func (m *InstanceManagerMonitor) syncOrphans(im *longhorn.InstanceManager, insta
...
+	if imuc.Status.CurrentNode != "" {
+		return true, nil
+	}
+	for _, info := range imuc.Status.Nodes {
+		if info.StartedAt == "" || info.IMUName == "" {
+			continue
+		}
+		imu, err := imc.ds.GetInstanceManagerUpgrade(info.IMUName)
+		if err != nil {
+			if datastore.ErrorIsNotFound(err) || types.ErrorIsNotFound(err) {
```
derekbit: Why do we need to do the two ErrorIsNotFound checks? What's the difference?

controller/instance_manager_controller.go
```diff
@@ -2792,6 +2922,115 @@ func (m *InstanceManagerMonitor) syncOrphans(im *longhorn.InstanceManager, insta
...
+	if err != nil {
+		return err
+	}
+
+	needsUpdate := false
+	if imuc.Spec.TargetImage != defaultIMImage {
+		log.Infof("Updating InstanceManagerUpgradeControl target image from %v to %v", imuc.Spec.TargetImage, defaultIMImage)
+		imuc.Spec.TargetImage = defaultIMImage
+		needsUpdate = true
+	}
```
derekbit, line 2995: Why is `imuc.Spec.TargetImage` different than `defaultIMImage`?
Is `imuc.Spec.TargetImage` set to `defaultIMImage` initially?

controller/instance_manager_upgrade_controller.go
```diff
@@ -0,0 +1,2031 @@
...
+	imuc.logger.Info("Starting Longhorn instance manager upgrade controller")
+	defer imuc.logger.Info("Shut down Longhorn instance manager upgrade controller")
+
+	if !cache.WaitForNamedCacheSync("longhorn instance manager upgrades", stopCh, imuc.cacheSyncs...) {
+		return
+	}
+
+	for i := 0; i < workers; i++ {
+		go wait.Until(imuc.worker, time.Second, stopCh)
+	}
```
derekbit: nit:
```suggestion
    for range workers {
        go wait.Until(imuc.worker, time.Second, stopCh)
    }
```

controller/instance_manager_upgrade_control_controller.go
```diff
@@ -0,0 +1,987 @@
...
+			return imu.Name, nil
+		}
+	}
+
+	for attempt := 0; attempt < 3; attempt++ {
+		imuName := generateIMUName(nodeID)
+		imu := &longhorn.InstanceManagerUpgrade{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:       imuName,
+				Finalizers: []string{longhornFinalizerKey},
```
derekbit: Move the finalizer to mutating webhook. In addition, the utating webhook also patches appropriate labels.

controller/instance_manager_upgrade_control_controller.go
```diff
@@ -0,0 +1,987 @@
...
+			continue
+		}
+		if imu.Spec.NodeID == nodeID && imu.Spec.TargetImage == targetImage &&
+			imu.Status.State != longhorn.InstanceManagerUpgradeStateFailed &&
+			imu.Status.State != longhorn.InstanceManagerUpgradeStateCompleted {
+			return imu.Name, nil
+		}
+	}
+
+	for attempt := 0; attempt < 3; attempt++ {
```
derekbit: Use

```suggestion
	for attempt := range 3 {
```

Define a constant for the magic number 3.

controller/instance_manager_upgrade_control_controller.go
```diff
@@ -0,0 +1,987 @@
...
+	defer func() {
+		if err == nil && !reflect.DeepEqual(existingStatus, &imuc.Status) {
+			if _, updateErr := c.ds.UpdateInstanceManagerUpgradeControlStatus(imuc); updateErr != nil {
+				log.WithError(updateErr).Warn("Failed to update InstanceManagerUpgradeControl status in deferred update")
+				err = updateErr
+			}
+		}
+	}()
+
+	active, err := c.reconcile(imuc, log)
```
derekbit: nit:
```suggestion
	active, err := c.reconcile(imuc)
```

controller/instance_manager_upgrade_control_controller.go
```diff
@@ -0,0 +1,987 @@
...
+	if active {
+		c.queue.AddAfter(key, imucRequeueAfter)
+	}
+
+	return nil
+}
+
+// reconcile is the flat, stateless reconcile function. It returns true when
+// a node upgrade is actively in progress (so the caller requeues periodically).
+func (c *InstanceManagerUpgradeControlController) reconcile(imuc *longhorn.InstanceManagerUpgradeControl, log *logrus.Entry) (bool, error) {
```
derekbit: nit:
```suggestion
func (c *InstanceManagerUpgradeControlController) reconcile(imuc *longhorn.InstanceManagerUpgradeControl) (bool, error) {

    log := c.logger.WithField("instanceManagerUpgradeControl", imuc.Name)
```

controller/instance_manager_upgrade_control_controller.go
```diff
@@ -0,0 +1,987 @@
...
+	var startAt time.Time
+	if imuc.Spec.StartAt != "" {
+		var err error
+		startAt, err = util.ParseTime(imuc.Spec.StartAt)
+		if err != nil {
+			return false, errors.Wrapf(err, "failed to parse startAt %v", imuc.Spec.StartAt)
+		}
+	}
+
+	repairedCurrentNode, err := c.repairIMUCInvariants(imuc, log)
```
derekbit:
```suggestion
	repairedCurrentNode, err := c.repairIMUCInvariants(imuc)
```

controller/instance_manager_upgrade_control_controller.go
```diff
@@ -0,0 +1,987 @@
...
+		return true, nil
+	}
+
+	return false, nil
+}
+
+func (c *InstanceManagerUpgradeControlController) repairIMUCInvariants(
+	imuc *longhorn.InstanceManagerUpgradeControl,
+	log *logrus.Entry,
+) (bool, error) {
```
derekbit: // repairIMUCInvariants restores the bookkeeping that the one-node-at-a-time rolling upgrade
// depends on: Status.CurrentNode must name an entry that is in-progress, and an in-progress entry
// must carry an IMU name to remain trackable. In-progress entries other than CurrentNode are left
// intact so recoverOrphanedNodes can fail their IMU as well. Returns true when anything was
// repaired, so the caller ends the round instead of scheduling from state it just rewrote.
```suggestion
func (c *InstanceManagerUpgradeControlController) repairIMUCInvariants(imuc *longhorn.InstanceManagerUpgradeControl) (bool, error) {
    log := c.logger.WithField("instanceManagerUpgradeControl", imuc.Name)
```

controller/instance_manager_upgrade_control_controller.go
```diff
@@ -0,0 +1,987 @@
...
+	}
+	imuc.Status.Nodes[nodeID] = info
+	return nil
+}
+
+// processCurrentNode manages the active IMU for imuc.Status.CurrentNode.
+// Clears CurrentNode when the node reaches a terminal outcome; leaves it set while waiting.
+// Returns active=true when the caller should requeue periodically.
+func (c *InstanceManagerUpgradeControlController) processCurrentNode(imuc *longhorn.InstanceManagerUpgradeControl, log *logrus.Entry,
+) (active bool, err error) {
```
derekbit:
```suggestion
func (c *InstanceManagerUpgradeControlController) processCurrentNode(imuc *longhorn.InstanceManagerUpgradeControl
) (active bool, err error) {
    log := ...
```

derekbit:
Notes while reviewing the PR

- IMUC controller logic
```mermaid
flowchart TD
    A["IMUC reconcile starts"] --> B{"CurrentNode is set?"}

    B -- "No" --> C["Skip processCurrentNode"]
    B -- "Yes" --> D["processCurrentNode"]

    D --> E{"IMUName is empty?"}
    E -- "Yes" --> F["Clear CurrentNode"]
    F --> G["Continue reconcile"]

    E -- "No" --> H["Get InstanceManagerUpgrade by IMUName"]
    H --> I{"IMU exists?"}

    I -- "Yes" --> J{"IMU state"}
    J -- "Completed" --> K["markNodeCompleted"]
    K --> L["Clear CurrentNode"]
    L --> G

    J -- "Failed" --> M["markNodeFailed"]
    M --> N["Clear CurrentNode"]
    N --> G

    J -- "Non-terminal" --> O{"Target image changed?"}
    O -- "Yes" --> P["Set AbortRequested on IMU"]
    P --> Q["Keep CurrentNode active"]
    O -- "No" --> Q
    Q --> R["Requeue after interval"]

    I -- "No / NotFound" --> S["IMU was deleted externally"]
    S --> T["handleNodeFailure"]
    T --> U["markNodeFailed"]
    U --> V["Clear CurrentNode"]
    V --> G

    G --> W["recoverOrphanedNodes"]
    W --> X["handleTargetImageChange"]
    X --> Y{"Auto upgrade allowed?"}

    Y -- "No" --> Z["Stop reconcile"]
    Y -- "Yes" --> AA["pickNextPendingNode"]

    AA --> AB{"Failed node's IMU is missing?"}
    AB -- "Yes" --> AC["resetNodeToPending"]
    AC --> AD["Node becomes pending candidate"]

    AB -- "No" --> AE{"Any pending node?"}
    AD --> AE

    AE -- "No" --> AF["retryFailedNode"]
    AE -- "Yes" --> AG["startNodeUpgrade"]

    AG --> AH["ensureIMUForNode"]
    AH --> AI{"Reusable non-terminal IMU exists?"}

    AI -- "Yes" --> AJ["Reuse existing IMU"]
    AI -- "No" --> AK["Create new IMU with generated name"]

    AJ --> AL["startNodeInProgress"]
    AK --> AL
    AL --> AM["New IMU is processed again"]
```


- If IMU CR is gone (IMUC controller: processCurrentNode())
```mermaid
flowchart TD
    A["External IMU deletion"] --> B{"Object still finalizing?"}
    B -- "Yes" --> C["IMU controller handles abort / restore / cleanup"]
    B -- "No, object is gone" --> D["IMUC sees NotFound"]
    D --> E["Mark node Failed"]
    E --> F["pickNextPendingNode sees missing failed IMU"]
    F --> G["Reset node to Pending"]
    G --> H["Create a new IMU"]
    H --> I["Process upgrade again"]
```

controller/instance_manager_upgrade_control_controller.go
```diff
@@ -0,0 +1,987 @@
...
+			"Node %v upgrade completed", nodeID)
+		return false, nil
+
+	case longhorn.InstanceManagerUpgradeStateFailed:
+		log.Warnf("Node %v upgrade failed: %v", nodeID, imu.Status.ErrorMsg)
+		if err := c.markNodeFailed(imuc, nodeID, imu.Status.ErrorMsg); err != nil {
+			return false, err
+		}
+		c.eventRecorder.Eventf(imuc, corev1.EventTypeWarning, constant.EventReasonFailed,
+			"Node %v upgrade failed: %v", nodeID, imu.Status.ErrorMsg)
```
derekbit: Could this result in flooding if the node upgrade keeps failing?

controller/instance_manager_upgrade_control_controller.go
```diff
@@ -0,0 +1,987 @@
...
+		return false, nil
+	}
+
+	// For non-terminal states, only enforce target-image changes here. The IMU
+	// controller owns timeout detection so there is a single writer for timeout
+	// abort transitions.
+	if imu.Spec.TargetImage != imuc.Spec.TargetImage {
```
derekbit, line 648: Is the change due to rollback?
Can we allow rollback operation?

controller/instance_manager_upgrade_control_controller.go
```diff
@@ -0,0 +1,987 @@
...
+	return nil
+}
+
+// processCurrentNode manages the active IMU for imuc.Status.CurrentNode.
+// Clears CurrentNode when the node reaches a terminal outcome; leaves it set while waiting.
+// Returns active=true when the caller should requeue periodically.
+func (c *InstanceManagerUpgradeControlController) processCurrentNode(imuc *longhorn.InstanceManagerUpgradeControl, log *logrus.Entry,
+) (active bool, err error) {
+	nodeID := imuc.Status.CurrentNode
+	nodeInfo := imuc.Status.Nodes[nodeID]
```
derekbit: It is better to check if `nodeInfo` exists in the map first.

controller/instance_manager_upgrade_control_controller.go
```diff
@@ -0,0 +1,987 @@
...
+// pickNextPendingNode prioritizes nodes that host v2 engines, then selects
+// lexicographically among remaining replica-only nodes. It discovers nodes
+// from IM CRs, then checks convergence using the active pod-backed IM.
+func (c *InstanceManagerUpgradeControlController) pickNextPendingNode(imuc *longhorn.InstanceManagerUpgradeControl) string {
+	// Register any nodes we haven't seen yet.
+	discoveredNodes := map[string]struct{}{}
+	// Exclude nodes whose active IM already uses the requested image. Keep their
+	// existing status unchanged so terminal upgrade history is retained.
+	excludedNodes := map[string]struct{}{}
+	ims, err := c.ds.ListInstanceManagersRO()
```
derekbit: nit:
Can we introduce `ListInstanceManagersROByDataEngine()`? We can improve it in the future.

controller/instance_manager_upgrade_control_controller.go
```diff
@@ -0,0 +1,987 @@
...
+			c.queue.AddAfter(c.namespace+"/"+imuc.Name, time.Until(startAt))
+			return false, nil
+		}
+	}
+
+	if nextNode == "" {
+		return c.retryFailedNode(imuc, log)
+	}
+
+	if err := c.startNodeUpgrade(imuc, nextNode, log); err != nil {
```
derekbit:
```suggestion
	if err := c.startNodeUpgrade(imuc, nextNode); err != nil {
```

controller/instance_manager_upgrade_control_controller.go
```diff
@@ -0,0 +1,987 @@
...
+	if imuc.Spec.StartAt != "" && !hasStartedNodeUpgrade(imuc) {
+		if time.Now().Before(startAt) {
+			log.Debugf("Upgrade scheduled for %v, waiting", imuc.Spec.StartAt)
+			c.queue.AddAfter(c.namespace+"/"+imuc.Name, time.Until(startAt))
+			return false, nil
+		}
+	}
+
+	if nextNode == "" {
+		return c.retryFailedNode(imuc, log)
```
derekbit:
```suggestion
		return c.retryFailedNode(imuc)
```

controller/instance_manager_upgrade_control_controller.go
```diff
@@ -0,0 +1,987 @@
...
+	if err != nil {
+		return false, err
+	}
+	if recoveringOrphan {
+		return true, nil
+	}
+
+	// --- Step 2: check for a target image change mid-cycle ---
+	// Always run, whether or not a node upgrade is in progress.
+	if err := c.handleTargetImageChange(imuc, log); err != nil {
```
derekbit:
```suggestion
	if err := c.handleTargetImageChange(imuc); err != nil {
```

controller/instance_manager_upgrade_control_controller.go
```diff
@@ -0,0 +1,987 @@
...
+		active, err = c.processCurrentNode(imuc, log)
+		if err != nil {
+			return false, err
+		}
+	}
+
+	// --- Step 1.5: detect and fix orphaned in-progress nodes ---
+	// A node can be stuck in "in-progress" state without being the currentNode
+	// if its IMU was deleted externally or if the currentNode was cleared unexpectedly.
+	recoveringOrphan, err := c.recoverOrphanedNodes(imuc, log)
```
derekbit:
```suggestion
	recoveringOrphan, err := c.recoverOrphanedNodes(imuc)
```

controller/instance_manager_upgrade_control_controller.go
```diff
@@ -0,0 +1,987 @@
...
+
+// recoverOrphanedNodes marks nodes stuck in "in-progress" state without being
+// the current node as failed. This can happen if:
+// - An IMU was deleted externally while the node was being upgraded
+// - The currentNode was cleared but the node state wasn't reset
+// - A controller restart interrupted state transitions
+func (c *InstanceManagerUpgradeControlController) recoverOrphanedNodes(
+	imuc *longhorn.InstanceManagerUpgradeControl,
+	log *logrus.Entry,
+) (bool, error) {
```
derekbit:
```suggestion
func (c *InstanceManagerUpgradeControlController) recoverOrphanedNodes(imuc *longhorn.InstanceManagerUpgradeControl) (bool, error) {
    log := ...
```

controller/instance_manager_upgrade_control_controller.go
```diff
@@ -0,0 +1,987 @@
...
+			imu.Status.ErrorMsg = "instance manager upgrade became orphaned"
+			if _, err := c.ds.UpdateInstanceManagerUpgradeStatus(imu); err != nil {
+				return false, errors.Wrapf(err, "failed to mark orphaned IMU %v as failed", imu.Name)
+			}
+		}
+		if err := c.markNodeFailed(imuc, nodeID, imu.Status.ErrorMsg); err != nil {
+			return false, err
+		}
+		c.eventRecorder.Eventf(imuc, corev1.EventTypeWarning, constant.EventReasonUpdate,
+			"Marked orphaned IMU %v for node %v as failed", imu.Name, nodeID)
```
derekbit: It might result in flooding events. We can log it instead.

upgrade/util/util.go
```diff
@@ -183,6 +183,18 @@ func CreateOrUpdateLonghornVersionSetting(namespace string, lhClient *lhclientse
...
+		}
```
derekbit: Should be
```suggestion
		if s.Value != "" && semver.Compare(s.Value, types.MinimumLonghornVersionForV2InstanceManagerLiveUpgrade) < 0 {			// Preserve the unsupported source marker until the cluster reaches a
			// release that supports v2 instance manager live upgrade.
			if s.Annotations == nil {
				s.Annotations = make(map[string]string)
			}
			s.Annotations[types.GetLonghornLabelKey(types.V2InstanceManagerLiveUpgradeUnsupported)] = "true"
		} else if semver.Compare(meta.Version, types.MinimumLonghornVersionForV2InstanceManagerLiveUpgrade) >= 0 {
			// Reaching a supported release through an offline upgrade makes subsequent
			// live upgrades safe again.
			delete(s.Annotations, types.GetLonghornLabelKey(types.V2InstanceManagerLiveUpgradeUnsupported))
		}
```
?

controller/instance_manager_upgrade_controller.go
```diff
@@ -0,0 +1,2031 @@
...
+		timeoutMinutes = 60
+	}
+	upgradeTimeout := time.Duration(timeoutMinutes) * time.Minute
+
+	if time.Since(startedAt) <= upgradeTimeout {
+		return false, nil
+	}
+
+	// Timeout detected: transition to RestoringEngines to revert engines back to original nodes.
+	// We use a single global timeout - StartedAt is never reset, so this provides predictable
```
derekbit: In the retry failed node flow, the `StartedAt` is set to `util.Now()`. It is inconsistent and the timeout falls into per-retry window. Should we use the global timeout instead?

upgrade/util/util.go
derekbit: Added test cases in util_test.go
```
func TestCreateOrUpdateLonghornVersionSettingMarksUnsupportedV2IMLiveUpgradeSource(t *testing.T) {
	lhClient := lhfake.NewSimpleClientset() // nolint: staticcheck
	setting := &longhorn.Setting{
			Name: string(types.SettingNameCurrentLonghornVersion),
		},
		Value: "v1.12.1",
	}
	_, err := lhClient.LonghornV1beta2().Settings(TestNamespace).Create(context.TODO(), setting, metav1.CreateOptions{})
	require.NoError(t, err)

	originalVersion := meta.Version
	meta.Version = "v1.13.0"
	t.Cleanup(func() { meta.Version = originalVersion })

	require.NoError(t, CreateOrUpdateLonghornVersionSetting(TestNamespace, lhClient))

	updated, err := lhClient.LonghornV1beta2().Settings(TestNamespace).Get(context.TODO(), string(types.SettingNameCurrentLonghornVersion), metav1.GetOptions{})
	require.NoError(t, err)
	assert.Equal(t, "v1.13.0", updated.Value)
	assert.Equal(t, "true", updated.Annotations[types.GetLonghornLabelKey(types.V2InstanceManagerLiveUpgradeUnsupported)])
}

func TestCreateOrUpdateLonghornVersionSettingClearsStaleUnsupportedV2IMLiveUpgradeMarker(t *testing.T) {
	lhClient := lhfake.NewSimpleClientset() // nolint: staticcheck
	setting := &longhorn.Setting{
			Name: string(types.SettingNameCurrentLonghornVersion),
			Annotations: map[string]string{
				types.GetLonghornLabelKey(types.V2InstanceManagerLiveUpgradeUnsupported): "true",
			},
		},
		Value: "v1.12.2",
	}
	_, err := lhClient.LonghornV1beta2().Settings(TestNamespace).Create(context.TODO(), setting, metav1.CreateOptions{})
	require.NoError(t, err)

	originalVersion := meta.Version
	meta.Version = "v1.13.0"
	t.Cleanup(func() { meta.Version = originalVersion })

	require.NoError(t, CreateOrUpdateLonghornVersionSetting(TestNamespace, lhClient))

	updated, err := lhClient.LonghornV1beta2().Settings(TestNamespace).Get(context.TODO(), string(types.SettingNameCurrentLonghornVersion), metav1.GetOptions{})
	require.NoError(t, err)
	assert.Equal(t, "v1.13.0", updated.Value)
	assert.NotContains(t, updated.Annotations, types.GetLonghornLabelKey(types.V2InstanceManagerLiveUpgradeUnsupported))
}
```

controller/volume_controller.go
```diff
@@ -1234,6 +1270,40 @@ func (c *VolumeController) cleanupReplicas(v *longhorn.Volume, es map[string]*lo
...
 }
 
+func (c *VolumeController) getPlannedDetachedReplicasForVolume(v *longhorn.Volume) (map[string]struct{}, error) {
+	plannedDetachedReplicas := map[string]struct{}{}
+	if !types.IsDataEngineV2(v.Spec.DataEngine) {
+		return plannedDetachedReplicas, nil
+	}
+	if v.Status.State != longhorn.VolumeStateAttached {
+		return plannedDetachedReplicas, nil
+	}
```
derekbit: Should we consider attaching and detaching?

controller/instance_manager_upgrade_controller.go
davidcheng0922: I will fix it, thank you

controller/instance_manager_controller.go
davidcheng0922: I will add it, thank you

controller/instance_manager_controller.go
davidcheng0922: You're right, I should only keep `datastore.ErrorIsNotFound(err)`

controller/instance_manager_controller.go
davidcheng0922: For a continuous upgrade, for example `v1.12.2 -> v1.13.0 -> v1.14.0`, the IMUC from the first cycle is retained. At the beginning of the second cycle, its `Spec.TargetImage` still points to the `v1.13.0` instance-manager image, while the manager startup flow updates `default-instance-manager-image` to the `v1.14.0` image.  

**The mismatch therefore identifies a new upgrade cycle.** 

This code updates the existing IMUC target to the new default image so the control controller can re-evaluate all nodes for `v1.14.0`.

controller/instance_manager_upgrade_control_controller.go
davidcheng0922: Move it, thank you

controller/instance_manager_upgrade_control_controller.go
davidcheng0922: Will fix it, thank you

datastore/longhorn.go
davidcheng0922: Yes. When the setting is changed to `false` during a V2 instance manager upgrade, Longhorn allows the current node upgrade to finish but does not start or retry upgrades on additional nodes. 

Changing it back to `true` resumes the rolling upgrade for the remaining eligible nodes.

controller/instance_manager_upgrade_control_controller.go
davidcheng0922: Normally, it should not. Once an IMU enters the Failed state, the node is retried at most five times.

However, to avoid duplicate Events when status updates are retried, I will keep the warning log and remove this Event.

controller/instance_manager_upgrade_control_controller.go
davidcheng0922: No. A normal Longhorn downgrade is rejected by the upgrade path, so this is not intended to support rollback.

This handles the case where the target instance-manager image changes while a live upgrade cycle is still in progress. For example, the original image is `v1.13.0`, the first target is `v1.14.0-alpha`, and the user then upgrades to `v1.14.0-beta`. The active IMU is aborted and engines are restored before the controller proceeds with the new target image.

controller/instance_manager_upgrade_control_controller.go
davidcheng0922: We already have `ListInstanceManagersBySelectorRO()`, I will use it instead

controller/instance_manager_upgrade_control_controller.go
davidcheng0922: Agree, fix it

upgrade/util/util.go
davidcheng0922: Correct, fix it

controller/instance_manager_upgrade_controller.go
davidcheng0922: For each retry, it should have its own timeout window. Therefore, I prefer to reset StartedAt for each retry and update the comment accordingly. Thank you.

controller/volume_controller.go
davidcheng0922: Modified, thank you

davidcheng0922:
- https://ci.longhorn.io/job/private/job/longhorn-e2e-test/9916/
- https://ci.longhorn.io/job/private/job/longhorn-e2e-test/9917/
- https://ci.longhorn.io/job/private/job/longhorn-e2e-test/9918/

derekbit (approved):

derekbit:
@mergify backport v1.13.x

PR #4690, opened 2026-04-29
fix(csi): return the created backup instead of timeout before backup sync

COLDTURNIP:

Issue longhorn/longhorn#11429

When waitForBackupControllerSync times out, the Backup CR may already exist but its SnapshotCreatedAt field is not yet reconciled (blocked by the other BackupController reconcile steps or slow S3 initiation). Previously this returned codes.DeadlineExceeded, causing the VolumeSnapshot to stay stuck in Error state.

Per the CSI spec, once the snapshot is cut and upload is in progress CreateSnapshot SHOULD return OK with ready_to_use=false. The sidecar then re-calls CreateSnapshot idempotently until ready_to_use=true.

https://github.com/container-storage-interface/spec/blob/master/spec.md#the-ready_to_use-parameter

Changes:

- `waitForBackupControllerSync`: continue polling `backup.SnapshotCreated != ""` for best effort. On timeout: if the Backup CR exists return it so the caller produces ready_to_use=false; if the CR is absent return `codes.Internal` (genuine creation failure).
- First-creation path: encode `snapshot_id` as `bak://volume/backup.Name` instead of `bak://volume/backup.Id` (S3 URL `bak://volumeName/s3://bucket/...`). Aligns with the idempotent retry path (line 1056) and fixes `DeleteSnapshot`, which passes the decoded id to `ActionBackupDelete` as the Backup CR name.

### DIFF
--- a/csi/controller_server.go
+++ b/csi/controller_server.go
@@ -1130,9 +1130,9 @@ func (cs *ControllerServer) createCSISnapshotTypeLonghornBackup(req *csi.CreateS
 		return nil, err
 	}
 
-	log.Infof("Volume %s backup %s of snapshot %s created", existVol.Name, backup.Id, csiSnapshotName)
-	snapshotID := encodeSnapshotID(csiSnapshotTypeLonghornBackup, existVol.Name, backup.Id)
-	rsp := createSnapshotResponseForSnapshotTypeLonghornBackup(existVol.Name, snapshotID, snapshotCR.CreationTime,
+	log.Infof("Volume %s backup %s of snapshot %s created", backup.VolumeName, backup.Name, csiSnapshotName)
+	snapshotID := encodeSnapshotID(csiSnapshotTypeLonghornBackup, backup.VolumeName, backup.Name)
+	rsp := createSnapshotResponseForSnapshotTypeLonghornBackup(backup.VolumeName, snapshotID, snapshotCR.CreationTime,
 		existVol.Size, backup.State == string(longhorn.BackupStateCompleted))
 	return rsp, nil
 }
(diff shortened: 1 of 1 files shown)


csi/controller_server.go
```diff
@@ -1515,9 +1512,20 @@ func (cs *ControllerServer) waitForBackupControllerSync(volumeName, snapshotName
 	for {
 		select {
 		case <-timeout:
-			msg := fmt.Sprintf("waitForBackupControllerSync: timeout while waiting for backup controller to sync for volume %s and snapshot %s", volumeName, snapshotName)
+			// Best effort: SnapshotCreated was not synced in time. If the Backup CR exists, return it so
+			// the caller can respond with ready_to_use=false per the CSI spec. The sidecar will retry
+			// CreateSnapshot idempotently until the backup completes and ready_to_use=true is returned.
```
derekbit, line 1517: Have you checked the `CreateSnapshot` is idempotent?

csi/controller_server.go
```diff
@@ -1515,9 +1512,20 @@ func (cs *ControllerServer) waitForBackupControllerSync(volumeName, snapshotName
...
+			backup, err := cs.getBackup(volumeName, snapshotName)
+			if err != nil {
+				return nil, err
+			}
+			if backup != nil {
+				logrus.Infof("waitForBackupControllerSync: backup CR for snapshot %s on volume %s exists but SnapshotCreated not yet synced; returning with ready_to_use=false", snapshotName, volumeName)
+				return backup, nil
+			}
```
derekbit, line 1525: Can we change it to Debug?

c3y1huang (commented):
Other than @derekbit comments, lgtm.

csi/controller_server.go
COLDTURNIP: To the `bak` case of this issue, before cutting the snapshot, we've checked if the Backup CR is created earlier at line 1042.

derekbit (approved):

c3y1huang (approved):

PR #4693, opened 2026-04-30
feat(metrics): add setting to toggle Kubernetes Metrics Server integr…

hookak:
…ation

Add kubernetes-metrics-server-metrics-enabled (Bool, default true) to let users disable metrics.k8s.io API calls on clusters without metrics-server installed.

https://github.com/longhorn/longhorn/issues/13011

### DIFF
--- a/controller/setting_controller.go
+++ b/controller/setting_controller.go
@@ -1685,6 +1685,16 @@ func (info *ClusterInfo) collectNodeCount() error {
 }
 
 func (info *ClusterInfo) collectResourceUsage() error {
+	enabled := true
+	if v, err := info.ds.GetSettingAsBool(types.SettingNameKubernetesMetricsServerMetricsEnabled); err != nil {
+		logrus.WithError(err).Warnf("Failed to get setting %v, defaulting to enabled", types.SettingNameKubernetesMetricsServerMetricsEnabled)
+	} else {
+		enabled = v
+	}
+	if !enabled {
+		return nil
+	}
+
 	componentMap := map[string]map[string]string{
 		"Manager":         info.ds.GetManagerLabel(),
 		"InstanceManager": types.GetInstanceManagerComponentLabel(),
--- a/types/setting.go
+++ b/types/setting.go
@@ -79,6 +79,7 @@ const (
 	SettingNameUpgradeResponderURL                                      = SettingName("upgrade-responder-url")
 	SettingNameManagerURL                                               = SettingName("manager-url")
 	SettingNameAllowCollectingLonghornUsage                             = SettingName("allow-collecting-longhorn-usage-metrics")
+	SettingNameKubernetesMetricsServerMetricsEnabled                    = SettingName("kubernetes-metrics-server-metrics-enabled")
 	SettingNameCurrentLonghornVersion                                   = SettingName("current-longhorn-version")
 	SettingNameLatestLonghornVersion                                    = SettingName("latest-longhorn-version")
 	SettingNameStableLonghornVersions                                   = SettingName("stable-longhorn-versions")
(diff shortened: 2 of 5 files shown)


types/setting.go
```diff
@@ -743,6 +746,21 @@ var (
 		Default:            "true",
 	}
 
+	SettingDefinitionKubernetesMetricsServerMetricsEnabled = SettingDefinition{
+		DisplayName: "Kubernetes Metrics Server Metrics Enabled",
+		Description: "Enabling this setting allows Longhorn to query the Kubernetes Metrics Server (metrics.k8s.io) for pod and node CPU/memory usage. " +
+			"Disable this on clusters where the Metrics Server is not installed to avoid repeated scrape errors and unnecessary API traffic. " +
+			"When disabled, the following metrics are not exposed: longhorn_node_cpu_usage_millicpu, longhorn_node_memory_usage_bytes, " +
+			"longhorn_instance_manager_cpu_usage_millicpu, longhorn_instance_manager_memory_usage_bytes, " +
+			"longhorn_manager_cpu_usage_millicpu, longhorn_manager_memory_usage_bytes.\n",
```
derekbit, line 755: @sushant-suse Could you review the description? Thanks.

derekbit (dismissed):
Technically, LGTM. Thanks @hookak

controller/setting_controller.go
```diff
@@ -1685,6 +1685,15 @@ func (info *ClusterInfo) collectNodeCount() error {
 }
 
 func (info *ClusterInfo) collectResourceUsage() error {
+	enabled, err := info.ds.GetSettingAsBool(types.SettingNameKubernetesMetricsServerMetricsEnabled)
+	if err != nil {
+		logrus.WithError(err).Warnf("Failed to get setting %v", types.SettingNameKubernetesMetricsServerMetricsEnabled)
+		return nil
+	}
```
c3y1huang, line 1693: For the setting lookups, since the [default is enabled](https://github.com/hookak/longhorn-manager/blob/10d46abe9c0bad401e639539e565318aece43ce7/types/setting.go#L761), Longhorn should fall back to `true` on lookup failure.

c3y1huang (changes requested):

controller/setting_controller.go
hookak: Agree! Thanks

types/setting.go
sushant-suse, line 755:
```suggestion
		DisplayName: "Kubernetes Metrics Server Metrics Enabled",
		Description: "Enabling this setting allows Longhorn to query the Kubernetes Metrics Server ('metrics.k8s.io') for pod and node resource usage. " +
			"Disable if Metrics Server is not installed to prevent repeated scrape errors and unnecessary API traffic. " +
			+
			"Disabling hides metrics for: longhorn_node_cpu_usage_millicpu, longhorn_node_memory_usage_bytes, " +
			"longhorn_instance_manager_cpu_usage_millicpu, longhorn_instance_manager_memory_usage_bytes, " +
			"longhorn_manager_cpu_usage_millicpu, longhorn_manager_memory_usage_bytes.\n",
```

sushant-suse (changes requested):

types/setting.go
derekbit: @hookak Could you check the comment from @sushant-suse? Thank you.

hookak: I think @sushant-suse version is better. Thanks for the comment!

derekbit (approved):
LGTM. Thanks @hookak

hookak:
Thanks for your kind review. @derekbit 
Additionally, I think it would be better to get metrics from kubelet directly, rather than depending on kube-metrics server. What do you think?

derekbit:
> Thanks for your kind review. @derekbit Additionally, I think it would be better to get metrics from kubelet directly, rather than depending on kube-metrics server. What do you think?

But need to handle the endpoint of each kubelet?

derekbit (approved):

PR #4697, opened 2026-04-30
fix: enforce snapshot max count for Snapshot CR creation

carterli0407-cell:
### Which issue(s) this PR fixes:

Issue longhorn/longhorn#12934

### What this PR does / why we need it:

This PR makes snapshot max count enforcement consistent across snapshot creation paths.
#### Before this change:

snapshotCreate was rejected when Snapshot Maximum Count was reached.
snapshotCRCreate and direct Snapshot CR creation could still be accepted.
#### After this change:

Snapshot creation is rejected consistently when snapshot usage is equal to or greater than snapshotMaxCount, including Snapshot CR paths.
Update flow is also guarded when createSnapshot changes from false to true, so patch-based bypass is prevented.
### Why we need this:

It closes a behavior gap between API paths.
It prevents accepted-then-fail-later behavior for Snapshot CR requests.
It aligns user-facing behavior with configured snapshot limits.

### Special notes for your reviewer:
Validation done:

Ran focused unit tests for snapshot webhook package.
Added/updated tests for snapshot count usage logic.
Manually verified behavior consistency:
At max count: snapshotCreate, snapshotCRCreate, and Snapshot CR creation are all rejected.
Below max count: snapshot creation continues to work as expected.
#### Scope notes:

Change is intentionally minimal and targeted to snapshot validation behavior.
No unrelated refactor or API surface change included.
#### Root cause summary:

Snapshot max count was enforced in one creation path, but Snapshot CR admission path lacked equivalent protection.
#### Design choice:

Enforce in snapshot validation layer so all Snapshot CR entry points are covered consistently and early.


webhook/resources/snapshot/validator.go
```diff
@@ -59,6 +59,10 @@ func (o *snapshotValidator) Create(request *admission.Request, newObj runtime.Ob
 		return werror.NewInvalidError(fmt.Sprintf("snapshot is not allowed for linked-clone volume %v", snapshot.Spec.Volume), "")
 	}
 
+	if err := o.validateSnapshotMaxCount(snapshot); err != nil {
+		return werror.NewInvalidError(err.Error(), "spec.createSnapshot")
+	}
```
derekbit: function is misleading. how about
```suggestion
	if err := o.validateSnapshotCount(snapshot); err != nil {
		return werror.NewInvalidError(err.Error(), "spec.createSnapshot")
	}
```

webhook/resources/snapshot/validator.go
```diff
@@ -84,5 +88,76 @@ func (o *snapshotValidator) Update(request *admission.Request, oldObj runtime.Ob
 		return werror.NewInvalidError(fmt.Sprintf("label %v is immutable", types.LonghornLabelVolume), "metadata.labels")
 	}
 
+	if !oldSnapshot.Spec.CreateSnapshot && newSnapshot.Spec.CreateSnapshot {
+		if err := o.validateSnapshotMaxCount(newSnapshot); err != nil {
+			return werror.NewInvalidError(err.Error(), "spec.createSnapshot")
+		}
+	}
```
derekbit: Why do we need the logic here? Can you elaborate on the meaning of `Snapshot.Spec.CreateSnapshot`?

webhook/resources/snapshot/validator.go
```diff
@@ -84,5 +88,76 @@ func (o *snapshotValidator) Update(request *admission.Request, oldObj runtime.Ob
...
+
 	return nil
 }
+
+func (o *snapshotValidator) validateSnapshotMaxCount(snapshot *longhorn.Snapshot) error {
+	if snapshot == nil || !snapshot.Spec.CreateSnapshot {
+		return nil
+	}
```
derekbit: 1. Is it possible that snapshot is `nil`?
2. Why do we need to check `snapshot.Spec.CreateSnapshot`?

webhook/resources/snapshot/validator.go
```diff
@@ -84,5 +88,76 @@ func (o *snapshotValidator) Update(request *admission.Request, oldObj runtime.Ob
...
+
+	volume, err := o.ds.GetVolumeRO(snapshot.Spec.Volume)
+	if err != nil {
+		return fmt.Errorf("failed to get volume %v: %v", snapshot.Spec.Volume, err)
```
derekbit:
```suggestion
		return errors.Wrapf(err, ....)
```

webhook/resources/snapshot/validator.go
```diff
@@ -84,5 +88,76 @@ func (o *snapshotValidator) Update(request *admission.Request, oldObj runtime.Ob
...
+	if !oldSnapshot.Spec.CreateSnapshot && newSnapshot.Spec.CreateSnapshot {
+		if err := o.validateSnapshotMaxCount(newSnapshot); err != nil {
+			return werror.NewInvalidError(err.Error(), "spec.createSnapshot")
+		}
+	}
+
 	return nil
 }
+
+func (o *snapshotValidator) validateSnapshotMaxCount(snapshot *longhorn.Snapshot) error {
```
derekbit: There are two snapshot count, one is the global setting https://longhorn.io/docs/1.12.0/references/settings/#snapshot-maximum-count and the other one is `volume.Spec.SnapshotMaxCount`. The check global setting check is missing. 
cc @COLDTURNIP

webhook/resources/snapshot/validator.go
```diff
@@ -84,5 +88,76 @@ func (o *snapshotValidator) Update(request *admission.Request, oldObj runtime.Ob
...
+
+	currentSnapshotCount := 0
+	for _, engine := range engines {
+		count := getSnapshotCountUsage(engine.Status.Snapshots)
+		if count > currentSnapshotCount {
+			currentSnapshotCount = count
+		}
+	}
+
+	return checkSnapshotMaxCount(snapshot.Spec.Volume, volume.Spec.SnapshotMaxCount, currentSnapshotCount)
```
derekbit: Why not just calculate the number of the existing snapshot CRs?

webhook/resources/snapshot/validator.go
```diff
@@ -84,5 +88,76 @@ func (o *snapshotValidator) Update(request *admission.Request, oldObj runtime.Ob
...
+	if currentSnapshotCount >= snapshotMaxCount {
+		return fmt.Errorf(
+			"cannot create snapshot for volume %s: snapshot count usage %d is greater than or equal to snapshotMaxCount %d; remove snapshots or increase snapshotMaxCount",
+			volumeName,
+			currentSnapshotCount,
+			snapshotMaxCount,
+		)
+	}
+	return nil
+}
```
derekbit: The helper function seems not helpful. We can remove this function and merge the codes into validateSnapshotMaxCount.

webhook/resources/snapshot/validator_test.go
```diff
@@ -0,0 +1,161 @@
...
+
+	"github.com/stretchr/testify/assert"
+
+	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
+)
+
+// ---------------------------------------------------------------------------
+// getSnapshotCountUsage
+// ---------------------------------------------------------------------------
+
```
derekbit: Remove the lines

webhook/resources/snapshot/validator_test.go
```diff
@@ -0,0 +1,161 @@
...
+//
+// Both Create (spec.createSnapshot=true on a new CR) and the Update
+// false→true transition ultimately call validateSnapshotMaxCount, which
+// delegates to checkSnapshotMaxCount once the volume and engine data have
+// been fetched from the datastore.  Because checkSnapshotMaxCount is a
+// pure function, we can test every boundary condition here without a real
+// DataStore, giving us full end-to-end coverage of the enforcement logic
+// for both admission paths.
+// ---------------------------------------------------------------------------
+
```
derekbit: Remove the test case because `checkSnapshotMaxCount` is not really necessary as mentioned above.

In addtion,
- the description should be in the function side rather than in the test case side.
- A descriptions should be started with `TestCheckSnapshotMaxCount ... `. 
- "// ---------------------------------------------------------------------------" is not necessary. Please check the other functions' descriptions.

webhook/resources/snapshot/validator.go
```diff
@@ -59,6 +59,10 @@ func (o *snapshotValidator) Create(request *admission.Request, newObj runtime.Ob
 		return werror.NewInvalidError(fmt.Sprintf("snapshot is not allowed for linked-clone volume %v", snapshot.Spec.Volume), "")
 	}
 
+	if err := o.validateSnapshotMaxCount(snapshot); err != nil {
+		return werror.NewInvalidError(err.Error(), "spec.createSnapshot")
```
derekbit: > return werror.NewInvalidError(err.Error(), "spec.createSnapshot")

Why use "spec.createSnapshot" here?

derekbit (commented):
Please go check with @COLDTURNIP for the global setting snapshot count max and per-volume snapshot count max. If the doc is missing in the official doc, please update the official doc as well.

webhook/resources/snapshot/validator.go
carterli0407-cell: From what I've understood as:
spec.createSnapshot is a one-shot trigger. A Snapshot CR can exist purely as a passive record, for example, tracking a backup-scheduler snapshot or it can actively request a new snapshot be taken. Setting spec.createSnapshot: true tells the snapshot controller to call the engine and take a snapshot. Once taken, status.creationTime is populated and the field is never acted on again.

The false to true guard on the Update path exists because a user can activate a previously passive CR by flipping the field via an update rather than a create. If we only checked during Create, this transition would bypass max-count enforcement entirely. The guard catches the exact moment a new snapshot is being triggered, regardless of whether that trigger comes from a fresh CR or an updated one.
I thinke that's the important reason why this issue happened before that.

webhook/resources/snapshot/validator.go
carterli0407-cell: I just add only for the nil check is purely a defensive guard against future callers - snapshot.Spec.CreateSnapshot.

IMO regarding the second question, as I mentioned A Snapshot CR with createSnapshot: false is a passive record, it tracks a snapshot that already exists on the engine.
No new snapshot is being taken, so there is nothing to count-limit. Enforcement only makes sense when createSnapshot: true, because that is the moment the controller will actually call the engine and consume a snapshot slot.

webhook/resources/snapshot/validator.go
carterli0407-cell: I've updated.

webhook/resources/snapshot/validator.go
carterli0407-cell: Yes, I know that there are two snapshot count configurations.
I've noticed that volume.Spec.SnapshotMaxcount is set based on global setting value when the volume is created. If the global setting value is changed afterward, it doesn't affect existing volumes
For example, when the global value is 250, a newly created volume gets volume.Spec.SnapshotMaxCount= 250.
If the global setting is later changed to 3 for testing, the existing volume still retains the value of 25. 
Based on this behavior, it seems unnecessary to check the global setting for existing volumes.
That's my understanding and please let me know if I'm missing anything.

webhook/resources/snapshot/validator.go
carterli0407-cell: I've updated.

webhook/resources/snapshot/validator.go
carterli0407-cell: Updated

webhook/resources/snapshot/validator_test.go
carterli0407-cell: Removed

webhook/resources/snapshot/validator_test.go
carterli0407-cell: Thx. I've updated.

webhook/resources/snapshot/validator.go
carterli0407-cell: Because that is the field the user set to trigger snapshot creation. When they set spec.createSnapshot: true and the max count is reached, that specific field is the root cause of the rejection, it's the action the user requested that we're blocking.

derekbit: Sounds weird because the second field usually means it is invalid. In this case, it is not the root cause of the error. Could you remove it to avoid the confusion?

webhook/resources/snapshot/validator.go
derekbit: > Is it possible that snapshot is nil?

If you go over the `Create` and `Update` codes. It is impossible that `snapshot` is `nil` because it is deferenced before. Please remove the unnecessary check.

webhook/resources/snapshot/validator.go
derekbit: > The false to true guard on the Update path exists because a user can activate a previously passive CR by flipping the field via an update rather than a create.

Can you clarify if it is a valid operation? If it is not valid and not expected, we should reject the update of `spec.createSnapshot`.
In addition, IMO, we don't need to validateSnapshotCount in the UPDATE, because the snapshot CR is already created. The validation is meaningless.

cc @shuo-wu @COLDTURNIP @PhanLe1010

webhook/resources/snapshot/validator.go
derekbit: I see. Could you add a comment to explain the validation?

webhook/resources/snapshot/validator.go
```diff
@@ -7,6 +7,7 @@ import (
 	admissionregv1 "k8s.io/api/admissionregistration/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 
+	"github.com/cockroachdb/errors"
 	"github.com/longhorn/longhorn-manager/datastore"
 	"github.com/longhorn/longhorn-manager/types"
 	"github.com/longhorn/longhorn-manager/util"
```
derekbit, line 17: Reorganize the imports as https://github.com/longhorn/longhorn/wiki/coding-convention#organize-package-imports

webhook/resources/snapshot/validator.go
derekbit: > Why do we need to check snapshot.Spec.CreateSnapshot?

https://github.com/longhorn/longhorn-manager/pull/4697#discussion_r3175792463

webhook/resources/snapshot/validator.go
```diff
@@ -84,5 +89,81 @@ func (o *snapshotValidator) Update(request *admission.Request, oldObj runtime.Ob
...
+
+func (o *snapshotValidator) validateSnapshotCount(snapshot *longhorn.Snapshot) error {
+	if snapshot == nil || !snapshot.Spec.CreateSnapshot {
+		return nil
+	}
+
+	volume, err := o.ds.GetVolumeRO(snapshot.Spec.Volume)
+	if err != nil {
+		return errors.Wrapf(err, "failed to get volume %v", snapshot.Spec.Volume)
+	}
```
derekbit, line 108:
```suggestion
	if err != nil {
		return errors.Wrapf(err, "failed to get volume %v for ....
	}
```

webhook/resources/snapshot/validator.go
```diff
@@ -84,5 +89,81 @@ func (o *snapshotValidator) Update(request *admission.Request, oldObj runtime.Ob
...
+
+	snapshotMaxCount := volume.Spec.SnapshotMaxCount
+	if snapshotMaxCount <= 0 {
+		configuredSnapshotMaxCount, err := o.ds.GetSettingAsInt(types.SettingNameSnapshotMaxCount)
+		if err != nil {
+			return errors.Wrapf(err, "failed to get setting %v", types.SettingNameSnapshotMaxCount)
```
derekbit:
```suggestion
			return errors.Wrapf(err, "failed to get setting %v for ...
```

webhook/resources/snapshot/validator.go
```diff
@@ -84,5 +89,81 @@ func (o *snapshotValidator) Update(request *admission.Request, oldObj runtime.Ob
...
+		}
+		snapshotMaxCount = getEffectiveSnapshotMaxCount(volume.Spec.SnapshotMaxCount, int(configuredSnapshotMaxCount))
+	}
+
+	engines, err := o.ds.ListVolumeEnginesRO(snapshot.Spec.Volume)
+	if err != nil {
+		return errors.Wrapf(err, "can't list engines for volume %s", snapshot.Spec.Volume)
```
derekbit:
```suggestion
		return errors.Wrapf(err, "can't list engines for volume %s for ...
```

webhook/resources/snapshot/validator.go
```diff
@@ -84,5 +89,81 @@ func (o *snapshotValidator) Update(request *admission.Request, oldObj runtime.Ob
...
+	}
+
+	snapshotMaxCount := volume.Spec.SnapshotMaxCount
+	if snapshotMaxCount <= 0 {
+		configuredSnapshotMaxCount, err := o.ds.GetSettingAsInt(types.SettingNameSnapshotMaxCount)
+		if err != nil {
+			return errors.Wrapf(err, "failed to get setting %v", types.SettingNameSnapshotMaxCount)
+		}
+		snapshotMaxCount = getEffectiveSnapshotMaxCount(volume.Spec.SnapshotMaxCount, int(configuredSnapshotMaxCount))
+	}
```
derekbit: The original logic is a bit messy
```
    snapshotMaxCount := volume.Spec.SnapshotMaxCount
	if snapshotMaxCount <= 0 {
		configuredSnapshotMaxCount, err := o.ds.GetSettingAsInt(types.SettingNameSnapshotMaxCount)
		if err != nil {
			return errors.Wrapf(err, "failed to get setting %v", types.SettingNameSnapshotMaxCount) 
		}
		snapshotMaxCount = getEffectiveSnapshotMaxCount(volume.Spec.SnapshotMaxCount, int(configuredSnapshotMaxCount))
	}
```

You can use the logic like

```suggestion
	snapshotMaxCount := o.getSnapshotMaxCount(..)
	...

func (o *snapshotValidator) getSnapshotMaxCount(...) ... {
    snapshotMaxCount := volume.Spec.SnapshotMaxCount
    if snapshotMaxCount > 0 {
        return snapshotMaxCount
    }

    snapshotMaxCount, err := o.ds.GetSettingAsInt(types.SettingNameSnapshotMaxCount)
    if err != nil {
        return errors.Wrapf(err, "failed to get setting %v", types.SettingNameSnapshotMaxCount) 
    }
    return snapshotMaxCount
}	
```

webhook/resources/snapshot/validator.go
derekbit: My question is not addressed. The logic is still to get the snapshots from the data path directly.
Why not just calculate the number of the existing and ready-to-use snapshot CRs? If the number of existing snapshot CRs exceed the threshold, it errors out.

Pseudo code. WDYT? @COLDTURNIP @shuo-wu 
```
       snapshots, err := ListVolumeSnapshotsRO()
       if err != nil {
          return ....
       }

       countNumberOfReadyToUseSnapshotCRs
       if  numberOfReadyToUseSnapshotCRs >= max count {
          return ...
       }

       ...
```

COLDTURNIP:
I think we should distinguish the case between system and user created snapshots, e.g. attach some annotation like `create-snapshot-from-longhorn`. If it reaches the snapshot count limit, we should still allow the system to create the new Snapshot CR, to finish the operations like volume expansion, or to reflect the actual file status.

webhook/resources/snapshot/validator.go
COLDTURNIP: Similar to `Spec.Volume`, I think `Spec.CreateSnapshot` is immutable. We shall focus on the Snapshot CR creation in this story, or mark as immutable in the CRD instead.

webhook/resources/snapshot/validator.go
COLDTURNIP: I think checking `Spec.CreateSnapshot` is more like to distinguish the case from system and user created snapshots. We should check the count for user created snapshot, and always allow the system created one. But, it is still possible that the system creates a snapshot with `Spec.CreateSnapshot = true`, e.g. system backup. Maybe we should introduce a new annotation like `create-snapshot-from-longhorn` for this check.

COLDTURNIP (changes requested):

webhook/resources/snapshot/validator.go
carterli0407-cell: Updated.

webhook/resources/snapshot/validator.go
carterli0407-cell: Updated.

webhook/resources/snapshot/validator.go
carterli0407-cell: Removed.

webhook/resources/snapshot/validator.go
carterli0407-cell: Done.

webhook/resources/snapshot/validator.go
carterli0407-cell: Apologies for the confusion in my previous reply. I've now replaced the entire engine iteration block with ListVolumeSnapshotsRO and count only entries where snap.Status.ReadyToUse == true.
This removes the dependency on the data path entirely. The getSnapshotCountUsage helper is also removed as it is no longer needed.

webhook/resources/snapshot/validator.go
carterli0407-cell: Agreed and fixed.

webhook/resources/snapshot/validator.go
carterli0407-cell: Done.

webhook/resources/snapshot/validator.go
carterli0407-cell: Updated

webhook/resources/snapshot/validator.go
carterli0407-cell: Updated

webhook/resources/snapshot/validator.go
carterli0407-cell: Done

webhook/resources/snapshot/validator.go
```diff
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"reflect"
 
+	"github.com/cockroachdb/errors"
+
 	admissionregv1 "k8s.io/api/admissionregistration/v1"
 	"k8s.io/apimachinery/pkg/runtime"
```
derekbit: Reorganize the imports

derekbit (commented):
In general, LGTM. Only one comment.

webhook/resources/snapshot/validator.go
```diff
@@ -86,3 +92,63 @@ func (o *snapshotValidator) Update(request *admission.Request, oldObj runtime.Ob
 
 	return nil
 }
+
+func (o *snapshotValidator) validateSnapshotCount(snapshot *longhorn.Snapshot) error {
+	if !snapshot.Spec.CreateSnapshot {
+		return nil
+	}
```
COLDTURNIP, line 103: This may [block the system backup](https://github.com/longhorn/longhorn-manager/blob/master/controller/system_backup_controller.go#L1010). Should we focus only on user-created snapshots? ([previous comments](https://github.com/longhorn/longhorn-manager/pull/4697#discussion_r3181003118))

> I think checking `Spec.CreateSnapshot` is more like to distinguish the case from system and user created snapshots. We should check the count for user created snapshot, and always allow the system created one. But, it is still possible that the system creates a snapshot with `Spec.CreateSnapshot = true`, e.g. system backup. Maybe we should introduce a new annotation like `create-snapshot-from-longhorn` for this check.

webhook/resources/snapshot/validator.go
```diff
@@ -86,3 +92,63 @@ func (o *snapshotValidator) Update(request *admission.Request, oldObj runtime.Ob
...
+	return int(snapshotMaxCount), nil
+}
+
+func countReadyToUseSnapshots(snapshots map[string]*longhorn.Snapshot) int {
+	count := 0
+	for _, s := range snapshots {
+		if s != nil && s.Status.ReadyToUse {
+			count++
+		}
```
COLDTURNIP: Do we really need this check? We may miss the in-progress snapshots.

derekbit: Good question. I'm thinking if we need to take the non-ReadyToUse snapshots into account. WDYT? @shuo-wu

webhook/resources/snapshot/validator.go
derekbit: Good idea!

webhook/resources/snapshot/validator.go
carterli0407-cell: Done

webhook/resources/snapshot/validator.go
carterli0407-cell: Good catch. Counting only ReadyToUse snapshots misses in-progress ones, allowing concurrent creates to bypass the limit.

I updated the condition with:
if s != nil && s.DeletionTimestamp == nil && !s.Status.MarkRemoved {
    count++
}

@COLDTURNIP , WDYT?

webhook/resources/snapshot/validator.go
carterli0407-cell: @COLDTURNIP, to fix this properly, we'd need to add a label to system backup snapshots so the admission webhook can recognize and skip them during the count check. That's my thought and I'd love to hear your opinion.

webhook/resources/snapshot/validator.go
```diff
@@ -86,3 +93,63 @@ func (o *snapshotValidator) Update(request *admission.Request, oldObj runtime.Ob
...
+	}
+
+	snapshotMaxCount, err := o.ds.GetSettingAsInt(types.SettingNameSnapshotMaxCount)
+	if err != nil {
+		return 0, errors.Wrapf(err, "failed to get setting %v for snapshot count validation", types.SettingNameSnapshotMaxCount)
+	}
+	return int(snapshotMaxCount), nil
+}
+
+func countReadyToUseSnapshots(snapshots map[string]*longhorn.Snapshot) int {
```
derekbit: Function name is incorrect

webhook/resources/snapshot/validator.go
```diff
@@ -86,3 +93,63 @@ func (o *snapshotValidator) Update(request *admission.Request, oldObj runtime.Ob
...
+	count := 0
+	for _, s := range snapshots {
+		if s != nil && s.DeletionTimestamp == nil && !s.Status.MarkRemoved {
```
derekbit: Don't we skip system snapshots as @COLDTURNIP mentioned?

webhook/resources/snapshot/validator.go
carterli0407-cell: Done.

webhook/resources/snapshot/validator.go
carterli0407-cell: I added s.Status.UserCreated condition.

webhook/resources/snapshot/validator.go
carterli0407-cell: @derekbit , @COLDTURNIP , I would like to discuss and hear your thoughts for this part.
Could you advise on the preferred approach here?

derekbit: How does snapshotCreate handle system snapshots?

webhook/resources/snapshot/validator.go
derekbit: I just realized the `UserCreated` flag is in the status. That means the `UserCreated` is always false while validating it.

webhook/resources/snapshot/validator.go
```diff
@@ -86,3 +93,63 @@ func (o *snapshotValidator) Update(request *admission.Request, oldObj runtime.Ob
...
+	if err != nil {
+		return 0, errors.Wrapf(err, "failed to get setting %v for snapshot count validation", types.SettingNameSnapshotMaxCount)
+	}
+	return int(snapshotMaxCount), nil
+}
+
+func countExistingSnapshots(snapshots map[string]*longhorn.Snapshot) int {
+	count := 0
+	for _, s := range snapshots {
+		if s != nil && s.DeletionTimestamp == nil && !s.Status.MarkRemoved && s.Status.UserCreated {
```
derekbit, line 153: As I mentioned, `s.Status.UserCreated` is always false. It is meaningless to validate it.
`s.Status.MarkRemoved` has the same issue.

derekbit: Can you check how the snapshotMaxCount is validated in the data path? Can you create snapshots by `SnapshotCreate` to hit `snapshotMaxCount` and then trigger the replica rebuilding? Check if the replica rebuilding will succeed or fail.

Should we really need to takes.Status.MarkRemoved and s.Status.UserCreated into consideration? WDYT @shuo-wu @COLDTURNIP

webhook/resources/snapshot/validator.go
derekbit: > @derekbit , @COLDTURNIP , I would like to discuss and hear your thoughts for this part. Could you advise on the preferred approach here?

@carterli0407-cell Please check https://github.com/longhorn/longhorn-manager/pull/4697#discussion_r3192183232 and do some testing first.

webhook/resources/snapshot/validator.go
shuo-wu: I think the key point is whether there is really a snapshot file/lvol there. As long as there is really a snap entity, I prefer to count it as an existing one without checking `Status.MarkRemoved` or `Status.UserCreated` or even `DeletionTimestamp`.

derekbit: @shuo-wu Good idea, but how to know if it is a snap entity while creating a CR?

shuo-wu: Just double-checked the code, it seems that the strongest evidence of a snap entity is the presence of ` engine.Status.Snapshots[snapshotName]`, which is directly reported by the engine/replica itself. I am not sure whether we can utilize it here

PR #4712, opened 2026-05-05
fix: backfill v2 expansion failures

davidcheng0922:

https://github.com/longhorn/longhorn/issues/12903

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced error handling for volume expansion operations in data engine v2. The system now accurately detects expansion failures and automatically records detailed failure information including error messages and exact failure timestamps, providing improved visibility and diagnostics for expansion-related issues.

### DIFF
--- a/controller/engine_controller.go
+++ b/controller/engine_controller.go
@@ -386,9 +386,16 @@ func (ec *EngineController) syncEngine(key string) (err error) {
 				return err
 			}
 		}
-	} else if ec.isMonitoring(engine) {
-		// engine is not running
-		ec.resetAndStopMonitoring(engine)
+	} else {
+		if ec.isMonitoring(engine) {
+			// engine is not running
+			ec.resetAndStopMonitoring(engine)
+		}
+		if types.IsDataEngineV2(engine.Spec.DataEngine) {
+			if engine.Status.CurrentState == longhorn.InstanceStateError {
+				ec.fillV2ExpansionFailureFromInstanceError(engine)
+			}
+		}
 	}
 
 	if err := ec.syncSnapshotCRs(engine); err != nil {
@@ -528,6 +535,35 @@ func failedCloneBefore(e *longhorn.Engine) bool {
 		}
 	}
 	return false
+}
+
+func (ec *EngineController) fillV2ExpansionFailureFromInstanceError(engine *longhorn.Engine) {
+	if engine.Status.LastExpansionError != "" || engine.Status.LastExpansionFailedAt != "" {
+		return
+	}
+
+	if !isV2ExpansionIncomplete(engine) {
+		return
+	}
+
+	errMsg := ""
+	instance, err := ec.GetInstance(engine)
+	if err != nil {
+		errMsg = fmt.Sprintf("engine instance could not be reached; unable to determine whether expansion completed successfully, and the result has not yet been reflected in engine status: %v", err)
+	} else if instance == nil {
+		errMsg = "engine instance could not be reached; unable to determine whether expansion completed successfully, and the result has not yet been reflected in engine status"
+	} else {
+		instanceErrMsg := instance.Status.ErrorMsg
+		if instanceErrMsg == "" {
+			errMsg = "unable to determine whether expansion completed successfully before the engine entered error, and the result has not yet been reflected in engine status"
+		} else {
+			errMsg = fmt.Sprintf("unable to determine whether expansion completed successfully before the engine entered error; instance error: %v", instanceErrMsg)
+		}
+	}
+	// The engine is already in error at this stage, and volume information is unavailable,
+	// so use the current engine instance error message to backfill the expansion failure.
+	engine.Status.LastExpansionError = errMsg
+	engine.Status.LastExpansionFailedAt = util.Now()
 }
 
 func (ec *EngineController) enqueueEngine(obj interface{}) {
@@ -3074,4 +3110,11 @@ func isV2ReplicaAddAlreadyInProgressError(err error) bool {
 
 func isV2ReplicaAddRestoreInProgressError(err error) bool {
 	return err != nil && strings.Contains(err.Error(), "restore is in progress")
+}
+
+func isV2ExpansionIncomplete(engine *longhorn.Engine) bool {
+	// Do not rely on IsExpanding here. If spdk_tgt is already down, that runtime
+	// state may be stale and stay true, which would prevent this fallback path
+	// from working.
+	return engine.Spec.VolumeSize > engine.Status.CurrentSize
 }


controller/engine_controller.go
```diff
@@ -386,9 +386,14 @@ func (ec *EngineController) syncEngine(key string) (err error) {
...
-		// engine is not running
-		ec.resetAndStopMonitoring(engine)
+	} else {
+		if engine.Status.CurrentState == longhorn.InstanceStateError {
+			ec.fillV2ExpansionFailureFromCondition(engine)
+		}
+		if ec.isMonitoring(engine) {
+			// engine is not running
+			ec.resetAndStopMonitoring(engine)
+		}
```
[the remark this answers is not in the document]
davidcheng0922: Fix

controller/instance_handler.go
```diff
@@ -485,7 +485,7 @@ func (h *InstanceHandler) ReconcileInstanceState(obj interface{}, spec *longhorn
 		if instance, exists := instances[instanceName]; exists {
 			// If instance is in error state and the ErrorMsg contains 61 (ENODATA) error code, then it indicates
 			// the creation of engine process failed because there is no available backend (replica).
-			if spec.DesireState == longhorn.InstanceStateRunning {
+			if spec.DesireState == longhorn.InstanceStateRunning || spec.DesireState == longhorn.InstanceStateStopped {
```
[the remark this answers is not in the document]
davidcheng0922: Remove it

controller/engine_controller.go
```diff
@@ -386,9 +386,14 @@ func (ec *EngineController) syncEngine(key string) (err error) {
...
-		// engine is not running
-		ec.resetAndStopMonitoring(engine)
+	} else {
+		if ec.isMonitoring(engine) {
+			// engine is not running
+			ec.resetAndStopMonitoring(engine)
+		}
+		if engine.Status.CurrentState == longhorn.InstanceStateError {
+			ec.fillV2ExpansionFailureFromInstanceError(engine)
+		}
```
derekbit, line 398: How about
```suggestion
		if types.IsDataEngineV2(engine.Spec.DataEngine) {
		    if engine.Status.CurrentState == longhorn.InstanceStateError {
			    ec.fillExpansionFailureFromInstanceError(engine)
			}
		}
```

controller/engine_controller.go
```diff
@@ -530,6 +535,28 @@ func failedCloneBefore(e *longhorn.Engine) bool {
 	return false
 }
 
+func (ec *EngineController) fillV2ExpansionFailureFromInstanceError(engine *longhorn.Engine) {
+	if engine == nil || !types.IsDataEngineV2(engine.Spec.DataEngine) {
+		return
+	}
```
derekbit: Remove the checks and move them to outside of the function

controller/engine_controller.go
```diff
@@ -530,6 +535,28 @@ func failedCloneBefore(e *longhorn.Engine) bool {
...
+	if engine.Spec.VolumeSize <= engine.Status.CurrentSize {
+		return
+	}
+
+	instance, err := ec.GetInstance(engine)
+	if err != nil || instance == nil || instance.Status.ErrorMsg == "" {
+		return
+	}
```
derekbit, line 562: How about 
```suggestion
	if err != nil {
        log.withError(err).Warnf(....)
        return
    }
    if instance == nil || instance.Status.ErrorMsg == "" {
		return
	}
```

Will the log message be flooding?

Another question: Should we update LastExpansionError and LastExpansionFailedAt if instance is not found?

controller/engine_controller.go
davidcheng0922: Address, thank you

controller/engine_controller.go
davidcheng0922: Address, thank you

controller/engine_controller.go
davidcheng0922: The log could be noisy, so I chose to still populate the engine expansion failure message even when fetching the instance fails, to avoid leaving the expansion failure information empty. WDYT?

controller/engine_controller.go
```diff
@@ -530,6 +537,28 @@ func failedCloneBefore(e *longhorn.Engine) bool {
...
+	if engine.Status.LastExpansionError != "" || engine.Status.LastExpansionFailedAt != "" {
+		return
+	}
+	if engine.Spec.VolumeSize <= engine.Status.CurrentSize {
+		return
+	}
+
+	errMsg := ""
+	instance, err := ec.GetInstance(engine)
+	if err != nil || instance == nil {
```
mantissahz: It can return another error when using `ec.GetInstance()`, such as `ec.ds.GetInstanceManagerRO(e.Status.InstanceManagerName)` in `ec.GetInstance()`.
Should we separate it as @derekbit suggested?

controller/engine_controller.go
```diff
@@ -386,9 +386,16 @@ func (ec *EngineController) syncEngine(key string) (err error) {
 				return err
 			}
 		}
-	} else if ec.isMonitoring(engine) {
-		// engine is not running
-		ec.resetAndStopMonitoring(engine)
+	} else {
+		if ec.isMonitoring(engine) {
+			// engine is not running
+			ec.resetAndStopMonitoring(engine)
```
derekbit, line 392: I just realized `resetAndStopMonitoring()` cleans up `LastExpansionError` and `LastExpansionFailedAt`. Does it matter?

controller/engine_controller.go
derekbit: If the instance is not found (nil). Does it mean the expansion must be failed? In this case, are we able to get the expansion error message?

davidcheng0922: If the instance is not found, it does not necessarily mean the expansion itself is the direct failure point. 

However, at this stage we can no longer reliably verify completion from engine/volume information, while the expansion result has not yet been reflected in Engine status. Therefore, we backfill `LastExpansionError` with a fallback message derived from the current instance error path.

controller/engine_controller.go
davidcheng0922: Thank you, refactor this part

controller/engine_controller.go
davidcheng0922: Yes, it does matter. `resetAndStopMonitoring()` clears `LastExpansionError` and `LastExpansionFailedAt`, so I  stop monitoring first and then backfill the expansion failure afterward.

controller/engine_controller.go
davidcheng0922: I refactored the log messages to distinguish the different cases. The messages are intended to make the uncertainty explicit to the user

controller/engine_controller.go
```diff
@@ -530,6 +537,33 @@ func failedCloneBefore(e *longhorn.Engine) bool {
...
+	switch {
+	case err != nil:
+		errMsg = fmt.Sprintf("engine instance could not be reached; unable to determine whether expansion completed successfully, and the result has not yet been reflected in engine status: %v", err)
+	case instance == nil:
+		errMsg = "engine instance could not be reached; unable to determine whether expansion completed successfully, and the result has not yet been reflected in engine status"
+	case instance.Status.ErrorMsg == "":
+		errMsg = "unable to determine whether expansion completed successfully before the engine entered error, and the result has not yet been reflected in engine status"
+	default:
+		errMsg = fmt.Sprintf("unable to determine whether expansion completed successfully before the engine entered error; instance error: %v", instance.Status.ErrorMsg)
+	}
```
derekbit: How about this? The current `switch` is ordering-sensitive. If someone changes the case order in the future, it could accidentally introduce a nil pointer dereference or make the logic harder to maintain.

```suggestion
	if err != nil {
		errMsg = fmt.Sprintf("engine instance could not be reached; unable to determine whether expansion completed successfully, and the result has not yet been reflected in engine status: %v", err)
	} else if instance == nil {
		errMsg = "engine instance could not be reached; unable to determine whether expansion completed successfully, and the result has not yet been reflected in engine status"
	} else {
		instanceErrMsg := instance.Status.ErrorMsg
		if instanceErrMsg == "" {
			errMsg = "unable to determine whether expansion completed successfully before the engine entered error, and the result has not yet been reflected in engine status"
		} else {
			errMsg = fmt.Sprintf("unable to determine whether expansion completed successfully before the engine entered error; instance error: %v", instanceErrMsg)
		}
	}
```

davidcheng0922: More reasonable, fix it. Thank you

controller/engine_controller.go
```diff
@@ -530,6 +537,34 @@ func failedCloneBefore(e *longhorn.Engine) bool {
 	return false
 }
 
+func (ec *EngineController) fillV2ExpansionFailureFromInstanceError(engine *longhorn.Engine) {
+	if engine.Status.LastExpansionError != "" || engine.Status.LastExpansionFailedAt != "" {
+		return
+	}
+	if engine.Spec.VolumeSize <= engine.Status.CurrentSize {
+		return
+	}
```
derekbit:
```suggestion
	if !isV2ExpansionIncomplete(engine) {
		return
	}
```

derekbit: Should we check `IsExpanding` as well?

```
func isV2ExpansionIncomplete(engine *longhorn.Engine) bool {
	return engine.Spec.VolumeSize > engine.Status.CurrentSize || engine.Status.IsExpanding
}
```

davidcheng0922: I don't think we should rely on `IsExpanding` here. If `spdk_tgt` is down, this field may become stale and remain true without being updated, which makes the backfill logic work incorrectly.

derekbit (approved):

derekbit:
@mergify backport v1.12.x

PR #4715, opened 2026-05-06
fix(snapshot): delete snapshot without attaching volume if possible

COLDTURNIP:

Issue longhorn/longhorn#12489

### DIFF
--- a/controller/snapshot_controller.go
+++ b/controller/snapshot_controller.go
@@ -51,6 +51,9 @@ type SnapshotController struct {
 	cacheSyncs             []cache.InformerSynced
 	engineClientCollection engineapi.EngineClientCollection
 
+	// for unit test
+	nowHandler func() string
+
 	proxyConnCounter util.Counter
 }
 
@@ -82,6 +85,7 @@ func NewSnapshotController(
 		eventRecorder:          eventBroadcaster.NewRecorder(scheme, corev1.EventSource{Component: "longhorn-snapshot-controller"}),
 		ds:                     ds,
 		engineClientCollection: engineClientCollection,
+		nowHandler:             util.Now,
 		proxyConnCounter:       proxyConnCounter,
 	}
 
@@ -442,6 +446,14 @@ func (sc *SnapshotController) reconcile(snapshotName string) (err error) {
 			return err
 		}
 
+		if snapshot.Spec.CreateSnapshot && snapshot.Status.RequestedTime == "" {
+			// The snapshot has not been taken at the engine side. It is safe to cancel the attachment and delete the CR.
+			if err = sc.handleAttachmentTicketDeletion(snapshot); err != nil {
+				return err
+			}
+			return sc.ds.RemoveFinalizerForSnapshot(snapshot)
+		}
+
 		if isEngineUpgrading(engine) {
 			// requeue the snapshot to wait for upgrading engine after 3 seconds
 			snapshot.Status.Error = fmt.Sprintf("snapshot deletion delayed: volume engine %v is upgrading from image %v to %v", engine.Name, engine.Status.CurrentImage, engine.Spec.Image)
@@ -455,7 +467,7 @@ func (sc *SnapshotController) reconcile(snapshotName string) (err error) {
 				return
 			}
 			if snapshotInfo, ok := engine.Status.Snapshots[snapshot.Name]; ok {
-				if err := syncSnapshotWithSnapshotInfo(snapshot, snapshotInfo, engine.Spec.VolumeSize); err != nil {
+				if err := syncSnapshotWithSnapshotInfo(sc.logger, snapshot, snapshotInfo, engine.Spec.VolumeSize); err != nil {
 					return
 				}
 			}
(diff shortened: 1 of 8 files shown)


controller/snapshot_controller.go
```diff
@@ -442,6 +446,14 @@ func (sc *SnapshotController) reconcile(snapshotName string) (err error) {
 			return err
 		}
 
+		if snapshot.Spec.CreateSnapshot && snapshot.Status.RequestedTime == "" {
+			// The snapshot has not been taken at the engine side. It is safe to cancel the attachment and delete the CR.
+			if err = sc.handleAttachmentTicketDeletion(snapshot); err != nil {
+				return err
+			}
+			return sc.ds.RemoveFinalizerForSnapshot(snapshot)
```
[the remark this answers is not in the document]
COLDTURNIP: `snapshot.Spec.CreateSnapshot == true` means the Snapshot CR is created by the user or system backup controller, and it should always record the timestamp when the Snapshot CR initiate a snapshot cutting to the engine. `RequestedTime == ""` implies the cutting has not been initiated.

controller/snapshot_controller.go
```diff
@@ -442,6 +446,14 @@ func (sc *SnapshotController) reconcile(snapshotName string) (err error) {
...
+		}
```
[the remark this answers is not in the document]
COLDTURNIP: - pre-v1.13 snapshots before backfill: backfilled by upgrade
- status update was missed: a warning message prints in `syncSnapshotWithSnapshotInfo`

controller/snapshot_controller.go
```diff
@@ -724,6 +736,12 @@ func syncSnapshotWithSnapshotInfo(snap *longhorn.Snapshot, snapInfo *longhorn.Sn
 	snap.Status.Size = size
 	snap.Status.Labels = snapInfo.Labels
 	snap.Status.RestoreSize = restoreSize
+	if snap.Status.RequestedTime == "" && snapInfo.Created != "" {
+		if snapInfo.UserCreated {
+			logger.Warnf("Backfilling RequestedTime for user-created snapshot %v from engine creation time %v; the snapshot controller may have lost track of the creation request", snap.Name, snapInfo.Created)
+		}
+		snap.Status.RequestedTime = snapInfo.Created
+	}
```
shuo-wu, line 744: NIT: 
Will this be triggered by below cases? 
    1. Somehow `engineClientProxy.SnapshotGet(engine, snapshot.Name)` errors out but the request actually gets handled by the engine instance?
    2. There are existing snapshot files before we have corresponding CRs

COLDTURNIP: > 1. Somehow `engineClientProxy.SnapshotGet(engine, snapshot.Name)` errors out but the request actually gets handled by the engine instance?

To the mentioned code block, it tries to fill back the `snapshot.Status.RequestedTime`. It typically happens for system-created snapshots. The `snapshot SnapshotInfo` instance here came from `engine.Status`, which is fetched using the engine proxy. To the worst case that [the Engine controller failed to fetch the snapshot list](https://github.com/longhorn/longhorn-manager/blob/master/controller/engine_controller.go#L1085-L1087), all the SnapshotInfos keep out-dated, then `snapInfo.Created` keeps empty.

> 2. There are existing snapshot files before we have corresponding CRs

In case that snapshot files are created before Snapshot CR creation, the Snapshot CR [is always created by Engine controller](https://github.com/longhorn/longhorn-manager/blob/master/controller/engine_controller.go#L1581-L1588) with `snapshot.Spec.CreateSnapshot = false`. This case is excluded in this PR while handling Snapshot CR deletion and follow the original path:

```go
		if snapshot.Spec.CreateSnapshot && snapshot.Status.RequestedTime == "" {
			// The snapshot has not been taken at the engine side. It is safe to cancel the attachment and delete the CR.
```

controller/snapshot_controller.go
```diff
@@ -771,7 +789,9 @@ func (sc *SnapshotController) handleSnapshotCreate(snapshot *longhorn.Snapshot,
 		return err
 	}
 	if snapshotInfo == nil {
-		sc.logger.Infof("Creating snapshot %v of volume %v", snapshot.Name, snapshot.Spec.Volume)
+		now := sc.nowHandler()
+		sc.logger.Infof("Creating snapshot %v of volume %v at %s", snapshot.Name, snapshot.Spec.Volume, now)
+		snapshot.Status.RequestedTime = now
```
davidcheng0922: Question, we assign the `snapshot.Status.RequestedTime` before calling create snapshot. 

What happens if snapshot creation fails? In that case, `RequestedTime` is still set even though the snapshot may never have been created on the engine side. Is it matters?

COLDTURNIP: Thanks, reordered after the request succeed.

davidcheng0922 (approved):
Overall, LGTM

derekbit (approved):

derekbit:
@mergify backport v1.12.x

PR #4716, opened 2026-05-07
fix(v2): remove hasRecentlyHealthyExtraReplica to unblock eviction cleanup

derekbit:

Issue Longhorn/longhorn#13058

The hasRecentlyHealthyExtraReplica function delays extra healthy replica cleanup by 1 minute for v2 volumes. During node drain/eviction, this delay prevents cleanupEvictionRequestedReplicas from removing the old evicting replica before the drain completes. After uncordon, the node controller cancels eviction (EvictionRequested=false) since the node is no longer Unschedulable, leaving a surplus replica that no cleanup path can remove.

This does not affect v1 behavior as the function was gated behind IsDataEngineV2.

## Summary by CodeRabbit

* **Refactor**
  * Simplified volume controller replica cleanup logic by removing the deferred cleanup mechanism for recently healthy replicas, eliminating unnecessary cleanup delays.
  * Updated related tests to align with the streamlined cleanup behavior.

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -51,8 +51,7 @@ var (
 	RetryInterval = 100 * time.Millisecond
 	RetryCounts   = 20
 
-	AutoSalvageTimeLimit             = 1 * time.Minute
-	RecentHealthyReplicaCleanupDelay = 1 * time.Minute
+	AutoSalvageTimeLimit = 1 * time.Minute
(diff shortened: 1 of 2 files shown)


davidcheng0922 (approved):

derekbit:
@mergify backport v1.12.x

PR #4717, opened 2026-05-07
fix: lazy umount LUKS device before attaching a volume to a node

mantissahz:

Issue # longhorn/longhorn#11510

We unmount the LUSK device in the volume controller that is the owner/attaching node of the volume if the engine crashed unexpectedly, and the volume was re-attached to the same node.

It can not do lazy unmount the LUKS device at the CSI plugin level before sending an attaching volume request to the Longhorn server in `ControllerPublishVolume` because the CSI request might be received by a CSI plugin pod on the node (maybe node A) where the LUKS device is not present (maybe node B).

And we need to unmount the LUKS device before attaching.

Ran the customized e2e test case 3 times and `PASSED`
~https://ci.longhorn.io/job/private/job/longhorn-e2e-test/7726~
~https://ci.longhorn.io/job/private/job/longhorn-e2e-test/7744/~
https://ci.longhorn.io/job/private/job/longhorn-e2e-test/7749

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -36,6 +36,7 @@ import (
 	imutil "github.com/longhorn/longhorn-instance-manager/pkg/util"
 
 	"github.com/longhorn/longhorn-manager/constant"
+	"github.com/longhorn/longhorn-manager/csi/crypto"
 	"github.com/longhorn/longhorn-manager/datastore"
 	"github.com/longhorn/longhorn-manager/engineapi"
 	"github.com/longhorn/longhorn-manager/scheduler"
@@ -1898,6 +1899,20 @@ func (c *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[strin
 				// See https://github.com/longhorn/longhorn/issues/9089
 				if err := c.handleDelinquentAndStaleStateForFaultedRWXVolume(v); err != nil {
 					return err
+				}
+				// When the engine crashed unexpectedly, the LUKS device might still be mounted and the underlayer device is gone if the volume is encrypted.
+				// This can cause the subsequent attach to fail because the LUKS device is stale in use and ISCSI target can not be logged out.
+				// We can not lazy unmount the LUKS device at the CSI plugin level before sending an attaching volume request
+				// to the Longhorn server in ControllerPublishVolume because the CSI request might be received
+				// by a CSI plugin pod on the node (maybe node A) where the LUKS device is not present (maybe node B).
+				// After an unexpected engine crash, a stale LUKS device mount may still be present.
+				// Lazy unmount it here to clear stale mount state before re-attaching the volume.
+				if v.Spec.Encrypted {
+					log.Infof("Volume is encrypted, lazy unmounting crypto device for volume %v if it is still mounted", v.Name)
+					cryptoDevice := crypto.VolumeMapper(v.Name, string(v.Spec.DataEngine))
+					if err := util.LazyUnmount(cryptoDevice); err != nil {
+						return errors.Wrapf(err, "failed to lazy unmount crypto device %v for volume %v", cryptoDevice, v.Name)
+					}
 				}
 			}
 		}
--- a/util/util.go
+++ b/util/util.go
@@ -945,3 +945,42 @@ func IsCryptsetupVerWithFixed16MiBHeaderSize() (bool, error) {
 
 	return nsexec.IsLuksFixed16MiBHeaderSize()
 }
+
+func LazyUnmount(path string) error {
+	if _, err := lhns.Stat(path); err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			return nil
+		}
+		logrus.WithError(err).Warnf("Failed to stat path %s, will continue lazy unmounting it", path)
+	}
+
+	namespaces := []lhtypes.Namespace{lhtypes.NamespaceMnt, lhtypes.NamespaceNet}
+	nsexec, err := lhns.NewNamespaceExecutor(lhtypes.ProcessNone, lhtypes.HostProcDirectory, namespaces)
+	if err != nil {
+		return err
+	}
+
+	_, err = nsexec.Execute(nil, "umount", []string{"-l", path}, lhtypes.ExecuteDefaultTimeout)
+	if err != nil {
+		if isUnmountedError(err) {
+			// The device is already unmounted. We can safely ignore the error and treat it as a success!
+			logrus.WithError(err).Debugf("Device %s is already unmounted.", path)
+			return nil
+		}
+		return errors.Wrapf(err, "failed to unmount %s", path)
+	}
+
+	logrus.Debugf("Lazy unmounted device %s without the error.", path)
+	return nil
+}
+
+func isUnmountedError(err error) bool {
+	if err == nil {
+		return false
+	}
+
+	errMsg := strings.ToLower(err.Error())
+	return strings.Contains(errMsg, "not mounted") ||
+		strings.Contains(errMsg, "no mount point specified") ||
+		strings.Contains(errMsg, "no such file or directory")
+}


util/util.go
```diff
@@ -945,3 +945,31 @@ func IsCryptsetupVerWithFixed16MiBHeaderSize() (bool, error) {
...
+	}
+
+	output, err := nsexec.Execute(nil, "umount", []string{"-l", path}, lhtypes.ExecuteDefaultTimeout)
+	if err != nil {
+		if strings.Contains(err.Error(), "not mounted") {
+			// The device is already unmounted. We can safely ignore the error and treat it as a success!
+			logrus.WithError(err).Debugf("Device %s is already unmounted.\n", path)
+			return nil
+		}
+		return errors.Wrapf(err, "failed to umount %s: %s", path, string(output))
```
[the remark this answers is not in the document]
mantissahz: modified. Thanks

controller/volume_controller.go
```diff
@@ -2133,6 +2134,14 @@ func (c *VolumeController) reconcileAttachDetachStateMachine(v *longhorn.Volume,
 				if err := c.openVolumeDependentResources(v, e, rs, efs, log); err != nil {
 					return err
 				}
+				// For unexpected engine crash, the LUKS device might still be there,
+				// we need to lazy unmount it before we can safely log out the ISCSI target and re-attach the volume successfully.
```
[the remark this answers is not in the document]
mantissahz: modified. thanks.

util/util.go
```diff
@@ -945,3 +945,31 @@ func IsCryptsetupVerWithFixed16MiBHeaderSize() (bool, error) {
...
+
+	_, err = nsexec.Execute(nil, "umount", []string{"-l", path}, lhtypes.ExecuteDefaultTimeout)
+	if err != nil {
+		if strings.Contains(err.Error(), "not mounted") || strings.Contains(err.Error(), "no mount point specified") || strings.Contains(err.Error(), "no such file or directory") {
+			// The device is already unmounted. We can safely ignore the error and treat it as a success!
+			logrus.WithError(err).Debugf("Device %s is already unmounted.", path)
+			return nil
+		}
+		return errors.Wrapf(err, "failed to umount %s", path)
+	}
```
derekbit, line 971:
```suggestion
	if _, err := nsexec.Execute(nil, "umount", []string{"-l", path}, lhtypes.ExecuteDefaultTimeout); err != nil {
		if isUnmountedError(err) {
			logrus.WithError(err).Debugf("Device %s is already unmounted", path)
			return nil
		}
		return errors.Wrapf(err, "failed to lazy unmount %s", path)
	}
```


Add a helper function
```
func isUnmountedError(err error) bool {
	if err == nil {
		return false
	}

	errMsg := strings.ToLower(err.Error())

	return strings.Contains(errMsg, "not mounted") ||
		strings.Contains(errMsg, "no mount point specified") ||
		strings.Contains(errMsg, "no such file or directory")
}
```

util/util.go
```diff
@@ -945,3 +945,31 @@ func IsCryptsetupVerWithFixed16MiBHeaderSize() (bool, error) {
...
+
+	logrus.Debugf("Lazy umount device %s without the error.", path)
```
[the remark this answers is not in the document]
derekbit: Agree with the comment cc @mantissahz

util/util.go
```diff
@@ -945,3 +945,31 @@ func IsCryptsetupVerWithFixed16MiBHeaderSize() (bool, error) {
...
 	return nsexec.IsLuksFixed16MiBHeaderSize()
 }
+
+func LazyUnmount(path string) error {
+	if _, err := lhns.Stat(path); err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			return nil
+		}
+		return err
+	}
```
derekbit, line 955: If the stat is somehow failed, can we log the error and continue the lazy unmount?

controller/volume_controller.go
```diff
@@ -1899,6 +1900,15 @@ func (c *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[strin
...
 				}
+				// After an unexpected engine crash, a stale LUKS device mount may still be present.
+				// Lazy unmount it here to clear stale mount state before re-attaching the volume.
+				if v.Spec.Encrypted {
+					log.Infof("Volume is encrypted, lazy unmounting crypto device for volume %v if it is still mounted", v.Name)
+					cryptoDevice := crypto.VolumeMapper(v.Name, string(v.Spec.DataEngine))
+					if err := util.LazyUnmount(cryptoDevice); err != nil {
+						return errors.Wrapf(err, "failed to lazy unmount crypto device %v for volume %v", cryptoDevice, v.Name)
+					}
+				}
```
derekbit, line 1916: Another question: why do we put it in the volume controller rather than in the CSI path?

mantissahz: 1. It can not do lazy unmount the LUKS device at the CSI plugin level before sending an attaching volume request to the Longhorn server in ControllerPublishVolume because the CSI request might be received by a CSI plugin pod on the node (maybe node A) where the LUKS device is not present (maybe node B).
2. When the engine crashes, the volume controller will detect the error first, and then we can handle it as soon as possible.

derekbit: Got it! But is it possible to put it in the csi/node_server.go and execute the lazy unmount right before mount it?

mantissahz: No, because it will fail at attaching the volume, and it might not send the node request yet to publish the volume at the node (`NodePublishVolume`)

derekbit: I see. Could you add a comment to explain the purpose of the logic and why we don't put it in the CSI plugin? Thanks.

util/util.go
mantissahz: Added. Thanks.

controller/volume_controller.go
mantissahz: Added the explanation in the volume controller. Thanks.

util/util.go
mantissahz: Added a log and continued lazy unmount.

controller/volume_controller.go
```diff
@@ -1899,6 +1900,20 @@ func (c *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[strin
...
+				// by a CSI plugin pod on the node (maybe node A) where the LUKS device is not present (maybe node B).
+				// After an unexpected engine crash, a stale LUKS device mount may still be present.
+				// Lazy unmount it here to clear stale mount state before re-attaching the volume.
+				if v.Spec.Encrypted {
+					log.Infof("Volume is encrypted, lazy unmounting crypto device for volume %v if it is still mounted", v.Name)
+					cryptoDevice := crypto.VolumeMapper(v.Name, string(v.Spec.DataEngine))
+					if err := util.LazyUnmount(cryptoDevice); err != nil {
+						return errors.Wrapf(err, "failed to lazy unmount crypto device %v for volume %v", cryptoDevice, v.Name)
+					}
+				}
```
COLDTURNIP, line 1916: Does it guarantee that the volume state is always reconciled on the mounted node? Maybe we should handle device mount/unmout only at CSI node server or instance manager instead?

mantissahz: https://github.com/longhorn/longhorn-manager/blob/2850ba05cad1039d876537d7c42a3cdffd74377d/controller/volume_controller.go#L5937
`Volume.Spec.NodeID` is the node to which the volume is attaching, so the volume controller should be on the node where the LUKS device is.

>Maybe we should handle device mount/unmout only at CSI node server

We cannot handle it at the mount stage because it failed at the attach stage before the mount action. And we can not guarantee that the CSI-plugin pod that handles the attach request is on the node where the LUKS device is.

> instance manager

Yes, we can handle it in the engine when attaching the volume as my first PR in 
https://github.com/longhorn/go-iscsi-helper/pull/169,
and then we need the same modification in `go-spdk-helper`.

util/util.go
```diff
@@ -945,3 +945,42 @@ func IsCryptsetupVerWithFixed16MiBHeaderSize() (bool, error) {
...
+
+	_, err = nsexec.Execute(nil, "umount", []string{"-l", path}, lhtypes.ExecuteDefaultTimeout)
+	if err != nil {
+		if isUnmountedError(err) {
+			// The device is already unmounted. We can safely ignore the error and treat it as a success!
+			logrus.WithError(err).Debugf("Device %s is already unmounted.", path)
+			return nil
+		}
+		return errors.Wrapf(err, "failed to umount %s", path)
```
derekbit: Update the commit and PR titles as well
```suggestion
		return errors.Wrapf(err, "failed to unmount %s", path)
```

util/util.go
```diff
@@ -945,3 +945,42 @@ func IsCryptsetupVerWithFixed16MiBHeaderSize() (bool, error) {
...
+	}
+
+	logrus.Debugf("Lazy umount device %s without the error.", path)
```
derekbit:
```suggestion
	logrus.Debugf("Lazy unmounted device %s without the error.", path)
```

derekbit (dismissed):
In general, LGTM. Only few comments.

util/util.go
mantissahz: Modified. Thanks.

util/util.go
mantissahz: Modified. Thanks.

derekbit:
@mergify backport v1.12.x

PR #4732, opened 2026-05-12
feat: refactor v2 linked clone feature

shuo-wu:

Issue https://github.com/longhorn/longhorn/issues/12552

Summarizations from Copilot:

### Phase 0 — Prerequisites

- Source volume must be **Attached** (`sourceVol.Status.State == Attached`). `checkAndInitVolumeClone` silently returns and retries on the next reconcile cycle if not.

---

### Phase 1 — Clone initialization (`checkAndInitVolumeClone`)

1. Clone volume CR is created with `Spec.DataSource = vol/<srcVol>[/snapshot/<name>]` and `Spec.CloneMode = LinkedClone`.

2. Volume controller detects `shouldInitVolumeClone` (`CloneStatus.State == Empty` or retriable `Failed`):
   - If no snapshot is in DataSource: creates a **deterministic snapshot** on the source volume (UUID hashed from src+dst UIDs — idempotent across retries).
   - Sets `v.Status.CloneStatus = {State: Initiated, Snapshot: <snapshotName>, SourceVolume: <srcVol>}`.
   - Sets `v.Labels[linked-clone-source-snapshot] = snapshotName` (protects snapshot from deletion).

---

### Phase 2 — Replica and engine creation (`reconcileVolumeCreation` → `replenishReplicas`)

3. Volume controller creates **engine CR**.

4. `getReplenishReplicasCount`: for N-replica linked-clone (`proxyAPIVersion >= MinProxyAPIVersionForNReplicaLinkedClone`), the first gate (`isCloneTargetNotCompletedAndNotCopyCompleted && !linkedCloneAllowsN`) is **bypassed** → **all N replica CRs created at once**.

---

### Phase 3 — Scheduling (replica scheduler)

5. For each replica CR, `buildLinkedCloneSrcNodeDiskMap`:
   - If `r.Spec.LinkedCloneSrcReplicaName` already set → **must** land on that src replica's exact node+disk.
   - If not yet set → allowed disks = union of all **healthy** src replica node+disk pairs.
   - Normal fit criteria (tags, space, etc.) applied within allowed set.
   - Replica scheduled: `r.Spec.NodeID` and `r.Spec.DiskID` set.

---

### Phase 4 — Src replica assignment (`syncLinkedCloneReplicaSourceFields`)

6. Once each replica is scheduled (NodeID/DiskID ≠ ""), volume controller assigns `LinkedCloneSrcReplicaName`:
   - Finds healthy src replicas on the **same node+disk** as the dst replica.
   - Picks the one with the **fewest existing clone children** (load balancing).
   - Sets `r.Spec.LinkedCloneSrcReplicaName` + label — **immutable once set**.
   - If no src replica is on that disk at all (src was deleted during scheduling window): marks dst replica failed so it gets rescheduled.

---

### Phase 5 — Clone trigger (`updateRequestedDataSourceForVolumeCloning`)

7. Volume controller sets `e.Spec.RequestedDataSource = {sourceVol, snapshotName}`, signalling the engine controller to start cloning.

---

### Phase 6 — Engine controller fires SnapshotClone (`cloneSnapshot`)

8. Engine controller detects `RequestedDataSource != ""` and empty `CloneStatus` → builds `dstReplicaSrcReplicaPairMap = {dstReplicaName → srcReplicaName}` from all replicas' `LinkedCloneSrcReplicaName`.

9. Calls `engineClientProxy.SnapshotClone(engine, snapshot, srcEngineURL, ..., pairMap)` → IM proxy gRPC → **`Engine.SnapshotClone` on spdk-engine**.

---

### Phase 7 — spdk-engine executes `snapshotCloneLinkedN`

10. Engine connects to source engine (via `srcEngineAddress`) to fetch the list of RW src replicas and their addresses — **this is why source must be attached**.

11. For each `(dstReplica, srcReplica)` pair in the map, spawns a goroutine calling `ReplicaSnapshotCloneDstStart` on the dst replica's SPDK server (direct gRPC, **not** via IM):

    **Inside `Replica.SnapshotCloneDstStart` (linked-clone, synchronous):**
    - Calls `srcReplica.ReplicaSnapshotCloneSrcStart(...)` — registers the in-flight clone on the src side.
    - Sets `r.isCloneReplica = true`, `r.cloneEntrypointLvolName`.
    - Calls `SnapshotCloneDstFinish(spdkClient, LinkedClone)`:
      - Creates the entrypoint lvol (`<srcReplicaName>-clone-ep-<dstReplicaName>-...`).
      - Calls **`BdevLvolSetParent(entrypointLvol, srcSnapshotLvol)`** — fast SPDK metadata op, no data copy.
      - Rebuilds the replica's lvol chain with the entrypoint as root.
    - Calls `srcReplica.ReplicaSnapshotCloneSrcFinish(...)` — clears src in-memory cache.
    - Returns — **entire linked-clone is complete synchronously within this call**.

12. `wg.Wait()` — engine blocks until **all N goroutines return**. All N replicas finish before `Engine.SnapshotClone` returns to the IM proxy.

> **Contrast with deep-copy:** `SnapshotCloneDstStart` returns `InProgress` and spawns `monitorSnapshotClone` goroutine; `Engine.SnapshotClone` returns immediately; engine controller must poll `SnapshotCloneStatus` for minutes until data transfer completes.

---

### Phase 8 — Status propagation (polling)

13. Engine controller separately polls `engineClientProxy.SnapshotCloneStatus(engine)` → IM proxy → `ReplicaSnapshotCloneDstStatusCheck` on each replica → `e.Status.CloneStatus` updated. For linked-clone, the first poll already shows all replicas `Complete`.

14. Volume controller aggregates `e.Status.CloneStatus` across all replicas:
    - Any errored → `CloneStatus.State = Failed`
    - All complete → `CloneStatus.State = CopyCompletedAwaitingHealthy`

---

### Phase 9 — Volume becomes healthy and clone completes

15. Replicas start up, pass validation in `syncWithBdevLvolMap`/`syncCloneReplicaInfo` (entrypoint chain verified), become RW.

16. `v.Status.Robustness == Healthy` → volume controller sets `CloneStatus.State = Completed`.

17. Source snapshot label removed from deletion protection; `e.Spec.RequestedDataSource` cleared.

---

## Part 2 — Replica Rebuilding Flow

There are **two sub-cases** within each case depending on whether `CloneStatus.State == Completed` (clone finished) or still `Initiated/CopyCompletedAwaitingHealthy` (clone in-flight).

---

### Case 1 — Reuse existing failed replica

**Trigger:** A replica has `FailedAt != ""` and `HealthyAt != ""` (was once healthy; eligible for reuse).

1. Volume controller `replenishReplicas` → `CheckAndReuseFailedReplica`: clears `r.Spec.FailedAt` and `r.Spec.HealthyAt` on the existing replica CR (does **not** create a new CR).
   - `r.Spec.LinkedCloneSrcReplicaName` is **kept** (immutable once set).
   - `r.Spec.NodeID` / `r.Spec.DiskID` are **kept** — no rescheduling.

2. Instance manager detects the replica CR is runnable again → starts the replica SPDK process.

3. **Sub-case A: Clone still in-flight (`State == Initiated`)**
   - `syncLinkedCloneReplicaSourceFields`: no-op (src name already set).
   - `updateRequestedDataSourceForVolumeCloning`: `RequestedDataSource` is still set.
   - Engine controller's `shouldClone` check: only triggers if **all** `CloneStatus` entries have empty State.
     - If no other replica has cloned yet → `SnapshotClone` called again with updated pairMap including this replica → linked-clone executed (BdevLvolSetParent).
     - If other replicas already cloned (non-empty CloneStatus) → `shouldClone = false` → this replica goes through **normal SPDK rebuild** (full data copy, no thin clone).

4. **Sub-case B: Clone completed (`State == Completed`)**
   - `RequestedDataSource = ""` (cleared by `updateRequestedDataSourceForVolumeCloning`).
   - Engine adds replica back → normal SPDK rebuild (full data copy).
   - `RebuildingDstFinish` → `construct(bdevLvolMap)` → `recoverCloneReplicaInfo`:
     - After full rebuild, `ActiveChain[0]` is a normal snapshot (not a clone entrypoint).
     - → `isCloneReplica = false` ✅ — replica is now a regular full-copy replica.

---

### Case 2 — Replenish a brand-new replica

**Trigger:** `len(rs) < NumberOfReplicas` and no eligible failed replica to reuse.

1. Volume controller creates a **new replica CR** (no `LinkedCloneSrcReplicaName`, no NodeID/DiskID yet).

2. **Scheduler** (`buildLinkedCloneSrcNodeDiskMap`):
   - Co-location constraint applies **regardless of `CloneStatus.State`** — the new replica must land on the same node+disk as a healthy src replica.
   - If no src-co-located disk is schedulable: replica scheduling fails, volume stays degraded.

3. `syncLinkedCloneReplicaSourceFields`: once scheduled (NodeID/DiskID set), assigns `LinkedCloneSrcReplicaName` (same load-balancing logic as initial creation).

4. **Sub-case A: Clone still in-flight (`State == Initiated`)**
   - `shouldClone` in engine controller: if any existing replica already has a non-empty CloneStatus State → `false` → **normal SPDK rebuild**, not linked-clone.
   - If all CloneStatus entries are empty (very early, no replica has cloned yet) → `SnapshotClone` fires including this new replica → linked-clone path.

5. **Sub-case B: Clone completed (`State == Completed`)**
   - `RequestedDataSource = ""` → engine never calls `SnapshotClone`.
   - **Normal SPDK rebuild** from existing healthy replicas.
   - `RebuildingDstFinish` → `construct` → `recoverCloneReplicaInfo`:
     - `ActiveChain[0]` is a normal snapshot → `isCloneReplica = false` ✅

---

### Key takeaway

**Replacement replicas are almost always rebuilt as full copies, not thin clones.** The thin-clone optimization (BdevLvolSetParent) is only guaranteed for the initial N replicas that participated in the first `SnapshotClone` call. A replacement replica rebuilt via normal SPDK rebuild has its data but loses the "thin" property — `isCloneReplica` is correctly cleared by `recoverCloneReplicaInfo` in `RebuildingDstFinish`. This is intentional and correct.


webhook/resources/snapshot/validator.go
```diff
@@ -53,12 +57,6 @@ func (o *snapshotValidator) Create(request *admission.Request, newObj runtime.Ob
 		return werror.NewInvalidError("spec.volume is required", "spec.volume")
 	}
 
-	if isLinkedClone, err := o.ds.IsVolumeLinkedCloneVolume(snapshot.Spec.Volume); err != nil {
-		return werror.NewInvalidError(fmt.Sprintf("failed to check IsVolumeLinkedCloneVolume: %v", err), "")
-	} else if isLinkedClone {
-		return werror.NewInvalidError(fmt.Sprintf("snapshot is not allowed for linked-clone volume %v", snapshot.Spec.Volume), "")
-	}
-
```
davidcheng0922, line 61: There is a inconsistent behavior, we removes the create-time snapshot/backup guard for linked-clone volumes, but the rest of the code still treats linked-clone as not supporting snapshot/backup operations

- https://github.com/shuo-wu/longhorn-manager/blob/e90a874bacbc390801d06f23f15fd63acbda4a48/controller/volume_controller.go#L5115
- https://github.com/shuo-wu/longhorn-manager/blob/e90a874bacbc390801d06f23f15fd63acbda4a48/webhook/resources/volume/validator.go#L787

controller/volume_controller.go
```diff
@@ -990,31 +990,51 @@ func (c *VolumeController) ReconcileEngineReplicaState(v *longhorn.Volume, es ma
...
+				v.Status.CloneStatus.State = longhorn.VolumeCloneStateFailed
+				c.eventRecorder.Eventf(
+					v, corev1.EventTypeWarning, constant.EventReasonVolumeCloneFailed,
+					"failed to clone snapshot %v with clone mode %v from source volume %v: %v",
+					v.Status.CloneStatus.Snapshot, v.Spec.CloneMode, v.Status.CloneStatus.SourceVolume, firstError,
+				)
+			} else {
+				// For linked-clone: at least one replica completed the clone; failed replicas
+				// can be rebuilt from the successful ones via normal Longhorn replica rebuild.
+				v.Status.CloneStatus.State = longhorn.VolumeCloneStateCopyCompletedAwaitingHealthy
```
davidcheng0922, line 1039: Question: I might be missing something here, but is `CopyCompletedAwaitingHealthy` intended to remain as the clone state until the volume eventually becomes `Healthy`? 

In the partial-success case, it looks like the volume may stay `Degraded` while waiting for rebuilding. If rebuilding later fails and the volume becomes `Faulted/Failed`, should the clone state also move back to `Failed`,would we expect the clone state to transition to Failed as well, or is the current behavior intentional?

controller/volume_controller.go
```diff
@@ -4421,6 +4450,105 @@ func shouldInitVolumeClone(v *longhorn.Volume, log *logrus.Entry) bool {
...
+			continue
+		}
+
+		// Pick the src replica with the fewest existing clone replicas (load balancing)
+		best := candidates[0]
+		for _, candidate := range candidates[1:] {
+			if srcReplicaCloneCount[candidate.Name] < srcReplicaCloneCount[best.Name] {
+				best = candidate
+			}
+		}
```
davidcheng0922: Nit: when multiple source replicas have the same clone count, should we use a sorting to make the selection behavior more predictable and deterministic?

or do we prefer to keep the current simpler behavior for not using sorting.

shuo-wu: I think sorting and making it deterministic should be better. Thank you for the suggestion~

webhook/resources/snapshot/validator.go
shuo-wu: Fixed these with a separate commit. Thank you!

controller/volume_controller.go
shuo-wu: Yes. For a partial success case, the volume's clone state will remain `CopyCompletedAwaitingHealthy`.
If the rebuilding fails, the volume will be stuck in state `CopyCompletedAwaitingHealthy` and keep retrying the rebuilding.

This kind of unlimited retry and wait is not a good idea. It's better to forcibly update the state to `Completed` when the retry count reaches the max count (5 for now) or the rebuild wait time is over the stale time. I will launch this improvement with a new commit.

shuo-wu:
My manual test for the current implementation is good. Feel free to review it!

controller/node_controller.go
```diff
@@ -1948,16 +1948,8 @@ func (nc *NodeController) syncReplicaEvictionRequested(node *longhorn.Node, kube
...
-					log.WithError(err).Warnf("Cannot evict replica %v", replica.Name)
-				} else if isLinkedClone {
-					log.Warnf("Cannot evict replica %v because the volume %v is a linked-clone volume. "+
-						"Please delete this volume to unblock the eviction", replica.Name, replica.Spec.VolumeName)
-				} else {
-					replica.Spec.EvictionRequested = shouldEvictReplica
-					replicasToSync = append(replicasToSync, replicaToSync{replica, reason})
-				}
+				replica.Spec.EvictionRequested = shouldEvictReplica
+				replicasToSync = append(replicasToSync, replicaToSync{replica, reason})
```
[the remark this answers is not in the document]
shuo-wu: Legacy linked-clone volume crash is not considered a serious issue. No need to fix this

controller/volume_controller.go
```diff
@@ -4856,6 +4953,182 @@ func shouldInitVolumeClone(v *longhorn.Volume, log *logrus.Entry) bool {
 	return false
 }
 
+func (c *VolumeController) syncLinkedCloneReplicaSourceFields(v *longhorn.Volume, rs map[string]*longhorn.Replica) error {
```
derekbit, line 4956: The function runs on every volume reconcile for a linked-clone volume and does one job: bind each new replica of the clone volume to the specific source-volume replica it should be linked to.

1. Skip if not a linked-clone volume, if it's a legacy linked-clone volume (old architecture, doesn't use this field), or if the clone snapshot hasn't been resolved yet.
2. List the source volume's replicas to know what's available to link against.
3. Count existing bindings per source replica (for load balancing across multiple source replicas).
4. For each clone replica that is scheduled but not yet bound:
  - Find healthy source replicas on the same node+disk.
  - If none found but a source replica still exists there (just unhealthy), wait and retry next reconcile.
  - If no source replica exists there at all (it was deleted mid-scheduling), delete the stranded clone replica so replenishment can reschedule it somewhere that actually has a source.
  - If candidates exist, pick the least-loaded one (load balancing) and set Spec.LinkedCloneSrcReplicaName + a matching label, then persist via UpdateReplica.

Nit: The function is a bit complicated, we can refactor it in the future.

controller/volume_clone_controller.go
```diff
@@ -236,6 +236,36 @@ func (vcc *VolumeCloneController) reconcile(volName string) (err error) {
...
+	// case 3: this volume is source of a linked-clone target that needs rebuild
+	// (post-initial-clone: clone completed or awaiting healthy, but volume is degraded)
+	var readyNodes map[string]*longhorn.Node
+	for _, v := range vols {
+		if !isLinkedCloneNeedingSourceForRebuild(v, vol.Name) {
+			continue
+		}
+		attachmentTicketID := longhorn.GetAttachmentTicketID(longhorn.AttacherTypeVolumeCloneController, v.Name)
+		// Prefer the current attached node to avoid unnecessary migration
+		nodeID := vol.Status.CurrentNodeID
```
derekbit, line 248: Note: The "migration" risk here isn't about the data itself moving. It's that multiple attachment tickets on the same volume requesting different nodes could make the control plane think the source volume needs to move, which is unrelated to how the clone data actually gets copied.

controller/volume_clone_controller.go
```diff
@@ -236,6 +236,36 @@ func (vcc *VolumeCloneController) reconcile(volName string) (err error) {
...
+			}
+			for n := range readyNodes {
+				nodeID = n
+				break
+			}
+			if nodeID == "" {
+				continue // no ready nodes, skip this ticket
+			}
+		}
+		createOrUpdateAttachmentTicket(va, attachmentTicketID, nodeID, longhorn.AnyValue, longhorn.AttacherTypeVolumeCloneController)
```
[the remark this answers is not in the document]
derekbit: @shuo-wu It looks a valid concern. Would you improve it?

derekbit (approved):
In general, LGTM except for the comments from Coplit.

derekbit:
Merged the PR. @shuo-wu You can create a  new PR to address the comments.

derekbit:
@mergify backport v1.12.x

webhook/resources/volume/validator.go
```diff
@@ -296,6 +301,17 @@ func (v *volumeValidator) Update(request *admission.Request, oldObj runtime.Obje
 		}
 	}
 
+	// Legacy linked-clone volumes only support attach, detach, deletion, and backup.
+	// Block spec changes that would trigger incompatible operations.
+	if types.IsLegacyLinkedCloneVolume(oldVolume) {
```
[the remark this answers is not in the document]
shuo-wu: Legacy linked-clone volume is designed for adopting to the 3rd party backup solution like Velero. Backup creation involves snapshot creation. Hence, I will disable the recurring jobs only for the legacy volumes right now.

controller/volume_controller.go
shuo-wu: > If none found but a source replica still exists there (just unhealthy), wait and retry next reconcile.
> If no source replica exists there at all (it was deleted mid-scheduling), delete the stranded clone replica so replenishment can reschedule it somewhere that actually has a source.

So the main concern you are talking about is this part: the no-available-src-replica case, right? Compared with your proposal, blindly deleting the clone replica without any backoff window is indeed not graceful and even lead to tons of unnecessary replica deletion+recreation. But the proposal may lead to extra delay. I am thinking if we can further improve it a little bit:
- When the src replicas on the current selected disk are unavailable:
    - If there are no src replicas on other disks available, or picking any of the other healthy src replicas on the corresponding disk/node breaks the affinity for the clone volume ==> Wait for retry?
    - Otherwise, delete the clone replica so that we can quickly make the clone volume healthy

But this makes things more complicated. And I am not sure if this will conflict with any replica-auto-balance implementations. WDYT?

PR #4734, opened 2026-05-13
fix: get final backup status before disabling monitor

mantissahz:

Issue #

longhorn/longhorn#12842

Get the final backup status if the previous backup state is `Error` and the backup continues and succeeds.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved backup status synchronization to accurately reflect final backup states from the engine, particularly in error scenarios.
  * Enhanced error handling during backup reconciliation to ensure proper cleanup and status updates.

[Review Change Stack](https://app.coderabbit.ai/change-stack/longhorn/longhorn-manager/pull/4734)

### DIFF
--- a/controller/backup_controller.go
+++ b/controller/backup_controller.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"reflect"
 	"strconv"
+	"strings"
 	"sync"
 	"time"
 
@@ -485,11 +486,12 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
 	}
 
 	// The backup creation is complete, then the source of truth becomes the remote backup target
-	backupTargetClient, err := newBackupTargetClientFromDefaultEngineImage(bc.ds, backupTarget)
+	engineClientProxy, backupTargetClient, err := getBackupTarget(bc.controllerID, backupTarget, bc.ds, log, bc.proxyConnCounter, longhorn.DataEngineTypeAll)
 	if err != nil {
-		log.WithError(err).Error("Error init backup target clients")
+		log.WithError(err).Error("Failed to init backup target clients")
 		return nil // Ignore error to prevent enqueue
 	}
+	defer engineClientProxy.Close()
 
 	backupURL := backupstore.EncodeBackupURL(backup.Name, canonicalBackupVolumeName, backupTargetClient.URL)
 	backupInfo, err := backupTargetClient.BackupGet(backupURL, backupTargetClient.Credential)
@@ -518,9 +520,24 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
 		}
 	}
 
+	volume, err := bc.ds.GetVolume(canonicalBackupVolumeName)
+	if err != nil {
+		if !apierrors.IsNotFound(err) {
+			return err
+		}
+	}
+
+	if backup.Status.State == longhorn.BackupStateError && volume != nil {
+		if err := bc.syncFinalEngineBackupStatus(engineClientProxy, volume, backup); err != nil {
+			return err
+		}
+	}
+
 	// Update Backup CR status
+	if backup.Spec.SnapshotName == "" || volume == nil {
+		backup.Status.State = longhorn.BackupStateCompleted
+	}
 	backup.Status.BackupTargetName = backupTargetName
-	backup.Status.State = longhorn.BackupStateCompleted
 	backup.Status.URL = backupInfo.URL
 	backup.Status.SnapshotName = backupInfo.SnapshotName
 	backup.Status.SnapshotCreatedAt = backupInfo.SnapshotCreated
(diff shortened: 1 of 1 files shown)


controller/backup_controller.go
```diff
@@ -518,6 +520,19 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
...
+		if !apierrors.IsNotFound(err) {
+			return err
+		}
+	}
+
+	if backup.Status.State == longhorn.BackupStateError && volume != nil {
+		if err := bc.syncFinalEngineBackupStatus(engineClientProxy, volume, backup); err != nil {
+			return err
+		}
+	}
```
derekbit, line 534: A bit confused. If the `backup.Status.State` is `Error`, why the following updates change the state to `Completedd`? Shouldn't we keep the `Error` state?

derekbit: Okay. I understand the issue https://github.com/longhorn/longhorn/issues/12842. 
Another question, why doesn't v1 have the issue?

mantissahz: > Another question, why doesn't v1 have the issue?

V1 should have the same issue.
When the backup process is interrupted, the backup state is updated to be `Error`.
After the backup process continues at the engine/replica level, the backup volume controller gets information on the last backup updated and requests the backup reconciliation again.
In the backup controller, the backup is in the final state, and then it skips starting the monitor again and sets the backup to be `Completed` directly.

derekbit: Thanks @mantissahz.
@chriscchien Could you try to reproduce the issue using v1 volume and update to https://github.com/longhorn/longhorn/issues/12842? Thank you.

controller/backup_controller.go
```diff
@@ -965,6 +980,48 @@ func (bc *BackupController) syncWithMonitor(backup *longhorn.Backup, volume *lon
...
+		return fmt.Errorf("no RW replica found for engine %v to sync backup %v status", engine.Name, backup.Name)
+	}
+
+	backupStatus, err := engineClientProxy.SnapshotBackupStatus(engine, backup.Name, strings.TrimPrefix(replicaAddress, "tcp://"), "")
+	if err != nil {
+		return err
+	}
+
+	backup.Status.Progress = backupStatus.Progress
+	backup.Status.Error = backupStatus.Error
```
derekbit, line 1027: Don't we sync update the `State`?
If the `backupStatus.Error` is not empty or state is not completed, the following update will set the `Error` state to `Completed`. Will it have any potential issue?

controller/backup_controller.go
chriscchien: Please see https://github.com/longhorn/longhorn/issues/12842#issuecomment-4438535862 for reference

controller/backup_controller.go
```diff
@@ -518,6 +520,19 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
...
+	volume, err := bc.ds.GetVolume(canonicalBackupVolumeName)
+	if err != nil {
+		if !apierrors.IsNotFound(err) {
+			return err
+		}
+	}
+
+	if backup.Status.State == longhorn.BackupStateError && volume != nil {
+		if err := bc.syncFinalEngineBackupStatus(engineClientProxy, volume, backup); err != nil {
+			return err
```
davidcheng0922, line 532: Just to double-check: with this method, which one should be treated as the source of truth here — the engine-side backup status or the remote backup target metadata (backupInfo)?

controller/backup_controller.go
davidcheng0922: Same question, do we need to update `backup.Status.State` as well?

mantissahz: Yes, we can synchronize the `State`, and keep it without updating to `Completed` as follows.

Modified. Thanks

controller/backup_controller.go
mantissahz: Both are the source. 
Actually, almost fileds of backup info are passed from the engine side.
And the backup info does not keep the progress and state.
We have some fields to point out that the backup is completed.
Therefore, we can have a simpler way to treat progress 100 if fileds, such as `backup.CreatedTime`, `BackupVolume.LastBackupName` is `backup.Name`, are not empty.
WDYT, @derekbit, @davidcheng0922?

controller/backup_controller.go
```diff
@@ -965,6 +982,60 @@ func (bc *BackupController) syncWithMonitor(backup *longhorn.Backup, volume *lon
...
+	if err != nil {
+		return err
+	}
+	replicaAddress := ""
+	for addr, rs := range rsMap {
+		if rs.Mode == longhorn.ReplicaModeRW {
+			replicaAddress = addr
+			break
+		}
+	}
```
COLDTURNIP, line 1012: We're communicate with the attached volume replica here; however, the `engineClientProxy` here is fetched from `getBackupTarget`, which implies that the proxy client point the the default IM rather than the actual IM that the volume is using, then it may not able to list the replicas from the engine.

PR #4736, opened 2026-05-13
fix: remove encrypted dm device when engine crashed unexpectedly for v2 encrypted volumes

mantissahz:

Issue #

longhorn/longhorn#11510

https://ci.longhorn.io/job/private/job/longhorn-e2e-test/7811/
https://ci.longhorn.io/job/private/job/longhorn-e2e-test/7897/

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -1913,6 +1913,14 @@ func (c *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[strin
 					if err := util.LazyUnmount(cryptoDevice); err != nil {
 						return errors.Wrapf(err, "failed to lazy unmount crypto device %v for volume %v", cryptoDevice, v.Name)
 					}
+					if types.IsDataEngineV2(longhorn.DataEngineType(v.Spec.DataEngine)) {
+						// For v2 data engine, the dm-encrypted device will block the removing of the linear dm device that Longhorn creates for the volume on SPDK engine side.
+						// so we remove the encrypted dm device directly here to clear the stale state.
+						// And linear dm device will be removed on SPDK engine side when restarting the initiator.
+						if err := util.RemoveDMDevice(cryptoDevice); err != nil {
+							return errors.Wrapf(err, "failed to remove dm-crypt device %v for volume %v", cryptoDevice, v.Name)
+						}
+					}
 				}
 			}
 		}
--- a/util/util.go
+++ b/util/util.go
@@ -46,6 +46,7 @@ import (
 	lhio "github.com/longhorn/go-common-libs/io"
 	lhns "github.com/longhorn/go-common-libs/ns"
 	lhtypes "github.com/longhorn/go-common-libs/types"
+	lhspdkutil "github.com/longhorn/go-spdk-helper/pkg/util"
 
 	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
 )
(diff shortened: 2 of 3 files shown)


controller/volume_controller.go
```diff
@@ -1913,6 +1914,16 @@ func (c *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[strin
...
+					if types.IsDataEngineV2(longhorn.DataEngineType(v.Spec.DataEngine)) {
+						// To remove /dev/mapper/[volumeName] and /dev/mapper/[volumeName]-encrypted simultaneously.
+						dmDev := filepath.Join(crypto.MapperFilePathPrefix, v.Name)
+						if err := util.RemoveDMDevice(dmDev); err != nil {
+							return errors.Wrapf(err, "failed to remove DM device %v for volume %v", dmDev, v.Name)
+						}
+						if err := util.RemoveDMDevice(cryptoDevice); err != nil {
+							return errors.Wrapf(err, "failed to remove DM device %v for volume %v", cryptoDevice, v.Name)
+						}
+					}
```
derekbit, line 1923: Why does the deletion logic only work for v2 data engine?

mantissahz: Because the v2 volume will add a dm device originally at the start of the initiator, the LUKS device will be added later with a suffix name `-encrypted`.
And the original dm device might not be cleaned up because of the LUKS device.
Therefore, we need to clean them up for v2.
The v1 volume only has one dm device.

derekbit: Got it. Deleting the DM device here is a bit weird. Could we delete it in the longhorn-spdk-engine instead?

mantissahz:
Moved the removing dm logic to SPDK engine, https://github.com/longhorn/go-spdk-helper/pull/282

util/util.go
```diff
@@ -984,3 +985,31 @@ func isUnmountedError(err error) bool {
...
+	namespaces := []lhtypes.Namespace{lhtypes.NamespaceMnt, lhtypes.NamespaceIpc}
+	nsexec, err := lhns.NewNamespaceExecutor(lhtypes.ProcessNone, lhtypes.HostProcDirectory, namespaces)
+	if err != nil {
+		return err
+	}
+
+	if err := lhspdkutil.DmsetupRemove(devicePath, true, true, nsexec); err != nil {
+		if isRemoveDMError(err) {
+			return nil
+		}
```
derekbit: Add debug log message

controller/volume_controller.go
```diff
@@ -1913,6 +1913,13 @@ func (c *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[strin
 					if err := util.LazyUnmount(cryptoDevice); err != nil {
 						return errors.Wrapf(err, "failed to lazy unmount crypto device %v for volume %v", cryptoDevice, v.Name)
 					}
+					if types.IsDataEngineV2(longhorn.DataEngineType(v.Spec.DataEngine)) {
+						// For v2 data engine, we remove the encrypted dm device directly here to clear the stale state.
+						// And liner dm device will be removed on SPDK engine side when restarting the initiator.
+						if err := util.RemoveDMDevice(cryptoDevice); err != nil {
+							return errors.Wrapf(err, "failed to remove DM device %v for volume %v", cryptoDevice, v.Name)
```
derekbit:
```suggestion
							return errors.Wrapf(err, "failed to remove dm-crypt device %v for volume %v", cryptoDevice, v.Name)
```

controller/volume_controller.go
```diff
@@ -1913,6 +1913,13 @@ func (c *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[strin
 					if err := util.LazyUnmount(cryptoDevice); err != nil {
 						return errors.Wrapf(err, "failed to lazy unmount crypto device %v for volume %v", cryptoDevice, v.Name)
 					}
+					if types.IsDataEngineV2(longhorn.DataEngineType(v.Spec.DataEngine)) {
```
derekbit, line 1916: Don't we need to delete the v1's dm-crypt device?

util/util.go
mantissahz: Added. Thanks.

controller/volume_controller.go
mantissahz: Modified. Thanks.

controller/volume_controller.go
mantissahz: No, we don't need to delete v1 dm-crypted.
V2 volumes have another linear dm device that would affect dm-crypted device,
but v1 volumes don't have another one.

Lazy unmount is enough.

util/util.go
```diff
@@ -984,3 +985,32 @@ func isUnmountedError(err error) bool {
...
+			logrus.WithError(err).Debugf("Dm device %s is already removed.", devicePath)
+			return nil
+		}
+		return err
+	}
+
+	return nil
+}
+
+func isRemoveDMError(err error) bool {
```
derekbit: Rename to a better function name

controller/volume_controller.go
```diff
@@ -1913,6 +1913,13 @@ func (c *VolumeController) ReconcileVolumeState(v *longhorn.Volume, es map[strin
...
+						// For v2 data engine, we remove the encrypted dm device directly here to clear the stale state.
+						// And liner dm device will be removed on SPDK engine side when restarting the initiator.
+						if err := util.RemoveDMDevice(cryptoDevice); err != nil {
+							return errors.Wrapf(err, "failed to remove dm-crypt device %v for volume %v", cryptoDevice, v.Name)
+						}
+					}
```
derekbit, line 1923: Update the comment

util/util.go
mantissahz: Renamed to `isIgnorableDMRemoveError`.
Thanks.

controller/volume_controller.go
mantissahz: Added the description that the dm-crypted device will block the removal of the linear dm device.
Thanks.

derekbit (approved):

derekbit:
@mergify backport v1.12.x

PR #4738, opened 2026-05-13
fix: detect and delete stale engine frontend instances with empty IM frontend

derekbit:

Issue Longhorn/longhorn#13084

When an EngineFrontend is recovered from a persisted record left over from a DisableFrontend phase, the IM may report it as running with an empty frontend. This mismatch causes the volume to remain stuck because the controller treats the instance as healthy while no actual frontend is serving I/O.

Add a defense-in-depth check in syncEngineFrontend that compares the IM-reported frontend against the spec. If the spec requires a real frontend but the IM has an empty one, delete the stale instance so it can be re-created with the correct frontend type.

### DIFF
--- a/controller/engine_frontend_controller.go
+++ b/controller/engine_frontend_controller.go
@@ -53,6 +53,9 @@ type EngineFrontendController struct {
 
 	instanceHandler *InstanceHandler
 
+	// nil means use DeleteInstance; set in tests.
+	deleteInstanceHandler func(obj interface{}) error
+
 	proxyConnCounter util.Counter
 
 	backoff *flowcontrol.Backoff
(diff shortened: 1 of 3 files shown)


derekbit:
https://ci.longhorn.io/job/private/job/longhorn-e2e-test/7825/

- v2 regression
  - https://ci.longhorn.io/job/private/job/longhorn-tests-regression/10763/ 
  - https://ci.longhorn.io/job/private/job/longhorn-tests-regression/10764/

- v2 robot
  - https://ci.longhorn.io/job/private/job/longhorn-e2e-test/7826/
  - https://ci.longhorn.io/job/private/job/longhorn-e2e-test/7827/

controller/engine_frontend_controller.go
```diff
@@ -370,6 +373,32 @@ func (efc *EngineFrontendController) syncEngineFrontend(key string) (err error)
...
+		if imErr != nil {
+			log.WithError(imErr).Warn("Failed to get instance manager for stale engine frontend detection, skipping")
+		} else if shouldDeleteStaleRunningEngineFrontend(ef, im) {
+			log.Warnf("EngineFrontend %v is running in IM with empty frontend (spec expects %v), deleting stale instance to trigger re-creation", ef.Name, ef.Spec.Frontend)
+			efc.eventRecorder.Eventf(ef, corev1.EventTypeWarning, constant.EventReasonStaleInstance,
+				"Deleting stale instance: IM reports empty frontend but spec expects %v", ef.Spec.Frontend)
+			if err := efc.deleteEngineFrontendInstance(ef); err != nil {
+				return err
+			}
+			return nil
```
davidcheng0922: NIT:
If we don't log here, is it better for 
```go
return efc.deleteEngineFrontendInstance(ef)
``` 
to be cleaner?

derekbit: You are right! Use errors.wrap instead. Thank you.

derekbit:
@mergify backport v1.12.x

davidcheng0922 (dismissed):
LGTM

davidcheng0922 (approved):

PR #4742, opened 2026-05-14
feat(upgrade): allow n-2 upgrades for configured distros

derekbit:

Issue #

Add support for LONGHORN_TWO_MINOR_UPGRADE_DISTROS in longhorn-manager upgrade path validation.

When the current LONGHORN_DISTRO is listed in the
LONGHORN_TWO_MINOR_UPGRADE_DISTROS env var, allow upgrades that skip one minor version within the same major release. Keep the existing validation behavior unchanged for all other distros.

Also add unit coverage for both the allowed and rejected two-minor upgrade cases.


upgrade/util/util.go
```diff
@@ -283,6 +290,21 @@ func checkLHUpgradePath(namespace string, lhClient lhclientset.Interface) error
...
+		return "", false
+	}
+
+	for _, candidate := range strings.Split(os.Getenv(types.EnvTwoMinorUpgradeDistros), ",") {
+		if strings.ToLower(strings.TrimSpace(candidate)) == distro {
+			return distro, true
+		}
+	}
+
+	return "", false
```
davidcheng0922, line 305:
```go
return distro, false`
```
It miss distro information, even though it is non-empty

derekbit: Good catch! Thanks.

derekbit: Fixed. Thank you.

davidcheng0922 (approved):
LGTM

derekbit:
@mergify backport v1.12.x

PR #3673, opened 2025-04-09
fix: cleanup K8s VA associated with force-deleted pods

c3y1huang:
#### Which issue(s) this PR fixes:

Issue longhorn/longhorn#10689

#### What this PR does / why we need it:

Proactively clean up the Kubernetes VA of force-deleted pods. Without this, Kubernetes will eventually clean up the stale VA, but users may experience a significant delay before the pod becomes running again.

#### Special notes for your reviewer:

`None`

#### Additional documentation or context

https://github.com/longhorn/longhorn/issues/10689#issuecomment-2788357346

PR #4743, opened 2026-05-14
feat(global-manager): add Deployment hosting cluster-wide controllers

hookak:
Issue longhorn/longhorn#13059

### DIFF
--- a/app/daemon.go
+++ b/app/daemon.go
@@ -197,7 +197,9 @@ func startWebhooksByLeaderElection(ctx context.Context, kubeconfigPath, currentN
 		}
 
 		var err error
-		clients, err = client.NewClients(kubeconfigPath, true, ctx.Done())
+		// Neither the admission webhook (longhorn-system Pods only, AWS-IAM secret path) nor the
+		// node-local controllers read workload Pods, so this process stays off a cluster-wide Pod watch.
+		clients, err = client.NewClientsForNodeLocal(kubeconfigPath, ctx.Done())
 		if err != nil {
 			return err
 		}
@@ -340,12 +342,14 @@ func startManager(cmd *cli.Command) error {
 
 	snapshotConcurrentLimiter := controller.NewSnapshotConcurrentLimiter()
 
-	wsc, err := controller.StartControllers(logger, clients,
+	wsc, err := controller.StartNodeLocalControllers(logger, clients,
 		currentNodeID, serviceAccount, managerImage, backingImageManagerImage, shareManagerImage, instanceManagerImage,
 		kubeconfigPath, meta.Version, proxyConnCounter, snapshotConcurrentLimiter)
 	if err != nil {
 		return err
 	}
+
+	go monitorGlobalManagerLease(ctx, logger, clients.Datastore)
 
 	m := manager.NewVolumeManager(currentNodeID, clients.Datastore, proxyConnCounter, snapshotConcurrentLimiter)
 
@@ -449,4 +453,109 @@ func initDaemonNode(ds *datastore.DataStore, nodeName string) error {
 		return err
 	}
 	return nil
+}
+
+const (
+	globalManagerLeaseCheckInterval  = 30 * time.Second
+	globalManagerLeaseStaleThreshold = time.Minute
+	globalManagerLeaseAbsenceGrace   = 10 * time.Minute
+	globalManagerLeaseWarnInterval   = 5 * time.Minute
+)
+
+// monitorGlobalManagerLease warns when the longhorn-global-manager Lease is
+// missing, released without a successor, or no longer changing — the
+// controllers it hosts are then paused cluster-wide, which nothing else
+// surfaces in the manager logs. Staleness is judged by locally-observed
+// record changes rather than by comparing the leader-written RenewTime with
+// this node's clock (clock skew); a released Lease also carries a fresh
+// RenewTime, so an empty holder is tracked separately. The absence grace
+// keeps install and upgrade windows quiet, and the staleness threshold stays
+// well above the lease duration so normal failovers do not warn. Reads only
+// the informer cache.
+func monitorGlobalManagerLease(ctx context.Context, logger logrus.FieldLogger, ds *datastore.DataStore) {
+	log := logger.WithField("monitor", "globalManagerLease")
+	startTime := time.Now()
+
+	var lastWarnedAt time.Time
+	warn := func(reason string) {
+		if time.Since(lastWarnedAt) < globalManagerLeaseWarnInterval {
+			return
+		}
+		lastWarnedAt = time.Now()
+		log.Warnf("Lease %v %v; volume KubernetesStatus sync and remount-triggered pod deletion are paused cluster-wide while volume I/O keeps working; check the %v Deployment",
+			types.LonghornGlobalManagerName, reason, types.LonghornGlobalManagerName)
+	}
+
+	// Lease state as observed by this node, with local observation times.
+	var (
+		observed         bool
+		lastHolder       string
+		lastRenewTime    time.Time
+		lastChangeSeenAt time.Time
+		emptyHolderSince time.Time
+	)
+
+	check := func() string {
+		lease, err := ds.GetLeaseRO(types.LonghornGlobalManagerName)
+		if datastore.ErrorIsNotFound(err) {
+			observed = false
+			emptyHolderSince = time.Time{}
+			if time.Since(startTime) > globalManagerLeaseAbsenceGrace {
+				return "is not found"
+			}
+			return ""
+		}
+		if err != nil {
+			log.WithError(err).Debug("Failed to check the global-manager lease")
+			return ""
+		}
+		if lease.Spec.RenewTime == nil {
+			if time.Since(startTime) > globalManagerLeaseAbsenceGrace {
+				return "has never been acquired"
+			}
+			return ""
+		}
+
+		holder := ""
+		if lease.Spec.HolderIdentity != nil {
+			holder = *lease.Spec.HolderIdentity
+		}
+		if !observed || holder != lastHolder || !lease.Spec.RenewTime.Time.Equal(lastRenewTime) {
+			observed = true
+			lastHolder = holder
+			lastRenewTime = lease.Spec.RenewTime.Time
+			lastChangeSeenAt = time.Now()
+		}
+
+		if holder == "" {
+			if emptyHolderSince.IsZero() {
+				emptyHolderSince = time.Now()
+			}
+			if time.Since(emptyHolderSince) > globalManagerLeaseStaleThreshold {
+				return fmt.Sprintf("has been released and not re-acquired for at least %v",
+					time.Since(emptyHolderSince).Truncate(time.Second))
+			}
+			return ""
+		}
+		emptyHolderSince = time.Time{}
+
+		if time.Since(lastChangeSeenAt) > globalManagerLeaseStaleThreshold {
+			return fmt.Sprintf("has not been renewed for at least %v (holder %q)",
+				time.Since(lastChangeSeenAt).Truncate(time.Second), holder)
+		}
+		return ""
+	}
+
+	ticker := time.NewTicker(globalManagerLeaseCheckInterval)
+	defer ticker.Stop()
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-ticker.C:
+		}
+		if reason := check(); reason != "" {
+			warn(reason)
+		}
+	}
 }
--- a/app/global.go
+++ b/app/global.go
@@ -1,0 +1,182 @@
+package app
+
+import (
+	"context"
+	"net/http"
+	"sync"
+	"time"
+
+	"github.com/cockroachdb/errors"
+	"github.com/rancher/wrangler/v3/pkg/signals"
+	"github.com/sirupsen/logrus"
+	"github.com/urfave/cli/v3"
+
+	"k8s.io/client-go/tools/leaderelection"
+	"k8s.io/client-go/tools/leaderelection/resourcelock"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/longhorn/longhorn-manager/controller"
+	"github.com/longhorn/longhorn-manager/types"
+	"github.com/longhorn/longhorn-manager/util"
+	"github.com/longhorn/longhorn-manager/util/client"
+)
+
+const (
+	// HealthzPortGlobalManager serves /v1/healthz, which returns 200 OK
+	// whenever the process is alive; it is not tied to leader status so
+	// standby pods stay Ready.
+	HealthzPortGlobalManager = ":9505"
+)
+
+// GlobalCmd returns the `longhorn-manager global` subcommand, which hosts the
+// controllers that need cluster-wide Pod visibility (KubernetesPVController,
+// KubernetesPodController) under leader election.
+func GlobalCmd() *cli.Command {
+	return &cli.Command{
+		Name:  "global",
+		Usage: "Run the longhorn-global-manager workload (leader-elected).",
+		Flags: []cli.Flag{
+			&cli.StringFlag{
+				Name:  FlagKubeConfig,
+				Usage: "Specify path to kube config (optional)",
+			},
+		},
+		Action: func(ctx context.Context, cmd *cli.Command) error {
+			if err := startGlobalManager(cmd); err != nil {
+				logrus.Fatalf("Error starting global manager: %v", err)
+			}
+			return nil
+		},
+	}
+}
+
+func startGlobalManager(cmd *cli.Command) error {
+	kubeconfigPath := cmd.String(FlagKubeConfig)
+
+	podName, err := util.GetRequiredEnv(types.EnvPodName)
+	if err != nil {
+		return errors.Wrap(err, "failed to detect the pod name")
+	}
+	podNamespace, err := util.GetRequiredEnv(types.EnvPodNamespace)
+	if err != nil {
+		return errors.Wrap(err, "failed to detect the pod namespace")
+	}
+
+	ctx := signals.SetupSignalContext()
+
+	logger := logrus.StandardLogger().WithFields(logrus.Fields{
+		"component": "longhorn-global-manager",
+		"pod":       podName,
+	})
+
+	// A controller-init or healthz-listener failure must not exit in place:
+	// recording the error and cancelling the election context lets RunOrDie
+	// unwind and release the Lease (ReleaseOnCancel) before the process exits
+	// through the normal return path. The mutex makes the error safe to read
+	// after RunOrDie returns — client-go does not wait for the
+	// OnStartedLeading goroutine.
+	electionCtx, cancelElection := context.WithCancel(ctx)
+	defer cancelElection()
+
+	var (
+		failureLock sync.Mutex
+		failure     error
+	)
+	fail := func(err error) {
+		failureLock.Lock()
+		if failure == nil {
+			failure = err
+		}
+		failureLock.Unlock()
+		cancelElection()
+	}
+
+	startHealthzServer(ctx, logger, func(err error) {
+		fail(errors.Wrap(err, "healthz server failed"))
+	})
+
+	// Build clients/informers before leader election so every replica keeps a
+	// warm cache; a new leader starts against a synced cache (no fresh LIST).
+	clients, err := client.NewClients(kubeconfigPath, true, electionCtx.Done())
+	if err != nil {
+		return errors.Wrap(err, "failed to build longhorn clients")
+	}
+
+	lock := &resourcelock.LeaseLock{
+		LeaseMeta: metav1.ObjectMeta{
+			Name:      types.LonghornGlobalManagerName,
+			Namespace: podNamespace,
+		},
+		Client: clients.K8s.CoordinationV1(),
+		LockConfig: resourcelock.ResourceLockConfig{
+			Identity: podName,
+		},
+	}
+
+	leaderelection.RunOrDie(electionCtx, leaderelection.LeaderElectionConfig{
+		Lock:            lock,
+		ReleaseOnCancel: true,
+		LeaseDuration:   20 * time.Second,
+		RenewDeadline:   10 * time.Second,
+		RetryPeriod:     2 * time.Second,
+		Callbacks: leaderelection.LeaderCallbacks{
+			OnStartedLeading: func(leaderCtx context.Context) {
+				logger.Info("Acquired leader lease; starting global controllers")
+				if err := controller.StartGlobalControllers(logger, clients, leaderCtx.Done()); err != nil {
+					fail(errors.Wrap(err, "failed to start global controllers"))
+					return
+				}
+				logger.Info("Global controllers started; serving as leader")
+				<-leaderCtx.Done()
+			},
+			OnStoppedLeading: func() {
+				// Runs on both lease loss and election cancel; RunOrDie
+				// returns after this, so the process exits via the normal
+				// return path.
+				logger.Info("Leader election stopped; shutting down")
+			},
+			OnNewLeader: func(identity string) {
+				if identity == podName {
+					return
+				}
+				logger.Infof("Standing by; current leader is %q", identity)
+			},
+		},
+	})
+
+	failureLock.Lock()
+	defer failureLock.Unlock()
+	return failure
+}
+
+// startHealthzServer backs the Deployment's liveness probe. The server shuts
+// down when ctx is cancelled so SIGTERM stops the listener promptly, and a
+// listener failure is reported through onFailure so the caller can release
+// leadership before exiting.
+func startHealthzServer(ctx context.Context, logger logrus.FieldLogger, onFailure func(error)) {
+	mux := http.NewServeMux()
+	mux.HandleFunc("/v1/healthz", func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusOK)
+	})
+	server := &http.Server{
+		Addr:              HealthzPortGlobalManager,
+		Handler:           mux,
+		ReadHeaderTimeout: 5 * time.Second,
+	}
+	go func() {
+		logger.Infof("Healthz server listening on %s", HealthzPortGlobalManager)
+		if err := server.ListenAndServe(); err != nil && !errors.Is(err, http.ErrServerClosed) {
+			logger.WithError(err).Error("Healthz server failed")
+			onFailure(err)
+		}
+	}()
+	go func() {
+		<-ctx.Done()
+		shutdownCtx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+		defer cancel()
+		if err := server.Shutdown(shutdownCtx); err != nil {
+			logger.WithError(err).Warn("Healthz server shutdown error")
+		}
+	}()
+}
--- a/controller/controller_manager.go
+++ b/controller/controller_manager.go
@@ -28,8 +28,10 @@ var (
 	longhornFinalizerKey = longhorn.SchemeGroupVersion.Group
 )
 
-// StartControllers initiates all Longhorn component controllers and monitors to manage the creating, updating, and deletion of Longhorn resources
-func StartControllers(logger logrus.FieldLogger, clients *client.Clients,
+// StartNodeLocalControllers runs the DaemonSet's node-local controllers. The
+// cluster-wide-Pod-informer controllers (KubernetesPV/Pod) run only in the
+// longhorn-global-manager Deployment.
+func StartNodeLocalControllers(logger logrus.FieldLogger, clients *client.Clients,
 	controllerID, serviceAccount, managerImage, backingImageManagerImage, shareManagerImage, instanceManagerImage,
 	kubeconfigPath, version string, proxyConnCounter util.Counter, snapshotConcurrentLimiter *SnapshotConcurrentLimiter) (*WebsocketController, error) {
 	namespace := clients.Namespace
@@ -166,15 +168,7 @@ func StartControllers(logger logrus.FieldLogger, clients *client.Clients,
 	}
 
 	// Kubernetes controllers
-	kubernetesPVController, err := NewKubernetesPVController(logger, ds, scheme, kubeClient, controllerID)
-	if err != nil {
-		return nil, err
-	}
 	kubernetesNodeController, err := NewKubernetesNodeController(logger, ds, scheme, kubeClient, controllerID)
-	if err != nil {
-		return nil, err
-	}
-	kubernetesPodController, err := NewKubernetesPodController(logger, ds, scheme, kubeClient, controllerID)
 	if err != nil {
 		return nil, err
 	}
@@ -229,15 +223,39 @@ func StartControllers(logger logrus.FieldLogger, clients *client.Clients,
 	go shardGroupController.Run(Workers, stopCh)
 
 	// Start goroutines for Kubernetes controllers
-	go kubernetesPVController.Run(Workers, stopCh)
 	go kubernetesNodeController.Run(Workers, stopCh)
-	go kubernetesPodController.Run(Workers, stopCh)
 	go kubernetesConfigMapController.Run(Workers, stopCh)
 	go kubernetesSecretController.Run(Workers, stopCh)
 	go kubernetesPDBController.Run(Workers, stopCh)
 	go kubernetesEndpointController.Run(Workers, stopCh)
 
 	return websocketController, nil
+}
+
+// StartGlobalControllers runs the cluster-wide-Pod-informer controllers
+// (KubernetesPVController, KubernetesPodController) inside the leader-elected
+// longhorn-global-manager Deployment. The leader is the single writer, so the
+// controllers carry no per-node sharding guard. stopCh must be bound to the
+// leadership lifetime so the workers stop as soon as leadership is lost.
+func StartGlobalControllers(logger logrus.FieldLogger, clients *client.Clients,
+	stopCh <-chan struct{}) error {
+	kubeClient := clients.K8s
+	ds := clients.Datastore
+	scheme := clients.Scheme
+
+	kubernetesPVController, err := NewKubernetesPVController(logger, ds, scheme, kubeClient)
+	if err != nil {
+		return err
+	}
+	kubernetesPodController, err := NewKubernetesPodController(logger, ds, scheme, kubeClient)
+	if err != nil {
+		return err
+	}
+
+	go kubernetesPVController.Run(Workers, stopCh)
+	go kubernetesPodController.Run(Workers, stopCh)
+
+	return nil
 }
 
 func ParseResourceRequirement(val string) (*corev1.ResourceRequirements, error) {
--- a/controller/kubernetes_pod_controller.go
+++ b/controller/kubernetes_pod_controller.go
@@ -38,11 +38,9 @@ const (
 	remountRequestDelayDuration = 5 * time.Second
 )
 
+// KubernetesPodController is hosted by the longhorn-global-manager Deployment.
 type KubernetesPodController struct {
 	*baseController
-
-	// use as the OwnerID of the controller
-	controllerID string
 
 	kubeClient    clientset.Interface
 	eventRecorder record.EventRecorder
@@ -56,8 +54,7 @@ func NewKubernetesPodController(
 	logger logrus.FieldLogger,
 	ds *datastore.DataStore,
 	scheme *runtime.Scheme,
-	kubeClient clientset.Interface,
-	controllerID string) (*KubernetesPodController, error) {
+	kubeClient clientset.Interface) (*KubernetesPodController, error) {
 
 	eventBroadcaster := record.NewBroadcaster()
 	eventBroadcaster.StartLogging(logrus.Infof)
@@ -68,8 +65,6 @@ func NewKubernetesPodController(
 
 	kc := &KubernetesPodController{
 		baseController: newBaseController("longhorn-kubernetes-pod", logger),
-
-		controllerID: controllerID,
 
 		ds: ds,
 
@@ -366,10 +361,6 @@ func (kc *KubernetesPodController) isControllerInBlacklist(resource *metav1.Owne
 // cleanupForceDeletedPodResources removes stale resources left behind when a pod
 // is force-deleted (i.e., deletion grace period is zero).
 func (kc *KubernetesPodController) cleanupForceDeletedPodResources(pod *corev1.Pod) error {
-	if !isControllerResponsibleFor(kc.controllerID, kc.ds, pod.Name, "", pod.Spec.NodeName) {
-		return nil
-	}
-
 	if pod.DeletionTimestamp.IsZero() {
 		return nil
 	}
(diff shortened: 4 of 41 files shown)


controller/kubernetes_pod_controller.go
```diff
@@ -57,7 +66,8 @@ func NewKubernetesPodController(
 	ds *datastore.DataStore,
 	scheme *runtime.Scheme,
 	kubeClient clientset.Interface,
-	controllerID string) (*KubernetesPodController, error) {
+	controllerID string,
+	globalManagerEnabled bool) (*KubernetesPodController, error) {
```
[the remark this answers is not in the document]
hookak: Added the new parameter to the test's NewKubernetesPodController call.

app/daemon.go
```diff
@@ -305,7 +310,17 @@ func startManager(c *cli.Context) error {
...
+	// system-state frame (globalManagerEnabled) once here, so the
+	// rest of the code reads consistently.
+	globalManagerEnabled := c.Bool(FlagNodeLocalOnly)
+
+	var clients *client.Clients
+	if globalManagerEnabled {
+		clients, err = client.NewClientsForNodeLocal(kubeconfigPath, ctx.Done())
+	} else {
+		clients, err = client.NewClients(kubeconfigPath, true, ctx.Done())
+	}
```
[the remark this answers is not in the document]
hookak: - The DaemonSet-side variable is renamed to nodeLocalOnly, matching the flag's frame.
- StartControllers no longer takes a topology boolean. The consolidated-mode call is now explicit at the entry point: if !nodeLocalOnly { StartGlobalControllersInDaemonSet(...) } in app/daemon.go.
- StartGlobalControllers is split into StartGlobalControllersInDaemonSet and StartGlobalControllersInDeployment, so each call site declares its mode in the function name rather than via a boolean literal.

app/global.go
```diff
@@ -0,0 +1,190 @@
...
+				if identity == podName {
+					return
+				}
+				logger.Infof("Standing by; current leader is %q", identity)
+			},
+		},
+	})
+
+	return nil
+}
```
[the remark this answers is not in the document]
hookak: startGlobalManager now creates a cancellable child context, captures the runAsLeader error in a closure variable, and calls cancel() on failure so RunOrDie unwinds and releases the lease via ReleaseOnCancel.

app/global.go
```diff
@@ -0,0 +1,190 @@
...
+		Handler:           mux,
+		ReadHeaderTimeout: 5 * time.Second,
+	}
+	go func() {
+		logger.Infof("Healthz server listening on %s", HealthzPortGlobalManager)
+		if err := server.ListenAndServe(); err != nil && !errors.Is(err, http.ErrServerClosed) {
+			logger.WithError(err).Fatal("Healthz server failed")
+		}
+	}()
+}
```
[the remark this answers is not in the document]
hookak: startHealthzServer now accepts the signal context and launches a goroutine that calls server.Shutdown(shutdownCtx) with a 5 s timeout when the context is cancelled, so the listener stops accepting new connections promptly on SIGTERM.

hookak:
@derekbit I've reworked this to the single-mode always-on design with warm-standby leader election.
Could we get a fresh Copilot review on the updated code?

derekbit:
> @derekbit I've reworked this to the single-mode always-on design with warm-standby leader election. Could we get a fresh Copilot review on the updated code?

Hello @hookak 
Triggered. Sorry for late reply. Thank you.

app/global.go
```diff
@@ -0,0 +1,185 @@
...
+		LeaseDuration:   20 * time.Second,
+		RenewDeadline:   10 * time.Second,
+		RetryPeriod:     2 * time.Second,
+		Callbacks: leaderelection.LeaderCallbacks{
+			OnStartedLeading: func(leaderCtx context.Context) {
+				logger.Info("Acquired leader lease; starting global controllers")
+				if err := controller.StartGlobalControllers(logger, clients, GlobalControllerID); err != nil {
+					// Init failure exits 1; OnStoppedLeading exits 0.
+					logger.WithError(err).Error("failed to start global controllers")
+					os.Exit(1)
```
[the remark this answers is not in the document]
hookak: Fixed — the init error now cancels the election context and is returned after `RunOrDie` unwinds, so the Lease is released before exit. The `os.Exit(0)` in `OnStoppedLeading` is also removed — it runs on the same cancel path and would swallow the error.

controller/kubernetes_pod_controller.go
```diff
@@ -366,10 +368,8 @@ func (kc *KubernetesPodController) isControllerInBlacklist(resource *metav1.Owne
 // cleanupForceDeletedPodResources removes stale resources left behind when a pod
 // is force-deleted (i.e., deletion grace period is zero).
 func (kc *KubernetesPodController) cleanupForceDeletedPodResources(pod *corev1.Pod) error {
-	if !isControllerResponsibleFor(kc.controllerID, kc.ds, pod.Name, "", pod.Spec.NodeName) {
-		return nil
-	}
-
+	// The single leader handles pods on every node; force-delete below
+	// (grace 0) already implies the node is gone.
 	if pod.DeletionTimestamp.IsZero() {
```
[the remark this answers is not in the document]
hookak: Agreed — reworded to drop the node-gone assumption.

hookak:
@derekbit No worries at all. I know how busy you all are.

app/global.go
```diff
@@ -0,0 +1,195 @@
...
+			Identity: podName,
+		},
+	}
+
+	// Controller-init failure is propagated by cancelling the election
+	// context instead of exiting inside the callback, so RunOrDie unwinds
+	// and releases the Lease (ReleaseOnCancel) before the process exits.
+	var controllerInitErr error
+	electionCtx, cancelElection := context.WithCancel(ctx)
+	defer cancelElection()
```
davidcheng0922: Question here

The global controllers seem to be bound to `ctx` (clients.StopCh) instead of `electionCtx`.

Since closing `clients.StopCh` only signals the workers to stop, could `ReleaseOnCancel: true` release the `Lease` before the controllers have fully stopped, allowing another `global-manager replica` to reconcile the same resources?

For example:
T0: SIGTERM
T1: process ctx is cancelled
    ├─ clients.StopCh is closed
    └─ electionCtx is cancelled
T2: controller workers begin stopping
    └─ an in-flight reconcile may still be running
T3: leader election exits renew()
T4: ReleaseOnCancel releases the Lease
T5: another replica may acquire the Lease
T6: RunOrDie returns and the old process exits

Could this result in a short period where two controllers are reconciling the same resource?

hookak: Thanks for the thorough review — good catch on the shutdown ordering.

I think that window does exist, and my understanding is that it's hard to prevent fundamentally: client-go leader election doesn't provide fencing, so even with a drain-before-release sequence a brief overlap could still occur. So rather than trying to prevent the overlap, I leaned on the Kubernetes approach of keeping the writes safe under it — and I believe both controllers here are: KubernetesPVController only mirrors PV/PVC state into the volume status (a stale write fails on a resourceVersion conflict and requeues), and KubernetesPodController only deletes Pods/VolumeAttachments, which is idempotent.

That said, if you see a better way to handle the shutdown ordering, or think this needs more hardening, I'd be glad to hear it.

app/daemon.go
```diff
@@ -330,7 +334,7 @@ func startManager(cmd *cli.Command) error {
 
 	snapshotConcurrentLimiter := controller.NewSnapshotConcurrentLimiter()
 
-	wsc, err := controller.StartControllers(logger, clients,
+	wsc, err := controller.StartNodeLocalControllers(logger, clients,
```
c3y1huang, line 345: This no longer starts the PV/pod controller, so suggest adding a runtime guard. At minimum, consider a watchdog for the global-manager lease. WDYT?
```go
    // The cluster-wide Pod controllers run in the longhorn-global-manager
    // Deployment. If that Deployment is missing or its leader stops renewing
    // the Lease, remount-triggered pod deletion and volume KubernetesStatus
    // sync silently stop. Warn loudly so the gap is visible in the manager logs.
    go watchGlobalManagerLease(...)
```

app/global.go
```diff
@@ -0,0 +1,195 @@
...
+	})
+	server := &http.Server{
+		Addr:              HealthzPortGlobalManager,
+		Handler:           mux,
+		ReadHeaderTimeout: 5 * time.Second,
+	}
+	go func() {
+		logger.Infof("Healthz server listening on %s", HealthzPortGlobalManager)
+		if err := server.ListenAndServe(); err != nil && !errors.Is(err, http.ErrServerClosed) {
+			logger.WithError(err).Fatal("Healthz server failed")
```
c3y1huang: IIUC, this calls os.Exit and bypasses ReleaseOnCanel, so the lease is never released before the process exits. Could we avoid this path so that other candidates don't have to wait for the lease to expire?

app/global.go
```diff
@@ -0,0 +1,195 @@
...
+		},
+	}
+}
+
+func startGlobalManager(cmd *cli.Command) error {
+	kubeconfigPath := cmd.String(FlagKubeConfig)
+
+	podName, err := util.GetRequiredEnv(types.EnvPodName)
+	if err != nil {
+		return fmt.Errorf("failed to detect the pod name (env %v)", types.EnvPodName)
```
c3y1huang:
```suggestion
		return errors.Wrap(err, "failed to detect the pod name")
```

app/global.go
```diff
@@ -0,0 +1,195 @@
...
+	}
+	podNamespace, err := util.GetRequiredEnv(types.EnvPodNamespace)
+	if err != nil {
+		return fmt.Errorf("failed to detect the pod namespace (env %v)", types.EnvPodNamespace)
```
c3y1huang:
```suggestion
		return errors.Wrap(err, "failed to detect the pod namespace")
```

datastore/datastore.go
```diff
@@ -134,8 +134,42 @@ type DataStore struct {
 	extensionsClient apiextensionsclientset.Interface
 }
 
-// NewDataStore creates new DataStore object
+// NewDataStore creates a DataStore with the cluster-wide PodInformer
+// (equivalent to NewDataStoreForGlobal).
 func NewDataStore(namespace string, lhClient lhclientset.Interface, kubeClient clientset.Interface, extensionsClient apiextensionsclientset.Interface, informerFactories *util.InformerFactories) *DataStore {
+	return NewDataStoreForGlobal(namespace, lhClient, kubeClient, extensionsClient, informerFactories)
+}
```
c3y1huang: Slightly confused, what is the purpose of duplicating the function?

app/global.go
```diff
@@ -0,0 +1,195 @@
...
+	if err != nil {
+		return errors.Wrap(err, "failed to build longhorn clients")
+	}
+
+	config, err := buildLeaseConfig(kubeconfigPath)
+	if err != nil {
+		return errors.Wrap(err, "failed to build kube config for lease")
+	}
+	leaseKubeClient, err := clientset.NewForConfig(config)
+	if err != nil {
```
c3y1huang: What is the reason for building a seperate client for the lease instead of using `clients.K8s`? If intentional, suggest adding a doc comment explaining. Otherwise this reads like accidential duplication.

controller/controller_manager.go
```diff
@@ -157,19 +159,11 @@ func StartControllers(logger logrus.FieldLogger, clients *client.Clients,
 		return nil, err
 	}
 
-	// Kubernetes controllers
-	kubernetesPVController, err := NewKubernetesPVController(logger, ds, scheme, kubeClient, controllerID)
-	if err != nil {
-		return nil, err
-	}
+	// Kubernetes controllers (excluding PV and Pod — see StartGlobalControllers)
```
c3y1huang: Redundant and duplicated doc comment.

controller/controller_manager.go
```diff
@@ -28,8 +28,10 @@ var (
 	longhornFinalizerKey = longhorn.SchemeGroupVersion.Group
 )
 
-// StartControllers initiates all Longhorn component controllers and monitors to manage the creating, updating, and deletion of Longhorn resources
-func StartControllers(logger logrus.FieldLogger, clients *client.Clients,
+// StartNodeLocalControllers runs the DaemonSet's node-local controllers. The
+// cluster-wide-Pod-informer controllers (KubernetesPV/Pod) run only in the
+// longhorn-global-manager Deployment (see StartGlobalControllers).
```
c3y1huang: > (see StartGlobalControllers)

Redundant reference.

util/client/client.go
```diff
@@ -36,7 +36,19 @@ type Clients struct {
 	StopCh        <-chan struct{}
 }
 
+// NewClients constructs a Clients with the cluster-wide PodInformer
+// when needDataStore is true. Existing callers retain this behavior.
```
c3y1huang: > Existing callers retain this behavior.

This is a PR description, meaningless to someone reading this later in project codebase.

app/global.go
```diff
@@ -0,0 +1,195 @@
...
+				}
+				logger.Info("Global controllers started; serving as leader")
+				<-leaderCtx.Done()
+				logger.Info("Leader context cancelled")
+			},
+			OnStoppedLeading: func() {
+				// Runs on lease loss and on election-context cancel. No
+				// os.Exit here — RunOrDie returns right after this and the
+				// process exits through the normal return path, so a demoted
+				// leader stops writing and the Deployment recreates the pod.
```
c3y1huang: Suggest trimming this to just state the design intent:

```suggestion
				// Runs on both lease loss and election cancel; RunOrDie returns after
                // this, so the process exits via the normal return path.
```

app/global.go
```diff
@@ -0,0 +1,195 @@
...
+				logger.Info("Leader election stopped; shutting down")
+			},
```
c3y1huang: On shutdown, will there be two log lines? If so, suggest drop one.

controller/kubernetes_pod_controller.go
```diff
@@ -366,10 +368,8 @@ func (kc *KubernetesPodController) isControllerInBlacklist(resource *metav1.Owne
 // cleanupForceDeletedPodResources removes stale resources left behind when a pod
 // is force-deleted (i.e., deletion grace period is zero).
 func (kc *KubernetesPodController) cleanupForceDeletedPodResources(pod *corev1.Pod) error {
-	if !isControllerResponsibleFor(kc.controllerID, kc.ds, pod.Name, "", pod.Spec.NodeName) {
-		return nil
-	}
-
+	// The single elected leader handles pods on every node, so no per-node
+	// ownership check is needed here.
```
c3y1huang: nit: suggest removing this comment. It's a little weird explaining code that no longer exists, and is already documented in StartGlobalControllers.

app/global.go
```diff
@@ -0,0 +1,195 @@
...
+
+	"github.com/longhorn/longhorn-manager/controller"
+	"github.com/longhorn/longhorn-manager/types"
+	"github.com/longhorn/longhorn-manager/util"
+	"github.com/longhorn/longhorn-manager/util/client"
+)
+
+const (
+	// GlobalControllerID is the controller identity for the controllers hosted
+	// in the longhorn-global-manager Deployment.
```
c3y1huang: The uninstall controller should also clean up the longhorn-global-manager Deployment and its Lease.

app/daemon.go
hookak: Good idea — added. Each manager now watches the Lease from the existing informer cache and warns when it is missing, released without a successor, or no longer renewed — judged by locally-observed changes rather than remote timestamps, with a startup grace and rate-limited warnings.

app/global.go
hookak: Right, thanks — fixed. A healthz listener failure now cancels the election context (same path as controller-init failure), so RunOrDie releases the Lease and the process exits non-zero through the normal return path.

app/global.go
hookak: Done.

app/global.go
hookak: Done.

datastore/datastore.go
hookak: It kept the legacy name for the other entry points (uninstall, system-rollout), but agreed it reads ambiguous — removed the alias; those callers now use NewDataStoreForGlobal explicitly.

app/global.go
hookak: No strong reason — the upgrade leader election also uses the client it already has. Switched to clients.K8s.

controller/controller_manager.go
hookak: Done.

controller/controller_manager.go
hookak: Done.

util/client/client.go
hookak: Removed.

app/global.go
hookak: Done.

app/global.go
hookak: Yes — dropped the "Leader context cancelled" line; OnStoppedLeading covers both paths.

controller/kubernetes_pod_controller.go
hookak: Done — also removed the same-style comment in enqueuePodChange.

app/global.go
hookak: Added — the uninstaller now removes the longhorn-global-manager Deployment alongside the manager DaemonSet and deletes its Lease.

hookak:
@c3y1huang Thank you so much for the thorough review! Some of the comments caught pretty basic oversights on my side, so I also swept the rest of the change for the same patterns and cleaned those up along the way. Everything is addressed in follow-up commits (no force-push, so the threads stay anchored) — I'll squash back into a single commit once the review settles. Further feedback is always welcome!

datastore/datastore.go
```diff
@@ -138,8 +138,37 @@ type DataStore struct {
 	extensionsClient apiextensionsclientset.Interface
 }
 
-// NewDataStore creates new DataStore object
-func NewDataStore(namespace string, lhClient lhclientset.Interface, kubeClient clientset.Interface, extensionsClient apiextensionsclientset.Interface, informerFactories *util.InformerFactories) *DataStore {
+// NewDataStoreForGlobal creates a DataStore whose Pod informer watches
+// cluster-wide. It is used by the longhorn-global-manager process and by the
+// uninstall and system-rollout entry points.
+func NewDataStoreForGlobal(namespace string, lhClient lhclientset.Interface, kubeClient clientset.Interface, extensionsClient apiextensionsclientset.Interface, informerFactories *util.InformerFactories) *DataStore {
```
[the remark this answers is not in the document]
hookak: Fixed — the branch is updated to master and the three callers added there (longhorn_test.go, shard_group_controller_test.go) now use NewDataStoreForGlobal.

app/global.go
```diff
@@ -0,0 +1,182 @@
...
+	leaderelection.RunOrDie(electionCtx, leaderelection.LeaderElectionConfig{
+		Lock:            lock,
+		ReleaseOnCancel: true,
+		LeaseDuration:   20 * time.Second,
+		RenewDeadline:   10 * time.Second,
+		RetryPeriod:     2 * time.Second,
+		Callbacks: leaderelection.LeaderCallbacks{
+			OnStartedLeading: func(leaderCtx context.Context) {
+				logger.Info("Acquired leader lease; starting global controllers")
+				if err := controller.StartGlobalControllers(logger, clients, types.LonghornGlobalManagerName); err != nil {
```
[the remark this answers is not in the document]
hookak: Good catch — fixed in e880ad38e: the workers are now bound to `leaderCtx`, so they stop as soon as leadership is lost.

app/global.go
```diff
@@ -0,0 +1,182 @@
...
+		cancelElection()
+	}
+
+	startHealthzServer(ctx, logger, func(err error) {
+		fail(errors.Wrap(err, "healthz server failed"))
+	})
+
+	// Build clients/informers before leader election so every replica keeps a
+	// warm cache; a new leader starts against a synced cache (no fresh LIST).
+	clients, err := client.NewClients(kubeconfigPath, true, ctx.Done())
```
c3y1huang:
```suggestion
	clients, err := client.NewClients(kubeconfigPath, true, electionCtx.Done())
```

controller/kubernetes_pv_controller.go
```diff
@@ -195,10 +196,6 @@ func (kc *KubernetesPVController) syncKubernetesStatus(key string) (err error) {
 		return err
 	}
 
-	if volume.Status.OwnerID != kc.controllerID {
```
c3y1huang, line 198: Is the `controllerID` in the struct still needed?

datastore/datastore.go
```diff
@@ -138,8 +138,37 @@ type DataStore struct {
...
+	ds := newDataStoreCommon(namespace, lhClient, kubeClient, extensionsClient, informerFactories)
```
c3y1huang, line 147: Do we need to build the full datastore? It seems a bit wasteful.

app/global.go
hookak: Good catch — applied. Now a failure recorded via `fail()` also unwinds the cache-sync wait instead of leaving the process blocked.

controller/kubernetes_pv_controller.go
hookak: It isn't — its only readers were the ownership guards this PR removes. Dropped the field and the constructor parameters from both controllers.

datastore/datastore.go
hookak: Agreed — these two controllers transitively need only a small subset of the caches (Longhorn Volume/Setting/Node/VolumeAttachment + Pod/PV/PVC). I did consider a trimmed datastore for this process, but held off for two reasons:

- A trimmed constructor would still return `*DataStore`, so every datastore method keeps compiling against it, while any method that internally reaches an unwired lister panics at runtime — and only in this process. Which listers are safe can only be determined by tracing the internal lister usage of each method, and that list goes stale silently as datastore methods evolve. Making this compile-safe would mean giving the controllers purpose-built interfaces/stores, which is a datastore-wide restructuring beyond this PR.
- The avoidable overhead is bounded: the cluster-wide Pod cache is what this Deployment exists for (already reduced by the Pod transform), so the waste is mostly the namespace-scoped Longhorn CR caches that every DaemonSet pod already holds, duplicated into two replicas regardless of cluster size.

So I leaned toward keeping the datastore uniform rather than trading a constant overhead for a new runtime failure mode. If you think the trimmed variant is worth it, I'm happy to do it as a follow-up.

app/global.go
```diff
@@ -0,0 +1,182 @@
...
+	leaderelection.RunOrDie(electionCtx, leaderelection.LeaderElectionConfig{
+		Lock:            lock,
+		ReleaseOnCancel: true,
+		LeaseDuration:   20 * time.Second,
+		RenewDeadline:   10 * time.Second,
+		RetryPeriod:     2 * time.Second,
+		Callbacks: leaderelection.LeaderCallbacks{
+			OnStartedLeading: func(leaderCtx context.Context) {
+				logger.Info("Acquired leader lease; starting global controllers")
+				if err := controller.StartGlobalControllers(logger, clients, leaderCtx.Done()); err != nil {
```
c3y1huang, line 126: Since handlers are registered after promotion and informers don't replay deletes, could Longhorn miss a pod delete during a no-leader window? If so, could that potential leak VA and block reattach because the cleanup won't be retried?

hookak: Thanks — you're right that the controllers only see events after promotion and that deletes aren't replayed.

For this controller, though, a missed delete event doesn't leave a VA without an owner to clean it up:

- The normal detach path doesn't go through this controller: when a pod is deleted, the Kubernetes attach-detach controller decides the detach and deletes the VA, which propagates to Longhorn through the regular CSI unpublish flow.
- The cleanups this controller performs act on pods that still exist (stuck terminating), so a newly promoted leader re-discovers them from the informer's initial listing and the periodic resync. In `handlePodDeletionIfNodeDown`, the VAs are deleted before the pod force-delete is issued, so the pod's disappearance means the cleanup already happened.
- The one exception is `cleanupForceDeletedPodResources`, which does need to catch the short-lived grace-period-0 state, so a no-leader window could miss it. The impact looks bounded, though: as noted in https://github.com/longhorn/longhorn-manager/pull/3673, Kubernetes eventually cleans up the stale VA on its own, so a miss means slower recovery for that pod rather than a permanently blocked reattach.

If you think it's worth hardening further, a follow-up could make that cleanup derivable from current state (an attached Longhorn VA with no live consumer on its node) instead of requiring the deletion to be observed. Happy to file an issue for it.

hookak:
@c3y1huang The suppressed Copilot comment in https://github.com/longhorn/longhorn-manager/pull/4743#pullrequestreview-4853161791 points at a real gap on the PV side: `KubernetesPVController` learns which volume a deleted PV belonged to only from the delete event, so a PV deleted while no controller instance is running never gets its `KubernetesStatus` cleaned up. This is pre-existing behavior (the historical topology had the same loss on an owner-manager restart), which is why I had left it out of scope so far — but the leader-failover window makes it a recurring case, and this concern has come up a few times in this review.

I have a commit ready that removes the delete-event dependency: cleanup is re-derived from the durable record (`Status.KubernetesStatus.PVName`), so missed deletions converge from current state instead of being lost.

Would you prefer this included in this PR, or tracked as a separate follow-up issue? I'm happy to push the commit here if you consider it in scope — otherwise I'll file a follow-up and keep this PR as is.

c3y1huang:
> The suppressed Copilot comment in [(earlier review)](https://github.com/longhorn/longhorn-manager/pull/4743#pullrequestreview-4853161791) points at a real gap on the PV side: `KubernetesPVController` learns which volume a deleted PV belonged to only from the delete event, so a PV deleted while no controller instance is running never gets its `KubernetesStatus` cleaned up. This is pre-existing behavior (the historical topology had the same loss on an owner-manager restart), which is why I had left it out of scope so far — but the leader-failover window makes it a recurring case, and this concern has come up a few times in this review.
> 
> I have a commit ready that removes the delete-event dependency: cleanup is re-derived from the durable record (`Status.KubernetesStatus.PVName`), so missed deletions converge from current state instead of being lost.
> 
> Would you prefer this included in this PR, or tracked as a separate follow-up issue? I'm happy to push the commit here if you consider it in scope — otherwise I'll file a follow-up and keep this PR as is.

Thanks for confirming the gap and preparing the fix. Since this is pre-existing behavior, I would prefer it as a separate followup issue, so it gets its own review and test coverage. Please create a new issue and link it there.

c3y1huang (commented):
LGTM in general. 

For othere reviewers, three trade offs/gaps worth noting:
- https://github.com/longhorn/longhorn-manager/pull/4743#discussion_r3577293699
- https://github.com/longhorn/longhorn-manager/pull/4743#discussion_r3702251026
- https://github.com/longhorn/longhorn-manager/pull/4743#discussion_r3711091761

And one pre-existing gap to be tracked:
- https://github.com/longhorn/longhorn-manager/pull/4743#issuecomment-5180743813

I'll defer my approval until the commit history is cleaned up to avoid premature auto-merge.

cc @derekbit @longhorn/dev-control-plane

hookak:
Filed the follow-up issue for the PV-side gap: https://github.com/longhorn/longhorn/issues/13654

Squashed the commit history into a single commit (`106fe22a3` — tree identical to the previous head `6899750`, nothing to re-review content-wise).

On merge coordination: the DaemonSet no longer runs the two migrated controllers, so the chart PR https://github.com/longhorn/longhorn/pull/13100 should merge right after this one — until then, a master-head install has no instance of the migrated controllers.

derekbit:
@hookak 
Before this change, `handlePodDeletionIfNodeDown` ran in every DaemonSet pod with no node guard, so force-deleting a pod stuck on a downed node was handled redundantly by all surviving managers.

Now it runs only in the single global-manager leader. It's still correct (after failover the new leader re-enqueues and force-deletes the stuck pod), but when the leader itself is on the downed node, cluster-wide node-down deletion / remount is paused until a new leader takes over and how long that pause is depends entirely on the deployment topology. Will it be a concern?

By the way, could you help resovle the conflicts? Thank you.

hookak:
Thanks for reviewing, @derekbit.

We think the pause is not a concern in practice: when the leader is on the downed node, failover completes long before this handler has any work to do. A standby takes over the lease within ~22s at worst (LeaseDuration 20s + RetryPeriod 2s; a graceful exit releases the lease, ~2–3s measured). `handlePodDeletionIfNodeDown` itself only becomes actionable much later: it requires the Longhorn node Ready condition to be false (mirrored from Kubernetes NotReady, `node-monitor-grace-period` 40s by default) and the pod to already be terminating with its grace expired — on a downed node the DeletionTimestamp comes from taint eviction (300s toleration by default). Both clocks start at the same node failure, so the new leader is in place well before any pod becomes eligible for force-deletion. Standbys run warm (caches pre-synced) and the Add replay on takeover re-enqueues every pod, so nothing depends on events missed during the window.

More generally, any pause on this path is bounded by that ~22s regardless of deployment topology, as long as one standby survives elsewhere — the chart's 3 replicas with pod anti-affinity keep them spread. This is also the model the node-down path already relies on: kube-controller-manager drives the NotReady/eviction chain and the CSI external-attacher performs the detach after this handler deletes the VolumeAttachment, both single leaders with lease settings of the same magnitude. If a tighter bound is preferred, we're happy to lower the lease timings (e.g. 15s/10s/2s, matching kube-controller-manager's defaults).

Resolved the conflicts and rebased onto the latest master.

derekbit (dismissed):
LGTM. I don't have questions now. I can build a custom longhorn-manager image and run the negative testing first before merging it. WDYT?

hookak:
Totally agree — given the size of this change, running negative testing before merging makes a lot of sense. Please go ahead, thank you!

We will also run negative tests on our side in parallel with the latest rebased branch and share the results here.

hookak:
Totally agree — given the size of this change, running negative testing before merging makes a lot of sense. Please go ahead, thank you!

We will also run negative tests on our side in parallel with the latest rebased branch and share the results here.

derekbit:
Running
- master-head with the feature
	- https://ci.longhorn.io/job/private/job/longhorn-e2e-test/8931
	- https://ci.longhorn.io/job/private/job/longhorn-e2e-test/8932
- master-head
	- https://ci.longhorn.io/job/private/job/longhorn-e2e-test/8921/
	- https://ci.longhorn.io/job/private/job/longhorn-e2e-test/8922/

derekbit:
> Running
> 
> * master-head with the feature
>   
>   * [ci.longhorn.io/job/private/job/longhorn-e2e-test/8931](https://ci.longhorn.io/job/private/job/longhorn-e2e-test/8931)
>   * [ci.longhorn.io/job/private/job/longhorn-e2e-test/8932](https://ci.longhorn.io/job/private/job/longhorn-e2e-test/8932)

@hookak 
The negative testing is unable to complete. 
Triggered it again to see if it is reproducible.
https://ci.longhorn.io/job/private/job/longhorn-e2e-test/8941
https://ci.longhorn.io/job/private/job/longhorn-e2e-test/8942

hookak:
@derekbit Would it be possible for me to get access to CI system?

derekbit:
> @derekbit Would it be possible for me to get access to CI system?

Hello @hookak Could you try if you can log in https://ci.longhorn.io/ using your github account?

hookak:
It works! Thanks for the access!

hookak:
@derekbit It looks like the `longhorn-global-manager` Deployment was not deployed in these runs — the job installs `deploy/longhorn.yaml` from `master`, which does not carry it yet (it comes with https://github.com/longhorn/longhorn/pull/13100), so the migrated controllers had no host process.

Could you re-run with `LONGHORN_REPO_URI=https://github.com/hookak/longhorn.git` and `LONGHORN_REPO_BRANCH=feat/global-manager-chart`? That branch has the manifest change, and `CUSTOM_LONGHORN_MANAGER_IMAGE` is substituted into the added Deployment as well.

hookak:
@derekbit A gentle follow-up — I know you are busy, so no rush.

Since I now have CI access, would it be OK if I trigger the re-run myself? I would reuse the parameters of https://ci.longhorn.io/job/private/job/longhorn-e2e-test/8941/

derekbit:
> @derekbit A gentle follow-up — I know you are busy, so no rush.
> 
> Since I now have CI access, would it be OK if I trigger the re-run myself? I would reuse the parameters of [ci.longhorn.io/job/private/job/longhorn-e2e-test/8941](https://ci.longhorn.io/job/private/job/longhorn-e2e-test/8941/)

Ah @hookak 
I missed the comment. No worries. Feel free to re-run and trigger it. Thank you.

hookak:
@derekbit Thank you. I re-ran the negative suite, installing from a branch that carries the `longhorn-global-manage` deployment 

- https://ci.longhorn.io/job/private/job/longhorn-e2e-test/9058/
- https://ci.longhorn.io/job/private/job/longhorn-e2e-test/9059/

One question: 8941 and 8942 are still marked BUILDING, wold you live me to abort them, or leave them to you?

derekbit:
> @derekbit Thank you. I re-ran the negative suite, installing from a branch that carries the `longhorn-global-manage` deployment
> 
> * [ci.longhorn.io/job/private/job/longhorn-e2e-test/9058](https://ci.longhorn.io/job/private/job/longhorn-e2e-test/9058/)
> * [ci.longhorn.io/job/private/job/longhorn-e2e-test/9059](https://ci.longhorn.io/job/private/job/longhorn-e2e-test/9059/)
> 
> One question: 8941 and 8942 are still marked BUILDING, wold you live me to abort them, or leave them to you?

Hello @hookak Cancelled 8941 and 8942. Thank you.

BTW, I forgot to mention that you can add your public key of your PC to the `CUSTOM_SSH_PUBLIC_KEY` field while creating a test job.

image

Then, you can acees the machine by `ssh ec2-user@<IP>`. The IP can be found in the console log (search `LONGHORN_CLIENT_URL=`). The kubeconfig is in `/etc/rancher/k3s/`.

hookak:
Got it. Thanks for the tip!

hookak:
@derekbit 9058 and 9059 passed 134 cases across 12 suites with no failures, then stalled at `Pull Backup From Another Longhorn` because my fork had no `v1.12.0` tag, and were cancelled on 08-28. I pushed the tag and re-triggered the full suite with the same parameters:

- https://ci.longhorn.io/job/private/job/longhorn-e2e-test/9148/
- https://ci.longhorn.io/job/private/job/longhorn-e2e-test/9149/

cc @yangchiu — if these need to be cancelled for a release or other jobs, please let me know and I will re-run later.

hookak:
The negative test reruns have completed. Both runs finished with identical results: 176 cases — **156 passed, 19 failed, 1 skipped** ([9148](https://ci.longhorn.io/job/private/job/longhorn-e2e-test/9148/), [9149](https://ci.longhorn.io/job/private/job/longhorn-e2e-test/9149/), ~36.5h each). 

The 19 failures all look like environment/configuration issues, identical across both runs: node ssh unreachable (`Service Restart`, 15), AWS EBS snapshot timeouts (`VM Snapshot`, 2), `LONGHORN_TRANSIENT_VERSION` unset (1), and one backup timeout on stock v1.12.0 before the upgrade step. Please let me know if any of them needs a closer look.

I've also submitted longhorn/longhorn-tests#3690 adding e2e coverage for the new component.

derekbit:
> The negative test reruns have completed. Both runs finished with identical results: 176 cases — **156 passed, 19 failed, 1 skipped** ([9148](https://ci.longhorn.io/job/private/job/longhorn-e2e-test/9148/), [9149](https://ci.longhorn.io/job/private/job/longhorn-e2e-test/9149/), ~36.5h each).
> 
> The 19 failures all look like environment/configuration issues, identical across both runs: node ssh unreachable (`Service Restart`, 15), AWS EBS snapshot timeouts (`VM Snapshot`, 2), `LONGHORN_TRANSIENT_VERSION` unset (1), and one backup timeout on stock v1.12.0 before the upgrade step. Please let me know if any of them needs a closer look.
> 
> I've also submitted [longhorn/longhorn-tests#3690](https://github.com/longhorn/longhorn-tests/pull/3690) adding e2e coverage for the new component.

Awesome! Thanks @hookak. I will review it next Monday. The PR should be ready to go!

derekbit (dismissed):
LGTM

derekbit:
@mergify backport v1.13.x

derekbit:
Hello @hookak 
CI failed.  Could you help fix it?
```
262.6 # github.com/longhorn/longhorn-manager/controller/monitor
262.6 # [github.com/longhorn/longhorn-manager/controller/monitor]
262.6 vet: controller/monitor/disk_monitor_test.go:56:18: undefined: datastore.NewDataStore
```

derekbit:
Hello @hookak 
We found some failures while running test_ha.py testing. Could you take a look?
- https://ci.longhorn.io/job/private/job/longhorn-tests-regression/12082 
- https://ci.longhorn.io/job/private/job/longhorn-tests-regression/12081

hookak:
@derekbit

- vet failure: fixed and pushed (`e357c7dc1`).
- test_ha.py: the runs installed the `master` manifest, which has no `longhorn-global-manager` Deployment yet (https://github.com/longhorn/longhorn/pull/13100), so the migrated controllers were not running. I re-ran the same two jobs with only `LONGHORN_REPO_URI=https://github.com/hookak/longhorn.git` / `LONGHORN_REPO_BRANCH=feat/global-manager-chart` changed:
  - https://ci.longhorn.io/job/private/job/longhorn-tests-regression/12094/ (test_ha.py)
  - https://ci.longhorn.io/job/private/job/longhorn-tests-regression/12095/ (test_salvage_auto_crash_all_replicas)

derekbit:
> @derekbit
> 
> * vet failure: fixed and pushed (`e357c7dc1`).
> * test_ha.py: the runs installed the `master` manifest, which has no `longhorn-global-manager` Deployment yet ([feat(chart): add always-on longhorn-global-manager Deployment longhorn#13100](https://github.com/longhorn/longhorn/pull/13100)), so the migrated controllers were not running. I re-ran the same two jobs with only `LONGHORN_REPO_URI=https://github.com/hookak/longhorn.git` / `LONGHORN_REPO_BRANCH=feat/global-manager-chart` changed:
>   
>   * [ci.longhorn.io/job/private/job/longhorn-tests-regression/12094](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/12094/) (test_ha.py)
>   * [ci.longhorn.io/job/private/job/longhorn-tests-regression/12095](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/12095/) (test_salvage_auto_crash_all_replicas)

So fast. Thank you @hookak. I will review it!

derekbit (approved):

derekbit:
@mergify backport v1.13.x

hookak:
Just a quick heads-up: regression run [12094](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/12094/) is still in progress. The `test_salvage_auto_crash_all_replicas` case that 12095 was intended to rerun has already passed in 12094. Would you still like me to follow up once the full run completes? @derekbit

derekbit:
> Just a quick heads-up: regression run [12094](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/12094/) is still in progress. The `test_salvage_auto_crash_all_replicas` case that 12095 was intended to rerun has already passed in 12094. Would you still like me to follow up once the full run completes? @derekbit

Sure! Thanks @hookak

hookak:
@derekbit [12094](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/12094/) finished: 33 passed, 1 failed, 2 skipped.

The one failure, `test_engine_image_not_fully_deployed_perform_dr_restoring_expanding_volume[nfs]`, also fails on master-head without this PR ([12077](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/12077/), [12078](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/12078/), [12117](https://ci.longhorn.io/job/private/job/longhorn-tests-regression/12117/)), so it doesn't look related to this change.

PR #4769, opened 2026-05-19
fix(scheduler): prevent unschedulable replicas during disk pressure auto-balance

derekbit:

Issue longhorn/longhorn#13152

Validate volume.Spec.DiskSelector when checking schedulable candidates for disk pressure auto-balance.

This prevents Longhorn from selecting disks with mismatched tags as valid migration targets and creating unschedulable placeholder replicas.

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -3433,6 +3433,20 @@ func (c *VolumeController) checkReplicaDiskPressuredSchedulableCandidates(volume
 		return errors.Errorf("%v setting is 0, skip auto-balance replicas in disk pressure", types.SettingNameReplicaAutoBalanceDiskPressurePercentage)
 	}
 
+	allowEmptyDiskSelectorVolume, err := c.ds.GetSettingAsBool(types.SettingNameAllowEmptyDiskSelectorVolume)
+	if err != nil {
+		return err
+	}
+
+	var biDiskSelector []string
+	if volume.Spec.BackingImage != "" {
+		bi, err := c.ds.GetBackingImageRO(volume.Spec.BackingImage)
+		if err != nil {
+			return err
+		}
+		biDiskSelector = bi.Spec.DiskSelector
+	}
+
 	nodes, err := c.ds.ListNodesRO()
 	if err != nil {
 		return err
@@ -3468,7 +3482,7 @@ func (c *VolumeController) checkReplicaDiskPressuredSchedulableCandidates(volume
 			continue
 		}
 
-		if !diskSpec.AllowScheduling {
+		if eligible, _, _ := c.scheduler.IsDiskEligibleForVolume(diskSpec, diskStatus, volume, allowEmptyDiskSelectorVolume, biDiskSelector); !eligible {
 			continue
 		}
 
--- a/scheduler/replica_scheduler.go
+++ b/scheduler/replica_scheduler.go
@@ -547,18 +547,9 @@ func (rcs *ReplicaScheduler) filterNodeDisksForReplica(node *longhorn.Node, disk
 			continue
 		}
 
-		isV1EngineFilesystemDisk := types.IsDataEngineV1(volume.Spec.DataEngine) && diskSpec.Type == longhorn.DiskTypeFilesystem
-		isV2EngineBlockDisk := types.IsDataEngineV2(volume.Spec.DataEngine) && diskSpec.Type == longhorn.DiskTypeBlock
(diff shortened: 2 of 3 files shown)


controller/volume_controller.go
carterli0407-cell: In [replica_scheduler.go](https://github.com/longhorn/longhorn-manager/blob/a6b529e1914e58953e49edb3d9d6a46f8477bdd2/scheduler/replica_scheduler.go#L174), diskSpec.EvictionRequested is checked, but it doesn't checked here.

carterli0407-cell (commented):
No unit test was added for checkReplicaDiskPressuredSchedulableCandidates?

controller/volume_controller.go
derekbit: Added. Thank you.

derekbit:
> No unit test was added for checkReplicaDiskPressuredSchedulableCandidates?

We can add a test case for it in the future and consolidate the similar functions. Currently, we can leverage e2e

controller/volume_controller.go
```diff
@@ -3468,7 +3473,22 @@ func (c *VolumeController) checkReplicaDiskPressuredSchedulableCandidates(volume
 			continue
 		}
 
-		if !diskSpec.AllowScheduling {
+		if !diskSpec.AllowScheduling || diskSpec.EvictionRequested {
+			continue
+		}
+
+		// Validate disk type compatibility with the volume's data engine.
```
carterli0407-cell: The function was not validate disk and engine compatibility. But now it's updated.

carterli0407-cell (commented):
LGTM.

controller/volume_controller.go
```diff
@@ -3468,7 +3473,22 @@ func (c *VolumeController) checkReplicaDiskPressuredSchedulableCandidates(volume
...
+		isV2EngineBlockDisk := types.IsDataEngineV2(volume.Spec.DataEngine) && diskSpec.Type == longhorn.DiskTypeBlock
+		if !isV1EngineFilesystemDisk && !isV2EngineBlockDisk {
+			continue
+		}
+
+		if !datastore.IsSupportedVolumeSize(volume.Spec.DataEngine, diskStatus.FSType, volume.Spec.Size) {
+			continue
+		}
+
+		if !types.IsSelectorsInTags(diskSpec.Tags, volume.Spec.DiskSelector, allowEmptyDiskSelectorVolume) {
```
davidcheng0922: Do we also need to handle the `volume.Spec.BackingImage != ""` case here?

davidcheng0922 (commented):
Overall LGTM.

One question, it looks like this path is mirroring part of the filtering logic in [filterNodeDisksForReplica()](https://github.com/longhorn/longhorn-manager/blob/master/scheduler/replica_scheduler.go#L372), but not the full set of checks. 

I'm wondering if we may still have other edge cases where this function accepts a disk candidate that the scheduler later rejects.

Should we consider moving the common validation into a shared helper so both paths stay aligned?

c3y1huang:
> One question, it looks like this path is mirroring part of the filtering logic in [filterNodeDisksForReplica()](https://github.com/longhorn/longhorn-manager/blob/master/scheduler/replica_scheduler.go#L372), but not the full set of checks.
> 
> I'm wondering if we may still have other edge cases where this function accepts a disk candidate that the scheduler later rejects.
> 
> Should we consider moving the common validation into a shared helper so both paths stay aligned?

Fair point. We could either extract to a shared helper or expose a dry-run on the scheduler.

controller/volume_controller.go
derekbit: Addressed. Thank you.

davidcheng0922 (dismissed):
LGTM

scheduler/replica_scheduler.go
```diff
@@ -307,7 +307,7 @@ func (rcs *ReplicaScheduler) getDiskCandidates(nodeInfo map[string]*longhorn.Nod
 	errs := multierr.NewMultiError()
 
 	biNodeSelector := []string{}
-	biDiskSelector := []string{}
+	var biDiskSelector []string
```
davidcheng0922: NIT: maybe unify the declaration style here?

derekbit: Updated. Thank you.

davidcheng0922 (approved):

derekbit:
@mergify backport v1.12.x v1.11.x v1.10.x

PR #4773, opened 2026-05-21
feat: support CPU list format for V2 Data Engine CPU Mask setting

derekbit:

Issue Longhorn/longhorn#13166

Support CPU list format for V2 Data Engine CPU Mask setting with automatic conversion to hex mask

### DIFF
--- a/types/setting.go
+++ b/types/setting.go
@@ -3,6 +3,7 @@ package types
 import (
 	"encoding/json"
 	"fmt"
+	"math/big"
 	"slices"
 	"strconv"
 	"strings"
@@ -41,6 +42,11 @@ const (
 	BackupBlockSize2Mi           = 2 * BackupBlockSizeMi
 	BackupBlockSize16Mi          = 16 * BackupBlockSizeMi
 	BackupBlockSizeInvalid int64 = -1
+
+	// maxCPU is the maximum supported CPU number.
+	// This is a reasonable upper bound to prevent resource exhaustion from malicious input
+	// while supporting modern high-core-count servers.
+	maxCPU = 1023
 )
 
 type SettingType string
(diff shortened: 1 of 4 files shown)


derekbit:
@mergify backport v1.12.x

types/setting.go
```diff
@@ -2858,3 +2864,138 @@ func parseSettingSingleString(definition SettingDefinition, value string) (map[l
 
 	return values, nil
 }
+
+// maxCPU is the maximum supported CPU number.
+// This is a reasonable upper bound to prevent resource exhaustion from malicious input
+// while supporting modern high-core-count servers.
+const maxCPU = 1023
```
davidcheng0922: Nit: should maxCPU be moved to the top-level const block?

types/setting.go
```diff
@@ -2858,3 +2864,142 @@ func parseSettingSingleString(definition SettingDefinition, value string) (map[l
...
+			return false
+		}
+	}
+	return true
+}
+
+// NormalizeCPUMask normalizes the CPU mask value. If it's already a hex mask, it is
+// validated to be non-zero and returned as-is. If it's a CPU list format (e.g., "1-3,5"),
+// it's converted to a hex mask.
+func NormalizeCPUMask(value string) (string, error) {
```
davidcheng0922, line 2975: CPU-list input is set at 1023, but larger hex masks are still accepted.

For example, 1024 is rejected while the equivalent hex mask is accepted. Should we enforce the same limit for hex input?

webhook/resources/setting/mutator.go
```diff
@@ -63,9 +64,47 @@ func (s *settingMutator) mutate(newObj runtime.Object) (admission.PatchOps, erro
...
 	}
 
+	// Normalize CPU mask: convert CPU list format (e.g., "1-3,5") to hex mask (e.g., "0x2e")
+	if types.SettingName(setting.Name) == types.SettingNameDataEngineCPUMask {
+		value, err = normalizeCPUMaskSettingValue(value)
+		if err != nil {
+			return nil, werror.NewInvalidError(fmt.Sprintf("failed to normalize CPU mask for setting %s: %v", setting.Name, err), "value")
+		}
+	}
+
```
davidcheng0922: For data-engine-cpu-mask, should customized default settings also go through the same normalization as the mutating webhook? 

Otherwise definition.Default may keep a CPU-list value like "1-3" while the persisted Setting value is converted to hex.

types/setting.go
derekbit: Updated. Thank you.

webhook/resources/setting/mutator.go
derekbit: Good catch! Addressed. Thank you.

types/setting.go
derekbit: Addressed. Thank you.

davidcheng0922 (approved):
LGTM

PR #4775, opened 2026-05-21
feat: restart IM pod when hugepage settings change and no instances are running

derekbit:

Issue Longhorn/longhorn#13170

When DataEngineHugepageEnabled or DataEngineMemorySize settings change,
the IM controller now detects the mismatch between the pod's hugepages-2Mi
resource limit / --spdk-memory-size arg and the current settings. If no
instances are running, the pod is deleted and recreated with the updated
values.

A safety guard prevents deletion when the node lacks sufficient hugepage
capacity to reschedule the replacement pod — avoiding a stuck Pending
state. On verification errors, the guard also blocks deletion (fail-safe)
until the next successful reconcile.

Non-hugepage danger-zone settings (TaintToleration, PriorityClass, etc.)
retain their existing behavior unchanged for V1 and V2 with sufficient
node capacity.

### DIFF
--- a/controller/instance_manager_controller.go
+++ b/controller/instance_manager_controller.go
@@ -222,7 +222,9 @@ func (imc *InstanceManagerController) isResponsibleForSetting(obj interface{}) b
 
 	return types.SettingName(setting.Name) == types.SettingNameKubernetesClusterAutoscalerEnabled ||
 		types.SettingName(setting.Name) == types.SettingNameDataEngineCPUMask ||
-		types.SettingName(setting.Name) == types.SettingNameOrphanResourceAutoDeletion
+		types.SettingName(setting.Name) == types.SettingNameOrphanResourceAutoDeletion ||
+		types.SettingName(setting.Name) == types.SettingNameDataEngineHugepageEnabled ||
+		types.SettingName(setting.Name) == types.SettingNameDataEngineMemorySize
 }
 
 func isInstanceManagerPod(obj interface{}) bool {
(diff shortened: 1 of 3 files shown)


controller/instance_manager_controller.go
```diff
@@ -657,12 +659,24 @@ func (imc *InstanceManagerController) handlePod(im *longhorn.InstanceManager) er
...
+	nodeHasEnoughHugepages, err = imc.nodeHasEnoughHugepages(im)
+	if err != nil {
+		log.WithError(err).Warnf("Failed to check node hugepage capacity for instance manager pod %v", im.Name)
+	} else {
+		hasHugepageCapacityInfo = true
+	}
+
+	shouldSkipDeletionForHugepageOnlyMismatch := hasHugepageCapacityInfo &&
+		!nodeHasEnoughHugepages &&
+		areOnlyHugepageSettingsUnsynced(unSyncedDangerSettings)
```
mantissahz: Is it possible to merge `imc.nodeHasEnoughHugepages(im)` and `case types.SettingNameDataEngineHugepageEnabled, types.SettingNameDataEngineMemorySize:` into an independent function as `imc.isDateEngineCPUMaskApplied` does?

controller/instance_manager_controller.go
```diff
@@ -657,12 +659,42 @@ func (imc *InstanceManagerController) handlePod(im *longhorn.InstanceManager) er
...
 	if err != nil {
 		return err
 	}
 
-	isPodDeletionNotRequired := (isSettingSynced && dataEngineCPUMaskIsApplied) || areInstancesRunningInPod || isPodDeletedOrNotRunning
+	if !hugepageSettingsSynced && !isPodDeletedOrNotRunning && !areInstancesRunningInPod {
+		unSyncedSettings = append(unSyncedSettings, types.SettingNameDataEngineHugepageEnabled, types.SettingNameDataEngineMemorySize)
+		im.Status.Conditions = types.SetCondition(im.Status.Conditions, longhorn.InstanceManagerConditionTypeSettingSynced,
+			longhorn.ConditionStatusFalse, longhorn.InstanceManagerConditionReasonSettingNotSynced, fmt.Sprintf("Settings %v are not synced", unSyncedSettings))
+	}
```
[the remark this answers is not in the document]
derekbit: The two settings are inherently coupled — DataEngineHugepageEnabled and DataEngineMemorySize jointly determine the pod's hugepages-2Mi limit (formula: enabled ? memorySize : 0). A change to either one causes the hugepage limit check to fail. They cannot be meaningfully attributed in isolation.

controller/instance_manager_controller.go
```diff
@@ -657,12 +659,47 @@ func (imc *InstanceManagerController) handlePod(im *longhorn.InstanceManager) er
...
+	// hugepageSettingApplied=true with hugepageSettingsSynced=false indicates
+	// exactly this case: not synced but node lacks capacity.
+	isHugepageDeletionBlocked := hugepageSettingApplied && !hugepageSettingsSynced && !isPodDeletedOrNotRunning && !areInstancesRunningInPod
+	if isHugepageDeletionBlocked {
+		im.Status.Conditions = types.SetCondition(im.Status.Conditions, longhorn.InstanceManagerConditionTypeSettingSynced,
+			longhorn.ConditionStatusFalse, longhorn.InstanceManagerConditionReasonSettingNotSynced,
+			fmt.Sprintf("Settings %v are not synced; skipping pod deletion because node lacks hugepage capacity for rescheduling", unSyncedSettings))
+		log.Warnf("Skipping deletion of instance manager pod %v because node lacks hugepage capacity for rescheduling", im.Name)
+		return nil
+	}
```
c3y1huang, line 698: Not sure if there could be a case for redundant write (dead work). Suggest consolidating to a single decision point.

controller/instance_manager_controller.go
```diff
@@ -657,12 +659,47 @@ func (imc *InstanceManagerController) handlePod(im *longhorn.InstanceManager) er
...
+	// settings are unsynced). Deleting the pod in this state would leave the
+	// replacement pod unschedulable (Pending) due to insufficient hugepages.
+	// hugepageSettingApplied=true with hugepageSettingsSynced=false indicates
+	// exactly this case: not synced but node lacks capacity.
+	isHugepageDeletionBlocked := hugepageSettingApplied && !hugepageSettingsSynced && !isPodDeletedOrNotRunning && !areInstancesRunningInPod
+	if isHugepageDeletionBlocked {
+		im.Status.Conditions = types.SetCondition(im.Status.Conditions, longhorn.InstanceManagerConditionTypeSettingSynced,
+			longhorn.ConditionStatusFalse, longhorn.InstanceManagerConditionReasonSettingNotSynced,
+			fmt.Sprintf("Settings %v are not synced; skipping pod deletion because node lacks hugepage capacity for rescheduling", unSyncedSettings))
+		log.Warnf("Skipping deletion of instance manager pod %v because node lacks hugepage capacity for rescheduling", im.Name)
```
c3y1huang: The condition and log message might be misleading. This looks more like a verification failure than an insufficient capacity issue.

derekbit: Updated. Thank you.

controller/instance_manager_controller.go
```diff
@@ -746,57 +781,64 @@ func (imc *InstanceManagerController) annotateCASafeToEvict(im *longhorn.Instanc
...
-			isSettingSynced, err = imc.isSettingStorageNetworkSynced(setting, pod)
-		case types.SettingNameV1DataEngine, types.SettingNameV2DataEngine:
-			isSettingSynced, err = imc.isSettingDataEngineSynced(settingName, im)
-		case types.SettingNameInstanceManagerPodLivenessProbeTimeout:
-			isSettingSynced, err = imc.isSettingInstanceManagerPodLivenessProbeTimeoutSynced(setting, pod)
-		case types.SettingNameLogPath:
-			// TODO: Support log path for v1 data engine.
-			if types.IsDataEngineV2(im.Spec.DataEngine) {
-				isSettingSynced, err = imc.isSettingLogPathSynced(setting, pod)
+		case types.SettingNameDataEngineHugepageEnabled, types.SettingNameDataEngineMemorySize:
```
mantissahz: I believe we don't need to add this case and the following `default` part, right?
We can keep it as original, but still send out the `unSyncedDangerSettings`.

derekbit: You are right! It is an old desgin.

controller/instance_manager_controller.go
```diff
@@ -943,6 +978,171 @@ func (imc *InstanceManagerController) isSettingInterruptModeEnabledSynced(settin
...
+	if len(pod.Spec.Containers) == 0 {
+		return false, nil
+	}
+
+	currentMemorySize := getContainerArgValue(pod.Spec.Containers[0].Args, "--spdk-memory-size")
+	expectedMemorySize := fmt.Sprintf("%d", memorySize)
+	return currentMemorySize == expectedMemorySize, nil
+}
+
+func getContainerArgValue(args []string, flag string) string {
```
c3y1huang: NIT: Suggest moving this to controller/utils.go.

derekbit: Moved. Thank you.

c3y1huang (approved):

mantissahz (approved):

derekbit:
@mergify backport v1.12.x

PR #4777, opened 2026-05-22
feat: reject v2 backing image creation and remove monitoring

derekbit:

Issue Longhorn/longhorn#13181

v2 backing image is removed since v1.12.0

### DIFF
--- a/controller/instance_manager_controller.go
+++ b/controller/instance_manager_controller.go
@@ -80,7 +80,6 @@ type InstanceManagerMonitor struct {
 	logger logrus.FieldLogger
 
 	Name         string
-	imName       string
 	controllerID string
 
 	ds                 *datastore.DataStore
@@ -93,8 +92,7 @@ type InstanceManagerMonitor struct {
 
 	nodeCallback func(nodeName string)
 
-	client      *engineapi.InstanceManagerClient
-	proxyClient engineapi.EngineClientProxy
+	client *engineapi.InstanceManagerClient
 }
 
 type instanceProcessMap map[string]longhorn.InstanceProcess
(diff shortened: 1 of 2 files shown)


derekbit:
@mergify backport v1.12.x

mantissahz (dismissed):
LGTM.
Just one question:
So we don't care about the old created bis for v2, right?

derekbit:
> LGTM. Just one question: So we don't care about the old created bis for v2, right?

If v2 bi is unable to created, bis won't be created. but yes, we can remove them all in the future. 
Currently, I just remove the monitoring and reject bi creation.

mantissahz (approved):

COLDTURNIP:
> > LGTM. Just one question: So we don't care about the old created bis for v2, right?
> 
> If v2 bi is unable to created, bis won't be created. but yes, we can remove them all in the future. Currently, I just remove the monitoring and reject bi creation.

If we allow volumes to use the existing v2 image copies, I think it would be dangerous to stop monitoring in the instance manager.

COLDTURNIP:
> > > LGTM. Just one question: So we don't care about the old created bis for v2, right?
> > 
> > 
> > If v2 bi is unable to created, bis won't be created. but yes, we can remove them all in the future. Currently, I just remove the monitoring and reject bi creation.
> 
> If we allow volumes to use the existing v2 image copies, I think it would be dangerous to stop monitoring in the instance manager.

Per discussion with @derekbit , we don't support v2 BI volumes to upgrade to v1.12.0, hence we don't maintain the image status anymore for v2 BI copies. Users need to recreate the volume to remove the dependency.

PR #4779, opened 2026-05-22
fix(csi): use in-place upddate to prevent 0-replica window during upg…

carterli0407-cell:
Issue [#13105](https://github.com/longhorn/longhorn/issues/13105)

During upgrade, `deploy()` deleted then recreated CSI Deployments (attacher/provisioner/resizer/snapshotter), causing a guaranteed window with 0 running replicas regardless of `maxUnavailable` setting.

Fix: use `Update()` for Deployments so Kubernetes rolling update strategy is preserved and at least 1 replica stays running.
E2E: `Test CSI Components Rolling Update Configuration During Upgrade`  passed 2/2 runs on v1.12.x Jenkins environment.
https://ci.longhorn.io/job/public/job/v1.12.x/job/v1.12.x-longhorn-e2e-tests-sles-amd64/72/
## Summary by CodeRabbit

* **Refactor**
  * Improved CSI component deployment to support in-place updates, reducing disruption during reconciliation cycles.

* **Tests**
  * Added comprehensive tests to verify proper update behavior during deployment operations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/longhorn/longhorn-manager/pull/4779?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/csi/deployment.go
+++ b/csi/deployment.go
@@ -15,8 +15,9 @@ import (
 	clientset "k8s.io/client-go/kubernetes"
 
 	"github.com/longhorn/longhorn-manager/datastore"
-	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
 	"github.com/longhorn/longhorn-manager/types"
+
+	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
 )
 
 const (
@@ -94,7 +95,7 @@ func NewAttacherDeployment(namespace, serviceAccount, attacherImage, rootDir str
 
 func (a *AttacherDeployment) Deploy(kubeClient *clientset.Clientset) error {
 	return deploy(kubeClient, a.deployment, "deployment",
-		deploymentCreateFunc, deploymentDeleteFunc, deploymentGetFunc)
+		deploymentCreateFunc, deploymentDeleteFunc, deploymentGetFunc, deploymentUpdateFunc)
 }
 
 func (a *AttacherDeployment) Cleanup(kubeClient *clientset.Clientset) {
@@ -155,7 +156,7 @@ func NewProvisionerDeployment(namespace, serviceAccount, provisionerImage, rootD
 
 func (p *ProvisionerDeployment) Deploy(kubeClient *clientset.Clientset) error {
 	return deploy(kubeClient, p.deployment, "deployment",
-		deploymentCreateFunc, deploymentDeleteFunc, deploymentGetFunc)
+		deploymentCreateFunc, deploymentDeleteFunc, deploymentGetFunc, deploymentUpdateFunc)
 }
 
 func (p *ProvisionerDeployment) Cleanup(kubeClient *clientset.Clientset) {
@@ -222,7 +223,7 @@ func NewResizerDeployment(namespace, serviceAccount, resizerImage, rootDir strin
 
 func (p *ResizerDeployment) Deploy(kubeClient *clientset.Clientset) error {
 	return deploy(kubeClient, p.deployment, "deployment",
-		deploymentCreateFunc, deploymentDeleteFunc, deploymentGetFunc)
+		deploymentCreateFunc, deploymentDeleteFunc, deploymentGetFunc, deploymentUpdateFunc)
 }
 
 func (p *ResizerDeployment) Cleanup(kubeClient *clientset.Clientset) {
@@ -279,7 +280,7 @@ func NewSnapshotterDeployment(namespace, serviceAccount, snapshotterImage, rootD
 
 func (p *SnapshotterDeployment) Deploy(kubeClient *clientset.Clientset) error {
 	return deploy(kubeClient, p.deployment, "deployment",
-		deploymentCreateFunc, deploymentDeleteFunc, deploymentGetFunc)
+		deploymentCreateFunc, deploymentDeleteFunc, deploymentGetFunc, deploymentUpdateFunc)
 }
 
 func (p *SnapshotterDeployment) Cleanup(kubeClient *clientset.Clientset) {
@@ -622,7 +623,7 @@ func NewPluginDeployment(namespace, serviceAccount, nodeDriverRegistrarImage, li
 
 func (p *PluginDeployment) Deploy(kubeClient *clientset.Clientset) error {
 	return deploy(kubeClient, p.daemonSet, "daemon set",
-		daemonSetCreateFunc, daemonSetDeleteFunc, daemonSetGetFunc)
+		daemonSetCreateFunc, daemonSetDeleteFunc, daemonSetGetFunc, nil)
 }
 
 func (p *PluginDeployment) Cleanup(kubeClient *clientset.Clientset) {
@@ -653,7 +654,7 @@ func NewCSIDriverObject(storageCapacityEnabled bool) *DriverObjectDeployment {
 
 func (d *DriverObjectDeployment) Deploy(kubeClient *clientset.Clientset) error {
 	return deploy(kubeClient, d.obj, "CSI Driver",
-		csiDriverObjectCreateFunc, csiDriverObjectDeleteFunc, csiDriverObjectGetFunc)
+		csiDriverObjectCreateFunc, csiDriverObjectDeleteFunc, csiDriverObjectGetFunc, nil)
 }
 
 func (d *DriverObjectDeployment) Cleanup(kubeClient *clientset.Clientset) {
--- a/csi/deployment_util.go
+++ b/csi/deployment_util.go
@@ -166,9 +166,10 @@ func getCommonDeployment(commonName, namespace, serviceAccount, image, rootDir s
 
 type resourceCreateFunc func(kubeClient *clientset.Clientset, obj runtime.Object) error
 type resourceDeleteFunc func(kubeClient *clientset.Clientset, name, namespace string) error
+type resourceUpdateFunc func(kubeClient *clientset.Clientset, obj runtime.Object) error
 
 func deploy(kubeClient *clientset.Clientset, obj runtime.Object, resource string,
-	createFunc resourceCreateFunc, deleteFunc resourceDeleteFunc, getFunc util.ResourceGetFunc) (err error) {
+	createFunc resourceCreateFunc, deleteFunc resourceDeleteFunc, getFunc util.ResourceGetFunc, updateFunc resourceUpdateFunc) (err error) {
 
 	objMeta, err := meta.Accessor(obj)
 	if err != nil {
@@ -205,6 +206,13 @@ func deploy(kubeClient *clientset.Clientset, obj runtime.Object, resource string
 			logrus.Infof("Detected %v %v CSI Git commit %v version %v has already been deployed",
 				resource, name, annos[AnnotationCSIGitCommit], annos[AnnotationCSIVersion])
 			return nil
+		}
+		// For Deployments, update in-place to let Kubernetes perform a rolling update,
+		// which respects maxUnavailable and avoids a 0-replica window.
+		// For other resource types (DaemonSet, CSIDriver), fall back to delete+recreate.
+		if updateFunc != nil && existingMeta.GetDeletionTimestamp() == nil {
+			logrus.Infof("Updating %s %s", resource, name)
+			return updateFunc(kubeClient, obj)
 		}
 		// otherwise clean up the old deployment
 		if err := cleanup(kubeClient, obj, resource, deleteFunc, getFunc); err != nil {
@@ -357,6 +365,24 @@ func deploymentCreateFunc(kubeClient *clientset.Clientset, obj runtime.Object) e
 		return fmt.Errorf("failed to convert back the object")
 	}
 	_, err := kubeClient.AppsV1().Deployments(o.Namespace).Create(context.TODO(), o, metav1.CreateOptions{})
+	return err
+}
+
+func deploymentUpdateFunc(kubeClient *clientset.Clientset, obj runtime.Object) error {
+	o, ok := obj.(*appsv1.Deployment)
+	if !ok {
+		return fmt.Errorf("failed to convert object to *appsv1.Deployment for update, got %T", obj)
+	}
+	_, err := util.RetryOnConflictCause(func() (interface{}, error) {
+		latest, err := kubeClient.AppsV1().Deployments(o.Namespace).Get(context.TODO(), o.Name, metav1.GetOptions{})
+		if err != nil {
+			return nil, err
+		}
+		latest.Annotations = o.Annotations
+		latest.Labels = o.Labels
+		latest.Spec = o.Spec
+		return kubeClient.AppsV1().Deployments(o.Namespace).Update(context.TODO(), latest, metav1.UpdateOptions{})
+	})
 	return err
 }
 
--- a/csi/deployment_util_test.go
+++ b/csi/deployment_util_test.go
@@ -3,8 +3,14 @@ package csi
 import (
 	"testing"
 
+	"k8s.io/apimachinery/pkg/runtime"
+
 	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	clientset "k8s.io/client-go/kubernetes"
+
+	longhornmeta "github.com/longhorn/longhorn-manager/meta"
 )
 
 func TestNeedToUpdatePodAntiAffinity(t *testing.T) {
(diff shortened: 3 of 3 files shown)


csi/deployment_util.go
```diff
@@ -206,6 +207,14 @@ func deploy(kubeClient *clientset.Clientset, obj runtime.Object, resource string
...
 			return nil
 		}
+		// For Deployments, update in-place to let Kubernetes perform a rolling update,
+		// which respects maxUnavailable and avoids a 0-replica window.
+		// For other resource types (DaemonSet, CSIDriver), fall back to delete+recreate.
+		if updateFunc != nil {
+			objMeta.SetResourceVersion(existingMeta.GetResourceVersion())
+			logrus.Infof("Updating %s %s", resource, name)
+			return updateFunc(kubeClient, obj)
+		}
```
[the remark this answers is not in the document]
carterli0407-cell: Updated

csi/deployment_util.go
```diff
@@ -360,6 +369,15 @@ func deploymentCreateFunc(kubeClient *clientset.Clientset, obj runtime.Object) e
 	return err
 }
 
+func deploymentUpdateFunc(kubeClient *clientset.Clientset, obj runtime.Object) error {
+	o, ok := obj.(*appsv1.Deployment)
+	if !ok {
+		return fmt.Errorf("failed to convert object to *appsv1.Deployment for update, got %T", obj)
+	}
+	_, err := kubeClient.AppsV1().Deployments(o.Namespace).Update(context.TODO(), o, metav1.UpdateOptions{})
+	return err
```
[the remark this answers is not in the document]
derekbit: Fair concern. @carterli0407-cell Please check this.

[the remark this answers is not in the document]
carterli0407-cell: @derekbit , I understand the concern now - deploymentUpdateFunc is new code in this PR and uses context.TODO(), which is a placeholder that blocks callers from cancelling or timing out the retry loop.

To fix this properly, the context needs to thread through three layers:

resourceUpdateFunc type -> add ctx context.Context as first parameter
deploy() -> add ctx parameter and pass it to updateFunc
All Deploy() methods in deployment.go -> add ctx context.Context parameter and pass it down
This also means updating all call sites in the manager that call .Deploy(kubeClient) to pass a real context.

Before I proceed with the full cascade - is this the direction you had in mind, or would you prefer a narrower fix (e.g., only change deploymentUpdateFunc internally and keep resourceUpdateFunc/deploy() signatures unchanged for now to limit blast radius)?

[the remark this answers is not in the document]
derekbit: @carterli0407-cell Not fully understand what you mentioned and if it is related to the issue. Is the `conflict` related to the `context`? 

- Could you elaborate more on their relationship?
- Could you explain when a conflict happen while performing a CR update?

IMO. the comment from @copilot is for the retry on conflict. Could you review the [comment](https://github.com/longhorn/longhorn-manager/pull/4779#discussion_r3300755340) and check `RetryOnConflict` mentioned in the comment again?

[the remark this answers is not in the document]
carterli0407-cell: I've reverted that thinking.
I've wrapped the Update() as the Copilot comment suggested.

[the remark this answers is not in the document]
derekbit: @carterli0407-cell could you explain this?

> Could you explain when a conflict happen while performing a CR update?

csi/deployment_util.go
```diff
@@ -360,6 +369,15 @@ func deploymentCreateFunc(kubeClient *clientset.Clientset, obj runtime.Object) e
...
+}
```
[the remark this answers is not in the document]
derekbit: @carterli0407-cell Unresolved. The comment seems not addressed.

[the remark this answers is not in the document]
carterli0407-cell: return retry.RetryOnConflict(retry.DefaultRetry, func() error {
		latest, err := kubeClient.AppsV1().Deployments(o.Namespace).Get(context.TODO(), o.Name, metav1.GetOptions{})
		if err != nil {
			return err
		}
		latest.Annotations = o.Annotations
		latest.Labels = o.Labels
		latest.Spec = o.Spec
		_, err = kubeClient.AppsV1().Deployments(o.Namespace).Update(context.TODO(), latest, metav1.UpdateOptions{})
		return err
	})

I've wrapped the update in RetryOnConflict as you can see.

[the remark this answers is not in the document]
derekbit: Why not use `util.RetryOnConflictCause`?

csi/deployment_util.go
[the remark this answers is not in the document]
carterli0407-cell: When deploy() runs, it calls Get() to fetch the existing Deployment and captures its ResourceVersion. Then it calls Update() with that ResourceVersion.

However, Kubernetes uses optimistic concurrency, the API server only accepts an Update() if the ResourceVersion in the request matches what is currently stored in etcd. If any other controller modifies the same Deployment between our Get() and Update() even just updating a status field, the ResourceVersion gets bumped. Our Update() then carries a stale ResourceVersion and the API server rejects it with HTTP 409 Conflict.
Without retry, this causes the entire CSI reconcile/upgrade to fail with no recovery.

The fix wraps the Update() in retry.RetryOnConflict, which on a 409 response automatically re-Get()s the latest Deployment (fresh ResourceVersion), re-applies our desired Annotations, Labels, and Spec, then retries the Update(). This is the standard client-go pattern for safe updates.

csi/deployment_util.go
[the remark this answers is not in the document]
carterli0407-cell: @derekbit That's a great point. util.RetryOnConflictCause is the project's own helper and is already used throughout this codebase, so it's more consistent to use it here instead of importing k8s.io/client-go/util/retry just for this one function. I'll update. Thanks.

csi/deployment.go
```diff
@@ -622,7 +623,7 @@ func NewPluginDeployment(namespace, serviceAccount, nodeDriverRegistrarImage, li
 
 func (p *PluginDeployment) Deploy(kubeClient *clientset.Clientset) error {
 	return deploy(kubeClient, p.daemonSet, "daemon set",
-		daemonSetCreateFunc, daemonSetDeleteFunc, daemonSetGetFunc)
+		daemonSetCreateFunc, daemonSetDeleteFunc, daemonSetGetFunc, nil)
 }
```
derekbit, line 627: Don't we set `deploymentUpdateFunc`?

csi/deployment.go
```diff
@@ -653,7 +654,7 @@ func NewCSIDriverObject(storageCapacityEnabled bool) *DriverObjectDeployment {
 
 func (d *DriverObjectDeployment) Deploy(kubeClient *clientset.Clientset) error {
 	return deploy(kubeClient, d.obj, "CSI Driver",
-		csiDriverObjectCreateFunc, csiDriverObjectDeleteFunc, csiDriverObjectGetFunc)
+		csiDriverObjectCreateFunc, csiDriverObjectDeleteFunc, csiDriverObjectGetFunc, nil)
```
derekbit, line 657: Same question as above

csi/deployment.go
carterli0407-cell: Good catch, 
I passed nil intentionally so it falls back to the existing delete+recreate path. If we want rolling updates for the DaemonSet as well, that would require a separate daemonSetUpdateFunc, but DaemonSets already perform rolling updates by default on pod template changes, so delete+recreate on version upgrade is the right behavior here.

csi/deployment.go
carterli0407-cell: Same reason.

csi/deployment_util.go
```diff
@@ -206,6 +207,14 @@ func deploy(kubeClient *clientset.Clientset, obj runtime.Object, resource string
 				resource, name, annos[AnnotationCSIGitCommit], annos[AnnotationCSIVersion])
 			return nil
 		}
+		// For Deployments, update in-place to let Kubernetes perform a rolling update,
+		// which respects maxUnavailable and avoids a 0-replica window.
+		// For other resource types (DaemonSet, CSIDriver), fall back to delete+recreate.
+		if updateFunc != nil && existingMeta.GetDeletionTimestamp() == nil {
+			objMeta.SetResourceVersion(existingMeta.GetResourceVersion())
```
derekbit: Why do we need to set `SetResourceVersion`?

csi/deployment_util.go
```diff
@@ -360,6 +369,24 @@ func deploymentCreateFunc(kubeClient *clientset.Clientset, obj runtime.Object) e
...
+	if !ok {
+		return fmt.Errorf("failed to convert object to *appsv1.Deployment for update, got %T", obj)
+	}
+	_, err := util.RetryOnConflictCause(func() (interface{}, error) {
+		latest, err := kubeClient.AppsV1().Deployments(o.Namespace).Get(context.TODO(), o.Name, metav1.GetOptions{})
+		if err != nil {
+			return nil, err
+		}
+		latest.Annotations = o.Annotations
+		latest.Labels = o.Labels
```
derekbit, line 382: Why do we update `Annotations` and `Labels`?

csi/deployment.go
derekbit: Got it. Then, we only need to handle csi sidecar components here.

csi/deployment_util.go
carterli0407-cell: The real `deploymentUpdateFunc` does a Re-GET inside itself, so it doesn't need this to be set, but a simpler future `updateFunc` that calls `Update(obj)` directly (without Re-GET) would absolutely require it

csi/deployment_util.go
carterli0407-cell: `Annotations` must be copied because `deploy()` writes the current build version into obj before calling `updateFunc`. If those annotations are not propagated onto latest before the Update call, the next run of deploy() will compare the annotations, see a mismatch, and trigger another update - causing an infinite update loop.

`Labels` is copied to propagate any label changes to the live resource, keeping it consistent with what the manager expects to select.

As you know, `Spec` is the core of the update - it carries the new container image and configuration.

csi/deployment_util.go
derekbit: IIUC, resouce version is automaticlly updated while doing update operation. What will happen if we don't it by ourselves?

csi/deployment_util.go
derekbit: Where are `latest` and `o` from? Is the content of the `latest. Annotations` different than `o.Annotations`?
Same question for `Labels`.

csi/deployment_util.go
derekbit: When we want to update the image of a deployment, don't we only update the spec.image? Do we need to update resource version?

carterli0407-cell: > IIUC, resouce version is automaticlly updated while doing update operation. What will happen if we don't it by ourselves?

You are right that Kubernetes updates the ResourceVersion after a successful write, but the ResourceVersion must be included in the update request itself for the API server to accept it. This is Kubernetes' optimistic concurrency control - the API server compares the submitted ResourceVersion against what's in etcd.
 If they don't match, it rejects with a 409 Conflict. 
 Also, if ResourceVersion is empty, the API server also rejects the request.

csi/deployment_util.go
carterli0407-cell: - `o` is the desired state, the Deployment object constructed in code with the new image, new version annotations (AnnotationCSIGitCommit, AnnotationCSIVersion), etc.

- `latest` is the current live state, fetched from the cluster via Re-GET, containing the old annotations and old image, but also carrying the live ResourceVersion, Status, and any fields managed by Kubernetes.

Yes, their Annotations are different, that difference is actually what triggered the update path in the first place. `deploy()` checks `AnnotationCSIGitCommit` and `AnnotationCSIVersion` before deciding to update, so we only reach `deploymentUpdateFunc` when `o.Annotations` (new version) differs from `latest.Annotations` (old version). We copy `o.Annotations` onto latest so the new version info is persisted, preventing deploy() from triggering the update again on the next reconcile loop.

csi/deployment_util.go
derekbit: Can you check other UPDATE APIs in Longhorns? Why don't they need to update resource version by manual? Resource version will update automatically. If there is any conflict, shouldn't we get the latest resource and update again?

carterli0407-cell: Thanks, I checked other update paths in Longhorn and I think your concern is right.

Existing Longhorn update APIs generally do not manually set `ResourceVersion`. They update an object that was already fetched from the API/datastore, so that object already contains the current ResourceVersion. When there is a conflict, the common pattern is to re-GET the latest object, apply the intended field changes again, and retry the update.

In my CSI deployment code, the desired object `o` is newly constructed, so it does not naturally have the live ResourceVersion. That is why I initially set ResourceVersion in deploy(). But after adding the retry path, `deploymentUpdateFunc` now re-GETs latest and calls `Update(latest)`. Since latest already has the correct `ResourceVersion`, manually setting it on obj in `deploy()` is unnecessary.

When I tried to remove `setResourceVersion` earlier in the PR, I hit a build error, so I was a bit confused and thought I might have made another mistake. After reviewing the update pattern again, I agree that we can remove the manual `ResourceVersion` setting and rely on the retry/update function to use the latest object from the API server.

I updated.

derekbit: > the desired object o is newly constructed

IMO, we should get the existing resource and only mutate the fields that need to be update and then do UPDATE rather than construct a new one. Could you apply the same pattern in your implementation?

carterli0407-cell: > > the desired object o is newly constructed
> 
> IMO, we should get the existing resource and only mutate the fields that need to be update and then do UPDATE rather than construct a new one. Could you apply the same pattern in your implementation?

Could you please clarify in more details?
The current implementation re-GETs the latest Deployment, but then it still copies whole Annotations, Labels, and Spec from the `o`. 
You mean that is broader than necessary and can overwrite fields we do not intend to manage.
So, adjust it to follow the existing Longhorn update pattern more closely: inside the retry loop, GET the current Deployment, mutate only the fields required for this PR, then call Update(latest)?

c3y1huang (changes requested):
Remember to use your SUSE email to sign off and clean up the commit message.

carterli0407-cell:
> Remember to use your SUSE email to sign off and clean up the commit message.

I updated the commit to use my SUSE email and cleaned up the duplicated commit message/sign-off.
Thanks.

csi/deployment_util.go
derekbit: > The current implementation re-GETs the latest Deployment, but then it still copies whole Annotations, Labels, and Spec from the o.

If so, we don't need to update resource version, isn't it?
Have you tried to remove the resource version update? Any error?

> So, adjust it to follow the existing Longhorn update pattern more closely: inside the retry loop, GET the current Deployment, mutate only the fields required for this PR, then call Update(latest)?

Yes, please follow the same pattern as mentioned.

carterli0407-cell: I've already remove the resource version update in [deploy()](https://github.com/longhorn/longhorn-manager/pull/4779/changes#diff-c2d7d8c4e81287204b752f93b57bf1958cf656001606b2167805846e10e87c7cR213)
I'm not sure what is the same pattern as you mentioned.

derekbit: https://github.com/longhorn/longhorn-manager/pull/4779#discussion_r3377095562

csi/deployment_util.go
```diff
@@ -360,6 +368,24 @@ func deploymentCreateFunc(kubeClient *clientset.Clientset, obj runtime.Object) e
...
+		latest.Spec = o.Spec
+		return kubeClient.AppsV1().Deployments(o.Namespace).Update(context.TODO(), latest, metav1.UpdateOptions{})
+	})
+	return err
+}
```
derekbit, line 387: After rethinking the retry on conflict logic, do we really need ` util.RetryOnConflictCause`?
If UPDATE failed, `deploymentUpdateFunc` returns error. Will longhorn try to get and update the deployment again? If yes, we don't need the `RetryOnConflictCause`. Another questgion, how long will it rety?

csi/deployment_util.go
```diff
@@ -360,6 +368,22 @@ func deploymentCreateFunc(kubeClient *clientset.Clientset, obj runtime.Object) e
...
+	latest, err := kubeClient.AppsV1().Deployments(o.Namespace).Get(context.TODO(), o.Name, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+	latest.Annotations = o.Annotations
+	latest.Labels = o.Labels
+	latest.Spec = o.Spec
+	_, err = kubeClient.AppsV1().Deployments(o.Namespace).Update(context.TODO(), latest, metav1.UpdateOptions{})
+	return err
+}
```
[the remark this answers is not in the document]
carterli0407-cell: Yes, you are right. However, as you can see this comment, when remove the `RetryOnConflictCause`, the Copilot concerns about "read-modify-write Get -> Update can still hit a 409 Conflict".

BTW, for your question about how long it retries: `util.RetryOnConflictCause` uses `APIRetryCounts = 10`, with `APIRetryInterval = 500ms`.

csi/deployment_util.go
carterli0407-cell: Yes, you are right. However, as you can see the below Copilot's comment, when remove the RetryOnConflictCause, the Copilot concerns about "read-modify-write Get -> Update can still hit a 409 Conflict".

BTW, for your question about how long it retries: util.RetryOnConflictCause uses APIRetryCounts = 10, with APIRetryInterval = 500ms.

derekbit: I'm not sure if Copilot is aware of this. Could you try to remove `RetryOnConflictCause` and test it? It works, then we can remove `RetryOnConflictCause` to make it simple.

csi/deployment_util.go
```diff
@@ -360,6 +368,22 @@ func deploymentCreateFunc(kubeClient *clientset.Clientset, obj runtime.Object) e
...
+		return fmt.Errorf("failed to convert object to *appsv1.Deployment for update, got %T", obj)
+	}
+	latest, err := kubeClient.AppsV1().Deployments(o.Namespace).Get(context.TODO(), o.Name, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+	latest.Annotations = o.Annotations
+	latest.Labels = o.Labels
+	latest.Spec = o.Spec
+	_, err = kubeClient.AppsV1().Deployments(o.Namespace).Update(context.TODO(), latest, metav1.UpdateOptions{})
```
derekbit:
```suggestion
func deploymentUpdateFunc(kubeClient *clientset.Clientset, obj runtime.Object) error {
	o, ok := obj.(*appsv1.Deployment)
	if !ok {
		return fmt.Errorf("failed to convert object to *appsv1.Deployment for update, got %T", obj)
	}
	
	latest, err := kubeClient.AppsV1().Deployments(o.Namespace).Get(context.TODO(), o.Name, metav1.GetOptions{})
	if err != nil {
		return err
	}
	
	if reflect.DeepEqual(latest.Annotations, o.Annotations) &&
	   reflect.DeepEqual(latest.Labels, o.Labels) &&
	   reflect.DeepEqual(latest.Spec, o.Spec) {
	    return nil
	}

	latest.Annotations = o.Annotations
	latest.Labels = o.Labels
	latest.Spec = o.Spec
	_, err = kubeClient.AppsV1().Deployments(o.Namespace).Update(context.TODO(), latest, metav1.UpdateOptions{})
```

carterli0407-cell: Thanks for the suggestion.
The idea of skipping a redundant Update is good, but I have a concern with some parts:
- First, deploy() already gates on real differences [(needToUpdateImage, needToUpdatePodAntiAffinity, and the CSI version annotations)](https://github.com/longhorn/longhorn-manager/blob/55d8f773c55f33b03bf3c054dc1053ba4f7b406c/csi/deployment_util.go#L199-L208) before calling `updateFunc`, so I think additional whole-object comparison here is redundant.

- Second, reflect.DeepEqual(latest.Spec, o.Spec): `latest` comes from the API server with server-defaulted fields that the constructed `o` doesn't set - was just built locally. So the DeepEqual would almost always be false and the early-return becomes dead code.

COLDTURNIP:
LGTM, please address the comments above.

derekbit:
@carterli0407-cell 
Please add `longhorn/longhorn#13105` in your commit message

```commit 012851c9b2066dd5811452a72e743d4afe60756a
Author: Carter Li <carter.li@suse.com>
Date:   Mon Jun 8 18:23:37 2026 -0700

    fix(csi): use in-place update to prevent 0-replica window during upgrade
    
    longhorn/longhorn#13105

    Signed-off-by: Carter Li <carter.li@suse.com>
```

csi/deployment_util.go
```diff
@@ -360,6 +368,29 @@ func deploymentCreateFunc(kubeClient *clientset.Clientset, obj runtime.Object) e
...
+		reflect.DeepEqual(latest.Labels, o.Labels) &&
+		reflect.DeepEqual(latest.Spec, o.Spec) {
+		return nil
+	}
+	latest.Annotations = o.Annotations
+	latest.Labels = o.Labels
+	latest.Spec = o.Spec
+	_, err = kubeClient.AppsV1().Deployments(o.Namespace).Update(context.TODO(), latest, metav1.UpdateOptions{})
+	return err
+}
```
derekbit, line 387: After rethinking the tradeoff with and without `RetryOnConflict`, I think we can keep it.

Without `RetryOnConflict`, a Conflict causes the entire deploy-driver process to Fatalf and the Pod restarts. This is expensive. All CSI components (attacher, provisioner, resizer, snapshotter, plugin) are re-deployed from scratch, typically taking tens of seconds. RetryOnConflict resolves the same conflict in-process within milliseconds, avoiding this unnecessary restart penalty.

BTW, we can remove `reflect.DeepEqual` as you explained.

WDYT for the above changes?

```suggestion
func deploymentUpdateFunc(kubeClient *clientset.Clientset, obj runtime.Object) error {
	o, ok := obj.(*appsv1.Deployment)
	if !ok {
		return fmt.Errorf("failed to convert object to *appsv1.Deployment for update, got %T", obj)
	}

	_, err := util.RetryOnConflictCause(func() (interface{}, error) {
		latest, err := kubeClient.AppsV1().Deployments(o.Namespace).Get(context.TODO(), o.Name, metav1.GetOptions{})
		if err != nil {
			return nil, err
		}

		latest.Annotations = o.Annotations
		latest.Labels = o.Labels
		latest.Spec = o.Spec
		return kubeClient.AppsV1().Deployments(o.Namespace).Update(context.TODO(), latest, metav1.UpdateOptions{})
	})
	return err
}
```

carterli0407-cell: Agreed, thanks for thinking it through.

derekbit (approved):

derekbit:
@mergify backport v1.12.x

PR #4798, opened 2026-05-26
fix(v2): do not clear Spec.NodeID during volume deletion to prevent SPDK bdev orphan

derekbit:

Issue Longhorn/longhorn#13198

Remove the Spec.NodeID = "" assignments for engine frontends, engines,
and replicas in the v2 volume deletion path.

Setting DesireState=Stopped is sufficient to prevent ReconcileInstanceState
from re-creating instances, since only the DesireState=Running case calls
createInstance(). Clearing NodeID is redundant and removes the fallback
path in DeleteInstance() that uses NodeID to locate the Instance Manager
when Status.InstanceManagerName is empty (cleared by
syncStatusWithInstanceManager after the instance stops). Without NodeID,
DeleteInstance skips the actual SPDK instance deletion, leaving the lvol
bdev as an orphaned resource in spdk_tgt.

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -413,7 +413,6 @@ func (c *VolumeController) syncVolume(key string) (err error) {
 			for _, ef := range engineFrontends {
 				if ef.DeletionTimestamp == nil {
 					ef.Spec.DesireState = longhorn.InstanceStateStopped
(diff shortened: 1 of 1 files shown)


davidcheng0922 (approved):

derekbit:
@mergify backport v1.12.x

PR #4811, opened 2026-05-27
fix(recurring-job): correct system backup retention to protect successful backups

issmirnov:
Issue longhorn/longhorn#13203

The `system-backup` `RecurringJob`'s retention (`cleanup()`) sorts `SystemBackup` CRs by `Status.CreatedAt`, but the controller only writes that field on the successful upload path (`controller/system_backup_controller.go` ~L566). CRs in `Error` state — and CRs still in flight when `ListSystemBackup` runs — have the zero value `time.Time{}` (year 1), which sorts to the front of `filterExpiredItems`, so retention pruned the wrong CRs.

This PR makes retention state-aware:

- **Skip in-flight backups.** `cleanup()` only considers CRs in a terminal state (`Ready` or `Error`). A backup still being created (`Syncing`/`Generating`/`Uploading`/…) is never deleted and does not count toward the `retain` quota.
- **Prune failed backups before successful ones.** `Error` CRs keep their zero `Status.CreatedAt` and sort ahead of `Ready` CRs, so they are pruned first. A run of failures (e.g. a backup-target outage) can no longer evict good backups.

- Implements @derekbit's suggestion to skip non-`Ready`/`Error` states in `cleanup()`. The earlier `creationTimestamp` fallback was dropped in favor of this, since it could let accumulating `Error` CRs evict successful backups during a sustained outage.
- `isSystemBackupEligibleForRetention` treats only `Ready`/`Error` as eligible — unlike the controller's `systemBackupInFinalState`, it excludes `Deleting` (those are already on their way out).
- The recurring-job Longhorn client is made an interface so `cleanup()` is testable against a fake clientset.

#### Testing

`app/recurringjob/` unit tests, incl. `TestSystemBackupJobCleanup` driving the real `cleanup()` against a fake clientset:
- `prunes_error_backup_before_ready`, `skips_in_flight_system_backups`, `repeated_failures_do_not_evict_ready_backups`, plus label-scoping and within-retain cases.

```
$ go test ./app/recurringjob/
ok  	github.com/longhorn/longhorn-manager/app/recurringjob
```

### DIFF
--- a/app/recurringjob/systembackup_test.go
+++ b/app/recurringjob/systembackup_test.go
@@ -1,0 +1,193 @@
+package recurringjob
+
+import (
+	"context"
+	"io"
+	"testing"
+	"time"
+
+	"github.com/sirupsen/logrus"
+	"github.com/stretchr/testify/assert"
+
+	"k8s.io/apimachinery/pkg/runtime"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/longhorn/longhorn-manager/types"
+
+	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
+	lhfake "github.com/longhorn/longhorn-manager/k8s/pkg/client/clientset/versioned/fake"
+)
+
+const (
+	testNamespace       = "longhorn-system"
+	testSystemBackupJob = "daily-system-backup"
+)
+
+// systemBackupRecurringJobLabels returns the label set the system-backup
+// RecurringJob stamps on the SystemBackup CRs it creates. cleanup() lists by
+// this exact label (see (*Job).ListSystemBackup), so seeded fixtures must carry
+// it to be considered for retention.
+func systemBackupRecurringJobLabels(jobName string) map[string]string {
+	return map[string]string{
+		types.GetRecurringJobLabelKey(types.LonghornLabelRecurringJob, string(longhorn.RecurringJobTypeSystemBackup)): jobName,
+	}
+}
+
+// newLabeledSystemBackup builds a SystemBackup owned by jobName (carrying the
+// recurring-job label) in testNamespace, reusing the shared fixture builder.
+func newLabeledSystemBackup(name, jobName string, creationTime, statusCreatedAt time.Time, state longhorn.SystemBackupState) *longhorn.SystemBackup {
+	sb := newSystemBackup(name, creationTime, statusCreatedAt, state)
+	sb.Namespace = testNamespace
+	sb.Labels = systemBackupRecurringJobLabels(jobName)
+	return &sb
+}
+
+// newSystemBackupJobForTest wires a SystemBackupJob to a fake Longhorn clientset
+// seeded with the given SystemBackups. This exercises the real cleanup() path
+// (ListSystemBackup -> filter retention-eligible states ->
+// systemBackupsToNameWithTimestamps -> filterExpiredItems -> DeleteSystemBackup)
+// without needing a cluster.
+func newSystemBackupJobForTest(retain int, objs ...*longhorn.SystemBackup) *SystemBackupJob {
+	runtimeObjs := make([]runtime.Object, 0, len(objs))
+	for _, o := range objs {
+		runtimeObjs = append(runtimeObjs, o)
+	}
+	lhClient := lhfake.NewSimpleClientset(runtimeObjs...) // nolint: staticcheck
+
+	logger := logrus.New()
+	logger.SetOutput(io.Discard)
+
+	return &SystemBackupJob{
+		Job: &Job{
+			lhClient:  lhClient,
+			logger:    logger,
+			name:      testSystemBackupJob,
+			namespace: testNamespace,
+			retain:    retain,
+		},
+		logger: logrus.NewEntry(logger),
+	}
+}
+
+// remainingSystemBackupNames returns the names of SystemBackups still present in
+// the fake cluster after a cleanup pass.
+func remainingSystemBackupNames(t *testing.T, job *SystemBackupJob) []string {
+	t.Helper()
+	list, err := job.lhClient.LonghornV1beta2().SystemBackups(testNamespace).List(context.TODO(), metav1.ListOptions{})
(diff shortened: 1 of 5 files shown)


app/recurringjob/util.go
```diff
@@ -159,9 +159,20 @@ func sliceStringSafely(s string, begin, end int) string {
...
+		if ts.IsZero() {
+			ts = systemBackup.CreationTimestamp.Time
+		}
 		result = append(result, NameWithTimestamp{
 			Name:      systemBackup.Name,
-			Timestamp: systemBackup.Status.CreatedAt.Time,
+			Timestamp: ts,
 		})
 	}
 	return result
```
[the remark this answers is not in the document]
issmirnov: Fixed in followup commit.

derekbit:
cc @carterli0407-cell @mantissahz to review the PR.

carterli0407-cell (commented):
The [snapshotCRsToNameWithTimestamps](https://github.com/longhorn/longhorn-manager/blob/4352861537c90c4f9ea86ab8f933ee9dfcfe1238/app/recurringjob/util.go#L66) still uses `result := []NameWithTimestamp{}` (no pre-allocation). Not a bug and not related with this issue, but now the two functions are inconsistent in style. Is it worth aligning them?

IMO, Overall, the fix is correct, and well-tested. LGTM

issmirnov:
Thanks! Added a test to showcase deterministic behavior.

app/recurringjob/systembackup_test.go
```diff
@@ -0,0 +1,150 @@
...
+	"github.com/sirupsen/logrus"
+	"github.com/stretchr/testify/assert"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+
+	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
+	lhfake "github.com/longhorn/longhorn-manager/k8s/pkg/client/clientset/versioned/fake"
+	"github.com/longhorn/longhorn-manager/types"
+)
```
derekbit, line 20:
```suggestion
import (
	"context"
	"io"
	"testing"
	"time"

	"github.com/sirupsen/logrus"
	"github.com/stretchr/testify/assert"

	"k8s.io/apimachinery/pkg/runtime"

	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"

	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
	lhfake "github.com/longhorn/longhorn-manager/k8s/pkg/client/clientset/versioned/fake"

	"github.com/longhorn/longhorn-manager/types"
)
```

app/recurringjob/type.go
```diff
@@ -15,7 +15,7 @@ import (
 // Job is a base job that contains the necessary clients, configuration, and general information.
 type Job struct {
 	api      *longhornclient.RancherClient // Rancher client used to interact with the Longhorn API.
-	lhClient *lhclientset.Clientset        // Kubernetes clientset for Longhorn resources.
+	lhClient lhclientset.Interface         // Longhorn clientset (interface so a fake can be injected in tests).
```
derekbit, line 18: Good catch! Thanks for the fix.

app/recurringjob/systembackup_test.go
derekbit, line 20: Reorganize the imports

```suggestion
import (
	"context"
	"io"
	"testing"
	"time"

	"github.com/sirupsen/logrus"
	"github.com/stretchr/testify/assert"

	"k8s.io/apimachinery/pkg/runtime"

	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"

	"github.com/longhorn/longhorn-manager/types"

	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
	lhfake "github.com/longhorn/longhorn-manager/k8s/pkg/client/clientset/versioned/fake"
)
```

derekbit (commented):
LGTM.

@issmirnov Only one comment. Could you fix it? Thank you.

derekbit:
@mergify backport v1.11.x v1.12.x

app/recurringjob/systembackup_test.go
issmirnov: Reorganized per your suggestion — thanks!

derekbit:
@issmirnov 
For this issue, if a system-backup's `status.state` is not in either `Error` or `Ready`, we can skip it in the cleanup() logic. Then, we won't mistakenly delete it. WDYT?

issmirnov:
Done — went with your suggestion: skip non-`Ready`/`Error` states in `cleanup()`, and dropped the `creationTimestamp` fallback so an `Error` backup (zero `CreatedAt`) is pruned ahead of `Ready` ones. Bonus: a run of failed backups can no longer evict good ones. Added tests for in-flight skipping + the failure case. PTAL.

c3y1huang (commented):
LGTM.

Nit: suggest cleaning up the commit history before merge. There are currently a few commits that fix previous commits or address review comments, which makes the history a bit confusing.

issmirnov:
> LGTM.
> 
> Nit: suggest cleaning up the commit history before merge. There are currently a few commits that fix previous commits or address review comments, which makes the history a bit confusing.

Thank you

Would you like to squash rebase on merge, or have a single commit sent up that rebases history on this branch?

derekbit:
> > LGTM.
> > Nit: suggest cleaning up the commit history before merge. There are currently a few commits that fix previous commits or address review comments, which makes the history a bit confusing.
> 
> Thank you
> 
> Would you like to squash rebase on merge, or have a single commit sent up that rebases history on this branch?

Hello @issmirnov 
How about a single commit or few commits? Thank you.

issmirnov:
Squashed down to a single commit as suggested — history's cleaned up. PTAL, thanks!

derekbit (approved):
LGTM

derekbit:
@mergify backport v1.11.x v1.12.x

c3y1huang (approved):

PR #4812, opened 2026-05-27
fix(v2): gate engine frontend switchover on volume SwitchoverState

derekbit:

Issue Longhorn/longhorn#13215

Previously the EngineFrontend controller would trigger an ANA multipath switchover on any target IP/port change, including benign engine restarts on the same node. This could interfere with in-flight operations such as volume expansion on old-engine volumes after an upgrade.

Now the controller only executes the explicit ANA switchover when the volume controller has signaled intent via a non-empty Volume.Status.SwitchoverState. Target changes without switchover intent (e.g. engine pod restart) are left to NVMe multipath auto-reconnection.

An already in-progress switchover (SwitchoverPhase != None) continues unconditionally without requiring a volume lookup.

### DIFF
--- a/controller/engine_frontend_controller.go
+++ b/controller/engine_frontend_controller.go
@@ -430,7 +430,23 @@ func (efc *EngineFrontendController) syncEngineFrontend(key string) (err error) 
 		targetChanged := ef.Status.TargetIP != ef.Spec.TargetIP || ef.Status.TargetPort != ef.Spec.TargetPort
 		switchoverInProgress := ef.Status.SwitchoverPhase != longhorn.EngineFrontendSwitchoverPhaseNone
 
-		if targetChanged || switchoverInProgress {
+		// Only execute the ANA multipath switchover when the volume controller
+		// has explicitly signaled switchover intent (SwitchoverState != Empty).
+		// A target IP change without switchover intent (e.g. engine restart on
+		// the same node) is handled by NVMe multipath auto-reconnection and
+		// does not require an explicit ANA state transition.
+		var volume *longhorn.Volume
+		if targetChanged && !switchoverInProgress {
+			volume, err = efc.ds.GetVolumeRO(ef.Spec.VolumeName)
+			if err != nil && !datastore.ErrorIsNotFound(err) {
+				return errors.Wrapf(err, "failed to get volume %v for engine frontend switchover evaluation", ef.Spec.VolumeName)
+			}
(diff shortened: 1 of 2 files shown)


derekbit:
https://ci.longhorn.io/job/private/job/longhorn-e2e-test/8112/
https://ci.longhorn.io/job/private/job/longhorn-e2e-test/8113/

controller/engine_frontend_controller.go
```diff
@@ -487,6 +503,9 @@ func (efc *EngineFrontendController) syncEngineFrontend(key string) (err error)
 				log.Infof("Successfully switched over target for engine frontend %v to %v", ef.Name, targetAddress)
 				efc.eventRecorder.Eventf(ef, corev1.EventTypeNormal, constant.EventReasonSwitchover, "Successfully switched over target to %v", targetAddress)
 			}
+		} else if targetChanged {
+			log.Debugf("Target changed (spec=%v:%v status=%v:%v) but switchover not executed (volume SwitchoverState is empty), waiting for data-plane path convergence",
+				ef.Spec.TargetIP, ef.Spec.TargetPort, ef.Status.TargetIP, ef.Status.TargetPort)
 		}
```
davidcheng0922, line 509: If the data plane can guarantee target convergence after an IM restart, this LGTM. Otherwise, it may get stuck reconnecting to the stale target.

davidcheng0922 (dismissed):

controller/engine_frontend_controller.go
derekbit: The data plane does not guarantee autonomous target convergence. The IM has no path health monitoring or auto-reconnect to a new IP. However, this code path is effectively unreachable under normal IM restart scenarios.

When the IM pod restarts, the engine goes through an `Error` state. The volume controller detects this and transitions the volume to Faulted. The recovery flow then:

Detaches the volume -> 
stops the EF (closeVolumeDependentResources, DesireState = Stopped) ->
EF instance is deleted (NVMe initiator disconnected) ->
Reattaches ->
EF is recreated with a fresh nvme connect to the new target IP
The EF is fully torn down and rebuilt during recovery.

So, no switchover or data-plane convergence is needed.

davidcheng0922: Clear, thank you!

davidcheng0922 (approved):

derekbit:
@mergify backport v1.12.x

PR #4815, opened 2026-05-28
fix(v2): clear stale EF switchover phase when volume SwitchoverState is reset

derekbit:

Issue Longhorn/longhorn#13215

After auto-salvage or fault recovery the volume controller clears Volume.Status.SwitchoverState, but the EngineFrontend's SwitchoverPhase may remain stuck in a non-None state (e.g. Preparing or Switching). Because shouldExecuteEngineFrontendSwitchover previously returned true unconditionally for any in-progress phase, the EF controller would keep attempting a switchover against an engine that no longer exists, preventing the volume from completing reattachment.

### DIFF
--- a/controller/engine_frontend_controller.go
+++ b/controller/engine_frontend_controller.go
@@ -435,8 +435,13 @@ func (efc *EngineFrontendController) syncEngineFrontend(key string) (err error) 
 		// A target IP change without switchover intent (e.g. engine restart on
 		// the same node) is handled by NVMe multipath auto-reconnection and
 		// does not require an explicit ANA state transition.
+		//
+		// Also fetch the volume when a switchover is already in progress to
+		// validate that the volume controller still wants the switchover.
+		// After auto-salvage or recovery the volume clears SwitchoverState
+		// but the EF's SwitchoverPhase may remain stale.
 		var volume *longhorn.Volume
-		if targetChanged && !switchoverInProgress {
+		if targetChanged || switchoverInProgress {
 			volume, err = efc.ds.GetVolumeRO(ef.Spec.VolumeName)
 			if err != nil && !datastore.ErrorIsNotFound(err) {
 				return errors.Wrapf(err, "failed to get volume %v for engine frontend switchover evaluation", ef.Spec.VolumeName)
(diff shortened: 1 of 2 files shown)


controller/engine_frontend_controller.go
```diff
@@ -503,6 +508,14 @@ func (efc *EngineFrontendController) syncEngineFrontend(key string) (err error)
 				log.Infof("Successfully switched over target for engine frontend %v to %v", ef.Name, targetAddress)
 				efc.eventRecorder.Eventf(ef, corev1.EventTypeNormal, constant.EventReasonSwitchover, "Successfully switched over target to %v", targetAddress)
 			}
+		} else if switchoverInProgress {
+			// shouldExecuteEngineFrontendSwitchover returned false while a
+			// switchover phase was still set. This means the volume no longer
+			// requires the switchover (e.g., after auto-salvage cleared
+			// SwitchoverState). Clear the stale phase.
+			log.Warnf("Clearing stale switchover phase %v: volume SwitchoverState is empty, switchover no longer required",
+				ef.Status.SwitchoverPhase)
```
davidcheng0922, line 517: NIT: `shouldExecuteEngineFrontendSwitchover` also return false when volume is nil. Maybe add the information here for logging more accurate.

controller/engine_frontend_controller.go
```diff
@@ -503,6 +508,14 @@ func (efc *EngineFrontendController) syncEngineFrontend(key string) (err error)
...
+			ef.Status.SwitchoverPhase = longhorn.EngineFrontendSwitchoverPhaseNone
```
davidcheng0922, line 518: Question: when we clear the switchover state here, is it possible that the data-plane switchover is still in the middle of switching? If so, do we need explicit ANA state cleanup/rollback, or is it safe to rely on the current multipath behavior here?

This may be an over-concern here

derekbit: IMO, clearing `SwitchoverPhase` without explicit ANA rollback is safe because this code path only triggers after the volume has already faulted and been auto-salvaged.
The old engine is already destroyed. You cannot roll back ANA state on a dead target. Furthermore, the subsequent recovery flow will fully tear down the EF instance (stop the NVMe initiator and disconnect all paths), then recreate it fresh with correct ANA state.

controller/engine_frontend_controller.go
derekbit: Added. Thank you.

controller/engine_frontend_controller.go
davidcheng0922: Fair point, no further concern from me for now.

derekbit:
@mergify backport v1.12.x

PR #4825, opened 2026-06-01
fix(setting): update DataEngineCPUMask description and default to recommend 2+ cores

derekbit:

Issue Longhorn/longhorn#13237

The previous description did not explain why multiple CPU cores matter for the SPDK target daemon. With a single core, the master reactor handles both I/O polling and management RPCs, so heavy I/O workloads can starve RPC processing and cause latency spikes, timeouts, and operational instability.

### DIFF
--- a/types/setting.go
+++ b/types/setting.go
@@ -1791,13 +1791,13 @@ var (
 
 	SettingDefinitionDataEngineCPUMask = SettingDefinition{
 		DisplayName:        "Data Engine CPU Mask",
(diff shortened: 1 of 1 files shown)


mantissahz (approved):

derekbit:
@mergify backport v1.12.x

innobead (approved):

innobead:
cc @tserong @WebberHuang1118 @Vicente-Cheng

PR #4834, opened 2026-06-03
feat: add setting to gate live engine upgrade on image tag change

derekbit:

Issue #

Commit a6b529e19 introduced support for live engine upgrades between engine images that share the same git commit but differ in image tag (e.g. 1.10.2 -> 1.10.2-4.12). This behavior is now gated behind a new setting "allow-live-engine-upgrade-on-image-tag-change" (default: false).

When disabled, the system retains the original behavior: same-commit images are treated as identical and upgrades are delayed until detach. When enabled, tag-only changes on matching version patterns are allowed to proceed as live upgrades.

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -4150,8 +4150,15 @@ func (c *VolumeController) checkOldAndNewEngineImagesForLiveUpgrade(v *longhorn.
 		return errors.Wrapf(err, "engine live upgrade to %v, but the image wasn't ready", newImage.Spec.Image)
 	}
 
-	if oldImage.Status.GitCommit == newImage.Status.GitCommit && !isRevisionedEngineImage(newImage.Spec.Image) {
-		return fmt.Errorf("engine image %v and %v are identical, delay upgrade until detach for volume", oldImage.Spec.Image, newImage.Spec.Image)
+	if oldImage.Status.GitCommit == newImage.Status.GitCommit {
+		allowSameCommitUpgrade, err := c.ds.GetSettingAsBool(types.SettingNameAllowLiveEngineUpgradeOnSameImageCommit)
+		if err != nil {
+			return errors.Wrapf(err, "failed to get setting %v", types.SettingNameAllowLiveEngineUpgradeOnSameImageCommit)
+		}
+		if !allowSameCommitUpgrade || !isRevisionedEngineImage(newImage.Spec.Image) {
+			return fmt.Errorf("engine images %v and %v share the same git commit (setting %v=%v, revisioned=%v), delay upgrade until detach for volume",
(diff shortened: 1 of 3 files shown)


derekbit:
@megify backport v1.12.x

derekbit:
@mergify backport v1.9.x v1.10.x v1.11.x v1.12.x

types/setting.go
```diff
@@ -180,6 +180,7 @@ const (
 	SettingNameNodeDiskHealthMonitoring                                 = SettingName("node-disk-health-monitoring")
 	SettingNameCSIAllowedTopologyKeys                                   = SettingName("csi-allowed-topology-keys")
 	SettingNameCSIStorageCapacityTracking                               = SettingName("csi-storage-capacity-tracking")
+	SettingNameAllowLiveEngineUpgradeOnImageTagChange                   = SettingName("allow-live-engine-upgrade-on-image-tag-change")
```
innobead: Suggest `SettingNameAllowLiveEngineUpgradeOnSameImageCommit`.

In the normal case, if tags differ, the live engine upgrade should always be enabled when `SettingNameConcurrentAutomaticEngineUpgradePerNodeLimit` is set to a value greater than 0. The original naming might cause confusion, leading people to think the live upgrade can be enabled when the tags differ, regardless of whether ' SettingNameConcurrentAutomaticEngineUpgradePerNodeLimit` is set.

controller/volume_controller.go
```diff
@@ -4150,8 +4150,14 @@ func (c *VolumeController) checkOldAndNewEngineImagesForLiveUpgrade(v *longhorn.
...
 
-	if oldImage.Status.GitCommit == newImage.Status.GitCommit && !isRevisionedEngineImage(newImage.Spec.Image) {
-		return fmt.Errorf("engine image %v and %v are identical, delay upgrade until detach for volume", oldImage.Spec.Image, newImage.Spec.Image)
+	if oldImage.Status.GitCommit == newImage.Status.GitCommit {
+		allowImageTagChangeUpgrade, err := c.ds.GetSettingAsBool(types.SettingNameAllowLiveEngineUpgradeOnImageTagChange)
+		if err != nil {
+			return errors.Wrapf(err, "failed to get setting %v", types.SettingNameAllowLiveEngineUpgradeOnImageTagChange)
+		}
+		if !allowImageTagChangeUpgrade || !isRevisionedEngineImage(newImage.Spec.Image) {
+			return fmt.Errorf("engine images %v and %v share the same git commit, delay upgrade until detach for volume", oldImage.Spec.Image, newImage.Spec.Image)
```
innobead: nit: Suggest adding the setting value to the log to understand why the upgrade is decided to be skipped.

innobead (changes requested):

types/setting.go
derekbit: Renamed to make it clear. Thank you.

controller/volume_controller.go
derekbit: Updated the message. Thank you.

innobead (approved):

PR #4835, opened 2026-06-04
fix(crypto): create v2 encrypted volumes

mantissahz:

Issue # longhorn/longhorn#13163

Normally create/expand/rebuild v2 encrypted volumes

E2E test for v2 volume:
https://ci.longhorn.io/job/private/job/longhorn-e2e-test/8193/


mantissahz:
@Mergifyio backport v1.12.x

controller/engine_frontend_controller.go
```diff
@@ -1100,6 +1109,9 @@ func (m *EngineFrontendMonitor) refresh(ef *longhorn.EngineFrontend) (err error)
 			"requestedSize":     ef.Spec.Size,
 			"expansionRequired": volume.Status.ExpansionRequired,
 		}).Trace("Skip engine frontend expansion because volume expansion is not required")
+		if volume.Spec.Encrypted {
+			ef.Status.CurrentSize = ef.Spec.Size
+		}
```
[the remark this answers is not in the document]
mantissahz: Modified. Thanks.

derekbit:
@mantissahz How do we handle the compatibility of the existing v2 encrypted volumes without extra 16-MiB header?

webhook/resources/volume/mutator.go
```diff
@@ -195,6 +197,10 @@ func (v *volumeMutator) Create(request *admission.Request, newObj runtime.Object
 			return nil, werror.NewInvalidError(fmt.Sprintf("get invalid size for volume %v: %v", currentBackupVolumeSize, err), "")
 		}
 
+		if types.IsDataEngineV2(volume.Spec.DataEngine) && volume.Spec.Encrypted {
+			size = size - lhtypes.Luks2EncryptionHeaderSize
+		}
```
derekbit: What don't we need this for v1 volume?

controller/engine_controller.go
```diff
@@ -1808,6 +1816,14 @@ func checkSizeBeforeRestoration(log logrus.FieldLogger, engine *longhorn.Engine,
 	// actual current size has also reached the backup volume size.
 	if types.IsDataEngineV2(engine.Spec.DataEngine) {
 		if engine.Status.CurrentSize < bvSize {
+			volume, err := ds.GetVolumeRO(engine.Spec.VolumeName)
+			if err != nil {
+				return false, err
+			}
+			if volume.Spec.Encrypted && engine.Status.CurrentSize == bvSize-lhtypes.Luks2EncryptionHeaderSize {
```
derekbit: Suggest adding a helper function for `bvSize-lhtypes.Luks2EncryptionHeaderSize`

webhook/resources/volume/mutator.go
mantissahz: Yes, we need to handle it with backups of v1 encrypted volumes.
Previously, we might consider that v2 encrypted volumes will be as officially starts from v1.12.1, so I will only handle it with v2 encrypted volumes.
But yes, we should handle it generally.

mantissahz:
> @mantissahz How do we handle the compatibility of the existing v2 encrypted volumes without extra 16-MiB header?

IIRC, we do not handle as we discussed. Here we just handle the new created v2 encrypted volumes.
If we want to handle existing volumes, it will be more complicated to expand the underlying SPDK lvol, that will be like what we discarded to expand the v2 volume before upgrading or what v1 does, checking the lvol size and expanding it when volume attaching.

derekbit:
> > @mantissahz How do we handle the compatibility of the existing v2 encrypted volumes without extra 16-MiB header?
> 
> IIRC, we do not handle as we discussed. Here we just handle the new created v2 encrypted volumes. If we want to handle existing volumes, it will be more complicated to expand the underlying SPDK lvol, that will be like what we discarded to expand the v2 volume before upgrading or what v1 does, checking the lvol size and expanding it when volume attaching.

I don't mean adding 16MiB to existing volumes, and what I mean is the compatibility.

As we discussed and emphasized many times before, we still need to handle the compatibility. Our conclusion is that **existing V2 encrypted volumes do not need the additional 16 MiB**. However, we must ensure that both existing and newly created V2 encrypted volumes continue to work (attachment, detachment, backup, restore, dr volume, rebuilding, and so on) correctly for all operations. 

Please make sure this works and is fully verified, and that the test coverage includes these scenarios.

webhook/resources/volume/mutator.go
mantissahz: Removed this mutation.
We will mutate the volume size at the Backup CR.
And v1 and v2 will have the same behavior.

controller/engine_controller.go
mantissahz: Removed this check statement (unnecessary).
Thanks.

mantissahz:
> I don't mean adding 16MiB to existing volumes, and what I mean is the compatibility.
> As we discussed and emphasized many times before, we still need to handle the compatibility. Our conclusion is that existing V2 encrypted volumes do not need the additional 16 MiB. However, we must ensure that both existing and newly created V2 encrypted volumes continue to work (attachment, detachment, backup, restore, dr volume, rebuilding, and so on) correctly for all operations.
> Please make sure this works and is fully verified, and that the test coverage includes these scenarios.

The existing v2 encrypted volumes will not be affected by the new implementation.
The existing v2 encrypted volumes will not be extended by 16 MiB for the LUKS header overhead.
Restoring backups of existing v2 encrypted volumes will still behave as before.

Rebuilding replicas, expanding volumes, and backup/restore operations work as expected.
Only new creating v2 encrypted volumes and their backups/restores after/include this version (v1.12.1 and newer) will be extended by 16 MiB overhead.

derekbit:
> > I don't mean adding 16MiB to existing volumes, and what I mean is the compatibility.
> > As we discussed and emphasized many times before, we still need to handle the compatibility. Our conclusion is that existing V2 encrypted volumes do not need the additional 16 MiB. However, we must ensure that both existing and newly created V2 encrypted volumes continue to work (attachment, detachment, backup, restore, dr volume, rebuilding, and so on) correctly for all operations.
> > Please make sure this works and is fully verified, and that the test coverage includes these scenarios.
> 
> The existing v2 encrypted volumes will not be affected by the new implementation. The existing v2 encrypted volumes will not be extended by 16 MiB for the LUKS header overhead. Restoring backups of existing v2 encrypted volumes will still behave as before.
> 
> Rebuilding replicas, expanding volumes, and backup/restore operations work as expected. Only new creating v2 encrypted volumes and their backups/restores after/include this version (v1.12.1 and newer) will be extended by 16 MiB overhead.

@mantissahz Could you test the existing and new v2 encrypted DR volume as well? Thank you.

webhook/resources/volume/mutator.go
```diff
@@ -147,6 +147,23 @@ func (v *volumeMutator) Create(request *admission.Request, newObj runtime.Object
 	}
 
 	moreLabels := map[string]string{}
+
+	// Use a new label to indicate that the volume is a v2 encrypted volume and LUKS2 header size can be extended.
+	// Existing v2 encrypted volumes that are created before (v1.12.1) the introduction of this label will not have this label.
+	if volume.Spec.Encrypted && types.IsDataEngineV2(volume.Spec.DataEngine) {
+		cureentLonghornVersion, err := v.ds.GetSettingValueExisted(types.SettingNameCurrentLonghornVersion)
```
c3y1huang:
```suggestion
		currentLonghornVersion, err := v.ds.GetSettingValueExisted(types.SettingNameCurrentLonghornVersion)
```

webhook/resources/volume/mutator.go
```diff
@@ -147,6 +147,23 @@ func (v *volumeMutator) Create(request *admission.Request, newObj runtime.Object
...
+		if err != nil {
+			return nil, werror.NewInvalidError(fmt.Sprintf("failed to get current longhorn version: %v", err), "")
+		}
+		isGreater, err := util.IsLonghornVersionGreaterThan(cureentLonghornVersion, util.LastV2EngineImageVerNotSupportExtendLuks2Header)
+		if err != nil {
+			return nil, werror.NewInvalidError(fmt.Sprintf("failed to compare longhorn versions: %v", err), "")
+		}
+		if isGreater {
+			moreLabels[types.LonghornLabelV2VolumeLuksHeaderExtend] = longhorn.TrueValue
```
c3y1huang: IIUC, this label stays fixed for the volume's lifetime? If true, should we add a guard to make it immutable?

controller/engine_frontend_controller.go
```diff
@@ -1100,6 +1108,13 @@ func (m *EngineFrontendMonitor) refresh(ef *longhorn.EngineFrontend) (err error)
 			"requestedSize":     ef.Spec.Size,
 			"expansionRequired": volume.Status.ExpansionRequired,
 		}).Trace("Skip engine frontend expansion because volume expansion is not required")
+		if volume.Spec.Encrypted {
+			desiredSize := ef.Spec.Size
+			if desiredSize == 0 {
+				desiredSize = ef.Spec.VolumeSize
+			}
+			ef.Status.CurrentSize = desiredSize
+		}
```
c3y1huang: Seems this overwrites the `.CurrentSize` unconditionally when the volume is encrypted. IIUC, the volume controller treats `.CurrentSize` as the observed frontend size and gates expansion completion on it (https://github.com/longhorn/longhorn-manager/blob/bc1d7ff798c791552777f7408d2b7b0365f22686/controller/volume_controller.go#L1993-L2019). Is it possible that this gets updated while the expansion is still in flight?

types/volumes.go
```diff
@@ -53,3 +53,23 @@ func IsVolumeReady(v *longhorn.Volume, vrs []*longhorn.Replica, volOp string) (r
...
+		return false
+	}
+	return GetVolumeLuksHeaderExtendLabel(v) == longhorn.TrueValue
+}
+
+// GetVolumeLuksHeaderExtendLabel returns the value of the v2-volume-luks-header-extend label for the given volume.
+func GetVolumeLuksHeaderExtendLabel(v *longhorn.Volume) string {
+	if v.Labels == nil {
+		return ""
+	}
```
c3y1huang: Any reason for duplicating the check in `IsVolumeLuksHeaderExtendLabelTrue`?

controller/engine_controller.go
```diff
@@ -1262,6 +1263,9 @@ func (m *EngineMonitor) refresh(engine *longhorn.Engine) error {
 	removeInvalidEngineOpStatus(engine)
 
 	// align 'engine.Status.CurrentSize' to 'engine.Spec.VolumeSize' if the backend size is expected.
+	if types.IsDataEngineV2(engine.Spec.DataEngine) && volume.Spec.Encrypted && cliAPIVersion == 0 {
+		cliAPIVersion = lhtypes.CliAPIVersionForSupportingExtendLuks2HeaderSize
+	}
```
c3y1huang: Should this check the extended label instead of `cliAPIVersion == 0`? Is it possible that legacy unlabeled volumes also get the +16 MiB expectation here?

types/types.go
```diff
@@ -211,6 +211,7 @@ const (
 	LonghornLabelInstanceManagerImage       = "instance-manager-image"
 	LonghornLabelVolume                     = "longhornvolume"
 	LonghornLabelVolumeEncrypted            = "volume-encrypted"
+	LonghornLabelV2VolumeLuksHeaderExtend   = "v2-volume-luks-header-extend"
```
davidcheng0922: This label actually affects how the volume is created and operated. Should we make it immutable once an encrypted volume is created to avoid accidental changes?

util/util.go
```diff
@@ -646,6 +656,39 @@ func IsKubernetesVersionAtLeast(kubeClient clientset.Interface, vers string) (bo
...
+	if longhornVersion == masterHeadImageTag {
+		return true, nil
+	}
+
+	if matches := engineImageBranchHeadVersionPattern.FindStringSubmatch(longhornVersion); matches != nil {
+		longhornVersion = "v" + matches[1] + ".999999"
+	}
+
+	return isVersionGreaterThan(longhornVersion, targetVersion)
+}
```
derekbit: Add a unit test case for the function.

webhook/resources/volume/mutator.go
mantissahz: Modified. Thanks.

webhook/resources/volume/mutator.go
mantissahz: Yes, added a validation for immutable in the volume validator when updating volume.
Thanks.

types/volumes.go
mantissahz: Removed. Thanks.

controller/engine_controller.go
mantissahz: Modified. Thanks.

types/types.go
mantissahz: Yes, added a validation for immutable in the volume validator when updating volume.
Thanks.

util/util.go
mantissahz: Added. Thanks.

controller/engine_frontend_controller.go
mantissahz: Here, the expansion is completed, and it has deleted the expansionBackoff entry.
The expansion should not be in flight.

types/types.go
derekbit:
```suggestion
	LonghornLabelV2EncryptedVolumeWithLuksHeader   = "v2-encrypted-volume-with-luks-header"
```

mantissahz: Modified. Thanks.

util/util_test.go
```diff
@@ -362,3 +362,120 @@ func TestTimestampAfterTimestamp(t *testing.T) {
...
+		"Current version with pre-release 2": {
+			currentVersion: "v1.2.4-distro",
+			minimumVersion: "v1.2.3",
+			expected:       true,
+		},
+		"Minimum version with pre-release 1": {
+			currentVersion: "v1.2.3",
+			minimumVersion: "v1.2.3-distro",
+			expected:       true,
+		},
```
derekbit: True?

types/volumes.go
```diff
@@ -53,3 +53,17 @@ func IsVolumeReady(v *longhorn.Volume, vrs []*longhorn.Replica, volOp string) (r
...
+	if v.Labels == nil {
+		return false
+	}
+
+	luksExtendedLabel, ok := v.Labels[LonghornLabelV2EncryptedVolumeWithLuksHeader]
+	if !ok {
+		return false
+	}
+
+	return luksExtendedLabel == longhorn.TrueValue
```
derekbit, line 68: When is the label false?

types/volumes.go
```diff
@@ -53,3 +53,17 @@ func IsVolumeReady(v *longhorn.Volume, vrs []*longhorn.Replica, volOp string) (r
 	}
 	return true, ""
 }
+
+// IsVolumeLuksHeaderExtendLabelTrue returns true if the volume has the label "v2-encrypted-volume-with-luks-header" set to "true".
+func IsVolumeLuksHeaderExtendLabelTrue(v *longhorn.Volume) bool {
```
derekbit: Need to update the function name because of the change of the label.
In addition, should we check of the volume is v2?

webhook/resources/volume/validator.go
```diff
@@ -354,6 +354,23 @@ func (v *volumeValidator) Update(request *admission.Request, oldObj runtime.Obje
...
+			if oldVolume.Labels != nil {
+				oldVolumeLuksHeaderExtend = oldVolume.Labels[types.LonghornLabelV2EncryptedVolumeWithLuksHeader]
+			}
+			if newVolume.Labels != nil {
+				newVolumeLuksHeaderExtend = newVolume.Labels[types.LonghornLabelV2EncryptedVolumeWithLuksHeader]
+			}
+			if oldVolumeLuksHeaderExtend != newVolumeLuksHeaderExtend {
+				err := fmt.Errorf("changing %v label for volume %v is not supported", types.LonghornLabelV2EncryptedVolumeWithLuksHeader, oldVolume.Name)
+				return werror.NewInvalidError(err.Error(), "")
+			}
```
derekbit, line 372: Update the variables to make them easy to understand.

derekbit (commented):
Don't we need to handle restoration?
How to distinguish v2 encrypted volumes with and without the extra header while restoration?

types/volumes.go
mantissahz: It will never be false.

webhook/resources/volume/mutator.go
```diff
@@ -147,6 +147,23 @@ func (v *volumeMutator) Create(request *admission.Request, newObj runtime.Object
...
+		if err != nil {
+			return nil, werror.NewInvalidError(fmt.Sprintf("failed to get current longhorn version: %v", err), "")
+		}
+		isGreater, err := util.IsLonghornVersionGreaterThan(currentLonghornVersion, util.LastV2EngineImageVerNotSupportExtendLuks2Header)
+		if err != nil {
+			return nil, werror.NewInvalidError(fmt.Sprintf("failed to compare longhorn versions: %v", err), "")
+		}
+		if isGreater {
+			moreLabels[types.LonghornLabelV2EncryptedVolumeWithLuksHeader] = longhorn.TrueValue
+		}
```
derekbit: Why do we need to check the version for new volumes?

util/util_test.go
mantissahz: Yes, `v1.2.3-distro` will be a pre-release version like `v1.2.3-rc1`

controller/engine_frontend_controller.go
c3y1huang: I'm slightly confused. IIUC, the flow looks like:
- T0: user requested and expansion to 2N. The controller sets [ExpansionRequired = true](https://github.com/mantissahz/longhorn-manager/blob/09b7cc87367f560bb3393d65a15c5c42ef72090f/controller/volume_controller.go#L3285); the real size is still `N` now.
- T1: Engine starts the expansion.
- T2: Engine frontend monitor runs while expanding inflight:
	- First [read the real frontend size](https://github.com/mantissahz/longhorn-manager/blob/09b7cc87367f560bb3393d65a15c5c42ef72090f/controller/engine_frontend_controller.go#L1070), still `N`
	- Then asks `shouldExpandEngineFrontend` and [returns false because `IsExpanding == true`](https://github.com/mantissahz/longhorn-manager/blob/09b7cc87367f560bb3393d65a15c5c42ef72090f/controller/engine_frontend_controller.go#L1133-L1135).
	- Falls into the else branch and [overwrites `CurrentSize`](https://github.com/mantissahz/longhorn-manager/blob/09b7cc87367f560bb3393d65a15c5c42ef72090f/controller/engine_frontend_controller.go#L1116) with `2N`, but the frontend might still be at `N`.
	- Then [defer updates the `2N` to the cluster](https://github.com/mantissahz/longhorn-manager/blob/09b7cc87367f560bb3393d65a15c5c42ef72090f/controller/engine_frontend_controller.go#L972-L975).
	
Would this cause the status to drift/run-ahead of reality? If so, could this potentially lead to problems when other operations overlap with the expansion, such as if a rebuild is triggered while the expansion is still inflight?

derekbit: cc @davidcheng0922

c3y1huang: How about updating `CurrentSize` only after the frontend has reached the expected size?
```go
expectedFrontendSize := ef.Spec.VolumeSize
if types.IsVolumeLuksHeaderExtendLabelTrue(volume) {
	expectedFrontendSize, err = util.GetActualBackendSize(ef.Spec.VolumeSize, volume.Spec.Encrypted, lhtypes.CliAPIVersionForSupportingExtendLuks2HeaderSize)
	# log and return
}

if volume.Spec.Encrypted && volumeInfo.Size >= expectedFrontendSize {
	ef.Status.CurrentSize = ef.Spec.VolumeSize
}
```

derekbit: Looks good!

mantissahz: > How about updating CurrentSize only after the frontend has reached the expected size?

Agreed.

Modified. Thanks.

util/util_test.go
mantissahz: Removed the unit test. Thanks.

util/util.go
```diff
@@ -646,6 +656,39 @@ func IsKubernetesVersionAtLeast(kubeClient clientset.Interface, vers string) (bo
...
+
+// isVersionGreaterThan checks if the current version is greater than the minimum version.
+func isVersionGreaterThan(currentVersion, minimumVersion string) (bool, error) {
+	parsedVer, err := version.ParseSemantic(currentVersion)
+	if err != nil {
+		return false, errors.Wrapf(err, "failed to parse testing version %q", currentVersion)
+	}
```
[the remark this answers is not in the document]
mantissahz: Modified. Thanks.

types/volumes.go
```diff
@@ -53,3 +53,17 @@ func IsVolumeReady(v *longhorn.Volume, vrs []*longhorn.Replica, volOp string) (r
 	}
 	return true, ""
 }
+
+// IsVolumeLuksHeaderExtendLabelTrue returns true if the volume has the label "v2-encrypted-volume-with-luks-header" set to "true".
```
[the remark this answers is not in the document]
mantissahz: Modified. Thanks.

webhook/resources/volume/mutator.go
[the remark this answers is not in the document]
mantissahz: Removed the version test. Thanks.

webhook/resources/volume/mutator.go
```diff
@@ -194,6 +211,14 @@ func (v *volumeMutator) Create(request *admission.Request, newObj runtime.Object
 			}
 		}
 
+		if backup == nil || backup.Status.Labels == nil {
+			delete(moreLabels, types.LonghornLabelV2EncryptedVolumeWithLuksHeader)
+		} else {
+			if encrypted, exists := backup.Status.Labels[types.LonghornLabelVolumeEncrypted]; !exists || encrypted != types.LonghornLabelValueEnabled {
+				delete(moreLabels, types.LonghornLabelV2EncryptedVolumeWithLuksHeader)
+			}
+		}
```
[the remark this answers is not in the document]
mantissahz: Modified. Thanks.

types/volumes.go
mantissahz: > In addition, should we check of the volume is v2? 

Yes. Modified. Thanks.

controller/engine_frontend_controller.go
davidcheng0922: I think the confusing part is the meaning of `shouldExpandEngineFrontend()`.

It returns `false` in at least two different cases:

1. The expansion has completed, so no further expansion is needed.
2. The expansion is still in progress, so another expansion should not be started.

However, both cases fall into the same `else` branch, which overwrites `ef.Status.CurrentSize` with the desired size. If the second case occurs, that run-ahead value may be persisted to the cluster and consumed by other controllers, even though the actual frontend size has not reached it yet.

webhook/resources/volume/validator.go
mantissahz: Updated. Thanks.

webhook/resources/volume/mutator.go
mantissahz: Removed the version comparison. Thanks.

mantissahz:
> Don't we need to handle restoration?
How to distinguish v2 encrypted volumes with and without the extra header while restoration?

It has been handled during the creation of the volume by `util.GetActualBackendSize`.
The creation volume process will preallocate the correct backend size and start the restore process.

davidcheng0922 (commented):
Overall LGTM

controller/engine_frontend_controller.go
```diff
@@ -1071,13 +1079,13 @@ func (m *EngineFrontendMonitor) refresh(ef *longhorn.EngineFrontend) (err error)
...
-			}
-			expectedExpansionSize, err := util.GetActualBackendSize(ef.Spec.VolumeSize, volume.Spec.Encrypted, cliAPIVersion)
-			if err != nil {
-				return err
+
+			expectedExpansionSize := ef.Spec.VolumeSize
+			if types.IsVolumeV2EncryptedVolumeWithLuksHeaderLabelTrue(volume) {
+				expectedExpansionSize, err = util.GetActualBackendSize(ef.Spec.VolumeSize, volume.Spec.Encrypted, lhtypes.CliAPIVersionForSupportingExtendLuks2HeaderSize)
+				if err != nil {
+					return err
```
c3y1huang:
```suggestion
					return errors.Wrapf(err, "failed to get actual backend size for engine frontend %v", ef.Name)
```

controller/engine_frontend_controller.go
```diff
@@ -1091,6 +1099,10 @@ func (m *EngineFrontendMonitor) refresh(ef *longhorn.EngineFrontend) (err error)
 			}
 			// Expansion succeeded; clear backoff.
 			m.expansionBackoff.DeleteEntry(ef.Name)
+
+			if volume.Spec.Encrypted && volumeInfo.Size >= expectedExpansionSize {
+				ef.Status.CurrentSize = ef.Spec.VolumeSize
```
c3y1huang: IIUC, this is unreachable in the `shouldExpandEngineFrontend == true` branch.

https://github.com/longhorn/longhorn-manager/blob/8b8debe286e627955b386785778c7e5fe4b4bdc1/controller/engine_frontend_controller.go#L1136

controller/engine_frontend_controller.go
mantissahz: Modified. Thanks.

controller/engine_frontend_controller.go
mantissahz: Moved this check to
```go
if shouldExpandEngineFrontend(ef, volume, engineForFrontend) {
    ...
} else {
    if volume.Spec.Encrypted && volumeInfo.Size > ef.Spec.VolumeSize && ef.Status.CurrentSize != ef.Spec.VolumeSize {
        ef.Status.CurrentSize = ef.Spec.VolumeSize
    }
}
```

controller/engine_frontend_controller.go
```diff
@@ -1100,6 +1108,9 @@ func (m *EngineFrontendMonitor) refresh(ef *longhorn.EngineFrontend) (err error)
 			"requestedSize":     ef.Spec.Size,
 			"expansionRequired": volume.Status.ExpansionRequired,
 		}).Trace("Skip engine frontend expansion because volume expansion is not required")
+		if volume.Spec.Encrypted && volumeInfo.Size > ef.Spec.VolumeSize && ef.Status.CurrentSize != ef.Spec.VolumeSize {
```
c3y1huang: > `&& ef.Status.CurrentSize != ef.Spec.VolumeSize`

Redundant?

c3y1huang: Suggest adding a doc comment and a line break above.
```
		// An encrypted volume's frontend carries a LUKS header, so the size it reports is
		// larger than the volume's logical size. Report the logical size once the header
		// is there.
```

mantissahz: Modified and added the doc comment. Thanks.

c3y1huang (dismissed):

engineapi/instance_manager.go
```diff
@@ -515,24 +523,27 @@ func (c *InstanceManagerClient) EngineInstanceCreate(req *EngineInstanceCreateRe
...
-	EngineCLIAPIVersion int
-	Encrypted           bool
+	Replica                     *longhorn.Replica
+	DiskName                    string
+	DataPath                    string
+	BackingImagePath            string
+	DataLocality                longhorn.DataLocality
+	EngineCLIAPIVersion         int
+	Encrypted                   bool
+	EncryptedV2HeaderExtendable bool
```
derekbit:
```
ExtraLUKS2HeaderSpaceRequired
```

engineapi/instance_manager.go
```diff
@@ -515,24 +523,27 @@ func (c *InstanceManagerClient) EngineInstanceCreate(req *EngineInstanceCreateRe
...
-	EngineName        string
+	EngineFrontend              *longhorn.EngineFrontend
+	VolumeFrontend              longhorn.VolumeFrontend
+	UblkQueueDepth              int
+	UblkNumberOfQueue           int
+	TargetIP                    string
+	TargetPort                  int
+	EngineName                  string
+	Encrypted                   bool
+	EncryptedV2HeaderExtendable bool
```
derekbit:
```suggestion
	ExtraLUKS2HeaderSpaceRequired bool
```

engineapi/instance_manager.go
```diff
@@ -561,6 +572,14 @@ func (c *InstanceManagerClient) EngineFrontendInstanceCreate(req *EngineFrontend
 		return nil, fmt.Errorf("engine frontend (initiator) requires instance manager API version >= 4")
 	}
 
+	volumeSize := getEngineFrontendInstanceSize(req.EngineFrontend)
+	if req.EncryptedV2HeaderExtendable {
+		volumeSize, err = util.GetActualBackendSize(getEngineFrontendInstanceSize(req.EngineFrontend), req.Encrypted, lhtypes.CliAPIVersionForSupportingExtendLuks2HeaderSize)
```
derekbit:
```suggestion
	if req.EncryptedV2HeaderExtendable {
		volumeSize, err = util.GetActualBackendSize(getEngineFrontendInstanceSize(req.EngineFrontend), req.Encrypted, lhtypes.CliAPIVersionExtraLUKS2HeaderReservation)
```

engineapi/instance_manager.go
mantissahz: Updated. Thanks.

engineapi/instance_manager.go
mantissahz: Updated. Thanks.

engineapi/instance_manager.go
mantissahz: Updated. Thanks.

derekbit (approved):

derekbit:
@mergify backport v1.12.x

PR #4849, opened 2026-06-08
feat(controller): tolerate v2 engine frontend IM unavailability in split topology

davidcheng0922:

https://github.com/longhorn/longhorn/issues/13223

### DIFF
--- a/controller/instance_handler.go
+++ b/controller/instance_handler.go
@@ -49,7 +49,7 @@ func NewInstanceHandler(ds *datastore.DataStore, instanceManagerHandler Instance
 	}
 }
 
-func (h *InstanceHandler) syncStatusWithInstanceManager(log *logrus.Entry, im *longhorn.InstanceManager, instanceName string, spec *longhorn.InstanceSpec, status *longhorn.InstanceStatus, instances map[string]longhorn.InstanceProcess) {
+func (h *InstanceHandler) syncStatusWithInstanceManager(log *logrus.Entry, obj runtime.Object, im *longhorn.InstanceManager, instanceName string, spec *longhorn.InstanceSpec, status *longhorn.InstanceStatus, instances map[string]longhorn.InstanceProcess) {
 	defer func() {
 		if status.CurrentState == longhorn.InstanceStateStopped && !status.Starting {
 			status.InstanceManagerName = ""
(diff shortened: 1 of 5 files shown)


controller/instance_handler.go
```diff
@@ -71,18 +72,21 @@ func (h *InstanceHandler) syncStatusWithInstanceManager(log *logrus.Entry, im *l
...
 		}
-		status.IP = ""
-		status.StorageIP = ""
-		status.Port = 0
-		status.UblkID = 0
-		status.UUID = ""
+		h.resetInstanceConnectionStatusFields(status)
 		h.resetInstanceErrorCondition(status)
 		return
 	}
```
mantissahz, line 78: Could we move `tolerateIMUnavailable := h.shouldTolerateEngineFrontendIMUnavailableInSplitTopology(obj, spec, status)` here?
Because we start using `tolerateIMUnavailable` from here.

controller/instance_handler.go
```diff
@@ -71,18 +72,21 @@ func (h *InstanceHandler) syncStatusWithInstanceManager(log *logrus.Entry, im *l
...
 	if im.Status.CurrentState == longhorn.InstanceManagerStateStopped ||
 		im.Status.CurrentState == longhorn.InstanceManagerStateError ||
 		im.DeletionTimestamp != nil {
+		if tolerateIMUnavailable {
+			status.CurrentState = longhorn.InstanceStateUnknown
+			status.CurrentImage = ""
+			h.resetInstanceConnectionStatusFields(status)
+			h.resetInstanceErrorCondition(status)
+			return
+		}
```
mantissahz:
```go
status.CurrentImage = ""
h.resetInstanceConnectionStatusFields(status)
h.resetInstanceErrorCondition(status)
```
They are also used in the code below.
I'm thinking about a simpler way like

```suggestion
if tolerateIMUnavailable {
  status.CurrentState = longhorn.InstanceStateUnknown
} else {
  if status.Started {
    ...
  } else {
    ...
  }
}
status.CurrentImage = ""
h.resetInstanceConnectionStatusFields(status)
h.resetInstanceErrorCondition(status)
return
```
Is it simpler and cleaner?

controller/instance_handler.go
```diff
@@ -92,34 +96,40 @@ func (h *InstanceHandler) syncStatusWithInstanceManager(log *logrus.Entry, im *l
...
 	}
 
 	if im.Status.CurrentState == longhorn.InstanceManagerStateStarting {
+		if tolerateIMUnavailable {
+			status.CurrentState = longhorn.InstanceStateUnknown
+			status.CurrentImage = ""
+			h.resetInstanceConnectionStatusFields(status)
+			h.resetInstanceErrorCondition(status)
+			return
+		}
```
mantissahz: The same as https://github.com/longhorn/longhorn-manager/pull/4849#discussion_r3542307415

controller/instance_handler.go
```diff
@@ -92,34 +96,40 @@ func (h *InstanceHandler) syncStatusWithInstanceManager(log *logrus.Entry, im *l
...
 
 	instance, exists := instances[instanceName]
 	if !exists {
+		if tolerateIMUnavailable {
+			status.CurrentState = longhorn.InstanceStateUnknown
+			status.CurrentImage = ""
+			h.resetInstanceConnectionStatusFields(status)
+			h.resetInstanceErrorCondition(status)
+			return
+		}
```
mantissahz: The same as https://github.com/longhorn/longhorn-manager/pull/4849#discussion_r3542307415

controller/utils.go
```diff
@@ -289,3 +289,25 @@ func getContainerArgValue(args []string, flag string) string {
...
+	}
+	if targetEngineNodeID == "" || v.Status.CurrentNodeID != nodeID || ef.Spec.NodeID != nodeID {
+		return false
+	}
+
+	// This uses volume status, so CurrentEngineNodeID may briefly lag while
+	// the engine switch is converging.
+	splitTopology := targetEngineNodeID != nodeID ||
+		(v.Status.CurrentEngineNodeID != "" && v.Status.CurrentEngineNodeID != nodeID)
+	return splitTopology
```
mantissahz: Maybe we can return it directly.
```suggestion
	return targetEngineNodeID != nodeID || (v.Status.CurrentEngineNodeID != "" && v.Status.CurrentEngineNodeID != nodeID)
```

or we can have another help function for the check statement.

controller/volume_controller.go
```diff
@@ -4613,6 +4613,30 @@ func isEngineFrontendReadyForNode(efs map[string]*longhorn.EngineFrontend, nodeI
 	return isEngineFrontendReady(ef)
 }
 
+// isEngineFrontendToleratedInSplitTopology returns true when a v2
+// EngineFrontend can still be treated as usable in split topology.
+func isEngineFrontendToleratedInSplitTopology(v *longhorn.Volume, ef *longhorn.EngineFrontend, nodeID string) bool {
```
mantissahz: Could you elaborate on why this function is here instead of in the volume attachment controller?

davidcheng0922: Thank you, move it

controller/utils.go
davidcheng0922: Thank you, update it

controller/instance_handler.go
davidcheng0922: Fix it, thank you

controller/instance_handler.go
davidcheng0922: Fix it, thank you

controller/instance_handler.go
davidcheng0922: Fix it, thank you

controller/instance_handler.go
davidcheng0922: Fix it, thank you

controller/utils.go
```diff
@@ -292,3 +292,23 @@ func getContainerArgValue(args []string, flag string) string {
 	}
 	return ""
 }
+
+// isEngineFrontendInSplitTopology returns true when a v2 EngineFrontend stays
+// on the attachment node while the engine target is on another node.
```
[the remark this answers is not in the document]
davidcheng0922: Update

controller/volume_attachment_controller.go
```diff
@@ -1026,7 +1026,12 @@ func (vac *VolumeAttachmentController) isVolumeAvailableOnNode(volumeName, node
 			return false
 		}
 		efForNode, err = getEngineFrontendForNode(efs, node)
-		if err != nil || !isEngineFrontendReady(efForNode) {
+		if err != nil {
+			return false
+		}
+
+		toleratedInSplitTopology := isEngineFrontendToleratedInSplitTopology(volume, efForNode, node)
```
c3y1huang, line 1033: In what scenario would this return true? Both callers of isVolumeAvailableOnNode pass CurrentMigrationNodeID, but IIUC this requires [v.Status.CurrentNodeID == nodeID](https://github.com/davidcheng0922/longhorn-manager/blob/eec10fe6a77687b38d9799e3fc05123b948d3a7e/controller/utils.go#L308-L310).

controller/volume_controller_test.go
```diff
@@ -2511,6 +2511,45 @@ func (s *TestSuite) TestProcessMigrationV2CreatesMigrationEngineFrontend(c *C) {
...
+	}
+
+	available := isEngineFrontendToleratedInSplitTopology(v, efs["vol-ef-0"], TestNode1)
+	c.Assert(available, Equals, true)
+
+	efs["vol-ef-0"].Status.Endpoint = ""
+	available = isEngineFrontendToleratedInSplitTopology(v, efs["vol-ef-0"], TestNode1)
+	c.Assert(available, Equals, false)
+
+	efs["vol-ef-0"].Status.Endpoint = "/dev/longhorn/vol"
```
c3y1huang, line 2547: This test sets `Endpoint: "/dev/longhorn/vol"` + `CurrentState: Unknown`, is this a valid flow? Given that monitor reset clears the endpoint when EF leaves Running? Why not build the EF state through `ReconcileInstanceState` & `ResetMonitoringEngineFrontendStatus`?

controller/volume_attachment_controller.go
```diff
@@ -1057,6 +1062,30 @@ func (vac *VolumeAttachmentController) isVolumeAvailableOnNode(volumeName, node
...
+		return false
+	}
+
+	// Disabled or empty frontends do not expose a user-visible block device,
+	// so there is no endpoint to validate on the attachment node.
+	if ef.Spec.DisableFrontend || ef.Spec.Frontend == longhorn.VolumeFrontendEmpty {
+		return true
+	}
+
+	return ef.Status.Endpoint != ""
```
c3y1huang, line 1086: IIUC the EF leaves Running -> EF controller stops monitoring and [clears Status.Endpoint](https://github.com/davidcheng0922/longhorn-manager/blob/eec10fe6a77687b38d9799e3fc05123b948d3a7e/datastore/longhorn.go#L2201). So by the time the EF is unknown, the endpoint should already be gone. In what scenario would this check return true?

controller/volume_attachment_controller.go
```diff
@@ -1026,7 +1026,12 @@ func (vac *VolumeAttachmentController) isVolumeAvailableOnNode(volumeName, node
...
 		efForNode, err = getEngineFrontendForNode(efs, node)
-		if err != nil || !isEngineFrontendReady(efForNode) {
+		if err != nil {
+			return false
+		}
+
+		toleratedInSplitTopology := isEngineFrontendToleratedInSplitTopology(volume, efForNode, node)
+		if !isEngineFrontendReady(efForNode) && !toleratedInSplitTopology {
 			return false
 		}
```
c3y1huang, line 1036: No need to check tolerance on healthy path.
```suggestion
		if !isEngineFrontendReady(efForNode) && !isEngineFrontendToleratedInSplitTopology(volume, efForNode, node) {
			return false
		}
```

controller/instance_handler.go
```diff
@@ -306,6 +288,41 @@ func (h *InstanceHandler) syncInstanceCondition(instance longhorn.InstanceProces
...
+
+	ef, ok := obj.(*longhorn.EngineFrontend)
+	if !ok {
+		return false
+	}
+
+	volume, err := h.ds.GetVolumeRO(spec.VolumeName)
+	if err != nil {
+		if !datastore.ErrorIsNotFound(err) {
+			logrus.WithError(err).Warnf("Failed to get volume %v while evaluating engine frontend IM tolerance", spec.VolumeName)
```
c3y1huang, line 306: How about pass the `log` in so it doesn't lose the context?

PR #4851, opened 2026-06-08
fix(migration): prevent initiate volume live migration if target node not ready

COLDTURNIP:

Issue longhorn/longhorn#13109

https://github.com/longhorn/longhorn/issues/13109#issuecomment-4677392901

### DIFF
--- a/controller/volume_attachment_controller.go
+++ b/controller/volume_attachment_controller.go
@@ -454,9 +454,31 @@ func (vac *VolumeAttachmentController) handleVolumeMigrationStart(va *longhorn.V
 	// Found one csi attachmentTicket that is requesting volume to attach to the current node
 
 	if attachmentTicket := getCSIAttachmentTicketNotRequestingNode(vol.Spec.NodeID, va, vol); attachmentTicket != nil {
+		log := getLoggerForMigratingLHVolumeAttachment(vac.logger, va, vol)
+
+		// Require the target node to be fully Ready before starting migration.
+		// If migration is initiated while the target node is still recovering (e.g.,
+		// after a node restart), controllers across the cluster may transiently observe
+		// inconsistent node readiness, causing rapid engine CR ownership flapping.
+		// During that window, the volume controller's cleanup logic can misidentify the
+		// engine actively serving I/O as an "extra" and delete it, causing immediate I/O
+		// failure for the workload. Deferring migration until Ready=True ensures the
+		// node's controllers are stable before a migration engine is created.
+		targetNode, err := vac.ds.GetNodeRO(attachmentTicket.NodeID)
+		if err != nil {
+			log.WithError(err).Warnf("Failed to get node %v for migration readiness check", attachmentTicket.NodeID)
+			vac.enqueueVolumeAttachmentAfter(va, 10*time.Second)
+			return
+		}
+		readyCond := types.GetCondition(targetNode.Status.Conditions, longhorn.NodeConditionTypeReady)
+		if readyCond.Status != longhorn.ConditionStatusTrue {
+			log.Infof("Pending migration attachment ticket %v: target node %v not ready (reason: %v)", attachmentTicket.ID, attachmentTicket.NodeID, readyCond.Reason)
+			vac.enqueueVolumeAttachmentAfter(va, 10*time.Second)
+			return
+		}
+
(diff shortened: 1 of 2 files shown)


controller/volume_attachment_controller.go
```diff
@@ -454,9 +454,29 @@ func (vac *VolumeAttachmentController) handleVolumeMigrationStart(va *longhorn.V
...
+		targetNode, err := vac.ds.GetNodeRO(attachmentTicket.NodeID)
+		if err != nil {
+			log.WithError(err).Warnf("Failed to get node %v for migration readiness check", attachmentTicket.NodeID)
+			return
+		}
+		readyCond := types.GetCondition(targetNode.Status.Conditions, longhorn.NodeConditionTypeReady)
+		if readyCond.Status != longhorn.ConditionStatusTrue {
+			log.Infof("Pending migration attachment ticket %v: target node %v not ready (reason: %v)", attachmentTicket.ID, attachmentTicket.NodeID, readyCond.Reason)
+			return
+		}
```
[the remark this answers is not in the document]
COLDTURNIP: We don't need to requeue this change event. The LHVA should be reconciled again when the node status is updated.

WebberHuang1118 (approved):
LGTM, thanks for the fix.

controller/volume_attachment_controller.go
derekbit, line 478: Can we add enqueueVolumeAttachmentAfter to avoid long wait?

controller/volume_attachment_controller.go
```diff
@@ -454,9 +454,31 @@ func (vac *VolumeAttachmentController) handleVolumeMigrationStart(va *longhorn.V
 	// Found one csi attachmentTicket that is requesting volume to attach to the current node
 
 	if attachmentTicket := getCSIAttachmentTicketNotRequestingNode(vol.Spec.NodeID, va, vol); attachmentTicket != nil {
+		log := getLoggerForMigratingLHVolumeAttachment(vac.logger, va, vol)
+
+		// Require the target node to be fully Ready before starting migration.
+		// If migration is initiated while the target node's instance manager is still
+		// initializing (e.g., after a node restart), the engine controller may fail to
+		// take stable CR ownership, causing rapid ownership flapping between nodes.
```
c3y1huang: > 		// If migration is initiated while the target node's instance manager is still
>		// initializing

The comment is a bit confusing. Does Node Ready==True mean the instance manager has finished initializing?

controller/volume_attachment_controller_test.go
```diff
@@ -560,6 +569,130 @@ func (s *TestSuite) runVolumeAttachmentTestCase(c *C, tc *volumeAttachmentTestCa
...
+				Satisfied: false,
+				Conditions: types.SetConditionWithoutTimestamp([]longhorn.Condition{},
+					longhorn.AttachmentStatusConditionTypeSatisfied, longhorn.ConditionStatusFalse, "",
+					fmt.Sprintf("the volume is currently attached to different node %v ", TestNode1)),
+			},
+		}
+	}
+
+	///////////////////////////////////////////////////////////////////
+	// Case A: target node not found in etcd -> IsNodeDownOrDeletedOrMissingManager=true -> blocked
```
c3y1huang, line 621: The comment references `IsNodeDownOrDeletedOrMissingManager`. Was this called in the path?

COLDTURNIP: Not directly related to the VolumeAttachment controller. `IsNodeDownOrDeletedOrMissingManager` is invoked in `isControllerResponsibleFor`, which is the key method to Engine controller to decide the ownership. When the preferred node is not ready, `isControllerResponsibleFor` may cause the ownership flipping while reconciling the migration engines. Hence, VolumeAttachment controller prevent starting the migration in this ambiguous situation.

controller/volume_attachment_controller.go
COLDTURNIP: Good question. During the node restarting or rejoining, the instance manager readiness needs to be reconciled. Volume controller is responsible to select the target IM to create migration engine. Just reworded this part.

c3y1huang (approved):

derekbit (approved):

derekbit:
@mergify backport v1.11.x v1.12.x

PR #4874, opened 2026-06-09
fix(replica): remove v2 replica from engine RAID before deleting its …

hookak:
…instance

On the v2 data engine, removing a replica from a running volume stalls foreground write I/O for ~10s. The replica controller tears down the replica's NVMe-oF target before the engine removes its base bdev from the RAID, so the orphaned bdev_nvme reconnect-loops and bdev_raid holds I/O until fast_io_fail_timeout_sec (default 10s).

Defer a v2 replica's instance deletion until no running engine (current or migration) still references it, and drop a deleting replica from the engine's replica address map so the engine removes its base bdev first. A 60s timeout bounds the wait. This mirrors the volume-deletion path.

Longhorn 13309

Issue # https://github.com/longhorn/longhorn/issues/13309

### DIFF
--- a/controller/replica_controller.go
+++ b/controller/replica_controller.go
@@ -27,6 +27,7 @@ import (
 	v1core "k8s.io/client-go/kubernetes/typed/core/v1"
 
 	imapi "github.com/longhorn/longhorn-instance-manager/pkg/api"
+	imutil "github.com/longhorn/longhorn-instance-manager/pkg/util"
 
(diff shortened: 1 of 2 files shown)


derekbit (approved):
LGTM

derekbit:
@hookak Thanks for the contribution!

derekbit:
@mergify backport v1.12.x

PR #4895, opened 2026-06-12
fix: align runtime paths with default-data-path

davidcheng0922:

https://github.com/longhorn/longhorn/issues/9083

This PR introduces `default-control-path` to separate Longhorn runtime/control-plane paths from `default-data-path`.

Previously, engine binaries, metadata, unix domain sockets, and logs still relied on `/var/lib/longhorn`, which could cause issues in environments where `/var` is mounted with `noexec`. With this change, those runtime artifacts are resolved from `default-control-path`, while `default-data-path` continues to represent the data path.

This PR also introduces `LONGHORN_CONTROL_PATH` and makes both `default-data-path` and `default-control-path` immutable after Longhorn has been initialized.

- default-data-path and default-control-path are installation-time settings.
- The chart writes them into the default setting ConfigMap and passes them to pods through `LONGHORN_DATA_PATH` and `LONGHORN_CONTROL_PATH`.
- Longhorn processes read these env vars at startup to determine data/runtime path resolution.

### DIFF
--- a/datastore/longhorn.go
+++ b/datastore/longhorn.go
@@ -7,6 +7,7 @@ import (
 	"math/rand"
 	"net"
 	"net/url"
+	"path/filepath"
 	"reflect"
 	"regexp"
 	"strconv"
@@ -660,6 +661,42 @@ func (s *DataStore) ValidateSetting(name, value string) (err error) {
 			_, err = s.ValidateV2DataEngineEnabled(dataEngineEnabled)
 			if err != nil {
 				return err
+			}
+		}
+
+	case types.SettingNameDefaultDataPath:
+		old, err := s.GetSettingWithAutoFillingRO(types.SettingNameDefaultDataPath)
+		if err != nil {
+			return err
+		}
+
+		oldPath := filepath.Clean(strings.TrimSpace(old.Value))
+		newPath := filepath.Clean(strings.TrimSpace(value))
+		if oldPath != newPath {
+			nodes, err := s.ListNodesRO()
+			if err != nil {
+				return err
+			}
+			if len(nodes) != 0 {
+				return errors.Errorf("cannot change %v after Longhorn has been initialized", types.SettingNameDefaultDataPath)
+			}
+		}
+
+	case types.SettingNameDefaultControlPath:
+		old, err := s.GetSettingWithAutoFillingRO(types.SettingNameDefaultControlPath)
+		if err != nil {
+			return err
+		}
+
+		oldPath := filepath.Clean(strings.TrimSpace(old.Value))
+		newPath := filepath.Clean(strings.TrimSpace(value))
+		if oldPath != newPath {
+			nodes, err := s.ListNodesRO()
+			if err != nil {
+				return err
+			}
+			if len(nodes) != 0 {
+				return errors.Errorf("cannot change %v after Longhorn has been initialized", types.SettingNameDefaultControlPath)
 			}
 		}
 
--- a/datastore/longhorn_test.go
+++ b/datastore/longhorn_test.go
@@ -15,6 +15,7 @@ import (
 	apiextensionsfake "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset/fake"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
+	"github.com/longhorn/longhorn-manager/types"
 	"github.com/longhorn/longhorn-manager/util"
 
 	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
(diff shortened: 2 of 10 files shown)


types/types.go
```diff
@@ -402,13 +401,41 @@ func GetDefaultManagerURL() string {
 	return "http://longhorn-backend:" + strconv.Itoa(DefaultAPIPort) + "/v1"
 }
 
+func GetLonghornDataPath() string {
+	path := filepath.Clean(strings.TrimSpace(os.Getenv(LonghornDataPathEnv)))
+	if path == "." || path == "" || !filepath.IsAbs(path) {
+		return DefaultDataPath
+	}
```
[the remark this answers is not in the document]
davidcheng0922: The intended path flow is:

1. Before installation, the user configures `default-data-path`.
2. During chart rendering / pod creation, that value is written into the `default-setting.yaml` ConfigMap.
3. The same value is also propagated to the `LONGHORN_DATA_PATH` environment variable for the relevant Longhorn manager-managed pods/processes.
4. The manager, jobs, and instance manager processes read this environment variable at startup.
5. `GetLonghornDataPath()` then derives the runtime root from that process-scoped value.

So this helper is intentionally env-based rather than reading the Setting dynamically at runtime. This aligns with the current design that path changes are not dynamically supported after initialization.

To make this contract clearer, I added a comment to `GetLonghornDataPath()` documenting that `LONGHORN_DATA_PATH` is expected to be populated from the configured `default-data-path` during installation or pod creation, and that the default path is retained as the backward-compatible fallback.

types/types.go
```diff
@@ -402,13 +401,52 @@ func GetDefaultManagerURL() string {
...
+}
+
+func GetUnixDomainSocketDirectoryOnHost() string {
+	return filepath.Join(GetLonghornDataPath(), UnixDomainSocketDirectorySubpath)
+}
+
+func GetUnixDomainSocketDirectoryInContainer() string {
+	return filepath.Join(ReplicaHostPrefix,
+		strings.TrimPrefix(GetUnixDomainSocketDirectoryOnHost(), string(filepath.Separator)))
+}
```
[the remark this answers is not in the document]
davidcheng0922: I don't think this case applies here because the path is already normalized by `filepath.Clean` in `GetLonghornDataPath()`, and the derived host paths are constructed via `filepath.Join(...)`. On Unix, that yields a single-rooted absolute path (e.g. `/data/longhorn/...`), so trimming a single leading separator is sufficient to make the path relative before joining with `/host`.

types/types.go
```diff
@@ -418,7 +456,11 @@ func GetEngineBinaryDirectoryForEngineManagerContainer(image string) string {
 
 func GetEngineBinaryDirectoryForReplicaManagerContainer(image string) string {
 	cname := GetImageCanonicalName(image)
-	return filepath.Join(filepath.Join(ReplicaHostPrefix, EngineBinaryDirectoryOnHost), cname)
+	return filepath.Join(
+		ReplicaHostPrefix,
+		strings.TrimPrefix(GetEngineBinaryDirectoryOnHost(), string(filepath.Separator)),
+		cname,
+	)
```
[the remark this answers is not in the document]
davidcheng0922: Same above

davidcheng0922:
This is a design tradeoff in the current approach.

When `default-data-path` is a directory, both data-plane and runtime/control-plane paths follow the same root. However, when `default-data-path` is a block-device-style path, only the data-plane semantics apply. In that case, runtime/control-plane paths fall back to `/var/lib/longhorn`, since they require a **directory-based** root for engine binaries, metadata, sockets, and logs.

Examples:

- If `default-data-path` is `/data`:
  - data path: `/data`
  - control/runtime path: `/data`

- If `default-data-path` is `/dev/nvmeXnY`:
  - data path: `/dev/nvmeXnY`
  - control/runtime path: `/var/lib/longhorn`

Another option would be to introduce a separate setting such as `default-control-path`. That would make the behavior more explicit, but I am not sure whether adding another setting would be too redundant or make the configuration more complex and confusion for users.

For now, I kept the current design. If you think introducing a separate control/runtime path setting would be the better direction, I can implement and test that as well. What do you think, @derekbit ?

types/types.go
```diff
@@ -107,16 +107,15 @@ const (
...
-	UnixDomainSocketDirectoryInContainer = "/host/var/lib/longhorn/unix-domain-socket/"
-	UnixDomainSocketDirectoryOnHost      = "/var/lib/longhorn/unix-domain-socket/"
-
-	DefaultLogDirectoryOnHost = "/var/lib/longhorn/logs/"
+	DefaultDataPath                  = "/var/lib/longhorn"
+	LonghornDataPathEnv              = "LONGHORN_DATA_PATH"
+	EngineBinaryDirectorySubpath     = "engine-binaries"
+	MetadataDirectorySubpath         = "metadata"
+	UnixDomainSocketDirectorySubpath = "unix-domain-socket"
+	DefaultLogDirectorySubpath       = "logs"
```
derekbit: Should we remove `Default` to make the naming consistent?

davidcheng0922: Fix it, thank you

derekbit:
@davidcheng0922 CI failed (go fmt error)

cc @COLDTURNIP @carterli0407-cell for reviewing the PR. Thanks.

types/types.go
```diff
@@ -403,13 +402,57 @@ func GetDefaultManagerURL() string {
...
+	if path == "." || path == "" || path == string(filepath.Separator) || !filepath.IsAbs(path) {
+		return DefaultDataPath
+	}
+	// default-data-path may be configured as a block device path for the V2 data
+	// engine, but runtime/control-plane paths require a directory root.
+	if path == "/dev" || strings.HasPrefix(path, "/dev/") {
+		return DefaultDataPath
+	}
+	return path
+}
```
COLDTURNIP, line 448: Suggest to reflect the calculated data path on Node CR. Instead of falling back to `DefaultDataPath` silently when the given path is not a directory, maybe be could have a condition to reflect the calculated `GetLonghornDataPath()` , along with the availability status (e.g. not a writable directory or mounted with the reported flag `noexec`).

davidcheng0922: Re-design this part

types/types.go
```diff
@@ -403,13 +402,57 @@ func GetDefaultManagerURL() string {
...
+func GetLonghornDataPath() string {
+	path := filepath.Clean(strings.TrimSpace(os.Getenv(LonghornDataPathEnv)))
+	if path == "." || path == "" || path == string(filepath.Separator) || !filepath.IsAbs(path) {
+		return DefaultDataPath
+	}
+	// default-data-path may be configured as a block device path for the V2 data
+	// engine, but runtime/control-plane paths require a directory root.
+	if path == "/dev" || strings.HasPrefix(path, "/dev/") {
+		return DefaultDataPath
+	}
```
carterli0407-cell, line 423: Does it work for the V2 Data Engine case?
If it does not fix for those users, we don't need to consider a separate default-control-path setting so V2 users can relocate runtime artifacts off /var independently of the block device?

types/types.go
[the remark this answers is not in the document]
shuo-wu: NIT: I am thinking whether we need a `node.Status` field to record this immutable path after reading for the env. Otherwise, users may have no idea that there is still a special path being used after changing the setting `SettingNameDefaultDataPath` value

shuo-wu (dismissed):
In general, LGTM

davidcheng0922:
The team’s concerns regarding the fallback strategy and potential compatibility issues for v2 users are completely valid.

After discussing this with @derekbit, we agreed that the cleanest and most robust way to fully support both v1 and v2 environments is to introduce a new setting: `default-control-path`. This gives users explicit control over where they configure both their control and data paths.

Therefore, I will re-implement this ticket. Thank you all for the feedback!

cc @COLDTURNIP  @carterli0407-cell  @shuo-wu

types/types.go
```diff
@@ -403,13 +404,94 @@ func GetDefaultManagerURL() string {
...
+
+// Defaults to /var/lib/longhorn/unix-domain-socket when LONGHORN_CONTROL_PATH is unset.
+func GetUnixDomainSocketDirectoryOnHost() string {
+	return filepath.Join(GetLonghornControlPath(), UnixDomainSocketDirectorySubpath)
+}
+
+// Defaults to /host/var/lib/longhorn/unix-domain-socket inside the container.
+func GetUnixDomainSocketDirectoryInContainer() string {
+	return filepath.Join(ReplicaHostPrefix,
+		strings.TrimPrefix(GetUnixDomainSocketDirectoryOnHost(), string(filepath.Separator)))
```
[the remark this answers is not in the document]
davidcheng0922: Fix it.

types/setting.go
```diff
@@ -2851,6 +2875,10 @@ func validateSettingString(name SettingName, definition SettingDefinition, value
 			if _, err := NormalizeCPUMask(strValue); err != nil {
 				return errors.Wrapf(err, "the value of %v is invalid", name)
 			}
+		case SettingNameDefaultDataPath:
+			if !IsValidLonghornDataPath(strValue) {
+				return fmt.Errorf("the value of %v is invalid", name)
+			}
 		}
```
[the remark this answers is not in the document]
davidcheng0922: Fix it.

types/types.go
```diff
@@ -419,7 +501,11 @@ func GetEngineBinaryDirectoryForEngineManagerContainer(image string) string {
 
 func GetEngineBinaryDirectoryForReplicaManagerContainer(image string) string {
 	cname := GetImageCanonicalName(image)
-	return filepath.Join(filepath.Join(ReplicaHostPrefix, EngineBinaryDirectoryOnHost), cname)
+	return filepath.Join(
+		ReplicaHostPrefix,
+		strings.TrimPrefix(GetEngineBinaryDirectoryOnHost(), string(filepath.Separator)),
+		cname,
+	)
```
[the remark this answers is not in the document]
davidcheng0922: Fix it.

types/setting.go
```diff
@@ -550,14 +553,35 @@ var (
...
-		Description:        "Default path to use for storing data on a host. An absolute directory path indicates a filesystem-type disk used by the V1 Data Engine, while a path to a block device indicates a block-type disk used by the V2 Data Engine.",
+		DisplayName: "Default Data Path",
+		Description: "Default path to use for storing data on a host. " +
+			"An absolute directory path indicates a filesystem-type disk used by the V1 Data Engine, " +
+			"while a path to a block device indicates a block-type disk used by the V2 Data Engine. " +
+			"Bare PCI identifiers such as 0000:00:1e.0 are not supported here because this setting may be " +
+			"used as a host path for pod mounts. " +
+			"When this setting is a block device path, runtime and control-plane paths are configured " +
+			"separately by default-control-path. This is an installation-time setting and cannot be changed " +
+			"after Longhorn has been initialized.",
```
derekbit: @sushant-suse Could you review the description? Thanks.

types/setting.go
```diff
@@ -550,14 +553,35 @@ var (
...
+		Description: "Default path to use for storing runtime and control-plane artifacts on a host. " +
+			"This setting must be an absolute directory path. Engine binaries, metadata, sockets, and logs " +
+			"are stored under this path. This is an installation-time setting and cannot be changed after " +
+			"Longhorn has been initialized.",
+		Category:           SettingCategoryGeneral,
+		Type:               SettingTypeString,
+		Required:           true,
+		ReadOnly:           false,
+		DataEngineSpecific: false,
+		Default:            DefaultControlPath,
```
derekbit, line 591: @sushant-suse Could you review the description? Thanks.

datastore/longhorn.go
```diff
@@ -663,6 +664,58 @@ func (s *DataStore) ValidateSetting(name, value string) (err error) {
...
+		}
+
+		oldPath := strings.TrimSpace(old.Value)
+		newPath := strings.TrimSpace(value)
+		if filepath.IsAbs(oldPath) {
+			oldPath = filepath.Clean(oldPath)
+		}
+		if filepath.IsAbs(newPath) {
+			newPath = filepath.Clean(newPath)
+		}
```
shuo-wu: NIT: `types.IsValidLonghornControlPath` already guarantees that the path is always abs path. Is this still necessary?

datastore/longhorn_test.go
```diff
@@ -240,3 +241,168 @@ func TestGetVolumeCurrentEngineFrontendReturnsErrorWhenMissing(t *testing.T) {
...
+
+func TestValidateSettingDefaultControlPath(t *testing.T) {
+	const testNamespace = "longhorn-system"
+
+	baseSetting := &longhorn.Setting{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      string(types.SettingNameDefaultControlPath),
+			Namespace: testNamespace,
+		},
+		Value: "/var/lib/longhorn",
```
shuo-wu: NIT: use const `DefaultControlPath`

types/setting.go
```diff
@@ -550,14 +553,35 @@ var (
...
-		Default:            "/var/lib/longhorn/",
+		Default:            DefaultDataPath,
+	}
+
+	SettingDefinitionDefaultControlPath = SettingDefinition{
+		DisplayName: "Default Control Path",
+		Description: "Default path to use for storing runtime and control-plane artifacts on a host. " +
+			"This setting must be an absolute directory path. Engine binaries, metadata, sockets, and logs " +
+			"are stored under this path. This is an installation-time setting and cannot be changed after " +
+			"Longhorn has been initialized.",
```
shuo-wu: NIT: We can mention that this path is necessary for both v1 and v2 engines

datastore/longhorn.go
```diff
@@ -663,6 +664,58 @@ func (s *DataStore) ValidateSetting(name, value string) (err error) {
...
+		if oldPath != newPath {
+			nodes, err := s.ListNodesRO()
+			if err != nil {
+				return err
+			}
+			if len(nodes) != 0 {
+				return errors.Errorf("cannot change %v after Longhorn has been initialized", types.SettingNameDefaultDataPath)
+			}
```
shuo-wu, line 682: After introducing the control path setting, should this still be immutable?

types/types.go
```diff
@@ -404,13 +405,94 @@ func GetDefaultManagerURL() string {
...
+		return DefaultControlPath
+	}
+	path = filepath.Clean(path)
+	if path == "." || path == "" || path == string(filepath.Separator) || !filepath.IsAbs(path) {
+		logrus.Warnf("Falling back to default control path %q because %s is unset or invalid", DefaultControlPath, LonghornControlPathEnv)
+		return DefaultControlPath
+	}
+	if path == "/dev" || strings.HasPrefix(path, "/dev/") {
+		logrus.Warnf("Falling back to default control path %q because %s cannot point to /dev", DefaultControlPath, LonghornControlPathEnv)
+		return DefaultControlPath
```
carterli0407-cell: `GetLonghornDataPath()` correctly delegates validation to `IsValidLonghornDataPath()`.
But `GetLonghornControlPath()` duplicates the same validation logic inline instead of calling `IsValidLonghornControlPath()`.

datastore/longhorn.go
davidcheng0922: You are right, fix it, thank you

types/setting.go
davidcheng0922: Fix it, thank you

datastore/longhorn_test.go
davidcheng0922: Fx it, thank you

types/types.go
davidcheng0922: You are right, thank you

datastore/longhorn.go
davidcheng0922: If we want to support changing this setting dynamically, we would also need to restart the `instance-manager` pods so they can pick up the new value, since it is passed through the `Env`.

If users need a different path after installation, they can remove the default disk and add a new disk with the desired path instead. In my opinion, it is probably not worth supporting this dynamically. WDYT?

types/setting.go
```suggestion
			"An absolute directory path indicates a filesystem-type disk used by the V1 Data Engine, " +
			"whereas a path to a block device indicates a block-type disk used by the V2 Data Engine. " +
			"Bare PCI identifiers (such as '0000:00:1e.0') are not supported here since this setting may be " +
			"used as a host path for pod mounts. " +
			"When this setting is a block device path, runtime and control-plane paths are configured " +
			"separately via the 'default-control-path' setting. Note: This is an installation-time setting " +
			"and cannot be changed after Longhorn is initialized.",
```

types/setting.go
```diff
@@ -550,14 +553,35 @@ var (
...
-		Default:            "/var/lib/longhorn/",
+		Default:            DefaultDataPath,
+	}
+
+	SettingDefinitionDefaultControlPath = SettingDefinition{
+		DisplayName: "Default Control Path",
+		Description: "Default path to use for storing runtime and control-plane artifacts on a host. " +
+			"This setting must be an absolute directory path. Engine binaries, metadata, sockets, and logs " +
+			"are stored under this path for both V1 and V2 engines. This is an installation-time setting and " +
+			"cannot be changed after Longhorn has been initialized.",
```
```suggestion
		Description: "Default path used for storing runtime and control-plane artifacts on a host. " +
			"This setting must be an absolute directory path. Engine binaries, metadata, sockets, and logs " +
			"are stored under this path for both V1 and V2 engines. Note: This is an installation-time " +
			"setting and cannot be changed after Longhorn is initialized.",
```

sushant-suse (changes requested):
Some minor suggestions for your consideration.
Thanks.

types/setting.go
davidcheng0922: Thank you, fix it

types/setting.go
davidcheng0922: Thank you, fix it

sushant-suse (approved):

datastore/longhorn.go
derekbit: Yes, we can make it immutable and see the feedback from users.

derekbit (approved):
LGTM

sushant-suse (approved):

PR #4910, opened 2026-06-18
fix: skip unscheduled replicas in volume

AoRuiAC:

Issue longhorn/longhorn#13355

For newly created volumes that have never been attached to a pod, expanding the volume fails with:

  cannot find the corresponding ready node and disk with disk UUID

  This happens because replicas for such volumes exist but have an empty `DiskID` (not yet scheduled to a disk). The `validateExpansionSize` webhook validator calls `GetReadyDiskNode` with an empty disk UUID, which fails
  since no disk matches.

### DIFF
--- a/webhook/resources/volume/validator.go
+++ b/webhook/resources/volume/validator.go
@@ -391,6 +391,13 @@ func (v *volumeValidator) validateExpansionSize(oldVolume *longhorn.Volume, newV
 	}
 
 	for _, replica := range replicaMap {
+		// An empty DiskID means the replica has not been scheduled to a disk yet.
+		// This can happen for newly created volumes that have never been attached.
+		// Since there is no underlying disk filesystem to check for size compatibility,
+		// it is safe to skip the validation for these unscheduled replicas.
+		if replica.Spec.DiskID == "" {
+			continue
+		}
 		diskUUID := replica.Spec.DiskID
 		node, diskName, err := v.ds.GetReadyDiskNode(diskUUID)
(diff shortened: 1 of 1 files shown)


webhook/resources/volume/validator.go
```diff
@@ -393,6 +391,9 @@ func (v *volumeValidator) validateExpansionSize(oldVolume *longhorn.Volume, newV
 	}
 
 	for _, replica := range replicaMap {
+		if replica.Spec.DiskID == "" {
```
COLDTURNIP, line 398: Need comments to explain the skipping reason, include the meaning of empty disk ID, and why it is safe to skip replica checking at this moment.

AoRuiAC: fixed

shuo-wu (dismissed):

derekbit:
@AoRuiAC Isn't the PR applied to master branch?

AoRuiAC:
> @AoRuiAC Isn't the PR applied to master branch?

It can be applied to master. I am currently using 1.10.x and hope to update it as well

derekbit:
Got it. v1.10.x is not actively supported, so we won't release a new v1.10.x patch version.
Could you update to PR to master branch? We can backport it v1.12.x and v1.11.x.
You can upgrade the cluster to v1.11.3 later that is upcoming.

derekbit (approved):

derekbit:
@mergify backport v1.11.x v1.12.x master

PR #4931, opened 2026-06-24
feat: introduce CPU manager for v2 volume SPDK

mantissahz:

Issue # longhorn/longhorn#13248

Introduce a new setting `data-engine-number-of-cpu-cores` instead of `data-engine-cpu-mask` to prevent creating IM pods failed from the CPU is pinned by other workloads.
When creating v2 instance manager pod, the CPU pinning will be managed by the K8S CPU manager with the setting `data-engine-number-of-cpu-cores` value.
And then IM pod will get pinned CPUs when starting, and send it to the SPDK process

### DIFF
--- a/controller/controller_manager.go
+++ b/controller/controller_manager.go
@@ -262,11 +262,13 @@ func GetInstanceManagerCPURequirement(ds *datastore.DataStore, imName string) (*
 	}
 
 	cpuRequest := 0
+	cpuRequestVal := ""
 	switch im.Spec.DataEngine {
 	case longhorn.DataEngineTypeV1, longhorn.DataEngineTypeV2:
 		// TODO: Currently lhNode.Spec.InstanceManagerCPURequest is applied to both v1 and v2 data engines.
 		// In the future, we may want to support different CPU requests for them.
 		cpuRequest = lhNode.Spec.InstanceManagerCPURequest
+		cpuRequestVal = fmt.Sprintf("%dm", cpuRequest)
 		if cpuRequest == 0 {
 			guaranteedCPUPercentage, err := ds.GetSettingAsFloatByDataEngine(types.SettingNameGuaranteedInstanceManagerCPU, im.Spec.DataEngine)
 			if err != nil {
@@ -274,12 +276,26 @@ func GetInstanceManagerCPURequirement(ds *datastore.DataStore, imName string) (*
 			}
 			allocatableMilliCPU := float64(kubeNode.Status.Allocatable.Cpu().MilliValue())
 			cpuRequest = int(math.Round(allocatableMilliCPU * guaranteedCPUPercentage / 100.0))
+			cpuRequestVal = fmt.Sprintf("%dm", cpuRequest)
+		}
+		if types.IsDataEngineV2(im.Spec.DataEngine) {
+			spdkCoreNumber, err := ds.GetSettingAsIntByDataEngine(types.SettingNameDataEngineNumberOfCPUCores, im.Spec.DataEngine)
+			if err != nil {
+				return nil, err
+			}
+			if spdkCoreNumber > 0 {
+				if cpuRequest < int(spdkCoreNumber)*1000 {
+					cpuRequestVal = fmt.Sprintf("%d", spdkCoreNumber)
+				} else {
+					cpuRequestVal = fmt.Sprintf("%d", ((cpuRequest-1)/1000)+1)
+				}
+			}
 		}
 	default:
 		return nil, fmt.Errorf("unknown data engine %v", im.Spec.DataEngine)
 	}
 
-	return ParseResourceRequirement(fmt.Sprintf("%dm", cpuRequest))
+	return ParseResourceRequirement(cpuRequestVal)
 }
 
 func isControllerResponsibleFor(controllerID string, ds *datastore.DataStore, name, preferredOwnerID, currentOwnerID string) bool {
--- a/controller/instance_manager_controller.go
+++ b/controller/instance_manager_controller.go
@@ -559,13 +559,22 @@ func (imc *InstanceManagerController) syncInstanceStatus(im *longhorn.InstanceMa
 	return nil
 }
 
-func (imc *InstanceManagerController) isDateEngineCPUMaskApplied(im *longhorn.InstanceManager) (bool, error) {
+func (imc *InstanceManagerController) isDateEngineCPUMaskCoreNumberApplied(im *longhorn.InstanceManager) (bool, error) {
 	if types.IsDataEngineV1(im.Spec.DataEngine) {
 		return true, nil
 	}
 
 	if im.Status.CurrentState != longhorn.InstanceManagerStateRunning {
 		return true, nil
+	}
+
+	spdkCoreNumber, err := imc.ds.GetSettingAsIntByDataEngine(types.SettingNameDataEngineNumberOfCPUCores, im.Spec.DataEngine)
+	if err != nil {
+		return false, errors.Wrapf(err, "failed to get %v setting for checking data engine CPU mask", types.SettingNameDataEngineNumberOfCPUCores)
+	}
+
+	if spdkCoreNumber > 0 {
+		return im.Status.DataEngineStatus.V2.CPUCoreNumber == spdkCoreNumber, nil
 	}
 
 	if im.Spec.DataEngineSpec.V2.CPUMask != "" {
@@ -653,7 +662,7 @@ func (imc *InstanceManagerController) handlePod(im *longhorn.InstanceManager) er
 		log.WithError(err).Warnf("Failed to sync log settings to instance manager pod %v", im.Name)
 	}
 
-	dataEngineCPUMaskIsApplied, err := imc.isDateEngineCPUMaskApplied(im)
+	dataEngineCPUMaskIsApplied, err := imc.isDateEngineCPUMaskCoreNumberApplied(im)
 	if err != nil {
 		log.WithError(err).Warnf("Failed to sync date engine CPU mask to instance manager pod %v", im.Name)
 	}
@@ -1874,20 +1883,31 @@ func (imc *InstanceManagerController) createInstanceManagerPodSpec(im *longhorn.
 			logFlags = strings.ToLower(logFlagsSetting)
 		}
 
-		// CPU mask is required for SPDK.
-		cpuMask := im.Spec.DataEngineSpec.V2.CPUMask
-		if cpuMask == "" {
-			value, err := imc.ds.GetSettingValueExistedByDataEngine(types.SettingNameDataEngineCPUMask, dataEngine)
-			if err != nil {
-				return nil, err
-			}
+		spdkCoreNumber, err := imc.ds.GetSettingAsIntByDataEngine(types.SettingNameDataEngineNumberOfCPUCores, im.Spec.DataEngine)
+		if err != nil {
+			return nil, err
+		}
+		dynamicCPUPinningEnabled := spdkCoreNumber != 0
 
-			cpuMask = value
+		// CPU mask is required for SPDK.
+		cpuMask := ""
+		if !dynamicCPUPinningEnabled {
+			cpuMask = im.Spec.DataEngineSpec.V2.CPUMask
 			if cpuMask == "" {
-				return nil, fmt.Errorf("failed to get CPU mask setting for data engine %v", dataEngine)
+				value, err := imc.ds.GetSettingValueExistedByDataEngine(types.SettingNameDataEngineCPUMask, dataEngine)
+				if err != nil {
+					return nil, err
+				}
+
+				cpuMask = value
+				if cpuMask == "" {
+					return nil, fmt.Errorf("failed to get CPU mask setting for data engine %v", dataEngine)
+				}
 			}
 		}
+		// When CPU-manager-based pinning is enabled, SPDK CPUs are determined at runtime inside the IM pod (cpuMask may be empty here).
 		im.Status.DataEngineStatus.V2.CPUMask = cpuMask
+		im.Status.DataEngineStatus.V2.CPUCoreNumber = spdkCoreNumber
 
 		// Hugepage or legacy memory preallocation is required for SPDK.
 		hugepageEnabled, err := imc.ds.GetSettingAsBoolByDataEngine(types.SettingNameDataEngineHugepageEnabled, im.Spec.DataEngine)
@@ -1909,9 +1929,10 @@ func (imc *InstanceManagerController) createInstanceManagerPodSpec(im *longhorn.
 		}
 
 		args := []string{
-			"instance-manager",
+			"start-spdk-tgt",
 			"--spdk-log", logFlags,
 			"--spdk-cpumask", cpuMask,
+			"--spdk-core-number", fmt.Sprintf("%d", spdkCoreNumber),
 			"--spdk-memory-size", fmt.Sprintf("%d", memory),
 			"--enable-spdk", "--debug",
 			"daemon",
@@ -1947,6 +1968,15 @@ func (imc *InstanceManagerController) createInstanceManagerPodSpec(im *longhorn.
 		}
 
 		podSpec.Spec.Containers[0].Resources.Limits[corev1.ResourceName("hugepages-2Mi")] = resource.MustParse(fmt.Sprintf("%vMi", hugepage))
+		if dynamicCPUPinningEnabled {
+			cpuQty, ok := podSpec.Spec.Containers[0].Resources.Requests[corev1.ResourceCPU]
+			if !ok {
+				cpuQty = resource.MustParse(fmt.Sprintf("%d", spdkCoreNumber))
+				podSpec.Spec.Containers[0].Resources.Requests[corev1.ResourceCPU] = cpuQty
+			}
+			podSpec.Spec.Containers[0].Resources.Limits[corev1.ResourceCPU] = cpuQty
+			podSpec.Spec.Containers[0].Resources.Limits[corev1.ResourceMemory] = resource.MustParse("128Mi")
+		}
 
 		podSpec.Spec.Containers[0].Lifecycle = &corev1.Lifecycle{
 			PreStop: &corev1.LifecycleHandler{
--- a/controller/monitor/environment_check_monitor.go
+++ b/controller/monitor/environment_check_monitor.go
@@ -2,7 +2,9 @@ package monitor
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
+	"os"
 	"reflect"
 	"strings"
 	"sync"
@@ -38,7 +40,20 @@ const (
 
 	kernelConfigDir = "/host/boot/"
 	systemConfigDir = "/host/etc/"
+
+	// Path to the kubelet CPU manager state file.
+	// If running in a pod, this path should match where you mounted the host's /var/lib/kubelet
+	DefaultKubeletRootDir      = "/var/lib/kubelet"
+	DefaultCpuManagerStateFile = "/cpu_manager_state"
 )
+
+// CPUManagerState represents the structure of the kubelet's CPU manager state file.
+type CPUManagerState struct {
+	PolicyName    string                       `json:"policyName"`
+	DefaultCPUSet string                       `json:"defaultCpuSet"`
+	Entries       map[string]map[string]string `json:"entries,omitempty"`
+	Checksum      uint32                       `json:"checksum"`
+}
 
 var (
 	kernelModules       = map[string]string{"CONFIG_DM_CRYPT": "dm_crypt"}
@@ -173,6 +188,10 @@ func (m *EnvironmentCheckMonitor) environmentCheck(kubeNode *corev1.Node) *Colle
 
 	if isV2DataEngine {
 		m.checkHugePages(kubeNode, collectedData)
+	}
+
+	if err := m.syncNodeCPUManagerPolicy(collectedData); err != nil {
+		m.logger.WithError(err).Debug("Failed to sync node CPU manager policy")
 	}
 
 	return collectedData
@@ -561,4 +580,50 @@ func (m *EnvironmentCheckMonitor) syncNFSClientVersion(kubeNode *corev1.Node, co
 	}
 
 	collectedData.conditions = types.SetCondition(collectedData.conditions, longhorn.NodeConditionTypeNFSClientInstalled, longhorn.ConditionStatusTrue, "", "")
+}
+
+func (m *EnvironmentCheckMonitor) syncNodeCPUManagerPolicy(collectedData *CollectedEnvironmentCheckInfo) error {
+	kubeletCPUManagerStateFilePath := getKubeletCPUManagerStateFilePath()
+	data, err := lhns.ReadFileContent(kubeletCPUManagerStateFilePath)
+	if err != nil {
+		m.logger.WithError(err).Debugf("failed to read CPU state file %s", kubeletCPUManagerStateFilePath)
+		collectedData.conditions = types.SetCondition(collectedData.conditions,
+			longhorn.NodeConditionTypeCPUManagerPolicy, longhorn.ConditionStatusFalse,
+			string(longhorn.NodeConditionReasonCPUManagerPolicyUnConfigured),
+			fmt.Sprintf("Failed to read CPU state file %s on node %s", kubeletCPUManagerStateFilePath, m.nodeName))
+		return nil
+	}
+
+	// Unmarshal the JSON data
+	var state CPUManagerState
+	if err := json.Unmarshal([]byte(data), &state); err != nil {
+		m.logger.WithError(err).Debugf("failed to parse CPU state file %s", kubeletCPUManagerStateFilePath)
+		collectedData.conditions = types.SetCondition(collectedData.conditions,
+			longhorn.NodeConditionTypeCPUManagerPolicy, longhorn.ConditionStatusFalse,
+			string(longhorn.NodeConditionReasonCPUManagerPolicyUnConfigured),
+			fmt.Sprintf("Failed to parse CPU state file %s on node %s", kubeletCPUManagerStateFilePath, m.nodeName))
+		return nil
+	}
+
+	if state.PolicyName == longhorn.CpuManagerStaticPolicy {
+		collectedData.conditions = types.SetCondition(collectedData.conditions,
+			longhorn.NodeConditionTypeCPUManagerPolicy, longhorn.ConditionStatusTrue,
+			string(longhorn.NodeConditionReasonCPUManagerPolicyConfigured),
+			fmt.Sprintf("Node %v CPU manager policy is static", m.nodeName))
+	} else {
+		collectedData.conditions = types.SetCondition(collectedData.conditions,
+			longhorn.NodeConditionTypeCPUManagerPolicy, longhorn.ConditionStatusFalse,
+			string(longhorn.NodeConditionReasonCPUManagerPolicyUnConfigured),
+			fmt.Sprintf("Node %v CPU manager policy is %v", m.nodeName, state.PolicyName))
+	}
+
+	return nil
+}
+
+func getKubeletCPUManagerStateFilePath() string {
+	kubeletRootDirPath := strings.TrimRight(strings.TrimSpace(os.Getenv(types.EnvKubeletRootDir)), "/")
+	if kubeletRootDirPath == "" {
+		kubeletRootDirPath = DefaultKubeletRootDir
+	}
+	return kubeletRootDirPath + DefaultCpuManagerStateFile
 }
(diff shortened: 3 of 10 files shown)


mantissahz:
@Mergifyio backport v1.12.x

derekbit:
@mantissahz CI failed

datastore/longhorn.go
```diff
@@ -707,6 +707,21 @@ func (s *DataStore) ValidateSetting(name, value string) (err error) {
...
+
+		for dataEngine, v := range values {
+			coreNumber, err := strconv.Atoi(v)
+			if err != nil {
+				return errors.Wrapf(err, "failed to convert CPU core number %s for data engine %s to integer", v, dataEngine)
+			}
+			if coreNumber < 0 {
+				return errors.Errorf("CPU core number for data engine %s cannot be negative", dataEngine)
+			}
+		}
```
derekbit: > Introduce a new setting for configuring the number of CPU cores. The setting is mutually exclusive with the existing CPU mask setting

1. The validation is missing. Both CPU cores and CPU mask need validation to ensure they are mutually exclusive.
2. If user wants to set non-zero CPU cores, need to ensure the CPU manager is using `static` policy.

controller/instance_manager_controller.go
```diff
@@ -1874,19 +1905,34 @@ func (imc *InstanceManagerController) createInstanceManagerPodSpec(im *longhorn.
...
+		nodeCPUManagerState, err := imc.getNodeCPUManagerState()
+		if err != nil {
+			return nil, err
+		}
+
+		spdkCoreNumber, err := imc.ds.GetSettingValueExistedByDataEngine(types.SettingNameDataEngineCPUCoreNumber, dataEngine)
+		if err != nil {
+			return nil, err
+		}
+		dynamicCPUPinningEnabled := spdkCoreNumber != "0" && nodeCPUManagerState.PolicyName == CPUManagerStaticPolicy
```
derekbit: `spdkCoreNumber != "0"` can be checked in the validating webhook.

```suggestion
		dynamicCPUPinningEnabled := nodeCPUManagerState.PolicyName == CPUManagerStaticPolicy
```

controller/instance_manager_controller.go
```diff
@@ -568,6 +585,20 @@ func (imc *InstanceManagerController) isDateEngineCPUMaskApplied(im *longhorn.In
 		return true, nil
 	}
 
+	nodeCPUManagerState, err := imc.getNodeCPUManagerState()
+	if err != nil {
+		return true, err
+	}
```
COLDTURNIP: Should we just log and skip this part when the cpu_manager_state is not available? Returning an error from `isDateEngineCPUMaskApplied` breaks `handlePod`.

controller/instance_manager_controller.go
```diff
@@ -1909,9 +1955,10 @@ func (imc *InstanceManagerController) createInstanceManagerPodSpec(im *longhorn.
 		}
 
 		args := []string{
-			"instance-manager",
+			"get-pod-cpu-mask-start-spdk",
```
COLDTURNIP: Maybe use a more representative name.
```suggestion
			"instance-manager-spdk-launcher",
```

datastore/longhorn.go
mantissahz: 1. The CPU mask will not be zero after the mutator. But yes, I can still add it in the validator.
2. That might need a condition added in the lhnode to save the CPU policy.

derekbit: > The CPU mask will not be zero after the mutator. But yes, I can still add it in the validator.

We should remove the mutation. The CPU mask and number of cores should be set by user itself. What we do is to validate the values to
1. make sure they are mutually exclusive. 
2. cpu cores should be zero if the policy is not static

> That might need a condition added in the lhnode to save the CPU policy.

Why do we need it?

derekbit: Wait, I might be know your concern. Let's discuss it tomorrow.

derekbit: Rethought the logic.
Currently, we have to two settings, cpu_mask and cpu_cores. By default, cpu_mask is 0x3 and cpu_cores is 0. There are three cases
1. User updates the cpu_mask to 0x3 and cpu_cores to 3. => Invalid update because the two values are mutually exclusive. Validating webhook should reject it.
2. User updates the cpu_mask to 0x0 and cpu_cores to 0. => Invalid update because the two values cannot be 0 at the same time. Validating webhook should reject it.
~~3. User updates the cpu_mask to 0xc and cpu_cores to 0. => Validating webhook should check if all nodes disables the `static` cpu policy. If one of nodes doesn't disable it, the update should be rejected.~~
4. User updates the cpu_mask to 0x0 and cpu_cores to 3. => Validating webhook should check if all nodes enables the `static` cpu policy. If one of nodes doesn't enable it, the update should be rejected.

Based on 3 and 4, we should introduce CPU policy in the lhnode resource as you mentioned.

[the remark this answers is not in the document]
davidcheng0922: This comment make sense to me

controller/controller_manager.go
```diff
@@ -274,12 +276,30 @@ func GetInstanceManagerCPURequirement(ds *datastore.DataStore, imName string) (*
...
+					cpuRequestVal = fmt.Sprintf("%d", ((cpuRequest-1)/1000)+1)
+				}
+			}
 		}
 	default:
 		return nil, fmt.Errorf("unknown data engine %v", im.Spec.DataEngine)
 	}
 
-	return ParseResourceRequirement(fmt.Sprintf("%dm", cpuRequest))
+	return ParseResourceRequirement(cpuRequestVal)
```
davidcheng0922, line 298: Is it possible for `lhNode.Spec.InstanceManagerCPURequest != 0` and `spdkCoreNumberStr == "0"`? If so, `cpuRequestVal` remains `""` and we eventually call ParseResourceRequirement("").

controller/instance_manager_controller.go
```diff
@@ -52,6 +52,23 @@ var (
 	mountPropagationHostToContainer = corev1.MountPropagationHostToContainer
 )
 
+const (
+	// Path to the kubelet CPU manager state file.
+	// If running in a pod, this path should match where you mounted the host's /var/lib/kubelet
+	kubeletCPUManagerStateFilePath = "/var/lib/kubelet/cpu_manager_state"
```
davidcheng0922: We hardcode the kubelet CPU manager state path here as `/var/lib/kubelet/cpu_manager_state`. 

Even though `FlagKubeletRootDir` is currently used for CSI, could this still conflict with environments where the kubelet root is not `/var/lib/kubelet`?

shuo-wu: IIRC, users can customize the root dir via flag `--root-dir`. Besides, I am not sure if `/var/lib/kubelet` fits all kinds of k8s distros.

types/setting.go
```diff
@@ -1803,6 +1806,17 @@ var (
 		Default:            fmt.Sprintf("{%q:\"0x3\"}", longhorn.DataEngineTypeV2),
 	}
 
+	SettingDefinitionDataEngineCPUCoreNumber = SettingDefinition{
+		DisplayName:        "Data Engine CPU Core Number",
+		Description:        "Applies only to the V2 Data Engine. Specifies the number of CPU cores allocated to the Storage Performance Development Kit (SPDK) target daemon. The daemon is deployed in each Instance Manager pod. Ensure that the assigned CPU cores do not exceed the guaranteed CPUs allocated to the V2 Data Engine Instance Manager. A minimum of 2 CPU cores is recommended. SPDK uses a busy-polling reactor model where the master reactor handles both I/O polling and management RPCs. When only a single core is assigned, heavy I/O workloads can delay or starve RPC processing, resulting in increased latency, timeout events, and operational instability. Assigning 2 or more cores allows I/O and management tasks to run on separate reactors, improving responsiveness and operational stability.",
```
shuo-wu: After introducing this, we need to update the description for the setting `SettingNameDataEngineCPUMask` as well

datastore/longhorn.go
shuo-wu: What's the final conclusion for this part? Will we introduce one more setting like `SettingNameDataEngineCPUPolicy`?

controller/controller_manager.go
```diff
@@ -262,6 +263,7 @@ func GetInstanceManagerCPURequirement(ds *datastore.DataStore, imName string) (*
 	}
 
 	cpuRequest := 0
+	cpuRequestVal := ""
 	switch im.Spec.DataEngine {
 	case longhorn.DataEngineTypeV1, longhorn.DataEngineTypeV2:
 		// TODO: Currently lhNode.Spec.InstanceManagerCPURequest is applied to both v1 and v2 data engines.
```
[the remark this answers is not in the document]
mantissahz: Modified. Thanks.

types/setting.go
```diff
@@ -1803,6 +1806,17 @@ var (
...
+		Category:           SettingCategoryDangerZone,
+		Type:               SettingTypeString,
+		Required:           true,
+		ReadOnly:           false,
+		DataEngineSpecific: true,
+		Default:            fmt.Sprintf("{%q:\"0\"}", longhorn.DataEngineTypeV2),
+	}
```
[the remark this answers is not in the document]

controller/instance_manager_controller.go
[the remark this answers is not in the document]
mantissahz: Modified. Thanks.

controller/instance_manager_controller.go
```diff
@@ -568,6 +585,20 @@ func (imc *InstanceManagerController) isDateEngineCPUMaskApplied(im *longhorn.In
...
+	}
+
+	spdkCoreNumber, err := imc.ds.GetSettingValueExistedByDataEngine(types.SettingNameDataEngineCPUCoreNumber, im.Spec.DataEngine)
+	if err != nil {
+		return true, errors.Wrapf(err, "failed to get %v setting for checking data engine CPU mask", types.SettingNameDataEngineCPUCoreNumber)
+	}
+
+	if spdkCoreNumber != "0" && nodeCPUManagerState.PolicyName == CPUManagerStaticPolicy {
+		return true, nil
+	}
```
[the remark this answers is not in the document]
mantissahz: Modified. Thanks.

controller/instance_manager_controller.go
```diff
@@ -1947,6 +1994,11 @@ func (imc *InstanceManagerController) createInstanceManagerPodSpec(im *longhorn.
 		}
 
 		podSpec.Spec.Containers[0].Resources.Limits[corev1.ResourceName("hugepages-2Mi")] = resource.MustParse(fmt.Sprintf("%vMi", hugepage))
+		if dynamicCPUPinningEnabled {
+			podSpec.Spec.Containers[0].Resources.Requests[corev1.ResourceCPU] = resource.MustParse(spdkCoreNumber)
+			podSpec.Spec.Containers[0].Resources.Limits[corev1.ResourceCPU] = resource.MustParse(spdkCoreNumber)
+			podSpec.Spec.Containers[0].Resources.Limits[corev1.ResourceMemory] = resource.MustParse("128Mi")
+		}
```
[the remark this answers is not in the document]
mantissahz: Modified. Thanks.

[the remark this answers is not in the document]
mantissahz: Modified. Thanks.

datastore/longhorn.go
mantissahz: > What's the final conclusion for this part? Will we introduce one more setting like SettingNameDataEngineCPUPolicy?

We will only have one new setting `data-engine-cpu-core-number`.
Users will control the CPU policy manually, and we will collect the CPU policy information in our node conditions.

> Currently, we have to two settings, cpu_mask and cpu_cores. By default, cpu_mask is 0x3 and cpu_cores is 0. There are three cases
> 1. User updates the cpu_mask to 0x3 and cpu_cores to 3. => Invalid update because the two values are mutually exclusive. Validating webhook should reject it.
> 2. User updates the cpu_mask to 0x0 and cpu_cores to 0. => Invalid update because the two values cannot be 0 at the same time. Validating webhook should reject it.
> 3. User updates the cpu_mask to 0xc and cpu_cores to 0. => Validating webhook should check if all nodes disables the static cpu policy. If one of nodes doesn't disable it, the update should be rejected.
> 4. User updates the cpu_mask to 0x0 and cpu_cores to 3. => Validating webhook should check if all nodes enables the static cpu policy. If one of nodes doesn't enable it, the update should be rejected.
  
> Based on 3 and 4, we should introduce CPU policy in the lhnode resource as you mentioned.

Due to the implementation of the setting on both our side and the Kubernetes side (where a setting can only be updated once at a time), we need to introduce a priority mechanism.

1. When CPU core number is not set, we use CPU mask.
2. When users try to set CPU core number, the kubelet CPU policy should be `static` on all nodes.
3. When CPU core number is set, we use CPU core number, whether the CPU mask is set or not.
4. Two settings can not be zero at the same time.

controller/instance_manager_controller.go
mantissahz: Modified. Thanks.
Here we consider it as enabled.

controller/instance_manager_controller.go
mantissahz: Removed. Thanks.

controller/instance_manager_controller.go
mantissahz: Modified. Thanks.

controller/controller_manager.go
mantissahz: Modified. Thanks.

controller/instance_manager_controller.go
mantissahz: Added an environment variable as the CSI part. Thanks

types/setting.go
```diff
@@ -1803,6 +1806,20 @@ var (
 		Default:            fmt.Sprintf("{%q:\"0x3\"}", longhorn.DataEngineTypeV2),
 	}
 
+	SettingDefinitionDataEngineCPUCoreNumber = SettingDefinition{
+		DisplayName:        "Data Engine CPU Core Number",
+		Description:        "Applies only to the V2 Data Engine. The Setting will overwrite the Data Engine CPU Mask setting. Specifies the number of CPU cores allocated to the Storage Performance Development Kit (SPDK) target daemon. The daemon is deployed in each Instance Manager pod. Ensure that the assigned CPU cores do not exceed the guaranteed CPUs allocated to the V2 Data Engine Instance Manager. A minimum of 2 CPU cores is recommended. SPDK uses a busy-polling reactor model where the master reactor handles both I/O polling and management RPCs. When only a single core is assigned, heavy I/O workloads can delay or starve RPC processing, resulting in increased latency, timeout events, and operational instability. Assigning 2 or more cores allows I/O and management tasks to run on separate reactors, improving responsiveness and operational stability.",
```
derekbit: > The Setting will overwrite the Data Engine CPU Mask setting

Actually, it doesn't overwrite. Please refine the description to mention the priority, ordering and avoid confusion.

controller/monitor/environment_check_monitor.go
```diff
@@ -562,3 +584,34 @@ func (m *EnvironmentCheckMonitor) syncNFSClientVersion(kubeNode *corev1.Node, co
...
+	data, err := lhns.ReadFileContent(kubeletCPUManagerStateFilePath)
+	if err != nil {
+		return errors.Wrapf(err, "failed to read CPU state file %s", kubeletCPUManagerStateFilePath)
+	}
+
+	// Unmarshal the JSON data
+	var state CPUManagerState
+	if err := json.Unmarshal([]byte(data), &state); err != nil {
+		return errors.Wrapf(err, "failed to parse CPU state file %s", kubeletCPUManagerStateFilePath)
+	}
```
derekbit, line 606: Should we return an error? If the CPUManagerPolicy is not parsable, can we set it to "unknow" and update the condition message? If it is "unknown", we can use cpu mask.

controller/monitor/environment_check_monitor.go
```diff
@@ -38,8 +40,24 @@ const (
...
 	kernelConfigDir = "/host/boot/"
 	systemConfigDir = "/host/etc/"
+
+	// Path to the kubelet CPU manager state file.
+	// If running in a pod, this path should match where you mounted the host's /var/lib/kubelet
+	DefaultKubeletRootDir      = "/var/lib/kubelet"
+	DefaultCpuManagerStateFile = "/cpu_manager_state"
+
+	CpuManagerStaticPolicy = "static"
+	CpuManagerNonePolicy   = "none"
```
derekbit: Move the definition to k8s/pkg/apis/longhorn/v1beta2/node.go

How about
```suggestion
	CpuManagerPolicyStatic = "static"
	CpuManagerPolicyNone   = "none"
	CpuManagerPolicyUnknown  = "unknown"
```

controller/monitor/environment_check_monitor.go
```diff
@@ -562,3 +584,34 @@ func (m *EnvironmentCheckMonitor) syncNFSClientVersion(kubeNode *corev1.Node, co
 
 	collectedData.conditions = types.SetCondition(collectedData.conditions, longhorn.NodeConditionTypeNFSClientInstalled, longhorn.ConditionStatusTrue, "", "")
 }
+
+func (m *EnvironmentCheckMonitor) syncNodeCPUManagerPolicy(collectedData *CollectedEnvironmentCheckInfo) error {
+	kubeletCPUManagerStateFilePath := getKubeletCPUManagerStateFilePath()
+	data, err := lhns.ReadFileContent(kubeletCPUManagerStateFilePath)
+	if err != nil {
+		return errors.Wrapf(err, "failed to read CPU state file %s", kubeletCPUManagerStateFilePath)
+	}
```
derekbit, line 595: Same as the following comment

datastore/longhorn.go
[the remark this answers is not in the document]
mantissahz: Modified. Thanks.

controller/monitor/environment_check_monitor.go
mantissahz: Modified. Thanks.
Set the status to false if it failed.

controller/monitor/environment_check_monitor.go
mantissahz: Moved. Thanks

controller/monitor/environment_check_monitor.go
mantissahz: Modified. Thanks.
Set the status to false if it failed.

types/setting.go

types/setting.go
mantissahz: Updated. Thanks.

controller/instance_manager_controller.go
```diff
@@ -568,6 +568,15 @@ func (imc *InstanceManagerController) isDateEngineCPUMaskApplied(im *longhorn.In
...
 	}
 
+	spdkCoreNumber, err := imc.ds.GetSettingAsIntByDataEngine(types.SettingNameDataEngineCPUCoreNumber, im.Spec.DataEngine)
+	if err != nil {
+		return false, errors.Wrapf(err, "failed to get %v setting for checking data engine CPU mask", types.SettingNameDataEngineCPUCoreNumber)
+	}
+
+	if spdkCoreNumber > 0 {
+		return true, nil
+	}
```
davidcheng0922, line 578: Question: it always return `true` for the case `>0`, if we change the core number from `2` to `3`, will the change be applied and recreate instance-manager pod?

mantissahz: Yes, you are right.
Refactor the implementation.
Added a new status field to save the CPU core number as the CPU mask in "InstanceManager.Status.DataEngineStatus.V2".
Updated. Thanks.

controller/instance_manager_controller.go
```diff
@@ -1874,19 +1905,34 @@ func (imc *InstanceManagerController) createInstanceManagerPodSpec(im *longhorn.
...
+					return nil, err
+				}
+
+				cpuMask = value
+				if cpuMask == "" {
+					return nil, fmt.Errorf("failed to get CPU mask setting for data engine %v", dataEngine)
+				}
 			}
 		}
+		// new setting spdk cpu count and exclusive or cpu mask
```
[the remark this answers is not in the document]
mantissahz: Updated. Thanks.

types/setting.go
```diff
@@ -1803,6 +1806,21 @@ var (
...
 	}
 
+	SettingDefinitionDataEngineCPUCoreNumber = SettingDefinition{
+		DisplayName: "Data Engine CPU Core Number",
+		Description: "Applies only to the V2 Data Engine. Only when the kubelet CPU policy is set to static, the setting can be applied. The Setting has higher priority than the Data Engine CPU Mask setting. Therefore, when specified, the CPU Mask setting will be ignored. " +
+			"Specifies the number of CPU cores allocated to the Storage Performance Development Kit (SPDK) target daemon. The daemon is deployed in each Instance Manager pod. Ensure that the assigned CPU cores do not exceed the guaranteed CPUs allocated to the V2 Data Engine Instance Manager. A minimum of 2 CPU cores is recommended. SPDK uses a busy-polling reactor model where the master reactor handles both I/O polling and management RPCs. When only a single core is assigned, heavy I/O workloads can delay or starve RPC processing, resulting in increased latency, timeout events, and operational instability. Assigning 2 or more cores allows I/O and management tasks to run on separate reactors, improving responsiveness and operational stability.",
```
derekbit: @sushant-suse Please review it. Thank you.

types/setting.go
```diff
@@ -1794,7 +1797,7 @@ var (
...
+		Description:        "Applies only to the V2 Data Engine. If the setting Data Engine CPU Core Number is specified, this setting will be ignored. Specifies the CPU cores on which the Storage Performance Development Kit (SPDK) target daemon runs. The daemon is deployed in each Instance Manager pod. Ensure that the assigned CPU cores do not exceed the guaranteed CPUs allocated to the V2 Data Engine Instance Manager. A minimum of 2 CPU cores is recommended. SPDK uses a busy-polling reactor model where the master reactor handles both I/O polling and management RPCs. When only a single core is assigned, heavy I/O workloads can delay or starve RPC processing, resulting in increased latency, timeout events, and operational instability. Assigning 2 or more cores allows I/O and management tasks to run on separate reactors, improving responsiveness and operational stability. Accepts either hexadecimal CPU masks (for example, 0x3 or 0xff) or CPU list format (for example, 0-1,2,5). CPU lists are automatically converted to hexadecimal masks. The default value is 0x3.",
```
derekbit: @sushant-suse Please review it. Thank you.

types/setting.go
```diff
@@ -157,6 +157,7 @@ const (
 	SettingNameDataEngineHugepageEnabled                                = SettingName("data-engine-hugepage-enabled")
 	SettingNameDataEngineMemorySize                                     = SettingName("data-engine-memory-size")
 	SettingNameDataEngineCPUMask                                        = SettingName("data-engine-cpu-mask")
+	SettingNameDataEngineCPUCoreNumber                                  = SettingName("data-engine-cpu-core-number")
```
derekbit: This is confusing. Could you rename it to `data-engine-number-of-cpu-cores`?

mantissahz: Updated. Thanks.

controller/instance_manager_controller.go
```diff
@@ -1947,6 +1968,15 @@ func (imc *InstanceManagerController) createInstanceManagerPodSpec(im *longhorn.
...
 		podSpec.Spec.Containers[0].Resources.Limits[corev1.ResourceName("hugepages-2Mi")] = resource.MustParse(fmt.Sprintf("%vMi", hugepage))
+		if dynamicCPUPinningEnabled {
+			cpuQty, ok := podSpec.Spec.Containers[0].Resources.Requests[corev1.ResourceCPU]
+			if !ok {
+				cpuQty = resource.MustParse(fmt.Sprintf("%d", spdkCoreNumber))
+				podSpec.Spec.Containers[0].Resources.Requests[corev1.ResourceCPU] = cpuQty
+			}
+			podSpec.Spec.Containers[0].Resources.Limits[corev1.ResourceCPU] = cpuQty
+			podSpec.Spec.Containers[0].Resources.Limits[corev1.ResourceMemory] = resource.MustParse("128Mi")
+		}
```
davidcheng0922, line 1979: Do we need to align with Request and Limits? 

```
podSpec.Spec.Containers[0].Resources.Requests[corev1.ResourceMemory] = resource.MustParse("128Mi")
```

types/setting.go
```suggestion
		Description:        "Applies only to the V2 Data Engine. If the Data Engine CPU Core Number setting is specified, this setting will be ignored. It specifies the CPU cores on which the Storage Performance Development Kit (SPDK) target daemon runs. The daemon is deployed in each Instance Manager pod. Ensure that the assigned CPU cores do not exceed the guaranteed CPUs allocated to the V2 Data Engine Instance Manager. A minimum of 2 CPU cores is recommended. SPDK uses a busy-polling reactor model where the master reactor handles both I/O polling and management RPCs. When only a single core is assigned, heavy I/O workloads can delay or starve RPC processing, resulting in increased latency, timeout events, and operational instability. Assigning 2 or more cores allows I/O and management tasks to run on separate reactors, improving responsiveness and operational stability. Accepts either hexadecimal CPU masks (for example, 0x3 or 0xff) or CPU list format (for example, 0-1,2,5). CPU lists are automatically converted to hexadecimal masks. The default value is 0x3.",
```

types/setting.go
```diff
@@ -1803,6 +1806,21 @@ var (
 		Default:            fmt.Sprintf("{%q:\"0x3\"}", longhorn.DataEngineTypeV2),
 	}
 
+	SettingDefinitionDataEngineNumberOfCPUCores = SettingDefinition{
+		DisplayName: "Data Engine Number of CPU Cores",
+		Description: "Applies only to the V2 Data Engine. Only when the kubelet CPU policy is set to static, the setting can be applied. The Setting has higher priority than the Data Engine CPU Mask setting. Therefore, when specified, the CPU Mask setting will be ignored. " +
```
```suggestion
		Description: "Applies only to the V2 Data Engine. It can be applied only when the kubelet CPU policy is set to static. It has higher priority than the Data Engine CPU Mask setting. Therefore, when specified, the CPU Mask setting will be ignored. " +
```

sushant-suse (approved):
A few suggestions for your consideration.
Thanks.

controller/instance_manager_controller.go
mantissahz: Yes, if we want to use the CPU manager, `Requests` and `Limits` should be the same.
And I have tested that if we remove the `Limits[corev1.ResourceMemory]`, the pod will not be assigned any CPUs.

ref: https://kubernetes.io/docs/concepts/workloads/resource-managers/

```yaml
spec:
  containers:
  - name: nginx
    image: nginx
    resources:
      limits:
        memory: "200Mi"
        cpu: "2"
      requests:
        memory: "100Mi"
        cpu: "1"
```
> The pod above runs in the Burstable QoS class because resource requests do not equal limits. It runs in the shared pool.

types/setting.go

types/setting.go

types/setting.go
```diff
@@ -157,6 +157,7 @@ const (
 	SettingNameDataEngineHugepageEnabled                                = SettingName("data-engine-hugepage-enabled")
 	SettingNameDataEngineMemorySize                                     = SettingName("data-engine-memory-size")
 	SettingNameDataEngineCPUMask                                        = SettingName("data-engine-cpu-mask")
+	SettingNameDataEngineNumberOfCPUCores                               = SettingName("data-engine-number-of-cpu-cores")
```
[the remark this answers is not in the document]
mantissahz: Updated. Thanks.

davidcheng0922 (approved):
LGTM

controller/instance_manager_controller.go
```diff
@@ -559,7 +559,7 @@ func (imc *InstanceManagerController) syncInstanceStatus(im *longhorn.InstanceMa
 	return nil
 }
 
-func (imc *InstanceManagerController) isDateEngineCPUMaskApplied(im *longhorn.InstanceManager) (bool, error) {
+func (imc *InstanceManagerController) isDateEngineCPUMaskCoreNumberApplied(im *longhorn.InstanceManager) (bool, error) {
```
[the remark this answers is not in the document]
mantissahz: Modified. Thanks. good catch

controller/instance_manager_controller.go
```diff
@@ -653,7 +662,7 @@ func (imc *InstanceManagerController) handlePod(im *longhorn.InstanceManager) er
 		log.WithError(err).Warnf("Failed to sync log settings to instance manager pod %v", im.Name)
 	}
 
-	dataEngineCPUMaskIsApplied, err := imc.isDateEngineCPUMaskApplied(im)
+	dataEngineCPUMaskIsApplied, err := imc.isDateEngineCPUMaskCoreNumberApplied(im)
 	if err != nil {
 		log.WithError(err).Warnf("Failed to sync date engine CPU mask to instance manager pod %v", im.Name)
 	}
```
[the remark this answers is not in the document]
mantissahz: Modified. Thanks.

derekbit (approved):

derekbit:
@mergify backport v1.12.x

PR #4933, opened 2026-06-25
fix: prevent false-positive v2 volume expansion success

davidcheng0922:

https://github.com/longhorn/longhorn/issues/13379

### DIFF
--- a/api/model.go
+++ b/api/model.go
@@ -1517,15 +1517,21 @@ func toVolumeResource(v *longhorn.Volume, vefs []*longhorn.EngineFrontend, ves [
 		endpoint := e.Status.Endpoint
 		controllerNodeID := e.Spec.NodeID
 		controllerSize := e.Status.CurrentSize
-		// For v2, the engine has no endpoint; use the matching EF's endpoint
-		// and size. The EF's CurrentSize reflects the frontend device size,
-		// which may lag behind the RAID (engine) size during expansion.
+		// For v2, the engine has no endpoint; use the matching EF's endpoint.
+		// Keep engine size as the source of truth, but for an attached volume
+		// do not expose a newer size until the frontend is also serving it.
 		if ef, ok := efByEngine[e.Name]; ok {
 			controllerNodeID = ef.Spec.NodeID
 			if ef.Status.Endpoint != "" {
 				endpoint = ef.Status.Endpoint
 			}
-			if ef.Status.CurrentSize > 0 {
+			// For an attached volume, also use the frontend size as a source
+			// of truth. If the frontend lags the backend engine, keep
+			// returning the older size until the frontend catches up.
+			if v.Status.State == longhorn.VolumeStateAttached &&
+				ef.Status.Endpoint != "" &&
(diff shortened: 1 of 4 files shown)


controller/volume_controller.go
```diff
@@ -2829,6 +2829,17 @@ func (c *VolumeController) reconcileVolumeSize(v *longhorn.Volume, e *longhorn.E
...
+		// requested size even though the target size already propagated to the
+		// Engine spec. Do not use EngineFrontend.CurrentSize here: detached or
+		// reattaching V2 frontends legitimately report 0 or temporarily lag the
+		// backend size, and treating that as expansion-incomplete can spuriously
+		// trigger volume-expansion attachment loops for otherwise idle volumes.
+		if e.Status.CurrentSize > 0 {
+			if e.Status.CurrentSize != v.Spec.Size {
+				v.Status.ExpansionRequired = true
+			}
+		}
```
carterli0407-cell, line 2850: The reassert block in `reconcileVolumeSize()` is not guarded by data-engine  type, so it also affects v1. The comment only talks about v2. Is the v1  behavior change intended? If yes, worth a v1 test + a note; if not, why not  wrap it in `types.IsDataEngineV2(...)`?

carterli0407-cell (commented):
LGTM overall.

api/model.go
```diff
@@ -1516,17 +1516,14 @@ func toVolumeResource(v *longhorn.Volume, vefs []*longhorn.EngineFrontend, ves [
 		endpoint := e.Status.Endpoint
 		controllerNodeID := e.Spec.NodeID
 		controllerSize := e.Status.CurrentSize
-		// For v2, the engine has no endpoint; use the matching EF's endpoint
-		// and size. The EF's CurrentSize reflects the frontend device size,
-		// which may lag behind the RAID (engine) size during expansion.
+		// For v2, the engine has no endpoint; use the matching EF's endpoint.
+		// Keep the controller size sourced from the engine so API/CSI callers
+		// observe backend expansion state rather than frontend device size.
```
derekbit: Why don't we use the size from engine frontend?

davidcheng0922: checking

controller/volume_controller.go
davidcheng0922: V1 does not have EngineFrontend, and it has always used the engine size for expansion progress. This change keeps the same behavior for V1.

The V2-specific part of the comment is only about why we intentionally do not use EngineFrontend.CurrentSize in the reassert path.

I also added a V1 unit test. Thank you

api/model.go
davidcheng0922: Fix it.

For an attached volume, we now require both the engine size and the engine frontend size to reach the new size before exposing the new size to the caller.

If the frontend size is still lagging, we keep returning the older size to avoid reporting expansion success too early.

derekbit (approved):
LGTM

shuo-wu (approved):

derekbit:
@mergify backport v1.12.x

PR #4934, opened 2026-06-25
fix: ublk volume attaching

carterli0407-cell:
Issue [#11977](https://github.com/longhorn/longhorn/issues/11977)

Fixes the V2 UBLK volume attach hang on kernel 6.17. Root cause: the EngineFrontend start gate in volume_controller.go required a non-zero engine port; UBLK never exposes one, so the EngineFrontend stayed stopped. Adds VolumeFrontendUblk to the bypass. Validated by rebuilding longhorn-manager only and confirming repeated UBLK attaches; NVMe unaffected.

ref: longhorn/longhorn#11977

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -2560,6 +2560,11 @@ func (c *VolumeController) openVolumeDependentResources(v *longhorn.Volume, e *l
 			// so we have the TargetIP available.
 			// For DR volumes (frontend disabled or empty), bypass the Port check
 			// because the engine may not expose a port when the frontend is off.
+			// For UBLK, also bypass the Port check: UBLK is a local block device
+			// exposed via ublk_drv, so the engine never reports a TCP port and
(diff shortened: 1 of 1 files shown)


controller/volume_controller.go
```diff
@@ -2561,7 +2561,7 @@ func (c *VolumeController) openVolumeDependentResources(v *longhorn.Volume, e *l
 			// For DR volumes (frontend disabled or empty), bypass the Port check
```
derekbit, line 2561: Could you update the comment for the ublk? Thanks.

derekbit:
@mergify backport v1.12.x

controller/volume_controller.go
carterli0407-cell: Updated. Thanks

derekbit (approved):
Well done! @carterli0407-cell

PR #4953, opened 2026-06-29
feat(v2): support CPU isolation

derekbit:

Issue Longhorn/longhorn#13417

### DIFF
--- a/controller/instance_manager_controller.go
+++ b/controller/instance_manager_controller.go
@@ -840,8 +840,8 @@ func (imc *InstanceManagerController) areDangerZoneSettingsSyncedToIMPod(im *lon
 			isSettingSynced, err = imc.isSettingInterruptModeEnabledSynced(setting, im)
 		case types.SettingNameDataEngineIobufLargePoolSize:
 			isSettingSynced, err = imc.isSettingIobufLargePoolSizeSynced(im, pod)
-		case types.SettingNameDataEngineIRQAffinityEnabled:
-			isSettingSynced, err = imc.isSettingIRQAffinityEnabledSynced(setting, im, pod)
+		case types.SettingNameDataEngineCPUIsolationEnabled:
+			isSettingSynced, err = imc.isSettingCPUIsolationEnabledSynced(setting, im, pod)
 		}
 		if err != nil {
 			return false, nil, false, false, err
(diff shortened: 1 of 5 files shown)


WebberHuang1118 (approved):
LGTM, thanks for the improvement.

WebberHuang1118 (approved):

controller/instance_manager_controller.go
```diff
@@ -989,6 +991,63 @@ func (imc *InstanceManagerController) isSettingInterruptModeEnabledSynced(settin
...
+	}
+	return val == "true", nil
+}
+
+// isSettingCPUIsolationEnabledSynced returns true if the effective CPU-isolation
+// value (Spec.DataEngineSpec.V2.CPUIsolationEnabled, falling back to the global
+// setting) matches what the V2 instance-manager pod was started with. The pod
+// always receives --longhorn-control-path (so we can reconcile stale state
+// across restarts even after toggling off); the actual toggle is the presence
+// of --enable-irq-affinity AND --enable-workqueue-affinity in the pod's args.
```
davidcheng0922, line 1021: Question: do we pass the arg `--longhorn-control-path`?

The control path is passed by env var

derekbit: Yes, I think passing the arg --longhorn-control-path should be fine. We can get the path from arg or from env. From the maintainance perspective, I perfer `--longhorn-control-path` arg.  WDYT?

davidcheng0922: I mean that I don't see the code where we pass it. Do I miss it?

derekbit: In the code changes in [controller/instance_manager_controller.go‎](https://github.com/longhorn/longhorn-manager/pull/4953/changes#diff-9c1df4b5ec912fbb4f6723505aef14b50bf80a6bdb86fb805a0b94e1d33d77a8)

davidcheng0922: I see, but it is `--spdk-control-path` not `--longhorn-control-path`, typo here?

derekbit: Good catch! It is --longhorn-control-path. Fixed. Thanks.

davidcheng0922 (approved):
LGTM

derekbit:
@mergify backport v1.12.x

PR #4958, opened 2026-07-01
feat(backing-image): honor system managed node selector

COLDTURNIP:

Issue longhorn/longhorn#12925


derekbit:
@COLDTURNIP Could you address the comments from Copilot?

PR #4960, opened 2026-07-03
fix: typo in validating data-engine-number-of-cpu-cores setting

mantissahz:

Issue # longhorn/longhorn#13248


mantissahz:
Will be fixed in another PR https://github.com/longhorn/longhorn-manager/pull/4975

PR #4974, opened 2026-07-05
fix(csi): reconcile CSI sidecar replica count on existing deployments

sebastiangaiser:

Issue https://github.com/longhorn/longhorn/issues/13461

The csi.*ReplicaCount Helm values only took effect when a CSI deployment was first created. deploy() short-circuited on a git-commit/version/image/anti-affinity match and never compared spec.replicas, so changing the count on a running cluster was a silent no-op. Add needToUpdateReplicas (Deployment-only, nil-safe) to the skip condition so a replica-count-only change is reconciled through the existing rolling-update path.


sebastiangaiser:
Fixed the commit message.

derekbit (approved):
LGTM. Thanks @sebastiangaiser for the contribution!

derekbit:
@mergify backport v1.12.x

PR #4978, opened 2026-07-08
feat: support per-component system-managed priority classes

swasabs:
## Why

Longhorn currently applies one `priority-class` value to all system-managed components. That was not enough for our node-memory-pressure case: `system-cluster-critical` did not protect node-local storage components reliably, so we needed finer control for instance-manager, engine-image, and CSI workloads.

## What changed

- Add a new optional setting: `system-managed-components-priority-classes`
- Keep `defaultSettings.priorityClass` as the global fallback
- Allow per-component overrides for:
  - `instance-manager`
  - `engine-image`
  - `longhorn-csi-plugin`
  - `csi-attacher`
  - `csi-provisioner`
  - `csi-resizer`
  - `csi-snapshotter`
- Apply the resolved priority class in the manager/controller code paths that create or reconcile these system-managed workloads
- Validate the new setting against existing PriorityClass objects

## Compatibility

- Existing installs keep current behavior if the new setting is unset
- No change to the meaning of `defaultSettings.priorityClass`
- The new setting is optional and only changes behavior when explicitly set

## Testing

- `go test ./types ./datastore`
- fork GitHub Actions run: build/test passed
- image publish failed only at Docker Hub login, which is expected on the fork

## Related

- Feature issue: https://github.com/longhorn/longhorn/issues/13476
- Chart companion PR: https://github.com/longhorn/longhorn/pull/13485


derekbit:
Hello @i3871 
Thanks for your contribution. Could you create a feature ticket for this PR in https://github.com/longhorn/longhorn/issues? Thanks.

controller/setting_controller_test.go
```diff
@@ -144,6 +146,69 @@ func TestCountCPUCoresFromMask(t *testing.T) {
 	}
 }
 
+func TestGetSystemManagedComponentFromRuntimeObject(t *testing.T) {
```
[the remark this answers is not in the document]
swasabs: Keeping this PR lean for now. The concrete validation bug from the other thread is fixed and covered; broader integrated controller coverage can be added if a human reviewer wants it.

types/deploy.go
```diff
@@ -22,6 +22,9 @@ const (
 	CSIResizerName     = "csi-resizer"
 	CSISnapshotterName = "csi-snapshotter"
 	CSIPluginName      = "longhorn-csi-plugin"
+
+	SystemManagedComponentInstanceManager = "instance-manager"
+	SystemManagedComponentEngineImage     = "engine-image"
```
derekbit, line 27: Don't we consider SystemManagedComponentLonghornManager?

swasabs: It doesn't have the longhorn.io/managed-by label the reconciler filters on, so this setting can't drive its priority class. Same gap exists for the existing global priority-class setting, not just this one.                                                                                                                                                                                              

It's wired differently: priority class for longhorn-manager is already settable, just separately, via longhornManager.priorityClass in the chart at install/upgrade time.                              

Can wire up dynamic support too in a follow-up, but it'd need its own path rather than reusing this one. Adding the shared label would also pull it into node selector/toleration/autoscaler reconcilers it shouldn't be in.

swasabs:
3.time asking conflicts resolution 
No plan on when this can be merged 
I don’t fancy keep on coming back to this

PR #4983, opened 2026-07-09
feat: tune host RPS to keep SPDK reactor cores off the network path

bachmanity1:
Add an opt-in node network tuning component in longhorn-manager that configures host Receive Packet Steering (RPS) on the physical interfaces so RX softirq processing is steered away from the CPU cores occupied by the SPDK target reactor. This prevents kernel network processing from stealing CPU from the busy-polling reactor cores, which can degrade V2 volume I/O under network load.

- New opt-in setting `data-engine-host-network-tuning` (default false).
- New controller/networktuning package: builds the RPS complement mask from the CPU mask the running V2 Instance Manager actually applied (Status.CPUMask) and writes it to every rx queue of the host physical interfaces via the host mount+net namespaces.
- Wired into InstanceManagerController: node-local, only for a running V2 Instance Manager, re-applied when the applied mask changes, and best-effort so it never fails the sync. Skips dynamic CPU-manager pinning (empty applied mask), where the cores are not known here.

Issue https://github.com/longhorn/longhorn/issues/13483

### DIFF
--- a/controller/instance_manager_controller.go
+++ b/controller/instance_manager_controller.go
@@ -1018,9 +1018,9 @@ func (imc *InstanceManagerController) resolveCPUIsolationEnabled(im *longhorn.In
 // setting) matches what the V2 instance-manager pod was started with. The pod
 // always receives --longhorn-control-path (so we can reconcile stale state
 // across restarts even after toggling off); the actual toggle is the presence
-// of --enable-irq-affinity AND --enable-workqueue-affinity in the pod's args.
-// Both flags are set together, so the effective state is "enabled" only when
-// both are present.
+// of --enable-irq-affinity, --enable-workqueue-affinity, AND --enable-rps in the
+// pod's args. All flags are set together, so the effective state is "enabled"
+// only when all are present.
 func (imc *InstanceManagerController) isSettingCPUIsolationEnabledSynced(setting *longhorn.Setting, im *longhorn.InstanceManager, pod *corev1.Pod) (bool, error) {
 	if types.IsDataEngineV1(im.Spec.DataEngine) {
 		return true, nil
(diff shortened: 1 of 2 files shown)


controller/instance_manager_controller.go
```diff
@@ -418,9 +426,88 @@ func (imc *InstanceManagerController) syncInstanceManager(key string) (err error
...
+			return
+		}
+		if err := networktuning.ResetRPS(log); err != nil {
+			// Do not cache the signature on failure, so the next reconcile retries.
+			log.WithError(err).Warn("Host network RPS reset did not fully succeed")
+			return
+		}
+		imc.lastNetworkTuningSig = ""
+		log.Info("Reset host network RPS to full CPU mask (host network tuning disabled)")
+		return
```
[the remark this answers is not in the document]
bachmanity1: The reset is best-effort, and I think this edge case is rare enough that it's not worth handling. Reset only runs when the feature was enabled and then transitioned to disabled, which means the host is already using RPS, so we reset it back to all-ones.

derekbit (commented):
@bachmanity1 
I'm working on the CPU isolation for v2 data engine. 
Is it possible to implement the RPS feature same as 
- https://github.com/longhorn/longhorn-instance-manager/pull/1509
- https://github.com/longhorn/longhorn-manager/pull/4953

The CPU mask update would be handled and set by the longhorn-instance-manager pre-start script, while longhorn-manager would control enabling and disabling the feature.

bachmanity1:
Hi @derekbit. I've aligned this PR with your CPU isolation code. I haven't changed the IM CRD, though, so RPS steering can only be enabled for all IMs or none. The IM-side code is [here](https://github.com/longhorn/longhorn-instance-manager/pull/1552).

controller/instance_manager_controller.go
```diff
@@ -842,6 +842,8 @@ func (imc *InstanceManagerController) areDangerZoneSettingsSyncedToIMPod(im *lon
 			isSettingSynced, err = imc.isSettingIobufLargePoolSizeSynced(im, pod)
 		case types.SettingNameDataEngineCPUIsolationEnabled:
 			isSettingSynced, err = imc.isSettingCPUIsolationEnabledSynced(setting, im, pod)
+		case types.SettingNameDataEngineHostNetworkTuning:
+			isSettingSynced, err = imc.isSettingHostNetworkTuningSynced(im, pod)
```
derekbit: @bachmanity1 
Can `SettingNameDataEngineCPUIsolationEnabled` also address this tunning? Or, is the independent setting better?

bachmanity1: Yes, I think it makes sense to fold this into `SettingNameDataEngineCPUIsolationEnabled`. Users who want to steer RPS away from the SPDK reactor cores would also want to steer IRQs away from them, so a single setting is a better fit than an independent one. I'll update the PR.

derekbit (approved):
LGTM. Thanks @bachmanity1

derekbit:
@mergify backport v1.12.x

derekbit:
cc @longhorn/dev

derekbit (approved):

PR #4984, opened 2026-07-09
fix: fix v2 data engine metrics missing

derekbit:

Issue Longhorn/longhorn#12941


mantissahz (approved):

derekbit:
I don't think the fix makes sense. Close it.

PR #4998, opened 2026-07-13
feat(csi): add volumeTopology parameter for zonal/regional volume placement

hookak:
Issue https://github.com/longhorn/longhorn/issues/13493


csi/controller_server.go
```diff
@@ -1691,6 +1743,100 @@ func (cs *ControllerServer) ControllerModifyVolume(ctx context.Context, req *csi
...
+// checkVolumeTopologyParameters rejects explicitly contradictory volume
+// parameters: a zonal volume keeps all of its replicas in one zone, which hard
+// zone anti-affinity can never satisfy for more than one replica. The volume
+// webhook enforces the same invariant on the volume itself (the mutator fills
+// replicaZoneSoftAntiAffinity enabled for zone-pinned volumes, the validator
+// rejects other values); this check just fails provisioning fast with a
+// parameter-level error when the StorageClass asks for both at once.
+func checkVolumeTopologyParameters(volumeParameters map[string]string) error {
+	if volumeParameters["volumeTopology"] == "zonal" &&
+		volumeParameters["replicaZoneSoftAntiAffinity"] == string(longhorn.ReplicaZoneSoftAntiAffinityDisabled) {
```
derekbit, line 1754: Why do we need to check `volumeParameters["replicaZoneSoftAntiAffinity"] == string(longhorn.ReplicaZoneSoftAntiAffinityDisabled)`? If `volumeTopology` is `zonal`, doesn't `volumeTopology` itself already constrain all replicas to the same zone, regardless of whether `replicaZoneSoftAntiAffinity` is `enabled` or `disabled?`

csi/controller_server.go
```diff
@@ -1691,6 +1743,100 @@ func (cs *ControllerServer) ControllerModifyVolume(ctx context.Context, req *csi
...
+// behavior), "zonal" resolves the zone of the first entry (the selected node's
+// zone under delayed binding), "regional" only its region. The resolved terms
+// are stored on the volume so every replica is scheduled in that failure
+// domain, and rendered back into the response topology (the PV node affinity)
+// so pods follow the volume.
+func volumeTopologyTerms(accessibleTopology []*csi.Topology, volumeTopology string) ([]longhornclient.VolumeTopologyTerm, error) {
+	switch volumeTopology {
+	case "", "any":
+		return nil, nil
+	case "zonal", "regional":
```
derekbit: nit: Could we introduce the following constants
```
VolumeTopologyAny
VolumeTopologyZonal
VolumeTopologyRegional
```

csi/controller_server.go
```diff
@@ -1691,6 +1743,100 @@ func (cs *ControllerServer) ControllerModifyVolume(ctx context.Context, req *csi
...
+	default:
+		return nil, fmt.Errorf("invalid volumeTopology %q, must be one of \"any\", \"zonal\", \"regional\"", volumeTopology)
+	}
+
+	if len(accessibleTopology) == 0 {
+		return nil, nil
+	}
+
+	segments := accessibleTopology[0].GetSegments()
```
derekbit, line 1781: Can `accessibleTopology` contain multiple entries at this point? If so, what does each entry represent, and is it safe to use only the first one?

derekbit (commented):
Some questions. I will contine reviewing the PR later today.

scheduler/shard_scheduler.go
```diff
@@ -67,6 +67,14 @@ func (ss *ShardScheduler) ScheduleShard(sg *longhorn.ShardGroup, vol *longhorn.V
 			continue
 		}
 
+		// Shards hold the volume data, so like replicas they must stay in one of
+		// the failure domains the volume resolved at creation. An empty topology
+		// requirement leaves the volume unconstrained.
+		if !types.NodeMatchesTopologyRequirement(node, vol.Spec.TopologyRequirement) {
+			skipReasons.Append("node is outside the volume topology requirement", fmt.Errorf("node %v", nodeName))
+			continue
+		}
```
derekbit, line 76: cc @c3y1huang

csi/controller_server.go
hookak: `volumeTopology` does not override zone anti-affinity — the scheduler enforces both independently. So with `replicaZoneSoftAntiAffinity: disabled` (hard zone anti-affinity) and `numberOfReplicas >= 2`, the two constraints are mutually unsatisfiable: the topology requirement keeps every replica in one zone while hard zone anti-affinity requires each replica in a different zone. The first replica schedules fine, and the conflict only surfaces later on scale-up or rebuild, leaving the volume permanently degraded.

csi/controller_server.go
hookak: Yes — each entry is one candidate failure domain from the CO's preferred list (requisite as fallback), filtered by `csi-allowed-topology-keys`. With delayed binding, external-provisioner puts the selected node's topology first, so `[0]` is where the pod landed; with immediate binding, it rotates the preferred order per volume, so `[0]` is its round-robin pick.

Using only `[0]` is safe because the response topology is narrowed to the same resolved term — the PV node affinity and the stored requirement are always the same single domain, so the pod cannot be scheduled into one of the other entries' zones. Idempotent retries return the stored term instead of re-resolving. This is the same model as GCE PD and ceph-csi.

Storing the whole list instead would lose which zone the volume actually lives in, and zone anti-affinity would then spread rebuilt replicas across the allowed zones — the opposite of what `zonal` promises.

csi/controller_server.go
hookak: Done in https://github.com/longhorn/longhorn-manager/pull/4998/commits/632e2191b34824c9e489f7996a83355b8fc2210e — added `types.VolumeTopologyAny` / `types.VolumeTopologyZonal` / `types.VolumeTopologyRegional` and replaced the literals.

derekbit (commented):
Do we need to update the node_server.go

```
     if kubeNode != nil {
         ns.ensureTopologyKey(kubeNode.Name, topologySegments, kubeNode.Labels, corev1.LabelHostname)
         ns.ensureTopologyKey(kubeNode.Name, topologySegments, kubeNode.Labels, corev1.LabelTopologyZone)
+		ns.ensureTopologyKey(kubeNode.Name, topologySegments, kubeNode.Labels, corev1.LabelTopologyRegion)
     }
```

hookak:
> Do we need to update the node_server.go

 Yes — we considered it too, and had hesitated to touch the block in this PR. I added it.

derekbit (approved):
LGTM

PR #5035, opened 2026-07-22
fix(recurring-job): retry and propagate execution errors

christophersherman:

Issue longhorn/longhorn#13587

Recurring-job pods currently report success through two failure paths:

- An initial Kubernetes API error while reading the RecurringJob CR is logged and returned as `nil`.
- Per-volume worker errors are assigned in a deferred function after the unnamed return value has already been evaluated.

In both cases the container exits 0, Kubernetes marks the Job complete, and the configured `restartPolicy: OnFailure` and `backoffLimit: 3` cannot take effect.

This patch:

- Retries transient initial RecurringJob GET failures for up to one minute with jittered exponential backoff. Retryable failures include transport timeouts, connection reset/refused, HTTP/2 connection loss, API timeouts, throttling, and service unavailability.
- Returns permanent GET errors immediately and preserves the final error after retry exhaustion.
- Returns `errgroup.Wait()` directly after all volume workers finish.

The command action exits nonzero after an unrecoverable or exhausted failure, allowing Kubernetes to apply its existing retry policy and report an accurate Job status. The startup retry matches the reliability model introduced for Longhorn manager API initialization by longhorn-manager#5032 and longhorn/longhorn#13567.

The per-volume change restores the behavior established by longhorn/longhorn#4255 and longhorn-manager#2099 before it regressed during the recurring-job package refactor.

The observed startup trigger was a transient failure reaching the Kubernetes API Service. This change is transport-agnostic and does not attribute the underlying connectivity failure to a specific network component.

The retry remains inside the existing Pod network namespace. It can tolerate delayed connectivity convergence but does not repair a permanently broken per-Pod endpoint. This PR does not change `restartPolicy`; changing it from `OnFailure` to `Never` would alter retry semantics for all recurring jobs and is intentionally out of scope.

A retry of a multi-volume job can repeat work for volumes that already succeeded; that separate behavior is tracked by longhorn/longhorn#10154.

Please backport this fix to v1.12.x.

Checks at `e0174a9e5`:

- Focused tests cover transient success, retry exhaustion, permanent errors, missing and successful RecurringJob lookups, retry classification, per-volume worker errors, and the no-volume success path.
- Linux `make ci` and the full race-enabled `make test` suite pass at https://github.com/christophersherman/longhorn-manager/actions/runs/30123414390. The overall workflow is red only because the fork cannot authenticate to Longhorn's Docker Hub publishing account after those steps pass.
- The branch is rebased onto current `master`.
- The two commits apply to current `v1.12.x` with a small adaptation for its older `urfave/cli` API; Linux-targeted compilation and `go vet ./app` pass there.

This PR does not address backup-store lock cleanup or the underlying network transient.

### DIFF
--- a/app/recurring_job.go
+++ b/app/recurring_job.go
@@ -15,6 +15,7 @@ import (
 	"github.com/longhorn/longhorn-manager/types"
 
 	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
+	lhtyped "github.com/longhorn/longhorn-manager/k8s/pkg/client/clientset/versioned/typed/longhorn/v1beta2"
 )
 
 func RecurringJobCmd() *cli.Command {
@@ -58,10 +59,9 @@ func recurringJob(cmd *cli.Command) (err error) {
 		return errors.Wrap(err, "failed to get clientset")
 	}
 
-	recurringJob, err := lhClient.LonghornV1beta2().RecurringJobs(namespace).Get(context.TODO(), jobName, metav1.GetOptions{})
+	recurringJob, err := getRecurringJob(lhClient.LonghornV1beta2().RecurringJobs(namespace), jobName)
 	if err != nil {
-		logrus.WithError(err).Errorf("Failed to get recurring job %v.", jobName)
-		return nil
+		return err
 	}
 
 	recurringJob.Status.ExecutionCount += 1
@@ -80,4 +80,13 @@ func recurringJob(cmd *cli.Command) (err error) {
 	default:
 		return recurringjob.StartVolumeJobs(job, recurringJob)
 	}
+}
+
+func getRecurringJob(client lhtyped.RecurringJobInterface, jobName string) (*longhorn.RecurringJob, error) {
+	recurringJob, err := client.Get(context.TODO(), jobName, metav1.GetOptions{})
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to get recurring job %v", jobName)
+	}
+
+	return recurringJob, nil
 }
--- a/app/recurring_job.go
+++ b/app/recurring_job.go
@@ -9,14 +9,21 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/urfave/cli/v3"
 
+	"k8s.io/apimachinery/pkg/util/wait"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	utilnet "k8s.io/apimachinery/pkg/util/net"
 
 	"github.com/longhorn/longhorn-manager/app/recurringjob"
 	"github.com/longhorn/longhorn-manager/types"
 
 	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
-	lhtyped "github.com/longhorn/longhorn-manager/k8s/pkg/client/clientset/versioned/typed/longhorn/v1beta2"
 )
+
+type recurringJobGetter interface {
+	Get(context.Context, string, metav1.GetOptions) (*longhorn.RecurringJob, error)
+}
 
 func RecurringJobCmd() *cli.Command {
 	return &cli.Command{
@@ -28,7 +35,7 @@ func RecurringJobCmd() *cli.Command {
 			},
 		},
 		Action: func(ctx context.Context, cmd *cli.Command) error {
-			if err := recurringJob(cmd); err != nil {
+			if err := recurringJob(ctx, cmd); err != nil {
 				logrus.WithError(err).Fatal("Failed to do a recurring job")
 			}
 			return nil
@@ -36,7 +43,7 @@ func RecurringJobCmd() *cli.Command {
 	}
 }
 
-func recurringJob(cmd *cli.Command) (err error) {
+func recurringJob(ctx context.Context, cmd *cli.Command) (err error) {
 	logger := logrus.StandardLogger()
 
 	var managerURL = cmd.String(FlagManagerURL)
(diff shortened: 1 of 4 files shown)


app/recurring_job.go
```diff
@@ -58,10 +59,9 @@ func recurringJob(cmd *cli.Command) (err error) {
 		return errors.Wrap(err, "failed to get clientset")
 	}
 
-	recurringJob, err := lhClient.LonghornV1beta2().RecurringJobs(namespace).Get(context.TODO(), jobName, metav1.GetOptions{})
+	recurringJob, err := getRecurringJob(lhClient.LonghornV1beta2().RecurringJobs(namespace), jobName)
```
mantissahz: Could we move getting the client into the `getRecurringJob`?
And then we don't have to import `lhtyped "github.com/longhorn/longhorn-manager/k8s/pkg/client/clientset/versioned/typed/longhorn/v1beta2"`

```suggestion
	recurringJob, err := getRecurringJob(namespace, jobName)
```

christophersherman: Thanks. I removed the generated typed-client import in e0174a9e5 by using a minimal local GET interface.

I kept client construction in recurringJob because that same clientset is immediately reused for UpdateStatus and NewJob, while the injected getter lets the tests deterministically exercise transient success, retry exhaustion, and permanent API errors. Moving construction entirely into getRecurringJob would either create a second clientset or couple these tests to in-cluster configuration.

Does this address the concern?

christophersherman:
Related follow-up: longhorn/longhorn-manager#5045 protects snapshots referenced by nonterminal Backup CRs during recurring cleanup.

This is kept separate because #5035 fixes retry/error propagation, while #5045 fixes a pre-existing cleanup correctness bug. The connection is that a fast container retry can begin cleanup after the previous process exited but while its engine-side backup is still asynchronous; without #5045, that retry can delete the prior Backup's source snapshot.

mantissahz (dismissed):
In general, LGTM

christophersherman:
@derekbit @mantissahz Thanks for the triage and review. This remains conflict-free, and the fork validation completed both `make ci` and the full `make test` suite. Mergify is still waiting for the upstream `Build binaries` check, a second approval, and an approval from the `longhorn/maintainer` team.

Could you help authorize/trigger the upstream workflow and route the remaining maintainer review? I have not force-pushed another rebase because I do not want to invalidate the existing approval; I can update the branch if you prefer.

app/recurringjob/volume.go
```diff
@@ -52,19 +52,14 @@ func StartVolumeJobs(job *Job, recurringJob *longhorn.RecurringJob) error {
...
-		}
-	}()
 	for _, volumeName := range filteredVolumes {
 		startJobVolumeName := volumeName
 		ewg.Go(func() error {
 			return startVolumeJob(job, recurringJob, startJobVolumeName, concurrentLimiter, jobGroups)
 		})
 	}
 
-	return err
```
derekbit, line 67: cc @c3y1huang

app/recurring_job.go
```diff
@@ -9,14 +9,21 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/urfave/cli/v3"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	utilnet "k8s.io/apimachinery/pkg/util/net"
+	"k8s.io/apimachinery/pkg/util/wait"
```
derekbit:
```suggestion
	"k8s.io/apimachinery/pkg/util/wait"

	apierrors "k8s.io/apimachinery/pkg/api/errors"
	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
	utilnet "k8s.io/apimachinery/pkg/util/net"
```

app/recurring_job_test.go
```diff
@@ -0,0 +1,198 @@
...
+
+	"github.com/stretchr/testify/assert"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/wait"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	k8stesting "k8s.io/client-go/testing"
```
derekbit:
```suggestion
	"k8s.io/apimachinery/pkg/runtime"
	"k8s.io/apimachinery/pkg/runtime/schema"
	"k8s.io/apimachinery/pkg/util/wait"

	apierrors "k8s.io/apimachinery/pkg/api/errors"
	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
	k8stesting "k8s.io/client-go/testing"
```

app/recurring_job.go
```diff
@@ -81,3 +87,60 @@ func recurringJob(cmd *cli.Command) (err error) {
...
+	ctx, cancel := context.WithTimeout(ctx, recurringjob.InitClientTimeout)
+	defer cancel()
+
+	// Nine attempts produce about 49-59 seconds of delay for immediate failures.
+	// The context deadline also bounds the time spent inside slow requests.
+	return getRecurringJobWithBackoff(ctx, client, jobName, wait.Backoff{
+		Duration: recurringjob.InitClientRetryInterval,
+		Factor:   1.5,
+		Jitter:   0.2,
+		Steps:    9,
```
derekbit, line 102: @christophersherman What considerations went into choosing them?

app/recurring_job.go
christophersherman: Addressed, thank you!

app/recurring_job_test.go
christophersherman: Addressed, thank you!

app/recurring_job.go
christophersherman: The one-minute timeout and one-second initial interval reuse the existing recurring-job startup bounds. The 1.5 factor keeps early retries close together while reducing API pressure during a longer outage, and 0.2 jitter spreads jobs that start together. Nine attempts keep the planned waits within that one-minute window; permanent API errors still return immediately.

derekbit (approved):
In general, LGTM. Thanks @christophersherman

derekbit:
@mergify backport v1.12.x

PR #5038, opened 2026-07-22
fix: keep running replica rebuilds from being blocked

derekbit:

Issue Longhorn/longhorn#13571

openVolumeDependentResources rebuilds the entire engine ReplicaAddressMap while the volume is already attached so newly added or rebuilt replicas can be picked up by the running engine.

Before this change, one replica that was not yet Running caused the function to return early. This preserved the initial attach safety behavior, but it also blocked already Running sibling replicas from being added to the engine's ReplicaAddressMap during steady-state reconcile. Since the engine only dials replicas listed in that map, those running replacements could remain undialed, never finish rebuilding, and keep rebuild concurrency slots occupied.

When the engine is already Running, preserve any existing address for a not-yet-Running replica and continue processing the rest of the replicas. This keeps existing connected replicas from being dropped as unknown while allowing newly Running sibling replicas to be added to the map. Initial attach behavior is unchanged: if the engine is not yet Running, Longhorn still waits for all potentially healthy replicas before starting the engine.

Add a regression test covering an attached degraded volume where a Running replacement replica must be added while another sibling replica is not Running.

### DIFF
--- a/controller/volume_controller.go
+++ b/controller/volume_controller.go
@@ -2686,6 +2686,19 @@ func (c *VolumeController) openVolumeDependentResources(v *longhorn.Volume, e *l
 		rs[r.Name] = r
 	}
 
+	// This function also runs while the volume is already Attached, to pick up replicas added or
+	// rebuilt afterward (see the "This is a stable state" call site in
+	// reconcileAttachDetachStateMachine). In that steady state, don't let one replica that is still
+	// starting/rebuilding block OTHER, already Running, sibling replicas from being (re-)added to
+	// e.Spec.ReplicaAddressMap below: that field is fully overwritten every call, so bailing out here
+	// would silently drop already-healthy replicas until every sibling becomes Running at the same
+	// time. That may never happen when rebuilds are serialized one at a time per volume/per node,
+	// leaving replacement replicas permanently un-dialed by the engine after a mass replica
+	// replacement event (https://github.com/longhorn/longhorn/issues/13571). Only wait for every
+	// replica to become Running when the engine itself hasn't started yet, since starting it for the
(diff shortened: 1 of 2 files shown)


shuo-wu (approved):
LGTM

controller/volume_controller.go
```diff
@@ -2710,8 +2723,19 @@ func (c *VolumeController) openVolumeDependentResources(v *longhorn.Volume, e *l
...
-		// wait for all potentially healthy replicas become running
 		if r.Status.CurrentState != longhorn.InstanceStateRunning {
+			if engineAlreadyRunning {
+				// The engine is already up. Keep this replica's existing address (if any) instead of
+				// waiting for it to become Running, so a still-starting or briefly-flapping replica
+				// does not block its running siblings from being kept in the map, while also not
+				// dropping a still-connected replica that would then be removed as unknown.
+				if addr, ok := e.Spec.ReplicaAddressMap[r.Name]; ok {
+					replicaAddressMap[r.Name] = addr
+				}
```
COLDTURNIP, line 2734: If the engine is running but the replica is dead, should we keep this dead replica address in the map?

derekbit: If the replica is failed, it will be skipped and won't be added to the map.
line 2720
```
		if r.Spec.FailedAt != "" {
			continue
		}
```

mantissahz (approved):
LGTM

derekbit:
> > If the engine is running but the replica is dead, should we keep this dead replica address in the map?


After discussing with @COLDTURNIP, the fix might be risky if a stopped but not failed replica is added to the engineReplicaMap. Instead, we should fail the replicas on the stopped node. @COLDTURNIP will investigate the possible fix.

PR #5054, opened 2026-07-28
fix: backup not starting when the responsible node is drained

carterli0407-cell:
Use the engine's node (engine.Spec.NodeID) instead of the backup controller's own node (bc.controllerID) when obtaining the instance manager for the backup engine client proxy. This prevents backups from failing with "failed to find a running instance manager for node ..." when the responsible node has been drained.

Issue longhorn/longhorn#12562


mantissahz:
I'm not sure whether we should always use the engine node to process the backup and send the snapshot to the remote backup server.
For example, 
1. Volume A is attached to node A with three replicas on nodes A, B, and C.
Node C is drained.
2. The backup controller on node B starts processing the backup of volume A.
3. The I/O load might be heavy on node A, and node B is idle.
4. We will still process the backup on node A because the `engineClientProxy` is using the instance manager on node A.

Of course, the situation would be reversed because we don't monitor I/O on every node.

I would prefer to select a running instance manager in the `getBackupTarget` function if the instance manager is running on the controller node.
I'm also fine with always selecting the engine node.
WDYT, @derekbit, @COLDTURNIP?

derekbit:
> I'm not sure whether we should always use the engine node to process the backup and send the snapshot to the remote backup server.

We shouldn't becuae the engine node might not have any replica. 

> I would prefer to select a running instance manager in the `getBackupTarget` function if the instance manager is running on the controller node.

Agree with @mantissahz's proposal!

carterli0407-cell:
> > I'm not sure whether we should always use the engine node to process the backup and send the snapshot to the remote backup server.
> 
> We shouldn't becuae the engine node might not have any replica.
> 
> > I would prefer to select a running instance manager in the `getBackupTarget` function if the instance manager is running on the controller node.
> 
> Agree with @mantissahz's proposal!

Thanks @mantissahz and @derekbit for the detailed feedback - that makes sense, and I agree with the concern.
You're right that always pinning to the engine node isn't ideal:
It ignores the fact that the controller node may be idle while the engine/attached node is under heavy I/O, so we'd lose the chance to offload the backup processing.
And as @derekbit noted, the engine node isn't guaranteed to host a replica, so it's not inherently the "best" node to anchor on either.
So I'll move the selection logic into `getBackupTarget` and make it pick a running instance manager instead of requiring one on a fixed node.

PR #5055, opened 2026-07-29
fix(backup): retain monitor until status persists

COLDTURNIP:

Issue longhorn/longhorn#13611

### DIFF
--- a/controller/backup_controller.go
+++ b/controller/backup_controller.go
@@ -392,15 +392,23 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
 		if bc.backupInFinalState(existingBackup) && (!existingBackup.Status.LastSyncedAt.IsZero() || existingBackup.Spec.SnapshotName == "") {
 			err = bc.handleAttachmentTicketDeletion(backup, canonicalBackupVolumeName)
 		}
-		if reflect.DeepEqual(existingBackup.Status, backup.Status) {
-			return
+		if !reflect.DeepEqual(existingBackup.Status, backup.Status) {
+			_, statusErr := bc.ds.UpdateBackupStatus(backup)
+			if statusErr != nil {
+				if apierrors.IsConflict(errors.Cause(statusErr)) {
+					log.WithError(statusErr).Debugf("Requeue %v due to conflict", backupName)
+					bc.enqueueBackup(backup)
+				} else {
+					err = errors.Join(err, statusErr)
+				}
+				return
+			}
 		}
-		if _, err := bc.ds.UpdateBackupStatus(backup); err != nil && apierrors.IsConflict(errors.Cause(err)) {
-			log.WithError(err).Debugf("Requeue %v due to conflict", backupName)
-			bc.enqueueBackup(backup)
-			err = nil // nolint: ineffassign
-			return
+
+		if bc.backupInFinalState(backup) {
+			bc.disableBackupMonitor(backup.Name)
 		}
+
 		if backup.Status.State == longhorn.BackupStateCompleted && existingBackupState != backup.Status.State {
 			if err := bc.syncBackupVolume(backupTargetName, canonicalBackupVolumeName); err != nil {
 				log.Warnf("Failed to sync backup volume %v for backup target %v", canonicalBackupVolumeName, backupTargetName)
@@ -469,11 +477,8 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
 		switch backup.Status.State {
 		case longhorn.BackupStateNew, longhorn.BackupStatePending, longhorn.BackupStateInProgress:
 			return nil
-		case longhorn.BackupStateCompleted:
-			bc.disableBackupMonitor(backup.Name)
 		case longhorn.BackupStateError, longhorn.BackupStateUnknown:
 			backup.Status.LastSyncedAt = syncTime
-			bc.disableBackupMonitor(backup.Name)
 			return nil
 		}
 	}
--- a/controller/backup_controller_test.go
+++ b/controller/backup_controller_test.go
@@ -1,0 +1,358 @@
+package controller
+
+import (
+	"context"
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/cockroachdb/errors"
+
+	"github.com/sirupsen/logrus"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes/fake"
+	"k8s.io/client-go/tools/record"
+
+	apiextensionsfake "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset/fake"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	k8stesting "k8s.io/client-go/testing"
+
+	"github.com/longhorn/longhorn-manager/datastore"
+	"github.com/longhorn/longhorn-manager/engineapi"
+	"github.com/longhorn/longhorn-manager/types"
+	"github.com/longhorn/longhorn-manager/util"
+
+	longhorn "github.com/longhorn/longhorn-manager/k8s/pkg/apis/longhorn/v1beta2"
+	lhfake "github.com/longhorn/longhorn-manager/k8s/pkg/client/clientset/versioned/fake"
+)
+
+const (
+	backupLifecycleTestSnapshotName = "missing-snapshot"
+	backupLifecycleTestBackupURL    = "backup://monitor"
+)
+
+type backupLifecycleTestEngineClientProxy struct {
+	*engineapi.EngineSimulator
+}
+
+func (p *backupLifecycleTestEngineClientProxy) Close() {}
+
+func (p *backupLifecycleTestEngineClientProxy) SnapshotBackupStatus(obj engineapi.DataEngineObject, backupName, replicaAddress, replicaName string) (*longhorn.EngineBackupStatus, error) {
+	return &longhorn.EngineBackupStatus{
+		Progress:     100,
+		BackupURL:    backupLifecycleTestBackupURL,
+		SnapshotName: backupLifecycleTestSnapshotName,
+		State:        "complete",
+	}, nil
+}
+
+var _ engineapi.EngineClientProxy = (*backupLifecycleTestEngineClientProxy)(nil)
+
+func newBackupLifecycleTestController(t *testing.T) (*BackupController, *datastore.DataStore, *lhfake.Clientset, *longhorn.Backup) {
+	t.Helper()
+
+	kubeClient := fake.NewSimpleClientset()                    // nolint: staticcheck
+	lhClient := lhfake.NewSimpleClientset()                    // nolint: staticcheck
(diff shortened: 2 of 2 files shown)


c3y1huang (commented):
As discussed in the internal meeting, I'm slightly confused. This issue sounds like a regrssion with 1.13.0 and 1.12.1 backport, but the [RCA](https://github.com/longhorn/longhorn/issues/13611#issuecomment-5101116798) doesn't seem to explain what introduced it. If this is a day one issue, does 1.11.x need the fix too?

COLDTURNIP:
This fix is verified by triggering e2e test several times:

- Job 8699, 8705, 8712, 8714: success
- job 8711, 8713, 8716: backup initialization failed because of `no such device or address` during opening nvme device (longhorn/longhorn#13188)

> As discussed in the internal meeting, I'm slightly confused. This issue sounds like a regrssion with 1.13.0 and 1.12.1 backport, but the [RCA](https://github.com/longhorn/longhorn/issues/13611#issuecomment-5101116798) doesn't seem to explain what introduced it. If this is a day one issue, does 1.11.x need the fix too?

v1.11.x backport seems to be needed. I can't find a reason why v1.11.x didn't have such issue, and it is reproducible only on v1.12.x with v2 data engine. Maybe it was shadowed by longhorn/longhorn#13188.

### Why not happening with v1 data engine

When a backup initialization is finished but the monitor is mistakenly dropped & recreated, the replica address becomes empty; instance manager receives a request without a meaningful replica address, then [it iterate over the existing replicas and tries to figure out an available address](https://github.com/longhorn/longhorn-instance-manager/blob/master/pkg/proxy/backup.go#L182-L209). V2 API simply using the [exactly match](https://github.com/longhorn/longhorn-instance-manager/blob/master/pkg/proxy/backup.go#L278) without checking the address.

controller/backup_controller.go
```diff
@@ -395,12 +395,12 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
 		if reflect.DeepEqual(existingBackup.Status, backup.Status) {
 			return
 		}
-		if _, err := bc.ds.UpdateBackupStatus(backup); err != nil && apierrors.IsConflict(errors.Cause(err)) {
-			log.WithError(err).Debugf("Requeue %v due to conflict", backupName)
-			bc.enqueueBackup(backup)
-			err = nil // nolint: ineffassign
+		updated, updateErr := bc.persistBackupStatus(backup, log)
+		err = updateErr
```
c3y1huang: This can potentially overwrite the err returned from `bc.handleAttachmentTicketDeletion`.

controller/backup_controller.go
```diff
@@ -395,12 +395,12 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
 		if reflect.DeepEqual(existingBackup.Status, backup.Status) {
 			return
 		}
-		if _, err := bc.ds.UpdateBackupStatus(backup); err != nil && apierrors.IsConflict(errors.Cause(err)) {
-			log.WithError(err).Debugf("Requeue %v due to conflict", backupName)
-			bc.enqueueBackup(backup)
-			err = nil // nolint: ineffassign
+		updated, updateErr := bc.persistBackupStatus(backup, log)
```
c3y1huang: Would it be possible for the backup to have already reached a final state before the monitor is disabled, causing the monitor to leak? I suppose not in the normal path, but it might be worth adding a cheap safty net to make this idempotent.

controller/backup_controller.go
```diff
@@ -544,6 +541,25 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
...
+		if apierrors.IsConflict(errors.Cause(err)) {
+			log.WithError(err).Debugf("Requeue %v due to conflict", backup.Name)
+			bc.enqueueBackup(backup)
+			return false, nil
+		}
+		return false, err
+	}
+
+	switch backup.Status.State {
+	case longhorn.BackupStateCompleted, longhorn.BackupStateError, longhorn.BackupStateUnknown:
```
c3y1huang: can we reuse `bc.backupInFinalState`?

controller/backup_controller.go
```diff
@@ -389,18 +389,27 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
...
-			err = nil // nolint: ineffassign
+		updated, statusErr := bc.persistBackupStatus(backup, log)
+		if attachmentErr != nil {
+			err = attachmentErr
+		} else {
+			err = statusErr
+		}
+		if !updated {
 			return
 		}
```
derekbit: Don't we need to enqueueBackup if we hit an error or updated is false?

controller/backup_controller.go
```diff
@@ -544,6 +550,24 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
 	return err
 }
 
+// Retain a terminal monitor until its status update succeeds. A conflict requeues the Backup with its monitor intact so stale non-final status cannot recreate the monitor.
+func (bc *BackupController) persistBackupStatus(backup *longhorn.Backup, log *logrus.Entry) (updated bool, err error) {
+	if _, err = bc.ds.UpdateBackupStatus(backup); err != nil {
+		if apierrors.IsConflict(errors.Cause(err)) {
+			log.WithError(err).Debugf("Requeue %v due to conflict", backup.Name)
+			bc.enqueueBackup(backup)
```
derekbit: `persistBackupStatus()` can only focus on the backup status persistence. Suggest moving `bc.enqueueBackup(backup)` outside the `persistBackupStatus()`. WDYT?

controller/backup_controller.go
COLDTURNIP: `persistBackupStatus` will requeue the failed CR. Will extract the requeue logic to here.

controller/backup_controller.go
```diff
@@ -389,18 +389,27 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
...
-		if _, err := bc.ds.UpdateBackupStatus(backup); err != nil && apierrors.IsConflict(errors.Cause(err)) {
-			log.WithError(err).Debugf("Requeue %v due to conflict", backupName)
-			bc.enqueueBackup(backup)
-			err = nil // nolint: ineffassign
+		updated, statusErr := bc.persistBackupStatus(backup, log)
+		if attachmentErr != nil {
+			err = attachmentErr
+		} else {
+			err = statusErr
+		}
```
c3y1huang: Consider using errors.Join.

controller/backup_controller.go
c3y1huang: Suggest also moving bc.disableBackupMonitor out to the call site so the function stays honest to its name. IMO, persisting status and tearing down the monitor are separate concerns.

controller/backup_controller.go
```diff
@@ -392,15 +392,23 @@ func (bc *BackupController) reconcile(backupName string) (err error) {
...
+		if !reflect.DeepEqual(existingBackup.Status, backup.Status) {
+			_, statusErr := bc.ds.UpdateBackupStatus(backup)
+			if statusErr != nil {
+				if apierrors.IsConflict(errors.Cause(statusErr)) {
+					log.WithError(statusErr).Debugf("Requeue %v due to conflict", backupName)
+					bc.enqueueBackup(backup)
+				} else {
+					err = errors.Join(err, statusErr)
+				}
+				return
```
c3y1huang, line 404: Is it possible this potentially introduces a regression?
1. The status update failed w/ non-conflict error -> monitor kept alive (status persisted).
2. Kept alive means the poll loop keeps polling. If the engine failed to answer the poll during this window -> [status flipped to Error](https://github.com/COLDTURNIP/longhorn-manager/blob/08e534a26f48c968283d016f0882c9ea8af3c789/engineapi/backup_monitor.go#L220).
3. The retry reconcile arrives and persists Error + [LastSyncedAt](https://github.com/COLDTURNIP/longhorn-manager/blob/08e534a26f48c968283d016f0882c9ea8af3c789/controller/backup_controller.go#L481) + [disabled monitor](https://github.com/COLDTURNIP/longhorn-manager/blob/08e534a26f48c968283d016f0882c9ea8af3c789/controller/backup_controller.go#L409). The [LastSyncedAt guard](https://github.com/COLDTURNIP/longhorn-manager/blob/08e534a26f48c968283d016f0882c9ea8af3c789/controller/backup_controller.go#L487-L490) then short-circuits all future reconciles.
4. [deleteSnapshotAfterBackupCompleted](https://github.com/COLDTURNIP/longhorn-manager/blob/08e534a26f48c968283d016f0882c9ea8af3c789/controller/backup_controller.go#L417) never fires.

COLDTURNIP: To the following scenario:

- 1st reconcile loop: backup completed, but failed to save CR status. The monitor keep running.
- Monitor observes some failure from IM, flips to error, and trigger a reconcile asynchronously.
- 2nd reconcile loop: controller fetch error state from the monitor, save the CR status with state `error`, then shutdown the monitor
- The following reconcile loops: the `deleteSnapshotAfterBackupCompleted` is never invoked

I think this is the same case that the backup creation failed. If we keep the succeed state in the monitor after completion, it is still possible that the IM is not able to properly handle the snapshot deletion. In addition, user is still able to delete the snapshot manually if they don't like to keep it. WDYT? @c3y1huang

derekbit:
@COLDTURNIP If we have https://github.com/longhorn/longhorn-instance-manager/pull/1589, do we really need the PR?

COLDTURNIP:
> @COLDTURNIP If we have [longhorn/longhorn-instance-manager#1589](https://github.com/longhorn/longhorn-instance-manager/pull/1589), do we really need the PR?

Indeed, the replica address issue can be resolved when the IM is able to handle the empty replica address. But prevent recreating the monitor after the backup creation completed would prevent potential issue. I think this PR is nice to have.

derekbit:
> > @COLDTURNIP If we have [longhorn/longhorn-instance-manager#1589](https://github.com/longhorn/longhorn-instance-manager/pull/1589), do we really need the PR?
> 
> Indeed, the replica address issue can be resolved when the IM is able to handle the empty replica address. But prevent recreating the monitor after the backup creation completed would prevent potential issue. I think this PR is nice to have.

It seems the longhorn-manager PR might introduce potential regression. We can create another ticket to improve and validate later.
Have you checked if the issue can fix the issue only with the IM PR on Jenkins?

COLDTURNIP:
To the issue longhorn/longhorn#13611, changed to `ready for testing` since it should be covered by [the IM change](https://github.com/longhorn/longhorn-instance-manager/pull/1589).

derekbit:
> To the issue [longhorn/longhorn#13611](https://github.com/longhorn/longhorn/issues/13611), changed to `ready for testing` since it should be covered by [the IM change](https://github.com/longhorn/longhorn-instance-manager/pull/1589).

@COLDTURNIP Could you add the link to the Jenkins testing result only with the IM PR? Thanks.

COLDTURNIP:
> > To the issue [longhorn/longhorn#13611](https://github.com/longhorn/longhorn/issues/13611), changed to `ready for testing` since it should be covered by [the IM change](https://github.com/longhorn/longhorn-instance-manager/pull/1589).
> 
> @COLDTURNIP Could you add the link to the Jenkins testing result only with the IM PR? Thanks.

Let me trigger one with the latest merged change.

COLDTURNIP:
To the issue longhorn/longhorn#13611, the IM PR covers: longhorn/longhorn-instance-manager#1589

We don't introduce this PR for this issue.

PR #5056, opened 2026-07-29
fix(recurringjob): skip snapshot purge while a replica is rebuilding

IggyGG:
## Which issue(s) this PR fixes

Issue longhorn/longhorn#13623

## What this PR does / why we need it

`purgeSnapshots` returns `ActionSnapshotPurge`'s error verbatim. When the engine declines to start a purge because a replica is mid-rebuild:

```
cannot purge snapshots because tcp://<replica>:<port> is rebuilding
```

that error propagates `doSnapshotCleanup` → `startVolumeJob` → the `errgroup` in `recurringJob()`, and `RecurringJobCmd()` turns any non-nil result into `logrus.Fatal`. The process exits, so **every volume that had not yet been processed is silently skipped for that run**.

A rebuild is transient and self-resolving — the next run purges the volume fine. But on a cluster large enough that *some* replica is usually rebuilding, the sweep can go for hours without ever completing: snapshots marked for removal never coalesce, replica directories grow well past their volume's nominal size, and nodes eventually hit `DiskPressure`.

Raising `concurrency` does **not** mitigate it — the workers share one process, so any worker's `Fatal` ends the run.

This PR skips the volume instead. That also makes the function self-consistent: the purge-status poll immediately below **already** tolerates per-replica purge errors, logging them as warnings and returning `nil`. Only the initial trigger was fatal.

### Why match on the message

The engine surfaces this through the proxy as a generic `Unknown` gRPC code, so there is no typed error or distinct status code to key on. The substring is deliberately narrow so genuine failures (timeouts, missing volumes, connection errors) still fail the job loudly. The test pins both directions.

### Observed impact

On a v1.7.2 cluster, the cleanup job failing intermittently over ~8 hours left **1805.7 GB unpurged across 92 snapshots**. One node reached 82% disk and `DiskPressure=True`, which evicted a Postgres pod ~503 times and took the dependent service down for ~2.5 days.

It was slow to diagnose because nothing pointed at snapshots: kubelet's image GC reported `Failed to free sufficient space by deleting unused images (freed 0 bytes)` — images were only 28 GB of the 683 GB used, while Longhorn replicas were 512 GB. A `75Gi` volume occupying **141.8 GB** (`actualSize` ≈ 2× nominal) turned out to be the reliable tell.

## Special notes for your reviewer

- The bug exists identically on `v1.7.2` (`app/recurring_job.go#L505-L508`, pre-refactor) — happy to open a backport if wanted.
- Related but distinct: longhorn/longhorn#12229 covers the *controller's* post-rebuild purge not retrying; this is the *recurring job* aborting its whole sweep.

Verified on linux/amd64 in `golang:1.26` (the package pulls Linux-only vendored code, so it cannot be built or tested on macOS):

- `go build ./app/...` — clean
- `go vet ./app/...` — clean
- `go test ./app/recurringjob/ -run TestIsTransientPurgeRejection` — **PASS**
- The same test with the matcher reverted to pre-fix behaviour — **FAIL**, confirming it guards the behaviour rather than passing vacuously.

### DIFF
--- a/app/recurringjob/volume.go
+++ b/app/recurringjob/volume.go
@@ -342,9 +342,53 @@ func (job *VolumeJob) eventCreate(eventType, eventReason, message string) error 
 	return nil
 }
 
+// isTransientPurgeRejection reports whether err is the engine declining to
+// start a purge because a replica is currently rebuilding.
+//
+// The engine guards ActionSnapshotPurge with a rebuild check and answers:
+//
+//	failed to purge snapshots: rpc error: code = Unknown desc =
+//	tcp://<replica>:<port>: cannot purge snapshots because
+//	tcp://<replica>:<port> is rebuilding
+//
+// That is a temporary, self-resolving condition rather than a fault: the next
+// scheduled run purges the volume once the rebuild finishes. It is matched on
+// the message because the engine surfaces it as a generic Unknown gRPC code
+// through the proxy, so there is no typed error or distinct status code to key
+// on. The substring is deliberately narrow so genuine failures still surface.
+func isTransientPurgeRejection(err error) bool {
+	if err == nil {
+		return false
+	}
+	return strings.Contains(err.Error(), "is rebuilding")
+}
+
 func (job *VolumeJob) purgeSnapshots(volume *longhornclient.Volume, volumeAPI longhornclient.VolumeOperations) error {
 	// Trigger snapshot purge of the volume
 	if _, err := volumeAPI.ActionSnapshotPurge(volume); err != nil {
+		// A rebuilding replica means "not now", not "broken". Returning the
+		// error here fails this volume's job, and because recurringJob()
+		// collects goroutine errors through an errgroup and RecurringJobCmd()
+		// turns any non-nil result into logrus.Fatal, ONE rebuilding volume
+		// aborts the whole run -- every volume not yet processed is silently
+		// skipped for that tick. Raising concurrency does not help, since the
+		// workers share a process.
+		//
+		// On a cluster large enough that some replica is usually rebuilding,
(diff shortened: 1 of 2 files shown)


app/recurringjob/volume.go
```diff
@@ -342,9 +342,53 @@ func (job *VolumeJob) eventCreate(eventType, eventReason, message string) error
...
+//	failed to purge snapshots: rpc error: code = Unknown desc =
+//	tcp://<replica>:<port>: cannot purge snapshots because
+//	tcp://<replica>:<port> is rebuilding
+//
+// That is a temporary, self-resolving condition rather than a fault: the next
+// scheduled run purges the volume once the rebuild finishes. It is matched on
+// the message because the engine surfaces it as a generic Unknown gRPC code
+// through the proxy, so there is no typed error or distinct status code to key
+// on. The substring is deliberately narrow so genuine failures still surface.
+func isTransientPurgeRejection(err error) bool {
```
mantissahz, line 359: Could we rename it to 

```suggestion
func isSnapshotPurgeRejectedByRebuilding(err error) bool {
```
to make it clearer for rebuilding?

davidcheng0922 (commented):
I have some comments here
https://github.com/longhorn/longhorn/issues/13623#issuecomment-5127732203

derekbit:
@IggyGG Could you help fix the conflict? Thanks.

derekbit:
ping @IggyGG

derekbit:
@IggyGG Thanks for your contribution! However, this PR has been pending for quite a while without any further response. Based on the feedback from the other reviewers, we've created a new [PR](https://github.com/longhorn/longhorn-manager/pull/5103) to address the issue. Let’s close this one for now. Thanks again!

PR #5057, opened 2026-07-29
fix: respect node/disk selector when attaching clone target

carterli0407-cell:
Issue longhorn/longhorn#12792


datastore/longhorn.go
```diff
@@ -3817,6 +3817,81 @@ func (s *DataStore) GetReadyNodeDiskForBackingImage(backingImage *longhorn.Backi
...
+		}
+		if !types.IsSelectorsInTags(node.Spec.Tags, volume.Spec.NodeSelector, allowEmptyNodeSelectorVolume) {
+			return false
+		}
+		for diskName, diskStatus := range node.Status.DiskStatus {
+			diskSpec, exists := node.Spec.Disks[diskName]
+			if !exists {
+				continue
+			}
+			if !diskSpec.AllowScheduling {
```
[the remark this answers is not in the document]
carterli0407-cell: Reused scheduler eligibility/capacity checks for clone attach node.

scheduler/replica_scheduler.go
```diff
@@ -1417,6 +1417,60 @@ func (rcs *ReplicaScheduler) FilterNodesSchedulableForVolume(nodes map[string]*l
...
+	if err != nil {
+		return "", err
+	}
+
+	// The target replica is already placed: keep its node (idempotent, capacity-safe).
+	for _, r := range replicas {
+		if r.Spec.NodeID != "" && r.Spec.FailedAt == "" {
+			return r.Spec.NodeID, nil
+		}
+	}
```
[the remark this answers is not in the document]
carterli0407-cell: Updated

controller/volume_clone_controller.go
```diff
@@ -219,7 +223,19 @@ func (vcc *VolumeCloneController) reconcile(volName string) (err error) {
 	// case 1: this volume is target of a clone and the cloning hasn't completed
 	if isCloneTargetActive(vol) {
 		cloningAttachmentTicketID := longhorn.GetAttachmentTicketID(longhorn.AttacherTypeVolumeCloneController, volName)
-		createOrUpdateAttachmentTicket(va, cloningAttachmentTicketID, vol.Status.OwnerID, longhorn.TrueValue, longhorn.AttacherTypeVolumeCloneController)
+
+		preferredNodeID := vol.Status.OwnerID
+		if existingTicket, ok := va.Spec.AttachmentTickets[cloningAttachmentTicketID]; ok && existingTicket.NodeID != "" {
+			preferredNodeID = existingTicket.NodeID
+		}
```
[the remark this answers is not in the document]
carterli0407-cell: updated.

scheduler/replica_scheduler.go
```diff
@@ -1417,6 +1417,84 @@ func (rcs *ReplicaScheduler) FilterNodesSchedulableForVolume(nodes map[string]*l
 	return filteredNodes
 }
 
+func (rcs *ReplicaScheduler) GetReadyNodeForVolumeAttach(volume *longhorn.Volume, preferredNodeID string) (string, error) {
+	// Fast path: the attachment is only truly unconstrained when the volume has no node/disk
+	// selector and no backing image AND empty selectors are allowed to match any node/disk. When
+	// either allow-empty-*-selector-volume setting is false, empty selectors still exclude tagged
+	// nodes/disks (see types.IsSelectorsInTags), so we must run the full candidate selection below.
+	if len(volume.Spec.NodeSelector) == 0 && len(volume.Spec.DiskSelector) == 0 && volume.Spec.BackingImage == "" {
```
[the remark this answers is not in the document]
carterli0407-cell: Updated.

controller/volume_clone_controller.go
```diff
@@ -329,12 +333,14 @@ func (vcc *VolumeCloneController) reconcile(volName string) (err error) {
...
 		if err != nil {
 			return err
 		}
-		if chosenNodeID != "" {
-			createOrUpdateAttachmentTicket(va, cloningAttachmentTicketID, chosenNodeID, longhorn.TrueValue, longhorn.AttacherTypeVolumeCloneController)
+
+		if attachNodeID != "" {
+			createOrUpdateAttachmentTicket(va, cloningAttachmentTicketID, attachNodeID, longhorn.TrueValue, longhorn.AttacherTypeVolumeCloneController)
 			expectedAttachmentTickets[cloningAttachmentTicketID] = true
 		}
```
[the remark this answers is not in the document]
carterli0407-cell: Updated

derekbit:
@carterli0407-cell Please address the comments from Copilot first. Thank you.

controller/volume_clone_controller.go
```diff
@@ -393,8 +397,19 @@ func (vcc *VolumeCloneController) reconcile(volName string) (err error) {
...
+			return err
+		}
+
+		if attachNodeID == "" {
+			vcc.enqueueVolumeAfter(vol, constant.LonghornVolumeAttachmentNotFoundRetryPeriod)
+		} else {
+			createOrUpdateAttachmentTicket(va, cloningAttachmentTicketID, attachNodeID, longhorn.TrueValue, longhorn.AttacherTypeVolumeCloneController)
+		  if longhorn.IsAttachmentTicketSatisfied(cloningAttachmentTicketID, va) {
+			  expectedAttachmentTickets[cloningAttachmentTicketID] = true
+      }
```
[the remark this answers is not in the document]
carterli0407-cell: Fixed.

scheduler/replica_scheduler.go
```diff
@@ -1417,6 +1417,132 @@ func (rcs *ReplicaScheduler) FilterNodesSchedulableForVolume(nodes map[string]*l
...
+		allowEmptyNodeSelectorVolume, err := rcs.ds.GetSettingAsBool(types.SettingNameAllowEmptyNodeSelectorVolume)
+		if err != nil {
+			return "", errors.Wrapf(err, "failed to get %v setting", types.SettingNameAllowEmptyNodeSelectorVolume)
+		}
+		allowEmptyDiskSelectorVolume, err := rcs.ds.GetSettingAsBool(types.SettingNameAllowEmptyDiskSelectorVolume)
+		if err != nil {
+			return "", errors.Wrapf(err, "failed to get %v setting", types.SettingNameAllowEmptyDiskSelectorVolume)
+		}
+		if allowEmptyNodeSelectorVolume && allowEmptyDiskSelectorVolume {
+			attachableNodes, err := rcs.ds.ListReadyNodesWithReadyInstanceManagerRO(volume.Spec.DataEngine)
```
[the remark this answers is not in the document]
carterli0407-cell: Fixed. This is old comment.

controller/volume_clone_controller.go
```diff
@@ -396,12 +399,13 @@ func (vcc *VolumeCloneController) reconcile(volName string) (err error) {
 		if longhorn.IsAttachmentTicketSatisfied(cloningAttachmentTicketID, va) {
 			expectedAttachmentTickets[cloningAttachmentTicketID] = true
 		} else {
-			chosenNodeID, err := pickNodeID(vol, va)
+			var attachNodeID string
+			attachNodeID, err = vcc.scheduler.GetReadyNodeForVolumeAttach(vol, vol.Status.OwnerID)
```
shuo-wu: Is it possible to improve `pickNodeID` logic rather than directly discarding it for this case?

> For strict-local volumes the single replica is then hard-pinned (HardNodeAffinity = vol.Spec.NodeID) to that owner node, which may not satisfy the volume's node/disk selector

Based on your statement, it's mainly for the strict-local case, right? Then, I think refactoring `pickNodeID` would benefit more:
- The clone target volume can pick up the Node that both having selector and local-replica satisfied 
- The clone src volume can pick up the Node containing replicas if strict-local is enabled as well

carterli0407-cell: Thanks for the suggestion. Yes, I agree.
I implemented it before using `pickNodeID()`.
I will review this logic and update as you mentioned.

scheduler/replica_scheduler.go
```diff
@@ -1417,6 +1417,132 @@ func (rcs *ReplicaScheduler) FilterNodesSchedulableForVolume(nodes map[string]*l
 	return filteredNodes
 }
 
+func (rcs *ReplicaScheduler) GetReadyNodeForVolumeAttach(volume *longhorn.Volume, preferredNodeID string) (string, error) {
+	// Fast path: the attachment is only truly unconstrained when the volume has no node/disk
+	// selector and no backing image AND empty selectors are allowed to match any node/disk. When
+	// either allow-empty-*-selector-volume setting is false, empty selectors still exclude tagged
+	// nodes/disks (see types.IsSelectorsInTags), so we must run the full candidate selection below.
+	if volume.Spec.DataLocality != longhorn.DataLocalityStrictLocal && len(volume.Spec.NodeSelector) == 0 && len(volume.Spec.DiskSelector) == 0 && volume.Spec.BackingImage == "" {
```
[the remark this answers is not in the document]
carterli0407-cell: Already updated.

derekbit:
@carterli0407-cell Could you address the comments from @shuo-wu?

scheduler/replica_scheduler.go
```diff
@@ -1437,6 +1437,120 @@ func (rcs *ReplicaScheduler) FilterNodesSchedulableForVolume(nodes map[string]*l
...
+		// Ordinary no-candidate placement: return empty node so caller can clean stale tickets.
+		if reasons := errs.JoinReasons(); reasons != "" {
+			logrus.Debugf("No schedulable node for clone attachment of volume %v: %v", volume.Name, reasons)
+		}
+		return "", nil
+	}
+	candidateNodes := map[string]bool{}
+	for _, disk := range diskCandidates {
+		candidateNodes[disk.NodeID] = true
+	}
```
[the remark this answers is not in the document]
carterli0407-cell: Updated.

controller/volume_clone_controller.go
```diff
@@ -328,53 +331,67 @@ func (vcc *VolumeCloneController) reconcile(volName string) (err error) {
...
-					candidates = []string{nodeID}
-				} else if count == maxCount {
-					candidates = append(candidates, nodeID)
-				}
+
+		needsScheduler := len(v.Spec.NodeSelector) > 0 ||
+			len(v.Spec.DiskSelector) > 0 ||
+			v.Spec.DataLocality == longhorn.DataLocalityStrictLocal ||
+			v.Spec.BackingImage != "" ||
+			v.Spec.CloneMode == longhorn.CloneModeLinkedClone
```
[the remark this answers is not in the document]
carterli0407-cell: Resolved.

scheduler/replica_scheduler.go
```diff
@@ -1437,6 +1437,125 @@ func (rcs *ReplicaScheduler) FilterNodesSchedulableForVolume(nodes map[string]*l
...
+			if preferredNodeID != "" && attachableNodes[preferredNodeID] != nil {
+				return preferredNodeID, nil
+			}
+			chosen := ""
+			for nodeName := range attachableNodes {
+				if chosen == "" || nodeName < chosen {
+					chosen = nodeName
+				}
+			}
+			return chosen, nil
```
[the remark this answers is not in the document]
carterli0407-cell: Fixed.

derekbit:
Hello @carterli0407-cell 
Please address the comments from Copilot first. If a comment is resolved, you can close (resolve) it. Thanks.

derekbit:
Hello @carterli0407-cell
Please address the comments from Copilot first. If a comment is resolved, you can close (resolve) it. Thanks.

controller/volume_clone_controller.go
```diff
@@ -328,66 +331,53 @@ func (vcc *VolumeCloneController) reconcile(volName string) (err error) {
...
-		if attachableNodes[v.Status.OwnerID] != nil {
-			log.Debugf("Picked node %v for volume %v clone attachment: volume owner node", v.Status.OwnerID, v.Name)
-			return v.Status.OwnerID, nil
 		}
 
-		// Otherwise, pick up the 1st node in sorted order for determinism
-		candidates := make([]string, 0, len(attachableNodes))
-		for n := range attachableNodes {
-			candidates = append(candidates, n)
+		chosenNodeID, err = vcc.scheduler.GetReadyNodeForVolumeAttach(v, preferredNodeID)
```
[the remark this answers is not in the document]
carterli0407-cell: https://github.com/longhorn/longhorn/issues/13681

scheduler/replica_scheduler.go
```diff
@@ -1437,6 +1437,147 @@ func (rcs *ReplicaScheduler) FilterNodesSchedulableForVolume(nodes map[string]*l
...
 }
 
+func (rcs *ReplicaScheduler) GetReadyNodeForVolumeAttach(volume *longhorn.Volume, preferredNodeID string) (string, error) {
+	// Fast path: unconstrained non-strict-local volume. Even here, the returned node must be
+	// attachable (Ready with a running instance manager).
+	if volume.Spec.DataLocality != longhorn.DataLocalityStrictLocal &&
+		volume.Spec.CloneMode != longhorn.CloneModeLinkedClone &&
+		len(volume.Spec.NodeSelector) == 0 &&
+		len(volume.Spec.DiskSelector) == 0 &&
+		volume.Spec.BackingImage == "" {
```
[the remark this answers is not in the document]
carterli0407-cell: Added.

controller/volume_clone_controller.go
derekbit: Should we use `ListReadyNodesWithReadyInstanceManagerRO`?

controller/volume_clone_controller.go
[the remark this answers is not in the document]
derekbit: I think this is a valid concern. 

@carterli0407-cell Could you split the node selection so that scheduler candidacy (selector enforcement) is only applied to the clone target (case 1), while source attachments (cases 2 and 3) preserve the source's current node or fall back to any attachable node hosting a healthy replica without going through `GetReadyNodeForVolumeAttach`? 

WDYT?

scheduler/replica_scheduler.go
```diff
@@ -1437,6 +1437,148 @@ func (rcs *ReplicaScheduler) FilterNodesSchedulableForVolume(nodes map[string]*l
...
+			return "", errors.Wrapf(err, "failed to get %v setting", types.SettingNameAllowEmptyDiskSelectorVolume)
+		}
+		if allowEmptyNodeSelectorVolume && allowEmptyDiskSelectorVolume {
+			attachableNodes, err := rcs.ds.ListReadyNodesWithReadyInstanceManagerRO(volume.Spec.DataEngine)
+			if err != nil {
+				return "", err
+			}
+			chosen := ""
+			for nodeName := range attachableNodes {
+				imageReady, err := rcs.ds.CheckDataEngineImageReadiness(volume.Spec.Image, volume.Spec.DataEngine, nodeName)
```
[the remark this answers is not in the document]
derekbit: This is also fair.

controller/volume_clone_controller.go
```diff
@@ -328,66 +331,53 @@ func (vcc *VolumeCloneController) reconcile(volName string) (err error) {
...
+				for nodeID, count := range nodeCount {
+					if count > maxCount {
+						maxCount = count
+						candidates = []string{nodeID}
+					} else if count == maxCount {
+						candidates = append(candidates, nodeID)
+					}
 				}
+				sort.Strings(candidates)
+				preferredNodeID = candidates[0]
```
derekbit: Can we add a helper function for the codes and add a unit test for the helper functino?

scheduler/replica_scheduler.go
```diff
@@ -1437,6 +1437,148 @@ func (rcs *ReplicaScheduler) FilterNodesSchedulableForVolume(nodes map[string]*l
...
+			if err != nil {
+				return "", errors.Wrapf(err, "failed to check data engine image readiness on node %v", nodeID)
+			}
+			if !imageReady {
+				continue
+			}
+			if reused == "" || nodeID < reused {
+				reused = nodeID
+			}
+		}
```
derekbit: Can we add a helper function for searching `reuse`?

scheduler/replica_scheduler.go
```diff
@@ -1437,6 +1437,148 @@ func (rcs *ReplicaScheduler) FilterNodesSchedulableForVolume(nodes map[string]*l
...
+				if !imageReady {
+					continue
+				}
+				if nodeName == preferredNodeID {
+					return preferredNodeID, nil
+				}
+				if chosen == "" || nodeName < chosen {
+					chosen = nodeName
+				}
+			}
```
derekbit: Can we add a helper function for searching `reuse`?

derekbit (commented):
In general, the solution makes sense to me.

If possible, could you create a ROBOT e2e test case to validate your implementation and attach the link to Jenkins job?

Thanks.

controller/volume_clone_controller.go
shuo-wu: A ready node may not be necessary for volume attachment, e.g., the longhorn-manager pod is not healthy while the instance manager pod is still running, and the volume is already attached to this node.

controller/volume_clone_controller.go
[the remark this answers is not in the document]
carterli0407-cell: Thanks @derekbit - agreed, this is a valid concern.
`GetReadyNodeForVolumeAttach` should only be used for the clone target (case 1), where node/disk selector candidacy must be respected.
For source attachments (cases 2 and 3), we should keep the previous behavior.
I'll split the node selection accordingly and update the PR.

controller/volume_clone_controller.go
carterli0407-cell: I think we should keep `ListNodesWithReadyInstanceManagerRO` here.
`ListReadyNodesWithReadyInstanceManagerRO` remains appropriate for clone-target placement via `GetReadyNodeForVolumeAttach`, where we do need a Ready/schedulable node.

scheduler/replica_scheduler.go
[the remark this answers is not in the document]
carterli0407-cell: Added. Thanks.

controller/volume_clone_controller.go
carterli0407-cell: I added helper function and unit test for it.

scheduler/replica_scheduler.go
carterli0407-cell: Added. Thanks.

scheduler/replica_scheduler.go
carterli0407-cell: Added.

scheduler/replica_scheduler.go
```diff
@@ -1437,6 +1437,151 @@ func (rcs *ReplicaScheduler) FilterNodesSchedulableForVolume(nodes map[string]*l
 	return filteredNodes
 }
 
+func (rcs *ReplicaScheduler) pickReusableNodeForVolumeAttach(
```
shuo-wu: 1. NIT: Maybe you can filter out `candidateNodes` based on `attachableNodes`. Then this function wouldn't need the `attachableNodes` input.
2. I do not get what "Reusable" means. It mainly picks up a node from the candidates list with a preference

scheduler/replica_scheduler.go
```diff
@@ -1437,6 +1437,151 @@ func (rcs *ReplicaScheduler) FilterNodesSchedulableForVolume(nodes map[string]*l
...
+		}
+		allowEmptyDiskSelectorVolume, err := rcs.ds.GetSettingAsBool(types.SettingNameAllowEmptyDiskSelectorVolume)
+		if err != nil {
+			return "", errors.Wrapf(err, "failed to get %v setting", types.SettingNameAllowEmptyDiskSelectorVolume)
+		}
+		if allowEmptyNodeSelectorVolume && allowEmptyDiskSelectorVolume {
+			attachableNodes, err := rcs.ds.ListReadyNodesWithReadyInstanceManagerRO(volume.Spec.DataEngine)
+			if err != nil {
+				return "", err
+			}
```
shuo-wu, line 1503: I am not sure whether we still need this `attachableNodes` here and below. 
The main case we may need to consider is what I mentioned above: the Longhorn manager pod is not ready, while the IM pod on this node is still running. IMO, it's better to consider this kind of node as attachable since there may be running replicas on it.

scheduler/replica_scheduler.go
```diff
@@ -1437,6 +1437,151 @@ func (rcs *ReplicaScheduler) FilterNodesSchedulableForVolume(nodes map[string]*l
...
+		}
+		// For strict-local, do not move away from existing placement if it is currently unavailable.
+		if volume.Spec.DataLocality == longhorn.DataLocalityStrictLocal {
+			return "", nil
+		}
+		// For non-strict-local, fall through to candidate selection.
+	}
+
+	// No replica scheduled on a reusable attachable node yet: ask scheduler for real candidates.
+	syntheticReplica := &longhorn.Replica{
```
shuo-wu: I am thinking about how this can happen. IIUC, all replicas should be scheduled before requesting a VA ticket. 
Then, if things fall into this, it means all non-failed replicas are not in ready IM pods. (Especially considering what I mentioned in https://github.com/longhorn/longhorn-manager/pull/5057/changes#r3876818783.) Is there any exception?

scheduler/replica_scheduler.go
carterli0407-cell: Good point, thanks.
On the signature: agreed. I'll refactor the helper to take only `candidateNodes` and filter internally against nodes with a running IM.
I will rename it to something like `pickPreferredAttachableNodeForVolumeAttach` (open to better names if you have).

scheduler/replica_scheduler.go
carterli0407-cell: For clone **target** placement (`GetReadyNodeForVolumeAttach`), I think `ListReadyNodesWithReadyInstanceManagerRO` is intentional.
This path is only used from the clone controller for the **target** (case 1). We need a node that is both scheduler-eligible and currently attachable for engine startup - Ready + running IM is the right bar for picking a *new* attach node with node/disk selectors.
The edge case you mentioned (longhorn-manager unhealthy but IM still running, volume already attached) is handled separately for **source** attachments in `volume_clone_controller.go`, where we deliberately use `ListNodesWithReadyInstanceManagerRO` and do not go through `GetReadyNodeForVolumeAttach`.
So:
- source attach (cases 2/3): preserve current/attachable node, no Ready-node requirement
- target attach (case 1): enforce selector candidacy + Ready/IM attachability
If we used `ListNodesWithReadyInstanceManagerRO` in `GetReadyNodeForVolumeAttach`, we could pick a non-Ready node for the clone target and still fail engine start / scheduling which is what #12792 is trying to avoid.

scheduler/replica_scheduler.go
carterli0407-cell: You're right that in the steady state, a clone target should have its replica(s) scheduled before attachment completes. 
But the clone controller can create/update the VA ticket earlier, while placement is still in progress or temporarily invalid.

A few cases reach the `FindDiskCandidates` path:
1. **No placed replica yet** - brand-new clone volume; `VolumeCloneController` reconciles before `ReplicaController` has assigned `spec.nodeID` (replica exists with empty nodeID, or no healthy placed replica yet).
2. **Wrong / stale placement during clone** - this is essentially #12792: volume has `spec.nodeID` or a placed replica on a node that does not satisfy node/disk selectors. Reusing that node is wrong; we need scheduler candidacy to find the correct hosting node.

So this block is the "no reusable placed node" path: ask the scheduler where the clone target *should* attach, respecting selectors, instead of guessing from stale `nodeID` or a non-attachable placed replica.

shuo-wu: 1. For regular brand new volumes with unscheduled replicas, IMO, it's better to ask the VA-related controllers to wait for the scheduling handling, rather than repeating the scheduling logic by themselves
2. Do we need to consider the existing clone-in-progress volumes on wrong nodes?  The clone is a one-time operation for each volume. I think we just need to focus on preventing the wrong placement from happening again. Besides, if the replicas are already on unsatisfied nodes/disks, how can this logic help rebuild replicas in the correct places?

cc @derekbit WDYT?

scheduler/replica_scheduler.go
shuo-wu: > we could pick a non-Ready node for the clone target and still fail engine start / scheduling which is what #12792 is trying to avoid

If the only **selector-satisfied** node contains a not-ready manager pod, a running IM pod, and a running replica on it. Your new logic will exclude this node, which then makes things more complicated.
Would you elaborate on this with another example?

carterli0407-cell: For example, only node-b has hosting, replica is on node-b, manager on node-b is restarting (ManagerPodDown) but IM/replica are still running. We requeue until node-b is Ready=True, then attach there. We do not fall back to other nodes. This is intentional and consistent with migration logic that also waits on ManagerPodDown. The alternative (attach while manager is down, or attach on a non-selector node) is riskier and was the source of #12792.

scheduler/replica_scheduler.go
carterli0407-cell: 1. Agreed. This PR does not schedule replicas.
VolumeCloneController only decides where to create the clone-target attach ticket.
FindDiskCandidates in this PR is not “schedule a replica.” It is only a helper to pick an attach node when reuse is not possible yet. But I agree with your direction: the main behavior should be wait for scheduling, not do scheduling again in VA logic.
2. No. Clone is one-time per volume. This PR only prevents the wrong attach from happening going forward. It is not a migration/repair for volumes that already have bad placement.
This logic does not move or rebuild replicas. Replica relocation is handled by scheduling/eviction/rebuild flows.
What #12792 actually is:
- replica is already on the correct selector-matching node
- but clone attach ticket was created on a different node
- attach and replica placement disagree -> clone fails
This PR fixes that attach mismatch, not wrong replica placement.

shuo-wu: Then, in other words, this PR is responsible for:
1. Blindly wait for replicas to be scheduled (probably by checking the condition `Scheduled`). No need to use an unreal `syntheticReplica` to predict which node is suitable for scheduling. 
2. For strict-local volumes, pick up a node containing a valid replica for the VA 

Right?

Then, the fix for this issue can be simplified to: add the replica nodes as the top preference for strict-local volumes to the original `pickNodeID` function

scheduler/replica_scheduler.go
shuo-wu: In your case, what if there is one more node satisfying all conditions (node/disk selectors) but has no replica? Then, your logic will exclude node-b (since its manager pod down) and pick up this new node (by using the `syntheticReplica`), right? This attachment will lead to an unnecessary rebuilding of the strict-local volumes on this new node.

carterli0407-cell: @shuo-wu ,
Your concern would apply only if we fell through to FindDiskCandidates while a placed replica already exists. For strict-local we already block that. If the only placed replica is on node-b and node-b is temporarily not Ready, we return empty and wait. We do not pick another selector-matching node via syntheticReplica, so we do not cause an unnecessary rebuild.

Relevant code:
```
if len(placedNodes) > 0 {
	reused, err := rcs.pickPreferredAttachableNodeForVolumeAttach(volume, placedNodes, preferredNodeID)
	...
	if reused != "" {
		return reused, nil
	}
	// For strict-local, do not move away from existing placement if it is currently unavailable.
	if volume.Spec.DataLocality == longhorn.DataLocalityStrictLocal {
		return "", nil
	}
```
So for strict-local we wait for node-b, and we do not attach on node-c or trigger an unnecessary rebuild.
WDYT?

scheduler/replica_scheduler.go
carterli0407-cell: @shuo-wu , mostly right, with one caveat.

For strict-local, replicas are scheduled only after attach (HardNodeAffinity = Spec.NodeID). Waiting for Scheduled before the clone VA ticket would deadlock.
“Prefer replica nodes in pickNodeID” alone is not enough for #12792; we still need a selector-aware attach choice on the first attach.

scheduler/replica_scheduler.go
shuo-wu: I see. Line 1535 prevents the unnecessary rebuild in my case. Then, I am fine with `ListNodesWithReadyInstanceManagerRO`

scheduler/replica_scheduler.go
shuo-wu: > For strict-local, replicas are scheduled only after attach (HardNodeAffinity = Spec.NodeID). 

OK. I missed this case. Then, can we still make `pickNodeID ` a general function? My current PR https://github.com/longhorn/longhorn-manager/pull/5140/changes/29ea54956ba5afa446d4eca77df8549dddf8298a will use it for the volume restore auto-attachment.

carterli0407-cell: @shuo-wu , 
Yes. Keeping pickAttachmentTicketNodeID general is the right approach for your restore PR.

However, for #12792, we should extend that same helper rather than adding a separate scheduler path: after CurrentNodeID / ticket majority, prefer placed replica nodes, and if none exist yet prefer a selector-compatible attachable node before falling back to OwnerID / first ready. That fixes strict-local first-attach without deadlocking on Scheduled, and keeps clone/restore on one shared picker.

shuo-wu: Is it possible for you to make `pickNodeID` a general function after modifying it then continue extending it inside volume_clone_controller reconcile?

carterli0407-cell: @shuo-wu ,
Yes, it's possible, use it directly for clone-source tickets, and only extend it inside reconcile for clone-target tickets.

carterli0407-cell:
e2e test passed for this ticket:
https://ci.longhorn.io/job/longhorn-e2e-test-pull-request-check/526/

scheduler/replica_scheduler.go
```diff
@@ -1437,6 +1437,137 @@ func (rcs *ReplicaScheduler) FilterNodesSchedulableForVolume(nodes map[string]*l
...
+		Spec: longhorn.ReplicaSpec{
+			InstanceSpec: longhorn.InstanceSpec{
+				VolumeName: volume.Name,
+				VolumeSize: volume.Spec.Size,
+				DataEngine: volume.Spec.DataEngine,
+				Image:      volume.Spec.Image,
+			},
+		},
+	}
+	diskCandidates, errs := rcs.FindDiskCandidates(syntheticReplica, replicas, volume)
```
shuo-wu, line 1556: NIT: Add a comment that explains why we need this replica scheduling foreknowledge
```
For strict-local, replicas are scheduled only after attach (HardNodeAffinity = Spec.NodeID).

```

shuo-wu (dismissed):
In general LGTM

scheduler/replica_scheduler.go
carterli0407-cell: Added a comment. Thanks.

derekbit (approved):

derekbit:
@mergify backport v1.13.x

PR #5062, opened 2026-08-04
fix(clone): validate clone attachment node has ready IM before creating ticket

shuo-wu:

Issue https://github.com/longhorn/longhorn/issues/13639

The volume clone controller previously used vol.Status.OwnerID as the attachment node for clone VA tickets. OwnerID is the reconciler owner which can be a not-ready node or a node without a running instance manager, causing clone volumes to be attached to invalid nodes.

Add ListReadyNodesWithReadyInstanceManagerRO to the datastore that returns only nodes which are both Longhorn-ready and have a running instance manager for the given data engine.

Refactor pickNodeID with a clear priority chain:
1. CurrentNodeID (where the volume is actually attached)
2. Majority node among highest-priority existing VA tickets
3. OwnerID (reconciler node)
4. First available ready node (sorted for determinism)

### DIFF
--- a/controller/volume_clone_controller.go
+++ b/controller/volume_clone_controller.go
@@ -3,6 +3,7 @@ package controller
 import (
 	"fmt"
 	"reflect"
+	"sort"
 	"time"
 
 	"github.com/cockroachdb/errors"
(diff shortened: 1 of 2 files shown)


controller/volume_clone_controller.go
```diff
@@ -254,30 +255,116 @@ func (vcc *VolumeCloneController) reconcile(volName string) (err error) {
 
 	expectedAttachmentTickets := make(map[string]bool)
 
+	var attachableNodes map[string]*longhorn.Node
+	log := getLoggerForVolume(vcc.logger, vol)
+	pickNodeID := func(v *longhorn.Volume, va *longhorn.VolumeAttachment) (chosenNodeID string, err error) {
```
c3y1huang, line 260: What happens to an in-progress copy if pickNodeID == ""? For example, if the attached node's Ready condition briefly flips and no other nodes passed the filter at that moment? IIUC, the live ticket gets deleted -> volume gets detached -> clone failed and consumes retry attempts. If that is the case, I am slightly confused, sine it seem the [VA controller treats mid-copy as non-interruptable](https://github.com/longhorn/longhorn-manager/blob/v1.12.1-rc3/controller/volume_attachment_controller.go#L673-L674).

controller/volume_clone_controller.go
```diff
@@ -254,30 +255,116 @@ func (vcc *VolumeCloneController) reconcile(volName string) (err error) {
...
+		for n := range attachableNodes {
+			candidates = append(candidates, n)
+		}
+		sort.Strings(candidates)
+		if len(candidates) > 0 {
+			log.Debugf("Picked node %v for volume %v clone attachment: first available ready node", candidates[0], v.Name)
+			return candidates[0], nil
+		}
+
+		log.Warnf("Cannot find a valid node for volume %v clone attachment, will clean up stale tickets and retry", v.Name)
```
c3y1huang, line 325: > will clean up stale tickets and retry

What schedules the retry?

controller/volume_clone_controller.go
```diff
@@ -254,30 +255,116 @@ func (vcc *VolumeCloneController) reconcile(volName string) (err error) {
...
+		if attachableNodes == nil {
+			attachableNodes, err = vcc.ds.ListReadyNodesWithReadyInstanceManagerRO(v.Spec.DataEngine)
+			if err != nil {
+				return "", err
+			}
+		}
+
+		// The CurrentNodeID holds the 1st priority if it is valid.
+		// The corresponding ticket is implicitly satisfied.
+		if attachableNodes[v.Status.CurrentNodeID] != nil {
```
c3y1huang, line 270: What happens when the volume is attached and healthy on CurrentNodeID, but that node briefly drops out of attachableNodes? Is moving the ticket in this window intended?

davidcheng0922 (approved):
Overall LGTM

datastore/longhorn.go
```diff
@@ -3738,6 +3738,28 @@ func (s *DataStore) ListReadyNodesRO() (map[string]*longhorn.Node, error) {
...
+		return nil, err
+	}
+
+	result := make(map[string]*longhorn.Node, len(readyNodes))
+	for nodeName, node := range readyNodes {
+		imMap, err := s.ListInstanceManagersByNodeRO(nodeName, longhorn.InstanceManagerTypeAllInOne, dataEngine)
+		if err != nil {
+			return nil, err
+		}
+		for _, im := range imMap {
```
c3y1huang, line 3753: Is any-image IM sufficient for v1? Will this potentially cause issue is there are old-image IM exist?

derekbit (approved):
LGTM

derekbit:
@mergify backport v1.12.x

controller/volume_clone_controller.go
```diff
@@ -254,30 +255,116 @@ func (vcc *VolumeCloneController) reconcile(volName string) (err error) {
...
+					maxCount = count
+					candidates = []string{nodeID}
+				} else if count == maxCount {
+					candidates = append(candidates, nodeID)
+				}
+			}
+			sort.Strings(candidates)
+			log.Debugf("Picked node %v for volume %v clone attachment: majority node among highest-priority tickets", candidates[0], v.Name)
+			return candidates[0], nil
+		}
```
c3y1huang, line 306: IIUC this duplicates the [VA ticket election](https://github.com/longhorn/longhorn-manager/blob/v1.12.1-rc3/controller/volume_attachment_controller.go#L731) with different rule. So the clone ticket can land on a node that VA doesn't pick, what is the expected behavior when this drifts?

controller/volume_clone_controller.go
```diff
@@ -254,30 +255,116 @@ func (vcc *VolumeCloneController) reconcile(volName string) (err error) {
...
+			}
 		}
 	}
 
 	// case 3: this volume is source of a linked-clone target that needs rebuild
 	// (post-initial-clone: clone completed or awaiting healthy, and has replicas pending rebuild)
-	var readyNodes map[string]*longhorn.Node
-	chosenNodeID := vol.Status.CurrentNodeID
+	// Use a single node for all clone tickets to avoid attaching the source to multiple nodes.
+	srcNodeID = ""
```
c3y1huang, line 367: Why re-pick for case 3 if case 2 already picked?

derekbit:
@shuo-wu I merged it too early. Could you check the comments from @c3y1huang? Thanks

controller/volume_clone_controller.go
shuo-wu: retry actually means reconcile later. I can update the wording if it's pretty confusing

controller/volume_clone_controller.go
shuo-wu: Case 3 is for:
1. Linked clone only
2. Post-clone phase, which guarantees that the linked clone volumes can be healthy before removing the VA ticket for the src volume

controller/volume_clone_controller.go
shuo-wu: Here, I prefer to ask the clone ticket to directly use a node that is already picked by most of the top-priority tickets. This means that **the clone ticket can be satisfied and handled as soon as possible**.

If this part is executed, it typically means the volume is currently not attached, or the currently attached Node suddenly becomes unavailable. Otherwise, we will directly choose `v.Status.CurrentNodeID`

controller/volume_clone_controller.go
shuo-wu: The most likely reason for this case I can find is the Longhorn Manager Pod down case. The old-image IM pod's existence may be another issue as well.

I am thinking whether it's better to:
- Either list the nodes that contain the default running IM pod & kube node ready instead.
- Or trust `CurrentNodeID`  unconditionally

WDYT?

controller/volume_clone_controller.go
shuo-wu: Nice catch. 
To avoid a potential interruption:
1. Call `createOrUpdateAttachmentTicket` only when the existing ticket is not satisfied. ==> This avoids suddenly reattaching the volume to another node if the clone ticket is satisfied (the clone is probably in progress at this moment)
2. Set `expectedAttachmentTickets[cloningAttachmentTicketID]` as long as the case is matched & the ticket is satisfied or newly set/updated.

PR #5063, opened 2026-08-04
fix(setting): clarify V2 guaranteed CPU must cover assigned SPDK CPU cores

davidcheng0922:

https://github.com/longhorn/longhorn/issues/13585

### DIFF
--- a/types/setting.go
+++ b/types/setting.go
@@ -1314,7 +1314,11 @@ var (
 			"  - One more set of instance manager pods may need to be deployed when the Longhorn system is upgraded. If current available CPUs of the nodes are not enough for the new instance manager pods, you need to detach the volumes using the oldest instance manager pods so that Longhorn can clean up the old pods automatically and release the CPU resources. And the new pods with the latest instance manager image will be launched then. \n\n" +
 			"  - This global setting will be ignored for a node if the field \"InstanceManagerCPURequest\" on the node is set. \n\n" +
 			"  - After this setting is changed, the instance manager pod using this global setting will be automatically restarted without instances running on the instance manager. \n\n" +
-			"  - For the V2 Data Engine, the Storage Performance Development Kit (SPDK) target daemon inside each instance manager pod uses one or more dedicated CPU cores. Setting a minimum CPU usage is critical to maintaining stability during periods of high node load.",
+			"  - For the V2 Data Engine, the Storage Performance Development Kit (SPDK) target daemon inside each instance manager pod uses one or more dedicated CPU cores. " +
+			"It is necessary that the effective instance manager pod CPU request reserve at least as many CPU cores as configured by the V2 Data Engine CPU Mask based on the node's allocatable CPU capacity. " +
+			"For example, reserving 2 CPU cores on a node with 8 allocatable CPUs requires an effective CPU request of at least 25% or 2 CPU cores. " +
+			"If the V2 Data Engine Number of CPU Cores setting is specified, Longhorn automatically raises the pod CPU request to at least that many CPU cores. " +
+			"Setting a minimum CPU usage is critical to maintaining stability during periods of high node load.",
 		Category:           SettingCategoryDangerZone,
 		Type:               SettingTypeFloat,
 		Required:           true,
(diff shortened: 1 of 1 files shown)


types/setting.go
```diff
@@ -1314,7 +1314,7 @@ var (
...
-			"  - For the V2 Data Engine, the Storage Performance Development Kit (SPDK) target daemon inside each instance manager pod uses one or more dedicated CPU cores. Setting a minimum CPU usage is critical to maintaining stability during periods of high node load.",
+			"  - For the V2 Data Engine, the Storage Performance Development Kit (SPDK) target daemon inside each instance manager pod uses one or more dedicated CPU cores. It is strongly recommended that the effective instance manager pod CPU request reserve at least as many CPU cores as configured by the V2 Data Engine CPU Mask based on the node's allocatable CPU capacity. For example, reserving 2 CPU cores on a node with 8 allocatable CPUs requires an effective CPU request of at least 25% or 2 CPU cores. If the V2 Data Engine Number of CPU Cores setting is specified, Longhorn automatically raises the pod CPU request to at least that many CPU cores. Setting a minimum CPU usage is critical to maintaining stability during periods of high node load.",
```
```suggestion
			"  - For the V2 Data Engine, the Storage Performance Development Kit (SPDK) target daemon inside each instance manager pod uses one or more dedicated CPU cores. It is necessary that the effective instance manager pod CPU request reserve at least as many CPU cores as configured by the V2 Data Engine CPU Mask based on the node's allocatable CPU capacity. For example, reserving 2 CPU cores on a node with 8 allocatable CPUs requires an effective CPU request of at least 25% or 2 CPU cores. If the V2 Data Engine Number of CPU Cores setting is specified, Longhorn automatically raises the pod CPU request to at least that many CPU cores. Setting a minimum CPU usage is critical to maintaining stability during periods of high node load.",
```

types/setting.go
```diff
@@ -1849,7 +1849,7 @@ var (
...
...mance Development Kit (SPDK) target daemon runs. The daemon is deployed in each Instance Manager pod. It is strongly recommended that the effective V2 instance manager pod CPU request reserve at least the same number of CPU cores configured here based on the node's allocatable CPU capacity. If the field `InstanceManagerCPURequest` on the node is set, it overrides the global Guaranteed Instance Manager CPU setting. If the effective CPU request is lower than the CPU cores assigned here, the V2 instance manager may become unstable under load. A minimum of 2 CPU cores is recommended. SPDK uses a busy-polling reactor model where the master reactor handles both I/O polling and management RPCs. When only a single core is assigned, heavy I/O workloads can delay or starve RPC processing, resulting in increased latency, timeout events, and operational instability. Assigning 2 or more cores allows I/O and management tasks to run on separate reactors, improving responsiveness and operational stability. Accepts either hexadecimal CPU masks (for example, 0x3 or 0xff) or CPU list format (for example, 0-1,2,5). CPU lists are automatically converted to hexadecimal masks. The default value is 0x3.",
```
```suggestion
		Description:        "Applies only to the V2 Data Engine. If the Data Engine CPU Core Number setting is specified, this setting will be ignored. It specifies the CPU cores on which the Storage Performance Development Kit (SPDK) target daemon runs. The daemon is deployed in each Instance Manager pod. It is necessary that the effective V2 instance manager pod CPU request reserve at least the same number of CPU cores configured here based on the node's allocatable CPU capacity. If the field `InstanceManagerCPURequest` on the node is set, it overrides the global Guaranteed Instance Manager CPU setting. If the effective CPU request is lower than the CPU cores assigned here, the V2 instance manager may become unstable under load. A minimum of 2 CPU cores is recommended. SPDK uses a busy-polling reactor model where the master reactor handles both I/O polling and management RPCs. When only a single core is assigned, heavy I/O workloads can delay or starve RPC processing, resulting in increased latency, timeout events, and operational instability. Assigning 2 or more cores allows I/O and management tasks to run on separate reactors, improving responsiveness and operational stability. Accepts either hexadecimal CPU masks (for example, 0x3 or 0xff) or CPU list format (for example, 0-1,2,5). CPU lists are automatically converted to hexadecimal masks. The default value is 0x3.",
```

derekbit (commented):
Use `necessary` WDYT?

types/setting.go
davidcheng0922:
```
	SettingDefinitionGuaranteedInstanceManagerCPU = SettingDefinition{
		DisplayName: "Guaranteed Instance Manager CPU",
		Description: "...",
		Category:           SettingCategoryDangerZone,
		Type:               SettingTypeFloat,
		Required:           true,
		ReadOnly:           false,
		DataEngineSpecific: true,
		Default:            fmt.Sprintf("{%q:\"12\",%q:\"12\"}", longhorn.DataEngineTypeV1, longhorn.DataEngineTypeV2),
		ValueFloatRange: map[string]float64{
			ValueFloatRangeMinimum: 0,
			ValueFloatRangeMaximum: 40,
		},
	}
```

I used `strongly recommended` because we currently cap this setting at `40%`. If a user dedicates half of the node CPUs to V2, describing this as strictly necessary would imply a configuration that the setting cannot always satisfy. Do we want to raise the maximum limit to cover those cases?

derekbit: Oh, you are right!
I think `necessary` might be mandatory to avoid any potential issues.
Can we create a ticket to raise the maximum limit? We can improve it in v1.13.0.
WDYT

davidcheng0922: Sure

davidcheng0922: Use `necessary` instead of `strongly recommended`

derekbit (dismissed):
LGTM.

derekbit:
@davidcheng0922 Build failed.

derekbit:
@mergify backport v1.12.x

derekbit (approved):

PR #5065, opened 2026-08-04
fix(clone): avoid unnecessary clone tickets changes

shuo-wu:

The comments from @c3y1huang in PR https://github.com/longhorn/longhorn-manager/pull/5062 mentioned that the current clone VA logic may update/remove VA tickets unexpectedly, especially when the clone ticket is already satisfied. 
Besides, it's better to consider the nodes having default IM pods when attaching the volume

Issue: https://github.com/longhorn/longhorn/issues/13639

### DIFF
--- a/controller/volume_clone_controller.go
+++ b/controller/volume_clone_controller.go
@@ -323,6 +323,7 @@ func (vcc *VolumeCloneController) reconcile(volName string) (err error) {
 		}
 
 		log.Warnf("Cannot find a valid node for volume %v clone attachment", v.Name)
+		vcc.enqueueVolumeAfter(v, constant.LonghornVolumeAttachmentNotFoundRetryPeriod)
 		return "", nil
 	}
 
--- a/controller/volume_clone_controller.go
+++ b/controller/volume_clone_controller.go
@@ -259,7 +259,7 @@ func (vcc *VolumeCloneController) reconcile(volName string) (err error) {
 	log := getLoggerForVolume(vcc.logger, vol)
 	pickNodeID := func(v *longhorn.Volume, va *longhorn.VolumeAttachment) (chosenNodeID string, err error) {
 		if attachableNodes == nil {
-			attachableNodes, err = vcc.ds.ListReadyNodesWithReadyInstanceManagerRO(v.Spec.DataEngine)
+			attachableNodes, err = vcc.ds.ListNodesWithReadyInstanceManagerRO(v.Spec.DataEngine)
 			if err != nil {
 				return "", err
 			}
--- a/controller/volume_clone_controller.go
+++ b/controller/volume_clone_controller.go
@@ -322,20 +322,24 @@ func (vcc *VolumeCloneController) reconcile(volName string) (err error) {
 			return candidates[0], nil
 		}
 
-		log.Warnf("Cannot find a valid node for volume %v clone attachment, will clean up stale tickets and retry", v.Name)
+		log.Warnf("Cannot find a valid node for volume %v clone attachment", v.Name)
 		return "", nil
 	}
 
 	// case 1: this volume is target of a clone and the cloning hasn't completed
 	if isCloneTargetActive(vol) {
 		cloningAttachmentTicketID := longhorn.GetAttachmentTicketID(longhorn.AttacherTypeVolumeCloneController, volName)
-		chosenNodeID, err := pickNodeID(vol, va)
-		if err != nil {
-			return err
-		}
-		if chosenNodeID != "" {
-			createOrUpdateAttachmentTicket(va, cloningAttachmentTicketID, chosenNodeID, longhorn.TrueValue, longhorn.AttacherTypeVolumeCloneController)
+		if longhorn.IsAttachmentTicketSatisfied(cloningAttachmentTicketID, va) {
 			expectedAttachmentTickets[cloningAttachmentTicketID] = true
+		} else {
+			chosenNodeID, err := pickNodeID(vol, va)
+			if err != nil {
+				return err
+			}
+			if chosenNodeID != "" {
+				createOrUpdateAttachmentTicket(va, cloningAttachmentTicketID, chosenNodeID, longhorn.TrueValue, longhorn.AttacherTypeVolumeCloneController)
+				expectedAttachmentTickets[cloningAttachmentTicketID] = true
+			}
 		}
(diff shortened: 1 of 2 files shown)


controller/volume_clone_controller.go
```diff
@@ -265,9 +265,10 @@ func (vcc *VolumeCloneController) reconcile(volName string) (err error) {
 			}
 		}
 
-		// The CurrentNodeID holds the 1st priority if it is valid.
-		// The corresponding ticket is implicitly satisfied.
-		if attachableNodes[v.Status.CurrentNodeID] != nil {
+		// The CurrentNodeID holds the 1st priority unconditionally.
+		// If the volume is already running there, that is the strongest signal the node works,
+		// even if it transiently drops from attachableNodes (e.g., manager pod restart).
+		if v.Status.CurrentNodeID != "" {
```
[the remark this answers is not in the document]
shuo-wu: @c3y1huang 
I am thinking we really need to unconditionally set the ticket Node ID to the current Node ID.
If this clone ticket is already satisfied, the 1st commit `7245f58a2a19812c72c1302568987cf45f413660` would help avoid changing it
If this clone ticket is not satisfied or is not even set yet, why not use a healthier Node ID instead?

controller/volume_clone_controller.go
```diff
@@ -322,20 +323,24 @@ func (vcc *VolumeCloneController) reconcile(volName string) (err error) {
 			return candidates[0], nil
 		}
 
-		log.Warnf("Cannot find a valid node for volume %v clone attachment, will clean up stale tickets and retry", v.Name)
+		log.Warnf("Cannot find a valid node for volume %v clone attachment", v.Name)
 		return "", nil
```
c3y1huang, line 390: The retry is removed from log, but IIUC, we still expect this to retry on the next reconcile. If that is the case, when this returns for a not-yet-satisfied clone -> ticket gets removed -> reconcile returns nil -> key forgotten. What re-enqueues the volume?

controller/volume_clone_controller.go
```diff
@@ -373,16 +382,21 @@ func (vcc *VolumeCloneController) reconcile(volName string) (err error) {
...
-		if srcNodeID != "" {
-			cloningAttachmentTicketID := longhorn.GetAttachmentTicketID(longhorn.AttacherTypeVolumeCloneController, v.Name)
-			createOrUpdateAttachmentTicket(va, cloningAttachmentTicketID, srcNodeID, longhorn.AnyValue, longhorn.AttacherTypeVolumeCloneController)
+
+		cloningAttachmentTicketID := longhorn.GetAttachmentTicketID(longhorn.AttacherTypeVolumeCloneController, v.Name)
+		if longhorn.IsAttachmentTicketSatisfied(cloningAttachmentTicketID, va) {
 			expectedAttachmentTickets[cloningAttachmentTicketID] = true
+		} else {
+			if srcNodeID == "" {
+				srcNodeID, err = pickNodeID(vol, va)
```
c3y1huang, line 454: > https://github.com/longhorn/longhorn-manager/pull/5062#discussion_r3715863643

Got it, but my question is only about the re-pick(reset) of srcNodeID. Given case 2 already picked N, can case 3 repick return a node other than N?

controller/volume_clone_controller.go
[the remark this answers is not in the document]
c3y1huang: IIUC, an unsatisfied ticket doesn't mean the current node is bad. The volume can be attached and healthy even if its node temporarily drops out of `attachableNodes`.

I agree on the satisfied case, but consider this potential unsatisfied case:
1. The source volume is attached on node A and serving I/O, backed by a satisfied ticket.
2. Node A temporary drops out of `attachableNodes`; engine keeps running.
3. A new clone target is created -> need new ticket. With the filter check, `pickNodeID` rejects A and picks node B.
4. The volume stays attached on A -> ticket on B not satisfied -> clone blocked.

So I think our choice is between:
1. Unconditional: a new ticket can briefly target a dying node -> failover clears the CurrentNodeID -> self-correct through the filter.
1. Filter check: a new clone can block silently for the duration of the flaps, indefinitly if the condition persists.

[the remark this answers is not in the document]
shuo-wu: > Unconditional: a new ticket can briefly target a dying node -> failover clears the CurrentNodeID -> self-correct through the filter.

I am worried about one race case: Once the clone ticket chooses the dying node and is just marked as satisfied. Then the Node down will corrupt the clone and make the case more unpredictable. 

> Filter check: a new clone can block silently for the duration of the flaps, indefinitly if the condition persists.

I am ok with having a looser check (checking if IM pod is running and not deleting only. Blocking a clone attachment is better than having a messy attachment & clone.

controller/volume_clone_controller.go
shuo-wu: > key forgotten. What re-enqueues the volume?

Nice catch. I am thinking if we need to add an informer so that it will enqueue all clone src & target volumes if `im.status.state` becomes `running`

controller/volume_clone_controller.go
shuo-wu: When the clone state is updated from init/in-progress (case 2) to CopyCompletedAwaitingHealthy (case 3), this reconcile loop will:
1. Skip case 2 logic since the check cannot pass now
2. Enter into case 3 handling. Then in line 391, `pickNodeID` will choose the CurrentNodeID, which is exactly what the case 2 VA ticket is using.
3. Remove the case 2 VA ticket in line 307

controller/volume_clone_controller.go
c3y1huang: Why not use [enqueueVolumeAfter](https://github.com/longhorn/longhorn-manager/pull/5065#discussion_r3725064007)?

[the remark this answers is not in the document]
c3y1huang: Ref https://github.com/longhorn/longhorn-manager/pull/5065#discussion_r3725071073

shuo-wu: @c3y1huang  Due to a race case, this controller still needs an IM informer. You can check the last comment message for details

c3y1huang (approved):

shuo-wu:
@mergify backport v1.12.x

derekbit (approved):
LGTM

PR #5066, opened 2026-08-05
fix(v2): reject attach when volume has no scheduled replicas 

c3y1huang:

Issue longhorn/longhorn#13655

See https://github.com/longhorn/longhorn/issues/13655#issuecomment-5187146472.

`None`

`None`

### DIFF
--- a/manager/volume.go
+++ b/manager/volume.go
@@ -245,6 +245,30 @@ func (m *VolumeManager) Delete(name string) error {
 	return nil
 }
 
+// validateVolumeHasScheduledReplica rejects attaching a volume that has no
+// scheduled replicas, since such a volume can never start.
+func (m *VolumeManager) validateVolumeHasScheduledReplica(v *longhorn.Volume) error {
+	// Sharded volumes place shards instead of replicas.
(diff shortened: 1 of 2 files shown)


c3y1huang:
Test result: https://ci.longhorn.io/job/private/job/longhorn-tests-regression/11666/

manager/volume.go
```diff
@@ -264,6 +288,10 @@ func (m *VolumeManager) Attach(name, nodeID string, disableFrontend bool, attach
 		return nil, err
 	}
 
+	if err := m.validateVolumeHasScheduledReplica(v); err != nil {
+		return nil, err
+	}
```
derekbit, line 293: Should we move the validation logic to the validating webhook (volume UPDATRE validation) instead?

c3y1huang: I think no. IIUC a ticket for a volume with no scheulded replica is a valid desired state, so admission is the wrong layer. This check only exists to fail the user's attach request. Is there a benefit moving to the webhook?

derekbit (approved):
LGTM

derekbit:
@mergify backport v1.12.x

